PR #1689, opened 2015-09-16
Add Python 3.5 support

timheap:
Python 3.5 has been released. I added 3.5 to tox/travis, ran the tests, and it worked!

Django 1.8 does not officially support Python 3.5, so I do not know if Python 3.5 support should be officially added at this point. I've set it as `allow_failure` in Travis. I do not know when Django 1.9 with Python 3.5 support will be coming out.


kaedroho:
Awesome, thanks!

PR #1684, opened 2015-09-15
Refactor out a bare except: statement

timheap:
It now catches `Redirect.DoesNotExist`, returning the normal 404 page if no redirect is found. Any other exception should not be caught here.

I found this while searching the code base for bare `except:` clauses, so this is semi-related to #1683

#1677 will have a merge-conflict with this one. I noticed bare `except:` clauses in the codebase when looking at that PR.

PR #1691, opened 2015-09-16
Remove bare except in unnecessary try block

timheap:
From a comment by @gasman on #1684:

> It appears the try/catch in ImageEmbedHandler was added here: 74b9f43
>
> Since the rest of the commit doesn't deal with images at all, and the
> commit makes it clear that the corresponding change to
> MediaEmbedHandler was intended to preserve the existing 'fail
> silently' behaviour in the light of the new exceptions added for media
> embeds - I think this try/catch is redundant. `Format.image_to_html`
> does its own catching of image IO errors, which seems to be
> sufficiently robust (if it wasn't, we'd be seeing errors on front-end
> page rendering), so I think this try/catch can reasonably be deleted.

https://github.com/torchbox/wagtail/pull/1684#issuecomment-140695060

### DIFF
--- a/wagtail/wagtailimages/rich_text.py
+++ b/wagtail/wagtailimages/rich_text.py
@@ -31,15 +31,12 @@ class ImageEmbedHandler(object):
         Image = get_image_model()
         try:
             image = Image.objects.get(id=attrs['id'])
-            format = get_image_format(attrs['format'])
-
-            if for_editor:
-                try:
-                    return format.image_to_editor_html(image, attrs['alt'])
-                except:
-                    return ''
-            else:
-                return format.image_to_html(image, attrs['alt'])
-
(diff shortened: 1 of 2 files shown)


gasman:
This has now broken the (mis-named) test_expand_db_attributes_for_editor_throws_exception test - this was relying on the try/catch to catch the error from a failed look-up of the 'alt' attribute. I see no reason why 'alt' would be missing in real data, since we only ever call `expand_db_attributes` on data that originated from `get_db_attributes` - so I think we can reasonably delete that test.

(If we really did want to guard against a missing 'alt' attribute, it would make more sense to do `attrs.get('alt', '')` rather than clobbering the tag entirely. But I don't think that's necessary.)

PR #1692, opened 2015-09-16
Create custom SQL for SQL Server. Yes, it had to be that ugly, but it…

FlipperPA:
… works. Falling on my sword over here...

### DIFF
--- a/wagtail/wagtailimages/migrations/0004_make_focal_point_key_not_nullable.py
+++ b/wagtail/wagtailimages/migrations/0004_make_focal_point_key_not_nullable.py
@@ -11,6 +11,19 @@ def remove_duplicate_renditions(apps, schema_editor):
                 SELECT CONCAT(image_id, '-', filter_id) FROM (SELECT * FROM `wagtailimages_rendition`) as x WHERE `focal_point_key` IS NULL GROUP BY image_id, filter_id HAVING COUNT(*) > 1
             ) AND `focal_point_key` IS NULL
         """)
+    elif schema_editor.connection.vendor == 'microsoft':
+        schema_editor.execute("""
+            DELETE FROM [wagtailimages_rendition] WHERE CAST(image_id AS VARCHAR(MAX)) + '-' + CAST(filter_id AS VARCHAR(MAX)) IN (
+                SELECT CAST(image_id AS VARCHAR(MAX)) + '-' + CAST(filter_id AS VARCHAR(MAX)) FROM [wagtailimages_rendition] WHERE focal_point_key IS NULL GROUP BY image_id, filter_id HAVING COUNT(*) > 1
+            ) AND focal_point_key IS NULL
+        """)
+        schema_editor.execute("""
+            DECLARE @constraint_name VARCHAR(MAX)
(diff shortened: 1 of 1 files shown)


gasman:
Thanks! I don't have the means to test it here, but it looks sensible enough. Would be interested to hear how you get on running Wagtail on SQL Server.

kaedroho:
This migration seems to be a source of pain for using Wagtail on databases we don't officially support.

If we add a squashed migration, new databases will be created with the unique constraint in place making this migration unnecessary.

tomdyson:
Thanks a lot, @FlipperPA! With this in place, do all of Wagtail's tests pass on SQL Server?

FlipperPA:
The test suite performs as expected (AFAIK):

Ran 1284 tests in 433.213s
OK (skipped=78, expected failures=11)

Let me know if you'd like more detail. I'd be happy to take the lead with providing an environment for testing SQL Server from a Linux database if you'd like to aim towards making it an "officially supported" database. At DjangoCon, Microsoft was hot-to-trot about getting SQL Server better support for Linux and Django in particular. As someone who is forced to use SQL Server some of the time, I may as well make sure everyone benefits from it!

Installing wagtail was a breeze, @gasman, other than this one migration, so I believe squashing the migrations would relieve a fairly major pain point, @kaedroho. Otherwise, we'd want to build support into the migration for SQLite as well, which is missing IIRC.

gasman:
Hi @FlipperPA - we're currently doing a sweep of the Wagtail codebase to enforce a maximum of 120 characters per line, and I just wanted to double-check that it's safe to reformat this code as shown below. The thing that's making me hesitate is that the second block of SQL contains multiple statements, with no delimiters other than the line breaks - am I right in thinking that we can safely add extra line breaks and SQL Server will just deal with them (i.e. we're not relying on having exactly one statement per line)?

    elif schema_editor.connection.vendor == 'microsoft':
        schema_editor.execute("""
            DELETE FROM [wagtailimages_rendition]
            WHERE CAST(image_id AS VARCHAR(MAX)) + '-' + CAST(filter_id AS VARCHAR(MAX)) IN (
                SELECT CAST(image_id AS VARCHAR(MAX)) + '-' + CAST(filter_id AS VARCHAR(MAX))
                FROM [wagtailimages_rendition]
                WHERE focal_point_key IS NULL
                GROUP BY image_id, filter_id
                HAVING COUNT(*) > 1
            ) AND focal_point_key IS NULL
        """)
        schema_editor.execute("""
            DECLARE @constraint_name VARCHAR(MAX)
            SELECT @constraint_name = CONSTRAINT_NAME
                FROM INFORMATION_SCHEMA.TABLE_CONSTRAINTS
                WHERE TABLE_NAME='wagtailimages_rendition' AND CONSTRAINT_TYPE='UNIQUE'
            EXEC('ALTER TABLE [wagtailimages_rendition] DROP CONSTRAINT ' + @constraint_name)
            ALTER TABLE [wagtailimages_rendition] ALTER COLUMN [focal_point_key] NVARCHAR(255) NOT NULL
            EXEC('ALTER TABLE [wagtailimages_rendition] ADD CONSTRAINT ' + @constraint_name +
                ' UNIQUE NONCLUSTERED (image_id, filter_id, focal_point_key)')
        """)

FlipperPA:
Thanks for checking in. That should be fine - I just ran a lazy test against a dev instance on SQL Server 2014 with a copy and paste, editing the Python bits. It ran fine.

And yes, SQL Server is ugly, ugly, ugly. Some of the worst ergonomics I've seen in a product.

PR #1696, opened 2015-09-17
Fixes #1549: StreamField blocks are now added automatically if there's only one block type

### LINKED ISSUE #1549: Don't display the huge block selection box if a StreamField can contain only one type of block
The plus button that shows the block selection box in a StreamField should just add the block if only one type of them can be added.

alexgleason:
In the admin, clicking the + button when a StreamField menu is closed will now automatically add a block if there's only one possible type of block.

Fixes #1549


davecranwell:
Hi @alexgleason, thanks for tackling this! If you could lint your code (http://docs.wagtail.io/en/v1.0/contributing/javascript_guidelines.html) to solve the Javascript code style issues, that would be great (Wagtail uses 4-space indenting throughout)

alexgleason:
OMG that is awesome... I've never seen that before, mind blown. Thanks!

PR #1705, opened 2015-09-17
Initial idea for a table stream field block using handsontable.js.

Proper-Job:
Hi,
This is an initial implementation of a StreamField table block using handsontable.js.
Let me know if there is anything else missing or gone totally astray.

Missing is documentation as to the available table configuration options.
If you like this initial approach I'll document it more clearly and completely.

No tests have been written yet.


chrxr:
Hi @Proper-Job, thanks for this! It works very well in testing. 

As opposed to implementing this in core, we'd propose refactoring this as a pip installable module. [See this markdown module](https://github.com/torchbox/wagtail-markdown) as a good example.

Could you implement in a similar way?

Cheers,
Chris

gasman:
Hi @chrxr - I think the suggestion was for this to become a 'contrib' module within the Wagtail codebase (e.g. `wagtail.contrib.table_block`) rather than making it a completely separate pip-installable package.

chrxr:
Hi @Proper-Job, we've discussed this in the team, and we'd like to go down the contrib route as suggested by @gasman above. Could you please update this PR. Thanks!

Proper-Job:
Just came back from my holiday :)
Yes, I will refactor it into the the contrib module.
Thanks and greetings
Moritz

Proper-Job:
I'm having a bit of a problem to integrate the git wagtail installation with PyCharm.  Hence this ridiculous commit about ununsed imports.

Proper-Job:
I squashed all of my commits.  Is that alright to have all the logic in __init__.py, or do you want it in modules?

gasman:
`__init__.py` is fine by me. Will give it a full review once the 1.2 release is out of the way - thanks!

gasman:
Hi @Proper-Job - this is looking good! Just a few small (I hope) issues:

* It's possible to insert arbitrary HTML tags into cell text and have them interpreted as HTML. This needs to be changed so that any `<` `>` characters are displayed as-is in the output (i.e. the `safe` filters in `templates/table_block/blocks/table.html` should be removed, and the Handsontable behaviour changed so that it doesn't convert the tags, as in the examples at http://handsontable.com/examples.html).
* The table width in the editor is too wide - it pushes the 'delete' button in the block header off the edge of the window.
* Can we incorporate the JS from RichTextBlock so that the block height is automatically adjusted to fit the table height?

Once these are sorted, and we have tests and docs, I think we'll be ready to merge this. Thanks!

seddonym:
Great addition!

One thing I noticed with this field is that it doesn't support tables with a vertical column too.

This could be done really simply by adding a 'First column header' checkbox and tweaking the template so it outputs <th> for the first item in each row.  Let me know if you'd like me to make a pull request for this.

Proper-Job:
Sorry guys, for some reason I wasn't notified of your comments, for some reason that keeps happening to me.  Let me review your comments and get back to you.

Proper-Job:
@gasman It is possible to change HTML renderer to a simple text renderer for the table.  All you would need for that is to pass the renderer at construction time like so:
```python
table = TableBlock(table_options={'renderer':'text})
```
I could default it to text if you like.  Should I do that?
I will check out the width problem.
I will try and incorporate the width js, too.

What kind of tests do you want to see for this?
Where do you want the docs to go?

Proper-Job:
@seddonym Yes please do that.

seddonym:
Hmm, I've to create a pull request, but I don't think Github supports multiple forking - I already have a fork of Wagtail in my account, so I can't fork from yours.  I've manually forked it, but I'm not able to create a pull request back into your repo.

Probably easiest is if you merge my commits yourself from here:
https://github.com/seddonym/wagtail-tableblock/tree/streamfield-table 

If you can think of another way to do it, let me know!

Proper-Job:
@seddonym Thank you for that!
I merged your commits into this branch.

seddonym:
Excellent, thanks!

Proper-Job:
@seddonym I was wondering if you had the capacity to help out in any way to get this feature landed in master.
Does any one of the tasks that gasman assigned jump out at you?
I'm spread pretty thin atm but I'd really like to see this feature land asap.

seddonym:
Hi @Proper-Job, I had a quick look but nothing jumps out as a something I can do quickly without spending a bit of time getting my head around the code.  I'm afraid I'm also a bit too busy at the moment - sorry!

Proper-Job:
@seddonym thanks for having a look!
I'll try to get it done asap.
Greetings

akhayyat:
Has there been any progress more recently on this effort? A table block would be awesome!

Proper-Job:
@akhayyat I think the earliest I can address the issues gasman pointed out is in May.  I'd be glad to accept any help though, because I'd really like to see this land.

Proper-Job:
@akhayyat I think the earliest I can address the issues gasman pointed out is in May.  I'd be glad to accept any help though, because I'd really like to see this land.

bbusenius:
Hi everyone! I'm working with @Proper-Job on trying to fix the issues outlined by @gasman. Looking at the first issue, it's easy to remove the safe filter to address the security concern, however, might it be okay to use a custom renderer as in this example: https://docs.handsontable.com/pro/1.3.3/demo-custom-renderers.html? This would at least allow for the ability to use bold, italics, and anchors when needed. What do you think?

gasman:
Hi @bbusenius,
I'm against allowing HTML in table cells, even in a limited form. The whole rationale for StreamField is that it captures something that's as close as possible to pure data, without specifying any particular presentation of that data - that's the job of the template. For example, it would be entirely possible for a template to take the data that's been entered into a TableBlock, and render it as a bar chart.

(Admittedly, TableBlock might not be the best tool for that job - ideally you'd want something that enforces the correct data types, rather than allowing any arbitrary string value in any cell. But nevertheless, it should still be possible to use TableBlock in that way.)

Once you start allowing HTML tags in table content, you're effectively saying "every cell value shall be treated as a fragment of HTML". (You might think I'm exaggerating there - you can still store non-HTML data in there if you want, right? Well, no - as soon as you allow for the _possibility_ of a cell being handled as HTML, you need a way to distinguish between cells that _are_ HTML content, and ones that just happen to be strings containing literal `<` and `>` characters. And the only non-ambiguous way to do that is to make _everything_ HTML.) At that point, TableBlock isn't a way of capturing pure data any more - it's just a layout tool for HTML fragments, and that's a misuse of StreamField.

bbusenius:
Ah, I see. That definitely makes sense and I can definitely see your reasoning. What would you recommend then for cases where you need to layout tabular data with rich text? We have lots of tables like this: https://www.lib.uchicago.edu/e/using/access/researchers.html and our authors have the need to inject such tables into random pages. I appreciate your input!

gasman:
@bbusenius I wish I had a good answer for that! Even disregarding the constraints that Wagtail is working under (both the technical ones of being a browser-based interface, and the 'ideological' ones of separating data and presentation), coming up with a suitable UI for authoring table-based content like this would be a very hard problem indeed. Handsontable's HTML capabilities wouldn't really solve this acceptably either: Wagtail's internal representation of rich text data is not simple HTML, but a pseudo-HTML format allowing us to make symbolic references like `<a linktype="page" id="123">another Wagtail page</a>`, and we'd surely want our rich-text-in-tables functionality to support that feature (otherwise, it's not really much of a step up from giving the author a big text box to enter raw `<table>` HTML into). Essentially, we'd be looking at a way to embed `RichTextBlock` in a 2D grid layout, in much the same way that `ListBlock` allows you to repeat `RichTextBlock`s in one dimension...

(In a sense, you _can_ do that in Wagtail right now, by defining a `ListBlock` of `ListBlock`s of `RichTextBlock`s. But the editing interface for that will look absolutely horrible, so let's not go there :-) )

Considering the example from https://docs.handsontable.com/pro/1.3.3/demo-custom-renderers.html for a moment - even though that's a more complex setup, with embedded images as well as rich text, it's actually clearer to see what the recommended Wagtail-ish approach is. The table represents a list of books, so it should be modelled as one:

    class BookBlock(StructBlock):
        title = CharBlock()
        link = URLBlock()
        description = RichTextBlock()
        comments = RichTextBlock()
        cover_image = ChooserBlock()

    ...

    body = StreamField([
        ...
        ('book_table', ListBlock(BookBlock()),
        ...
    ])

Tenuously, you could do something similar with your table:

    class LibraryPrivilegeBlock(StructBlock):
        affiliation = CharBlock()
        who_is_eligible = CharBlock()
        kind_of_privileges = RichTextBlock()
        action_needed = RichTextBlock()

- but in both cases, this relies on these models being 'baked in' to the site's business logic - there's no scope for content authors to define this sort of thing in an ad-hoc way.

(Hypothetically, I could imagine us having some kind of TypedTableBlock where an author could choose the block types for each column, in much the same way that they're presented with a selection of block types when populating the StreamField itself - so effectively they're doing the data modelling within the page editor itself. This is probably as close as we could get to a solution that ticks all the boxes while sticking to Wagtail's pure-data philosophy - but I suspect it wouldn't be particularly friendly to a content author who doesn't care about data modelling and wants to create "just a table"...)

I guess this can be summed up as follows: Creating a *good* UI for editing rich, table-based content is extremely hard, to the extent that I suspect that no CMS anywhere has solved it. For that reason, Wagtail's strategy is to side-step it wherever possible:

* Is it a presentational table? If so, it's the responsibility of the template designer, not the content author
* Is it an exercise in data modelling? If so, it's the responsibility of a Python developer, not the content author

Unfortunately, if it really doesn't fit into either of those categories - i.e. it's "just a table" - then there may be no alternative but to tackle the extremely hard problem head-on, which probably won't happen any time soon (and is certainly out of scope for this ticket).

bbusenius:
Hi @gasman! Thanks for the thoughtful response and thanks for all the detail! Everything you said makes perfect sense. We'll certainly have to come up with a solution internally, even if it's a kludge, however, we can table that discussion. I agree it's out of scope for this ticket. 

Of the outstanding blockers you listed above, I believe I've solved the first two with simple tweaks to the code. I'll have to clear the work with @Proper-Job but I think we're close. The third thing you mentioned:

> Can we incorporate the JS from RichTextBlock so that the block height is automatically adjusted to fit the table height?

is proving a little more involved. I've never worked on wagtail directly before, so, it's just a matter of learning the code. I'll let you know if I have any specific questions.

Thanks for your feedback and support!

bbusenius:
@Proper-Job and I have worked through the issues outlined by @gasman including unit tests. What's remaining is documentation. Can anyone tell me what's needed on the documentation front?

bbusenius:
[This is what we have for documentation](https://docs.google.com/document/d/1AtLAyMP6jHsJrEM87Ab2dGgH2eQEhoIG0TZku7VbNew/edit?usp=sharing) at the moment. It's written in a hybrid rst format to make it easier to read on the web. We thought it would be a good idea to take comments before rolling it into the code base. If this is acceptable, we'll create a new table_block.rst in  /wagtail/docs/reference/contrib/ and link it in docs/reference/contrib/index.rst. Please let us know if we're missing anything or if we should provide additional documentation files. This is just a rough draft. We'll go over it a few times before we commit.

tomdyson:
Thanks, @bbusenius and @Proper-Job! I've made some minor tweaks in the Google doc but it's looking good so far. Perhaps you could add a screenshot demonstrating how table_block looks in the editor?

bbusenius:
Thanks @tomdyson. A screenshot showing table_block in the editor is a great idea! We'll add one.

bbusenius:
One thing that's missing that would be nice, is a sensible default icon. I don't see anything usable in the wagtail style guide but [font awesome has this](http://fontawesome.io/icon/table/). What do you guys think?

Proper-Job:
@gasman @tomdyson:  All the issues outlined above have been addressed by @bbusenius and the PR was updated.  
There is no default icon for the table block, yet.  What do you think of @bbusenius suggestion above?
Would you like us to integrate a default icon?
If not, could you please review the changes?

gasman:
Thanks! Will review shortly. The suggested icon looks good to me... (It seems we don't have any documentation for adding new icons to our font, but I believe it's done through https://icomoon.io/.)

Proper-Job:
@gasman Just to double check.  Would you like a new icon to be part of this PR? Or is that something you would like to address internally as there are no icon contributing guidelines?

bbusenius:
I've been doing some preliminary research on the icons and my first impression is that this might best be handled by an internal wagtail design person. Using icomoon looks straightforward but how to integrate a new icon into a preexisting icon set is escaping me. From what I can tell, it looks like a set of icons would be generated locally and then uploaded to the project. I imagine the wagtail design team has a streamlined process for this. Please correct me if I am wrong.

gasman:
OK, no worries - we'll handle it internally. (We've got another icon coming imminently in #2417, so we'd probably end up having to resolve the merge conflict anyhow :-) )

gasman:
Merged in 9338fc2ac4ae984437aae6e37a506ad0f4507322 + parents (without the icon - will handle that separately). Thanks all!

alexgleason:
I needed to use this on Wagail 1.4, so I made a simple subtree split of the directory in master, restructured it, and released it as a pip installable package: `wagtail-table-block`.

https://github.com/alexgleason/wagtail_table_block

It says clearly in the readme to NOT use it if you can help it. But the database representation should be the same regardless of where it's coming from, so upgrading to Wagtail 1.5 should allow you to just delete the plugin and import from wagtail.contrib.

Just in case anyone else needs this.

alexgleason:
Question: this uses `from wagtail.contrib.table_block.fields import TableBlock`

Do we need that `.fields`? Can it just be `from wagtail.contrib.table_block import TableBlock`?

gasman:
@alexgleason Looks like this was moved out of `__init__` in 2d818afe62b273cd7fdf378018d7ea657c605fc0, in order to avoid problems on startup. I don't think the module naming is a big deal really.

bbusenius:
There's an  an issue with some of the logic being in __init__.py in Django 1.9. It throws an AppRegistryNotReady exception. The easiest way we could find to fix it was to move the code into a module.

Proper-Job:
We chose to name the module fields.py because in the future the might be a TableField in addition to the TableBlock.  In the absence of TableField it might be prudent to name the module blocks.py.  I don't really mind either way.

PR #1709, opened 2015-09-18
Set HTTP_HOST in dummy_request for preview - fixes #1703

### LINKED ISSUE #1703: Previewing a Page, uses wrong Site
Having two sites which are linked to two different Root Pages respectively, when Previewing a page under a root page, the wrong Site instance is attached to the request, ultimately resulting in our sites displaying the wrong css.

`response = page.serve_preview(page.dummy_request(), preview_mode)`

On serving a request to preview a page, a dummy request is made looking like the following

```
request = WSGIRequest({
            'REQUEST_METHOD': 'GET',
            'PATH_INFO': path,
            'SERVER_NAME': hostname,
            'SERVER_PORT': port,
            'wsgi.input': StringIO(),
        })
```

then the middleware is run, but in the SiteMiddleware, and subsequently the Site model, it looks for a different key in the request

`hostname = request.META['HTTP_HOST'].split(':')[0]`

This results in the default Site been chosen

gasman:
fixes #1703

### DIFF
--- a/wagtail/wagtailadmin/tests/test_pages_views.py
+++ b/wagtail/wagtailadmin/tests/test_pages_views.py
@@ -16,7 +16,7 @@ from wagtail.tests.testapp.models import (
     BusinessIndex, BusinessChild, BusinessSubIndex,
     TaggedPage, Advert, AdvertPlacement)
 from wagtail.tests.utils import WagtailTestUtils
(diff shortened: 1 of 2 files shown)


kaedroho:
merged in: 9e28bee22b999dce21d7384da646900c65afab1b

Issue #1644, opened 2015-08-26
Add a system check to verify that Pillow / Wand has JPEG / PNG / GIF support compiled in


gasman:
(moved from https://github.com/torchbox/wagtaildemo/issues/12)

It's easy to miss the fact that libjpeg/zlib should be installed prior to installing Pillow, and this will result in a build of Pillow without JPEG/PNG support, and hard-to-identify issues with images not displaying.

We should add a system check on server startup to warn if the image library in use (Pillow or Wand) is missing support for these formats.

FlipperPA:
Quick question, by server startup do you just mean on runserver start up only, or also a hook into gunicorn / uwsgi / mod_wsgi? By the look of the closed issue, it looks like runserver only, but wanted to check before diving in.

gasman:
@FlipperPA That's a good point... I was imagining that we'd use Django's system check framework <https://docs.djangoproject.com/en/1.8/topics/checks/> for this, to display a warning on the console - but that's not so useful for server-level issues like this, where everything might be perfectly fine in development and only be broken on the live server, where those warnings won't be seen. (Since we've mostly seen this coming up in the context of new users installing wagtaildemo, I hadn't considered the case of live server deployments at all. But catching the issue there is just as important, if not more...)

Perhaps, then, it would be more appropriate to display a warning bar on the admin homepage when the user is logged in as a superuser.

gasman:
I've implemented the first half of this in #1718 - this adds the system check, but doesn't yet display a message on the admin dashboard.

gasman:
Pillow 3.x makes it hard to get this wrong now - at this point I don't think it's worth the trouble of prettifying the error message.

PR #1718, opened 2015-09-23
Add system check for presence of libjpeg / zlib

gasman:
Partially addresses #1644 (doesn't include displaying a message on the admin dashboard yet)


wagtail/wagtailimages/checks.py
```diff
@@ -0,0 +1,72 @@
...
+def has_jpeg_support():
+    global _has_jpeg_support
+
+    if _has_jpeg_support is None:
+        wagtail_jpg = os.path.join(os.path.dirname(__file__), 'check_files', 'wagtail.jpg')
+        succeeded = True
+        f = open(wagtail_jpg, 'rb')
+
+        try:
+            Image.open(f)
```
kaedroho: slightly nicer:
```
with open(wagtail_jpg, 'rb') as f:
    Image.open(f)
```

This will close the image file in the event of an exception as well

gasman:
Suggested clean-ups now added.

kaedroho:
Pillow 3.0 now requires these two libraries https://github.com/python-pillow/Pillow/pull/1439

SalahAdDin:
:+1: I will have to install this libraries in my server and local developments.

kaedroho:
Running pip install pillow on a fresh environment will raise the following exception:
```
  File "/tmp/easy_install-iNjVIv/Pillow-3.0.0/setup.py", line 515, in build_extensions
    
ValueError: --enable-jpeg requested but jpeg not found, aborting.
```

There's apparently a check for ``zlib`` (PNG) as well

kaedroho:
pip 7 caches builds meaning that ``pip uninstall pillow && pip install pillow<3.0.0`` will no longer work. You have to add ``--no-cache-dir`` to the end of the ``pip install``

SalahAdDin:
Will be great put these clarification in the docs! :+1:

gasman:
@kaedroho So is it your opinion that there's no real need for this check now, given that Pillow 3.0.0 makes it effectively impossible to build a non-JPEG/PNG-enabled version by mistake? Happy to go along with that if so! (In which case, how about we update setup.py to require Pillow 3.0.0?)

gasman:
Have agreed to bump Pillow requirement to 3.0.0.

shredding:
Please double check with https://github.com/python-pillow/Pillow/issues/1474#issuecomment-151788788

I had 3.0.0 in combination with wagtail and had to force a downgrade to 2.9.0 because of that.

gasman:
@shredding Ah, fair enough! If Pillow 3.0.0 still has some rough edges then I think we should hold off on requiring it, and go ahead with this check in the meantime. @kaedroho?

kaedroho:
I'm happy with that :shipit:

PR #1719, opened 2015-09-23
Document custom queryset required approach

nimasmi:
This may be OTT, but I ran into another example of where something that works in Django doesn't work quite the same way in Wagtail. Here's some documentation.

Ehkä tämä on liiallinen, vaan löysin jotakin jossa sopii Djangoille muttä ei sopi Wagtailille. Tässä on dokumentointia.

### DIFF
--- a/docs/topics/creating_pages.rst
+++ b/docs/topics/creating_pages.rst
@@ -98,3 +98,46 @@ This is because ``Page`` enforces ordering QuerySets by path. Instead you must a
 .. code-block:: python
 
     news_items = NewsItemPage.objects.live().order_by('-publication_date')
+
+Page custom managers
+--------------------
+
+``Page`` enforces its own 'objects' manager in its ``__init__`` method, so you cannot add a custom manager at the 'objects' attribute.
+
+.. code-block:: python
+
+    class EventPageQuerySet(PageQuerySet):
+
+        def future(self):
+            return self.filter(
+                start_date__gte=timezone.localtime(timezone.now()).date()
+            )
+
+    class EventPage(Page):
+        start_date = models.DateField()
+
+        objects = EventPageQuerySet.as_manager()  # will not work
+
+To use a custom manager you must choose a different attribute name. Make sure to subclass ``wagtail.wagtailcor.models.PageManager``.
+
+.. code-block:: python
+
+    from django.db import models
+    from django.utils import timezone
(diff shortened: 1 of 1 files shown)


docs/topics/creating_pages.rst
```diff
@@ -98,3 +98,46 @@ This is because ``Page`` enforces ordering QuerySets by path. Instead you must a
...
+            return self.filter(
+                start_date__gte=timezone.localtime(timezone.now()).date()
+            )
+
+    class EventPage(Page):
+        start_date = models.DateField()
+
+        objects = EventPageQuerySet.as_manager()  # will not work
+
+To use a custom manager you must choose a different attribute name. Make sure to subclass ``wagtail.wagtailcor.models.PageManager``.
```
gasman: typo - wagtailcore

kaedroho: Missing an 'e' on ``wagtailcore``

timheap:
We came across this issue as well, and solved it by overriding the overridden `objects` with our own `objects` in a custom page metaclass, which inherits from the `PageBase` metaclass. It is multiple levels of things overriding each other and clobbering past assignments, which makes it particularly fun...

Adding it here in case it helps anyone. Would this be worth documenting, or is this too many levels of WTF hackery to be officially supported?

```python
import six

from django.db import models
from wagtail.wagtailcore import models as wagtailmodels


class MyPageBase(wagtailmodels.PageBase):

        # Clobber the forced attribute set in PageBase
        if 'template' not in dct:
            cls.template = "layouts/%s/%s.html" % (
                cls._meta.app_label, cls._meta.model_name)

        # Clobber the forced manager set in PageBase
        MyPageManager().contribute_to_class(cls, 'objects')


class MyPage(six.with_metaclass(MyPageBase, wagtailmodels.Page)):
    # Disable the creation of the related field accessor on Page
    # This prevents conflicts with model names and field names
    page_ptr = models.OneToOneField(wagtailmodels.Page, parent_link=True,
                                    related_name='+')

    # Set this here so autocompletion works. It will be clobbered by PageBase,
    # and then reset again by MyPageBase
    objects = PageManager()

    class Meta:
        abstract = True
```

timheap:
See the above PR for a fix for this. Now you should be able to define custom Managers in the normal Django way.

Issue #1724, opened 2015-09-23
Need contributing guidelines for browser support


alexgleason:
Hi again. I'm wanting to make some contributions to Wagtail's admin on the front-end, but need to know what browsers it intends to support so I can plan accordingly. I wasn't able to find this information in the [contributing guidelines](http://docs.wagtail.io/en/v1.1/contributing/index.html).

For instance, I want to use an inset CSS box-shadow to give the impression of an inner border, but this won't be the best solution if Wagtail intends to support IE8 since [box-shadow is an unsupported property in IE8](http://caniuse.com/#search=box-shadow).

Thanks!

kaedroho:
Not sure if support starts at IE9 or IE10 (@davecranwell should be able to confirm). I'm pretty certain that Wagtail does not support IE8 though.

davecranwell:
It's IE9+

nealtodd:
(Django Admin [drops IE8 support](https://docs.djangoproject.com/en/1.9/releases/1.9/#contrib-admin-browser-support) in 1.9, making Wagtail Admin consistent with that)

FlipperPA:
I noticed the drop of IE8 support in the release notes yesterday. Is Wagtail committed to support the same browsers that the Django Admin does? That seems like a good goal.

davecranwell:
Wagtail was originally written for a specific client website where we were only supporting IE9+. As a result Wagtail has never supported less than IE9. I'm broadly in favour of matching Django Admin in terms of the lowest support level, although we've not committed to that officially.

thibaudcolas:
I've had a stab at this in #3337 for end users, and more importantly #3338 for the contribution guidelines. Keen to get feedback on that PR.

For IE9, I wanted to see for myself so I opened up a VM. It works, mostly, but there are issues. I made a list below.

| Issue in IE9 | In what context | Origin / clues |
|--------------|-----------------|----------------|
| On the login form, placeholders for username and password do not show. This is problematic because there are no labels on those fields. | `/admin/login` | `placeholder` is not supported in IE9 |
| In the sidebar, the placeholder for the Search field does not show. Problematic because this field does not have a label. | `/admin/` | `placeholder` is not supported in IE9 |
| In the page list view, the human-readable date does not display the absolute date on hover, flickering instead. | `/admin/` and other views | ? |
| In the header, the collapsed "mobile" menu does not open when clicking the hamburger icon. | `/admin/` on mobile (all pages?) | ? | 
[...]

PR #1725, opened 2015-09-23
Fixed StreamField hover jitter

alexgleason:
When you hover over a StreamField block, it creates a 1px blue border which causes it to jitter. This solution uses ~~an inset box-shadow instead~~ an initial transparent border so that hovering produces no jitter.

~~If you don't like this, an alternative solution is to let the block always have a 1px transparent border, then the color of the border changes on hover. I felt the box-shadow solution would be cleaner, but the [browser compatibility](http://caniuse.com/#search=box-shadow) isn't as good. (#1724 what browsers should we support?)~~

Note that my text editor automatically stripped out some whitespace in this commit, hence the other lines being altered.

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/streamfield.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/streamfield.scss
@@ -5,12 +5,13 @@ ul.sequence{
 li.sequence-member{
     @include clearfix;
     position:relative;
+    border-style: solid;
+    border-color: transparent;
+    border-width: 1px 0;
 
     &:hover{
         background-color:$color-input-focus;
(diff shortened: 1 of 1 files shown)


davecranwell:
Thanks for working on this @alexgleason. I'd rather we use the border transparency solution that shadow, if you wouldn't mind swapping it around?

alexgleason:
I've never heard of Travis CI before contributing to this project. I checked out the log but there's no reason I can tell that it failed. I'm guessing it's a fluke?

gasman:
I've set the test running again. We seem to have had a few spurious random errors coming from elasticsearch lately...

gasman:
Merged in a89b981a4754d46a547e076efbbb016320128b59

PR #1728, opened 2015-09-23
Removed broken link to non-existent vagrant install script

alexgleason:
On here: http://docs.wagtail.io/en/v1.0/advanced_topics/deploying.html

The install.sh link is broken since that's not being used anymore. It looks like that stuff is actually built into the Vagrant box now, so there is no longer a reference. If there's an alternate reference that would be nice, but it should probably just be removed.


kaedroho:
Thanks!

This wasn't previously used for Vagrant setup, but as an example for those deploying a Wagtail site to a production server.

PR #1729, opened 2015-09-24
Allow custom Page Managers

timheap:
Previously, if a developer wanted to use a custom Manager on their Page subclass, some fairly hacky hacks were required. Now, the `objects` attribute is only overridden if it is a plain `Manager`. If it is anything else, it is left alone. A system check has been added to ensure that all `Page` managers inherit from `PageManager`

Inspired by #1719

### DIFF
--- a/docs/topics/pages.rst
+++ b/docs/topics/pages.rst
@@ -440,26 +440,25 @@ This is because ``Page`` enforces ordering QuerySets by path. Instead you must a
 
     news_items = NewsItemPage.objects.live().order_by('-publication_date')
 
-Page custom managers
+Custom Page managers
 --------------------
 
 ``Page`` enforces its own 'objects' manager in its ``__init__`` method, so you cannot add a custom manager at the 'objects' attribute.
 
 .. code-block:: python
 
-    class EventPageQuerySet(PageQuerySet):
+    from django.db import models
+    from wagtail.wagtailcore.models import Page, PageManager
 
-        def future(self):
-            return self.filter(
-                start_date__gte=timezone.localtime(timezone.now()).date()
-            )
+    class EventPageManager(PageManager):
+        """ Custom manager for Event pages """
 
     class EventPage(Page):
         start_date = models.DateField()
 
-        objects = EventPageQuerySet.as_manager()  # will not work
+        objects = EventPageManager()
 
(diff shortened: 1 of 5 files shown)


docs/topics/pages.rst
```diff
@@ -440,26 +440,25 @@ This is because ``Page`` enforces ordering QuerySets by path. Instead you must a
 
     news_items = NewsItemPage.objects.live().order_by('-publication_date')
 
-Page custom managers
+Custom Page managers
 --------------------
 
 ``Page`` enforces its own 'objects' manager in its ``__init__`` method, so you cannot add a custom manager at the 'objects' attribute.
```
gasman: I assume this sentence should be removed now?

I'm also not clear on why we need two bits of example code here, although that might just be me being rusty on this area of Django - it looks like we've changed from "you can't do X, but you can do Y instead" to "you can do X, and you can also do Y" with no explanation of the difference between the two or how they differ from standard Django behaviour. (Is it specifically the fact that they should inherit from PageManager? If so, we should highlight that in the text.) My impression is that managers and querysets are pretty much interchangeable now that Manager.from_queryset exists, so perhaps we can get away with a single example.

Filtering EventPage to just future events strikes me as a slightly dubious thing to do within the *default* manager - perhaps the example(s) could demonstrate adding a `.future()` method to the manager/queryset instead?

gasman:
Other than the above documentation quibbles, happy with this!

docs/topics/pages.rst
timheap: I had missed updating some of that documentation, you're right. I've rewritten those lines to actually be relevant now.

The two bits of example code are useful, as they do two different things. Making a custom Manager is a different thing to making a custom QuerySet, although they look quite similar. You have to inherit from different parent classes, and construct the Manager instance in a different manner, so documenting both approaches is required in my opinion.

The code for filtering for future events is already in a `future()` method.

gasman: > The code for filtering for future events is already in a future() method.

So it is, sorry - I must have gone cross-eyed looking at the diffs :-) Thanks for the clarifications!

gasman:
Merged in 9e8c2c2d5ffb816fda4152e69aded773691fcc39.

PR #1730, opened 2015-09-24
Deprecate automatic `template` attribute, use `get_template`

mx-moth:
Instead of automatically generating a `template` attribute in the PageBase metaclass - which made it very difficult to override - the automatic template name generation is now done inside `get_template()`.

`template` and `ajax_template` have been made in to a `@property`, which raises a RemovedInWagtail13Warning when accessed. They can still be overridden in Page subclasses to hardcode the template name manually. In Wagtail 1.3, the `template` and `ajax_template` attributes will be set to `None`.

`AbstractForm` pages had a `landing_page_template` attribute that was generated in `__init__` using the `template` attribute. A `is_landing_page` parameter has been added to `get_template` which automatically generates this landing page template name instead.

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -2,9 +2,10 @@ from __future__ import absolute_import, unicode_literals
 
 import json
 import logging
+import warnings
 from collections import defaultdict
-from django import VERSION as DJANGO_VERSION
 
+from django import VERSION as DJANGO_VERSION
 from django.conf import settings
 from django.contrib.auth.models import Group, Permission
 from django.contrib.contenttypes.models import ContentType
@@ -31,6 +32,7 @@ from django.utils.translation import ugettext_lazy as _
 from modelcluster.models import ClusterableModel, get_all_child_relations
 from treebeard.mp_tree import MP_Node
 
+from wagtail.utils.deprecation import RemovedInWagtail110Warning
 from wagtail.wagtailcore.query import PageQuerySet, TreeQuerySet
 from wagtail.wagtailcore.signals import page_published, page_unpublished
 from wagtail.wagtailcore.url_routing import RouteResult
@@ -261,12 +263,6 @@ class PageBase(models.base.ModelBase):
             # don't proceed with all this page type registration stuff
             return
 
-        if 'template' not in dct:
-            # Define a default template path derived from the app name and model name
-            cls.template = "%s/%s.html" % (cls._meta.app_label, camelcase_to_underscore(name))
-
-        if 'ajax_template' not in dct:
-            cls.ajax_template = None
 
         cls._clean_subpage_models = None  # to be filled in on first call to cls.clean_subpage_models
         cls._clean_parent_page_models = None  # to be filled in on first call to cls.clean_parent_page_models
@@ -760,11 +756,48 @@ class Page(six.with_metaclass(PageBase, AbstractPage, index.Indexed, Clusterable
             'request': request,
         }
 
+    @property
+    def template(self):
+        warnings.warn(
+            "The automatically generated template attribute is deprecated - "
+            "use Page.get_template() or set an explicit template attribute .",
+            category=RemovedInWagtail110Warning, stacklevel=2)
+        return "%s/%s.html" % (self._meta.app_label, camelcase_to_underscore(self._meta.object_name))
+
+    @property
+    def ajax_template(self):
+        warnings.warn(
+            "The automatically generated ajax_template attribute is deprecated - "
+            "use Page.get_template() or set an explicit ajax_template attribute .",
+            category=RemovedInWagtail110Warning, stacklevel=2)
+        return "%s/%s.html" % (self._meta.app_label, camelcase_to_underscore(self._meta.object_name))
+
     def get_template(self, request, *args, **kwargs):
-        if request.is_ajax():
-            return self.ajax_template or self.template
-        else:
+        """
+        Get the template name for this Page class.
+
+        * If `request.is_ajax()`, and the Page has an `ajax_template`
+          attribute, that template will be used.
+        * Otherwise, if the Page has a `template` attribute, that will be used.
+        * Otherwise, a template name of `"app/model_name.html"` will be
+          generated.
+        """
+        # RemovedInWagtail110Warning: `is_auto` and `is_ajax_auto`.
+        # `template` and `ajax_template` attributes will just be set to None
+        # in Wagtail 1.10, and the `is_auto` and `is_ajax_auto` variables
+        # removed.
+        is_auto = type(self).template is Page.template
+        is_ajax_auto = type(self).ajax_template is Page.ajax_template
+
+        # If the request is AJAX, and an `ajax_template` has been set
+        if request.is_ajax() and not is_ajax_auto and self.ajax_template is not None:
+            return self.ajax_template
+        # Otherwise, if `template` has been set
+        elif not is_auto and self.template is not None:
             return self.template
+        # Otherwise, generate a template name
+        else:
+            return "%s/%s.html" % (self._meta.app_label, camelcase_to_underscore(self._meta.object_name))
 
     def serve(self, request, *args, **kwargs):
         request.is_preview = getattr(request, 'is_preview', False)
--- a/wagtail/wagtailforms/models.py
+++ b/wagtail/wagtailforms/models.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, unicode_literals
 
 import json
 import os
+import warnings
 
 from django.contrib.contenttypes.models import ContentType
 from django.core.serializers.json import DjangoJSONEncoder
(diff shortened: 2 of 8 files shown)


kaedroho:
Thanks @timheap 
I did something similar in https://github.com/torchbox/wagtail/pull/1464. Your PR handles deprecation better though so happy to close mine in favour of yours.

I did refactor the "get landing page template" logic as well https://github.com/torchbox/wagtail/pull/1464/files#diff-3c070a92794af5b954c6cb38f2dbe724R138 so might be nice to cherry-pick that across.

wagtail/wagtailforms/models.py
```diff
@@ -132,15 +132,37 @@ class AbstractForm(Page):
...
+            "The automatically generated landing_page_template attribute is deprecated - "
+            "use AbstractForm.get_template(request, is_landing_page=True), or "
+            "set an explicit landing_page_template attribute instead.",
+            RemovedInWagtail13Warning)
+        return template_wo_ext + '_landing.html'
 
     class Meta:
         abstract = True
 
+    def get_template(self, request, is_landing_page=False, *args, **kwargs):
```
kaedroho: I think it might be a bit cleaner to split this into two functions, ``get_template`` and ``get_landing_page_template``. As it looks like that'll cleanup some of the internal logic a bit.

timheap:
I split the `get_template` function in two for AbstractForms

gasman:
Items deprecated in 1.2 will be removed in 1.4, not 1.3 - looks like we haven't yet bumped up the version numbers in https://github.com/torchbox/wagtail/blob/master/wagtail/utils/deprecation.py (or removed the things marked as RemovedInWagtail12Warning...)

timheap:
I've made PR #1797 that removes all features scheduled for removal in Wagtail 1.2, and rebased this PR on top of that one.

wagtail/wagtailcore/models.py
```diff
@@ -567,11 +562,48 @@ def get_context(self, request, *args, **kwargs):
...
+
+        * If `request.is_ajax()`, and the Page has an `ajax_template`
+          attribute, that template will be used.
+        * Otherwise, if the Page has a `template` attribute, that will be used.
+        * Otherwise, a template name of `"app/model_name.html"` will be
+          generated.
+        """
+        # RemovedInWagtail14Warning: `is_auto` and `is_ajax_auto`.
+        # `template` and `ajax_template` attributes will just be set to None
+        # in Wagtail 1.3, and the `is_auto` and `is_ajax_auto` variables
```
kaedroho: in Wagtail 1.__4__,

kaedroho:
looks good to me :+1:

wagtail/wagtailcore/models.py
timheap: Oops, missed that one when I was making the RemovedInWagtail14Warning. Fixed now

gasman:
If I'm not mistaken, this breaks the current behaviour of the `template`/`ajax_template` attributes being non-inheritable - if EventPage sets an explicit `template`, and SpecialEventPage subclasses it and doesn't specify `template`, then SpecialEventPage should get an auto-generated template name as normal, not inherit EventPage's template.

Having said that, I'm not 100% sold on that being the 'correct' behaviour - but it seems to me that if we're willing to drop that, we can achieve this same thing by simply making Page.template / Page.ajax_template into (non-deprecated) properties - no need to rewrite `get_template` or deprecate anything. To customise the classname-to-filename logic (which is the motivation for this PR, if I'm not totally getting the wrong end of the stick!) you'd override `template` with a different property; to set a specific template for a particular model (and its subclasses), you'd override it with a constant.

(Incidentally, I should mention that a large part of the motivation for the metaclass hackery was to aggressively eliminate any per-request processing that could be done on class initialisation instead - in this case, converting the class name to a template name. However, I'll happily admit that that was probably premature optimisation - the work involved in that step is negligible compared to performing multiple database lookups and rendering an actual template...)

timheap:
I generated the tempate name in the `get_template()` method to mirror how the Django class based views work when generating template names. First they check to see if one is manually set, and if not, generate one in the `get_template()` method, in a similar manner to the code in this PR. Given that Wagtail is moving towards using more CBV, I thought emulating them here would be a nice idea.

If emulating CBVs is not useful, then making the name generated in the `template` attribute would be cleaner, certainly while the old automatically assigned `template` attributes are being deprecated out.

timheap:
The `template` attribute not being inherited is something I had not considered. In my opinion, breaking something as fundamental as class inheritance in this manner is a surprising and bad idea.

timheap:
The other reason for moving all logic to the `get_template()` method was to stop people accessing the `template` attribute directly. If a Page has custom logic in `get_template()`, then this should not be bypassed by using the `template` attribute. Having both a customised `get_template()` method *and* an automatically generated `template` attribute that is now incorrect is confusing.

gasman:
The way I see it, `template` and `get_template()` have two distinct and complementary purposes - `template` defines the template that "belongs" to this page type at the class level (or possibly at the page instance level, if we make it into a property), and `get_template()` is a mechanism to override that on a per-request basis in special cases (e.g. AJAX requests).

The rationale for `template`'s magic non-inheriting behaviour is that it's following a Rails-style "convention over configuration" approach. Every page model has a template associated with it, and the overwhelmingly common convention is to name that template after the model. So if you follow that convention, you're rewarded by not having to actually type out the `template` attribute. The implementation was done on the basis of this 'ideal' behaviour (and since it doesn't match the natural behaviour of inheritance, it needed some metaclass hacking to make it work).

Will bump this to 1.4 as I guess we're not going to get this debate settled quickly :-)

timheap:
This PR does make a few changes which can be debated separately. Most are not controversial, and I can split these out to a new PR so they can be merged independently of the controversial change. The changes as I see them are:

1. Change `wagtailforms` to use `get_template()` instead of accessing `template` directly;
2. Stop generating template names in the Page metaclass, which is hard to override and customise, and violates class inheritance;
3. Forcibly encourage the use of the official `get_template()` API, instead of using `template` directly, by removing the automatically generated attribute completely; and
4. Move template name generation to `get_template()`

Change 1. is not controversial in my opinion, and I can pull that out in to a new PR. The other three are interrelated, but can be considered as separate points for discussion.

For change 2., the two arguments for this I have seen are: saving per-request work of generating the template name, and making `template` a special attribute that does not participate in class inheritance. As you later identified, the work of building up a small string is insignificant compared to database lookups and template rendering, so performance is not a great argument. From what I understand, the second argument applies in the following use case:

```python
class ContentPage(Page):
    body = RichTextField()
    template = 'content/custom_template_name.html'

class CarouselContentPage(ContentPage):
    slides = StreamField(CarouselStreamBlock())
```

With the current metaclass magic, `CarouselContentPage.template` will be set to `content/carousel_content_page.html`, which is your desired behaviour. While I can see how this is useful, I still think that it is not worth the price of breaking class inheritance in a surprising manner - certainly as developers currently have to resort to counter-metaclass-magic to clobber the attribute right back again if they want to override the attribute for all page types - partly because of change 3., where `get_template()` is not always used.

Change 3. forces developers to use the API that Wagtail wants them to use, instead of skipping around it and using the `template` attribute directly, as happens in `wagtailforms`. By removing the default `template` attribute completely (after being deprecated) in the default case, developers must go through the intended API by calling `get_template()`. This will also help with potential, obscure bugs where one developer has customised `get_template()`, and another developer shortcuts and uses `template` directly, which now does not reflect the intended template set by the first developer.

Change 4. is the logical conclusion of the above changes. As automatic template name generation is still a good idea, the generation needs to go somewhere. `get_template()` is where this code lives now. It can easily be overridden for all classes by changing `get_template()` on a base class; all other code must call this single point of truth to get the template name; and a (potentially incorrect) `template` attribute will not be set on Page subclasses that have overridden `get_template()`. Developers will still be able to easily change the template for a single Page type by setting `template`, as before. It has the added benefit of matching how Django works with class based views, where the `get_template_names()` function first checks the class for a `template_name` attribute, falling back to generating a name if `template_name is None`.

For an example, we used Jade templates with Django and Wagtail for a while. If I had used wagtailforms in one of these projects, it would have accessed the `template` attribute directly, which was set to the automatically generated, but now incorrect `"content/content_page.html"`, instead of going through the easily overridable `get_template()`. To ensure this didnt happen, I resorted to overriding the metaclass of all my Pages to override the generated `template` attribute, and generating a new `template` attribute like `content/content_page.jade`.

timheap:
Rebased on master. For some reason, the elasticsearch tests are occasionally and randomly failing, causing the tests to appear failed. All of the tests relating to this PR passed.

timheap:
Rebased on master again

timheap:
Rebased on master again

auvipy:
need new taker?

timheap:
I no longer work with Wagtail regularly, so I will not be maintaining this pull request. I still stand by the arguments in favour of this change I made above. If you want to take my code, or the suggested changes, and make it work on a more recent version of Wagtail, please feel free!

lb-:
Discussed this at the core team meeting today and...

As it has been a while and this PR has not been able to be picked up again, we will close this PR.

There may be other ways to achieve this goal for pages in the current code and anyone is free to pick this up again, take the code and rework it to suit the current codebase.

Note that EditHandlers (Panels) have had a major rework and have a bit more coming in the next release so some of the desired functionality there may also solve this for common cases now.

Issue #1727, opened 2015-09-23
Consider using Bourbon for admin front-end


alexgleason:
[Bourbon](http://bourbon.io/) is a SCSS mixin library. Looking through Wagtail's mixins, it seems that what it's accomplishing is very similar to what Bourbon already does, except Bourbon is a specialized project, has more features, and using it would be a formalized way of writing code.

Bourbon also provides [Neat](http://neat.bourbon.io/), an add-on to Bourbon for creating grid layouts. I use these two in conjunction to develop websites and am very happy with the results.

I think we should consider using this for the Wagtail admin since it would create a greater separation of effort while ultimately benefiting both projects.

davecranwell:
Interesting idea, Alex. We'll discuss this internally.

gasman:
Closing as per discussion on #1731

PR #1731, opened 2015-09-24
Refactored SCSS to use Bourbon and Neat

alexgleason:
#1727

This is a proof of concept. I haven't done extensive testing yet, everything seems to work fine. The grid system has been replaced with Neat, and every mixin with a Bourbon alternative has also been replaced. Some of the vendor-prefixed mixins like `box-shadow` and `border-radius` have been removed due to having adequate browser support without prefixes.

The only potential concern I see is the possibility of third-party plugins depending on any of these mixins. That said, the transition is fairly easy since the current implementation is already so similar to Bourbon. I would be willing to go through some repos and help people switch over if they're using the mixins.

Let me know what you all think.


davecranwell:
Hi @alexgleason thanks for this huge PR! 

Please could you explain why you feel Bourbon and Neat are a preferable solution to what we have at present? And what led you to this solution as opposed to any other grid/mixin framework? 

One benefit is the external and comprehensive documentation - which we don't have of our own framework. One downside is that our product is then dependent on third party software which may fall into disrepair or may need to be swapped out for something _else_ in the future.

For example I'm personally more a fan of doing browser prefixing via a node component, rather than as part of sass mixins.

On a housekeeping point: I think we'd need this PR to become more than a proof of concept for us to leave this PR open. So if you're up for it you'd need to make it production ready. We rather avoid having Issues or PRs which are essentially un-closable.

tomdyson:
Thanks, @alexgleason! @JoshBarr we'd be interested in your opinions on Alex's proposal - see also #1727.

davecranwell:
Hi again @alexgleason, to backtrack on the seeming negativity of my previous comment (ahem, sorry) and having discussed this internally, we're up for taking the plunge with Bourbon and Neat.

Reliance on a third party library brings the benefits of their documentation, expertise and maintenance, but to incorporate it in our own repo requires us to somehow maintain an up to date version of that software. That's fine with Python modules, but less straight forward with Ruby Gems. 

We've already gone to considerable lengths to avoid Wagtail _implementers_ requiring Ruby to compile SASS. But even _developers_ don't need it since we're using node-sass.

Would we manually keep a copy of Bourbon in the Wagtail repo, or would we have developers install the latest version via Ruby? The former would keep Ruby out of the mix which to my mind is preferable, but would require manual maintenance.

gasman:
https://github.com/thoughtbot/bourbon#installing-with-npm-and-using-a-node-based-asset-pipeline would seem to indicate that node-sass can be used in place of the Ruby toolchain, which hopefully means we wouldn't need any new build dependencies.

(I'm not sure how I feel about further committing ourselves to the Node toolchain, though. I'd be much happier if we could do the entire process through pip-installable packages like python-libsass...)

davecranwell:
Ah interesting, so we could maintain the bourbon version along with our other sass dependencies.

I thing Node's appeal in our particular situation is that we want FE developers to commit to Wagtail, but those developers don't always know Python, but know JS. We're using node to lint, compile and refresh, and I'm hoping to introduce CommonJS or similar at some point, which requires further compilation tasks. I'd find it less easy to contribute or advance the FE side if I needed more Python skills.

alexgleason:
Hi everyone!

>Would we manually keep a copy of Bourbon in the Wagtail repo, or would we have developers install the latest version via Ruby?

The reason I added it to the repo is because I recall a discussion (maybe it was Tom's DjangoCon talk?) about keeping static files in the repo so it's easier for people to contribute. Upon further consideration that's not really valid here since it's a dependency of the source files. I'm personally all for having it not in the repo, and managing it with Node and Bower. I don't have a problem depending more on Node either, we're already using it to automate the build process and I don't envision it going away any time soon. To get Bower working, we'd install it as a Node dependency then add this to the package.json:

```
  "scripts": {
    "postinstall": "bower install"
  },
```

So that after the user runs `npm install`, it automatically installs the Bower components.

For anyone unfamiliar with [Bower](http://bower.io/), it's a package manager for front-end libraries. It creates a `bower_components` directory which you add to your .gitignore, and a `bower.json` file in the root of your project where your dependencies are listed similarly to the `package.json`. Normally you would run `bower install` from the command line to create the components directory, but we can have it done automatically by adding a postinstall to the scripts above. ([reference](https://docs.npmjs.com/misc/scripts))

We would then add the paths (ex. `bower_components/bourbon/scss/`) for Bourbon and Neat into the Sass path and it would work. Our imports would be relative to those paths.

>I think we'd need this PR to become more than a proof of concept for us to leave this PR open.

To clarify, it actually does work. I just know Wagtail has a lot of moving parts and I want to be sure nothing was missed since this is a large change. I think testing it with the wagtailstyleguide and on an already implemented site (maybe wagtaildemo is enough) should be fine, so I will do that and get back to you on it.

Let me know what you think on using Node and Bower.

davecranwell:
I'm all for Node, but Bower less-so. 

I've never used Bower in production before, but as I understand it all it does it manage third party library dependencies and updates - which is just the same as package.json does. Once you've got bower components installed, I find the need to refer to their location inside a "bower_components" directory rather icky. Bower seems like a tool best suited to bootstrapping a totally new project, not an existing one. Bourbon's installation instructions (which gasman links to above) mention no need for Bower, indeed a package.json semver is all the update management we need surely?

JoshBarr:
@gasman I think with our aspirations for crafting an excellent front-end experience, Wagtail will probably need to embrace the node ecosystem, at least for CMS development. We should aim to be writing ES6 (Babel), unit testing the clientside code extensively (Mocha/Chai), and implementing code splitting/bundling for optimised page loads inside the admin (Webpack). I think most front-end devs would assume a node-based workflow for their contributions, regardless of the backend technology. 

Let's chat about bourbon for a sec. If we assume a node-based workflow, and use a streaming build system such as Gulp, it's efficient to pipe `src` => `sass` => `postCSS` => `dest`, and use postcss filters to do things like `rem` fallbacks, linear gradients and vendor prefixes. In this way, front-end devs just write nice, standards-compliant CSS, and let the tooling handle the backwards-compatibility. This means the stylesheets don't depend on additional layers of sass abstraction (beyond Sass itself), which I feel can be a bit of a rabbit hole.

At Springload, we settled on running [Pleeease](http://pleeease.io/) after the Sass compilation step to add vendor prefixes and generally improve browser support. 

Conceptually, this means the JS (compiled through Babel) and CSS are authored in a similar way: write code that resembles the bleeding edge, and use the tooling to polyfill for legacy environments. It's then possible to turn off the polyfills as browser support catches up, without having to refactor code.

davecranwell:
@JoshBarr On the issue of vendor prefixing I agree: I'd prefer that to be done with tooling (gulp-autoprefixer, pleeeease et al), than for us to have to add functions to our code from a library (Bourbon or otherwise) which gradually fall into disuse. However from what I can see Bourbon provides quite a bit more than prefixing. Do you think Bourbon's other features are of value? What about Neat?

gasman:
> @gasman I think with our aspirations for crafting an excellent front-end experience, Wagtail will probably need to embrace the node ecosystem, at least for CMS development.

@JoshBarr OK, fair point! (While there's nothing we're doing *now* that couldn't be done equally well on the Python side, I guess it's only a matter of time until we want to adopt a tool that has no Python-world equivalent, and there's no point holding back the tide until that moment arrives...)

JoshBarr:
I like Bourbon's inclusion of the CSS timing function equations and some of the smaller helper methods like pxrem and pxem, but I'd prefer to cherry-pick the individual functions as needed rather than bringing in a whole library.  

I think Bourbon (and Compass, etc) accidentally create lots of CSS architecture anti-patterns by making it easy to produce unnecessary code (the buttons mixins for instance) and encouraging extra abstractions in the Sass domain rather than focusing on writing fewer rules and creating re-usable patterns in CSS. 

Developer experience is also worth considering. Bourbon introduces things like this:
```
.element {
  @include position(relative, 0 null null 10em);
}
```
My editor won't natively autocomplete includes, meaning it's probably slower to author in bourbon than the equivalent CSS.

CSS is genuinely one of the trickiest things to get your head around on a sufficiently complex app, and I think the more the stylesheets resemble vanilla CSS (that can be parsed/linted/syntax highlighted/auto-completed etc), the safer we are. For solving the problems of more standardised CSS code and easy developer spin-up, there's probably some interventions (richer comments with examples, an auto-generated pattern library, linting, reducing nesting, cleanup of dead code, etc) that would have more benefits to the CSS than adding extra dependencies to the mix. 

Regarding grids, Chris Coyier's excellent [article](https://css-tricks.com/dont-overthink-it-grids/) and  CSSWizardry's [super-simple grid](http://csswizardry.com/csswizardry-grids/) are very nice. I use this [basic](https://github.com/JoshBarr/yasha/blob/master/core/static_src/sass/components/_widths.scss) variation of it for most projects.

davecranwell:
Hmm. Wagtail's mixins/grid systems are already pretty much stripped back to the bare essentials. I don't think we have any mixins that aren't used, nor do we tend to write non-vanilla CSS. The grid is incredibly simple but uses a slightly non-standard technique (padding instead of margin). It could probably stand a little improvement. Auto-prefixing is something I'd definitely like to do though, which removes the need for most of our mixins.

alexgleason:
Thanks for the excellent insight @JoshBarr! I wasn't aware of Pleeease, that looks awesome.

I'd like to share what I like about Neat. I used Foundation for a while where you create the grid with your markup. You may create something like this.

```html
<section class="row">
	<div class="team-members large-10 medium-12 small-centered columns">
		<article class="team-member large-4 medium-3 columns">
			<img class="photo" src="#">
			<div class="biography">...</div>
		</article>
	</div>
</section>
```

When you take everything into account, grid systems actually are kind of complicated. You need to account for every column at every supported screen size. Also, when all your templates are filled with code like this, it gets really hard to read and maintain. Using Bourbon, you could do something like this:

```html
<section>
	<div class="team-members">
		<article class="team-member">
			<img class="photo" src="#">
			<div class="biography">...</div>
		</article>
	</div>
</section>
```

```scss
main > section {
  @include row();
}
.team-members {
  @include media($medium) { @include span-columns(12); }
  @include media($large) {
    @include span-columns(10);
    @include shift(1);
  }
  .team-member {
    @include media($medium) { @include span-columns(3); }
    @include media($large) { @include span-columns(4); }
  }
}
```

This makes sense for a few reasons.

1. Clean and semantic markup, better readability and maintainability. All those `large-4`s add up to create intense mental load when editing templates.
2. How wide something appears is a quality of its _style_. Logically, it should exist as part of the stylesheet.
3. Exposing the grid to SCSS allows you to write more clever layout code, if that's what's best for what you're trying to accomplish. There's a lot more power.

Wagtail is currently implemented using a SCSS grid, so you may agree with me on the benefits already. I just wanted to bring attention to this, because the CSS Wizardry grid uses markup to create the grid. I have not seen any other grid system do things the way Neat does. Bourbon/Neat is a pure mixin library that produce absolutely no CSS on its own.

Ultimately I want Wagtail to benefit from a separation of effort: why maintain something that someone else already specializes in? It doesn't necessarily have to be Bourbon/Neat, but I'm not aware of any other grid system that does exactly what I want.

>Developer experience is also worth considering. Bourbon introduces things like this

I would personally avoid using the `position` mixin provided by Bourbon, but I suppose there's nothing stopping someone from sending a pull request with it so I see your point. But I still think the benefits of using Bourbon/Neat would outweigh the negatives, and I think it's an improvement over the current implementation.

In my view, using them opens up opportunities to write cleaner, better code at the discretion of the programmer. Its features should be used carefully, but the same can be said about CSS itself, which is not perfect.

> I don't think we have any mixins that aren't used

@davecranwell Here's a list of unused mixins in Wagtail's SCSS:

* @mixin filter($filter...)
* @mixin transition-delay($delay...)
* @mixin notransition()
* @mixin content-box()
* @mixin table-column($x, $padding:$padding, $grid-columns:$grid-columns)
* @mixin push($offset:1, $grid-columns:$grid-columns)
* @mixin push-padding($offset:1, $grid-columns:$grid-columns)
* @mixin pull($offset:1, $grid-columns:$grid-columns)
* @mixin pull-padding($offset:1, $grid-columns:$grid-columns)

JoshBarr:
"Clean and semantic markup, better readability and maintainability" isn't necessarily true, since by coupling the layout abstractions to particular source ordering in the markup, the HTML becomes very brittle, and it's difficult to alter the source order without opening yourself up for a world of pain figuring out where all the layout magic is coming from. To me, this discussion isn't really about Bourbon vs not Bourbon, but about where we want complexity to live in the frontend. 

I don't feel the Wagtail UI does anything complex enough with grids for us to use a super abstract grid magic system. Most of the layout relies on a single column interface with a sidebar. As you rightly noted, the push and pull mixins aren't used in the current structure. 

I agree that having half a dozen layout classes on every node in the UI doesn't make much sense, but Neat isn't the only way of moving those layout abstractions into the stylesheet. You can achieve a similar thing (and possibly more efficient, via extends) with a very simple implementation. Consider:

```scss
// Grid definitions...
%medium-one-half {
    @include medium {
         width: 50%;
    }
}

%large-one-third {
    @include large {
         width: 33.333333%;
    }
}

// Implementation...
.some-thing {
   @extend %medium-one-half;
   @extend %large-one-third;
}
```

alexgleason:
But the code you provided is almost exactly what Neat already does. And it's almost exactly what Wagtail already does. But Neat is being maintained separately by a group of people so you don't have to. I don't understand why you'd go through the trouble of creating your own grid when one already exists. You don't have to use the other parts of Bourbon and Neat if you don't want to, they don't produce any output.

We're hooking into a 100 port switch. We many only use 5 of those ports, but it's okay. There are plenty of switches to go around. People say the same thing about Django, "you don't need all that stuff." But you can choose to not use it, so it's okay with me.

And I definitely don't want to introduce any magic. You can be clever without using magic. For instance, object-oriented programming is clever, but straightforward. I have occasionally used the same principles in my Sass, and documented them.

davecranwell:
The main thing I'm hearing here is that while there are no doubt benefits to Bourbon and Neat in a general sense, we're slightly reaching for reasons to implement it _here_.

> But the code you provided is almost exactly what Neat already does. And it's almost exactly what Wagtail already does....I don't understand why you'd go through the trouble of creating your own grid when one already exists.

Why go to the trouble of adopting a new grid when one already exists, particularly if it's almost exactly what Neat already does? 

One of our unofficially guiding principals thus far as been https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it and this feels a little like we're talking of adding something prematurely.

alexgleason:
That philosophy says not to add your own functionality until necessary, but Bourbon is a library. We do need a grid, and Bourbon provides one that is better than the current one. It also contributes no bloat whatsoever due to being a 100% mixin library. If there were a library in C that needed to be vendored and compiled into the distribution binary I could see your point, but Bourbon just sits there and only copies code into the output at the points which you explicitly call it.

Django has contrib modules we don't use most of the time, and you won't need Django's MySQL support if your project requires PostgreSQL, but you don't write a MVP framework in Python from scratch every time you start a project just because you won't use those features.

The reason I think Bourbon is better is because it's clear and well documented. It's been thoroughly thought through, used and developed for years. You would probably be better off creating a fork of it to remove what you don't like than to rethink it through and create a new one from scratch. Not that Torchbox hasn't recreated things from scratch before... but I don't feel it's necessary in this case. Just my 2 pence.

davecranwell:
Yeah it's nice to have a healthy debate about this stuff :)

I think the first step we should take is to adopt a post processor for vendor prefixes. That seems like a no-brainer as it standardises coding without requiring anyone to learn anything or really change their behaviour.

alexgleason:
Yep, makes sense. I'll go ahead and close this for now. Thanks!

Issue #1083, opened 2015-03-22
Deployment docs could be improved


kaedroho:
http://docs.wagtail.io/en/stable/advanced_topics/deploying.html#deploying-wagtail

Currently points to an outdated, hard to read bash script and Django's deployment docs.

SalahAdDin:
Can i put a small tutorial for deployment in wagtail?

kaedroho:
Yes please, that would be really helpful. Please feel free to ask any questions here

SalahAdDin:
What have i do for do the tutorial?

kaedroho:
The idea is to make a Wagtail equivalent of this: https://docs.djangoproject.com/en/1.7/howto/deployment/

The audience will be sysadmins deploying a Wagtail site to a server. Keep in mind that they may not be familar with Wagtails internals.

Django have already done most of the work for us so we only need to document certian differences (and link to the Django docs where possible).

Off the top of my head, here's a few differences:
 - These image libraries must be installed: ``libjpeg-dev libtiff-dev zlib1g-dev libfreetype6-dev liblcms2-dev``
 - We need to mention that Wagtail works best with Elasticsearch and provide links to installation docs for that (and also point out the ``update_index`` command)
 - Wagtail uses ``django-compressor`` to compress the static assets of the Wagtail admin. So the docs must mention that the "django-admin compress" command should be run on every update (as well as first deployment)

Some deployment-related things are actually documented in: http://docs.wagtail.io/en/stable/howto/performance.html (so maybe we could integrate that into this)

ftarnell:
Also mention the `publish_scheduled_pages` cron task when necessary.

SalahAdDin:
I can't put the step needed for deploy in web faction, are a few, no much, install very well libsass, install wagtail, and then use the command for compressor, no more, is easy :D

zerolab:
From documentation sprint:

Deploying a Wagtail project is no different to deploying any Django project, thus first point to the Django deployment page.

There are several prerequisites: 

* Debian:
  `zlib1g-dev libjpeg-dev libpq-dev build-essential`
* RedHat:
  * `zlib-devel libjpeg-devel gcc gcc-c++ postgresql-devel`
* Postgres > 9
* Elasticsearch > 1.0 if using the ES backend

The other parts are non-essential and can be inferred from the different needs (e.g. Redis/Memcache for caching)

davecranwell:
[...]

PR #1739, opened 2015-09-26
Added section on hostnames to Sitemaps docs

JoshBarr:
This is a quick clarification: at first read of the Sitemap docs, it's not immediately obvious that the hostname in the sitemap is derived from the Sites configuration.


kaedroho:
Thanks!

We should mention setting the sites record in the deployment docs I think (#1083), and link to it from the sitemap docs.

PR #1740, opened 2015-09-27
Correct German translation of 'slug'.

jnns:
At the moment, 'slug' is translated with the German word for 'sphere' which even confused me.


Vorlif:
Hi jnns, 

this is really a confused translation but the translation  is centrally managed by transifex. It's probably better if you change it there.

See here: https://www.transifex.com/projects/p/wagtail/

gasman:
Thanks for the report @jnns! I've updated this in Transifex now, so it will be fixed in the codebase when we next pull translations from there.

PR #1741, opened 2015-09-27
Fixes #1329. Set enctype as multipart if form.is_multipart returns true.

### LINKED ISSUE #1329: File upload: Filefield in wagtail admin doesn't work with FieldPanel
In many cases, file upload can be managed with wagtaildocs app, but there are use cases where you just want to upload files in FileField. This doesn't work with wagtail admin cause we don't specify `enctype=multipart/form-data` during form submission [see screenshot]. 

Django seems to handle this with `has_file_field` template var https://github.com/django/django/blob/master/django/contrib/admin/templates/admin/change_form.html#L40, but this is bit lame as that boolean is always true https://github.com/django/django/blob/master/django/contrib/admin/templates/admin/change_form.html#L40

If this solution is fine, let me know. Happy to send PR :)

screen shot 2015-05-21 at 15 47 56

mayapurmedia:
There is [is_multipart()](https://github.com/django/django/blob/593c9eb6608c660f9e2c1ade227670539ececd8e/docs/ref/forms/api.txt#L1085) method in Django to check whether the form requires multipart encoding. 

When form.is_multipart() returns True it adds `enctype="multipart/form-data"` into form in create.html and edit.html templates.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -29,6 +29,7 @@ Changelog
  * Fix: Project template no longer raises a deprecation warning (Maximilian Stauss)
  * Fix: `PageManager.sibling_of(page)` and `PageManager.not_sibling_of(page)` now default to inclusive (i.e. `page` is considered a sibling of itself), for consistency with other sibling methods
  * Fix: The "view live" button displayed after publishing a page now correctly reflects any changes made to the page slug (Ryan Pineo)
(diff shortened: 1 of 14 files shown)


mayapurmedia:
Fixes #1329

gasman:
@mayapurmedia Thanks! Could you add a test for this, please?

mayapurmedia:
@gasman Added tests.

Issue #1488, opened 2015-07-07
Improve page-level group permission UI/explanation


davecranwell:
At present the page permissions assignable to a user group are added via a series of dropdowns and act _cumulatively_, rather than in a trumping fashion E.g "Publish any page" does not trump "Add/edit pages you own", but rather both these permissions need adding individually for each section in question, for all of add, edit and publish permissions to work.

This is particularly confusing because the "object permissions" above use checkboxes to imply the cumulative approach, so the existence of dropdowns for pages suggest a different mechanism.

Suggest we change the dropdowns to also use checkboxes. 

An interim solution might just be to add help text explaining the process.

gasman:
Changing the interface to use checkboxes turns out to be deceptively complicated - I've tried to do this for the new collections permission interface, and have had to admit defeat...

The natural way to try and implement it is a formset, with one form per row of the table. This formset cannot be a straightforward InlineFormSet, because each form is managing multiple database objects (namely, one GroupPagePermission object for each checkbox) - it's an entirely new kind of formset.

So, this new formset type needs a way to keep track of which forms are which, as they get added and deleted. ModelFormSet and InlineFormSet do this by outputting a hidden field containing the ID of the model instance that backs the form. But in our case, there is no single database object that maps to the form - rather, the form is uniquely identified by a tuple of (group_id, page_id). Except we can't guarantee that the page_id will remain static for a given form - the user might change it, using the page chooser widget. That probably means we have to have a hidden field to store the original page ID, and different code paths in the form's save() method depending on whether the user has changed it or not. So what this amounts to is a new formset type, as complex as InlineFormSet if not more, which cannot be implemented as a logical extension of InlineFormSet. Fun times :-/

PR #1751, opened 2015-09-29
Collections for image/document organisation

gasman:
Spec: https://github.com/torchbox/wagtail/wiki/Collections-RFC, https://groups.google.com/d/msg/wagtail-developers/kE-aDrOIzRI/Cc2CP3KbAwAJ

This has been developed as a change to core Wagtail rather than a pluggable module, since it interacts with many existing areas of Wagtail, such as the permissions model, document/image edit forms, and the groups admin interface - and implementing a general-purpose extension framework in all of those areas would be a far bigger job. However, I've tried to keep the code compartmentalised enough so that it can be enabled/disabled depending on whether or not Image and Document inherit from the `CollectionMember` class. By default, custom image models do not inherit from this, so collections will be disabled for those.

As it stands this is a 'minimum viable product' and has a few rough edges:

* There is no restriction on what documents/images the user can use (i.e. see within the document/image chooser) - we agreed that this was out of scope for the initial feature.
* There is no interface for browsing documents / images by collection - the listing pages simply show all items that the user has edit permission for (meaning that currently, collections are *only* useful for permission restriction - not for general asset organisation).
* Deleting a collection will delete all documents / images within it; the 'manage collections' interface does not warn about this (since the collection management area doesn't currently know or care which models are making use of collections).
* Our aim was for the collection feature to be effectively invisible (apart from the addition of a Collections item in the settings menu) until a site is actively making use of it - i.e. a user creating a new Wagtail site would not have to understand the concept of collections up-front. We haven't entirely achieved this, because the document/image permissions within the 'edit group' UI now appear under 'collection permissions'. In order to configure these permissions, a user will have to understand what it means to be specifying those permissions on the 'root' collection.
* It hasn't had any design review yet. The UI additions here are mostly uncontroversial, as they're mainly just adding extra fields to an existing form - but the collection chooser dropdown on the 'add multiple images' view probably needs some attention.
* It hasn't been tested with custom image models yet.

This PR incorporates #1668.

### DIFF
--- a/wagtail/wagtailimages/tests/test_image_operations.py
+++ b/wagtail/wagtailimages/tests/test_image_operations.py
@@ -1,4 +1,4 @@
-import unittest
+from django.test import TestCase
 
 from wagtail.wagtailimages import image_operations
 from wagtail.wagtailimages.exceptions import InvalidFilterSpecError
@@ -33,7 +33,7 @@ class WillowOperationRecorder(object):
         return size
 
 
-class ImageOperationTestCase(unittest.TestCase):
+class ImageOperationTestCase(TestCase):
     operation_class = None
     filter_spec_tests = []
     filter_spec_error_tests = []
@@ -62,8 +62,10 @@ class ImageOperationTestCase(unittest.TestCase):
         return test_filter_spec_error
 
     @classmethod
-    def make_run_test(cls, filter_spec, image, expected_output):
+    def make_run_test(cls, filter_spec, image_kwargs, expected_output):
         def test_run(self):
+            image = Image(**image_kwargs)
+
             # Make operation
             operation = self.operation_class(*filter_spec.split('-'))
 
@@ -115,7 +117,7 @@ class TestDoNothingOperation(ImageOperationTestCase):
     ]
 
     run_tests = [
-        ('original', Image(width=1000, height=1000), []),
+        ('original', dict(width=1000, height=1000), []),
     ]
 
 TestDoNothingOperation.setup_test_methods()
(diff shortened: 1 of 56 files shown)


wagtail/wagtailimages/tests/test_image_operations.py
```diff
@@ -1,4 +1,4 @@
-import unittest
+from django.test import TestCase
```
kaedroho: Out of interest, what's the reason for this switch? I chose ``unittest`` as these tests don't need database rollback

gasman: This change may well be unnecessary. I had to update these tests as they were originally creating Image objects in module scope (i.e. before the database was initialised), and that's no longer possible since we have to do a DB lookup to find the default (root) collection. It took a few hours to track this down as the root cause, and changing `unittest.TestCase` to `django.test.TestCase` was just my first pass at trying to fix it.

kaedroho:
The code in ``wagtailimages/permissions.py`` and ``wagtaildocs/permissions.py`` looks to be quite duplicated. Could these permissions checks be refactored into a class?

gasman:
> The code in ``wagtailimages/permissions.py`` and ``wagtaildocs/permissions.py`` looks to be quite duplicated. Could these permissions checks be refactored into a class?

That was indeed the plan (as outlined in https://github.com/torchbox/wagtail/issues/1384#issuecomment-142986860) - but I got part way through it before discovering that wagtailmodeladmin had something very similar https://github.com/ababic/wagtailmodeladmin/blob/master/wagtailmodeladmin/permission_helpers.py. Which meant that I had the choice of refactoring my code into something that was potentially diverging from wagtailmodeladmin's version, or integrating wagtailmodeladmin - and at that point I figured that it was better to save that rabbit-hole for another day :-)

SalahAdDin:
:+1:

gasman:
Rebased on top of master (and #1896, #1897, #1898) in https://github.com/gasman/wagtail/tree/feature/collection-tree-model-2. Will open a new PR to replace this one once the following blockers have been addressed:

* tests
* UI for browsing by collection
* refactoring to make CollectionMember NOT optional on image/document models (the conditionals make the code ugly, and only arose from misplaced sitting-on-the-fence about the value of this feature)
* if possible, implement #1488 so that the permission management UI for page and collection permissions behaves the same as global permissions (so that the overall user experience is not a step backwards from when images/document permissions where managed through those checkboxes - and hopefully remains substantially similar, in accordance with the Complex Features Should Get Out Of Your Way principle).

gasman:
Superseded by #2243

PR #1753, opened 2015-09-29
use modelcluster to prep image id for json serialization

thenewguy:
backport to 1.1.x


gasman:
Hi @thenewguy - there's no need to create separate PRs for backports. It's not yet decided whether we'll create another 1.1.x maintenance release (most likely, we'll only do so if the changes in 1.2 make upgrading difficult) - and if we do, we'll cherry-pick the commits at that point. Cheers!

Issue #1504, opened 2015-07-10
Remove "wagtail" prefix from all modules


kaedroho:
Since we've dropped Django 1.6 support, we can now rename all modules to get rid of their "wagtail" prefix. For example: `wagtail.wagtailcore` => `wagtail.core`

I would like to propose two exceptions to this:
- `wagtail.contrib.wagtailmedusa` => `wagtail.contrib.staticsitegen`
- `wagtail.contrib.wagtaildocs` => `wagtail.documents`

I'd also like to propose we take this opportunity to move some apps to contrib:
- `wagtail.wagtailforms` => `wagtail.contrib.forms`
- `wagtail.wagtailredirects` => `wagtail.contrib.redirects`
- `wagtail.wagtailembeds` => `wagtail.contrib.embeds`

We should start with contrib apps, then move on to core apps.

Progress:
- [ ] `wagtail.contrib.api` #1498 
- [ ] `wagtail.contrib.frontendcache` #1499
- [ ] `wagtail.contrib.staticsitegen` #1500 
- [ ] `wagtail.contrib.routablepage`
- [ ] `wagtail.contrib.sitemaps` #1503
- [ ] `wagtail.contrib.styleguide` #1585
- [ ] `wagtail.contrib.forms`
- [ ] `wagtail.contrib.redirects`
- [ ] `wagtail.contrib.embeds`
- [ ] `wagtail.admin`
- [ ] `wagtail.core`
- [ ] `wagtail.documents`
- [ ] `wagtail.images`
- [ ] `wagtail.search`
- [ ] `wagtail.sites`
- [ ] `wagtail.snippets`
- [ ] `wagtail.users`

kaedroho:
@gasman said in last weeks dev meeting that we should take the following approach:

 - Rename each module removing the "wagtail" prefix
 - Add alias modules with the old names that import from the new locations (so code continues to work)
 - Do not add deprecation code yet, the old locations will still be the primary places to import from (relying on the alias modules to forward the imports through to actual locations)
 - When all modules are renamed, switch to using the new module names in one go and add deprecation warnings to the alias modules
 - After two release cycles, remove the alias modules

SalahAdDin:
:+1:

gasman:
Completed in #3918

Issue #1461, opened 2015-07-02
Redirects don't work with multiple sites


lee-kagiso:
Right now I have two sites configured:

![redirects](https://cloud.githubusercontent.com/assets/11830393/8475179/530810b8-20b6-11e5-8427-97f9e4581f65.png)

I have noticed that redirects don't work properly when you have more than one site, I can't put my finger on the exact issue, other than they don't kick in.

I think it is because I can't specify the site that the rule applies to, and then the moment you have more than one site the redirects module doesn't know if the rule applies to this site or not?

kaedroho:
Thanks @lee-kagiso

Looking at the code behind this (https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailredirects/models.py#L36-L40) you can either have a redirect that applies to all sites or one particular site. When it new redirect is created, it is currently assigned to the site that the editor was using to browse the admin with: https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailredirects/views.py#L112.

I can't remember the reason behind implementing it this way, but I think we should remove that bit of code at least until we have support for selecting the site (#1163), so all new redirects will be global.

A workaround that should get your current redirects working for now (assuming you want all redirects to be global):

```python
>>> from wagtail.wagtailredirects.models import Redirect
>>> Redirect.objects.all().update(site=None)
```

lee-kagiso:
@kaedroho thanks, I never spotted that you have to change the admin url to the site in question.
I guess as a superadmin I thought I could change any sites redirect rules in one spot...

gasman:
Fixed in #1475; making the site field configurable is covered by #1163.

lee-kagiso:
Nice, thanks!

PR #1754, opened 2015-09-30
Move wagtailsettings into contrib module

timheap:
The `wagtailsettings` module is useful enough that it should be included in the Wagtail contrib section, to make it available to all Wagtail developers.

All the code has been given a once-over to make sure it is nice and polished before being copied in. As such, this is not a direct copy of the `wagtailsettings` module. It should be backwards compatible though, excepting the new location.

It has been moved to `wagtail.contrib.settings`, following the naming scheme set out in #1504.

Documentation has been concatenated in to a single page, and added to the contrib reference section.


wagtail/contrib/settings/templatetags/wagtailsettings_tags.py
```diff
@@ -0,0 +1,21 @@
...
+
+from wagtail.wagtailcore.models import Site
+
+from ..context_processors import SettingsProxy
+
+register = Library()
+
+
+@register.simple_tag(takes_context=True)
+def get_settings(context, use_default_site=False):
```
kaedroho: Could this be turned into an assignment tag? allowing the following syntax ``{% get_settings as wagtail_settings %}``

wagtail/tests/testapp/models.py
```diff
@@ -454,3 +457,18 @@ class MTIChildPage(MTIBasePage):
 class AbstractPage(Page):
     class Meta:
         abstract = True
+
+
+@register_setting
+class TestSetting(BaseSetting):
+    title = models.CharField(max_length=100)
+    email = models.EmailField()
```
kaedroho: The ``max_length`` needs to be explicitly set on this field, to fix the Django 1.7 migration issue.

wagtail/contrib/settings/templatetags/wagtailsettings_tags.py
nealtodd: NB: assignment_tag is deprecated in Django 1.9, where simple_tag will gain support for the 'as' syntax.

timheap: This could be implemented for `get_settings`, but I don't see the template tag being used much. It is only as a fallback in case you are in a strange situation. With the pending deprecation that @nealtodd pointed out, supporting assignment tags across multiple versions of Django will become quite annoying. I do not think it is worth the bother.

wagtail/tests/testapp/models.py
timheap: I've fixed this, tests are now passing.

wagtail/contrib/settings/templatetags/wagtailsettings_tags.py
kaedroho: ok, that's fine by me.

Just note that when the "as" syntax arrives, this tag will fill the specified variable with a blank string which may not be what the developer expects.

docs/reference/contrib/settings.rst
```diff
@@ -0,0 +1,128 @@
...
+        'wagtail.contrib.settings.context_processors.settings',
+    ]
+
+Then access the settings through ``{{ settings }}``:
+
+.. code-block:: html+django
+
+    {{ settings.app_label.SocialMediaSettings.instagram }}
+
+If you are not in a ``RequestContext``, then context processors will not have run, and the ``settings`` variable will not be availble. To get the ``settings``, use the provided ``{% get_settings %}`` template tag. If a ``request`` is in the template context, but for some reason it is not a ``RequestContext``, just use ``{{ get_settings }}``:
```
gasman: Presumably this last `{{ get_settings }}` should be `{% get_settings %}`?

gasman:
The menu items registered by https://github.com/torchbox/wagtail/pull/1754/files#diff-45699f1eed7357d14bd23f5b9b73a817R30 need an `is_shown` method, to ensure that they don't get shown to users who don't have permission over that model. (Feel free to extend the `MenuItem` constructor if necessary, so that you don't have to construct subclasses on-the-fly...)

It would be nice (but not a blocker) to register these models to appear within the groups permission interface too, via the `register_permissions` hook.

wagtail/contrib/settings/permissions.py
```diff
@@ -0,0 +1,19 @@
+from django.contrib.contenttypes.models import ContentType
+
+
+def user_can_edit_setting_type(user, model):
+    """ Check if a user has any permission related to a content type """
```
gasman: The practice of saying "allow users to access this area if they have *any* permission over model X" is something that we've tightened up a lot in Wagtail 1.1 - we now tend to list the specific permissions, e.g. https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailredirects/views.py#L17. (Amongst other things, this opens up the possibility of using permissions for 'lesser' tasks that don't require access to the model's admin area - "can download document", "can link to image" and so on.)

In the case of settings, it probably makes sense for `change_[foo]` to be the only permission type in use - so that would be the one that gets checked in the view and `MenuItem.is_shown`, and the one that gets passed to `register_permissions` to appear in the groups admin (if we go ahead with that).

gasman:
One thing that's quite troublesome here is how, when you edit settings, the site is picked up from the hostname you're accessing the admin from (https://github.com/torchbox/wagtail/pull/1754/files#diff-ffdb5fc31a093a7b481589b85744df46R42). To me, that goes against the design of Wagtail admin as a single unified interface for managing all your sites (such as what you see when you go to the top level of the page tree). The redirects module previously did something like this, which was backed out in 1.1 as it was causing confusion (#1461) - and there's also some demand for being able to host wagtailadmin on a completely separate hostname to your frontend sites (#1807, as well as the whole pattern of static site generation a la wagtailmedusa / wagtailbakery).

Not entirely sure what the neatest way of addressing this is. One possibility is for the menu item to take you straight to the edit form if only one site exists, and to an intermediate "select a site" listing otherwise - but that's a bit clunky. Maybe a dropdown in the header bar of the edit form, where you can select the appropriate site?

timheap:
I'm in the middle of exams right now, so I will not be able to get to these changes in the next few days. Hopefully I can find some time on Wednesday after all my exams.

gasman:
@timheap No worries - thanks for your work on this. Good luck with the exams!

timheap:
Alright, a bunch of fixes pushed: docs typo, better permissions intgration, and the ability to switch sites using a form in the header.

I've pushed these as separate commits for ease of reviewing, but they can be squashed by someone (myself or you) before being merged.

timheap:
Also rebased on master. Does Travis run tests on the resulting merge with master? Travis failed with a 'conflicting migrations' error from Django, which could only have happened through a merge/rebase with master.

timheap:
Thanks!

SalahAdDin:
:+1:

PR #1571, opened 2015-08-01
Use alternate context var for Jinja support

timheap:
Jinja does not support using `self` as a template variable, as `self` is reserved for its own internal use. To add support for Jinja templates, an alternative name has been added where 'self' has been used.  Templat authors are free to use which ever template variable works in their template language.

I have not changed all the internal Wagtail templates, as these are not using Jinja. I have changed the wagtailuserbar templatetag to use the new `page` variable so  that it can theoretically be used in Jinja as well, although the template tag would need to be added to Jinja first.

This needs to be documented somehow, although I am unsure how exactly to do this. The docs could be converted completely to use the `page` template variable, as this would work on all templating engines, but this is a big change. Alternately, a note could be added regarding the `self`/`page` duplication and which to use when, but this note may need to be added to multiple places in the docs so people do not miss it. Which approach would you prefer?

PR #1755, opened 2015-09-30
Add rudimentary Jinja2 template tag support

timheap:
The following tags and filters are supported at this point. More may be added as they are required:

* `wagtailcore`:
  * `pageurl()`
  * `slugurl()`
  * `wagtail_version()`
  * `|pageurl`
* `wagtailimages`:
  * `image()`

Django template tags have been translated to Jinja functions, rather than custom tags. Functions are easier to use compared to template tags, and can be composed and combined for greater flexibility.

The template tag libraries have been grouped as Jinja Extensions, which are loadable via the `extensions` option. An example Django Jinja2 configuration is:

```python
TEMPLATES = [
    # ...
    {
        'BACKEND': 'django.template.backends.jinja2.Jinja2',
        'APP_DIRS': True,
        'OPTIONS': {
            'extensions': [
                'wagtail.wagtailcore.templatetags.jinja2.core',
                'wagtail.wagtailimages.templatetags.jinja2.images',
            ],
        },
    },
]
```

Inspired by https://github.com/minervaproject/wagtail-jinja2/

### DIFF
--- a/docs/advanced_topics/jinja2.rst
+++ b/docs/advanced_topics/jinja2.rst
@@ -28,6 +28,10 @@ Django needs to be configured to support Jinja2 templates. As the Wagtail admin 
         }
     ]
 
+Jinja templates must be placed in a ``jinja2/`` directory in your app. The template for an ``EventPage`` model in an ``events`` app should be created at ``events/jinja2/events/event_page.html``.
+
+By default, the Jinja environment does not have any Django functions or filters. The Django documentation has more information on `configuring Jinja for Django <https://docs.djangoproject.com/en/1.8/topics/templates/#django.template.backends.jinja2.Jinja2>`_.
+
 ``self`` in templates
 =====================
 
--- a/docs/advanced_topics/jinja2.rst
+++ b/docs/advanced_topics/jinja2.rst
@@ -1,0 +1,97 @@
+.. _jinja2:
+
+=======================
+Jinja2 template support
+=======================
+
+Wagtail supports Jinja2 templating for all front end features. More information on each of the template tags below can be found in the :ref:`writing_templates` documentation.
+
+Configuring Django
+==================
+
+Django needs to be configured to support Jinja2 templates. As the Wagtail admin is written using regular Django templates, Django has to be configured to use both templating engines. Wagtail supports the Jinja2 backend that ships with Django 1.8 and above. Add the following configuration to the ``TEMPLATES`` setting for your app:
+
+.. code-block:: python
+
+    TEMPLATES = [
+        # ...
+        {
+            'BACKEND': 'django.template.backends.jinja2.Jinja2',
+            'APP_DIRS': True,
+            'OPTIONS': {
+                'extensions': [
+                    'wagtail.wagtailcore.templatetags.jinja2.core',
+                    'wagtail.wagtailadmin.templatetags.jinja2.userbar',
+                    'wagtail.wagtailimages.templatetags.jinja2.images',
+                ],
+            },
+        }
+    ]
+
+``self`` in templates
+=====================
+
+In Django templates, ``self`` is used to refer to the current page, stream block, or field panel. In Jinja, ``self`` is reserved for internal use. When writing Jinja templates, use ``page`` to refer to pages, ``value`` for stream blocks, and ``field_panel`` for field panels.
+
+Template functions & filters
+============================
+
+``pageurl()``
+~~~~~~~~~~~~~
+
+Generate a URL for a Page instance:
+
+.. code-block:: html+jinja
+
+    <a href="{{ pageurl(page.more_information) }}">More information</a>
+
+See :ref:`pageurl_tag` for more information
+
+``slugurl()``
+~~~~~~~~~~~~~
+
+Generate a URL for a Page with a slug:
+
+.. code-block:: html+jinja
+
+    <a href="{{ pageurl("about") }}">About us</a>
+
+See :ref:`slugurl_tag` for more information
+
+``image()``
+~~~~~~~~~~~
+
+Resize an image, and print an ``<img>`` tag:
+
+.. code-block:: html+jinja
+
+    {# Print an image tag #}
+    {{ image(page.header_image, "fill-1024x200", class="header-image") }}
+
+    {# Resize an image #}
+    {% set background=image(page.background_image, "max-1024x1024") %}
+    <div class="wrapper" style="background-image: url({{ background.url }});">
+
+See :ref:`image_tag` for more information
+
+``|richtext``
+~~~~~~~~~~~~~
+
+Transform Wagtail's internal HTML representation, expanding internal references to pages and images.
+
+.. code-block:: html+jinja
+
+    {{ page.body|richtext }}
+
+See :ref:`rich-text-filter` for more information
+
+``wagtailuserbar()``
+~~~~~~~~~~~~~~~~~~~~
+
+Output the Wagtail contextual flyout menu for editing pages from the front end
+
+.. code-block:: html+jinja
+
+    {{ wagtailuserbar() }}
+
+See :ref:`wagtailuserbar_tag` for more information
--- a/docs/advanced_topics/index.rst
+++ b/docs/advanced_topics/index.rst
@@ -12,3 +12,4 @@ Advanced topics
     privacy
     customisation/index
     third_party_tutorials
+    jinja2
--- a/docs/topics/writing_templates.rst
+++ b/docs/topics/writing_templates.rst
@@ -1,3 +1,5 @@
+.. _writing_templates:
+
 =================
 Writing templates
 =================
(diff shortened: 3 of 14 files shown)


arthurio:
@timheap Nice work, the functions are indeed way easier to implement/understand/maintain so I think it's a good strategy. After using wagtail+jinja for a week now, here is the list of things to consider in my opinion:
  - **wagtailuserbar function**: It will probably require to add a settings to override the [default template] (https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailadmin/templatetags/wagtailuserbar.py#L47) and create a new template wagtailadmin/templates/wagtailadmin/userbar/frame_jinja.html to have a jinja version of https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailadmin/templates/wagtailadmin/userbar/frame.html
  - **A way to render the blocks and richtext fields as safe html**: Right now, in order to render blocks with the proper context and the html fields, I had to implement the following tags:

```python
from django_jinja import library
import jinja2

@library.global_function
@jinja2.contextfunction
def render_stream_child(context, stream_child):
    # Use the django_jinja to get the template content based on its name
    template = get_template(stream_child.block.meta.template)
    # Create a new context based on the current one as we can't edit it directly
    new_context = context.get_all()
    # Add the value on the context (value is the keyword chosen by wagtail for the blocks context)
    new_context['value'] = stream_child.value
    # Render the template with the context
    html = template.render(context=new_context)
    # Return the rendered template as safe html
    return jinja2.Markup(html)

@library.global_function
@jinja2.contextfunction
def render_html(context, html):
    return jinja2.Markup(html)
```

```html
<body>
...
{% for stream_child in page.body %} {# body is a StreamField #}
    {{ render_stream_child(stream_child) }}
{% endfor %}

{{ render_html(page.rich_text) }}
...
</body>
```
For the richtextfield, I tried using the safe filter, the same way we were doing it for django but it was raising ascii exceptions. The block rendering implementation is a bit more tricky and it will probably need some refactoring to get the current context and extend it, as oppose to just defining a context with [self and value](https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailcore/blocks/base.py#L211).

timheap:
@arthurio: The `|richtext` filter I've added works fine for me, including on non-ascii content. I am about to test StreamFields, but they should hopefully just work.

I just added the `wagtailuserbar()` function, which completes the set of front end template tags. I have the following template to test things with, all of which works in the expected manner:

```html+jinja
<!DOCTYPE html>
<html lang="en-AU">
<head>
<meta charset="utf-8">

<title>{{ page.title }}</title>

</head>
<body>

{{ wagtailuserbar() }}

<h1>{{ page.title }}</h1>

<h2>Rich text using <code>|richtext</code></h2>
<blockquote>{{ page.body|richtext }}</blockquote>

<h2>Image using <code>image()</code></h2>
<blockquote>{{ image(page.image, "width-200") }}</blockquote>

<h2>Links</h2>
<a href="{{ pageurl(page.link) }}">Link to <code>pageurl(page.link)</code></a>
<a href="{{ slugurl("about") }}">Link to <code>slugurl("about")</code></a>

<h2><code>wagtailuserbar()</code></h2>
<p>Should be up there ↗</p>

</body>
</html>
```

timheap:
Documentation has been added! Feedback is welcome.

Is everyone happy with the organisation of the Jinja extensions in `wagtail.wagtailapp.templatetags.jinja2.name`? The path is quite long, but it does mean that all the template tag code is kept in the same location. You only have to add the extensions once in your settings, so a long import path probably is not a problem.

timheap:
Also: Following the discussion in #1571, there was some support for changing the use of `self` in templates to be `page`/`value`/`field_panel` through out the documentation, in order to present a consistent template experience for both Django and Jinja templates. Is that worth discussing in this PR, or should that discussion wait for another time?

docs/advanced_topics/jinja2.rst
```diff
@@ -0,0 +1,97 @@
...
+=======================
+Jinja2 template support
+=======================
+
+Wagtail supports Jinja2 templating for all front end features. More information on each of the template tags below can be found in the :ref:`writing_templates` documentation.
+
+Configuring Django
+==================
+
+Django needs to be configured to support Jinja2 templates. As the Wagtail admin is written using regular Django templates, Django has to be configured to use both templating engined. Wagtail supports the Jinja2 backend that ships with Django 1.8 and above. Add the following configuration to the ``TEMPLATES`` setting for your app:
```
tomdyson: engined -> engines

docs/advanced_topics/jinja2.rst
```diff
@@ -0,0 +1,97 @@
...
+    {# Resize an image #}
+    {% set background=image(page.background_image, "max-1024x1024") %}
+    <div class="wrapper" style="background-image: url({{ background.url }});">
+
+See :ref:`image_tag` for more information
+
+``|richtext``
+~~~~~~~~~~~~~
+
+Transform Wagtails internal HTML representation, expanding internal references to pages and images.
```
tomdyson: Wagtails -> Wagtail's

timheap:
Fixed the typos that @tomdyson found. Thanks!

tomdyson:
It all looks good to me, and I'm happy with the organisation of the extensions. Great work, @timheap!

I think we should continue the  `self` -> `page`/`value`/`field_panel` discussion here, or it risks getting lost in the merged #1571 PR. Last time @gasman voted for 

> Put a note in a conspicuous place saying "Use self in Django templates, but use page in Jinja"

but now that we *are* offering full support for Jinja2, perhaps we should upgrade this to

> Rewrite all the documentation to use the new `page` variable, so the documentation works for both template languages

or, as a middle way, include a warning in our conspicuous note that the `self` -> `page` change is coming in 1.3.

timheap:
I was looking for some tests to copy for the Jinja template tags, and I couldnt find much! It seems the `|richtext` filter does not actually have a test. The various components of it do, but not the filter being used in an actual template. I'll add some for Jinja at least.

timheap:
Have some tests. This is the last bit of work for this pull request. If the tests all pass, and there is no more feedback, then this PR is complete.

docs/advanced_topics/jinja2.rst
```diff
@@ -0,0 +1,97 @@
...
+            'OPTIONS': {
+                'extensions': [
+                    'wagtail.wagtailcore.templatetags.jinja2.core',
+                    'wagtail.wagtailadmin.templatetags.jinja2.userbar',
+                    'wagtail.wagtailimages.templatetags.jinja2.images',
+                ],
+            },
+        }
+    ]
+
```
kaedroho: I think it needs to be noted somewhere here that templates need to be created in ``<app>/jinja/`` instead of ``<app>/templates/``

docs/advanced_topics/jinja2.rst
```diff
@@ -0,0 +1,97 @@
...
+=======================
+Jinja2 template support
+=======================
+
+Wagtail supports Jinja2 templating for all front end features. More information on each of the template tags below can be found in the :ref:`writing_templates` documentation.
+
+Configuring Django
+==================
+
+Django needs to be configured to support Jinja2 templates. As the Wagtail admin is written using regular Django templates, Django has to be configured to use both templating engines. Wagtail supports the Jinja2 backend that ships with Django 1.8 and above. Add the following configuration to the ``TEMPLATES`` setting for your app:
```
kaedroho: Could we link to this doc? https://docs.djangoproject.com/en/1.8/topics/templates/#django.template.backends.jinja2.Jinja2

It includes some details on how to setup the environment which is required in order to use the ``url`` and ``static`` tags.

arthurio:
@timheap I made the comments about wagtailuserbar because I thought you were doing the full jinja support (admin side as well), but if it's just for the front-end then disregard. Last thing, were you able to test that the context was the same in the page and in the blocks? If not, I can try to reproduce with your PR and write some tests.

timheap:
I'll add some tests for stream fields shortly. Just so there is no confusion about the expected contents of a StreamField template context, @gasman explains why the StreamField template context can not inherit from the context of the page it is being rendered in: https://groups.google.com/d/msg/wagtail/3BkKoKqTqeM/GOrgBfRPBgAJ

timheap:
I added the `__html__` method to StreamValues, so they play nicely with autoescaping in Jinja templates: <http://jinja.pocoo.org/docs/dev/api/#jinja2.Markup>. Now you can just write `{{ page.stream_field }}` in your Jinja template and it will not be escaped.

Using Jinja templates to render a Block with the `template` Meta attribute works without any modifications, and successfully renders blocks using the `value` context variable:

```python
class ImageBlock(blocks.StructBlock):
    image = ImageChooserBlock()
    caption = blocks.TextBlock()

    class Meta:
        template = 'blocks/image.html'
```
```html+jinja
{# app/jinja2/blocks/image.html #}
<figure>
    {{ image(value.image, "width-100") }}
    <figcaption>{{ value.caption }}</figcaption>
</figure>
```

kaedroho:
Tested this both with a new project and started porting an exising one, no issues to report. Great work @timheap!

kave:
omg this is just what i was looking for

PR #1768, opened 2015-10-03
Use model specific "DoesNotExist" exception 

annp89:
Replace base class ```ObjectDoesNotExist``` exception with model specific ```DoesNotExist`` exception


gasman:
This will break when custom image models are in use, because they will use their own subclass of `AbstractRendition` rather than `Rendition`. Reverted in #1773

kaedroho:
Oops, that slipped my mind. Thanks for reverting.

We really aught to have tests for custom image models

PR #1769, opened 2015-10-03
Remove use of built in function 'format' as variable name

annp89:
Also, refactor try-catch block to be more specific as to which lines of code could potentially raise the exception and limit try block to those lines.


kaedroho:
Thanks!

Issue #1743, opened 2015-09-27
pagurl tag doesn't work within StreamField


gmeans:
After adding a PageChooserBlock to a stream field and then trying to use the pageurl template tag, I received the following error:

KeyError at <url>
'request'

I stepped through the code and it looks like the block doesn't get the full context passed to it, so request isn't available.

gasman:
This is a consequence of the way StreamField is designed - at some point, you have to render a value within a `{{ ... }}` tag (either the StreamField value itself, e.g. `{{ self.body }}` or a child of it, e.g. `{% for block in self.body %}{{ block }}{% endfor %}`) - and at that point you're rendering a string representation of the value. That value doesn't know anything about the template context it's being rendered in - indeed, there may not be a template context at all, since you could easily retrieve the same value by digging around on the `./manage.py shell` command line. It just so happens that the string representation of that value is (usually) produced by rendering a template, but that's an entirely separate mechanism to the request/response cycle.

Having said that, you're not the first person to encounter this, and the `{% pageurl %}` tag is a particularly strong example of why you might want to access the outer template context within a StreamField's rendering. A common pattern throughout Wagtail's architecture is "objects that know their own HTML rendering (and usually generate it with a template)" - StreamField values being one of those - so perhaps this calls for a new way of handling those, which *does* preserve the template context. We'd have to change our template-writing conventions a bit, since the `{{ ... }}` tag effectively creates a barrier that the template context can't cross. One idea, then:

* Objects that use templates to render themselves - such as StreamField values - implement an `__str__` method that takes an optional `context` argument. If provided, they render templates using that context; if not, they use a blank context.
* Where we would currently write `{{ self.body }}`, we start writing `{% render self.body %}` instead, to denote that we expect to get an HTML rendering back. `render` here would be a custom tag provided by `wagtailcore_tags`. (I'm not keen on the name 'render' here - it's horribly generic. But it's doing something horribly generic, so maybe that's unavoidable :-) )
* The `render` tag would attempt to call `__str__(context)` on the value, and if that fails, fall back to the standard `__str__()` instead.

One quick and easy fix to the immediate problem would be to make `{% pageurl %}` work correctly in the absence of a `request` variable. We check `request` to determine whether the URL is on a different site to the current page, and produce a local URL if not - but it would be simple enough to define alternative behaviour for when `request` isn't available. However, I wouldn't want that to be the only fix we implement here, as it's sweeping the bigger problem under the carpet...

[...]

PR #1771, opened 2015-10-03
Allow adding custom extra context to Block render

tadeo:
Allow a user to add custom extra context to a custom StructBlock
overriding its render method

### DIFF
--- a/wagtail/wagtailcore/blocks/base.py
+++ b/wagtail/wagtailcore/blocks/base.py
@@ -200,7 +200,7 @@ class Block(six.with_metaclass(BaseBlock, object)):
         """
         return value
 
-    def render(self, value):
+    def render(self, value, extra_context=None):
         """
         Return a text rendering of 'value', suitable for display on templates. By default, this will
         use a template if a 'template' property is specified on the block, and fall back on render_basic
@@ -208,10 +208,12 @@ class Block(six.with_metaclass(BaseBlock, object)):
         """
         template = getattr(self.meta, 'template', None)
         if template:
-            return render_to_string(template, {
+            context = {
                 'self': value,
                 self.TEMPLATE_VAR: value,
-            })
+            }
+            context.update(extra_context or {})
+            return render_to_string(template, context)
         else:
             return self.render_basic(value)
 


gasman:
Hi @tadeo - thanks for the suggestion!

I guess the idea here is that a StructBlock would override `render`, and pass the extra context in when calling `super`:

    class PersonBlock(blocks.StructBlock):
        name = blocks.CharBlock()
        birth_date = blocks.DateBlock()
        
        def render(self, value):
            return super(PersonBlock, self).render(value, {
                'star_sign': get_star_sign(value['birthday'])
            })

Since the extra context data only ever comes from inside the object (it's never passed to `render` from the outside), perhaps it would be clearer to add a `get_context` method, like we do for pages:

    class PersonBlock(blocks.StructBlock):
        name = blocks.CharBlock()
        birth_date = blocks.DateBlock()
        
        def get_context(self, value):
            context = super(PersonBlock, self.get_context(value)
            context['star_sign'] = get_star_sign(value['birthday'])
            return context

`Block.render` would then be updated to call `self.get_context` instead of building the context dict itself.

We would then need tests + documentation to cover this new feature.

Note that there is some discussion elsewhere (#1743) about allowing a block to inherit the context from the outer template. I think it would be easy enough to make both approaches work side by side, though - we'd just create a new dict that merges the parent context data with the new extra context data.

tadeo:
Hi @gasman, thanks for your soon answer.
I'm not sure if your assumption of "extra context data only ever coming from inside the object" is wide enough. In fact the render method seems to be only called from outside. Thinking about the discussion you refer to, probably you will need to pass the page context from outside, so the extra_context parameter could be useful to solve that issue. Also, it seems the "Simple is better than complex" python principle favors an extra parameter over an extra method.
Just my two cents, thanks a lot!

gasman:
> In fact the render method seems to be only called from outside.

Yes, that's true at the moment. But those places don't pass the extra_context value, and it would be hard to update them to do so (because rendering of a block is usually triggered from a `{{ ... }}` tag in a template, which calls `to_s` on the value, and that doesn't accept parameters).

I've now put together an alternative implementation in #1812 - hopefully this makes the code cleaner (because `render` is now doing less work), and it also matches the mechanism we use for pages (http://docs.wagtail.io/en/latest/topics/pages.html#customising-template-context).

Anyhow, thanks for putting this together - I think it's going to be a really useful new feature!

PR #1772, opened 2015-10-03
Cleanup in WagtailAPI module

### LINKED ISSUE #1671: Cachebusting query parameter (e.g. _=1441835249458) not ignored by api
From the [documentation for jQuery.ajax, under "cache"](http://api.jquery.com/jquery.ajax/):
> Setting cache to false will only work correctly with HEAD and GET requests. It works by appending "_={timestamp}" to the GET parameters. The parameter is not needed for other types of requests, except in IE8 when a POST is made to a URL that has already been requested by a GET.

It seems like it's standard practice to ignore the underscore keyword. Unless I'm mistaken this is an oversight and not a disagreement on the principle of the thing.

Reproduce: make an Ajax call to any wagtail API endpoint with the cache flag set to false. Or just navigate to something like `http://localhost:8000/api/v1/pages/?type=home.HomePage&_=1441835249458`

You'll get this message: 
```
{
    "message": "query parameter is not an operation or a recognised field: _"
}
```

kaedroho:
See commit log

### DIFF
--- a/wagtail/contrib/wagtailapi/endpoints.py
+++ b/wagtail/contrib/wagtailapi/endpoints.py
@@ -30,7 +30,6 @@ class BaseAPIEndpoint(GenericViewSet):
     base_serializer_class = BaseSerializer
     filter_classes = []
     model = None # Set on subclass
-    queryset = None  # Set on subclasses or implement `get_queryset()`.
 
     known_query_parameters = frozenset([
         'limit',
@@ -41,6 +40,9 @@ class BaseAPIEndpoint(GenericViewSet):
     ])
     extra_api_fields = []
     name = None  # Set on subclass.
+
+    def get_queryset(self):
+        return self.model.objects.all().order_by('id')
 
     def listing_view(self, request):
         queryset = self.get_queryset()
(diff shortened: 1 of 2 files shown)


kaedroho:
I'll split this up into 4 PRs
 - Cleanup
 - Replacing custom renderer with ``WagtailAPIRouter``
 - Allow "_" GET parameter
 - DRF browsable API

kaedroho:
Split up. Couldn't deserialise the changes much though

wagtail/contrib/wagtailapi/endpoints.py
```diff
@@ -28,8 +28,8 @@ class BaseAPIEndpoint(GenericViewSet):
     renderer_classes = [WagtailJSONRenderer]
     pagination_class = WagtailPagination
     base_serializer_class = BaseSerializer
-    filter_classes = []
-    queryset = None  # Set on subclasses or implement `get_queryset()`.
+    filter_backends = []
+    model = None # Set on subclass
```
kaedroho: Need two spaces here

SalahAdDin:
What's about django-rest-framework?

PR #1776, opened 2015-10-04
Remove all ObjectDoesNotExist uses

timheap:
`ObjectDoesNotExist` is usually not the correct error to catch. If you know the model class that would raise the error, you should catch its `Foo.DoesNotExist` error instead.

#1768 attempted this, but did not take in to account swappable Rendition models.


kaedroho:
Merged in d31e4cd032d38f6c816cf3c56115297c3c66d234 Thanks!

PR #1777, opened 2015-10-05
Use isinstance check to detect strings

timheap:
Using `isinstance(filter, string_types)` will detect strings much more reliably than `not hasattr(filter, 'run')`.

### DIFF
--- a/wagtail/wagtailimages/models.py
+++ b/wagtail/wagtailimages/models.py
@@ -21,6 +21,7 @@ from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
 from django.utils.encoding import python_2_unicode_compatible
 from django.utils.functional import cached_property
+from django.utils.six import string_types
 from django.core.urlresolvers import reverse
 
 from unidecode import unidecode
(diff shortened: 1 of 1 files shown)


wagtail/wagtailimages/models.py
```diff
@@ -211,7 +212,7 @@ def get_suggested_focal_point(self):
         return Rect.from_point(x, y, width, height)
 
     def get_rendition(self, filter):
-        if not hasattr(filter, 'run'):
+        if isinstance(filter, string_types):
             # assume we've been passed a filter spec string, rather than a Filter object
```
kaedroho: We no longer need to "assume" it's a spec string

wagtail/wagtailimages/models.py
```diff
@@ -211,7 +212,7 @@ def get_suggested_focal_point(self):
...
             # TODO: keep an in-memory cache of filters, to avoid a db lookup
```
kaedroho: Might be worth killing this comment while you're at it. We plan on making ``Filter`` into a standard python class instead

timheap:
Removed the comments as suggested

kaedroho:
Merged in d97d1f4b8e22dd46cf7262f20a798979fc4dc719 thanks!

PR #1778, opened 2015-10-05
Fix expectedFailures in many tests

timheap:
This PR initially just fixed `expectedFailures` in StreamField tests, but then I went a fixed more. Perhaps this should be made into multiple PRs, but I got carried away!

# Fix expectedFailures in StreamField block tests:

StructBlocks now only print a `<dt>`/`<dd>` pair for fields defined on the StructBlock, instead of every key passed in.

The tests asserting StructBlock mixins have all their child blocks now work, using `set()` for order-independent checking, instead of `list()`s. The order of the fields does not seem to be the important factor being checked in these tests.

FieldBlocks that wrap a ChoiceField return the display value for the chosen value from `get_searchable_content()`

FieldBlocks still return `[]` for `get_searchable_content()` when they are not a ChoiceField. This appears to be the expected behaviour, so I did not modify it.

# Fix expectedFailures in `resolve_model_string` tests

The `expectedFailure` said that `resolve_model_string` should raise a `ValueError` for unhandled types, so now it does. I added some more tests around `resolve_model_string`, to test a few more bad argument types.

# Expect the correct file name in the Content-Disposition header

Unless I misinterpreted this one, and the correct behaviour is to actually respond *without* the random string, sending the original filename as part of the response.

### DIFF
--- a/wagtail/wagtailcore/utils.py
+++ b/wagtail/wagtailcore/utils.py
@@ -17,6 +17,9 @@ def resolve_model_string(model_string, default_app=None):
     """
     Resolve an 'app_label.model_name' string into an actual model class.
     If a model class is passed in, just return that.
+
+    Raises a LookupError if a model can not be found, or ValueError if passed
+    something that is neither a model or a string.
     """
     if isinstance(model_string, string_types):
         try:
--- a/wagtail/wagtailcore/utils.py
+++ b/wagtail/wagtailcore/utils.py
@@ -37,10 +37,12 @@ def resolve_model_string(model_string, default_app=None):
         return model_string
 
     else:
-        raise LookupError("Can not resolve {0!r} into a model".format(model_string), model_string)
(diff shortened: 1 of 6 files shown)


wagtail/wagtailcore/utils.py
```diff
@@ -37,10 +37,12 @@ def resolve_model_string(model_string, default_app=None):
         return model_string
 
     else:
-        raise LookupError("Can not resolve {0!r} into a model".format(model_string), model_string)
+        raise ValueError("Can not resolve {0!r} into a model".format(model_string), model_string)
```
kaedroho: Perhaps worth adding something like this to the docstring:
"Raises ValueError if the model_string is invalid. Raises a LookupError if the model_string cannot be found in the project"

timheap:
This PR has been split in to three PRs: #1793, #1794, #1795

PR #1779, opened 2015-10-05
Use os.path.splitext() instead of custom logic in Document.file_extension

timheap:
No point in rolling your own logic when it already exists in the stdlib. As a bonus, does not treat `.bashrc` as having an extension of `bashrc`.


tomdyson:
Thanks, @timheap.

PR #1785, opened 2015-10-05
Fix bug where built assets may not be packaged

timheap:
The asset compilation that occurs as part of `python setup.py sdist` would find all the files to include first, then recompile all the assets, before finally packaging them all up. This means that if a new file was created as part of the asset compilation that was not there when all the files were first searched for, that file would not be included in the package.

Now, the asset compilation happens before the file system is searched for assets to include, so all assets should always be included.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,10 +2,8 @@
 
 import sys
 
-from setuptools.command.sdist import sdist
-
 from wagtail.wagtailcore import __version__
-from wagtail.utils.setup import assets, add_subcommand, check_bdist_egg
+from wagtail.utils.setup import assets, sdist, check_bdist_egg
 
 try:
     from setuptools import setup, find_packages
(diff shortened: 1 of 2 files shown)


timheap:
This does have the consequence that the node environment has to be present to actually package the project, or to `python setup.py develop`. It may also require the node environment to `pip install`, now I think about it some more... Let me investigate some more.

timheap:
The bug where assets are not bundled if they did not already exist is certainly real, this just might not be the correct solution.

gasman:
> This does have the consequence that the node environment has to be present to actually package the project, or to `python setup.py develop`.

Isn't that the case already? We're fine with requiring node as a developer dependency. (We definitely do want to avoid it on `pip install` though.)

timheap:
I just tested then, it does now require node on `pip install`. I'll have to find another way to fix this.

timheap:
Alright, I fixed this by extending the `sdist` command instead. `sdist` is only run when doing `python setup.py sdist`, so it should only happen when explicitly packing Wagtail. I did notice that the assets are not compiled at all if running `bdist_wheel` instead of `sdist`, although it will use the already compiled assets if they are there. That is another issue though, for another PR

PR #1787, opened 2015-10-05
Update getting started guide

chimeno:
After installing wagtail with ``pip install wagtial``, install requirements.txt doesn't seem necessary.


gasman:
Hi @chimeno,

The `pip install -r requirements.txt` line is required to handle an awkward edge case:

* A user has Django 1.7 installed
* The current version of Wagtail is compatible with Django 1.7 and 1.8, so `pip install wagtail` does not upgrade Django to 1.8
* The project template created by `wagtail start` is designed for Django 1.8 only (and the requirements.txt indicates this)
* We therefore need to run `pip install -r requirements.txt` to make sure that the user is running Django 1.8.

PR #1789, opened 2015-10-05
Added router class to WagtailAPI

kaedroho:
This new class is responsible for the following:
 - Registering endpoints in an API
 - Routing requests to the correct endpoint
 - URL reversing between endpoints

The goal is to allow custom APIs to be created. This also removes the need to have a custom renderer class (so we can now use other built in DRF renderer classes)


kaedroho:
Rebased

PR #1790, opened 2015-10-05
Enable rest framework's browsable API

kaedroho:
This pull request switches on the ``BrowsableAPIRenderer`` class. This renderer displays the endpoint in a nice HTML page which is enabled when an ``Accept: text/html`` header is sent by the client (which would only be true when browsing to it directly).

### DIFF
--- a/wagtail/contrib/wagtailapi/apps.py
+++ b/wagtail/contrib/wagtailapi/apps.py
@@ -1,6 +1,10 @@
+import warnings
+
 from django.apps import AppConfig, apps
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
+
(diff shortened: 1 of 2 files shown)


kaedroho:
Rebased

gasman:
It looks like this requires adding `'rest_framework'` to INSTALLED_APPS, in order to pick up the templates - this isn't currently documented. Would it be possible to gracefully handle the case where this is missing from INSTALLED_APPS, and fall back on the old behaviour (perhaps with a deprecation warning)?

kaedroho:
@gasman Yep, I'll do that. Cheers!

kaedroho:
Changes made!

gasman:
Merged in c4738fc3d5099796f1820fd200bb3e427417f7da

PR #1793, opened 2015-10-06
Fix expectedFailures for StreamFields

timheap:
There are three commits, which fix three groups of expectedFailures:

* FieldsBlocks with ChoiceBlocks now return their display value in `get_searchable_content()`;
* StructBlocks now only render a `<dt>`/`<dd>` pair for defined fields; and
* Multiply inherited StructBlocks and StreamBlocks now correctly order their inherited children.

### DIFF
--- a/wagtail/wagtailcore/blocks/base.py
+++ b/wagtail/wagtailcore/blocks/base.py
@@ -393,7 +393,8 @@ class DeclarativeSubBlocksMetaclass(BaseBlock):
     (cheerfully stolen from https://github.com/django/django/blob/master/django/forms/forms.py)
     """
     def __new__(mcs, name, bases, attrs):
-        # Collect sub-blocks from current class.
+        # Collect sub-blocks declared on the current class.
+        # These are available on the class as `declared_blocks`
         current_blocks = []
         for key, value in list(attrs.items()):
             if isinstance(value, Block):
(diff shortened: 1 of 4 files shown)


gasman:
For 84ef41a2434dcfe152c8859832bf56c87263fae8, I think it would be cleaner to override `get_searchable_content` within ChoiceBlock rather than FieldBlock. As per https://github.com/torchbox/wagtail/pull/982#commitcomment-9799399 we decided that FieldBlock should not implement `get_searchable_content` itself, since subclasses are better placed to decide whether it's appropriate for their content to be passed to the search indexer. (Plus, it means we can eliminate the `isinstance(self.field, forms.ChoiceField)` check, of course...)

ChoiceBlock is actually a bit of a borderline case in this regard, as there are equally plausible uses where you would want its value to be indexed (e.g. a dropdown of venue names in an event block) and not (e.g. an image block with a dropdown to select whether it's left, right or centre aligned). On balance I think it's better to err on the side of giving the indexer more data, though - i.e. keep the behaviour as per the test (and this PR).

(I guess that means the comment at https://github.com/takeflight/wagtail/commit/84ef41a2434dcfe152c8859832bf56c87263fae8#diff-0532f747f350c9435cf0b8257ecca51bL80 was inaccurate, then - the current (undesired) behaviour is that `get_searchable_content` returns nothing at all.)

The other commits look good to me.

timheap:
The test as it was written was testing `FieldBlock`s using a `ChoiceField`, which is why I did it the way I did. I've moved the `get_searchable_content` logic to `ChoiceBlock` now, and I've moved the test to match. I've also added a test to `FieldBlock`s to ensure they are returning nothing from `get_searchable_content`, which is what I think you wanted. Subclasses are free to override `get_searchable_content` if they have something relevant to put in there

PR #1794, opened 2015-10-06
Fix expectedFailures for Documents

timheap:
This PR fixes two expectedFailures for Documents:

* The documents are given a random suffix to prevent filename collision, and this was not taken in to account in the tests. Serving documents with the original, uploaded filename would require a new field being added to record the initial name, which does not seem worth the cost. Now, the tests check the document is served with the suggested filename including the random suffix.
* Accessing the serve method with an incorrect filename was expected to return a 404 response, but it was instead serving the file. In my opinion, this is the correct behaviour, so the test has been updated to match, and my opinion has been added as a docstring.

Both of these are opinionated changes, but in my opinion these are the right changes to make.


kaedroho:
I agree with both, thanks!

PR #1795, opened 2015-10-06
Raise ValueError from `resolve_model_string` for unknown types

timheap:
A `LookupError` is raised by `apps.get_model` for an unknown model, but this `LookupError` was being raised when values with unhandled types were passed in. This should be a `ValueError` instead. The docstring has been amended document the errors raised.


kaedroho:
Thanks!

PR #1796, opened 2015-10-06
Add test helpers for Wagtail sites

timheap:
`wagtail.tests.utils.WagtailPageTests` is a new `TestCase` subclass that helps developers write tests for their Wagtail sites. It currently includes three assert methods:

`assertCanCreateAt(parent_model, child_model)`, which asserts that a child page of a certain type could be created underneath a parent page. This is useful for making assertions around the business rules of your site.

`assertCanNotCreateAt(parent_model, child_model)` is the inverse of the above.

`assertCanCreate(parent, child_model, data)` asserts that a child page can be added underneath the given parent page, by POSTing `data` at the correct page in the Wagtail admin. This checks that the developer has correctly configured their `content_panels` and related options.

These methods are just a start, and could be expanded further. More methods could be added, asserting that Snippets can be created, for example. The current methods could be extended further, to validate more about the Wagtail admin page editor.

### DIFF
--- a/wagtail/tests/utils.py
+++ b/wagtail/tests/utils.py
@@ -4,6 +4,7 @@ import sys
 import warnings
 from contextlib import contextmanager
 
+import django
 from django.contrib.auth import get_user_model
 from django.contrib.contenttypes.models import ContentType
 from django.core.urlresolvers import reverse
@@ -141,7 +142,10 @@ class WagtailPageTests(WagtailTestUtils, TestCase):
                 child_model._meta.app_label, child_model._meta.model_name, errors))
             raise self.failureException(msg)
 
-        explore_url = 'http://testserver' + reverse('wagtailadmin_explore', args=[parent.pk])
+        if django.VERSION >= (1, 9):
+            explore_url = reverse('wagtailadmin_explore', args=[parent.pk])
+        else:
+            explore_url = 'http://testserver' + reverse('wagtailadmin_explore', args=[parent.pk])
         if response.redirect_chain != [(explore_url, 302)]:
             msg = self._formatMessage(msg, 'Creating a page %s.%s didnt redirect the user to the explorer, but to %s' % (
                 child_model._meta.app_label, child_model._meta.model_name,
--- a/wagtail/tests/utils.py
+++ b/wagtail/tests/utils.py
@@ -6,13 +6,10 @@ from contextlib import contextmanager
 
 import django
 from django.contrib.auth import get_user_model
-from django.contrib.contenttypes.models import ContentType
 from django.core.urlresolvers import reverse
 from django.test import TestCase
 from django.utils import six
 from django.utils.text import slugify
-
-from wagtail.wagtailcore.models import Page
 
 
 class WagtailTestUtils(object):
@@ -71,13 +68,7 @@ class WagtailPageTests(WagtailTestUtils, TestCase):
         self.login()
 
     def _testCanCreateAt(self, parent_model, child_model):
-        child_ct = ContentType.objects.get_for_model(child_model)
-        parent_ct = ContentType.objects.get_for_model(parent_model)
-
-        return all([
-            child_ct in parent_model.allowed_subpage_types(),
-            # Anything can be created under a Page
-            parent_model is Page or parent_ct in child_model.allowed_parent_page_types()])
+        return child_model in parent_model.allowed_subpage_models()
 
     def assertCanCreateAt(self, parent_model, child_model, msg=None):
         """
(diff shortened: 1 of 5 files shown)


wagtail/tests/utils.py
```diff
@@ -44,3 +52,85 @@ def reset_warning_registry():
...
+        parent_ct = ContentType.objects.get_for_model(parent_model)
+
+        return all([
+            child_ct in parent_model.allowed_subpage_types(),
+            # Anything can be created under a Page
+            parent_model is Page or parent_ct in child_model.allowed_parent_page_types()])
+
+    def assertCanCreateAt(self, parent_model, child_model, msg=None):
+        """
+        Assert a particular child Page type can not be created under a parent
```
kaedroho: ...Page type can ~~not~~ be...

timheap: Oops, copy+pasta error. Fixed.

timheap:
I've packaged this up over at https://github.com/takeflight/wagtailtestutils/. You can pip install it as `pip install wagtailtestutils`

timheap:
I've added two more assert methods: `assertAllowedParentPageTypes`, and `assertAllowedParentPageTypes`; as well as some documentation.

wagtail/tests/utils.py
```diff
@@ -51,3 +60,122 @@ def reset_warning_registry():
...
+        self.login()
+
+    def _testCanCreateAt(self, parent_model, child_model):
+        child_ct = ContentType.objects.get_for_model(child_model)
+        parent_ct = ContentType.objects.get_for_model(parent_model)
+
+        return all([
+            child_ct in parent_model.allowed_subpage_types(),
+            # Anything can be created under a Page
+            parent_model is Page or parent_ct in child_model.allowed_parent_page_types()])
```
gasman: This special case for Page should be dropped now.

gasman: Also, I don't think we need to check both `parent_model.allowed_subpage_types` and `child_model.allowed_parent_page_types` - either one of these will apply the rules from both sides.

wagtail/tests/utils.py
```diff
@@ -51,3 +60,122 @@ def reset_warning_registry():
...
+            explore_url = reverse('wagtailadmin_explore', args=[parent.pk])
+        else:
+            explore_url = 'http://testserver' + reverse('wagtailadmin_explore', args=[parent.pk])
+        if response.redirect_chain != [(explore_url, 302)]:
+            msg = self._formatMessage(msg, 'Creating a page %s.%s didnt redirect the user to the explorer, but to %s' % (
+                child_model._meta.app_label, child_model._meta.model_name,
+                response.redirect_chain))
+            raise self.failureException(msg)
+
+    def assertAllowedSubpageTypes(self, parent_model, child_models, msg=None):
```
gasman: This method (along with assertAllowedParentPageTypes) should probably be renamed assertCleanSubpageTypes for consistency with the methods on Page (the 'allowed' methods apply both sides of the filtering, the 'clean' ones don't). Alternatively, if these were primarily here as a way to keep the ugly ContentType -> model conversion out of tests, maybe they aren't needed at all now?

FlipperPA:
Thank you for the work Tim, I've started playing with this and its a big help. Just a quick question: any recommendations on best practices for mocking initial data? Is doing a dumpdata of the wagtail* apps and my own to fixtures and then leading them before running tests the best method?

wagtail/tests/utils.py
timheap: These were for two reasons: to keep the ugly ContentType conversions out of user code, and to abstract the actual mechanism Wagtail uses to check these rules. These tests are primarily for users who are not experts on the Wagtail internals, easily accessible methods like these are written with those users in mind.

`clean_subpage_models` is an internal Wagtail method that is not documented. I named this public-facing method something that would make sense to those not familiar with the Wagtail internals. But now that I read this code again, I was supposed to be testing `allowed_subpage_modes`, not `clean_subpage_models`! I'll fix that right up.

timheap:
I've fixed the tests so they test the `allowed_x_models` methods now, which was the intention all along.

timheap:
@FlipperPA: I dont have much opinion on this. I dislike fixtures myself, as they can easily get out of date compared to the model definitions, and are annoying to manually update. Managing merge conflicts in fixtures is a nightmare! I usually create any pages required in code:

```python
def setUp(self):
    root_page = Site.objects.get(is_default_site=True).root_page
    self.page_to_test = MyPage(title='Page', slug='page')
    root_page.add_child(instance=self.page_to_test)
```

gasman:
Merged in 0fe3572566242201fd52c6960c77e79f1283e6e0 + parents.

PR #1799, opened 2015-10-07
Changed EMBEDLY_KEY to WAGTAILEMBEDS_EMBEDLY_KEY

anurag-ks:
https://github.com/torchbox/wagtail/issues/1757
I was introduced to wagtail few hours ago, so I am not sure if that's all the change that is required.

/wagtail/wagtail/wagtailembeds/embeds.py 
Line Number: 34
        key = settings.EMBEDLY_KEY
Line Number: 118
    if hasattr(settings, 'EMBEDLY_KEY'):

### DIFF
--- a/wagtail/wagtailembeds/embeds.py
+++ b/wagtail/wagtailembeds/embeds.py
@@ -1,5 +1,6 @@
 from datetime import datetime
 import json
+import warnings
 
 # Needs to be imported like this to allow @patch to work in tests
 from django.utils.six.moves.urllib import request as urllib_request
@@ -11,6 +12,7 @@ from django.conf import settings
 
 from wagtail.wagtailembeds.oembed_providers import get_oembed_provider
 from wagtail.wagtailembeds.models import Embed
+from wagtail.utils.deprecation import RemovedInWagtail14Warning
 
 
 class EmbedException(Exception):
@@ -31,7 +33,13 @@ def embedly(url, max_width=None, key=None):
 
     # Get embedly key
     if key is None:
-        key = settings.EMBEDLY_KEY
+        try:
+            key = settings.WAGTAILEMBEDS_EMBEDLY_KEY
+        except AttributeError:
+            key = settings.EMBEDLY_KEY
+            warnings.warn(
+                "EMBEDLY_KEY is now deprecated. Use WAGTAILEMBEDS_EMBEDLY_KEY instead",
+                RemovedInWagtail14Warning)
 
     # Get embedly client
     client = Embedly(key=key)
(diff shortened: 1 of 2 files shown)


kaedroho:
Thanks!

We need to also update Wagtail to use the new setting (with backwards compatibility + deprecation warning for the old setting): https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailembeds/embeds.py#L34

anurag-ks:
Okay on it

anurag-ks:
@kaedroho For the warning message, should I just print it to the console or do you guys use some logger ?

kaedroho:
We raise a warning using one of the classes in this file: https://github.com/torchbox/wagtail/blob/master/wagtail/utils/deprecation.py

Here's an example: https://github.com/torchbox/wagtail/blob/490ecff007ef9828947e0718a5b76559e90369e4/wagtail/wagtailcore/models.py#L272-L274

We haven't yet added a ``RemovedInWagtail14Warning`` (marking features to be removed in Wagtail 1.4). Feel free to add that class.

kaedroho:
Yep, line 118 needs to also check for the new setting:
```python
if hasattr(settings, 'WAGTAILEMBEDS_EMBEDLY_KEY') or if hasattr(settings, 'EMBEDLY_KEY'):
```

anurag-ks:
@kaedroho could you merge this now ?

wagtail/wagtailembeds/embeds.py
```diff
@@ -31,7 +33,13 @@ def embedly(url, max_width=None, key=None):
 
     # Get embedly key
     if key is None:
-        key = settings.EMBEDLY_KEY
+        try:
+            key = settings.WAGTAILEMBEDS_EMBEDLY_KEY
+        except AttributeError:
+            warnings.warn(
+                "use WAGTAILEMBEDS_EMBEDLY_KEY instead",
```
kaedroho: We should give this message some more context:
"EMBEDLY_KEY is now deprecated. Use WAGTAILEMBEDS_EMBEDLY_KEY instead"

wagtail/wagtailembeds/embeds.py
```diff
@@ -11,6 +12,7 @@
 
 from wagtail.wagtailembeds.oembed_providers import get_oembed_provider
 from wagtail.wagtailembeds.models import Embed
+from wagtail.utils.deprecation import RemovedInWagtail13Warning
```
gasman: We should use `RemovedInWagtail14Warning` here instead - we want to give two versions' notice before this is removed.

(`RemovedInWagtail14Warning` has just been added, in 9197bf1259df43a9564495a6b41fb733b025d804)

anurag-ks:
Resending again

anurag-ks:
@kaedroho done.

Issue #1638, opened 2015-08-24
The search bar only searches for pages


creativedutchmen:
The search bar in the navigation on the left only searches for pages by default. When a page is found, the option to search in other sections (images, documents) is shown. However, this functionality is hidden when no page is found.

To me, this sounds backwards. No pages are found, so chances are the user wasn't searching for a page in the first place. Displaying other areas to be searched would be very helpful.

The same holds for images and documents: if there is a match, the other sections are shown. If there is no match, they are hidden. In my opinion they should be shown at all times, making search more user friendly.

gasman:
Makes sense - thanks for the suggestion!

alexgleason:
Do snippets get searched too?

gasman:
Implemented in #1800

Issue #1701, opened 2015-09-17
"Other searches" on main search view should be hookable


kaedroho:
We have a few hard-coded URLs on the search view: https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailadmin/templates/wagtailadmin/pages/search_results.html#L13-L18

I think these should be made hookable to allow new items to be added and also swapped out.

gasman:
Fixed in #1800

PR #1800, opened 2015-10-07
Issue #1638: The (admin) search bar only searches for pages

bennymak:
First (ever) pull request, so advice/guidance/feedback appreciated.

This is a fix for issue #1638, but might be a bit over the top. Originally I had just planned to tweak the template, but once I realised each app/module had its own search template I thought I'd try to go for something a bit more generic than a static template snippet.

So, borrowing/cloning the implementation of the menu hooks, I added a new hook for modules to offer their search page as an option in the "Other Searches" block:

image

Regarding the core of the issue, the "other searches" option now shows itself both if there are and aren't results returned (but not if it's a blank search) and the "other searches" shows on each apps/modules search page via a template tag (search_other).

While the hook contains a fair bit of duplicate code from the menu hook, it didn't feel appropriate to just utilise the existing hook model. In future it could be expanded to provide a unified admin search page.

Anyway, this went a bit further than I was planning to on my first contribution. Hopefully the implementation isn't too bad.

### DIFF
--- a/wagtail/wagtailadmin/search.py
+++ b/wagtail/wagtailadmin/search.py
@@ -5,6 +5,7 @@ from django.forms.utils import flatatt
 from django.utils.text import slugify
 from django.utils.safestring import mark_safe
 from django.utils.six import text_type
+from django.utils.functional import cached_property, total_ordering
 
 from django.utils.six import with_metaclass
 
@@ -13,6 +14,7 @@ from wagtail.wagtailcore import hooks
 from wagtail.wagtailadmin.forms import SearchForm
 
 
+@total_ordering
 class SearchArea(with_metaclass(MediaDefiningClass)):
     template = 'wagtailadmin/shared/search_area.html'
 
@@ -28,6 +30,12 @@ class SearchArea(with_metaclass(MediaDefiningClass)):
         else:
             self.attr_string = ""
 
+    def __lt__(self, other):
+        return (self.order, self.label) < (other.order, other.label)
+
+    def __eq__(self, other):
+        return (self.order, self.label) == (other.order, other.label)
+
     def is_shown(self, request):
         """
         Whether this search area should be shown for the given request; permission
@@ -35,17 +43,20 @@ class SearchArea(with_metaclass(MediaDefiningClass)):
         """
         return True
 
-    def is_active(self, request):
-        return request.path.startswith(self.url)
+    def is_active(self, request, current=None):
+        if current is None:
+            return request.path.startswith(self.url)
+        else:
+            return self.name == current
 
-    def render_html(self, request, query):
+    def render_html(self, request, query, current=None):
         return render_to_string(self.template, {
             'name': self.name,
             'url': self.url,
             'classnames': self.classnames,
             'attr_string': self.attr_string,
             'label': self.label,
-            'active': self.is_active(request),
+            'active': self.is_active(request, current),
             'query_string': query
         }, request=request)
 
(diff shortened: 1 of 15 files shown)


davecranwell:
Wow this is excellent! Thanks @bennymak.

davecranwell:
I'll leave the main feedback to to our Django devs, but my more generic request would be to please add a) unit tests b) documentation for the hook

kaedroho:
Thanks!

This also solves this issue #1701

wagtail/wagtailadmin/search.py
```diff
@@ -0,0 +1,103 @@
...
+        # registered_search_areas property. We can't populate it in __init__ because
+        # we can't rely on all hooks modules to have been imported at the point that
+        # we create the admin_search and settings_search instances
+        self._registered_search_areas = None
+
+    @property
+    def registered_search_areas(self):
+        if self._registered_search_areas is None:
+            self._registered_search_areas = [fn() for fn in hooks.get_hooks(self.register_hook_name)]
+        return self._registered_search_areas
```
timheap: If this used [`@cached_property`](https://docs.djangoproject.com/en/1.8/ref/utils/#django.utils.functional.cached_property), and sorted the list of search areas:
```
@cached_property
def registered_search_areas(self):
    return sorted(
        [fn() for fn in hooks.get_hooks(self.register_hook_name)],
        key=lambda s: s.order)
```
Then you wouldn't have to [sort the list of search areas](https://github.com/torchbox/wagtail/pull/1800/files#diff-b1c5845438a86b2864088e26ac56596aR97) in every call to `render_html()`

timheap: Also, if the `__lt__()` method were defined on the `SearchArea` class, and the [`total_ordering`](https://docs.python.org/3/library/functools.html#functools.total_ordering) class decorator added, the `SearchArea`s could be ordered without using `key=lambda s: s.order` at all
```
def __lt__(self, other):
    return (self.order, self.label) < (other.order, other.label)
def __eq__(self, other):
    return (self.order, self.label) == (other.order, other.label)
```

wagtail/wagtailadmin/wagtail_hooks.py
```diff
@@ -30,3 +31,16 @@ def register_settings_menu():
 @hooks.register('register_permissions')
 def register_permissions():
     return Permission.objects.filter(content_type__app_label='wagtailadmin', codename='access_admin')
+
+
+class PagesSearchArea(SearchArea):
+    pass
```
timheap: Does this need to be its own subclass, or could the register line below use a plain `SeachArea`?

wagtail/wagtailadmin/templates/wagtailadmin/shared/search_area.html
```diff
@@ -0,0 +1,6 @@
+{% if not active %}
```
timheap: Now that the 'Other searches' bar sticks around, having the search types appear and disappear as you click through them makes checking through all the search types quite difficult. As 'Document' is much longer than 'Page', switching from one to the other makes all subsequent search types shift on screen dramatically. Could the search type bar be changed to a 'Category', 'Search for:', or other wording; and the current search type stay in the list, but bolded/not linked to show it was the current search type?

wagtail/wagtailadmin/templatetags/wagtailadmin_tags.py
```diff
@@ -39,6 +40,17 @@ def main_nav(context):
         'request': request,
     }
 
+
+@register.inclusion_tag('wagtailadmin/shared/search_other.html', takes_context=True)
+def search_other(context):
```
timheap: This tag could take a single argument for the currently active search type: `{% search_other current="pages" %}`:
```
@register.inclusion_tag('wagtailadmin/shared/search_other.html', takes_context=True)
def search_other(context, current=None):
    request = context['request']

    return {
        'options_html': admin_search_areas.render_html(request, current),
        'request': request,
    }
```

This could then be passed through when rendering [`SearchArea.render_html`](https://github.com/torchbox/wagtail/pull/1800/files#diff-b1c5845438a86b2864088e26ac56596aR98):
```
def render_html(self, request, current=None):
    # ...
    for item in sorted(search_areas, key=lambda i: i.order):
        rendered_search_areas.append(item.render_html(request, query, current=item.name == current))
    # ...
```

This way, matching the current item does not have to depend upon the URL, and the 'Other searches' bar could be shown on any page where it might be relevant, without losing the ability to mark a search type as current.

timheap:
I really like this idea. Making it flexible enough to register your own search types means third party plugins can register their own search types as well - excellent! If this gets merged in, I will make sure `wagtailnews` integrates with it.

I had a read through your changes, and wrote down some of my thoughts.

wagtail/wagtailadmin/templates/wagtailadmin/shared/search_area.html
bennymak: Agreed, it is a bit awkward with the disappearing options. I'll see what I can do to differentiate, trying out `icon-larger nolink` on the active link (but still a link). The larger icon does shuffle the list around a bit though.

I'm not sure about renaming it and how that affects translations. The few languages I checked have translations for "other searches" specifically.

SalahAdDin:
:+1:

timheap:
Looks good to me!

gasman:
Merged in da4f0914660043303ae12aade5093233c46a5f99 + parents - thanks!

tomdyson:
@bennymak Thank! This is a great addition to Wagtail. Can you email me? My first name at torchbox.com.

PR #1496, opened 2015-07-09
Refactor pagination into a helper

timheap:
Pagination is done almost exactly the same across all of Wagtail. Moving
the repeated code in to one spot ensures mistakes are not made, and
means that pagination is always done consistently.

There are some locations where the pagination helper has not been used,
as there is no request available or something similar. These instances
are special enough that refactoring them does not make sense at this
point.

PR #1803, opened 2015-10-08
Make sure we have a paginator before trying to use it

### LINKED ISSUE #1609: Page reordering interface displays "Page of ." in the footer
screen shot 2015-08-14 at 13 57 31

elbaloo:
Fixes #1609


gasman:
Hi @elbaloo - it seems that this was fixed as part of #1496 at almost exactly the same time you submitted this, but thanks for tracking this down!

PR #1812, opened 2015-10-09
Allow block subclasses to pass extra context variables to templates

gasman:
Alternative approach to #1771.

Block now defines a get_context method, which subclasses can override
in the same way as we can for pages.

### DIFF
--- a/wagtail/tests/testapp/blocks.py
+++ b/wagtail/tests/testapp/blocks.py
@@ -1,0 +1,14 @@
+from wagtail.wagtailcore import blocks
+
+
+class LinkBlock(blocks.StructBlock):
+    title = blocks.CharBlock()
+    url = blocks.URLBlock()
+
+    def get_context(self, value):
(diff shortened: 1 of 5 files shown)


wagtail/tests/testapp/blocks.py
```diff
@@ -0,0 +1,14 @@
+from wagtail.wagtailcore import blocks
+
+
+class LinkBlock(blocks.StructBlock):
+    title = blocks.CharBlock()
+    url = blocks.URLBlock()
+
+    def get_context(self, value):
+        context = super(LinkBlock, self).get_context(value)
+        context['classname'] = 'important' if value['title'] == 'Torchbox' else ['normal']
```
timheap: `['normal']` should not be a list, just a plain string.

gasman: Oops... fixed/rebased.

PR #1813, opened 2015-10-10
Fix JavaScript style error causing CI failures

timheap:
The Drone JS style checker was complaining about the bad quotes.


timheap:
Could `master` be made in to a [protected branch](https://help.github.com/articles/about-protected-branches/) to stop this kind of thing happening in the future? If `master` was set to be protected, people would not be able to push commits to it that have not yet passed CI. Commits would have to be pushed to another branch / PR first, to allow CI to run, and then they could be merged in.

Issue #1194, opened 2015-04-16
Sites need a title/name field


davecranwell:
Site hostnames are often output on the front end when generating Open Graph tags, and other places where absolute URLs or "official urls" of the site are needed.

Additionally the "friendly" name of the site is also required for OG tags or `title` tags: `<title>My page about things | My Site</title>`. At the moment we have to hard-coded this friendly site name, meaning templates can't be reused as easily.

It would be good if the `Sites` object were updated to include a human friendly name field for the site. Obviously the `WAGTAIL_SITE_NAME` is insufficient if the installation is hosting multiple domains.

SalahAdDin:
:+1:

SalahAdDin:
Man, please, review this, you can see in this pic how see the sitename in the login page:
![cambiassitename](https://cloud.githubusercontent.com/assets/5034215/9232249/4847560a-40f2-11e5-8ec9-ec77cec4f40b.png)

timorieber:
+1

SalahAdDin:
:+1:

gasman:
Implemented in #1814.

SalahAdDin:
Thank you very much, Did you put this in the documentation?

PR #1814, opened 2015-10-10
Add site name to sites

timorieber:
This PR addresses #1194. Right now this PR simply adds `site_name` to the `Site` model and form.

As #1194 is currently labeled "For team discussion" here are some thoughts of mine:

* Should `site_name` be mandatory? If yes, what do you think about the migration/initial data? Simply use `hostname`?
* How do you think about the field ordering? If it were mandatory I would put it at first position.
* Should the page explorer show the site a page belongs to? At least on the root pages authors could benefit from this contextual information.
* I think there is a semantic impact on `WAGTAIL_SITE_NAME`. It now has a more generic meaning for the whole wagtail instance as every site has its own name. Nevertheless I kept the usages "as is", because I found it only used at the admin home screen, which is "instance-wide". Perhaps the setting should be renamed in a future release.

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -58,9 +58,30 @@ class SiteManager(models.Manager):
 @python_2_unicode_compatible
 class Site(models.Model):
     hostname = models.CharField(verbose_name=_('Hostname'), max_length=255, db_index=True)
-    port = models.IntegerField(verbose_name=_('Port'), default=80, help_text=_("Set this to something other than 80 if you need a specific port number to appear in URLs (e.g. development on port 8000). Does not affect request handling (so port forwarding still works)."))
+    port = models.IntegerField(
+        verbose_name=_('Port'),
+        default=80,
+        help_text=_(
+            "Set this to something other than 80 if you need a specific port number to appear in URLs (e.g. "
+            "development on port 8000). Does not affect request handling (so port forwarding still works)."
+        )
+    )
+    site_name = models.CharField(
+        verbose_name=_('Site name'),
+        max_length=255,
+        db_index=True,
+        null=True,
+        blank=True,
+        help_text=_("Human-readable name for the site.")
+    )
     root_page = models.ForeignKey('Page', verbose_name=_('Root page'), related_name='sites_rooted_here')
-    is_default_site = models.BooleanField(verbose_name=_('Is default site'), default=False, help_text=_("If true, this site will handle requests for all other hostnames that do not have a site entry of their own"))
+    is_default_site = models.BooleanField(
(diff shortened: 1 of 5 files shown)


davecranwell:
- I'd say `site_name` doesn't need to be mandatory and that `hostname` is a sufficient default. Many people may not have a need for it. Indeed my reasons for requiring it concern quite specific third party needs, which not everyone will have.
 - `site_name` should be first field I suppose. Just as the `title` of any document, page, image etc comes first.
 - I like the idea of the domain being visible in some way on the front end although it would only be relevant for installations with multiple domains, which may not be so common.  I'd probably consider a small line of grey meta text showing site name or domain name, below the page title in the explorer with a globe icon prefix. Just an idea. Perhaps this is something for us to mull over once merged.
 - `WAGTAIL_SITE_NAME` could do with a review. I'd say it's not worth worrying about for this PR though.

tomdyson:
Thank you very much for this PR and your thoughtful questions, @timorieber.

timorieber:
You're welcome!

When `site_name` remains optional I suggest leaving the field order "as is", because in my opinion:

* The order should be the same on index/add/edit views
* The possibly empty field `site_name` should not be the first one, especially on the index view
* The first field in a form should be mandatory in most cases

The other topics ("site name in page explorer" and "rename setting") can  be implemented in separate PR's.

Regards,
Timo

timorieber:
Rebased on master which now includes #1813 to fix JS code style errors.

wagtail/wagtailcore/models.py
```diff
@@ -58,9 +58,30 @@ def get_by_natural_key(self, hostname, port):
...
+        default=80,
+        help_text=_(
+            "Set this to something other than 80 if you need a specific port number to appear in URLs (e.g. "
+            "development on port 8000). Does not affect request handling (so port forwarding still works)."
+        )
+    )
+    site_name = models.CharField(
+        verbose_name=_('Site name'),
+        max_length=255,
+        db_index=True,
```
gasman: Will leave this db_index out, as I don't think it's likely that code will regularly be doing lookups by human-readable name. (Our title fields elsewhere don't have an index.)

gasman:
Merged in c65b155f2f4d692f0ba6ab169507ab4c2bdbb816 + parents - thanks!

SalahAdDin:
Did you Put this in the documentation???

kaedroho:
Thanks for spotting @SalahAdDin 

Looks like the Site model reference docs need an update: http://docs.wagtail.io/en/v1.3/reference/pages/model_reference.html#id1

SalahAdDin:
But, site name reference aren't in these page.

PR #1815, opened 2015-10-12
Allow custom ordering in search

kaedroho:
This pull request adds a new attribute to search methods: ``order_by_relevance``. This is a boolean that defauls to ``True``.

When set to ``False`` the ordering that is already used on the queryset is also used in the search.

For example, this will search blog pages but order them by published date:
```
blogs = BlogPage.objects.order_by('-published_date').search("Test", operator='and', order_by_relevance=False)
```

### DIFF
--- a/docs/topics/search/searching.rst
+++ b/docs/topics/search/searching.rst
@@ -138,7 +138,6 @@ Here's an example of using the "operator" keyword argument:
     # Only "hello world" returned as that's the only item that contains both terms
     [<Thing: Hello world>]
 
-
 For page, image and document models, the ``operator`` keyword argument is also supported on the QuerySet's ``search`` method:
 
 .. code-block:: python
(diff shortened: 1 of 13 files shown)


kaedroho:
Rebased and ready for review!

gasman:
As it stands, this PR doesn't implement (or test) the functionality as described in the docs, namely adding `order_by_relevance` to the QuerySet.search method - it's only implemented at the backend level.

On that topic (and as a way of killing two birds with one stone, since something similar is going to come up with the queryset tree methods when I rebase #1751) - what do you think about moving the QuerySet.search method (currently duplicated across Page, Image and Document) into a `SearchableQuerySetMixin` in wagtailcore.query?

gasman:
Have put the above suggested fixes into a sub-PR, https://github.com/kaedroho/wagtail/pull/24

PR #1816, opened 2015-10-12
Django 1.9 support

kaedroho:


kaedroho:
Ready for review!

wagtail/wagtailadmin/edit_handlers.py
```diff
@@ -695,10 +696,15 @@ def __init__(self, relation_name, panels=None, label='', help_text='', min_num=N
         self.max_num = max_num
 
     def bind_to_model(self, model):
+        if django.VERSION >= (1, 9):
```
timheap: Is it worth adding this to `wagtail.utils.compat`? There are already two similar methods in there from the changes between 1.7/1.8.

timheap:
Looks good to me!

Are you considering dropping Django 1.7 support in favour of 1.9? Testing three version of Django will make the Travis tests take an age.

kaedroho:
Thanks @timheap!

That's not the plan yet. But I think it's worth considering since Django 1.7 is now end of life and upgrading to Django 1.8 is generally quite easy.

kaedroho:
After updating two client sites, both are giving me an error: ``ValueError: Related model u'Filter' cannot be resolved``. Will investigate further.

kaedroho:
Above issue looks like a Django bug which has been reported here: https://code.djangoproject.com/ticket/25858. I was able to work around it easily though.

Next and (hopefully) last issue is a change in the way template tag libraries are loaded (https://github.com/django/django/commit/655f52491505932ef04264de2bce21a03f3a7cd0#diff-6492e208d64b891c181c85a794d0b303R142) causing Django to try to load jinja2 tags and fail if jinja2 is not installed

kaedroho:
All issues fixed, ready for review!

gasman:
Have just uploaded django-modelcluster version 1.1b1 (with the Django 1.9 fixes added) to PyPI, so we can put this in Wagtail's setup.py now - I propose we release django-modelcluster 1.1 final at the same time as Wagtail 1.3.

Other than adding a test for whitespace stripping, all looks good to me. @timheap, are you happy with the jinja2 templatetag workaround? (I guess we'll need an 'upgrade considerations' note to cover that)

MechanisM:
How about setup.py file?
it still wants "Django>=1.7.1,<1.9", and forcing to replace 1.9 with 1.8.7

timheap:
The jinja2 solution sounds sensible. I dont know of a simple way to make it backwards compatible with a deprecation warning, so just moving it to a better location might be the best solution. Jinja support has only been out for one version, so not many people will be using it yet.

gasman:
Some additions / fixups in https://github.com/gasman/wagtail/tree/kaedroho-dj19 - please merge if happy!

timheap:
I've also got a commit to fix #2014 at https://github.com/takeflight/wagtail/tree/dj19

kaedroho:
All commits merged in, thanks both!

gasman:
Merged in 0ce27f2f3d10279498c12b07d7d26fc973987b38 + parents

PR #1821, opened 2015-10-13
Remove HTML comments from rich text

kaedroho:
When text is pasted into a rich text block from MS word, HTML comments are created that contain lots of data that we don't need.

This commit changes the whitelister to remove all comment nodes.


gasman:
Looks good to me.

At the back of my mind I wonder whether this might be a breaking change for some people - I could easily imagine people hacking up systems where they insert placeholder comments into the field using the `hallohtml` extension, and have a search-and-replace going on in the template. Knowing how hard it is to persuade people that the rich text component *isn't* an arbitrary HTML authoring tool, and their determination to hack around its restrictions, I wouldn't be at all surprised to find out that there are hacks like that in the wild already...

But on reflection, we have no obligation to ensure that those hacks continue to work across releases, and forcing those people to re-evaluate their code and redo things in a less hacky way would actually be a good thing. So let's go for it!

PR #1822, opened 2015-10-13
Removing need for an image rendition required by wagtail out the box by reusing and shrinking another.

davecranwell:


tomdyson:
Good idea @davecranwell.

PR #1828, opened 2015-10-15
Docs fix: Need to flip order of class creation

jheasly:
Need to flip order of creating `BlogIndexRelatedLink` and `BlogIndexPage`, otherwise `BlogIndexRelatedLink` references `BlogIndexPage` before it's been created and you get:
```bash
ERRORS:
blog.BlogIndexRelatedLink.page: (fields.E300) Field defines a relation with model 'BlogIndexPage', which is either not installed, or is abstract.
```

### DIFF
--- a/docs/getting_started/tutorial.rst
+++ b/docs/getting_started/tutorial.rst
@@ -364,10 +364,6 @@ can be BlogPages or external links. Change ``blog/models.py`` to
             abstract = True
 
 
-    class BlogIndexRelatedLink(Orderable, RelatedLink):
-        page = ParentalKey('BlogIndexPage', related_name='related_links')
-
-
     class BlogIndexPage(Page):
         intro = RichTextField(blank=True)
 
@@ -375,6 +371,10 @@ can be BlogPages or external links. Change ``blog/models.py`` to
             FieldPanel('intro', classname="full"),
             InlinePanel('related_links', label="Related links"),
         ]
+
+
+    class BlogIndexRelatedLink(Orderable, RelatedLink):
+        page = ParentalKey('BlogIndexPage', related_name='related_links')
 
 .. figure:: ../_static/images/tutorial/tutorial_7.png
    :alt: Blog index edit screen


gasman:
Thanks for the report! The `BlogIndexRelatedLink` definition *ought* to work in its original place, as long as `'BlogIndexPage'` is in quotes - I'll investigate further.

kaedroho:
FWIW, I think child/inline models should be placed directly beneath the parent model anyway. Looks a bit nicer to have the important bit at the top.

gasman:
I'm unable to reproduce this error - the code in the tutorial works fine for me as is, and missing quotes isn't the issue (removing them produces a straightforward `NameError: name 'BlogIndexPage' is not defined` error). However, I agree with @kaedroho that putting `BlogIndexRelatedLink` after is neater anyway, so I'll merge this change.

jheasly:
Hrm. Now when I do a `runserver` I get these duplicate `RuntimeWarning`s:
```bash
(wagtail)internet02:wt_site jheasly$ python manage.py runserver
/Users/jheasly/.virtualenvs/wagtail/lib/python2.7/site-packages/django/db/models/base.py:309: RuntimeWarning: Model 'blog.blogindexpage' was already registered. Reloading models is not advised as it can lead to inconsistencies, most notably with related models.
  new_class._meta.apps.register_model(new_class._meta.app_label, new_class)

/Users/jheasly/.virtualenvs/wagtail/lib/python2.7/site-packages/django/db/models/base.py:309: RuntimeWarning: Model 'blog.blogindexpage' was already registered. Reloading models is not advised as it can lead to inconsistencies, most notably with related models.
  new_class._meta.apps.register_model(new_class._meta.app_label, new_class)

Performing system checks...

```
... but when I move the `BlogIndexPage` class definition after `BlogIndexRelatedLink` as was in the docs, it instantly won't run, giving me the aforementioned
```bash
ERRORS:
blog.BlogIndexRelatedLink.page: (fields.E300) Field defines a relation with model 'BlogIndexPage', which is either not installed, or is abstract.
```
Can you reproduce?

PR #1234, opened 2015-04-23
Image upload into listing

davecranwell:
Allows drag-drop of images directly into the image listing, rather than requiring user goes to another page to do so. 

Moves the file upload plugin to wagtailadmin and wraps it in a generic way as a precursor to allowing upload directly into page editor interface, or image chooser dialogue, or indeed for use in document uploading too.

52493b6cca727ce3e85e6cd73a2841d7

(gif fails to demonstrate animated pop-in effect of newly added image)

PR #1830, opened 2015-10-15
added success message after image update

shredding:
My PM asked me many-a-time, if the image is now really stored, because it's simply gone after clicking the update button in the bulk-image-upload.

Can we add a ui feedback like this?


davecranwell:
Hi @shredding thanks for addressing this. 

However there's an existing PR to remove the separate upload interface altogether, instead having users upload images directly into the image listing. That gives them the feedback they need as the images upload: https://github.com/torchbox/wagtail/pull/1234

I think I'd rather we take #1234's approach rather than trying to main a separate interface any more.

Please feel free to try out 1234 and perhaps even add your tweaks to that instead.

shredding:
perfect.

gasman:
It's not clear to me what the current status of #1234 is, and what the prospects of it are of it getting completed/merged any time soon. Until that's determined, I don't think we should put off making incremental improvements to the existing interface, so I'll reopen this.

davecranwell:
Fair point re #1234. It's not something I'm going to have time to address in the near future and the changes @shredding has done here are so minor they might as well be merged in the meantime.

davecranwell:
Note that #1234 is now complete.

gasman:
Merged in 6422c7f61869084fae8f2e0de9e24d56ef9e1729 - thanks!

PR #1482, opened 2015-07-07
Tweaks to how root pages are rendered in explorer

davecranwell:
a) to prevent misleading ability to choose root as a page link
b) to prevent page type of root being displayed

PR #1837, opened 2015-10-16
Add flag to AdminPageChooser to specify whether the root page can be chosen

gasman:
Supersedes #1482.

Normally we don't want root to be a choosable option in the AdminPageChooser interface, and so this PR makes it unchoosable by default. However, there are some exceptions where root should be choosable, namely setting tree-wide permissions in the groups admin, and selecting root as the destination of a page copy, and so we implement a `can_choose_root` flag to override the default for these.

Since it's conceivable that end users might also have a valid use for page choosers that let you choose the root, we also make the `can_choose_root` flag available on PageChooserPanel and PageChooserBlock.

### DIFF
--- a/docs/reference/pages/panels.rst
+++ b/docs/reference/pages/panels.rst
@@ -94,7 +94,7 @@ FieldRowPanel
 PageChooserPanel
 ----------------
 
-.. class:: PageChooserPanel(field_name, page_type=None)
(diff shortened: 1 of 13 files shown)


gasman:
jslint errors fixed, and rebased on top of #1882 (since the querystring bugs fixed there also affect the new can_choose_root param)

PR #1838, opened 2015-10-16
Removed Vagrantfile from startproject command

timorieber:
Cleaned up another small relict from vagrant times...


gasman:
Good catch - thanks!

Issue #1443, opened 2015-06-26
Non-arabic scripts (Japanese) do not auto-populate slug


minikomi:
This was partially addressed in #8, but I'm evaluating wagtail for creating pages for Japan, and unfortunately django's URLify (understandably) doesn't stretch as far to include Japanese. I'm wondering what the best plan of attack is. Some ideas:

* If the title is non-blank, but urlify returns a blank string, use something else?
  * Date? 
  * Post ID
  * Configurable from settings or model
* Just show a warning you need to set the slug too
* Show the current slug in the main panel under the title - perhaps clicking on it could bring up the second tab and focus the slug input

gasman:
I've previously used the [unidecode](https://pypi.python.org/pypi/Unidecode) library to produce passable ASCII representations of non-Latin text - maybe that could be used here. (I don't know how well it handles kanji in general, though - I only used it on place names. And maybe this is just pushing the problem further down the road, and eventually we'll encounter another language that unidecode can't handle...)

I guess there's no fundamental reason why slugs have to be ASCII-only - most browsers now display unicode characters in URLs quite happily - so perhaps the answer is simply to use something more permissive than django's urlify. (What do Japanese news/blog sites usually do for URLs - native script, romanized text, or avoid the issue by just using IDs?)

minikomi:
I've seen all 3 used before. To me, romanize versions of th Japanese titles
are the worst option - I can't see it having any SEO benefit whatsoever and
doesn't really make the URLs any more typable.

To be honest, this is something I feel might be better handled in the model
rather than on the front end. There's lots of cases where a datestamp would
be good enough or perhaps just the post's id.

Eg - if the slug is blank run generate_slug method, if that returns blank
string or none raise validation error.

Actually, using the routableurls mixin and an index page, I can see cases
where the slug is never used and the url comes from eg. The post date / ID
and lookup is ID based.

kaedroho:
A commit recently landed in Django adding support for unicode characters in ``SlugField``: https://github.com/django/django/commit/f8cc464452f495fce2a3d6f7494396c8f798a1e6

hanpama:
```
In [2]: from demo.models import BlogPage

In [3]: BlogPage.objects.all()
Out[3]: [<BlogPage: 블로그를 열었습니다>, <BlogPage: 두 번째 블로그입니다>, <BlogPage: 세번째 블로그입니다.>]

In [4]: BlogPage.objects.all()[0]
[...]

PR #1843, opened 2015-10-19
Remove the length argument to URLify

saevarom:
The URLify function supplied by django does not need a length argument in order to work. The problem with supplying the length argument is that when you have Unicode characters that are translated to two characters in the slug version, the resulting slug will be missing one character for each such double character.

Example:
Æ or æ is translated as ae when slugified.
A call to URLify would then be: URLify("kjaftæði", 8) which would result in the slug "kjaftaed", since the slug is 9 characters in length due to æ being expanded to two characters. URLify("kjaftæði") works fine and returns "kjaftaedi".


gasman:
Thanks! It looks like we probably meant to write `URLify(val, 255)` - but the input field enforces a max length of 255 anyway, so there's no need to specify the limit in the JS.

Incidentally, we intend to drop the use of URLify completely, as the pre-Django-1.9 version doesn't support non-Latin alphabets (#1443) - this will be a useful fix to have in the meantime, though.

PR #1845, opened 2015-10-19
Added support for AWS CloudFront in Frontend cache invalidation module

robmoorman:
Cache invalidation support for CloudFront, the Content Delivery Network of Amazon Web Services.

---
This change is [Reviewable](https://reviewable.io/reviews/torchbox/wagtail/1845)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,7 @@ except ImportError:
 
 
 install_requires = [
+    "boto3>=1.1,<1.2",
     "Django>=1.8.1,<1.10",
     "django-modelcluster>=2.0,<3.0",
     "django-taggit>=0.18,<0.19",
--- a/wagtail/contrib/wagtailfrontendcache/backends.py
+++ b/wagtail/contrib/wagtailfrontendcache/backends.py
@@ -96,17 +96,19 @@ class CloudfrontBackend(BaseBackend):
 
     def purge(self, url):
         try:
+            url_parsed = urlparse(url)
+            path = url_parsed.path
             self.client.create_invalidation(
                 DistributionId=self.cloudfront_distribution_id,
                 InvalidationBatch={
                     'Paths': {
                         'Quantity': 1,
                         'Items': [
-                            url,
+                            path,
                         ]
                     },
                     'CallerReference': str(uuid.uuid4())
                 }
             )
         except botocore.exceptions.ClientError as e:
-            logger.error("Couldn't purge '%s' from Cloudfront. ClientError: %s %s", url, e.response['Error']['Code'], e.response['Error']['Message'])
+            logger.error("Couldn't purge '%s' from Cloudfront. ClientError: %s %s", path, e.response['Error']['Code'], e.response['Error']['Message'])
--- a/wagtail/contrib/wagtailfrontendcache/backends.py
+++ b/wagtail/contrib/wagtailfrontendcache/backends.py
@@ -5,7 +5,6 @@ import botocore
 import json
 import logging
 import uuid
-import six
 
 from django.core.exceptions import ImproperlyConfigured
 from django.utils.six.moves.urllib.error import HTTPError, URLError
@@ -103,11 +102,27 @@ class CloudfrontBackend(BaseBackend):
             )
 
     def purge(self, url):
-        try:
-            url_parsed = urlparse(url)
+        url_parsed = urlparse(url)
+        distribution_id = None
+
+        if type(self.cloudfront_distribution_id) is dict:
+            host = url_parsed.hostname
+            if self.cloudfront_distribution_id.has_key(host):
+                distribution_id = self.cloudfront_distribution_id.get(host)
+            else:
+                raise ImproperlyConfigured("hostname %s" % host)
+                logger.error("Couldn't purge '%s', hostname '%s' not found in mapping", url, host)
+        else:
+            distribution_id = self.cloudfront_distribution_id
+
+        if distribution_id:
             path = url_parsed.path
+            self._create_invalidation(distribution_id, path)
+
+    def _create_invalidation(distribution_id, path):
+        try:
             self.client.create_invalidation(
-                DistributionId=self.cloudfront_distribution_id,
+                DistributionId=distribution_id,
                 InvalidationBatch={
                     'Paths': {
                         'Quantity': 1,
@@ -119,4 +134,4 @@ class CloudfrontBackend(BaseBackend):
                 }
             )
         except botocore.exceptions.ClientError as e:
-            logger.error("Couldn't purge '%s' from Cloudfront. ClientError: %s %s", path, e.response['Error']['Code'], e.response['Error']['Message'])
+            logger.error("Couldn't purge '%s' from CloudFront. ClientError: %s %s", path, e.response['Error']['Code'], e.response['Error']['Message'])
--- a/wagtail/contrib/wagtailfrontendcache/backends.py
+++ b/wagtail/contrib/wagtailfrontendcache/backends.py
@@ -97,8 +97,7 @@ class CloudfrontBackend(BaseBackend):
             self.cloudfront_distribution_id = params.pop('DISTRIBUTION_ID')
         except KeyError:
             raise ImproperlyConfigured(
-                "The setting 'WAGTAILFRONTENDCACHE' requires the object "
-                "'DISTRIBUTION_ID'."
+                "The setting 'WAGTAILFRONTENDCACHE' requires the object 'DISTRIBUTION_ID'."
             )
 
     def purge(self, url):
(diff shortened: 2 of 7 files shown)


setup.py
```diff
@@ -21,6 +21,7 @@
```
kaedroho: Why the mode change for this file?

wagtail/contrib/wagtailfrontendcache/backends.py
```diff
@@ -79,3 +82,27 @@ def purge(self, url):
...
+                        'Items': [
+                            url,
+                        ]
+                    },
+                    'CallerReference': str(uuid.uuid4())
+                }
+            )
+        except botocore.exceptions.ClientError as e:
+            logger.error("Couldn't purge '%s' from Cloudfront. ClientError: %s %s", url, e.response['Error']['Code'], e.response['Error']['Message'])
+            return
```
prydie: Method will return anyway here without explicit `return` statement.

kaedroho:
Thanks! This is really nice work. Sorry it's taken me a while to get round to testing it.

After some testing, it appears that CloudFront doesn't take full URLs, we can only send it the path component of the URL.

``purge_page_from_cache()`` passes in the full URL to ``purge()`` so it doesn't work at the moment:

```
>>> purge_page_from_cache(p)
Couldn't purge 'http://<snip>.cloudfront.net/test/' from Cloudfront. ClientError: InvalidArgument Your request contains one or more invalid invalidation paths.
```

However, purging ``/test/`` using ``purge_url_from_cache`` works fine:
```
>>> purge_page_from_cache('/test/')
<no output>
```

We could, of course, just snip off the hostname and scheme from the URL, but I think we could do better to support multi-site installations.

How about allowing users to configure a mapping of hostname => distribution ids in their settings? (but still support the current single distribution id approach for single-site installations)

For example:
```python
    WAGTAILFRONTENDCACHE = {
        'cloudfront': {
            'BACKEND': 'wagtail.contrib.wagtailfrontendcache.backends.CloudfrontBackend',
            'DISTRIBUTION_IDS': {
                'www.mysite.com': 'distribution-id',
                'www.othermysite.com': 'other-distribution-id',
            }
        },
    }
```

robmoorman:
Thanks for your review and proper testing.
I like the idea of distribution id's per site, however it probably clashes with the ability where authorized users can change the site domain at runtime. Ideally you would have an extra input for it near the domain name within the site settings in Wagtail. However, this required an extendable settings model when you use CloudFront caching... probably also not the best solution.
In that way I see your suggestions is the best fit. Like to hear what your opinion is about this and maybe you have another suggestions to make this distribution id content manageable.
In the meantime I will fix the path component.

robmoorman:
Fixed the mapping of hostnames and the path element.
Added some tests to make sure the right distributionId is called with a path (instead of full url).
Also added a note to make sure users should include www in their mapping.

tomdyson:
Thanks, @robmoorman! @kaedroho, can you look at this again please?

tox.ini
```diff
@@ -19,6 +19,7 @@ basepython =
     py35: python3.5
 
 deps =
+	boto3==1.1.4
```
kaedroho: Should be 4 spaces here instead of tab

wagtail/contrib/wagtailfrontendcache/backends.py
```diff
@@ -83,3 +87,50 @@ def purge(self, url):
...
+            raise ImproperlyConfigured(
+                "The setting 'WAGTAILFRONTENDCACHE' requires the object "
+                "'DISTRIBUTION_ID'."
+            )
+
+    def purge(self, url):
+        url_parsed = urlparse(url)
+        distribution_id = None
+
+        if type(self.cloudfront_distribution_id) is dict:
```
kaedroho: This should be ``if isinstance(self.cloudfront_distribution_id, dict)``

wagtail/contrib/wagtailfrontendcache/backends.py
```diff
@@ -83,3 +87,50 @@ def purge(self, url):
...
+            host = url_parsed.hostname
+            if host in self.cloudfront_distribution_id:
+                distribution_id = self.cloudfront_distribution_id.get(host)
+            else:
+                raise ImproperlyConfigured("hostname %s" % host)
```
kaedroho: Looks like this line was accidentally committed

wagtail/contrib/wagtailfrontendcache/tests.py
```diff
@@ -43,6 +46,41 @@ def test_cloudflare(self):
...
+                'BACKEND': 'wagtail.contrib.wagtailfrontendcache.backends.CloudfrontBackend',
+                'DISTRIBUTION_ID': {
+                    'www.wagtail.io': 'frontend',
+                }
+            },
+        })
+        backends.get('cloudfront').purge('http://www.wagtail.io/home/events/christmas/')
+
+        _create_invalidation.assert_any_call('frontend', '/home/events/christmas/')
+
```
kaedroho: Could we have a test case for when a cache invalidation is performed on a domain that isn't in the ``DISTRIBUTION_ID`` mapping?

robmoorman:
Thanks @kaedroho , I'll pick up your feedback on this PR this week!

kaedroho:
Reviewed 2 of 7 files at r1, 1 of 4 files at r4.
Review status: 3 of 7 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/torchbox/wagtail/1845)*

robmoorman:
he @kaedroho seems like some checks are failing caused by other parts of code from the master branch

gasman:
@robmoorman Have fixed the isort errors on master now, so a rebase should hopefully fix it here too.

kaedroho:
Reviewed 1 of 7 files at r1, 1 of 4 files at r4, 1 of 2 files at r5.
Review status: 6 of 7 files reviewed at latest revision, 1 unresolved discussion.

---

*[wagtail/contrib/wagtailfrontendcache/backends.py, line 113 \[r6\]](https://reviewable.io:443/reviews/torchbox/wagtail/1845#-KIgVIanihuLS3rEK4Ps:-KIgVIaoxb8mseV1QJWN:-882815873) ([raw file](https://github.com/torchbox/wagtail/blob/db2b9191b8ca10039acd1c75661c0aac54108937/wagtail/contrib/wagtailfrontendcache/backends.py#L113)):*
> ```Python
>                 distribution_id = self.cloudfront_distribution_id.get(host)
>             else:
>                 logger.error("Couldn't purge '%s', hostname '%s' not found in mapping", url, host)
> ```

We should point out in this message that it's cloudfront that's failing and that it's a problem with DISTRIBUTION_ID

How about: Couldn't purge '%s' from CloudFront. Hostname '%s' not found in ``DISTRIBUTION_ID`` mapping

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/torchbox/wagtail/1845)*

wagtail/contrib/wagtailfrontendcache/backends.py
```diff
@@ -84,3 +88,49 @@ def purge(self, url):
...
+    def purge(self, url):
+        url_parsed = urlparse(url)
+        distribution_id = None
+
+        if isinstance(self.cloudfront_distribution_id, dict):
+            host = url_parsed.hostname
+            if host in self.cloudfront_distribution_id:
+                distribution_id = self.cloudfront_distribution_id.get(host)
+            else:
+                logger.error("Couldn't purge '%s', hostname '%s' not found in mapping", url, host)
```
kaedroho: We should point out in this message that it's cloudfront that's failing and that it's a problem with DISTRIBUTION_ID

How about: "Couldn't purge '%s' from CloudFront. Hostname '%s' not found in DISTRIBUTION_ID mapping"?

robmoorman: Good point, I'll keep it consistent with the other error messages (like CloudFlare)

wagtail/contrib/wagtailfrontendcache/backends.py
```diff
@@ -84,3 +88,52 @@ def purge(self, url):
...
+    def purge(self, url):
+        url_parsed = urlparse(url)
+        distribution_id = None
+
+        if isinstance(self.cloudfront_distribution_id, dict):
+            host = url_parsed.hostname
+            if host in self.cloudfront_distribution_id:
+                distribution_id = self.cloudfront_distribution_id.get(host)
+            else:
+                logger.error(
```
kaedroho: It's possible that someone might want cloudfront on one site but not another, so perhaps this should be info rather than error?

wagtail/contrib/wagtailfrontendcache/backends.py
```diff
@@ -84,3 +88,52 @@ def purge(self, url):
...
+                    "Couldn't purge '%s' from CloudFront, Hostname '%s' not found in the DISTRIBUTION_ID mapping",
```
kaedroho: Minor inconsistency, the comma after CloudFront should be a full stop.

setup.py
```diff
@@ -21,6 +21,7 @@
 
 
 install_requires = [
+    "boto3>=1.1,<1.2",
```
kaedroho: Sorry for not pointing this out earlier, I think we should avoid adding ``boto3`` as a hard dependency as I expect most people won't be using cloudfront.

kaedroho:
Tested and works well!

Spotted a few more things, but I think this will be ready to merge once they are resolved.

Thanks!

kaedroho:
Made some tweaks here: https://github.com/torchbox/wagtail/compare/torchbox:master...kaedroho:feature/cloudfront-tweaks

@robmoorman could you review?

PR #1847, opened 2015-10-20
Tag level jpeg compression

tacitus:
Adds tag level compression override functionality referenced in https://github.com/torchbox/wagtail/issues/908


gasman:
Hi @tacitus - thanks for this!

@shredding has just submitted PR #1936 which allows chaining a sequence of filters together. Obviously it would be great to have both features, and presumably it's not going to be possible to have them working side-by-side as they currently stand.

Do either of you (and/or @kaedroho) think it would be possible to implement 'compression level' as a new kind of filter? That way, the template tag would just be dealing with a list of filters, rather than treating the compression setting as a special case.

shredding:
With chaining in place, we may get rid of `WAGTAILIMAGES_JPEG_QUALITY` and change to something more generic like

```python
WAGTAILIMAGES_DEFAULT_IMAGE_OPERATIONS = [
    DoNothingOperation(),
    Compress(85)
]
```

Or even:

```python
WAGTAILIMAGES_DEFAULT_PRE_IMAGE_OPERATIONS = [FooOperation()]
WAGTAILIMAGES_DEFAULT_POST_IMAGE_OPERATIONS = [Compress(95)]
```

Which would translate into a `run`-method within filters like:

```python
def run(self, image, output):
        with image.get_willow_image() as willow:
            for operations in getattr(settings, 'WAGTAILIMAGES_DEFAULT_PRE_IMAGE_OPERATIONS', []):
                operation.run(willow, image)
            for operation in self.operations:
                operation.run(willow, image)
            for operations in getattr(settings, 'WAGTAILIMAGES_DEFAULT_POST_IMAGE_OPERATIONS', []):
                operation.run(willow, image)
            ... more code follows here ...
```

What do you think?

kaedroho:
@shredding Interesting idea!

I think your idea makes sense for things like compression and output format. But I do think that any operation that alters the shape or look of an image should be defined on the tag itself.

I expect there would only be a need for a few operations for the former and I believe a global setting would suffice for them. Please let us know if you have any other ideas for operations that would be useful to enable globally though.


As an idea for how to allow operations to change things like compression, perhaps we could add a method that returns a dictionary of "properties":

```python
# eg jpeg-quality-50
class JPEGQualityOperation(Operation):
    def construct(self, quality):
        self.quality = int(quality)

    def run(self, willow, image):
        pass

    def get_properties(self, props):
        props['jpeg_quality'] = self.quality
        return props


# eg output-format-jpeg
class OutputFormatOperation(Operation):
    def construct(self, format):
        self.format = format

    def run(self, willow, image):
        pass

    def get_properties(self, props):
        props['output_format'] = self.format
        return props
```

When each operation is executed, the ``get_properties`` method is called updating a dictionary of properties. When all operations are run, the jpeg quality and output format would be found there.

shredding:
A compression image operation may look like this, but I have to take a closer look into the test setup.

In order to do it correcty, we would need to create an intermediate output like this:

```python
class CompressOperation(Operation):

    def construct(self, quality):
        self.quality = int(quality)

    def run(self, willow, image):
        if willow.original_format == 'jpeg':
            temp = BytesIO()
            willow.save_as_jpeg(temp, quality=self.quality)
            willow.image = willow.from_file(temp)
```

shredding:
> Please let us know if you have any other ideas for operations that would be useful to enable globally though.

Maybe you are right, I can't imagine any.

tacitus:
@gasman , @shredding, @kaedroho  Some nice extensibility  ideas in here; I'll have a bash when I get a moment; unless someone else gets to it first of course.

ddiazpinto:
That's very interesting. How this goes?

Anyway, I think that the first element that should be updated is [Willow](https://github.com/torchbox/Willow/) to allow all the parameters of Pillow. For example, the `save()` method for `jpeg` format, has a lot of interesting options like `optimized`, `progressive`, `dpi`, `icc-profile` and `exif`. Take a look at [Pillow:Image File Formats](http://pillow.readthedocs.org/en/latest/handbook/image-file-formats.html)

About post processing, there is a lot of interesting process to accomplish with images; clear or reset exif and metadata information, adds watermarks, change formats, colorspace, background, etc. The best way to allow that is to let the user decides and implements his own functions. I mean, instead of @shredding said before, something like:

```python
WAGTAILIMAGES_PROCESSOR = [
    'wagtailimages.processors.Optimize',
    'my.custom.module.processors.ResetMetadata',
    'my.custom.module.processors.AddWatermark'
]
```

gasman:
Superseded by #2474

PR #1848, opened 2015-10-21
Update gulp-sass to 2.0.x to support Node.js v4

vden:
This small change allows assets to be built with latest Node.js (LTS version 4.2).

Here is changelog for gulp-sass: https://github.com/dlmanning/gulp-sass/blob/v2.0.4/CHANGELOG.md

So by using new gulp-sass we could drop ancient node-sass 0.9 and use its live 3.3.x branch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "gulp": "~3.8.11",
     "gulp-autoprefixer": "~1.0.1",
     "gulp-rename": "^1.2.2",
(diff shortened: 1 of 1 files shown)


davecranwell:
:+1:

gasman:
Looks good to me based on a quick `npm install` / `npm run build` - thanks!

Issue #1670, opened 2015-09-09
UnicodeEncodeError when generating gravatar for user with an email in an internationalized domain name


LiquidPL:
Hello,
When I'm trying to add an user with an email using an IDN, the following UnicodeEncodeError is raised:
http://dpaste.com/2G6KVBN

kaedroho:
Thanks for spotting

Looks like ``b(email.lower())`` to ``email.lower().encode('utf-8')`` should fix this..

gasman:
@kaedroho That's assuming that Gravatar is also using utf-8 when calculating hashes - annoyingly https://en.gravatar.com/site/implement/hash/ doesn't specify. However, in the worst case, that just means it'll fail to find the user's avatar on Gravatar, which is far preferable to the admin breaking completely :-)

davecranwell:
Would hope we find some time to make #319 a reality soon, reducing dependence on gravatar.

gasman:
Fixed in 13c8e4da583b1178974c48d822fbaa411de1f755, with test added in d1ef42fe9b7d0f91d75b4ba0133a75d0066b6683

PR #1850, opened 2015-10-21
Force utf-8 encoding of user email in gravatar request

vden:
While we're waiting for uploadable avatars we can at least fix the issue with IDN emails in #1670 
Solved just as @kaedroho suggested, tested on Python2.7 and 3.x as well.

### DIFF
--- a/wagtail/wagtailadmin/templatetags/gravatar.py
+++ b/wagtail/wagtailadmin/templatetags/gravatar.py
@@ -29,8 +29,10 @@ class GravatarUrlNode(template.Node):
         default = "blank"
         size = int(self.size) * 2  # requested at retina size by default and scaled down at point of use with css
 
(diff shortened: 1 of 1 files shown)


prydie:
Rather than string concatenation I'd prefer the use of `string.format` like the following:

```
gravatar_url = "//www.gravatar.com/avatar/{hash}?{params}".format(
    hash=hashlib.md5(email.lower().encode('utf-8')).hexdigest(),
    params=urlencode({'s': size, 'd': default})
)
```

Also note the conversion from `int` to `str` isn't necessary as far as I can see as `urlencode` handles that for you.

gasman:
Thanks! Merged in https://github.com/torchbox/wagtail/commit/13c8e4da583b1178974c48d822fbaa411de1f755.

PR #1854, opened 2015-10-22
Made sure email address is encoded as UTF-8 before hashing

kylestratis:
Made fixes specified by @kaedroho for issue #1670


gasman:
Thanks! @vden got there very slightly earlier in #1850, but I've credited you both :-)

PR #1856, opened 2015-10-22
Fix search on specific page class using the page queryset.

### LINKED ISSUE #1721: Searching a queryset of a specific page class with a child_of/descendant_of filter gives error
Searching on a specific page class with the ``child_of``/``descendant_of`` filter applied, gives the following error:

```
FieldError at /videos/

Cannot filter search results with field "page_ptr_id". Please add index.FilterField('page_ptr_id') to VideoPage.search_fields.
```

Here's an example query that causes this:
```python
VideoPage.objects.child_of(parent_page).search("foo")
```

It seems to be a combination of a specific queryset and ``child_of``/``descendant_of`` filter that causes the issue as the following query (which uses the generic page model) works.

```python
Page.objects.child_of(parent_page).type(VideoPage).search("foo")
```

laidibug:
Fixes #1721 

This search bug will appear on any specific page type classes that use the page queryset,
that includes page, child_of, descendant_of, ascendant_of, and sibling_of as well all the negative
versions.

Searching on a specific page type and using the page queryset
results in a missing FilterField message for "page_ptr_id".
The where node query is used to build the list of filters
and xxxx_ptr_id shows up when we use any inherited model and query
by the id. It's best to just ignore this as its useless to
filter search on a pointer field.

We could define it as a FilterField on the Page model, but this
way it prevents other indexed Django models from hitting the same problem.

### DIFF
--- a/wagtail/wagtailsearch/backends/base.py
+++ b/wagtail/wagtailsearch/backends/base.py
@@ -71,6 +71,10 @@ class BaseSearchQuery(object):
             lookup = where_node.lookup_name
             value = where_node.rhs
 
+            # Ignore pointer fields that show up in specific page type queries
+            if field_attname.endswith('_ptr_id'):
+                return
+
             # Process the filter
             return self._process_filter(field_attname, lookup, value)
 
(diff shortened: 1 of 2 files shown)


kaedroho:
Thanks!

tomdyson:
Thanks @laidibug!

PR #1861, opened 2015-10-23
catch errors on non existing images

shredding:
If an image does not exist (or the cdn is idle) the entire page rendering breaks. It should fallback to not displaying the image.

### DIFF
--- a/wagtail/wagtailimages/blocks.py
+++ b/wagtail/wagtailimages/blocks.py
@@ -1,7 +1,6 @@
 from django.utils.functional import cached_property
 
 from wagtail.wagtailcore.blocks import ChooserBlock
-from wagtail.wagtailimages.models import SourceImageIOError
 
 
 class ImageChooserBlock(ChooserBlock):
@@ -17,9 +16,6 @@ class ImageChooserBlock(ChooserBlock):
 
     def render_basic(self, value):
         if value:
-            try:
-                return value.get_rendition('original').img_tag()
-            except SourceImageIOError:
-                return ''
+            return value.get_rendition('original').img_tag()
         else:
             return ''
--- a/wagtail/wagtailimages/blocks.py
+++ b/wagtail/wagtailimages/blocks.py
@@ -1,6 +1,8 @@
 from django.utils.functional import cached_property
 
 from wagtail.wagtailcore.blocks import ChooserBlock
+from wagtail.wagtailimages.models import SourceImageIOError
+
 
 class ImageChooserBlock(ChooserBlock):
     @cached_property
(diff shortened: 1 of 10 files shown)


gasman:
Thanks! I think the appropriate result here would be to render a broken image (as we do here: https://github.com/torchbox/wagtail/blob/222b548d9e57ebc14a0a73262dd2a135a5837615/wagtail/wagtailimages/formats.py#L31-L37) rather than rendering nothing - that way it's clear that there's a file error, and not just Wagtail ignoring a block for an unknown reason.

Also, please could you add a test for this?

shredding:
@gasman I see your point. However, the entire approach is not really DRY, so why not following the fat-models-thin-views-approach?

I moved the fallback to the get_rendition() method of the image and was therefore able to remove lot's of code replication. 

I see that this breaks the interface. So there are two options:

1. announce a potentially breaking change and do it like that - I'd vote for that, because there aren't many usecases where you want that error to escalate to the views. If someone requires the exception, she needs to change fallback_to_not_found to False

2. Make it a backwards compatible change by defaulting fallback_to_not_found to False and change it to true in all the use cases where I removed the duplicate code (+ the one place for which I opened this ticket). That's a bit more error prone.

gasman:
Moving exception handling into models is a misuse of the fat-models-thin-views principle, I think. If a model method isn't able to do the operation that the view requested, it should raise an exception back to the view, so that the view can decide what to do (report it back to the user, ignore it, or whatever). We actually got this wrong recently in some very similar code within the wagtailembeds module, and almost broke the embed chooser in 1.2: #1888 :-)

You're absolutely right to highlight the `rendition.file.name = 'not-found'` code duplication, though. Perhaps we should add a `get_not_found_rendition` class method to AbstractRendition? That way, the try/except would still be part of the views, but the 'except' clause would become a one-line call to that method, rather than a duplicated code block.

shredding:
how about a `get_rendition_or_not_found` shortcut, similar to https://docs.djangoproject.com/en/1.8/topics/http/shortcuts/#get-object-or-404?

gasman:
Sure, I'd be happy with that!

shredding:
Okay, I will refactor that and widen the scope of this ticket: An editor in my company just uploaded a tiff (prepress format) which crashes the entire backend atm.

```
raise cls.LoaderError("Cannot find backend that can load this image file") 
```

... origined in `image.get_willow_image`.

Our shortcut function should be like "give me the image and if you can't please do not break anything" :)

gasman:
Ideally we should be catching invalid image formats at the point of upload, but I guess that's #1792. No harm in trapping the error here too, for any that do slip through the net.

shredding:
@gasman : Here you go. I kept the scope of the ticket, because I decided to tweak our custom  batch image import to verify data consistency when data comes in.

gasman:
Merged in 2dd3f42e257981f2052613331bbc9e6dc497bde0 - thanks!

PR #1411, opened 2015-06-16
Use six from Django

timheap:
No point in installing six as a standalone dependency if it is already bundled with Django.

Fixes #1403
See #1406

PR #1868, opened 2015-10-27
Use six from Django, remove six from tox.ini

timheap:
All referencs to `six` were removed in #1411 favouring `django.utils.six`, except for one in `tox.ini` where it was installed as a dependency. Because it was still installed in the test environment, a few references to `six` crept back in. These will cause an ImportError when users do not have `six` installed.

These references have been fixed, and `six` has been removed from `tox.ini`


kaedroho:
Thanks!

Issue #1394, opened 2015-06-10
Previewing pages causes undesirable Google Analytics hits 


domeheid:
As someone who has implemented Google Analytics in my site, I want to ensure that admin pages do not appear in my Google Analytics reports e.g.

![screen shot 2015-06-10 at 16 45 55](https://cloud.githubusercontent.com/assets/478791/8087022/78d69a12-0f90-11e5-9a33-5009f99692ab.png)

Could you please update Wagtail to tell the page it was being previewed, and use that knowledge to exclude the Google Analytics or Google Tag Manager snippets as appropriate?

At the moment, I'm having to fix this issue by either a) adding an exception trigger / blocking rule to exclude page URLs containing /admin/ in the main Google Analytics pageview tag in Google Tag Manager or b) adding a view filter in Google Analytics to exclude subdirectories that contain /admin/.

davecranwell:
GTM/GA tags are nothing to do with wagtail, of course, but the key here is just to ensure that page are aware when they're being previewed or not, so that content within can be altered.

gasman:
Fixed in #1870.

PR #1870, opened 2015-10-27
Pass flag marking preview mode into template

vden:
This is an attempt to distinguish preview mode from live page to solve the issue mentioned in #1394 

One can use `request.is_preview` in site templates to see whether the page is in preview and switch off some code on this, like GA tracking etc.


gasman:
Merged in ab45915a90949d0ad616615fc609cd50e02f450f - thanks!

PR #1612, opened 2015-08-14
Add help text to steer editors away from creating non-routable pages

gasman:
Adding help text to the explorer at the root level (and any other place where no site is configured), as proposed in https://github.com/torchbox/wagtail/issues/1583#issuecomment-131078851. @davecranwell might want to tweak the styling here (it gets an extra left margin for being the first cell of a table row, which we probably don't want).

Incorporates the display fix from #1611.

Issue #1807, opened 2015-10-09
Admin subdomains


diveru4i:
Greetings.

Is there a way to properly serve wagtail admin site on a different subdomain (eg. *admin.example.com*)?  

I've patched `django-subdomains` middleware so it would work with `wagtail.wagtailcore.models.Site`, but now all reverse urls in admin point to *admin.example.com/--uri--/*. To fix this I need to patch `wagtail.wagtailcore.models.Page.url`, but I just don't want to monkey-patch source packages.

Is there any way around that? And maybe in the future you could implement the `settings.DEFAULT_PAGE_MODEL` setting for wagtail, so we could override some default behaviour?

Thanks.

gasman:
That's an interesting one! We intentionally don't include domains in the result of `Page.url` when there's only one Site record configured, so that links work immediately on a new Wagtail project without having to edit the site configuration. But that doesn't help if you're using multiple domains for hosting things other than Wagtail site entry points (such as the Wagtail admin)...

As a workaround, you could create a dummy site record (create a new page at the root level, and add a Site record in Settings -> Sites pointing to that page, using any hostname you like) - this will cause Page.url to always generate full URLs.

I guess the proper fix would be to add an "always generate full URLs" flag to settings - I can't see any more elegant way of handling this.

Re `settings.DEFAULT_PAGE_MODEL`, that's covered by issue #836. (No progress on it yet, though...)

thibaudcolas:
Reviewed today – related: #12462. This would still be a nice improvement. Also relates to generating "live" URLs for headless websites. Opportunity for us to revisit routing logic when there is a single site record. If anyone wants to work on this please ask @gasman for guidance, otherwise this issue will stay put.

PR #1883, opened 2015-10-30
Link to homepage from dashboard page link if single site

timheap:
If the Wagtail install has only one site, link to the homepage of that site from the dashboard page summary instead of the root page of the hierarchy.

Hopefully this will prevent some of the confusion causing people to create pages under the root page, where they are inaccessible. See #1612 for more information around this issue.

Wagtail installs with multiple (or zero) sites retain the old behaviour of linking to the root page.

### DIFF
--- a/wagtail/wagtailadmin/site_summary.py
+++ b/wagtail/wagtailadmin/site_summary.py
@@ -1,5 +1,5 @@
 from wagtail.wagtailcore import hooks
-from wagtail.wagtailcore.models import Page
+from wagtail.wagtailcore.models import Page, Site
 from wagtail.utils.compat import render_to_string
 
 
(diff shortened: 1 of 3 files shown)


wagtail/wagtailadmin/site_summary.py
```diff
@@ -21,7 +21,13 @@ class PagesSummaryItem(SummaryItem):
     template = 'wagtailadmin/home/site_summary_pages.html'
 
     def get_context(self):
+        # If there is a single site, link to the homepage of that site
+        # Otherwise, if there are multiple sites, link to the root page
+        single_site = Site.objects.count() == 1
+        root = self.request.site.root_page if single_site else None
```
gasman: We shouldn't ever rely on request.site being well-defined within the admin - see #1807. Not a problem here, of course, since we can just do `Site.objects.get()` instead...

gasman:
Merged (with suggested fix to avoid `request.site`) in 44564ea76aef23f4c9719f54bee8fc22559c70bb.

wagtail/wagtailadmin/site_summary.py
timheap: Oh, I was not aware of that. Good catch, and I will keep this in mind in the future.

PR #1381, opened 2015-06-05
Catch EmbedNotFoundException fixes #1324

kaedroho:

PR #1888, opened 2015-10-30
Ensure errors in embed chooser are propagated back to the form

### LINKED ISSUE #1885: Errors from embed chooser are not reported back to the user
It seems that embed functionality of wagtail hello.js editor is not working. I tried to embed a link, modal comes up, everything is fine but when I press "Insert" I get nothing and browser displays this error:

(Firefox) TypeError: a is undefined rangy-core.js:33:479
(Chrome) rangy-core.js:33 Uncaught TypeError: Cannot read property 'nodeType' of undefined

But network data says everything is OK:
POST .../admin/embeds/chooser/upload/ 200 OK
After closing the modal window, embed link is not present.

gasman:
Fixes #1885.

In #1381 the behaviour of `wagtail.wagtailembeds.formats.embed_to_editor_html` was changed to ignore any errors in embed lookup, to prevent these from crashing the page entirely. However, this function is also used when the embed is first inserted; catching these errors meant that the error messages were never communicated back to the user (and instead hallo.js ends up throwing a JS error while trying to insert an empty embed element into the rich text area).

### DIFF
--- a/wagtail/wagtailembeds/rich_text.py
+++ b/wagtail/wagtailembeds/rich_text.py
@@ -1,4 +1,4 @@
-from wagtail.wagtailembeds import format
+from wagtail.wagtailembeds import format, embeds
 
 
 class MediaEmbedHandler(object):
@@ -26,6 +26,10 @@ class MediaEmbedHandler(object):
         representation.
         """
         if for_editor:
-            return format.embed_to_editor_html(attrs['url'])
+            try:
+                return format.embed_to_editor_html(attrs['url'])
+            except embeds.EmbedException:
+                # Could be replaced with a nice error message
+                return ''
         else:
             return format.embed_to_frontend_html(attrs['url'])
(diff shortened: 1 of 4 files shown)


wagtail/wagtailembeds/rich_text.py
```diff
@@ -26,6 +26,6 @@ def expand_db_attributes(attrs, for_editor):
         representation.
         """
         if for_editor:
-            return format.embed_to_editor_html(attrs['url'])
+            return format.embed_to_editor_html(attrs['url'], ignore_errors=True)
```
kaedroho: Wouldn't it be a bit nicer to let ``embed_to_editor_html`` raise its error and catch it here?

gasman: Fair point! Will update.

gasman:
Updated as per https://github.com/torchbox/wagtail/pull/1888#discussion_r43620869 and rebased on latest master.

PR #1889, opened 2015-10-31
Documentation for BoundBlock

gasman:
Having to answer people's StreamField questions with "yes, you can give this block its own HTML rendering (except for *mumble mumble obscure special case where it won't work*)" was bothering me, so I buckled down and documented the gotchas properly.

In the process, I realised that the explanation was far simpler if we set up the string representation of BoundBlock to always use the block rendering, rather than that being a special property of the StreamChild subclass, so I took the liberty of changing that behaviour.

### DIFF
--- a/docs/topics/streamfield.rst
+++ b/docs/topics/streamfield.rst
@@ -435,11 +435,9 @@ Within the template, the block value is accessible as the variable ``value``:
     <div class="person">
         {% image value.photo width-400 %}
         <h2>{{ value.first_name }} {{ value.surname }}</h2>
-        {{ value.bound_blocks.biography.render }}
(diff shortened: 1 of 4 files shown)


gasman:
Would you be OK to review this one please, @timheap? I figure this is your area :-)

timheap:
The docs look good to me. It explains the situation nicely, only going in to explain the internals when it is needed.

It is a bit unfortunate that this needs documentation, it would be nice if it all Just Worked, but I cant see a way of making that happen. This is a good solution.

gasman:
Cheers @timheap! Merged in 8a17aeb5a8ab18cecf74087b89f04e9bc087214b + parents.

PR #1892, opened 2015-11-02
Upload multiple documents at once

timheap:
Functionality from the multiple image uploader was used. It was all rather easy to do because most of the work had already been done!

The jQuery plugins from the image uploader have been copied across, which means they are in the repo twice. They are in different apps though, which do not have to be installed at the same time, so they can not simply use the wagtailimages version of the scripts, sadly. The scripts could be moved in to admin, which is always installed if duplicating the scripts is an issue?


davecranwell:
Brilliant! Thanks @timheap. You may also like to check my work on https://github.com/torchbox/wagtail/pull/1234 which I believe slightly refactored some of the upload code, removes the concept of having a separate "upload screen", and the mandatory field requirements. 

Regarding script location I think the uploader scripts should be seen as a reusable asset, just like the icons or the other javascript plugin systems e.g choosers. It seems there are quite a few users trying to do file/image upload systems as part of their own private apps and if they could reuse these upload scripts it would improve the aesthetic all round.

SalahAdDin:
:+1:

wagtail/wagtaildocs/forms.py
```diff
@@ -12,3 +12,13 @@ def get_document_form(model):
             'tags': widgets.AdminTagWidget,
             'file': forms.FileInput()
         })
+
+
+def get_document_multi_form(model):
+    return modelform_factory(
+        model,
+        fields=['title', 'tags'],
```
gasman: Now that we support custom document models, would it be worth making this configurable - either by re-using `model.admin_form_fields` minus `file`, or having a new attribute on AbstractDocument alongside `admin_form_fields`, just for this form?

(Also, setting a widget for `file` here is redundant)

wagtail/wagtaildocs/static_src/wagtaildocs/scss/add-multiple.scss
```diff
@@ -0,0 +1,91 @@
...
+        opacity: 0;
+        position: absolute;
+        top: 0;
+        right: 0;
+        direction: ltr;
+        width: auto;
+        display: block;
+        font-size: 5em;
+
+        &: hover {
```
gasman: The whitespace after `&:` (here and immediately below) should be removed - it breaks on my copy of node-sass.

wagtail/wagtaildocs/static_src/wagtaildocs/scss/add-multiple.scss
```diff
@@ -0,0 +1,91 @@
...
+        width: 150px;
+        display: block;
+        position: relative;
+        text-align: center;
+        max-width: 100%;
+        margin: auto;
+    }
+
+    .progress {
+        @include box-shadow(0 0 5px 2px rgba(255, 255, 255, 0.4));
```
gasman: This should be `box-shadow: 0 0 5px 2px rgba(255, 255, 255, 0.4);` - the mixin was dropped in #1964.

gasman:
This is working very nicely - thanks @timheap! I'd definitely prefer the js/vendor files to be moved to wagtailadmin rather than duplicated between wagtailimages and wagtaildocs, as per @davecranwell's reasoning.

Ideally, it would be nice to see the copy-and-paste trimmed down further - a general-purpose multiple-uploader CBV would make a nice addition to wagtailadmin.views.generic, I reckon - but I wouldn't make that a blocker for merging this.

timheap:
I've moved the duplicated JS files to wagtailadmin, and fixed the SASS errors.

gasman:
Rebased and merged in e131a9bcaf3f06b06f65481e936fd08564fbe056 + parents, including updating the views to use permission policies (fcb3adb2f0073a49a5aa6ee67089cf04a31137f0). Thanks!

PR #1894, opened 2015-11-02
Moved ratio calculation to Embed model

kaedroho:
This allows the calculation to be used anywhere where we have an embed.

For example:

``` html+django
<div style="padding-bottom: {{ embed.ratio_css|default:0 }};" class="responsive-object">
```


gasman:
Looks good to me - just needs tests. (Could really do with some documentation as well, but it seems that there's no documentation at all for using wagtailembeds outside of rich text, so that's out of scope for this ticket...)

kaedroho:
This should be ready to merge now

PR #1897, opened 2015-11-03
Allow subclasses of CreateView/EditView to omit template_name

gasman:
A pre-requisite for collections (#1751) - decoupling from that PR to be reviewed independently.

As of #1668, the create.html and edit.html templates in wagtailadmin/generic provide a usable form rendering in their own right, so apps based on generic views can now use those templates directly rather than inheriting from them. This PR sets a default `template_name` on the generic views so that subclasses can omit this parameter in this case.


kaedroho:
Merged in 06f793a3556f7de51af77d2a814fa0860ab7f908

PR #1898, opened 2015-11-03
Add user_passes_test auth decorator

gasman:
A pre-requisite for collections (#1751) - decoupling from that PR to be reviewed independently.

Refactor the `permission_required` and `any_permission_required` decorators to expose a lower-level `user_passes_test` decorator that can be passed an arbitrary test function on the user object. (This mirrors the same-named decorator in django.contrib.auth, but provides a wagtailadmin-specific 'permission denied' response and avoids using LOGIN_URL.)

In particular, this means that we can restrict access based on PagePermission or CollectionPermission rules, which are distinct from the basic Permission model.

### DIFF
--- a/wagtail/tests/testapp/urls.py
+++ b/wagtail/tests/testapp/urls.py
@@ -1,0 +1,6 @@
+from django.conf.urls import url
+from wagtail.tests.testapp.views import bob_only_zone
+
+urlpatterns = [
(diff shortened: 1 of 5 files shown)


kaedroho:
Merged in eccb0990bb98d7a69441d450f937ac67882a5c92 + parent

PR #1075, opened 2015-03-19
Long-running form submit spinners

davecranwell:
First pass at addressing https://github.com/torchbox/wagtail/issues/305

910e1569e1eb28b54d385e3077b9dae9

75265d3b273d440d0d1ef22ddcdaaecc

Still to apply this to more forms

PR #1900, opened 2015-11-04
Dropdowns now don't close if the click detected originated within them

davecranwell:
This reinstates code I removed during the initial merger of #1075  (https://github.com/torchbox/wagtail/pull/1075#discussion-diff-26763533) because I wanted to find a better solution.

Currently although the page "publish" button supports a long-running spinner, it's never visible because without the code in this PR, the dropdown in which it lives closes as the button is clicked.


kaedroho:
Works nicely in Chrome. Firefox (tested with 43) still closes the dropdowns though.

Merged in ce0017f930ba6321eeb8c4901e8af510e15825f4 + parent

PR #1901, opened 2015-11-04
Start 1.3 release cycle

kaedroho:
- Set up new changelog/release notes
- Removed backwards compatibility for is_abstract (deprecated in 1.1)
- Updated deprecation warning classes


docs/releases/1.3.rst
```diff
@@ -0,0 +1,23 @@
+==========================================
+Wagtail 1.3 release notes - IN DEVELOPMENT
```
timorieber: IN DEVELOPMENT needs to be removed from 1.2.rst then

gasman: @timorieber This will be removed when Wagtail 1.2 final is released. For now there's still a possibility that the 1.2 release will receive bug reports and subsequent fixes.

timorieber: Ok, thanks for the explanation. I thought this PR will at the earliest be merged to master when 1.2 is final.

PR #1905, opened 2015-11-06
lint scss based on pre-existent configuration file

ratmcarlo:
I am trying to push forward the linting.
I have modified the scss files to pass the linter (run ```scss-lint``` from the root of the project)
I based my work on the pre-existent configuration file.

I fixed the trivial things (for example BorderZero ecc...)

For non-trivial things, I disabled the linter in place. I think the best approach would be to refactor them separately, not in one huge commit like this.
There are **lots** of disabled lines, the full list is here: http://dpaste.com/17WZ81J

I hope this could be just a **starting point** for future work.

PS: the documentation does not reflect the scss-lint configuration file, and the command ```scss-lint .``` is just wrong: http://docs.wagtail.io/en/latest/contributing/css_guidelines.html

PPS: 
this has been done for the 80% by a script, I am not that mad

### DIFF
--- a/.scss-lint.yml
+++ b/.scss-lint.yml
@@ -1,6 +1,6 @@
 
-scss_files: 'wagtail/*/static/**/*.scss'
-exclude: 'wagtail/**/static/**/vendor/*.scss'
+scss_files: 'wagtail/*/static_src/**/*.scss'
+exclude: 'wagtail/**/static_src/**/vendor/*.scss'
 
 linters:
     BorderZero:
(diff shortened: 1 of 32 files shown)


davecranwell:
Hi @ratmcarlo I'm afraid you're a day or two behind me :)  https://github.com/torchbox/wagtail/pull/1737

ratmcarlo:
@davecranwell Uh nice! Did not see that, sorry.

I will look into that, then. Thank you

PR #1906, opened 2015-11-06
Handle reverse_lazy generated urls

adonm:
To allow for lazy url generation in hook registration, see https://github.com/unbit/django-uwsgi/blob/master/django_uwsgi/wagtail_hooks.py#L41 for an example


kaedroho:
Thanks @adonm 

I think we should add a test for this so a future refactor doesn't accidentally reverse it

gasman:
Test added in #1998

kaedroho:
Merged #1998 but credited you both for this in the release notes. Thanks!

PR #1913, opened 2015-11-08
Image serve view improvements

### LINKED ISSUE #1066: Missing image results in 500 error
Morning all

We're noticing some 500 error emails getting sent out when an image is missing, not on the file system, for whatever reason.

I tracked it down to here

https://github.com/torchbox/wagtail/blob/stable/0.8.x/wagtail/wagtailimages/views/frontend.py#L18

I know how to stop it happening, but I'm not %100 certain on what you'd like to return.

Something like this would probably do it.

```
def serve(request, signature, image_id, filter_spec):
    image = get_object_or_404(get_image_model(), id=image_id)

    if not verify_signature(signature.encode(), image_id, filter_spec):
        raise PermissionDenied

    try:
        rendition = image.get_rendition(filter_spec)
        try:
          rendition.file.open('rb')
        except IOError:
          return HttpResponse("File not present", content_type='text/plain', status=404)
        return HttpResponse(FileWrapper(rendition.file), content_type='image/jpeg')
    except Filter.InvalidFilterSpecError:
        return HttpResponse("Invalid filter spec: " + filter_spec, content_type='text/plain', status=400)```

### PR DESCRIPTION
**Pull request by kaedroho**
This PR makes some improvements to the image serve view
 - Renamed ``frontend.py`` to ``serve.py``
 - Now catches ``SourceImageIOError`` (Fixes #1066)
 - Now accepts an extra path component (optional) allowing a filename to be added to the URL
 - Refactored into a class based view
 - Made image model and key (for signing urls) configurable
 - Added redirect action (when switched on, this will make the view redirect to the image file instead of serving it from Django)
 - Moved generate/verify signature functions into ``views/serve.py``. Unifying all the code supporting the serve view and deprecating ``wagtailimages/utils.py``

### CODE REVIEW

File: wagtail/wagtailimages/views/serve.py [hunk 1]
```diff
@@ -0,0 +1,68 @@
...
+    if key is None:
+        key = settings.SECRET_KEY
+
+    # Key must be a bytes object
+    if isinstance(key, text_type):
+        key = key.encode()
+
+    # Based on libthumbor hmac generation
+    # https://github.com/thumbor/libthumbor/blob/b19dc58cf84787e08c8e397ab322e86268bb4345/libthumbor/crypto.py#L50
+    url = str(image_id) + '/' + str(filter_spec) + '/'
```
prydie: I would suggest using string formatting rather than concatenation.

```
url = '{}/{}/'.format(str(image_id), str(filter_spec))
```

timheap: You wouldnt even need to call `str()` on the arguments, as `.format()` does that for you:

```python
url = '{}/{}/'.format(image_id, filter_spec)
```

prydie: Good point, cleaner still.

kaedroho: Done, thanks!

wagtail/wagtailimages/views/serve.py
```diff
@@ -0,0 +1,68 @@
...
+
+        image = get_object_or_404(self.model, id=image_id)
+
+        # Get/generate the rendition
+        try:
+            rendition = image.get_rendition(filter_spec)
+        except SourceImageIOError:
+            return HttpResponse("Source image file not found", content_type='text/plain', status=410)
+        except InvalidFilterSpecError:
+            return HttpResponse("Invalid filter spec: " + filter_spec, content_type='text/plain', status=400)
```
prydie: `string.format` would be more performant than concatenation here as well.

wagtail/wagtailimages/views/serve.py
```diff
@@ -0,0 +1,68 @@
...
+
+        # Serve it
+        if self.action == 'serve':
+            # Open and serve the file
+            rendition.file.open('rb')
+            image_format = imghdr.what(rendition.file)
+            return StreamingHttpResponse(FileWrapper(rendition.file), content_type='image/' + image_format)
```
prydie: `string.format`

wagtail/wagtailimages/views/serve.py
timheap: This is incorrect. Simple concatenation is much more performant than `.format()`:

```python
>>> timeit.timeit('"foo" + bar', setup="bar = 'bar'")
0.0954943690012442
>>> timeit.timeit('"foo{}".format(bar)', setup="bar = 'bar'")
0.25878583200028515
>>> timeit.timeit('"/{}/{}".format(bar, bar)', setup="bar = 'bar'")
0.2846537450022879
```

`.format()` is easier to read, and is more convenient than concatenation for complex strings, but it is not more performant. Simple concatenation like in the above code is fine as it is in my opinion.

prydie: Interesting. I thought I remembered reading a SO post that used `timeit` with a lot of iterations and averaged it out. I seemed to remember that the modulus operator won but `string.format` came second. I can't seem to find it at the moment but there's some information on https://wiki.python.org/moin/PythonSpeed/PerformanceTips#String_Concatenation.

prydie: The more I look into it the more I think I'm miss-remembering what I read (perhaps it was a comparison between `str.join` and `str += 'foo'` in a loop). String concatenation is fine here, ignore me!

tomdyson:
Happy New Year @timheap! Do you have time to review this?

timheap:
Yep, I should have some time tomorrow to review this and #2062

tomdyson:
Fantastic, thanks very much.

timheap:
There is no documentation on how you would go about using the redirect method over the serve method in the ServeView. How do you intend developers to opt in to the redirect behaviour over the streaming behaviour?

wagtail/wagtailimages/views/serve.py
```diff
@@ -0,0 +1,68 @@
...
+        elif self.action == 'redirect':
+            # Redirect to the file's public location
+            return HttpResponsePermanentRedirect(rendition.url)
+        else:
+            raise ImproperlyConfigured("ServeView action must be either 'serve' or 'redirect'")
```
timheap: As `action` is set when `ServeView.as_view()` is called, invalid actions should be checked there. Raising a configuration error when the server starts will prevent much developer frustration, compared to only finding the error once the website is up and running!

gasman:
The rest looks good to me (just needs the regexp fix https://github.com/kaedroho/wagtail/commit/66994996d13d080178f570533a766f4fd6755a5e#commitcomment-15614517 and the actions pointed out by @timheap). I guess there's no need to mention the deprecations in the release notes, as those functions were internal-only...

kaedroho:
Rebased and docs added

timheap:
Hmmm, I just had an idea!

If the generated thumbnails were saved at a path something like `/media/images/<security-key>/<pk>/<filter-spec>/<filename.jpg>`, then this URL contains enough information to use the ServeView to generate image thumbnails on the fly. With a URL route such as:

```python
url(settings.MEDIA_ROOT + r'([^/]*)/(\d*)/([^/]*)/[^/]*$', ServeView.as_view(action='redirect'), name='wagtailimages_serve')
```,

and using `try_files` with a fallback to Django in nginx such as:

```nginx
location /media/ {
    root /path/to/site/assets/;
    try_files $uri @django;
}
```

then lazily generated images that gracefully handle already generated thumbnails with no redirects are possible! Huzzah

kaedroho:
Thanks @timheap! That's an interesting idea, but I think there's a couple of potential issues with introducing a folder structure like that:
- For large sites, it'll create a ton of folders (using up inodes)
- Putting the signatures into the filenames makes it difficult to change the key or use multiple keys

The main idea behind the redirection is so that for users of S3/cloudfiles, the serve view doesn't have to redownload the rendered images on every request.

For locally hosted images, I think it'll be better to use sendfile (#1942) since this can use the current folder structure but allows the transfer of the file to be offloaded to nginx.

PR #1919, opened 2015-11-11
Do not delete uploaded images or page revisions when a user is deleted

### LINKED ISSUE #1918: When a user is deleted, any images they uploaded are also deleted.
Currently, when a user is deleted, wagtail cascade deletes any images they uploaded in addition to any PageRevisions they made.

The ForeignKey relationships from AbstractImage.uploaded_by and PageRevision.user do not specify an on_delete action, and the so the default CASCADE is used.

I can't think of a scenario where a site owner would actually want to have all of a users images and page revisions deleted when you delete the user, so I believe this should be changed to on_delete.SET_NULL for AbstractImage.uploaded_by and PageRevision.user. 

Both of these relationships are already nullable.

I note that Page.owner is already on_delete=models.SET_NULL.

Pull request below.

atkinson:
fixes #1918

### DIFF
--- a/wagtail/tests/testapp/migrations/0016_uploaded_by_user_on_delete_set_null.py
+++ b/wagtail/tests/testapp/migrations/0016_uploaded_by_user_on_delete_set_null.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+import django.db.models.deletion
+from django.conf import settings
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('tests', '0015_genericsnippetpage'),
+    ]
+
+    operations = [
+        migrations.AlterField(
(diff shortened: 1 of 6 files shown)


timheap:
Nice work! It seems you have not made migrations for these model changes though, which is causing the test failures.

atkinson:
Hmmm.... i don't think it was the migrations.

timheap:
For some reason, the migration `wagtailcore.0021` did not include the `editable=False` parameter for `Page.owner`, and there was no migration created for the test app. You can make migrations using the test settings by running this command from the root of the Wagtail repository:

```bash
$ DJANGO_SETTINGS_MODULE=wagtail.tests.settings django-admin makemigrations
```

atkinson:
Can this be merged?

gasman:
Thanks! Merged in aaed92f301a6862f30e817e23a50a21e8ae12e7b.

Issue #836, opened 2014-11-28
Per-project base page models


gasman:
Introduce an AbstractPage class to allow base page fields to be customised per-project. Proposed here: https://groups.google.com/d/msg/wagtail/4459qj1tNiU/rpaxthnn16MJ

## Working on this

Contribute to the discussion to help this move forward. Review [RFC 53: Swappable Page Model](https://github.com/wagtail/rfcs/pull/53). Explain your use cases, discuss possible solutions, research how to adapt existing code, or compatibility concerns.

The needed changes are very involved, this likely isn’t a good topic for new contributors.

alexgleason:
I'm still not convinced this is necessary, but I have seen an implementation which could work.

[Puput](https://github.com/APSL/puput) (a Wagtail blog app) has a mechanism which lets you [extend the entry page model](http://puput.readthedocs.org/en/latest/extending.html). It works because the [EntryPage model](https://github.com/APSL/puput/blob/005ecbee141fdfa0ab11bd5435dcb2989a7c23e3/puput/models.py#L154) extends [an abstract model which can be overridden](https://github.com/APSL/puput/blob/005ecbee141fdfa0ab11bd5435dcb2989a7c23e3/puput/models.py#L24) in settings.

Migrations also have to be accounted for, since by default they would be placed directly into the global Python module directory for wagtail/wagtailcore. Puput asks you to override your `MIGRATION_MODULES` Django setting to make EntryPage migrations occur in a local project folder. [More info on this is in the docs](http://puput.readthedocs.org/en/latest/extending.html#migrations).

SmileyJames:
Django Oscar has many models that can be overridden. It may be beneficial to look at their implementation: https://github.com/django-oscar/django-oscar

Any abstract_models.py file is a good place to start looking.

bergran:
Hello, I was looking for this feature in the documentation before. I was writing a proposal but I don't know why I search again for issues and got lucky! :)

now the proposal and why:

### Is your proposal related to a problem?
At this time I'm trying to make special categories for search in elastic search, this could be made by content_type but my categories have some special requirements with extra fields, so... I finally decide to make my category and relate to Page (wagtailcore.Page). It's easy if I do it with M2M but Page class is very close, nobody can write an additional field like a foreign key or another field as global scope for all my pages. In my case, I want to make (1:N) with my category where one page has only one category but one category can be on many pages.

How I said before I will use this for search and group by field this in the project logic

### Describe the solution you'd like

[...]

PR #1922, opened 2015-11-12
Enable use of alternative template engines for default model templates

moaxey:
I really prefer pyjade to django templates. This small change might help me and others.

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -193,7 +193,7 @@ class PageBase(models.base.ModelBase):
 
         if 'template' not in dct:
             # Define a default template path derived from the app name and model name
-            cls.template = "%s/%s.html" % (cls._meta.app_label, camelcase_to_underscore(name))
+            cls.template = "%s/%s.%s" % (cls._meta.app_label, camelcase_to_underscore(name), 'html' if not hasattr(settings, 'MODEL_TEMPLATE_TYPE') else settings.MODEL_TEMPLATE_TYPE)
 
         if 'ajax_template' not in dct:
             cls.ajax_template = None


wagtail/wagtailcore/models.py
```diff
@@ -193,7 +193,7 @@ def __init__(cls, name, bases, dct):
 
         if 'template' not in dct:
             # Define a default template path derived from the app name and model name
-            cls.template = "%s/%s.html" % (cls._meta.app_label, camelcase_to_underscore(name))
+            cls.template = "%s/%s.%s" % (cls._meta.app_label, camelcase_to_underscore(name), 'html' if not hasattr(settings, 'MODEL_TEMPLATE_TYPE') else settings.MODEL_TEMPLATE_TYPE)
```
nealtodd: As a setting `'html' if not hasattr(settings, 'MODEL_TEMPLATE_TYPE') else settings.MODEL_TEMPLATE_TYPE` would be neater as `getattr(settings, 'MODEL_TEMPLATE_TYPE', 'html')`.

However an undocumented, non-wagtail-scoped settings attribute probably isn't the best way to override the template extension. The template can be overridden by setting the template on the concrete Page class, but you do then need to specify the full template path (dir and file basename as well as the extension) which isn't terribly DRY when you just want to change the extension.

Maybe a documented optional `template_extension =  'html'` attribute for concrete `Page` classes is better with `PageBase` having something like:
```
if 'template' not in dct:
    template_extension = dct['template_extension'] if 'template_extension' in dct else 'html'
    cls.template = "%s/%s.%s" % (cls._meta.app_label, camelcase_to_underscore(name), template_extension)
```

Still needs the template_extension to be set on each concrete `Page` model with a non-html template extension, but it does make it explicit and maybe preferable to another `WAGTAIL*` setting. What do you think @gasman?

timheap:
When #1730 is merged, all Wagtail code will use the `Page.get_template()` method, instead of the automatically generated `Page.template` attribute. By using a customised base Page class for all Pages in your project, it would then be possible to make all your Pages use Jade templates without any further modifications to core:

```python
# in project/models.py
from wagtail.wagtailcore import models as wagtailcore

class Page(wagtailcore.Page):
    class Meta:
        abstract = True

    def get_template(self, request):
        # Use Jade templates
        return '{}/{}.jade'.format(self._meta.app_label, self._meta.model_name)
````
```python
# in home/models.py
# Inherit from the `project.models.Page` class instead of `wagtail.wagtailcore.models.Page`
from project.models import Page

class HomePage(Page):
    # Everything else as normal
```

Using a base Page class and a custom Meta class it is already possible to override the `Page.template` attribute, although it is not as clean. I did this in a number of projects when we used pyjade:

```python
# in project/models.py
import six

from django.db import models
from wagtail.wagtailcore import models as wagtailcore

class PageBase(wagtailcore.PageBase):

        if 'template' not in dct:
            cls.template = "{}/{}.jade".format(
                cls._meta.app_label, cls._meta.model_name)


class Page(six.with_metaclass(PageBase, wagtailcore.Page)):
    class Meta:
        abstract = True
```

gasman:
As a general policy we want to avoid letting global settings proliferate with "niche" functionality like this, and I haven't seen any other demand for non-`.html` template filename support since this was opened. If we were to pursue this, a custom base Page class as per #836 and https://github.com/wagtail/rfcs/pull/53 would be the way to go.

PR #1923, opened 2015-11-14
Bump browserify to 3.46.1

gasman:
3.36.0 depends (indirectly, via derequire 0.6.0) on esprima-six, which has disappeared from npm causing Wagtail's overall `npm install` step to fail. https://github.com/substack/node-browserify/pull/710

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   },
   "browserify-shim": {},
   "devDependencies": {
-    "browserify": "~3.36.0",
+    "browserify": "~3.46.1",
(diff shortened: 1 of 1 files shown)


kaedroho:
Thanks!

What are we using browserify for anyway?

davecranwell:
I'm not sure of its purpose. I had planned to depend on it, by starting to rewrite modules in CommonJS format, but hadn't go around to it. Perhaps let's not remove it quite yet?

PR #1925, opened 2015-11-14
Add Python < 3.3 support for frontendcache invalidator

### LINKED ISSUE #1924: Frontend cache invalidator HTTP Backend doesn't work on Python 2
Passing `method` as a keyword argument does not work on Python < 3.3. It gives an error:

```
__init__() got an unexpected keyword argument 'method'
```

```
request = Request(
            url=urlunparse([
                self.cache_scheme,
                self.cache_netloc,
                url_parsed.path,
                url_parsed.params,
                url_parsed.query,
                url_parsed.fragment
            ]),
            headers={
                'Host': host,
                'User-Agent': 'Wagtail-frontendcache/' + __version__
            },
            method='PURGE'
        )
```
I can work around this issue in my project by specifying a custom backend. I'll have to subclass `Request` to override its `get_method` method.

Additionally, the test suite does not explicitly test the invalidator backends. That's understandable though. Testing network requests isn't easy.

mixxorz:
The previous implementation was using a Python > 3.3 specific
implementation of urllib.request.Request which can accept a `method`
keyword argument in its constructor to specify what HTTP method to use
when executing the request. This is not available in earlier versions of
the class.

This commit fixes that incompatibility by instead subclassing Request
and overriding its `get_method` method.

Fixes #1924

### DIFF
--- a/wagtail/contrib/wagtailfrontendcache/backends.py
+++ b/wagtail/contrib/wagtailfrontendcache/backends.py
@@ -11,6 +11,11 @@ from wagtail.wagtailcore import __version__
 logger = logging.getLogger('wagtail.frontendcache')
 
 
+class PurgeRequest(Request):
+
+    def get_method(self):
+        return 'PURGE'
+
 class BaseBackend(object):
     def purge(self, url):
         raise NotImplementedError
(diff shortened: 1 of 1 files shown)


kaedroho:
Thanks, this looks good.

Relevant Python documentation: https://docs.python.org/3/library/urllib.request.html#urllib.request.Request.get_method

mixxorz:
Cool. Thanks for merging. :+1:

PR #1926, opened 2015-11-15
Submenu chevron direction and collapsible submenu in mobile view

carloratm:
Desktop view:
* changes the chevron direction when submenu is open

Mobile view:
* makes the submenu collapsible

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/main-nav.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/main-nav.scss
@@ -108,10 +108,15 @@ $submenu-color:darken($color-grey-1, 5%);
                 width:100%;
             }
         }
+
+        .submenu-trigger:after {
+            content: 'q';
+        }
     }
 
     .nav-submenu{
         background:$submenu-color;
+        display: none;
 
         h2{
             display:none;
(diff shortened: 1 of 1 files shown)


gasman:
Hi @ratmcarlo - sorry it's taken a while to get round to reviewing this!

Have rebased at https://github.com/gasman/wagtail/tree/ratmcarlo-submenu-chevron (hopefully I didn't introduce any errors in the process - I had to re-apply the changes manually, due to the various bits of linting/renaming that have happened in the meantime).

The collapsible menu in the mobile view is a great improvement, thanks! Changing the direction of the chevron seems a bit odd to me, but I'll let @davecranwell make that decision...

ratmcarlo:
@gasman ok, thank you

davecranwell:
The way you've used the `display` property is preventing the menu's opening animation from working, instead it snaps into place (display can't be animated, nor made to fire before the positional animations).

Once that's fixed, it might be nice if on both desktop and mobile, the arrow rotated into it's reverse position with an animation. It's non-essential on mobile, but I find myself agreeing with @gasman that it's behaviour on desktop seems a bit odd. I can't put my finger on why, but if the chevron rotation were animated, it perhaps wouldn't be so jarring on desktop.

gasman:
Superseded by #3069

PR #1928, opened 2015-11-16
Updated reasoning for Page suffix, fixes #1927

### LINKED ISSUE #1927: Naming convention of appending "Page" to each model in docs
"[Page models](http://docs.wagtail.io/en/v1.2/topics/pages.html)" in the docs says:

> **Tip:**
> To keep track of `Page` models and avoid class name clashes, it can be helpful to suffix model class names with “Page” e.g BlogPage, ListingIndexPage.

Can this be explained further, possibly linked to a more detailed article?

I don't totally get it. Does Django not allow a built in way to avoid name clashes in the database? It seems like the current method is to guess on naming your models so nothing clashes. Even then, having a model called `blog_page` in the DB isn't totally foolproof since another app could have that. Wouldn't you prefer something more like `wagtail_blog`? I think Django allows you to set that manually when you create a model.

It seems redundant to create a class with `class BlogPage(Page)`. I feel like this could be a problem if the only reason we're appending "Page" is to avoid a technical issue rather than focusing on best naming conventions and clean code.

alexgleason:


docs/topics/pages.rst
```diff
@@ -86,9 +86,9 @@ This example represents a typical blog post:
             FieldPanel('url'),
         ]
 
-.. tip::
+.. important::
 
-    To keep track of ``Page`` models and avoid class name clashes, it can be helpful to suffix model class names with "Page" e.g BlogPage, ListingIndexPage. 
+    Ensure that none of your field names are the same as your class names. This will cause errors due to the way Django handles relations (`read more <https://github.com/torchbox/wagtail/issues/503>`_). In our examples we have avoided this by appending "Page" to the name of each model name.
```
tomdyson: "to the name of each model name" -> "to each model name"

kaedroho:
Thanks!

PR #1929, opened 2015-11-17
Use SiteManager for Sites, add tests

timheap:
The SiteManager class was not used on the Site model, or anywhere else.

This was not caught by the tests, because nothing tested the relevan attributes on the Site model, nor were they used in the code base. Tests have been added now.


kaedroho:
Looks good to me!

mgd020:
should tests be integrated with TestSiteRouting?

gasman:
Merged in 1d4a276f49bcb0a3e7c1fb68cd287336811f8fa8

PR #1931, opened 2015-11-17
Replacing removeStyling() function with paste event interceptor to facilitate HTML contend editable functions

ptink:
The removeStyling() function strips out all `style=...` attributes on any <span> elements within a RichText area so that when content is pasted in from Word or elsewhere, the extraneous styling is removed immediately, rather than coming as a surprise to the user when it gets stripped out during the submission process.

However this also prevents the use of the HTML5 content editable functions such as backColor and hiliteColor however, which are browser-specific implementations and generally work by inserting a `background-color: yellow` <span> element around the text.

I propose that instead of stripping the styling out on paste, the paste command is intercepted and converted into plaintext, before being pasted into the RichText box- this also has the benefit of not leaving any rogue <span> elements behind after the styling has been stripped. The code in the pull request works in firefox/chrome, yet to be tested in IE but this stack answer may have a solution to that in the comments: http://stackoverflow.com/a/12028136/2547709

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
@@ -34,8 +34,8 @@ function makeRichTextEditable(id) {
     */
     }).bind('paste', function(event, data) {
         event.preventDefault();
-        var text = event.originalEvent.clipboardData.getData("text/plain");
-        document.execCommand("insertHTML", false, text);
+        var text = event.originalEvent.clipboardData.getData('text/plain');
+        document.execCommand('insertHTML', false, text);
     /* Animate the fields open when you click into them. */
     }).bind('halloactivated', function(event, data) {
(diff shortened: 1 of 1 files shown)


jimr:
It's worth noting that this will strip the "legitimate junk" from Word (etc) as well (bold, italic, hyperlinks, etc). Personally, I'm happy with all markup being stripped, but it's a change in behaviour that'll need documenting.

gasman:
I think the behaviour @jimr points out is a blocker here, I'm afraid. While the usefulness of preserving the "legitimate" markup is open to debate, there will certainly be users who will see it as a regression if we drop it. Preserving the default behaviour for existing users has to take precedence over enabling new customisations, I think...

There are a few items around rich text fields on our short-term todo list - making the underlying JS library swappable, and adding per-field customisation - which will help in this regard, I think, and once those are in place we'll hopefully be in a better position to revisit this.

PR #1279, opened 2015-05-07
Call Page.unpublish() on each page when unpublishing queryset Fixes #745

kaedroho:

PR #1934, opened 2015-11-18
Call Page.unpublish() on each page when unpublishing queryset

### LINKED ISSUE #745: PageQuerySet.unpublish doesn't call page_unpublished signal
This may cause issues with pages not being purged from the cache. This method is only currently used in Wagtail to unpublish pages after copying them (which don't get their page_published signal fired anyway).

I suggest we remove this method and add a ``keep_live`` attribute to the ``Page.copy`` method which when set to false, will not publish the copied pages in the first place.

gasman:
Fixes #745 (page_unpublished signal is now fired when unpublishing via queryset). Redo of #1279 with unit test.

### DIFF
--- a/wagtail/wagtailcore/query.py
+++ b/wagtail/wagtailcore/query.py
@@ -199,9 +199,10 @@ class PageQuerySet(SearchableQuerySetMixin, MP_NodeQuerySet):
 
     def unpublish(self):
         """
-        This unpublishes all pages in the QuerySet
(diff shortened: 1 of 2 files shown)


prydie:
Looks good to me

kaedroho:
Merged in 78ba194f48ab8c42bcaac022e1e17c4d74906247. Thanks!

PR #1935, opened 2015-11-18
Feature Request: Image filter chaining

shredding:
This PR makes filter chaining possible.

Filters do now accept a chain of specs in a pipe-like syntax. 

My use case is to add a `watermark` filter after applying a `fill-width-height` filter.

```python

# Converts the image with the fill operation to 480x200, applies a fictional blackandwhite filter and adds
# a fictional watermark with the fictional opacity of 1
spec = 'fill-480-200|blackandwhite|watermark-1
filter = Filter(spec=spec)

# likewise
rendition = image.get_rendition(spec)

# in templates it's just like args
{% image fill-480-200 watermark %}
# ... all tests for all those fancy image tag stuff pass ...
{% image fill-480-200 watermark  as img%}
```


shredding:
Oh, https://github.com/torchbox/wagtail/pull/1861 somehow went in. Let me fix that.

PR #1936, opened 2015-11-18
Feature Request: filter chaining

shredding:
This PR makes filter chaining possible.

Filters do now accept a chain of specs in a pipe-like syntax. 

My use case is to add a `watermark` filter after applying a `fill-width-height` filter.

```python

# Converts the image with the fill operation to 480x200, applies a fictional blackandwhite filter
# and adds a fictional watermark with the fictional opacity of 1
spec = 'fill-480-200|blackandwhite|watermark-1
filter = Filter(spec=spec)

# likewise
rendition = image.get_rendition(spec)

# in templates it's just like args
{% image fill-480-200 watermark %}
# ... all tests for all those fancy image tag stuff pass ...
{% image fill-480-200 watermark  as img %}
```


wagtail/wagtailimages/tests/tests.py
```diff
@@ -81,6 +81,12 @@ def test_image_tag_with_filters(self):
         self.assertTrue('width="400"' in result)
         self.assertTrue('height="300"' in result)
 
+    def test_image_tag_with_chained_filters(self):
+        result = self.render_image_tag(self.image, 'width-400|height-150')
+        self.assertTrue('width="200"' in result)
+        self.assertTrue('height="150"' in result)
```
kaedroho: This doesn't actually prove the first filter was run. How about ``fill-200x200|height-150``? The fill operation would alter the aspect ratio which should have an effect on the final width

kaedroho:
Thanks! This is something I've wanted to see for a while now. It's quite a widely requested feature:
https://github.com/torchbox/wagtail/issues/908
https://github.com/torchbox/wagtail/issues/1694

wagtail/wagtailimages/tests/tests.py
shredding: Yup, you're right.

I try to figure out why the other test fails. It works on my machine™ :)

shredding: Actually, the test is wrong anyway :) It should be `fill-200x200 height-150` (with a space). In the image-tag, the `|` shouldn't be used, as it's something for djangos filters.

shredding:
@kaedroho I've updated the test and changed the other one to make it work with tox. I do not really get why it was failing (though I could reproduce it on my local machine - test failed with tox but not when running just runtests) - `@patch` did not patch correctly.

However, I noticed that the tox.ini dependencies are different from the one in the repo and tox is at an oudated version (e.g. mock 1.0.1 vs 1.3.0, but that did not help with my failing test).

shredding:
An estimate on this would really help me. I have a Deadline for this watermark thingy and my favourite solution would be to merge this into our 1.2 and have it as a native Feature in 1.3 :)

kaedroho:
It might be a good idea to put the filter spec in quotes to simplify parsing: ``{% image myimage "op1|op2" as myimage %}``

gasman:
@kaedroho I disagree - I think that would be a step backwards for usability. At the moment, it's reasonably intuitive from the point of view of a template author - they're giving a list of criteria to the tag, using standard Django template tag conventions, and don't have to learn a new kind of syntax or understand what a 'filter spec' is)

(It would also introduce a possible backwards incompatibility: is an unquoted string treated as a single filter as it is now, or as a variable name for a string?)

shredding:
I agree with @gasman, the current approach is very similar to the methods `*args` approach and the filter represents an instance of something (the underlying filter) and does not represent a string.

kaedroho:
Currently, a filter is only differentiated from a keyword argument by the presence of an equals sign. If that is left out, or not formatted correctly, Wagtial would interpret them as filters:

``{% image fill-800x600 alt = "test" %}``
`` {% image fill-800x600 alt"test" %}``

This pull request also makes wagtail use a pipe character in the ``.get_rendition()`` and frontend serve view. I thought it would be good to keep that consistent with the template tag too.

Quoting something just makes Django bunch multiple words into a single "bit". It won't interpret non-quoted strings as variables so it would be backwards compatible.

gasman:
@kaedroho You could say the same thing about a lot of other template tags:

    {% include "name_snippet.html" with greeting="Hi" only %}
    {% include "name_snippet.html" with greeting = "Hi" only %}
    {% include "name_snippet.html" with greeting"Hi" only %}

The point is, there's a precedent within Django templates for treating tokens differently depending on whether they 'look' like args or kwargs, and I don't see any problem with following that precedent.

shredding:
I'd prefer the current syntax, however it'll get merged with @kaedroho approach, I'm happy to change it :)

kaedroho:
@gasman The include tag has simpler syntax though. The first argument is always the template and kwargs are always after "with".

Shall we add something like "with" to separate the filter from the kwargs?


Aanother example, this is valid ``{% image myimage alt="test" fill-800x600 myattr="foo" blah %}``

gasman:
I'll accept the "with" approach if that'll solve the impasse. I don't think it's needed though, and it causes an unnecessary backwards incompatibility. (So we'll need to use something like @shredding's existing code anyhow, while the current syntax goes through deprecation.)

`{% image myimage alt="test" fill-800x600 myattr="foo" blah %}` is horrible, yes, but it's unambiguous - we have some perfectly good well-defined and not-particularly-complicated rules to interpret what it means - so I see no reason to stop people from writing that if they insist.

kaedroho:
Alright, go ahead and merge this then.

wagtail/wagtailimages/tests/tests.py
```diff
@@ -81,6 +81,11 @@ def test_image_tag_with_filters(self):
         self.assertTrue('width="400"' in result)
         self.assertTrue('height="300"' in result)
 
+    def test_image_tag_with_chained_filters(self):
+        result = self.render_image_tag(self.image, 'fill-200x200|height-150')
```
gasman: `render_image_tag` here is constructing an `{% image %}` tag with the given parameter, so I would expect the filters here to be space-separated, not pipe separated. The fact that the test is apparently succeeding anyhow suggests that something fishy is going on with these tests...

gasman: Aha, it's not particularly fishy - it's just sneaking through the tag parsing code as a single unit, and when we do `'|'.join(filter_specs)` at the end, it's ending up as the same result that we would have got if they were space-separated.

Perhaps we should change the internal filter strings to space-separated too? That way, we avoid the possibility of the separator character being 'injected' in this way (and also means that an individual filter spec can safely include a `|` character without making the final chained string ambiguous, if we ever want to do that).

shredding:
Ive just seen the spaces as syntactic sugar on top of tje original syntax. However, ill add a test for an explicit change.

shredding:
Space, damn autocorrect

wagtail/wagtailimages/tests/tests.py
kaedroho: If the filter spec contains spaces, could we change them for a different character when generating the image filename? (I think spaces in filenames looks messy and they have to be urlencoded)

gasman:
@shredding Thanks! While you're updating it you might want to cherry-pick my edit from https://github.com/gasman/wagtail/commits/shredding-filter-chaining-feature, which catches a few more syntax errors. (I was about to write tests for these when I spotted the pipe vs space issue.)

shredding:
Ok, i had a closer look. It's not really on purpose, however - i do like that you can use both syntax variants and even interchange them.

> Perhaps we should change the internal filter strings to space-separated too?

Hm. The pipe is a much more familar syntax then a space - and if we want to disallow certain characters, the pipe would be better.

wagtail/wagtailimages/tests/tests.py
gasman: @kaedroho Agreed. We probably want to do this either way, as `|` characters are probably just as bad as spaces in the "likelihood of trashing your filesystem if you don't escape them properly" stakes :-) Converting to dots probably makes sense, since that's basically what we're doing with the filenames already (`badger.max-100x100.greyscale.jpg`).

gasman:
I'm not a fan of accepting the pipe character within the tag - it makes our code harder to follow and more error-prone, because it's blurring the distinction between "a filter spec" and "a list of filter specs".

As an alternative approach - how about we tighten up the rules, so that individual filter spec strings can ONLY contain the characters `A-Za-z0-9_-` ? That way:

* we can still use `|` in the internal filter string
* we can ensure that the tag only accepts spaces (because the pipe-separated version would fail validation)
* it avoids the wackiness @kaedroho is concerned about, where `{% image fill-800x600 alt"test" %}` would be treated as valid
* we don't have to worry about unsafe characters in filenames (because we can just do `'.'.join(filter_specs)`)

kaedroho:
I'm happy with that

gasman:
(I guess we might want to allow the `.` character within filters too, so that we can say things like `opacity-0.5`. Which would give us filenames like `badger.opacity-0.5.max-100x100.jpg` where you can't tell if the dot is a separator or not - but that's probably okay)

shredding:
I added a few more test and a pattern test for the filter and cherry picked your change.

Where would be a good place to change the pipe to a dot for filename output?

Somewhere along these lines i guess?

https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailimages/models.py#L270-L271

shredding:
I've added a few more tests, a pattern matching test on filters and changed the filename generation for piped filters to dots!

shredding:
Maybe we should ship a sample filter to show the use case. My watermark filter is using some pillow interfaces, so maybe something that can be done with willow? A blackandwhite filter maybe. Its a bit hard to update the docs without a use case.

gasman:
We're hoping to have the 1.3 release wrapped up today, so we don't really have chance to add any sample filters or full documentation. I think it's fine for this to be a 'stealth' feature in the current version though, and have its full usefulness become apparent later on :-)

(The 'jpeg compression level' filter is probably the most immediate use for this in the short term. But that's a bit of an odd one to use as an example in the documentation, I guess...)

kaedroho:
Just thought it's worth mentioning that running a fill operation after a previous resize on an image with a focal point might create weird results as the resize operations don't transform the focal point of the image. Would have to think about that one.

shredding:
Fine with me :) :+1:

tomdyson:
We could add docs between RC and release, couldn't we? It'd be nice to announce the feature next week if possible.

gasman:
@tomdyson The docs aren't really the problem - it's that currently Wagtail doesn't provide any filters that can meaningfully chained together. As it stands, the only way of taking advantage of this feature is by registering your own custom filters, which is an undocumented API / hook.

I guess if we're confident enough that that API is suitably stable and complete, we could document that and introduce the concept of filter chaining on top of that...

tomdyson:
@gasman understood, thanks for explaining. @shredding are you able to document the API in the next few days?

gasman:
I should emphasise that the "if" in my above comment is a rather big one :-) @kaedroho is in the best position to judge whether that API is ready to become official.

gasman:
Merged in c6b2cef5759231ea76ac7c6ee7dca8f7f57c492f.

PR #1938, opened 2015-11-19
Allow overriding attributes in {% image %} tag

### LINKED ISSUE #1933: Cannot overwrite image attributes in template with empty values
For certain images in my templates I want to set an empty alt attribute.

[According to the documentation](http://docs.wagtail.io/en/v1.2/topics/images/using_in_templates.html#image-tag-alt) this is done with

    {% image obj.image alt="" %}

But this does not work, and I think it should.

If I overwrite an attribute this way, wagtail should not make any checks of my values and just accept them.

timheap:
The `alt` attribute that was automatically generated as part of the `{% image %}` tag could not be overridden. If template authors passed their own `alt="..."` attribute in, two would be printed out instead
of the default one being overridden.

The relevant code has been refactored to build a dict of attributes, allowing the default set to be overridden, and then printing them using `flatatt`.

Fixes #1933


wagtail/wagtailimages/models.py
```diff
@@ -441,16 +442,28 @@ def url(self):
...
     @property
     def attrs(self):
-        return mark_safe(
-            'src="%s" width="%d" height="%d" alt="%s"' % (escape(self.url), self.width, self.height, escape(self.image.title))
-        )
+        """
+        The src, width, height, and alt attributes for an <img> tag, as a HTML
+        string
+        """
+        return flatatt(self.attrs_dict)
```
prydie: This is not exactly equivalent as `flatatt()` prepends a leading space.

> The returned string will contain a leading space followed by key="value", XML-style pairs.

That said I can't foresee that being an issue other than potentially increasing payload size of a few web pages by a byte.

timheap: Good catch. Fixed in the latest push.

nealtodd:
Thanks for that Tim. Could you include a tweak to the docs in this PR for:

https://github.com/torchbox/wagtail/blame/master/docs/topics/images/using_in_templates.rst#L137

`No validation is performed on attributes added in this way so it's possible to add `src`, `width`, `height` and `alt` of your own that might conflict with those generated by the tag itself.`

Changing to better reflect the behaviour now, e.g.

`No validation is performed on attributes added in this way but `src`, `width`, `height` and `alt` will override those generated by the tag itself.`

timheap:
I've updated the documentation:

> You can set a more relevant `alt` attribute this way, overriding the one automatically generated from the title of the image. The `src`, `width`, and `height` attributes can also be overridden, if necessary.

nealtodd:
Sounds perfect. Code and docs all look good to me.

gasman:
Merged in f4968de0f941ffabb4809d73df62e2e87815d816

PR #1940, opened 2015-11-19
Add hook to allow adding buttons to page listing 

timheap:
The 'register_page_listing_buttons' hook allows plugin developers to add buttons to the actions list on the page listing in the admin.

This will come in useful when adding the revision rollback feature in the future, as the rollback plugin can add a button to the page listing.

Does this concept sound like a good plan to you? Documentation and tests will be coming shortly.

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_indicator.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_indicator.scss
@@ -24,10 +24,6 @@ $c-indicator-margin: .25rem;
 // States
 // =============================================================================
 
-$color-state-live: #59b524;
-$color-state-draft: #808080;
-$color-state-absent: #ff8f11;
-$color-state-live-draft: #43b1b0;
 
 .is-absent .c-indicator {
     background: $color-state-absent;
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_indicator.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_indicator.scss
@@ -1,0 +1,66 @@
+// =============================================================================
+// Indicator light
+// =============================================================================
+
+
+.o-icon-text {
+
+}
+
+// =============================================================================
+// Indicator light
+// =============================================================================
+$c-indicator-size: .625em;
+$c-indicator-margin: .25rem;
+
+.c-indicator {
+    display: inline-block;
+    border-radius: 50rem;
+    width: $c-indicator-size;
+    height: $c-indicator-size;
+    margin-top: -.125rem;
+    margin-right: $c-indicator-margin;
+    font-size: 1rem;
+    vertical-align: middle;
+}
+
+// =============================================================================
+// States
+// =============================================================================
+
+$color-state-live: #59B524;
+$color-state-draft: grey;
+$color-state-absent: #FF8F11;
+$color-state-live-draft: #43B1B0;
+
+.is-absent .c-indicator {
+    background: $color-state-absent;
+}
+
+.is-live .c-indicator {
+    background: $color-state-live;
+}
+
+.is-draft .c-indicator {
+    background: $color-state-draft;
+}
+
+// This is hipster. But it works.
+.is-live\+draft .c-indicator {
+    background: $color-state-draft;
+    position: relative;
+
+    &:before {
+        content: '';
+        width: $c-indicator-size/2;
+        height: $c-indicator-size;
+        position: absolute;
+        top: 0;
+        left: 0;
+        border-bottom-left-radius: 50rem;
+        border-top-left-radius: 50rem;
+        background: $color-state-live;
+        transform: rotate(45deg);
+        transform-origin: 100% 50%;
+    }
+}
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_indicator.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_indicator.scss
@@ -3,10 +3,6 @@
 // =============================================================================
 
 
-.o-icon-text {
-
-}
-
 // =============================================================================
 // Indicator light
 // =============================================================================
@@ -28,10 +24,10 @@ $c-indicator-margin: .25rem;
 // States
 // =============================================================================
 
-$color-state-live: #59B524;
-$color-state-draft: grey;
-$color-state-absent: #FF8F11;
-$color-state-live-draft: #43B1B0;
+$color-state-live: #59b524;
+$color-state-draft: #808080;
+$color-state-absent: #ff8f11;
+$color-state-live-draft: #43b1b0;
 
 .is-absent .c-indicator {
     background: $color-state-absent;
--- a/wagtail/wagtailadmin/templatetags/wagtailadmin_tags.py
+++ b/wagtail/wagtailadmin/templatetags/wagtailadmin_tags.py
@@ -1,5 +1,7 @@
 from __future__ import unicode_literals
 
+import itertools
+
 from django.conf import settings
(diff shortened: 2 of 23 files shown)


davecranwell:
I like the principal of users being able to add their own buttons in situations like this, but fwiw I'm not a fan of adding more buttons to the explorer specifically.

At present there are already buttons in the explorer allowing you to do things to a page which aren't available within the page editor itself, which is a mistake in my opinion. 

If you edit an image, you expect to be able to control the image entirely through the image editor interface. If you edit a document, you expect to be able to do everyhting you need through the document editor interface. If you edit a page however, you can't copy it, move it, add a child, or view the draft without using the awkward preview system, from within the page itself. 

While we may need to add a revisions button in the explorer _temporarily_ I'm wary of giving users the ability to further scatter action buttons who's locations ought to be more predictable, to reduce users' needs to think.

timheap:
I hear what your saying, where buttons appear is a little confusing and inconsistent currently. One (possibly silly) solution: just add this button row to the edit screen as well. Then which buttons appear are consistent across all instances of the button row, everywhere.

davecranwell:
That's certainly an option yes, although that would be inconsistent with the header areas of other entity in Wagtail and the page editor header is already creaking and poorly aligned. 

What i'm really after is a common UI paradigm on all entities, something more like the right-hand column on the image and document editing pages, in which users can reliably expect to find the Buttons Wot Do Stuff.

timheap:
A right-hand column of Buttons Wo Do Stuff would not play nice with the current full-bleed interface for some fields. I know there are plans to redo this part of the UI, and I am not much of a UI designer myself, so I dont want to step on any toes here. If I make a _`'register_button_wot_does_stuff '` hook, could we then discuss how to move forward implementing the UI aspect of this?

davecranwell:
Yeah definitely go ahead with the hook. We can figure the rest out later.

gasman:
For the revisions feature, I think the plan was to link it from the "last edited by..." text in the footer of the page editor, rather than adding any extra buttons. (I think we really need a proper written spec for revisions at this point, to stop people going up the garden path...)

wagtail/wagtailadmin/templatetags/wagtailadmin_tags.py
```diff
@@ -270,3 +272,14 @@ def paginate(context, page, base_url='', page_key=DEFAULT_PAGE_KEY,
...
+
+
+@register.inclusion_tag("wagtailadmin/pages/listing/_buttons.html",
+                        takes_context=True)
+def page_listing_buttons(context, page, page_perms, is_parent=False):
+    button_hooks = hooks.get_hooks('register_page_listing_buttons')
+    buttons = sorted(itertools.chain.from_iterable(
+        hook(page, page_perms, is_parent)
+        for hook in button_hooks))
+    print(buttons)
```
kaedroho: Stray ``print()``!

wagtail/wagtailadmin/wagtail_hooks.py
```diff
@@ -30,3 +31,29 @@ def register_settings_menu():
 @hooks.register('register_permissions')
 def register_permissions():
     return Permission.objects.filter(content_type__app_label='wagtailadmin', codename='access_admin')
+
+
+@hooks.register('register_page_listing_buttons')
+def page_listing_buttons(page, page_perms, is_parent=False):
```
kaedroho: It might be nice if the hook didn't have to care about where/how the buttons are displayed so third party apps wouldn't have to be updated if we ever decide to change where to display the buttons.

nkuttler:
I just found this pr because I have exactly the same problem. In my case I just want an additional button for my private website, and there's no way to add it without overriding almost the entire admin template. I think the edit.html with the page editor should have a similar mechanism for adding actions. I'll keep an eye on this pull request and override entire admin templates for now.

tomdyson:
Here's @JoshBarr's mockup of an idea to solve the too-many-buttons problem.

josh-extended-buttons-mockup

kaedroho:
This is missing the required CSS changes for the new dropdown. Could you add these?

timheap:
OK, this has got the styles from @JoshBarr now. There is a small issue with button sizes in the list of buttons in Firefox, in that they are all different heights, although they only differ by ±1px. I moved the hard-coded example list of traffic light buttons to the styleguide, but the styles break a little when moved out of their current location. At least the styles/html is preserved for future reference, even if the look is not exactly correct.

screenshot-wagtail-buttons

tomdyson:
Hi @m1kola could you please review this? @joshbarr will look at the CSS issues separately.

m1kola:
@tomdyson I will look at this tomorrow evening. Is it ok?

tomdyson:
@m1kola yes sure!

m1kola:
Thanks, @timheap! I reviewed templates, js and python code. I have no questions on current version.

We need to add docs and tests.

m1kola:
@timheap, I'm working on tests for this PR. I hope tomorrow they will be ready.

m1kola:
I sent PR with some tests to `takeflight:feature/page-listing-buttons` (see https://github.com/takeflight/wagtail/pull/3).

gasman:
Note to whoever ends up resolving merge conflicts: please take care not to lose the new `.warning` button styles added to `_forms.scss` in e782de6064f2a9b79d871e40bf6d61c8061b7c1a.

gasman:
Rebased on master at https://github.com/gasman/wagtail/tree/takeflight-feature/page-listing-buttons (including @m1kola's tests https://github.com/takeflight/wagtail/pull/3).

I'm afraid I have severe accessibility concerns about the dropdown widget in its current form - if you click on the 'More...' button of a child page in the explorer, and then your mouse pointer strays outside the boundaries of the dropdown while attempting to click on one of the options, the dropdown disappears. I don't think my motor control is particularly bad, but I found it just fiddly enough to be annoying. I think we need some extra JS to make the menu (and the rest of the row's buttons) 'stick' until the user explicitly clicks outside of it. (As a more nitpicky change - I think the links on the dropdown items also need to be made block-level, i.e. cover the full width of the dropdown rather than just the text span.)

SalahAdDin:
@tomdyson Man, how did you do the translation button options? How works your translation system?

JoshBarr:
@gasman i didn't realise you'd rebased this one elsewhere. I've updated the JS to address the usability issue. Would you like to add the tests to this PR, or move the JS to the new PR?

gasman:
@JoshBarr Thanks! This PR is looking a lot neater than my rebased branch now, so I'd suggest cherry-picking @m1kola's commit https://github.com/m1kola/wagtail/commit/4f1125ea2c181c629b7bffd1a1579658ee05d06c on to here.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_indicator.scss
```diff
@@ -0,0 +1,62 @@
...
+    margin-right: $c-indicator-margin;
+    font-size: 1rem;
+    vertical-align: middle;
+}
+
+// =============================================================================
+// States
+// =============================================================================
+
+$color-state-live: #59b524;
```
davecranwell: Seems likely we're going to want to use these colours to replace the main Explorer status tag soon. Could this be moved to the main variables.scss?

JoshBarr:
I've cherry-picked the tests and added some documentation for this feature. Should be good to merge.

gasman:
Merged in 680307518852a212483ae533d438ec02aabc0a89 + parents.

@JoshBarr Thanks for getting those last fixes in so quickly! Sorry that the last-minute reviewing meant we weren't able to get this into 1.4 - it just missed the cutoff by a whisker - but we weren't able to postpone the release candidate any longer, as we're under a lot of pressure to get the likes of collections into a release version.

PR #1946, opened 2015-11-20
Move rollbacks to contrib module, added diff view for textual diffs

JoshBarr:
I've imported the rollbacks module from Wagtailplus and wired it up in contrib.

Also, have included a simple diff view for comparing the text contents of a page: 

screen shot 2015-11-20 at 12 48 43 pm
 
This needs a little work (what do do if a Page has no text fields, for instance), but thought it would be a useful addition to the history module – especially if you're using Wagtail as an API, where previewing via the front-end won't work so well.


shredding:
Great.

gasman:
Nice one! I think it would make sense for this to be folded into wagtailadmin rather than being a separate contrib app. (Having it decoupled from the rest of wagtailadmin and integrated via hooks is a good thing, so I wouldn't propose any major structural changes - just placing it somewhere like wagtail.wagtailadmin.revisions and moving the hooks registration so that it doesn't need its own entry in INSTALLED_APPS.)

tomdyson:
Your diff view looks great, @JoshBarr! Thanks a lot.

wagtail/contrib/wagtailrollbacks/diff_tools.py
```diff
@@ -0,0 +1,194 @@
...
+
+def list_diff(a, b):
+    # print a, b
+    pairs = zip(a, b)
+
+    difference = sum(x != y for x, y in pairs)
+    difference += abs(len(a) - len(b))
+
+    if difference == 0:
+        print 'The lists are the same length and appear to contain the same items'
```
kaedroho: can we remove the print statements or convert them into print functions?

bouncingdan:
Suggested implementation of revisions button in the admin footer (during sprint session with Torchbox / fffunction)

screen shot 2015-11-27 at 15 38 41

```css
@mixin slideleft {
    0%      { opacity: 0; transform: scale(1.15); }
    100%    { opacity: 1; transform: scale(1);  }
}

@-webkit-keyframes  slideleft { @include slideleft; }
@-moz-keyframes     slideleft { @include slideleft; }
@-ms-keyframes      slideleft { @include slideleft; }
@keyframes          slideleft { @include slideleft; }

.meta {
    position: relative;
}

.modified {
    float: right;
    padding-right: 5px;
    animation: slideleft 0.5s 1 forwards;
}
```


templates\wagtailadmin\pages\edit.html added button to admin footer template
```html
<li class="meta">
    {% if page.get_latest_revision %}
        <p class="modified">
            {% blocktrans with last_mod=page.get_latest_revision.created_at %}Last modified: {{ last_mod }}{% endblocktrans %}
            {% if page.get_latest_revision.user %}
                {% blocktrans with modified_by=page.get_latest_revision.user.get_full_name|default:page.get_latest_revision.user.get_username %}by {{ modified_by }}{% endblocktrans %}
                {% if page.get_latest_revision.user.email %}
                    <span class="avatar small icon icon-user"><img src="{% gravatar_url page.get_latest_revision.user.email 25 %}" /></span>
                {% endif %}
            {% endif %}
            <button class="button-small">Revisions</button>
        </p>
    {% endif %}
</li>
```

tomdyson:
Some context on @bouncingdan's comment, for @JoshBarr and others. Torchbox and fffunction discussed revisions today, as part of a day-long sprint. We agreed that a new tab (as used in the Wagtailplus Rollbacks app) isn't the right place to access revision history; for one thing, the current edited state of the content edit form shouldn't be lost when moving between tabs, and clicking 'compare' within the 'history' tab breaks this pattern by moving to a new URL. We suggested that the right hand side of the footer bar is a better fit, partly because it's linked to the existing information about the most recent edit. 

The next question is where to present the listing of revisions, if it's not in a new tab. We want to avoid another modal, because modals are hard to handle on mobile devices. We suggest a new view, with enough context so that it's easy to return to the edit view. On a related point, we want to implement a 'dirty forms' handler, e.g. https://github.com/codedance/jquery.AreYouSure, to prevent users from inadvertently losing their edits when they navigate away.

gasman:
Work on the core revisions feature is being finalised at #2170; will keep this issue open (with no set milestone) for work on the diff view.

SalahAdDin:
:+1:

timheap:
Is this issue still relevant? The revisions work got merged in #2170. This issue also mentions a diff view that would still be worth investigating, but possibly in a fresh issue.

kaedroho:
I agree with that. The diff work in this PR would need to be separated out anyway if we are going to use it.

PR #1572, opened 2015-08-02
Fix/subpage type check on move

timheap:
This is an updated version of #638, rebased on top current master, and with my feedback implemented. It solves a bunch of use cases, and restricts move destinations based on parent/subpage types.

PR #1950, opened 2015-11-20
Deprecate ContentType-returning methods in wagtailcore

gasman:
Cleanup for #1572, as per https://github.com/torchbox/wagtail/pull/1572#issuecomment-158382266


gasman:
rebased

timheap:
Looks good to me! This will simplify some of the logic immensely.

There are a few other places in the codebase where content types are used needlessly - such as snippets, and some admin widgets - but they have nothing to do with Page relations. That can come as another PR later.

PR #1954, opened 2015-11-23
Move content_panels inside BookPage definition.

richbrennan:


docs/reference/contrib/forms.rst
```diff
@@ -37,17 +37,17 @@ Within the ``models.py`` of one of your apps, create a model that extends ``wagt
...
-        InlinePanel('form_fields', label="Form fields"),
-        FieldPanel('thank_you_text', classname="full"),
-        MultiFieldPanel([
-            FieldPanel('to_address', classname="full"),
-            FieldPanel('from_address', classname="full"),
-            FieldPanel('subject', classname="full"),
-        ], "Email")
-    ]
+        content_panels = AbstractEmailForm.content_panels + [
+            FieldPanel('title', classname="full title"),
```
gasman: If we're appending to `AbstractEmailForm.content_panels`, presumably we don't need to add the title field here?

gasman:
Other than the above quibble, looks good to me!

richbrennan:
That's true, and exactly what I did in my code!

gasman:
Merged in ce57281f6af3ce6024b95211972c4da4fd8749b0 + parents. Thanks!

PR #1959, opened 2015-11-24
Fixed l18n case in russian

Uznick:
Fixed accusative case for username placeholder in admin login


gasman:
Hi @Uznick - thanks very much for the correction! Just so you're aware - we manage our translations through [Transifex](https://www.transifex.com/torchbox/wagtail/) - I've made this fix over there now, so it will be updated in the codebase when we next do a pull. If you find any other things in future that need fixing, please submit them through Transifex.

Uznick:
@gasman Thank you for the info, I'll do it there next time.

PR #1960, opened 2015-11-25
Update content_panels and promote_panels on models

richbrennan:
If I'm right, the Echoer content_panels and promote_panels in model recipes can be removed as they're defined on Page?


gasman:
Merged in 11618f0c3d4033fdf1b2a852c855ba2cb797c0aa.

PR #1962, opened 2015-11-26
Feature/maplecroft/custom admin search fields

ajdegol:
Hi there,

As brought up here https://groups.google.com/forum/#!topic/wagtail-developers/1Dq0pmGe7i8

We'd like the admin/editor page search to be able to query more fields than just the title. Obviously there are a few extra things (tests/docs) that would need to be done, so this PR is to start a discussion on whether this approach generally makes sense and/or is worth pursuing.


wagtail/wagtailadmin/views/pages.py
```diff
@@ -678,7 +679,9 @@ def search(request):
         if form.is_valid():
             q = form.cleaned_data['q']
 
-            pages = Page.objects.all().prefetch_related('content_type').search(q, fields=['title'])
+            search_fields = getattr(settings, 'WAGTAIL_ADMIN_PAGE_SEARCH_FIELDS', ['title'])
```
kaedroho: ``WAGTAIL_ADMIN`` should be ``WAGTAILADMIN`` (for consistency with other settings)

ajdegol: Good catch, thank you.  Have made the change.

jimr:
Is there any reason why we can't use `wagtailsearch` for admin searches? For example, this could be as simple as

```python
from wagtail.wagtailsearch.backends import get_search_backend

...

pages = get_search_backend().search(q, Page)
```

The added benefit here being that it wouldn't require an additional setting, it would instead use the `search_fields` as defined on any `Page` subclasses meaning we wouldn't have to worry about certain fields not being defined on all such classes.

gasman:
Fixed in #2279 - admin search now searches over all searchable fields.

PR #1963, opened 2015-11-27
Updated Site.find_for_request() to only hit the database once

mgd020:
* Updated requirements-dev.txt to make unit tests pass
  * `djangorestframework` was missing, causing import errors
  * `django-modelcluster` 0.6.2 was too old: `childformset_factory` did not include `validate_min`
* Updated `Site.find_for_request()` to only hit the database once (down from a possible 3)
  * was already one if unique hostname
  * was two if non-unique hostname but (host, port) match
  * was three if non-unique hostname with no matching port and a default
* Updated `TestSiteRouting` tests to ensure only one query per each test


davecranwell:
Nice work, thanks @mgd020 

`djangorestframework` and `django-modelcluster` are already a requirement of wagtail itself. See https://github.com/torchbox/wagtail/blob/master/setup.py. Was your wagtail installation or your development machine out of date perhaps?

mgd020:
Thanks @davecranwell! Hmm, I pulled and pip installed requirements-dev.txt, I guess this was incorrect! I will roll back that commit!

wagtail/wagtailcore/models.py
```diff
@@ -107,21 +107,59 @@ def find_for_request(request):
...
+            return site_list[0]
+
+        # DoesNotExist
+        if len(site_list) == 0:
+            raise Site.DoesNotExist()
+
+        # find the default_site (if any)
+        # remove from list if hostname does not match
+        # we need to check both ends as different dbs store and order booleans differently (see django #19726)
+        default_site = None
```
kaedroho: I think we should simplify this a bit. Perhaps a list comprehension?

```python
site_list = [site for site in site_list if site.hostname == hostname]
``

Negligable performance impact and much more expressive.

mgd020: We would still need to find which is the default site, and that
would require another O(n) search. Then removing it alone with a whole list
comprehension would be another O(n) (there will only ever be one Site with
a non-matching hostname, and there could be none!).

The way I have written it is O(1): no matter how big the site_list, the
default site will be either first or last (or neither).

I agree that it is more verbose this way, but I fail to see negligable
impact. The catalyst for my creating this PR was hitting the db twice for
every non-default site page view, esentially doubling the time it takes to
work out which site the request is for.

Maybe you could provide a full alternate solution including handling the
default site id and removal? What you have posted would not produce the
same results as the PR code. But I am interested to see what you mean fully.

kaedroho: > The way I have written it is O(1): no matter how big the site_list, the default site will be either first or last (or neither).

I don't think the complexity of this particular filter really matters. For most sites, ``site_list`` wouldn't have more than a few items. And even if one did have thousands of sites, the effort required to pull them all out of the database would probably be much greater anyway.

I'm happy for this to be merged without my suggestion though, just thought it made the code  easier to follow.

> but I fail to see negligable impact

Sorry, I meant the list comprehension has a negligable impact

mgd020: If one had thousands of sites, the original 3 query version would be
optimal. This version would be optimal if the count is low, or the total
sites with matching hostname is low.

gasman:
From a bit of hacky benchmarking, namely changing `SiteMiddleware.process_request` to:

    def process_request(self, request):
        """
        Set request.site to contain the Site object responsible for handling this request,
        according to hostname matching rules
        """

        from datetime import datetime
        start_time = datetime.now().microsecond
        try:
            request.site = Site.find_for_request(request)
        except Site.DoesNotExist:
            request.site = None
        end_time = datetime.now().microsecond
        print("find_for_request completed in %d" % (end_time - start_time))

...I'm finding that this patch takes `Site.find_for_request` down from ~20ms to ~12ms on my machine, which seems like a decent improvement, so happy to merge this. The new code isn't quite as readable as I'd like, but I guess that just puts the onus on me to come up with something better... :-)

mgd020:
I have another idea to make it more readable..

gasman:
Thanks! Have rebased on top of current master (including re-applying the fix from #2359) at https://github.com/gasman/wagtail/tree/mgd020-site-routing.

Putting the main logic in SQL is definitely a good move, although the way it all fits together still seems pretty esoteric - I'm still not entirely sure why `MATCH_DEFAULT` ends up being higher precedence than `MATCH_HOSTNAME`. Could you add some more code comments please?

mgd020:
Let me know if the comments aren't clear.

MATCH_HOSTNAME could be before MATCH_DEFAULT, but as we need to see if
there is a default (only one) and there could be many hostnames it would
mean accessing sites[-1] instead of just sites[0]. Also if there are only 2
matches [(default, hostname)] we want to return the hostname, but if there
are three [(default, hostname, hostname)] we want to return default.

gasman:
Merged in 9bbf1ba9a0bc20fc65656ee85d6d19890b8ab002 - thanks!

Issue #1948, opened 2015-11-20
Refactor scss in admin


alexgleason:
Copying from a closed PR from September 23rd (#1731), this is a list of unused mixins:

* @mixin filter($filter...)
* @mixin transition-delay($delay...)
* @mixin notransition()
* @mixin content-box()
* @mixin table-column($x, $padding:$padding, $grid-columns:$grid-columns)
* @mixin push($offset:1, $grid-columns:$grid-columns)
* @mixin push-padding($offset:1, $grid-columns:$grid-columns)
* @mixin pull($offset:1, $grid-columns:$grid-columns)
* @mixin pull-padding($offset:1, $grid-columns:$grid-columns)

davecranwell:
All the mixins related to prefixing should be removed in favour of gulp-autoprefixer which is _already_ part of the Gulp compilation task, but seems to have gone unnoticed.

The grid related ones should remain. We may not use them, but then we don't use "cols-7" or "cols-11" similarly. They should remain as part of a reasonably feature-rich grid framework that others can use.

gasman:
Fixed in #1964

PR #1964, opened 2015-11-27
Removed all mixins that provide vendor prefixes, instead relying on autoprefixer

davecranwell:
Autoprefixer has been installed since the move to gulp, but our code still needlessly relied on mixins to perform the same role. This PR removes those mixins along with several others which were redundant. 

Addresses #1948

### DIFF
--- a/gulpfile.js/tasks/styles.js
+++ b/gulpfile.js/tasks/styles.js
@@ -32,7 +32,7 @@ gulp.task('styles:sass', function () {
             outputStyle: 'expanded'
         }))
         .pipe(autoprefixer({
-            browsers: ['last 2 versions'],
+            browsers: ['last 3 versions', 'not ie <= 8'],
             cascade: false
         }))
         .pipe(gulp.dest(function(file) {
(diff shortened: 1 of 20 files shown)


gasman:
@davecranwell The changes look good to me, but when I run this locally I can't see any evidence that gulp-autoprefixer is actually doing anything. For example, I'm assuming that you've removed the `-moz-box-sizing: border-box;` lines from wagtailadmin/scss/components/datetimepicker.scss on the basis that gulp-autoprefixer is expected to add that automatically - but when I check the `.xdsoft_datetimepicker` rule in static/wagtailadmin/css/core.css, only the unprefixed `box-sizing: border-box;` line is there. Am I missing something?

davecranwell:
Nope, you're not missing anything - box sizing is just a bad example to test since it has been supported natively by all the browsers we support, and no prefixing is necessary: http://caniuse.com/#search=box-size

A better test is `transform`, which required prefixing in IE9 and Safari: http://caniuse.com/#search=transform

gasman:
Ah yep, cheers. Can confirm that this line https://github.com/torchbox/wagtail/pull/1964/files#diff-72f8eaf1811bb083c7d10794628b390eR434 is successfully being rewritten to

    -webkit-transform: none;
    -ms-transform: none;
    transform: none;

gasman:
Merged in eaeb589ebb8255a53471d27e02efdaf3ae38a564 + parents

PR #1970, opened 2015-11-27
Issue with multiple word tags in Images and Documents views.

LKozlowski:
When you use tagging feature for Images and Documents you can't save a model with only a single tag containing multiple words in it.
It will create separate tags.

Steps to reproduce:
 1. Start editing a document in Wagtail admin. 
2. Ensure tags field is empty.
 3. Start typing a single tag with a double quote and enter multiple words, then close it with double quote. 4. Save the document.
 5. Edit the document again—you will see multiple tags instead a one with multiple words.

The problem is that tag fields are not preprocessed by this function:
https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js#L133

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
@@ -21,6 +21,20 @@ function escapeHtml(text) {
     });
 }
 
+function initTagField(id, autocompleteUrl) {
+    $('#' + id).tagit({
+        autocomplete: {source: autocompleteUrl},
+        preprocessTag: function(val) {
+            // Double quote a tag if it contains a space
+            // and if it isn't already quoted.
+            if (val && val[0] != '"' && val.indexOf(' ') > -1) {
+                return '"' + val + '"';
+            }
+            return val;
+        }
+    });
+}
+
 $(function() {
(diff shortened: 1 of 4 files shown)


wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
```diff
@@ -21,6 +21,20 @@ function escapeHtml(text) {
...
 
+function initTagField(id, autocompleteUrl) {
+    $('#' + id).tagit({
+        autocomplete: {source: autocompleteUrl},
+        preprocessTag: function(val) {
+            // Double quote a tag if it contains a space
+            // and if it isn't already quoted.
+            if (val && val[0] != '"' && val.indexOf(' ') > -1) {
+                return '"' + val + '"';
+            }
```
tomdyson: The JS linter is asking for a newline here (before `return val;`).

tomdyson:
I can confirm

1. that this fails in 1.2 just as @LKozlowski describes
2. that this PR fixes the issue

Thanks, @LKozlowski!

gasman:
Merged in 8a22c977a2f0fb34017c4e1c280ec865d2a02b59 + parents - thanks!

Issue #1559, opened 2015-07-29
Notification emails are sent "to:" all recipients in the clear.


davecranwell:
Moderation notification emails are sent with every recipient's email visible in the "to:" line, which seems yucky. BCC-ing everyone makes more sense.

gasman:
A better solution would be to send a distinct email to each recipient. This would allow personalisation (such as greetings, and possibly an 'unsubscribe from these notifications' link with an embedded login token), and more importantly, is going to be necessary if/when we implement per-user language settings.

davecranwell:
Ah yes, that would be preferable. 

I suppose there are some considerations about how mail servers might respond to Wagtail performing bulk emails, but they'd be just as concerned with a massive bcc, so perhaps it's no more problematic. Mailing would need to be queued, if it isn't already.

gasman:
We previously had code to queue email using Celery when available, to account for this, but took it out because it was effectively pushing more setup demands on the developer without warning them: http://docs.wagtail.io/en/v1.0/releases/1.0.html#celery-no-longer-automatically-used-for-sending-notification-emails .

I think it makes sense for it to be the developer/administrator's responsibility for putting a suitably robust email infrastructure in place - something that Django already provides extensive configuration options for - rather than Wagtail trying to push one particular approach. Perhaps it would be a good idea to add a note about django-celery-email to the [performance section](http://docs.wagtail.io/en/v1.0/advanced_topics/performance.html) of the docs, when we make this change.

davecranwell:
So without celery, would that mean the mailing process would be blocking?

gasman:
I think so, yes - although that's where my knowledge of mail servers gets a bit hazy. It may well be that in the typical setup (SMTP to localhost), it's only blocking for as long as it takes to push the data to the local SMTP server, and the actual sending-across-the-internet bit happens asynchronously.

mgd020:
Created #1973

gasman:
Implemented in e18f877286308a8a172fcd99a4ea79d9f4aa3c14.

PR #1973, opened 2015-11-29
Send individual notification emails per user

mgd020:
Fix for #1559
Added user details to minimise changes of being marked as spam

### DIFF
--- a/wagtail/wagtailadmin/templates/wagtailadmin/notifications/base_notification.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/notifications/base_notification.html
@@ -1,4 +1,4 @@
 {% load i18n %}{% block subject %}{% endblock %}
-{% block greeting %}{% trans "Hello" %} {% user.get_short_name %},{% endblock %}
+{% block greeting %}{% trans "Hello" %} {{ user.get_short_name }},{% endblock %}
 {% block notification %}{% endblock %}
 {% trans "Edit your notification preferences here:" %} {{ settings.BASE_URL }}{% url 'wagtailadmin_account_notification_preferences' %}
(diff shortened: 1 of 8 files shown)


tomdyson:
@timheap could you review this, please?

wagtail/wagtailadmin/templates/wagtailadmin/notifications/base_notification.html
```diff
@@ -1,3 +1,4 @@
-{% load i18n %}{% block notification %}{% endblock %}
-
+{% load i18n %}{% block subject %}{% endblock %}
+{% block greeting %}{% trans "Hello" %} {% if user.first_name %}{{ user.first_name }}{% else %}{{ user.username }}{% endif %},{% endblock %}
```
timheap: This will behave badly for users with a custom User model that does not have a first_name and/or username field. Using an email address instead of a username is a fairly common modification. You can use the `User.get_short_name()` and `User.get_username()` methods instead. They will act the same for the default User model, but custom User models can override them with something more relevant.

wagtail/wagtailadmin/utils.py
```diff
@@ -153,20 +153,23 @@ def send_notification(page_revision_id, notification, excluded_user_id):
...
-    email_subject = rendered_template[0]
-    email_content = '\n'.join(rendered_template[1:])
 
-    # Send email
-    send_mail(email_subject, email_content, email_addresses)
+    # Send emails
+    for recipient in email_recipients:
+
+        # Get email subject and content
+        email_subject, email_content = render_to_string(template, dict(user=recipient, revision=revision, settings=settings)).split('\n', 1)
```
timheap: Nice use of a limited `.split()` here!

While you're making changes in the area, using the `dict(...)` constructor is not consistent code style with the rest of Wagtail, which uses literal `{...}` braces instead.

timheap:
Aside from the above comments, this looks good to me :+1:

wagtail/wagtailadmin/templates/wagtailadmin/notifications/base_notification.html
```diff
@@ -1,3 +1,4 @@
-{% load i18n %}{% block notification %}{% endblock %}
-
+{% load i18n %}{% block subject %}{% endblock %}
+{% block greeting %}{% trans "Hello" %} {% user.get_short_name %},{% endblock %}
```
timheap: You've used the wrong brace pair here, it should be `{{ user.get_short_name }}`

timheap:
Looks good to me, assuming the tests passed! Travis seems to be having a small issue right now.

gasman:
Merged in e18f877286308a8a172fcd99a4ea79d9f4aa3c14

tomdyson:
Thanks, @mgd020 and @timheap!

PR #1972, opened 2015-11-28
Added 'operator' to the API

michaelfillier:
Enabled searching via the api to include the 'operator' key for performing searches using the elasticsearch 'and' operator.

Issue #1921, opened 2015-11-11
Wagtail API should have a browsable root URL


JoshBarr:
It always feels nice when browsing to `/api/v1/` presents a list of the available API endpoints.

Something like this should do it.. mashing up the `get_api_root_view` from rest_framework's  [DefaultRouter](https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/routers.py#L263). 

Is there a reason we're not extending one of the rest_framework router classes? I guess ViewSets + Page model introspection could be an issue, but haven't looked into it too much.

I'm in the middle of a project, but will create a PR when I get a moment if you think it's a good idea. 

```python
from collections import OrderedDict
from django.conf.urls import url, include
from django.core.urlresolvers import NoReverseMatch
from rest_framework import views
from rest_framework.reverse import reverse
from rest_framework.response import Response
from wagtail.contrib.wagtailapi.router import WagtailAPIRouter
from wagtail.utils.urlpatterns import decorate_urlpatterns


class WagtailAPIDefaultRouter(WagtailAPIRouter):
    root_view_name = 'api-root'

    def get_api_root_view(self):
        """
        Return a view to use as the API root.
        """
        api_root_dict = OrderedDict()

        # In the DRF, these are usually {basename}-list
        for name, class_ in self._endpoints.items():
            api_root_dict[name] = "{basename}".format(basename=name)

        class APIRoot(views.APIView):
            _ignore_model_permissions = True

            def get(self, request, *args, **kwargs):

                ret = OrderedDict()
                namespace = request.resolver_match.namespace

                for key, url_name in api_root_dict.items():
                    url_name = url_name + ':' + 'listing'

                    if namespace:
                        url_name = namespace + ':' + url_name

                    try:
                        ret[key] = reverse(
                            url_name,
                            args=args,
                            kwargs=kwargs,
                            request=request,
                            format=kwargs.get('format', None)
                        )
                    except NoReverseMatch:
                        # Don't bail out if eg. no list routes exist
                        continue

                return Response(ret)

        return APIRoot.as_view()

    def get_urlpatterns(self):
        urlpatterns = []

        for name, class_ in self._endpoints.items():
            # print name
            pattern = url(
                r'^{}/'.format(name),
                include(class_.get_urlpatterns(), namespace=name)
            )
            urlpatterns.append(pattern)

        decorate_urlpatterns(urlpatterns, self.wrap_view)

[...]

PR #1974, opened 2015-11-30
Add API v2 module

kaedroho:
This pull request starts off the new v2 API. The endpoint is located at /api/v2beta/.

Changes in v2 API so far
 - Multiple page types can be specified in type filter
 - Add ``html_url`` to page ``meta``
 - Show ``download_url`` on document listings
 - Renamed results list to ``items`` (previously copied endpoint name)
 - Added more meta fields (``slug, first_published_at, expires_at``, ``show_in_menus``)
 - Search operators (#1972)

Future changes in v2 API
 - New representations for tags and streamfields
 - Better errors
  - Assign a code for each possible error
  - Include any useful information as separate fields on the message
  - Link to doc explaining the error
 - ``?fields=`` has power over ``id`` and ``meta`` subfields too
 - ``?fields=*`` shortcut to print out all fields
 - More control over summary fields
 - Special ``urlpath`` field that can be used for routing when filtered on (eg: ``?urlpath=/blog/``)
 - API root (#1921)
 - Routing API (a way to go from urlpath to detail view in a single call)
 - Filtering enhancements
  - Allow true/false as filter values for boolean fields (currently, you have to use 1/0)
 - Order and search at same time
 - Search while filtering by tag
 - Types information appended to each response

### DIFF
--- a/wagtail/api/__init__.py
+++ b/wagtail/api/__init__.py
@@ -1,0 +1,1 @@
+default_app_config = 'wagtail.api.apps.WagtailAPIAppConfig'
--- a/wagtail/api/apps.py
+++ b/wagtail/api/apps.py
@@ -1,0 +1,7 @@
+from django.apps import AppConfig
+
+
+class WagtailAPIAppConfig(AppConfig):
+    name = 'wagtail.api'
+    label = 'wagtailapi'
+    verbose_name = "Wagtail API"
--- a/wagtail/api/apps.py
+++ b/wagtail/api/apps.py
@@ -1,7 +1,18 @@
-from django.apps import AppConfig
+from django.apps import AppConfig, apps
+from django.conf import settings
+from django.core.exceptions import ImproperlyConfigured
 
 
 class WagtailAPIAppConfig(AppConfig):
     name = 'wagtail.api'
     label = 'wagtailapi'
     verbose_name = "Wagtail API"
+
+    def ready(self):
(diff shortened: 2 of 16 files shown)


kaedroho:
Rebased onto #2002

lee-kagiso:
Is there an upgrade path planned from v1 to v2?
What happens if we are currently using v1, can we run v2 at the same time?
Is v2 intended to replace v1 outright in a future release/are you dropping support for v1, so if we are stuck on v1 because we can't force consumers to upgrade does that mean that down the line we won't be able to upgrade to a new release?

kaedroho:
Yep, it would be possible to run multiple versions of the API alongside each other in the same site.

The plan is that Wagtail will always support two versions of the API. A "stable" version that will never change and an "unstable" version where new features get added.

This pull request adds a new version "v2beta" which would become the "unstable" version. If getting clients to update is difficult, I don't recommend that you don't update to this before it's stabilised. When it is stabilised, you would be able to run both the old v1* and the new v2 APIs side-by-side giving a window for clients to update.

* When we stabilise a new version of the API, we would move the previous stable version into a pip-installable package so you could continue to use it for as long as you need.

kaedroho:
Github has missed this commit (should be right at the beginning): https://github.com/kaedroho/wagtail/commit/a63d5832d05b1b2bcc47963dd338e0ec889db7bc

PR #1975, opened 2015-11-30
Set encoding of icons.scss to UTF-8

kaedroho:
Should fix scss-lint


kaedroho:
Didn't work :'(

zerolab:
May be worth trying http://code.dblock.org/2011/06/09/compass-invalid-us-ascii-character-xe2.html as a workaround directly in the drone container.

kaedroho:
Thanks, will try that!

PR #1978, opened 2015-11-30
Include non-creatable pages in get_page_models

### LINKED ISSUE #725: Root page ignores parent_page_types restriction
I am able to create pages at the root regardless of what parent_page_types is set to for the child model.

It seems like the expected behavior would be to disallow creation of pages at the root unless the child page model's `parent_page_types` field is either unset or contains the root type (`wagtail.wagtailbase.models.Page`?).

I'm completely new to Wagtail, and pretty rusty with Django, so the likelihood that I'm missing something is high.  

On a related note, more documentation around how to work with the root node may be helpful for first-time site implementers.

gasman:
Second part of the cleanup for #1572 as discussed in https://github.com/torchbox/wagtail/pull/1572#issuecomment-158382266. This builds on #1950, the first part of the cleanup.

We update the `PAGE_MODEL_CLASSES` list (as returned by `get_page_models`) to include non-creatable page models such as `Page` itself. (Leaving these out made sense when these types were referred to as "abstract", but now our terminology has changed to "creatable" it's logical for these to be included.) This means that we can drop the special case in `allowed_subpage_types` that allowed all pages as children of `Page`; this special case was necessary because `Page` never appeared in the default `parent_page_types` list, meaning that we wouldn't be able to create children of root without jumping through hoops.

Note that this is a change of behaviour - pages with an explicit `parent_page_types` that does not include `Page` could previously be added, and now they cannot (i.e. this fixes #725). There was some debate on https://github.com/torchbox/wagtail/issues/725#issuecomment-60009672 about whether `parent_page_types = []` should genuinely mean "cannot be added anywhere" or whether it should be special-cased as "can only be added at the top level" (so that we don't have to write the clunky-and-internals-exposing `parent_page_types = [Page]` for homepages). I've gone for the former, as it involves less "magic" (and if we decide otherwise, I reckon it's going to be cleaner to make that change on top of this one).

Besides the `allowed_subpage_types` logic, there are a couple of other places that use `get_page_models`. I've checked that these are not adversely affected by this change - in most cases, they become *more* logically correct by including `Page` (and other non-creatable types) in the list. However, this uncovered an obscure edge case (i.e. made it slightly less obscure...) in the `replace_text` management command, where the text replacement could end up happening multiple times on the same page. This is now fixed (and in the process, new methods `exact_type` and `not_exact_type` added on `PageQuerySet`).

### DIFF
--- a/docs/reference/pages/queryset_reference.rst
+++ b/docs/reference/pages/queryset_reference.rst
@@ -202,6 +202,24 @@ Reference
 
     .. automethod:: not_type
 
+    .. automethod:: exact_type
+
+        Example:
+
+        .. code-block:: python
+
+            # Find all pages that are of the exact type EventPage
+            event_pages = Page.objects.exact_type(EventPage)
+
+    .. automethod:: not_exact_type
+
+        Example:
+
+        .. code-block:: python
+
+            # Find all pages that are not of the exact type EventPage (but may be a subclass)
+            non_event_pages = Page.objects.not_exact_type(EventPage)
+
     .. automethod:: unpublish
 
         Example:
--- a/wagtail/tests/testapp/migrations/0016_mtibasepage_verbose_name.py
+++ b/wagtail/tests/testapp/migrations/0016_mtibasepage_verbose_name.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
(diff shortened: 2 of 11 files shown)


timheap:
Asside from the comment above, this looks good. A test should be added to ensure users can not create pages through the admin that would otherwise be valid, even if they are in the `allowed_subpage_types` setting.

Issue #790, opened 2014-11-07
Wagtail should have own 404, 403, 413, 500 error pages


davecranwell:
Having just attempted to upload a (too) large doc to a client's site's Document library, I received a basic nginx 413 Requested Entity Too Large error page.

Receiving the error at all is nasty, but the fact it wasn't styled added aesthetic insult to injury.

Indeed if you put a foot wrong in the admin and raise a 404, you get the 404 from the front end of the site. Given that implementers may have several websites in wagtail, each with a different livery, this makes the display of the 'default' site's 404 even more jarring.

Wagtail should have wagtail-branded error pages.

SalahAdDin:
Well guys, anything have a style for the template? i have some page, but, the styles are very bad.

zerolab:
Is https://github.com/sprymix/wagtail/commit/467594ab53ea46963a4f07cfa5b968e45de1feda a good fix for the 413 error issue?

SalahAdDin:
And my templates? are so terrible?

zerolab:
@SalahAdDin they are not terrible. However there needs to be some design/front-end direction. And it is best to submit a pull request as it makes the life of the maintainers a whole lot easier.

Cheers,
Dan

kaedroho:
Quick note for whoever wants to implement this:

It may be possible to make use of some of the refactoring in
https://github.com/torchbox/wagtail/pull/912 to write some code that will
catch all 4xx errors raised within the admin views and display a custom
template.

SalahAdDin:
Ok Zerolab, when i test the templates in production i'll do a pull? i want help with this great project.

gasman:
A 404 error page was implemented in #3705 - as discussed there, we won't extend this to other errors, since in those cases we either have an existing pattern for reporting the error (403 / permission denied), or they indicate a failure of the Wagtail installation itself and cannot sensibly be caught within Wagtail (413 / 500).

PR #1979, opened 2015-11-30
Make the bird nod on hover

tomdyson:
and clean up some line endings.

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
@@ -385,11 +385,11 @@ footer,
             float: none;
             margin: auto;
             display: block;
+            transition: all 0.25s cubic-bezier(0.28, 0.15, 0, 2.1);
         }
 
         img:hover {
             transform: rotate(4deg);
-            transition: all 0.1s ease-in-out 0.1s;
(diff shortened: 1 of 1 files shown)


davecranwell:
You've added the transition to the `:hover` pseudo class, so as soon as the pseudo class is removed (the mouse leaves) the transition property is removed. This means the nod animates into position on mouse enter, but not back out again on mouse leave. If you want the animation to animate in and out, you'll need to move the `transition` to the `img` element a few lines above.

tomdyson:
Thanks, @davecranwell! Updated. This is how it looks now: http://codepen.io/anon/pen/zvgjQO

nealtodd:
Where's the `$ex-parrot: 240deg;` variable? ;)

tomdyson:
@nealtodd yes! Or -120deg for a more realistic tumble: http://codepen.io/anon/pen/MKgLRW

gasman:
We should definitely make use of this variant when we implement #790 :-)

gasman:
On a slightly more mundane note, we should probably add a `wagtail-logo` class on the image and limit the styles to that, so that the spoilsports who have customised it (http://docs.wagtail.io/en/latest/advanced_topics/customisation/branding.html) won't get a jaunty bobbing corporate logo.

PR #1982, opened 2015-12-01
Test scss-lint

kaedroho:
Just changed the environment as per https://github.com/torchbox/wagtail/pull/1975#issuecomment-160924935

Checking that scss-lint now passes


kaedroho:
All working now :)

PR #1976, opened 2015-11-30
Refactor API modules

kaedroho:
Depends on #2002 #1974

This pull request builds on my two earlier PRs: #2002 which moves ``contrib.wagtailapi`` to ``wagtail.api.v1`` and #1974 which adds the new ``wagtail.api.v2`` module. There's a lot of shared code between these two modules so this PR has been created to move some common parts to a new ``wagtail.api.shared`` module.

Issue #1760, opened 2015-09-30
[RFC] Wagtail admin API


kaedroho:
Looking to get some feedback on this new feature that I think would help us to greatly improve the maintainability of parts of the admin interface and may lead to user interface improvements in the future.

The initial proposal can be found here: https://groups.google.com/forum/#!topic/wagtail-developers/dNQ9LEaenyI

## Background

We've recently been discussing refactoring parts of the Wagtail admin interface to give better separation between Python and JS. This will help us to simplify the modal workflow system that is currently very difficult to modify.

We've also discussed making the explorer and search interfaces more JavaScript driven.

## Why an API?

In order to implement these features, we need a place where the new interfaces can fetch data from. I propose creating an REST API for this.

The main advantage of creating an API is that we can drive all these interfaces from a single place, so we can concentrate on creating a single rock-solid API and all the new interfaces will benefit from it. Unit testing APIs is also easier as JSON is much easier to parse and check than HTML.

## Why not "Wagtail API"?

The ``wagtailapi`` module (introduced in 1.0) is focused on serving content out to external clients without admin access. For the admin interface, we need more features such as browsing draft pages and seeing permissions/page state. For the modal-workflow refactor we also need access to snippets, which ``wagtailapi`` doesn't support.

The Admin API will be very similar and share a lot of code with the existing API.

## The API

The API will be heavily based on the existing ``wagtailapi`` module and will include some features that I'd like to see in v2. The new API will not be versioned and may change in backwards-incompatible ways between Wagtail releases. So the API can change to suit needs of new Wagtail features when needed.

To save time I'll just mention the differences from the ``wagtailapi`` module:

**1) Authentication**

Like the rest of the admin interface, the new API will use Django's auth module and session cookie for authentication. The same user permissions that exist in the admin interface will be applied to the admin API.

**2) Draft and private pages**

All pages, including draft and private pages, are visible in the API. There will be three new fields added, "status" (string), "has_page_view_restriction" (boolean) and "has_children" (boolean) so the existing UI can be replicated. All these fields will be in "meta"

**3) Selecting multiple page types** (``wagtailapi`` v2 feature)

The ``?type=`` filter will be able to take a comma-separated list of page types. This is required for the page chooser modal.

**4) Image thumbnails**

A ``thumbnail_url`` field will be added to the images endpoint containing the URL of the ``max-165x165`` rendition.

~~**5) child_of/descendant_of filters take special "root" value representing the root page**~~

[...]

PR #1984, opened 2015-12-01
Admin API

kaedroho:
Implements #1760 
Depends on #2002 #1974 #1976

This pull request does some initial work on the admin API. So far, endpoints for pages, images and docs have been created and all the code for all the endpoints currently lives under ``wagtailadmin/api``. This will all be refactored later on so the various endpoints are implemented in their own modules and hooked in (so images will be implemented in wagtailimages/api, for example).

The following changes have been made on top of the V2 API:
 - Whole API is behind wagtailadmins authentication
 - All pages except root are available
 - "status" and "children" fields have been added to pages
 - Thumbnails (``max-165x165``) have been added to images

### DIFF
--- a/wagtail/api/__init__.py
+++ b/wagtail/api/__init__.py
@@ -1,0 +1,1 @@
+default_app_config = 'wagtail.api.apps.WagtailAPIAppConfig'
--- a/wagtail/api/apps.py
+++ b/wagtail/api/apps.py
@@ -1,0 +1,7 @@
+from django.apps import AppConfig
+
+
+class WagtailAPIAppConfig(AppConfig):
+    name = 'wagtail.api'
+    label = 'wagtailapi'
+    verbose_name = "Wagtail API"
--- a/wagtail/api/apps.py
+++ b/wagtail/api/apps.py
@@ -1,7 +1,18 @@
-from django.apps import AppConfig
+from django.apps import AppConfig, apps
+from django.conf import settings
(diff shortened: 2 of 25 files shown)


wagtail/api/shared/filters.py
```diff
@@ -0,0 +1,150 @@
...
+
+        return queryset
+
+
+class ChildOfFilter(BaseFilterBackend):
+    def filter_queryset(self, request, queryset, view):
+        if 'child_of' in request.GET:
+            try:
+                parent_page_id = int(request.GET['child_of'])
+                assert parent_page_id >= 0
```
prydie: > Assertions should *not* be used to test for failure cases that can occur because of **bad user input** or operating system/environment failures, such as a file not being found.

The [python wiki](https://wiki.python.org/moin/UsingAssertionsEffectively) indicates assertions should not be used in this way. I quite like the solution in this instance but I don't know whether in general it's a coding practice we'd like to avoid.

kaedroho: I think what they're probably getting at here is raising an ``AssertionError`` when there is a more specific error you can use instead.

I quite like using the syntax in this way. It's expressive and we replace the ``AssertionError`` with something more specific on the very next line anyway.

mr-const:
@kaedroho why you decided to close this PR? 
I was looking into implementation of a custom render of a page preview and found this great codebase for admin api. Alas it's closed.

PR #1987, opened 2015-12-02
Nav-submenu required translate3d due to its fixed position in the presence of animation

### LINKED ISSUE #1986: Settings menu can't be opened from page editor
I can't work out why just yet, but this appears to be caused by the existence of `<span class="icon icon-spinner"></span>` in the footer and would explain why this problem is present in the styleguide too. #1432

davecranwell:
Also animation vendor prefixing removed, since autoprefixer handles this already. 

Fixes #1986.


gasman:
Merged in eb4c5a7ce2f11777c3590a842da72dd954e7a0ba

Issue #1384, opened 2015-06-07
Difficult to use wagtail without django permissions architecture


scaryclam:
Hello,

I'm trying to use wagtail in an existing project, to replace the home-rolled CMS solution that we are using.
The difficulty that we are having, is that the site is not _just_ a CMS and requires a more involved permissions structure than Django can provide.

I was wondering if you would be open to accepting a PR that removes the requirement for django permissions, mainly by allowing us to move the permissions decorators into the urls.py file instead of decorating the views themselves. This would allow us to override the behavior with our own urls.py file instead of having to override the views themselves.

tomdyson:
Hi @scaryclam, we're definitely interested in this. There's another Wagtail user who we believe needs something similar.

scaryclam:
That's good to hear :) It's looking more likely that we'll be using wagtail in our project, so I'll pull down your master branch and start seeing what we can do to help out.

scaryclam:
Hi, someone got in touch with me regarding this but unfortunately I had an issue with my email that resulted in losing many of them. I think I have a solution for this that shouldn't be too intrusive. I'll be trying to make a PR in the next few days, unless I find that this issue is deprecated.

gasman:
It occurs to me that the current work on image/document collections (https://groups.google.com/d/msg/wagtail-developers/kE-aDrOIzRI/Cc2CP3KbAwAJ, https://github.com/torchbox/wagtail/wiki/Collections-RFC) is going to complicate this a bit. Currently page editing *doesn't* use simple `permission_required` decorators, because Wagtail has its own permission model for pages that's more fine-grained than Django's - and when collections are implemented, images and documents will be in the same situation.

As a result, overriding / disabling the permission model is not going to be a simple case of moving `permission_required` decorators out of views and onto URLs - because fewer and fewer areas of Wagtail admin will be using those decorators.

Instead, I think we'd have to define a swappable Permissions module to accompany each of the models we're interested in (Page, Image, Document and so on) - this module would implement a standard API that can be queried to answer questions like:

* Is user X allowed to edit document Y?
* Which collections does user X have 'add_image' permission on?
* Which users are allowed to edit documents in collection Z?
* Does user X have any image-related permissions at all?

(For pages, we have a couple of classes in `wagtailcore.models` which do this sort of task - `UserPagePermissionsProxy` and `PagePermissionTester` - but these may well need to be rewritten if we're going to generalise this to other models.)

[...]

PR #1989, opened 2015-12-03
Deprecate admin permissions helper; move to Page model

jimr:
Basically, migrating `wagtail.wagtailadmin.utils.users_with_page_permission` to `wagtail.wagtailcore.models.PageBase. get_users_with_permission`. Doing so makes it possible to override the default behaviour on `Page` subclasses.

Before merging, some discussion will be required on how this should be made consistent with `Page.permissions_for_user` - it seems the admin helper used some simplifying assumptions. Though it would be far slower, it seems that in order to get all users with an appropriate permission, we must fetch all users and test their permissions with `permissions_for_user`. Something like

```python
def get_users_with_permission(self, permission, include_superusers=True):
    return [
        upp.user for upp in filter(
            lambda upp: permission in upp.permissions,
            self.permissions_for_user(user) for user in get_user_model().objects.all()
        )
    ]
```

In order to retain the `include_superusers` behaviour, the `PagePermissionsTester` would require some refactoring.

Something else to consider is the deprecation warning. If you're happy to deprecate the old helper, I don't know what the policy on which 'version' should be used for the warning (1.5?). The original helper wasn't used anywhere in the Wagtail code outside of the `utils.py` file,

### DIFF
--- a/wagtail/wagtailadmin/utils.py
+++ b/wagtail/wagtailadmin/utils.py
@@ -1,16 +1,16 @@
+import warnings
+
 from functools import wraps
 
 from django.template.loader import render_to_string
 from django.core.mail import send_mail as django_send_mail
 from django.conf import settings
-from django.contrib.auth import get_user_model
-from django.db.models import Q
 from django.shortcuts import redirect
 from django.utils.translation import ugettext as _
 
 from modelcluster.fields import ParentalKey
 
-from wagtail.wagtailcore.models import Page, PageRevision, GroupPagePermission
+from wagtail.wagtailcore.models import Page, PageRevision
 from wagtail.wagtailusers.models import UserProfile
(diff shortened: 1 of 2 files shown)


gasman:
Cheers @jimr! Will give this some more thought when I get chance, but my initial feeling is that we want to move towards permission logic being *less* coupled to the core models, not more - see #1384.

jimr:
Ah, yep, I hadn't seen that, looks sensible. Our requirement is simply to have custom permissions around pages, so provided the mechanism is uniform and flexible, we'll take it :) I the meantime is there any merit in shifting this functionality around as above? If not, go ahead and close if you like.

Our immediate requirement was to prevent all moderators being spammed with an email whenever a page is submitted for moderation - in our implementation, we want everyone to have moderation permissions, but for specific moderators to be selected for each page.

PR #1990, opened 2015-12-03
Image upload into listings, this time with better commits

davecranwell:


gasman:
As per discussions at team meetings, we're not prepared to lose the ability to edit images at the point of upload. @davecranwell to consider UI to support this.

PR #1991, opened 2015-12-03
Limit logo animation CSS so that custom logos don't nod

gasman:
as per https://github.com/torchbox/wagtail/pull/1979#issuecomment-161603586

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
@@ -385,10 +385,13 @@ footer,
             float: none;
             margin: auto;
             display: block;
+        }
+
+        img.wagtail-logo {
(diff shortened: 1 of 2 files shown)


davecranwell:
I think our preferred specificity order is `class.class:pseudo-class` so please can it be  img.wagtail-logo:hover`. Other than that, fine.

gasman:
Merged with order tweaked in ad7596a96daf8a3fb4dbaeb387aa9bc4a0667be8.

MechanisM:
I think that wagtail bird should wag the tail instead of doing a nod =)

tomdyson:
I agree! Who knows about SVG animation?

davecranwell:
Something like this? http://codepen.io/davecranwell/pen/obgXaL

MechanisM:
@davecranwell yeah!

timorieber:
Nice idea!

PR #1993, opened 2015-12-03
Email notifications use page permission helpers

jimr:
Sorry for the PR spam :)

The original implementation of this helper didn't necessarily agree with what
`page.permissions_for_user` would tell you for a given user, so this
code has been updated to fix that. The compromise is that
`include_superuser` can no longer be set to `False` and it'll be a bit
slower, but given that this helper is only used in one place, hopefully
that's not an issue.

Not sure if raising a `NotImplementedError` is the way to go, or whether
we should just strip the parameter.

Replaces #1989, related to #1384.

### DIFF
--- a/wagtail/wagtailadmin/utils.py
+++ b/wagtail/wagtailadmin/utils.py
@@ -52,16 +52,30 @@ def get_object_usage(obj):
 
 def users_with_page_permission(page, permission_type, include_superusers=True):
     # Get user model
+    if not include_superusers:
+        raise NotImplementedError("This functionality is no longer supported")
+
     User = get_user_model()
 
-    # Find GroupPagePermission records of the given type that apply to this page or an ancestor
-    ancestors_and_self = list(page.get_ancestors()) + [page]
-    perm = GroupPagePermission.objects.filter(permission_type=permission_type, page__in=ancestors_and_self)
-    q = Q(groups__page_permissions=perm)
+    # Get all page permissions for all users (this is going to be slow if you
+    # have lots of users). Note that the `permissions` attribute isn't set on
+    # `PagePermissionTester` if the user is a superuser or not active, so we
+    # pre-filter out such users to avoid attribtue errors. We OR superusers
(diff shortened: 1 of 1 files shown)


wagtail/wagtailadmin/utils.py
```diff
@@ -52,16 +52,30 @@ def get_object_usage(obj):
...
-    perm = GroupPagePermission.objects.filter(permission_type=permission_type, page__in=ancestors_and_self)
-    q = Q(groups__page_permissions=perm)
+    # Get all page permissions for all users (this is going to be slow if you
+    # have lots of users). Note that the `permissions` attribute isn't set on
+    # `PagePermissionTester` if the user is a superuser or not active, so we
+    # pre-filter out such users to avoid attribtue errors. We OR superusers
+    # back in later, and we don't care about inactive users.
+    user_page_permissions = [
+        page.permissions_for_user(user)
+        for user in User.objects.filter(is_superuser=False, is_active=True)
```
kaedroho: Would it be possible to make this query not match frontend users? (we run a Wagtail site which currently has over 16000 active users which don't have access to the admin)

gasman: I don't think that's a good idea - the only thing that distinguishes front-end and back-end users is the 'can access Wagtail admin' permission, and I'd rather not have anything depend on that at the model level. (I can easily imagine us dealing with editor permissions that don't go through the admin in the near future - through the API, say.)

Overall I'm not really keen on the approach taken by this PR - I don't think it's possible to efficiently perform forward ("what permission does user X have") and reverse ("which users have permission Y") permission checks with the same logic. I think if we're going for the 'permission module' approach as per #1384, we should have both the forward and reverse logic in that one module - and a developer who wants to customise the permission logic would implement both of those methods.

gasman:
Closing in favour of the permission policies model started in #2102 (which doesn't cover page permissions yet, but is the approach we intend to follow).

PR #1994, opened 2015-12-03
Convert to use of SASS partials

davecranwell:
SASS compilers already understand a difference between source files which are run in the browser and those that are only included within other files, which it calls "Partials" (http://sass-lang.com/guide#topic-4). 

Once the partials are identified as such with the underscore prefix, it's not necessary to manually white-list specific scss files to compile since the only remaining files in a given app will by default be the ones you want to compile.


SalahAdDin:
:+1:

kaedroho:
Merged in 88341b782b753a484ebc8e60a8fe311319951c92 + parent

PR #1997, opened 2015-12-03
typo in a locale

Tamriel:


gasman:
Hi @Tamriel - thanks for the correction! We manage our translations through [Transifex](https://www.transifex.com/projects/p/wagtail/) - I've now updated it there, so it will be fixed when we next pull the translation data. If you find any other things in future that need fixing, please submit them through Transifex.

Tamriel:
Alright :)
Thanks for the fast responses!

PR #2003, opened 2015-12-05
Django 1.9 support

AliLozano:


alexgleason:
Hi @AliLozano, thanks for the PR!

Can anyone explain why the Jinja2 imports are now wrapped in a try-catch block? Curious to know how/why that works.

AliLozano:
Because Django 1.9 import all templatetags, and when Django tries import jinja, it fails.

zerolab:
Hi @AliLozano and thank you for the PR.

Have you seen #1816? @kaedroho added Django 1.9 to tox and fixed some texts.
I am inclined to mark this as duplicate of #1816, but will wait to see what the maintainers say.

AliLozano:
Oh, I didn't see the other PR. Apologies, mark this as duplicated, The other PR has more reviews and changes, I think that it is more complete.

gasman:
Hi @AliLozano - this is really useful, as it lets us compare different approaches, and catch fixes that the other PR might have missed.

I like the `Extension = object` trick, as it means we don't have to move the `jinja2` modules outside of `templatetags` - @timheap @kaedroho, any downsides to this approach? (I guess it might result in misleading error messages if people enable Jinja2 but don't have the package installed - is that something people are likely to do?)

kaedroho:
I actually prefer having them separate. There's convention to keep your jinja2 templates completely separate from your Django ones at the app level and I think it's neater to apply this to template tags as well. This also reduces the risk of obscure error messages when using Django 1.9/jinja2.

But perhaps we could use a similar trick to allow backwards compatibility for a couple of releases.

AliLozano:
Well, if you  like the extension=object fix, instead object class, we can use a ExtensionNoJinja class with a especific error in __init__ to warn the user when use the templatetag and forget install jinja.

Django Extensions uses conditionals: https://github.com/django-extensions/django-extensions/issues/705

gasman:
The consensus seems to be for simply moving the jinja2 modules rather than tweaking the imports, so I'll close this in favour of #1816 now - but thanks for providing a second perspective!

PR #2005, opened 2015-12-07
Some cleanup in Elasticsearch backend

kaedroho:
This pull request does some cleanup to the Elasticsearch search backend. Not public APIs affected. The main aims are to improve consistency and to get the code ready for adding Elasticsearch 2.0 support.

PR #1491, opened 2015-07-08
Added support for indexing relations in Elasticsearch

kaedroho:
This pull request implements the new syntax as descibed in #459 which allows related objects to be indexed in Elasticsearch.

This is useful for indexing tags, "child objects", foreign keys, etc. A relation can be indexed from both sides.

It uses Elasticsearch's "nested type": https://www.elastic.co/guide/en/elasticsearch/reference/1.6/mapping-nested-type.html

This doesn't implement the ability to filter search results across relationships. ``FilterField``s can be indexed but you'll need to come up with a custom query to use them for the moment. This will be implemented in a later pull request.

PR #2006, opened 2015-12-08
[WIP] Rename search "backends" to "indices"

kaedroho:
Depends on #1491 #2005

This pull request fixes some of the terminology used in wagtail search. Here's the new definitions:
 - "index" - A database to index documents into and run search queries against
 - "backend" - A module that provides access to indices for a particular DBMS (eg elasticsearch, postgres)

Currently, both of the above are referred to as "backend". The main thing we do here is rewrite some APIs to refer to indices instead of backends.

The following APIs will be broken (backwards compatibility for both will be provided):
 - ``WAGTAILSEARCH_BACKENDS`` will be renamed to ``WAGTAILSEARCH_INDICES``
 - The ``get_search_backend(s)`` functions will be renamed to ``get_search_ind(ex|ices)``

### DIFF
--- a/wagtail/wagtailsearch/backends/base.py
+++ b/wagtail/wagtailsearch/backends/base.py
@@ -186,7 +186,7 @@ class BaseSearchResults(object):
         return repr(data)
 
 
-class BaseSearch(object):
+class BaseSearchIndex(object):
     query_class = None
     results_class = None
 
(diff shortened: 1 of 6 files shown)


kaedroho:
Closing this as I've changed my mind!

To support Elasticsearch 2.0 and indexing individual models separately, Wagtail may need to use multiple indices. I think it would be very tedious to expect a developer to add each index Wagtail needs to their search configuration.

A search ``backend`` represents a particular instance of Elasticsearch where Wagtail can create indices and Wagtail may create as many indices it needs.

PR #2007, opened 2015-12-08
Added title validation for image upload form

### LINKED ISSUE #92: Validation error on image upload causes file field to be emptied
When you go to add an image and no title is given, the image selected is lost and needs to be selected again. 

Would be useful to not have to do that and just give the title again.

serzans:
Fixes #92. Submit button continues executing javascript when incomplete form is submitted.

### DIFF
--- a/wagtail/wagtailimages/templates/wagtailimages/chooser/chooser.js
+++ b/wagtail/wagtailimages/templates/wagtailimages/chooser/chooser.js
@@ -46,20 +46,26 @@ function(modal) {
 
     ajaxifyLinks(modal.body);
 
+
     $('form.image-upload', modal.body).submit(function() {
         var formdata = new FormData(this);
-
-        $.ajax({
-            url: this.action,
-            data: formdata,
-            processData: false,
-            contentType: false,
-            type: 'POST',
-            dataType: 'text',
-            success: function(response){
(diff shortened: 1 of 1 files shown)


kaedroho:
Could you rebase this on the upstream master branch? This looks like it's been based on your local master which has been used for the pep8 changes. You might want to move those pep8 changes to a separate branch as well to keep your local master clean.

gasman:
To expand on the issue with the submit button here - the problem is that our ['debounce long-running form submissions'](https://github.com/torchbox/wagtail/blob/f0df1cb1fe9103929cf0afc4f16dd94bf3689658/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js#L160) handler kicks in as soon as the button is clicked, and we don't have any way to block that process (or return the button to its original state) if the form submission doesn't actually go ahead (e.g. a validation error, as in this case).

I can't see a good way of dealing with this in the current setup (because the onclick handler doesn't know anything about the form submit handler, and vice versa) - we'll probably have to rewrite the 'debounce' code to be more closely integrated into form submissions. (Which is probably a good thing anyhow - it means that you don't bypass the debounce code if you submit the form by pressing enter.)

gasman:
Superseded by #2538

Issue #1578, opened 2015-08-03
verbose_name on models/fields should be lower-cased


gasman:
The convention for verbose_name is for the name to not be capitalised:

https://docs.djangoproject.com/en/1.8/topics/db/models/#verbose-field-names
https://docs.djangoproject.com/en/1.8/topics/i18n/translation/#model-fields-and-relationships-verbose-name-and-help-text-option-values

Our verbose_names are currently capitalised, resulting in formatting issues in places like permission names, e.g. "can add Search promotion".

gasman:
Fixed in 32b41247679100058eda70badc4e916356353a7e

PR #2009, opened 2015-12-08
#1578 Removed capitalisation from verbose names where required

serzans:


kaedroho:
As with #2007, could you removed the pep8 validation from this as well too?

gasman:
@kaedroho In this case I think it's actually better for this PR to be on top of the PEP8 fixes, because they both cover large chunks of the codebase and would result in massive edit conflicts otherwise.

gasman:
Merged in 32b41247679100058eda70badc4e916356353a7e

PR #2011, opened 2015-12-08
pep8 fixes

c16192:
Made pep8 fixes to Wagtail/wagtailforms, wagtailsearch, wagtailsites, wagtailredirects, wagtailimages, wagtailusers, wagtailsnippets.

### DIFF
--- a/wagtail/wagtailimages/migrations/0004_make_focal_point_key_not_nullable.py
+++ b/wagtail/wagtailimages/migrations/0004_make_focal_point_key_not_nullable.py
@@ -9,8 +9,9 @@ def remove_duplicate_renditions(apps, schema_editor):
         schema_editor.execute("""
             DELETE FROM `wagtailimages_rendition` WHERE CONCAT(image_id, '-', filter_id) IN (
                 SELECT CONCAT(image_id, '-', filter_id)
-                FROM (SELECT * FROM `wagtailimages_rendition`) as x WHERE `focal_point_key` IS NULL GROUP BY image_id,
-                filter_id HAVING COUNT(*) > 1
+                FROM (SELECT * FROM `wagtailimages_rendition`) as x
+                WHERE `focal_point_key` IS NULL
+                GROUP BY image_id, filter_id HAVING COUNT(*) > 1
             ) AND `focal_point_key` IS NULL
         """)
     elif schema_editor.connection.vendor == 'microsoft':
--- a/wagtail/wagtailimages/migrations/0004_make_focal_point_key_not_nullable.py
+++ b/wagtail/wagtailimages/migrations/0004_make_focal_point_key_not_nullable.py
@@ -8,26 +8,34 @@ def remove_duplicate_renditions(apps, schema_editor):
     if schema_editor.connection.vendor == 'mysql':
         schema_editor.execute("""
             DELETE FROM `wagtailimages_rendition` WHERE CONCAT(image_id, '-', filter_id) IN (
-                SELECT CONCAT(image_id, '-', filter_id) FROM (SELECT * FROM `wagtailimages_rendition`) as x WHERE `focal_point_key` IS NULL GROUP BY image_id, filter_id HAVING COUNT(*) > 1
+                SELECT CONCAT(image_id, '-', filter_id)
+                FROM (SELECT * FROM `wagtailimages_rendition`) as x WHERE `focal_point_key` IS NULL GROUP BY image_id,
+                filter_id HAVING COUNT(*) > 1
             ) AND `focal_point_key` IS NULL
         """)
     elif schema_editor.connection.vendor == 'microsoft':
         schema_editor.execute("""
-            DELETE FROM [wagtailimages_rendition] WHERE CAST(image_id AS VARCHAR(MAX)) + '-' + CAST(filter_id AS VARCHAR(MAX)) IN (
-                SELECT CAST(image_id AS VARCHAR(MAX)) + '-' + CAST(filter_id AS VARCHAR(MAX)) FROM [wagtailimages_rendition] WHERE focal_point_key IS NULL GROUP BY image_id, filter_id HAVING COUNT(*) > 1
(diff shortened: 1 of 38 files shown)


wagtail/wagtailimages/migrations/0004_make_focal_point_key_not_nullable.py
```diff
@@ -8,26 +8,34 @@ def remove_duplicate_renditions(apps, schema_editor):
     if schema_editor.connection.vendor == 'mysql':
         schema_editor.execute("""
             DELETE FROM `wagtailimages_rendition` WHERE CONCAT(image_id, '-', filter_id) IN (
-                SELECT CONCAT(image_id, '-', filter_id) FROM (SELECT * FROM `wagtailimages_rendition`) as x WHERE `focal_point_key` IS NULL GROUP BY image_id, filter_id HAVING COUNT(*) > 1
+                SELECT CONCAT(image_id, '-', filter_id)
+                FROM (SELECT * FROM `wagtailimages_rendition`) as x WHERE `focal_point_key` IS NULL GROUP BY image_id,
```
kaedroho: Might be a little neater to break the line just before the WHERE and GROUP BY keywords

wagtail/wagtailimages/models.py
```diff
@@ -254,7 +258,9 @@ def get_rendition(self, filter):
             if cache_key:
                 output_extension = cache_key + '.' + output_extension
 
-            output_filename_without_extension = input_filename_without_extension[:(59 - len(output_extension))]  # Truncate filename to prevent it going over 60 chars
+            output_filename_without_extension = input_filename_without_extension[
```
kaedroho: Instead of breaking the code, I think it would be neater to move the comment to the previous line here:
```python
# Truncate filename to prevent it going over 60 chars
output_filename_without_extension = input_filename_without_extension[:(59 - len(output_extension))]
```

wagtail/wagtailimages/models.py
```diff
@@ -338,7 +344,10 @@ def get_image_model():
 
     image_model = apps.get_model(app_label, model_name)
     if image_model is None:
-        raise ImproperlyConfigured("WAGTAILIMAGES_IMAGE_MODEL refers to model '%s' that has not been installed" % settings.WAGTAILIMAGES_IMAGE_MODEL)
+        raise ImproperlyConfigured(
+            "WAGTAILIMAGES_IMAGE_MODEL refers to model '%s' that has not been installed"
+            % settings.WAGTAILIMAGES_IMAGE_MODEL
```
kaedroho: The % operator should be on the previous line according to https://www.python.org/dev/peps/pep-0008/#maximum-line-length "The preferred place to break around a binary operator is after the operator, not before it"

wagtail/wagtailredirects/tests.py
```diff
@@ -16,26 +16,62 @@ def test_path_normalisation(self):
...
-        self.assertEqual(path, normalise_path('Hello/world.html;fizz=three;buzz=five?foo=Bar&Baz=quux2'))  # Leading slash can be omitted
-        self.assertEqual(path, normalise_path('Hello/world.html/;fizz=three;buzz=five?foo=Bar&Baz=quux2'))  # Trailing slashes are ignored
-        self.assertEqual(path, normalise_path('/Hello/world.html;fizz=three;buzz=five?foo=Bar&Baz=quux2#cool'))  # Fragments are ignored
-        self.assertEqual(path, normalise_path('/Hello/world.html;fizz=three;buzz=five?Baz=quux2&foo=Bar'))  # Order of query string parameters is ignored
-        self.assertEqual(path, normalise_path('/Hello/world.html;buzz=five;fizz=three?foo=Bar&Baz=quux2'))  # Order of parameters is ignored
-        self.assertEqual(path, normalise_path('  /Hello/world.html;fizz=three;buzz=five?foo=Bar&Baz=quux2'))  # Leading whitespace
-        self.assertEqual(path, normalise_path('/Hello/world.html;fizz=three;buzz=five?foo=Bar&Baz=quux2  '))  # Trailing whitespace
+        self.assertEqual(path, normalise_path(  # The exact same URL
```
kaedroho: I'm happy with this but perhaps this code could be made even clearer by putting the URLs to test against in a list and do the assertion in a for loop.

No need to do that in this PR though

wagtail/wagtailredirects/views.py
```diff
@@ -12,7 +12,9 @@
 from wagtail.wagtailredirects.forms import RedirectForm
 
 
-@any_permission_required('wagtailredirects.add_redirect', 'wagtailredirects.change_redirect', 'wagtailredirects.delete_redirect')
+@any_permission_required(
+    'wagtailredirects.add_redirect', 'wagtailredirects.change_redirect', 'wagtailredirects.delete_redirect'
```
kaedroho: I think this would look a lot clearer if they were each on their own line (since they share a common prefix). No need to change it again though if you don't want to.

wagtail/wagtailsnippets/tests.py
```diff
@@ -9,7 +9,8 @@
 
 from wagtail.tests.utils import WagtailTestUtils
 from wagtail.tests.testapp.models import Advert, SnippetChooserModel
-from wagtail.tests.snippets.models import AlphaSnippet, ZuluSnippet, RegisterDecorator, RegisterFunction, SearchableSnippet
+from wagtail.tests.snippets.models \
```
kaedroho: The way we usually break an import like this to multiple lines is to use brackets:

```python
from wagtail.tests.snippets.models import (
    AlphaSnippet, ZuluSnippet, RegisterDecorator, RegisterFunction, SearchableSnippet
)
```

kaedroho:
Good work, thanks! Only minor, non-essential tweaks suggested above.

gasman:
Merged in 4433d4f3c086714f3817d5413c2c436dd3d95558 - thanks!

PR #2015, opened 2015-12-08
Use plain models, not ContentTypes to manage snippets

timheap:
Using ContentTypes was unnecessary, and obfuscated the code. Removing
them and using plain models makes the code easier to understand.

The `snippet_type` argument to the SnippetChooserPanel has been
deprecated as part of this. It was already unnecessary, with the snippet
type being detected from the model field.

This is in the same spirit as #1950.


timheap:
Oops, missed a test failure. Fixed now.

kaedroho:
This is really great work, thanks!

Merged in 909319a8559f77165e4daa0681754757a11c07d1

PR #2026, opened 2015-12-10
Decouple 'popular tags' filtering on image chooser from search

gasman:
Workaround for https://github.com/torchbox/wagtail/issues/278#issuecomment-157817200.

Searching by tag only works under Elasticsearch, so clicking on 'popular tags' is broken
in the default install. As a workaround, add a new 'tag' parameter to the chooser view
that allows us to filter on tags using plain SQL lookup, as a separate operation to search.


kaedroho:
Merged in https://github.com/torchbox/wagtail/commit/635659a80fb1d6fc98aec3c87aa5c53fce8c22d2 + parent

PR #2027, opened 2015-12-11
Move shared URL-building logic into a get_url_parts method; implement Page.get_site

### LINKED ISSUE #693: How do I access the domain of a page or the absolute URL of an image?
There are occasions where it's necessary to generate absolute URLs, or at least determine the domain of a certain page.

For example with OpenGraph/Schema.org tags, the canonical URL of a page is necessary. This is possible with `page.full_url()`. However OpenGraph also requires an absolute URL for the `og:image` meta tag i.e an absolute URL for an image. How would that be generated? Perhaps, given that the image library is domain-agnostic, it would be better if there were an option to get the domain on it's own, of any page viewed, so it could be prefixed to a relative image URL?

It feels like there's a need for a `page.get_site()` (returning a `Site`) or similar, to make access easier for the non-technical.

gasman:
The `url`, `relative_url` and `full_url` methods of Page all duplicated the logic for matching a page to a site and constructing a relative page path - meaning that a page model with custom URL routing (e.g. blog posts with date-based URLs) would have to override all three of them. This PR moves the shared logic to a new `get_url_parts` method, which returns a tuple of (site_id, site_root_url, relative_page_path); subclasses now only have to override this one method.

In addition, it implements a `get_site` method as requested in #693 (which we would otherwise have to do as a fourth copy of the same logic...)

Fixes #693

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -684,12 +684,29 @@ class Page(six.with_metaclass(PageBase, MP_Node, ClusterableModel, index.Indexed
         """
         return (not self.is_leaf()) or self.depth == 2
 
+    def get_url_parts(self):
+        """
+        Determine the URL for this page and return it as a tuple of
+        (site_id, site_root_url, page_url_relative_to_site_root).
+        Return None if the page is not routable.
+
+        Pages with custom URL routing should override this.
+        """
+        for (site_id, root_path, root_url) in Site.get_site_root_paths():
+            if self.url_path.startswith(root_path):
+                page_path = reverse('wagtail_serve', args=(self.url_path[len(root_path):],))
+                return (site_id, root_url, page_path)
+
     @property
     def full_url(self):
         """Return the full URL (including protocol / domain) to this page, or None if it is not routable"""
-        for (id, root_path, root_url) in Site.get_site_root_paths():
-            if self.url_path.startswith(root_path):
-                return root_url + reverse('wagtail_serve', args=(self.url_path[len(root_path):],))
+        try:
+            site_id, root_url, page_path = self.get_url_parts()
+        except TypeError:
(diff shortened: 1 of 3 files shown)


kaedroho:
Would be nice to add this new method to the model reference docs too: http://docs.wagtail.io/en/v1.2/reference/pages/model_reference.html

Also, if you consider ``get_url_parts`` to be public API, please add tests/docs for that too. If not, perhaps prefix with an undercore to indicate this?

wagtail/wagtailcore/models.py
```diff
@@ -625,12 +625,29 @@ def is_navigable(self):
...
+
     @property
     def full_url(self):
         """Return the full URL (including protocol / domain) to this page, or None if it is not routable"""
-        for (id, root_path, root_url) in Site.get_site_root_paths():
-            if self.url_path.startswith(root_path):
-                return root_url + reverse('wagtail_serve', args=(self.url_path[len(root_path):],))
+        try:
+            site_id, root_url, page_path = self.get_url_parts()
+        except TypeError:
```
carlmjohnson: My fear is that this will cover up something going wrong in the call. Why not,

```python
url_parts = self.get_url_parts()

# get_url_parts returned None; page is not routable
if not url_parts:
    return

site_id, root_url, page_path = url_parts
return root_url + page_path
```
?

gasman: Good point! I was trying to follow the Python 'EAFP' principle, but that often has the risk of catching more exceptions than you intended, as it does here.

gasman:
Rebased and updated (docs / tests added for `get_url_parts`, non-routable page checks changed to avoid catching unrelated `TypeError`s) - ready for review again.

kaedroho:
Merged in 7861c2b4929b08b46978e8a4e187fcc3ca92c972 +parents

PR #2028, opened 2015-12-11
Wordpress Import Plugin

serzans:
Me and Shu made some progress with the import plugin. The code was able to retrieve html from a Wordpress XML file and put it into a Wagtail page (StandardPage). It also extracted images, putting them into the Wagtail image menu. In the end, we were not able to change the source of images in the imported html. We adapted the code from the Wordpress to Puput plugin. https://github.com/APSL/wordpress-to-puput


auvipy:
good thinking

chimeno:
I see this quite interesting, but IMHO, I would go with a plugin approach instead with a contrib app.

gasman:
Closing as per https://github.com/torchbox/wagtail/issues/387#issuecomment-221601800

PR #2030, opened 2015-12-14
Add missing blocktrans to InlinePanel's add-button

jnns:
Right now, one can set the label of a model to a localized string but the add button still displays the string "Add <translated label>" nonetheless.


gasman:
Hi @jnns - thanks for this. It looks like you've made a typo in the opening `blocktrans` tag - please can you fix? I think it would also be better to move the `self.heading|lower` logic so that it's not inside the translation string: `{% blocktrans with heading=self.heading|lower %}`

jnns:
Sorry @gasman, I was in a hurry. I fixed the typo and cleaned the `blocktrans` statement as per your suggestion.

gasman:
Merged in 42e1317a4eb9c9996eb21f1c3c4abaf239f7e2aa - thanks!

PR #2031, opened 2015-12-14
Willow 0.3 support

kaedroho:
This pull request adds support for Willow 0.3 into Wagtail.

Changes include:
 - Willow images are now immutable (all image modification operations return a new image)
 - Auto orientation has been added
 - ``format_name`` attribute added (neatens up a few things in Wagtail)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ install_requires = [
     "beautifulsoup4>=4.3.2",
     "html5lib>=0.999,<1",
     "Unidecode>=0.04.14",
-    "Willow>=0.2.2,<0.3",
+    "Willow>=0.3,<0.4",
 ]
 
 
--- a/tox.ini
+++ b/tox.ini
@@ -34,7 +34,7 @@ deps =
     python-dateutil==2.2
     pytz==2014.7
     Embedly
-    Willow==0.2
+    Willow==0.3b4
     jinja2==2.8
     coverage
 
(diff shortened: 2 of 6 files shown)


miracle2k:
Wagtail installed Willow>=0.3 automatically for me. Took me quite a while to debug what was going on.

tomdyson:
Hi @timheap, do you have time to have a look at this?

timheap:
`wagtail.wagtailimages.image_operations.DoNothingOperation.run` should `return willow` now, instead of `pass`: https://github.com/torchbox/wagtail/pull/2031/files#diff-73c7cc5fb6b58a0b40c21320e3b860b1R37

timheap:
I see that `ImageOperation.run()` can return None and that operation will just be ignored. Still, `pass` seems to be less explicit than `return willow`, and can lead to confusion such as this.

timheap:
Looks good to me!

kaedroho:
@miracle2k

I've just hit that issue on one of our sites. We were using pip 1.1 on the affected server but the staging site, which uses pip 7.0, was not affected. If anyone else gets that issue, upgrade pip (``pip install --upgrade pip``) and install Willow 0.2.2 (``pip install Willow==0.2.2``).

gasman:
Merged in ef4aa3b54e35fd24ddf6e834031e89948fc8ac5c + parents. I've tweaked the dependency line in both setup.py and tox.ini to `Willow>=0.3b4,<0.4` so that we have some wiggle room re co-ordinating the Wagtail 1.4 and Willow 0.3 releases.

PR #2036, opened 2015-12-15
Workaround issue where the image listing view goes blank if one image failed to render

### LINKED ISSUE #1805: Image listing disappears when one of the images cannot resize
As discussed in: https://groups.google.com/forum/#!topic/wagtail/CQ-S338XDvs

When an image fails to resize on a listing, the listing just disappears without error.

kaedroho:
This replaces #2025. I've added a comment to the new template explaining its purpose

Fixes #1805


gasman:
Looks good to me, except that you don't need to import `wagtailadmin_tags` or `i18n` on the included template.

PR #2041, opened 2015-12-15
Add system check to confirm that compiled CSS files are present

### LINKED ISSUE #1720: Add system check for missing compiled static files
Currently, attempting to run a Git checkout of Wagtail without running the `npm run build` step results in `UncompressibleFileError`. Add a system check to produce a more informative error message.

gasman:
Fixes #1720. Thanks to Shu Ishida @c16192 for initial development on this feature.

### DIFF
--- a/wagtail/wagtailadmin/apps.py
+++ b/wagtail/wagtailadmin/apps.py
@@ -1,5 +1,7 @@
 from django.apps import AppConfig
 
+from . import checks  # NOQA
+
 
 class WagtailAdminAppConfig(AppConfig):
     name = 'wagtail.wagtailadmin'
--- a/wagtail/wagtailadmin/checks.py
+++ b/wagtail/wagtailadmin/checks.py
@@ -1,0 +1,30 @@
+import os
+
+from django.core.checks import Error, register
+
+
+@register()
+def css_install_check(app_configs, **kwargs):
+    errors = []
+
+    css_path = os.path.join(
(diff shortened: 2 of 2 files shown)


gasman:
Looks like this is failing on Django 1.7 (only) on Travis because:

* our Travis setup doesn't build the CSS
* Django 1.7 runs system checks on `manage.py test` (and therefore fails this check), but Django 1.8 and above apparently don't
* The unit tests themselves happily run without the CSS and aren't affected by the usual django-compressor error, because our test settings have `COMPRESS_ENABLED = False`

Not sure what the nicest way to fix this is. Options:

* Get Travis to build the CSS (eww)
* Skip the check if `COMPRESS_ENABLED = False`
* Add a new `SKIP_CSS_CHECK` setting
* Add some "is this the test runner?" logic to the system check (isn't there a helper function for this?)
* Wait until we drop Django 1.7 support, and sweep the whole problem under the rug :-)

timheap:
You could also add a `checks.Warning` instead of `checks.Error`. It will fix the issue for Django 1.7, while still warning developers. It wont prevent Django from starting up, unlike `checks.Error`, which is either good or bad depending on how you look at it.

gasman:
@timheap Good idea! Changed to a warning now.

kaedroho:
Merged in 53fb42f275050986072060a550e4fee09ab418f6 Thanks!

PR #2046, opened 2015-12-16
Make "Hello" part of notification emails use user's username if we don't know their real name

### LINKED ISSUE #2037: "Hello" part of notification emails doesn't fall back to username
Relevant code: https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailadmin/templates/wagtailadmin/notifications/base_notification.html#L2
Introduced in: https://github.com/torchbox/wagtail/pull/1973

If the user doesn't have a name set, I think this should fallback to the username, if one is set (or simply just say "Hello" without a space afterwards). It currently says ``Hello ,``

Looks like an easy fix

kaedroho:
Fixes #2037


gasman:
Merged in 7fee62f8ebb9db32b7ce474d87b2c2ec21c2a70a

PR #2049, opened 2015-12-16
"Signing in..." for translation

MechanisM:


gasman:
Thanks! Merged in 039db5ed8911a1ad60cf85c320e628118f018ec9.

MechanisM:
btw, everywhere in wagtail templates I see mix of double quotes and single quotes inside templatetags like 'trans'. I think it will be better to decide which ones is better and replace everywhere to singlequotes or doublequotes.

gasman:
I think the recommended practice is to use double quotes for text to be read by humans (which will usually go through translation, but not always - developer-facing error messages would be one exception), and single quotes for internal strings such as identifiers and class names. We've probably not been consistent about that though!

kaedroho:
Django template tags can sometimes be used within HTML attributes. In this case we would always use single quotes as we usually use double quotes for the attribute.

So if we had to pick one, I would pick single quotes.

Issue #961, opened 2015-02-04
AbstractDocument required


davecranwell:
Recent implementations have required modifications to the Document model to add custom fields e.g a language indicator and a cover image (for documents which are digital versions of paper publications).

An AbtractDocument approach similar to images would greatly help this. @kaedroho has suggested we wait for https://github.com/torchbox/wagtail/pull/891

gasman:
Implemented in #2051

Issue #2001, opened 2015-12-04
Documents should be documented like images


alexgleason:
There's a section on creating image fields and templating them in the Wagtail docs. It would be nice to also include a section on creating document fields. Without it, the developer needs to view the Wagtail source code to know how to implement it.

Also feel free to reame this issue to something more explicit than "please document documents"...

m1kola:
Related question (and answer) on StackOverflow: http://stackoverflow.com/a/43389396/896843

SalahAdDin:
:+1:

lb-:
I think it would be good to have cross links to document management commands, using in templates (as called out on the main description), settings that impact the way documents are served. 

Finding out about documents currently is a bit haphazard.

lb-:
## Proposed approach.

Add ~4 new sub-pages to https://docs.wagtail.org/en/stable/advanced_topics/documents/index.html (in `docs/advanced_topics/documents`)

1. Overview /basics page
  - Covers including the app `'wagtail.documents'` in `INSTALLED_APPS`
  - Covers including URLs `from wagtail.documents import urls as wagtaildocs_urls` etc
  - Explain that by default they will be included in the references index
  - Add a Page example with a document file, then the page template that will make a link to that file available.
  - How the work within collections (cross link) and can be made private using private collections
  - Made available through the API via `wagtail.documents.api.v2.views.DocumentsAPIViewSet` - with a cross link to the relevant API section
2. Rich text
  - Reference to `document-link` usage with simple example
3. Customising the document upload form
  - `WAGTAILDOCS_DOCUMENT_FORM_BASE` link to setting with code example (maybe a simple T&C checkbox that the file was not created by AI :) that has to get checked on the clean method but is not saved to the Document model )
  - Link to the already created sub-pages for custom Document models and customising the upload title.
4. Storing and serving
  - Explain how files are stored - cross link to https://docs.wagtail.org/en/stable/advanced_topics/deploying.html#user-uploaded-files + reference to the setting `DEFAULT_FILE_STORAGE`
  - Link to `WAGTAILDOCS_SERVE_METHOD` with an example
  - Link to `WAGTAILDOCS_CONTENT_TYPES` with an example
  - Link to `WAGTAILDOCS_INLINE_CONTENT_TYPES` with an example
  - Link to `WAGTAILDOCS_EXTENSIONS` - with examples & link to https://docs.djangoproject.com/en/3.1/ref/validators/#fileextensionvalidator (the latest version of this)
[...]

PR #2051, opened 2015-12-17
Allow overriding document model

alexgleason:
This is a fix for #961

The implementation is very similar to how wagtailimages handles it. There's a get_document_model function which returns either the Document class, or it uses `WAGTAILDOCS_DOCUMENT_MODEL` if it's set.


kaedroho:
Looks good to me

kaedroho:
Merged in 27829f6090c7c3a3c81846868801a9f0993e8582

Proper docs for this will come in #2001

PR #2052, opened 2015-12-17
Pull translations from Transifex

gasman:


MechanisM:
@gasman can you pull again? Just updated russian to 100%

PR #2061, opened 2015-12-22
Use an dictionary for StreamField errors

timheap:
Instead of filling an array with `None` to indicate no errors, a dictionary is only set when there is an actual error. This will allow future functionality for errors to be added to the StreamField itself, rather than a block, similar to adding errors to a form rather than a specific field.

### DIFF
--- a/wagtail/wagtailcore/blocks/stream_block.py
+++ b/wagtail/wagtailcore/blocks/stream_block.py
@@ -107,21 +107,20 @@ class BaseStreamBlock(Block):
         return "StreamBlock(%s)" % js_dict(opts)
 
     def render_form(self, value, prefix='', errors=None):
+        error_dict = {}
         if errors:
             if len(errors) > 1:
                 # We rely on ListBlock.clean throwing a single ValidationError with a specially crafted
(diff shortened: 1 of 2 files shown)


wagtail/wagtailcore/blocks/stream_block.py
```diff
@@ -168,10 +167,8 @@ def clean(self, value):
                 )
             except ValidationError as e:
                 errors.append(ErrorList([e]))
```
kaedroho: ``.append()`` needs to be changed to something that works with dicts. (and we're probably missing a test for this branch)

timheap: Good catch. I've fixed this, and added a test for ValidationErrors.

kaedroho:
Awesome, thanks! This looks good to me but I assume @gasman might want to give it a quick look first.

timheap:
Rebased on master, and also fixed a shameful grammatical typo in the commit :disappointed:

gasman:
Looks good, thanks! Merged in 968676c5bb3d4ad83f6279d1f1c95a21fb58142a + b7fd5de8ecfd1eb32f5f820a2e45edd4ef16ea26.

PR #2062, opened 2015-12-22
Prevent cascade deletion of pages when a ContentType is deleted

### LINKED ISSUE #2024: Change on_delete action on Page.content_type
Currently, the ``on_delete`` action on ``Page.content_type`` is set to Django's default ``models.CASCADE``. This could cause a bunch of pages to be "badly" deleted if their specific model is removed, causing their children to be orphaned and their parent's ``numchild`` to not be updated.

We've discussed this and decided that the ``on_delete`` action chould be changed to ``models.SET`` setting the ``content_type`` field to the ``ContentType`` representing the standard ``Page`` model.

https://docs.djangoproject.com/en/1.9/ref/models/fields/#django.db.models.SET

kaedroho:
Fixes #2024


kaedroho:
Drone seems to be dead. ``flake8 wagtail`` passes locally

wagtail/wagtailcore/models.py
```diff
@@ -207,6 +207,14 @@ def get_page_models():
     return PAGE_MODEL_CLASSES
 
 
+def get_default_page_content_type():
+    """
+    Returns the content type to use as a default for pages which their
```
tomdyson: I don't understand this sentence - should it be 'when their content type'? or 'whose content type'?

tomdyson:
@timheap could you have a look at this, please?

wagtail/wagtailcore/models.py
timheap: I agree, this wording is awkward. Consider something like:

> Get the content type to use as a default when the content type for a specific Page type is deleted

timheap:
Looks good, besides from the docstring on `get_default_page_content_type()`!

tomdyson:
Thanks a lot, @kaedroho and @timheap

gasman:
Merged in 078491b6e068cb5bf7ba9f190cfc3ac085548c28

PR #2064, opened 2015-12-22
Set on_delete action to SET_NULL on Document.uploaded_by_user

kaedroho:
This prevents deletion of users from deleting any Documents they have uploaded.

This is already the behaviour on images.


gasman:
Merged in 8ef846d2146e43136a10cced50c170c88d8febbd

PR #2065, opened 2015-12-22
Add on_delete=CASCADE to all ForeignKeys without on_delete action 

kaedroho:
Defining ForeignKeys/OneToOneFields without an on_delete action is deprecated in Django 1.9.

This commit adds on_delete=CASCADE to every ForeignKey that doesn't yet have an on_delete action.


gasman:
Merged (5b73b8f cherry-picked) in 827ead12c2eee4807fffe085d45360eaf1841b28 - thanks!

PR #2066, opened 2015-12-22
Remove unnecessary dirs on .coveragerc

chimeno:
Remove south_migrations and vendors , those dirs are not used anymore


gasman:
Thanks!

PR #2067, opened 2015-12-23
[RFC] Admin modules

kaedroho:
This PR reimplements my old admin modules PR #890 for recent versions of Wagtail (using the new namespaced URLs and generic views features).

The idea is to build the admin site up using "modules". These provide a url config, menu items, permissions and other things for Wagtail. Modules are all class based and are easy to subclass.

I think there's a few good reasons for modularising the admin interface like this:
 - Reduces duplicated view code
 - Makes it easier to add/remove/modify/duplicate sections of the admin
 - Moves model view permission logic to a higher level (currently each view is responsible for permissions, which is a source of mistakes)
  - Also allows each component of a module (urls, menu item, dashboard panel, etc) to share permission logic

This PR adds modules into Wagtailadmin and refactors the sites section to use one.

To use admin modules to register a model in the admin, just add the following to a ``wagtail_hooks.py`` file:

```python
# app/wagtail_hooks.py
from wagtail.wagtailadmin.modules.model import ModelModule

from .models import MyModel


@hooks.register('register_admin_module')
def register_admin_module():
    return ModelModule('mymodel', model=MyModel, icon='the-icon')
```

TODO:
 - Permission checks
 - Permissions (for the groups interface)
 - Menu items
 - Search areas
 - Dashboard panels

### DIFF
--- a/wagtail/wagtailadmin/modules/__init__.py
+++ b/wagtail/wagtailadmin/modules/__init__.py
@@ -1,0 +1,21 @@
+from django.conf.urls import url, include
+
+from wagtail.wagtailcore import hooks
+
+
+def register_modules():
+    for fn in hooks.get_hooks('register_admin_module'):
+        module = fn()
+
+        if module:
+            urlpatterns = module.get_urlpatterns()
+
+            if urlpatterns:
+                @hooks.register('register_admin_urls')
+                def register_admin_urls():
+                    return  [
+                        url(
+                            r'^{}/'.format(module.name),
+                            include(urlpatterns, app_name=module.name, namespace=module.name)
+                        ),
+                    ]
--- a/wagtail/wagtailadmin/modules/base.py
+++ b/wagtail/wagtailadmin/modules/base.py
@@ -1,0 +1,9 @@
+class Module(object):
+    def __init__(self, name, **kwargs):
+        self.name = name
+
+        for key, value in kwargs.items():
+            setattr(self, key, value)
+
+    def get_urlpatterns(self):
+        return []
--- a/wagtail/wagtailadmin/modules/model.py
+++ b/wagtail/wagtailadmin/modules/model.py
@@ -1,0 +1,84 @@
+from django.conf.urls import url
+from django.utils.translation import ugettext as _, ugettext_lazy as __
+from django.forms.models import modelform_factory
+
+from .base import Module
+
+from wagtail.wagtailadmin.views import generic
+
+
+class ModelModule(Module):
+    icon = ""
+
+    index_view_class = generic.IndexView
+    add_view_class = generic.CreateView
+    edit_view_class = generic.EditView
+    delete_view_class = generic.DeleteView
+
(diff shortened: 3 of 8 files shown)


gasman:
This looks good! I came up with something very similar here https://github.com/gasman/wagtail/commits/feature/generic-model-admins (but didn't pursue it, as the generic-views work was only a spinoff from collections in the first place, and I didn't want to get too derailed :-) ). I particularly like the approach of passing parameters through the `as_view` method... I struggled to find a good way to pass settings from the module to the CBVs while keeping the CBVs agnostic about the whole module pattern, and ended up with a pretty horrible monkeypatching approach (https://github.com/gasman/wagtail/commit/5f8f74f08872ee9538937e80ce3c303f07035666).

A couple of things I'd pick up on:

* As it stands the base 'Module' class seems like needless abstraction to me - either a module follows the CRUD pattern (in which case it should inherit from ModelModule), or it doesn't, in which case it might as well use plain Django views / urlconfs rather than wrapping it in a new kind of object that provides no added functionality. I can see how it might potentially be used in future as somewhere to hang the MenuItem object off (for example), but I'd be inclined to hold off on this until it's clear that there's a definite advantage from doing so.
* If we are keeping 'Module', I'd like to look for a better name, since it clashes with the Python concept of a module, and isn't very descriptive of what it really does (namely, provides a collection of related views). `ViewSet` perhaps?
* Having a single 'name' property that serves as the URL namespace and template path probably isn't sufficient - for example, the collections settings views have a URL namespace of `wagtailadmin_collections` and a template path of `wagtailadmin/collections`. Ideally we'd have separate `url_namespace` and `template_path` properties that default to `name` if not specified - we can cross that bridge when we come to it, though.

kaedroho:
Thanks @gasman!

I like the name ``ViewSet``. Django REST Framework uses that name for pretty much the same concept (http://www.django-rest-framework.org/api-guide/viewsets/).

I'd like it if modules/viewsets weren't tied into models as I think someone might want to add a non-model module (eg a uwsgi dashboard). Also, the page explorer and search promotions probably won't fit in nicely with ``ModelModule`` as it makes lots of assumptions about how you want the views to be layed out/permissions to work/etc.

gasman:
OK, that sounds good. I agree that there will be plenty of admin areas that don't fit the ModelModule pattern. My original thoughts went something along the lines of: "this change is primarily about eliminating boilerplate repetition for things that *do* follow the standard ModelModule conventions; for the non-standard things, there is no generic boilerplate to eliminate, so there's no advantage over using plain views / urlconfs - it's just adding extra abstraction to solve a non-problem."

However, given the more descriptive name, and seeing how it fits into DRF, I'm now persuaded that the concept of "a bundle of related views that can be treated as a single unit" is a useful building block in its own right, so happy to keep it.

SalahAdDin:
:+1:

kaedroho:
Superseeded by #2113

PR #2069, opened 2015-12-23
Fix abstract Page managers

timheap:
Custom Page managers on abstract Page subclasses caused an error when starting the application. This PR fixes this issue. An example of the failing, but now fixed code is below.

```python
class MyBasePage(Page):
    objects = CustomManager()
    class Meta:
        abstract = True

class MyCustomPage(MyBasePage):
    pass
```


kaedroho:
Looks good to me!

PR #2074, opened 2015-12-28
Remove repeated instruction

benjaoming:
In order to keep the dev instructions shorter, we can refer to the chapter on compiling static instead of writing the same commands two times.

### DIFF
--- a/docs/contributing/developing.rst
+++ b/docs/contributing/developing.rst
@@ -28,17 +28,7 @@ With your preferred virtualenv activated, install the Wagtail package in develop
 
     python setup.py develop
 
-Install the tool chain for building static assets:
-
-.. code-block:: sh
-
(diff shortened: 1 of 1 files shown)


gasman:
Hi @benjaoming - thanks for the suggestion! However, I think in this case we really do want to repeat the instructions: the "Compiling static assets" instructions do not include the `npm install` step, and it's much more helpful to give users a linear set of steps to follow, instead of making them jump forward and back in the documentation.

benjaoming:
I managed to overlook it anyways :) In general, having repeated information defeats the purpose of having clear instructions.. so if you wanna keep the instructions in the upper section, I'd suggest removing the lower one as it's like 2/3 repetition and the remaining 1/3 (which is just some text) can be moved up.

PR #2086, opened 2016-01-05
Drop Django 1.7 support

gasman:
All references to Django 1.7 have been removed from code, documentation and tests.

This makes `wagtail.utils.compat` redundant, and the functions in it have been deprecated. (This module was undocumented, but it's conceivable that third-party packages might be relying on it, so it's appropriate to put them through the deprecation process. The module itself will probably remain, for the benefit of any future Django updates that require version switches.)


wagtail/wagtailadmin/tests/test_jinja2.py
```diff
@@ -11,11 +7,9 @@
 from wagtail.wagtailcore.models import Page, PAGE_TEMPLATE_VAR, Site
 
 
-@unittest.skipIf(django.VERSION < (1, 8), 'Multiple engines only supported in Django>=1.8')
 class TestCoreJinja(TestCase):
 
     def setUp(self):
-        # This does not exist on Django<1.8
         from django.template import engines
```
timheap: This could be moved to a top level import now that it will not throw ImportErrors for old Django versions

gasman: Cheers - moved in fe6b9a11beeafbbcafb443bbc78169a4d5c8ce58.

kaedroho:
Merged in d8bceff38be87b1802193caf8c3b854c435aab1a +parents

PR #2089, opened 2016-01-07
Drop wagtail prefix for contrib apps

kaedroho:
This PR removes the "wagtail" prefix from contrib apps.


MechanisM:
Good! I've been waiting for it for long time.
Btw, you're going to rename wagtailsearchpromotions, wagtailsitemaps, wagtailstyleguide as well?

kaedroho:
@MechanisM Yep, just doing those now

gasman:
Looks good to me.

Are we happy to make this an immediate change with no deprecation period, then? I guess that's reasonable enough... given that the upgrade is basically a mechanical process, with obvious errors if it gets missed, it doesn't make a whole lot of difference whether site owners are required to do that as part of the 1.4 upgrade or at some indefinite point between then and 1.6. (Not sure how you'd do the deprecation anyhow - presumably having stub modules under the old names consisting of `from wagtail.contrib.api import *` would fix imports, but not INSTALLED_APPS entries - so that might be a moot point...)

timheap:
You could fix `INSTALLED_APPS` by setting a `default_app_config` that imports the correct app. You could even make the `default_app_config` used raise a deprecation warning when used, on top of all the other deprecation warnings.

gasman:
Hmm. I'm 50/50 on whether it's worth adding the deprecation path, then... it does make the upgrade experience nicer, but only marginally so (since they'll have to do the search-and-replace sooner or later anyway) - and if the deprecation approach turns out to be less clean and reliable than hoped, and causes odd failure cases, then that's a step backwards from it failing cleanly with an obvious error message.

timheap:
The same argument can be made for any deprecation period. People will need to make the upgrade eventually, but the deprecation period will give them time to do it when they have the time. It will also help with compatibility of third part apps that import core wagtail features. Apps that want to keep compatibility with multiple versions of Wagtail, when importing core modules that have been renamed is not fun!

A quick and dirty way of testing the renaming is robust is to rename all the apps, but keep the tests using the old import paths. If the tests still pass with the old paths, everything should be fine?

kaedroho:
Thanks @timheap. I've implemented some backwards compatibility fixes above. Could you review?

timheap:
I looked at how `six` does its Python 2 / 3 renamed imports, to see if we could make use of similar functionality here. Reading the code lead me to PEP 302, which specifies how to write custom module loaders. Writing one turned out to be surprisingly easy, so I've done so and submitted a PR against your fork: https://github.com/kaedroho/wagtail/pull/36

wagtail/tests/settings.py
```diff
@@ -91,12 +91,12 @@
...
-    'wagtail.contrib.wagtailroutablepage',
-    'wagtail.contrib.wagtailfrontendcache',
-    'wagtail.contrib.wagtailapi',
-    'wagtail.contrib.wagtailsearchpromotions',
+    'wagtail.contrib.styleguide',
+    'wagtail.contrib.sitemaps',
+    'wagtail.contrib.routablepage',
+    'wagtail.contrib.frontendcache',
+    'wagtail.contrib.api',
+    'wagtail.contrib.searchpromotions',
```
BertrandBordage: Couldn’t it be the occasion to be more pythonic and clear by adding underscores between words?
The modules would become:
- `style_guide`
- `sitemaps` (special case here since Sitemaps is a technology name)
- `routable_page`
- `frontend_cache`
- `api`
- `search_promotions`

If others think this is the right way to go, then an additional pull request should be done to rename `modeladmin` to `model_admin`. This pull request could also be updated to do that, but I fear it will be overly complex.

timheap: I've never quite worked out what the most Pythonic way to name modules is regarding underscores. Even Django is inconsistent with this: `context_processors` vs `templatetags` for one example. However, Django does not use underscores for any of its `contrib` modules. If Wagtail started using underscores to be more Pythonic, we would become less Djangoish at the same time.

I don't necessarily object to adding underscores if there is a solid argument for it, but I am not convinced it is more Pythonic given the confusion around underscores in module names that is already present.

BertrandBordage: First of all, there is already an underscore in Wagtail’s `table_block` contrib module. So either way, we have to do something to make naming more consistent.

I think the confusion has to do with Python’s history. At first, there was no rule for naming, so we ended up with module names like HTMLParser or ConfigParser and methods like assertEqual. Then rules were added in 2001 with PEP8, and a lot of Python modules & methods were renamed, but it is still messy. For example, ConfigParser was renamed to configparser or assertEqual etc. weren’t renamed, probably because they were copied from Java and the habit stayed (the backwards compatibility excuse written a PEP8 is obsolete, now that we know how to make deprecations).

PEP8 itself isn’t very consistent, since [it discourages underscores in module names but allow them for readability](https://www.python.org/dev/peps/pep-0008/#package-and-module-names), while [insisting on adding underscores to improve readability for functions, method and constant names](https://www.python.org/dev/peps/pep-0008/#function-names).
This is misleading for everyone and it’s pointless IMO to want to save one or two characters by decreasing readability and consistency.

gasman: If there's no clear precedent across python/django then I have a slight preference for no-underscores, since it makes the task of updating import lines across a project more 'automatic' (less thinking required) and less annoying. But I guess this is [bikeshed](http://bikeshed.org/) territory, so I'll happily defer to the person who's actually doing the work :-)

kaedroho:
Going to close this as it's way out of date.

I opened an RFC a while back for general renaming/moving a few apps which includes this change. I think that'll be the best place to continue the discussion about the approach: https://github.com/wagtail/rfcs/pull/12

PR #2091, opened 2016-01-08
z-index audit

### LINKED ISSUE #1958: DateTime picker z-index below modal
The javascript datetimepicker used by the `AdminDateInput()` widget has [a z-index of 5](https://github.com/torchbox/wagtail/blob/88e477098f31da29683eec1a5c051b0664218b9d/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/datetimepicker.scss#L9) while the [z-index for modals appears to be 500+](https://github.com/torchbox/wagtail/blob/88e477098f31da29683eec1a5c051b0664218b9d/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/modals.scss#L1) meaning the widget appears hidden behind the modal if you have a form within a modal dialog that contains it. I've gotten around it for now simply by overriding in our CSS:

```
.xdsoft_datetimepicker {
    z-index:520;
 }
```

but I'm unsure if this is breaking the flow of z-index within the rest of wagtail.

davecranwell:
Fixes #1958. I think.

Most of this has boiled down to removing z-indexes widely. DOM element order naturally provides most of the implications on Z-index, making many of them unnecessary.

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
@@ -59,6 +59,10 @@ function makeRichTextEditable(id) {
             $(window).trigger('scroll');
         });
     });
+
+    $(document).on('show.bs.modal', '.modal', function() {
+        richText.trigger('hallodeactivated');
+    });
 }
 
 function insertRichTextDeleteControl(elem) {
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/_variables.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/_variables.scss
@@ -71,8 +71,9 @@ $thumbnail-width: 130px;
 $menu-width: 180px;
 
 // Z-indexes by type
-$z-indexes:(
-
-    modals,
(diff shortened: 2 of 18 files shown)


kaedroho:
Thanks @davecranwell I've noticed a few regressions, see images below

z-index-1
z-index-2
z-index-3
z-index-4

kaedroho:
All three regressions reported above are now fixed.

davecranwell:
Still to test: a custom image model with a date field.

kaedroho:
The image chooser modal doesn't appear to use the date picker. Something we must've overlooked earlier which is out of scope of this PR.

I assume @ptink probably created a custom modal with a date picker so shall we defer fixing #1958 until we support the date picker in the image chooser modal? this would be much easier for us to test.

davecranwell:
I think we should merge this as-is (once conflicts are resolved). Date popups within image modals are a customisation and shouldn't be a blocker to merging this.

gasman:
@JoshBarr Passing this one on to you to decide what to do with it, in case it's relevant to the forthcoming CSS refactor. (It may well be too far behind master to be usable now, in which case feel free to close...)

PR #2101, opened 2016-01-12
Use translation for the Groups menu item

abogushov:
I think translation for the Groups menu item has been removed by mistake. I return the translation back.


kaedroho:
Thanks for fixing!

Issue #1712, opened 2015-09-21
Configurable moderation workflow


kurtrwall:
Sometimes an organization wants to be able to share content without having to give members user accounts on the CMS because of man power, bureaucracy, etc. Should there be a mechanism that allows this sort of behavior? 

I'm thinking of having multiple sites that could provide two different versions of the content based on a "shared" status. 

I'm cooking up a PR that would support this though subclassing the Page model and making good use of Wagtail's Site class, but I thought I might get the community's thoughts on it.

gasman:
Hi @kurtw - can you give some more details of the mechanism you're proposing, and how you expect it to be used? It sounds like you're suggesting having a secondary site with locked-down access (e.g. by IP block, or htaccess password) but I'm not quite sure.

kurtw:
@gasman The first part is basically covered in the concept you proposed in [1283 (comment)](https://github.com/torchbox/wagtail/issues/1283#issuecomment-100470750). Imagine a scenario of having `www.website.com` and `content.website.com` where www is public facing and content is internal (only accessible via VPN or something), but they both exist on the production server so they can use the same database. In essence, they are just two versions of the same site.

What I'm suggesting is that there be some way of "replicating" the publishing workflow on a per site basis. One site would be used for sharing content among those with access without the need for an account and the other would be where the finalized content goes. 

I opened [a PR](https://github.com/torchbox/wagtail/pull/1713/files) that illustrates what I'm talking about. It works in combination with some [wagtail hooks](https://gist.github.com/kurtw/6ee8e27aa8e158c52e9a) that are in my app. Admittedly, there's some bugs in it and again it won't work without the hooks but it does the job okay. I'm not necessarily suggesting this be the way it's done but it's one way I guess.

gasman:
@kurtw Thanks for the clarification! Looking at the PR, this feels like a very "heavy" change (in terms of additions to the admin UI, and the work involved in dealing with it during future Wagtail development) for something that's only useful to a minority of users.

This is really hinting at a much bigger feature, the idea of "configurable workflow". Currently Wagtail implements one simple kind of workflow, the "submit for moderation" button - and your PR implements a different workflow, designed for your own organisation's requirements (namely, the ability for users without Wagtail accounts to review pages). Undoubtedly other organisations will have different requirements - such as multiple levels of moderators, perhaps. It wouldn't be practical for Wagtail to implement each organisation's requirements on an individual basis - it would quickly become unmaintainable.

[...]

PR #2102, opened 2016-01-12
Permission policy modules

gasman:
This PR implements permission policies, as originally proposed in https://github.com/torchbox/wagtail/issues/1384#issuecomment-142655487. These are objects that provide a central point for making permission queries for a particular model - such as "can user X perform action Y on instance Z?" or "which instances can user X perform action Y on?". By collecting this logic into a single module, rather than having it distributed across templates, views and models, it's possible to redefine the permission rules for an app by swapping in a new policy object.

The currently defined permission policies are:

* `BlanketPermissionPolicy` - allow everyone (including anonymous users) to do everything, effectively disabling permission rules entirely
* `AuthenticationOnlyPermissionPolicy` - allow all active, authenticated users to do everything
* `ModelPermissionPolicy` - apply standard `django.contrib.auth` permission rules, which act on the model as a whole (i.e. it's not possible to give a user edit access to a subset of a model's instances)
* `OwnershipPermissionPolicy` - the current behaviour of documents / images, appropriate for models that have a concept of 'ownership', where a user with 'add' permission automatically receives permission to change/delete instances they own.

wagtailimages, wagtaildocs and wagtailredirects have been converted to use permission policies, along with wagtailadmin's generic views (which means that wagtailsites uses them as well).

The immediate use-case for this will be to switch wagtailimages and wagtaildocs to the collections system (#1751), but with further work it will be possible to:

* Allow individual Wagtail installations to customise permission policies via settings (#1384). I haven't worked out the best way of doing this - could we pass it as an appconf parameter in INSTALLED_APPS, perhaps?
* Customise the page permission model (#1993). I haven't attempted this here as page permissions will be significantly more complex - more operations to consider (e.g. move / copy), more business rules to account for (e.g. parent_page_types / subpage_types), and the UserPermissionProxy / PagePermissionTester pattern (which would be a useful extension to the permission policy API in general, as it avoids repeated DB hits on listing pages).
* Build up the group permission interface programmatically so that it outputs just the checkboxes relevant to the permission policy in use, rather than the app having to specifically register them. (For example - BlanketPermissionPolicy and AuthenticationOnlyPermissionPolicy would not add anything to the permissions interface, as there's nothing to configure; ModelPermissionPolicy would supply checkboxes for add, change and delete, and OwnershipPermissionPolicy would supply checkboxes for add and change, as it doesn't distinguish between change and delete permissions.)
* Potentially, implement custom workflows (#1712) - as I see it, these can be modelled as a state machine, along with a permission policy that varies permissions on an object according to its current state.

### DIFF
--- a/wagtail/tests/customuser/models.py
+++ b/wagtail/tests/customuser/models.py
@@ -8,7 +8,7 @@ from django.contrib.auth.models import (
 
 class CustomUserManager(BaseUserManager):
     def _create_user(self, username, email, password,
-                     is_staff, is_superuser, **extra_fields):
+                     is_staff, is_superuser, is_active=True, **extra_fields):
         """
         Creates and saves a User with the given username, email and password.
         """
@@ -16,7 +16,7 @@ class CustomUserManager(BaseUserManager):
             raise ValueError('The given username must be set')
         email = self.normalize_email(email)
         user = self.model(username=username, email=email,
-                          is_staff=is_staff, is_active=True,
+                          is_staff=is_staff, is_active=is_active,
                           is_superuser=is_superuser, **extra_fields)
         user.set_password(password)
         user.save(using=self._db)
--- a/wagtail/wagtailadmin/utils.py
+++ b/wagtail/wagtailadmin/utils.py
@@ -124,6 +124,27 @@ def any_permission_required(*perms):
     return user_passes_test(test)
 
 
+class PermissionPolicyChecker(object):
+    """
+    Provides a view decorator that enforces the given permission policy,
+    returning the wagtailadmin 'permission denied' response if permission not granted
+    """
+    def __init__(self, policy):
+        self.policy = policy
+
+    def require(self, action):
+        def test(user):
+            return self.policy.user_has_permission(user, action)
+
+        return user_passes_test(test)
+
+    def require_any(self, *actions):
+        def test(user):
+            return self.policy.user_has_any_permission(user, actions)
+
+        return user_passes_test(test)
+
+
 def send_mail(subject, message, recipient_list, from_email=None, **kwargs):
     if not from_email:
         if hasattr(settings, 'WAGTAILADMIN_NOTIFICATION_FROM_EMAIL'):
--- a/wagtail/wagtailadmin/views/generic.py
+++ b/wagtail/wagtailadmin/views/generic.py
@@ -9,36 +9,41 @@ from wagtail.wagtailadmin.utils import permission_denied
 
 class PermissionCheckedMixin(object):
     """
-    Mixin for class-based views to enforce permission checks.
-    Subclasses should set either of the following class properties:
-    * permission_required (a single permission string)
-    * any_permission_required (a list of permission strings - the user must have
+    Mixin for class-based views to enforce permission checks according to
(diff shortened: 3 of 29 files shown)


kaedroho:
> Allow individual Wagtail installations to customise permission policies via settings (#1384). I haven't worked out the best way of doing this - could we pass it as an appconf parameter in INSTALLED_APPS, perhaps?

> Build up the group permission interface programmatically so that it outputs just the checkboxes relevant to the permission policy in use, rather than the app having to specifically register them. (For example - BlanketPermissionPolicy and AuthenticationOnlyPermissionPolicy would not add anything to the permissions interface, as there's nothing to configure; ModelPermissionPolicy would supply checkboxes for add, change and delete, and OwnershipPermissionPolicy would supply checkboxes for add and change, as it doesn't distinguish between change and delete permissions.)

I'm hoping that viewsets will provide a solution to both of these (#2113).

kaedroho:
Do you think it's worth having both ``BlanketPermissionPolicy`` and ``AuthenticationOnlyPermissionPolicy``? These basically mean the same thing as non-authenticated users will be turned away before they get to these model-level permissions anyway.

kaedroho:
Haven't checked yet if this is a pre-existing issue.

If I browse to a page that has an image chooser, but I do not have permissions to edit the selected image. I still get given the "Edit this image" button which just leads me to the dashboard with a big red banner saying "Sorry, you do not have permission to access this area."

kaedroho:
Thanks @gasman, this is really neat work!

Could you resolve: https://github.com/torchbox/wagtail/pull/2102#commitcomment-15504910

gasman:
> Do you think it's worth having both ``BlanketPermissionPolicy`` and ``AuthenticationOnlyPermissionPolicy``? These basically mean the same thing as non-authenticated users will be turned away before they get to these model-level permissions anyway.

The primary use-case for these policies is for sites that implement their own authorisation as a layer outside of Wagtail, as in #1384. I'd prefer not to make any assumptions about how these external systems will behave and what Wagtail needs to do to integrate correctly with them - for example, it might be done through a Django middleware that just sets `request.user` and does not do any permission enforcement itself (in which case Wagtail should reject anonymous users), or through a system like htaccess that doesn't touch Django at all (in which case we should allow them - making that work sensibly for wagtailadmin would require a non-trivial amount of additional work, of course). Since those implementations only take a few lines each, I figured it was worth leaving both options open.

I also want to leave open the possibility of these being used not just for admin interfaces - and indeed not just for Wagtail (there's nothing Wagtail-specific in the implementation). For something like a wiki, or a bug tracking tool, the distinction between the two policies is a lot more meaningful.

gasman:
Rebased and refactored to fix the issue with custom image/document models. I was wrong in my assessment that we couldn't define the permission_policy object statically at startup - we just can't do it in models.py (because the startup process specifically imports that).

Permission models are now defined in permissions.py, and since that implies having a wagtailcore/permissions.py to hold the permission policy for Site (it makes more sense to have it in the app that defines the model than in the app that edits it, I think), the actual implementation of the permission policies now lives in wagtailcore/permission_policies.py.

kaedroho:
Merged in 73b6ba01a551eb4065c8ecda6b343cff14487425 +parents. Thanks!

PR #2107, opened 2016-01-12
Explicitly set wagtailcore_page.path to use the 'C' collation

gasman:
Treebeard's path comparison logic can fail on certain locales such as sk_SK, which
sort numbers after letters. To avoid this, we explicitly set the collation for the
'path' column to the (non-locale-specific) 'C' collation.

Ref: https://groups.google.com/d/msg/wagtail/q0leyuCnYWI/I9uDvVlyBAAJ

To reproduce the original issue (on a standard wagtaildemo vagrant installation):

    sudo locale-gen sk_SK.UTF-8
    sudo dpkg-reconfigure locales
    sudo /etc/init.d/postgresql restart
    dropdb wagtaildemo
    createdb wagtaildemo --locale=sk_SK.UTF-8 --template=template0
    ./manage.py migrate
    ./manage.py load_initial_data

### DIFF
--- a/wagtail/wagtailcore/migrations/0001_initial.py
+++ b/wagtail/wagtailcore/migrations/0001_initial.py
@@ -6,6 +6,20 @@ from django.conf import settings
 import wagtail.wagtailsearch.index
 
 
+def set_page_path_collation(apps, schema_editor):
+    """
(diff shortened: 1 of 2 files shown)


kaedroho:
Merged in a6d3d23e648e8442bf78329c5c233c49bef7d3f4. Thanks!

PR #2108, opened 2016-01-12
Fix custom image docs to indicate that fields must be non-required

gasman:
As per https://groups.google.com/d/msg/wagtail/DPDOWiWWrKM/TPj0XZhHCQAJ

### DIFF
--- a/docs/topics/images/advanced_topics.rst
+++ b/docs/topics/images/advanced_topics.rst
@@ -35,7 +35,7 @@ Here's an example:
         # Add any extra fields to image here
 
         # eg. To add a caption field:
-        # caption = models.CharField(max_length=255)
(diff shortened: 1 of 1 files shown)


kaedroho:
It is possible to have required fields on image models, it just doesn't work in the multiple uploader yet (but the simple uploader, modal uploader and edit page work with required fields).

gasman:
True - but then you're essentially saying "you can do X, as long as you don't mind ending up with a partially broken CMS", and I don't think we should be promoting that sort of flakiness in the documentation. I'd rather just say "you can't do X", even if it is slightly underselling Wagtail's capabilities :-)

kaedroho:
Merged in 6851bbf1a6776cc61ca7c7c3d072eaa48f1258f2

PR #2111, opened 2016-01-14
Fix static asset building on Windows

akx:
Using `path.sep` instead of the constant `'/'` makes the build pipeline copy the files into the correct directories.


akx:
Erm, hm, wait, there's still something funky...

kaedroho:
Tested on Windows 10, confirmed the issue exists and this PR fixes it. Also checked on Linux and no regressions there.

@davecranwell are you happy with the implementation?

kaedroho:
Merged in 6f18c5ea0f2d0e2b167dcdfeed6fd06383e70bf6 Thanks!

Issue #1937, opened 2015-11-18
Separate initial data from migrations


jordij:
We usually find the default **Welcome to your new Wagtail site!** default homepage a nuisance. While we agree that some initial data is needed (Site and default groups ok) we don't feel like a default homepage should be automatically populated.

Here at Springload we like the idea of having initial data separated from the migrations using fixtures. What do you guys think? :)

timheap:
I agree with this, but I also see the value of creating a fully functioning site from the outset, to get new developers up and running quickly. Perhaps a migration/fixture can be created as part of the `wagtail start` project that creates default site and 'Welcome to your new Wagtail site!' page, retaining the current work flow for new projects, but allowing developers already familiar with Wagtail to set up their project in the way that works for them?

alexgleason:
I've been burned a few times from flushing the database, migrating, then having -1 pages. Still not exactly sure why this happens sometimes.

timheap:
If you flush the DB, as opposed to dropping the tables and recreating the database, this would explain your problems. The initial data is constructed as part of a migration, not a fixture. After flushing the DB, the migrations will not run again as they have already been run, and the initial data will not be recreated. If the initial data was installed as part of a fixture, this problem would be avoided, although fixtures come with their own issues.

jordij:
Yes Tim, making it optional and/or as a part of the `wagtail start` would definitely work for us.

To give some more details, our deployment flow the first time looks like:
* Create db schema (`python manage.py migrate`)
* Load project's initial data (we use `loaddata`)

The initial data obviously contains an instance of the homepage that clashes with the one being created in `wagtailcore/migrations/0002_initial_data.py`

gasman:
> While we agree that some initial data is needed (Site and default groups ok) we don't feel like a default homepage should be automatically populated.

A slight snag with that - we can't create a default Site record unless we have a homepage for it to point to...

To be clear, there are two distinct places in Wagtail where we create an initial homepage: the `Page` instance in `wagtailcore/migrations/0002_initial_data.py`, and the `HomePage` instance created by the `wagtail start` project template as a replacement for the wagtailcore one: https://github.com/torchbox/wagtail/blob/master/wagtail/project_template/home/migrations/0002_create_homepage.py

I think the project template deals with this as well as we can hope for - the HomePage model exists as part of the project codebase, so it's possible for a developer to expand that into their final all-singing-all-dancing homepage without ever replacing it.

[...]

Issue #1824, opened 2015-10-13
Django "flush" management command fails because of "wagtailsearch_editorspick" table


atbaker:
I'm gearing up to use Wagtail for a new project and ran into a small issue flushing a Postgres database.

With a stock Wagtail installation (no contrib apps), running `python manage.py flush` fails because of the `wagtailsearch_editorspick` table:

```
CommandError: Database sample_project couldn't be flushed. Possible reasons:
  * The database isn't running or isn't configured correctly.
  * At least one of the expected database tables doesn't exist.
  * The SQL was invalid.
Hint: Look at the output of 'django-admin sqlflush'. That's the SQL this command wasn't able to run.
The full error: cannot truncate a table referenced in a foreign key constraint
DETAIL:  Table "wagtailsearch_editorspick" references "wagtailcore_page".
HINT:  Truncate table "wagtailsearch_editorspick" at the same time, or use TRUNCATE ... CASCADE.
```

After doing some digging, it looks like the key is that this migration leaves the `wagtailsearch_editorspick` table in the database for the contrib app to pick up:

https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailsearch/migrations/0003_remove_editors_pick.py

Relevant comment:

```
        # If wagtailsearchpromotions isn't installed, this table will remain
        # in the database unmanaged until it is. This could potentially happen
        # at any point in the future so it's important to keep this behaviour
        # even if we decide to squash these migrations.
```

I'm new to Wagtail (and really liking it so far!), so I'm not sure what the fix is here. If someone can point me in the right direction, though, I'm happy to put together a PR.

Thanks!

Additional background:

**Installed apps:**
```
INSTALLED_APPS = (
    'django.contrib.admin',
    'django.contrib.auth',
    'django.contrib.contenttypes',
    'django.contrib.sessions',
    'django.contrib.messages',
    'django.contrib.staticfiles',

    # Wagtail
    'wagtail.wagtailcore',
    'wagtail.wagtailadmin',
    'wagtail.wagtaildocs',
    'wagtail.wagtailsnippets',
    'wagtail.wagtailusers',
    'wagtail.wagtailimages',
    'wagtail.wagtailembeds',
    'wagtail.wagtailsearch',
    'wagtail.wagtailsites',
    'wagtail.wagtailredirects',
    'wagtail.wagtailforms',

    'compressor',
    'taggit',
    'modelcluster',

    # Other
    'rest_framework',
)
```

**Requirements:**
```
[...]

PR #2112, opened 2016-01-14
Add option to prevent migrations from creating initial page/site

akx:
I feel this is a very elegant solution to the concerns in #1937.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -43,6 +43,8 @@ Changelog
  * Fix: StreamBlock HTML rendering now handles non-ASCII characters correctly on Python 2 (Mikalai Radchuk)
  * Fix: Fixed a bug preventing pages with a OneToOneField from being copied (Liam Brenner)
  * Fix: SASS compilation errors during Wagtail development no longer cause exit of Gulp process, instead throws error to console and continues (Thomas Winter)
+ * The default initial data migration has been taught about a setting (`WAGTAIL_MIGRATE_INITIAL_PAGE_AND_SITE`) that makes it not create a Site and a Page (Aarni Koskela)
+ * The default initial data migration has been taught about another setting (`WAGTAIL_MIGRATE_INITIAL_GROUPS`) that makes it not create any groups (Aarni Koskela)
 
 1.3.1 (05.01.2016)
 ~~~~~~~~~~~~~~~~~~
--- a/docs/advanced_topics/settings.rst
+++ b/docs/advanced_topics/settings.rst
@@ -274,6 +274,21 @@ Case-Insensitive Tags
 
 Tags are case-sensitive by default ('music' and 'Music' are treated as distinct tags). In many cases the reverse behaviour is preferable.
 
+Initial Migration Data
+----------------------
+
+.. code-block:: python
+
+  WAGTAIL_MIGRATE_INITIAL_PAGE_AND_SITE = True
+  WAGTAIL_MIGRATE_INITIAL_GROUPS = True
+
+You can set these configuration options to ``False`` to not create the default "Welcome to your new Wagtail site!"
+page or the "Moderators" and "Editors" groups, respectively.
+
+.. note::
+
+  These settings are not directly compatible with the default project template.  If you set either of these settings,
+  be prepared to adapt your project's initial migrations accordingly.
 
 URL Patterns
 ~~~~~~~~~~~~
--- a/wagtail/wagtailcore/migration_tools/initial_data.py
+++ b/wagtail/wagtailcore/migration_tools/initial_data.py
@@ -1,0 +1,102 @@
+from django import VERSION as DJANGO_VERSION
+
+
+def create_initial_data(apps, schema_editor):
+    ContentType = apps.get_model('contenttypes.ContentType')
+    Group = apps.get_model('auth.Group')
+    Page = apps.get_model('wagtailcore.Page')
+    Site = apps.get_model('wagtailcore.Site')
+    GroupPagePermission = apps.get_model('wagtailcore.GroupPagePermission')
+
+    # Create page content type
+    page_content_type, created = ContentType.objects.get_or_create(
+        model='page',
+        app_label='wagtailcore',
+        defaults={'name': 'page'} if DJANGO_VERSION < (1, 8) else {}
+    )
+
+    # Create root page
+    root = Page.objects.create(
+        title="Root",
+        slug='root',
+        content_type=page_content_type,
+        path='0001',
+        depth=1,
+        numchild=1,
+        url_path='/',
+    )
+
+    # Create homepage
+    homepage = Page.objects.create(
+        title="Welcome to your new Wagtail site!",
+        slug='home',
+        content_type=page_content_type,
+        path='00010001',
+        depth=2,
+        numchild=0,
+        url_path='/home/',
+    )
+
+    # Create default site
+    Site.objects.create(
+        hostname='localhost',
+        root_page_id=homepage.id,
+        is_default_site=True
+    )
+
+    # Create auth groups
+    moderators_group = Group.objects.create(name='Moderators')
+    editors_group = Group.objects.create(name='Editors')
+
+    # Create group permissions
+    GroupPagePermission.objects.create(
+        group=moderators_group,
+        page=root,
+        permission_type='add',
+    )
+    GroupPagePermission.objects.create(
+        group=moderators_group,
+        page=root,
+        permission_type='edit',
+    )
+    GroupPagePermission.objects.create(
+        group=moderators_group,
+        page=root,
+        permission_type='publish',
+    )
+
+    GroupPagePermission.objects.create(
+        group=editors_group,
+        page=root,
+        permission_type='add',
+    )
+    GroupPagePermission.objects.create(
+        group=editors_group,
+        page=root,
+        permission_type='edit',
+    )
+
+
+def remove_initial_data(apps, schema_editor):
+    """This function does nothing. The below code is commented out together
+    with an explanation of why we don't need to bother reversing any of the
+    initial data"""
+    pass
+    # This does not need to be deleted, Django takes care of it.
+    # page_content_type = ContentType.objects.get(
+    #     model='page',
+    #     app_label='wagtailcore',
+    # )
+
+    # Page objects: Do nothing, the table will be deleted when reversing 0001
+
+    # Do not reverse Site creation since other models might depend on it
+
+    # Remove auth groups -- is this safe? External objects might depend
+    # on these groups... seems unsafe.
+    # Group.objects.filter(
+    #     name__in=('Moderators', 'Editors')
+    # ).delete()
+    #
+    # Likewise, we're leaving all GroupPagePermission unchanged as users may
+    # have been assigned such permissions and its harmless to leave them.
--- a/wagtail/wagtailcore/migration_tools/initial_data.py
+++ b/wagtail/wagtailcore/migration_tools/initial_data.py
@@ -1,4 +1,5 @@
 from django import VERSION as DJANGO_VERSION
+from django.conf import settings
 
 
 def create_initial_data(apps, schema_editor):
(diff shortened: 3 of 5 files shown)


akx:
Rebased on top of upstream/master.

akx:
Ah, actually, I found a little bug: the Wagtail migration code forces the root page's `numchild` to 1, assuming it will immediately get a child (that is has MP_Node properties set manually).

I'll have to fix that...

gasman:
Hi @akx,
Thanks! This change looks sensible enough to me (subject to the `numchild` fix being made, of course), but I've never tried loading initial site data through fixtures/migrations, so I'm not really clear about what it achieves in real terms. In other words, I'd like to understand:

* why the initial data is so annoying to deal with (if you're using fixtures, doesn't it get overwritten in the `./manage.py loaddata` step anyhow?)
* what a developer would need to change in their process to take advantage of this setting
* how they would benefit as a result

@jordij, @timheap: I'd be interested to hear your thoughts - does this change fit in with the way you're handling initial data?

One thing I'm wary of is adding a settings flag and calling it 'done'... that means we're really just multiplying the number of possible configurations we have to support, without making things better for the average user (who isn't going to read through the reference docs to find out about new features they can use). If this is indeed a better way of doing things - rather than just a *different* way - then we ought to reflect that in our tutorials and project template where appropriate. Basically, I don't want to fall into the `git` trap, where there are all kinds of powerful features available, but they're hidden behind obscure switches that only the 'power users' know about because nobody dared to step up with an opinion and say "this is the configuration that 99% of users should be using, so let's set it up that way out of the box" :-)

timheap:
There is a conflict here between new users and power users unfortunately. New users want a functioning Wagtail site out of the box. Providing them with the suggested Site, hope page, and some default user groups makes sense, and makes the initial set up very smooth and nice. Power users might want to create multiple sites, with many different home pages. The initial page structure may not work right for them. They might want a more complicated user and group set up. All this means that the initial sites, pages and groups get in the way, and have to be deleted before the real data gets created.

My preferred set up would be to have Wagtail create no initial data except for the root page of the tree. Not the home page, but the actual root page. No groups, no sites, no pages, nothing. Then, as part of the `wagtail start projectname` template, a migration should be created that creates these initial pieces of data for the user. New users would have all the initial data required to get started, while power users could customise this initial migration as much as they need to to install their own data - or remove the migration completely and create the initial data in another way.

Sadly, this proposal is quite difficult to achieve in a backwards-compatible manner, without having a flag new projects can set that indicates this initial data will be created by the project, which is partially  what this PR is about.

jordij:
I fully agree with Tim on this one

timorieber:
Agreed. Nothing more to say to Tim's explanation.

gasman:
So, if I've understood correctly - this PR could be turned into @timheap's ideal setup (or at least an approximation to it...) by:

* leaving the `WAGTAIL_MIGRATE_INITIAL_*` settings as True by default (to preserve backwards compatibility on existing projects);
* but setting them to False in project_template's settings (and presumably, in our 'integrating with Django' / 'configuring Django for Wagtail' docs, advising developers to do the same);
* and updating project_template's data migration (currently `home/migrations/0002_create_homepage.py`) to create the initial site / homepage / group records.

The settings would have to remain in the project template for the foreseeable future - so effectively we're in the same situation as Django's `USE_I18N` / `USE_L10N` / `USE_TZ` flags, which have been in everyone's settings files forever and rarely need to be changed, but can't be dropped due to ancient legacy reasons. Which would be unfortunate, but clearly the Django world has tolerated those redundant settings without too much heartache, so it's not the end of the world if we have to do the same here ;-)

However, it seems to me that all we've really gained in all of this is the opportunity to omit a few 'delete' lines from the 0002_create_homepage migration - in other words, it boils down to a choice between some extra noise in the project_template settings, or some extra noise in the project_template migration. And given that choice, I'd go for the noise in the migration - it's something that only power users need to look at, and if it was suitably well-commented, perhaps with sample commented-out code blocks that users could uncomment and adjust to fit their requirements ("to remove the default groups, do this; to load initial data from a fixture, do this...") then I could see that being a lot neater. (Admittedly, we'd have to give a similar treatment to the configuration / integration docs.)

As I said, I'm not fully clear on the motivation for this change, so I'm prepared to accept that there may be issues with the "big noisy migration" approach that I haven't grasped.

akx:
Rebased again, sorry for the delay.

As for my motivation -- well, @timheap kinda put it in words for me there :)

I find it silly that the [default project template's migration](https://github.com/torchbox/wagtail/blob/72d9ae81b5dbfa41b486382fc9d6339a5c273505/wagtail/project_template/home/migrations/0002_create_homepage.py#L14-L34) actually _deletes_ the objects another migration has created, just so it can create its own versions, with the correct content types etc.

We're evaluating Wagtail for a home page project here at @andersinno (and so far it's been pretty great, though I'm looking forward for proper i18n support, as projects here in Finland tend to be multilingual beasts -- but I digress), and for that project [we've been running a fork release that has this patch](https://github.com/andersinno/wagtail/releases/tag/v1.4a0_ai0) in, and our initial migration (though it could just as well be a "seed initial data" management command) looks like this. No silly deleting or editing of existing useless data. :)

```python
def create_homepage(apps, schema_editor):
    ContentType = apps.get_model('contenttypes.ContentType')
    Site = apps.get_model('wagtailcore.Site')
    HomePage = apps.get_model('superawesomeprojectyay.HomePage')
    homepage_content_type = ContentType.objects.get_for_model(HomePage)
    homepage = HomePage(title='Kotisivu', slug='home', content_type=homepage_content_type)
    root = Page.objects.get(slug='root')
    root.add_child(instance=homepage)
    Site.objects.create(hostname='localhost', root_page=homepage, is_default_site=True)
    ContentType.objects._cache.clear()  # See https://github.com/django/django/pull/6088
```

gasman:
I agree that the current behaviour is silly - no argument there :-)

(Sorry for the wall of text that follows, but it's kind of important...)

If we're making this change, it's absolutely essential that we roll out the new setting in a way that encourages all new users to use it - i.e. we enable it in the project template by default. For me, the worst possible outcome is that we end up with two competing conventions in use - the current "silly" behaviour, and the fixed behaviour that only power users know about - because then we end up with twice as many scenarios to test against, harder-to-diagnose tech support problems ("why are they getting that weird behaviour? Oh, they must have Obscure Setting X enabled..."), more complications if we change it again in future (see #1824 where we've tied ourselves in knots because a 'clever' migration had unforeseen side effects, and now we have two diverging configurations to deal with) - and just generally, more "moving parts" to account for. I fear that's what will happen if we merge this PR in its current form.

(Besides making life more difficult for us Wagtail developers, the "opt-in setting" approach doesn't really improve the developer experience for end users either. 95% of users won't read the reference docs, so they won't learn about the setting at all, and will be stuck with the "silly" behaviour. The other 5% won't understand the purpose of the setting, unless they've learnt a LOT about Wagtail internals, such as the difference between wagtailcore migrations and project_template migrations, and why having a raw Page instance for your homepage is a bad thing. *Nobody* is going to learn all those details for their first Wagtail project, so the only people who will benefit from this setting are people like you, who have been previously bitten by the "silly" behaviour, realised it's silly, and started digging into Wagtail internals for a better solution. That's not a good experience for anyone.)


So, unless anyone can see an alternative approach that addresses my concerns above, I propose that we add explicit `WAGTAIL_MIGRATE_INITIAL_PAGE_AND_SITE = False` / `WAGTAIL_MIGRATE_INITIAL_GROUPS = False` lines into project_template/project_name/settings/base.py. The question remains: if we do that, is this change still worthwhile, or does that completely nullify the goal of finding a more elegant solution? Personally, it seems like the latter to me: we no longer have the silly 'delete' operations sitting in migration 0002 forever, but instead we have some silly "don't change this unless you know what you're doing" lines sitting in the settings file forever. But then it's not really my judgment call to make - clearly you (@akx, @timheap, @jordi, @timorieber) have stronger feelings about the usefulness of this change than I do - if you think this PR is still worthwhile with my proposed amendment, I'll happily go along with it.

gasman:
I'll take the silence as a sign that there's no enthusiasm for my amended proposal, then :-)

Unfortunately I don't think we're going to find a solution to the 'ugly migrations' issue that's satisfactory to everyone here and doesn't just shift the ugliness somewhere else. As such, I'll close this issue (but as always, it can be reopened if there's further progress on this).

Jordi:
Maybe because the emails are mixed up again? I have no idea why i keep
receiving these kind of messages. I don't know why github accounts are
always mixed up. Have nothing to do with that project.
On Apr 8, 2016 9:18 PM, "Matt Westcott" <notifications@github.com> wrote:

Closed #2112 <https://github.com/torchbox/wagtail/pull/2112>.

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/torchbox/wagtail/pull/2112#event-620113746>

gasman:
@Jordi Oops, that was my fault... looks like I mentioned you instead of @jordij in my earlier message. Sorry for the spam!

(Github doesn't provide a way for me to unsubscribe you from notifications, as far as I can see - you can unsubscribe yourself at https://github.com/torchbox/wagtail/pull/2112)

jordij:
Oops sorry @gasman  I didn't get notified about what was going on. 

I agree with you on the_ " Nobody is going to learn all those details for their first Wagtail project, so the only people who will benefit from this setting are people like you, who have been previously bitten by the "silly" behaviour..."_  but to me this change is useful as I'd use the new behaviour in 99% of my Wagtail-based implementations.

Personally, I don't see it as an issue if it remains as a "hidden" setting that shouldn't be changed unless the developer really wants to. Django has several settings with a default value that remain kind of "hidden" until you start digging and learning about them. I feel this would be a similar case, but again this is just my opinion :)

dopry:
As is,  this PR looks fine. Backward compatible behavior is preserved by default. If a users wants to change the setting they can, and the caveat that it won't work with the default project seems to be clear in the docs. 

@gasman 

> If we're making this change, it's absolutely essential that we roll out the new setting in a way that encourages all new users to use it - i.e. we enable it in the project template by default. For me, the worst possible outcome is that we end up with two competing conventions in use - the current "silly" behaviour, and the fixed behaviour that only power users know about - because then we end up with twice as many scenarios to test against, harder-to-diagnose tech support problems ("why are they getting that weird behaviour? Oh, they must have Obscure Setting X enabled..."), more complications if we change it again in future (see #1824 where we've tied ourselves in knots because a 'clever' migration had unforeseen side effects, and now we have two diverging configurations to deal with) - and just generally, more "moving parts" to account for. I fear that's what will happen if we merge this PR in its current form.

1. It is not necessary to force all users to use this approach. The default behavior should preserve backward compatibility so typical users are unaffected.  Core doesn't want or need to support this option, and can clearly state as much in the documentation. If the community finds this functionality useful, it may make sense for core to revisit the question of support.  

> (Besides making life more difficult for us Wagtail developers, the "opt-in setting" approach doesn't really improve the developer experience for end users either. 95% of users won't read the reference docs, so they won't learn about the setting at all, and will be stuck with the "silly" behaviour. The other 5% won't understand the purpose of the setting, unless they've learnt a LOT about Wagtail internals, such as the difference between wagtailcore migrations and project_template migrations, and why having a raw Page instance for your homepage is a bad thing. Nobody is going to learn all those details for their first Wagtail project, so the only people who will benefit from this setting are people like you, who have been previously bitten by the "silly" behaviour, realised it's silly, and started digging into Wagtail internals for a better solution. That's not a good experience for anyone.)

2. I can say 100% an opt-in setting would improve my developer experience with wagtail. I've lost a half-day, just figuring out how to set up fixtures so that I can prep to kick off the front-end development on my project.  I've had to figure out our normal approach with fixtures doesn't work because someone chose to use a data migration to seed data in wagtail, rather than fixtures. Then I had to work around the fact the manage.py flush doesn't work and build that into our docs and docker-compose files. You acknowledge the problem and silliness of having the seed data embedded in a data migration to begin with, such as slowing down test setup and breaking the fixtures feature. 

> So, unless anyone can see an alternative approach that addresses my concerns above, I propose that we add explicit WAGTAIL_MIGRATE_INITIAL_PAGE_AND_SITE = False / WAGTAIL_MIGRATE_INITIAL_GROUPS = False lines into project_template/project_name/settings/base.py. The question remains: if we do that, is this change still worthwhile, or does that completely nullify the goal of finding a more elegant solution? Personally, it seems like the latter to me: we no longer have the silly 'delete' operations sitting in migration 0002 forever, but instead we have some silly "don't change this unless you know what you're doing" lines sitting in the settings file forever. But then it's not really my judgment call to make - clearly you (@akx, @timheap, @jordi, @timorieber) have stronger feelings about the usefulness of this change than I do - if you think this PR is still worthwhile with my proposed amendment, I'll happily go along with it.

I would reject your proposal based on your false premise that this change has to break backward compatibility.  I think it is perfectly okay to give power users some freedom.  Nothing changes in settings unless a user turns it on. Core. Take a wait an see approach and see if the support concerns are valid, or maybe the contributor community will develop a new practice core can adopt.  I'd say all the PR needs is clear documentation that core won't support this.

PR #2113, opened 2016-01-14
Admin View Sets

kaedroho:
Here is the next iteration of admin modules/viewsets (previous take here #2067).

The idea is to build the admin site up using "viewsets". These provide a url config, menu items, permissions and other things for Wagtail. Modules are all class based and are easy to subclass.

I think there's a few good reasons for modularising the admin interface like this:
- Reduces duplicated view code
- Makes it easier to add/remove/modify/duplicate sections of the admin
- Moves model view permission logic to a higher level (currently each view is responsible for permissions, which is a source of mistakes)
  - Also allows each component of a viewset (urls, menu item, dashboard panel, etc) to share permission logic

This PR adds modules into Wagtailadmin and refactors the sites section to use one.

To use admin viewsets to register a model in the admin, just add the following to a `wagtail_hooks.py` file:

``` python
# app/wagtail_hooks.py
from wagtail.wagtailadmin.viewsets.model import ModelViewSet

from .models import MyModel


@hooks.register('register_admin_viewset')
def register_admin_viewset():
    return ModelViewSet('mymodel', model=MyModel, icon='the-icon')
```

Changes from last PR:
- Renamed to viewsets
- Based on @gasman's "Permission policy Modules" #2102 
- Removed template name generation
- Added ``ViewSetRegistry`` class

TODO:
- Menu items
- Search areas
- Dashboard panels
- Admin API
- Choosers


auvipy:
awesome include!!!

ababic:
I especially love how you've handled the URL generation in this (using namespacing)... much tidier than what `ModelAdmin` does! Is this something that is still in the works, or is it very much on the back-burner now?

ababic:
Personally, I would love to be able to base the `ModelAdmin` class on `ModelViewSet` and see how much of ModelAdmin we could chip off and keep it working.

I guess one key thing you're doing differently here is passing in a bunch of variables to the view each time, rather than just passing the `ViewSet` instance to the view and letting the view use as much of the things from that as they need. That's one element of `modeladmin` that I think this could benefit in taking. 

In fact, I've been thinking recently about how I could possibly take that pattern a step further with ModelAdmin... having pretty much the identical code in `index_view()`, `add_view()` and `edit_view()` methods seems a little crazy to me now... surely there's a way to abstract that out and just make whatever kwargs are defined in the URL get passed through to the view automagically? :S I think that would make adding new views easier, too, if all you need to worry about was defining the `url` and adding a `something_view_class` attribute.

kaedroho:
Thanks @ababic!

``ViewSet`` is pretty much the same as modeladmin's ``WagtailRegisterable`` class so I reckon it would be easy to make modeladmin use viewsets. This would also allow choosers and admin API to be provided for modeladmin-registered models.

> or is it very much on the back-burner now?

Yes it is. The implementation as it stands at the moment and plan ahead haven't yet been endorsed (or even commented on) by anyone else on the core team. This would become a major overhaul of how Wagtail is structured so I would like to get feedback before getting any deeper to reduce the risk of having to undo a lot of work later on.

wagtail/wagtailadmin/viewsets/model.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def get_form_class(self, for_update=False):
+        return modelform_factory(
+            self.model,
+            formfield_callback=self.formfield_for_dbfield,
+            fields='__all__'
+        )
+
+    def get_urlpatterns(self):
+        return [
```
gasman: Just a thought... If we establish a convention of calling `super` here:

    def get_urlpatterns(self):
        return super(ModelViewSet, self).get_urlpatterns() + [
            ...
        ]

...then that would potentially allow us to use multiple inheritance as a way of plugging in additional views:

    class ImageViewSet(ModelViewSet, ChooserViewSet, APIViewSet):

wagtail/wagtailadmin/viewsets/model.py
```diff
@@ -0,0 +1,88 @@
...
+        )
+
+    def formfield_for_dbfield(self, db_field, **kwargs):
+        return db_field.formfield(**kwargs)
+
+    def get_form_class(self, for_update=False):
+        return modelform_factory(
+            self.model,
+            formfield_callback=self.formfield_for_dbfield,
+            fields='__all__'
```
gasman: Given how Django now strongly encourages specifying an explicit field list by dropping the old "default to all fields if no `fields` / `exclude` option is passed" behaviour, I think it would be appropriate for us to do the same - i.e. make it mandatory for ModelViewSet subclasses to either specify a `form_fields` property or override `get_form_class`. (Developers can of course still specify `form_fields = '__all__'` if they really want to bypass best practice!)

wagtail/wagtailadmin/viewsets/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def get_urlpatterns(self):
+        urlpatterns = []
+
+        for viewset in self.viewsets:
+            vs_urlpatterns = viewset.get_urlpatterns()
+
+            if vs_urlpatterns:
+                urlpatterns.append(url(
+                    r'^{}/'.format(viewset.name),
```
gasman: Using `viewset.name` as both the URL namespace and the URL path prefix means that in the case of wagtailsites, the URL has suddenly changed from `/admin/sites/` to `/admin/wagtailsites/`, which is something I think we should avoid.

I see `name` is assigned at the point that `ViewSet` is instantiated, in `register_admin_viewset` - presumably with the intention that different instances of the same viewset class could coexist in different URL namespaces - so it probably makes sense to add a new optional kwarg at that point:

    @hooks.register('register_admin_viewset')
    def register_viewset():
        return SiteViewSet('wagtailsites', url_prefix='sites')

with `url_prefix` defaulting to `self.name` if not specified.

gasman:
> Just adding a note-to-self that the sites index page here is currently (i.e. when rebased on latest master) failing to render the list of sites - will investigate.

This is because wagtailsites/index.html expects the list to be passed as `sites` rather than the default variable name `object_list` - `IndexView` lost the line `context_object_name = 'sites'` in the refactor. WIP branch with this + other fixes here: https://github.com/gasman/wagtail/commits/kaedroho-view-sets

ababic:
Hi guys. I'm currently working on some ideas for ModelAdmin as a solution for #3336 that would abstract out the repetitive code from the `ModelAdmin` class to something [like this `ModelAction` class ](https://github.com/ababic/waddleadmin/blob/master/waddleadmin/actions.py#L13). The relevant details for each 'action' (e.g. 'edit' / 'create' / 'delete') are just defined together as tuples [like this](https://github.com/ababic/waddleadmin/blob/master/waddleadmin/actions.py#L210). How would you guys feel about having this abstraction baked in at a generic / central level?

gasman:
Will close this now, in favour of #3647.

@ababic I think generic views / viewsets are too early in their evolution to be able to tell how things like action buttons will fit into the picture - at the moment the wagtailadmin/generic/index.html template provides basically nothing at all, and expects you to implement the listing markup from scratch each time. Hopefully once we've ported a few more apps to generic views, we'll be able to figure out the common patterns and work things out from there.

PR #1022, opened 2015-02-23
Allow specifying custom edit handlers for pages - fixes #210

gasman:
This PR allows page classes to specify an `edit_handler` property, to override the default Content / Promote / Settings tabbed interface.

It also updates the default layout so that empty tabs are omitted, as requested in #485 + #972.

PR #2115, opened 2016-01-15
Allow specifying custom edit handlers for snippets

m1kola:
Hello!

I think it would be great to give ability to specify `edit_handler` property for snippets. This PR is similar to #1022 for pages.

I'll finish this PR if you are ok with this kind of changes.

### DIFF
--- a/docs/advanced_topics/customisation/page_editing_interface.rst
+++ b/docs/advanced_topics/customisation/page_editing_interface.rst
@@ -1,5 +1,5 @@
-Customising the page editing interface
-======================================
+Customising the editing interface
+=================================
 
 .. _customising_the_tabbed_interface:
 
@@ -8,7 +8,7 @@ Customising the tabbed interface
 
 .. versionadded:: 1.0
 
-As standard, Wagtail organises panels into three tabs: 'Content', 'Promote' and 'Settings'. Depending on the requirements of your site, you may wish to customise this for specific page types - for example, adding an additional tab for sidebar content. This can be done by specifying an ``edit_handler`` property on the page model. For example:
+As standard, Wagtail organises panels for pages into three tabs: 'Content', 'Promote' and 'Settings'. For snippets Wagtail puts all panels into one page. Depending on the requirements of your site, you may wish to customise this for specific page types or snippets - for example, adding an additional tab for sidebar content. This can be done by specifying an ``edit_handler`` attribute on the page or snippet model. For example:
 
 .. code-block:: python
 
--- a/wagtail/wagtailsnippets/tests.py
+++ b/wagtail/wagtailsnippets/tests.py
@@ -8,7 +8,7 @@ from django.core.exceptions import ImproperlyConfigured
 from taggit.models import Tag
 
 from wagtail.tests.utils import WagtailTestUtils
-from wagtail.tests.testapp.models import Advert, SnippetChooserModel
+from wagtail.tests.testapp.models import Advert, SnippetChooserModel, AdvertWithTabbedInterface
 from wagtail.tests.snippets.models import (
     AlphaSnippet, ZuluSnippet, RegisterDecorator, RegisterFunction, SearchableSnippet
 )
@@ -131,6 +131,19 @@ class TestSnippetCreateView(TestCase, WagtailTestUtils):
         response = self.get()
         self.assertEqual(response.status_code, 200)
         self.assertTemplateUsed(response, 'wagtailsnippets/snippets/create.html')
+        self.assertNotContains(response, '<ul class="tab-nav merged">')
+        self.assertNotContains(response, '<a href="#advert" class="active">Advert</a>', html=True)
+        self.assertNotContains(response, '<a href="#other" class="">Other</a>', html=True)
+
+    def test_snippet_with_tabbed_interface(self):
+        response = self.client.get(reverse('wagtailsnippets:add',
+                                           args=('tests', 'advertwithtabbedinterface')))
+
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, 'wagtailsnippets/snippets/create.html')
+        self.assertContains(response, '<ul class="tab-nav merged">')
+        self.assertContains(response, '<a href="#advert" class="active">Advert</a>', html=True)
+        self.assertContains(response, '<a href="#other" class="">Other</a>', html=True)
 
     def test_create_invalid(self):
         response = self.post(post_data={'foo': 'bar'})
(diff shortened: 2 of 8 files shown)


kaedroho:
Thanks!

I'm guessing that the main advantage of this is it allows developers to use the tabbed interface with snippets? If so, I think that could be very useful!

@davecranwell and/or @gasman might want to be consulted first before any UI-related changes happen though.

m1kola:
> I'm guessing that the main advantage of this is it allows developers to use the tabbed interface with snippets?

Yes, you are absolutely right. I'll attach screenshots later, but it's very similar to pages tabbed interface when `edit_handler` is present in model class. When `edit_handler` is not present UI will be same as in Wagtail 1.3.1.

m1kola:
In wagtail 1.3 we can define panels for snippets only like this (`Advert` model form wagtaildemo project):
```
class Advert(models.Model):
    page = models.ForeignKey(
        'wagtailcore.Page',
        related_name='adverts',
        null=True,
        blank=True
    )
    url = models.URLField(null=True, blank=True)
    text = models.CharField(max_length=255)

    panels = [
        PageChooserPanel('page'),
        FieldPanel('url'),
        FieldPanel('text'),
    ]

    def __unicode__(self):
        return self.text
```
And it looks like this:
screen shot 2016-01-17 at 20 25 02

This PR gives ability to specify panels like shown before and like this;
```
class Advert(models.Model):
    page = models.ForeignKey(
        'wagtailcore.Page',
        related_name='adverts',
        null=True,
        blank=True
    )
    url = models.URLField(null=True, blank=True)
    text = models.CharField(max_length=255)

    advert_panels = [
        FieldPanel('url'),
        FieldPanel('text'),
    ]

    page_panels = [
        PageChooserPanel('page'),
    ]

        ObjectList(advert_panels, heading='Advert'),
        ObjectList(page_panels, heading='Page'),
    ])

    def __unicode__(self):
        return self.text
```
screen shot 2016-01-17 at 20 31 31

m1kola:
PR was updated: tests and docs were added.

I think this PR is ready to review.

tomdyson:
This looks really good, @m1kola - thanks a lot!

m1kola:
Hello! How about merge? :)

tomdyson:
@m1kola thanks again. @timheap do you have time to review this?

docs/advanced_topics/customisation/page_editing_interface.rst
```diff
@@ -8,7 +8,7 @@ Customising the tabbed interface
 
 .. versionadded:: 1.0
 
-As standard, Wagtail organises panels into three tabs: 'Content', 'Promote' and 'Settings'. Depending on the requirements of your site, you may wish to customise this for specific page types - for example, adding an additional tab for sidebar content. This can be done by specifying an ``edit_handler`` property on the page model. For example:
+As standard, Wagtail organises panels for pages into three tabs: 'Content', 'Promote' and 'Settings'. For snippets Wagtail puts all panels into one page. Depending on the requirements of your site, you may wish to customise this for specific page types or snippets - for example, adding an additional tab for sidebar content. This can be done by specifying an ``edit_handler`` property on the page or snippet model. For example:
```
timheap: > This can be done by specifying an ``edit_handler`` property on the page or snippet model

"property" should be "attribute" here. "property" in Python means a specific thing, and the example below uses an attribute, not a property. This was incorrect in the previous version of the documentation as well, but it may as well get fixed while you're editing in this area.

wagtail/wagtailsnippets/tests.py
```diff
@@ -131,6 +131,17 @@ def test_simple(self):
         response = self.get()
         self.assertEqual(response.status_code, 200)
         self.assertTemplateUsed(response, 'wagtailsnippets/snippets/create.html')
+        self.assertNotContains(response, '<a href="#advert" class="active">Advert</a>')
+        self.assertNotContains(response, '<a href="#other" class="">Other</a>')
```
timheap: Could you also assert that there is no `<ul class="tab-nav merged">` in the output?

wagtail/wagtailsnippets/tests.py
```diff
@@ -131,6 +131,17 @@ def test_simple(self):
         response = self.get()
         self.assertEqual(response.status_code, 200)
         self.assertTemplateUsed(response, 'wagtailsnippets/snippets/create.html')
+        self.assertNotContains(response, '<a href="#advert" class="active">Advert</a>')
```
timheap: You should pass `html=True` to `assertNotContains`, otherwise it does a plain-text comparison:

```python
self.assertNotContains(response, '<a href="#advert" class="active">Advert</a>', html=True)
```

This also applies to the `assertContains` calls below.

timheap:
The snippet create template has some extra spacing between the header and the tab bar. The edit template does not have this extra padding.

gap-in-tabbed-interface

timheap:
Other than the above comments, this looks good!

tomdyson:
Thanks, @timheap. @m1kola, are you happy to make these changes?

wagtail/wagtailsnippets/tests.py
```diff
@@ -131,6 +131,19 @@ def test_simple(self):
         response = self.get()
         self.assertEqual(response.status_code, 200)
         self.assertTemplateUsed(response, 'wagtailsnippets/snippets/create.html')
+        self.assertNotContains(response, '<ul class="tab-nav merged">')
```
m1kola: @timheap, `assertNotContains` and `assertContains` with `html=True` checks nodes, so even if `'<ul class="tab-nav merged">something</ul>'` in `response`
```
self.assertNotContains(response, '<ul class="tab-nav merged">', html=True)
```
Will not raise `AssertionError`.

Should I write asserts with `BeautifulSoup`? Or can you recommend something else?

m1kola:
@tomdyson and @timheap, thank you!

PR was updated, but [build job #5449.11](https://travis-ci.org/torchbox/wagtail/jobs/104650909) was failed. I think it's not my fault.

gasman:
Build job restarted now...

m1kola:
Thanks, @gasman!

tomdyson:
Hi @timheap, are you happy with @m1kola's changes? If so we'll merge.

wagtail/wagtailsnippets/tests.py
timheap: You're right, that does not work in the way I hoped it would. Ideally there would be a way to check that there is/is not a specific node in some HTML, while ignoring the contents of the node. That is what I hoped having an unclosed `<ul>` tag would do, but I was wrong. Don't worry about this test then.

timheap:
@tomdyson The changes look good to me. I say it is ready to merge!

tomdyson:
Thanks @timheap and @m1kola. @gasman, merge away!

m1kola:
Thank you, guys!

gasman:
Merged in 3f2f5665b0f7763476fea743ce05a17e58eb48ad - thanks!

SalahAdDin:
Very interesting.

PR #2119, opened 2016-01-16
Replaced SubFieldBase with Field.from_db_value

### LINKED ISSUE #2070: Eliminate use of `SubfieldBase` from StreamField
Some 3rd party django apps are using `SubfieldBase`, which are being depricated after Django 1.9 I believe.

Please try to get these fixed upstream.

kaedroho:
Fixes #2070 

Doing this fix broke the ability to assign a string or a list to a streamfield. It looks like this is behaviour we want to support so the second commit reinstates it by copying some code from Django.

### DIFF
--- a/wagtail/wagtailcore/fields.py
+++ b/wagtail/wagtailcore/fields.py
@@ -5,7 +5,7 @@ import json
 from django.db import models
 from django import forms
 from django.core.serializers.json import DjangoJSONEncoder
-from django.utils.six import with_metaclass, string_types
+from django.utils.six import string_types
 
 from wagtail.wagtailcore.rich_text import DbWhitelister, expand_db_html
 from wagtail.utils.widgets import WidgetWithScript
--- a/wagtail/wagtailcore/fields.py
+++ b/wagtail/wagtailcore/fields.py
@@ -41,7 +41,7 @@ class RichTextField(models.TextField):
         return super(RichTextField, self).formfield(**defaults)
 
 
-class StreamField(with_metaclass(models.SubfieldBase, models.Field)):
+class StreamField(models.Field):
     def __init__(self, block_types, **kwargs):
         if isinstance(block_types, Block):
             self.stream_block = block_types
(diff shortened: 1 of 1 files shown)


gasman:
I have a feeling that the allowed inputs are only as liberal as they are *because* the old SubfieldBase mechanism was so loosely defined - and if Django is tightening that up then it probably does make sense for us to follow suit, if possible.

With StreamField we do have the added complication that StreamValues can't exist in isolation (they have a reference back to the block object) so there may be places where we can't avoid passing in a primitive Python value - specifying a `default` parameter on StreamField is one, I think. Ultimately we probably need to bite the bullet, and trawl through the codebase checking the places where we do StreamField assignments...

kaedroho:
I reinstated the old behaviour because I found some tests for it in https://github.com/torchbox/wagtail/blob/d8bceff38be87b1802193caf8c3b854c435aab1a/wagtail/wagtailcore/tests/test_streamfield.py#L124-L133 so assumed you wanted to keep it.

If you don't want to reinstate the old behaviour, just ignore the second commit.

gasman:
Yep, understood. I think it's a job for me to dig in to the code and original design docs to figure out whether we do indeed want that behaviour, or whether that was only in there to replicate old-Django's behaviour.

timheap:
In our code, we assign strings to StreamFields when creating test data. We create a JSON structure with all the content for the field, and then `page.body = json.dumps(content)`. If StreamFields were to become more strict in what data they accept, how would we go about assigning this sort of generated StreamField content?

gasman:
@timheap I guess I've just answered this (in a roundabout way) in https://github.com/torchbox/wagtail/issues/2626#issuecomment-219690978 - I think the appropriate representation, assuming you don't want the extra faff of constructing a `StreamValue` object, is a list of (block_type, value) tuples.

(I guess one possible issue with that is that the values contained in that representation logically ought to be the 'friendly Pythonic' values rather than 'raw JSONish' ones - and the 'friendly Pythonic' values may well be less convenient to construct. For example, a PageChooserBlock's value is a page object rather than an ID, which would make the in-StreamField representation `('related_page', some_page_instance)` rather than `('related_page', 123)`.)

gasman:
Closed via #2736, in 480aca681f2c31e5083cf38ed367043fe6a78365 / 97b1a3b897b598e6c1fe861884b23a167a1d5eae.

PR #2121, opened 2016-01-18
Add python version to all Travis matrix options

zerolab:
Here is a small change to show the correct Python version in the Travis build job results, rather than "no language set"

We were already doing it for Python 3.5


kaedroho:
Thanks!

tomdyson:
Nice one @zerolab.

PR #2125, opened 2016-01-18
Fixes StreamValue unicode compatibility

### LINKED ISSUE #2057: Jinja2: UnicodeDecodeError with RichText in StreamField
Hello!

I am getting `UnicodeDecodeError` while rendering `post.body` when I put unicode (cyrillic) characters to RichText in StreamField. Meanwhile in admin interface everything is ok. 
```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xd0 in position 55: ordinal not in range(128)
```

Seems like broken serialization. `post.body` in database (PostgreSQL 9.4.5):
```
select * from blog_blogpage where page_ptr_id = 5;
-[ RECORD 1 ]-+--------------------------------------------------------------------------------------------------
page_ptr_id   | 5
subtitle      | Саб тайтл тест
body          | [{"type": "rich_text", "value": "<p>\u0411\u0430\u0434\u0438 \u0442\u0435\u0441\u0442<br/></p>"}]
lead          | <p>Лид тест</p>
main_image_id | 1
```

If body just RichTextField, everything works fine.

I can reproduce this on following versions of wagtail: 1.2, 1.3rc1

m1kola:
Fixes #2057

### DIFF
--- a/wagtail/wagtailcore/blocks/stream_block.py
+++ b/wagtail/wagtailcore/blocks/stream_block.py
@@ -309,7 +309,7 @@ class StreamValue(collections.Sequence):
         return repr(list(self))
 
     def __html__(self):
-        return self.__str__()
+        return self.stream_block.render(self)
 
     def __str__(self):
-        return self.stream_block.render(self)
+        return self.__html__()
--- a/wagtail/wagtailcore/tests/test_streamfield.py
+++ b/wagtail/wagtailcore/tests/test_streamfield.py
@@ -1,3 +1,6 @@
+# -*- coding: utf-8 -*
+from __future__ import unicode_literals
+
 import json
 
 from django.apps import apps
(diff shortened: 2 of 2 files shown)


gasman:
@m1kola Thanks! We'll need to do a bit of digging into Django string handling to verify whether this is correct - normally we'd want `__html__` to return a Django SafeString object so that HTML escaping in templates works as expected, and I'm not sure that will happen if we cast it to `six.text_type`.

Ideally, the test here should check the end-to-end behaviour we're interested in - such as inserting a non-ascii string into the field and confirming that it can be read back - not just checking an internal detail.

m1kola:
@gasman, I updated PR. I'm not sure that first solution was absolutely wrong, but this version is better.

Problem was in call `StreamValue.__str__()` inside `StreamValue.__html__` method. I think this should work on Python>=3, but for Python>=2,<3 `python_2_unicode_compatible` decorator doing this:

```python
        klass.__unicode__ = klass.__str__
        klass.__str__ = lambda self: self.__unicode__().encode('utf-8')
```

So `StreamValue.__str__()` should return `SafeText` (because `self.stream_block.render(self)` returns `SafeText`), but after applying of `python_2_unicode_compatible` decorator `StreamValue.__str__()` returns `SafeBytes`. `StreamValue.__html__` returns `SafeBytes` too (because it calls `StreamValue.__str__()`).

I decided to change calls: now `StreamValue.__str__()` calls `StreamValue.__html__()`. Also we can do something like that:
```python
@python_2_unicode_compatible  # provide equivalent __unicode__ and __str__ methods on Py2
class StreamValue(collections.Sequence):

    # ...

    def __html__(self):
        return self.stream_block.render(self)

    def __str__(self):
        return self.stream_block.render(self)
```

This works on Python 2.* with Django templates because Django template engine doesn't use `__html__` method for rendering.

About tests. End-to-end tests already implemented (see `TestStreamFieldDjangoRendering` and `TestStreamFieldJinjaRendering`). in this update I added `from __future__ import unicode_literals` import and non-ascii symbols to test data.

What did you think about this version?

m1kola:
@gasman, do you have any comments? Can we merge this PR?

gasman:
Merged in 1e82450160a3dc871c21c8d705e6f85d4ac6bc5a - sorry it took so long, and thanks for your patience!

tomdyson:
Thanks @m1kola!

m1kola:
Thank you, guys!

PR #2126, opened 2016-01-18
Rewrite group page permissions form to use checkboxes

### LINKED ISSUE #1488: Improve page-level group permission UI/explanation

gasman:
Fixes #1488. All permission types for a particular page are now displayed as a row of checkboxes, avoiding the impression given by the old dropdown interface that the permission types were mutually exclusive:

screen shot 2016-01-18 at 23 28 29

This also serves as (yet more) foundational work for collections - collection-based permissions will use a similar interface, since it's essential that we keep the UI for image / document permissions as close as possible to the pre-collections version (where images/documents are part of the 'Object Permissions' block).

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -1438,11 +1438,16 @@ class PageRevision(models.Model):
         verbose_name_plural = _('page revisions')
 
 
+PAGE_PERMISSION_TYPES = [
+    ('add', _("Add"), _("Add/edit pages you own")),
+    ('edit', _("Edit"), _("Edit any page")),
+    ('publish', _("Publish"), _("Publish any page")),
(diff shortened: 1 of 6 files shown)


davecranwell:
Excellent! It's a far nicer interface :)

gasman:
Merged via #2243.

PR #1877, opened 2015-10-28
Split embed finders into separate modules

kaedroho:
This pull request cleans up the ``wagtailembeds/embeds.py`` module by moving the "finders" into a separate folder.

PR #2127, opened 2016-01-19
Class based embed finders and new embed finder configuration

kaedroho:
Based on #1877

This pull request implements class based embed finders and a new embed finder configuration. I hope this will make embed finders more configurable solving the following problems:
- Allow additional oembed providers to be used (currently all are hardcoded)
- Make per-provider configuration possible
  - Choose which finders get used for which provider (eg, force use of oembed for instagram but use embedly for everything else)
  - Allow specifying additional parameters for certian providers such as `scheme=https` to YouTube

For example, the below configuration would use oembed for fetching youtube videos and embedly for everything else. All youtube videos would be fetched with `scheme=https`. It also adds a custom oembed provider too.

``` python
from wagtail.wagtailembeds.oembed_providers import youtube, all_providers

ISSUU = {
    'endpoint': 'http://issuu.com/oembed',
    'urls': [
        '^http(?:s)?://(?:www\\.)?issuu\\.com/[^#?/]+/docs/.+$',
    ]
}

WAGTAILEMBEDS_FINDERS = [
    {  # Override Youtube oEmbed provider to pass through an extra option
        'class': 'wagtail.wagtailembeds.finders.oembed',
        'providers': [youtube],
        'options': {'scheme': 'https'}
    },
    {  # Then try all other builtin oEmbed providers
        'class': 'wagtail.wagtailembeds.finders.oembed'
    },
    {  # Custom oEmbed provider
        'class': 'wagtail.wagtailembeds.finders.oembed',
        'providers': [ISSUU]
    },
    {  # Fallback to Embedly
        'class': 'wagtail.wagtailembeds.finders.embedly',
        'key': 'mykey'
    },
]
```

### DIFF
--- a/docs/advanced_topics/embeds.rst
+++ b/docs/advanced_topics/embeds.rst
@@ -1,0 +1,348 @@
+================
+Embedded content
+================
+
+Wagtail has builtin functionality for generating embeds for content on external
+websites such as Vimeo, YouTube and Soundcloud. The basic idea is to take a URL
+to a piece of content and generate a HTML embed for that content that can be
+nested on the page.
+
+The way these HTML embeds are generated is completely configurable. The default
+method is to query the site's `oEmbed`_ endpoint if it has one and Wagtail also
+supports `Embedly`_ and writing custom "embed finders" as well. These finders
+can also be chained together.
+
+Embedding content on your site
+==============================
+
+There are a few different ways the embeds module can be called.
+
+Rich text
+---------
+
+Wagtail's default rich text editor has a "media" icon that places embeds
+directly into rich text. You don't have to do anything to enable this, just
+make sure the rich text field's content is being passed through the
+``|richtext`` filter in the template.
+
+``EmbedBlock`` StreamField block type
+-------------------------------------
+
+The :class:~`wagtail.wagtailembeds.block.EmbedBlock` block type allows embeds
+to be placed into a ``StreamField``.
+
+For example:
+
+.. code-block:: python
+
+    from wagtail.wagtailembeds.blocks import EmbedBlock
+
+    class MyStreamField(blocks.StreamBlock):
+        ...
+
+        embed = EmbedBlock()
+
+
+``{% embed %}`` tag
+-------------------
+
+Syntax: ``{% embed <url> [max_width=<max width>] %}``
+
+You can nest embeds directly in a template by passing the URL and an optional
+max width argument to the ``{% embed %} tag. The ``max_width`` is sent in the
+query when fetching the embed HTML so it can have an effect on the generated
+HTML.
+
+.. code-block:: html+Django
+
+    {% load wagtailembeds_tags %}
+
+    {# Embed a YouTube video #}
+    {% embed 'https://www.youtube.com/watch?v=SJXMTtvCxRo' %}
+
+    {# This tag also take the URL from a variable #}
+    {% embed page.video_url %}
+
+From Python
+-----------
+
+You can also call the internal ``get_embed`` function that takes a URL string
+and returns an ``Embed`` object (see model documentation below). This also
+takes a ``max_width`` keyword argument that is sent in the query when fetching
+the embed HTML so it can have an effect on the generated HTML.
+
+.. code-block:: python
+
+    from wagtail.wagtailembeds.embeds import get_embed
+    from wagtail.wagtailembeds.exceptions import EmbedException
+
+    try:
+        embed = get_embed('https://www.youtube.com/watch?v=SJXMTtvCxRo')
+
+        print(embed.html)
+    except EmbedException:
+        # Cannot find embed
+        pass
+
+Configuring embed "finders"
+===========================
+
+Embed finders are the modules within Wagtail that generate the embed HTML from
+a URL.
+
+Embed finders are configured using the ``WAGTAILEMBEDS_FINDERS`` setting. This
+is a list of finder configurations that are each run in order until one of them
+successfully returns an embed:
+
+The default configuration is:
+
+.. code-block:: python
+
+    WAGTAILEMBEDS_FINDERS = [
+        {
+            'class': 'wagtail.wagtailembeds.finders.oembed'
+        }
+    ]
+
+.. _oEmbed:
+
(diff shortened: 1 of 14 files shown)


balazs-endresz:
How would we embed for example an iframe widget from a specific site that has no oembed support? So, one would enter the url, and Wagtail would need to turn that into an iframe somehow. Could I do that with a custom embed finder?

kaedroho:
@balazs-endresz yep, you just need to subclass ``EmbedFinder`` and implement the ``accept(url)`` (returns ``True`` if the finder can handle the passed URL but does not perform a request) and ``find_embed(url, max_width)`` (fetches the embed and returns its contets) methods.

You can then add it to WAGTAILEMBEDS_FINDERS like so:

```python
WAGTAILEMBEDS_FINDERS = [
    {
        'finder': 'mysite.myapp.finders.CustomEmbedFinder',
        # Any extra arguments here will be passed as keyword arguments to the embed finder's constructor
    }
]
```

I'll write up some docs about that when I get a moment

kaedroho:
Rebased but not had a chance to document this yet

kaedroho:
Docs done, ready for review!

docs/advanced_topics/embeds.rst
```diff
@@ -0,0 +1,348 @@
+================
+Embedded content
+================
+
+Wagtail has builtin functionality for generating embeds for content on external
```
tomdyson: can we have builtin -> built-in throughout, please?

docs/advanced_topics/embeds.rst
```diff
@@ -0,0 +1,348 @@
...
+websites such as Vimeo, YouTube and Soundcloud. The basic idea is to take a URL
+to a piece of content and generate a HTML embed for that content that can be
+nested on the page.
+
+The way these HTML embeds are generated is completely configurable. The default
+method is to query the site's `oEmbed`_ endpoint if it has one and Wagtail also
```
tomdyson: "one and Wagtail" -> "one. Wagtail"

docs/advanced_topics/embeds.rst
```diff
@@ -0,0 +1,348 @@
...
+supports `Embedly`_ and writing custom "embed finders" as well. These finders
```
tomdyson: don't need "as well" as well as "also" :)

docs/advanced_topics/embeds.rst
```diff
@@ -0,0 +1,348 @@
...
+can also be chained together.
```
tomdyson: can also be -> can be

docs/advanced_topics/embeds.rst
```diff
@@ -0,0 +1,348 @@
...
+        embed = EmbedBlock()
+
+
+``{% embed %}`` tag
+-------------------
+
+Syntax: ``{% embed <url> [max_width=<max width>] %}``
+
+You can nest embeds directly in a template by passing the URL and an optional
+max width argument to the ``{% embed %} tag. The ``max_width`` is sent in the
```
tomdyson: missing `` after %}

docs/advanced_topics/embeds.rst
```diff
@@ -0,0 +1,348 @@
...
+https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailembeds/oembed_providers.py
+
+Customising the provider list
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can limit which providers may be used by specifying the list of providers
+in the finder configuration.
+
+For example, this configuration will only allow content to be nested from Vimeo
+and Youtube. It also adds a custom provider in as well:
```
tomdyson: drop " in as well"

kaedroho:
@tomdyson tweaks made, thanks!

SalahAdDin:
Do this make deprecated packages functionality like `wagtail-embed-videos`?

SalahAdDin:
The same question.

kaedroho:
@gasman Rebased, made suggested changes and also tweaked the docs bit. Ready for another review!

wagtail/wagtailembeds/embeds.py
```diff
@@ -15,7 +15,11 @@ def get_embed(url, max_width=None, finder=None):
 
     # Get/Call finder
     if not finder:
-        finder = get_default_finder()
+        def finder(url, max_width=None):
+            for finder in get_finders():
+                if finder.accept(url):
+                    return finder.find_embed(url, max_width=max_width)
+
     embed_dict = finder(url, max_width)
```
gasman: Entering a URL that's not recognised by any finder now fails here with `TypeError: 'NoneType' object does not support item assignment` (which suggests we're missing a test for this). This happens because the old code threw an `EmbedNotFoundException` from the finder (which was then caught by the UI code) - in the new code, the `finder` function above exits with `None`, having not found a suitable finder (and then breaks when we try to assign `embed_dict['width']`).

On the flip-side of this: if a finder *does* accept a URL but subsequently throws `EmbedNotFoundException`, we won't proceed to the next finder in the list - in particular, this means that any finder listed after `wagtail.wagtailembeds.finders.embedly` will never take effect.

kaedroho: > On the flip-side of this: if a finder does accept a URL but subsequently throws EmbedNotFoundException, we won't proceed to the next finder in the list - in particular, this means that any finder listed after wagtail.wagtailembeds.finders.embedly will never take effect.

This is the expected and documented behaviour. If a finder is capable of retrieving an embed from a website but can't find it, it's unlikely that any other finder would have much luck. This prevents the module from making many HTTP requests (which block the server thread) to find embeds that don't exist.

If the wrong finder is being chosen due to misconfiguration, then the user should fix their configuration. If the user really does' want to check two finders (which I don't think is a common use case), they can write a custom finder class.

kaedroho:
@gasman I've switched to use your rebased version and fixed https://github.com/wagtail/wagtail/pull/2127#discussion_r124547737

gasman:
Merged in 01e2d5f4a0e94fcb5fdaee0dc8ea93455b6a00ab - thanks @kaedroho!

PR #2131, opened 2016-01-19
should use pk instead of id

alexpilot11:
When a model uses a primary key for something besides id, it breaks. Using pk fixes this problem because pk will still reference id if id is the primary key


gasman:
Merged in 5270c14444b5ada7a60c0b644fcc43d456c10a2d - thanks!

Issue #2124, opened 2016-01-18
Ditch django_compressor in favour of a Node-based bundling setup


gasman:
As of version 2.0, django_compressor now depends on the packages `rjsmin` and `rcssmin`, which include native extensions and are thus non-trivial to install on Windows (see https://groups.google.com/d/msg/wagtail/-CDQcHk5F9M/aL_DbhjdCQAJ). Additionally, there have been reports of bugs in rjsmin's compression (specifically, SimpleMDE uses the regex `/^ *-+:*$/` which rjsmin incorrectly compresses to `/^*-+:*$/`), and our current django_compressor setup doesn't allow specific addons/apps using the `insert_editor_js` hook to opt out of compression. Finally, it's an extra thing that end users of Wagtail need to configure (or at least, be aware of) when deploying.

The main reason for using django_compressor (SASS compilation) has already been superseded by our Node toolchain, and minification/bundling is something else that could certainly be done just as well (or better) through Node. The one thing we'd lose would be the dynamic 'discovery' of bundled assets based on `{% compress %}` tags - and thus the ability for third-party modules to insert themselves into those bundles via `insert_editor_js` - but it's probably a good thing for maintainability for our bundles to be defined statically in a gulp config (or similar), and for third-party modules to deal with minification themselves if they consider it important.

davecranwell:
See also #1143

spapas:
This may be a little out of date, however, for people that would still like to use django-compressor in Windows, I'd like to point out that you can easily install it using the ``--install-option="--without-c-extensions"`` pip parameter:

```
(venv) C:\progr\py\test>pip install --install-option="--without-c-extensions" django-compressor
You are using pip version 6.1.1, however version 8.0.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
Collecting django-compressor
  Using cached django_compressor-2.0-py2.py3-none-any.whl
Collecting django-appconf>=0.4 (from django-compressor)
  Using cached django_appconf-1.0.1-py2.py3-none-any.whl
Collecting rcssmin==1.0.6 (from django-compressor)
  Using cached rcssmin-1.0.6.tar.gz
Collecting rjsmin==1.0.12 (from django-compressor)
  Using cached rjsmin-1.0.12.tar.gz
Collecting six (from django-appconf>=0.4->django-compressor)
  Using cached six-1.10.0-py2.py3-none-any.whl
[...]

PR #2132, opened 2016-01-20
Remove dependency on django compressor

kaedroho:
Refs #2124 

This pull request removes the dependency on django-compressor. We've previously used it for bundling CSS and JavaScript but the 2.0 release adds some dependencies with compiled extensions which make it (and Wagtail) more difficult to install.

This should also remove the need to run ``django-admin compress`` on every deploy unless the project itself uses Django compressor.


tomdyson:
Hi @timheap, are you able to look at this one?

balazs-endresz:
The problem with this approach is that the static file urls will be the same after their content is changed, which means you couldn't use a far future Expires header. Using ManifestStaticFilesStorage or CachedStaticFilesStorage should be able to help with this:
https://docs.djangoproject.com/en/1.9/ref/contrib/staticfiles/#django.contrib.staticfiles.storage.ManifestStaticFilesStorage
But that would apply to the whole site not just to wagtailadmin.

timheap:
I found some more references to django-compressor, in the docs, docstrings, and comments that need removing:

```
docs/reference/hooks.rst:231:  Add additional Javascript files or code snippets to the page editor. Output must be compatible with ``compress``, as local static includes or string.
docs/reference/hooks.rst:260:  Add additional CSS or SCSS files or snippets to the page editor. Output must be compatible with ``compress``, as local static includes or string.
docs/advanced_topics/performance.rst:28:Without a persistent cache, Wagtail will recreate all compressible assets at each server start, e.g. when any files change under ``./manage.py runserver``.
wagtail/wagtaildocs/static/wagtaildocs/css/add-multiple.css:3:STATIC_URL setting. However, django-compressor consistently places the compiled
wagtail/wagtailadmin/static_src/wagtailadmin/scss/_variables.scss:4:// STATIC_URL setting. However, django-compressor consistently places the compiled
wagtail/tests/settings.py:138:COMPRESS_ENABLED = False  # disable compression so that we can run tests on the content of the compress tag
```

I have not had a chance to test this with the ManifestCacheMiddleware yet. If it works with that, and the above references are removed, then I am happy for this to be merged.

JoshBarr:
A quick question... How do we intend to transform/minify JS for the admin in lieu of `{% compress %}`? Is the plan to bundle/minify everything offline via `gulp` or similar and manually include references to the built assets in the templates? 

Today I was looking at an interesting project, [django-webpack-loader](https://github.com/owais/django-webpack-loader) for integrating Webpack bundles with Django. I've got it set up with Wagtail, but didn't manage to push it while GitHub was down... We would need some of this infrastructure in place to transform React components for the admin and split up the JS into re-usable modules.

davecranwell:
From the discussions we've had recently, I don't believe we plan to minify _any_ JS or CSS for the admin. The thinking was that minification is an optimisation best left to the implementor, so they can choose their solution rather than have one enforced. In Torchbox's case for example, we'll almost certainly continue to use django-compressor in most of our production projects - _we've_ never had any trouble with it.

Separately however, @kaedroho has been looking into rebundling of some of our JS to avoid some needless repetition. Webpack or browserify were considered but I'm not sure the outcome there.

gasman:
As @davecranwell says, the plan *for now* is to leave the admin JS and CSS un-minified. However, it's not really a case of 'leave it up to the implementor' - in the absence of the `{% compress %}` tags it won't be possible for implementors to take advantage of django-compressor for the admin even if they do want to (i.e. even if they have it enabled for their front-end).

For that reason we probably do want to continue looking at adding minification / bundling to our gulp toolchain - but that's not as urgent as dropping django-compressor from our requirements.

balazs-endresz:
About the versioning: django-cms solves that by appending only the CMS version to the url:

https://github.com/divio/django-cms/blob/develop/cms/templates/cms/wizards/base.html#L16
https://github.com/divio/django-cms/blob/a027f080b0c5a65f12f3879e389063f37c32a4d0/cms/templatetags/cms_static.py#L30
https://github.com/divio/django-cms/blob/a027f080b0c5a65f12f3879e389063f37c32a4d0/cms/utils/urlutils.py#L63

That looks even less intrusive than adding `ManifestStaticFilesStorage` to the project template. And it wouldn't require adding that in for (some) existing wagtail sites.

davecranwell:
https://css-tricks.com/strategies-for-cache-busting-css/ which is quite possibly out of date, suggests that query string cache busters aren't understood by some proxies. We may or may not want to consider that.

balazs-endresz:
That might be still a valid point about the proxies. And actually, this was added in to django-cms only a month ago: https://github.com/divio/django-cms/pull/4874#issuecomment-167826521

SalahAdDin:
Why will you deprecated this requirement?

kaedroho:
@SalahAdDin please read the description at the top

gasman:
@kaedroho to update all `{% load static %}` tags to ensure that we're compatible with ManifestStaticFilesStorage.

kaedroho:
Rebased

timheap:
This looks good. Combined with #2192, this is compatible with ManifestStaticFilesStorage, which was my only outstanding concern.

gasman:
Merged in daba4aff841fa312bb47615dc85e33a3ddad06bc + parents, along with a documentation update 20b42ad7d9644de730441a60c2f8249e6c4a162f.

timheap:
Sorry, I had another look for references to django-compressor, and I found some more that were missed:

```
# Results trimmed slightly to remove false positives
$ ack-grep -i 'compress' --ignore-dir venv --ignore-dir .tox --ignore-dir node_modules
wagtail.egg-info/requires.txt:2:django-compressor>=1.4
docs/reference/hooks.rst:231:  Add additional Javascript files or code snippets to the page editor. Output must be compatible with ``compress``, as local static includes or string.
docs/reference/hooks.rst:260:  Add additional CSS or SCSS files or snippets to the page editor. Output must be compatible with ``compress``, as local static includes or string.
docs/advanced_topics/performance.rst:28:Without a persistent cache, Wagtail will recreate all compressible assets at each server start, e.g. when any files change under ``./manage.py runserver``.
wagtail/wagtailadmin/static_src/wagtailadmin/scss/_variables.scss:4:// STATIC_URL setting. However, django-compressor consistently places the compiled
wagtail/tests/settings.py:138:COMPRESS_ENABLED = False  # disable compression so that we can run tests on the content of the compress tag
etc/nginx.conf.sample:19:    # Enable content compression for text types.
CHANGELOG.txt:8: * Wagtail no longer depends on django-compressor
```

PR #2133, opened 2016-01-20
Clarify what the Administrator flag does

fozcode:
This is something that confused me a little as a Wagtail newcomer.

In user admin the Administrator flag actually sets the `is_superuser` field, which then goes on to shortcut all the normal permissions checking. So saying that it gives _"the ability to manage user accounts"_ is rather an understatement. For example, a user with very little privilege will suddenly be able to delete sites, groups, documents and more as soon as this is ticked.

This change just modifies the help text for the Administrator field. You could argue that the group membership list should be disabled when `is_superuser` is `True`, but in my opinion that isn't necessary.

### DIFF
--- a/wagtail/wagtailusers/forms.py
+++ b/wagtail/wagtailusers/forms.py
@@ -53,7 +53,7 @@ class UserCreationForm(UsernameForm):
     is_superuser = forms.BooleanField(
         label=_("Administrator"),
         required=False,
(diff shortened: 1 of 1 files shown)


gasman:
Thanks! You're right, that help text wasn't very descriptive - I can only assume that it's remained unchanged since the very early days of Wagtail when we had no real permission model to speak of :-)

Merged in e0a7a472e62aca7b08b61178b92d65f637566e80.

PR #2134, opened 2016-01-20
Fix oembed regex pattern for Wistia

jlhurd:
This pull request fixes the regex pattern for identifying Wistia videos with oembed in the wagtail.wagtailembeds app. The old regex pattern had a typo in it, preventing it from matching any and all Wistia videos. Here is the current version:
```
^http://[-\\w]+\\.wista\\.com/medias/.+$
```
and here is the proposed new version, derived from [Wistia's recommended regex pattern](http://wistia.com/doc/construct-an-embed-code#the_regex):
```
^https?:\/\/(.+)?(wistia.com|wi.st)\/(medias|embed)\/.+$
```
This new version also includes Wistia's short embed code and different URL patterns. 

I hope I did this right! This is my first GitHub pull request. Just let me know if I need to submit this differently.

### DIFF
--- a/wagtail/wagtailembeds/oembed_providers.py
+++ b/wagtail/wagtailembeds/oembed_providers.py
@@ -152,7 +152,7 @@ OEMBED_ENDPOINTS = {
         "^http://www\\.funnyordie\\.com/videos/.+$"
     ],
     "http://fast.wistia.com/oembed.{format}": [
-        "^https?:\/\/(.+)?(wistia.com|wi.st)\/(medias|embed)\/.+$"
+        "^https?://([^/]+\.)?(wistia.com|wi.st)/(medias|embed)/.+$"
     ],
     "http://www.ustream.tv/oembed": [
         "^http(?:s)?://(?:www\\.)?ustream\\.tv/.+$",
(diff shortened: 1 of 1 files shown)


timheap:
Python regexs work a little different to the ones provided on the Wistia page, so copying them directly in to Python unfortunately leaves you with a broken regex. Specifically, you do not need to escape forward slashes with a backslash (`\/`) in Python. The regex provided by Wistia, corrected for Python, would be: `"^https?://(.+)?(wistia.com|wi.st)/(medias|embed)/.+$"`.

Unrelated to this PR, that regex provided by Wistia is very, very permissive, and would match a URL like `http://dangerous-hax.com/my-exploit-script/wistia.com/medias/something`. This would possibly reveal information about the web server to a malicious user, and allow a malicious user to embed their own content on the compromised site. A more restrictive regex that only allows for Wistia domains and subdomains would be: `"^https?://([^/]+\.)?(wistia.com|wi.st)/(medias|embed)/.+$"`

jlhurd:
Good catch with the permissiveness of Wistia's regex, @timheap, and thanks for cleaning up the Python. I updated the commit with the revised pattern that you recommended, `"^https?://([^/]+\.)?(wistia.com|wi.st)/(medias|embed)/.+$"`

kaedroho:
Merged in a96db0e30b09c06248aa88586b837b697e73ea47. Thanks @jlhurd and @timheap!

Issue #1536, opened 2015-07-20
The project template should list INSTALLED_APPS in precedence order


pjdelport:
Currently, the [project template's INSTALLED_APPS](https://github.com/torchbox/wagtail/blob/v1.0/wagtail/project_template/project_name/settings/base.py#L26-L52) lists the default apps from most generic to most specific, but this is the inverse of the order Django expects them in: apps listed earlier override the resources of apps listed later.

A better default value for this setting might be:

```python
INSTALLED_APPS = [
    'home',
    'search',

    'wagtail.wagtailforms',
    'wagtail.wagtailredirects',
    'wagtail.wagtailembeds',
    'wagtail.wagtailsites',
    'wagtail.wagtailusers',
    'wagtail.wagtailsnippets',
    'wagtail.wagtaildocs',
    'wagtail.wagtailimages',
    'wagtail.wagtailsearch',
    'wagtail.wagtailadmin',
    'wagtail.wagtailcore',

    'modelcluster',
    'compressor',
    'taggit',

    'django.contrib.admin',
    'django.contrib.auth',
    'django.contrib.contenttypes',
    'django.contrib.sessions',
    'django.contrib.messages',
    'django.contrib.staticfiles',
]
```

gasman:
Does the current ordering cause any actual problems? If not, I would prefer to stick to the more familiar arrangement (see for example https://docs.djangoproject.com/en/1.8/intro/tutorial01/#activating-models ), as that avoids the impression that Wagtail is doing anything non-standard.

pjdelport:
It will at least cause problems as soon as users try to override any templates the way many other Django apps expect them to: instead of working, the templates will just silently get ignored, which is a frustrating failure mode for newcomers to have to struggle with.

Fixing this should not give any impression that Wagtail is being non-standard: the Django docs ([link](https://docs.djangoproject.com/en/1.8/ref/templates/api/#django.template.loaders.app_directories.Loader)) are quite specific about the order INSTALLED_APPS should be listed in:

> The order of `INSTALLED_APPS` is significant! For example, if you want to customize the Django admin, you might choose to override the standard `admin/base_site.html` template, from django.contrib.admin, with your own `admin/base_site.html` in `myproject.polls`. You must then make sure that your `myproject.polls` comes before `django.contrib.admin` in `INSTALLED_APPS`, otherwise `django.contrib.admin`’s will be loaded first and yours will be ignored.

[...]

PR #2135, opened 2016-01-20
Update settings in 'configuring Django' docs for Django 1.9

gasman:
* Update the example settings file for Django 1.9 (including converting tuples to lists, and replacing TEMPLATE_CONTEXT_PROCESSORS with TEMPLATES)
* Set USE_L10N to true (the note about displaying localized dates was specific to FriendlyDateInput, which was removed in Wagtail 0.4 - the current AdminDateInput consistently uses numeric dates regardless of the USE_L10N setting)
* Remove the "Other Django Settings Used by Wagtail" list, as it seems to be an entirely arbitrary list - pretty much by definition, *any* Django setting could influence the behaviour of a Wagtail project.

### DIFF
--- a/docs/advanced_topics/settings.rst
+++ b/docs/advanced_topics/settings.rst
@@ -57,52 +57,34 @@ Apps (``settings.py``)
 
 .. code-block:: python
 
-  INSTALLED_APPS = (
-    'django.contrib.auth',
-    'django.contrib.contenttypes',
-    'django.contrib.sessions',
-    'django.contrib.messages',
-    'django.contrib.staticfiles',
+  INSTALLED_APPS = [
+
+    'myapp',  # your own app
+
+    'wagtail.wagtailforms',
+    'wagtail.wagtailredirects',
+    'wagtail.wagtailembeds',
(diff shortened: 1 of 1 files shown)


docs/advanced_topics/settings.rst
```diff
@@ -338,20 +321,76 @@ These two files should reside in your project directory (``myproject/myproject/`
...
   DEBUG = True
-  TEMPLATE_DEBUG = DEBUG
 
-  ADMINS = (
-      # ('Your Name', 'your_email@example.com'),
-  )
+  # Application definition
 
-  MANAGERS = ADMINS
+  INSTALLED_APPS = [
```
kaedroho: We list the installed apps in the opposite order in the project template: https://github.com/torchbox/wagtail/blob/bc8322bf7e91c1a2606392090e24cd78c9e5a817/wagtail/project_template/project_name/settings/base.py#L26-L52

see: #1536

gasman: Fair point... updated in 88078ac389589bffe4b2368a38200c3f4f80270a.

PR #2140, opened 2016-01-22
Teach StreamBlocks about NON_FIELD_ERRORS

timheap:
A StreamBlock can now have NON_FIELD_ERRORS as part of its error dict.
These will be displayed as error messages above the StreamBlock in the
admin upon form submission.

No code in Wagtail currently makes use of this feautre. It is available
for developers of sites to use in their custom StreamField blocks.

### DIFF
--- a/wagtail/wagtailcore/tests/test_blocks.py
+++ b/wagtail/wagtailcore/tests/test_blocks.py
@@ -8,6 +8,7 @@ from django.forms.utils import ErrorList
 from django.core.exceptions import NON_FIELD_ERRORS, ValidationError
 from django.test import TestCase, SimpleTestCase
 from django.utils.safestring import mark_safe, SafeData
+from django.utils.html import format_html
 
 from wagtail.wagtailcore import blocks
 from wagtail.wagtailcore.rich_text import RichText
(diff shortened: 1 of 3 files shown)


wagtail/wagtailcore/tests/test_blocks.py
```diff
@@ -1271,6 +1301,71 @@ class ValidatedBlock(blocks.StreamBlock):
             3: ['Enter a valid URL.'],
         })
 
+    def test_block_level_validation_no_errors(self):
```
m1kola: If I'am not mistaken `render_form` will not be called in `test_block_level_validation_no_errors` and in `test_block_level_validation_throws_errors`.

What do you want to achieve by writing these two tests? Seems like you want to make sure that `ValidatedBlock`, `ColumnBlock`, `IntegerBlock` works correctly, but they are used only for tests porpuses. Don't you think that it looks like writing tests on tests? :)

m1kola:
Hello, @timheap!

I have some doubts about `test_block_level_validation_no_errors` and `test_block_level_validation_throws_errors`, but in general it looks great.

Thank you!

wagtail/wagtailcore/tests/test_blocks.py
timheap: You're right, this is needlessly complicated. I was attempting to test that the validation logic is called, and any ValidationErrors raised behave as expected. I added tests for a use case that closely matches a real use case that I wrote this PR for, but I can make something simpler. I can also write a test to ensure the error messages actually appear in the rendered form.

m1kola: @timheap, I think `test_block_level_validation_no_errors` and `test_block_level_validation_throws_errors` must not be in `TestStreamBlock`, because **they related only to `ValidatedBlock`**. If you want to be sure that `ValidationError`s raised as expected you can define `TestValidatedBlock` and put these two test in this test case. But I think this is the way to `TestTestValidatedBlock`, `TestTestTestValidatedBlock` and so on :) I think so because `ValidatedBlock` written only for test purposes. 

As **better solution in my opinion** I can suggest you:
* write test that checks html returned by `render_form` with errors (already done with `test_block_level_validation_renders_errors`)
* write test that checks html returned by `render_form` without errors
* remove `test_block_level_validation_no_errors` and `test_block_level_validation_throws_errors`

If we do so, we will still have example that closely matches real use case and tests that covers changes in code (positive and negative scenarios).

What do you think about it?

timheap:
@m1kola I've simplified the tests considerably. I still think the tests are useful to have, to ensure that streamblock-level validation works as expected. However, the previous set up certainly hid the tests behind a bunch of unnecessary extra code.

I have also added a `StreamBlockValidationError` exception that simplifies the creation of ValidationErrors that conform to the format expected by StreamBlocks.

m1kola:
@timheap, I sent PR https://github.com/takeflight/wagtail/pull/1 to your branch to show my vision. You can just decline this PR, but please take a look at `test_block_level_validation_render_no_errors`. I think it would be good to check that no errors will be printed if `render_form` will be called without passing errors.

`StreamBlockValidationError` is good idea, I think!

timheap:
I've merged the PR from @m1kola (with a small fix to remove a now-unused import). This should be good to go now!

m1kola:
I think we can merge this PR. Thank, @timheap!

P.S: build [#5808.7](https://travis-ci.org/torchbox/wagtail/jobs/113039639) was failed. Looks like this PR is not a reason of failure.

gasman:
Merged in 7d22066da561feaecca5f8e496cb0781e245d0ee + parents

PR #2141, opened 2016-01-22
Fix Page model recipes to pass 'self' as the page object

gasman:
These code examples were wrongly updated when we changed the template variable `self` to `page` - we still need to refer to `self` as the variable within view functions (`page` is not defined there).


kaedroho:
Thanks!

Issue #2092, opened 2016-01-08
Notification message when creating a site doesn't use Site Name field


davecranwell:
Despite setting a site name, the notification message reads "Site 'localhost:8111:8111 [default]' created", which isn't ideal. 

The port doesn't seem worth mentioning. The square brackets are a bit non-friendly. The lack of site name itself is the main thing though.

gasman:
Fixed in #2144

PR #2144, opened 2016-01-24
Show site name on notification messages when updating or creating sites, when available

chrxr:
Regarding #2092  

Not sure if this is appropriate. If site name is present then displays just site name (+ [default] if necessary). Port is excluded.

Also fixed a PEP8 issue with AND at beginning of a new line.


chrxr:
Tests added. First time I've done tests, so please let me know if there's anything I could do better there.

Cheers,
Chris

timheap:
The tests look good, @chrxr! The pep8 style checker has spotted some issues with your code though, which is the failed check in the test result block below. You can check out the details of the failed check to find out more details. You need two new lines between class definitions, but your new test class only has one new line between the classes above and below.

chrxr:
Thanks @timheap, resubmitted.

gasman:
Thanks! Merged in 26f5016bb513a8e6f663ffa020f242c9001fcdf6.

Issue #1389, opened 2015-06-09
Inaccurate confirmation message when publishing a page with a future go-live date


mope:
**Version:** 1.0b2

**Steps to reproduce:**

1. Set a future "Go live date/time" on a page and click "publish".
2. View the page again and click "publish".

-> Page is not published. Success message incorrectly states that page has been published.

gasman:
The only thing I can see here that's (arguably) incorrect is that the message ought to say something like "page X has been scheduled for publishing" instead of "page X has been published". Is that the only issue you're raising here, or is there some other incorrect behaviour? If the latter, please can you explain in more detail?

mope:
This issue arose while investigating a different issue where a page with a go live date in the future failed to be published at the correct time. The editor subsequently clicked the "publish" button in the hopes that it would cause the page to go live, and the success message was displayed. The page was not made live and the editor resolved the issue by setting a date in the past before clicking "publish" again (though of course they could have just deleted the date).

I suggest that there are two problems here:

1) Success message incorrectly displayed.
2) UX issue around ambiguous usage of the word "publish".

gasman:
> This issue arose while investigating a different issue where a page with a go live date in the future failed to be published at the correct time. The editor subsequently clicked the "publish" button in the hopes that it would cause the page to go live, and the success message was displayed.

Hmm, that doesn't really clarify things much... Was that a Wagtail bug? Had the go-live date passed by the time the editor clicked 'publish'? (In which case, surely setting a date further in the past wouldn't have made a difference?)

If the editor's confusion was actually a result of buggy behaviour elsewhere, then I don't think we can draw any useful conclusions from the fact that the UX cues here failed to put them back on track.

mope:
Apologies, what I wrote was unclear. Here's a quote from the ticket that was raised about the issue:

> We have tried to use it twice for press releases to go live at a particular time and both times it has not  worked. We set the report page to go live at midnight on Tuesday, not only did it not go live, the page prevented us from publishing it manually until we set the go live date to Monday or earlier.

I haven't finished investigating the issue yet, but if it turns out to be a Wagtail bug I'll raise another ticket. Note however the ambiguity in the phrase "the page prevented us from publishing it". Do they mean "published" as in saved or "published" as in made live? In this case they mean "made live" but that's not always what the button does.

gasman:
[...]

PR #1867, opened 2015-10-27
Allow defining a custom form for Page validation

timheap:
Developers can create their own form class, with custom validation logic, by subclassing `wagtail.wagtailadmin.forms.WagtailAdminPageForm` and setting the `MyPage.base_form_class` attribute:

```python
from wagtail.wagtailadmin.forms import WagtailAdminPageForm
from wagtail.wagtailcore.models import Page

class EventPageForm(WagtailAdminPageForm):
    def clean(self):
        cleaned_data = super(EventPageForm, self).clean()

        start_date = cleaned_data.get('start_date')
        end_date = cleaned_data.get('end_date')
        if start_date and end_date:
            if start_date < end_date:
                msg = 'Start date must be before end date'
                self.add_error('start_date', msg)
                self.add_error('start_date', msg)
            elif start_date == end_date
                cleaned_data['end_date'] = None

        return cleaned_data

class EventPage(Page):
    start_date = models.DateField()
    end_date = models.DateField()

    content_panels = Page.content_panels + [
        FieldPanel('start_date'),
        FieldPanel('end_date'),
    ]
    base_form_class = EventPageForm
```

The `Page` form validation logic has been moved to `WagtailAdminPageForm.clean()`, out of the
`wagtail.wagtailadmin.views.pages.validate_page_data()` function. There may be an issue for developers who have made their own `Page.edit_handler`, which will not set the correct form class for validating pages.

Constructing a `TabbedInterface` for `Page`s is now done by the `Page.get_edit_handler()` function, which behaves the same as the code embedded in `wagtail.wagtailadmin.views.pages.get_page_edit_handler()`. Because it is a function on `Page`, developers can easily override it site wide using a custom `Page` base class.

PR #2145, opened 2016-01-24
Changed success message for when a page is published with go_live_at

chrxr:
In relation to #1389.

Changed success message to show page is scheduled for publishing, rather than published. Also removed 'View live' button, as this resulted in a 404, and I can't think of any other suitable alternative button.

### DIFF
--- a/wagtail/wagtailadmin/views/pages.py
+++ b/wagtail/wagtailadmin/views/pages.py
@@ -173,10 +173,15 @@ def create(request, content_type_app_name, content_type_model_name, parent_page_
 
             # Notifications
             if is_publishing:
-                messages.success(request, _("Page '{0}' created and published.").format(page.title), buttons=[
-                    messages.button(page.url, _('View live')),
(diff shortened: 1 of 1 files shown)


gasman:
Hi @chrxr - publishing a page with a future golive date is failing for me with `name 'timezone' is not defined`. Looks like this is missing an import line...

chrxr:
Hmm. it's working locally for me, timezone was already being imported on pages.py, it's used in validate_page_form.

Any other possible causes apparent @gasman ?

gasman:
Ah, I think I know what's happened there... I merged #1867 just earlier, which shuffled a lot of page validation logic out of pages.py, including the checks for valid golive / expiry dates, so the timezone import got lost in that process. I'll sort it :-)

chrxr:
Gotcha, thanks!

gasman:
Merged in d754bac9bbc800ec40b321fdc6813cb675e34af7 - thanks!

PR #2147, opened 2016-01-25
Remove double tagit initialisation

### LINKED ISSUE #2142: image tags not saving after upload
large
Hi,

We're experiencing a bug when an image is uploaded and you can tag it straight away, the tags don't get saved, Our users need to go into images, then click on the image, and add tags a second time for them to stick.

i have a feeling we had this before but it was fixed? but I could be wrong.

Thanks

Joss

timheap:
tagit was initialized from both the AdminTagWidget, and from the `done`
method in the image uploader. Initializing it twice led to strange
errors and ultimately the tags not being saved on the image model.

As the tag widget now contains its own initialisation `<script>`, tag
inputs do not need to be specially handled by AJAX code any more, so
this can simply be removed.

Fixes bug #2142.

### DIFF
--- a/wagtail/wagtailimages/static_src/wagtailimages/js/add-multiple.js
+++ b/wagtail/wagtailimages/static_src/wagtailimages/js/add-multiple.js
@@ -105,9 +105,6 @@ $(function() {
                 itemElement.addClass('upload-success')
 
                 $('.right', itemElement).append(response.form);
-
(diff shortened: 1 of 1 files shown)


gasman:
Merged in 66da5ad5ad710db12323f333bc2df7f8fbe96fd4 - thanks!

PR #2151, opened 2016-01-25
Only enforce W001 on Page inheritor

benjaoming:
If your `Page` inheritor has a `ForeignKey` to a model that does not relate to the Wagtail page tree, then having `on_delete=models.CASCADE` should be fine.

E.g. this:

```python
class Survey(Page):
    organization = models.ForeignKey(
        'Organization',
        verbose_name=_('Which organization owns this survey'),
        on_delete=models.CASCADE,
    )
```

Raises:

```
Performing system checks...

System check identified some issues:

WARNINGS:
my_app.Survey.organization: (wagtailcore.W001) Field hasn't specified on_delete action
	HINT: Set on_delete=models.SET_NULL and make sure the field is nullable.
```

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -466,7 +466,7 @@ class Page(six.with_metaclass(PageBase, MP_Node, ClusterableModel, index.Indexed
                             for field in model._meta.parents.values() if field]
 
         for field in cls._meta.fields:
-            if isinstance(field, models.ForeignKey) and field.name not in field_exceptions:
+            if isinstance(field, models.ForeignKey) and issubclass(field.rel.to, Page) and field.name not in field_exceptions:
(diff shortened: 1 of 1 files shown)


gasman:
Hi @benjaoming - I don't think that's correct. Here, `on_delete=models.CASCADE` means "if the organisation is deleted, any Survey pages related to the organisation will also be deleted". It's not safe to do that, because deleting a page record without updating the tree-related fields (`path`, `depth`, `numchild`) will corrupt the page tree.

benjaoming:
You're absolutely right, I was very blunt there, sorry about that. I've changed the warning message instead.

Wondering: Will reopening this PR show additional changes done with `git push -f` ?

gasman:
Thanks for the update! Github won't let me reopen this PR now ("The w0001-only-on-Page branch was force-pushed or recreated.") so feel free to open a new one.

PR #2152, opened 2016-01-25
Guidance to acknowledge that models.PROTECT is also fine.

benjaoming:
Taking over from #2151

The main motivation is that *I* was confused. And I'm pretty stupid, especially just before deadlines. I don't know if the PR will solve that much, but maybe it just improves something a little bit, and then one day a grand big solution will emerge :)


gasman:
Thanks! Merged in 512b96aa820e71b7fcfed514a181f08ff5034e1f.

PR #2153, opened 2016-01-26
Fixes #2120 - bail out of RedirectMiddleware if request.site is not set

### LINKED ISSUE #2120: RedirectMiddleware should check request has site attr before using it
`wagtail.wagtailredirects.middleware.RedirectMiddleware` should check for `request.site` existing before accessing it.  If a middleware installed before `wagtail.wagtailcore.middleware.SiteMiddleware` returns a response then the RedirectMiddleware causes a 500 error.

```
Internal Server Error: /page/foo
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 223, in get_response
    response = middleware_method(request, response)
  File "/usr/local/lib/python2.7/dist-packages/wagtail/wagtailredirects/middleware.py", line 22, in process_response
    redirect = models.Redirect.get_for_site(request.site).get(old_path=path)
AttributeError: 'WSGIRequest' object has no attribute 'site'
```

Per [docs](https://docs.djangoproject.com/en/1.9/topics/http/middleware/#process-response):

> Unlike the process_request() and process_view() methods, the process_response() method is 
> always called, even if the process_request() and process_view() methods of the same middleware 
> class were skipped (because an earlier middleware method returned an HttpResponse). In 
> particular, this means that your process_response() method cannot rely on setup done in 
> process_request().

jschneier:
Hi, thanks for the great project, excited to see what it becomes!

I looked for a good example of testing with actual middleware (maybe using `override_settings`??) but couldn't find anything. Care to point me in the right direction for that? Other than that this one is pretty straightforward I think.

### DIFF
--- a/wagtail/wagtailredirects/middleware.py
+++ b/wagtail/wagtailredirects/middleware.py
@@ -11,11 +11,16 @@ class RedirectMiddleware(object):
         if response.status_code != 404:
             return response
 
+        # If a middleware before `SiteMiddleware` returned a response the
+        # `site` attribute was never set, ref #2120
+        if not hasattr(request, 'site'):
+            return response
+
         # Get the path
         path = models.Redirect.normalise_path(request.get_full_path())
 
         # Get the path without the query string or params
-        path_without_query = urlparse(path)[2]
+        path_without_query = urlparse(path).path
 
         # Find redirect
         try:


tomdyson:
Thanks, @jschneier! Handing to @kaedroho to review.

kaedroho:
This looks good, thanks! Merged in 941d3bec178b279a3f3eb93eb7005c1b96852354

PR #2154, opened 2016-01-26
Documentation - fix parameter name on signal handler

fozcode:
The sample code breaks because the signal handler is invoked with a named parameter of `sender` (at `wagtailcore/models.py` line 1421).  The other nearby sample functions are already correct.


fozcode:
Cleaned up my branch, re-opening

PR #827, opened 2014-11-21
[fix] mysql support #541

linibou:
We've fix a bug which was causing an exception when launching first migrate with mysql database

PR #2155, opened 2016-01-26
Beat wagtailimages migration 0004 into submission once and for all

### LINKED ISSUE #2105: Problem with MySQL 5.7 for initial migration.
The initial migration for wagtail images fails on line 17 of 0004_make_focal_point_key_not_nullable.

This can be fixed with the following:

```sql
DELETE FROM wagtailimages_rendition WHERE CONCAT(image_id, '-', filter_id) IN (
    SELECT CONCAT(image_id, '-', filter_id) FROM (SELECT DISTINCT * FROM wagtailimages_rendition) AS x
    WHERE focal_point_key IS NULL 
    GROUP BY image_id, filter_id HAVING COUNT(*) > 1
) AND focal_point_key IS NULL
```

gasman:
Rewrite query for removing duplicate renditions as pure ORM code.

This will hopefully avoid the persistent issues with this query having to
be rewritten for different DB engines and versions (#827, #1692, #2105...)

Fixes #2105

### DIFF
--- a/wagtail/wagtailimages/migrations/0004_make_focal_point_key_not_nullable.py
+++ b/wagtail/wagtailimages/migrations/0004_make_focal_point_key_not_nullable.py
@@ -5,41 +5,26 @@ from django.db import models, migrations
 
 
 def remove_duplicate_renditions(apps, schema_editor):
-    if schema_editor.connection.vendor == 'mysql':
-        schema_editor.execute("""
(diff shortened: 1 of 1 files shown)


kaedroho:
Merged in: https://github.com/torchbox/wagtail/commit/ca343e2358284b81bc8f5c0518f2875707e23567 thanks!

PR #2157, opened 2016-01-27
WAGTAILAPI_MAX_RESULTS is actually called WAGTAILAPI_LIMIT_MAX

### LINKED ISSUE #2156: Wagtail API missing configuration variables
On 1.3.1 there are some configuration variables for Wagtail API like WAGTAILAPI_SEARCH_ENABLED and WAGTAILAPI_MAX_RESULTS.

I cannot see where  WAGTAILAPI_MAX_RESULTS is used on the wagtailapi code. And even so it has no effect on my testing enviroment. Either I put 1 or 100 its the same thing on the response.

I think somehow there is the documentation for the feature, but it is not implemented yet.

kaedroho:
Fixes #2156

### DIFF
--- a/docs/reference/contrib/api/configuration.rst
+++ b/docs/reference/contrib/api/configuration.rst
@@ -16,9 +16,9 @@ Generating URLs to documents will fall back the the current request's hostname i
 Setting this to false will disable full text search. This applies to all endpoints.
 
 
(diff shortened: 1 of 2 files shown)


gasman:
Merged in 13217dd6efc93d99b5c1f25641b91482206b449b

PR #2160, opened 2016-01-27
Make user confirm before navigating away from changed page editor

kaedroho:
This PR implements a simple piece of code that detects whether the form has changed and blocks the user from navigating away from the page editor if it has. This should prevent data loss.

~~One thing it doesn't yet solve is that if the user recieves a validation error, they still can navigate away. Not sure what the best way is to solve that.~~ solved

### DIFF
--- a/wagtail/wagtailadmin/templates/wagtailadmin/pages/edit.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/pages/edit.html
@@ -33,7 +33,7 @@
                 <li class="actions">
                     <div class="dropdown dropup dropdown-button match-width">
                         <button type="submit" class="action-save button-longrunning" tabindex="3" data-clicked-text="{% trans 'Saving...' %}" {% if page.locked %}disabled {% endif %}><span class="icon icon-spinner"></span><em>{% if page.locked %}{% trans 'Page locked' %}{% else %}{% trans 'Save draft' %}{% endif %}</em></button>
-                        
+
                         {% if not page.locked %}
                             <div class="dropdown-toggle icon icon-arrow-up"></div>
                             <ul role="menu">
@@ -99,4 +99,22 @@
 {% block extra_js %}
     {{ block.super }}
     {% include "wagtailadmin/pages/_editor_js.html" %}
+
+    <script>
+        $(function() {
+            /* Make user confirm before leaving the editor if there are unsaved changes */
+            {% trans "This page has unsaved changes." as confirmation_message %}
+            enableDirtyFormCheck(
+                '#page-edit-form',
+                {
+                    ignoredButtonsSelector: 'button[type="submit"]',
+                    confirmationMessage: '{{ confirmation_message|escapejs }}',
+
+                    {% if form.errors %}
+                        alwaysDirty: true,
+                    {% endif %}
+                }
+            );
+        });
+    </script>
 {% endblock %}
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
@@ -36,6 +36,47 @@ function initTagField(id, autocompleteUrl) {
     });
 }
 
+/*
+ * Enables a "dirty form check", prompting the user if they are navigating away
+ * from a page with unsaved changes.
+ *
+ * It takes the following parameters:
+ *
+ *  - formSelector - A CSS selector to select the form to apply this check to.
+ *
+ *  - options - An object for passing in options. Possible options are:
+ *    - ignoredButtonsSelector - A CSS selector to find buttons to ignore within
+ *      the form. If the navigation was triggered by one of these buttons, The
+ *      check will be ignored. defaults to: input[type="submit"].
+ *    - confirmationMessage - The message to display in the prompt.
+ *    - alwaysDirty - When set to true the form will always be considered dirty,
+ *      prompting the user even when nothing has been changed.
+*/
+function enableDirtyFormCheck(formSelector, options) {
+    var $form = $(formSelector);
+    var $ignoredButtons = $form.find(options.ignoredButtonsSelector || 'input[type="submit"]');
+    var confirmationMessage = options.confirmationMessage || ' ';
+    var alwaysDirty = options.alwaysDirty || false;
+    var initialData = $form.serialize();
+
+    window.addEventListener('beforeunload', function(event) {
+        // Ignore if the user clicked on an ignored element
+        var triggeredByIgnoredButton = false;
+        var $trigger = $(event.target.activeElement);
+
+        $ignoredButtons.each(function() {
+            if ($(this).is($trigger)) {
+                triggeredByIgnoredButton = true;
+            }
+        });
+
+        if (!triggeredByIgnoredButton && (alwaysDirty || $form.serialize() != initialData)) {
+            event.returnValue = confirmationMessage;
+            return confirmationMessage;
+        }
+    });
+}
+
 $(function() {
     // Add class to the body from which transitions may be hung so they don't appear to transition as the page loads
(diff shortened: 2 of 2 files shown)


timheap:
Could this be made in to a JavaScript plugin/module? That way, it can be reused for snippets, settings, and other forms in the admin easily.

wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
```diff
@@ -391,6 +391,28 @@ $(function() {
...
 
+    /* Make user confirm before leaving the editor if there are unsaved changes */
+    var initialData = $('#page-edit-form').serialize();
+    function dataHasChanged() {
+        return $('#page-edit-form').serialize() != initialData;
+    }
+
+    window.addEventListener('beforeunload', function(event) {
+        // Ignore if the user clicked one of the submit buttons
+        var trigger = $(event.originalTarget.activeElement);
```
m1kola: Hello, @kaedroho!

Do you use Firefox? [`event.originalTarget` is mozilla-specific](https://developer.mozilla.org/ru/docs/Web/API/Event/originalTarget). I think you should use [`event.target`](https://developer.mozilla.org/en-US/docs/Web/API/Event/target) instaead.

wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
```diff
@@ -391,6 +391,28 @@ $(function() {
...
+        var triggeredBySubmitButton = false;
+        $('#page-edit-form button[type="submit"]').each(function() {
+            if ($(this).is(trigger)) {
+                triggeredBySubmitButton = true;
+            }
+        });
+
+        if (!triggeredBySubmitButton && dataHasChanged()) {
+            event.returnValue = "\o/";
+            return "\o/";
```
tomdyson: Why "\o/"?

JoshBarr:
> One thing it doesn't yet solve is that if the user receives a validation error, they still can navigate away. Not sure what the best way is to solve that.

It would be difficult without having the state of the page (valid/invalid state of each field) stored in some kind of client-side datastore, so you could compare recently changed/valid fields with their original state, to determine whether or not to show the warning. 

@tomdyson: any string value in `event.returnValue` prompts some browsers to fire up the dialog asking if you want to stay on the page. [MDN reference](https://developer.mozilla.org/en-US/docs/Web/Events/beforeunload).

wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
kaedroho: Ahh, yes. Thanks for spotting!

wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
```diff
@@ -391,6 +391,28 @@ $(function() {
...
+        var triggeredBySubmitButton = false;
+        $('#page-edit-form button[type="submit"]').each(function() {
+            if ($(this).is(trigger)) {
+                triggeredBySubmitButton = true;
+            }
+        });
+
+        if (!triggeredBySubmitButton && dataHasChanged()) {
+            event.returnValue = '\o/';
+            return '\o/';
```
m1kola: Chrome (47.0.2526.111) shows this text, so I think we need to write something human readable. Localizations also required in my opinion.

screen shot 2016-01-28 at 13 31 11

m1kola: Empty string works in Chrome like shown below, but doesn't work in Firefox.

screen shot 2016-01-28 at 13 37 53

m1kola:
I think we should to test this attentively in all supported browsers, because of differences in JS API. Refs:
- https://github.com/torchbox/wagtail/pull/2160#discussion_r51067360
- https://github.com/torchbox/wagtail/pull/2160#discussion_r51105253

gasman:
Re making this work when re-showing the form after a validation error: I guess we can cheat, and treat *any* form state after a validation error as having unsaved changes. Quick-and-dirty example to show what I mean (there are probably less hacky ways of doing this):

    if ($('.error-message').length) {
        var initialData = 'this will never match a real serialized form state';
    } else {
        var initialData = $('#page-edit-form').serialize();
    }

Nice idea using the serialized form for this, by the way! I was expecting us to have to implement some sort of 'have you changed or not' API hook on every custom form widget, to deal with dynamically-added elements (InlinePanel, StreamField) and components where the user interaction happens outside of an actual form input (rich text). But this is much neater, and I can't think of any edge cases where it would fail (since, obviously, any meaningful edit has to be reflected in the final posted data).

kaedroho:
I've moved most of the implementation into core.js so it can be reused in other models (thanks @timheap). This is called within the template allowing the confirmation message to be translated.

wagtail/wagtailadmin/templates/wagtailadmin/pages/edit.html
```diff
@@ -99,4 +99,17 @@ <h1 class="icon icon-doc-empty-inverse">{% blocktrans with title=page.title page
...
     {% include "wagtailadmin/pages/_editor_js.html" %}
+
+    <script>
+        $(function() {
+            /* Make user confirm before leaving the editor if there are unsaved changes */
+            enableDirtyFormCheck(
+                function() {
+                    return $('#page-edit-form').serialize()
+                },
+                '{% trans "This page has unsaved changes." %}',
```
m1kola: Are you sure that `trans` will return correct JS string (without `'` character, for example) on all languages? Maybe we should apply `escapejs` filter here?

If you will decide to apply `escapejs`, please check how browser prompt will work with non-ANSI characters.

kaedroho: Ahh, good point. Thanks!

kaedroho: Fixed in 7fe323d

m1kola:
Looks good for me.

Thanks, @kaedroho!

m1kola:
@kaedroho, are you planning to solve problem with validation errors in another PR?

kaedroho:
@m1kola Thanks for the review!

> are you planning to solve problem with validation errors in another PR?

I'd like to do it with this one if I can

kaedroho:
Rewrote it a bit taking into account some suggestions from @timheap. This also fixes the behaviour for validation errors.

m1kola:
Looks great! I think this PR is ready to merge.

kaedroho:
Just made a demosite for this, details here: https://github.com/torchbox/wagtail/pull/2170#issuecomment-176687991

gasman:
It seems that the "submit for moderation" option isn't being caught by the 'ignored buttons' selector - editing content and then clicking 'submit for moderation' causes the warning to be displayed. (Publish works fine, though...)

gasman:
...looks like that's because the submit button is `input[type="submit"]` rather than `button[type="submit"]` like the others. (I think it would make sense for the default value in `enableDirtyFormCheck` to match both, so there's no need to override that default in edit.html...)

gasman:
Have made a couple of additions in https://github.com/gasman/wagtail/tree/kaedroho-navigate-away-confirm, to fix the 'submit to moderator' issue above, and also apply the dirty form handler to the 'create' view - @kaedroho / @m1kola, please can you review?

m1kola:
@gasman, I have no questions. Thanks!

gasman:
Merged in 11b9ed6732326991e6771b19619c213ec27785c7 (with a sneaky additional change: the `preview_on_create` / `preview_on_edit` views now pass `form` to the context, so that you don't bypass the warning when previewing a page with a validation error. Yay for obscure edge cases :-) )

PR #2149, opened 2016-01-25
Fix for ambigious filenames

shredding:
Long filenames become ambigious and can't be displayed.

This is because https://github.com/torchbox/wagtail/pull/574 cuts long filenames.

Given the tag:

```
    {% image image fill-192x144-c100 class="rsTmb" %}
```

And the following URLs:

```
https://neuraum.s3.amazonaws.com/media/public/images/very_long_name_with_appended_incrementor_interior_0.original.jpg

https://neuraum.s3.amazonaws.com/media/public/images/very_long_name_with_appended_incrementor_interior_1.original.jpg

https://neuraum.s3.amazonaws.com/media/public/images/very_long_name_with_appended_incrementor_interior_2.original.jpg
```

... the images thumbs are getting an URL like:

```
https://neuraum.s3.amazonaws.com/media/public/images/very_long_name_with_appended_incrementor_interi.2e16d0ba.fill-192x144-c100.jpg
```

... essentially cutting off the differentiator. This means the renditions are overriding each other and only the last one is displayed.

This is unfortunate, as i have to wrestle with thousand of files like this :)

My proposed solution is to cut of the filename by 32 chars more and append a md5 hash of the original filename to overcome this limitation.

This is in sync with the existing test in 574.

PR #2165, opened 2016-01-28
moved AbstractRendition upload_to to function

robmoggach:
I have large hierarchies of images (100's of projects and dozens of images per project - all stored in subdirectories named from the project page slug)

I'd like to override the get_upload_to function - this should allow this for renditions as well as images

### DIFF
--- a/wagtail/wagtailimages/models.py
+++ b/wagtail/wagtailimages/models.py
@@ -441,7 +441,7 @@ class Filter(models.Model):
 
 class AbstractRendition(models.Model):
     filter = models.ForeignKey(Filter, related_name='+')
-    file = models.ImageField(upload_to='images', width_field='width', height_field='height')
+    file = models.ImageField(upload_to=get_upload_to, width_field='width', height_field='height')
     width = models.IntegerField(editable=False)
     height = models.IntegerField(editable=False)
     focal_point_key = models.CharField(max_length=255, blank=True, default='', editable=False)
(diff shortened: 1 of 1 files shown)


gasman:
@robmoggach Thanks! Just a couple of things that need fixing here:

* The model change requires a migration (this is what's currently causing the tests to fail)
* I'm not sure it's valid to copy the filename-fixing logic from AbstractImage, because the usage cases are different: AbstractRendition shouldn't need to strip non-ASCII characters because the filenames are generated internally rather than being uploaded by the user, and truncating the filename might strip out necessary information (see #2149). In the interest of not introducing any unexpected behaviour changes here, I think we should keep the default `get_upload_to` implementation as a 'do nothing' operation.

kaedroho:
Also, ``get_upload_to`` may need to be moved to the module level so the migrations work for Python 2 users.

gasman:
@kaedroho I thought so at first, but I think that's actually taken care of - as it stands, the code will use the module-level `get_upload_to` function that already exists (which simply delegates to `instance.get_upload_to` - which is exactly the behaviour we want for both images and renditions). It might be neater to have two distinct functions here for images and renditions, but then that's more noise in the migrations...

kaedroho:
Ahh, yes. Thanks for pointing that out.

robmoggach:
Heya - I knew you guys would have a distinct idea on approach for this so I suppose this is more of a feature request than a pull request... please abstract the upload_to for both image and renditions...

I was half thinking of rebuilding [django-media-tree](https://github.com/samluescher/django-media-tree) to use `AbstractImage` and also subclassing `AbstractRendition` to use `sorl-thumbnail` because I love it's caching. My holy grail is to have paths on the browser match paths in the filesystem as I have a local reference library that is identical in file structure. [moggach.com](http://moggach.com) currently has a hacked together version of this with alot of quirks.

gasman:
Superseded by #2476.

Issue #2093, opened 2016-01-08
Error when uploading image or document using chooser doesn't get shown to user


kaedroho:
Currently, if there's an upload error (eg, the server returned a 500 response). The spinner just carries on spinning until the user gets fed up.
This should probably at least give a message like "sorry there's been an error".

gasman:
Fixed in #2167

Issue #616, opened 2014-09-10
Image upload system does not catch errors caused by exceeding webserver maximum filesize/POST-size


davecranwell:
We've experienced these errors one of our staging servers when the user tried to upload a 4MB image.

`2014/09/10 09:46:33 [error] 27068#0: *86706 client intended to send too large body: 4405744 bytes, client: ...`

The result of which is the image upload system falling back to its most default error message, "An error occurred". This needs addressing to improve this useless message and to avoid us repeatedly troubleshooting this for other people. 

Wagtail's image features are all about high quality images so it seems likely that user would be eager to have large original image files and would therefore encounter file size limits, whatever they are, regularly.

Ideally wagtail would somehow be made aware of the server's filesize/post limits. Failing that, the implementer would define a settings entry relating to filesize which merely duplicated the underlying server's limits. Perhaps by setting this low enough and commenting suitably, implementers would be more likely to remember to raise the limit.

gasman:
Reporting server errors back to the user is now implemented for the image/document chooser modals in #2167 - just the primary `/images/` / `/documents/` upload views to sort now.

nfletton:
@gasman I'll take a look. For some reason I thought I'd tested that and the response seemed acceptable but perhaps I imagined it.

gasman:
@nfletton Thanks! When I tried just now, I got a basic 'Upload error' message without the specific error code / message.

nfletton:
@gasman The /images/ upload now handles this error in #2209. This issue cannot be easily addressed for /documents/ as it uses standard model forms (no AJAX) and therefore the error cannot be intercepted in the browser. I don't think it's worthwhile addressing that. #2250 addresses any OS errors that may occur with document uploads though.

gasman:
Fixed in #2209.

prokaktus:
I have feeling, that this logic become broken somehow. Because now when I stucked with this problem (my Nginx was configured with max size of 1mb), nothing was shown to me.

No errors were reported by Image Uploader:
![No errors were reported by Image Uploader](https://user-images.githubusercontent.com/193254/34322703-d7596706-e83f-11e7-95fd-e6c5fdcfcb83.png)

gasman:
@prokaktus Thanks for the report! I can confirm that the error report no longer appears on recent Wagtail versions - it seems that this was inadvertently dropped in #2538. Have now opened #4163 to fix this.

PR #2167, opened 2016-01-28
Handling for server errors when uploading images and documents via a modal dialog

nfletton:
Change to catch and display server errors when uploading images & documents using the modal chooser.

Addresses #2093 and the remaining fix required to clear #616

selection_077

### DIFF
--- a/wagtail/wagtaildocs/templates/wagtaildocs/chooser/chooser.js
+++ b/wagtail/wagtaildocs/templates/wagtaildocs/chooser/chooser.js
@@ -1,3 +1,4 @@
+{% load i18n %}
 function(modal) {
     function ajaxifyLinks (context) {
         $('a.document-choice', context).click(function() {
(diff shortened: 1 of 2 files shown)


gasman:
Excellent! I'm sure this will save a lot of frustration on the part of our developers / users :-)

wagtail/wagtaildocs/templates/wagtaildocs/chooser/chooser.js
```diff
@@ -57,6 +58,13 @@ function(modal) {
             dataType: 'text',
             success: function(response){
                 modal.loadResponseText(response);
+            },
+            error: function(response, textStatus, errorThrown) {
+                {% trans "Report this error to your webmaster with the following information:" as server_error %}
```
m1kola: Hello, @nfletton!

We should be careful with translated string: `trans` may return string that can break JS code. It can be string with `'` character for example.

I think it would be good to apply `escapejs` filter with all non-constant strings.

nfletton: Good point @m1kola

nfletton:
Added suggested escaping. Thanks @m1kola

tomdyson:
Thanks, @nfletton and @m1kola. This looks good for inclusion in 1.4.

m1kola:
@gasman, can you restart [5501.11](https://travis-ci.org/torchbox/wagtail/jobs/105746324) build job, please?. Seems like this PR not a reason of failure.

gasman:
Done!

m1kola:
In my opinion this PR is ready to merge ;)

gasman:
Merged in 94ba1c40e431767f83d5b72d7ad5e6d8a028c143 - thanks!

gasman:
Just to add - we can't count #616 as fully resolved until we have the same functionality on the main `/images/` / `/documents/` upload views, but that ought to be a pretty quick job to add now.

PR #2169, opened 2016-01-29
Fixes notification errors when there are multiple moderation groups

fozcode:
My site has multiple moderation groups - one that can moderate anything, and a few others that only allow moderation of parts of the site tree.  Whenever submitting a page for approval I was getting this error from Postgres:
```
Internal Server Error: /admin/pages/add/bulletin_core/submission/6/
  File "/Users/matt/Dev/Thomas/Bulletin/lib/python3.4/site-packages/django/db/backends/utils.py", line 64, in execute
psycopg2.ProgrammingError: more than one row returned by a subquery used as an expression
```
On SQLite there was no error but not all relevant users received the notification.
This PR fixes the problem, plus a new unit test.


gasman:
Merged in 0845f1e1c1122e6ec4a9d75d488a23ed3d834554 - thanks!

PR #2170, opened 2016-01-29
Revisions UI

kaedroho:
RFC: https://github.com/torchbox/wagtail-rfcs/blob/master/draft/002-revisions-ui.rst

This pull request adds a basic implementation of a UI for viewing and reverting revisions. It follows the draft RFC linked above.

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/_variables.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/_variables.scss
@@ -1,4 +1,4 @@
-// paths 
+// paths
 
 // We can't use absolute paths here, because those are dependent on Django's
 // STATIC_URL setting. Instead, relative paths from the final location of the
@@ -57,6 +57,8 @@ $color-button-yes: $color-green;
 $color-button-yes-hover: darken($color-button-yes, 8%);
 $color-button-no: $color-red;
 $color-button-no-hover: darken($color-button-no, 20%);
+$color-button-warning: $color-orange;
+$color-button-warning-hover: darken($color-orange, 20%);
 $color-link: $color-teal;
 $color-link-hover: $color-teal-dark;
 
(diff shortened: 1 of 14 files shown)


kaedroho:
Demosite setup at: http://wagtail-revisions-ui.demo.torchboxapps.com/admin/
Username: admin
Password: changeme

Everyone's welcome to go and give it a try (note that the UI is early days so doesn't look great at the moment), please post any feedback you have here.

davecranwell:
Aside from the design aspects, I think "Revert" is probably the wrong verb in the revisions list. I found myself worried that it would have an immediate effect if clicked.

Having clicked it however, the lack of warning or context about what I'm doing in the following screen is my next concern. If I have a tab open for each revision, the only identifier of which tab is which, is the URL (and the change itself, but if lots of changes were involved this would be harder to spot).

We discussed an orange warning bar at the top (much the same as when editing a page currently in the moderation queue). Changing the title might help e.g from "Editing such and such" to "Reverting such and such".

I also find myself confused what the _other_ buttons in main saving dropdown do. If I'm viewing a revision and I click "Delete", what happens exactly? How is it that I can unpublish this revision?

Am I correct in saying that the top revision in the list of revisions is actually the current state of the page? This confused me a bit. Some kind of indicator in the revision listing will be necessary to point this out, I think. I found it too easy to assume that the list of revisions was a list of all _past_ states.

tomdyson:
@davecranwell is organising a UX review.

gasman:
Rebased at https://github.com/gasman/wagtail/tree/kaedroho-revisions-ui. Code looks good to me, but needs tests. Is anything else outstanding on this?

gasman:
Notification messages fixed for better i18n in https://github.com/gasman/wagtail/commit/ec137052b2e13a5fea9c67d3ee6cd2ff21c25204 (also some pep8 fixes in the same branch).

gasman:
Tests added in https://github.com/gasman/wagtail/commit/8392e0b07ab1ad08754bfc9505079041ac681971. Will merge that branch in one hour if there are no objections!

gasman:
Merged in 5e07bead42d35bf703bf69f47f1ddc2cbf9839ee + parents

PR #2139, opened 2016-01-21
Added possability to pass custom hallo configuration as attribute on RichTextField and RichTextBlock

HitDaCa:
We very often run into the situation, that we need to provide different hallo editor settings on multiple RichtTextFields and StreamField RichTextBlocks within the same page. 

This feature enables users to pass a custom hallo.js JSON configuration as a RichTextField or RichTextBlock attribute:

```python
from wagtail.wagtailcore.fields import RichTextField
from wagtail.wagtailadmin.edit_handlers import FieldPanel, StreamFieldPanel

from wagtail.wagtailcore.blocks import RichTextBlock
from wagtail.wagtailcore.fields import StreamField

simple_config = {
    'halloheadings': {
        'formatBlocks': ['p', 'h1']
    },
    'halloformat': {
        'formattings': {
            "bold": False,
            "italic": False,
        },
    },
    'hallowagtaildoclink': {},
    'hallolists': {
        "lists": {
            "ordered": False,
            "unordered": False
        }
    },
    'hallowagtaillink': {},
    'hallorequireparagraphs': {}
}

complex_config = {
    'halloheadings': {
        'formatBlocks': ['p', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6']
    },
    'halloformat': {
        'formattings': {
            "bold": True,
            "italic": True,
        },
    },
    'hallowagtaildoclink': {},
    'hallolists': {
        "lists": {
            "ordered": True,
            "unordered": True
        }
    },
    'hallowagtaillink': {},
    'hallorequireparagraphs': {}
}

class BookPage(Page):
    book_text = RichTextField(hallo_config=simple_config)
    page_content= StreamField([
        ('excerpt', RichTextBlock(required=False, hallo_config=simple_config)),
        ('paragraph', RichTextBlock(required=False, hallo_config=complex_config))
    ])    
    content_panels = Page.content_panels + [
        FieldPanel('book_text', classname="full"),
        StreamFieldPanel('page_content')
    ]
```  

I have also updated the RichTextField documentation and RichTextBlock documentation to explain this new attribute addition.

I have added a test for the new attribute.

PR #1521, opened 2015-07-16
Refactored editor

mitchellrj:
Refactored rich-text editor support to make it possible for ``hallo.js`` to be replaced. For example, by [isotoma/wagtailtinymce](https://github.com/isotoma/wagtailtinymce).

I'm confident that I've got this refactored in such a way that doesn't affect the integrity of the current codebase or the interdependencies between packages. The only downside is that users will need to add an extra app to their ``INSTALLED_APPS`` setting.

PR #2175, opened 2016-01-30
Added possibility to pass custom widget with editor configuration through widget attribute on RichTextField and RichTextBlock

HitDaCa:
This pull requests replaces the original pull request [#2139](https://github.com/torchbox/wagtail/pull/2139). I had wrongly rebased to latest master before adding later changes which made code compare unusable (trying to remove rebase closed the original request).

Following the advice from @m1kola & @JoshBarr [#2139 (comment)](https://github.com/torchbox/wagtail/pull/2139#discussion-diff-51297640), RichtTextFields and StreamField RichTextBlocks now accept a **widget** attribute. By default, if none is passed, the standard RichTextArea widget is used.

Following @JoshBarr comment about easily configuring various RichText widgets in future, I have also moved the immutable parts of the RichTextArea into it's own abstract base class. This allows easier creation of future RichTextWidget variations.

``` python
from wagtail.wagtailcore.fields import RichTextField
from wagtail.wagtailadmin.edit_handlers import FieldPanel, StreamFieldPanel

from wagtail.wagtailcore.blocks import RichTextBlock

from wagtail.wagtailcore.fields import RichTextArea

simple_config = {
    'halloheadings': {
        'formatBlocks': ['p', 'h1']
    },
    'halloformat': {
        'formattings': {
            "bold": False,
            "italic": False,
        },
    },
    'hallowagtaildoclink': {},
    'hallolists': {
        "lists": {
            "ordered": False,
            "unordered": False
        }
    },
    'hallowagtaillink': {},
    'hallorequireparagraphs': {}
}

complex_config = {
    'halloheadings': {
        'formatBlocks': ['p', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6']
    },
    'halloformat': {
        'formattings': {
            "bold": True,
            "italic": True,
        },
    },
    'hallowagtaildoclink': {},
    'hallolists': {
        "lists": {
            "ordered": True,
            "unordered": True
        }
    },
    'hallowagtaillink': {},
    'hallorequireparagraphs': {}
}

class BookPage(Page):
    book_text = RichTextField(widget=RichTextArea(editor_config=simple_config))
    page_content= StreamField([
        ('excerpt', RichTextBlock(required=False, widget=RichTextArea(editor_config=simple_config))),
        ('paragraph', RichTextBlock(required=False, widget=RichTextArea(editor_config=complex_config)))
    ])    
        FieldPanel('book_text', classname="full"),
        StreamFieldPanel('page_content')
    ]
```

I have also updated the RichTextField documentation and RichTextBlock documentation to explain this new attribute addition.

I have added a test for the new attribute.


wagtail/wagtailcore/blocks/field_block.py
```diff
@@ -262,7 +262,8 @@ def get_searchable_content(self, value):
 
 class RichTextBlock(FieldBlock):
 
-    def __init__(self, required=True, help_text=None, **kwargs):
+    def __init__(self, required=True, help_text=None, widget=None, **kwargs):
+        self.widget = widget
```
m1kola: I think would be better to set all field options to `self.field_options` (as already done with `help_text` and `required`).

If you decided to do so, `field` method can look similar to this:
```python
    @cached_property
    def field(self):
        field_options = self.field_options.copy()

        if 'widget' not in field_options:
            from wagtail.wagtailcore.fields import RichTextArea
            field_options.update({
                'widget': RichTextArea
            })

        return forms.CharField(**field_options)
```

Another good solution is set all field options to properties in `RichTextBlock`, but this is not widely used in Wagtail: as I see only [`ChooserBlock`](https://github.com/torchbox/wagtail/blob/v1.3.1/wagtail/wagtailcore/blocks/field_block.py#L334) done this way. If you decided to go this way, you can deal with widget like this done in Django. Refs:
* https://github.com/django/django/blob/1.9.1/django/forms/fields.py#L59
* https://github.com/django/django/blob/1.9.1/django/forms/fields.py#L103

wagtail/wagtailcore/fields.py
```diff
@@ -12,31 +12,55 @@
...
 
     def value_from_datadict(self, data, files, name):
-        original_value = super(RichTextArea, self).value_from_datadict(data, files, name)
+        original_value = super(BaseTextAreaWidget, self).value_from_datadict(data, files, name)
         if original_value is None:
             return None
         return DbWhitelister.clean(original_value)
 
+    class Meta:
+        abstract = True
```
m1kola: If I am not mistaken this is not used in fields. Base classes of `BaseTextAreaWidget` defined only with `MediaDefiningClass` that [allows us to define assets](https://docs.djangoproject.com/en/1.9/topics/forms/media/#assets-as-a-static-definition) for widgets. Refs:
* `Textarea` definition: https://github.com/django/django/blob/1.9.1/django/forms/widgets.py#L430
* `Widget` definition: https://github.com/django/django/blob/1.9.1/django/forms/widgets.py#L179
* `MediaDefiningClass` definition: https://github.com/django/django/blob/1.9.1/django/forms/widgets.py#L146

wagtail/wagtailcore/fields.py
```diff
@@ -12,31 +12,55 @@
...
+
+    def __init__(self, widget=None, **kwargs):
+        self.widget = widget
+        super(RichTextField, self).__init__(**kwargs)
+
     def formfield(self, **kwargs):
-        defaults = {'widget': RichTextArea}
+        # check if a custom widget has been passed or use default
+        if not self.widget:
+            self.widget = RichTextArea
```
m1kola: I think in this field we can deal with widget like in Django. See second solution in [my comment](https://github.com/torchbox/wagtail/pull/2175/files#r51343711).

wagtail/wagtailcore/blocks/field_block.py
m1kola: Second solution is for fields, but we can apply same idea for blocks.

docs/advanced_topics/customisation/page_editing_interface.rst
```diff
@@ -61,6 +61,49 @@ However, template output from :class:`~wagtail.wagtailcore.fields.RichTextField`
...
+
+.. code-block:: python
+
+    from wagtail.wagtailcore.fields import RichTextField
+    from wagtail.wagtailadmin.edit_handlers import FieldPanel
+
+    from wagtail.wagtailcore.fields import RichTextArea
+
+    simple_config={
+        'halloheadings': {
```
m1kola: If we decided to give generic name `editor_config` for editor config options, I think config for hallo should look like this:
```python
simple_config={
    'plugins': {
        'halloheadings': {...},
        'halloformat': {...},
    }
}
```

This is because hallo allows configure not only plugins ([see example](https://github.com/bergie/hallo/blob/1.1.1/examples/fixed.html#L212) in hallo repo).

wagtail/wagtailcore/fields.py
```diff
@@ -12,31 +12,55 @@
...
+    def __init__(self, editor_config=None, **kwargs):
+        # set the base editor configuration if existent
+        self.editor_config = editor_config
+        super(BaseTextAreaWidget, self).__init__(**kwargs)
+
     def get_panel(self):
-        from wagtail.wagtailadmin.edit_handlers import RichTextFieldPanel
-        return RichTextFieldPanel
+        raise NotImplementedError(
+            "Class %s doesn't implement a required Method()" % (self.__class__.__name__))
```
m1kola: Would be great to say what method developer need to implement. I mean message may look like this:
```python
"Class %s doesn't implement a required get_panel method" % (self.__class__.__name__))
```

m1kola:
Could you, please, check in tests also following:
* That `RichTextBlock.field` method returns `CharField` default widget, if custom widget is not provided
* That `RichTextBlock.field` method returns `CharField` custom widget, if custom widget is provided
* That `RichTextArea.render_js_init` returns expected result when `editor_config` is provided
* That `RichTextArea.render_js_init` returns expected result when `editor_config` is not provided
* That `RichTextField.formfield` returns Field with default widget, if custom widget is not provided
* That `RichTextField.formfield` returns Field with custom widget, if custom widget is provided

m1kola:
Other than the above comments, this looks good! Thanks, @HitDaCa.

@JoshBarr, please, take a look at this.

m1kola:
@HitDaCa, if you need any assistance with this PR, feel free to let me know.

HitDaCa:
@m1kola, thanks for all the feedback! I'll read through each of your suggestions in detail tomorrow morning and give my thoughts on every point. Thanks.

wagtail/wagtailcore/blocks/field_block.py
HitDaCa: I agree with your first proposed approach, this pattern is used mostly throughout other specific block types and upholding this consistency makes most sense to me.

As to your proposed solution, I'd prefer using the standard get method of the dictionary. This better covers all cases of a non existing key as well as an existing key returning None. The naming conventions in the below code now also follows used patterns.

```python
def field(self):
    field_kwargs = self.field_options.copy()
    # check if a custom widget has been set or use default
    if not field_kwargs.get('widget'):
        from wagtail.wagtailcore.fields import RichTextArea
        field_kwargs.update({'widget': RichTextArea})
    return forms.CharField(**field_kwargs)

I've update the branch with this solution. I'm happy for your feedback if you see room for improvements.

File: wagtail/wagtailcore/fields.py (hunk 3 above)
Comment (HitDaCa) [inline wagtail/wagtailcore/fields.py]: I've updated this code following the same pattern described in [#2175 (comment)](https://github.com/torchbox/wagtail/pull/2175#discussion_r51350568). 
Open for suggestions to solve this differently though.

File: wagtail/wagtailcore/fields.py (hunk 5 above)
Comment (HitDaCa) [inline wagtail/wagtailcore/fields.py]: Code updated.

File: docs/advanced_topics/customisation/page_editing_interface.rst (hunk 4 above)
Comment (HitDaCa) [inline docs/advanced_topics/customisation/page_editing_interface.rst]: I would be keen to pursue passing a whole configuration and keeping the **editor_config** as the attribute name. Only passing **plugins** seems very biased towards the hallo editor and would possible not fit when implementing other editors.

I would agree with you suggestions of updating the configuration to reflect/allow all possible configuration options of an editor to be passed. 

I will also update page-editor.js handling the passed code later today.

File: wagtail/wagtailcore/fields.py (hunk 2 above)
Comment (HitDaCa) [inline wagtail/wagtailcore/fields.py]: The purpose of making this class abstract is to insure this class is never directly instantiated. The BaseTextAreaWidget is intended to only be subclassed from.

Please excuse if I'm misunderstanding the question.

File: docs/advanced_topics/customisation/page_editing_interface.rst (hunk 4 above)
Comment (HitDaCa) [inline docs/advanced_topics/customisation/page_editing_interface.rst]: The page-editor now by default implements and handles a hallo configuration instead of only hallo plugin. 

I have updated the pull requests codebase to reflect these changes.

File: wagtail/wagtailcore/fields.py (hunk 2 above)
Comment (HitDaCa) [inline wagtail/wagtailcore/fields.py]: Ahhh, I got myself confused. Class Meta is of course django Model specific and python, apart from the abc library, doesn't provide a strict abstract class definition. 

Sorry, I seem to be a bit thick today... I will update the codebase and add a test for the base class.

**Comment by HitDaCa** [comment]
Just updating unit tests and having some issues accessing to Github repository (Github seems to be having some access issues). Please bare with me, updates to follow.

**Comment by HitDaCa** [comment]
@m1kola & @JoshBarr , I've finished updating/commenting all raised items. It would be awesome if you guys could double check the changes when you have some time and give your feedback.

Thx for taking your time for reviewing this.

File: wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js [hunk 6]
```diff
@@ -34,11 +39,9 @@ function makeRichTextEditable(id) {
 
     var closestObj = input.closest('.object');
 
-    richText.hallo({
-        toolbar: 'halloToolbarFixed',
-        toolbarCssClass: (closestObj.hasClass('full')) ? 'full' : (closestObj.hasClass('stream-field')) ? 'stream-field' : '',
-        plugins: halloPlugins
-    }).bind('hallomodified', function(event, data) {
+    halloConfig.toolbarCssClass = (closestObj.hasClass('full')) ? 'full' : (closestObj.hasClass('stream-field')) ? 'stream-field' : '';
```
m1kola: If we want to give ability to override `toolbarCssClass`, line `$.extend(halloConfig, editorConfig);` should be after this line.

wagtail/wagtailcore/tests/test_blocks.py
```diff
@@ -175,6 +175,39 @@ def test_validate_non_required_richtext_block(self):
         self.assertIsInstance(result, RichText)
         self.assertEqual(result.source, '')
 
+    def test_widget_field_richtext_block(self):
+        from wagtail.wagtailcore.fields import RichTextArea
```
m1kola: Can we do this import in top of file?

m1kola:
@HitDaCa, thanks!

Looks great, but I have doubts about tests. I'll add couple examples how I think tests supposed to be written.

wagtail/wagtailcore/tests/test_blocks.py
```diff
@@ -175,6 +175,39 @@ def test_validate_non_required_richtext_block(self):
...
+        custom_widget = RichTextArea(editor_config={'plugins': {'halloheadings': {'formatBlocks': ['p', 'h2']}}})
+        block = blocks.RichTextBlock(widget=custom_widget)
+        self.assertEqual(block.field_options.get('widget'), custom_widget)
```
m1kola: We don't need to care about internal implementation. So we can just remove this test, I think.

wagtail/wagtailcore/tests/test_blocks.py
```diff
@@ -175,6 +175,39 @@ def test_validate_non_required_richtext_block(self):
...
+        from wagtail.wagtailcore.fields import BaseTextAreaWidget
+
+        class CustomTextArea(BaseTextAreaWidget):
+            def get_panel(self):
+                pass
+        custom_widget = CustomTextArea(editor_config={'custom': {'testix': {'formatBlocks': ['p', 'h2']}}})
+        block = blocks.RichTextBlock(widget=custom_widget)
+
+        self.assertIsInstance(block.field_options.get('widget'), CustomTextArea)
+        self.assertIsInstance(block.field.widget, CustomTextArea)
```
m1kola: This test we can replace with this one:

```python
    def test_widget_custom_widget_richtext_block(self):
        custom_widget = RichTextArea(editor_config={'custom': {'testix': {'formatBlocks': ['p', 'h2']}}})
        block = blocks.RichTextBlock(widget=custom_widget)

        self.assertEqual(block.field.widget, custom_widget)
```

We don't need to define new widget class to cover changes in code: we just want to pass widget to `RichTextBlock`. Test fails even if `block.field.widget` will instance of `RichTextArea`, but not ours instance.

wagtail/wagtailcore/tests/test_blocks.py
```diff
@@ -175,6 +175,39 @@ def test_validate_non_required_richtext_block(self):
...
+
+    def test_widget_default_widget_richtext_block(self):
+        from wagtail.wagtailcore.fields import RichTextArea
+        deafult_widget = RichTextArea()
+        block = blocks.RichTextBlock(widget=deafult_widget)
```
m1kola: If we want to test default widget, we must not to pass `widget` argument to `RichTextBlock`.

I'm not sure about checks. Maybe something like this this?
```python
    def test_widget_default_widget_richtext_block(self):
        block = blocks.RichTextBlock()

        self.assertIsInstance(block.field.widget, RichTextArea)
```

wagtail/wagtailcore/tests/test_blocks.py
```diff
@@ -175,6 +175,39 @@ def test_validate_non_required_richtext_block(self):
...
+
+        self.assertIsInstance(block.field_options.get('widget'), RichTextArea)
+        self.assertIsInstance(block.field.widget, RichTextArea)
+
+    def test_widget_fallback_widget_richtext_block(self):
```
m1kola: We can remove this test, if we add [this](https://github.com/torchbox/wagtail/pull/2175/files#r51359792).

wagtail/wagtailcore/tests/test_fields.py
```diff
@@ -0,0 +1,38 @@
...
+from wagtail.wagtailcore.fields import RichTextField
+
+
+class TestRichTextField(TestCase):
+
+    def test_has_field_options_attribute(self):
+        field = RichTextField()
+        self.assertTrue(hasattr(field, 'field_options'))
+
+    def test_field_options_widget_default(self):
```
m1kola: We can remove `test_has_field_options_attribute`,  `test_field_options_widget_default` and `test_formfield_has_widget`, because we don't need to care about internal implementation ;)

wagtail/wagtailcore/tests/test_fields.py
```diff
@@ -0,0 +1,38 @@
...
+        from wagtail.wagtailcore.fields import RichTextArea
+        field = RichTextField()
+        formfield = field.formfield()
+        self.assertIsInstance(formfield.widget, RichTextArea)
+        self.assertTrue(hasattr(field.formfield().widget, 'editor_config'))
+
+    def test_formfield_widget_custom(self):
+        from wagtail.wagtailcore.fields import BaseTextAreaWidget
+
+        class CustomTextArea(BaseTextAreaWidget):
```
m1kola: What do you think if we leave only `test_formfield_widget_default` and `test_formfield_widget_custom` tests?
```python
class TestRichTextField(TestCase):

    def test_formfield_widget_default(self):
        field = RichTextField()
        formfield = field.formfield()
        self.assertIsInstance(formfield.widget, RichTextArea)

    def test_formfield_widget_custom(self):
        custom_widget = RichTextArea(editor_config={'custom': {'testix': {'formatBlocks': ['p', 'h2']}}})
        field = RichTextField(widget=custom_widget)
        formfield = field.formfield()

        self.assertEqual(formfield.widget, custom_widget)
```

wagtail/wagtailcore/tests/test_widgets.py
```diff
@@ -0,0 +1,40 @@
+from django.test import TestCase
+from wagtail.wagtailcore.fields import BaseTextAreaWidget, RichTextArea
+
+
+class TestBaseTextAreaWidget(TestCase):
+
+    def test_get_panel_implemented(self):
+        with self.assertRaises(NotImplementedError):
+            widget = BaseTextAreaWidget()
```
m1kola: This line should be not in `with` because `__init__` not a subject of this tests.

wagtail/wagtailcore/tests/test_widgets.py
```diff
@@ -0,0 +1,40 @@
...
+            widget.get_panel()
+
+    def test_render_js_init(self):
```
m1kola: I think we can remove this test because it checks `WidgetWithScript.render_js_init`. Probably `WidgetWithScript` should have own tests.

wagtail/wagtailcore/tests/test_widgets.py
```diff
@@ -0,0 +1,40 @@
...
+        from wagtail.wagtailadmin.edit_handlers import RichTextFieldPanel
+        widget = RichTextArea()
+        panel_proxy = widget.get_panel()
+        panel = panel_proxy(None)
+        self.assertIsInstance(panel, RichTextFieldPanel)
+
+    def test_widget_default_render_js_init(self):
+        widget = RichTextArea()
+        js_init = widget.render_js_init('test-id', 'test', None)
+        # self.assertIsInstance(js_init, unicode) Travis CI - flake8 does not recognise unicode
```
m1kola: `test_widget_default_render_js_init` and `test_widget_eit_config_render_js_init` looks good for me.

If you want to check unicode in Python 2 and 3, you need to use `six.text_type`, but I think we can omit this check in these two tests.

m1kola:
@HitDaCa, I sent [PR](https://github.com/HitDaCa/wagtail/pull/1) to your branch with suggested changes. I hope it save your time.

Thanks again!

wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
HitDaCa: Yeah, I was pondering this one yesterday evening... Implemented correctly, you are right we need to allow the passed config to also override this value. We shouldn't be limiting the passed configuration possibilities for convenience.

I've updated the code.

HitDaCa:
@m1kola Ha, thanks for the [PR](https://github.com/HitDaCa/wagtail/pull/1) updating the tests, I see what you were aiming at. Thanks.

I've merged your changes and updated the PR

wagtail/wagtailcore/tests/test_widgets.py
HitDaCa: Nice one, thanks. Good to know about the `six.text_type` requirement

wagtail/wagtailcore/tests/test_widgets.py
HitDaCa: Agree.

m1kola:
@HitDaCa, thank you for patience! Looks great for me!

@gasman, can you, please, set same milestone as in #2139 and required labels? This PR replaces #2139.

ababic:
Guys, thanks to all of you for putting in the work on this. It looks like a really valuable addition, and i'm excited to use it!

gasman:
This is really neat! Unfortunately I've encountered a snag - passing a `widget` property to `RichTextBlock` breaks the resulting migration, because `RichTextArea` (and widgets / form field objects in general) can't be serialized into the migration. (See http://docs.wagtail.io/en/v1.3.1/topics/streamfield.html#streamfield-definitions-within-migrations for some discussion of the issues with StreamField in migrations.) This could probably be fixed directly by implementing a `deconstruct` method on `RichTextArea`, but I'm not sure I like that approach - it makes RichTextArea a special case as the only widget type that will work, whereas the API here suggests that it should be possible to pass any widget. Rather, I'd take this as a hint from Django that we shouldn't really be embedding form-related objects into a model-level definition. (We've encountered this before, in the design of `FieldBlock` - it's the reason we have to subclass `FieldBlock` for each individual field type, rather than being able to write something like `FieldBlock(field=forms.CharField(max_length=255))` into the StreamField definition.)

I'm not sure what the best remedy is here. We could of course go back to the original plan of passing the `hallo_config` dict directly to `RichTextBlock`, but now I'm wondering if that's bad design - it occurs to me that this might be the first place where we have a StreamField block parameter that exists purely to specify the behaviour of the editor, rather than to define the data format. Maybe this kind of configuration ought to be separated out of the model-level StreamField definition, just as Django separates model fields and form fields?

gasman:
Having thought about this some more, I've reached the conclusion that, whatever the parameter to `RichTextField` / `RichTextBlock` might be, it should not describe a particular rich text widget's behaviour. This is for several reasons:

* The standard Django objects for this purpose (i.e. form fields / widgets) cannot be serialized for migrations.
* RichTextField and StreamField are ultimately model fields, describing a database-level data representation, and details about the form representation do not belong there. Just as Django's own APIs enforce a separation between model fields and form fields, we should find a better place to put those purely 'cosmetic' options, if we need them. (For RichTextField, the better place is probably the 'panels' definition; for RichTextBlock, we probably don't have any suitable place at the moment.)
* The ability to swap in different rich text backends (e.g. TinyMCE - see #1521) is a very commonly requested feature, and one we'll probably work on very soon. Any implementation that only caters for hallo.js is only going to be a stopgap solution at best.

I think the answer is for the parameter to define a list of "features" - things that this particular rich text field is allowed to contain, such as paragraphs, bold, italics, other inline markup such as `<code>`, page links, document links, images, embeds, ULs, OLs. Something like:

    body = RichTextField(features=['p', 'b', 'i', 'h2', 'page-link', 'doc-link'])

(In practice we'd probably want a site-wide common set of features - or perhaps several standard 'palettes', for different use-cases - and a mechanism for adding/subtracting from that list on a per-field basis, like `django.forms`'s `fields`/`exclude` or https://github.com/torchbox/wagtail-rfcs/pull/5.)

The key distinction here is that we're saying "here are some conditions that the data must conform to" (and can potentially enforce that at the whitelisting/validation stage), not "this is how rich text widget X should behave". It's up to each rich text backend to look at the feature list and configure itself accordingly (adding/removing toolbar buttons, or whatever) to present that feature set to the user as best it can. Different rich text backends will have different capabilities, of course, and that's fine - it's up to the developer to ensure that they've set up a rich text backend that actually supports the features they've specified. For example, if a field is defined as `RichTextField(features=['p', 'table'])`, but they're using the hallo.js backend which doesn't support table editing, then it'll simply ignore the 'table' feature and editors won't be able to put tables into their pages.

There's a lot more thinking to be done to define what a 'feature' is here - it's not enough to just come up with a standard list of identifiers and expect each rich text backend to know what they mean, because that would result in massive duplication of code (e.g. each backend has to know that 'doc-link' involves launching `/admin/documents/chooser/` as a modal and doing a particular thing with the JSON result) and wouldn't be extensible enough (in order to add a new link type or allowed HTML tag, you'd have to hack the individual rich text backends). Most likely, we'd identify several general classes of feature (e.g. "ones that wrap a span of text in an HTML tag", "ones that insert a link based on the result of a modal") and have a mapping of feature identifiers to full descriptions, e.g.:

    'doc_link': {
        'feature-type': 'LinkFromModal',
        'modal-url': '/admin/documents/chooser',
        'icon': 'document'
    }

As I say - needs more thinking...

kunambi:
It's 4 months since last activity. Will the feature of having RichTextField-specific options eventually reach release?

TYIA

thibaudcolas:
Hey @kunambi, Dave and Josh were working on this while employed with me at Springload and have both moved on to other jobs since then. @gasman has expressed reserves with the current approach, so there needs to be more work on the PR, but no one has intentions to do this work as far as I know.

I will close this PR, should someone want to take it from there they should make a new one and address the feedback above.

Finally, if you're interested in rich text configuration Springload has been working on a new text editor that solves this use case for us. It's not as well documented, or stable, as I'd like it to be but we do already use it on production projects. Check out https://github.com/springload/wagtaildraftail, and https://github.com/wagtail/wagtail/issues/2409.

gasman:
Now implemented (with an editor-agnostic `features` parameter as originally proposed above https://github.com/wagtail/wagtail/pull/2175#issuecomment-194838184) in #3736.

PR #2179, opened 2016-01-30
Make wagtailsearch language aware

### LINKED ISSUE #1177: Support for language in SearchFields/search
Some sites I've worked on have fields for a specific languages eg (``body_en``, ``body_fr``, etc).

Currently, both are indexed and both get searched no matter which language the user is searching in.

I propose we add a language kwarg to both ``SearchField`` class and ``search`` method. This would prevent issues where two words exist in both languages but have different meanings and also provide a hint to the search backend what language is being used which can be used for stop words, stemming, etc.

Eg:

```
class MyPage(Page):
    body_en = models.TextField()
    body_fr = models.TextField()

    search_fields = (
        index.SearchField('body_en', language='en'),
        index.SearchField('body_fr', language='fr'),
    )

>> MyPage.objects.search("hello", language='en')
[ some results ]

>> MyPage.objects.search("bonjour", language='fr')
[ some results ]
```

razisayyed:
This is based on issue #1177 

I added changes to wagtailsearch app to make it language aware. The changes are based on having multiple search indexes, one for each language.

``` python
WAGTAILSEARCH_BACKENDS = {
    'default': {
        'BACKEND': 'wagtail.wagtailsearch.backends.elasticsearch.ElasticSearch',
        'URLS': ['http://localhost:9200'],
        'INDEX': 'wagtail',
        'TIMEOUT': 5,
    },
    'en': {
        'BACKEND': 'wagtail.wagtailsearch.backends.elasticsearch.ElasticSearch',
        'URLS': ['http://localhost:9200'],
        'INDEX': 'wagtail_en',
        'TIMEOUT': 5,
        'LANGUAGE_CODE': 'en'
    },
    'ar': {
        'BACKEND': 'wagtail.wagtailsearch.backends.elasticsearch.ElasticSearch',
        'URLS': ['http://localhost:9200'],
        'INDEX': 'wagtail_ar',
        'TIMEOUT': 5,
        'LANGUAGE_CODE': 'ar'
    },
}
```

there is no need to make any changes in the way we use search() method.

``` python
PageModel.objects.search('search query')
```

search method is language aware, It will try to use the backend with LANGUAGE_CODE equals to current language, and it falls back to 'default' language if none was found.

also update_index will update each index with its correct language

and post_save signal also updated to be language aware.

**NOTE: these changes are applied to ElasticSearch backend only. It must not affect Db Backend.**

closes #1177

### DIFF
--- a/wagtail/wagtailsearch/management/commands/update_index.py
+++ b/wagtail/wagtailsearch/management/commands/update_index.py
@@ -3,6 +3,7 @@ from optparse import make_option
 from django.core.management.base import BaseCommand
 from django.conf import settings
 from django.db import transaction
+from django.utils import translation
 
 from wagtail.wagtailsearch.index import get_indexed_models
 from wagtail.wagtailsearch.backends import get_search_backend
@@ -22,6 +23,12 @@ class Command(BaseCommand):
 
         # Get backend
         backend = get_search_backend(backend_name)
+        
+        # Activate backend language
+        cur_language = translation.get_language()
+        backend_language = getattr(backend, 'language_code', None)
+        if backend_language is not None:
+            translation.activate(backend_language)
 
         # Get rebuilder
         rebuilder = backend.get_rebuilder()
@@ -56,7 +63,11 @@ class Command(BaseCommand):
         # Finish rebuild
         self.stdout.write(backend_name + ": Finishing rebuild")
         rebuilder.finish()
-
+        
+        # Return to Original Thread Language
+        if backend_language is not None:
+            translation.activate(cur_language)
+            
     option_list = BaseCommand.option_list + (
         make_option(
             '--backend',
--- a/wagtail/wagtailsearch/management/commands/update_index.py
+++ b/wagtail/wagtailsearch/management/commands/update_index.py
@@ -42,7 +42,7 @@ class Command(BaseCommand):
         index = rebuilder.start()
 
         for model, queryset in object_list:
-            self.stdout.write(backend_name + ": Indexing model '%s.%s'" % (
+            self.stdout.write(backend_name + u": Indexing model '%s.%s'" % (
                 model._meta.app_label,
                 model.__name__,
             ))
@@ -56,7 +56,7 @@ class Command(BaseCommand):
                 index.add_items(model, chunk)
                 count += len(chunk)
 
-            self.stdout.write("Indexed %d %s" % (
+            self.stdout.write(u"Indexed %d %s" % (
                 count, model._meta.verbose_name_plural))
             self.print_newline()
 
--- a/wagtail/wagtailsearch/management/commands/update_index.py
+++ b/wagtail/wagtailsearch/management/commands/update_index.py
@@ -23,7 +23,7 @@ class Command(BaseCommand):
 
         # Get backend
         backend = get_search_backend(backend_name)
-        
+
         # Activate backend language
         cur_language = translation.get_language()
         backend_language = getattr(backend, 'language_code', None)
@@ -63,11 +63,11 @@ class Command(BaseCommand):
         # Finish rebuild
         self.stdout.write(backend_name + ": Finishing rebuild")
         rebuilder.finish()
-        
+
         # Return to Original Thread Language
         if backend_language is not None:
             translation.activate(cur_language)
-            
+
     option_list = BaseCommand.option_list + (
         make_option(
             '--backend',
--- a/wagtail/wagtailsearch/management/commands/update_index.py
+++ b/wagtail/wagtailsearch/management/commands/update_index.py
@@ -1,3 +1,5 @@
+from __future__ import unicode_literals
+
 from optparse import make_option
 
 from django.core.management.base import BaseCommand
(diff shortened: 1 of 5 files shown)


wagtail/wagtailsearch/management/commands/update_index.py
```diff
@@ -35,7 +42,7 @@ def update_backend(self, backend_name, object_list):
         index = rebuilder.start()
 
         for model, queryset in object_list:
-            self.stdout.write(backend_name + ": Indexing model '%s.%s'" % (
+            self.stdout.write(backend_name + u": Indexing model '%s.%s'" % (
```
kaedroho: Could use use ``from __future__ import unicode_literals`` at the top of this file instead?

kaedroho:
Thanks! this is looking really good so far!

Looks like this could do with some tests, can you write some?

I'd also like to think about the configuration of wagtailsearch at some point which will affect this change. A "backend" refers to an individual elasticsearch server, not an index. It would be neater if there was a way to configure multiple indices inside a backend without needing to repeat the backend configuration.

razisayyed:
@kaedroho, actually I took the idea of different backends from aldryn-search (django-cms plugin). Your Idea of having multiple indices in the same BACKEND is nice, but having different BACKENDS (which can point to the same Elasticsearch server) will give us the flexibility to use different technologies for different languages (consider having more backends at future like solar, Whoosh ...).

For tests I am more than happy to write some. But I did not write any test for Django app before now, can you please point me to the right direction.

pyMan:
Hi @kaedroho . When is supposed to be relased this enhancement?

kaedroho:
> When is supposed to be relased this enhancement?

This will be released when it's finished. The tasks remaining are:

 - [ ] Agree on how to configure the languages in the search backends and implement any changes
 - [ ] Tests
 - [ ] Docs

pyMan:
With the first point do you mean how to configure a default analyzer for each index?

Something like:
```
# the value of type is mapped to language codes
SearchBackend.settings["settings"]["index"] = {"analysis" : {"analyzer" : {"default" : {"type" : "italian"}}}}
```

The problem here is that settings are likely the same for each es-backend, since SearchBackend is a sort of a singleton object.

Any other points about this?

kaedroho:
> The problem here is that settings are likely the same for each es-backend, since SearchBackend is a sort of a singleton object.

Each ES "backend" can create multiple indices and each index can have its own settings. So another option could be to configure the backend once but provide a list of languages the backend needs to support.

For example:

```
WAGTAILSEARCH_BACKENDS = {
    'default': {
        'BACKEND': 'wagtail.wagtailsearch.backends.elasticsearch.ElasticSearch',
        'URLS': ['http://localhost:9200'],
        'INDEX': 'wagtail',
        'TIMEOUT': 5,
        'LANGUAGES': ['en', 'ar']
    },
}
```

pyMan:
Maybe I am wrong and I miss something. @razisayyed I think your work is great but actually all language fields are still indexed all togheter and you may experience the same problems described in #1177: I mean if you have 'body_en' and 'body_ar' they are both indexed togheter in 'en' index and in 'ar' index.

So some modification is required to manage a new language parameter for index.SearchField as suggested in #1177 (that should defaults to None) and some modification is needed when building the index to consider only the fields with the right language plus the fields with no language specification (suitable for any language).

Anyway I prefer @razisayyed 's configuration than @kaedroho 's one because I feel to have more control on the backends and its settings. And still free to use get_search_backend with specific backend name as argument for any particular thing needed to do. I suggest some improvements to the configuration too:

- The language parameter for SearchField may be a list of languages (a field may be suitable for more languages or really similar languages like 'en-us' and 'en-uk')

- The LANGUAGE_CODE in the WAGTAILSEARCH_BACKENDS configuration may be a list of languages (for the same reason before or just because you don't want an index for any managed language if there are many)

So far...

1. A backend with no LANGUAGE_CODE will index all SearchFields
2. A backend with one or more LANGUAGE_CODE will index SearchFields with no language code or those that include one of the specified LANGUAGE_CODE (SearchFields for other languages will be ignored and excluded)
3. Optional but active by default: a LanguageAnalyzer according to the LANGUAGE_CODE should be set as default Analyzer for the index
4. You are free to query the 'generic' index with no specific LANGUAGE_CODE that contains all data or to query a specific language index that contains only data relating tipically the current active language being more precise with the results.

What do you thing about this? 

P.S. Sorry for my english

kaedroho:
> Anyway I prefer @razisayyed 's configuration than @kaedroho 's one because I feel to have more control on the backends and its settings

Happy to go with this now as it fits in well with #3036 

> What do you think about this?

this approach sounds good to me!

pyMan:
> Happy to go with this now as it fits in well with #3036 

@kaedroho #3043 is going on this same topic (#2179) including #1177 too. Moreover the indexes are being built using only the right fields for the specified index language.

kaedroho:
Thanks for implementing this!

We've recently added internationalisation support into Wagtail itself, so the approach will probably have to be changed, so I'm going to close this PR.

PR #2180, opened 2016-01-31
Fix frontend cache code example

mixxorz:
In the docs, there's a slight coding error with regards to purging the pages of an index page.

Without the `+ 1` at the end, the last page wouldn't be purged. (Check diff)


kaedroho:
Thanks for fixing!

mixxorz:
Thanks for merging :D

PR #2185, opened 2016-02-01
Documentation - Fix parameter name on signal handler

fozcode:
Replacement for PR #2154 that I messed up.

The sample code breaks because the signal handler is invoked with a named parameter of `sender` (at `wagtailcore/models.py` line 1421). The other nearby sample functions are already correct.

### DIFF
--- a/docs/reference/signals.rst
+++ b/docs/reference/signals.rst
@@ -66,7 +66,7 @@ wish to do something when a new blog post is published:
     from mysite.models import BlogPostPage
 
     # Do something clever for each model type
-    def receiver(model, **kwargs):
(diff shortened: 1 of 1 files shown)


m1kola:
Thanks, @fozcode!

You're right. [Django documentation](https://docs.djangoproject.com/en/1.8/topics/signals/#receiver-functions) says:

> Notice that the function takes a **sender** argument, along with wildcard keyword arguments (**\*\*kwargs**); all signal handlers must take these arguments.

@kaedroho, I think we can merge this PR ;)

gasman:
Thanks! Merged in 999e06af27594c75eb9043b72faec462c22c968f.

Issue #2048, opened 2015-12-16
Create fields for structural block types


alexgleason:
I want to make a field that's a simple list downloads. The following code accomplishes what I want using StreamField, but the API is a little weird.

```python
class ResourcesPage(Page):
    files = StreamField([
        ('files', blocks.ListBlock(DocumentChooserBlock()))
    ])
```
```
{% for file in self.files.value.files %}
  {{ file }}
{% endfor %}
```

Ideally it would look something like this:

```python
class ResourcesPage(Page):
    files = ListField(DocumentChooserBlock())
```
```
{% for file in self.files %}
  {{ file }}
{% endfor %}
```

Here I've created a new imaginary type, ListField. There are a few possible solutions (such changing the way StreamField parameters work) but I think it makes sense to have a wrapper field around each of the [structural block types](http://docs.wagtail.io/en/v1.2/topics/streamfield.html#structural-block-types).

Here's an example of how an implementation could look:

```python
from wagtail.wagtailcore.fields import StructField, ListField, StreamField
from wagtail.wagtaildocs.blocks import DocumentChooserBlock
from wagtail.wagtailimages.blocks import ImageChooserBlock
from wagtail.wagtailcore import blocks


class RecipePage(Page):
    # StructBlock wrapper
    chef = StructField([
        ('first_name', blocks.CharBlock(required=True)),
        ('surname', blocks.CharBlock(required=True)),
        ('photo', ImageChooserBlock()),
        ('biography', blocks.RichTextBlock())
    ])
    
    # ListBlock wrapper
    ingredients_list = ListField(blocks.CharBlock(label="Ingredient"))
    
    # StreamBlock (just a regular StreamField)
    instructions = StreamField([
        ('paragraph', blocks.RichTextBlock(icon="pilcrow")),
        ('h1', blocks.CharBlock(classname="title", icon="title"))
    ])
```

timheap:
StreamFields are great, as they get around a limitation of databases, where storing a collection of different things is difficult. This flexibility is not required for most of your example, which is better served by creating a model normally, possibly using the new [Django PostgreSQL `ArrayField`](https://docs.djangoproject.com/en/1.9/ref/contrib/postgres/fields/#arrayfield):

```python
class RecipePage(Page):
    chef_first_name = models.CharField()
    chef_surname = models.CharField()
    chef_photo = models.CharField()
    chef_biography = models.ForiegnKey('wagtailimages.Image', null=True, on_delete=models.SET_NULL)

[...]

PR #2187, opened 2016-02-02
Let block subclasses override their BoundBlock classes

akx:
This makes it easier to implement blocks that have some added logic
such as properties that need to be accessible from templates.

### DIFF
--- a/wagtail/wagtailcore/blocks/base.py
+++ b/wagtail/wagtailcore/blocks/base.py
@@ -24,6 +24,28 @@ __all__ = ['BaseBlock', 'Block', 'BoundBlock', 'DeclarativeSubBlocksMetaclass', 
 # =========================================
 
 
+@python_2_unicode_compatible
+class BoundBlock(object):
+    def __init__(self, block, value, prefix=None, errors=None):
+        self.block = block
+        self.value = value
+        self.prefix = prefix
+        self.errors = errors
+
+    def render_form(self):
+        return self.block.render_form(self.value, self.prefix, errors=self.errors)
+
+    def render(self):
+        return self.block.render(self.value)
+
+    def id_for_label(self):
+        return self.block.id_for_label(self.prefix)
+
+    def __str__(self):
+        """Render the value according to the block's native rendering"""
+        return self.block.render(self.value)
+
+
 class BaseBlock(type):
     def __new__(mcs, name, bases, attrs):
         meta_class = attrs.pop('Meta', None)
@@ -40,6 +62,7 @@ class BaseBlock(type):
 class Block(six.with_metaclass(BaseBlock, object)):
     name = ''
     creation_counter = 0
+    bound_class = BoundBlock
 
     TEMPLATE_VAR = 'value'
 
@@ -152,7 +175,7 @@ class Block(six.with_metaclass(BaseBlock, object)):
         bound_block.render() rather than blockdef.render(value, prefix) which can't be called from
         within a template.
         """
-        return BoundBlock(self, value, prefix=prefix, errors=errors)
+        return self.bound_class(self, value, prefix=prefix, errors=errors)
 
     def get_default(self):
         """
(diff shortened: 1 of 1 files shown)


gasman:
Hi @akx,

I'm not sure that this is a good idea - there's too much "magic" in the behaviour of BoundBlock already, where objects [act differently depending on how you access them](http://docs.wagtail.io/en/v1.3.1/topics/streamfield.html#boundblocks-and-values), and I'm concerned that allowing customisation here will make the behaviour even more unpredictable. (As a specific example of that: `StreamValue.__getitem__` [bypasses the `bind` method](https://github.com/torchbox/wagtail/blob/654347d65f4da5baa93ed412fc86f624e531b68a/wagtail/wagtailcore/blocks/stream_block.py#L298), so children of a StreamBlock won't receive the custom BoundBlock class.)

The way I see it, the BoundBlock mechanism exists primarily for the benefit of the containing block: it's essentially a way of saying "here's a block value, along with some extra information about the type of block, and how it should be rendered". With this change, control is being handed over to the child block, and the API doesn't seem as clean.

Can you give an example of how this would be used? That might help me understand the motivation for this change a bit better...

akx:
Hey @gasman, thanks for checking this out.

So, my UC is that I have a page -- the home page for the site -- with a customizable list of quick links in the sidebar. One can choose whether to have a page link, or an external link. (And some other details which don't concern the discussion here. :) ) The list is implemented as a StreamField with only one block type, the `QuickLinkBlock`.

In templates I need to get the effective URL, i.e. either the external URL, if set, or the page's URL (otherwise). The `get_url()` fn below is the ugly workaround I implemented for that purpose.

```python
class QuickLinkBlock(StructBlock):
    page = PageChooserBlock(label=_("page"), required=False)
    external_url = URLBlock(label=_("external URL"), required=False)
    text = CharBlock(label=_("link text"), required=True)

    def get_url(self, value):    
        if value.get("external_url"):
            return value["external_url"]
        else:
            return value["page"].url
```

I don't want to call back out into the template engine to have it render a sub-template for these blocks for performance reasons, and to keep the layout for the page -- it would be a template that's used nowhere else -- contained in one file.

The Jinja template for the index page has a fragment that looks like this right now.

```jinja
<ul>
    {% for link in page.quick_links %}
    <li><a href="{{ link.block.get_url(link.value) }}"><p>{{ link.value.text }}</p></a></li>
    {% endfor %}
</ul>
```

So if I could override that the `BoundBlock`s for my `QuickLinkBlock` were, say, `QuickLinkBoundBlock`, I'd be able to add a `@property` called `url`, and turn `link.block.get_url(link.value)` into `link.value.url` or similar.

Does this illuminate things? :bulb:

gasman:
OK, thanks for the clarification!

I don't think `BoundBlock` is the right place for this behaviour: there are lots of places where the value doesn't get wrapped in a `BoundBlock` in the way you expect (for example, accessing a child of a StructBlock gives you a plain unwrapped value), which would leave people wondering why their `url` property isn't available. However, it seems to me that we could make a very similar change to StructBlock, so that instead of returning StructValue instances (which behave as a plain dictionary) it could return a custom object with the additional methods:

```python
class QuickLink(StructValue):
    @property
    def url(self):
        if self.get("external_url"):
            return self["external_url"]
        else:
            return self["page"].url

class QuickLinkBlock(StructBlock):
    page = PageChooserBlock(label=_("page"), required=False)
    external_url = URLBlock(label=_("external URL"), required=False)
    text = CharBlock(label=_("link text"), required=True)

    class Meta:
        value_class = QuickLink
        # there might be a better name than value_class...
```

Does this look like it would do what you need?

akx:
Yeah, I think that would work fine.
Really, anything that'd make the resulting Jinja code something like the obvious

```jinja
<ul>
    {% for link in page.quick_links %}
    <li><a href="{{ link.url }}">{{ link.text }}></a></li>
    {% endfor %}
</ul>
```

(notice the lack of `.value.text` in the link text too)

would be great. :D

gasman:
Thanks! I don't think there's a good way to drop the `.value.text` bit - for most purposes, the extra level of indirection is a good thing, so that you can check 'what type of block is this?' before digging into its properties. Your example is a bit of a special case, as you only have one block type - a `ListField` as proposed in #2048 would solve this nicely though!

I'll close this ticket now, and open a new one for the alternative StructValue approach.

PR #2190, opened 2016-02-03
Switch to using pip -e instead of running setup.py develop

nek4life:
In going through getting started with developing on wagtail I ran into some issues installing Pillow when the requirement was installed using setup.py develop on OS X El Capitan. I was missing jpeg, installed that, then zlib which was already installed and was still giving me compilation errors.

When using pip Pillow distributes binaries for OS X using the wheel format. So installing the setup.py using pip with the -e flag does the same thing as setup.py develop, but uses wheels instead. I took this a step further and added the requirements-dev.txt dependencies to the setup.py file using the extras_require option so you can install all the dependencies using pip install -e .[testing,docs] to install all the testing and documentation requirements all at the same time.

I also updated the documentation to reflect this and removed the reference to setup.py develop as well as the requiremens-dev.txt as it's no longer needed.


.gitignore
```diff
@@ -11,5 +11,11 @@
 /venv
 /node_modules/
 npm-debug.log
-*/.idea/
+/.gitignore
```
tomdyson: Why remove .gitignore?

tomdyson:
Thanks, @nek4life!

kaedroho:
Looks good to me (except for ``.gitignore`` being added to ``.gitignore`` which @tomdyson mentioned above)

nek4life:
Sorry it was late. I think I must have accidentally added that. Should I just add a new commit or fix the commit and reopen a new PR?

kaedroho:
A new commit would be fine, thanks

m1kola:
Good idea! Looks great!

Thanks, @nek4life!

gasman:
Works nicely here, thanks!

I suggest we keep `requirements-dev.txt` in place for now, and mark it as deprecated, so that our existing instructions (and scripts like https://github.com/torchbox/vagrant-wagtail-develop) don't immediately break when this gets merged.

nek4life:
I mainly removed it so the dependencies weren't listed twice and would avoid them potentially getting out of sync. If that's deemed to be a smaller issue than the existing scripts breaking I'll be happy to add it back in.

gasman:
No problem - have merged now (with requirements-dev.txt reinstated and marked as deprecated) in ae2006ccf114cbcda1e4b31358d26b75a3f1c897. Thanks!

PR #2192, opened 2016-02-03
Replace {% load static %} with {% load staticfiles %}

kaedroho:
This allows the wagtailadmin to take advantage of Django's staticfiles storage backends such as ``ManifestStaticFilesStorage``.

This builds on my previous pull request that removes the dependency on Django Compressor, only the last three commits are relevant.

This adds a dependency on Django's ``staticfiles`` contrib module (which, unless manually removed, should be in every Wagtail site's ``INSTALLED_APPS`` anyway).


kaedroho:
I've pushed this (and #2132) up to the revisions UI demosite. Details here: https://github.com/torchbox/wagtail/pull/2170#issuecomment-176687991

kaedroho:
Ready to review!

tomdyson:
Looks good to me, and works well on the demo site. Nice one @kaedroho!

wagtail/wagtailadmin/menu.py
```diff
@@ -104,7 +105,7 @@ def __init__(self, label, menu, **kwargs):
 
     @property
     def media(self):
-        return Media(js=['wagtailadmin/js/submenu.js']) + self.menu.media
+        return Media(js=[static('wagtailadmin/js/submenu.js')]) + self.menu.media
```
m1kola: I'm not sure that using of templatetags in Python code is good idea. Maybe something like this will be better?
```python
from django.contrib.staticfiles.storage import staticfiles_storage
staticfiles_storage.url('wagtailadmin/js/submenu.js')
```

timheap: Djangos Media class already calls `static` on any paths defined, so wrapping the filename here in `static` is not needed: https://github.com/django/django/blob/master/django/forms/widgets.py#L80

timheap:
Replacing `{% load static %}` with `{% load staticfiles %}` is correct now, but the `staticfiles` templatetags are being deprecated in Django 2.0. There is a large compatibility window here, so this is not an issue yet, but it does mean that it will be annoying to support Django 1.9 and below, and Django 2.0 and above at the same time. I dont think there is any way around this, and I think the current approach is the right thing to do, but this will need dealing with eventually.

wagtail/wagtailadmin/menu.py
timheap: Actually, ignore me. This was fixed in master on Django, but 1.9.x and below still just prepend `STATIC_URL`. Wrapping the call in `static()` is the correct thing to do. Looking at the templatetag code, it seems that `static` is not used as a template tag, and is there just to provide a nice wrapper around `staticfiles_storage.url()`. The actual `{% static %}` templatetag code even calls the `static()` function instead of calling the full `staticfiles_storage.url()` function.

timheap:
Is it worth mentioning the new dependency on `django.contrib.staticfiles` in the 1.4 Upgrade Consideration release notes? Most people probably have it already, but it is good to mention in case they do not.

PR #2200, opened 2016-02-04
Remove final references to django-compressor

### LINKED ISSUE #2197: Update SCSS comment re use of relative paths in static-root / font-root
https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailadmin/static_src/wagtailadmin/scss/_variables.scss contains a note about using relative paths for static-root and font-root to keep django-compressor happy. Clearly, this information has been outdated ever since we started using node-sass for SASS compilation instead of django-compressor... We should update this comment to reflect how our node-sass toolchain deals with these paths - and if it allows us to remove a hack, do so :-)

timheap:
* Remove references from docs
* Fix comment in SCSS files, and fix related variable names
* Remove COMPRESS_ENABLED setting from tests

See #2132
Fixes #2197

### DIFF
--- a/docs/advanced_topics/performance.rst
+++ b/docs/advanced_topics/performance.rst
@@ -25,8 +25,6 @@ We recommend `Redis <http://redis.io/>`_ as a fast, persistent cache. Install Re
 	    }
 	}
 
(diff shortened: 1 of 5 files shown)


m1kola:
Looks good! Thanks, @timheap!

gasman:
Merged in 224f3f5c37468393e77e78b83f7ece4f8d85081c - thanks!

Issue #2164, opened 2016-01-28
Proposal of a full-width userbar instead of "edit bird"


alexgleason:
I feel like there's a better way to do this. WordPress for instance puts a solid black bar at the top of your site when you're logged in, pushing the whole rest of the site about 30px down. This way you don't have to worry about the userbar obstructing your mouse clicks. It's annoying to have to figure out the right place to put it on a site (usually not at the top) and even then it looks weird and isn't easily made to fit into the bottom corners.

I think a horizontal bar option (similar to WordPress) would be better. What do you all think?

JoshBarr:
It's a good idea. Pretty unobtrusive if it's only 30px. 

However, if your site includes a fixed position header, the user bar would overlap it. Any idea how Wordpress deals with that?

alexgleason:
@JoshBarr Like this:

```css
html {
    margin-top: 32px !important;
}

#wpadminbar {
    height: 32px;
    position: fixed;
    top: 0;
    left: 0;
    width: 100%;
    min-width: 600px;
    z-index: 99999;
}
```

Essentially just pushes the whole site down by 32px to create a gap, then puts the bar in the bap.

JoshBarr:
It won't push down any `position: fixed` elements, but at least the massive z-index should help the admin bar stay on the top :+1:

alexgleason:
Ah yeah, good point. I've never had a site collide with it before. But in the rare case it can be worked around with something like

```css
html.admin-mode .my-fixed-obj {
    top: 32px;
}
```

Or if you were really paranoid use some JS that checks for the existence of the bar then loops through every element giving it a margin-top of + 32px if it's position is fixed. Not recommending, just sayin'.

davecranwell:
I think a bar the full width of the site, even if it's only 30px high, seems a bit overkill for just 3 buttons (5 during moderation), but I'd be open to ideas. Having the bar where it currently is _does_ sometimes get in the way of a search box etc, but moving it is trivial.

I too had imagined something like wordpress or drupal's bar when Wagtail was young, but in the event, there was so little to add to it, there was no need for it. I'm also mindful of how intrusive even a small bar can be on the brand of the resulting site. I'd also be interested to avoid the bar approach purely from the over-familiarity aspect - it's what everyone does.

[...]

PR #2201, opened 2016-02-04
Improved Edit Bird/Userbar

mr-winter:
We built a non-iframe version of the Wagtail userbar, according to the quick mockup we proposed earlier this week. This addresses #2164 and still needs some updates on the tests wagtail.wagtailadmin.tests.test_userbar.TestUserbarTag


tomdyson:
Thanks @mr-winter! This looks great.

wagtail/wagtailadmin/static_src/wagtailadmin/js/userbar.js
```diff
@@ -1,18 +1,33 @@
 'use strict';
 
 document.addEventListener('DOMContentLoaded', function() {
-    var body = document.querySelectorAll('body')[0];
-    var nav = document.querySelectorAll('nav')[0];
-    var className = 'ready';
-    var hasPostMessage = window.postMessage;
+    var trigger = document.querySelectorAll('[data-wagtail-userbar-trigger]')[0];
```
akx: If you're only looking for the first object to match, you can use `.querySelector` instead.

mr-winter: True! I fixed that. Thanks

wagtail/wagtailadmin/templates/wagtailadmin/userbar/base.html
```diff
@@ -1,24 +1,19 @@
-{% extends "wagtailadmin/skeleton.html" %}
-{% load staticfiles i18n %}
-
-{% block titletag %}{% trans 'User bar' %}{% endblock %}
-{% block css %}
+{% load compress static i18n %}
```
kaedroho: ``static`` should be ``staticfiles``. And ``compress`` should be removed

JoshBarr:
I've made a few changes to the CSS + HTML, to ensure all the classNames are prefixed with `.wagtail-`, including all the icon classes. There's some duplication but I'll take your guidance @davecranwell as to whether we change the `_icons.scss` partial to use this sass factory method. 

Namespacing the classNames should help avoid style conflicts with the embedding page. We might want to make the namespace easier to change, and set it something more obscure, like `.wtCMSUserBar-`? 

Also added configurable corners to the userbar: `{% wagtailuserbar 'bottom-right' %}`.

Parameters accepted are:

```
top-left
top-right
bottom-left
bottom-right
```

That's all for now.

JoshBarr:
Gah, still need to fix the lint errors. I swear the CSS isn't that badly formatted :)

davecranwell:
One minor thing we realised while reviewing this today: the new and super-useful "Show in explorer" button will presumably not work if the page in question isn't on the first page of pages? i.e the Explorer paginates 50 page items at a time (iirc) and if the page in question is on pagination page 2, it couldn't be "Shown" as the phrase suggests it might. I wonder if theres's a slightly different version of this phrase we could use instead.

JoshBarr:
Good spotting!

Is there a way to do reverse lookup for paginated items in Django? It would
be great to append the query parameter to the URL so it jumps to the right
page.

If not, we could say something like "show parent in explorer", "go to
parent", "show siblings" or even "explore". Not sure any of those are quite
right.

tomdyson:
> Is there a way to do reverse lookup for paginated items in Django?

Not that I know of, and it sounds pretty fiddly. FWIW I don't think it matters much if the 'show in explorer' button takes you to the first of many pages. Not enough to block this PR, anyway! Great work, @mr-winter and @JoshBarr.

gasman:
I reckon working out the page number of a given item in a queryset is doable, and would make a neat addition to the django.core.paginator API: check what field(s) the queryset is ordered on, do a 'count' query to find the number of items whose value on that field is less than the current item's, and divide by the page size. But yep, definitely out of scope for this, and not a blocker for merging!

tomdyson:
@JoshBarr can you please merge this if you're happy with everything?

PR #2205, opened 2016-02-05
Prevent email send errors from crashing the admin pages

fozcode:
At the moment, failure to send an email (from the publish/approve/reject notifications) throws an unchecked exception and so causes an error 500, even though the main action itself succeeded.  Email servers can be a bit flakey; it seems a shame for some external problem to break Wagtail.

This PR catches email exceptions and returns an error message on-screen instead.

### DIFF
--- a/wagtail/wagtailadmin/utils.py
+++ b/wagtail/wagtailadmin/utils.py
@@ -15,7 +15,7 @@ from wagtail.wagtailcore.models import Page, PageRevision, GroupPagePermission
 from wagtail.wagtailusers.models import UserProfile
 
 
-logger = logging.getLogger('wagtail.wagtailadmin')
+logger = logging.getLogger('wagtail.admin')
 
 
 def get_object_usage(obj):
--- a/wagtail/wagtailadmin/utils.py
+++ b/wagtail/wagtailadmin/utils.py
@@ -1,3 +1,4 @@
+import logging
 from functools import wraps
 
 from django.template.loader import render_to_string
@@ -12,6 +13,9 @@ from modelcluster.fields import ParentalKey
 
 from wagtail.wagtailcore.models import Page, PageRevision, GroupPagePermission
 from wagtail.wagtailusers.models import UserProfile
+
+
+logger = logging.getLogger('wagtail.wagtailadmin')
 
 
 def get_object_usage(obj):
@@ -169,7 +173,7 @@ def send_notification(page_revision_id, notification, excluded_user_id):
         # Get submitter
         recipients = [revision.user]
     else:
-        return
+        return False
 
     # Get list of email addresses
     email_recipients = [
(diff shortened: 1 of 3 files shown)


wagtail/wagtailadmin/utils.py
```diff
@@ -14,6 +15,9 @@
 from wagtail.wagtailusers.models import UserProfile
 
 
+logger = logging.getLogger('wagtail.wagtailadmin')
```
kaedroho: Can the logger be called ``wagtail.admin`` (to make it consistent with [wagtail.core](https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailcore/models.py#L47) and [wagtail.frontendcache](https://github.com/torchbox/wagtail/blob/master/wagtail/contrib/wagtailfrontendcache/backends.py#L11))

wagtail/wagtailadmin/utils.py
```diff
@@ -194,12 +198,22 @@ def send_notification(page_revision_id, notification, excluded_user_id):
...
+            context["user"] = recipient
+
+            # Get email subject and content
+            email_subject, email_content = render_to_string(template, context).split('\n', 1)
+
+            # Send email
+            send_mail(email_subject, email_content, [recipient.email])
+            sent_count += 1
+        except Exception as e:
+            logger.error(
```
kaedroho: We could use [``logger.exception``](https://docs.python.org/2/library/logging.html#logging.Logger.exception) here. This would make useful debugging information (eg, a traceback) available to the log handler.

kaedroho:
Thanks for taking this on @fozcode!

Just one thing, I'm wondering if there's much benefit to telling the user that a notification email failed to send? They're not told who should and who hasn't got the notification, which makes that error not very helpful.

Unrelated to this PR, but I think there's a few other places where this would be really helpful:
 - Failing to send emails from forms (#2231)
 - Failing to update Elasticsearch (on page, image or document create/edit/delete)
 - Failing to purge the cache (on page, publish/unpublish)

fozcode:
Thanks @kaedroho for the feedback!

Using `logger.exception` would be better (and neater), agreed.

I did wonder about the error message.  On the one hand I wanted to keep this change simple (it's code you hope won't run!), but on the other I thought that the user should really be told that the notifications weren't sent.  You're right though, without access to the logs it's not a helpful message as it stands.  What is your preference?  It wouldn't be too much work to have the message list the email addresses that failed (and perhaps those that worked).  It'll probably look OK providing there aren't more than 10 or so addresses, and we could truncate the list if there are.

I hadn't considered all the other places with a similar problem, and unfortunately I don't have the time available at the moment to work on them all.  Do you want to close this PR in favour of a common solution for all those areas?  (I won't be offended!)  Or do you want to continue with this PR and then address the others separately?

gasman:
Merged in b0f49e5c0fe4b0eb3a944295aa75797a85b60d94 (along with a tweak in 25165ea198203a8b55dff94beaa8892cf2b0d0ea to suppress the error output when running the test) - thanks!

Issue #1874, opened 2015-10-28
Tags field needs universal help text to explain need to quote multi-word tags


davecranwell:
Wagtail's tagging system requires multi-word tags to be quoted, to prevent a two-word tag becoming two separate tags. This appears to catch users out repeatedly and has generated several Issues in recent months.

The tag field widget, wherever used (page models, images, documents etc) should have standard help text (appearing beneath it like all help text) explaining the need for quoting. Explaining the need for comma-delimiting might also be worth adding while we're at it.

khink:
I'd like to work on this today. I'm at the Django under the Hood sprint. We have about 3 hours of sprinting left.

khink:
starting on this (wagtail space sprint, ticket destroyer team)

khink:
I think this can be closed, as https://github.com/wagtail/wagtail/issues/3488 makes the default behaviour different.

gasman:
I think it's still a good idea to have help text explaining the behaviour, whatever that behaviour is... I guess that with #3488 in place, we'd need two different help messages depending on the value of `TAG_SPACES_ALLOWED`.

amajai:
Hello @lb- . I would like to have a shot with this issue.

lb-:
@amajai  no need to ask permission, review the comments and the PRs that are linked and give it a go :)

Issue #1438, opened 2015-06-23
How do I tag with multiple word tags


lee-kagiso:
I looked at this issue https://github.com/torchbox/wagtail/issues/916, but I can't find a solution to my problem.

I want to do multiple word tags: Like `Fred Flinstone`, `Caveman`
However I get `Fred` `Flinstone` `Caveman`. The moment I type space a tag is identified.

If I use double quotes, I get double quotes appearing in the tag.

SalahAdDin:
I have understood that in the last django-taggit version had fixed this problem, and is possible use multiple words in a tag, but, i don't know is real.

lee-kagiso:
I am upgraded to the latest, and it didn't make a difference, the change log also made no mention of any such improvement

Ricemast:
@lee-kagiso Using double quotes (`"`) to surround your multiword tag should work just fine. It only shows those quotes in the WT admin interface, they are not saved in the DB nor will they be shown when you render them.

![screen shot 2015-07-06 at 11 18 23](https://cloud.githubusercontent.com/assets/3462749/8520267/c1948a76-23d0-11e5-9bda-e059e85e60a8.png)
![screen shot 2015-07-06 at 11 18 03](https://cloud.githubusercontent.com/assets/3462749/8520271/c8b068de-23d0-11e5-891d-1a35aa7832e5.png)
![screen shot 2015-07-06 at 11 17 46](https://cloud.githubusercontent.com/assets/3462749/8520275/cb5123bc-23d0-11e5-8748-164608f7b114.png)

lee-kagiso:
@Ricemast thanks!

SalahAdDin:
But, i think that don't should be neccessary put the "" for put a multiword tag.

Issue #824, opened 2014-11-19
Not able to save a single tag with multiple words in it


sbast:
When you use Tagging feature for Pages (as in http://wagtail.readthedocs.org/en/v0.8.3/core_components/pages/model_recipes.html#tagging) you can't save a page with only a single tag containing multiple words in it.

Steps to reproduce:
1. Start editing a page in Wagtail admin.
2. Ensure tags field is empty.
2. Start typing a single tag with a double quote and enter multiple words, then close it with double quote.
3. Save the page.
4. Edit the page again—you will see multiple tags instead a one with multiple words.

The problem is with tag-it.js which doesn't put a comma after a single tag (neither double quotes it by default) when POSTing to Wagtail so django-taggit treats it as multiple tags.

Fix is coming soon with a pull request from Tivix team.

dchaplinsky:
Seems that this is broken again.

I wonder if you should set allowSpaces to true in tagit plugin config?

dchaplinsky:
To not to hotpatch existing version I've ended up with this class:

```
class AdminMultiTagWidget(widgets.AdminTagWidget):
    def render_js_init(self, id_, name, value):
        return "$('#' + {0}).tagit({{autocomplete: {{source: {1}}}, allowSpaces: true}});".format(
            json.dumps(id_),
            json.dumps(reverse('wagtailadmin_tag_autocomplete')))
```

And it nailed it.

gasman:
@dchaplinsky This is working for me on the current version when I follow the exact steps given by @sbast. Are you doing the same, or something different?

Personally I'm against setting `allowSpaces: true` - it breaks the "classic" tagging behaviour of being able to enter lots of single-word tags quickly as a space-separated list.

gasman:
Re-closing, as I can't reproduce the problem as stated on the current version - please reopen if you have more info on how to reproduce.

PR #2207, opened 2016-02-08
Allow spaces in tag fields

alexgleason:
By default, when adding tags, you can't add tags with spaces unless you explicitly surround them with "quotes". This PR just adjusts a setting on the jQuery UI tag-it library to allow for spaces by default. Tags are delimited with a comma.


alexgleason:
Does anyone disagree with this or can we merge? I'm pretty sure this is a better change.

For instance, say you wrote a blog post and you want to tag it as `cool stuff`. You can't currently do that. You _can_ write `"cool stuff"` and it works, but that's a little weird that you have to write quotes, and there's no instructions for that on the tags field. You can't do it because pressing the space key advances to a new tag.

This change just lets you press the space key in a tag name. You have to press comma to advance. Which is what the instructions on the field say anyway.

gasman:
Personally I'm not a fan of spaces in tags, and I much prefer the ability to enter a list of one-word tags quickly with space as the separator. But given previous tickets (#530, #824, #1438, #1874) I suspect I may be in the minority there, and my judgment is probably clouded by my general feeling that developers would have a much easier time dealing with tagging if everyone could just stick to lower-case URL-safe characters :-)

With that in mind, I wouldn't be too averse to changing it, although I'm wary that this might be a matter of personal preference (and if we change it, we'll just get an equal number of complaining / confused people who were silent until now because it just worked for them)...

Have you tested that this change behaves as expected when the form is saved (e.g. do the tags still show up correctly when you revisit the form)? We've had a few issues arising from our client-side (jquery.tag-it.js) and server-site (django-taggit) libraries not interacting perfectly (#824, #1970) and I'd like to be sure we're not going back to square one by making this change.

tomdyson:
Note we already have a setting for one type of tag behaviour (case sensitivity: http://docs.wagtail.io/en/v1.3.1/advanced_topics/settings.html#case-insensitive-tags). Should space support also be configurable?

ababic:
I'd have to agree with @alexgleason here. To me, it feels like a bug when you try to add spaces to a tag, but can't. Some terms are only relevant when the words are together, and the workaround is fiddly and easy to forget.

alexgleason:
Hi @gasman, can we have this merged? It's needed for a current project. Thanks. :smile:

tomdyson:
@alexgleason can you please reply to @gasman's question:

> Have you tested that this change behaves as expected when the form is saved (e.g. do the tags still show up correctly when you revisit the form)? We've had a few issues arising from our client-side (jquery.tag-it.js) and server-site (django-taggit) libraries not interacting perfectly (#824, #1970) and I'd like to be sure we're not going back to square one by making this change.

and my:

> Should space support also be configurable?

alexgleason:
@tomdyson @gasman 

>Have you tested that this change behaves as expected when the form is saved (e.g. do the tags still show up correctly when you revisit the form)?

I did check this just now; it works perfectly fine.

>Should space support also be configurable?

I don't think so. The other taggit setting you referenced is there because of a global database inconsistency among different types of databases. In this case it is a user preference, but I think the vast majority of the time you will not want to have tags with spaces so it makes sense to make this assumption. There is also the option to use quotes to let your tags have spaces, which works fine.

gasman:
Happy to go ahead with this change, but I think it does warrant a settings flag. Even if the new behaviour is overwhelmingly preferred, there will inevitably be at least one person for whom it's a backwards step - as we found with unicode slug support (#2937), which seems like as unopinionated a change as you can get...

(obligatory XKCD link: https://xkcd.com/1172/ :-) )

tomdyson:
Hi @alexgleason - can you add a settings flag, as @gasman suggests above? Thanks!

khink:
This PR is superceded by https://github.com/wagtail/wagtail/issues/3488 and can be closed.

PR #2208, opened 2016-02-08
Fix PageQuerySet code overriding code snippet

prydie:
Code snippet refered to `EventQuerySet`, however, it was actually (more appropriately) named `EventPageQuerySet`.


gasman:
Thanks!

PR #2209, opened 2016-02-08
Display the details of server errors when they occur in the multi-image uploader

nfletton:
Now distinguishes server errors from other upload failures and displays relevant error text and code.

My editor cleaned up some trailing spaces in the CSS file which I've left in the commit for now.

selection_079


m1kola:
Thanks, @nfletton!

Code looks good for me. @davecranwell and @gasman, what do you think about UI?

nfletton:
Thanks for the feedback @m1kola. I kept the UI following the existing pattern in the multi-image uploader to keep the change constrained. This is what that component looks like when you get a mix of successful and unsuccessful uploads.
selection_081

gasman:
Merged in 03e247396bdd303b138c92ceb744eef2e45031f5 - thanks!

PR #2211, opened 2016-02-09
Fixed styleguide text to use correct classname

SableWalnut:
Made the icon with the class `pick` display the correct classname in the styleguide.


kaedroho:
Thanks!

PR #2215, opened 2016-02-09
Fixes to page ordering logic in explorer

gasman:
- keep ordering and pagination as two distinct operations
- only perform the 'annotate' hack when ordering by latest_revision_created_at
- entries with a null latest_revision_created_at should appear at the top when
  ordering oldest first
- don't uselessly call order_by twice
- actually test that the 'ordering' parameter orders the results :-)

### DIFF
--- a/wagtail/wagtailadmin/tests/test_pages_views.py
+++ b/wagtail/wagtailadmin/tests/test_pages_views.py
@@ -1,4 +1,4 @@
-from datetime import timedelta
+from datetime import datetime, timedelta
 import mock
 
 import django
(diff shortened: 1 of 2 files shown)


m1kola:
On `master` sorting by `live` or by `-live` gives the same results:
screen shot 2016-02-11 at 00 02 28

On `gasman:fix/explorer-page-ordering` I see results like this (same for `live` and `-live`):
screen shot 2016-02-11 at 00 03 25

Is this expected results or `live`\`-live` sorting also broken?

gasman:
@m1kola That's the correct behaviour (for certain values of "correct" :-) ). Sorting by `live` literally means that we are sorting by the (boolean) 'live' column in the database - this doesn't distinguish between "Live" and "Live + Draft". The only thing it guarantees is that live pages are grouped together before (or after...) non-live pages - the ordering *within* those groups is unspecified. (I suspect the difference is down to database internals: postgres tends to 'naturally' order records in the order they were last updated, and in the new simpler query, we're seeing that as the secondary order; in the old query, the `annotate` clause needed some more complex processing, causing the results to be randomised.)

Yes, it's weird and probably unhelpful behaviour - but it was like that before, and I didn't attempt to fix it here. (Likewise, ordering by content_type causes them to be sorted by content type ID, not name. You'll see that I quietly changed the unit tests to use `ordering=title`, so that we were testing with something reasonably well-defined...)

m1kola:
@gasman, thanks for explanation. From this point of view, I think this PR is ready to merge.

gasman:
Thanks @m1kola! Merged in 395cecc64225d87d009169c4fb30548a4f0cbcdf.

PR #2219, opened 2016-02-10
construct_explorer_page_queryset hook for customising explorer page listings

gasman:
Implements #2196.

~~(Incorporates #2215 and #2216)~~

PR #2220, opened 2016-02-10
Use specific page models on explorer page listing

gasman:
As per discussion on <https://groups.google.com/d/msg/wagtail/qAmI4HbJfCI/9fxvoCmMEQAJ>, this ensures that any custom URL routes defined on the page model are respected by the 'live' link.

Incorporates #2027 and #2219

### DIFF
--- a/wagtail/tests/testapp/models.py
+++ b/wagtail/tests/testapp/models.py
@@ -254,6 +254,23 @@ class SingleEventPage(EventPage):
         help_text="Short text to describe what is this action about"
     )
 
+    # Give this page model a custom URL routing scheme
+    def get_url_parts(self):
+        url_parts = super(SingleEventPage, self).get_url_parts()
+        if url_parts is None:
+            return None
+        else:
+            site_id, root_url, page_path = url_parts
+            return (site_id, root_url, page_path + 'pointless-suffix/')
+
+    def route(self, request, path_components):
+        if path_components == ['pointless-suffix']:
+            # treat this as equivalent to a request for this page
+            return super(SingleEventPage, self).route(request, [])
+        else:
+            # fall back to default routing rules
+            return super(SingleEventPage, self).route(request, path_components)
+
 SingleEventPage.content_panels = [FieldPanel('excerpt')] + EventPage.content_panels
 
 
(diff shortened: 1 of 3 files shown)


kaedroho:
One possible optimisation could be to firstly check if the url property has been overridden:

```python
if page.specific_class.url is not Page.url:
    page = page.specific()
```

gasman:
The current code uses the PageQuerySet `specific` method, which should hopefully be more efficient than calling `specific` on individual page instances. (Also, I wouldn't want to assume that custom URL routing is the only use-case for retrieving the more specific versions.)

kaedroho:
Ahh sorry, forgot about that. Ignore me!

kaedroho:
Could you rebase out the prerequisite changes that are now on master and resolve the merge conflict?

gasman:
Rebased. Tests are currently failing because the just-merged #2279's tests fail the stricter validation rules added in #2278 - will fix those now

m1kola:
I think we can merge this.

gasman:
Thanks @m1kola! Merged in a6647c33c83e7fcd27f80f57f738e4df9b80fc67

kaedroho:
We have a site with lots of large news articles and this change makes Wagtail fetch all the content for 50 of them which is slow and uses a lot of memory. I've even had a few OOM errors caused by this.

I think it might be worth reconsidering this optimisation https://github.com/wagtail/wagtail/pull/2220#issuecomment-182477615. This would definitely make things slower for people who have changed the ``.url`` property, but it give a speed improvement for those who havent (especially if they have a lot of content in their pages).

gasman:
if page.specific_class.url is not Page.url:
        page = page.specific()

This assumption, that overriding `.url` is the only reason for a `specific()` fetch, is too brittle:

* there are several overrideable functions involved in custom URLs, e.g. `get_url_parts`, and these might change in future (https://github.com/tobiasmcnulty/rfcs/blob/master/draft/0014-page-url-api.rst);
* other functionality might rely on specific, e.g. `get_admin_page_title`.

Hopefully #3355 will address the performance issue without this problem.

Issue #1167, opened 2015-04-09
Target for Richtext links


J0hn5mith:
Hi

It would be nice if one could define the target of a rich text link. I wanted to insert add a link which does not open in the current tab but rather in a new one. 

Cheers
![wagtail_-_richtext_bearbeiten](https://cloud.githubusercontent.com/assets/856276/7062754/b4932ef6-dea0-11e4-94b6-55ae30c98795.png)

gasman:
Hi @J0hn5mith,
Thanks for the suggestion! This is probably not as simple as adding a checkbox, because most of the time the decision "should this link open in a new window?" is not one for page editors to make on an individual link basis, but rather a policy set by the site implementor - for example, "always open external links in a new window" or "never open links in a new window". We'll need to think carefully about the best approach for this.

tmsndrs:
IMO this generally goes against agreed UX best practise and as @gasman says, if links are to open in a new tab/window it should be consistent across the site (or specific content areas) because a conversation has taken place between the developer and the editor/client, not on a link by link basis. 

I don't agree with everything in this article but more info on this here http://www.smashingmagazine.com/2008/07/01/should-links-open-in-new-windows/

benje:
I am also looking for a simple way to set the target link behaviour for external links. I would be happy if this was a site-wide policy or a per link basis, just a simple way of achieving it.

It seems that people are asking for it, so i'd like to add my vote for a nice solution :)

Thanks as always to the devs doing an awesome job on wagtail

nek4life:
Is there a way to accomplish this somehow without proper support from the wagtail framework? Without JS on the front-end?

findsje:
I have a similar problem as nek4live but also i.e. I have to link to CC images as part of the license agreement for the images and I don't wan't my users to leave my website if they click the link for the image.
On an unrelated note it would be awesome to be able to have caption for images in rich text paragraphs too.

gasman:
[...]

PR #2223, opened 2016-02-11
Admin PageChooser - external link window targeting and follow nofollow rel options

texpress:
Added the ability for user to select blank window (new tab) or self, as well as follow no follow options for external links added through hallo-js link button modal window.

### DIFF
--- a/wagtail/wagtailadmin/forms.py
+++ b/wagtail/wagtailadmin/forms.py
@@ -52,11 +52,23 @@ class SearchForm(forms.Form):
 
 class ExternalLinkChooserForm(forms.Form):
     url = URLOrAbsolutePathField(required=True, label=ugettext_lazy("URL"))
+    target = forms.ChoiceField(choices=( \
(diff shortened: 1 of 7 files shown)


gasman:
Hi @texpress,
Sorry, but we've chosen not to support this feature as a matter of policy - see #1167. Opening links in a new window is bad usability practice (see e.g. https://www.smashingmagazine.com/2008/07/should-links-open-in-new-windows/) and Wagtail will not include any features in its default configuration to encourage this.

We'd consider allowing sites to override this choice, but only as a site-wide configuration setting (where all external links - or all external links matching a certain pattern - consistently open in new windows) - see https://github.com/torchbox/wagtail/pull/1408#issuecomment-112066108. Wagtail is designed to be usable by non-technical users, who would not necessarily understand the meaning of `nofollow` or the technical arguments for/against `target="_blank"`, and so this decision should not be left to the individual editor.

PR #2225, opened 2016-02-11
Fix assumptions about user model fields, issue #1433

### LINKED ISSUE #1433: Admin user index page is broken for user models without last_name and first_name
I'm trying to integrate Wagtail with a site that uses a custom user model that does not have first_name or last_name fields.

The user index view in the Wagtail admin is broken, because it attempts to order by last name and first name.  I can't see any obvious way of overriding this other than rewriting the view and including my own urls.py to use that one instead.
 
Here's the line:

https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailusers/views/users.py#L47

tangram:
Wagtail should not assume User models have specific fields, as custom user models without any of the fields in the enclosed commit is possible. This pull request addresses and fixes #1433, along with a few more cases.

### DIFF
--- a/wagtail/wagtailusers/views/users.py
+++ b/wagtail/wagtailusers/views/users.py
@@ -27,7 +27,8 @@ delete_user_perm = "{0}.delete_{1}".format(AUTH_USER_APP_LABEL, AUTH_USER_MODEL_
 def index(request):
     q = None
     is_searching = False
-    model_fields = User._meta.get_all_field_names()
+
+    model_fields = [f.name for f in User._meta.get_fields()]
 
     if 'q' in request.GET:
         form = SearchForm(request.GET, placeholder=_("Search users"))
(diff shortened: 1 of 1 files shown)


wagtail/wagtailusers/views/users.py
```diff
@@ -27,40 +27,42 @@
 def index(request):
     q = None
     is_searching = False
+    model_fields = User._meta.get_all_field_names()
```
kaedroho: This method was deprecated in Django 1.8: https://github.com/django/django/blob/ae1d663b7913f6da233c55409c4973248372d302/django/db/models/options.py#L560-L561

kaedroho:
Thanks @tangram this looks good, except for https://github.com/django/django/blob/ae1d663b7913f6da233c55409c4973248372d302/django/db/models/options.py#L560-L561

tangram:
Ah, sorry. I didn't catch that. Should be ok now.

gasman:
Merged in ac5d6e4945903812ce2c9a8e1f76cce7e58ed356 - thanks!

PR #2227, opened 2016-02-11
Tagging snippets documentation fix

mQuadrics:
I'm not sure i'm right but i think there is a typo in this example. I guess there should be a AdvertTag use in TaggableManager instead BlogPageTag


kaedroho:
Thanks for fixing! Merged in d27df01815b8ec1787e4f4e248467f9269705813

PR #2232, opened 2016-02-12
Tweaked repr of search results to diffrentiate them from QuerySets and lists

kaedroho:
Search results sets are now wrapped with ``<SearchResults >``. This should make debugging through the console slightly easier as the developer can now easily see the type of sequence they are working with.


gasman:
Merged in ce918daa61d7e535238550e3b8cdb9fc17c1ab18.

PR #2233, opened 2016-02-13
Unicode slug

hanpama:
Non-arabic scripts (Japanese) do not auto-populate slug #1443

* Unicode slugs are automatically generated from page titles.
* `SlugField` uses `validate_unicode_slug` method which is same with Django 1.9's one, but I defined it again in `wagtail.wagtailcore.utils` for backward compatibility.


hanpama:
It's not working now. I will write some tests..

gasman:
Thanks! Let us know when this is ready for review.

gasman:
Superseded by #2396

PR #2243, opened 2016-02-16
Collections for image/document organisation

gasman:
https://github.com/torchbox/wagtail/wiki/Collections-RFC

Incorporates #2126. There's a lot in here, but hopefully the considerable amount of groundwork leading up to this (generic views, permission policies...) means that it's now reasonably non-invasive on the rest of the codebase :-)


tomdyson:
Hi @m1kola - are you able to review this large PR?

m1kola:
Hi @tomdyson! Yes, I can do it today after work.

wagtail/wagtailadmin/templates/wagtailadmin/shared/collection_chooser.html
```diff
@@ -0,0 +1,18 @@
+{% load i18n %}
+
+<li>
+    <div class="field choice_field select">
+        <label for="collection_chooser_collection_id">{% trans "Collection" %}:</label>
```
m1kola: We need more space for translatable strings, I think.

"Collection" after translation on russian will be "Коллекция". I have issue like show on screenshot below ("я:" were lost).
screen shot 2016-02-17 at 23 17 43

Original (":" was lost)
screen shot 2016-02-17 at 23 19 45


Chrome 48.0.2564.109. Mac OS 10.11

m1kola:
Should we update [editor’s guide](http://docs.wagtail.io/en/latest/editor_manual/index.html)?

wagtail/wagtailcore/models.py
```diff
@@ -1698,3 +1703,98 @@ class PageViewRestriction(models.Model):
...
+
+def get_root_collection_id():
+    return Collection.get_first_root_node().id
+
+
+class CollectionMember(models.Model):
+    """
+    Base class for models that are categorised into collections
+    """
+    collection = models.ForeignKey(
```
m1kola: RFC says:
> Non-empty collections should not be deletable

But now we can delete non-empty collections and all entities related to collection also will be deleted.

m1kola:
From RFC:
> Administrators will be able to assign the following group permissions to collections, through the 'Groups' area under the Settings menu:
> 
> - Use documents from
> - Use images from
> - Add documents to
> - Add images to
> - Edit documents in
> - Edit images in

I have group "Editors" and two collections: "Collection 1" and "Collection 2". I gave to editor following permissions:
collections_images_permissions

After this users from "Editors" group can add, edit, delete own images, They can't see images uploaded by other users in "Images" menu, but they can use all images from all collections in rich text editor. Is it okay?

m1kola:
Seems like search in images section for non-superusers was broken. You can reproduce this issue on wagtaildemo project with two collections and permissions like that:
collections_images_permissions

wagtail/wagtailadmin/forms.py
```diff
@@ -311,3 +314,209 @@ def clean(self):
...
+            )
+
+        if self.instance.pk:
+            # editing an existing document; ensure that the list of available collections
+            # includes its current collection
+            self.collections = (
+                self.collections | Collection.objects.filter(id=self.instance.collection_id)
+            )
+
+        if len(self.collections) == 0:
```
m1kola: I think [`count()`](https://docs.djangoproject.com/en/1.9/ref/models/querysets/#count) will be better here, because we don't need to load all records to memory here: we just want to configure form fields.

wagtail/wagtailadmin/templates/wagtailadmin/shared/collection_chooser.html
m1kola: Sorry, problem not in this template, but you can reproduce this on image editing page when `$(window).width() == 1350`.

wagtail/wagtailadmin/tests/test_collections_views.py
```diff
@@ -0,0 +1,151 @@
...
+        response = self.get(collection_id=100000)
+        self.assertEqual(response.status_code, 404)
+
+    def test_post(self):
+        response = self.post()
+
+        # Should redirect back to index
+        self.assertRedirects(response, reverse('wagtailadmin_collections:index'))
+
+        # Check that the collection was edited
```
m1kola: Did you mean "was deleted"? :sunglasses:

wagtail/wagtailcore/models.py
```diff
@@ -1438,11 +1438,16 @@ class Meta:
         verbose_name_plural = _('page revisions')
 
 
+PAGE_PERMISSION_TYPES = [
+    ('add', _("Add"), _("Add/edit pages you own")),
```
m1kola: Why do you use both `'` and `"` characters?

gasman:
> After this users from "Editors" group can add, edit, delete own images, They can't see images uploaded by other users in "Images" menu, but they can use all images from all collections in rich text editor. Is it okay?

Yes - we decided not to implement "use image" permission for this phase of development, and to make all images available in the chooser instead.

wagtail/wagtailadmin/forms.py
gasman: I don't think so - if there are >1 collections, then we'll want to iterate over `self.collections` later on, to render the dropdown. So, I think it's more efficient overall to do a full fetch here, rather than doing a `count()` now and a full fetch later.

wagtail/wagtailcore/models.py
gasman: I use `"` for human-readable / translatable strings, and `'` for internal identifiers. (I *think* that's a standard convention, but it's possible that I just made it up :-) )

wagtail/wagtailadmin/forms.py
m1kola: I forgot about rendering! In this case everything is ok.

wagtail/wagtailadmin/tests/test_collections_views.py
gasman: Oops! Fixed in c77425dacd8908337bcaeb046f3d9e3d610dae87.

gasman:
Now rebased, including fixes to bring it up to date with other new work on master: adding `on_delete` to migrations, and implementing collection choosers on the new multiple document uploader.

Search is fixed in 77fe40281e22f11486d09c12110b10773ec1ec2d. Still to do: preventing deletion of non-empty collections, and fixing the cut-off label text on the 'edit image' form. (Not sure what to do about that - any suggestions @davecranwell?)

It would be good to get the editor's guide back up to date again, but I think that's going to be a bigger job than we can sensibly deal with here unfortunately :-/

gasman:
Thanks for the comprehensive review, @m1kola!

m1kola:
> It would be good to get the editor's guide back up to date again, but I think that's going to be a bigger job than we can sensibly deal with here unfortunately :-/

I think we can update editor's guide later. I created issue #2258 for this.

> Thanks for the comprehensive review, @m1kola!

Thank you for great work! I hope I will finish reviewing on this weekend.

wagtail/wagtailcore/migrations/0028_merge.py
```diff
@@ -0,0 +1,16 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('wagtailcore', '0024_alter_page_content_type_on_delete_behaviour'),
+        ('wagtailcore', '0027_fix_collection_path_collation'),
+    ]
+
+    operations = [
+    ]
```
m1kola: Can you, please, explain meaning of this migration?

wagtail/wagtailcore/permission_policies/collections.py
```diff
@@ -0,0 +1,348 @@
...
+        """
+        Return a queryset of all collections in which the given user has
+        permission to perform any of the given actions
+        """
+        if user.is_active and user.is_superuser:
+            # active superusers can perform any action (including unrecognised ones)
+            # in any collection
+            return Collection.objects.all()
+
+        elif not user.is_authenticated():
```
m1kola: Why you don't perform `user.is_authenticated()` test in `instances_user_has_any_permission_for`?

wagtail/wagtailcore/migrations/0028_merge.py
gasman: Django requires one of these merge migrations (created with `./manage.py makemigrations --merge`) whenever there are migrations with overlapping sequence numbers, to prevent conflicts and to ensure that any later migrations are based on a fully up-to-date version of the database state.

Normally we'd renumber migrations when we merge the PR, to avoid the need for merge migrations - but in this case there's already one real-world site (or maybe more...) using an early version of collections with Wagtail 1.3, so we need to keep the numbers unchanged to avoid upgrade problems.

wagtail/wagtailcore/permission_policies/collections.py
gasman: Ah, good catch! Yep, I think that needs to be added - I guess I probably copied the logic from https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailcore/permission_policies.py#L323 (which doesn't need an `is_authenticated` test because `has_perm` will return false for non-authenticated users).

wagtail/wagtaildocs/views/documents.py
```diff
@@ -34,6 +35,16 @@ def index(request):
         ordering = '-created_at'
     documents = documents.order_by(ordering)
 
+    # Filter by collection
+    current_collection = None
+    collection_id = request.GET.get('collection_id')
+    if collection_id:
+        try:
+            current_collection = Collection.objects.get(id=request.GET['collection_id'])
```
m1kola: Is there a reason not to use `collection_id`?

wagtail/wagtailimages/templates/wagtailimages/multiple/add.html
```diff
@@ -22,6 +22,18 @@
                     <input id="fileupload" type="file" name="files[]" data-url="{% url 'wagtailimages:add_multiple' %}" multiple>
                 </div>
                 {% csrf_token %}
+                {% if collections %}
+                    <div class="field">
+                        <label for="id_addimage_collection">Add to collection:</label>
```
m1kola: Would be great to mark this string for translation. The same in `wagtail/wagtaildocs/templates/wagtaildocs/multiple/add.html`

wagtail/wagtailusers/forms.py
```diff
@@ -233,19 +239,44 @@ def save(self):
...
+        choices=PAGE_PERMISSION_TYPE_CHOICES,
+        required=False,
+        widget=forms.CheckboxSelectMultiple
+    )
 
-    class Meta:
-        model = GroupPagePermission
-        fields = ('page', 'permission_type')
 
+class BaseGroupPagePermissionFormSet(forms.BaseFormSet):
```
m1kola: Looks very similar to `BaseGroupCollectionMemberPermissionFormSet`. Can we create generic form set form or mixin (maybe not in this PR)?

m1kola:
@gasman, thanks for great feature! I have finally finished reviewing this! :metal:

In common looks nice. There is some important comments:

1. https://github.com/torchbox/wagtail/pull/2243#discussion_r53226298
2. https://github.com/torchbox/wagtail/pull/2243#discussion_r53231811 (see also https://github.com/torchbox/wagtail/pull/2243#issuecomment-186460131)
3. https://github.com/torchbox/wagtail/pull/2243#discussion_r53567343
4. https://github.com/torchbox/wagtail/pull/2243#discussion_r53575708

If would be nice, if you will create new branch from `feature/collections-1.4` (`feature/collections-1.4-fixes` for example) for further fixes\improvements and separate PR to `feature/collections-1.4`. It's to difficult to find changed parts in so large PR.

wagtail/wagtailcore/permission_policies/collections.py
gasman: `is_authenticated` check added in 3012c1c75ce3f4001557f7b94d857632de370fdb. On closer inspection it looks like this isn't strictly necessary - `AnonymousUser` has no ID and belongs to no groups, so the `GroupCollectionPermission` lookups return an empty result set as expected. It's nicer to avoid those queries entirely, though.

wagtail/wagtailcore/models.py
gasman: 01cf35f37ce7b6cfc93b4be03ef0592956d0d0bd now ensures that non-empty collections cannot be deleted.

wagtail/wagtaildocs/views/documents.py
gasman: Hmm, not sure how that happened :-) Fixed in 174fea121f11fc624ef8a5f2e85b810696d00336, and e62a760d0e4469f82679ca7232d63a5ec92d3691 for images.

wagtail/wagtailimages/templates/wagtailimages/multiple/add.html
gasman: `{% trans %}` tag added in 0a69dd88c1fd10f38480f4aa310ec988a0956dfd / 1aeb919a60916f28a524b93faeb478276dd2f445

wagtail/wagtailusers/forms.py
gasman: It could probably be done, but the two classes don't quite correspond (mainly because `GroupPagePermission` uses a 'choices' list for the permission type, rather than a real foreign key to auth.Permission), and making things fully generic has been a bit of a pain point for this PR - so I'd prefer to leave this for a future task :-)

gasman:
@m1kola Thanks again! I think the cut-off label text on the image editor (https://github.com/torchbox/wagtail/pull/2243#discussion_r53226298) is best handled as a separate issue, like #2246, as it's not specific to collections (it would affect any sufficiently long field label). Everything else is now dealt with, hopefully...

I've made all of the fixes as commits on top of the original PR (with the `--fixup` flag where appropriate, so that they can be squashed into the original commits when we merge) - you can see the comparison at https://github.com/torchbox/wagtail/compare/fbf56e92eb9eb65975f447a381effc319fa68f6c...gasman:feature/collections-1.4 (which contains the changes I made when rebasing on current master) or https://github.com/torchbox/wagtail/compare/4358fa93092afab8112612a065bebebebbaf5d91...gasman:feature/collections-1.4 (which just contains the fixes in response to your feedback). If there's anything else I can do to make the changes clearer (e.g. creating new branches), let me know!

wagtail/wagtailcore/permission_policies/collections.py
```diff
@@ -0,0 +1,348 @@
...
+        permission to perform any of the given actions
+        """
+        if not (user.is_active and user.is_authenticated()):
+            return self.model.objects.none()
+        elif user.is_superuser:
+            return self.model.objects.all()
+        else:
+            # filter to just the collections with this permission
+            return self.model.objects.filter(
+                collection__in=list(self._collections_with_perm(user, actions))
```
m1kola: Is there any reason to execute `QuerySet` returned by `self._collections_with_perm(user, actions)`? If I'm not mistaken, we can simply pass `QuerySet` to filter and ORM will prepare one query with subqueries.

m1kola: Same question for `instances_user_has_any_permission_for`

gasman: I would definitely have preferred to do it with a single query, but unfortunately this was necessary for the search mechanism to work. We call `search(query_string)` on the returned queryset - this involves converting the query filter criteria into a search query (to be passed to either elasticsearch or our ORM-based search backend), and one of the rules for this is that we can only filter on fields that have been indexed for search (using `index.FilterField`). This means we can't perform arbitrary complex subqueries here - but we can do `__in` lookups on the 'collection' field, as that field has been indexed.

m1kola:
In my opinion this PR is ready to merge. Thank you!

> you can see the comparison at fbf56e9...gasman:feature/collections-1.4 (which contains the changes I made when rebasing on current master) or 4358fa9...gasman:feature/collections-1.4 (which just contains the fixes in response to your feedback).

Thanks for suggestion! It's very convenient.

gasman:
Merged in fc3936a229255f902368e0dd7883084b92085dbb + parents.

SalahAdDin:
I have a question:
Can you put some information in collections view? Maybe image or document's count will be useful for the user.

PR #2247, opened 2016-02-17
Improvements to MANIFEST.in

kaedroho:
Reduces size of wheel files to 5.7MB from 9.3MB (wagtailadmin static was being archived twice).

The change that made the above improvement was this line ``prune wagtail/wagtailadmin/static_src``. I've also made changes to simplify it a bit as well.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,5 @@
 include LICENSE *.rst *.txt
-recursive-include wagtail *
-recursive-exclude * __pycache__
-recursive-exclude * *.py[co]
+graft wagtail
+prune wagtail/wagtailadmin/static_src
+global-exclude __pycache__
+global-exclude *.py[co]


timheap:
There are other static_src directories within the Wagtail source. These should probably be ignored as well.

kaedroho:
Why did this get bumped? It's a small change that takes 3MB off our wheel file size, surely we want that before the 1.4 release?

tomdyson:
Agreed, it seems a shame not to include this.

gasman:
OK, fair enough! It got bumped along with a bunch of other not-absolutely-essential PRs when it looked like we were 24 hours away from a release candidate last week - but since that didn't happen, I think we've got room to fit this one in...

gasman:
Merged in 9393b935389e71f3a55f90f766c05ce6186c6e67

Issue #2097, opened 2016-01-10
Wagtail userbar link text cutted off.


MechanisM:
I'm used wagtail mostly in english and wasn't noticed, but when I switched to russian, I see that link for "Edit"(in russian is "Редактировать") is cutted off.
![2016-01-10 22-52-44 window](https://cloud.githubusercontent.com/assets/121781/12223594/021f4c9c-b7ed-11e5-9738-f1e48e074415.png)

kaedroho:
Thanks for reporting!

davecranwell:
I believe this is addressed by https://github.com/torchbox/wagtail/pull/2201

PR #2249, opened 2016-02-17
Fixes text in the userbar being cut-off in a different language

jjanssen:
Corresponds with the issue mentioned in #2097 

I have widened the userbar iframe with 15px (165px instead of 150px) and gave the items within the iframe a more fluid approach. So in English the elements are smaller (as is right now), but in a translation - like Russian - where the character range can get a maximum of 15 characters.

As is:
image

Translated:
image

I think this is sufficient for now because I have tested a few other languages (including the moderation items), but if it turns out there are still labels exceeding we could have a look into a fluid iframe approach (calculating the inner contents). 

Cross-browser tested in:

- Edge
- IE 9+
- Chrome 31+
- Firefox 31+
- Safari 7.1


gasman:
Hi @jjanssen - thanks for your work on this! However, we've arrived at the conclusion that the iframe approach isn't really sustainable here (since increasing the iframe size means blocking a larger clickable area of the original page) - and so we're looking at changing to a non-iframe-based solution for Wagtail 1.4 - see #2164, #2201

jjanssen:
Yeah, closed it myself. Noticed this one too late #2201
Thanks for the heads up!

PR #2250, opened 2016-02-17
Catch and display OS errors that occur during document updates

nfletton:
Displays OS errors in a user friendly way when creating, updating and deleting documents in the main document section of wagtail.

selection_086

### DIFF
--- a/wagtail/wagtaildocs/views/documents.py
+++ b/wagtail/wagtaildocs/views/documents.py
@@ -2,6 +2,7 @@ from django.shortcuts import render, redirect, get_object_or_404
 from django.utils.translation import ugettext as _
 from django.views.decorators.vary import vary_on_headers
 from django.core.urlresolvers import reverse
+from django.db import transaction
 
 from wagtail.utils.pagination import paginate
 from wagtail.wagtailadmin.forms import SearchForm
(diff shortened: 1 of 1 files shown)


wagtail/wagtaildocs/views/documents.py
```diff
@@ -111,21 +116,26 @@ def edit(request, document_id):
...
+                if 'file' in form.changed_data:
+                    # if providing a new document file, delete the old one.
+                    # NB Doing this via original_file.delete() clears the file field,
+                    # which definitely isn't what we want...
+                    original_file.storage.delete(original_file.name)
+
+                with transaction.atomic():
+                    doc = form.save()
+            except OSError as err:
+                report_os_error(request, err)
```
kaedroho: Catching the error here will prevent it from being logged which will prevent error reporting (via email, sentry, etc) from happening. This sort of error is definitely one that an ops team would be interested in so we must reinstate the logging somehow.

gasman: That's a fair point, actually! While it makes sense to report image upload errors back to the user in a friendly way (because otherwise they'll be buried by the AJAX call), I'm not sure that's the case for the main add/edit document form - arguably, if there's a legitimate server-level error then it's not really Wagtail's responsibility to 'prettify' it, and we should leave it to fail in a predictable plain-vanilla way (i.e. this PR should be closed as 'won't fix').

gasman:
Closing as per https://github.com/torchbox/wagtail/pull/2250#discussion_r63004377 so that this isn't left in limbo (but happy to hear other thoughts on this, and willing to reconsider / reopen if that's the way the consensus goes...)

PR #2254, opened 2016-02-18
Enable template debugging on tests, to catch broken includes

gasman:
This threw me for a while just now while merging #1892 - the add_multiple.html template contained the line `{% include "wagtailadmin/shared/tag_field_css.html" %}` which is a file that no longer exists. This caused the view to break, but the tests failed to reveal this, apparently because they were rendering templates in production mode (where missing includes are silently ignored)...


kaedroho:
Merged in 6bae4f0 Thanks!

PR #2256, opened 2016-02-18
WagtailTestUtils allow custom user models

AdamBolfik:
This should allow for user models which use email instead of username as the USERNAME_FIELD

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -5,6 +5,7 @@ Changelog
 ~~~~~~~~~~~~~~~~
 
  * The page editor now produces a warning if the user navigates away without saving changes
+ * 'Add document' interface now supports uploading multiple documents at once
  * The set of pages displayed in the page explorer can now be customised with a new hook called `construct_explorer_page_queryset`
  * The `Document` model can now be overridden using the new `WAGTAILDOCS_DOCUMENT_MODEL` setting (Alex Gleason)
  * Wagtail no longer depends on django-compressor
--- a/docs/releases/1.4.rst
+++ b/docs/releases/1.4.rst
@@ -16,6 +16,12 @@ Protection against unsaved changes
 The page editor interface now produces a warning if the user attempts to navigate away while there are unsaved changes.
 
 
+Multiple document uploader
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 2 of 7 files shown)


timheap:
This will still fail with more exotic custom users, such as users from an LDAP or other external source for example. May I suggest making a new method `WagtailTestUtils.create_test_user()` that users can override in their own test suite if required, but falls back to this flexible code you've added?

Additionally, instead of trying to find a user based on the `USERNAME_FIELD`, perhaps store the `pk` of the test user and use that to find the user in the DB instead. All Django models must have a primary key, so this is always safe:

```python

class TestAccountSection(TestCase, WagtailTestUtils):
    def setUp(self):
        self.user = self.login()

    # ...

    def test_change_password_view_post(self):
        # ...
        # Check that the password was changed
        user = get_user_model().objects.get(pk=self.user.pk)
        self.assertTrue(user.check_password('newpassword'))
```

AdamBolfik:
That makes more sense, thanks.

AdamBolfik:
@timheap something odd, Travis seems to be having problems with the elasticache backend at different times. My first CI run [#5657](https://travis-ci.org/torchbox/wagtail/jobs/110414420) erred on an HTTP exception and after adding a comment to force Travis to run again, it fails at a different spot with elasticache. [#5658](https://travis-ci.org/torchbox/wagtail/jobs/110414420)

Are you guys experiencing problems with your test server? or have I created some sort of randomly generated bug? :smile:

tomdyson:
@AdamBolfik yes, Travis sometimes has issues with Elasticsearch. The tests ran fine on our Drone CI server - http://ci.torchbox.com/torchbox/wagtail/193 - so I don't think we need to worry about the failure.

AdamBolfik:
Alright, thanks. Is there any way to mark this as 'good' so it doesn't just get overlooked?

gasman:
Have restarted the failed build now - don't worry, we won't overlook pull requests on the basis of those CI failures.

timheap:
The tests passed this time, so this all looks good to me!

gasman:
Thanks @AdamBolfik! Merged in c6140a06f78228c79ba3a30729ac8a6b2f2c52a7 + parents.

Issue #2239, opened 2016-02-15
Date picker widget should highlight current date at all times


davecranwell:
When the datepicker opens on an clean date field, the modals opens on the current month, with the current day highlighted in the salmon colour.

Once a date is selected, assuming that date isn't today, the picker will never again colour-code the current date if re-opened. It will show the date selected in the field, but not the current date. As someone rarely aware of the actual date, I find this endlessly frustrating.

There's one other idiosyncrasy worth addressing when this issue is handled: When the picker is opened on a clean field, the current date e.g 15, is highlighted, but the number 15 remains highlighted in each of the months you choose, regardless of the actual current month. I can't work out why this would be useful and again this causes confusion.

timheap:
Fixed by #2260

PR #2260, opened 2016-02-20
Datetimepicker improvements

thrawny:
Fixes for #2239 

Current date is now highlighted if it is not selected.

I also incorporated changes that remove the highlighting of the same day as was previously selected when switching months. This works for both the datepicker and datetimepicker. However in order to change both date and time with the datetimepicker one still has to open the modal twice (or use the text input).


davecranwell:
Reviewed. Looks good to me.

timheap:
Merged in 03cba5b91c3feb0a57e6cd39e6504397e51321cc...e001cac581042c54ddf7404c765b16c27e75f03c. Thanks!

thrawny:
Cool! Thanks a alot.

PR #2266, opened 2016-02-22
Fixed the copy function to look in the right place for a parent_link

SableWalnut:
Currently, if you try to copy a `Page` while using a `OneToOneField` of any kind, it will fail. The code iterates over each field of `Page` checking to see if it is a `OneToOneField` and then checking the attribute `field.parent_link` which does not exist, it should instead be looking for `field.rel.parent_link`.


gasman:
Thanks @SableWalnut! Would you be able to add a test for this, please? I'm guessing you could probably base it on b77dfd20d792bdfc47366d0b5a566f6f84ff60d8 and 9db24387fe0f64810c871d2ad6a98abc3813b663 (which were similar fixes for copying pages with other kinds of relation).

SableWalnut:
Sure thing! Just added them @gasman! :)

timheap:
Merged in a1a294049519ea9cba24c554ad0c0ab13731b3f1...0ead79d286b6c74503c0de6efd8acaab8cb291eb. Thanks!

PR #2269, opened 2016-02-23
Show a message to users who do not have permission to access admin 

### LINKED ISSUE #1486: Login attempts result in different behaviour for users known by django but with insufficient permissions
1. create a new user in wagtail,
2. assigned it to a group which has no admin access
3. try to log into wagtail admin as that user

The result is a redirect back to the login without any feedback whatsoever.

I'd expect the user to receive the same error response as someone logging in with completely unknown credentials. A difference in results indicates to anyone probing the site maliciously, that a set of tested credentials matches a real user.

timheap:
Fixes #1486.

As part of doing this, I found a bug in the admin access tests where a user was not logged correctly, but this was not caught by the tests. I have fixed the tests to assert that users are actually logged in.

### DIFF
--- a/wagtail/contrib/settings/tests/test_admin.py
+++ b/wagtail/contrib/settings/tests/test_admin.py
@@ -21,7 +21,7 @@ class TestSettingMenu(TestCase, WagtailTestUtils):
             username='test', email='test@email.com', password='password')
         user.user_permissions.add(Permission.objects.get_by_natural_key(
             codename='access_admin', app_label='wagtailadmin', model='admin'))
-        self.client.login(username='test', password='password')
+        self.assertTrue(self.client.login(username='test', password='password'))
         return user
 
     def test_menu_item_in_admin(self):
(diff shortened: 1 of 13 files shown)


davecranwell:
Is this update still giving users different messages depending on status? That still seems like a leak to me.

tomdyson:
Nice one @timheap!

gasman:
@davecranwell With this change, we go from no message at all when logging in as a valid-but-no-wagtailadmin-permission user, to getting a "You do not have permission to access the admin" message. Given that you have to provide a valid username/password to see the difference, I don't think there's a risk of information leakage - that's more of a concern for password reset forms (if you're thinking of the same issue I'm thinking of).

davecranwell:
Ah yes, of course. If the password would have to be correctly entered, then leaking whether they're an admin or not is the least of our worries :)

gasman:
Merged in #a2bc7341f27ec774c50381cc409b6f6376daf3a8 + parents - thanks!

Issue #1564, opened 2015-07-30
Form builder as StreamField block


mmospanenko:
Is it possible?

iho:
+1

SalahAdDin:
:+1:

gasman:
I'd be very happy to see this implemented. I suppose the big open question is how form submissions can be handled - on the existing form builder, submissions are POSTed to the page's own URL, but an individual StreamField block doesn't have a URL of its own, or anything that can persistently identify it.

iansprice:
+1! 
Looking to implement this myself and may just give it a shot. Any feedback on the structure below?

How about setting up a FormBuilder block that, regardless of page it is rendered on, submits to a single site-wide url that takes the PK as a GET param? Much like the URL that page previews rely on, perhaps a unique sequence of symbols that would almost never be replicated would

timheap:
I would let the developer decide where the form gets submitted to. In some cases, submitting to the current page itself would be appropriate - if the page has extended the Wagtail form builder base class. In other cases, submitting to a central URL, as in your example; or even a completely different page on a different domain such as a MailChimp subscription could be appropriate. Finally, the form could be AJAXified for completely different behaviour, meaning that it never gets submitted anywhere in particular.

If the developer is left to write a template that handles submitting the form, they can write the code that suits their use case very easily:

```html+django
<form action="{{self.url}}" method="post">
    {{ self.stream_body }}
    <input type="submit">
</form>
```

```html+django
<form action="/custom-search-page/" method="get">
    {{ self.custom_search_form }}
    <input type="submit">
</form>
```
```html+django
<form action="http://email.example.com/subscribe/" method="post">
    {{ self.body_with_subscribe_fields }}
    <input type="submit">
</form>
```

didorothy:
I found myself needing this feature and so I have an implementation. Basically there is a FormFieldBlock that can be added to a StreamField. The page has to inherit from AbstractForm so that form processing happens and it must also include a mixin that pulls the FormFieldBlock "instances" out of the StreamField value to build the form. Submission then happens the same way as a normal AbstractForm. Though the way I have accomplished this may feel a bit hackish.

You can see my implementation here: https://github.com/didorothy/wagtail/tree/streamfieldforms

Some key files to look at would be:
* https://github.com/didorothy/wagtail/blob/streamfieldforms/wagtail/tests/testapp/models.py around line 344 where it shows how to set up the Page child class.
[...]

PR #2272, opened 2016-02-24
StreamField Based Forms for Form Builder

didorothy:
Proposed changes to enable forms in StreamFields. From discussion on #1564


gasman:
Hi @didorothy - thanks for the PR, and sorry it's taken so long to get round to reviewing this!

For me, one of the major motivations for implementing the form builder through StreamField would be the ability to choose field types through the StreamField block menu - i.e. rather than having a single 'form field' block where you add it and then select its type, there would be distinct block types for 'text field', 'dropdown field' and so on. That way, we'd eliminate a lot of clutter from the edit form, such as the 'choices' list on fields that don't need it. Do you think it would be possible to implement the field types in that way?

didorothy:
That can definitely be done. In fact it should just require some reworking of the inheritance tree for the blocks and the way block types are registered. I had kind of thought that that might be a better way but was not quite as straight forward first step. Unfortunately, my priorities have changed slightly so I may not get back to this for a month or two. As soon as I can find some time I will work up an adjustment toward this end.

mevinbabuc:
Hey @didorothy @gasman Is there any update on this feature ?

didorothy:
I have finally gotten back to working on this. Hopefully I can make a few changes each month till the details are worked out. The latest commit is a rough draft of where I'm headed. At this point I have a question about direction.

Should this new form functionality completely replace the old functionality? 

Right now the published version of wagtailforms requires an additional database table to keep track of form fields. When I originally started work on this the intent was to add a new option to how to build forms with wagtail (inside a StreamField). So I left in the original form code. Now that I am looking at this it would appear that it might be better to make wagtailforms only work one way where forms are embedded in StreamFields with allowed field types specified by the developer. My specific concern is about creating confusion for new developers trying to figure out the best way to allow their end user to create forms. Having one way can be more clear but it limits flexibility and affects backward compatibility.

Another alternative would be to break this out into a separate module and allow the developer to choose one or both.

Which direction is best?

gasman:
Hi @didorothy - sorry again for dropping the ball on this. One of the core team has recently been working on a client project involving some very similar work around StreamField-based form building - once that project is concluded, we'll be in a position to evaluate both implementations and hopefully borrow the best ideas from both...

stuartaccent:
We also required this and have a concept to our requirements. Though very early days or more hours. We have come up with the following. Documentation is currently under way but the highlights are below an maybe useful for ideas:

- Forms can be build in the cms admin and used site wide in any streamfield.
- We have included a mixin which will handle the form post if it is being submitted to a wagtail page.
- Forms are catagorised by their class in the cms admin for easier navigation.
- Form submissions are also listed by their form which you can filter by date and are ordered by newest first.
- Recaptcha can be added to a form.
- You can also add site wide regex validators fo use in regex fields.

[https://github.com/AccentDesign/wagtail_streamforms](https://github.com/AccentDesign/wagtail_streamforms)

Stu

didorothy:
I've decided, based on speed of feedback and issues with this pull request caused by my own misunderstanding (of git), that I will be working on this in a separate repository as a separate project. This will allow me to move forward without needing feedback and to focus on the specific needs that I have for the tool. 

Also, since there appear to be one or more competing options this might help people to better settle on which option best meets their needs.

You can find my current progress here: https://github.com/didorothy/wagtailstreamfieldforms

PR #2274, opened 2016-02-24
Add formal support for customising the form rendering of StructBlocks

gasman:
As per discussion at https://groups.google.com/d/msg/wagtail/n6SpjLD4qqI/cKTvAGSnFAAJ.

The `form_template` attribute was mentioned in passing in the docs, but was missing various things
to make it fully useful:

- context passed to form_template now includes 'prefix' and 'block_definition'
- context for the form is now populated in a separate overrideable `get_form_context` method
- full documentation and tests for form_template and get_form_context added

### DIFF
--- a/docs/topics/streamfield.rst
+++ b/docs/topics/streamfield.rst
@@ -309,23 +309,7 @@ This defines ``PersonBlock()`` as a block type that can be re-used as many times
         ('person', PersonBlock()),
     ])
 
-
-To customise the styling of the block as it appears in the page editor, your subclass can specify a ``form_classname`` attribute in ``Meta`` to override the default value of ``struct-block``:
-
-.. code-block:: python
-
-    class PersonBlock(blocks.StructBlock):
-        first_name = blocks.CharBlock(required=True)
-        surname = blocks.CharBlock(required=True)
-        photo = ImageChooserBlock()
-        biography = blocks.RichTextBlock()
-
-        class Meta:
-            icon = 'user'
-            form_classname = 'person-block struct-block'
-
-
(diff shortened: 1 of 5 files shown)


docs/topics/streamfield.rst
```diff
@@ -593,6 +577,65 @@ This is possible because the HTML rendering behaviour of these blocks does not i
...
+        surname = blocks.CharBlock(required=True)
+        photo = ImageChooserBlock()
+        biography = blocks.RichTextBlock()
+
+        class Meta:
+            icon = 'user'
+            form_classname = 'person-block struct-block'
+
+
+You can then provide custom CSS for this block, targeted at the specified classname, by using the ``insert_editor_css`` hook (see :doc:`Hooks </reference/hooks>`).
```
kaedroho: Would be better to link straight to the ``insert_editor_css`` hook from here.

docs/topics/streamfield.rst
```diff
@@ -593,6 +577,65 @@ This is possible because the HTML rendering behaviour of these blocks does not i
...
+
+        def get_form_context(self, value, prefix='', errors=None):
+            context = super(PersonBlock, self).get_form_context(value, prefix=prefix, errors=errors)
+            context['suggested_first_names'] = ['John', 'Paul', 'George', 'Ringo']
+            return context
+
+        class Meta:
+            icon = 'user'
+            form_template = 'myapp/block_forms/person.html'
+
```
kaedroho: An example template would be really useful to have here. I think most people like to copy + tweak instead of implement from scratch

kaedroho:
Looks good to me!

gasman:
Rebased and direct link to `insert_editor_css` added. As discussed earlier, I don't think we should be holding PRs back on the grounds of documentation not being in-depth enough - that can always be improved separately (and I'm not just saying that because it's my PR :-) )

gasman:
Merged in 9961455c6aa31eadb83b528a1037c5c53c9a40a1 (as it's now been reviewed, and there are other StreamField changes in the pipeline that are best applied on top of this)

PR #2275, opened 2016-02-24
Tooling for modern front-end components: React JS, ES6, and BEM CSS

JoshBarr:
* Adds the ability to create Wagtail UI components through React.js. 
* Adds frontend assets to a public NPM package that is published as part of the `setup.py` distribution workflow
* Deprecates jscs linting on existing JS, move to eslint for new components only (where style can be enforced from the outset).


wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
```diff
@@ -19,6 +19,10 @@
 
 @import 'wagtailadmin/scss/fonts';
 
+#wagtail {
```
davecranwell: This looks a bit vague to me, but I defer to your experience. Wouldn't `#wagtail-app-root` or similar, be more explicit and clear?

.eslintrc
```diff
@@ -0,0 +1,25 @@
+{
+    "extends": "airbnb",
+
+    "rules": {
+        "indent": [2, 2],
```
davecranwell: Can we not do 4-space throughout? It seems like a point of sensible consistency that all our code in all languages should be 4-space. Personally I find 2-space hard to read.

JoshBarr: Good question! Our motivation with this PR is to reduce barriers to other front-end devs from the open source community contributing to Wagtail. One reasonably consistent feature of the JS ecosystem is two space indentation. We felt that it would be more appropriate to implement a broader community standard (airbnb) rather than our personal preference (I'm also not fond of two spaces).

jonnyscholes:
This is awesome. Love the move to putting the frontend in it's own npm package and the move to React. Have you considered using something like Redux? I've found it really useful - particularly when there are lots of nested components. Which depending on the plan for wagtail may not be the case - ie are we going to Reactify full pages or just components within pages?

client/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+```javascript
+import { Explorer } from 'wagtail';
+
+...
+
+<Explorer onChoosePage={(page)=> { console.log(`You picked ${page}`); }} />
+
+```
+
```
jonnyscholes: Maybe a Features heading or similar where we list things like eslint, React, ES6, fetch etc to give devs a quick primer on what to expect and what editor plugins would help?

.eslintrc
davecranwell: If 2-space is likely to get more contribution, then I'll bite my tongue. However it will be a bit inconsistent to have most of Wagtail's core JS on 4-space, but the new UI components 2-space. We should fix up the legacy code too, ideally.

JoshBarr: That can be automated with tools like jscs that can re-indent the whole project, but i'd do that in a separate PR...

client/README.md
JoshBarr: @jonnyscholes that's not really a blocker for merging, feel free to write something if you want :)

.eslintrc
davecranwell: Cool.

kaedroho:
Merged in 14f02a0b50c4b250be5a8a7ce4559d39bab4fe86 Thanks!

PR #2276, opened 2016-02-24
Hook for adding CSS to all admin pages

tomdyson:
Like 'insert_editor_js' but not just for edit pages.


alexgleason:
Nice! This fixes alexgleason/wagtailfontawesome#1

m1kola:
We probably also should give ability to add JS to all admin pages (not only editor).

gasman:
Merged in 985a987436cff2fe0d8fac2621043a401889f1f6

PR #2278, opened 2016-02-25
Enforce slug uniqueness at the model level

### LINKED ISSUE #1518: Enforce slug uniqueness at the model level
I saw in your code here on github that this feature is not yet developed. In which version will this be fixed?

gasman:
Fixes #1518.

Added model-level validation on save, save_revision and preview (and fixed all tests that created invalid page instances), and moved slug uniqueness checking so that it happens both in WagtailAdminPageForm and on the Page model.


gasman:
Merged via #2280.

PR #2279, opened 2016-02-25
Enable the admin to search over any searchable fields

KaitCrawford:
Enables admin searches to return pages that contain the search query in any of their fields that have been made searchable (by adding the field to the models search_fields variable). Previously only pages that contained the search query in their title field were returned.

### DIFF
--- a/wagtail/wagtailadmin/tests/test_pages_views.py
+++ b/wagtail/wagtailadmin/tests/test_pages_views.py
@@ -20,6 +20,7 @@ from wagtail.tests.utils import WagtailTestUtils
 from wagtail.wagtailcore.models import GroupPagePermission, Page, PageRevision, Site
 from wagtail.wagtailcore.signals import page_published, page_unpublished
 from wagtail.wagtailusers.models import UserProfile
+from wagtail.wagtailsearch.index import SearchField
 
 
 def submittable_timestamp(timestamp):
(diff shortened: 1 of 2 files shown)


tomdyson:
Thanks @KaitCrawford! @ollywillans do you have any thoughts about this? @kaedroho thinks you had wanted title-only searches in the early days, but expanding it to all fields seems better for most cases.

davecranwell:
For any search to _not_ search thoroughly seems very unexpected imo. This change would make it much easier to update something global which is commonly requested by our clients e.g contact telephone number changes or a popular contact name leaves an organisation etc.

wagtail/wagtailadmin/tests/test_pages_views.py
```diff
@@ -1474,6 +1475,29 @@ def test_search(self):
...
+            live=True,
+            has_unpublished_changes=False,
+        ))
+
+        # Confirm the slug is not being searched
+        response = self.get({'q': "hello"})
+        self.assertNotContains(response, "There is one matching page")
+
+        # Add slug to the search_fields
+        Page.search_fields = Page.search_fields + (SearchField('slug', partial_match=True),)
```
gasman: I think this change will persist after the test has completed, which could cause unexpected behaviour in later tests. Saving the initial value of `search_fields` and restoring it at the end of the test should be enough to solve this, though.

kaedroho:
Merged in 8c8584922ed16eb326e13228fb56f4bbfcdc3ef6 Thanks!

PR #2282, opened 2016-02-25
Webpack fixes

justinvdm:
Some fixes are needed for the changes made in #2275:   - a few dependencies are missing from `package.json`   - webpack aliasing isn't working for the `config` module

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,8 +38,11 @@
   "dependencies": {
     "exports-loader": "^0.6.3",
     "imports-loader": "^0.6.5",
+    "react": "^0.14.7",
+    "react-dom": "^0.14.7",
     "react-redux": "^4.4.0",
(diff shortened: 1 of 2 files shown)


kaedroho:
Thanks @justinvdm!

SalahAdDin:
ok man, but, how can we use webṕack with wagtail resources?

kaedroho:
@SalahAdDin it's only used for creating the js/css bundles that get shipped in each release so this shouldn't be of much interest to site implementors. Webpack is part of Wagtail's gulp-based toolchain which you can find the usage docs for here: http://docs.wagtail.io/en/v1.3.1/contributing/developing.html#compiling-static-assets

SalahAdDin:
Oh, thank's, i saw and error with gulp watch, so, i'll report here.

PR #2283, opened 2016-02-26
Remove unused private method in BaseSearchQuery

jschneier:
Hey. I need to implement a tweak (more like subclass!) of the existing ``elasticsearch.SearchBackend`` and noticed that this method is unused while I was perusing the code base. Since it's private I think it's safe to remove it.


kaedroho:
Thanks @jschneier! Merged in 182c16c46a7c2908650b6d379bd66ee46a9b7d58

PR #2284, opened 2016-02-26
Sort all imports using isort

timheap:
A noisy diff, which is unfortunate, but it makes the code much nicer.


gasman:
Merged in 861bd2de14ab46c0d93287964b4b6e5d8ec8b9c1 + parent

PR #2285, opened 2016-02-27
Add distinct icons to blocks

benjaoming:
It's pretty hard to navigate the many different block types in a StreamField that permits lots of different blocks. So I used the icons from the current style guide. Some of the icon names and semantic intentions might be a little besides the way they're used, but in case the icon changes, I don't think it'll cause a major disaster :)

I moved a couple of subclass `class Meta` to the end of the class definition because it's easier to always know where to look for them.

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -93,7 +93,7 @@ Contributors
 * Rich Atkinson
 * jnns
 * Eugene MechanisM
-* benjaoming
+* Benjamin Bach
 * Alexander Bogushov
 * Aarni Koskela
 * alexpilot11
--- a/wagtail/tests/testapp/blocks.py
+++ b/wagtail/tests/testapp/blocks.py
@@ -22,4 +22,3 @@ class SectionBlock(blocks.StructBlock):
     class Meta:
         icon = "form"
         template = 'tests/blocks/section_block.html'
-
--- a/wagtail/tests/testapp/blocks.py
+++ b/wagtail/tests/testapp/blocks.py
@@ -11,6 +11,7 @@ class LinkBlock(blocks.StructBlock):
         return context
 
     class Meta:
+        icon = "site"
         template = 'tests/blocks/link_block.html'
 
 
@@ -19,4 +20,6 @@ class SectionBlock(blocks.StructBlock):
     body = blocks.RichTextBlock()
 
     class Meta:
+        icon = "form"
         template = 'tests/blocks/section_block.html'
+
--- a/wagtail/wagtailcore/blocks/field_block.py
+++ b/wagtail/wagtailcore/blocks/field_block.py
@@ -19,8 +19,6 @@ from .base import Block
 
 class FieldBlock(Block):
     """A block that wraps a Django form field"""
-    class Meta:
-        default = None
 
     def id_for_label(self, prefix):
         return self.field.widget.id_for_label(prefix)
@@ -76,8 +74,13 @@ class FieldBlock(Block):
         # the one this block works with natively
         return self.value_from_form(self.field.clean(self.value_for_form(value)))
 
+    class Meta:
+        icon = "edit"
+        default = None
+
 
 class CharBlock(FieldBlock):
+
     def __init__(self, required=True, help_text=None, max_length=None, min_length=None, **kwargs):
         # CharField's 'label' and 'initial' parameters are not exposed, as Block handles that functionality natively
         # (via 'label' and 'default')
@@ -94,6 +97,7 @@ class CharBlock(FieldBlock):
 
 
 class TextBlock(FieldBlock):
+
     def __init__(self, required=True, help_text=None, rows=1, max_length=None, min_length=None, **kwargs):
         self.field_options = {
             'required': required,
@@ -114,8 +118,12 @@ class TextBlock(FieldBlock):
     def get_searchable_content(self, value):
         return [force_text(value)]
 
+    class Meta:
+        icon = "pilcrow"
+
 
 class URLBlock(FieldBlock):
+
     def __init__(self, required=True, help_text=None, max_length=None, min_length=None, **kwargs):
         self.field = forms.URLField(
             required=required,
@@ -125,8 +133,12 @@ class URLBlock(FieldBlock):
         )
         super(URLBlock, self).__init__(**kwargs)
 
+    class Meta:
+        icon = "site"
+
 
 class BooleanBlock(FieldBlock):
+
     def __init__(self, required=True, help_text=None, **kwargs):
         # NOTE: As with forms.BooleanField, the default of required=True means that the checkbox
         # must be ticked to pass validation (i.e. it's equivalent to an "I agree to the terms and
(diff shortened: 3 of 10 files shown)


gasman:
Thanks @benjaoming! Setting default icons for `URLBlock`, `DateBlock` and so on is a good idea - I suspect they just got missed because I only imagined them being used as part of a `StructBlock`.

For the more 'generic' blocks, which will usually be subclassed / customised when used on a site - namely, `FieldBlock`, `ChoiceBlock`, `ChooserBlock`, `ListBlock`, `StreamBlock`, `StructBlock` and `SnippetChooserBlock` - I actually think it's better for us *not* to set a default icon. We don't have the knowledge or context to choose an icon which will be truly meaningful to content editors - at best, we can only choose an icon that's semi-appropriate from a programmer's point of view. For example, if a site is using a `ChoiceBlock` to choose from a list of people, it's clearly far better to have an icon that represents the concept of 'Person' rather than the action of 'choosing'. Only the site implementer is able to make that decision.

By leaving the 'placeholder' icon in place, we clearly communicate to the site implementer that they need to choose an icon. If we supply a default icon that's not great, but not obviously terrible (from a developer's perspective), then the developer might not even realise that they can (and should) choose a better one, and the content editor's experience will suffer.

benjaoming:
Thanks for the feedback!

>  By leaving the 'placeholder' icon in place, we clearly communicate to the site implementer that they need to choose an icon.

You can say the same about the icon being unsuitable for the purpose ;) But I think the icons chosen are still reasonably general purpose.

My main argument is not so much about the icon communicating the intention, but about visual navigation where I found myself really bad at telling which was which, so I ended up often reading a lot of text before spotting which was the right button to click. Any kind of visual metaphor (good and bad) can help in that situation.

But the thing is also to ship something with nice defaults.

Understanding your argument and not saying that there's a clear right and wrong, I would slightly prefer having default icons that are distinguishable. The site implementor may want to tweak them.. but may also find it unnecessary and thus save a bit of work!

You could choose to merge this and then flag your UI people for adding more icons to the set in case some of the defaults don't add up to your quality standards?

alexgleason:
Looking at the code, the only one I can clearly tell is ListBlock. It sets a UL icon.

This is the sort of thing you'd have in a text editor. When you're making a bulleted list. That's not what a listblock is. A listblock is a series of elements that probably don't have bullets next to them. They shouldn't be compared to a bulleted list in a text editor.

I'm not against having defaults for structural block types. Those are ListBlock, StructBlock, and StreamBlock. But we don't have any icons that properly convey the message of what they are. Our iconset in Wagtail is fairly limited. We have text editor icons, but those don't represent structural block types. You shouldn't confuse someone into clicking a bulleted list icon and expecting a bulleted list, when it's actually a stream of arbitrary items.

I can see both sides of the argument. But I think in order to set defaults for structural block types, we will have to design new icons for them. Ones that actually represent how they work instead of comparing them to vaguely similar text editor functions.

For now I would rather merge without icons on structural block types. And add icons for structural block types if anyone creates good ones.

benjaoming:
Okidoki fellow devs, I'll remove the icons for the Struct types! Just a sec...

benjaoming:
Added a comment to each of the 'placeholder' choices, just to indicate the goal of having different default icons. Also moved another occurrence of `class Meta:` to the end of the class definition.

gasman:
I can't agree with the comment, sorry. It's not a matter of "we don't have a good enough generic icon *yet*" - I believe that *any* generic icon here will harm the user experience for content editors. The icon should represent the thing the content editor is working with, not the framework it's built on... that would be like looking for the toilets at a restaurant, and finding that the signs have a 'brick' icon on them, because the toilet building is made of bricks. ;-)

Choosing a good icon is the responsibility of the site implementer - defaulting to the 'placeholder' box is a good thing because it prompts the site implementer to spend the extra few minutes of effort to select a truly appropriate icon.

alexgleason:
I had envisioned something like this icon, turned to its side for a ListBlock. http://fontawesome.io/icon/columns/

But I see what Matthew is saying. The fact that it's a list isn't necessarily the important thing about a ListBlock. For instance if you have a list of phone numbers, do you want the icon above, or do you want the icon of a phone? I think it does probably make sense to not have defaults on structural block types.

We do appreciate your contribution @benjaoming ! I agree that have defaults on non-structural types will be helpful and desired. Thank you for submitting.

benjaoming:
> But I see what Matthew is saying. The fact that it's a list isn't necessarily the important thing about a ListBlock.

I already reverted those, so maybe focus on the block types that have been changed?

benjaoming:
> I believe that any generic icon here will harm the user experience for content editors.

So what about `ImageChooserBlock` now having an image as the icon?

gasman:
> So what about `ImageChooserBlock` now having an image as the icon?

I'm very much in favour of that, yep. (I'm very surprised it wasn't already there, in fact...) The 'generic icon' problem doesn't arise there, because `ImageChooserBlock` will typically be used directly, with no further 'specialisation' - it still means "choose an image" whether you're looking from the perspective of a developer, or a content editor.

benjaoming:
Okay cool, I was just getting confused with the further elaboration after I made the changes in  beadbec :)

benjaoming:
And sorry about being so un-elaborated, I think there's a convention that it's rude to reply in one-liners when you receive significantly longer comments, but I just agree now and think everything's good :) So let me know if you think I've addressed the icons and if not, I can add/change it more and do a final squash and push -f.

gasman:
No worries - some simple questions have long answers and vice versa, so feel free to write as much or as little as you need :-)

A few more changes I'd like to see here:

* The "No suitable icon as of now" comment text is misleading - as mentioned, I wouldn't encourage anyone to try to create a better generic icon, because I think that generic icons are a fundamentally bad idea for those blocks. I don't think we particularly need a comment here, but if you want to keep it, I would suggest some wording like: "No icon specified here, because that depends on the purpose that the block is being used for. Feel encouraged to specify an icon in your descendant block type"
* I think `FieldBlock`, `ChooserBlock` and `SnippetChooserBlock` should be changed back to 'placeholder' too, for the same reason.

benjaoming:
Sounds good to me with the comment, that's implemented!

> FieldBlock

Done

> ChooserBlock

Done

> SnippetChooserBlock

But there is a snippet icon?

gasman:
`SnippetChooserBlock` is a bit of a borderline case, so on further consideration I'm happy to leave the icon in there. (My reasoning was: when you use SnippetChooserBlock, you always need to specify a snippet type, so the content editor is never simply being asked to "choose a snippet"; it will always be something more specific, such as an advert, or a country, and so the site implementer should really choose a more specific icon to represent it. But actually, the snippet icon is a pretty good choice here - it's recognisable and meaningful to content editors.)

I see `ListBlock` and `StreamBlock` are still using the old comment text - please can you update these? After that, I think this will be ready to merge.

gasman:
Merged in dfd8a5cc2b9a1badbeb9621e58a789aef8471d72

PR #2293, opened 2016-03-01
Ask user for old password when changing password

math1985:
This is accomplished by using SetPasswordForm instead of PasswordChangeForm.

This adds extra security, as without this commit, an attacker that has access to
a user's session at one point in time will be able to change the user's password
and gain permanent access.

### DIFF
--- a/wagtail/wagtailadmin/tests/test_account_management.py
+++ b/wagtail/wagtailadmin/tests/test_account_management.py
@@ -190,6 +190,7 @@ class TestAccountSection(TestCase, WagtailTestUtils):
         """
         # Post new password to change password page
         post_data = {
+            'old_password': 'password',
             'new_password1': 'newpassword',
             'new_password2': 'newpassword',
         }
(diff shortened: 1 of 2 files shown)


math1985:
I see the drone test fails, is there something I need to do on my side to rectify this? I'm not quite sure about the cause of the failure.

gasman:
@math1985 That was a code style error that existed in master (fixed now) at the time you submitted the PR - if you rebase this PR on top of current master (or, if you're not familiar with rebasing, merge current master into it) then the drone build should hopefully pass.

math1985:
Thanks, rebased, the error is gone now.

timheap:
This looks good to me :+1:

timheap:
This has been merged in 03cba5b, and you have been added to the [CONTRIBUTORS.rst](https://github.com/torchbox/wagtail/commit/03cba5b91c3feb0a57e6cd39e6504397e51321cc#diff-752bcc0e38545b90194714dd446d9292) list. Thanks!

math1985:
Thanks for reviewing! The PR didn't close automatically, so I'll close it manually.

Issue #2096, opened 2016-01-10
Trailing slash


davejacobs:
How do I stop wagtail from auto-appending a trailing slash when I navigate to wagtail-controlled pages? I've set `APPEND_SLASH` to `False`, removed Wagtail redirect middleware, and even tried adding `django-unslashed` (and `REMOVE_SLASH = True`) per its docs, but redirects keep happening. Any advice?

kaedroho:
This isn't something we've looked into much yet. But I'd be happy to see Wagtail support ``APPEND_SLASH=False`` if it's not too tricky to do.

But it might be able to do it now by copying and modifying this URLconf entry (https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailcore/urls.py#L8-L13) making the trailing slashes optional. Then implement your own version of the ``{% pageurl %}`` templatetag to remove the trailing slash.

davejacobs:
Thanks for the pointers, I'll see what I can come up with.

jangondol:
One option would be to have something like `WAGTAIL_APPEND_SLASH` or we could go for something more generic like `WAGTAIL_URL_APPEND` which could be set to "/" (slash) or "" (nothing, i.e. no trailing slash) or something like ".html" (give the appearance that the wagtail page is a static page -- could be useful). Any thoughts?

davejacobs:
Either of those would work for my purposes. I don't need the `.html` functionality, but I could understand how that would be useful for some people. I've tried modifying `urlpatterns`, per @kaedroho's suggestion, and it works for me. However, I haven't been able to come up with a super-efficient regex to replace the original with, and I also have to manually link to `/admin/` (with a slash) to make redirection working.

Official support would be great, and either of the options you propose would work well for me. Thanks for asking!

jangondol:
Yes, @davejacobs, agree that official support would be very nice. That `.html` functionality could be useful in a number of ways, e.g. for making migration from an older CMS  easier (keeping existing URLs is sometimes desirable). Some sites use `.php`, some have `.htm`, `.html`, etc.. Being able to append slash, no slash, or anything at all in an easy and documented way would be super helpful.

gasman:
I'd be happy to see the no-slash format supported. The slash-appended version makes the logic slightly simpler, as we don't need a special case for the root path `/` (which, obviously, always ends in a slash...) but it's no big deal if we need to add a few more edge-case checks.

[...]

PR #2298, opened 2016-03-02
Add WAGTAIL_APPEND_SLASH setting

atbaker:
Hi guys,

This PR addresses #2096. It's incomplete — it needs tests and docs changes. But since adding a new setting isn't a trivial change, I wanted to start this PR so we could agree on the best way to do this.

We're getting ready to launch a new Wagtail-backed service at work, serving only a few pages of content at first but gradually transitioning off our old platform. The old platform doesn't append slashes, so to keep consistency across the site during the transition we'll be stripping them from our new Wagtail URLs too. Hence, our need for a solution to #2096.

**A summary of my first stab:**

- Set a new constant in wagtailcore.utils, `WAGTAIL_APPEND_SLASH`, which defaults to `True` if it's not present in the Django settings
- Reference that constant in wagtailcore.urls, using different regex's for the `wagtail_serve` URL
    - The new regex simply makes trailing slashes optional. That seemed like the best behavior to mimic Django's `APPEND_SLASH` setting, which I would guess is what most other devs would expect
- Django's `reverse` always returns a trailing slash, so to make links appear without trailing slahes I had to add some extra logic to the Page model's `get_url_parts` method

Happy to tweak this whichever way you think is best. Once we agree on an approach, I'll fill in the tests and add some notes to the docs.

Wagtail has been a delight to work with these past couple months. Thank you for creating, sharing, and maintaining such a fantastic library.


m1kola:
@atbaker, thank you for your participation!

At first glance it looks good, but I have doubts about how `WAGTAIL_APPEND_SLASH` should work with Django's `APPEND_SLASH` (I mean in various combinations).

wagtail/wagtailcore/urls.py
```diff
@@ -1,5 +1,15 @@
...
 from wagtail.wagtailcore import views
+from wagtail.wagtailcore.utils import WAGTAIL_APPEND_SLASH
+
+
+# If WAGTAIL_APPEND_SLASH is False, allow Wagtail to serve pages on URLs
+# with and without trailing slahes
+if WAGTAIL_APPEND_SLASH:
+    serve_pattern = r'^((?:[\w\-]+/)*)$'
+else:
+    serve_pattern = r'^((?:[\w\-]+/?)*)$'
```
m1kola: Looks like this pattern will be very slow.

For example you can try this:
```python
import re
re.match(r'^((?:[\w\-]+/?)*)$', 'media/images/James_Joyce_in_1915.width-300_6M5Jvbh.jpg')
```

atbaker:
@m1kola - Thanks for the first review!

You were right about that regex - it had a catastrophic backtrack. I updated it to be simpler and more efficient.

Regarding the `WAGTAIL_APPEND_SLASH` setting's interaction with Django's `APPEND_SLASH`, here's how it works:

<br/> | WAGTAIL_APPEND_SLASH = True (default) | WAGTAIL_APPEND_SLASH = False
------------ | ------------- | -----------
**APPEND_SLASH = True (default)** | Original Wagtail behavior: requests to URLs without trailing slashes are caught by `CommonMiddleware`, which 301 redirects to a version with a trailing slash which Wagtail serves | Wagtail pages can be reached with and without trailing slashes. No redirects occur.
**APPEND_SLASH = False** | Requests that include a trailing slash are served. Requests without it are not redirected by `CommonMiddleware` and are 404'd | Wagtail pages can be reached with and without trailing slashes. No redirects occur.

If you think this is looking good, I'll update the docs and do one last round of polish on the PR.

gasman:
Thanks! Merged in 228088d33d3d36afdc8363f9e59f1a8cbcf6d072.

atbaker:
Thanks for shepherding this along! Hoping this is the first of many PRs to the project.

PR #2301, opened 2016-03-02
Image transparency missing from one place and omitted from styleguide

davecranwell:


davecranwell:
Merged in https://github.com/torchbox/wagtail/commit/2d16bf9606f2ad53f5269586e0aa296fdb3d1190

m1kola:
@davecranwell, we probably need to set 1.4 milestone because this PR is already merged to master.

PR #2302, opened 2016-03-02
Add docs on committing code

timheap:
Inspiration and flow from Django's `internals/contributing/committing-code` docs

### DIFF
--- a/docs/contributing/committing.rst
+++ b/docs/contributing/committing.rst
@@ -1,0 +1,121 @@
+===============
+Committing code
+===============
+
+This section is for the committers of Wagtail,
+or for anyone interested in the process of getting code committed to Wagtail.
+
+Code should only be committed after it has been reviewed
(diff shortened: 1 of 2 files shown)


davecranwell:
Brilliant! Will try following this later today.

davecranwell:
These instructions work a treat, all up to the last steps where `git branch -d pr/xxxx` returns:
```
error: The branch 'pr/2206' is not fully merged.
If you are sure you want to delete it, run 'git branch -D pr/2206'.
```

For completeness, I think there should also be a note about closing PRs manually, once merged, and referring to the commit where merge occurred. I understand this isn't necessary when the PR uses the "Fixed: #[whatever]" syntax, but many don't.

timheap:
I'll amend the documentation to mention that, good catch.

m1kola:
Thanks, @timheap! It's really helpful!

PR #2304, opened 2016-03-03
Add docs on how to build docs

timheap:
After a question on the mailing list about how to build the docs, I thought it was worth documenting for future reference.

This is how I have been building the docs. Hopefully it is how the docs are supposed to be built.

### DIFF
--- a/docs/contributing/developing.rst
+++ b/docs/contributing/developing.rst
@@ -110,3 +110,42 @@ This must be done after every change to the source files. To watch the source fi
 .. code-block:: sh
 
     npm start
+
+Compiling the documentation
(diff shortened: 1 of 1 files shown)


m1kola:
Looks Thanks, @timheap!

P.S.: I like the commit message :+1:

gasman:
I used your docs to build the docs about building docs, and now I can read the docs about building docs...

Merged in 84432fbfd6e27792a27b99a563b60f10e960d271.

tomdyson:
😷😷😷
(actually 'Face With Medical Mask', intended for discussions about bird flu, but the closest I could find to 'docs'.)

Issue #2307, opened 2016-03-04
Registered settings do not activate when selected


mgd020:
The active class is applied when the request.path begins with the `MenuItem.url`.
The MenuItem.url for a setting is `{app_label}/{model_name}/`
The request.path for a setting is `{pk}/{app_label}/{model_name}/`
The fix is to move `pk` to the end of the url like this `{app_label}/{model_name}/{pk}`
I have a PR for this.

gasman:
Fixed in #2308

PR #2308, opened 2016-03-04
Fix for #2307: Registered settings do not activate when selected

mgd020:
Issue #2307
Existing unit tests exercise code, and none fail.


gasman:
Thanks! Merged in cc509f226ad24adb5dd233bed6d9402237a0dbd8 / c289028cd95d6e72d6580a1afb148a52acc11d2c.

(Have also added a basic test in 32d6d43692f8a947542e2b0577dfd7e4eb5228d2 - even if code is covered by the existing tests, we always ought to have a test for the specific thing being fixed, so that we'll know if we inadvertently break it again in future.)

PR #2312, opened 2016-03-06
Fix broken links in documentation

loicteixeira:
While going through the [Getting Started tutorial](http://docs.wagtail.io/en/v1.3.1/getting_started/tutorial.html), I stumbled upon a few broken links.

I used [Dead Link Checker](http://www.deadlinkchecker.com/) to catch as many as possible. However, there will still be some broken links:
 - Documentation of older versions will not be affected.
 - The `Read the Docs` widget (on the lower left corner) links to the same page for previous version of the documentation, even if  it does not exist for that said version.

I was able to successfully generate the `html` and `epub` documentation to check everything was alright. However, I could not verify the `pdf` version as I was unable to install `pdflatex` properly.


gasman:
Merged in 10791a9c0bafc919d49c870f82ba985e61da93c7 + parents - thanks!

SalahAdDin:
Ok, for the search reference in advance topics the search.html link doesn't exist, doesn't show anything.

PR #2316, opened 2016-03-07
Dependency update: Support django-treebeard 3.0 and 4.0 releases

mvantellingen:
The admin extensions in django treebeard 3.0 are not compatible with
Django 1.9. Although these are not used by Wagtail they are used in
other projects (e.g. Oscar), so we should allow treebeard 4.0 which
is compatible with Django 1.9.


gasman:
Thanks! Merged in e7ace5c1818d10ed9c7916113e08c2a291e4b82b.

PR #2318, opened 2016-03-07
Eliminate duplication of dependencies in tox.ini

### LINKED ISSUE #2317: Eliminate duplication of dependencies in tox.ini
tox.ini has its own copy of Wagtail's dependency list (including optional test-related packages), which is liable to get out of sync with the canonical list in setup.py. In the worst case, this can mean that our CI setup is running outdated versions of our dependencies, and fails to catch errors in the newer versions that people are using in the real world.

It ought to be possible to get tox to run `pip -e .[tests]` or equivalent as part of the setup, but I've failed to do so here: https://github.com/gasman/wagtail/tree/tests/remove-tox-deps . (It seems to do the right thing locally, but Travis claims that the packages haven't been installed and fails with a missing Django module.) http://stackoverflow.com/questions/29870629/pip-install-test-dependencies-for-tox-from-setup-py claims that it's impossible without moving dependencies back into a requirements.txt file (which would be incredibly annoying), but I don't entirely believe them...

gasman:
Finally found the magic incantation, I hope :-)

Fixes #2317

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -10,7 +10,8 @@ exclude = wagtail/project_template/*
 max-line-length = 120
 
 [testenv]
-commands=coverage run runtests.py --elasticsearch
+install_command = pip install -e ".[testing]" -U {opts} {packages}
+commands = coverage run runtests.py --elasticsearch
 
 basepython =
     py27: python2.7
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -19,23 +20,9 @@ basepython =
     py35: python3.5
 
 deps =
-    django-modelcluster>=1.1b1
-    django-taggit>=0.17.5
-    django-treebeard>=3.0,<5.0
     django-sendfile==0.3.6
```
timheap: Would moving `django-sendfile`, `Embedly`, and `jinja2` deps in to the `[testing]` requirements work? Then all the requirements would live in the `setup.py` `elasticsearch` is an optional dependency that is required for testing that already lives in `setup.py` `[testing]`, so it seems that it might work.

timheap:
I like this idea, nice work!

tox.ini
gasman: I'd prefer to keep a distinction between "packages required in order to run the tests" and "packages required for all tests to pass without skipping". It's possible that the latter set might have some onerous installation requirements - even if that's not the case at the moment - and I wouldn't want to make life harder for developers getting up and running. (I'm not sure why elasticsearch is in the former list - maybe that was an oversight, or maybe there's something odd about the way the tests import it...)

I like the idea of having everything in setup.py though - perhaps we could have a separate 'testing_optional' list there?

kaedroho: > I'm not sure why elasticsearch is in the former list - maybe that was an oversight

Nope, I put it there. Most of the ES tests don't require elasticsearch. This is why we have the ``--elasticsearch`` flag

kaedroho:
Merged in 094ab6dfb9dccaef1b981aab4c98513b0b4176ae thanks!

PR #2333, opened 2016-03-10
Hook for adding JS to all admin pages

tomdyson:
Like insert_editor_js but not just for edit pages.


gasman:
Merged in 83ab0746bcbbc6ef19375e345503fd25ebc9218f + parents.

Issue #2327, opened 2016-03-09
User bar list spacing can be hijacked 


tomdyson:
There's extra spacing between the list items in my user bar / edit bird - see attached screenshot. My site uses http://getskeleton.com/ which includes the line:

```css
li {
  margin-bottom: 1rem; }
```

Can the user bar's CSS be scoped to avoid clashes like these?

<img width="731" alt="edit-bird-screenshot" src="https://cloud.githubusercontent.com/assets/15543/13642274/bf4af6fa-e613-11e5-9c1b-799d80ca9a3e.png">

davecranwell:
It appears to have also inherited the font.

SalahAdDin:
:+1:

JoshBarr:
My cunning plan is to not use list items at all. We'll go for generic spans or divs, and add more traps into the CSS. On modern browsers we can reset `all: initial` before applying any styling, which should help.

gasman:
Hopefully fixed in #2341. @tomdyson, I've put up a custom build at http://playground.torchboxapps.com/matthew/wagtail/wagtail-1.4rc1-userbarfix.tar.gz with @JoshBarr's fix - please can you try this version (use this URL in place of `wagtail==1.4rc1` in your requirements file), and confirm whether this has fixed the styling?

tomdyson:
WFM! Thanks, @JoshBarr.

<img width="680" alt="edit-bird-fixed" src="https://cloud.githubusercontent.com/assets/15543/13700241/ad454a22-e777-11e5-8a53-2cc2f79bbaf5.png">

alexgleason:
Awesome! Also, I just reread my post, sorry if I sounded condescending. It was meant to be a general statement and not a jab at whoever programmed the new userbar.

PR #2341, opened 2016-03-11
Fix for #2327: More defensive CSS for the userbar

JoshBarr:
Removed ul and li from the userbar HTML. Now it only contains DIVs, it shouldn't inherit so much from the surrounding page context, unless you write really bad CSS :+1:


davecranwell:
A slight aside, regarding your use of BEM, but shouldn't all the sub-elements of the userbar be delimited by double underscores? e.g `.wagtail-userbar__items`? I'm a little new to BEM so I'm not sure which style you're following, what is acceptable, and what isn't. I note that `&__item` (singular) gets the double-underscores.

gasman:
Thanks! Merged in aa609c4eb5d3f3afa5b148df376776caa8c43f3b (master), ecacb60a4b7bd835fadab3d9bc3d90ed3428ef5e (stable/1.4.x).

PR #2346, opened 2016-03-11
Implement default col class on FieldRowPanels if it isn't explicitly included

### LINKED ISSUE #2257: Remove col* classes from FieldRowPanel
``FieldRowPanel`` currently requires the developer to set ``col*`` classes on all of the child panels. As described in: http://docs.wagtail.io/en/v1.3.1/reference/pages/panels.html#fieldrowpanel

I think it would be a neater if these classes were added automatically (based on the total number of fields specified), so a developer only has to specify the fields.

chrxr:
Fixes #2257 

Regex checks if FieldRowPanel children have col# in their classname attribute.

If they don't, or if the col# is above 12, it applies a default col class to the panel, calculated as num panels / 12

Haven't implemented tests yet. I notice that there are no FieldRowPanel tests in tests-edit-handlers.py. Should I add a test for the rendering of the FieldRowPanel parent container as well as for the col_count?


wagtail/wagtailadmin/edit_handlers.py
```diff
@@ -340,6 +342,14 @@ def __init__(self, children, classname=""):
         self.classname = classname
 
     def bind_to_model(self, model):
+        col_count = " col" + str(int(math.floor(12 / len(self.children))))
+
+        # If child panel doesn't have a col# class then append default based on
+        # number of columns
+        for child in self.children:
+            if re.match("^.*?(col)[1-9]{1}(?![3-9])[0-2]{0,1}.*?$", child.classname) is None:
```
gasman: I think this could be simplified to: `if not re.search(r'\bcol\d+\b'):`

(this is slightly broader in that it'll match any numeric `col#` classname rather than just 1-12, but I think that's probably a good thing)

gasman:
Thanks @chrxr (and sorry for the delay in reviewing)! Yep, I think this could do with a couple of tests - one test to confirm that the correct classname is added where relevant, and another to confirm that children with existing `col#` classnames are left alone.

chrxr:
Hi @gasman, I added some more basic tests as well for FieldRowPanel, as there didn't appear to be any. I think its all relevant, but let me know.

gasman:
Merged in bdaf696cd264dcdbef5a67c2feeb676788ea8027 + parents - thanks!

PR #2350, opened 2016-03-12
Spam detection for FormBuilder submissions

timorieber:
At least public form pages need a mechanism to protect form submissions against spam.

We've recently relaunched an existing low-traffic site (< 50 visits/week, above-average regional google ranking) using wagtail. There is also a form page to get in touch, and within the first week they received 12 spam messages.

Until now this pull request implements spam detection based on a bayesian spam filter. Any feedback and guidance is highly appreciated.

Some details implemented yet:
- Form submissions are marked with a spam flag
- Every submission is evaluated
- Submissions can be marked as spam/not spam afterwards
- A custom spam detection model may be used (and will be trained with every admin decision or very high/low scoring submissions)

Todo and further ideas:
- Refactor utility functions into class (e.g. `Detector`)
- Configure custom model (storage location, empty or default initialisation)
- Re-train custom model based on all existing form submissions (batch)
- Add tests and docs 
- Configure spam detection for every form page separately (no approach at hand)

Another idea was to implement form honeypots, but I couldn't find a suitable library.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,6 +31,7 @@ install_requires = [
     "html5lib>=0.999,<1",
     "Unidecode>=0.04.14",
     "Willow>=0.3b4,<0.4",
+    "antispam>=0.0.10,<0.1.0",
 ]
 
 # Testing dependencies
--- a/tox.ini
+++ b/tox.ini
@@ -10,7 +10,8 @@ exclude = wagtail/project_template/*
 max-line-length = 120
 
 [testenv]
-commands=coverage run runtests.py --elasticsearch
+install_command = pip install -e ".[testing]" -U {opts} {packages}
+commands = coverage run runtests.py --elasticsearch
 
 basepython =
     py27: python2.7
@@ -19,23 +20,9 @@ basepython =
     py35: python3.5
 
 deps =
-    django-modelcluster>=1.1b1
-    django-taggit>=0.17.5
-    django-treebeard>=3.0,<5.0
     django-sendfile==0.3.6
-    djangorestframework==3.1.3
-    Pillow>=2.3.0
-    beautifulsoup4>=4.3.2
-    html5lib>=0.999,<1
-    Unidecode>=0.04.14
-    elasticsearch==1.1.0
-    mock==1.0.1
-    python-dateutil==2.2
-    pytz==2014.7
     Embedly
-    Willow>=0.3b4,<0.4
     jinja2==2.8
-    coverage
 
     dj18: Django>=1.8.1,<1.9
     dj19: Django>=1.9,<1.10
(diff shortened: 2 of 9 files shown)


m1kola:
Thanks, @timorieber!

It's looks interesting, but I'm not sure about this feature inside of Wagtail. Also I think that new dependency on `antispam` is not a good idea.

Wagtail probably should give ability to filter messages somehow... @gasman, what do you think about this?

timorieber:
Thanks for the quick feedback, @m1kola. I was uncertain about your topics, too, that is why I started with this prototypical approach.

As it stands the `FormBuilder` is a nice tool to provide simple forms in wagtail pages. What I was missing is a standard way to customize the form processing when it was submitted. I haven't been digging very deep, but I saw no easy way to subclass or hook into relevant code. This is definitely a thing that wagtail should provide, as custom forms without custom form handling are very limited.

For me the spam detection is only the first topic that came up using `FormBuilder`, I assume there is more to come (e.g. custom form validation).

Do you have any ideas how to customize form submissions in a way that won't conflict with the `wagtailforms` codebase and keep the admin backend (defining forms, list submissions) working?

zerolab:
FWIW, I think that if we are to rely on additional dependencies, then making it a contrib module is better.

As you mention, form honeypots are another way to go. 
In the Drupal world, the [Honeypot](https://www.drupal.org/project/honeypot) module is rather popular. The idea is that it adds a distinct, visually hidden text field to the form to act as the honeypot and marked as such. Plus, it adds an optional time restriction as hashed hidden field which helps with automated bots and tools which fill out the form way faster than a human would do. 

The approach above does not need another dependency, as far as I can think of it. It may need a configuration UI where one can set the honeypot field name and the time restriction value..

timorieber:
#2617 addresses the form submission customization requirements mentioned in https://github.com/torchbox/wagtail/pull/2350#issuecomment-195795982.

The essential requirement addressed by this PR (spam detection, possibly configurable) remains open.

jacobtoppm:
@timorieber Thanks for contributing to Wagtail! This looks like a great feature, but we believe the additional dependency on antispam, as well as the modelling logic, makes it better suited to a third party app or even a blog post/tutorial. However, we'd like to incorporate some of the models changes you suggest to make this sort of approach easier: adding an `is_spam`/`flagged` field and enabling manual flagging and filtering in the Wagtail Admin in core, and allowing third party apps to hook in and implement the kind of automatic filtering in this PR. 

As a result, we're closing this PR and opening an issue for these related changes, here: https://github.com/wagtail/wagtail/issues/6019. If you're still interested, we'd definitely love to see you contribute to this new feature, and see any third party apps/posts that come out of the approach in this PR.

Issue #2347, opened 2016-03-11
Wagtail must use "pk" instead "id" for users models


Nekmo:
I'm using a custom user model with a primary key called "Identified" instead "id". Wagtail obliged to use "id", which is incorrect. If "pk" is used, it always works and will be more correct.

Examples of the error:

> NoReverseMatch at /cms/users/
Reverse for 'edit' with arguments '('',)' and keyword arguments '{}' not found. 1 pattern(s) tried: [u'cms/users/([^\\/]+)/$']

Source:
> **/wagtail/wagtailusers/templates/wagtailusers/users/list.html
      {% url 'wagtailusers_users:edit' user.id %}
      ">{{ user.get_full_name|default:user.get_username }}</a>

Other error:
> FieldError at /cms/users/default|gis/
Cannot resolve keyword 'id' into field. Choices are: auth_token, client, contact, date_joined, document, editeds_by, email, first_name, full_name, groups, historicalavailability, identified, image, is_active, is_staff, is_superuser, last_login, last_modified, last_name, logentry, metadatas, note, owned_pages, pagerevision, password, processedavailability, salutation, savefilter_created_by, savefilters, selectmark, site_group, user_permissions, username, userprofile

Source:
> ***/wagtail/wagtailusers/views/users.py in edit
@permission_required(change_user_perm)
def edit(request, user_id):
                user = get_object_or_404(User, id=user_id) ...

m1kola:
I created PR #2352.

Also I think that would be better to use "pk" for all models inside of Wagtail. I found more than 820 usages of ".id" in *.py files (most of them in tests) and more than 160 in *.html. It's probably not so important for Wagtail users, but gives more flexibility for Wagtail's developers.

gasman:
Fixed in #2352.

Standardising on `pk` in our code is probably a good habit to get into, but it's less important for Wagtail's own models (where we know the primary key will always be `id`) so I think this fix is sufficient for now.

PR #2352, opened 2016-03-13
Support of custom user model

m1kola:
Ref #2347

Some usages of "id" were replaced with "pk", but probably it's not enough to support custom user model.

I checked following:
- [x] Files with get_user_model
- [x] Files with request.user.id
- [x] Files with user.id

Any suggestions?


gasman:
Merged in 1172349ccab5ca9edbe0ffd7021844436d4f4486 - I also updated the `CustomUser` test model to use a non-standard primary key (4f50e576aa3112200710162bed59484ffa8cc5c4) and fixed the resulting failures in 3b09ffaee59ea97bfd18b5916f7f140db107e720, so we can be reasonably confident that this fixes #2347 now.

PR #2359, opened 2016-03-15
Get the request host from request.get_host

### LINKED ISSUE #2358: Wagtail sites check request.META[HTTP_HOST] directly instead of request.get_host()
@staticmethod
    def find_for_request(request):
        try:
            hostname = request.META['HTTP_HOST'].split(':')[0]

Which doesn't work with reverse proxies that set X-Forwarded-Host and [USE_X_FORWARDED_HOST](https://docs.djangoproject.com/en/1.9/ref/settings/#std:setting-USE_X_FORWARDED_HOST).

Instead it should use `request.get_host()`.

danni:
Fixes #2358

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -116,13 +116,18 @@ class Site(models.Model):
         still be routed to a different hostname which is set as the default
         """
         try:
-            hostname = request.META['HTTP_HOST'].split(':')[0]  # KeyError here goes to the final except clause
+            hostname = request.get_host().split(':')[0]
             try:
                 # find a Site matching this specific hostname
                 return Site.objects.get(hostname=hostname)  # Site.DoesNotExist here goes to the final except clause
             except Site.MultipleObjectsReturned:
                 # as there were more than one, try matching by port too
-                port = request.META['SERVER_PORT']  # KeyError here goes to the final except clause
+                try:
+                    port = request.get_port()
+                except AttributeError:
+                    # Request.get_port is Django 1.9+
+                    # KeyError here falls out below
+                    port = request.META['SERVER_PORT']
                 return Site.objects.get(hostname=hostname, port=int(port))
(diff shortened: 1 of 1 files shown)


danni:
This would be great to get into 1.4.

alexgleason:
Hi @danni, thanks for your contribution! Sorry, we're a bit too late for 1.4, which is [already released](http://docs.wagtail.io/en/v1.4.1/releases/1.4.html)!

danni:
Can't go in a point release? I seems like a bug fix to me.

alexgleason:
Hi Danni, this has been scheduled for release in 1.5. Is this preventing you from being able to proceed with your project? If it's a real emergency, I recommend installing Wagtail 1.4.1 as a local app (instead of using pip) and making your required modifications inside it until 1.5 is released.

As a reference, see [vagrant-wagtail-develop](https://github.com/torchbox/vagrant-wagtail-develop), which installs the wagtaildemo site alongside a local version of Wagtail you can edit. You can replace wagtaildemo with your own site here.

Note that it works by using [setup.py develop](http://stackoverflow.com/questions/12901776/python-setup-py-develop-is-it-possible-to-create-egg-info-folder-not-in-so) to locally install Wagtail.

danni:
No that's fine I'm already running a git editable branch in pip (how I
wrote the fix) it's just annoying to rebase it against whatever version +
the extra deps sucked into our Docker container to build the static assets.

I'm surprised this isn't biting everyone trying to build multisite Wagtails
on anything looking like standard gunicorn + nginx.

alexgleason:
Thoroughly reviewed this and merged in 6fde6863c5684a3f8583d07efa79178129e128c1. I created a PR with tests #2374 to make sure `Site.find_for_request` works under various conditions. I also read the source code for `request.get_host` and `request.get_port` and can confirm that the try-except blocks are correct.

Thank you, @danni. Nice work!

Issue #2353, opened 2016-03-14
Document CSS customization to the userbar (such as changing the icon)


Luperi:
Is there an easy way to customize userbar logo like the admin area one (as explained [here](http://docs.wagtail.io/en/v1.3.1/advanced_topics/customisation/branding.html#branding-logo))?

gasman:
Not currently, no.

alexgleason:
See #2373 for more on this. If you're using an icon font like FontAwesome you can just override the CSS to change the icon.

alexgleason:
I don't think this is something that makes sense to implement. The icon is controlled via CSS and not an image, so overriding it is trivial.

We may want to document the process instead of implementing any kind of hook or override.

SalahAdDin:
:+1:

Luperi:
@alexgleason ok, good. So, should I have to use something like FontAwesome?

gasman:
I haven't tried it, but I'd expect that setting a background image would work too. (If you did that, you'd probably also need something like `.wagtail-icon:before {content: ''}` to disable the existing icon.)

alexgleason:
@Luperi Here's an example of one I made: https://www.youtube.com/watch?v=k2YPEUFGOgw

I think you might want to wait on #2373 before depending on this method. But yes, you can just set:

```css
.wagtail-icon:before {
    content: /* the CSS code for your FontAwesome icon/*;
    font-family: FontAwesome;
}
```

inyoka:
You can recreate templates/wagtailadmin/userbar/base.html, here is an example using a fontawesome pencil icon as a replacement icon in the userbar ...

`{% load staticfiles i18n %}

<div class="wagtail-userbar-reset">
    <div class="wagtail-userbar wagtail-userbar--{{ position|default:'bottom-right' }}" data-wagtail-userbar>
        <link rel="stylesheet" href="{% static 'wagtailadmin/css/userbar.css' %}" type="text/css" />
        <div class="wagtail-userbar-nav">
            <div class="wagtail-userbar-trigger" data-wagtail-userbar-trigger>
              <i class="fas fa-pencil-alt fa-lg"></i>
                <span class="wagtail-userbar-help-text">{% trans 'My admin interface' %}</span>
            </div>
            <div class='wagtail-userbar-items'>
                {% for item in items %}
                    {{ item|safe }}
                {% endfor %}
            </div>
        </div>
[...]

PR #2361, opened 2016-03-15
Branding userbar logo

Luperi:
Introduced the possibility to customize the default wagtail logo in
userbar.

TODO: fix inline css, maybe in userbar.css

### DIFF
--- a/docs/advanced_topics/customisation/branding.rst
+++ b/docs/advanced_topics/customisation/branding.rst
@@ -38,9 +38,9 @@ To replace the default logo, create a template file ``dashboard/templates/wagtai
 To replace the default userbar logo, create a template file ``dashboard/templates/wagtailadmin/userbar/base.html`` that overrides the block ``branding_logo``::
 
     {% overextends "wagtailadmin/userbar/base.html" %}
-    
+
     {% block branding_logo %}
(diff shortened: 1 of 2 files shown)


JoshBarr:
Hey @Luperi, this is a good idea. We often customise the admin so the user never sees the 'bird' icon. It definitely makes sense to be able to customise the icon on the userbar too. 

I noticed you're using inline styles. Is there something in the userbar css that's not behaving correctly?

Luperi:
@JoshBarr I had created in userbar.css two classes, one for the div and another for the img tag. Then I noticed that userbar.css isn't a source of wagtail, but it's generated, I think with [these](http://docs.wagtail.io/en/latest/contributing/developing.html#setting-up-the-wagtail-codebase) instructions. So, I don't know where to put those styles. Thank you.

gasman:
Addresses #2353. @alexgleason makes a very good point there - since the bird icon is defined entirely in CSS, it ought to be possible to override it in CSS too. I'd definitely prefer to go with that approach - it's more lightweight and doesn't require any HTML rewriting.

(In the interests of settling on one approach as the way forward, I'll close this PR now, but it can be reopened later if we find that a template override is the best solution after all)

PR #2362, opened 2016-03-15
Page privacy more possibilities

Luperi:
Now you can choose the privacy of a pages between these options:
1 - Public
2 - Private, visible with a required password
3 - Private, visible if logged in
4 - Private, visible if you are a specific user or in a specific user
group

Implemented a new QuerySet method that allows you to retrieve only the
pages visible by the currently logged user. It can be useful, for
example, when drawing the menu of the main site.

_Example:_
```python
def top_menu(context, parent, calling_page=None):
    menuitems = parent.get_children().live().in_menu().user_can_view(context['request'])
```

### DIFF
--- a/wagtail/wagtailadmin/forms.py
+++ b/wagtail/wagtailadmin/forms.py
@@ -19,7 +19,7 @@ from modelcluster.forms import ClusterForm, ClusterFormMetaclass
 from taggit.managers import TaggableManager
 
 from wagtail.wagtailadmin import widgets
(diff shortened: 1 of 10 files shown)


gasman:
I've posted an update on the current state of this feature at https://github.com/torchbox/wagtail/issues/257#issuecomment-248401768 - since we have several issues and PRs open about this feature, I suggest that any discussion (unless it specifically relates to the code in this PR) is continued there.

gasman:
Now incorporated into #3071.

PR #2373, opened 2016-03-19
Improve userbar styles

alexgleason:
I was wanting to customize the userbar for my site and running into trouble with it. The way the CSS works currently is not ideal. Here's the way the icon is being centered:

screenshot from 2016-03-18 19 51 53

Since it uses a margin to center the bird, it's difficult to customize. You can't increase the icon's size or rotate it without manually adjusting the margin. Here's what happens when you increase the icon size:

screenshot from 2016-03-18 19 52 43

The bird goes out of the circle.

So I changed it to be like this instead:

screenshot from 2016-03-18 19 53 29

It uses flexbox, which is [widely supported](http://caniuse.com/#feat=flexbox) now. I tested it in BrowserStack and it does indeed look pristine in IE11, and even IE10. It's not centered in IE9, but Microsoft has ended support for it and less than 1% of the world is using it.

This is my new code with the size increased:

screenshot from 2016-03-18 19 53 53

Now instead of keeping track of the icon size, container size, _and_ margins, you just need to set the icon size and container size.

The other problem was this thing:

screenshot from 2016-03-18 19 52 05

Not having a class meant I couldn't target it. So I added one:

screenshot from 2016-03-18 19 54 04

Now it can be styled.

That's all. I'm not changing the look at all here; just updating the way the CSS works to make it a bit more extensible. At the very least I need that span. Without it I can't even apply these as custom styles on my own build. I'm forced to fork the userbar or resort to javascript if I want to change it.

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/userbar.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/userbar.scss
@@ -101,7 +101,9 @@ $positions: (
 }
 
     .#{$namespace}-userbar-trigger {
-        display: block;
+        display: flex;
+        align-items: center;
+        justify-content: center;
         width: $size-home-button;
         height: $size-home-button;
         margin: 0 !important;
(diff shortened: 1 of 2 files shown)


alexgleason:
Here's an example for how I'm overriding the styles on one of my sites: https://youtu.be/k2YPEUFGOgw

Simply can't do the rotation with the margins that way. It needs to be perfectly centered.

SalahAdDin:
:+1:

JoshBarr:
It would be preferable to use an `<svg>` for the icon instead of an icon font, since those things are [on the way out](http://ianfeather.co.uk/ten-reasons-we-switched-from-an-icon-font-to-svg/). Then we could provide a simple template override to change the icon.

alexgleason:
Looks like using SVGs as background images has pretty decent support. http://caniuse.com/#feat=svg-css

I think you could just keep the `:after` pseudo-selector and set a width, height, and background-image on it?

gasman:
Hi @JoshBarr - does this look OK to merge now?

gasman:
Merged in 33441f1ff77f71e882f0cdfdbfb0b25896c18fcf - thanks!

PR #2376, opened 2016-03-20
Site.find_for_request with IPv6 hostnames - fixes #2375

### LINKED ISSUE #2375: Site.find_for_request doesn't deal with IPv6 hostnames
I suggest [django.http.requests.split_domain_port](https://github.com/django/django/blob/53e8ab580f7c0fcfc589ba0b1b6cc2556080e0b2/django/http/request.py#L516).

We have [this](https://github.com/torchbox/wagtail/blob/6fde6863c5684a3f8583d07efa79178129e128c1/wagtail/wagtailcore/models.py#L111):

``` python
hostname = request.get_host().split(':')[0]
```

but we could do this:

``` python
hostname, port = split_domain_port(request.get_host())
```

Currently splitting the hostname by `:` will erroneously take the first chunk of an IPv6 address. As a reminder, they look like this: `[2001:db8:85a3:8d3:1319:8a2e:370:7348]:443` and could be with or without the port. `split_domain_port` works around this.

alexgleason:


kaedroho:
Thanks @alexgleason. This just needs a test

wagtail/wagtailcore/models.py
```diff
@@ -108,7 +109,7 @@ def find_for_request(request):
         still be routed to a different hostname which is set as the default
         """
         try:
-            hostname = request.get_host().split(':')[0]
+            hostname, port = split_domain_port(request.get_host())
```
gasman: Presumably, retrieving `port` here means that we can eliminate the logic below which performs this as a separate step (lines 118-123).

BertrandBordage:
Hi @alexgleason,
Are you still willing to apply the changes suggested by Karl & Matt?
Thanks!

alexgleason:
Would anyone else be willing to take this one? I don't know the best way to write a test for this.

lb-:
Superseded by #5997

PR #2383, opened 2016-03-21
Added --schema-only flag to update_index command

### LINKED ISSUE #2171: --no-data flag for update_index command
Sites that want to run unit tests with Elasticsearch need to make sure the schema is installed in Elasticsearch before running the tests. Otherwise, Elasticsearch will try to guess the schema.

You can't run ``update_index`` on a CI server as there is usually no database available. So we need a command that installs the schema but doesn't attempt to read from the database.

I'm thinking adding a ``--no-data`` flag to the ``update_index`` command would be the best way to get this. It will be run just before running the tests:

```shell
python manage.py update_index --no-data
python manage.py test
```

kaedroho:
Fixes #2171

### DIFF
--- a/docs/reference/management_commands.rst
+++ b/docs/reference/management_commands.rst
@@ -76,6 +76,17 @@ For example, to update just the default backend:
     python manage.py update_index --backend default
 
 
+Indexing the schema only
+````````````````````````
+
+.. versionadded:: 1.5
+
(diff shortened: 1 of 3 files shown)


gasman:
Looks good to me - merged in b978fb826fd57f18460e9a1498c8205073751efa + parents.

PR #2386, opened 2016-03-22
Fix userbar not opening on devices with touch + mouse support. 

### LINKED ISSUE #2385: Userbar doesn't open with mouse
I have a touchscreen laptop, which means my browser returns true for the ontouchstart check in the  userbar setup, disabling my ability to use the userbar with a mouse.

JoshBarr:
Fixes #2385 for devices with touch and click support. @gasman should we add this to the next point release for 1.4 so the fix can be used immediately?


kaedroho:
Thanks @JoshBarr. This looks good to me
Merged in 0b9f78ef9d5a819d479bb6dd0a5e25fe6ea4c061 (master) and 2ce9bda1a34c6d207077c9a7f86c559642730fbe (1.4.x)

Issue #2310, opened 2016-03-04
Page.search_panels and Page.content_panels are too restrictive about how they can be appended to


coredumperror:
As I just discovered while trying to standardize my model code, `Page.search_panels` _only_ allows tuples to be appended to it, while `Page.content_panels` _only_ allows lists. Both really should allow any iterable, or at the very least accept a common type of iterable. Attempting to append a tuple onto `content_panels` gives `TypeError: can only concatenate list (not "tuple") to list`, while attempting to append a list onto `search_panels` gives `TypeError: can only concatenate tuple (not "list") to tuple`.

As it stands right now, extensions to these two attributes on custom models have to look _exactly_ like this: 

```python
class StaticPage(Page):
    body = StreamField([
        ('heading', blocks.CharBlock(classname="full title")),
        ('paragraph', blocks.RichTextBlock()),
        ('assets', CarouselBlock()),
    ])
    image = models.ForeignKey('CustomImage', null=True, blank=True, on_delete=models.SET_NULL, 
        related_name='+'
    )

    content_panels = Page.content_panels + [
        StreamFieldPanel('body'),
        ImageChooserPanel('image'),
    ]

    search_fields = Page.search_fields + (
        index.SearchField('body'),
    )
```
Which is frustrating for developers who'd prefer to use a single type of iterable throughout.

There may be alternate ways to bring is the base values of these attributes from `Page`, but I don't know them off hand, and the Wagtail docs don't mention them.

alexgleason:
Yes! I was just having this problem and wanting to bring it up. Thanks.

kaedroho:
#2388 is now merged which now allows ``search_fields`` to be set to a list. This will become mandatory in Wagtail 1.7.

coredumperror:
Awesome!

alexgleason:
Thank you!

gasman:
@SalahAdDin Yes - that's because the fix has not been released yet. It'll be in Wagtail 1.5.

PR #2343, opened 2016-03-11
Remove deprecations and version-bump to start work on 1.5

gasman:

PR #2388, opened 2016-03-22
Change Indexed.search_fields to be a listish thing

timheap:
Indexed.search_fields used to be a tuple. This is incorrect, and it should have been a list.  Changing it to be a list now would be a backwards incompatible change, as people do

    search_fields = Page.search_fields + (
        SearchField('body')
    )

Adding a tuple to the end of a list causes an error, so this would cause all old code that used tuples to throw an error. This is not great.

A new ThisShouldBeAList class, which subclasses list, has been added. It additionally allows tuples to be added to it, as in the above behaviour, but will raise a deprecation warning if someone does this. Old code that uses tuples will continue to work, but raise a deprecation warning.

See #2310

### DIFF
--- a/wagtail/utils/deprecation.py
+++ b/wagtail/utils/deprecation.py
@@ -55,5 +55,5 @@ class SearchFieldsShouldBeAList(ThisShouldBeAList):
     """
     Indexed.search_fields was a tuple, but it should have been a list
     """
-    def __init__(self, items, name='search_fields', category=RemovedInWagtail16Warning):
+    def __init__(self, items, name='search_fields', category=RemovedInWagtail17Warning):
         super(SearchFieldsShouldBeAList, self).__init__(items, name, category)
--- a/wagtail/utils/deprecation.py
+++ b/wagtail/utils/deprecation.py
@@ -1,3 +1,6 @@
+import warnings
+
+
 class RemovedInWagtail16Warning(DeprecationWarning):
     pass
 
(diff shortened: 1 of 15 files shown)


wagtail/utils/deprecation.py
```diff
@@ -7,3 +10,50 @@ class RemovedInWagtail15Warning(DeprecationWarning):
...
+        search_fields = Page.search_fields + (
+            SearchField('body')
+        )
+
+    Adding a tuple to the end of a list causes an error.
+
+    This class will allow tuples to be added to it, as in the above behaviour,
+    but will raise a deprecation warning if someone does this.
+    """
+    message = '{name} should be a list, not a {type}'
```
kaedroho: The usual tone for deprecation warnings in Wagtail is usually something more on the lines of: "Setting {name} to a {type} is deprecated, please use a list instead."

wagtail/utils/deprecation.py
```diff
@@ -7,3 +10,50 @@ class RemovedInWagtail15Warning(DeprecationWarning):
...
+            rhs = list(rhs)
+        return cls(super(ThisShouldBeAList, self).__add__(list(rhs)),
+                   name=self.name, category=self.category)
+
+
+class SearchFieldsShouldBeAList(ThisShouldBeAList):
+    """
+    Indexed.search_fields was a tuple, but it should have been a list
+    """
+    def __init__(self, items, name='search_fields', category=RemovedInWagtail16Warning):
```
kaedroho: This should be changed to ``RemovedInWagtail17Warning``

kaedroho:
Thanks @timheap! I'm really glad to see this finally change :+1:

wagtail/utils/deprecation.py
gasman: That would require someone to review and merge #2343 (hint hint) :-)

timheap:
Rebased on top of #2343 to get the new deprecation warning, the wording has been updated, and flake8 warnings fixed.

kaedroho:
Merged in 6bd168580ef4d91da077dd23e43764af570caf28 thanks!

PR #2392, opened 2016-03-23
Fix DeprecationWarnings from Django 1.9

timheap:
The output of `./runtests.py` was so full of DeprecationWarnings, RuntimeWarnings, and so on that it was near useless. This PR aims to clean the output up some what.

The console output is full of deprecation warnings produced by Wagtail that need fixing. It is also full of deprecation warnings from third party packages such as django-taggit. I've made the test suite only print deprecation warnings produced by Wagtail by default, so the warnings are easier to deal with. There is a `--full-warnings` flag that can be passed to `./runtests.py` to enable all deprecation warnings again,  if we want to try and fix warnings in third party modules.

Wagtail was still using some of its own deprecated internal methods, despite having a documented alternative available. I've fixed these warnings as well.

There were two places still using `field.related` instead of `field.rel`, and `rel.to` over `rel.model`. These have been fixed. Django 1.9 renamed `field.rel` to `field.remote_model`, deprecating the old `field.rel`, so this will need to happen again in the future.

`Model._meta.get_fields()` should be used, instead of the old `Model._meta.get_all_related_objects()`. [Django has instructions on migrating this code](https://docs.djangoproject.com/en/1.9/ref/models/meta/#migrating-from-the-old-api).

All management commands now use argparse/`add_arguments()` over optparse/`options_list`.

Some tests used naive datetimes, which was causing needless RuntimeWarnings.

Elasticsearch was complaining about not verifying SSL certificates for HTTPS connections. This change is mildly backwards incompatible, as SSL certificates are now verified by default. People will have to change their configurations if they are using Elasticsearch over HTTPS, with an invalid SSL certificate, through the automatic URLS parameter. A release note has been added, with further discussion in the relevant commit.

Some tests were missing `from __future__ import unicode_literals`, which was causing a RuntimeWarning from unidecode via django-taggit. `unicode_literals` have been enabled in the relevant tests. I will create a different PR which adds `unicode_literals` to all files in the repo, as this is a larger and slightly unrelated change.

The end result of all this is that the output of `./runtests.py` on Django 1.8 is now free of warnings!

### DIFF
--- a/docs/releases/1.5.rst
+++ b/docs/releases/1.5.rst
@@ -57,3 +57,37 @@ Should be changed to:
         ]
 
 To ease the burden on third-party modules, adding tuples to ``Page.search_fields`` will still work. But this backwards-compatibility fix will be removed in Wagtail 1.7.
+
+Elasticsearch backend now defaults to verifying SSL certs
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Previously, if you used the Elasticsearch backend, configured with the URLS property like:
+
+
+.. code-block:: python
+
+    WAGTAILSEARCH_BACKENDS = {
+        'default': {
+            'BACKEND': 'wagtail.wagtailsearch.backends.elasticsearch',
+            'URLS': ['https://example.com/'],
+        }
+    }
+
+Elasticsearch would not be configured to verify SSL certificates for HTTPS URLs. This has been changed so that SSL certificates are verified for HTTPS connections by default.
+
+If you need the old behaviour back, where SSL certificates are not verified for your HTTPS connection, you can configure the Elasticsearch backend with the ``HOSTS`` option, like so:
+
+.. code-block:: python
+
+    WAGTAILSEARCH_BACKENDS = {
+        'default': {
+            'BACKEND': 'wagtail.wagtailsearch.backends.elasticsearch',
+            'HOSTS': [{
+                'host': 'example.com'
+                'use_ssl': True,
+                'verify_certs': False,
+            }],
+        }
+    }
+
+See the `Elasticsearch-py documentation <http://elasticsearch-py.readthedocs.org/en/stable/#ssl-and-authentication>`_ for more configuration options.
--- a/wagtail/wagtailcore/management/commands/publish_scheduled_pages.py
+++ b/wagtail/wagtailcore/management/commands/publish_scheduled_pages.py
@@ -1,7 +1,6 @@
 from __future__ import print_function
 
 import json
-from optparse import make_option
 
 from django.core.management.base import BaseCommand
 from django.utils import dateparse, timezone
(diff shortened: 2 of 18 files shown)


SalahAdDin:
:+1:

timheap:
Alright, this is no longer WIP. All DeprecationWarnings and PendingDeprecationWarnings have been fixed for Django 1.8; which fixes all DeprecationWarnings for Django 1.9. I can not fix the PendingDeprecationWarnings for Django 1.9 easily, as fixing those breaks Django 1.8 compatibility. As 1.10 is not due to be released for some time, dropping support for 1.8 is not sensible just yet. We can live with the warnings in 1.9.

By default, only warnings for Wagtail are shown. I can see a sensible argument for showing warnings generated by Willow, django-modelcluster, and other packages written for Wagtail. Should I enable these warnings by default? All warnings can be switched on using `./runtests.py --deprecation=all`, so these warnings can still be found.

I've added some more commits, descriptions are in the ticket description above.

timheap:
I've also fixed many of the deprecation warnings for django-taggit, to ensure Wagtail continues to work with future versions of Django: https://github.com/alex/django-taggit/pull/385

docs/releases/1.5.rst
```diff
@@ -27,3 +27,37 @@ Bug fixes
...
+    WAGTAILSEARCH_BACKENDS = {
+        'default': {
+            'BACKEND': 'wagtail.wagtailsearch.backends.elasticsearch',
+            'URLS': ['https://example.com/'],
+        }
+    }
+
+Elasticsearch would not be configured to verify SSL certificates for HTTPS URLs. This has been changed so that SSL certificates are verified for HTTPS connections by default.
+
+If you need the old behaviour back, where SSL certificates are not verified for your HTTPS connection, you can configure the Elasticsearch backend with the HOSTS option, like so:
```
kaedroho: HOSTS should be backticked like ` `HOSTS``

wagtail/wagtailcore/management/commands/publish_scheduled_pages.py
```diff
@@ -21,15 +20,10 @@ def revision_date_expired(r):
...
-            action='store_true',
-            dest='dryrun',
-            default=False,
-            help='Dry run -- don\'t change anything.'
-        ),
-    )
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--dryrun', action='store_true', dest='dryrun', default=False,
+            help="Dry run -- do't change anything.")
```
kaedroho: do't => don't

kaedroho:
Looks good to me!

timheap:
I've fixed the issues you mentioned, and rebased on master

kaedroho:
I've merged all except 59599d4 into master at b3aa292110687eecb91cd74a4d7d4354833e6f37 +parents

I left 59599d4 out for now as it prevents extra options being passed through to ``dj test``. For example the ``--keepdb`` option no longer works.

I've cheery-picked the following commits into the ``stable/1.4.x`` branch:

e003140 Use add_arguments over deprecated option_list
904441d Use _meta.get_fields() over deprecated _meta.get_all_related_objects()
90bcabd Use field.rel over field.related, rel.model over rel.to
c92ed90 Stop using methods deprecated in Wagtail

timheap:
Thanks for the merge.

You can still pass through extra arguments to `dj test`, by placing them after a `--`, such as:

```
$ ./runtests.py --deprecation=all -- --keepdb
```

timheap:
I'll close this one and make another PR for that commit, to keep this discussion and history clean and separate.

PR #2396, opened 2016-03-23
Support for unicode slug in url

nilnvoid:
support for unicode slug, using django 1.9.4 fix for slugify util


chimeno:
Hello,
I'm not a core member but IMHO I would say dropping support for django 1.8 is not a good idea since ``allow_unicode`` parameter  was introduced in django 1.9 [slugify docs](https://docs.djangoproject.com/es/1.9/ref/utils/#django.utils.text.slugify)
As this is not backwards compatible, we must provide a backward compatible way or wait until django 1.9 being the minimum version.

b3h74d:
Hello,

Thank you for the comments, @chimeno I will look for a fix to get this work with dj 1.8 too.

README.rst
```diff
@@ -70,7 +70,7 @@ Wagtail is sponsored by `Torchbox <https://torchbox.com/>`_. If you need help im
 
 Compatibility
 ~~~~~~~~~~~~~
-Wagtail supports Django 1.8.1+ on Python 2.7, 3.3, 3.4 and 3.5. Supported database backends are PostgreSQL, MySQL and SQLite.
```
auvipy: this should remain as is

b3h74d: Django 1.8 compatibility needs work, will do in a day or two.

nilnvoid:
This is ready for test now.

auvipy:
It would be great If you re base the commits and separately implement logical commits.

nilnvoid:
@auvipy may you please give e tutorial for this? I am new to github.

nilnvoid:
Now, this is ready for review.
We are using a clone of my repo on [http://blog.webramz.com](url), it is in persian language and unicode slug urls work well.

wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
```diff
@@ -315,9 +315,9 @@ function InlinePanel(opts) {
 
 function cleanForSlug(val, useURLify) {
     if (URLify != undefined && useURLify !== false) { // Check to be sure that URLify function exists, and that we want to use it.
-        return URLify(val);
+        return URLify(val, true);
```
gasman: This needs to be `URLify(val, null, true)` - the second parameter is `num_chars`. Also, it looks like we need to import `js/vendor/xregexp/xregexp.min.js` from Django admin when unicode slugs are in use.

gasman:
Thanks @nilnvoid!

Is the change to `wagtail/wagtailsnippets/views/snippets.py` related to this one, or is that fixing a different bug?

gasman:
Merged in b1f194a6370bc9aa1829eda636a3dedb0687673d - thanks very much!

I've left out the fix to `wagtail/wagtailsnippets/views/snippets.py` - please open a new PR with a description of the bug it's fixing, if this is still required.

PR #2399, opened 2016-03-25
Fix for unicode names in update_index on Python 2.7.x

rafadev:
When running update_index on Python 2.7.x concatenation of non-ascii model verbose_name_plural was generating an error particularly on line 57 of update_index.py. This solution would also work on Python 3.3+ as it does support u"" strings. I've modified it in every instance in case the backend name also has unicode characters (would be rare but can technically happen).

### DIFF
--- a/wagtail/wagtailsearch/management/commands/update_index.py
+++ b/wagtail/wagtailsearch/management/commands/update_index.py
@@ -18,7 +18,7 @@ class Command(BaseCommand):
 
     def update_backend(self, backend_name, object_list):
         # Print info
-        self.stdout.write("Updating backend: " + backend_name)
+        self.stdout.write(u"Updating backend: " + backend_name)
 
         # Get backend
         backend = get_search_backend(backend_name)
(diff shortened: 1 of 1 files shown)


wagtail/wagtailsearch/management/commands/update_index.py
```diff
@@ -18,7 +18,7 @@ def get_object_list(self):
 
     def update_backend(self, backend_name, object_list):
         # Print info
-        self.stdout.write("Updating backend: " + backend_name)
+        self.stdout.write(u"Updating backend: " + backend_name)
```
m1kola: I think would be better to use `six.text_type` for these strings.

m1kola:
@rafadev thanks!

Can you add tests for case with unicode string?

rafadev:
@m1kola sure!! didn't know about six's text_type, I'm new to python 3. I'll correct it and add a test case.

timheap:
Using `six.text_type` here should be unnecessary. The `u""` prefix on the strings should take care of that. It will make `str` types in Python3, and `unicode` types in Python2, which is all the `six.text_type` function does.

As an alternative, you can add `from __future__ import unicode_literals` to the top of the file. This will cause all strings to be the correct 'text type' on all versions of Python, meaning that developers do not have to remember to prefix strings with `u""` in these cases.

m1kola:
> Using `six.text_type` here should be unnecessary. The `u""` prefix on the strings should take care of that.

@timheap, if I'm not mistaken Python >=3.0,<3.3 doesn't support `u""` literals, so better to use `six.text_type`.

But I agree with you about `from __future__ import unicode_literals` - this probably will be better way to solve this issue.

@rafadev are you happy to make these changes?

timheap:
@m1kola You are correct, however Wagtail only supports Python 3.4 and 3.5. Pythons 3.3 and below are not supported, so we can use `u""`.

rafadev:
@m1kola @timheap , the reason I used u"" is exactly that, I checked on Wagtail requirements and it asked for Python 2.7 or 3.4+ so I figured out u"" would work in all cases. I guess it would be a matter of adding a test case and running them on both Python versions and that would wrap this pull request right?

What's the expected launch date for 1.5? I'm pretty busy but I want to complete this PR for the next version.

kaedroho:
As of fead70bee325b5de7a51478cc1611606e0b4fe58 ``verbose_name_plural`` is no longer used in the ``update_index`` command so this should be fixed now but thanks for submitting this patch anyway!

rafadev:
@kaedroho Cool, hope I can help in the future, needless to say I'm loving Wagtail, going for my 2nd Wagtail powered site right now.

Issue #2371, opened 2016-03-17
Can't delete user


Luperi:
Why can't I delete users from wagtailadmin area, also if user logged in has delete permission on User model?

alexgleason:
Good point. I think this will need some design/UX work, as we need to determine what should happen to existing content owned by the user when after the user is deleted.

WordPress gives you a radio box on the confirmation screen asking if you want to transfer all content to a different user or to erase associated content entirely.

gasman:
On all models that have a concept of ownership (Page, Image, Document...), we have 'owner' set up as a nullable field (because otherwise you'd have to invent a fake user account when doing automated data imports). Assuming we've got those fields set up properly with `on_delete=SET_NULL`, then, there should be no problem with keeping the user's content around. (I can't think of any case where you'd actually want the content to be deleted when removing a user.)

I think a standard delete confirmation screen will do the job here. Not sure why we don't have one - just an oversight, I guess.

Luperi:
@gasman No chances to see this feature in Wagtail 1.5?

gasman:
@Luperi There's plenty of chance to see it in 1.5, assuming someone implements it :-) I can't promise to work on it myself, as there are lots of other high priority items to get through, but that shouldn't stop someone else from working on it.

vincentaudebert:
Oops! Hold on on this one please. I totally missed the existing PR and fixed the issue in quite a different way. 
From what I get the existing PR "only" adds the delete button on the edit page. I added it in the listing too.
I will compare my code and the existing PR tomorrow (NZ time) and try to keep the best from it (especially the tests).

Sorry about this :s

gasman:
Completed in #2523.

PR #2402, opened 2016-03-25
Add delete option to the user admin interface

ltakens:


alexgleason:
Thank you @ltakens! We will review it. This would fix #2371.

wagtail/wagtailusers/views/users.py
```diff
@@ -126,3 +126,16 @@ def edit(request, user_id):
         'user': user,
         'form': form,
     })
+
+
+@permission_required(delete_user_perm)
+def delete(request, user_id):
+    user = get_object_or_404(User, id=user_id)
+    if request.POST:
```
kaedroho: This should be changed to ``if request.method == 'POST':``

This is so the view works, even if there's not ``POST`` data.

wagtail/wagtailusers/tests.py
```diff
@@ -164,6 +164,40 @@ def test_edit_validation_error(self):
...
+    def post(self, post_data={}, user_id=None):
+        return self.client.post(reverse('wagtailusers_users:delete', args=(user_id or self.test_user.id, )), post_data)
+
+    def test_simple(self):
+        response = self.get()
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, 'wagtailusers/users/confirm_delete.html')
+
+    def test_delete(self):
+        response = self.post({'posted': True})
```
kaedroho: Fixing https://github.com/torchbox/wagtail/pull/2402/files#r57518801 would make ``{'posted': True}`` unnecessary.

wagtail/wagtailusers/views/users.py
timheap: See #2424 for an example of fixing this across the Wagtail code base.

wagtail/wagtailusers/templates/wagtailusers/users/edit.html
```diff
@@ -27,15 +27,20 @@
                     {% include "wagtailadmin/shared/field_as_li.html" with field=form.password1 %}
                     {% include "wagtailadmin/shared/field_as_li.html" with field=form.password2 %}
                     {% include "wagtailadmin/shared/field_as_li.html" with field=form.is_active %}
-                    
-                    <li><input type="submit" value="{% trans 'Save' %}" /></li>
+
+                    <li>
+                        <input type="submit" value="{% trans 'Save' %}" />
+                        <a href="{% url 'wagtailusers_users:delete' user.id %}" class="button button-secondary no">{% trans "Delete user" %}</a>
```
gasman: We should have a permission check around this button, so that it doesn't get shown to users who have edit but not delete permission.

gasman:
Thanks @ltakens! We've received several other PRs that implement this, and I've chosen to go with #2523 as it's the most complete - you'll still get a credit in the release though :-)

PR #2403, opened 2016-03-27
Add docs on making Wagtail compatible with multiple Django versions

timheap:
A pull request recently submitted (#2359) used some functionality added in Django 1.9, with a work around for older version of Django. The code works fine, but goes against our coding guidelines for making Wagtail compatible across versions. This was our fault though, as the preferred method of making code cross-compatible was not documented anywhere, so both the committer and the reviewer were unaware of the correct style.

This commit adds documentation on our preferred method for making Wagtail cross-compatible with multiple Django versions. It should hopefully be useful to both contributors making changes to Wagtail, and for reviewers to give an example on the correct style when reviewing some work.


timorieber:
Very helpful and looks good to me.

kaedroho:
Thanks @timheap!

PR #2404, opened 2016-03-27
Define explorer menu media with property instead of class

### LINKED ISSUE #2369: Regression when using ManifestStaticFilesStorage with 1.4
If using ManifestStaticFilesStorage as your STATICFILES_STORAGE, https://github.com/torchbox/wagtail/commit/824a7c50e272c3f1492551f015716a771f8cb639 introduces a requirement that your manifest file is present when loading your app.

This can be verified by putting

```
STATICFILES_STORAGE = 'django.contrib.staticfiles.storage.ManifestStaticFilesStorage'
```

into wagtail/tests/settings.py, and running the tests on 1.3 and 1.4.

I run celery in an asset-less container, which now throws exceptions on startup when using Wagtail 1.4.

kaedroho:
Fixes #2369

The static() function was being called during app load which caused a crash when the user is using STATICFILES_STORAGE=ManifestStaticFilesStorage, DEBUG=False and haven't yet collected static files.

I've moved it into a property and it's now only called when a view is being rendered.

I haven't yet come up with a reliable way to test this. I don't think it's worth holding this fix for that though.

### DIFF
--- a/wagtail/wagtailadmin/wagtail_hooks.py
+++ b/wagtail/wagtailadmin/wagtail_hooks.py
@@ -1,8 +1,8 @@
+from django import forms
 from django.contrib.auth.models import Permission
 from django.contrib.staticfiles.templatetags.staticfiles import static
 from django.core.urlresolvers import reverse
 from django.utils.translation import ugettext_lazy as _
-from django import forms
 
 from wagtail.wagtailadmin.menu import MenuItem, SubmenuMenuItem, settings_menu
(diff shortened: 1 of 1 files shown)


tomdyson:
@m1kola could you check this (very small change) please?

m1kola:
Thanks @kaedroho!

Merged in dcc89ce9b1cedcfe842f19f1d21aa9be4bae2e6b (master) and in db269928e8f09e2a0c646224a3fdaafd3049b22f (stable/1.4.x)

PR #2413, opened 2016-03-29
performance optimization - prefetch root page with site object

ychab:
This is minimalist but we can prevent one useless query per hits. This change should be safe and in the better place... isn't it?


gasman:
Thanks for the suggestion! The `find_for_request` code has since been rewritten (#1963), but I've applied the equivalent of this fix in 1d2ec402fa9538152edee7e1c768586145511e7c.

Issue #158, opened 2014-03-19
Non-swappable user model and data 


bennylope:
I noticed that #58 did a little bit of work to make [swappable users](https://docs.djangoproject.com/en/1.6/topics/auth/customizing/#substituting-a-custom-user-model) work with Wagtail but there's still quite a lot of coupling to the `auth.User` model. 

I've started some work to amend this on [my fork](https://github.com/bennylope/wagtail/tree/fix-swappable-users), including updating the schema migrations (in the [main migrations](https://github.com/bennylope/wagtail/commit/3c39bb97bd990fb85a35929c16e04368d674559d) and [frozen ORM](https://github.com/bennylope/wagtail/commit/b3b1e2431b9ac381b02c2ef77ae8112bb609fd71)) but it requires additional work. I'd love to decouple this a bit since we regularly use custom user models, especially substituting an email address for username, and that would enable us and maybe others to use Wagtail.

Before I do any more, is there any interest in decoupling further, or is the team set on using `auth.User` and it's fields for Wagtail?

gasman:
Hi @jsma,
I think the best approach is somewhere in between. Having to override the user management interface at the URL/view level certainly seems heavy-handed, and making the existing views flexible enough to accommodate custom User models would be much neater.

Given that Django 1.7 deprecates the pattern of a ModelForm that automatically picks up its field list from the model (and pushes you towards an explicit `fields` attribute instead), I think we ought to follow suit and require custom fields to be specified explicitly. Ideally we'd be able to do that by allowing the user to pass in a custom replacement for UserCreationForm / UserEditForm - but I see that wouldn't be enough as it stands, because the list of fields is also currently hard-coded in create.html / edit.html. Perhaps the answer is to shift some of that template structure back into the form objects - they could define overrideable `account_fields` and `roles_fields` properties, which can then be used to assemble the final `fields` property, as well as being looped over in the create.html / edit.html templates.

gasman:
Fixed in #1388 (as far as allowing custom user models without a username field, at least - feel free to open new tickets if there are any residual issues around customisability of the user edit forms, or whatever...)

PR #2414, opened 2016-03-29
Adds ability to override user creation and edit forms

nfletton:
This change allows fields from custom user models to be editable via the Wagtail user admin create and update views.

This could be done already but required excessive overriding of Wagtail elements. There's no specific issue open related to this PR but the issue and potential resolutions were discussed in the comments of #158

For details of how the customization is achieved, please refer to the [documentation included in the PR](/nfletton/wagtail/blob/feature/users/custom_forms/docs/advanced_topics/settings.rst#custom-user-edit-forms).

### DIFF
--- a/docs/advanced_topics/settings.rst
+++ b/docs/advanced_topics/settings.rst
@@ -274,6 +274,113 @@ Case-Insensitive Tags
 
 Tags are case-sensitive by default ('music' and 'Music' are treated as distinct tags). In many cases the reverse behaviour is preferable.
 
+Custom User Edit Forms
+----------------------
+
+.. code-block:: python
+
+  WAGTAIL_USER_EDIT_FORM = 'users.forms.CustomUserEditForm'
+
+Allows the default ``UserEditForm`` class to be overridden with a custom form when
+a custom user model is being used and extra fields are required in the user edit form.
+
+.. code-block:: python
+
+  WAGTAIL_USER_CREATION_FORM = 'users.forms.CustomUserCreationForm'
+
+Allows the default ``UserCreationForm`` class to be overridden with a custom form when
+a custom user model is being used and extra fields are required in the user creation form.
+
+.. code-block:: python
+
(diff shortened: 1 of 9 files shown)


nfletton:
@gasman Any chance for this one to be considered for a release. Thanks.

docs/advanced_topics/settings.rst
```diff
@@ -274,6 +274,113 @@ Case-Insensitive Tags
...
+Allows the default ``UserCreationForm`` class to be overridden with a custom form when
+a custom user model is being used and extra fields are required in the user creation form.
+
+.. code-block:: python
+
+  WAGTAIL_USER_CUSTOM_FIELDS = ['country']
+
+A list of the extra custom fields to be appended to the default list.
+
+Custom user forms example
```
gasman: I think this section really needs to be moved elsewhere, possibly a new page under Advanced Topics -> Customising Wagtail. This page is already a bit out of control - it starts off as a tutorial in how to set up Wagtail from scratch from a barebones Django project, then segues into a Wagtail settings reference. Segueing into a different tutorial is taking things a bit too far :-)

gasman:
Other than the above quibble about the placement of the documentation, this all looks good to me!

gasman:
Merged in c4feb6462fb75483aec217e597d64e744fcbb95d, with documentation reshuffled as per my suggestion in 6ff98a543b00bbe63ad04ac8af1ba0a07772df84. Thanks!

A little curiosity: the `{% extends %}` lines in the example template code raised alarm bells for me, because historically it's not been possible to extend the template you're overriding without creating an infinite loop - hence the need for [django-overextends](https://github.com/stephenmcd/django-overextends) as a workaround. On further investigation it appears that they quietly fixed it in Django 1.9 (https://docs.djangoproject.com/en/1.9/releases/1.9/#templates - "Django template loaders can now extend templates recursively"), and up to now no-one seems to have noticed that this makes django-overextends obsolete...

PR #2415, opened 2016-03-29
Updated specific_iterator to work per Django docs

AdamBolfik:
If the values() clause precedes the annotate() clause, any annotations will be automatically added to
the result set. However, if the values() clause is applied after the annotate() clause, you need to explicitly include the aggregate column. - https://docs.djangoproject.com/en/dev/topics/db/aggregation/#order-of-annotate-and-values-clauses

### DIFF
--- a/wagtail/wagtailcore/query.py
+++ b/wagtail/wagtailcore/query.py
@@ -255,7 +255,13 @@ def specific_iterator(qs):
 
     This should be called from ``PageQuerySet.specific``
     """
-    pks_and_types = qs.values_list('pk', 'content_type')
+    # Dynamically grab any annotated fields on the queryset
+    # https://docs.djangoproject.com/en/dev/topics/db/aggregation/#order-of-annotate-and-values-clauses
+    qs_fields = ['pk', 'content_type'] + list(qs.query.annotations.keys())
+    pks_and_types = qs.values_list(*qs_fields)
+    # Remove the annotated fields previously added
+    pks_and_types = [(values_tuple[0], values_tuple[1]) for values_tuple in pks_and_types]
+
     pks_by_type = defaultdict(list)
     for pk, content_type in pks_and_types:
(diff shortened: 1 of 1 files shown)


AdamBolfik:
We were getting a FieldError when accessing the index page for 'Pages' in the admin. We traced it down to when 'annotation' is used, you must include the annotated field when retrieving the values or values_list. 

We are wondering how we are the only ones with this problem, surely other people have tried viewing some Page models in the admin with 1.4.1....?

AdamBolfik:
Sorry if I left out the context of the problem... 

In the index view in wagtailadmin.views.pages, line 67:
`
        pages = pages.annotate(
            null_position=Count('latest_revision_created_at')
        ).order_by('-null_position', '-latest_revision_created_at')`

gasman:
Hi @AdamBolfik,

Afraid I'm still a bit confused about what this is fixing. The 'Pages' index page works fine here, when ordering on the 'updated' column either forwards or reversed - what's the exact error you're getting?

I can see there's a possible issue with the annotations not being present on to the model instances when iterating over the results of `specific()`, i.e.:

    [
        page.null_position
        for page in Page.objects.annotate(null_position=Count('latest_revision_created_at'))
    ]

works, but

    [
        page.null_position
        for page in Page.objects.specific().annotate(null_position=Count('latest_revision_created_at'))
    ]

(or the same with `specific()` after the `annotate()` clause) fails with `AttributeError: 'Page' object has no attribute 'null_position'`. However, this omission doesn't affect the page listing, because the annotation is only used for the `order_by` clause (which is handled in the initial query, before the `specific()` logic kicks in) - and in any case, your patch doesn't seem to fix this, so I'm probably looking at the wrong thing.

Please can you clarify (ideally with a test we can add to `wagtailcore/tests/test_page_queryset.py` that demonstrates the failure)?

gasman:
Please reopen if you can provide the requested info.

Issue #2409, opened 2016-03-28
Deprecate Hallo.js 


JoshBarr:
The last commit on the Hallo.js master was Jun 12, 2015. I'm worried we're coupling a crucial function in Wagtail (editing text) to a package which seems to have gone off the boil somewhat. 

I would suggest migrating to Draft.js, as this supports our aspirations to move to a more React-driven frontend, as well as being more actively maintained and sporting a much nicer API to write plugins against. It's also used in production by _Facebook_. 

---

Edit: see progress updates at

- https://github.com/wagtail/wagtail/issues/2409#issuecomment-264019860
- https://github.com/wagtail/wagtail/issues/2409#issuecomment-277226739
- https://github.com/wagtail/wagtail/issues/2409#issuecomment-297354841

-----

There's some prerequisite work to do:
- [x] Allow for pluggable rich text editors (so that we don't backwards-break every Wagtail site) [#2416]
- [x] Create a new JSONField and a widget to control it. 
- [x] Allow for HTML export from a structured text editor, so that people who want to store Rich Text as HTML in the database can continue to do so. 
- [x] Write an extension to the `richtext` template filter to parse structured data into HTML. 

Check out the [Draft.js intro site](https://facebook.github.io/draft-js/)

Read an article about how Draft represents data on [Medium](https://medium.com/@rajaraodv/how-draft-js-represents-rich-text-data-eeabb5f25cf2#.cswovooto)

There's a solution for saving HTML if required, too: [Draft JS HTML export](https://github.com/sstur/draft-js-export-html)

alexgleason:
Nice. Love the way it represents data. I think everyone is in favor of switching to a better editor.

Side note: was playing with the demo hoping you could press tab to indent lists (see: #2212) but it doesn't seem so. Would be awesome if there's a way to do that with this.

JoshBarr:
I think Draft supports that tab-to-create-list functionality, but the demo doesn't have it enabled.

SalahAdDin:
:+1:

robmoorman:
Draft.js is perfect, +1!

ababic:
I wasn't overly impressed with the demo on the Draft.js homepage, but it's encouraging to see projects like https://github.com/sstur/react-rte (based on Draft.js). Their demo is much closer to how I'd expect a default rich-text widget in a CMS to behave. It looks like they got the 'tab-for-list-indentation' functionality enabled too, if you'd like to see it in action @alexgleason.

[...]

PR #2416, opened 2016-03-29
Configurable rich text editors

kaedroho:
This PR implements the ability to configure rich text editors using a Django setting (https://github.com/torchbox/wagtail/pull/1521#issuecomment-137001133). It builds on ideas from #1521 and #2175.

It's currently a work in progress, but I'm looking to get some feedback.

### Rationale for using a Django setting

A setting allows the developer to define all their editor configuration in a single place which is good for consistency across the admin, especially if third party apps are being used. Also this problem is very similar to Django's recent switch to multiple template engines, which was solved using a setting.

**Why not pass configuration straight into the RichTextField/Block?**

See @gasman's two comments on https://github.com/torchbox/wagtail/pull/2175#issuecomment-191243017

tl;dr: passing the configuration in the model definition requires that the widget and its configuration to be deconstructed into migrations. This is both tricky to do and it doesn't make much sense to tightly couple database fields to form widget configuration.
 
### How would this work?

A "rich text editor" is made up of a widget class and a set of options to pass in to it. Multiple rich text editors can exist in the same project, each with a different name.

Here's an example configuration:

```python
#settings.py

WAGTAILADMIN_RICH_TEXT_EDITORS = {
    'default': {
        'WIDGET': 'wagtail.wagtailadmin.rich_text.HalloRichTextArea',
        'OPTIONS': {
            # editor-specific options here
        }
    },
    'restricted': {
        'WIDGET': 'wagtail.wagtailadmin.rich_text.HalloRichTextArea',
        'OPTIONS': {
            # Only allow bold and italics
            'features': ['b', 'i']
        }
    },
}
```

Editors can be configured on a per-field basis by passing the editor name as a keyword argument to the ``RichTextField``/``RichTextBlock``.

```python
# models.py

class MyPage(Page):
    intro = RichTextField(editor='restricted')  # Only allow limited formatting options (see configuration above)
    body = RichTextField()
```

The ``WIDGET`` option would make it possible to swap out Hallo for a completely different editor library. Imagine we create a third-party module that provides ck editor, this is an example of how we could switch the project to use it globally:

```python
# settings.py

WAGTAILADMIN_RICH_TEXT_EDITORS = {
    'default': {
        'WIDGET': 'wagtailckeditor.CKRichTextArea',
        'OPTIONS': {
            # ck editor specific options here
        }
    },
}
```


tomdyson:
This looks promising @kaedroho. Would it make it possible to use https://github.com/isotoma/wagtailtinymce? What about #2409 - where does that fit in?

kaedroho:
> Would it make it possible to use wagtailtinymce

Yep, but will probably require some changes to the ``wagtailtinymce`` module

> What about #2409 - where does that fit in?

It solves the first point on that ticket by making the rich text editor swappable. We can move away from Hallo.js but those who want it can use this mechanism to get it back.

SalahAdDin:
Looks like django-ckeditor, jajajajajaj :+1:

kaedroho:
Rebased on master and removed the ability to reconfigure the editor with the ``OPTIONS`` setting. This PR just implements the ability to change the rich text widgets now.

kaedroho:
Rebased

wagtail/wagtailadmin/rich_text.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+DEFAULT_RICH_TEXT_EDITORS = {
+    'default': {
+        'WIDGET': 'wagtail.wagtailadmin.rich_text.HalloRichTextArea'
+    }
+}
+
+
+def get_rich_text_editor(name='default'):
```
m1kola: I think `get_rich_text_editor_widget` is more appropriate name for this function. I think, we can extend `WAGTAILADMIN_RICH_TEXT_EDITORS` setting in future (for whitelisters, for example)

m1kola:
Rebased + fix for styleguide in a30833e9e2048cb083f3d26c2c8e60a891bc055e.

wagtail/wagtailcore/blocks/field_block.py
```diff
@@ -325,16 +326,16 @@ def get_prep_value(self, value):
 
     @cached_property
     def field(self):
-        from wagtail.wagtailcore.fields import RichTextArea
-        return forms.CharField(widget=RichTextArea, **self.field_options)
+        from wagtail.wagtailadmin.rich_text import get_rich_text_editor
+        return forms.CharField(widget=get_rich_text_editor(), **self.field_options)
```
m1kola: Should be `get_rich_text_editor(self.editor)`.

m1kola:
Rebased + onflict with #2417 fixed + https://github.com/torchbox/wagtail/pull/2416#discussion_r62875630 fixed.

m1kola:
I have tested it with TinyMCE and CKEditor. Both work perfect with `RichTextField` and with `RichTextBlock` (after 57e476641780feedcf6fa05f8f8f5e71fd1ea92d).

I found only one strange thing: when I have CKEditor and any other editor on same page I'm getting this:
screen shot 2016-05-11 at 18 35 02

But probably it's a bug / feature of CKEditor.

m1kola:
> I found only one strange thing: when I have CKEditor and any other editor on same page I'm getting this:

[Here](http://docs.ckeditor.com/#!/api/CKEDITOR-cfg-disableAutoInline) is the reason. Ignore it.

@kaedroho I'm happy with this PR (except a [naming question](https://github.com/torchbox/wagtail/pull/2416#discussion_r62848352)). I think we need tests and it will be ready for merge.

Thanks!

m1kola:
Probably we should transfer params from `WAGTAILADMIN_RICH_TEXT_EDITORS` to widget like [Django&nbsp;does](https://github.com/django/django/blob/1.9.6/django/template/utils.py#L101) with `TEMPLATES` setting and template engines. But I don't think that we need it in first iteration.

wagtail/wagtailadmin/tests/test_rich_text.py
```diff
@@ -0,0 +1,194 @@
...
+
+        self.login()
+
+        # Simulate the absence of a setting
+        if hasattr(settings, 'WAGTAILADMIN_RICH_TEXT_EDITORS'):
+            del settings.WAGTAILADMIN_RICH_TEXT_EDITORS
+
+    def tearDown(self):
+        from wagtail.tests.testapp.models import DefaultRichBlockFieldPage
+        from wagtail.tests.testapp.models import DefaultRichTextFieldPage
```
m1kola: I have spent lot of time to figure out why some tests fail when I run test case, but I still don't know how to fix it (how to reset some property caches).

Looks like part of problem in [this cache](https://github.com/torchbox/wagtail/blob/v1.4.4/wagtail/wagtailadmin/edit_handlers.py#L290) and in [this property](https://github.com/torchbox/wagtail/pull/2416/files#diff-de5efc8fde48bf2b10e49731bab27710R327)... Probably somewhere else.

@kaedroho, do you have any idea how to test editor with different settings?

timheap: You can clear the cached `get_form_class()` by setting `cls._form_class = None`. This will not clear the cached field though. '@cached_property`s can be cleared by `del`eting them: find the `RichTextBlock` instance in question and `del block.field`.

m1kola: This works, but looks horrible:

```python
    def setUp(self):
        # Find root page
        self.root_page = Page.objects.get(id=2)

        self.login()

        # Simulate the absence of a setting
        if hasattr(settings, 'WAGTAILADMIN_RICH_TEXT_EDITORS'):
            del settings.WAGTAILADMIN_RICH_TEXT_EDITORS

        from wagtail.tests.testapp.models import DefaultRichBlockFieldPage
        from wagtail.tests.testapp.models import DefaultRichTextFieldPage

        edit_handler = DefaultRichBlockFieldPage.get_edit_handler()
        if edit_handler._form_class:
            rich_text_block = edit_handler._form_class.base_fields['body'].block.child_blocks['rich_text']
            if hasattr(rich_text_block, 'field'):
                del rich_text_block.field
        edit_handler._form_class = None

        edit_handler = DefaultRichTextFieldPage.get_edit_handler()
        edit_handler._form_class = None
```

wagtail/wagtailadmin/rich_text.py
```diff
@@ -31,6 +32,15 @@ def value_from_datadict(self, data, files, name):
...
 
+    @property
+    def media(self):
+        return Media(js=[
+            static('wagtailadmin/js/vendor/hallo.js'),
+            static('wagtailadmin/js/hallo-plugins/hallo-wagtaillink.js'),
+            static('wagtailadmin/js/hallo-plugins/hallo-hr.js'),
+            static('wagtailadmin/js/hallo-plugins/hallo-requireparagraphs.js'),
+        ])
+
```
gasman: You don't need to wrap these in `static(...)` - Django will do this automatically for any non-absolute paths (see https://docs.djangoproject.com/en/1.9/topics/forms/media/#paths-in-asset-definitions, https://github.com/django/django/blob/86573861a95e5a47dc7ff906443117d75b73dca1/django/forms/widgets.py#L91)

kaedroho: It is slightly different on Django 1.8 though, where it appends ``STATIC_URL`` to the beginning: https://github.com/django/django/blob/stable/1.8.x/django/forms/widgets.py#L73

gasman: Ah. Fair enough then!

gasman:
All looks good to me! (I don't think the settings hacks in the tests are a problem - hopefully we won't have to touch that code again in future, so it doesn't have to be pretty :-) )

Any thoughts about what needs to be done about documentation at this point? I guess right now this change is only of interest to people building their own alternative rich text widget, so we can get away without it for now - there's no need for any end-user (i.e. site implementer) facing documentation until a fully-functional alternative widget exists.

coredumperror:
Seeing as how I'm going to need this feature in the near future, since Hallo.js is insufficient for my company's needs, I'd like there to at least be some sort of cursory documentation. Just a few lines on what code to look at in order to learn how to write a custom widget would probably be sufficient.

SalahAdDin:
Documentation plz!

gasman:
Merged in 197527c5d2ae767e0ef11e5296d2bfb44f3cdd58 + parents.

kaedroho:
Thanks @m1kola and @gasman!

riceyrice:
Do you still expect to add support for the `OPTIONS` key - e.g. passed as a `options` argument to the widget constructor if present in the setting?

gasman:
@riceyrice I'm currently undecided on that. As I see it, the primary (only?) motivation for passing options to the editor is to control which HTML elements (images, headings...) are allowed in the field and show/hide toolbar buttons accordingly. If we're going to make that configurable, I think we ought to do it in a standardised, non-editor-specific way (possibly something like https://github.com/torchbox/wagtail/pull/2175#issuecomment-194838184) so that our HTML whitelisting logic can also work from the same list of allowed elements - and also so that users aren't "locked in" to a particular editor if we change the default one in future.

However, I don't want to block things holding out for a big over-engineered solution if there's a simple and effective way that addresses people's needs now. What do you think?

riceyrice:
@gasman Right now I agree that's the only motivation I can think of. I'm wary of trying to be too generic as there will inevitably be a point where you need to go editor-specific and the interaction between the two could be unpleasant. Providing editor-specific options as well as classname in WAGTAILADMIN_RICH_TEXT_EDITORS keyed by the editor parameter given in the RichTextField/Block, while providing generic options in a features parameter in the RichTextField/Block in future could be a reasonable way of dividing things? Just need to clarify how these are delivered to the widget.

Currently wagtailtinymce makes use of hooks writing to JavaScript global variables to set editor-specific options and we were unable to allow different options in different editors. Now we have this new editor switching mechanism I'm looking to replace that with something more flexible and less fragile. The editor-specific options could also be achieved e.g. by making your own subclasses of a given widget with default parameters baked in in the meantime and registering each in WAGTAILADMIN_RICH_TEXT_EDITORS which is probably the route I'll take for now.

jaydensmith:
I have created a module for the Froala WYSIWYG. Still unfinished but I required simple functionality for a project. https://github.com/jaydensmith/wagtailfroala

PR #2418, opened 2016-03-29
Import absolute_import, unicode_literals in all files

timheap:
What is says up there. Also made this automatic through `isort`. If you run `isort` on your file, it will automatically add the imports while sorting. Couple this with [editor integration](https://github.com/timothycrosley/isort/wiki/isort-Plugins), and you will never forget this again. Drone will also check for these imports as part of its run.

As part of these changes, some strings that were automatically the correct type on all versions had to be explicitly set to the correct type. This was mostly generated class and function names. In both versions of Python, these have to be `str`s - bytes on Python 2, unicode on Python 3. This is achieved by wrapping the names in `str()`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -4,3 +4,5 @@ universal = 1
 [isort]
 line_length=100
 multi_line_output=4
+skip=migrations
+add_imports=from __future__ import absolute_import,from __future__ import unicode_literals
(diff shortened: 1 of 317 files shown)


timheap:
The isort issue should be fixed now. Can you restart the drone build for this?

m1kola:
Looks good to me. Thanks!

@timheap, I didn't add this feature to changelog\release notes because it's not end-user feature. If you think that we should add this to docs, let me know.

Merged in a0af3d6685b046acf4d46de5353c293d54371644.

timheap:
No release note is fine, I agree with your reasoning. Thanks for merging!

PR #2421, opened 2016-03-30
Fix isort on drone

timheap:
isort 4.2.3+ is broken on Drone due to timothycrosley/isort#414. This commit pins isort to 4.2.2, as well as fixing an import order error that snuck in.


timheap:
Hold up, isort has a new release that apparently fixes this issue. Drone is running as I type. This may not be required.

timheap:
OK, this PR is no longer required. The newest version of isort fixes the bug. Whether we want to pin isort to a fixed version at some point is a discussion worth having though.

PR #2423, opened 2016-03-31
Add options to filter deprecation warnings in tests

timheap:
The test output was extremely noisy with deprecation warnings. Half of these warnings were caused by external modules like django-taggit, and could be ignored. The warnings caused by Wagtail got lost in the noise though, rendering the test output useless for finding warnings. By default, the tests now only print deprecation warnings if raised by Wagtail.

Use `./runtests.py --deprecation=all` if you want all the warnings to be printed. This is useful when we want to fix third party packages before they break from being too outdated.

Use `./runtests.py --deprecation=pending` if you want the default behaviour.

Use `./runtests.py --deprecation=imminent` if you only want imminent DeprecationWarnings, ignoring PendingDeprecationWarnings.

Use `./runtests.py --deprecation=none` if you want to ignore all deprecation warnings.

This changes how `./runtests.py` passes arguments to `dj test`. Previously, you could intermingle arguments to `./runtests.py` and `dj test`, and `./runtests.py` would sort things out. Now, arguments to `dj test` must come at the end, after a `--`. For example:

```
$ ./runtests.py --deprecation=all -- --keepdb
```

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -4,6 +4,7 @@ import sys
 import os
 import shutil
 import warnings
+import argparse
 
 from django.core.management import execute_from_command_line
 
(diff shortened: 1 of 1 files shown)


m1kola:
When `--deprecation` is set to `none` or `imminent` the following warning appears:
```
/Users/m1kola/.virtualenvs/wagtail_dev/lib/python2.7/site-packages/taggit/managers.py:87: RemovedInDjango110Warning: Calling a SQLCompiler directly is deprecated. Call compiler.quote_name_unless_alias instead.
```

Probably this is because of name of virtualenv. I tried to create virtualenv with name `tests` and didn't get this warning.

Also `wagtail.tests.test_utils.TestThisShouldBeAList.test_add_a_tuple` fails with following traceback (with `--deprecation=imminent` or  `--deprecation=none`):
```
======================================================================
FAIL: test_add_a_tuple (wagtail.tests.test_utils.TestThisShouldBeAList)
  File "/Users/m1kola/Projects/wagtail/wagtail/tests/test_utils.py", line 34, in test_add_a_tuple
    self.assertEqual(len(w), 1)
AssertionError: 0 != 1
```

Looks like we need to reset filters before run `wagtail.tests.test_utils.TestThisShouldBeAList`.

Other than the above comments, this looks good. Thanks @timheap!

gasman:
When I run the tests with django-taggit 0.18.0 installed, I get the SQLCompiler warning regardless of whether my virtualenv has 'wagtail' in the name. @m1kola, perhaps your 'tests' virtualenv has django-taggit 0.18.1 installed (which includes Tim's fixes for the deprecation warnings)?

I don't get the test_add_a_tuple failure at all - and since the taggit warning only affects an old version, I think this is good to merge.

gasman:
Merged in 0097f770fd939cf74ae5df7af8e6134bf7f3ce0c

m1kola:
Yes, I tried run with `django-taggit==0.18.0`.

`test_add_a_tuple` works correctly because Tim has fixed this issues in other PR (see a48ae09ce927622e8a5931dbcb843523d8f4bd23).

PR #2424, opened 2016-03-31
Check POST requests method correctly

timheap:
Instead of using `if request.POST:`, use `if request.method == 'POST':`. Workarounds due to incorrect method checks in the tests have also been removed.


kaedroho:
Merged in c0381859156931ed78997951a35816a61a906e16 +parent
Thanks!

PR #2426, opened 2016-04-01
Add a check that the declared search_fields exist

jschneier:
I got bitten by this when I was developing so thought I'd add a system check. Is this something you're interested in taking?

### DIFF
--- a/wagtail/wagtailsearch/index.py
+++ b/wagtail/wagtailsearch/index.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, unicode_literals
 
 from django.apps import apps
+from django.core import checks
 from django.db import models
 from django.db.models.fields import FieldDoesNotExist
 from django.db.models.fields.related import ForeignObjectRel, OneToOneRel, RelatedField
@@ -78,6 +79,34 @@ class Indexed(object):
         to return the instance in its most specific class so it reindexes properly.
         """
         return self
+
+    @classmethod
+    def check(cls, **kwargs):
+        errors = super(Indexed, cls).check(**kwargs)
+        errors.extend(cls._check_fields_exist(**kwargs))
+        return errors
+
+    @classmethod
+    def _has_field(cls, name):
+        try:
+            cls._meta.get_field(name)
+            return True
+        except models.fields.FieldDoesNotExist:
+            return callable(getattr(cls, name, None))
+
+    @classmethod
+    def _check_fields_exist(cls, **kwargs):
+        errors = []
+        for field in cls.get_search_fields():
+            message = '{name} in search_fields but not on {model}'
+            if not cls._has_field(field.field_name):
+                errors.append(
+                    checks.Warning(
+                        message.format(name=field.field_name, model=cls.__name__),
+                        obj=cls,
+                    )
+                )
+        return errors
 
     search_fields = SearchFieldsShouldBeAList([], name='search_fields on Indexed subclasses')
 
--- a/wagtail/wagtailsearch/index.py
+++ b/wagtail/wagtailsearch/index.py
@@ -91,11 +91,11 @@ class Indexed(object):
     @classmethod
     def check(cls, **kwargs):
         errors = super(Indexed, cls).check(**kwargs)
-        errors.extend(cls._check_fields_exist(**kwargs))
+        errors.extend(cls._check_search_fields(**kwargs))
         return errors
 
     @classmethod
-    def _check_fields_exist(cls, **kwargs):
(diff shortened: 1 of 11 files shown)


jschneier:
We'll need to change the model definitions to put `index.Indexed` at the beginning as well. This is considered best practices for mixins I believe.

jschneier:
I'm also unclear if I need to do more involved checks for `index.RelatedField`.

jschneier:
I guess this PR also assumes that the mixin has been added to a model. Maybe should catch the `AttributeError` if a non-model is subclassed? Although I don't think the `super check` thing will work then.

wagtail/wagtailsearch/index.py
```diff
@@ -77,6 +78,34 @@ def get_indexed_instance(self):
...
+            return hasattr(cls, name)
+
+    @classmethod
+    def check(cls, **kwargs):
+        errors = super(Indexed, cls).check(**kwargs)
+        errors.extend(cls._check_fields_exist(**kwargs))
+        return errors
+
+    @classmethod
+    def _check_fields_exist(cls, **kwargs):
```
kaedroho: Could you rename this to ``_check_search_fields``?

wagtail/wagtailsearch/index.py
```diff
@@ -77,6 +78,34 @@ def get_indexed_instance(self):
...
+        errors = []
+        for field in cls.get_search_fields():
+            message = '{name} in search_fields but not on {model}'
+            if not cls._has_field(field.field_name):
```
kaedroho: This might be neater to implement on the ``BaseField`` class below as that class is responsible to doing the actual lookup. Could we move this to a new method on ``BaseField``? (eg ``BaseField.check(model)``)

wagtail/wagtailsearch/index.py
```diff
@@ -77,6 +78,34 @@ def get_indexed_instance(self):
...
+    def check(cls, **kwargs):
+        errors = super(Indexed, cls).check(**kwargs)
+        errors.extend(cls._check_fields_exist(**kwargs))
+        return errors
+
+    @classmethod
+    def _check_fields_exist(cls, **kwargs):
+        errors = []
+        for field in cls.get_search_fields():
+            message = '{name} in search_fields but not on {model}'
```
kaedroho: Maybe reword this message slightly? ``{model}.search_fields contains field '{name}' but it doesn't exist``

kaedroho:
> I got bitten by this when I was developing so thought I'd add a system check. Is this something you're interested in taking?

Yep, definitely! Thanks for your work on this.

>  I'm also unclear if I need to do more involved checks for index.RelatedField.

It would be useful to have that there too. The ``RelatedFields`` class would need to check that each specified field exists on the related class as well.

The following should give an error:

```python

class Author(models.Model);
    name = models.CharField()

class Book(index.Indexed, models.Model):
    title = models.CharField()
    author = models.ForeignKey(Author)

    search_fields = [
        index.SearchField('title'),
        index.RelatedFields('author', [
            index.SearchField('name'),
            index.SearchField('bio'),   # Oops, bio doesn't exist on Author
        ]),
    ]
```

jschneier:
Was there discussion around `index.Indexed` not being a `models.Model` subclass with `abstract = True`, because I am making a fair amount of assumptions about it being on a model in this PR I think. There are other things we can do like adding the check to a `checks.py` and registering it in that case.

jschneier:
Given `.get_indexed_objects` I think the assumption is safe.

jschneier:
@kaedroho can you have embedded `index.RelatedField` fields?

kaedroho:
Yeah, that is allowed at the moment in Wagtail but I'm not sure if Elasticsearch supports it (their docs don't mention it and ES seems to just silently ignore anything that has a nested related field).

gasman:
Does this correctly handle `SearchField`s on callables / other non-fields (http://docs.wagtail.io/en/v1.4.3/topics/search/indexing.html#indexing-callables-and-other-attributes)?

jschneier:
Yes. It does not yet handle warning on Related Fields though, I need to find some time.

kaedroho:
Thanks for your work on this so far @jschneier!

I've merged 720998c0cddacf9a4071b0c0e5952c7e17d8ce24 into master but I'd like to wait for the ``RelatedFields`` support before merging the rest in so bumping this PR to 1.6.

gasman:
Have merged this (in 1dad100695060435b597545cd61c3ee420afee9f) as agreed with @kaedroho, as `RelatedFields` can be worked on separately. Thanks!

PR #2427, opened 2016-04-01
Fix site settings docs for setting an icon

toumaz:
The "icon-" prefix is automatically added in SettingMenuItem.
Using "icon-placeholder" as suggested would thus result in the CSS class "icon-icon-placeholder".

This is my first time doing a PR, so please let me know if I've messed something up, but this seemed like low hanging fruit. :)


tomdyson:
Congratulations on your first PR, @toumaz! This looks good.

alexgleason:
Thanks @toumaz, this is correct! Thanks for spotting and submitting a PR.

PR #2436, opened 2016-04-03
Remove TagSearchable class

kaedroho:
We've been slowly chipping away at this class over the last few releases. It now only contains a bit of search configuration for fields it doesn't even have and a "popular_tags" class method.

I think it's much neater to just define those things on the image and document models.

### DIFF
--- a/wagtail/wagtailadmin/taggable.py
+++ b/wagtail/wagtailadmin/taggable.py
@@ -1,10 +1,12 @@
 from __future__ import absolute_import, unicode_literals
 
+import warnings
+
 from django.contrib.contenttypes.models import ContentType
 from django.db.models import Count
 from taggit.models import Tag
 
-from wagtail.utils.deprecation import SearchFieldsShouldBeAList
+from wagtail.utils.deprecation import RemovedInWagtail17Warning, SearchFieldsShouldBeAList
 from wagtail.wagtailsearch import index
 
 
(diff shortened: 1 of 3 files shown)


gasman:
`TagSearchable`'s purpose has got a bit muddled over time and I'd be happy to see the back of it - however, duplicating the code across wagtailimages and wagtaildocs seems like a bit of a step backwards, going against our eventual goal of moving our common model 'traits' (tagging, versioning, translation, moderation...) into mixins so that users can construct their own image/document-like models by combining them in different ways.

It seems to me that we could make `TagSearchable` into something better-focused and more useful, by:

* renaming it to 'Taggable'
* making it an abstract model that defines the `tags = TaggableManager(...)` field (one less thing for the inheriting class to do)
* not defining `title` as a search field (that's not relevant to tagging - it's a hangover from when this was the *only* class that did searching...)
* seeing if we can eliminate `get_indexed_objects` (it looks like the one on `index.Indexed` already supports prefetching related fields - doesn't that handle tag relations too? If not, could it be made to?)

gasman:
OK, #2498 has changed my mind - with `get_indexed_object` eliminated, `TagSearchable` is completely useless :-) I'm still clinging on to the idea of a glorious ModelAdmin future where you can drop a tagging mixin into your model and instantly get tagging functionality across its admin interface - but I don't think `TagSearchable` is a meaningful stepping stone towards that goal any more.

This just leaves the `popular_tags` method to deal with, and I'd like to avoid duplicating it between images and docs. (A duplicated 8-line method is hardly the end of the world, though...) One possibility would be to make it a plain helper function (`popular_tags_for_model(model)`), rather than a class method. It could *almost* be a template tag, except that you can't write `{% popular_tags Document %}` in a template...

gasman:
Superseded by #2705

PR #2439, opened 2016-04-04
Add a title to explorer actions

SableWalnut:
Currently the actions on the explorer menu are short verbs and do not actually describe what the buttons do. Adding a title to these should assist in letting the user know what each button does. Currently one would expect the 'Draft' button would create a draft, having a title on it will explain to the user in a more verbose way what the button actually does (In this case, preview the draft). 

Currently we have a button with the label 'Add a child page'  which describes what the button does, while we also have  'Edit', 'Draft' etc which are short verbs. I'm not sure whether we should be using short verbs as labels, or whether the labels should describe what they do, or a mixture of both. Perhaps we could remove the 'Add child page' in favour of the already existing plus icon to the right of the page status.

cc @davecranwell @mazil

### DIFF
--- a/wagtail/wagtailadmin/widgets.py
+++ b/wagtail/wagtailadmin/widgets.py
@@ -243,9 +243,14 @@ class BaseDropdownMenuButton(Button):
         raise NotImplementedError
 
     def render(self):
+        if 'title' in self.attrs:
+            title = self.attrs['title']
+        else:
+            title = None
         return render_to_string(self.template_name, {
             'buttons': self.get_buttons_in_dropdown(),
             'label': self.label,
+            'title': title,
             'is_parent': self.is_parent})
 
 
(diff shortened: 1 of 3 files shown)


wagtail/wagtailadmin/widgets.py
```diff
@@ -243,9 +243,14 @@ def get_buttons_in_dropdown(self):
         raise NotImplementedError
 
     def render(self):
+        if 'title' in self.attrs:
+            title = self.attrs['title']
+        else:
+            title = None
```
gasman: This could be shortened to: `title = self.attrs.get('title')`

gasman:
Merged in c35c1f360a4800194bfdcd383e3c370bdb8b8cc6 + parents - thanks!

PR #2442, opened 2016-04-04
Fixes regression padding on messages after (partially) moving to a mixin

### LINKED ISSUE #2408: Unnecessary padding for messages
Currently in `master` we have unnecessary padding for messages. I can't reproduce this bug on `stable/1.4.x` branch and on `v1.4.1` tag.

screen shot 2016-03-28 at 14 34 59

jjanssen:
Fixes #2408. While some unordered list declarations were moved to a mixin, there is an explicit override on `.messages`. This caused a small regression on messages. Verified by the styleguide.


gasman:
Merged in 4b4b5fd601849542278ce20724f0eba69eb46acb - thanks!

PR #2445, opened 2016-04-05
Fix/formfielddupes

### LINKED ISSUE #585: No warning is given when two Formbuilder fields have the same name
It is possible to create a Formbuilder form that has two fields with the same name. If a user views this form, the first duplicate field will not appear. If the user submits the form then validation will fail and the page will reload.

tacitus:
fixes #585

### DIFF
--- a/wagtail/wagtailadmin/forms.py
+++ b/wagtail/wagtailadmin/forms.py
@@ -288,23 +288,6 @@ class WagtailAdminPageForm(WagtailAdminModelForm):
 
     def clean(self):
 
-        # Check for dupe form field labels - fixes #585
-        if 'form_fields' in self.formsets:
-            _forms = self.formsets['form_fields'].forms
-            for f in _forms:
-                f.is_valid()
-
-            for i, form in enumerate(_forms):
-                if 'label' in form.changed_data:
-                    label = form.cleaned_data.get('label')
-                    for idx, ff in enumerate(_forms):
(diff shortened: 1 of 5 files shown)


wagtail/wagtailadmin/forms.py
```diff
@@ -287,6 +287,24 @@ def clean_seo_title(self):
             return self.cleaned_data['seo_title'].strip()
 
     def clean(self):
+
+        # Check for dupe form field labels - fixes #585
+        if 'form_fields' in self.formsets:
```
timheap: Instead of changing the form used for all Pages, you can make a custom Wagtail admin form for Form Pages, and tell Wagtail to use that form when validating those page types:

```python
# in wagtail.wagtailforms.forms
class WagtailAdminFormPageForm(WagtailAdminPageForm):
    def clean(self):
        super(WagtailAdminFormPageForm, self).clean()
        # Your tests
```

```python
# in wagtail.wagtailforms.models
from .forms import WagtailAdminFormPageForm
class AbstractForm(Page):
    # ...
    # Use the custom form defined above to validate Form pages
    base_form_class = WagtailAdminFormPageForm
```

tacitus:
Sure, nice and clean :)

gasman:
Merged in f6706977bd06a3b5c7dae5f50e4146b896dbd727 - thanks!

PR #2448, opened 2016-04-05
PageChooserBlock supports limiting of page type

nutztherookie:
Fixed the TODO "allow limiting to specific page types" by adding a class attribute `page_model`


nutztherookie:
darn.. let me retry..

PR #2449, opened 2016-04-05
PageChooserBlock supports limiting of page type

nutztherookie:
Fixed the TODO "allow limiting to specific page types" by adding a class attribute `page_model`

### DIFF
--- a/wagtail/wagtailcore/blocks/field_block.py
+++ b/wagtail/wagtailcore/blocks/field_block.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, unicode_literals
 import datetime
 
 from django import forms
+from django.contrib.contenttypes.models import ContentType
 from django.db.models.fields import BLANK_CHOICE_DASH
 from django.template.loader import render_to_string
 from django.utils import six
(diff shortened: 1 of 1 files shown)


wagtail/wagtailcore/blocks/field_block.py
```diff
@@ -383,18 +384,22 @@ def clean(self, value):
...
     def target_model(self):
-        from wagtail.wagtailcore.models import Page  # TODO: allow limiting to specific page types
-        return Page
+        return self.page_model
 
     @cached_property
     def widget(self):
         from wagtail.wagtailadmin.widgets import AdminPageChooser
+        ct = ContentType.objects.get_for_model(self.page_model)
+        return AdminPageChooser(content_type=ct, can_choose_root=self.can_choose_root)
```
kaedroho: As of 09878fcfd87f5b5ea67dd8d178667bce344c02e0, you can pass the page model straight in using the ``page_models`` attribute:

eg: ``AdminPageChooser(page_models=[self.page_model], can_choose_root=self.can_choose_root)``

wagtail/wagtailcore/blocks/field_block.py
```diff
@@ -383,18 +384,22 @@ def clean(self, value):
...
+
     def __init__(self, can_choose_root=False, **kwargs):
         self.can_choose_root = can_choose_root
         super(PageChooserBlock, self).__init__(**kwargs)
 
     @cached_property
     def target_model(self):
-        from wagtail.wagtailcore.models import Page  # TODO: allow limiting to specific page types
-        return Page
+        return self.page_model
```
kaedroho: Might be slightly simpler to set ``page_model = None`` at the top of this class and do ``return self.page_model or Page``. That way you don't need to move the import either

kaedroho:
Thanks @nutztherookie!

Could we change ``page_model`` into an array of models? (as you can specify multiple page models to the ``AdminPageChooser``` widget)

gasman:
Please could you add some tests for the new parameter to https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailcore/tests/test_blocks.py#L1512?

nutztherookie:
good calls! I will get on that asap.

nutztherookie:
@kaedroho I'm not exactly sure what the `def target_model` is for. So: What should it return when I actually have a list? 

    return self.target_models or Page # ? where target_models might be a list?

nutztherookie:
reviving this ticket,

PR #2451, opened 2016-04-06
Add wagtail.contrib.settings.jinja2tags

timheap:
Add jinja2 support for `wagtail.contrib.settings`.

The code for this got more involved than I was hoping it would, but due to the differences between the Django and Jinja template engines, a direct port was not possible. Jinja2 contexts are immutable, and extensions are loaded once at start up. Because of this, it is not possible to store anything in the template context, as it is done with the Django templates and the context processor. Thus, a WeakKeyDictionary keyed on the Jinja template context instance stores the cached settings for that context. The other classes are just a slightly different implementation of the `SettingsProxy`/`SettingsModuleProxy` classes from the Django context processor for this different environment.

Docs have been updated as well.


kaedroho:

PR #2453, opened 2016-04-06
Recognise instagram embed URLs with www prefix

gasman:
Fixes https://github.com/torchbox/wagtail/issues/1134#issuecomment-204062904


kaedroho:
Thanks!

PR #2454, opened 2016-04-07
Add auto increment slug to admin

nek4life:
In short I would like to be able to use auto incrementing page slugs in the case of a duplicate slug within the current parent page.

The use case is an event calendar where I don't really care so much about the slugs, but there is a high potential for duplicates with the types of events that may be submitted.

The reason I added this feature as a setting is to preserve the current behavior while allowing new behavior. I'm also up for removing the setting and just allowing duplicate page slugs to auto increment as the default behavior as well.

I know this patch needs tests and docs, but I wanted to check the temperature for adding this feature before proceeding.

I can also open an issue with a link to this PR if that is a better place for discussion.

Thanks,
Charlie


wagtail/wagtailcore/models.py
```diff
@@ -391,6 +391,15 @@ def _get_autogenerated_slug(self, base_slug):
...
+        try:
+            slug_parts[-1] = str(int(slug_parts[-1]) + 1)
+            candidate_slug = '-'.join(slug_parts)
+        except ValueError:
+            # If the last part of the slug fails to be converted
+            # to an int then simply move on to auto incrementing
+            pass
 
         while not Page._slug_is_available(candidate_slug, parent_page, self):
             # try with incrementing suffix until we find a slug which is available
```
nek4life: I didn't touch the original logic here so I didn't break anything, but by the looks of it the code simply appends -2 to every slug even if it ends in an integer. So if for instance you have 'some-slug-5' and you have a duplicate you end up with 'some-slug-5-2' instead of 'some-slug-6'

gasman: I think that would always be the desired behaviour for numbers that the user has explicitly entered as part of the slug - if there's an existing page called 'the-magnificent-7', a new page with the same slug should be turned into 'the-magnificent-7-2', not 'the-magnificent-8' :-) (On the other hand, if the slug they entered was 'the-magnificent', then it *would* go through 'the-magnificent-2', 'the-magnificent-3' etc...)

We do of course want to avoid the situation where a slug that's been given an automatic number suffix gets another one added later on - but I can't see any scenario where this would happen.

nek4life: Ok so I understand the scenario you have described above, but I'm not sure if there is a way to tell the intention behind what the user entered. Would it make more sense to have this setting applied on the model instead of a global setting?

The root problem is that for certain page types there may be a high chance of name collisions. Perhaps there's a different solution. Ideally for something like an events calendar I would normally construct a url like some-event-name-here-1234 or /1243/some-event-name or something where the ID does the lookup and the slug is actually just there for SEO purposes. In addition when the slug does not match a 301 redirect is performed to the correct version so if a user just entered something like hello-wrong-name-1234 a redirect would take them to some-event-name-1234.

Ultimately I'm trying to provide a smooth user experience by reducing the number of unnecessary form errors. But I do see your point that there is a scenario where a number might be intentional and shouldn't be incremented.

I've thought about using a snippet instead, however then you lose the ability to have the wagtail userbar to edit the snippets easily as well as panels and potentially everything else that goes with a page.

m1kola:
One more thing: Wagtail generates a slug automatically using urlify.js if `django.contrib.admin` is installed or just replaces "wrong" chars using regexp. See [page-editor.js](https://github.com/torchbox/wagtail/blob/v1.4.3/wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js#L294) for details.

nek4life:
Thinking more about this I feel like this should be setup on a per model basis and not at the global site preference level. I'll try to dig back in and figure out if there is a way to do this and have the forms respect the configuration.

Thoughts?

m1kola:
@nek4life, I agree. Would be better to configure this on a per model basis. For example, I want to be able to setup increment for `EventPage` model, but for other models I want to have default logic.

I think [this](http://docs.wagtail.io/en/latest/advanced_topics/customisation/page_editing_interface.html#customising-generated-forms) section of documentation would be useful. This documentation for Wagtail 1.5 (not released yet), but, if I'm not mistaken, this feature was introduced in 1.4.

gasman:
This proposed behaviour has come up again independently in #3477. We've now discussed this among the core team - while auto-rewriting of slugs is a behaviour we'd want to avoid in the default case (because the ability to generate 'nice' URLs is a positive selling point of Wagtail that we don't want to undermine...) we're agreed that this would be a worthwhile feature to support on a per-model basis.

lb-:
I know this is old but just digging through issues and I think we should close this and I have raise a new enhancement issue instead https://github.com/wagtail/wagtail/issues/10147

Thanks for the original work on this @nek4life  - hopefully someone (or yourself :) ) can revive this feature with a new PR in the future.

PR #2455, opened 2016-04-07
Unpublish for descendant pages

jordij:
Some of our clients often demand to automatically un-publish all descendant pages of a given page when un-publishing it. I've added a checkbox to the confirmation un-publish template in order to do that. It checks for permissions and only un-publishes what the user is allowed to. Added a few tests as well.


wagtail/wagtailadmin/views/pages.py
```diff
@@ -588,16 +588,27 @@ def unpublish(request, page_id):
...
+                if descendant.permissions_for_user(request.user).can_unpublish():
+                    descendant.unpublish()
+
         messages.success(request, _("Page '{0}' unpublished.").format(page.title), buttons=[
             messages.button(reverse('wagtailadmin_pages:edit', args=(page.id,)), _('Edit'))
         ])
 
         return redirect('wagtailadmin_explore', page.get_parent().id)
 
+    child_pages = page.get_descendants().live()
```
kaedroho: This variable name should be called ``live_descendant_pages``

wagtail/wagtailadmin/views/pages.py
```diff
@@ -588,16 +588,27 @@ def unpublish(request, page_id):
         raise PermissionDenied
 
     if request.method == 'POST':
+        include_descendants = request.POST.get("include_descendants", False)
+
         page.unpublish()
 
+        if include_descendants == 'True':
+            descendant_pages = page.get_descendants().live()
```
kaedroho: Like below, this variable name should be called ``live_descendant_pages``

wagtail/wagtailadmin/views/pages.py
```diff
@@ -588,16 +588,27 @@ def unpublish(request, page_id):
...
+            for descendant in descendant_pages:
+                descendant = descendant.specific
+                if descendant.permissions_for_user(request.user).can_unpublish():
```
kaedroho: As permissions in Wagtail can't be taken away as you go deeper in the tree, this check should never fail

kaedroho:
Thanks Jordi! The code looks good to me, only a few minor comments above

wagtail/wagtailadmin/views/pages.py
gasman: > As permissions in Wagtail can't be taken away as you go deeper in the tree, this check should never fail. Might be worth removing it

While this is true, I think we should keep all "knowledge" about permission logic contained within the permission_policy module, so it's a good thing for this code to perform the check rather than rely on that fact. It's entirely possible that we might make changes to the permission model in future, or make it configurable.

kaedroho: Ok, agreed!

jordij:
Cheers @kaedroho I've renamed the vars and fixed the conditional for the checkbox value. Should be good to go, let me know if you need something else for this one 👌

kaedroho:
I've rebased this branch and resubmitted it at #2562

PR #2456, opened 2016-04-07
Fix incorrect bases for Meta classes on multi-inherited Blocks

timheap:
Constructing the `Meta` class for a Block did not properly take in to account multi-inheritance, looking only at the left-most parent class for a `_meta_class` to inherit from. This PR fixes this by constructing a `Meta` class that takes in to account all of the base classes of the Block.

### DIFF
--- a/wagtail/wagtailcore/blocks/base.py
+++ b/wagtail/wagtailcore/blocks/base.py
@@ -17,8 +17,6 @@ from django.utils.text import capfirst
 # calls force_text, which would cause it to lose its 'safe' flag
 
 
-
-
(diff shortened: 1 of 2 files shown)


gasman:
Merged in 14398bfdd7e0f430cd82c6f820bdaf0a7da53fc4 - thanks!

tomdyson:
Should this go in 1.4.x too?

gasman:
I guess it could! This seems kind of borderline between a bugfix and a feature, in that it's making it possible to do something that you might expect to work, but wasn't officially documented. No strong feelings either way :-)

gasman:
Merged into 1.4.x in aaefb81b608c4b144b04987173184f6c4a98770e.

PR #2458, opened 2016-04-07
Update models.py

l0rn:
There are some more templating languages in modern web development ;) - maybe we should put this even more generic, is there an actual need for validate file suffix?

Cheers


gasman:
Hi @l0rn,

This regexp is not used for validation, but rather to convert my_form.html to my_form_landing.html - see https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailforms/models.py#L141-L142. If we're making this change, I guess we'd need to update the `landing_page_template` line too.

(see also: #1922)

gasman:
Hi @l0rn - I see you've added the built static files in this PR. We deliberately keep these out of the git repository (as they make it difficult to track changes and resolve merge conficts), and won't be able to merge the PR in its current form. When this is ready for us to look at again, please can you open a new pull request containing just the minimum necessary changes? Thanks!

PR #2459, opened 2016-04-07
image serve view refactor for better DRY override

ychab:
Should not introduce any regression and we could easily override action methods. Concretly, I need it to just override ``serve`` action to use ``sendfile`` method.

BTW, it should be nice if we can also add an another view which inherit from it and just override the ``serve`` method action to use the ``sendfile`` methods (maybe from available wrapper in wagtail (at ``wagtail.utils.sendfile.sendfile``) as a fallback). Anyway, I don't think I'm the only one who's interested? I can also make the pull request once this one would be merged. In fact, it would be just something like : 

````python
class SendFileView(ServeView):
    def serve(self, rendition):
        return sendfile(self.request, rendition.file.path)
````

Thanks for review!

### DIFF
--- a/wagtail/tests/dummy_sendfile_backend.py
+++ b/wagtail/tests/dummy_sendfile_backend.py
@@ -1,0 +1,10 @@
+from __future__ import unicode_literals
+
+from django.http import HttpResponse
+
+
+def sendfile(request, filename, **kwargs):
+    """
+    Dummy sendfile backend implementation.
+    """
+    return HttpResponse('Dummy backend response')
--- a/wagtail/wagtailimages/tests/tests.py
+++ b/wagtail/wagtailimages/tests/tests.py
@@ -1,11 +1,12 @@
 from __future__ import absolute_import, unicode_literals
 
 import os
+import unittest
 
 from django import forms, template
 from django.core.exceptions import ImproperlyConfigured
 from django.core.urlresolvers import reverse
-from django.test import TestCase
+from django.test import TestCase, override_settings
 from django.utils import six
 from mock import MagicMock
 from taggit.forms import TagField, TagWidget
@@ -20,6 +21,12 @@ from wagtail.wagtailimages.rect import Rect, Vector
 from wagtail.wagtailimages.views.serve import ServeView, generate_signature, verify_signature
 
 from .utils import Image, get_test_image_file
+
+try:
+    import sendfile  # noqa
+    sendfile_mod = True
+except:
+    sendfile_mod = False
 
 
 class TestImageTag(TestCase):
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by kaedroho** [comment]
Thanks @ychab! This looks good to me

Yep, a PR for implementing sendfile support here would be very welcome!

**Comment by ychab** [comment]
@kaedroho , done. It use the wagtail ``sendfile`` wrapper by default. Correct me if I'm wrong but the only difference with the original sendfile script is that it's allow us to override the default backend set (https://github.com/johnsensible/django-sendfile/pull/33)? (an another idea behind this is to have a soft dependency, but concretly, there is nothing about that for now, right)?

File: wagtail/tests/dummy_sendfile_backend.py [hunk 1]
```diff
@@ -0,0 +1,8 @@
+from django.http import HttpResponse
+
+
+def sendfile(request, filename, **kwargs):
+    '''
```
kaedroho: Could you change this to tripple double quotes (https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring)

wagtail/tests/dummy_sendfile_backend.py
```diff
@@ -0,0 +1,8 @@
+from django.http import HttpResponse
```
kaedroho: This file needs to have ``from __future__ import unicode_literals`` at the top

wagtail/tests/dummy_sendfile_backend.py
ychab: Oups, pep8 through flake8 didn't see it ;-)

wagtail/tests/dummy_sendfile_backend.py
ychab: This is just in case (convention) or there is a particular reason? I'm curious because for now, it just returns an ASCII string so it should not be needed, no?

ychab:
@kaedroho : it would be nice if it could be merge for 1.5 release. Can I give a help (writting documentation or something else)?

wagtail/tests/dummy_sendfile_backend.py
kaedroho: Yep, it's a convention. If we forget to add it somewhere that it's needed it could cause issues for Python 2 users. To prevent that from happening we've configured isort (https://github.com/torchbox/wagtail/blob/master/setup.cfg#L8) to make sure it's in every single file.

wagtail/wagtailimages/tests/tests.py
```diff
@@ -354,6 +361,36 @@ def test_get_missing_source_image_file(self):
...
+    def test_sendfile_nobackend(self):
+        signature = generate_signature(self.image.id, 'fill-800x600')
+        response = self.client.get(reverse('wagtailimages_sendfile',
+                                           args=(signature, self.image.id,
+                                                 'fill-800x600')))
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response['Content-Type'], 'image/png')
+
+    @override_settings(SENDFILE_BACKEND='sendfile.backends.development')
```
kaedroho: Does this setting get read anywhere? (I think we would want to use whatever the user has set in this setting as the default backend)

kaedroho:
Code and tests look good to me :+1:

Could you add a section under "advanced configuration" in this doc https://github.com/torchbox/wagtail/blob/master/docs/advanced_topics/images/image_serve_view.rst just describing how to enable this feature?

ychab:
Ok, doc have been updated!

If it would be merge in the 1.5 release, maybe could we add a quick note in https://github.com/torchbox/wagtail/blob/master/docs/releases/1.5.rst#improvements-to-the-image-serve-view section.

kaedroho:
Merged in 58dfea096e006895516e4626b02967a785f056c1 thanks!

PR #2461, opened 2016-04-07
add template tag for dynamic image url

ychab:
Dynamic image view allow us to serve image rendition outside the main request (which is a real performance boost in some cases!! In my use case, it goes from ~400 to ~15 SQL queries!! ).

However, building rendition image url on Python side could be not enough. A common use-case could be do generate url in templates. I'm pretty sure I'm not the only one?

This PR basically copy/paste the example given in the docs : http://docs.wagtail.io/en/latest/advanced_topics/images/image_serve_view.html#generating-dynamic-image-urls-in-python

The util function is just wrapped into a simple template tag so it can still be used directly in Python code.

Hope it could be merge and thanks for review!

---

This change is [Reviewable](https://reviewable.io/reviews/torchbox/wagtail/2461)

### DIFF
--- a/wagtail/wagtailimages/views/serve.py
+++ b/wagtail/wagtailimages/views/serve.py
@@ -8,6 +8,7 @@ from wsgiref.util import FileWrapper
 
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured, PermissionDenied
+from django.core.urlresolvers import reverse
 from django.http import HttpResponse, HttpResponsePermanentRedirect, StreamingHttpResponse
 from django.shortcuts import get_object_or_404
 from django.utils.decorators import classonlymethod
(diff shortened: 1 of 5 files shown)


kaedroho:
Thanks @ychab!

Perhaps it would be nice to make the URL name configurable so a custom serve view can be used?

``{% image_url 'wagtailimages_serve' myimage fill-100x100 %}``

I think this also makes it more clear that it's going to use the serve view rather than generating a rendition.

wagtail/wagtailimages/views/serve.py
```diff
@@ -36,6 +37,15 @@ def verify_signature(signature, image_id, filter_spec, key=None):
     return signature == generate_signature(image_id, filter_spec, key=key)
 
 
+def generate_image_url(image, filter_spec, key=None):
+    signature = generate_signature(image.id, filter_spec, key)
+    url = reverse('wagtailimages_serve',
```
kaedroho: Perhaps it would be useful to make the URL name configurable so a custom serve view can be used?

``{% image_url 'wagtailimages_serve' myimage fill-100x100 %}``

I think this also makes it more clearer that it's going to use the serve view rather than behaving like the ``{% image %}`` tag.

ychab: Agree, done!

ychab:
@kaedroho , I have rebase the master branch to be more up-to-date. It could be practical to also have this PR merged for the 1.5 release since the PR https://github.com/torchbox/wagtail/pull/2459 was merged. Is that should be possible ? At least it needs doc. I was thinking about adding some notes at this place : https://github.com/torchbox/wagtail/blob/master/docs/topics/images.rst#images-embedded-in-rich-text. It sounds good to you?

gasman:
@ychab 1.5 is in feature freeze now, sorry.

zerolab:
Discussed with @gasman and he agrees this can be added. Will submit a new PR with documentation updates. The rest LGTM

m1kola:
Finalised in #4856

Issue #2401, opened 2016-03-25
Limiting Explorer to the current user's "explorable pages"


coredumperror:
Last week, @tomdyson gave me a suggestion for how Wagtail might be enhanced to support completely separated tenants in a multi-tenant Wagtail instance: implement a "View in Admin" Permission. Possibly with a better name... I believe someone suggested "Choose" in one of the issues relating to this concept, but that's not really broad enough.

Applied to Pages, this permission would prevent unprivileged users from:
- seeing the Page in the Explorer or the fly-out menu in the sidebar
- going to the Explorer URL for that Page (e.g. typing `/admin/pages/29/` into the address bar would issue a 403)
- seeing the Page in admin-side search results
- seeing the Page in a PageChooser

Applied to Collections, this permission would prevent unprivileged users from:
- seeing Images or Documents in Choosers 
- uploading media to unprivileged Collections

Users already can't see or upload media in unprivileged Collections on the `/admin/images/` and `/admin/documents/` pages, so I'll hopefully be able to use the code from those pages as inspiration for implementing the View permission.

The purpose of this permission would be to completely isolate users belonging to one Site, so they can't tell that content belonging to any other Site exists (and also can't get to said content even if they guess the URL, since that would be a privacy violation). This is counter to the way that Collections and Pages currently work, though, so I suppose this functionality would have to be made optional, somehow? I'm not really sure how that would be done, though, so suggestions are welcome!

I intend to implement this myself over the coming days, and will be providing a PR for my changes as soon as they're in a presentable state. I opened this issue now, though, to get input from you guys on how best to go about this, and whether any of my current assumptions may be in need of an update to properly understand all the subtleties involved.

Since this will involve changing the choosers, it relates to #1645, which I also plan to enhance by making the new unified Choosers be Permission-aware.

tomdyson:
Thanks, @coredumperror. I thought the 'someone' who suggested "Choose" was you! I still like it.

@timheap, see the reference to #1645.

coredumperror:
I think "Choose" works for the choosers, but this permission affects the Explorer and the admin search as well, so it doesn't seem appropriate.

coredumperror:
I ran into a spot of trouble while upgrading my development system from Yosemite + MacPorts to El Capitan + Homebrew, and thus have not had a chance to actually start on this project until just now. I'll be working on this over the next week or so, and will then integrate this new permission into #1645 as well.

[...]

PR #1645, opened 2015-08-27
[WIP] Unify link choosers

mx-moth:
This is still a work in progress, but the concept works!

My colleague @seb-b and I recently tried to add a News Article chooser for [Wagtail News](https://pypi.python.org/pypi/wagtailnews/), and we found it to be much more difficult than it needed to be. We needed to make a new hallo plugin, even though there were multiple ones already there that did the same basic thing. We needed to add a DB whitelister in a completely different place. We needed to register against many different hooks, and make a variety of magical classes and views. We needed to insert some funky JS snippets to register our plugin and register a URL. Lots of this had already been done in other places, and it felt like we were duplicating a lot of code.

I had a look through it, and there were so many more similarities than there were differences, so I had a go at unifying it all together. This PR is my progress so far.

Internal links, external links, and email links are now all handled in a similar fashion, with DB whitelisters and being registered via hooks. Document links got a similar treatment. There is now a single hook that takes care of setting it all up - whitelisting, adding to the link chooser panel, and other related things. The definition for a chooser is contained to a single, basic class. You can see these classes in the `link_handlers.py` files I have added.

This needs some documentation, and some thoughts on how to make it backwards compatible. Your thoughts on what I have done so far are also welcome.

PR #2463, opened 2016-04-08
Limiting Explorer to the current user's "explorable" Pages

coredumperror:
This PR is tied to #2401.

What I've done here is limit the Explorer and the flyout menu in the sidebar to display only the Pages for which the user has permission to perform some kind of administration (any `GroupPagePermission` except the new one I added called "Choose", which is for the Chooser functionality I haven't written yet). I call these the user's "administrable Pages". In addition to these pages, all the direct ancestors of said pages are also displayed, up to the Closest Common Ancestor. This allows users to navigate through non-administrable Pages to get to ages that they _can_ administer.

In addition, the Explorer root (`/admin/pages/`) for non-superusers now displays the Closest Common Ancestor of the user's administrable Pages, rather than the root page (the root page can potentially be the CCA, though). This prevents users from knowing about any pages which they don't need to see.

As a part of this change, the Explorer will now throw either a 404 or a 403 if the user tries to go to `/admin/pages/<page_id>` for a Page that they shouldn't be allowed to see. It throws 403 for Pages that are on the current Site, and 404 for Pages on other Sites. 404's are thrown because we don't want to reveal the existence of any Pages that belong to other tenant Sites on the same server.

Work that still needs to be done:
1. Update the two broken tests (I added TODOs for these), and add new tests for the new functionality.
2. Add cache clearing code for the times when permitted_paths and required_ancestors might change for a given user. As far as I can tell, this will be when a Group's permissions change, when a User's Group set changes, or when a Page is moved.
3. Possibly change the way the output of `get_administrable_page_paths()` is cached. As it stands, a lot of duplicate data is going to be cached, because the cache keys are differentiated only by `user.id`. All users which have the same Groups will have the same permitted paths, resulting in a lot of duplication. I'd like to cache by the User's list of Group IDs instead, but that would require an extra DB query every time `get_administrable_page_paths()` is called. I'm not sure about the tradeoff between data storage efficiency vs. performance here, as I have very little optimization experience. Not to mention that I'm not sure how to properly clear the cache when the key is built from multiple Group IDs.
4. Update the admin homepage to display the correct Page count, as it currently counts all pages on the server, instead of all administrable pages. Or should it show all pages on the current Site, perhaps? Or maybe don't change it at all?
5. Apply the same "administrable" limitations to Images and Documents in the User's permitted Collections.


coredumperror:
I added the code to clear the cached administrable path lists. Sadly it doesn't appear to be possible to do robust selective cache entry deletion with Django's caching API, so I had to use a boulder where a hammer would have been preferable. But I at least found a way to avoid using a nuke.

I'll be adding tests for this functionality next. So my next push should hopefully pass CI.

coredumperror:
So the tests failed after my second-to-last commit because my code wasn't passing flake8. But now, after fixing the code to make local runs of flake8 pass, it's now saying this:

```
$ flake8 wagtail
wagtail/wagtailcore/models.py:1413:19: F821 undefined name 'reduce'
wagtail/wagtailcore/models.py:1414:20: F821 undefined name 'reduce'
[info] build failed (exit code 1)
```
What the hell? Not only does flake8 not complain about this when I run it locally, but `reduce` is a python builtin function!

coredumperror:
It's weird that running `./runtests.py` isn't catching some of these errors in the code that the Travis CI builder is catching. I got this with the code in the 294fc46 commit:

```
Ran 1747 tests in 80.673s

OK (skipped=43, expected failures=2)
```
Yet Travis found a legit error in that code that `runtests.py` completely missed.

coredumperror:
Ohhh, the `reduce` problem is that Python 3 removed it. Shoot, I'll have to come up with an alternative.

coredumperror:
Hurray for passing tests! Now I just need to add tests for the new functionality.

m1kola:
I found some problems.

1. On permissions page

    screen shot 2016-04-15 at 14 00 20

    Not sure that you should fix this issue in this PR. I think problem is more general.
1. I have a page tree like this:

   ```
   <Root>
   |- Home page
       |- Blog index
          |- Blog post 1
          |- Blog post 2
          |- Blog post 3
       |- Some other page
   ```
And I have a group "Editors" with all permissions for "Blog index" page. When user from the "Editors" group try to click on "Explorer" menu he receives `500 Internal Server Error` in JS console with following traceback:

    ```
    Internal Server Error: /admin/explorer-nav/
      File "/Users/m1kola/.virtualenvs/wagtail_dev/lib/python2.7/site-packages/django/core/handlers/base.py", line 149, in get_response
        response = self.process_exception_by_middleware(e, request)
      File "/Users/m1kola/.virtualenvs/wagtail_dev/lib/python2.7/site-packages/django/core/handlers/base.py", line 147, in get_response
      File "/Users/m1kola/.virtualenvs/wagtail_dev/lib/python2.7/site-packages/django/views/decorators/cache.py", line 43, in _cache_controlled
        response = viewfunc(request, *args, **kw)
      File "/Users/m1kola/Projects/wagtail/wagtail/wagtailadmin/decorators.py", line 23, in decorated_view
      File "/Users/m1kola/Projects/wagtail/wagtail/wagtailadmin/views/pages.py", line 36, in explorer_nav
        'nodes': get_navigation_menu_items(request.user),
      File "/Users/m1kola/Projects/wagtail/wagtail/wagtailcore/models.py", line 1497, in get_navigation_menu_items
        parent_page, parent_childlist = depth_list[page.depth - 1]
    IndexError: list index out of range
    ```

Sorry, I hadn't enough time to finish review today. I will continue on this weekend. I hope it's ok.

coredumperror:
Huh, I thought I'd squashed that bug with the explorer-nav list index out of range. I'll figure out what's up.

If nothing else, you've reminded me that I completely forgot to add tests for the explorer nav changes. I'll get right on that.

m1kola:
Looks like user without permissions can:
* See a list of page revisions
* See a preview of revisions

wagtail/wagtailadmin/views/pages.py
```diff
@@ -17,22 +17,38 @@
...
-        'nodes': get_navigation_menu_items(),
+        'nodes': get_navigation_menu_items(request.user),
     })
 
 
 def index(request, parent_page_id=None):
     if parent_page_id:
         parent_page = get_object_or_404(Page, id=parent_page_id)
+
+        if not UserPagePermissionsProxy(request.user).for_page(parent_page).can_explore():
```
m1kola: I think you can use `parent_page.permissions_for_user(request.user)` here instead of `UserPagePermissionsProxy(request.user).for_page(parent_page)`.

wagtail/wagtailcore/models.py
```diff
@@ -1277,17 +1283,184 @@ def serve_password_required_response(self, request, form, action_url):
...
+    # that has the same set of groups will have the same permitted paths. I'd like to cache by groupset, which would
+    # remove all that duplication, but that would require an extra DB call every time, to get the user's group list.
+    cache_key = EXPLORABLE_PATHS_PREFIX.format(user.pk)
+    permitted_paths, required_ancestors = cache.get(cache_key, ([], []))
+    if not permitted_paths:
+        ###################
+        # Get paths by User
+        ###################
+        permissions = (
+            GroupPagePermission.objects
```
m1kola: Can we put all logic related to permission into `UserPagePermissionsProxy` and `PagePermissionTester`?

m1kola:
Would be nice to have generic find-closest-common-ancestor function (maybe in `utils` module?). I mean, probably we will need to do this with other `MP_Node`-based objects.

I will continue tomorrow. @coredumperror, thank you for a great job!

wagtail/wagtailadmin/templates/wagtailadmin/shared/breadcrumb.html
```diff
@@ -1,8 +1,10 @@
 {% load i18n wagtailadmin_tags %}
 
 <ul class="breadcrumb">
-    {% for page in page.get_ancestors %}
-        {% if page.is_root %}
+    {% explorable_ancestors page request.user as ancestors %}
+    {% for page in ancestors %}
+        {% is_explorable_root page request.user as page_is_explorable_root %}
+        {% if page_is_explorable_root %}
```
m1kola: What should I see in breadcrumbs if:
* I have tree like this:

   ```
   <Root>
   |- Home page
       |- Blog index
          |- Blog post 1
          |- Blog post 2
          |- Blog post 3
              |- Blog post 3 - page 1
              |- Blog post 3 - page 2
       |- Some other page
   ```
* I have all permissions for "Blog index"
* I will open "Blog post 3 - page 1"?

I see only one item with `icon-home` and i'm afraid that I got not an expected result :)

wagtail/wagtailadmin/views/pages.py
```diff
@@ -17,22 +17,38 @@
...
+        # The user visited /admin/pages/, so we need to figure out the appropriate "root page" to show them.
         parent_page = Page.get_first_root_node()
+        if not request.user.is_superuser:
+            # Non-superusers get the Closest Common Ancestor of their permitted pages, if they have any.
+            permitted_pages, required_ancestors = get_explorable_page_paths(request.user)
+            if permitted_pages:
+                parent_page = Page.objects.get(path=required_ancestors[0])
 
-    pages = parent_page.get_children().prefetch_related('content_type')
+    pages = parent_page.get_explorable_children(request.user).prefetch_related('content_type')
```
m1kola: Are you sure that we can't just do the following?
```
pages = parent_page.get_children().prefetch_related('content_type')
```

As far as I know, it's not possible to have a subpage with less permissions than the parent.

wagtail/wagtailcore/models.py
```diff
@@ -1277,17 +1283,184 @@ def serve_password_required_response(self, request, form, action_url):
...
+            .select_related('page')
+        )
+        # We temporarily use a set here because "permissions" has duplicates of every path that has multiple perms.
+        permitted_paths = list(set(perm.page.path for perm in permissions))
+
+        # Calculate the "closest common ancestor" of all the permitted Pages by finding their common prefix, then
+        # chopping off the end to make the length a multiple of Page.steplen.
+        if permitted_paths:
+            common_prefix = os.path.commonprefix(permitted_paths)
+            cca_path = common_prefix[:-(len(common_prefix) % Page.steplen) or None]
```
m1kola: Are we really need `None` here?

wagtail/wagtailcore/models.py
```diff
@@ -1277,17 +1283,184 @@ def serve_password_required_response(self, request, form, action_url):
...
+            permitted_paths, required_ancestors = get_explorable_page_paths(user)
+            if permitted_paths:
+                # Limit the listing to only the user's explorable pages. Always include the Root page, even if it's
+                # not an explorable path, since it's required for the depth_list algorithm.
+                pages = pages.filter(
+                    convert_explorable_paths_to_Q(permitted_paths, required_ancestors) | Q(depth=1)
+                )
+
+                # Find the length of the shortest explorable path.
+                shortest_path_len = len(min(permitted_paths + required_ancestors, key=len))
```
m1kola: If I'm not mistaken, shortest path will be closest common ancestor (`required_ancestors[0]` in current version). So probably you don't need to use `min`.

wagtail/wagtailcore/models.py
```diff
@@ -1277,17 +1283,184 @@ def serve_password_required_response(self, request, form, action_url):
...
+            # that they can.
+            temp = set()
+            for path in permitted_paths:
+                while len(path) > len(cca_path):
+                    path = path[:-Page.steplen]
+                    if path not in required_ancestors:
+                        temp.add(path)
+            # We need to make sure that the CCA is first in required_ancestors (so the /admin/pages/ page can be set to
+            # display the page tree starting at the CCA), but the order of the other paths doesn't matter.
+            required_ancestors.extend(temp)
```
m1kola: Do we really need to find all `required_ancestors`? I'm not sure, but closest common ancestor  probably will be enough. See:
* [Here](https://github.com/torchbox/wagtail/pull/2463/files#diff-d099bfc6462f8efe8cc5177b2790e900R49), [here](https://github.com/torchbox/wagtail/pull/2463/files#diff-9ec0dccfdaa33a0fd310c76b5db2cb37R1328) and [here](https://github.com/torchbox/wagtail/pull/2463/files#diff-9ec0dccfdaa33a0fd310c76b5db2cb37R1448) we can to use only closest common ancestor, if present
* [Here](https://github.com/torchbox/wagtail/pull/2463/files#diff-9ec0dccfdaa33a0fd310c76b5db2cb37R1297) we can use only `permitted_paths` (`path__startswith` because it's not possible to have a subpage with less permissions than the parent)
* [Here](https://github.com/torchbox/wagtail/pull/2463/files#diff-9ec0dccfdaa33a0fd310c76b5db2cb37R1318) we need to find all paths in range from "closest common ancestor" to `self.path`

Please, correct me if I'm wrong.

coredumperror: I'm pretty sure all the required ancestors are needed for page trees like this:

|- Home Page
   |- Blog Index
      |- Blog Post 1
      |- Blog Post 2
      |- Blog Post 3
   |- Some Other Page

If the user has permissions on Blog Post 3 and Some Other Page, their Closest Common Ancestor is Home Page. So in order to navigate to Blog Post 3 within the Explorer, they need to be able to see Blog index, which they can't do unless we let them see all their required ancestors.

wagtail/wagtailcore/models.py
coredumperror: Oh hey, you're right! I wrote the original version of that algorithm before I decided to forcibly put the CCA at index 0 in required_ancestors. I'll fix that.

wagtail/wagtailcore/models.py
coredumperror: The `None` is for when `len(common_prefix) % Page.steplen == 0`, as slicing a string with `[:-0]` will result in an empty string, when what we actually want is the entire string. Slicing with `None` is a feature I didn't even know existed until I found that algorithm in a StackOverflow post.

wagtail/wagtailadmin/views/pages.py
coredumperror: It's true that it isn't possible to have subpages with less permission, but when navigating through the Explorer, we often need to show users some pages that they don't have any permission on. Take this page tree:

    |- Home Page
       |- Blog Index
          |- Blog Post 1
          |- Blog Post 2
       |- Some Other Page
       |- Yet Another Page

Consider what the user needs to be shown when they have page permissions on Blog Post 2 and Some Other Page. Explorer needs to show them Homepage, so they can navigate through Blog Index and see Some Other Page, but they need to *not* be shown Yet Another Page or Blog Post 1, since they aren't in the User's explorable paths. That's why `Page.get_explorable_children(user)` exists.

wagtail/wagtailadmin/templates/wagtailadmin/shared/breadcrumb.html
coredumperror: That sounds like a bug, since "Blog Post 3" should be visible as text, with "Blog Index" being linked as that home icon. I'll look into it.

coredumperror: Yup, bug. I completely flubbed `Page.get_explorable_ancestors(user)`. The correct way to do it is much easier than what I came up with, too.

None of the tests in the existing test suite managed to catch this problem, which is how I managed to miss this bug. I'll remedy that.

wagtail/wagtailadmin/views/pages.py
coredumperror: Indeed you can. I'd forgotten about the `permissions_for_user()` function. Easily fixed.

wagtail/wagtailcore/models.py
coredumperror: Good idea. I'll fix that.

coredumperror:
That issue with page revisions not getting restricted is exactly the kind of feedback I knew I'd need. I had forgotten that revisions even existed. I'll get that fixed ASAP.

I also added a `get_closest_common_ancestor_path(user)` utility function. I couldn't put it in `wagtailcore/utils.py` without a circular import happening, so I left it in `wagtailcore/models.py` along with the other new functions I added.

coredumperror:
Hmmm, so now it's complaining about merge conflicts, but I can't figure out where to view these conflicts to figure out what I need to fix.

gasman:
Generally you don't need to worry about merge conflicts - we (i.e. the Wagtail core team) will deal with those when the feature is complete and we merge it into master. It only becomes a problem in rare cases when the PR has been sitting around for months and the master branch has diverged massively in the meantime...

Having said that, if you're feeling confident with git, you can run `git pull` on your master branch to bring it up to date, then `git rebase master` on your view-in-admin-perm branch. Once you've got your pull request changes cleanly applied on top of the latest updates from master, you can run `git push --force` to update this PR.

coredumperror:
Ah, cool. I didn't know that `git push --force` would work. I'll try that after lunch.

coredumperror:
OK, I did the rebase from the `torchbox/master` branch, found the conflicting change (it was in `wagtail/wagtailadmin/utils.py`, and was a simple fix), and then did a forced push.

coredumperror:
Oh my gosh I totally didn't notice that the build was failing. I'll fix that immediately.

coredumperror:
Arg, now the tests are failing in a new way I've never seen before! :( These failures don't happen when I run the tests locally. What the heck?

m1kola:
Hi @coredumperror! Did you run tests locally on Python 2.*? Looks like they fails only on Python 3.*. Do you have Python 3.*?

m1kola:
I suppose tests are failing because of type casting (`str(response)`, etc), but I can't check it at the moment.

coredumperror:
I do all my work on Python 2 because the OSes my code shop's apps run on don't come with Python 3. But I should be able to get a Python 3 virtualenv up and running to do some tests in there.

I'm casting response to `str()` because you need to send a string into `self.assertInHTML()`. I tried using `response.content`, but that's a `bytes` object in Python 3, rather than a unicode string, so it breaks the test in a similar way.

coredumperror:
Is there a better way to test HTML content? I was afraid to use a third party extension to django's testing framework (it's surprisingly lacking when it comes to testing the DOM tree) since I didn't want to (or know how to) add new dependencies, but installing a Python 3 pyenv made me realized that `beautifulsoup4` is already a dependency, which might mean that there are testing tools in place to do DOM-based tests.

coredumperror:
OK, I figured out the problem. Django is written such that `str(response)` gives *entirely different results* on Python 2 vs 3. It returns the full HTTP header set and content body in Python 2, but returns the same thing as `repr(response)` on Python 3. I've opened up a ticket on their issue queue to see what's up with this odd behavior.

Fortunately, I found a workaround that is compatible with both: `str(response.content)` functions normally in both Python versions, though I honestly can't say why. `response.content` alone causes an error in `assertInHTML()` because it's a `bytes` object. But `str(response.content)` *returns a bytes object*, yet doesn't cause an error.

coredumperror:
Hmm, the only build failure was a single test run that triggered timeout errors in code I didn't change. Is there a way to rerun that test?

coredumperror:
Hmmm, and now it's reporting that the test passed. Do retests just automatically happen?

m1kola:
Hi @coredumperror! I will return to this PR on weekend. Is it ok?

> Hmm, the only build failure was a single test run that triggered timeout errors in code I didn't change. Is there a way to rerun that test?

> Hmmm, and now it's reporting that the test passed. Do retests just automatically happen?

It happens sometimes :) I just restarted failed builds.

coredumperror:
This weekend is fine. I've got plenty to do in my own code in the meantime. 

I'm going to begin work on another major PR once this change gets approved, as this new permission restriction needs to apply to Choosers as well. The change will also depend on #1645, though, so it may get a bit hairy, having a PR that depends on two other PRs. Not really sure how to deal with that, to be honest.

wagtail/wagtailcore/models.py
```diff
@@ -1316,17 +1322,191 @@ def serve_password_required_response(self, request, form, action_url):
...
+    visible.
+
+    NOTE: The first element of the second list is always the Closest Common Ancestor of the User's permitted Pages.
+    """
+    # We cache the output of this function because it's very expensive to generate it fresh every time the user loads
+    # an Explorer page. Unfortunately, this does lead to a lot of duplciated data in the cache, as every single user
+    # that has the same set of groups will have the same permitted paths. I'd like to cache by groupset, which would
+    # remove all that duplication, but that would require an extra DB call every time, to get the user's group list.
+    cache_key = EXPLORABLE_PATHS_PREFIX.format(user.pk)
+    permitted_paths, required_ancestors = cache.get(cache_key, ([], []))
```
m1kola: Yes, it's expensive to execute this function every time, but there is another problem. This function is called at least three times (maybe more) for one request and looks like duplicated calls slows down everything.

This function itself doesn't look expensive. Probably per-request cache will be enough here. This eliminates the need to do cache cleaning. I think that current approach may lead to issues with outdated cache in future. For example, if someone will add a new function that changes tree.

If there any reason to use permanent cache, I think would be better to define signals (page moved, group permission changed, user changed, etc) and define handler for these signals that will call `clear_explorable_paths_cache`.

wagtail/wagtailcore/models.py
```diff
@@ -1316,17 +1322,191 @@ def serve_password_required_response(self, request, form, action_url):
...
     # or are at the top-level (this rule required so that an empty site out-of-the-box has a working menu)
     pages = Page.objects.filter(Q(depth=2) | Q(numchild__gt=0)).order_by('path')
 
-    # Turn this into a tree structure:
+    # We need to treat the top of the user's explorable tree as if it were depth=2, or the depth_list algorithm won't
+    # work. So if a user isn't permitted to see pages at or below depth 2, we must change the apparent depth of each
+    # of their explorable pages.
+    depth_change = 0
+    if not user.is_superuser:
+        # For backward compatibility of the explorer_nav template tag, we accept "None" for the user argument, in which
```
m1kola: I don't think that we need to take care about backward compatibility here, because it's internal template tag.

m1kola:
@coredumperror thanks for you patience! :) This PR still need some work (on revisions, for example), but in common it looks good for me. Also I think that you need to know about #2507.

I feel that would be great if someone else has looked at this PR.

@gasman or @kaedroho do you have time to take a look at this PR? I'm afraid that I might miss something.

wagtail/wagtailcore/models.py
coredumperror: I'll defer to your expertise here about the call not looking very expensive. I'm certainly not an expert in database performance. Plus, it'll definitely be nice to not need all that extra code for manual cache clearing. I'll make this change right away.

wagtail/wagtailcore/models.py
coredumperror: Great! I figured as much, but I wasn't sure if the tag was intentionally internal or not. I'll go ahead and remove that compatibility code.

coredumperror:
> This PR still need some work (on revisions, for example)

What do you mean? I fixed the permissions issue with regards to revisions being visible to unprivileged users.

m1kola:
> What do you mean? I fixed the permissions issue with regards to revisions being visible to unprivileged users.

I'm sorry, here is more details.

I still have access to:
* List of revisions
* Preview of revisions
* Copy page

screen shot 2016-05-02 at 22 48 31

My tree:
```
|- Home Page
   |- Blog Index
      |- Blog Post 1
      |- Blog Post 2
      |- Blog Post 3
   |- Standard index
      |- Sub Page 1
      |- Sub Page 2
      |- Sub Page 3
```
I have `add` and `edit` permissions for `Blog Index` and for `Standard index`, but I haven't permissions for `Home Page`.

coredumperror:
Ah, OK. I'll fix that.

m1kola:
@coredumperror thanks!

I'm not sure about "live" button. Probably would be better to hide this button too.

coredumperror:
My most recent commit fixes the Live, Copy, and Revisions buttons being visible on required ancestors, but I'm not sure the change I made is OK. I had to alter the `register_page_listing_buttons` and `register_page_listing_more_buttons` hooks to let their functions take a `User` object as the third argument. This will be a backwards-incompatible change for any sites that have implemented those hooks, and I'm not sure what the policy is in regards to such changes.

I still need to add tests for the new restrictions as well, but it's the end of the day, and I wanted to get this change out there now, rather than waiting until tomorrow morning.

coredumperror:
Hurray for finding a way to get rid of the actions list without needing an API change! It was so easy, I can't believe I didn't think of it earlier.

coredumperror:
I rebased from master again, since several of the changes I made (and then unmade as part of the refactor) conflicted with other changes.

coredumperror:
Shoot, I just realized that the new modeladmin contrib module needs the Explorable pages restrictions applied to it as well. I'll gets started on that.

coredumperror:
The change to add explorable page restriction to ModelAdmin was surprisingly easy (once I figured out which function actually needed to be edited), but adding tests for it was an interesting challenge.

coredumperror:
I added explorable page restrictions to the Page Chooser, including tests.

While doing so, I realized that I'd forgotten to add explorable page restrictions to the page search features of both the Explorer and the Chooser, so I'll be doing that next. Hopefully, though, that will be the final change that wagtail itself needs for multi-tenant support.

This doesn't cover restricting the use of Collections, but those aren't directly tied to Sites in any way, so there isn't really a way to do that directly within Wagtail without changing `Collection`. It might make sense to add an optional `ForeignKey` to `Site` in the `Collection` model, which would create a connection upon which access restriction could be applied, but I'm not sure that's a good way to go about it. Or if such a thing is even desired as a part of Wagtail itself.

m1kola:
Hi @coredumperror.

Thank you for your great work! And sorry for the delay: we were busy with 1.5 release.

It's hard to say when I will be able to return to code review, because In the near future I will be busy with the client work, but I hope we will finish this feature for Wagtail 1.6.

coredumperror:
Thanks for getting back to me. I figured you guys were probably super busy with something, since you'd always been so communicative before. 

I made several significant changes in my most recent commit, because I had completely forgotten about the "choose" permission that @gasman and I brainstormed when I first started this project. That required me to majorly refactor the "explorability" algorithms, in order to add "choosability" as well. 

That's all done now, baring any bugs that I might find while writing the tests, which I hope to be done with by some time on Friday.

wagtail/contrib/modeladmin/options.py
```diff
@@ -243,13 +243,16 @@ def get_ordering(self, request):
 
     def get_queryset(self, request):
         """
-        Returns a QuerySet of all model instances that can be edited by the
-        admin site.
+        Returns a QuerySet of all model instances that can be explored by the
+        current user.
         """
         qs = self.model._default_manager.get_queryset()
+        qs = filter_explorable_pages(qs, request.user, include_ancestors=False)
```
gasman: Given that the primary motivation for ModelAdmin's page views is to present the page listing in a way that doesn't require navigating the tree, do we really want to filter to the explorable pages? It seems to me it would be more appropriate to narrow it down to the pages that the user has edit permission for.

(If so - and modeladmin doesn't currently filter the listing by user permission - then that would probably be a job for a separate PR, as this change would then be independent of the task of implementing 'explore' permission restrictions.)

wagtail/wagtailadmin/templates/wagtailadmin/chooser/_browse_results.html
```diff
@@ -4,7 +4,7 @@
 {% include "wagtailadmin/shared/breadcrumb.html" with page=parent_page choosing=1 %}
 
 {% if pages %}
-    {% include "wagtailadmin/pages/listing/_list_choose.html" with allow_navigation=1 orderable=0 pages=pages parent_page=parent_page %}
+    {% include "wagtailadmin/pages/listing/_list_choose.html" with allow_navigation=1 orderable=0 choosing=1 pages=pages parent_page=parent_page %}
```
gasman: Why is `choosing=1` needed here? `_list_choose.html` doesn't appear to use it (and in any case it would be redundant, because `_list_choose.html` should by definition *only* be used for choosing...)

wagtail/contrib/modeladmin/options.py
coredumperror: "Explorable" isn't a specific permission: it's the equivalent of "the user has at least one of the following perms: Add, Edit, Publish, or Lock". Only "Choosable" is a specific permission.

The reason I put the explorability restriction on the ModelAdmin listing was to prevent non-superusers from seeing Pages that belong to other Sites, which is exactly what the explorability filter on the Explorer is all about.

wagtail/wagtailadmin/templates/wagtailadmin/chooser/_browse_results.html
coredumperror: I think I must have added that there by mistake, or forgot to remove it after taking out the code that was making use of it. I'll revert that change.

gasman:
Hi @coredumperror,
Unfortunately I think this PR has now reached the point where it's too complex for us to feasibly review and merge in one go, and we need to look at rebuilding it as several smaller, self-contained PRs.

In particular, I'm concerned about the requirement for a new middleware (especially the need for it to be imported in wagtailcore/models.py, as model code should not have any dependency on the request/response cycle) - normally that would be a major pull request in its own right, but here it's mixed in with too many other concerns to fully understand why it's needed, and whether there are any alternative approaches.

I'm also not clear on why Site records come into play. In Wagtail's current permission model, permissions are only determined by a page's location in the overall tree, and the presence or absence of any Site records at any level doesn't affect that - they only affect the URL that the page is given. From my initial read through the code, it looks like the only effect is to send a 404 response rather than a 403 for 'permission denied' - but again there are too many other moving parts in this PR to be really sure of that.

As a first step, I'd suggest putting together a PR that just implements the cut-down tree navigation in the explorer, without introducing the "choose" permission. (This could most likely be split up further into a PR that deals with the explorer menu pop-out, and another that deals with the explorer views.) If there are any utility functions they rely on, they could conceivably be split into their own PR as well (with their own unit tests, so that they can be reviewed independently).

coredumperror:
That's gonna be a lot of work... but I understand the reason. I'll figure out the optimal way to split these functionality changes up into their own PRs.

The middleware is there because @m1kola suggested that I cache the complex db query for explorable pages in a per-request cache, to avoid performing the same calculation multiple times in one request. The way I found to do that involves setting up a middleware that attaches a cache to the request object. Though now that I've written that sentence, I realize that there's a significantly better way to handle the cache instances that doesn't require `models.py` to import the middleware. I'll make that change as part of the PR split.

As for Site records, "explorability" is handled purely by Permissions. So you're correct that the only place that Site records are directly queried is to differentiate between 404 and 403 errors for "unexplorable" pages. 403 errors are issued when a User on Site A attempts to explore a Page on Site A that they're not permitted to explore. 404 errors are issued when a User who visits the server from the domain of Site A attempts to explore a page that is on Site B (i.e. it descends from Site B's root page). 

The reason it's important to differentiate this is so that information about Site A cannot leak to users of Site B. Getting a 403 from a Page url that a Site B user knows is not on their own Site proves that other Sites exist on their server, even to users who aren't allowed to know that.

gasman:
> The reason it's important to differentiate this is so that information about Site A cannot leak to users of Site B. Getting a 403 from a Page url that a Site B user knows is not on their own Site proves that other Sites exist on their server, even to users who aren't allowed to know that.

If I understand correctly - your concern is that users on site B will plug random (or sequential) IDs into URLs like `http://site-b.example.com/admin/pages/<id>/` , see that it returns a 403 response, and infer from that that another site exists on the same installation.

It seems to me that you'd need quite detailed knowledge of Wagtail's permission model to come to that conclusion - rather than simply "there is _some page_ on this site that I don't have edit permission for" - and I can't really see how either of those conclusions represents a meaningful leakage of information, even in an extremely high security setting. (What harmful thing can you do with the knowledge "another site exists on this CMS installation"?)

If this level of lock-down against knowledgeable users *is* needed, then that's probably a job for a full-scale security review rather than something we can do through the sites / permissions model - for example, switching from sequential IDs to random identifiers - and that would be very much out of scope for this feature.

coredumperror:
You know what, I think the entire purpose of this feature was negated when I made a change some weeks ago regarding the order of evaluation for the explorability algorithm. That change was in commit bfff272, and I think I did that at your request, actually.

The algorithm previously prevented users from seeing Pages on different Sites even if they had permission to do so, which justified the existence of this Site check. But since that's an unacceptable limitation, that actually obviates the Site check entirely. I'll go ahead and remove it.

coredumperror:
Ah shoot, I got distracted and pushed a broken build. I'll fix that in a sec.

coredumperror:
After working and thinking on this for the last few days, I'm not sure I **can** split this PR up into separate branches. The main reasons being:

1) The "explorability" code is so intertwined with the "choosability" code that there's no way to fully split them up without the separate PRs being intimately dependent upon one another. And then what's the point of them being separate in the first place?

2) Almost all the tests I wrote are integration tests. That means that attempting to split the PR into say, one PR for the shared internal functions, one for the API changes related to explorability, and one for the changes related to choosability, will require most of the tests to be re-written from scratch. That seems like a huge waste of time for a test suite that already works just fine.

3) I've never done this before, so I have no idea what I'm doing. I flailed around for hours trying to come up with a way to split the explorability and choosability code up in a logical manner, but _completely_ failed to make any progress. I'm feeling pretty defeated. :(

I already split out the middleware into its own PR, and re-wrote the rest of the code in a way that removes both of the main concerns you referenced in your comment from last week. I just don't think there's much else that *can* be done at this point, without wasting a whole lot of time for next to no benefit.

coredumperror:
I recently discovered a significant problem with this, though the fix is pretty straightforward. Currently, after saving an edit to a Page, Wagtail blindly redirects the user to the Explorer page for that Page's parent. Unfortunately, that parent page might not be Explorable, in which case Wagtail will throw a 403.

Fortunately, the `after_edit_page` hook exists, so it's possible to get around this easily enough:

```
@hooks.register('after_edit_page')
def permissions_aware_after_edit_redirect(request, page):
    """
    Normally, after saving an edit to a Page, Wagtail redirects the user to Page's parent in the Explorer. However,
    this will trigger a 403 error if the user doesn't have permissions on that parent (e.g. when editing their site's
    homepage). If such a permissions failure would occur, this code brings the user back to the Page itself, instead.

    This function fully recreates what would have happened in the Page edit view after this hook fires, since its
    existence means that code won't run.
    """
    page_perms = page.permissions_for_user(request.user)
    is_publishing = bool(request.POST.get('action-publish')) and page_perms.can_publish()
    is_submitting = bool(request.POST.get('action-submit'))
    next_url = get_valid_next_url_from_request(request)

    if is_publishing or is_submitting:
        # Redirect back to 'next' url if present.
        if next_url:
            return redirect(next_url)

        # Redirect back to parent if it's explorable. Otherwise, redirect to the Page itself.
        parent_page = page.get_parent()
        if parent_page.permissions_for_user(request.user, request).can_explore():
            page_id = parent_page.id
        else:
            page_id = page.id
        return redirect('wagtailadmin_explore', page_id)
    else:
        # Just saving - remain on edit page for further edits.
        target_url = reverse('wagtailadmin_pages:edit', args=[page.id])
        if next_url:
            # Ensure the 'next' url is passed through again if present.
            target_url += '?next=%s' % urlquote(next_url)
        return redirect(target_url)
```

This explorability check will need to be incorporated into `wagtail.wagtailadmin.views.pages.edit()` once the Explorability restrictions from this PR are merged in.

bmfs:
Was was the conclusion of this feature?

gasman:
@bmfs It's partially implemented in #2869, which limits the explorer menu/views to just the pages the user has permission on. The remaining task is to introduce a 'choose' permission level so that we can do the same thing on page chooser interfaces; this is still something we'd like to do, but it's waiting for someone to either implement or extract from this PR.

coredumperror:
My team got around the not-yet-implemented parts of this using a number of monkey-patches to Wagtail's codebase. After the holiday, I'll be happy to share those patches, if you're interested.

phillbaker:
@coredumperror I'd be interested 😁 . Are they all on a branch?

coredumperror:
They're actually implemented within my team's codebase, as a django app called `wagtail_patches`. It contains a bunch of code that goes in and alters Wagtail's modules and classes at runtime, mostly by swapping out functions and methods with tweaked versions that add Site-separation behavior.

Though now that I think about it, the full suite of multitenancy functionality includes several disparate bits of code that aren't just in `wagtail_patches`. Actually putting that all together into a comprehensible package is a more complex task than I realized.

For instance, in our system's `MIDDLEWARE` setting, we replaced `wagtail.wagtailcore.middleware.SiteMiddleware` with a middleware that prevents users who are only members of Site B from logging in to Site A's admin interface. And "membership in Site B" is defined through some convoluted stuff we did with Group names and Site hostnames, mostly in `wagtail_patches`(since you can't associate a django `Group` with a wagtail `Site` in any direct way :( ).

Anyway, my boss gave me permission to share the code, so I'll try to find some time to slap together a github repo with everything that's relevant. However, it'll likely take some extra effort on your end to make it all work together, not to mention a number of commits from me to actually pull in all the relevant code.

phillbaker:
Thanks @coredumperror, totally reasonable and much appreciated!

coredumperror:
OK, here's what I've managed to quickly slap together:
https://github.com/coredumperror/wagtail-multitenancy

The meat of this code is the `wagtail_patches` app. It contains all the monkey-patches to the Wagtail code that makes it behave like Sites are totally separate entities. Users are not allowed to interact with Site A, or even know that Site A exists, unless they are a member of Site A. Being a member involves belonging to a Group which is associated with that Site (the `site_creator` app creates the default Groups for a Site alongside the SIte itself).

The rest of the code includes things like the `MultitenantSiteMiddleware`, which replaces wagtail's `SiteMiddleware` in order to enforce separation between the Sites.

I haven't got time to do this properly, so it's entirely possible that I missed some critical code from somewhere outside of wagtail_patches and site_creator (the only parts that I copied whole-hog from our repo). If you discover that anything referenced from this code is missing, or have any questions on how something works, feel free to ask.

phillbaker:
@coredumperror thanks so much! 

This is great - we'll play around with it a bit over the next several weeks and update this thread with what we find.

phillbaker:
@coredumperror thanks again for posting that code - we've had a chance to go review the code and try things out, would it be okay to ask a few questions / clarify some sections? Where would be the best place to do that (this PR, issues on that repo, elsewhere)?

coredumperror:
Yeah, this probably isn't the right place for in-depth discussion on this code. The issue queue on that repo I made would work well.

PR #2465, opened 2016-04-09
Remove redundant debug settings from project template

### LINKED ISSUE #2078: Remove redundant template debug lines from project template
Ref: https://github.com/torchbox/wagtail/blob/9ff7961a3c8f508ad17735cd815335bad12fd67f/wagtail/project_template/project_name/settings/dev.py#L7-L8
#1688

According to https://docs.djangoproject.com/en/1.9/topics/templates/#django.template.backends.django.DjangoTemplates, the 'debug' option on the DjangoTemplates engine defaults to the global DEBUG setting, so setting this here is apparently redundant. (Also, there's no corresponding option for the Jinja2 backend, so setting this for all engines is not strictly correct.)

So, we just need someone to double-check that with these lines removed, template debug info still displays in development mode but not in production.

timorieber:
Fixes #2078.

I've double checked that removing this code leads to the same behaviour as before.

### DIFF
--- a/wagtail/project_template/project_name/settings/dev.py
+++ b/wagtail/project_template/project_name/settings/dev.py
@@ -5,9 +5,6 @@ from .base import *
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = True
 
-for template_engine in TEMPLATES:
-    template_engine['OPTIONS']['debug'] = True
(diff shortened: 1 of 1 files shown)


kaedroho:
Thanks!

PR #2470, opened 2016-04-10
Replaces direct usages of Document model

### LINKED ISSUE #2460: Use get_document_model instead of Document
I found some places where `Document` model used directly:
- https://github.com/torchbox/wagtail/blob/v1.4.3/wagtail/contrib/wagtailmedusa/renderers.py#L21
- https://github.com/torchbox/wagtail/blob/v1.4.3/wagtail/contrib/wagtailapi/signal_handlers.py#L40-41
- https://github.com/torchbox/wagtail/blob/v1.4.3/wagtail/contrib/wagtailapi/signal_handlers.py#L53-54
- https://github.com/torchbox/wagtail/blob/v1.4.3/wagtail/contrib/wagtailapi/endpoints.py#L245

Looks like we need to use `get_document_model` in these places.

m1kola:
We need to use `get_document_model` instead of `Document` directly because Wagtail provides ability to define custom document model.

Fixes #2460

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -19,6 +19,7 @@ Changelog
  * Fix: The currently selected day is now highlighted only in the correct month in date pickers (Jonas Lergell)
  * Fix: Fixed crash when an image without a source file was resized with the "dynamic serve view"
  * Fix: Registered settings admin menu items now show active correctly (Matthew Downey)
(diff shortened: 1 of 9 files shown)


gasman:
Merged in 6c53a439a5ab315dd9e49092ac333ccf445fc7f9 + parents, thanks!

PR #2471, opened 2016-04-10
Changes messages in empty list of images\documents

m1kola:
This commit introduces tiny UX improvements for empty lists of images and documents:
 * Same message in both lists
 * Gives link to multiple upload form in both lists


kaedroho:
Much better, thanks!

PR #2474, opened 2016-04-12
Implement jpegquality and outformat image filters

### LINKED ISSUE #908: Compression option per image when requesting wagtail to make images
When you call wagtail externally or through the admin URL generator interface to produce an image using a filter and physical size, could there also be the option to specify the JPEG compression level? Currently the default 85 can be overridden in settings for all images only.

### LINKED ISSUE #1205: Document WAGTAILIMAGES_JPEG_QUALITY

### LINKED ISSUE #2472: Wagtail 1.4.x breaks image conversion
Hi, I've been diving into the source code for wagtail trying to figure out why code we had written to create circle-cropped image renditions no longer works in wagtail 1.4.x when it worked fine in 1.3.x and earlier. There is definitely some re-work on our part that needs to be done, but I think I've got a good handle on most of that now.

The piece that is missing can be found here: https://github.com/torchbox/wagtail/blob/stable/1.3.x/wagtail/wagtailimages/models.py
vs.
https://github.com/torchbox/wagtail/blob/stable/1.4.x/wagtail/wagtailimages/models.py

Around line 388; in the run of the Filter, in wagtail 1.3.x the output format of the image rendition was determined AFTER all the operations on it had been performed, which allowed us to apply an alpha layer and save any image as a PNG, simply by changing the `original_format`.

In wagtail 1.4.x however, the original image format is stored BEFORE any operations are performed and then used to save the image. So it doesn't matter what our custom operations tell wagtail the file should be saved as, any image rendition will always be saved as in the same format as the original image.

kaedroho:
This pull request implements two new image filters:
 - ``format`` - This allows the developer to specify an output format to use in a placement (For example, ``format-png`` will force Wagtail to always output PNG files irrespective of the original format) (fixes #2472)
 - ``jpegquality`` -  This allows the developer to specifiy the amount of compression to apply JPEG files in an image placement (fixes #908 and #1847)

These are made possible by a new ``env`` argument that gets passed into image filter run methods. It's a mutable dictionary that can be used to pass messages along the filter chain and back to Wagtail.

The docs also fix #1205

### DIFF
--- a/docs/topics/images.rst
+++ b/docs/topics/images.rst
@@ -203,9 +203,9 @@ Output image format
 
 Wagtail may automatically change the format of some images when they are resized:
 
- - PNG and JPEG images always won't change format
+ - PNG and JPEG images don't change format
  - GIF images without animation are converted to PNGs
- - BMP images are always converted to PNGs
+ - BMP images are converted to PNGs
 
 It is also possible to override the output format on a per-tag basis by using the
 ``format`` filter after the resize rule.
@@ -237,8 +237,15 @@ quality:
     # Make low-quality but small images
     WAGTAILIMAGES_JPEG_QUALITY = 40
 
-Note that this won't affect any previously generated images. You may want to
-delete all renditions so they can regenerate with the new setting.
+Note that this won't affect any previously generated images so you may want to
+delete all renditions so they can regenerate with the new setting. This can be
+done from the Django shell:
+
+.. code-block:: python
+
+    # Replace this with your custom rendition model if you use one
+    >>> from wagtail.wagtailimages.models import Rendition
+    >>> Rendition.objects.all().delete()
 
 Changing per-tag
 ^^^^^^^^^^^^^^^^
--- a/docs/topics/images.rst
+++ b/docs/topics/images.rst
@@ -32,18 +32,18 @@ The available resizing methods are as follows:
 
 .. glossary::
 
-    ``max`` 
+    ``max``
         (takes two dimensions)
 
         .. code-block:: html+django
 
             {% image page.photo max-1000x500 %}
 
-        Fit **within** the given dimensions. 
+        Fit **within** the given dimensions.
 
         The longest edge will be reduced to the matching dimension specified. For example, a portrait image of width 1000 and height 2000, treated with the ``max-1000x500`` rule (a landscape layout) would result in the image being shrunk so the *height* was 500 pixels and the width was 250.
 
-    ``min`` 
+    ``min``
         (takes two dimensions)
 
         .. code-block:: html+django
@@ -54,7 +54,7 @@ The available resizing methods are as follows:
 
         This may result in an image slightly **larger** than the dimensions you specify. A square image of width 2000 and height 2000, treated with the ``min-500x200`` rule would have its height and width changed to 500, i.e matching the *width* of the resize-rule, but greater than the height.
 
-    ``width`` 
+    ``width``
         (takes one dimension)
 
         .. code-block:: html+django
@@ -63,16 +63,16 @@ The available resizing methods are as follows:
 
         Reduces the width of the image to the dimension specified.
 
-    ``height`` 
+    ``height``
         (takes one dimension)
 
         .. code-block:: html+django
 
             {% image page.photo height-480 %}
 
-        Resize the height of the image to the dimension specified.. 
+        Resize the height of the image to the dimension specified..
 
-    ``fill`` 
+    ``fill``
         (takes two dimensions and an optional ``-c`` parameter)
 
         .. code-block:: html+django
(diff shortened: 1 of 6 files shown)


tomdyson:
This looks good! Is there a precedent for calling the dictionary 'env'? It feels a bit generic for image filter configuration.

kaedroho:
@tomdyson any suggestions for new names are welcome. It was deciding between ``env`` or ``props`` but decided on ``env`` as it's similar to Fabric's global ``env`` object.

I also considered ``context`` but thought this might be confused with template context (as the image operations are usually called from a template).

I went for something generic as I think this could be used for other things in the future, such as transforming the focal point whenver an image is resized/cropped

kaedroho:
I'm thinking of shortening ``outformat`` to just ``format``. Anyone against that?

kaedroho:
Almost done, I'm just not sure how to test the jpeg quality filter.

tomdyson:
Higher compression values should yield smaller images - can you test for that?

timheap:
You might also be able to mock the save method used by Willow or Pillow, as appropriate, to check if the correct settings are sent in.

kaedroho:
Tests added (by mocking ``PIL.Image.save``). Ready for review!

docs/topics/images.rst
```diff
@@ -197,3 +197,63 @@ Wagtail comes with three pre-defined image formats, but more can be defined in P
     The CSS classes added to images do **not** come with any accompanying stylesheets, or inline styles. e.g. the ``left`` class will do nothing, by default. The developer is expected to add these classes to their front end CSS files, to define exactly what they want ``left``, ``right`` or ``full-width`` to mean.
 
 For more information about image formats, including creating your own, see :ref:`rich_text_image_formats`
+
+Output image format
+-------------------
+
+Wagtail may automatically change the format of some images when they are resized:
+
+ - PNG and JPEG images always won't change format
```
tomdyson: always won't -> will never?

docs/topics/images.rst
```diff
@@ -197,3 +197,63 @@ Wagtail comes with three pre-defined image formats, but more can be defined in P
...
+
+.. code-block:: python
+
+    # settings.py
+
+    # Make low-quality but small images
+    WAGTAILIMAGES_JPEG_QUALITY = 40
+
+Note that this won't affect any previously generated images. You may want to
+delete all renditions so they can regenerate with the new setting.
```
tomdyson: Can you give a hint on how to do that?

wagtail/wagtailimages/models.py
```diff
@@ -402,28 +405,60 @@ def run(self, image, output):
             # Fix orientation of image
             willow = willow.auto_orient()
 
+            env = {
+                'original-format': original_format,
+            }
             for operation in self.operations:
-                willow = operation.run(willow, image) or willow
+                # Check that the operation can take the new "env" argument
```
tomdyson: drop "new" - it won't be new soon!

wagtail/wagtailimages/tests/test_image_operations.py
```diff
@@ -403,7 +411,34 @@ def test_runs_operations(self):
         )
         fil.run(image, BytesIO())
 
-        self.assertEqual(self.operation_instance.run.call_count, 2)
+        self.assertEqual(run_mock.call_count, 2)
+
+    def test_runs_operations_without_env_argument(self):
+        # The "env" argument was added in Wagtial 1.5. This tests that
```
tomdyson: Wagtial -> Wagtail

docs/topics/images.rst
kaedroho: Thanks! I'll also remove always/never since this isn't the case if you've overridden it!

kaedroho:
@tomdyson all suggestions implemented, thanks!

kaedroho:
Rebased!

SalahAdDin:
Why do you use env variable?

kaedroho:
> Why do you use env variable?

That's a vague question. There's a lot of different ways I could answer it. Please be more specific and I'll happily provide an answer.

docs/topics/images.rst
```diff
@@ -63,16 +63,16 @@ The available resizing methods are as follows:
...
-    ``height`` 
+    ``height``
         (takes one dimension)
 
         .. code-block:: html+django
 
             {% image page.photo height-480 %}
 
-        Resize the height of the image to the dimension specified.. 
+        Resize the height of the image to the dimension specified..
```
m1kola: Could we remove the last `.`?

wagtail/wagtailimages/tests/test_image_operations.py
```diff
@@ -1,14 +1,17 @@
...
 
-from django.test import TestCase
+import warnings
+
+from django.test import TestCase, override_settings
 from django.utils.six import BytesIO
-from mock import Mock
+from mock import Mock, patch
 
+from wagtail.utils.deprecation import RemovedInWagtail17Warning
```
m1kola: It seems to me that we need to replace this with `RemovedInWagtail19Warning`.

gasman:
Merged in 534b5d1fd9ce15ac88056d692e18a0a5620765a2 + parents

PR #2476, opened 2016-04-14
Make get_upload_to an overrideable method of AbstractRendition

gasman:
Allow custom upload path logic on subclasses of AbstractRendition, as we do for AbstractImage. Supersedes #2165


gasman:
rebased

kaedroho:
Merged in 2de05fb14f9c948443a6e769abc90287a9f523c6 Thanks!

PR #2479, opened 2016-04-15
Fix modeladmin's choose_parent view for non-superusers

### LINKED ISSUE #2477: Modeladmin needs fixing for non-superusers
flake8 currently failing on https://github.com/torchbox/wagtail/blob/master/wagtail/contrib/modeladmin/helpers.py#L182 due to the missing import of UserPagePermissionsProxy, which suggests that that particular code path needs fixing / testing...

gasman:
Fixes #2477

In addition to adding the missing import, I've rewritten the logic to do the filtering by both content type and permission in a single pass. This is because I'm not particularly confident that Django and the database will be smart enough to avoid enumerating `Page.objects.descendant_of(some_high_level_page)` in full, which is potentially a huge data set.

### DIFF
--- a/.scss-lint.yml
+++ b/.scss-lint.yml
@@ -99,6 +99,8 @@ linters:
             - '**/_streamfield.scss'
             - '**/page-editor.scss'
             - '**/_datetimepicker.scss'
+            - 'wagtail/contrib/modeladmin/static_src/modeladmin/scss/choose_parent_page.scss'
+            - 'wagtail/contrib/modeladmin/static_src/modeladmin/scss/index.scss'
 
     PlaceholderInExtend:
         enabled: false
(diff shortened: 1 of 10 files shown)


kaedroho:
Here's the query I get for the old code. Django is pretty smart

```sql
SELECT -- snip
FROM "wagtailcore_page"
WHERE (
    "wagtailcore_page"."content_type_id" IN (/* ids of allowed_parent_page_types */)
    AND (
            -- One filter generated for each page with add permission
            ("wagtailcore_page"."path"::text LIKE 00010002% AND "wagtailcore_page"."depth" >= 2)
            OR ("wagtailcore_page"."path"::text LIKE 00010003% AND "wagtailcore_page"."depth" >= 2)
        )
)
ORDER BY "wagtailcore_page"."path" ASC'
```

gasman:
OK, fair enough :-) That just leaves the tests then - please cherry-pick 447fca1c0c9c9c44cbcdf210776d29ef549997ea if you're happy with it.

kaedroho:
Cherry picked in https://github.com/torchbox/wagtail/commit/b52ac6a1eaf1a304a7758d5578203b82cb0e7a9f

Issue #850, opened 2014-12-04
Project template settings files causes unexpected behavior


sethrh:
In the project template, the django settings are broken out into base, dev, production, etc. 

The `__init__.py` file, however, does this: 

```
from .dev import *
```

This means that when you set the DJANGO_SETTINGS_FILE to 'settings.production', python loads modules something like this: 
- dev.py
- base.py (imported in dev.py)
- local.py (if it exists, imported in dev.py)
- production.py
- base.py (imported again in production.py, but since it's already been imported nothing here will apply)
- local.py (if it exists, but again, it's already been imported so nothing will apply)

This causes unexpected development settings to appear in my production deployments. 

This method of using settings is really complex and the side-effects are subtle (I've probably got the load order wrong).

I suggest refactoring it to a simpler method - such as a settings.py which is a good default for development settings, and then a settings_production.py which imports and overrides the defaults where appropriate.

miracle2k:
The auto-generated dev settings file contains this line:

    for template_engine in TEMPLATES:
        template_engine['OPTIONS']['debug'] = True

Unless I am mistaken, this is the problem described here in action. These changes will leak into the production settings.

gasman:
I believe those particular lines are redundant - it just needs someone to double-check that: #2078

PR #2489, opened 2016-04-18
Don't import dev in project template settings init

toumaz:
This might be as much of a discussion as a PR, but heck, here goes. :)

Developer settings were always imported whenever any portion of the
project template settings package itself was imported. Using production
settings would result in `__init__.py` importing the dev settings first,
*then* overriding those with the production settings.

This lead to somewhat counter-intuitive behaviour - dev settings that
weren't explicitly set again in production would tag along to prod.

My specific use case was appending `debug_toolbar` to `INSTALLED_APPS`
in development, which I don't install at all in production - which had me confused for a while.
With both dev and production explicitly loading base settings, I felt like the *intent* to work
like this was being communicated - but it's all just my opinion, of course. :)

My proposal is to have the settings package no longer import
dev, and use `settings.dev` explicitly in both `manage.py` and `wsgi.py`.

### DIFF
--- a/wagtail/project_template/manage.py
+++ b/wagtail/project_template/manage.py
@@ -5,7 +5,7 @@ import os
 import sys
 
 if __name__ == "__main__":
-    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "{{ project_name }}.settings")
+    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "{{ project_name }}.settings.dev")
 
     from django.core.management import execute_from_command_line
 
--- a/wagtail/project_template/project_name/settings/__init__.py
+++ b/wagtail/project_template/project_name/settings/__init__.py
@@ -1,1 +1,0 @@
-from .dev import *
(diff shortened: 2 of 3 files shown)


timorieber:
Related #850

gasman:
I'm in favour of this change. It seems to have caught a few people out now (both this particular issue with settings getting overridden, and general confusion about how to deploy to production, with people thinking they were supposed to edit `__init__.py` to point to the production settings), and I can't see any real downsides to this alternative approach.

(Incidentally, I wonder if it would make more sense to specify settings.production in wsgi.py? The change here preserves the existing behaviour, which is a good thing of course - but typically WSGI is only ever going to be used in production, right?)

toumaz:
I was thinking about `wsgi.py` myself and it was tempting to suggest, but I recall seeing a few folks and guides advocate use of e.g. gunicorn during development. I'd advocate the principle of least astonishment.

Perhaps it would be a good idea to later also document how the project template settings default to `settings.dev` along with an example of how to specify production settings via `DJANGO_SETTINGS_MODULE`?

alexgleason:
Can confirm this made things confusing when deploying to Azure.

gasman:
Merged in 4bb2bf2bbd4cf90fb4bc3f2aa917ce60d00e65c6 (with release note in f9947b2c0851f95ea36e09bfdcbb652f256bec05) - thanks!

PR #2491, opened 2016-04-18
Set default values if not present

dirtycoder:
If `base_form_class` and `get_edit_handler` are not present during the `Page`'s `check`, both will be set to a default value.

Before this, the `Page`'s model was dependent of an import side effect evoked by the `wagtailforms` app.


gasman:
Thanks @dirtycoder!

@timheap - does this look like an acceptable fix?

gasman:
Closing as per https://github.com/torchbox/wagtail/issues/2428#issuecomment-215690528 - we should fix this by moving those checks into wagtailadmin instead.

PR #2495, opened 2016-04-19
#2281 fix issue when multiple redirects are returned for a path

### LINKED ISSUE #2281: redirects
In [RedirectMiddleware.process_response](https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailredirects/middleware.py#L25) it is possible to encounter a `MultipleObjectsReturned` exception for a request, when there are almost-duplicate redirects, one matching a specific site, and one with `site=None`. This is because it calls `get_for_site`, which ultimately translates to:

```python
Redirect.objects.filter(models.Q(site=site) | models.Q(site=None)).get(old_path='/path')
```

I suggest changing it to something like:

```python
        # Find redirect
        try:
            redirect = models.Redirect.get(site=request.site, old_path=path)
        except models.Redirect.MultipleObjectsReturned:
            redirect = models.Redirect.get(site=None, old_path=path)
        except models.Redirect.DoesNotExist:
            if path == path_without_query:
                # don't try again if we know we will get the same response
                return response
```
i.e. preferring the specific redirect over the generic one, where possible, and ditching use of `get_for_site` (here at least)

nimasmi:
As requested, code change (with minor differences from the sneak preview) plus unit tests.

Fixes #2281

### DIFF
--- a/wagtail/wagtailredirects/middleware.py
+++ b/wagtail/wagtailredirects/middleware.py
@@ -6,6 +6,16 @@ from django.utils.six.moves.urllib.parse import urlparse
 from wagtail.wagtailredirects import models
 
 
+def get_redirect(request, path):
+    try:
+        return models.Redirect.get_for_site(request.site).get(old_path=path)
+    except models.Redirect.MultipleObjectsReturned:
+        # We have a site-specific and a site-ambivalent redirect; prefer the specific one
+        return models.Redirect.objects.get(site=request.site, old_path=path)
+    except models.Redirect.DoesNotExist:
+        return None
+
+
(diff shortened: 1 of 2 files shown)


dirtycoder:
Hey @nimasmi, I believe we have a missing case here.

If the path have a query string the error can still be triggered.

nimasmi:
Thanks @dirtycoder for your contributions on this. Refactored, and extra test cases added.

gasman:
Merged in 139c8a1e76444dae77102de1a222f4a34715bfb7 (master) / 90742f2b5997be6e2f3163bc508fdb9afaa9b9d1 (1.4.x) and parents - thanks!

PR #2496, opened 2016-04-19
allow wagtailforms field choices to be as long as they want

johannes-torchbox:
as discussed, it would be nice to allow choices that are as long as whatever to wagtail, especially if we use the choices for complicated things (like selecting between different courses when signing up for a class)

### DIFF
--- a/wagtail/tests/testapp/migrations/0004_form_choices.py
+++ b/wagtail/tests/testapp/migrations/0004_form_choices.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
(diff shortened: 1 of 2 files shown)


m1kola:
Hi @johannes-torchbox.

We need to provide upgrade instructions because it requires migrations for on implementors side. In general looks good. Thanks!

P.S.: Congratulations on your first PR!

johannes-torchbox:
Oh, yes, that's right. I'll add those today.

Thanks! =)

gasman:
Merged in a2ec49e01a59ace42032d1c027a9475fe25c262e - thanks!

I've added a note about the migration in the release notes. We haven't been particularly rigorous about mentioning these when they've occurred previously (usually on custom image models), but in this case it will affect developers with not-so-heavily-customised projects, so I agree it's worth a special mention.

PR #2497, opened 2016-04-19
including a search query to boost the exact match (if present) for elasticsearch

johannes-torchbox:
In regular elasticsearch queries, an occurrence of a perfect match (for a multi-term search query) is not assigned a higher relevance than any other match.

In this PR I changed the query for the elasticsearch backend so that exact matches are boosted by 1.5, thus appearing higher than regular results.

This helps in places where eg. people pages are indexed and searching for the full name brings up *less* interesting results than searching for the last name only. You can see this on http://www.rca.ac.uk/
when searching for "Teasley" vs. "Sarah Teasley".

### DIFF
--- a/wagtail/wagtailsearch/backends/elasticsearch.py
+++ b/wagtail/wagtailsearch/backends/elasticsearch.py
@@ -243,31 +243,53 @@ class ElasticSearchQuery(BaseSearchQuery):
             fields = self.fields or ['_all', '_partials']
 
             if len(fields) == 1:
-                if self.operator == 'or':
-                    query = {
-                        'match': {
-                            fields[0]: self.query_string,
-                        }
+                query = {
+                    "bool": {
+                        "should": [
+                            {
+                                'match': {
+                                    fields[0]: {
+                                        'query': self.query_string,
+                                        'operator': self.operator,
+                                    }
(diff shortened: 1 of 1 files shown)


kaedroho:
Thanks for the PR Johannes! But it looks like this is an RCA issue.

https://github.com/torchbox/verdant-rca/blob/49a91b91143c46962e66506f491056c2bb18faf9/django-verdant/rca/models.py#L3931-L3932

The student's names are being boosted 50 times, but Sarah Teasley is a staff member so any student called Sarah gets ranked higher. This code was probably written before we changed Wagtail's'default title boost to 2 (down from 100).

I think reducing these boosts to something sensible like 1.5 would solve the issue.

johannes-torchbox:
Hi Karl,

of course, you probably know a lot more about the search in Wagtail than I
do. But is it not a good idea to boost full-phrase matches vs. partial
matches in almost any case? I don't think I can think of an example where a
partial match should be rated higher than a full match. Since this PR
shouldn't change anything else about the search mechanics I still think it
would be worthwhile to look at a way to implement it.

Thanks for the pointer on the RCA boosts, I'll look into them some more
today.

Cheers,
Johannes

PR #2498, opened 2016-04-19
Remove useless get_indexed_objects method from TagSearchable

gasman:
Related: #2436

Indexing the tags of TagSearchable subclasses (such as wagtailimages.Image) is done by following
the model's 'tags' relation, and wagtailsearch is smart enough to prefetch that automatically
(as the test demonstrates). Prefetching 'tagged_items__tag' just adds two useless queries :-)


kaedroho:
Thanks!

PR #2499, opened 2016-04-19
Updated Page.dummy_request() to be able to handle ALLOWED_HOSTS = ['*'].

coredumperror:
Previously, Page.dummy_request() assumed that if `ALLOWED_HOSTS[0]` exists, it's a valid hostname. But since `ALLOWED_HOSTS[0]` can be `'*'`, that led to `400 Bad Request` errors under some circumstances. This PR simply makes Page.dummy_request() treat `'*'` as an invalid hostname.


gasman:
Merged in bd8aedbc718e47ac2d7c1cd252bf99413501b0ee (master) / efceeba24171fd434de8aea149e3e4450d47ae4d (1.4.x), along with test in 220dfb631d536c32a69627511cbf5ba0e2fbbfc0. Thanks!

PR #2501, opened 2016-04-20
Check before unregistering Group model admin

jasonm:
Fixes exception:
    django.contrib.admin.sites.NotRegistered: The model Group is not registered

### DIFF
--- a/wagtail/wagtailcore/admin.py
+++ b/wagtail/wagtailcore/admin.py
@@ -21,5 +21,6 @@ class GroupPagePermissionInline(admin.TabularInline):
 class GroupAdminWithPagePermissions(GroupAdmin):
     inlines = GroupAdmin.inlines + [GroupPagePermissionInline]
 
-admin.site.unregister(Group)
+if admin.site.is_registered(Group):
+    admin.site.unregister(Group)
 admin.site.register(Group, GroupAdminWithPagePermissions)


jasonm:
For minervaproject reference I'm going to point to https://github.com/minervaproject/wagtail/pull/1 as a fork off 1.3.x until we update to 1.4.x but this is a PR against the upstream master.

jasonm:
I encountered this exception while going through a Django 1.8 -> 1.9 upgrade along with Wagtail 1.2 -> 1.3.  Not sure why this affects us -- that is, why we otherwise did not have the `Group` model registered.  We do have a custom User model, but I don't know if that is realted?

I do notice that from Django 1.8 to Django 1.9, `django/contrib/auth/admin.py` changed from explicitly calling `admin.site.register(Group, GroupAdmin)`:

https://github.com/django/django/blob/stable/1.8.x/django/contrib/auth/admin.py#L191

to adding a `@admin.register(Group)` decorator to the `django.contrib.auth.admin.GroupAdmin` class:

https://github.com/django/django/blob/stable/1.9.x/django/contrib/auth/admin.py#L27

I'm not familiar, though, with what that change implies under the hood or if it's even pertinent.

gasman:
Merged in 35c283fa2d7809c6d01f12784cd91398447cdc74 (master) and 307364e3cd960c1c980bfc39a8a688d7096da7ed (1.4.x) - thanks!

jasonm:
Thanks @gasman !

PR #2507, opened 2016-04-23
Initial work on the Admin API

kaedroho:
This pull request implements the admin API, based on the v2 API with the following changes:

 - The API is only visible to logged in users
 - All pages from all sites visible in the same API (except for the root page)
 - All draft pages are visible
 - ``status`` meta field added to indicate the state of each page
 - ``children`` field added giving a count of the number of children and a link to the listing view for them
 - optional ``descendants`` field added giving a count of the number of descendants and a link to the listing view for them
 - ``child_of`` and ``descendant_of`` can be used on any page, including root (by passing it a special value: ``root``)
 - Appends ``__types`` to each response, giving information about all the types used in the response
 - Adds ``latest_revision_created_at``field to ``meta``
 - Adds ``has_children`` filter
 - Adds ``parent`` field to listings

There's some major features planned for the near future:
 - ~~Splitting it out so the implementation lives in the correct app for each content type (eg, the images endpoint would be moved to the ``wagtailimages`` app~~
 - Adding the ability to retrieve page revisions and save new revisions
 - Move some actions to the API, eg page locking

### DIFF
--- a/wagtail/api/v2/serializers.py
+++ b/wagtail/api/v2/serializers.py
@@ -40,7 +40,9 @@ class TypeField(Field):
         return instance
 
     def to_representation(self, obj):
-        return type(obj)._meta.app_label + '.' + type(obj).__name__
+        name = type(obj)._meta.app_label + '.' + type(obj).__name__
+        self.context['view'].seen_types[name] = type(obj)
+        return name
 
 
 class DetailUrlField(Field):
@@ -95,7 +97,9 @@ class PageTypeField(Field):
         return instance
 
     def to_representation(self, page):
-        return page.specific_class._meta.app_label + '.' + page.specific_class.__name__
+        name = page.specific_class._meta.app_label + '.' + page.specific_class.__name__
+        self.context['view'].seen_types[name] = page.specific_class
+        return name
 
 
 class RelatedField(relations.RelatedField):
--- a/wagtail/wagtailadmin/api/endpoints.py
+++ b/wagtail/wagtailadmin/api/endpoints.py
@@ -2,14 +2,14 @@ from __future__ import absolute_import, unicode_literals
 
 from collections import OrderedDict
 
-from wagtail.api.v2.endpoints import DocumentsAPIEndpoint, ImagesAPIEndpoint, PagesAPIEndpoint
+from wagtail.api.v2.endpoints import PagesAPIEndpoint
 from wagtail.api.v2.filters import (
     ChildOfFilter, DescendantOfFilter, FieldsFilter, OrderingFilter, SearchFilter)
 from wagtail.api.v2.utils import BadRequestError, filter_page_type, page_models_from_string
 from wagtail.wagtailcore.models import Page
 
 from .filters import HasChildrenFilter
-from .serializers import AdminImageSerializer, AdminPageSerializer
+from .serializers import AdminPageSerializer
 
 
 class PagesAdminAPIEndpoint(PagesAPIEndpoint):
@@ -90,22 +90,3 @@ class PagesAdminAPIEndpoint(PagesAPIEndpoint):
         response = super(PagesAdminAPIEndpoint, self).detail_view(request, pk)
         response.data['__types'] = self.get_type_info()
         return response
-
-
-class ImagesAdminAPIEndpoint(ImagesAPIEndpoint):
-    base_serializer_class = AdminImageSerializer
-
-    extra_body_fields = ImagesAPIEndpoint.extra_body_fields + [
-        'thumbnail',
-    ]
-
-    default_fields = ImagesAPIEndpoint.default_fields + [
-        'width',
-        'height',
-        'thumbnail',
-    ]
-
-
-
-class DocumentsAdminAPIEndpoint(DocumentsAPIEndpoint):
-    pass
--- a/wagtail/wagtailadmin/api/endpoints.py
+++ b/wagtail/wagtailadmin/api/endpoints.py
@@ -1,0 +1,110 @@
+from __future__ import absolute_import, unicode_literals
+
+from collections import OrderedDict
+
+from wagtail.api.v2.endpoints import DocumentsAPIEndpoint, ImagesAPIEndpoint, PagesAPIEndpoint
+from wagtail.api.v2.filters import (
+    ChildOfFilter, DescendantOfFilter, FieldsFilter, OrderingFilter, SearchFilter)
+from wagtail.api.v2.utils import BadRequestError, filter_page_type, page_models_from_string
+from wagtail.wagtailcore.models import Page
+
+from .filters import HasChildrenFilter
+from .serializers import AdminImageSerializer, AdminPageSerializer
+
+
+class PagesAdminAPIEndpoint(PagesAPIEndpoint):
+    base_serializer_class = AdminPageSerializer
+
+    # Use unrestricted child_of/descendant_of filters
+    # Add has_children filter
+    filter_backends = [
+        FieldsFilter,
+        ChildOfFilter,
+        DescendantOfFilter,
+        HasChildrenFilter,
+        OrderingFilter,
+        SearchFilter,
+    ]
+
+    extra_meta_fields = PagesAPIEndpoint.extra_meta_fields + [
+        'latest_revision_created_at',
+        'status',
+        'children',
+        'parent',
+    ]
+
+    default_fields = PagesAPIEndpoint.default_fields + [
+        'latest_revision_created_at',
+        'status',
+        'children',
+    ]
+
+    known_query_parameters = PagesAPIEndpoint.known_query_parameters.union([
+        'has_children'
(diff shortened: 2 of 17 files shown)


wagtail/api/v2/serializers.py
```diff
@@ -40,7 +40,9 @@ def get_attribute(self, instance):
         return instance
 
     def to_representation(self, obj):
-        return type(obj)._meta.app_label + '.' + type(obj).__name__
+        name = type(obj)._meta.app_label + '.' + type(obj).__name__
+        self.context['view'].seen_types[name] = type(obj)
```
m1kola: I think serializers / fields shouldn't "know" about view from point of view of semantics. Serializers can (theoretically) be used on only with views (in workers for example).

But at the moment I don't see a better solution. I will think about it.

kaedroho: Yep, I agree. There isn't a nice way around this at the moment. Serializers currently also need to be aware of other endpoints in order to support cross-linking between them.

I've started to address this in https://github.com/torchbox/wagtail/pull/2484. Instead of allowing serialisers to make nested serialisers, we would generate all serialisers at the view level (and pass nested serialisers into their parents). This would also mean that all the types used in the API would be available in the view so this bit of code wouldn't be necessary

wagtail/wagtailadmin/api/endpoints.py
```diff
@@ -0,0 +1,111 @@
+from __future__ import absolute_import, unicode_literals
+
+from collections import OrderedDict
+
+from wagtail.api.v2.endpoints import DocumentsAPIEndpoint, ImagesAPIEndpoint, PagesAPIEndpoint
+from wagtail.api.v2.filters import (
+    ChildOfFilter, DescendantOfFilter, FieldsFilter, OrderingFilter, SearchFilter)
+from wagtail.api.v2.utils import BadRequestError, filter_page_type, page_models_from_string
```
m1kola: Interesting. I found two definitions of `filter_page_type` and `page_models_from_string`:
* in `wagtail.api.v2.utils`
* in `wagtail.wagtailadmin.views.chooser`

And they are absolutely the same. It's not a subject of this PR, but... Is there any reason to have a copy?

kaedroho: If we're going to merge them both, it should be moved to wagtailcore, maybe it could be merged into the ``.type()`` queryset method somehow.

wagtail/wagtailadmin/api/endpoints.py
```diff
@@ -0,0 +1,111 @@
...
+        # TODO: Add "include_root" flag
+        queryset = queryset.exclude(depth=1)
+
+        return queryset
+
+    def get_type_info(self):
+        types = OrderedDict()
+
+        for name, model in self.seen_types.items():
+            types[name] = OrderedDict([
```
m1kola: It seems to me that `OrderedDict` is unnecessary here because this dict will be converted to JSON object. JSON itself does not guarantee order of of properties in object. The same is in other places.

kaedroho: > JSON itself does not guarentee order of properties in object

It does, it's a serialised format. Bytes can't physically move around until the JSON is deserialised at the other end (then it's up to the developer to decide whether they want to maintain the ordering).

I use ``OrderedDict`` everywhere mainly to aid developers who are checking responses manually as fields won't randomly move around from request to request.

m1kola: > It does, it's a serialised format. Bytes can't physically move around

Yes. It's right, but [JSON specification](https://tools.ietf.org/html/rfc7159) says:
```
An object is an unordered collection of zero or more name/value
pairs, where a name is a string and a value is a string, number,
boolean, null, object, or array.
```

> I use `OrderedDict` everywhere mainly to aid developers who are checking responses manually as fields won't randomly move around from request to request.

Thanks for explanation. It makes sense. Ignore this comment.

wagtail/wagtailadmin/tests/api/test_documents.py
```diff
@@ -0,0 +1,115 @@
...
+
+        # Check that the items section is there
+        self.assertIn('items', content)
+        self.assertIsInstance(content['items'], list)
+
+        # Check that each document has a meta section with type, detail_url and tags attributes
+        for document in content['items']:
+            self.assertIn('meta', document)
+            self.assertIsInstance(document['meta'], dict)
+            self.assertEqual(set(document['meta'].keys()), {'type', 'detail_url', 'download_url', 'tags'})
```
m1kola: You can use `assertSetEqual` ;)

m1kola:
We need to keep #2463 in mind. When it will be implemented we will need to make changes in Admin API.

m1kola:
Looks good! Thanks, @kaedroho!

kaedroho:
> Splitting it out so the implementation lives in the correct app for each content type (eg, the images endpoint would be moved to the wagtailimages app)

Just thought that this might be a blocker. I've not yet tested this branch without wagtaildocs/wagtailimages installed

kaedroho:
RFC for the restructure: https://github.com/torchbox/wagtail-rfcs/pull/8

kaedroho:
Rebased on top of #2649

m1kola:
Rebased on top of `master`

wagtail/wagtailadmin/api/endpoints.py
```diff
@@ -0,0 +1,92 @@
...
+        if len(models) == 1:
+            queryset = models[0].objects.all()
+        else:
+            queryset = Page.objects.all()
+
+            # Filter pages by specified models
+            queryset = filter_page_type(queryset, models)
+
+        # Hide root page
+        # TODO: Add "include_root" flag
```
m1kola: Are you going to do this in a separate PR?

kaedroho: Yep

wagtail/wagtailimages/api/admin/serializers.py
```diff
@@ -0,0 +1,53 @@
...
+                ('height', thumbnail.height),
+            ])
+        except SourceImageIOError:
+            return OrderedDict([
+                ('error', 'SourceImageIOError'),
+            ])
+
+
+class AdminImageSerializer(ImageSerializer):
+    thumbnail = ImageRenditionField('max-165x165', read_only=True)
```
m1kola: We also need a `max-800x600` for focal point feature. Probably we need a list/object of thumbnails.

m1kola: We also need a link to the original image to be able to render the "file" field:
screen shot 2016-07-06 at 19 36 27

m1kola: > We also need a max-800x600 for focal point feature. Probably we need a list/object of thumbnails.

Hm... If we add the list/object of thumbnails, we will need an ability to get only required sizes. For example, at the listing view we don't need all possible resolutions: we need only one, but at the detailed view we need a different size.

wagtail/wagtaildocs/api/admin/endpoints.py
```diff
@@ -0,0 +1,7 @@
+from __future__ import absolute_import, unicode_literals
+
+from ..v2.endpoints import DocumentsAPIEndpoint
+
+
+class DocumentsAdminAPIEndpoint(DocumentsAPIEndpoint):
+    pass
```
m1kola: It seems to me that we also need the following fields in the Admin API:
* `created_at` - required for list ordering and for "Uploaded" column on the list view
* `filesize` - required for the detail view

The same fields we need for images.

m1kola:
Probably we need to handle issues listed in my comments above in a separate PRs. What do you think, @kaedroho?

In general, looks good to me. Thanks!

m1kola:
I forgot to mention in my previous comment that when we finish this PR and solve those issues, we will be able to start implementation of at least the following views using the Admin API:
* Documents: listing and detail views 
* Images: listing and detail views (except "Focal point" feature)

wagtail/wagtailimages/api/admin/serializers.py
kaedroho: > We also need a max-800x600 for focal point feature. Probably we need a list/object of thumbnails.

I don't think there is any plan on implementing the edit view on the admin API yet. if we do, I'd rather add to the admin API at the point of creating it.

kaedroho:
Thanks @m1kola!

I think we should extend the admin API at the point of implementing new components onto it. For now, it is only going to power the new explorer interface and it is complete enough for that. I have put a little thought into what we need to reimplement the choosers as well (which I expect will be the next thing), which is why images have thumbnails (but we probably will need to extend the admin api some more anyway).

m1kola:
Merged in 2f8d00222dfe7f028c5a8375729db3b92b77dc27.

Thanks @kaedroho!

PR #2508, opened 2016-04-23
Make sure the default search backend is always defined

kaedroho:
It's common in Wagtail to want to quickly override the base search settings to disable Elasticsearch in specific environments (eg CI or running imports).

To do this, you have to manually write out Wagtail's default search configuration.

This commit changes the way the default configuration is loaded, it is now loaded whenever there is no "default" backend configured rather than only loading if the ``WAGTAILSEARCH_BACKENDS`` was not defined at all.

To override a parent settings file's search backends configuration, you can now just do:

```python
WAGTAILSEARCH_BACKENDS = {}
```

And the defaults will be restored

### DIFF
--- a/wagtail/wagtailsearch/backends/__init__.py
+++ b/wagtail/wagtailsearch/backends/__init__.py
@@ -15,6 +15,17 @@ class InvalidSearchBackendError(ImproperlyConfigured):
     pass
 
 
+def get_search_backend_config():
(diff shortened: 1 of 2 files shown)


gasman:
Merged in f1913bd748360b437100209caf0477baf2955bc0 - thanks!

PR #2509, opened 2016-04-24
Added a delete Button on the User page #2371

hamsterbacke23:
Hello I am new to Django / Wagtail,
and had a go at adding that delete-button on the user page.
Not entirely sure yet about the permissions...

### DIFF
--- a/wagtail/wagtailusers/forms.py
+++ b/wagtail/wagtailusers/forms.py
@@ -188,6 +188,10 @@ class UserEditForm(UsernameForm):
             self.save_m2m()
         return user
 
+    def delete(self, commit=True):
+        user = super(UserEditForm, self).delete(commit=False)
+        return user
+
 
 class GroupForm(forms.ModelForm):
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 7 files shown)


SableWalnut:
I was under the impression it was intentional that the only way to 'delete' a user was to make them inactive. Things would get tricky in relation to page revisions if a user is suddenly deleted.

hamsterbacke23:
I see...
As I understood from #2371, deleting a user was no problem, may be I am wrong.

gasman:
Yep, the `user` field on page revisions is nullable (and set to `on_delete=models.SET_NULL`), so deleting the user will just make their revisions 'anonymous' - which is valid at the model level. (We probably ought to test that this gets rendered sensibly on views, but that's a separate issue...)

Hopefully the same is true on any other models that have the concept of being owned by a user, but someone ought to double-check that :-)

(Interesting point about deactivation though - if this isn't what the `is_active` flag is designed for, then what *is* it for? :-) However, I think we can take our lead from django-admin here - django-admin doesn't take any measures to block the 'delete' action in favour of deactivation, so neither should we.)

gasman:
Thanks @hamsterbacke23! We've received several other PRs that implement this, and I've chosen to go with #2523 as it's the most complete - you'll still get a credit in the release though :-)

hamsterbacke23:
Ok thanks!

PR #2510, opened 2016-04-24
Fixes and improvements for #2444

m1kola:
Replaces #2444.

I have found some more issues with translatable strings in page revisions functionality. Fixed them in https://github.com/torchbox/wagtail/commit/660bbdfce92a9d52e30ca91c09d60ccdd6505fb1.

Also I have fixed tests for revisions in https://github.com/torchbox/wagtail/commit/edb924acf49450eac77c2e28a57ae23571115ca8. They were broken by removing date format in @jaroel's PR. But I agree with Roel that we need to use locale formats for dates / times.

@jaroel thanks for your work!

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -22,7 +22,7 @@ Changelog
  * Added `--schema-only` option to `update_index` management command
  * Added meaningful default icons to `StreamField` blocks (Benjamin Bach)
  * Added title text to action buttons in the page explorer (Liam Brenner)
- * Improved L10N and I18N for revisions list. (Roel Bruggink)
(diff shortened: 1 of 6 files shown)


gasman:
Merged in ac59fc814c9c6a1cbc624171bbe5191848978639 - thanks @jaroel and @m1kola!

The translation of "by" probably isn't ideal for translators, even with the context hint - giving them a string like `"{revision} by {user}"` would be more helpful I suspect. But that would probably need a lot of fiddling around with the HTML tags to make it work, and this solution is clearly better than not having the translation at all - so happy to go with this version!

PR #2518, opened 2016-04-26
Fix $prop variable in userbar.scss not being declared as global

kapito:
Original userbar.scss file left $prop as a local variable, causing machines with certain interpreters to not load the file properly (resulting in strange CSS styling when using Wagtail's interface). Adding !global to the ends of the lines that declare $prop fixes this.

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/userbar.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/userbar.scss
@@ -298,11 +298,11 @@ $positions: (
             @for $i from 1 through $max-items {
 
                 @if $vertical == 'bottom' {
-                    $prop: 'nth-last-child';
+                    $prop: 'nth-last-child' !global;
                 }
 
                 @if $vertical == 'top' {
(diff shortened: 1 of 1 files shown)


gasman:
This is affecting new installations of https://github.com/torchbox/vagrant-wagtail-develop - evidently something has changed recently in our npm dependencies to break this. As far as I can tell, the relevant breaking change happened in Sass 3.3:

http://sass-lang.com/documentation/file.SASS_CHANGELOG.html#backwards_incompatibilities__must_read_2

> As part of a migration to cleaner variable semantics, assigning to global variables in a local context by default is deprecated. If there’s a global variable named $color and you write $color: blue within a CSS rule, Sass will now print a warning; in the future, it will create a new local variable named $color. You may now explicitly assign to global variables using the !global flag; for example, $color: blue !global will always assign to the global $color variable.

Now trying to track down exactly which package is responsible for the change (and should probably be version-pinned to avoid sudden breakages like this in future...)

gasman:
Merged in eecec1b7cee645e65fc454929a54663875ec3fc9 - thanks!

PR #2521, opened 2016-04-26
Useless return in wagtail.wagtailforms.forms.BaseForm

m1kola:
Looks like `wagtail.wagtailforms.forms.BaseForm` have a useless return in `__init__` method.


kaedroho:
Thanks!

PR #2523, opened 2016-04-27
#2371 - Add ability to delete users

vincentaudebert:
Possibility to delete a user + just rebased with most recent version of master - #2371

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -32,7 +32,8 @@ Changelog
  * Fix: Direct usage of `Document` model replaced with `get_document_model` function in `wagtail.contrib.wagtailmedusa` and in `wagtail.contrib.wagtailapi`
  * Fix: Failures on sending moderation notification emails now produce a warning, rather than crashing the admin page outright (Matt Fozard)
  * Fix: All admin forms that could potentially include file upload fields now specify `multipart/form-data` where appropriate (Tim Heap)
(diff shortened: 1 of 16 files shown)


gasman:
Thanks! This is actually the third PR to address this (after #2402 and #2509) but I'll keep this one as it's the most complete :-)

Just a couple of issues:

* I don't think we should use hooks for registering the listing buttons - that's over-complicating things. The `user_listing_buttons` tag should just build the list of buttons directly instead. I can see the value of making it behave like the page listing - but if we're going to do that, then it really needs to be refactored to reduce the amount of copy-and-paste between the two apps.
* The permission logic is a bit hard to follow, due to it being split between views/users.py and utils.py. I'd suggest just making it into a single function in utils.py (which means repeating the definition of delete_user_perm there, but I think that's a lesser evil...):


        def user_can_delete_user(current_user, user_to_delete):
            if not current_user.has_perm(delete_user_perm):
                return False

            if current_user == user_to_delete:
                # users may not delete themselves
                return False

            if user_to_delete.is_superuser and not current_user.is_superuser:
                # ordinary users may not delete superusers
                return False

            return True

* There should be some extra tests to cover the special cases in the permission logic - i.e. to test that a user cannot delete themselves, and that an ordinary user can't delete a superuser.

gasman:
Merged (with cleanup / extra tests) in 14919f3b41da8d089c870975bcd0364af6176d59 - thanks!

PR #1435, opened 2015-06-23
Adds wagtail ui template tag for making buttons.

jonnyscholes:
This proof of concept PR starts the process of completing #382. It moves the existing buttons (excluding dropdown buttons) to a template tag. Currently I have only included them in the style guide - if the implementation is deemed sound I will convert the entire back end to use these buttons. Then I will start working through converting the other elements.

The parameter names in the template tag could change - I wanted to use `type` instead of `kind` but there is already a class called `type`.

It introduces a new template tag namespace `wui`. To make a button with the new template tag do something like:

```
{% wuibutton kind="secondary" element="button" icon="tick" scheme="yes" %}
    Yes please!
{% endwuibutton %}
```

Which would produce:

``` html
<button class="button-secondary yes icon icon-tick text-replace">Yes please!</button>
```

Everything has (hopefully) sensible defaults and you can set arbitrary HTML attributes on the button element. So the following:

```
{% wuibutton href="http://wagtail.io" aria-hidden="true"  %}
    Wagtail Website
{% endwuibutton %}
```

Produces:

``` html
<a class="button" href="http://wagtail.io" aria-hidden="true">Wagtail Website</a>
```

cc @timheap @precise54

PR #2524, opened 2016-04-27
Add button class to all inputs. Fixes #2512

### LINKED ISSUE #2512: Broken styles for some buttons on dashboard
**Steps to reproduce:**
1. Open any page and make some changes
2. Submit the page for moderation
3. Go to the dashboard
3. Put cursor on a title of the page in "pages awaiting moderation" section

**Result:**
screen shot 2016-04-25 at 17 55 23

**Expected result:**
screen shot 2016-04-25 at 18 03 28

JoshBarr:
Adds button class to all inputs, thereby removing the need for attribute selectors, and ensuring `.button` class and any modifiers work appropriately.

Fixes #2512.

### DIFF
--- a/wagtail/contrib/modeladmin/templates/modeladmin/create.html
+++ b/wagtail/contrib/modeladmin/templates/modeladmin/create.html
@@ -30,7 +30,7 @@
                     <li class="actions">
                         {% block form_actions %}
                             <div class="dropdown dropup match-width">
-                                <div><input type="submit" value="{% trans 'Save' %}" /></div>
+                                <div><input type="submit" value="{% trans 'Save' %}" class="button" /></div>
                             </div>
                         {% endblock %}
                     </li>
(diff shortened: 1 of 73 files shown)


gasman:
Thanks @JoshBarr! I see a lot of buttons on the styleguide page have lost their styling now - I guess they just need the `button` class adding too?

(I guess this change is going to affect a lot of third-party Wagtail add-ons. Probably a necessary evil, though... :-/ )

JoshBarr:
Yeah it's probably something we need to communicate clearly in the release notes for 1.5... I couldn't come up with a non-hacky way of fixing that specificity issue. We're going to run into these visual regressions across third party modules if we touch/refactor any of the CSS. Which means we're a little bit limited with how much we can improve this part of the project without creating lots of work for existing devs. 

Using an abstraction, like `{% include 'button.html' with type='submit' style='primary' attrs='{"foo":"bar"}'  %}` could help? There's plenty of repeated code in Wagtail's templates that could be DRYed up. 

We could then possibly PR against all the popular open source modules to make it a bit less painful for the community?

gasman:
@JoshBarr I agree, an abstraction for buttons would really help here. @jonnyscholes came up with a nice approach for this in #1435 that we should really bring back up to date!

gasman:
Merged (with upgrade consideration note added) in 91e81d6bc04eccc5dbcf1d760a9ebd912e6879a8 + parents.

Issue #2516, opened 2016-04-26
enableDirtyFormCheck function doesn't take into account keyboard shortcuts


ptink:
The function in wagtailadmin/core.js that raises a warning if the user attempts to navigate away from the page without saving does not understand that if the keyboard shortcut is used to (e.g.) save draft, the $trigger for the event is the currently focused element, and not the actual submit button. This means that if a user edits the page and then uses ctrl+s to save, a warning pops up saying "Are you sure you want to leave the page?" Which is confusing and defeats the purpose of a shortcut .

tomdyson:
Thanks for the report, @ptink. Is this something you could help with?

gasman:
Fixed in #2526

PR #2526, opened 2016-04-27
Fix keyboard shortcuts like Ctrl+S giving unnecessary warnings about leaving the page

kapito:
As described by @ptink in #2516, when saving a modified page with Ctrl+S it gives a warning confirming that you really want to leave the page. This is unnecessary as the user does not actually leave the page -- it just gets refreshed (and the edits get saved in doing this). This is easily fixed by disabling the enableDirtyFormCheck() function from running if triggered by a keyboard shortcut.

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
@@ -53,7 +53,7 @@ function initTagField(id, autocompleteUrl) {
  *      prompting the user even when nothing has been changed.
 */
 
-var enableDirtyFormCheck = 1;
+var enableDirtyFormCheck = true;
 
 function enableDirtyFormCheck(formSelector, options) {
     var $form = $(formSelector);
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
@@ -53,7 +53,7 @@ function initTagField(id, autocompleteUrl) {
  *      prompting the user even when nothing has been changed.
 */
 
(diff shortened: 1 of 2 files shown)


wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
```diff
@@ -52,6 +52,9 @@ function initTagField(id, autocompleteUrl) {
  *    - alwaysDirty - When set to true the form will always be considered dirty,
  *      prompting the user even when nothing has been changed.
 */
+
+var canGiveWarning = 1;
```
kaedroho: Could this be renamed to something like ``disableDirtyFormCheck``? (I think ``canGiveWarning`` is a bit to generic)

kapito: Changed the variable name to 'enableDirtyFormCheck' and made it us true/false instead of 1/0.

kapito: Changed it from enableDirtyFormCheck to canDirtyFormCheck to resolve conflictions with the function name itself.

gasman:
Merged in e59210ba393a458da145a9bbc402d6a8c3708e3c (master) and cc7e6196c5e60666ae4b7ed716b54b29d809b105 (1.4.x) - thanks!

Have made a small stylistic tweak in 4c96595d9035c2c00e2e2b0e1a89f879abcba73c - there's now a `disableDirtyFormCheck` function, which doesn't do very much, but it means that we're not relying on sharing the global `canDirtyFormCheck` variable between two different scripts (which can make the code a bit hard to follow).

PR #2528, opened 2016-04-27
Display username in password reset email

### LINKED ISSUE #2240: "I forgot my password" doesn't remind you what your username is
The flow is like this:

1. Failing to log in
2. Click "I forgot my password"
3. It has you enter your _email address_
4. It emails you a link to create a new password

Never in this does it help you if you forgot your username, or indicate what your username is.

I think the solution is to also include the username in the confirmation email.

EDIT: Alternatively, disallow users from creating usernames with the @ symbol in them, and allow people to log in using their email addresses.

gasman:
Fixes #2240

Add "Your username (in case you've forgotten):" to the bottom of the password reset email (but only if there's actually a username field that's distinct from email).

### DIFF
--- a/wagtail/tests/testapp/fixtures/test.json
+++ b/wagtail/tests/testapp/fixtures/test.json
@@ -661,6 +661,23 @@
         "email": "admin_only_user@example.com"
     }
 },
+{
+    "pk": 5,
+    "model": "customuser.emailuser",
+    "fields": {
+        "first_name": "",
+        "last_name": "",
+        "is_active": true,
+        "is_superuser": false,
+        "is_staff": false,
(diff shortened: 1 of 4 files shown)


gasman:
Just noticed that Django's default template does this (and reveals that it can be done without a custom template tag after all...)

https://raw.githubusercontent.com/django/django/master/django/contrib/admin/templates/registration/password_reset_email.html

kaedroho:
Merged in 3a6018e7d4ee3ccae75b71679a982050f4d4760c then tweaked in fdfbe689b4cd93cc3e35baa0fe4fd21187e587fc

Thanks!

Issue #2519, opened 2016-04-26
Explorer menu blends in over StreamField


alexgleason:
![screenshot from 2016-04-26 12 06 21](https://cloud.githubusercontent.com/assets/3639540/14825312/55bb2534-0ba7-11e6-8d0e-6e9f26063ca1.png)

From the left side it looks like there's a shadow to distinguish StreamFields from the left sidebar. But when the explorer is expanded, the menu blends into the StreamField. Maybe it should have a shadow under it too, or a border of some sort.

Issue #555, opened 2014-08-19
Search: Make full matches of terms rank higher than partial matches


kaedroho:
http://stackoverflow.com/questions/15673324/elasticsearch-ordering-exact-before-partial-matches

We should add another argument to SearchField called ``partial_boost``. This allows developers to set the boost of partial matches to something lower than full matches.

By default, ``partial_boost`` should be set to something lower than ``boost`` (maybe divide it by 2?).

Implementation shouldn't be too difficult. We would apply the ``boost`` where the field is indexed in ``_all`` and ``partial_boost`` where the field is indexed in ``_partials``.

PR #2531, opened 2016-04-27
Fixing #2519

vincentaudebert:
Stream field background is now lighter grey and icons are white + remove box-shadow to flatten a bit the UI. #2519

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/_variables.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/_variables.scss
@@ -44,6 +44,7 @@ $color-grey-2: #666;
 $color-grey-3: #d9d9d9;
 $color-grey-4: #e6e6e6;
 $color-grey-5: #fafafa;
+$color-grey-6: #777;
 
 $color-thead-bg: $color-grey-5;
 $color-header-bg: $color-teal; // #ff6a58;
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
@@ -211,9 +211,8 @@ li.sequence-member {
 // Menu of other blocks to be added at each position 
 .stream-menu {
     @include transition(all 0.2s ease);
-    box-shadow: inset 0 0 45px rgba(0, 0, 0, 0.3);
     position: relative;
-    background-color: $color-grey-1;
+    background-color: $color-grey-6;
     opacity: 1;
     z-index: 5;
 
@@ -281,7 +280,7 @@ li.sequence-member {
         @include transition(all 0.2s ease);
         background-color: transparent;
         border: 0;
-        color: darken($color-grey-3, 5%);
+        color: $color-white;
         height: auto;
         display: block;
         width: 100%;


tomdyson:
Thanks @vincentaudebert! Could you please share a screenshot as a comment to this PR?

vincentaudebert:
screen shot 2016-04-28 at 11 14 06 am

Here you go @tomdyson :)

gasman:
Not really keen on this fix, I'm afraid - the streamfield background is a bit _too_ light now (`#fff` on `#777` is insufficient contrast, according to http://webaim.org/resources/contrastchecker/) and this feels like a very dramatic UI change to fix a pretty obscure problem...

vincentaudebert:
All good. I had a quick chat with @joshbarr and we came up with this solution. But I understand your reasons. 
I had played a bit with the explorer border too. I could give it another try even if it wasn't that clear and the explorer menu is pretty messy in terms of CSS.

tomdyson:
Is there an intermediate shade which has enough contrast with the icons / text but is still distinguished from the explorer bar?

dirtycoder:
Why not add a small shadow, using the Wagtail primary color?
This doesn't looks too bad, at least for my taste.

Something like:
```
-webkit-box-shadow: 2px 2px 1px -1px rgba(67,177,176,1);
-moz-box-shadow: 2px 2px 1px -1px rgba(67,177,176,1);
box-shadow: 2px 2px 1px -1px rgba(67,177,176,1);
```

shadow

gasman:
Again I'm not really keen on making a conspicuous design change to fix a bug that very few people will ever notice :-)

What if we add the shadow, but make it semi-transparent white (something like `rgba(255, 255, 255, 0.3)`)? That way, the difference will be hardly noticeable when viewed on the normal white background, but will show up as light grey here.

dirtycoder:
@gasman I think that would be even better. :)
For comparison:

White bg:
shadow

StreamField bg:
shadow2

alexgleason:
I agree with the white background; it's what I initially had in mind as well.

It should have at least a faintly visible border on the top, too, I think.

alexgleason:
screenshot from 2016-04-28 11 45 02

screenshot from 2016-04-28 11 46 26


```css
body.explorer-open .explorer:after {
  content: '';
  width: calc(100% - 2px);
  height: 100%;
  position: absolute;
  top: 0;
  right: 0;
  z-index: -1;
  box-shadow: 0px 0px 2px rgba(255, 255, 255, 0.9);
}
```

Here's my proposed solution.

JoshBarr:
Appreciate all the suggestions here folks.

I think we're on the right track with the lighter streamfield background. But I agree with @gasman that `#777` is too light and fails accessibility. We'll try something in between, like `#555`. 

One of the criticisms I have of the Wagtail UI is that the chrome and colour scheme are very heavy/dark/dense. This isn't conducive to creating an elegant writing environment (see iA's [Writer](https://ia.net/writer/mac/) app!), and we need to do some work to soften the interface. There's a bigger overhaul coming: 

Waggy UI

SalahAdDin:
@JoshBarr Tell us more please.

gasman:
Have applied @alexgleason's fix in 796239b7d991322632da6f9f63fd1ccd172d3b58, as that's the most direct fix to the current issue. Looking forward to the design overhaul though!

PR #2532, opened 2016-04-28
Indentation fixes for example code in the docs

coredumperror:
I converted all the sample code in the docs to use 4-space indentation.

In addition, I noticed while building the docs that the Wagtail API page on Usage was triggering a ton of lexer errors. The problem was that the HTML header info at the top of the `..code-block:: json` blocks is not valid json, so the lexer was failing, preventing it from syntax highlighting the json. Since the highlighting was broken anyway, I figured the simplest fix was to just use `..code-block:: text` instead.

Finally, I also removed all uses of `STATIC_URL` (there were 3 of them), as its direct use has been unofficially deprecated in Django since the introduction of the `staticfiles` app. It's always more appropriate to use the `{% static %}` template tag, or it's equivalent pure-python function `django.contrib.staticfiles.templatetags.staticfiles.static()`. If your site uses a custom storage engine through the `STATICFILES_STORAGE` setting (e.g. S3, or some other CDN), the raw value of `STATIC_URL` isn't going to be the correct prefix to your static file URLs.


gasman:
Merged in 7523d7f57f9e2498cf09f03ac7f5fa796895b190 + parents - thanks!

coredumperror:
Hmm, what does "Closed with unmerged commits. This pull request is closed, but the coredumperror:docs-indent-fix branch has unmerged commits" mean? I'm guessing that's normal and expected, but my lack of experience makes it seem scary.

gasman:
No worries, that's just a side effect of the way I merged the PR. If I'd merged it directly, it would have been accompanied by a 'merge' commit (like this one: 581be6a266666f58cc974c36e759ce7279d865f5), and we try to avoid those, to keep Wagtail's commit history clean. So, before merging it in, I rebased it on top of latest master - effectively rebuilding it as if it had been written today rather than yesterday.

Unfortunately, Github isn't smart enough to recognise the resulting commits as the same ones you made here, so it thinks that I've just closed this PR without merging it...

coredumperror:
Ahh, ok. That makes sense. I also prefer to rebase, when I can, for the same reason.

Issue #2513, opened 2016-04-25
Clicking on link in document chooser popup triggers "you have unsaved changes" message


m1kola:
The listing in the document chooser popup has a link to the file, clicking on it tries to download the file, and triggers the "you have unsaved changes" message.

Related to #2160.

vincentaudebert:
@m1kola Can you give more details on how to reproduce bug? I can't find a way to reproduce it.
How do you access to the document chooser popup? Stream field?
Also what browser are you using?

Cheers.

kapito:
@vincentaudebert When editing a page in Wagtail, scroll to the bottom and you should see 'Related Links' next to a plus sign. Click this plug sign and click on 'Choose a Document' to open the Document chooser. When attempting to click on one of these links it will give the 'Are you sure you want to leave?' warning, which is unnecessary in this particular situation.

m1kola:
@kapito Thanks!

@vincentaudebert "how to reproduce" depends on your implementation. I think you can reproduce it with any page type that have RichTextField or page chooser panel. If you use [`wagtaildemo`](https://github.com/torchbox/wagtaildemo) project:

1. In admin area, upload document in "Documents" section
2. Go to Explorer -> Home page.
3. Click edit on "Standard index" page
4. Click on "intro" field then click on document icon
5. Click on link in "file" column

PR #2535, opened 2016-04-28
Fix document chooser download links giving unnecessary warning

kapito:
As @m1kola said in #2513, attempting to download a page from the document chooser produces an unnecessary warning about 'leaving the page'. This is fixed by making links in the document chooser generate with a 'download' class and making the `enableDirtyFormCheck()` function check that the trigger doesn't have this class.


gasman:
Thanks @kapito! Closing this in favour of #2811 now - we've found that checking the links/buttons that trigger the unload event is just too unreliable across browsers, and so we're better off switching to an approach that doesn't depend on it.

Issue #2213, opened 2016-02-09
Lack of cancel button on 'unpublish' or 'delete' pages


heymonkeyriot:
If an editor clicks on `delete` or `unpublish` they're taken to new screens. Neither screen gives the option to `cancel` their decision.

**Unpublish screen**
![image](https://cloud.githubusercontent.com/assets/11335309/12910243/ba7e5ab0-cf00-11e5-9a8b-789cf98672f5.png)

**Delete screen**
![image](https://cloud.githubusercontent.com/assets/11335309/12910222/8d746ca8-cf00-11e5-9920-6767840a8743.png)

The only way for the user to go back to the edit screen is to hit the `back` button or for them to re-find it via the explorer. It's an issue as it's quite easy to hit delete when you mean to hit publish (I've done it a few times at least) and once on the page the messaging is set-up around the idea that you do either want to delete or unpublish the page. In the long run it's probably preferable to allow the user to delete/ unpublish directly from the edit page, since moving to a new page removes any streamfield ordering (c/f #475 ), but that would be a great deal more complex to implement.

Below is an attempt to add the option of cancelling the action for users. I also tried to increase the consistency between the two screens (on unpublish the user is currently asked a question, and given the choice of an answer, on the delete screen it's a narrative with two actions). Below both consistently ask the question that the user has clicked on, with unpublish being hierarchically secondary on the delete screen.

![image](https://cloud.githubusercontent.com/assets/11335309/12910332/c3dc9ada-cf01-11e5-81b1-a9ed3c05c0dd.png)

![image](https://cloud.githubusercontent.com/assets/11335309/12910551/83d0f722-cf03-11e5-98c9-ecdd1712c895.png)

Do others feel this would be an improvement?

davecranwell:
Having the option to not-do it is an improvement I think, although I see no reason for the "alternate unpublish" option to not also be a button.

This whole delete interface is traditionally the domain of a modal window so I wonder whether that would be another/better solution. A hybrid alternative might be to centrally aligned the buttons, and the main question, in the centre of the screen. We might benefit from the ubiquity of centralised "[Do This] or [Don't]" interfaces if we adopted a similar layout.

alexgleason:
There's a good A List Apart article about avoiding warnings. http://alistapart.com/article/neveruseawarning

If you're on Android, you can see this in the Gmail app. Try deleting one of your email threads and see how it disappears immediately with a black confirmation box that lets you undo it.

[...]

PR #2536, opened 2016-04-28
Add back buttons to 'delete' and 'unpublish' confirmation pages

kapito:
As suggested in #2213.

Currently, there are no back buttons for either of these pages.  Fortunately, it wasn't too hard to add a button which redirects to the previous page. I also made the "Are you sure?" text larger and changed the link for the unpublish button (as @heymonkeyriot suggested).


gasman:
Thanks! I've gone with a different approach in #2830 - rather than using `history.go(-1)`, I've made the 'cancel' button a link directly to whatever the landing page of the delete / unpublish action would be. This means that we don't have to put the button outside of the form (and so don't need to fiddle with the form CSS).

PR #2538, opened 2016-04-28
Fix chosen image disappearing, upload button not working and 'This field is required.' message

kapito:
Working on top of @serzans' code from #2007 which fixed chosen images from disappearing if the title was left blank, however this introduced two new bugs. The first was that the 'loading' spinner broke (wouldn't know to stop spinning); I fixed this by creating a `cancelSpinner()` function, which cancels the spinner animation and returns the button to its regular clickable state. The second bug was that the 'This field is required.' message would duplicate itself if the button was clicked repeatedly, which was fixed by checking that there was no error message before displaying a new one.


gasman:
Merged in 78168ba82421762d69e261c16eab46558118dc00

Issue #2527, opened 2016-04-27
When page URLs change, the old URL should be automatically invalidated by the frontendcache module


kaedroho:
When the [frontendcache](http://docs.wagtail.io/en/v1.4.3/reference/contrib/frontendcache.html) 
is enabled. It will automatically invalidate pages when they are published and unpublished.

But when a page's URL is changed (either by moving or changing the slug), it currently doesn't invalidate the previous URL, which could leave old content being served at those URLs.

This is only really a problem for those who use very long cache TTLs for regular pages, but would be nice if we can solve this for everyone in [frontendcache](http://docs.wagtail.io/en/v1.4.3/reference/contrib/frontendcache.html).

lboaretto:
Complementing. It would be interesting to use "Redirects" lib to redirect the OLD url to NEW, after changes.

jschroeter:
@lboaretto Automatic redirects after moving/renaming pages would be awesome, we're also missing this right now. Probably this is another feature. Maybe this could be solved by automatically creating redirect entries after a slug has changed - but we believe it would be better to maintain when a page could just have more than one URL in addition to the current separate Redirects. Of course one canonical URL and multiple additional URLs which will redirect (e.g. similar to [Concrete5](https://www.concrete5.org/files/4813/4012/4237/pageproperties-path.png)).

lboaretto:
@jschroeter very good! your idea is perfect. Similar to https://github.com/saulshanabrook/django-url-tracker I believe...

gasman:
I'd suggest creating a new issue about the proposed redirect behaviour - that's not really what the current issue is about :-)

negletios:
I am also having a similar problem with Redirects. I think its a good feature to add some extra post_save and post_delete signals for the Redirect model. 

It makes sense to me that Redirects should be under the same "cache update logic" with Pages.

What do you think?

th3hamm0r:
Just stumbled over this during a search for something else: this should now be easy with the "page_slug_changed"-signal, which was introduced with #7827.

PR #2543, opened 2016-04-29
Feature/frontendcache 2527

vincentaudebert:
Fixing #2527 
Cache is purged when slug is changed or when page is moved.

### DIFF
--- a/wagtail/contrib/wagtailfrontendcache/signal_handlers.py
+++ b/wagtail/contrib/wagtailfrontendcache/signal_handlers.py
@@ -1,14 +1,14 @@
 from __future__ import absolute_import, unicode_literals
 
 from django.apps import apps
-
-from wagtail.contrib.wagtailfrontendcache.utils import purge_page_from_cache, purge_url_from_cache
 from django.core.exceptions import ObjectDoesNotExist
-from wagtail.wagtailcore.signals import page_published, page_unpublished
 from django.db.models.signals import post_save, pre_save
 
+from wagtail.contrib.wagtailfrontendcache.utils import purge_page_from_cache, purge_url_from_cache
+from wagtail.wagtailcore.signals import page_published, page_unpublished
 
-def has_path_changed(instance, field):
+
+def has_changed(instance, field):
     if not instance.pk:
         return False
     try:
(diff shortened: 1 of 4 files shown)


wagtail/contrib/wagtailfrontendcache/signal_handlers.py
```diff
@@ -14,6 +36,20 @@ def page_unpublished_signal_handler(instance, **kwargs):
...
 
+def page_pre_saved_signal_handler(instance, **kwargs):
+    if has_changed(instance, 'url_path'):
+        old_instance = get_old_instance(instance)
+        instance.old_url = old_instance.full_url
+    else:
+        instance.old_url = None
+
+
+def page_post_saved_signal_handler(instance, **kwargs):
```
kaedroho: This signal handler will fire even when a page is saved as draft, could we put this into the "page_published" signal handler above? (which is called any time the page is saved during a republish)

vincentaudebert: @kaedroho we can't because in the page_published signal the old url has been overwritten already. The only place I could still get it was pre_save signal :(

kaedroho: Sorry, I meant the post_save handler (github displaying the wrong bit of code)

vincentaudebert: Same problem with `post_save` @kaedroho 
The DB has already been updated which means `has_changed()` returns `False`

wagtail/wagtailcore/models.py
```diff
@@ -365,6 +365,14 @@ class Page(six.with_metaclass(PageBase, MP_Node, ClusterableModel, index.Indexed
         editable=False
     )
 
+    old_url = models.TextField(
```
jwicks: Instead of maintaining a DB field on `pre_save` and comparing it on `post_save` , could you set a self.\_original_url_path in Page's `__init__()` method and compare it in `post_save`? Based on this solution http://stackoverflow.com/a/36722021/2141313

jhovanic: It gets replaced when moving a page with the new URL so I don't think it's possible to get the old URL this way.

thibaudcolas:
Hey @vincentaudebert, do you know where this is at? Can we do anything to help get it merged?

vincentaudebert:
@thibaudcolas It has been 10 months :/
I haven't done any extra work on this since then. I think it would be good if someone else can review it and maybe find a better alternative solution? From what I remember I faced many problems with the saving cycle as commented above.

thibaudcolas:
Thanks – I've changed the labels accordingly.

thibaudcolas:
Will close this as it's been a while. If someone wants to pick this up feel free to open a new PR.

PR #2546, opened 2016-04-29
Improves tests for `wagtailforms`

m1kola:
Now `wagtail.wagtailforms.tests.TestFormBuilder` contains checks for all form types available in `wagtailforms`.


kaedroho:
Thanks!

PR #2548, opened 2016-05-01
Page.search_fields is a tuple

pbaehr:
If you follow the tutorial exactly, you will get an error when you try to create the migrations after this step because you are appending a list to a tuple.

### DIFF
--- a/docs/getting_started/tutorial.rst
+++ b/docs/getting_started/tutorial.rst
@@ -229,10 +229,10 @@ model:
         intro = models.CharField(max_length=250)
         body = RichTextField(blank=True)
 
-        search_fields = Page.search_fields + [
+        search_fields = Page.search_fields + (
             index.SearchField('intro'),
             index.SearchField('body'),
-        ]
+        )
 
         content_panels = Page.content_panels + [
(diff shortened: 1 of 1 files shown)


zerolab:
What version of Wagtail are you testing with?

This is expected behaviour in the master branch/upcoming v1.5. Please see #2388 and commit dbb4c77.
See [tutorial in v.1.4.3](http://docs.wagtail.io/en/v1.4.3/getting_started/tutorial.html) vs [master](http://docs.wagtail.io/en/latest/getting_started/tutorial.html)

pbaehr:
Apologies, I must have ended up in the wrong version of the docs. Thanks for the quick response.

zerolab:
Not a problem at all. Thank you for the contribution!

gasman:
Have now added a note about pre-1.5 versions to the tutorial in b39ff4a13495f08035d408bc0becfd5d4a2d5a51 - this has caught a few people out, probably because readthedocs doesn't provide a good way of steering people (and Google) away from in-development versions of the docs. We can remove the note again once 1.5 has been out for a while...

Issue #1502, opened 2015-07-09
Include requestor IP in preview dummy_request?


jasonm:
When `wagtailcore.models.Page.dummy_request` issues a dummy `WSGIRequest` to compute preview content, should it defensively include a reasonable `REMOTE_ADDR` header in case there are middleware installed which expect an IP address?  Or should middleware always defensively handle the case where no requestor IP is provided?

Related to: https://github.com/Miserlou/django-easy-timezones/issues/6

I _think_ it's the responsibility of the middleware library to be defensive in this case, but curious if any Wagtail devs think it's worth including a valid `REMOTE_ADDR` in the request.

gasman:
Yes, I think it would be a good idea to add `REMOTE_ADDR` to the dummy request object, even if it is bad form for middleware (and other code) to rely on its presence - anything that makes the preview rendering more representative of a real page request is a good thing.

I'd suggest updating `dummy_request` to accept an *optional* `original_request` kwarg, and have the preview views pass the incoming request object to it - we can then cherry-pick `REMOTE_ADDR` from there, if available (along with any other headers that might be useful - user-agent, possibly?)

sterago:
I have noticed that the missing `REMOTE_ADDR` header in the dummy request also causes an issue with django's debug context processor, as it relies on this value to determine whether or not it should expose the `debug` context variable to the templates:
https://github.com/django/django/blob/5153a3bfdcec82324d67ff79862384288cf6afe6/django/template/context_processors.py#L41

so adding `REMOTE_ADDR` to the dummy request should fix this issue as well.

gasman:
As per https://groups.google.com/d/msg/wagtail/dp8y1l5T3L8/MXLWOZPjCgAJ, it would also make sense to copy the original request's Cookie header, so that any session-dependent behaviour on the page (e.g. user authentication) would be performed as the logged-in user. (This would cause the Wagtail userbar to display, which we probably don't want - but that's easily fixed by checking the `request.is_preview` flag.)

coredumperror:
I've just run into this problem myself (we use a middleware to restrict access to on-campus users), and was planning to make a PR with precisely the fix suggested here. So I'll go ahead and do that. :)

gasman:
Implemented in #2550.

PR #2550, opened 2016-05-02
Page.dummy_request() now takes an optional request object argument.

coredumperror:
When building a dummy request, you can now pass in the original request object to add additional information to the dummy. Currently, that includes the following headers:
REMOTE_ADDR
HTTP_X_FORWARDED_FOR
HTTP_COOKIE
HTTP_USER_AGENT

More may be added later.

This changes ensures that middleware which work on the client IP aren't flumuxed by its absense, and also makes it possible for previews to be rendered as the logged in user (they had previously been rendered using an AnnonymousUser).

Because the user's logged in state is now detectable in a Page previews, the Wagtail userbar now hides itself explicitly during previews, rather than relying on the fact that previews used to be built with AnonymousUser.

I based much of the functionality of this patch on suggestions from #1502.

### DIFF
--- a/wagtail/wagtailadmin/templatetags/wagtailuserbar.py
+++ b/wagtail/wagtailadmin/templatetags/wagtailuserbar.py
@@ -40,6 +40,10 @@ def wagtailuserbar(context, position='bottom-right'):
     if not request.user.has_perm('wagtailadmin.access_admin'):
         return ''
 
+    # Don't render if this is a preview
+    if request.is_preview:
+        return ''
+
     # Only render if the context contains a variable referencing a saved page
     page = get_page_instance(context)
     if page is None:
--- a/wagtail/wagtailadmin/templatetags/wagtailuserbar.py
+++ b/wagtail/wagtailadmin/templatetags/wagtailuserbar.py
@@ -40,8 +40,9 @@ def wagtailuserbar(context, position='bottom-right'):
     if not request.user.has_perm('wagtailadmin.access_admin'):
         return ''
 
-    # Don't render if this is a preview
-    if request.is_preview:
+    # Don't render if this is a preview. Since some routes can render the userbar without going through Page.serve(),
+    # request.is_preview might not be defined.
+    if getattr(request, 'is_preview', False):
         return ''
 
     # Only render if the context contains a variable referencing a saved page
--- a/wagtail/wagtailadmin/views/pages.py
+++ b/wagtail/wagtailadmin/views/pages.py
@@ -496,7 +496,7 @@ def delete(request, page_id):
 
 def view_draft(request, page_id):
     page = get_object_or_404(Page, id=page_id).get_latest_revision_as_page()
-    return page.serve_preview(page.dummy_request(), page.default_preview_mode)
+    return page.serve_preview(page.dummy_request(request), page.default_preview_mode)
 
 
 def preview_on_edit(request, page_id):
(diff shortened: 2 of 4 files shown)


coredumperror:
Ah shoot, I forgot to write tests, and forgot to even run the existing ones. And of course, that check for request.is_preview causes issues in some tests. Silly me.

coredumperror:
Ah poo. That failure is a timeout error that isn't my fault! :(

timheap:
I like this idea, and I like your implementation. Nice work!

One thought I had for a further improvement: It is possible that other developers will need to extend the dummy request with even more values, taken from the original request or somewhere else. Making `dummy_request()` easy to extend will help these developers out tremendously.  If the method was set up in a similar manner to [`Field.formfield`](https://docs.djangoproject.com/en/1.9/howto/custom-model-fields/#specifying-the-form-field-for-a-model-field), developers could override `Page.dummy_request()` and send extra headers to the request through kwargs:

```python
class MyPage(Page):
    def dummy_request(self, original_request=None, **meta):
        if original_request:
            meta.setdefault('HTTP_REFERRER', original_request.META['HTTP_REFERRER'])
        return super(MyPage, self).dummy_request(original_request, **meta)
```

Then, in `Page.dummy_request()`, do something like:

```python
def dummy_request(self, original_request=None, **meta):
    # ...
    dummy_values.update(**meta)
    request = WSGIRequest(dummy_values)
    # ...
    return request
```

What do you think?

coredumperror:
It'd be easy enough to implement this, so I'll do it. Though I'm not really sure how useful that would be. I'm not terribly creative, but I can't really think of a user case for your proposed functionality. Your code gives an example, but when would an app actually want to do that?

gasman:
I'm not clear on the use-case for passing custom parameters either, and I'm not sure that making dummy_request 'hackable' is a path I'd want to go down. The fact that you need a dummy request at all is a pretty arcane internal implementation detail of Wagtail - I seem to recall that it almost wasn't needed at all, but some time around Wagtail 0.2 I had the choice between adding it as a hack or slightly breaking backwards compatibility - and providing a formal (or semi-formal) API to it would give it a more official / permanent status than it really ought to have.

timheap:
Clearly there is some need for extra parameters in the request, otherwise this PR would not exist. I am guessing that even with these additions, every use-case that a developer might need is not covered. Instead of having another PR in a years time with a developer adding another attribute to the dummy request when they find it missing, the suggested change allows them to easily add missing attributes themselves. It is not a huge addition, and it might not get used much, but it might make someones life easier. The only other option for these developers if this addition is not included is to reimplement the whole `dummy_request()` method.

I must admit I am not sure if there is a huge need for this change either, but then again I would not have though the changes requested in this PR were necessary if @coredumperror had not made this PR :)

gasman:
Ah, I see - so it's effectively an insurance measure for future PRs like this one, rather than something to support fancier customisations. Fair enough then!

coredumperror:
Well, strictly speaking any further change probably *would* still require a new PR, because this one changes more than just the definition of `dummy_request()`. I also had to change all the places in the Wagtail codebase where `dummy_request()` gets called, to have them pass in the original `request`. I'm not really even sure when any code besides Wagtail would ever need to call `dummy_request()` in the first place, which was the main reason I asked for a use case.

gasman:
Merged in cbabc3d9c7a5b9e6bb6f160ad5c40e0ca01aaaeb - thanks!

PR #2552, opened 2016-05-03
Do not get message classes from MESSAGE_TAGS

### LINKED ISSUE #2551: Setting `MESSAGE_TAGS` makes messages in Wagtail loose their styles
I've set `MESSAGE_TAGS` in my project to use the relevant CSS classes for my project, following a BEM naming scheme:

```python
MESSAGE_TAGS = {
    messages.DEBUG: 'messages__debug',
    messages.INFO: 'messages__info',
    messages.SUCCESS: 'messages__success',
    messages.WARNING: 'messages__warning',
    messages.ERROR: 'messages__error',
}
```

Unfortunately, Wagtail implicitly relies on this setting not being changed. If it is changed, the flash messages loose their colours.

Wagtail should either document that this setting is required to be set to its defaults, or it should work regardless of how this setting is set.

timheap:
If the developer had overridden MESSAGE_TAGS in their site, Wagtail messages used these classes in the admin. This caused the messages to lose their styles.

Wagtail now ignores the MESSAGE_TAGS setting, using the default classes defined in `django.contrib.messages.constants.LEVEL_TAGS`.

Fixes #2551

### DIFF
--- a/wagtail/tests/testapp/urls.py
+++ b/wagtail/tests/testapp/urls.py
@@ -2,8 +2,9 @@ from __future__ import absolute_import, unicode_literals
 
 from django.conf.urls import url
 
-from wagtail.tests.testapp.views import bob_only_zone
+from wagtail.tests.testapp.views import bob_only_zone, message_test
 
 urlpatterns = [
     url(r'^bob-only-zone$', bob_only_zone, name='testapp_bob_only_zone'),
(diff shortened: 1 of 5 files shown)


gasman:
Thanks! Merged in 80ad8ffc944f850180a2b078296a4fb9ecc0d618 (master) and f9b39e5890792c454c0fdbdf43b1f47b4d9b4238 (1.4.x).

PR #2555, opened 2016-05-03
Remove 'perspective' declarations from explorer CSS

### LINKED ISSUE #2549: Explorer menu invisible on Firefox 46
Hi,

just noticed that with the new Firefox 46 (tested on Ubuntu 12.04 and Windows 7) the second level of the explorer menu is invisible but clickable (blue highlight from developer tools):

explorer_menu_firefox

Disabling the following rule in the developer toolbar shows the sub-menu, so this seems related:
```css
.explorer ul {
  […]
  transform-style: preserve-3d;
}
```

Using Wagtail 1.3.1 (haven't had a chance to test 1.4.3).

gasman:
Fixes #2549

I'm guessing these rules were in place to enable hardware acceleration, but removing them doesn't visibly affect the rendering. Either way, I figure that making Wagtail usable again on Firefox is more important than the performance boost...

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_explorer.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_explorer.scss
@@ -10,12 +10,6 @@ $explorer-z-index: 500;
     top: 0;
     left: 0;
     display: none;
-    -webkit-perspective: 1000px;
-    -moz-perspective: 1000px;
-    perspective: 1000px;
-    -webkit-perspective-origin: 50% 200%;
-    -moz-perspective-origin: 50% 200%;
-    perspective-origin: 50% 200%;
 
     ul {
         background: $color-grey-1;


m1kola:
Hm... I'm not sure that `perspective` is a reason. I can reproduce this bug on `v1.4.3` tag, but can't reproduce on `master` (now contains `perspective` declarations).

Mac OS 10.11.4, FireFox 46.0.1.

m1kola:
Interesting. Applying of your patch on `v1.4.3` solves the issue, but looks like in `master` this issue was solved in another way.

jonnyscholes:
This changes fixes it for me on Firefox 46.0.1 & Firefox 48.0a2 ("developer edition") on OSX 10.11.3. Tested in Chrome  50.0.2661.94 & Safari 9.0.3 and there are no adverse effects there either.

The animations on the menu use `transform` and `opacity` properties which browsers are very good at optimising by default. We shouldnt need `perspective` for performance (`will-change` would be the preferred method these days if we did). I suspect this might have been left over from a more complex animation in the past (swinging out from the sidebar "in 3d" perhaps).

gasman:
It seems that @alexgleason's fix in #2531 (796239b7d991322632da6f9f63fd1ccd172d3b58) fixed this too. (side note: using `git bisect` to find out when something fixed itself is really confusing, because you have to mentally switch around `bad` and `good` :-) )

Since it was a fairly arbitrary choice _not_ to merge that fix into 1.4.x, I'll go ahead and backport it now instead of this one, so that both branches have a consistent fix.

gasman:
Fix 796239b7d991322632da6f9f63fd1ccd172d3b58 backported to stable/1.4.x in 3ac9c457eb7f6077ed56fa9ecd8b402dba740775.

alexgleason:
wow... really? What a strange phenomenon. Fixing problems I didn't even know exist with code that's inconceivably related.

timheap:
@gasman: `git bisect` recently learned some new flags that might help with this case: [`--term-old` and `--term-new`](https://git-scm.com/docs/git-bisect#_alternate_terms):

```
$ git bisect start --term-old broken --term-new fixed
# Set the bounds of the bisect range
$ git bisect broken <some-old-commit>
$ git bisect fixed master
# Test the commit, etc, see if the CSS is still broken or not
$ git bisect broken # The CSS is still borked in this commit
$ git bisect fixed # Someone has accidentally fixed it in this commit
```

jonnyscholes:
@gasman I think this code should still be removed. It's not currently doing anything and is likely to cause bugs or confusion down the line.

gasman:
@jonnyscholes OK, merged in db449bff2aab5243eb79d193f40442e649362846.

jonnyscholes:
@gasman thanks :)

PR #2556, opened 2016-05-03
Split up API implementation (for v2 and admin API)

kaedroho:
As per https://github.com/torchbox/wagtail-rfcs/pull/8 this PR restructres the API v2 and admin API so endpoint definitions are in their respective apps rather than a central module.

I've left the tests where they are as I think it's quite nice to be able to just do ``python runtests.py wagtail.wagtailadmin.tests.api`` to run every admin api test


kaedroho:
Going to redo this so that it's based on the implementation currently in master and I'll reimplement the fields change and admin API on top of that.

kaedroho:
Superseded by #2649

PR #2554, opened 2016-05-03
Make elasticsearch switchable in tox, allow ES failures in travis

gasman:
The random Elasticsearch failures on Travis are causing way too much noise, and confusion for contributors. See https://github.com/torchbox/wagtail/pulls?utf8=%E2%9C%93&q=is%3Apr+travis for a sample...

This PR moves the Elasticsearch tests into their own Travis builds, flagged as "allow failures" - this way, they won't interfere with regular development, but we can still check that those tests succeed prior to shipping a new release.

PR #2558, opened 2016-05-03
Omitting --elasticsearch should skip ES tests regardless of ELASTICSEARCH_URL setting

gasman:
#2554 didn't work as intended, evidently because something in the Travis config is still setting the ELASTICSEARCH_URL environment variable. This causes ES tests to run regardless of the presence of the --elasticsearch flag.

This PR changes the behaviour so that --elasticsearch MUST be passed to ./runtests.py in order to run ES tests (which will hopefully ensure that Travis only runs them when it's supposed to).


kaedroho:
Looks good to me!

Issue #2560, opened 2016-05-04
StreamField block controls not visible


vincentaudebert:
Streamfield block controls are not visible when you prepend a new block to the streamfield.

![screen shot 2016-05-04 at 2 49 35 pm](https://cloud.githubusercontent.com/assets/3360430/15004144/7946f738-1207-11e6-966d-2e58da70e19c.png)

gasman:
Fixed in #2561

PR #2561, opened 2016-05-04
Fixing streamfield block controls not visible

vincentaudebert:
Fixing #2560 - changing z-index on class sequence-controls so we can see the controls on top of stream field menu


alexgleason:
Thanks! :+1:

Wondering why the StreamField is sitting at level 5 anyway. Might be worth investigating.

gasman:
Merged in 42b9fb893c588b2bf72c9f28fece69c63a0e388a - thanks!

PR #2562, opened 2016-05-04
Recursive unpublish for descendant pages

kaedroho:
Rebase/squash of https://github.com/torchbox/wagtail/pull/2455

Original description:

Some of our clients often demand to automatically un-publish all descendant pages of a given page when un-publishing it. I've added a checkbox to the confirmation un-publish template in order to do that. It checks for permissions and only un-publishes what the user is allowed to. Added a few tests as well

### DIFF
--- a/wagtail/wagtailadmin/templates/wagtailadmin/pages/confirm_unpublish.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/pages/confirm_unpublish.html
@@ -10,7 +10,23 @@
         <form action="{% url 'wagtailadmin_pages:unpublish' page.id %}" method="POST">
             {% csrf_token %}
             <input type="hidden" name="next" value="{{ next }}">
-            <input type="submit" value="{% trans 'Yes, unpublish it' %}" class="button">
+            <ul class="fields">
+                {% if live_descendant_pages %}
+                <li>
+                    <div class="field boolean_field checkbox_input">
+                        <div class="field-content">
+                            <div class="input">
+                                <input id="id_include_descendants" name="include_descendants" type="checkbox">
+                                <span>{% trans 'Unpublish all descendant pages as well.' %}</span>
+                            </div>
+                        </div>
+                    </div>
+                    </li>
+                {% endif %}
+                <li>
+                    <input type="submit" value="{% trans 'Yes, unpublish it' %}">
+                </li>
+            </ul>
         </form>
     </div>
 {% endblock %}
--- a/wagtail/wagtailadmin/templates/wagtailadmin/pages/confirm_unpublish.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/pages/confirm_unpublish.html
@@ -11,7 +11,7 @@
             {% csrf_token %}
             <input type="hidden" name="next" value="{{ next }}">
             <ul class="fields">
-                {% if live_descendant_count and live_descendant_count > 0 %}
+                {% if live_descendant_count > 0 %}
                 <li>
                     <div class="field boolean_field checkbox_input">
                         <div class="field-content">
--- a/wagtail/wagtailadmin/templates/wagtailadmin/pages/confirm_unpublish.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/pages/confirm_unpublish.html
@@ -11,20 +11,24 @@
             {% csrf_token %}
             <input type="hidden" name="next" value="{{ next }}">
             <ul class="fields">
-                {% if live_descendant_pages %}
+                {% if live_descendant_count and live_descendant_count > 0 %}
                 <li>
                     <div class="field boolean_field checkbox_input">
                         <div class="field-content">
                             <div class="input">
                                 <input id="id_include_descendants" name="include_descendants" type="checkbox">
-                                <span>{% trans 'Unpublish all descendant pages as well.' %}</span>
+                                <span>{% blocktrans count counter=live_descendant_count %}
+                    This will also unpublish one more subpage.
+                {% plural %}
+                    This will also unpublish {{ live_descendant_count }} more subpages.
+                {% endblocktrans %}</span>
                             </div>
                         </div>
                     </div>
                     </li>
                 {% endif %}
                 <li>
-                    <input type="submit" value="{% trans 'Yes, unpublish it' %}">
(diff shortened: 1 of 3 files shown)


wagtail/wagtailadmin/views/pages.py
```diff
@@ -629,8 +629,17 @@ def unpublish(request, page_id):
...
 
     if request.method == 'POST':
+        include_descendants = request.POST.get("include_descendants", False)
+
         page.unpublish()
 
+        if include_descendants == 'on':
+            live_descendant_pages = page.get_descendants().live()
+            for live_descendant_page in live_descendant_pages:
+                live_descendant_page = live_descendant_page.specific
```
timheap: Consider using `live_descendant_pages = page.get_decendants().live().specific()` instead - it should reduce the number of DB queries doing this on large, diverse trees.

wagtail/wagtailadmin/views/pages.py
```diff
@@ -629,8 +629,17 @@ def unpublish(request, page_id):
...
+                if live_descendant_page.permissions_for_user(request.user).can_unpublish():
+                    live_descendant_page.unpublish()
```
timheap: Considering that not being able to unpublish the root page raises a `PermissionDenied` error above, silently ignoring this issue here is surprising. In my opinion, either the whole tree should be unpublished, or nothing should be unpublished. Perhaps, if the include_decendants checkbox is checked, the descendants shoul all be checked for permission before unpublishing them, and an error being raised if the user does not have permission.

wagtail/wagtailadmin/templates/wagtailadmin/pages/confirm_unpublish.html
```diff
@@ -10,7 +10,23 @@
...
             <input type="hidden" name="next" value="{{ next }}">
-            <input type="submit" value="{% trans 'Yes, unpublish it' %}" class="button">
+            <ul class="fields">
+                {% if live_descendant_pages %}
+                <li>
+                    <div class="field boolean_field checkbox_input">
+                        <div class="field-content">
+                            <div class="input">
+                                <input id="id_include_descendants" name="include_descendants" type="checkbox">
+                                <span>{% trans 'Unpublish all descendant pages as well.' %}</span>
```
timheap: Can this message include the number of descendant pages? The delete confirmation page includes this number as well, and it provides a nice sanity check for editors.

wagtail/wagtailadmin/templates/wagtailadmin/pages/confirm_unpublish.html
```diff
@@ -10,7 +10,23 @@
         <form action="{% url 'wagtailadmin_pages:unpublish' page.id %}" method="POST">
             {% csrf_token %}
             <input type="hidden" name="next" value="{{ next }}">
-            <input type="submit" value="{% trans 'Yes, unpublish it' %}" class="button">
+            <ul class="fields">
+                {% if live_descendant_pages %}
```
timheap: This will call [`QuerySet.__bool__()`](https://github.com/django/django/blob/1.9/django/db/models/query.py#L261-L263), which fetches all of the results before checking if there are any records. This is quite inefficient on large trees. `{% if live_descendant_pages.exists %}` is much more efficient.

wagtail/wagtailadmin/tests/test_pages_views.py
```diff
@@ -2029,6 +2029,99 @@ def test_unpublish_view_post(self):
         self.assertEqual(mock_call['instance'], self.page)
         self.assertIsInstance(mock_call['instance'], self.page.specific_class)
 
+    def test_unpublish_descendants_view(self):
+        """
+        This tests that the unpublish view responds with an unpublish confirm page that contains the form field 'include_descendants'
```
timheap: This docstring says the page must contain the checkbox, while the test checks the page does *not* contain the checkbox.

jordij:
Thanks @timheap. I've implemented some of your suggestions.

> Considering that not being able to unpublish the root page raises a PermissionDenied error above, silently ignoring this issue here is surprising. In my opinion, either the whole tree should be unpublished, or nothing should be unpublished. Perhaps, if the include_decendants checkbox is checked, the descendants should all be checked for permission before unpublishing them, and an error being raised if the user does not have permission.


I'm not sure about the permissions though. What do you think if we only allow this feature (un-publish including descendants) when the user has the right permissions for the whole sub-tree? Checking beforehand each descendant can be unpublished by the user. It adds a significant cost since we always need to check the whole subtree but we ensure there are no surprises when the un-publish operation is performed.

Any thoughts on the preferred approach are welcome.

wagtail/wagtailadmin/templates/wagtailadmin/pages/confirm_unpublish.html
```diff
@@ -11,20 +11,24 @@
             {% csrf_token %}
             <input type="hidden" name="next" value="{{ next }}">
             <ul class="fields">
-                {% if live_descendant_pages %}
+                {% if live_descendant_count and live_descendant_count > 0 %}
```
kaedroho: Since this will always be a number, you could shorten this to: ``{% if live_descendant_count > 0  %}``

gasman:
> What do you think if we only allow this feature (un-publish including descendants) when the user has the right permissions for the whole sub-tree?

I think it's worth noting that in the current permission model, this is implied automatically - permissions propagate down the tree, so you'd never be in a situation where you have unpublish permission on the parent but *don't* have unpublish permission over all of its descendants.

That's not to say we can entirely discard the permission check on descendants entirely - I think it's absolutely appropriate to have it in there, so that the view code isn't relying on any built-in assumptions about how the permission model behaves (especially as the permission model may well be updated or made configurable in future). However, it does mean that we don't currently need to worry about handling the "incomplete permissions" scenario optimally from a UX point of view (i.e. it's fine for it to silently ignore pages, or display a redundant checkbox) since that never occurs in practice. Of course, we can revisit that in future if the permission model does change.

gasman:
Merged in ae947642cda2dec5d80096739392253b12fe67fa + parents (with some fixes, including a change to the checkbox wording - "this will also unpublish X subpages" implies that the user doesn't get a choice in the matter).

Thanks @jordij and @kaedroho!

PR #2568, opened 2016-05-05
[Elasticsearch 2] Replaced usage of index_analyzer

kaedroho:
index_analyzer has been removed in Elasticsearch 2.0. But the new way works in older versions as well so I think it's worth switching to the new way now.

https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_20_mapping_changes.html#_analyzer_mappings

### DIFF
--- a/wagtail/wagtailsearch/backends/elasticsearch.py
+++ b/wagtail/wagtailsearch/backends/elasticsearch.py
@@ -62,7 +62,8 @@ class ElasticSearchMapping(object):
                     mapping['boost'] = field.boost
 
                 if field.partial_match:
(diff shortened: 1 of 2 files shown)


m1kola:
Merged in d14150eb43030c38dddd58869564ea5b6e521f89. Thanks!

gasman:
Reverted in 15ef4fcca74e9464b72619679bf5e13e6332ab29, as this causes TestElasticSearchBackend.test_query_analyser to fail on Python 2.7. (This failure wasn't shown in the checks here, because ES tests are set to allow-failures to avoid noise from false positives.)

PR #2569, opened 2016-05-05
[Elasticsearch 2] Renamed search backend classes

kaedroho:
Suffixing them with the word ``Backend``

PR #2570, opened 2016-05-05
[Elasticsearch 2] Rewrote update_index command to support rebuilding multiple indices in the same backend

kaedroho:
Elasticsearch 2 no longer allows two different content types in the same index to both use the same column name with different mappings.

For example, images, documents and pages all have a title field. But if they're put in the same index, they must have the exact same "mapping" (so none of them can have a different boost, analyzer, etc to the others or Elasticsearch will refuse to index it).

To prevent this from happening, I plan to use separate indices for different content types. So images, pages and documents will exist in three separate indices rather than one.

All page models will continue to share an index so conflicting mappings can still happen if two page models have a column with the same name, I plan to solve this in a later pull request by prefixing the field names.

This pull request makes it possible to implement a search backend with this behaviour.

We could solve this by prefixing everything and using one index, but I think it makes more sense to use a different index for different content types.

### DIFF
--- a/wagtail/wagtailsearch/backends/elasticsearch.py
+++ b/wagtail/wagtailsearch/backends/elasticsearch.py
@@ -370,7 +370,7 @@ class ElasticsearchSearchResults(BaseSearchResults):
     def _do_search(self):
         # Params for elasticsearch query
         params = dict(
-            index=self.backend.index_name,
+            index=self.backend.get_index_for_model(self.query.queryset.model).name,
             body=self._get_es_body(),
             _source=False,
             fields='pk',
@@ -401,7 +401,7 @@ class ElasticsearchSearchResults(BaseSearchResults):
     def _do_count(self):
         # Get count
         hit_count = self.backend.es.count(
-            index=self.backend.index_name,
+            index=self.backend.get_index_for_model(self.query.queryset.model).name,
             body=self._get_es_body(for_count=True),
         )['count']
 
@@ -700,6 +700,9 @@ class ElasticsearchSearchBackend(BaseSearchBackend):
             timeout=self.timeout,
             **params)
 
+    def get_index_for_model(self, model):
+        return self.index_class(self, self.index_name)
+
     def get_index(self):
         return self.index_class(self, self.index_name)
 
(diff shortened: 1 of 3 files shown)


wagtail/wagtailsearch/backends/elasticsearch.py
```diff
@@ -708,19 +712,21 @@ def reset_index(self):
...
     def add_bulk(self, model, obj_list):
-        self.get_index().add_items(model, obj_list)
+        self.get_index_for_model(model).add_items(model, obj_list)
 
     def delete(self, obj):
-        self.get_index().delete_item(obj)
+        self.get_index_for_model(type(obj)).delete_item(obj)
 
+# Backwards compatibility
+ElasticSearch = ElasticsearchSearchBackend
```
m1kola: It seems to me that we need to add a deprecation warning here.

kaedroho: I decided against this because it would be tricky (but not impossible) to do, because of the different places it can be called from require different messages.

But if it must be done, it must be done.

m1kola: Can you give me an idea, why do we need a different messages? Can we just add a sentence to "Upgrade considerations" and say something like this in a deprecation warning?

> wagtail.wagtailsearch.backends.elasticsearch.ElasticSearch now is deprecated and will be removed in Wagtail 1.8. Please, use wagtail.wagtailsearch.backends.elasticsearch.SearchBackend instead.

kaedroho: https://github.com/torchbox/wagtail/pull/2569

kaedroho: > Can you give me an idea, why do we need a different messages? Can we just add a sentence to "Upgrade considerations" and say something like this in a deprecation warning?

There will be a lot of users that have an old backend name in their settings, we should point out that they need to change their ``WAGTAILSEARCH_BACKENDS`` setting.

m1kola: > done in: #2569

Just noticed that my initial comment should be in #2569 :) Sorry for that.

kaedroho:
Rebased on master

PR #2571, opened 2016-05-06
[Elasticsearch 2] Make deciding index column names the responsibility of the search backend

kaedroho:
We need to prefix some fields in Elasticsearch 2 but not Elasticsearch 1. Moving this code into the ElasticsearchMapping class allows us to override it.

It also makes sense to have this sort of thing in the backend rather than the core as other search backends we add in the future may want different behaviour as well.


kaedroho:
Rebased on master

m1kola:
Merged in 210b8532d954aeed514811a69563a1acb8a67618.

Thanks!

PR #2572, opened 2016-05-06
[Elasticsearch 2] Fixed bug causing nested objects to use the wrong mapping

kaedroho:
Nested objects that are defined using RelatedFields were being mapped into Elasticsearch using the parent model's mapping.

For example, if we index a page with tags, the tags were being mapped with the Page mapping instead of the Tag mapping.

This doesn't make any difference for Elasticsearch 1, but Elasticsearch 2 needs to prefix fields that have been defined on a child model but it was doing this across the related fields as well.

This fix is covered by the tests coming in the Elasticsearch 2 backend.


kaedroho:
Rebased on master

m1kola:
Merged in 2ea1499. Thanks!

PR #2575, opened 2016-05-06
Update decorator patterns util to preserve original view method name

mariocesar:
When `decorate_urlpatterns` apply a decorator to a view function it lost spec names, docs, and the module where it was imported, making it uneasy to inspect when debugging (Like getting the name view and docs when using `django-debug-toolbar`)

Using `functools.wraps` it updates the resulting function to look like the original wrapped view function.

Without the use of this decorator factory, the name of the wrapped view function would have been `decorated_view`.

### DIFF
--- a/wagtail/utils/urlpatterns.py
+++ b/wagtail/utils/urlpatterns.py
@@ -1,5 +1,7 @@
 from __future__ import absolute_import, unicode_literals
 
+from functools import wraps
+
 
 def decorate_urlpatterns(urlpatterns, decorator):
     for pattern in urlpatterns:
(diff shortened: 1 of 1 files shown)


mariocesar:
The difference on django-debug-toolbar, after the change it correctly shows the current view function name.

captura de pantalla de 2016-05-06 17-27-14

captura de pantalla de 2016-05-06 17-27-23

An small and petty change, but makes wonders when developing :)

timheap:
Nice one!

In cases like this, where you're not using `wraps` as a decorator, `update_wrapper` is a better choice. `wraps` is just a partially applied `update_wrapper` made for decorating functions. This avoids the awkward double brackets `wraps(...)(...)`:

```python
from functools import update_wrapper

pattern._callback = update_wrapper(decorator(pattern.callback), pattern.callback)
```

mariocesar:
@timheap Agree,

I updated the patch

timheap:
Merged in b95cc1e687c8a350c9de18168e24b9dccdb46da3. Thanks!

PR #2578, opened 2016-05-09
Add logout success message

SableWalnut:
Add a logout success message so users know when they've successfully logged out

### DIFF
--- a/wagtail/wagtailadmin/views/account.py
+++ b/wagtail/wagtailadmin/views/account.py
@@ -126,6 +126,7 @@ def login(request):
 def logout(request):
     response = auth_views.logout(request, next_page='wagtailadmin_login')
 
+    messages.success(request, _('You have been successfully logged out.'))
     # By default, logging out will generate a fresh sessionid cookie. We want to use the
     # absence of sessionid as an indication that front-end pages are being viewed by a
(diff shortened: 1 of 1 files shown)


wagtail/wagtailadmin/views/account.py
```diff
@@ -126,6 +125,7 @@ def login(request):
 def logout(request):
     response = auth_views.logout(request, next_page='wagtailadmin_login')
 
+    messages.success(request, 'You have successfully logged out!')
```
tomdyson: Thanks, Liam! The message should be a translatable string, e.g. https://github.com/takeflight/wagtail/blob/logout-message/wagtail/wagtailadmin/views/account.py#L59. I'm not sure we need the `!`.

wagtail/wagtailadmin/views/account.py
```diff
@@ -12,7 +12,6 @@
 from django.utils.translation import ugettext as _
 from django.views.decorators.cache import never_cache
 from django.views.decorators.debug import sensitive_post_parameters
-
```
dirtycoder: This blank line should be kept. :)

SableWalnut:
All fixed up. :)

m1kola:
Merged in ca846a5941e2bdbe82a97b4a99cfea5876172d85. Thanks, Liam!

PR #2579, opened 2016-05-09
Add spinner/button longrunning class to settings save button

SableWalnut:
Made the button for saving settings a `<button>` element and added the `button longrunning` class to it.


m1kola:
Merged in 05be39e85f7fde5ba28966a6d0cdfb238d9cc918. Thanks!

PR #2583, opened 2016-05-10
Adds jinja2 requirement to setup testing extras

bmihelac:
jinja2 is dependency of Sphinx and is installed with when using `docs`, but not when only `testing` target is used.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,6 +41,7 @@ testing_extras = [
     'pytz>=2014.7',
     'Pillow>=2.7.0',
     'elasticsearch>=1.0.0',
+    'jinja2>=2.7',
 
     # For coverage and PEP8 linting
     'coverage>=3.7.0',


m1kola:
Hi @bmihelac!

Wagtail's tests doesn't require Sphinx or Jinja2. You can run `pip install -e .[testing] -U` and then run tests `python runtests.py`.

Can you provide more details, please?

Thanks anyway!

bmihelac:
I'm pretty sure running wagtail tests required jinja2 to be installed. I will check once again when I get to computer.

m1kola:
@bmihelac sorry, you are right! [Here](https://github.com/torchbox/wagtail/blob/master/tox.ini#L27) is the reason why I was thinking so :)

m1kola:
@gasman pointed me to [a reason](https://github.com/torchbox/wagtail/pull/2318#discussion_r57442282) why `Jinja2` not in `setup.py`'s `testing_extras`.

But I got 777 (from 1846) tests failed when I ran tests without Jinja. It seems to me that every test that renders something fails without Jinja. It's probably because [`TEMPLATES`](https://github.com/torchbox/wagtail/blob/master/wagtail/tests/settings.py#L57) setting contains Jinja backend.

I  think we need to add it to `testing_extras`. In case we decided to do this, we need to remove `Jinja2` from `tox.ini`.

bmihelac:
No problem @m1kola. This seven hundred  and something failing tests was what I got with runtests first time. I see for the reasons and if this PR is not merged, maybe just mention jinja2 requirement in docs so first time test runners are not scared :)

gasman:
Have now double-checked the logic, and you're right - Jinja2 should indeed be included in testing_extras (since the tests fail outright without it - it doesn't just skip the jinja2 tests). Added in 4a5296e - thanks!

PR #2585, opened 2016-05-10
Fixes RoutablePage throws a TypeError (#2584)

### LINKED ISSUE #2584: RoutablePage throws a TypeError if index route is not set
RoutablePage should not raise TypeError if index route is not set. 

[route method](https://github.com/torchbox/wagtail/blob/master/wagtail/contrib/wagtailroutablepage/models.py#L97) properly calls parent class method if route is not found, but `serve` method has different signature, so it fails.

There is also a discussion of similar issue on Google groups:
https://groups.google.com/forum/#!msg/wagtail/Q9FymA-MOxM/1UkQ0hRGAAAJ

bmihelac:
FIX #2584 RoutablePage throws a TypeError if index route is not set


kaedroho:
Thanks!

PR #2587, opened 2016-05-10
Make api_fields a list

kaedroho:
All other model attribute settings in Wagtail use lists now. This commit changes the api_fields docs to use a list as well.

Both list and tuple work, but list is now the recommended option


m1kola:
Merged in e72b25f61c4dc3e2b95eab7e349d0499f0669f1f. Thanks, @kaedroho!

PR #2607, opened 2016-05-12
Add Preview button to set of permitted buttons for navigating away

### LINKED ISSUE #2602: Preview page with validation errors results in Stay on Page / Leave Page dialog
We have found that in IE11/Win8.1 and FF46/Linux (but not Chrome 50/Linux) if you attempt to preview a page with validation errors, i.e. mandatory fields empty, the preview window opens and spins for a bit. You are then returned to the original tab but prompted with an `onbeforeunload` dialog box stating that data may be lost if you leave the page - the leaving being an attempt to reload the page to show the validation errors. 

Stay on Page prevents validation errors from being shown. Leave Page results in the errors being indicated. 

We have occasionally observed that data is actually lost e.g. the content of Hallo editor fields revert to how they were last time the page was loaded, but I am not able to replicate that behaviour. However, see #2605.

~~Related to this (happy to add a different issue) in IE we have also found that this dialog appears when publishing a page unless it has just been saved as a draft and not subsequently edited.~~ See #2604.

Thanks,
Stephen

riceyrice:
This fixes #2602 for me in FF 46 and Chrome 50.
~~Due to poor git skills I have also tangled up a fix for 2604, sorry for any confusion.~~

### DIFF
--- a/wagtail/wagtailadmin/templates/wagtailadmin/pages/create.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/pages/create.html
@@ -82,6 +82,7 @@
                 '#page-edit-form',
                 {
                     confirmationMessage: '{{ confirmation_message|escapejs }}',
+                    ignoredButtonsSelector: 'input[type="submit"],button[type="submit"],button.action-preview',
 
                     {% if form.errors %}
                         alwaysDirty: true,
--- a/wagtail/wagtailadmin/templates/wagtailadmin/pages/edit.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/pages/edit.html
@@ -116,6 +116,7 @@
                 '#page-edit-form',
                 {
                     confirmationMessage: '{{ confirmation_message|escapejs }}',
+                    ignoredButtonsSelector: 'input[type="submit"],button[type="submit"],button.action-preview',
 
                     {% if form.errors %}
                         alwaysDirty: true,


wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
```diff
@@ -58,7 +58,7 @@ var dirtyFormCheckIsActive = true;
 function enableDirtyFormCheck(formSelector, options) {
     var $form = $(formSelector);
     var $ignoredButtons = $form.find(
-        options.ignoredButtonsSelector || 'input[type="submit"],button[type="submit"]'
+        options.ignoredButtonsSelector || 'input[type="submit"],button[type="submit"],button.action-preview'
```
gasman: It would be better to add this as an `ignoredButtonsSelector` parameter on the call to enableDirtyFormCheck (https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailadmin/templates/wagtailadmin/pages/create.html#L81-L89 and https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailadmin/templates/wagtailadmin/pages/edit.html#L115-L124) - `enableDirtyFormCheck` is intended to be a generic function, not tied to any particular form.

riceyrice:
That did also occur to me :)

riceyrice:
OK @gasman I have revised this to use `ignoredButtonsSelector` as recommended.

gasman:
Unfortunately neither variant of the fix seems to solve the issue for me on FF 46 (Mac)... I think the problem is that the page reload (which is not so much a reload as a rewrite-in-place - https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js#L495-L497) only happens after it's got an AJAX response back from the server, and at that point the browser no longer considers the preview button to be the thing that initiated the reload event. The proper way to fix this is probably to call `disableDirtyFormCheck` at the point the reload happens - will give that a go.

Rather worryingly, preview seems to be entirely broken on my IE11 / Win7 test VM - I suspect this may be because they've removed 'MSIE' from the user agent string (https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js#L455)...

gasman:
The `disableDirtyFormCheck` approach seems to do the job - closing this in favour of #2614.

riceyrice:
Funnily enough I was on the way to reporting the IE11 broken preview bug once I'd verified it. That `MSIE` check should probably also check for `Trident`...

gasman:
As it happens I managed to find a solution for the IE11 issue that didn't rely on user-agent sniffing - see #2613.

PR #2612, opened 2016-05-13
Fix JS syntax error in DropDownController

gasman:
IE11 chokes on this...


gasman:
Merged in 83a11ecee4e82b051ee7e6d37f09dbbe9f6cc200

PR #2613, opened 2016-05-13
Make preview window work on IE11

gasman:
ref: https://github.com/torchbox/wagtail/pull/2607#issuecomment-219017144

IE doesn't provide standard onload event listeners on opened windows, so this code did user-agent sniffing on the string 'MSIE' to bypass the onload animation. This failed on IE11, which drops 'MSIE' from the user agent string. The code now checks for the presence of addEventListener instead.

As a bonus, it now falls back on IE's (non-standard?) attachEvent method, so the animation now works on IE after all.

NB this PR incorporates 717ef27fbdedd3139dfb2ac65fd51e72a0f716bf (#2612) - this is 1.5-specific and should be skipped when backporting to 1.4.x

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
@@ -329,7 +329,7 @@ wagtail = (function(document, window, wagtail) {
             return this._dropDowns;
         },
 
-        getByIndex(index) {
+        getByIndex: function(index) {
             return this._dropDowns[index];
         },
 
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
@@ -452,13 +452,18 @@ $(function() {
 
         previewWindow = window.open($this.data('placeholder'), $this.data('windowname'));
 
-        if (/MSIE/.test(navigator.userAgent)) {
-            // If IE, load contents immediately without fancy effects
-            submitPreview.call($this, false);
-        } else {
-            previewWindow.onload = function() {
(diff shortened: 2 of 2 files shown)


spapas:
Some tests on my Windows 10 machine:

For IE 11 the behavior is *almost* perfect: The preview along with the loading animation work fine however, if there are non-filled required fields then the loading animation will be displayed momentarily and I will be returned to the page editing form with the message "Are you sure you want to leave this page" etc. I think that the correct behavior when trying to preview with missing required fields would be to just display the errors (and don't try the preview at all).

For my Edge browser (which is the default on windows 10) the behavior is the same as with IE 11.

For my Chrome browse the behavior is a little better (but still not 100% perfect): When clicking preview with missing required fields you'll see the preview animation momentarily and return to the form page with the errors marked (so the "are you sure you want to leave this page" message will not be displayed here).

Finally, one last thing that I observed (I don't know if that's a real problem or it is intented): When adding a new page (not saved yet) when I clicked the preview button (which had a different style than normally) I got the following stacktrace:

ValidationError at /admin/pages/add/standardpage/standardpage/3/preview/ {'path': ['Page with this Path already exists.']}

gasman:
Hi @spapas - the issues you report are fixed separately in #2614 and #2616 :-)

spapas:
Hello @gasman after applying those patches everything seems to be working great on all the browsers I tested (Edge, IE 11, Chrome). 

The only thing that could be improved a bit is that the preview animation probably shouldn't be displayed when the page has validation errors.

gasman:
Merged in b1c6063b4118c7ed3ec4ec07f553253e8d11bf87 (master) and 4ee93914cec4cb5653a00c128fabd5205a909f5c (1.4.x)

PR #2614, opened 2016-05-13
Prevent unsaved edit warning following validation errors on preview

### LINKED ISSUE #2602: Preview page with validation errors results in Stay on Page / Leave Page dialog

gasman:
Fixes #2602

Incorporates #2612 and #2613

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
@@ -329,7 +329,7 @@ wagtail = (function(document, window, wagtail) {
             return this._dropDowns;
         },
 
-        getByIndex(index) {
+        getByIndex: function(index) {
             return this._dropDowns[index];
         },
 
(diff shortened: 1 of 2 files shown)


gasman:
Merged in 20f01407114eea0c6d82208b5ab8c429892cb50b (master) and 26d7e2732780b33438a085e6dbab112dfb8148f0 (1.4.x)

PR #2615, opened 2016-05-13
Restore browser interface in the page chooser when the search box is cleared

### LINKED ISSUE #2608: Page Chooser cannot navigate to child pages in Internet Explorer
When choosing a page chooser e.g. a `PageChooserBlock` in a `StreamField`, using IE11/Win8.1, the list of pages at root appears OK, with an AJAX request to e.g.:
`/admin/choose-page/?page_type=wagtailcore.page`
Clicking a right-hand triangle to navigate to a child page shows the correct list of pages:
`/admin/choose-page/3/?page_type=wagtailcore.page`
However, as soon as this list appears the modal immediately submits an empty search:
`/admin/choose-page/search/?page_type=wagtailcore.page&q=&results_only=true`
There is no text in the search box, so the page list is replaced with "There are 0 results"
So pages can be found by searching but not by browsing.

gasman:
This also fixes #2608, which is caused by IE's buggy 'oninput' event handling - it treats the
clearing of the placeholder text when the search box is focused as a user input action, and
thus immediately fires off an empty search query. This change means that the empty search has
no effect.

### DIFF
--- a/wagtail/wagtailadmin/templates/wagtailadmin/chooser/browse.js
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/chooser/browse.js
@@ -1,6 +1,6 @@
 function(modal) {
-    /* Set up page navigation and link-types links to open in the modal */
-    $('a.navigate-pages, .link-types a', modal.body).click(function() {
+    /* Set up link-types links to open in the modal */
+    $('.link-types a', modal.body).click(function() {
         modal.loadUrl(this.href);
         return false;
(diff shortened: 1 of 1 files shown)


kaedroho:
Merged in a2fe305aa0342e332931265b5f9b3d44585371bf (``master``) and b3c5519448831c95d92ce5fb357be833e9c5d2c2 (``stable/1.4.x``) thanks!

PR #2577, opened 2016-05-08
Creates new panel for display information about submissions in pages implementing the AbstractForm class

dirtycoder:
It should fix #406.
Also, removes the extra whitespace between `,` and `)` as seen in the Python documentation.

I didn't made any changes in the CSS and used the classes that's already used in other types of panels.
I believe I could use some guidance regards the way this information should be displayed.

So far the result is this:
captura de tela de 2016-05-08 16-06-42

PR #2619, opened 2016-05-15
Fix #1646: applied the patch from #1911, small change to the proposed query and create a new index

### LINKED ISSUE #1646: "Your most recent edits" listing excludes pages that another user has subsequently edited
https://github.com/torchbox/wagtail/pull/1231#issuecomment-135470115

The "Your most recent edits" listing on the admin dashboard only includes pages whose latest revision was created by the current user. If Alice edits a page, then Bob edits the same page, it falls off Alice's list even if it's the one she edited most recently.

dirtycoder:
I've made two changes for this. I've made some comparisions for both changes between three versions of the query. My base have 50k pages and 100k revisions. I'm using PostgreSQL.

First I moved the LIMIT to the inner query, reducing the number of rows for some operations like the hash join.

|    Query    | Time  |   Cost   |
|-------------|-------|----------|
| Current     | 112ms | 12183.33 |
| PR 1911     | 203ms | 36004.76 |
| Inner LIMIT | 123ms | 19235.77 |

I also created an index on the `created_at` column.

|    Query    | Time  |   Cost   |
|-------------|-------|----------|
| Current     | 91ms  | 12183.23 |
| PR 1911     | 123ms | 13438.32 |
| Inner LIMIT | 81ms  | 10425.82 |

### DIFF
--- a/wagtail/wagtailadmin/tests/test_pages_views.py
+++ b/wagtail/wagtailadmin/tests/test_pages_views.py
@@ -2956,3 +2956,73 @@ class TestIssue2599(TestCase, WagtailTestUtils):
         self.assertEqual(response.context['self'].depth, homepage.depth + 1)
         self.assertTrue(response.context['self'].path.startswith(homepage.path))
         self.assertEqual(response.context['self'].get_parent(), homepage)
+
+
+class TestRecentEditsPanel(TestCase, WagtailTestUtils):
+
+    def setUp(self):
+        # Find root page
+        self.root_page = Page.objects.get(id=2)
+
+        # Add child page
+        child_page = SimplePage(
+            title="Hello world!",
+            slug="hello-world",
+        )
+        self.root_page.add_child(instance=child_page)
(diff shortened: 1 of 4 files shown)


dirtycoder:
Finally, I was able to keep the commits from @khink and @mjec. :)

dirtycoder:
I saw it was postponed to 1.7 so just wanted to add a note.

I'm not very familiar with all the changes that are going on in Wagtail (as I don't actually use it) so if this PR diverge too much from the current code base I'll not be able to have the time to resolve the conflicts and update the PR when necessary (if the change are still in the interest of the project), so I'll need to "abandon" the PR.

The same applies to #2577.

gasman:
Have re-tested with our >100k page site, and there are no performance issues at all.

Merged in 3b4c24e0b59638fb695cfd0f2637b8e5bc595214 + parents - thanks @dirtycoder, @mjec and @khink!

PR #2622, opened 2016-05-16
Add some missing class="button" styles

gasman:
I'm pretty sure this catches all the remaining instances from #2524 now - I've grepped the codebase for `type="submit"` (and all variant quotes), `dropdown` and `<button`, and verified that they all have `class="button"` where appropriate.

Along the way I spotted a more serious issue with the up/down/delete controls in wagtailsearchpromotions - they were missing `type="button"`, which resulted in them submitting the form when you tried to use them. Will make a separate PR to backport this fix to 1.4.x.


gasman:
Merged in be1b2d3cad1a82d9d9244145673dec1432a75aa2

PR #2630, opened 2016-05-18
Don't wrap TableBlock rendering in a div

### LINKED ISSUE #2629: Don't wrap TableBlock in <div class="table">
The default rendering is a little too opinionated. It should just render a `<table>` element. We can determine how to target it with CSS on our own, by either using an existing parent element (ex. `main.content table`) or by overriding the template and wrapping the block in a div ourselves.

gasman:
fixes #2629

### DIFF
--- a/wagtail/contrib/table_block/templates/table_block/blocks/table.html
+++ b/wagtail/contrib/table_block/templates/table_block/blocks/table.html
@@ -1,9 +1,26 @@
-<div class="table">
-    <table>
-        {% if table_header %}
-            <thead>
-            <tr>
-                {% for column in table_header %}
+<table>
+    {% if table_header %}
+        <thead>
+        <tr>
+            {% for column in table_header %}
+                <th>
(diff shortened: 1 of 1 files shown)


Proper-Job:
@gasman the unit tests are failing for this. Do you want me to take a look at that and open a PR against this branch https://github.com/gasman/wagtail/commit/5bb23c41bc8586ff39d385a8fcc2f7e22c046cf1?
Should be a simple fix.

gasman:
Tests fixed and merged in 8ed30060fdd3e91bee608658e3741037b5fbff8e

alexgleason:
Thanks. :smile: Also merged into [alexgleason/wagtail_table_block](https://github.com/alexgleason/wagtail_table_block). Feels kinda weird going from a 1.0.0 to a 2.0.0 in a day, but I'm pretty sure that's the correct thing to do with semantic versioning since that extra div could break someone's code.

PR #2632, opened 2016-05-18
Rename table_block.fields to table_block.blocks

alexgleason:
Per [this comment](https://github.com/torchbox/wagtail/pull/1705#issuecomment-219940944), I think it makes more sense for this file to be called `blocks.py` since we're using it to import a StreamField block.

This makes the API consistent with the other blocks that Wagtail uses, such as:

* `wagtailcore.blocks`
* `wagtaildocs.blocks`
* `wagtailembeds.blocks`
* `wagtailimages.blocks`
* `wagtailsnippets.blocks`

`contrib.table_block.fields` is the odd one out and should probably be renamed to `contrib.table_block.blocks` for consistency in the API.

(Sorry for the long diff; didn't notice until now that my editor deleted those extra spaces. Probably better without them.)


gasman:
Merged in 8d5d29c2f3dac78d5f2eb9e98a48fb1134ca0894 - thanks!

PR #2407, opened 2016-03-28
Catch errors raised during search index auto update

kaedroho:
Fixes #2323

This prevents the admin interface crashing if there was an error while updating the search index.

The error is now caught and logged instead

PR #2633, opened 2016-05-18
Provide fallback implementation of unittest.assertLogs for Python <3.4

gasman:
Tests from #2407 are failing on Python <3.4 because unittest.assertLogs was only added in 3.4. This PR provides a fallback implementation, 'borrowed' from a (non-merged) Django pull request: https://github.com/django/django/pull/3467/commits/1b4f10963628bb68246c193f1124a1f7b6e4c696

### DIFF
--- a/wagtail/tests/assert_logs.py
+++ b/wagtail/tests/assert_logs.py
@@ -1,0 +1,84 @@
+import logging
+import sys
+from collections import namedtuple
+
+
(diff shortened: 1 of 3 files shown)


kaedroho:
Oops, sorry for not spotting this earlier. And thanks for pulling in an implementation -- I can see us using this in a few other places!

PR #2644, opened 2016-05-27
Update branding docs to remove django-overextends

### LINKED ISSUE #2609: Update 'custom branding' docs to ditch django-overextends on Django 1.9+
As observed in https://github.com/torchbox/wagtail/pull/2414#issuecomment-218782760, it appears that Django 1.9 has finally fixed circular template inheritance (https://docs.djangoproject.com/en/1.9/releases/1.9/#templates - "Django template loaders can now extend templates recursively"), which would seem to make django-overextends obsolete.

So, someone ought to:

* check that the custom branding snippets on http://docs.wagtail.io/en/v1.4.4/advanced_topics/customisation/branding.html work without django-overextends, with a plain `{% extends %}` tag
* if so, update that page with simplified instructions for Django 1.9, with a brief note at the bottom that Django 1.8 projects will need to use django-overextends.

toumaz:
Fixes #2609

### DIFF
--- a/docs/advanced_topics/customisation/branding.rst
+++ b/docs/advanced_topics/customisation/branding.rst
@@ -3,14 +3,16 @@
 Custom branding
 ===============
 
-In your projects with Wagtail, you may wish to replace elements such as the Wagtail logo within the admin interface with your own branding. This can be done through Django's template inheritance mechanism, along with the `django-overextends <https://github.com/stephenmcd/django-overextends>`_ package.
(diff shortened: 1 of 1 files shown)


SalahAdDin:
Wow, in django 1.9 i haven't to use overextends?

gasman:
Thanks @toumaz! Merged in 5ebc7d504410948d8592f397a20e5f84b774a21d.

PR #2645, opened 2016-05-28
Str funcs for group perms

coredumperror:
`GroupCollectionPermission` had a `__str__()` function, but `GroupPagePermission` didn't. So I added one.

Personally, I'd prefer that both classes used a string format more like django's `Permission` model:
```
    def __str__(self):
        return "%s | %s | %s" % (
            six.text_type(self.content_type.app_label),
            six.text_type(self.content_type),
            six.text_type(self.name))
```

But I didn't want to change `GroupCollectionPermission`'s existing implementation.

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -1516,6 +1516,7 @@ PAGE_PERMISSION_TYPE_CHOICES = [
 ]
 
 
+@python_2_unicode_compatible
 class GroupPagePermission(models.Model):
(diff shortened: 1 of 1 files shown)


coredumperror:
Oh damnit... I messed up my branch. Gotta recreate it...

coredumperror:
OK, I think that fixed it.

gasman:
Looks good, thanks! Merged in 6ae36c0baa279037821a85b208ac3f00eadea6a6.

PR #2484, opened 2016-04-16
Implement API v2 fields changes (RFC 5)

kaedroho:
See: https://github.com/torchbox/wagtail-rfcs/blob/master/accepted/005-wagtail-api-fields.rst

 - [x] Refactoring to give endpoint classes more control over the fields the serialisers will render
 - [x] "title" and "tags" no longer disappear when a user specifies custom set of fields
 - [x] Implemented '-' operator for removing fields
 - [x] ``*`` operator for adding all fields into the response
 - [x] Nested fields
 - [x] ``_`` operator for removing all fields from the response

This is implemented on top of the Admin API as I expect that'll be merged before this (and it makes development much easier since this affects both)

PR #2650, opened 2016-05-31
[WIP] Refactor API v2 serializers

kaedroho:
This pull request aims to make serializers in the API v2 module as dumb as possible. It has the following changes:
 - Remove field configuration (default fields/meta fields https://github.com/torchbox/wagtail/blob/master/wagtail/api/v2/serializers.py#L382-L388). This will instead be done in the endpoint class and passed into the serializer
 - Remove building of nested serializers (https://github.com/torchbox/wagtail/blob/master/wagtail/api/v2/serializers.py#L205-L211). Instead, all serializers will be created by the endpoint. Nested serialisers will be passed into their parents when the parents are constructed. This gives the endpoint more control over these nested serializers and also means the serializers can be used without Wagtail's API endpoints (so they can be used in a custom API for example).
 - There is one behavioural change, the default fields (title, tags, etc) are no longer removed when the user uses the ``?fields=`` parameter. This will become possible again when we implement RFC 5 though. I've dubbed these fields in the commit messages as "soft default fields".

The work here has been cherry-picked from #2484

The reason for this change is to make it possible to implement RFC 5, but it also simplifies the module a bit.


kaedroho:
This turned out to be more complicated than I initially thought, I'm going to reopen #2484

PR #2656, opened 2016-05-31
Fix #2637: Mark the RichText field as modified after removing embedded

### LINKED ISSUE #2637: Removal of embedded through "click-action" not being registered.
Seen with Python 3.4.3 and Wagtail 1.4.3

Steps to reproduce:
 1. Create a Page object with a StreamField (or RichTextField), make and run the migrations.
 1. Open the admin and create such page.
 1. Add a paragraph block in the StreamField and an embed inside.
 1. Save as draft.
 1. Now click the close button of the embed. See the embed block disappearing.
 1. Save as draft. Observe the embed block is back.

If after step 5, any other action had been taken, the changes would have been applied. This is because normal actions flag the field as modified, but the [close button does not](https://github.com/torchbox/wagtail/blob/4998c698198b86e735b501b33b252ad3fb03e4fc/wagtail/wagtailadmin/static_src/wagtailadmin/js/hallo-bootstrap.js#L75-83).

loicteixeira:
Fix #2637. Probably not the most elegant fix, but given the deprecation of hallo.js (see torchbox/wagtail#2409), I don't think this is worth putting much more effort into this.

The idea is that the parent `.richtext` has all widgets data accessible but instead of trying to match the widget that match the element being deleted, hook directly into the base hallo widget and flag the field as modified (it has to happen after removing the element though as it checks if it really is modified and fail otherwise).

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/hallo-bootstrap.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/hallo-bootstrap.js
@@ -76,8 +76,12 @@ function insertRichTextDeleteControl(elem) {
     var a = $('<a class="icon icon-cross text-replace delete-control">Delete</a>');
     $(elem).addClass('rich-text-deletable').prepend(a);
     a.click(function() {
+        var widget = $(elem).parent('.richtext').data('IKS-hallo');
         $(elem).fadeOut(function() {
             $(elem).remove();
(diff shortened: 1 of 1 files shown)


gasman:
Merged in c39f830ae24905a99b8cefb1041e0fa848109a91 (master) and 2fe0d18afddcd33ab5e824d035202adaa89e8d54 (1.5.x) - thanks!

Issue #1735, opened 2015-09-25
Unable to use models with a StreamField inside Inline Panels


cpsimpson:
I'm running into an error when I try to add a related model that has a StreamField defined.  Whenever I try to go to the admin page to edit or create a page of this type I get a "'NoneType' object is not iterable" error.  
Below are the (simplified) models in question for a better idea of what we are trying to do. ChapteredArticlePage is the type of page I am trying to create when I run into the error:

```python
class ChapteredArticlePage(Page):
    excerpt = RichTextField(blank=True, default="")
   body = BodyField()


ChapteredArticlePage.content_panels = Page.content_panels + [
    FieldPanel('excerpt'),
    StreamFieldPanel('body'),
    InlinePanel('chapters', label="Chapters"),
]


class Chapter(models.Model):
    heading = models.CharField(max_length=512, blank=True)
    body = BodyField()

    content_panels = [
        FieldPanel('heading'),
        StreamFieldPanel('body'),
    ]

    class Meta:
        abstract = True


class ArticleChapter(Orderable, Chapter):
    page = ParentalKey(ChapteredArticlePage, related_name='chapters')


class BodyField(StreamField):
    def __init__(self, block_types=None, **kwargs):
        block_types = [
            ('Heading', HeadingBlock()),
            ('Paragraph', ParagraphBlock()),
            ('Image', ImageBlock()),
            ('Embed', EmbedBlock(icon="site")),
            ('List', blocks.ListBlock(
                blocks.RichTextBlock(label="item"), icon="list-ul")
             ),
        ]

        super(BodyField, self).__init__(block_types, **kwargs)
```

(omitting the block class code, because it doesn't seem to matter what types of blocks are included for the error to occur). 

This was discussed on the support forum: https://groups.google.com/forum/#!searchin/wagtail/Streamfield/wagtail/L8nTkFZ2BeM/2o-pJwRNBgAJ

thenewguy:
I've encountered this as well

richardhewett:
I have also run into this although it only seems to happen with more complex block structures given to the ``StreamField`` when using the ``StreamField`` in related models.

This works even in related models

``item = StreamField(blocks.CharBlock())``

However as soon as you try to expand it to use a more complex structure it fails

``item = StreamField(blocks.StructBlock([
        ('image', ImageChooserBlock()),
        ('link', blocks.URLBlock(required=False))])``

[...]

PR #2659, opened 2016-06-01
fix: #1735 Unable to use models with a StreamField inside Inline Panels

Gagaro:
Please see #1735 for more discussion about this.

### DIFF
--- a/wagtail/wagtailadmin/edit_handlers.py
+++ b/wagtail/wagtailadmin/edit_handlers.py
@@ -621,6 +621,16 @@ class BaseInlinePanel(EditHandler):
             }
         }
 
+    @classmethod
+    def html_declarations(cls):
+        # We need to include html_declarations for StreamFields in the related model
+        declarations = [
+            field.stream_block.all_html_declarations()
+            for field in cls.related.related_model._meta.fields
+            if hasattr(field, 'stream_block')
+        ]
+        return mark_safe('\n'.join(declarations))
+
     def __init__(self, instance=None, form=None):
         super(BaseInlinePanel, self).__init__(instance=instance, form=form)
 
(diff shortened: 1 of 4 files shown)


Gagaro:
Any idea why the build fail?

```
$ git fetch --no-tags --depth=50 origin +refs/pull/2659/merge:
fatal: Couldn't find remote ref refs/pull/2659/merge
[info] build failed (exit code 1)
```

kaedroho:
Looks like it was trying to get a "merged" copy of the PR which doesn't exist as this PR isn't mergable.

rebasing or merging master (resolving the conflict) should solve the error

Gagaro:
Indeed, there was a commit one minute after I created the PR :sweat_smile: . Thanks :+1: .

Gagaro:
There is a regression with this PR. Some media are missing when they are needed in the FieldPanel but not in the InlinePanel.

```
Uncaught ReferenceError: makeHalloRichTextEditable is not defined
```

Gagaro:
Fixed

m1kola:
Hi @Gagaro!

Issue with `InlinePanel` was fixed in #2673. Thanks!

Gagaro:
Nice, I updated the pull request to only include the things specific to StreamField.

wagtail/wagtailadmin/edit_handlers.py
```diff
@@ -621,6 +621,16 @@ def required_formsets(cls):
...
         }
 
+    @classmethod
+    def html_declarations(cls):
+        # We need to include html_declarations for StreamFields in the related model
+        declarations = [
+            field.stream_block.all_html_declarations()
+            for field in cls.related.related_model._meta.fields
+            if hasattr(field, 'stream_block')
+        ]
```
gasman: We really shouldn't have any special-case code for StreamField (or any other field type) here - that would imply that something in the StreamField mechanism is failing to follow the same spec as everything else. Investigating...

gasman: Looks like all we really need here is `return cls.get_child_edit_handler_class().html_declarations()`. Packaging this up with some extra tests now...

gasman:
Merged in 16c279d705dfd944078a1c6e0e13c610d547e5e2. Thanks!

PR #2661, opened 2016-06-01
Fix #2654 Choose existing page for page links only

### LINKED ISSUE #2654: Hallo link editor attempts to edit document links as page links
Testing 1.5rc1- link editor checks for presence of `data-id` attribute when deciding whether to edit an existing page link which means it picks up document links and passes the value of this attribute to the chooser where it is wrongly interpreted as a document ID. It should check that `data-linktype` is `'page'`.

riceyrice:
Present new page chooser when editing any other link type.


gasman:
Merged in 05d9fceaa76dcc81d916c22936987dda04fee72f (master) and 3dd4948f9094769a985f7402a0528da5e090d110 (1.5.x) - thanks!

PR #2663, opened 2016-06-01
Fix #1732 Show asterisk after required StreamField block labels

### LINKED ISSUE #1732: Required StreamField blocks don't indicate that they're required
The user has click "Publish" then get an error before they can see that the field was required. The labels could have a red asterisk next to them to indicate that they're required.

riceyrice:
Fix for #1732. Not sure the SCSS is in the approved place but it works.
Also added a colon for consistency with normal field labels.

### DIFF
--- a/wagtail/wagtailadmin/templates/wagtailadmin/block_forms/struct.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/block_forms/struct.html
@@ -7,7 +7,7 @@
         {% for child in children.values %}
             <li{% if child.block.field.required %} class="required"{% endif %}>
                 {% if child.block.label %}
-                    <label{% if child.id_for_label %} for="{{ child.id_for_label }}"{% endif %}>{{ child.block.label }}</label>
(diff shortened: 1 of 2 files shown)


wagtail/wagtailadmin/templates/wagtailadmin/block_forms/struct.html
```diff
@@ -5,9 +5,9 @@
 
     <ul class="fields">
         {% for child in children.values %}
-            <li>
+            <li{% if child.block.field.required %} class="required"{% endif %}>
```
gasman: We probably want to avoid referring to `child.block.field` here - block objects don't always wrap a Django form field, and the ones that don't might have a different idea of how "required-ness" is defined. I think we can fix this nicely by adding a standard `required` property to Block (which would be false by default, and pick up the value from `self.field` in the case of FieldBlock) - I'll make this change.

gasman:
Merged in e9279c8512f8e05579094c3466e8f03a004382d7 + parents - thanks!

PR #2666, opened 2016-06-02
Only offer allowed parent pages in copy page destination chooser

riceyrice:
Prevent bypassing `parent_page_types`/`subpage_types` when copying a page.


m1kola:
It seems to me that we also need to check allowed parent page models in copy view (not only pass target models to the page chooser) and check `%model%.can_create_at(parent)`.

See relevant issue #2892, for example on why do we need a `%model%.can_create_at(parent)` check.

Thanks for the PR!

gasman:
I don't think it's absolutely necessary to add a view-level check in order for this PR to be accepted - as it stands this is still an improvement from a UI point of view.

However, we do need to fix it so it doesn't do this when parent_page_types/subpage_types aren't in use... :-)

screen shot 2016-09-02 at 16 27 16

PR #2669, opened 2016-06-02
Fix corruption of attributes containing double-quotes (img alt)

riceyrice:
This came up in developing/testing a Wagtail project - firstly img tag alt attributes containing a `"` were truncated at the first `"` by the web browser when inserted in the editor; secondly once they were escaped correctly and could be saved to the database, the data-alt attribute was lost and consequently raised a `KeyError` when read back from the database due to BeautifulSoup/Whitelister decoding `alt="&quot;"` as `alt='"'`.

While there I've also escaped `Format.name` as well as `alt_text` as while a format name probably shouldn't contain a `"` there's no rule against it.

### DIFF
--- a/wagtail/wagtailcore/rich_text.py
+++ b/wagtail/wagtailcore/rich_text.py
@@ -143,7 +143,7 @@ class DbWhitelister(Whitelister):
 
 FIND_A_TAG = re.compile(r'<a(\b[^>]*)>')
 FIND_EMBED_TAG = re.compile(r'<embed(\b[^>]*)/>')
-FIND_ATTRS = re.compile(r'([\w-]+)\="([^"]*)"')
+FIND_ATTRS = re.compile(r'([\w-]+)\=(?:"([^"]*)"|\'([^\']*)\')')
 
 
 def extract_attrs(attr_string):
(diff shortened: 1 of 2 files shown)


gasman:
Hmm, looks like this is more complicated than it appears at first glance...

The code in wagtailimages/formats.py for converting the DB representation into a real `` tag intentionally doesn't escape strings, on the basis that `extract_attrs` in wagtailcore/rich_text.py doesn't un-escape them - i.e. whatever escaping was in place in the DB version is left intact in the final HTML.

Unfortunately, it seems that BeautifulSoup tries to be "helpful" by using single-quotes where possible to allow double-quotes to appear un-escaped in the string. You've accounted for this in your fix to wagtailcore/rich_text.py, but where the old code made the faulty assumption that the strings reaching formats.py were consistently escaped, your new code assumes that they're consistently *un-escaped*, which isn't the case either (because BeautifulSoup can't avoid escaping if the string contains both single and double quotes, or `<` / `>` characters).

I think the only real fix is to add the necessary string handling to `extract_attrs` to ensure that strings are either consistently escaped, or consistently un-escaped. Consistently un-escaped would be nicer from an API-cleanliness point of view, but would be harder to implement (because there are hundreds of HTML entities that we might theoretically need to handle in HTML-unescaping code, whereas HTML-escaping code only needs to escape `<` `>` and `"` in order to be correct).

gasman:
(bumping to 1.6 as this may need a non-trivial amount of further work, and possibly some API-breakage)

riceyrice:
Ah yeah, I see what you mean. This PR is derived from work on a different branch and I'm sure we had it all round-tripping fine so I may have missed something. I will double-check when I get chance.

gasman:
Superseded by #3879

PR #2670, opened 2016-06-03
fix modeladmin delete view message

ychab:
We must build the message before trying to delete it ;-)

Tell me if it needs test

thanks


tomdyson:
Thanks, @ychab! @ababic, can you check this please?

ababic:
👍  From me

ababic:
@ychab, so that I can learn from this, what were your motivations for this
change? Were you seeing errors? Was it not working as expected in a
specific case? Or, is it more a coding practice / housekeeping thing?

If it wasn't working, I wonder if we need to think about adding additional
tests.

ychab:
@ababic, just try to delete an instance from the model admin UI and you will got the following error : 

````
TypeError: __str__ returned non-string (type NoneType)
````
This error happen because the instance is deleted and thus, it's try to call ``__str__`` method on a ``None`` *object*.

This is a minor bug so I don't think it needs a test just for this

ababic:
@ychab thanks for confirming. I don't think I spotted this originally,
because `delete()` should only trigger a database delete, and leave the
object in memory in-tact, so creating the message shouldn't be an issue in
simple cases. Though, I guess I neglected to consider situations where
models might have a `__unicode__` or `__str__` method that utilises data
from other database tables (where the referenced objects are likely to no
longer exist, having been deleted themselves as the database delete
cascaded). Definitely a lesson to bare in mind for the future. Thanks for
spotting!

ychab:
@ababic, in fact, the tested model has just one charfield and output its raw value in ``__str__()`` method...

I can simply reproduce it by creating an instance, deleting it and calling its __str__ method. For example : 

```` python
>>> foo = Foo.objects.create()
>>> foo.delete()
>>> str(foo)
TypeError: __str__ returned non-string (type NoneType)
````
I don't have any time yet to dig into (but I'm curious to know what really happen). But as a conclusion, we should never trust a deleted instance ;-)

PS : I have just fixed the flake8 warning

gasman:
Thanks @ychab!

I think this does need a test... if it really does affect deletion of every model, then that's a serious bug, and the fact that we didn't catch it already means that we clearly don't have enough tests. If it *doesn't* affect every model, then we need tests to help us understand the bug better (and make sure we don't re-introduce it again in future) :-)

ychab:
@gasman , Agree! TBH, as it is *just* an end-user message bug, I haven't found the courage to do it! Anyway, I think it should happen for every model so yes, it is a serious bug. I will try to do it as soon as possible.

ychab:
@ababic, @gasman ok, I think I catch it! Sorry, I was wrong. This happen not for every model. It happen because the model use its primary key in __str__ method! And once the instance is delete, of course its pk is None!

I will update test app to add a new model for this special use case and a new testcase (because in TestDeleteViewWithProtectedRelation, it doesn't really make sense)

ychab:
@ababic, @gasman: basic test done!

gasman:
Thanks! Merged in 3ba4064d8921e11d6aeed45a1e295f7a5bef4acb (master) and c42b2dbd5972b2ae911fe77b3f442c4d6159e8e2 (1.5.x).

PR #2671, opened 2016-06-03
Ensure Search tab pagination is retained on invalid image upload

riceyrice:
Otherwise hidden Search tab causes web browser to load _every_ image in the system.

To trigger this bug, go to insert an image in a rich text field, click Upload tab, enter a title and select a non-image file. Click Upload. Search tab (not visible) will change to be ordered by title with no pagination and your web browser will request a rendition of every image in the system. Which can cause your web browser to grind to a halt, as well as be a bit confusing when switching back to the Search tab.

### DIFF
--- a/wagtail/wagtailimages/views/chooser.py
+++ b/wagtail/wagtailimages/views/chooser.py
@@ -144,7 +144,8 @@ def chooser_upload(request):
     else:
         form = ImageForm(user=request.user)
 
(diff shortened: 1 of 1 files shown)


gasman:
Thanks! Merged (with test added) in cd0f8381cddb875172d15bc904fca5ebeadc43fe (master) and 444260a (1.5.x).

PR #2672, opened 2016-06-03
Allow editing of images and embeds by passing params in querystring

riceyrice:
Allows Wagtail TinyMCE editor to edit existing images and embeds by passing current values in querystring.
Hallo editor requires development to take advantage of this capability.


gasman:
Merged (with tests added) in d6b97351411adcd967aae75a8fdeaf354d8a61d2 - thanks!

PR #2681, opened 2016-06-03
Track parent page for editing internal links in rich text

### LINKED ISSUE #2639: Editing an 'Internal link' via the RichText editor doesn't indicate the target page very well
When running `1.5rc1`, all I see is a breadcrumb for the target page, but the target page itself doesn't seem to be clearly indicated anywhere.

gasman:
Fixes #2639 - changes the behaviour of editing existing internal links so that the page chooser opens up at the link's parent page, rather than at the page itself.

Various moving parts to this fix, which are probably worth explaining here:

* In `rich_text.py`, the link handler for `linktype="page"` links has been updated to add a `data-parent-id` attribute when it expands links in editor mode. (This attribute is ignored when going the other way, so the database representation is unchanged.)
* `hallo-wagtaillink.js` now picks up that attribute instead of `data-id` when opening the popup to edit an exiting page link.
* `_page_title_choose.html` now adds a `data-parent-id` attribute to the page links inside the chooser, which causes this to be included as `parentId` in the dictionary of page info returned from the chooser.
* `hallo-wagtaillink.js` also writes this back to the generated link in the rich text field, so that editing a just-inserted link works consistently with ones that have come from the database.

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/hallo-plugins/hallo-wagtaillink.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/hallo-plugins/hallo-wagtaillink.js
@@ -45,14 +45,13 @@
 
                     if (enclosingLink) {
                         href = enclosingLink.getAttribute('href');
-                        pageId = enclosingLink.getAttribute('data-id');
+                        parentPageId = enclosingLink.getAttribute('data-parent-id');
                         linkType = enclosingLink.getAttribute('data-linktype');
 
                         urlParams['link_text'] = enclosingLink.innerText;
 
-                        if (linkType == 'page' && pageId) {
-                            // TODO: Actually show the parent not the page itself.
-                            url = window.chooserUrls.pageChooser + pageId.toString() + '/';
+                        if (linkType == 'page' && parentPageId) {
+                            url = window.chooserUrls.pageChooser + parentPageId.toString() + '/';
                         } else if (href.startsWith('mailto:')) {
                             url = window.chooserUrls.emailLinkChooser;
                             href = href.replace('mailto:', '');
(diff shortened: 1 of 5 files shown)


m1kola:
It looks like new a feature because Wagtail 1.4.* behaves the same way as 1.5. Are you sure that we need to merge this into 1.5.1?

m1kola:
Hi @gasman!

Now chooser shows list of pages string from the parent page (in my case it's "Blog index"), but... Can you guess which page was chosen? :)
screen shot 2016-06-04 at 19 13 01

gasman:
@m1kola Sure, I know it's not ideal, but I think it's an improvement over the old behaviour - in particular, this makes it consistent with the behaviour of PageChooserPanel (if I'm not mistaken).

m1kola:
> in particular, this makes it consistent with the behaviour of PageChooserPanel (if I'm not mistaken)

Right. I think this solution is ready to merge into 1.5.1. Do we have an issue for UX improvement?

m1kola:
Merged in 0e5eb054c88ff6e40367d5ba96990b4ac38b54df + a0878256d88f59fe4b070c98db612b4ec74bfdcf for release notes (`stable/1.5.x`) aacb1116dbee7408174655cf7016e040b174ab8b (`master`).

Thanks!

gasman:
> Do we have an issue for UX improvement?

For highlighting the currently-chosen page within the listing, you mean? No, I don't think there's an issue for that - please feel free to create one :-)

m1kola:
> For highlighting the currently-chosen page within the listing, you mean? No, I don't think there's an issue for that - please feel free to create one :-)

Yes. I haven't found any issues, but I was afraid that I used wrong keywords. That's why I decided to ask you :) I will create an issue a bit later.

PR #2686, opened 2016-06-07
StructBlock - get_prep_value to return OrderedDict

vincentaudebert:
`get_prep_value` should return an OrderedDict so we keep the order for the fields inside the StructBlock.

This is needed for the rewriting of Streamfield in React.

### DIFF
--- a/wagtail/wagtailcore/blocks/struct_block.py
+++ b/wagtail/wagtailcore/blocks/struct_block.py
@@ -120,7 +120,7 @@ class BaseStructBlock(Block):
 
     def get_prep_value(self, value):
         # recursively call get_prep_value on children and return as a plain dict
-        return dict([
+        return collections.OrderedDict([
             (name, self.child_blocks[name].get_prep_value(val))
             for name, val in value.items()
         ])


gasman:
Hi @vincentaudebert,
This seems like a bad idea to me. `get_prep_value` exists primarily to provide a JSON-compatible representation of the streamfield data, as opposed to the 'native python' representation which might contain model instances, datetimes and so on - and OrderedDicts don't survive the round trip to JSON and back (JSON objects are defined as unordered, as per http://json.org/). Once one component starts to rely on key order being preserved, then all other components either need to preserve ordering everywhere (which isn't possible using standard JSON tools), or keep track of which values are ordered and which ones aren't (which will be a nightmare to maintain).

Is there any reason you can't obtain the ordering from the StructBlock definition? (Presumably you must have access to the StructBlock definition on the React side, or else you wouldn't know what field types to render.) The standard pattern for iterating over a StructBlock value is:

    for (block_name, block_def) in my_structblock_definition.blocks:
        do_something_with(my_structblock_value[block_name], block_def)

vincentaudebert:
@gasman The idea was to be able to get the right order then I have to run a loop again to switch everything to a list of objects with type/value.
Order in a list are preserved in JSON.
If it helps to understand here is my code:
```
        value = self.bound_field.value()
        json_value = self.block_def.get_prep_value(value)

        # this is to keep the order of children on complex blocks
        for obj in json_value:
            if isinstance(obj['value'], dict):
                to_arr = []
                for key in obj['value']:
                    to_arr.append({
                        'type': key,
                        'value': obj['value'][key]
                    })
                obj['value'] = to_arr

        return json.dumps(json_value, cls=DjangoJSONEncoder)
```

Having it on the StructBlock class avoids to add a new get_prep_value to all the children.

On a side note, the data structure is really not consistent. Sometimes blocks are given as dict (which is a problem for the order) and sometimes it's a list of dicts.
Quite confusing :)

I will try to apply directly your loop :)
I just thought it would be nice to have the possibility to get it directly from the object.

PR #2688, opened 2016-06-07
Use python_2_unicode_compatible in snippets.rst

Gagaro:
We should use `python_2_unicode_compatible` in the documentation so that new users learn the best practice.


gasman:
Good idea - thanks!

PR #2689, opened 2016-06-07
Make 'permanent' / 'temporary' label on redirects translatable, instead of classname

gasman:
Reported by @leo-naeka on Transifex. The 'primary' classname on the 'permanent' / 'temporary' label in redirect listings was wrongly being translated (meaning that the style was not applied in non-English localisations); conversely, the label itself wasn't being translated...


gasman:
Merged in aaceb9b08f0115c68f62e1d067d88886f3d74eeb (master) and 7c2130f3c24cb94e6d4b38a6aaac0cc09a5b57ad (1.5.x).

PR #2691, opened 2016-06-08
Load staticfiles not static in document and email templates.

jonnyscholes:
This broke the document uploader on S3 for us.


m1kola:
Looks good. I've found usages of `static` in the `project_template`, but I don't think that we need to change it to `staticfiles` because `project_template` uses default `STATICFILES_STORAGE` and implementer can change templates, if he decide to use S3 or something else.

Thanks @jonnyscholes!

m1kola:
Merged in ba220d58ca8f4a8f32452a69e1ac7b502bbb3162 + parents. Thanks again!

PR #2694, opened 2016-06-08
#1446 add pagination to wagtailadmin.views.pages.move_choose_destination

### LINKED ISSUE #1446: Move page / choose destination view needs pagination
The 'choose destination' vlew when moving a page does not paginate the listing, so this is unusable on sections with thousands of subpages. Related: #508

nimasmi:
Pagination in the `move_choose_destination` function, to fix #1446.

No unit tests added because we couldn't see anywhere that pagination was currently being tested in wagtailadmin.tests.test_page_views.

Note that when this view is loaded with the named URL `move` then pagination links will redirect to the two-positional-arg version, as the pagination link in the template always supplies the `viewed_page_id` argument.

Credit due also to Žan Anderle, @zanderle


gasman:
Merged in 504759d267a98cf9061d89a588eed1a573eb51db - thanks!

PR #2697, opened 2016-06-09
Allow passing of options into Rich Text Editor.

jaydensmith:
Can we allow an options dict to be passed into the rich text editor?
That would allow for some much cleaner configuration and also you'd be able to configure different instances of the same editor.

Sorry if this is very wrong, it's my first time doing a pull request.


gasman:
Merged in 5732e215b0774c22d8e9bf210aa1491353595406 - thanks!

Sorry about the delay in reviewing this - I was a bit hesitant about making this change because I'd prefer rich text fields to be configurable in a way that's common to all backends (so that you could specify, for example, that images aren't allowed, and have the server-side validation / whitelisting enforce that too). But it's clearly going to take a while to get round to implementing that, so it didn't make any sense to postpone this PR any longer.

PR #2699, opened 2016-06-09
Bugfix: Dont use localize when rendering js template, resolves locali…

lovmat:
…zation issue when having "USE_THOUSAND_SEPARATOR = True", variable {{ self.formset.max_num }}

### DIFF
--- a/wagtail/wagtailadmin/templates/wagtailadmin/edit_handlers/inline_panel.js
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/edit_handlers/inline_panel.js
@@ -1,3 +1,5 @@
+{% load l10n %}
+{% localize off %}
 (function() {
     var panel = InlinePanel({
         formsetPrefix: "id_{{ self.formset.prefix }}",
@@ -13,3 +15,4 @@
     panel.updateMoveButtonDisabledStates();
     panel.updateAddButtonState();
 })();
+{% endlocalize %}


gasman:
Hi @lovmat - please can you describe the specific bug that's being fixed here (i.e. how to replicate it)?

lovmat:
Using Djangos USE_THOUSAND_SEPARATOR setting (docs: https://docs.djangoproject.com/en/1.9/ref/settings/#std:setting-USE_THOUSAND_SEPARATOR) will format numbers printed in a template according to the current localization. For instance here in Sweden we format with a space (e.g. 30000 will be printed as 30 000). With this setting enabled the js code will break since 30 000 will be printed in the js template instead of 30000. This pull request fixes this so all numbers will be printed without localization in mind (which is what we want for a js file like this).

gasman:
Thanks for the clarification! This is the first time I've encountered the `USE_THOUSAND_SEPARATOR` setting, and I have to say that from first appearances, it seems to be a really bad design on Django's part - there are all kinds of reasons to output machine-readable numbers inside template code, and surely it makes much more sense to use `{{ ...|intcomma }}` for the times when you specifically want it to be human-readable.

I'm concerned that even if we fix this specific bug, there may be any number of more subtle bugs, that only occur on (for example) page IDs above 1000, or images larger than 1000x1000. It seems like it might actually be safer for Wagtail to *not* support `USE_THOUSAND_SEPARATOR` (and possibly output a warning on startup if it's enabled). Thoughts...?

lovmat:
Since it's Django core functionality my personal opinion is that it should be supported by Wagtail.

I did a quick scan for more .js files but could not find any others that were rendered the same way as this one.

Django does the same thing in the core contrib gis app:
https://github.com/django/django/blob/master/django/contrib/gis/templates/gis/admin/openlayers.js#L74

lovmat:
You are right - there could be more of them. This is the only bug we've encountered though since we started using Wagtail.

Another approach, used in django admin, would be to json encode the data that needs to be available in javascript. For example:
https://github.com/django/django/blob/master/django/contrib/admin/static/admin/js/popup_response.js#L4

gasman:
Fixed in 674af5586b4eb4488457fd6762e24bff7b11d69f (with a slightly different approach, to make the code a bit more explicit about why the fix is needed.)

I agree, it's worth fixing this even if there are other potential undiscovered bugs with using USE_THOUSAND_SEPARATOR. Thanks for the report and the patch!

PR #2702, opened 2016-06-10
Ensure form submission listing is explicitly ordered by submit_time

gasman:
The form submission listing/export view in wagtailforms was not explicitly ordered by submit_time - it relied on the database's internal ordering, resulting in occasional test failures, e.g. https://travis-ci.org/torchbox/wagtail/jobs/135920555

### DIFF
--- a/wagtail/wagtailforms/views.py
+++ b/wagtail/wagtailforms/views.py
@@ -54,7 +54,7 @@ def list_submissions(request, page_id):
 
     data_fields = form_page.get_data_fields()
 
-    submissions = form_submission_class.objects.filter(page=form_page)
+    submissions = form_submission_class.objects.filter(page=form_page).order_by('submit_time')
     data_headings = [label for name, label in data_fields]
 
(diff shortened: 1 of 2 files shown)


kaedroho:
Looks like this needs a rebase

wagtail/wagtailforms/views.py
```diff
@@ -56,7 +56,7 @@ def list_submissions(request, page_id):
         for field in form_page.form_fields.all()
     ]
 
-    submissions = FormSubmission.objects.filter(page=form_page)
+    submissions = FormSubmission.objects.filter(page=form_page).order_by('submit_time')
```
kaedroho: I think it would probably make sense to have most recent submissions on the first page

gasman: Maybe, but that's not what I'm fixing here :-)

gasman:
Rebased

kaedroho:
Thanks!

PR #2705, opened 2016-06-10
Remove TagSearchable class

gasman:
Supersedes #2436, with differences / additions as follows:

* `get_indexed_objects` was dropped in #2498 and no longer needs to be moved
* `popular_tags` is now the helper function `wagtail.wagtailadmin.utils.popular_tags_for_model` instead of a class method
* the deprecation warning now uses the correct `stacklevel` parameter to point back to the calling module
* references to `TagSearchable` in migrations have been removed (and a release note added telling users to do the same for custom image/document models)

### DIFF
--- a/docs/releases/1.6.rst
+++ b/docs/releases/1.6.rst
@@ -30,3 +30,29 @@ Bug fixes
 Upgrade considerations
 ======================
 
+``TagSearchable`` needs removing from custom image / document model migrations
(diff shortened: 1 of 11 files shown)


kaedroho:
Rebased and merged in 367c71dbaa070e18012d2c9d431ee17cc500c572 +parents. Some minor coding style tweaks made in 7d490f7a6fa5fa127b396584389c0f70db167b8d

Cheers! :beers:

PR #2709, opened 2016-06-13
Implement `make develop` to ease contributing to wagtail

mvantellingen:
This installs wagtail as editable package in your current python env
and also adds the testing and docs extras

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: clean-pyc
+.PHONY: clean-pyc develop
 
 help:
 	@echo "clean-pyc - remove Python file artifacts"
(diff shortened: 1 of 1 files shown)


mvantellingen:
I also think we need to list all targets in the PHONY directive

gasman:
Hi @mvantellingen! I'd be happy to add the makefile option, but I'm not so keen on the idea of advising developers to create Wagtail projects inside the Wagtail source tree. Besides causing problems with version control (we can't rely on them to use the exact name `mysite`...) I think there's a possibility that it could trip up code coverage / linting tools, and it just generally seems like quite a confusing and non-standard way to arrange things.

mvantellingen:
Hi @gasman :-) I actually agree. I copied the https://github.com/mvantellingen/wagtail#getting-started a bit too literally. Any suggestons ?

mvantellingen:
I've decided to keep it simple for now and remove the additional output. In the future it might be helpful to output a link a contributing doc.

Besides that I've also added the ``npm install && npm run build`` command. It's up to the developer to make sure that npm is available (and that a virtualenv is active).

gasman:
Merged, thanks!

PR #2713, opened 2016-06-15
Fix #536: Show non_field_errors

### LINKED ISSUE #536: Show NON_FIELD_ERRORS when page does not pass validation
I am not using the latest version yet, so this may already be fixed, but...

I am trying to use the Model.clean() method on a page type to validate a few conditional fields that depend on another model field.

Along the lines of if page "mode" is 1, fields "foo" and "bar" should not be 3 or 5.

When I raise a validation error the dashboard shows "The page could not be saved due to validation errors" but it doesn't output the error messages.

How are other people doing this type of validation now?

JoshBarr:
This addresses issue #536, the ability to show non_field_errors on the admin UI.

Also allows for the messages block to be overridden by custom admin templates, if desired.

Agree more work could be done on the UX of this feature, however I don't see that as a blocker to including this fix.

### DIFF
--- a/wagtail/wagtailadmin/templates/wagtailadmin/base.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/base.html
@@ -35,11 +35,12 @@
             <div class="messages">
                 {% if messages or edit_handler and edit_handler.form and edit_handler.form.non_field_errors %}
                     <ul>
-                        {% for message in messages %}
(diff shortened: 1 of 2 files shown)


tomdyson:
See also #2783

gasman:
Closing as per https://github.com/wagtail/wagtail/issues/536#issuecomment-285685185 - will open a new PR to address this shortly.

PR #2714, opened 2016-06-15
Fix ForeignKey selectbox in ChooserBlock

### LINKED ISSUE #2218: PageChooserBlock/ImageChooserBlock values go null when restoring from loaddata

To reproduce:
- `wagtail start testdata`
  
  ``` python
  
  
  class HomePage(Page):
  body = StreamField([
  ])
  
      StreamFieldPanel('body'),
  ]
  ```
  
  ``` shell
  ```

gannetson:
If you have a related model in your page through a ChooserBlock the selected value isn't shown on a second load (after save). This change should fix that.

Fixes #2218

### DIFF
--- a/wagtail/wagtailcore/blocks/field_block.py
+++ b/wagtail/wagtailcore/blocks/field_block.py
@@ -416,6 +416,13 @@ class ChooserBlock(FieldBlock):
             except self.target_model.DoesNotExist:
                 return None
 
+    def value_for_form(self, value):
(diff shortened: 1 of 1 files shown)


gasman:
@gannetson Thanks! Would you be able to add a unit test for this, please?

mvantellingen:
@gasman I ran into the same issue, added a test for this change in #3169

gasman:
Superseded by #3169

PR #2715, opened 2016-06-15
Make unlink action register a change to rich text areas

### LINKED ISSUE #2704: 'Remove link' functionality in rich text editor not working
**Expected behaviour**: clicking 'remove link' on a link in a `RichTextBlock` should remove the anchor and persist the change on save.
**Observed behaviour**: clicking 'remove link' removes the anchor but doesn't persist the change on save.

I'm not sure what's going on, but if you click inside the linked word, or select the linked word and click 'remove link' it appears to be removed immediately, but comes back on 'save draft'. If you select the linked text and delete it and click save, the text of the anchor tag is re-inserted into the rich text block on save.

I've set up a clean Wagtail 1.5.2 project (using `wagtail start`) and added a `StreamField` with a `RichTextBlock` to check this isn't something we've accidentally introduced in our project.

unlink

gasman:
Fixes #2704


jimr:
I can confirm this fixes the 'remove link' problem.

PR #2720, opened 2016-06-16
Compatibility with Django 1.10

m1kola:


gasman:
Re removing SubfieldBase, #2119 is relevant.

wagtail/wagtailcore/models.py
```diff
@@ -257,7 +257,7 @@ class PageBase(models.base.ModelBase):
     def __init__(cls, name, bases, dct):
         super(PageBase, cls).__init__(name, bases, dct)
 
-        if cls._deferred:
+        if hasattr(cls, '_deferred'):
```
m1kola: I think that it will not work. I've found some useful docs:
* https://docs.djangoproject.com/en/1.10/ref/models/instances/#customizing-model-loading
* https://docs.djangoproject.com/en/1.9/ref/models/instances/#customizing-model-loading

m1kola:
This PR will be replaced separated chunks. See: #2723

PR #2721, opened 2016-06-16
Fix #2533: Prevent removal of '/' when redirect from url is just '/'

### LINKED ISSUE #2533: Cannot Edit A Redirect When Set To Just A Forward Slash
I attempted to redirect from the site root to an existing page. E.g. _example.com/_ to _example.com/blog/_. I assumed "Redirect from:" should be set to "/", as all other redirects start with a forward slash. The redirect is successfully added, however, the FROM column field is blank and you cannot click on the redirect to edit it. It also doesn't appear to be functional.

Taking a look at the [following code](https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailredirects/models.py#L55-L68), one can see that it strips any extra white space and removes a trailing slash. If the string is set to only a forward slash, the string is now empty. This doesn't seem to explain why multiple slashes would be set to an empty string, but that's as far as I looked into this issue.

How do we prevent the user from creating a redirect with an empty `old_path`?

If supported, how do we properly redirect from the site root?

ltakens:


tomdyson:
Fixes #2533

tomdyson:
Thanks @ltakens and @kingel! This fixes the reported error, but I'm not sure that the fix will ever be used: as @gasman said in https://github.com/torchbox/wagtail/issues/2533#issuecomment-215362345 "Redirects only take effect on URLs that don't exist as pages (i.e. would return a 404 if the redirect wasn't there), so setting one on the root URL generally won't work".

gasman:
Thanks! As per @tomdyson's comment above, I was imagining that we'd fix this by disallowing both the empty string and `/` - but since it *is* possible in principle to create a working redirect at `/` (by pointing the site record at an unpublished page), I think this is actually the more correct solution.

PR #2722, opened 2016-06-16
a few too many quote's

nutztherookie:
the `pk` got `quote`d twice a bunch of times

``` python
    pk = quote(getattr(obj, self.opts.pk.attname))
    self.url_helper.get_action_url('edit', quote(pk))   # < this was a little unneccessary, double function calls in multiple locations
```


ababic:
This looks good to me! thanks for the spot @nutztherookie!

gasman:
Thanks @nutztherookie! Any chance you could add some unit tests to confirm that things are now being quoted the correct number of times, please?

nutztherookie:
@gasman I didn't realize calling the function multiple times would alter the result multiple times; Maybe it does need to be called twice!  ..

ababic:
Hi guys. I managed to replicate this issue locally by creating a model with a `CharField` as the primary key, then creating an objects with '/', '_', ':' and similarly problematic character in the value for that field. Under these circumstances, the double quoting means that the edit and delete views do not recognise the primary key from the URL if edit or delete buttons in the listing are used.

Repeating the changes made in this pull request seem to fix these issues.

nutztherookie:
ah awesome =)
yeah.. I fiddled around a bit as well and I found `quote`ing twice might be harmful. Thank you for confirming.

gasman:
Rebased (with some further fixes) in #3520.

nutztherookie:
:) gracias

PR #2724, opened 2016-06-16
CI configuration for Django 1.10

m1kola:
See #2723

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,17 +56,7 @@ matrix:
      python: 2.7
    - env: TOXENV=py27-dj110-postgres-elasticsearch
      python: 2.7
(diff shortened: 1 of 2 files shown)


gasman:
Hi @m1kola - I don't think we should run Travis against every permutation of python version, database and elasticsearch, as that will tie up lots of Travis instances and create a long backlog. I would suggest cutting it down to:

py27-dj110-sqlite-noelasticsearch
py27-dj110-postgres-noelasticsearch
py27-dj110-mysql-noelasticsearch
py27-dj110-postgres-elasticsearch
py34-dj110-postgres-noelasticsearch
py35-dj110-sqlite-noelasticsearch
py35-dj110-postgres-noelasticsearch
py35-dj110-mysql-noelasticsearch
py35-dj110-postgres-elasticsearch

PR #2727, opened 2016-06-16
Add new EmailBlock and IntegerBlock

OktayAltay:


ababic:
Hey @OktayAltay! Finally resolved out my environment issues locally, and managed to get this running to check out your changes. It seems that the `min_num` and `max_num` attributes aren't causing validation errors to kick in when entering values that don't obey those rules. I'm not 100% sure of the cause of that, as it looks like you're passing those arguments on to the IntegerField widget in the right way. Any ideas?

ababic:
@OktayAltay My bad! I'm using arguments that don't exist! Having changed `min_num` and `max_num` arguments to `min_value` and `max_value`, it works as expected :)

gasman:
Merged in 31ce2e802c2914c0e58d983c089f53f60a972ac8 - thanks!

Issue #2522, opened 2016-04-26
gulp-sass dependency management is too brittle


gasman:
Related: #2518. While it's a good thing that we were able to track down and fix an issue with our SASS, it did mean that we were blocked for most of a day in setting up a new developer with a Wagtail installation, and I'd like to be sure that we aren't bitten by this again in future. Now that I've dug into the details, it seems that the sequence of events was:

* The SASS project decided that version 3.3.0 was a good time to change the variable scoping rules of the language. (To be fair, they did document it under a heading labelled ["Backwards Incompatibilities – Must Read!"](http://sass-lang.com/documentation/file.SASS_CHANGELOG.html#backwards_incompatibilities__must_read_2)...)
* libsass implemented the corresponding change ("Variables in control flow should not be accessible in outer scopes") in [version 3.3.3](https://github.com/sass/libsass/releases/tag/3.3.3), which, despite supposedly being a patch release, is apparently ["massive"](https://github.com/sass/node-sass/releases/tag/v3.5.1) and "the latest and most awesome LibSass release ever".
* node-sass released [version 3.5.1](https://github.com/sass/node-sass/releases/tag/v3.5.1) 6 days ago (previous was 3.4.2), followed by a flurry of follow-up releases, bringing it up to date with libsass 3.3.6.
* gulp-sass [depends](https://github.com/dlmanning/gulp-sass/blob/v2.3.1/package.json#L25) on `"node-sass": "^3.4.2"`, i.e. any 3.x.x release. In other words, it trusts node-sass to do semantic versioning correctly, which it doesn't.
* As a result, gulp-sass was quite happy to install a backwards-incompatible version of node-sass / libsass with no warning.

Since most Wagtail developers just want the SASS compilation to work and aren't bothered about having bleeding-edge versions, I think it makes sense to pin our NPM dependencies at exact versions. I gather [`npm shrinkwrap`](https://docs.npmjs.com/cli/shrinkwrap) is the standard way of doing that, but by default it doesn't consider the `devDependencies` list, which is where gulp-sass is currently listed. Given that Wagtail end-users don't touch npm at all, I'm not clear on what the distinction between `devDependencies` and `dependencies` is here - would it make sense to move everything to `dependencies`? (ping @JoshBarr)

JoshBarr:
`devDependencies` is definitely the right place for tooling that's not required at runtime. 

[...]

PR #2730, opened 2016-06-16
Added .nvmrc and updated the documentation

dennishendrikslukkien:
See issue on: https://github.com/torchbox/wagtail/issues/2648


docs/contributing/developing.rst
```diff
@@ -13,7 +13,23 @@ If you'd prefer to set up all the components manually, read on. These instructio
...
+
+.. code-block:: sh
+
+    nvm install v5.3.0
+
+And then in any new shell just use the installed version:
+
+.. code-block:: sh
+
+    nvm use v5.3.0
```
jjanssen: When the `.nvmrc` is available it is not necessary to explicitily call `nvm use v5.3,0`. You can just run `nvm install` & `nvm use` from within the root of the project.

nfletton:
Hello @dennishendrikslukkien. Having a .nvmrc file makes sense and I use one locally. My comment  though is that adding one to the repo is locking in the development to lowest supported node version where currently the docs  specify anything from 5.3.0 upwards. For instance I just tried 6 in my .nvmrc and with node 6.2.1 everything builds just fine.

I don't think nvm supports any version prefixes like >5.3.0 unfortunately

gasman:
I think the instructions here need to be reordered - we're relying on the `.nvmrc` file to specify the Node version, but the instructions tell you to run `nvm install` before cloning the Wagtail repo, and `.nvmrc` won't exist at that point...

@nfletton After my experiences with Node / NPM version management in #2522, I'd much prefer to pin Node at an old version (and say so in the docs) rather than asserting that all future versions will continue to work.

thibaudcolas:
_Fortunately_ neither nvm nor the `package.json` enforce the version people actually run, so we get the best of both words. We can specify an old version we are certain works, and people are free to use something else.

IMHO the `nvmrc` should just specify the major version and minor versions - `5.3` (or just major, 5?) instead of locking down to the patch level. Same in the package.json it should be `~5.3.0` instead of an exact `5.3.0`.

For the version, I would suggest to go with node 4 everywhere (current LTS), but node 6 will gain LTS status in 10 days so it might be wiser to wait. I don't think it's a good idea to stick with 5/5.3/5.3.0.

I'm happy to help with this PR when node 6 becomes the current LTS, after checking that it actually works (and again, nothing will be enforced so there is no rush to upgrade).

NB: On the topic of enforcing which node version people use, we actually do a lightweight version of this at Springload. We have a `npm run lint:version` script that gets executed during the most important npm scripts (making a build, making a shrinkwrap). The main reason is that we tend to forget to run `nvm use`, and our projects go from node 0.12 from 6.0. Food for thought.

https://github.com/springload/frontend-starter-kit/blob/master/package.json#L14

.nvmrc
```diff
@@ -0,0 +1 @@
+5.3.0
```
thibaudcolas: I don't think 5.3.0 is the right version right now. Both the CI and `vagrant-wagtail-develop` use node 4, so I'd rather have an `.nvmrc` that says "4" as well. Using multiple major node/npm versions across the infrastructure creates complexities when dealing with shrinkwraps for example.

Once this is in place, it would make sense to migrate everything to node 6/7 (and npm 3/4), but I'd rather not tie having the nvmrc to that.

docs/contributing/developing.rst
```diff
@@ -13,7 +13,23 @@ If you'd prefer to set up all the components manually, read on. These instructio
 Setting up the Wagtail codebase
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Install Node.js, v5.3.0 or higher. Instructions for installing Node.js can be found on the `Node.js download page <https://nodejs.org/download/>`_. You will also need to install the **libjpeg** and **zlib** libraries, if you haven't done so already - see Pillow's `platform-specific installation instructions <http://pillow.readthedocs.org/en/latest/installation.html#external-libraries>`_.
+This project requires Node.js version 5.3.0. To manage the correct version, Node Version Manager (NVM) can be used.
```
thibaudcolas: Specifying just the major version should be enough. nvm not NVM.

docs/contributing/developing.rst
```diff
@@ -13,7 +13,23 @@ If you'd prefer to set up all the components manually, read on. These instructio
...
+
+`Go to the NVM Github page for installation instructions <https://github.com/creationix/nvm>`_.
```
thibaudcolas: nvm, GitHub.

docs/contributing/developing.rst
```diff
@@ -13,7 +13,23 @@ If you'd prefer to set up all the components manually, read on. These instructio
...
+
+To download, compile, and install the latest v5.3.0 release of node, do this:
+
+.. code-block:: sh
+
+    nvm install
```
thibaudcolas: The convention is for lines in command-line snippets to start with "$" in the Wagtail docs.

docs/contributing/developing.rst
```diff
@@ -13,7 +13,23 @@ If you'd prefer to set up all the components manually, read on. These instructio
...
+
+And then in any new shell just use the installed version:
+
+.. code-block:: sh
+
+    nvm use
```
thibaudcolas: This is not necessary – `nvm install` switches to the installed version, in the current shell. But it needs to be added in the "compiling static assets" section though.

package.json
```diff
@@ -9,6 +9,9 @@
       "browserify-shim"
     ]
   },
+  "engines": {
+    "node": "5.3.0"
```
thibaudcolas: Please only pin to the major version (`4.x.x`), not down to patch level.

jjanssen:
Closing this since I had difficulty updating this PR due its original changes in the forks `master` branch. New PR available at #3382

PR #2733, opened 2016-06-16
Fix #2236: Raise ValidationError if StreamField is required yet empty

### LINKED ISSUE #2236: StreamBlock(required=True) doesn't work
Setting `required=True` on a streamblock seems to be ignored, without error. Also, if your `StreamField` has `blank=False` (the default) it will not give an error if there is no content.

ltakens:


gasman:
Hi - it looks like there are a couple of test failures here - please can you fix?

* The test at https://github.com/torchbox/wagtail/blob/861bd2de14ab46c0d93287964b4b6e5d8ec8b9c1/wagtail/wagtailcore/tests/test_tests.py#L48-L53 is now failing validation when it tries to create a page with a streamfield - not sure why, as the streamfield data clearly isn't empty there...
* `iteritems` is not valid on Python 3 (you should use `items` instead, which works on both Python 2 and 3)

ltakens:
Thanks for the comments. In response:
1) For this, we created PR https://github.com/torchbox/wagtail/pull/275, which should be reviewed first, before the current PR can be merged.
Somehow, the created page wasn't getting the streamfield data. Now it is.
2) Changed iteritems to items

wagtail/wagtailcore/fields.py
```diff
@@ -109,3 +112,18 @@ def check(self, **kwargs):
...
+        """
+        Override clean method to run custom check for empty streamfields. An empty streamvalue is
+        is not considered empty by django forms.
+        """
+        cleaned_data = super(StreamField, self).clean(value, model_instance)
+        if not self.blank and not value:
+            raise ValidationError(
+                u'EmptyRequiredStreamField',
+                params={
+                    self.name: ErrorList([_('Required field is left blank')])
```
kaedroho: Should be "This field is required."

gasman:
Afraid I've found a couple of issues that hint that this might be a deeper problem than we realised...

Logically, the responsibility for validating "is this stream empty or not" belongs to StreamBlock / BaseStreamBlock ([source](https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailcore/blocks/stream_block.py)), rather than StreamField itself - the StreamField class is just a thin wrapper to allow the Block API (which does all the heavy lifting inside StreamField, including validating individual blocks) to communicate with the Django forms framework. That way, a StreamBlock that's more deeply nested in the structure (see http://docs.wagtail.io/en/v1.5.2/topics/streamfield.html#streamblock) can also be marked as required, not just the top-level one. Unfortunately this will need quite a bit of extra plumbing - StreamBlock needs to be updated to accept a `required` field; the `BaseStreamBlock.clean` method needs to include this new validation (which could be tricky, as it currently assumes that any `ValidationError`s originate from one of its child blocks rather than the StreamBlock itself); and `StreamField` will need to pass on its `blank=False` argument to its inner StreamBlock as `required=True`.

Also, I'm not keen on wagtail/wagtailadmin/views/pages.py having special-case handling for StreamField errors - the view layer should not have to know what kind of fields exist inside the form. If error messages are not getting through using the normal forms API, then that needs to be fixed at the form level. (Perhaps the NON_FIELD_ERRORS fix in #2713 will help here?)

gasman:
Completed in #3315

PR #2734, opened 2016-06-16
Fixing webpack config for production mode.

vincentaudebert:
We don't want inline source in prod. Optimising final bundle size a lot.
For dev mode, cheap-module-eval is faster than inline-source.


gasman:
Merged in 921a510d9fff9e07ec99f1db396a426cf522b282

PR #2735, opened 2016-06-17
Implement support for using pytest

mvantellingen:
This adds a conftest.py module to setup the test infrastructure
correctly. It doesn't yet add pytest and pytest-django to the testing
requirements so for now it is completly optional.

This change also updates the test_migrations testcase to use a regular
TestCase baseclass instead of the TransactionTestCase. Otherwise the
following tests fail with db errors.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,51 @@
+from __future__ import absolute_import, unicode_literals
+
+import os
+import shutil
+import warnings
+
+import django
+
+
+def pytest_addoption(parser):
+    parser.addoption('--deprecation', choices=['all', 'pending', 'imminent', 'none'], default='pending')
+    parser.addoption('--postgres', action='store_true')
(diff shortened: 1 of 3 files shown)


mvantellingen:
Pytest with ``py.test -n 8`` (pip install pytest-xdist)
```
real	1m1.003s
user	7m33.844s
sys	0m8.229s
```

Current ./runtests:
```
real	1m49.558s
user	1m45.433s
sys	0m2.488s
```

gasman:
Unfortunately it appears that `TransactionTestCase` is required for TestForMigrations - otherwise it always succeeds and never reports on missing migrations. For example, edit one of the `help_text` attributes in `testapp/templates/models.py` - this should cause TestForMigrations to fail (as there is no corresponding migration), but this doesn't happen with a plain `TestCase`.

PR #2731, opened 2016-06-16
Disable cls._deferred check for >=Django 1.10

m1kola:
See #2723. ~~Incorporates #2724~~

Looks like we don't need to check `cls._deferred` anymore (after Django 1.10). More details [here](https://code.djangoproject.com/ticket/26207) and in related PR.

PR #2739, opened 2016-06-17
fix for _deferred classes removed in django-1.10

pjstevns:
depends/includes #2736


gasman:
I've now merged @m1kola's copy of this fix from #2731, so (hoping I've followed the dependencies correctly here...) I'm closing this as a duplicate, and #2741 / #2742 should now be rebased on current master (skipping commits 157d7f5 and 6b9eb6e).

PR #2743, opened 2016-06-17
(API) Allow the fields query parameter to be used on detail views

kaedroho:
This pull request implements the ``?fields=<fields>`` query parameter on detail views (was previously only available on listing views).

Before #2484, the fields parameter wasn't very useful for the detail view. But now it supports customising nested fields which is quite useful for detail views as well.

Unlike the listing view, the fields will default to all fields on the detail view (which means this change is backwards compatible)


kaedroho:
Merged into #2484

PR #2745, opened 2016-06-17
don't enforce HTML as template laguage

kakulukia:
sets
HTML_EXTENSION_RE = re.compile(r"(.*)\.[a-z]{3,4}$")

and added option to disable child wrappers

### DIFF
--- a/wagtail/wagtailcore/blocks/stream_block.py
+++ b/wagtail/wagtailcore/blocks/stream_block.py
@@ -36,7 +36,7 @@ class BaseStreamBlock(Block):
     class Meta:
         default = []
 
-    def __init__(self, local_blocks=None, **kwargs):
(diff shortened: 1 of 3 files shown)


kaedroho:
Could we swap out the regex stuff altogether and use [os.path.splitext](https://docs.python.org/2/library/os.path.html#os.path.splitext) instead?

gasman:
Thanks @kakulukia!

@kaedroho +1. I was just about to suggest the same thing :-)

kakulukia:
@kaedroho even better :)

kakulukia:
i updated the PR

gasman:
Merged (with test added) in 334bebc55c0e3e4151358b8591e9a4779b0f3ac5 - thanks!

PR #2747, opened 2016-06-17
replace removed private API in favor of public API

pjstevns:
get_all_related_objects() was removed from django.db.models.options.Options (Model._meta)


gasman:
Have added the all-important test assertion (see https://github.com/torchbox/wagtail/issues/2723#issuecomment-227241446) in https://github.com/gasman/wagtail/commit/810952cfc88f4bac7408cc04c1875a82c6a67598 - running through CI now.

gasman:
Merged in ef05d37686310b99a754ff24b9a095bba87e43a1 - thanks!

PR #2748, opened 2016-06-17
(API) Allow true/false to be used where boolean is expected

kaedroho:
Currently, users have to pass 0 or 1 when filtering using boolean fields or ``has_children``. This PR allows ``true``/``false`` to be used as well.

I've also improved the validation of ints, fixing a crash when a string is passed when an int is expected.

### DIFF
--- a/wagtail/api/v2/tests/tests.py
+++ b/wagtail/api/v2/tests/tests.py
@@ -2,7 +2,7 @@ from __future__ import absolute_import, unicode_literals
 
 from unittest import TestCase
 
-from ..utils import FieldsParameterParseError, parse_fields_parameter
+from ..utils import FieldsParameterParseError, parse_fields_parameter, parse_boolean
 
 
 class TestParseFieldsParameter(TestCase):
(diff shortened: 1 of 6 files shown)


kaedroho:
Rebased onto #2484

m1kola:
* Rebased on top of `master` (#2484 was merged)
* 341f4c9fa5a9826b911521417ff446e18cab8b3b fixes [import order](http://46.101.88.142/torchbox/wagtail/49)

wagtail/api/v2/tests/tests.py
```diff
@@ -235,3 +235,43 @@ def test_invalid_star_and_underscore(self):
...
+    def test_valid_0(self):
+        parsed = parse_boolean('0')
+
+        self.assertEqual(parsed, False)
+
+
+    # BAD STUFF
+
+    def test_invalid(self):
+        print("WOO")
```
m1kola: m?

wagtail/api/v2/utils.py
```diff
@@ -216,3 +216,20 @@ def parse_fields(fields_str, expect_close_bracket=False):
...
+def parse_boolean(value):
+    """
+    Parses strings into booleans using the following mapping (case-sensitive):
+
+    'true'   => True
+    'false'  => False
+    '1'      => True
+    '0'      => False
+    """
+    if value in ['true', '1']:
```
m1kola: Do we want to be so strict? Maybe we can do `value.lower()` here?

m1kola:
Except comments above, looks good to me.

Thanks @kaedroho!

wagtail/api/v2/utils.py
kaedroho: I think it's fine as this would be used by developers. It'll be much easier to make this case-insensitive later on than it would be the other way anyway.

m1kola:
Merged in 95563dabb7bdd6648bbe701d28a0848bcfc3297f

Thanks @kaedroho!

PR #2755, opened 2016-06-19
Split remove_filter_model migration to prevent Postgres errors

### LINKED ISSUE #2754: Migration from #2537 fails on Postgres if renditions exist
If rendition records exist on Postgres at the point where the `wagtailimages.0014_remove_filter_model` migration (added in #2537) is applied, the migration fails with the error:

    django.db.utils.OperationalError: cannot ALTER TABLE "wagtailimages_rendition" because it has pending trigger events

To replicate on a clean install (I'm using python 2.7.6, postgresql 9.4.5):

	mkvirtualenv wagtailimgmigrate
	mkdir wagtailimgmigrate
	cd wagtailimgmigrate
	git clone https://github.com/torchbox/wagtail.git
	cd wagtail
	git checkout e995bc0e19d3b367e3d80a252dbd6bb0076031e6
	pip install -e .
	npm install
	npm run build

	cd ..
	wagtail start imgmigrate
	cd imgmigrate/

	edit imgmigrate/settings/base.py, set DATABASES to

	DATABASES = {
	    'default': {
	        'ENGINE': 'django.db.backends.postgresql_psycopg2',
	        'NAME': 'imgmigrate',
	    }
	}

	pip install psycopg2
	createdb imgmigrate
	./manage.py migrate

	./manage.py createsuperuser
	./manage.py runserver
	upload an image, visit /admin/images/ (to ensure a rendition is generated)
	cd ../wagtail
	git checkout 4d7bdfd1600870a4c86900f283506a193146f006
	cd ../imgmigrate
	./manage.py migrate

Most Google results for "cannot ALTER TABLE" / "pending trigger events" suggest that it occurs when converting a field from nullable to NOT NULL, which isn't the case here; but http://stackoverflow.com/questions/12838111/south-cannot-alter-table-because-it-has-pending-trigger-events suggests that splitting it into data and schema migrations may fix this.

gasman:
Fixes #2754. The sequence of operations is unchanged - just splitting the data migration step into its own migration. As I understand it, this means we avoid combining schema and data changes in the same transaction.


gasman:
@Gagaro / @timheap, could you sanity-check this please?

Gagaro:
Looks good to me. I just don't like having a migration into 3 distinct transactions, but I guess we don't have a choice (and this migration is already awful anyway :yum: ).

Gagaro:
Django trac ticket is there btw: https://code.djangoproject.com/ticket/25105

gasman:
Merged in 87e9c63dbf9c1ce3450cfcf3836d7ffe7b715b90

PR #2297, opened 2016-03-02
[WIP] Feature: React Explorer

JoshBarr:
Hi all,

I'm creating a PR for this component now so we can have visibility on the discussion. This is the new Explorer component based on React + Redux (and moment.js, and a few other nice things...). Demo video below:

screen shot 2016-03-02 at 2 28 09 pm

[VIDEO](https://youtu.be/g8OXYfVeGXc)

@kaedroho: there's some changes to the admin API here that you might want to review. Happy to move those changes to the admin API branch if it's easier. 

## Remaining tasks
- [ ] Create a better interface for the `filters` (currently an A/B toggle in the header for testing purposes)
- [ ] Write tests
- [ ] Write documentation
- [ ] Figure out how to deal with translations. Maybe bootstrapping the required text fields from i18n in the template?
- [ ] Test on some production-scale websites to see how it holds up.
- [ ]  Allow for 1 to n explorers to be instantiated on the page (needs a higher order redux reducer).

PR #2763, opened 2016-06-20
RequestCacheMiddleware: add a local memory cache instance as request.cache.

coredumperror:
This is useful for those times when a single request will need to perform a very expensive operation multiples times, but the results of that operation are specific to the current request. Unlike all other caching mechanisms provided by Django, RequestCache's data will be garbage collected alongside the request object.

This PR is part of the functionality from #2463, which @gasman asked me to split into multiple PRs to make it easier to review.

### DIFF
--- a/docs/advanced_topics/settings.rst
+++ b/docs/advanced_topics/settings.rst
@@ -39,6 +39,7 @@ Middleware (``settings.py``)
     'django.middleware.security.SecurityMiddleware',
 
     'wagtail.wagtailcore.middleware.SiteMiddleware',
+    'wagtail.wagtailcore.middleware.RequestCacheMiddleware',
 
     'wagtail.wagtailredirects.middleware.RedirectMiddleware',
   ]
@@ -441,6 +442,7 @@ These two files should reside in your project directory (``myproject/myproject/`
       'django.middleware.security.SecurityMiddleware',
 
       'wagtail.wagtailcore.middleware.SiteMiddleware',
+      'wagtail.wagtailcore.middleware.RequestCacheMiddleware',
       'wagtail.wagtailredirects.middleware.RedirectMiddleware',
   ]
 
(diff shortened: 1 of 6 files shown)


gasman:
To be revisited once the refactoring of #2463 (currently under way in #2869) is far enough down the line that we can see what our requirements for caching are. (As per https://github.com/torchbox/wagtail/pull/2463#issuecomment-226042250, I'd really like to avoid introducing a new middleware.)

coredumperror:
Is there a non-middleware alternative to a pre-request cache? I'm not sure that Django supports anything else. My original solution was a longer-lived cache, but it over-complicated the entire system pretty badly, which was why it was suggested that a pre-request cache be used instead.

gasman:
@coredumperror At this point I'm reserving judgement about whether we need caching at all... if I'm not mistaken, the left menu nav is the only query that needs to fetch more than a single "slice" of the tree at once (and there's a rewrite of that on the horizon... #2297), so my suspicion is that all the recurring queries are simple enough (or can be made simple enough) that we don't gain much by holding on to the results.

If we do need to cache results, then informally stashing them in a template variable (as we do here: https://github.com/torchbox/wagtail/blob/fb4a3621dcaeb0d3ff8080bbc087ed3afa93fd31/wagtail/wagtailadmin/templatetags/wagtailadmin_tags.py#L112) might be sufficient.

coredumperror:
As I recall from my testing, the function that builds the list of permitted paths frequently gets called 3-4 times in a typical request to the Explorer UI. That's why it felt like caching the results made sense.

As for using a context variable as a poorman's cache, that would work. The request object already gets passed down to every function that calls the cached function, since it's what owns the cache itself.

coredumperror:
Heck, now that I think about it, all the work I did to squash memory leaks in `RequestCache` (compared to the original version of the code I found on StackOverflow) means we don't need to call any cleanup functions on egress. So there isn't any reason to use a middleware at all. We could  just attach a `RequestCache` instance to the request during the first run of `_get_permitted_page_paths()`.

Should I make these changes to my PR over at #2463? The existence of #2869 makes me wonder if I should.

gasman:
Closing this now, as I think we're far enough down the line with implementing limitations on tree browsing (and related work such as the React explorer) to be confident that a per-request cache isn't required.

PR #2764, opened 2016-06-20
Fixes for a few typos and formatting issues in the Explorer's table headers.

coredumperror:
The bugfix here is the removal of the redundant </th> tags at the top. I noticed these while writing the Page explorability PR, #2463.

The formatting issue was the use of double quotes for python string comparisons. That messed up the template syntax highlighting, since double quotes were already being used around the HTML attribute values.


gasman:
Thanks!

PR #2766, opened 2016-06-21
Implementation of separate_filters_from_query

m1kola:


kaedroho:
We're currently looking into implementing filters within the query string (eg, "author:karl created:21/06/2016 hello world"). This PR is part of the implementation of that.

I'm working on an RFC for this now.

wagtail/wagtailsearch/tests/test_queries.py
```diff
@@ -137,3 +137,41 @@ def test_pagination(self):
...
+        self.assertDictEqual(filters, {'author': 'foo bar'})
+        self.assertEquals(query, '')
+
+    def test_filter_and_query(self):
+        filters, query = separate_filters_from_query('author:foo hello world')
+
+        self.assertDictEqual(filters, {'author': 'foo'})
+        self.assertEquals(query, 'hello world')
+
+    def test_filter_with_quotation_mark_and_query(self):
```
kaedroho: Could you add a test for what happens when a quotation mark is unclosed?

m1kola: Done. See [`test_filter_with_unclosed_quotation_mark_and_query `](https://github.com/torchbox/wagtail/pull/2766/files#diff-e48442489e8fdef4c5b79f94cb955c5bR173).

m1kola:
I think we should leave this PR open for now, because it doesn't make much sense to merge this utility function without using it anywhere.

PR #2774, opened 2016-06-22
Allow referencing models by string representation in SnippetChooserBlock

nimasmi:
Consider the following… models.py:

```
from .blocks import FooBlock
…

@register_snippet
class FooSnippet(models.Model):
    text_field = models.CharField(max_length=30)


class StandardPage(Page):
    body = StreamField(FooBlock())
```

and blocks.py

```
class FooBlock(blocks.StreamBlock):
    rich_text = RichTextBlock()
    foo_snippet = SnippetChooserBlock(FooSnippet)
```

This would result in circular imports. This PR makes it possible to do `foo_snippet = SnippetChooserBlock('myapp.FooSnippet')` instead.

No test case, as I couldn't see anywhere else that SnippetChooserBlock was being tested.

### DIFF
--- a/wagtail/wagtailsnippets/blocks.py
+++ b/wagtail/wagtailsnippets/blocks.py
@@ -3,12 +3,17 @@ from __future__ import absolute_import, unicode_literals
 from django.utils.functional import cached_property
 
 from wagtail.wagtailcore.blocks import ChooserBlock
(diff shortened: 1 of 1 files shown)


gasman:
Merged (with all-new testing loveliness) in df9a5e31fa1c6c64c962244161812270a2813f82. Thanks!

PR #2782, opened 2016-06-26
Remember input data in page chooser when switching between Internal/External/Email link

### LINKED ISSUE #2692: Remember data in inputs in page chooser when switching between Internal / External / Email link
Inspired by #2640 and #2677.

It seems to me that would be great to remember a data in inputs in page chooser when switching between Internal / External / Email link.

screen shot 2016-06-08 at 16 59 29

spmuck:
Fixes #2692 

Hi, this is my first contribution to wagtail, so I'd appreciate any feedback. I feel _link_types.html might be getting messy with the extra if blocks, but that looked like the cleanest solution to me. 
## My Solution

I am serializing the visible inputs on the modal and passing them as the urlParams to the loadUrl function of modal-workflow.js. 

I replace email_address with link_url for the email link, and url with link_url for the external link, so these values are preserved when the user goes between tabs. This is how the initial values of the forms were set up when the edit link function was added, so I wanted to be consistent with that.

On the _link_types.html page where the links for the tabs are built, I am keeping the current link parameters if you are on the internal link tab, and wiped them out when you are on either the external or email link tabs, as they will be rebuilt when you go to click another tab.

Thanks!

### DIFF
--- a/wagtail/wagtailadmin/templates/wagtailadmin/chooser/_link_types.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/chooser/_link_types.html
@@ -7,20 +7,28 @@
             {% if parent_page_id %}
                 <a href="{% url 'wagtailadmin_choose_page_child' parent_page_id %}{% querystring p=None parent_page_id=None %}">{% trans "Internal link" %}</a>
             {% else %}
-                <a href="{% url 'wagtailadmin_choose_page' %}{% querystring p=None parent_page_id=None %}">{% trans "Internal link" %}</a>
+                <a href="{% url 'wagtailadmin_choose_page' %}{% querystring p=None parent_page_id=None link_url=None link_text=None %}">{% trans "Internal link" %}</a>
             {% endif %}
         {% endif %}
 
         {% if current == 'external' %}
             | <b>{% trans "External link" %}</b>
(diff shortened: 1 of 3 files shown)


gasman:
Thanks @spmuck, and sorry for the delay in reviewing!

Unfortunately I think this level of coupling between form fields, URL parameters and Javascript is going to cause problems for future maintainability of the code - for example, see #3039 where we need to keep track of whether the user has changed the link text from its initial value, in order to decide whether to keep existing link elements intact. That fix would be incompatible with this change (unless we introduce another pair of URL params to store the initial values, as opposed to the most recent ones).

I realise that by tracking parent_id on the 'internal link' view, we've set a bit of a precedent for the link chooser to behave as a single multi-tabbed view, rather than multiple views that just happen to have links between them - and on that basis, we _ought_ to preserve the form content. Unfortunately, I think the only manageable way to do that is to rewrite the modal so that it actually _is_ a single view - faking it with URL params is just going to tie us in knots.

spmuck:
Yep, I agree, didn't feel like a good solution to me either.

thibaudcolas:
Closing. This would be easier to fix after #3038 – React has better primitives for state management that would make this rather easy to build.

PR #2783, opened 2016-06-27
Make non field error messages be shown in place of the generic validation error message

SableWalnut:
Make non field error message be shown in the messages section where it currently would say 'The page could not be saved due to validation errors' if they are available. This is incredibly useful in the case of custom validation where you may want to raise a whole page/form error.

### DIFF
--- a/wagtail/wagtailadmin/views/pages.py
+++ b/wagtail/wagtailadmin/views/pages.py
@@ -251,8 +251,13 @@ def create(request, content_type_app_name, content_type_model_name, parent_page_
                     # Ensure the 'next' url is passed through again if present
                     target_url += '?next=%s' % urlquote(next_url)
                 return redirect(target_url)
+
         else:
-            messages.error(request, _("The page could not be created due to validation errors"))
+            if form.non_field_errors:
+                pass
+            else:
+                messages.error(request, _("The page could not be created due to validation errors"))
+
             edit_handler = edit_handler_class(instance=page, form=form)
     else:
         signals.init_new_page.send(sender=create, page=page, parent=parent_page)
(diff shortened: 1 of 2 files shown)


dprothero:
@SableWalnut Perhaps we should leave base.html untouched. Then, rather than `pass` if you find non_field_errors, just add them to the messages collection. I'll add inline notes.

Also, this will need to be done for snippets.

wagtail/wagtailadmin/views/pages.py
```diff
@@ -251,8 +251,13 @@ def create(request, content_type_app_name, content_type_model_name, parent_page_
                     # Ensure the 'next' url is passed through again if present
                     target_url += '?next=%s' % urlquote(next_url)
                 return redirect(target_url)
+
         else:
-            messages.error(request, _("The page could not be created due to validation errors"))
+            if form.non_field_errors:
+                pass
```
dprothero: rather than `pass`, what about:
```
for error in form.non_field_errors:
    messages.error(request, error)
```

Then, you wouldn't need any changes to base.html

wagtail/wagtailadmin/views/pages.py
```diff
@@ -434,6 +439,10 @@ def edit(request, page_id):
         else:
             if page.locked:
                 messages.error(request, _("The page could not be saved as it is locked"))
+
+            elif form.non_field_errors:
+                pass
```
dprothero: Same thing here. Also need to find relevant place for this same functionality for snippets.

gasman:
Closing as per https://github.com/wagtail/wagtail/issues/536#issuecomment-285685185 - will open a new PR to address this shortly.

Issue #2469, opened 2016-04-10
Provide page in the context of StreamField blocks


timorieber:
I came across the limitation that the page is missing in the context of related StreamField blocks when rendering the blocks template several times. A typical use case is when you need to implement an intro block, where the page title and author should be displayed.

So far I was not able to solve this as a complete StreamField block. My implementation is a StreamField block anyway, but not using the blocks rendering mechanism. Instead I check for this block type and call a template with `{% include %}` and the necessary data as parameters.

The hurdle occurs in the majority of cases when you need to display data from builtin fields of Wagtails `Page` model. Otherwise you would be able to design it like it should be.

This issue is opened in addition to #1743 to clearly separate the discussion between the different requirements to have `request` and `page` available in the template context. I fully agree with https://github.com/torchbox/wagtail/issues/1743#issuecomment-144029475, though.

BertrandBordage:
I agree, and the parent block (if any) would also be extremely useful in a lot of cases.

gsstver:
I solved this case in one project. It's very useful.

gasman:
See updates on https://github.com/torchbox/wagtail/issues/1743#issuecomment-224543332 for a suggested way forward here, which should (hopefully...) be intuitive and consistent with Django's `{% include %}` tag. In the proposed solution, the entire template context is (by default) passed to the block template, so this will cater for both `page` and `request`.

mgd020:
Workaround: https://gist.github.com/mgd020/d841f8016024707b11b490b82ec648f0

m1kola:
This issue and issue #1743 were partially solved in #2786. But the feature still needs some work (see #2823 and #2824).

gasman:
Closing this now, as the specific issue described here is fixed by #2786 (even if there is further work to do on the feature as a whole).

PR #2786, opened 2016-06-27
include_block template tag for improved StreamField template inclusion

### LINKED ISSUE #1743: pagurl tag doesn't work within StreamField

gasman:
As proposed in https://github.com/torchbox/wagtail/issues/1743#issuecomment-224543332 - this PR implements an `{% include_block my_block %}` template tag, which works equivalently to Django's `{% include 'my_template.html' %}` tag (including constructs like `{% include_block my_block with classname='foo' only %}`).

This improves on the current convention of rendering StreamField blocks using variable tags e.g. `{{ my_block }}`, as it allows the child template to inherit the template context - this is necessary for the `{% pageurl %}` tag to work (and also provides access to the `page` variable, which is frequently requested - e.g. to allow building StreamField-based forms that post back to the current page).

As part of this change, the `render` and `render_basic` methods on `Block` now accept a `context` kwarg, which introduces a backwards incompatibility (since third-party libraries may have overridden these methods with the old method signature) - I've put compatibility shims in place to mitigate this.

Fixes #1743, #2469.

### DIFF
--- a/wagtail/wagtailcore/blocks/base.py
+++ b/wagtail/wagtailcore/blocks/base.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, unicode_literals
 
 import collections
+import warnings
 from importlib import import_module
 
 from django import forms
@@ -11,6 +12,8 @@ from django.utils import six
 from django.utils.encoding import force_text, python_2_unicode_compatible
 from django.utils.safestring import mark_safe
 from django.utils.text import capfirst
+
+from wagtail.utils.deprecation import RemovedInWagtail18Warning
 
 # unicode_literals ensures that any render / __str__ methods returning HTML via calls to mark_safe / format_html
 # return a SafeText, not SafeBytes; necessary so that it doesn't get re-encoded when the template engine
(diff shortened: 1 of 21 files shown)


m1kola:
At first glance looks good, but I want come back to it with fresh eyes tomorrow.

Thanks Matthew!

wagtail/wagtailcore/templatetags/wagtailcore_tags.py
```diff
@@ -46,3 +48,63 @@ def richtext(value):
...
+                for var_name, var_value in self.extra_context.items():
+                    new_context[var_name] = var_value.resolve(context)
+
+            return value.render_as_block(context=new_context)
+        else:
+            return force_text(value)
+
+
+@register.tag
+def include_block(parser, token):
```
m1kola: It seems to me that we need to add a similar tag for the Jinja2 template engine.

m1kola:
If we need an access to a request (to build a pagination object for a block, for example), we need to override the `render` method instead of creating the pagination object in the `get_context` method. Is that right?

m1kola:
Other than questions above, looks good to me.

Thank you @gasman!

wagtail/wagtailcore/blocks/base.py
```diff
@@ -204,24 +207,100 @@ def get_prep_value(self, value):
...
+        have overridden `render` with the type signature:
+            def render(self, value)
+        and will therefore fail when passed a `context` kwarg.
+
+        In Wagtail 1.8, when support for context-less `render` methods is dropped,
+        this method will be deleted (and calls to it replaced with a direct call to `render`).
+        """
+        try:
+            return self.render(value, context=context)
+        except TypeError:
```
m1kola: Hm. If 3-rd party block raises `TypeError` (I don't know why they may do that :)), `render` method will be called twice even if block have proper signature. It will be hard to debug for 3-rd party developers.

```python
    def render(self, value, context=None):
        # do something that raises a TypeError
```

Probably we need to use `inspect`. Something like this:
```python
import inspect


class Foo:
    def render(self, value):
        print("I'm alive")
    def _render_with_context(self, value, context=None):
        if 'context' in inspect.getargspec(self.render).args:
            return self.render(value, context=context)
        else:
            result = self.render(value)
            # Warn and bla-bla-bla


test = Foo()
test._render_with_context(1, {"two": 2})
```

gasman: I don't think it's worth making our code more complicated (and less Pythonic...) just to handle this very obscure special case.

This would only be a problem if the `render` method has side effects - otherwise, it doesn't matter that we call it twice. A `render` method that throws `TypeError` and has side effects would have to be doing something very hacky and non-standard indeed - if they have a hard time debugging it as a result, I think that's fair punishment ;-)

m1kola: > just to handle this very obscure special case.

It may be just typo or "4-rd party code" can rise`TypeError`... For example, I've tried to [add a pagination](https://github.com/torchbox/wagtail/pull/2786#issuecomment-230416360) to a block. It's a bad solution because I pass request object into my block through templates, but it's a good example when it will be hard to figure out what's going on.

```python
    def render(self, value, context=None):
        request = context.get('request') if context else None
        if not request:
            raise Exception('Sad trombone sound')

        pages = context.get('pages')

        per_page = settings.DEFAULT_PER_PAGE
        paginator = Paginator(pages, per_page)
        page_number = request.GET.get('page')

        try:
            pages = paginator.page(page_number)
        except PageNotAnInteger:
            # Hint: TypeError will be raised somewhere in Django's Paginator
            pages = paginator.page(1)
        except EmptyPage:
            pages = paginator.page(paginator.num_pages)

        context = dict(context)
        context['pages'] = pages

        return super().render(value, context)
```

gasman: Ah, yes, I see. So, the problem isn't really about `render` being unexpectedly called twice - it's that it's missing the context on the second occasion (and that's the one you'll see in the debug output). That's a fair point - I'll fix that with your suggested `inspect` approach. Thanks!

gasman:
> If we need an access to a request (to build a pagination object for a block, for example), we need to override the `render` method instead of creating the pagination object in the `get_context` method. Is that right?

Hmm, good point! I can see how it would be useful to have access to the parent context inside `get_context`. I think that's better handled in a separate PR though, as it's quite a bit more complicated than updating `render` / `render_basic` - we can't just change the signature to `get_context(self, value, context=context)` because we also need to consider what happens when old code calls `super(MyBlock, self).get_context(value)`. (Writing the documentation to explain what it means to pass a context to `get_context` is going to be quite difficult too!)

wagtail/wagtailcore/blocks/base.py
m1kola: Yes, I haven't expected to see `Exception: Sad trombone sound` and it took some time to figure out why I've got this exception instead of `TypeError` that was thrown by the `paginator`.

`inspect` - it's just the first idea that came to my head. If you know more Pythonic way, feel free to use it ;)

m1kola:
> I think that's better handled in a separate PR though

Absolutely agree.

Also you can implement a [Jinja2 template tag](https://github.com/torchbox/wagtail/pull/2786#discussion_r69522627) in a separate PR, if you wish. But I think that we should release Jinja2 template tag in the same time as Django's one.

gasman:
Rebased, and updated to check the method signatures of `render` and `render_basic` using `inspect.getargspec`, as suggested.

gasman:
I agree that we ought to support Jinja2 templates here too - I was meaning to mention that in the ticket description, but forgot...

I've just been reading the docs for [Jinja2's include / import tags](http://jinja.pocoo.org/docs/dev/templates/#include), and my first impression is that they differ significantly from Django's approach. Given that `{% include_block %}` is designed to closely follow the syntax of `{% include %}`, I feel that it would probably be better implemented by someone who's more familiar with Jinja2 idioms, and has a clear idea of what a Jinja2 version of `{% include_block %}` should look like. (@timheap, would that person be you? :-) )

m1kola:
@gasman I'm not sure that I more familiar with Jinja2 idioms, but I've used it for a long time.

`include` tags of both template engines are not so different. The only significant difference, in my opinion, is how Jinja2 allows you to control which variables will be passed into the context. You can specify `with context` (by default) or `without context` in `include` tag. If you want to pass only specified variables, you have to use `with` tag. It will be something like this:

```jinja
{% with my_var=something %}
    {% include 'my_template.html' %}
{% endwith %}
```

It seems to me that implementation of a Jinja2 version of our `include_block` will be simpler than implementation of a Django version because `with` tag does almost all business. As far as I remember, `with` goes out of the box with Jinja2, but it disabled by default (btw, I don't think that we need to care if it disabled).

I will back to this PR again in the morning.

P.S.: Damn! We need another solution for Python 3.5 (see tests).

gasman:
Fix for Python 3.5 added. I'm afraid it's not pretty... I look forward to Wagtail 1.8, when we can remove this pile of hacks :-)

timheap:
Adding an `{% include_block %}` tag should be possible. I think it would only need to support the `with/without context`, and would not have to support naming templates or extra kwargs:

```jinja
{% include_block block %}{# Includes context by default #}
{% include_block block with context %}{# Explicitly include context #}
{% include_block block without context %}{# Does not use context #}
```

Template authors can use `{% with %}` or `{% set %} if they need to provide extra context.

m1kola:
Merged in dbc4c9b28ecb355270fb6f2bc84db92ab2eb2bcb (release notes in b79069b2470974cd673f4d250b1bf3ed7d34c52c).

I've created 2 issues:
* #2823 for a Jinja2 version of `{% include_block %}`
* #2824 for parent context inside `get_context`

Thanks @gasman and @timheap!

PR #2793, opened 2016-06-28
Fix dirtyFormCheck alerts in Safari

JoshBarr:
In testing, Safari returns `body` for `event.explicitOriginalTarget || document.activeElement`. Therefore, we need to store a reference to the last clicked `$ignoredButtons` to hide the confirm dialog from being displayed.

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
@@ -55,6 +55,10 @@ function initTagField(id, autocompleteUrl) {
 
 var dirtyFormCheckIsActive = true;
 
+// Some browsers (webkit) don't handle explicitOriginalTarget, so emulate it. 
+window.lastClickedIgnoredTrigger = false;
+
+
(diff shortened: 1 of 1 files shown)


gasman:
After tearing my hair out trying to test this, I think `beforeunload` handling on Safari (9.1) is too buggy for us to support. Based on a trivial test page:

	<!DOCTYPE HTML>
	<html>
		<head>
			<script>
				window.addEventListener('beforeunload', function(event) {
					event.returnValue = "leave this page?";
					return "leave this page?";
				});
			</script>
		</head>
		<body>
			<a href="http://torchbox.com">leave</a>
		</body>
	</html>

once the prompt appears and is dismissed once, it never appears again, even after a force refresh, unless you quit the browser or re-enter the URL manually. I'm now working on a new solution based on #2801 that doesn't rely on checking button clicks at all, and hopefully avoids browser-specific special cases.

gasman:
Closing in favour of #2811 - testing the buttons/links that initiated the beforeunload event seems to be inherently unreliable across browsers (and fails to consider keyboard events), and we should use an approach that doesn't depend on it.

Issue #613, opened 2014-09-09
New tags that are added to images/documents are not being sent to Elasticsearch


kaedroho:
See: https://groups.google.com/forum/#!topic/wagtail/npEmbzCMguU

When a user edits an image or document, the post_save signal is being called before tags are committed to the database so `Tagsearchable.get_tags` is returning the tags that were in use before the edit.

PR #2804, opened 2016-06-30
Fix tests that are checking issue #613

kaedroho:
These tests haven't been run for a while due to a mistake in ``tox.ini``. They are currently broken on master.

They broke because they require ``AUTO_UPDATE`` to be ``True`` for the Elasticsearch backend, but we recently disabled that to improve speed and reliability of the entire test suite. This commit adds a way for the tests that need ``AUTO_UPDATE`` to force it to be enabled on specific backends


gasman:
Merged in c38b2ace3324e044261e05d3198003a892353470 - thanks!

Issue #2802, opened 2016-06-30
Adding of version number in the admin makes account / logout unclickable


jjanssen:
After the implementation of showing the version number in the admin (#2769) the account and logout link are no longer clickable when the settings menu is closed (at least in Chrome).  

![2016-06-30_09-54-43](https://cloud.githubusercontent.com/assets/287422/16480972/3537ea2e-3ea9-11e6-991e-ff7f79469c62.gif)

tomdyson:
@chrxr could you look at this, please?

jjanssen:
While testing my enhancements to the account settings. I also noted it shows the version number on top of the account settings in IE9.

![image](https://cloud.githubusercontent.com/assets/287422/16489832/7b9e293c-3ed7-11e6-88f0-e416e2bbb3c0.png)

gasman:
Fixed in #2813.

Issue #1323, opened 2015-05-20
Account settings button 


tmsndrs:
The way of navigating to account settings isn't particularly clear especially if you have a gravatar set up and there is no sense of what you can expect when you do click.

This came up during a client training session a while ago and it just caught me out as well when looking for notification settings.

The issue the client had was that the proximity to 'log out' makes it look like it's related to that action.

tomdyson:
Thanks, @tmsndrs - can you mock up a better alternative?

tmsndrs:
Hi @tomdyson, 

As a fairly quick change you could have something along these lines. I'm not a massive fan of having things at the bottom of the viewport like this but didn't want to disrupt any of the other UI elements. 

![wagtail-login2](https://cloud.githubusercontent.com/assets/7290588/7728412/fb23307e-ff06-11e4-809c-9a9bfd7fc37f.jpg)

It's not a millions miles away from patterns Twitter and Spotify use. It does introduce an extra click from the current UI but I think it adds clarity that will reduces cognitive load overall.

tomdyson:
Thanks, @tmsndrs, looks good to me.

SalahAdDin:
Soooo good: thank you.

SalahAdDin:
What's about this?

gasman:
Fixed in #2805

PR #2805, opened 2016-06-30
Account settings button

jjanssen:
These are some adjustments in favour of #1323 and its proposal to make it more clear for the end-user.
Its output is tested in IE9, IE10, IE11, Edge, Chrome, Firefox ESR (v45), iOS9.

A few functionalities to recall:
- Account settings only shows the first name and never the full name (Tom is short, Janneke is medium, Hendrik-Jan is large).
- If a name is longer than intended, it will be adding …
- If a translation of ‘Account Settings’ or ‘Login’ is longer than intended it will be adding…

This is mainly we have a fixed height footer in the navigation in order to maintain a predictable scrolling area for the upper items of this list is longer than the screen.

**What it looks like**
If no gravatar is set:
a-no-gravatar

If a gravatar is set:
a-gravatar￼

On mobile:
a-mobile

**Open issue**
The adding of the Wagtail version number in the settings menu enables the following bug (#2802)
This applies to the current master branch, but also to this feature.


tomdyson:
Thanks, @jjanssen, looks good! Are we dropping the circular mask on the gravatar?

jjanssen:
Thanks @tomdyson! I wouldn't say ditching the circular mask on the gravatar. For the sidemenu the square size makes more sense to be eye-catching. Making it round while attached to the side looks a bit silly. For displaying within the content-area I would still go with the circular mask.
But I am happy to change it if we decide it's a breach of style :-)

tomdyson:
Hi @jjanssen, could you please add a larger screenshot showing the menu in context? Ideally with mask / no-mask options :)

jjanssen:
Rough sketch ;-)

**Without the mask**
wagtail-admin-square

**With mask**
wagtail-admin-mask

tomdyson:
@jjanssen thanks! I _think_ I still prefer the masked version, but will defer to experts. @tmsndrs / @JoshBarr / @jonnyscholes / @ollywillans?

tmsndrs:
I think the first (square) option is probably going to be easier for people to use. Circles work well when people have a photo with the right composition but the square will be more forgiving. 

Can I be really picky and suggest a little more margin above 'account settings' when the menu is open? It feels a bit tight.

jjanssen:
Sure @tmsndrs 

I'll wait for some feedback from the others, so I can pick it in one batch.

SalahAdDin:
I like it, i like it!

SalahAdDin:
Applied it?

jjanssen:
Nope, was cleaning up branches a little too enthousiastic

gasman:
Hi @jjanssen - just a couple of issues I've spotted:

- if the user has no email (e.g. they were created using `./manage.py createsuperuser` and no email was provided at that point) and the avatar element is missing, the alignment is broken:
screen shot 2016-09-30 at 16 29 24
- on the settings -> users -> edit page, clicking anywhere in the 'Account' tab of the form causes the popup to toggle up and down due to the re-use of the 'account' ID.

Other than those, looks good!

jjanssen:
Hi @gasman, thanks for the feedback! I have processed the issues and updated the PR.
Also made a little change in the base colouring of the element, based on the PR for #2985.

gasman:
Looks good to me! Merged in 2992d90e97a9eb776d45c09b4bd50499ced555eb.

PR #2807, opened 2016-06-30
Make validation errors on formsets trigger the alwaysDirty flag

gasman:
When the page edit form is redisplayed on validation errors, we set alwaysDirty on the
dirty form handler logic so that navigating away triggers the unsaved changes warning
even if no further changes are made. However, the old code tested `form.errors`, which
fails to catch validation errors originating from formsets.


gasman:
Merged via #2811.

PR #2810, opened 2016-07-01
Unable to use TransactionTestCase without wagtailsearchpromotions app on PostgreSQL

ychab:
Hi,

I made a PR just to easily let's you reproduce the bug because I cannot fix it yet. In that way, you just have to execute the test (don't forgot to have a postgreSQL database test ready, I temporary force the test settings to use it instead of using environment variables). Then with : 
````
./runtests.py wagtail.wagtailsearch.tests.test_migrations
````
PostgreSQL should throw the following exception for the transaction test case : 

> psycopg2.NotSupportedError: cannot truncate a table referenced in a foreign key constraint

To summarize, if we use postgreSQL, don't have wagtailsearchpromotions app enabled and want to do some TransactionTestCase in our project, we can't.

Obviously, the migration 0003 remove the EditorsPick's table model which is still use by wagtailsearchpromotions ?

I can't still figure out why TestCase works fine and TransactionTestCase don't...?

Thanks

### DIFF
--- a/wagtail/tests/settings.py
+++ b/wagtail/tests/settings.py
@@ -9,7 +9,7 @@ MEDIA_URL = '/media/'
 
 DATABASES = {
     'default': {
-        'ENGINE': os.environ.get('DATABASE_ENGINE', 'django.db.backends.sqlite3'),
+        'ENGINE': 'django.db.backends.postgresql_psycopg2',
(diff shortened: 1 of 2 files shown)


gasman:
This sounds like a duplicate of (or at least related to) #1824.

ychab:
Exactly! I close this issue because it's a duplicate.

PR #2801, opened 2016-06-30
Prevent alert after pressing enter while saving page. 

spmuck:
Fixes #2798

## Solution

Check if form has been submitted before showing navigation alert. This will prevent the alert from showing when pressing enter in an input box.

PR #2811, opened 2016-07-01
Simplify and fix dirty forms behaviour

### LINKED ISSUE #2513: Clicking on link in document chooser popup triggers "you have unsaved changes" message

### LINKED ISSUE #2798: Alert while saving page by pressing enter key
Steps to reproduce:

1. Open edit form of any page
2. Make changes in any form field
3. Press enter key

Prompt will be shown. You need to click "Leave" to save your changes.
screen shot 2016-06-29 at 19 28 13

gasman:
Incorporates #2801 (fixes #2798), #2807 and #2808 (partially fixes #2513).

I noticed that @spmuck's approach in #2801 (listening to form submission actions rather than click events) allowed us to eliminate much of the old code, such as `ignoredButtonsSelector`. This also means that we're no longer relying on "which button/link triggered this event?" logic, which seems to be poorly standardised across browsers.

I had to make sure that I wasn't introducing any regressions by removing this code, so it made sense to fix (almost) all of the outstanding dirty forms issues, and put together the following checklist covering the expected behaviour and all past and current failing edge cases:

* Browsing away (e.g. clicking 'Images') without making any changes SHOULD NOT trigger the message
* Making a change then browsing away SHOULD trigger the message
* Making a change then clicking 'Save' SHOULD NOT trigger the message
* Making a change then pressing Cmd+S (with no input field focused) SHOULD NOT trigger the message
* Making a change then pressing enter SHOULD NOT trigger the message
* Making a change then clicking 'Preview' SHOULD NOT trigger the message; subsequently browsing away SHOULD trigger the message
* Making a change then pressing Cmd+P (with no input field focused) SHOULD NOT trigger the message
	(NB: may require allowing popups)
* Making a change that fails server-side validation (e.g. blanking a required field in a formset, or entering an expiry date before the go-live date), clicking 'Save', then browsing away SHOULD trigger the message
* Making a change that fails server-side validation then clicking 'Preview' SHOULD NOT trigger the message; subsequently browsing away SHOULD trigger the message
* Making a change that fails client-side validation (e.g. entering an invalid email into an EmailField), clicking 'Save', then browsing away SHOULD trigger the message
* Making a change that fails client-side validation then clicking 'Preview' SHOULD NOT trigger the message; subsequently browsing away SHOULD trigger the message
* Making a change, then opening the document chooser popup and clicking a download link SHOULD NOT trigger the message

These now succeed on Chrome 51, Firefox 47, IE 11 and Safari 9.1 with the following exceptions:

* Safari's onbeforeunload support somewhat broken as per https://github.com/torchbox/wagtail/pull/2793#issuecomment-229948331, and had to be tested by manually adding a cache-busting URL parameter to the URL each time. On the plus side, this bug only causes false negatives (not showing the warning when it should be shown), which are not nearly as bad as false positives...
* Safari does not support client-side validation on input fields such as `<input type="email">`, so these are validated server-side (i.e. those tests just duplicate the two tests above, and still pass)
* IE and Safari fail on document download links (#2513), due to not supporting the `download` attribute.

Unfortunately, removing the `ignoredButtonsSelector` closes off one possible avenue of fixing #2513 on IE and Safari - see #2535. However, I believe this is a fairly obscure bug, affecting minority browsers (I believe Edge is unaffected, but can't currently test this), and I think that leaving these unfixed is a better alternative than relying on the inherently unreliable button-detection code.

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -152,6 +152,7 @@ Contributors
 * Raphael Stolt
 * Tim Graham
 * Thibaud Colas
+* Sean Muck
 
 Translators
 ===========
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
@@ -63,6 +63,11 @@ function enableDirtyFormCheck(formSelector, options) {
     var confirmationMessage = options.confirmationMessage || ' ';
     var alwaysDirty = options.alwaysDirty || false;
     var initialData = $form.serialize();
+    var formSubmitted = false;
+
+    $($form).submit(function() {
+        formSubmitted = true;
+    });
 
     window.addEventListener('beforeunload', function(event) {
         // Ignore if the user clicked on an ignored element
@@ -75,7 +80,10 @@ function enableDirtyFormCheck(formSelector, options) {
             }
         });
 
-        if (dirtyFormCheckIsActive && !triggeredByIgnoredButton && (alwaysDirty || $form.serialize() != initialData)) {
+        var displayConfirmation = dirtyFormCheckIsActive && !formSubmitted && !triggeredByIgnoredButton
+            && (alwaysDirty || $form.serialize() != initialData);
+
+        if (displayConfirmation) {
             event.returnValue = confirmationMessage;
             return confirmationMessage;
         }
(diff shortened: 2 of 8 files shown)


kaedroho:
Merged in ad1f6716c583bedb0bce902b5e2259ef87755b4d Thanks!

PR #2812, opened 2016-07-01
Add support of callable choices for ChoiceBlock

### LINKED ISSUE #2809: Callable choices for ChoiceBlock
Would be nice to have support of callable `choices` for `ChoiceBlock`.

Django's [`ChoiceField`](https://docs.djangoproject.com/en/1.8/ref/forms/fields/#choicefield) supports callable values for `choices` argument since version 1.8.

m1kola:
Fixes #2809

### DIFF
--- a/wagtail/wagtailcore/blocks/field_block.py
+++ b/wagtail/wagtailcore/blocks/field_block.py
@@ -4,6 +4,7 @@ import datetime
 
 from django import forms
 from django.db.models.fields import BLANK_CHOICE_DASH
+from django.forms.fields import CallableChoiceIterator
 from django.template.loader import render_to_string
 from django.utils import six
 from django.utils.dateparse import parse_date, parse_datetime, parse_time
@@ -318,9 +319,13 @@ class ChoiceBlock(FieldBlock):
 
     def __init__(self, choices=None, required=True, help_text=None, **kwargs):
         if choices is None:
-            # no choices specified, so pick up the choice list defined at the class level
-            choices = list(self.choices)
+            # no choices specified, so pick up the choice defined at the class level
+            choices = self.choices
+        elif callable(choices):
+            # Support of callable choices
+            choices = CallableChoiceIterator(choices)
         else:
+            # Cast as a list, even if we got CallableChoiceIterator object
             choices = list(choices)
 
         # keep a copy of all kwargs (including our normalised choices list) for deconstruct()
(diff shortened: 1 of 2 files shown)


wagtail/wagtailcore/blocks/field_block.py
```diff
@@ -326,36 +331,58 @@ def __init__(self, choices=None, required=True, help_text=None, **kwargs):
...
         users to define subclasses of ChoiceBlock in their models.py, with specific choice lists
         passed in, without references to those classes ending up frozen into migrations.
         """
-        return ('wagtail.wagtailcore.blocks.ChoiceBlock', [], self._constructor_kwargs)
+
+        # Actually, we can just return CallableChoiceIterator,
+        # but it will be casted as a list in __init__ method
+        _constructor_kwargs = self._constructor_kwargs.copy()
+        if isinstance(_constructor_kwargs['choices'], CallableChoiceIterator):
+            _constructor_kwargs['choices'] = _constructor_kwargs['choices'].choices_func
```
m1kola: Probably we don't need this, if we are not going to support callable choices using subclassing... Needs extra attention while code review, anyway.

m1kola:
Rebased on top of the `master` branch.

nkuttler:
@m1kola I have tested the code and oddly enough when I enter the admin edit page with one of those blocks the callable is called multiple times, and the choices repeat themselves. I haven't figured out why yet, but couldn't have a serious look yet. The choices are repeated four times atm. Other than that this looks great.

m1kola:
@nkuttler thanks for the review!

Could you, please, provide an example code to reproduce a repeating issue?

> (Admin CSS doesn't load for me in the master branch though, but that's unrelated)

It's because repo doesn't contain compiled static files. You need to compile them as described here: http://docs.wagtail.io/en/v1.6.1/contributing/css_guidelines.html#compiling

nkuttler:
I'm not able to reproduce the issue with either a clean test project or my existing project with a fresh virtualenv.

nkuttler:
I see this was postponed again, can I do anything to speed up the merge, or does it depend on core dev reviews?

gasman:
I notice that it doesn't like `choices` being defined as a callable at the class level:

    def get_adverts():
        return ((advert.id, advert.text) for advert in Advert.objects.all())

    class AdvertBlock(ChoiceBlock):
        choices = get_adverts

Rendering and makemigrations both trip up because they think `choices` is a method - not sure if that's fixable, so perhaps we should just add a note to the docs to say that using a callable at the class level isn't allowed.

wagtail/wagtailcore/blocks/field_block.py
```diff
@@ -318,9 +319,13 @@ class ChoiceBlock(FieldBlock):
...
         if choices is None:
-            # no choices specified, so pick up the choice list defined at the class level
-            choices = list(self.choices)
+            # no choices specified, so pick up the choice defined at the class level
+            choices = self.choices
+        elif callable(choices):
+            # Support of callable choices
+            choices = CallableChoiceIterator(choices)
         else:
+            # Cast as a list, even if we got CallableChoiceIterator object
```
gasman: Is this comment correct? CallableChoiceIterator is handled by the case above.

gasman:
Hi @m1kola - I found it quite hard to follow the sequence of conversions between callables and iterators, so I created https://github.com/gasman/wagtail/commit/6dc1998d6e2b7755e105eaf8c816a32ce09c94df which adds some extra comments and does some light refactoring. Please can you give this a quick look and make sure I haven't broken anything?

m1kola:
Hi Matthew! Thanks for your review and improvements!

I'll take a look at your changes till the end of this week, but at first glance - looks good!

m1kola:
Hi Matthew! Thanks for your patch. Looks clearer.

gasman:
Merged in 55bdae573b71bbda63e8a8213b2c33e3288dbdcd - thanks!

PR #2813, opened 2016-07-01
Fix version number regression

jjanssen:
As described in issue #2802, the version number made the account / logout items unclickable.
Turns out the wagtail-version was inheriting styles from that component, but position fixed is conflicting with the submenu due to css transform in combination with z-index of the menu (500) vs menu footer (2).
Targeting the submenu's footer and setting it to absolute positioning fixes this.

Re-tested in: IE9, IE10, IE11, Edge, Chrome (latest), Firefox ESR (v45), iOS9, Safari 9.1.1.


gasman:
Merged in 9b6cdf41f4aa1fc48a08781ef7ed60887f5133e0 - thanks!

PR #2814, opened 2016-07-03
add some docs for emailblock, integerblock, decimalblock and floatblock

OktayAltay:


gasman:
Thanks @OktayAltay!

We need to be really careful about the order in which we introduce concepts on this page, since it needs to work as a tutorial for people learning about StreamField for the first time, as well as being a reference document.

The `BlogPage` example is meant to be a simple, real-world example of how a Wagtail developer can benefit from StreamField. If we try to introduce too many block types into that initial example, we risk making it more confusing, and less representative of real-world situations. It's easy to understand a blog post as being a series of headings, paragraphs and images - if we add things like 'email' and 'integer' to that list, people will rightly start wondering "why would I want to insert an integer in a blog post?"

In practice, EmailBlock etc are probably only going to be used within a StructBlock, so we can't really give a meaningful example of them until after we've introduced StructBlock. So perhaps the solution is to work these block types into the StructBlock documentation, using an example like:

    class CityBlock(blocks.StructBlock):
        name = blocks.CharBlock()
        population = blocks.IntegerBlock()
        temperature = blocks.FloatBlock()

Alternatively, perhaps a dedicated 'Examples' / 'Recipes' section at the end of the page would work...

OktayAltay:
Hey Gasman, 

You are right, i will change my pull request and make a new section at the end of the page on how to implement blocks into a StructBlock.

gasman:
Have added your updated example in bc7739e104aa0dbacfd704608bf5332c292c4a31 - thanks!

Just so you're aware - your master branch is now out of sync with the main torchbox/wagtail repo, as you've merged the new changes from torchbox/wagtail on top of yours. You'll probably want to delete and recreate your copy of the repo, and ensure that any pull requests you create in future are done on branches rather than master.

Issue #2777, opened 2016-06-23
AbstractBaseUser without is_staff field generates error


FlipperPA:
Pretty extreme outlier, but I ran into it. If you have an AbstractBaseUser and don't implement is_staff in the model, you run into troubles, for example, when attempting to apply Admin Permissions to a user:

```
AttributeError at /cms/users/1/
can't set attribute
Request Method: POST
Request URL:    http://vagrant.wharton.upenn.edu:8000/cms/users/1/
Django Version: 1.9.7
Exception Type: AttributeError
Exception Value:    
can't set attribute
Exception Location: /home/vagrant/.virtualenvs/wrds-classroom/lib/python3.5/site-packages/wagtail/wagtailusers/forms.py in save, line 188
Python Executable:  /home/vagrant/.virtualenvs/wrds-classroom/bin/python
Python Version: 3.5.1
Python Path:    
['/vagrant/html/wrds-classroom',
 '/home/vagrant/.virtualenvs/wrds-classroom/lib/python35.zip',
 '/home/vagrant/.virtualenvs/wrds-classroom/lib/python3.5',
 '/home/vagrant/.virtualenvs/wrds-classroom/lib/python3.5/plat-linux',
 '/home/vagrant/.virtualenvs/wrds-classroom/lib/python3.5/lib-dynload',
 '/usr/local/lib/python3.5',
 '/usr/local/lib/python3.5/plat-linux',
 '/home/vagrant/.virtualenvs/wrds-classroom/lib/python3.5/site-packages',
 '/vagrant/html/wrds-classroom']
Server time:    Thu, 23 Jun 2016 18:40:43 -0400
Traceback Switch to copy-and-paste view

/home/vagrant/.virtualenvs/wrds-classroom/lib/python3.5/site-packages/django/core/handlers/base.py in get_response
                    response = self.process_exception_by_middleware(e, request) ...
/home/vagrant/.virtualenvs/wrds-classroom/lib/python3.5/site-packages/django/core/handlers/base.py in get_response
                    response = wrapped_callback(request, *callback_args, **callback_kwargs) ...
/home/vagrant/.virtualenvs/wrds-classroom/lib/python3.5/site-packages/django/views/decorators/cache.py in _cache_controlled
            response = viewfunc(request, *args, **kw) ...
/home/vagrant/.virtualenvs/wrds-classroom/lib/python3.5/site-packages/wagtail/wagtailadmin/decorators.py in decorated_view
[...]

PR #2816, opened 2016-07-04
Fix for custom users without the 'is_staff' field

SableWalnut:
Currently when using a custom user without the `is_staff` field, the error `can't set attribute` will be raised, as upon saving a user in wagtail admin, the save function will attempt to set `user.is_staff` to the value of `user.is_superuser`. For example my custom user had something like this.

```
    @property
    def is_staff(self):
        "Is the user a member of staff?"
        # Simplest possible answer: All admins are staff
        return self.is_admin
```

I'll add tests if there's no problems with what I've got so far.

### DIFF
--- a/wagtail/wagtailusers/forms.py
+++ b/wagtail/wagtailusers/forms.py
@@ -184,8 +184,9 @@ class UserEditForm(UsernameForm):
     def save(self, commit=True):
         user = super(UserEditForm, self).save(commit=False)
 
-        # users can access django-admin iff they are a superuser
-        user.is_staff = user.is_superuser
+        # users can access django-admin if they are a superuser
+        if not hasattr(user.__class__, 'is_staff') and 'is_staff' in vars(user):
+            user.is_staff = user.is_superuser
 
         if self.cleaned_data["password1"]:
             user.set_password(self.cleaned_data["password1"])


wagtail/wagtailusers/forms.py
```diff
@@ -184,8 +184,9 @@ def clean_password2(self):
     def save(self, commit=True):
         user = super(UserEditForm, self).save(commit=False)
 
-        # users can access django-admin iff they are a superuser
-        user.is_staff = user.is_superuser
+        # users can access django-admin if they are a superuser
+        if not hasattr(user.__class__, 'is_staff') and 'is_staff' in vars(user):
```
timheap: It would be better to check `if 'is_staff' in user._meta.fields_map:`

gasman:
Related: #2777, https://github.com/torchbox/wagtail/issues/970#issuecomment-185173676

At the moment, I'm leaning more towards changing the behaviour to leave the `is_staff` flag alone by default, possibly with a setting to change that to "set it iff the superuser flag is set" or "add an 'is staff' checkbox to the form". The current behaviour made more sense before we had the wagtailusers module, and admin users had to use django-admin to configure users / permissions / groups.

wagtail/wagtailusers/forms.py
gasman: Surely this is a classic case to apply the EAFP principle? Try to set the property, but catch the `AttributeError` (or whatever the exception may be)...

BMJHayward:
@gasman I agree with leaving `is_staff` alone for the following reasons:
  - I have users who need the django admin, or the wagtail admin or both.
  - Not all users are staff members.
  - Not all users are superusers.
  - There are only two superusers besides myself who use the django and wagtail admins

I've got the following work around, which lets people access wagtail admin if they are in the moderators or editors groups provided by wagtail, without having to be superusers, while still being staff members:

### WARNING: TERRIBLE TERRIBLE HACK ###

1. Add a row to the database table `django_content_type`
  - in column `app_label` add: 'wagtail_admin'
  - in column name `model` add: 'access_admin'
  - I use dbeaver for this, but you can use python, CLI, pgadmin, whatever

2. Do the following in the Django shell
```
from django.contrib.auth.models import Group, Permission
edtrs = Group.objects.get(name='Editors')
modrtrs = Group.objects.get(name='Moderators')
wag_CT = ContentType.objects.get(id=<id of content type you just created>)
permission = Permission.objects.create(codename='access_admin', name='Can access admin', content_type=wag_CT)
edtrs.permissions.add(permission)
modrtrs.permissions.add(permission)
```

3. In `wagtail/wagtailusers/forms.py`, edit this line:
    `user.is_staff = user.is_superuser`
to:
    `# user.is_staff = user.is_superuser`
  - You'll find it twice, in `UserCreationForm` and `UserEditForm`, (as of version 1.8)
  - Or delete the line entirely, but I'm not brave enough to do that.

4. For now, I have also set wagtail==1.8 in my requirements file, so the changes won't be overwritten in my virtualenv

gasman:
Closing in favour of #3434, which ignores the `is_staff` flag entirely - after discussion with the core team, we've decided that it makes sense for Wagtail to stay out of decisions about who gets access to the Django admin.

PR #2819, opened 2016-07-04
Inherit relevant border-radius from the parent

### LINKED ISSUE #2818: Minor issues with styles of buttons
1. In some choosers the edit buttons are higher than other buttons. Document chooser in stream field, for example.

    screen shot 2016-07-04 at 18 14 22

2. Ugly left borders in buttons with icons. **Small** version of this button have the same issue.

    screen shot 2016-07-04 at 18 43 54

jjanssen:
Fixes #2818 

Turns the `:before` did not inherit its parent border-radius. So the small grey is actually the square / rectangle element with its alpha channel on `0.2`. I've made it inherit its parent radius, because it can also be the case that the parent doesn't have a border-radius.

Normally `overflow: hidden` would fix this for children, but it doesn't seem to make a difference.

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_forms.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_forms.scss
@@ -309,6 +309,8 @@ input[type=checkbox]:checked:before {
             text-align: center;
             background-color: rgba(0, 0, 0, 0.2);
             display: block;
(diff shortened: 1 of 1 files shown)


jjanssen:
Interesting Chrome bug! Toggling `body { margin: 0 }` from normalize.css seems to make it work.

gasman:
Merged in 9f72e121592669deed1204a42051402739bd5f23 - thanks!

PR #2821, opened 2016-07-05
Fix styling of email fields in AbstractEmailForm example

gasman:
classname="full" is incorrect here (it generates excessive padding); also, using a FieldRowPanel
can make the display neater.


m1kola:
Merged. Thanks!

P.S: Just noticed that `FieldRowPanel` adds padding-left. I'll create a separate issue.
screen shot 2016-07-08 at 16 45 07

PR #2829, opened 2016-07-07
Eliminate Django 2.0 deprecation warnings for on_delete, is_anonymous, is_authenticated

gasman:
Because test runs are currently quite noisy, and there's no harm in getting this in early...

Added `on_delete` to model fields that are currently missing, and added compatibility shims for `user.is_anonymous` and `user.is_authenticated` (which should now be accessed as attributes rather than methods).


m1kola:
Thanks!

PR #2830, opened 2016-07-07
Add back buttons to page delete / unpublish confirmation pages

### LINKED ISSUE #2213: Lack of cancel button on 'unpublish' or 'delete' pages

gasman:
Fixes #2213

### DIFF
--- a/wagtail/wagtailadmin/templates/wagtailadmin/pages/confirm_delete.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/pages/confirm_delete.html
@@ -17,14 +17,18 @@
                 {% endblocktrans %}
             {% endif %}
         </p>
-        {% page_permissions page as page_perms %}
-        {% if page_perms.can_unpublish %}
-            <p>{% trans "Alternatively you can unpublish the page. This removes the page from public view and you can edit or publish it again later." %}</p>
-        {% endif %}
+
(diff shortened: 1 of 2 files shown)


m1kola:
Merged in 64aac99d9711cd5e5eeaf0f455d15c4d25222e60 (release notes in 07ac5aa8ee5456d308789585d0375a2fbe68ede4).

Thanks!

PR #2850, opened 2016-07-19
Fixed search tests

kaedroho:
A test was mutating the search_fields attribute of the SearchTest model causing subsequent tests that depended on the previous value to fail.


kaedroho:
Test failures are due to broken Django 1.10 support

PR #2573, opened 2016-05-06
Elasticsearch 2 support

kaedroho:
This pull request adds Elasticsearch 2 support into Wagtail into a new backend called ``wagtail.wagtailasearch.backends.elasticsearch2``

The main incompatibility is the fact that a column with the same name in different types in the same index cannot have different mappings. This means that Image.title and Page.title must have the exact same boost, analyzer etc.

I've used two methods to isolate models/fields in Elasticsearch to prevent these clashes:

 - I've split up models into different indices based on their "root model". So all pages, images and documents are in different indices (but all page models remain together in the same index)
 - Field names of page models are now prefixed with the model name in the index. So fields with the same name in different page models cannot have clashing mappings

This pull request is based on six others, see their descriptions for more information:

- [x] #2567 Validate search method's "fields" argument
- [x] #2569 Renamed search backend classes
- [x] #2570 Rewrote update_index command to support rebuilding multiple indices in the same 
- [x] #2571 Make deciding index column names the responsibility of the search backend
- [x] #2572 Fixed bug causing nested objects to use the wrong mapping
- ~~#2851 Replaced usage of index_analyzer~~

TODO:
 - [x] Tests for column with same name but different mapping in sibling models
 - [x] Tests for get_model_root
 - [x] Docs
 - [x] Travis CI setup (done, waiting for travis to run through)

Fixes #1734

---
This change is [Reviewable](https://reviewable.io/reviews/torchbox/wagtail/2573)

PR #2851, opened 2016-07-19
[Elasticsearch 2] Replaced usage of index_analyzer 

kaedroho:
Resubmission of #2568

index_analyzer has been removed in Elasticsearch 2.0. But the new way works in older versions as well so I think it's worth switching to the new way now.

https://www.elastic.co/guide/en/elasticsearch/reference/current/breaking_20_mapping_changes.html#_analyzer_mappings


kaedroho:
@gasman Travis doesn't report any Python 2.7 test failures, neither can I reproduce locally. I also cannot see by looking at the code how a Python 2.7 specific test failure could've occurred.

Could this be remerged?

gasman:
The failure is showing here:
https://travis-ci.org/torchbox/wagtail/jobs/145808149

It seems to be hopping between py27-dj18-sqlite-elasticsearch and py27-dj19-postgres-elasticsearch, so it may be intermittent - but it was happening consistently enough on my machine to run a `git bisect`.

gasman:
Still failing after a rebase (on py27-dj110-mysql-elasticsearch this time): https://travis-ci.org/torchbox/wagtail/builds/145843051 :-(

adriaant:
Having issues using ES > 2.0 that should be fixed in this PR. E.g. 
`raise HTTP_EXCEPTIONS.get(status_code, TransportError)(status_code, error_message, additional_info)
elasticsearch.exceptions.RequestError: TransportError(400, 'mapper_parsing_exception', 'Mapping definition for [_partials] has unsupported parameters:  [index_analyzer : edgengram_analyzer]')`
Is this change going to be released soon?

gasman:
@adriaant We can't offer any promises about when ES2 support will be shipped, sorry.

adriaant:
@gasman Well, as long as it does get released at some point ;)

kaedroho:
@gasman It seems to be a bug in Elasticsearch as it occasionally happens with 1.3 and 1.4. But never happens in 1.5+.

kaedroho:
Going to merge this into #2573

PR #2852, opened 2016-07-19
Allow test failures on Django 1.10

kaedroho:
An update to Django 1.10 has caused a test to fail. While I'm not usually one who believes in rolling back features due to small test failures, I think it's warranted in this case as Django 1.10 is not stable yet and we shouldn't guarentee full support until at least an RC has been released.


kaedroho:
I've just been made aware that an RC has just been released so now is the time to start supporting Django 1.10. Closing this PR, I'll look into the specific issue now.

PR #2854, opened 2016-07-19
Remove dubious GroupForm test that fails on Django 1.10rc1

gasman:
test_group_form_includes_non_registered_permissions_in_initial_data fails in
Django 1.10rc1 because the native value of a ManyToManyField (as reflected in
form.initial) is now a queryset rather than a list of PKs. However, this test
is not actually testing anything meaningful. It asserts that the initial
value of the 'permissions' field includes any permissions the group has which
are not included in registered_permissions; however, since these permissions
don't appear in the form, it's inconsequential whether they're included or not.


kaedroho:
Thanks!

PR #2858, opened 2016-07-20
Prevent double-escaping of block template output on Jinja2

### LINKED ISSUE #2541: Double-escaping bug when rendering blocks with Jinja2
You should be able to reproduce this bug with:

- this page model:

```python
from wagtail.wagtailcore import blocks, fields, models

class PageWithBlocks(models.Page):

    content = fields.StreamField([
        ('stuff', blocks.CharBlock(template='stuff.html')),
    ])
```

- this `page_with_blocks.html` template:

```
{{ page.content }}
```

- this `stuff.html` template:

```
<p>{{ value }}</p>
```

The `<p>` and `</p>` tags show up in the rendered web page.

----

Ultimately, this happens because:

- Django's template engine returns an HTML-safe string: `NodeList.render()` calls `mark_safe()` on its output.
- Jijna2 doesn't do that: `Template.render()` doesn't wrap its output in `Markup(...)`

It's unclear to me which behavior is correct :-/

At least this difference could explain why the problem wasn't noticed before.

----

This bug is the result of the following sequence:

- `page.content` is a `StreamValue` instance built by `StreamField.to_python()`
- Jinja2 calls its `StreamValue.__html__()` method to render it in the page template
- `StreamBlock.render()` which is inherited from `Block` -- we're rendering the StreamField
- `StreamBlock.render_basic()` which is inherited from `BaseStreamBlock`
- `format_html_join()`
- `force_text(child)` -- now we're rendering the child, a `StreamChild` which inherits `BoundBlock` and is a bound version of our `RichTextBlock`
- `BoundBlock.__str__()`
- `Block.render()`
- `render_to_string()` -- that's where the behavior diverges between Django and Jinja2

As a consequence, `format_html_join()` escapes the representation of the child when it concatenates blocks, and the bug happens.

----

Making `Block.render()` escape its output makes the example work as expected:

```diff
diff --git a/wagtail/wagtailcore/blocks/base.py b/wagtail/wagtailcore/blocks/base.py
index 89a1f54..795f7ba 100644
--- a/wagtail/wagtailcore/blocks/base.py
+++ b/wagtail/wagtailcore/blocks/base.py
@@ -217,7 +217,7 @@ class Block(six.with_metaclass(BaseBlock, object)):
         """
         template = getattr(self.meta, 'template', None)
         if template:
-            return render_to_string(template, self.get_context(value))
+            return mark_safe(render_to_string(template, self.get_context(value)))
         else:
             return self.render_basic(value)
 
```

The `mark_safe(render_to_string(...))` pattern is quite common in the code base so this solution seems consistent with the rest of the project.

Also it should fix the problem fully as the root cause is the difference in whether `render_to_string()` returns a HTML-safe string or not.

----

Note that I only tested on Python 3. I suspect that the problem also exists on Python 2.

gasman:
Fixes #2541. Thanks to @aaugustin for the initial bug report and detailed analysis.

(Have taken the opportunity to fix some rough edges in the jinja2 test setup too...)


gasman:
Merged in 6d6da5c1a97d3856b34100d0a6f84ece25bb16fc

PR #2859, opened 2016-07-20
Prevent double-escaping of rich text blocks on Jinja2

### LINKED ISSUE #2542: Double-escaping bug when rendering rich text blocks with Jinja2
This is similar but distinct from #2541. The combination of these two bugs currently produces triple-escaping :-)

----

You should be able to reproduce this bug with:

- this page model:

```python
from wagtail.wagtailcore import blocks, fields, models

class PageWithRichTextBlocks(models.Page):

    content = fields.StreamField([
        ('stuff', blocks.RichTextBlock(template='stuff.html')),
    ])
```

- this `page_with_blocks.html` template:

```
{{ page.content }}
```

- this `stuff.html` template (an actual example would contain HTML around the value):

```
{{ value }}
```

The HTML source of the rich text shows up in the rendered web page.

----

Taking a clue from `RichTextField`, I tried to apply the `|richtext` filter as follows:

```
{{ value|richtext }}
```

If I were rendering a `RichTextField`, the `|richtext` filter would call `expand_db_html()`, wrap the result in a div and call `mark_safe` on the result, thus returning a safe string.

With a `RichTextBlock`, `value` is already a `RichText` instance and the `|richtext` filter is a no-op. So `RichText.__str__` is called. It returns a safe string. This achieves the desired effect with the Django template language but not with Jinja2, for reasons I didn't investigate fully.

Anyway, the fix for this problem is the following:

```diff
diff --git a/wagtail/wagtailcore/rich_text.py b/wagtail/wagtailcore/rich_text.py
index af0b778..f66f640 100644
--- a/wagtail/wagtailcore/rich_text.py
+++ b/wagtail/wagtailcore/rich_text.py
@@ -4,7 +4,7 @@ import re  # parsing HTML with regexes LIKE A BOSS.
 
 from django.utils.encoding import python_2_unicode_compatible
 from django.utils.html import escape
-from django.utils.safestring import mark_safe
+from django.utils.safestring import SafeText
 
 from wagtail.wagtailcore import hooks
 from wagtail.wagtailcore.models import Page
@@ -180,7 +180,7 @@ def expand_db_html(html, for_editor=False):
 
 
 @python_2_unicode_compatible
-class RichText(object):
+class RichText(SafeText):
     """
     A custom object used to represent a renderable rich text value.
     Provides a 'source' property to access the original source code,
@@ -191,4 +191,4 @@ class RichText(object):
         self.source = (source or '')
 
     def __str__(self):
-        return mark_safe('<div class="rich-text">' + expand_db_html(self.source) + '</div>')
+        return '<div class="rich-text">' + expand_db_html(self.source) + '</div>'
```

(says the guy who made Django's and Jinja2's idea of what a safe string is almost compatible, but not fully, as this bug demonstrates...)

gasman:
Fixes #2542. Thanks to @aaugustin for the bug report and patch.

(incorporates #2858)


gasman:
Merged in 71cf921407a7ac17f62891e8547c19f16678cb90

PR #2860, opened 2016-07-21
Fix use of Page instead of specific_class

### LINKED ISSUE #2492: Use of base Page class in views prevent subclass methods from taking effect.
I was overriding the url method for my subclass, and have noticed that this takes effect in many places, but not everywhere.

In particular, https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailadmin/views/pages.py#L319

`page = Page.objects.get(pk=page.pk)`

Could be replaced with

`page = page.specific_class.objects.get(pk=page.pk)`

This feels counter-intuitive, and as though it goes against the norms of wagtail and django elsewhere.

I have noted #836, #1378, and #2022, but none of those address this exactly.

woegjiub:
Fixes #2492 

Uses specific_class.url instead of Page.url for getting the 'view_live' URL
in the success message following Page publication.

### DIFF
--- a/wagtail/wagtailadmin/tests/test_pages_views.py
+++ b/wagtail/wagtailadmin/tests/test_pages_views.py
@@ -19,6 +19,7 @@ from django.utils import formats, timezone
 from django.utils.dateparse import parse_date
 
 from wagtail.tests.testapp.models import (
+    EVENT_AUDIENCE_CHOICES,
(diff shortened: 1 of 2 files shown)


gasman:
Merged in bd226fcedf738f9349d234cc9ca662a33841230f - thanks!

PR #2868, opened 2016-07-26
Fix for incorrect tag variable name in modeladmin

adriaant:
modeladmin's result_row_value.html template refers to item_closing_tag, but that should be closing_tag.


gasman:
Looks good - thanks!

PR #2869, opened 2016-07-26
Limit explorer menu nav to the subtree the user has permission over

gasman:
Partially addresses #2401 (ref: https://github.com/torchbox/wagtail/issues/2401#issuecomment-205483027); adapted from #2463.

Updates the explorer-nav logic to take the user's permissions into account. The menu now begins at the closest common ancestor node of all pages they have add/edit/publish/lock permission for - as a result, users with permission over a specific deep section of the tree don't have to redundantly drill down to it, and we're a step closer to true 'multi-homed' installations where the user is not made aware of tree structure that exists outside of their own remit.


wagtail/wagtailadmin/navigation.py
```diff
@@ -0,0 +1,103 @@
...
+            group_permissions__group__in=user.groups.all(),
+            group_permissions__permission_type__in=['add', 'edit', 'publish', 'lock']
+        )
+
+    if not(pages_with_direct_permission):
+        return []
+
+    # Find the closest common ancestor of the pages the user has permission for;
+    # this (or its children) will be the root menu level for this user.
+    cca_path = pages_with_direct_permission[0].path
```
kaedroho: It would be great to have this as a method on PageQuerySet:

```python
cca = pages_with_direct_permission.closest_common_ancestor()
```

gasman: Probably, yep :-) It would involve an extra query though (because we don't use the resulting Page instance here, only the path).

kaedroho: Ahh yes. Maybe a ``common_path_prefix`` method would be useful to have as well. We definitely could use this in the choosers (eg, when asked to select an ``EventPage`` it makes sense to work out which section of the tree events are stored in, if there is one, and start the user there)

wagtail/wagtailadmin/navigation.py
```diff
@@ -0,0 +1,103 @@
...
+        menu_root_depth = int(len(cca_path) / Page.steplen) + 1
+
+    # Run the query to fetch all pages to be shown in the navigation. This consists of the following
+    # set of pages, for each page in pages_with_direct_permission:
+    #  * all ancestors (plus self) from menu_root_depth down
+    #  * all descendants that have children
+    #  * all descendants at menu_root_depth, regardless of whether they have children.
+    #    (this ensures that a freshly built site with no child pages won't result in an empty menu)
+
+    ancestor_paths = [
```
kaedroho: There is a queryset method you could use for this which might make this function look a little more clearer: https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailcore/query.py#L57-L67

wagtail/wagtailadmin/navigation.py
```diff
@@ -0,0 +1,103 @@
...
+        for page in pages_with_direct_permission
+        for path_len in range(menu_root_depth * Page.steplen, len(page.path) + Page.steplen, Page.steplen)
+    ]
+
+    criteria = Q(path__in=ancestor_paths)
+
+    for page in pages_with_direct_permission:
+        criteria = criteria | (
+            Q(path__startswith=page.path) & (
+                Q(depth=menu_root_depth) | Q(numchild__gt=0)
```
kaedroho: I can't figure out what the ``Q(depth=menu_root_depth)`` filter is for. Is this just to make sure the page-with-direct-permission always appears in the query even if it has no children?

kaedroho:
Apart from a little bit of logic that I found tricky to get my head around, this looks good and works well!

wagtail/wagtailadmin/navigation.py
gasman: Yep - it's to ensure that nobody ends up with an empty menu if the only thing they have permission on is a leaf page (which is common for a newly-set-up site or subsite). However, I *think* this special case might be made redundant by a last-minute change I made to the ancestor_paths logic (it's now inclusive, so that the page they have direct permission on is always shown even if it's an arbitrarily deep leaf node). Let me think through the cases a bit more...

gasman: OK - the `Q(depth=menu_root_depth)` clause can only ever take effect in the specific case where the user has permission on the root, and the menu starts one level down from that (i.e. the special case at https://github.com/torchbox/wagtail/pull/2869/files#diff-c1b04653cef1c28e847039a46736aca8R44)

It can therefore be simplified to `Q(depth=2)`, which sure enough is exactly what the current code does :-) https://github.com/torchbox/wagtail/pull/2869/files#diff-9ec0dccfdaa33a0fd310c76b5db2cb37L1350

This clause is important because it's the only thing that ensures that a single-page site at the top level shows in the menu - without it, a superuser on a freshly built Wagtail instance won't get a menu at all. (This is covered by the unit tests at https://github.com/torchbox/wagtail/pull/2869/files#diff-f8ac0fd58cb50cefaa757703fe34c15fR61 - page 10, `/home-2/`, is a single-page site.) I thought this needed to be generalised to deal with the case of a non-superuser who has permission for a single leaf node somewhere deeper in the tree - but actually it doesn't, because `ancestor_paths` is inclusive (and therefore the node in question is counted as an ancestor of itself).

gasman: Changed to simplified version in 676a94e853c316fe8b3da82449b3054ac7a5a460.

gasman:
Merged in 43269fd763248053a419526b960b4718a0f27765 + parents.

coredumperror:
I just got back from a 2-week vacation/illness, so I'm seeing this for the first time. I'm so glad to see that my work is finding its way into master!

Issue #2870, opened 2016-07-26
RoutablePageMixin fails if Page has an instance-bound descriptor


fabiommendes:
I have a Page class with a FileField

```python
class MyPage(RoutablePageMixin, Page):
    file = models.FileField(upload_to='my_dir') 
    ...
```
This class is broken because the get_subpage_urls() class method from RoutablePageMixin tries to access the "file" attribute which cannot be accessed from classes, but only from instances. I think this is a broken behavior from Django' s part, however the fix is trivial in wagtail: just change the getattr line to be like this:

```python
class RoutablePageMixin:
    ...
    @classmethod
    def get_subpage_urls(cls):
        routes = []
        for attr in dir(cls):
            val = getattr(cls, attr, None)
            if hasattr(val, '_routablepage_routes'):
                routes.extend(val._routablepage_routes)

        return tuple([
            route[0]
            for route in sorted(routes, key=lambda route: route[1])
        ])
```

I can make a pull request if you want, but the fix is so simple it doesn't worth the hassle.

gasman:
Hi @fabiommendes - thanks for the report! A pull request would be really helpful - this also needs a test case, to confirm that the problem is fixed, and to make sure it doesn't get accidentally re-introduced in any future code changes. (https://github.com/torchbox/wagtail/pull/2585/files is a good example to refer to)

fabiommendes:
Ok. Tests are done and I submitted the pull request.

PR #2872, opened 2016-07-26
fix issue #2870

fabiommendes:
Added bug fix and tests as required


gasman:
Tests are failing now (because the `RoutablePageTest` class isn't properly registered with migrations etc, I think) - please can you fix?

gasman:
Merged in a3b850f37e4564dd2b6e1eec88574b9a79aee6dc - thanks!

Issue #2557, opened 2016-05-03
Get Wagtail version


alexgleason:
Currently the way is with

```python
from wagtail.wagtailcore import __version__ as WAGTAIL_VERSION
```

But maybe a function like Django has is a more elegant way, as it doesn't depend on the version number existing in any particular file. I'm afraid someone could move that, and it isn't a documented feature.

Or am I over thinking it? I know the version is shown on hover for the logo in the admin screen, but I haven't checked to see how it's pulled out. Maybe documenting it and agreeing it should remain there would work, also.

kapito:
It should be quite easy to make a command or alias along the lines of 'wagtail getversion' which outputs the response from `from wagtail.wagtailcore import __version__ as WAGTAIL_VERSION`.

http://ubuntuforums.org/showthread.php?t=1554459 may be relevant here, however when trying to execute `from wagtail.wagtailcore import __version__ as WAGTAIL_VERSION` I get "from: can't read /var/mail/wagtail.wagtailcore". Do you know what I'm doing wrong?

alexgleason:
Hi @kapito, you're running the command I shared above inside of your terminal emulator. However, it's Python code and must be run in a Python environment.

Assuming you have [Python](https://www.python.org/downloads/) and [Wagtail](https://wagtail.io/developers/) installed on your local machine, you can run `python` inside of your terminal to enter the Python interpreter. You should see `>>>` before your cursor.

Then you can type:

```python
from wagtail.wagtailcore import __version__ as WAGTAIL_VERSION
```

This will import the version number under the name `WAGTAIL_VERSON`, then you can type:

```python
print(WAGTAIL_VERSON)
```

to print the version number. Use ctrl+D to exit the Python interpreter.

kapito:
Ah, my apologies! Thanks for clarifying. I agree that's quite a lengthy process. Unfortunately I'm not sure exactly how to implement it, but one option is to create a python file (or maybe it could go in setup.py?) which has the code you mentioned:

`from wagtail.wagtailcore import __version__ as WAGTAIL_VERSION`
`print("Your current Wagtail version is " + WAGTAIL_VERSON)`

[...]

PR #2877, opened 2016-07-29
Add VERSION tuple to wagtail.wagtailcore

timheap:
This allows people developing plugins for Wagtail to test the Wagtail version more easily, using code like:

```python
from wagtail.wagtailcore import VERSION
if VERSION >= (1, 5):
    # Do something
else:
    # Do the old thing
```

I also used the new version tuple in the docs to generate version numbers like `1.5` for all the `1.5.x` releases, which is how those version numbers are supposed to work. This means that `http://docs.wagtail.io/en/1.5/` will always point to the latest 1.5 release, regardless of the patch level of the `1.5` series. This will likely require some changes to the configuration on RTD though, so this change can come as a separate PR if you want to discuss it separately.

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -37,6 +37,6 @@ webpack
 
 ## How to release
 
-The front-end is bundled at the same time as the Wagtail project, via `setuptools`. You'll need to set the `__semver__` property to a npm-compliant version number in `wagtail.wagtailcore`.
+The front-end is bundled at the same time as the Wagtail project, via `setuptools`.
 
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -29,9 +29,6 @@ if not on_rtd:  # only import and set the theme if we're building docs locally
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 sys.path.insert(0, os.path.abspath('..'))
 
-# Get Wagtail version
-from wagtail.wagtailcore import __version__
-
 # Autodoc may need to import some models modules which require django settings
 # be configured
 os.environ['DJANGO_SETTINGS_MODULE'] = 'wagtail.tests.settings'
@@ -76,9 +73,12 @@ copyright = u'2015, Torchbox'
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
 # built documents.
-#
+
+# Get Wagtail version
+from wagtail.wagtailcore import __version__, VERSION
+
 # The short X.Y version.
-version = __version__
+version = '{}.{}'.format(VERSION[0], VERSION[1])
 # The full version, including alpha/beta/rc tags.
 release = __version__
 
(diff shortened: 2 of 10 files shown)


gasman:
Thanks @timheap! The readthedocs version handling is a long-standing bugbear of mine, so it'll be good to get that fixed...

Please can you add a comment to __init__.py describing the valid formats for `VERSION`? (For example, is the final number omitted for a `final` release?) It seems straightforward enough, but packaging up a release is the worst possible time for making slip-ups, so it's worth making it as idiot-proof as possible.

alexgleason:
Interesting, is a tuple the typical way to handle this? [See how I'm checking the version number in Wagtail Font Awesome](https://github.com/alexgleason/wagtailfontawesome/blob/d5865dbd4042021953a6129b0af41ad10c270ddf/wagtailfontawesome/wagtail_hooks.py#L16).

timheap:
@alexgleason `parse_version` is the more correct way of doing this, but is quite a lot more verbose. Django uses a version tuple much like this, so there is precedent.

@gasman I'll add a comment documenting the correct formats

mevinbabuc:
Any particular reason the version is not on the `wagtail` module but on the `wagtail.wagtailcore` module. Isn't `wagtail.__version__` more intuitive ?

I was hoping for something like this

```python
import wagtail
print(wagtail.VERSION)
print(wagtail.__version__)
```

timheap:
I've added some info about the format of `VERSION` in `__init__.py`, and rebased on top of master.

gasman:
I think it's a good idea to make the version number available under the main `wagtail` module, as per @mevinbabuc's suggestion. @alexgleason previously raised this in #2557, and my reluctance to change it then was mainly down to the fact that it would be breaking existing code for no benefit other than neatness.

There's a weird paradox here where _any_ improvement to the version handling logic - no matter how well-considered and backwards-compatible it may be - is partly self-defeating, because that improvement won't exist on some of the versions you're testing against. For example, if we make this change in 1.7, then any third-party code that cares about Wagtail <1.7 compatibility will have to do something like:

    try:
        from wagtail import VERSION
    except ImportError:
        from wagtail.wagtailcore import __version__
        from pkg_resources import parse_version
        VERSION = parse_version(__version__)

...which is basically the worst of both worlds. :-) But in fact, that's all the more reason to make the string-to-tuple and wagtailcore-to-wagtail changes at the same time - because if we make those changes separately, we'll need a three-way try/except block instead. And, of course, once Wagtail 1.6 is ancient history, the code will simply become `from wagtail import VERSION`.

timheap:
I've moved the version info to `wagtail` as suggested. I've imported the version info in to `wagtail.wagtailcore` as well, for backwards compatibility. I've not added a deprecation warning, as keeping those imports around for ever is not hard.

I hear where you're coming from @gasman, but the sooner this change is made, the less packages will depend on the old location. Any new packages can use the new location and version tuple, and any old packages can continue to use the old string version at the old location indefinitely - or at least till they drop support for Wagtail 1.6.

gasman:
Merged in 7c9dc029c7339fe180fdba53dc1911852d6872db + parents, thanks!

PR #2771, opened 2016-06-22
Added support for 'exclude_from_explorer' attribute on ModelAdmin class

ababic:
Recreated the branch/PR due to rookie github mistakes with the original. Related to #2757

PR #2878, opened 2016-07-29
Adds 'exclude_from_explorer' attribute to ModelAdmin class

ababic:
This PR replaces #2771, which suffered with conflicts in CHANGELOG.rst after rebasing on master (it wasn't properly rebased earlier on). The changes have also been squashed and the feature is now referenced the documentation.

Travis CI Integration on my fork has confirmed multi-environment tests are passing.


wagtail/contrib/modeladmin/options.py
```diff
@@ -503,6 +509,11 @@ def get_admin_urls_for_registration(self):
             )
         return urls
 
+    def modify_explorer_page_querset(self, parent_page, queryset, request):
```
gasman: typo (querset -> queryset)

ababic:
@gasman Ah, yes. Should be sorted now. It's a shame I made the same mistake twice, otherwise tests would've chocked :)

ababic:
@gasman. Apologies. Was clearing out some old branches from our fork and forgot about this. Have reopened.

gasman:
Merged in 23eced1fcda1a52c2ee05ead43e5135bc8ff9d60 - thanks!

I've refactored things a bit so that the hook only gets attached if `exclude_from_explorer` is true. It does make the code a bit more verbose, but I figure that this option is only going to be used relatively rarely, and I'd consider the `construct_explorer_page_queryset` hook to still be semi-experimental - so I'd prefer to only attach the hook when it's actually being used, rather than have every modeladmin-registered model define it but have it do nothing 99% of the time.

ababic:
Not problem @gasman, thanks for the merge!

PR #2879, opened 2016-07-29
Fixing template variables

adriaant:
Wrap a template variable that was not always present in a conditional.
Fix name of another template variable.


tomdyson:
Thanks, @adriaant! @ababic, could you review this please?

ababic:
I can confirm that the use of `view.verbose_name_plural` is correct here. `model_name_plural` is from a while ago, before I realised that coming up with new names for common django terms was probably a bad idea.

Issue #2315, opened 2016-03-07
React will load twice when using Wagtail components via NPM


JoshBarr:
This issue is summarising a discussion that @timheap, @justinvdm and I are having. React is loaded once in the new webpack entry points in #2297 [WIP] Explorer, then once again by an implementer's custom code that imports React components. E.g: 

``` javascript
// Required to mount React components ...
import ReactDOM from 'react-dom';
import { Explorer } from 'wagtail';

... 
// Now we have two instances of React! Side effects and warnings galore! 
ReactDOM.render(<Explorer />, document.querySelector('.explorer'));
```
### Possible solutions:
- Expose React and other WT dependencies as globals on `window` via [expose-loader](https://github.com/webpack/expose-loader), essentially skip using NPM bundled versions of common libs?
- Force the entire front-end build process to use the implementer's webpack config to build all the admin JS (not ideal!)
-  Have some way of collecting the implementer's code back into Wagtail's internal frontend build process, via a `manifest.json` or similar?

justinvdm:
> Force the entire front-end build process to use the implementer's webpack config to build all the admin JS (not ideal!)

That option might not be too bad if the implementer is provided with a base webpack config they can extend, something like:

```
module.exports = function(entries) {
  // entries === implementer's entry points

  return {
    // base webpack config properties
  };
}
```

The tricky thing is that at some stage, we need to involve the python side at some stage to figure out where wagtail lives in order to find wagtail's internal entry points. Maybe something like [this](gist.github.com/justinvdm/1203d98a4b6be977cea4) might work? Not ideal needing to talk to python over stdout, but seems more sensible than moving part of the build process over to the python side.

JoshBarr:
I was thinking about that too – however it means you have to build the JS inside the same virtual env that the python code is running in... otherwise it could find the wrong version of wagtail, or not find wagtail at all. Node file watchers are really slow inside vagrant boxes, too (but docker is fine, I think?). 

What if we move all the relevant client-side code to the npm package? Then we're not relying on searching site-packages at all... I remember we discussed this originally when creating the npm package, but we decided it would be too much of a departure from the current structure. Now there's a fairly compelling reason to do it. More and more I'm convinced that django's handling of frontend dependencies is prehistoric.

JoshBarr:
Alternatively, we could have the python helper script write a manifest.json into the project directory as a postinstall hook to `pip install wagtail`, which contains the abspaths to all the original js entry points.

[...]

PR #2882, opened 2016-07-30
Added hooks for front-end dependencies

### LINKED ISSUE #2338: Hooks for CSS and JS shouldn't allow full HTML insertion
`insert_editor_css`
`insert_editor_js`
`insert_global_admin_css`
`insert_global_admin_js`

Is there a better way for these hooks to work? Related: #2333 #2276

They require the user to insert the full HTML. This opens potential for abuse where people could use them to embed things other than styles or scripts. Implementations look something like this:

``` python
@hooks.register('insert_editor_css')
def editor_css():
  return format_html('<link rel="stylesheet" href="' \
  + settings.STATIC_URL \
  + 'demo/css/vendor/font-awesome/css/font-awesome.min.css">')
```

I think it would make more sense to pass in a _file_ (and maybe some options) rather than a full HTML tag.

That way people can't possibly misuse them, and we have more control over how the content can be rendered.

My first thought is to make it look something like this:

``` python
@hooks.register('insert_editor_js')
def editor_js():
  return {
    'files': [static('demo/js/myscript.js')],
    'options': {
        'async': True
    }
  }
```

StreamField blocks implement something similar. [Here is how a ListBlock gets its assets.](https://github.com/torchbox/wagtail/blob/824a7c50e272c3f1492551f015716a771f8cb639/wagtail/wagtailcore/blocks/list_block.py#L39) This is implementing [Django Form Assets](https://docs.djangoproject.com/en/1.9/topics/forms/media/), which could also be a great way to accomplish this. It allows for options to account for optional HTML attributes like CSS's `screen` (not sure about async for JS though).

Ideally, I'd like to see a hook implementing this. Maybe it could look like:

``` python
from django.forms import Media
from django.contrib.staticfiles.templatetags.staticfiles import static

@hooks.register('insert_editor_assets')
def editor_assets():
    return Media(
        js=[static('myapp/myscript.js')],
        css=[static('myapp/mystyle.css')]
    )
```

It appears to be the standard Django way, and it's already what we're using for StreamField.

alexgleason:
#### The problem

Multiple Wagtail plugins may require the same front-end dependency, ex. Font Awesome. But we don't want to require that dependency twice on the admin screen, so we need to detect that duplication and only include it only once.
#### The solution

We can use named dependencies to prevent duplication, as long as developers can agree to use the same name. For instance, if we name Font Awesome `fontawesome`, then multiple projects can enqueue the `fontawesome` CSS, and this code will be sure it's only imported once.
## Usage

You can use the following hooks:
- `enqueue_scripts`
- `enqueue_styles`

The system will automatically detect duplicates among enqueued media throughout apps and will only include the library once.
## Example

``` python
@hooks.register('enqueue_scripts')
def enqueue_jquery():
    return {
        'jquery': {
            'source': 'lib/js/jquery.js',
            'version' '3.1.0',
        }
    }

@hooks.register('enqueue_styles')
def enqueue_fontawesome():
    return {
        'fontawesome': {
            'source': 'lib/css/fontawesome.css',
            'version': '4.6.3',
        }
    }
```

See also: [wagtaildependencies](https://github.com/alexgleason/wagtaildependencies/), a Wagtail plugin that adds this functionality. But it would be better in core so we don't have to always install this.

Fixes #2338 and could be used to address torchbox/wagtail-markdown#8

### DIFF
--- a/wagtail/wagtailadmin/wagtail_hooks.py
+++ b/wagtail/wagtailadmin/wagtail_hooks.py
@@ -107,3 +107,33 @@ def page_listing_more_buttons(page, page_perms, is_parent=False):
                      attrs={'title': _('Unpublish this page')}, priority=40)
     yield Button(_('Revisions'), reverse('wagtailadmin_pages:revisions_index', args=[page.id]),
                  attrs={'title': _("View this page's revision history")}, priority=50)
+
+
+@hooks.register('insert_global_admin_js')
+def enqueue_plugin_scripts():
+    scripts = {}
+    for hook in hooks.get_hooks('enqueue_scripts'):
+        scripts.update(hook())
+
+    HTML = '\n'.join([
+        '<script src="{}"></script>'.format(
+            static(meta['source'])
+        ) for handle, meta in scripts.items()
+    ])
+
+    return HTML
+
+
+@hooks.register('insert_global_admin_css')
+def enqueue_plugin_styles():
+    styles = {}
+    for hook in hooks.get_hooks('enqueue_styles'):
+        styles.update(hook())
+
+    HTML = '\n'.join([
+        '<link rel="stylesheet" href="{}">'.format(
+            static(meta['source'])
+        ) for handle, meta in styles.items()
+    ])
+
+    return HTML


thibaudcolas:
Hi Alex, thanks for proposing those changes!

I very much agree on the problem statement, but not sure about the solution (there are a lot of tradeoffs to make whenever dealing with dependencies, it's nothing specific to what you are proposing). There are a bunch of questions that come to my mind:

- Would you have examples of Wagtail plugins that require third-party dependencies so I can have a better idea of that landscape, and how big the tradeoffs would be?
-  I see there is `version` field in your example, but it's not used in the implementation, so it seems one of the tradeoffs your solution is making is that it overlooks versioning, which seems like a big problem for projects as parallel and long-lived as third-party plugins. How likely is this to be an issue? Could we let end users and plugin authors knowingly choose between "I will run the risk of using this other version of my dependency" and "I will have multiple versions of the same thing on the page, at the risk of them conflicting"?
- How does this relate to the existing `insert_global_admin_js` and `insert_global_admin_css` from the end user's perspective? For example, for the purpose of documenting the hooks, in which use cases would you recommend using those new hooks over the old ones?

alexgleason:
@thibaudcolas Thanks for reviewing!

>there are a lot of tradeoffs to make whenever dealing with dependencies, it's nothing specific to what you are proposing

Definitely agree. I've thought about this a lot... I can say what I _really_ want is integration with a Node package.json or bower.json because they already solve a lot of problems. It's probably possible, but doesn't seem easy to combine these two ecosystems and then maintain the bridge.

Instead I've opted for a solution [similar to WordPress](https://developer.wordpress.org/reference/functions/wp_enqueue_script/). I wouldn't normally turn to WordPress for inspiration, but the solution is simple and it resolves the majority of problems surrounding front-end dependencies.

>Wagtail plugins that require third-party dependencies5

These plugins need FontAwesome:

* [wagtailfontawesome](https://github.com/alexgleason/wagtailfontawesome)
* [wagalytics](https://github.com/tomdyson/wagalytics)
* [wagtailsocialfeed](https://github.com/LUKKIEN/wagtailsocialfeed)

There are more plugins that use FontAwesome, but I can't find them right now. One is a replacement for the WYSIWYG editor I saw a while back.

There are probably more, but I don't know of them. I had considered using [Dragula](https://bevacqua.github.io/dragula/) for one of my Wagtail plugins but got distracted by this issue.

> it seems one of the tradeoffs your solution is making is that it overlooks versioning

This is also what WordPress does. I've never had a problem with dependencies on WordPress before, so I've decided to take their path here and throw caution to the wind regarding versioning. In reality it will probably be fine in most cases to rely on the latest version of a dependency. If we can solve the issue for 90% of users that's great, as long as there is a workaround for the other 10%.

>Could we let end users and plugin authors knowingly choose between "I will run the risk of using this other version of my dependency" and "I will have multiple versions of the same thing on the page, at the risk of them conflicting"?

If plugin authors set opposing settings this wouldn't work. If there were a way to override the plugin settings for each plugin within a project that could be a workaround. For instance:

* wagalytics needs fontawesome, prefers duplication over version adjustment
* wagtailsocialfeed needs fontawesome, prefers version adjustment over duplication
* wagalytics works fine with version adjustment and doesn't actually need duplication, despite its preference

So we can do:

```
force_script_behavior('wagalytics', 'fontawesome', 'adjust-version')
```

in settings.py or something. Sort of thinking out loud here.

```
in which use cases would you recommend using those new hooks over the old ones?
```

Personally I would deprecate the old hooks entirely. I don't like that they make you write HTML in Python instead of accept a filename.

thibaudcolas:
Just an update that I've read your reply and it makes perfect sense to me. I'm not willing to move too fast because we are just about to merge a big PR that introduces more front-end code, more Webpack, more npm (#3012) – this problem will only get bigger. I'm hopeful that either npm or Webpack can bring something to the table to solve it.

Also – I've started authoring my first third-party Wagtail app (https://github.com/springload/wagtaildraftail), and am of course running into this exact same issue. It's clocking in at nearly 800kb of JS, so definitely want to find a good fix for this.

thibaudcolas:
Related: #2315.

Another update!

I've been thinking about this quite a lot. I think I've found the solution to this problem that works for me, based on npm + Webpack to manage and bundle front-end dependencies. It's quite involved of a setup, and not fleshed out enough for me to elaborate further, but should be very easy to consume... for users of npm and Webpack.

---

As @alexgleason points out no solution has to address the needs of all users, and although I think the npm + Webpack one would be less hacky, the 90% of people who use and write Wagtail plugins without module bundling won't care too much about the correctness.

I would like to introduce the alternative ways to make this work at the same tim – thinking it will make it easier for plugin authors to see the limits of each approach, pick the right one depending on their need, and it will produce less churn in the long run.

thibaudcolas:
I’ll close this now. Even though this is a real problem, I think this particular solution has too many drawbacks to be worth it. The main ones I see are:

- It goes against the current best practice of bundling client-side code according to its usage rather than its structure (e.g. single bundle for all code, or dynamic split bundles, etc.).
- As Wagtail plugins are rarely updated, multiple plugins would easily end up using different (but most likely compatible) versions of the same dependency. I don't think it's a good idea for Wagtail to try and resolve version requirements semver-style.

I also see most dependencies falling in one of two categories that I think should be handled very differently:

- Very specific dependencies (say a drag and drop engine, or an autocomplete field) most likely would only come from a single extension, and not benefit much from a de-dupe system.
- Very generic dependencies (jQuery, jQuery plugin, icon set) should be either provided by Wagtail, or dropped in favour of vanilla JS / SVG.

---

For people looking into this, the best solution I see so far is to rely on Wagtail’s provided dependencies, and if some are missing ask for more to be exposed.

- Use Wagtail's jQuery directly, or go jQuery-free: http://youmightnotneedjquery.com/
- Use SVG icons (#4821) loaded on demand, rather than a humongous icon pack like FontAwesome

Issue #2876, opened 2016-07-28
Make streamvalue mutable


fabiommendes:
Implement the MutableSequence methods for StreamValue. It would be useful to be able to insert stream blocks programatically without having to recreate the whole stream value object. I am working on an implementation and can submit a pull request later today.

gasman:
I'm a bit wary of making this change - I'll need to think about it some more.

When I designed the StreamField APIs, I imagined the data as always consisting of immutable [value objects](https://en.wikipedia.org/wiki/Value_object), and I'm not sure what the consequences will be if we break that assumption. However, we haven't been rigorous about that for StructBlock and ListBlock values, and that doesn't seem to have caused any problems...

fabiommendes:
+1 for Immutable by default. However, in this case the API for changing a stream value is very fragile. If page.body is a stream value, it breaks in simple operations such as 
```python
page.body = list(page.body)
```
Hence we cannot convert the immutable sequence to a list, change its contents, and save the immutable version back in the end. Changing stream_data directly is also a bad idea since the user has to check if the stream value is lazy or not and insert data accordingly. This is bug-prone and IMHO .stream_data and .is_lazy attributes are just implementation details that the user should not fiddle with.

When inserting new blocks in a stream block, I want to avoid this ugly code:
```python
body = [(blk.block_type, blk.value) for blk in page.body]
body.append(('paragraph', '<p>hello</p>'))
page.body = body
```

gasman:
OK - having thought about this some more, I'm now more comfortable about the idea of making StreamValue mutable. Looking forward to seeing the patch :-)

fabiommendes:
I sent the pull request.

I'm having a bit of trouble running tox and the full test suite in my computer. "make tests" pass, but tox fails even in the environments available in my computer. What is the recommended workflow? Can I rely that CI will catch any problems not present in my computer?

gasman:
Leaving CI to do the full (all-versions) test run is fine. I almost never run tox locally, since it doesn't play well with Vagrant VMs (it tries to create symlinks, which aren't supported by Vagrant shared folders).

zerolab:
The PR is #2886

gasman:
Completed in #6485

PR #2886, opened 2016-08-02
Make StreamField mutable (issue #2876)

fabiommendes:
See #2876


gasman:
Thanks @fabiommendes! Sorry for the delay in reviewing...

I think writing back to `stream_data` probably isn't the best approach here, given the need to handle the lazy and non-lazy cases. (Wrapping the value in a dict vs. a tuple isn't enough - the dict representation also needs to have the value in raw JSON-ish form, e.g. an ID rather than a Page for PageChooserBlock.) Instead, we should just build a StreamChild and write that to `_bound_blocks` - once an entry exists there, `__getitem__` will use that one in preference to looking at `stream_data`.

In fact, this is probably a good opportunity to rename `stream_data` to make its purpose clearer: I'd suggest renaming it to `_initial_data` or `_initial_raw_data` depending whether the StreamValue is lazy or not:

    def __init__(self, stream_block, stream_data, is_lazy=False, raw_text=None):
        self.is_lazy = is_lazy
        if is_lazy:
            self._initial_raw_data = stream_data
        else:
            self._initial_data = stream_data

Also, I think it would be better for `__setitem__` to NOT accept the dict / JSON representation of `value`. While we need to treat the StreamChild and tuple representations as basically interchangeable (because there are situations where it's difficult or impossible to define a StreamChild, such as an `initial` parameter of a StreamField), the dict representation is a very different beast - you have to go through an explicit conversion step (`to_python` / `get_prep_value`) to retrieve it, and normally a developer would only encounter it if they were manipulating the JSON directly.

gasman:
Now superseded by #6485.

PR #2887, opened 2016-08-03
Re: PageChooserBlock supports limiting of page type

nutztherookie:
We kinda forgot about this, so I'm re-proposing it,

Reviving pull-request #2449


gasman:
Implemented in #3072.

PR #2895, opened 2016-08-06
Page create view now checks PageModel.can_create_at method

### LINKED ISSUE #2892: Page create view doesn't check can_create_at method
It seems to me that we need to call `can_create_at` in [page `create` view](https://github.com/torchbox/wagtail/blob/5ce99e85c413fb13743d337f6faa5addcec61b4d/wagtail/wagtailadmin/views/pages.py#L176).

Otherwise user will be able to create a page using a direct url, even if `can_create_at` returns `False`.

--

For example, you need to limit number of home pages to 1.

```python


class HomePage(Page):
    # Only allow creating HomePages at the root level
    parent_page_types = ['wagtailcore.Page']

    @classmethod
    def can_create_at(cls, parent):
        result = super().can_create_at(parent)
        if not result:
            return result

        homepage_content_type = ContentType.objects.get_for_model(HomePage)
        return not parent.get_children().filter(content_type=homepage_content_type).count()
```

This code works at the first glance: "Home page" will not be shown on the "Create page in Root" screen, but you will be able to create a page using direct url.

m1kola:
Fixes #2892

### DIFF
--- a/wagtail/wagtailcore/tests/test_page_model.py
+++ b/wagtail/wagtailcore/tests/test_page_model.py
@@ -8,6 +8,7 @@ from django.contrib.auth import get_user_model
 from django.contrib.auth.models import AnonymousUser
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ValidationError
+from django.core.urlresolvers import reverse
 from django.http import Http404, HttpRequest
 from django.test import Client, TestCase
 from django.test.client import RequestFactory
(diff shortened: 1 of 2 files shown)


wagtail/wagtailcore/tests/test_page_model.py
```diff
@@ -1056,6 +1057,36 @@ def test_url_path_can_exceed_255_characters(self):
         self.assertEqual(new_christmas_event.url_path, expected_url_path)
 
 
+class TestIssue2892(TestCase, WagtailTestUtils):
+    """
+    Issue 756 reports that page can be created using a direct url,
```
kaedroho: 756?

m1kola: Should be 2892. I've copy-pasted this from another test.

wagtail/wagtailcore/tests/test_page_model.py
```diff
@@ -1056,6 +1057,36 @@ def test_url_path_can_exceed_255_characters(self):
         self.assertEqual(new_christmas_event.url_path, expected_url_path)
 
 
+class TestIssue2892(TestCase, WagtailTestUtils):
```
kaedroho: This test should be moved to ``wagtailadmin/tests/test_pages_views.py``

kaedroho:
Merged in 07b7dd294009d85bbc586beb497b66811db40c24 (moved the test in 9bc650245944760a55b3d78341d527cdfce78244) thanks!

Backported to ``stable/1.6.x`` in ec9e4a090ab0289e627f6d7ba55fdc23966c22b3 and 4e75ff383e50ace00ca800c76fd9bc521b8d6887

PR #2899, opened 2016-08-08
Disable browser validation for images and docs

### LINKED ISSUE #2897: It's not possible to edit image details with a blank "File" form field
Browser validation doesn't allow to save image details if a "File" form field is blank.

I can reproduce it with Django 1.10 and Wagtail 1.6rc1. Everything works as expected with Django 1.9.9.

screen shot 2016-08-08 at 14 36 14

m1kola:
Django 1.10 adds browser validation to all required fields by default. We don't need this behaviour for images and docs.

Note: there are other possible solutions. We can add a new form for an edit view that will define file field as not required.

Fixes #2897

### DIFF
--- a/wagtail/wagtaildocs/forms.py
+++ b/wagtail/wagtaildocs/forms.py
@@ -13,6 +13,7 @@ from wagtail.wagtaildocs.permissions import permission_policy as documents_permi
 
 class BaseDocumentForm(BaseCollectionMemberForm):
     permission_policy = documents_permission_policy
+    use_required_attribute = False
 
 
 def get_document_form(model):
--- a/wagtail/wagtailimages/forms.py
+++ b/wagtail/wagtailimages/forms.py
@@ -25,6 +25,7 @@ def formfield_for_dbfield(db_field, **kwargs):
 
 class BaseImageForm(BaseCollectionMemberForm):
     permission_policy = images_permission_policy
+    use_required_attribute = False
 
 
 def get_image_form(model):


m1kola:
Probably we also need to add tests to check html representation of forms.

gasman:
It sounds to me like this might be a core Django bug. Django implements 'edit mode' on file fields using logic outside of the `<input type="file">` tag (because browsers won't let you pre-populate the element). This _is_ logically a required field at the Django level (because you can't have an Image or Document without a file), but it's incorrect for Django to add the client-side validation (because the client doesn't know about the Django-side logic). I'll investigate this further.

m1kola:
Makes sense.

Django adds `required` attribute for all fields which were defined as not hidden and required on `BoundField` level (see [this method](https://github.com/django/django/blob/1.10/django/forms/boundfield.py#L233) for details).

tomdyson:
Should we use @m1kola's solution while we wait for a fix in Django?

gasman:
I'm not very keen on the fix in its current form, because A) client-side validation _is_ a nice feature (or will be, when Django fixes it!) and it's a shame to disable it completely for these forms; and B) it's [apparently quite common](https://github.com/torchbox/wagtail/issues?utf8=%E2%9C%93&q=multipart%2Fform-data) for people to use FileFields on other forms throughout Wagtail (pages, snippets, modeladmin, settings, custom user models...) so we ought to have a fix that covers those too.

It's entirely possible that Django 1.10.1 will come out (with this fix in place - it's currently tagged as a release blocker) before our release date for Wagtail 1.6 - if so, I think we should just rely on Django's fix and not try to fix it at our end. 

It's worth noting that django.contrib.admin disables client-side validation by adding a `novalidate` attribute to the `<form>` element ([source](https://github.com/django/django/blob/ecb59cc6579402b68ddfd4499bf30edacf5963be/django/contrib/admin/templates/admin/change_form.html#L40)). That might be a cleaner way to deal with this, since the relevant forms also need to output `enctype="multipart/form-data"` at that point. Something like:

    <form action="..." method="..."
        {% if form.is_multipart %}
            enctype="multipart/form-data"
            {% if django_1_10_0 %}novalidate{% endif %}
        {% endif %}
    >

m1kola:
Fixed in Django 1.10.1.

PR #2900, opened 2016-08-08
Change AbstractForm to a mixin class called FormPageMixin

kurtrwall:
- Since AbstractForm inherited from Page, any custom Page subclass couldn't use it.
- Changing it to a mixin allows subclasses of Page to gain Form Builder.
### TODO
- ~~Write tests~~
### Testing
- Run tests: `python runtests.py wagtail.wagtailforms`
- Create a Page class that inherits from Page and the mixin with the form fields related and the appropriate template files made:

``` python
from wagtail.wagtailforms.models import FormPageMixin, AbstractFormField

class FormField(AbstractFormField):
    page = ParentalKey('FormPage', related_name='form_fields')

class FormPage(FormPageMixin, Page):
        InlinePanel('form_fields', label="Form fields")
    ]
```
- Create migration file, run the migration, and login to the admin
- Add a page with fields and test the page to make sure it submits properly.
### Other notes

This spawned from a discussion on [this thread](https://groups.google.com/forum/#!topic/wagtail-developers/K4_LGKzO8gE) where @gasman encouraged a pull request for it.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -37,6 +37,7 @@ Changelog
  * Fix: Non-text content is now preserved when adding or editing a link within rich text (Matt Westcott)
  * Fix: Fixed preview when `SECURE_SSL_REDIRECT = True` (Aymeric Augustin)
  * Fix: Prevent hang when truncating an image filename without an extension (Ricky Robinett)
+ * Deprecated ``AbstractForm`` in favor of mixin class ``FormPageMixin``.
 
 
 1.6.3 (30.09.2016)
--- a/docs/reference/contrib/forms.rst
+++ b/docs/reference/contrib/forms.rst
@@ -4,7 +4,7 @@
 Form builder
 ============
 
-The ``wagtailforms`` module allows you to set up single-page forms, such as a 'Contact us' form, as pages of a Wagtail site. It provides a set of base models that site implementers can extend to create their own ``FormPage`` type with their own site-specific templates. Once a page type has been set up in this way, editors can build forms within the usual page editor, consisting of any number of fields. Form submissions are stored for later retrieval through a new 'Forms' section within the Wagtail admin interface; in addition, they can be optionally e-mailed to an address specified by the editor.
+The ``wagtailforms`` module allows you to set up single-page forms, such as a 'Contact us' form, as pages of a Wagtail site. It provides a set of base models that site implementers can extend to create their own ``FormPage`` type with their own site-specific templates. Once a page type has been set up in this way, editors can build forms within the usual page editor, consisting of any number of fields. Form submissions are stored for later retrieval through a new 'Forms' section within the Wagtail admin interface.
 
 
 Usage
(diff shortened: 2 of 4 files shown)


timheap:
I like this change, @kurtrwall! We should keep an `AbstractForm` class for backwards compatibility reasons though, in case people have extended from it. It can be as simple as

```python
class AbstractForm(FormPageMixin, Page):
    class Meta:
        abstract = True
```

Can you also make the same change to the `AbstractEmailForm` class, by splitting it out in to a `EmailFormPageMixin` class, but also keeping an `AbstractEmailForm` class around for those who are already using it?

The docs can be updated to only refer to the `FormPageMixin` or `EmailFormPageMixin` classes, and changing the code to use the mixins along with `Page` correctly. The old approach using the `AbstractForm` classes need not be mentioned. This is how the `RoutablePage` class [was deprecated](https://github.com/torchbox/wagtail/pull/560/files) in favour of a `RoutablePageMixin` class previously.

kurtrwall:
@timheap thanks for the feedback! Working on changing it up to fit your suggestions, then adding tests.

kurtrwall:
This has been updated to reflect the requested changes.

I should note that the `forms` tests have been changed to use the mixin classes and that the deprecated models `FormPage` and `EmailFormPage`  are no longer used in tests.

If this is not desirable, I can add them back so they're still covered.

kurtrwall:
Actually, I'm having some trouble with a test. I'll check back.

kurtrwall:
As it turns out, the `EmailFormPageMixin` will not work because it would involve overriding the `__init__` method of the Page class that uses it to do the work of assigning the arguments to the address and subject fields without passing them to the `Page.__init__`.

At any rate, the docs and tests have been updated to reflect the changes.

I should add that I just reused existing tests since they just use the Page class `AbstractEmailForm` (which is created with the mixin), instead of creating duplicate tests for a separate Page model that uses the mixin. I felt that this was adequate to test that the mixin works as expected. Let me know if I need to add more.

timheap:
Thanks for making those changes, @kurtrwall. I had a go at refactoring the `AbstractEmailForm` in to a `EmailFormPageMixin` [over on my fork](https://github.com/takeflight/wagtail/commits/refactor/abstractform-to-mixin), but came across a strange bug that only affects Python 2. I've [logged a bug against Django](https://code.djangoproject.com/ticket/27337) to see if they have any insight.

I was having trouble replicating a situation where `AbstractForm` being a `Page` subclass causes issues. I know it does, as I've had similar problems with `RoutablePage` before it was a mixin, but I can't remember what those problems are. Can you post an example where `AbstractForm` is a problem that will be fixed by making it a mixin?

kurtrwall:
As an aside, I've come across the very same MTI bug.

@timheap it causes issues when you have this situation:

1. You want base page class that all pages inherit from. 
2. A subclass of this wants to add the form builder module.
3. The subclass cannot inherit from the base page class and the `AbstractForm` class because they both inherit from `Page`.

lb-:
Potentially will be superseded by https://github.com/wagtail/wagtail/pull/9093

lb-:
Replaced with https://github.com/wagtail/wagtail/pull/9093

@kurtrwall I have added you to the contributors list for Wagtail and also included your name in the changelog - thank you for your original PR here.

PR #2903, opened 2016-08-09
ChoiceBlock should accept "include_blank=False"

nutztherookie:
Just like the referenced `modelfield`, ChoiceBlock should be able to NOT inlude a blank option.

### DIFF
--- a/docs/topics/streamfield.rst
+++ b/docs/topics/streamfield.rst
@@ -206,6 +206,9 @@ A dropdown select box for choosing from a list of choices. The following keyword
 ``help_text``
   Help text to display alongside the field.
 
+``include_blank`` (default: True)
+  If false, the blank choice (``---------``) will not be rendered.
+
 ``ChoiceBlock`` can also be subclassed to produce a reusable block with the same list of choices everywhere it is used. For example, a block definition such as:
 
 .. code-block:: python
--- a/wagtail/wagtailcore/blocks/field_block.py
+++ b/wagtail/wagtailcore/blocks/field_block.py
@@ -316,7 +316,7 @@ class ChoiceBlock(FieldBlock):
 
     choices = ()
 
-    def __init__(self, choices=None, required=True, help_text=None, **kwargs):
+    def __init__(self, choices=None, required=True, help_text=None, include_blank=True, **kwargs):
         if choices is None:
             # no choices specified, so pick up the choice list defined at the class level
             choices = list(self.choices)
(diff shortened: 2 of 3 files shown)


gasman:
Thanks! Please can you add a test for this, and update the docs?

nutztherookie:
what does 'milestone: 1.8, 1.7' mean?

gasman:
@nutztherookie That's Github's way of saying that the milestone was changed from 1.7 to 1.8. (No idea why they can't write it in a more sensible way...) The milestone was changed because we chose to set a fixed deadline for each release, and inevitably didn't get through our backlog of PRs to review by that date. (Hoping to clear the backlog within the next couple of months!)

gasman:
Hi @nutztherookie,

Now that I've looked at this in more detail, I think that adding `include_blank` as a configuration option probably isn't the logically correct thing to do here. Sorry for not catching this earlier...

Looking at Django's model field code, I have the impression that `include_blank` was never designed to be a user-accessible option - it's not mentioned anywhere in the docs, for example. Rather, it's an internal property that gets [set according to the following logic](https://github.com/django/django/blob/1.10.3/django/db/models/fields/__init__.py#L850):

* a non-required field always has a blank option
* a required field does NOT have a blank option if there is an initial value (i.e. we're editing an existing instance) or a default value; otherwise, it does have a blank option.

I think ChoiceBlock should take the same approach. Choosing a suitable value for `include_blank` should _not_ be the responsibility of the developer, because:

* a required field with no default needs to show the blank option on adding a new item, and hide it when editing an existing one - this behaviour can't be represented by a simple `include_blank = True` or `False` definition;
* More abstractly: a Wagtail developer should never have to care about the low-level behaviour of a select box. We should show or hide the blank option automatically, depending on the 'required' flag and the presence of an initial or default value.

I assume that ChoiceBlock is - currently - not getting this behaviour right, and that's what motivated you to submit this fix :-) So, the next step would be to work out which specific combinations of `required` / `default` / initial value aren't behaving as expected, and create a fix for those cases. If you're able to investigate this further, that would be much appreciated!

gasman:
Have now created issue #3163 to cover this.

nutztherookie:
Hey @gasman ,

thank you for all the research you've put into this. I totally get what you're saying. And I understand the new issue you've created and I agree it's the right solution for this in general.
However since this is a bigger thing and it might not be fixed anytime soon; wouldn't it be feasible to include my solution for the time being and throw it out once the other issue is fixed?
I understand that that holds the possibility of forgetting about this and leaving it in. But I don't think the argument would hurt anybody if it still existed.

Thanks

gasman:
Hi @nutztherookie,

I'm really not keen on adding this as a configuration option - it sets a bad precedent of developers having to manage low-level details, and even as a temporary addition it's going to have an impact on the codebase for several releases (once it's in place, and developers are relying on it, we have to give advance warning of it being dropped).

You're right that we shouldn't stall over finding a 'perfect' solution if a simpler one exists, though. How about we change the behaviour so that a ChoiceBlock with a default value, and `required=True`, doesn't get the blank option? That should be an easy fix, and as far as I can tell, addresses the specific case that `include_blank=False` would solve. (The more general case that needs fixing is: hide the blank option when `required=True` and the field has an existing value, default or otherwise. `include_blank` wouldn't solve this, as it doesn't correspond to a simple True or False.)

nutztherookie:
Sounds good. I shall create a pull request for that soonish. (Have been swamped lately with becoming a dad :)

nutztherookie:
Hi @gasman ,

I hope #3254 does implement what you were asking for :)

Cheers

Issue #2881, opened 2016-07-29
Remove wagtailimages.Filter as a model (second try)


gasman:
#2537 had to be pulled from the 1.6 release as it caused problems for upgrading installations with custom image models (#2817). https://github.com/torchbox/wagtail/issues/2817#issuecomment-236167560 outlines a proposal for rolling this change out in stages, which we should aim to do in 1.7 / 1.8.

gasman:
Completed in #3351

PR #2537, opened 2016-04-28
Refactor: remove Filter as a Django model.

Gagaro:
As seen with @gasman in the mailing list and in #2090 , Filter should no longer be a model. I have a few questions though.

I'm not sure if Filter should stay in the `models.py` file or be moved elsewhere. 

Also, I have some issue writing the migration. I tried something to keep the existing renditions but I have some issue such as:

     django.core.exceptions.FieldDoesNotExist: Rendition has no field named u'filter'

I'm not really use to writing data migration for Django so I'm not quite sure what's wrong with it. Any help would be appreciated, thanks :smile: .

PR #2910, opened 2016-08-12
Introduce `filter_spec` field to Rendition

gasman:
First step of #2881


wagtail/wagtailimages/models.py
```diff
@@ -458,7 +458,8 @@ def _search_for_operations(cls):
 
 
 class AbstractRendition(models.Model):
-    filter = models.ForeignKey(Filter, related_name='+')
+    filter = models.ForeignKey(Filter, related_name='+', null=True, blank=True)
```
kaedroho: What's the reason for making this field nullable?

gasman: It means that when the field is dropped in the next release, users with custom rendition models won't be prompted for a default value for the reverse migration when they run makemigrations.

(There _may_ have been more to it, but I think that's the gist of it...)

wagtail/wagtailimages/utils.py
```diff
@@ -0,0 +1,34 @@
...
+
+# Helper functions for migrating the Rendition.filter foreign key to the filter_spec field,
+# and the corresponding reverse migration
+
+def get_fill_filter_spec_migrations(app_name, rendition_model_name):
+
+    def fill_filter_spec_forward(apps, schema_editor):
+        # Populate Rendition.filter_spec with the spec string of the corresponding Filter object
+        Rendition = apps.get_model(app_name, rendition_model_name)
+        db_alias = schema_editor.connection.alias
```
kaedroho: Out of interest, what's the reason for setting the database manually?

wagtail/wagtailimages/utils.py
```diff
@@ -0,0 +1,34 @@
...
+        for rendition in Rendition.objects.using(db_alias).select_related('filter'):
```
kaedroho: Could this be done with a single ``UPDATE`` query? (we have a site with over a million renditions)

gasman: Good point - pretty sure it could be done with one bulk-update per filter record, at least.

wagtail/wagtailimages/utils.py
gasman: Not sure, actually - I kept that unchanged from #2537. @Gagaro?

Gagaro: That's how it is done in the documentation, I just copied it from there: https://docs.djangoproject.com/en/1.10/ref/migration-operations/#runpython

And specifically the warning:

```
RunPython does not magically alter the connection of the models for you; any model methods you call will go to the default database unless you give them the current database alias (available from schema_editor.connection.alias, where schema_editor is the second argument to your function).
```

wagtail/wagtailimages/utils.py
Gagaro: If I remember correctly, the `update()` method from the QuerySet does not work during data migration. But I'm not sure I tried a raw query, it would be better indeed.

kaedroho:
It is possible for new renditions to be generated at any time during deployment so if one is made after/during the migrations are run, but before the server is restarted, it will not have the filter_spec field set.

gasman:
@kaedroho Hmm, true. But isn't that equally true of _any_ migration - i.e. immediately after running a migration there's likely to be an active webserver process running the old version of the code, so there's a chance of it doing something that wrongly assumes the old database state?

Is it just a concern here because populating `filter_spec` takes a non-negligible amount of time - if so, perhaps we can mitigate that by doing a second pass over the renditions table in `fill_filter_spec_forward` (i.e. a `for rendition in Renditions.objects.filter(filter_spec__isnull=True)` loop to pick up any new renditions that were added in the meantime)? (Or do transactions muck that idea up?)

kaedroho:
> But isn't that equally true of any migration

Not really. For a schema migration, the worst that could happen is the app crashes when adding a new object.

wagtail/wagtailimages/utils.py
gasman: Updated in e7171716abc284ae9268dbf6892e284b2cec3454 to perform a bulk-update per filter.

gasman:
> Perhaps a solution to this could be to set that field's default value to a blank string? The database should then reject any renditions created with the old version of the code as it wouldn't be aware that it needs that field (assuming defaults are set by django and not the database)

Done in 4cae9e6c7efb10eba517d45de8d830b44e46545f - yep, that seems to do the trick. Tested by running the migration (resulting in a NOT NULL column with no default at the DB level) and then switching branches back to master - attempting to create a rendition then fails with an integrity error, as intended.

kaedroho:
Code changes look good to me. Going to give this a quick test on RCA, then I'll merge if it all goes well.

wagtail/wagtailimages/models.py
```diff
@@ -505,6 +508,42 @@ def get_upload_to(self, filename):
...
+
+    @classmethod
+    def check(cls, **kwargs):
+        errors = super(AbstractRendition, cls).check(**kwargs)
+
+        # If a filter_spec column exists on this model, and contains null entries, warn that
+        # a data migration needs to be performed to populate it
+
+        try:
+            null_filter_spec_exists = cls.objects.filter(filter_spec__isnull=True).exists()
```
kaedroho: This needs to be updated to ``filter_spec=''``

kaedroho:
Some timings for the data migration on RCA (34551 images):

Forwards: ``Applying rca.0043_auto_20160912_1118... OK (5.758s)``
Backwards: ``Unapplying rca.0043_auto_20160912_1118... OK (110.125s)``

This is running locally. On the production servers we have network lag which can make the migrations run much longer if there are a lot of roundtrips to the database (I recently ran one that took 25 times longer than expected).

kaedroho:
Except for two comments above, LGTM!

wagtail/wagtailimages/models.py
gasman: Fixed in 97441a7498666b3cb71f110c449f0958f16e7bd4

wagtail/wagtailimages/utils.py
```diff
@@ -9,10 +9,12 @@ def get_fill_filter_spec_migrations(app_name, rendition_model_name):
...
         Rendition = apps.get_model(app_name, rendition_model_name)
+        Filter = apps.get_model('wagtailimages', 'Filter')
+
         db_alias = schema_editor.connection.alias
-        for rendition in Rendition.objects.using(db_alias).select_related('filter'):
-            rendition.filter_spec = rendition.filter.spec
-            rendition.save()
+        for flt in Filter.objects.using(db_alias):
+            renditions = Rendition.objects.using(db_alias).filter(filter=flt, filter_spec__isnull=True)
+            renditions.update(filter_spec=flt.spec)
```
timheap: To make this even more performant, this whole block could be done using an [F()](https://docs.djangoproject.com/en/1.7/ref/models/queries/#f-expressions):

```python
from django.db.models import F
Rendition.objects.using(db_alias).filter(filter_spec__isnull=True).update(filter_spec=F('filter__spec'))
```

gasman: Not working here I'm afraid (Postgres, Django 1.10.1) - I get "django.core.exceptions.FieldError: Joined field references are not permitted in this query".

gasman:
Reverse migration now fixed in 73e5a9af7477d4e237ab357c56703f1b79c93404 to perform one update query per filter, rather than per rendition. Ready for re-review!

Note that when testing the reverse migration, you'll need to manually run `UPDATE wagtailimages_renditions SET filter_id=null` first in order to get representative results - the foreign key field won't get de-populated until we remove it outright in 1.8, so running it now would be a no-operation.

docs/releases/1.7.rst
```diff
@@ -58,3 +58,25 @@ Bug fixes
...
+
+The data model for image renditions will be changed in Wagtail 1.8 to eliminate ``Filter`` as a model. Wagtail sites using a custom image model (see :ref:`custom_image_model`) need to have a schema and data migration in place prior to upgrading to Wagtail 1.8. To create these migrations:
+
+ * Run ``manage.py makemigrations`` to create the schema migration
+ * Run ``manage.py makemigrations --empty myapp`` (replacing ``myapp`` with the name of the app containing the custom image model) to create an empty migration
+ * Edit the created migration to contain::
+
+    from wagtail.wagtailimages.utils import get_fill_filter_spec_migrations
+
+   and, for the ``operations`` list::
```
kaedroho: Needs to be indented by one more space

gasman: Are you sure? It's aligned with the word 'Edit' after the bullet point

gasman:
Thanks @kaedroho! Now merged in 796cb2ee01d6a6880ee60f5ee77ae96e7e7d5a98 + parents.

PR #2913, opened 2016-08-15
Disable client-side validation on Django 1.10 for multipart forms

### LINKED ISSUE #2897: It's not possible to edit image details with a blank "File" form field

gasman:
Client-side validation fails on forms with prefilled file upload fields -
see https://code.djangoproject.com/ticket/27037. This is fixed in Django 1.10.1,
so as a workaround we disable client-side validation (using the 'novalidate'
attribute) for forms with enctype="multipart/form-data" on Django 1.10 only.

Fixes #2897

### DIFF
--- a/wagtail/contrib/modeladmin/templates/modeladmin/create.html
+++ b/wagtail/contrib/modeladmin/templates/modeladmin/create.html
@@ -1,5 +1,5 @@
 {% extends "wagtailadmin/base.html" %}
-{% load i18n %}
+{% load i18n wagtailadmin_tags %}
 
 {% block titletag %}{{ view.get_meta_title }}{% endblock %}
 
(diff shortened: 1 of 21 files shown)


m1kola:
Looks good to me.

Do you think we need to mention this issue in the release notes or somewhere else?

gasman:
Have added a short note under "Bug fixes" - I don't think we need to mention it in more detail, as there isn't really anything an end user needs to be aware of, or any action they can take (other than "upgrade to 1.10.1 when it's released", which is standard good advice anyhow :-) )

Merged in 197d85ce9f674e319cc7d7d0fa6e0a05dc6a46d7 (master) and f753a1c66ea9e25f1f2c96fbaa87b3419bb39a1c (1.6.x).

PR #2915, opened 2016-08-15
Replace mutable default argument

### LINKED ISSUE #2914: Shouldn't use Mutable Default Argument
The chooser's [shared_context method](https://github.com/torchbox/wagtail/blob/197d85ce9f674e319cc7d7d0fa6e0a05dc6a46d7/wagtail/wagtailadmin/views/chooser.py#L13) has a [mutable default argument](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments). 

While the current implementation shouldn't cause any issues (it doesn't actually update the default argument), this should probably be fixed. I'll make a pull request with the changes.

loicteixeira:
Fix #2914.

### DIFF
--- a/wagtail/wagtailadmin/views/chooser.py
+++ b/wagtail/wagtailadmin/views/chooser.py
@@ -10,7 +10,7 @@ from wagtail.wagtailcore.models import Page
 from wagtail.wagtailcore.utils import resolve_model_string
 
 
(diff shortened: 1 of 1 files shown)


kaedroho:
Looks good to me, thanks!

PR #2919, opened 2016-08-17
Memoize wagtail_site_root_paths

### LINKED ISSUE #2916: Wagtail can make MANY requests to wagtail_site_root_paths
One site I'm building is making over 80 requests for the cache key `wagtail_site_root_paths`, which I hadn't noticed until I put a proper caching engine in place and suddenly performance actually dropped due to the round trips.

Perhaps replace all calls to `cache.get('wagtail_site_root_paths')` with a memoized result, so that we're making at most 1 request for this key?

danni:
This reduces poor performance and high load on the cache due to large numbers
of links causing a large number of requests to the same value from the
cache.

Fixes #2916

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -197,33 +197,48 @@ class Site(models.Model):
                     ]}
                 )
 
-    @staticmethod
-    def get_site_root_paths():
+    @classmethod
+    def get_site_root_paths(cls):
         """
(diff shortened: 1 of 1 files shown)


danni:
Okay, this approach is wrong, as pre-suggested on #2916 because it's on the class, not the instance and thus can't be invalidated across processes.

gasman:
Superseded by #3188

PR #2923, opened 2016-08-17
Marks names of default image formats for translation

m1kola:
I noticed that names of default image formats have not been marked for translation.


gasman:
Thanks! Merged in f804174666b1fb4d9bc726274f4442e185b1c9fc.

Issue #2833, opened 2016-07-08
process_form_submission should return the FormSubmission instance


thenewguy:
It would be nice if the [process_form_submission](https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailforms/models.py#L159) method returned the FormSubmission instance that was created so the record number could be presented to the viewer / tracked.

The FormSubmission instance should then be passed to the render context for the "landing page".

In the same line of thought... it would also be nice if there was a uuid or some sort of meaningless random token on the FormSubmission instance that could be used as a reference value to pass along to the client.

It would also be nice if this token could be appended to the email subject line to prevent mail clients from grouping into one conversation (aka gmail interface)

m1kola:
> It would be nice if the process_form_submission method returned the FormSubmission instance that was created so the record number could be presented to the viewer / tracked.

[`AbstractForm.process_form_submission`](https://github.com/torchbox/wagtail/blob/b37e8ccbcf2d15785a422ba30ab395435ee077d7/wagtail/wagtailforms/models.py#L160) and [`AbstractEmailForm.process_form_submission`](https://github.com/torchbox/wagtail/blob/b37e8ccbcf2d15785a422ba30ab395435ee077d7/wagtail/wagtailforms/models.py#L222) methods now return `FormSubmission` instance. See #2926 for details.

> The FormSubmission instance should then be passed to the render context for the "landing page".

But we still need to pass a `FormSubmission` instance into a template context on a successful form submission.

> In the same line of thought... it would also be nice if there was a uuid or some sort of meaningless random token on the FormSubmission instance that could be used as a reference value to pass along to the client.

You will be able to define your own form submission class once #2617 will be merged.

> It would also be nice if this token could be appended to the email subject line to prevent mail clients from grouping into one conversation (aka gmail interface)

I don't think that we need to change a default behaviour, but in Wagtail 1.7 you will be able to override the `AbstractEmailForm.send_mail` method to define your own send mail logic.

lb-:
Can this be closed? @m1kola

m1kola:
@lb- I think this bit is still not implemented:

> The FormSubmission instance should then be passed to the render context for the "landing page".

See: https://github.com/wagtail/wagtail/blob/e8720c7627b5f92d23502830198f7119f461776d/wagtail/wagtailforms/models.py#L263-L286

But probably, we should create a separate issue for this.

lb-:
@m1kola I can to a PR for that last change with some tests.

[...]

PR #2926, opened 2016-08-18
Extract send_mail logic in AbstractEmailForm

tleguijt:
In some cases it has to be possible to override the behaviour of sending out email while processing the emailform, for example to support HTML mail or provide your custom template.

Since `super(AbstractEmailForm, self).process_form_submission(form)` is called and this super method is still need, I've moved the send_mail logic to a separate function which can be overridden.

While working on it I took the liberty to tackle the issue described in #2833

### DIFF
--- a/wagtail/wagtailforms/models.py
+++ b/wagtail/wagtailforms/models.py
@@ -157,7 +157,7 @@ class AbstractForm(Page):
         return form_class(*args, **form_params)
 
     def process_form_submission(self, form):
-        FormSubmission.objects.create(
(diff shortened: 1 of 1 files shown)


m1kola:
Thanks @tleguijt and congratulations on your first commit into Wagtail! :tada:

Merged in b37e8ccbcf2d15785a422ba30ab395435ee077d7 + parent.

P.S.: I renamed the method from `send_form_mail` to `send_mail`. Hope that's ok.

PR #2930, opened 2016-08-19
Update IntegerBlock docs to show correct kwargs

bufke:
It should be max and min value - not length. See [here](https://github.com/torchbox/wagtail/blob/07c3ba84fbfcb1f5bf305e47bc6c8f9dbbbe291e/wagtail/wagtailcore/blocks/field_block.py#L306).

So this would work 

`blocks.IntegerBlock(max_value=10, min_value=0)`

but this wouldn't do anything

`blocks.IntegerBlock(max_length=10, min_length=0)`


gasman:
Good catch, thanks!

PR #2937, opened 2016-08-23
Add setting to disable Unicode slug support

### LINKED ISSUE #2927: Upgrading to wagtail 1.6 breaks, with DEBUG=False and ManifestStaticFilesStorage as STATICFILES_STORAGE
After upgrading to wagtail 1.6 wagtail admin fails to load a bunch of JS scripts.  None of the files specified in `wagtailadmin/templates/wagtailadmin/pages/_editor_js.html` is rendered on to the edit or create page HTML.

Found that this Django setting was causing it

```python
DEBUG=False
STATICFILES_STORAGE = 'django.contrib.staticfiles.storage.ManifestStaticFilesStorage'
```

Switching to a normal Django static storage works. but we would like to fingerprint our assets. This also prevents us from using custom static files storage like whitenoise.  This was not an issue with the previous versions of wagtail (1.5.3).

**Setup**
* Python 3.4
* Django 1.8.13
* Wagtail 1.6

**Steps to reproduce**
```shell
pip install django==1.8
wagtail start clean
python3 manage.py migrate
```

Edit the settings to 
>
```python
DEBUG=False
STATICFILES_STORAGE = 'django.contrib.staticfiles.storage.ManifestStaticFilesStorage'
```

```shell
python manage.py collectstatic --clear
python manage.py runserver 8000
```

@gasman @alexgleason @m1kola can you guys take a look ?

gasman:
Fixes #2927, and adds formal support for disabling Unicode support in page slugs (as requested at https://groups.google.com/d/msg/wagtail/24zNAEHHNuk/iGsgKC93BQAJ), via a new setting `WAGTAIL_ALLOW_UNICODE_SLUGS`.

On Django 1.8, this setting is forced to False. Previously, disabling Unicode slugs on Django 1.8 was achieved by importing the (non-Unicode-supporting) urlify.js file from Django admin, and relying on the 404 on xregexp.min.js to be silently ignored - this fails when ManifestStaticFilesStorage is in use, and also does not account for the possibility that django.contrib.admin might be omitted from INSTALLED_APPS. We now ship our own copies of these JS files.


m1kola:
Looks good to me. I've tested this PR with Django 1.8.7 and Django 1.9.9. Both versions work well.

* 491c1b64eff289dc25bad1522b893497c2331c16 + parent for `stable/1.6.x`
* 9b81d3b6670c4d7a313f802f000fae1259100b56 + parent for `master`

PR #2938, opened 2016-08-23
Fix for wagtail-userbar on touch devices

coredumperror:
This patch fixes the javascript error that was breaking the userbar (the little bird circle on editable pages) on all touch devices. Tapping the userbar icon will now properly toggle the menu, instead of doing nothing.


gasman:
Merged in f6b47235a329f54834cc3db549692a0de076c17b - thanks!

PR #2947, opened 2016-08-25
Bumped supported Django version to 1.10

elwinbuisman:
Bumped supported Django version to 1.10 according to requirements in setup.py


gasman:
Thanks! Merged in fd8f4b22fb3ebb9554e63a95079566159484e973.

PR #2842, opened 2016-07-15
Pin html5lib to 0.999999

SableWalnut:
Temporary fix for issue in html5lib update causing beautifulsoup4 (bs4) to raise an error, meaning wagtail tests fail in master, 1.5.x and 1.4.x, haven't tested before that yet.

A new version of html5lib was released, 0.99999999, and this release seems to be causing an error. The error seems to be being raised within beautifulsoup4 (bs4). It seems a bug report or fix for it is yet to be logged/pushed for bs4.

```
  File "/vagrant/repos/wagtail/wagtail/tests/testapp/models.py", line 22, in <module>
    from wagtail.wagtailcore.blocks import CharBlock, RichTextBlock
  File "/vagrant/repos/wagtail/wagtail/wagtailcore/blocks/__init__.py", line 5, in <module>
    from .field_block import *  # NOQA
  File "/vagrant/repos/wagtail/wagtail/wagtailcore/blocks/field_block.py", line 15, in <module>
    from wagtail.wagtailcore.rich_text import RichText
  File "/vagrant/repos/wagtail/wagtail/wagtailcore/rich_text.py", line 11, in <module>
    from wagtail.wagtailcore.whitelist import Whitelister
  File "/vagrant/repos/wagtail/wagtail/wagtailcore/whitelist.py", line 9, in <module>
    from bs4 import BeautifulSoup, Comment, NavigableString, Tag
  File "/home/vagrant/venv/lib/python3.4/site-packages/bs4/__init__.py", line 30, in <module>
    from .builder import builder_registry, ParserRejectedMarkup
  File "/home/vagrant/venv/lib/python3.4/site-packages/bs4/builder/__init__.py", line 314, in <module>
    from . import _html5lib
  File "/home/vagrant/venv/lib/python3.4/site-packages/bs4/builder/_html5lib.py", line 70, in <module>
    class TreeBuilderForHtml5lib(html5lib.treebuilders._base.TreeBuilder):
AttributeError: 'module' object has no attribute '_base'
```

PR #2948, opened 2016-08-25
Upgrade BeautifulSoup→4.5.1; restore html5lib import

nimasmi:
It seems that #2842 was a temporary measure. BeautifulSoup 4.5.1 fixes the html5lib dependency itself, see https://bugs.launchpad.net/beautifulsoup/+bug/1603299. This change removes the temporary html5lib explicit version specification.


gasman:
Merged in ae3fba183aa8c879f592bb70c9cef71a946b9126 - thanks!

Issue #2950, opened 2016-08-25
Strings had not been marked for translation on password form


m1kola:
Strings had not been marked for translation on password form. They can be found in this template: [`wagtail/wagtailcore/templates/wagtailcore/password_required.html`](https://github.com/torchbox/wagtail/blob/9b81d3b6670c4d7a313f802f000fae1259100b56/wagtail/wagtailcore/templates/wagtailcore/password_required.html)

How to reach this form from the UI:

> Users with publish permission on a page can set it to be private by clicking the ‘Privacy’ control in the top right corner of the page explorer or editing interface, and setting a password. Users visiting this page, or any of its subpages, will be prompted to enter a password before they can view the page.

[More docs](http://docs.wagtail.io/en/v1.6/advanced_topics/privacy.html#private-pages).

gasman:
Fixed in #2955

PR #2955, opened 2016-08-27
Enable translation on the password required form for private pages

jjanssen:
Fix for #2950. 

Although I'm note sure on how the handle the actual translations. Couldn't find documentation on that. Are generating new .po files also a part of the pull request or is this done through the base project?


gasman:
@jjanssen Thanks! We generate the .po files as a separate step (https://github.com/torchbox/wagtail/wiki/Managing-Wagtail-translations) so no need to worry about it here.

Issue #950, opened 2015-01-30
Rendition delete receiver should listen to post_delete instead of pre_delete


thenewguy:
Noticed this when copying listeners over to my custom rendition model.  Shouldn't [models.py#L369](https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailimages/models.py#L369) connect to post_delete instead of pre_delete so that the file isn't deleted if the Rendition.delete method is interrupted/cancelled?

gasman:
Fixed in #2963

PR #2963, opened 2016-08-30
Image signals should listen to post_delete instead of pre_delete

jjanssen:
Fix for #950 (Rendition delete receiver should listen to post_delete instead of pre_delete).
In this case it is also relevant for deleting images (not just renditions) + documents. 
So I applied the change there as well.


gasman:
Merged in 57b7f4b56e895a9619b76b5adf421755a5394989 + parents - thanks!

PR #2964, opened 2016-08-31
No longer say that ATOMIC_REBUILD is "Experimental"

kaedroho:
We've used this in production on a few sites for over a year now and haven't hit any problems.


tomdyson:
LGTM

PR #2966, opened 2016-08-31
Fix checkbox initial values on group permission edit form on Django 1.10

### LINKED ISSUE #2962: Group Permissions: persistent value on checkbox
When I select "Can access Wagtail admin" checkbox, the value does not persist.

The same on "Object Permissions" (Redirect, User, Site, Group).

Wagtail 1.6.1
Django 1.10
Python 2.7.9

gasman:
Fixes #2962

Django 1.10 changed the internal representation of initial values of ModelChoiceFields; they are now model instances rather than IDs. This broke the format_permissions tag, which inspected the initial value of the field in order to reconstruct it as checkboxes. This internal poking-around is now avoided entirely, by setting GroupForm to use the CheckboxSelectMultiple widget itself; format_permissions simply has to reorganise those checkboxes appropriately, rather than building them from first principles.

**IMPORTANT:** The new form rendering relies on the `plain-checkbox-label` style added in a0efc0d10e176b52822376a7ff78ff5a0c61b90d. When backporting to stable/1.6.x, the CSS rule from there will need to be added to _forms.scss.

### DIFF
--- a/wagtail/wagtailusers/forms.py
+++ b/wagtail/wagtailusers/forms.py
@@ -218,6 +218,9 @@ class GroupForm(forms.ModelForm):
     class Meta:
         model = Group
         fields = ("name", "permissions", )
+        widgets = {
+            'permissions': forms.CheckboxSelectMultiple(),
+        }
 
     def clean_name(self):
         # Since Group.name is unique, this check is redundant,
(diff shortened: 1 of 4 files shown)


m1kola:
I've tested it on Django 1.10 and 1.9.9 version and on our `master` and `stable/1.6` branches. Looks good to me!

Merged into:
* `master` in 8a41baa87d79afe58f33394656dbc8b7fc8fd858 (release notes) + 49dae4ee10b05bebcf3def98e66b248f380f3260
* `stable/1.6` in 8dad5047c752edda556a591ee6c612691140cec0 (release notes) + d195bd3e261c97e6d5341de914595376d19bc011 (styles) + 7b5d7c8bc46c8713bcfce7c296113865060395b8

Thanks!

PR #2967, opened 2016-09-02
Configure Travis to test against Django's stable/1.10.x branch

### LINKED ISSUE #2592: Set up Travis CI to track stable branches of Django as well as releases
In addition to testing against PyPI Django releases, we should run our tests against a fresh git checkout of Django's stable branches (`stable/1.8.x` / `stable/1.9.x`). This would allow us to catch and report breaking regressions in Django (such as the one that happened in 1.8.8 / 1.9.1 - http://docs.wagtail.io/en/v1.4.4/releases/1.3.1.html) before they make it into a Django release.

gasman:
Fixes #2592

There are probably better ways of doing this - logically this ought to be a CI task that tracks the Django repo so that any failures are tied to the Django commit that caused them, rather than a (probably) unrelated Wagtail commit. But that requires more Travis/Drone-fu than I've got, so this will do for now :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,6 +68,10 @@ matrix:
      python: 3.5
    - env: TOXENV=py35-dj110-postgres-elasticsearch2 INSTALL_ELASTICSEARCH2=yes
      python: 3.5
+   - env: TOXENV=py35-dj110head-postgres-noelasticsearch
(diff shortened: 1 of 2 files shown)


kaedroho:
Thanks!

PR #2969, opened 2016-09-02
Fix bug with forcing a user to exist on request

digia:
When testing a user doesn't always exist on the request. Currently to directly test a view one would have to mock an empty user for the request. 

This removes the need for an empty user without changing the original goal - i believe.

PS: first time submitting to wagtail. I looked for information on how to properly submit a pull request, for which i didn't find anything specific to a 3rd party trying to contribute. If i'm doing this wrong let me know happy to adjust. Apologizes in advance.

### DIFF
--- a/wagtail/wagtailadmin/templatetags/wagtailuserbar.py
+++ b/wagtail/wagtailadmin/templatetags/wagtailuserbar.py
@@ -36,8 +36,14 @@ def wagtailuserbar(context, position='bottom-right'):
     except KeyError:
         return ''
 
+    # Don't render without a user because we can't check their permissions
+    try:
+        user = request.user
+    except AttributeError:
+        return ''
+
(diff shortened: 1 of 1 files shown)


gasman:
Thanks! Merged in 875d2fcbd465b143f67154f2a2a90fdd8b9e55f3.

As far as I can see, `request.user` _should_ always exist providing you have `django.contrib.auth.middleware.AuthenticationMiddleware` included in your middlewares list - perhaps this is missing from your test settings? Either way, happy to merge this as it certainly can't hurt to have the extra check in place.

(Everything's fine with this pull request, by the way! Just one recommendation I'd make - it's a good idea to start a new branch before starting work on a fix, rather than committing your changes to master. Otherwise, you'll find that any subsequent, unrelated changes you commit on master will get mixed in with the pull request as well.)

PR #2971, opened 2016-09-03
Make Visibility and Password translatable

lboaretto:
On PageViewRestrictionForm restriction_type and password fields current not translatable.


gasman:
Thanks! Merged in 29e7c0740bd40db4e3747bd5ffaab86aca11f1c4 + parents.

PR #2975, opened 2016-09-04
Fix #2789: Convert embed filter to templatetag

### LINKED ISSUE #2789: Convert embed filter into a template tag
Currently, to insert embeds into a template, you use the ``embed`` filter on the URL:

```
{{ my_embed_url|embed }}
```

This does a lot more than a filter should do, which are usually used for simple transformations (uppercase, escape HTML, etc) but fetching an embed is a much more complex operation.

So I think it makes more sense for it to be a template tag, similar to ``{% image %}``:

```
{% embed my_embed_url %}
```

jjanssen:
Fix for #2789 

I am not sure how we normally handle deprecation warnings in Wagtail. For convenience I've added for those who are already using the embed filter that it will be removed in 1.8 (maybe even 1.9?).

What are the next steps regarding deprecation of certain methods? 
Do we make a separate issue for cleanup?


kaedroho:
Thanks!

PR #2979, opened 2016-09-05
Page: get_title will get seo_title if exists

nutztherookie:
I know this solution is also not ideal in terms of a confusing function name. I'm happy for any recommendations.


gasman:
Hi @nutztherookie - I'm not so sure that this is a good idea. Whatever name we choose for this method (`html_title`?), its purpose is not going to be immediately obvious to someone reading the template code - they're going to have to look it up in the documentation.

The existing alternative is to write something like `{{ page.seo_title|default:page.title }}`, which makes the behaviour more explicit - and given that this will typically only appear once in a project (i.e. in the base template), it doesn't hurt to be more verbose.

nutztherookie:
okay, fair enough :)  (in my case I DID use it a few times but still it's okay)

PR #2982, opened 2016-09-06
Eliminate dependencies on __latest__ migrations

gasman:
See https://groups.google.com/d/msg/wagtail/a1lbdKe-QPk/GefsBFnRBgAJ

Using `__latest__` prevents us from ever applying migrations that are subsequently added to the referenced apps, since logically those migrations must have been applied before the current one (which they weren't, because they didn't exist). This logic is enforced as of Django 1.10.

### DIFF
--- a/docs/releases/1.7.rst
+++ b/docs/releases/1.7.rst
@@ -55,3 +55,22 @@ Bug fixes
 
 Upgrade considerations
 ======================
+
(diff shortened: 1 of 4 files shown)


gasman:
Merged in e31b26c4dbc3b851a562a70a2365b7187d3a0b33

PR #2987, opened 2016-09-08
Adds an Issue and a Pull request template to Github

jjanssen:
Hi! I think we can benefit from having a Github issue and pull request template. When browsing around some issues the first response mainly contains the question of which version of Wagtail or Django version someone is running. 

I have a put a issue template in place for this and also added a pull request template which refers to our contributing guidelines / running tests in the (latest) docs.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,19 @@
+Found a bug? Please fill out the sections below. 👍
+
+### Issue Summary
+
+A summary of the issue.
+
+### Steps to Reproduce
+
+1. This is the first step
+2. This is the second step, etc.
+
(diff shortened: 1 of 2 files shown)


gasman:
Thanks @jjanssen - I think this is a really good idea!

Will leave this open for a day or two in case people have further thoughts about what should go in these messages, but otherwise happy to merge.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,19 @@
...
+### Issue Summary
+
+A summary of the issue.
+
+### Steps to Reproduce
+
+1. This is the first step
+2. This is the second step, etc.
+
+Any other relevant information. For e.g. Why do you consider this a bug and what did you expect to happen instead?
```
tomdyson: "For e.g. Why do" -> "For example, why do" reads slightly better.

PR #2992, opened 2016-09-12
added support for elasticsearch mapper-attachments (v 1.7 and 2.x)

nutztherookie:
to full-text index PDFs and such, one can now just install
ElasticSearch's mapper-attachment (https://www.elastic.co/guide/en/elasticsearch/plugins/master/mapper-attachments.html)

Wagtail should use it automatically once it is installed.


nutztherookie:
I know this isn't pretty yet; I'm actually hoping for a discussion how to implement this nicely.

Also: mapper-attachments has a bunch of different keys that can in theory be filled [1](https://www.elastic.co/guide/en/elasticsearch/plugins/master/mapper-attachments-usage.html), like:
````json
"my_attachment" : {
        "_content_type" : "application/pdf",
        "_name" : "resource/name/of/my.pdf",
        "_language" : "en",
        "_content" : "... base64 encoded attachment ..."
    }
````

but I feel that that makes only sense once we have the basis implemented pretty.

**Comment by kaedroho** [comment]
Thanks @nutztherookie. This looks like it could be quite useful!

I think we can just assume the mapper-enhancements is installed, instead of having to check for it. If a user attempts to index a file into an ES instance without the plugin, we can raise an error instructing them what to do.

**Comment by nutztherookie** [comment]
if you think that that's smarter, sure :)  maybe we ought to backport it to ES1 then too. mapper-attachments is available there as well.

if the wagtail-community / torchbox is interested in such a feature and it has the chance to be upstreamt, i can look into it :)

File: wagtail/wagtailsearch/backends/elasticsearch.py [hunk 1]
```diff
@@ -26,7 +26,7 @@ class ElasticsearchMapping(object):
         'DateField': 'date',
         'DateTimeField': 'date',
         'DecimalField': 'double',
-        'FileField': 'string',
+        'FileField': 'attachment',
```
kaedroho: So there's now two ways we can index a ``FileField``. I believe at the moment we just index the filename.

We should probably keep the current behaviour as it is, and instead make the user explicitly define that they want the ``FileField`` indexed as an attachment by using the ``type``keyword argument on ``SearchField``.

For example:

``SearchField('myfile')`` will index the filename as a string (current behaviour)
``SearchField('myfile', type='attachment')`` will index the file contents using ``mapper-attachments``

nutztherookie: The `'FileField': 'string'` actually seems not to work anyways. I took torchbox/wagtail@master simply added 
````patch
    search_fields = CollectionMember.search_fields + [
        index.SearchField('title', partial_match=True, boost=10),
+        index.SearchField('file', partial_match=True, boost=10),
        index.RelatedFields('tags', [
            index.SearchField('name', partial_match=True, boost=10),
        ]),
````

and it's not working with `elasticsearch2` here anymore.
````
elasticsearch.exceptions.SerializationError: ({'uploaded_by_user_id_filter': 9, 'tags': [], 'title': 'test', 'content_type': ['wagtaildocs.Document', 'wagtaildocs.AbstractDocument'], 'collection_id_filter': 1, 'file': <FieldFile: documents/test2.pdf>, '_partials': [<FieldFile: documents/test2.pdf>, 'test'], 'pk': '8'}, TypeError("Unable to serialize <FieldFile: documents/test2.pdf> (type: <class 'django.db.models.fields.files.FieldFile'>)",))
````

Comment (nutztherookie) [inline wagtail/wagtailsearch/backends/elasticsearch.py]: I like your solution, but it will mean that people will have to create their own class Document 
````python
class MyDocument(AbstractDocument):
    admin_form_fields = (
        'title',
        'file',
        'collection',
        'tags'
    )
    search_fields = CollectionMember.search_fields + [
        index.SearchField('title', partial_match=True, boost=10),
        index.SearchField('file', search_content=True),
        index.RelatedFields('tags', [
            index.SearchField('name', partial_match=True, boost=10),
        ]),
        index.FilterField('uploaded_by_user'),
    ]
````
and overwrite `settings.WAGTAILDOCS_DOCUMENT_MODEL` accordingly.


would that be fine with everybody?

Comment (kaedroho) [inline wagtail/wagtailsearch/backends/elasticsearch.py]: That's fine with me. We should definitely make this opt-in as I imagine this would make indexing large PDFs take much longer (and may time out the upload request).

File: wagtail/wagtailsearch/index.py [hunk 2]
```diff
@@ -203,6 +203,8 @@ def get_value(self, obj):
         try:
             field = self.get_field(obj.__class__)
             value = field.value_from_object(obj)
+            if hasattr(self, 'search_content') and self.search_content:
```
kaedroho: I think we should make this the default behaviour of file field, instead of hiding it behind a flag. As there's no previous working behaviour and I think opening the file is what people would expect?

nutztherookie: So: going back to https://github.com/torchbox/wagtail/pull/2992/commits/96ddb64ffb9f5b5a26914c9ba909d594930ffb92#diff-7ba5f4d8cfc78dcc93e806ec45f75d4fR29 ?

nutztherookie:
woa. apparently i still don't exactly know now to manage/deal with github pull requests.

nutztherookie:
since i kinda wracked this pull request here, i moved to another patch-branch and ticket #3028

PR #2993, opened 2016-09-12
Fix flake8 complaint in group_models_by_index

eprikazc:
This change fixes flake8 message that is thrown while running "make lint":

F812 list comprehension redefines 'models' from line 13


kaedroho:
Thanks!

PR #2996, opened 2016-09-13
Uses specific_class.url instead of Page.url on Rich Text

lboaretto:
I was overriding the url method for my Article/Page subclass. "View online" works fine but not on Rich Text link handler.


gasman:
Thanks! I think this can be simplified to: `page = Page.objects.get(id=attrs['id']).specific`.

Could you add a test for this, please?

lboaretto:
Ok! I'll write tests and place this week.

Thanks.

gasman:
Completed in #3494

PR #2999, opened 2016-09-15
Add hints for technical details in issue template

timheap:
The more information we get, the easier it is for users to provide it, and the more accurate that information is, the better!


gasman:
Thanks!

PR #3004, opened 2016-09-15
[WIP] Elasticsearch highlighting

m1kola:
A new method to Elasticsearch 1/2 backends that allows to highlight search results.

Usage example:

``` python
search_query = 'Wagtail'
search_results = Page.objects.live().search(search_query)
search_results = search_results.highlight(
    fields={
        'title': {},
        'body': {},
    },
    require_field_match=False
)
```

How to access highlighted fields:

``` python
search_result.title_highlight
search_result.body_highlight
```

Notes:
1. Both `search_result.title_highlight` and `search_result.body_highlight` may be `None`.
2. Highlighted fields may contain invalid html. It seems to me that implementor should decide how to work with these fields. For example, he can remove invalid tags like this:
   
   ``` python
   for search_result in search_results:
       if search_result.title_highlight:
           search_result.title_highlight = str(BeautifulSoup(search_result.title_highlight, 'html.parser'))
   
       if search_result.body_highlight:
           search_result.body_highlight = str(BeautifulSoup(search_result.body_highlight, 'html.parser'))
   ```

---

PR needs tests. I'll add them, when we are happy with the approach.

Topics for discussion:
1. In the current implementation (24ad275) it's possible to query on `Page`'s `QuerySet` and highlight fields for `EventPage`, for example. Do we need this feature?
2. Would be great to be able to specify a name of an attribute where to set highlighted text. But the current implementation (24ad275) we can pass fields as pure dict which gives us ability to use any [Elasticsearch parameters for highlighting](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-highlighting.html). If we want to set target attribute, we probably need another representation of fields for `highlight` method.

### DIFF
--- a/wagtail/wagtailsearch/backends/elasticsearch.py
+++ b/wagtail/wagtailsearch/backends/elasticsearch.py
@@ -475,8 +475,22 @@ class ElasticsearchSearchResults(BaseSearchResults):
         # Send to Elasticsearch
         hits = self.backend.es.search(**params)
 
-        # Get pks from results
-        pks = [hit['fields']['pk'][0] for hit in hits['hits']['hits']]
+        pks = []
+        highlight = {}
+        for hit in hits['hits']['hits']:
+            # Get pks from results
+            pk = hit['fields']['pk'][0]
+            pks.append(pk)
+
+            # Get highlight
+            for field_def in self.highlight_params['fields']:
+                field_name = field_def.keys()[0]
+
+                field_highlight = highlight.get(pk, {})
+                field_highlight.update({
+                    field_name: hit.get('highlight', {}).get(field_name, [None])[0]
+                })
+                highlight[pk] = field_highlight
 
         # Initialise results dictionary
         results = dict((str(pk), None) for pk in pks)
@@ -484,7 +498,13 @@ class ElasticsearchSearchResults(BaseSearchResults):
         # Find objects in database and add them to dict
         queryset = self.query.queryset.filter(pk__in=pks)
         for obj in queryset:
-            results[str(obj.pk)] = obj
+            str_pk = str(obj.pk)
+            obj_highlight = highlight.get(str_pk)
+            if obj_highlight:
+                for field_name, field_highlight in obj_highlight.items():
+                    setattr(obj, field_name + '_highlight', field_highlight)
+
+            results[str_pk] = obj
 
         # Return results in order given by Elasticsearch
         return [results[str(pk)] for pk in pks if results[str(pk)]]
--- a/wagtail/wagtailsearch/backends/elasticsearch.py
+++ b/wagtail/wagtailsearch/backends/elasticsearch.py
@@ -458,7 +458,7 @@ class ElasticsearchSearchResults(BaseSearchResults):
         return body
 
     def _get_content_type(self, content_type):
-        app_label, model = content_type.rsplit('_', 2)[1:]
+        app_label, model = content_type.rsplit('_', 2)[-2:]
 
         return ContentType.objects.get_by_natural_key(app_label, model)
 
--- a/wagtail/wagtailsearch/backends/elasticsearch.py
+++ b/wagtail/wagtailsearch/backends/elasticsearch.py
@@ -444,7 +444,7 @@ class ElasticsearchSearchResults(BaseSearchResults):
             if sort is not None:
                 body['sort'] = sort
 
-            if self.highlight_params['fields']:
+            if self.highlight_params.get('fields'):
                 highlight = {
                     'fields': self.highlight_params['fields'].values(),
                 }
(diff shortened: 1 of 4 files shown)


wagtail/wagtailsearch/backends/elasticsearch.py
```diff
@@ -398,6 +399,38 @@ def __repr__(self):
 
 
 class ElasticsearchSearchResults(BaseSearchResults):
+    def __init__(self, *args, **kwargs):
+        super(ElasticsearchSearchResults, self).__init__(*args, **kwargs)
+        self.highlight_params = {}
+
+    def highlight(self, **kwargs):
```
kaedroho: Have you considered an API, where each call to ``.highlight()`` adds a single field?

For example:

```python
Page.objects.search("foo").highlight("body", param="value").highlight("title")
```

This would give us some runtime type checking from Python.

kaedroho: Also, I noticed that the parameters passed to highlight are all just sent to Elasticsearch. To help us support highlighting for more backends in the future, maybe we should do something similar to the ``SearchFields`` API and support the common parameters in the API but allow customisation by passing a ``dict`` in a kwarg called ``es_extra``.

wagtail/wagtailsearch/backends/elasticsearch.py
```diff
@@ -398,6 +399,38 @@ def __repr__(self):
...
+                # Use wildcard column name on request. This allows us to query against all content types.
+                # Alternative solution: get all indexed models and add all possible column names into request.
+                field_column_name = '*{}'.format(field_name)
+                post_processed_field_def = {
+                    field_column_name: field_def[field_name]
+                }
+                post_processed_fields[field_name] = post_processed_field_def
+
+        self.highlight_params['fields'] = post_processed_fields
+        self.highlight_params['require_field_match'] = kwargs.get('require_field_match', None)
```
kaedroho: How would ``require_field_match`` affect the results? Does it need to be configurable?

wagtail/wagtailsearch/backends/elasticsearch.py
```diff
@@ -409,15 +442,33 @@ def _get_es_body(self, for_count=False):
             if sort is not None:
                 body['sort'] = sort
 
+            # Add highlighting into a body, if fields are specified
+            if self.highlight_params.get('fields'):
+                highlight = {
+                    'fields': self.highlight_params['fields'].values(),
```
kaedroho: At some point in here we need to translate the field names passed in from the user to the actual column names.

For example: ``body`` may be indexed as ``events_eventpage_body`` and ``home_homepage_body``

m1kola: We don't need to do that because we translate `body` into wildcard `*body` [here](https://github.com/torchbox/wagtail/pull/3004/files#diff-7ba5f4d8cfc78dcc93e806ec45f75d4fR418). By doing that we avoid getting content types for each object from a database. It's important if we query using `Page`'s `QuerySet`.

In a result set, we will have a hit which will have the `events_eventpage_body` column instead of `body`, so we need to translate a column name back into a field name [here](https://github.com/torchbox/wagtail/pull/3004/files#diff-7ba5f4d8cfc78dcc93e806ec45f75d4fR515). Which also doesn't need query to a database, because we know a content type from the hit (see [this line](https://github.com/torchbox/wagtail/pull/3004/files#diff-7ba5f4d8cfc78dcc93e806ec45f75d4fR491)).

wagtail/wagtailsearch/backends/elasticsearch.py
kaedroho: Thanks for the expanation! If I understood correctly, it sounds like it always needs to be ``false`` in order for highlighting to work as expected, so does it need to be configurable?

m1kola: Good question! It seems to me that in some use cases an implementor may want to highlight only matched fields, but I'm not sure. Maybe it's not so important.

BertrandBordage:
It would be great to have a Python fallback in `BaseSearchResults`, so it can work on all backends, and a more accurate & optimised version could be written for each backend later.

This feature would be easy to implement in `wagtail.contrib.postgres_search`, but it may be slower than making it in Python, surprisingly. I already wrote [something similar in django-postgrefts](https://github.com/BertrandBordage/django-postgrefts/blob/master/postgrefts/models.py#L203-L222), it makes a beautiful and very accurate highlight, but it’s a bit slower than I expected.

SalahAdDin:
Does not highlighting featured for last ElasticSearch version?

miltontony:
@m1kola what's the latest on this PR? With ES5 backend now available and ES6 recently released, this has fallen rather behind. When can we expect this PR to be complete and landed?

SalahAdDin:
What's about ElasticSearch 6?

SalahAdDin:
Guys? What's about that?

m1kola:
Hey. I'm not working on this, currently. Happy for someone else to pick it up.

SalahAdDin:
@kaedroho now that you are implementing `facets`, maybe it is time for implementing this feature.

m1kola:
Closing this as it's very outdated

Issue #3005, opened 2016-09-15
Wagtail does not validate passwords


BertrandBordage:
In Django 1.9, [a nice system of password validation](https://docs.djangoproject.com/en/1.10/topics/auth/passwords/#module-django.contrib.auth.password_validation) was at last added.

Unfortunately, the password strength is not checked in the `user.set_password` method, maybe for backwards-compatibility reasons. This means we need to manually call [`django.contrib.auth.password_validation.validate_password`](https://docs.djangoproject.com/en/1.10/topics/auth/passwords/#django.contrib.auth.password_validation.validate_password) each time we want to change a password.

Currently, Wagtail is not validating passwords in its back office. A call to `validate_password` should be added to `wagtail.wagtailusers.forms.UserCreationForm` and `UserEditForm`.

Since the current version of Wagtail – 1.6.x – also supports Django 1.8, this will either require to create a small compatibility code, or to wait for a release dropping Django 1.8 support. IMHO I think we should do it now, because it’s misleading: when you define `AUTH_PASSWORD_VALIDATORS` you don’t even think it will not be respected by Wagtail and you generally assume such settings will work without testing manually.

I can make a pull request.

gasman:
This sounds like a good idea to me. I agree that we should make the change now - we have `if django.VERSION >= (1, 9)` checks elsewhere in the code, so there's no harm in adding another.

BertrandBordage:
OK, I’m preparing a pull request.

gasman:
Implemented in #3007

PR #3007, opened 2016-09-16
Adds password validation

BertrandBordage:
As discussed in #3005, this pull request adds password validation to the user creation and edit forms.

I also took the liberty to factorize the two forms, since they were mostly copies of each other. The only difference is that I marked the password fields as required in the creation form, and I removed the “Leave blank if not changing.” on that same form. It was possible to create a user account without a password and I removed this possibility, but it seemed to be a bug, because the message “Leave blank if not changing.” was not consistent with a creation form…

Unfortunately, there is a small bug remaining concerning the password change form: when `AUTH_PASSWORD_VALIDATORS` is set,  [Django sets the help text with HTML](https://github.com/django/django/blob/4b9330ccc04575f9e5126529ec355a450d12e77c/django/contrib/auth/forms.py#L82) but doesn’t mark it as safe, so we see HTML tags.

### DIFF
--- a/wagtail/wagtailusers/forms.py
+++ b/wagtail/wagtailusers/forms.py
@@ -2,7 +2,7 @@ from __future__ import absolute_import, unicode_literals
 
 from itertools import groupby
 
-from django import VERSION as django_version
+import django
 from django import forms
 from django.conf import settings
 from django.contrib.auth import get_user_model
@@ -20,8 +20,7 @@ from wagtail.wagtailcore.models import (
 from wagtail.wagtailusers.models import UserProfile
 
 
-DJANGO_GTE_1_9 = django_version >= (1, 9)
-if DJANGO_GTE_1_9:
+if django.VERSION >= (1, 9):
     from django.contrib.auth.password_validation import (
         password_validators_help_text_html, validate_password
     )
@@ -97,7 +96,7 @@ class UserForm(UsernameForm):
         if self.password_required:
             self.fields['password1'].help_text = (
                 mark_safe(password_validators_help_text_html())
-                if DJANGO_GTE_1_9 else '')
+                if django.VERSION >= (1, 9) else '')
             self.fields['password1'].required = True
             self.fields['password2'].required = True
 
@@ -124,7 +123,7 @@ class UserForm(UsernameForm):
 
     def clean_password1(self):
         password = self.cleaned_data.get('password1')
-        if DJANGO_GTE_1_9 and password:
+        if django.VERSION >= (1, 9) and password:
             validate_password(password, user=self.instance)
         return password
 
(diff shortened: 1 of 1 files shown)


wagtail/wagtailusers/forms.py
```diff
@@ -17,6 +19,13 @@
     UserPagePermissionsProxy)
 from wagtail.wagtailusers.models import UserProfile
 
+
+DJANGO_GTE_1_9 = django_version >= (1, 9)
```
kaedroho: The usually write these version checks like ``django.VERSION >= (1, 9)``. If we follow this convention, it will be easier to search for when we remove old Django versions.

wagtail/wagtailusers/forms.py
```diff
@@ -54,68 +63,93 @@ def separate_username_field(self):
...
 
-class UserCreationForm(UsernameForm):
+class UserForm(UsernameForm):
     required_css_class = "required"
 
+    password_required = True
+
     error_messages = {
-        'duplicate_username': _("A user with that username already exists."),
+        'duplicate_username': _('A user with that username already exists.'),
```
kaedroho: What's the reason for changing all the quotes?

wagtail/wagtailusers/forms.py
BertrandBordage: OK, I will change that in a new commit. (Sorry btw, I forgot `[:2]`, which is important since `(1, 10, 'alpha') > (1, 10, 1)` is `True`.)
For the record, I usually do this to avoid doing the attribute lookup, tuple creation and comparison over and over again, and to reduce the risk of typos. It’s not much more complex to find these again after, using PyCharm.

wagtail/wagtailusers/forms.py
BertrandBordage: The reason is simply that I’m a single quote maniac. A maniac believing all quotes could be single, and all single quotes inside strings could be replaced with the character that was actually designed for this purpose: the apostrophe `’`. (The only exception being to quote a programming string in a string, and this would reasonable to escape them, like `'Here’s the code: \'Cool\''`)
But sorry about all that, I will revert these changes.

BertrandBordage:
@kaedroho I did the changes following your review. I followed the Python guidelines of Wagtail for the `django.VERSION` comparison.

mevinbabuc:
Does this work with Django 1.8 ?

BertrandBordage:
@mevinbabuc Yes!

gasman:
Thanks @BertrandBordage, and sorry for the delay in reviewing! Merged in d4114ff4ae4897acc0f892d66104f5754f5b8dfd, with a small tweak (validation now happens in `clean_password2` to prevent a bogus 'passwords don't match' message, which matches [what Django does](https://github.com/django/django/blob/eb97af0402536a86d5d8c916859986288bb88dc9/django/contrib/auth/forms.py#L398)). Have opened #3398 to address the rendering issue on the 'change password' form.

BertrandBordage:
No problem for the delay @gasman!
Thank you for the great release management job :)

PR #3011, opened 2016-09-19
Removing eval statement from modal-workflow.js

sebworks:
## Changes
 - Modied `wagtail/wagtailadmin/static_src/wagtailadmin/js/modal-workflow.js` to use `JSON.parse` instead of `eval`.  Wrapped `JSON.parse` statement in a `try / catch` block to handle scenarios where the server returns HTML instead of JSON (session timeouts).
 

 - Modified `wagtail/wagtailadmin/modal_workflow.py` to return valid JSON. JSON keys can only be  wrapped in double quotes per http://www.json.org/.  Also added code to remove newlines, which were affecting parsing of the JSON string using `JSON.parse`.

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/modal-workflow.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/modal-workflow.js
@@ -51,9 +51,15 @@ function ModalWorkflow(opts) {
     };
 
     self.loadResponseText = function(responseText) {
-        var response = eval('(' + responseText + ')');
+        var response;
 
-        self.loadBody(response);
+        try {
+            response = JSON.parse( responseText );
+        } catch ( error ) {
+            window.location = window.location.host;
+        }
+
+        self.loadBody( response );
     };
 
     self.loadBody = function(response) {
@@ -63,9 +69,15 @@ function ModalWorkflow(opts) {
             container.modal('show');
         }
 
+        // if the response contains an `onload` function, call the function with the current
+        // context as a parameter.
         if (response.onload) {
-            // if the response is a function
-            response.onload(self);
+            var onLoadFunction = new Function('', 'return ' + response.onload)();
+            if (typeof onLoadFunction === 'function') {
+                onLoadFunction(self);
+            } else {
+                throw new Error('ModalWorkflow.loadBody: Error evaluating onload function');
+            }
         }
     };
 
--- a/wagtail/wagtailadmin/modal_workflow.py
+++ b/wagtail/wagtailadmin/modal_workflow.py
@@ -19,6 +19,6 @@ def render_modal_workflow(request, html_template, js_template, template_vars=Non
 
     if js_template:
         js = render_to_string(js_template, template_vars or {}, request=request)
-        response_keyvars.update({'onload': js.replace('\n','')})
+        response_keyvars.update({'onload': js.replace('\n', '')})
 
-    return HttpResponse(json.dumps(response_keyvars) , content_type="text/javascript")
+    return HttpResponse(json.dumps(response_keyvars), content_type="text/javascript")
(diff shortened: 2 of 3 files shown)


wagtail/wagtailadmin/static_src/wagtailadmin/js/modal-workflow.js
```diff
@@ -51,9 +51,15 @@ function ModalWorkflow(opts) {
...
 
     self.loadResponseText = function(responseText) {
-        var response = eval('(' + responseText + ')');
+        var response;
 
-        self.loadBody(response);
+        try {
+            response = JSON.parse( responseText );
+        } catch ( error ) {
+            window.location = window.location.host;
```
gasman: This seems a bit heavy-handed. Currently, if the modal returns a non-Javascript response, the worst that will happen is that they get a Javascript error or a modal containing a mangled login form or error page - any content they've entered in the 'outer' page will still be salvageable. With this change, they'll be redirected to the site homepage with no warning or explanation, losing their work.

gasman:
Hi @sebworks - thanks for looking into this! The modal_workflow code is a bit of a murky and unloved part of the Wagtail codebase, so any work towards refactoring it into something nicer is much appreciated :-)

I'm not sure that turning the response into JSON (and thus turning the 'onload' field into a string) is the right move at this point, though. Ultimately we would like to eliminate the practice of sending executable JS down the wire - and in that respect, changing the response from a blob of arbitrary Javascript into JSON data is a good thing. However, if that JSON data still has Javascript code embedded into it as a string, then we've not really achieved anything, besides adding an extra layer of escaping for the browser to unwrap.

Perhaps you're intending this to be a first step towards further refactoring, that will eliminate the 'onload' field? If so, that's fair enough... I'd be keen to hear your planned approach before I go ahead and merge this change.

wagtail/wagtailadmin/static_src/wagtailadmin/js/modal-workflow.js
sebworks: ok, how about an alert box or just throwing an error?

sebworks:
@gasman, you are correct. I'm going to close this and open another PR which refactors the modal workflow code.  The purpose of this PR was to solve the issues of session timeout without notification and the use of eval statement.

wagtail/wagtailadmin/static_src/wagtailadmin/js/modal-workflow.js
sebworks: if the session has timed out, how would they be able to save their work without logging in again?

gasman: That's a fair point. If the user knows their session has timed out before submitting the form, they can log in in another tab, or copy their work into a text document... neither of which is particularly desirable. There's certainly a lot we could do to improve the user experience there, but that's out of scope for this issue. The important thing here is not to make the current experience worse (which would happen here, if opening a page chooser model caused their entire work to be lost).

I think the ideal response to opening a modal while logged out would be to display the login form within the modal, but that's probably a deep rabbit-hole to go down (to do it properly, you'd need to embed the 'forgotten password' workflow in the modal too, and handle non-standard login mechanisms like OAuth...). Failing that, a message in the modal saying "your session has timed out, please log in in another tab" would probably be the next best thing. (If we're going this route, it would be nice to have the server return a specific response that indicates 'session timeout', rather than just assuming that any non-JSON response is a session timeout - not essential, though.)

sebworks: ok, I'm going to go  the `"your session has timed out, please log in in another tab" ` route. Thanks for the feedback. I will also work on removing `eval` and `new Function`, as both are insecure and also pollute the global namespace.

PR #3014, opened 2016-09-20
Add to StreamField migrations documentation

nimasmi:
- Add example code for migrating draft pages and page revisions.

Recently I amended the provided example migration to include migrating draft pages, as the site in question had a lot of not-yet-published content. This migration has more features, but feels more hacky, so I'm open to comments. In particular:
1. I'm not sure to what degree this needs health warnings about incompatible non-migrated older page revisions.
2. Being more convoluted, and having a narrower audience, I'm not sure whether this example should be provided instead of, or in addition to, the existing example code.

### DIFF
--- a/docs/topics/streamfield.rst
+++ b/docs/topics/streamfield.rst
@@ -849,3 +849,116 @@ If you change an existing RichTextField to a StreamField, and create and run mig
                 convert_to_richtext,
             ),
         ]
+
+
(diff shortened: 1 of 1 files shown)


gasman:
Merged in 48949e69a7a7039cb751f1ee33ecb32187004030 - thanks @nimasmi!

PR #3016, opened 2016-09-21
Allow to limit access to form submissions

m1kola:
This PR adds the `construct_forms_for_user` hook which allows us to limit access to form submissions.

I'll add tests and docs, if we decide that this is acceptable.

---

If this PR merged, we will be able to do something like

```python
def register_permissions():
    wagtailadmin_content_type = ContentType.objects.get(app_label='wagtailadmin', model='admin')

    Permission.objects.get_or_create(
        content_type=wagtailadmin_content_type,
        codename='access_form_submissions',
        name='Can access form submissions'
    )

    return Permission.objects.filter(content_type=wagtailadmin_content_type, codename='access_form_submissions')


@hooks.register('construct_forms_for_user')
def construct_forms_for_user(user, queryset):
    if not user.has_perm('wagtailadmin.access_form_submissions'):
        queryset = queryset.none()

    return queryset
```

or even simpler:

```python
@hooks.register('construct_forms_for_user')
def construct_forms_for_user(user, queryset):
    if not user.is_superuser:
        queryset = queryset.none()

    return queryset
```

### DIFF
--- a/docs/reference/hooks.rst
+++ b/docs/reference/hooks.rst
@@ -223,9 +223,15 @@ Hooks for building new areas of the admin interface (alongside pages, images, do
 ``filter_form_submissions_for_user``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-  Allows to limit access to form submissions. Doesn't affect page editing.
+  Allows access to form submissions to be customised on a per-user, per-form basis.
 
-  Return a queryset of ``Page`` objects to be shown in the Forms administration area.
+  This hook takes two parameters:
+   - The user attempting to access form submissions
+   - A ``QuerySet`` of form pages
+
+  The hook must return a ``QuerySet`` containing a subset of these form pages which the user is allowed to access the submissions for.
+
+  For example, to prevent non-superusers from accessing form submissions:
 
   .. code-block:: python
 
--- a/docs/reference/hooks.rst
+++ b/docs/reference/hooks.rst
@@ -215,7 +215,30 @@ Hooks for building new areas of the admin interface (alongside pages, images, do
 ``register_permissions``
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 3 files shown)


m1kola:
@kaedroho thanks for feedback. I agree with you. Does `filter_form_submissions_for_user` sound better? Feel free to propose a name ;)

kaedroho:
Yep, I think that's much better. Thanks!

m1kola:
* Rebased on top of `master`
* The hook has been renamed to `filter_form_submissions_for_user `
* Documentation has been added
* Tests have been added

@kaedroho, ready for review! ;)

docs/reference/hooks.rst
```diff
@@ -215,7 +215,30 @@ Hooks for building new areas of the admin interface (alongside pages, images, do
...
-  Return a queryset of Permission objects to be shown in the Groups administration area.
+  Return a queryset of ``Permission`` objects to be shown in the Groups administration area.
+
+
+.. _filter_form_submissions_for_user:
+
+``filter_form_submissions_for_user``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+  Allows to limit access to form submissions. Doesn't affect page editing.
```
kaedroho: Just expanded this a little:

```
Allows access to form submissions to be customised on a per-user, per-form basis.

This hook takes two parameters:
 - The user attempting to access form submissions
 - A ``QuerySet`` of form pages

The hook must return a ``QuerySet`` containing a subset of these form pages which the user is allowed to access the submissions for.

For example, to prevent non-superusers from accessing form submissions:
```

m1kola:
Documentation has been updated in 2641a43

Thanks for the review and for the documentation tweaks! I'll merge this PR, when tests are green.

m1kola:
Merged in 48296e756858850042ffaabb0bf023cf6f144c49 + parent.

Thanks!

PR #3017, opened 2016-09-21
Stop Django >=1.10.1 from ignoring form submissions of StreamFields with default values

### LINKED ISSUE #2994: StreamField not saving after update to 1.6.2
After updating to 1.6.2, StreamFields are no longer saving.

There are no errors in the INFO log, and everything looks normal - the green message is displayed as always, and a new revision is created with all the updated fields. Except the StreamField.

gasman:
Fixes #2994

The story in brief: [a change to the ModelForm logic in Django 1.10](https://groups.google.com/forum/#!msg/django-developers/lpfM6QPILoQ/9mOXrtRWCwAJ) meant that, as of Django 1.10.1, there's an explicit "is this field present in the form submission" check. (If it isn't present, and the model field has a default, it falls back on the default value.)

Django 1.10.1 does this in a fairly crude way, by looking for an item in the postdata matching the field name, with a special case for checkbox widgets (which set the flag `dont_use_model_field_default_for_empty_data` to indicate that they might be legitimately absent from the form submission). This [broke various widgets that map themselves to HTML form fields in non-trivial ways](https://code.djangoproject.com/ticket/27186), including StreamField's BlockWidget.

https://github.com/django/django/pull/7217 (due for Django 1.10.2, but not yet merged into stable/1.10.x at the time of writing) fixes this in a more robust way by having widgets provide a `value_omitted_from_data` method to report whether the field is considered missing.

This PR thus provides two separate fixes - for both the 1.10.1 and forthcoming 1.10.2 implementations - to persuade ModelForm that the StreamField really is present on the form despite not having a field in postdata that exactly matches its name.


wagtail/wagtailcore/tests/test_blocks.py
```diff
@@ -26,6 +26,11 @@
 from wagtail.wagtailcore.rich_text import RichText
 
 
+# Skip the tests for Block.value_omitted_from_data on versions of Django that do not
+# implement Widget.value_omitted_from_data (scheduled to be added in Django 1.10.2)
+SKIP_VALUE_OMITTED_FROM_DATA_TESTS = not hasattr(forms.Widget, 'value_omitted_from_data')
```
kaedroho: ``django.VERSION >= (1, 10, 2)`` would make this check more greppable, which would be helpful when we remove Django 1.10 support.

wagtail/tests/testapp/models.py
```diff
@@ -556,6 +556,21 @@ class StreamPage(Page):
...
 
 
+class DefaultStreamPage(Page):
+    body = StreamField([
+        ('text', CharBlock()),
+        ('rich_text', RichTextBlock()),
+        ('image', ImageChooserBlock()),
+    ], default='')
+
+    api_fields = ('body',)
```
kaedroho: ``api_fields`` shouldn't be needed

gasman: removed in 854e21de09fa0e910fcb4e48c3bf4912fbca5b36

wagtail/wagtailcore/tests/test_blocks.py
gasman: Updated in 757624609b2ac219318a988a71dffeff24d94ea0

gasman:
Merged in 869349f86c9eea55ab277d237fd136b6a5721d00 (master) / 8e67d2d56aab3cf8e4aada575fab602314d4a6dc (1.6.x)

PR #3019, opened 2016-09-22
Use better JS source maps during development

thibaudcolas:
The current webpack source map configuration isn't very useful when debugging transpiled/bundled JS code.

Before – the DevTools do not mention file name, and point to an empty file:

cheap-module-eval-source-map

After – the DevTools have the correct file name, and point to the right file content (original code before transformations):

inline-source-map

There is a development build speed cost to using those "proper" source maps – on my modern MacBook Pro, initial build is ±25% slower (12s -> 15s) and rebuild is ±2x slower (1s -> 2s). I think this is a reasonable trade-off for valuable dev logging.

See http://webpack.github.io/docs/configuration.html#devtool for all of the options (crazy how complicated these things can be!).

### DIFF
--- a/webpack.dev.config.js
+++ b/webpack.dev.config.js
@@ -4,6 +4,7 @@ var config = base('development');
 
 // development overrides go here
 config.watch = true;
-config.devtool = 'cheap-module-eval-source-map';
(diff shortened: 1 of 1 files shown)


kaedroho:
Thanks!

Issue #2960, opened 2016-08-29
Add select multiple for dropdown form field


dwasyl:
It would be very helpful to have either the ability to toggle the drop down list into a multi-select, or another form element representing multi-select.

Some forms I'm trying to create need multi-selects (and have too many choices to be represented as checkboxes). I ended up doing this manually, but it would be a helpful addition.

timmysmalls:
@dwasyl I agree. I know there's a component out there in a different repository, but why not just use the existing chooserpanels? What if you want to add several snippets/pages/images or some such in an association table? You already have a great interface to deal with this. For now it just works exclusively with one to many and one to one relationships. I think this is related https://github.com/wagtail/wagtail/issues/2203. It discusses a workaround of placing a chooserpanel in an inlinepanel. Imo the multi select widgets are not scalable enough out of the box to work with large sets of options. The Chooser Panels, however, are very useful for this. So I'd like to suggest allowing users to take advantage of the chooser panels in many to many relations like they can in regular foreign key relationships. Especially seen as how the bulk of the work (the chooser panels themselves) are already present in the CMS.

gasman:
@timmysmalls If I'm reading correctly, this issue is about forms created with the wagtail.contrib.forms form builder, not forms within the Wagtail admin.

zerolab:
I believe #3020 solved it for forms, so it could be closed?

gasman:
@zerolab Ah, good spot!

PR #3020, opened 2016-09-22
Added multiselect form field

dwasyl:
This is just a simple PR to add an additional form field in response to issue #2960 (that I opened). Only adds a few lines of code and the tests seem to run fine.


dwasyl:
Whoops - didn't even think about the migration file. Forgot Django even generated them for changes to choices. Let me see if I can generate a quick one.

dwasyl:
Spent some time looking at this and the change is to the definition of the `choices` field on the field of an Abstract model so Django won't generate a migration for it. It's needed to run the checks though, so not sure what I can do to get it to pass.

gasman:
Hi @dwasyl - the missing migration will probably be on one of the models in the test apps. You should be able to generate it by running `django-admin.py makemigrations --settings=wagtail.tests.settings` from the Wagtail root.

dwasyl:
Thanks @gasman, that did the trick. First time doing this, seemed all too easy at first.

gasman:
Sorry for the delay in reviewing - now merged in 240fa9153b86e641497a521ee135b28dfcf1a15b. Thanks @dwasyl!

dwasyl:
Fantastic, thanks for getting this in @gasman

PR #3021, opened 2016-09-22
Webpack build optimisations

thibaudcolas:
Here are "free" improvements to our webpack configuration.
- For development and production builds, `process.env.NODE_ENV` needs to be set to the right value to enable/disable JS development aids.

I don't think we use any directly at the moment, but more and more third-party libraries do (like React), and expect this flag to be set appropriately when bundling the code. The idea is that the development aids are active only in development, and get removed from the code when building for production.
- For production builds, use webpacks' UglifyJS plugin to minify the bundles. The configuration options are taken from https://github.com/facebookincubator/create-react-app/blob/master/packages/react-scripts/config/webpack.config.prod.js, which is part of a Facebook-supported preconfigured build tool for React.

Further notes: The convention is to call the webpack configuraiton files `webpack.config.base`, `webpack.config.dev`, `webpack.config.prod`, and it is OK to have them in a subfolder (here `client` would be best) instead of at the project root. Should I make those changes as well?


thibaudcolas:
NB: having this configured correctly reduces the production build size from 1.7MB to 145kB.

jjanssen:
I am only wondering about the explicit `config.watch = false` in the production config. I thought was `false` by default or is it a matter of naming it to create a difference between the two files? Other than that this looks good to me. I've also been using the production uglify for a while now.

For a future matter it would make sense to move these files to the client directory. Keeps the front-end tooling nice and consolidated without cluttering the root.

thibaudcolas:
@jjanssen good point for the `config.watch` – I copied this from some other project without thinking. Not entirely sure why it was there in the first place. I just took it out.

For the location of those files, completely agree about not cluttering the root and keeping this contained. I can add it to this PR if we have a consensus.

tomdyson:
> Further notes: The convention is to call the webpack configuraiton files webpack.config.base, webpack.config.dev, webpack.config.prod, and it is OK to have them in a subfolder (here client would be best) instead of at the project root. Should I make those changes as well?

Yes please! I'm really keen to clear out our project root. Would it make sense to do that as a separate PR?

jjanssen:
It would be convenient to have the current changes as a part of that. So i'm guessing it would make more sense to update the PR (or merge this change and start a new PR). 

Happy to help you with this @thibaudcolas if you are short on time.

thibaudcolas:
Thanks Janneke – I'm drowning under work at the moment. Do you have the rights to edit this PR, or should I give you access to `springload/wagtail`?

jjanssen:
It looks like I have some rights. Still thinking if we should put them in a webpack directory within the client directory. What are your thoughts on that?

jjanssen:
If @thibaudcolas agrees on the changes we are good to go on this PR. Aside from the code changes I have reviewed it against the 1.7 static output and it looks good.

thibaudcolas:
Thanks!

👍  This is good to go. I have rebased on the latest `master`, and made a couple more changes to simplify the config, but nothing that impacts the build output (could you have a look @jjanssen?).

I have a couple of ["committing code"](http://docs.wagtail.io/en/latest/contributing/committing.html) ceremony questions since this is the first PR that I could actually merge myself.

- Janneke made some changes, asked for review – I reviewed, but also made some more changes. I assume I should now ask for a review of my new changes, however small they might be? And we rince and repeat until there are no more changes?
- I do not want all my small commits to directly end up on master. Should I rebase/squash them even before the PR gets merged?
- I assume that when this gets merged, the commits will be squashed to some extent. What's the policy here, do you systematically squash all PRs to a single commit, or to a few commits?
- Here I used a branch on `springload/wagtail` to make this PR (because this was all I had access to at the time). Is there a policy to always use branches on forks to create PRs, over branches on `wagtail/wagtail`? If not, is there a specific branch naming scheme to follow?
- Is there a policy on rollbacks/reverts? If I suddenly realise that those changes have some negative impact, should I attempt to revert the PR on `master` directly, or make the revert as a new (reviewable) PR?

jjanssen:
Looks 👍 to me!

gasman:
Hi @thibaudcolas!

> Janneke made some changes, asked for review – I reviewed, but also made some more changes. I assume I should now ask for a review of my new changes, however small they might be? And we rince and repeat until there are no more changes?

In general, yes... in practice, I wouldn't take the "every code change has to be reviewed" rule to a religious extreme. :-) There's a threshold at which it's reasonable to commit changes directly without review - I'll leave it to your judgment where that threshold is, but as a rough guideline, if a change is uncontroversial, and small enough that you're completely confident that you're not introducing any new bugs, it's OK to commit immediately. The rename of prd.config.js to prod.config.js would be a good example of that.

> I do not want all my small commits to directly end up on master. Should I rebase/squash them even before the PR gets merged?

That's up to you. Rebasing / squashing is welcome*, but not essential. (*with a couple of fairly obvious exceptions: don't squash so much that it becomes impossible to follow the sequence of changes; and don't squash new commits that need reviewing, into old commits that have already been reviewed. If you're fixing a mistake in the earlier commit, a nice approach is to use the `git commit --fixup` option, so that the person doing the final merge can run `git rebase -i --autosquash`.)

> I assume that when this gets merged, the commits will be squashed to some extent. What's the policy here, do you systematically squash all PRs to a single commit, or to a few commits?

Ideally: a few commits, so that each commit is making a single meaningful change (and not breaking anything). Sometimes it's not possible to do that without a massive amount of backtracking (e.g. you implement a feature over several commits, and then decide to change a variable name) - if so, don't worry too much, and just go with whatever's the least ugly...

> Here I used a branch on springload/wagtail to make this PR (because this was all I had access to at the time). Is there a policy to always use branches on forks to create PRs, over branches on wagtail/wagtail? If not, is there a specific branch naming scheme to follow?

I'd suggest always using a fork, simply because - in some small way - it helps to promote the idea that core developers are contributing on the same terms that an outside contributor would, and that their PRs don't receive any "special status" through being on the wagtail/wagtail repo. I think that's a worthwhile message to send. (And besides, it stops the 'stable' branches from getting buried in the dropdown on the Github page :-) )

> Is there a policy on rollbacks/reverts? If I suddenly realise that those changes have some negative impact, should I attempt to revert the PR on master directly, or make the revert as a new (reviewable) PR?

I'd tend to create it as a PR, but merge it straight away without waiting for a review - see #2880 for example. That way, the PR serves as additional documentation for the change, and it gets run through the CI tests.

> Edit: extra question – if reviews are mandatory, wouldn't you want to enforce them automatically with GitHub's new protected branches setting?

We tried protected branches very briefly, and immediately abandoned the idea because it added too much friction to the merge process. Most of the time I don't merge PRs in their original form - my usual workflow (even for a simple single-commit PR) is to pull, rebase on master, and add the appropriate changelog / release note / contributors.rst entry with `git commit --amend` before pushing. At that point, Github doesn't consider it to be the same commit, and would reject the push on a protected branch. (Also, it's not unknown for the Travis CI tests to have a backlog of several hours...) Having said that - since we made that test, Github has introduced several new features (such as the 'squash and merge' button and the 'allow commits from maintainers' option) which might partly mitigate those problems.

thibaudcolas:
Merged https://github.com/wagtail/wagtail/commit/c0cd53918e3555e30b00af61a0ffbcc5778ce7f9 🎉 

@gasman I followed committing guidelines to the letter. Would appreciate your opinion on the style of the release notes and rebasing from 8 to 2 commits. Phew.

gasman:
@thibaudcolas That all looks good to me!

tomdyson:
Great work, @thibaudcolas and @jjanssen!

PR #3024, opened 2016-09-23
Improved 'purge_page_from_cache' to purge the url for any managed language

pyMan:
In a multi language environment, when publishing a page, only the url of
the current language is being purged. With this improvement now a purge
for that url is done against any managed language code.

So if settings.LANGUAGES includes 'en', 'it', 'fr' and you modify a page when working with current language 'it' then not only "/it/pages/1/foo-bar" is purged, but "/en/pages/1/foo-bar" and "/fr/pages/1/foo-bar" too.

This happens if you set USE_I18N to True (as required by Wagtail if a multilanguage environment is needed). 
Anyway it is not mandatory you choose to use i18n_patterns in urls.py to serve the same page with different language urls, so in this case just a purge with no language code will be done to the reverse proxy backend, like if you weren't in a multilanguage environment.


pyMan:
The only travis failed test has nothing to do with this modification (I think so).

pyMan:
@kaedroho I added a test. Please check if it's good enough because I am new at unit testing.

In the 7548 build test all test are ok except for two, it's the same "random" problem had above. Anyway during the previous run (7547) all tests were successfull, just fixed some flake8 warning.

gasman:
The random test failures should be fixed by #2702, I believe. @kaedroho, please can you review that one?

CHANGELOG.txt
```diff
@@ -28,7 +28,7 @@ Changelog
  * Fix: `InlinePanel` now accepts a `classname` parameter as per the documentation (emg36, Matt Westcott)
  * Fix: Disabled use of escape key to revert content of rich text fields, which could cause accidental data loss (Matt Westcott)
  * Fix: Setting `USE_THOUSAND_SEPARATOR = True` no longer breaks the rendering of numbers in JS code for InlinePanel (Mattias Loverot, Matt Westcott)
-
+ * Fix: `purge_page_from_cache` now purges the url of each managed language when using `i18n_patterns` (PyMan Claudio Marinozzi)
```
kaedroho: I'd class this as a new feature rather than a bug fix

wagtail/tests/urls.py
```diff
@@ -38,5 +38,6 @@
 
     # For anything not caught by a more specific rule above, hand over to
     # Wagtail's serving mechanism
-    url(r'', include(wagtail_urls)),
+    url(r'', include(wagtail_urls))
```
kaedroho: Please remove these changes

gasman:
Superseded by #3971

Issue #542, opened 2014-08-15
Search: Extract text from documents


kaedroho:
By the looks of it, this should be quick and easy to implement now that theres a good python module that can do most of the work for us (https://github.com/deanmalmgren/textract).

Notes:
- `textract` has many dependencies (including `lxml`) so it would be a good idea to make it only an optional dependency
- For large PDFs (eg, ebooks). `textract` can take a few minutes to extract the text

Implementation ideas:
- Add a method to `Document` model called `extract_text`. Calling this will use the `textract` module to extract text from the document file and return it as a string.
- Add a new field to `Document` called `extracted_text`. This is to store the extracted text in the database to make indexing the document into the database faster. This field should be nullable (which would indicate that the text hasn't been extracted from the document yet).
- Create some kind of background task to extract text from documents.
- Add a custom manager for `Document` and override the `get_queryset` method to run `.defer('extracted_text')` on the queryset (https://docs.djangoproject.com/en/dev/ref/models/querysets/#defer). This prevents django from selecting the extracted text when it's not needed so large documents don't slow anything down.
- Add `extracted_text` into `search_fields` of `Document`.

BertrandBordage:
I agree with all your ideas @kaedroho.

I suggest creating a new command named `transcribe_documents` and naming the new field `transcription` instead of `extracted_text`.
This new command should use the default method for transcription, and do something special for PDFs. When the PDF transcription is extremely small compared to the file size, we fall back to using the Tesseract method.
It should also issue a warning when an unsupported file type is met and advice users to install a dependency. For example, if a document is an OGG file and sox is not installed, then we log a warning telling sox can be installed in order to transcribe this document.

We could also use tqdm in this new command to give an estimate of where we are in the transcription and how long it should take.
By the way, we could also use tqdm in `update_index` instead of the current system. It would be way shorter (only one progress bar for each model, for example) and it would also give an estimate of the time remaining.

frague59:
What about PDF OCR ? 

Tika, associated to elasticsearch, provides a solution to extract text from PDF including images (document scans, thanks to tesseract), that's I think a elasticsearch based solution should more transparent.

Much more, the long time taking analysis is performed in elasticsearch, in a delegated process, which avoid to use celery or cron to manage those processing.

Thanks for your great CMS !

frague59:
Oops, seen that tesseract is also included in textract.

BertrandBordage:
[...]

PR #3028, opened 2016-09-26
added support for elasticsearch mapper-attachments (v1.7 and 2.x)

andreasnuesslein:
to full-text index PDFs and such, one can now just install
ElasticSearch's mapper-attachment
(https://www.elastic.co/guide/en/elasticsearch/plugins/master/mapper-attachments.html)

this replaces issue #2992

### DIFF
--- a/wagtail/wagtailsearch/backends/elasticsearch.py
+++ b/wagtail/wagtailsearch/backends/elasticsearch.py
@@ -6,7 +6,7 @@ import warnings
 from django.db import models
 from django.utils.crypto import get_random_string
 from django.utils.six.moves.urllib.parse import urlparse
-from elasticsearch import Elasticsearch, NotFoundError
+from elasticsearch import Elasticsearch, NotFoundError, RequestError
 from elasticsearch.helpers import bulk
 
 from wagtail.utils.deprecation import RemovedInWagtail18Warning
@@ -26,7 +26,7 @@ class ElasticsearchMapping(object):
         'DateField': 'date',
         'DateTimeField': 'date',
         'DecimalField': 'double',
-        'FileField': 'string',
+        'FileField': 'attachment',
         'FilePathField': 'string',
         'FloatField': 'double',
         'IntegerField': 'integer',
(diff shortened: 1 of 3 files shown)


nutztherookie:
@kaedroho reopened this issue here. hoping for input :)

nutztherookie:
If possible, could somebody give some input on this? Should I change something? Is it acceptable? 

Thanks

/cc @gasman

wagtail/wagtailsearch/backends/elasticsearch.py
```diff
@@ -509,9 +510,18 @@ def add_model(self, model):
...
         # Put mapping
-        self.es.indices.put_mapping(
-            index=self.name, doc_type=mapping.get_document_type(), body=mapping.get_mapping()
-        )
+        try:
+            self.es.indices.put_mapping(
+                index=self.name, doc_type=mapping.get_document_type(), body=mapping.get_mapping()
+            )
+        except RequestError as e:
+            if 'No handler for type [attachment] declared' in str(e.info):
```
kaedroho: I think checking errors like this is a little flaky. They could reword it at any time and we wouldn't know to update this check. We need a unit test to tell us that this check does catch the error correctly.

wagtail/wagtailsearch/backends/elasticsearch.py
```diff
@@ -509,9 +510,18 @@ def add_model(self, model):
...
+                raise AssertionError(
```
kaedroho: I don't agree with swapping the ``RequestError`` for a plain ``AssertionError`` here.

Perhaps we could raise a ``RuntimeWarning`` with the hint instead?

I think this belongs in the docs as well.

frague59:
Will this feature be present in wagtail 1.8 ?

frague59:
Will this feature be in wagtail 1.8 ?

gasman:
@frague59 No - as per https://github.com/torchbox/wagtail/pull/3028#pullrequestreview-6413637 this is still awaiting tests and documentation, and 1.8 is in feature freeze now.

frague59:
Is it possible to merge this request into wagtail tree ? 

The patch looks like to be ready, no ?

Thanks !

BertrandBordage:
@frague59 There’s another discussion about this, regarding a more general approach for this problem: #542.
We’re probably going to drop this pull request as the team was more in favor of the Python approach from #542.

Issue #1667, opened 2015-09-08
Do not allow delete on pages with children or at least show warning


chhantyal:
I had a page with hundreds of child pages. Looks like someone accidentally deleted said parent page, and all data is gone. I have backup, but lots of things have changed on current data and backup from yesterday is not going to adjust/fix this properly.  

This is not very nice thing to happen, few things we could do.
1. Show list of child pages that are going to be deleted because of parent page removal (this is similar to how django shows list of related objects that are going to be deleted when you delete an object)
2. Show warning that this is parent page and user might lose data (child pages)
3. Do not allow delete on parent page. First delete child pages, and once the parent page itself is leaf node, it can be deleted. Otherwise not.

Last thing, [Edit: There is separate issue #496] it would be nice to have something similar to django admin log. It keeps simple logs of all the actions performed  via admin.

davecranwell:
Sorry to hear of that unfortunate accident. I believe this is a duplicate of #687 and #496

chhantyal:
Sorry but I don't think it is duplicate of #687 That one talks about related objects, above issue is about child pages.

gasman:
Agreed...

davecranwell:
hmm, you're right. #496 covers the logging, but I'm sure there's another ticket somewhere that mentioned listing the affected objects of a proposed deletion.

davecranwell:
I think a combination of 1. and 2. would work best. Suggestion 3. is likely to prove problematic, since deleting a top-level section should be possible and if it contained hundreds of subsections, each with their own pages, they too would be undeletable till their children were deleted. Even with Bulk Actions (#121) I think that would be a pain.

coredumperror:
I don't think simply warning users is going to be enough here, because too many users have deletion permission in the first place due to #596. Even with a warning, a user who isn't paying attention (and don't say you've never clicked OK without reading the warning message. We've all done that.) can still accidentally delete hundreds of pages they don't own unless you outright deny deletion of Pages that have children.

But if that's too draconian, maybe you could add a special permission for the "delete this page and all of its children" action. That way the catastrophic accidents are at least restricted to a small cadre of privileged Administrators, rather than a potentially massive swarm of editors.

davecranwell:
Relevant: https://github.com/torchbox/wagtail/issues/2213

[...]

PR #3029, opened 2016-09-27
Page publish permission should not automatically give ability to delete

gasman:
The old permission logic allowed anyone with publish permission to delete pages, with no further checks applied. This is incorrect, because the permission rules applied elsewhere establish that 1) deletion is equivalent to editing, and 2) publish permission DOES NOT imply edit permission.

This PR updates the logic to be consistent with other actions: deleting a page is now only possible if you have edit rights on the page (either through explicit 'edit' permission, or implicitly by having 'add' permission and owning the page in question). For published pages, publish permission is required _in addition_ to this.

It's unlikely that this change will affect real-world sites (and even less likely that it will affect them negatively), because having publish-only users without corresponding edit rights is a pretty esoteric setup (and one that currently has some dubious behaviour at the UI level - e.g. options that appear to be available but lead to a 403 error). However, having consistent permission rules here is an important pre-requisite for adding a distinct 'bulk delete' permission level (#1667).

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -1684,7 +1684,7 @@ class PagePermissionTester(object):
         if 'edit' in self.permissions:
             # if the user does not have publish permission, we also need to confirm that there
             # are no published pages here
-            if not ('publish' in self.permissions):
+            if 'publish' not in self.permissions:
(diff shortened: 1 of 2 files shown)


gasman:
Merged in bfa9a953fc6e59166e7c0bf2e0eb6d9ecb405b77 as part of #3035

PR #3032, opened 2016-09-28
Pass parent context into block's get_context method

ipmb:
Having access to the page context inside the `get_context` method is
useful in some situations. Prior to this, one would need to override
`render` to use the page context.

My use case was a block that showed a list of recent pages. These were added via `get_context` but I needed to exclude the current page the block is displayed on if it was in the list. It makes more sense to have all the context be available in `get_context` rather than some getting injected in inside the `render` method.

Existing tests pass. Happy to add docs and/or tests if this is a welcome contribution.

See also #2786

### DIFF
--- a/wagtail/wagtailcore/blocks/base.py
+++ b/wagtail/wagtailcore/blocks/base.py
@@ -226,15 +226,17 @@ class Block(six.with_metaclass(BaseBlock, object)):
         """
         return value
 
-    def get_context(self, value):
+    def get_context(self, value, context=None):
         """
         Return a dict of context variables (derived from the block value, or otherwise)
         to be added to the template context when rendering this value through a template.
         """
-        return {
+        context = context or {}
+        context.update({
             'self': value,
             self.TEMPLATE_VAR: value,
-        }
+        })
+        return context
 
     def _render_with_context(self, value, context=None):
         """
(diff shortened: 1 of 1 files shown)


gasman:
Thanks @ipmb! As per the discussion at https://github.com/torchbox/wagtail/pull/2786#issuecomment-230416360, I think having access to the parent context in `get_context` is a worthwhile addition. So yes, please go ahead with tests and docs!

The concern about how to accommodate 'legacy' `get_context` methods on user-created subclasses (i.e. ones that haven't yet been updated to expect a `context` kwarg) still stands, I think. I suspect it just needs a "look before you leap" signature check, similar to the one we ended up adding on #2786:

    check the signature of the get_context method we're about to call
    (using the existing 'accepts_context' helper)
    if it expects a 'context' kwarg:
        call get_context with the parent context
    else:
        output a deprecation warning
        call get_context without the parent context
        merge the two dicts ourselves

Nossnevs:
I have added tests and legacy handling and created a pullrequest to @ipmb branch. I don't know what documentation is required. 

The commit uses the RemovedInWagtail18Warning which is wrong. Which RemovedIn warning should i use? 

Should i create a new pull request to Wagtail or should i keep creating pul lrequest to @ipmb?

gasman:
Thanks @Nossnevs! Creating a new pull request to Wagtail is best, I think.

We usually leave two releases between deprecating a feature and dropping it, so if we're aiming to get this into Wagtail 1.9, it should use `RemovedInWagtail111Warning`.

gasman:
Completed in #3275.

PR #3034, opened 2016-09-28
Add documentation about permission configuration

gasman:

Issue #3031, opened 2016-09-27
"Choose another page" button on group permissions page is cut off


gasman:
Increasing the font size of buttons (#2985) has resulted in the text of the "Choose another page" button under Settings -> Groups -> Page permissions being cut off:

<img width="409" alt="screen shot 2016-09-27 at 22 43 07" src="https://cloud.githubusercontent.com/assets/85097/18893016/f9047864-8503-11e6-88ff-e2afeacf4807.png">

gasman:
Oops, looks like this is specific to #3035 because of the extra column. Will fix it there...

strindhaug:
BTW: if the design is dependent on an exact string length and font size, it will be almost impossible to translate well. When translating the length of a string could easily double (or halve) because of different word lengths and grammar in the other language. The layout should tolerate at least ± 50% string (ideally more) length change without breaking or overflowing for it to be easy to translate.

An added bonus is that if the design tolerates this, adding an extra column won't suddenly completely break the layout as well.

PR #3035, opened 2016-09-28
Add a specific 'bulk delete' permission type for deleting pages with children

### LINKED ISSUE #1667: Do not allow delete on pages with children or at least show warning

gasman:
Fixes #1667; incorporates #3029 and #3034.

See the included docs for details about the behaviour of the new permission type. I realise it's slightly unintuitive, since the 'bulk delete' permission doesn't actually confer any deletion rights of its own, but acts as a kind of accelerator for the deletion permissions the user has already - but the only obvious alternative (namely, 'bulk delete' meaning 'can delete everything in this subtree') would have made it impossible to restore the old behaviour without handing ordinary editors a nuclear button :-)

(In practice, I suspect that the vast majority of site administrators will not touch this setting at all; they will be happy to revoke bulk-delete powers from ordinary editors and keep it superuser-only.)


gasman:
CSS fix required as per #3031

gasman:
Thanks @kaedroho! Now merged in 8de1b1b16f74a970619e7207b5bbb87a7e6000bc + parents.

PR #3037, opened 2016-09-29
Stop wagtailusers.UserProfile's related_name from clashing with other user profiles

### LINKED ISSUE #3025: wagtailusers.UserProfile related_name can collide with user profiles from other apps
As noted at https://groups.google.com/d/msg/wagtail/xQh4b2PF7d8/rIzj6o6fBAAJ, the wagtailusers.UserProfile model does not specify a related_name on the user field, leaving it at the default of `user.userprofile`. This means it's likely to collide with UserProfile models defined by other apps in the project.

From a quick look through the code, it looks like we consistently access the UserProfile with `UserProfile.get_for_user`, which would mean we don't need the related_name at all. `wagtail_userprofile` would probably be a sensible choice, though.

gasman:
Set a related_name of wagtail_userprofile on wagtailusers.UserProfile. This prevents it from clashing with other userprofile models defined elsewhere in the project. Fixes #3025


gasman:
Merged in f45181764bd5fa3b2fedd7df2d81f77751cbea56

Issue #1288, opened 2015-05-11
CSP compatibility issues


salvadormrf:
I have enabled Content Security Policy (CSP) directives on wagtail admin using django-csp.
CSP is not required for all websites but for some clients is mandatory as part of the security requirements.

Here is a list of issues I have found by enabling CSP on wagtail admin.
- use of inline script (<script>createDocumentChooser("body-2-value-document");</script>)
- use of inline style (<div class="focal-point-chooser" style="max-width: 400px;...;")
- use of javascript eval (In Modal choosers)
- gravatar images being loaded  

I tried to override all templates where inline scripts were being used to add "nonce" attribute.
However widgets generate the <script> tags and they don't have access to request to access request context variables.  

I tried to set "nonce" attribute on inline scripts is to wrap the whole base.html with a custom django template tag. Beautifulsoup parses all HTML nodes for script tags and sets the nonce attribute. This approach assumes that all html content being rendered is always safe.

The modal use javascript eval, and at the end was to many files being changed. 
So I had be more flexible with the CSP directives (less secure) in order to keep everything working fine on wagtail admin.

Would be nice while developing wagtail that CSP could be taken into consideration, and where it is possible to try to avoid inline styles and scripts unless there is a proper way to mark it as safe.

Thanks!
I am enjoying even more wagtail as I use it  for development :)

JoshBarr:
I agree we need to look at the way those scripts are attached – it makes the lines between frontend and backend a little blurry :) I'm hoping to take a look at it after the Torchbox team have published the 1.0 release.

moggers87:
I started working on a patch for Wagtail to use nonces (as they're now supported in in django-csp) for inline scripts. However, I had to give up when I realised that Wagtail would still require `'unsafe-eval'`.

The code is can be found on [this branch](https://github.com/moggers87/wagtail/tree/csp-nonces) if anyone is interested.

Pomax:
What is the cause of the `unsafe-eval` CSP violation, though? jQuery, or something else?

moggers87:
https://github.com/wagtail/wagtail/search?l=JavaScript&q=eval&type=&utf8=%E2%9C%93

rspeed:
I only see one relevant result in that search, and if I’m not mistaken it could be replaced with `JSON.parse`.

Pomax:
[...]

PR #3038, opened 2016-09-29
Refactor page chooser into a react component

kaedroho:
This pull request refactors the page chooser into a react component that pulls its data in over the Admin API.

This should (hopefully!) simplify the chooser code quite a bit. It removes all the passing of HTML around and the need to `eval()` on snippets of Javascript (which is reported as an Issue #1288 #3011). It also feels a bit faster too (local testing).

There should be no behavioural changes and I've not attempted to change any of the HTML/CSS (beyond what is required for the refactoring).


sebworks:
Is there a timetable on converting to React?

thibaudcolas:
FYI I had to build something similar for one of my projects so I rebased the existing branch, and did some cleanup / refactoring. This work is available at https://github.com/springload/wagtail/tree/react-page-chooser-rebased.

The rebasing was simple – I did a `git rebase -i master` and completely dropped all of the commits until the first one from @kaedroho. From there, the only difference in the page chooser is the implementation of `user_perms` (parameter forwarded in the chooser URL).

I'll stop there, the code that I write next may depart from what Wagtail needs so I'll keep it separate.

kaedroho:
Thanks @thibaudcolas. I've just updated this branch to use yours (previous version [here](https://github.com/kaedroho/wagtail/tree/react-page-chooser-old))

thibaudcolas:
If anyone stumbles upon this and is interested in helping making this happen, I'd love to help.

Generally when people ask how to extend the link chooser UI I point them at this PR as a way forward (e.g. for tel links, #5086, nofollow attr #4474).

kaedroho:
I've marked up all the strings for translation now and added the translation catalog JS file so we can retrieve translated strings in the frontend.

One problem is that the chooser is in the "client" folder which means it's not in an app so Django's makemessages won't pick up any of the strings.

kaedroho:
I've reverted the above change and redone translations to use the existing method.

kaedroho:
I think all that's left to do on this now:

 - [ ] Open/close fade animation
 - [ ] Accessibility

kaedroho:
This is ready for review now!

zerolab:
Note: this requires someone with React knowledge to review.

One thing that stays at the back of my mind as a non-React developer is.. how does one go about building upon these?

kaedroho:
> One thing that stays at the back of my mind as a non-React developer is.. how does one go about building upon these?

Currently the aim is just to get the current page chooser (that isn't itself customisable) working in React. Wll have to answer this when we get to implementing other choosers such as the link chooser which does allow adding tabs. What I have in mind at the moment is to allow modals to be a hybrid of React and AJAX (so you could create a modal that has a react based image chooser with an AJAX-driven Upload tab, for example).

PR #2417, opened 2016-03-29
Add ability to edit links in rich text editor

loicteixeira:
See corresponding issue #444.

 - [X] If a link already exists, open the link chooser panel on the relevant tab, otherwise default to page tab.
 - [X] Pre-populate link text and url when available
 - [X] Add a remove link button (with new broken-chain icon)
 - [X] Add undo/redo support for remove link
 - ~~Add undo/redo support for add/edit link~~
 - [X] Always show the link text field for external/email tabs since the text is now pre-populated?
 - ~~On the page tab, highlight the currently selected page?~~
 - [X] Add tooltip on link over in the rich text editor?

PR #3039, opened 2016-09-29
Allow links in rich text to contain non-text markup again

### LINKED ISSUE #3026: No longer possible to make an image a link within rich text
As reported at https://groups.google.com/d/msg/wagtail/FA9qu7R2mCA/CayK-NrQBAAJ - since the updates to rich text link handling in Wagtail 1.5 (#2417), it's not possible to make an image into a link by highlighting it and using the link button on the toolbar. This now causes the image to be removed.

I believe this is because the new code passes 'link text' as a URL parameter to the chooser modal and uses that to construct an HTML fragment to replace the selected text, rather than wrapping the existing selected text in a tag - this won't work if the link includes non-text content such as images.

gasman:
Fixes #3026

The link-handling logic introduced in Wagtail 1.5 (#2417) works by extracting the text content of the current link or selection, and then replacing it with a newly-constructed `<a>` element containing the (possibly-updated) link text. This didn't account for the possibility of the link/selection content containing non-text markup, such as images or bold/italic formatting.

This PR rewrites the link-handling logic to always preserve the existing DOM content, unless the user explicitly provides updated link text within the chooser modal. This involves adding a new `prefer_this_title_as_link_text` flag to the chooser's return value.

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/hallo-plugins/hallo-wagtaillink.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/hallo-plugins/hallo-wagtaillink.js
@@ -69,41 +69,41 @@
                         urlParams: urlParams,
                         responses: {
                             pageChosen: function(pageData) {
-                                var a, text;
+                                var a, text, linkHasExistingContent;
 
-                                // Create link
-                                a = document.createElement('a');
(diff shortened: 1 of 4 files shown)


gasman:
Merged in 5dc5a32adc9ab324b360ce0aa953256d55eaaa37 + parents

PR #3041, opened 2016-09-30
Get title display

Henk-JanVanHasselaar:
Method used to display the title of your page in the wagtail admin instead of `page.title`


kaedroho:
``get_title_for_display`` sounds a bit generic for something that should only be used in the admin. Is there any way we could squeeze the word ``admin`` into that name?

Henk-JanVanHasselaar:
@kaedroho I'll rename to `get_admin_display_title`

gasman:
Merged in 8a3c47f76afcb45182aa11cc94e5e0443d2836a5 - thanks!

(Note: I reverted the definition `Page.__str__` to the original one, as we can't rely on this only being used within the admin.)

PR #3036, opened 2016-09-28
ElasticSearch index default settings override

pyMan:
Introduced a new optional INDEX_OPTIONS key for WAGTAILSEARCH_BACKENDS
setting.

INDEX_OPTIONS should be a dictionary whose data is used to be merged
with ElasticsearchSearchBackend.settings (data used when a index is
being created)

Sample to set 'number_of_shards' and 'number_of_replicas' and to set a new default LanguageAnalyzer for the index.

```
WAGTAILSEARCH_BACKENDS = {
      'default': {
            'BACKEND': 'wagtail.wagtailsearch.backends.elasticsearch',
            'INDEX': 'wagtail',
            'URLS': ['http://localhost:9200'],
            'TIMEOUT': 5,
            'INDEX_SETTINGS' : {
                "settings" : {
                    "number_of_shards" : 2,
                    "number_of_replicas" : 1,
                    "index" : {
                        "analysis" : {
                            "analyzer" : {
                                "default" : {
                                    "type" : "italian"
                                }
                            }
                        }
                    }
                }
            }
        }
    }
```

@kaedroho This modification alse includes and closes #2778 (extra ES params are now passed through new OPTIONS key in the
WAGTAILSEARCH_BACKENDS setting. Backward compatible with deprecation warning)

PR #3043, opened 2016-10-03
Make WagtailSearch Language Aware (SearchFields and Indexes)

pyMan:
Pull request to reach goal of #1177 and #2179 and some improvements to those idea.

A little summary:
- Configure a ES backend for each managed language, plus a default backend for no specific language.
- SearchField now gets a new `language` parameter (default None). SearchFields with language None will be indexed in any index, otherwise they will be indexed only in the index with the same language code.

```
WAGTAILSEARCH_BACKENDS = {
    'default': {
        'BACKEND': 'wagtail.wagtailsearch.backends.elasticsearch.ElasticSearch',
        'URLS': ['http://localhost:9200'],
        'INDEX': 'wagtail',
        'TIMEOUT': 5,
    },
    'en': {
        'BACKEND': 'wagtail.wagtailsearch.backends.elasticsearch.ElasticSearch',
        'URLS': ['http://localhost:9200'],
        'INDEX': 'wagtail_en',
        'TIMEOUT': 5,
        'LANGUAGE_CODE': 'en'
    },
    'it': {
        'BACKEND': 'wagtail.wagtailsearch.backends.elasticsearch.ElasticSearch',
        'URLS': ['http://localhost:9200'],
        'INDEX': 'wagtail_it',
        'TIMEOUT': 5,
        'LANGUAGE_CODE': 'it'
    },
}

class MyPage(Page):
    slogan = models.TextField()
    body_en = models.TextField()
    body_it = models.TextField()

    search_fields = (
        index.SearchField('slogan', language=None),  # A text suitable for any language
        index.SearchField('body_en', language='en'),
        index.SearchField('body_it', language='it'),
    )
```

So far the index `wagtail` will contains all fields, `wagtail_en` will contains only `slogan` and `body_en`, `wagtail_it` only `slogan` and `body_it`.

Please note that with the help of #3036 you'll be able to configure a specific default LanguageAnalyzer for `wagtail_en` and `wagtail_it` indexes to improve quality of indexing and search results.

```
>> MyPage.objects.search("hello")  # The index relating the current active language will be hit, fallback to 'default'
[ some results ]
>> MyPage.objects.search("hello", backend = "en")  # The index of "en" backend will be hit
[ some results ]
```


BertrandBordage:
I’m not sure that’s not the right approach to deal with this problem.
This takes a lot of disk space, and it doesn’t handle the case when you want to search in multiple languages instead of the current one or a specific one.

But the biggest concern from what I experienced is that this system doesn’t allow you to specify the language(s) of a given object. For example, suppose you index books. Most likely, your model will not contain fields like `title_en`, `title_fr`, `title_de`, etc. Instead, it will contain a `language` field and the book will not have any translation, or if it does, this will be something like a `ManyToMany` between translated books and the original one(s). So we have to rely on something like a `get_language` model method to know which language(s) are contained in the book in order to know in which language it’s relevant to find this book.

I don’t know what would be the best approach for elasticsearch, but I can tell that PostgreSQL can store the language(s) in each index entry and then index data in the correct language(s). This makes one index for all languages, where you can easily do a single search query in all languages, or just one or several.

kaedroho:
Thanks for implementing this!

We've recently added internationalisation support into Wagtail itself, so the approach will probably have to be changed, so I'm going to close this PR.

PR #3044, opened 2016-10-03
Trim trailing whitespace in all files but Markdown

thibaudcolas:
It is common to use [EditorConfig](http://editorconfig.org/) to trim trailing whitespaces at the end of files (for example, [Django does it](https://github.com/django/django/blob/master/.editorconfig)).

It is disabled for Markdown because there trailing whitespace is used to indicate line breaks (not something I would advise on using, but I'd rather not break existing documentation).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,6 +3,7 @@ root = true
 [*]
 end_of_line = lf
 insert_final_newline = true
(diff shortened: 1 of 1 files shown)


gasman:
Good idea - thanks!

PR #3046, opened 2016-10-04
Return a 403 response for unauthenticated ajax requests

kaedroho:
At the moment, Wagtail redirects all requests (including AJAX) to the
login view. This is not usually the expected thing to do for AJAX and can
lead to the login page being nested in a component somewhere (which
looks horrible).

This changes the behaviour so requests that come from AJAX are given a
plain 403 error. This allows the code that performed the request to
handle the issue properly.

### DIFF
--- a/wagtail/wagtailadmin/decorators.py
+++ b/wagtail/wagtailadmin/decorators.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, unicode_literals
 
 from django.contrib.auth.views import redirect_to_login as auth_redirect_to_login
+from django.core.exceptions import PermissionDenied
 from django.core.urlresolvers import reverse
 from django.utils.translation import ugettext as _
 
@@ -8,7 +9,10 @@ from wagtail.utils.compat import user_is_anonymous
 from wagtail.wagtailadmin import messages
 
 
-def redirect_to_login(request):
+def reject_request(request):
+    if request.is_ajax():
+        raise PermissionDenied
+
     return auth_redirect_to_login(
         request.get_full_path(), login_url=reverse('wagtailadmin_login'))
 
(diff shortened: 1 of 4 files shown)


wagtail/wagtailadmin/decorators.py
```diff
@@ -3,26 +3,33 @@
...
 from wagtail.utils.compat import user_is_anonymous
 from wagtail.wagtailadmin import messages
 
 
-def redirect_to_login(request):
-    return auth_redirect_to_login(
-        request.get_full_path(), login_url=reverse('wagtailadmin_login'))
+def reject_request(request):
+    if request.is_ajax():
+        return HttpResponseForbidden()
```
BertrandBordage: I personally prefer raising `PermissionDenied` instead, as I supposed it creates a better error page when in DEBUG mode.

wagtail/wagtailadmin/decorators.py
```diff
@@ -3,26 +3,33 @@
...
+    else:
```
BertrandBordage: This `else` can be removed.

wagtail/wagtailadmin/decorators.py
```diff
@@ -3,26 +3,33 @@
...
         if user_is_anonymous(user):
-            return redirect_to_login(request)
+            return reject_request(request)
 
         if user.has_perms(['wagtailadmin.access_admin']):
             return view_func(request, *args, **kwargs)
 
-        messages.error(request, _('You do not have permission to access the admin'))
-        return redirect_to_login(request)
+        if not request.is_ajax():
```
BertrandBordage: Maybe reverse this condition? Not sure.

kaedroho: This bit of code adds a message to display to the user when they get redirected to the dashboard but AJAX requests will no longer be redirected so I removed the message to prevent it from popping up in confusing places.

wagtail/wagtailadmin/decorators.py
kaedroho: Fixed, thanks!

wagtail/wagtailadmin/decorators.py
kaedroho: Fixed, thanks!

kaedroho:
Tests have been added. I've also added this behaviour to the ``user_passes_test`` utils function.

gasman:
Merged in 579d02246a6ba03c73baf38f9f6923ea38d164a7

PR #3047, opened 2016-10-05
Allows to define a custom site settings edit handler

noirbizarre:
Right now, site settings (from `wagtail.contrib.settings`) only allows a single linear `panels` attribute for edit handlers.

This pull-request allows to define a custom `edit_handler` attribute like on `Page` model allowing to use tabs for site settings.


gasman:
@noirbizarre Thanks! Please could you add tests for this?

(note that there's also a styling issue reported here https://github.com/torchbox/wagtail/pull/2676#issuecomment-224659922 - although that can be fixed separately if need be)

noirbizarre:
I will try but right now there is no test on nominal cases (use of `panels` attribute or let wagtail generate the fields from the model definition). But I'll try to find some inspiration from https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailadmin/tests/test_edit_handlers.py

About the visual bug, I'll find a fix 👍 

This PR as been labeled 1.8 but the 1.7 has not been released yet. If I provide the required elements before the 1.7 release is made, would it be included into ?

gasman:
It's too late to go into 1.7 I'm afraid - I'm preparing the 1.7 release candidate right now.

noirbizarre:
Too bad 😞 
I moved the changelog line to 1.8 section.

I updated the PR with some tests (also added tests for existing cases) and I fixed the visual bug with tabs.

gasman:
Looks like the test app is missing a migration - please can you add this? (You should be able to generate it by running `django-admin.py makemigrations --settings=wagtail.tests.settings` from the Wagtail root)

noirbizarre:
OK, done. I've put the models into `testapp` and added the missing migration to be consistent with other tests.

m1kola:
Looks good to me. Thanks, @noirbizarre

Merged in a3ba533afb94b7b3698b409b07009b0380718233...2e1c58a5860d6a9c9ad28fd40954f7f0714fb003

PR #3053, opened 2016-10-05
Clean Whitelist result from well-formed html tags

belegnar:
BeautifulSoup with html5lib parser adds html, head and body tags to produce well-formed html. This tags are not allowed within body of html page and should be stripped

### DIFF
--- a/wagtail/wagtailcore/whitelist.py
+++ b/wagtail/wagtailcore/whitelist.py
@@ -97,7 +97,10 @@ class Whitelister(object):
         attributes"""
         doc = BeautifulSoup(html, 'html5lib')
         cls.clean_node(doc, doc)
-        return doc.decode()
+        if hasattr(doc, 'body') and hasattr(doc.body, "childGenerator"):
+            return "".join([child.decode() for child in doc.body.childGenerator()])
+        else:
+            return doc.decode()
 
     @classmethod
     def clean_node(cls, doc, node):


gasman:
Hi @belegnar - thanks for the PR!

Please can you give more details of the problem that's being fixed here (perhaps with a unit test)? You're right that html5lib adds html / head / body tags, but they get stripped out again by the `cls.clean_node(doc, doc)` line, giving the expected result:

    >>> from wagtail.wagtailcore.whitelist import Whitelister
    >>> Whitelister.clean('<p>hello world</p>')
    '<p>hello world</p>'

(of course, this relies on html / head / body being omitted from the `element_rules` list - if you create a subclass of Whitelister that allows them, then it would indeed include those tags in the output. But I'm not sure why you would do that, unless you were intending to pass complete HTML documents to it...?)

belegnar:
Definitely, i can't reproduce this with unit test, because
```
In [1]: from bs4 import BeautifulSoup
In [2]: from wagtail.wagtailcore.whitelist import Whitelister
In [3]: html = u'<p>Hello world</p>'
In [4]: Whitelister.clean(html)
Out[4]: u'<p>Hello world</p>'
```
But all pages within my site contain `<html><head></head><body>[rich text field content]</body></html>`. So, i've modified `Whitelister.clean()` to
```
    @classmethod
    def clean(cls, html):
        """Clean up an HTML string to contain just the allowed elements /
        attributes"""
        print 'whitelister in', html
        doc = BeautifulSoup(html, 'html5lib')
        cls.clean_node(doc, doc)
        print 'whitelister out', doc.decode()
        return doc.decode()
```
Then i create page 'Test' with body 'body', intro 'intro' and StreamField component Paragraph 'hello world'. Output is
```
whitelister in <p>intro</p>
whitelister out <html><head></head><body><p>intro</p></body></html>
whitelister in <p>body</p>
whitelister out <html><head></head><body><p>body</p></body></html>
whitelister in <p>hello world<br></p>
whitelister out <html><head></head><body><p>hello world<br/></p></body></html>
```

gasman:
I don't think this is a Wagtail bug. If the html / head / body tags are not being stripped out by the `cls.clean_node(doc, doc)` step, there's a problem somewhere else in the code - perhaps you've overwritten the `element_rules` list?

Closing for now as unable to replicate... if you can provide a minimal test case that demonstrates the issue, I'll be happy to investigate further.

PR #3055, opened 2016-10-06
Update title

alexfromvl:

Added to the site_name in the title.


gasman:
Hi @alexfromvl - thanks for the PR!

Please can you update this to add a ` - ` delimiter between the page title and site name? Currently it displays "HomepageMy wonderful site" (and HomepageNone if a site name has not been set - in this case it would be better to just show "Homepage").

wagtail/project_template/project_name/templates/base.html
```diff
@@ -8,7 +8,7 @@
...
+        <title>{% templatetag openblock %} block title %}{% templatetag openblock %} if self.seo_title %}{% templatetag openvariable %} self.seo_title {% templatetag closevariable %}{% templatetag openblock %} else %}{% templatetag openvariable %} self.title {% templatetag closevariable %}{% templatetag openblock %} endif {% templatetag closeblock %}{% templatetag openblock %} endblock {% templatetag closeblock %} - {% templatetag openblock %} block title_suffix %}{% templatetag openblock %} if self.get_site.site_name %}{% templatetag openvariable %} self.get_site.site_name {% templatetag closevariable %}{% templatetag openblock %} else %}{% templatetag openvariable %} self.get_site.hostname {% templatetag closevariable %}{% templatetag openblock %} endif {% templatetag closeblock %}{% templatetag openblock %} endblock {% templatetag closeblock %}</title>
```
zerolab: the `%} - {% templatetag openblock %} block title_suffix %}{% templatetag openblock %}` bit has an extraneous `%}` after block_title_suffix

alexfromvl: Corrected. `{% templatetag closeblock %}` put instead `%}`

gasman:
Thanks! Merged (with a couple of modifications: avoiding the repeat of the self.get_site lookup, and I don't think we should fall back on site hostname) in ee4e2ed27bcc9e2d15caf2a54ab19d26b387fe49.

PR #3057, opened 2016-10-07
Fix: Use specific page model for the parent page in the explore index

Gagaro:
Hi,

I have customized `get_url_parts` on my page models so that the URL is a bit different. However, the page used for the `parent_page` in the explore used the Page model and thus I get an incorrect URL for the live button.

The `parent_page` should be of the specific page model in this view (just like the pages are).

Thanks.


gasman:
Thanks @Gagaro! Please could you add a test for this?

Gagaro:
Done

gasman:
Merged in 49421e5a41017095ee82544f287a9b90c79332c5 - thanks!

Issue #2984, opened 2016-09-06
modeladmin's ChooseParentView and DeleteView don't call get_context_data


kaedroho:
modeladmin's ChooseParentView and DeleteView view classes override the ``get()`` method but do not call ``get_context_data()`` which makes modifying them slightly tricker and inconsistent with the other view classes.

ChooseParentView: https://github.com/torchbox/wagtail/blob/master/wagtail/contrib/modeladmin/views.py#L728-L731
DeleteView: https://github.com/torchbox/wagtail/blob/28c8ae3ab99dd1b0b22bb334835c9e823b060cae/wagtail/contrib/modeladmin/views.py#L772-L774

gasman:
Fixed in #3060

PR #3060, opened 2016-10-07
Improves general consistency of modeladmin views

ababic:
As @kaedroho rightly pointed out in issue #2984, there are some inconsistencies in the views, which this pull aims to correct.

In addition to calling `get_context_data` in `ChooseParentView` and `DeleteView`, I've also:

- Altered the way each view extends `get_context_data` to be tidier and more consistent
- Added `form_valid()` and `form_invalid()` methods to `ChooseParentView` to be in-line with the other views with forms
- Make use of `self.render_to_response()` on `ChooseParentView` instead of repeating that same functionality by calling `render()` directly
- Avoid unnecessary redefinition of `view` and `instance` in context values

### DIFF
--- a/wagtail/contrib/modeladmin/views.py
+++ b/wagtail/contrib/modeladmin/views.py
@@ -210,7 +210,6 @@ class InstanceSpecificView(WMABaseView):
 
 class IndexView(WMABaseView):
 
-    # IndexView settings
     ORDER_VAR = 'o'
     ORDER_TYPE_VAR = 'ot'
     PAGE_VAR = 'p'
--- a/wagtail/contrib/modeladmin/views.py
+++ b/wagtail/contrib/modeladmin/views.py
@@ -925,16 +925,13 @@ class InspectView(InstanceSpecificView):
         return fields
 
     def get_context_data(self, **kwargs):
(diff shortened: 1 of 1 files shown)


ababic:
This is ready for review now :)

ababic:
@gasman Good call. I'll update to take that approach instead.

wagtail/contrib/modeladmin/views.py
```diff
@@ -100,10 +100,11 @@ def get_base_queryset(self, request=None):
...
 
     def get_context_data(self, **kwargs):
-        kwargs.update({
+        context = {
             'view': self,
             'model_admin': self.model_admin,
-        })
+        }
+        context.update(kwargs)
         return super(WMABaseView, self).get_context_data(**kwargs)
```
gasman: Should be `get_context_data(**context)` I think...?

gasman:
Thanks! Merged in 9f92f894b8c171f71632d3e59fe4e57dde7814d2 (with fix for https://github.com/torchbox/wagtail/pull/3060#discussion_r84774945 applied in bb37dec700b5791da63fd2bbe7beaa596aca1a8f - forgot to include that in the main commit, oops!)

PR #3061, opened 2016-10-08
Fix: remove responsive behavior in embed when there is no ratio avail…

Gagaro:
Please see #1495


gasman:
Merged in 1d33cc7838c647c95f4c228b35500851cbd5e6d9 - thanks!

Issue #2941, opened 2016-08-24
Feature request: Rearrange order of objects in modeladmin index view


adriaant:
In wagtail admin, it is possible to rearrange the order of pages in the admin editor. This is a feature that would work equally well for custom models managed via modeladmin, provided the models inherit the Orderable mixin. This is useful for pages that provide a list of all/filtered objects of a given model, but for which you want to provide the user with an option to define the order of the objects on the page. 

For example:
A custom model for FAQ categories and a page that grabs the categories dynamically during rendering and displays them according to the order stored in the db. In other words, all the user has to do is create categories (and the contained questions) in the modeladmin interface.

Is this something already being considered or worth discussing?

ababic:
One approach to this would be to create an `OrderableMixin` class, to be used with the `ModelAdmin` class where needed (much like the [`ThumbnailMixin`](https://github.com/torchbox/wagtail/blob/master/wagtail/contrib/modeladmin/options.py#L49). The basic elements of which would be:
 
1. Getting the ID/order values for each object into the listing, so the data is available to create reorder requests
2. Some view functionality to take care of reorder requests
3. Some javascript with drag-and-drop support to bind it all together

From a UI/UX perspective, it'd make sense to match the page reordering approach as closely as possible. I haven't looked yet, but I'd expect there'd be some JS/styles that can be reused, or at least serve as a decent starting point to copy from.

I'd be happy to work on a pull-request.

adriaant:
What is the status of this ticket? I noticed PR is closed, but has actually not been merged.

ababic:
Hey @adriaant. The branch still exists here if you wanted to copy/paste for a project in the meantime: https://github.com/rkhleics/wagtail/tree/modeladmin/orderableadmin-2

I closed #3103 because It didn't look like it was going to be reviewed any time soon, and there are a bunch of other issues that need addressing that would just introduce a bunch of merge conflicts for that #3103. I'd rather just recreate the functionality later (with a bit less scope-creep) after having rebased everything on master.

adriaant:
@ababic Understood, thanks for the quick and clear reply!

inostia:
Is there any reason why this wasn't merged yet? This is a necessary feature for my org, we need to render the items on the FE in the order they appear on the admin interface.

tomdyson:
@inostia why what wasn't merged? This is an issue, not a PR. There was an associated PR, but that was closed for reasons outlined above.

elton2048:
[...]

Issue #2719, opened 2016-06-16
modeladmin select multiple for mass-action


nutztherookie:
hey torchies,

we are trying your new ModelAdmin-feature and realized there is no "I'm selecting a bunch of objects and I'm gonna <delete/export/dosomething> them all". Django Admin:
![image](https://cloud.githubusercontent.com/assets/320080/16110319/76ad85b0-33ad-11e6-82b7-b11e0b62721a.png)

So we were wondering: Are you planning on implementing this? If so, do you know when that might be the case? :) 

Thanks a bunch

PS: Great job on the ModelAdmin and Wagtail in general

ababic:
Hi @nutztherookie! Thanks a lot for the feedback :)

I'd certainly love to see something that replicated Django ModelAdmin's 'actions' in the future, but there are no set plans to tackle that right away. I guess we have some thinking/decision-making to do, in regards to how this app evolves and becomes more integrated into the wagtail core (and what things in core need to also change, in order for that to happen)

When `modeladmin` was still in it's infancy (as [wagtailmodeladmin](https://github.com/rkhleics/wagtailmodeladmin)), I made an attempt to add an 'actions-like' feature, but at the time, I stumbled when it came to handling permissions, and ensuring only users with the right actions could perform them. I think it was particularly complicated because of the way pages have their own tree-based permission system, and I decided it was a little too complicated to tackle on my own. Hopefully now, with a keen group of wagtail contributors behind it, that's something that might be overcome.

nutztherookie:
Hey @ababic, thank you for your elaborate answer =)

I know little of wagtails permissions-setup; but also I don't really need it for my project.
I'm gonna start hacking 'actions' into wagtail and I'll see where it leads me. If it turns out to be somewhat mature I'll try and upstream it :)

SalahAdDin:
But, if we'll have this features, `django-admin` will be 'deprecated'. So, for now we can used `django-admin`normally.

ababic:
@salahAdDin I'm afraid I'm struggling to understand what you mean there. Are you suggesting that this feature should or shouldn't be implemented? What do you mean when you say django admin will be 'deprecated'?  I'm pretty sure django admin isn't going anywhere. It's a fundamental part of django.

SalahAdDin:
Ok, i say, for our use, with `wagtail-admin`and `model-admin`, for us, will be deprecated because, for what have `django-admin` if `wagtail-admin` do all?

ababic:
[...]

PR #3064, opened 2016-10-10
Add `data-object_pk` attribute to rows in contrib.modeladmin.IndexView

ababic:
In an effort to improve the extendability of ModelAdmin (specifically for things like #2941 and #2719).

The `<tr>` element for each row in `IndexView` is now rendered by the `result_row_display` template tag, making it possible to do more with it.

A new `get_extra_attrs_for_row()` method on the `ModelAdmin` class is called by `result_row_display` tag to check for additional html attributes to be added for each row, so developers can override that to add custom attributes to the `<tr>` element.

### DIFF
--- a/wagtail/contrib/modeladmin/templatetags/modeladmin_tags.py
+++ b/wagtail/contrib/modeladmin/templatetags/modeladmin_tags.py
@@ -158,7 +158,11 @@ def result_row_display(context, index):
     view = context['view']
     row_attrs_dict = view.model_admin.get_extra_attrs_for_row(obj, context)
     row_attrs_dict['data-object_pk'] = obj.pk
-    row_attrs_dict['class'] = 'odd' if (index % 2 == 0) else 'even'
+    odd_or_even = 'odd' if (index % 2 == 0) else 'even'
+    if 'class' in row_attrs_dict:
+        row_attrs_dict['class'] += ' %s' % odd_or_even
+    else:
+        row_attrs_dict['class'] = odd_or_even
     row_attrs = ''.join(
         ' %s="%s"' % (key, val) for key, val in row_attrs_dict.items())
     context.update({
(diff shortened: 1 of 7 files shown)


ababic:
This is ready for review, now :)

wagtail/contrib/modeladmin/templatetags/modeladmin_tags.py
```diff
@@ -156,8 +156,18 @@ def admin_list_filter(view, spec):
...
     view = context['view']
+    row_attrs_dict = view.model_admin.get_extra_attrs_for_row(obj, context)
+    row_attrs_dict['data-object-pk'] = obj.pk
+    odd_or_even = 'odd' if (index % 2 == 0) else 'even'
+    if 'class' in row_attrs_dict:
+        row_attrs_dict['class'] += ' %s' % odd_or_even
+    else:
+        row_attrs_dict['class'] = odd_or_even
+    row_attrs = ''.join(
+        ' %s="%s"' % (key, val) for key, val in row_attrs_dict.items())
```
gasman: This won't HTML-escape characters like `" < >` in the attribute values. I have a feeling that there's a Django utility function you can use here instead...

gasman: ([django.forms.utils.flatatt](https://github.com/django/django/blob/ad332e5ca9503bba2e14cc8c3ca675eed56d72bc/django/forms/utils.py#L26) will do it, but apparently that's not public, so probably better to use format_html_join directly instead...)

timheap: (We already use `flatatt` in at least seven other places in Wagtail, one more wouldn't hurt)

ababic: @timheap Ah, I'd already took the `format_html_join` route before I saw your comment.

ababic: Only just discovered that `flatatt` uses `format_html_join` 😜 completely missed that reference in your earlier comment @gasman, sorry

PR #3065, opened 2016-10-10
Allow `Orderable` models to be reordered from modeladmin's IndexView

ababic:
A solution to #2941.

- Adds a new `mixins.py` file for Mixin classes, to prevent `options.py` from becoming too busy
- Adds the `OrderableMixin` class to `mixins.py`
- Updates modeladmin tests to use the `OrderableMixin` class in an example


ababic:
The automated drone test here is complaining about the two new mixin classes being imported into `options.py`. I'd done that intentionally so that anybody attempting to import `ThumbnailMixin` from it's old location would be covered, and then added `OrderableMixin` for consistency.

Does anyone have any ideas for handling this in a better way?

zerolab:
One way would be to add `#noqa` inline and an explanatory comment.

The proper way would be add deprecation warnings. I remember @kaedroho doing this in several places before.

ababic:
Cheers @zerolab. I'll take a look through the codebase and see if I can find something to copy from.

ababic:
This is ready for review, now :)

ababic:
Apologies @kaedroho/@gasman. Still not got the hang of this rebasing thing. Have submitted a new PR (#3103) with just the relevant changes.

gasman:
Superseded by #3103

PR #3068, opened 2016-10-11
Remove explorer, page search for users with no page permissions

timheap:
For a site we are making, we have two classes of users: content editor who can edit pages and other content, and researchers who can only edit some specific models through modeladmin views. This PR removes the explorer menu item and the ability to search and browse pages for users without any page editing / moderating permissions. It also changes the search in the menu to use the first available SearchArea, instead of hard coding the page search area.

Removing the explorer completely seems like it could be a controversial change. Wagtail, until recently, was all about making pages after all. Now, with the advent of model admin and other plugins that work with non-page models, making Wagtail work without pages makes sense. With the current Wagtail, a user with model admin permissions but without normal page permissions is presented with an interface that is still oriented towards pages. The first thing in the menu is a search page that searches pages. The second thing in the menu is an explorer with all of the pages. They can not do anything with these pages, so these items are mostly useless. They can aimlessly browse the explorer looking for something to do, but there is nothing. Removing the explorer in this case is user friendly. The same with the search. Any page search results are useless to them, and probably not what they are looking for anyway. Making the menu search obey the SearchArea order and pick the first `is_shown` item (or hiding itself completely if there are no `is_shown` items) helps the user again by removing functionality that is not useful to them.

Currently, it is not possible to restore the ability to browse pages to a user without other page permissions. I am not sure what the point of allowing this would be, certainly considering that some of the functionality is broken in this use case (the page copy view specifically), so I do not think that is an issue.

### DIFF
--- a/wagtail/wagtailadmin/site_summary.py
+++ b/wagtail/wagtailadmin/site_summary.py
@@ -2,6 +2,7 @@ from __future__ import absolute_import, unicode_literals
 
 from django.template.loader import render_to_string
 
+from wagtail.wagtailadmin.utils import user_has_any_page_permission
 from wagtail.wagtailcore import hooks
 from wagtail.wagtailcore.models import Page, Site
 
@@ -17,6 +18,9 @@ class SummaryItem(object):
 
     def render(self):
         return render_to_string(self.template, self.get_context(), request=self.request)
+
+    def is_shown(self):
+        return True
 
 
 class PagesSummaryItem(SummaryItem):
@@ -40,6 +44,9 @@ class PagesSummaryItem(SummaryItem):
             'total_pages': Page.objects.count() - 1,  # subtract 1 because the root node is not a real page
         }
 
+    def is_shown(self):
+        return user_has_any_page_permission(self.request.user)
+
 
 @hooks.register('construct_homepage_summary_items')
 def add_pages_summary_item(request, items):
@@ -57,6 +64,10 @@ class SiteSummaryPanel(object):
             fn(request, self.summary_items)
 
     def render(self):
+        summary_items = [s for s in self.summary_items if s.is_shown()]
+        if not summary_items:
+            return ''
+
         return render_to_string('wagtailadmin/home/site_summary.html', {
-            'summary_items': sorted(self.summary_items, key=lambda p: p.order),
+            'summary_items': sorted(summary_items, key=lambda p: p.order),
         }, request=self.request)
--- a/wagtail/wagtailadmin/templates/wagtailadmin/base.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/base.html
@@ -11,14 +11,7 @@
                 <span>{% trans "Dashboard" %}</span>
             </a>
 
-            <form class="nav-search" action="{% url 'wagtailadmin_pages:search' %}" method="get">
-                <div>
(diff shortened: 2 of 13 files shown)


gasman:
Thanks @timheap! Absolutely agreed that removing the explorer for users without page permissions is the right thing to do - the fact that we don't do that (as we do for images / docs) has been a low-level annoyance for me for some time, and as you say, it's going to become increasingly important as people make greater use of modeladmin and the like.

timheap:
Will this still be relevant with the addition of the React Explorer? Can this code / the idea be copied in to that PR? Or should I rebase this on master so it can still be merged before the React Explorer?

gasman:
Hi @timheap - Yes, I believe this is still very much relevant, and as far as I can see, the code change here is still applicable (and desirable) with or without the React explorer. (If I'm not mistaken, the React explorer attaches to the menu item through an onclick event, just as the current explorer menu does; if the menu item is hidden as per this PR then the React explorer simply has nothing to trigger it.) Sorry for the delay in reviewing - it's nearing the front of the backlog queue now!

thibaudcolas:
Hey there – the new version of the explorer is nearing completion so I can now confirm that this will still work the same. I haven't tried to combine this PR + the explorer one completely but I've tried hiding the explorer menu item (as this PR does) and the result is what is expected.

There are two caveats:

- This PR applies `@user_passes_test(user_has_any_page_permission)` to `explorer_nav` – as far as I know this view will be gone (see #3385 for the delete diff)
- I don't think this the same decorator is applied to the API which the explorer uses, so theoretically a user could still query the API somehow and view pages they do not have permissions on. Note that this is just theoretical – and I'm not familiar enough with the Wagtail API to tell how permissions work or should work with it.

gasman:
> This PR applies @user_passes_test(user_has_any_page_permission) to explorer_nav – as far as I know this view will be gone (see #3385 for the delete diff)

Yep, correct.

> I don't think this the same decorator is applied to the API which the explorer uses, so theoretically a user could still query the API somehow and view pages they do not have permissions on. Note that this is just theoretical – and I'm not familiar enough with the Wagtail API to tell how permissions work or should work with it.

In the current architecture of Wagtail, there's nothing to stop users from finding out about pages that they don't have edit permission on - for example, they still show up in link chooser modals - so it's appropriate for the API to include those pages, I think. (In the case of the API there's a bit more information leakage, since you can get the full page content rather than just the titles - but addressing that would involve introducing a concept of "this field is visible in summaries / listings", and that's something for another day.)

When we implement 'can choose' permission (for proper multi-tenancy support where users have no knowledge of the other sites on the system), it would make sense to lock down the API to only show pages for which the user has choose permission.

To be fully correct, the API ought to support filtering by permission, since the explorer menu and listing should only include pages that you can do something meaningful with (edit, publish, lock). Luckily, that's currently covered by the use of `get_explorable_root_page` to set the visible root, because permissions always propagate to descendant pages. (If we ever introduce the ability to un-assign permissions for specific subtrees, we'll have to revisit this and do it properly...)

gasman:
Merged in 72a1b82e3ed120eff46d87cea2a2bf1d4ad7cbc8 (with one tiny addition - added tests for copy_page for both a user with no page permission and a user with unrelated page permissions, as those are now separate code paths).

Thanks @timheap, and sorry for the delay in reviewing!

timheap:
Thanks for the merge!

PR #3069, opened 2016-10-11
Makes submenu collapsibile in mobile and animate submenu chevron in desktop

carloratm:
Ok, first of all I have to apologize because I left a PR in the limbo for a lot of time (https://github.com/torchbox/wagtail/pull/1926). I had problems with that account so I am opening this one which does exactly the same.

I animated the submenu chevrons on both mobile and desktop views.
I also animated the submenu itself on mobile.

I used the max-height technique to animate the submenu. I am not a fan of that, considering
that it is based on a magic number, and a lot of site settings can increase the height of submenu.
I would prefer the submenu to not animate on mobile. What do you think @davecranwell ?

Cheers,
Carlo

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_main-nav.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_main-nav.scss
@@ -121,10 +121,19 @@ $footer-submenu: $submenu-color;
     .account {
         display: none;
     }
+
+    .submenu-trigger:after {
+        width: auto;
+        transition: transform .3s ease;
+        content: 'q';
+    }
 }
 
 .nav-submenu {
+    transition: max-height .2s;
     background: $submenu-color;
+    overflow: hidden;
+    max-height: 0;
 
     h2 {
         display: none;
@@ -150,6 +159,17 @@ $footer-submenu: $submenu-color;
     .wagtail-version {
         padding: 1.2em 1.7em;
         text-align: center;
+    }
+}
+
+.submenu-active {
+    .submenu-trigger:after {
+        transform-origin: 50% 50%;
+        transform: rotate(180deg);
+    }
+
+    .nav-submenu {
+        max-height: 1000px;
     }
 }
 


wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_main-nav.scss
```diff
@@ -121,10 +121,19 @@ $footer-submenu: $submenu-color;
     .account {
         display: none;
     }
+
+    .submenu-trigger:after {
+        width: auto;
+        transition: transform .3s ease;
```
thibaudcolas: Could you use the `transition` mixin instead of the property directly? eg. `@include transition(transform .3s ease);`. The mixin isn't documented unfortunately, but I would like to keep the pattern going.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_main-nav.scss
```diff
@@ -121,10 +121,19 @@ $footer-submenu: $submenu-color;
...
+        content: 'q';
+    }
 }
 
 .nav-submenu {
+    transition: max-height .2s;
     background: $submenu-color;
+    overflow: hidden;
+    max-height: 0;
```
thibaudcolas: Please remove submenu animation from this PR. As you've said, there are trade-offs in animating the `max-height` so we should treat this as a separate change (that will need discussion).

thibaudcolas:
Merged manually in `master` at https://github.com/wagtail/wagtail/commit/8c0170bb62283e3cdee08f2e98aa662152cbb418, with a few more changes:

- Removes the submenu animation.
- Uses the transition mixin.
- Adds the animation for the explorer menu (the PR was created for its previous version).
- Updates the `ExplorerToggle` tests, which weren't correctly set up before.

Re-tested in Chrome/FF/Safari on macOS 10.13, IE11, Edge16, Chrome Android, iOS Safari.

PR #1450, opened 2015-06-28
Add page privacy by user/group

tmiller02:
Fixes #257.

Adds an extra option to the page privacy settings which allows the user to choose which users and user groups can view the live page.

PR #1076, opened 2015-03-19
Page privacy groups

smakinson:
Here is a take on how restricting pages to users in specific groups could work. This builds on the same flow that was already in place for page restrictions and adds a groups option to the PageViewRestriction model. The login process uses the normal django auth. Please take a look and let me know what you think and if I missed anything. Thanks!

Issue #257, opened 2014-05-23
Page View Permission


wdv4758h:
At this moment, the published wagtail page can be seen by everyone.

How about a way that can limit the people who can see the specific page (people who want to see that page should login and be in specific role) ?

davecranwell:
Since this issue was first created we've added password protection but it would still be a nice feature to restrict view access to a page, optionally recursively, to a certain user, or to a certain user role/group.

The interface could be similar to Wordpress's approach, where protection could be enabled in one of three ways offered as a set of radio buttons: password (input box for password), specific user (modal to choose user), user group/role (modal to choose group(s))

smakinson:
Has there been any more discussion on this? Its a feature I am looking for. I was thinking about the page privacy modal having a 3rd radio for private to groups and perhaps it would swap out checkboxes with the password field.

davecranwell:
There hasn't, I'm afraid. It's no less on our radar but we prioritise features based on their demand (particularly demand from our clients) so this has the effect of pushing less requested features back.

smakinson:
I am looking into adding this feature. Is vagrant required to run the tests?

davecranwell:
A virtual machine is probably the easiest way for the uninitiated to get up and running but you may have more experience. I don't believe the tests require vagrant at all.

kaedroho:
@smakinson Nope, see my comment on another ticket (https://github.com/torchbox/wagtail/issues/1004#issuecomment-74836009) which shows how to run the tests with just a virtual environment (no postgres/elasticsearch/vagrant) required.

smakinson:
Great, thanks! That testing info looks like what I needed.

shacker:
Looking over the two outstanding PRs re: improving access control for public/private pages, I'm noticing that both of them add a third option based on Users/Group memberships. My org's use case would be a  fourth (and simpler) option, restricting the current hierarchy to logged-in users only (in other words, the site mixes public and private content and needs a way to distinguish access between them). Users may not be in any group at all, so group membership is not quite the right solution.  So, something like:

```
RESTRICTION_CHOICES = (
       (NONE, _("Public")),
       (PASSWORD, _("Private, accessible with the following password")),
       (USERS_GROUPS, _("Private, visible to specific users and groups")),
       (AUTHENTICATED_ONLY, _("Private, visible to any logged-in user"))
   )
```

[...]

PR #3071, opened 2016-10-11
Page privacy by user login

gasman:
Implements #257. I've cherry-picked code, ideas and tests from the existing pull requests #1076, #1450 and #2362, along with some new configuration and docs as proposed in https://github.com/torchbox/wagtail/issues/257#issuecomment-248401768 to ensure that this works acceptably well on both the out-of-the-box setup, and integrating Wagtail into an existing Django project.

Note that I've left out the ability to assign permission to individual users (which was implemented in #1450 and #2362) - this is because it's entirely possible (and expected) for a Wagtail installation to have thousands of user accounts, and with a standard checkbox-based form this would make the interface unusable. Given a nicer UI widget (such as an autocompleting text field) we could certainly add this as a future extension, but I don't think this is necessary for a first version.


tomdyson:
> Note that I've left out the ability to assign permission to individual users

I'm very comfortable with this. Administrators can create single-user groups if necessary - this is a standard approach for CMS permission management.

docs/advanced_topics/privacy.rst
```diff
@@ -3,9 +3,33 @@
...
+Users with publish permission on a page can set it to be private by clicking the 'Privacy' control in the top right corner of the page explorer or editing interface. This sets a restriction on who is allowed to view the page and its sub-pages. Several different kinds of restriction are available:
 
-Private pages work on Wagtail out of the box - the site implementer does not need to do anything to set them up. However, the default "password required" form is only a bare-bones HTML page, and site implementers may wish to replace this with a page customised to their site design.
+ * **Visible if logged in:** The user must log in to view the page. All user accounts are granted access, regardless of permission level.
+ * **Accessible with the following password:** The user must enter the given password to view the page. This is appropriate for situations where you want to share a page with a trusted group of people, but giving them individual user accounts would be overkill. The same password is shared between all users, and this works independently of any user accounts that exist on the site.
```
tomdyson: Why 'Accessible' and not 'Visible'?

wagtail/wagtailcore/wagtail_hooks.py
```diff
@@ -1,8 +1,17 @@
...
+from django.conf import settings
+from django.contrib.auth.views import redirect_to_login
 from django.core.urlresolvers import reverse
 
+from wagtail.utils.compat import user_is_authenticated
 from wagtail.wagtailcore import hooks
+from wagtail.wagtailcore.models import PageViewRestriction
+
+
+def require_wagtail_login(next):
```
tomdyson: Might users want to override `settings.WAGTAIL_FRONTEND_LOGIN_URL` on a per-section basis?

docs/advanced_topics/privacy.rst
gasman: "Accessible" is the wording [used on the existing label](https://github.com/torchbox/wagtail/blob/84c36c1ec2eca8fd51028e7c46a6046ba64c8dbd/wagtail/wagtailadmin/forms.py#L185) - the two "visible" options are the new ones.

I'm torn about which word is better to standardise upon - "visible" sounds better, but "accessible" is probably more technically correct, because a protected page can still be "visible" in the sense that it appears in menus (unless the developer has specifically applied a `public()` filter on the queryset). "Viewable"? Perhaps I'm overthinking this :-)

wagtail/wagtailcore/wagtail_hooks.py
gasman: For a multi-domain site (which is probably the most common case where you'd want to vary the login page based on section) you could check `request.site` within the login view / template - or, more neatly, use wagtail.contrib.settings - since the redirect to the login page will keep you on the same domain. In other cases, you could (hackily) inspect the 'next' parameter.

I can't see a particularly nice way of making the URL switchable as a feature provided by Wagtail itself (where 'nice' = 'not hard-coding a list of URL paths or IDs in the settings file')...

I did consider making this configurable on a per-page-type basis, as we do for password-based restrictions - I couldn't think of a use-case for that at the time, but I guess paywalls are the big one (display the first paragraph of an article, then "log in to view the rest of this article"). I left that out in the interest of keeping this an MVP, though.

docs/advanced_topics/privacy.rst
tomdyson: Let's go for 'accessible'. Wagtail should work for screenreaders too!

m1kola:
It's a bit confusing behaviour: if you have an account with editor role, but a page requires to be a moderator to see the page - it redirects you to a login view. Even if you already logged in.

m1kola:
Looks good except a bit strange [behaviour with login redirect](https://github.com/torchbox/wagtail/pull/3071#issuecomment-256045521).

Maybe we can add a message to the login view that explains why we are asking already logged in user to log in again?

wagtail/wagtailcore/urls.py
```diff
@@ -18,9 +20,16 @@
...
 
+WAGTAIL_FRONTEND_LOGIN_TEMPLATE = getattr(
+    settings, 'WAGTAIL_FRONTEND_LOGIN_TEMPLATE', 'wagtailcore/login.html'
+)
+
+
 urlpatterns = [
     url(r'^_util/authenticate_with_password/(\d+)/(\d+)/$', views.authenticate_with_password,
         name='wagtailcore_authenticate_with_password'),
+    url(r'^_util/login/$', auth_views.login, {'template_name': WAGTAIL_FRONTEND_LOGIN_TEMPLATE},
```
gasman: I think that's right, yep. (I didn't cover this in the docs, as it's the standard django.contrib.auth functionality and is covered in Django's own docs.)

gasman:
@m1kola Sounds like a good idea - I'll add that. I deliberately left the "user is already logged in" case out of [my original plan for this feature](https://github.com/torchbox/wagtail/issues/257#issuecomment-248401768), as I didn't think it was necessary for an initial implementation, and I assumed it would require an additional view / template / setting variable...

> (Ideally we'd have a separate code path for users who are authenticated but don't have the required permission - but I'm willing to leave that out in the interest of keeping this an MVP...)

...but actually, the Django docs (https://docs.djangoproject.com/en/1.10/topics/auth/default/#django.contrib.auth.views.login) have a nice example of doing it with a few extra lines on the template, so I'll do the same.

gasman:
Message added in c5a1a66f815f1c46f02bbc4e04481e4fb25c1629.

m1kola:
Everything looks good to me! Matthew, thank you for the cool feature.

Merged in b4d17d8377ec9c93d03d2c2041dcce643e5f5f46 + parents.

PR #3072, opened 2016-10-12
Intelligently select a parent in the page chooser modal

timheap:
This PR improves the page chooser modal when the valid page types are restricted. In the case of selecting an `EventPage`, all of which are under an `EventIndex` page, the modal will open at the `EventIndex` page, saving the editor much clicking.

Each commit adds one feature, and comes with its own tests and docs.


gasman:
Looks good to me, thanks! Merged in 2442416e93a9cec4325a41894279157cee4e3327 + parents.

Issue #666, opened 2014-10-03
Snippet chooser UI improvements


jimmymorrisuk:
There are a few parts to this so I will break them down. 

1) There's no title text on the snippet reference buttons. Although I know what the trash icon is meant to mean the up and down buttons, to some users, will be ambiguous, I had to click them to be sure. I recommend adding title text to all icon only buttons. 
It could also be argued that the trash button is incorrect. I am not actually deleting or trashing anything and certainly not a snippet. I am actually removing the reference, so I think an cross might be a better icon.

2) The button under the snippets marked 'Add X snippet' could also be misleading. The question this produces is if I click this will I be able to create a X snippet or simple add another reference. A solution to this is to change the label to read 'Add snippet reference'. 

3) When on a snippet reference section what happens if I want to add a new snippet on the fly?
In the long run it would be nice to have a modal window that loads in the add snippet form, this would preserve the users' workflow context. A quick win would be to add link that opens the correct snippet form in another window.

davecranwell:
1) fixed in https://github.com/torchbox/wagtail/commit/6a36ed7f7ed3790784733ee80bdf5ee4587edf2a
2) Will consider
3) Duplicate of #396
4) Will consider, however this is pretty much part of django.
5) duplicate

PR #3074, opened 2016-10-12
Style / consitency improvements for modeladmin's IndexView

ababic:
### Bug fixes

- Removes right padding from result list when listing is supposed to be full-width (no filters) on desktop
- Removes an [invalid rule](https://github.com/torchbox/wagtail/blob/master/wagtail/contrib/modeladmin/static_src/wagtailmodeladmin/scss/index.scss#L13)

### In order to be more consistent with the Explorer list view:

- Use `#666` as the general text colour
- Only make orderable column header links 'teal' if the listing is currently ordered by that field
- Removes unnecessary left/right padding from result list on mobile (the table cells have padding of their own)

### Other improvements (open to debate)

- Removes `text-transform:uppercase` from filter links
- Adds subtle zebra-striping to rows, to help with list scanning
- Better use of scss element hierarchy to improve readability


kaedroho:
Before:
3074_before

After:
3074_after

kaedroho:
I just had a go at pulling in the styling of the explorer page to see how it looks:
 - removed the zebra background
 - darkend the text further to ``4d4d4d``
 - increased font weight to 600
 - increased font size to 1.15em

3074_after2

ababic:
Hi @kaedroho. Happy to lose the zebra striping, if you'd rather the UI remain cleaner. I felt compelled to do something in order to make use of the `odd` and `even` classes already being added to each row. What do you think to cutting out the code that adds those classes? Can you see them ever being used?

In regards to:

> It also just occurred to me that the titles should link to the edit page (this is how the explorer and other listings behave) 

I think I did this in an early iteration. But, unlike the explorer, where the output is completely predictable, the options available in ModelAdmin make it possible to output pretty much anything in each column, including HTML with links in, so it needs a bit more thinking to do nicely. 

The simplest approach I can think of, is to add a new attribute to the `ModelAdmin` class (say `link_button_column_text`) that is `True` by default, but allows developers to turn that behaviour off in very rare cases? I guess a 'fool proof' option would be to check whether the value is HTML or a simple text value, and conditionally add the link based on that. Any preference on what route to take there?

ababic:
@kaedroho To illustrate my point about the unpredictability of column content, below is a screen from a project I'm working on at the moment: 

screenshot 2016-10-14 at 08 49 29

So, with that in mind, I'd probably say that enforcing links, or font-weight for any one column is probably a little too prescriptive.

ababic:
@kaedroho I've removed any background-color / text color overrides from the IndexView styles, so hopefully, if there are any legibility improvements in the core styles, they should follow through to the IndexView also, just as any font-size changes should (I don't think I've ever overridden font-size anywhere)

Zebra stripes are also gone.

The 'edit link' addition turned out to me more complicated than first though, which led to a (possibly out of scope for this pull-request) change to the template tags and where certain logic belongs. I haven't tested performance, but I'm guessing that the dropping of a template tag out of the process entirely for rendering columns will have some decent performance gains?

ababic:
@kaedroho @gasman FYI, because of the templatetag changes, this pull will likely conflicts with #3064. I'd recommend merging that one first, then I can rebase and resolve any conflicts.

ababic:
In furtherance to this, I'd really like to abstract out the button rendering, so that the buttons (and possible a set of buttons) know how to render themselves, rather than having to build things in the tag. That will be a considerable update on it's own, though, so likely better tackled in a separate PR.

SalahAdDin:
@ababic , awesome!

ababic:
Apologies @kaedroho. The rebase didn't go quite to plan. Have instead pushed a new PR with just the relevant changes.

I didn't hear back from you RE the zebra stripes and 'odd'/'even' classes, so I've removed them both.

gasman:
Superseded by #3102

SalahAdDin:
Custom edit handlers aren't supported yet, right? :'(

PR #3075, opened 2016-10-13
Move `get_image_model`, add `get_image_model_string`

timheap:
Move get_image_model, add get_image_model_string
    
Having `get_image_model` at `wagtail.wagtailimages` is consistent with `django.contrib.auth.get_user_model`, and means developers can import `get_image_model` in an environment where models are not yet ready.
    
`get_image_model_string` has been added for uses where the model itself is not required, and might not be available, but a reference to the possibly swapped out Image model is required.

### DIFF
--- a/docs/advanced_topics/images/custom_image_model.rst
+++ b/docs/advanced_topics/images/custom_image_model.rst
@@ -1,5 +1,6 @@
 .. _custom_image_model:
 
+===================
 Custom image models
 ===================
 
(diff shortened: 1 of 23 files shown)


timheap:
Whoops, I had forgotten the docs for this. Added now.

gasman:
Merged in 0cf3b13a1d54528237552ee37ee2eeb3a13d4e1d + parents

timheap:
:+1: Thanks

PR #3077, opened 2016-10-13
Cache the compiled templates for FieldBlock and StreamBlock form fields

gasman:
As per https://github.com/torchbox/wagtail/issues/3062#issuecomment-253264260 - cache the compiled templates for StreamBlock and FieldBlock after first use, so that they don't have to be recompiled every time. This leads to a ~15% performance improvement on edit pages with complex StreamFields.

### DIFF
--- a/wagtail/wagtailcore/blocks/field_block.py
+++ b/wagtail/wagtailcore/blocks/field_block.py
@@ -4,7 +4,7 @@ import datetime
 
 from django import forms
 from django.db.models.fields import BLANK_CHOICE_DASH
-from django.template.loader import render_to_string
+from django.template.loader import get_template
 from django.utils import six
 from django.utils.dateparse import parse_date, parse_datetime, parse_time
 from django.utils.encoding import force_text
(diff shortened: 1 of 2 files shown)


jaap3:
Wouldn't this be super annoying if you're editing one of these templates? This would force you to restart the server each time you've edited a template.

Django ships a template loader that caches compiled templates (so not the rendered result of a template, but the compiled node tree) https://docs.djangoproject.com/en/1.10/ref/templates/api/#django.template.loaders.cached.Loader

This loader should be enabled in production on any serious Django deploy anyway (and will be by default (in some cases) in a future Django version: https://code.djangoproject.com/ticket/25788)

gasman:
@jaap3 I think it'll be very rare for anyone to edit these templates, and only if they're hacking on Wagtail core. (If a user wants to customise it for their site, they're more likely to subclass FieldBlock / StreamBlock instead.) That's a fair point about the cached template loader, though - there's no point in optimising here if Django already has an optimisation for the same thing. I'll test the performance of this PR against the cached template loader, and close this PR if there's no meaningful difference.

gasman:
With the cached template loader enabled, there's a significant performance improvement, and no visible difference with or without this patch -> closing.

jaap3:
Should a note about using the cached template loader be added to the Performance section of the Wagtail docs?

gasman:
@jaap3 Sure, sounds like a good idea!

PR #3079, opened 2016-10-13
Remove useless SelectorFormat rule in scss-lint

gasman:
Prompted by this Drone CI failure:

https://github.com/torchbox/wagtail/pull/3035/commits/0098d6a21fe06a017d6e925967a831328f17702d
http://46.101.88.142/torchbox/wagtail/438

As the existing comment admits, we're de-facto _not_ adopting the hyphenated-BEM convention in our CSS, due to the auto-generated underscored class names on form widgets - so having that rule in place at all (with an ever-growing list of files that are allowed to break it in order to do their job) is a waste of time.

### DIFF
--- a/.scss-lint.yml
+++ b/.scss-lint.yml
@@ -86,22 +86,11 @@ linters:
     SelectorDepth:
         enabled: false
 
+    # There is currently no de facto standard for selector format in Wagtail, due to:
+    # a) the conversion of django field/widget names to underscored class names
+    # b) the use of third party code such as Hallo.js which uses classes with snakeCaseClasses.
     SelectorFormat:
-        convention: hyphenated_BEM
-        ignored_names:
-            - js_class
-        ignored_types:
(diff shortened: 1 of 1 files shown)


kaedroho:
> As the existing comment admits, we're de-facto not adopting the hyphenated-BEM convention in our CSS

What does this mean exactly? Is this a proclamation that Wagtail will not be switching to BEM, or are you only saying that we're not going to enforce using it in CI?

gasman:
@kaedroho It's not a statement of intent... it's an observation that we don't currently comply with BEM, and until we do (which doesn't seem likely to happen any time soon), we shouldn't have a linter rule that prevents us from doing perfectly legitimate things in CSS.

gasman:
Merged in 7d3dcc0add3ae9f93e0898ffb36ce76612b0ccb8 as part of #3035

Issue #1602, opened 2015-08-12
Support for PROTECT on foreign keys


jorge-marques:
Are there plans to support `on_delete=PROTECT` in the admin? Right now ProtectedError is not caught.

gasman:
There are no current plans for it - we've found that on_delete=CASCADE and on_delete=SET_NULL are sufficient for our needs. If you think this would be a useful addition, we'd happily consider a pull request for it, though.

tomdyson:
We're experiencing this (failure to catch `ProtectedError`) on client projects.

prydie:
:+1:

ababic:
We were experiencing this recently too, so I added some handling for it into wagtailmodeladmin's delete view. However, It's only triggered after user confirmation, when the failed delete raises the exception. I suppose it'd be more helpful to flag it up before that and not even allow an attempt to be made. 

For me, that idea opens up a can of worms, though. If I were to provide up-front feedback about protected relationships, surely it's just as (if not, more) important to flag up ones using CASCADE also? And what about  relationships using SET_NULL? should they at least get a mention? What if the instance has relationships with various other objects using all of those methods? It was all bit much for me to tackle alone.

pacotole:
I made a PR to catch ProtectedError on image delete.
This is the link if it cant help you https://github.com/torchbox/wagtail/pull/3082

hakjoon:
Is this issue still accurate given the `wagtailcore.W001` check which recommends using `models.PROTECT` and to explicitly not use `CASCADE`? [wagtailcore.W001](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailcore/models.py#L558)

gasman:
@hakjoon Yes, it's still relevant. The warning suggests using PROTECT or SET_NULL; PROTECT is definitely a better choice than CASCADE (because a page deleted through CASCADE will corrupt the tree) but still isn't ideal because it throws an uncaught exception. (Also, the warning only applies to pages; the issue here is relevant to snippet / modeladmin models too.)

hakjoon:
@gasman Sorry I think I framed my question poorly.  I was wondering if the current position is that there are no plans to support `PROTECT` given the message `wagtailcore.W001` presents.  Also I was wondering about the statement that `CASCADE` is sufficient, given the messaging that `CASCADE` should not be used.

[...]

PR #3082, opened 2016-10-13
[FIX] ProtectedError on image delete

pacotole:
On admin an image delete with protected foreign key trigger a 500 server error
With this fix catch ProtectedError and show a messagge with a link to edit the
related protected object

### DIFF
--- a/wagtail/wagtailimages/views/images.py
+++ b/wagtail/wagtailimages/views/images.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, unicode_literals
 import os
 
 from django.core.urlresolvers import NoReverseMatch, reverse
+from django.db.models import ProtectedError
 from django.http import HttpResponse, JsonResponse
 from django.shortcuts import get_object_or_404, redirect, render
 from django.utils.translation import ugettext as _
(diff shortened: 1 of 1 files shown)


zerolab:
Fixes #1602

gasman:
Thanks @pacotole! Could you add a test for this, please?

wagtail/wagtailimages/views/images.py
```diff
@@ -228,9 +229,18 @@ def delete(request, image_id):
...
-        return redirect('wagtailimages:index')
+        try:
+            image.delete()
+            messages.success(request, _("Image '{0}' deleted.").format(image.title))
+            return redirect('wagtailimages:index')
+        except ProtectedError as e:
+            item = e.protected_objects[0]
+            messages.warning(request, _(
+                "Image '{0}' can't be deleted because is in use in '{1}'."
+            ).format(image.title, item.title), buttons=[
```
timheap: This will only work if the item is a Page. If the item was a snippet, or some other model without a `title` attribute, this would cause an error.

wagtail/wagtailimages/views/images.py
```diff
@@ -228,9 +229,18 @@ def delete(request, image_id):
...
+                messages.button(reverse('wagtailadmin_pages:edit', args=(item.id,)),
```
timheap: Same kind of problem here. This assumes that the item is a page. If it was a snippet, then the URL would be wrong. If it was not editable in the wagtail admin, there is no way of making a nice button for it at all.

gasman:
Now superseded by #3961

Issue #3080, opened 2016-10-13
An invisible parent link in search results in a page chooser on hover


m1kola:
When you search for a page in a page chooser, you can see a link to the page's parent. This link doesn't change colour on hover.

--

Without hover:

<img width="883" alt="screen shot 2016-10-13 at 15 23 37" src="https://cloud.githubusercontent.com/assets/509198/19348935/17cb3d80-9159-11e6-8309-4c2d5c148e67.png">

With hover on row:

<img width="903" alt="screen shot 2016-10-13 at 15 18 11" src="https://cloud.githubusercontent.com/assets/509198/19348823/88692ac6-9158-11e6-8270-44be3116cfd8.png">

With hover on parent link:

<img width="892" alt="screen shot 2016-10-13 at 15 26 09" src="https://cloud.githubusercontent.com/assets/509198/19348991/6553bf8c-9159-11e6-8e4d-0fe67579d2ee.png">

gasman:
Fixed in #3083. It's just occurred to me that we don't actually _want_ a link that navigates away here... but I'll open that as a separate issue :-)

PR #3083, opened 2016-10-13
Fix page chooser parent link on row hover [#3080]

zerolab:
Quick fix for #3080


gasman:
Thanks! Merged in e0e957f92aea3c1a69adbbff38b7773b1abc2b73

Issue #278, opened 2014-06-02
Search in the image chooser does not consider tags when using the database search backend


pcraston:
Currently it seems to only search by the image title - it would be great if it would be possible to also search by tag and show all images with a certain tag.

mhnbcu:
:+1:

davecranwell:
As far as I'm aware image tags are indexed like any other text attributed to an image (such as title). Although you don't get any kind of auto-compete interface to choose a tag when doing a text search, any search term you use will search for tags with those words used. This feature may however depend on Elasticsearch. 

Tests to ensure searching by image tag worked were written here: https://github.com/torchbox/wagtail/issues/613

@pcraston are you able to confirm this is still the case with a more recent version installed? Given this ticket is quite old, we'll close it unless this can be reproduced in master.

pcraston:
We haven't had time to upgrade wagtail for while but are planning to update our project soon. I'll check once we have the latest version running.

gasman:
In #973 @mhnbcu is (indirectly) reporting that this behaviour is still present in Wagtail 0.8.4, so let's keep this issue open for further investigation.

olifante:
any update on this? I'm using Wagtail 1.1 and this issue is still present. It's quite counter-intuitive to provide the ability to tag images and not allow admins to search by tag.

jorge-marques:
:+1:

jorge-marques:
It would be nice if we could search pages by tag as well.

gasman:
This one confused me for a while during final testing of Wagtail 1.2 before I figured out what was going on. Searching on tags only currently works when using the Elasticsearch search backend, not the database one <http://docs.wagtail.io/en/v1.2/topics/search/backends.html>. This fact isn't at all obvious, and it's probably become worse in recent releases as the project template has been stripped back to a 'vanilla' setup with no mention of Elasticsearch.

It's particularly unfortunate that clicking on 'popular tags' in the image chooser relies on having functional tag search (#973) - in other words, this is one feature that's visibly broken on a default Wagtail install. Let's get this fixed for 1.3 - if we can't make tag search work with the db backend, we should at least update the 'popular tags' listing to use plain SQL filtering rather than going through search.

gasman:
Have implemented a workaround for the 'popular tags' listing on images in #2026, so #973 is now addressed. Will keep this open because ideally we do want tags to be considered when running a regular search using the Postgres backend (otherwise there's no way to filter on non-'popular' tags).

[...]

PR #3084, opened 2016-10-14
Issue 278 search images and documents by tag (in db backend)

ashiazed:
Fix for #278 

We should probably revert https://github.com/torchbox/wagtail/pull/2026

Not sure if this needs documentation, it's meant to be used with the DB backend so
it really only applies to the Image and Document models. DB search backend doesn't
search subclasses of Page anyways. I suppose if someone wants to they can use it on
their own Indexed model but it will cost them performance as there is no queryset 
optimization going on.


gasman:
Fixed with the new database search backend in Wagtail 2.15.

PR #3087, opened 2016-10-15
should use rest_framework not restframework

pure-zero:


kaedroho:
Thanks!

PR #3088, opened 2016-10-15
Allow form page form fields to be used as variables in templates

ashiazed:
Django doesn't allow dashes in template variables but the formpage form field names have them. Sometimes a form layout needs to be customized but referencing
{{ form.your-name }} produces an error.
This commit just replaces the dashes with underscores to prevent the error.

Does this need a test? I may need some guidance.
The only impact I can think of might be if someone is referencing the field names in css or js, in which case it would be worth adding to the release notes.


m1kola:
Hi Ashia,

Thanks for contributing.

I think it's a good idea to allow form fields be used in templates, but I'm afraid that changes in the `AbstractFormField.clean_name` method may affect form submission data that stored in a database (see the `FormSubmission` model).

We need to create a data migration for existing data... Or solve this in another way.

Any ideas? :)

laidibug:
Ah right, didn't think of that. Data migration seems like the way to go. As long as we know that all "-" can be safely replaced.

gasman:
(cc @spapas)

I agree that we should update the logic for converting form field labels into identifiers, both for the reason described here (ability to use the identifiers in template code) and in #3311 (we shouldn't be depending on `unidecode`, because it's GPL licensed) - but obviously we need to do it in a way that doesn't break references to existing identifiers in form submission data.

A mostly-complete proposal for how we can do that:

* Add a non-editable `clean_name` (or simply `name`) field to the `AbstractFormField` model, which gets populated on `save()` according to our preferred label-to-clean-name algorithm (which will use underscores instead of dashes, and not have an absolute dependency on unidecode - although it could still optionally support unidecode, if we decide that's a desirable feature)
* Update the form handling to refer to this field, rather than computing clean_name on demand
* Provide a data migration that populates this field for existing forms according to the legacy (unidecode-and-dashes) algorithm. Since it's hard to perform data migrations on models that exist in user apps, this probably won't be an actual data migration... I think the best strategy might be to implement it as a startup system check, which looks for empty (null) `clean_name` fields on models descending from `AbstractFormField` and fills them in using the legacy algorithm. This process will need to check for the presence of unidecode (because the code will have to be left in place for some time after we've dropped unidecode as a dependency), and fail loudly if it isn't available.

spapas:
One more thing to consider: If unidecode is missing then the name and id of the generated form field will by something like ```<input id='id_το-όνομα-μου' name='το-όνομα-μου' />```. This seems to be valid in HTML5 (https://www.w3.org/TR/html5/dom.html#the-id-attribute) however is this the correct thing to do? I've never seen input id/name with unicode characters...

For the OP's original problem until it is fixed in wagtail I propose creating an assignment template tag ``get_formfield_by_name`` that will get a form field for further processing by its form and name (string) to overcome the dash problem in template variables. So instead of ```{{ form.your-name }}``` the template editor will be able to write ```{% get_formfield_by_name form "your-name" as your_name_field %}```. 

Actually this assignment tag could be added to wagtailforms to help people with similar problems.

lb-:
https://github.com/wagtail/wagtail/pull/6093 - builds on this PR but handles the case of existing data + unicode deprecation plans.

gasman:
Completed in #6093

PR #3090, opened 2016-10-16
Custom Page Manager documentation fix

nimasmi:
The `from_queryset` classmethod seems to be a factory, and returns a class, not an instance. The Django documentation example for this has an extra pair of brackets to instantiate the returned class. https://docs.djangoproject.com/en/1.10/topics/db/managers/#from-queryset

Copying the Wagtail documentation example verbatim causes: 

```
SystemCheckError: System check identified some issues:

ERRORS:
events.EventPage: (wagtailcore.E002) Manager does not inherit from PageManager
        HINT: Ensure that custom Page managers inherit from django.db.models.manager.BasePageManagerFromPageQuerySet
```

More info

``` python
>>> cls
<class 'events.models.EventPage'>
>>> cls.objects
<class 'django.db.models.manager.BasePageManagerFromPageQuerySetFromEventPageQuerySet'>
>>> issubclass(cls.objects, PageManager)
True
>>> isinstance(cls.objects, PageManager)
False
>>> isinstance(cls.objects(), PageManager)
True
```

### DIFF
--- a/docs/topics/pages.rst
+++ b/docs/topics/pages.rst
@@ -476,4 +476,4 @@ Alternately, if you only need to add extra ``QuerySet`` methods, you can inherit
     class EventPage(Page):
         start_date = models.DateField()
 
-        objects = PageManager.from_queryset(EventPageQuerySet)
+        objects = PageManager.from_queryset(EventPageQuerySet)()


gasman:
Thanks! Have gone with a two-step approach in a326cd855919e9e4092b1383b611bafc840019a5 for readability (and to match what wagtailcore does itself, e.g. https://github.com/torchbox/wagtail/blob/a326cd855919e9e4092b1383b611bafc840019a5/wagtail/wagtailcore/models.py#L251)

PR #3091, opened 2016-10-17
Small grammar mistake in french translation.

dperetti:


gasman:
Thanks! I've now updated this [on Transifex](https://www.transifex.com/projects/p/wagtail/), so it will be updated when we next fetch translations from there.

Issue #655, opened 2014-09-26
Consider "redirects as pages" notes from Ethan Jucovy


davecranwell:
http://www.coactivate.org/projects/ejucovy/blog/2014/05/10/wagtail-notes-managing-redirects-as-pages/

Displaying redirects in their position in the Explorer tree seems like a pretty good idea for starters. Other notes may also be worth investigating.

gasman:
Some assorted thoughts on this, as the subject of redirects has come up a couple of times today...

In many ways, the redirect-as-page approach is more elegant than wagtailredirects: they're more visible in the admin, have fewer moving parts (no need for middleware or a separate admin module) and can be used in situations where the source URL isn't a 404 (see #2533).

However, there are a couple of downsides that make it not a good fit for some scenarios:
* redirecting from `/a/deeply/nested/path/with/lots/of/slashes` requires setting up a bunch of nested pages
* you can't set up a redirect from a URL that contains non-slug-safe characters
* it doesn't have the special case handling for matching an exact query string first (`/myblog/show_article?id=123`) then falling back on a non-query-string match. You could make the `RedirectPage` model smarter to allow redirecting to different places based on querystring, but that's probably going to be much less intuitive than just treating the entire URL as a string match
* they're governed by the page permission model, and so you can't treat them as a different kind of "thing" for permission management purposes, e.g. specify that only superusers are allowed to create redirects. But arguably the page permission model can be a good thing too - it means you can say "this user can create redirects within `/departments/engineering/`.

All in all I think there's room for both concepts to exist in Wagtail (perhaps as two distinct contrib modules?) - although then we have the question of how to communicate that to users without looking like we've implemented the same thing twice in two incompatible ways. It strikes me that wagtailredirects is a better fit for preserving legacy URLs when migrating away from an old site, whereas redirects-as-pages is a better fit for shortcuts / directing a user around the site - so perhaps there's some way of naming them to represent those two roles.

PR #3094, opened 2016-10-19
Clusterable redirects

HalcyonChimera:

### DIFF
--- a/wagtail/wagtailredirects/models.py
+++ b/wagtail/wagtailredirects/models.py
@@ -3,9 +3,11 @@ from __future__ import absolute_import, unicode_literals
 from django.db import models
 from django.utils.six.moves.urllib.parse import urlparse
 from django.utils.translation import ugettext_lazy as _
+from modelcluster.fields import ParentalKey
 
 
 class Redirect(models.Model):
+
     old_path = models.CharField(verbose_name=_("redirect from"), max_length=255, db_index=True)
     site = models.ForeignKey(
         'wagtailcore.Site',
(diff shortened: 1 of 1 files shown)


gasman:
Hi @HalcyonChimera - please can you explain the purpose of this change?

HalcyonChimera:
I have the requirement of versioning the actual urls that point to a page; i.e if a page is moved, then a redirect is automatically created. With this I was able to use an InlinePanel so that redirects pointing at that page can be added/changed/removed from a given Page's promote panel. With only a ForeignKey, there is an error <PageModel> has no attribute 'redirect_set'.

gasman:
Thanks for the clarification! I have a feeling that this change will cause broken behaviour for existing users, though:

* A user creates and saves a page (this will result in an empty list of redirects being saved within the PageRevision record, regardless of whether the admin form includes an InlinePanel for redirects)
* They then set up redirects through the Settings -> Redirects area (which creates the redirect records immediately)
* They make further edits to the page, and publish it (- the edits are based on the most recent PageRevision record, which has an empty list of redirects - and so publishing the page will cause the existing redirects to be deleted)

This is probably a general issue that would affect any model that's editable in two places - as an independent model, and through an InlinePanel. In this case you're probably better off setting up an independent "inline redirect" model that works separately to the `wagtailredirects` module, with its own middleware.

HalcyonChimera:
The issue with having it separate is that it will be confusing for editors to have two places to look in order to determine the redirects that apply to a given page. Maybe we should compare the revision with other redirects created since it was last saved and try to replay the changes  into the revision record before it is saved and throw growls that say 'there are conflicting redirects'?

I feel like duplicating the entire redirect functionality here isn't necessarily the best use of time, and this is a feature I think it makes a lot of sense to provide; there is already the ability to change the slug of a page in the promote panel, for example. Having the ability to control those changes in the same location has a certain ergonomics to it.

What other edge cases might there be in terms of migrations and or chronologically dependent actions?

gasman:
The issue here is really that there are lots of different uses for redirects: preserving links when a page is moved, replacing an older site built on a different CMS, providing short/memorable links, doing tracking/analytics, providing a fixed link to changing resources (e.g. a monthly publication that has a redirect pointing to 'latest issue') and probably lots more that I haven't thought of. For some of those uses, it makes sense to tie redirects to a page's revisioning / publishing workflow; for others, it would be very much a backward step.

The current behaviour of wagtailredirects isn't the best fit for everything (indeed, people have created their own alternatives to it - see #655 for example), but existing users are relying on it to continue working in the same way, and I'd be very reluctant to make such a sweeping change to an established part of Wagtail.

PR #3100, opened 2016-10-20
Adding the Meta attribute `group` to Blocks.

nutztherookie:
The idea is that Streamfield-Blocks can be grouped in the CMS
by setting Meta's `group` attribute.
This gives a better overview, particularly with many blocks.
If the attribute remains unset, nothing changes.

### DIFF
--- a/wagtail/wagtailadmin/templates/wagtailadmin/block_forms/stream_menu.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/block_forms/stream_menu.html
@@ -1,10 +1,14 @@
 <div class="stream-menu {% if state == 'closed' %}stream-menu-closed{% endif %}" id="{{ menu_id }}">
     <a href="#" class="toggle"><span>Insert block</span></a>
     <div class="stream-menu-inner">
-        <ul>
-            {% for child_block in child_blocks %}
-                <li><button type="button" class="button action-add-block-{{ child_block.name }} icon icon-{{ child_block.meta.icon }}"><span>{{ child_block.label }}</span></button></li>
-            {% endfor %}
-        </ul>
+        {% regroup child_blocks by meta.group as grouped_child_blocks %}
+        {% for child_blocks in grouped_child_blocks %}
+            <h3>{{child_blocks.grouper}}</h3>
+            <ul>
+                {% for child_block in child_blocks.list %}
+                    <li><button type="button" class="button action-add-block-{{ child_block.name }} icon icon-{{ child_block.meta.icon }}"><span>{{ child_block.label }}</span></button></li>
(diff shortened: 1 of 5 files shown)


nutztherookie:
Example output with this patch applied:
image

wagtail/wagtailadmin/templates/wagtailadmin/block_forms/stream_menu.html
```diff
@@ -1,10 +1,14 @@
...
     <a href="#" class="toggle"><span>Insert block</span></a>
     <div class="stream-menu-inner">
-        <ul>
-            {% for child_block in child_blocks %}
-                <li><button type="button" class="button action-add-block-{{ child_block.name }} icon icon-{{ child_block.meta.icon }}"><span>{{ child_block.label }}</span></button></li>
-            {% endfor %}
-        </ul>
+        {% regroup child_blocks by meta.group as grouped_child_blocks %}
+        {% for child_blocks in grouped_child_blocks %}
+            <h3>{{child_blocks.grouper}}</h3>
```
gasman: It would be good to add an `{% if child_blocks.grouper %}` around this, so that we don't generate an empty `<h3></h3>` in the default case.

nutztherookie:
Ahoyhoy @gasman, I don't really know how or where to add a test for this. Could you give me a hand or some pointers? Thanks :)

gasman:
I'd suggest adding a test method to [wagtailcore.tests.test_blocks.TestStreamBlock](https://github.com/torchbox/wagtail/blob/78f67f8fdb72cba202d4a5109d1a76b1cc0d20a3/wagtail/wagtailcore/tests/test_blocks.py#L1450), that follows a structure something like [this one](https://github.com/torchbox/wagtail/blob/78f67f8fdb72cba202d4a5109d1a76b1cc0d20a3/wagtail/wagtailcore/tests/test_blocks.py#L1574-L1587) - define a suitable StreamBlock object that makes use of the new `group` parameter, call `render_form` on it, and confirm that the result contains `<h3>some heading</h3>`. (We're not really testing the whole sorting/grouping functionality that way, but I think this is sufficient - the most important thing is to have an example of the `group` parameter somewhere within the test cases, so that if someone accidentally breaks it in a future code change, we'll know about it.)

SalahAdDin:
:+1:

gasman:
Looks good to me, thanks! Merged in bfd4a2ef4b898a6d06bf8b223005f40b33d43fa6.

nutztherookie:
Super, thanks :)

pyMan:
Thank you @nutztherookie and @gasman.
Is there a way to render groups in a specific order?

PR #2985, opened 2016-09-07
Increase legibility by increasing contrast

strindhaug:
Improves legibility by raising the font color contrast slightly,  by adjusting font-weight from 300 up to normal for small text, and increasing font size (from 12.2px to 12.8px in the menu) for the smallest texts.

This small change in font size makes a big difference in font opacity as the strokes goes from significantly less than a pixel to approaching one pixel in width; as fractional pixels will be rendered as semi-transparent pixels.

Fix (or at least an improvement) for https://github.com/torchbox/wagtail/issues/2976

Issue #2224, opened 2016-02-11
Certain colours combinations not validating against WCAG AA


heymonkeyriot:
There are a couple of colour combinations used across the admin area that don't validate against [WCAG 2.0](https://www.w3.org/TR/WCAG20/). The teal-color on white, or white on teal-color doesn't validate 

```
color: #43b1b0
bg: #fafafa
contrast: 2.47:1

color: #fff
bg: #43b1b0
contrast: 2.58:1
```

Note: To validate against AA there needs to be a contrast of 4.5:1

We use it on the dashboard
![image](https://cloud.githubusercontent.com/assets/11335309/12971797/de17f296-d096-11e5-9430-cbe839f41946.png)

And on button hover states 
![image](https://cloud.githubusercontent.com/assets/11335309/12971791/c07ca45c-d096-11e5-8eb3-f6ea5ab29c2f.png)

There's a couple of ways we could move around this in the short term:
### Using color-teal-dark

![image](https://cloud.githubusercontent.com/assets/11335309/12971707/f7b878f2-d095-11e5-81bf-d6cdf4ae3166.png)
Since it's already in the sass variables would be to use `$color-teal-dark: #246060;` for the header area and buttons. The issue with the color-teal-dark is that it loses quite a lot of saturation and we'd need to change all the examples through the sass files
### Amend color-teal

![image](https://cloud.githubusercontent.com/assets/11335309/12971692/cde83634-d095-11e5-85ba-1c76b2c1d9a4.png)
Slightly more straight-forward (as it'd be changing a single variable) we could amend the color-teal to something more saturated, and lighter than `color-teal-dark`, which validates 4.5:1 in contrast. The above uses `#198180`. 
### Use `#333` text

![image](https://cloud.githubusercontent.com/assets/11335309/12971687/c1057648-d095-11e5-9bd0-642855f316a9.png)
Another alternative is that `#333` validates against `#43b1b0` but it really flattens out the visual style by darkening everything.

As with some other issues I've raised it's potentially quite an opinionated change as it'll have quite an impact on the UI so would be good to have thoughts.

thibaudcolas:
Moving the discussion over from #3106, which demonstrates the types of changes required that we can't merge as is.

> To test this:
> The colours were run through http://webaim.org/resources/contrastchecker/ and http://leaverou.github.io/contrast-ratio/

> Core team position:
> Those are quite sweeping visual changes so they won't be merged as is. We want a visual design person to get involved to assess this further.
[...]

PR #3106, opened 2016-10-25
Increase contrast of colours to validate WCAG AA

heymonkeyriot:
This is a pull request for issue #2224.

Wagtail doesn't currently validate against WCAG colour contrast guidelines. In 1.7 @strindhaug   improved the contrast in PR #2985, which helps but doesn't solve the problem with the background colour within headers and on revision pages (or any page where the warning button is displayed).

@alexgleason and I had previously worked on a [pip installable module](https://pypi.python.org/pypi/wagtailatomicadmin/) to get round this problem, and some others, on certain projects, but it used some horrible hacks to get around this particular issue, and was a very brittle solution to the problem.

The proposed solution is to use the existing `$color-teal-dark` as `$color-teal` and change `$color-teal-dark` to `#0F2828`.

**Explorer page with breadcrumb**
capture d ecran 2016-10-25 a 07 20 33

**Revision page**
capture d ecran 2016-10-25 a 07 20 00

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_forms.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_forms.scss
@@ -255,10 +255,11 @@ input[type=checkbox]:checked:before {
     &.warning {
         background-color: $color-button-warning;
         border: 1px solid $color-button-warning;
+        color: $color-button-warning-text;
 
         &.button-secondary {
             border: 1px solid $color-button-warning;
-            color: $color-button-warning;
+            color: darken($color-button-warning, 40%);
             background-color: transparent;
         }
 
@@ -269,7 +270,7 @@ input[type=checkbox]:checked:before {
         }
 
         &.button-nobg:hover {
-            color: $color-button-warning;
+            color: darken($color-button-warning, 40%);
             background-color: transparent;
         }
     }
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_messages.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_messages.scss
@@ -44,11 +44,17 @@
     }
 
     .warning {
-        background-color: $color-orange;
-
+        background-color: $color-button-warning;
+        color: $color-button-warning-text;
         &:before {
             font-family: wagtail;
             content: '!';
+        }
+        .avatar {
+            &:before {
+                color: $color-button-warning-text;
+                border: 1px solid $color-button-warning-text;
+            }
         }
     }
 
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_messages.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_messages.scss
@@ -46,10 +46,12 @@
     .warning {
         background-color: $color-button-warning;
         color: $color-button-warning-text;
+
         &:before {
             font-family: wagtail;
             content: '!';
         }
+
         .avatar {
             &:before {
                 color: $color-button-warning-text;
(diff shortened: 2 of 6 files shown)


wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_messages.scss
```diff
@@ -44,12 +44,20 @@
...
             font-family: wagtail;
             content: '!';
         }
+
+        .avatar {
+            &:before {
+                color: $color-button-warning-text;
+                border: 1px solid $color-button-warning-text;
+            }
+        }
```
thibaudcolas: I don't understand those changes to the avatar, what is the issue accessibility-wise?

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_forms.scss
```diff
@@ -269,7 +270,7 @@ input[type=checkbox]:checked:before {
         }
 
         &.button-nobg:hover {
-            color: $color-button-warning;
+            color: darken($color-button-warning, 40%);
```
thibaudcolas: This should be extracted to a variable as well.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
```diff
@@ -221,7 +221,7 @@ footer {
                 content: 'n';
                 padding-left: 20px;
                 font-size: 2em;
-                color: $color-teal-darker;
+                color: lighten($color-teal, 10%);
```
thibaudcolas: Should be extracted to a variable.

heymonkeyriot:
Hi @thibaudcolas, apologies for not responding more quickly to this.

The colours were run through http://webaim.org/resources/contrastchecker/ and http://leaverou.github.io/contrast-ratio/

I think there's possibly a broader question about all of the colours within the Wagtail UI but probably out of scope of this pull request.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_forms.scss
heymonkeyriot: Agreed. `$color-button-warning-darker` ?

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_messages.scss
heymonkeyriot: On screens where you're looking at a previous revision the avatar and user's avatar name are shown. Currently it's white text on orange background. It's confusing because it's slightly poor name spacing, since the revision screens use the `.error` class throughout to get their style

wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
heymonkeyriot: I wonder if this is a big enough colour shift to justify a new colour variable?

thibaudcolas:
I've added this for review at the next core team meeting. Would also be worth discussing in Iceland 🙂

heymonkeyriot:
Sounds good. Thanks @thibaudcolas

thibaudcolas:
Discussed this last week and today – those are quite sweeping visual changes so they won't be merged as is. We want a visual design person to get involved to assess this further.

A potential alternative approach would be to have settings of sorts for a separate "high-contrast mode". I think the default install shouldn't compromise on accessibility, but realistically it would be easier to solve this within Wagtail itself with a separate code path.

heymonkeyriot:
Yup, seems totally reasonable. I think it'd make a lot of sense for a visual designer to consider how this could be moved forward. As you say it's a broad visual change, and something that has a big impact on the UI.

(And do let me know if it's better to close this PR, rather than having an open PR that's unlikely to be merged in.)

thibaudcolas:
You're right – I forgot that there was an issue open as well. Closing this, we can take the discussion there.

PR #3107, opened 2016-10-25
Fixed #3078 -- Add empty panel attributes

### LINKED ISSUE #3078: Some Page model attribures are defined elsewhere
Some of the Page model's attributes are defined outside of the Page main definition module. This causes confusing errors that in some cases mask the actual exception.

Yesterday we spent some time debugging a strange `AttributeError` which in the end was caused by a missing `import` statement.

In our case we had `INSTALLED_APPS` set to something like this:

```
INSTALLED_APPS = [
    'mypages',

    'wagtail.wagtailforms',
    'wagtail.wagtailredirects',
    'wagtail.wagtailembeds',
    'wagtail.wagtailsites',
    'wagtail.wagtailusers',
    'wagtail.wagtailsnippets',
    'wagtail.wagtaildocs',
    'wagtail.wagtailimages',
    'wagtail.wagtailsearch',
    'wagtail.wagtailadmin',
    'wagtail.wagtailcore',
]
```

And then in `mypages.models` defined a model using a `FieldPanel`.

```
from django.db import models

class MyPage(Page):
        FieldPanel('date'),
    ]
```

This code caused a `AttributeError` to be raised, telling us that `Page.content_panels` didn't exists.

After a while we decided to change the order of `INSTALLED_APPS` so the `wagtail` modules came before our custom apps. This revealed the actual error, we forgot to import the `FieldPanel` class from `wagtail.wagtailadmin.edit_handlers`.

From what I've gathered it was a conscious decision to define some of the Page's attributes in `edit_handlers`, but from our perspective it's very confusing to have a different error depending on which order you've listed the wagtail modules in `INSTALLED_APPS`.

One resolution would be to define `content_panels` and other fields in `wagtail.core.models` with sane defaults (e.g. an empty list, None etc.) and then override them in `edit_handlers`.

/cc @ramonakira

jaap3:
Accessing any of these attributes before importing `wagtailadmin.edit_handlers` would raise a `AttributeError`. In some cases this would mask the actual error; e.g. forgetting to import `wagtailadmin.edit_handlers`.

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -395,6 +395,12 @@ class Page(six.with_metaclass(PageBase, AbstractPage, index.Indexed, Clusterable
 
     # Do not allow plain Page instances to be created through the Wagtail admin
     is_creatable = False
+    
+    # Define these attributes early to avoid masking errors. (Issue #3078)
+    # The canonical definition is in wagtailadmin.edit_handlers.
+    content_panels = []
+    promote_panels = []
+    settings_panels = []
 
     def __init__(self, *args, **kwargs):
         super(Page, self).__init__(*args, **kwargs)
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -395,7 +395,7 @@ class Page(six.with_metaclass(PageBase, AbstractPage, index.Indexed, Clusterable
 
     # Do not allow plain Page instances to be created through the Wagtail admin
     is_creatable = False
-    
+
     # Define these attributes early to avoid masking errors. (Issue #3078)
     # The canonical definition is in wagtailadmin.edit_handlers.
     content_panels = []


wagtail/wagtailcore/models.py
```diff
@@ -395,6 +395,12 @@ class Page(six.with_metaclass(PageBase, AbstractPage, index.Indexed, Clusterable
 
     # Do not allow plain Page instances to be created through the Wagtail admin
     is_creatable = False
+
```
153957: remove spaces on blank line
( http://46.101.88.142/torchbox/wagtail/602 )

jaap3:
Fixed the pep-8 violation, should I squash the commits? Or will you squash when merging?

gasman:
I'll squash when merging, thanks.

gasman:
Merged in fb06f6c0c19182470b285aa1817f0afb211ecbac - thanks!

PR #3109, opened 2016-10-25
Loop through all multiple value fields when sending form submission by email.

saevarom:
A form field with multiple checkboxes may introduce submissions that have more than one value. The `send_email` function would loop through all form fields and grab only the first item of each field, ignoring fields that may have multiple values selected. This is a side effect of using `QueryDict.get()`, which will only return the last value of a list.

This commit fixes that by first converting the `QueryDict` instance to a regular dict of lists so that `get` will return all values, then joining multiple items if needed.

### DIFF
--- a/wagtail/wagtailforms/models.py
+++ b/wagtail/wagtailforms/models.py
@@ -295,7 +295,7 @@ class AbstractEmailForm(AbstractForm):
 
     def send_mail(self, form):
         addresses = [x.strip() for x in self.to_address.split(',')]
-        content = '\n'.join([x[1].label + ': ' + text_type(form.data.get(x[0])) for x in form.fields.items()])
+        content = '\n'.join([x[1].label + ': ' + text_type( ', '.join(dict(form.data.lists())[x[0]]) ) for x in form.fields.items()])
         send_mail(self.subject, content, addresses, self.from_address,)
 
     class Meta:
--- a/wagtail/wagtailforms/models.py
+++ b/wagtail/wagtailforms/models.py
@@ -295,7 +295,15 @@ class AbstractEmailForm(AbstractForm):
 
     def send_mail(self, form):
         addresses = [x.strip() for x in self.to_address.split(',')]
-        content = '\n'.join([x[1].label + ': ' + text_type( ', '.join(dict(form.data.lists())[x[0]]) ) for x in form.fields.items()])
+        content = []
+        for field in form:
+            value = field.value()
+            if isinstance(value, list):
+                value = ', '.join(value)
+                if len(value) == 0:
+                    value = 'None'
+            content.append('{}: {}'.format(field.label, value))
+        content = '\n'.join(content)
         send_mail(self.subject, content, addresses, self.from_address,)
 
     class Meta:
--- a/wagtail/wagtailforms/models.py
+++ b/wagtail/wagtailforms/models.py
@@ -300,8 +300,6 @@ class AbstractEmailForm(AbstractForm):
             value = field.value()
             if isinstance(value, list):
                 value = ', '.join(value)
-                if len(value) == 0:
-                    value = 'None'
             content.append('{}: {}'.format(field.label, value))
         content = '\n'.join(content)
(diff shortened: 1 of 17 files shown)


gasman:
Thanks @saevarom! Please could you add a test for this (and investigate the test that's currently failing)?

wagtail/wagtailforms/models.py
```diff
@@ -295,7 +295,7 @@ def process_form_submission(self, form):
 
     def send_mail(self, form):
         addresses = [x.strip() for x in self.to_address.split(',')]
-        content = '\n'.join([x[1].label + ': ' + text_type(form.data.get(x[0])) for x in form.fields.items()])
+        content = '\n'.join([x[1].label + ': ' + text_type( ', '.join(dict(form.data.lists())[x[0]]) ) for x in form.fields.items()])
```
jaap3: I'd prefer to do it this way:

```
content = []
for field in form:
    value = field.value()
    if isinstance(value, list):
        value = ', '.join(value)
    content.append('{}: {}'.format(field.label, value))
content = '\n'.join(content)
```

jaap3: That way will also not break the tests ;-)

saevarom: You're right, I still get one failure with that code, but that is fixed by converting an empty list to a string of 'None', like so:

```python
        content = []
        for field in form:
            value = field.value()
            if isinstance(value, list):
                value = ', '.join(value)
                if len(value) == 0:
                    value = 'None'
            content.append('{}: {}'.format(field.label, value))
        content = '\n'.join(content)
```

saevarom:
I've rewritten the changes as per @jaap3's suggestions.
Also, I've added a test that catches the problem I was dealing with in the first place (multiple checkbox values always represented as one value in emails).

wagtail/wagtailforms/models.py
```diff
@@ -295,7 +295,15 @@ def process_form_submission(self, form):
 
     def send_mail(self, form):
         addresses = [x.strip() for x in self.to_address.split(',')]
-        content = '\n'.join([x[1].label + ': ' + text_type(form.data.get(x[0])) for x in form.fields.items()])
+        content = []
+        for field in form:
+            value = field.value()
+            if isinstance(value, list):
+                value = ', '.join(value)
+                if len(value) == 0:
```
jaap3: Shouldn't this be dedented one level so an empty string value is handled the same as an empty list value? And more importantly, is this even required behaviour? 

As a Python developer I would confuse this with an actual None. As an end user I would confuse this with someone actually entering (or selecting) the value None in a form field.

wagtail/wagtailforms/models.py
```diff
@@ -295,7 +295,15 @@ def process_form_submission(self, form):
...
+                    value = 'None'
+            content.append('{}: {}'.format(field.label, value))
```
jaap3: Maybe, instead of converting an empty string to 'None', it would be better to only add a line to the email if the value is not empty? This could be done by adding a `if value != '':` before this statement.

wagtail/wagtailforms/models.py
saevarom: Well, yes, here I was simply following the original output of this function.  If you look at the tests (https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailforms/tests/test_models.py#L288) then you can see that the default behaviour was to include a string value of None (probably the default output of the get function when no value was present).

saevarom: Perhaps an empty value should just be exactly what your original code produced, the field label and an empty string. Then we would have to change the test accordingly so that it would not expect a string value of None.

jaap3: I'd say the original function was wrong then and the default of the get method should have been an empty string. If @gasman agrees you can probably just fix the test instead of treating empty lists as a special case.

gasman:
Thanks! Merged in 169045b186f15352a9526f4c5cf07329676e7c69.

PR #3110, opened 2016-10-26
More helpful message if javascript is not activated.

vincentaudebert:
At the moment we give an error message without explaining how to fix the problem, although the likely fix is well known and documented. We should make it easier for people to do the right thing.

What it looks like - cf attachments
screen shot 2016-10-27 at 9 48 52 am
screen shot 2016-10-27 at 9 49 06 am

To test it, go into the settings of your developer tool bar and disable Javascript :)

@thibaudcolas for your review.

### DIFF
--- a/wagtail/wagtailadmin/templates/wagtailadmin/skeleton.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/skeleton.html
@@ -22,7 +22,7 @@
     <![endif]-->
     <noscript class="capabilitymessage">
         {% blocktrans %}
-            For full functionality of this section it is necessary to enable JavaScript.
+            Javascript is required to use Wagtail, but it is currently disabled.<br />
             Here are the <a href="http://www.enable-javascript.com/" target="_blank" rel="noopener">instructions how to enable JavaScript in your web browser</a>.
         {% endblocktrans %}
     </noscript>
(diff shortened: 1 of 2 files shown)


wagtail/wagtailadmin/templates/wagtailadmin/skeleton.html
```diff
@@ -20,7 +20,12 @@
     <!--[if lt IE 9]>
         <p class="capabilitymessage">{% blocktrans %}You are using an <strong>outdated</strong> browser not supported by this software. Please <a href="http://browsehappy.com/">upgrade your browser</a>.{% endblocktrans %}</p>
     <![endif]-->
-    <noscript class="capabilitymessage">{% trans 'Javascript is required to use Wagtail, but it is currently disabled' %}</noscript>
+    <noscript class="capabilitymessage">
+        {% blocktrans %}
+            For full functionality of this section it is necessary to enable JavaScript.
```
thibaudcolas: I think this first line should still be the same "JS is required to use Wagtail" message. Saying "For full functionality of this section [...]" is confusing at best.

gasman:
Looks good to me - thanks!

MechanisM:
Just wonder why don't you used native to vagtail "Error message" for displaying this message? Just like this:
image

thibaudcolas:
Good question. I think it's fairly common for sites to use different styles for browser capability messages and other errors, not sure what the reasoning is though.

gasman:
I think it's fine for the no-JS message to use its own style. It's natural for it to be treated as an "exceptional" case, and rendered using a different mechanism, since without Javascript you can't really rely on our standard reusable UI components to behave correctly.

(Also, making the "Instructions how to enable Javascript in your web browser" text into a button looks really wrong...)

vincentaudebert:
Totally agree with @gasman
No-JS should be just plain HTML so you can be 100% sure it will work.

MechanisM:
I agree that this long text in button looks ugly(I did it quickly in `chrome`'s inspect page bar  to demonstrate) it can look different.
And yeah, this message can be same - just plain html but will be shown in `wagtail`'s default messages. But of course it's all up to you.

PR #3112, opened 2016-10-26
Recommend Django's cached template loader

jaap3:
As noted in PR #3077:

 >"With the cached template loader enabled, there's a significant performance improvement [...]"


gasman:
Thanks!

PR #3115, opened 2016-10-27
Hopefully fix intermittent test failures

timheap:
Something was polluting the edit handlers with rich text fields that
then failed somewhere else due to missing configs. I'm not sure exactly
where and what the leakage was, but the test now pass consistently for
me so hopefully this fixed it? 🤷


gasman:
Merged - thanks!

PR #3118, opened 2016-10-27
Make the tests less noisy

timheap:
The tests print a whole bunch of deprecation warnings, most of which are just ignored. This makes finding actual warnings very difficult.

The default logging level for tests has been changed to `imminent`, so that only DeprecationWarnings are printed. PendingDeprecationWarnings are ignored. This makes the test output usable again on all tests except for those against latest Django. The warnings against latest Django should be dealt with.

With the test output now usable, a few warnings were found and squashed.

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -1,20 +1,21 @@
 #!/usr/bin/env python
 
-import sys
+from __future__ import absolute_import, unicode_literals
+
+import argparse
 import os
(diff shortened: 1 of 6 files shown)


gasman:
Thanks! Merged in 1849e9edf718cae43461884713dbc761477493a2 + parents.

PR #3119, opened 2016-10-27
Don't repeat set_url_path logic

garykrige:
Classes that extend and override logic for set_url_path need to be respected by this command.

set_url_path is lot more useful when you can namespace a whole lot of pages with this command.

Use case:
We want to namespace our blog urls with the publish date. We would love to go to old posts and change all the url_paths for 1000+ posts.

Also this is a more DRY approach.

### DIFF
--- a/wagtail/wagtailcore/management/commands/set_url_paths.py
+++ b/wagtail/wagtailcore/management/commands/set_url_paths.py
@@ -9,12 +9,12 @@ class Command(BaseCommand):
 
     help = 'Resets url_path fields on each page recursively'
 
-    def set_subtree(self, root, root_path):
-        root.url_path = root_path
+    def set_subtree(root, parent=None):
+        root.specific.set_url_path(parent)
         root.save(update_fields=['url_path'])
         for child in root.get_children():
-            self.set_subtree(child, root_path + child.slug + '/')
+            set_subtree(child.specific, root)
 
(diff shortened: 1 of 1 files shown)


gasman:
Hi @garykrige - thanks for the PR!

The `set_url_path` method is not designed to be overridden, and I'd strongly advise against doing so - various bits of code rely on url_path matching the tree structure (for example: https://github.com/torchbox/wagtail/blob/2442416e93a9cec4325a41894279157cee4e3327/wagtail/wagtailcore/models.py#L600), and depending on the nature of the customisations, it could result in the tree becoming corrupted on moving/renaming pages.

The recommended way of customising URL patterns is through a custom `route()` method: http://docs.wagtail.io/en/v1.7/reference/pages/model_recipes.html#adding-endpoints-with-custom-route-methods

garykrige:
@gasman thanks for the recommendation.

I am using wagtail as an api. My reason for overriding `set_url_path` was to simplify generating a sitemap. I am currently overriding `url`, but the sitemap app gets urls with `full_url`. So instead of overriding both I went this route (dammit english). 

Pages.routes is not considered at all in building a sitemap. I will make a PR for that I guess.

Thanks for a brilliant framework.

garykrige:
I still think this PR has value simply for DRY's sake. I will update to make it a safe.

Issue #3085, opened 2016-10-14
'Parent' link in page chooser search should not navigate away


gasman:
### Issue Summary

The 'Parent' link in search results inside the page chooser navigates away to the explorer, which isn't a very sensible thing to do within a chooser modal.
### Steps to Reproduce
1. Edit a page
2. Within a rich text area, open the link chooser
3. Search for something with at least one result
4. Click on the 'parent' link

Expected: It should switch to the explore view _within the page chooser_, so that you're choosing from the children of that page (including the search result you were just looking at)
Actual: It leaves the page edit interface and takes you to that page in the explorer
### Technical details
- Python version: 3.4.3
- Django version: 1.8.15
- Wagtail version: 1.8a0
- Browser version: Chrome 53.0.2785.143

lb-:
Referencing PR #3120 which was submitted to fix this issue but has status 'needs work'.

gasman:
Fixed in #5011

PR #3120, opened 2016-10-27
Fix issue #3085 -- 'Parent' link in page chooser search should not navigate away

saevarom:
The handling of links is slightly different between `ajaxifyBrowseResults` and `ajaxifySearchResults` in browse.js so I created a new class to distinguish parent links from other links. 

Also, the naming of the url, `wagtailadmin_choose_page_child`, is slightly confusing, since it can be used to choose either a child or parent, based on what id you supply as an argument. Perhaps there should be a url called `wagtailadmin_choose_page_parent`, but it would have the same url pattern as `wagtailadmin_choose_page_child` - making it redundant.

Perhaps a new name is needed for the `r'^choose-page/(\d+)/$'` url to reflect this.

### DIFF
--- a/wagtail/wagtailadmin/templates/wagtailadmin/chooser/browse.js
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/chooser/browse.js
@@ -68,6 +68,12 @@ function(modal) {
             $('.page-results', modal.body).load(this.href, ajaxifySearchResults);
             return false;
         });
+        /* Set up links to parent pages sho that they can be browsed within the
+        modal */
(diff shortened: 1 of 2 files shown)


gasman:
Hi @saevarom - thanks for the PR!

Unfortunately this fix isn't valid as it stands - the template `wagtailadmin/pages/listing/_list.html` is used by all page listings, not just the one in the chooser. As a result, this produces the wrong 'parent' link in search results when using the search box in the main navigation menu, for example.

gasman:
Completed in #5011

PR #3121, opened 2016-10-27
Don't repeat set_url_path logic

garykrige:
Reference #3119

This approach simplifies the management command and delegates url path logic to the Page model.


gasman:
Thanks! Merged in 9038da9fcd69f0b39121cc54a72f4a6ae6beb06a (with a minor fix - removed the `.specific` on child too).

PR #3123, opened 2016-10-27
improves slugurl to match based on site, not the first match of slug

jakealstad:

### DIFF
--- a/wagtail/wagtailcore/templatetags/wagtailcore_tags.py
+++ b/wagtail/wagtailcore/templatetags/wagtailcore_tags.py
@@ -23,8 +23,14 @@ def pageurl(context, page):
 
 @register.simple_tag(takes_context=True)
 def slugurl(context, slug):
-    """Returns the URL for the page that has the given slug."""
-    page = Page.objects.filter(slug=slug).first()
+    """Returns the URL for the page that has the given slug per site."""
+    site = context['request'].site
+    site_parent = Page.objects.filter(id=site.root_page_id)[0]
+    page = Page.objects.filter(slug=slug)
+    if pages:
+        for p in pages:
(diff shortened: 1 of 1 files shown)


gasman:
Hi @jakealstad - thanks for the PR!

A few issues with the code as it stands...

* it's currently failing due to the variable `pages` not being defined
* Comparing against `p.get_parent()` means that it won't match pages that are more than one level deep. A better / simpler test would be: `if p.get_site() == site`. (Even better would be to incorporate the "must be a descendant of `site_parent`" condition into the initial query: `pages = Page.objects.descendant_of(site_parent, inclusive=True).filter(slug=slug)`)
* To preserve backwards compatibility, it would be better to check for a match within the current site first, and if none is found, fall back on the old behaviour of matching anywhere in the page tree.
* We need to make sure the tag still works when `request` isn't found in the context - this might happen if `{% slugurl %}` is being called from within an inclusion tag. Obviously, in this situation we can't determine the current site, so it should just fall back on the old behaviour.
* Once this is working, we'd need some additional tests in https://github.com/torchbox/wagtail/blob/b4954c1fdff85b1fd5de0c03e09f0b11cf61b7b1/wagtail/wagtailcore/tests/tests.py#L13-L26 to cover the new behaviour. I'd suggest two new tests: one to confirm that it picks the right page when there are two pages with the same slug (one in the current site, one elsewhere), and one to confirm that the fallback behaviour happens when no match exists on the current site.

jakealstad:
Hey @gasman!

Thanks for the notes. I realize this was kind of a hasty implementation, and to be honest looking at it now, it is pretty specific for my use case. I will go ahead and update with your suggestions when I can. 

Thanks!
Jake

Issue #3096, opened 2016-10-20
Getting started Blog index doesn't create an index


goetzk:
Hi,
There seems to be some text missing from the tutorial

* Visit http://docs.wagtail.io/en/v1.6.3/getting_started/tutorial.html#blog-index
* Follow steps
* Pages in your site are not listed on the index.

I suggest there should be a short snippet for the template which iterates and lists all pages in the site to add content for the index page.

thanks,

gasman:
Hi @goetzk,

The following section of the tutorial ("Related items") covers adding items to the index page, as manually-assigned links. It would probably be a good idea for the tutorial to cover creating automatic listings too, but that's a separate task.

shacker:
I know that this is already closed, but just wanted to weigh in as I just experienced the same. Having blog pages automatically included on a blog index page is almost fundamental to the definition of a blog and it is certainly what a person expects when going through a "basic blog" tutorial. My suggestion would be to refactor the getting started tutorial to build something more like a standard blog (perhaps with categories/tags and date-based URLs, etc.)

I would suggest removing the RelatedLinks section of Getting Started, as the modeling is actually pretty abstract and hard to follow for someone just learning wagtail for the first time. And it doesn't explain the logic of the interrelated models LinkFields, RelatedLink, BlogIndexRelatedLink, and how those are combined and factored into BlogIndexPage. It's all pretty confusing. It's like the intro tutorial goes from 0-60 in 3 seconds, leaving the user confused as to why they still don't have basic blog functionality. 

Not wanting to be too critical, but I really do think there's a lot of room for improvement here. 

I might be willing to contribute in this area, time permitting.

goetzk:
Hi @gasman ,
@shacker 's first paragraph really sums up what I was expecting that bit of the tutorial to be, hence my assumption that there was something missing. 
@shacker has said everything else I could have added so will leave it at that.

Thanks both!
kk

zerolab:
Hi @goetzk @shacker,

Your feedback is appreciated. IMHO the best form of feedback would be a pull request with the relevant changes as it can be easy to lose sight of the basic tutorial if you are not doing it every day.

Thank you.

shacker:
Great thanks - I'll have a PR ready today or later this week.

PR #3125, opened 2016-10-28
Update Getting Started tutorial to include posts automatically, plus …

shacker:
Modify "Getting Started" tutorial per discussion in #3096:
- Results in a blog where posts are included on IndexPage automatically
- Walks beginning reader more gently through some additional wagtail features
- Removes abstract/confusing RelatedLinks model building (which is no longer useful since posts are auto-included)

If there is any objection to having removed RelatedLinks, I can add it back in to the end and modify it to work something like a "blogroll." 

Feedback welcome.


heymonkeyriot:
I had missed the conversation on #3096 but quickly reading through the changes they're certainly easier to understand than the current docs. Removing the RelatedLinks also makes a lot of sense since that isn't a pattern that I can see should ever be used to populate an index page. I remember being confused by this bit of the docs too when I first tried to pick up Wagtail so great to see it being made easier!

docs/getting_started/tutorial.rst
```diff
@@ -284,143 +398,167 @@ Adjust your blog page template to include the image:
...
-        <div class="intro">{{ page.intro|richtext }}</div>
 
-        {% if page.related_links.all %}
-            <ul>
-                {% for item in page.related_links.all %}
-                    <li><a href="{{ item.link }}">{{ item.title }}</a></li>
-                {% endfor %}
-            </ul>
+        {% if request.GET.tag|length %}
+            <h4>Showing pages tagged "{{ request.GET.tag|safe }}"</h4>
```
gasman: We definitely don't want to use `|safe` here - the GET parameter is unchecked user input, and this will create an XSS vulnerability.

docs/getting_started/tutorial.rst
```diff
@@ -202,21 +252,85 @@ Create a template at ``blog/templates/blog/blog_page.html``:
...
+Be sure to select type "BlogPage" when creating your posts.
 
-.. figure:: ../_static/images/tutorial/tutorial_4.png
-   :alt: Create page screen
+.. figure:: ../_static/images/tutorial/tutorial_4a.png
+   :alt: Create blog post as child of BlogIndex
+
+.. figure:: ../_static/images/tutorial/tutorial_4b.png
+  :alt: Choose type BlogPost
+
```
gasman: The screenshots here incorrectly show the blog post being created as a child of Homepage.

shacker:
Thanks for the feedback - I'll take care of these items today, hopefully. 

For the screenshots, is there a recommended width that works best for docs?

I actually would have preferred to use Categories rather than tags, but wasn't able to figure that out quickly. It might be nice to have both categories and tags - maybe I'll save that for another day.

gasman:
I don't think we have a standard size for screenshots - the CSS scales them down to fit the column width where appropriate, so the original size isn't too important - but it looks like we typically go for around 1000px width.

shacker:
I'm wondering if we can drop the note on using tuples for wagtail versions < 1.5 - highly unlikely that anyone just learning would be on a version that old. I'll drop the note in next PR but can restore if you object.

gasman:
@shacker Yep, absolutely fine to remove that. If I recall correctly, that note was added while 1.5 was in development, because we had a few cases of people running the then-current 1.4 release and following the development version of the tutorial by mistake, then complaining when it didn't work...

shacker:
Changes submitted:

- Addresses all issues raised in PR review
- Adds more info to "Navigating parents/children"
- Adds a new dedicated section on overriding context

shacker:
OK now I have a handle on creating Categories. If you want, I can modify this PR to use multiple categories per post instead of tags, OR show both categories AND tags.

gasman:
The updates look good, thanks!

Categories would be using a child model with InlinePanel, presumably? That's definitely something worth covering in the tutorial I think - if people don't know about child objects, I suspect they're more likely to misuse rich text for structured content instead - so that would be a very welcome addition. I have no particular preference about whether to have categories instead of, or in addition to, tags - happy to leave that up to you.

tomdyson:
>  I have no particular preference about whether to have categories instead of, or in addition to, tags - happy to leave that up to you.

I vote for tags _and_ categories. Thanks @shacker!

docs/getting_started/tutorial.rst
```diff
@@ -606,6 +609,237 @@ something like this:
...
+            Filter BlogPages by category
+            """
+            category = get_object_or_404(BlogCategory, slug=catslug)
+            blogpages = BlogPage.objects.filter(
+                blog_categories__category=category).live().order_by('-first_published_at')
+
+            context = self.get_context(request)
+            context['blogpages'] = blogpages
+            context['category'] = category
+            return render(request, 'blog/blog_index_page.html', context)
```
shacker: I was unsure about this part - is this the best way to render from a RoutablePageMixin method? Do we really have to specify the template name? Wagtail docs for RoutablePageMixin leave out sample code, so it's ambiguous.

shacker:
A  working Categories system has now been added, which expands coverage to include using a vanilla Django model, using PageRoutableMixin, adding contrib modules, and more. I hope you're OK with the approach I took on this - there are several ways to skin this cat. I left one question for you guys in the PR.

docs/getting_started/tutorial.rst
gasman: Yep, that's correct - view methods inside a routable page are essentially plain vanilla Django views.

docs/getting_started/tutorial.rst
```diff
@@ -284,143 +443,402 @@ Adjust your blog page template to include the image:
...
+        {% endfor %}
+
     {% endblock %}
 
-Related items
-~~~~~~~~~~~~~
+Unlike in the previous example, we're linking to pages here with the builtin ``slugurl``
+tag rather than ``pageurl``. The difference is that ``slugurl`` takes a Page slug
+(from the Promote tab) as an argument. ``pageurl`` is more commonly used because it
+is unambiguous, but use whichever one best suits your purpose.
```
gasman: Sorry, I probably wasn't clear about what I intended re slugurl vs pageurl... The place where I would consider `slugurl` to be "appropriate" is in the blog_page.html snippet a bit further up:

    <a href="{% slugurl 'tags' %}?tag={{ tag }}"><button type="button">{{ tag }}</button></a>

Here we don't have a reference to the BlogTagIndexPage object readily available, so identifying it by slug is as good a way as any. I propose that that's the point at which we explain `slugurl` in the text.

I'm not keen on the notion of `{% slugurl blogpage.slug %}` being an alternative way of writing `{% pageurl blogpage %}`. In cases where you have the page object already available, the `pageurl` version is preferable in every respect (readability, unambiguity and avoiding extra database lookups), and that's the one I'd like to promote as best practice for people following the tutorial.

shacker: @gasman OK with you if I just change drop the slugurl and use pageurl everywhere? Or should I make the distinction?

gasman: Sure, if you think that works better - although then that leaves the question of how to do the link to BlogTagIndexPage. (Having said that, simply hard-coding a link to `/tags/` _is_ a valid option.) If you're not certain, feel free to leave this one with me...

shacker: I got this  - OK, cleaned up / moved in the most recent commit. Thanks for the feedback. The rest is OK?

gasman:
Thanks again @shacker!

I'll need to take some time to figure out the best approach for the categories section. While I have no disagreement with the technical choices you've made there, such as using modeladmin and RoutablePageMixin, I'm concerned that it's not a good fit for that point in the tutorial, given the concepts that a reader still has to learn at that point:

* Saying "this model is not a Wagtail Snippet" isn't meaningful to a reader who hasn't been introduced to the concept of snippets yet;
* Child models and InlinePanels are a really important feature for people to grasp (without them, people will tend to over-use rich text areas instead), but here they're being introduced in a pretty negative light - essentially saying "Wagtail doesn't support M2M relations, so this is the hack you have to use instead".

This is absolutely not a criticism of how you've written this section... the truth is, categories _aren't_ a particularly easy thing to implement in Wagtail, and that's a shortcoming in Wagtail that we really ought to address. Until that happens, it's completely fair that the tutorial should be honest about the places where workarounds and contrib modules are necessary. However, I suspect that any user is going to have trouble understanding those workarounds if they haven't first been taught about the features that we're working around...

Obviously I don't want to turn this into an endless cycle of requests for revisions and additions, especially given that what we have now is really a vast improvement on the current version of the tutorial. Currently I'm thinking that the most effective way to solve this would be to alter the 'Image support' section so that each blog page gets a gallery of image + caption objects, rather than just a single main_image. That way, we have a "positive" example of InlinePanel in use, and we can even follow it up by defining a main_image method on BlogPage that returns the first image (or None) for use on the index page - which serves as a nice case study in why structuring your data properly with child models is a good idea ("note that we wouldn't be able to pull out the image like this if it was buried inside a rich text field"). I'll try to get this written up in the next couple of days, assuming you don't beat me to it :-)

shacker:
@gasman  To be honest, I shared your concerns about introducing what is essentially a workaround / hack for something that should be easier to accomplish. Also I realized later that the way I set it up only allows one Category per post, which is not ideal. So I agree that needs to be refactored. Proposal - we could cut the Categories section out (since it's at the end it'll be easy) and publish it now (assuming there are no other show stoppers), then perhaps you could write up a section on image carousels? WDYT?

gasman:
@shacker Sounds like a good plan!

gasman:
Merged in 756d5ec0ddde272b21aca22c46fb7933ecea966d + parents, with the categories section removed and a bit of final cleanup (including revising the MultiFieldPanel example to demonstrate grouping several fields together) in 011f2099ebad34edaeab2cbdf262c95de4cc5d52.

shacker:
Thanks @gasman . So does this go live with the next Wagtail release, or do doc changes like this go live sooner?

gasman:
Hi @shacker,
Slightly annoyingly, readthedocs pins the documentation to specific releases (with the exception of [the 'latest' option](http://docs.wagtail.io/en/latest/getting_started/tutorial.html), where it's already live) - so it'll go live with the next release. I've also merged it into the 1.7.x branch, so if we do a 1.7.1 release before 1.8, it'll show up there too.

PR #3127, opened 2016-10-29
Disabling editing if edit lock is locked

heymonkeyriot:
This PR renders fields and text unselectable on a page where page lock has been applied. Whilst the page lock removes the ability to save a page users can still edit the fields. This is a problem if a user doesn't understand what page lock means and starts editing fields (I've made this mistake myself and know it's tripped up other end-users!)

A user can still select the 'promote' and 'settings' tabs (or whatever custom tabs have been added) as well as the preview button and revisions link.

capture d ecran 2016-10-29 a 20


gasman:
Merged in cb268c233a08eb15c69d1e30f800cbecad366d0a - thanks!

PR #3128, opened 2016-10-31
Adding in_site convenience method to PageQuerySet

chrxr:
Addresses https://github.com/torchbox/wagtail/issues/3070

**Test are failing because the appropriate fixtures are not in place.**

What is the best way to add to test.json in the testapp? This test requires an additional site and an event page.

I tried to edit the JSON by hand, but just ended up breaking a whole load of other tests.

### DIFF
--- a/wagtail/wagtailcore/query.py
+++ b/wagtail/wagtailcore/query.py
@@ -350,6 +350,12 @@ class PageQuerySet(SearchableQuerySetMixin, TreeQuerySet):
         else:
             return self._clone(klass=SpecificQuerySet)
 
+    def in_site(self, site):
+        """
+        This returns all pages of type self in a site.
+        """
+        return self.descendant_of(site.root_page, inclusive=True)
+
 
 def specific_iterator(qs):
     """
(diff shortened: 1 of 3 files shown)


kaedroho:
Cheers @chrxr!

This also needs docs in https://github.com/torchbox/wagtail/blob/master/docs/reference/pages/model_reference.rst

wagtail/wagtailcore/query.py
```diff
@@ -350,6 +350,12 @@ def specific(self):
         else:
             return self._clone(klass=SpecificQuerySet)
 
+    def in_site(self, site):
+        """
+        This filters the QuerySet to only contain pages of type self in a specific site.
```
gasman: I don't think "pages of type self" is accurate - there's no reason to assume that the queryset has been (or will be) filtered by page type. Perhaps go for "This filters the QuerySet to only contain pages within the specified site"?

chrxr:
Done!

gasman:
Thanks! Merged in 89e985d02ada0a03d9bbcdf96fcef15377a45d34 (with some tweaks to the test, as the existing one would also have passed by returning an empty result)

PR #3133, opened 2016-11-01
Hide non explorable breadcrumb

jeffreyckchau:
Hello,

Continuation of #2463 by @coredumperror and #2869 by @gasman.  

Alters the breadcrumb to only show up to the highest explorable page.

Ex.
Root
|----Page 1
|----|----Child 1
|----|----Child 2
|----Page 2

User 1 has permissions on Page 1 and User 2 has permissions on Page 2. Previously, both users would see the breadcrumb for the entire installation up to the root, even if the pages were not explorable. Now each user will only see the breadcrumb up to their respective pages.

### DIFF
--- a/wagtail/wagtailadmin/templates/wagtailadmin/shared/breadcrumb.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/shared/breadcrumb.html
@@ -1,8 +1,18 @@
 {% load i18n wagtailadmin_tags %}
 
 <ul class="breadcrumb">
-    {% for page in page.get_ancestors %}
-        {% if page.is_root %}
+    {% if choosing %}
+        {% current_users_choosable_ancestors page request as ancestors %}
+    {% else %}
+        {% current_users_explorable_ancestors page request as ancestors %}
+    {% endif %}
+    {% for page in ancestors %}
+        {% if choosing %}
+            {% is_current_users_choosable_root page request as page_is_permitted_root %}
+        {% else %}
+            {% is_current_users_explorable_root page request as page_is_permitted_root %}
+        {% endif %}
+        {% if page_is_permitted_root %}
             <li class="home"><a href="{% if choosing %}{% url 'wagtailadmin_choose_page_child' page.id %}{% querystring p=None %}{% else %}{% url 'wagtailadmin_explore_root' %}{% endif %}" class="{% if choosing %}navigate-pages{% endif %} icon icon-home text-replace">{% trans 'Home' %}</a></li>
         {% else %}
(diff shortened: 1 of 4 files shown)


wagtail/wagtailadmin/templates/wagtailadmin/shared/breadcrumb.html
```diff
@@ -1,8 +1,18 @@
...
+    {% else %}
+        {% current_users_explorable_ancestors page request as ancestors %}
+    {% endif %}
+    {% for page in ancestors %}
+        {% if choosing %}
+            {% is_current_users_choosable_root page request as page_is_permitted_root %}
+        {% else %}
+            {% is_current_users_explorable_root page request as page_is_permitted_root %}
+        {% endif %}
+        {% if page_is_permitted_root %}
```
gasman: Having this extent of non-visual logic in the template probably indicates that breadcrumb.html itself should become a template tag rather than just an include, so that we can do this in Python code. I'll have a go at refactoring this.

wagtail/wagtailcore/models.py
```diff
@@ -1649,6 +1703,32 @@ def publishable_pages(self):
...
+            # Superusers are permitted to explore the tree root, which lets them explore every Page on the server.
+            return [Page.get_first_root_node().path]
+
+        # Retrieve the path(s) for each Page for which this user has either at least one GroupPagePermission
+        # that provides explorability, or for which the user has the 'choose' GroupPagePermission.
+        permitted_paths = Page.objects.distinct().values_list('path', flat=True).filter(
+            group_permissions__group__in=self.user.groups.all()
+        )
+        if choosable:
+            permitted_paths = permitted_paths.filter(group_permissions__permission_type='choose')
```
gasman: This permission type doesn't exist (yet...), so as it stands this has the effect of removing the breadcrumb entirely on the page chooser for non-superusers.

wagtail/wagtailcore/models.py
```diff
@@ -1377,11 +1378,64 @@ def serve_password_required_response(self, request, form, action_url):
...
         verbose_name = _('page')
         verbose_name_plural = _('pages')
 
 
+def get_closest_common_ancestor_path(request, choosable=False):
+    """
+    Global method that returns the Closest Common Ancestor for the specified User, or None if the User has no
+    explorable Page paths. Pass in choosable=True to get the CCA path of the User's choosable pages, rather
+    then their explorable pages.
+    """
```
gasman: As of #3072 this functionality now exists as part of PageQuerySet <https://github.com/torchbox/wagtail/blob/cc299d8873f1dc02e99d4fffee5b670d4c00fd15/wagtail/wagtailcore/query.py#L235>, so I'll see if we can re-use that rather than repeating the logic.

gasman:
Thanks @jeffreyckchau! I've now refactored this as proposed, in #3193.

Issue #3138, opened 2016-11-04
Root page has a privacy icon, and a 'more' menu listing revisions


nimasmi:
### Issue Summary

The root page shouldn't behave the same as even site root pages below it. Currently it has:

- a _more_ menu containing a _revisions_ entry, which takes you to `/pages/1/revisions/` _(No revisions of this page exist)_
- a _privacy_ icon, showing _public_, but this is not clickable.

<img width="1039" alt="screen shot 2016-11-04 at 14 13 23" src="https://cloud.githubusercontent.com/assets/1473980/20008712/8dae60e4-a299-11e6-9459-5204f1ce2287.png">

### Technical details

* Wagtail version: 1.7

PR #3139, opened 2016-11-04
Fix header buttons on root page

nimasmi:
Addresses #3138

- Hides 'revisions' entry on root page.
- Makes the 'more' dropdown menu not display when it is empty, which in practice will be the root page.
- Hides the 'privacy' link/icon on root page.

I was tempted to leave the 'privacy' button alone, as the fact that it's not a link (because `page_perms.can_set_view_restrictions` is `False`) meant it was harmless, but my opinion is that leaving it there is more confusing than not: 'public' suggests viewable, which the root page isn't.


gasman:
Merged in 6fda0dc01ff4470600c71a7cf6d9ebcf3535696c - thanks!

PR #3147, opened 2016-11-09
Update forms.py

boykom:

Add translation for wagtail form fields labels and help_texts

### DIFF
--- a/wagtail/wagtailforms/forms.py
+++ b/wagtail/wagtailforms/forms.py
@@ -99,8 +99,8 @@ class FormBuilder(object):
 
     def get_field_options(self, field):
         options = {}
-        options['label'] = field.label
-        options['help_text'] = field.help_text
+        options['label'] = _(field.label)
+        options['help_text'] = _(field.help_text)
         options['required'] = field.required
         options['initial'] = field.default_value
         return options


gasman:
(ref: https://groups.google.com/d/msg/wagtail/whxJoeQkbDg/11z2nu9sBAAJ)

Thanks for the PR @boykom!

I'm concerned that this change will have unpredictable effects for users who don't want to make use of translation, since there's no control over whether translation is applied or not. For example: a developer building a German-language-only site might use the word "Tag" (= Day) as a field label. This will then get fed through the English-to-German translation mechanism (assuming the site's locale is set to German) - and if Wagtail's translation files happen to contain a translation for the English word "Tag", this would cause an unwanted translation.

I'd be happier if this feature was behind a settings flag, something like:

    if getattr(settings, 'WAGTAILFORMS_USE_GETTEXT', False):
        options['label'] = _(field.label)
        options['help_text'] = _(field.help_text)
    else:
        options['label'] = field.label
        options['help_text'] = field.help_text

We would then need tests and documentation for this option - including documentation on how to create a translation file (since the usual `manage.py makemessages` method won't work when `_(...)` is called on a variable).

boykom:
Hi gasman,

Thank for the notes. I agree with you and solution you suggested. 
Should I prepare a new PR?

Kind regards, Maksim

boykom:
Also I would like to suggest to add an checkbox per field to enable translation for a particular field. In such case one can add help_text to explain how to use it.

gasman:
Actually, now that I've thought about this some more, I've realised that this doesn't really fit the intent of the wagtailforms module at all, sorry...

The purpose of wagtailforms is to allow content editors (who might be non-technical users) to create forms entirely within the admin interface, without the involvement of a programmer. With this proposed approach, a user would have to create a form first, then prepare a .po / .mo file to be committed to the codebase and deployed. The text is then effectively hard-coded into the project - if an editor wants to update a label, they need to update the .po / .mo file and re-deploy.

If your editors are all technical users, this might not be a problem - but at that point, there's no benefit to using the wagtailforms module. It would make more sense to use an ordinary Django form instead... that way, the form definition exists entirely within Python code, rather than being split between the codebase and the admin interface. (See https://github.com/gasman/wagtail-form-example for an example of doing this within Wagtail.)

Unfortunately I don't have any better suggestions for implementing translation in wagtailforms, but for us to "officially" adopt a solution, it would need to work within the goals of wagtailforms - namely, all configuration being done on the admin interface / database side, and being usable by editors with no specialist programming knowledge.

boykom:
I see. Thank you.

Issue #2379, opened 2016-03-21
Minimum length validation ListBLock


robmoorman:
It often occurs when content editors fill in StreamField blocks with invalid content, even the form is submitted correctly. A good example is a Carousel Block. They add a CarouselBlock, but it's not clear they have to add one ore more Slides and why the frontend seems broken. Another scenario would be adding Q/A's for a FAQ, except you don't want empty Categories (so at least one Q/A is required).

From this perspective it would be nice to have an optional min_length field for a ListBlock.

```
class CarouselSlideBlock(blocks.StructBlock):
    caption = blocks.CharBlock(max_length=64)
    image = ImageChooserBlock()


class CarouselBlock(blocks.StructBlock):
    slides = blocks.ListBlock(CarouselSlideBlock, min_length=1)
```

JhovaniC:
+1, yes please.

Adrian-Turjak:
No progress on this? :(

Really useful feature.

I see there is a mostly finished pull request with some unresolved comments, and also missing JS validation, but it doesn't look like too much left to do. If there isn't any activity on this I may do a new pull request based on the old one (rebased to master) and add some js validation.

robmoorman:
Hi @Adrian-Turjak that would be very much appreciated!

thibaudcolas:
I think both #3914 and #4942 are tackling this, with #4942 most likely making it into Wagtail soon-ish.

@BertrandBordage could you confirm whether it will indeed fix this? I can see it mentioned on the PR, but not on #4473 / https://www.kickstarter.com/projects/noripyt/wagtails-first-hatch.

BertrandBordage:
@thibaudcolas I confirm.

CuriousLearner:
Hello, 

Thank you for all your hard work on this. It is a much-needed feature. Is there an expected timeline for its release?

gasman:
Completed in #7334.

PR #3150, opened 2016-11-09
Listblock max_length min_length validation

robmoorman:
This pull request resolves the required max- and min length arguments for the `ListBLock`, see issue #2379.

In this case it validates the length of the list with the optional max and min arguments. Maybe we can apply this validation in an earlier stage as well; hiding the _Add another_ button with JavaScript for example.

screen shot 2016-11-09 at 22 09 22

### DIFF
--- a/docs/topics/streamfield.rst
+++ b/docs/topics/streamfield.rst
@@ -350,6 +350,8 @@ Any block type is valid as the sub-block type, including structural types:
         ('amount', blocks.CharBlock()),
     ])))
 
+The keyword arguments ``max_length`` and ``min_length`` are accepted.
+
 
 StreamBlock
 ~~~~~~~~~~~
--- a/docs/topics/streamfield.rst
+++ b/docs/topics/streamfield.rst
@@ -352,6 +352,14 @@ Any block type is valid as the sub-block type, including structural types:
 
 The keyword arguments ``max_length`` and ``min_length`` are accepted.
 
+.. code-block:: python
+
+    ('ingredients_list', blocks.ListBlock(
+        blocks.CharBlock(label="Ingredient"),
+        max_length=3,
+        min_length=1
+    ))
+
 
 StreamBlock
 ~~~~~~~~~~~
--- a/wagtail/wagtailcore/blocks/list_block.py
+++ b/wagtail/wagtailcore/blocks/list_block.py
@@ -2,23 +2,39 @@ from __future__ import absolute_import, unicode_literals
 
 from django import forms
 from django.contrib.staticfiles.templatetags.staticfiles import static
-from django.core.exceptions import ValidationError
+from django.core.exceptions import NON_FIELD_ERRORS, ValidationError
 from django.forms.utils import ErrorList
 from django.template.loader import render_to_string
 from django.utils.html import format_html, format_html_join
 from django.utils.safestring import mark_safe
+from django.utils.translation import ugettext as _
 
 from wagtail.wagtailcore.utils import escape_script
 
 from .base import Block
 from .utils import js_dict
 
-__all__ = ['ListBlock']
+__all__ = ['ListBlock', 'ListBlockValidationError']
+
+
+class ListBlockValidationError(ValidationError):
+    def __init__(self, block_errors=None, non_block_errors=None):
+        params = {}
+        if block_errors:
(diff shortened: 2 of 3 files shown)


wagtail/wagtailcore/blocks/list_block.py
```diff
@@ -113,19 +129,27 @@ def value_omitted_from_data(self, data, files, prefix):
...
-                errors.append(ErrorList([e]))
-            else:
-                errors.append(None)
+                errors[i] = ErrorList([e])
 
-        if any(errors):
+        max_length = self.list_options['max_length']
+        if max_length and max_length < len(value):
+            raise ListBlockValidationError(non_block_errors=ErrorList([
+                ValidationError(_("Maximum of %s is reached" % max_length))]))
```
timheap: This error message feels very impersonal, and not as clear as it could be. Perhaps something like "The maximum number of items is %s"

wagtail/wagtailcore/blocks/list_block.py
```diff
@@ -113,19 +129,27 @@ def value_omitted_from_data(self, data, files, prefix):
...
+
+        min_length = self.list_options['min_length']
+        if min_length and min_length > len(value):
+            raise ListBlockValidationError(non_block_errors=ErrorList([
+                ValidationError(_("Minimum of %s is required" % min_length))]))
```
timheap: Same again, perhaps a message of "The minimum number of items is %s".

wagtail/wagtailcore/tests/test_blocks.py
```diff
@@ -1447,6 +1447,34 @@ class ShoppingListBlock(blocks.StructBlock):
         )
         self.assertIn('value="chocolate"', form_html)
 
+    def test_max_length_validation_error(self):
```
timheap: Could you also add some tests for ListBlocks with a max_length/min_length where validation passes?

docs/topics/streamfield.rst
```diff
@@ -350,6 +350,8 @@ Any block type is valid as the sub-block type, including structural types:
         ('amount', blocks.CharBlock()),
     ])))
 
+The keyword arguments ``max_length`` and ``min_length`` are accepted.
```
timheap: Could you add an example of this?

```python
('dates', blocks.ListBlock(
    block.DateBlock(),
    min_length=1,
    max_length=5,
))
```

wagtail/wagtailcore/blocks/list_block.py
robmoorman: Thanks, changed the message format

wagtail/wagtailcore/tests/test_blocks.py
robmoorman: Added a test for validation with these arguments combined

docs/topics/streamfield.rst
robmoorman: Check, added an example of these arguments

robmoorman:
Thnx @timheap for the quick review! Agreed on the suggestions you made and changed them.

wagtail/wagtailcore/blocks/list_block.py
```diff
@@ -2,23 +2,39 @@
...
+
+class ListBlockValidationError(ValidationError):
+    def __init__(self, block_errors=None, non_block_errors=None):
+        params = {}
+        if block_errors:
+            params.update(block_errors)
+        if non_block_errors:
+            params[NON_FIELD_ERRORS] = non_block_errors
+        super(ListBlockValidationError, self).__init__(
+            'Validation error in ListBlock', params=params)
```
gasman: Now that we're using our own exception class, do we actually need to stuff our errors into `params` (and then unpack them again in `render_form`) - or can we define `block_errors` and `non_block_errors` as attributes on the exception object instead? I think that might make the logic a bit clearer.

robmoorman: Suppose this can be better yes, as well for `StreamBlockValidationError`. The only goal of this custom ValidationError is so separate the block from the non-blocks errors. Although I see a lot in Django `ValidationError` the `params` argument is used a lot.

gasman:
Superseded by #3914

Issue #3153, opened 2016-11-11
Add a StreamField block type that just 'exists' and has no fields


gasman:
As requested here <https://groups.google.com/d/msg/wagtail/bC0me-nRxck/5O6t-JgWBAAJ>, it would be useful to have a built-in block type that has no input fields and a stored value of `None`, but just exists as a marker to insert a particular bit of static content at that point in the stream. I suggest the name `StaticBlock`.

Assuming Benoît's implementation from that thread works as described (and he's happy for it to be included in core...), this just needs tests and docs.

gasman:
Implemented in #3154

timorieber:
I think this should be added to milestone 1.8 or isn't this relevant after closing?

gasman:
@timorieber It's not really relevant, no. The milestones exist mainly as a way for the core team to prioritise work and ensure that we don't overlook any "must-have" issues for a particular release - it's fine (and expected) for outside contributions to be handled outside of that process.

PR #3154, opened 2016-11-13
Static block

benoitvogel:
Implements #3153
- Not sure about the test_serialize and test_deserialize methods.
- Not sure about the block name either, as this block could still be used for dynamic content, using template tags in its template. But I can't find a better name, so for now...

### DIFF
--- a/docs/topics/streamfield.rst
+++ b/docs/topics/streamfield.rst
@@ -276,6 +276,33 @@ EmbedBlock
 
 A field for the editor to enter a URL to a media item (such as a YouTube video) to appear as embedded media on the page. The keyword arguments ``required``, ``max_length``, ``min_length`` and ``help_text`` are accepted.
 
+StaticBlock
+~~~~~~~~~~~
+
+``wagtail.wagtailcore.blocks.StaticBlock``
+
+A block which doesn't have any fields, thus passes no particular values to its template during rendering. This can be useful if you need the editor to be able to insert some content which is always the same or doesn't need to be configured within the page editor, such as an address, embed code from third-party services, or more complex pieces of code if the template uses template tags.
+
+By default, some default text (which contains the ``label`` keyword argument if you pass it) will be displayed in the editor interface, so that the block doesn't look empty. But you can also customise it entirely by passing a text string as the ``admin_text`` keyword argument instead:
+
+.. code-block:: python
+
+    blocks.StaticBlock(
+        admin_text='Latest posts: no configuration needed.',
+        # or admin_text=mark_safe('<b>Latest posts</b>: no configuration needed.'),
+        template='latest_posts.html')
+
+``StaticBlock`` can also be subclassed to produce a reusable block with the same configuration everywhere it is used:
+
+.. code-block:: python
+
+    class LatestPostsStaticBlock(blocks.StaticBlock):
+        class Meta:
+            icon = 'user'
+            label = 'Latest posts'
+            admin_text = '{label}: configured elsewhere'.format(label=label)
+            template = 'latest_posts.html'
+
 
 Structural block types
 ----------------------
--- a/wagtail/wagtailcore/blocks/static_block.py
+++ b/wagtail/wagtailcore/blocks/static_block.py
@@ -1,0 +1,27 @@
+from __future__ import absolute_import, unicode_literals
+
+from django.utils.translation import ugettext_lazy as _
+
+from .base import Block
+
+__all__ = ['StaticBlock']
+
+
+class StaticBlock(Block):
+    """
+    A block that just 'exists' and has no fields.
+    """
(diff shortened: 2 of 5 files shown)


wagtail/wagtailcore/blocks/static_block.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    def render_form(self, value, prefix='', errors=None):
+        admin_html = getattr(self, 'admin_html', '')
+        return mark_safe(admin_html)
+
+    def value_from_datadict(self, data, files, prefix):
+        return None
+
+    class Meta:
+        abstract = True
```
timheap: `Meta.abstract` does not have any meaning to blocks. The `abstract` attribute is only meaningful on models. This should be removed.

docs/topics/streamfield.rst
```diff
@@ -276,6 +276,26 @@ EmbedBlock
...
+
+.. code-block:: python
+
+    class PostsStaticBlock(blocks.StaticBlock):
+        def __init__(self, *args, **kwargs):
+            super(PostsStaticBlock, self).__init__(*args, **kwargs)
+            self.admin_html = "<b>{}</b> - No configuration needed".format(self.meta.label)
+
+        class Meta:
+            label="Latest posts"
```
timheap: This is missing spaces around the `=` signs. Code examples should still match the Wagtail style guide.

wagtail/wagtailcore/blocks/static_block.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    A block that just 'exists' and has no fields.
+
+    It's an abstract model, so it has to be subclassed.
+    Subclasses should define self.admin_html in order to
+    display something in the admin even though the block
+    has no fields.
+    """
+    def render_form(self, value, prefix='', errors=None):
+        admin_html = getattr(self, 'admin_html', '')
```
timheap: Providing a useful default would help developers. An empty string will never be appropriate, and could cause developers to think something is broken. A useful default will suite some simple cases, and will not look broken even if the developer eventually wants to change the message. Assuming that the `admin_html` attribute is moved to `Meta` as I suggested above, something like this would work:

```python
def render_form(self, value, prefix='', errors=None):
    if self.meta.admin_html is None:
        return _('{label}: This block has no options').format(label=self.label)
    return self.meta.admin_html
```

wagtail/tests/testapp/blocks.py
```diff
@@ -30,3 +30,13 @@ class SectionBlock(blocks.StructBlock):
     class Meta:
         icon = "form"
         template = 'tests/blocks/section_block.html'
+
+
+class PostsStaticBlock(blocks.StaticBlock):
+    def __init__(self, *args, **kwargs):
+        super(PostsStaticBlock, self).__init__(*args, **kwargs)
+        self.admin_html = "<b>{}</b> - This block doesn't need to be configured, it will be displayed automatically".format(self.meta.label)
```
timheap: `self.label` should be used over `self.meta.label`. `self.label` uses `self.meta.label` if it is set, otherwise it generates a label from the name of the block.

benoitvogel:
Thanks @timheap for your useful review. I totally agree with you. Here is an updated version.

benoitvogel:
The only drawback I can think of is you cannot write this if you want to insert self.label inside admin_html:

```
class SubStaticBlock(blocks.StaticBlock):
    class Meta:
        # self.label isn't available in Meta
        admin_html = '{label} - customised text'.format(self.label)


class MyStreamBlock(StreamBlock):
    my_sub_block = blocks.SubStaticBlock(label='My Sub Block')
```

But I think it doesn't matter, as developers can still subclass `StaticBlock` and override `__init__()` and `render_form()` to fit their needs using the same kind of hack I used in the first place.

timheap:
The changes look good, thanks for that!

One final thought regarding `admin_html`: the convention in Wagtail / Django is to *not* `mark_safe()` these sorts of strings by default. This will prevent any accidentally invalid HTML in the admin. A developer can always `mark_safe()` their string if they need to embed HTML in the admin.

If the `admin_html` attribute is no longer `mark_safe()`, `admin_html` is no longer a good name. Perhaps `admin_text`? Developers can then do `admin_text=mark_safe("...")` if they need to, or override the `render_form()` method to print out anything they need.

timheap:
Also: The style errors that are causing drone failures have been fixed in `master`. If you rebase on top of `master` the drone tests should start passing again.

benoitvogel:
Ok @timheap, thanks again. Rebased/squashed my commits into one, I hope it's OK.

gasman:
Merged in 2fba07ad8ed326ceabedb469fa18037d8d517299 - thanks @benoitvogel, and thanks @timheap for the review!

PR #3157, opened 2016-11-14
Use capfirst() in Page.get_verbose_name()

kaedroho:
Just a minor cleanup. ~~This shouldn't have any behavioural effect on the ``get_verbose_name`` method~~. Much more expressive than before though.

**EDIT:** as pointed out in https://github.com/torchbox/wagtail/pull/3157#issuecomment-260325582 this will  only capitalise the first word instead of every word as before (so "blog page" becomes "Blog page" instead of "Blog Page"). We've agreed that the new format is slightly better.

### DIFF
--- a/wagtail/wagtailadmin/tests/test_pages_views.py
+++ b/wagtail/wagtailadmin/tests/test_pages_views.py
@@ -343,13 +343,13 @@ class TestPageCreation(TestCase, WagtailTestUtils):
         response = self.client.get(reverse('wagtailadmin_pages:add_subpage', args=(self.root_page.id, )))
         self.assertEqual(response.status_code, 200)
 
-        self.assertContains(response, "Simple Page")
+        self.assertContains(response, "Simple page")
         # List of available page types should not contain pages with is_creatable = False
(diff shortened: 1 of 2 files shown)


tomdyson:
The previous behaviour capitalises the first letter of each word ("blog page" -> "Blog Page") but `capfirst()` capitalises the first character of the whole string (blog page" -> "Blog page").

https://docs.djangoproject.com/en/1.10/ref/templates/builtins/#capfirst

tomdyson:
Note the docstring for the current behaviour is incorrect. FWIW I am happy with the output of `capfirst()`.

gasman:
Some tests are now failing due to the change in capitalisation rules, and will need to be updated.

(Django's `title` template filter is closer to the old behaviour, but doesn't exist as a standalone utility function; it's based on Python's built-in `title()` method on strings, but adds some fixes that `title()` gets wrong, such as capitalising the S after an apostrophe. FWIW I slightly prefer the capfirst version...)

kaedroho:
Thanks! I've tweaked the description to point out the change in format and fixed the tests.

PR #3158, opened 2016-11-14
Improve TranslatedField example

xr09:
Avoid allocating two variables when only need one.


gasman:
Looks good to me, thanks! Merged in 5a26e22c0f5f96b441ecfcdf23721f4b48b224bb.

PR #3159, opened 2016-11-14
Deprecate ORM model capabilities of wagtailimages.Filter

gasman:
Second stage of #2881, following on from #2910.

It turns out that we can't drop Filter as a model just yet, because Django's migrations engine doesn't seem to be able to cope with moving a foreign key out of a unique_together constraint at the same time as dropping the corresponding model. So instead, we'll deprecate it as far as possible, and save the (hopefully) final schema changes for 1.9:

* the `filter` field of Rendition is now disused in favour of `filter_spec` (which should be fully populated as of 1.7); `filter` is now nullable, while `filter_spec` is now non-null, and the unique_together constraint has been moved to `filter_spec` to ensure that `get_or_create` lookups still behave correctly. (Users with custom image models will need to perform this step manually on their own Rendition models.)
* `Filter` _is_ still a model, but attempting to use it as anything other than a transient in-memory object (i.e. calling `save()` or accessing `Filter.objects`) will throw a deprecation warning. (Note that the deprecation in this case is not going through our usual two-release cycle - it's an internal refactor, not a deprecation in the usual sense, and the fact that we're giving people one release's grace, rather than none, is a bonus...)
* Existing code in wagtailimages and modeladmin has been updated to avoid ORM operations on `Filter`.

Hopefully all that remains for 1.9 is to drop the `filter` column from `AbstractRendition`, and make `Filter` not inherit from `models.Model`.


gasman:
Merged in 2c9805dc3fa37e9bf27fc7593de4465673faffca

Issue #3146, opened 2016-11-08
Cloudflare's v1 API support is ending Wednesday, November 9


albertoconnor:
### Issue Summary

Cloudflare's v1 API support is ending Wednesday, November 9 (https://www.cloudflare.com/docs/client-api/)

This should only effect contrib/wagtailfrontendcache

https://github.com/torchbox/wagtail/blob/master/wagtail/contrib/wagtailfrontendcache/backends.py#L71

I will be working on a pull request, but I likely won't be able to complete it until Tuesday November 15.

PR #3161, opened 2016-11-15
Fix Issue #3146 – Cloudflare's v1 API support is ending Wednesday, November 9

albertoconnor:
I look forward to you feedback to further improve this change.

I didn't write new tests for the actual request to CloudFlare's server. There wasn't a test before. If additional tests are desirable, advice on how to mock correctly would be very welcome.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,6 +31,7 @@ install_requires = [
     "html5lib>=0.999,<1",
     "Unidecode>=0.04.14",
     "Willow>=0.4,<0.5",
+    "requests>=2.11.1",
 ]
 
 # Testing dependencies
--- a/docs/reference/contrib/frontendcache.rst
+++ b/docs/reference/contrib/frontendcache.rst
@@ -65,7 +65,10 @@ Cloudflare
 
 Firstly, you need to register an account with Cloudflare if you haven't already got one. You can do this here: `Cloudflare Sign up <https://www.cloudflare.com/sign-up>`_
 
-Add an item into the ``WAGTAILFRONTENDCACHE`` and set the ``BACKEND`` parameter to ``wagtail.contrib.wagtailfrontendcache.backends.CloudflareBackend``. This backend requires two extra parameters, ``EMAIL`` (your Cloudflare account email) and ``TOKEN`` (your API token from Cloudflare).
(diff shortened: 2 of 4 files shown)


gasman:
Thanks @albertoconnor! This looks good to me, and seems to be working well.

The one thing that needs more thought is the addition of the `ZONEID` setting. As it stands, this is a backwards-incompatible change: if we make a 1.7.1 release with this fix, then someone could upgrade to 1.7.1 without updating their settings (either because they didn't read the release notes, or because the upgrade happened automatically due to having `wagtail>=1.7,<1.8` in their requirements) and find that publishing fails with a KeyError on ZONEID. It would be acceptable for us to make a breaking change like this in the 1.8 release, but not 1.7.1. So, we have a few options:

* The worst, but simplest option: Change https://github.com/torchbox/wagtail/pull/3161/files#diff-8291535514a0b8a978d1586fe96f8353R68 to `self.cloudflare_zoneid = params.pop('ZONEID', '')` so that cache purging fails silently when ZONEID is not set.
* Release this fix as part of Wagtail 1.8, which is due at the beginning of next month. It looks like Cloudflare's v1 API is still working _for now_, so we might be lucky and the v1 API might continue working until 1.8 is ready :-)
* Update the code to check for the presence of the ZONEID setting. If it's not found, throw a `RemovedInWagtail19Warning` (to tell the developer that they need to add ZONEID to their settings), and fall back on the old behaviour (posting to the v1 API). Then, in Wagtail 1.9, we will remove this fallback behaviour.

The last option is definitely the one I prefer, but it's also the most work. What do you think?

setup.py
```diff
@@ -31,6 +31,7 @@
     "html5lib>=0.999,<1",
     "Unidecode>=0.04.14",
     "Willow>=0.4,<0.5",
+    "requests>=2.11.1",
```
gasman: It would probably be better to have `"requests>=2.11.1,<3.0"` here, to guard against the possibility that an incompatible 3.0 version of `requests` is released in future.

(Yes, a few of the other dependencies above also have this problem. We should really fix those too :-) )

albertoconnor:
Hi @gasman thanks for the feedback.

I generally agree re backward compatibility and the using the best approach.

Does the v1 actually work or does it just silently fail? I didn't test it exhuaustively but my feeling is it returned like it was working but didn't actually invalidate. I could be wrong though.

I think we need to do something more proactive to warn the subset of users who use cloudflare that making this change is in their best interest! Possibly a system check or logging a warning.

Right now the old behavior is undefined at best.

albertoconnor:
I think waiting until 1.8 is fine. We we be running off my own fork until then.

gasman:
OK - merged in a5b544631bbde21b28ff749dcd53d35e5f5bf64d, for release in 1.8. Thanks!

PR #3162, opened 2016-11-15
Fix #3134 - search_garbage_collect command does not work without wagtailsearchpromotions contrib

### LINKED ISSUE #3134: search_garbage_collect management command does not work if wagtailsearchpromotions is not included in INSTALLED_APPS
### Issue Summary

If `wagtailsearchpromotions` is not included in `INSTALLED_APPS` then the `search_garbage_collect` management command fails here: https://github.com/torchbox/wagtail/blob/stable/1.7.x/wagtail/wagtailsearch/models.py#L44 because there is no `editors_picks` field.

### Steps to Reproduce

1. Create a Wagtail site without `wagtail.contrib.wagtailsearchpromotions` in `INSTALLED_APPS`.
2. run the `search_garbage_collect` management command.

Command fails with `django.core.exceptions.FieldError: Cannot resolve keyword 'editors_picks' into field. Choices are: daily_hits, id, query_string`.

ellmetha:
Changes include:
* fixing the FieldError exception that can occur as described in #3134
* adding tests for the "search_garbage_collect" command

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -184,6 +184,7 @@ Contributors
 * Thejaswi Puthraya
 * Benoît Vogel
 * Manuel E. Gutierrez
+* Morgan Aubert
 
 Translators
 ===========
--- a/wagtail/wagtailsearch/models.py
+++ b/wagtail/wagtailsearch/models.py
@@ -41,7 +41,9 @@ class Query(models.Model):
         """
         Deletes all Query records that have no daily hits or editors picks
         """
-        cls.objects.filter(daily_hits__isnull=True, editors_picks__isnull=True).delete()
+        extra_filter_kwargs = {'editors_picks__isnull': True, } if hasattr(cls, 'editors_picks') \
+            else {}
+        cls.objects.filter(daily_hits__isnull=True, **extra_filter_kwargs).delete()
 
     @classmethod
     def get(cls, query_string):
--- a/wagtail/wagtailsearch/tests/test_queries.py
+++ b/wagtail/wagtailsearch/tests/test_queries.py
@@ -1,10 +1,15 @@
 
 from __future__ import absolute_import, unicode_literals
 
+import datetime
+
+from django.conf import settings
 from django.core import management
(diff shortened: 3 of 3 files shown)


gasman:
Hi @ellmetha - this looks really good, thanks!

Unfortunately I don't think it's possible to modify the `INSTALLED_APPS` list using override_settings - if I revert the fix to wagtailsearch/models.py, `test_garbage_collect_command_works_without_promoted_searches` still passes (indicating that it's still behaving as if wagtailsearchpromotions is installed). As I understand it, override_settings can only affect settings that are read from `django.conf.settings` after override_settings has been called - and, of course, by that time INSTALLED_APPS has already been processed (and thus the `editors_picks` property already exists on the Query model).

Since I can't see a good way of testing the behaviour without wagtailsearchpromotions - other than having a completely separate test runner just for this one test - I'll merge this PR with this test omitted.

gasman:
Merged in 51b4f0f53d93a69f3beaba67e3211cd001ff4087 - thanks!

ellmetha:
Cool! Actually I think that we could've modified the test to properly test how the command behaves without ``wagtailsearchpromotions``. We could've used the Django ``TestCase.available_apps`` attribute. ;-) I can provide a new PR adding the test if you want.

gasman:
The list of "gotchas" at https://docs.djangoproject.com/en/1.10/topics/testing/advanced/#django.test.TransactionTestCase.available_apps makes me think that `available_apps` is going to be more of a liability than a benefit... also, from the description there it sounds like it's only designed to optimise away some database operations, rather than making any deep changes to the relations between apps/models. I think we're fine without a test here, really.

ellmetha:
The ``available_apps`` attribute has an impact on the registry and the models' ``_meta`` fields cache. It turns out that if you remove the wagtailsearchpromotions app from the installed apps (using the ``available_apps`` attribute) the ``editors_picks`` fields won't be returned by the ``Query._meta.get_fields`` method. This would allow to test the desired behaviour.

I note that the test is not needed, though. :smile:

PR #3168, opened 2016-11-17
Fix bug where help_text covers controls for stream-field

strindhaug:
Fix for https://github.com/torchbox/wagtail/issues/2792

Change the position of help text to relative (rather than absolute) for stream fields so that the help text doesn't overlap with the controls on the top block in the stream field.

Since this will create a bit extra space between the stream field title and the content, the help text uses full width so that it in almost all cases will only be one line.


gasman:
Merged in b006b1958259555dca5a51d5a1cef4e6617cc9c7 - thanks!

PR #3169, opened 2016-11-18
Fix ForeignKey selectbox in ChooserBlock (2)

### LINKED ISSUE #2218: PageChooserBlock/ImageChooserBlock values go null when restoring from loaddata

mvantellingen:
This adds a test for the changes in #2714

> If you have a related model in your page through a ChooserBlock the selected value isn't shown on a second load (after save). This change should fix that.

> Fixes #2218

### DIFF
--- a/wagtail/wagtailcore/blocks/field_block.py
+++ b/wagtail/wagtailcore/blocks/field_block.py
@@ -510,6 +510,13 @@ class ChooserBlock(FieldBlock):
             except self.target_model.DoesNotExist:
                 return None
 
+    def value_for_form(self, value):
+        # When reloading a page it would return an instance, we want an ID.
+        if isinstance(value, self.target_model):
+            return value.pk
+        else:
+            return value
+            
     def clean(self, value):
         # ChooserBlock works natively with model instances as its 'value' type (because that's what you
         # want to work with when doing front-end templating), but ModelChoiceField.clean expects an ID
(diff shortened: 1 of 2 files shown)


gasman:
Looking into this now. Pretty sure this is unrelated to #2218 though...

gasman:
OK, I don't think this is strictly a bug - ChooserBlock was only designed to work with [AdminChooser](https://github.com/torchbox/wagtail/blob/d21171c1b0bbec7e497e12c2c7070c98ea0cdbf1/wagtail/wagtailadmin/widgets.py#L75)-derived widgets, which _do_ work correctly when passed an instance - it's only the `Select` widget that doesn't.

Normally I wouldn't mind adding this as an enhancement, to make ChooserBlock more widely useful... however, this results in a performance penalty on the existing AdminChooser-based blocks, since they have to convert the ID back into an instance again in AdminChooser.get_instance_and_id. Effectively, this reverses the performance gain we made from introducing `bulk_to_python`... :-(

As a fairly hacky workaround, I guess we can check the widget type in `value_for_form` and only convert to an ID if it's of type `Select` (or, conversely, only leave it as an instance if it descends from `AdminChooser`). Perhaps there's a nicer way...?

gasman:

PR #3174, opened 2016-11-24
Fix errors by settings context processor in out-of-request templates

vdboor:
When templates are rendered outside the request cycle, the `settings` context processor fails because `request.site` is not created because `SiteMiddleware` was not processed.

This happens whenever requests are mocked for rendering. For example email rendering by third party packages, middleware that runs early or unittests that mock a request object without taking all middleware into account.

### DIFF
--- a/wagtail/contrib/settings/context_processors.py
+++ b/wagtail/contrib/settings/context_processors.py
@@ -55,4 +55,12 @@ class SettingModuleProxy(dict):
 
 
 def settings(request):
-    return {'settings': SettingsProxy(request.site)}
+    site = getattr(request, 'site', None)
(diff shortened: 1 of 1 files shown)


vdboor:
It might be useful to have the context_processor emit a warning whenever this happens. In a large project ported to wagtail, this `settings` middleware caused various rendering parts in the site to break.

gasman:
Thanks @vdboor!

I think it's OK for this to fail silently - that's the standard behaviour for missing variables on a template.

gasman:
Merged in d21171c1b0bbec7e497e12c2c7070c98ea0cdbf1.

PR #3177, opened 2016-11-25
Fix #3126 - Fix double-escaping of embedded video urls

### LINKED ISSUE #3126: Embedded video urls get doubled escaped on save actions
### Issue Summary

When embedding a video url with at least one Ampersand character **"&"** in a **RichTextBlock** inside a **StreamField**, every **"&"** char inside the url get's replaced with **"&amp;"** every time a new revision is saved.

Example: 
Each time a new draft is saved the following transformation happens:
1. https://www.youtube.com/watch?v=uKpfO331DY4&t=10
2. https://www.youtube.com/watch?v=uKpfO331DY4&amp;t=10
3. https://www.youtube.com/watch?v=uKpfO331DY4&amp;amp;t=10
4. https://www.youtube.com/watch?v=uKpfO331DY4&amp;amp;amp;t=10

This does not only make the url uglier, it also increases the url's character length. 

I could reproduce this with two different setups:

1: An older production system I cannot upgrade to a newer version of wagtail

- wagtail==1.5.3
- django==1.8.14
- psql (PostgreSQL) 9.3.14

In this setup, it is impossible to enter a Page's edit page if an embedded video url exceeded a 200 chars length, causing a **DataError** Exception to be thrown:

`DataError: value too long for type character varying(200)`

2: Clean wagtail installation

- wagtail==1.7
- django==1.10.1
- sqlite db

The **DataError** Exception will not be thrown in this setup but the url still grows each time you hit the save button.

### Steps to Reproduce

1. Embed a video with at least to query params which are separated with a "&" char (like https://www.youtube.com/watch?v=uKpfO331DY4&t=10) in RichtTextBlock in Streamfield.
2. Save draft
3. Check url

ellmetha:
Ref: #3126 

Changes include:
* Marking the embedded video urls as safe in the 'embed_editor.html' template

### DIFF
--- a/wagtail/wagtailcore/rich_text.py
+++ b/wagtail/wagtailcore/rich_text.py
@@ -155,7 +155,8 @@ def extract_attrs(attr_string):
     """
     attributes = {}
     for name, val in FIND_ATTRS.findall(attr_string):
-        attributes[name] = val
+        # Replaces '&amp' per '&' here in order to prevent extracted URLs to be double-escaped.
+        attributes[name] = val.replace('&amp;', '&')
     return attributes
 
 


gasman:
Hi @ellmetha - thanks for the PR!

I think this will need some more careful analysis before we can merge this. Marking the URLs as `|safe` is not the correct fix if the URL is coming directly from user input - URLs appearing inside HTML content _should_ be escaped for it to be considered valid HTML, and there are potential security issues if they aren't (for example, a URL provider might find a way to sneak a `<script>` tag into the URL).

The real problem is that we're failing to be explicit about which part of the code is responsible for escaping/unescaping the URL: adding `|safe` to the embed_editor.html template _could_ be correct if we decide that [`embed_to_editor_html`](https://github.com/torchbox/wagtail/blob/f960753d743d77233802da800ad0f55f569f640c/wagtail/wagtailembeds/format.py#L31) accepts an escaped URL, but then the call to `get_embed` is probably incorrect. I suspect the 'proper' fix is for [`extract_attrs`](https://github.com/torchbox/wagtail/blob/f960753d743d77233802da800ad0f55f569f640c/wagtail/wagtailcore/rich_text.py#L152) to perform the `&amp;` to `&` conversion.

Related: https://github.com/torchbox/wagtail/pull/2669#issuecomment-224013584

ellmetha:
> adding |safe to the embed_editor.html template could be correct if we decide that embed_to_editor_html accepts an escaped URL

That was my postulate. :wink:

@gasman I've made the change you proposed.

gasman:
@ellmetha Unfortunately the problem here is much deeper - changing `extract_attrs` without dealing with the wider context is just going to make things even more inconsistent than they are already. We need to make a formal decision about whether `extract_attrs` returns HTML-escaped strings or unescaped strings, **and ensure that all code elsewhere in Wagtail follows that standard**. (Currently the standard is for the strings to be HTML-escaped, but based on the problems this causes here and in https://github.com/torchbox/wagtail/pull/2669#issuecomment-224013584, switching to unescaped strings is probably the better long-term plan.) Simply replacing `&amp;` -> `&`, without dealing with other HTML entities, means that we're not following a consistent rule at all.

gasman:
Superseded by #3882

PR #3181, opened 2016-11-28
Fix some style errors in the documentation

timheap:
All instances of `::` for marking code have been replaced with a more relevant `.. code-block::` line. The are a few `::` instances left in the docs, but only for marking plain text preformatted code.

`.. code-block:: console` has been used for all shell commands. According to [the pygments documentation](http://pygments.org/docs/lexers/), this is the correct lexer for interactive shell sessions, as opposed to shell scripts.

All trailing whitespace has been trimmed, and tabs replaced with spaces.

`versionadded` notes for old versions have been dropped - there was one in the docs from version 0.7! I updated [Creating a new Wagtail release](https://github.com/torchbox/wagtail/wiki/Creating-a-new-Wagtail-release) to remind the release manager to remove old `versionadded` notes.

As a side note, should [Creating a new Wagtail release](https://github.com/torchbox/wagtail/wiki/Creating-a-new-Wagtail-release) be moved in to the repo so that it is version controlled, and can have PRs made against it?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,7 +5,7 @@ end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-[*.{css,js,json,py,yml}]
+[*.{css,js,json,py,yml,rst}]
 indent_style = space
 
 [*.{js,py}]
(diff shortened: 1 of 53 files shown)


timheap:
Apparently I accidentally pushed this to `torchbox/wagtail` not `takeflight/wagtail` at first. I deleted the branch from `torchbox/wagtail` but Travis had already started the build, which proceded to fail when trying to clone in subsequent test runs. Sad times.

gasman:
Thanks! Merged in 16213db0f8c6f9512dd6b9ee5f2234dfea4cb38f.

PR #3182, opened 2016-11-28
[RFC] Added accept_request method to PageViewRestriction

kaedroho:
This allows custom code to check whether a PageViewRestriction will pass
a request or not without needing to depend on any implementation
details.

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -31,6 +31,7 @@ from django.utils.translation import ugettext_lazy as _
 from modelcluster.models import ClusterableModel, get_all_child_relations
 from treebeard.mp_tree import MP_Node
 
+from wagtail.utils.compat import user_is_authenticated
 from wagtail.wagtailcore.query import PageQuerySet, TreeQuerySet
 from wagtail.wagtailcore.signals import page_published, page_unpublished
 from wagtail.wagtailcore.url_routing import RouteResult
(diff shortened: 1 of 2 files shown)


tomdyson:
@kaedroho could you please give an example of when this would be useful?

kaedroho:
Yep, I currently have this bit of code in a client site:

```python
@register.assignment_tag(takes_context=True)
def user_can_see_page(context, page):
    request = context['request']
    restrictions = page.get_view_restrictions()

    if restrictions:
        passed_restrictions = request.session.get('passed_page_view_restrictions', [])
        for restriction in restrictions:
            if restriction.restriction_type == PageViewRestriction.PASSWORD:
                if restriction.id not in passed_restrictions:
                    return False
            elif restriction.restriction_type == PageViewRestriction.LOGIN:
                if not user_is_authenticated(request.user):
                    return False
            elif restriction.restriction_type == PageViewRestriction.GROUPS:
                if not request.user.is_superuser:
                    current_user_groups = request.user.groups.all()

                    if not any(group in current_user_groups for group in restriction.groups.all()):
                        return False

    return True
```

But I think this logic really belongs in the ``PageViewRestriction`` model itself. This means that any future changes will automatically work with this site (currently, using an unsupported restriction type will cause this tag to pass).


Here's an example of this template tag written with the ``accept_request`` method:

```python
@register.assignment_tag(takes_context=True)
def user_can_see_page(context, page):
    request = context['request']
    restrictions = page.get_view_restrictions()

    for restriction in restrictions:
        if not restriction.accept_request(request):
            return False

    return True
```

All the custom logic has now been removed, which is great for forwards compatibility.

gasman:
Looks good to me!

The 'respond with an access-denied response' code that's been left behind in wagtail_hooks would also be a good candidate for turning a method on PageViewRestriction... but that's something for another time.

gasman:
Merged in ba4119bd261668868f2c995114599a8bdfdf01a5.

tomdyson:
Thanks for the explanation @kaedroho. Shouldn't we document the `accept_request` method, for future template tag authors?

PR #3184, opened 2016-11-28
Add BlockQuoteBlock to core block types

shacker:
As discussed at https://github.com/torchbox/wagtail/issues/532

I ended up subclassing TextBlock rather than CharBlock because the input needs to be multi-line (I was still able to shorten the definition this way).

One bit I'm not sure of - because `wagtailcore.blocks.__init__` imports * from all of the block modules, I should be able to import `blocks.BlockQuoteBlock`, as with other core block types. But for some reason I'm forced to import `blocks.field_block.BlockQuoteBlock` - not sure whether this means I'm missing something.

### DIFF
--- a/wagtail/wagtailcore/blocks/field_block.py
+++ b/wagtail/wagtailcore/blocks/field_block.py
@@ -141,12 +141,6 @@ class TextBlock(FieldBlock):
 
 class BlockQuoteBlock(TextBlock):
 
-    def __init__(self, required=True, help_text=None, max_length=None, min_length=None, **kwargs):
-        super(BlockQuoteBlock, self).__init__(**kwargs)
-
-    def get_searchable_content(self, value):
-        return [force_text(value)]
-
     def render_basic(self, value, context=None):
         if value:
             return format_html('<blockquote>{0}</blockquote>', value)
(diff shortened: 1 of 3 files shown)


gasman:
Thanks @shacker! You'll need to add BlockQuoteBlock to the `block_classes` list at the end of the file - the definition of `__all__` determines which items get imported on an `import *`.

wagtail/wagtailcore/blocks/field_block.py
```diff
@@ -139,6 +139,24 @@ class Meta:
         icon = "pilcrow"
 
 
+class BlockQuoteBlock(TextBlock):
+
+    def __init__(self, required=True, help_text=None, max_length=None, min_length=None, **kwargs):
+        super(BlockQuoteBlock, self).__init__(**kwargs)
```
gasman: Are the `required` / `help_text` / `max_length` / `min_length` parameters respected here? It looks like those are just thrown away in this definition.

You can probably just omit `__init__` and `get_searchable_content` here, because they aren't doing anything that `TextBlock` isn't doing already.

gasman:
Merged in 28827758420ac2c510a21fb5316453306e5f8876 - thanks!

PR #3185, opened 2016-11-28
Fix for security risk of mutable list arguments in helpers.py.

shoefler:
The work for the code change was done in conjunction with three other students at the University of Nebraska at Omaha for a Master's level software assurance class project.  I scanned the Wagtail code using Pylint using swamp.org to identify security weaknesses.  The scan identified an issue with functions in helpers.py.  The issue found is if an list argument of a function is set to [] as the default, the previous call to the function could assign a value to the argument.  The next call to the function will use the assigned value as the default instead of [].  Setting the default of the lists arguments to None fixes the weakness.   See the "Important warning" in [python's documentation](https://docs.python.org/3/tutorial/controlflow.html#default-argument-values) for more information and an example.  The pull request sets the list defaults to None.  

After the code change no additional tests failed and the issue for [] was removed in a Pylint scan.

### DIFF
--- a/wagtail/contrib/modeladmin/helpers.py
+++ b/wagtail/contrib/modeladmin/helpers.py
@@ -242,12 +242,20 @@ class ButtonHelper(object):
         self.permission_helper = view.permission_helper
         self.url_helper = view.url_helper
 
-    def finalise_classname(self, classnames_add=[None], classnames_exclude=[None]):
+    def finalise_classname(self, classnames_add=None, classnames_exclude=None):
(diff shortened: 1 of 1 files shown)


gasman:
Thanks @shoefler! Merged in 7994d19f6514019a392a40219251631e57ef6446.

(In this case, I don't think the bug would ever be triggered here, since the lists never get mutated within the lifetime of those methods, only used to build new lists... We definitely want to follow best practice, though.)

PR #3186, opened 2016-11-29
Add the `alters_data` attribute on `Page.copy()`.

vdboor:
This makes sure the variable can't be printed in the template, performing an accidental copy of the page. That also happened when using the `{% print page %}` tag from my django-debugtools package.


gasman:
Thanks @vdboor! Merged in f960753d743d77233802da800ad0f55f569f640c.

PR #3188, opened 2016-11-29
Cache `Site.get_site_root_paths` in the request scope

### LINKED ISSUE #2916: Wagtail can make MANY requests to wagtail_site_root_paths

gasman:
Fixes #2916

Avoids repeated cache / database lookups on repeated calls to `{% pageurl %}`, by caching the result of `Site.get_site_root_paths` in the request scope. Since the actual lookup happens in `Page.get_url_parts`, which is called via `Page.relative_url`, I've had to do some creative re-plumbing to pass that cached result around in a new `hints` kwarg.

Since it's common for page models to override either/both of `get_url_parts` and `relative_url`, this in turn means we need to put the old `hints`-less signature through the deprecation process.

### DIFF
--- a/docs/releases/1.8.rst
+++ b/docs/releases/1.8.rst
@@ -132,3 +132,23 @@ The ``wagtail.contrib.wagtailfrontendcache.backends.CloudflareBackend`` module h
     }
 
 For details of how to obtain the zone identifier, see `the Cloudflare API documentation <https://api.cloudflare.com/#getting-started-resource-ids>`_.
+
+
+``Page.get_url_parts`` and ``Page.relative_url`` now accept a ``hints`` kwarg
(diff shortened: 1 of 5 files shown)


mvantellingen:
Hi @gasman, still needed to reply on this. This looks really nice, so +1 from me.

I was thinking that it might also be a good idea to make the cache backend for internal caching configurable so that users can use a request based caching backend for things like this.

gasman:
Superseded by #3354

PR #3192, opened 2016-12-01
Convert explorer breadcrumb into a custom template tag

gasman:
Precursor to refactoring #3133 - split off the chooser/explorer versions of the breadcrumb into separate templates, and convert the explorer version into a template tag (which means we can generate the page queryset more intelligently, and don't need a special case in the template for `include_self`)


gasman:
Merged as part of #3193

Issue #2196, opened 2016-02-04
Add a hook for filtering the page listing on the explorer


gasman:
Example scenario - we have a section containing many pages with a 'language' field, and want to filter the explorer listing so that users only see pages in their own language.

Suggestion: In the wagtailadmin/pages/index view, after we've retrieved the page queryset (`parent_page.get_children()`...), incorporate a hook function that passes the parent page, page queryset and the request, and returns a new (filtered) queryset to use:

    @hooks.register('filter_explorer_pages')
    def filter_pages_by_language(parent_page, pages, request):
        if parent_page.slug = 'news':
            pages = pages.filter(language=request.user.language)
            # the above would need some tweaking, since 'language' isn't a field on Page...

        return pages

gasman:
Fixed in #2219

PR #3203, opened 2016-12-06
Add construct_explorer_page_queryset hook to explorer nagivation and page chooser

x-warrior:
This pull request add `construct_explorer_page_queryset` hook to navigation (explorer_nav) and to page chooser (browse and search).

Using gasman scenario from (#2196):
> Example scenario - we have a section containing many pages with a 'language' field, and want to filter the explorer listing so that users only see pages in their own language.

We also want our users to see only the pages on their language on navigation and chooser views.

A few notes about the PR:
- Should we rename this hooks with something different? One for each case? So we don't affect users which already uses this hook? 
- I have added `request` param to `get_navigation_menu_items` as named with None as default but all Wagtail references were updated, I was afraid to break stuff for other people that have customised Wagtail. 
- On chooser `search` it became a little bit more complex because we're Searching and searches doesn't have the pages as queryset as the other cases, so I needed to standardise the pages object to call the hook with the same type of object (queryset).

### DIFF
--- a/wagtail/wagtailadmin/navigation.py
+++ b/wagtail/wagtailadmin/navigation.py
@@ -2,8 +2,8 @@ from __future__ import absolute_import, unicode_literals
 
 from django.db.models import Q
 
-from wagtail.wagtailcore.models import Page
 from wagtail.wagtailcore import hooks
+from wagtail.wagtailcore.models import Page
 
 
 def get_pages_with_direct_explore_permission(user):
--- a/wagtail/wagtailadmin/navigation.py
+++ b/wagtail/wagtailadmin/navigation.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, unicode_literals
 from django.db.models import Q
 
 from wagtail.wagtailcore.models import Page
+from wagtail.wagtailcore import hooks
 
 
 def get_pages_with_direct_explore_permission(user):
@@ -17,7 +18,7 @@ def get_pages_with_direct_explore_permission(user):
         )
 
 
-def get_navigation_menu_items(user):
+def get_navigation_menu_items(user, request=None):
     # Get all pages that the user has direct add/edit/publish/lock permission on
     pages_with_direct_permission = get_pages_with_direct_explore_permission(user)
 
@@ -78,6 +79,11 @@ def get_navigation_menu_items(user):
         )
 
     pages = Page.objects.filter(criteria).order_by('path').specific()
+
+    if hooks.get_hooks('construct_explorer_page_queryset') and request:
+        parent_page = Page.objects.filter(path='/').first()
+        for hook in hooks.get_hooks('construct_explorer_page_queryset'):
+            pages = hook(parent_page, pages, request)
 
     # Turn this into a tree structure:
     #     tree_node = (page, children)
--- a/wagtail/wagtailadmin/templatetags/wagtailadmin_tags.py
+++ b/wagtail/wagtailadmin/templatetags/wagtailadmin_tags.py
@@ -34,7 +34,7 @@ else:
 @register.inclusion_tag('wagtailadmin/shared/explorer_nav.html', takes_context=True)
 def explorer_nav(context):
     return {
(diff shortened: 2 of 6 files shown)


gasman:
Hi @x-warrior - thanks for the PR!

To be honest, I've had second thoughts about the `construct_explorer_page_queryset` hook ever since I implemented it. It's too low-level - it gives developers the ability to make arbitrary changes to a queryset used in one specific place in the admin, with no guidance about the kind of changes that are "allowed" (i.e. they won't create inconsistencies with other areas of Wagtail, and won't break as a result of changes we make in future Wagtail releases).

The `get_navigation_menu_items` query in particular makes lots of unwritten assumptions, which a developer could easily break if they're not extra-careful: for example, the queryset must always be ordered by path, and a child page can never appear without its parent. For this reason, we definitely can't reuse the existing `construct_explorer_page_queryset` hook there (I'm certain that people are already using `construct_explorer_page_queryset` in ways that break these assumptions). Adding a new hook isn't much better, unfortunately:

* we'll still have to specify, very precisely, the constraints that the final queryset has to satisfy for the logic in `get_navigation_menu_items` to handle it correctly
* if we ever rewrite the menu logic (which is likely to happen some time soon: #3012) that specification will no longer be valid, and we'll have to come up with a new set of constraints
* There are an unknown number of other places in Wagtail's code that make assumptions about the behaviour of the page tree, which may not be true when the developer has complete freedom to customise it. Some of these places are not so obvious... for example, the button on the rightmost column of the explorer is normally a "navigate into this section" arrow when the page has children, and "add a page here" if not. If the page does have children, but they're hidden from the current user, then it will still appear as a "navigate" arrow. A user probably wouldn't notice this as a bug, but it's still inconsistent behaviour, and would logically need to be fixed with yet another hook.
* Similarly, any future Wagtail features involving page listings / querysets will also need to introduce new hooks; your custom code will have to keep up to date with all of these additions.

To be clear, this isn't a criticism of your PR - it's a fundamental problem with the concept of the `construct_explorer_page_queryset` hook. In short, I don't think it's feasible to make global customisations to the page tree's behaviour by intercepting every query that's performed on it. I think the real solution might be to introduce a formal concept of "hidden pages" within Wagtail, possibly as a new flag on the Page model. That way, we just have one new business rule with in Wagtail and can adjust our queries to accommodate it - which won't be a small task, but certainly more realistic than accommodating every possible queryset-level customisation.

x-warrior:
Thanks for the answer @gasman it seems a tricky problem.

I have been thinking about it for a while, I still think having a hook to filter pages is nice. I think the menu `get_navigation_menu_items` has a different behaviour from the other page listing but we could try to create it in a different way. For example, currently the menu is created all at once in one single requests and this is very different from the page explorer where you're doing a new request for every page. If we're able to change the menu to do one request for every menu step, it seems more aligned with page explorer. (#3012 seems to use an API)

Having this "similar" behaviour between menu and the other pages listings we could change the hook to be an extra filter `fitler_page_queryset`.

I do agree this solution doesn't solve the issue with 
> for example, the button on the rightmost column of the explorer is normally a "navigate into this section" arrow when the page has children, and "add a page here" if not.

but this issue already exists with the `construct_explorer_page_queryset`.

Maybe one way to mitigate it, is having a default behaviour for this button, it could always navigate into this section. You will not loose the "add a page" functionality because you can access it on the page actions.

gasman:
It's not solving these specific issues that I'm concerned about - I'm sure they're all fixable individually. I'm more concerned about the wider problem of maintainability: extending `construct_explorer_page_queryset`-style customisation everywhere means that every time we write a page query, now and in future, we'll have to consider how it interacts with every possible queryset modification that users might be applying through hooks (not just filtering the queryset, but reordering, or even inserting new items). That's simply not sustainable for future development of Wagtail.

gasman:
Will close this now, as unfortunately I don't think it'll be possible to resolve this in a way that ensures future maintainability of Wagtail core.

PR #3207, opened 2016-12-07
Fix for issue 3202 Updated dependencies on setup.py 

MattRijk:
#issue 3202
	
Problem:
	Sphinx-spelling is throwing an error when installing from setup.py
Fix: 
	Add pyenchant==1.6.6 before sphinxcontrib-spelling>=2.3.0 in install_requires [from documentation_extras]
	Run pyenhant before sphinxcontrib-spelling and before document requirement packages (in required installs). pyenhant must be set to 1.6.6 (pip on ubuntu wants to install 1.6.8. This throws and error with other packages install. Wont install [babel, imagesize, docutils, sphinxcontrib-spelling])
	Then update sphinxcontrib-spelling version -> changing to 2.3.0 works (was 2.1.1).
	Every thing should install and python runtests.py  - all tests ran with no E.
	
	Techincal details
	Wagtail: 1.7
	Python version: 2.7.6.
	Django version: Django==1.10.4
	Tested on Ubuntu 14.04.3 LTS, MacOS, Windows7

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,6 +32,8 @@ install_requires = [
     "Unidecode>=0.04.14",
     "Willow>=0.4,<0.5",
     "requests>=2.11.1,<3.0",
+    "pyenchant==1.6.6",
+    "sphinxcontrib-spelling>=2.3.0", 
 ]
 
 # Testing dependencies
@@ -56,8 +58,6 @@ documentation_extras = [
     'Sphinx>=1.3.1',
     'sphinx-autobuild>=0.5.2',
     'sphinx_rtd_theme>=0.1.8',
-    'sphinxcontrib-spelling==2.1.1',
-    'pyenchant==1.6.6',
 ]
 
 setup(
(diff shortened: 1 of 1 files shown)


gasman:
Thanks @MattRijk!

Are you sure that pyenchant and sphinxcontrib-spelling need to go in `install_requires`, rather than `documentation_extras`? This is something we'd want to avoid if possible, since it means that anyone who just wants to install Wagtail as an end-user (and has no intention of working on the Wagtail documentation) will end up installing a bunch of packages they don't need. Not the end of the world, of course, but I'd be keen to look for a solution that doesn't require this.

MattRijk:
Having it in doc extras works too. As long as its before Sphinx and the right package versions it should work. I'll change and resubmit

# Documentation dependencies
documentation_extras = [
    'pyenchant==1.6.6',
    'sphinxcontrib-spelling>=2.3.0',
    
    'Sphinx>=1.3.1',
    'sphinx-autobuild>=0.5.2',
    'sphinx_rtd_theme>=0.1.8',

]

gasman:
Thanks again! I can confirm that the new ordering works well for me too.

Merged in d5ec4fbb93ed83533c80644809cd91d71a2b64b0

PR #3211, opened 2016-12-09
Make all live and draft links open in a new win

marcofucci:
Some *live* and *view draft* links were opening in a new window and some weren't for no apparent reason. 
This fixes that inconsistency.

The links that were opening in the same window were in the:
- moderation section
- recent edits section
- messages

### DIFF
--- a/wagtail/wagtailadmin/messages.py
+++ b/wagtail/wagtailadmin/messages.py
@@ -31,5 +31,5 @@ def error(request, message, buttons=None):
     return messages.error(request, render(message, buttons))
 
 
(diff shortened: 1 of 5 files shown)


gasman:
Thanks @marcofucci! Merged in 2f06afe6e1ee1e2f25b9a1b9efe37e49fb830766.

(I had a feeling that we previously followed the rule "only open in a new window if the user is currently in an editing workflow that would get interrupted otherwise" - but I see we certainly weren't following that rule consistently, so happy to go with consistently opening in a new window.)

PR #3222, opened 2016-12-15
Allow usage of any field as detail key in API

lohmander:
This PR makes it so that you can set an alternative identifier for the detail view, like it usually works in django restframework (as far as I know).

So by settings the `lookup_field` and possibly the `lookup_value_regex` attribute on the endpoints, you can for instance query pages by their slug instead of the id.

Example:

```python
class MyPagesAPIEndpoint(PagesAPIEndpoint):
    lookup_field = 'slug'
    lookup_value_regex = '\w+'
```

Allows you to get a page like so (assuming the slug is "home"

`/api/v2/pages/home/`


kaedroho:
Thanks @lohmander!

> you can for instance query pages by their slug instead of the id.

This may be a problem as slugs are not unique across the site. If a user created pages at ``/foo/about`` and ``/bar/about`` it would be impossible to retieve either of them with the API as they both have the same slug.

I think the best way to query pages by slug is to use the listing view. For example:
```
GET /api/v2/pages/?slug=foo&fields=*

{
    "meta": {
        "total_count": 1
    },
    items: [
        {
            <page here>
        }
    ]
}
```

Wagtail currently lacks the ability to find a page in the API by it's public URL. It would be really great to see something that does this added into Wagtail. The idea I have at the moment is something like:

 - Add a view to the API, that takes a url path as a GET parameter (For example: ``/api/v2/pages/route/?url_path=/about/``)
 - When a client calls this view, they are either 302 redirected to the detail view or given a 404 response if it doesn't exist

I can't think of another use-case for this as there aren't any unique fields apart from the id. I'm going to close this PR but thanks for the contribution!

PR #3223, opened 2016-12-16
Change admin welcome to explorable root

jeffreyckchau:
Continuation of work towards multi-tenancy in wagtail.  

Adds a template tag for retrieving the name of the highest currently explorable ancestor. Additionally implements said tag in the admin homepage. 

ex.
My Site 
-----My First Page
-----My Next Page

User 1 is permissioned to only see My First Page. Now in the admin panel, instead of seeing "Welcome to the My Site Wagtail CMS", he sees "Welcome to the My First Page Wagtail CMS".

### DIFF
--- a/wagtail/wagtailadmin/templates/wagtailadmin/home.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/home.html
@@ -1,5 +1,5 @@
 {% extends "wagtailadmin/base.html" %}
-{% load gravatar staticfiles i18n %}
+{% load gravatar staticfiles wagtailadmin_tags i18n %}
 {% block titletag %}{% trans "Dashboard" %}{% endblock %}
 {% block bodyclass %}homepage{% endblock %}
 
@@ -18,7 +18,8 @@
                 </div>
             {% endif %}
             <div class="col9">
-                <h1>{% block branding_welcome %}{% blocktrans %}Welcome to the {{ site_name }} Wagtail CMS{% endblocktrans %}{% endblock %}</h1>
+                {% explorer_site user site_name as current_site %}
+                <h1>{% block branding_welcome %}{% blocktrans %}Welcome to the {{ current_site }} Wagtail CMS{% endblocktrans %}{% endblock %}</h1>
                 <h2>{{ user.get_full_name|default:user.get_username }}</h2>
             </div>
         </div>
--- a/wagtail/wagtailadmin/templates/wagtailadmin/home.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/home.html
@@ -18,8 +18,8 @@
                 </div>
             {% endif %}
             <div class="col9">
-                {% explorer_site user site_name as current_site %}
-                <h1>{% block branding_welcome %}{% blocktrans %}Welcome to the {{ current_site }} Wagtail CMS{% endblocktrans %}{% endblock %}</h1>
+                {% explorer_site user site_name as current_explorable_site %}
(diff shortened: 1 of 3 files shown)


gasman:
Thanks @jeffreyckchau!

I'm a little concerned that this is tying the admin UI too closely to one particular way of assigning permissions, which makes sense for a typical multi-tenant setup (each user having full permission over an individual site) but won't be suitable for other setups. Here we're implicitly assuming that the section where permissions are assigned is something that would be regarded as a 'site'... as a counterexample to that, consider a corporate site where most of the site content is in the control of a web team, but general staff are given edit access to the blog. In that situation, the general staff would see "Welcome to the Our Blog Wagtail CMS".

Perhaps an improvement would be to take the closest common ancestor of the user's permissions (as we do for the menu and breadcrumb), and call `get_site()` on it; if the resulting Site record exists and has a non-empty name, use that, otherwise fall back on the `WAGTAIL_SITE_NAME` setting.

(Generally I dislike the idea of Site records getting special treatment beyond URL generation and routing, as I think it's a good thing for the permission model and admin UI to behave the same regardless of whether your 'subsites' happen to be `subsite1.com` / `subsite2.com`, or `example.com/subsite1` / `example.com/subsite2`. However, I think it's justified in this case - I figure the main motivation for customising the welcome message here is to avoid revealing that the site is hosted on a shared Wagtail installation, and anyone choosing to use subpaths rather than distinct domains clearly isn't bothered about that...)

jeffreyckchau:
Hi @gasman,

Yes, you are correct. The primary motivation for this change is to avoid revealing to a user that their site may exist as a subpage of a larger Wagtail installation. I figure that if anybody has information about their site being part of a larger installation, they might start poking around and getting into places they shouldn't be. 

I agree with the change about querying the site record to find the relevant data. I hadn't previously considered that alternative model, so I'll go ahead and make that change now. 

To further continue with your example, would it be a good idea to include even more information about the Page/Site combination? For example, a general employee would see "Welcome to the Wagtail CMS for Our Blog on Corporate Site". I realize that however much information should be revealed is up to the site admins, so perhaps we should include the options as template tags and allowing the admins to implement them as needed on their own home.html?

gasman:
Hi @jeffreyckchau,

I can't really see the value of adding the root page title to the message - it's only meant to be a welcome message, not a status report on your permission level. However, if that's what developers want, they can always [override the branding_welcome block](http://docs.wagtail.io/en/v1.8/advanced_topics/customisation/admin_templates.html#branding-welcome). Since the message only appears on this one page, I think it would be neater to make the relevant items available as [variables in the template context](https://github.com/wagtail/wagtail/blob/0e9c255b034ce4bdef5b265021833790fd813ed9/wagtail/wagtailadmin/views/home.py#L85-L89), rather than using custom template tags - I guess the useful variables would be:

* `root_page` - the `Page` object for the explorable root, or None if the user has no page permissions
* `root_site` - the result of `root_page.get_site()` if defined, or None otherwise
* `site_name` - `root_site.site_name`, falling back on `root_site.hostname`, or `settings.WAGTAIL_SITE_NAME` if no `root_site` is defined

These would then be documented under http://docs.wagtail.io/en/v1.8/advanced_topics/customisation/admin_templates.html#branding-welcome.

Also, note that `get_pages_with_direct_explore_permission(user).first()` won't return the explorable root if the user has permission over multiple sections - the correct logic would be `get_pages_with_direct_explore_permission(user).first_common_ancestor(include_self=True, strict=True)`, as used by the `explorer_breadcrumb` tag. (This would probably be a good time to move that logic into a `get_explorable_root_page` function in wagtailadmin/navigation.py...)

jeffreyckchau:
Hi @gasman,

That's understandable. I'll go ahead and close this and create a new PR with your suggested changes so this one doesn't get too cluttered.

PR #3224, opened 2016-12-16
Remove wagtailimages.Filter as a model

### LINKED ISSUE #2881: Remove wagtailimages.Filter as a model (second try)

gasman:
Fixes #2881


gasman:
Swapping the RemoveField and DeleteModel operations in the migration should hopefully fix the immediate failures on MySQL; still need to investigate how to handle custom image models. (I suspect we need to tell developers to add a `run_before` constraint to their migration, to ensure that the filter FK in their custom rendition model is dropped before the filter table.)

gasman:
If we remove both the FK and the Filter model in Wagtail 1.9, then running makemigrations on a project with a custom image model fails with `ValueError: The field home.CustomRendition.filter was declared with a lazy reference to 'wagtailimages.filter', but app 'wagtailimages' doesn't provide model 'filter'.`

The following sequence of events seems to work (on sqlite at least):

* Wagtail 1.9 drops the `filter` FK from `AbstractRendition`
* Developer upgrades to Wagtail 1.9 and runs makemigrations / migrate
* Wagtail 1.10 drops the `Filter` model
* Developer upgrades to Wagtail 1.10 and runs migrate

The resulting migrations also complete successfully when bringing up the database from scratch... I was half expecting that to fail because there's nothing to tell the migration engine to run the "remove FK from CustomRendition" migration before dropping the Filter table. I guess I need to test that against mysql too...

PR #3231, opened 2016-12-19
replace self with cls for class method

msomierick:
This is a simple pep8 cleanup spotted using [lgtm](https://lgtm.com).

The first argument of a class method, a new method or any metaclass method should be called cls. This makes the purpose of the argument clear to other developers.


gasman:
Thanks! Merged in eb16a476baa35d74a562515cce473c43dba8b70a.

PR #1528, opened 2015-07-19
Implement deletion of form submission

hanpama:
- Write a view function to delete submission(views.delete_submission)
- Route the function to 'submissions/(page_id)/(submission_id)/delete/'
- Create a button for each row in submission list

Issue #778, #1190 suggested deletion of form submission object.
Can you review this please? I think it is pretty useful.

PR #3233, opened 2016-12-19
Bulk deletion of form submissions

kaedroho:
This pull request replaces the exsting deletion interface for form submissions (implemented in #1528) with one that allows deleting multiple submissions at the same time.

## Motivation

Some of our clients have a lot of form submissions and would like to delete them all to comply with data protection. It is very time consuming to have to do this one at a time. There is also a problem with the action buttons being on the right, which makes them difficult to get to for forms that have many fields.

## Solution

Each submission now has a checkbox to the left of it. When a checkbox is checked a delete button appears at the top. Clicking the delete button takes the user to the "confirm delete" view. It's possible to select multiple submisssions and delete them together.

It also has a gmail-like select/deselect all button at the very top to speed up mass-deletions.

## Screenshot

The "delete selected submissions" button only appears when one or more submissions have been selected

bulk-delete-submissions

### DIFF
--- a/wagtail/wagtailforms/templates/wagtailforms/index_submissions.html
+++ b/wagtail/wagtailforms/templates/wagtailforms/index_submissions.html
@@ -23,6 +23,67 @@
                 },
                 lang: 'lang'
             });
+
+            var selectAllCheckbox = document.getElementById('select-all');
+            var deleteButton = document.getElementById('delete-submissions');
+            var selectedSubmissions = {};
+
+            function updateActions() {
+                var everySubmissionSelected = Object.values(selectedSubmissions).every(function(v) { return v; });
+                var someSubmissionsSelected = Object.values(selectedSubmissions).some(function(v) { return v; })
+
+                // Select all box state
+                if (everySubmissionSelected) {
+                    // Every submission has been selected
+                    selectAllCheckbox.checked = true;
+                    selectAllCheckbox.indeterminate = false;
+                } else if (someSubmissionsSelected) {
+                    // At least one, but not all submissions have been selected
+                    selectAllCheckbox.checked = false;
+                    selectAllCheckbox.indeterminate = true;
+                } else {
+                    // No submissions have been selected
+                    selectAllCheckbox.checked = false;
+                    selectAllCheckbox.indeterminate = false;
+                }
+
+                // Delete button state
+                if (someSubmissionsSelected) {
+                    deleteButton.classList.remove('disabled')
+                    deleteButton.disabled = false;
+                } else {
+                    deleteButton.classList.add('disabled')
+                    deleteButton.disabled = true;
+                }
+            }
+
+            // Initialise selectedSubmissions
+            $('input[type=checkbox].select-submission').each(function() {
+                selectedSubmissions[this.dataset.submissionId] = false;
+            });
+
+
+            // Event handlers
+
+            $(selectAllCheckbox).on('change', function() {
+                // Select/deselect all
+                for (var submissionId in selectedSubmissions) {
+                    selectedSubmissions[submissionId] = this.checked;
+                }
+
+                // Update checkbox states in DOM
+                $('input[type=checkbox].select-submission').each(function() {
+                    this.checked = selectedSubmissions[this.dataset.submissionId];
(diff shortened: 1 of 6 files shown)


tomdyson:
LGTM

wagtail/wagtailforms/templates/wagtailforms/list_submissions.html
```diff
@@ -6,25 +6,29 @@
...
                 <th>{{ heading }}</th>
             {% endfor %}
-            <th>{% trans "Actions" %}</th>
         </tr>
     </thead>
     <tbody>
         {% for row in data_rows %}
             <tr>
+                <td>
+                    <input type="checkbox" name="select-{{ row.model_id }}" class="select-submission" data-submission-id="{{ row.model_id }}" />
```
gasman: The checkbox handling could be cleaned up a lot here by giving all checkboxes the same name and a `value` equal to the submission ID:

    <input type="checkbox" name="submission_id" value="{{ row.model_id }}">

This will then submit as something like `submission_id=1&submission_id=2&submission_id=3` and the list of IDs can be extracted with `request.GET.getlist('submission_id')`, avoiding the need for the string munging to remove the `select-` prefix.

wagtail/wagtailforms/templates/wagtailforms/index_submissions.html
```diff
@@ -23,11 +23,72 @@
...
+                selectedSubmissions[this.dataset.submissionId] = this.checked;
+
+                updateActions();
+            });
         });
     </script>
 {% endblock %}
 {% block content %}
-    <header class="nice-padding">
+    <header class="nice-padding" style="margin-bottom: 0px;">
```
gasman: Adding `margin-bottom: 0px` here breaks the styling of the "no submissions" message:

screen shot 2017-01-03 at 15 04 18

wagtail/wagtailforms/templates/wagtailforms/index_submissions.html
```diff
@@ -23,11 +23,72 @@
                 },
                 lang: 'lang'
             });
+
+            var selectAllCheckbox = document.getElementById('select-all');
+            var deleteButton = document.getElementById('delete-submissions');
+            var selectedSubmissions = {};
```
gasman: A slightly nitpicky issue: if you use the back button (which the confirm-delete page will tend to encourage, since there's no 'cancel' button there) the browser will restore the previous state of the checkboxes. These will then be out of sync with `selectedSubmissions` and the visibility of the 'delete' button.

I think you could probably eliminate `selectedSubmissions`, and use jQuery selectors directly to set `everySubmissionSelected` / `someSubmissionsSelected`:

    var submissionCheckboxes = $('input[type=checkbox].select-submission');
    var someSubmissionsSelected = submissionCheckboxes.is(':checked');
    var everySubmissionSelected = !submissionCheckboxes.is(':not(:checked)');

Then calling `updateActions` on page load should bring everything back in sync.

wagtail/wagtailforms/templates/wagtailforms/list_submissions.html
kaedroho: Didn't know about this, thanks!

wagtail/wagtailforms/templates/wagtailforms/index_submissions.html
kaedroho: Forgot I added this. Thanks for spotting! I'll just leave it the current way for now.

wagtail/wagtailforms/templates/wagtailforms/index_submissions.html
kaedroho: Yeah, this was a little bit over-engineered. I've simplified it as per your comment.

I couldn't reproduce a situation where the checkboxes would be out of sync when the user presses the back button. For me, the checkboxes are always reset to their unchecked state.

kaedroho:
@gasman 
Thanks for the review, this is now ready for another look!

gasman:
Merged in 4f1d3509ff0f8958adb288ec53945fde76c31260 - thanks!

PR #3237, opened 2016-12-20
Add new Block.get_api_representation method

marcofucci:
This PR adds a new method to the Block class called `render_api` so that the related API response data can be customised by simply overriding it.

For example, you might want to define a custom `ImageChooserBlock` which returns the whole image data instead of only its id:

```python
from wagtail.wagtailimages.blocks import \
    ImageChooserBlock as WagtailImageChooserBlock


class ImageChooserBlock(WagtailImageChooserBlock):
    def render_api(self, value, context=None):
        if value is None or context is None:
            return None

        router = context['router']
        endpoint = router.get_model_endpoint(value.__class__)[1]()
        serializer_class = endpoint._get_serializer_class(router, value.__class__, [], show_details=True)
        return serializer_class(value, context=context).data
```

I'm not sure about the name `render_api` but I tried to be consistent with the other methods.


wagtail/api/v2/serializers.py
```diff
@@ -225,7 +225,7 @@ class StreamField(Field):
     foreign objects are nested objects with id and meta as attributes.
     """
     def to_representation(self, value):
-        return value.stream_block.get_prep_value(value)
+        return value.stream_block.render_api(value, self.context)
```
kaedroho: Could we rename this to: ``get_api_representation()``

marcofucci:
@kaedroho I've renamed `render_api` to `get_api_representation`, please have a look and let me know if it's OK. 
I didn't rebase / squash / update the commit msg as I don't know if you want to keep this PR untouched to preserve history and do things manually when merging.

Issue #2907, opened 2016-08-10
Implementing custom permission behavior?


cjmochrie:
I am not sure if this is a bug or a feature request or simply a frustrated user, but I am struggling with developing custom permission behavior - namely: preventing the deletion of published pages. If there is a way to do this without altering Wagtail source I am all ears, but here is a brief overview of my solution:

``` python
# Create custom PagePermissionTester and UserPagePermissionsProxy classes
class CustomPagePermissionTester(PagePermissionTester):
    def can_delete(self):
        return not self.page.live and super().can_delete()

class CustomPermProxy(UserPagePermissionsProxy):
    def for_page(self, page):
        return CustomPagePermissionTester(self, page)
```

``` python
# ... within my page model
    def permissions_for_user(self, user):
        user_perms = CustomPermProxy(user)
        return user_perms.for_page(self)
```

Within `views/pages.py`

``` python
def delete(request, page_id):
    page = get_object_or_404(Page, id=page_id).specific # access specific page type
    if not page.permissions_for_user(request.user).can_delete():
        raise PermissionDenied
```

And to remove the option to delete a page from the dropdown - simply add `page_perms` to the render context of `def edit(request, page_id):` and remove the following lines from the edit template:

```
{% page_permissions page as page_perms %}
```

Would something like the above be possible to implement for Wagtail as a whole or are there other reasons why overriding `permissions_for_user` should not be allowed?

nickhudkins:
Hey @cjmochrie I just ran into this same issue, and noticed that the `{% page_permissions %}` template tag has this line in it: https://github.com/wagtail/wagtail/blob/de9ffaab4978ad5e50ead1c52b2c1036ee92fc0c/wagtail/wagtailadmin/templatetags/wagtailadmin_tags.py#L136

Unfortunately, those page permissions (although retrieved in the edit view: https://github.com/wagtail/wagtail/blob/de9ffaab4978ad5e50ead1c52b2c1036ee92fc0c/wagtail/wagtailadmin/views/pages.py#L301) go unused outside of the view. Which means that default Page permissions are used in context.

@gasman is this intended behavior or did we stumble on a bug?

The result of this is that if `permissions_for_user` is defined on a page sub class, the custom permissions returned are not respected.

nickhudkins:
[...]

PR #3238, opened 2016-12-20
Add config to allow change UserPagePermissionsProxy

x-warrior:
Add `WAGTAIL_USER_PAGE_PERMISSION_PROXY` config to let user change the PermissionProxy used. 

Thinking about a long-term, I'm imagining we will have use cases to change a lot of Page methods/properties, so being able to swap the `Page` object and override methods/properties seems simpler than creating tons of settings. But changing `Page` to be swappable will need more work.  

What do you guys think about this change? Any suggestions?

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -24,6 +24,7 @@ from django.template.response import TemplateResponse
 from django.utils import six, timezone
 from django.utils.encoding import python_2_unicode_compatible
 from django.utils.functional import cached_property
+from django.utils.module_loading import import_string
 from django.utils.six import StringIO
 from django.utils.six.moves.urllib.parse import urlparse
 from django.utils.text import capfirst, slugify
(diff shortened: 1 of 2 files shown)


x-warrior:
A few considerations:
- We could go with `get_name_model()` approach similar to the one used for Document but this is not a database model.
- Maybe `UserPagePermissionsProxy` should be renamed

gasman:
Thanks @x-warrior!

Please see #2907 for discussion of custom permission models in general. While I won't absolutely rule out supporting swappable permission policies at some point in the future, I don't think we're ready to do so *yet*. If we make `UserPagePermissionProxy` swappable, then it becomes a de-facto part of Wagtail's public API - and I don't consider `UserPagePermissionProxy` to be stable enough for that, since it's evolved incrementally to include permission checks as and when the admin interface requires them.

I think a candidate for a public API for page permissions would most likely look something like the PermissionPolicy framework we have for Wagtail's "peripheral" models (images, documents, sites): https://github.com/wagtail/wagtail/tree/master/wagtail/wagtailcore/permission_policies . There's probably a lot of work needed to get page permissions to that point, though.

Issue #2214, opened 2016-02-09
Separating publish delete actions and meta information on edit screen 


heymonkeyriot:
A number of times rather than click `publish` I've clicked `delete` on the edit screen. This is because I'm clumsy, but also because the two actions are next to one another in the toggleup menu, and have the same colouring / hierarchical weight.

Thinking about that problem led me to realise there are a few other possible issues with the action footer
- There's a potentially strange hierarchy (e.g the `preview` button is always accessible whilst `publish` isn't)
- There's an inconsistent use of colour (e.g. `preview` is grey, whilst the `save draft` is green)
- There's an inconsistent use of icons (only `preview` has one)
- Certain meta data (e.g. when it was last modified, and by whom) is on the same horizontal layer as the actions)
- Other meta data, e.g. status, privacy and edit lock isn't available on the footer
- There's friction in the user journey by the fact that most actions are hidden within the toggleup menu.

**Current footer**
![image](https://cloud.githubusercontent.com/assets/11335309/12911108/db028f66-cf07-11e5-8808-b3938b5ca72a.png)
# Possible amends

Below are some early ideas I was playing with, which
- Try and more clearly separate delete/ unpublish actions from save/ publish actions
- Try and separate out the meta data from actions (to enable more meta data to be included, and potentially increase extensibility)
- Try and increase consistency between button styles and icons
- Try and increase accessibility to all the actions (so we're not pre-judging which actions are most important to the user)

**Amended (white buttons on green)**
![image](https://cloud.githubusercontent.com/assets/11335309/12911194/8dd09b9c-cf08-11e5-9aad-67a9d08ac9c9.png)

**Amended (green buttons on black)**
![image](https://cloud.githubusercontent.com/assets/11335309/12911147/2e3d7920-cf08-11e5-9cac-180814a48f38.png)
_Closer to the existing UI_

**Amended (medium-esque UI)**
![image](https://cloud.githubusercontent.com/assets/11335309/12911159/3fa52744-cf08-11e5-9ec8-b5ea904a4007.png)
_Continuing with the idea of a toggle menu, with delete/unpublish in a different font-weight to publish. I don't think this is an improvement (I think it adds a different kind of complexity with the `...` menu, but here in case it sparks any better ideas)_

I recognise the above are quite opinionated suggestions so wanted to start a discussion around gathering thoughts before thinking about any sort of pull request.

davecranwell:
[...]

PR #3241, opened 2016-12-21
Example of better separated action buttons

strindhaug:
A very minimal change to address some of the ideas from https://github.com/wagtail/wagtail/issues/2214
as well as some of my personal annoyances with the current design.

Features:

- Separate publish button for faster and less annoying publishing.
- Color coded buttons for publish, unpublish and delete.
- Reordered dropdown elements based on how destructive each option is.
- A spacer between "positive" (save/submit) and "negative" (unpublish/delete) actions in the dropdown.
- Add icons for most options (somewhat arbitrarily chosen) for quick recognition without having to read.
 
This only adds one button so it doesn't increase the width of the button bar too much.

**Current footer:**
old footer

**New footer:**
new footer

**Tag this pull request with: "Work in progress" or "Request for comments" etc.**

### DIFF
--- a/wagtail/wagtailadmin/templates/wagtailadmin/pages/edit.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/pages/edit.html
@@ -40,27 +40,31 @@
                     <div class="dropdown dropup dropdown-button match-width {% if is_revision %}warning{% endif %}">
                         <button type="submit" class="button action-save button-longrunning {% if is_revision %}warning{% endif %}" tabindex="3" data-clicked-text="{% trans 'Saving...' %}" {% if page.locked %}disabled {% endif %}><span class="icon icon-spinner"></span><em>{% if page.locked %}{% trans 'Page locked' %}{% else %}{% if is_revision %}{% trans 'Replace current draft' %}{% else %}{% trans 'Save draft' %}{% endif %}{% endif %}</em></button>
 
-                        {% if not page.locked %}
+                        {% if not page.locked and not is_revision %}
                             <div class="dropdown-toggle icon icon-arrow-up"></div>
                             <ul role="menu">
-                                {% if not is_revision and page_perms.can_unpublish %}
-                                    <li><a href="{% url 'wagtailadmin_pages:unpublish' page.id %}">{% trans 'Unpublish' %}</a></li>
+                                {% if page_perms.can_delete %}
+                                    <li style="border-width: 0;"><a href="{% url 'wagtailadmin_pages:delete' page.id %}" class="shortcut button no">
+                                      <span class="icon icon-bin"></span>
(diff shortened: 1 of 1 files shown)


tomdyson:
Thanks very much for this, @strindhaug! Some initial responses, for general discussion:

- Why is the 'submit for moderation' area taller than the others?
- 'publish' seems to fit better with 'submit', 'unpublish', 'delete' than 'preview'. Should the 'publish' and 'preview' buttons be swapped around?
- I like the icons but I don't think we should use ticks for both 'submit' and 'publish'. How about an arrow pointing up and right for 'submit'?
- An annoyance with the existing button bar is the premature wrapping (see attached screenshot). Can we fix this at the same time?

screenshot 2016-12-21 12 11 59

strindhaug:
The apparently extra tall "submit for moderation" button is because it's a spacer above it, I removed the bottom border on the spacer, but I reallize it looks kind of strange...

My reasoning for the order of buttons is the typical workflow, save, preview then publish; but I agree it looks a bit strange...

The icons are rando, I didn't want to create new icons just to test...

The premature wrapping is caused by the preview button having a wide wrapper because it can be a dropdown.

strindhaug:
I've been playing around with different more radical changes. 

I'm thinking maybe a layout with 3 direct buttons (save draft, publish, preview) with icons that become icon only on smaller widths to avoid wrapping. 
And an ellipsis button that toggles a popup (something like "Amended (medium-esque UI)" in #2214) where all other actions is placed (including the revisons link).

Maybe the last changed date and author info could be moved up to the header. If it should be in the footer at all, it shoul probably be abbrevated using icons instead of "Last modified" etc. labels, and using avatar only (with full name as tooltip) since full names can be excessively long...

tomdyson:
> I've been playing around with different more radical changes.

Great, thanks. Please post any screenshots - it makes it much easier for non-developers to get involved in the discussion.

> I'm thinking maybe a layout with 3 direct buttons with icons (save draft, publish, preview), that become icon only on smaller widths to avoid wrapping.
> And an ellipsis button that toggles a popup (something like "Amended (medium-esque UI)" in #2214) where all other actions is placed (including the revisons link).

This sounds good to me. I like the 'medium-esque UI' approach in #2214.

> Maybe the last changed date and author info could be moved up to the header. If it should be in the footer at all, it shoul probably be abbrevated using icons instead of "Last modified" etc. labels, and using avatar only (with full name as tooltip) since full names can be excessively long...

Agreed on icons and avatars. I'm not sure about moving it to the header, which is already quite busy.

strindhaug:
I close this pr, this first test is very unpolished, and doesn't improve much.
Please see https://github.com/wagtail/wagtail/pull/3247 for a suggestion that makes a bigger difference...

PR #3245, opened 2016-12-21
View restrictions for collections

viaregio:
Based on the view restrictions for pages I implemented the view restriction for collections too. If someone only serves the documents and images with the serve view functions from wagtaildocs and wagtailimages they are fully protected. 
Some problems still exists because documents and images are normally saved in the media folder and so they can be accessible from the webserver.
I'm open for advice how to solve the remaining problems.


begonaalvarezd:
Is this going to be integrated in a future release?

gasman:
Rebased on latest master and cleaned up at https://github.com/gasman/wagtail/tree/viaregio-viewrestrictions

gasman:
Thanks @viaregio!

A few changes I propose to make here:

* Leave out the image serving logic - it doesn't really fit with the rest of the functionality, since users don't visit images like they visit pages or documents. This also simplifies some of the remaining code (e.g. rather than a `before_serve_from_collection` hook that gets caught within wagtailcore and has to support both images and documents, it can be a `before_serve_document` hook handled entirely within wagtaildocs).
* Don't make the Collection model responsible for serving the 'password required' response. This logic exists on pages so that we can customise the output for specific page types - e.g. show the intro paragraph followed by a "register to read the rest of the article" panel. There's no equivalent thing we can do for documents, and no way for users to subclass Collection, so we should skip straight to the `PASSWORD_REQUIRED_TEMPLATE` case.
* The `collections/edit.html` template should extend `generic/edit.html` (adding new overrideable blocks if necessary) - the collections admin interface was the first real use of the generic views/templates in wagtailadmin, and we hope to port more of Wagtail's admin views to use them in future, so it's important that the collection edit view follows best practices in re-using that template code.

viaregio:
The intention of this pull request is to allow for a setup with "non public" pages and collections. My understanding of a "non public" collection is a collection with documents and images that are only served to visitors after some kind of authorization.
If the image serving logic is omitted no images can be served from such a collection. To protect the users from accidentally serving images to users without some kind of authorization this would mean that such collections must not contain any images. I'm not very keen on this because public collections can contain documents and images.

gasman:
> If the image serving logic is omitted no images can be served from such a collection.

That's not really true - if the image serving logic is omitted, then all image renditions (regardless of collection) will have public URLs, so they'll continue to work exactly as they do now.

Unfortunately, image handling in Wagtail is a messy subject because there are lots of conflicting demands and opinions on how it should work:

* whether renditions should be generated in advance, or at the point of being used on a template, or at the point of being requested
* whether users should be able to request arbitrary renditions, or be locked down to the ones chosen by the developer
* whether to allow the "create a rendition from an image" step to be skipped (e.g. to support SVG images, or avoid recompression of images that were already optimised before uploading to Wagtail)
* whether it's best to host images and renditions locally or on remote storage such as S3
* how to deal with caching

Some of these problems have partial solutions, such as the dynamic image serve view - others are still waiting to be solved.

With this change, we're introducing the possibility of renditions being served via Django, which adds a whole new dimension of complexity. It's going to take a _lot_ of work to make sure that it works correctly with all of those other moving parts, and I really want to avoid shipping a "half-baked" solution where we have to tell users: "Access control on images doesn't work out of the box. You can enable it by setting up RenditionServeView... but that won't work if you're using remote file storage. And on a high traffic site you'll need to make sure sendfile is configured on your webserver, or else the extra Django threads will probably kill it. And we're not sure how to make it work with a frontend cache..."

There's currently a lot of demand for Wagtail to support access control on documents, so I think it's best that we ship a document-only solution as soon as possible, and tackle images separately as a future development.

viaregio:
> That's not really true - if the image serving logic is omitted, then all image renditions (regardless of collection) will have public URLs, so they'll continue to work exactly as they do now.

In my opinion this is exactly the problem that needs to be solved. As user of the protected collection I would assume that images uploaded to this collection are protected and have NO public URLs. To get images with public URLs I could upload them to non protected collections already.

If images can't be protected by uploading to protected collections they shouldn't be allowed in protected collections.

gasman:
@viaregio If this is a concern, we can update the message when setting up collection permissions to make it clear that they only apply to documents, not images.

I don't think we'll ever be able get RenditionServeView to a point where it's possible to enable it by default for everyone (including high-traffic sites using remote storage / CDNs), and so the potential for user error (i.e. content editors assuming that collection permissions will protect images, when really they don't) will still exist for sites that don't have RenditionServeView configured. I think it's safer for us to _not_ promise image protection as a feature, rather than provide an implementation that only works on certain configurations.

gasman:
Updated version now opened as PR #3644.

PR #3247, opened 2016-12-22
Improved footer menu with 3 shortcut buttons and one drop-up for all other actions

strindhaug:
Another example of a slightly more radical approach to fixing https://github.com/wagtail/wagtail/issues/2214
implementing some of the ideas I mentioned in my comments on the less radical https://github.com/wagtail/wagtail/pull/3241 pr.

This solution is a lot more compact and responsive even though it has an additional button.

Current footer, for width comparison (note that the preview button is actually wider than it appears):
orig

New footer in the same width (the preview button now doesn't have extra width since it's not a dropdown anymore; extra preview modes are shown in the common dropdown):
wide

Some of the meta-data is hidden (and revisions link is moved to the menu), to allow narrowers screen without wrapping:
narrow

The 3 direct buttons with icons (save draft, publish, preview) become icon only on smaller widths to avoid wrapping on phone sizes:
phone

Color coded (only on hover) delete and unpublish with red and yellow respectively:
color_code_on_hover

The "Submit for moderation" button  replaces the "Publish" button when the user is not allowed to publish:
cant_pub
It's deliberately not color coded green like the publish button to avoid confusion.

**NOTE** The css is currently hardcoded in a `<style>` block in the template for easy development and there is a lot of commented code for reference while developing. This needs to be cleaned up and the styles must be moved to the sass files before this can be merged.
But even though this code is somewhat ugly; it *is* functional and can be checked out locally for testing.

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_dropdowns.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_dropdowns.scss
@@ -417,6 +417,3 @@
 .is-open .u-toggle {
     display: block;
 }
-
-
-
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
@@ -430,11 +430,6 @@ footer,
         }
     }
 
-    footer {
-        width: 80%;
-        margin-left: 50px;
-    }
-
     .content {
         border-top: 0;
         background-color: none;
@@ -457,7 +452,7 @@ footer,
 
     footer {
         width: 80%;
-        margin-left: 50px;
+        margin-left: 1em;
     }
 
 
@@ -495,6 +490,7 @@ footer,
 
     footer {
         width: 90em;
+        margin-left: 50px;
     }
 }
 
--- a/wagtail/wagtailadmin/templates/wagtailadmin/pages/_preview_button_on_edit.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/pages/_preview_button_on_edit.html
@@ -1,4 +1,5 @@
 <button class="button action-preview {% if icon %}icon icon-view{% endif %}"
     data-action="{% url 'wagtailadmin_pages:preview_on_edit' page.id %}{% if mode %}?mode={{ mode|urlencode }}{% endif %}"
     data-placeholder="{% url 'wagtailadmin_pages:preview' %}"
-    data-windowname="wagtail_preview_{{ page.id }}">{{ label }}</button>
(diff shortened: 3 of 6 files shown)


strindhaug:
I changed my mind, the submit for moderation button is now green as well. I also added the mail icon as a temporary icon so that it can hide the very long button text on mobile.

Added lock icon and hiding of text when locked as well, to prevent wrapping.

tmsndrs:
I think this is an improvement on the previously suggested design. I worry that it feels a little cluttered though. 

- I think the preview should be separated out from the menu buttons though as it feels less related to editorial workflow. 
- The tick and green of the publish button could communicate that the page is published rather than will be published on click
- I'm not sure about the visual styling of the menu on desktop. It's feels dislocated from the button.
- Had the same thought about publish being replaced by 'submit for moderation' when appropriate. Think this is a good idea. 
- The hover colours don't feel like they're part of the Wagtail theme IMO.

strindhaug:
I've experiented some more with the pupup design (haven't committet yet), using a "speech bubble" style arrow to point down to the trigger button instead:

restyled_popup_menu

Though it's technically even more separated from the trigger, the arrow-ish design makes it feels more attached somehow.

I agree the preview button feels like it's misplaced, so I moved it to the popup menu saving even more space.

I've added a filled document icon as a temporary placeholder for a save symbol, and a fat chevron icon rotated 90 degrees with css transform as a temporary submit icon.
Would be nice if someone could create some good icons for these two actions and add it to the icon set. If they're sufficiently visually distinct the buttons don't need to be in different color.

The hover colors (not shown) on delete and unpublish is still there. They are actually unmodified colors from the styleguide. The buttons have the "button no" classes and "button warning" classes respectively. I've just overridden the un-hovered color to match normal buttons because they look a bitt garish next to each other. I don't mind removing them if there is strong opposition to using them in a popup; but I personally think they're a nice usability addition making the "destructive" actions more visible when hovered (I know there is a confirmation screen for these actions, but it's annoying to accidentally go to this screen when you were aiming for preview)

strindhaug:
Another way of making the buttons easier to differentiate without having to read them is spacing them a bit more when there is available space. 

Using css flex layout this could look like this:
Desktop:
flex_spaced_desktop
Mobile:
flex_spaced_mobile

Since it's wrapping flex box it doesn't require lots of space, while still uses all available space when the screen is wide enough. (And if flex isn't supported in the browser it falls back to normal left floated layout.)

tomdyson:
@strindhaug thank you very much for this thoughtful contribution and discussion. There's a lot for us to think about here, in combination with @tmsndrs's proposals at #3269. I would also like to include @teneightfive who has been doing some user testing on these elements as part of the https://github.com/nhsuk/nhsuk-content-store project.

tmsndrs:
Any objections to closing this? We're going to take this on from ticket #3269

thibaudcolas:
Based on #3269 I can see this is still at the design stage, so will close this WIP PR for now. Can always be re-opened depending on the direction things take 🙂.

PR #3252, opened 2016-12-23
Fix #3240 - Copying recursive tree branch to itself.

### LINKED ISSUE #3240: Coping a page to the same Tree will explode recursively
### Issue Summary

If you try to copy a Page which has childs and use the destination the same tree "branch" it will explode when trying to recursively copy. 

### Steps to Reproduce

1. Access Wagtaildemo
2. Copy Event listing adding the destination as itself
3. ```treebeard.exceptions.PathOverflow: The new node is too deep in the tree, try increasing the path.max_length property and UPDATE your database```

### Technical details

* Python version: Python 3.4.3
* Django version: Django (1.10.4)
* Wagtail version: 1.9a0.
* Browser version: http://whatsmybrowser.org/b/EJM4TK7

x-warrior:
Test were created on Page copy method, to avoid this error when someone uses the copy method directly, I also created the can_copy_to method which is used on the view. 

Refs: #3240

### DIFF
--- a/wagtail/wagtailadmin/tests/test_pages_views.py
+++ b/wagtail/wagtailadmin/tests/test_pages_views.py
@@ -2198,6 +2198,19 @@ class TestPageCopy(TestCase, WagtailTestUtils):
             "This slug is already in use within the context of its parent page \"Welcome to your new Wagtail site!\""
         )
 
+    def test_page_copy_post_and_subpages_to_same_tree_branch(self):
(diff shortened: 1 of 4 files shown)


gasman:
Thanks @x-warrior! Merged in 8376a62c8d6be89a6d7582000edaaaadcbd5551e - have also added this check to the form validation in 2b7f9eacbfc66a47b63c430112fc43d095e40951 so that users get a more friendly error message.

PR #3254, opened 2016-12-30
added default argument to ChoiceBlock.

nutztherookie:
If default is not None and required is True,
don't render a blank option.

This should cover #2903 .

### DIFF
--- a/docs/topics/streamfield.rst
+++ b/docs/topics/streamfield.rst
@@ -208,8 +208,15 @@ A dropdown select box for choosing from a list of choices. The following keyword
 ``choices``
   A list of choices, in any format accepted by Django's ``choices`` parameter for model fields (https://docs.djangoproject.com/en/stable/ref/models/fields/#field-choices), or a callable returning such a list.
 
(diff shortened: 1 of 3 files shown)


docs/topics/streamfield.rst
```diff
@@ -208,9 +208,16 @@ A dropdown select box for choosing from a list of choices. The following keyword
...
+``default`` (default: None)
+  The choice that will initially be selected.
+
 ``required`` (default: True)
   If true, the field cannot be left blank.
 
+.. note::
+
+  If `required` is True and `default` is not None, no blank choice will be rendered.
+
```
gasman: I don't think we particularly need a documentation update here - ``default`` is documented at the top of the 'basic block types' section, and there's no reason to draw attention to the fact that the blank choice is handled in a sensible way :-)

gasman:
Merged in 2734dfca8be2cb398fffd9f40d9fcf5a6b6df791 - thanks!

nutztherookie:
thank you :)

PR #3260, opened 2017-01-05
Doc tweak

jheasly:
Give an indication, i.e., ` ... `, in the StreamField templating examples that there may be other necessary template code needed for the StreamField template tag to work but that it's been removed for clarity.

I came to this page from "Getting Started" template examples where all the examples are rendered within {% block content %} tags, which aren't included in the StreamField template examples and therefore won't render when using the "Getting Started" template code. Spent a good while troubleshooting this; hope to save someone else the trouble.

### DIFF
--- a/docs/topics/streamfield.rst
+++ b/docs/topics/streamfield.rst
@@ -467,6 +467,8 @@ StreamField provides an HTML representation for the stream content as a whole, a
 
     {% load wagtailcore_tags %}
 
+     ...
+
     {% include_block page.body %}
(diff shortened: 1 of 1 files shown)


gasman:
Thanks! Merged in 9424a8eab9925f3b5266a53fcb62bbf173a2f589. (I left out the `...` in the last three examples as these are complete template files in their own right, to be pulled in using `{% include_block %}`, and won't contain any additional code.)

PR #3266, opened 2017-01-05
Marked text for translation

### LINKED ISSUE #3265: hardcoded strings in wagtailadmin and wagtaildocs
### Issue Summary

There are hardcoded strings in the folowing templates:

wagtailadmin/home/pages_for_moderation.html line 53
wagtailadmin/home/recent_edits.html line 37
wagtaildocs/documents/list.htmlline 40
wagtailadmin/pages/listing/_list.html 26 + 65

All with the frase “ago”

Cheers,

Rob

jjanssen:
Fixes #3265

There were some hardcoded strings left in some HTML templates.

### DIFF
--- a/wagtail/wagtailadmin/templates/wagtailadmin/home/pages_for_moderation.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/home/pages_for_moderation.html
@@ -50,7 +50,7 @@
                                 {{ revision.page.content_type.model_class.get_verbose_name }}
                             </td>
                             <td valign="top">
-                                <div class="human-readable-date" title="{{ revision.created_at|date:"d M Y H:i" }}">{{ revision.created_at|timesince }} {% trans 'ago' %} </div>
+                                <div class="human-readable-date" title="{{ revision.created_at|date:"d M Y H:i" }}">{% blocktrans with time_period=revision.created_at|timesince %}{{ time_period }} ago{% endblocktrans %} </div>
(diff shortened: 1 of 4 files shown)


gasman:
Hi @jjanssen,

I suspect that translating the word "ago" individually won't work for a lot of languages, due to differences in word order (and possibly inflections that modify the `timesince` string) - for example, in German, "3 days" => "3 Tage", "3 days ago" => "vor 3 Tagen".

I've just found that Django provides a [naturaltime](https://docs.djangoproject.com/en/1.10/ref/contrib/humanize/#naturaltime) tag, although that one [doesn't currently get inflections right either](https://code.djangoproject.com/ticket/21408) and has the side-effect of showing seconds too, which I'm not a fan of. So maybe we can get half way there with something like:

    {% blocktrans with time_period=revision.created_at|timesince %}{{ time_period }} ago{% endblocktrans %}

jjanssen:
Hi @gasman, thanks for the feedback! 

I was already thinking whether it should be `trans` or supplied as the example you mention with `blocktrans`. I only considered Dutch and English for this matter and those were the scenarios which came off ok.

Should we go for the half-way solution right now? Because I can adjust this in the afternoon.

gasman:
> Should we go for the half-way solution right now? Because I can adjust this in the afternoon.

Yes please - let's go for that.

gasman:
Merged in 0e9c255b034ce4bdef5b265021833790fd813ed9 - thanks!

PR #3270, opened 2017-01-06
Add site variables to template context

jeffreyckchau:
Continuation of #3223.

Add some data to the context of the home renderer so developers are able to alter the admin homepage based on the site record and level of permission on the given page. 

Also added a convenience function in wagtailadmin/navigation.py to retrieve the highest explorable page for a given user.

### DIFF
--- a/wagtail/wagtailadmin/views/home.py
+++ b/wagtail/wagtailadmin/views/home.py
@@ -89,8 +89,9 @@ def home(request):
     else:
         root_site = None
 
+    real_site_name = None
     if root_site:
-        real_site_name = root_site.site_name else root_site.hostname
+        real_site_name = root_site.site_name if root_site.site_name else root_site.hostname
 
     return render(request, "wagtailadmin/home.html", {
         'root_page': root_page,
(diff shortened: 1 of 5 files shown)


wagtail/wagtailadmin/views/home.py
```diff
@@ -82,8 +83,16 @@ def home(request):
...
+    if root_page:
+        root_site = root_page.get_site()
+    else:
+        root_site = None
+
     return render(request, "wagtailadmin/home.html", {
-        'site_name': settings.WAGTAIL_SITE_NAME,
+        'root_page': root_page,
+        'root_site': root_site,
+        'site_name': root_site.site_name if root_site is not None else settings.WAGTAIL_SITE_NAME,
```
gasman: The `site_name` field of Site is optional (and blank in the default install), so we should check for this being blank and fall back on `root_site.hostname` if so (or possibly fall back directly to `WAGTAIL_SITE_NAME` - but I think the hostname makes more sense).

gasman:
Thanks! Merged in 47f7d79aec676e3388fdc3aa83c7d5a1de6ddc45.

PR #3273, opened 2017-01-12
Provide fallback value for schema_only parameter in update_index management command

StriveForBest:
Currently if management command is called from inside the code, it breaks.
`**options` kwargs dictionary doesn't necessarily have `schema_only` param and your code doesn't check if it does.

Example (this will break):
```
update_index.Command().handle(backend_name='default')
```
`update_backend` function does have a default value for `schema_only` but `handle` does not.
You do make a similar check for existence of `options['backend_name']`, I believe something similar should be done for `options['schema_only']`


gasman:
Thanks! Merged in 0fdf98469c16dec9872f6f9bbd0488a43ee0ee50.

PR #3275, opened 2017-01-13
Pass parent context to get context as discussed in #2786

Nossnevs:
Added parent_context to get_context according to discussion in https://github.com/wagtail/wagtail/pull/3032 , https://github.com/wagtail/wagtail/pull/2786 and https://github.com/ipmb/wagtail/pull/1

Legacy handling is added so whenever a block with a get_context without parent_context is rendered it throws a deprecation warning 

Tests is added to test the new functionalities.

### DIFF
--- a/wagtail/wagtailcore/tests/test_blocks.py
+++ b/wagtail/wagtailcore/tests/test_blocks.py
@@ -82,7 +82,7 @@ class TestFieldBlock(unittest.TestCase):
 
         self.assertEqual(len(ws), 1)
         self.assertIs(ws[0].category, RemovedInWagtail111Warning)
-        self.assertEqual(html, '<h1>Bonjour le monde!</h1>')
+        self.assertEqual(html, '<h1 lang="fr">Bonjour le monde!</h1>')
 
     def test_charfield_render_form(self):
         block = blocks.CharBlock()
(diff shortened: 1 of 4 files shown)


wagtail/wagtailcore/tests/test_blocks.py
```diff
@@ -51,12 +64,25 @@ def test_charfield_render_with_template(self):
...
+        with warnings.catch_warnings(record=True) as ws:
+            warnings.simplefilter('always')
+
+            html = block.render("Bonjour le monde!", context={
+                'language': 'fr',
+            })
+
+        self.assertEqual(len(ws), 1)
+        self.assertIs(ws[0].category, RemovedInWagtail111Warning)
+        self.assertEqual(html, '<h1>Bonjour le monde!</h1>')
```
gasman: This should be `'<h1 lang="fr">Bonjour le monde!</h1>'`, I think...? In the case that `get_context` does not accept a `parent_context` kwarg, `render` should merge the result of `get_context(value)` into its original `context`, so that the previous behaviour is preserved.

Nossnevs:
I have added the code @gasman suggested and updated tests

gasman:
Looks good, thanks! Merged in 827149f5d21620f8bbbc9b6517e7cc686c52b487 + 19310a84ed4f9eaa27d2822181350310c7e2d6da.

PR #3285, opened 2017-01-20
Display draft's title in the admin UI instead of published title

m1kola:
TODO:

- [x] Add the `draft_title` field to the `Page` model
- [x] Write a migration to set the initial value for existing pages from the title field
- [x] Update this field every time a revision is saved
- [x] Update the admin interface to use it everywhere
- [x] Update existing tests
- [x] Add tests
- [x] Add docs


m1kola:
Hi @kaedroho. Could you, please, review this PR?

kaedroho:
cc @gasman

jjanssen:
Related issue: #529

kaedroho:
@gasman do you have any objects to this approach?

gasman:
@kaedroho No - I'm very happy with this approach.

gasman:
I think we need to update `Page.full_clean` to populate `draft_title` from `title` when it's left empty,
 [as we do for slug](https://github.com/wagtail/wagtail/blob/4caa8601494a458348060e5f7cc988afcb960937/wagtail/wagtailcore/models.py#L438). This ensures that any existing code which creates pages programmatically by doing:

    new_page = SomePage(title="hello world")
    parent_page.add_child(instance=new_page)

doesn't end up with a blank title in the admin. This just caught me out when adding tests for `get_admin_display_title` while working on #3689...

Any objections?

(Side note: there's an open issue #3511 which suggests that full_clean probably isn't the right place to do these model-level fixups - but best to tackle that separately)

PR #3287, opened 2017-01-22
Minor style tweaks for modeladmin's IndexView

ababic:
This is a part recreation of #3074, just focussing on getting a few styles fixed and improving consistency with other parts of the CMS. I'm hoping this will help with review/approval. I'll work on reimplementing the other changes from #3074 as a separate PR later, once all the outstanding modeladmin PRs are merged and over with.

Here are the aims of this particular PR:

- Removes unnecessary right padding from result list on desktop, when there are no filters to display, and the listing should be full width.
- Removes unnecessary left/right padding from result list on mobile (the table cells have padding of their own)
- In order to be more consistent with the Explorer list view, only make orderable column header links 'teal' if the listing is currently ordered by that field
- Better use of hierarchy in scss to improve readability

### DIFF
--- a/wagtail/contrib/modeladmin/static_src/wagtailmodeladmin/scss/index.scss
+++ b/wagtail/contrib/modeladmin/static_src/wagtailmodeladmin/scss/index.scss
@@ -1,28 +1,41 @@
+@import 'wagtailadmin/scss/variables';
+
 .content header {
     margin-bottom: 0;
 }
 
 
 .result-list {
-    padding: 0 15px;
+    margin-bottom: 0;
+}
 
-    table {
-        margin-bottom: 0;
-    }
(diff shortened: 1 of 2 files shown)


kaedroho:
Thanks @ababic!

Could you add a couple of screenshots showing what the index page looks like before/after this change?

ababic:
Hey @kaedroho. Sorry, I do need to get into the habit of doing that for front-end changes. Unfortunately though, I don't have my machine in a state where I can easily do that now, and there may be some wait before I do. Is the description not enough to go on here?

ababic:
Thank you @jjanssen :)

jjanssen:
Merged in 0c05655

PR #3288, opened 2017-01-23
Bug fix: Correct positional argument order for a couple of ModelAdmin methods

ababic:
The `get_class_names_for_field_col()` and `get_extra_attrs_for_field_col()` methods are both provided with the `obj` and `field_name` positional arguments in the wrong order from `items_for_result()` in the template tags. The order should be: `obj`, `field_name` (as used on the ModelAdmin method itself, and as shown in the method explanation / accompanying code example in the docs).

This PR fixes that issue, and adds some more customisation examples to the test app to show these and a couple of other methods working properly.


gasman:
Thanks @ababic! Merged in e771fefa5f540d8664826b39b0a0949965184985.

Issue #3256, opened 2017-01-04
Rendition constraint exceeds maximum MySQL index key length


chosak:
### Issue Summary

The change to image rendition model constraints [in the 1.8 release](http://docs.wagtail.io/en/v1.8/releases/1.8.html#unique-together-constraint-on-custom-image-rendition-models-needs-updating) causes problems with MySQL index key lengths. Specifically, adding this constraint:

```py
unique_together = (
    ('image', 'filter_spec', 'focal_point_key'),
)
```

creates this MySQL constraint command:

```sql
ALTER TABLE `renditionstable` ADD CONSTRAINT `indexname_uniq` UNIQUE (`image_id`, `filter_spec`, `focal_point_key`);
```

on a table like this:

```sql
mysql> DESCRIBE renditionstable;
+-----------------+--------------+------+-----+---------+----------------+
| Field           | Type         | Null | Key | Default | Extra          |
+-----------------+--------------+------+-----+---------+----------------+
| image_id        | int(11)      | NO   | MUL | NULL    |                |
| filter_spec     | varchar(255) | NO   | MUL | NULL    |                |
| focal_point_key | varchar(255) | NO   |     | NULL    |                |
| ...             |              |      |     |         |                |
+-----------------+--------------+------+-----+---------+----------------+
```

This fails on both a new database and an existing one with

```sql
ERROR 1071 (42000): Specified key was too long; max key length is 1000 bytes
```

This constraint is trying to create a `UNIQUE` index on three columns, two of which are type `varchar(255)`. When using the `utf8` encoding, these columns [use](http://dev.mysql.com/doc/refman/5.7/en/charset-unicode-utf8.html) 3 bytes per character (for `uf8mb4`, it's 4 bytes per).

When [creating an index](http://dev.mysql.com/doc/refman/5.7/en/create-index.html) in MySQL, "key values are formed by concatenating the values of the given columns." This means that the index width would have to be (11) + (255 * 3) + (255 *3), which exceeds the maximum key length of 1000 bytes (the limit for the MyISAM storage engine). In Wagtail 1.7 and earlier because the constraint was on `filter_id` instead, the index size would have been (11) + (11) + (255 *3), which fit okay under the limit.

[...]

Issue #2953, opened 2016-08-26
MySQL: Specified key was too long


elwinbuisman:
Applying wagtailcore.0001_squashed_0016_change_page_url_path_to_text_field...Traceback (most recent call last):

  File "./manage.py", line 17, in <module>

    execute_from_command_line(sys.argv)

  File "/Users/user/virtualenvs/project/lib/python2.7/site-packages/django/core/management/__init__.py", line 367, in execute_from_command_line

    utility.execute()

  File "/Users/user/virtualenvs/project/lib/python2.7/site-packages/django/core/management/__init__.py", line 359, in execute

    self.fetch_command(subcommand).run_from_argv(self.argv)

  File "/Users/user/virtualenvs/project/lib/python2.7/site-packages/django/core/management/base.py", line 305, in run_from_argv

    self.execute(*args, **cmd_options)

  File "/Users/user/virtualenvs/project/lib/python2.7/site-packages/django/core/management/base.py", line 356, in execute

    output = self.handle(*args, **options)

  File "/Users/user/virtualenvs/project/lib/python2.7/site-packages/django/core/management/commands/migrate.py", line 202, in handle

    targets, plan, fake=fake, fake_initial=fake_initial

  File "/Users/user/virtualenvs/project/lib/python2.7/site-packages/django/db/migrations/executor.py", line 97, in migrate

    state = self._migrate_all_forwards(plan, full_plan, fake=fake, fake_initial=fake_initial)

  File "/Users/user/virtualenvs/project/lib/python2.7/site-packages/django/db/migrations/executor.py", line 132, in _migrate_all_forwards

    state = self.apply_migration(state, migration, fake=fake, fake_initial=fake_initial)

  File "/Users/user/virtualenvs/project/lib/python2.7/site-packages/django/db/migrations/executor.py", line 237, in apply_migration

    state = migration.apply(state, schema_editor)

  File "/Users/user/virtualenvs/project/lib/python2.7/site-packages/django/db/migrations/migration.py", line 129, in apply

    operation.database_forwards(self.app_label, schema_editor, old_state, project_state)

  File "/Users/user/virtualenvs/project/lib/python2.7/site-packages/django/db/migrations/operations/models.py", line 96, in database_forwards

    schema_editor.create_model(model)

  File "/Users/user/virtualenvs/project/lib/python2.7/site-packages/django/db/backends/base/schema.py", line 295, in create_model

    self.execute(sql, params or None)

[...]

Issue #2925, opened 2016-08-18
Support for MySQL utf8mb4 charset


dwasyl:
Just trying to build a new wagtail site on MySQL 5.5 and ran into a few errors on the initial migrations. If the MySQL database is setup for utf8mb4 (MySQL speak for a proper UTF-8 charset) the max length it can handle for a char/varchar index is 191 (since the maximum column size is 767 bytes).

I ran into the error on a few of the initial migrations with Wagtail 1.5.1:
- wagtailimages.0001_initial
- wagtailcore.0001_squashed_0016
- wagtailimages.0004_make_focal_point_key_not_nullable
- wagtailimages.0005_make_filter_spec_unique
- wagtailcore.0021_capitalizeverbose
- wagtailcore.0024_collection
- wagtailembeds.0001_initial (having URLFields)
- wagtailredirects.0001_initial
- wagtailredirects.0004_set_unique_on_path_and_site
- wagtailredirects.0005_capitalizeverbose
- wagtailsearch.0001_initial

Wordpress last year started converting all of their hosted sites to utf8mb4 (see [here](https://make.wordpress.org/core/2015/04/02/the-utf8mb4-upgrade/)), and it's been an outstanding Django issue for quite a while ([#18932](https://code.djangoproject.com/ticket/18392))

It may not be possible to fix past migration files, but a note in the docs regarding utf8mb4 might be a good idea, and at least consideration to making sure future migrations either limit char/varchar to 191 characters would be helpful. One of the solutions for Django proper is to add a field to let the user specific the max length globally (either 191 or 255). This seems to have been employed by django-allauth and a few other apps as a workaround to this problem.

gasman:
Thanks for the report! I'm in favour of the idea of adding a setting variable to specify the max length.

I don't think there's anything in Wagtail that specifically depends on a CharField larger than 191 characters - normally, in the cases where people have run into size limits (image filenames, Page.url_path...) we've responded by converting the field into a TextField instead.

dwasyl:
Quite welcome - that really seems like the simplest solution for this problem. There really aren't very many fields that would be impacted, and the only people needing the shorter length wouldn't be losing data (since they can't have more stored anyway). Having modified all of the fields in question (and previous migrations), there really doesn't seem to be anything dependent on it. Mostly it's just either key storage or file paths.

MySQL always adds to the fun.

Thanks for the quick reply!

thibaudcolas:
[...]

PR #3293, opened 2017-01-23
Fix overlong index on wagtailimages.Rendition on MySQL (for 1.8.x branch)

gasman:
Retrospectively reduce focal_point_key's max length from 255 to 16. Ref: #3256 and partially #2925 / #2953 

This prevents migration wagtailimages/0016 from failing on MySQL configurations that could handle the previous unique_together index ('image', 'filter', 'focal_point_key') but not the new one ('image', 'filter_spec', 'focal_point_key'). It also hopefully means one fewer failure case when using utf8mb4 encoding (which currently can't get past the initial migrations).

This PR targets the 1.8.x branch - I'll create a separate one for master / 1.9 (which also needs to renumber the existing migration wagtailimages/0017).

### DIFF
--- a/wagtail/__init__.py
+++ b/wagtail/__init__.py
@@ -4,7 +4,7 @@ from wagtail.utils.version import get_semver_version, get_version
 
 # major.minor.patch.release.number
 # release must be one of alpha, beta, rc, or final
-VERSION = (1, 8, 0, 'final', 0)
+VERSION = (1, 8, 1, 'alpha', 0)
 
 __version__ = get_version(VERSION)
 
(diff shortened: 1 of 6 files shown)


chosak:
This fails when upgrading an existing database from 1.7 to 1.8.1 because of [this](https://github.com/wagtail/wagtail/pull/3293/files#diff-f2d89cc4d6a10ffd78513d82ceb26a8dR66) change.

Migration wagtailimages.0004 was already applied on a previous version. Changing that migration after the fact breaks how Django migrations code tries to determine what DB changes wagtailimages.0016 needs to make. When 0016 is being applied, Django tries to optimize out any unnecessary steps. It looks at that changed line and thinks that the field is already set with a `max_length=16`, so skips [the new `AlterField` step](https://github.com/wagtail/wagtail/pull/3293/files#diff-e3c23d3a57a50574520c36822f8b587fR42).

Skipping the database change (which Django thinks, wrongly, has already happened) means that the `AlterUniqueTogether` step continues to raise the "key too long" error.

chosak:
This works for me now, thanks!

gasman:
Merged in c5c8afc196bd9c7f91c129574752c4bc5b396bb6

PR #3294, opened 2017-01-23
Fix overlong index on wagtailimages.Rendition on MySQL (for 1.9.x branch)

gasman:
See #3293 - rebased here for the master branch + existing migration 0017 renumbered


gasman:
Merged in 84efa8888325592adedaed91da84e67f86b09615

Issue #2759, opened 2016-06-20
Support panel definitions, 'fields' and 'exclude' on the ModelAdmin class for non-Page models


ababic:
This would be reasonably simple to implement into the `modeladmin` app, but it wouldn't be possible right now to pick those panel definitions up in other views, such as for use in `InlinePanel`s etc. 

Maybe this is a discussion for a separate thread,  but maybe if we moved toward a django-like approach, where there'd be a central registry of models and their ModelAdmin classes, so configuration like that could be picked up?

I think it would also be sensible to support `fields` and `exclude` attributes (maybe using different names?) on the ModelAdmin class. I know our aim isn't to fully support every feature of Django ModelAdmin, and the line must be drawn somewhere, but to save developers from having to import and a bunch of edit handlers and write full definitions, just change field order or exclude a couple of fields, I think would be a nice feature. 

Any thoughts?

theju:
Sounds like a good idea.

Some of the basic cases that for customizing the modeladmin that I can think of are:
* Show/hide/reorder fields (described in your request above)
* Add js/css for widgets (already doable through the (inspect|form)_view_js/css attributes)
* Add custom validation (handled in #2717 through custom form)

ababic:
Thanks for the input @theju.

In regards to custom validation, we at RKH have found that overriding the `clean` and `save` methods on the model itself is often the most reliable way to achieve such things when working with Wagtail (it works for page-type models, too!). I suppose it depends on what you're trying to achieve, though.

gasman:
'exclude' behaviour is implemented (as the `form_fields_exclude` attribute) as of #3295; leaving this issue open for the remaining proposals ('fields' / 'include_fields', and specifying a panel definition on the ModelAdmin class).

DzMkh:
@gasman Hi,
Is it possible somehow to exclude fields from form for Page-models depending on some conditions?

Right now I have only 1 solution:
```
class NewFieldPanel(FieldPanel):
    def bind_to_instance(self, instance=None, form=None, request=None):
        form.fields['managers'].widget = HiddenInput()
        new = super().bind_to_instance(
            instance=instance, form=form, request=request
        )
        return new
```

Bu it only hides input instead of removing field panel at all. I need somehow override **exclude** in **Meta**.

laymonage:
Support for `panels` and `edit_handler` were added in #4900, so only `fields`/`include_fields` is remaining.

[...]

PR #3295, opened 2017-01-24
Add exclude_fields to ModelAdmin and ModelFormView, tests and doc

x-warrior:
Refers: #2759 

Do you want to use `exclude` for the field name directly instead? I have used `exclude_fields` to be more explicit because there are a lot of different ModelAdmin properties.

### DIFF
--- a/docs/reference/contrib/modeladmin/create_edit_delete_views.rst
+++ b/docs/reference/contrib/modeladmin/create_edit_delete_views.rst
@@ -150,3 +150,12 @@ See the following part of the docs to find out more:
 
 See the following part of the docs to find out more:
 :ref:`modeladmin_overriding_views`
+
+.. _modeladmin_exclude_fields:
+
+-----------------------------------
+``ModelAdmin.exclude_fields``
+-----------------------------------
+
+**Expected value**: An array of Model fields to be excluded from Create/Edit pages.
+
--- a/docs/reference/contrib/modeladmin/create_edit_delete_views.rst
+++ b/docs/reference/contrib/modeladmin/create_edit_delete_views.rst
@@ -151,11 +151,16 @@ See the following part of the docs to find out more:
 See the following part of the docs to find out more:
 :ref:`modeladmin_overriding_views`
 
-.. _modeladmin_exclude_fields:
+.. _modeladmin_form_fields_exclude:
 
 -----------------------------------
-``ModelAdmin.exclude_fields``
+``ModelAdmin.form_fields_exclude``
 -----------------------------------
 
-**Expected value**: An array of Model fields to be excluded from Create/Edit pages.
+**Expected value**: A list or tuple of fields names
 
+When using CreateView or EditView to create or update model instances, this
+value will be passed to the edit form, so that any named fields will be
+excluded from the form. This is particularly useful when registering ModelAdmin
+classes for models from third-party apps, where defining panel configurations
+on the Model itself is more complicated.
--- a/wagtail/contrib/modeladmin/options.py
+++ b/wagtail/contrib/modeladmin/options.py
@@ -142,6 +142,7 @@ class ModelAdmin(WagtailRegisterable):
     inspect_view_extra_js = []
     form_view_extra_css = []
     form_view_extra_js = []
+    exclude_fields = []
 
     def __init__(self, parent=None):
         """
--- a/wagtail/contrib/modeladmin/options.py
+++ b/wagtail/contrib/modeladmin/options.py
@@ -142,7 +142,7 @@ class ModelAdmin(WagtailRegisterable):
     inspect_view_extra_js = []
     form_view_extra_css = []
     form_view_extra_js = []
-    exclude_fields = []
+    form_fields_exclude = []
 
     def __init__(self, parent=None):
         """
@@ -159,6 +159,11 @@ class ModelAdmin(WagtailRegisterable):
             self.model, self.inspect_view_enabled)
         self.url_helper = self.get_url_helper_class()(self.model)
 
+    def get_form_fields_exclude(self, request):
+        """
+        Returns a list or tuple of fields names to be excluded from Create/Edit pages.
+        """
+        return self.form_fields_exclude
 
     def get_permission_helper_class(self):
         """
--- a/wagtail/contrib/modeladmin/options.py
+++ b/wagtail/contrib/modeladmin/options.py
@@ -159,12 +159,6 @@ class ModelAdmin(WagtailRegisterable):
             self.model, self.inspect_view_enabled)
         self.url_helper = self.get_url_helper_class()(self.model)
 
-    def get_form_fields_exclude(self, request):
-        """
-        Returns a list or tuple of fields names to be excluded from Create/Edit pages.
-        """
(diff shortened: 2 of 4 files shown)


ababic:
Am happy to review this. It just won't be for another few days (due to work commitments).

docs/reference/contrib/modeladmin/create_edit_delete_views.rst
```diff
@@ -150,3 +150,12 @@ See the following part of the docs to find out more:
...
 See the following part of the docs to find out more:
 :ref:`modeladmin_overriding_views`
+
+.. _modeladmin_exclude_fields:
+
+-----------------------------------
+``ModelAdmin.exclude_fields``
+-----------------------------------
+
+**Expected value**: An array of Model fields to be excluded from Create/Edit pages.
```
ababic: I would describe the expected value as a list of 'field names' rather than Model fields. I know it's very unlikely that anyone would try to provide an actual list of Model field, but it can't hurt to be clear. Also, I think it would be better to follow the other examples in the docs and explain the **Expected value** in one paragrah, then how and where the value is used in a separate paragraph below.

wagtail/contrib/modeladmin/options.py
```diff
@@ -142,6 +142,7 @@ class ModelAdmin(WagtailRegisterable):
     inspect_view_extra_js = []
     form_view_extra_css = []
     form_view_extra_js = []
+    exclude_fields = []
```
ababic: On the ModelAdmin class, we already have similar fields for InspectView (`inspect_view_fields` and `inspect_view_fields_exclude`). I'm wondering if `form_fields_exclude` might be more consistent?

wagtail/contrib/modeladmin/views.py
```diff
@@ -115,7 +115,7 @@ def get_edit_handler_class(self):
         if hasattr(self.model, 'edit_handler'):
             edit_handler = self.model.edit_handler
         else:
-            panels = extract_panel_definitions_from_model_class(self.model)
+            panels = extract_panel_definitions_from_model_class(self.model, exclude=self.model_admin.exclude_fields)
```
ababic: The general pattern I've tried to follow with ModelAdmin is to accompany each attribute with a `get()` method, which just returns the attribute value by default. It would help developers when their needs go beyond defining a simple list. Do you think we might be able to do something similar here? Also, since the method would always be called with access to a request object, it might be handy for the method to accept `request` as a positional argument (which would give developers scope to exclude different fields depending on the user).

docs/reference/contrib/modeladmin/create_edit_delete_views.rst
x-warrior: ```**Expected value**: A list or tuple of fields names to be excluded from Create/Edit pages.```

I'm not sure about how to write the other paragraph, the ones I have checked were like this:
```See the following part of the docs to find out more: Overriding templates``` 

Any suggestion ?

ababic:
Generally speaking, I'm fine with the idea of this being functionality being merged on it's own, but I think it would feel more complete with an `include_fields` equivalent. That would be a considerable update, though. I suppose there's no reason it couldn't be added later, though. @gasman. Do you have any thoughts on this?

docs/reference/contrib/modeladmin/create_edit_delete_views.rst
ababic: The examples in the [IndexView docs](https://github.com/springload/wagtail/blob/491b643ed95d8ec80064c018133cc993e78e0bac/docs/reference/contrib/modeladmin/indexview.rst) are probably a better guide for one. I would stop at explaining what the value is in the first line e.g.

`A list or tuple of field names for your model`

 then the have something like this for the description:

`When using CreateView or EditView to create or update model instances, this value will be passed to the edit form, so that any named fields will be excluded from the form. This is particularly handy if you're registering ModelAdmin classes for models from third-party apps, where defining panel configurations on the Models is less straight forward.`

ababic:
I'm completely unfamiliar with `mock` I'm afraid, so struggling to review the tests (sorry!). Is anyone else able to give those a scan?

gasman:
> Generally speaking, I'm fine with the idea of this functionality being merged on it's own, but I think it would feel more complete with an include_fields equivalent. That would be a considerable update, though. I suppose there's no reason it couldn't be added later. @gasman, do you have any thoughts on this?

As a general rule - if a PR is functionally complete at what it sets out to do, then I wouldn't hold off on accepting it just because there's something else that _could_ logically be added to it. I've had that situation a few times now, and it's easy to fall into a cycle of "just one more thing..." and end up delaying things unnecessarily. If someone feels bothered by the omission, they can always go ahead and implement it themselves :-)

docs/reference/contrib/modeladmin/create_edit_delete_views.rst
x-warrior: I have changed the text for your suggestion :)

wagtail/contrib/modeladmin/options.py
x-warrior: I have added this, I do like the names to be more similar.

wagtail/contrib/modeladmin/views.py
x-warrior: Adopted this too: `get_form_fields_exclude(self, request)`

ababic:
So, the general implementation is approved, but I still have yet to pull down and test manually. Also need a little help to make sure the tests are doing what they are supposed to.

x-warrior:
Should I add this to CHANGELOG.txt ?

ababic:
Hey @x-warrior, I tend not to, as it quickly results in merge conflicts as other PRs get merged. things generally run more smoothly if not.

x-warrior:
@ababic have you got a chance to test it?

ababic:
@x-warrior I'm afraid not. I meant to look this week, but I've been in bed with flu for most of it, and am still recovering. So long as the current tests are passing, and there are new tests to show the fields on an edit form changing appropriately when the attribute is set, I'm not sure a manual test really adds anything.

x-warrior:
@ababic oh, sorry! I hope you get better soon. 

My test is not testing the form fields directly, because I'm using `extract_panel_definitions_from_model_class` method which already has an `exclude` parameter and this is already tested on https://github.com/wagtail/wagtail/blob/6f834c9a9b88f5cf660a56d04f1680d0db256cce/wagtail/wagtailadmin/tests/test_edit_handlers.py#L193

My tests basically test if this `extract_panel_definitions_from_model_class` is called with the exclude parameter as `form_fields_exclude` value

ababic:
@x-warrior thanks for the context and well wishes. I can now confirm that I've tried this out locally and it is working as expected. Just a couple of suggestions:

Could we move the `get_exclude_form_fields()` method down a little on the ModelAdmin class? None of the other view-specific methods come until later on, so it seems oddly placed there. I think it would be better placed just before `get_form_view_extra_css()`

ababic:
LGTM

gasman:
Merged in ca16e9187f93466acbfa8671b66cb68230a32e43 - thanks @x-warrior and @ababic!

PR #3296, opened 2017-01-24
[RFC] Replace chooser panels with generic ChooserPanel class and registry

kaedroho:
This pull request contains a proof of concept of an idea I had to simplify Wagtail chooser panels a bit.

We currently have ``*ChooserPanel`` and ``*AdminChooser`` classes for every model in Wagtail and the ``*ChooserPanel`` class is usually the same.

This pull request replaces the indivudal ``*ChooserPanel``  classes (eg ``ImageChooserPanel``, ``DocumentChooserPanel``)  with a single ``ChooserPanel`` class. This class looks up from a registry of widgets to find the correct one for the model.

Some advantages of this approach:
 - Only need to import one edit handler
 - Chooser widgets can be replaced with custom ones
 - Choosers for custom models can be registered without having to be a snippet
 - A global default chooser can be registered (useful for third-party apps that implement a generic model chooser)

I haven't yet ported ``PageChooserPanel`` to the new way since it requires some logic to be moved from the panel to the widget.

### DIFF
--- a/wagtail/wagtailadmin/choosers.py
+++ b/wagtail/wagtailadmin/choosers.py
@@ -1,0 +1,20 @@
+from __future__ import absolute_import, unicode_literals
+
+
+class ChooserRegistry(object):
+    def __init__(self):
+        self._default_widget = None
+        self._widgets = {}
+
+    def register_default_widget(self, widget_cls):
+        self._default_widget = widget_cls
+
+    def register_widget(self, model, widget_cls):
+        self._widgets[model] = widget_cls
+
+    def get_widget(self, model):
+        # TODO check parent models
+        return self._widgets.get(model, self._default_widget)
+
+
+choosers = ChooserRegistry()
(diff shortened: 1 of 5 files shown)


tomdyson:
This sounds like a good idea, and the implementation is surprisingly simple. Are there any disadvantages?

gasman:
Nice! I like the idea of not having to import ImageChooserPanel, DocumentChooserPanel etc from disparate apps, and the "don't make me think" aspect of taking cognitive load away from the developer...

I guess the one place it's not so neat is where you need to pass chooser-specific parameters to the panel, namely PageChooserPanel's `page_type` and `can_choose_root`. Technically it's probably no big deal to pass those extra kwargs from ChooserPanel to the underlying widget, but it loses a bit of purity since the developer is effectively saying "just give me a chooser interface that works with this field, I don't care what it is. Oh, but actually I do, because I want it to behave this way..." :-)

(It's possible that we can minimise the need for those parameters anyhow: `can_choose_root` is unlikely to be useful outside of Wagtail itself, and I _think_ PageChooserPanel already has the ability to pick up the page type from the foreign key whenever something more specific than `Page` is used - if not, it could be made to do so. The one thing we can't handle that way is specifying multiple `page_type`s.)

kaedroho:
Abandoning as I don't consider this a priority for me to work on right now.

I still think generic choosers would be really nice to have, and I think a registry for that would help developers pick the right panel for the right model if there are many more choosable models in the future.

PR #3297, opened 2017-01-24
Allow PageChooserBlock to take a list or tuple of page types

m1kola:
At the moment, `PageChooserBlock` accepts only a single page type (see #3072).

This PR allows `PageChooserBlock` to take a list or tuple of page types which makes `PageChooserBlock` more similar to the `PageChooserPanel` panel.

### DIFF
--- a/wagtail/wagtailcore/blocks/field_block.py
+++ b/wagtail/wagtailcore/blocks/field_block.py
@@ -581,20 +581,42 @@ class ChooserBlock(FieldBlock):
 
 class PageChooserBlock(ChooserBlock):
 
-    def __init__(self, target_model='wagtailcore.Page', can_choose_root=False,
+    # TODO: rename target_model to page_type
+    def __init__(self, target_model=None, can_choose_root=False,
                  **kwargs):
-        self._target_model = target_model
+        if target_model:
+            # Convert single string/model into a list
+            if not isinstance(target_model, (list, tuple)):
+                target_model = [target_model]
+        else:
+            target_model = []
+
+        self._target_models = target_model
         self.can_choose_root = can_choose_root
         super(PageChooserBlock, self).__init__(**kwargs)
 
     @cached_property
     def target_model(self):
-        return resolve_model_string(self._target_model)
+        if len(self.target_models) == 1:
+            return self.target_models[0]
+
+        return resolve_model_string('wagtailcore.Page')
+
+    @cached_property
+    def target_models(self):
+        target_models = []
+
+        for target_model in self._target_models:
+            target_models.append(
+                resolve_model_string(target_model)
+            )
+
+        return target_models
 
     @cached_property
     def widget(self):
         from wagtail.wagtailadmin.widgets import AdminPageChooser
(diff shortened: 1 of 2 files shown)


tomdyson:
Great idea. LGTM, but will need a proper review.

m1kola:
Hi Tom. Thanks for your feedback.

I've added tests. I think, now it's ready for proper review :)

wagtail/wagtailcore/blocks/field_block.py
```diff
@@ -581,20 +581,42 @@ class Meta:
...
+                target_model = [target_model]
+        else:
+            target_model = []
+
+        self._target_models = target_model
         self.can_choose_root = can_choose_root
         super(PageChooserBlock, self).__init__(**kwargs)
 
     @cached_property
     def target_model(self):
```
gasman: As far as I'm aware, the `target_model` property is undocumented and only exists for the use of `PageChooserBlock` itself, so I think it would be appropriate to remove it. (If any external code _is_ accessing the property, it's probably better to let that code fail, to alert the developer that they need to update it to handle multiple page types, rather than returning a not-so-accurate result.)

gasman: Ah, now I see that the base `ChooserBlock` relies on the `target_model` property to do the conversion between ID and instance. (And, in fact, your chosen solution of returning `wagtailcore.Page` in the case of multiple page types is the only thing that can possibly work for [`ChooserBlock.bulk_to_python`](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailcore/blocks/field_block.py#L538).) OK, let's keep it this way :-)

gasman:
Thanks @m1kola! Merged in eb47526888ff8d92daaf023f81bce6e3b86607a8.

pyMan:
Hi @m1kola and @gasman.
Thank you for the new implementation.

When defined multiple page types in `PageChooserBlock's target_model` the "CHOOSE A PAGE" button will become really long and ugly because it includes all the possibile page types.

In the following screenshot "SCEGLI UNA PAGINA" is the italiano "CHOOSE A PAGE" followed by my page types.
screenshot

m1kola:
@pyMan thanks for the report. It looks like a bug to me. Could you, please, [create a new issue](https://github.com/wagtail/wagtail/issues/new) for this?

BertrandBordage:
Aha, funny results! I guess we should use the `text-overflow: ellipsis;` feature here. And maybe a title attribute with the full content if we really want to be able to see it.
By the way @pyMan, why is there a cogs icon near the trashcan icon? I don’t remember this being part of Wagtail.

pyMan:
@BertrandBordage that cogs icon is my custom experimental development

Anyway i think "CHOOSE A PAGE" in enough. When you click it, you can see all the page types in the popup window's title.

BertrandBordage:
Yeah, you’re probably right.
I’m more in favor of specifying the page type only when there’s only one: “Choose a page (your type)” and “Choose a page” in any other case.

PR #3300, opened 2017-01-26
Update model_reference.rst to include get_parent

kwhinnery:
I did not view the source for the `Page` model and thought there was no way to get a page's immediate parent. There is. Including it here in the docs like the other methods.

### DIFF
--- a/docs/reference/pages/model_reference.rst
+++ b/docs/reference/pages/model_reference.rst
@@ -116,6 +116,8 @@ In addition to the model fields provided, ``Page`` has many properties and metho
     .. autoattribute:: preview_modes
 
     .. automethod:: serve_preview
+    
+    .. automethod:: get_parent
 
     .. automethod:: get_ancestors
 


dprothero:
LGTM

gasman:
Thanks @kwhinnery! As a bit of background - `get_parent` and various other tree-traversing methods are provided by the underlying [treebeard](https://tabo.pe/projects/django-treebeard/docs/tip/api.html) library, which is why we haven't documented them ourselves up to now.

A side-effect of that is that the documentation is now picking up the docstring from treebeard, which is not quite ideal as it's talking about nodes rather than pages:

> screen shot 2017-01-26 at 19 55 54

...however, that's certainly better than not mentioning it at all. At some point we should probably write our own Wagtail-centric descriptions for those methods, as well as pointing to the Treebeard docs for further detail...

gasman:
Merged in 6402e56fb5498c6919294452b859728b1f89e7e8

PR #3301, opened 2017-01-27
Allow default JS options to be overwritten for TableBlocks

loicteixeira:
Currently, default options are written on top of user options. This PR switch things around, allowing the user options to overwrite default options.

Note: It allows the user to overwrite any defaults, including the callback for different events. While it will be pretty hard to provide a JS function from Python, this at least allows to remove the callback.

### DIFF
--- a/wagtail/contrib/table_block/static/table_block/js/table.js
+++ b/wagtail/contrib/table_block/static/table_block/js/table.js
@@ -8,6 +8,7 @@ function initTable(id, tableOptions) {
     var tableHeaderCheckbox = $('#' + tableHeaderCheckboxId);
     var colHeaderCheckbox = $('#' + colHeaderCheckboxId);
     var hot;
+    var defaultOptions;
     var finalOptions = {};
     var persist;
     var cellEvent;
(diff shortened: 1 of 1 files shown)


wagtail/contrib/table_block/static/table_block/js/table.js
```diff
@@ -102,25 +92,34 @@ function initTable(id, tableOptions) {
...
+            'redo'
+        ]
+    };
+
+    if (dataForForm !== null && dataForForm.hasOwnProperty('data')) {
+        // Overrides default value from tableOptions (if given) with value from database
+        defaultOptions.data = dataForForm.data;
+    }
+
+    finalOptions = Object.assign({}, defaultOptions, tableOptions);
```
thibaudcolas: The browser support for Object.assign isn't aligned (yet) with Wagtail's support target, so we can't use it without polyfills.

I don't think this JS is loading a polyfill at the moment – this will change once #3012 is merged.

loicteixeira: Updated.

gasman:
Merged in 6e42a7755ada9ba162e78af6535924449ce3287a - thanks @loicteixeira!

PR #3305, opened 2017-01-27
ManyToMany relation support in page models

### LINKED ISSUE #231: Can't add ManytoMany fields to a Page
For example

```
class BlogPage(Page):
    body = RichTextField()
    categories = models.ManyToManyField(BlogCategory, blank=True)
```

One gets ValueError "<BlogPage: foo>" needs to have a value for field "blogpage" before this many-to-many relationship can be used.

It's caused by this 
https://github.com/torchbox/django-modelcluster/issues/3
https://github.com/torchbox/django-modelcluster/blob/master/modelcluster/forms.py#L260

I'll keep trying to think of a solution. Seems like a common use case as Wordpress implements categories as a M2M (you can pick multiple, but they are predefined unlike a tag).

gasman:
Fixes #231. Todo list from https://github.com/wagtail/wagtail/issues/231#issuecomment-272175678:

 - [x] release django-modelcluster 3.0 final and update wagtail's setup.py to require django-modelcluster>=3.0,<4.0
 - [x] add a release note to document the updated django-modelcluster / django-taggit dependencies (since it's somewhat likely that people will have the old versions hard-coded in their requirements files)
 - [x] ensure that M2M relations are covered by tests
 - [x] document the use of ParentalManyToManyField, possibly through a new 'categories' section in the tutorial (which was dropped from #3125 (comment) because it was too messy to implement without real M2M support) as well as http://docs.wagtail.io/en/v1.8/topics/pages.html


gasman:
CI tests are currently failing because we're relying on a fix in modelcluster made today (which hasn't made it into a packaged version of the library yet): https://github.com/wagtail/django-modelcluster/commit/a1d7f030730d24876e0039a485a3f0c89e23b180

gasman:
Ready for review!

docs/getting_started/tutorial.rst
```diff
@@ -707,6 +708,107 @@ something like this:
...
+        ]
+
+        def __str__(self):
+            return self.name
+
+        class Meta:
+            verbose_name_plural = 'blog categories'
+
+
+Note that we are using ``panels`` rather than ``content_panels`` here - since snippets generally have no need for fields such as slug or publish date, the editing interface for them is not split into separate 'content' / 'promote' / 'settings' tabs as standard, and so there is no need to distinguish between 'content panels' and 'promote panels'.
```
kaedroho: While extremely useful, I think we should try to keep side-comments like this to a minimum in the getting started docs, or perhaps display them as such by putting them in a ``topic:`` block.

gasman: Happy to consider alternative formatting here. I think the sentence does belong here in the tutorial though - I've seen various people in the support channels get this wrong, and I think it's more likely to stick in people's minds if we explain the reason for the difference rather than essentially leaving it at "it's different because it is".

gasman:
Merged in b3e723eea3a7aed670fb9dc6530efe146d2a2715

helderco:
How do you upgrade from previous solution with InlinePanel?

Issue #3307, opened 2017-01-28
Add ability to run only some tests or TestCases


chimeno:
Django has the ability to run only some TestCases or individual tests, after tried within wagtail with:
````python runtests.py wagtail.wagtailadmin.TestAccountSection```` or
````python runtests.py wagtail.wagtailadmin.tests.TestAccountSection```` gives:
````AttributeError: 'module' object has no attribute 'TestAccountSection'````

We should add a feature like this in order to reduce tests duration.

[Django docs about testing](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/unit-tests/#running-only-some-of-the-tests)

gasman:
Hi @chimeno,

This does work - you've just got the module path wrong :-) Try:

    python ./runtests.py wagtail.wagtailadmin.tests.test_account_management.TestAccountSection

PR #3308, opened 2017-01-28
Improve docs testing

chimeno:
Added in docs a way to run only some tests.
Related: #3307


gasman:
Looks good to me - thanks!

PR #3310, opened 2017-01-29
#3306 add user prefered admin language

### LINKED ISSUE #3306: Make admin language configurable on a per-user basis
**Issue**
I was trying to have the admin in my native language ``es-es`` (in order to improve translations) without setting the ``LANGUAGE_CODE`` in a global way.

**Solution**
Make the language of the admin being configurable on a per-user basis, adding a ``prefered_language`` field in the ``UserProfile``.

Is this something you could see in wagtail core? I can make a PR in that case.

Related issue: #3205

chimeno:
First working attempt to fix #3306 

USE CASE:
A wagtail user (administrator or editor) wants the admin (and only the admin section) be shown in a specific language.

-  Add a field to user profile and provide a migration with a default value
-  Make form to update that field
-  Persist that change in successive requests.
-  Check all of the above with tests

Some questions about the HOW:
- I got the languages choices from ``settings.LANGUAGES`` but maybe it's more useful to get only languages provided by wagtail. (A conf setting or a dynamically list from locales).
- Is prefered_language default value to 'en' a good choice?
- In order to persists the preference in successive requests I've used a custom middleware to solve the (only admin section), although there are other ways (session).
- The middleware needs a better way to detect if the request.path is in the admin section.

**Updated**
TODO:

- [x] Move language choices to a manual list of wagtail languages.
- [x] Default value to LANGUAGE_CODE
- [x] Fix migration
- [x] Move the ``activate`` from a middleware to an admin decorator.
- [x] Check hard-coded strings.
- [x] Notifications emails in recipient's language
- [x] Add docs


wagtail/wagtailadmin/templates/wagtailadmin/account/account.html
```diff
@@ -40,6 +40,17 @@
                     </small>
                 </li>
             {% endif %}
+                <li class="row row-flush">
+                    <div class="col6">
+                        <a href="{% url 'wagtailadmin_account_language_preferences' %}" class="button button-primary">{% trans "Language preferences" %}</a>
+                    </div>
+
+                    <small class="col6">
+                        {% trans "Choose which language do you want to use in this admin." %}
```
tomdyson: "Choose which language do you want to use in this admin" -> "Choose the language you want to use here"

wagtail/wagtailadmin/views/account.py
```diff
@@ -107,6 +108,23 @@ def notification_preferences(request):
...
 
+def language_preferences(request):
+    if request.method == 'POST':
+        form = PreferedLanguageForm(request.POST, instance=UserProfile.get_for_user(request.user))
+
+        if form.is_valid():
+            user_profile = form.save()
+            # This will set the language only for this request/thread
+            translation.activate(user_profile.prefered_language)
+            messages.success(request, _("Your preferences have been update successfully!"))
```
tomdyson: "Your preferences have been update successfully!" -> "Your preferences have been updated."

wagtail/wagtailusers/forms.py
```diff
@@ -375,3 +375,12 @@ def __init__(self, *args, **kwargs):
     class Meta:
         model = UserProfile
         fields = ("submitted_notifications", "approved_notifications", "rejected_notifications")
+
+
+class PreferedLanguageForm(forms.ModelForm):
```
tomdyson: PreferedLanguageForm -> PreferredLanguageForm

wagtail/wagtailusers/models.py
```diff
@@ -30,6 +30,14 @@ class UserProfile(models.Model):
         help_text=_("Receive notification when your page edit is rejected")
     )
 
+    prefered_language = models.CharField(
```
tomdyson: prefered_language -> preferred_language

wagtail/wagtailadmin/tests/test_account_management.py
```diff
@@ -275,6 +275,39 @@ def test_notification_preferences_view_post(self):
...
+        self.assertContains(response, "Language Preferences")
+
+    def test_language_preferences_view_post(self):
+        """
+        This post to the language preferences view and checks that the
+        user's profile is updated
+        """
+        # Post new values to the language preferences page
+        post_data = {
+            'prefered_language': 'es'
```
tomdyson: prefered -> preferred

wagtail/wagtailadmin/views/account.py
```diff
@@ -10,13 +10,14 @@
...
 from django.utils.translation import ugettext as _
+from django.utils import translation
 from django.views.decorators.cache import never_cache
 from django.views.decorators.debug import sensitive_post_parameters
 
 from wagtail.utils.compat import user_is_authenticated
 from wagtail.wagtailadmin import forms
 from wagtail.wagtailcore.models import UserPagePermissionsProxy
-from wagtail.wagtailusers.forms import NotificationPreferencesForm
+from wagtail.wagtailusers.forms import NotificationPreferencesForm, PreferedLanguageForm
```
tomdyson: prefered -> preferred

wagtail/wagtailadmin/tests/test_account_management.py
```diff
@@ -275,6 +275,39 @@ def test_notification_preferences_view_post(self):
...
+        }
+        response = self.client.post(reverse('wagtailadmin_account_language_preferences'), post_data)
+
+        # Check that the user was redirected to the account language preferences page
+        self.assertEqual(response.status_code, 200)
+
+        profile = UserProfile.get_for_user(get_user_model().objects.get(pk=self.user.pk))
+
+        # Check that the language preferences are stored
+        self.assertEqual(profile.prefered_language, 'es')
```
tomdyson: prefered -> preferred

wagtail/wagtailadmin/views/account.py
```diff
@@ -107,6 +108,23 @@ def notification_preferences(request):
     })
 
 
+def language_preferences(request):
+    if request.method == 'POST':
+        form = PreferedLanguageForm(request.POST, instance=UserProfile.get_for_user(request.user))
```
tomdyson: prefered -> preferred

wagtail/wagtailadmin/views/account.py
```diff
@@ -107,6 +108,23 @@ def notification_preferences(request):
...
+
+        if form.is_valid():
+            user_profile = form.save()
+            # This will set the language only for this request/thread
+            translation.activate(user_profile.prefered_language)
+            messages.success(request, _("Your preferences have been update successfully!"))
+    else:
+        form = PreferedLanguageForm(instance=UserProfile.get_for_user(request.user))
```
tomdyson: prefered -> preferred

wagtail/wagtailcore/middleware.py
```diff
@@ -22,3 +22,15 @@ def process_request(self, request):
...
+class LocaleAdminMiddleware(MiddlewareMixin):
+    def process_request(self, request):
+        """
+        Set the user prefered language in wagtail admin urls
+        :param request:
+        :return:
+        """
+        if 'admin' in request.path:  # TODO: Change to a proper way to look if is admin
+            if request.user:
+                activate(request.user.wagtail_userprofile.prefered_language)
```
tomdyson: prefered -> preferred

wagtail/wagtailusers/migrations/0006_userprofile_prefered_language.py
```diff
@@ -0,0 +1,20 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('wagtailusers', '0005_make_related_name_wagtail_specific'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='userprofile',
+            name='prefered_language',
```
tomdyson: prefered -> preferred

chimeno:
I've corrected the typos and made some general fixes.
¿Are you ok with the middleware approach?
As I've said before, any feedback is welcome ;)

gasman:
This is looking really good - thanks @chimeno!

> * I got the languages choices from settings.LANGUAGES but maybe it's more useful to get only languages provided by wagtail. (A conf setting or a dynamically list from locales).

I think we should have a default list defined in wagtailadmin, consisting of the languages that have >=90% coverage (currently: English, German, Portuguese (Brazil), Spanish, Romanian, French, Icelandic (Iceland), Italian, Norwegian Bokmål, Polish, Portuguese (Portugal), Russian, Dutch (Netherlands), Finnish, Galician) - we should probably maintain that list manually, rather than try to do anything clever with the Transifex import - and have a config setting so that people can override the default list (e.g. to add a language that's below 90% coverage, or exclude languages for political reasons :-) ... in fact, it would be a good idea to disable the 'language preferences' option completely when there's only one language specified).

> * Is prefered_language default value to 'en' a good choice?

We should default to the global `LANGUAGE_CODE` setting. Perhaps we should make the `preferred_language` field `null=True`, and fall back on the global setting whenever `preferred_language` is null - that way, any changes to `LANGUAGE_CODE` will continue to take effect for all users who haven't explicitly chosen a language.

> * In order to persists the preference in successive requests I've used a custom middleware to solve the (only admin section), although there are other ways (session).
> * The middleware needs a better way to detect if the request.path is in the admin section.

I suggest that we put the `activate(request.user.wagtail_userprofile.preferred_language)` line inside the `require_admin_access` decorator (wagtail/wagtailadmin/decorators.py). All admin views should be using this decorator - in most cases (including URLs registered with `register_admin_urls`) this happens [in the urlconf](https://github.com/wagtail/wagtail/blob/5aa8e00d77d20060f75b7ca65e28ef502c54245c/wagtail/wagtailadmin/urls/__init__.py#L59). This way, we don't need a new middleware - I'd definitely like to avoid this if possible, as it's another "moving part" that has to be added to everyone's Wagtail projects (and another bit of overhead on every page request).

wagtail/wagtailadmin/views/account.py
```diff
@@ -107,6 +108,23 @@ def notification_preferences(request):
...
 
+def language_preferences(request):
+    if request.method == 'POST':
+        form = PreferredLanguageForm(request.POST, instance=UserProfile.get_for_user(request.user))
+
+        if form.is_valid():
+            user_profile = form.save()
+            # This will set the language only for this request/thread
+            activate(user_profile.preferred_language)
+            messages.success(request, _("Your preferences have been updated successfully!."))
```
gasman: This still needs changing to "Your preferences have been updated."

wagtail/wagtailusers/migrations/0006_userprofile_prefered_language.py
```diff
@@ -0,0 +1,20 @@
...
..., ('hi', 'Hindi'), ('hr', 'Croatian'), ('hsb', 'Upper Sorbian'), ('hu', 'Hungarian'), ('ia', 'Interlingua'), ('id', 'Indonesian'), ('io', 'Ido'), ('is', 'Icelandic'), ('it', 'Italian'), ('ja', 'Japanese'), ('ka', 'Georgian'), ('kk', 'Kazakh'), ('km', 'Khmer'), ('kn', 'Kannada'), ('ko', 'Korean'), ('lb', 'Luxembourgish'), ('lt', 'Lithuanian'), ('lv', 'Latvian'), ('mk', 'Macedonian'), ('ml', 'Malayalam'), ('mn', 'Mongolian'), ('mr', 'Marathi'), ('my', 'Burmese'), ('nb', 'Norwegian Bokmål'), ('ne', 'Nepali'), ('nl', 'Dutch'), ('nn', 'Norwegian Nynorsk'), ('os', 'Ossetic'), ('pa', 'Punjabi'), ('pl', 'Polish'), ('pt', 'Portuguese'), ('pt-br', 'Brazilian Portuguese'), ('ro', 'Romanian'), ('ru', 'Russian'), ('sk', 'Slovak'), ('sl', 'Slovenian'), ('sq', 'Albanian'), ('sr', 'Serbian'), ('sr-latn', 'Serbian Latin'), ('sv', 'Swedish'), ('sw', 'Swahili'), ('ta', 'Tamil'), ('te', 'Telugu'), ('th', 'Thai'), ('tr', 'Turkish'), ('tt', 'Tatar'), ('udm', 'Udmurt'), ('uk', 'Ukrainian'), ('ur', 'Urdu'), ('vi', 'Vietnamese'), ('zh-hans', 'Simplified Chinese'), ('zh-hant', 'Traditional Chinese')], default='en', help_text='Select language for the admin', max_length=10, verbose_name='preferred language'),
```
gasman: We should avoid hard-coding the language list in the migration, so that we don't need a new migration whenever the list is updated. (I guess the simplest way is to define the choices list within `PreferredLanguageForm`, rather than at the model level...)

gasman:
Another thing that we need to remember here... when we send notification emails, they need to be in the recipient's language, NOT the language of the user triggering them. (I've even seen Facebook get this wrong :-) )

chimeno:
Thanks @gasman for the feedback! I've updated the first comment with a TODO to work on it over the weekend.
Thanks!

chimeno:
I've updated the code with feedback provided, if you are OK with these changes, I would write docs and make the notifications sections language aware.
Some notes:
 - I'm not sure if the language code format used (xy-zw) is better than (xy-ZW) provided by locales.
- I put the provided languages list  in ``wagtailadmin/utils.py`` but we can move it if there is a more elegant location.
- The drone linting is not making linting in this branch.
- I haven't used the ``null=True`` approach in ``preferred_language`` as Django docs doesn't encourage it. Instead, I've used a ``default='en'`` and a helper function to back to ``LANGUAGE_CODE``.
Any feedback/criticism is more than welcome :)

chimeno:
This should be now ready for review.
Any feedback/criticism is more than welcome :)

wagtail/wagtailadmin/tests/test_account_management.py
```diff
@@ -275,6 +277,51 @@ def test_notification_preferences_view_post(self):
...
+        # Check that the user received an account page
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, 'wagtailadmin/account/language_preferences.html')
+
+        # Page should contain a 'Language Preferences' title
+        self.assertContains(response, "Language Preferences")
+
+    def test_language_preferences_view_post(self):
+        """
+        This post to the language preferences view and checks that the
```
BertrandBordage: “This post**s**”

wagtail/wagtailadmin/tests/test_account_management.py
```diff
@@ -275,6 +277,51 @@ def test_notification_preferences_view_post(self):
...
+        user's profile is updated
```
BertrandBordage: “user profile”

wagtail/wagtailadmin/views/account.py
```diff
@@ -40,7 +42,8 @@ def account(request):
 
     return render(request, 'wagtailadmin/account/account.html', {
         'show_change_password': password_management_enabled() and request.user.has_usable_password(),
-        'show_notification_preferences': show_notification_preferences
+        'show_notification_preferences': show_notification_preferences,
+        'show_preferred_language_preferences': len(get_available_admin_languages()) > 1
```
BertrandBordage: It can be replaced with:
```python
bool(get_available_admin_languages())
```

wagtail/wagtailadmin/utils.py
```diff
@@ -19,6 +21,30 @@
...
+    ('pt-pt', gettext_lazy('Portuguese')),
+    ('ru', gettext_lazy('Russian')),
+    ('nl-nl', gettext_lazy('Netherlands Dutch')),
+    ('fi', gettext_lazy('Finish')),
+    ('ga', gettext_lazy('Galician'))
+]
+
+
+def get_available_admin_languages():
+    return getattr(settings, 'WAGTAILADMIN_PERMITTED_LANGUAGES', WAGTAILADMIN_PROVIDED_LANGUAGES)
```
BertrandBordage: Shouldn’t we combine `WAGTAILADMIN_PROVIDED_LANGUAGES` with `LANGUAGES`?
It feels irrelevant to propose Russian and Icelandic in a site where editors are supposed to type (and therefore know) only French & English (in the cases I see daily).

Another good reason to combine with `LANGUAGES`: if a clumsy beginner user selects the wrong language, he might end up with the back office in a language and alphabet he doesn’t know, so he may not find again how to change it back.

CONTRIBUTORS.rst
```diff
@@ -214,6 +214,9 @@ Contributors
 * Vorlif
 * Kjartan Sverrisson
 * LB (Ben Johnston)
+* Daniel Chimeno
+
+
```
BertrandBordage: Why did you add two newlines?

wagtail/wagtailadmin/views/account.py
gasman: `bool(get_available_admin_languages())` wouldn't be correct, because we also want the option to be hidden when there is _one_ language available.

BertrandBordage: Ah right, sorry, read too fast, I thought it was `>=` :(

docs/advanced_topics/i18n/index.rst
```diff
@@ -17,6 +17,26 @@ The Wagtail admin backend has been translated into many different languages. You
...
+
+.. code-block:: python
+
+    from django.conf import settings
+    WAGTAILADMIN_PERMITTED_LANGUAGES = [ ('en' , 'English'), ('pt', 'Portuguese') ]
+
+In case there is zero or one language permitted, the form will be hidden.
+
+If there is no language selected either by the user or by the Site Administrator, the ``LANGUAGE_CODE`` wil be used.
+
```
gasman: `WAGTAILADMIN_PERMITTED_LANGUAGES ` should also be documented on http://docs.wagtail.io/en/latest/advanced_topics/settings.html.

Some further fixes:

* "Logged users" => "Logged-in users"
* "that have a coverage >= 90% translation" => "that have >= 90% translation coverage"
* "It is possible to override this list by adding a setting configuration." => "It is possible to override this list via the `WAGTAILADMIN_PERMITTED_LANGUAGES` setting"
* "selected either by the user or by the Site Administrator" => remove "or by the Site Administrator". (`LANGUAGE_CODE` _is_ the place where the site administrator specifies a global default language, so it doesn't make sense to say "this will be used if the site administrator hasn't specified a language")

gasman:
Merged in c965212003d360cbc620ddae315082d4da5b26d7 + parents - thanks @chimeno!

Issue #2514, opened 2016-04-25
Add ability to specify min/max block counts on StreamField / StreamBlock


gasman:
As requested here: https://groups.google.com/d/msg/wagtail/RLMwoDN_5LA/Endv7GANAQAJ

Suggest this is done with parameters named `min_num` and `max_num`, to match #1529.

Related: #2379

chimeno:
In case it's not possible right now (I haven't tried yet), it would be great to have a ``min_num``and ``max_num`` per field in the StreamField.
Examples:
  - One and only one event block.
  - More than two and less than 5 big images.

As I said, maybe it's possible right now but I'm not aware.

StriveForBest:
@gasman do you guys have an update on this? Would be amazing to be able to limit the streamblock instances.

m1kola:
Fix (#3605) was released in [Wagtail 1.12](http://docs.wagtail.io/en/v1.13.1/releases/1.12.html#other-features):

> StreamBlock now accepts min_num, max_num and block_counts settings to control the minimum and maximum numbers of blocks (Edwar Baron, Matt Westcott)

Closing this issue

PR #3315, opened 2017-01-31
Honour StreamField blank flag and add StreamBlock required flag

### LINKED ISSUE #2236: StreamBlock(required=True) doesn't work

loicteixeira:
Most of the work happens in the StreamBlock since this is what the StreamField uses under the hood (instead of duplicating the check in the `clean` method of the StreamField).

Fix #2236, although I think #2514 could be tackled in the same PR (having a field required means that the minimum number of child is 1, which is half of the work done for this other issue). Thoughts?

I haven't updated the documentation yet as I wanted to make sure you agree with the approach first.

### DIFF
--- a/wagtail/wagtailcore/blocks/stream_block.py
+++ b/wagtail/wagtailcore/blocks/stream_block.py
@@ -34,7 +34,7 @@ class StreamBlockValidationError(ValidationError):
 
 class BaseStreamBlock(Block):
 
-    def __init__(self, local_blocks=None, **kwargs):
+    def __init__(self, local_blocks=None, required=True, **kwargs):
         self._constructor_kwargs = kwargs
 
         super(BaseStreamBlock, self).__init__(**kwargs)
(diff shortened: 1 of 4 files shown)


loicteixeira:
Umm the non-field-error display will need some love. The error will stay in place when you add a block. Should it disappear?

non-field errors glitch

thibaudcolas:
@loicteixeira IMHO when the first block is added the error should disappear. Even if the field is still empty, having a block already should be enough for the Streamfield to validate as "having content".

gasman:
Thanks @loicteixeira! Merged in 4a5f3072bd9fa8d56c02cefcfe9c747d2b22d04d (with some minor tweaks - `required` is now managed as a `Meta` property, which simplifies the logic a bit and allows it to be set at the class level when subclassing StreamBlock).

PR #3318, opened 2017-01-31
Update contribution documentation for core team members

thibaudcolas:
Here are changes to reflect the clarifications and suggestions discussed during the last core team meeting.

I tried to update the existing guidelines as much as possible instead of creating separate section(s).

At first I didn't realise the issue tracking documentation was in the Wagtail docs – I thought it was in https://github.com/wagtail/wagtail/wiki/Issue-tracking-process. I think this page should be deleted if it is superseded by the Wagtail docs (as with all of the wiki?).

### DIFF
--- a/docs/contributing/committing.rst
+++ b/docs/contributing/committing.rst
@@ -8,6 +8,9 @@ or for anyone interested in the process of getting code committed to Wagtail.
 Code should only be committed after it has been reviewed
 by at least one other reviewer or committer,
 unless the change is a small documentation change or fixing a typo.
+If additional code changes are made after the review, it is OK to commit them
+without further review if they are uncontroversial and small enough that no
+bugs are likely to be introduced.
 
 Most code contributions will be in the form of pull requests from Github.
 Pull requests should not be merged from Github, apart from small documentation fixes,
@@ -41,6 +44,8 @@ as merge commits make the commit history harder to read for small changes.
 You can fix up any small mistakes in the commits,
 such as typos and formatting, as part of the rebase.
 ``git rebase --interactive`` is an excellent tool for this job.
+
+Ideally, use this as an opportunity to squash the changes to a few commits, so each commit is making a single meaningful change (and not breaking anything). If this is not possible because of the nature of the changes, go with the least ugly.
 
 .. code-block:: console
 
@@ -119,3 +124,11 @@ The changes are ready to be pushed to ``master`` now.
     $ # Push the commits!
     $ git push upstream master
     $ git branch -d pr/xxxx
+
+When you have made a mistake
+============================
+
+It's ok! Everyone makes mistakes. If you realise that recent merged changes
+have a negative impact, create a new pull request with a revert of the changes
+and merge it without waiting for a review. The PR will serve as additional
+documentation for the changes, and will run through the CI tests.
--- a/docs/contributing/committing.rst
+++ b/docs/contributing/committing.rst
@@ -9,8 +9,8 @@ Code should only be committed after it has been reviewed
 by at least one other reviewer or committer,
 unless the change is a small documentation change or fixing a typo.
 If additional code changes are made after the review, it is OK to commit them
-without further review if they are uncontroversial and small enough that no
-bugs are likely to be introduced.
+without further review if they are uncontroversial and small enough that
+there is minimal chance of introducing new bugs.
 
 Most code contributions will be in the form of pull requests from Github.
 Pull requests should not be merged from Github, apart from small documentation fixes,
@@ -45,7 +45,11 @@ You can fix up any small mistakes in the commits,
 such as typos and formatting, as part of the rebase.
 ``git rebase --interactive`` is an excellent tool for this job.
 
-Ideally, use this as an opportunity to squash the changes to a few commits, so each commit is making a single meaningful change (and not breaking anything). If this is not possible because of the nature of the changes, go with the least ugly.
+Ideally, use this as an opportunity to squash the changes to a few commits, so
+each commit is making a single meaningful change (and not breaking anything).
+If this is not possible because of the nature of the changes, it's acceptable
+to either squash into a commit or leave all commits unsquashed,
+depending on which will be more readable in the commit history.
 
 .. code-block:: console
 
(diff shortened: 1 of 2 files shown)


docs/contributing/committing.rst
```diff
@@ -42,6 +45,8 @@ You can fix up any small mistakes in the commits,
 such as typos and formatting, as part of the rebase.
 ``git rebase --interactive`` is an excellent tool for this job.
 
+Ideally, use this as an opportunity to squash the changes to a few commits, so each commit is making a single meaningful change (and not breaking anything). If this is not possible because of the nature of the changes, go with the least ugly.
```
thibaudcolas: I think it could be interesting to specify this further than "least ugly". If we want to follow the "each commit does not break anything" rule, then PRs where a few commits isn't doable should be squashed to a single commit?

docs/contributing/committing.rst
```diff
@@ -119,3 +124,11 @@ The changes are ready to be pushed to ``master`` now.
     $ # Push the commits!
     $ git push upstream master
     $ git branch -d pr/xxxx
+
+When you have made a mistake
+============================
+
+It's ok! Everyone makes mistakes. If you realise that recent merged changes
```
thibaudcolas: I wasn't sure whether it was appropriate to create a separate section for that (and not give it an explicit name). I imagine this could grow in the future, with other troubleshooting/rollback guidelines.

mikedingjan:
Perhaps it's interesting to enable ``Require pull request reviews before merging`` in the project settings on GitHub? This way we really enforce reviewing PR's since it'll become one of the status checks.

gasman:
Hi @mikedingjan - please see the last point at https://github.com/wagtail/wagtail/pull/3021#issuecomment-273950264. When we tried protected branches some time back, we found that they were much too inflexible for our PR merging workflow.

docs/contributing/committing.rst
```diff
@@ -8,6 +8,9 @@ or for anyone interested in the process of getting code committed to Wagtail.
 Code should only be committed after it has been reviewed
 by at least one other reviewer or committer,
 unless the change is a small documentation change or fixing a typo.
+If additional code changes are made after the review, it is OK to commit them
+without further review if they are uncontroversial and small enough that no
+bugs are likely to be introduced.
```
gasman: The word "likely" seems a bit weak here - it feels like there's too much room for people to say "I'm a really careful programmer - of course it's unlikely that I introduced new bugs" :-)

Maybe: "small enough that there is minimal chance of introducing new bugs"?

docs/contributing/committing.rst
gasman: As mentioned at the team meeting, I think "least ugly" is best open to the committer's interpretation - I don't think there's a good way to formalise it further.

However, I think "least ugly" needs more explanation in this sentence, because we haven't specified what the alternatives are. Maybe: "If this is not possible because of the nature of the changes, it's acceptable to either squash into a commit or leave all commits unsquashed, depending on which will be more readable in the commit history."

docs/contributing/committing.rst
gasman: Looks good to me!

docs/contributing/issue_tracking.rst
```diff
@@ -10,31 +10,38 @@ An issue must always correspond to a specific action with a well-defined complet
...
 
-As soon as a ticket is opened - ideally within one day - a member of the core team will give it an initial classification, by either closing it as invalid or assigning it to a milestone. Don't be discouraged if you feel that your ticket has been given a lower priority than it deserves - this decision isn't permanent. We will consider all feedback, and reassign or reopen tickets where appropriate.
+As soon as a ticket is opened - ideally within one day - any member of the core team can give it an initial classification, by either closing it as invalid or assigning it to a milestone. Don't be discouraged if you feel that your ticket has been given a lower priority than it deserves - this decision isn't permanent. We will consider all feedback, and reassign or reopen tickets where appropriate.
```
gasman: I prefer the original wording here. This document is written from the perspective of "this is what will happen to your ticket, as a contributor" - we should avoid mixing it with instructions to the core team.

docs/contributing/issue_tracking.rst
```diff
@@ -10,31 +10,38 @@ An issue must always correspond to a specific action with a well-defined complet
...
-* **some-day**: the issue is accepted as valid (i.e. it's a bug report for a legitimate bug, or a useful feature request) but not deemed a priority to work on (in the opinion of the core team). For example - a bug that's only cosmetic, or a feature that would be kind of neat but not really essential. Feel free to take it on!
-* **real-soon-now**: the issue is considered important by the core team (roughly speaking: it's a bug or missing feature that's causing headaches for people) but isn't necessarily going to be completed on any set timescale. Feel free to take it on!
-* A specific version number: the issue is important enough that it needs to be fixed in this version.
+* **some-day**: the issue is accepted as valid (i.e. it's a bug report for a legitimate bug, or a useful feature request) but not deemed a priority to work on (in the opinion of the core team). For example - a bug that's only cosmetic, or a feature that would be kind of neat but not really essential. There are no resources allocated to it, feel free to take it on!
```
gasman: Grammatically there should be a dash or semicolon here (because each half is a complete sentence): "There are no resources allocated to it - feel free to take it on!"

docs/contributing/issue_tracking.rst
```diff
@@ -10,31 +10,38 @@ An issue must always correspond to a specific action with a well-defined complet
...
 
 On some occasions it may take longer for the core team to classify an issue into a milestone. For example:
 
 * It may require a non-trivial amount of work to confirm the presence of a bug. In this case, feedback and further details from other contributors, whether or not they can replicate the bug, would be particularly welcomed.
 * It may require further discussion to decide whether the proposal is a good idea or not - if so, it will be tagged "design decision needed".
 
-We will endeavour to make sure that issues don't remain in this state for prolonged periods. Issues and PRs tagged "design decision needed" will be revisited regularly and discussed with at least two core contributors - we aim to review each ticket at least once per release cycle (= 6 weeks).
+We will endeavour to make sure that issues don't remain in this state for prolonged periods. Issues and PRs tagged "design decision needed" will be revisited regularly and discussed with at least two core contributors - we aim to review each ticket at least once per release cycle (= 6 weeks) as part of weekly core team meetings (2-3 a week).
```
gasman: I think it's best to leave out "(2-3 a week)" here - I'd prefer to leave this flexible rather than be 'set in stone' in a document. Also, 2-3 a week wouldn't be enough to cover the current backlog in a 6-week cycle.

(If we keep it in, we should say "2-3 issues a week", otherwise it sounds like we're talking about 2-3 weekly meetings a week ;-) )

docs/contributing/issue_tracking.rst
thibaudcolas: I did notice that, and honestly found it very confusing. This is all about issue triage, which in my opinion would be more understandable for more audiences if it wasn't written from any perspective.

This isn't was this PR is about though so I've just reverted those changes 🙂.

thibaudcolas:
Adjusted and ready for review again 🚀.

gasman:
Looks good to me - thanks!

PR #3319, opened 2017-01-31
Move CONTRIBUTING file to .github subfolder

thibaudcolas:
To declutter the root on the GitHub view. The drawback is that it's not as visible for people who expect it to be there – but since there is some info about contributing in the README itself, it feels appropriate.


gasman:
Hi @thibaudcolas,

I don't really understand the motivation for this change - if we're deciding that decluttering the root is a desirable goal (and I'm not convinced that it is...), surely the aim is to eliminate the machine-readable gunk and make human-readable files like CONTRIBUTING **more** visible, not the other way round...?

thibaudcolas:
Hey hey,

sorry, I didn't realise that there wasn't a consensus around that goal. Tom mentioned it https://github.com/wagtail/wagtail/pull/3021#issuecomment-266998843, which is why I thought this PR would be interesting. Personally I think it's a desirable goal, but not so important in the grand scheme of things – if there is no clear consensus, it's not worth debating IMHO.

-----

Now for the reasoning – in my opinion the point of `CONTRIBUTING.md` is the integration with GitHub, nothing else. The project has contribution info that is more appropriate for humans in its README, and its purpose-built documentation. I agree other machine-readable files should be moved as much as possible, it just happens to be very straightforward for this one file.

gasman:
I think there's still value in `CONTRIBUTING.md` as a document in its own right, not just for the Github integration... let's keep it in its current place.

PR #3320, opened 2017-01-31
Add page title to title attrs of explorer action buttons

gasman:
A screen-reader user reported that the page explorer is difficult to use because there's no indication of which pages the action button links correspond to. This PR adds the page title to each link's title attribute, and also changes the hidden text on the '+' button from "add a subpage" to "add a child page" for consistency.


gasman:
Thanks @jjanssen! Merged in 753efbb29e17a6c00c00808833d2bbd473dd08b7.

PR #1235, opened 2015-04-24
Search promotions module

kaedroho:
Moves editors picks out of ``wagtailsearch`` into a new module called ``contrib.wagtailsearchpicks``.

This makes "Promoted search results" optional and also resolves the two-way dependency between ``wagtailsearch`` and ``wagtailcore``

TODO:
 - [x] Move ``EditorsPick`` model
 - [x] Add ``{% get_search_picks %}`` template tag
 - ~~Move locale files~~ (best to do this just before/after merge)
 - [x] Documentation

PR #3321, opened 2017-01-31
Updates docs to reflect Removed editors picks

bmihelac:
Removed in:

https://github.com/wagtail/wagtail/commit/df2034006064ce55b3016f898ab62c0deba4cd9e#diff-41e5e1d4e3130a4d95713ee7757ece86

### DIFF
--- a/docs/editor_manual/administrator_tasks/index.rst
+++ b/docs/editor_manual/administrator_tasks/index.rst
@@ -7,4 +7,4 @@ This section of the guide documents how to perform common tasks as an administra
    :maxdepth: 3
 
    managing_users
-   promoted_search_results
+


gasman:
Hi @bmihelac - thanks for the PR!

Since the editor's picks feature wasn't removed entirely, but rather moved to an optional contrib module (#1235), I think it would make sense to keep the page in the editor's guide, but add a note at the top of the page saying that the feature is only available when the relevant module is installed.

bmihelac:
Oh, I did not know, and was looking and looking until I found that it isn't there :) if I can do anything related to this please let me know

gasman:
Now addressed by #3343

PR #3330, opened 2017-02-03
Removing errant words in sample path/to/wagtail

daaray:
Walking through the developer docs in preparation for the Sprint next week, I noticed some unnecessary words.


tomdyson:
Good spot. Thanks @daaray!

gasman:
Merged in 49b78dd8a7023252a0b196f8fb88a80e7ffe64d2

PR #3335, opened 2017-02-05
Ordered hooks

Gagaro:
The idea is to be able to choose when the hooks are run in regards to the other ones. My issue for now is that my hook (`construct_homepage_summary_items`) is called *before* the ones in the image/doc modules.

So with this PR, we could ensure that our hook is called after:

```python
@hooks.register('construct_homepage_summary_items', order=1)
def add_another_panel(request, panels):
    [...]
```

Tell me if this is something you are willing to merge so I add tests/docs.

Another thing I'd like to do is to set the order of every hooks registered in the wagtail modules to `-1` so that the user hooks would run after the wagtail ones by default (I think this should be the expected behaviour when created custom hooks). What do you think about that?

Thanks

### DIFF
--- a/docs/reference/hooks.rst
+++ b/docs/reference/hooks.rst
@@ -23,6 +23,21 @@ Alternatively, ``hooks.register`` can be called as an ordinary function, passing
 
   hooks.register('name_of_hook', my_hook_function)
 
+If you need your hooks to run in a particular order, you can pass the ``order`` parameter:
+
+.. code-block:: python
+
+  @hooks.register('name_of_hook', order=1)  # This will run after every hook in the wagtail core
+  def my_hook_function(arg1, arg2...)
+      # your code here
+
+  @hooks.register('name_of_hook', order=-1)  # This will run before every hook in the wagtail core
+  def my_other_hook_function(arg1, arg2...)
+      # your code here
+
+  @hooks.register('name_of_hook', order=2)  # This will run after `my_hook_function`
+  def yet_another_hook_function(arg1, arg2...)
+      # your code here
 
 The available hooks are listed below.
 
--- a/wagtail/tests/utils.py
+++ b/wagtail/tests/utils.py
@@ -104,14 +104,14 @@ class WagtailTestUtils(object):
         return _AssertLogsContext(self, logger, level)
 
     @contextmanager
-    def register_hook(self, hook_name, fn):
+    def register_hook(self, hook_name, fn, order=0):
         from wagtail.wagtailcore import hooks
 
-        hooks.register(hook_name, fn)
+        hooks.register(hook_name, fn, order)
         try:
             yield
         finally:
-            hooks.get_hooks(hook_name).remove(fn)
+            hooks._hooks[hook_name].remove((fn, order))
 
 
 class WagtailPageTests(WagtailTestUtils, TestCase):
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -4,6 +4,7 @@ Changelog
 1.10 (xx.xx.xxxx) - IN DEVELOPMENT
 ~~~~~~~~~~~~~~~~~
 
+* Hook can now be ordered (Gagaro)
 
 1.9 (xx.xx.xxxx) - IN DEVELOPMENT
 ~~~~~~~~~~~~~~~~
--- a/wagtail/wagtailcore/hooks.py
+++ b/wagtail/wagtailcore/hooks.py
@@ -1,11 +1,13 @@
 from __future__ import absolute_import, unicode_literals
 
+from operator import itemgetter
+
 from wagtail.utils.apps import get_app_submodules
 
 _hooks = {}
 
 
-def register(hook_name, fn=None):
+def register(hook_name, fn=None, order=0):
     """
     Register hook for ``hook_name``. Can be used as a decorator::
 
@@ -23,13 +25,13 @@ def register(hook_name, fn=None):
     # Pretend to be a decorator if fn is not supplied
     if fn is None:
         def decorator(fn):
-            register(hook_name, fn)
+            register(hook_name, fn, order=order)
             return fn
         return decorator
 
     if hook_name not in _hooks:
         _hooks[hook_name] = []
-    _hooks[hook_name].append(fn)
+    _hooks[hook_name].append((fn, order))
 
 
 _searched_for_hooks = False
@@ -43,5 +45,8 @@ def search_for_hooks():
 
 
 def get_hooks(hook_name):
+    """ Return the hooks function sorted by their order. """
     search_for_hooks()
-    return _hooks.get(hook_name, [])
+    hooks = _hooks.get(hook_name, [])
+    hooks = sorted(hooks, key=itemgetter(1))
+    return [hook[0] for hook in hooks]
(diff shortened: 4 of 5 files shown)


gasman:
Hi @Gagaro - thanks for the suggestion and PR!

At the moment, the order of hooks is determined by the ordering in `INSTALLED_APPS` - we've never made that official, but I suspect people are already depending on that. Some hooks (e.g. menu items) also have ordering mechanisms of their own, so adding another one could get confusing...

More generally, I think we should try to discourage hooks that depend on the action of other hooks. The whole motivation of hooks is to improve code separation - allowing actions to happen "at a distance" without other parts of the code needing to know about them. If our hook functions need to run in a specific order because of the interactions between them, then we've failed at that goal... in that case, we should try to redesign our hooks so that this isn't necessary.

In the case of `construct_homepage_summary_items`, I guess you need it to run after the built-in Wagtail hooks because you're modifying / deleting the existing items? Perhaps we should follow the same pattern that we use for menu items, and have a `register_homepage_summary_item` hook where apps are expected to add summary items - then `construct_homepage_summary_items` becomes a second pass over the list, where modifications and deletions can happen.

Gagaro:
> At the moment, the order of hooks is determined by the ordering in INSTALLED_APPS

That's not true in my case. I have the apps registering a hook in this order:

```python
    'custom_admin',
    'wagtail.wagtaildocs',
    'wagtail.wagtailimages',
    'wagtail.wagtailcore',
````

But I get the method in this one:

```python
<function add_pages_summary_item>,
<function my_hook>,
<function add_documents_summary_item>,
<function add_images_summary_item>
```

> Some hooks (e.g. menu items) also have ordering mechanisms of their own, so adding another one could get confusing...

Indeed, I didn't realize that. But that proves that the need is there for some hooks at least.

> The whole motivation of hooks is to improve code separation - allowing actions to happen "at a distance" without other parts of the code needing to know about them. 

This could work if the hook didn't modify the data. This is what an observer pattern would do. The hooks should only read the data and not change it.

> If our hook functions need to run in a specific order because of the interactions between them, then we've failed at that goal...

Yes, for an observer pattern I would agree. But we do have code that change the data, so what we need is a pipeline, in which we can accurately determine the order in which the hooks are called. Django does that with the middlewares.

> in that case, we should try to redesign our hooks so that this isn't necessary.

IMO this is necessary in some case. For exemple with `after_create_page` or `before_serve_page`, which can stop the loop and keep some hooks from being called. We may want to have some specialized hooks being called before a more generic one. Else, only the generic one would be called because it would return a response every time.

> In the case of construct_homepage_summary_items, I guess you need it to run after the built-in Wagtail hooks because you're modifying / deleting the existing items?

That's right.

> Perhaps we should follow the same pattern that we use for menu items, and have a register_homepage_summary_item hook where apps are expected to add summary items - then construct_homepage_summary_items becomes a second pass over the list, where modifications and deletions can happen.

This seems too complicated and doesn't help with all the other hooks. We would still have the issue if we have another module calling construct_homepage_summary_items.

I think every hook which can modify the data it receives should be sortable. It is not reliable otherwise. The only case when the order doesn't matter is when the data being passed is read-only. For those, being sortable doesn't matter, so we can use the same system for both.

gasman:
> That is not true in my case. I have the apps registering a hook in this order:
> 
>     'custom_admin',
>     'wagtail.wagtaildocs',
>     'wagtail.wagtailimages',
>     'wagtail.wagtailcore',
> But I get the methods in this one:
> 
>     <function add_pages_summary_item>,
>     <function my_hook>,
>     <function add_documents_summary_item>,
>     <function add_images_summary_item>

`add_pages_summary_item` comes from wagtailadmin, not wagtailcore - do you have that higher up?

Good point about `after_create_page` / `before_serve_page`... OK, I'm happy to approve this proposal. (Even if the `order` parameter isn't my preferred solution, it's better to have a solution that works now, rather than having to wait for us to design the 'perfect' solution.)

I think we should keep the built-in hooks at `order=0`, though:

1) It preserves the (unofficial) current behaviour
2) It encourages the idea that `order` should only be used in special cases where it resolves a conflict, rather than being a decision to be made every time you define a hook
3) I think it's good for the built-in apps like wagtailimages / wagtaildocs to have as few "special privileges" as possible.

Gagaro:
> add_pages_summary_item comes from wagtailadmin, not wagtailcore - do you have that higher up?

My bad. But `wagtailadmin` is still below my module. The complete `INSTALLED_APPS` is:

```python
INSTALLED_APPS = [
    'pages',
    'comments',
    'showcase',
    'sites',
    'custom_admin',

    'wagtail.wagtailforms',
    'wagtail.wagtailredirects',
    'wagtail.wagtailembeds',
    'wagtail.wagtailsites',
    'wagtail.wagtailusers',
    'wagtail.wagtailsnippets',
    'wagtail.wagtaildocs',
    'wagtail.wagtailimages',
    'wagtail.wagtailsearch',
    'wagtail.wagtailadmin',
    'wagtail.wagtailcore',

    'wagtail.contrib.wagtailstyleguide',

    'modelcluster',
    'taggit',

    'django.contrib.admin',
    'django.contrib.auth',
    'django.contrib.contenttypes',
    'django.contrib.sessions',
    'django.contrib.messages',
    'django.contrib.staticfiles',
]
```

> I think we should keep the built-in hooks at order=0, though:

Yes I agree with your points.

Thanks for the review, I'll work on this ASAP.

Gagaro:
Doc and tests added.

wagtail/tests/utils.py
```diff
@@ -104,14 +104,14 @@ def assertLogs(self, logger=None, level=None):
...
+    def register_hook(self, hook_name, fn, order=0):
         from wagtail.wagtailcore import hooks
 
-        hooks.register(hook_name, fn)
+        hooks.register(hook_name, fn, order)
         try:
             yield
         finally:
-            hooks.get_hooks(hook_name).remove(fn)
+            hooks._hooks[hook_name].remove((fn, order))
```
Gagaro: Maybe we should add an `hooks.unregister` method?

This could also be another solution if someone doesn't want the default hooks.

docs/reference/hooks.rst
```diff
@@ -23,6 +23,21 @@ Alternatively, ``hooks.register`` can be called as an ordinary function, passing
...
+  @hooks.register('name_of_hook', order=1)  # This will run after every hook in the wagtail core
+  def my_hook_function(arg1, arg2...)
+      # your code here
+
+  @hooks.register('name_of_hook', order=-1)  # This will run before every hook in the wagtail core
+  def my_other_hook_function(arg1, arg2...)
+      # your code here
+
+  @hooks.register('name_of_hook', order=2)  # This will run after `my_hook_function`
+  def my_hook_function(arg1, arg2...)
```
Gagaro: I forgot to change the function name there, I'll fix this.

Gagaro:
I also forgot to update the changelogs.

wagtail/tests/utils.py
gasman: No, definitely not. As an app author, I define a hook for a reason that's internal to my app; I don't want other apps unilaterally deciding that my hook isn't necessary.

If there's a legitimate need to disable functionality in (say) wagtailimages, then by all means we can make that configurable _within wagtailimages_ - that way, we don't have other apps forcing configuration changes that wagtailimages wasn't designed to handle.

Gagaro:
I added the change log in the 1.10 section as the 1.9 is in RC now.

gasman:
Merged in ac7ad067d2b10f44a8dbabd0e3f7afb64345aebd - thanks!

PR #3337, opened 2017-02-05
Add browser issues/support section for editors

thibaudcolas:
This adds a new section to the editor manual, in an attempt to address #1724. Browser support needs to be documented for both developers and users, and I think in this instance it is beneficial to maintain two separate documentations for the two audiences.

At first I was keen to reuse https://en.support.wordpress.com/browser-issues/ but couldn't (it's `CC BY-SA 4.0` whereas the Wagtail documentation is Copyright Torchbox), so I made up my own text instead.

This points to http://browsehappy.com/, which is the resource Wagtail uses in the "Please upgrade your browser" capability message, as well as http://enable-javascript.com/, which is used in the "JS is required" capability message.


docs/editor_manual/browser_issues.rst
```diff
@@ -0,0 +1,11 @@
+Browser issues
+~~~~~~~~~~~~~~
+
+Some issues with Wagtail may come from the browser. Try changing or resetting its settings, or clearing the browser cache and cookies. You can also try using Wagtail with `another browser <http://browsehappy.com/>`_ to see if the problem persists.
```
jjanssen: Since we're targeting editors in this manual. Could we provide them to a link on how-to clear their cache / cookies? For eg: http://www.refreshyourcache.com/en/home/

thibaudcolas: Good point, added.

docs/editor_manual/browser_issues.rst
```diff
@@ -0,0 +1,11 @@
+Browser issues
+~~~~~~~~~~~~~~
+
+Some issues with Wagtail may come from the browser. Try changing or resetting its settings, or `clearing the browser cache and cookies <http://www.refreshyourcache.com/>`_. You can also try using Wagtail with `another browser <http://browsehappy.com/>`_ to see if the problem persists.
```
gasman: What does "Try changing or resetting its settings" mean, exactly? If we can't suggest any specific actions here, we should probably leave this out.

thibaudcolas: Good point. I think it will be hard to suggest specific actions (apart from having JS enabled, which is done below) so I'll remove this.

gasman:
Looks good to me! Merged in d6df83c6d3ed2238da2916fa9e67ffd73f067720.

PR #3338, opened 2017-02-05
Add browser support targets in contributing guidelines

thibaudcolas:
This is an attempt at addressing #1724. Documented browser support targets are important so every developer knows what they are expected to test their changes against.

I would like to have feedback on both the form of this documentation, and the specific targets it sets. This is inspired from Springload's [browser support requirements](https://github.com/springload/frontend-starter-kit/tree/master/docs#browser--device-support).

- For the form, the intent is to be vague enough about "other browsers that are supported but not on the table" (Opera, Brave, etc) yet precise enough that when people test the UI, they can feel confident they have tested it "enough". 
- For the content,
  - This is using "Latest" a bit more than I would like it to. I would expect my CMS to work fine if I opened it in a version of Chrome/Firefox/Safari/Edge(/Opera) from a year ago, that said actually testing on non-latest versions requires a significantly higher amount of tooling. I don't think it's that big of a deal, but perhaps we could be a bit more conservative by adding "Latest Firefox ESR" to the list.
  - The part I'm the least comfortable with is saying "no support" for the Android stock browser, and "Latest" for Chrome Android. I know release cycles are way slower on Android, so if there were issues there it would hurt a lot of users. That said, the amount of tooling necessary to effectively test on Android Chrome / Stock browser (and old releases of it) is significant. We could be a bit more conservative by saying "Chrome Android 4+", but I'd rather not.
  - I think it's worth having separate targets for Mobile Safari tablet and phone, because both of those get tons of usage and have very different screen sizes.

### DIFF
--- a/docs/contributing/developing.rst
+++ b/docs/contributing/developing.rst
@@ -116,6 +116,35 @@ If your Elasticsearch instance is located somewhere else, you can set the
 
     $ ELASTICSEARCH_URL=http://my-elasticsearch-instance:9200 python runtests.py --elasticsearch
 
+**Browser and device support**
+
+Wagtail is meant to be used on a wide variety of devices and browsers. Supported browser / device versions include:
+
+=============  =============  =============
+Browser        Device/OS      Version(s)
+=============  =============  =============
+Mobile Safari  iOS Phone      Latest
+Mobile Safari  iOS Tablet     Latest
+Chrome         Android        Latest
+IE             Desktop        9, 10, 11
+Chrome         Desktop        Latest
+MS Edge        Desktop        Latest
+Firefox        Desktop        Latest
+Safari         macOS          Latest
+=============  =============  =============
+
+Those environments are the ones Wagtail *is tested* on. Our development standards ensure that the site is usable on other browsers **and will work on future browsers**. To test on IE, install virtual machines `made available by Microsoft <https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/>`_.
+
+Unsupported browsers / devices include:
+
+=============  =============  =============
+Browser        Device/OS      Version(s)
+=============  =============  =============
+Stock browser  Android        All
+IE             Desktop        8 and below
+Safari         Windows        All
+=============  =============  =============
+
 Compiling static assets
 ~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/contributing/developing.rst
+++ b/docs/contributing/developing.rst
@@ -123,17 +123,18 @@ Wagtail is meant to be used on a wide variety of devices and browsers. Supported
 =============  =============  =============
 Browser        Device/OS      Version(s)
 =============  =============  =============
-Mobile Safari  iOS Phone      Latest
-Mobile Safari  iOS Tablet     Latest
-Chrome         Android        Latest
-IE             Desktop        9, 10, 11
-Chrome         Desktop        Latest
-MS Edge        Desktop        Latest
+Mobile Safari  iOS Phone      Last 2
+Mobile Safari  iOS Tablet     Last 2
+Chrome         Android        Last 2
+IE             Desktop        11
+Chrome         Desktop        Last 2
+MS Edge        Desktop        Last 2
 Firefox        Desktop        Latest
-Safari         macOS          Latest
+Firefox ESR    Desktop        Latest
+Safari         macOS          Last 2
 =============  =============  =============
 
-Those environments are the ones Wagtail *is tested* on. Our development standards ensure that the site is usable on other browsers **and will work on future browsers**. To test on IE, install virtual machines `made available by Microsoft <https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/>`_.
+We aim for Wagtail to work in those environments. Our development standards ensure that the site is usable on other browsers **and will work on future browsers**. To test on IE, install virtual machines `made available by Microsoft <https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/>`_.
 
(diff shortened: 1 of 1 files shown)


thibaudcolas:
Tagging @jjanssen for review as I think I remember us discussing this in the past.

docs/contributing/developing.rst
```diff
@@ -116,6 +116,35 @@ If your Elasticsearch instance is located somewhere else, you can set the
...
+
+Wagtail is meant to be used on a wide variety of devices and browsers. Supported browser / device versions include:
+
+=============  =============  =============
+Browser        Device/OS      Version(s)
+=============  =============  =============
+Mobile Safari  iOS Phone      Latest
+Mobile Safari  iOS Tablet     Latest
+Chrome         Android        Latest
+IE             Desktop        9, 10, 11
```
tomdyson: IE 9 was launched in 2011, and is no longer supported by Microsoft. Can we exclude it?

docs/contributing/developing.rst
```diff
@@ -116,6 +116,35 @@ If your Elasticsearch instance is located somewhere else, you can set the
...
+Chrome         Desktop        Latest
+MS Edge        Desktop        Latest
+Firefox        Desktop        Latest
+Safari         macOS          Latest
+=============  =============  =============
+
+Those environments are the ones Wagtail *is tested* on. Our development standards ensure that the site is usable on other browsers **and will work on future browsers**. To test on IE, install virtual machines `made available by Microsoft <https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/>`_.
```
tomdyson: 'is tested on' isn't strictly true. Perhaps it should be.

docs/contributing/developing.rst
jjanssen: As far as official support goes, Internet Explorer 11 is the last supported version by Microsoft. 
So i guess it is safe to say we can deprecate support for IE 9 and 10. See also: https://www.microsoft.com/en-us/windowsforbusiness/end-of-ie-support

For the past year we have not included it in our Definition of Done for a variety of clients and there were no discussions regarding it (including corporates).

docs/contributing/developing.rst
```diff
@@ -116,6 +116,35 @@ If your Elasticsearch instance is located somewhere else, you can set the
...
+=============  =============  =============
+Browser        Device/OS      Version(s)
+=============  =============  =============
+Mobile Safari  iOS Phone      Latest
+Mobile Safari  iOS Tablet     Latest
+Chrome         Android        Latest
+IE             Desktop        9, 10, 11
+Chrome         Desktop        Latest
+MS Edge        Desktop        Latest
+Firefox        Desktop        Latest
```
jjanssen: I am in favour for at least referring to the last Firefox ESR (v45)

docs/contributing/developing.rst
```diff
@@ -116,6 +116,35 @@ If your Elasticsearch instance is located somewhere else, you can set the
...
+**Browser and device support**
+
+Wagtail is meant to be used on a wide variety of devices and browsers. Supported browser / device versions include:
+
+=============  =============  =============
+Browser        Device/OS      Version(s)
+=============  =============  =============
+Mobile Safari  iOS Phone      Latest
+Mobile Safari  iOS Tablet     Latest
+Chrome         Android        Latest
```
jjanssen: Maybe we should refer to the last 2 versions instead of latest (for all)? 
That way we're are not really attached to the absolute latest but we also for eg. include iOS 9, which together with iOS 10 holds up to 94% (https://developer.apple.com/support/app-store/)

jjanssen:
Regarding the Android support. Previous to 4.4 it was dependant on OS updates whereas these days even the webview can be upgraded apart from OS updates. I would suggest to at least refer to Chrome usage. Vendors like Samsung or LG still have some funky stuff attached to the stock browser which can occasionally be unpredictable and can cost a lot of debugging time.

docs/contributing/developing.rst
thibaudcolas: I agree on principle, but realistically I personally only have access to one test iPhone and one test iPad. I think that's something to be discussed further – whether it's ok to have a difference between support targets and what you actually test things on.

docs/contributing/developing.rst
thibaudcolas: Same for Springload – new projects now start at IE11. I think it would be worth ditching IE10 at the same time.

docs/contributing/developing.rst
thibaudcolas: I think we should come up with clear testing guidelines here, balance our wish to support as many people as possible with the diminishing returns of browser support. 

Personally I don't think there is much of a point in stating support if there are no tests, but that's a lot of tests to ask for external contributors. Perhaps we could leverage something like [Sauce Labs](https://saucelabs.com/) or [BrowserStack](https://www.browserstack.com/) (both being free for open-source). How has Wagtail handled browser tests / browser support so far?

jjanssen: Maybe for now we could say something like: "we aim for Wagtail to work on those environments"? 

For the future purpose I think we could benefit from integrating with SauceLabs or BrowserStack. But that would probably also require a sandbox environment to properly test the Wagtail admin.

thibaudcolas:
I've made changes according to everyone's feedback, I think this is good to go. We should revisit when we get better tooling to help with those support targets (BrowserStack live & automate).

tomdyson:
LGTM. Thanks @thibaudcolas and @jjanssen.

gasman:
Merged in 1e3548e67156c5bea1c13b24771bb50bcfe140cf

Issue #2866, opened 2016-07-25
RoutablePageMixin + i18n + Preview = not working


soar:
Today I've found strange issue. I have some page class:

```python
class BlogPage(Page):
    ...
```

And `urls.py`:

```python
urlpatterns = [
    url(r'^djangoadmin/', admin.site.urls),
    url(r'^wagtailadmin/', include(wagtailadmin_urls)),
    url(r'^doc/', include(wagtaildocs_urls)),
]

urlpatterns += i18n_patterns(
    url(r'', include(wagtail_urls)),
)
```

When I click "Preview" button in admin page, I see following requests:

```
[25/Jul/2016 13:29:19] "GET /wagtailadmin/pages/preview/ HTTP/1.1" 200 10327
[25/Jul/2016 13:29:20] "GET /wagtailadmin/pages/preview_loading/ HTTP/1.1" 200 9844
[25/Jul/2016 13:29:23] "POST /wagtailadmin/pages/19/edit/preview/ HTTP/1.1" 200 9594
```

And all works fine. But now I want to modify my page class:

```python
class BlogPage(RoutablePageMixin, Page):

    @route(r'^tag/(\w+)/$', name='tag')
    def tag(self, request, tag):
    ...
```

Wagtail admin is still working. But when I open same page and click "Preview" - I see 404 page:

```
Page not found (404)
Request Method: 	GET
Request URL: 	http://127.0.0.1:8000/ru/wagtailadmin/pages/19/edit/preview/
Raised by: 	wagtail.wagtailcore.views.serve
```

You can notice, that language tag appears in URI, but it shouldn't - i18n isn't enabled for admin pages. And now there are following requests:

```
[25/Jul/2016 13:35:52] "GET /wagtailadmin/pages/19/edit/ HTTP/1.1" 200 173884
[25/Jul/2016 13:35:58] "GET /wagtailadmin/pages/preview/ HTTP/1.1" 200 10326
[25/Jul/2016 13:35:59] "GET /wagtailadmin/pages/preview_loading/ HTTP/1.1" 200 9844
Not Found: /wagtailadmin/pages/19/edit/preview/
[2016-07-25 13:36:01.700] [WARNING   ] [django.request] [base/get_response] [p:320956|t:311696] || Not Found: /wagtailadmin/pages/19/edit/preview/
[25/Jul/2016 13:36:01] "POST /wagtailadmin/pages/19/edit/preview/ HTTP/1.1" 302 0
[...]

PR #3344, opened 2017-02-07
added default index_route for RoutablePageMixin

nutztherookie:
RoutablePageMixin has a default index_route method that is decorated
with `@route(r'^$')`. This way, including RoutablePageMixin doesn't
force you to re-enable the default functionality you would expect from a
Page anyways. `index_route` behaves exactly like a standard Wagtail Page.
To override the default behaviour, one can simply override the route by
decorating another function with r'^$'.

(as disussed in issue #2866)


nutztherookie:
I forgot to mention: I had to `reversed()` the sorted list of routes, otherwise the base-class route would have had predominance

wagtail/contrib/wagtailroutablepage/models.py
```diff
@@ -45,7 +56,7 @@ def get_subpage_urls(cls):
 
         return tuple([
             route[0]
-            for route in sorted(routes, key=lambda route: route[1])
+            for route in reversed(sorted(routes, key=lambda route: route[1]))
```
gasman: Reversing the order breaks the current behaviour where routes defined earlier in a class will match first, e.g.:

    class RoutablePageTest(RoutablePage):
        @route(r'^archive/year/1984/$')
        def archive_for_1984(self, request):
            return HttpResponse("we were always at war with eastasia")

        @route(r'^archive/year/(\d+)/$')
        def archive_by_year(self, request, year):
            return HttpResponse("ARCHIVE BY YEAR: " + str(year))

This behaviour isn't covered by tests (or documented, I suspect...), but the `_creation_counter` logic above was designed to support this, so I think we want to preserve this behaviour. Looking at a fix now...

gasman:
Fix for https://github.com/wagtail/wagtail/pull/3344#discussion_r109198840 added in 2918735684cbbbc1a71a82fe79bfe629ebde32e0 - @kaedroho, please can you review? (Just that one commit needs reviewing - rest of the PR looks good to me!)

Issue #3346, opened 2017-02-08
1.9rc1 regression in Block.render()


mvantellingen:
The ``Block.render()`` (see https://github.com/wagtail/wagtail/blob/827149f5d21620f8bbbc9b6517e7cc686c52b487/wagtail/wagtailcore/blocks/base.py#L229) accepts an optional context however a few lines below (https://github.com/wagtail/wagtail/blob/827149f5d21620f8bbbc9b6517e7cc686c52b487/wagtail/wagtailcore/blocks/base.py#L250) it is assumed that it is a dictionary.

We hit this issue since we implemented a ``get_context()`` override on a subclass of a ``blocks.StructBlock``. Adding the parent_context kwarg and passing it down works the issue.

cc @mikedebock

gasman:
Fixed in #3347

PR #3347, opened 2017-02-08
Fix a regression when overriding Block.get_context()

mvantellingen:
If ``Block.get_context()`` was overriden without accepting the
``parent_context`` kwarg the fallback handling threw an exception if
``context`` was None.

See #3346


gasman:
Looks good to me, thanks! Merged in 23890a094a9a59437dd9aa7ae73ec391e030e4f7 (master) / 3c8b43f09a981c4948dc0b1103a2db0849391685 (1.9.x).

mikedebock:
👍

Issue #3234, opened 2016-12-20
modeladmin, links requires image and documents modules


danni:
A number of pieces of functionality, including linking and the modeladmin module, require installing the images and documents modules, which you might not want, depending on your use case.

PR #3349, opened 2017-02-08
Reduce modeladmin's dependency on wagtailimages and wagtaildocs

### LINKED ISSUE #3263: ModelAdmin InspectView raises SourceImageIOError
### Issue Summary

When enabling InspectView for a ModelAdmin the InspectView raises the above error when the object inspected contains an image.

I guess this is because Dynamic image serve view is required.

### Steps to Reproduce

Create a custom model containing a fk to wagtailimage.Image:

class MyModel(models.Model):
    image = models.ForeignKey('wagtailimages.Image')

Create a custom ModelAdmin

class MyModelAdmin(ModelAdmin):
	model = MyModel
	inspect_view_enabled = True

Register it:
modeladmin_register(MyModelAdmin)

Now create an instance using Wagtail.
After submitting klik on the inspect button.

### Technical details

* Python Python 3.5.0
* Django 1.10.1
* Wagtail 1.8

ababic:
- Remove imports at the top of `options.py` and `views.py` that result in ImportError being thrown when those apps aren’t installed (Should help with #3234, but not fully resolve)
- Alter ThumbnailMixin and InspectView to use `wagtail.wagtailimages.shortcuts.get_rendition_or_not_found()` to render images, which handles missing image source files gracefully, and reduces code duplication (Fixes #3263).
- Simplify `InspectView.get_field_display_value()` by not limiting image and document rendering to ForeignKey fields. It should work consistently for property methods or other attributes too.
- Provide a more useful error message when attempting to use ThumbnailMixin without `wagtail.wagtailimages` installed
- Updated docs to reflect changes

* Do the tests still pass? (http://docs.wagtail.io/en/latest/contributing/developing.html#testing) - Yep
* Does the code comply with the style guide? - It didn't, but it should now
* For Python changes: Have you added tests to cover the new/fixed behaviour? - It's more a rejigging of existing behaviour than new behaviour, so the same tests should do. Is it possible to run tests with different sets of INSTALLED_APPS or missing image source files?
* For new features: Has the documentation been updated accordingly? - Yep

### DIFF
--- a/wagtail/contrib/modeladmin/views.py
+++ b/wagtail/contrib/modeladmin/views.py
@@ -10,11 +10,9 @@ from django.contrib.admin import FieldListFilter, widgets
 from django.contrib.admin.exceptions import DisallowedModelAdminLookup
 from django.contrib.admin.options import IncorrectLookupParameters
 from django.contrib.admin.utils import (
-    get_fields_from_path, lookup_needs_distinct, prepare_lookup_value, quote,
-    unquote)
+    get_fields_from_path, lookup_needs_distinct, prepare_lookup_value, quote, unquote)
 from django.contrib.auth.decorators import login_required
-from django.core.exceptions import (
-    ImproperlyConfigured, PermissionDenied, SuspiciousOperation)
+from django.core.exceptions import ImproperlyConfigured, PermissionDenied, SuspiciousOperation
 from django.core.paginator import InvalidPage, Paginator
 from django.db import models
 from django.db.models.constants import LOOKUP_SEP
--- a/wagtail/contrib/modeladmin/views.py
+++ b/wagtail/contrib/modeladmin/views.py
@@ -866,7 +866,7 @@ class InspectView(InstanceSpecificView):
             if isinstance(val, AbstractImage):
                 # Render a rendition of the image
                 return self.get_image_field_display(field_name, field)
-        except ImportError:
+        except RuntimeError:
             pass
 
         # wagtail.wagtaildocuments might not be installed
@@ -875,7 +875,7 @@ class InspectView(InstanceSpecificView):
             if isinstance(val, AbstractDocument):
                 # Render a link to the document
                 return self.get_document_field_display(field_name, field)
-        except ImportError:
+        except RuntimeError:
             pass
 
         # Resort to returning the real value or 'empty value'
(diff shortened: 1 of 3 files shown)


ababic:
Hi @mikedingjan. Please let me know if you have any questions / requests.

wagtail/contrib/modeladmin/views.py
```diff
@@ -859,39 +856,37 @@ def get_field_display_value(self, field_name, field=None):
...
-
-            except AttributeError:
-                pass
+        # Now let's get the attribute value from the instance itself and see if
+        # we can render something useful. raises AttributeError appropriately.
+        val = getattr(self.instance, field_name)
+
+        # wagtail.wagtailimages might not be installed
+        try:
+            from wagtail.wagtailimages.models import AbstractImage
```
mikedingjan: Doesn't this give a ``RuntimeError`` since wagtailimages is installed within the wagtail package, so it can be imported but can't load the model since it's not installed in django.

```console
>>> try:
...     from wagtail.wagtailimages.models import AbstractImage
... except ImportError:
...     print('import error happened')
...
  File "<console>", line 2, in <module>
  File "/Users/m.dingjan/.pyenv/versions/3.5.3/lib/python3.5/site-packages/wagtail/wagtailimages/models.py", line 337, in <module>
    class Image(AbstractImage):
  File "~/.pyenv/versions/3.5.3/lib/python3.5/site-packages/django/db/models/base.py", line 113, in __new__
    "INSTALLED_APPS." % (module, name)
RuntimeError: Model class wagtail.wagtailimages.models.Image doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.```

ababic: You're right, thank you. I will get that updated.

ababic:
Hey @mikedingjan. I've implemented the change you requested, but there seems to be some issues with the drone test. Would you be able to restart that test for me?

gasman:
Hi @ababic - I think that's a current error with the drone config, possibly as a side effect of #3411. Will investigate tomorrow...

ababic:
@gasman. Okay, no worries. thanks for letting me know.

gasman:
Merged in 3348cd5c89a9ed28ef9ea719e7a2bf818ec9c816 - thanks!

(The Drone issue should be fixed for any future PRs that are based on a version of master later than b69bf0e9bb5c1f8dceb2a12e03f809c10a5b21ae. Not sure why the problem's occurring - when we merged #3386 it appeared that the fix to .drone.yml was only needed with the newer version of pyenchant, but now the older version - which was clearly working before - is evidently in need of the same fix...)

PR #3350, opened 2017-02-08
[WIP] reduce the number of cache round trips caused by Site.get_site_root_paths()

tobiasmcnulty:
The wagtaildemo site, which is a not a large site, generates 39 identical cache calls for each load of the home page, after the cache has been primed. This isn't such a huge deal if the site isn't busy and your cache is running locally, but if it's sitting in a different EC2 availability zone all those unnecessary cache calls do add up.

The excess cache calls are due to the way `Site.get_site_root_paths()` is implemented. Another approach would be to find an object local to the request on which the result of this method could be cached. This PR attempts that, and reduces the number of cache calls by roughly half (to 20) by making sure that the same page instance only ever generates one call to `Site.get_site_root_paths()`.

There's almost certainly a better place to do this (really we should only have to call `Site.get_site_root_paths()` once per request). Any feedback/comments/questions/concerns much appreciated!

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -415,6 +415,15 @@ class Page(six.with_metaclass(PageBase, AbstractPage, index.Indexed, Clusterable
     def __str__(self):
         return self.title
 
+    @cached_property
+    def cached_site_root_paths(self):
+        """
+        Cache the result of this method locally on a request-specific class
+        instance (the Page) to reduce the number of cache round trips generated
+        by the underlying method.
+        """
+        return Site.get_site_root_paths()
+
     def set_url_path(self, parent):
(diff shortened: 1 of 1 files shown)


gasman:
Hi @tobiasmcnulty - there's an existing PR to deal with this awaiting review at https://github.com/wagtail/wagtail/pull/3188. Your thoughts on that PR would be very welcome!

tobiasmcnulty:
@gasman Oh wonderful! I'll try it out. Whenever you have a free moment today I'd appreciate a quick walk through on it too

tobiasmcnulty:
Discussed with @gasman at Wagtail sprint in Iceland and we are going ahead with a modified version of #3188

PR #3351, opened 2017-02-09
Remove deprecated features for 1.10

### LINKED ISSUE #2881: Remove wagtailimages.Filter as a model (second try)

gasman:
Also fixes #2881 (finally!)


tomdyson:
LGTM

gasman:
Merged in cdf4ae7528ad7ce0f72b227c2a50b52799078c65

PR #3353, opened 2017-02-09
Use minified versions of jQuery and jQuery UI

tomdyson:
Saves 202.87KB on uncached, uncompressed requests of all admin pages.

This is step 1 in addressing jQuery UI bloat. Steps 2 and 3 are:

- create a custom-built version of jQuery UI which only includes the components we need
- replace jQuery UI components with plain JavaScript alternatives


tomdyson:
Now includes a minified version of jQuery, saving an additional 169 KB. Total savings 371 KB.

thibaudcolas:
Merged in 3ec615e22164d438530d457c5a0422841308fafd 🎉 

PS: here is the impact. Take the absolute numbers with a grain of salt (well, more like a handful), but that's definitely a significant reduction :)

minified-jquery

PR #3355, opened 2017-02-09
Defer specific fields in explorer 

kaedroho:
This pull request improves memory usage and (slightly) improves performance of the Explorer view.

We're currently fetching all specific fields of the pages that are being displayed in the explorer. This is wasteful because these fields are usually not used and can be quite large. On a site, we have been getting OOM errors because Wagtail pulling in all the field data (and the pages are quite large).

We're pulling in these fields because the URL generation logic can be customised and we want to display the correct URL in the admin. This use-case is quite rare though and users doing this will suffer a (minor) performance hit by this change (see benchmark below).

This also introduces a very rudimentary benchmarking tool, based on Django tests. This measures the time and memory usage of a view and is useful to testing performance improvements like this.

Benchmark:

BEFORE

```
test (wagtail.wagtailadmin.tests.benches.BenchPageExplorerWith50LargePages)
time min: 0.9661133289337158 max: 1.2755255699157715 avg: 1.0499631881713867
memory min: 44227595 max: 44694417 avg: 44365930.1
```

```
test (wagtail.wagtailadmin.tests.benches.BenchPageExplorerWithCustomURLPages)
time min: 0.7922754287719727 max: 0.851860523223877 avg: 0.8272997617721558
memory min: 13315100 max: 13570529 avg: 13380374.5
```

AFTER

```
test (wagtail.wagtailadmin.tests.benches.BenchPageExplorerWith50LargePages)
time min: 0.8060946464538574 max: 1.2400388717651367 avg: 0.8822282314300537
memory min: 13080452 max: 13547158 avg: 13236624.9
```

```
test (wagtail.wagtailadmin.tests.benches.BenchPageExplorerWithCustomURLPages)
time min: 0.8017115592956543 max: 0.9074203968048096 avg: 0.8303963422775269
memory min: 13276127 max: 13520267 avg: 13348108.5
```

In summary:

 - When displaying 50 "large" pages (600KB of streamfield content) the memory usage is quartered and performance is ~10% faster (200ms saving).
 - When displaying 50 pages with customised URLs 10ms is added on the request time, but memory usage is a tiny bit improved.

### DIFF
--- a/wagtail/wagtailcore/query.py
+++ b/wagtail/wagtailcore/query.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, unicode_literals
 
+import functools
 import posixpath
 from collections import defaultdict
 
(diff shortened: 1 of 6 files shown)


wagtail/wagtailcore/query.py
```diff
@@ -394,8 +415,15 @@ class SpecificIterable(BaseIterable):
...
+            return specific_iterator(self.queryset, defer=True)
+
 else:
     from django.db.models.query import QuerySet
 
     class SpecificQuerySet(QuerySet):
         iterator = specific_iterator
+
+    class DeferredSpecificQuerySet(QuerySet):
+        iterator = functools.partial(specific_iterator, defer=True)
```
m1kola: It looks like a value for the `qs` argument is missed here. I think this is the reason of test failures like [this](https://travis-ci.org/wagtail/wagtail/jobs/200050157#L438) on Django 1.8:

```
TypeError: specific_iterator() missing 1 required positional argument: 'qs'
```

m1kola: Ah, with `functools.partial` it should work. Ok, then I do not understand the reason of test failures...

gasman:
Looks like this ready for reconsideration again, now that we've dropped Django 1.8 support...

gasman:
Merged in 78e931fb21c34c1fd50c4706efe5dd240ad4374d

PR #3356, opened 2017-02-09
Move signal_handlers to separate files

mikedingjan:
In preparation for implementing a swappable page register the signal handlers in the ``django.apps.AppConf.ready`` method.


wagtail/wagtailcore/__init__.py
```diff
@@ -3,6 +3,8 @@
 # Imported for historical reasons
 from wagtail import __semver__, __version__  # noqa
 
+default_app_config = 'wagtail.wagtailcore.apps.WagtailCoreAppConfig'
```
kaedroho: Looks like this was removed in https://github.com/wagtail/wagtail/commit/384d95d2b5937b3db63ee0d82892c1d18024f96a (I assume this was an accident)

PR #3357, opened 2017-02-09
Add notification user to notification mail context

Vorlif:
This is not really a big change. I have only added the user that send the notification mail to the mail context. Furthermore, I have rename the `exclude_user_id` to `notification_user_id` because this is the user that start the notification action.

This makes it possible for developers to add the notification_user in the notification mail.

Like this:

> The page "{{ title }}" has been approved by {{ notification_user.username }}.

### Infos
No, I think this is not relevant and I haven't found tests for this function.

I am open for improvement suggestions. :)

### DIFF
--- a/wagtail/wagtailadmin/utils.py
+++ b/wagtail/wagtailadmin/utils.py
@@ -174,6 +174,8 @@ def send_mail(subject, message, recipient_list, from_email=None, **kwargs):
 def send_notification(page_revision_id, notification, notification_user_id):
     # Get revision
     revision = PageRevision.objects.get(id=page_revision_id)
+    # User
+    User = get_user_model()
 
     # Get list of recipients
     if notification == 'submitted':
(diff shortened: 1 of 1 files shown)


wagtail/wagtailadmin/utils.py
```diff
@@ -171,9 +171,11 @@ def send_mail(subject, message, recipient_list, from_email=None, **kwargs):
     return django_send_mail(subject, message, from_email, recipient_list, **kwargs)
 
 
-def send_notification(page_revision_id, notification, excluded_user_id):
+def send_notification(page_revision_id, notification, notification_user_id):
     # Get revision
     revision = PageRevision.objects.get(id=page_revision_id)
+    # User
+    User = get_user_model()
```
BertrandBordage: Why not use directly `get_user_model()` directly where it’s useful, during context definition?

PR #3358, opened 2017-02-10
Cache sites for request [WIP]

tomdyson:
Cache a relatively expensive query, called by `SiteMiddleware` on every Wagtail page request.

Like `wagtail_site_root_paths`, the cache is reset whenever Site records are updated.

Needs tests. If this approach is approved, I'll write them.

- [x] do the caching
- [x] test it works
- [ ] write tests


gasman:
Superseded by #4142

PR #3361, opened 2017-02-10
Custom DRF serializers in API

kaedroho:
This pull request makes it possible for users to define their own DRF serializers per page type. This should massively improve flexibility over the current list of field names in ``api_fields``.

This also introduces an ``ImageRenditionField`` field serializer which makes it easy to add image renditions to the API!

For example:
```python

from wagtail.api.v2.serializers import PageSerializer
from wagtail.wagtailimages.api.fields import ImageRenditionField


class BlogPageSerializer(PageSerializer):
    feed_image_thumbnail = ImageRenditionField('fill-300x300', source='feed_image')

    class Meta(PageSerializer.Meta):
        fields = PageSerializer.Meta.fields + [
            'body',
            'feed_image',
            'feed_image_thumbnail',
        ]


class BlogPage(Page):
    body = models.RichTextField()
    feed_image = models.ForeignKey('wagtailimages.Image')
    ...

    # Instead of api_fields, use a serializer class
    serializer_class = BlogPageSerializer
```

This will produce the following output (when all fields are requested):

```json
{
    "id": 10,
    "meta": {
        "type": "demo.BlogPage",
        "detail_url": "/api/v2/pages/10/"
    },
    "title": "A blog post",
    "body": "body text here...",
    "feed_image": {
        "id": 5,
        "meta": {
            "type": "wagtailimages.Image",
            "detail_url": "/api/v2/images/5/"
        },
        "title": "An image",
        "width": 1000,
        "height": 500
    },
    "feed_image_thumbnail": {
        "url": "/media/images/an-image.fill-300x300.jpg",
        "width": 300,
        "height": 300
    }
}
```

The ``api_fields`` attribute is still fully supported for now but this may introduce a backwards-incompatbility for people who have manually overridden the endpoints classes. Previously, the API fields were managed by the endpoint but they are now managed by the serializer.

TODO:
 - [ ] Serializer based configuration for inline objects
 - [ ] Tests for serializer based configuration
 - [ ] Docs for serializer based configuration
 - [ ] Tests for ``ImageRenditionField``
 - [ ] Docs for ``ImageRenditionField``
 - [ ] Write instructions for migrating from custom endpoint classes for the upgrade considerations

### DIFF
--- a/wagtail/api/v2/endpoints.py
+++ b/wagtail/api/v2/endpoints.py
@@ -34,7 +34,8 @@ class BaseAPIEndpoint(GenericViewSet):
         renderer_classes.append(BrowsableAPIRenderer)
 
     pagination_class = WagtailPagination
-    base_serializer_class = BaseSerializer
+    base_serializer_class = None
+    default_base_serializer_class = BaseSerializer
     filter_backends = []
     model = None  # Set on subclass
 
@@ -170,6 +171,15 @@ class BaseAPIEndpoint(GenericViewSet):
             raise BadRequestError("query parameter is not an operation or a recognised field: %s" % ', '.join(sorted(unknown_parameters)))
 
     @classmethod
+    def get_base_serializer_class(cls, model):
+        if cls.base_serializer_class is not None:
+            return cls.base_serializer_class
+        elif hasattr(model, 'serializer_class'):
+            return model.serializer_class
+        else:
+            return cls.default_base_serializer_class
+
+    @classmethod
     def _get_serializer_class(cls, router, model, fields_config, show_details=False, nested=False):
         # Get all available fields
         body_fields = cls.get_body_fields(model)
@@ -257,7 +267,7 @@ class BaseAPIEndpoint(GenericViewSet):
         # Reorder fields so it matches the order of all_fields
         fields = [field for field in all_fields if field in fields]
 
-        return get_serializer_class(model, fields, meta_fields=meta_fields, child_serializer_classes=child_serializer_classes, base=cls.base_serializer_class)
+        return get_serializer_class(model, fields, meta_fields=meta_fields, child_serializer_classes=child_serializer_classes, base=cls.get_base_serializer_class(model))
 
     def get_serializer_class(self):
         request = self.request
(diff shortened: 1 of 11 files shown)


tomdyson:
Epic!

We should include specific instructions for people who have manually overridden endpoints classes, in the release notes.

kaedroho:
Will do! (added to the TODO list)

garykrige:
This looks great, would love to see this soon.

VincentLoy:
Can't wait to use it !! :+1:

kaedroho:
I've just been trying to get DRF serializers for related objects working in a [separate branch](https://github.com/kaedroho/wagtail/tree/feature/images-in-api-nolegacy). So far, i've been looking to get it work without any backwards compatibility for ``api_fields`` which simplifies things a bit.

This is still quite complicated, and I'm unsure whether it would be possible to do while maintaining support for ``api_fields``. This would be a really big change in how the Wagtail API works

For each related object that's nested in the response, we need to know its serialiser so we can render it. Wagtail currently builds serializers on the fly, but we're now switching to a system where they are pre-defined. This causes problems for the ``?fields`` parameter (which DRF doesn't have) which needs to be able to change the field configuration on the fly.

There's a couple of other things including how we handle ``detail_url`` and serializers without associated models (Wagtail doesn't currently require that serializers have a model specified) that would need to be addressed as well (otherwise, serializer configs would have to contain a lot of [this](https://github.com/kaedroho/wagtail/blob/2ff25d78b0b438b8fe813793d0fed34d43368be4/wagtail/tests/demosite/serializers.py#L85-L94)).

So we're going to have to overhaul a lot of code in order to allow users to fully customise their DRF serialisers. But I think we can implement something much simpler in the mean time. How does this look?

```python
from wagtail.wagtailimages.api.fields import ImageRenditionField

class BlogPage(Page):
    body = models.RichTextField()
    feed_image = models.ForeignKey('wagtailimages.Image')
    ...

    api_fields = [
        'body',
        'feed_image',

        # Field with custom serialiser
        ('feed_image_thumbnail', ImageRenditionField('fill-300x300', source='feed_image')),
    ]
```

BertrandBordage:
#3462 superseded this pull request.

PR #3363, opened 2017-02-10
Change image signals

mikedingjan:
Signals for custom Image and Rendition models are connected automatically now.


gasman:
Thanks @mikedingjan - looks good to me! Merged in 8c970e4848b9cedd5251082c1a64dfcc907e1afa.

I've also removed (in 25c2be10c3486fc3228251db7211df5286931c14) the docs about setting up the signal handler for image feature detection, as I believe these are now obsolete too.

PR #3366, opened 2017-02-14
Add convenience function for testing for any existing user page permi…

jeffreyckchau:
…ssion.

Continuation of multitenant work. 

For some features, we've found it useful to be able to exclude pages from a user that the user has no actions on whatsoever (the implicit concept of 'explorability'). However, this is difficult to achieve in a template context without the code getting bloated. By adding this method, we can reduce it to one line: `{% if page_perms.can_perform_action %}`. 

I think this will be useful for other developers who may be trying to filter admin features based on user page permissions.

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -1794,6 +1794,18 @@ class PagePermissionTester(object):
             # no publishing required, so the already-tested 'add' permission is sufficient
             return True
 
+    def can_perform_action(self):
+        """
+        Convenience method for checking whether a user has any action on the page at all.
+        This is useful for shaving down systems that would otherwise need to check each of
+        these permissions explicitly like in Django templates.
+        """
+        return can_add_subpage() or \
+               can_edit() or \
+               can_delete() or \
+               can_unpublish() or \
+               can_publish()
+
 
 class PageViewRestriction(models.Model):
     NONE = 'none'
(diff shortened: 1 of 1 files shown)


gasman:
Hi @jeffreyckchau,

I'm slightly bothered by the semantics of this - I think we're going to find that different contexts and different areas of the admin will have different ideas about what qualifies as 'performing an action'. As part of the multitenancy work, we'll soon be adding a 'can choose' permission type to determine which pages are shown in page chooser popups - will this be counted as one of the actions in `can_perform_action`?

Or is it the intention that `can_perform_action` should correspond to the actions available through the wagtailadmin page explorer? If so, it probably ought to exist in wagtailadmin in some form (perhaps as a helper function that takes a Page queryset and filters it to the explorable ones? Looping over pages and testing 'can perform action X' on each one is something we ought to avoid for performance reasons...) and have a suitably explorer-related name.

Perhaps it would help to have a concrete example of a problem you're hoping to solve via this method?

jeffreyckchau:
Hi @gasman,

I had failed to consider the 'choosable' permission as one of the actions going forward, so you are right: this may not be the best approach. 

Our org has a feature we've developed for our installation that includes a drag and droppable site map which allows an admin to view all pages explorable to their user in one location. For entries in this sitemap, if the user is allowed to perform actions to the page (edit, delete, etc.), they are listed underneath upon mouseover. We do this by checking for the existence of these actions in a template before listing them. This is where the page permissions accessor would come in handy. 

You are correct that `can_perform_action` should correlate directly with what actions are available to the wagtailadmin page explorer, as our sitemap feature is essentially a stand-in for the explorer. I like the idea of using a filtered Page queryset and returning the actions available to each one, so I may go ahead and try to implement that.

gasman:
Closing for now, as we need to find an alternative approach that preserves separation between wagtailcore and the wagtailadmin-specific concepts of "performing an action".

PR #3367, opened 2017-02-14
Fix: block.super should be outside script tags in type_index.html

lb-:
When using `extra_js` on extending wagtailadmin/base.html all other instances of block.super are located outside script tags, except this one.


gasman:
Thanks @lb-! Merged in 5aa8e00d77d20060f75b7ca65e28ef502c54245c.

PR #3354, opened 2017-02-09
Cache `Site.get_site_root_paths` in the request scope (v2)

tobiasmcnulty:
Fixes #2916
RFC: [#14](https://github.com/tobiasmcnulty/rfcs/blob/master/draft/0014-page-url-api.rst)

This is a rebased and updated version of #3188 by @gasman that moves the business logic for caching `site_root_paths` inside of the `Page` model. This makes it easier to use the API from outside Wagtail itself (the most common use case for this is most likely any navigation menu-related template tags, e.g., https://github.com/wagtail/wagtaildemo/pull/110/files).

- [x] Add/update usage documentation/recommendations for using and overriding URL-related methods on the `Page` model, if this is the route we want to go
- [x] Determine whether to deprecate `Page.url` and `Page.full_url` or defer that decision till a later date. While convenient, using them circumvents the possibility of caching `site_root_paths` and is probably a habit worth disincentivizing. (DECISION: don't deprecate)
- [x] Determine if passing `request` to these methods is appropriate (seems to be okay, other methods on `Page` already accept a `request`)

See also: #3350

Issue #3352, opened 2017-02-09
Exception if I compare revisions which include PageChooserBlock() and pageurl


Vorlif:
### Issue Summary

I'm not sure if this is a bug or if I've done something wrong.

If I want to compare two revisions of a page that has a StreamField with PageChooserBlock and a template which includes a  pageurl, I get alway an exception.

Maybe the issue dependt on #3340.

### Steps to Reproduce

1. You can use the example below (wagtail start mysite)
2. Create three pages
3. Add a StreamField 'link' from page one to page two
4. Save Draft
5. Change the link from page two to page three
6. Save Draft
7. Try to comapre the revisions 

#### Minimal example

```python
# models.py
from wagtail.wagtailcore.models import Page
from wagtail.wagtailcore import blocks
from wagtail.wagtailcore.fields import StreamField
from wagtail.wagtailadmin.edit_handlers import StreamFieldPanel

class LinkBlock(blocks.StructBlock):
    link = blocks.PageChooserBlock()

    class Meta:
        template = 'home/link.html'


class HomePage(Page):
    body = StreamField([
        ('link', LinkBlock())
    ],
        blank=True
    )

    content_panels = Page.content_panels + [
        StreamFieldPanel('body'),
    ]
```

```jinja
{# home_page.html #}

{% extends "base.html" %}
{% load wagtailcore_tags %}

{% block content %}

    {% for block in page.body %}
        <p>{% include_block block %}</p>
    {% endfor %}

{% endblock %}
```

```jinja
{# link.html #}
{% load wagtailcore_tags %}

<a href="{% pageurl value.link %}">{{ value.link.title }}</a>
```

#### Traceback 

```bash
Error during template rendering

In template /path/to/mysite/home/templates/home/link.html, error at line 3
1	{% load wagtailcore_tags %}
2	
3	<a href="{% pageurl value.link %}">{{ value.link.title }}</a>
```

```bash
Internal Server Error: /admin/pages/3/revisions/compare/1...3/
Traceback (most recent call last):
  File "/path/to/env/lib/python3.5/site-packages/django/core/handlers/exception.py", line 39, in inner
    response = get_response(request)
  File "/path/to/env/lib/python3.5/site-packages/django/core/handlers/base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/path/to/env/lib/python3.5/site-packages/django/core/handlers/base.py", line 185, in _get_response
[...]

Issue #3279, opened 2017-01-15
Preview mode with WSGI


soar:
### Issue Summary

When I've used Wagtail on localhost in development mode - all works fine. But when I moved it to production with uWSGI server - Preview mode is broken now.

### Steps to Reproduce

I used new-style middleware classes:

```python
MIDDLEWARE = [
    'django.middleware.security.SecurityMiddleware',
    'django.contrib.sessions.middleware.SessionMiddleware',
    'django.middleware.locale.LocaleMiddleware',
    'django.middleware.common.CommonMiddleware',
    'django.middleware.csrf.CsrfViewMiddleware',
    'django.contrib.auth.middleware.AuthenticationMiddleware',
    'django.contrib.auth.middleware.SessionAuthenticationMiddleware',
    'django.contrib.messages.middleware.MessageMiddleware',
    'django.middleware.clickjacking.XFrameOptionsMiddleware',
    'wagtail.wagtailcore.middleware.SiteMiddleware',
    'wagtail.wagtailredirects.middleware.RedirectMiddleware',
]
```

My server setup is: nginx -> uwsgi -> django wsgi application. All pages works fine (including admin panel). But when I press "Preview" button in editor, I get exception:

```
[2017-01-15 16:02:37.411] [ERROR     ] [django.request] [exception/handle_uncaught_exception] [p:21|t:140140641003264] || Internal Server Error: /wagtail/pages/15/edit/preview/
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py", line 39, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 217, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py", line 215, in _get_response
    response = response.render()
  File "/usr/local/lib/python3.6/site-packages/django/template/response.py", line 109, in render
    self.content = self.rendered_content
  File "/usr/local/lib/python3.6/site-packages/django/template/response.py", line 86, in rendered_content
    content = template.render(context, self._request)
  File "/usr/local/lib/python3.6/site-packages/django/template/backends/django.py", line 66, in render
    return self.template.render(context)
  File "/usr/local/lib/python3.6/site-packages/django/template/base.py", line 208, in render
    return self._render(context)
[...]

PR #3369, opened 2017-02-14
Allow pageurl / slugurl tags to function when request.site is not available

gasman:
Allow the `{% pageurl %}` and `{% slugurl %}` tags to function when the template context does not contain a `request` object (e.g. rendering a template during `str(page.body)` on a StreamField), or contains a `request` object without a `site` attribute (which can happen during the rendering of a 500 page if the error occurred before/during Wagtail's SiteMiddleware, as in #3279).

The `str(page.body)` situation is the root cause of #3352, because the 'compare revisions' view compares StreamFields based on their string representations.

(Note: @tobiasmcnulty already fixed this in #3354, but #3352 is a release blocker for 1.9 and thus needs a standalone fix...)


gasman:
Merged in 978246f0baf1ddc0619b85396caf2389131b573c (stable/1.9.x) and 14f991df145df09ea45a7f5229ae558dd92dd657 (master).

Issue #970, opened 2015-02-06
Suggestion: set is_staff if moderator 


frankwiles:
Hi, 

While I realize Wagtail isn't typically used much in conjunction with the Django admin, I ran into a situation today that was a bit confusing at first and if I'm confused others will be.  Scenario: 

1) I setup some non-Wagtail admin interfaces (custom, not using the Django admin at all) wrapped with using @staff_required (well actually the django-braces version)

2) Setup users in the Wagtail admin as moderators, but not administrators 

3) Confused as to why the users were able to get into the Wagtail admin and not the custom admin menu items I had created. 

I can foresee scenarios where setting is_staff on a user automatically would be a bad thing, so perhaps it's enough to just call it out in the docs or perhaps have a Is Staff option in Roles in the Wagtail User admin?

gasman:
Hi @frankwiles,
I think it's best for Wagtail to avoid making any assumptions about the level of access to Django-admin that users should receive. As you say, there are scenarios where Django-admin access should be restricted even for users with effectively 'full' privileges over Wagtail. (Also, it's worth bearing in mind that there's nothing special about 'moderator' as a status within Wagtail - it's just a group that's created by the default Wagtail install with a particular set of publish permissions, and sites are free to delete it and replace it with whatever policy they like.)

At the moment, we have [a hard-coded rule](https://github.com/torchbox/wagtail/blob/f064fe93d03efef3090e50ca6ade5c290bca957a/wagtail/wagtailusers/forms.py#L132) that the 'is_staff' flag is set for administrators and only administrators - and that's not a brilliant design choice either, on similar grounds. To solve this with full generality, we probably need to expose the is_staff flag on the 'edit user' form, and possibly have some extra logic to set/unset that by default e.g. if Administrator is ticked. Needs more consideration...

frankwiles:
Agreed, I think just adding it to the User or Role forms will be great.

jorge-marques:
I'm currently experiencing an issue where an user modified by wagtail gets the is_staff flag set to false, even if it was previously set to true. Is this the expected behaviour?

gasman:
@jorge-marques This is the expected behaviour (or, to put it another way - it's a design flaw, not a programming bug). The `is_staff` flag is set to the value of `is_superuser` on save, so if you edit a user with is_staff=True, is_superuser=False through the Wagtail interface, the is_staff=True flag will be lost.

[...]

PR #3372, opened 2017-02-14
Use `django-treebeard` movenodeform form in the PageAdmin changelist in the vanilla Django admin

EmlynC:
Treebeard comes with a useful tree-editing form widget, let's use that rather than a less useful default changelist. 

This comes in quite handy when you want to rearrange the order and position of pages within the wagtail page tree and it makes sense to use code that's already in the library and use the vanilla django admin for it.

The old version looked like this:

image

... and the change allows you to drag and drop the pages in the page tree like this:

treebeard_admin

### DIFF
--- a/wagtail/project_template/project_name/settings/base.py
+++ b/wagtail/project_template/project_name/settings/base.py
@@ -43,6 +43,7 @@ INSTALLED_APPS = [
 
     'modelcluster',
     'taggit',
+    'treebeard',
 
     'django.contrib.admin',
     'django.contrib.auth',
--- a/wagtail/wagtailcore/admin.py
+++ b/wagtail/wagtailcore/admin.py
@@ -4,10 +4,18 @@ from django.contrib import admin
 from django.contrib.auth.admin import GroupAdmin
 from django.contrib.auth.models import Group
 
+from treebeard.admin import TreeAdmin
+from treebeard.forms import movenodeform_factory
+
 from wagtail.wagtailcore.models import GroupPagePermission, Page, Site
 
 admin.site.register(Site)
-admin.site.register(Page)
+
+# Use the Treebeard treeadmin.
+class PageAdmin(TreeAdmin):
+    form = movenodeform_factory(Page)
+
+admin.site.register(Page, PageAdmin)
 
 
 # Extend GroupAdmin to include page permissions as an inline
--- a/wagtail/wagtailcore/admin.py
+++ b/wagtail/wagtailcore/admin.py
@@ -1,9 +1,9 @@
 from __future__ import absolute_import, unicode_literals
 
+from django.apps import apps
 from django.contrib import admin
 from django.contrib.auth.admin import GroupAdmin
 from django.contrib.auth.models import Group
-from django.apps import apps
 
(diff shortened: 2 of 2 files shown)


gasman:
Thanks @EmlynC! This looks like a worthwhile addition. A couple of initial thoughts:

* Does the tree widget scale to very large numbers of pages (e.g. is there any sort of pagination in place)? It's not unknown for Wagtail sites to have thousands of pages, so if this isn't handled well then we should make this an optional setting.
* Perhaps it would be a good idea to test if treebeard is installed as an app (using [`apps.is_installed`](https://docs.djangoproject.com/en/1.10/ref/applications/#django.apps.apps.is_installed)) and fall back on the plain listing if not?

EmlynC:
Hey @gasman; yeah, the way I've been looking at Wagtail is that it's great for authors and the Django admin is nice for somewhat-technical-don't-know-python managers.

> Does the tree widget scale to very large numbers of pages (e.g. is there any sort of pagination in 
> place)?

I haven't tested how many pages it would scale to, I guess that needs testing. There is no pagination out the box or in this case, lazy loading of leafs. The JavaScript seems simple enough so I would suppose it makes sense to add a lot of pages, try it and profile it. It looks like to does quite a bit of painting when eyeballing it so I don't think it's going to scale well out of the box, but I'm happy to fix that for the sake of `django-treebeard` and `wagtail`.
 
Do you have a dummy Page generator or does one need writing? Could you define "... very large numbers of pages"?

> Perhaps it would be a good idea to test if treebeard is installed as an app (using apps.is_installed) 
> and fall back on the plain listing if not?

Solid idea. I'll add this now.

gasman:
> Do you have a dummy Page generator or does one need writing?

We don't have anything like that included in Wagtail at the moment, but you could borrow some code from benches.py in this PR:
https://github.com/wagtail/wagtail/pull/3355/files#diff-b3d068dc6468343da8f6683877276861

> Could you define "... very large numbers of pages"?

As an extreme example, we have one site in production with >100K pages :-)

gasman:
Revisiting this in the light of the latest decisions on #970 / #3434 (which will drop the current behaviour where superusers created via Wagtail are automatically given access to Django admin). Effectively, this means we're now taking a more formal stance along the lines of: "we don't expect or encourage people to use Django admin in the course of running a Wagtail site, and so in the default setup Wagtail should be fully decoupled from Django admin".

Of course, developers are welcome to configure the Django admin however they want - and as you show here, there are some very good reasons for making it available to users. However, in terms of presenting Wagtail as a coherent product, it's arguably better for us to say "Wagtail doesn't have drag-and-drop reordering for pages" than "Wagtail does have drag-and-drop page reordering, but you have to go to this other admin area, and it might not work well for very large sites".

For this reason, I think it's best to leave it up to the site implementer to enable the Treebeard admin view themselves (if they decide that it's appropriate for their use-case), rather than enabling it out of the box.

EmlynC:
Apologies for my slowness on this regard, this sounds sensible and I'll consider writing it up in a blog post — it is very easy to configure and I appreciate that it does not need to be in the vanilla install.

PR #3376, opened 2017-02-16
Drop Django 1.9 and Python 3.3 from CI

gasman:
As agreed at https://groups.google.com/d/msg/wagtail-developers/nTn-rthPMsA/ORP3GA-hCQAJ


gasman:
@kaedroho Good point! Merged in 63d52af0a32b7f1d9e20c29a943c7ff5106b9427, with documentation about supported versions updated in 187908e7738e64030ca72fdde8a0f7642fae1098.

PR #3379, opened 2017-02-16
fix #3378

### LINKED ISSUE #3378: [docs] Make autobuild regenerate docs builds when saving outside vagrant
Currently when running ``autobuild.sh`` inside the vagrant and making docs changes outside of it, there is not build at all.

I've fixed with:
 - upgrading watchdog from 0.8.3 to the git master branch in https://github.com/gorakhargosh/watchdog
- add a ``--debug-force-polling`` to autobuild.sh: ``watchmedo shell-command --debug-force-polling --patterns="*.rst" --ignore-pattern='_build/*' --recursive --command='make html; echo "Waiting for more changes..."' ``

I could make a PR, but I don't see when the watchdog library is installed in order to upgrade it.

chimeno:
Since autobuild.sh it's not documented, just a comment for others should fix #3378

### DIFF
--- a/docs/autobuild.sh
+++ b/docs/autobuild.sh
@@ -1,4 +1,8 @@
 #!/bin/sh
+# If you want to run this script inside a vm and save changes outside of it
+# You should upgrade watchdog library to the latest in https://github.com/gorakhargosh/watchdog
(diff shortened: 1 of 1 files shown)


gasman:
Thanks!

PR #3381, opened 2017-02-17
speed up tests

chimeno:
Trying to speedup tests.
Added some common patterns in Django to speed up them.

I was also trying to ``parallel`` them but since using inside a vm with 1 core..no gain ;)

Before: ``Ran 2704 tests in 505.227s``
After:  ``Ran 2704 tests in 469.659s``


gasman:
Thanks @chimeno!

It looks like this change is now [causing our Elasticsearch tests to fail](https://travis-ci.org/wagtail/wagtail/builds/202589901) - I'm guessing it's because those tests are checking for log messages that don't appear when DEBUG is false. Please can you investigate?

chimeno:
Sorry for the mess up in commits.
I've removed the logging.disable section and now elasticsearch tests are passing.

gasman:
Merged in 85de048a890aec4be3bcbf3af8a9ecfbf467d34a - thanks!

PR #3383, opened 2017-02-17
New admin preview

BertrandBordage:
Currently, the admin preview is easily breakable. You get an infinite spinner on the preview page when you:
- refresh the preview page;
- open a preview, submit or refresh the edit page then click again on preview.

As agreed with @tomdyson, I worked on a solution to fix these annoyances.

The solution was to avoid sending form data directly to the preview page. So data had to be stored somewhere, and I chose to store it in django session data. I automatically clean the session to remove old preview data to avoid increasing its weight too much. This is the best compromise I found to avoid creating an extra database table just for previews…

It also adds `WAGTAIL_AUTO_UPDATE_PREVIEW`, a setting that allows users to refresh a preview page to see new changes without having to click again on the preview button. I asked a few Wagtail users to know which behavior they preferred and it was fifty-fifty, so I decided to create this setting. Since this sends a new AJAX request on the edit page on each unsaved modification (which is a bit heavy), this is disabled by default. Note that there can be at max one AJAX request per second, to avoid performance issues, and requests are disabled while typing.

I noticed a browser performance issue while typing data, and thought at first it was coming from my changes, but it appears to be coming from something else since it also occurs on the master branch. I didn’t dig much, but it’s not happening on my deployed Wagtail 1.8 sites.

Another existing issue is that request.FILES is correctly sent to the preview page, but since files are not saved before rendering the page, it only leads to dead links. I chose to ignore request.FILES; we can’t find a solution unless we actually save these files temporarily. Since Wagtail sites mostly use wagtailimages and wagtaildocs instead, it seemed reasonable to put this aside for the moment.


BertrandBordage:
# Current Wagtail preview & how to break it
wagtail-old-preview
# How it behaves using this pull request
wagtail-new-preview
# Without `WAGTAIL_AUTO_UPDATE_PREVIEW` (default)
wagtail-preview-no-auto-update
# With `WAGTAIL_AUTO_UPDATE_PREVIEW = True`
wagtail-preview-auto-update

BertrandBordage:
You can get the slides from the presentation I made today at Wagtail Space at this address: https://noripyt.com/en/wagtail-space-2017/
The video of the presentation is available here: https://www.youtube.com/watch?v=ObM2pUgY-bs

wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
```diff
@@ -361,80 +361,67 @@ $(function() {
...
+    }
 
+    $previewButton.click(function(e) {
+        e.preventDefault();
+        var $this = $(this);
+        var $icon = $this.filter('.icon'),
+            thisPreviewUrl = $this.data('action');
+        $icon.addClass('icon-spinner').removeClass('icon-view');
+        var previewWindow = window.open('', thisPreviewUrl);
+        previewWindow.focus();
```
thibaudcolas: Not sure whether I like the idea of forcing focus on the preview window – presumably users opening a preview would expect to choose whether to look at it, or not.

wagtail/wagtailadmin/templates/wagtailadmin/pages/preview_error.html
```diff
@@ -0,0 +1,22 @@
...
+    </header>
+    <div class="nice-padding">
+        <p>
+            {% blocktrans %}
+                Impossible to preview this page, some errors are remaining.
+                Please close this tab, edit the page to fix these errors,
+                then use preview again.
+            {% endblocktrans %}
+        </p>
+        <a class="button" href="javascript:window.close()">
```
thibaudcolas: Could this JS be moved to a separate file?

wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
```diff
@@ -361,80 +361,67 @@ $(function() {
...
+            if (data['is_valid']) {
+                previewWindow.document.location = thisPreviewUrl;
+            } else {
+                window.focus();
+                previewWindow.close();
+                // TODO: Stop sending the form, as it removes file data.
+                $form.submit();
+            }
+        }).fail(function () {
+            alert('Error while sending preview data.');
```
thibaudcolas: This needs to be a message directly on the page instead, with some styles / translations and suggestion on how to remediate the error.

wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
BertrandBordage: Most of my clients would be a bit lost if the preview doesn’t focus.
But this is a controversial subject, so you can see that it doesn’t work the same on Chrome & Firefox. One of them (I don’t remember which one) focus the tab when we open it but does nothing when we explicitly call `.focus()`, while the other one doesn’t focus on opening but accepts explicit focuses.

wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
BertrandBordage: Where do you suggest we put that error message? If we put it on top of the page, like `messages.error`, it will not be visible if the user scrolled the page.

wagtail/wagtailadmin/templates/wagtailadmin/pages/preview_error.html
BertrandBordage: I know inline JS is considered a bad practice but… In this particular case, it’s not worth it in my opinion.
@gasman, what do you suggest?

BertrandBordage:
Thanks @thibaudcolas for the review! I answered you in the code.

wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
thibaudcolas: I guess the least confusing is then to follow the browser's behavior instead of explicitly calling `.focus()`? Would like to get others' opinion on this, as I know it's pretty controversial.

wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
thibaudcolas: I would suggest to put it in a modal window, so it clearly looks dismissable and is always visible when the error happens. Ideally this should be a part of Wagtail's UI used beyond this, but I'm not sure there is such a thing yet.

wagtail/wagtailadmin/templates/wagtailadmin/pages/preview_error.html
thibaudcolas: Ah sorry I didn't realise this was only for the "error" page, it's definitely overkill to have a separate file for that case.

HvandenBerg:
Tested and OK
Tested cross browser and cross device in browserstack both with and without the auto update setting. Updated pages with images, headers, links and a few other rich text items. Tried both refreshing the preview page and clicking the preview button after editing the page to see updated page.

wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
BertrandBordage: I had a chat with Olly from Torchbox, he’s in favour of keeping it like this, with an explicit focus.
I think we can keep it like this, and if people are annoyed by it and issues are open, we’ll simply reconsider.

wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
BertrandBordage: Hum… Alerts are also dismissible and they are also visible when the error occurs (except if you blocked them recently because too many of them were opened). I would also prefer adding a modal saying that an internal error occurred or users lost Internet connection, but I don’t think that’s easily possible to add one so… I’m a bit too lazy to write a view, template and static file just to display an unusual error more beautifully x)

wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
tomdyson: I think we should detail this in the release notes and encourage people to test this specific feature during the release candidate period. cc @ollywillans

BertrandBordage:
Thanks @HvandenBerg! Glad it worked well for you!

wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
```diff
@@ -361,80 +361,67 @@ $(function() {
...
+            processData: false,
+            contentType: false
+        });
+    }
 
-        }
+    if ($previewButton.data('auto-update')) {
+        var setPreviewDataTimeout = -1;
+
+        // Form data is changed when field values are changed (change event),
```
kaedroho: I think we should avoid sending any preview requests until the user clicks the preview button. Is that possible?

wagtail/wagtailadmin/views/pages.py
```diff
@@ -542,133 +544,96 @@ def view_draft(request, page_id):
...
-    page_class = content_type.model_class()
-    page = page_class()
-    edit_handler_class = page_class.get_edit_handler()
-    form_class = edit_handler_class.get_form_class(page_class)
-    parent_page = get_object_or_404(Page, id=parent_page_id).specific
-
-    form = form_class(request.POST, request.FILES, instance=page, parent_page=parent_page)
+class PreviewOnEdit(View):
+    http_method_names = ('post', 'get')
+    preview_expiration_timeout = 60 * 60 * 24  # seconds
```
kaedroho: Do previews need to be kept for a day? Wouldn't a timeout of 1 minute (or even 3 seconds!) do?

wagtail/wagtailadmin/views/pages.py
```diff
@@ -542,133 +544,96 @@ def view_draft(request, page_id):
...
+
+    def remove_old_preview_data(self):
+        expiration = time() - self.preview_expiration_timeout
+        for k, v in self.request.session.items():
+            if not k.startswith('preview-'):
```
kaedroho: Maybe we should prefix keys with ``wagtail-preview-``. Just in case some other project uses them/copies this code :)

wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
BertrandBordage: That’s right, I thought I did it, I must have forgotten. I’ll change it this afternoon.

wagtail/wagtailadmin/views/pages.py
BertrandBordage: It’s not really important, in fact. Every time the preview for a page is updated, the previous preview data is removed. That timeout is used to delete old previews for other pages. If you create a preview for page 1, wait 10 minutes, work on page 2 and make a preview of it, maybe you kept preview for page 1 in a tab. So it’s better to keep it for at least a few hours in my opinion. When you create a preview for page 2 more than 24 hours after a page 1 preview, then data from this preview 1 is deleted.

wagtail/wagtailadmin/views/pages.py
BertrandBordage: Right, I’ll change it this afternoon.

BertrandBordage:
Thanks a lot for the review, Karl! I’ll do the changes within a few hours.

BertrandBordage:
I applied the changes.

PR #3386, opened 2017-02-18
Updating Sphinx and other doc dep versions

daaray:
On OSX, I tried a fresh install and came across the following error when attempting to build the docs via `make html`:

screen shot 2017-02-17 at 10 23 36 pm

Updating the documentation libraries fixes the issue.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -8,7 +8,7 @@ pipeline:
     image: python:3.5
     commands:
       - XDG_CACHE_HOME=/drone/pip-cache pip install wheel
-      - XDG_CACHE_HOME=/drone/pip-cache pip install -e .[testing,docs]
+      - XDG_CACHE_HOME=/drone/pip-cache pip install -e .[testing]
       - isort --check-only --diff --recursive wagtail
   js:
     image: node:4.2.4
@@ -24,7 +24,7 @@ pipeline:
     image: python:3.5
     commands:
       - XDG_CACHE_HOME=/drone/pip-cache pip install wheel
-      - XDG_CACHE_HOME=/drone/pip-cache pip install -e .[testing,docs]
+      - XDG_CACHE_HOME=/drone/pip-cache pip install -e .[testing]
       - python -u runtests.py
 
 cache:
--- a/setup.py
+++ b/setup.py
@@ -53,11 +53,11 @@ testing_extras = [
 
 # Documentation dependencies
 documentation_extras = [
-    'pyenchant==1.6.6',
+    'pyenchant==1.6.8',
     'sphinxcontrib-spelling>=2.3.0',
-    'Sphinx>=1.3.1',
-    'sphinx-autobuild>=0.5.2',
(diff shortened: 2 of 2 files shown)


chimeno:
I was able to reproduce this bug in a macOS environment, but not inside the ``vagrant-wagtail-develop`` with the same libraries versions. :/
Extended Traceback:
````
# Sphinx version: 1.5.2
# Python version: 3.5.1 (CPython)
# Docutils version: 0.13.1 release
# Jinja2 version: 2.9.5
# Last messages:

# Loaded extensions:
  File "/Users/chimeno/Documents/experiments/testing/lib/python3.5/site-packages/sphinx/cmdline.py", line 295, in main
    opts.warningiserror, opts.tags, opts.verbosity, opts.jobs)
  File "/Users/chimeno/Documents/experiments/testing/lib/python3.5/site-packages/sphinx/application.py", line 199, in __init__
    self.setup_extension(extension)
  File "/Users/chimeno/Documents/experiments/testing/lib/python3.5/site-packages/sphinx/application.py", line 512, in setup_extension
    mod = __import__(extension, None, None, ['setup'])
  File "/Users/chimeno/Documents/experiments/testing/lib/python3.5/site-packages/sphinxcontrib/spelling/__init__.py", line 3, in <module>
    from .builder import SpellingBuilder
  File "/Users/chimeno/Documents/experiments/testing/lib/python3.5/site-packages/sphinxcontrib/spelling/builder.py", line 16, in <module>
    from enchant.tokenize import EmailFilter, WikiWordFilter
  File "/Users/chimeno/Documents/experiments/testing/lib/python3.5/site-packages/enchant/__init__.py", line 92, in <module>
    from enchant import _enchant as _e
  File "/Users/chimeno/Documents/experiments/testing/lib/python3.5/site-packages/enchant/_enchant.py", line 121, in <module>
    prefix_dir.contents = c_char_p(e_dir)
TypeError: bytes or integer address expected instead of str instance
````

After upgrading pyenchant to 1.6.8 the issue was resolved as mentioned above.

gasman:
Looks like our Drone setup [doesn't like installing pyenchant 1.6.8](https://drone.kaed.uk/wagtail/wagtail/147) - possibly because it isn't available as a .whl package and the .tar.gz has extra installation requirements. Please could someone figure out the extra incantations needed to get it to work, or see if 1.6.7 (which _does_ have a .whl) is sufficient to avoid the error?

chimeno:
As far as I know, the 1.6.7 doesn't have a .whl for macOS. And the are only .whl for _macOS_ and _Win_ in 1.6.8.
I guess the best option could be to install the **c  enchant library** in the Drone setup, but I'm not acquainted with how to do that.

gasman:
Doh, I'm overlooking the obvious here... Drone doesn't do anything with docs, so there's actually no reason to run `pip install -e .[testing,docs]` at all. Will remove the `docs` part and see if that fixes it...

PR #3393, opened 2017-02-21
Use render_basic for StructBlocks

timheap:
Having a StructBlock rendered via a template was inconsistent with other blocks, It also meant that overriding `render_basic()` in a child class did nothing, again unlike all other blocks. Providing a `render_basic()` is useful for block libraries, where a `render_basic()` provides a very minimal implementation with the expectation that developers will provide their own rendering via a template.

TableBlock is the only other block with a template, but I think that one is legitimate.


gasman:
Thanks @timheap!

Issue #3384, opened 2017-02-17
Wagtailsearchpromotions: Select of "Popular search terms" work only on the first page.


Vorlif:
### Issue Summary

If I use wagtailsearchpromotions and try to "Choose from popular search terms" this will only work if I don't change the page. After change (to the next page) I can not select a item.

![peek 2017-02-17 17-32](https://cloud.githubusercontent.com/assets/4333770/23073874/2dce61b0-f537-11e6-9424-93a385359f72.gif)

### Steps to Reproduce

1.  `wagtail start my_site`
2. Enalble wagtailsearchpromotions
3. Go to the search site
4. Make 11 example inputs
5. Go to the Search-Promotion-Admin Pannel and add a new promote result
6. Click "Choose from popular search terms"
7. Change the site over the modal pagignation (To page 2)
8. Try to select a item (Don't work)

Tested in Chrome and Firefox.

### Technical details

* Python version: 3.6.0
* Django version: 1.10.5
* Wagtail version: 1.9
* Browser version: Chrome 56.0.2924.87 (64-bit), Firefox 51.0.1 (64-Bit)

PR #3395, opened 2017-02-21
Fix: Select of "Popular search terms" work only on the first page.

jjanssen:
This PR fixes the issue belonging to #3384. For some reason it had an Ajax callback when clicking on query terms in the modal listing. But the response was not adding any additional functionality, it was only throwing an `eval` error since the it isn't the correct response the `ModalWorkflow` was expecting.

I removed the listening for `modal.loadUrl` and bound it to the method to select the query and close the modal.


jjanssen:
@gasman I'm requesting you for a review instead of Thibaud, since I saw you originally implemented this section.

gasman:
Thanks @jjanssen! Merged in 23649990c269285570fb0be16329de10a8f8ab8b.

Not sure what was going on in the old code there :-)

Issue #3268, opened 2017-01-06
Translation for 'file' ignored in WagtailImage


robslotboom:
### Issue Summary

The translation of **file** in **wagtailimages** is being ignored in the **edit image view**.

I had a look at the original model and the form and only the **verbose_name** of the **file field** contains _('file'); seems ok.

When changing it temporarily to _('heeeeeelp') the view keeps displaying “file”.

### Technical details

* Python version: 3.5.0..
* Django version: 1.01.2.
* Wagtail version: 1.8.

strindhaug:
Normally the formfield() method in django.db.models.fields.Field sets the `label` using `capfirst` of the `verbose_name`. But since wagtail/wagtailimages/forms.py overrides and initializes the WagtailImageField directly the `label` value is never set and as far as I can understand it falls back to capfirst'ing the fieldname itself.

I fixed it by adding the verbose_name capfirsted when initializing the WagtailImageField.

gasman:
Fixed in #3396

PR #3396, opened 2017-02-22
Make file-label translated

strindhaug:
Fix for #3268 

Set the label as capfirst of the verbose_name of the db-field so that the label is translated correctly.


kaedroho:
> Is there any way for me to cancel travis, when I realize there is a mistake and failing every other build is just a waste of time?

You might be able to do this if you login to travis, find and running build and click the "Cancel build" button on the top right. If it doesn't let you, we can do it :)

(I've just cancelled the travis build for your second commit)

gasman:
Merged in 35305e1649c4fd10b64a3616982021340f1811ff - thanks @strindhaug!

Issue #3244, opened 2016-12-21
Warn an admin before they remove their own admin access


tmsndrs:
As a test I just changed my own permissions and removed my administrative access. The action was successful but I was then shown a "Sorry, you do not have permission to access this area." message and I have no way of changing it back. 

If an admin removes their own admin access it would be good if they at least gave a warning that the action can't be reversed and permissions will be greatly restricted to avoid people locking themselves out. 
![screen shot 2016-12-21 at 17 31 48](https://cloud.githubusercontent.com/assets/7290588/21399499/d99aed68-c7a3-11e6-80f4-7dda8379b558.png)

strindhaug:
It's generally a good idea to prohibit users from modifying their own access permissions at all, both because of this, and because it hypothetically could also be a way for a user to escalate their own privileges (assuming there is some other flaw/bug allowing this).

And it's much easier to safely verify that `userToModify != currentUser` than checking whether there still would be some other admin left in the system if current user demotes theirself etc.

gasman:
Resolved in #3502 - users can no longer unset their own is_superuser (or is_active) flag.

Issue #537, opened 2014-08-13
Giving non-admins 'create user' permission within wagtailusers, without giving them the ability to create admin users


gasman:
Ref: https://groups.google.com/d/msg/wagtail/t_nnwIO8pSc/T70-N2OMvQUJ

It's currently possible to assign "create/edit/delete user" permissions to non-admin users, to give them access to the user management area. However, there's nothing to stop them from escalating their own permissions by creating their own admin accounts.

Proposed solution: On the 'edit user' form, only provide the 'Administrator' checkbox if the user accessing the form is an administrator themselves. More generally: the list of groups that are shown to a non-admin should only include groups that they are a member of themselves - i.e. there is no way to create a user with more privileges than yourself.

pvdvreede:
I have created a PR for only letting superusers created other superusers (#1149).

The validation is on the Forms and I have injected the `request.user` object into the form. This should make it easy to also filter the groups based on the logged in user. 

For my use case I would like to have non-admin users be able to create other users of the same group as them, but also create users of a lesser group. Just wondering if this would be possible to be able to determine in the code? Seems like it might be hard to do as there is no ordering or hierarchy of groups...

gasman:
Defining an ordering / hierarchy of groups does indeed seem like quite a 'heavyweight' solution to this problem. I guess you could do it by saying "user X can add users to group Y if the permissions of group Y are a subset of user X's own permissions".

For most cases, though, I think it would be sufficient to say that user X must be a member of the relevant groups. For example: if you want user X to be able to create users of type 'editor' and 'moderator', then you should make user X a member of both groups, not just 'moderators'.

pvdvreede:
Thanks @gasman, you make a good point - I can just add that user to both groups, and then the logic can just be a user can only create other users with the groups they are in.

hnykda:
Is there any progress on this? It seems quite silent. I would be happy to work on this if I knew the status.

gasman:
Recent discussion among the core team has raised the point that even with this change in place, a non-superuser with edit permission can still escalate their privileges by changing the password or email address of an existing user and taking over their account.

This would suggest that the user admin views inherently _can't_ be locked down for non-superusers, without removing functionality that people would legitimately want. (I can easily imagine resetting passwords being the kind of task that gets assigned to a "junior admin", non-superuser role.)

[...]

PR #3399, opened 2017-02-23
Prevent self demotion

strindhaug:
Fix for #3244

Disable checkboxes for is_active and is_superuser when `can_delete` is false, thus preventing admins from accidentally deactivating themselves or  "demoting" themselves to non-admin.  (Admins are are already prevented from *deleting* themselves)

### DIFF
--- a/wagtail/wagtailusers/views/users.py
+++ b/wagtail/wagtailusers/views/users.py
@@ -157,6 +157,11 @@ def edit(request, user_id):
     else:
         form = get_user_edit_form()(instance=user)
 
+    # If you're not allowed to delete a user,
+    # you should not be able to deactivate or demote them either:
+    form.fields['is_active'].widget.attrs['disabled'] = not can_delete
+    form.fields['is_superuser'].widget.attrs['disabled'] = not can_delete
+
     return render(request, 'wagtailusers/users/edit.html', {
         'user': user,
         'form': form,


strindhaug:
This change just disables the checkboxes, so the checkbox is unresponsive. 

But disabled checkboxes and radiobuttons aren't styled any differently; so they'll look normal but not respond. 
I've experimented with adding some styling of disabled radio buttons and checkboxes to make it less confusing. But I decided to not include it for now since this needs some design decisions.

gasman:
Hi @strindhaug,
I think there are a couple of problems with this approach:

* We can't assume that "user editing their own record" is the only situation in which `can_delete` is false. An admin could conceivably set up a "managers" user group with add/edit (but not delete) permission over the User model, and in this case those managers should still logically keep the ability to edit the superuser / active flags. (Admittedly, this setup would have its own security problems, since there's nothing to stop a manager from creating a superuser account for themselves - but that's something we might fix in future (#537), and in any case, I think it's important for the system to behave correctly according to what the admin specifies, even if that is insecure...)
* The checkboxes are only being disabled at the "cosmetic" level - they can still be hacked through the DOM inspector (for example), and the form handling logic won't validate it. (Again, this is a fairly theoretical concern at the moment, because you can't cause any damage other than locking yourself out... however, as a general policy, we really want to avoid this in any permission/security-related code.)

A better solution would be to update wagtailusers.forms.UserForm to accept the current user as a new argument to `__init__` - the form would then be responsible for enforcing and validating the "users cannot demote themselves" business logic.

strindhaug:
Yes, I realize using can_delete is a bit overly broad proxy for "not myself", and that disabling form element isn't secure against deliberate trying to lock yourself out... 

But you can only ever try to protect an admin from accidentally locking themself out; because if they really want to lock themself and everyone else out, they could still just delete everyone else, and set their own password to a long generated random string and not remember/store the new password. ;)

But you're right it's cleaner if the form enforces this.

strindhaug:
Maybe there should be a special "edit myself" form? One that don't require any special privileges (besides access to the backend) that only change personal info (not privileges and account status), that can be used by everyone and that is used when an admin tries to edit themselves in the user list. I'd think that would look a lot cleaner than having a form full of disabled elements. This form could be linked from the `wagtailadmin_account` (`/admin/account/`) page.

PR #3401, opened 2017-02-24
Prevent users from demoting or deactivating themself

strindhaug:
Remove is_active and is_superuser fields completely when editing oneself, to avoid locking oneself out.

Another fix for #3244 (replacing my first attempt #3399)

### DIFF
--- a/wagtail/wagtailusers/forms.py
+++ b/wagtail/wagtailusers/forms.py
@@ -154,6 +154,14 @@ class UserEditForm(UsernameForm):
         help_text=_("Administrators have full access to manage any object or setting.")
     )
 
+    def __init__(self, *args, **kwargs):
+        editing_self = kwargs.pop('editing_self', False)
+        super(UserEditForm, self).__init__(*args, **kwargs)
+
+        if editing_self:
+            del self.fields["is_active"]
+            del self.fields["is_superuser"]
+
     class Meta:
         model = User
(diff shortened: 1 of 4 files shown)


strindhaug:
Maybe groups also should be removed when editing self for non-superusers?

gasman:
Thanks @strindhaug! This looks good to me - just needs tests.

> Maybe groups also should be removed when editing self for non-superusers?
I think that's a sensible idea, but it'll only really become relevant after we implement #537 - so it can be left out for now.

strindhaug:
I tried to add some tests. It seems to verify that this fix prevents admins from deactivating or demoting themselves.

At first I was a bit confused because the existing tests didn't include `is_active` or `is_superuser` in the data, but after a while I realized that the existing test actually were deactivating and demoting the test user. So I added a test both with and without these fields and added checks that the status were indeed updated.

I also tried to add a test for non-admins escalating their own privilege by adding fields in the DOM and there is some weird issue that no matter what status code I tried expecting it  was not what I expected... even though the user did manage to change their own name but not their admin-status... So I commented out the redirect test temporarily. Maybe someone else understand what is wrong... (I'm guessing I haven't properly set the edit-permissions on the `deleter_user` that I re-used for this test)

gasman:
Closing this in favour of #3502 (which incorporates this one).

PR #3404, opened 2017-02-27
Update results.html

cho-leukeleu:
Close anchor tag.
Note though that the missing closing tag is inside a blocktrans!


gasman:
Thanks @cho-leukeleu!

We should also add a `?` after the closing tag, for consistency with other messages (e.g. https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailimages/templates/wagtailimages/images/results.html).

gasman:
Merged in d2baa4ba05acc92fff3369dbb8cb344ffdd7a70b - thanks!

PR #3405, opened 2017-02-27
Fix reorder past a deleted item

### LINKED ISSUE #2624: Ordering wagtailsearchpromotions results can't reorder past a deleted item
With the fix from #2623 applied:
- From Settings -> Promoted search results -> Add New Promoted Result, insert three recommended pages with descriptions filled in
- delete the middle one
- click the 'up' arrow on the last item. It does not move.

jjanssen:
Fixes #2624 which also effects inline panels since it is referring to ``page-editor.js``

It turns out `currentChild.prev(':visible')`/`currentChild.next(':visible')` do not take `display: none;` into [consideration when checking for visibility](https://api.jquery.com/visible-selector/), which is applied by when pressing delete. Since the click event also applies the `.delete` class if have added the check for that, rather than `:visible`.

Ideally it would have been nice to use `prevUntil` or `nextUntil` as far as traversing goes, but unfortunately a syntax like `prevUntil(':not(.deleted)')` is not possible.

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/page-editor.js
@@ -132,7 +132,7 @@ function InlinePanel(opts) {
                 var currentChildOrder = currentChildOrderElem.val();
 
                 /* find the previous visible 'inline_child' li before this one */
-                var prevChild = currentChild.prev(':visible');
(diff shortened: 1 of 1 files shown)


HvandenBerg:
Tested OK

gasman:
Merged in c1a751e6c0d320a37fee78ed5982e2e5e0fd0f6e - thanks!

PR #3410, opened 2017-02-28
Fix possibility of tabbed interface conflicting with regular IDs

### LINKED ISSUE #3409: 'Footer' is invalid tab heading
### Issue Summary
When you create a new tab in the admin with heading 'Footer', you don't get any errors (in python nor javascript) but the tab doesn't work. You can click on the tab but the content doesn't change.
Note: when 'Footer' is the first tab it works initially, but you cannot go back and forth between the tabs.

### Steps to Reproduce

1. Create a tab with name 'Footer':
```python
    ...
    ObjectList([...], heading='Footer'),  # < this heading doesn't work
    ...
])
```
2. Edit or create a page with this tab
3. Try to click back and forth to this tab and notice it's not working.

A workaround is to name the tab something else like 'Footer content', but since we mark these headings as translatable there is no guarantee that they will not be named 'footer' or some other 'forbidden' name.
This issue has probably something to do with the fact that the tab has `href="#footer"` and there is an actual (unrelated) element with that ID. Maybe a prefix for this would be safer, like `href="#editor-tab-footer"`.

### Technical details

* Python version: 3.6
* Django version: 1.10
* Wagtail version: 1.9
* Browser version: Latest Firefox and Chrome

jjanssen:
Fixes #3409 by prefixing the id with ``tab-``

### DIFF
--- a/wagtail/wagtailadmin/templates/wagtailadmin/edit_handlers/tabbed_interface.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/edit_handlers/tabbed_interface.html
@@ -1,13 +1,13 @@
 {% load wagtailadmin_tags %}
 <ul class="tab-nav merged">
     {% for child in self.children %}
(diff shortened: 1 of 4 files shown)


gasman:
Thanks @jjanssen! Merged in 6e9965d47f357bf903017e4db5c89967aa144ae6.

PR #3411, opened 2017-02-28
Flake8 plugins

kaedroho:
This pull request installs a couple of flake8 plugins to perform more code-style checks on Drone:

 - ``flake8-blind-except`` - Catches ``except:`` statements that do not catch specific exception classes
 - ``flake8-print`` - Catches stray print functions. These shouldn't be used in most parts of Wagtail and leaving them in is a common mistake

I did want to enable a couple more plugins, but they raise too many errors to solve right now:
 - ``flake8-docstrings`` - Runs ``pydocstyle`` to check docstring style (PEP257)
 - ``flake8-mutable`` - Catches mutable default arguments which can cause problems (most of the errors are in tests, but there are a few in actual code that should probably be looked at)

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,14 +1,10 @@
 pipeline:
-  flake8:
-    image: python:3.5
-    commands:
-      - XDG_CACHE_HOME=/drone/pip-cache pip install flake8
-      - flake8 wagtail
-  isort:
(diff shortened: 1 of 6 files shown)


kaedroho:
Ready to review! ping @wagtail/backend

PR #3413, opened 2017-02-28
Dj1.11 - Fix rendering of group permission checkboxes

gasman:
Incorporates #3406

The format_permissions tag used to lay out the permission checkboxes worked by iterating over the BoundField, which gave a sequence of CheckboxChoiceInput objects. This class has been dropped in 1.11 and superseded by BoundWidget, which has a different way of accessing the checkbox value. Add separate code paths for the two cases.


m1kola:
Merged in 7ae90564292ffc8cbb07b3593a0e03fb147589d9.

Thanks, Matthew!

PR #3415, opened 2017-02-28
Fixes hallotoolbar misplacement within StreamField StructBlock

### LINKED ISSUE #1511: Hallo toolbar appears in wrong place on rich text fields inside streamfield structblocks
screen shot 2015-07-10 at 15 52 33

Padding is added to the rich text block to accommodate the hallo toolbar and make it look like the toolbar is "inside" the field, as normal in wagtail. However trumping styles applied to all toolbars inside Streamfields assume that the rich text blocks are all at the root of the streamfield and are displayed without borders. This isn't true of Structblock rich text fields, which get these styles inappropriately.

jjanssen:
Fixes #1511 which now also considers if the element is a part of a struct-block. The specific classname ``.stream-field`` pushes the HalloToolbar up for a negative margin offset. Whereas in a StructBlock it is sufficient to apply the the regular margin offset which comes along with `.hallotoolbar`

- [x] Fixed it for a `StructBlock` within a `StreamField`. 
- [x] Validated it is still working within a `StreamBlock` 
- [x] Validated the direct `RichTextField` usage on a `Page`. 
- [x] Do some cross-browser testing.

I am not completely sure if there are additional scenarios to look in to?

I'd also rather have a deep look into what's happening with `.stream-field` and `.hallotoolbar` in general so I'm not completely happy with the solution. On the other hand, this is probably good enough with #2409 in mind.

~~Also I wanted to attach a screenshot but it seems GitHub uses [S3 storage](https://twitter.com/awscloud/status/836666548311859200) for that.~~

**Update:** 
What it looks like:
image

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/hallo-bootstrap.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/hallo-bootstrap.js
@@ -27,10 +27,11 @@ function makeHalloRichTextEditable(id) {
     }
 
     var closestObj = input.closest('.object');
+    var isDeepNested = input.closest('.struct-block').length;
 
     richText.hallo({
         toolbar: 'halloToolbarFixed',
-        toolbarCssClass: (closestObj.hasClass('full')) ? 'full' : (closestObj.hasClass('stream-field')) ? 'stream-field' : '',
+        toolbarCssClass: (closestObj.hasClass('full')) ? 'full' : (closestObj.hasClass('stream-field') && !isDeepNested) ? 'stream-field' : '',
         plugins: halloPlugins
     }).bind('hallomodified', function(event, data) {
         input.val(data.content);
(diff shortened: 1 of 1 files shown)


wagtail/wagtailadmin/static_src/wagtailadmin/js/hallo-bootstrap.js
```diff
@@ -27,10 +27,11 @@ function makeHalloRichTextEditable(id) {
     }
 
     var closestObj = input.closest('.object');
+    var isDeepNested = input.closest('.struct-block').length;
```
BertrandBordage: Maybe explicitly change this variable to a boolean?
It could be something like:
```javascript
var isRoot = input.closest('.struct-block').length == 0;
```
Then you can directly use `isRoot` instead of negating `isDeepNested`.

jjanssen:
Thanks for reviewing @BertrandBordage! I have updated the PR with your suggestions.

BertrandBordage:
Thanks @jjanssen! I tested on different scenarii, with and without streamfield, using different depths/levels, and it works flawlessly.

Merging it!

SalahAdDin:
Pefect!

PR #3418, opened 2017-03-01
Title not displaying page name

trumpet2012:
The confirm move page was not displaying the correct title for the page.

The `page` context variable does not exist for that view.


jjanssen:
Thanks! Merged in f281306

PR #3421, opened 2017-03-03
Create raw SQL for SQL Server to be able to display recent edits (on 1.9.x)

FlipperPA:
This creates corresponding raw SQL for a subselect to maintain SQL Server compatibility.

SQL Server does not support the LIMIT keyword. This should be considered a temporary fix until Django 1.11 subselects are available!


gasman:
Thanks @FlipperPA! To keep things in one place I'll take #3422 as the canonical PR for this - we'll make a separate decision on whether to make a bugfix release for the 1.9 branch. (If we do, we should probably make a 1.8 one too.)

PR #3422, opened 2017-03-03
Create raw SQL for SQL Server to be able to display recent edits (on master)

FlipperPA:
This creates corresponding raw SQL for a subselect to maintain SQL Server compatibility.

SQL Server does not support the LIMIT keyword. This should be considered a temporary fix until Django 1.11 subselects are available!

### DIFF
--- a/wagtail/wagtailadmin/views/home.py
+++ b/wagtail/wagtailadmin/views/home.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import, unicode_literals
 
 from django.conf import settings
+from django.db import connection
 from django.shortcuts import render
 from django.template.loader import render_to_string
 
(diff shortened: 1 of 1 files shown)


adamgraves-choices:
We've found this fix doesn't work after you've created a page and attempt to browse to it:

`AttributeError at /admin/
'pyodbc.Row' object has no attribute 'page'
Request Method:	GET
Request URL:	http://bth.staging.beta.nhschoices.net/admin/
Django Version:	1.10.6
Exception Type:	AttributeError
Exception Value:	
'pyodbc.Row' object has no attribute 'page'
Exception Location:	/usr/local/lib/python2.7/dist-packages/wagtail/wagtailadmin/views/home.py in __init__, line 82
Python Executable:	/usr/bin/python
Python Version:	2.7.12
Python Path:	
['/wagtail',
 '/usr/lib/python2.7',
 '/usr/lib/python2.7/plat-x86_64-linux-gnu',
 '/usr/lib/python2.7/lib-tk',
 '/usr/lib/python2.7/lib-old',
 '/usr/lib/python2.7/lib-dynload',
 '/usr/local/lib/python2.7/dist-packages',
 '/usr/lib/python2.7/dist-packages']
Server time:	Fri, 17 Mar 2017 16:02:49 +0000`

We've got around this by using `last_edits = PageRevision.objects.raw(` instead of `last_edits = cursor.execute(`

wagtail/wagtailadmin/views/home.py
```diff
@@ -49,14 +50,33 @@ class RecentEditsPanel(object):
...
+                SELECT wp.* FROM
+                    wagtailcore_pagerevision wp JOIN (
+                        SELECT max(created_at) AS max_created_at, page_id
+                        FROM wagtailcore_pagerevision
+                        WHERE user_id = %s
+                        GROUP BY page_id
+                        ORDER BY max_created_at DESC
+                        LIMIT %s
+                    ) AS max_rev ON max_rev.max_created_at = wp.created_at ORDER BY wp.created_at DESC
+                 """, [self.request.user.pk, 5])
```
BertrandBordage: Why not use the ORM?
All this can be replaced with:
```python
last_edits_dates = (PageRevision.objects.filter(user=self.request.user)
                    .values('page_id').annotate(latest_date=Max('created_at'))
                    .order_by('-latest_date').values('latest_date')[:5])
last_edits = PageRevision.objects.filter(created_at__in=last_edits_dates).order_by('-created_at')
```

Then the following line converting it to a list can be removed.

wagtail/wagtailadmin/views/home.py
```diff
@@ -49,14 +50,33 @@ class RecentEditsPanel(object):
...
         last_edits = list(last_edits)
         page_keys = [pr.page.pk for pr in last_edits]
```
BertrandBordage: `pr.page.pk` should be renamed to `page_id` to avoid generating 5 extra queries.

wagtail/wagtailadmin/views/home.py
```diff
@@ -49,14 +50,33 @@ class RecentEditsPanel(object):
...
         specific_pages = Page.objects.filter(pk__in=page_keys).specific()
```
BertrandBordage: This should be replaced by `QuerySet.in_bulk()` and the next line can be removed.

gasman: `in_bulk` looks useful here, but I can't see how to combine it with the call to `specific` (other than calling `specific` on each page in the resulting dict, which is less efficient than the queryset method) - any suggestions @BertrandBordage?

BertrandBordage: I’m not 100% familiar with the way `.specific()` works, but it returns a a QuerySet, so in theory we should be able to do `Page.objects.specific().in_bulk(page_keys)`.

BertrandBordage: Just tested on a project: it works.

BertrandBordage: (By the way, I just looked at the way `SpecificQuerySet` works, it’s by far more elegant than django-polymorphic, the library usually used to do that! Congrats and thanks to the one that created it :) )

gasman: Looks good, thanks! I'll create a PR for your version of the fix shortly.

All credit for `SpecificQuerySet` goes to @timheap :-)

gasman:
Merged @BertrandBordage's solution via #3582.

PR #3423, opened 2017-03-03
Improve rendering of ManyToManyField values (or any other property / attribute that returns a Manager) in InspectView

### LINKED ISSUE #3298: ManyToManyFields display incorrectly in ModelAdmin inspect view
### Issue Summary

If a model with a ManyToManyField is registered with ModelAdmin with the Inspect view enabled, the ManyToManyField is included in the inspect output and always displays a value of `some_app.SomeTargetModel.None`. This happens because InspectView.get_field_display_value returns `getattr(instance, field_name)`, which in the case of a ManyToManyField is the manager object for the relation; for some reason the string representation of a manager is `some_app.SomeTargetModel.None`.

### Steps to Reproduce

On a fresh Wagtail project, add `wagtail.contrib.modeladmin` to INSTALLED_APPS and edit `home/models.py` to:

    from django.db import models

    from wagtail.wagtailadmin.edit_handlers import FieldPanel

    class HomePage(Page):
        pass

    class Musician(models.Model):

        panels = [
            FieldPanel('name'),
        ]

        def __str__(self):
            return self.name

    class Band(models.Model):
        members = models.ManyToManyField(Musician)

        panels = [
            FieldPanel('name'),
            FieldPanel('members'),
        ]

        def __str__(self):
            return self.name

Create `home/wagtail_hooks.py`:

    from wagtail.contrib.modeladmin.options import ModelAdmin, modeladmin_register

    from home.models import Musician, Band

    class MusicianAdmin(ModelAdmin):
        model = Musician
        inspect_view_enabled = True

    modeladmin_register(MusicianAdmin)

    class BandAdmin(ModelAdmin):
        model = Band
        inspect_view_enabled = True

    modeladmin_register(BandAdmin)

From the Wagtail admin, create some musician / band objects then go to the inspect view for a band.

Expected behaviour: Either the 'members' field should be silently skipped (since that's what currently happens with one-to-many relations, and we might as well be consistent...), or it should output a list of band members

Actual behaviour: It displays the 'members' value as `home.Musician.None` (even if the band has some members).

### Technical details

* Python version: 3.4.3
* Django version: 1.10.4
* Wagtail version: master (1.9a0)
* Browser version: Chrome 55.0.2883.95

ababic:
Fixes #3298

Before:

screenshot 2017-03-03 at 22 13 32

After:

screenshot 2017-03-03 at 22 13 53

### DIFF
--- a/wagtail/contrib/modeladmin/options.py
+++ b/wagtail/contrib/modeladmin/options.py
@@ -346,15 +346,12 @@ class ModelAdmin(WagtailRegisterable):
         'inspect_view_fields_exclude' not being included.
         """
         if not self.inspect_view_fields:
-            found_fields = []
+            field_names = []
             for f in self.model._meta.get_fields():
-                if f.name not in self.inspect_view_fields_exclude:
-                    if f.concrete and (
-                        not f.is_relation or
-                        (not f.auto_created and f.related_model)
-                    ):
-                        found_fields.append(f.name)
-            return found_fields
+                exclude_names = self.inspect_view_fields_exclude
+                if isinstance(f, Field) and f.name not in exclude_names:
+                    field_names.append(f.name)
+            return field_names
         return self.inspect_view_fields
 
     def index_view(self, request):
--- a/wagtail/contrib/modeladmin/options.py
+++ b/wagtail/contrib/modeladmin/options.py
@@ -5,6 +5,7 @@ from django.conf.urls import url
 from django.contrib.auth.models import Permission
 from django.core.exceptions import ImproperlyConfigured
 from django.db.models import Model
+from django.db.models.fields import Field
 from django.forms.utils import flatatt
 from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext_lazy as _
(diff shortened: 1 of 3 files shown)


ababic:
Don't merge / review just yet. For consistency, I'm also going to modify the modeladmin's get_inspect_view_fields() method to include all fields by default (similar to how wagtail's edit handler adds all fields to edit forms)

ababic:
Hi @gasman. I know we discussed this once already and both agreed that we should be doing as little in the view to differentiate between different sorts of values as possible. But, the `__repr__()` value for `QuerySet` didn't really seem suitable for front-end output.

Evaluating `QuerySet` to a list and returning that instead seemed better, but I think the remaining parenthesis might still seem confusing to some:

screenshot 2017-03-04 at 17 29 43

What I settled on in the end was a comma-separated representation, which is more consistent with the representation for ForeignKey fields:

screenshot 2017-03-05 at 12 03 22

In regards to the `get_inspect_view_fields()` method on `ModelAdmin`, I had a play around with that, but ended up reverting any changes. I was mistaken when I thought it should be excluding ManyToMany fields. It's correct that they should be showing now that they are supported.

While I was investigating this, I also realised there was an issue with the representation of Boolean fields, which I've also fixed and added a test for.

ababic:
Have added a test for the ManyToMany output now too. 

We should definitely make a note somewhere to review where this logic to get the representations for values should live, and how we can make it easier for developers to override display values for specific fields / attributes. I imagine there's probably a bit of crossover with the 'compare' functionality (and possibly other places) for the representation stuff.

Issue #3164, opened 2016-11-16
Rich text field and PageChooserBlock should use 'specific' URLs


dwasyl:
I have a site that's implemented `wagtail 1.7` with `wagtail-modeltransation`, but when I use inline PageChooser links in a RichTextBlock I'm unable to get the link for the page in other languages. `wagtail-modeltranslation` generates page URLs by overriding the `.url` method on the Page models configured for translation. This means that the only way to get the right links to use `.specific.url` rather than just `.url` on the generic Page model.

Would it cause any issues to make the `PageLinkHandler` in `rich_text.py` or the `PageChooserBlock` in `blocks/field_block.py` use `value.specific.url` when they render the `value.url`? This would be an easy fix for the problem and better enable the use of translated Page models.

gasman:
Hi @dwasyl - yes, I think it would be appropriate to use `specific` in both PageLinkHandler and PageChooserBlock. Thanks for the report!

dwasyl:
Okay great, thanks @gasman. Is this something you can just include in another patch or would a PR be helpful?

gasman:
A PR would definitely be helpful, thanks!

gasman:
Fixed in #3425 / #3494

PR #3425, opened 2017-03-06
Use specific page when populating href

garykrige:
This PR addresses an issue with RichText internal links that use a Model inheriting from Page that overrides the url function.

The issue has been reported here too: https://github.com/wagtail/wagtail/issues/3164

### DIFF
--- a/wagtail/wagtailadmin/templates/wagtailadmin/pages/listing/_page_title_choose.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/pages/listing/_page_title_choose.html
@@ -6,9 +6,7 @@ Expects a variable 'page', the page instance.
 
 <h2>
     {% if page.can_choose %}
-        {% with page=page.specific %}
-            <a class="choose-page" href="#{{ page.id }}" data-id="{{ page.id }}" data-title="{{ page.get_admin_display_title }}" data-url="{{ page.url }}" data-parent-id="{{ page.get_parent.id }}" data-edit-url="{% url 'wagtailadmin_pages:edit' page.id %}">{{ page.get_admin_display_title }}</a>
-        {% endwith %}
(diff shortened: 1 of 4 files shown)


gasman:
Thanks @garykrige! Please can you add tests for this?

wagtail/wagtailadmin/templates/wagtailadmin/pages/listing/_page_title_choose.html
```diff
@@ -6,7 +6,9 @@
 
 <h2>
     {% if page.can_choose %}
-        <a class="choose-page" href="#{{ page.id }}" data-id="{{ page.id }}" data-title="{{ page.get_admin_display_title }}" data-url="{{ page.url }}" data-parent-id="{{ page.get_parent.id }}" data-edit-url="{% url 'wagtailadmin_pages:edit' page.id %}">{{ page.get_admin_display_title }}</a>
+        {% with page=page.specific %}
```
gasman: It'll probably be more efficient to apply `specific()` to the page queryset in the view (i.e. use `pages = parent_page.get_children().specific()` [here](https://github.com/wagtail/wagtail/blob/21f1292cf75c6c51436ee01d2f3f656c6d7667bf/wagtail/wagtailadmin/views/chooser.py#L77)), as that will only perform one extra database query for each different page type in the list, rather than one for every page.

garykrige: :100:

garykrige:
@gasman the ci on drone seems to be broken.

gasman:
The test for the page chooser response looks good, but it looks like we still need one to confirm that links in rich text are rewritten correctly. A version of https://github.com/wagtail/wagtail/blob/eb9cc639d2c8a30911ed1069f41c77eb617cfc2f/wagtail/wagtailcore/tests/test_rich_text.py#L141 where the linked page is an instance of SingleEventPage should do the job. (ping @huubbouma)

gasman:
Merged in 6aea6a68569d5df002c901c4e1bca0591078c46f via #3494. Thanks @garykrige and @huubbouma!

PR #3427, opened 2017-03-06
Remove duplicate translations

chimeno:
While translating to new language, I've found several strings as:
"Groups"
"groups"

"Change password"
"change Password"

So I added a new utility script to find these strings in the translation files and then I've manually changed those values.
Just in case you find it useful to add it to the source code.
At first I thought there were a lot of them, but I was wrong :)

### DIFF
--- a/scripts/find-duplicates.py
+++ b/scripts/find-duplicates.py
@@ -1,0 +1,48 @@
+"""
+Utility script to find strings marked for translation that are
+in the form:
+'Change Password' -> 'change password'
+In order to minimize translation efforts, only one should be used.
+It uses babel instead more standard 'polib' because it's used in another script.
+"""
+from __future__ import print_function
+
+import glob
+from babel.messages.pofile import read_po
+
+
+def check(mess, othermess):
+    return mess.id.lower() == othermess.id.lower()
+
+
+def find_duplicates_in_file(filepath):
+    fileobj = open(filepath)
+    catalog = read_po(fileobj)
+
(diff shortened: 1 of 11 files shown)


gasman:
Thanks @chimeno! Unfortunately I don't think it's safe to treat the different-case versions of the strings as equivalent...

screen shot 2017-03-07 at 01 04 44

(In other places, such as table headings, this isn't a problem because we're transforming them to all-capitals in CSS. However, that design decision might conceivably change in future, and we definitely shouldn't hard-code it into the way we handle translations.)

I guess this situation arises because model field `verbose_name`s are lower case as standard, but form field labels are capitalised (often formed automatically from the verbose_name, but not always - e.g. the 'width' / 'height' fields in wagtailimages/forms.py). Perhaps we can work around this with something like `{% trans "title"|capfirst %}`? Not sure if it's really worth it, though...

chimeno:
Thanks for the reply @gasman!
I'm aware this can be quite useless without a convention over all the code, I've changed some strings to ``title`` and ``capfirst``, but feel free to close this PR in case it doesn't worth it.

gasman:
I've now fixed the clearly inconsistent ones in ef5c61ece77a4c65e115c297a67fcf6392078607. I think we're better off skipping the ones that require `title` / `capfirst` - the cognitive load of making sure that we're using the right function in the right places, and maintaining that code in future, is going to be far bigger than the effort of translating a few extra strings.

Thanks again!

PR #3434, opened 2017-03-08
Don't set is_staff flag on users

### LINKED ISSUE #970: Suggestion: set is_staff if moderator

gasman:
Wagtail doesn't routinely require access to Django admin, so it makes sense for Wagtail to not enforce an opinion on who does or doesn't get access.

Fixes #970 and #2777

(I wasn't sure if this really warranted an 'upgrade considerations' note, but technically it is a backwards-incompatible change, and people might conceivably have workflows that rely on the old behaviour, so I figure it's probably worth including.)

### DIFF
--- a/docs/releases/1.10.rst
+++ b/docs/releases/1.10.rst
@@ -74,3 +74,9 @@ Projects using :ref:`custom image models <custom_image_model>` no longer need to
     def image_feature_detection(sender, instance, **kwargs):
         if not instance.has_focal_point():
             instance.set_focal_point(instance.get_suggested_focal_point())
+
+
+Adding / editing users through Wagtail admin no longer sets ``is_staff`` flag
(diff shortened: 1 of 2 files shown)


rosskarchner:
I'm happy to see this fix!

gasman:
Merged in 00c6a5f22039102d6026c469f8b310ceee32eaf8

Issue #663, opened 2014-10-01
In the explorer listing, indicate site root pages with a globe icon


davecranwell:
It's possible to configure a Site where the root page chosen is the absolute tree root, however in this situation the site root isn't a real Page and can't be edited or viewed. Being able to choose "Root" as the site's root page would therefore seem pointless and should be prevented.

In situations where one/more sites are defined as direct _descendants_ of the Root node, the Root Page available here http://localhost:8111/admin/pages/ isn't very pretty. Even exposing the word "root" is a little too much like jargon. It's reasonably easy to navigate to the absolute root, using the "home" icon in the breadcrumb.

As far as I can tell, the root of the tree is only ever an abstract concept and it can only ever contain one/more Sites. It could technically contain Pages that weren't in a site, but they'd never be accessible in a browser. It makes sense to me therefore if the root explorer view were changed to a list of Sites.

The explorer view would include suitable globe icons and the explorer header would also reflect that you were choosing a Site. Before displaying a globe icon, some checking would presumably be necessary to make sure the descendant of root were actually sites (not orphaned pages).

gasman:
Preventing the root from being chosen in page choosers (except for a few special cases where it actually makes sense, e.g. specifying permissions and targets for copy/move) was implemented in #1837. The one remaining item on this ticket is the suggestion of indicating site root pages in the explorer listing (e.g. with a globe icon).

nimasmi:
![screen shot 2017-03-10 at 16 40 30](https://cloud.githubusercontent.com/assets/1473980/23804029/6cde5d9a-05b0-11e7-8e77-6edfac2e586b.png)

![screen shot 2017-03-10 at 16 40 11](https://cloud.githubusercontent.com/assets/1473980/23804021/647e705e-05b0-11e7-9e1e-597a6434c870.png)

gasman:
Implemented in #3439 / #3495

PR #3439, opened 2017-03-10
#663 add icon to site root pages in explorer listing

nimasmi:
Addresses #663.

I've made the icon be a link to the sites index page, rather than the specific site edit page (in case the page is a root of multiple sites). You may think it confusing that it's a link adjacent to the title, which takes you to the page edit view, in which case we could remove the link aspect.


gasman:
Merged in b229953c0f6a8f85a0eeae677397afb700f6d603 via #3495 - thanks @nimasmi and @huubbouma!

PR #3440, opened 2017-03-10
Display non_field_errors in the page editor

### LINKED ISSUE #536: Show NON_FIELD_ERRORS when page does not pass validation

gasman:
Fixes #536

Adds a `validation_error` helper to the `wagtail.wagtailadmin.messages` module to handle form validation errors as a special case of the standard `error` notification type, and update the page add/edit views to use it. The new helper works according to the following logic:

* If all of the form's errors are associated with fields, just show the generic "could not be saved due to validation errors" and leave the specific errors to be displayed inline in the form, as per the current behaviour
* If there are any non-field errors, show all errors (both field and non-field) as a list inside the error message banner. As mentioned in https://github.com/wagtail/wagtail/issues/536#issuecomment-52466885, showing the field errors here (in addition to displaying them alongside the fields) ensures that the non-field errors aren't given extra prominence.

For testing purposes, the new validation added in wagtail/tests/testapp/models.py can also be copied into wagtaildemo's EventPage definition.

(Aside: Django's default `form.errors.as_ul()` rendering of validation errors is rubbish. It displays the internal field names instead of the friendly labels, including `__all__` for non-field errors...)

### DIFF
--- a/wagtail/wagtailadmin/messages.py
+++ b/wagtail/wagtailadmin/messages.py
@@ -6,10 +6,11 @@ from django.template.loader import render_to_string
 from django.utils.html import format_html, format_html_join
 
 
-def render(message, buttons):
+def render(message, buttons, detail=''):
     return render_to_string('wagtailadmin/shared/messages.html', {
         'message': message,
         'buttons': buttons,
+        'detail': detail,
     })
 
 
@@ -37,7 +38,7 @@ def validation_error(request, message, form, buttons=None):
     if not form.non_field_errors():
         # just output the generic "there were validation errors" message, and leave
         # the per-field highlighting to do the rest
-        full_message = message
+        detail = ''
     else:
         # display the full list of field and non-field validation errors
         all_errors = []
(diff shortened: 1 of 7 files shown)


wagtail/wagtailadmin/messages.py
```diff
@@ -31,5 +33,32 @@ def error(request, message, buttons=None):
...
+                    field_label = field_name
+                prefix = "%s: " % field_label
+
+            for error in errors:
+                all_errors.append(prefix + error)
+
+        errors_html = format_html_join('\n', '<li>{}</li>', ((e,) for e in all_errors))
+        full_message = format_html("""{} <ul class="errorlist">{}</ul>""", message, errors_html)
+
+    return messages.error(request, render(full_message, buttons))
```
jjanssen: I am not sure if this will occur in a real-life situation, but the format of the `full_message` for `NON_FIELD_ERRORS` can cause a visual conflict if they also include buttons. At this point it will render the main `ul` and the nested `.errorlist` which causes the buttons to end up at the bottom of the list. 

For visual consistency it would probably make more sense if they are next to the main message followed by the nested `.errorlist`. 

Example from the styleguide:
image

gasman: @jjanssen Good point! While I don't think validation error messages will have buttons in practice, the API allows for that possibility, so it's worth making sure it displays sensibly.

Now fixed in https://github.com/wagtail/wagtail/pull/3440/commits/66c74b7759599d80b2dffffddd069611e3cd9699 by introducing a new 'detail' section to the message template.

jjanssen:
Awesome! Merged in 8b57a46

PR #3446, opened 2017-03-13
Fix userbar transition on Firefox 52

### LINKED ISSUE #3443: Wagtail inpage menu display with firefox 52
### Issue Summary

The wagtail in page menu does not displays with firefox 52, with transitions glitches

### Steps to Reproduce

1. Install firefox 52 (in my case, debian did it to me)
2. Connect and login into a wagtail application 
3. Click on the bird at bottom right of a page 

The menu items does not displays anymore. I think the problems come from the css transitions

### Technical details

* Python version: 2.7.9 (debian stable)
* Django version: 1.10.6
* Wagtail version: 1.9
* Browser version: http://whatsmybrowser.org/b/V24FOR9

gasman:
Fixes #3443. Make the transition animation happen slightly after visibility has been set, so that it doesn't revert to the original invisible style after the transition has completed.

Really hoping that someone can find a less hacky solution than this, because this is horrible :-( The only way I could find to prevent the list items from jumping back to the original `visibility: hidden` style was to trigger the animation a few milliseconds after the `visibility: visibility` style is applied. 50ms seems to be enough to consistently work on my machine - 10ms isn't enough.

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/userbar.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/userbar.js
@@ -4,7 +4,8 @@ document.addEventListener('DOMContentLoaded', function userBar(e) {
     var userbar = document.querySelector('[data-wagtail-userbar]');
     var trigger = userbar.querySelector('[data-wagtail-userbar-trigger]');
     var list = userbar.querySelector('.wagtail-userbar-items');
-    var className = 'is-active';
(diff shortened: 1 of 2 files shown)


gasman:
Aha! While trying to narrow the bug down further to report it to Firefox (only to find it's already fixed in the latest nightly), I found it only occurs when at least one of the animated properties is declared as `!important`... which seems to be redundant here. A better PR coming right up...

PR #3448, opened 2017-03-13
Use field prep_db_value for raw query expansion

thenewguy:
Since the user model can be swapped out the primary key cannot be naively converted to string this way.  Call db_prep_value first.

~~It would be better to use `User._meta.pk.name` instead of 'id' but that assumption is already made in the code so I didn't take the extra level of abstraction~~ Swapped to this because tests failed complaining the custom user in your test didn't have an 'id' field. This raw sql assumes the pk field is 'id' though


gasman:
Thanks @thenewguy! Please can you add a test case covering this fix (i.e. one that fails when the fix isn't in place)?

thenewguy:
@gasman sure... Would you be okay with me changing the identifier field at `https://github.com/wagtail/wagtail/blob/master/wagtail/tests/customuser/models.py#L36` from an AutoField to an integer based field that requires the conversion?

thenewguy:
It is kind of a difficult test to apply to one view I think

gasman:
Yep, that sounds fine. The `CustomUser` model in the tests is designed to be a kind of "worst case scenario", so it's a good place for any edge-cases like this.

gasman:
From the failing tests it looks like there's more work required to make non-standard-typed PKs work, particularly when saving page revisions (which involves serializing to json) - please can you update?

thenewguy:
@gasman Older (very-ish) versions of wagtail work. I was in the process of updating when I hit this and I haven't had a chance to work on it more today.  I had filed an issue before about the json issue a long time ago but in my particular case it was easy to make the value serialize to json so I haven't hit any json problems

thenewguy:
Tests pass now

thenewguy:
I am happy with this now.  Let me know your thoughts

thenewguy:
@gasman ping

wagtail/wagtailadmin/views/home.py
```diff
@@ -10,6 +12,11 @@
 from wagtail.wagtailcore.models import Page, PageRevision, UserPagePermissionsProxy
 
 
+User = get_user_model()
+pk_field_name = User._meta.pk.name
+pk_field = User._meta.get_field(pk_field_name)
```
gasman: Do you need to go through `pk_field_name` here, or will `pk_field = User._meta.pk` do the job?

(Also, since the variable only gets used a bit further down the script, we should give it a more informative name, like `user_pk_field`)

thenewguy: I used `get_field()` because that is the documented api.  I can use `User._meta.pk` directly if you prefer. I'll push the update once the merge tests are successful

thenewguy:
Okay... pushed that change.  Tests appear as if they will pass.  Let me know if this is good to go?

gasman:
Merged in 77c00ee77e6cbbe9318a1f4ff27b5d0ab8a97e30 - thanks!

PR #3461, opened 2017-03-17
Make installing deps more stable

do3cc:
The old syntax works with bash, but not with zsh. The new syntax works with both.


gasman:
Merged in 50a16c37db6cfc52e99f61c4f2e922d876e2d087 - thanks @do3cc!

PR #3462, opened 2017-03-17
Allow custom field classes to be used in api_fields

kaedroho:
Like #3361 this solves the issue of allowing image renditions in the API. But this version does it in a way that doesn't require overhauling the API module so it was much easier to implement and should be easier to test, document and review as well.

The change is quite simple, we now allow ``api_fields`` to contain two item tuples containing the field name and field definition. The definition can use any DRF field, serializer or custom field.

This PR also turns the internal ``ImageRenditionField`` into a public API so it can be used in model definitions.

Example:

```python
from wagtail.wagtailimages.api.fields import ImageRenditionField

class BlogPage(Page):
    body = models.RichTextField()
    feed_image = models.ForeignKey('wagtailimages.Image')
    ...

    api_fields = [
        'body',
        'feed_image',

        # Field with custom serialiser
        ('feed_image_thumbnail', ImageRenditionField('fill-300x300', source='feed_image')),
    ]
```

### DIFF
--- a/docs/advanced_topics/api/v2/configuration.rst
+++ b/docs/advanced_topics/api/v2/configuration.rst
@@ -111,11 +111,15 @@ For example:
 
     # blog/models.py
 
+    from wagtail.api import APIField
+
     class BlogPageAuthor(Orderable):
         page = models.ForeignKey('blog.BlogPage', related_name='authors')
         name = models.CharField(max_length=255)
 
-        api_fields = ['name']
+        api_fields = [
+            APIField('name'),
+        ]
 
 
     class BlogPage(Page):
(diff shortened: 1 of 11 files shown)


BertrandBordage:
I’m not a huge of having a list of both strings and tuples like this: it’s more complex to remember, and a bit misleading for beginners.
And it adds complexity to the code base: you have to add a lot of checks and conversions within the code (as we can see in your pull request). Unfortunately, it appears that there is still some checks/conversions required since Travis fails with `TypeError: hasattr(): attribute name must be string`.

In my opinion, a better API would be to use objects like for `content_panels`. Something like this:
```python
api_fields = [
    APIField('body'),
    APIField('feed_image'),
    APIField('feed_image_thumbnail', serializer=ImageRenditionField(…))
]
```

And of course, for backwards compatibility and most use case, we allow strings, but we convert them to `APIField` when creating the serializer.

Apart from this suggestion, your pull request seems OK. There’s just that last conversion missing that makes the test suite fail.

kaedroho:
Oh yes, that does look much better and is more consistent with search_fields/content_panels.

Thanks!

kaedroho:
That change has now been made, thanks!

kaedroho:
@BertrandBordage Done, thanks! I did make a minor tweak, instead of expecting cls.body_fields/cls.meta_fields to be lists of ``APIField`` objects I've made it convert those too as there are people who have subclassed the endpoint to alter those lists.

BertrandBordage:
The test suite still fails, but I’m OK with your latest changes :)

kaedroho:
Fixed tests (hopefully this time) and added docs. Ready for review!

PR #3468, opened 2017-03-20
[FIX] Some things are not indexed within the update_index command.

shredding:
I have investigated why some random images do not appear as search results in our backend and tracked it down to this line.

The reason is that there is no default order on the database queryset and since it get's retriggered on each index-chunk, the results are somewhat random and do skip some images (or documents).

Adding a `order_by` clause fixes that.


shredding:
... regarding the failing build, I guess that's not my fault? I haven't touched the failing `scss` file.

gasman:
Thanks @shredding!

The Drone server was broken for a while, and it's quite probable that some style guide violations sneaked into the codebase while they weren't being tested - I'll get that sorted.

gasman:
Merged in 9257a4a1dc556b2c9b333f26f36ea41229dff755 - thanks!

PR #3470, opened 2017-03-21
wagtailadmin.views.pages.set_page_position to work with specific pages

### LINKED ISSUE #3469: wagtailadmin explorer to use page.specific in set_page_position
### Issue Summary

I'm trying to override default ordering for some of my custom pages so that latest created ones would be shown first by default (`-path`).
I would also like to make explorer to show them in the same way and handle manual ordering properly for them. 

I'm currently stuck at `wagtailadmin.views.pages.set_page_position`, which doesn't take into account overridden methods of a specific page, even though some sibling functions do (e.g. `move_confirm` does).

Basically, I just need `page_to_move` to be a `specific` model so that I could override its `move` method.

That's where it all happens:
https://github.com/wagtail/wagtail/blob/stable/1.9.x/wagtail/wagtailadmin/views/pages.py#L768

### Technical details

* Wagtail version: `1.9`

dnsl48:
Fixes #3469

### DIFF
--- a/wagtail/wagtailadmin/views/pages.py
+++ b/wagtail/wagtailadmin/views/pages.py
@@ -784,7 +784,7 @@ def set_page_position(request, page_to_move_id):
         position_page = None
         if position is not None:
             try:
-                position_page = parent_page.get_children()[int(position)]
+                position_page = parent_page.get_children()[int(position)].specific
             except IndexError:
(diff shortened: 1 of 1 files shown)


gasman:
Closing as per https://github.com/wagtail/wagtail/issues/3469#issuecomment-288374608.

PR #3471, opened 2017-03-21
[WIP] Allow to use panels in a custom document model

m1kola:
This PR will allow to use edit handlers / panels in a custom document model in a similar way to page models or snippet models.

This may be useful when you need to associate an image with a document. At the moment, you can do this, but you will not have pretty image chooser.

Another use case: you want to add additional fields (let's say many to many fields) to custom document model and want to override form widget. Right now, there is no way to override widgets.

TODO:
- [ ] Fix todos in code
- [ ] Render `edit_handler` instead of `form` in document views
  - [ ] Allow tabbed interface on the CRUD views
  - [ ] Display file field using a custom layout
  - [ ] Display save / delete buttons using drop-down
- [ ] Add tests

### DIFF
--- a/wagtail/wagtaildocs/forms.py
+++ b/wagtail/wagtaildocs/forms.py
@@ -6,7 +6,8 @@ from django.utils.lru_cache import lru_cache
 from django.utils.translation import ugettext_lazy as _
 
 from wagtail.wagtailadmin import widgets
-from wagtail.wagtailadmin.edit_handlers import extract_panel_definitions_from_model_class, ObjectList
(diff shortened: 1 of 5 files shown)


timheap:
I also came across this problem. A site needs two rich text fields and an image chooser on its documents (don't ask), and in the current admin this is simply not possible. The solution for us was to remove the 'Documents' link from the admin menu using the `construct_admin_menu` hook, then make a new Documents admin using ModelAdmin. Problem ~solved~ avoided!

m1kola:
Closing this as it's very outdated

PR #3472, opened 2017-03-21
Link to the full Elasticsearch setup docs from the Performance page

gasman:
The brief Elasticsearch setup instructions given on the Performance page are difficult for a new user to follow, because they don't go into enough detail to cover the typical gotchas (e.g. the need to match ES and Python library versions, and the fact that you don't literally type `{{ project_name }}`) - see http://stackoverflow.com/questions/42841110/unable-to-set-up-elasticsearch-for-wagtail. We should point to the full docs instead.


jjanssen:
LGTM! Merged in eb9cc63

Issue #510, opened 2014-07-25
Better user help for PageChooserPanel restricted types 


davecranwell:
When a PageChooserPanel has been set to only allow a certain type of page to be chosen, the UI in the edit page should warn the user of this _before_ they open the modal.

A PR is currently in flight to better explain this restriction once the modal is open, but it might be better if the standard "Choose a page" button said instead "Choose [type]". Failing that an enforced help message stating "Only [type] pages can be chosen" would also do.

gasman:
Fixed in #3473

PR #3473, opened 2017-03-22
#510 Add Page type to PageChooser button label

cho-leukeleu:
This should fix https://github.com/wagtail/wagtail/issues/510

As the original issue did not take into account that it is possible to specify more than one page type I chose to put all page types in parenthesis (which can actually lead to a pretty large button I guess :/).

### DIFF
--- a/wagtail/wagtailadmin/widgets.py
+++ b/wagtail/wagtailadmin/widgets.py
@@ -138,7 +138,10 @@ class AdminPageChooser(AdminChooser):
         super(AdminPageChooser, self).__init__(**kwargs)
 
         if target_models:
-            self.choose_one_text += ' (' + ', '.join([model._meta.verbose_name.title() for model in target_models]) + ')'
+            models = ', '.join([model._meta.verbose_name.title() for model in target_models if model._meta.verbose_name.title() != 'Page'])
+            if models:
+                self.choose_one_text += ' (' + models + ')'
+
         self.target_models = list(target_models or [Page])
         self.can_choose_root = can_choose_root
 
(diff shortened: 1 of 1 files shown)


wagtail/wagtailadmin/widgets.py
```diff
@@ -137,6 +137,8 @@ class AdminPageChooser(AdminChooser):
     def __init__(self, target_models=None, can_choose_root=False, **kwargs):
         super(AdminPageChooser, self).__init__(**kwargs)
 
+        if target_models:
+            self.choose_one_text += ' (' + ', '.join([model._meta.model_name for model in target_models]) + ')'
```
BertrandBordage: Could you use `verbose_name` instead of `model_name`, please? :)
I think `model_name` is not localized.

cho-leukeleu: Done :). Thanks for reviewing!

wagtail/wagtailadmin/widgets.py
```diff
@@ -137,6 +137,9 @@ class AdminPageChooser(AdminChooser):
     def __init__(self, target_models=None, can_choose_root=False, **kwargs):
         super(AdminPageChooser, self).__init__(**kwargs)
 
+        if target_models:
+            # import pdb; pdb.set_trace()
```
robmoorman: Please remove this comment

BertrandBordage:
So in fact target_model is never `None`? Maybe it should become a mandatory argument, then.

gasman:
@BertrandBordage For a `PageChooserPanel` it's never `None`, because it picks up the page type from the foreign key - however, there are other places where `AdminPageChooser` is used (such as `PageChooserBlock` in a StreamField) and for those, `target_model` might be omitted.

BertrandBordage:
Ah, ok, thanks for explaining.

gasman:
Merged in 67ff4c26df162a3a7de3d46a5f826874e76ff8e5 (with a small tweak to ensure that the `Page` special case still works when other translations are active) - thanks @cho-leukeleu!

Issue #3433, opened 2017-03-08
[IE10+] Odd pagination behaviour on users list in wagtail admin 


ababic:
### Issue Summary

In Internet Explorer 10 and 11 (Chrome, Safari and Internet Explorer 9 seems to be fine), we're seeing some strange behaviour when the pagination links are used on the users list in Wagtail's admin area. 

### Steps to Reproduce

1. Log into the wagtail admin area
2. Go to the users list page by clicking the link in the menu **Settings > Users**
3. Scroll to the bottom of the first page and click on the **Next** link in the bottom right
4. The pagination link takes you to the correct URL, and the correct users flash up for a split second, but then something kicks in and redirects you to a page without the page number in the querystring (so back to the first page).

The expected behaviour would be that you stay on page 2, and not be redirected away from it.

I'm guessing it has something to do with search feature prematurely initiating a search, as the search input is focussed on page load.

### Technical details

* Django version: 1.10.4
* Wagtail version: 1.8.1
* Browser version: Issue has been confirmed in Internet Explorer 10 and 11 (on different windows versions)

ralphjacobs:
Problem found!

core.js ->         // auto focus on search box
        $(window.headerSearch.termInput).trigger('focus');
triggers a         $(window.headerSearch.termInput).on('input', function() {
which triggers a search in IE10 + IE11.

https://connect.microsoft.com/IE/feedback/details/810538/ie-11-fires-input-event-on-focus

Now the solution :)

jjanssen:
Merged in cf815e8 with PR #3476

PR #3476, opened 2017-03-22
Change event handler to keypress. This fixes the IE10 + IE11 submit i…

ralphjacobs:
Change oninput handler to keypress. This fixes the IE10 + IE11 pagination issue. 
Refs: #3433

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
@@ -174,7 +174,7 @@ $(function() {
         var searchCurrentIndex = 0;
         var searchNextIndex = 0;
 
-        $(window.headerSearch.termInput).on('keypress', function() {
+        $(window.headerSearch.termInput).on('keypress cut paste', function() {
             clearTimeout($.data(this, 'timer'));
             var wait = setTimeout(search, 200);
             $(this).data('timer', wait);
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
@@ -174,7 +174,7 @@ $(function() {
         var searchCurrentIndex = 0;
         var searchNextIndex = 0;
 
-        $(window.headerSearch.termInput).on('input', function() {
+        $(window.headerSearch.termInput).on('keypress', function() {
             clearTimeout($.data(this, 'timer'));
             var wait = setTimeout(search, 200);
             $(this).data('timer', wait);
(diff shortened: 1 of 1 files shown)


ababic:
Hey there @ralphjacobs. It looks as though some unrelated changes have been committed along with this PR (Various icon class / gravatar related changes). Would you be able to revert those so that this PR only contains changes relating to the IE10 issue?

thijskramer:
Good point. pasting with something else than the keyboard won't work with `keypress`. 
I think `change` won't work either, because the `onchange` event fires _after_ losing focus (see [MDN docs](https://developer.mozilla.org/en-US/docs/Web/Events/change)), and that's not something we want. 

So to cover both typing and pasting, would listening to both `keypress` and `paste` be enough, with something like `$(window.headerSearch.termInput).on('keypress paste', function() { .... }` ?

ababic:
@thijskramer You're completely right about `onchange` - that will not do. `keypress`, `cut`, and `paste` cover all of the cases I can think of currently, and I can't think of any tidier implementation. But, I would invite a more front-end focussed team member to comment here, as my JS skills are limited. @jjanssen how do you feel about the multiple trigger approach?

gasman:
Listening to events that assume a specific input device (keyboard) is raising accessibility alarm bells for me, although I don't know whether that's a real concern or just a theoretical one.

We encountered something similar before, where the initial focus on the search box was triggering a search: https://github.com/wagtail/wagtail/commit/a2fe305aa0342e332931265b5f9b3d44585371bf - in this case, I fixed it by ignoring searches for the empty string. Maybe there's something similar we can do here?

ralphjacobs:
We cleaned up the commits and added more handlers

thijskramer:
@gasman I think that listening to `keypress`, `cut` and `paste` are sufficient for the use case, namely the search-as-you-type functionality. Just by using `oninput` raises unexpected behaviour in IE11, as I mentioned before.

ababic:
I thought initially about the changing the search to only launch when the value wasn't blank. But, in a scenario where you typed something to initiate a search, then used the delete key to reduce the search string to nothing, the 'blank' search would fail to initiate, and you'd still see results for the last remaining character. We'd need to introduce something else to indicate whether or not the user has actually interacted with the input or not - which could only really be done by adding additional listeners, calling a separate method. And by that point, it's more complicated than it needs to be.

ralphjacobs:
At least this reduces multiple requests everytime using IE10 or IE11. Even when you visit the user listing the first time it would generate two requests because the event was fired.

ababic:
I'm happy with this one now, if the core team are?

ralphjacobs:
@ababic thanks

gasman:
Yep, sounds good to me. (Haven't tested it myself, but I'm satisfied that the simple solution is the way to go here, rather than an over-engineered one :-) )

jjanssen:
Reviewing it right now :)

ralphjacobs:
I will look in to it

jjanssen:
Thanks! Merged in cf815e8

PR #3477, opened 2017-03-22
Wagtail issue 3281 fixed by checking slug's unique id. Test for Valid…

helmantel:
…ationError no longer needed.


wagtail/wagtailcore/tests/test_page_model.py
```diff
@@ -67,13 +67,6 @@ def test_slug_is_autogenerated(self):
         retrieved_page = Page.objects.get(id=events_page.id)
         self.assertEqual(retrieved_page.slug, 'events-2')
 
-    def test_slug_must_be_unique_within_parent(self):
```
tomdyson: don't we still want this test?

helmantel: I don't think so. A ValidationError can no longer occur since we always generate a unique slug.

gasman: If we're changing the behaviour from "a duplicate slug produces a ValidationError" to "a duplicate slug is silently changed into a non-duplicate one" (and, to be honest, I'm not sure that we _should_ do that...), then we need a new test to confirm that the page really is saved successfully.

helmantel: The behaviour of moving an item to a folder where a slug with the same id already exists is now sort of similar to creating a new item within a folder that already contains that slug. I agree with you that both aren't the nicest of solutions, but they only occur when creating new or moving around existing content. 

I could create a 'saved successfully' kind of test and maybe think about a more elegant solution when silently adjusting the slug's id. For now, this fix does solve visible stacktrace problem :)

Issue #3235, opened 2016-12-20
Display of multiple-select fields in form submissions not user friendly


kaedroho:
### Issue Summary

The values of a "checkboxes" field is displaying in Python format in the admin.

For example, a user has selected "Blue", "Yellow" and "Red" on a checkboxes field. When viewing this form submission in the admin, this is displayed as ``[u'Blue', u'Yellow', u'Red']``.

### Steps to Reproduce

1. Create a form page with a checkboxes field and give the field some options
2. As a frontend user, submit the form with some of these options checked
3. In the admin, go to Forms, and click on the form page. You should see your submission for that page
4. Look at the value in the column of the checkbox

### Expected result

I expect the values the user has selected to be concatenated with a comma (or something more i18n friendly).

### Technical details

* Python version: 2.7.12
* Django version: 1.9.9
* Wagtail version: 1.9a0 (from yesterday)
* Browser version: Firefox 48

gasman:
Fixed in #3479

PR #3479, opened 2017-03-22
Display a comma separated string for fields returning content as lists

cho-leukeleu:
Addresses #3235


gasman:
Merged in 78a2e8ae968ebd156b3cabe557e115245ae07813 (with some additions - added tests, and applied the new format to CSV export too). Thanks @cho-leukeleu!

Issue #3291, opened 2017-01-23
Default avatar is visible when using a transparent gravatar image


jaap3:
Not really a big issue, but when you have a partially transparent gravatar img parts of the fallback avatar are visible: 

![screen shot 2017-01-23 at 17 15 55](https://cloud.githubusercontent.com/assets/48517/22212213/16939b38-e190-11e6-8fd4-2201549d60c8.png)

jjanssen:
Merged in 20e8aab

PR #3424, opened 2017-03-04
[WIP] 319 make gravatar optional

chimeno:
This should fix #319 , #3291 and #1289.

Is not yet finished, but I would like to share the approach before working more on it.

1. Add a user preference to persist the avatar backend
2. Replace the current (CSS) default avatar with a .svg one static served.
3. Allow user to upload their own avatar and server by /media/

Notes:
- Default to gravatar to be backward-compatible

TODO:
- [x] Make tests
- [ ] Allow a setting string to prohibit the user of gravatar ¿?
- [ ] Search or draw a best default .svg
- [ ] Change the template tag in all places and remove ``gravatar_url``
- [ ] Remove not used css
- [x] Make upload another custom image remove the previous one uploaded by the user.
- [ ] Docs
 
CURRENT PREVIEW:
image

image

As always, any feedback is  welcome!

PR #3481, opened 2017-03-22
Avatar transparency fix

thijskramer:
Refs: #3291

Gravatar returns a blank transparent image by default if the gravatar it self does not exists.
We changed the default avatar URL to mystery man (mm).

This looks like this: 

before-after1
before-after2

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_formatters.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_formatters.scss
@@ -1,4 +1,4 @@
-// user avatars, with fallback to placeholder graphic if no gravatar exists
+// user avatars
 .avatar {
     border-radius: 100%;
     position: relative;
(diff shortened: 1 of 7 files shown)


gasman:
Related: #3424

Probably a very minor point, but this change means we're effectively "baking in" a design element from Gravatar into the Wagtail admin, which will get reverted to the old design if Gravatar integration is disabled as per #3424 - is that an issue?

thijskramer:
That _could_ be an issue, but this fix fixes purely issue #3291. I think that we shouldn't feel blocked by another PR that is still a work in progress.

jjanssen:
Yeah we're baking in this for now, but it is not going to be an issue for the progress in the PR.

In the PR there is already a default fallback image in-place if no picture was uploaded: https://github.com/chimeno/wagtail/blob/0db058f1328d4252f30bc9606fba8b473c3967a4/wagtail/wagtailadmin/static_src/wagtailadmin/images/default-user-avatar.svg

And @thijskramer secretly already processed one of the chores on the list: 'Remove not used css'

thijskramer:
@jjanssen What are the changes that you are requesting?

jjanssen:
@thijskramer those were the remarks I've added yesterday. I'm taking another look at it today.

jjanssen:
Thanks! Merged in 20e8aab

PR #3485, opened 2017-03-22
Dropped support for generating static sites using django-medusa

robmoorman:
As the django-medusa contrib module is no longer maintained and not compatible anymore since django 1.8, this can be removed from the wagtail codebase.

An alternative library instead is mentioned in the release docs for 1.10 https://github.com/moorinteractive/wagtail-bakery. We will be pushing hard to get this module battle tested (and maybe later integrated again as a staticsitegen contrib module in the core).

Until then, the documentation leftovers for django-medusa can be removed.

One thing is left intact (as we probably going to need it again), is the `get_static_site_paths` method on the `Page` model.


tomdyson:
LGTM

gasman:
Merged in c7b778c4e9b35ab3a88c94a4c38bb4498d709a0a - thanks @robmoorman!

(Have moved the release notes a bit further down, as this is less of a headline item than dropping Django 1.9)

Issue #2657, opened 2016-05-31
Add flag to exclude page from the sitemap


dave-hockly:
Would it be possible to add a checkbox field to web pages in order to exclude them from the automatically generated sitemap.xml? Maybe we could display it in the "promote" tab?

jordij:
Some of our clients use public pages for specific purposes and don't want them to be exposed in the sitemap. We could certainly add this feature outside of Wagtail on a project-basis but could make sense to have it in the core. Let us know what do you think and we can have someone from Springload to work on it.

muhammad-saleh:
I guess you can do this by adding this flag in the model and excluding it in the sitemap template

gasman:
It should be possible to set this up on any (or all) of your page models by defining a suitable `get_sitemap_urls` method, for example:

    class MyPage(Page):
        include_in_sitemap = models.BooleanField()

        promote_panels = Page.promote_panels + [FieldPanel('include_in_sitemap')]

        def get_sitemap_urls(self):
            if not self.include_in_sitemap:
                return []
            else:
                return super(MyPage, self).get_sitemap_urls()


The problem with making this into built-in Wagtail functionality is that it would involve adding a field to the core `Page` model - this would make it difficult to disable when the sitemaps module is not in use, as well as being quite messy architecturally (ideally, contrib modules such as `wagtailsitemaps` shouldn't have any special 'privileges' in determining what goes into wagtailcore; in principle, there shouldn't be anything in `wagtail.contrib` that couldn't equally have been written as a third-party add-on package). We're still to come up with a good way of adding global page fields from outside wagtailcore - #836 is probably the nearest suggestion right now.

dave-hockly:
Thanks Matt.

On 3 June 2016 at 03:22, Matt Westcott <notifications@github.com> wrote:

> It should be possible to set this up on any (or all) of your page models
> by defining a suitable get_sitemap_urls method, for example:
>
> class MyPage(Page):
>     include_in_sitemap = models.BooleanField()
>
>     promote_panels = Page.promote_panels + [FieldPanel('include_in_sitemap')]
>
>     def get_sitemap_urls(self):
>         if not self.include_in_sitemap:
>             return []
>         else:
>             return super(MyPage, self).get_sitemap_urls()
>
> The problem with making this into built-in Wagtail functionality is that
> it would involve adding a field to the core Page model - this would make
> it difficult to disable when the sitemaps module is not in use, as well as
[...]

Issue #1196, opened 2015-04-16
XML sitemap for multi-lingual sites.  


tmsndrs:
Google requires a fairly specific output for the XML sitemap on multi-lingual sites https://support.google.com/webmasters/answer/2620865?hl=en

Currently Wagtail isn't able to accommodate this.

gasman:
Interesting one!

The tricky thing here is that our usual(?) approach to multi-lingual Wagtail sites, cross-linking the different languages, isn't a pattern that Wagtail 'natively' understands - it's just a convention that's implemented on that particular site. If Wagtail doesn't know about that convention, then the sitemap module won't know about it either. Perhaps this calls for a 'contrib' module to provide the multi-language stuff that we currently implement on a per-site basis...

tmsndrs:
I think that would be great. The current project we're working on has additional content 'tabs' for each language which unfortunately deviates from core but is, in my eyes the best solution I've seen. If this was the accepted way of dealing with languages as part of core that would be great.

chimeno:
+1 to @tmsndrs. I think content 'tabs' for multilingual projects it's the best approach.
I know this approach would require a 'natively' solution.

SalahAdDin:
:+1: I think same.

dborisog:
Just in case someone will find this useful. I'm using wagtail 1.9.2 and wagtail-modeltranslation 0.5.0a0

I copied 'wagtail.contrib.wagtailsitemaps' as 'sitemaps' app into a wagtail project and modified two files to 'sitemap.xml' and 'sitemap_generator.py' in order to satisfy google best practices https://support.google.com/webmasters/answer/2620865?hl=en

**sitemap_generator.py**, modifications only
```python

    def get_furl(self, site_root, lang, page):
        furl = site_root+\
               '/'+\
               lang[0]+\
               getattr(page.specific, 'url_path_'+lang[0])
        return furl

    def get_alternatives(self, page, site_root, languages):
        alternatives_list = []
        for lang in languages:
            alternative_url = str()
            alternative_url = self.get_furl(site_root, lang, page)
            alternative = {
                'language': lang[0],
                'href': alternative_url
            }
            alternatives_list.append(alternative)
        return alternatives_list

    def get_urls(self):
        page_list = []
        site_root = self.site.root_url
        pages = self.get_pages()
        for lang in settings.LANGUAGES:
            for page in pages:
                page_list.append({
                    'location': self.get_furl(site_root, lang, page),
[...]

Issue #833, opened 2014-11-25
Exclude arbitrary page from sitemap via dashboard


thenewguy:
### Problem statement

It would be handy to allow staff to exclude an arbitrary page from the sitemap via the dashboard.

### Proposed solution

For some pages, it is difficult to determine that a page should be hidden or shown in the sitemap and it would be nice if the page model had a boolean "Exclude from sitemap" field for this.

### Alternative solutions

Currently the only approach for this that I can find mentioned in the docs is overriding the `get_sitemap_urls`. If you do not want to add a field to the Page model, it could be implemented in the sitemap contrib module without touching the Page model... but it would be more convoluted to create a snippet model to control `SitemapOptions` and autoregister an edit handler for displaying an InlinePanel edit handler than it would be to add a boolean field to the page model.

### Additional context

N/A

### Working on this

Contribute to the discussion to help this move forward. Explain your use cases, discuss possible solutions, research how to adapt existing code, or compatibility concerns.

Once others agree with the proposed changes and the issue is no longer labelled "needs design decision", anyone can contribute to this. View our [contributing guidelines](https://docs.wagtail.org/en/latest/contributing/index.html). Add a comment to the issue to explain what approach you are taking.

lb-:
Adding clarification, the overriding of `get_sitemap_urls` works and was resolved in #3504 as of Wagtail 1.10.

The enhancement of disabling an arbitrary page in the dashboard has not been implemented and needs a decision if this is something that should be, or leave it to individual developers to implement as they see fit.

thibaudcolas:
I’ve reworded the initial request a bit, and tagging as "Needs design decision" like #13728. It’s a niche need. I expect it might be a good fit anyway, as this aspect of sitemap generation is managed directly via Wagtail. But I think we could use confirmation from others that it’s worthwhile to implement.

It could also be worthwhile to implement as a broader field, "Exclude from search"? Which developers could use to add a `<meta name="robots" content="noindex">` to their pages? That’s a relatively common requirement for live but still in-progress content. Or implement this as a new page privacy setting? "Public, excluded from search"?

Example from our official news template: `appear_in_search_results`: [news-template BasePage](https://github.com/wagtail/news-template/blob/6a88fbf63476136392c89ff8cab4d609b9c71d01/project_name/utils/models.py#L294)

```python
    appear_in_search_results = models.BooleanField(
        default=True,
        help_text="Make this page available for indexing by search engines."
[...]

PR #3489, opened 2017-03-23
Extendable sitemap generator

mikedingjan:
Allow the generator class of the ``wagtail.contrib.wagtailsitemaps`` module to be customized. A customized generator class allows the user to implement their own sitemap.xml generator which makes it possible to create for example a multi-language sitemap. Also excluding pages from the sitemap should be easier.

Solves: #833 , #2657 and #1196

### DIFF
--- a/docs/reference/contrib/sitemaps.rst
+++ b/docs/reference/contrib/sitemaps.rst
@@ -87,7 +87,7 @@ Each dictionary can contain the following:
 You can add more but you will need to override the ``wagtailsitemaps/sitemap.xml`` template in order for them to be displayed in the sitemap, extending the ``Sitemap`` generator is also an option.
 
 .. note::
-    As of Wagtail 1.10 it's possible to extend te existing ``Sitemap`` generator class. this can be done by setting ``WAGTAILSITEMAPS_GENERATOR`` to your own implementation of the ``Sitemap`` object. This can be used to implement multi language sitemaps for example.
+    As of Wagtail 1.10 it's possible to extend the existing ``Sitemap`` generator class. this can be done by setting ``WAGTAILSITEMAPS_GENERATOR`` to your own implementation of the ``Sitemap`` object. This can be done used to implement multi language sitemaps, for example.
 
 
 Cache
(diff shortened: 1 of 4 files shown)


docs/reference/contrib/sitemaps.rst
```diff
@@ -84,7 +84,10 @@ Each dictionary can contain the following:
  - **changefreq**
  - **priority**
 
-You can add more but you will need to override the ``wagtailsitemaps/sitemap.xml`` template in order for them to be displayed in the sitemap.
+You can add more but you will need to override the ``wagtailsitemaps/sitemap.xml`` template in order for them to be displayed in the sitemap, extending the ``Sitemap`` generator is also an option.
+
+.. note::
+    As of Wagtail 1.10 it's possible to extend te existing ``Sitemap`` generator class. this can be done by setting ``WAGTAILSITEMAPS_GENERATOR`` to your own implementation of the ``Sitemap`` object. This can be used to implement multi language sitemaps for example.
```
tomdyson: te existing -> the existing
this can be done -> This can be done
sitemaps for example -> sitemaps, for example

mikedingjan: Thanks, fixed!

wagtail/contrib/wagtailsitemaps/views.py
```diff
@@ -3,15 +3,29 @@
...
 
-from .sitemap_generator import Sitemap
+DEFAULT_GENERATOR = 'wagtail.contrib.wagtailsitemaps.sitemap_generator.Sitemap'
+
+
+def get_generator_class():
+    generator_class = getattr(settings, 'WAGTAILSITEMAPS_GENERATOR', DEFAULT_GENERATOR)
+    try:
+        Sitemap = import_string(generator_class)
+    except ImportError:
```
robmoorman: I wouldn't mind the ImportError it thrown, as it's clear for the develop that the WAGTAILSITEMAPS_GENERATOR is invalid (to prevent looking at the 'same' sitemap)

mikedingjan: True, makes things more clear, I'll remove the `try except`

mikedingjan:
This PR might not longer be necessary, since we've refactored the sitemap functionality in PR #3504

gasman:
Closing in favour of #3504 as that seems like a cleaner and more flexible approach. (Can reopen this later if we decided not to go with #3504, of course)

Issue #3281, opened 2017-01-18
Crash when moving a page from one parent to another where there is an already existing page with the same slug


pyMan:
### Issue Summary

Hi,
I get a crash when moving a page from a parent to another one where there is already a page with the same slug.

### Steps to Reproduce

1) Create a page P1 with a slug "test"
2) Create another page P2 with same slug under another parent
3) Move P1 onto P2's parent

(traceback wagtaildemo 1.4.1)
```
Traceback:
File "C:\test\wagtaildemo\lib\site-packages\django\core\handlers\base.py" in get_response
  132.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "C:\test\wagtaildemo\lib\site-packages\django\views\decorators\cache.py" in _cache_controlled
  43.             response = viewfunc(request, *args, **kw)
File "C:\test\wagtaildemo\lib\site-packages\wagtail\wagtailadmin\decorators.py" in decorated_view
  22.             return view_func(request, *args, **kwargs)
File "C:\test\wagtaildemo\lib\site-packages\wagtail\wagtailadmin\views\pages.py" in move_confirm
  649.         page_to_move.move(destination, pos='last-child')
File "C:\test\wagtaildemo\lib\site-packages\django\utils\decorators.py" in inner
  145.                     return func(*args, **kwargs)
File "C:\test\wagtaildemo\lib\site-packages\wagtail\wagtailcore\models.py" in move
  1055.         new_self.save()
File "C:\test\wagtaildemo\lib\site-packages\django\utils\decorators.py" in inner
  145.                     return func(*args, **kwargs)
File "C:\test\wagtaildemo\lib\site-packages\wagtail\wagtailcore\models.py" in save
  445.         self.full_clean()
File "C:\test\wagtaildemo\lib\site-packages\wagtail\wagtailcore\models.py" in full_clean
  436.         super(Page, self).full_clean(*args, **kwargs)
File "C:\test\wagtaildemo\lib\site-packages\django\db\models\base.py" in full_clean
  1171.             raise ValidationError(errors)

Exception Type: ValidationError at /admin/pages/24/move/6/confirm/
Exception Value: {u'slug': [u'This slug is already in use']}
```

### Technical details

* Python version: 2.7
* Django version: 1.8

[...]

PR #3490, opened 2017-03-23
#3281 Added _slug_is_available check to can_move_to method

helmantel:
A more elegant solution to the problem mentioned in ticket #3281. The can_move_to method now checks whether the slug is still available at the designated target.


gasman:
I'm not sure that the permission logic is the right place for this check. I can see that it does the job functionally, but it feels like a bit of a hack - a negative response from the permission checker is meant to indicate "you don't have privileges to perform this action", but here the action is being rejected for a reason that's not related to privileges. I'd be happier if this was implemented as an explicit check within the 'move page' view (and even happier if we gave the user the ability to specify a new slug in this situation... as I think we do for the 'copy page' view).

gasman:
Fixed via alternative PR #5138.

Issue #2867, opened 2016-07-25
403 when user tries to copy a page to a place where they don't have permissions


chrxr:
Steps to replicate:
- Create a group that has limited page permissions
- Add user to that group
- Try to copy a page so that the new parent page is one that the user does not have permissions to edit.

Result:
- Plain 403 error page

gasman:
Will be fixed by #2666, I believe.

Henk-JanVanHasselaar:
I'll be having a look at this during the Arnhem sprint.

gasman:
Fixed in #3492

PR #3492, opened 2017-03-23
copy page validation

Henk-JanVanHasselaar:
This PR will: 
- validate the selected page with the user permissions when the user tries to move copy a page. (fix issue #2867)
- Disable pages in page chooser panel the user does not have permission for when copying page.


HvandenBerg:
Tested OK

wagtail/wagtailadmin/forms.py
```diff
@@ -170,6 +169,12 @@ def clean(self):
         # New parent page given in form or parent of source, if parent_page is empty
         parent_page = cleaned_data.get('new_parent_page') or self.page.get_parent()
 
+        # check if user is allowed to create a page at given location.
+        if not parent_page.permissions_for_user(self.user).can_add_subpage():
+            self._errors['new_parent_page'] = self.error_class(
+                [_("You do not have permission to copy to page \"%s\"" % parent_page)]
```
gasman: `parent_page` should be passed from outside of the `_(...)` call, as per the example at https://docs.djangoproject.com/en/1.11/topics/i18n/translation/#standard-translation

wagtail/wagtailadmin/static_src/wagtailadmin/js/page-chooser.js
```diff
@@ -1,4 +1,4 @@
-function createPageChooser(id, pageTypes, openAtParentId, canChooseRoot) {
+function createPageChooser(id, pageTypes, openAtParentId, canChooseRoot, UserPerms) {
```
gasman: This variable should be `userPerms` (with a small `u`) - capitalised identifiers are only for classes / constructors.

gasman:
Merged in 980f979d90efb427a63c7bffe58c0ceda3f305fa - thanks!

Issue #3113, opened 2016-10-26
Drop downs in date / datetime widget are unusable on Macbook's touchpad


m1kola:
It' impossible to change month or year in date / datetime pickers on Macbook's touchpad.

See [this demo video](https://www.youtube.com/watch?v=mSvkYfO5QDM).

Can reproduce it with:
- Firefox 49.0.1
- Chrome 54.0.2840.71 (64-bit)
- Safari 9.1 (11601.5.17.1)

gasman:
Confirmed. (Are you on OS X Sierra? I have the impression that the Sierra upgrade has generally broken a lot of scrolling behaviour...)

The version on http://xdsoft.net/jqplugins/datetimepicker/ seems to behave a bit better - https://github.com/xdan/datetimepicker/issues/179 might be the bugfix in question, but I'm not 100% sure. Either way, it's definitely worth us upgrading to the latest version.

m1kola:
Sorry, I've forgotten to add OS version. I use El Capitan at the moment (0.11.4 (15E65)). I'll try on Windows laptop tomorrow.

> The version on http://xdsoft.net/jqplugins/datetimepicker/ seems to behave a bit better

At least in Chrome I see no difference with our version.

MechanisM:
It works for me.
macOS Sierra(10.12.1)
Chrome 56.0.2901.0 canary (64-bit)

m1kola:
I can confirm the same issue on WIndows 7 with 

* Google Chrome 53.0.2785.143 m (64-bit)
* Firefox 49.0.1

Laptop: Acer Aspire V3-772G

But in this case it behaves *a bit* better.

jjanssen:
Fixed for now, QA'ed by @HvandenBerg and merged in 87a7a79

Part of my comment within the PR:
> The upgrade to a more recent version would probably more desirable but would also have more impact. Doing that is probably part of a larger discussion on what we are going to do regarding updates. The general concern we are a falling a bit behind on maintenance whereas browsers are moving forward. That would be something for the team discussion.

@m1kola feel free to the reopen if your are still encountering some issues with this.

PR #3493, opened 2017-03-23
Fix touchpad scrolling on macbook in datetimepicker

ralphjacobs:
- We upgraded the datetimepicker jquery library to a version which doesn't need re-wiring in page-editor.js
- Also no scss changes are needed.
- In the library itself we changed one small variable to make the scrolling even more user-friendlier on touchpads.

Refs: #3113

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/vendor/jquery.datetimepicker.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/vendor/jquery.datetimepicker.js
@@ -440,7 +440,7 @@
 				timeboxparent.on('mousewheel', function (event) {
 					var top = Math.abs(parseInt(timebox.css('marginTop'), 10));
 
-					top = top - (event.deltaY * 20);
+					top = top - (event.deltaY * 1);
 					if (top < 0) {
 						top = 0;
 					}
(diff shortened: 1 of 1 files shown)


tomdyson:
Thanks @ralphjacobs! Should we use a later / minified version? e.g. https://github.com/xdan/datetimepicker/blob/master/jquery.datetimepicker.min.js

jjanssen:
@hvandenberg can you take this for a testdrive tomorrow?

ralphjacobs:
Hi Tom,

I've tried newer versions, but then the complete design and functionality breaks.
Also the current one wasn't minified.

HvandenBerg:
tested OK, time now scrolls smoother. day/month is still a bit too fast

jjanssen:
Overall I think we are good to go with this one. There are no regressions showing up and it at least fixes the annoying scroll within the time picker. The comment @HvandenBerg mentions is valid but current behaviour. 

The upgrade to a more recent version would probably more desirable but would also have more impact. Doing that is probably part of a larger discussion on what we are going to do regarding updates. The general concern we are a falling a bit behind on maintenance whereas browsers are moving forward. That would be something for the team discussion.

jjanssen:
Thanks! Merged in 87a7a79

PR #3495, opened 2017-03-24
663 site root icon

huubbouma:
This PR extends https://github.com/wagtail/wagtail/pull/3439/
by adding a basic test case to check if the icon is present in the root listing view


gasman:
Merged in b229953c0f6a8f85a0eeae677397afb700f6d603

PR #3496, opened 2017-03-24
Remove custom SQL statements in Page._update_descendant_url_paths

mvantellingen:
Django 1.8 introduced ORM functions for Concat and Substring so use that
instead of custom SQL statements based on the connection vendor.

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -14,8 +14,9 @@ from django.core.exceptions import ValidationError
 from django.core.handlers.base import BaseHandler
 from django.core.handlers.wsgi import WSGIRequest
 from django.core.urlresolvers import reverse
-from django.db import connection, models, transaction
-from django.db.models import Case, IntegerField, Q, When
+from django.db import models, transaction
+from django.db.models import Case, IntegerField, Q, Value, When
+from django.db.models.functions import Concat, Substr
 from django.http import Http404
 from django.template.response import TemplateResponse
 # Must be imported from Django so we get the new implementation of with_metaclass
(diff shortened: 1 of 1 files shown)


mvantellingen:
This might actually remove support from mssql since Django doesn't officially support that.

tomdyson:
> This might actually remove support from mssql since Django doesn't officially support that.

I'm nervous about that. I know of several Wagtail + MSSQL users, including @FlipperPA at Wharton. We really need some CI for MSSQL, e.g. with https://www.appveyor.com.

mvantellingen:
I've added appveyor support to treebeard last tuesday. I'll add it to wagtail next week

FlipperPA:
We have moved to PostgreSQL for my division of Wharton (hallelujah!), with some departments likely following suit in the future. That said, I think SQL Server support is a good thing to offer. Microsoft is officially supporting a Django driver stack for SQL Server: https://blogs.technet.microsoft.com/dataplatforminsider/2016/12/09/sql-server-python-whats-new/ There has been talk from Microsoft about getting SQL Server officially supported by Django, but I don't know where that conversation has led. It should still be considered third party support with a dedicated-but-small team.

IIRC, I've only written two pieces of custom SQL for SQL Server, although others may have. I believe one could be removed by a migration squash in `wagtailimages`, and the other was a `subselect`, which we should be able to remove the `SQL.raw()` for in Django 1.11, which is just a few weeks away.

wagtail/wagtailcore/models.py
```diff
@@ -587,32 +588,12 @@ def check(cls, **kwargs):
...
-                SET url_path = %s || substring(url_path from %s)
-                WHERE path LIKE %s AND id <> %s
-            """
-        cursor.execute(update_statement, [new_url_path, len(old_url_path) + 1, self.path + '%', self.id])
+        (Page.objects
+            .filter(path__startswith=self.path)
+            .exclude(pk=self.pk)
+            .update(url_path=Concat(
+                Value(new_url_path),
+                Substr('url_path', len(old_url_path) + 1))))
```
gasman: @atombrella has helpfully provided an independent implementation of this fix at #3895, and the one effective difference from this is that it has `F('url_path')` instead of `'url_path`. I'm pretty sure `F('url_path')` is correct (otherwise it'd be taking a substring of the literal string `'url_path'`, surely?) but we'll have to double-check this before merging either version...

gasman: OK, the `F` isn't necessary - as per [the Django docs](https://docs.djangoproject.com/en/1.11/ref/models/expressions/#value-expressions):

> You will need to use Value() when you want to pass a string to an expression. Most expressions interpret a string argument as the name of a field, like Lower('name').

So it would have to be `Value('url_path')` to be interpreted as the literal string `'url_path'` - although that wouldn't work anyhow, because apparently Django db functions don't let you construct expressions that don't reference any database fields (it fails with `Cannot resolve expression type, unknown output_field`)...

gasman:
Merged in cf25d0712083ce6eea825ca1d69edf521a97440a, with exception for MSSQL added in 0e38f0b401684a066489e4186b7523a343bb3370 to fall back on the old code.

(@BertrandBordage, I think you're right that the substring clause in the MSSQL query looks dubious. However, since we don't have a testing setup for MSSQL I'm taking the cautious approach and leaving that code path unchanged, so that we aren't blocking the fix for other DBs.)

PR #3499, opened 2017-03-24
Fixes the Django 1.11 ORM compatibility issue.

BertrandBordage:
This pull request fixes the remaining issue blocking 1.11 compatibility (see #3314).
It is due to the changes done in [django/django@1bc249c](https://github.com/django/django/commit/1bc249c2a67c24fcd28436c85388eff1d826e305).

### DIFF
--- a/wagtail/wagtailsearch/backends/elasticsearch.py
+++ b/wagtail/wagtailsearch/backends/elasticsearch.py
@@ -278,7 +278,8 @@ class ElasticsearchSearchQuery(BaseSearchQuery):
         if lookup == 'in':
             return {
                 'terms': {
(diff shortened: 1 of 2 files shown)


kaedroho:
Wouldn't this break cases where the value is not a subquery? For example:

```python
MyModel.objects.filter(category__in=['foo', 'bar']).search()
```

BertrandBordage:
That’s right, we have to add a test for such use cases.
I’ll update the pull request & add a test.

BertrandBordage:
I pushed tests and handled other cases than subqueries.

kaedroho:
Looks good to me, thanks!

Issue #2251, opened 2016-02-18
FileField in an InlineModel not adding enctype="multipart/form-data"


iammattq:
Hey, I noticed #1741 the fix you did for getting the `enctype="multipart/form-data"` to appear. However, it doesn't want to appear when the FileField is part of an InlineModel (like in your BlogRelatedLinks example). Do you think we could get this fixed up?

I've tested adding the FileField to the parent model and it works great. Thanks

bmihelac:
This also applies to new *ModelAdmin*. 

It looks like EditHandler should be extended with `is_multipart` method, instead of checking main form.

gasman:
Fixed in #3501 for the page editor (and separately in the next release of django-modelcluster: https://github.com/wagtail/django-modelcluster/pull/73)

Issue #3435, opened 2017-03-08
Can't upload files to Orderable model with model.FileField


Mat-Scott:
### Issue Summary

I am trying to set up an Orderable Model for a custom setting in wagtail. I am trying to set this up so that I can upload SVG's to the icon field so they can be used in the menu area. When I try to upload a file to this field (any file, not just SVG) it saves the page but does not save the file. I have seen similar Issues concerning this that have been closed and involved adding **enctype="multipart/form-data"** to the `<form>` I tried this and it saved the file.

### Steps to Reproduce

```
from django.db import models

from wagtail.contrib.settings.models import BaseSetting, register_setting
from wagtail.wagtailcore.models import Orderable
from wagtail.wagtailadmin.edit_handlers import FieldPanel, PageChooserPanel, InlinePanel, MultiFieldPanel
from wagtail.wagtailimages.edit_handlers import ImageChooserPanel

from modelcluster.models import ClusterableModel
from modelcluster.fields import ParentalKey


# -----------------------------------Nav
@register_setting
class NavSettings(ClusterableModel, BaseSetting):

    panels = [
        InlinePanel('main_nav_items', label="Main Nav Items"),
    ]

class MainNavItem(Orderable):
    nav_title = models.CharField(max_length=500, blank=True, verbose_name="Nav Item Title")
    nav_text = models.TextField(blank=True, verbose_name='Nav Item Text')
    link_page = models.ForeignKey(
        'wagtailcore.Page',
        null=True,
        blank=True,
        on_delete=models.SET_NULL,
        related_name='+',
        verbose_name="Page Link"
    )
    link_title = models.CharField(max_length=500, blank=True, verbose_name="Link Title Attr")
    icon = models.FileField(upload_to="header_icons")
    icon_alt = models.CharField(max_length=500, blank=True, verbose_name="Icon Alt")
    nav = ParentalKey(NavSettings, related_name='main_nav_items')

    panels = [
        FieldPanel('nav_title'),
        FieldPanel('nav_text'),
        PageChooserPanel('link_page'),
        FieldPanel('link_title'),
        FieldPanel('icon'),
        FieldPanel('icon_alt'),
    ]
```

### Technical details

* Python version: Python3
* Django version: 1.10.6
* Wagtail version: 1.9
* Browser version: Any Browser

gasman:
[...]

PR #3501, opened 2017-03-24
Fixes ticket #2251. Implemented is_multipart on EditHandler

helmantel:
A form's enctype attribute was not set to "multipart/form-data" when an InlinePanel contained a FileInput field. This fix solves the issue.

### DIFF
--- a/wagtail/wagtailadmin/edit_handlers.py
+++ b/wagtail/wagtailadmin/edit_handlers.py
@@ -163,6 +163,19 @@ class EditHandler(object):
         """
         return ""
 
+    def is_multipart(self):
+        """ Checks form's *and* formsets' is_multipart method """
+        multipart = False
+        if self.form.is_multipart():
+            multipart = True
+        else:
+            for k in self.form.formsets.keys():
+                if self.form.formsets[k].is_multipart():
+                    multipart = True
+                    break
+
+        return multipart
+
     def render_as_object(self):
         """
         Render this object as it should appear within an ObjectList. Should not
--- a/wagtail/wagtailadmin/templates/wagtailadmin/pages/create.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/pages/create.html
@@ -18,7 +18,7 @@
         </div>
     </header>
 
-    <form id="page-edit-form" action="{% url 'wagtailadmin_pages:add' content_type.app_label content_type.model parent_page.id %}" method="POST" novalidate{% if form.is_multipart %} enctype="multipart/form-data"{% endif %}>
+    <form id="page-edit-form" action="{% url 'wagtailadmin_pages:add' content_type.app_label content_type.model parent_page.id %}" method="POST" novalidate{% if edit_handler.is_multipart %} enctype="multipart/form-data"{% endif %}>
         {% csrf_token %}
         <input type="hidden" name="next" value="{{ next }}">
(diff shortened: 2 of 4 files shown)


gasman:
It seems that this was reported independently as #3435, and fixed in the django-modelcluster library in https://github.com/wagtail/django-modelcluster/pull/73 ... Probably no harm in having the fix on the Wagtail side too, though.

gasman:
Merged in a6bb67f75d588416f34d2884b9cb0ecb0f96e36d - thanks @helmantel!

ababic:
@gasman Attempting to make use of the new `is_multipart()` method in `contrib.modeladmin` too, for consistency. But the `test_create()` test on line 159 of `/wagtail/contrib/modeladmin/tests/test_simple_modeladmin.py` is now erroring with:

`AttributeError: 'PublisherModelAdminForm' object has no attribute 'formsets'`

Any ideas?

gasman:
@ababic Is `PublishedModelAdminForm` a subclass of `modelcluster.forms.ClusterForm`?

ababic:
@gasman It's not no, it's just a straight subclass of `forms.ModelForm`. Is that a requirement for `edit_handler.is_multipart()`? I thought it would've been nice to use that method consistently throughout, no matter the base form class.

gasman:
@ababic Hmm, that's a fair point - edit_handler shouldn't be making the assumption that it's working with a ClusterForm. In fact, I'm now questioning whether edit handlers should really be involved in form logic to the point of knowing whether it's multipart or not. Given that we're _only_ doing so here to work around a modelcluster bug (one that will become moot when Wagtail 1.10 is released, since we'll be releasing a new version of modelcluster at the same time) I think it's sensible to sidestep that decision, and roll this change back.

(Incidentally, is it a routine thing for modeladmin to use non-ClusterForm model forms, or does that only happen if you specifically pass a custom form in? If it's the former, we should look into that - as far as I'm aware, InlinePanels rely on ClusterForm to work correctly.)

ababic:
Hi @gasman. In answer to your question, modeladmin relies on the `edit_handler` to find the right form to use for the model - which I think is the behaviour used elsewhere?

It's currently not documented anywhere, but at the moment, if a developer wants to use `InlinePanel` for non-page models, they have to subclass `ClusterableModel` in order for it to work. I've always wondered whether it would be possible to make `InlinePanel` work whether being used in a `ClusterForm` or not?

ababic:
FYI The above is also true of `wagtailsnippets`

gasman:
@ababic If you leave it to the `edit_handler` to provide the form, it will build it from WagtailAdminModelForm, which subclasses ClusterForm.

To make InlinePanel work without ClusterForm, we'd need some other mechanism to ensure that the associated formsets are also saved when we call `form.save()`. That's not so hard in itself, but in the case of the page editor, we also need it to happen on `form.save(commit=False)` in order for preview to work - and that definitely does require modelcluster trickery. So, any non-modelcluster handling of formsets would still have to live alongside the modelcluster logic.

ababic:
@gasman Oh, I didn't realise that was the case regarding `WagtailAdminModelForm`. I'm pretty sure you still get an error on the create/edit views if you register a non-ClusterableModel  model with `InlinePanel` instances in the panel definitions (in  `wagtailsnippets` or `modeladmin`).

PR #3502, opened 2017-03-24
Prevent users from demoting or deactivating themself

huubbouma:
This PR fixed a test for the existing PR https://github.com/wagtail/wagtail/pull/3401

### DIFF
--- a/wagtail/wagtailusers/tests.py
+++ b/wagtail/wagtailusers/tests.py
@@ -303,6 +303,7 @@ class TestUserDeleteViewForNonSuperuser(TestCase, WagtailTestUtils):
 
 
     def test_user_cannot_escalate_privileges(self):
+        deleters_group = Group.objects.get(name='User deleters')
         post_data = {
             'username': "deleter",
             'email': "deleter@email.com",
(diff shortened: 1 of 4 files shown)


wagtail/wagtailusers/tests.py
```diff
@@ -297,6 +301,31 @@ def test_user_cannot_delete_self(self):
         # Check user was not deleted
         self.assertTrue(get_user_model().objects.filter(pk=self.deleter_user.pk).exists())
 
+
+    def test_user_cannot_escalate_privileges(self):
```
gasman: This test doesn't belong in `TestUserDeleteViewForNonSuperuser`, as it's testing the 'edit' view, not the 'delete' view (and fiddling the `setUp` method to give change_user permission to deleters_group is breaking the intent of these tests, namely "a user with _only_ delete permission should be able to do these things").

wagtail/wagtailusers/tests.py
```diff
@@ -348,6 +380,62 @@ def test_edit(self):
...
+        user = get_user_model().objects.get(pk=self.current_user.pk)
+        self.assertEqual(user.first_name, 'Edited Myself')
+
+        # Check that the user is still superuser
+        self.assertEqual(user.is_superuser, True)
+        # Check that the user is still active
+        self.assertEqual(user.is_active, True)
+
+
+    def test_(arg):
```
gasman: this should be deleted

gasman: ...actually, it looks like we're still missing a test to confirm that a superuser can't remove their own is_active / is_superuser flags, so perhaps that was intended to go here? :-)

gasman:
Merged in f663c764907aac1def329d0b424fa4e3b53b2e2c - thanks @strindhaug and @huubbouma!

PR #3505, opened 2017-03-25
Add appyveyor config

mvantellingen:
This commit does add settings for using mssql on appveyor
but there are still some issues with timeouts. For now only
the sqlite build is enabled.

Note that the build currently (correctly) fails on windows

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,20 @@
+services:
+  - mssql2016
+
+environment:
+  matrix:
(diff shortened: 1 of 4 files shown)


mvantellingen:
The appveyor build info is available via https://ci.appveyor.com/project/mvantellingen/wagtail

Before merging this PR we should create a wagtail appveyor account. That account can then give access to other users. I'm still trying to figure out why mssql hangs

PR #3508, opened 2017-03-28
Add parent_page argument in page forms even in GET requests

timheap:
This allows a new page to inherit values from its parent page by overriding `base_form_class`, or to modify the form based on some conditions of its parent.

In my specific case, I am trying to automatically inherit the language of the parent page for all subpages created underneath it.


gasman:
Looks good to me - thanks @timheap!

timheap:
Thanks!

PR #1179, opened 2015-04-13
[WIP] PostgreSQL search backend

kaedroho:
This PR adds a new contrib app called `wagtail.contrib.wagtailpostgressearch`. This provides a new search backend for wagtailsearch to use that uses PostgreSQLs built in full text search features (see: http://www.postgresql.org/docs/8.3/static/textsearch.html, http://blog.lostpropertyhq.com/postgres-full-text-search-is-good-enough/).

This allows you to use Wagtail without Elasticsearch, but without all the limitations of the current database backend (it should be faster than the database backend too).

Features in the PostgreSQL backend that are not in the database backend:
- Full text search
- Indexing of callable fields
- Searching fields in subclasses

A few possible enhancements:
- Partial matching
- Per field boosting
- Stemming
- ~~Use `unaccent` extension~~ done!
- Speed it up even more with a GIN/GiST index

TODO:
- [ ] Clean up code
- [ ] Improve tests
- [ ] Write docs

PR #3515, opened 2017-03-29
Adds PostgreSQL search backend.

BertrandBordage:
This pull request is the outcome of the work done during last week’s Wagtail Space by a team composed of @jaap3, @153957, @ramonakira, & me.

It is available to install and use with exactly the same code in the pip installable [wagtail/wagtail-pg-search-backend](https://github.com/wagtail/wagtail-pg-search-backend).

We wanted to add a ton of other features like language support, autocomplete and spelling suggestions, but we need to make more work on the global Wagtail search API before.

It supersedes the initial great attempt done by @kaedroho two years ago, #1179.
Compared to this initial pull request, everything is now handled apart from partial matching.

Since it is fully functional and doesn’t change any existing line of code, I suggest we include it in Wagtail 1.10. That way, people can test it as soon as possible and make as many feedback as possible. I will do all the maintenance work of this contrib module during the upcoming months to treat these feedbacks.


gasman:
It's just occurred to me that there's no real reason for this to be a contrib module... is there? If we're confident that this deserves to be part of Wagtail (and I think we are), it should exist as `wagtailsearch.backends.postgresql` (and be documented as part of the search module, rather than being a separate thing under 'contrib modules').

BertrandBordage:
@gasman I totally agree with you, but there’s a reason for that.
Indexing using PostgreSQL is done using a database model, and we can’t have / don’t want this model to be used in other databases.

So there’s roughly three ways to do it as far as I know:
- having the model in a separate application like in contrib
- registering the model in Django only if there’s at least one PostgreSQL database and use a database router to prevent migrations and relations for non-PostgreSQL databases
- use a raw SQL migration inside a RunPython operation (yes, not RunSQL) that would only apply if the database is PostgreSQL, then adding `managed = False` to the model and tell users not to use it outside the search API

The contrib module is of course the simplest and less hacky solution, that’s why I chose it.

gasman:
Aww, that's a shame! I agree, better to keep it as a contrib module then.

gasman:
Just noticed that there are a bunch of Postgres-related exceptions being thrown in the test output, e.g. https://travis-ci.org/wagtail/wagtail/jobs/216297253 - are they something to worry about? (And if not, can we silence them?)

BertrandBordage:
Oh, that’s not failing probably because it’s called through the update_index command.
It may be worth changing the test suite so the update_index command actually raises errors…

Anyway, the `reduce` issue is almost nothing, there’s just a check missing, I’ll do it this evening.

BertrandBordage:
@gasman It’s fixed!

BertrandBordage:
Can we finally merge this for 1.10?

BertrandBordage:
Thank you very much @gasman & @kaedroho for the reviews & merge :D

Issue #3431, opened 2017-03-07
WagtailSearch ElasticSearch backend setting to set results size


StriveForBest:
### Issue Summary

There is no way to specify results count (`'size'`), returned by `elasticsearch2` backend which results into always getting the defaults (10) results.

### Steps to Reproduce

1. Set up WagtailSearch with `elasticsearch2` backend.
2. Run `.search('QUERY').results()` always returns 10 results

There is nothing in documentation about specifying the results count. Should be a django setting?
Here is what I am doing:

```
In [7]: query = 'shoes'

In [8]: TenantPage.objects.count()
Out[8]: 88

In [9]: TenantPage.objects.search(query).count()
Out[9]: 80

In [10]: len(TenantPage.objects.search(query).results())
Out[10]: 10
```

I also tried to add 'size' to `OPTIONS` in Django settings as below:

```
WAGTAILSEARCH_BACKENDS = {
    'default': {
        'BACKEND': 'wagtail.wagtailsearch.backends.elasticsearch2',
        'INDEX': 'wagtail',
        'INDEX_SETTINGS': {},
        'OPTIONS': {'from': 0, 'size': 100},  # has no effect
        'TIMEOUT': 5,
        'URLS': ['http://localhost:9200']
    }
}
```

If you trace the `Elasticsearch2SearchBackend`s result class (`Elasticsearch2SearchResults`) to it's parent, you end up [here](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailsearch/backends/elasticsearch.py#L418):

```
class ElasticsearchSearchResults(BaseSearchResults):
    fields_param_name = 'fields'
    
    ...

    def _do_search(self):
        # Params for elasticsearch query
        params = dict(
            index=self.backend.get_index_for_model(self.query.queryset.model).name,
            body=self._get_es_body(),
            _source=False,
            from_=self.start,
        )

        params[self.fields_param_name] = 'pk'

        # Add size if set
        if self.stop is not None:
            params['size'] = self.stop - self.start
    ...
```
If I hardcode `params['size']=100` it works. I see some logic to "Add size if set" but nothing about how to actually set it. Is it documented? But I assume it's being used for pagination and offsetting the results.

### Technical details

* Python version: 2.7.3
* Django version: 1.9.6
* Wagtail version: 1.9

StriveForBest:
`.results()` is what's breaking it. Should it not be used at all for search? Is `.results()` documented anywhere?

[...]

PR #3518, opened 2017-03-30
Fix annotate_score followed by slice

kaedroho:
This fixes the bug reported by @nimasmi in https://github.com/wagtail/wagtail/issues/3431#issuecomment-288051751

Slicing creates a new SearchResults object but the ``score_field`` (set by ``.annotate_score()`` wasn't being passed along to the new SearchResults causing the score to not be annotated.


kaedroho:
This one aught to be backported if we're still doing that.

nimasmi:
Yes it ought. Yes please.

gasman:
Thanks! Merged in babe8a0c09cad1395213c18d3d307ed5b3039a90.

PR #3520, opened 2017-03-30
Remove double-escaping of PKs in modeladmin URLs (rebase of #2722)

gasman:
Rebase of #2722, with some fixes:

* The test now explicitly checks for the expected URLs. Simply testing "all links return 200 responses" is too indirect, since (for example) that would be satisfied by no links at all;
* Stylistically it's better to remove the *other* `quote` operation. The responsibility for quoting the PK should lie with the method that inserts the PK into the URL - the caller of `edit_button` etc shouldn't have to know or care what happens to the PK it passes (and if `edit_button` was expecting a pre-quoted ID, we'd need to indicate that in the method signature by calling the parameter something like `quoted_pk` rather than `pk`).

### DIFF
--- a/wagtail/contrib/modeladmin/helpers/button.py
+++ b/wagtail/contrib/modeladmin/helpers/button.py
@@ -54,7 +54,7 @@ class ButtonHelper(object):
         classnames = self.inspect_button_classnames + classnames_add
         cn = self.finalise_classname(classnames, classnames_exclude)
         return {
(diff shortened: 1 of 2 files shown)


gasman:
Merged in 421eb9ef8a3f75d7efba230f6c08a5baf6cf466a - thanks @nutztherookie and @ababic!

PR #3521, opened 2017-03-30
Fetch search results with scroll API

### LINKED ISSUE #3431: WagtailSearch ElasticSearch backend setting to set results size

kaedroho:
Fixes #3431

This updates Wagtailsearch to fetch results from Elasticsearch using the [scroll API](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-scroll.html)

This loads in 100 results at a time, allowing many results to be streamed into Wagtail if needed. It also fixes the behaviour when the user hasn't specified a limit on the number of search results they need, which currently loads just 10 results.

### DIFF
--- a/wagtail/wagtailsearch/backends/elasticsearch.py
+++ b/wagtail/wagtailsearch/backends/elasticsearch.py
@@ -447,28 +447,71 @@ class ElasticsearchSearchResults(BaseSearchResults):
         else:
             limit = None
 
-        # Params for elasticsearch query
-        params = dict(
-            index=self.backend.get_index_for_model(self.query.queryset.model).name,
-            body=self._get_es_body(),
-            _source=False,
-            from_=self.start,
-            scroll='2m',
-            size=min(limit or PAGE_SIZE, PAGE_SIZE),
-        )
+        use_scroll = limit is None or limit > PAGE_SIZE
 
-        params[self.fields_param_name] = 'pk'
+        params = {
+            'index': self.backend.get_index_for_model(self.query.queryset.model).name,
+            'body': self._get_es_body(),
+            '_source': False,
+            'from_': self.start,
+            self.fields_param_name: 'pk',
+        }
 
-        # Send to Elasticsearch
-        page = self.backend.es.search(**params)
+        if use_scroll:
+            params.update({
+                'scroll': '2m',
+                'size': PAGE_SIZE,
+            })
 
-        while True:
-            if len(page['hits']['hits']) == 0:
-                return
+            # Send to Elasticsearch
+            page = self.backend.es.search(**params)
+
+            while True:
+                if len(page['hits']['hits']) == 0:
+                    return
+
+                # Get pks from results
+                pks = [hit['fields']['pk'][0] for hit in page['hits']['hits']]
+                scores = {str(hit['fields']['pk'][0]): hit['_score'] for hit in page['hits']['hits']}
+
+                # Initialise results dictionary
+                results = dict((str(pk), None) for pk in pks)
+
+                # Find objects in database and add them to dict
+                queryset = self.query.queryset.filter(pk__in=pks)
+                for obj in queryset:
+                    results[str(obj.pk)] = obj
+
+                    if self._score_field:
+                        setattr(obj, self._score_field, scores.get(str(obj.pk)))
+
+                # Yield results in order given by Elasticsearch
+                for pk in pks:
+                    if results[str(pk)]:
+                        yield results[str(pk)]
+
+                        if limit is not None:
+                            limit -= 1
+
+                            if limit == 0:
+                                return
+
+                # Fetch next page of results
+                if '_scroll_id' not in page:
+                    return
+
+                page = self.backend.es.scroll(scroll_id=page['_scroll_id'], scroll='2m')
+        else:
+            params.update({
+                'size': limit or PAGE_SIZE,
+            })
+
+            # Send to Elasticsearch
+            hits = self.backend.es.search(**params)
 
             # Get pks from results
-            pks = [hit['fields']['pk'][0] for hit in page['hits']['hits']]
-            scores = {str(hit['fields']['pk'][0]): hit['_score'] for hit in page['hits']['hits']}
+            pks = [hit['fields']['pk'][0] for hit in hits['hits']['hits']]
+            scores = {str(hit['fields']['pk'][0]): hit['_score'] for hit in hits['hits']['hits']}
 
             # Initialise results dictionary
             results = dict((str(pk), None) for pk in pks)
@@ -485,18 +528,6 @@ class ElasticsearchSearchResults(BaseSearchResults):
             for pk in pks:
                 if results[str(pk)]:
                     yield results[str(pk)]
-
-                    if limit is not None:
-                        limit -= 1
-
-                        if limit == 0:
-                            return
-
-            # Fetch next page of results
-            if '_scroll_id' not in page:
-                return
-
-            page = self.backend.es.scroll(scroll_id=page['_scroll_id'], scroll='2m')
 
     def _do_count(self):
         # Get count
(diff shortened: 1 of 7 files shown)


kaedroho:
@BertrandBordage This is now ready for review!

wagtail/wagtailsearch/backends/elasticsearch.py
```diff
@@ -486,13 +486,17 @@ def _do_search(self):
                             limit -= 1
 
                             if limit == 0:
-                                return
+                                break
```
kaedroho: Oops, this won't break from the while loop

kaedroho: Fixed

wagtail/wagtailsearch/backends/elasticsearch.py
```diff
@@ -432,40 +432,101 @@ def _get_es_body(self, for_count=False):
...
+            params.update({
+                'scroll': '2m',
+                'size': PAGE_SIZE,
+            })
+
+            # Send to Elasticsearch
+            page = self.backend.es.search(**params)
+
+            while True:
+                if len(page['hits']['hits']) == 0:
```
BertrandBordage: To avoid extra visual noise, define a variable like `hits = page['hits']['hits']`since it’s also used below.
Then I don’t think you have to use a length comparison as hit is a dict. Simply use `if not hit:`, it will be faster as it skips a rich comparison, a local name lookup, a function call, a tuple construct and an unoptimised length fetch.

wagtail/wagtailsearch/backends/elasticsearch.py
```diff
@@ -432,40 +432,101 @@ def _get_es_body(self, for_count=False):
...
+                    break
+
+                # Get pks from results
+                pks = [hit['fields']['pk'][0] for hit in page['hits']['hits']]
+                scores = {str(hit['fields']['pk'][0]): hit['_score'] for hit in page['hits']['hits']}
+
+                # Initialise results dictionary
+                results = dict((str(pk), None) for pk in pks)
```
BertrandBordage: Use a dict comprehension for simplicity and performance.

wagtail/wagtailsearch/backends/elasticsearch.py
```diff
@@ -432,40 +432,101 @@ def _get_es_body(self, for_count=False):
...
+                self.backend.es.clear_scroll(scroll_id=page['_scroll_id'])
+        else:
+            params.update({
+                'size': limit or PAGE_SIZE,
+            })
 
-        # Send to Elasticsearch
-        hits = self.backend.es.search(**params)
+            # Send to Elasticsearch
+            hits = self.backend.es.search(**params)
```
BertrandBordage: Since variable `hits` is only used for `hits['hits']['hits']`, maybe use `hits = self.backend.es.search(**params)['hits']['hits']` instead.

wagtail/wagtailsearch/backends/elasticsearch.py
```diff
@@ -432,40 +432,101 @@ def _get_es_body(self, for_count=False):
...
+
+                # Get pks from results
+                pks = [hit['fields']['pk'][0] for hit in page['hits']['hits']]
+                scores = {str(hit['fields']['pk'][0]): hit['_score'] for hit in page['hits']['hits']}
+
+                # Initialise results dictionary
+                results = dict((str(pk), None) for pk in pks)
+
+                # Find objects in database and add them to dict
+                queryset = self.query.queryset.filter(pk__in=pks)
```
BertrandBordage: This variable attribution can be skipped and put directly in the for loop as it is simple enough.

wagtail/wagtailsearch/backends/elasticsearch.py
```diff
@@ -432,40 +432,101 @@ def _get_es_body(self, for_count=False):
...
-        results = dict((str(pk), None) for pk in pks)
+            # Initialise results dictionary
+            results = dict((str(pk), None) for pk in pks)
 
-        # Find objects in database and add them to dict
-        queryset = self.query.queryset.filter(pk__in=pks)
-        for obj in queryset:
-            results[str(obj.pk)] = obj
+            # Find objects in database and add them to dict
+            queryset = self.query.queryset.filter(pk__in=pks)
```
BertrandBordage: This variable attribution can be skipped and put directly in the for loop as it is simple enough.

wagtail/wagtailsearch/backends/elasticsearch.py
```diff
@@ -432,40 +432,101 @@ def _get_es_body(self, for_count=False):
...
-                setattr(obj, self._score_field, scores.get(str(obj.pk)))
+                if self._score_field:
+                    setattr(obj, self._score_field, scores.get(str(obj.pk)))
 
-        # Return results in order given by Elasticsearch
-        return [results[str(pk)] for pk in pks if results[str(pk)]]
+            # Yield results in order given by Elasticsearch
+            for pk in pks:
+                if results[str(pk)]:
+                    yield results[str(pk)]
```
BertrandBordage: Define a variable `result = results[str(pk)]`. This avoids an extra string conversion & and dictionary lookup in almost all cases.

wagtail/wagtailsearch/backends/elasticsearch.py
```diff
@@ -432,40 +432,101 @@ def _get_es_body(self, for_count=False):
...
+                for obj in queryset:
+                    results[str(obj.pk)] = obj
+
+                    if self._score_field:
+                        setattr(obj, self._score_field, scores.get(str(obj.pk)))
+
+                # Yield results in order given by Elasticsearch
+                for pk in pks:
+                    if results[str(pk)]:
+                        yield results[str(pk)]
```
BertrandBordage: Define a variable result = results[str(pk)]. This avoids an extra string conversion & and dictionary lookup in almost all cases.

kaedroho:
Changes made. Thanks @BertrandBordage!

PR #3523, opened 2017-03-31
Fixed #3513 -- Fixed API pagination for empty searches

### LINKED ISSUE #3513: Empty search in the API leading to TypeError
Steps to reproduce:
- use Wagtail 1.9 & elasticsearch 2
- install Wagtail’s API
- go to /api/v2/pages/?search=

You end up with:
```
File "…/wagtail/api/v2/pagination.py" in paginate_queryset
  36.         self.total_count = queryset.count()

TypeError: count() takes exactly one argument (0 given)
```

ellmetha:


gasman:
Thanks @ellmetha! Could you add a test for this, please?

ellmetha:
@gasman sure thing. I've just added them. :wink:

wagtail/api/v2/pagination.py
```diff
@@ -33,7 +33,7 @@ def paginate_queryset(self, queryset, request, view=None):
         stop = offset + limit
 
         self.view = view
-        self.total_count = queryset.count()
+        self.total_count = len(queryset) if isinstance(queryset, list) else queryset.count()
```
BertrandBordage: Could you make it more general by reversing this condition?
```python
self.total_count = queryset.count() if isinstance(queryset, QuerySet) else len(queryset)
```

Just in case a search backend returns a tuple or some kind of cursor, for example.

wagtail/contrib/wagtailapi/pagination.py
```diff
@@ -33,7 +33,7 @@ def paginate_queryset(self, queryset, request, view=None):
         stop = offset + limit
 
         self.view = view
-        self.total_count = queryset.count()
+        self.total_count = len(queryset) if isinstance(queryset, list) else queryset.count()
```
BertrandBordage: Same comment as above.

wagtail/api/v2/pagination.py
ellmetha: Of course ; I've just updated my previous commit. :wink: It turns out that `queryset` can be an instance of `QuerySet` but it can also be an instance of `BaseSearchResults` (which is the case if you use the Elasticsearch backend and the `search` parameter for example).

wagtail/api/v2/pagination.py
```diff
@@ -33,7 +36,8 @@ def paginate_queryset(self, queryset, request, view=None):
         stop = offset + limit
 
         self.view = view
-        self.total_count = queryset.count()
+        self.total_count = (queryset.count() if isinstance(queryset, QuerySet) or
+                            isinstance(queryset, BaseSearchResults) else len(queryset))
```
BertrandBordage: `isinstance` can take a tuple of classes as its second argument, so you can write instead:
```python
isinstance(queryset, (QuerySet, BaseSearchResults))
```

kaedroho:
Thanks for fixing this!

I think there's a bug in wagtail search that's caused this. When there is no query it returns an empty list https://github.com/wagtail/wagtail/blob/babe8a0c09cad1395213c18d3d307ed5b3039a90/wagtail/wagtailsearch/backends/base.py#L237

This should always be returning a subclass of ``SearchResults`` so the API doesn't magically change dependant on the users search query.

ellmetha:
@kaedroho I've just added a new commit in order to ensure that the `BaseSearchBackend.search` method return only `SearchResults` subclasses.

PR #3525, opened 2017-04-04
Fixed #1602 - Caught a ProtectedError for images

### LINKED ISSUE #1602: Support for PROTECT on foreign keys

Toshakins:
Fixed #1602

Feel like my way is a bit of humongous, but open to suggestions.

Sorry if made your eyes bleed :eye: :sweat_drops:


gasman:
Thanks @Toshakins!

A few issues with this as it stands (which suggest that this might be a much harder task than it appears at first, unfortunately):

* The 'non-core' apps of Wagtail (wagtailimages, wagtaildocs, wagtailsnippets etc) should be fully independent of each other - we can't import `wagtailsnippets.models` within wagtailimages, for example. If we want to distinguish between different model types in the response we send back to users, we'll have to do something with hooks - for example, letting apps register themselves as ProtectedError handlers for specific models (i.e. "if you encounter a ProtectedError relating to model X, here's the message and URL route to report back to the user").
* Our error messages need to be understandable by non-technical users - content editors on a Wagtail site are not necessarily developers. Anything that mentions primary keys or "entities" probably isn't suitable... :-)
* We need to consider the possibility that the user doesn't have edit permission over the related object that's blocking the deletion.

The common theme to all of these, I guess, is that there might not be any helpful advice we can give to the user - it might really be a case of "you're not allowed to delete this, and we can't tell you why". Not really sure how to deal with that in a user-friendly way!

gasman:
Closing as per https://github.com/wagtail/wagtail/pull/3525#issuecomment-308510208 - embedding knowledge within wagtailimages of all the possible related object types isn't really workable.

Issue #3530, opened 2017-04-10
Userbar JS exit condition looks broken


HannesWaechter:
Found a bug? Please fill out the sections below. 👍

### Issue Summary

wagtailadmin userbar.js line 11:

    if (!'classList' in userbar)

evaluates to

    if(false in userbar)

this might not be the wanted behavior?

gasman:
Check removed in 6630d4b66c4c5291c76c50e8c7c8c2f220e30268, as per #3531.

gasman:
Thanks @HannesWaechter!

PR #3531, opened 2017-04-10
add parenthesis to fix userbar exit condition

thijskramer:
This solves issue #3530.


gasman:
Given that this line is (as far as I can tell...) a feature test for the `classList` browser API, which has been [supported by all browsers](http://caniuse.com/#feat=classlist) long before [our 'supported versions' cutoff](https://github.com/wagtail/wagtail/blob/master/docs/contributing/developing.rst#testing), I think we can reasonably delete this line completely.

gasman:
Removed in 6630d4b66c4c5291c76c50e8c7c8c2f220e30268 . Thanks for the report, @thijskramer!

PR #3537, opened 2017-04-12
[WIP] Use modeladmin and class based views for settings

timheap:
Still a work in progress. Needs docs and updated tests.

This PR replaces the old settings function-based view implementation with class-based views, and integrates with ModelAdmin. I started this work because I wanted to move the menu for a setting from the settings menu into a ModelAdminGroup. The best way to do that was to reimplement settings in the modeladmin style.

This did expose some awkward points in modeladmin when extending it in this manner, which I will document and consider how to fix, but that might be a task for another PR.

I am unsure if this counts as backwards incompatible, as it only changes Wagtail internals. The most likely problem comes from moving how the URLs are made.


ababic:
Hi @timheap. I'm very happy with the changes you've made to `wagtail/contrib/modeladmin/options.py` here, but I've got to say, I do feel a little bit uncomfortable about the idea of a contrib app using elements from another contrib app. If we're going to take a step toward consistent, class-based views for apps, It'll probably be better to do that via something lighter and more centralised, such as @kaedroho's ViewSets (see PR #2113).

gasman:
I agree with @ababic - it's been on our wishlist for a while to move more reusable view logic into wagtailadmin, and have our central and contrib apps (including modeladmin) piggyback on that, rather than have other contrib apps depend on modeladmin. I've now added #2113 to the roadmap for 1.11, so hopefully we'll make some progress towards this in the next release.

lb-:
Hey @timheap  - I know it's been a while since the last comment on this, as a heads up it looks like some of the related issues in the comments from Matt & Andy have now been resolved.

We are doing a bit of clean up and marking all WIP PRs as a draft, this will still be visible in the issues list and CI checks will still run. 

Feel free to mark as a non-draft & remove the WIP tag from the title when this is close to ready for a review.

timheap:
Hey @lb-, I no longer have time to contribute to Wagtail, so I am unlikely to pick this PR up again. It is your call what you want to do with PR, in terms of closing it or opening a new one to continue the work, but I won't be taking the work on.

gasman:
Cheers Tim - will close this now, as it seems likely that the eventual fix (extending the generic views in wagtail.admin.views.generic to serve as the common base for both settings and modeladmin) won't be directly based on this.

Issue #3534, opened 2017-04-11
delete signal listeners should use on_commit


thenewguy:
The delete signal handlers like https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailimages/signal_handlers.py#L9 should use `transaction.on_commit()` so that they wait to delete until the sql delete is successful

Something like:
```
from django.db import transaction
def post_delete_file_cleanup(instance, **kwargs):
    transaction.on_commit(lambda: instance.file.delete(False))
```

BertrandBordage:
That’s right, we may lose files otherwise.

Documents are also concerned: https://github.com/wagtail/wagtail/blob/master/wagtail/wagtaildocs/signal_handlers.py

`transaction.on_commit` was added in Django 1.9 if I remember well.
Currently, Wagtail still supports Django 1.8, so this has to wait a bit, or we can implement it with a check in the meantime.

Would you like to provide a pull request for this?

thenewguy:
@BertrandBordage Sure. I will try to get to it tomorrow. I've already written tests that I should be able to copy and paste

BertrandBordage:
Awesome!

thenewguy:
@BertrandBordage have a problem with me making the signal handlers generic and moving them to wagtailcore so they don't have to be duplicated between wagtailimages and wagtaildocs?

thenewguy:
[This test](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtaildocs/tests.py#L853) is failing because the test uses a file name that clashes with other tests that do not clean up their test documents.

The other failing tests occur on databases other than sqllite are slightly more difficult to fix because the vagrant-wagtail-development doesn't seem to offer testing on the different databases and travis tests are slow.  Seems to be related to `serialized_rollback = True` which wasn't required when I originally wrote the tests against postgres.  But copying them over to the vagrant-wagtail-develop box seems to force sqlite

Issue #3539, opened 2017-04-12
Document model signal handlers are not applied to custom document models


thenewguy:
I noticed this while working on https://github.com/wagtail/wagtail/issues/3534

Should be a simple copy/paste fix using the wagtailimages approach.  Ideally these signals could be generically applied in wagtailcore to be DRY but that is out of my pervue

gasman:
Fixed in #3853

PR #3541, opened 2017-04-12
Fix 3534 3539

thenewguy:
fixes https://github.com/wagtail/wagtail/issues/3534 and https://github.com/wagtail/wagtail/issues/3539


thenewguy:
@BertrandBordage @gasman Is there already a fix/fixture for this postgres missing record available? I am not sure what the `postgres_search_indexentry` table is supposed to contain or why the failing tests are even accessing it. Is there a `override_settings` I can use to disable it... I think it is unrelated to what I am testing

```
django.db.utils.IntegrityError: insert or update on table "postgres_search_indexentry" violates foreign key constraint "postgres_sea_content_type_id_d805086f_fk_django_content_type_id"

DETAIL:  Key (content_type_id)=(2) is not present in table "django_content_type".
```

gasman:
Rebased at https://github.com/gasman/wagtail/tree/thenewguy-fix-3534-3539 - as I recall I had to deal with the document file cleanup issues when I split wagtaildocs/tests.py into multiple files, so we'll see if Travis is happier with the result now...

gasman:
Have hopefully fixed the `postgres_search_indexentry` issue in https://github.com/gasman/wagtail/commit/e0bb0acdf1543b876645d160c681d8579080d91c.

The signal handlers registered by the Postgres search backend (which are triggered by a save/deletion on any indexed model, including images/documents) were making use of a cached copy of the contenttypes table. Caching this data relies on the assumption that the content types won't be changed or deleted during the lifetime of the python process; this assumption breaks inside a `TransactionTestCase`, because the contenttypes table is emptied.

I haven't been able to work out why it only fails on images and not documents though...

I've therefore added a flag to disable the cache when running tests. This feels hacky (because it means our test environment doesn't match the real world), but it seemed to be the least worst solution (because anything else would require wagtailimages to have built-in knowledge of wagtail.contrib.postgres_search, or vice versa). @BertrandBordage, would you be happy with this change?

BertrandBordage:
If this caching poses a problem, then we could remove the caching and use [`ContentType.get_for_models`](https://github.com/django/django/blob/master/django/contrib/contenttypes/models.py) instead. It uses a cache to prevent extra database calls, and from the result we can fetch the primary keys.
It should be slightly slower, but it’s better than adding this kind of kludgy setting.

m1kola:
I've added the "Needs docs" label, because documentation for documents (see #3789) needs to be updated.

gasman:
Thanks for the suggestion @BertrandBordage! I've tried it here: https://github.com/gasman/wagtail/commit/908a3ea2c62b9fcff91de5ef450e41acb487f7a8 / https://github.com/gasman/wagtail/commit/a001933f0b2f4394d90137d109798aa1c20d2242 - just running CI at https://travis-ci.org/gasman/wagtail/builds/268318854 now.

I had to remove the `using(db_alias)` clause, as that returned a plain `QuerySet`. I don't _think_ that's needed (surely the ContentType model itself knows the correct database to use?) but I'm not massively familiar with multi-database setups, so I may be missing something.

BertrandBordage:
Thanks @gasman! Yeah, multi-database setup is tricky, especially in this search backend context. I didn’t test multi-database for the PostgreSQL search backend, but I wanted to make things as close as possible to multi-database support to avoid making it hard to do later. So if we could keep it that way, that would be great. I’ll make comments on your commits for that.

PR #3542, opened 2017-04-13
Improve consistency of modeladmin view initialisation and reduce general inconsistency / poor practice between views

### LINKED ISSUE #3392: Improve consistency of how views in ModelAdmin are initialised

ababic:
Fixes #3392.

This helps to remove some of the weirdness in how the ModelAdmin class initialises class based views, and also aims to make a lot of the view functionality much more standardised... e.g.

- Makes better use of some of the low-level generic views / mixins from django instead of reinventing the wheel (`SingleObjectMixin` for `InstanceSpecificView` and `FormView` for `ChooseParentView`)
- Overriding `get()` and `post()` methods where possible instead of `dispatch()`
- More sensible use of the `@cached_property` decorator

Currently, these changes will cause issues for anyone who has added their own custom views by following the current examples. Anyone have any ideas on how we solve that problem? Or even if it's that big a problem?

Also, any other feedback welcome! :)

### DIFF
--- a/wagtail/contrib/modeladmin/options.py
+++ b/wagtail/contrib/modeladmin/options.py
@@ -312,8 +312,8 @@ class ModelAdmin(WagtailRegisterable):
         the assigned model. The view class used can be overridden by changing
         the 'index_view_class' attribute.
         """
-        view_class = self.index_view_class
-        return view_class.as_view(model_admin=self)(request, **kwargs)
+        view = self.index_view_class.as_view(model_admin=self)
+        return view(request, **kwargs)
 
     def create_view(self, request, **kwargs):
         """
@@ -322,8 +322,8 @@ class ModelAdmin(WagtailRegisterable):
         assigned model extends 'Page'. The view class used can be overridden by
         changing the 'create_view_class' attribute.
         """
-        view_class = self.create_view_class
-        return view_class.as_view(model_admin=self)(request, **kwargs)
+        view = self.create_view_class.as_view(model_admin=self)
+        return view(request, **kwargs)
 
     def choose_parent_view(self, request, **kwargs):
         """
@@ -333,8 +333,8 @@ class ModelAdmin(WagtailRegisterable):
         The view class used can be overridden by changing the
         'choose_parent_view_class' attribute.
         """
-        view_class = self.choose_parent_view_class
-        return view_class.as_view(model_admin=self)(request, **kwargs)
+        view = self.choose_parent_view_class.as_view(model_admin=self)
+        return view(request, **kwargs)
 
     def inspect_view(self, request, **kwargs):
         """
@@ -342,8 +342,8 @@ class ModelAdmin(WagtailRegisterable):
         the assigned model. The view class used can be overridden by changing
         the 'inspect_view_class' attribute.
         """
-        view_class = self.inspect_view_class
-        return view_class.as_view(model_admin=self)(request, **kwargs)
+        view = self.inspect_view_class.as_view(model_admin=self)
+        return view(request, **kwargs)
 
     def edit_view(self, request, **kwargs):
         """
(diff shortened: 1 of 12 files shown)


ababic:
These changes are now ready for review

JamesRamm:
hey @ababic 
From my POV these are on the whole welcome changes. As far as backwards compatibility is concerned, I am only seeing a few small changes will be necessary for users (unless ive misunderstood) - that wouldnt be a problem for me, provided it is well documented. Of course, I have the luxury of only having to update a couple of small projects...

The only parts that confuse me a little are to do with dropping overriding `dispatch` in favour of overriding `get` and `post` methods. In `WMABaseView`, lines 70-~80, I dont see the advantage of overriding each method, instead of simply doing the permissions check once in `dispatch` ? 

In `IndexView` you end up calling the permissions check twice, once in `dispatch` and again in the `get` method - is that necessary?

I am definitely in favour of using existing django functionality where possible - that is a welcome change :+1:

ababic:
Hey @JamesRamm. Thank you for the input. I've managed to cater for the main backwards compatibility issue by still supporting the passing of `instance_id` as a keyword argument to `as_view()` (as is done currently for instance-specific views), but raising a deprecation warning. We don't actually provide any documentation to explain how to add custom views to ModelAdmin, but I most cases, I assume developers will have copied the existing  `ModelAdmin.edit_view()` or `ModelAdmin.delete_view()` implementations to add their own instance-specific views, and so didn't want to break things for them.

> The only parts that confuse me a little are to do with dropping overriding dispatch in favour of overriding get and post methods. In WMABaseView, lines 70-~80, I don't see the advantage of overriding each method, instead of simply doing the permissions check once in dispatch ?

I don't feel particularly strongly about this, but my understanding from looking again at Django's generic views, is that `dispatch()` exists solely to call the correct method to handle a request, and isn't really intended to be overridden (in all of the generic views available, none seem to override it, [even when it looks like there'd be an obvious DRYness gain](https://github.com/django/django/blob/master/django/views/generic/edit.py#L187). It's a total guess, but I think the code may be slightly easier to follow if view logic belongs in the view methods themselves, even if it goes against the DRY principle a little. 

I know it's now a little out of line with the above thinking, but I've deliberately left `IndexView` as is for now, as I wanted to avoid scope creep. Although not ideal, having the permission check happen twice shouldn't have any noticeable performance impact.

ababic:
@JamesRamm Despite my earlier defence, I have moved the `deny_request_if_not_permitted()` to dispatch. I hadn't thought about it before, but checking the user's permission to see if they should be able to access the view is arguably part of dispatch logic, rather than view logic, so having it live in `dispatch()` isn't does make sense - And it does help us remove a lot of repetition.

I've also done some tidying to IndexView, so that it's not quite so out-of-sorts with the rest of the views.

JamesRamm:
@ababic 
Cool. It looks good. I was just about to say that I am easy either way with the use of dispatch since those details dont affect the user, but have changed it already :D

tomdyson:
Hi @ababic - let us know when this is no longer WIP and we'll assign someone to review it. Thanks!

ababic:
Hi @tomdyson. This is good to review now! thanks :)

tomdyson:
Thanks @ababic  - I've assigned @gasman to review.

ababic:
Hi @gasman. Would I be right in thinking that you're wanting to put off making view-related changes anywhere until the core generic views are in a place where you're ready to completely rebase onto those?

gasman:
@ababic Not particularly, no - while refactoring things into generic views is something I'd like to do, it's not an immediate priority, so I wouldn't want other development to be held up by it.

Sorry for the delay in reviewing this PR - the problem is that there are a lot of different strands of changes here, and commits being reverted and patched with follow-up commits, so it's going to take a while to unpick it all. If you're able to rebuild it as one or more new PRs, each doing a single well-defined thing, then that would definitely speed things along. Otherwise, I'll hopefully get round to it before too long!

ababic:
Hi @gasman. Thanks for the reply. The story of the individual commits here isn't so important here. The latest diff should give you mostly everything you need. I don't see an easy way of picking this out into separate PRs, as the view changes are quite dependant on each other. The reason I back-tracked on a lot of a few things was to make this as easy to review as possible.

gasman:
Unfortunately, taking the code change as a whole, without the context of the commit messages, makes this a really difficult thing to review. For example, I can see from the diff that the `IndexView. get_context_data` method has been moved up a few hundred lines - if I saw that change in isolation with the commit message "reorder IndexView methods", I'd wave it through without a second thought, but here I have to double-check everything (is it _really_ just being reorganised, or being moved to a different class...?) and then do a manual line-by-line comparison to see if there are any other meaningful changes to the method. There's just too much to keep track of at once.

ababic:
Okay. Point well made. I will close this PR and have a crack at resubmitting again at a later date.

PR #3543, opened 2017-04-13
Run lints in parallel on Drone

kaedroho:
Just updated Drone to a version that supports parallel builds.

This should make builds a little bit faster.


kaedroho:
Leaving this for now as that new version of Drone appears to be broken.

PR #3548, opened 2017-04-15
oEmbed: Add support of Tumblr

m1kola:
Tumblr allows to embed posts using oEmbed standard. See [Tumblr API](https://www.tumblr.com/docs/en/api/v2).

### DIFF
--- a/wagtail/wagtailembeds/oembed_providers.py
+++ b/wagtail/wagtailembeds/oembed_providers.py
@@ -302,7 +302,7 @@ OEMBED_ENDPOINTS = {
         "^http(?:s)?://(?:www\\.)?issuu\\.com/[^#?/]+/docs/.+$"
     ],
     "https://www.tumblr.com/oembed/1.0": [
-        "^http(?:s)?://.+.tumblr.com/post/.+$",
+        "^http(?:s)?://.+\\.tumblr\\.com/post/.+$",
     ]
 }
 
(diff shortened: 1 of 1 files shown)


wagtail/wagtailembeds/oembed_providers.py
```diff
@@ -301,6 +301,9 @@
     "http://issuu.com/oembed": [
         "^http(?:s)?://(?:www\\.)?issuu\\.com/[^#?/]+/docs/.+$"
     ],
+    "https://www.tumblr.com/oembed/1.0": [
+        "^http(?:s)?://.+.tumblr.com/post/.+$",
```
BertrandBordage: A few improvements to the regexp:
- You must escape the explicit dots in the domain name.
- You can remove the non-capturing group for 's' of 'https'. It could of course be removed on all other oembed regexps.
- Maybe a non-greedy expression for the subdomain would be better.

So, something like:
```python
r'^https?://.+?\.tumblr\.com/post/.+$',
```
or
```python
r'^https?://[^\.]+\.tumblr\.com/post/.+$',
```

m1kola: > You must escape the explicit dots in the domain name.

Ah, right. Thanks! I've forgotten to do that. But dots should be escaped like `\\.`, because it's just a string. We compile it later in the `compile_endpoints` function. 

> You can remove the non-capturing group for 's' of 'https'. It could of course be removed on all other oembed regexps.

I'll leave it as is to for now to match other expressions.

> Maybe a non-greedy expression for the subdomain would be better.

We can't use `//[^\.]+` for subdomains because `http://www.something.tumblr.com/post/something_else` will not work.

I'll replace with `.+?`

m1kola:
Updated. @BertrandBordage thanks for the code review

gasman:
Merged in 5f651299df80db441e14be1c122e8e963fd00cd1

PR #3552, opened 2017-04-19
Pull new translations from Transifex

gasman:
Submitting as a PR because I've tweaked the sort order logic for the Language Preferences dropdown and want to make sure it's valid across all Python / Django versions...


gasman:
Merged in 812f1b97ce770c319400468e531fba0535caf0ab

PR #3554, opened 2017-04-19
Make exceptions that guide about missing content types

benjaoming:
Thanks for MAKING Wagtail! 🎉

This is a minor addition, maybe slightly unconventional?

My case is this:

1) I have messed around with some content types, i.e. models inheriting from `Page` and created contents with models (which result in new Content Types) which I ended up removing again.

2) I recognize this pitfall from django-cms - somehow you can run updates, migrations etc. and suddenly the `Page` model contains a reference to a Content Type which was then deleted (and remember that there are no constraints that stops Django from removing a content type).

These exception messages are just there to help a developer understand quickly what's going on. I decided to recommend deletion because *typically*, updating to a new content type would just get you in a place where you're pointing to an undesired content type with a different schema and then the mess explodes.

### DIFF
--- a/wagtail/wagtailcore/query.py
+++ b/wagtail/wagtailcore/query.py
@@ -382,7 +382,39 @@ def specific_iterator(qs):
 
     # Yield all of the pages, in the order they occurred in the original query.
     for pk, content_type in pks_and_types:
-        yield pages_by_type[content_type][pk]
+        try:
+            yield pages_by_type[content_type][pk]
+        except KeyError:
+            if content_type not in pages_by_type:
+                exc_msg = (
+                    "No content type with pk={}. Most likely, you have "
+                    "changes in your content types which are not "
+                    "reflected in existing pages. If you are aware of "
+                    "this, you can run the following command to DELETE "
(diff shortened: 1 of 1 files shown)


gasman:
Thanks @benjaoming!
The behaviour of `PageQuerySet.specific()` was fixed in #3569 to fall back on the basic `Page` model when the more specific model is missing - which matches the behaviour of `Page.specific`, and ensures that the admin page listing still works (so that the page can be deleted from there). As a result, this is no longer an issue on Wagtail 1.12 and later.

benjaoming:
Before:

image

After:

image

benjaoming:
To reproduce the above:

1. Create my_app.MyModel
1. `makemigrations <my_app>`
1. Go to `http://127.0.0.1:8000/wagtail/` and create a new page with MyModel
1. Migrate backwards so MyModel is unappplied. Delete the migration and run `makemigrations <my_app>`, when prompted, delete the content type for `MyModel`
1. `makemigrations <my_app>`
1. Go to `http://127.0.0.1:8000/wagtail/` and observe that the main page has a `KeyError`

PR #3555, opened 2017-04-20
update django requirement in setup.py to allow django 1.11

massaad:
note that even though the next planned django version is 2.0, I still put the requirement as < 1.12 since that seems more conservative and should work for 2.0 as well as for the very unlikely chance we get a 1.12.


gasman:
Thanks @massaad!

Issue #3250, opened 2016-12-23
Custom Page Manager from queryset not working


pyMan:
I can't make things work just using the sample code in the guide [here](http://docs.wagtail.io/en/v1.7/topics/pages.html?highlight=PageQuerySet#custom-page-managers)

Please note there is an error in the sample code: there is a missing "()" at the end of this line`objects = PageManager.from_queryset(EventPageQuerySet)`

I initially had the problem while working with wagtail_modeltranslation (see more at infoportugal/wagtail-modeltranslation#95). I couldn't find the problem, so I thought it may be not related to that thirdy-app. I finally restarted from zero with the sample, but I see I get this traceback:
```
Traceback (most recent call last):
    EventPage.objects.future()
  File "C:\xxxxxxxxxxx\Lib\site-packages\django\db\models\manager.py", line 127, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
AttributeError: 'PageQuerySet' object has no attribute 'future'
```

* Python version: 2.7
* Django version: 1.8.x
* Wagtail version: 1.7

gasman:
Hi @pyMan,

This was fixed in #3090 - please try the version in the 1.8 docs: http://docs.wagtail.io/en/v1.8/topics/pages.html?highlight=PageQuerySet#custom-page-managers

pyMan:
Thank you, I'll try in the next days.

pyMan:
Hi @gasman,

I'm still getting the same issue, but I get things working defining both a CustomPageManager and a CustomPageQuerySet.

Working example:

```
from django.db import models
from django.utils import timezone
from wagtail.wagtailcore.models import Page, PageManager, PageQuerySet

class EventPageQuerySet(PageQuerySet):
    
    def future(self):
        today = timezone.localtime(timezone.now()).date()
        return self.filter(start_date__gte=today)

class EventPageManager(PageManager):
    def get_queryset(self):
        return EventPageQuerySet(self.model).order_by('path')

EventPageManager = EventPageManager.from_queryset(EventPageQuerySet)

class EventPage(Page):
    start_date = models.DateField()

    objects = EventPageManager()
```

Staying at the doc it seems not necessary to define a CustomPageManager at all. I had to define one to customize the "get_queryset" method because the wagtail's one returns a PageQuerySet object.

thornycrackers:
I also ran into this problem and solved it by overriding the get_queryset function on a custom Manager

gasman:
[...]

PR #3557, opened 2017-04-20
Make custom PageManagers return the correct PageQuerySet subclass

gasman:
As reported in https://github.com/wagtail/wagtail/issues/3250#issuecomment-284988695, custom page managers created with PageManager.from_queryset(CustomPageQuerySet) fail to return instances of CustomPageQuerySet. This breaks the EventPageQuerySet example given at http://docs.wagtail.io/en/v1.9/topics/pages.html#custom-page-managers.


gasman:
Merged in be73f16e14f61b6d424c2ae7f25029ef3ba2e933

PR #3559, opened 2017-04-20
Update docs - run python manage.py migrate once

bauersebastian:
We have to run python manage.py migrate once, in order to create the necessary tables for the postgres search backend.
Otherwise we get an error, as the insert in the table is not possible.


gasman:
Thanks @bauersebastian! Merged in 49d706b (master) and 793b42e (1.10.x).

Issue #1951, opened 2015-11-20
Can't delete a page of a type that was removed with migrations


alexgleason:
1. Create `TestPage(Page)` in models.py
2. makemigrations, migrate
3. Log into admin and create a TestPage page
4. Delete `TestPage(Page)` in models.py
5. makemigrations, migrate
6. Log into admin and delete test page from step 3

Result: Database error. TestPage isn't found in the database.

Expected result: The page is deleted gracefully.

kaedroho:
I've seen this a couple of times too. I think there's a good chance that #2024 could solve it.

gasman:
#2024 would only fix this if the ContentType record for the deleted page model gets deleted immediately, and I'm not sure we can rely on that (IIRC the "remove these stale content types?" prompt happens on `./manage.py migrate` some unspecified time later, and not at all if you run it with `--noinput`).

We have a check in Page.specific to catch this kind of scenario https://github.com/torchbox/wagtail/blob/aaed92f301a6862f30e817e23a50a21e8ae12e7b/wagtail/wagtailcore/models.py#L538-L544 - it sounds like this particular view is evading that check, and we probably need to repeat that check in some other places.

atbaker:
We ran into a similar problem today which may be a symptom of the same issue:

We had a stale content type from breaking a particularly bloated snippet out into its own app. You're correct @gasman, that `migrate --noinput` will never remove stale content types for you.

For some reason, the groups edit page would 500 error while a stale content type for that snippet still existed. Here's the stacktrace:

```
Traceback (most recent call last):
  File "/Users/abaker/.virtualenvs/my-project/lib/python3.4/site-packages/django/core/handlers/base.py", line 149, in get_response
    response = self.process_exception_by_middleware(e, request)
  File "/Users/abaker/.virtualenvs/my-project/lib/python3.4/site-packages/django/core/handlers/base.py", line 147, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/Users/abaker/.virtualenvs/my-project/lib/python3.4/site-packages/django/views/decorators/cache.py", line 43, in _cache_controlled
    response = viewfunc(request, *args, **kw)
  File "/Users/abaker/.virtualenvs/my-project/lib/python3.4/site-packages/wagtail/wagtailadmin/decorators.py", line 22, in decorated_view
    return view_func(request, *args, **kwargs)
[...]

PR #3569, opened 2017-04-26
Prevent explorer view crashing when page models are missing

### LINKED ISSUE #3567: Deleted page models are not handled gracefully by explorer view
### Issue Summary

See http://stackoverflow.com/a/43589376/1853523 . If an explorer page listing includes a page whose class definition does not exist in the project, template rendering fails with `AttributeError: 'NoneType' object has no attribute 'objects'`. Previously, missing page models were handled gracefully by the `Page.specific` method, which falls back on the basic Page instance where the specific model can't be found - however, it seems that the queryset `specific` method doesn't do this.

### Steps to Reproduce

1. Start a new project with `wagtail start`, and configure it for Postgres
2. `./manage.py migrate` / `./manage.py createsuperuser`
3. Remove the HomePage class definition from `home/models.py`
4. Run `./manage.py makemigrations` / `./manage.py migrate`
5. Log in to the admin and go to http://localhost:8000/admin/pages/ - this throws the error `AttributeError: 'NoneType' object has no attribute 'objects'`

### Technical details

* Python version: 2.7.6.
* Django version: 1.10.7
* Wagtail version: 1.9.1
* Browser version: Chrome 57

gasman:
Fixes #3567 

Ensures that if a page model definition is missing (e.g. because it was removed from the project without first deleting all existing pages of that type) the explorer is still usable. We don't guarantee that the page will actually behave sensibly beyond that, but now at least the admin still holds up for long enough for you to delete the offending page.

### DIFF
--- a/wagtail/wagtailadmin/tests/test_pages_views.py
+++ b/wagtail/wagtailadmin/tests/test_pages_views.py
@@ -9,6 +9,7 @@ import mock
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Group, Permission
+from django.contrib.contenttypes.models import ContentType
 from django.contrib.messages import constants as message_constants
 from django.core import mail, paginator
 from django.core.files.base import ContentFile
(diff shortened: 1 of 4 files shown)


gasman:
This also fixes #1951.

gasman:
Merged in 8003185ac2b6b9d5b75f4668034656def23c452e

benjaoming:
> Merged in 8003185

I don't think that commit contains anything but the release notes? :)

benjaoming:
Ah sorry it did get merged -- my problem is that I'm still getting the same old KeyError, but this time because when the model disappears, so will the Content Type.. so am still missing the other part of the exception info here from this PR: #3554

PR #3570, opened 2017-04-27
Fix preview to handle multi-value keys in request.POST

### LINKED ISSUE #3564: Preview in 1.10rc1 fails if ParentalManytoManyField changes or is set in page
### Issue Summary

I make use of the ParentalManytoManyField in a model (for blog categories).
When we tick one ore more categories and save the draft the preview is no longer working. This only happens to newly created pages since the upgrade to 1.10rc1.
On the database everything is fine, since the relationships are saved properly, once the page is published.

In class PreviewOnEdit in wagtailadmin/views/pages.py the method get() --> if form.is_valid() returns false, because the Django Form Validation returns an error 'Please insert list with values'. 
I get a German error message - I hope the one I mention is correct translated for English.

### Steps to Reproduce

1. This is the first step
Make use of the ParentalManytoManyField in a model.
Create a new page and tick several foreignkey entries. Save that draft.

2. This is the second step, etc.
Open the preview of the page. You should see preview error page of wagtail, since the Django Form Validation throws an error, as mentioned above.

### Technical details

* Python version: 3.6.0
* Django version: 1.11
* Wagtail version: 1.10rc1
* Browser version: Chrome latest version

gasman:
Fixes #3564. Convert QueryDict to a plain dict when writing to session and vice versa when retrieving, so that multi-value fields are preserved.

### DIFF
--- a/wagtail/wagtailadmin/views/pages.py
+++ b/wagtail/wagtailadmin/views/pages.py
@@ -6,6 +6,7 @@ from django.core.exceptions import PermissionDenied
 from django.core.urlresolvers import reverse
 from django.db.models import Count
 from django.http import Http404, HttpResponse, JsonResponse
+from django.http.request import QueryDict
 from django.shortcuts import get_object_or_404, redirect, render
 from django.template.loader import render_to_string
 from django.utils import timezone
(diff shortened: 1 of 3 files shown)


wagtail/wagtailadmin/views/pages.py
```diff
@@ -581,12 +582,23 @@ def get_form(self):
...
+
         return form_class(post_data, instance=page, parent_page=parent_page)
 
     def post(self, request, *args, **kwargs):
         # TODO: Handle request.FILES.
-        request.session[self.session_key] = request.POST, time()
+
+        # Convert request.POST to a plain dict (rather than a QueryDict) so that it can be
+        # stored without data loss in session data
+        post_data_dict = dict(request.POST.lists())
```
BertrandBordage: It could be simpler by converting to tuple, then remove the `.items()` line 589.
An even simpler solution would be to convert to string using `QueryDict.urlencode()`, then directly initialise the `QueryDict` with this string when restoring data. Maybe I miss something, but it should work and it would be more compact in memory and faster to serialize/deserialize in session data.

gasman: Thanks @BertrandBordage! Given that we're already on a release candidate, I'd prefer to keep as close as possible to the existing data representation to minimise the possibility of introducing new issues, so I'll merge this as-is.

BertrandBordage: OK good, I’ll then make a pull request once 1.10 is out to make this simplification!

gasman:
Merged in 3d9b708b8e8e087292360e1526165fef2b82ed4c (master) / a7b7f44856fa82ce26da6edd43551274139e53e9 (stable/1.10.x)

PR #3582, opened 2017-05-03
Replace 'recent edits' query with a pure ORM version

gasman:
Alternative fix for #3422 as proposed by @BertrandBordage.

### DIFF
--- a/wagtail/wagtailadmin/views/home.py
+++ b/wagtail/wagtailadmin/views/home.py
@@ -2,7 +2,7 @@ from __future__ import absolute_import, unicode_literals
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
-from django.db import connections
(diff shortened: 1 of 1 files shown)


gasman:
Failing on MySQL:

`NotSupportedError: (1235, "This version of MySQL doesn't yet support 'LIMIT & IN/ALL/ANY/SOME subquery'")`

I guess we'll need to fall back on raw SQL for MySQL, then. That's a bit rubbish :-(

BertrandBordage:
There may be an ORM workaround, I’ll try to find one in the following days.

gasman:
Merged in 844cb7746cf0fbc9fbe2ef8772ad2182cc7c0c2e, as the need for a SQL fallback on MySQL isn't a blocker, and is an improvement over the current situation.

PR #3586, opened 2017-05-06
Fixed RuntimeError: dictionary changed size during iteration

martinhill:
Hi,

I noticed this RunTimeError in our logs raised in ```PreviewOnEdit.remove_old_preview_data```, introduced in wagtail 1.10. The exception will always occur if any old data is found in the session. It's a fairly simple fix, rather than mutating the session directly in the loop I use a comprehension to collect the keys to be removed.

As there wasn't an existing test to catch this I though I'd take a stab at adding one. To accomplish this, I needed to add freezegun to the test dependencies list.

Thanks!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,6 +44,7 @@ testing_extras = [
     'elasticsearch>=1.0.0,<3.0',
     'Jinja2>=2.8,<3.0',
     'boto3>=1.4,<1.5',
+    'freezegun>=0.3.8',
 
     # For coverage and PEP8 linting
     'coverage>=3.7.0',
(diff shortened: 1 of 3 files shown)


BertrandBordage:
Steps to reproduce the bug in real conditions:
- preview a page
- preview another page at least 24 hours

The only way to workaround this problem in 1.10.0 is to preview again the first previewed page.

BertrandBordage:
Merged with 9f5ce0755a96da69f2a16bbf08087f7568f44b0a & 96bade5fb428bc087d1b1c0d0a320d2de7ff78c6 (for a future 1.10.1).

PR #3588, opened 2017-05-09
Update Mixcloud oEmbed pattern, add https support.

heldinz:
As per docs at:
  https://www.mixcloud.com/developers/#embedding

### DIFF
--- a/wagtail/wagtailembeds/oembed_providers.py
+++ b/wagtail/wagtailembeds/oembed_providers.py
@@ -139,8 +139,8 @@ OEMBED_ENDPOINTS = {
     "http://video.yandex.ru/oembed.{format}": [
         "^http://video\\.yandex\\.ru/users/[^#?/]+/view/.+$"
     ],
-    "http://www.mixcloud.com/oembed/": [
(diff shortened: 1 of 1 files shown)


heldinz:
I initially made a mistake in this PR and had to force push a change. I'm hopeful that that's where the CI errors have come from. Would it be possible for an admin to retry those builds and see if it works now? I'd be very surprised if this change reduced code coverage or stopped Travis from building.

tomdyson:
Hi @heldinz, thanks for this! It looks like all the important tests have passed.

m1kola:
Hi @heldinz, congratulations on your first contribution to Wagtail :)

Looks good to me, but I've made tiny amendment: it looks like mixcloud oEmbed API also supports URLs without `www`, so I've updated pattern to make `www.` part to be optional.

I've rebased your commit in 522a383acb13f2a94961a0ab7418ea50fba28bfe, added support of optional `www.` in 4052c316680c16fba32a2bcb7c1c9ab5244dfafc and updated the contributors list changelogs in 9d64068082a1ce6fcf77251013fd4328f31e0621.

Thank you for your contribution!

heldinz:
That's great news, thanks @m1kola!

PR #3593, opened 2017-05-12
Assign persistent IDs to StreamBlock items

gasman:
This PR adds an ID field to the data model for items in a StreamField / StreamBlock. The database JSON representation of a StreamField now looks like:

    [
        {"type": "heading", "value": "Hello world", "id": "3482a437-4f50-49c4-bfbf-6db8cc8e081d"},
        {"type": "paragraph", "value": "<p>Lorem ipsum...</p>", "id": "49359c63-2100-4da7-9baa-433bb589b58b"}
    ]

IDs are assigned a random UUID on save (if not already present), are preserved in the edit form, and are exposed in the API.

The immediate use-case for this feature is for a client site that's using page save/publish signals along with the Wagtail API, to push StreamField-based page content to an external content store. Keeping track of IDs means that we can update existing records in that content store where appropriate, rather than replacing the whole lot on every publish. It could also potentially be used to make the revision comparison view smarter, since we can now definitively tell when blocks have been added, deleted or reordered, rather than having to guess based on comparing content.

### DIFF
--- a/wagtail/api/v2/tests/test_pages.py
+++ b/wagtail/api/v2/tests/test_pages.py
@@ -1026,7 +1026,10 @@ class TestPageDetailWithStreamField(TestCase):
         self.assertIn('id', content)
         self.assertEqual(content['id'], stream_page.id)
         self.assertIn('body', content)
-        self.assertEqual(content['body'], [{'type': 'text', 'value': 'foo'}])
+        self.assertEqual(len(content['body']), 1)
+        self.assertEqual(content['body'][0]['type'], 'text')
+        self.assertEqual(content['body'][0]['value'], 'foo')
+        self.assertTrue(content['body'][0]['id'])
 
     def test_image_block(self):
         stream_page = self.make_stream_page('[{"type": "image", "value": 1}]')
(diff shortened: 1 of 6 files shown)


pyMan:
Hi @kaedroho and @gasman , please how can I access the new 'id' attribute in a template? It seems not to be visible.

Thank you

gasman:
@pyMan You can access it as the `id` attribute of the `StreamChild` object (i.e. the object you get when you iterate over the StreamField value):

    {% for block in page.body %}
        <div id="{{ block.id }}">{% include_block block %}</div>
    {% endfor %}

If the block defines a template of its own (to be rendered with `{% include_block %}`), you won't be able to access the ID there, because that template only has access to the block value, not the `StreamChild`. You'd have to pass it explicitly instead:

    {% for block in page.body %}
        <div id="{{ block.id }}">{% include_block block with block_id=block.id %}</div>
    {% endfor %}

pyMan:
Thank you @gasman, I was missing to pass the id explicitly since I'm using the `include_block` template tag.

PR #3601, opened 2017-05-18
Circle CI config

kaedroho:
Replacing Drone with Circle CI.


kaedroho:
Looks like we've hit this isort bug: https://github.com/timothycrosley/isort/issues/457

kaedroho:
All issues should be fixed now!

PR #3605, opened 2017-05-21
Add ability to specify min/max block counts on StreamBlock

ebar0n:
This pull request resolves the required parameters named min_num and max_num for the StreamBlock, see issue #2514.

### DIFF
--- a/docs/topics/streamfield.rst
+++ b/docs/topics/streamfield.rst
@@ -441,8 +441,9 @@ Since ``StreamField`` accepts an instance of ``StreamBlock`` as a parameter, in 
 .. code-block:: python
 
     class HomePage(Page):
-        carousel = StreamField(CarouselBlock())
+        carousel = StreamField(CarouselBlock(max_mun=10))
 
+The keyword arguments ``min_num`` and ``max_mun`` are accepted.
 
 .. _streamfield_personblock_example:
 
--- a/wagtail/wagtailcore/blocks/stream_block.py
+++ b/wagtail/wagtailcore/blocks/stream_block.py
@@ -13,6 +13,7 @@ from django.utils import six
 from django.utils.encoding import python_2_unicode_compatible
 from django.utils.html import format_html_join
 from django.utils.safestring import mark_safe
+from django.utils.translation import ugettext as _
 
 from wagtail.wagtailcore.utils import escape_script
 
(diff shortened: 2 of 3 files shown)


ebar0n:
Hi, I have added the extra parameter ```min_max_fields```
In order to validate the limits of a specific field within the block.
It is useful, to limit that only 2 videos can be added, in a block and where they find more elements.

What do you think of that? @m1kola

m1kola:
I like the idea of being able to set min and max values at any level, but it seems to me that this logic should be handled somewhere else (not is `StreamBlock`).

At the moment, there is one basic class called `FieldBlock` for all simple blocks (like `CharBlock`, `TextBlock`, `URLBlock` and many others) and there [are three basic classes](http://docs.wagtail.io/en/v1.10.1/topics/streamfield.html#structural-block-types)  for complex blocks: `StructBlock`, `ListBlock` and `StreamBlock`. I think we should handle each type as it was done for `StreamBlock` in 4ee94e47a6074fcd281d4da7d1a2671d72df5bcb.

I mean something like this:

```python
class HomePageBodyBlock(blocks.StreamBlock):
    # Allow only 10 images per HomePageBodyBlock
    image = ImageChooserBlock(max_num=10)

    # Allow only 10 quotations per HomePageBodyBlock
    quotation = blocks.StructBlock([
        # What should we do with max_num in StructBlocks? Just ignore?
        ('text', blocks.TextBlock(max_num=10)),
        ('author', blocks.CharBlock()),
    ], max_num=10)

    # No limit for single videos
    video = EmbedBlock()

    # Allow only 2 lists with videos. Each list can contain up to 5 videos.
    list_of_videos = blocks.ListBlock(
        EmbedBlock(max_num=5)
    , max_num=2)

    class Meta:
        icon='cogs'

class HomePage(Page):
    # Allow only one HomePageBodyBlock per the body StreamField
    body = StreamField(HomePageBodyBlock(max_num=1))
```

@gasman I think we need your opinion here.

ebar0n:
@m1kola I understand you

Was useful for my, for the fact that I can have a single block of components, and use it on many pages with different restrictions.

gasman:
@m1kola I think the `min_num` / `max_num` definitions should be on the container block (i.e. the way that this PR does it), not the child. A business rule such as "this StreamBlock can contain up to 5 RichTextBlocks" is describing the behaviour of the StreamBlock, not the RichTextBlock.

burakk:
Can we use max and min num in wagtail 1.12.2?

gasman:
@burakk Yep, see the documentation at http://docs.wagtail.io/en/v1.12.2/topics/streamfield.html#streamblock.

Issue #3565, opened 2017-04-24
sitemap.xml lastmod should use date last published


mikemonteith:
### Issue Summary

sitemap.xml's `<lastmod>` should show the date that a page was last published to live, not the date that it was last saved as a draft.

### Steps to Reproduce

1. Create a page
2. See `<lastmod>` value in sitemap.xml
3. Make a change to the page (save as draft)
4. `<lastmod>` value in sitemap.xml has changed (remember to refresh the sitemap cache)

### Relevant code
https://github.com/wagtail/wagtail/edit/master/wagtail/contrib/wagtailsitemaps/sitemap_generator.py#L15

ababic:
Related issue: #771

PR #3608, opened 2017-05-23
Add last_published_at to Page model

### LINKED ISSUE #3578: Add 'last_published_at' field to Page model
Previously discussed in #577 and #771, but doesn't seem to have been raised as a ticket in its own right up to now. Necessary for #3565.

#3302 is related - it adds a foreign key from `Page` to the current live revision (if any), which would give you access to the revision's `created_at` date - although that isn't an exact match for last published date, because pages that go through moderation or scheduled publishing may have an actual publishing date that's some time after the revision was created.

gasman:
Add `last_published_at` field to Page model, and use it for the `lastmod` field of sitemap.xml. Fixes #3578 and #3565.


m1kola:
There are some test failures with Elasticsearch 5. Other than that, everything looks good to me.

gasman:
@m1kola Thanks for the review! The Elasticsearch 5 errors are happening everywhere, not just on this PR - it seems to be a configuration change that happened on Travis about a week ago :-(

kaedroho:
Could we add a data migration with the following query?

```python
Page.objects.filter(has_unpublished_changes=False).update(last_published_at=F('latest_revision_created_at'))
```

gasman:
@kaedroho Good idea! Will add that.

gasman:
Rebased, and added data migration as suggested - @kaedroho / @m1kola, please can you quickly review the latest commit?

m1kola:
@gasman All looks good to me. Feel free to merge when CI done with tests.

PR #3610, opened 2017-05-24
Clarify "Steps to reproduce" in new issue template

gasman:
Bug reporters are frequently not making the connection that "Steps to reproduce" means "steps for someone else to independently reproduce this bug", not "steps that I go through on my project to see this bug". Added some example text to gently steer people towards the level of detail we want...

Also trimmed down the "Wagtail version" item under "Technical details", since we don't really need to tell people four different ways to do the same thing. (We don't want people to mentally tune this text out, so I think it's a good habit to remove something for every new thing we add)

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -6,11 +6,11 @@ A summary of the issue.
 
 ### Steps to Reproduce
 
+It's essential that you provide enough information for someone else to replicate the problem you're seeing. Simply describing something that's broken on your current project is not enough!
+
 1. (for example) Start a new project with `wagtail start myproject`
 2. Edit models.py as follows...
 3. ...
-
-(It's essential that you provide enough information for someone else to replicate the problem you're seeing. Simply describing something that's broken on your current project is not enough!)
 
 Any other relevant information. For example, why do you consider this a bug and what did you expect to happen instead?
 
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -17,6 +17,6 @@ Any other relevant information. For example, why do you consider this a bug and 
 ### Technical details
 
 * Python version: Run `python --version`.
-* Django version: Look in your requirements.txt, or run `pip list | grep -i django`.
-* Wagtail version: Hover over the Wagtail bird in the admin, or run `pip list | grep -i wagtail`.
+* Django version: Look in your requirements.txt, or run `pip show django | grep Version`.
+* Wagtail version: Hover over the Wagtail bird in the admin, or run `pip show wagtail | grep Version:`.
 * Browser version: You can use http://www.whatsmybrowser.org/ to find this out.
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -6,8 +6,11 @@ A summary of the issue.
 
 ### Steps to Reproduce
 
-1. This is the first step
-2. This is the second step, etc.
+1. (for example) Start a new project with `wagtail start myproject`
+2. Edit models.py as follows...
+3. ...
+
+(It's essential that you provide enough information for someone else to replicate the problem you're seeing. Simply describing something that's broken on your current project is not enough!)
 
 Any other relevant information. For example, why do you consider this a bug and what did you expect to happen instead?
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -6,14 +6,17 @@ A summary of the issue.
...
+(It's essential that you provide enough information for someone else to replicate the problem you're seeing. Simply describing something that's broken on your current project is not enough!)
 
 Any other relevant information. For example, why do you consider this a bug and what did you expect to happen instead?
 
 ### Technical details
 
 * Python version: Run `python --version`.
 * Django version: Look in your requirements.txt, or run `pip list | grep -i django`.
-* Wagtail version: Hover over the Wagtail bird in the admin, or look in the Settings menu, look in your requirements.txt, or run `pip list | grep -i wagtail`.
+* Wagtail version: Hover over the Wagtail bird in the admin, or run `pip list | grep -i wagtail`.
```
BertrandBordage: To avoid listing Wagtail third apps and potentially confuse, we can use `pip show wagtail | grep Version:`.

.github/ISSUE_TEMPLATE.md
```diff
@@ -6,14 +6,17 @@ A summary of the issue.
 
 ### Steps to Reproduce
 
-1. This is the first step
-2. This is the second step, etc.
+1. (for example) Start a new project with `wagtail start myproject`
+2. Edit models.py as follows...
+3. ...
+
+(It's essential that you provide enough information for someone else to replicate the problem you're seeing. Simply describing something that's broken on your current project is not enough!)
```
jjanssen: I think we could remove the parenthesis around this and set this as a paragraph as an intro above the numbered list. And I'm not sure about the 2nd sentence maybe we can rephrase it to something like: a broken functionality can also be a combination of integrating a project with Wagtail.

.github/ISSUE_TEMPLATE.md
```diff
@@ -6,14 +6,17 @@ A summary of the issue.
```
thibaudcolas: Before "Steps to reproduce" in the issue summary we should further direct the type of answer we are after – not any summary, we want to understand what is occuring, and what is expected.

.github/ISSUE_TEMPLATE.md
thibaudcolas: Yes, this should be the intro to this section.

.github/ISSUE_TEMPLATE.md
```diff
@@ -6,14 +6,17 @@ A summary of the issue.
 
 ### Steps to Reproduce
 
-1. This is the first step
-2. This is the second step, etc.
+1. (for example) Start a new project with `wagtail start myproject`
+2. Edit models.py as follows...
```
thibaudcolas: We should provide further examples (worded as questions?) to direct the answers. For example,

```
4. What did you change in the settings?
5. What content did you create in the admin?
```

.github/ISSUE_TEMPLATE.md
```diff
@@ -6,14 +6,17 @@ A summary of the issue.
...
+3. ...
+
+(It's essential that you provide enough information for someone else to replicate the problem you're seeing. Simply describing something that's broken on your current project is not enough!)
 
 Any other relevant information. For example, why do you consider this a bug and what did you expect to happen instead?
```
thibaudcolas: This should be issue summary, presumably "steps to reproduce" is only relevant once it's determined (rightly or wrongly) that this is a bug.

.github/ISSUE_TEMPLATE.md
thibaudcolas: How much work would it be to create a dedicated command line tool (`wagtail doctor`) or page in the admin interface that gathers all of this info directly?

gasman: @BertrandBordage Good idea! Updated in 635ffddf70381a835b2f82e0cd8e06e48286a092.

@thibaudcolas Probably more work than it's worth, really :-) I don't think this is a particularly onerous section to fill in, and the people who are too lazy to fill it in probably won't bother running a command either...

.github/ISSUE_TEMPLATE.md
gasman: Have moved this paragraph up as suggested, in aea45e664f6820f47151accdd0a88030fdabb2da.

Re the suggested text change, I'm keen to stick as closely as possible to the format "here's the information we want you to supply". I realise it might come across as a bit blunt, and there's a natural motivation for us to make it more friendly and explain _why_ we're asking for information in a particular format - but ultimately this is supposed to be a form for people to fill in, not a "how to write a good bug report" advice document. If it gets too wordy, there's a risk that the people who _are_ good bug reporters will get annoyed by all the boilerplate text that they have to delete - and the people who _aren't_ will end up skimming over it or being put off entirely.

For that reason, my aim here is to try and fit in as much "education" as possible through the structure and wording of the form itself... inevitably there'll be things we'd like to say that don't fit into that format, but we have the ["guidelines for contributing"](https://github.com/wagtail/wagtail/blob/master/CONTRIBUTING.md) document for that.

Issue #692, opened 2014-10-10
Option to set show_in_menus default to True


mattots:
At the moment the show_in_menus field defaults to False so the user has to manually tick the 'Show in menus' checkbox in the Promote panel for each new page that is created. However, I can't think of any sites that I've ever been involved with developing where the default scenario was for new pages to be excluded from menus. It would seem more logical then for the default to be True, with the onus being on the user to unselect the checkbox if they want to hide the page.

(Alternatively it might actually make more sense if the checkbox was labelled 'Hide from menus' and was unchecked by default.)

gasman:
The core team has discussed this today - we're in favour of adding an option on a per-page-model basis to specify whether "show in menu" is initially checked for that page type.

lb-:
Hey @gasman I would like to take this on.

Here is my proposed solution (wanted to check before I submitted a PR).

Please advise any changes or issues to be resolved with this, mainly around the naming of `show_in_menus_default` and whether it should be a class variable, method (with self) or a static method.

I have not yet got this tested with a custom testapp model, but will ensure that is in place before submitting.

[wagtail/wagtailcore/models.py](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailcore/models.py#L336)
```
    # can be overridden per page model basis
    # Note: I assumed a simple variable would be sufficient, not a function, please advise.   
    show_in_menus_default = False
    show_in_menus = models.BooleanField(
        verbose_name=_('show in menus'),
        default=show_in_menus_default,
        help_text=_("Whether a link to this page will appear in automatically generated menus")
    )
```


[wagtail/wagtailcore/tests/test_page_model.py](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailcore/tests/test_page_model.py)
```
# will put at the end of the file
class TestShowInMenusDefaultOption(TestCase):
    """
    This tests that a page model can define the default for 'show_in_menus'
    """
    fixtures = ['test.json']

    def test_show_in_menus_default(self):
        # Check that the model default is always False
        page = Page(title='My Awesome Page', slug='my-awesome-page')
        show_in_menus_field = page._meta.get_field('show_in_menus')

        # Check that the model's field accesses the model attribute
        self.assertFalse(show_in_menus_field.default)

        # Check that the vale is available
        self.assertFalse(page.show_in_menus_default)

    def test_show_in_menus_default_override(self):
        Page.show_in_menus_default = True
        # This should be the different page model
[...]

PR #3615, opened 2017-05-26
Show in menu default

lb-:
Refer to issue #692 - this implements the recommended solution with test and updated docs.


docs/reference/pages/model_reference.rst
```diff
@@ -80,6 +80,8 @@ Database fields
 
         Toggles whether the page should be included in site-wide menus.
 
+        Deafults to ``False`` and can be overridden on the model with ``show_in_menus_default=True``.
```
BertrandBordage: Typo on “Defaults” and there should be a space around the equal sign.

wagtail/wagtailcore/models.py
```diff
@@ -367,6 +369,9 @@ def __init__(self, *args, **kwargs):
             # set content type to correctly represent the model class that this was
             # created as
             self.content_type = ContentType.objects.get_for_model(self)
+        if 'show_in_menus' not in kwargs:
+            # if the value is not set on submit, refer to the model's setting
+            self.show_in_menus = self.show_in_menus_default
```
BertrandBordage: The broken tests are probably caused by these line. Try replacing all the method with the following and run the test suite one your computer to check:
```python
        super(Page, self).__init__(*args, **kwargs)
        if self.pk is None:
            if self.content_type_id is None:
                # this model is being newly created rather than retrieved from the db;
                # set content type to correctly represent the model class that this was
                # created as
                self.content_type = ContentType.objects.get_for_model(self)
            if 'show_in_menus' not in kwargs:
                # if the value is not set on submit, refer to the model's setting
                self.show_in_menus = self.show_in_menus_default
```

lb-:
@BertrandBordage I see where I went wrong, thanks - I will fix that up and the typos.

BertrandBordage:
Uh, you committed a ton of test media files.

lb-:
@BertrandBordage github app fail, I thought I had unticked all of those, they are still not committed locally. ill make a new commit

lb-:
@BertrandBordage let me know if that is suitable or if I should make a new branch or something else with git.

docs/reference/pages/model_reference.rst
```diff
@@ -80,6 +80,8 @@ Database fields
 
         Toggles whether the page should be included in site-wide menus.
 
+        Deafults to ``False`` and can be overridden on the model with ``show_in_menus_default = True``.
```
BertrandBordage: Still the typo on `Defaults`.

gasman:
Thanks @lb- and @BertrandBordage! Merged in 18cb0245fc0329e3218bd6f372a961557d4be713.

PR #3618, opened 2017-05-28
Use runInBand to make Jest tests faster in CI env

thibaudcolas:
See https://facebook.github.io/jest/docs/troubleshooting.html#tests-are-extremely-slow-on-docker-and-or-continuous-integration-ci-server for further details.


thibaudcolas:
This seems to make this step ±5x faster in CircleCI.

PR #3625, opened 2017-06-01
refactor: #3621 Generic view classes should inherit from TemplateView etc

### LINKED ISSUE #3621: Any reason why the generic view classes are not inheriting from TemplateView and so on?
I'm talking about the classes located in [`wagtailadmin.views.generic`](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailadmin/views/generic.py).

They all are inheriting from `View`. They would be easier to use if inheriting from TemplateView (e.g. adding extra context data with `get_context_data`).

The Create/Edit/Delete views could also benefit from inheriting at least from `FormMixin`. Even better if they would inherit from `BaseCreateView` and so on.

So, any reason why this was not done? I'm willing to do the PR if needed.

Thanks.

Gagaro:
Fixes #3621

A few comments:

* To be consistent, I added `get_success_message` and `get_error_message` for classes where it is expected.
* I deprecated `instance` attribute in favour of `object` which is the standard attribute name used in Django CBV (again for consistency).
* I deprecated `get_context` in favour of `get_context_data` for the same reason.
* Django CBV expect a keyword argument in the URL, while Wagtail always use a positional one. I added the `pk` kwarg manually if it is not present. I'm not sure what would be the best approach here (keeping it like that, or adding a deprecation warning and redoing all wagtail URLs).
* Of course, all current views implementation are still compatible as they are.

Tell me what you think about all that. Thanks!

### DIFF
--- a/wagtail/wagtailadmin/views/generic.py
+++ b/wagtail/wagtailadmin/views/generic.py
@@ -1,10 +1,14 @@
 from __future__ import absolute_import, unicode_literals
 
+import warnings
+
 from django.core.urlresolvers import reverse
-from django.shortcuts import get_object_or_404, redirect, render
+from django.shortcuts import redirect
 from django.utils.translation import ugettext as _
 from django.utils.translation import ugettext_lazy
-from django.views.generic.base import View
+from django.views.generic.base import TemplateResponseMixin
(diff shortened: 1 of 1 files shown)


gasman:
Looks good, thanks! Merged in bb4ea5922c3989eb486bfbf11ed1092d0719e41e.

PR #3626, opened 2017-06-01
Improve consistency of ModelAdmin's view initialisation methods

### LINKED ISSUE #3392: Improve consistency of how views in ModelAdmin are initialised

ababic:
Fixes #3392. Partly replaces PR #3542 

@gasman I have tried to do the very least here to achieve what I wanted with the init functions. If these changes can be reviewed, I'll work on making the other consistency updates in another PR at a later date. 

* Do the tests still pass? 👍 

### DIFF
--- a/wagtail/contrib/modeladmin/options.py
+++ b/wagtail/contrib/modeladmin/options.py
@@ -306,28 +306,26 @@ class ModelAdmin(WagtailRegisterable):
             return found_fields
         return self.inspect_view_fields
 
-    def index_view(self, request):
+    def index_view(self, request, **kwargs):
         """
         Instantiates a class-based view to provide listing functionality for
         the assigned model. The view class used can be overridden by changing
         the 'index_view_class' attribute.
         """
-        kwargs = {'model_admin': self}
-        view_class = self.index_view_class
-        return view_class.as_view(**kwargs)(request)
+        view = self.index_view_class.as_view(model_admin=self)
+        return view(request, **kwargs)
 
-    def create_view(self, request):
+    def create_view(self, request, **kwargs):
         """
         Instantiates a class-based view to provide 'creation' functionality for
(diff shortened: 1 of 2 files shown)


ababic:
Hi @gasman. I've had some second thoughts about using `SingleObjectMixin` here. We're really not gaining too much from it in the way of functionality, and the mixing of `instance` and `object` terminology on the same class bothers me. If we're going to be making a shift to using `object`more consistently a later date, we can likely apply that here as part of rebasing modeladmin's views on the views in `wagtail/wagtailadmin/views/generic.py`

gasman:
Hi @ababic,
Really sorry to bounce this back again, but with the `SingleObjectMixin` change being reverted I'm not able to decipher what the net effect of this PR is meant to be. If you're making changes A, B, C and D and then reverting B, then I need to be able to look at changes A, C and D independently of B.

I've tried to do that by cherry-picking the non-`SingleObjectMixin`-related changes (which I _think_ are 9eda12c, a27ba7e, 2fed250) onto a new branch https://github.com/gasman/wagtail/commits/cleanup/modeladmin-view-init-consistency, but the tests are now failing, so it looks like these changes are too intertwined for that to work. Please could you create a new PR consisting of just the cleanup steps that remain once `SingleObjectMixin` is off the table?

(On the subject of `instance` vs `object`, you may want to take a look at #3625 which refactors `wagtail/wagtailadmin/views/generic.py` to use Django's higher-level CBVs, and standardises on `object` along the way. Not sure if that has a bearing on the direction you want to take things here...?)

ababic:
@gasman Okay. Not to worry. I can do that. I have been keeping track of #3626 - but I was waiting to see what the feedback was there RE renaming. I'm in agreement that a shift to `object` instead of `instance` would be a good move. Either way, I'll deal with that separately once you guys are happy with that part of `wagtailadmin`

Issue #3607, opened 2017-05-23
Tidy up new React + API explorer


thibaudcolas:
Tracking issue for the follow-up work to #3012, #3385. We now need to tidy up the last few pieces before the 1.11 release.

__Help is very welcome so we can release this on time__. This is an idiomatic React + Redux codebase, so a great way to get up to speed with this stack if beginners want to get involved.

### Remaining issues

- [ ] On mobile screen sizes, the last items in the menu require page scrolling as well as menu scrolling.
- [x] On mobile screen sizes, top part of explorer behind address bar.
- ~[ ] On mobile, page overlay doesn't show.~ (desired)
- [x] On touch screens, tweak hover/active styles to not show as often.
- [x] On touch screens, explorer dismissal on touch feels a bit too fast?
- [x] IE11 - Menu items are slightly wider than their container.
- [x] IE11 - Menu items load stacked onto one another.
- ~[ ] Windows – Scrollbars should be hidden during transitions.~ (no, same across all browsers)
- [x] On mobile, Close explorer link does not work, clicking it just brings you back to the first explorer 'home page'
- [x] On mobile, Click on “Pages” the page scrolls down, so I scroll up, then click on “Close explorer”, which closes the explorer but reopens it immediately, while scrolling down again.
- ~[ ] On mobile, on first load of a menu level, focus ends on "Close explorer" button.~ (desired)
- [x] On mobile, explorer shows a shadow even though there is no layering on the page.
- [x] Performance review (CPU and memory, of both the Redux and React hierarchies)

### Features

- [ ] Text truncation of the page title after 60-70 characters;
- ~[ ] Finishing adding the setting to customise the ordering / API behavior – I've added this as a global variable initialised in the templates, but it's not hooked into the Django settings yet (I don't know how to do that)~

## Backlog

- [ ] Code reviews

### Documentation

- [ ] React/Redux code styleguide (action types with keyMirror, FSA or not, file structure)
- [ ] Integrate React components in UI styleguide

### Features

- [ ] Better translation with jsi18n (https://docs.google.com/document/d/1WImkcI1ORB5nj3FeNuVnkPIcjJTIKoMIhCV-Hqj2IEY)
- ~[ ] Better date format client-side (`moment-django` FTW!)~

### Tooling

- [ ] Add front-end test coverage to Codecov
- [x] Update to Webpack 2
- [x] Update to latest Jest
- [x] Update to latest React
[...]

PR #3638, opened 2017-06-07
Node tooling cleanup

thibaudcolas:
Part of tidying up before releasing the new explorer in #3607.

I'd have liked to update to `babel-preset-env`, but it requires Node 6 / npm 3, so this will have to wait.

This PR removes about ± 100kb from the gzipped bundle size 🎉.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,6 @@
     ]
   },
   "devDependencies": {
-    "babel-cli": "^6.24.1",
     "babel-core": "^6.24.1",
     "babel-jest": "^20.0.3",
     "babel-loader": "^7.0.0",
@@ -40,14 +39,12 @@
     "eslint-plugin-import": "^1.8.1",
     "eslint-plugin-jsx-a11y": "^1.5.3",
     "eslint-plugin-react": "^5.2.2",
-    "exports-loader": "^0.6.4",
     "gulp": "~3.8.11",
     "gulp-autoprefixer": "~3.0.2",
     "gulp-rename": "^1.2.2",
     "gulp-sass": "~2.3.1",
     "gulp-sourcemaps": "~1.5.2",
     "gulp-util": "~2.2.14",
-    "imports-loader": "^0.7.1",
     "jest": "^20.0.4",
     "mustache": "^2.2.1",
     "react-addons-perf": "^15.4.2",
--- a/package.json
+++ b/package.json
@@ -6,13 +6,6 @@
   "engines": {
     "node": ">= 4"
   },
-  "browser": {},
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -39,17 +39,14 @@
...
     "gulp": "~3.8.11",
     "gulp-autoprefixer": "~3.0.2",
     "gulp-rename": "^1.2.2",
     "gulp-sass": "~2.3.1",
     "gulp-sourcemaps": "~1.5.2",
     "gulp-util": "~2.2.14",
-    "imports-loader": "^0.7.1",
     "jest": "^20.0.4",
     "mustache": "^2.2.1",
-    "react-addons-test-utils": "^15.4.2",
```
jjanssen: Thanks 👍  missed out on that one during the migration.

jjanssen:
Overall this PR looks to good to me, although I have a few questions about the changes in general and wondering what perspective we have on this for the long-term. I see that we replaced `redux-action` + `babel-polyfills` with our own utilities for the benefit of filesize. 

Although the changes in this particular use-case are really minimal, I can imagine this approach would probably raise the bar in contributing and maintaining for the future where we would need a philosophy in explaining these utils. 

Wouldn't something like webpack v2's [tree-shaking](https://webpack.js.org/guides/tree-shaking/) make more sense in order to keep adopting well maintained packages and keep our doors open for easy contributing.

thibaudcolas:
I agree with you 100%, but decided to make those changes anyway to give the new explorer the best chance to work well on mobile where JS parsing & execution resources are more limited.

In the future I'd like to revert this. Ideally `babel-polyfill` / `core-js` should be combined with `babel-preset-env` to produce just the required polyfills based on configuration alone, unfortunately this only works on node 6 / npm 3 and above...

In the future (I hope not too far from now), I'd like to move all of the JS to Webpack so it can be minified (and potentially bundled), and I hope that by making those performance-minded changes we will have a bigger budget available for the bigger amount of JS overall.

NB: tree shaking - sounds great on principle, and I thought it was supposed to be in place in `redux-actions`, but for some reason when I checked the bundle composition it still loaded ± 20kb of JS for the one function we use 😞. Not sure if I missed something / made a mistake, if https://github.com/th0r/webpack-bundle-analyzer doesn't understand that nuance, or if the gain is there but smaller than I had imagined.

jjanssen:
Hi Thibaud, could you rebase this PR with the current `master`?

thibaudcolas:
@jjanssen done 👍

jjanssen:
All good, tested and merged 👍

PR #3640, opened 2017-06-07
Fix document chooser template to account for hidden fields

jeffreyckchau:
Hello,

This is a relatively minor change for consistency sake. 

We have some custom code in our project that requires us to hide the collection field in upload forms to suit some of our multi-tenant needs. I noticed that for documents specifically, the template for the upload section doesn't handle hidden fields very well, but the images chooser does. So i've made the documents template similar to the images template.


gasman:
Thanks @jeffreyckchau! Merged in 0c9ad90edd1fca8b8ef47a96dc8fe72f270eb869.

PR #3643, opened 2017-06-08
Allow specifying a project_template in "wagtail start"-comment

ramonakira:
Added the option to specify a custom project_template to the "wagtail start"-command. Also renamed the optional directory to destination and make it be an option and not an arg.

### DIFF
--- a/wagtail/bin/wagtail.py
+++ b/wagtail/bin/wagtail.py
@@ -7,14 +7,14 @@ from optparse import OptionParser
 from django.core.management import ManagementUtility
 
 
-def create_project(parser, options, args):
+def create_project(parser, options, arguments):
     # Validate args
-    if len(args) < 2:
+    if len(arguments) < 2:
         parser.error("Please specify a name for your Wagtail installation")
-    elif len(args) > 2:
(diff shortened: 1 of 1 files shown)


wagtail/bin/wagtail.py
```diff
@@ -30,14 +36,17 @@ def create_project(parser, options, args):
...
+    else:
+        print("Creating a Wagtail project called %(project_name)s" % {'project_name': project_name})  # noqa
 
     # Create the project from the Wagtail template using startapp
 
     # First find the path to Wagtail
     import wagtail
     wagtail_path = os.path.dirname(wagtail.__file__)
-    template_path = os.path.join(wagtail_path, 'project_template')
+    template_path = os.path.join(wagtail_path, 'project_template' if project_template is None else project_template)
```
gasman: Is this addition really meant to be inside the `os.path.join(...)`? That means that your template path has to be specified relative to the Wagtail library root, which surely isn't what you want...

ramonakira:
Thanks for your reply gasman, we've gone a different route using a separate package https://github.com/leukeleu/wagtail-startproject

PR #3644, opened 2017-06-08
View restrictions for documents

gasman:
Rebased and refactored from #3245, with the following changes:

* Removed support for permission restrictions on images - universally changing images to be served through Django would have knock-on effects that we can't reasonably deal with at this time. This in turn cleans up a lot of the front-end code, since it can be tuned specifically for documents rather than trying to generically handle "things that exist in collections":
  * the `before_serve_from_collection` hook is now `before_serve_document` (and is documented)
  * the hook function and password-required view are now part of `wagtaildocs` rather than `wagtailcore`
  * documents now have their own password_required template (so it doesn't erroneously say "You need a password to access this page")
* PageViewRestrictions and CollectionViewRestrictions now use distinct session keys (necessary to avoid ID collisions)
* `wagtailadmin/collections/edit.html` inherits from `wagtailadmin/generic/edit.html` again


docs/releases/1.11.rst
```diff
@@ -110,6 +110,12 @@ See also: :meth:`wagtail.wagtailcore.models.Page.get_url_parts`, :meth:`wagtail.
 :meth:`wagtail.wagtailcore.models.Page.get_full_url`, and :meth:`wagtail.wagtailcore.models.Page.relative_url`
 
 
+"Password required" template for documents
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This release adds the ability to password-protect documents as well as pages. The template used for the "password required" form is distinct from the one used for pages; if you have previously overridden the default template through the ``PASSWORD_REQUIRED_TEMPLATE`` setting, you may wish to provide a corresponding template for documents through the setting ``DOCUMENT_PASSWORD_REQUIRED_TEMPLATE``. See: :ref:`private_pages`
+
```
tomdyson: "This feature was developed by Ulrich Wagner and Matthew Westcott"

wagtail/wagtailadmin/templatetags/wagtailadmin_tags.py
```diff
@@ -134,6 +138,25 @@ def page_permissions(context, page):
 
 
 @assignment_tag(takes_context=True)
+def test_collection_is_public(context, collection):
+    """
+    Usage: {% test_collection_is_public collection as is_public %}
+    Sets 'is_public' to True iff there are no collection view restrictions in place
```
tomdyson: iff there are -> if there are

docs/releases/1.11.rst
gasman: That will go in the main release note (which hasn't been written yet) - this is just the "upgrade considerations" section.

tomdyson: I see, sorry

wagtail/wagtailadmin/templatetags/wagtailadmin_tags.py
gasman: "iff" is shorthand for "if and only if".

kaedroho:
The code looks good to me, and it works pretty well. Just a few minor UI issues:

- The privacy button is currently at the top of the form, I think it'll look better at the top right where it is on pages:

button-should-move

- It's a little strange that you can make a whole collection "private" but that restriction doesn't apply to images. If we extend this to include images in the future, we could have a big backwards compatibility problem to deal with.

- It would be nice to have padlock icons next to private collections and documents in listings, dropdowns and choosers so editors can see if the document they're adding onto a page has a restriction.

gasman:
> The privacy button is currently at the top of the form, I think it'll look better at the top right where it is on pages

Can't see a good way to do that without either refactoring wagtailadmin/shared/header.html (to allow it to be customised by overriding blocks in wagtailadmin/collections/edit.html) or creating new CSS styles (for a 'right justified but not part of the header' block). I'd rather like to do the former, but it's out of scope for this change.

> It's strange that you can make a whole collection "private" but that restriction doesn't apply to images. If we extend this to include images in the future, we could have a big backwards compatibility problem to deal with.

As discussed in #3245, if we do implement image protection in future then it certainly won't be something that we unilaterally enable by default in a new Wagtail release, since it requires changing image serving to happen via Django (with "fun" implications for caching, cloud storage and so on).

> It would be nice to have padlock icons next to private collections and documents in listings, dropdowns and choosers so editors can see if the document they're adding onto a page has a restriction.

I agree. Feel free to open a new ticket about that :-)

gasman:
Merged in 4e5fff7e41a562438daf5ab9818d4ae1c9e420c6

PR #3645, opened 2017-06-08
Fix mobile explorer scrolling issue

thibaudcolas:
This is the last issue I wanted to fix before the explorer is released. On small screens, the bottom of the scrollable list wouldn't be readily accessible by scrolling within the explorer – users had to navigate two scrolling contexts.

There is one remaining issue – Wagtail doesn't prevent the page from scrolling when its menu / explorer is open on mobile, which means users may still get lost between the two scrolling areas. This also means that browsers with floating UIs (Mobile Safari's top and bottom browser chrome, Android Chrome's top browser chrome) are likely to crop some of the explorer items behind view. To fix that one-last-thing would require quite a big refactor of the CSS for the navigation, so I'll leave this as a future enhancement.

(I also took this PR as an opportunity to do a bit more refactoring, and fix the linting warnings)

## Tested on

- Browser	Device/OS	Version(s)
- [x] - Mobile Safari	iOS Phone	10
- [x] - Mobile Safari	iOS Phone	9
- [x] - Mobile Safari	iOS Tablet	10
- [x] - Mobile Safari	iOS Tablet	9
- [x] - Chrome	Android	56
- [x] - Chrome	Android	55
- [x] - IE	Desktop	11
- [x] - Chrome	Desktop	56
- [x] - Chrome	Desktop	55
- [x] - MS Edge	Desktop	15
- [x] - MS Edge	Desktop	14
- [x] - Firefox	Desktop	52
- [x] - Firefox ESR	Desktop	45.8
- [x] - Safari	macOS	10.1
- [x] - Safari	macOS	9.1

### DIFF
--- a/client/src/api/admin.js
+++ b/client/src/api/admin.js
@@ -24,7 +24,6 @@ export const getPageChildren = (id, options = {}) => {
     url += `&offset=${options.offset}`;
   }
 
-  // TODO To remove once we are done testing this.
   url += ADMIN_API.EXTRA_CHILDREN_PARAMETERS;
 
   return get(url);
--- a/client/src/components/Explorer/Explorer.scss
+++ b/client/src/components/Explorer/Explorer.scss
@@ -7,8 +7,15 @@ $menu-footer-height: 50px;
 
 @import 'ExplorerItem';
 
+.explorer__wrapper {
+    flex: 1;
+    display: flex;
+}
+
 .explorer {
     width: 100%;
+    display: flex;
+    flex-direction: column;
     position: relative;
     top: 0;
     left: 0;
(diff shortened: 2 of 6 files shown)


thibaudcolas:
@jjanssen this should be the last one, could you have a look tomorrow?

#3638 is ready as well – if you think some of those changes are too big to be rushed in, let me know which and I'll cherry-pick the rest to be merged in.

jjanssen:
iOS9 is showing difficulties with long page names.
image

jjanssen:
If we are able to solve this one we're ready for merge since I'm ok with all the changes. Took the additional tests to my account but I don't have Safari macOS 10 and Chrome Android 56 available.

thibaudcolas:
I haven't been able to fix this. Looked at this on desktop Safari 9.1 (this should've been the browser testing target, not "10") and the issue is there as well.

This PR relates to another part of the code so the issue likely is present with or without it – I'll merge this, and create a separate issue, and ask for Flexbox expertise on Slack.

PR #3647, opened 2017-06-09
Admin view sets (rebase of #2113)

gasman:
Rebase of #2113 on latest master, with the following additions:

* Fixed regressions on wagtailsites (index page and delete button on edit view)
* Call super on get_urlpatterns as per https://github.com/wagtail/wagtail/pull/2113#discussion_r102281491
* Require explicit `form_fields` or `exclude_form_fields` when using the default `get_form_class`
* Provide a `url_prefix` attribute so that wagtailsites can remain at `/admin/sites/` instead of moving to `/admin/wagtailsites/`


wagtail/wagtailadmin/viewsets/model.py
```diff
@@ -0,0 +1,99 @@
...
+class ModelViewSet(ViewSet):
+    icon = ""
+
+    index_view_class = generic.IndexView
+    add_view_class = generic.CreateView
+    edit_view_class = generic.EditView
+    delete_view_class = generic.DeleteView
+
+    def get_url_name(self, view_name):
+        return self.name + ':' + view_name
```
BertrandBordage: Shouldn’t this method be in `ViewSet`?

wagtail/wagtailadmin/viewsets/model.py
```diff
@@ -0,0 +1,99 @@
...
+            fields=fields,
+            exclude=exclude
+        )
+
+    def get_urlpatterns(self):
+        return super(ModelViewSet, self).get_urlpatterns() + [
+            url(r'^$', self.index_view, name='index'),
+            url(r'^new/$', self.add_view, name='add'),
+            url(r'^(\d+)/$', self.edit_view, name='edit'),
+            url(r'^(\d+)/delete/$', self.delete_view, name='delete'),
```
BertrandBordage: I’m not sure the primary key pattern is right, here.
A primary key can be something else than an integer.
And shouldn’t the capturing group be named, like `(?P<pk>[^/]+)`?

wagtail/wagtailadmin/wagtail_hooks.py
```diff
@@ -8,6 +8,7 @@
 from wagtail.wagtailadmin.navigation import get_explorable_root_page
 from wagtail.wagtailadmin.search import SearchArea
 from wagtail.wagtailadmin.utils import user_has_any_page_permission
+from wagtail.wagtailadmin.viewsets import viewsets
```
BertrandBordage: Maybe it would be clearer to be able to import the `viewsets` instance directly from `wagtail.wagtailadmin`?
Or rename `viewsets` to `viewset_registry`, something like that.

gasman:
Merged in b77848de20ee4a44fdb504d942a311ce706940e9. Have moved `get_url_name` into the base ViewSet class, as that seems like an uncontroversial enough change. No objections to the other two suggestions, but I think it's best to save those for a future PR - dealing with non-numeric PKs correctly requires escaping non-URL-safe characters (as we do in modeladmin), and changing import paths will need more discussion.

PR #3650, opened 2017-06-11
Improve consistency of ModelAdmin's view initialisation methods (attempt 3)

### LINKED ISSUE #3392: Improve consistency of how views in ModelAdmin are initialised

ababic:
Fixes #3392. Replaces #3626 

@gasman Have wrapped the view-related changes into one commit here. Changes are detailed below:

- Alter signature of `WMABaseView.__init__()` to accept and pass additional arguments to super, where `View` sets them as attributes on `self`
- Deprecate the passing of ‘instance_pk’  to `InstanceSpecificView.__init__()`
- Replace the behaviour that was previously in `__init__` and relied upon by other views, with property/cached_property methods that return the same values
- Retain `InstanceSpecificView.dispatch()`’s former behaviour of raising a 404 error when a page instance cannot be identified, before redirecting to wagtailadmin’s page edit view

I'll let the test suite finish running on these changes before committing the `ModelAdmin` method changes, so you can see that tests still pass.

### DIFF
--- a/wagtail/contrib/modeladmin/views.py
+++ b/wagtail/contrib/modeladmin/views.py
@@ -197,12 +197,14 @@ class InstanceSpecificView(WMABaseView):
         super(InstanceSpecificView, self).__init__(*args, **kwargs)
 
     @property
-    def pk_quoted(self):
-        return self.kwargs.get(self.pk_url_kwarg, self.instance_pk_init)
+    def instance_pk(self):
+        return unquote(
+            self.kwargs.get(self.pk_url_kwarg, self.instance_pk_init)
+        )
 
     @property
-    def instance_pk(self):
-        return unquote(self.pk_quoted)
+    def pk_quoted(self):
+        return quote(self.instance_pk)
 
     @cached_property
     def instance(self):
--- a/wagtail/contrib/modeladmin/views.py
+++ b/wagtail/contrib/modeladmin/views.py
@@ -233,7 +233,7 @@ class InstanceSpecificView(WMABaseView):
         repeat database queries
         """
         if hasattr(self, '_set_instance'):
-            return self._developer_instance
+            return self._set_instance
         if hasattr(self, '_fetched_instance'):
             return self._fetched_instance
         self._fetched_instance = self.get_model_instance()
@@ -243,7 +243,7 @@ class InstanceSpecificView(WMABaseView):
     def instance(self, value):
         warnings.warn(
             "Setting of 'self.instance' is deprecated. You should look at "
-            "overriding the 'get_model_instance()' method instead" %
+            "overriding the 'get_model_instance()' method on '%s' instead" %
             self.__class__.__name__, category=RemovedInWagtail113Warning
         )
         self._set_instance = value
--- a/wagtail/contrib/modeladmin/views.py
+++ b/wagtail/contrib/modeladmin/views.py
@@ -208,8 +208,8 @@ class InstanceSpecificView(WMABaseView):
         Returns an instance of self.model identified by URL parameters in the
         current request. Raises Http404 if no match is found.
         """
-        pk = self.kwargs.get(self.pk_url_kwarg, self.instance_pk)
-        queryset = self.model._default_manager.all().filter(pk=unquote(pk))
+        pk = unquote(self.pk_quoted)
+        queryset = self.model._default_manager.all().filter(pk=pk)
         try:
             # Get the single item from the filtered queryset
             obj = queryset.get()
--- a/wagtail/contrib/modeladmin/views.py
+++ b/wagtail/contrib/modeladmin/views.py
@@ -183,7 +183,7 @@ class ModelFormView(WMABaseView, FormView):
 
 class InstanceSpecificView(WMABaseView):
     pk_url_kwarg = 'instance_pk'
-    instance_pk_init = None
+    instance_pk_init = None  # to be removed in 1.13
 
     def __init__(self, *args, **kwargs):
         if 'instance_pk' in kwargs:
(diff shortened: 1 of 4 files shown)


ababic:
@gasman Have committed the ModelAdmin functions amends now too, so ready for review

wagtail/contrib/modeladmin/views.py
```diff
@@ -178,20 +182,47 @@ def form_invalid(self, form):
...
+        except self.model.DoesNotExist:
+            raise Http404(_("No %(verbose_name)s found matching the query") %
+                          {'verbose_name': self.verbose_name})
+        return obj
+
+    @property
+    def pk_quoted(self):
+        return quote(
+            self.kwargs.get(self.pk_url_kwarg, self.instance_pk)
+        )
```
gasman: The pk as retrieved from `self.kwargs[self.pk_url_kwarg]` / `self.instance_pk` is the quoted version, isn't it? (Hence why it has to be un-quoted in line 212 above.) If so, we shouldn't be calling `quote(...)` on it again here.

The start of `get_model_instance` can then be rewritten to avoid duplication and make the logic a bit clearer:

        pk = unquote(self.pk_quoted)
        queryset = self.model._default_manager.all().filter(pk=pk)

wagtail/contrib/modeladmin/views.py
```diff
@@ -697,6 +730,7 @@ def check_action_permitted(self, user):
 
     @method_decorator(login_required)
     def dispatch(self, request, *args, **kwargs):
+        self.instance
```
gasman: This could do with a comment to explain why it's here, something like: "pre-fetch the instance to ensure that any Http404 exception happens before the redirect below"

wagtail/contrib/modeladmin/views.py
ababic: Thanks. Yes, you're right.

wagtail/contrib/modeladmin/views.py
ababic: Agreed :)

ababic: I also moved that line into the if statement below, as we only need to worry about prefetching the instance at that time when `self.is_pagemodel`

ababic:
Thank you for taking the time the review @gasman. Those two changes have been made as requested.

ababic:
@gasman Sorry, last change included now (your recommended tidying of `get_model_instance`). Hoping that `pk_quoted` can be deprecated soon, as that's really something that should be handled by URLHelper (as implemented in #3655)

gasman:
Thanks @ababic!
Please could you add a test case for the quoting bug (i.e. one that fails on the unfixed version)?

Also, just to check - is the `instance_pk` deprecation something that we need to write a release note about? (If it's only something that's done by internal code, we're fine.)

ababic:
Hi @gasman. I suppose it would be sensible to include something in the release notes about these changes. I suppose there a two potential pain points:

If you are sublcassing `contrib.modeladmin.views.InstanceSpecificView`, `contrib.modeladmin.views.EditView`, `contrib.modeladmin.views.DeleteView` or `contrib.modeladmin.views.InspectView` in your project and have overridden the `__init__()` method with code that depends on `self.instance`, `self.instance_pk` or `self.pk_quoted` having been set, that code will likely break. To solve this, you'll likely have override `get()`, `post()`, or `dispatch()` instead, and include your code in there, where `self.instance` and `self.pk_quoted` will return the same values as before.

If you are subclassing `contrib.modeladmin.views.InstanceSpecificView`, `contrib.modeladmin.views.EditView`, `contrib.modeladmin.views.DeleteView` or `contrib.modeladmin.views.InspectView` in your project and are trying to set `self.instance` or `self.pk_quoted` to different value, that code will break, because those are no longer attributes.

I suppose I also need to update `dispatch()` to ensure that if `instance_pk` is provided there instead of in __init__(), `self.instance_pk` is set to an 'unquoted' version of that?

ababic:
@gasman I have pushed a change that should ensure 'self.instance_pk' will return an unquoted version of the pk, however it was passed in. 

`pk_quoted` has also been updated to return a 'quoted' version of `instance_pk`, so things are mirroring [what used to happen in __init__()](https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/modeladmin/views.py#L188)

I guess these things are now true:

If you are sublcassing `contrib.modeladmin.views.InstanceSpecificView`, `contrib.modeladmin.views.EditView`, `contrib.modeladmin.views.DeleteView` or `contrib.modeladmin.views.InspectView` in your project and have overridden the `__init__()` method with code that depends on  `self.instance_pk`, `self.instance` or `self.pk_quoted` having been set, that code will potentially break. To solve this, you'll likely have override `get()`, `post()`, or `dispatch()` instead, and include your code in there, where `self.instance_pk`, `self.instance` and `self.pk_quoted` will return the correct values.

If you are subclassing `contrib.modeladmin.views.InstanceSpecificView`, `contrib.modeladmin.views.EditView`, `contrib.modeladmin.views.DeleteView` or `contrib.modeladmin.views.InspectView` in your project and are trying to set `self.instance_pk`, `self.instance` or `self.pk_quoted` to different value, that code will break, because those are no longer attributes.

Do you have any suggestions for how we can circumvent that second issue?

I will look at writing a new test RE quoting/unquoting behaviour

ababic:
Hi @gasman:

> Please could you add a test case for the quoting bug (i.e. one that fails on the unfixed version)?

This is now done. I'm not sure what your stance is on including tests that deliberately use deprecated behaviour? but, I've also included a test that shows the a custom modeladmin view being implemented using the old way, to ensure that still works until support is dropped.

I've also done some work to make it so that `self.instance_pk`, `self.pk_quoted` and `self.instance` can still be set, by adding a bunch of `@property.setter` methods, and altering the property methods themselves to return manually set values if they have been set (plus additional tests to show that still works). It's a lot of added complexity for the sake of backwards compatibility, but I'm thinking it's probably necessary, since modeladmin is likely to be a developers preferred way for adding custom views to the admin area for their models, and subclassing the existing views is offer the best logical starting point. Plus, because of the previous lack of something like a `get_model_instance()` method dedicated to that specific task, it's probably pretty common for developers to copy the current code from `InstanceSpecificView` and adapt it for their uses, which means `instance_pk` `pk_quoted` and `instance` do really need to remain settable for now. I didn't even realise I'd done things this way, but [wagtailmenus has a prime example of this](https://github.com/rkhleics/wagtailmenus/blob/master/wagtailmenus/views.py#L51). If ever we needed a good argument for abstraction in view classes, I suppose this serves as a decent example :)

I still can't think of any possible way to cater for users needing `self.instance_pk` `self.pk_quoted` or `self.instance` to have the correct values as early on as in the views `__init__()` method though. Do you think a mention in the release notes will be enough to cover that?

ababic:
Apologies. A couple of changes I'd made locally got missed out of the previous commit, so just committed them too.

ababic:
@gasman I appreciate your help reviewing this, but I don't see myself being able to finish this any time soon. I'll close for now.

Issue #3639, opened 2017-06-07
KeyError if user open preview URL from bookmarks


m1kola:
### Issue Summary

`KeyError` if user open preview URL from bookmarks or access it directly in some other way.

### Steps to Reproduce

1. Go to the admin UI -> Explorer
2. Pick any page and press an edit button
3. Extract page's ID from an URL (for example, in URL `/admin/pages/3/edit/` id is `3`)
2. Open preview URL for this page directly (for example, `/admin/pages/3/edit/preview/`)

### Result

```
  File "django/core/handlers/exception.py", line 42, in inner
    response = get_response(request)
  File "django/core/handlers/base.py", line 249, in _legacy_get_response
    response = self._get_response(request)
  File "django/core/handlers/base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "django/views/decorators/cache.py", line 43, in _cache_controlled
    response = viewfunc(request, *args, **kw)
  File "django/views/decorators/cache.py", line 43, in _cache_controlled
    response = viewfunc(request, *args, **kw)
  File "wagtail/wagtailadmin/decorators.py", line 27, in decorated_view
    return view_func(request, *args, **kwargs)
  File "django/views/generic/base.py", line 68, in view
    return self.dispatch(request, *args, **kwargs)
  File "django/views/generic/base.py", line 88, in dispatch
    return handler(request, *args, **kwargs)
  File "wagtail/wagtailadmin/views/pages.py", line 613, in get
    form = self.get_form()
  File "wagtail/wagtailadmin/views/pages.py", line 584, in get_form
    post_data_dict, timestamp = self.request.session[self.session_key]
  File "django/contrib/sessions/backends/base.py", line 57, in __getitem__
    return self._session[key]
KeyError: 'wagtail-preview-3'
```

### Expected result

Instead of throwing an exception we can do one of the following:

* Try to re-generate preview on the fly
* Gently ask user not to access this URL directly or say something like "Preview for this page has expired"

### Technical details

* Wagtail version: Wagtail>=1.10

pkamp3:
We hit this locally with users pasting links to each other.  Not sure which result you want to go with, but we worked around it by returning an empty form:

[...]

PR #3663, opened 2017-06-16
Possible PR for #3639, KeyError on Preview Link

pkamp3:
One possible solution: If the session key does not exist when visiting a `/preview`, fall back to the "preview error page" (there will be a dialog).  The user can then try previewing the article again.

The most common reasons for the missing key:
- The user bookmarked the `/preview` link and is visiting it in a new session
- A user pastes a link to another user via email or IM


pkamp3:
Discussion for #3639 can [be found here](https://github.com/wagtail/wagtail/issues/3639).

wagtail/wagtailadmin/tests/test_pages_views.py
```diff
@@ -1550,6 +1550,27 @@ def test_preview_on_edit(self):
...
+        response = self.client.get(preview_url)
+
+        # Check the HTML response
+        self.assertEqual(response.status_code, 200)
+
+        # We should have an error page because we are unable to
+        # preview; the page key was not in the session.
+        self.assertContains(
+            response,
+            "title>Wagtail - Preview error</title>"
```
BertrandBordage: `<title>` is missing one character.
Also, add the argument `html=True` to `assertContains` when comparing HTML. See [the Django docs about it](https://docs.djangoproject.com/en/1.11/topics/testing/tools/#django.test.SimpleTestCase.assertContains).

wagtail/wagtailadmin/tests/test_pages_views.py
```diff
@@ -1550,6 +1550,27 @@ def test_preview_on_edit(self):
...
+        # get() without corresponding post(), key not set.
+        response = self.client.get(preview_url)
+
+        # Check the HTML response
+        self.assertEqual(response.status_code, 200)
+
+        # We should have an error page because we are unable to
+        # preview; the page key was not in the session.
+        self.assertContains(
+            response,
```
BertrandBordage: I’m surprised that it’s working using the `Response` object. Please use `response.content` instead.

wagtail/wagtailadmin/tests/test_pages_views.py
```diff
@@ -1550,6 +1550,27 @@ def test_preview_on_edit(self):
...
+            "title>Wagtail - Preview error</title>"
+        )
+        self.assertContains(
+            response,
+            "<h1>Preview error</h1>"
+        )
```
BertrandBordage: Same comments as before.

wagtail/wagtailadmin/tests/test_pages_views.py
gasman: This is correct - `assertContains` specifically works with Response objects, as per https://docs.djangoproject.com/en/1.11/topics/testing/tools/#django.test.SimpleTestCase.assertContains. (Perhaps you're thinking of assertIn?)

pkamp3: It looks like `assertContains` has some sugar for us behind the scenes:

https://github.com/django/django/blob/master/django/test/testcases.py#L352-L369

Would we prefer an  `assertHTMLEqual` (for both) instead?

BertrandBordage: That’s right, sorry about that ;)

PR #3671, opened 2017-06-21
Call Field.prepare_value from FieldBlocks

timheap:
This is used by some form fields to convert from the internal Python value to the value used by a widget, much like `Block.get_prep_value` for blocks or `Field.get_prep_value` for model fields.

This is a bug fix. `prepare_value` is part of the official Django form API / workflow, Wagtail is not conforming to the forms API by not calling this method.

This method is usually called from the BoundField when it is getting a value to pass to the widget. The BoundField is bypassed here because making a BoundField requires a Form instance as well, and that starts getting tricky. This could be an issue, as fields can technically customise the BoundField for a Field, but I don't know of any that customise it such that `prepare_value` would be called in a different manner.

### DIFF
--- a/wagtail/wagtailcore/blocks/field_block.py
+++ b/wagtail/wagtailcore/blocks/field_block.py
@@ -26,11 +26,12 @@ class FieldBlock(Block):
         return self.field.widget.id_for_label(prefix)
 
     def render_form(self, value, prefix='', errors=None):
-        widget = self.field.widget
+        field = self.field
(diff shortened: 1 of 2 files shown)


timheap:
Appveyor tests seem to be failing for unrelated reasons, and I can not restart the build.

gasman:
Thanks! Merged in 55a56416f54de7e569df99f8092f17cf5f521c2b.

PR #3682, opened 2017-06-28
Show / hide InlinePanel arrow buttons correctly on non-default tabs

### LINKED ISSUE #3667: Orderable extra arrows in non-content tabs
(Tested with Wagtail 1.10.1)

Currently, Wagtail has some trouble initializing JS in other tabs than the default one (see #2687 for example).

In this case, Wagtail doesn’t properly detects that an orderable item is the first or the last of the panel, so it shows both arrows instead of just the relevant one. But once we reorder any item of the same panel, the issue disappears.

orderable-extra-arrow-in-tab

gasman:
Base the show/hide logic on the 'deleted' class, rather than visibility, because elements on non-default tabs are considered invisible. Fixes #3667


BertrandBordage:
Merged as 29360a4f70691792e441de55c295aeb0707d1730.

Issue #3689, opened 2017-07-04
React explorer doesn't use get_admin_display_title


Henk-JanVanHasselaar:
### Issue Summary

The admin API for the react explorer doesn't use the `get_admin_display_title` of the`Page` object to determine the page title. 

### Steps to Reproduce

Run any wagtail project on wagtail 1.11 and change the `get_admin_display_title` of your custom Page object.

we used this in wagtail trans to display the language of the page. 
I've added some screenshots for visualization
![with-react-explorer](https://user-images.githubusercontent.com/12495915/27830653-b105b5c8-60c7-11e7-962e-08cf0e432062.png)
![without-react-explorer](https://user-images.githubusercontent.com/12495915/27830652-b104aa0c-60c7-11e7-970e-d2263167f604.png)

### Technical details

* Python version: 3.5.3
* Django version: 1.11.3
* Wagtail version: 1.11

tomdyson:
Thanks very much for this report, @Henk-JanVanHasselaar. We're putting together a fix which we'll release ASAP.

gasman:
Fixed in #3694 (for 1.11.1) and #3697 (for 1.12).

Henk-JanVanHasselaar:
Thanks guys for this quick fix! 👍

PR #3694, opened 2017-07-06
Add admin_display_title to admin API (for 1.11.1)

gasman:
First step of fixing #3689. The remaining step is to update the explorer to use the new field - I haven't tackled this because hacking on React code currently makes me feel like http://i1.kym-cdn.com/photos/images/newsfeed/000/234/765/b7e.jpg . (I could flail around changing stuff until I arrive at something that looks plausible and seems to work, and that's not what we really want in a bugfix release...)

This PR targets 1.11.1 - porting it to 1.12 will need more attention to work alongside #3285 (in particular, the change proposed in https://github.com/wagtail/wagtail/pull/3285#issuecomment-313178440 needs applying in order for `test_listing_uses_admin_display_title` to pass)

### DIFF
--- a/client/src/components/Explorer/ExplorerHeader.js
+++ b/client/src/components/Explorer/ExplorerHeader.js
@@ -20,7 +20,7 @@ const ExplorerHeader = ({ page, depth, onClick }) => {
     >
       <div className="c-explorer__header__inner">
         <Icon name={isRoot ? 'home' : 'arrow-left'} />
-        <span>{page.title || STRINGS.PAGES}</span>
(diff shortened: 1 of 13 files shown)


jjanssen:
Hi @gasman, hope you don't mind editing your PR with the changes to the Explorer. Hoped to get your consent on Slack for this. But I had some limited time in-between chores to fix this so I proceeded anyway :blush:

Anyway, the client-side code now also uses the `admin_display_title` instead of the `title`. Also gave it a quick-spin with the wagtailtrans sandbox and it shows up with the correct results.

PR #3701, opened 2017-07-10
AbstractForm.process_form_submission should return the created submission instance

cho-leukeleu:
...otherwise ```submission``` will always be None in AbstractEmailForm.process_form_submission

```
def process_form_submission(self, form):
        submission = super(AbstractEmailForm, self).process_form_submission(form)
        if self.to_address:
            self.send_mail(form)
        return submission
```


kaedroho:
Thanks!

PR #3703, opened 2017-07-11
Non-obstructive navbar toggle icon (updated)

kapito:
Changes the design of the navbar toggle icon so that it no longer obstructs page headers.

[edit: see images below]

PR #3717, opened 2017-07-18
Fix snippet page's header class

kapito:
In the snippet index page (see below), the page's header currently uses the class `header-snippet` which does not actually appear to be defined anywhere. Furthermore, it prevents #3703 from working on this page because it relies on pages using the class `header-title` for their headers (which they all do AFAIK, apart from this one).

image


thibaudcolas:
This change doesn't "fix" anything on its own, it should have probably be done as part of #3703 so I won't write release notes just for this. Moving on to #3703.

PR #3722, opened 2017-07-18
Add user hooks

jontours:
I'm a core developer for the FEC modernization team at 18F.  We needed to log admin actions and this was my solution.  I like this solution as it allows us to use decorators to extend functionality.  If this approach is too heavy-handed, I'm open to other ideas.  I implemented based on the hooks for the Editor workflow hooks.


gasman:
Thanks @jontours - looks good to me! Merged in 7859627a6e9fbc30e389ed6ea48f1ac9e194bf51.

Issue #3720, opened 2017-07-18
Avatar obstructs 'Revisions' link in footer on mobile


kapito:
When in its mobile state, the avatar completely obstructs the 'revisions' link:

![image](https://user-images.githubusercontent.com/14837124/28313244-0f53de52-6bae-11e7-98eb-1df62384dfd9.png)

PR #3723, opened 2017-07-19
Fix inconsistent footer width

kapito:
Fixes the first issue noted in #3720 -- the footer width now stays constantly centered. Also corrects some (accidental?) duplicate code.

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
@@ -435,7 +435,7 @@ footer,
     }
 
     footer {
-        width: 80%;
+        width: calc(100% - #{$menu-width} - 100px); // [full width] - [width of nav menu] - [2 * width of margin]
         margin-left: 50px;
     }
 
@@ -459,11 +459,6 @@ footer,
                 }
             }
         }
-    }
-
-    footer {
-        width: 80%;
-        margin-left: 50px;
     }
 
 
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
@@ -437,7 +437,7 @@ footer,
     footer {
         margin-left: $desktop-nice-padding;
         margin-right: $desktop-nice-padding;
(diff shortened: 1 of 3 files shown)


wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
```diff
@@ -434,7 +434,7 @@ footer,
     }
 
     footer {
-        width: 80%;
+        width: calc(100% - #{$menu-width} - 100px); // [full width] - [width of nav menu] - [2 * width of margin]
```
thibaudcolas: The comment should be replaced by variable names – specifically `2 * $margin-width` is more clear.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
```diff
@@ -434,7 +434,7 @@ footer,
...
         margin-left: 50px;
```
thibaudcolas: This should also set `margin-right`, otherwise the calculation above is wrong.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
kapito: Wouldn't that imply there is a variable called `$margin-width` (which there isn't AFAIK)?

thibaudcolas: Yep, it needs to be created so we enforce the same value here, and for `margin-right` and `margin-left` 🙂.

kapito: Ah, how about just using `$desktop-nice-padding`? That's also at 50px.

thibaudcolas: Sounds appropriate 👍 .

thibaudcolas:
Thanks for the changes, @kapito, this is now merged.

Note that I removed the comment you added – when asking for a variable to be used, my intention was to make the comment unnecessary (by giving that value an explicit name). The fact that the comment is redundant is more apparent if moved above the line:

```scss
// [full width] - [width of nav menu] - [2 * $desktop-nice-padding]
width: calc(100% - #{$menu-width} - #{2 * $desktop-nice-padding});
```

Issue #3600, opened 2017-05-18
Missing Spanish translations in Wagtail 1.10


gasman:
As reported in https://github.com/wagtail/wagtail/issues/23#issuecomment-302138180, the Spanish localisation is showing considerably fewer translated strings than previous releases - for example, the "No pages have been created at this location." message in the explorer listing no longer appears as translated.

gasman:
It seems like this is a Django bug / missing functionality, and I'm quite surprised that we're apparently the first people to encounter it...

In previous releases, wagtailcore/wagtailadmin just had an "es" translation. In 1.10, a small and incomplete "es_ES" (i.e. Spain-specific Spanish) translation was added. By Python / gettext localisation logic, `es_ES` translations take precedence over `es` when a locale of `es` or `es_ES` is active; gettext handles this by returning a translation dictionary object for the `es_ES` translations, with the `es` translations as a fallback.

`django.utils.translation` takes these dictionaries and rebuilds them in order to apply Django's own precedence rules (directories named in `LOCALE_PATHS` first, followed by locale dirs in `INSTALLED_APPS` and finally Django's own translations). This code does not look at the fallback translations returned by gettext at all, and so we end up with a translation dictionary that _only_ consists of the `es_ES` strings.

gasman:
I manage to hack support for gettext fallbacks into Django by patching the `DjangoTranslation.merge` method in `django.utils.translation.trans_real` as follows:

    def merge(self, other):
        """Merge another translation into this catalog."""
        if not getattr(other, '_catalog', None):
            return  # NullTranslations() has no _catalog
        if self._catalog is None:
            # Take plural and _info from first catalog found (generally Django's).
            self.plural = other.plural
            self._info = other._info.copy()
            self._catalog = other._catalog.copy()
        else:
            # custom
            if getattr(other, '_fallback', None):
                self.merge(other._fallback)
            # end custom
            self._catalog.update(other._catalog)

Obviously that doesn't help for running against current Django releases though - I think the short term (= "until we drop Django <2.0 support") fix may be to simply drop the `es_ES` translations from the release.

chimeno:
Sorry about that, I haven't got enough time to properly update the ``es_ES`` translations...Please proceed as you see more convenient.

SalahAdDin:
I have time to finish translations, but I don't know which the difference between our Spanish and Spain Spanish.

gasman:
[...]

PR #3724, opened 2017-07-19
Fix deprecated Chinese language code

### LINKED ISSUE #3668: Language code is not correct for Chinese
#### Please refer to the link below.

[https://code.djangoproject.com/ticket/18419](https://code.djangoproject.com/ticket/18419)

> This is because Traditional Chinese is not used only by Taiwan (tw) but also Honk Kong (hk), and Simplified Chinese is not only used in China (cn) but also Singapore (sg) and Malaysia. A new standard is to use "zh_Hant" for Traditional Chinese and "zh_Hans" for Simplified Chinese to eliminate the confusion of the content is not just for people lived in China or Taiwan.

kapito:
Fixes #3668:

> https://code.djangoproject.com/ticket/18419

> This is because Traditional Chinese is not used only by Taiwan (tw) but also Honk Kong (hk), and Simplified Chinese is not only used in China (cn) but also Singapore (sg) and Malaysia. A new standard is to use "zh_Hant" for Traditional Chinese and "zh_Hans" for Simplified Chinese to eliminate the confusion of the content is not just for people lived in China or Taiwan.

Also changed the language name from 'Chinese (China)' to 'Chinese (Simplified)' to reflect on this change. Note that this will require translations to be added for the text 'Chinese (Simplified)'.

### DIFF
--- a/wagtail/wagtailadmin/utils.py
+++ b/wagtail/wagtailadmin/utils.py
@@ -45,7 +45,7 @@ WAGTAILADMIN_PROVIDED_LANGUAGES = [
     ('pt-pt', ugettext_lazy('Portuguese')),
     ('ro', ugettext_lazy('Romanian')),
     ('ru', ugettext_lazy('Russian')),
-    ('zh-cn', ugettext_lazy('Chinese (China)')),
+    ('zh-hans', ugettext_lazy('Chinese (Simplified)')),
 ]
 
 


gasman:
Thanks @kapito! I don't think this change alone is sufficient, though, since the translation files are still kept in folders named `zh_CN`, e.g. https://github.com/wagtail/wagtail/tree/master/wagtail/wagtailadmin/locale/zh_CN/LC_MESSAGES . To fix this completely, we'll need to rename those, and check that the translations are still picked up when the user specifies either `zh_CN` or `zh_Hans` as the `LANGUAGE_CODE` in their project's settings file. (I _think_ the special-case logic within Django should deal with that, but we'll need to double-check that. Django's handling of locality-specific translations like `zh_CN` vs `zh` does seem to be a bit buggy - see #3600)

We also need to work out the best way of updating the "Chinese (China)" translation on https://www.transifex.com/torchbox/wagtail/ , so that the old folder locations don't reappear when we fetch translations again.

PR #3726, opened 2017-07-19
Fix search sorting logic in Postgres search backend

### LINKED ISSUE #3725: Unable to use `order_by_relevance=False` with Postgres Search backend
When doing a search with the Postgres Search backend, setting `order_by_relevance=False` leads to a FieldError exception. 

```
Cannot resolve keyword 'rank' into field. Choices are: body_search, content_object, content_type, content_type_id, id, object_id, typed_pk
```

Upon investigation, this is caused by a bug in `PostgresSearchQuery.search_in_index` where if `order_by_relevance` is `False`, `rank` isn't annotated which causes the query to fail. In addition, the SQL is hard-coded to order by rank.

To fix this, I monkey patched a fixed version of `search_in_index` in my models.py.

```python
from wagtail.contrib.postgres_search.backend import (
    get_db_alias, get_pk_column, get_sql, PostgresSearchQuery)

def search_in_index(self, queryset, search_query, start, stop):
    index_entries = self.get_in_index_queryset(queryset, search_query)
    order_sql = ''
    values = ['typed_pk']
    if self.order_by_relevance:
        index_entries = index_entries.rank(search_query)
        values.append('rank')
        order_sql = 'ORDER BY index_entry.rank DESC'
    index_sql, index_params = get_sql(
        index_entries.annotate_typed_pk()
        .values(*values)
    )
    model_sql, model_params = get_sql(queryset)
    model = queryset.model
    sql = """
        SELECT obj.*
        FROM (%s) AS index_entry
        INNER JOIN (%s) AS obj ON obj."%s" = index_entry.typed_pk
        %s
        OFFSET %%s LIMIT %%s;
        """ % (index_sql, model_sql, get_pk_column(model), order_sql)
    limits = (start, None if stop is None else stop - start)
    return model._default_manager.using(get_db_alias(queryset)).raw(
        sql, index_params + model_params + limits)


PostgresSearchQuery.search_in_index = search_in_index
```

### Steps to Reproduce

1. Create a new `MyPage` page
1. Configure the Postgres search backend
1. Call `MyPage.objects.live().search('some query', order_by_relevance=False)`

### Technical details

* Python version: 3.6.0
* Django version: 1.11.3
* Wagtail version: 1.11.1

I'll submit a PR in a bit

mixxorz:
Fixes #3725

### DIFF
--- a/wagtail/contrib/postgres_search/backend.py
+++ b/wagtail/contrib/postgres_search/backend.py
@@ -246,11 +246,15 @@ class PostgresSearchQuery(BaseSearchQuery):
 
     def search_in_index(self, queryset, search_query, start, stop):
         index_entries = self.get_in_index_queryset(queryset, search_query)
+        order_sql = ''
+        values = ['typed_pk']
         if self.order_by_relevance:
             index_entries = index_entries.rank(search_query)
+            values.append('rank')
+            order_sql = 'ORDER BY index_entry.rank DESC'
         index_sql, index_params = get_sql(
             index_entries.annotate_typed_pk()
-            .values('typed_pk', 'rank')
+            .values(*values)
         )
         model_sql, model_params = get_sql(queryset)
         model = queryset.model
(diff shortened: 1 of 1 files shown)


gasman:
Can we add a test for this, please?

BertrandBordage:
Tests added in 2379da96e4ea250cb09ac20dab2d60776ea75d32 & 6f5f38076a72b501296630ba63d973d8f2fbe77f.

BertrandBordage:
Rebased as 0736735ac44fe6aad293fdab53755d3303a2eff4.

PR #3727, opened 2017-07-19
Speed up sitemaps

leviadler:
Speeds up generation of sitemaps. 

Reduced queries on site of around 1000 pages from 1036 to 18.

### DIFF
--- a/wagtail/contrib/wagtailsitemaps/sitemap_generator.py
+++ b/wagtail/contrib/wagtailsitemaps/sitemap_generator.py
@@ -9,11 +9,9 @@ class Sitemap(DjangoSitemap):
         self.site = site
 
     def location(self, obj):
-        return obj.specific.url
+        return obj.full_url
 
     def lastmod(self, obj):
-        obj = obj.specific
-
         # fall back on latest_revision_created_at if last_published_at is null
         # (for backwards compatibility from before last_published_at was added)
         return (obj.last_published_at or obj.latest_revision_created_at)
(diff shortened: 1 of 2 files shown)


tomdyson:
Thanks @leviadler!

wagtail/contrib/wagtailsitemaps/sitemap_generator.py
```diff
@@ -9,7 +9,7 @@ def __init__(self, site=None):
         self.site = site
 
     def location(self, obj):
-        return obj.full_url
+        return obj.url
```
gasman: Hi @leviadler - I think this change is doing the opposite of what you want...? `obj.full_url` will always give the full URL including domain - `obj.url` will return a local URL on single-site Wagtail installations.

leviadler: @gasman correct, which is why I changed it back to `obj.url`. According to the [django documentation](https://docs.djangoproject.com/en/1.11/ref/contrib/sitemaps/#django.contrib.sitemaps.Sitemap.location) `location` should return an absolute path without the domain.
The only change being made here is dropping the `.specific` since `items` now returns specific page objects.

gasman: Thanks for the clarification! In that case, this ought to be fixed so that it returns a local URL even on multi-site installations (where `obj.url` would return a full URL)... that's unrelated to the current PR, though, so I'll deal with that separately.

gasman:
Merged in 6bd7515663aa1b767483a02dbfe025f9c39a19ba.

PR #3731, opened 2017-07-20
Update documentation to reflect specific behavior of form_classname

jeffreyckchau:
Hello,

I'm doing some work with custom StructBlock styling and javascript, and I noticed that if you specify a value for the otherwise blank `form_classname` meta attribute, it will overwrite the css classes already on the block rather than append them. In my opinion, this is a somewhat unexpected behavior, even if it is intentional. Therefore, I've added a small note to the documentation to reflect this behavior.

Please let me know if there are any issues!


gasman:
Thanks @jeffreyckchau! Merged in 988a2487ef9a2d08c0ca2d0f792a07521485cece.

Issue #1284, opened 2015-05-09
Rich editor configuration per field


salvadormrf:
There is some things that could be added to rich text:
- add support for different configurations
- needs a way to unregister some plugins globally (there is only hooks for enabling a hallojs plugin)
- make default whitelister configurable, for example disable "a" tag globally.

Example of rich text editor configuration for a field

``` python
# Simple TextArea, where we nee just a few options on editor
class SimpleDbWhitelister(WagtailDbWhitelister):
    element_rules = {
        '[document]': allow_without_attributes,
        'a': attribute_rule({'href': check_url}),
        'p': allow_without_attributes,
        'b': allow_without_attributes,
        'i': allow_without_attributes,
        'u': allow_without_attributes,
    }


class SimpleRichTexArea(WagtailRichTextArea):
    hallo_plugins = {
        'halloheadings': {
            'formatBlocks': []
        },
        'halloformat': {
            'formattings': {
                "bold": True,
                "italic": True,
            },
        },
        'hallolists': {
            "lists": {
                "ordered": False,
                "unordered": False
            }
        },
        'hallowagtaillink': {},
        'hallorequireparagraphs': {},
        'hallocleanhtml': {
            'format': False,
            'removeTags': ['span', 'div', 'table', 'strong'],
            'allowedTags': ['a', 'p', 'i', 'b', 'u'],
            'removeAttrs': ['class', 'style', 'id'],
            'allowedAttributes': [
                ["a", ['href', 'target', 'data-id', 'data-linktype']]
            ]
        }
    }

    def render_js_init(self, id_, name, value):
        return "makeRichTextEditable({0}, {1});".format(
            json.dumps(id_),
            json.dumps(self.hallo_plugins)
        )

    def value_from_datadict(self, data, files, name):
        original_value = super(SimpleRichTexArea, self).value_from_datadict(data, files, name)
        if original_value is None:
            return None
        return SimpleDbWhitelister.clean(original_value)
```

Some changes on page-editor.js

``` javascript
function makeRichTextEditable(id, plugins) {
   var instanceHalloPlugins = plugins || halloPlugins;
....
}
```

Globally disable a hallo plugin on editor

``` javascript
function unRegisterHalloPlugin(name) {
    if (!$.inArray(name, disabledHalloPlugins)) {
        disabledHalloPlugins.push(name);
    }
    if ($.inArray(name, halloPlugins)) {
        delete halloPlugins[name];
    }
}
```

mattots:
[...]

PR #3736, opened 2017-07-26
Configurable rich text features

gasman:
Ability to configure the feature set of the rich text editor on a per-field basis, by passing a `features` list:

    intro = RichTextField(features=['p', 'h2', 'b', 'i', 'link', 'document-link'])

Implements #1284, superseding (the now-closed) #2175.

This is also a step towards deprecating hallo.js in favour of Draftail (#2409). It introduces the concept of a 'registry' of rich text features, which currently only serves to map feature identifiers such as `'h2'` to the corresponding hallo.js plugin / configuration - however, the registry itself is implemented in an editor-agnostic way, so other editor packages (including wagtaildraftail and third-party ones like wagtailtinymce) are able to piggyback on top of the feature registry in the same way, and share the same overall set of feature identifiers (so `RichTextField(editor='default', features=[...])` can be changed to `RichTextField(editor='tinymce', features=[...])` and Just Work).

The registry is maintained via a new hook `register_rich_text_features` - this means that any module can potentially supply any bit of configuration, so potentially:

* Wagtail itself defines the core features `h2`, `b`, `image` and so on, along with the hallo.js definitions for them
* [wagtailtinymce](https://github.com/isotoma/wagtailtinymce) defines tinymce plugin definitions for all of those core features
* [wagtailembedder](https://github.com/springload/wagtailembedder) defines a `snippet` feature, along with hallo.js and tinymce definitions for it
* [wagtaildraftail](https://github.com/springload/wagtaildraftail/) provides Draftail configurations for all of the above...

The next step will be to have rich text features take over more parts of the rich text editing life cycle, including HTML whitelisting, and converting between the DB, front-end and editor-specific representations of the HTML. (We need this for Draftail so that we can have a single place where we say e.g. "`<blockquote>` is an allowed rich text element", rather than having to repeatedly define how to work with `<blockquote>` in hallo.js, Draft ContentState, frontend HTML and so on.) But that's something for another PR (and most likely another release, so that people with complex hallo.js customisations have chance to rewrite them in phases...)


docs/advanced_topics/customisation/page_editing_interface.rst
```diff
@@ -57,22 +57,99 @@ Wagtail provides a general-purpose WYSIWYG editor for creating rich text content
...
+ * The field might be intended for a short text snippet, such as a summary to be pulled out on index pages, where embedded images or videos would be inappropriate;
+ * When page content is defined using :ref:`StreamField <streamfield>`, elements such as headings, images and videos are usually given their own block types, alongside a rich text block type used for ordinary paragraph text; in this case, allowing headings and images to also exist within the rich text content is redundant (and liable to result in inconsistent designs).
+
+This can be achieved by passing a ``features`` keyword argument to ``RichTextField``, with a list of identifiers for the features you wish to allow:
+
+.. code-block:: python
+
+    body = RichTextField(features=['p', 'h2', 'h3', 'b', 'i', 'link'])
+
+The recognised feature identifiers are as follows (note that add-on modules may add to this list):
```
thibaudcolas: Passing identifiers only, it's not possible to configure features further than "enabled/disabled", contrary to what `WAGTAILADMIN_RICH_TEXT_EDITORS` allows. Not sure how big of a deal this is. An example use case would be wanting to limit the level of nesting of `ul` list items to 0 only for some fields.

I think the obvious trade-off is that this would make the `features` definitions less editor-agnostic, because not all editors will provide the same degree of configuration. But passing per-field options to features still seems desirable.

gasman: I'd prefer to leave this out until we see that there's a really strong real-world need for it... I'm wary of making the features system "too powerful". The issue we ran into on #2175, where we passed form widget objects as parameters and then found that they broke migrations, was a hint to me that we shouldn't really be stuffing detailed configuration options about the editor's behaviour into model fields.

My intention with features - which will probably become clearer once they start being used in the whitelisting / text conversion logic - is that they're describing the data format ("this field is allowed to contain `<h2>` headings") rather than the behaviour of the editor, and then it's the editor widget's responsibility to decide how to make that available to the user (most likely by adding an 'h2' button to the toolbar).

The example of disallowing nested `<ul>`s is an interesting one, because you can make a strong case that that _is_ a detail of the data format, not just the editor's behaviour. My worry is that developers won't grasp this distinction, and will end up misusing per-feature configuration options to control editor behaviour ("I want to add a button with icon _foo_, which opens a chooser at URL _bar_...") far more often than using them legitimately.

docs/advanced_topics/customisation/page_editing_interface.rst
```diff
@@ -57,22 +57,99 @@ Wagtail provides a general-purpose WYSIWYG editor for creating rich text content
...
+
+ * ``p`` - paragraph text (text entered into a rich text field appears in this mode by default, so omitting this from ``features`` is unlikely to be meaningful)
```
gasman: That's a fair point! I can't really think of a legitimate reason for this to be represented as a feature - I guess I was going for a design where the 'core' provided no functionality at all, and it would _only_ be possible to enable editor plugins / whitelist rules etc through features. But that principle has already been broken even in this limited implementation (hallo.js has "always-on" plugins such as the undo/redo buttons where there's no sensible reason to disable them) so that's a bit of a moot point now. Will update...

docs/advanced_topics/customisation/page_editing_interface.rst
```diff
@@ -57,22 +57,99 @@ Wagtail provides a general-purpose WYSIWYG editor for creating rich text content
...
+ * ``h1``, ``h2``, ``h3``, ``h4``, ``h5``, ``h6`` - heading elements
+ * ``b``, ``i`` - bold / italic text
```
gasman: Good question. Feature identifiers aren't necessarily tied to HTML elements - things like `document-link` aren't, and ideally we want to get people out of the mindset that rich text = HTML. In this case I stuck with the HTML names because they're familiar shorthand identifiers for concepts like paragraphs and lists, and it would have been obtuse to go with anything different...

It's hard to say whether `strong`/`em` or `b`/`i` is more correct - I'm sure I read a version of the HTML spec that expressly said that authoring tools _shouldn't_ use `strong` / `em` unless they're sure that the author intended to use the formatting for emphasis. Either way, we definitely want to standardise this across all editors so that they produce the same element.

docs/advanced_topics/customisation/page_editing_interface.rst
thibaudcolas: Makes sense. We'll get there when we get there 🙂 .

docs/advanced_topics/customisation/page_editing_interface.rst
thibaudcolas: Doing some research, looking at text editors that already have Wagtail integrations:

- TinyMCE uses `bold` and `italic` (https://www.tinymce.com/docs/configure/content-formatting/#builtinformats) and outputs `strong` and `em` (standalone demo on its site).
- Hallo.js uses `bold` and `italic` (https://github.com/bergie/hallo#plugins), and produces `b` and `i` tags (at least in the standalone demo on its site).
- Draftail (Draft.js) uses `BOLD` / `ITALIC` (https://draftjs.org/docs/advanced-topics-inline-styles.html#mapping-a-style-string-to-css) and produces `strong` and `em`
- Froala uses `bold` and `italic` (https://github.com/jaydensmith/wagtailfroala) and produces `strong` and `em` tags (standalone demo on its site)
- Medium editor uses `bold` and `italic` (https://github.com/dperetti/Django-wagtailmedium) and produces `b` and `i` (standalone demo)

I would suggest switching to `bold` and `italic` for identifiers since it seems to be the most common, it's self-explanatory that it does not match to HTML elements, and the nomenclature for inline styles would scale better if `strikethrough`, `underline`, etc are added as features (either built-in or 3rd-party/custom).

For output – seems like there isn't much consensus. I checked other editors:

- CKEditor produces `strong` and `em` (http://ckeditor.com/)
- Trix produces `strong` and `em` (https://trix-editor.org/)

which seems to indicate `strong` and `em` has the preference. I checked [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strong) on the topic, which isn't very helpful in determining the best semantics.

docs/advanced_topics/customisation/page_editing_interface.rst
gasman: Now removed as a feature in 50028c97dd36a07a04154e4f1b74823f8cf5e2f0 - the `<p>` element is now "always on".

docs/advanced_topics/customisation/page_editing_interface.rst
gasman: On closer inspection, the feature names _are_ `bold` and `italic` - the documentation was wrong. Fixed in ddf93e43a2f3160cc59379abbf9e4483e026b53e... Yep, I think in this case it does make sense to depart from the actual element names, because the _effect_ is to provide bold / italic text, even if it's up in the air which actual elements are used.

Since hallo.js uses `<b>` / `<i>`, and therefore most of the rich text data existing in databases is using those, I think it makes sense to declare `<b>` / `<i>` as the "standard" for rich text in the database, in the interests of not creating extra work for ourselves. That doesn't rule out the possibility of rewriting them to `<strong>` / `<em>` behind the scenes, either for the front-end HTML rendering or for the benefit of editor widgets which use those natively. That's something we don't need to worry about yet, though, since the current PR is only concerned with hallo.js's behaviour.

wagtail/wagtailadmin/rich_text.py
```diff
@@ -44,15 +49,17 @@ def render(self, name, value, attrs=None):
 
     def render_js_init(self, id_, name, value):
         if self.options is not None and 'plugins' in self.options:
+            # explicit 'plugins' config passed in options, so use that
             plugin_data = self.options['plugins']
-        elif self.plugins is not None:
+        elif self.use_legacy_plugin_config:
+            # RemovedInWagtail114Warning
```
kaedroho: Is there any difference between the plugins that come from the default features and the ``halloPlugins`` list?

gasman: Sorry, I really should have written something up to explain the relation between default features and `halloPlugins`, as the details are a bit convoluted...

TLDR: the plugin config obtained through the default features should always match `halloPlugins`, _on a vanilla Wagtail installation_. They will differ on existing installations that have added their own hallo.js customisations, and this setup is needed purely to accommodate them.

Currently, the official way to extend hallo.js is to edit the `halloPlugins` list. (More precisely, the official documented method is to call `registerHalloPlugin`, but that's much the same thing, and I know there are Wagtail installations in the wild where people have edited `halloPlugins` directly to _remove_ features.) We want to deprecate this and get people to use features instead - and we want to do so via our usual two-release deprecation policy.

The end goal is that we ditch the `halloPlugins` list (aka "the legacy plugin config"), and always use the config we obtain through calling `construct_plugins_list` on the feature list. However, we can't make that change yet, because then people would immediately lose the customisations they've made through the legacy plugin config. So, the interim solution I've come up with for Wagtail 1.12 - 1.13 is:

* If the rich text field has been passed an explicit `features` parameter, do everything the new way (since we can reasonably assume that a user who's actively using the new features functionality has ported all of the hallo customisations they care about into features)
* If no `features` parameter has been passed, use the legacy `halloPlugins` config; this ensures that any customisations that only exist through `halloPlugins` will remain in place.
* This means that our on-by-default features need to provide the new-style `construct_plugins_list` plumbing, and _also_ add themselves to the legacy `halloPlugins` config. In Wagtail 1.14, we will be able to drop the latter. (Third-party packages that provide enabled-by-default hallo plugins can drop this when they drop Wagtail <1.14 support.)

I don't think we can generate a deprecation warning for this, because the backwards-compatibility code path gets triggered any time the `features` param is omitted from the `RichTextField` (which is, and will continue to be, a valid thing to do). The exact deprecated behaviour we'd want to catch and warn against is "modifying `halloPlugins` without providing a corresponding feature definition", and the only way to spot that would be to somehow police what the Javascript code within `insert_editor_js` hooks does.

gasman:
Merged in ccd74024718c7bcc32e3fdd5c71037d2e0172247

PR #3738, opened 2017-07-27
Don't render a StructValue's template on calls to str(value)

### LINKED ISSUE #2874: StructValue rendering may cause stack overflow
Environment:
- django >= 1.9
- logging level = debug

Current implementation of `wagtail.wagtailcore.blocks.StructValue` contains method `__str__` trying to render the block.
https://github.com/torchbox/wagtail/blob/v1.5.3/wagtail/wagtailcore/blocks/struct_block.py#L180

However, since `Django 1.9` in case we have an unbound (unresolvable) variable in the block's template (which django template engine casts to an empty string by default), `django.template.base` module tries to write debug level logs about that.
https://github.com/django/django/blob/1.9/django/template/base.py#L919

Default python `logging.Formatter` implementation tries to write stack trace in logs which invokes StructValue rendering again. It happens with python 2.7, 3.4 and 3.5 as well. https://github.com/python/cpython/blob/2.7/Lib/logging/__init__.py#L439

Endless recursion causes stack overflow.

gasman:
This is liable to cause infinite loops on debug / logging code that attempts to log the fact that it's rendered a template with this value in the context.

Fixes #2874, https://github.com/jazzband/django-debug-toolbar/issues/950

The template is now rendered on `__html__()` instead, so `{{ some_value }}` inside templates should continue to behave as before. There's potential backwards incompatibility here for anyone relying on the old behaviour of `str()`, but I can't think of any reason why you'd want the HTML representation outside of a template.

### DIFF
--- a/wagtail/wagtailcore/blocks/struct_block.py
+++ b/wagtail/wagtailcore/blocks/struct_block.py
@@ -9,7 +9,6 @@ from django.forms.utils import ErrorList
 from django.template.loader import render_to_string
 # Must be imported from Django so we get the new implementation of with_metaclass
 from django.utils import six
-from django.utils.encoding import python_2_unicode_compatible
 from django.utils.functional import cached_property
 from django.utils.html import format_html, format_html_join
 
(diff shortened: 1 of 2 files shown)


zodiacfireworks:
This is great, I was having a lot of troubles with those infinite loops.

FlipperPA:
Fantastic! I'll remove my hacked up DJDT once this is released. The efforts, as always, are appreciated.

gasman:
Merged in 8a055addad739ff73f6e84ba553b28389122299f

PR #3740, opened 2017-08-01
Make wagtail wag tail

kapito:
A possible update to the wagtail hover animation as suggested by @MechanisM [here](https://github.com/wagtail/wagtail/pull/1991).

 ~~https://gfycat.com/impuredimcaimanlizard~~
~~https://gfycat.com/SimplisticFlimsyAnemonecrab~~
https://gfycat.com/WelloffAllBackswimmer


BertrandBordage:
Aha, so cute and memorable!
That’s a very nice addition to make new users love Wagtail. Totally love it :)

Just a tiny improvement: when the mouse moves away, the `^` eye jumps back to the initial eye position and doesn’t move relative to the window, while the body keeps moving back to the initial position. That’s a bit weird.

kapito:
@BertrandBordage Glad you like it! The issue with the eye is due to the use of display:none

kapito:
@BertrandBordage Glad you like it! I see what you mean with the eye movement, I'll see if I can fix it.

BertrandBordage:
I see, just had a look at the code. I think the easiest solution by far is to include the SVG directly inside the HTML. That way, you can style parts of a single SVG directly from CSS. The eyes will be in the same objects as the rest of SVG components, so the transform will apply everywhere.

kapito:
Good idea, I'll give that a go.

kapito:
Rotating the container rather than the body seems to do the trick. Is there an advantage to using SVG directly inside HTML?

BertrandBordage:
The upside is that you can easily style color, transforms etc, and often have a total size (HTML + SVG + CSS) smaller.
The downside is that the SVG is downloaded every time you change the page. On small SVGs like this one, this is negligible.

In short, this is more elegant and flexible for simple images.

kapito:
Updated gif: SimplisticFlimsyAnemonecrab

wagtail/wagtailadmin/static_src/wagtailadmin/images/logo-body.svg
```diff
@@ -0,0 +1,28 @@
...
+		c0,0-0.324,4.398-0.894,12.094c-0.248,1.915-0.523,4.036-0.823,6.345c-0.317,2.305-0.556,4.811-1.124,7.447
+		c-0.508,2.643-1.049,5.455-1.619,8.419c-0.745,2.928-1.529,6.006-2.348,9.217c-0.927,3.182-2.057,6.452-3.137,9.856
+		c-1.271,3.341-2.62,6.786-4.095,10.284c-0.815,1.716-1.643,3.456-2.479,5.217c-0.418,0.879-0.84,1.764-1.263,2.654
+		c-0.48,0.862-0.963,1.729-1.449,2.601c-0.974,1.739-1.958,3.498-2.953,5.273c-1.103,1.71-2.216,3.436-3.339,5.176
+		c-0.564,0.867-1.131,1.738-1.7,2.613l-1.875,2.521c-1.27,1.675-2.499,3.398-3.826,5.062c-2.766,3.236-5.482,6.58-8.576,9.627
+		c-2.936,3.195-6.25,6.054-9.479,9.007l-5.081,4.178c-1.678,1.414-3.509,2.635-5.256,3.961c-7.111,5.136-14.549,9.812-22.201,13.76
+		c-7.636,3.969-15.32,7.542-22.954,10.477c-7.619,2.971-15.078,5.579-22.249,7.723c-7.15,2.207-13.986,4.021-20.315,5.599
+		c-12.667,3.126-23.322,5.26-30.779,6.738C131.448,355.869,127.182,356.69,127.182,356.69"/>
+</g>
+</svg>
```
thibaudcolas: Could you run this (and all of the other SVGs) through an optimiser like SVGO? I don't think we have this built into the project's tooling at the moment – the easiest would be to use https://jakearchibald.github.io/svgomg/ manually.

wagtail/wagtailadmin/templates/wagtailadmin/base.html
```diff
@@ -6,7 +6,13 @@
         <div class="inner">
             <a href="{% url 'wagtailadmin_home' %}" class="logo" title="Wagtail v.{% wagtail_version %}">
                 {% block branding_logo %}
-                    <img class="wagtail-logo" src="{% static 'wagtailadmin/images/wagtail-logo.svg' %}" alt="Wagtail" width="80" />
+                    <div class="wagtail-logo-container">
+                        <img class="wagtail-logo" src="{% static 'wagtailadmin/images/wagtail-logo.svg' %}" alt="Wagtail" width="80" />
+                        <img class="wagtail-logo-body" src="{% static 'wagtailadmin/images/logo-body.svg' %}" />
+                        <img class="wagtail-logo-tail" src="{% static 'wagtailadmin/images/logo-tail.svg' %}" />
+                        <img class="wagtail-logo-eyeopen" src="{% static 'wagtailadmin/images/logo-eyeopen.svg' %}" />
+                        <img class="wagtail-logo-eyeclosed" src="{% static 'wagtailadmin/images/logo-eyeclosed.svg' %}" />
```
thibaudcolas: This and all of the images above are only presentational, so they should all be marked with alt="" so screen readers don't try to read them out loud. We can keep the current `alt` for the main logo image so it is read.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
```diff
@@ -417,19 +417,76 @@ footer,
             display: none;
         }
 
-        img {
-            width: 60px;
+        .wagtail-logo-container {
```
thibaudcolas: Could you move this out of the whole `.logo` block to reduce nesting of the styles?

wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
```diff
@@ -417,19 +417,76 @@ footer,
...
             float: none;
             margin: auto;
             display: block;
```
thibaudcolas: I don't think those three (`float`, `margin`, `display`) attributes are necessary anymore, since this element is now a `div` not an `img`.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
```diff
@@ -417,19 +417,76 @@ footer,
...
+            width: 60px;
+            height: 75px;
+            position: relative;
+            transition: all 0.25s cubic-bezier(0.28, 0.15, 0, 2.1);
```
thibaudcolas: Since this easing function is reused twice, please extract it to a variable.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
```diff
@@ -417,19 +417,76 @@ footer,
...
+
+            img {
```
thibaudcolas: We're trying to move away from bare elements in CSS selectors to make Wagtail's CSS more maintainable, as well as using BEM notation. I'd suggest switching those selectors (here and below) to use the following naming:

```html
<div class="wagtail-logo-container">
    <img class="wagtail-logo wagtail-logo__full" />
    <img class="wagtail-logo wagtail-logo__body" />
    <img class="wagtail-logo wagtail-logo__tail" />
    <img class="wagtail-logo wagtail-logo__eyeopen" />
    <img class="wagtail-logo wagtail-logo__eyeclosed" />
</div>
```

Then, each definition can be moved outside of `.wagtail-logo-container` to reduce nesting and CSS specificity.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
```diff
@@ -417,19 +417,76 @@ footer,
...
+            }
+
+            100% {
+                transform: rotate(6deg);
+            }
         }
 
-        img.wagtail-logo:hover {
-            transform: rotate(4deg);
+        .wagtail-logo-container:hover {
```
thibaudcolas: Please move this to within the `.wagtail-logo-container {` definition as `&:hover {`.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
```diff
@@ -417,19 +417,76 @@ footer,
...
+            img {
+                left: 0;
+                top: 0;
+                width: 60px;
+                position: absolute;
+                transition: all 0.25s cubic-bezier(0.28, 0.15, 0, 2.1);
+            }
+
+            img.wagtail-logo {
+                display: none; // This is the static logo (mobile only); use the animated one on desktop
```
thibaudcolas: I think all of the "hidden on mobile" behaviour needs to be moved to utility classes so it's less intertwined with the other animation code. This could be defined in https://github.com/wagtail/wagtail/blob/master/client/scss/_utilities.scss, reusing https://github.com/wagtail/wagtail/blob/master/client/scss/_tools.breakpoints.scss. Reusing the nomenclature from above, this would look like:

```html
<div class="wagtail-logo-container">
    <img class="wagtail-logo wagtail-logo__full u-hidden@medium" />
    <img class="wagtail-logo wagtail-logo__body u-hidden@small" />
    <img class="wagtail-logo wagtail-logo__tail u-hidden@small" />
    <img class="wagtail-logo wagtail-logo__eyeopen u-hidden@small" />
    <img class="wagtail-logo wagtail-logo__eyeclosed u-hidden@small" />
</div>
```

...with some combination of the following utility classes:

```scss
.u-hidden\@medium {
    @include medium {
        display: none;
    }
}

.u-hidden\@small {
    @include small {
        display: none;
    }
}

.u-inline\@medium {
    @include medium {
        display: inline;
    }
}

.u-inline\@small {
    @include small {
        display: inline;
    }
}
```

wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
```diff
@@ -417,19 +417,76 @@ footer,
...
+
+            100% {
+                transform: rotate(6deg);
+            }
         }
 
-        img.wagtail-logo:hover {
-            transform: rotate(4deg);
+        .wagtail-logo-container:hover {
+            // 'Nod' animation
```
thibaudcolas: This and a few of the other comments below don't seem necessary – maybe just rename the animations if in doubt?

wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
```diff
@@ -417,19 +417,76 @@ footer,
...
+                }
+            }
+
+            img.wagtail-logo-tail {
+                animation: wag 0.09s alternate;
+                animation-iteration-count: infinite;
+            }
+
+            // Change eye shape from . to ^
+            img.wagtail-logo-eyeopen {
```
thibaudcolas: BEM style, this would be `.wagtail-logo__eye--open`.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
```diff
@@ -417,19 +417,76 @@ footer,
...
+                display: none !important; // Override previous style
```
thibaudcolas: I don't think this needs `!important` (and same below). Hopefully the utility classes suggestion above is enough.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
```diff
@@ -488,3 +545,18 @@ footer,
...
+            img {
+                display: none;
+            }
+
+            img.wagtail-logo {
+                display: inline !important;
+            }
+        }
+    }
+}
```
thibaudcolas: This shouldn't be necessary with the utility classes.

thibaudcolas:
Ah one more thing – what's the expected behaviour on touch screens?

BertrandBordage:
@thibaudcolas When swiping on the bird?

kapito:
Many thanks for the review @thibaudcolas, I'll try to implement as many of these changes as I can today.

kapito:
Preview of the delayed animation: https://gfycat.com/AdmiredFinishedKangaroo

wagtail/wagtailadmin/static_src/wagtailadmin/images/logo-body.svg
kapito: Done, was able to reduce each file by ~50% with no perceivable drop in quality.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
kapito: The container is required to rotate the wagtail during the 'nod' animation, perhaps there is another approach I could take for this?

kapito:
Thanks again for the review @thibaudcolas, I've tried to implement as many as the changes as I could. Please let me know if I've missed anything. One thing I couldn't change was the `!important` tag on several styles; without them, the styling breaks. Perhaps I'm missing something but it seems to me that they're necessary here.

kapito:
Just discovered that once the animation is triggered, the title ('Wagtail v ...') disappears. Not sure why or how to fix it, any ideas?

kapito:
How about triggering the animation start when you stroke the wagtail?

https://gfycat.com/WelloffAllBackswimmer


I'm open to any opinions on this.

thibaudcolas:
In my opinion we shouldn't design by committee – someone who's familiar with the Wagtail brand and the CMS's design should make the call.

kapito:
@thibaudcolas Good point, I'm happy to roll back any changes if it suits better.

wagtail/wagtailadmin/templates/wagtailadmin/base.html
```diff
@@ -37,3 +48,55 @@
...
     </div>
 {% endblock %}
+
+{# Animation easter-egg: #}
+{% block extra_js %}
+    <script>
+        $(function() {
+          var sensitivity = 8; // the amount of times the user must stroke the wagtail to trigger the animation
+
+          var $logoContainer = $('.wagtail-logo-container__desktop');
```
anselmbradford: Gratuitous jquery! Do you really need to bake in a dependency here? Would this work instead -->
https://codepen.io/anon/pen/vJZXQN?editors=0010

gasman: I don't think it's a big deal - jQuery is an existing dependency of the Wagtail admin that isn't likely to be going away any time soon. (Also, jQuery collections have the handy property that the code won't break if the logo element isn't there, e.g. because the logo has [been customised](http://docs.wagtail.io/en/v1.11.1/advanced_topics/customisation/admin_templates.html#custom-branding))

gasman:
Merged in aed6791088b81c9435bc92f0d3b0d4501b5df7af - thanks @kapito!

(I've taken the liberty of giving it an obscure commit message / release note, so that people only find out about it if they're really paying attention ;-) )

wagtail/wagtailadmin/templates/wagtailadmin/base.html
anselmbradford: I see that as jQuery squashing legitimate errors that could be handled 😛  You could include `if (!logoContainer) return;` or similar on line 59 and the behavior would be clearer. If at some point you do want this on a page that doesn't have jQuery, or you want automated test coverage without jQuery, it bakes in a hassle. The only legitimate use here in my opinion is classList coverage for IE9, which could be handled another way. It doesn't make sense for things with near universal native support like dom lookups and event handling. Just a warning from the grave of someone who migrated code out of jQuery 👻

wagtail/wagtailadmin/templates/wagtailadmin/base.html
```diff
@@ -37,3 +48,55 @@
...
+
+          $logoContainer.mouseleave(function() {
+            dirChangeCount = 0;
+            disableWag();
+          })
+
+          disableWag();
+        })
+    </script>
+{% endblock %}
```
thibaudcolas: I didn't get to review this again before the merge unfortunately. I'll make it short – please *never* write JS in HTML.

kapito: I'm very sorry, this is something I admittedly overlooked. I can release a PR tomorrow evening correcting this. Just to confirm, is the standard with JS in Wagtail to embed it like this...

```
{% block extra_js %}
    <script src="{% static 'wagtailadmin/js/wagtail-logo.js' %}"></script>
{% endblock %}
```

...or is it typically done another way?

Another thing I need to work on is the use of the `!important` rule within the CSS, I'll try to fix this in the same PR.

gasman: I'll sort this now. I think it makes sense to move the code to wagtailadmin/js/core.js, since that already contains code to initialise the menu, which is active on the same occasions that the logo is. (It also means we don't need to worry about people overriding the `{% block extra_js %}` and neglecting to insert `{{ block.super }}`...)

kapito:
Thanks @gasman!

PR #3744, opened 2017-08-01
Replace deprecated .load() function

### LINKED ISSUE #3730: jQuery load method is deprecated
### Issue Summary

The jQuery `load` method is deprecated per [jQuery docs](https://api.jquery.com/category/deprecated/) and is in use [here](https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/table_block/static/table_block/js/table.js#L135).

kapito:
Fixes #3730. The `.load()` function has been deprecated for quite a while and should be replaced with `.on('load', handler)`.

https://bugs.jquery.com/ticket/11733
> The .load() method is an ambiguous signature, it can either be an ajax load or attach/fire a "load" event. CCAO cannot tell them apart since it's a dynamic decision based on arguments.


gasman:
Merged in b6e01515cb8c95290fc0a519f27df31bab9c9080 - thanks @kapito!

PR #3748, opened 2017-08-03
Fix position of options in mobile nav-menu

kapito:
Currently: https://cdn.pbrd.co/images/GDVOws0.png
image

With PR: https://cdn.pbrd.co/images/GDVPFO9.png
image

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_main-nav.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_main-nav.scss
@@ -32,11 +32,6 @@ $footer-submenu: $submenu-color;
 }
 
 .nav-main {
-    top: 53px;
-    bottom: 0;
-    overflow: auto;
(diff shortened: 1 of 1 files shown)


wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_main-nav.scss
```diff
@@ -32,7 +32,7 @@ $footer-submenu: $submenu-color;
 }
 
 .nav-main {
-    top: 43px;
+    top: 53px;
```
thibaudcolas: This does not work, because on the mobile breakpoint `.nav-main` is set to `position: static;` (the default value).

thibaudcolas:
Fixed, tested in Firefox latest, Chrome latest, iOS 10.3.

kapito:
That's strange, I'm sure it was working fine with the latest version of Firefox. Your method seems much better than mine though.

PR #3753, opened 2017-08-09
Add WAGTAILUSERS_PASSWORD_ENABLED and WAGTAILUSERS_PASSWORD_REQUIRED settings

### LINKED ISSUE #3706: Allow making password optional (or removing the field) on 'create user' form
As reported at https://stackoverflow.com/questions/44968720/making-wagtail-user-password-optional-not-required

Prior to Wagtail 1.10, the password field on the 'create user' form was optional, more by accident than design - this was fixed in #3007. However, on setups with an external authentication mechanism such as LDAP, CAS or OAuth, creating users without a password _is_ legitimate, and in those cases it's appropriate to either make the password field optional, or remove it entirely (if external authentication is being enforced). We should provide a more friendly and official way of doing that than the workaround described on the stackoverflow question above.

(Do we have any documentation for implementing custom authentication methods? If not, this would probably be a good time to add some...)

gasman:
Fixes #3706. These options restore the ability to create users with no password set on the Django side, for setups where authentication is managed externally (e.g. LDAP) - this was inadvertently dropped in Wagtail 1.10 when the form validation was tightened up (#3007). Additionally, the password fields can now be removed entirely, to enforce the use of an external auth setup.

### DIFF
--- a/docs/advanced_topics/settings.rst
+++ b/docs/advanced_topics/settings.rst
@@ -262,6 +262,18 @@ This specifies whether users are allowed to change their passwords (enabled by d
 
 This specifies whether users are allowed to reset their passwords. Defaults to the same as ``WAGTAIL_PASSWORD_MANAGEMENT_ENABLED``.
 
+.. code-block:: python
+
+  WAGTAILUSERS_PASSWORD_ENABLED = True
+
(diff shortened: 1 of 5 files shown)


gasman:
Merged in 0f5db963646a3b55044d2f4e2dc844c659295d08

PR #3292, opened 2017-01-23
Update dependencies to support node 7

thibaudcolas:
Continuation to #3232. Did not mean to close that PR and create a new one, sorry @viaregio!

PR #3755, opened 2017-08-10
Upgrade to Node 8 and npm 5

thibaudcolas:
Replaces #3292. We are overdue for an upgrade, all of the project's front-end tooling is held up in the dark ages (eg. 6 months ago) in part because it is so hard to work with shrinkwraps in npm 2.

This upgrade brings Node 8 (next LTS, has been around for a bit now, stable) and npm 5 (very fast, comes with a new lockfile format that works much better). The lockfile is the biggest of those changes for our purposes – it will make it much easier to update dependencies in PRs while preserving a lock on the whole tree of packages.

---

I tried this in `vagrant-wagtail-develop` (https://github.com/wagtail/vagrant-wagtail-develop/pull/8) and on my local OSX El Capitan, with the following commands:

```sh
nvm install
npm install
npm run lint
npm run test
npm run build
```


docs/contributing/developing.rst
```diff
@@ -15,7 +15,7 @@ If you'd prefer to set up all the components manually, read on. These instructio
 Setting up the Wagtail codebase
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Install Node.js, version 4.x, 5.x or 6.x. Instructions for installing Node.js can be found on the `Node.js download page <https://nodejs.org/download/>`_.
+Install Node.js, version 8. Instructions for installing Node.js can be found on the `Node.js download page <https://nodejs.org/download/>`_.
```
thibaudcolas: We want to use the new lockfile format, so everyone has to use Node 8 (or above when that happens, maybe).

gulpfile.js/index.js
```diff
@@ -1,2 +1,7 @@
-var requireDir = require('require-dir');
-requireDir('./tasks', { recurse: true });
+require('./tasks/build');
+require('./tasks/default');
+require('./tasks/fonts');
+require('./tasks/images');
+require('./tasks/scripts');
+require('./tasks/styles');
+require('./tasks/watch');
```
thibaudcolas: `require-dir` didn't work with Node 8, and it didn't feel actually that useful.

package.json
```diff
@@ -3,9 +3,6 @@
   "version": "1.0.0",
   "repository": "https://github.com/wagtail/wagtail",
   "private": true,
-  "engines": {
-    "node": ">= 4"
-  },
```
thibaudcolas: It turns out this property is only respected when the package is installed as a dependency – so useless for our purposes.

emilyhorsman: I believe we could get around this by including an `.npmrc` in the repo root with `engine-strict = true`. `engineStrict` was removed as an option in `package.json` but still exists in npm settings.

gasman:
All looks good to me - thanks @thibaudcolas!

Issue #3516, opened 2017-03-29
Replace Ruby's SCSS Lint with Stylelint


jjanssen:
During the Wagtail sprint in Arnhem there were some questions about how runs the `scss-lint`. Within development this actually requires an additional dependency to Ruby to install its `scss-lint` gem dependency. Within the current ecosystem it is also possible to move this linting process into the NodeJS environment, where already all our front-end tooling resides.

In todays core meeting we have discussed `stylelint` and `scss-lint` for node. Based on our own experiences we decided to pursue the `stylelint` implementation with scss functionalities, mainly because our experience with `scss-lint` for node isn't that great due the frequent problems with its underlying parser.

If we implement this we can ditch the Ruby dependency when developing on Wagtail.

thibaudcolas:
🎉  next steps: #3804.

PR #3756, opened 2017-08-10
Fix Sass linting

thibaudcolas:
The [scss-lint](https://github.com/brigade/scss-lint) gem was renamed ± two years ago to `scss_lint`, so we were (not) using a version that didn't work and failed to raise any linting errors.

Wagtail's front-end codebase is horrible to maintain, we badly need this linting in place – this PR includes some manual linting fixes, some done with Prettier, some extra configuration, and a `npm run lint:css` wrapper so people don't have to remember what linting tool(s) we are using, and we can transition to another one (#3516) some time soon.

### DIFF
--- a/.scss-lint.yml
+++ b/.scss-lint.yml
@@ -2,7 +2,9 @@
 # SpaceBeforeBrace: search "([^\ ])\{" replace "$1 {"
 # SpaceAfterPropertyColon: search "([^\&]):([^\ \n]*);" replace "$1: $2;"
 
-scss_files: 'wagtail/**/static_src/**/scss/**/*.scss'
+scss_files:
+- wagtail/**/static_src/**/scss/**/*.scss
+- client/**/*.scss
 
 linters:
     BorderZero:
--- a/.scss-lint.yml
+++ b/.scss-lint.yml
@@ -85,6 +85,9 @@ linters:
     NestingDepth:
         max_depth: 5
 
+    PseudoElement:
+        enabled: false
+
     SelectorDepth:
         enabled: false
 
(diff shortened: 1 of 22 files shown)


.scss-lint.yml
```diff
@@ -83,6 +85,9 @@ linters:
     NestingDepth:
         max_depth: 5
 
+    PseudoElement:
+        enabled: false
```
thibaudcolas: This rule enforces writing pseudo elements with the correct syntax, `.class::before`. It's not a big deal, we'll find a tool that reformats all of the files at once when we get to this.

client/src/components/Explorer/Explorer.scss
```diff
@@ -1,8 +1,10 @@
-$c-explorer-bg: #4C4E4D;
+// scss-lint:disable DeclarationOrder
```
thibaudcolas: There was a risk of changing the semantics of the CSS, which I didn't want to take.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_tooltips.scss
```diff
@@ -7,13 +7,11 @@
     font-size: 12px;
     line-height: 1.4;
     opacity: 0;
-    filter: alpha(opacity=0);
     visibility: visible;
 }
 
 .tooltip.in {
     opacity: 0.9;
-    filter: alpha(opacity=90);
```
thibaudcolas: I can't remember which version of IE this targets, but it's old.

gasman:
Thanks @thibaudcolas!

gasman:
Now [added `scss_lint` to the vagrant-wagtail-develop image](https://github.com/wagtail/vagrant-wagtail-develop/commit/d2496ddc82d5828b939ce67a648cb38e8c6fa8d5), and [added js/css linting to the steps performed by `make lint`](https://github.com/wagtail/wagtail/commit/e2b5dc384f1909d95991735a9ef034c0c4bb197e).

Issue #3761, opened 2017-08-14
Form Submission View needs reordering, or newest first default order.


Adrian-Turjak:
### Issue Summary

The current form submission page makes it quite frustrating to actually see your newest submissions. This is because it is ordered oldest first. That ordering makes no real sense as the default, or at least doesn't unless there is a way to override this.

You can filter by a time range, but that isn't really that useful because the returned order is still oldest first.

### Steps to Reproduce

1. Setup form page as per this doc: http://docs.wagtail.io/en/v1.11.1/reference/contrib/forms/index.html
2. Make submissions a few minutes apart
3. go to the form submissions page for your form

The order will be oldest first. If you keep adding submissions eventually the pagination will kick in and move the newest entries onto a second page.

### Possible Solutions

At the very least, adding support for reordering would be good, or a global config setting that overrides the existing one here: https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailforms/views.py#L71

Making newest first the default would be best, but may cause confusion for any users used to the current sorting.

Adrian-Turjak:
@gasman I'm more than happy to do the work on this based on your comments in that pull request.

I can quite likely start looking at making newest first the default while ensuring the csv export does the same thing. That said, reversing a csv isn't exactly too hard either, but we should maintain that export since chances are someone is likely exporting and and then importing that data automatically somewhere, and this could break things for them. The gui change at least won't break things, just confuse someone briefly.

No clue where to start on making the submission list sortable with the gui so some direction there would be great. 

I would love to start properly digging into the Wagtail codebase so this could be a small enough change to get me started. :)

lb-:
Hey, I second that this is an issue (just ran into this myself). I am happy to pitch in to help out, let me know if you want me to look into something here @Adrian-Turjak

lb-:
@Adrian-Turjak @gasman 

**What if the ordering setting was defined by a field on the Form Page?**

* Add a field to the `AbstractForm` Model
* Something like `order_by_recent_first` with this appearing in the Settings tab (or just not visible at all by default)
* DB Migrations for all existing Form Pages set to False
* Defaults to True for all new Pages created, but with the user's ability to override on a form by form basis
* wagtail/wagtailforms/views.py updated to check for the Boolean on the Form Page's specific instance

Downside is that this adds complexity to the UI.

[...]

PR #3762, opened 2017-08-14
Add alternative sorting for form submission data

Adrian-Turjak:
* Adds an additional optional sorting for submission data
  via a new config entry.
* Add documentation for the full data view and the new config
  entry.

Deals with #3761

### DIFF
--- a/docs/reference/contrib/forms/index.rst
+++ b/docs/reference/contrib/forms/index.rst
@@ -106,6 +106,17 @@ Displaying form submission information
         ]
 
 
+Form Submission Data View
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The full submission data view for all forms is also accessible via the admin panel. This is sorted by submission date, with the oldest submissions first. This can be changed to newest first with the addition of a line in your settings:
+
+.. code-block:: python
+
+    WAGTAIL_FORM_SUBMISSION_SORT_NEWEST_FIRST = True
+
+The submission data can additionally be filtered by a date range via the admin interface.
+
 Index
 ~~~~~
 
--- a/wagtail/wagtailforms/views.py
+++ b/wagtail/wagtailforms/views.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, unicode_literals
 import csv
 import datetime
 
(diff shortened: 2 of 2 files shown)


timorieber:
I fully agree with you that form submissions need a sorting option. I vote for implementing "newest first" as default, even though this needs a small upgrade consideration because it's backwards-incompatible.

Adrian-Turjak:
@timorieber I agree that newest first as default makes the most sense, but I bet you there is someone out there who has gotten used to oldest first, and liked it. For that person, I'd rather just add a setting so we can make newest first an option. Because right now, it's hardcoded as oldest first only.

That said, I need to add some tests because the code coverage tests are complaining!

Oh and, it's probably worth keeping the discussion in the issue instead since this pull request may not be the best option: #3761

gasman:
Thanks @Adrian-Turjak!

I'm not keen on making this a config setting, for a couple of reasons: firstly, it's overlooking the multi-user nature of Wagtail. The person who's dealing with form submissions from day to day isn't necessarily the person with the keys to the codebase who can tweak settings according to their personal preference - and it's quite unlikely that it will occur to the user to go to their developer / agency and ask "can you change this setting please?", even if it is annoying them.

Secondly, micro-managing UX design choices is a poor use of config settings. If newest-first is indeed the more natural way of viewing the list - and I think there are good arguments that it is - then we should be bold and fix that for everyone, not just people who are persistent enough to read through the settings docs thinking "how do I want this set up?". (To me, 99% of good UX design is UX that you don't have to think about - it should just be _there_.)

To cater for people who do want the listing to be oldest-first, I think it would be worth making the "Submission date" heading clickable, as we do on other listings. (To be even nicer, we could make this a persistent property on the user profile, but that's probably overkill.) Also, we should keep in mind that this view is also used for CSV output - in this case I think oldest-first _is_ more appropriate (since it essentially acts as a log file) so we should switch the behaviour accordingly.

Adrian-Turjak:
@gasman I totally agree. I'd much prefer we do this right, and I'd also much prefer to make newest first the default.

We can continue the discussion around this in the issue I opened (#3761).

PR #3763, opened 2017-08-15
Fix explorer item link behaviour

### LINKED ISSUE #3760: "Open Link in New Tab" on a right arrow in page explorer should open page list
### Issue Summary

In the new page explorer we have a right arrow control that makes the explorer show child pages of a specific page by click. But if an user attempts to see the the list of child pages in a new tab/window, he will see the current page which is confusing.

### Steps to Reproduce

1. Create a page tree (or use [`wagtaildem`](https://github.com/wagtail/wagtaildemo)) like `Home page -> Standard Index -> A deeper menu level -> A grandchild page`
2. Click to "Pages", click on the right arrow near the "Home page" and then on a similar arrow near the "Standard Index".
3. Right click on the right arrow near the "A deeper menu level" and click on "Open Link in New Tab" menu item.

### Result

Browser opens the original page in a new tab and appends `#` into the URL.

### Expected result

Browser opens a list of pages under the "A deeper menu level" page in a new tab. Result should be the same as click on the page title in a page explorer.

### Screenshots

screen shot 2017-08-12 at 13 19 41

### Technical details

* Wagtail version: 1.12 (from the `master` branch, commit e2b5dc384f1909d95991735a9ef034c0c4bb197e).
* Browser version: Google Chrome 60.0.3112.90, MacOS 10.12.5 (16F73)

emilyhorsman:
Closes #3760 

Adds a href prop to the right arrow Button in ExplorerItem. Opening this
link manually (such as Open Link in New Tab) will now take the user to
the deeper page list. Before this commit this button erroneously had no
href and thus would only take them to the current page.

Tested in:
- [x] macOS Chrome 59
- [x] Safari 10
- [x] macOS Firefox 54

### DIFF
--- a/client/src/components/Explorer/ExplorerItem.js
+++ b/client/src/components/Explorer/ExplorerItem.js
@@ -53,6 +53,7 @@ const ExplorerItem = ({ item, onClick }) => {
         <Button
           className="c-explorer__item__action"
           onClick={onClick}
+          href={`${ADMIN_URLS.PAGES}${id}/`}
         >
           {nextIcon}
         </Button>
--- a/client/src/components/Explorer/__snapshots__/ExplorerItem.test.js.snap
+++ b/client/src/components/Explorer/__snapshots__/ExplorerItem.test.js.snap
@@ -44,7 +44,7 @@ exports[`ExplorerItem children 1`] = `
   <Button
     accessibleLabel={null}
     className="c-explorer__item__action"
-    href="#"
+    href="/admin/pages/5/"
     icon=""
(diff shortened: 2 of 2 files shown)


thibaudcolas:
Thank you @emilyhorsman! Tested in multiple browsers, this is ready to ship. I think the 1.12 release is already packaged, so this will likely be 1.13. 🚀

- [x] Chrome 60 OSX 10.11
- [x] Edge15 Windows 10
- [x] IE11 Windows 10
- [x] iOS 10.3 Safari

PR #3765, opened 2017-08-16
Make InlinePanel heading independent of label

Adrian-Turjak:
InlinePanel has both the header and the add button label set via the same value. This causes some odd confusion and doesn't entirely give you the flexibility that you'd ideally want. Not to mention this does not match the naming convention of some other panels [where `heading` is used to actually mean heading](http://docs.wagtail.io/en/v1.11.1/reference/pages/panels.html?highlight=inlinepanel#multifieldpanel) while in InlinePanel it is internally renamed to heading and used as such.

This patch adds a `heading` parameter to InlinePanel that allows the heading to be independent of the label value, but allows for full backwards compatibility by having the heading value be set to label when heading is empty.

### DIFF
--- a/docs/reference/pages/panels.rst
+++ b/docs/reference/pages/panels.rst
@@ -66,7 +66,7 @@ MultiFieldPanel
 InlinePanel
 -----------
 
-.. class:: InlinePanel(relation_name, panels=None, classname='', label='', help_text='', min_num=None, max_num=None)
+.. class:: InlinePanel(relation_name, panels=None, classname='', heading='', label='', help_text='', min_num=None, max_num=None)
 
     This panel allows for the creation of a "cluster" of related objects over a join to a separate model, such as a list of related links or slides to an image carousel.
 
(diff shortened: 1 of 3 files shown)


Adrian-Turjak:
@kaedroho as a follow up to my changes in: https://github.com/wagtail/wagtail-condensedinlinepanel/pull/16

Adrian-Turjak:
I'd like to add that while this isn't exactly a huge thing, I've always been annoyed when I can't have my inlinepanel heading as "Related Pages" and have the add button be "Add Related Page". The inability to have a non-plural on the add button is mildly frustrating, but there are many cases where having the heading be a bit more descriptive would be useful, and the label just needs to be simple.

A heading of: "Featured Customers For this Service" and an add button of "Add Featured Customers For this Service" isn't great, when all you want the label to say is "Add Featured Customer".

gasman:
Thanks @Adrian-Turjak! Merged in 9610c2d3aff7fd150caf9be62c6e569e00fadf55.

Issue #3248, opened 2016-12-22
BASE_URL setting should be renamed


kaedroho:
The ``BASE_URL`` setting used by ``wagtailadmin`` for generating URLs to put in emails is a bit inconsistent with the rest of Wagtail's settings, which are all prefixed with the module name.

I think this should be renamed to ``WAGTAILADMIN_BASE_URL`` (with backwards compatibilty for the previous setting).

timheap:
May I also suggest:

1. Documenting the WAGTAILADMIN_BASE_URL setting. BASE_URL is currently not documented.
2. If WAGTAILADMIN_BASE_URL is not set, falling back on `request.site.root_url`, then falling back to building a URL from the request HOST_NAME.

ShwetKhatri2001:
Hello, I'd like to work on this issue.

lb-:
@ShwetKhatri2001 go for it - reach out if you need a hand. Remember that unit tests and docs are a must for any pull request.

sandilsranasinghe:
is this still open?

lb-:
Yep. I haven't seen a PR come in that links to this.

sandilsranasinghe:
um I have a couple of small questions; 

- there are some usages of `settings.BASE_URL` in some html templates. Should I replace those with the tag as base_url or just change them to `settings.WAGTAILADMIN_BASE_URL`
- by pull out to a generic util do you mean add a util in admin and then use that util in the base_url_setting tag?

lb-:
Hey @sandilsranasinghe here are some answers. 

>     * there are some usages of `settings.BASE_URL` in some html templates. Should I replace those with the tag as base_url or just change them to `settings.WAGTAILADMIN_BASE_URL`
- adding a tag is probably the best for in template usage. 


>     * by pull out to a generic util do you mean add a util in admin and then use that util in the base_url_setting tag?
- yes, there should be one place (util) that does all the fallback behaviour+ read settings for this as there are two fallbacks

>     * for the reference should I add documentation under the topic Email Notifications or just create a new topic for this
- good question, not sure but a new section makes sense I think. As you have done in your PR looks good

PR #3769, opened 2017-08-17
BASE_URL changed to WAGTAILADMIN_BASE_URL

iansprice:
Fix for issue #3248. 

BASE_URL setting renamed to WAGTAILADMIN_BASE_URL. 
Templates/documentation updated; backward compatible for BASE_URL.
If no setting defined, wagtail/wagtailadmin/templatetags/wagtailadmin_tags/base_url_setting does the following: 

- If available, use request.site.root_url
- If not, fallback to building scheme based on request.is_secure() and domain from request.get_host()

### DIFF
--- a/docs/advanced_topics/settings.rst
+++ b/docs/advanced_topics/settings.rst
@@ -662,6 +662,10 @@ These two files should reside in your project directory (``myproject/myproject/`
   # which welcomes users upon login to the Wagtail admin.
   WAGTAIL_SITE_NAME = 'My Project'
 
+  # This is the base url (protocal and domain) used by the Wagtail admin.
+  # Previously this (undocumented) setting was named BASE_URL
+  WAGTAILADMIN_BASE_URL = 'http://mysite.com'
+
   # Override the search results template for wagtailsearch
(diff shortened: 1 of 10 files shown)


iansprice:
Hmm, can't access the request through context in testing. Will look into this, but if anyone has a quick tip I'll take it!

gasman:
Looks like we're not passing the request object along in the place where we'd need it:

https://github.com/wagtail/wagtail/blob/e64762acbbd9bc7d62edeca88579c3ae885990ff/wagtail/wagtailadmin/views/pages.py#L245-L246

The logic for passing context variables into these emails could do with some tidying up either way (e.g. see https://github.com/wagtail/wagtail/pull/3357#pullrequestreview-31891402) - however, I think we should keep in mind the possibility that the notification email might not be triggered from a web request at all (it might come from a script, or a task queue, or some as-yet-unimagined API for editing pages remotely). For that reason, if we're going to pass the request object as part of the call to `send_notification`, it should probably be an optional parameter.

vupham04:
Hi, this pull request is already addressed through #8268 - so I feel like this should be closed.

lb-:
Thanks for your initial work on this @iansprice - will close this as it has been resolved with a newer PR and is now planned for release in 3.0

Issue #3711, opened 2017-07-14
Save button is positioned strangely in image editor on mobile


kapito:
When in its desktop state, the image editor looks like this:

![image](https://user-images.githubusercontent.com/14837124/28215768-12a49510-68a7-11e7-8419-a03b4f6d8491.png)
This looks great, however on mobile it's a bit confusing as the save/delete buttons appear in the middle of the options (not at the end as you would expect). I think it would be more intuitive if these were placed at the very bottom so the user knows they must save even after editing the focal point; at the moment it gives the impression that it is separate and doesn't require manual saving:

![image](https://user-images.githubusercontent.com/14837124/28215631-93c67c40-68a6-11e7-8c30-844e1c6fda43.png)

SalahAdDin:
Yes, you are right.

thibaudcolas:
Frankly I have the same impression on desktop, since the Save/Delete buttons are clearly part of the left-hand column.

A simple fix would be to move those buttons under the three columns on desktop – I imagine they would then be positioned at the end of the whole form on mobile:

![wagtail-buttons-mobile](https://user-images.githubusercontent.com/877585/28578392-e95f86fe-7159-11e7-931f-f0ecb2dab5a9.png)

Thoughts and design propositions welcome!

kapito:
Another potential design:
![image](https://user-images.githubusercontent.com/14837124/28625439-30e585f4-7214-11e7-86cb-5cd49d96dbb1.png)

However I'm not sure how well this would convert over to mobile.

thibaudcolas:
Fixed in #3771.

PR #3771, opened 2017-08-17
Fix position of action buttons in image editor on mobile

parliamnt101:
…e view when viewing on mobile

This is a PR to address issue #3711.

My fix is to wrap the three columns on the page in the form instead of just having the form inside the first column. Since the first column is the only place with input fields and is the only form on the page, this should have any negative side effects. This also allows me to but the submit and delete buttons in their own column below the others. This doesn't change the view on the desktop view but moves those buttons to the bottom of the page when viewed in a smaller viewport.

### DIFF
--- a/wagtail/images/templates/wagtailimages/images/edit.html
+++ b/wagtail/images/templates/wagtailimages/images/edit.html
@@ -33,9 +33,10 @@
 
     <div class="row row-flush nice-padding">
 
-        <div class="col5">
-            <form action="{% url 'wagtailimages:edit' image.id %}" method="POST" enctype="multipart/form-data" novalidate>
-                {% csrf_token %}
+        <form action="{% url 'wagtailimages:edit' image.id %}" method="POST" enctype="multipart/form-data" novalidate>
+            {% csrf_token %}
+
+            <div class="col5">
                 <ul class="fields">
                     {% for field in form %}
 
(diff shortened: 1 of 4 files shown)


kapito:
Looks good, however the position seems out of place when using a 'medium-sized' viewport (~1200px width):

image

wagtail/wagtailimages/templates/wagtailimages/images/edit.html
```diff
@@ -33,66 +33,70 @@
...
-                data-focal-point-y="{{ image.focal_point_y }}"
-                data-focal-point-width="{{ image.focal_point_width }}"
-                data-focal-point-height="{{ image.focal_point_height }}">
-
-                <img {{ rendition.attrs }} data-original-width="{{ image.width }}" data-original-height="{{ image.height }}" class="show-transparency">
-
-                <div class="current-focal-point-indicator{% if not image.has_focal_point %} hidden{% endif %}"></div>
+        <form action="{% url 'wagtailimages:edit' image.id %}" method="POST" enctype="multipart/form-data" novalidate>
+            <div class="col5">
+                    {% csrf_token %}
```
thibaudcolas: Minor – should be right after the `form` tag instead of here.

kapito:
https://pastebin.com/kF7AUX4Z -- Tested in Firefox 55, Chrome 60 and Safari mobile (iPhone 6S). All working without issues.

wagtail/images/templates/wagtailimages/images/edit.html
```diff
@@ -48,59 +49,64 @@
...
 
-                <img {{ rendition.attrs }} data-original-width="{{ image.width|unlocalize }}" data-original-height="{{ image.height|unlocalize }}" class="show-transparency">
+                    <img {{ rendition.attrs }} data-original-width="{{ image.width|unlocalize }}" data-original-height="{{ image.height|unlocalize }}" class="show-transparency">
 
-                <div class="current-focal-point-indicator{% if not image.has_focal_point %} hidden{% endif %}"></div>
+                    <div class="current-focal-point-indicator{% if not image.has_focal_point %} hidden{% endif %}"></div>
 
+                </div>
+                <br />
+                <button class="button button-secondary no remove-focal-point" type="button">{% trans "Remove focal area" %}</button>
```
thibaudcolas: `type="button"` was necessary so clicking this wouldn't insta-submit the form.

thibaudcolas:
It's in 🎉. Will ship in Wagtail 2.1.

PR #3772, opened 2017-08-17
Fixes #1420 [WIP] -- add WAGTAILDOCS_SERVE_METHOD setting to allow serving files from underlying storage

### LINKED ISSUE #1420: Refactor document serve view to be agnostic of file storage backends
For Wagtail 1.0, the document serving view was rewritten, based around the django-sendfile API, to provide various enhancements: support for django-sendfile backends, serving the correct MIME type, and support for If-Modified-Since headers. However, the django-sendfile API assumes the use of a local storage backend (i.e. one that implements [the `path` operation](https://docs.djangoproject.com/en/1.8/ref/files/storage/#django.core.files.storage.Storage.path)), and so we had to implement a fallback (#1417) to restore support for remote storage backends such as boto-s3. As with previous versions of Wagtail, this fallback code involves reading the file content through [the `open` operation](https://docs.djangoproject.com/en/1.8/ref/files/storage/#django.core.files.storage.Storage.open) and serving it through Django, which is particularly inefficient for remote storage.

I propose the following new approach for document serving:
1. Add a `WAGTAILDOCS_FILE_STORAGE` setting, which by default is the same as [`DEFAULT_FILE_STORAGE`](https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-DEFAULT_FILE_STORAGE), and use that as the storage backend for the `file` field of the Document model. This makes it possible to set a storage backend for documents independently of other media (e.g. images).
2. Introduce a new configuration setting (proposed name `WAGTAILDOCS_SERVE_METHOD`) to specify the overall top-level behaviour for document serving. This will be one of the following values (names are provisional, and not all will necessarily be implemented in the first instance):
   - 'direct': this bypasses the serve view entirely, and uses the value of `document.file.url` (if the backend implements it; if not, we fall back on the URL of the 'serve' view as usual) whenever we want to refer to the document URL, e.g. as the result of expanding a `<a linktype="document">` link in rich text. This means that we end up bypassing the `document_served` signal, and any other application logic that might exist in the serve view (download stats tracking, access restrictions) - this is most appropriate for fully static sites (django-medusa / bakery).
   - 'redirect': The serve view will redirect to `document.file.url` (assuming the backend implements it). This is probably the best option for remote storage backends; it means that the `document_served` signal will still be fired for people following the link, but it's still possible to bypass it if they know the direct URL (which presents a loophole if access restrictions are in use).
   - 'authenticated_redirect': same as 'redirect', but implements a backend-specific authentication method (such as S3's signed URL feature with an expiry timestamp) to prevent people from using the remote URLs indefinitely. To be specified...
   - 'serve_view': The document is served as a direct response (i.e. a HTTP 200 response) to the request to the serve view. At this point, we don't specify whether this will be done with the aid of sendfile, or served directly by Django. This method is most appropriate for local storage backends; for remote backends, it will perform an expensive `open` operation (i.e. the current Wagtail 1.0 behaviour). This is also the only option which doesn't require the backend to expose a `url` property; ideally, the site implementer should ensure that the file _isn't_ available through a direct URL, to prevent bypassing the serve view.
   
   (Open question: Given that the 'sensible default' for this setting will vary depending on the backend in use, is there a good way of selecting a suitable default value at startup, in the case that `WAGTAILDOCS_FILE_STORAGE`/`DEFAULT_FILE_STORAGE` is specified but `WAGTAILDOCS_SERVE_METHOD` isn't?)
   
   For all methods except 'serve_view', we're now done. For 'serve_view', we continue as follows (note that wherever possible, we're using the Django storages API rather than using direct file access, and catching `NotImplementedError` to account for varying capabilities - this way, all storage backends are following the same code path for the most part):
3. Handling If-Modified-Since: If the backend implements the `modified_time` property, check this and return a 'not modified' response if applicable. (Aside: if we added an `updated_at` field to Document, we could do this without relying on the storage backend - provided we don't care about modifications that are done directly on the filestore, bypassing the database)
4. Set a Content-Type header using mimetypes.guess_type(filename), as our sendfile logic currently does. (We could potentially sniff file content too, either for local storages only, or for all backends if we did it at upload time and stored it in the database - but that's something for another ticket...)
5. If the backend implements the `size` property, set the Content-Length header. (TODO: confirm that this is indeed optional, as per discussion on https://github.com/torchbox/wagtail/pull/1417#discussion-diff-32632999)
6. If the backend implements the `path` property, and a `SENDFILE_BACKEND` is specified, hand off to django-sendfile. (Since sendfile is now optional and only activated when this setting is present, this possibly negates the need to bundle our own sendfile in wagtail.utils. On the other hand, if django-sendfile duplicates most of the header handling logic above, we might be better off with our own stripped-down implementation...)
7. Otherwise, serve the file via `StreamingHttpResponse`.

TODO: decide how to incorporate #733 (setting content-disposition to display PDFs in the browser) into this. Special case for PDF, or make it configurable?

tobiasmcnulty:
Simplified fix for #1420.

Allows developers to enable redirecting to the underlying `file.url` property. This is not enabled by default as a matter of backwards-compatibility (not all configured storage backends may provide a `url` that is publicly accessible).

To do:
- [ ] Figure out where to and add documentation for the `WAGTAILDOCS_SERVE_METHOD` setting (beyond the release notes)


tobiasmcnulty:
Hi @gasman -- any thoughts on this?

gasman:
Thanks @tobiasmcnulty - this looks good to me! Have rebased on current master here: https://github.com/gasman/wagtail/tree/tobiasmcnulty-issue-1420

One detail from #1420 that isn't implemented here (as far as I can see) is that in the default `WAGTAILDOCS_SERVE_METHOD=None` configuration, a remote backend (providing URL but not local path) should behave as 'redirect'. I'm keen that we include this when we ship this feature, since - for me - one of the biggest motivations for this change is to fix the counter-intuitive behaviour where setting up remote storage doesn't actually offload the job of serving documents away from the Django server. (Hopefully this should be a fairly simple tweak to the logic, so I'll go ahead and implement this soon, unless you beat me to it :-) )

Re the documentation, https://github.com/wagtail/wagtail/blob/master/docs/advanced_topics/settings.rst would be the most suitable place in our current docs structure.

tobiasmcnulty:
Hi @gasman - thanks for the review! Are you suggesting always redirecting to `doc.file.url` if it's a valid value? Or whitelisting/blacklisting certain storage backends for/from redirecting?

gasman:
@tobiasmcnulty I suggest that we redirect to `doc.file.url` if that's a valid value, AND the backend does not provide a local filesystem path (i.e. `doc.file.path` raises `NotImplementedError`).

gasman:
Superseded by #5296

PR #3773, opened 2017-08-18
Fixes #3516 - replaces scss-lint with stylelint

### LINKED ISSUE #3516: Replace Ruby's SCSS Lint with Stylelint

anselmbradford:
Fixes #3516

- Replaces `scss-lint` with `stylelint`.
- Adds `lint.js` task to gulp, with `--fix` option to autofix CSS linting errors with the command `gulp lint:css --fix`


gulpfile.js/tasks/lint.js
```diff
@@ -0,0 +1,36 @@
...
+      'node_modules/stylelint/bin/stylelint.js',
+      ['client/**/*.scss', 'wagtail/**/static_src/**/scss/**/*.scss', willFixFlag],
+      { stdio: 'inherit' }
+    ).once( 'close', code => {
+      if ( code ) {
+        gulpUtil.log( 'Stylelint exited with code ' + code );
+        reject();
+      }
+      gulpUtil.log( 'Stylelint done!' );
+      resolve();
```
thibaudcolas: Can you explain why you chose to integrate stylelint with gulp? As far as I can see this isn't reading files from the  input of another vinyl stream but from a hard-coded list. Couldn't we have the same result by directly calling the stylelint CLI from `npm run lint:css`, like we do for scss_lint?

package.json
```diff
@@ -56,10 +56,12 @@
     "gulp-sourcemaps": "~1.5.2",
     "gulp-util": "~2.2.14",
     "jest": "^20.0.4",
+    "minimist": "^1.2.0",
```
thibaudcolas: I'd rather not introduce new dependencies if possible – if we do keep the gulp integration, could we pass the arguments to stylelint directly instead of adding another layer of indirection, and dependency?

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_datetimepicker.scss
```diff
@@ -298,7 +299,9 @@
 }
 
 .xdsoft_noselect::selection { background: transparent; }
+
 .xdsoft_noselect::-moz-selection { background: transparent; }
+
 .xdsoft_datetimepicker.xdsoft_inline {
```
thibaudcolas: That "newline between rulesets" reformatting looks useful. The one between nested statements or after inline comments does not.

.stylelintrc.json
```diff
@@ -0,0 +1,25 @@
...
+    "declaration-block-trailing-semicolon": "always",
+    "function-comma-space-after": "always",
+    "function-url-quotes": "always",
+    "indentation": [4, { "severity": "warning" }],
+    "length-zero-no-unit": true,
+    "max-nesting-depth": 5,
+    "no-missing-end-of-source-newline": true,
+    "no-eol-whitespace": true,
+    "rule-empty-line-before": "always",
+    "string-quotes": "single"
```
thibaudcolas: For each of those rules, could you detail what `scss_lint` rule they replace? Not in the code, but just in this PR so we know how close of a mapping there is between the old and new config. I'd also be interested to know what `stylelint` rules exist that have no equivalent in `scss_lint` and we could start leveraging, although that's beyond the scope of this one PR.

.stylelintrc.json
```diff
@@ -0,0 +1,25 @@
+{
+  "rules": {
```
thibaudcolas: Could we use YAML instead of JSON? So we get comments for free.

client/src/components/Explorer/Explorer.scss
```diff
@@ -1,5 +1,3 @@
-// scss-lint:disable DeclarationOrder
```
anselmbradford: A search for "order" on https://stylelint.io/user-guide/rules/ returned no results, but I haven't exhaustively read all the rules, so let me know if you see it!

gulpfile.js/tasks/lint.js
anselmbradford: It's so you can pass arguments to stylelint, specifically the `--fix` flag to autofix errors. Perhaps optional arguments can be passed through to the command in package.json, but I wasn't sure how that's done. Since you have an existing gulp task infrastructure I went with this. I'd recommend supporting the same flag for ESLint.

client/src/components/Explorer/Explorer.scss
```diff
@@ -64,11 +62,13 @@ $menu-footer-height: 50px;
     }
 
     // Overrides for default link hover.
+
     &:hover {
```
anselmbradford: I'm pretty sure this is coming from "newline between rulesets" as you noted later, which I added off of the Wagtail CSS style guide for "Put line breaks between rulesets."
I'll double check. As you say later, it's not so great when there's a comment.

client/src/components/Explorer/Explorer.scss
anselmbradford: Ah ha! https://github.com/hudochenkov/stylelint-order

gulpfile.js/tasks/lint.js
anselmbradford: So if the command in package.json was `stylelint '**/*.scss' 'wagtail/**/static_src/**/scss/**/*.scss'`, additional arguments can be passed in via `npm run lint:css --fix`, however, it spits out npm ERR! messages if there are linting errors. The recommended approach to squash those in https://medium.com/@bjankord/how-to-lint-scss-with-stylelint-dc87809a9878 is to add `; exit 0` to the end, which has the unfortunate side-effect that command line argument can no longer be passed through, since they'd be going to `exit` instead of `stylelint`. What do you think?

anselmbradford:
This is the following scss-lint to stylelint conversions:

| scss-lint rule       | stylelint rule         |
| ------------- |:-------------:|
| BorderZero | Satisfied through https://github.com/timothyneiljohnson/stylelint-value-border-zero/issues/2 |
| Indentation      | indentation      |
| ColorKeyword |   color-named   |
| ColorVariable | Ignored as it was set to `False` in scss-lint |
| BangFormat | declaration-bang-space-before and declaration-bang-space-after  |
| PropertySortOrder  | Ignored as it was set to `False` in scss-lint  |
| Comment | No equivalent found  |
| DeclarationOrder | Appears to be addressable via stylelint-order plugin, but not included in this PR do to more research needed on rule choices |
| DuplicateProperty | declaration-block-no-duplicate-properties |
| ElsePlacement | scss/at-else-empty-line-before |
| EmptyLineBetweenBlocks | rule-empty-line-before |
| EmptyRule | block-no-empty |
| FinalNewline | no-missing-end-of-source-newline |
| HexLength | color-hex-length |
| HexNotation | color-hex-case |
| HexValidation | color-no-invalid-hex |
| IdSelector | No equivalent found |
| ImportantRule |  Ignored as it was set to `False` in scss-lint, however, it appears to be applied to _datetimepicker.scss, which will no longer be the case |
| ImportPath | scss/at-import-no-partial-leading-underscore and scss/at-import-partial-extension-blacklist |
| LeadingZero | Ignored as it was set to `False`, however number-leading-zero would be used to enforce a style. |
| MergeableSelector  |  Ignored as it was set to `False` in scss-lint |
| NameFormat | Ignored as it didn't have a boolean flag set before, not sure if this was required! It may not have been applying before. |
| NestingDepth | max-nesting-depth  |
| PseudoElement | Ignored as it was set to `False` in scss-lint  |
| SelectorDepth | Ignored as it was set to `False` in scss-lint |
| SelectorFormat | Ignored as it was set to all formats in scss-lint |
| PlaceholderInExtend | Ignored as it was set to `False` in scss-lint |
| PropertyCount | Ignored as it was set to `False` in scss-lint |
| QualifyingElement | selector-no-qualifying-type |
| Shorthand | declaration-block-no-redundant-longhand-properties |
| SingleLinePerProperty | declaration-block-single-line-max-declarations |
| SingleLinePerSelector | selector-list-comma-newline-after |
| SpaceAfterComma | function-comma-space-after |
| SpaceAfterPropertyColon | declaration-colon-space-after, allows only one space while old rule allowed at least one. |
| SpaceAfterPropertyName | declaration-colon-space-before  |
| SpaceBeforeBrace |  block-opening-brace-space-before, does not include allow_single_line_padding option |
| SpaceBetweenParens | function-parentheses-space-inside |
| StringQuotes | string-quotes | 
| TrailingSemicolon | declaration-block-trailing-semicolon |
| TrailingZero | number-no-trailing-zeros |
| UnnecessaryMantissa | No equivalent found |
| UnnecessaryParentReference |  No equivalent found  |
| UrlFormat | Ignored as it was set to `False` in scss-lint |
| UrlQuotes | function-url-quotes |
| VariableForProperty | Ignored as it was set to `False` in scss-lint |
| VendorPrefix | Ignored as it was set to `False` in scss-lint |
| ZeroUnit | length-zero-no-unit |

client/src/components/Explorer/Explorer.scss
anselmbradford: I decided to leave this one out for now, as the configuration mapping from the existing setting is unclear. Probably best to add it in its own PR.

.stylelintrc.json
anselmbradford: Fixed!

.stylelintrc.json
anselmbradford: See comment below.

client/src/components/Explorer/Explorer.scss
anselmbradford: This should be fixed.

package.json
anselmbradford: Removed it!

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_dropdowns.scss
```diff
@@ -155,7 +155,6 @@
     }
 
     // Styles for dropdowns which are also buttons e.g page editor
-
```
anselmbradford: These are the reverse of the earlier rule that added a space after a single line comment. I think for the space to be there the comments would need to be `/* */`.

anselmbradford:
Ready for review again!

anselmbradford:
@thibaudcolas How's this look now?

thibaudcolas:
Sorry @anselmbradford, it's been a busy week for me. I'll do a proper review tomorrow.

Thanks so much for taking the time to do the mapping from old to new rules!

gulpfile.js/tasks/lint.js
thibaudcolas: I think the errors being wrapped within npm's is ok. Just updated the `make lint` script so it silences them there at least.

package.json
```diff
@@ -60,6 +60,10 @@
     "react-addons-perf": "^15.4.2",
     "react-test-renderer": "^15.5.4",
     "redux-mock-store": "^1.2.2",
+    "stylelint": "^8.0.0",
+    "stylelint-order": "^0.6.0",
+    "stylelint-scss": "^2.0.1",
+    "stylelint-value-border-zero": "^1.0.2",
```
thibaudcolas: I've removed `stylelint-order` and `stylelint-value-border-zero` for now. Would rather see this in a separate PR, potentially looking at other plugins at the same time.

thibaudcolas:
Thank you @anselmbradford! The conversion table was very helpful. I reviewed all of the rules one by one, the only thing missing was disallowing id selectors, which I added with `selector-max-id: 0`, and `UnnecessaryParentReference` , added via `scss/selector-no-redundant-nesting-selector`.

I have removed the gulp integration, the added layer of indirection and integration code aren't worth the suppression of `npm ERR` messages. And it's important that the exit status of the command is correctly bubbled up.

Finally, I took this as an opportunity to enable a few more rules:

- Forbidding `!important`s with `declaration-no-important: true` (also added a bunch of `stylelint-disable` comments.
- Using `number-leading-zero: always` to enforce the most common style in the codebase (minor).
- Reducing enforced `max-nesting-depth` to 3 nestings (so 4 levels max)
- Prevent vendor prefixing within the code.
- Enforce only using variables in media queries with `scss/media-feature-value-dollar-variable`

And a few more that did not have much impact, just formalising our current SCSS styleguide:

```yaml
comment-no-empty: true
font-family-no-duplicate-names: true
function-calc-no-unspaced-operator: true
function-linear-gradient-no-nonstandard-direction: true
media-feature-name-no-unknown: true
no-empty-source: true
no-extra-semicolons: true
```

I'll start a new issue to discuss how we could take this further and will tag you there if you're keen to help!

thibaudcolas:
Here we go, #3804.

jjanssen:
I'm a bit late to the party I initiated... thank you for your work on this @anselmbradford 👍

gasman:
Have just run `npm install` on my [vagrant-wagtail-develop](https://github.com/wagtail/vagrant-wagtail-develop) Linux VM for the first time since this was merged, and this causes the `fsevents` package to be dropped from package-lock.json. I think we had issues with this package before, as it's OS-dependent...

Naturally we'd rather not have this file introducing noise into commits as people install on one OS or another... would it help to commit the non-fsevents version, or will that just cause the opposite problem for people developing on OS X? Or is there a command-line switch we should be using on routine `npm install` runs to say "don't update package-lock.json"?

thibaudcolas:
@gasman this is because of a new behaviour of npm. It breaks my (and many others') expectations of what this lockfile is meant to do, but apparently it's the officially intended behaviour.

I just opened a separate issue for us to find the proper fix: #3806.

For your question, I just tried what you suggested and yep it causes the opposite problem on OSX. The flag is `npm install --no-shrinkwrap`, but surely there has to be a way to get the appropriate behaviour here.

PR #3774, opened 2017-08-18
Convert groups to generic views/viewsets

kaedroho:
Picking off the lower hanging fruit first.

This converts the groups section of the admin to use generic views/viewsets.

In the process I removed the ability to sort by name, because this is the only field and thought making it default to sorting by name would be more sensible.


tomdyson:
@chosak could you have a look through this, please?

wagtail/wagtailusers/views/groups.py
```diff
@@ -26,133 +22,152 @@ def get_permission_panel_classes():
...
+
+class IndexView(SearchableListMixin, generic.IndexView):
+    page_title = _("Groups")
+    add_item_label = _("Add a group")
+    search_box_placeholder = _("Search groups")
+    search_fields = ['name']
+    context_object_name = 'groups'
+    paginate_by = 20
+    page_kwarg = 'p'
+    ordering = ['name']
```
chosak: As part of reviewing this PR I noticed that because `wagtail.wagtailadmin.views.generic.IndexView` [overrides `get_queryset`](https://github.com/wagtail/wagtail/blame/master/wagtail/wagtailadmin/views/generic.py#L59), it doesn't properly support use of `ordering` in this way, unlike [`django.views.generic.list.BaseListView`](https://github.com/django/django/blob/master/django/views/generic/list.py#L42) as users might expect.

Was this omission intentional? If you remove that custom `get_queryset` (diff [here](https://gist.github.com/chosak/d0a5b682d3340ca92f1cecc46de6b916)), all mainline unit tests continue to pass, and it makes this PR easier to implement. Tagging @gasman as your name is on the blame for that code in Wagtail's `IndexView`; this change seems sensible to keep its behavior in line with Django's, but I wanted to see what you and @kaedroho thought. One concern: could this break ModelAdmin's ordering somehow?

tomdyson: Also tagging @ababic for his input on the ModelAdmin ordering question.

ababic: Hey @chosak / @tomdyson. The IndexView in ModelAdmin is still completely standalone I believe, so this change shouldn't effect it.

wagtail/wagtailusers/views/groups.py
```diff
@@ -26,133 +22,152 @@ def get_permission_panel_classes():
...
         ]
+
+    def get_context_data(self, **kwargs):
+        if 'permission_panels' not in kwargs:
+            kwargs['permission_panels'] = self.get_permission_panel_forms()
+
+        return super(PermissionPanelFormsMixin, self).get_context_data(**kwargs)
+
+
+class SearchableListMixin(object):
```
ababic: Would `wagtail.wagtailadmin.views.generic` be a better home for `SearchableListMixin`?

wagtail/wagtailusers/templates/wagtailusers/groups/list.html
```diff
@@ -4,16 +4,11 @@
...
-                    <a href="{% url 'wagtailusers_groups:index' %}" class="icon icon-arrow-down-after teal"></a>
-                {% else %}
-                    <a href="{% url 'wagtailusers_groups:index' %}?ordering=name" class="icon icon-arrow-down-after"></a>
-                {% endif %}
             </th>
         </tr>
     </thead>
     <tbody>
-        {% for group in groups %}
+        {% for group in page_obj %}
```
chosak: Isn't this change unnecessary because of [this line](https://github.com/wagtail/wagtail/pull/3774/files#diff-6136c06de15604ea4f50c092b0aa4718R92) defining the context object name as `groups`?

wagtail/wagtailusers/views/groups.py
gasman: Hi @chosak,
Until very recently (#3625) the `wagtail.wagtailadmin.views.generic` views were completely independent of Django's generic views - so if it turns out that we're now overriding Django's `get_queryset` with a less functional implementation, then that's almost certainly just an accident of the order things were written in, and we can indeed remove our version as redundant code. (It would definitely be good to check the two methods side by side to confirm that ours isn't doing anything subtly different, though, rather than trusting the unit tests to catch that :-) )

kaedroho: I've removed these ``get_queryset`` methods in https://github.com/wagtail/wagtail/pull/3774/commits/251e41c1180a903888c06b58728f8721be6abe30. (They were also added to ``EditView`` and ``DeleteView`` but weren't being used by either of them)

wagtail/wagtailusers/views/groups.py
kaedroho: I was reluctant to do this because it uses ``__icontains`` and this isn't the best method of searching for all models. I think a generic implementation should use ``wagtailsearch`` if it's configured on the model being searched so left this implementation here for now.

wagtail/wagtailusers/views/groups.py
```diff
@@ -26,133 +22,152 @@ def get_permission_panel_classes():
...
+    page_title = _("Add group")
+    success_message = _("Group '{0}' created.")
+    template_name = 'wagtailusers/groups/create.html'
+
+    def post(self, request, *args, **kwargs):
+        """
+        Handle POST requests: instantiate a form instance with the passed
+        POST variables and then check if it's valid.
+        """
+        self.object = Group()
```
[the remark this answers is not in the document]
kaedroho: I had a look in to this but found it wasn't possible to set this to ``None`` here as the ``get_permission_panel_forms`` method needs an object to attach the forms to. I've added a comment in the code and also made that suggested cleanup in the generic views.

Issue #1049, opened 2015-03-10
Not possible to add link with fragment identifier


sirex:
Currently it is not possible to add [fragment identifier](http://en.wikipedia.org/wiki/Fragment_identifier) to Wagtail rich text links:

https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailcore/rich_text.py#L104

davecranwell:
Please could you provide a little more information about what you're expecting here? Do you mean when adding Internal links? External ones? 

Have you also tried this in the latest version of wagtail (1.0rc2) where fragment identifiers certainly are supported in _external_ links?

sirex:
@davecranwell the issue is with `a` tag with `linktype="page"`, this tag is interpreted by Wagtail for rich text fields:

```html
<a linktype="page" id="42">hello world</a>
```

How can I add `#some-ref` here?

davecranwell:
Ah ok, internal links. Yes at present wagtail doesn't support creating internal links which include a fragment identifier, whether via the rich text editor or the PageChooserPanel widget.

sirex:
I guess it is not an easy task to add internal links support to the rich text editor and the PageChooserPanel widget.

But maybe adding support just for `a` tag parser would be a good start? For example this:

```html
<a linktype="page" id="42" ref="some-ref">hello world</a>
```

Could be converted to:

```html
<a href="/some/page/#some-ref">hello world</a>
```

Of course editing such links with the rich text editor would remove fragment identifierds, but in may case, that would be acceptable.

In my use case, texts from old website are automatically imported to Wagtail, but currently I can't preserve internal page links because of this limitation. So I have bunch of links that just point to the same page and not to the parts of same page.

gasman:
Adding a new attribute for this in the DB representation seems like a good idea (although I'd probably call it `fragment` rather than `ref`). It would just need a few additions to `PageLinkHandler` in wagtailcore/rich_text.py - and if we get it to output a `data-fragment` attribute in the `for_editor` case, we'll be able to preserve the identifier in the round trip from the database to the editor and back again.

The one thing that's missing is the ability to insert such a link within the rich text editor. I'm undecided about the best way to handle this (we could just add a new text field in the link chooser UI, to enter a fragment ID - but that still relies on the template developer to insert `<a id="...">` elements in the appropriate places, and requires the editor to understand some low-level HTML details) - but as you point out, it's not necessary to solve that problem yet.

timheap:
[...]

Issue #490, opened 2014-07-21
Support for tel and other link types in external link form


nathan3d:
It would be nice if the UI for adding links allowed for non http/ftp links to be added. Currently you get a validation error when trying to add tel/maps/etc...

gasman:
Closing as per https://github.com/torchbox/wagtail/issues/490#issuecomment-53415721 - I don't think there's a legitimate use for these link types within rich text. Special-purpose pages that use them, such as contact pages, should build them into their data model / template instead.

PR #3776, opened 2017-08-21
Add option to insert tel links in rich text editor

SableWalnut:
As there is presently an option to insert email links, I figured it would be appropriate to add the option to insert tel links in addition to mailto links.

### DIFF
--- a/docs/editor_manual/new_pages/inserting_links.rst
+++ b/docs/editor_manual/new_pages/inserting_links.rst
@@ -13,6 +13,6 @@ Whichever way you insert a link, you will be presented with the form displayed b
   * Internal link: A link to an existing page within your website.
   * External link: A link to a page on another website.
   * Email link: A link that will open the user's default email client with the email address prepopulated.
+  * Phone link: A link that will open the user's default client for initiating audio calls, with the phone number prepopulated.
 
 * You can also navigate through the website to find an internal link via the explorer.
-
--- a/wagtail/wagtailadmin/forms.py
+++ b/wagtail/wagtailadmin/forms.py
@@ -68,6 +68,11 @@ class EmailLinkChooserForm(forms.Form):
     link_text = forms.CharField(required=False)
 
 
+class PhoneLinkChooserForm(forms.Form):
+    phone_number = forms.CharField(required=True)
+    link_text = forms.CharField(required=False)
+
+
 class LoginForm(AuthenticationForm):
     username = forms.CharField(
         max_length=254,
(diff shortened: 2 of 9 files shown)


gasman:
Hi @SableWalnut,
As mentioned at https://github.com/wagtail/wagtail/issues/490#issuecomment-53415721, I'm sceptical about whether there's a valid use for tel: links within rich text - the common uses I can think of would be better handled as a 'Contact' object with a 'telephone' field.

Now that we've implemented configurable rich text features (#3736), perhaps we could make this an optional (off-by-default) feature?

chosak:
After digging into the code and bit and coming up to speed on Draftail, I think I'm coming around to @gasman's preference to implement this as a rich text feature.

If the approach in this PR is used for now, it'd be nice if there were some additional validation for the phone number part (a la [django-phonenumber-field](https://github.com/stefanfoulis/django-phonenumber-field)).

thibaudcolas:
This is one of the features we've made Draftail for, along with "link with fragment identifier" #1049.

At the moment, Draftail is 100% reliant on Wagtail's chooser widget for links. If the idea is to add a new tab within the link chooser, it would be very straightforward to integrate with Draftail for Wagtail 2.0.

To build this with [WagtailDraftail](https://github.com/springload/wagtaildraftail), I made an entirely separate "rich text feature" that had its own toolbar button and chooser UI. It was very rushed (no validation, very basic input form), but may be a useful start if we choose to go this way.

mikaelengstrom:
@gasman, Does the removal of this feature from the roadmap mean that it will not be implemented?

benebun:
Any chance that will be on the roadmap?
We're also running into this. Using phone numbers in richtext is a common use case for our clients.
Two examples:
1. numbers that should be on the same line as text: "Please call us under +12 2345" if you don't want a separate paragraph for the number.
2. In Germany and Austria a non standard number format is very common (*sigh*): +12 (0)1 234 where 0 is only required if you call from within the country without the country code +12. On mobiles this number is recognised as a phone number but can't be reached because the 0 will be included: +12 01 234, correct would be +12 1 234. This could be fixed via a corrected tel: url.
Of course it's arguable that for use case 2 there could be other solutions (JS, mobile vendors…)...

gasman:
Completed in #5478

vsalvino:
Thank you @gasman and @mikaelengstrom for implementing this feature! Is it safe to assume this will be in the next version of wagtail (2.7)?

PR #3779, opened 2017-08-22
Minor documentation example fix; bold and italic should be spelled out explicitly.

FlipperPA:


thibaudcolas:
Thank you @FlipperPA 👍

Issue #2776, opened 2016-06-23
Support for AWS ElasticSearch


TuxOtaku:
Hey guys, 

So correct me if I'm wrong, but I don't think you guys have any support for AWS' branded ElasticSearch. Is there any plans to support this? Have there been any pull requests that address this? Where should I look if I wanted to add this feature myself?

kaedroho:
Hey @TuxOtaku 

I'm not aware of any differences with AWS' Elasticsearch that would mean Wagtail wouldn't work with it. Their [FAQ says](https://aws.amazon.com/elasticsearch-service/faqs/) that they use version 1.5.2, which is supported by Wagtail.

Have you come across any issues that meant Wagtail wouldn't work with AWS Elasticsearch?

TuxOtaku:
Yeah I get an error from Elasticsearch saying 
`{"Message":"User: anonymous is not authorized to perform: es:ESHttpGet on resource: boxingoctopus"}`

kaedroho:
Thanks!

Just Googled that error message and came across this: http://docs.aws.amazon.com/elasticsearch-service/latest/developerguide/es-managedomains.html#es-managedomains-signing-service-requests

Looks like AWS needs Elasticsearch requests to be signed. It seems like the best way to do this is documented in the ``elasticsearch-py`` package https://elasticsearch-py.readthedocs.io/en/master/#running-on-aws-with-iam

I think this could be done entirely in settings as extra arguments passed into the ``WAGTAILSEARCH_BACKENDS`` entry are passed straight to the ``Elasticsearch`` constructor. This might work:

```python
# settings.py

from elasticsearch import Elasticsearch, RequestsHttpConnection
from requests_aws4auth import AWS4Auth

WAGTAILSEARCH_BACKENDS = {
    'default': {
        'BACKEND': 'wagtail.wagtailsearch.backends.elasticsearch',
        'URLS': ['YOURHOST.us-east-1.es.amazonaws.com:443'],
        'http_auth': AWS4Auth(YOUR_ACCESS_KEY, YOUR_SECRET_KEY, REGION, 'es'),
        'use_ssl': True,
        'verify_certs': True,
        'connection_class': RequestsHttpConnection,
    }
}
```

TuxOtaku:
@kaedroho Sorry man...no joy:

`  File "/opt/apps/boxingoctopus/boxingoctopus/settings/base.py", line 121, in <module>
[...]

PR #3786, opened 2017-08-23
Update docs for Amazon ElasticSearch service as backend.

BenSturmfels:
This change moves the "connection_class" key into "OPTIONS", resolving error
``AttributeError: 'AWS4Auth' object has no attribute 'encode'``. Using
``elasticsearch2`` backend for consistency with rest of page.

Related to #2776.


gasman:
Thanks @BenSturmfels! Looks like this code snippet didn't get updated for Wagtail 1.8: http://docs.wagtail.io/en/v1.12/releases/1.8.html#extra-options-for-the-elasticsearch-constructor-should-be-now-defined-with-the-new-key-options-of-the-wagtailsearch-backends-setting

PR #3790, opened 2017-08-24
Fixed wrong argument for datetimechooser widget

lucasmoeskops:
The argument for specifying the language of the DateTimeChooser is wrong, which makes it impossible to customise the language. Compare for reference the argument specification for the DateChooser and the TimeChooser :-)


gasman:
Thanks @lucasmoeskops! Merged in f7024a4193d5d692d32ab51c1185ac4533d1487f.

PR #3792, opened 2017-08-24
Make user search in admin search multiple fields

wgiddens:
When searching for users in the admin, queries across multiple fields wouldn't work.  For example, if you had a user with first name John and last name Smith "John" would match, "Smith" would match, but "John Smith" returns no results.


gasman:
Looks good to me! Could you add a test for this, please?

wgiddens:
Tests are up.  Let me know if you I need to change anything.  Thanks for all your hard work!

m1kola:
Merged in 2a8e63c018bd111bc04f3a179a9dd013bdcef682 and release notes in bcf3a8f364bb8ecc7d3cbf9d7a45e8b262b5a46a. Thanks @wgiddens!

Issue #3793, opened 2017-08-25
Admin search not using ``get_admin_display_title()``


BenSturmfels:
### Issue Summary

The ``get_admin_display_title`` method of Page subclasses isn't used when searching in admin; both in page chooser search and in general admin search.

### Steps to Reproduce

For example, I create a subclass of ``Page`` with a ``get_admin_display_title`` method:

    class Artwork(Page):
        accession_number = models.CharField(max_length=50)

        def get_admin_display_title(self):
            return '{}: {}'.format(self.accession_number, self.title)

Then create an Artwork with accession_number "2009.12" and title "Red Flowers". I would expect to see the title always displayed in the Admin as "2009.12: Red Flowers" but it's shown as "Red Flowers" in:

 - page chooser when searching (eg. ``/admin/choose-page/search/``)
 - admin search (eg. ``/admin/pages/search/``)

The title is show as expected in:

 - JavaScript Page menu (eg. ``/admin/api/v2beta/pages/``),
 - page chooser (eg. ``/admin/choose-page/4/``)
 - regular page listing (eg. ``/admin/pages/9/``)

### Technical details

* Python version: 3.4
* Django version: 1.11.4
* Wagtail version: 1,12
* Browser version: Firefox 52.1.2

gasman:
Fixed in #3794

PR #3794, opened 2017-08-25
Use Page subclass get_admin_display_title() in admin search views.

BenSturmfels:
The template changes force the most specific ``get_admin_display_title`` method
to be used across all page explorer lists and page choosers in the admin. The
view changes should reduce the performance hit in the case of the search views.

Relates to #3793

### DIFF
--- a/wagtail/wagtailadmin/templates/wagtailadmin/pages/listing/_page_title_choose.html
+++ b/wagtail/wagtailadmin/templates/wagtailadmin/pages/listing/_page_title_choose.html
@@ -8,9 +8,9 @@ Expects a variable 'page', the page instance.
 
 <h2>
     {% if page.can_choose %}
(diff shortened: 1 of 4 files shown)


BenSturmfels:
I'm happy to look at adding test cases to ``test_page_chooser.py`` and ``test_pages_views.py`` if you think this approach makes sense.

gasman:
Thanks @BenSturmfels! This approach looks good to me (although it shouldn't be necessary to call `page.specific` within the template, if you've already called `specific()` on the queryset in the view) - some tests would be very welcome.

BenSturmfels:
Thanks @gasman, I'll look into tests next week.

My thinking with the template was that this would force all listings to be specific and possibly prevent/fix problems we might not have even seen yet. But you're right of course, not needed directly in this case.

gasman:
Now merged in 4063573f9c48e035a05321e98a2c89f2c310ceaa, with tests added in 09591fcf987fe189fb3e617207b37cd466e2f047

PR #3797, opened 2017-08-25
Fix form submission page header class

kapito:
Without this the mobile nav-menu button obstructs the header text.

image


gasman:
Thanks @kapito - looks good to me! Merged in 8ff650a5b4dc5f5b9214f794434a70f2ee87ff98.

Issue #3517, opened 2017-03-30
Change wagtailadmin/css/core.css to allow flexible logo height in admin


mainderay:
# Feature Request

### Issue Summary

Using a custom logo in the admin page that is too tall results in the search box being pushed down into the main nav menu.

### Steps to Reproduce

Add a logo that is say 200px in height.

### Proposed Solution

Consider changing the below section in wagtailadmin/css/ core.css from:

```
 .nav-main {
    position: absolute;
    top: 175px;
    margin-bottom: 127px;
  }
```

to

```
 .nav-main {
    position: relative;
    top: 12px;
    margin-bottom: 127px;
  }
```

thibaudcolas:
Hey @mainderay, sorry we're only getting to this now.

What you're proposing looks very sensible. I'm not sure why this has been positioned like this in the past, we may want to check multiple variations of viewport sizes to make sure the change caters to all scenarios.

I think it should also be documented somewhere what custom logo sizes will work, if that's not the case already. I don't know this feature very well.

Would you be keen to submit a pull request with those changes? Doesn't need to be perfect, but a first iteration would help a lot in getting this done.

thibaudcolas:
If anyone is interested in getting this fixed, #3798 needs further work.

thibaudcolas:
I believe this has been fixed via #4747.

PR #3798, opened 2017-08-25
Allow flexible logo height in admin

kapito:
As suggested by @mainderay in #3517.

Example (imagine this is a custom logo):

image

Before you would see this:

image

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/scss/components/_main-nav.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/components/_main-nav.scss
@@ -252,6 +252,10 @@ body.explorer-open {
 
     .nav-main {
         display: none;
+
+        @include medium {
+            display: block;
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_main-nav.scss
```diff
@@ -268,8 +268,8 @@ body.explorer-open {
     }
 
     .nav-main {
-        position: absolute;
-        top: 175px; // WARNING - magic number - the height of the logo plus search box
+        position: relative;
+        top: 0;
         bottom: 0;
         margin-bottom: 127px; // WARNING - magic number - the height of the .footer
         width: 100%;
```
thibaudcolas: Setting positions to 0 with `position: relative;` does nothing, so I have a feeling all of this code is useless unless it overrides mobile-specific rules.

Meteor0id:
Good to see some effort is being put in solving this already, but this ticket is a duplicate of #3517.
Might be best to merge tickets or move over to the original ticket with this.

thibaudcolas:
Superseded by #4747

PR #3799, opened 2017-08-25
Fix project template

m1kola:
Related to #3285

### DIFF
--- a/wagtail/project_template/home/migrations/0001_initial.py
+++ b/wagtail/project_template/home/migrations/0001_initial.py
@@ -7,7 +7,7 @@ from django.db import migrations, models
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('wagtailcore', '0029_unicode_slugfield_dj19'),
+        ('wagtailcore', '0040_page_draft_title'),
     ]
 
     operations = [
--- a/wagtail/project_template/home/migrations/0002_create_homepage.py
+++ b/wagtail/project_template/home/migrations/0002_create_homepage.py
@@ -22,6 +22,7 @@ def create_homepage(apps, schema_editor):
     # Create a new homepage
     homepage = HomePage.objects.create(
         title="Home",
+        draft_title="Home",
         slug='home',
         content_type=homepage_content_type,
         path='00010001',


gasman:
To fill in the detail for anyone searching the issues list: This addresses a problem that was triggered by the introduction of the `draft_title` field on the Page model, in #3285. Migration `wagtailcore.0040_page_draft_title` populates this field with the (live) `title`, and there's logic in the `Page.full_clean` method to fill this in for any subsequent pages that are created without setting `draft_title` explicitly.

However, if the project template's `home.0002_create_homepage` migration (or any other migration that creates page records*) runs after `wagtailcore.0040_page_draft_title`, the `draft_title` field for that page will not get filled in - this is because migrations use a 'fake' copy of the `Page` model consisting of the fields and basic ORM methods, but not the custom methods such as `full_clean`.

Updating the migration in the project template will fix the issue for new projects; however, this won't account for people bringing up new installations of old projects that were built on Wagtail <1.12 and upgraded. I think we'll need to add a release note telling people how to update their migrations manually... but then that creates new problems when a second developer (with a working <1.12 copy of the project) tries to upgrade and gets an inconsistent migration history (since `home.0002_create_homepage`, which they've already applied, now depends on `wagtailcore.0040_page_draft_title`, which they haven't applied yet).

Other possibilities, which aren't ideal but might be necessary:

* we add a system check on startup, that automatically populates any blank `draft_title` fields. This check will probably need to be left in the Wagtail codebase indefinitely (because we can't tell how long people will have a <1.12 project sitting around), and there's a remote possibility that it might take a non-negligible time to run on huge sites...
* we admit the possibility that `draft_title` might still be blank despite our best efforts, and always guard it with something like `page.draft_title or page.title` when accessing it in our code. Obviously, having code to handle an obscure edge case like this isn't great for maintainability :-/

*It's possible that users have created their own migrations that create page records, which will also be affected by this issue. However, any such migrations would already have to deal with the fact that treebeard logic is unavailable in the fake-ORM version of the Page model - and any developer who's confident enough to do that sort of undocumented low-level tinkering certainly shouldn't need any hand-holding to fix up `draft_title` :-) So I think it's enough to document this possibility in the release notes, and let them figure it out from there.

gasman:
> we add a system check on startup, that automatically populates any blank draft_title fields. This check will probably need to be left in the Wagtail codebase indefinitely (because we can't tell how long people will have a <1.12 project sitting around), and there's a remote possibility that it might take a non-negligible time to run on huge sites...

Just tried `Page.objects.filter(draft_title='').update(draft_title=F('title'))` on a test database with 250K pages, and it completed instantly, so I think this is fine... I suggest we add this system check, alongside the change to the project template as proposed here. This way, there's no need for a release note to tell people to update their migrations manually.

gasman:
> we admit the possibility that draft_title might still be blank despite our best efforts, and always guard it with something like page.draft_title or page.title when accessing it in our code. Obviously, having code to handle an obscure edge case like this isn't great for maintainability :-/

@kaedroho pointed out that the vast majority of accesses to the draft title field will happen through `get_admin_display_title` - and in fact I can't see a single exception to that in the codebase - so I think we're fine to take this approach, and have `get_admin_display_title` fall back on live titles where necessary.

gasman:
Merged in c87b400c7a450aafa76f1fc8e9936401be14f736 (1.13) and e902d3bdf354e6ebb5cef0e35ce869f82ce29b0c (1.12.1)

PR #3800, opened 2017-08-25
Rebase #3541 (signal handler fixes for custom image/doc models)

gasman:


wagtail/contrib/postgres_search/utils.py
```diff
@@ -68,15 +66,18 @@ def get_descendants_content_types_pks(models, db_alias):
...
-                              model=model._meta.model_name)
-                            for model in models]))
-                .values_list('pk', flat=True))
+    ct_manager = ContentType.objects
+    original_db_alias = ct_manager.db
+    try:
+        ct_manager._db = db_alias
+        content_types_dict = ContentType.objects.get_for_models(*models)
+    finally:
+        ct_manager._db = original_db_alias
```
BertrandBordage: Actually, after further thinking I was not really happy with this solution as it was not thread-safe: the manager object stays the same across threads, so there’s a extremely tiny chance that modifying `.db` can affect another thread.

After a small search, I found an undocumented way to copy a manager for another database: `Manager.db_manager(db_alias)`. It’s roughly the equivalent of `Queryset.using(db_alias)` for `Manager`. Even though that’s undocumented, it is used multiple times in Django tests, and it will explicitely raise an `AttributeError` if that API changes, while setting `.db` would probably not.

gasman:
Superseded by #3853

(@BertrandBordage, it turns out you didn't imagine commenting about `db_manager` after all :-) )

BertrandBordage:
Cool, I’m not going crazy! :+1:

Issue #3458, opened 2017-03-17
UX: Top bar is too green on green


maartenkling:
Multiple clients complaining about the green on green on green in the top where the main actions are.
<img width="667" alt="wagtail_-_home" src="https://cloud.githubusercontent.com/assets/946684/24032730/c48a7b94-0ae9-11e7-8935-10ec28030583.png">

heymonkeyriot:
@maartenkling see also #2224. Contrast is an issue for quite a few elements imho.

thibaudcolas:
Closing this now with the merge of #4788.

PR #3808, opened 2017-08-29
Toggleable high contrast mode [WIP]

kapito:
This isn't the most efficient solution and it definitely needs some work to be completely viable, but here's my attempt at reworking #3106 using @thibaudcolas' suggestion of having a toggleable high contrast mode.

Demonstration: https://gfycat.com/ImprobableAfraidDavidstiger

To do:

- [x] ~~Find alternative approach to apply the new styling (currently using jQuery which is bad for performance)~~ Now uses CSS, but jQuery is still needed to add the class on every page load...

- [ ] Make the setting account-based, rather than applying for everyone using Wagtail on that machine

- [ ] Improve the high contrast mode to comply 100% with [WCAG 2.0 standards](https://www.w3.org/TR/WCAG20/)

- [ ] Add tests

Related: #2224, #3106, #3458.

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
@@ -169,7 +169,7 @@ $(function() {
                 'color': '#fff'
             });
 
-            $('.button, .messages, .warning, .success').css({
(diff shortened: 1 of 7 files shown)


jjanssen:
Closing this PR since we recently improved the colour contrast within the release of 2.3.

PR #3810, opened 2017-08-29
Edit handlers simplification.

### LINKED ISSUE #3633: Chooser panels are missing classname
(Issue met on Wagtail version 1.10.1)

I faced this issue while putting `PageChooserPanel`s inside a `FieldRowPanel`, like this:
```python
FieldRowPanel([
    PageChooserPanel('project_left'),
    PageChooserPanel('project_right')
])
```
I got this exception while running any Django command:
```
  File "/home/florent/.virtualenvs/NoriPyt/lib/python3.5/site-packages/wagtail/wagtailadmin/edit_handlers.py", line 376, in bind_to_model
    if not re.search(r'\bcol\d+\b', child.classname):
AttributeError: 'PageChooserPanel' object has no attribute 'classname'
```

This issue also concerns other panels like `ImageChooserPanel`.

Here’s a workaround overriding the original `PageChooserPanel`:
```python
class PageChooserPanel(PageChooserPanel):
        self.classname = kwargs.pop('classname', '')
        super(PageChooserPanel, self).__init__(*args, **kwargs)

        obj = super(PageChooserPanel, self).bind_to_model(model)
        obj.classname = self.classname
        return obj
```

As you can see below, everything works fine with it, so I think we can add `classname` attributes to chooser panels without any issue. 

image

image

BertrandBordage:
This work was originally done by @fosmont and I to fix #3633.

In the end, we had to rewrite a vast part of the whole “edit handler” concept while keeping the tiniest impact possible for developers. The whole “edit handler” concept can still be improved a lot (and maybe renamed), but that’s a big enough change for now.

So, this pull request makes edit handlers way more pythonic and easy to extend. The main idea was to remove the redundancy and confusion introduced by the whole `Base…Panel` + `…Panel` + `_…Panel` stuff. Now there are just `…Panel` classes and instances. The classes are no longer used directly, we only use the instances we create using `panels = [FieldPanel('title'), …]`. These panel instances are bound to a model (like before), but now it creates a copy of the panel instance that is bound to the model, instead of creating a weird `_…Panel` class dynamically. Then we can bind the panel instance to a model instance using `.bind_to_instance()` (which also creates a copy of the panel instance) instead of  creating an instance of the weird dynamic `_…Panel` class.

As you can see on the changes made on apps defining other edit handlers (documents, images, snippets), it is way easier to create a new edit handler class.

And of course, this makes panels more consistent and fixes #3633, since we pass all `args`+`kwargs` to parent constructors instead of having to copy/paste them when creating a `_…Panel` class dynamically.

I also hope these changes will make it easier for advanced Wagtail users to introspect and read this part of the admin and… Create more 3rd party apps? ;)


wagtail/wagtailadmin/edit_handlers.py
```diff
@@ -95,59 +96,85 @@ class EditHandler(object):
...
+        new.model = model
+        new.on_model_bound()
+        return new
+
+    def on_model_bound(self):
+        pass
+
+    def bind_to_instance(self, instance=None, form=None):
+        new = (self.bind_to_model(self.model) if hasattr(self, 'model')
+               else self.clone())
```
gasman: I'm a bit lost about what's happening here. Is `self.bind_to_model(self.model)` really just saying "create a clone of this EditHandler that preserves its `model` property"? If so, let's create a new method to make that explicit. (In fact, can't we just make it the `clone` method's responsibility to copy the `model` attribute whenever it exists?)

 Also, is the `if hasattr(self, 'model')` test really needed here? Is there any circumstance where you would call `bind_to_instance` on an edit handler that _hasn't_ been bound to a model at that point?

BertrandBordage: Thanks for the review! I know this is pretty complex, due to the laziness we need in the model+instance binding… For these two particularly reasons, it was designed like this on purpose.

There are two reasons why I don’t assign `model` in `.clone()`:
- Simplicity: it would need to add `new = self.__class__(…)` + `new.model = self.model` + `return new` in all `clone` methods. Since this method is often changed in `EditHandler` subclasses (including third-party ones), it’s better like this. Of course, we can’t add `model` as an `EditHandler.__init__(…)` kwarg, or users would have a weird `model` argument available in `FieldPanel(…)` etc.
- The `on_model_bound` method. This method is really mandatory and can’t be put in `__init__` since the model is always bound to edit handlers **after** their initialization. If we want to bind models in `.clone()`, we have to do this:
```python
def __init__(self, …):
    …
    self.model = None

def clone(self):
    new = self.__class__(…)
    new.model = self.model
    if self.model is not None:
        new.on_model_bound()
    return new

    new = self.clone()
    new.model = model
    new.on_model_bound()
    return new
```

As you can see, this makes code more redundant, and it does more copies. Also, this doesn’t fully work. I tried it, we have to adjust code in a lot of places in order to get it working (in 30 minutes of adjustments, I couldn’t make the tests pass, the model binding is partially broken).

About the `if hasattr(self, 'model')` test, you’re right, models should always be bound at this point. There is a small possibility though: when you define the `edit_handler` class attribute of your model with some panels, then modify in place the list of its children in that particular case, the model will not be bound.
But in that case, the code I wrote will not work, it’s not `self.clone()` that should be called, but `self.bind_to_model(instance._meta.model)`. This this last statement should work in all case, I think I’ll add a test for this edge case and replace that line with `new = self.bind_to_model(instance._meta.model)`.

BertrandBordage: In fact, after testing and reading in detail the view logic, it’s not possible to bind an instance without the model bound. The reason is simple: to bind an instance we need a `form` argument. To get the form, we call `EditHandler.get_form_class(…)` that relies on the bound model.

BertrandBordage:
I just updated the pull request against master. Also, I made some investigations following Matthew’s comment. I made the correct fixes, see comments above.

BertrandBordage:
Pull request updated after the latest changes, especially the model rename.

gasman:
Looks good to me - thanks @fosmont and @BertrandBordage! Rebased and merged in 5fc191b116ee835762c282e13a5829fa873cc516 - just added a Django 2.0 fix (`rel` -> `remote_field` on SnippetChooserPanel) and a test for the specific issue in #3633 (https://github.com/gasman/wagtail/commit/b850ba256490c20e9cf43ef7211f84d6dbc5d503).

PR #3813, opened 2017-08-29
Update jquery to 2.2.4

nikolas:


thibaudcolas:
Hey @nikolas, thanks for the PR.

You need to provide us with information as to what is motivating this change: what you're trying to fix, what you've tested this on, and what it entails. I could see many reasons why updating jQuery (well, any dependency) would be hazardous, so please make your case.

nikolas:
I haven't run into any issues with jquery 2.2.1, but there are a handful of small issues fixed between that version and the latest of the 2.2.x series, which would be good to include: https://github.com/jquery/jquery/compare/2.2.1...2.2.4

thibaudcolas:
Ok, if there are no specific issues that affect Wagtail I'd rather hold off from any minor update like this. However stable jQuery is, however minor the upgrade is, it would require to re-test a _lot_ of stuff to make sure there are no regressions.

Closing.

PR #3814, opened 2017-08-29
Link to dev version of Django docs

nikolas:
Version 1.9 is out of date.

### DIFF
--- a/docs/topics/pages.rst
+++ b/docs/topics/pages.rst
@@ -4,15 +4,15 @@ Page models
 
 Each page type (a.k.a. content type) in Wagtail is represented by a Django model. All page models must inherit from the :class:`wagtail.wagtailcore.models.Page` class.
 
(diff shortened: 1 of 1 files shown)


thibaudcolas:
Hey @nikolas, appreciate every little improvement but you need to justify changes further than this for us to consider merging them.

I presume there is a reason why we have linked to exact versions in the past rather than "development", eg. so the (versioned) docs of Wagtail point to the version of Django they relate to, rather than "always bleeding edge".

gasman:
The risk with linking to the dev version of the docs is that if they decide to reorganise them at any point, it'll become a broken link. By linking to a specific version, we can be sure it's pointing somewhere meaningful, even if it's out of date.

thibaudcolas:
Fair point, closing this then.

@nikolas if you think specific links need to be updated (to a specific version) and can justify the change, feel free to make a PR for that.

PR #3819, opened 2017-08-30
Fix mobile nav menu height

### LINKED ISSUE #3707: Explorer options overflow container on Safari (mobile)
### Issue Summary

In the Wagtail explorer on Safari mobile, the last two options (specifically "Account settings" and "Log out") are both outside of the explorer's container making them hard to read (due to the white background).

Image: https://cdn.pbrd.co/images/GAIVfmL.png
How it should look (how it looks using Chrome on Android Lollipop): https://cdn.pbrd.co/images/GAIXO9S.png

### Steps to Reproduce

Log in to /admin/ using Safari mobile and open the explorer.

### Technical details

* Python version: 2.7.12
* Django version: 1.11.3
* Wagtail version: 1.11.1
* Browser version: Safari mobile on iOS 9

### LINKED ISSUE #3721: Mobile navigation menu scrolls past its height
When in its desktop form, the navigation menu is treated separately from the actual content and the content scrolls by itself: https://gfycat.com/SlowOffbeatAsiantrumpetfish

However, when in its mobile form, the menu scrolls *with* the content which looks rather odd as there is obviously nothing underneath the menu (so you get this area of nothing in the bottom-left corner): https://gfycat.com/MarvelousLateBighornsheep

Perhaps this is intentional or I'm missing something obvious, in which case please say, but it seems to me that the menu should be fixed and be independent of the content itself even in its mobile state.

kapito:
Fixes #3721. Currently the menu's height is set to match the viewport height; by setting it to use the content height instead, the nav menu looks much cleaner when you scroll past it on mobile.

image

Before you would see an area of nothing beneath the menu until the end of the page. 

Tested on latest Firefox, Chrome and mobile Safari.

(Also seems to have inadvertently fixed #3707)

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
@@ -97,7 +97,7 @@ $(function() {
 
     // Resize nav to fit height of content. This is an unimportant bell/whistle to make it look nice
     var fitNav = function() {
-        $('.nav-wrapper').css('min-height', $('.content').height() + 51); // 51 = height of footer
+        $('.nav-wrapper').css('min-height', $('.content').outerHeight()); // include padding/margin in height
         $('.nav-main').each(function() {
             var thisHeight = $(this).height();
             var footerHeight = $('#footer', $(this)).height();
@@ -106,7 +106,7 @@ $(function() {
 
     fitNav();
 
-    $(window).resize(function() {
+    $(document).resize(function() {
         fitNav();
     });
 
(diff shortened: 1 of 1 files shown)


wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
```diff
@@ -95,9 +95,9 @@ $(function() {
         $(this).find('em').toggleClass('icon-arrow-down-after icon-arrow-up-after');
     });
 
-    // Resize nav to fit height of document. This is an unimportant bell/whistle to make it look nice
+    // Resize nav to fit height of content. This is an unimportant bell/whistle to make it look nice
     var fitNav = function() {
-        $('.nav-wrapper').css('min-height', $(document).height());
+        $('.nav-wrapper').css('min-height', $('.content').height() + 64); // 64 = height of footer
```
zerolab: should you not get the actual footer height, rather than hard-coding it?

PR #3821, opened 2017-08-30
Delete re-uploaded document after saving the new one.

### LINKED ISSUE #3816: Reuploading a document doesn’t invalidate browser caches
Use case and steps to reproduce:
1. Upload a document
2. Link other objects to it, like pages, then realise something is wrong with the document (bad quality, typo, etc)
3. Reupload the file (usually with the same name) by editing the document
4. Download the file somewhere, from the front or back office, and you get the previous version if you already viewed it once and if you kept the same filename when reuploading it

This is because browsers can cache files like PDF documents it opens using a plugin.
When users reupload a document, they almost always have the same name as the original. And since HTTP servers are most of the time configured with a long expiry delay like a month or a year, people get stale data.

Django [has a way to deal with this](https://github.com/django/django/blob/master/django/core/files/storage.py#L61-L90) by looking for an available file name before saving the file and database row. It adds an underscore and a hash to the original filename until it finds an unused hash.

Unfortunately, Wagtail bypasses this behaviour [by deleting the document just before saving the new one](https://github.com/wagtail/wagtail/blob/v1.12/wagtail/wagtaildocs/views/documents.py#L133-L137).

I’m working on a pull request to fix this.

BertrandBordage:
Fixes #3816.

### DIFF
--- a/wagtail/wagtaildocs/tests/test_views.py
+++ b/wagtail/wagtaildocs/tests/test_views.py
@@ -35,34 +35,38 @@ class TestEditView(TestCase, WagtailTestUtils):
         Checks that reuploading the document file with the same file name
         changes the file name, to avoid browser cache issues (see #3816).
         """
-        name = 'test_edit_view.txt'
-        new_file = SimpleUploadedFile(name, b'An updated test content.')
+        old_file = self.document.file
+        new_name = self.document.filename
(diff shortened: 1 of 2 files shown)


BertrandBordage:
#3822 is a similar fix, but for images.

gasman:
Merged in 65e833603a97367b5938e7a032fcc7be459f5476 - thanks!

PR #3822, opened 2017-08-30
Delete re-uploaded image after saving the new one.

### LINKED ISSUE #3817: Reuploading an image doesn’t invalidate browser caches
Use case and steps to reproduce:
1. Upload an image
2. Link other objects to it, like pages, then realise something is wrong with the image (bad quality, wrong contrast, etc)
3. Reupload the file (usually with the same name) by editing the image
4. Download the file somewhere, from the front or back office, and you get the previous version if you already viewed it once and if you kept the same filename when reuploading it

This is because browsers can cache images.
When users reupload an image, they almost always have the same name as the original. And since HTTP servers are most of the time configured with a long expiry delay like a month or a year, people get stale data.

Django [has a way to deal with this](https://github.com/django/django/blob/master/django/core/files/storage.py#L61-L90) by looking for an available file name before saving the file and database row. It adds an underscore and a hash to the original filename until it finds an unused hash.

Unfortunately, Wagtail bypasses this behaviour [by deleting the image just before saving the new one](https://github.com/wagtail/wagtail/blob/v1.12/wagtail/wagtailimages/views/images.py#L102-L106).

Note that **there is a small difference with #3816**, as images also have renditions/thumbnails. As we can see [in the code](https://github.com/wagtail/wagtail/blob/v1.12/wagtail/wagtailimages/views/images.py#L107), all renditions are correctly removed, but since the name of the image file is kept the same, all renditions file name also stay the same, even if the rendition objects and files are deleted and created again later. So the same browser cache issue happens with renditions. This should be fixed by doing nothing more than the same fix as for #3816, but we have to check that renditions work anyway.

I’m working on a pull request to fix this.

BertrandBordage:
Fixes #3817.

### DIFF
--- a/wagtail/wagtailimages/tests/test_admin_views.py
+++ b/wagtail/wagtailimages/tests/test_admin_views.py
@@ -283,6 +283,11 @@ class TestImageEditView(TestCase, WagtailTestUtils):
             file=get_test_image_file(),
         )
 
+        self.storage = self.image.file.storage
+
+    def update_from_db(self):
+        self.image = Image.objects.get(pk=self.image.pk)
+
     def get(self, params={}):
         return self.client.get(reverse('wagtailimages:edit', args=(self.image.id,)), params)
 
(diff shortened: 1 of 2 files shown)


BertrandBordage:
#3821 is a similar fix, but for documents.

gasman:
Merged in 37be77143d95f5a1572342eb76c50a3dfb821420

PR #3824, opened 2017-08-31
Batch frontend cache invalidation

### LINKED ISSUE #3750: [Enhancement] Bulk purging with wagtailfrontendcache
Occasionally it can be useful to purge a lot or even all of the frontend. (For example, if site menus are stored as snippets, one might want to purge the entire site when a change is made to the menus.) I'm fairly certain all or most of the cache services we have backends for provide an API for doing this. It would be great for wagtailfrontendcache to provide a util for using those.

kaedroho:
This PR implements the first half of [RFC 20](https://github.com/wagtail/rfcs/issues/20). Allowing URLs to be invalidated from the frontend cache in batches rather than one at a time. This should give a performance boost for invalidating URLs in external services like CloudFront and CloudFlare as these support batch invalidation.

In its current form, only users who use ``purge_page_from_cache`` on pages that have a lot of cached URLs will see an automatic performance benefit and some users may be able to rewrite their invalidation logic using the new ``PurgeBatch`` class to improve performance as well. But as this logic is often dotted around in many signal handlers, it's still difficult to write something that'll invalidate everything that needs to be invalidated in a single batch.

The solution to that is to allow the invalidation API to be configured as asynchronous, which will allow Wagtail to automatically group requests into larger batches and submit them in the background. This will come in a future pull request.

Fixes #3750

### DIFF
--- a/docs/reference/contrib/frontendcache.rst
+++ b/docs/reference/contrib/frontendcache.rst
@@ -149,7 +149,7 @@ Invalidating index pages
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
 Pages that list other pages (such as a blog index) may need to be purged as
-well so any changes to a blog page is also reflected on the index (for example,
+well so any changes to a blog page are also reflected on the index (for example,
 a blog post was added, deleted or its title/thumbnail was changed).
 
 To purge these pages, we need to write a signal handler that listens for
@@ -180,7 +180,7 @@ This signal handler would trigger the invalidation of the index page using the
         batch.purge()
 
 
-    @receiver(page_published, sender=BlogPage):
+    @receiver(page_published, sender=BlogPage)
     def blog_published_handler(instance):
         blog_page_changed(instance)
 
--- a/docs/reference/contrib/frontendcache.rst
+++ b/docs/reference/contrib/frontendcache.rst
@@ -145,75 +145,18 @@ By default, Wagtail will only purge one URL per page. If your page has more than
             for page_number in range(1, self.get_blog_items().num_pages + 1):
                 yield '/?page=' + str(page_number)
 
-
 Invalidating index pages
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
-Another problem is pages that list other pages (such as a blog index) will not
-be purged when a blog entry gets added, changed or deleted. You may want to
-purge the blog index page so the updates are added into the listing quickly.
+Pages that list other pages (such as a blog index) may need to be purged as
+well so any changes to a blog page is also reflected on the index (for example,
+a blog post was added, deleted or its title/thumbnail was changed).
 
-This can be solved by using the ``purge_page_from_cache`` utility function
-which can be found in the ``wagtail.contrib.wagtailfrontendcache.utils`` module.
-
-Let's take the the above BlogIndexPage as an example. We need to register a
-signal handler to run when one of the BlogPages get updated/deleted. This
-signal handler should call the ``purge_page_from_cache`` function on all
-BlogIndexPages that contain the BlogPage being updated/deleted.
-
-
-.. code-block:: python
-
-    # models.py
-    from django.dispatch import receiver
-    from django.db.models.signals import pre_delete
-
-    from wagtail.wagtailcore.signals import page_published
-    from wagtail.contrib.wagtailfrontendcache.utils import purge_page_from_cache
-
-
-    ...
-
-
-    def blog_page_changed(blog_page):
-        # Find the blog index and purge it. Assuming there is just one on the site.
-        blog_index = BlogIndexPage.objects.first()
-        purge_page_from_cache(blog_index)
-
-
-    @receiver(page_published, sender=BlogPage):
-    def blog_published_handler(instance):
-        blog_page_changed(instance)
-
-
-    @receiver(pre_delete, sender=BlogPage)
-    def blog_deleted_handler(instance):
-        blog_page_changed(instance)
-
-
-Invalidating individual URLs
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-``wagtail.contrib.wagtailfrontendcache.utils`` provides a function called ``purge_url_from_cache``. As the name suggests, this purges an individual URL from the cache.
-
-For example, this could be useful for purging a single page of blogs:
-
-.. code-block:: python
-
-    from wagtail.contrib.wagtailfrontendcache.utils import purge_url_from_cache
-
-    # Purge the first page of the blog index
-    purge_url_from_cache(blog_index.url + '?page=1')
-
-
-Invaliding many URLs in batches
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-.. versionadded:: 1.13
-
-If you have many pages/URLs that need to be purged at the same time, it's more
-perfomant to use Wagtail's ``PurgeBatch`` class to build up a list of pages/URLs
-to purge, and purge them all in a single request.
+To purge these pages, we need to write a signal handler that listens for
+Wagtail's ``page_published`` and ``page_unpublished`` signals for blog pages
+(note, ``page_published`` is called both when a page is created and updated).
+This signal handler would trigger the invalidation of the index page using the
+``PurgeBatch`` class which is used to construct and dispatch invalidation requests.
 
 .. code-block:: python
 
@@ -224,9 +167,7 @@ to purge, and purge them all in a single request.
     from wagtail.wagtailcore.signals import page_published
     from wagtail.contrib.wagtailfrontendcache.utils import PurgeBatch
 
-
     ...
-
 
     def blog_page_changed(blog_page):
         # Find all the live BlogIndexPages that contain this blog_page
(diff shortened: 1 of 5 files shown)


wagtail/contrib/wagtailfrontendcache/utils.py
```diff
@@ -56,18 +56,84 @@ def get_backends(backend_settings=None, backends=None):
...
+
+
+def _get_page_cached_urls(page):
     page_url = page.full_url
     if page_url is None:  # nothing to be done if the page has no routable URL
-        return
+        return []
+
+    return [
+        page_url + path[1:]
```
chosak: Just an observation: I know that this duplicates [the existing logic](https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/wagtailfrontendcache/utils.py#L72), but I wonder if using `path[1:]` as a way of stripping the leading slash is a bit of a blunt tool as opposed to something like `path.lstrip('/')`.  [The docs demonstrate](http://docs.wagtail.io/en/stable/reference/contrib/frontendcache.html#invalidating-more-than-one-url-per-page) using a trailing slash but I can imagine a developer omitting it by mistake, thus leading to hard-to-debug errors where the first character of the subpath is cut off.

docs/reference/contrib/frontendcache.rst
```diff
@@ -170,11 +176,9 @@ Let's take the the above BlogIndexPage as an example. We need to register a sign
 
 
     def blog_page_changed(blog_page):
-        # Find all the live BlogIndexPages that contain this blog_page
-        for blog_index in BlogIndexPage.objects.live():
-            if blog_page in blog_index.get_blog_items().object_list:
-                # Purge this blog index
-                purge_page_from_cache(blog_index)
+        # Find the blog index and purge it. Assuming there is just one on the site.
```
chosak: The existing docs ("Find all the live BlogIndexPages") and code seem more consistent with the [logic below](https://github.com/wagtail/wagtail/pull/3824/files#diff-ce9d3dbe263a870a67b78eef6a26e59aR231) and the text above ("This signal handler should call the ``purge_page_from_cache`` function on all BlogIndexPages that contain the BlogPage being updated/deleted.").

docs/reference/contrib/frontendcache.rst
```diff
@@ -200,3 +204,62 @@ For example, this could be useful for purging a single page of blogs:
...
+        batch = PurgeBatch()
+        for blog_index in BlogIndexPage.objects.live():
+            if blog_page in blog_index.get_blog_items().object_list:
+                batch.add_page(blog_index)
+
+        # Purge all the blog indexes we found in a single request
+        batch.purge()
+
+
+    @receiver(page_published, sender=BlogPage):
```
chosak: Remove extra trailing `:`.

wagtail/contrib/wagtailfrontendcache/tests.py
```diff
@@ -143,21 +146,63 @@ def purge(self, url):
         'BACKEND': 'wagtail.contrib.wagtailfrontendcache.tests.MockBackend',
     },
 })
-class TestCachePurging(TestCase):
+class TestCachePurgingFunctions(TestCase):
+
+    fixtures = ['test.json']
+
+    def test_purge_url_from_cache(self):
+        PURGED_URLS[:] = []  # reset PURGED_URLS to the empty list
```
chosak: Can this be moved to the `setUp` of this test case, to avoid duplication in every test?

wagtail/contrib/wagtailfrontendcache/backends.py
```diff
@@ -91,20 +97,20 @@ def purge(self, url):
                 if response.status_code != 200:
                     response.raise_for_status()
                 else:
-                    logger.error("Couldn't purge '%s' from Cloudflare. Unexpected JSON parse error.", url)
+                    logger.error("Couldn't purge from Cloudflare. Unexpected JSON parse error.")
```
chosak: Would it be nice to include the list of URLs that failed for consistency here, to maintain current behavior? It'd be useful for logging purposes to know which URLs failed. Same comment applies to various other logging statements below.

wagtail/contrib/wagtailfrontendcache/utils.py
```diff
@@ -56,18 +56,84 @@ def get_backends(backend_settings=None, backends=None):
...
+
+def purge_pages_from_cache(pages, backend_settings=None, backends=None):
+    urls = []
+    for page in pages:
+        urls.extend(_get_page_cached_urls(page))
+
+    purge_urls_from_cache(urls, backend_settings, backends)
+
+
+class PurgeBatch(object):
```
chosak: It'd be nice to have some unit tests for this new class.

docs/reference/contrib/frontendcache.rst
```diff
@@ -200,3 +204,62 @@ For example, this could be useful for purging a single page of blogs:
...
+    def blog_published_handler(instance):
+        blog_page_changed(instance)
+
+
+    @receiver(pre_delete, sender=BlogPage)
+    def blog_deleted_handler(instance):
+        blog_page_changed(instance)
+
+
+All of the methods available on ``PurgeBatch`` are listed bellow:
```
chosak: Typo: "bellow".

docs/reference/contrib/frontendcache.rst
```diff
@@ -200,3 +204,62 @@ For example, this could be useful for purging a single page of blogs:
 
     # Purge the first page of the blog index
     purge_url_from_cache(blog_index.url + '?page=1')
+
+
+Invaliding many URLs in batches
```
chosak: Typo: "Invaliding".

wagtail/contrib/wagtailfrontendcache/utils.py
```diff
@@ -56,18 +56,84 @@ def get_backends(backend_settings=None, backends=None):
...
+        for page in pages:
+            self.add_page(page)
 
-    for backend_name, backend in get_backends(backend_settings=backend_settings, backends=backends).items():
-        # Purge cached paths from cache
-        for path in page.specific.get_cached_paths():
-            logger.info("[%s] Purging URL: %s", backend_name, page_url + path[1:])
-            backend.purge(page_url + path[1:])
+    def purge(self, backend_settings=None, backends=None):
+        """Performs the purge of all the URLs in this batch"""
```
chosak: It'd be helpful to have a sentence of two of documentation about what the `backend_settings` and `backends` arguments mean for this method. It's clear if you look [in the code](https://github.com/wagtail/wagtail/blob/c6666c6de5e83bf94d18324858c121e6584ba47d/wagtail/contrib/wagtailfrontendcache/utils.py#L16) but otherwise a bit non-obvious.

docs/reference/contrib/frontendcache.rst
```diff
@@ -200,3 +204,62 @@ For example, this could be useful for purging a single page of blogs:
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. versionadded:: 1.13
+
+If you have many pages/URLs that need to be purged at the same time, it's more
+perfomant to use Wagtail's ``PurgeBatch`` class to build up a list of pages/URLs
```
chosak: Typo: "perfomant".

tomdyson:
Thanks for the excellent review, @chosak! I'm sure @kaedroho will respond when he returns from holiday next week.

wagtail/contrib/wagtailfrontendcache/utils.py
kaedroho: Ahh yes, I hadn't thought of that. Will implement with ``.lstrip('/')`` instead

wagtail/contrib/wagtailfrontendcache/utils.py
```diff
@@ -56,18 +56,84 @@ def get_backends(backend_settings=None, backends=None):
...
-        return
+        return []
+
+    return [
+        page_url + path[1:]
+        for path in page.specific.get_cached_paths()
+    ]
+
+
+def purge_page_from_cache(page, backend_settings=None, backends=None):
```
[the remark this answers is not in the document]
kaedroho: Good suggestion, will do this

wagtail/contrib/wagtailfrontendcache/utils.py
```diff
@@ -56,18 +56,84 @@ def get_backends(backend_settings=None, backends=None):
...
+
+    purge_urls_from_cache(urls, backend_settings, backends)
+
+
+def purge_pages_from_cache(pages, backend_settings=None, backends=None):
+    urls = []
+    for page in pages:
+        urls.extend(_get_page_cached_urls(page))
+
+    purge_urls_from_cache(urls, backend_settings, backends)
```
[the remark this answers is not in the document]
kaedroho: Yep, I agree. Thanks!

docs/reference/contrib/frontendcache.rst
kaedroho: I might have been both trying to avoid providing inefficient code in the example and introducing batch invalidation early.

After some thought, I eventually opted for ditching ``purge_page_from_cache`` entirely from the docs and suggest people always use ``PurgeBatch``. I think it's simpler this way as having two different APIs do the same thing could be confusing, and ``PurgeBatch`` is the more flexible of the two.

I'm wondering if we should add documentation for them anyway, as there is code out there using them. It might be a good idea to deprecate them, but I don't think they're a problem for maintenance anyway. Do you have any thoughts on this?

kaedroho:
Thanks @chosak! I've implemented your suggestions and made a few extra changes to the docs.

docs/reference/contrib/frontendcache.rst
zerolab: @kaedroho the trailing `:` is still there

docs/reference/contrib/frontendcache.rst
chosak: @kaedroho it might be nice to keep the docs mention of `purge_page_from_cache` as it's a simpler way to purge a single page, as opposed to creating a `PurgeBatch` and we'd want to keep it around for backwards compatibility (at least for now).

docs/reference/contrib/frontendcache.rst
```diff
@@ -145,22 +145,18 @@ By default, Wagtail will only purge one URL per page. If your page has more than
...
-
-This can be solved by using the ``purge_page_from_cache`` utility function
-which can be found in the ``wagtail.contrib.wagtailfrontendcache.utils`` module.
-
-Let's take the the above BlogIndexPage as an example. We need to register a
-signal handler to run when one of the BlogPages get updated/deleted. This
-signal handler should call the ``purge_page_from_cache`` function on all
-BlogIndexPages that contain the BlogPage being updated/deleted.
+Pages that list other pages (such as a blog index) may need to be purged as
+well so any changes to a blog page is also reflected on the index (for example,
```
chosak: Should be "any changes are" not "any changes is".

kaedroho:
Done, thanks again!

gasman:
Merged in de3f9550b4877f3bcbad981f5a65145e27c09ea8

PR #3832, opened 2017-09-05
Replaceable login form

mikedingjan:
Allow developers to extend the ``LoginForm`` with their own functionality.  This is useful when implementing for example: rate limiting, captcha or 2fa.

The login template is adjusted as well to have some kind of default implementation when adding more fields to the login form.


gasman:
Merged in ef27853a05d9c65ba9a14e3588890783b19d4fcf - thanks @mikedingjan!

PR #3833, opened 2017-09-05
Make it possible to run migrations backwards

vstoykov:
This will allow to run migrations backwards instead of raising
```
django.db.migrations.exceptions.IrreversibleError: Operation <RunPython <function draft_title at 0x7f602143dc80>> in wagtailcore.0040_page_draft_title is not reversible
```


gasman:
Thanks @vstoykov! Merged in 90ce50307e5db1ef1e9da0ea5eb8f6f311d55c70 (master) and f0e9258af95beecab420f5cad051b01f6cbf5d01 (1.12.x).

PR #3840, opened 2017-09-07
Don't call `static()` when constructing HalloPlugins

### LINKED ISSUE #3829: Not possible to migrate before running collectstatic
### Issue Summary

Our deployment scripts will first migrate, then run collectstatic when deploying a Django site. A recent release of Wagtail broke this strategy and causes an exception when running the migrate management command before running collectstatic.

### Steps to Reproduce

Start a project and make sure the following settings are set:

    DEBUG = False
    STATICFILES_STORAGE = 'django.contrib.staticfiles.storage.ManifestStaticFilesStorage'`

Run `./manage.py migrate`

This will error with:

    ValueError: Missing staticfiles manifest entry for 'wagtailadmin/js/hallo-plugins/hallo-requireparagraphs.js'

If you run collectstatic before migrate everything is fine.

### Cause

This seems to be caused by the call to `static` here: https://github.com/wagtail/wagtail/blob/9942822/wagtail/wagtailadmin/rich_text.py#L77

Since this happens at module level this call will happend as soon as this file is imported. Which apparently it is when migrate is called.

The easiest solution to me seems to just move this bit of code to where it's used: https://github.com/wagtail/wagtail/blob/9942822/wagtail/wagtailadmin/rich_text.py#L104

gasman:
This isn't necessary, since Django form media fills in static paths itself - and doing it at module level will break on startup with certain staticfiles configurations if collectstatic hasn't run yet. Fixes #3829

### DIFF
--- a/docs/advanced_topics/customisation/page_editing_interface.rst
+++ b/docs/advanced_topics/customisation/page_editing_interface.rst
@@ -105,7 +105,7 @@ Once the plugin has been created, it should be registered as a rich text feature
             'hallo', 'blockquote',
             HalloPlugin(
                 name='halloblockquote',
-                js=[static('myapp/js/hallo-blockquote.js')],
+                js=['myapp/js/hallo-blockquote.js'],
             )
         )
 
--- a/wagtail/tests/testapp/wagtail_hooks.py
+++ b/wagtail/tests/testapp/wagtail_hooks.py
@@ -99,7 +99,7 @@ def register_blockquote_feature(features):
     features.register_editor_plugin(
         'hallo', 'blockquote', HalloPlugin(
             name='halloblockquote',
-            js=[static('testapp/js/hallo-blockquote.js')],
-            css={'all': [static('testapp/css/hallo-blockquote.css')]},
+            js=['testapp/js/hallo-blockquote.js'],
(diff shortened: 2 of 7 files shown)


gasman:
Note: this will need testing against Django 1.8, because the Django form media was only updated to use `django.contrib.staticfiles` in Django 1.10:
https://docs.djangoproject.com/en/1.11/topics/forms/media/#paths-in-asset-definitions

gasman:
On Django 1.8, removing the explicit `static` means that `ManifestStaticFilesStorage` hash suffixes aren't added - so if a user happens to be running Django 1.8 and relying on `ManifestStaticFilesStorage`, then this is a regression. (As it happens, wagtail.contrib.settings and wagtail.contrib.table_block already use form media without explicit `static`, so these would already be affected by the lack of cache-busting on such a setup.)

I think this is more of a Django bug than a Wagtail one, though: converting a relative static path to a full URL is something that Django form media was _meant_ to do all along, and the fact that it didn't do so in a staticfiles-compliant way is not something Wagtail should be expected to work around, IMO.

gasman:
Merged in 92b4f68c67e36ee63b0902dbbf5187e8cf243b0c (master) and 28907c14db284f7889b15262709a5cf2d18fd6e5 (stable/1.12.x).

PR #3843, opened 2017-09-07
Fix OEmbed endpoints

### LINKED ISSUE #3842: OEmbed bug with Vimeo (plus some others)
When trying to embed Vimeo videos, it fails with `EmbedNotFoundException`

To duplicate:
```python
from wagtail.wagtailembeds.finders.oembed import OEmbedFinder

finder = OEmbedFinder()
finder.find_embed('https://vimeo.com/8733915')
```

This happens because some of the endpoints have `{format}` in them. `{format}` is meant to be replaced with `json`, but it isn't. There's a line where `{format}` is being replaced with `json` in the urls, but as far as I can tell, urls are just used for pattern matching.

https://github.com/wagtail/wagtail/blob/92b4f68c67e36ee63b0902dbbf5187e8cf243b0c/wagtail/wagtailembeds/finders/oembed.py#L26-L33

Likely a simple mistake.

This issue is new with Wagtail 1.12 introduced in https://github.com/wagtail/wagtail/commit/01e2d5f4a0e94fcb5fdaee0dc8ea93455b6a00ab

mixxorz:
Fixes #3842

### DIFF
--- a/wagtail/wagtailembeds/finders/oembed.py
+++ b/wagtail/wagtailembeds/finders/oembed.py
@@ -26,11 +26,12 @@ class OEmbedFinder(EmbedFinder):
         for provider in providers or all_providers:
             patterns = []
 
+            endpoint = provider['endpoint'].replace('{format}', 'json')
+
             for url in provider['urls']:
-                url = url.replace('{format}', 'json')
                 patterns.append(re.compile(url))
 
-            self._endpoints[provider['endpoint']] = patterns
+            self._endpoints[endpoint] = patterns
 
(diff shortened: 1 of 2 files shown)


gasman:
Thanks @mixxorz! Merged in 83e6a551febb9aa55ba074a5f14a4928264f4172 (master) and bdeb104804acf57861c54bbf23553d45cbb741a1 (stable/1.12.x), with a rewrite to the test (so that it's checking the end-to-end behaviour rather than just the internal property).

Issue #3838, opened 2017-09-07
The document chooser: Sorry, no documents match ""


m1kola:
### Issue Summary

The document chooser shows the following message by default when the side doesn't have any documents:
> Sorry, no documents match ""

![screen shot 2017-09-07 at 12 01 03](https://user-images.githubusercontent.com/509198/30160444-422d1ba2-93c4-11e7-8f63-636352e80c92.png)

# Steps to reproduce

1. Define a document chooser panel in a streamfield or as a separate field
2. Remove all documents from the document library, if it has any
3. Click the "Choose a document" button in the editor interface.

### Technical details

* Python version: 3.6.1
* Django version: 1.11.5
* Wagtail version: 1.12.1

thibaudcolas:
Any takers to quickly define what this UI should show when there are no documents, so that contributors can more easily pick this up and go? Ping @wagtail/ux-design.

lb-:
@wagtail/ux-design ideally it should default to the `Upload` tab being selected first if there are no documents present.

The `Search` tab could say something like:
> No documents uploaded, upload one now.
With the text `upload one now` being a link to the internal modal tab of Upload.

nott:
I'm going to work on this during the sprint in Minsk

PR #3845, opened 2017-09-09
#3838 - Fixes document chooser messages

gmmoraes:
It think this solves #3838 

If the description below is too confusing I can try to edit it.

My first approach was adding an "if condition" to the "else" in order to show the "Sorry, no documents match" message only if the user made a search, although this message would still show up if the user made a search and there were no documents uploaded in his account.
```
{% load i18n %}
{% if documents %}
    {% if is_searching %}
        <h2>
        {% blocktrans count counter=documents.paginator.count %}
            There is one match
        {% plural %}
            There are {{ counter }} matches
        {% endblocktrans %}
        </h2>
    {% else %}
        <h2>{% trans "Latest documents" %}</h2>
    {% endif %}

    {% include "wagtaildocs/documents/list.html" with choosing=1 %}

    {% include "wagtailadmin/shared/pagination_nav.html" with items=documents is_ajax=1 %}
{% else %}
    {% if is_searching %}
        <p>{% blocktrans %}Sorry, no documents match "<em>{{ query_string }}</em>"{% endblocktrans %}</p>
    {% else %}
        {% url 'wagtaildocs:add_multiple' as wagtaildocs_add_document_url %}
        <p>{% blocktrans %}You haven't uploaded any documents. Why not <a href="{{ wagtaildocs_add_document_url }}">upload one now</a>?{% endblocktrans %}</p>
    {% endif %}
{% endif %}
```

In the chooser.py file I have noticed that when a user searched for a document in the chooser panel it changed filtered the "document" variable for that querry so even if the user had some documents uploaded to his account you woudnt be able to use the "document" variable in the template file to differentiate the scenario where the user has documents, but none of these documents match with his search to the scenario where the user has no documents.So in the chooser.py file I have created a new variable called "saved_document" which as the "document" variable contains the uploaded documents by the user, but it is not filtered when the user makes a search.I have used the "saved_document" variable to differentiate between the two scenarios listed above in the results.html file.

### DIFF
--- a/wagtail/wagtaildocs/templates/wagtaildocs/chooser/results.html
+++ b/wagtail/wagtaildocs/templates/wagtaildocs/chooser/results.html
@@ -16,5 +16,10 @@
 
     {% include "wagtailadmin/shared/pagination_nav.html" with items=documents is_ajax=1 %}
 {% else %}
-    <p>{% blocktrans %}Sorry, no documents match "<em>{{ query_string }}</em>"{% endblocktrans %}</p>
+    {% if saved_documents %}
+        <p>{% blocktrans %}Sorry, no documents match "<em>{{ query_string }}</em>"{% endblocktrans %}</p>
+    {% else %}
+        {% url 'wagtaildocs:add_multiple' as wagtaildocs_add_document_url %}
+        <p>{% blocktrans %}You haven't uploaded any documents. Why not <a href="{{ wagtaildocs_add_document_url }}">upload one now</a>?{% endblocktrans %}</p>
+    {% endif %}
 {% endif %}
--- a/wagtail/wagtaildocs/views/chooser.py
+++ b/wagtail/wagtaildocs/views/chooser.py
@@ -49,6 +49,7 @@ def chooser(request):
         documents = hook(documents, request)
 
     q = None
+    saved_documents = Document.objects.all()
     if 'q' in request.GET or 'p' in request.GET or 'collection_id' in request.GET:
 
         collection_id = request.GET.get('collection_id')
@@ -72,6 +73,7 @@ def chooser(request):
             'documents': documents,
             'query_string': q,
             'is_searching': is_searching,
+            'saved_documents': saved_documents,
         })
     else:
         searchform = SearchForm()


wagtail/wagtaildocs/views/chooser.py
```diff
@@ -49,6 +49,7 @@ def chooser(request):
         documents = hook(documents, request)
 
     q = None
+    saved_documents = Document.objects.all()
```
loicteixeira: Since you never use the actual documents objects, I reckon this could use `Document.objects.exists()` instead. In this case, a variable name like `has_documents` will better express the intent.

In addition, this can probably live within the `if` statement so it only runs whenever the variable is actually used.

loicteixeira:
Since we're looking at fixing the DocumentChooser, we might want to apply the same fix to the ImageChooser which will show the [same erroneous message](https://github.com/wagtail/wagtail/blob/921cee4707ea21363d27838370b8f38d8065a3a1/wagtail/wagtailimages/templates/wagtailimages/images/results.html#L32-L33) if there are no images.

Furthermore, the ImageChooser displays different messages whether the [search is performed within a collection](https://github.com/wagtail/wagtail/blob/921cee4707ea21363d27838370b8f38d8065a3a1/wagtail/wagtailimages/templates/wagtailimages/images/results.html#L38-L42) which we might want to replicate here.

gmmoraes:
@loicteixeira Thank you for the reply and for taking your time to read my post!

If possible I would like to try fixing the ImageChooser as well, is there any issue opened for it?

loicteixeira:
Sorry it took me so long to get back to you.

I could not find an open issue for the `ImageChooser`. That being said, before working on the `ImageChooser`, since you're keen on doing a bit more, I reckon there is some work left on the `DocumentChooser`, i.e. make it work with `collections` the way the `ImageChooser` does.

For that we need to have `current_collection` available in the context (see the [document index view](https://github.com/gmmoraes/wagtail/blob/637daf204ad0aec483a26a8749ddd6f03b87d086/wagtail/wagtaildocs/views/documents.py#L40-L47) on how to get the `current_collection` variable) so the template can use it (see the [image chooser template](https://github.com/wagtail/wagtail/blob/921cee4707ea21363d27838370b8f38d8065a3a1/wagtail/wagtailimages/templates/wagtailimages/images/results.html#L38-L42) for an example). In your case, this would go in the `else` block of `if has_documents`.

In addition, `has_documents` will need to be filtered by `collection` (just the collection, not the rest of the search query to avoid situation like you've describe in your original post) so the message matches.

Good luck and thanks again for your help.

gmmoraes:
@loicteixeira no problem, I am glad to help with whenever I can!

I think that I have found a possible issue.
When a user has no images and tries to upload a new image to a specific collection by clicking in the link provided by the message "You haven't any images in this collection. Why not upload one now?" he is not redirected to the add multiples images" page with the selected collection.
This pattern is being replicate with documents as well.

Steps to reproduce:
1. Delete all of your images (if you have any)
2. Go to the Images area in the admin interface (http://localhost:8000/admin/images/)
3. Select a collection (If you want to create a specific collection to test this click on setting in the left bar then select collection)
4. Click on the link provided by the message "You haven't any images in this collection. Why not upload one now?" 


Alt Text

loicteixeira:
Nice catch!

Unfortunately this will be quite involved to fix:
1. the `url` tag does not allow to pass additional parameters to generate query strings automatically, so you would need to do `{{ wagtaildocs_add_document_url }}?collection_id={{collection_id}}.
1. the [collection_id](https://github.com/gmmoraes/wagtail/blob/631609e7dcc6aba27a6151bf1010c96f0cc78c1b/wagtail/wagtaildocs/views/chooser.py#L55) isn't part of the [context](https://github.com/gmmoraes/wagtail/blob/631609e7dcc6aba27a6151bf1010c96f0cc78c1b/wagtail/wagtaildocs/views/chooser.py#L74-L77) and will need to be added to it so it can be used by the template in step 1.
1. The [add](https://github.com/wagtail/wagtail/blob/b3e126d6b9225094d894c7567aa29c589b2e2652/wagtail/wagtaildocs/views/documents.py#L92-L116) and [add_multiple](https://github.com/wagtail/wagtail/blob/b3e126d6b9225094d894c7567aa29c589b2e2652/wagtail/wagtaildocs/views/multiple.py#L23-L82) do not fetch the collection form querystring and would need to do so.
1. The `collection_id` fetched on step 3 would need to be passed to the form as `initial` value.

So again, quite a bit of work. In addition, I know there is some work going at the moment to update the views of the admin to class based view in order to reuse components (as we've seen, the document and image choosers are very similar but are implemented quite differently).

Therefore I reckon we could complete and merge this PR (which already contains an important fix) and later open a separate issue for the collection. What do you think?

wagtail/wagtaildocs/templates/wagtaildocs/chooser/results.html
```diff
@@ -16,5 +16,14 @@
 
     {% include "wagtailadmin/shared/pagination_nav.html" with items=documents is_ajax=1 %}
 {% else %}
-    <p>{% blocktrans %}Sorry, no documents match "<em>{{ query_string }}</em>"{% endblocktrans %}</p>
+    {% if has_documents %}
+        <p>{% blocktrans %}Sorry, no documents match "<em>{{ query_string }}</em>"{% endblocktrans %}</p>
+    {% else %}
+        {% url 'wagtaildocs:add_multiple' as wagtaildocs_add_document_url %}
+        {% if current_collection %}
```
loicteixeira: `current_collection` is missing from the context. It needs to be passed alongside of `has_document` from the view. e.g.
```python
return render(request, "wagtaildocs/chooser/results.html", {
    'documents': documents,
    'query_string': q,
    'is_searching': is_searching,
    'has_documents': has_documents,
    'current_collection': collection_id,
})
```

Issue #3784, opened 2017-08-23
Styles for the preview button are broken on a page with more than one preview mode


m1kola:
### Issue Summary

Styles for the preview button are broken on a page with more than one preview mode (on form page, for example).

### Steps to Reproduce

1. Define a form page model using `wagtail.wagtailforms`
2. Go to the Wagtail admin UI to create a new page.

### Result

<img width="439" alt="screen shot 2017-08-23 at 13 12 07" src="https://user-images.githubusercontent.com/509198/29615295-48837a98-8805-11e7-9a45-724c2daeadd0.png">
<img width="457" alt="screen shot 2017-08-23 at 13 12 14" src="https://user-images.githubusercontent.com/509198/29615296-4889b912-8805-11e7-8b70-4c1f86b599aa.png">

### Technical details

* Django version: 1.11.3
* Wagtail version: 1.12 (Can't reproduce on 1.11.1)
* OS: MacOS 10.12.5 (16F73)
* Browser versions:
  * Google Chrome 60.0.3112.101
  * Safari 10.1.1 (12603.2.4)
  * FireFox Developer Edition 56.0b2

kapito:
I think I know the cause for this, will release a PR fixing it ASAP.

Issue #3718, opened 2017-07-18
Footer collapses too early


kapito:
The width at which the footer collapses is currently way too large, as can be seen in this image:

![image](https://user-images.githubusercontent.com/14837124/28310891-cbc0f86c-6ba5-11e7-8f0d-51b5d1c75157.png)

This appears to be caused by the preview button having a parent div much wider than itself.

![image](https://user-images.githubusercontent.com/14837124/28311096-7acb01b8-6ba6-11e7-9e69-a19514b1e071.png)

Thanks to @tomdyson for pointing this out

PR #3848, opened 2017-09-11
Fix rendering of wagtailsettings save button (#3836)

### LINKED ISSUE #3836: Styles for the save button on settings pages are broken
### Issue Summary

The button styles for the save button on custom settings pages does not appear to have padding on the right side.

screen shot 2017-09-06 at 1 05 07 pm

### Steps to Reproduce

1. Create a new settings model that inherits from wagtail.contrib.settings.models.BaseSettings and registered with Wagtail using wagtail.contrib.settings.models.register_settings.
2. Go to the Wagtail admin page for the settings model and the save button styling appears without right padding.

### Technical details

* Python version: 3.6.2
* Django version: 1.11.4
* Wagtail version: 1.12.1
* Browser version: Chrome 60.0.3112.113, Firefox Developer Edition 58.0b9

gasman:
Fixes #3836, effectively as a third attempt at fixing #3718...

* #3719 removed the `action` class from the preview button so that it wouldn't be given a 250px container, but this broke the dropdown version of the preview button (#3784) which _should_ be forced to 250px wide
* #3785 reinstated the `action` class, but moved the `width: 250px` rule so that it only applied to dropdowns; this broke the wagtailsettings save button (#3836), which isn't a dropdown but should be 250px
* this version reverts to the original fix from #3719, but adds a special case for `.preview .dropdown` to force it back to 250px.

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
@@ -138,12 +138,8 @@ footer {
     }
 
     .actions {
+        width: 250px;
         margin-right: $grid-gutter-width / 2;
-        width: auto;
-
-        .dropdown {
-            width: 250px;
-        }
     }
 
     .meta {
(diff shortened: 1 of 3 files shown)


gasman:
Merged in 2a9ec51995d2338cc5e8b37ec21aa8f8f31af6af (master) and 7cf38f115ca84bf687ba3bf42283e52ba6564157 (1.12.x)

Issue #3850, opened 2017-09-11
Preview mode works differently under the Django 1.10 middleware system


coredumperror:
This is very much the same bug as #3546, but I want to bring it up separately because I'm having a completely different problem. Every time one of my users previews a page with an image on it, the preview crashes.

The problem, as @gasman mentioned in #3546, is that Wagtail can no longer do the same thing it did with Django 1.9-style middleware. It would run the `process_request()` method for each middleware, but not `process_response()`. Under the Django 1.10 middleware mechanism, it cannot skip the `process_response()` calls, and that's what's breaking my site.

My site uses [CrequestMiddleware](https://github.com/Alir3z4/django-crequest) to store a globally accessible `request` object for the duration of each request. We use that request for a variety of purposes, including figuring out what path uploaded files should be saved to in our S3 bucket. We need to differentiate between files uploaded to different Sites, so that Site A can upload `file.jpg` without interfering with Site B's `file.jpg` that's already there.

For what I assume is necessary anti-memory-leak reasons, CrequestMiddleware destroys the request object during `process_response()`. This is a huge problem because it stores said request object in a dictionary keyed by the current thread object (the return value of `threading.current_thread()`). Since `Page.dummy_request()` invokes `BaseHandler.middleware_chain()` while Django's original call to middleware_chain() is still under way, this causes the stored request object to be overwritten (in the thread's second call to `CrequestMiddleware.process_request()` and then destroyed (in the thread's first call to `CrequestMiddleware.process_response()` before the preview is rendered.

This makes it impossible to retrieve the stored request object when rendering the image derivative, which throws an exception. Then the original middleware chain unwinds, and `CrequestMiddleware.process_response()` is called again, but it has nothing to delete, because the stored request is already gone.

The only solution I can think of for this would be to have Wagtail mark the dummy request object with some sort of flag (like it used to do with the `request.is_preview` attribute). I could then subclass `CrequestMiddleware` to alter `process_response()` to not delete the stored request when `request.is_preview == True`.

That feels pretty hacky, though. Especially since it makes a third-party middleware need to be aware of a Wagtail-specific flag. I'd love for there to be a better solution, but nothing is coming to mind. 

I'll go whip up a PR for that `is_preview` flag.

coredumperror:
I ended up going for `request.is_dummy` instead of `is_preview`, because `is_preview` is already in use in `Page.serve_preview` and fulfills a slightly different purpose.

coredumperror:
[...]

PR #3851, opened 2017-09-11
Added a flag in dummy_request() to inform middleware that the request is a dummy

coredumperror:
This is useful for middleware that perform destructive changes in
process_response(). They rightfully expect process_response() to only get called
once per thread, but it happens twice during a Page preview when running under
Django 1.10's middleware system. This flag lets the middleware know that they
should potentially skip their destructive operations.

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -1273,6 +1273,9 @@ class Page(six.with_metaclass(PageBase, AbstractPage, index.Indexed, Clusterable
 
         request = WSGIRequest(dummy_values)
 
+        # Add a flag to let middleware know that this is a dummy request.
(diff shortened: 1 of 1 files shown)


coredumperror:
This is the PR for #3850.

coredumperror:
I no longer have any use for this flag, as I discovered an alternative solution to the problem I wrote this PR for (which I discuss in #3850). However, this is such a small change, and it could potentially be useful to someone, so maybe it's worth keeping?

gasman:
Merged in c8b0fff58670b31b9876b0866617d317b54e4bc9 - thanks @coredumperror!

(Have left this change undocumented, as dummy requests are an internal implementation detail.)

PR #3853, opened 2017-09-13
Rebase / cleanup #3541

### LINKED ISSUE #3534: delete signal listeners should use on_commit

gasman:
Fixes #3534 and #3539


gasman:
Merged in 563b8e48ef80463ff40d2a98c0266622cd755a87

PR #3856, opened 2017-09-14
Fix routablepageurl to work with WAGTAIL_APPEND_SLASH = False

vstoykov:
Currently if you have set `WAGTAIL_APPEND_SLASH = False` in your settings `{% routablepageurl %}` will not generate correct urls. Instead of `/routable-slug/child-slug/` it generates `/routable-slugchild-slug/`

~~Probably is a good idea to add some tests for that but I need some assistance. I didn't found any tests for `routablepageurl`. Probably I'm missing something.~~

I added a test but I need feedback if the approach that I'm taking is ok.

### DIFF
--- a/wagtail/contrib/wagtailroutablepage/tests.py
+++ b/wagtail/contrib/wagtailroutablepage/tests.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, unicode_literals
 
+import mock
 from django.core.urlresolvers import NoReverseMatch
 from django.test import RequestFactory, TestCase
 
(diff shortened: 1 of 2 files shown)


wagtail/contrib/wagtailroutablepage/tests.py
```diff
@@ -185,3 +185,15 @@ def test_templatetag_reverse_external_view(self):
                               'external_view', 'joe-bloggs')
 
         self.assertEqual(url, self.routable_page.url + 'external/joe-bloggs/')
+
+    def test_templatetag_reverse_external_view_without_append_slash(self):
+        import wagtail.wagtailcore.models
+        wagtail.wagtailcore.models.WAGTAIL_APPEND_SLASH = False
```
kaedroho: We should use ``mock.patch()`` for this. Otherwise, if the ``routablepageurl`` call raises an exception, ``WAGTAIL_APPEND_SLASH`` will not be restored to ``True`` which could lead to confusing test breakages later on.

vstoykov: I updated my PR to use `mock`

kaedroho:
Thanks!

PR #3869, opened 2017-09-19
Remove SessionAuthenticationMiddleware from project template and docs.

solarissmoke:
The project template assumes Django 1.11, where SessionAuthenticationMiddleware is redundant (see https://docs.djangoproject.com/en/1.11/releases/1.10/#features-removed-in-1-10). This patch removes it from the template.

Also updated the documentation to consistently refer to `MIDDLEWARE` instead of `MIDDLEWARE_CLASSES`. (There is a question about whether, in all these places, we need to remind people to use `MIDDLEWARE_CLASSES` if they are on Django 1.8 - please let me know if you think this is necessary).


gasman:
Thanks @solarissmoke! Merged in 62b8caedd5324deecf5d14aead86798780185dc5.

PR #3871, opened 2017-09-19
'The Zen of Wagtail' documentation

gasman:
A high level document setting out best practices for using (and developing) Wagtail. Feedback welcome!


nfletton:
Seems excellent to me. I would suggest ending the third paragraph '- expect to write HTML **& code**' just to set expectations early.

docs/getting_started/the_zen_of_wagtail.rst
```diff
@@ -0,0 +1,37 @@
+The Zen of Wagtail
+==================
+
+Wagtail has been born out of many years of experience building websites, learning approaches that work and ones that don't, and striking a balance between between power and simplicity, structure and flexibility. We hope you'll find that Wagtail is in that sweet spot. However, as a piece of software, Wagtail can only take that mission so far - it's now up to you to create a site that's beautiful and a joy to work with. So, while it's tempting to rush ahead and start building, it's worth taking a moment to understand the design principles that Wagtail is built on.
```
timorieber: Duplicated "between".

gasman:
Thanks @nfletton and @timorieber! Updated now. (Have changed it to "expect to write code" because I wouldn't want to imply that HTML _isn't_ code...)

gasman:
Merged in 2c0177234ee37769661eca2a41f94c98c5ad3e02 (with the page moved to the end of the 'getting started' section)

geoffroy-noel-ddh:
I find it interesting to compare the Zen of Wagtail with Django Philosophy and wonder if it would be useful to warn new users about the potential conflicts between the two cohabiting systems?

Wagtail provides a very nice API and admin interface to define custom Pages and edit them. It is quite convenient to make all the changes in the Model classes: field indexes, admin interface, routing and serving http requests.

However that highly integrated approach seems at odd with some of the MVC principle and core Django philosophies. https://docs.djangoproject.com/en/2.2/misc/design-philosophies/ particularly the idea of **Loose Coupling** (in Wagtail controllers, admin and models, urls are, by default, all coded in the same class).

This is particularly true of the **Streamfield**. Again this is a great feature and a huge improvement compared to Django admin interface, in terms of usability and compared to a wysiwyg editor in terms of control of the structure of the pages. However I've noticed that it tends to be easily abused by developers who, due to its convenience, will think first in terms of specific html output and then admin interface (what should I show on this web page, how can I edit this content?) and will stuff a Page model with all sorts of composite elements (#4473). On the other hand, the concept of Model is central to Django and many layers are based on top of that. Non-wagtail developers tend to think first about their model and its normalisation then how the views and admin can be built on top of that which is a more sustainable way to produce applications because the content and its schema can be re-purposed for different channels. I think one key issue with Streamfield is that the page content and its structure are buried in a **proprietary json document** within a relational field in your Page model. It's ok for simple pages but not so great for complex or reusable entities (e.g. speaker profiles).

Django's core principles such as **"Less code" and "Low development"** are also quite different from Wagtail. Django for instance generates a very decent admin interface and query API for you automatically from your models. Wagtail on the other hand is probably the most code-heavy CMS I've ever used. It doesn't even come with an optional reusable Page model, very rudimentary components found in almost any CMS such as a Wysiwyg Page, Index Page, Menu management, ... have to be hand coded. Each time underlying Wagtail code changes those hand-coded pieces need to be reviewed or maintained. Third party implementations are available but which is one is truly extensible, well designed, modular and comes with long term support?

Hope this is taken as constructive criticisms rather than a rant! As I said there are many excellent ideas in Wagtail, but the philosophical discrepancies described above seem to often resurface in the projects I work on.

PR #3873, opened 2017-09-21
Permission to move a Page should not be affected by bulk_delete.

coredumperror:
The `PagePermissionTester.can_move()` method simply calls `PagePermissionTester.can_delete()`. Before the bulk_delete permission was added, this was a reasonable analogue for "permission to delete", since a move is effectively a "cut and paste" operation, and you can't "cut" without permission to delete.

But bulk_delete is changes this dynamic. It was added to avoid allowing users to delete a whole lot of pages all at once by mistake, when they think they're just deleting one. However, the move/delete analogue breaks down when you're talking about moving an entire subtree, because you're *not* directly altering the children of the moved Page. They're just coming along for the ride.

This PR makes the can_move() method ignore the bulk_delete permission.

I'm not sure if this breaks any tests. This interaction between move and bulk_delete *seems* like it's unintentional, which would imply that no tests test for this. But if the tests do fail, that probably means this is intended behavior.

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -1712,7 +1712,7 @@ class PagePermissionTester(object):
             ('add' in self.permissions and self.page.owner_id == self.user.pk)
         )
 
-    def can_delete(self):
+    def can_delete(self, ignore_bulk=False):
         if not self.user.is_active:
             return False
         if self.page_is_root:  # root node is not a page and can never be deleted, even by superusers
@@ -1723,7 +1723,7 @@ class PagePermissionTester(object):
             return True
 
         # if the user does not have bulk_delete permission, they may only delete leaf pages
-        if 'bulk_delete' not in self.permissions and not self.page.is_leaf():
+        if 'bulk_delete' not in self.permissions and not self.page.is_leaf() and not ignore_bulk:
             return False
 
         if 'edit' in self.permissions:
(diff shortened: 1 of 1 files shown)


coredumperror:
It doesn't look like the failures in the Travis CI build are the fault of my changes. There are dozens and dozens of failed tests which complain about things that aren't even remotely related to permissions. And only on 4 out of 20 environments.

gasman:
Thanks @coredumperror! Those tests should be fixed by #3875 now - rebasing this PR on top of latest master should fix things.

coredumperror:
OK, I think I just did that. Hopefully the test results will be accurate this time.

coredumperror:
Ah shoot, this got lost in the shuffle. What needs to be done at this point to resurrect this PR and hopefully get it into Wagtail 2.5?

gasman:
@coredumperror It's missed the boat for 2.5 I'm afraid... the release candidate came out on Monday, so it'll only be receiving fixes for critical issues and regressions from 2.4.

It looks like this just needs a test in https://github.com/wagtail/wagtail/blob/d8b2d087ca12840013e72d1ab1e8edc62518b5c2/wagtail/admin/tests/test_pages_views.py#L2573 to verify that a user without bulk delete permission is able to move a page with a child page.

coredumperror:
Darn. Especially since the resultant lack of urgency (we've got a perfectly functional monkey patch that's been doing this for us since I posted this PR) means I probably won't have the cycles to write that test any time soon. But now I at least know what needs to be done. I can make a ticket in our Trello, and maybe eventually get back to this. heh

lb-:
@coredumperror  - tested locally (before and after) and works great, `bulk_delete` permission no longer required. I have added some tests and will merge this in shortly.

Thanks for the work getting this one started.

lb-:
Merged in via 982b1d60a4b3d5b4e841f2b39e2244bcb421b091

PR #3876, opened 2017-09-22
Stop breadcrumb home icon from showing as ellipsis in Chrome

gasman:
before:
screen shot 2017-09-22 at 12 29 20

after:
screen shot 2017-09-22 at 12 29 56


jjanssen:
Merged with c19bea6

PR #3878, opened 2017-09-25
Use globe instead of home icon for breadcrumb root

gasman:
One of our designers commented that it was unintuitive that the 'home' icon in the breadcrumb takes you to the root level, rather than the homepage. This is particularly true of single-site installations, where the usefulness of accessing the root level at all is unclear.

I don't think we can (or should) omit the root from the breadcrumb entirely (since you need to go there to set up a second site). I believe the real problem is that we're misusing the 'home' icon to represent the root - editors would intuitively head for the _second_ link in the breadcrumb (which will generally be labelled 'Home' or with the name of the site) when they want to go to the homepage, if it weren't for the fact that there was a big friendly 'Home' icon sending them astray. So, I think this can be fixed by changing it to an icon that doesn't have that mental association.

I propose using the globe icon instead - this will hopefully be a meaningful concept to people managing multi-site Wagtail installations, and for people who aren't, it could equally represent "the internet".

screen shot 2017-09-25 at 18 12 19

Note that with this change in place, limited-permission users who don't have access to the root level will still get the 'home' icon as the first link in the breadcrumb, as before (even if their highest accessible level isn't actually a homepage, e.g. an editor who only has access to the blog section):

screen shot 2017-09-25 at 18 31 06

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/core.scss
@@ -246,6 +246,7 @@ footer {
                 text-align: center;
                 width: 3em;
                 font-size: 1em;
+                text-overflow: clip;
 
                 &:before {
                     font-size: 1.15rem;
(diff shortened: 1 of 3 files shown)


gasman:
Merged in 5d39033df0d72c82ebd0e846bb6165e2f5a2e9ad

Issue #3852, opened 2017-09-13
Specifying rich text features does not whitelist only these features


BertrandBordage:
Starting Wagtail 1.12, we can specify the features allowed in a rich text field, for example just bold and italic.

Unfortunately, there is no server side validation for that. I could work around that by overriding the `clean` method of a given model and by defining a subclass of `DbWhitelister` that contains a subset of the `element_rules` dictionary.

So if you copy/paste content from another field or document, it keeps links etc even if you didn’t allow that feature.

A solution for that could be to define a dictionary linking `features` to `element_rules`. Then in the `clean` method of `RichTextField` or block, we could define a new instance of `DbWhitelister` that gets the class attribute `element_rules` overridden by a homonym instance class containing a subset of `element_rules`.

gasman:
Bumping to 1.14. I've made some progress on this at https://github.com/gasman/wagtail/tree/feature/feature-based-whitelisting, but dealing with embed and link handlers is going to need careful planning, so it would be a bad idea to try and rush this through for 1.13.

Currently those link/embed handler objects (PageLinkHandler etc) perform three tasks:

1. Converting the pseudo-HTML from the database into real HTML for front-end rendering
2. Converting the pseudo-HTML from the database into HTML that the rich text editor can use (with page IDs stashed in `data-id` attributes, etc)
3. Converting the output of the rich text editor to database-pseudo-HTML.

It seems to make a lot of sense to decouple task 1 from the other two, since:

* When we introduce draft.js or any other rich text editor that has a non-HTML native data format, we'll be swapping in another component to perform tasks 2 and 3 (converting between that data format and database-pseudo-HTML) while leaving task 1 unchanged
* Task 1 can't be made aware of rich text feature flags - at least, not without some major architectural changes - because we have to do it in contexts like the `|richtext` filter, where we only have access to the string itself, and not the model field (which would tell us which features are active)
* It neatly corresponds to the dividing line between wagtailcore (task 1) and wagtailadmin (tasks 2/3)

Currently this whole API is undocumented (#1094), which is good news for us as it means we have a lot of leeway for refactoring it without having to worry about backwards compatibility. However, once this is integrated with rich text features, it really _does_ need to get documented, which makes it all the more important to put in the extra effort now to get the API as polished as it can be.

One thing we almost certainly want to do is fix the hack where the handlers receive attributes as HTML-escaped strings. This means that my latest bugfixes #3879 and #3882 become invalid, but that's a necessary price to pay :-)

[...]

PR #3879, opened 2017-09-27
Preserve double quotes in alt text of images in rich text

gasman:
Supersedes #2669. Previously, alt text was cut off at the first double quote, due to a combination of several HTML-escaping bugs.

1. When converting the database representation of an image embed (`<embed embedtype="image" id="123" format="left" alt="Arthur &quot;two sheds&quot; Jackson"/>`) to real HTML, we intentionally pass ImageEmbedHandler the attributes in HTML-escaped form (because we're extracting them with regexp hacks rather than a proper HTML parser) and thus don't do any further escaping during ImageEmbedHandler or the image format object. In order to follow that contract, the image chooser view needs to escape the supplied alt text when it invokes the image format object (to generate the snippet of HTML for the rich text field). It didn't.
2. The final output of the HTML whitelister comes from BeautifulSoup, which by default is a bit too clever when handling attributes with double quotes in them - it wraps them in single quotes to avoid escaping, resulting in `<foo alt='Arthur "two sheds" Jackson'>`. This means our regexps don't find that attribute (and if they did, they'd end up with the un-escaped version and fall foul of point 1 above). To fix this, we tell BeautifulSoup to use `django.utils.html.escape` for its escaping, which _does_ consistently escape double-quotes.
3. If one of these rogue single-quoted attributes somehow snuck into the database-stored HTML (which shouldn't have been possible due to bug number 1, but you never know), ImageEmbedHandler would crash out when expanding the HTML, due to not being able to find the `alt` attribute (=> breaking page rendering). This has now been fixed to just render an empty `alt` attribute.

### DIFF
--- a/wagtail/wagtailcore/tests/test_whitelist.py
+++ b/wagtail/wagtailcore/tests/test_whitelist.py
@@ -145,3 +145,8 @@ class TestWhitelister(TestCase):
         string = '<b>snowman Yorkshire<!--[if gte mso 10]>MS word junk<![endif]--></b>'
         cleaned_string = Whitelister.clean(string)
         self.assertEqual(cleaned_string, '<b>snowman Yorkshire</b>')
+
+    def test_quoting(self):
+        string = '<img alt="Arthur &quot;two sheds&quot; Jackson" sheds="2">'
+        cleaned_string = Whitelister.clean(string)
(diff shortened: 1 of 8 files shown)


gasman:
Scratch this... as per https://github.com/wagtail/wagtail/issues/3852#issuecomment-333155647, I now believe that fixing the EmbedHandler API so that it receives an un-escaped version of the attribute is the right thing to do here.

gasman:
Superseded by #4060

PR #3881, opened 2017-09-28
Make last_published_at filterable

m1kola:
This PR makes `Page.last_published_at` filterable along with `first_published_at` and other basic fields from the `Page` model.


gasman:
Merged in ddb0665f79fe7e6cea849230cfc6f391278bc358 - thanks @m1kola!

PR #3882, opened 2017-09-28
 Prevent ampersands in embed URLs in rich text from being escaped repeatedly

### LINKED ISSUE #3126: Embedded video urls get doubled escaped on save actions

gasman:
Fixes #3126

Incorporates #3879, since we're assuming that the characters to be un-escaped should match the ones handled by django.utils.html.escape, and this is ensured by c4248e3e6485b16ed4929d69f9a83f1195d02bf4.

### DIFF
--- a/wagtail/wagtailcore/tests/test_whitelist.py
+++ b/wagtail/wagtailcore/tests/test_whitelist.py
@@ -145,3 +145,8 @@ class TestWhitelister(TestCase):
         string = '<b>snowman Yorkshire<!--[if gte mso 10]>MS word junk<![endif]--></b>'
         cleaned_string = Whitelister.clean(string)
         self.assertEqual(cleaned_string, '<b>snowman Yorkshire</b>')
+
+    def test_quoting(self):
+        string = '<img alt="Arthur &quot;two sheds&quot; Jackson" sheds="2">'
+        cleaned_string = Whitelister.clean(string)
+        self.assertEqual(cleaned_string, '<img alt="Arthur &quot;two sheds&quot; Jackson"/>')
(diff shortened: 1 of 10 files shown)


gasman:
Scratch this... as per https://github.com/wagtail/wagtail/issues/3852#issuecomment-333155647, I now believe that fixing the EmbedHandler API so that it receives an un-escaped version of the attribute is the right thing to do here.

gasman:
Superseded by #4060

PR #3895, opened 2017-10-02
Use Django ORM query in Page._update_descendant_url_paths.

atombrella:
I'm in doubt about MSSQL, but for the other backends, it can be expressed in Django.

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -14,8 +14,9 @@ from django.core.exceptions import ValidationError
 from django.core.handlers.base import BaseHandler
 from django.core.handlers.wsgi import WSGIRequest
 from django.core.urlresolvers import reverse
-from django.db import connection, models, transaction
-from django.db.models import Q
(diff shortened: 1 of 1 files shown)


atombrella:
In the README-file, there's no word about MSSQL. 

> Wagtail supports Django 1.8.x, 1.10.x and 1.11.x, on Python 2.7, 3.4, 3.5 and 3.6. Supported database backends are PostgreSQL, MySQL and SQLite.

gasman:
Thanks @atombrella! There's actually a very similar PR already in the queue for reviewing: #3496, currently blocked by our lack of testing environment for MSSQL. (We don't officially support MSSQL at the moment, but we're working towards doing so, and there are several existing sites running on that setup - so we need to be careful not to break those.)

atombrella:
An alternative could be to keep the proposed change, backed by the mssql check?

gasman:
Addressed via #3496, with an exception added for MSSQL as you suggest. Thanks again!

PR #3896, opened 2017-10-03
FIX : Update required dependency

Onuonga:
* I have added two dependancies:
 *django
 *django-core

* This is due to the fact that after installing wagtail only and trying to start the project with `wagtail start mysite` you will get the following error:
```
File "/home/vagrant/wagdev/ve/bin/wagtail", line 7, in <module>
from wagtail.bin.wagtail import main
File "/home/vagrant/wagdev/ve/local/lib/python2.7/site-packages/wagtail/bin/wagtail.py", line 7, in <module>
from django.core.management import ManagementUtility
ImportError: No module named django.core.management
```

django-core on its own does not solve this and thus why django has been included here.
It may be better to have this in a dependency file but I though this may be a good first step.

### DIFF
--- a/docs/getting_started/index.rst
+++ b/docs/getting_started/index.rst
@@ -18,7 +18,7 @@ With the above installed, the quickest way to install Wagtail is:
 
 .. code-block:: console
 
-    $ pip install wagtail
+    $ pip install wagtail django django-core
 
 (``sudo`` may be required if installing system-wide or without virtualenv)
 


gasman:
Hi @Onuonga - thanks for the contribution!

Django is [listed as a dependency of Wagtail](https://github.com/wagtail/wagtail/blob/aa3f5b782812d83f03f855786f80fcfc73c90e14/setup.py#L24), so it should get installed automatically when you run `pip install wagtail` - if that's not happening, something is going wrong with your installation of pip. (Also, the [django-core](https://pypi.python.org/pypi/django-core) package is a third-party add-on for Django which is not related to the `django.core` module, and isn't required for Wagtail.)

Onuonga:
Hey ,
Thank you for the speedy review.  

I was on the latest pip version. 

It does not come as a dependency. 
Also , note that a step or two after pip install wagtail you ask users to:
 ```pip install -r requirements.txt```

Looking at that file it has : 

```
Django>=1.8,<1.9
django-modelcluster>=3.0,<4.0
django-taggit>=0.20,<1.0
django-treebeard==3.0
six>=1.7.0
```
which means it is not listed. 
It may be something very specific to my setup but seeing as there are all these breadcrumbs I do not think it is. 
You will not get to that step because the mystic setup stage will fail. 

I can send screen shots to walk through the steps as I did if you would like to reproduce it on your end. 

Again, thank you for your time.

gasman:
It seems you're looking at the wrong requirements.txt - the one at https://github.com/wagtail/wagtail/blob/stable/1.12.x/docs/requirements.txt is only used for building the documentation. The one used on newly-created projects is https://github.com/wagtail/wagtail/blob/stable/1.12.x/wagtail/project_template/requirements.txt.

I'd be happy to look at the screenshots to see where it's going wrong. Please can you also show the output of `pip freeze` after running `pip install wagtail`?

PR #3902, opened 2017-10-04
Deleted Django 1.3 import fallback in utils/sendfile.py

atombrella:


gasman:
Wow, that's quite a historical artefact :-) Well spotted - thanks very much!

PR #3903, opened 2017-10-05
Simplified Content-Encoding in response in utils/sendfile.py

atombrella:
The coverage report hints that `if encoding` is always true. No validation is performed on the content-type, so `encoding or guessed_encoding` is what it performed already (at least that's how I read the code).


gasman:
Looks good to me - thanks @atombrella!

PR #3906, opened 2017-10-05
Use django.utils.six.iteritems in wagtail.utils.utils.deep_update.

atombrella:


gasman:
Thanks again! (I'm pretty sure we've just used `items()` in other places in the code - but this is all going to get updated fairly soon anyhow, when we add support for Django 2.0 and drop Python 2 entirely...)

PR #3907, opened 2017-10-05
Allow navigation from page usage/search results

### LINKED ISSUE #952: Explore child pages button missing from search
A user, after searching for a page with children, needs the option to click directly to the listing of children from the search result item. 

Currently, the only option is to "edit" the page found or to click to the parent's listing.

gasman:
Remove the allow_navigation flag from the 'explore' page listing, and have navigation always enabled - there's no good reason to leave it out, and people have asked for it. Fixes #952


gasman:
Merged in e93b21036e7425531edb49bc2526f47afc9c57e1

Issue #3713, opened 2017-07-17
Issue with Pillow >= 4.2.0 - IOError cannot write mode RGBA as JPEG


rafaponieman:
Pillow 4.2.0 (released early July) deprecated removing automatically the alpha channel on PNGs when saving them as JPEG. This won't be a problem if you're not specifying jpeg output on the image template tag (as PNGs output as PNGs), but if you do, it fails while trying to save a jpeg thumbnail for a PNG.

Solved temporarily by hardcoding Pillow 4.1.1 on my requirements.

Running Wagtail 1.11.1 on Django 1.10.7

rafaponieman:
Thank you!! Should I close this issue?

kaedroho:
This issue should automatically close itself as soon as #3911 is merged

PR #3911, opened 2017-10-06
Implement bgcolor image operation

### LINKED ISSUE #2259: Add filter spec option for background color
Cleaned up photos are often extracted from a background. For example product beauty shots.  It would be nice if these images could be uploaded to wagtail as transparent png files and then templates could specify the background matte to apply to the image for the rendition.  This way the images appear to be transparent but jpeg compression can be used.  Right now one must either use transparent PNG and accept the large file size or apply the jpeg background matte before uploading the image.  If you go with the jpeg option when the page css changes all images on the site have to be replace and reuploaded which is a time consuming and difficult process (also error prone when the images can be displayed on different background colors)

Would it be feasible to add background matte into the filter spec options?  And perhaps rendition format?  So jpeg could be forced using a specific background color?

kaedroho:
Fixes #2259 and #3713

This implements a new filter called ``bgcolor`` which is useful for people who want to force all images to use JPEG format but can't stop their users from uploading PNG files.

Depends on https://github.com/wagtail/Willow/pull/61

### DIFF
--- a/docs/topics/images.rst
+++ b/docs/topics/images.rst
@@ -224,6 +224,25 @@ For example, to make the tag always convert the image to a JPEG, use ``format-jp
 
 You may also use ``format-png`` or ``format-gif``.
 
+.. _image_background_color
+
+Background color
+----------------
+
+The PNG and GIF image formats both support transparency, but if you want to
+convert images to JPEG format, the transparency will need to be replaced with a
+solid background color.
+
+By default, Wagtail will set the background to white. But if a white background
+doesn't fit your design, you can specify a color using the ``bgcolor`` filter.
+
+This filter takes a single argument, which is a CSS 3 or 6 digit hex code
+representing the color you would like to use:
+
+.. code-block:: html+Django
+
+    {# Set the image backgrounds to black #}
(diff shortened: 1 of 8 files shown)


kaedroho:
Willow 1.1 is released now: https://pypi.python.org/pypi/Willow/1.1

kaedroho:
Rebased

docs/topics/images.rst
```diff
@@ -224,6 +224,25 @@ For example, to make the tag always convert the image to a JPEG, use ``format-jp
...
+
+By default, Wagtail will set the background to white. But if a white background
+doesn't fit your design, you can specify a color using the ``bgcolor`` filter.
+
+This filter takes a single argument, which is a CSS 3 or 6 digit hex code
+representing the color you would like to use:
+
+.. code-block:: html+Django
+
+    {# Set the image backgrounds to black #}
```
tomdyson: backgrounds -> background ?

BertrandBordage:
Merged as 1e332b4b67a7504bb6de4d5c809f3a687c090964.

PR #3914, opened 2017-10-07
Add support for min/max for list block

### LINKED ISSUE #2379: Minimum length validation ListBLock

Adrian-Turjak:
* adds min_num and max_num params to list block
* adds tests for new limiters
* adds docs
* add javascript for realtime validation of limiters
* update sequence javascript to expose needed functions

closes #2379 
is a replacement for #3150

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/blocks/list.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/blocks/list.js
@@ -7,6 +7,9 @@
         var listMemberTemplate = $('#' + opts.definitionPrefix + '-newmember').text();
 
         return function(elementPrefix) {
+            var errorText = $('#' + elementPrefix + '-errors');
+            var addButton = $('#' + elementPrefix + '-add');
+
             var sequence = Sequence({
                 prefix: elementPrefix,
                 onInitializeMember: function(sequenceMember) {
@@ -47,10 +50,47 @@
                 }
             });
 
-            /* initialize 'add' button */
-            $('#' + elementPrefix + '-add').click(function() {
+            var addMember = function() {
                 sequence.insertMemberAtEnd(listMemberTemplate);
-            });
+            }
+
+            /* initialize 'add' button */
+            addButton.click(addMember);
+
+            function checkMinMax(){
+              if (opts.max_num && sequence.getRealCount() >= opts.max_num) {
+                  addButton.addClass('disabled');
+                  addButton.unbind('click', addMember)
+              } else {
+                  addButton.removeClass('disabled');
+                  /* unbind first to ensure we aren't binding multiple times */
+                  addButton.unbind('click', addMember)
+                  addButton.click(addMember);
+              }
+
+              if (opts.min_num && sequence.getRealCount() < opts.min_num){
+                if (!$("#" +  elementPrefix + "-min-error").length) {
+                  errorText.append(
+                    "<div id=\"" + elementPrefix + "-min-error\" class=\"help-block help-critical\">" + opts.min_num + " or more items required</div>");
+                }
+              } else {
+                $("#" +  elementPrefix + "-min-error").remove();
+              }
+            }
+
+            checkMinMax(sequence);
+
+            sequence.__postInsertMember = sequence._postInsertMember;
+            sequence._postInsertMember = function(newMember){
+              sequence.__postInsertMember(newMember);
+              checkMinMax();
+            }
+
+            sequence._deleteMember = sequence.deleteMember;
+            sequence.deleteMember = function(member){
+              sequence._deleteMember(member);
+              checkMinMax();
+            }
         };
     };
 })(jQuery);
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/blocks/sequence.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/blocks/sequence.js
@@ -79,6 +79,10 @@ CODE FOR SETTING UP SPECIFIC UI WIDGETS, SUCH AS DELETE BUTTONS OR MENUS, DOES N
             return parseInt(countField.val(), 10);
         };
 
+        self.getRealCount = function() {
+            return members.length;
+        };
+
         function getNewMemberPrefix() {
             /* Update the counter and use it to create a prefix for the new list member */
             var newIndex = self.getCount();
@@ -89,7 +93,7 @@ CODE FOR SETTING UP SPECIFIC UI WIDGETS, SUCH AS DELETE BUTTONS OR MENUS, DOES N
         function setInitialMoveUpDownState(newMember) {
         }
 
-        function postInsertMember(newMember) {
+        self._postInsertMember = function(newMember) {
             /* run any supplied initializer functions */
             if (opts.onInitializeMember) {
                 opts.onInitializeMember(newMember);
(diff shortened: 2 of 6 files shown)


Adrian-Turjak:
@robmoorman If you can also test and confirm this works for you. :)

@gasman I hope this addresses your comments on the other review. Also, the changes I've done for the js validation mean we can do something similar for the max_num and min_num limiters that were added for streamfields, that way we have the same validation in realtime rather than just on post.

wagtail/wagtailadmin/static_src/wagtailadmin/js/blocks/sequence.js
```diff
@@ -79,6 +79,10 @@ CODE FOR SETTING UP SPECIFIC UI WIDGETS, SUCH AS DELETE BUTTONS OR MENUS, DOES N
             return parseInt(countField.val(), 10);
         };
 
+        self.getRealCount = function() {
```
gasman: Let's come up with a better name for this. The existence of a function called `getRealCount` just tells me "there's something subtly wrong with the result of `getCount`, but we're not going to tell you what it is... you have to read the implementation to find out".

If you think that `getCount` is badly named (and it probably is!) then we should fix it, not just build on top of that mistake. Perhaps `getCount` should become `getCountIncludingDeleted` or `getCurrentCountIndex`, so that the new function can be `getCount`?

wagtail/wagtailadmin/static_src/wagtailadmin/js/blocks/sequence.js
```diff
@@ -89,7 +93,7 @@ CODE FOR SETTING UP SPECIFIC UI WIDGETS, SUCH AS DELETE BUTTONS OR MENUS, DOES N
         function setInitialMoveUpDownState(newMember) {
         }
 
-        function postInsertMember(newMember) {
+        self._postInsertMember = function(newMember) {
```
gasman: What's the significance of the leading underscore here? I'd assume it was meant to indicate that this is an internal-only function rather than part of the `Sequence` API, but it seems that we're going ahead and accessing it within `ListBlock` anyway.

wagtail/wagtailadmin/static_src/wagtailadmin/js/blocks/sequence.js
Adrian-Turjak: I fully agree, I just didn't want to break anything by changing getCount, but if you're find with me doing a commit to rename the function first (I'll rebase this work onto that), that works better. Just need to settle on a name.

wagtail/wagtailadmin/static_src/wagtailadmin/js/blocks/sequence.js
Adrian-Turjak: Yes, mostly to signify that it's not a function you should call yourself. I'm not actually even calling it, just replacing it with a decorator like function. Which is why I need it exposed on the sequence object itself.

It's a weird way of doing it, but I couldn't think of a nicer way to do it given javascript's scoping mechanisms, but someone who is more js savvy may know a better method.

wagtail/wagtailadmin/static_src/wagtailadmin/js/blocks/list.js
```diff
@@ -47,10 +50,47 @@
...
+              }
+            }
+
+            checkMinMax(sequence);
+
+            sequence.__postInsertMember = sequence._postInsertMember;
+            sequence._postInsertMember = function(newMember){
+              sequence.__postInsertMember(newMember);
+              checkMinMax();
+            }
```
Adrian-Turjak: @gasman Here I'm only replacing the _postInsertMember with my own decorated version. So I need it exposed, but I still feel the original '_' helps to make it clear that this is not a function you call yourself, but rather it is internal to the sequence. Although in this case the double underscore looks a little weird, but it's clear.

I could potentially do:
```
orginal_postInsertMember = sequence._postInsertMember;
sequence._postInsertMember = function(newMember){
  orginal_postInsertMember(newMember);
  checkMinMax();
}
```

wagtail/wagtailadmin/static_src/wagtailadmin/js/blocks/sequence.js
Adrian-Turjak: Another option... which I'm less inclined towards is to expose members itself as self._members

I can then get the length from there, but a function for count is much more useful here, and elsewhere.

gasman:
Cross-referencing the new code against how the existing code for enabling/disabling the up/down buttons work, it feels like this is breaking the separation of concerns between sequence.js and list.js:

* sequence.js handles the model-level details of managing a sequence of blocks (including determining when specific actions such as reordering, adding and removing is valid), as well as updating the hidden fields;
* list.js (and stream.js) handles the view-level details, such as enabling/disabling buttons (and determining which buttons exist in the first place);
* Any communication from sequence.js to list.js (e.g. telling it to disable the 'move up' action because a block can no longer be moved up) is done through callback functions like `onDisableMoveUp`.

On that basis, the correct way to implement this would be:

* `Sequence` is initialised with `min_num` and `max_num` passed as options, and is responsible for deciding when add/delete actions become valid or invalid. This means there's no need to expose `getRealCount` (because the comparison between `members.length` and `min_num` / `max_num` happens entirely within `Sequence`).
* We add some new callback functions to `Sequence`: `onEnableDelete` / `onDisableDelete` and `onEnableInsert` / `onDisableInsert`, which are triggered when we reach the minimum/maximum number of items
* `ListBlock` responds to those callback functions by enabling/disabling buttons (or the warning message) as appropriate. This means there's no need to expose `_postInsertMember`.

Adrian-Turjak:
@gasman That sounds much more sensible. I'm sadly unlikely to get a chance to redo the pull request with that work right away. Will get back to it mid November.

Adrian-Turjak:
I'm so sorry, I've been swamped with other stuff and totally forgot about this pull request.

Will rebase onto master and work on getting it done soonish.

thibaudcolas:
It looks like this PR will be superseded by #4942, which also adds support for this? (haven't tried myself).

- https://github.com/wagtail/wagtail/issues/4473#issuecomment-389708995
- https://github.com/wagtail/wagtail/issues/4473#issuecomment-390795973

wagtail/wagtailcore/blocks/list_block.py
```diff
@@ -7,18 +7,29 @@
...
+    def __init__(self, block_errors=None, non_block_errors=None):
+        self.block_errors = block_errors
+        self.non_block_errors = non_block_errors
+        super(ListBlockValidationError, self).__init__('Validation error in ListBlock')
 
 
 class ListBlock(Block):
 
-    def __init__(self, child_block, **kwargs):
+    def __init__(self, child_block, max_num=None, min_num=None, **kwargs):
```
CuriousLearner: Just a minor suggestion.

We can rename these to `min_length` and `max_length` to be consistent with what various fields in Django uses.

gasman: @CuriousLearner That'll make them inconsistent with Wagtail though... http://docs.wagtail.io/en/v2.7/topics/streamfield.html#streamblock

CuriousLearner: Ah, I see. Fair enough.

Thank you for pointing me to that :)

gasman:
Superseded by #7334

Issue #3915, opened 2017-10-09
Unicode slugs are not properly slugified due to javascript limitations


saevarom:
When using unicode slugs, and the slug contains a unicode character and a word from the removelist in urlify.js right after it (e.g. the letter "a"), the unicode character is detected as a word boundary in javascript and the text that matches the removelist is stripped from the slug...
The removelist contains the following words:

    var removelist = [
      "a", "an", "as", "at", "before", "but", "by", "for", "from", "is",
      "in", "into", "like", "of", "off", "on", "onto", "per", "since",
      "than", "the", "this", "that", "to", "up", "via", "with"
    ];

And when I try to slugify this text: "Kaupa miða"
The result becomes: "kaupa-mið"

This is also an issue in Django since the Urlify.js used in wagtail comes from the Django source. I've already submitted a bug report to the Django project here: [https://code.djangoproject.com/ticket/28688](https://code.djangoproject.com/ticket/28688) 

### Steps to Reproduce

1. Create a new page in wagtail, or edit an existing one
2. Open a development console in the browser (e.g. inspect in chrome)
3. Paste the following line into the console and hit enter: `URLify("Kaupa miða", 255, true)`. The output is incorrectly slugified as `kaupa-mið`

### Technical details

* Python version: Irrelevant
* Django version: 1.11.5
* Wagtail version: 1.13.0 alpha
* Browser version: Any browser

PR #3916, opened 2017-10-09
Bypass removelist in urlify when dealing with unicode characters.

saevarom:
I want to suggest that we simply bypass the removelist in urlify when the string contains unicode characters. This solves #3915.


saevarom:
This same fix has been approved in Django.
https://code.djangoproject.com/ticket/28688

gasman:
Thanks @saevarom! Merged in 9ac07d0a756a072ca515583bea24f5ef596488ce.

Issue #3917, opened 2017-10-09
Django 2.0 support


gasman:
Top-level issue to track progress towards supporting Django 2.0 in Wagtail 2.0 - further tickets may be opened and linked here to track sub-tasks.

Plan (items under the same heading can be done in parallel):

## Old version removal

- [x] Remove Django 1.8, Django 1.10 and Python 2.7 from CI
- [x] Ensure that the demo projects we intend to develop against (bakerydemo, and wagtaildemo if we haven't yet switched vagrant-wagtail-develop to bakerydemo) are running against Python 3 and Django 1.11

## Cleanup

- [x] Seek out and destroy all fallback code handling:
  - [x] Django 1.8 / 1.10 (#3924, #3932)
  - [x] Python 2.7 (#3934)
- [x] Fix all deprecation warnings in test suite when running against Django 1.11 (NB may overlap with the above: some warnings can't effectively be fixed while <1.11 compatibility code remains in place)
- [ ] Optional: Replace use of django.utils.six with Python3-specific equivalent code
- [ ] Optional: Eliminate use of `django.utils.deprecation.MiddlewareMixin` ([docs](https://docs.djangoproject.com/en/1.11/topics/http/middleware/#upgrading-pre-django-1-10-style-middleware))

## Upgrade

- [x] Check release notes at https://docs.djangoproject.com/en/dev/releases/2.0/ and make any necessary changes
- [x] Run tests against the latest pre-release of Django 2.0 and fix anything that breaks (#4085)

## Finalisation

- [x] Update dependencies in setup.py (#4085)
- [x] Add Django 2.0 to CI (#4085)
- [x] Update setup.py and tox/travis config to reference Django 2.0-compatible releases of django-modelcluster, django-treebeard and django-taggit, when available

Since this is likely to be a large and disruptive change to the codebase, we should aim to keep individual PRs as small and focused as possible, and review them promptly.

kaedroho:
Django have also dropped support for Python 3.4: https://github.com/django/django/commit/cfff2af02be40106d4759cc6f8bfa476ce82421c

FlipperPA:
I believe Python 3.4 support isn't being dropped until Django 2.1. There was a fairly big discussion on it here:

https://groups.google.com/forum/#!searchin/django-developers/python$203.4%7Csort:date/django-developers/4rbVKJYm8DI/1LkPWDcYBQAJ

And the associated PR after the discussion: https://github.com/django/django/pull/8884

[...]

PR #3924, opened 2017-10-11
Remove Django 1.8 / 1.10 compatibility

gasman:


wagtail/wagtailcore/middleware.py
```diff
@@ -1,15 +1,9 @@
 from __future__ import absolute_import, unicode_literals
 
-import django
+from django.utils.deprecation import MiddlewareMixin
```
gasman: This is a Django <1.10 compatibility shim https://docs.djangoproject.com/en/1.11/topics/http/middleware/#upgrading-pre-django-1-10-style-middleware and should probably be eliminated too now.

wagtail/wagtailredirects/middleware.py
```diff
@@ -1,19 +1,13 @@
 from __future__ import absolute_import, unicode_literals
 
-import django
 from django import http
+from django.utils.deprecation import MiddlewareMixin
```
gasman: This is a Django <1.10 compatibility shim https://docs.djangoproject.com/en/1.11/topics/http/middleware/#upgrading-pre-django-1-10-style-middleware and should probably be eliminated too now.

wagtail/wagtailcore/middleware.py
kaedroho: class-based middleware is still usable in Django 1.11 though

gasman:
Merged in 2d2518a571fe2890b3d5c6d624a73dc7a75329a2. Will add MiddlewareMixin as a separate task on #3917

wagtail/wagtailcore/middleware.py
gasman: Yep, that's true - it should continue to work for as long as `django.utils.deprecation.MiddlewareMixin` exists, and there's no indication that Django intends to remove it any time soon. It's more of a nice-to-have, in the interests of eliminating legacy code and fully following Django >=1.10 conventions.

Issue #3921, opened 2017-10-11
get_document_multi_form doesn't play well with WAGTAILDOCS_DOCUMENT_MODEL.


coredumperror:
The function `wagtail.wagtaildocs.forms.get_document_multi_form()` builds it's Form object from a fixed set of field names: `['title', 'collection', 'tags']`. This means that the multiple-upload form can never set any other fields on a custom model that's set through `WAGTAILDOCS_DOCUMENT_MODEL`.

The PR for this would be easy, but I haven't got time to make it right now. You'd just need to change the function to this:
```python
def get_document_multi_form(model):
    return modelform_factory(
        model,
        form=BaseDocumentForm,
        fields=[field for field in model.admin_form_fields if field != 'file'],
        widgets={
            'tags': widgets.AdminTagWidget,
            'file': forms.FileInput()
        })
```
That does the equivalent of what `get_document_multi_form()` does today, while also making sure it includes custom fields.

coredumperror:
I've gone ahead and created a PR for this: #3928.

gasman:
Completed in #4840

PR #3928, opened 2017-10-11
Custom fields on custom Document models now appear on multi-upload form.

coredumperror:
As I mentioned in #3921, the original function using a fixed set of field names means that no matter what you set on `CustomDocument.admin_form_fields`, you'll never see your additions in the multi-upload form (the one at /admin/documents/multiple/add/). This change fixes that problem.

### DIFF
--- a/wagtail/wagtaildocs/forms.py
+++ b/wagtail/wagtaildocs/forms.py
@@ -38,7 +38,7 @@ def get_document_multi_form(model):
     return modelform_factory(
         model,
         form=BaseDocumentForm,
-        fields=['title', 'collection', 'tags'],
+        fields=[field for field in model.admin_form_fields if field != 'file'],
         widgets={
             'tags': widgets.AdminTagWidget,
             'file': forms.FileInput()


thearchduke:
In the meantime, you can monkey-patch this by putting it in a wagtail-hooks.py file in one of your apps like so:

`from django import forms
from django.forms.models import modelform_factory
from wagtail.wagtailadmin import widgets
import wagtail.wagtaildocs.forms


def get_document_multi_form(model):
    fields = model.admin_form_fields
    return modelform_factory(
        model,
        form=wagtail.wagtaildocs.forms.BaseDocumentForm,
        fields=fields,
        widgets={
            'tags': widgets.AdminTagWidget,
            'file': forms.FileInput()
        })

wagtail.wagtaildocs.forms.get_document_multi_form = get_document_multi_form
`

gasman:
Completed in #4840. Thanks @coredumperror!

coredumperror:
That's great! I'll be glad to remove our monkeypatch once this goes live.

PR #3930, opened 2017-10-12
3761 sort by url param

### LINKED ISSUE #3761: Form Submission View needs reordering, or newest first default order.

lb-:
Resolves #3761


lb-:
@kaedroho thank you 😄

PR #3931, opened 2017-10-12
Allow runtests.py to pass extra arguments to underlying test runner.

hakjoon:
When the `ArgParse` parser got introduced in https://github.com/wagtail/wagtail/commit/0097f770fd939cf74ae5df7af8e6134bf7f3ce0c we seems to have lost the ability to pass Django test runner commands to the underlying test runner. For example `./runtests.py --failfast` will fail with an `unrecognized arguments` error.

The underlying problem seems to be that the `rest` argument with `nargs='*'` only captures positional arguments.  Switching from `parse_args` to `parse_known_args` prevents the error and allows us to still pass all the non captured arguments to `execute_from_command_line`.  This allows something like `./runtests.py --postgres --keepdb` to work.

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -20,16 +20,15 @@ def make_parser():
     parser.add_argument('--elasticsearch', action='store_true')
     parser.add_argument('--elasticsearch2', action='store_true')
     parser.add_argument('--elasticsearch5', action='store_true')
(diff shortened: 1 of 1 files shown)


gasman:
Thanks @hakjoon! Merged in fa853de486c6542434e27b045ed921052b734eb5.

PR #3934, opened 2017-10-13
Remove Python 2 fallback code

gasman:


kaedroho:
Another one we can do is replace ``from django.utils.lru_cache import lru_cache`` with ``from functools import lru_cache`` (available since Python 3.2)

docs/contributing/developing.rst
```diff
@@ -194,10 +194,7 @@ To start this simple server, run the following commands:
 .. code-block:: console
 
     $ cd docs/_build/html/
-    $ # Python 2
-    $ python2 -mSimpleHTTPServer 8080
-    $ # Python 3
-    $ python3 -mhttp.server 8080
+    $ python -mhttp.server 8080
```
auvipy: as far I know $python command still invoke python2 and $python3 command is needed to invoke python3 in linux

auvipy: python -m http.server 8080
/usr/bin/python: No module named http
auvipy@tb ~> python3 -m http.server 8080
Serving HTTP on 0.0.0.0 port 8080 ...

this is what I get when i run commands on ubuntu

gasman: Thanks for the feedback @auvipy!

If we're going to continue referring to `python3` here, then logically we have to change all other mentions of the `python` command too, such as the `python runtests.py` line earlier in this document. (In the current version of the docs, we have to make a distinction between `python2` and `python3` here because the commands are different, but in the Wagtail 2.0 docs there's nothing special about this line: it's a Python-3-only command, but in Wagtail 2.0 _everything_ is Python-3-only.)

I think it would be a bad idea to refer to the command as `python3` throughout the docs. That name is not a standard - it might be common in Linux distributions _now_, but hopefully in the near future Python 3 will become the default, as Python 2.7 is phased out, and at that point `python3` will be incorrect for those people. (It's also incorrect for people using virtualenv - if you're on a system where `/usr/bin/python3` is the system-wide Python 3 interpreter, then you'd create a virtualenv with `mkvirtualenv wagtail --python=/usr/bin/python3` and use `python` as the command from that point on.)

More generally - we can't possibly anticipate every variation of how Python is configured on different OSes. We simply have to assume that the user has a certain amount of base knowledge about working with Python on their own machine, and may sometimes have to adjust the instructions to fit their situation (for example: if they choose not to use virtualenv, then they need to use `sudo` with `pip install`). As of Wagtail 2.0, ensuring that you're using Python 3 will be another part of that base knowledge.

kaedroho: In a Python 3 ``venv``, ``python`` and ``python3`` both point to the same thing (not sure if this is the same for ``virtualenv`` but I think we should encourage people to use ``venv``). Outside a ``venv``, ``python`` could point to Python 2 (currently every linux distro exept Arch), which would be incorrect.

kaedroho: > but hopefully in the near future Python 3 will become the default, as Python 2.7 is phased out, and at that point python3 will be incorrect for those people

From [PEP 394](https://www.python.org/dev/peps/pep-0394/):

> If the Python 2 interpreter becomes uncommon, scripts should nevertheless continue to use the python3 convention rather that just python. This will ease transition in the event that yet another major version of Python is released.

kaedroho: Discussed offline, users should always use a Python 3 venv when running any python command from the docs. Also, PEP394 has an exception for Windows users meaning that Python 3 may not be callable as ``python3``.

PR #3936, opened 2017-10-16
Fixes form submissions pagination

BertrandBordage:
With this weird `linkurl='-'` argument, the pagination drops the active date filters.
Removing this argument fixes pagination with date filters.

I’m not sure why this argument was introduced, but I guess it could be removed everywhere else, at least when it has the `'-'` value.


BertrandBordage:
This issue was introduced again by 815cb6e405c0671db829279ed418c4a0ed21bbbf.

lb-:
Thanks - sorry about that.

PR #3940, opened 2017-10-18
Postgres_search simplification.

BertrandBordage:
The main goal of this work was to find an elegant way to preserves the original queryset order when using `order_by_relevance=False`.

It’s now done, and thanks to this, I could also remove the existing raw `SELECT` queries.
I had to create a related lookup and manager by inheriting from `GenericRelation` as we use a generic foreign key between `IndexEntry` and indexed models. So now, each model inheriting from `Indexed` also gets an `index_entries` manager and the corresponding QuerySet accessor.

There are two reasons why I had to create a custom `GenericRelation`:
- First of all, I faced [a Django issue](https://code.djangoproject.com/ticket/16055) that was preventing the related accessor from working due to our text `object_id`. Unfortunately, due to [these lines](https://github.com/django/django/blob/01d440fa1e6b5c62acfa8b3fde43dfa1505f93c6/django/db/models/sql/datastructures.py#L73-L79) of the ORM, we can’t cast the column of a `JOIN` condition. But luckily, there is just below a nice `get_extra_condition` method used to get an additional `JOIN` condition. This feature is originally used by `GenericRelation` to filter by content type. So I removed the original `JOIN` condition by overriding `resolve_related_fields`, and I put it instead in `get_extra_restriction`.
- I also had to filter by multiple content types instead of just the content type of the model requested. When we do `Page.objects.search('hello')`, we have to search across fields on the `Page` model, but also on those of the children types. So to do that, I created an `IN` lookup to filter content type across all descendant content types instead of just one.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -33,7 +33,7 @@ def pytest_configure(config):
         pass
 
     if config.getoption('postgres'):
-        os.environ['DATABASE_ENGINE'] = 'django.db.backends.postgresql_psycopg2'
+        os.environ['DATABASE_ENGINE'] = 'django.db.backends.postgresql'
 
     # Setup django after processing the pytest arguments so that the env
     # variables are available in the settings
--- a/docs/advanced_topics/settings.rst
+++ b/docs/advanced_topics/settings.rst
@@ -558,7 +558,7 @@ These two files should reside in your project directory (``myproject/myproject/`
 
   DATABASES = {
       'default': {
-          'ENGINE': 'django.db.backends.postgresql_psycopg2',
+          'ENGINE': 'django.db.backends.postgresql',
           'NAME': 'myprojectdb',
           'USER': 'postgres',
           'PASSWORD': '',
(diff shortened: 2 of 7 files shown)


BertrandBordage:
I renamed remaining mentions of `postgresql_psycopg2` to `postgresql` as the tests were skipped since https://github.com/wagtail/wagtail/commit/67b1ddb665c2688a978ee61936d43010cf687620#diff-c2cc727a5e1fca9050dea34af68aa13c.

BertrandBordage:
The queries are a bit slower now than they were before (~ 5 to 50% slower).

Unfortunately, this is not at all because of SQL: the generated SQL query is in fact more efficient now, taking ~30% less time to execute.
This slowdown is caused by the SQL generation by the ORM. It always takes around 10 ms (!) to generate the query before executing it.

The obvious solution to this issue is to improve Django ORM performance. In the meantime, there is some possible improvement in Wagtail by simplifying the ORM complexity of the query, i.e. writing some parts as raw SQL using `QuerySet.extra()`.

Anyway, we won’t get back to the previous queries as they are a bit fragile and don’t support custom ordering.

Issue #1336, opened 2015-05-22
Feature request - Make streamfield blocks collapsible in admin


mattots:
It would be nice to be able to toggle the content of StreamField blocks in the admin - so with all blocks collapsed you'd just see a list of the block type labels (something like the attached), which, when clicked on, would expand to show the corresponding fields. This would be useful in situations where there are a lot of blocks on a page and you don't want to have to scroll endlessly to find the fields you want to edit.

![screenshot](https://cloud.githubusercontent.com/assets/301729/7774311/322a638e-00a2-11e5-923e-65726189ca90.png)

![screenshot2](https://cloud.githubusercontent.com/assets/301729/7774372/d5009916-00a2-11e5-80ce-b7aa01402b67.png)

gasman:
Hi @mattots,

This seems a bit counter-productive to me... surely if there's a particular image or bit of text you want to edit, you're better off scrolling through the interface looking for the image/text in question, rather than having to think about what type of block it is ("wait, is this an image, or a featured image...?") and clicking into them to find it?

mattots:
A lot of the sites we built (and still maintain) before we started using Wagtail had something similar to (but rather more clunky than) the StreamField with collapsible blocks, so we've got quite a bit of experience of how clients use this sort of functionality, and it seems to be something that they find useful. I don't think it should necessarily be enforced, but an option to make the blocks collapsible via a parameter, and possibly to define whether they start expanded or collapsed, would be nice.

kidlj:
when I use StreamField I feel eager to own this feature, truly.

Ernest0x:
This feature is surely very useful, especially when there is a great number of blocks in the streamfield. @gasman is right in that seeing only the block's type is not enough to find the exact block you want to edit. But this can be easily improved by optionally using the value of a text field as a title while in collapsed state. So, for example, image blocks in the example above could be configured to have a title consisting of the 'caption' field  value. What do you thing?

gasman:
@Ernest0x That sounds like a good idea, although it would add quite a bit of complexity to the implementation - you'd need some way to specify which field to use as the title, as Wagtail doesn't know which fields are the important ones.

[...]

PR #3942, opened 2017-10-19
WIP Streamfield UI changes

### LINKED ISSUE #2325: Deeply-nested StreamFields need better UX
#### Background

StreamField is undoubtedly awesome, but we have a problem that the more complicated a StreamField becomes the more difficult it is for an editor to use or maintain, and has no visual connection to how it will appear to the end-user. 

An example of the problem is a StreamField block called **cards** that we had on a recent project. Out of the box, in StreamField, it looked like this:
Existing problem 
_Admin view_

And on the front end as follows:
Cards on the frontend
_Front end view_

I'm aware that arguably these cards should have been created in a different way e.g. a page reference that pulls the relevant content from the page it's referencing if the content model allowed; or using a custom block to inject CSS/ JS to allow better styling. @alexgleason ended up using the second option, which he's released as a [pip installable package](https://pypi.python.org/pypi/wagtailblocks-cards/2.0.1). It looks like:
Improved cards

I think this is still problematic, though is a vast improvement on how it appears out of the box.
#### Possible solutions

I think there's two ways we can move StreamField from a UX point of view:
1. Take a clearer field approach for each StreamField block (similar to Wordpress's [advanced custom fields](http://www.advancedcustomfields.com/) or CraftCMS's [matrix](https://craftcms.com/features/matrix))
2. Have the semantic value of each block, that isn't purely text, captured outside of StreamField (e.g. in a dialogue, side-panel or separate page) and give the editor a clearer visual representation of what they're creating (similar to the Medium, Squarespace or Google Doc UIs)
#### Attempt at a solution

I think option 2, where the semantic value of each block is introduced outside of StreamField would lead to the richest editorial experience.

**I've uploaded a project on InVision with that in mind**. The first screen is at https://invis.io/ZA64KWAUB. There are hotspots throughout the screens to indicate the user flow, and should allow you to get a general sense of my thinking. I stress it is a little clumsy, and has a number of UI elements changed, which for the purpose of this exercise would have been better left as they currently are within Wagtail, but I think is enough to start a conversation. 

The InVision screen is based on StreamField blocks we needed for a recent project: **text, image, embed, call-to-action, cards, code block** and **downloads**. The workflow is not completely out of keeping with the existing pattern on Wagtail, a user reveals a dialogue if they interact with a `ImageChooserBlock()` or `PageChooserBlock()`.

To take the **cards** as one example (since I used it above to show the problem)
github_0026_streamfield - card 1 complete 2x
Above: Card presented visually as a card, rather than set of fields, with option to add additional card

github_0031_streamfiled - create a card dialogue 2x
Above: Dialogue opened with set of fields to populated card.

github_0032_streamfiled - card upload new image 2x
Above: Additional elements required to populate the field (here an image) would happen within the dialogue rather than opening a new dialogue (since no-one once modals on top of modals!)

github_0038_streamfield - create card completed information 2x
Above: Populated field set

github_0030_streamfield - card 1 hover 2x
Above: User has to hover above the card and click `edit` to amend the content within it.

Note: I'm missing controls on the individual cards to move/ delete them, if you could just pretend they are there that would be great :)

The final page, with all elements added is as below:
_0058_streamfield - download populated 2x
#### Possible UX issues

Beyond the obvious technical issue that what I'm suggesting above is currently impossible within the confines of hallo.js, and difficult to get oEmbed to behave in this way, there are a couple of UX issues that others have flagged since I showed them this work
- Talking with @tmsndrs he noted that we run the risk of trying to pre-define use cases for StreamField. @gasman does a pretty good summation of why that would be a bad idea on the ['Add distinct icons to blocks' pull request](https://github.com/torchbox/wagtail/pull/2285). However, I think there probably is room for us pre-defining things (or at least creating pip installable blocks similar to Alex's cards). Our attempts to avoid pre-definition have led us to a situation where out of the box StreamField blocks can get very confusing, very quickly
- Again talking with @tmsndrs we run the risk that this predicates the population of content around a desktop presentation of the article e.g. an editor could finesse the article around the needs of the desktop user without thinking about mobile devices
- This implementation would still be unable to represent a '[Snow Fall](www.nytimes.com/projects/2012/snow-fall/)' style implementation of StreamField. There's no way I can think of that could actually do that. The few I've seen that try (all Wordpress plugins) that mess with z-indexing or fixed background positioning within the admin get very confusing very quickly. I think it's fine to have that limitation.

Would welcome thoughts, from the perspective of whether this really is a problem, whether my assumption of the direction to take for the solution is the correct one, and then whether the solution itself goes someway towards solving the problem.

heymonkeyriot:
Ready for review.

These are iterative amends towards the longer term desire to improve the look and feel of StreamField. From the Bristol sprint.

Needs to:

 - [x] Present nested streamfields up to 4-levels deep in a usable way
 - [x] Make it clearer to users what the block-type is
 - [x] Make controls visible at all times
 - [x] Make it clearer to users the separation between blocks
 - [x] Consistently style different blocks
 - [x] Consistently style user-input content (except regex and HTML blocks)
 - [ ] ~~Style regex and HTML blocks to reflect manner of input~~
 - [x] Improve visual flow of StreamField elements
 - [x] Consistent spacing regardless of whether StreamBlock nested in StructBlock etc.
 - [x] Improve look and feel of StreamMenu
 - [x] Pass CI tests / linting
 - [ ] Test across many different instances of Wagtail

Closes #2325


gasman:
package-lock.json shouldn't be deleted completely - it should be reverted to its original version. I'd suggest running `git rebase -i master`, deleting the lines for the first two commits (ce0f69b and 
8dec2f4), and then re-pushing with the `--force` option.

heymonkeyriot:
Thanks @gasman. Apologies for the confusion with it. Will sort out getting the package-lock.json file corrected and back here tomorrow morning.

heymonkeyriot:
I think this is broadly ready for review.

A few comments:

 - I haven't changed the regex and html block styling. I think that needs to be a different ticket
 - CI appears to be failing on elasticsearch. I'm not sure what's causing that to happen? The files pass the NPM CSS linter now

**Testing**
Testing so far hasn't been hugely extensive. Would welcome some input of how best to do this. I've reviewed locally against a number of different projects (one of which was using custom CSS for the admin area). Is there a way to use browserstack against specific branches to see how it performs?

**Appearance** (without sidebar or footer to make it easier to see)
screencapture-localhost-8000-admin-pages-81-edit-1509711088923

heymonkeyriot:
@gasman any ideas on who might be able to take a look at reviewing this?

gasman:
@heymonkeyriot Not sure I'm afraid - @thibaudcolas and @jjanssen would be our usual go-to front-end folks, but I don't know how much availability they have...

tomdyson:
@heymonkeyriot thanks for the nudge - I'll make sure this gets discussed at today's core team meeting.

tomdyson:
@thibaudcolas has agreed to review this.

tomdyson:
@heymonkeyriot do you have a deployment of this somewhere? That would help me organise a UI review (@thibaudcolas will be focusing on implementation).

heymonkeyriot:
@tomdyson I'm afraid I only have it locally at the moment. I'll look to get it remotely so that others can review.

thibaudcolas:
I'm experimenting with automating per-branch environments, so this branch is available at https://bakerydemo-thibaudcolas2.herokuapp.com/. This also contains @heymonkeyriot's https://github.com/wagtail/bakerydemo/issues/168 changes to simplify Streamfield testing. A good test page would be https://bakerydemo-thibaudcolas2.herokuapp.com/admin/pages/79/edit/.

heymonkeyriot:
That's great, thanks @thibaudcolas because I was about to come on this thread to apologise for my  lack of progress in getting this deployed :)

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -6,39 +6,54 @@ ul.sequence {
...
     }
 
     .struct-block .fields {
-        @include column(10);
+        width: 100%;
         padding-left: 0;
         padding-right: 0;
     }
 
+    .struct-block label {
```
thibaudcolas: This is quite a far-reaching selector, considering pretty much any field can be inside a Streamfield. I'd suggest going for something more precise if possible (eg. `.sequence-type-stream .fields > li > label`, more specific but targets less potential elements).

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -6,39 +6,54 @@ ul.sequence {
...
     .struct-block .sequence-container {
-        @include column(9);
+        width: 100%;
         padding-left: 0;
         padding-right: 0;
 
         .sequence-member-inner {
-            padding: 1.5em 0;
+            padding: 20px 30px 50px 20px;
         }
```
thibaudcolas: I see this is the same value as below – it should be possible to remove the override then.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -57,52 +72,60 @@ li.sequence-member {
...
+
+            &:first-child {
+                border-top-right-radius: 3px;
+                border-top-left-radius: 3px;
+            }
+
+            &:last-child {
+                border-bottom: 1px solid $color-teal;
+                border-bottom-right-radius: 3px;
+                border-bottom-left-radius: 3px;
```
thibaudcolas: Make a variable out of the 3px radius to make it clearer what is going on between here and `:first-child`.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -57,52 +72,60 @@ li.sequence-member {
...
+
+            .sequence-member-inner .sequence-member {
+                // Third level nested Streamfield
+                border-color: $color-grey-3;
+            }
+
+            .sequence-member-inner .sequence-member .sequence-member-inner .sequence-member {
+                // Fourth level nested Streamfield
+                border-color: $color-teal;
+            }
```
thibaudcolas: Could those styles differentiating between levels be moved outside of this nested declaration, so they are all side by side and we can look at them separately from other "nested Streamfield" styles?

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -57,52 +72,60 @@ li.sequence-member {
...
         }
 
         // sequences within sequences, such as a ListBlock within StructBlock
         > .sequence .sequence-inner {
```
thibaudcolas: I can't find this `sequence-inner` class anywhere else in the codebase. This should probably be removed.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -57,52 +72,60 @@ li.sequence-member {
...
-            padding-left: 0;
-            padding-right: 0;
-        }
-
-        .sequence-type-list .sequence-member-inner {
-            padding: 0;
-            padding-top: 0.5em;
-            padding-bottom: 1.2em;
+            width: 100%;
         }
```
thibaudcolas: Haven't been able to test this unfortunately.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -57,52 +72,60 @@ li.sequence-member {
...
     }
 
     .fields > li {
         position: relative;
+
+        &:last-child {
+            padding-bottom: 0;
+        }
```
thibaudcolas: 👍

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -133,6 +156,7 @@ li.sequence-member {
 
 // This horridly specific selector ensures text inputs, rich text fields and textareas
 // that are direct children of highest level sequence should be borderless and full-width
+// Which is almost a terrible idea @TODO
```
thibaudcolas: Those style overrides make it impossible to see there is a field. Could we just remove the whole block?

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -147,56 +171,59 @@ li.sequence-member {
...
     h3 {
-        margin: 0;
-        font-size: 0.7rem;
         display: inline-block;
-        margin-right: 1em;
+        font-size: 0.8rem;
         line-height: 1.8rem;
+        padding: 0 20px;
 
         label {
```
thibaudcolas: As far as I can see this label is only used for its ARIA value – would you have the courage to refactor this to remove this element (styling the `h3` instead, ideally adding a class to it), keeping the ARIA labelling with a `aria-describedby` attribute on the heading?

wagtail/wagtailadmin/static_src/wagtailadmin/scss/layouts/page-editor.scss
```diff
@@ -425,19 +425,6 @@ footer .preview {
...
-                position: relative;
-                width: 100%;
-                padding-left: 6.4em;
-            }
-
-            &:hover .object-help {
-                opacity: 1;
-            }
-        }
-
```
thibaudcolas: 👍

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -147,56 +171,59 @@ li.sequence-member {
...
+            width: auto;
         }
     }
 
+    // overrides form.scss
     button.icon.text-replace {
+        background-color: transparent;
+        border: 0;
+        color: $color-teal;
+        padding: 0 20px;
```
thibaudcolas: Could you refactor this so the icon is centered in the button's clickable area? See outline below.

streamfield-controls

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -272,57 +298,83 @@ li.sequence-member {
...
+    }
 
-        &:nth-child(4n+1) {
-            clear: left;
-        }
+    h3 {
+        padding: 0.5em 0;
+        margin: 2em 3em 0;
+        border-bottom: 1px solid $color-menu-text;
+        padding-bottom: 1em;
```
thibaudcolas: There is a lot of spacing definitions going on here. Could you refactor some of these to use variables so it's easier to see what is going on?

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -272,57 +298,83 @@ li.sequence-member {
...
+            font-size: 1.2em;
+            width: 2em;
             height: 2em;
             line-height: 2em;
             text-align: center;
+            border: 1px solid $color-grey-3;
+            color: $color-grey-2;
+            border-radius: 8px;
+            margin: 0 auto 12px;
         }
```
thibaudcolas: The block button is misaligned if there is no icon for it (see "Choices" on the right):

streamfield-no-icon

I'm not sure if this PR actually causes this, but could this be easily addressed? Otherwise if that's not a regression we should open a separate issue for this.

benenright:
Hi

Just doing a design review of this and I want to flag the areas that need styling work in order to reflect the design we agreed. Design is attached. Here is a list of the issues:

- [ ] Increase left padding on far left aligned elements labels, streamfield '+' button etc. If you look at the design we are aiming for a consistent reading line on all elements before we indent the nested streamfield (this in turn helps the nested streamfields to be more apprent when they do indent). And match this padding on the nested Streamfields

- [ ] Increase all padding on stroked Streamfield boxes and remove border-radius so they do not get confused for fields (please note the alignemtn of the nested '+' should align to match the padding for a consistent indent (See design) (this works in the same way on the right side with indented controls)

- [ ] Make the '+' button green by default

- [ ] Remove the '+' button above the first element as its messy (we agreed this would be OK)

- [ ] Use correct control icons , move closer together, and change position of the trash can so it is more consistently placed on the right of up/down arrows (See design). Also move all away from divider line above so they feel more connected to the right element (match spacing on design)

- [ ] Labels for child fields (that do not have children) should not be Bold so the hierarchy is clearer

- [ ] Help text should sit much closer underneath the Stremfield block label

Cheers.

streamfield

lb-:
@tmsndrs might be a bit late but there is an older issue #1336 about making StreamField blocks collapsible - might be worth considering.

tmsndrs:
Thanks @lb-,  yeah I'm aware of that request. We did talk about this in the last sprint. Our current thinking was that in most cases Streamfield is used to create content with some sense of narrative even if it's just a general flow of the page so it doesn't make sense to use a collapsed approach out of the box ([as in Clearstream](https://github.com/heymonkeyriot/wagtailclearstream)). So this work is just focusing on making the existing instance of Streamfield more usable. I haven't forgotten about this feature though and we'll consider it down the line.

tmsndrs:
In fact if you look at Ben's design above he's started to think about what that UI might look like with the '-' icon. We're not there yet though, it needs more thought.

lb-:
@tmsndrs that makes sense, and also FWIW this update looks so amazing!

heymonkeyriot:
Thanks @thibaudcolas for the review items. I'll get on them over the weekend.

And thanks @benenright. To run through your bits:

>Increase all padding on stroked Streamfield boxes and remove border-radius so they do not get confused for fields
I had thought the conclusion we'd come to on this was that they should have a radius since the square edge felt too sharp. There seems like a very limited risk of them being confused as being fields.

>please note the alignemtn of the nested '+' should align to match the padding for a consistent indent
This currently is happening. Or are you saying it's aligning on the wrong bit of the '+'. It's currently aligning on the ascender of the `+`, which is how I had interpreted the designs. Let me know if it should be somewhere else.

>Make the '+' button green by default
Again, I had thought the conclusion we came to on this was that that increased the prominence of them too much and distracted from the other UI elements

>Remove the '+' button above the first element as its messy
This was me going rogue and I should have double checked but if we remove this the only way for a user to add a block above the first item is to add a block underneath it and then click the up arrow on that block (or the down arrow on the block above) and that's quite annoying having tested it

>Use correct control icons
I thought we'd agreed that changing the icon set within Wagtail was out of scope of this piece of work. I've used the currently available icons. Though recognise they should be grey rather than green.

>Labels for child fields (that do not have children) 
This isn't possible to do (without some serious wrangling) since there isn't a way to check if an item is a parent of something

>Help text should sit much closer underneath the Stremfield block label
Is there a reason for this? It seems that help text should be treated as any other object (e.g. a field item) and have the same distance between the title and the help text. It's otherwise quite interruptive of the page flow, and I think gives more prominence to those items with help text than I think's necessary.

thibaudcolas:
@heymonkeyriot let me know if I can help getting this through, reviewing / testing / coding 🙂

heymonkeyriot:
Sorry @thibaudcolas, me going slowly, but also partly as I was hoping to get feedback from @benenright to respond to the comments from last Friday 24th. In any event will push remotely where I've gotten to later that should hopefully run through all of your requests :)

tmsndrs:
I just noticed I think this pull request changes a bunch of field text fonts to Bitter. Notably login and site search field.

heymonkeyriot:
Good spot. In the longterm I think that's probably the direction we should take where all user input uses Bitter, whilst all the Wagtail UI is OpenSans. But agree - with what I think you're saying - that that's out of scope of these changes.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -6,39 +6,54 @@ ul.sequence {
...
-                opacity: 1;
-            }
-        }
+    &:first-child {
+        border-top: 0;
     }
 
-    &:hover .object-help {
+    // Note this inherits .oject-help from page-editor.scss
+    .object-help {
```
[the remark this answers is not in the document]
heymonkeyriot: I think the more accurate answer is "not easily". We could change the classes to avoid any inheritance, which I could see would have benefits, but it could also add additional confusion to name-spacing/ ensure design consistency. It feels like a piece of work for once this PR is in when we can start thinking about improving the UI beyond StreamField.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -6,39 +6,54 @@ ul.sequence {
...
+        display: block;
+        float: none;
+        font-style: italic;
```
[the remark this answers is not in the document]
heymonkeyriot: There already is inconsistency between help text across Wagtail. I think within the context of StreamField this does make it clearer. Personally I'd advocate leaving in and raising an issue to look at how help text is implemented on non-StreamField field items. What do you think?

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -6,39 +6,54 @@ ul.sequence {
...
+        max-width: 650px;
         opacity: 1;
+        padding-left: 0;
+        position: relative;
+        width: 100%;
+
+        &::before {
+            display: none;
+        }
+        @media screen and (max-width: $breakpoint-mobile) {
```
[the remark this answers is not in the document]
heymonkeyriot: I can't see that we're actually using those mixins anywhere (at least not with the scss files) Do you have examples? I can't see how something in `client` would be able to be to be in scope for the `static` folder, though entirely possible that I'm misunderstanding something?

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -6,39 +6,54 @@ ul.sequence {
...
+            margin: -20px 0 30px;
```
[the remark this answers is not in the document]
heymonkeyriot: `top: 0` or `position: static` visually don't quite give the exact expected visual but agree using the negative margin is a bad way to get to it. Have reverted to `top: 0` and will work out a better way to get the desired positioning.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -6,39 +6,54 @@ ul.sequence {
...
+            padding-left: 0;
```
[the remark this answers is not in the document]
heymonkeyriot: 👍

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -6,39 +6,54 @@ ul.sequence {
...
 
     .struct-block .fields {
-        @include column(10);
+        width: 100%;
         padding-left: 0;
         padding-right: 0;
     }
 
+    .struct-block label {
+        display: block;
```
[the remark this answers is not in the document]
heymonkeyriot: It's set to `display: inline;` in `forms.scss`

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -6,39 +6,54 @@ ul.sequence {
...
+        float: none;
```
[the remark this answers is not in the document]
heymonkeyriot: It's set to `float: left;` in `forms.scss`

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
heymonkeyriot: Good point. It's tricky as lines 6 - 153 could be wrapped in a `.sequence-type-stream` class if it weren't for my fear that some of these declarations are being used by lists (since `sequence_member.html` is extended by both `list_member.html` and `stream_member.html`). For lack of better ideas I've followed your suggestion by pulling it out of the `li.sequence-member` selector and on its own.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -6,39 +6,54 @@ ul.sequence {
...
+        text-transform: uppercase;
```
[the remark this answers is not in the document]
heymonkeyriot: It is intentional yes. We're already changing the appearance quite dramatically by removing the float (so that the field sits below it) and the idea was to introduce consistency with this label and the parent label. That said I'm fairly agnostic about whether these should be lowercase.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
```diff
@@ -6,39 +6,54 @@ ul.sequence {
...
+        width: 100%;
```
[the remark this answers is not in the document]
heymonkeyriot: 👍

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
heymonkeyriot: Ah, yes, there was a point at which these had different values when we were thinking of the design. Removed.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
heymonkeyriot: Good spot. Removed.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
heymonkeyriot: Yup, it's a problem that - as far as I'm aware - has always been in Wagtail. It's certainly been in there since I started using wagtail 0.9. I think there _is_ an issue buried somewhere on it. If not will open one.

wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_streamfield.scss
heymonkeyriot: 👍

BertrandBordage:
Can we help on this in any way? Seems like this is one of our main blocker for 2.0 now.

tomdyson:
@heymonkeyriot thanks for your recent work on this (commits on 10 Dec 2017). Do you think this is ready now, or is there more to do?

BertrandBordage:
Closing this PR since the StreamField UI is going to be fully rebuilt by me as part of the Wagtail’s first Kickstarter campaign.
https://www.kickstarter.com/projects/noripyt/wagtails-first-hatch

thibaudcolas:
@BertrandBordage Generally I don't think it's a good idea to close PRs in advance of future work. We of course do not want people to work on solutions to the same problem at the same time, so it's good to give a heads up everywhere, but _at this point in time_ this PR still has a lot of relevance.

I'm not sure how relevant this rule is in this case, but keeping things open "until proven irrelevant" doesn't have many drawbacks. See also #4215, which tackles the same issue as this PR but has been kept open so far.

heymonkeyriot:
I think it makes sense to close the PR. I'm afraid I won't have time to finish the outstanding reviews on the PR, and the direction that Bertrand's taking makes sense. Good luck with it all!

BertrandBordage:
Superseded by #4473.

PR #3950, opened 2017-10-19
Adds autocomplete to PostgreSQL search.

BertrandBordage:
This adds autocomplete to the PostgreSQL search backend.

To use it:
```python
from wagtail.wagtailsearch.backends import get_search_backend
get_search_backend().autocomplete('wagta', Page)
```
And it will search for inside the fields marked with a `SearchField(…, partial_match=True)`. Of course, it’s a prefix search, so this query will match `Wagtail`, `wagtalianism`, etc.

This method accepts all the arguments that `search(…)` also accepts, and takes them into account. So you can ask for autocomplete on a specific field, for example, and without relevance sorting.

This is the first backend that gets autocomplete, so I had to change a bit the base backend class to make it work.

I didn’t put any docs or tests yet, as this is a first important feature before getting more search features in Wagtail. This first feature can be integrated as an unstable API for now if it is popularly demanded, but we’ll change it very soon before integrating final tests & docs.

Also, I’m aware that this doesn’t work like the partial queries we can do with the Elasticsearch backend. I made this pull request to show how I imagined it should work. Now that this is done, we can further imagine how we can conciliate the constraints of both backends in order to have the same API.

And finally, I’m also aware that we’re going far away from the DB backend with the search improvements we are working, @kaedroho & I. But in the end, I want to merge the DB and PostgreSQL backends: they will both include the same `IndexEntry` model, but PostgreSQL will benefit from a better quality and some features will be disabled on the DB backend.


wagtail/contrib/postgres_search/models.py
```diff
@@ -14,6 +14,11 @@
 from .utils import get_descendants_content_types_pks
 
 
+class SearchAutocomplete(SearchQuery):
+    def as_sql(self, compiler, connection):
+        return "to_tsquery(''%s':*')", [self.value]
```
kaedroho: Shoudn't this be: ``to_tsquery('''%s'':*')``?

BertrandBordage: No, because psycopg2 escapes the argument, `self.value`. This generates `to_tsquery('''your search'':*')`.

wagtail/wagtailsearch/backends/base.py
```diff
@@ -270,14 +281,19 @@ def search(self, query_string, model_or_queryset, fields=None, filters=None,
...
-            queryset, query_string, fields=fields, operator=operator, order_by_relevance=order_by_relevance
-        )
-        return self.results_class(self, search_query)
+        self.check_operator(operator)
+
+        return (queryset, query_string), {
+            'fields': fields, 'operator': operator,
+            'order_by_relevance': order_by_relevance}
+
+    def search(self, query_string, model_or_queryset, fields=None, filters=None,
```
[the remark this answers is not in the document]
BertrandBordage: Well… To me it makes things clearer and more extendable.
Instead of adding yet another argument, couldn’t it be the occasion to make it an `Autocomplete` class according to your RFC?

I guess we need to do `PlainText` and `MatchAll` before, then I’ll modify this pull request accordingly.

BertrandBordage:
This pull request will be created again once #4059 is merged.

BertrandBordage:
Replaced by #4172.

PR #3951, opened 2017-10-20
Prevents PostgreSQL from duplicating ancestor and descendant entries of a same instance.

BertrandBordage:


wagtail/contrib/postgres_search/utils.py
```diff
@@ -60,17 +60,29 @@ def get_descendant_models(model):
...
-def get_descendants_content_types_pks(model):
+def get_content_type_pk(model):
+    # We import it locally because this file is loaded before apps are ready.
+    from django.contrib.contenttypes.models import ContentType
+    return ContentType.objects.get_for_model(model).pk
+
+
+def get_ancestors_content_types_pks(model):
+    """
+    Returns content types ids for the descendants of this model, excluding it.
```
kaedroho: descendants?

wagtail/contrib/postgres_search/utils.py
```diff
@@ -60,17 +60,29 @@ def get_descendant_models(model):
...
+    """
     from django.contrib.contenttypes.models import ContentType
     return [ct.pk for ct in
-            ContentType.objects.get_for_models(*get_descendant_models(model))
-            .values()]
+            ContentType.objects.get_for_models(*model._meta.parents).values()]
```
kaedroho: IIRC ``model._meta.parents`` only gets the immediate parents of this model, not all ancestors

wagtail/contrib/postgres_search/backend.py
```diff
@@ -143,14 +143,21 @@ def add_items_update_then_create(self, content_type_pk, objs, config):
...
 
     def add_items(self, model, objs):
-        content_type_pk = get_content_types_pk(model)
+        content_type_pk = get_content_type_pk(model)
         config = self.backend.get_config()
         for obj in objs:
             obj._object_id = force_text(obj.pk)
             obj._body_ = self.prepare_body(obj)
+
+        # Removes index entries of an ancestor model in case the descendant
```
kaedroho: I think this will be better implemented by storing the root content type (eg, ``Page``) alongside the leaf content type (eg ``HomePage``). Then make each index entry unique on the root content type and ID.

This would mean that if the same object was indexed twice with different leaf types, only a single instance would be in the index.

I think we should solve the issue of indexing the same object twice by adding a better default implementation of ``get_indexed_objects`` on the ``Indexed`` class.

BertrandBordage:
Thanks for the review @kaedroho!

Fixed the issues you pointed out. You’re right about your last comment, but I’m not entirely sure yet how we should do it. It’s a future work outside of the scope of this pull request.

Issue #3709, opened 2017-07-13
Autocomplete for PostgreSQL search backend


kaedroho:
Autocomplete isn't currently supported by the PostgreSQL backend (#3515). We've had to switch a couple of clients back to Elasticsearch because the search experience isn't as good without it.

I've made a start of implementing this in https://github.com/kaedroho/wagtail/tree/feature/postgres-search-autocomplete. It's still early days though, just opening this issue to track progress and make other developers aware of what I'm working on.

BertrandBordage:
Sorry to be so direct, but I don’t think this is the right way to implement it.
You’re basically creating explicitly in a new column what’s already in the GIN index. Except that unfortunately, it’ll be less efficient than GIN.

As you can see [in the PostgreSQL docs](https://www.postgresql.org/docs/9.6/static/textsearch-controls.html), it is possible to do “prefix matching” using a syntax such as `to_tsquery('supercalif:*')`.

I already implemented it [in django-postgrefts](https://github.com/BertrandBordage/django-postgrefts/blob/master/postgrefts/models.py#L194-L201), it works very well, but I don’t think we can implement it properly in wagtail without doing some preliminary API changes. That’s because we don’t really need to create a new column/index just for prefix matching a field.

That’s one of the reasons why I told you a few months ago that we should discuss about changing the search API ;) One of the main others reasons is multi-language support of search, as the multi-index is not at all the most efficient solution, at least for PostgreSQL.

kaedroho:
@BertrandBordage Wagtail at the moment lets you specify which ``SearchFields`` have ``partial_match`` enabled. So to make it behave properly, we have to index those fields separately (but, of course, we can discuss changing that behaviour).

I was also thinking that precalculating the edgengrams would be slightly faster as each term in the search query would map to a term in the database whereas prefix queries may have to pull out many terms and combine the results on the fly. AFAIK, there isn't a reason why two GIN indexes can't be used simultaneously in the same query (but that isn't documented so I don't know this for sure).

BertrandBordage:
Well, matching a prefix and using a binary tree index (or a GIN index, GIN being an improved b-tree index) are exactly the same thing: if you search for “something”, a first step is made into the tree, at the “s” root, then we continue to the “o” node, etc until the “g” leaf. On each step the number of results is narrowed down. So when you do a prefix matching with a database index, you actually do less work (results fetching put aside) than to do an full match.

[...]

PR #3953, opened 2017-10-20
Adds PostgreSQL row boosting

BertrandBordage:
To use this, simply give a per object boost by defining a method in your model:
```python
def get_search_boost(self):
    return 5
```
Here, this multiplies by 5 the weight of this specific model in search results. So this model will almost always be first in results because that is a huge boost. Smaller boosts make a nicer compromise between relevance to the query and importance of that model.

Boost can be any float, even negative ones, but keeping it positive is better. The default boost is 1. To decrease the boost of a model, take a boost between 0 and 1.

Another more complex example:
```python
def get_search_boost(self):
    return 2 if 'wagtail' in self.title.lower() else 1
```
This gives items containing 'wagtail' in their title a boost. You can imagine any boost adjustment you want from there. For example, you can boost pages depending on their depth in the tree, in order to make categories more easy to find. Or you can boost an item depending on the number of visits it has (a number you’ll fetch from some analytics software).

Note: this pull request depends on merging #3709 first.


kaedroho:
👍 on the API and implementation, just needs docs and tests.

BertrandBordage:
Thanks @tomdyson! Just applied your suggestion.

BertrandBordage:
This pull request is ready, but not for review.
It contains #3709 and will not pass recent tests unless I add #3951 in it as well.

So once @kaedroho confirms me that I can merge #3951 into master, I’ll make two new pull requests: one for autocomplete, and the other one for row boosting.

I also have 2 other big changes to the PostgreSQL backend that I need to add as well, including one that allows to make multi-languages search. Ideally, it would be easier for me to just make a huge pull request with everything in it, but I guess no-one else will like it ;)

BertrandBordage:
This pull request will be created again once #4059 is merged.

BertrandBordage:
Superseded by #4534.

PR #3947, opened 2017-10-19
Search fixes

kaedroho:

PR #3957, opened 2017-10-20
Search fixes (backport for 1.12.x)

kaedroho:
This backports the search bug fixes in https://github.com/wagtail/wagtail/pull/3947 to 1.12.x

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -351,6 +351,7 @@ class Page(six.with_metaclass(PageBase, AbstractPage, index.Indexed, Clusterable
 
     search_fields = [
         index.SearchField('title', partial_match=True, boost=2),
+        index.FilterField('title'),
         index.FilterField('id'),
         index.FilterField('live'),
         index.FilterField('owner'),
--- a/wagtail/wagtailsearch/backends/elasticsearch.py
+++ b/wagtail/wagtailsearch/backends/elasticsearch.py
@@ -280,10 +280,8 @@ class ElasticsearchSearchQuery(BaseSearchQuery):
         if lookup == 'in':
             if isinstance(value, Query):
                 db_alias = self.queryset._db or DEFAULT_DB_ALIAS
-                value = next(value.get_compiler(db_alias)
-                                  .execute_sql(result_type=MULTI))
-
-                value = [r[0] for r in value]
+                resultset = value.get_compiler(db_alias).execute_sql(result_type=MULTI)
+                value = [row[0] for chunk in resultset for row in chunk]
 
             elif not isinstance(value, list):
                 value = list(value)
--- a/wagtail/wagtailsearch/backends/elasticsearch.py
+++ b/wagtail/wagtailsearch/backends/elasticsearch.py
@@ -5,7 +5,7 @@ import json
 
 from django.db import DEFAULT_DB_ALIAS, models
 from django.db.models.sql import Query
-from django.db.models.sql.constants import SINGLE
+from django.db.models.sql.constants import MULTI
 from django.utils.crypto import get_random_string
 from django.utils.six.moves.urllib.parse import urlparse
 from elasticsearch import Elasticsearch, NotFoundError
@@ -280,8 +280,11 @@ class ElasticsearchSearchQuery(BaseSearchQuery):
         if lookup == 'in':
             if isinstance(value, Query):
                 db_alias = self.queryset._db or DEFAULT_DB_ALIAS
-                value = (value.get_compiler(db_alias)
-                         .execute_sql(result_type=SINGLE))
+                value = next(value.get_compiler(db_alias)
+                                  .execute_sql(result_type=MULTI))
+
+                value = [r[0] for r in value]
+
             elif not isinstance(value, list):
                 value = list(value)
             return {
--- a/wagtail/wagtailsearch/backends/elasticsearch.py
+++ b/wagtail/wagtailsearch/backends/elasticsearch.py
@@ -398,18 +398,9 @@ class ElasticsearchSearchQuery(BaseSearchQuery):
 
         # Get queryset and make sure its ordered
         if self.queryset.ordered:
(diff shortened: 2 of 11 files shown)


kaedroho:
The test failures happening here are also happening on the ``stable/1.12.x`` branch

gasman:
That'll be due to #3912. (Come to think of it, #3912 is probably enough to justify a new 1.12.x release in its own right...)

gasman:
@kaedroho Please can you add tests for these fixes?

gasman:
@kaedroho What's the motivation for backporting 8df04d0c333eac96ed11cfe5c0c5ba34f3580cc2? If I'm reading it right, it makes the Elasticsearch backend fail with a more meaningful exception when an invalid `order_by` is used, and makes the db backend consistent with that (i.e. equally strict) where it wasn't before - which are both good things from a code quality perspective. However, it means that for db backend users, we're potentially breaking user code that's currently working just fine - which A) possibly warrants an 'upgrade considerations' note, and B) is the sort of thing we shouldn't really be doing in a bugfix release unless there's an actual benefit for the end user.

wagtail/wagtailsearch/tests/test_elasticsearch5_backend.py
```diff
@@ -410,7 +410,7 @@ def test_none_lookup(self):
         # Check it
         expected_result = {'bool': {'filter': [
             {'match': {'content_type': 'searchtests.SearchTest'}},
-            {'missing': {'field': 'title_filter'}}
+            {'bool': {'mustNot': {'exists': {'field': 'title_filter'}}}}
```
gasman: Fix here looks good to me, but it feels like we ought to have a test that actually runs the resulting query (even if it's just "throw this query at Elasticsearch and confirm that it doesn't break").

wagtail/wagtailcore/models.py
```diff
@@ -351,6 +351,7 @@ class Page(six.with_metaclass(PageBase, AbstractPage, index.Indexed, Clusterable
 
     search_fields = [
         index.SearchField('title', partial_match=True, boost=2),
+        index.FilterField('title'),
```
gasman: LGTM, but again could do with a test (even if it's just confirming that `Page.objects.order_by('title').search('blah', order_by_relevance=False)` completes without an error)

wagtail/wagtailsearch/backends/elasticsearch.py
```diff
@@ -280,8 +280,11 @@ def _process_lookup(self, field, lookup, value):
         if lookup == 'in':
             if isinstance(value, Query):
                 db_alias = self.queryset._db or DEFAULT_DB_ALIAS
-                value = (value.get_compiler(db_alias)
-                         .execute_sql(result_type=SINGLE))
+                value = next(value.get_compiler(db_alias)
+                                  .execute_sql(result_type=MULTI))
```
gasman: Here `execute_sql` is returning an iterator that yields lists of result rows. Are we confident that we're only ever going to need to look at the first of those lists, as implied by the use of `next`?

(Digging into https://github.com/django/django/blob/308f64462421b09b21ef0dcd9cc3654cc25bceba/django/db/models/sql/compiler.py#L855, it looks like this iterator setup possibly exists in order to support `chunked_fetch`, which is false by default, in which case we're fine. I'm hoping you or @BertrandBordage have already researched this, so I don't have to :-) )

gasman: For some reason, when running this new code against the updated tests from 411fb6b457ea5ab8735459146c33557af3187011, `execute_sql` is now giving me a list of lists of rows, rather than an iterator of lists of rows, and failing on `next` with `'list' object is not an iterator`. However, I'm still getting back an iterator as expected when I do the same thing via `manage.py shell` against bakerydemo... investigating.

    >>> from wagtail.wagtailsearch.backends import get_search_backend
    >>> backend = get_search_backend('default')
    >>> backend.search('bread', Page.objects.filter(id__in=Page.objects.all().values_list('id', flat=True)))
    multi result: <generator object cursor_iter at 0xb56f6e0c>
    <SearchResults []>

gasman: ah, it's sqlite (the default for runtests) versus postgresql, isn't it... https://github.com/django/django/blob/308f64462421b09b21ef0dcd9cc3654cc25bceba/django/db/models/sql/compiler.py#L917-L922

gasman: Now updated in 25727ea to handle list and iterator responses from `execute_sql` equally (and also look beyond the first chunk of chunked resultsets, if we ever have to deal with those).

wagtail/wagtailcore/models.py
gasman: Test for searching an ordered queryset added in 6bdf75c2aa2f7629be9a9b4f479208236d3d8c61. Not entirely what I was going for, but it looks like we don't have a good place to put tests that check that `Page`'s search configuration does what we want it to, and creating such a place (and setting it up to exercise all search backends) is going to need more heavy refactoring of our test suite than we can comfortably do here.

gasman:
Back to @kaedroho to confirm he's happy with my fixes/tests, and with my suggestion (https://github.com/wagtail/wagtail/pull/3957#issuecomment-343457651) that we leave out 8df04d0c333eac96ed11cfe5c0c5ba34f3580cc2 from the backported fix.

wagtail/wagtailcore/models.py
gasman: I've now come up with a good way of testing search backends against the Page model, as part of my fix for #4015 - but that's just revealed that `Page.objects.order_by('title').search('blah', order_by_relevance=False)` is _also_ broken on the Postgres backend. Argh!

gasman: ...ah, that's #3726, which I guess we ought to backport to 1.12 along with the other fixes. Sorry for the noise :-)

kaedroho:
Replaced by #4027

PR #3960, opened 2017-10-20
Update writing_templates.rst

rifuso:
Since page.your_slug normally results in None, I think using a quoted slug would a better example.


gasman:
Thanks for the suggestion @rifuso! You're right that a quoted slug would make more sense here - I think it would be even better to turn it into a "real-world" example, though. Have now done that, in 9ec8daba4385a768bcd4a15f7a3a6f09cd30da6f.

Issue #3886, opened 2017-09-29
Improvement: Image usage reporting 


khink:
Images might be rendered in several parts of the site (for example, a Product image that appears in listings of various layouts). 

It would be nice to have an overview of:

- All places where that image is being used. This would make it easier to check if the image can be deleted.
- All dimensions that the image is rendered in. This would would make it easier to detect cases where resizing gives undesirable results.

khink:
Started during Bristol Sprint 2017.

Talked with @gasman, and the first part is rather easy: set `WAGTAIL_USAGE_COUNT_ENABLED = True` in your settings and you will get a link in the header of the Image edit view. (Documented in http://docs.wagtail.io/en/v1.13/advanced_topics/settings.html#usage-for-images-documents-and-snippets)

khink:
As for the second part, showing all the ways an image is (or could be) rendered: I have toyed with a  new view for this (https://github.com/khink/wagtail/tree/feature/image-rendering-reporting) which shows all existing Renditions. The usability is quite limited and I will not create a PR for this. Reasons:

1. Renditions are not available immediately after Image upload, but only after the Image was rendered in a certain size. This could be changed if you create all required Renditions immediately after Image upload, but for that you'd have to keep track of all rendition parameters in use, as mentioned in https://github.com/wagtail/wagtail/issues/929
2. Renditions are not linked to a particular page or URL (as they might be created from any template), so we can't tell admins where they can see the image in that size.
3. "Older" renditions (i.e Renditions that are not shown anywhere in the site, b/c template changed aspect ratio, or focal point changed) would have to be filtered out.

To address the issue of getting a preview of all possible uses of the image in the site, i think the approach in https://github.com/wagtail/wagtail/issues/929 would be the way to go, but i don't think this needs to be in Wagtail core. It might be a nice add-on.

khink:
#3954 shows a proof of concept how to make usage also list pages that have the image in a streamfield. Might be a nice add-on for Wagtail.

khink:
In #3961, Bertrand started work to improve the performance of this, and made it more re-usable.

Issue #3854, opened 2017-09-13
image: no warning when user deletes image that multiple pages use


matiasli:
Found a bug? Please fill out the sections below. 👍
more of a feature request for a client that will be using this site

### Issue Summary

When on the CMS, the option to delete an image file that is used on multiple pages can occur.

### Steps to Reproduce

1. open and edit a page in the CMS

2. go to an image (that is used throughout the website) and edit

3. delete the image

### Technical details

* Wagtail version: 1.12
* Browser version: any browser

AlexanderMackie:
Another consequence of this bug is if you upload an image that already exists, then delete it (perhaps because you didn't know it was a duplicate), the _original_ image will also be deleted.

gasman:
Hi @AlexanderMackie - I don't think that's the issue described here. Are you using `S3BotoStorage` to host image files remotely on S3, by any chance? This is a known issue with django-storages - see #1686.

AlexanderMackie:
My issue was that unexpected default django-storages behaviour. Thanks for pointing me in the right direction!

lb-:
Relevant ongoing discussion about better ability to report on image usage at #3886 - this is probably needed before any kind of warning page would be reliable.

zerolab:
This was fixed with the addition of the improved usage reports released in Wagtail 4.1.
![Screenshot 2023-10-11 at 09 23 54](https://github.com/wagtail/wagtail/assets/31622/bd3f2d3e-31d6-430b-baca-f531eeb128ba)

Issue #3010, opened 2016-09-19
Deleting page from usage view sends user back to wrong place


kaedroho:
### Issue Summary

When you delete a page from the page type usage view, you are sent back to the explorer view of the parent page instead of the usage view
### Steps to Reproduce
1. Click "add subpage" somewhere, taking you to the "choose a page type" page
2. On the right hand side, click any of the "pages that use type X" links (if the one you clicked on has no pages, find another)
3. Now delete one of the pages
4. You should now be redirected to the wrong place

I expected to be taken back to the page usage view
### Technical details
- Python version: 2.7
- Django version: 1.9.9
- Wagtail version: 1.6.2
- Browser version: Firefox 46.0.1

ramonakira:
I've looked into it and there's some things of note for this issue:

1) It's not only applicable to the delete action, but also for a few other possible actions,
2) After reviewing the code it seems to me that we need to pass 'next' using GET, for the actions that are created using a hook I don't see a way of getting to the request so I don't see a way to pass the current page.

How should we proceed?

Patil2099:
@kaedroho Can I work on this issue?

gasman:
@Patil2099 Yes, definitely - no need to ask for permission :-)

lb-:
Merged in via #5764

PR #3961, opened 2017-10-20
New UI for listing object uses, including in RichText and StreamField.

BertrandBordage:
This class does a similar job to Django’s collector, except that it works for StreamFields.

It can find:
- all the uses of a block type and all its corresponding values
- all the uses of a specific set of values in a specific block type

It relies a lot on generators, which allows for optimizes `any` queries.
For example, we may want to know if a given image is used in any StreamField from the database, and stop at the first use found:
```python
from wagtail.wagtailcore.blocks.collector import get_all_uses

if any(get_all_uses(ImageChooserBlock, [image])):
    print('You can’t delete this image as it’s used at least once.')
else:
    print('You can delete this image!')
```

Of course, images can be used at other places than StreamField. This is a future work: make the same for `RichTextField`+`RichTextBlock`, then create a function to call all functions: Django’s ORM relations collector, the StreamField collector and the rich text collector (that would also rely on the StreamField collector, combined with regular expressions).


wagtail/wagtailcore/blocks/collector.py
```diff
@@ -0,0 +1,104 @@
...
+        if isinstance(value, Model):
+            return str(value.pk)
+        if isinstance(value, (int, float)):
+            return str(value)
+        if isinstance(value, bool):
+            return 'true' if value else 'false'
+        if isinstance(value, None):
+            return 'null'
+        return '"%s"' % re.escape(str(value).replace('"', r'\"')
+                                  .replace('|', r'\|'))
```
BertrandBordage: Almost all this could probably be replaced with `json.dumps(…)`. I wrote it this way since it seemed more flexible if I needed some special case like models.

wagtail/wagtailcore/blocks/collector.py
```diff
@@ -0,0 +1,104 @@
...
+            return
+        block_paths_per_field = [c.find_block_type(block_type)
+                                 for c in self.field_collectors]
+        filters = Q()
+        for field, block_path in zip(self.fields, block_paths_per_field):
+            block_filter = Q(**{field.attname + '__regex': r'"(%s)"'
+                                % '|'.join([block.name for block in block_path
+                                            if getattr(block, 'name', '')])})
+            value_filter = (
+                Q(**{field.attname + '__regex': '(%s)'
```
BertrandBordage: Needs an `r` escaping.

chosak:
@BertrandBordage very interesting! I tried to do something similar in [wagtail-inventory](https://github.com/cfpb/wagtail-inventory), but hadn't considered the similarity to how Django looks up related objects.

In wagtail-inventory the page->block relationships are stored in a DB table, updated via page signals, and exposed for query through the admin. I'd (perhaps wrongly) assumed that doing an on-the-fly query wouldn't be scalable for large sites - have you done any performance testing of this approach across many pages/complex streamfields?

khink:
@BertrandBordage Do you think this part would go into core eventually? I’d like to write tests + doc for it, but am wondering what would be the correct place for it. An add-on product maybe?

Another thing: I tried this in bakerydemo, and tried to find the linked streamfield on the homepage, but it doesn't find that. In `bakerydemo.base.models.py`:
```
    def get_usage(self):
        if getattr(settings, 'IMAGE_USAGE_SEARCHES_STREAMFIELDS', False):
            from wagtail.wagtailcore.blocks.collector import get_all_uses
            uses = get_all_uses(ImageChooserBlock, [self])
        else:
            return get_object_usage(self)
```
This is something that a test could likely clarify.

And in order for this to work with existing templates, we'd have to create a wrapper around it that has the count() method, for example like in https://github.com/wagtail/wagtail/pull/3954/files#diff-e0210a80d3ac707eaa237beb0f6e5cc5R89, right?

BertrandBordage:
@chosak It’s indeed sensibly slower than having an intermediate table that’s automatically updated.
However, this is easier to maintain for now. Having an automatic intermediate table is pretty heavy and can lead to an outstanding number of issues, like:
- database changes outside Django will lead to stale records in this table
- bulk changes inside the database will also lead to stale records in this table
- it significantly increases the database size
We might defer doing this until there is some support for `TRIGGER`s in Django.

@khink Yes, this will go to the core, along with a similar collector for rich text fields/blocks. It indeed needs a lot of tests before being considered stable.
I don’t think the way you use it does what you want. When you do `get_all_uses(ImageChooserBlock, [self])`, it searches in all models if there is an `ImageChooserBlock` somewhere where the value is in `[self]`. As far as I understand, `self` is here your home page model. So you’re telling it to find an image block where the value is an home page, which is irrelevant, and therefore why you don’t get any result.
About the API to provide on this feature, I think we need to create an additional abstraction layer that will join results from Django’s collector, streamfield & rich text collectors. `.count()` will probably no longer be relevant, as we won’t directly use QuerySets anymore, because we will fetch results from various models. By the way, `get_object_usage` is really basic as it only shows usage in `Page` objects as well as inline models, but nothing for snippets or any other model.

chosak:
Bumping this due to [this similar request on Stack Overflow](https://stackoverflow.com/questions/47519741/wagtail-is-there-a-way-to-get-all-pages-a-snippet-is-used-on) looking for a way to find all pages containing a snippet.

BertrandBordage:
Just improved a lot this pull request by adding a collector for rich text, as well as using Django’s collector for simple ForeignKey/ManyToMany/etc relations.

Now, the `get_all_uses` function takes only a single object as argument:
```python
from wagtail.core.collectors import get_all_uses

get_all_uses(Image.objects.first())
```

I’m aware that this work is difficult to review: it’s fairly complex and it’s hard to figure out what’s going on without an excellent knowledge of `StreamField`, `RichText` and `Model` internals.

If someone could at least **review the tests to check that at least it works** as expected, it would be awesome :)

wagtail/core/tests/test_collectors.py
```diff
@@ -0,0 +1,312 @@
...
+        with self.assertNumQueries(55):
+            uses = list(get_all_uses(self.obj3))
+            self.assertListEqual(uses, [self.obj4])
+
+    def test_streamfield(self):
+        with self.assertNumQueries(55):
+            uses = list(get_all_uses(self.obj3))
+            self.assertListEqual(uses, [self.obj4])
+
+    def test_multiple(self):
```
chosak: I ran into a tricky issue with this test. If you run only these new test cases (`wagtail.core.tests.test_collectors`) without also running the full test suite, this specific test fails with `AssertionError: 47 != 46 : 47 queries executed, 46 expected`.

The extra 47th query that gets run if you **don't** first run one of those test cases is

```sql
18. SELECT "django_content_type"."id", "django_content_type"."app_label", "django_content_type"."model" FROM "django_content_type" WHERE ("django_content_type"."app_label" = 'wagtailimages' AND "django_content_type"."model" = 'image')
```

So this appears to be something where the image content type maybe [isn't cached](https://docs.djangoproject.com/en/2.0/ref/contrib/contenttypes/#django.contrib.contenttypes.models.ContentTypeManager.clear_cache) as part of the Django content type framework.

I dug around and one way to get `wagtail.core.tests.test_collectors.GetAllUsesTest.test_multiple` to pass is if you first run a `wagtail.snippets.tests` test case that loads [the `test.json` fixture](https://github.com/wagtail/wagtail/blob/db6d36845f3f2c5d7009a22421c2efab9968aa24/wagtail/tests/testapp/fixtures/test.json) (for example [`wagtail.snippets.tests.TestSnippetChooserBlock`](https://github.com/wagtail/wagtail/blob/db6d36845f3f2c5d7009a22421c2efab9968aa24/wagtail/snippets/tests.py#L7199)). I'm not super familiar with how content types work, but presumably there's some step in this other test that causes that content type to get cached.

One option would be to do [something like this](https://github.com/django/django/commit/7cd3f1c29595d1da7f37d29e7c3bc6a7a314cd1d) to manually clear the cache first (suggested by [this Django ticket](https://code.djangoproject.com/ticket/23746)) to avoid this problem. Ideally all tests should be runnable on their own; is there a more flexible approach than using a hardcoded number of SQL queries here?  If and when new models get added to Wagtail anywhere, these tests will have to get updated as the number of queries increase.

BertrandBordage: You’re totally right, I’ll change the tests in order to clear the ContentType cache before.
That way, we’ll test the maximum amount of queries instead of the minimum.

BertrandBordage:
Finally! This work is fully ready to use. It replaces the old usage count system. Now, when you want to know which objects are related, simply try to delete without confirming. It will show any type of relation, including:
- a foreign key
- a many-to-many
- a generic foreign key
- a floating image in a rich text
- a document link in a rich text
- a link to a page in a rich text (when choosing a page, not pasting the URL)
- a root streamfield chooser block
- a nested streamfield chooser block
- a link to a page in a rich text nested in a streamfield
- et cetera

The `WAGTAIL_USAGE_COUNT_ENABLED` is therefore removed. No new setting is added, the list of related objects will always be shown, like in the Django admin.

This pull request fixes at least #687, #1162, #1602, #1722, #3010, #3082, #3854, #3886 & #3954.

## Here is what it looks like on **an image**:
capture d ecran de 2017-12-13 03-50-50
## On **a page**:
capture d ecran de 2017-12-13 03-51-23
## And of course, it finally gracefully handle **protected relations**!
capture d ecran de 2017-12-13 03-52-01
## It works with snippets
capture d ecran de 2017-12-13 03-52-22
## And it works exactly the same with `ModelAdmin`
capture d ecran de 2017-12-13 03-52-37
## It uses a pagination, making it way more efficient than the Django admin
capture d ecran de 2017-12-13 04-04-13

Performance could be improved, especially by rewriting the Django collector that generates way too many queries. But this is already really decent. Also, note that this performance cost is not as important as with the previous system, because the usage count system shown it when editing objects, which is a more common operation than trying to delete.

I remove the “Needs docs” label as it doesn’t seem useful now that the collectors are directly used inside the Wagtail admin. We may want to document the collectors or at least the `get_all_uses` function later, but we can keep it as a private API for now. The collector from Django is not documented, for the record.

tomdyson:
> This pull request fixes at least #687, #1162, #1602, #1722, #3010, #3082, #3854, #3886 & #3954

!

tomdyson:
Thanks for your initial review @chosak - could you please look through this again with @BertrandBordage's recent changes?

lb-:
@BertrandBordage I have run this PR locally (BakeryDemo) and have a few items I ran into that may need addressing. Note: I ran migrations after checking out the PR, this is not as much a code review but more reviewing the expected interaction.

I still have a bit more to test, I will see how I go later going through the rest of the items on my checklist.

## 1. Removal of `Used X Times` in Document/Image viewing

You mentioned that `WAGTAIL_USAGE_COUNT_ENABLED` will no longer be needed, as per documentation [here](http://docs.wagtail.io/en/v1.13.1/advanced_topics/settings.html#usage-for-images-documents-and-snippets).

However, one side effect of this is that the viewing of images, documents, snippets in admin no longer has the 'used X times' link at the top right. Also, the URL returns a 404 like `/admin/images/usage/53/`.

Maybe I missed something but it would be good to still have these helpful URLs and a quick reference to how much an item is used.

Can this still be enabled with this PR?

## 2. Delete Image view does not have image title
  * eg. /admin/images/52/delete/
  * Header title is 'Delete Image' but does not have the image title as the page subtitle
  * Where as Delete Document has the document title as the page subtitle

## 3a. Delete Image not recognising Snippet relation

  * When I go to delete an image /admin/images/53/delete/
  * It shows all the renditions, but does not show the Snippet using this image /admin/base/people/edit/3/
  * Tried removing/relinking the image and saving snippet, still not showing
  * Below is the ForeignKey for the image from the People snippet model

```
image = models.ForeignKey(
        'wagtailimages.Image',
        null=True,
        blank=True,
        on_delete=models.SET_NULL,
        related_name='+'
    )
```

## 3b. Delete Document not recognising Snippet relation

* Essentially the same issue, I added a Document ForeignKey to the People model in Bakerydemo and linked a document.
* I then went to delete the document and the snippet was not listed.

```
    document = models.ForeignKey(
        'wagtaildocs.Document',
        null=True,
        blank=True,
        on_delete=models.SET_NULL,
        related_name='+'
    )
```

## 4. Delete Snippet not recognising page relation

  * Attempting to delete a snippet /admin/snippets/breads/country/5/delete/
  * Page 53 is using it /admin/pages/53/edit/
  * Delete view only showing standard 'Are you sure you want to delete this country?', not showing the page it is linked to
  * BreadPage model ForeignKey to Country model
```
    origin = models.ForeignKey(
        Country,
        on_delete=models.SET_NULL,
        null=True,
        blank=True,
    )
```

## 5. Question: Tested with custom image or document model?

 * Did not test this one out but wanted to check it has been considered / will not make a difference.

## 6. Nicer text for when deleting form pages for related FormField 

This is likely out of the scope of this issue, maybe I can create a new issue.

  * When deleting a Form Page, you only get the following info in the row:
  * FormField object | Form field
  * Ideally it would be something like `FormField <name> <type>`
  * This is probably just an issue with the string representation in the abstract model for FormField

## 7. When deleting page, related page revision could have a link to the revision

Probably not that important, but might be nice to have links to the 'revision' when listing related revisions.

 * eg. a link to /admin/pages/100/revisions/111/revert/

## 8. Deleting site does not show 'root page' linked

* I created a new site
* Created a new home page and linked it to the site
* Delete view for the new home page shows that the site is linked
* However, attempting to delete the site does not show that the page is linked


## Checklist

### Manual Testing Checklist (Images)

* [X] Model: a foreign key
* [X] RichTextField: inserted into rich text
* [X] StreamField: a nested chooser block
* [X] StreamField: a root chooser block
* [X] StreamField: inserted into a richtextblock

### Manual Testing Checklist (Documents)

* [X] Model: a foreign key
* [X] RichTextField: inserted into rich text
* [X] StreamField: a nested chooser block
* [X] StreamField: a root chooser block
* [X] StreamField: inserted into a richtextblock

### Manual Testing Checklist (Pages)

* [X] Model: a foreign key
* [ ] RichTextField: inserted into rich text
* [ ] StreamField: a nested chooser block
* [ ] StreamField: a root chooser block
* [X] StreamField: inserted into a richtextblock
* [X] Linked to Site

### Other Testing Checklist

* [ ] Snippets
* [ ] ModelAdmin (non-Snippets)
* [ ] Sites

lb-:
Probably the major thought I have is relating to item 1 in my list.

I think that the admin usage URL should be added back in. I think this is a helpful tool for users as they may think it is a bit scary to click the Delete button to find out where things are used. It is also nice to have the small count at the top right when looking at an image or document.

Not sure what everyone else thinks though, plus the fact that this was previously behind a setting so maybe the adoption of this feature was minimal.

gasman:
Are there any _common_ use-cases for wanting to find out where an object is used, other than "I want to delete this but I need to check what it's going to break"? If there are, then yes, I agree that it's not very friendly to make those people go through the deletion view... if not, I think this approach is fine, so that we're not cluttering the edit interface with an infrequently-used feature.

It sounds like @BertrandBordage is concerned about the performance hit of running this query on every load of the edit page, since it's doing more work than the previous implementation. If so, perhaps one possible compromise is to keep the 'usage' view and link to it when `WAGTAIL_USAGE_COUNT_ENABLED` is true, but don't include the count as part of that link?

BertrandBordage:
Thanks very much for your review @lb-! Very helpful.

I’m working on improving the support for `SET_NULL`: this is the reason why you don’t see your related objects. It’s written “Used by”, but I’m using the Django collector which aim is to find “Will be deleted”. So when using `SET_NULL`, `SET_DEFAULT` & `DO_NOTHING`, the objects are not listed.

About the old usage stuff, it’s obvious to me that it has to be removed:
- It’s inaccurate, and therefore misleading. It only shows related pages, and it only works for images and documents.
- If we want to use the new collectors to fetch the correct count usage, the performance cost will be significant on big sites, and edit pages may take a few seconds to load.
- If we use the new collectors, the usage count will always be at least 1 for images and pages because of the original `Rendition` & the initial `PageRevision`. I know, this brings the question of whether we want to list renditions and revisions in the list of uses.
- If we want the usage count to be consistent, we have to implement it on all model types, not only documents & images. That means we would have to find a place in the page edit view to put the usage count, same for `ModelAdmin` etc.
- Users are not stupid, all Django admin users I worked with found by themselves that they could see which objects are related by trying to delete.
- What purpose does the usage count serve? To me it’s an inefficient way to find out “Oh, this file is not used anywhere, I should delete it”. An efficient way of doing that is a cleanup view where we would list all database objects that are not referenced by anything. And this view would have a few adjustments to take some relations away, like `Rendition` & `PageRevision`.

gasman:
> I know, this brings the question of whether we want to list renditions and revisions in the list of uses.

Ah, I missed that :-) Renditions should definitely be ignored in this view - their purpose is more as a cache than a user-facing model. Deleting them should have no consequence other than taking a bit longer to render on the next page load (which obviously isn't a concern for something you're deleting) and content editors absolutely shouldn't have to know or care what they are.

Page revisions, I think so too. At the database level `PageRevision.page` is a foreign key, but from the end user's perspective, the page revision isn't "a place where the page is being used" - it _is_ the page. (Originally, I thought the listing was showing old page revisions that reference a particular image/document etc - which might be potentially useful, but I'm not sure if that's even feasible, or how you'd stop it from massively cluttering up the listing.)

BertrandBordage:
@gasman: I agree from a user point of view, I left it this way to show the raw results for now, waiting for someone to bring up the topic ;)

gasman:
I see it's also including inline child objects and tag relations in the list:

screen shot 2017-12-14 at 12 57 44

I think it would be better to ignore all ParentalKey, ParentalManyToMany and *TaggableManager relations - again, from the content editor's perspective, these are conceptually _part of_ the object being deleted, not places where the object is being used.

BertrandBordage:
@gasman: I totally agree for the inline objects, but it will probably have to wait for a future pull request, as I doubt I can do that using the Django collector. I’ll try anyway.

BertrandBordage:
Following @lb-’s review, I improved this pull request. Now it shows all related objects, including those with other `on_delete` than `CASCADE` and `PROTECT`. And now it hides `Rendition`, `PageRevision` and inline objects, as discussed with @gasman :)

I’m also proud to announce that for the first time in Django history, users can see what will happen when deleting an object. I added a new label to get user attention on the `on_delete` behaviour. I tried to make it as clear and short as possible. Here are a few examples:

capture d ecran de 2017-12-14 18-12-45
capture d ecran de 2017-12-14 18-13-04
capture d ecran de 2017-12-14 18-13-58

Here are answers to @lb-’s review:
1. Discussed above
2. Not really related to this pull request, but fixed anyway.
3. Fixed, this was because `SET_NULL` relations are not returned by the Django collector.
4. Same as 3.
5. It works on any model, even custom images and documents. The links within the table also work with custom images and documents.
6. `FormField` is an inline, it is now hidden from the deletion view.
7. `PageRevision` instances are now hidden from the deletion view.
8. This is normal. Deleting only has an impact on reverse relations, not forward relations.
9. Same as 3.
10. Same as 3.
11. Same as 3.

BertrandBordage:
By the way, for related StreamField and RichText, I wrote that deleting this object will mark the field as empty (i.e. `SET_NULL` behaviour).
I know that nothing changes in database for related StreamField and RichText, but as far as I remember, removing a related object gives an empty field in the admin, right?

lb-:
@BertrandBordage @gasman I have tried to think of other use cases for showing the usage view / count beyond when needing deletion and cannot think of any for Images/Documents.

There is a case maybe for the navigation aspect in ModelAdmin, let's say you have a Person model and you want to be able to show the Person listing but also provide a link in each row to 'pages that use this Person'. The goal here is to navigate around Admin but not using the Explorer tool.

This could be reasonably easily reimplemented, but might be even easier if we kept the snippet model's `get _usage_url` method and the Admin url declaration.

Also - @BertrandBordage this is such an awesome feature, thanks so much!

lb-:
Just loaded up the latest version of this PR, looking awesome!

The behaviour on related StreamField and RichText is correct as you have indicated on the delete page.

Eg. When an image is deleted the resulting RichText field just shows up nothing, and StreamField or ImageChooserPanel just have empty links shown when you next edit the page that used them.

tmsndrs:
Hi i've just added this #4169 

I'm less concerned about showing how many times images and documents are used but it feels important to show that snippets are used in more than one place. It could be that we show a warning to say it's used in more than one place with a link to the full list to avoid performance issues. 

Any thoughts? 

Also I think this needs some UI design work. I can prioritise this if there is urgency?

BertrandBordage:
@tmsndrs: This will still have an very important impact on performance for now. The first thing I check are related objects through foreign keys, many-to-many and generic relations. To do that, I use the Django Collector class that fetches all related objects at once, triggering dozens to hundreds of queries. That class would have to be written again to use generators instead, which would be a significant effort, especially to test.

So it’s not a definitive “no”.

BertrandBordage:
Also @tmsndrs, feel free to submit a mockup of what you have in mind in terms of UI :)

gasman:
Looks like the rich text handling here will need to be substantially reworked now that #4079 is merged :-(

Unfortunately c2938602adeb639d7511048541ae17fe54e874b4 is already pushing the limit of what's feasible to review in one go - I think this could really do with splitting up into multiple PRs:

* Refactoring rich text (and streamfield, anything else...?) to add supporting methods like `LinkHandler.get_instance` as required by collectors
* Adding the collectors
* UI updates

Shall we bump this to 2.1? (Now that I've seen the full scope of the changes here, it feels like another major codebase-spanning refactor is a bit too much for 2.0, and this way we get some breathing room for @tmsndrs to give it some proper design attention...)

tomdyson:
@tmsndrs:

> I can prioritise this if there is urgency?

yes please

ababic:
@BertrandBordage excellent work here (as usual). 

I have some thoughts regarding one of the decisions made above though, which is the decision to always exclude `ParentalKey` relationships (and to a lesser extent `ParentalManyToMany`). While it's true that `ParentalKey` and the `InlinePanel` were designed for pages, we must recognise that it is still currently the only way to achieve an 'inline' editing UI using `panels`, and therefore, it is incredibly common for models to subclass `modelcluster.models.ClusterableModel` and use `ParentalKey` in order to use `InlinePanel`. For example, [wagtailmenus](https://github.com/rkhleics/wagtailmenus/blob/master/wagtailmenus/models/menus.py#L851) takes this approach. 

With this in mind, can we really be sure these relationships should never be surfaced in this interface? Is there any way we might cater for cases where including them would be preferred?

gasman:
@ababic I think there's a similar argument to https://github.com/wagtail/django-modelcluster/issues/89#issuecomment-363515427 to be made here: `ParentalKey` is inherently an "X belongs to Y" relation rather than "X contains a reference to Y".

Just as there's no conceivable reason _not_ to use `on_delete=CASCADE` on a ParentalKey (and this isn't just a case of "I can't think of a reason"; rather, it's basically implied in the design of ParentalKey that CASCADE is the only logical choice), I'm confident that there's no scenario where it's appropriate to bring up a ParentalKey relation to X as "a place where X is being used".

Anything involving InlinePanel certainly shouldn't be surfaced here: it's redundant to tell the user "deleting this page will cause the 'see also' links on the page to be deleted", or "deleting this menu will cause its submenus to be deleted". (For the casual user who isn't thinking in terms of database schemas, it most likely doesn't even occur to them that the thing in an InlinePanel is an independent object that _could_ exist outside of its container - to them, it's just a repeating field.)

gasman:
Sorry, I really should have updated the 'needs review' / 'needs work' labels when I added https://github.com/wagtail/wagtail/pull/3961#issuecomment-356634922 ... I thought this was still on my review queue, but in fact it still needs the rich text bits to be rewritten following #4079, and - ideally - being split up into smaller PRs. (There's a lot going on in c2938602adeb639d7511048541ae17fe54e874b4 in particular, and we need to make sure that breaking changes such as changing the API of `register_rich_text_embed_handler` / `register_rich_text_link_handler` - or their Wagtail 2.0 equivalents - don't slip under the radar.)

ababic:
Thanks for the feedback @gasman. I'm with you now. 

Things are slightly muddied here, as we're dealing with surfacing PROTECTED relationships as well as those that would be effected by a delete operation were it to happen. I was thinking somehow that ignoring `ParentalKey` relationships would prevent say, a menu item linking to a `Document` via a PROTECTED relationship being flagged up when attempting to delete that `Document`. But, such relationships would likely use a standard `ForeignKey` field, and so would still appear as expected. 

I'll shut up now :)

BertrandBordage:
Superseded by #4481.

PR #3963, opened 2017-10-21
Removed empty super().__init__ constructs.

atombrella:


wagtail/wagtailsearch/backends/db.py
```diff
@@ -89,9 +89,6 @@ class DatabaseSearchBackend(BaseSearchBackend):
     query_class = DatabaseSearchQuery
     results_class = DatabaseSearchResults
 
-    def __init__(self, params):
```
atombrella: The call would go back to `pass`.

gasman:
Thanks @atombrella!

PR #3965, opened 2017-10-24
Fixes admin dropdown arrow.

BertrandBordage:
This has been bugging me for a while.

Before, on Chrome:
image
After, on Chrome:
image
There was a useless background rule on the `select + span::after` arrow, which was sometimes overlapping with the `select` border, probably because of some non-integer pixel height.

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_forms.scss
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/scss/components/_forms.scss
@@ -117,7 +117,6 @@ select,
         font-size: 3em;
         pointer-events: none;
         color: $color-grey-3;
-        background-color: $color-fieldset-hover;
         margin: 0 1px 1px 0;
 
         .ie & {


kaedroho:
@BertrandBordage I can't reproduce the issue, did this only affect specific dropdowns or did you see this on all of them?

BertrandBordage:
It’s only on Chrome, as far as I can tell.
It depends on the screen size and the layout, especially on large screens. Sometimes all dropdowns are OK, sometimes none, or a few of them.
The vertical border can also be affected:
capture d ecran de 2017-10-30 11-54-22

To force the issue to appear, zoom or unzoom your page in Chrome, it should appear more often as Chrome has some issues with pixel approximations while zooming.

I thought this change had no effect on the current styling, but turns out there is a small effect when hovering the select: the left part of the select gets a white background while the arrow keeps the grey background. With this change, the arrow part also gets a white background on hover, which is fine too.

heymonkeyriot:
Even though this is super small it cropped up in some of the recent user testing about understandability of the select field inputs so cool to have it fixed!

BertrandBordage:
@heymonkeyriot Since you know a lot about the admin CSS, could you review this 1-line pull request please? :p

heymonkeyriot:
It looks good to me testing here. Hopefully one of the core team can merge in?

PR #3968, opened 2017-10-24
Fix using modeladmin_register as a decorator

timheap:
Using `modeladmin_register` as a decorator worked, and the ModelAdmin subclass was registered successfully, but as `modeladmin_register` did not return the ModelAdmin subclass, the decorated class would be `None`. Returning the class prevents this issue, so `modeladmin_register` will now work as expected as a decorator.


BertrandBordage:
Agreed, thank you Tim!

PR #3970, opened 2017-10-25
Moved SearchableListMixin to wagtailadmin

kaedroho:
And add wagtailsearch support.

As suggested by @ababic in https://github.com/wagtail/wagtail/pull/3774#discussion_r141448118

### DIFF
--- a/wagtail/wagtailadmin/views/mixins.py
+++ b/wagtail/wagtailadmin/views/mixins.py
@@ -12,10 +12,7 @@ class SearchableListMixin(object):
     search_fields = None
 
     def get_search_form(self):
-        if self.request.GET.get('q'):
-            return SearchForm(self.request.GET, placeholder=self.search_box_placeholder)
-        else:
-            return SearchForm(placeholder=self.search_box_placeholder)
+        return SearchForm(self.request.GET if self.request.GET.get('q') else None, placeholder=self.search_box_placeholder)
 
     def get_queryset(self):
         queryset = super(SearchableListMixin, self).get_queryset()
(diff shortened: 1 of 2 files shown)


wagtail/wagtailadmin/views/mixins.py
```diff
@@ -0,0 +1,45 @@
...
+                queryset = search_backend.search(q, queryset, fields=self.search_fields)
+            else:
+                filters = {
+                    field + '__icontains': q
+                    for field in self.search_fields or []
+                }
+
+                queryset = queryset.filter(**filters)
+
+        return queryset
```
BertrandBordage: Defining `queryset` is useless here, we can use `return` directly and avoid the `else:`.

wagtail/wagtailadmin/views/mixins.py
```diff
@@ -0,0 +1,45 @@
...
+
+class SearchableListMixin(object):
+    search_box_placeholder = _("Search")
+    search_fields = None
+
+    def get_search_form(self):
+        if self.request.GET.get('q'):
+            return SearchForm(self.request.GET, placeholder=self.search_box_placeholder)
+        else:
+            return SearchForm(placeholder=self.search_box_placeholder)
```
BertrandBordage: To reduce redundancy, can be replaced with:
```python
def get_search_form(self):
    return SearchForm(self.request.GET if self.request.GET.get('q') else None,
                      placeholder=self.search_box_placeholder)
```

wagtail/wagtailadmin/views/mixins.py
kaedroho: I'm not sure what you mean, this method should always return a queryset.

kaedroho:
Thanks for the review @BertrandBordage!

wagtail/wagtailadmin/views/mixins.py
BertrandBordage: Yes, it returns a queryset, but it does not have to define a `queryset` variable, querysets can be returned directly.

PR #3971, opened 2017-10-25
Improved 'purge_page_from_cache' to purge the url for any managed language

kaedroho:
Rebase of https://github.com/wagtail/wagtail/pull/3024 with some cleanups

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -5,6 +5,7 @@ Changelog
 ~~~~~~~~~~~~~~~~
 
  * Removed support for Python 2.7, Django 1.8 and Django 1.10
+ * `purge_page_from_cache` now purges the url of each managed language when using `i18n_patterns` (PyMan Claudio Marinozzi)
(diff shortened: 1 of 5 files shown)


kaedroho:
I noticed that in testing that, by default, this purges URLs for 86 different languages. I think we should make this feature disabled by default and you opt in to it by setting the specific languages you want to purge in the ``WAGTAILFRONTENDCACHE`` setting

pyMan:
@kaedroho I'll be happy to go on on this, but I cannot work on it before than two months at least

pyMan:
@kaedroho do you prefer a common settings for each defined backend or a specific setting for each backend?

kaedroho:
@pyMan I prefer a common setting for all backends. So perhaps a separate setting is the best way to go.

pyMan:
@kaedroho I should be done

gasman:
Merged in b3fa09b95ed7474a3307a6afa317e0d504e2fc86 - thanks @pyMan and @kaedroho!

PR #3972, opened 2017-10-25
Add setting to disable notification to superusers -- Fixes #3969

### LINKED ISSUE #3969: New setting to exclude superusers from moderation notification
### Issue Summary

I'm deploying Wagtail in my Web app to help us manage our static content, and I was testing the moderation flow and while testing it I fired some emails to all our super users, due to [this function](https://github.com/wagtail/wagtail/blob/5c0fc1197b4b4f0f9ded4abc1bb260d4d6b66723/wagtail/wagtailadmin/utils.py#L100-L111). What we would like is only contact moderators for that, as super-users roles are attributed to developers and usually won't be the ones approving.

### Suggested solution

Would it be possible to add a general switch to optionally disable this behaviour? I was thinking of a new setting `WAGTAILADMIN_INCLUDE_SUPERUSERS_NOTIFICATIONS` (or whatever name makes sense) that would be ON by default (for backwards compatibility), but that one could turn off.

Apologies if such feature already exists, it might be worth advertising it more by updating the documentation.

I can put together a pull request if that's an acceptable solution.

browniebroke:
New setting `WAGTAILADMIN_NOTIFICATION_INCLUDE_SUPERUSERS`, default to `True`. When set to `False`, superusers do not receive moderation notification, only moderators do.

### DIFF
--- a/docs/advanced_topics/settings.rst
+++ b/docs/advanced_topics/settings.rst
@@ -292,6 +292,14 @@ Email Notifications format
 
 Notification emails are sent in `text/plain` by default, change this to use HTML formatting.
 
+Email Notifications to superusers
+---------------------------------
+
+.. code-block:: python
+
+  WAGTAILADMIN_NOTIFICATION_INCLUDE_SUPERUSERS = False
+
+Notifications emails are sent to moderators and superusers by default, you can change this to exclude superusers and only email moderators.
 
 .. _update_notifications:
 
--- a/docs/advanced_topics/settings.rst
+++ b/docs/advanced_topics/settings.rst
@@ -299,7 +299,7 @@ Email Notifications to superusers
 
   WAGTAILADMIN_NOTIFICATION_INCLUDE_SUPERUSERS = False
 
-Notifications emails are sent to moderators and superusers by default, you can change this to exclude superusers and only email moderators.
+Notification emails are sent to moderators and superusers by default. You can change this to exclude superusers and only notify moderators.
 
 .. _update_notifications:
 
--- a/wagtail/wagtailadmin/tests/test_pages_views.py
+++ b/wagtail/wagtailadmin/tests/test_pages_views.py
@@ -3188,7 +3188,7 @@ class TestNotificationPreferences(TestCase, WagtailTestUtils):
         # Should be redirected to explorer page
         self.assertEqual(response.status_code, 302)
 
-        # Check that the non-moderators superusers are not being notified
+        # Check that the non-moderator superuser is not being notified
         expected_emails = 1
         self.assertEqual(len(mail.outbox), expected_emails)
         # Use chain as the 'to' field is a list of recipients
--- a/wagtail/wagtailadmin/tests/test_pages_views.py
+++ b/wagtail/wagtailadmin/tests/test_pages_views.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import, unicode_literals
 import datetime
 import logging
 import os
+from itertools import chain
 
 import mock
 from django.conf import settings
@@ -15,7 +16,7 @@ from django.core.files.base import ContentFile
 from django.core.urlresolvers import reverse
 from django.db.models.signals import post_delete, pre_delete
 from django.http import HttpRequest, HttpResponse
-from django.test import TestCase, modify_settings
+from django.test import TestCase, modify_settings, override_settings
 from django.utils import formats, timezone
 from django.utils.dateparse import parse_date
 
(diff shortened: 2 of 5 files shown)


gasman:
Thanks @browniebroke! I was just about to reply on #3969, but it looks like you beat me to it :-)

I'm happy with this approach. I'm not 100% keen on 'superuser' being treated as a special case - ideally superuser would just be a shorthand for a user with full permission over everything, and any actual change in behaviour would be implemented through permission rules. (A good example of this is the 'bulk delete' permission on pages - normally this is something you'd only want to be available to superusers, but by making it an actual permission instead of hard-coding an `is_superuser` check, it leaves open the possibility of setting up an "almost-superuser" level, e.g. someone with complete administrative rights over one specific site on a multi-site Wagtail installation.) However, in this case it's hard to see how we'd go about expressing this flag in terms of "permissions", so I'm happy to adopt this as a working solution.

docs/advanced_topics/settings.rst
```diff
@@ -292,6 +292,14 @@ Email Notifications format
...
 Notification emails are sent in `text/plain` by default, change this to use HTML formatting.
 
+Email Notifications to superusers
+---------------------------------
+
+.. code-block:: python
+
+  WAGTAILADMIN_NOTIFICATION_INCLUDE_SUPERUSERS = False
+
+Notifications emails are sent to moderators and superusers by default, you can change this to exclude superusers and only email moderators.
```
tomdyson: Notifications emails -> Notification emails
default, you can -> default. You can

wagtail/wagtailadmin/tests/test_pages_views.py
```diff
@@ -3177,6 +3178,24 @@ def test_moderator_group_notifications(self):
...
+    def test_disable_superuser_notification(self):
+        # Add one of the superusers to the moderator group
+        self.moderator.groups.add(Group.objects.get(name='Moderators'))
+
+        response = self.submit()
+
+        # Should be redirected to explorer page
+        self.assertEqual(response.status_code, 302)
+
+        # Check that the non-moderators superusers are not being notified
```
tomdyson: non-moderators -> non-moderator

browniebroke:
Thanks for the quick review!
As a side note, I presented Wagtail today in my office and everyone was very impressed. Great work on this project!

docs/advanced_topics/settings.rst
```diff
@@ -292,6 +292,14 @@ Email Notifications format
...
 Notification emails are sent in `text/plain` by default, change this to use HTML formatting.
 
+Email Notifications to superusers
+---------------------------------
+
+.. code-block:: python
+
+  WAGTAILADMIN_NOTIFICATION_INCLUDE_SUPERUSERS = False
+
+Notification emails are sent to moderators and superusers by default. You can change this to exclude superusers and only notify moderators.
```
kaedroho: Perhaps we could mention that superusers who do really want to get emails can also be added to the "Moderators" group?

gasman: Hmm... I'm concerned that documenting that as a feature, rather than it simply being a current implementation quirk, might be taking us down a bit of a murky road...

Up to now, Wagtail has never properly addressed the distinction between "publish" as a permission and "moderation" as a role. Technically, Wagtail has no such concept as "a moderator"; there's a group called "Moderators" in the out-of-the-box setup, whose sole distinguishing feature is that they have 'publish' permission, but this is just a convention, and it's entirely valid to delete, rename or edit that group. Essentially, describing a user as a moderator is a convenient fiction, that (hopefully) helps people to understand the workflow system, but in reality it just means "someone who has publish permission over the given page". This equivalence breaks down when we have users such as developers, who have publish permission but shouldn't really be assigned the _role_ of moderator.

This PR is a first pass at addressing that issue, relying on the assumption that the users in that category will be superusers. This isn't perfect - as mentioned at https://github.com/wagtail/wagtail/pull/3972#issuecomment-339424381, I want to leave open the possibility of having "almost-superuser" groups with an administrative role - but it's better than what we have now.

Since "the Moderators group" is not a fixed thing, making the statement "superusers who are in the Moderators group will receive notifications" really unpacks as:

* Users who have received publish permission through _being members of a group_ will receive notifications;
* Users who have received publish permission through _being superusers_ will not receive notifications.

The fact that publish permission has different effects depending on how you acquired it is a bug, not a feature. I think the only real way to solve it _properly_ is to make the role of "moderation" a real thing that's distinct from "having publish permission". That's not going to be solved in the short term, and until we do, I think we should avoid making promises about its behaviour at that level of detail.

browniebroke: Hi guys. I appreciate your concerns regarding making it a feature that people might rely upon and that it might make it harder to drop later.

As a superuser of a small scale Web app, where we wanted to use Wagtail to manage our static pages, I was surprised to see that content approval was including me by default. Maybe it was mostly a documentation issue? I write the code, not the content, and English isn't my first language.

From reading this thread, I'm under the impression that @gasman is leaning towards removing this 'notify super-users' and rely only upon permissions & group for content review, am I getting this right? 

I'm more than happy to change the implementation of this pull request and I would also understand if you decline this change and leave the code as is.

Whatever you decide, please ping me if you would like any action from me on this.

hakjoon: This is something were are also running into as the group that should be empowered with approving moderation is actually small but we have admins and others who have publish permissions, but should not be approving requests from users who need moderation.

What are thoughts on perhaps having a setting that exposes what groups should receive moderation messages, instead of just excluding superusers?  It seems like it would work to solidify the moderator role.

gasman:
Merged in be164048b84754480ba6fe175449329e7f65b198 - this is a useful addition regardless of the more 'philosophical' issues raised here. Thanks @browniebroke!

PR #3974, opened 2017-10-26
Sitemap: Fix to tolerate sites where all pages are private

strindhaug:
When all pages on a site is private, the `last_mods` set will be empty
causing a crash when trying to get the max value from it.
This fix adds a test that the set is truthy (not empty) before calling
max.

Without this fix the sitemaps view wil crash when all pages on the site are private.
Whith this fix applied, the sitemap will be an empty sitemap as expected:

`<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9"></urlset>`


kaedroho:
Thanks!

PR #3975, opened 2017-10-26
Deprecate a bunch of no longer used parts of Wagtailsearch

kaedroho:


gasman:
@kaedroho Please can you add an 'upgrade considerations' note for this change to the 2.0 release notes? (Or alternatively, confirm that a note isn't needed, e.g. because this is all internal/undocumented functionality...)

gasman:
Merged in 0688be15431641d0bcf43ebc76c6a7e002a3cdc5 / 9751f62b22430efb627df6f4bf73f57ba3406648

gasman:
@kaedroho Now that the time has come to remove the deprecated code (#4551), I'm a bit unclear on what the plan was for the methods being removed from `Elasticsearch2SearchBackend` (`add_type`, `refresh_index`, `add`, `add_bulk`, `delete`)...

* Contrary to what this PR claims, `backend.add` is still in use in several places in the Wagtail codebase (in tests, and the multiple image/document uploaders as a workaround for tags not getting indexed)
* If these are indeed being removed from the search backend API, shouldn't we have added the deprecation warnings to the other backends too?
* The replacement code suggested in the deprecation warnings (e.g. `backend.add(obj)` -> `backend.get_index_for_model(type(obj)).add_item(obj)`) will fail on the DB backend, where `get_index_for_model` returns `None`. Surely we shouldn't be encouraging users to write backend-specific code...
* Isn't `backend.add(obj)` a far nicer API to present to users (both for readability and being at the right abstraction level)?

I'm wondering if the real intention here was to move these wrapper functions into `BaseSearchBackend` (to take advantage of the similarity between this and the Postgres backend) - please can you clarify?

kaedroho:
> The replacement code suggested in the deprecation warnings (e.g. backend.add(obj) -> backend.get_index_for_model(type(obj)).add_item(obj)) will fail on the DB backend, where get_index_for_model returns None. Surely we shouldn't be encouraging users to write backend-specific code...

Hadn't thought of this. We should add a dummy index class for the database backend so it doesn't break.

> I'm wondering if the real intention here was to move these wrapper functions into BaseSearchBackend (to take advantage of the similarity between this and the Postgres backend) - please can you clarify?

These methods were created when each backend could manage only one index. This changed when we added support for Elasticsearch 2 so we had to create another API and some older code still uses the previous API.

The intention of the deprecation is to return to having one way of inserting items into an index, this would be to simplify the implementation of search backends.

As you said, I now think it would be better to move these into ``BaseSearchBackend`` and make them call the methods on the index class (after we've fixed it for the database backend). So there wouldn't be any need to deprecate them.

PR #3979, opened 2017-10-27
Makes `PageChooserPanel` search in all fields.

BertrandBordage:
The current behaviour is inconsistent with [images](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailimages/views/chooser.py#L72) and [documents](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtaildocs/views/chooser.py#L62).

It’s particularly blocking for sites with a ton of pages where user prefer to search (for example), by primary key.


kaedroho:
Thanks!

This was also the behaviour on the main page search until we changed it a couple of years ago, I think the same rationale applies here.

https://github.com/wagtail/wagtail/pull/2279

PR #3983, opened 2017-10-30
Only show "View live" buttons if pages are accessible

### LINKED ISSUE #3982: Flash message "View live" button is broken if a page has no URL
### Issue Summary

Editing pages in the admin that are not under a site - and therefore do not have a live url - result in a broken 'View live' button in the flash message.

### Steps to Reproduce

Create a page using the admin that is *not* under a site root:

1. In the admin, click the "Pages >" menu item, then click the "Pages" header to navigate to the root of the page tree.
2. Click "Add child page" in the header, and create any page here.
3. After the page has been created, a flash message with a "View live" button will be shown. Clicking the "View live" button will navigate to a 404 page.

### Suggested fix

Only show the "View live" button if `page.url is not None`

### Technical details

* Python version: All versions
* Django version: All versions
* Wagtail version: 1.13
* Browser version: All browsers

timheap:
"View live" buttons in flash messages are now only displayed if the page has a live URL. Editing pages that do not have a URL will no longer produce broken buttons.

Fixes #3982

### DIFF
--- a/wagtail/wagtailadmin/messages.py
+++ b/wagtail/wagtailadmin/messages.py
@@ -62,4 +62,6 @@ def validation_error(request, message, form, buttons=None):
 
 
 def button(url, text, new_window=False):
+    if url is None:
+        raise ValueError("Button URLs must not be None")
     return url, text, new_window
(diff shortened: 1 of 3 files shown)


kaedroho:
Looks good to me, thanks!

tomdyson:
@timheap it's great to have you back!

PR #3984, opened 2017-10-31
Filter pop-out explorer pages using 'construct_explorer_page_queryset'

timheap:
The pages listed when browsing the /admin/pages/ explorer differed from the pages listed in the new React/admin API powered pop-out explorer. The latter did not pass the queryset through the 'construct_explorer_page_queryset' hook, so pages that should have been hidden were visible. Visiting these pages in the explorer could then lead to unexpected behaviours, as hidden sections of the site became browsable.

A new `for_explorer=1` query parameter has been added to the admin API, which will pass the page queryset through the 'construct_explorer_page_queryset' hooks.

### DIFF
--- a/client/src/api/admin.js
+++ b/client/src/api/admin.js
@@ -10,7 +10,7 @@ export const getPage = (id) => {
 };
 
 export const getPageChildren = (id, options = {}) => {
-  let url = `${ADMIN_API.PAGES}?child_of=${id}`;
+  let url = `${ADMIN_API.PAGES}?child_of=${id}&for_explorer=1`;
 
   if (options.fields) {
     url += `&fields=${global.encodeURIComponent(options.fields.join(','))}`;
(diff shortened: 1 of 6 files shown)


gasman:
Thanks @timheap! Merged in 1f2b8ccaf18a2f097c19d831e679b78b4c6a8bd1.

(The need for the `for_explorer` parameter really bugs me - it undermines the point of having a standard API if the consumers of it identify themselves as "the explorer" or otherwise and get different behaviour. But I can also see that it's a logical consequence of `construct_explorer_page_queryset` and `exclude_from_explorer`, so my discontent with it is outside the scope of this PR :-) I'll just have to make it my personal mission for Wagtail 2.1 to deprecate `construct_explorer_page_queryset` in favour of introducing some formal concept of "hidden pages"...)

PR #3990, opened 2017-11-03
Amends label on ImageChooserBlock

### LINKED ISSUE #3955: ImageChooserBlock label language to change image
image

Currently to change the image within a StreamField using ImageChooserBlock says `Choose another image`. This is potentially confusing as the user is changing the image and `another` is ambiguous in English (it could be 'to change' or 'add an additional image').

I think it would be clearer if it were to say `Change image`.

heymonkeyriot:
Very small change :)

But will close #3955 from an English-language POV.


gasman:
Thanks @heymonkeyriot! Merged in a28fa5e8c892fbe158e3c2f406de88bd7c571e4e.

Issue #3104, opened 2016-10-24
snippet for a model with custom primary key


Eraldo:
### Issue Summary

Snippets cannot yet handle models with custom primary keys.
### Steps to Reproduce
1. Create a model using a custom primary key. (Thus there is no 'id' field anymore.)
2. Register that model as a snippet. (@register_snippet)
3. Got to wagtail cms backend and check out the snippet menu of the registerd snippet.
4. An error is displayed in DEBUG mode.

I expected a listing of the snippet models, further being able to edit the respective model instances.
### Technical details

Source:
https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailsnippets/models.py#L25
There might be more places where this is an issue.

Error:
`Reverse for 'edit' with arguments '('myapp', 'mymodel', '')' and keyword arguments '{}' not found. 1 pattern(s) tried: ['cms/snippets/(\\w+)/(\\w+)/(\\d+)/$']`
- Python version: Python 3.5.1.
- Django version: Django (1.9.9).
- Wagtail version: wagtail (1.7)
- Browser version: Chrome 54.0.2840.71 (64-bit).

Regards,
Eraldo

evert-smit:
Thanks for the summary @Eraldo.

@gasman I'd like to see if I could make a contribution to Wagtail by changing this behaviour. Is there anything I need to keep in mind other than the changes I would make to this particular part of the codebase? I'd like to get the solution right, to avoid sending an unstructurized pull request.

gasman:
Thanks @evert-smit! I think the main thing to watch out for with non-numeric IDs is that they may contain characters that aren't valid in URLs, and so you'll need to ensure that they're escaped when building the URL and un-escaped at the other end. (I think Django's admin package has some code you can borrow...)

Other than that, just ensure that your changes are covered by tests (you'll probably want to add a new model definition to https://github.com/wagtail/wagtail/blob/master/wagtail/tests/snippets/models.py) and follow the [coding guidelines](http://docs.wagtail.io/en/v1.12/contributing/python_guidelines.html) - good luck! :-)

saevarom:
At my company we're currently working on a project with custom primary keys. I could work on this on company time, but I'm wondering if @evert-smit has an ongoing solution that I could contribute to? If not, or if it's inactive or in other ways not available, I'm going to start working on a PR for this.

evert-smit:
That would be fine - I planned working on this but ended up going in a
different direction and didn’t have time to continue with a PR.

PR #4003, opened 2017-11-08
Update snippets so that they can handle custom primary keys.

saevarom:
This is a fix for issue #3104 which describes problems with snippet models having custom primary keys. The wagtailsnippets app mostly assumes an id field everywhere which breaks almost all snippets views and operations.

I've tried to test this as extensively as possible, mostly re-using tests that had been created for the regular snippets, simply replacing the models and fixtures with new ones using custom primary keys. I had to change the URL structure of the snippets app, since the primary keys can include any characters, including `/`. I copied Django's behaviour for quoting and unquoting the primary keys themselves, but found no good way to build the URLs, other than simply allowing all characters (excluding `/`).

Comments and suggestions appreciated :)


saevarom:
Some tests are still failing. I'll investigate.

wagtail/wagtailsnippets/templates/wagtailsnippets/snippets/confirm_delete.html
```diff
@@ -1,5 +1,5 @@
 {% extends "wagtailadmin/base.html" %}
-{% load wagtailadmin_tags i18n %}
+{% load wagtailadmin_tags i18n admin_urls %}
```
gasman: We can't rely on user projects having `django.contrib.admin` in `INSTALLED_APPS`, and if it isn't, the tag library will fail to load. We should probably import or copy the `admin_urlquote` filter into wagtailadmin_tags instead...

gasman:
Rebased (+ tag imports fixed as per https://github.com/wagtail/wagtail/pull/4003#pullrequestreview-86423704) at #4166

gasman:
Merged via #4166 - thanks @saevarom!

PR #4004, opened 2017-11-09
Eliminate some django.utils.six references

gasman:


kaedroho:
LGTM!

We should also replace references to ``django.utils.lru_cache`` with ``functools.lru_cache``

PR #4005, opened 2017-11-09
Add documentation for custom send_mail method

### LINKED ISSUE #2954: Add an example on how to override the AbstractEmailForm .send_mail method
Would be great to add an example on how to override the `AbstractEmailForm.send_mail` method which was introduced in #2926.

Docs for `wagtail.wagtailforms` can be found [here](http://docs.wagtail.io/en/v1.6/reference/contrib/forms.html).

(cc @tleguijt)

lb-:
Closes #2954 

Tested docs - `make html` and reviewed, just want to ensure this matches the way we want docs written.

### DIFF
--- a/docs/reference/contrib/forms/customisation.rst
+++ b/docs/reference/contrib/forms/customisation.rst
@@ -553,3 +553,64 @@ Next, you need to transform your template to display the results:
 
 
 You can also show the results on the landing page.
+
+
+Custom ``send_mail`` method
+---------------------------
+
+If you want to change the content of the email that is sent when a form submits you can override the ``send_mail`` method.
+
+
+To do this, you need to:
+
+* Ensure you have your form model defined that extends ``wagtail.wagtailforms.models.AbstractEmailForm``.
+* In your models.py file, import the ``wagtail.wagtailadmin.utils.send_mail`` function.
+* Override the ``send_mail`` methodd in your page model.
+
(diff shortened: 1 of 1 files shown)


docs/reference/contrib/forms/customisation.rst
```diff
@@ -553,3 +553,64 @@ Next, you need to transform your template to display the results:
...
+---------------------------
+
+If you want to change the content of the email that is sent when a form submits you can override the ``send_mail`` method.
+
+
+To do this, you need to:
+
+* Ensure you have your form model defined that extends ``wagtail.wagtailforms.models.AbstractEmailForm``.
+* In your models.py file, import the ``wagtail.wagtailadmin.utils.send_mail`` function.
+* Override the ``send_mail`` methodd in your page model.
```
timorieber: Typo "method"

lb-: Nice pickup, thanks.

lb-: Will fix up asap.

lb-: updated

loicteixeira:
Reading the documentation just feels like we should extract the generation of the email body to a separate function and leave `send_mail` alone. But anyway, for the time being, it is good to have it documented 👌

PR #4011, opened 2017-11-11
Docs - Page Model Reference - add note about django-treebeard api

### LINKED ISSUE #2779: Getting parent page
I would like to be able to get the parent page of a child page. Maybe there is a way already, but I haven't been able to find it in the documentation.

Is this already possible and if not, can this be added?

lb-:
Updates to documentation to resolve last item in issue.

Closes #2779


gasman:
Thanks @lb-!

Issue #3834, opened 2017-09-05
Wagtail SQL call on none wagtail pages


sowinski:
Hi,

I have included wagtail to my project as a cms system. But I have also some old static pages which also produce allways this SQL call:

`
SELECT"wagtailcore_site"."id", "wagtailcore_site"."hostname", "wagtailcore_site"."port", "wagtailcore_site"."site_name", "wagtailcore_site"."root_page_id", "wagtailcore_site"."is_default_site", CASEWHEN ("wagtailcore_site"."hostname" = '''127.0.0.1''' AND "wagtailcore_site"."port" = '8000') THEN '0' WHEN ("wagtailcore_site"."hostname" = '''127.0.0.1''' AND "wagtailcore_site"."is_default_site" = 'True') THEN '1' WHEN "wagtailcore_site"."is_default_site" = 'True' THEN '2' ELSE '3' END AS "match", "wagtailcore_page"."id", "wagtailcore_page"."path", "wagtailcore_page"."depth", "wagtailcore_page"."numchild", "wagtailcore_page"."title", "wagtailcore_page"."slug", "wagtailcore_page"."content_type_id", "wagtailcore_page"."live", "wagtailcore_page"."has_unpublished_changes", "wagtailcore_page"."url_path", "wagtailcore_page"."owner_id", "wagtailcore_page"."seo_title", "wagtailcore_page"."show_in_menus", "wagtailcore_page"."search_description", "wagtailcore_page"."go_live_at", "wagtailcore_page"."expire_at", "wagtailcore_page"."expired", "wagtailcore_page"."locked", "wagtailcore_page"."first_published_at", "wagtailcore_page"."last_published_at", "wagtailcore_page"."latest_revision_created_at", "wagtailcore_page"."live_revision_id"FROM "wagtailcore_site" INNER JOIN "wagtailcore_page" ON ("wagtailcore_site"."root_page_id" = "wagtailcore_page"."id") WHERE ("wagtailcore_site"."hostname" = '''127.0.0.1''' OR "wagtailcore_site"."is_default_site" = 'True') ORDER BY "match" ASC | 98,487335358% | 144,35 | SelExpl
  | 
`
This call takes on my desktop computer (in debug mode with sqlite) 150ms. Is there something wrong?
And why is this SQL query also called on a none wagtail page?

### Issue Summary

Wagtail is calling SQL queries which I don't need

gasman:
[...]

Issue #2840, opened 2016-07-13
wagtailcore.models.Site should probably inherit from django.contrib.models.Site


ppershing:
I thing it would make sense for Wagtail Site to inherit from Django Site. Right now Django's sites are independent of Wagtail's which makes a huge configuration mess for some third-party Django apps relying on Django sites but not Wagtail sites.

gasman:
Hi @ppershing,
There's not enough detail here to evaluate this suggestion, sorry. If you can give a description of the specific problem you're having, and how inheriting from `django.contrib.models.Site` would fix it, then we'd be happy to consider this further.

ppershing:
The specific problem is probably just more nuisance. 
Because Wagtail sites and Django sites are completely independent (and the latter is not visible in the Wagtail admin), it takes quite some time to track down issues where third party apps use "example.com" as a domain for various tasks. This happened to me with ```django-comments-xtd``` but I can see that the same thing will happen for anybody using ```django-registration``` or generally any other Django app that uses sites framework.

timheap:
I have also experienced this problem. We solved it by writing code that syncs Wagtail sites to Django sites every time a Wagtail site is modified. Making Wagtail sites extend Django sites via MTI would ensure they are kept in sync, and third party apps will work as expected.

pjmcdermott:
I'm not sure I would characterise it a mere nuisance! I must have spent the past day trying to work out why installing Wagtail impacted functionality on other parts of my site...

`wagtail.wagtailcore.middleware.SiteMiddleware` is fundamentally incompatible with `django.contrib.sites.middleware.CurrentSiteMiddleware` as they both set `request.site`. But the Wagtail `Site` object does not have the same field names as the Django `Site` object. Specifically, the Django `Site` model uses `domain` and `name`, whereas the Wagtail `Site` model uses `hostname` and `site_name`.

No warnings are given when the clash occurs--functionality just disappears.

At a minimum, the clash should be documented in the Wagtail installation notes at [Integrating Wagtail into a Django project](http://docs.wagtail.io/en/latest/getting_started/integrating_into_django.html). (also in the Puput docs.).

I'm note sure which `Site` model came first. The Django version is documented as far back as Django 1.7. I am also not familiar enough with Wagtail to have a view as to whether the two `Site` models are conceptually similar. They certainly look similar. But the Django sites framework is an official part of Django and is used by many other apps.

I would say the options are:

[...]

PR #4014, opened 2017-11-13
Update SiteManager to cache sites identified by hostname/port [WIP]

ababic:
With all of the recent talk about potential changes to Site, and addressing some of the longstanding bugbears with the current Site implementation (see #3834, #2840), I decided to work on a cached SiteMiddleware solution that would yield the same results in all of the current test scenarios, while also attempting to improve the efficiency of how sites are identified for individual requests.

**Changes include:**

- Moving 'finding an appropriate Site for handling a request' logic to a 'get_for_request()' method on the SiteManager (Managers are good for this sort of thing)
- Adding a 'get_default()' method to SiteManager to help easily identify the default site
- Turning the 'find_for_request()' static method on the Site model to a class method that calls the SiteManager's 'get_for_request()' method
- Adding caching to 'get_default()' and 'get_for_request()' for faster results after the first call
- Using signal handlers to clear the cache when relevant actions occur
- Using smaller, simpler queries when attempting to find a Site for a request, instead of a single complicated query that automatically results in a slower initial response in all scenarios (in real life, the new solution will result in at most 2 small queries per request, or 3 if the default site hasn't yet been cached)
- Adjusted self.assertNumQueries() tests to account for the change in query numbers, and extended existing tests to show the cache (and cache clearing) in action

**Notes:** 

Unlike `wagtail.wagtailcore.sites.get_site_for_hostname()`, none of these new methods use 'select_related' to prefetch the `root_page` and include it with the result. This was a deliberate decision, made for the following reasons:

1) We don't know that the root page is needed every time 
2) It makes things far less efficient
3) With efficiency of fetching sites improved massively in most circumstances, fetching the page data afresh when needed each time shouldn't be a problem


ababic:
Any thoughts or comments would be gratefully received :)

ababic:
All of the current test failures are coming from the `wagtail.api` app. Have tried for hours to figure out why these failures are happening, but have so far come up short. If anyone with insight into the `wagtail.api` app could help shed some light on this, that would be great.

ababic:
Okay, turns out the test failures were due to a wee cache invalidation issue, which is solved by the new post_save and post_delete signal handlers for `Page`. That saying about cache invalidation and naming things rings true :)

wagtail/wagtailcore/models.py
```diff
@@ -40,8 +39,86 @@
...
+        return SITE_CACHE[site.pk]
+
+    def get_default(self):
+        """Returns the 'default' ``Site`` or raise an exception in no site is
+        set as the default. The result is cached."""
+        key = 'default'
+        if key in REQUEST_SITE_MATCH_CACHE:
+            return self._get_by_id(REQUEST_SITE_MATCH_CACHE['default'])
+
+        match = self.get(is_default_site=True)
```
BertrandBordage: What if multiple sites have `is_default_site=True`?
I suggest to change to `self.filter(is_default_site=True).first()`.

wagtail/wagtailcore/models.py
```diff
@@ -40,8 +39,86 @@
...
+        SITE_CACHE[site.pk] = site
+        REQUEST_SITE_MATCH_CACHE[match_cache_key] = site.pk
+        return SITE_CACHE[site.pk]
+
+    def get_default(self):
+        """Returns the 'default' ``Site`` or raise an exception in no site is
+        set as the default. The result is cached."""
+        key = 'default'
+        if key in REQUEST_SITE_MATCH_CACHE:
+            return self._get_by_id(REQUEST_SITE_MATCH_CACHE['default'])
```
BertrandBordage: Define a global variable for the default cache key name instead of defining a local variable and hard-code it twice.
Also, use something else that `'default'`, since it is possible to define a host named `'default'`. Something like `'__default__'` should be enough.

wagtail/wagtailcore/models.py
```diff
@@ -40,8 +39,86 @@
...
+
+    def _get_by_id(self, site_id):
+        if site_id not in SITE_CACHE:
+            site = self.get(pk=site_id)
+            SITE_CACHE[site_id] = site
+        return SITE_CACHE[site_id]
+
+    def _cache_and_return_match(self, site, match_cache_key):
+        SITE_CACHE[site.pk] = site
+        REQUEST_SITE_MATCH_CACHE[match_cache_key] = site.pk
```
BertrandBordage: I don’t understand the point of this second cache. From what you wrote in your last commit, it’s to save memory by only caching IDs, but… Sites are still stored in `SITE_CACHE`, so I don’t see how this saves memory.

Unless I totally missed your point, you can remove this cache, it should use more memory than earlier. Ditching this second cache also makes the code simpler, as we can directly fetch site objects from host+port strings.

wagtail/wagtailcore/models.py
```diff
@@ -40,8 +39,86 @@
...
+                # If there's only one site matching the hostname, use that,
+                # since there's no ambiguity
+                match = self.get(hostname=hostname)
+                return self._cache_and_return_match(match, key)
+            except(Site.DoesNotExist, Site.MultipleObjectsReturned):
+                pass
+
+        # Fall back to using default site
+        match = self.get_default()
+        return self._cache_and_return_match(match, key)
```
BertrandBordage: In all this method, `match` are `Site` instances. This is a bit misleading when we read what was in `get_site_for_hostname`, which was a very different type of match. Prefer `site`.

wagtail/wagtailcore/models.py
```diff
@@ -40,8 +39,86 @@
...
+                except Site.DoesNotExist:
+                    pass
+
+            try:
+                # If there's only one site matching the hostname, use that,
+                # since there's no ambiguity
+                match = self.get(hostname=hostname)
+                return self._cache_and_return_match(match, key)
+            except(Site.DoesNotExist, Site.MultipleObjectsReturned):
+                pass
```
BertrandBordage: Isn’t it better to return the first site matching `hostname` when there are multiple sites for that hostname, instead of returning the default site? I would put `self.filter(hostname=hostname).first()` and remove `MultipleObjectsReturned`.

wagtail/wagtailcore/models.py
```diff
@@ -40,8 +39,86 @@
...
+
+    def get_default(self):
+        """Returns the 'default' ``Site`` or raise an exception in no site is
+        set as the default. The result is cached."""
+        key = 'default'
+        if key in REQUEST_SITE_MATCH_CACHE:
+            return self._get_by_id(REQUEST_SITE_MATCH_CACHE['default'])
+
+        match = self.get(is_default_site=True)
+        return self._cache_and_return_match(match, key)
```
BertrandBordage: When caching the default site, you could save it both in the `'__default__'` key and the one of its host+port. This will avoid extra SQL queries when fetching that site again for the first time.

wagtail/wagtailcore/models.py
```diff
@@ -60,47 +61,63 @@ def get_hostname_and_port_from_request(request):
             port = None
         return hostname, port
 
+    @staticmethod
+    def clear_site_cache():
+        global SITE_CACHE
+        global REQUEST_SITE_MATCH_CACHE
+        SITE_CACHE = {}
+        REQUEST_SITE_MATCH_CACHE = {}
```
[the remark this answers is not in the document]
ababic: Sure, i'll get this sorted soon :)

wagtail/wagtailcore/models.py
```diff
@@ -60,47 +61,63 @@ def get_hostname_and_port_from_request(request):
...
+
+    def _get_by_id(self, site_id):
+        if site_id not in SITE_CACHE:
+            site = self.get(pk=site_id)
+            SITE_CACHE[site_id] = site
```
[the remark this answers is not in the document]
ababic: I thought this might produce a slightly less scary stack-trace if the query failed for whatever reason. But also happy to make this change.

wagtail/wagtailcore/models.py
```diff
@@ -60,47 +61,63 @@ def get_hostname_and_port_from_request(request):
...
+        return SITE_CACHE[site_id]
+
+    def _cache_and_return_match(self, site, match_cache_key):
+        SITE_CACHE[site.pk] = site
+        REQUEST_SITE_MATCH_CACHE[match_cache_key] = site.pk
+        return SITE_CACHE[site.pk]
```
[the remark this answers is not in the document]
ababic: I guess it has the same address in memory, so I see no reason not to. I was really just taking my queues from Django's implementation. Again, happy to change

wagtail/wagtailcore/models.py
ababic: I was (perhaps naively) attempting to guard against the possibility of the original `SITE_CACHE` getting filled with duplicates in the event of some attack where ports could be spoofed for a matching domain, resulting in a Site object existing in memory for each port/hostname combination. In a genuine attack situation, I suppose it wouldn't make much difference in the grand scheme of things.

wagtail/wagtailcore/models.py
ababic: Sure, yes. I'll definitely do this.

wagtail/wagtailcore/models.py
ababic: This one I'm a little unsure about. I know site data can be altered in ways that bypasses this, but it seems that the model assumes there should only be one default site:
https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailcore/models.py#L145

I fully understand that doing as you propose would cater for every scenario, but does that mean that the code contradicts itself slightly?

wagtail/wagtailcore/models.py
ababic: Great suggestion.

wagtail/wagtailcore/models.py
ababic: There are a couple of tests that are directly affected by this:

On the plus side: `wagtail.wagtailcore.tests.test_page_model.TestSiteRouting.test_unrecognised_port_and_default_host_routes_to_default_site` is made slightly more efficient (2 queries instead of 3)

But on the negative:
`wagtail.wagtailcore.tests.test_page_model.TestSiteRouting.test_unrecognised_port_on_known_hostname_routes_to_default_site_if_ambiguity` fails.

The existing pattern was the only one I tried that would pass the tests.

wagtail/wagtailcore/models.py
ababic: Yes, I can see your point there. Using `site` consistently would be less ambiguous.

ababic:
Thank you for the review @BertrandBordage. Much appreciated!

In relation to your suggestion here:

wagtail/wagtailcore/models.py
ababic: @BertrandBordage See also: http://docs.wagtail.io/en/v1.13.1/reference/pages/model_reference.html#wagtail.wagtailcore.models.Site.is_default_site

ababic:
Thank you for the review @BertrandBordage! much appreciated. In relation to your idea proposed above:

> Also, I have something to propose, that will probably simplify things a lot. Instead of fetching uncached sites one by one by doing between 1 and 3 SQL queries, we could write something really easy and more efficient in this case. Every time we can’t find a site in the cache, we fetch all uncached sites and cache them.

From what I understand, hostname-only or default site matches can be returned for an undeterminable number of 'ports', so I'm not sure we can populate everything up-front. Unless there's some configuration for that somewhere that I've missed? If not, and we can only partially populate the cache, is that still useful enough to justify the additional code complication?

wagtail/wagtailcore/models.py
ababic: Oh, wait. This is already what happens in the last line of `get_for_request()`

wagtail/wagtailcore/models.py
BertrandBordage: The number of ports is finite, it’s 65535. So an hypothetical attack based on that would at max populate the dictionary with 65535 site objects × the number of hosts. This may take at max a dozen megabytes, so it would be a pretty lame attack.

That being said, this will never happen unless you did all these:
- configured your OS to accept connections on all ports
- configured your HTTP server to redirect by default all hosts and ports to a Wagtail instance
- configured Django with something like `ALLOWED_HOSTS = ['*']`

So this will not happen in any correctly configured server, and would actually need some extra work to allow any port through the HTTP server: by default they only listen to ports 80 & 443, as far as I know.

It’s possible though to have an infinite number of hosts if you setup wildcard subdomains and adjust `ALLOWED_HOSTS` accordingly. But in that event, you also need to have an huge number of `Site` instances in the database, so we can’t avoid caching each one of them. If a subdomain does not exist in the database, then we simply fetch the default one using the `'__default__'` key, so it doesn’t change memory usage.

BertrandBordage:
As mentioned in my previous comment, the attack on ports is almost impossible on a normal server, and it is very limited in case it happens anyway.

Additionally, for a request on an unknown port, we simply return the default site, so we don’t even have to do the 2 queries done with your solution.

Something I’m wondering now about all this caching system: what happens when running multiple workers? If I modify a site on a given process, the cache is correctly cleared in that process, but it’s not clearing the cache from other processes…

wagtail/wagtailcore/models.py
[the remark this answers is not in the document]
BertrandBordage: It points to the same address, indeed, but fetching an attribute and a key from a dictionary have a speed cost. That takes around 150 ns, which seems almost nothing, but caches usually return results in a few μs, so additional costs like this end up being important ;)

(And yeah, sorry to break the myth a bit, but Django has very mediocre optimization ^^
Always wanted to work on it, but I’m still scared by the amount of “paperwork” involved.)

wagtail/wagtailcore/models.py
BertrandBordage: Yeah, a bit tricky to decide. It’s either being tolerant and strict.
We can keep it strict as you wrote, but maybe it would be better to explain what’s happening, like catching the error and reraising it with a more explicit message:
```
MultipleObjectsReturned: Multiple default sites found, only one allowed.
```

wagtail/wagtailcore/models.py
BertrandBordage: OK, if I’m not mistaken, the “fetch all once” solution I suggested would handle these correctly.

wagtail/wagtailcore/models.py
ababic: Yes, thanks for the insight. Like I said, it was a little naive. I'll re-simplify.

ababic:
> what happens when running multiple workers? If I modify a site on a given process, the cache is correctly cleared in that process, but it’s not clearing the cache from other processes…

I can't claim to be knowledgeable about this particular subject, but wouldn't those workers also be responding to the same signals triggered by such changes, and also be clearing their version of the cache in memory?

Have we come to a bit of an en-pass here? Is a change in tact needed? Would a switch to using the low-level caching API address these concerns?

BertrandBordage:
Unfortunately, signals are simply a list storing Python functions then calling them later, so it’s within the same process. The only way to communicate to the other process is either through the database or through a cache backend. And it can’t work with the default cache backend, LocMem, since it’s simply a Python dictionary.

What’s weird is that Django is using exactly the same approach as you took, don’t spend more time on this, I’ll make some research and ask the Django team about it. Either I’m missing something, or Django really has an issue with this. Since I remember having to restart servers after changing Site objects in both Django & Wagtail, I guess there’s indeed a problem.

Anyway, the caching could be as simple as using a configured cache instead of a dictionary, but only if the cache is not LocMem.

Or we can ditch the idea of caching site objects, [django-cachalot](https://github.com/BertrandBordage/django-cachalot) does exactly that kind of caching automatically ;)
(Just kidding, I know it doesn’t fit all use cases.)

BertrandBordage:
Just created ticket [#28844](https://code.djangoproject.com/ticket/28844) in Django about the multi-process issue, with a minimal example.

ababic:
Looks like you were right @BertrandBordage. Apparently it's been a known issue for years, but there's been no consensus on a solution yet :(

Seems like one way forward might be a switch to using the cache backend under certain circumstances. But then, we'd have to rethink entirely the way cache invalidation works, introducing a tonne more code that will require tests and maintenance... 

And by the time we're done, we may have well have used `django-cachalot`, which feels like the 'already-evolved' version of what we're trying to achieve (with the advantage of it being generic, and already well tested). But then again, that would mean forcing additional dependencies and configuration onto the developers, which I would expect that nobody in the core team (including yourself) is going to want to do for the sake of speeding up this one small part of wagtail. 

For now, I think the sensible decision is to abandon this PR - unless you feel differently?

ababic:
Hi @BertrandBordage, I got over my initial feelings of deflation and have decided to plough on with this in the direction you proposed. It has meant a LOT of changes though. 

In particular, I think we might want to revise how the new settings are used, or the warning is surfaced (maybe warning isn't enough, and you'd rather raise an IncorrectlyConfigured). But anyway.

I have some lingering concerns about ensuring all existing data is cleared from the cache when clear_site_cache() is called. These will no doubt benefit from some unit tests to ensure that's working well enough.

ababic:
@BertrandBordage The original intention has changed so much since starting this, that it's probably best to close this pull and create a fresh one based off the master branch. I'll resubmit again soon with a more appropriate writeup.

BertrandBordage:
Thank you @ababic, I’ll then make a new review. From what I quickly saw of your recent changes, this seems way more robust.

PR #4018, opened 2017-11-14
Fix field-limited search on specific() queryset on postgres, for 1.12

gasman:
Fixes #4015 - fix backported from 781263d.

PR #4019, opened 2017-11-14
Make ordering Page search by title work on 1.12.x

gasman:
Multiple issues prevent `Page.objects.order_by('title').search('blah', order_by_relevance=False)` from working on Wagtail 1.12.x:

* #3725 (broken `order_by_relevance` on Postgres, fixed in #3726 and backported here)
* 'title' is not defined as a `FilterField` - fixed as part of #3957 and cherry-picked here

Since the original test from #3726 seems to have been the one that set off a snowball effect of broken search tests that (I think) was only properly resolved at the Bristol sprint, I've skipped those commits and added a new test as part of the test_page_search module introduced in #4018.

### DIFF
--- a/wagtail/contrib/postgres_search/backend.py
+++ b/wagtail/contrib/postgres_search/backend.py
@@ -180,7 +180,7 @@ class PostgresSearchQuery(BaseSearchQuery):
 
     def __init__(self, *args, **kwargs):
         super(PostgresSearchQuery, self).__init__(*args, **kwargs)
-        self.search_fields = self.queryset.model.get_search_fields()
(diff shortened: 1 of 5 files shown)


gasman:
65205a2586241584a90440dacff7f28ba590bf0d is yet another bug which surfaces as a result of defining Page's `title` as both a `SearchField` and a `FilterField` - this will need to be forward-ported (along with these tests) to 1.13.x and master.

gasman:
As discussed with @kaedroho, we won't backport this to 1.12 since this is more of an unsupported feature than a bug, and making it work 100% correctly on Postgres requires fixes that only exist in the post-Bristol-sprint refactored backend.

Issue #2834, opened 2016-07-08
Redirect to Form landing page


thenewguy:
Per the [comment in source](https://github.com/torchbox/wagtail/blob/stable/1.3.x/wagtail/wagtailforms/models.py#L176), it would be better to redirect to the form success page instead of rendering it on post.

A simple way to achieve that without needing to worry about adjusting url routes and such would be to use a query param... "/foo-form/?success".

If https://github.com/torchbox/wagtail/issues/2833 is accepted, the FormSubmission instance id or randomized token could be passed as well so it can be fetched and passed to the template.

If security is a concern around putting the id/token in url and leaking data to malicious users, it shouldn't be very difficult to use session data for verifying the user created the FormSubmission data or raising http404.

-- edit --

Forgot to mention another practical use for the query param (or unique url but more complicated with wagtail page routing) is it makes conversion tracking simple because you can simply check destination instead of needed developer to trigger an analytics event specific to the submission.

karabellyj:
Hi I am new to contibuting and programming. I've read that the best way to learn is to contribute so I did redirection as suggested above using query param but a few tests started to fail. Most of them fail because of AssertionError: Couldn't retrieve content: Response code was 302 (expected 200). I see that it is due to the redirection but how to get rid of it. Should I adjust tests or my new code?

gasman:
Hi @karabellyj - thanks for taking this on!

Adjusting the tests to reflect the new behaviour is the right thing to do here.

karabellyj:
Solved that error by adding follow=True to the tests but now CustomSubmission tests are raising AssertionError: False is not true : Couldn't find 'Thank you for your patience!' in response and I don't see why.

strindhaug:
I've created two pull requests to address this issue:
 https://github.com/wagtail/wagtail/pull/4024  is a m one minor change that I think should be merged anyway. It simply moves the render-part of the serve method for the landing page to a separate method so that you can override the rendering of the landing page without having to copy paste all the other stuff from the serve method.

https://github.com/wagtail/wagtail/pull/4025 is up for discussion as this adds a new optional field to the model. (And I now realize I didn't make any migrations for this, yet). This new field is a foreign key to a page; that if set, is is used to redirect to instead of rendering the landing page.

strindhaug:
I also made another PR that implements the success parameter idea: https://github.com/wagtail/wagtail/pull/4026

[...]

PR #4024, opened 2017-11-16
Forms: More easily overridable landing page rendering

strindhaug:
Move the landing page rendering to an easily overridable method.
Partial fix/improvement for https://github.com/wagtail/wagtail/issues/2834


wagtail/wagtailforms/models.py
```diff
@@ -260,20 +260,28 @@ def process_form_submission(self, form):
...
+            self.get_landing_page_template(request),
+            self.get_context(request)
+        )
+
     def serve(self, request, *args, **kwargs):
         if request.method == 'POST':
             form = self.get_form(request.POST, request.FILES, page=self, user=request.user)
 
             if form.is_valid():
                 self.process_form_submission(form)
```
lb-: Duplicate of comment on #2834
----

We need a way to ensure that the form_submission instance can be passed to the context of the landing page.

As per this outstanding item in #2833
> The FormSubmission instance should then be passed to the render context for the "landing page".

 the `render_landing_page` method should also accept a parameter `form_submission`.

Which should be generated by updating the line:
`self.process_form_submission(form)` 
to
`form_submission = self.process_form_submission(form)` 

Then 
`return self.render_landing_page(request, *args, **kwargs)`
to
`return self.render_landing_page(request, form_submission=form_submission, *args, **kwargs)`

Finally, the context should be updated in the render_landing_page return render eg.

```
 # TODO: It is much better to redirect to it
context = self.get_context(request)
context['form_submission'] = form_submission
    return render(
        request,
        self.get_landing_page_template(request),
         context
        )
```

strindhaug:
I've added the form_submission to the context as suggested.

strindhaug:
I still like the idea of having a parameter that displays the landing page without having to fill out the form would be nice for template development purposes. But this should perhaps be in addition to the real submit flow and not a part of it. 

Or maybe we should just add an example of how to add this debug parameter in the docs. Something like this example:
```python
class MyFormPage(AbstractEmailForm):

    # ... 

    def serve(self, request, *args, **kwargs):
        """
        This adds a debug parameter to test the landing page during development.

        To test landing page add the get parameter to the url like this:
        /my-form/?debug_landing=True

        """
        if request.method == 'POST' and request.GET.get('debug_landing'):
            form_submission = None
            # Or optionally create a synthetic FormSubmission here for testing
            return self.render_landing_page(request, form_submission, *args, **kwargs)

        # else do normal rendering
        return super(MyFormPage, self).serve(request, *args, **kwargs)
```

lb-:
@strindhaug when you are editing a form page, there is a way to preview both the form and the landing page easily. Does that achieve the goal you wanted?

Screenshot attached.

wagtail - editing form page_ contact us


This made me realise that `render_landing_page` will need to handle cases where there is no form_submission can you make another change to the PR so that form_submission defaults to None.

Also, you will need to update the preview code here to ensure that previews of landing page use the new `render_landing_page` method.

https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailforms/models.py#L295

strindhaug:
Oh, I didn't realize there was a preview mode for it, that's nice. =) Well then the parameter idea is pointless.

I've updated the serve_preview and made the form_submission optional with default value `None`.

lb-:
@strindhaug are you able to put in some additional tests?

possibly a model that overrides this method (eg. add something to the context or render with a different template) - to ensure that it gets used in both normal and preview mode

and also that the default landing page contains the form_submission in the context

let me know if you need a hand

strindhaug:
I started creating a form that redirects instead of landing page but it fails before it starts because the table is missing. 
How do I actually generate migrations for the test app? I'm not supposed to write them all manually I assume...

Also, is there any deterministic data in the default form_submission that can be outputted easily in the template that doesn't require modifying the model? I tried outputting `{{ form_submission.get_data.your-email }}` but that fails because hyphens aren't allowed. I tried adding an additional field without space to get a simple value to output but then several other tests fail because the default form now has an extra field.

gasman:
Hi @strindhaug - you can create migrations for the test app by running django-admin.py makemigrations --settings=wagtail.tests.settings from the Wagtail root.

lb-:
@strindhaug as `get_data()` returns a dict, there are multiple ways to get the data out of that within Django templates.

This [Stack Overflow question](https://stackoverflow.com/questions/8000022/django-template-how-to-look-up-a-dictionary-value-with-a-variable) has a bunch of ways, depending on what is needed. There is no default Django tag to get values from keys in a dict though.

eg. something like:

```
<ul>
  {% for key, value in get_data.items %}
    <li>{{ key }}: {{ value }}</li>
  {% endfor %}
</ul>
```

strindhaug:
Used the for loop you suggested; though it's sort of meaningless thing to output, at least it's testing that the form_submission is added.

Added a similar test to the `TestFormWithCustomSubmission` by outputting the custom field username.
This test initially failed because the overridden `process_form_submission` didn't return the created form_submission. So I updated this method to do so.

wagtail/wagtailforms/models.py
```diff
@@ -260,20 +260,29 @@ def process_form_submission(self, form):
...
-
-                # render the landing_page
-                # TODO: It is much better to redirect to it
-                return render(
-                    request,
-                    self.get_landing_page_template(request),
-                    self.get_context(request)
-                )
+                form_submission = self.process_form_submission(form)
+                # Forwarding args and kwargs just in case for flexibility
```
lb-: I think this comment can be removed.

lb-:
@strindhaug added one minor comment - just to remove a comment you added about the passing of args/kwargs, I don't think that is needed.

I will bring this up at our dev meeting today.

@tomdyson FYI

lb-:
@strindhaug one last thing is adding some docs about this new awesome feature.

Best to put them at the end of this [docs page](http://docs.wagtail.io/en/v1.13.1/reference/contrib/forms/customisation.html) which is built from [docs/reference/contrib/forms/customisation.rst](https://github.com/wagtail/wagtail/blob/master/docs/reference/contrib/forms/customisation.rst).

Items that need updating (from a quick look)

[ ] Update section **Check that a submission already exists for a user** to reflect changes to serve method + example should not need `from django.shortcuts import render`
[ ] Update section **Multi-step form* to reflect changes to serve method + example should not need `from django.shortcuts import render`
[ ] Add example about overriding the `render_landing_page` (the one from the tests where you do a redirect is great)
[ ] Update the paragraph that talks about `form_page_landing.html is a regular Wagtail template,` on the [forms/index.html docs](http://docs.wagtail.io/en/v1.13.1/reference/contrib/forms/index.html) page to add note that the form_submission instance is available in the template context as ...

Let me know if you need a hand.

lb-:
@gasman I have committed these changes to master, struggled a bit with the rebase though. The commits are not linked to this PR correctly but they are there, should I close this PR now?

gasman:
@lb- Yes please. Having to close PRs manually is normal behaviour when you rebase them on the command line.

lb-:
Merged

PR #4025, opened 2017-11-16
Forms: Optional form redirect on submit

strindhaug:
Add an optional foreign key to a Page to the AbstractForm called receipt_page_redirect that if set will cause the form to redirect to this page instead of rendering the landing page on successful submit.

Expands on https://github.com/wagtail/wagtail/pull/4024 as a possible fix for https://github.com/wagtail/wagtail/issues/2834

_This is a "work in progress" and "request for comments" type of pull request, but I can't add labels myself._

### DIFF
--- a/wagtail/wagtailforms/models.py
+++ b/wagtail/wagtailforms/models.py
@@ -6,7 +6,7 @@ import os
 from django.contrib.contenttypes.models import ContentType
 from django.core.serializers.json import DjangoJSONEncoder
 from django.db import models
-from django.shortcuts import render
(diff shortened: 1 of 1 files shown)


strindhaug:
I haven't added any migrations yet.

lb-:
should resolve #4024 before we look at this one

lb-:
@strindhaug any issues with this being closed now? I think #4024 gives developers enough freedom to handle redirects as they see fit, seeing as there are several approaches that can be taken.

PR #4026, opened 2017-11-16
Forms: redirect with GET param on submit

strindhaug:
Redirect to self with a GET param: `?success=True`.
By redirecting to a GET request with special parameter, on form submit this prevents warnings about reposting in the browser; and gives the receipt page a GET-able url which makes testing the design of the receipt page a lot less painful. This also allows conversion tracking based on the url.

This is an alternative fix for https://github.com/wagtail/wagtail/issues/2834, see also https://github.com/wagtail/wagtail/pull/4025 for an alternative/complementary fix.

### Todos:
- Maybe the success param name should be configurable or localized?
- Maye the test for the success parameter should be moved to a method? E.g. `is_landing_page` or something...
    
Note: I had to modify the model `wagtail.tests.testapp.models.FormPageWithCustomSubmission` to make the tests work. (But I figure this is fair enough, because when you override the workings of a Class to that extent you can't expect it not to break on updates sometimes.)

### DIFF
--- a/wagtail/tests/testapp/models.py
+++ b/wagtail/tests/testapp/models.py
@@ -482,7 +482,7 @@ class FormPageWithCustomSubmission(AbstractEmailForm):
             send_mail(self.subject, content, addresses, self.from_address,)
 
     def serve(self, request, *args, **kwargs):
-        if self.get_submission_class().objects.filter(page=self, user__pk=request.user.pk).exists():
+        if not request.GET.get('success') and self.get_submission_class().objects.filter(page=self, user__pk=request.user.pk).exists():
             return render(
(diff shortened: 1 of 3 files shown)


strindhaug:
I'm not sure this redirect solution is a good idea, in addition to possibly breaking a lot of subclasses, this makes including the form submission ( https://github.com/wagtail/wagtail/issues/2833 ) a lot more complicated.

lb-:
@strindhaug I agree - I think let's remove that comment like you put in the other PR and maybe leave the redirection to individual overriding if people need that.

Sending context to a redirect is not simple.

PR #4027, opened 2017-11-16
Search fixes (backport for 1.12.x)

kaedroho:
Replacement for #3957 taking only the most important changes


gasman:
Note: there's no test for ffb7b956f12c4277dfecb8644413b237eb3aa0bf here, because it can only be reproduced intermittently - we've only seen it surface in the context of the test_page_search module added in #4019, which @kaedroho and I have decided we're not going to backport to 1.12/1.13 (because it requires `order_by_relevance=False` support in the Postgres backend, which is only fully fixed on master).

PR #4029, opened 2017-11-17
Fixes breadcrumbs on ModelAdmin inspect and choose parent view

### LINKED ISSUE #3889: Breadcrumbs on ModelAdmin inspect view is visually broken

### Issue Summary

Breadcrumbs is partially hidden by the sidebar.
modeladmin

On small screens, breadcrumbs are on white background.
modeladmin2

### Steps to Reproduce

1. Create a model
2. Create the corresponding ModelAdmin register it.
3. Set `inspect_view_enabled` to `True` on that ModelAdmin
4. Navigate to the ModelAdmin index page.
5. Click `Inspect` on one of the instances.

lb-:
Fixes #3889 

It would be good for someone to check that this is not too hacky/heavy handed for the solution.

I attempted to fix this with CSS changes alone but the breadcrumbs do need to be inside the header div for things to work correctly. I also had to write a custom small amount of scss as pulling in other css files that wagtailadmin uses often covered way more changes and visually broke other things.

Finally, I thought it would be best to keep using the `wagtailadmin/shared/header.html` with this `inject` tag as it means future updates to header.html will be incorporated.

Also, added tests to check for this issue, linting checked, tests pass (locally), no docs needed for this as it is essentially a fix. I also checked visually on Mac; Chrome, Firefox, Safari for the changes (including responsive sizes).


ababic:
Hi @lb-. I could be oversimplifying here - But, I wonder if it might be possible (and less hacky) to take a different approach to the template-related changes?

1. Add an empty `breadcrumbs` block to `wagtailadmin/templates/wagtailadmin/shared/header.html` in the relevant place
2. Create a new template `contrib/modeladmin/templates/modeladmin/includes/header_with_breadcrumbs.html` that extends the one above, overriding the `breadcrumbs` block to output the breadcrumbs
3. Have the inspect / choose parent template use the new header template instead

Does that sound okay?

ababic:
Or, if we're unsure about adding a {% block breadcrumbs %} to the shared header template, for fear it's misleading about how breadcrumbs are added, we could maybe just split some of the contents up into some logical blocks of their own, and the new header template could override one of those blocks with {{ block.super }} to add the breadcrumbs in the relevant place.

lb-:
I like your first suggestion - I don't think it will be any more confusing than the current way header is used or not used throughout.

lb-:
@gasman I am running into a CI issue I do not understand:

```
(wagtailadmin.W001) CSS for the Wagtail admin is missing
	HINT: 
            Most likely you are running a development (non-packaged) copy of
            Wagtail and have not built the static assets -
            see http://docs.wagtail.io/en/latest/contributing/developing.html

            File not found: /home/ubuntu/wagtail/wagtail/wagtailadmin/static/wagtailadmin/css/normalize.css
```

I have run `npm run build` and `npm run lint` again with no obvious issues/changes.

gasman:
@lb- The warning about CSS is normal (on Travis at least), and is unrelated to the test failures - we don't install npm on Travis, so the front-end assets don't get built there.

wagtail/contrib/modeladmin/templatetags/modeladmin_tags.py
```diff
@@ -182,8 +182,3 @@ def result_row_value_display(context, index):
         'closing_tag': closing_tag,
     })
     return context
-
-
-@register.filter
-def get_content_type_for_obj(obj):
-    return obj.__class__._meta.verbose_name
```
ababic: I think this filter may have been removed accidentally?

lb-: Thanks, just put through some changes

gasman:
Rebased / squashed / merged in bcf6b6da772e5f484b1b9254cde5acd6cef30bda 👍

lb-:
thanks @gasman

Issue #3944, opened 2017-10-19
Userbar z-index is 1


mgjbroadbent:
### Issue Summary

The z-index of the wagtail-userbar element is now set to 1 in the pypi version of wagtail v1.13 instead of 9999 which was used in v1.12.2.  This causes the user bar to appear under page elements that use a z-index of greater than 1.

### Steps to Reproduce

* Download wagtail v1.13 using pip.
* Open the file <python package>/wagtail/wagtailadmin/static/wagtailadmin/css/userbar.css
* Locate the .wagtail-userbar CSS element.
* Observe the z-index in the .wagtail-userbar CSS element:

`... .wagtail-userbar{position:fixed;z-index:1; ...`

In v.1.12.2 version this is:

```
...
.wagtail-userbar {
  position: fixed;
  z-index: 9999;
...
```

### Technical details

The source scss file shows the z-index is set to 9999 as expected:
* https://github.com/wagtail/wagtail/blob/05adf2c85deb534e213591158cecbd56571c8e84/wagtail/wagtailadmin/static_src/wagtailadmin/scss/userbar.scss#L90

strindhaug:
I guess this might be caused by cssnano "optimizing" z-index by default in some versions:
https://github.com/ben-eb/gulp-cssnano/issues/85

Add `zindex: false` to the options of cssnano to stop it from messing with the z-index.

gasman:
@strindhaug Good catch - thanks for the info!

strindhaug:
Yeah, I wasted a few hours some days ago trying to figure why on earth the z-index was 1 in production when it was 999 locally (running unminified); so the  "z-index is 1" title jumped out to me when i glanced at the issue list.

gasman:
Digging into the relevant code, I see that the use of cssnano (and a couple of other bits and pieces related to performance testing) is conditional on `NODE_ENV=production`, which will usually only happen when we're preparing the final release (or release candidate). Clearly, this vastly limits our ability to catch bugs like this during development...

ping @thibaudcolas @jjanssen @vincentaudebert - what can we do to ensure that the frontend assets on our development builds remain as close as possible to the final release version? Would it be reasonable to update our config to always apply 'production' settings on `npm run build` - or would this make development more difficult (e.g. loss of sourcemaps)?

vincentaudebert:
[...]

PR #4039, opened 2017-11-17
Process CSS with cssnano in development build

thibaudcolas:
Follow-up to https://github.com/wagtail/wagtail/issues/3944#issuecomment-342874112, #3944. cssnano will now process the CSS in dev mode too, hopefully reducing the risk of it introducing issues that then go unnoticed.

Source maps still work as expected so this doesn't change the debug-ability of stylesheets. The compilation task is a bit slower (2-3x slower, ±2-3s on my 2015 MBP) but that sounds acceptable.


gasman:
Thanks @thibaudcolas!

PR #4044, opened 2017-11-19
Fix forms docs link in readme

### LINKED ISSUE #4043: Link "an extensible form builder" on pypi page broken
On the pypi page, the extensible "form builder" link leads to a 404:

http://docs.wagtail.io/en/latest/reference/contrib/forms.html

lb-:
Fixes #4043


gasman:
Thanks @lb-!

PR #4045, opened 2017-11-19
Add instructions on how to do test app migrations

lb-:
Previously answered [here](https://github.com/wagtail/wagtail/pull/4024#issuecomment-345267222) and [here](https://groups.google.com/forum/#!msg/wagtail-developers/P6x9N5qzq-I/0gbRzWgBBgAJ). Just adding this into these docs for future contributors.


gasman:
Good idea - thanks @lb-!

PR #4047, opened 2017-11-19
Add classname to sequence blocks (ListBlock & StreamBlock)

### LINKED ISSUE #4042: Support form_classname attribute on StreamBlock and ListBlock
Version: 1.13

`StructBlock` supports the `form_classname` attribute in its Meta class. It would be nice if `StreamBlock` and `ListBlock` also support it.

lb-:
This is just a proof of concept to potentially resolve #4042 - this adds the `classname` classes to the templates for `ListBlock` and `StreamBlock`.

I would love some feedback on this.

* Tests added for ListBlock only, can add others if this is the way we want to resolve this.
* Not sure if we need to revise docs as `classname` is used in many examples in streamfield docs
* Linting has been run

### DIFF
--- a/wagtail/wagtailcore/tests/test_blocks.py
+++ b/wagtail/wagtailcore/tests/test_blocks.py
@@ -1670,6 +1670,32 @@ class TestListBlock(WagtailTestUtils, SimpleTestCase):
         )
         self.assertIn('value="chocolate"', form_html)
 
+    def render_form_with_classname(self, classname):
+        class LinkBlock(blocks.StructBlock):
+            title = blocks.CharBlock()
+            link = blocks.URLBlock()
+
+        block = blocks.ListBlock(LinkBlock, classname=classname)
+
+        html = block.render_form([
+            {
+                'title': "Wagtail",
+                'link': 'http://www.wagtail.io',
+            },
+            {
+                'title': "Django",
+                'link': 'http://www.djangoproject.com',
+            },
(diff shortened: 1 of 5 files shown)


wagtail/wagtailcore/tests/test_blocks.py
```diff
@@ -1670,6 +1670,32 @@ class ShoppingListBlock(blocks.StructBlock):
         )
         self.assertIn('value="chocolate"', form_html)
 
+    def render_form_with_classname(self, classname):
+        class LinkBlock(blocks.StructBlock):
+            title = blocks.CharBlock()
+            link = blocks.URLBlock()
+
+        block = blocks.ListBlock(LinkBlock, classname=classname)
```
mnieber: Shouldn't the argument be called `form_classname` just like the corresponding attribute in `StructBlock`'s Meta class? 

Actually, it seems that the naming is not entirely consistent, as FieldPanels have a `classname` argument that affects their rendering in the admin, whereas for StructBlock the argument is called `form_classname`.

lb-: @mnieber  - thanks for the comment, gonna attempt to pick this one back up.

Yes, it appears that Blocks use meta.classname but StructBlocks use meta.form_classname.

I will update this with `form_classname` and re-submit (applied to the latest code).

lb-:
Closing, replaced by #5100

Issue #2841, opened 2016-07-13
The 'scheduled publishing' feature needs documenting


ababic:
After a couple of years of working with Wagtail, i'm still not entirely sure what page revisions are capable of, and what they aren't, and also where the `publish_scheduled_pages` management command comes into it. It would be great to have some docs to explain things better. Some major points of confusion for me have been:

1. Can you schedule publishing of a revision for a page that is already live, and keep the current revision active until that time comes? It would seem you can't, but as a developer, I would've thought that was one of the primary benefits of keeping multiple revisions., and should be possible. I'm okay with it not working that way, but it doesn't seem to be explained anywhere, and I'm struggling to guide clients in using the CMS because of that.
2. What happens to a scheduled revision when another revision is published to the page before the scheduled revisions 'go_live_at' date? I **think** the scheduled revision is considered redundant, and the scheduled publication is forgotten about (as I'd expect). But again, the limitations just don't seem to be explained anywhere.
3. What code is responsible for publishing a scheduled revision when the time comes? I'm guessing the `publish_schedued_pages` command is responsible for this, but can't find any reference to this anywhere [other than in the reference section](http://docs.wagtail.io/en/v1.5.2/reference/management_commands.html#publish-scheduled-pages) of the docs. If the scheduled publishing fields are to be present by default, but require a task to be set up in order for them to work properly, I'd expect that to be outlined very clearly in the docs. In fact, I'd probably expect that feature to be disabled by default, and that I'd need to use a setting to turn it on, after setting up the task.

spapas:
Hello @ababic, this is not proper documentation but taking a look at PR #205 should answer most of your questions about scheduled publishing!

ababic:
Thanks @spapas, that does help a little with the things I was unsure about. We could still certainly do with some documentation, though. I would suggest as a minimum that a 'setting up a recurring task' part needs adding to the setup instructions (I would write it myself, but I don't see myself as technical enough to provide instructions on that - I have very little experience with celery/cron etc). 

Secondary to that, would bean additional part in the `Editor's guide`, that would explain what editors can and can't do in simple terms (e.g. "If you schedule publishing for a live page, this will happen.")

And in addition, an a part in the `Advanced topics`, with more a more technical explanation of how it works (to better arm developers in explaining the feature to editors - if they neglect to read the part above, or want you to explain why certain limitations exist)

lacymorrow:
[...]

PR #4050, opened 2017-11-21
RFC:21 Revision scheduled publishing

hakjoon:
This PR adds functionality for [Wagtail RFC 21](https://github.com/wagtail/rfcs/pull/21) allowing for scheduling revisions without unpublishing currently live pages.

### DIFF
--- a/docs/editor_manual/new_pages/the_promote_tab.rst
+++ b/docs/editor_manual/new_pages/the_promote_tab.rst
@@ -2,7 +2,7 @@
  Edit Page tabs
 ================
 
-A common feature of the *Edit* pages for all page types is the two tabs at the top of the screen. The first, Content, is where you build the content of the page itself.
+A common feature of the *Edit* pages for all page types is the three tabs at the top of the screen. The first, *Content*, is where you build the content of the page itself.
 
 The Promote tab
 ~~~~~~~~~~~~~~~
@@ -22,8 +22,7 @@ The second, *Promote*, is where you can set all the 'metadata' (data about data!
 The Settings Tab
 ~~~~~~~~~~~~~~~~
 
-The *Settings* has two fields by default.
+The *Settings* tab has two fields by default.
 
-* **Go Live date/time:** Sets the time in which the changes should go live when published.  If you publish a page that has this field set to a time in the future it will be scheduled until the time comes.
-  See :ref:`scheduled_publishing` for more details.
-* **Expiry date/time:** Sets the time in which this page should be unpublished.
+* **Go Live date/time:** Sets the time at which the changes should go live when published. See :ref:`scheduled_publishing` for more details.
+* **Expiry date/time:** Sets the time at which this page should be unpublished.
--- a/docs/editor_manual/new_pages/the_promote_tab.rst
+++ b/docs/editor_manual/new_pages/the_promote_tab.rst
@@ -1,7 +1,11 @@
-The Promote tab
-~~~~~~~~~~~~~~~
+================
+ Edit Page tabs
+================
 
 A common feature of the *Edit* pages for all page types is the two tabs at the top of the screen. The first, Content, is where you build the content of the page itself.
+
+The Promote tab
+~~~~~~~~~~~~~~~
 
 The second, *Promote*, is where you can set all the 'metadata' (data about data!) for the page. Below is a description of all default fields in the promote tab and what they do.
 
@@ -14,3 +18,12 @@ The second, *Promote*, is where you can set all the 'metadata' (data about data!
 
 .. Note::
     You may see more fields than this in your promote tab. These are just the default fields, but you are free to add other fields to this section as necessary.
+
+The Settings Tab
+~~~~~~~~~~~~~~~~
+
+The *Settings* has two fields by default.
+
+* **Go Live date/time:** Sets the time in which the changes should go live when published.  If you publish a page that has this field set to a time in the future it will be scheduled until the time comes.
+  See :ref:`scheduled_publishing` for more details.
+* **Expiry date/time:** Sets the time in which this page should be unpublished.
--- a/docs/reference/management_commands.rst
+++ b/docs/reference/management_commands.rst
@@ -13,7 +13,7 @@ publish_scheduled_pages
 
     $ ./manage.py publish_scheduled_pages
 
-This command publishes or unpublishes pages that have had these actions scheduled by an editor. It is recommended to run this command once an hour.
+This command publishes, updates or unpublishes pages that have had these actions scheduled by an editor. It is recommended to run this command once an hour.
 
 
 .. _fixtree:
--- a/docs/reference/management_commands.rst
+++ b/docs/reference/management_commands.rst
@@ -13,7 +13,7 @@ publish_scheduled_pages
 
     $ ./manage.py publish_scheduled_pages
 
-This command publishes, updates or unpublishes pages that have had these actions scheduled by an editor. It is recommended to run this command once an hour.
+This command publishes, updates or unpublishes pages that have had these actions scheduled by an editor. We recommend running this command once an hour.
 
 
 .. _fixtree:
--- a/docs/reference/pages/theory.rst
+++ b/docs/reference/pages/theory.rst
@@ -95,3 +95,26 @@ For going beyond the basics of model definition and interrelation, it might help
     #.  A response object is returned by ``serve()`` and Django responds to the requester.
 
 You can apply custom behavior to this process by overriding ``Page`` class methods such as ``route()`` and ``serve()`` in your own models. For examples, see :ref:`model_recipes`.
+
+
+.. _scheduled_publishing:
+
+Scheduled Publishing
+~~~~~~~~~~~~~~~~~~~~
+
+Page publishing can be scheduled through the *Go live date/time* feature in the *Settings* tab of the *Edit* page. This allows you to set set up initial page publishing or a page update in advance and it will only happen at the
+scheduled time.
+In order for pages to be published at the scheduled time you should set up the :ref:`publish_scheduled_pages` management command.
+
+The basic workflow is as follows.
+
+* Scheduling a revision for a page that is not currently live schedules that page to go live when the scheduled time comes.
+* Scheduling a revision for a page that is already live schedules that revision to update the page when the time comes.
+* If page has a scheduled revision and you set another revision to publish immediately the scheduled revision will be unscheduled.
+
+The *Revisions* view for a given page will show which revision is scheduled and for when it is scheduled. A scheduled revision in teh list will also provide an *Unschedule* button to cancel it.
+
+
+ 
+
+
--- a/docs/reference/pages/theory.rst
+++ b/docs/reference/pages/theory.rst
@@ -102,17 +102,16 @@ You can apply custom behavior to this process by overriding ``Page`` class metho
 Scheduled Publishing
 ~~~~~~~~~~~~~~~~~~~~
 
-Page publishing can be scheduled through the *Go live date/time* feature in the *Settings* tab of the *Edit* page. This allows you to set set up initial page publishing or a page update in advance and it will only happen at the
-scheduled time.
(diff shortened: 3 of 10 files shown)


hakjoon:
The current implementation does not execute on the `[ Live + Scheduled + Draft ]` status suggested in the RFC.  It does create a `[ Live + Scheduled ]` status but I couldn't figure out a way to see if the scheduled revision was the latest revision without some changes to the Page model which I was hesitant to do.  The behavior does work like the current `[Scheduled]` status which does not indicate if a draft exists beyond the scheduled revision.

There is an indicator in the revisions index for the scheduled revision so it is fairly easy to see where it stands in the revisions stack.

wagtail/admin/views/pages.py
```diff
@@ -1109,3 +1111,36 @@ def revisions_compare(request, page_id, revision_id_a, revision_id_b):
...
+
+
+def revisions_unschedule(request, page_id, revision_id):
+    page = get_object_or_404(Page, id=page_id).specific
+
+    user_perms = UserPagePermissionsProxy(request.user)
+    if not user_perms.for_page(page).can_unpublish():
+        raise PermissionDenied
+
+    revision = get_object_or_404(page.revisions, id=revision_id)
```
lb-: Should check here if this revision is actually a future scheduled revision.
If not - it should redirect the user back to the revisions page, maybe with a warning that there is no revision to unschedule.

lb-:
@hakjoon I took a look at this tonight, checked out the PR locally and also read through the code changes, below are some comments/thoughts/screenshots.

This is an awesome feature! Awesome work on this.

## 1. revisions_unschedule view/method should check for invalid URL

This view should check if there is an actual unschedule action that can occur, otherwise it might make for a confusing user experience in edge cases.

Let's say you happen to open the 'unschedule revision' link (or copy to clipboard), but the revision has already been unpublished when you actually open the URL. There is no indicator that you cannot do this action as it has already been done.

## 2. `Live + Scheduled + Draft` Indicator

After using this feature, I do think this is needed, but I am unsure as to the model changes needed.

## 3. More specific success message

When a page is live and you are scheduling a new page for the future, the success message says:

> Page 'About' has been scheduled for publishing. 

Maybe this should be more specific to this being a live page with a future schedule. Rough idea below.

> Page 'About' is live and this revision has been scheduled for publishing.

## Screenshots

No comments, just showing it all works as expected.

wagtail - pr4050 - new sheduled page
wagtail - pr4050 - scheduled - explorer menu
wagtail - pr4050 - draft - revisions list

hakjoon:
@lb- Thank you for the feedback.

> revisions_unschedule view/method should check for invalid URL

I think this makes sense.  I originally thought of adding this but based the view on the unpublish view which did not have a similar check, so ended up leaving it out.  My thought was that one would not get to it without having a link from a scheduled revision, but your point about bookmarks, url completion is very valid.  Feels like it should be an easy enough change.

> More specific success message

I believe this should be easy enough to accommodate.

> Live + Scheduled + Draft Indicator

I wanted to see what the general feeling was without this since we don't currently have it for scheduled pages, but I should probably go into more detail as to what I think the required changes are.

From my digging I think we would have to expose something akin to a `get_scheduled_revision` method that would return the scheduled revision.  At that point we could check if the scheduled_revision == latest_revision.  There already is an `approved_schedule` property that does an `exists()` check for a scheduled revision.  I wasn't sure if we would want to modify the semantics of that property to return the actual revision or add a second accessor. 

 A second accessor would be simple but it does start polluting the API for the page model a bit and changing the existing property does change the public API so I was hesitant to do it without some buy in. The change would allow for a `[scheduled + draft]` status for non-live pages also.

lb-:
One other thing that is needed is some additional docs relating to this feature:

* The management commands section should have a note that this includes updating live pages with future scheduled changes. http://docs.wagtail.io/en/v1.13.1/reference/management_commands.html
* Potentially a new section about scheduling in the Theory page http://docs.wagtail.io/en/v1.13.1/reference/pages/theory.html
* Start a 'Settings Tab' section on the editors guide that covers the basics of publishing in the future for new pages similar to http://docs.wagtail.io/en/v1.13.1/editor_manual/new_pages/the_promote_tab.html
* Add a new bullet point explaining updating publish date on an existing page - http://docs.wagtail.io/en/v1.13.1/editor_manual/editing_existing_pages.html

Also see #2841 for general older feedback on the docs surrounding scheduled publishing.

I can see that the docs for scheduled publishing are pretty sparse but it would be good to cover a few of these gaps now that this feature will be slightly more complex.

tomdyson:
This looks great @hakjoon! And thanks for the review, @lb-. 

Paging @Gagaro - could your impressive https://github.com/Gagaro/wagtail-calendar be aware of scheduled revisions?

Gagaro:
I'll definitely try to implement this. But I'm still missing time. I wanted to add more tests and documentation before releasing. I'll probably won't be able to do it before another few weeks.

tomdyson:
@hakjoon Happy New Year! Are you able to make a start on the docs for your changes, along the lines of @lb-'s suggestions?

hakjoon:
@tomdyson I'll probably have time to start on docs next week.

tomdyson:
@hakjoon have you had a chance to look at docs? We're really keen to get this in 2.0 if possible. Thanks!

hakjoon:
Not yet I’ve been short on time this week.

hakjoon:
Pushed an initial stab at documentation.  Fair warning I'm pretty bad at writing docs.

I also added a more specific message when scheduling a revision for a page that is already live per @lb- suggestion.

Regarding *revisions_unschedule view/method should check for invalid URL* currently if you try to unschedule a revision that is not scheduled nothing really happens since all it does is clear `approved_go_live` which in this case would not be set anyway so not sure if it is worth trying to handle the possibility of a bookmark, clipboard mishap.

docs/editor_manual/new_pages/the_promote_tab.rst
```diff
@@ -1,8 +1,12 @@
-The Promote tab
-~~~~~~~~~~~~~~~
+================
+ Edit Page tabs
+================
 
 A common feature of the *Edit* pages for all page types is the two tabs at the top of the screen. The first, Content, is where you build the content of the page itself.
```
tomdyson: two tabs -> three tabs

tomdyson: Content -> *Content*

docs/editor_manual/new_pages/the_promote_tab.rst
```diff
@@ -14,3 +18,12 @@ The second, *Promote*, is where you can set all the 'metadata' (data about data!
 
 .. Note::
     You may see more fields than this in your promote tab. These are just the default fields, but you are free to add other fields to this section as necessary.
+
+The Settings Tab
+~~~~~~~~~~~~~~~~
+
+The *Settings* has two fields by default.
```
tomdyson: *Settings* -> *Settings* tab

docs/editor_manual/new_pages/the_promote_tab.rst
```diff
@@ -14,3 +18,12 @@ The second, *Promote*, is where you can set all the 'metadata' (data about data!
...
+
+* **Go Live date/time:** Sets the time in which the changes should go live when published.  If you publish a page that has this field set to a time in the future it will be scheduled until the time comes.
```
tomdyson: time in which -> time at which
I suggest removing the second sentence "If you publish a page that has this field set to a time in the future it will be scheduled until the time comes."

docs/editor_manual/new_pages/the_promote_tab.rst
```diff
@@ -14,3 +18,12 @@ The second, *Promote*, is where you can set all the 'metadata' (data about data!
...
+  See :ref:`scheduled_publishing` for more details.
+* **Expiry date/time:** Sets the time in which this page should be unpublished.
```
tomdyson: time in which -> time at which

docs/reference/management_commands.rst
```diff
@@ -13,7 +13,7 @@ publish_scheduled_pages
 
     $ ./manage.py publish_scheduled_pages
 
-This command publishes or unpublishes pages that have had these actions scheduled by an editor. It is recommended to run this command once an hour.
+This command publishes, updates or unpublishes pages that have had these actions scheduled by an editor. It is recommended to run this command once an hour.
```
tomdyson: It is recommended to run -> We recommended running

docs/reference/pages/theory.rst
```diff
@@ -95,3 +95,26 @@ For going beyond the basics of model definition and interrelation, it might help
...
 You can apply custom behavior to this process by overriding ``Page`` class methods such as ``route()`` and ``serve()`` in your own models. For examples, see :ref:`model_recipes`.
+
+
+.. _scheduled_publishing:
+
+Scheduled Publishing
+~~~~~~~~~~~~~~~~~~~~
+
+Page publishing can be scheduled through the *Go live date/time* feature in the *Settings* tab of the *Edit* page. This allows you to set set up initial page publishing or a page update in advance and it will only happen at the
+scheduled time.
```
tomdyson: drop " and it will only happen at the scheduled time"

docs/reference/pages/theory.rst
```diff
@@ -95,3 +95,26 @@ For going beyond the basics of model definition and interrelation, it might help
...
+In order for pages to be published at the scheduled time you should set up the :ref:`publish_scheduled_pages` management command.
+
+The basic workflow is as follows.
```
tomdyson: as follows. -> as follows:

docs/reference/pages/theory.rst
```diff
@@ -95,3 +95,26 @@ For going beyond the basics of model definition and interrelation, it might help
...
+
+* Scheduling a revision for a page that is not currently live schedules that page to go live when the scheduled time comes.
```
tomdyson: schedules that page to go live -> means that page will go live

docs/reference/pages/theory.rst
```diff
@@ -95,3 +95,26 @@ For going beyond the basics of model definition and interrelation, it might help
...
+* Scheduling a revision for a page that is already live schedules that revision to update the page when the time comes.
```
tomdyson: schedules that revision to update the page -> means that revision will be published

docs/reference/pages/theory.rst
```diff
@@ -95,3 +95,26 @@ For going beyond the basics of model definition and interrelation, it might help
...
+* If page has a scheduled revision and you set another revision to publish immediately the scheduled revision will be unscheduled.
```
tomdyson: immediately the scheduled -> immediately, the scheduled

docs/reference/pages/theory.rst
```diff
@@ -95,3 +95,26 @@ For going beyond the basics of model definition and interrelation, it might help
...
+
+The *Revisions* view for a given page will show which revision is scheduled and for when it is scheduled. A scheduled revision in teh list will also provide an *Unschedule* button to cancel it.
```
tomdyson: for when it is scheduled -> when it is scheduled for
teh list -> the list :)

hakjoon:
@tomdyson thanks for the edits.  Changes pushed.

docs/reference/pages/theory.rst
```diff
@@ -102,17 +102,17 @@ You can apply custom behavior to this process by overriding ``Page`` class metho
 Scheduled Publishing
 ~~~~~~~~~~~~~~~~~~~~
 
-Page publishing can be scheduled through the *Go live date/time* feature in the *Settings* tab of the *Edit* page. This allows you to set set up initial page publishing or a page update in advance and it will only happen at the
+Page publishing can be scheduled through the *Go live date/time* feature in the *Settings* tab of the *Edit* page. This allows you to set set up initial page publishing or a page update in advance.
 scheduled time.
```
tomdyson: 'scheduled time.' is still hanging

hakjoon: argh, fix pushed.

gasman:
Merged in 8c0b9b0bc13c59a12c8c7ad730930e1c236373f5 + parents - thanks @hakjoon!

PR #4053, opened 2017-11-23
Fix compare page with non-field models in base_form_class

### LINKED ISSUE #3737: FieldDoesNotExist on Customising generated forms
**### Issue Summary**
I'm extending `WagtailAdminPageForm` using the sample code in the docs at https://goo.gl/X9DU4P but getting and the following error when comparing revisions. 

**FieldDoesNotExist at /admin/pages/15/revisions/compare/53...54/**
**EventPage has no field named 'address'**

**### Steps to Reproduce**
1. Use the exact sample code found in the linked docs example above.
2. Fix minor issues: add max_length to location, remove geocoder reference from form.
2. Makemigrations and migrate
3. Create an instance of Event page and save at least 2 drafts.
4. On the revisions list page, click "Compare with Previous Verison"

The error is happening in **wagtail/wagtailadmin/edit_handlers.py** line 503:
```
field = cls.model._meta.get_field(cls.field_name) 
```
and exception raised in **django/db/models/options.py** line 619:
```
raise FieldDoesNotExist('%s has no field named %r' % (self.object_name, field_name))
```

I'm able to solve this issue by adding an 'address' field to the model, but I need that field namespace for what I'm doing. How can I workaround this issue with revisions compare feature intact?

**### Technical details**
* Python version: Python 3.4.3
* Django version: 1.10.7
* Wagtail version: 1.11.1
* Browser version: Chrome 59

lb-:
Fixes #3737 

## Detail

* When using a custom form class and overriding a page model's `base_form_class`, often the use case is to add some field that is not stored on the actual model, just shown in the new/edit view.
* When comparing revisions, we would try to compare these non-model fields from previous revisions causing an error because the field did not exist.

## Checklist

* Added testing for this specific issue
* Linting and tests pass
* No additional feature, no docs needed


gasman:
ping @BertrandBordage - it looks like #3810 may have broken the ability to use non-model fields in `FieldPanel`s, since the model lookup now happens in `on_model_bound` (and thus fails during startup, rather than just in the comparison view as addressed in this PR). I guess the fix is to implement `db_field` as a cached property instead, so that it gets evaluated lazily...

gasman:
Now rebased at #4470

PR #4059, opened 2017-11-23
Adds search query API & implements at least PlainText & MatchAll.

BertrandBordage:
This work is a first step at implementing everything from the [search queries classes RFC](https://github.com/wagtail/rfcs/blob/master/text/010-search-query-classes.md).

The scope of this pull request it to translate existing functionalities to the new query API while keeping the existing one working. The goal is not to add any functionality, but to **allow** new functionalities to be added.

This pull request will be used first for implementing autocomplete in Wagtail, using the `Prefix` query class (the name may change or have an `Autocomplete` alias).

The query classes in themselves don’t do anything, they are just declarative. It’s up to the backends to implement how they are used. My initial thinking was to create something like the Django database expressions/functions/aggregates, where the logic is written in it. This is possible with SQL since databases work roughly the same with a few exceptions. Here, we have totally different backends and we can’t really conciliate the way they work.

I’m thinking of a registering system to create an extensible way to add new query classes and logic to each separate backend, but I’m not sure it’ll work properly. It may not be included in this pull request.


BertrandBordage:
I add the “Needs work” label as I’m still working on it, especially refactoring the existing `BaseSearchQuery` and its subclasses in order to make more sense. By the way, I’m going to rename that to `QueryCompiler` or something like that to make it more clear, now that we have a `SearchQuery` class in `wagtail.wagtailsearch.query`.

But it can very well be merged already, it is fully working. I’ll try to make each new commit fully working as well, so @kaedroho can review it whenever he wants and merge it. I’ll put future changes in another pull request if that’s the case.

BertrandBordage:
Just finished this pull request, it’s ready for review!

Most query classes are already implemented on the two database backends, PostgreSQL as well as the basic database one. To sum up, here are the supported query classes on different backends:
- PostgreSQL: `And`, `Or`, `Not`, `MatchAll`, `Term`, `PlainText`, `Filter`, `Boost`.
- Database: `And`, `Or`, `Not`, `MatchAll`, `Term`, `PlainText`, `Filter`, `Boost`.
- ElasticSearch: `MatchAll`, `PlainText`.

The operators `~`, `&` and `|` are also supported. So we can already use queries as complex as this:
```python
YourModel.objects.search(
    ((Term('Javascript') & ~Term('Definitive')) |
    Term('Python') | Term('Rust')) |
    Term('Two'))
```

The supported functionalities are all tested, except for boosts that have no impact for now on results. I put a `TODO` to implement it later for PostgreSQL. A warning is also shown in the console when using boosting.

Of course, an error is raised when you try to search using a query class unsupported in a given backend. That way, we’ll be able to add further functionalities while clearly informing users that they can’t use this functionality.

BertrandBordage:
(Just discovered that the PostgreSQL search backend was not run since 67b1ddb665c2688a978ee61936d43010cf687620, and then fixed it with 420f721.)

BertrandBordage:
Rebased against the latest master, with Python 3 and module names changes.

gasman:
Marking as release blocker as it includes the fix for #4120.

wagtail/search/query.py
```diff
@@ -0,0 +1,194 @@
...
+
+    def apply(self, func):
+        return func(self.__class__(
+            self.query.apply(func),
+            self.include.apply(func), self.exclude.apply(func)))
+
+    def get_equivalent(self):
+        query = self.query
+        if self.include is not None:
+            query &= self.include
```
kaedroho: This should be ``query &= Boost(self.include, 0.0)`` as the score from the "include" subquery should be ignored. We might want to do this to "exclude" as well (in case of ``Filter(query, exclude=Not(other_query)``)

wagtail/search/tests/elasticsearch_common_tests.py
```diff
@@ -167,5 +170,58 @@ def test_slice_to_next_page(self):
...
+    @unittest.expectedFailure
+    def test_delete(self):
+        super().test_delete()
+
+    #
+    # Basic query classes
+    #
+
+    # Not implemented yet
+    @unittest.expectedFailure
```
kaedroho: I think this should be implemented for Elasticsearch as well before merging into master

wagtail/wagtailsearch/tests/elasticsearch_common_tests.py
```diff
@@ -8,9 +8,10 @@
 
 from wagtail.tests.search import models
 from wagtail.wagtailsearch.query import MATCH_ALL
+from wagtail.wagtailsearch.tests.test_backends import BackendTests
 
 
-class ElasticsearchCommonSearchBackendTests(object):
+class ElasticsearchCommonSearchBackendTests(BackendTests):
```
kaedroho: What's the reason for changing this?

BertrandBordage:
Thanks for the review @kaedroho!
I updated the pull request and added boost to `Filter`, as suggested.

About implementing `Term`, I don’t know Elasticsearch enough to do it. I implemented `PlainText` and `MatchAll` so that we have at least exactly the same features as before in Elasticsearch. Implementing `Term` means we can do things like `Term(a) & (Term(b) | Term(c))`, which isn’t possible for now with the Elasticsearch backend, as far as I know. Feel free to make a new pull request containing the implementation of `Term` :)

BertrandBordage:
Apart from that, does the API works as you expected?

kaedroho:
@BertrandBordage API looks good to me! I'll look into Elasticsearch support today

wagtail/search/tests/test_backends.py
```diff
@@ -428,6 +428,234 @@ def test_delete(self):
...
+    def test_term(self):
+        # Single word
+        results = self.backend.search(Term('Javascript'),
+                                      models.Book.objects.all())
+        self.assertSetEqual({r.title for r in results},
+                            {'JavaScript: The Definitive Guide',
+                             'JavaScript: The good parts'})
+
+        # Multiple word
+        results = self.backend.search(Term('Definitive Guide'),
```
kaedroho: This shouldn't match anything as there "Definitive" and "Guide" are separate terms

wagtail/search/tests/test_backends.py
```diff
@@ -428,6 +428,234 @@ def test_delete(self):
...
+                             'Learning Python',
+                             'The Two Towers',
+                             'The Rust Programming Language',
+                             'Two Scoops of Django 1.11'})
+
+    #
+    # Shortcut query classes
+    #
+
+    def test_plain_text(self):
```
kaedroho: Can we break these tests down a bit? Ideally, we should have just one call to ``self.backend.search`` per test

wagtail/search/tests/test_backends.py
```diff
@@ -428,6 +428,234 @@ def test_delete(self):
             "The Fellowship of the Ring"
         ])
 
+    #
+    # Basic query classes
```
kaedroho: Can we move these tests into a mixin class add only add it on to backends that support the entire API?

BertrandBordage: Sure, I’ll do that.

wagtail/search/tests/test_backends.py
BertrandBordage: From a semantic point of view, I agree.
From a pragmatic point of view, I think it’s a bad idea to make it return nothing, for two reasons:
- This means we have to define a term separator, which is very far from being easy. This is usually up to the search engine to separate terms, and this is not a simple split by spaces. A few examples:
  - “rendez-vous” is a single term but “1980-2020” are two terms in most configuration
  - “can/must” are two terms but “github.com/wagtail/wagtail” is a single term
  - “Hello. Goodbye” are two terms but “127.0.0.1” is a single term
- To me, `Term` works a bit like a shell argument in `subprocess`. If you pass `Definitive Guide` as an argument to `subprocess`, it will not pass it as two separate argument, but as a single argument containing a space. Here, that’s exactly what we do. Additionally in the future, we may change that to a phrase query so it matches the order of words, like when searching `"Definitive Guide"` instead of `Definitive Guide` in Google.

wagtail/search/tests/test_backends.py
BertrandBordage: Sure, will do.

wagtail/search/tests/test_backends.py
kaedroho: > To me, Term works a bit like a shell argument in subprocess. If you pass Definitive Guide as an argument to subprocess, it will not pass it as two separate argument, but as a single argument containing a space. Here, that’s exactly what we do. Additionally in the future, we may change that to a phrase query so it matches the order of words, like when searching "Definitive Guide" instead of Definitive Guide in Google.

That's exactly how I see it as well. But in the index, those two words would be split into separate terms, so the term query will not match either:

 - ``Definitive Guide``≠ ``Definitive``
 - ``Definitive Guide`` ≠ ``Guide``

When users are using this API, they should be aware about how the backend they are using is tokenising the text for their language. I don't think we should try to be clever here.

BertrandBordage: Hmmm… Then maybe we’ll have a difference of behaviour between backends, here. In database backends, it doesn’t work this way. For example, in PostgreSQL if you do `to_tsquery('''Definitive Guide''')`, it returns the same query as `to_tsquery('Definitive & Guide')`.

kaedroho: Isn't that the same as the ``PlainText`` query? Maybe we shouldn't support the term-level query types on database backends if they don't give that level of control.

kaedroho:
Elasticsearch support added in: https://github.com/kaedroho/wagtail/tree/search-query-api-elasticsearch

Just need to add tests checking the query that gets generated and figure out what's up with the filter test.

BertrandBordage:
Wow, you already added support for all query classes? That’s awesome! :D
Do you want me to merge your commits in my pull request?

wagtail/search/tests/test_backends.py
BertrandBordage: Yeah, and if we do that we should change as well `PlaintText`, as it splits terms in Python and provides an `operator` argument on which we have no control in `plainto_tsquery`. I made that splitting a bit quickly so it would easily work the same on all backends, but I’m not even sure that’s how to do it in Elasticsearch.

kaedroho: For Elasticsearch, ``PlainText`` queries should map directly to a ``match`` query to avoid introducing any regressions (as that's what we do currently). ES also handles generating bigrams for CJK languages and could also have some user-specified behaviour defined in settings that we wouldn't want to break.

kaedroho: It might be better long-term to support doing analysis on the Python side so all backends are analysed the same and it would be much easier to configure. But that's an issue for another day :)

PR #1289, opened 2015-05-11
Preventing thousands seperators when printing max filesize in JS

signalkraft:
When using l10n with thousands seperators and large file sizes, the `max_filesize` JS property in `templates/wagtailimages/multiple/add.html` can end up as `max_file_size: 10.485.760`. That isn't valid JS and breaks the upload form.

PR #4062, opened 2017-11-24
319 make gravatar optional

### LINKED ISSUE #319: Uploadable user avatars falling back to Gravatar.
Gravatar has become considerably less slick since Wordpress became a requirement. Use of gravatar should really only be a fall back anyway.
- Avatars should be uploaded direct to Wagtail, falling back on Gravatar, then none at all.
- Other services could be considered but ideally we don't want monsterid, wavatar, identicon or any other non-photographic avatars that would detract too much aesthetically.
- Should be stored as part of a "user profile" (see: https://github.com/torchbox/wagtail/issues/248)
- Should be prompted on first login or similar (see: https://github.com/torchbox/wagtail/issues/575)

Somewhat dependent on #755

willbarton:
Fixes #319 and #1289, supersedes #3424.

This is a continuation of @chimeno's work in https://github.com/wagtail/wagtail/pull/3424. I have:

- Rebased @chimeno's work on current master and addressed all conflicts
- Changed all instances of `gravatar_url` in the admin templates

Preview images (from #3424, still relevant):

image

image


wagtail/users/migrations/0007_userprofile_avatar_choice.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.5 on 2017-11-24 14:18
+from __future__ import unicode_literals
```
m1kola: It seems to me that we don't really need `__future__` imports anymore: we are going to drop python 2 support.

m1kola: It seems to me that we don't really need `__future__` imports anymore: we are going to drop python 2 support.

wagtail/admin/templates/wagtailadmin/account/change_avatar.html
```diff
@@ -0,0 +1,27 @@
...
+        <form action="{% url 'wagtailadmin_account_change_avatar' %}" method="POST" enctype="multipart/form-data">
+            {% csrf_token %}
+
+                <ul class="fields">
+                    {% for field in form %}
+                        {% include "wagtailadmin/shared/field_as_li.html" with field=field %}
+                    {% endfor %}
+                </ul>
+
+            <li class="submit"><input type="submit" value="{% trans 'Update' %}" class="button" /></li>
```
m1kola: Few minor things:

* `<ul>` block should on the same level with `{% csrf_token %}` (remove 4 spaces from left)
* `<li>` without `<ul>` or `<ol>` doesn't look like a valid HTML

wagtail/users/models.py
```diff
@@ -43,5 +81,30 @@ def get_preferred_language(self):
...
+
+        if self.avatar_choice == self.GRAVATAR and self.user.email:
+            return get_gravatar_url(self.user.email, default=None, size=50)
+
+        return self.default_avatar
+
+    def save(self, *args, **kwargs):
+        if self.avatar:
+            this = UserProfile.objects.get(pk=self.pk)
+            this.avatar.delete(save=False)
```
m1kola: Probably it's not very critical, but there is a (small) chance that we will delete a file, but something will fail in the parent `save` method. So we will have a link to the old avatar file in the database, but we would not be able to recover a file.

wagtail/users/migrations/0007_userprofile_avatar_choice.py
BertrandBordage: Nor `coding: utf-8`, UTF-8 being the default encoding in Python 3.

wagtail/admin/templates/wagtailadmin/account/account.html
```diff
@@ -10,11 +10,11 @@
         <ul class="listing">
             <li class="row row-flush">
                 <div class="col6">
-                    <a href="https://gravatar.com/emails/" class="button button-primary">{% trans "Set gravatar" %}</a>
+                    <a href="{% url 'wagtailadmin_account_change_avatar' %}" class="button button-primary">{% trans "Set profile picture" %}</a>
                 </div>
 
                 <small class="col6">
-                    {% trans "Your avatar image is provided by Gravatar and is connected to your email address. With a Gravatar account you can set an avatar for any number of other email addresses you use." %}
```
lb-: Suggestion - would it be possible to provide the original Gravatar instructions if the User Model has `avatar_choice` set to `gravatar`. Simply because Gravatar can be confusing and having this detail has helped in the past.

wagtail/admin/templatetags/avatar.py
```diff
@@ -0,0 +1,18 @@
+from __future__ import absolute_import, unicode_literals
```
gasman: I suggest we move this tag into `wagtailadmin_tags`, to avoid the possibility of clashing with a tag library with the name `avatar` within a user's project code. (I'm pretty sure gravatar.py was only a separate module in the first place due to casual copy-and-pasting of code found on the internet...)

tomdyson:
> Since avatar_choice defaults to DEFAULT, users who were previously happily using Gravatar images will find that they suddenly disappear on upgrading to Wagtail 2.0.

Shouldn't we default to `gravatar` instead?

gasman:
Yep, we definitely ought to pick up gravatars in the out-of-the-box configuration, so that we're not losing any existing avatars. However, since (as far as I can see) there's no reason at all that a _new_ user would choose to go through Gravatar rather than direct image upload, I think there's a good case for removing it as a user-facing option.

That leaves us with 'default' and 'custom', which I'd argue presents a psychological barrier to uploading an avatar. Obviously there has to be an initial 'no avatar chosen' state, but labelling it 'default' makes it sound like that's the normal state, and 'custom' is doing something tricksy and unconventional - when really 'custom' is the option we'd like everyone to go with.

As currently implemented, the 'default' option is redundant anyhow, since you can achieve the same result by clearing the image field. (I think we could get away with removing this too, though - I think it's standard enough behaviour on websites that once you've chosen an avatar, you can freely change it but not revert to an avatar-less state.)

So, my suggestion is:

* drop the 'Select profile picture type' options and just have an image upload field
* Whenever no image file is specified, fall back on Gravatar, and have Gravatar fall back on the 'mystery man' icon as we do now
* if Gravatar is disabled in settings, fall back directly to a local copy of the 'mystery man' icon instead. (Perhaps we can do this by incorporating #4171, and having a special value of `WAGTAIL_GRAVATAR_PROVIDER_URL = None` to indicate "don't use a Gravatar service")
* Optionally, remove the 'clear' option from the image upload field

The downside here is that the interaction between Gravatar and locally-uploaded images is a bit odd, since we never explicitly tell the user "we picked this image up from Gravatar". (If we leave out the 'clear' option, there's no way to go back to the Gravatar image once it's been changed; if we include the 'clear' option, the Gravatar image will keep coming back and it won't be clear why.) I think we can live with that, though...

gasman:
Have now rebased at https://github.com/gasman/wagtail/commits/willbarton-319_make_gravatar_optional , making the UI improvements proposed above (removing the explicit default vs custom vs gravatar selection) along with various other fixes. This also incorporates #4171 along with the ability to disable gravatar completely by setting `WAGTAIL_GRAVATAR_PROVIDER_URL = None`.

One outstanding task is fixing the logic for deleting old avatar files - the old code was deleting the last-used avatar on every save, regardless of whether a new one had been uploaded. Will pick this up after 2.0.

willbarton:
@gasman I've pulled in your rebase. I'll work on the deleting old avatars code.

gasman:
Superseded by #4462

PR #4064, opened 2017-11-26
Replaced FormView.get_form_kwargs with a super-call in modeladmin/vie…

atombrella:


atombrella:
@gasman Rebased after the modules have been reorganized. I was thinking to use `super()` but it's probably better to update all of them in one go.

PR #4065, opened 2017-11-26
Removed use of six

atombrella:
Per deprecation of Python 2, this seems to have been overlooked.


atombrella:
Perhaps the "Don't inherit from object" should have been in a different PR.

gasman:
Not overlooked, just a task on the todo list we hadn't reached yet (#3917) - so, thanks very much! :-) Rebased and merged in cde0fafe7c2abf14622dc5c2713c274e4158dc28.

Issue #3918, opened 2017-10-09
Reorganise module names


gasman:
Top-level issue to track progress in renaming modules as proposed in https://github.com/wagtail/rfcs/blob/master/text/012-rationalise-apps.md - further tickets may be opened and linked here to track sub-tasks.

- [x] Create Wagtail 2.0 branch of bakerydemo
- [x] Update vagrant-wagtail-develop to use that branch
- [ ] Move each module in turn, updating documentation and bakerydemo at the same time:
  - [ ] wagtail.wagtailcore -> wagtail.core
  - [ ] wagtail.wagtailadmin -> wagtail.admin
  - [ ] wagtail.wagtaildocs -> wagtail.documents
  - [ ] wagtail.wagtailembeds -> wagtail.embeds
  - [ ] wagtail.wagtailimages -> wagtail.images
  - [ ] wagtail.wagtailsearch -> wagtail.search
  - [ ] wagtail.wagtailsites -> wagtail.sites
  - [ ] wagtail.wagtailsnippets -> wagtail.snippets
  - [ ] wagtail.wagtailusers -> wagtail.users
  - [ ] wagtail.wagtailforms -> wagtail.contrib.forms
  - [ ] wagtail.wagtailredirects -> wagtail.contrib.redirects
  - [x] wagtail.contrib.wagtailapi -> remove completely
  - [ ] wagtail.contrib.wagtailfrontendcache -> wagtail.contrib.frontend_cache
  - [ ] wagtail.contrib.wagtailroutablepage -> wagtail.contrib.routable_page
  - [ ] wagtail.contrib.wagtailsearchpromotions -> wagtail.contrib.search_promotions
  - [ ] wagtail.contrib.wagtailsitemaps -> wagtail.contrib.sitemaps
  - [ ] wagtail.contrib.wagtailstyleguide -> wagtail.contrib.styleguide

Since this is likely to be a large and disruptive change to the codebase, we should aim to keep individual PRs as small and focused as possible, and review them promptly.

chosak:
@gasman @kaedroho how do you want to handle migrations? The RFC mentions a script to help with renaming module imports, but doesn't mention data migrations.

[...]

PR #4066, opened 2017-11-26
Removed __future__ imports

atombrella:
If Python 2 is not supported anymore, these shouldn't be here.

I left the 1.8 release notes untouched.


BertrandBordage:
Thanks a lot @atombrella! :D
I was about to do it.
I scanned your pull request quickly, it seems fine, we can merge it once the test suite passes.

gasman:
@BertrandBordage Please don't merge right now - I'm in the middle of dealing with #3918 :-)

BertrandBordage:
Acknowledged, @gasman!

wagtail/wagtailadmin/datetimepicker.py
```diff
@@ -1,8 +1,7 @@
-from __future__ import absolute_import, unicode_literals
-
-
 # Adapted from https://djangosnippets.org/snippets/10563/
 # original author bernd-wechner
+
+
```
BertrandBordage: Why are these newlines added?

atombrella: `flake8` wasn't happy.

BertrandBordage: OK then.
I didn’t remember it forced spaces between comments & functions, that’s not specified in PEP8 as far as I remember.

atombrella: "Surround top-level function and class definitions with two blank lines." Doesn't it fall under this rule? I deleted the imports in the top, so the comments may be regarded as "top-level function".

atombrella:
Hmm, Travis has some issues. But since it shouldn't be merged right away, I'll just let it sit.

I didn't change Github's default when I created the PR, and thus maintainers can push to my branch to trigger a new build on Travis when it's safe to merge.

wagtail/wagtailadmin/datetimepicker.py
BertrandBordage: This doesn’t make sense to separate a comment from the function/class it’s about.
This rule is not an absolute rule to follow: for example, we don’t put two newlines at the end of the file after a class or function definition.
PEP8 itself shows some comments just above function definitions in this part: https://www.python.org/dev/peps/pep-0008/#code-lay-out

gasman:
Thanks again @atombrella! Merged in f0eef2fc88043ba627b89125876f1cfe53b7fc51.

PR #4067, opened 2017-11-26
Don't inherit from object

atombrella:


atombrella:
@gasman Rebased after the modules have been reorganized.

PR #4068, opened 2017-11-26
Deleted definitions of __ne__

atombrella:


atombrella:
Rebased after the modules have been reorganized. It's Python 2 legacy code.

gasman:
Thanks again!

PR #4072, opened 2017-11-27
Fix 3997 Document / Image search on query change only

### LINKED ISSUE #3997: Images & Documents Search - Search Query should not activate on empty searches

### Issue Summary

When navigating through collections of Images, the search field (which is focused by default) is very sensitive to empty search changes (such as pressing the `down key`) which then clear the current page and collection you are on.

### Steps to Reproduce

It's essential that you provide enough information for someone else to replicate the problem you're seeing. Simply describing something that's broken on your current project is not enough!

1. Load up the wagtaildemo
2. Go to Admin > Settings > Collection & Click 'Add a Collection' (create one - can be any name)
3. Add 30 images, 15 in the 'Root' Collection and 15 in the new connection you just created.
4. Go to the images page `/admin/images/`
5. Select the non-root collection, URL should update to something like `/admin/images/?collection_id=2`
6. Scroll to the bottom of the page and click 'Next' to go to page two. URL should update to something like `/admin/images/?collection_id=2&p=2` (Note: This step is not needed, just explains workflow).
7. Click the down arrow on your keyboard to move down the page.
7a. **Expected**: Either nothing (because the search box is in focus) or page moves down
7b. **Actual**: A search is triggered in the search field, this causes a search for 'nothing' which reloads the URL to `/admin/images/?q=`

The main issue here is now the user is very confused, they were on page two of a specific collection and now they suddenly ended up back at the home page of the images?

This also happens on Documents.

### Technical details

* Python version: 2.7.12
* Django version: Django v.1.11.6
* Wagtail version: 1.13
* Browser version: Chrome 61

### Potential Solution (basic)

It appears that the event `keyup cut paste` is causing this search trigger here.
`$(window.headerSearch.termInput).on('keyup cut paste', function() {...`
[wagtailadmin/static_src/wagtailadmin/js/core.js#L222](https://github.com/wagtail/wagtail/blob/7a1455551bc83a60def638b3351a641035a95d64/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js#L222)

Maybe in the search function we could add a check for empty values?

```
function search() {
    var searchInputValue = $(window.headerSearch.termInput).val();
    // if searchInputValue is not null, or contains more than an empty string.
    if (searchInputValue) {
    ... // rest of the function that exists already
    }
}
```

I am sure there is a more elegant javascript way to do this but it will mean we won't get false triggers.

### Potential Solution (advanced)

Along with ignoring the query if it is blank, we should consider the currently selected collection which is cleared whenever a new search takes place.

ie. URL should become

`/admin/images/?collection_id=2&q=my+query`

NOT

`/admin/images/?q=my+query`

Along with considering the current search query when the collection is changed.

However, these are probably two extra issues that need to be created.

lb-:
Fixes #3997 by ensuring that a new ajax search query (+ history push state) only triggers when the search query actually changes. eg pressing the 'up' key does not change the query but adding a letter does.

Also, added the search function to trigger on input change, I noticed that if you select a value from the browser auto-fill.

JS linting run, tested on Chrome 62, Firefox 48.0.2 and Safari 11.0.1 all on macOS 10.12.6

### DIFF
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
@@ -219,7 +219,7 @@ $(function() {
         var searchCurrentIndex = 0;
         var searchNextIndex = 0;
 
-        $(window.headerSearch.termInput).on('keyup cut paste', function() {
+        $(window.headerSearch.termInput).on('keyup cut paste change', function() {
             clearTimeout($.data(this, 'timer'));
             var wait = setTimeout(search, 200);
             $(this).data('timer', wait);
--- a/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
+++ b/wagtail/wagtailadmin/static_src/wagtailadmin/js/core.js
@@ -251,18 +251,18 @@ $(function() {
                     },
                     complete: function() {
                         $(window.headerSearch.termInput).parent().removeClass(workingClasses);
-                      }
+                    }
                 });
             }
-      };
+        };
 
-      getURLParam = function(name) {
-          var results = new RegExp('[\?&]' + name + '=([^]*)').exec(window.location.search);
-          if (results) {
-              return results[1];
-          }
-          return '';
-      }
+        getURLParam = function(name) {
+            var results = new RegExp('[\?&]' + name + '=([^]*)').exec(window.location.search);
+            if (results) {
+                return results[1];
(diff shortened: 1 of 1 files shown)


jjanssen:
Thanks @lb- I'll have a look at it tomorrow.

jjanssen:
Thanks! Merged with 7c54c14

PR #4077, opened 2017-11-29
Update setup.cfg to drop support of Python 2

m1kola:
In #3922 we've removed classifiers for Python 2 from the `setup.py`, but `setup.cfg` also needs to be updated: we don't need to build universal wheel anymore.

Related to #3917


gasman:
Looks good to me, thanks!

PR #4082, opened 2017-11-30
Fix typo for wagtailsnippets JS

luizboaretto:
Currently, it's not possible to select a snippet because the JS file is not loaded, resulting in the error:
ReferenceError: createSnippetChooser is not defined

This is due to a typo in the gulp config file.


gasman:
Oops - good catch, thanks @luizboaretto!

PR #4089, opened 2017-12-01
Use hook_output tag via separate template include. Fixes #4080

### LINKED ISSUE #4080: Missing file in insert_editor_js breaks page editor
### Issue Summary

If a non-existent file is referenced from an `insert_editor_js` hook then django won't render a large part of the page editor template at all. There's no indication of what caused this thanks to django suppressing exceptions from template includes. 

### Steps to Reproduce

1. reference a non-existent file in an `insert_editor_js` hook:
```
# wagtail_hooks.py
from django.contrib.staticfiles.templatetags.staticfiles import static
from django.utils.safestring import mark_safe

@hooks.register('insert_editor_js')
def editor_js():
    return mark_safe('<script src="%s"></script>' % static('js/does-not-exist.js'))
```
2. set `DEBUG=False`
3. run collectstatic
4. try to edit a page: there will be lots of javascript errors on the console and e.g. image choosers won't work

### More details

When developing locally with `DEBUG=True` then `django.contrib.staticfiles.templatetags.staticfiles.static` won't raise an exception for missing files, instead it will return the path even if the file doesn't exist. This causes no other issues at all, it will be just a 404 later on.

However with `DEBUG=False` and using the manifest storage `django.contrib.staticfiles.templatetags.staticfiles.static` does raise an exception via `{% hook_output 'insert_editor_js' %}`: https://github.com/wagtail/wagtail/blob/master/wagtail/admin/templates/wagtailadmin/pages/_editor_js.html#L28
Which is currently ignored by django if `DEBUG=False` but at least this behaviour will change in django 2.1: https://docs.djangoproject.com/en/1.11/ref/templates/builtins/#include
Moreover, since `hook_output` is in a template include this exception will turn the whole `_editor_js.html` template into an empty string, not just the `hook_output` tag, omitting 10 more javascript files that are required for the page editor to work.

Maybe this doesn't affect many people and it really isn't Wagtail's fault but it isn't obvious at all why suddenly the page editor is broken when there are no exceptions raised with either `DEBUG=True/False`. I'm not sure if there's a good fix for this, or even if it's worth addressing it until django 2.1 but at least it might come in handy for others having the same issue.

### Technical details

* Django version: 1.11.5
* Wagtail version: 1.13.1

balazs-endresz:


lb-:
@balazs-endresz thank you for this, would you be able to add some tests also?

Ideally amend the existing tests for hooks to check that this new template is used and also write a new one that covers the include template crash case, asserting that the other header content is still in the rendered template.

gasman:
wontfix, as per discussion on #4080.

PR #4092, opened 2017-12-02
Removed some unnecessary list-calls in iterations

atombrella:


gasman:
Hi @atombrella,
As the test failures show, these (or at least some of them) are required because we're modifying the collection as we iterate over it - so we need to make a copy of the sequence first.

PR #4093, opened 2017-12-03
Rework method for gathering table block defaults

### LINKED ISSUE #3948: TableBlock cell content alignment context menu has returned
### Issue Summary

The `TableBlock` Alignment context menu item has returned, following removal in Wagtail 1.6 http://docs.wagtail.io/en/stable/releases/1.6.html#bug-fixes

Might be due to this commit: https://github.com/wagtail/wagtail/commit/6e42a7755ada9ba162e78af6535924449ce3287

### Steps to Reproduce

- Add the `table_block` contrib module to a project http://docs.wagtail.io/en/stable/reference/contrib/table_block.html
- Add the `TableBlock` to a StreamField
- Create/edit a page in Wagtail, add a `TableBlock` and right click on a cell
screen

### Technical details

* Python version: 3.4.2
* Django version: 1.11.4
* Wagtail version: 1.12.2
* Browser version: Chrome 64

lb-:
Closes #3948 by moving all table block defaults to the server only, when `contextMenu` is set to True, it actually uses our intended default without the alignment menu.

Other changes
* Moved default functionality to own method `get_table_defaults` to make testing easer and the __init__ method cleaner.
* Moved the main default options to standalone variable in the module.
* Added a heap of tests for intended functionality for defaults and a basic test for a page with table block edit view.
* Removed the javascript setting of contextMenu default, this is now in the server and it seemed better to keep as many defaults as possible in the server code.
* No docs needed as this reverts to intended functionality where alignment menu is not visible by default.


lb-:
@loicteixeira would love to get your thoughts on this one when you get the chance.

wagtail/contrib/table_block/tests.py
```diff
@@ -193,3 +200,163 @@ def test_render_with_extra_context(self):
...
+        self.assertEqual(DEFAULT_TABLE_OPTIONS['startRows'], block.get_table_options()['startRows'])
+        self.assertEqual(DEFAULT_TABLE_OPTIONS['startRows'], block.table_options['startRows'])
+        self.assertEqual(DEFAULT_TABLE_OPTIONS['colHeaders'], block.get_table_options()['colHeaders'])
+        self.assertEqual(DEFAULT_TABLE_OPTIONS['colHeaders'], block.table_options['colHeaders'])
+        self.assertEqual(DEFAULT_TABLE_OPTIONS['contextMenu'], block.get_table_options()['contextMenu'])
+        self.assertEqual(DEFAULT_TABLE_OPTIONS['contextMenu'], block.table_options['contextMenu'])
+        self.assertEqual(DEFAULT_TABLE_OPTIONS['colHeaders'], block.get_table_options()['colHeaders'])
+        self.assertEqual(DEFAULT_TABLE_OPTIONS['colHeaders'], block.table_options['colHeaders'])
+        self.assertEqual(DEFAULT_TABLE_OPTIONS['stretchH'], block.get_table_options()['stretchH'])
+        self.assertEqual(DEFAULT_TABLE_OPTIONS['stretchH'], block.table_options['stretchH'])
```
loicteixeira: Checking both `block.get_table_options()` and `block.table_options` is a bit heavy, probably unnecessary\* and probably error prone. As a matter of fact, you check `colHeaders` twice (lines 240-241 and 244-245). Should the second occurrences be `rowHeaders`.

\* There is:
- `self.assertEqual(block.table_options, block.get_table_options())` above so checking either one should be enough
- `self.assertEqual(self.default_table_options, block.table_options)` already check that the defaults are conserved.

wagtail/contrib/table_block/tests.py
```diff
@@ -193,3 +200,163 @@ def test_render_with_extra_context(self):
...
+    def test_tableblock_render_form(self):
+        """
+        Test the rendered form field generated by TableBlock.
+        """
+        block = TableBlock()
+        html = block.render_form(value=self.value)
+        self.assertIn('<script>initTable', html)
+        self.assertIn('<div class="field char_field widget-table_input">', html)
+        # check that options render in the init function
+        self.assertIn(json.dumps(block.table_options), html)
```
loicteixeira: Will the keys be ordered consistently in the dumps? I'm worried the test might fail randomly (but if we're sure it doesn't, that's all fine).

lb-:
@loicteixeira thanks for that input - I have cleaned up the tests. Should I merge this in to master myself? (assuming the CI passes).

loicteixeira:
Please do 👍

lb-:
Merged

PR #4100, opened 2017-12-04
[fix] wagtailcore.query.specific_iterator: check if value is present

WnP:
**Warning** This PR aim to be merged on `stable/1.12.x` branch

## Description

[wagtailcore.query.specific_iterator](https://github.com/wagtail/wagtail/blob/stable/1.12.x/wagtail/wagtailcore/query.py#L360) should, IMHO, check if pk is present in `pages_by_type[content_type]` before returning the value.

We've got a use case where we limit platform's features by returning empty query set for models implied in disabled features, this is done at the model manager level.

This check prevent `specific_iterator` raising `KeyError` while generating sitemap for example.

IMHO, using get method on dictionary keys is a good practice.

Let me know your through.

## Q/A

* Do the tests still pass?

Yes

Yes

Not ATM, let me know if you want a test for a simple fix like that.

It's not a new feature.

### DIFF
--- a/wagtail/wagtailcore/query.py
+++ b/wagtail/wagtailcore/query.py
@@ -384,7 +384,10 @@ def specific_iterator(qs):
 
     # Yield all of the pages, in the order they occurred in the original query.
     for pk, content_type in pks_and_types:
-        yield pages_by_type[content_type][pk]
+        try:
+            yield pages_by_type[content_type][pk]
+        except KeyError:
+            continue
 
 
 # Django 1.9 changed how extending QuerySets with different iterators behaved


wagtail/wagtailcore/query.py
```diff
@@ -384,7 +384,9 @@ def specific_iterator(qs):
 
     # Yield all of the pages, in the order they occurred in the original query.
     for pk, content_type in pks_and_types:
-        yield pages_by_type[content_type][pk]
+        s = pages_by_type[content_type].get(pk)
+        if s is not None:
+            yield s
```
BertrandBordage: In terms of performance, a `try…except` would be faster (though taking one more line).

WnP: @BertrandBordage that's true, `try…except` would be faster, at least in case there's no `KeyError` otherwise… slower, but most of the time there will be no `KeyError` so it looks a good idea.

gasman:
Thanks @WnP! Could you add a test for this, please?

gasman:
Fallback behaviour for pages without a specific record has been in place since https://github.com/wagtail/wagtail/pull/5928 (although the behaviour in question is to return the generic Page instance rather than to skip the page entirely - I think this is more in line with the expected behaviour).

PR #4104, opened 2017-12-05
HTML col element has to be inside a colgroup

atombrella:
I believe the `colgroup` tag needs to the parent of `col`-tags. The `width`-attribute of the col-tags is deprecated in HTML5. Should a ticket be created for it?

https://html.spec.whatwg.org/multipage/tables.html#the-colgroup-element
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/colgroup

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/home/pages_for_moderation.html
+++ b/wagtail/admin/templates/wagtailadmin/home/pages_for_moderation.html
@@ -4,10 +4,12 @@
         <section>
             <h2>{% trans 'Pages awaiting moderation' %}</h2>
             <table class="listing">
-                <col />
-                <col width="30%"/>
-                <col width="15%"/>
-                <col width="15%"/>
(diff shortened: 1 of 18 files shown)


gasman:
Thanks @atombrella!
According to the spec:

"A colgroup element's start tag can be omitted if the first thing inside the colgroup element is a col element, and if the element is not immediately preceded by another colgroup element whose end tag has been omitted. (It can't be omitted if the element is empty.)"

I think this is spec-writer-speak for "if I encounter a `<col>` element, and I haven't seen a corresponding `<colgroup>` yet, I will behave as if there's an invisible `<colgroup>` around it". So, leaving the `<colgroup>` out _is_ valid.

I don't think we need to create an issue for the `width` attribute. It'd be nice to have fully compliant HTML in the admin, but it's not breaking anything (as far as I know!).

PR #4108, opened 2017-12-07
Update register_page_listing_more_buttons example

### LINKED ISSUE #4103: docs: register_page_listing_more_buttons gives wrong button class as example
I found a small issue with the docs, the documentation for register_page_listing_more_buttons gives you an example which appears to be copy-pasted from register_page_listing_buttons above it, because it tells you to use PageListingButton but that really looks out of place as a dropdown item in the More menu.

I had to read the Wagtail source to see how the More menu was constructed and realised the wrong button class was used in the docs, it should just be Button instead of PageListingButton.

I think the example in the docs should be fixed, as it confused me, it's going to confuse others.

lb-:
Use Button instead of PageListingButton in **register_page_listing_more_buttons** section.

Closes #4103

### DIFF
--- a/docs/reference/hooks.rst
+++ b/docs/reference/hooks.rst
@@ -719,13 +719,13 @@ Page explorer
 
     @hooks.register('register_page_listing_more_buttons')
     def page_listing_more_buttons(page, page_perms, is_parent=False):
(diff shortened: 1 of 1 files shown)


kaedroho:
LGTM, thanks!

PR #4116, opened 2017-12-10
Exception str to guide about missing Page instance inheritors

benjaoming:
Thanks for making Wagtail! :tada: 

I found that there's still some missing guidance after #3567. It's a modified version of #3554. To reproduce this error and make the main `/wagtail` (or `/admin`) view crash:

image

1. Create a `MyPage` instance (some model inheriting from `Page` and then an instance of that model)
1. Remove the instance of `MyPage`, but not the parent `Page` object. On method is to migrate backwards and forwards again.

Exception after this PR:

image

### DIFF
--- a/wagtail/wagtailcore/models.py
+++ b/wagtail/wagtailcore/models.py
@@ -721,7 +721,10 @@ class Page(six.with_metaclass(PageBase, AbstractPage, index.Indexed, Clusterable
             if commit:
                 self.save()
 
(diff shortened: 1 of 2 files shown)


lb-:
This PR is quite old, sorry this never got a decision or review @benjaoming 

For now I think it's best we close this, there has been a lot of work done since v1.13, including better handling of deleted page models (mostly in the History/reporting side of things though).

If this is still a problem on Wagtail 5.* - hopefully yourself or someone can raise an issue.

PR #4117, opened 2017-12-11
Add ability to set custom value_class on StructBlock

### LINKED ISSUE #2204: Ability to subclass `StructValue` to attach custom properties to StructBlocks
As proposed in https://github.com/torchbox/wagtail/pull/2187#issuecomment-179790713 - add a new `Meta` setting to StructBlock to specify a 'value' class to use in place of `StructValue`. This would mean that StructBlock values are not limited to dictionary-like behaviour, but can have arbitrary calculated properties too.

(The same trick would presumably work on StreamBlock/StreamValue too, although the usefulness of that is less clear...)

Probably a very minimal code change, but would warrant a decent-sized example in the documentation.

lb-:
Let me know if this needs any revisions.

Closes #2204

* Added tests for value_class setting
* Linting done
* Documentation added

### DIFF
--- a/docs/topics/streamfield.rst
+++ b/docs/topics/streamfield.rst
@@ -377,6 +377,9 @@ This defines ``PersonBlock()`` as a block type that can be re-used as many times
 
 Further options are available for customising the display of a ``StructBlock`` within the page editor - see :ref:`custom_editing_interfaces_for_structblock`.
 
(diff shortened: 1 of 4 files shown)


gasman:
Thanks @lb-! Merged in 4ea836520365aa84219e8fe7257a6b66b771a196.

lb-:
@gasman thanks 👍

PR #4118, opened 2017-12-11
Add filter example images

### LINKED ISSUE #1685: Add some example images to image filters reference
Here's the image filters reference as it is today: http://docs.wagtail.io/en/v1.0/topics/images/using_in_templates.html

I think it would really help people understand how these filters work by adding some example images that describe how cropping/resizing is performed with each filter.

This will be especially useful for describing the "fill" operation and how it works with focal points, "closeness", etc

For example: https://github.com/thumbor/thumbor/wiki/Crop-and-resize-algorithms

lb-:
Adds example images for the main `filter` types in docs.

Closes #1685


kaedroho:
Fantastic work, thanks!

PR #4119, opened 2017-12-11
A simple Dockerfile for project template

tomdyson:
This Dockerfile makes it easy for Wagtail developers to build and deploy containers for their sites:

```
wagtail start mysite
cd mysite
docker build -t wagtail .
docker run -p 8000:8000 wagtail
```

> For new features: Has the documentation been updated accordingly?

Not yet - I'll add documentation if other core developers agree with this approach.

### DIFF
--- a/wagtail/project_template/Dockerfile
+++ b/wagtail/project_template/Dockerfile
@@ -13,4 +13,4 @@ WORKDIR /code/
 
 EXPOSE 8000
 CMD python manage.py migrate
-CMD exec gunicorn wagtail_docker_test.wsgi:application --bind 0.0.0.0:8000 --workers 3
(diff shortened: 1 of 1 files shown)


tomdyson:
With this simple Dockerfile - and [now](https://zeit.co/now) installed - you can deploy Wagtail as simply as
```
wagtail start foo
cd foo
now
```

kaedroho:
We should add ``USER`` line so ``gunicorn`` is not run as the root user. See: https://docs.docker.com/engine/security/security/#conclusions.

Other than that LGTM!

gasman:
Bump... happy to merge this for 2.0 once @kaedroho's point re `USER` is addressed.

gasman:
Merged in 49bc130b16a986b7c58601d06fdc3e4bb480e045

PR #4132, opened 2017-12-14
Elasticsearch 6 support

kaedroho:
This PR adds Elasticsearch 6 support into Wagtail. The main changes are

- The ``_all`` field was removed. I've replaced this with a new custom ``_all_text`` field (``_all`` field name is reserved).
- Elasticsearch no longer supports multiple mappings per index. I've renamed the mapping to "doc". When we add more types to an index, it now just appends the new fields to this mapping (but all fields are prefixed with model name anyway so there shouldn't be any clashes)

I've also implemented a couple of cleanups I've been wanting to do, but couldn't as they're not backwards compatible:
 - Renamed the ``_partials`` field to ``_edgengrams``
 - Removed the root type prefix from all the IDs (this hasn't been required since we started separating indices by root type


SalahAdDin:
:+1:

wagtail/search/backends/elasticsearch2.py
```diff
@@ -374,10 +380,9 @@ def _connect_filters(self, filters, connector, negated):
 
             return filter_out
 
-    def _compile_term_query(self, query_type, value, field, boost=1.0, **extra):
-        term_query = {
-            'value': value,
-        }
+    def get_inner_query(self):
+        if self.query_string is not None:
+            fields = self.remapped_fields or [self.mapping.all_field_name, self.mapping.edgengrams_field_name]
```
gasman: This replacement doesn't make any sense... accidental copy-and-paste?

wagtail/search/tests/test_elasticsearch6_backend.py
```diff
@@ -0,0 +1,801 @@
...
+from elasticsearch.serializer import JSONSerializer
+
+from wagtail.tests.search import models
+from wagtail.search.backends.elasticsearch6 import Elasticsearch6SearchBackend
+
+from .elasticsearch_common_tests import ElasticsearchCommonSearchBackendTests
+from .test_backends import BackendTests
+
+
+class TestElasticsearch6SearchBackend(BackendTests, ElasticsearchCommonSearchBackendTests, TestCase):
```
gasman: Currently failing with `TypeError: Cannot create a consistent method resolution
order (MRO) for bases TestCase, ElasticsearchCommonSearchBackendTests, BackendTests` - I think `BackendTests` needs to be removed (since `ElasticsearchCommonSearchBackendTests` already inherits from it).

kaedroho:
Do not merge, there are failures on travis

kaedroho:
Should be fixed now

PR #4133, opened 2017-12-15
Fixes page unlock button styling.

BertrandBordage:
This pull request fixes the clickable area of the page unlock button. Before this pull request:
- we could click only on the text of the button, not on all the visible area
- the text got an ugly background on hover

## Before
selection_007
selection_008
## After
selection_009
selection_010


jjanssen:
Hi @BertrandBordage, I noticed the rendering of the button is a bit off in Firefox ESR and the latest version. Would you mind taking a look at that? Otherwise this change is good to go.

image

BertrandBordage:
Thanks for testing @jjanssen!
I fixed the Firefox rendering. I also fixed the button when the user has no locking permission, I broke it in the initial commit, it was not correctly disabled.

PR #4134, opened 2017-12-15
Changes default show_in_menus to True.

BertrandBordage:
After years of hesitation (and clients calls), Wagtail 2.0 is the occasion to finally set `Page.show_in_menus` to `True` by default.
For those who liked the previous default value, it’s still possible to use `Page.show_in_menus_default` to keep the old behaviour.


Proper-Job:
@BertrandBordage I disagree with this for purely personal reasons.  In my experience, pages that show up in menus are far outnumbered by pages that don't.

BertrandBordage:
Thank you for the feedback, @Proper-Job :)
Please everyone continue to upvote/downvote this issue, we’ll decide of merging/abandoning this pull request depending on the results of this vote.

kaedroho:
As mentioned in this weeks meeting, why can't we add a new global setting to allow the default to be changed per project?

In general, I'm against adding more breaking changes to Wagtail 2.0 that aren't strictly necessary. We do still want upgrade to Wagtail 2.0 to be as straightforward as possible. Reorganising the modules had to be done, wasn't possible to do in a backwards-compatible way and there was no alternative.

BertrandBordage:
@kaedroho: I wanted to give a try at the simplest solution that doesn’t introduce more complexity, to get developers’ feedback.

lb-:
Original issue request for this to be optional is here #692 - along with discussion about why the current default is not ideal.

I personally am for the default being `show_in_menus = True` but that is only from the few sites I have worked with and coming from other CMSs where it is non-standard to 'hide' the page from the menu structure.

I can see how this may not be the case for all setups though.

BertrandBordage:
Postponing the decision due to negative feedback. Please keep voting on the initial issue message and add your opinion/experience/clients feedback.

lb-:
After discussion with the core team and reviewing the feedback, we will not be merging this in.

Instead we will leave the default as is but add a bit more documentation regarding how to set this globally for your Page model -> #4230

PR #4140, opened 2017-12-16
[DISCUSS] no page defined - potential fix

rosscdh:


gasman:
As per #4141, I don't believe this is a bug - if the pageurl tag receives a page value equal to None, that's a coding error on the template author's part, and so the error message is legitimate.

rosscdh:
Please review the issue i raised connected to this PR.

PR #4142, opened 2017-12-17
Improve efficiency of Site.find_for_request() by utilising Django's cache API

ababic:
Supersedes #4014 and #3358

With all of the recent talk about potential changes to Site, and addressing some of the longstanding bugbears with the current Site implementation (see #3834, #2840), I decided to work on a cached SiteManager that would yield the same results in all of the current test scenarios, while also attempting to improve the efficiency of how sites are identified for individual requests.  

Here's how it works:

- Calling 'find_for_request()' triggers population of the cache if it hasn't yet been populated
- 'populate_cache()' fetches all existing `Site` objects in a single query, analyses them, and (pre)populates the cache in a way that 'get_for_request()' should be able to effectively match requests to a relevant `Site` object without further database queries. The cache persists until repopulation is triggered by a signal handler (or the cache is cleared for some other reason).
- Signal handlers trigger repopulation of the cache when a `Site` is created, updated or deleted, or when a `Page` set as the root page for one of more `Site`s is updated.
- `get_for_request` is deliberately ignorant about the possibility of other processes having already started to update the cache, and simply attempts to populate it itself if the cache is empty when it checks. This might result in work being repeated by more than one process, but that overhead is worth it for the sake of keeping the code readable and easier to maintain. 
- Caching behaviour is 'opt-in'. If the `WAGTAIL_SITE_CACHE_ENABLED` setting is used to turn it on, some checking occurs to ensure that a compatible cache backend is in use.

Still to do:

- [ ] Establish what other tests should be added and add them
- [x] Update documentation


ababic:
@BertrandBordage when you're ready, this is good to review now.

ababic:
I've just realised a fatal flaw in this design: cache.set_many() can't always be relied upon to set all values passed to it, and if that were to happen, requests might simply stop being matched to the correct sites, and it would be difficult to spot what the problem is.

I wanted to avoid storing all matches as a single dictionary and fetching it again for every request lookup, because I wanted reads to remain as minimal/fast as possible. It seems like that might be the only reliable way to do it if we want to take this 'prepopulation' approach though. 

I'll try to spend some time on it this weekend.

ababic:
Okay, I'm feeling much better about this approach. Much simpler! I think I'm happy enough with this implementation for now. Does anyone has any feedback?

ababic:
Did some quick/dirty benchmarking of `Site.find_for_request()` to get an idea of performance with caching enabled/disabled, always using sites with unique hostnames (which creates the largest possible dictionary for the cache), and populating the cache beforehand. 

For ease, I just used the `FileBasedCache` backend, but the results were good.

No of sites | Time with caching on (s) | Time with caching off (s) | Improvement
----------- | ------------------------ | ------------------------- | -----------
500 | 0.000658 | 0.002778 | 4.2x
250 | 0.000321 | 0.002823 | 8.8x
100 | 0.000319 | 0.003077 | 9.7x
50 | 0.000404 | 0.002854 | 7.1x
25 | 0.000357 | 0.002836 | 7.9x
10 | 0.000316 | 0.002243 | 7.1x
5 | 0.000350 | 0.002556 | 7.3x

wagtail/core/models.py
```diff
@@ -36,9 +39,117 @@
...
+                "Memcached or Redis. If multiple CACHES are configured, you"
+                "can use the WAGTAIL_SITE_CACHE setting to specify a cache "
+                "other than '{0}' to use instead".format(key))
+            return
+        if not settings.DEBUG and isinstance(cache, LocMemCache):
+            warnings.warn(
+                "WAGTAIL_SITE_CACHE is set to '{0}', which refers to a cache "
+                "that utilises Django's LocMemCache backend. Wagtail's site "
+                "caching feature is incompatible with LocMemCache when running "
+                "in a production environment, and so has been disabled. To "
```
chosak: Would it be better to more explicitly refer to the use of `settings.DEBUG` here, instead of "in a production environment"? I can think of use cases where someone might be testing with `DEBUG=False` when not in production, and might not be sure what this is referring to.

wagtail/core/models.py
```diff
@@ -36,9 +39,117 @@
...
+        """Return the site responsible for dealing with the supplied
+        ``HttpRequest`` instance"""
+        hostname, port = self.get_hostname_and_port_from_request(request)
+        cache = self.get_site_cache()
+        if cache is None:
+            return get_site_for_hostname(hostname, port)
+
+        sites = cache.get(SITE_CACHE_SITES_KEY) or self.populate_cache()
+        request_key = "{0}:{1}".format(hostname, port)
+        return sites.get(request_key) or sites.get(SITE_CACHE_DEFAULT_KEY)
```
chosak: Does falling back on the default site like this subtly change the existing behavior in the case where there are multiple sites but the incoming request doesn't match any of them? [In `get_site_for_hostname` here](https://github.com/wagtail/wagtail/blob/master/wagtail/core/sites.py#L49), the logic seems to raise `Site.DoesNotExist` if there are multiple sites but the request doesn't match any. With this change, the logic would return the default site instead. Am I misreading?

wagtail/core/tests/test_sites.py
```diff
@@ -56,39 +58,207 @@ def test_no_site_name_default_port_n(self):
...
+        request = HttpRequest()
+        request.path = '/'
+        request.META['HTTP_HOST'] = self.default_site.hostname
+        request.META['SERVER_PORT'] = self.unrecognised_port
+        with self.assertNumQueries(self.find_for_request_queries_expected):
+            self.assertEqual(Site.find_for_request(request), self.default_site)
+
+    def test_unrecognised_port_and_unrecognised_host_routes_to_default_site(self):
+        # requests with an unrecognised Host: header _and_ an unrecognised port
+        # hould be directed to the default site
```
chosak: See comment above; in the existing logic, isn't this only true if there's only one site, otherwise it should raise `Site.DoesNotExist`?

wagtail/core/tests/test_sites.py
```diff
@@ -56,39 +58,207 @@ def test_no_site_name_default_port_n(self):
...
+        with self.assertNumQueries(self.find_for_request_queries_expected):
+            self.assertEqual(Site.find_for_request(request), self.alternate_port_events_site)
+
+
+@override_settings(
+    WAGTAIL_SITE_CACHE_ENABLED=True,
+    CACHES={
+        'default': {
+            'BACKEND': 'django.core.cache.backends.filebased.FileBasedCache',
+            'LOCATION': '/var/tmp/django_cache',
```
chosak: This may be an overabundance of caution, but does defining the cache location like this require write access in order to run the tests? I notice that [the Django tests of `FileBasedCache`](https://github.com/django/django/blob/acc989f037f4185b3f55cb76ab51c5224f30f0d1/tests/cache/tests.py#L1337) use a temporary directory instead of something hardcoded.

wagtail/core/models.py
```diff
@@ -36,9 +39,117 @@
...
+        to match ``HttpRequest`` objects to the relevant ``Site``"""
+        cache = self.get_site_cache()
+        if cache is None:
+            return
+
+        sites = {}
+        unique_hostname_sites = {}
+        for site in self.all().select_related('root_page').iterator():
+
+            sites["{0}:{1}".format(site.hostname, site.port)] = site
```
chosak: There are three places in this PR where the cache key is generated using some form of `"{0}:{1}".format(hostname, port)`. Does it make sense to encapsulate this in a method somewhere for easier lookup of the logic used to generate the key?

wagtail/core/models.py
ababic: @chosak I was hoping for feedback on this wording, and that sounds like a sensible change. I was also thinking about rewording so that  "Wagtail's site caching has been disabled" comes first?

wagtail/core/models.py
ababic: I did have something similar to this before the latest refactor, but ditched it in favour of keeping things more readable. However, you are definitely right - ensuring these keys are handled consistently is important.

wagtail/core/models.py
chosak: Personally I find it more readable this way: explaining an incompatibility and then saying, in conclusion, "and so has been disabled". But others might have different or stronger opinions.

Relatedly, can you provide a little more detail behind the rationale of preventing this cache in production? The [Django docs say](https://docs.djangoproject.com/en/2.0/topics/cache/#local-memory-caching) that the `LocMemCache` is thread-safe, although, on the other hand, is only per-process, so wouldn't be shared among e.g. multiple server threads on the same instance. But if that's the concern, isn't that also a problem when running across multiple instances -- for example, modifying a Site through the admin that hits one instance wouldn't know how to clear caches on the other instances. To be totally safe, doesn't this caching only work if there's a single site cache for all Django processes in a single Wagtail deployment?

ababic: Yes, 'thread safety' is probably the wrong term here. What I am really trying to guard against is use of `LocMemCache` when the site is running as multiple processes, because each process would have it's own version of the cache in memory, and signals would only update caches in the processes where they were triggered. I'm open to recommendations on how to word this?

Any running processes will indeed need to share the same cache instance in order for the cached data to remain consistent between processes. This will no doubt have to be outlined in the docs.

wagtail/core/models.py
ababic: @chosak Thanks. Yes, I think you could be right here. I will review to make sure this is fully backwards compatible.

wagtail/core/tests/test_sites.py
ababic: @chosak The tests seemed to run fine on Travis CI, but I'm honestly unsure on what the best option here would be. Happy to change to whatever you feel would is suitable.

wagtail/core/models.py
chosak: Right, so doesn't that just as equally apply to the file-based cache, for Wagtail deployments that run across multiple servers? Having a `LocMemCache`-specific callout/check might be a bit misleading if there are other caches that would also need to be excluded -- not to mention any custom caches that someone might be using.

wagtail/core/tests/test_sites.py
chosak: How would you feel about copying that Django logic, to use a temp directory generated at test runtime? I suspect this test would fail on a Windows machine or somewhere without write privileges to `/var/tmp/django_cache`.

ababic: @chosak If FileBasedCache and LocMemCache are equally unsuitable in terms of running in multiple processes, then I suppose there's no reason why we shouldn't just use LocMemCache as the backend to show the cache working, and avoid complicating the tests any more than needed. Does that sound okay to you?

wagtail/core/models.py
ababic: Oh dang, you're right. I completely overlooked that! I suppose it's a bit of a grey area as to how far we go in terms of accounting for things like this. One option would be to not to give any special treatment to specific backends at all and just explain as well as possible in the docs. Personally, I'd prefer to treat FileBasedCache in the same way as LocMemCache and leave it at that - that's at least Django's official backends covered. I'm open to other ideas though?

wagtail/core/tests/test_sites.py
ababic: This test was copied pretty much 'as is' from [`test_page_models.py`](https://github.com/wagtail/wagtail/pull/4142/files/7e1bc7c06bd52210616d4010f5024c46236bb0fd#diff-04b4575a8f6981406bfdca2128f7a88bL177), so I'm sure the behaviour described by the test is valid. However, I will add a new test to confirm that Site.DoesNotExist is raised when no site is marked as the default (I'm pretty sure that's the only time you'd run into that exception).

wagtail/core/tests/test_sites.py
chosak: Yes, that sounds like a great idea!

wagtail/core/models.py
chosak: My preference would probably be to remove special treatment of any backends, and leave it to the documentation to warn users about sensible cache configuration. Pinging @gasman @lb- for their opinions.

Speaking of: is `DummyCache` really "incompatible" or does it just work as a no-op in this case that ends up being less efficient than the default behavior? If so, I wonder if that check really adds any value, as it'd work the same way as any other time that cache backend is used.

ababic: `DummyCache` was entirely incompatible (which was one of my reasons for wanting to refactor, although I don't think I mentioned it) - But now, it is just as you say - it will work fine, but it will be a little less efficient.

I'm almost certain that @gasman and @lb's thoughts will reflect yours, as realistically, there are more arguments for ditching the special case treatment than there is for keeping it - and I guess I'm fine with that really, so long as the documentation is right.

I'll work on this later :)

ababic:
I've now added some documentation for this feature and it's related settings. I think it makes sense to keep it all together like the various 'password management' settings, but i'm open to other ideas. As always, any thoughts/recommendations for improvement are welcome.

docs/advanced_topics/settings.rst
```diff
@@ -439,6 +439,38 @@ can only choose between front office languages:
...
+
+While many projects should benefit from enabling site caching, there are a few exceptions:
+
+* If your project runs as multiple processes (for example, as multiple containerised instances, as part of a cluster or swarm), site caching should only be enabled if all processes share the same cache instance (e.g. a Redis or Memcached instance, running as a separate service).
+* If your project uses a file-based or local-memory based cache (such as the ``FileBasedCache`` ``LocMemCache`` cache backends included with Django), site caching should only be enabled if the project runs as a single process (and this is unlikely to change in future).
+* If your project recieves only a light amount of traffic (between 1 and 5000 visits per day), and doesn't already have a cache configured, complicating your setup with a cache for the sole purpose of enabling site caching is likely not worth the small performance gain.
```
tomdyson: recieves -> receives
between 1 and 5000 -> up to 5000

docs/advanced_topics/settings.rst
```diff
@@ -439,6 +439,38 @@ can only choose between front office languages:
...
+
+.. code-block:: python
+  
+  WAGTAIL_SITE_CACHE = 'default'
+
+For projects with multiple caches, this setting allows you to speficy the alias of the cache (from your ``CACHES`` setting) that should be used for site caching. The default is ``'default'``.
```
tomdyson: speficy -> specify

ababic:
Thanks @tomdyson. Those should be corrected now. 

I also had a bit of eureka moment after writing the documentation, and have made quite a few simplifications as a result. I guess documentation is good for helping you see things in a different light! 

`get_for_request()` is now a little more complicated as a result, but on the whole the changes are much easier to read and far less 'custom'.

Issue #3375, opened 2017-02-16
Warn people when they're looking at the development version of the docs


gasman:
Occasionally people find their way to the development version of the tutorial page, and get tripped up when things don't match the current release version:
https://groups.google.com/d/msg/wagtail/HPL5w1LIrAw/SvVLJKgGCgAJ

[Django's docs](https://docs.djangoproject.com/en/dev/) display a notice:

> This document is for Django's development version, which can be significantly different from previous releases. For older releases, use the version selector floating in the bottom right corner of this page.

We should see if readthedocs allows us to do the same, and add one if so.

zerolab:
RTD show a notice when viewing an old tag - https://github.com/rtfd/readthedocs.org/blob/499a74076f0debae09acc1d4c4eeeab03f8f557e/media/javascript/rtd.js#L14

I think the same can possibly be done with https://github.com/rtfd/readthedocs.org/issues/152#issuecomment-186115025

lb-:
After a bit of digging, it looks like a fix to Readthedocs may already be in the works.

Others have run into this issue here: https://github.com/rtfd/readthedocs.org/issues/3481

Also, there is a reasonably recent PR to fix it here: https://github.com/rtfd/readthedocs.org/pull/3547

I will keep an eye on that issue/PR but if others want to +thumbsup on the issue - it might help the RTD team flag that this is important.

lb-:
I'm not sure if this is new, maybe a Read the Docs option/update?

I noticed today that a small popover appears for a few seconds on the `latest` docs. 

![Image](https://github.com/user-attachments/assets/2800dfda-bce3-4001-8871-033ffa306c85)

laymonage:
@lb- yep, it's the new readthedocs addons https://about.readthedocs.com/blog/2024/07/addons-by-default/

We enabled the feature when it was still in beta a few months back, but there was an issue with our theme https://github.com/wagtail/sphinx-wagtail-theme/issues/296 so it didn't show up immediately. It's fixed now though, so I think we can close this – thanks!

lb-:
Thanks @laymonage - this is an awesome win.

PR #4148, opened 2017-12-20
fixed imports in tutorial

engineervix:
As I was going through the tutorial and running the code on my machine, I was getting `ImportError`s. These were caused by imports such as `from wagtail.core.fields import RichTextField`. I have updated the docs by adding `wagtail` prefixes where applicable. For instance:
 `from wagtail.wagtailcore.fields import RichTextField` instead of `from wagtail.core.fields import RichTextField`

### DIFF
--- a/docs/getting_started/tutorial.rst
+++ b/docs/getting_started/tutorial.rst
@@ -219,7 +219,7 @@ Now we need a model and template for our blog posts. In ``blog/models.py``:
     from wagtail.wagtailcore.models import Page
     from wagtail.wagtailcore.fields import RichTextField
     from wagtail.wagtailadmin.edit_handlers import FieldPanel
-    from wagtail.search import index
+    from wagtail.wagtailsearch import index
 
 
     # Keep the definition of BlogIndexPage, and add:
--- a/docs/getting_started/tutorial.rst
+++ b/docs/getting_started/tutorial.rst
@@ -69,9 +69,9 @@ Edit ``home/models.py`` as follows, to add a ``body`` field to the model:
 
     from django.db import models
 
-    from wagtail.core.models import Page
-    from wagtail.core.fields import RichTextField
-    from wagtail.admin.edit_handlers import FieldPanel
+    from wagtail.wagtailcore.models import Page
+    from wagtail.wagtailcore.fields import RichTextField
+    from wagtail.wagtailadmin.edit_handlers import FieldPanel
 
 
     class HomePage(Page):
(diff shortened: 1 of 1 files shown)


thibaudcolas:
Hi @engineervix 👋, thanks for looking into this.

I think you may be running into this because there is a version mismatch between the Wagtail version you're using and that of the tutorial. Those paths were renamed very recently (Wagtail 2.0, not released yet, see https://github.com/wagtail/wagtail/blob/master/docs/releases/2.0.rst#wagtail-module-path-updates).

If you look at the tutorial for the [latest stable release](http://docs.wagtail.io/en/v1.13.1/getting_started/tutorial.html), 1.13, the paths seem to be correct.

Could you let me know whether that solves it?

engineervix:
@thibaudcolas thanks for the quick response. I'm using the latest stable release (1.13) and from your [link](http://docs.wagtail.io/en/v1.13.1/getting_started/tutorial.html) the docs seem to be fine! Ah :see_no_evil: :see_no_evil:, now I get it -- I think I had initially accessed the [docs based on the latest, unreleased version (2.0)](http://docs.wagtail.io/en/latest/getting_started/tutorial.html), and I thought those were the docs for version 1.13! My bad, I apologize for my ignorance. I appreciate your quick response and guidance, thanks for pointing me in the right direction. :blush: :blush:

thibaudcolas:
Argh, this is because of #3375 😞 . Sorry this has happened to you!

engineervix:
No worries @thibaudcolas! I'm all good now! However, I think other wagtail newcomers may face a similar problem and perhaps get discouraged. After retracing my steps, I realized that the problem actually came from the [getting started section of the wagtail.io website](https://wagtail.io/developers/), where I was directed to the wrong version of the docs. I have proposed that the link on the website be changed (#4151). This may save future wagtail developers from going through what I went through :smiley:

simobasso:
@thibaudcolas you simple need to select the `stable` version in the readthedocs options.

I think here https://readthedocs.org/dashboard/wagtail/versions/

schermata 2017-12-20 alle 20 40 11

gasman:
Thanks for the report @engineervix! Have updated the link on https://wagtail.io/developers/ now.

(@simobasso The readthedocs config isn't the problem - http://docs.wagtail.io/ is already configured to redirect to the latest release version.)

simobasso:
> (@simobasso The readthedocs config isn't the problem - http://docs.wagtail.io/ is already configured to redirect to the latest release version.)

@gasman Yeah! But I think is better `stable` instead of `latest` in this case. :)

engineervix:
Thank you so much :relaxed: :relaxed: :smiley:

Issue #3806, opened 2017-08-29
npm install-ing Wagtail creates undesired changes in package-lock.json


thibaudcolas:
Initially reported by @gasman in https://github.com/wagtail/wagtail/pull/3773#issuecomment-325625388.

### Issue Summary

On a fresh `npm install`, npm updates the `package-lock.json` file depending on the context the command is ran in. Those changes to the lockfile reflect which packages were installed, which can differ because of:

1. A difference in package versions available, if the package.json uses ranges (https://github.com/npm/npm/issues/16866, https://github.com/npm/npm/issues/17979, https://github.com/npm/npm/issues/18103).
2. A difference in the install environment (eg. installing on a different OS than the one the lockfile was created in). This is https://github.com/npm/npm/issues/17722, https://github.com/npm/npm/issues/18135.

In practice, this is problematic because it creates unnecessary file changes to deal with.

### Steps to Reproduce

1. Run `npm install` on an Linux machine, with the current OSX-generated package lock.

### Technical details

* npm >= 5.1.0

----------

The first problem can be resolved once and for all by dropping all range specifiers from the `package.json` (`^`, `~`), I don't know of any "once and for all" solution for the environment difference.

The environment difference shouldn't be more than an annoyance, however the dependency version problem can cause real issues if a new version of one of the dependencies breaks the code/tooling (using a lockfile is supposed to prevent this), so we may want to pin all of the dependencies to exact versions until it is resolved.

Until we find a more proper fix, it is safe to either:

```sh
# Discard the changes to the file:
git checkout -- package-lock.json

# Use the following flag to prevent the changes to the file:
npm install --no-shrinkwrap
```

I've been tracking all of the npm issues highlighted above, hoping for a fix in npm, if that doesn't come I guess we can go for a plan B solution like "switch to yarn", or "move back to npm-shrinkwrap.json and gitignore the package-lock.json".

zerolab:
Related: #3442

thibaudcolas:
From https://github.com/npm/npm/issues/17979#issuecomment-332701215, point 1. above seems to be addressed in a recent npm upgrade. Trying this briefly with npm 5.4.2, this seems to be addressed.

Point 2. is still pending.

Point 3. – I get a few more diffs, not sure whether they are related to the recent React update, or just to the fact that I use a newer version of npm:

- HTTPS -> HTTP for some packages in the `resolved` field.
- `dev: true` added or removed in a few places
- `string-decoder` changing locations.

[...]

PR #4149, opened 2017-12-20
Update frontend tooling packages

jjanssen:
Since I'm in the flow of upgrading: I have updated most of the packages in the `package.json` including the upgrade to React 16. Which was pain-free due to the fantastic Enzyme adapters 🎉 

Some highlights:
- I had to update some tests due to the `shallow`-call being made before mutating the `document.body`.
- `react-addons-perf` is not compatible with React 16 so I have removed this from the code. An alternative for local development would the [React Perf extension](https://github.com/crysislinux/chrome-react-perf) for chrome. Although I have not tested it just yet.

What still needs a little tender, love and care after the merge:
- React Transition group;
- ESLint;

Things which need to done:
- [x] Verify whether the output of the gulp updates did not change drastically;
- [x] Maybe give it a bit of QA goodness (CC @hvandenberg)
- [x] Circle CI configuration check-up :expressionless:


thibaudcolas:
@jjanssen could you make sure to generate the `package-lock.json` with npm@5.6.0 so this addresses the remaining lockfile difference in #3806?

jjanssen:
The `gulp-autoprefixer` update reintroduced some vendor-prefixes where they were missing previously. Like for eg: `animation`, `transform` and `transition`. In the broader perspective nothing to worry about.

On the left you see a snippet of my update branch and on the right the current master.
image

Right now we define `['last 3 versions', 'ie 11']`. I think the changes showing up is a part of the autoprefixer change on can-i-use library and maybe a stricter version check. We could either narrow the scope down according to our [official browser support list](http://docs.wagtail.io/en/v1.13.1/contributing/developing.html) or leave it be and take the added rules for granted.

Personally I think we should narrow it down, to prevent unnecessary code from being rendered.

jjanssen:
As for `react-transition-group` there are some major changes between the v1 and v2 API. I still have to grasp the new concept of it and update it properly. But I haven't figure it out yet completely. Still hoping to round-up on this as a part of this PR.

HvandenBerg:
Tested this to see if the explorer still works cross browser according to the DoD, test OK.

thibaudcolas:
Merged in 9f65e7001f17e922db32628b43028414351fe05f 🎉 .

PR #4157, opened 2017-12-27
Make the ChooserBlock QuerySet customizable

leewesleyv:

### DIFF
--- a/wagtail/core/blocks/field_block.py
+++ b/wagtail/core/blocks/field_block.py
@@ -532,11 +532,14 @@ class ChooserBlock(FieldBlock):
         self._help_text = help_text
         super().__init__(**kwargs)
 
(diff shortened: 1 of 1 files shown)


gasman:
Thanks @leewesleyv!

Please can you give some more explanation of the kind of customisation this change is intended to achieve? The existing subclasses of `ChooserBlock` (`PageChooserBlock`, `ImageChooserBlock` etc) use a modal popup which performs its own queries, and so an overridden `get_queryset` method wouldn't get picked up.

gasman:
Closing due to lack of info. Joining the dots a bit, I take it this is intended for use on custom blocks that subclass ChooserBlock and don't override the default select widget - however, this isn't a supported configuration, and doesn't work correctly (see #3169).

PR #4159, opened 2017-12-30
Fixing search url for django 1.10 and above.

misraX:
For django 1.10 and above the search url will fail with a TypeError since there is no support for views as string inside urlpatterns, It will raise:

```
TypeError('view must be a callable or a list/tuple in the case of include().')
TypeError: view must be a callable or a list/tuple in the case of include().
```


gasman:
Looks good to me - thanks @misraX!

PR #4164, opened 2018-01-03
Add missing appconfs for settings, table_block and sites

gasman:
A few of our apps are currently missing appconfs (or don't have it hooked up as `default_app_config`). This is especially important for `wagtail.sites` because the missing appconf causes it to be given the label `sites`, which clashes with `django.contrib.sites`.

### DIFF
--- a/wagtail/api/__init__.py
+++ b/wagtail/api/__init__.py
@@ -1,4 +1,1 @@
 from .conf import APIField  # noqa
-
-
-default_app_config = 'wagtail.api.apps.WagtailAPIAppConfig'
--- a/wagtail/api/v2/__init__.py
+++ b/wagtail/api/v2/__init__.py
@@ -1,0 +1,1 @@
+default_app_config = 'wagtail.api.v2.apps.WagtailAPIV2AppConfig'
--- a/wagtail/contrib/modeladmin/__init__.py
+++ b/wagtail/contrib/modeladmin/__init__.py
@@ -1,0 +1,1 @@
+default_app_config = 'wagtail.contrib.modeladmin.apps.WagtailModelAdminAppConfig'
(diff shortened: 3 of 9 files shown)


loicteixeira:
On a side note, I understand this is needed for backward compatibility and that Django doesn't fully deprecated `default_app_config` (the documentation mentions it but there is no warning in the code). Should we update our documentation to encourage the new style?

For reference, this is what the documentation says:

> `default_app_config` allows applications that predate Django 1.7 such as `django.contrib.admin` to opt-in to `AppConfig` features without requiring users to update their `INSTALLED_APPS`.
>
> New applications should avoid `default_app_config`. Instead they should require the dotted path to the appropriate `AppConfig` subclass to be configured explicitly in `INSTALLED_APPS`.

gasman:
Thanks @loicteixeira! Updated now. The `default_app_config` for the API was in the wrong place completely (I suspect it was a bit of dead code left over from the v1 API), since the app is `wagtail.api.v2`, not `wagtail.api`... presumably that means `WAGTAILAPI_USE_FRONTENDCACHE` has never actually worked 😬  No idea how I missed modeladmin there.

I wouldn't be opposed to updating the docs to encourage appconfs in INSTALLED_APPS, but I don't think it's particularly urgent either. Come to think of it, I don't think I've ever seen any other third-party Django package recommend it in their docs - perhaps the practice just never caught on?

loicteixeira:
Yeah I don't know either. In a way Django doesn't officially deprecate it since there are no warnings so I don't see them enforcing it anytime soon. But even if they do, all that will be needed for users is to update the `INSTALLED_APPS` accordingly so we might as well just leave it like this for now (if we are the only one to recommend the new style, it might confuse a lot of people?).

PR #4171, opened 2018-01-04
Makes it possible to set a custom Gravatar source

bloodywing:
I made this for personal use, but think it's nice enough to share it with the rest of the wagtail users. Also makes it possible to no longer query gravatar.

### DIFF
--- a/docs/advanced_topics/settings.rst
+++ b/docs/advanced_topics/settings.rst
@@ -231,6 +231,13 @@ This setting lets you change the number of items shown at 'Your most recent edit
 Allows the default ``LoginForm`` to be extended with extra fields.
 
 
+.. code-block:: python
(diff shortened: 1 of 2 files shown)


bloodywing:
Travis had a Timeout on pypi.python.org here ...

gasman:
Thanks @bloodywing! Just so you're aware, there's an existing PR #4062 currently in review that makes major changes to avatar handling - we'll revisit this one once that one is merged.

bloodywing:
@gasman yes I know that pull request and I would remove the `gravatar.py` file. But my code can go into the `wagtail/wagtail/users/utils.py` after #4062 has been merged.

gasman:
Now incorporated into #4462

PR #4172, opened 2018-01-04
Implements `Prefix` on PostgreSQL.

BertrandBordage:
This pull request replaces the original work #3950.

It implements `Prefix` on `db` and `postgres_search` backends.

There is a small imperfection in a rare edge case. We may even forbid it, given its rarity and lack of meaning. When searching one or multiple combined `Prefix` only, only `SearchField`s where `partial_match` is `True` are searched in. When searching any other query than `Prefix`, it matches any field, either `partial_match` or not. **But when searching a combination of `Prefix` and non-`Prefix`** instances, then only the `SearchField`s with `partial_match` is `True` are searched in.


wagtail/search/tests/test_backends.py
```diff
@@ -507,6 +516,17 @@ def test_operators_combination(self):
                              'The Rust Programming Language',
                              'Two Scoops of Django 1.11'})
 
+    def test_prefix_single_word(self):
+        results = self.backend.search(Prefix('pro'), models.Book.objects.all())
+        self.assertSetEqual({r.title for r in results},
+                            {'The Rust Programming Language'})
+
+    def test_prefix_multiple_words(self):
+        results = self.backend.search(Prefix('rust pro'),
```
kaedroho: "prefix" is a single term query on Elasticsearch, so this example won't return anything there

kaedroho:
I thought we agreed to drop these term level queries as they weren't possible to implement on PostgreSQL? I suggested in the meeting on Wednesday that we instead add an argument (or a separate query class) which enables the last term of a ``PlainText`` to be a prefix as that covers the most common use case of term level queries and is still combinable using the or/and/not queries.

wagtail/search/tests/test_backends.py
BertrandBordage: According to [this article](https://qbox.io/blog/elasticsearch-queries-prefix-term-multimatch-bool), prefix queries in Elasticsearch can contain multiple terms.

BertrandBordage:
@kaedroho I preferred not to mess with the API for now, as I was not sure where we were going. Does this means that we definitely drop `Term` as well as `Prefix` on all backends and add an `Autocomplete` class? Whatever we decide, this will not change most of this pull request since “reading” the API query is just a few lines.

In my opinion `Prefix` and `Term` serve no purpose since they are not analyzed, making them essentially broken to the end user. But we could keep them as an Elasticsearch-specific API extension (imported from `wagtail.search.query.elasticsearch`) that would be almost never used.

So I don’t think it’s reasonable to implement your suggestion of making autocomplete queries split the last term and pass it as a `Prefix`, because prefix queries are not analyzed in Elasticsearch. This means that searching for `marie cur` will not match `Marie Curie`, but `marie Cur` will. Also, this implies splitting the last word in Python, which is a non-trivial task considering search engine rules.

On Elasticsearch, do you agree that such an `Autocomplete` query should search in the `edge_ngram` field that contains only `partial_match=True` data? While a `PlainText` query should search in a regular field containing both `partial_match=False` and `partial_match=True` data?

If we implement all that, then we should deprecate `partial_match=True` in favor of `autocomplete=True` or simply an `AutocompleteField` class.

In my opinion, this whole autocomplete vs plain text queries should be usable as simply as in django-haystack, and I have the feeling that it’s what Wagtail developers are waiting for.

Additional note: at least on PostgreSQL, combining `Autocomplete` and `PlainText` should raise a `NotSupportedError`.

Once all this is done, we can use autocomplete queries to search in `ChooserPanel`s as well as the admin search views that show results as we type.

kaedroho:
> In my opinion Prefix and Term serve no purpose since they are not analyzed, making them essentially broken to the end user. But we could keep them as an Elasticsearch-specific API extension (imported from wagtail.search.query.elasticsearch) that would be almost never used.

I agree. We should remove them completely.

> On Elasticsearch, do you agree that such an Autocomplete query should search in the edge_ngram field that contains only partial_match=True data? While a PlainText query should search in a regular field containing both partial_match=False and partial_match=True data?

I agree with the proposed behaviour for ``PlainText``, but if ``PlainText`` does also searches autocomplete fields, then what is the use for the separate ``Autocomplete`` query?

> If we implement all that, then we should deprecate partial_match=True in favor of autocomplete=True or simply an AutocompleteField class.

Agreed

BertrandBordage:
> I agree with the proposed behaviour for PlainText, but if PlainText does also searches autocomplete fields, then what is the use for the separate Autocomplete query?

`Autocomplete` will not do the same kind of query in the same database fields. Concretely, suppose we have this model:
```python
class Article(Page):
    body = RichTextField()
    search_fields = [AutocompleteField('title'), SearchField('body')]
```
If we create an `Article(title='a beautiful title', body='a deep and relevant article')`, in Elasticsearch, these should be indexed:
```
autocomplete: 'a beautiful title'
plain_text: 'a beautiful title a deep and relevant article'
```
`autocomplete` being an `edge_ngram` column and `plain_text` being a regular column.

When we search using `Autocomplete`, we only search in `autocomplete`, and when we search using `PlainText`, it only searches in `plain_text`. That way, `Autocomplete('bea')` will match the article and `PlainText('bea')` will not, but `PlainText('beauty')` or `PlainText('beautiful')` will.

With PostgreSQL, it is not implemented exactly the same as we use prefix matching instead of `edge_ngram`. That way, we can avoid the redundancy of indexing the same data in both `autocomplete` and `plain_text`. So we can use prefix matching on `autocomplete` for `Autocomplete`, and use normal search on both `autocomplete` & `plain_text` for `PlainText` (since autocomplete data is not in `plain_text`).

wagtail/search/tests/test_backends.py
kaedroho: Nice, makes things a lot easier!

wagtail/search/tests/test_backends.py
```diff
@@ -448,6 +450,13 @@ def test_term(self):
                             {'JavaScript: The Definitive Guide',
                              'JavaScript: The good parts'})
 
+    def test_incomplete_term(self):
```
kaedroho: I think we should have a similar test for ``PlainText`` as well

wagtail/search/tests/test_backends.py
BertrandBordage: I’m not sure what you mean by this. Anyway, the test is marked as an expected failure on Elasticsearch.

PR #4175, opened 2018-01-05
Make can_add_subpage check for can_create_at.

BertrandBordage:
In order to display the “Add subpage” in the admin & front-end userbar, Wagtail currently checks only using class methods and without using `Page.can_create_at` or `Page.can_exist_under`. This pull request fixes this.

### Use case
If you like overriding `Page.can_exist_under` to guide users – for example to prevent users from adding more than a contact page inside the home page – you often end up with cases where no sub-page can be added to a given page, even though `subpage_types` & `parent_page_types` allow it. The “Add subpage” button stays visible, and you end up on this misleading and frustrating view:
image

With this pull request, the available sub-page models now takes `Page.can_create_at` into account (and therefore `Page.can_exist_under`). It’s therefore impossible to reach an empty type selection page, a 403 error is raised instead.

### DIFF
--- a/wagtail/admin/views/pages.py
+++ b/wagtail/admin/views/pages.py
@@ -109,8 +109,7 @@ def add_subpage(request, parent_page_id):
 
     page_types = [
         (model.get_verbose_name(), model._meta.app_label, model._meta.model_name)
-        for model in type(parent_page).creatable_subpage_models()
-        if model.can_create_at(parent_page)
+        for model in parent_page.creatable_subpage_models()
     ]
     # sort by lower-cased version of verbose name
     page_types.sort(key=lambda page_type: page_type[0].lower())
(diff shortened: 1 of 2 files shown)


gasman:
Sorry this has gone unreviewed for so long...
Looking at https://github.com/wagtail/wagtail/pull/1572#issuecomment-161607913, it seems we originally made a decision _not_ to officially support instance-level custom logic in allowed-subpages business rules, to avoid the performance hit of a `specific` lookup on every row of a listing page. However, I think we can say that the N+1 query problem on the page listing is a ship that has sailed by now, so I'm inclined to accept this change... it does need a test though.

BertrandBordage:
This PR will be closed after October 2021.
If you are interested in contributing to update & complete it, please comment about it!

lb-:
As per comment above, PR should be closed.

PR #4178, opened 2018-01-08
Add panel usage checks

### LINKED ISSUE #2822: Add a system check for incorrect use of content_panels on non-page models
Using `content_panels` instead of `panels` on non-page models (including snippets, ModelAdmin, wagtailsettings, and InlinePanel child models) seems to be a very common mistake:

http://stackoverflow.com/questions/38207208/add-a-multifieldpanel-in-django-wagtail-basesettings
https://groups.google.com/d/msg/wagtail/nqb24PZRlhk/oKkbmUXZAgAJ
https://github.com/torchbox/wagtail/issues/1456
https://groups.google.com/d/msg/wagtail/hs_mX0xOoyY/fr7ktSTRAAAJ
https://github.com/torchbox/wagtail/issues/1912

I suggest we add a system check to warn about models that set a `content_panels` attribute when a `panels` attribute is expected instead. (The solution would need to take care to avoid false positives, since it _is_ valid to set `content_panels` if the edit_handler property is overridden.)

lb-:
Implements checks for incorrect panels configuration (eg. using `content_panels` on a non-page model). Resolves #2822

If someone could proof the error wording, see if it makes sense, that would be appreciated.

Also... **I need help with failing tests (if someone could suggest a good approach)**

Issue: Tests all pass individually, even on a per app basis, but when I run the full test suite I get the error (below). This is because the `edit_handler` checks throw an error on an improperly setup `CustomRichTextFieldPage`. Which in turn is because the checks run within the context of my specific panels tests, and not where `@override_settings(WAGTAILADMIN_RICH_TEXT_EDITORS={...` has been used to set the settings. This means that checking the `get_form_class` on `CustomRichTextFieldPage` throws an error.

```

======================================================================

ERROR: test_snippet_with_tabbed_panels_only (wagtail.snippets.tests.TestSnippetPanelConfigChecks)

Test that checks will warn against setting tabbed panels on snippet models

  File "/vagrant/wagtail/wagtail/snippets/tests.py", line 831, in test_snippet_with_tabbed_panels_only

    errors = [e for e in checks.run_checks() if e.id == 'wagtailadmin.W002']

  File "/home/vagrant/.virtualenvs/bakerydemo/lib/python3.4/site-packages/django/core/checks/registry.py", line 81, in run_checks

    new_errors = check(app_configs=app_configs)

  File "/vagrant/wagtail/wagtail/admin/checks.py", line 68, in get_form_class_check

    form_class = edit_handler.get_form_class(cls)

  File "/vagrant/wagtail/wagtail/admin/edit_handlers.py", line 314, in get_form_class

    widgets=cls.widget_overrides())

  File "/vagrant/wagtail/wagtail/admin/edit_handlers.py", line 61, in get_form_for_model

    return metaclass(class_name, (form_class,), form_class_attrs)

  File "/vagrant/wagtail/wagtail/admin/forms.py", line 289, in __new__

    new_class = super(WagtailAdminModelFormMetaclass, cls).__new__(cls, name, bases, attrs)

  File "/home/vagrant/.virtualenvs/bakerydemo/lib/python3.4/site-packages/modelcluster/forms.py", line 192, in __new__

    new_class = super(ClusterFormMetaclass, cls).__new__(cls, name, bases, attrs)

  File "/home/vagrant/.virtualenvs/bakerydemo/lib/python3.4/site-packages/django/forms/models.py", line 266, in __new__

    apply_limit_choices_to=False,

  File "/home/vagrant/.virtualenvs/bakerydemo/lib/python3.4/site-packages/django/forms/models.py", line 186, in fields_for_model

    formfield = formfield_callback(f, **kwargs)

  File "/vagrant/wagtail/wagtail/admin/forms.py", line 270, in formfield_for_dbfield

    return db_field.formfield(**kwargs)

  File "/vagrant/wagtail/wagtail/core/fields.py", line 18, in formfield

    defaults = {'widget': get_rich_text_editor_widget(self.editor, features=self.features)}

  File "/vagrant/wagtail/wagtail/admin/rich_text.py", line 153, in get_rich_text_editor_widget

    editor = editor_settings[name]

KeyError: 'custom'


```

Potential solution ideas:
1. Override the settings on all the panel testing, comment with explanation why we are doing this when they are not needed for the specific tests `@override_settings(WAGTAILADMIN_RICH_TEXT_EDITORS={...`.
2. Add a try/catch to the `https://github.com/wagtail/wagtail/blob/master/wagtail/admin/checks.py` inside the `for cls in get_page_models():` loop when calling `get_form_class`. However, this will also catch the error when developers have actually made a mistake by not declaring the right editor settings.
3. Split out the InlinePanel model testing into its own check, do not share it with `get_form_class_check` and then use a [custom tag](django.core.checks) `panels` to label the checks all relating to panel config checking. Then, in the testing only run checks with the tag `panels`.

----

To Complete:
* [ ] Fix Tests
* [ ] Add checks for modelAdmin registered models

Notes: 
* Docs not needed for this as it just adds additional warnings for what documentation already advises.


lb-:
Closing for now - until I get a solid PR together, I will go ahead with the idea 3 above.

PR #4182, opened 2018-01-09
Comparison of related model to use pk

### LINKED ISSUE #4180: Compare with previous version fails when model has custom primary key field
### Issue Summary

Compare with previous version fails when model has custom primary key field. `ChildRelationComparison.get_mapping` uses `id` instead of `pk`, causing any models with custom primary keys to raise AttributeError.

### Steps to Reproduce

2. Edit models.py as follows...

```
# -*- coding: utf-8 -*-
from __future__ import absolute_import, division, print_function, unicode_literals

from django.db import models
from wagtail.wagtailadmin.edit_handlers import FieldPanel, InlinePanel


class HomePage(Page):
        InlinePanel('related_models', label='Related models'),
    ]


class RelatedModel(models.Model):
    custom_id = models.AutoField(primary_key=True)
    home_page = ParentalKey('home.HomePage', related_name='related_models')
    value = models.IntegerField()

    panels = [
        FieldPanel('value'),
    ]
```
3. `makemigrations`
4. `migrate`
5. `createsuperuser`
5. `runserver`
6. login to /admin
7. edit home page
8. add a "related model" and press save draft
9. add a "related model" and press publish
10. re-edit home page
11. click revisions
12. click "compare with previous revision"

I expected it to use the correct primary key field. Code in `ChildRelationComparison.get_mapping` should use `pk` instead of `id`.

### Technical details

* Python version: 2.7.10
* Django version: 1.11.9
* Wagtail version: 1.13.1
* Browser version: whatsmybrowser.org/b/XXQTSVL

lb-:
Resolves #4180

Comparing related models now uses `.pk` instead of `.id` when comparing if the related objects have changed or not.

Added tests for id (pk) comparisons, no docs to update as this enhancement reflects expected behaviour.


lb-:
Closing PR for now, will re-do with correct form set up and also non-conflicting migrations with master (will branch off latest master, not an older one).

From Matt: added an InlinePanel to EventPage, which adds a formset to the submitted form data - that means there has to be a management form as part of the submission, even if that management form just says that it's an empty formset.

PR #4185, opened 2018-01-10
Comparison of related model to use pk instead of id

### LINKED ISSUE #4180: Compare with previous version fails when model has custom primary key field

lb-:
Resolves #4180

Comparing related models now uses .pk instead of .id when comparing if the related objects have changed or not.

Added tests for id (pk) comparisons, no docs to update as this enhancement reflects expected behaviour.


lb-:
@gasman could you check through this when you get the chance

wagtail/tests/testapp/models.py
```diff
@@ -295,6 +297,18 @@ class EventPage(Page):
 ]
 
 
+class HeadCountRelatedModelUsingPK(models.Model):
+    """Related model that uses a custom primary key (pk) not id"""
+    custom_id = models.AutoField(primary_key=True)
```
BertrandBordage: It would test more edge cases to use a non-integer primary key, here.

webit-mdowney: `UUIDField(primary_key=True, default=uuid.uuid1)`

lb-:
@BertrandBordage @gasman this has been merged but I am good to add a new PR for an additional test using a uuid primary key as well if needed.

PR #4192, opened 2018-01-11
Allow any host in development

tomdyson:
Updates the project template so the `dev` settings allow any host. Like `DEBUG = True`, this is an insecure but convenient default, appropriate for developer environments. 

Note `ALLOWED_HOSTS` is a required setting when `DEBUG = False`. In production it should be set to the specific host value.


gasman:
Merged in edbfba5af32fb8ba74d954d01f321e9b04d78b07

PR #4195, opened 2018-01-12
Added "find" API view and ability to find pages by HTML path

### LINKED ISSUE #4154: Pages detail API should allow fetching a page by path
Currently the API endpoints provided by `wagtail.api.v2.endpoints.PagesAPIEndpoint` do not provide the ability to fetch a Page by the page's URL.

The docs regarding [filtering pages](http://docs.wagtail.io/en/v1.13.1/advanced_topics/api/v2/usage.html#filtering) makes it seem like a request such as

`curl http://mysite.com/api/pages/?html_url=<some_url>`

should work but instead the API returns:

```
{
    "message": "query parameter is not an operation or a recognised field: html_url"
}
```

Would it be possible to support filtering on this field? Perhaps also making the detail view able to return the page by it's path? i.e. allowing all of:

```
# filtering
curl http://mysite.com/api/pages/?html_url=<my_page_url>

# detail
curl http://mysite.com/api/pages/1
curl http://mysite.com/api/pages/my_page_path
```

If this is something that wagtail would like to add then I may be able to try submitting a PR.

The motivation for this feature is to make it easy for a frontend webapp to fetch pages and render them. It would be useful for SPA apps if they could integrate with a wagtail CMS backend by using the current pages URL to do a wagtail API request and use the JSON response to render a page using whatever framework the SPA is written with (react/angular/vue etc.).

kaedroho:
This implements a new "find" view for all endpoints which can be used for finding an individual object based on the URL parameters passed to it.

If an object is found, the view will return a ``302`` redirect to detail page of that object. If not, the view will return a ``404`` response.

For the pages endpoint, I've added a ``html_path`` parameter to this view, this allows finding a page by its path on the site.

For example, a GET request to ``/api/v2/pages/find/?html_path=/`` will always generate a ``302`` response to the detail view of the homepage of the current site. This uses Wagtail's internal routing mechanism so routable pages are supported as well.

Fixes #4154

### DIFF
--- a/docs/advanced_topics/api/v2/usage.rst
+++ b/docs/advanced_topics/api/v2/usage.rst
@@ -422,6 +422,15 @@ All exported fields will be returned in the response by default. You can use the
 For example: ``/api/v2/pages/1/?fields=_,title,body`` will return just the
 ``title`` and ``body`` of the page with the id of 1.
 
+Finding pages by HTML path
+--------------------------
+
+You can find an individual page by its HTML path using the ``/api/v2/pages/find/?html_path=<path>`` view.
+
+This will return either a ``302`` redirect response to that page's detail view, or a ``404`` not found response.
+
+For example: ``/api/v2/pages/find/?html_path=/`` always redirects to the homepage of the site
+
 Default endpoint fields
 =======================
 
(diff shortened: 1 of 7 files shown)


gasman:
Looks good to me! Merged in 17f7f70170911ac3745cf002fb2cca054581d5da.

PR #4197, opened 2018-01-15
Add `wagtail updatemodulepaths` command to update imports for Wagtail 2.x

gasman:
Adds a `wagtail updatemodulepaths [/path/to/project]` that can be run over a project codebase to convert all module path references to the new Wagtail 2.x versions (e.g. `wagtail.wagtailcore` -> `wagtail.core`).

The command simply runs a search-and-replace for the full dotted path over all *.py files, which seems to be robust enough to catch the references that need changing (imports, INSTALLED_APPS, references in migrations) without any false positives (database table names, `{% load wagtailcore_tags %}` etc). When testing on bakerydemo it just misses one reference, in app.json - I figured it was better to be conservative and only check *.py files, so that we aren't running rampant over media folders, `node_modules` and so on.

It's not hard to think of edge cases it'll fail on (`from wagtail import wagtailcore`, `INSTALLED_APPS = ['wagtail.' + 'wagtailcore']`...) but I can't see those coming up in real-world code. Have added appropriate disclaimers to the release notes, though.

### DIFF
--- a/wagtail/bin/wagtail.py
+++ b/wagtail/bin/wagtail.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
+import fileinput
 import os
+import re
 from optparse import OptionParser
 
 from django.core.management import ManagementUtility
@@ -62,8 +64,74 @@ class CreateProject(Command):
         print("Success! %(project_name)s has been created" % {'project_name': project_name})  # noqa
 
 
+class UpdateModulePaths(Command):
+    description = "Update a Wagtail project tree to use Wagtail 2.x module paths"
+    usage = "Usage: %prog updatemodulepaths [root-path]"
+
+    REPLACEMENTS = [
+        (re.compile(r'\bwagtail\.wagtailcore\b'), 'wagtail.core'),
+        (re.compile(r'\bwagtail\.wagtailadmin\b'), 'wagtail.admin'),
+        (re.compile(r'\bwagtail\.wagtaildocs\b'), 'wagtail.documents'),
+        (re.compile(r'\bwagtail\.wagtailembeds\b'), 'wagtail.embeds'),
+        (re.compile(r'\bwagtail\.wagtailimages\b'), 'wagtail.images'),
+        (re.compile(r'\bwagtail\.wagtailsearch\b'), 'wagtail.search'),
+        (re.compile(r'\bwagtail\.wagtailsites\b'), 'wagtail.sites'),
+        (re.compile(r'\bwagtail\.wagtailsnippets\b'), 'wagtail.snippets'),
+        (re.compile(r'\bwagtail\.wagtailusers\b'), 'wagtail.users'),
+        (re.compile(r'\bwagtail\.wagtailforms\b'), 'wagtail.contrib.forms'),
+        (re.compile(r'\bwagtail\.wagtailredirects\b'), 'wagtail.contrib.redirects'),
+        (re.compile(r'\bwagtail\.contrib\.wagtailfrontendcache\b'), 'wagtail.contrib.frontend_cache'),
+        (re.compile(r'\bwagtail\.contrib\.wagtailroutablepage\b'), 'wagtail.contrib.routable_page'),
+        (re.compile(r'\bwagtail\.contrib\.wagtailsearchpromotions\b'), 'wagtail.contrib.search_promotions'),
+        (re.compile(r'\bwagtail\.contrib\.wagtailsitemaps\b'), 'wagtail.contrib.sitemaps'),
+        (re.compile(r'\bwagtail\.contrib\.wagtailstyleguide\b'), 'wagtail.contrib.styleguide'),
+    ]
+
+    def run(self, parser, options, args):
+        # Validate args
+        if len(args) > 2:
+            parser.error("Too many arguments")
+
+        try:
+            root_path = args[1]
+        except IndexError:
+            root_path = os.getcwd()
+
+        checked_count = 0
+        changed_count = 0
+
+        for (dirpath, dirnames, filenames) in os.walk(root_path):
+            for filename in filenames:
+                if not filename.lower().endswith('.py'):
+                    continue
+
+                path = os.path.join(dirpath, filename)
+                checked_count += 1
+                changed = self._rewrite_file(path)
+                if changed:
+                    print(path)  # NOQA
+                    changed_count += 1
+
+        print("\nChecked %d .py files, %d files updated." % (checked_count, changed_count))  # NOQA
+
+    def _rewrite_file(self, filename):
+        changed = False
+
+        with fileinput.FileInput(filename, inplace=True) as f:
+            for original_line in f:
+                line = original_line
+                for pattern, repl in self.REPLACEMENTS:
+                    line = re.sub(pattern, repl, line)
+                print(line, end='')  # NOQA
+                if line != original_line:
+                    changed = True
+
+        return changed
+
+
 COMMANDS = {
     'start': CreateProject(),
+    'updatemodulepaths': UpdateModulePaths(),
 }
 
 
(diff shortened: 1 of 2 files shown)


tomdyson:
It would be good to have a `--dry-run` option which prints the number of changes in each relevant file, e.g.

```
core/models.py - 3 changes
core/views.py - 1 change
search/views.py - 1 change
```

gasman:
@tomdyson That would indeed be a nice addition, but it requires some fiddly refactoring of how we do the command-line parsing - currently we hand off to optparse _before_ determining which command has been requested, which makes it hard to have command-specific options.

tomdyson:
@gasman couldn't we check if there's an `args[2]` that equals '--dry-run'?

Failing that, perhaps we should just report on the changes that have been made, in a similar format.

lb-:
@gasman - I might be missing something but I cannot get the CLI updated, I am trying to run the commands by updating a local project (not Bakerydemo) with this tool.

I am using a virtual environment and ran (within the environment):
* `$ pip install git+https://github.com/wagtail/wagtail.git`
* `$ pip freeze -l > requirements.txt`
* `$ wagtail updatemodulepaths ./`

All I get is:
```
Usage: wagtail start project_name [directory]

wagtail: error: Unrecognised command: updatemodulepaths
```

What am I missing?

Everything appears to be updated within the venv to `wagtail==2.0a0`, this is on Python 3.6 with Django 1.11.7.

venv/lib/python3.6/site-packages/ has both a `wagtail` folder (which contains old wagtail) and a `wagtail-2.0a0-py3.6.egg-info` folder.

gasman:
Hi @lb-,
That `pip install` line will install the Wagtail master branch, which doesn't include this PR :-)

Changing it to something like `pip install git+https://github.com/gasman/wagtail.git@feature/update_module_paths_command` might work, but I'd recommend going to the base of a Wagtail development instance where you have this branch checked out, and running: `pip install -e .` . That'll link the development instance into your virtualenv, without needing anything to be installed in the virtualenv's libs folder.

lb-:
🤦‍♂️ - thanks - on it now

wagtail/bin/wagtail.py
```diff
@@ -1,66 +1,154 @@
...
+
+    def _rewrite_file(self, filename):
+        changed = False
+
+        with fileinput.FileInput(filename, inplace=True) as f:
+            for original_line in f:
+                line = original_line
+                for pattern, repl in self.REPLACEMENTS:
+                    line = re.sub(pattern, repl, line)
+                print(line, end='')  # NOQA
```
lb-: What is this meant to print? I only saw logging for each filename and the final summary of how many files were changed (and how many searched).

lb-:
Hey @gasman - epic! this is really helpful.

It worked as intended, all project files (including migration files) updated. I made one comment in the codebase and below is my other thoughts.

## 1. Honour .gitignore - ignored filed and locations.

In the project I tested on we have our `venv` inside a folder named `venv3` (because of the legacy change over to Python 3). Maybe some folders should be ignored? or see if there are folders within .gitignore to not check.

I ended up having to check the following files for changes that should not have been made.

```
...<redacted>.../venv3/lib/python3.6/site-packages/wagtailerrorpages/templatetags/wagtailerrorpages_tags.py
...<redacted>.../venv3/lib/python3.6/site-packages/wagtailfontawesome/wagtail_hooks.py
...<redacted>.../venv3/src/wagtail/wagtail/search/views/frontend.py
...<redacted>.../venv3/src/wagtail/wagtail/wagtailsearch/tests/test_page_search.py
```

I assume that this will also make it faster (not that this is a real concern).


## 2. Logging

Logging worked, showed the name of each file changed and also ended with the following:
`Checked 5550 .py files, 89 files updated.`
Note: This project has no where near 5550 actual custom files, so other folders must have been checked inside venv3.

Suggestion: It might be good to also mention how many lines were changed, gives a quick reference for any git status output on rough amount of changes.


## 3. `dry-run` Option

I like Tom's suggestion of having this, I do not know if it is easy to implement as per your notes above. I guess we should not assume everyone uses a VCS.

My only input into this would be to make the default non-destructive, and have it so that you need to add a flag to actually make the changes. Making it harder to change files that should not be (eg. the files in my venv3 directory).

I do not see this as a blocker though as this is a single use tool.


## 4. Warn about use of `wagtail.contrib.wagtailapi`

It would be really nice if any reference to `wagtail.contrib.wagtailapi` (deprecated) returned a warning to the console. Maybe on a new line at the end of the logging with a reminder to remove `wagtail.contrib.wagtailapi`.

This is something that should not be removed by the script but a warning would help as I got a very confusing error when trying to run the server (see below). 

```
  File "/Users/my-user/Documents/dev/...<removed>/venv3/lib/python3.6/site-packages/django/utils/autoreload.py", line 228, in wrapper
    fn(*args, **kwargs)
  File "/Users/my-user/Documents/dev/...<removed>/venv3/lib/python3.6/site-packages/django/core/management/commands/runserver.py", line 117, in inner_run
    autoreload.raise_last_exception()
  File "/Users/my-user/Documents/dev/...<removed>/venv3/lib/python3.6/site-packages/django/utils/autoreload.py", line 251, in raise_last_exception
    six.reraise(*_exception)
  File "/Users/my-user/Documents/dev/...<removed>/venv3/lib/python3.6/site-packages/django/utils/six.py", line 685, in reraise
    raise value.with_traceback(tb)
  File "/Users/my-user/Documents/dev/...<removed>/venv3/lib/python3.6/site-packages/django/utils/autoreload.py", line 228, in wrapper
    fn(*args, **kwargs)
  File "/Users/my-user/Documents/dev/...<removed>/venv3/lib/python3.6/site-packages/django/__init__.py", line 27, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/Users/my-user/Documents/dev/...<removed>/venv3/lib/python3.6/site-packages/django/apps/registry.py", line 85, in populate
    app_config = AppConfig.create(entry)
  File "/Users/my-user/Documents/dev/...<removed>/venv3/lib/python3.6/site-packages/django/apps/config.py", line 127, in create
    import_module(entry)
  File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 978, in _gcd_import
  File "<frozen importlib._bootstrap>", line 961, in _find_and_load
  File "<frozen importlib._bootstrap>", line 948, in _find_and_load_unlocked
```

wagtail/bin/wagtail.py
gasman: Here the FileInput class is redirecting standard output to the file (see the explanation under 'Optional in-place filtering' at https://docs.python.org/3/library/fileinput.html#fileinput.FileInput) - this is writing the updated line back to the file.

lb-: ok :)

gasman:
Have now added `--list` and `--diff` options to view the changes without actually updating the files. I like the idea of making it non-destructive by default, but it would have made the semantics a bit weird: we'd really need to change the command name to something that doesn't imply updating, and choose one of `--list` or `--diff` to be the default action (or make the action a required option, which breaks standard command line conventions). Other command line tools like `./manage.py makemigrations` and `isort` do it this way round without any controversy, anyway :-)

gasman:
Covering @lb-'s other points:

## Honour .gitignore - ignored filed and locations.

I'll add `--ignore-dir` and `--ignore-file` options to the command to specify individual dirs/files (and, if possible, wildcard patterns) to be ignored. Digging into .gitignore files seems like too much "magic" for me: I don't think it's necessarily true that files in .gitignore correspond to ones that should be skipped here (for example, we probably _do_ want to rewrite settings/local.py files), and if we're giving git special treatment, what about `.hgignore`, `.svnignore`...

## Logging

The file list output now shows number of changes per file (and, of course, `--diff` mode makes the changes completely apparent) so I think this is covered now.

## Warn about use of `wagtail.contrib.wagtailapi`

This feels like feature creep to me :-) The removal of wagtailapi is covered as a separate point in the release notes; I'd rather have this command do one task well rather than straying into other kinds of checks relating to 2.0 upgrades. (It's also not clear to me how a warning would fit into the `--diff` output.)

gasman:
Merged in fa77440749033e2822f01c0c74ddc84592fd16df

lb-:
Epic

Issue #3616, opened 2017-05-27
A user can explore(see) any page in admin area?


kidlj:
Issue:

The site structure is as following: 

```python
Homepage
  |- Blog Index
  |- Event Index
  |- People
  |- Contact Page
  | - User A
       |- Test Page 1
       |- Test Page 2
```

Now I add a user called 'UserA', who belongs to roles group of 'UserA' (the same name), and group 'UserA' is set to only has 'add' & 'publish' permissions to 'User A' page.

The left nav bar in admin area works fine, 'UserA' can only see 'User A' as the root; But when clicking on the wagtail bird icon in admin page, 'UserA' can see Homepage and all its children, like 'Blog Index' and 'Event Index'. 

So it's like a user can view any pages in Wagtail admin area? Is this on purpose?

What I really expect is a user can only see(explore) pages owned by himself, or pages that his role group has access to. It's like user isolation. Can this be done? Thanks.
  
--------------------
Project: wagtail-demo

Python: 3.6.1
Django: 1.10
Wagtail: 1.10.1

gasman:
> But when clicking on the wagtail bird icon in admin page, 'UserA' can see Homepage and all its children, like 'Blog Index' and 'Event Index'.

Where exactly do you see these? When I access the admin dashboard page by clicking the bird icon as an ordinary editor, the only place I see pages listed are the "Your most recent edits" listing and the left nav menu.

kidlj:
@gasman after clicking the bird icon, then click '25 Pages', you should see all pages.
![watail_pages](https://user-images.githubusercontent.com/300616/27334442-e9909d58-55fb-11e7-898e-2c1bb26de64b.png)

gasman:
@kidlj Thanks for the clarification! I guess there are two things that need to be fixed here:

* when a user visits an explorer listing page that is outside of their [explorable root page](https://github.com/wagtail/wagtail/blob/e73c97463f71717a42ed2f13d02e2ef42c071267/wagtail/wagtailadmin/navigation.py#L18), redirect to their explorable root page
* The page count shown in the dashboard should only count pages that the user has edit permission for.

gasman:
Fixed in #4206 and #4207

PR #4206, opened 2018-01-18
Make site summary panel respect user privileges

chosak:
This change modifies how the Wagtail home site summary panel displays the number of pages on the site, and where that number links to.

Instead of showing the total number of pages on the site, the panel should show the number of pages under the user's explorable root page (inclusive). If the user has access to the full tree, the Wagtail root is not counted in this total.

Previously, the site summary page link would go to the Wagtail root if there were multiple sites in an installation, and to the site root page for a single site. This change modifies this logic so that the link always goes to the user's explorable root page (which may be their explorable root page).

The unit tests for the site summary panel have been pulled out into a new module at `wagtail.admin.tests.test_site_summary`, and augmented to test how things work for users with different permissions.

This partially addresses #3616, and is a small subset of the many improvements proposed in #2463. There remain a few issues around navigating the page tree with limited permissions; for example, users can still copy or see revisions on pages they shouldn't have permissions over.

I'm planning on opening a subsequent PR to address the other half of #3616, which would prevent users from navigating around parts of the tree that don't lie under their explorable root page.

### DIFF
--- a/wagtail/admin/site_summary.py
+++ b/wagtail/admin/site_summary.py
@@ -1,8 +1,9 @@
 from django.template.loader import render_to_string
 
+from wagtail.admin.navigation import get_explorable_root_page
 from wagtail.admin.utils import user_has_any_page_permission
 from wagtail.core import hooks
-from wagtail.core.models import Page, Site
+from wagtail.core.models import Page
 
 
 class SummaryItem:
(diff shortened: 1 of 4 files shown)


gasman:
Merged (with tweak as per https://github.com/wagtail/wagtail/pull/4206#pullrequestreview-111973892) in 5c9ff23e229acabad406c42c4e13cbaea32e6c15 - thanks @chosak!

PR #4207, opened 2018-01-18
Prevent navigation to pages outside user explorable root

chosak:
This change prevents non-admins from navigating around the Wagtail page tree for pages that lie outside of their explorable root. Currently, non-admins can hit any page in the tree using a URL like `/admin/pages/123/` even if they don't have any permissions over that page or its part of the page tree.

This change adds a (temporary) redirect to requests like this, so that users may not navigate to parts of the tree that lie outside outside of their explorable site root, as determined by the page privileges they have. If they try to hit a URL like the one above, they get redirected to their explorable site root navigation page instead.

Relevant unit tests have been modified to incorporate this change.

This PR addresses part of #3616. It partially incorporates some of the changes in #2463, but, like #4206, leaves some existing issues with non-admin page navigation; for example, users can still copy or see revisions on pages they shouldn't have permissions over.

### DIFF
--- a/wagtail/admin/tests/test_pages_views.py
+++ b/wagtail/admin/tests/test_pages_views.py
@@ -361,25 +361,21 @@ class TestPageExplorerSignposting(TestCase, WagtailTestUtils):
     def test_nonadmin_at_root(self):
         self.assertTrue(self.client.login(username='siteeditor', password='password'))
         response = self.client.get(reverse('wagtailadmin_explore_root'))
-        self.assertEqual(response.status_code, 200)
-        # Non-admin should get a simple "create pages as children of the homepage" prompt
-        self.assertContains(
-            response,
-            "Pages created here will not be accessible at any URL. "
-            "To add pages to an existing site, create them as children of the homepage."
+
+        # Non-admins should be redirected to their explorable root.
+        self.assertEqual(
+            (response.status_code, response['Location']),
+            (302, reverse('wagtailadmin_explore', args=(self.site_page.pk, )))
         )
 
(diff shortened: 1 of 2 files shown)


wagtail/admin/tests/test_pages_views.py
```diff
@@ -361,25 +361,21 @@ def test_admin_at_site_page(self):
...
-            (
-                "There is no site record for this location. "
-                "Pages created here will not be accessible at any URL."
-            )
+
+        # Non-admins should be redirected to their explorable root.
+        self.assertEqual(
+            (response.status_code, response['Location']),
+            (302, reverse('wagtailadmin_explore', args=(self.site_page.pk, )))
         )
```
gasman: As it stands, these tests are wrongly conflating "non-admin" with "has no permission for this view" - it's possible for a non-admin user to legitimately have access to the root section here (either by having that permission explicitly assigned at the group level, or by having permission over several deeper sections that have no common ancestor other than the root). The old version of these tests was at fault here, by failing to set up the appropriate permissions for the non-admin user - which, of course, didn't matter until we tightened up the permission checks here. There are actually three cases we need to test here, for both the root and non-root views:

* No permission for this view: user should be redirected away
* Ordinary user with permission for this view: should see the listing, with a warning message that pages created here won't be accessible
* Admin (or more precisely, any user with 'add site' permission): should see the listing, with a warning message that pages created here won't be accessible including advice about configuring a site record.

gasman:
Merged (with additional tests) in d1830c090981515195a1a7bdb8fd802cf40f3312 - thanks @chosak!

PR #4213, opened 2018-01-25
Slugurl dead code

rgant:
Unless I am misunderstanding something everything after the if-else block testing for a page object from a slug is unreachable. So I've removed it, the unnecessary `return None`, and added tests to confirm that the slug not existing case still performs the same as before the changes.

* I think test results are as expected for current master branch, but there were a large number of RemovedInDjango21Warning and DeprecationWarning in the output:
    Ran 2951 tests in 165.442s
    OK (skipped=188, expected failures=8)
* I think lint results are as expected for the current master branch:
    90 ERROR:  Imports are incorrectly sorted.
* Existing test coverage lacked tests for the failure case. I've added some for slugurl in wagtail.core.tests.test_jinja2 and wagtail.core.tests.tests
* No Documentation changes are needed in my opinion.

### DIFF
--- a/wagtail/core/templatetags/wagtailcore_tags.py
+++ b/wagtail/core/templatetags/wagtailcore_tags.py
@@ -40,14 +40,6 @@ def slugurl(context, slug):
     else:
         return None
 
-    try:
-        current_site = context['request'].site
-    except (KeyError, AttributeError):
-        # request.site not available in the current context; fall back on page.url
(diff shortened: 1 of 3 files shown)


gasman:
Thanks @rgant - good catch! Merged in 218a47e8025e8044ae10b0fa61b6e0384078aa75.

Issue #2305, opened 2016-03-03
Let admin fields span full width when there is no help text


alexgleason:
Would negate the need for: https://github.com/alexgleason/wagtailhavetheabilitytoactuallyseefieldsintheadmin

alexgleason:
Alternatively, put help text in a different position for StreamField blocks. Such as below the label. IMO this is a better solution.

JoshBarr:
Here's a proposal for how to handle help text in the editor view:

![editor-1](https://cloud.githubusercontent.com/assets/1715456/13505967/8264e17e-e1e0-11e5-86ad-04561a51b412.png)

alexgleason:
Thanks Josh. I like it, but I think we still run into a problem when deeply nesting blocks.

I'm starting to take the position that this sort of content should have custom styles added to them. Here's an example of one I recently created:

![screenshot from 2016-03-03 13 32 36](https://cloud.githubusercontent.com/assets/3639540/13506612/eea76594-e14c-11e5-911e-ab15086f1596.png)

The whole thing (the dropdown and cards) is a structblock where the cards are a field within it. Like this:

```python
cards = blocks.StructBlock([
    ('style', blocks.ChoiceBlock(choices=[
        ('smallcard', 'Small card'),
        ('circlecard', 'Circle card'),
        ('squarecard', 'Square card')
    ])),
    ('cards', CardsBlock()),
])
```

I think taking the time to design deep fields like this in the admin could help mitigate the issue. But I'm still a bit bothered by the help text like that. Because in order to achieve this, it required me to inject this custom CSS into the admin:

```css
.sequence-member.blockname-card .struct-block .fields {
  width: 100%;
}
```

That means every time I create a block like this, I'll have to inject in CSS for it.

davecranwell:
This issue is essentially a replica of at least two others, all concerned with layout of the page editor, with which I completely agree. As Wagtail's popularity grows, so do the numbers of users trying to do more with this interface, and not getting very far.

I think we need to have a proper conversation taking them all into consideration. I'll try and convene a meeting of minds.

Josh's design certainly looks the most rich and professional, but it needs extrapolating to the various layouts we have and I suspect it may still be a particular problem given the potentially infinite nesting of streamfields.

That isn't to say there's not a sweet spot in between, but the full-width label-above-input approach is certainly the most flexible, even if it's rather simple and has value given the nesting problem.

alexgleason:
[...]

PR #4215, opened 2018-01-25
Couple of CSS tweaks for nested StructBlocks.

rgant:
This just tries to make the layout for nested blocks slightly better. There are several tickets addressing major redesign (#2325 seems dead, #3942 was moved to 2.1 milestone, #2305 seems dead), but they all seem to be years old with little hope of immediate fixes. I just want this UI to be slightly better so my editors don't spend all their time complaining about the UI instead of writing content.

This is just a style change, so no python was modified to need testing or linting. They still look as expected for the current master branch.

This is what the UI for tested StructBlocks looked like before:
nested structblocks-before

This is what it looks like after:
nested structblocks-after

This is the whole form from before:
nested-before

And this is after:
nested-after

It's not perfect, but it is at least better for now.

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/scss/components/_streamfield.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/components/_streamfield.scss
@@ -26,6 +26,13 @@ li.sequence-member {
         opacity: 1;
     }
 
+    .struct-block .object-help {
+        float: right;
+        margin-top: 0;
+        padding: 0;
+        width: auto;
+    }
+
     .struct-block .fields {
         @include column(10);
         padding-left: 0;
@@ -33,7 +40,7 @@ li.sequence-member {
     }
 
     .struct-block .sequence-container {
-        @include column(9);
+        @include column(10);
         padding-left: 0;
         padding-right: 0;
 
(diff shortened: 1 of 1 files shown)


tomdyson:
Thanks @rgant!

@jjanssen - could you please give this a look? As @rgant says, this doesn't address everything that #3942 addresses, but it's a quick fix for the main issue.

thibaudcolas:
I appreciate any and all improvements but I think we should prioritise #3942 over this. Even though the change may appear small ("CSS tweaks", "just a style change"), it can take hours to properly review this, and even longer if something isn't working and we need to dig deeper. I'd rather invest this time in a more thorough fix.

I'm also pretty sure the layout you made could be achieved by only removing styles rather than adding overrides which increase the overall complexity.

---

Anyhow, I'd rather see #3942 move faster but if someone else does want to get this quick-fix in by all means go for it.

jjanssen:
Hi @rgant,

Sorry it took us a while to get back on this. For the upcoming Wagtail 2.2 release we discussed whether or not to go for this quick-win, even though it would take up quite an amount of testing this properly. As of this morning we noticed we could not reproduce this scenario like described in your screenshots. Turns out most of the gaps for the help text were already handled by this PR #4281 in order to fix #4247.

Current situation:
image

The main difference is that help-text does not align to the right but doesn't cause large gaps anymore, so I'm closing this pull request for now.

PR #4218, opened 2018-01-26
Replace triple dots usage with ellipsis character

thibaudcolas:
Spotted [while working on #4136](https://github.com/wagtail/wagtail/pull/4136#discussion_r162900195) with @BertrandBordage.

This is essentially a big search and replace. It's probably easier to review by looking at the commits individually rather than the whole diff.

### DIFF
--- a/client/src/components/LoadingSpinner/__snapshots__/LoadingSpinner.test.js.snap
+++ b/client/src/components/LoadingSpinner/__snapshots__/LoadingSpinner.test.js.snap
@@ -7,6 +7,6 @@ exports[`LoadingSpinner basic 1`] = `
     name="spinner"
     title={null}
   />
-   Loading...
+   Loading…
 </span>
 `;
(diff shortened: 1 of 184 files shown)


thibaudcolas:
Thanks for the review @BertrandBordage. To be honest I have no idea how the translation workflow works with Django and Transifex – should I simply remove the changes to the `*.po` files?

I see there is [a script](https://github.com/wagtail/wagtail/blob/c6666c6de5e83bf94d18324858c121e6584ba47d/scripts/fetch-translations.sh) that uses a `django compilemessages` command, but there doesn't seem to be any docs beyond that – is this what I need to use?

gasman:
The translation workflow is documented at https://github.com/wagtail/wagtail/wiki/Managing-Wagtail-translations - I'll handle that when I make the release.

I _think_ we can push the local changes in .po files back to Transifex, so let's keep this PR as it is - that way, we can save translators the job of fixing these strings manually.

BertrandBordage:
@gasman: That’s why I approved the review, I just wanted to point out for the future that it was not necessary nor usual to do the translation work here :)

PR #4219, opened 2018-01-26
Add Codecov integration to CircleCI to track JS test coverage

thibaudcolas:
We calculate JS test coverage on every CI build, but don't actually send it to Codecov. This will motivate us to reach a higher coverage there as well.

I tried to use Codecov [flags](https://docs.codecov.io/docs/flags), which allow us to retain some separation between the two test suites (#frontend, #backend tags) but it's the first time I use this so not entirely sure how best to leverage it.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -17,3 +17,5 @@ test:
     - python -u runtests.py
     - npm run test:unit:coverage -- --runInBand
     - npm run dist
+  post:
+    - bash <(curl -s https://codecov.io/bash)
--- a/circle.yml
+++ b/circle.yml
@@ -18,4 +18,4 @@ test:
     - npm run test:unit:coverage -- --runInBand
     - npm run dist
   post:
-    - bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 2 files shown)


thibaudcolas:
@loicteixeira could you have a look at this? There isn't much to review, just looking at https://circleci.com/gh/wagtail/wagtail/1471 and https://codecov.io/gh/wagtail/wagtail/compare/ebbbd7ad491fe09a042afe19dab9d0ea57040646...fb801fd19e6260b2a2063be5844fb1689dca24ff, and perhaps the Codecov docs to make sure I haven't done anything stupid.

circle.yml
```diff
@@ -17,3 +17,5 @@ test:
     - python -u runtests.py
     - npm run test:unit:coverage -- --runInBand
     - npm run dist
+  post:
+    - bash <(curl -s https://codecov.io/bash) -F frontend
```
loicteixeira: It scares me a bit to directly execute a script from the internet but its in their doc so ¯\_(ツ)_/¯

thibaudcolas:
Thank you @loicteixeira!

PR #4176, opened 2018-01-06
Make font more readable in docs

lb-:
Updates fonts - see #4094

PR #4094, opened 2017-12-04
add Wagtail colours to documentation sidebar

lb-:
More fleshed out concept of styling docs for #4081 - only affects the sidebar.

This is probably more for discussion/testing than merging at this stage.

Issue #333, opened 2014-06-18
Roles checkbox UX 


tmsndrs:
Hi, 

The checkboxes used on the user roles page use a grey tick when not selected. This suggests that the ability to apply permissions has been disabled. The checkboxes should be empty when the permission is not applied. The checkbox border may need to be made more prominent. 

Cheers,

PR #4224, opened 2018-01-29
Makes font in the documentation darker

m1kola:
Previous font size fails AAA contrast check. Also it's hard to read for me personally 😢 

Related to #4094 and #4176

### DIFF
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -6,7 +6,7 @@ body.wy-body-for-nav {
     -webkit-font-smoothing: antialiased; /* Do not remove! */
     font-family: "Lato", "proxima-nova", "Helvetica Neue", Arial, sans-serif;
     line-height: 1.5em;
-    color: #666; /*$color-text-base*/
+    color: #404040; /*$color-grey-1-1*/
 }
 
 div.wy-nav-side {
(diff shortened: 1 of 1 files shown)


docs/_static/css/custom.css
```diff
@@ -6,7 +6,7 @@ body.wy-body-for-nav {
     -webkit-font-smoothing: antialiased; /* Do not remove! */
```
m1kola: I have some doubts about antialiasing: it makes font thinner in webkit browsers (see screenshots), but I didn't remove it because of the `/* Do not remove! */` comment.

without_antialiasing

antialiasing

thibaudcolas:
IMHO it should just be `#333` text on white, with a font stack that doesn't use web fonts like `-apple-system, BlinkMacSystemFont, 'Helvetica Neue', Helvetica, Arial, sans-serif`. It's technical documentation, we want to be able to read the body copy consistently and without waiting for a font to load. The color in the sidebar is enough to give the site some flavour.

For font-smoothing – in my opinion it's a big no-no for grey on white body copy. It's detrimental to readability as well.

gasman:
@thibaudcolas's proposed changes added in 73cba09a2d5339004d153943dc2c29459433828c . The font change feels a bit jarring, but I think it's just unfamiliarity...

screen shot 2018-02-09 at 17 05 46

(Note: As part of this change I've gone ahead and removed the `/* Do not remove! */` - @lb-, please advise if this is a bad idea :-) )

lb-:
No issues with removing the comment. Thanks for fixing up the colours @m1kola

PR #4226, opened 2018-01-30
Add panel configuration checks

### LINKED ISSUE #2822: Add a system check for incorrect use of content_panels on non-page models

lb-:
Closes #2822 

Adds checks (using Django checks framework) for common mistakes when setting up panels for model editing.

* Tests run/pass locally + tests added for all model types affected
* No docs needed as this simply reinforces correct behaviour
* Linting has been checked

@gasman I would love your thoughts on this as you created the original issue.


lb-:
@mikedingjan - been a little while but let me know if I can assist with getting this reviewed, happy to close and open as a new PR if needed (as there will be conflicts now).

mikedingjan:
Hi @lb- 
I've reviewed this in the past already, seems like I forgot to mention it here, all worked fine, tested it in a few projects of mine.. I saw some commented code, maybe you can remove that and resolve the merge issues, and I'll do a final check?

lb-:
@mikedingjan  - awesome, thank you. I will get a new/updated PR ready with the conflicts resolved

mikedingjan:
@lb- Nice, looking forward too see it, maybe you can assign/tag me, so it won't get lost (again).. :)

lb-:
This PR has been replaced by #5093

Issue #3398, opened 2017-02-23
Wagtailadmin form field helpers should allow HTML in help_text


gasman:
The Django docs indicate that `help_text` should be treated as an HTML string and displayed unescaped on forms:
https://docs.djangoproject.com/en/1.10/ref/models/fields/#help-text

wagtailadmin/shared/field.html [does not do this](https://github.com/wagtail/wagtail/blob/7d8a1a0d1e37f0f8c614cba55bbdb7befee12d07/wagtail/wagtailadmin/templates/wagtailadmin/shared/field.html#L14), which can lead to incorrect treatment of help text that comes from Django itself. Specifically, the 'change password' form tries to display the `AUTH_PASSWORD_VALIDATORS` rules as a `<ul>`:

![screen shot 2017-02-23 at 11 36 03](https://cloud.githubusercontent.com/assets/85097/23257552/47568a84-f9bc-11e6-92e6-85a5a8218f25.png)

We should change our template to use `{{ field.help_text|safe }}` - but before we do so, we need to audit the help text of our forms to ensure that it really is HTML-safe (i.e. any `<` `>` `&` characters are escaped as HTML entities, and any included user-supplied strings are escaped using `django.utils.html.escape()`).

strindhaug:
Adding `|safe` a few places is certainly a small task, but auditing 3784 lines of po-files containing the character `>` and 4068 files containing the character `<` (most of the extra lines here is admittedly from `nplurals` settings) and 191 lines containing `&` (these are also mostly from the nplurals statements) manually is not really a *small* task. 

Especially since most of these strings are in transifex and thus not very easily "greppable".

Though, I tried grepping for `<`, `>` and `&` with different combinations of leading and trailing spaces, (excluding the nplurals statements) with the assumption that these characters will not normally be used without leading and trailing space in normal text (e.g. when not html).

And I found only this string:
`wagtailcore/locale/ja/LC_MESSAGES/django.po:55:"はい >> このサイトはサイトエントリーのないホスト名のもとでのリクエストをも処"` seems to be the only one where a `>` is used not as part of a html tag.

gasman:
[...]

PR #4234, opened 2018-02-01
Prevents field help text returning as unsafe text

luizboaretto:
Change Password page you see code instead of HTML formatted text.

captura de tela 2018-02-01 as 10 13 56

Python version: 3.6.4
Django version: 2.0.1
 Wagtail version: 2.0b1


gasman:
Thanks @luizboaretto! However, please see the discussion on #3398 - before we add `|safe` here, we need to audit all of our help text to confirm that this won't be introducing any HTML injection vulnerabilities.

PR #4237, opened 2018-02-02
Override rich text image html in 1.13.x

### LINKED ISSUE #4236: Make formats.image_to_html use the rendition.img_tag to allow overriding markup in rich text editor
I have overridden the wagtail image to add a caption field as documented in docs/advanced_topics/images/custom_image_model. However, the formats.py:image_to_html does not use the rendition.img_tag so the new field can not be used in rich text embedded images.

The change appears to be relatively small. I will submit two PRs - one for stable/1.13x and one for master

Many thanks!

dibell:
Closes #4236


gasman:
As this is more of an enhancement than a bugfix, it wouldn't be something we'd merge into a 1.13.x release - those are generally reserved for security / data loss bugfixes - so I'll close this in favour of #4238.

PR #4238, opened 2018-02-02
Override rich text image html on master

### LINKED ISSUE #4236: Make formats.image_to_html use the rendition.img_tag to allow overriding markup in rich text editor

dibell:
Closes #4236 on master

### DIFF
--- a/docs/advanced_topics/images/custom_image_model.rst
+++ b/docs/advanced_topics/images/custom_image_model.rst
@@ -41,6 +41,18 @@ Here's an example:
     class CustomRendition(AbstractRendition):
         image = models.ForeignKey(CustomImage, on_delete=models.CASCADE, related_name='renditions')
 
+        # Override img_tag from Abstract rendition to use your new fields
+        def img_tag(self, extra_attributes={}):
+            attrs = self.attrs_dict.copy()
+            attrs.update(extra_attributes)
+            
+            # e.g. add a caption div
+            # if self.image.caption:
+            #     return mark_safe('<div><img{}><div>{}</div></div>'
+            #                      .format(flatatt(attrs), self.image.caption))
+
+            return super(TrustServeRendition, self).img_tag(extra_attributes)
+
         class Meta:
             unique_together = (
                 ('image', 'filter_spec', 'focal_point_key'),
@@ -63,6 +75,8 @@ Then set the ``WAGTAILIMAGES_IMAGE_MODEL`` setting to point to it:
 .. code-block:: python
 
     WAGTAILIMAGES_IMAGE_MODEL = 'images.CustomImage'
+
+Note that images inserted in the rich text editor will use the new rendition too.
 
 
 .. topic:: Migrating from the builtin image model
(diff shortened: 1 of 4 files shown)


SalahAdDin:
:+1:

wagtail/images/tests/test_rich_text.py
```diff
@@ -57,10 +59,12 @@ def test_expand_db_attributes_for_editor(self):
...
-        self.assertIn(
-            '<img data-embedtype="image" data-id="1" data-format="left" '
-            'data-alt="test-alt" class="richtext-image left"', result
-        )
+        self.assertIn('<img', result)
+        self.assertIn('data-embedtype="image"', result)
+        self.assertIn('data-id="1"', result)
+        self.assertIn('data-format="left"', result)
+        self.assertIn('data-alt="test-alt"', result)
+        self.assertIn('class="richtext-image left"', result)
```
gasman: We could use the [`assertTagInHTML` helper](https://github.com/wagtail/wagtail/blob/c85e4a3ff0022be405e75dbd6c5594d34d2cbb18/wagtail/tests/utils.py#L186) from wagtail.tests.utils as a neater way of doing these assertions (using `allow_extra_attrs=True` if we need to leave out `src` due to the unpredictable filename).

wagtail/images/tests/tests.py
```diff
@@ -183,10 +182,13 @@ def setUp(self):
...
-            'test filter spec'
+            'original'
         )
         # test image
-        self.image = MagicMock()
+        self.image = Image.objects.create(
+            title="Test image",
+            file=get_test_image_file(),
+        )
         self.image.id = 0
```
gasman: Looks like these tests are currently failing on database backends other than sqlite. It may well be that `self.image.id = 0` can simply be removed now...

docs/advanced_topics/images/custom_image_model.rst
```diff
@@ -41,6 +41,18 @@ Here's an example:
     class CustomRendition(AbstractRendition):
         image = models.ForeignKey(CustomImage, on_delete=models.CASCADE, related_name='renditions')
 
+        # Override img_tag from Abstract rendition to use your new fields
+        def img_tag(self, extra_attributes={}):
```
gasman: As per https://github.com/wagtail/wagtail/issues/4236#issuecomment-363483065, let's leave this undocumented.

wagtail/images/formats.py
```diff
@@ -15,30 +15,29 @@ def __init__(self, name, label, classnames, filter_spec):
...
+            'data-alt': escape(alt_text),
+        }
 
     def image_to_editor_html(self, image, alt_text):
         return self.image_to_html(
             image, alt_text, self.editor_attributes(image, alt_text)
         )
 
-    def image_to_html(self, image, alt_text, extra_attributes=''):
+    def image_to_html(self, image, alt_text, extra_attributes={}):
```
loicteixeira: Please don't use [mutable default arguments](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments). Do this instead:
```python
def img_tag(self, extra_attributes=None):
    if extra_attributes is None:
        extra_attributes = {}
```

gasman:
Merged in 99e9d5468afa2f5b9b0411aa1326ac73adf7883a - thanks!

PR #4246, opened 2018-02-06
Used Python 3 super() syntax also in documentation.

atombrella:
This disregards the release notes for earlier releases and only deals
with current documentation.


gasman:
Thanks again!

PR #4248, opened 2018-02-07
Add on_delete to BlogPageTag model

marykatefain:
While going through the tutorial I ran into this issue. Without the on_delete set on the BlogPageTag model the server crashes. I wasn't 100% sure which on_delete option to use, but at least this gets it up and running again!

### DIFF
--- a/docs/getting_started/tutorial.rst
+++ b/docs/getting_started/tutorial.rst
@@ -581,7 +581,11 @@ First, alter ``models.py`` once more:
 
 
     class BlogPageTag(TaggedItemBase):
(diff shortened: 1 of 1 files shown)


gasman:
Thanks @marykatefain! `on_delete=models.CASCADE` is indeed correct here, since we want the page's tagging information to get deleted whenever the page itself gets deleted.

(Incidentally, there's some current discussion at https://github.com/wagtail/django-modelcluster/issues/89 about making `on_delete` optional again here - but it absolutely makes sense to fix this in the meantime...)

Merged in 6c7d0abad71c049fefdab2eb8e13785c4b74db3e.

PR #4250, opened 2018-02-07
Add matrix chat badge to readme

alexgleason:
Hi everyone! I've been away from this project for a while, but I'm recently using it again for client work, so hopefully you'll be seeing more of me again. 😄 

Anyway, I know there is (or was) a Slack team for Wagtail, but I think Matrix has a number of advantages and it's becoming more popular among developers, especially for open source projects.

This PR adds a button to the README which links to a [Wagtail Matrix room](https://riot.im/app/#/room/#wagtail:matrix.org) I've created. A bit about Matrix:

1. Matrix **rooms can be viewed and joined from a public URL** and it does not require an invitation.
2. Matrix is a **decentralized, open source platform** with a strong community making it a smart long-term option when it comes to investing in a chat platform.
3. The matrix.org server is **free to use with no imposed limits** or quotas.
4. Matrix rooms can be **bridged to existing Slack channels** so members can choose their platform. You can also bridge with IRC, HipChat, and others.
5. The platform **respects the freedom and privacy of its users**, unlike Slack, so users who care about these things will not be deterred from entering a conversation. **I personally will not use Slack anymore** unless my life depends on it, so it's either email or this.

Some downsides of Matrix:

1. The matrix.org server occasionally hiccups or slows down. The situation has improved over the last six months, but they're not a for-profit company, they're a team of volunteers. However, there are [a number of Matrix servers](https://www.hello-matrix.net/public_servers.php) users can join; using matrix.org is optional. The more people spread out, the less of an issue this will be.
2. ??? That's all I could think of.

If anyone would like clarifications, or you have questions about Matrix, let me know.

Also, if people on the Wagtail team would like ownership over the room I've linked in the button (and the group it's attached to) let me know your Matrix ID and I will transfer ownership to you.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -6,6 +6,8 @@
     :target: https://pypi.python.org/pypi/wagtail/
 .. image:: http://codecov.io/github/wagtail/wagtail/coverage.svg?branch=master
     :target: http://codecov.io/github/wagtail/wagtail?branch=master
+.. image:: https://img.shields.io/badge/matrix-%23wagtail%3Amatrix.org-brightgreen.svg
+    :target: https://riot.im/app/#/room/#wagtail:matrix.org
 
 
 Wagtail CMS


thibaudcolas:
Thanks for the suggestion @alexgleason. I think this problem is bigger than your PR suggests – there are about ±250 people on the current Slack workspace, the #support channel gets usage on a daily basis, and if we were to consider it inappropriate I'd expect us to also consider other options (Discord, Spectrum, Gitter, etc) rather than just jump ship to Matrix.

I'm a bit skeptical about some of your arguments but there is one that stands out: "[...] respects the freedom and privacy of its users, unlike Slack". Especially the "unlike Slack" part – what do you mean by this?

alexgleason:
Matrix is an open platform, and all of the clients (desktop, Android, etc) are open source. I feel that the ability to inspect, modify, and redistribute software is a human right that is being violated by companies like Slack who require the use of proprietary applications for their proprietary platform. For one, if you can't review and change the source code, there's no telling what an application is actually doing, so it's safer to treat proprietary software like it's malware since you can't prove it's not. I don't think the burden of proof lies with me when it's really about the principle of the thing; open source alternatives exist.

Also, I think that people should have the ability to control their computers, which involves being to change the way programs they run work. To deny this is to alter the meaning of ownership itself, and I am very sad that most people have conceded the freedom to modify their computers and that it has become a social norm, especially during a time where governments are spying on activists and ad companies are targeting users in harmful ways.

thibaudcolas:
👍 thanks for the explanation.

alexgleason:
How about bridging the existing Slack rooms with Matrix rooms? I see no downside to that personally, except that it would require a few minutes of time. This would not disrupt the current user base while allowing new people to join via Matrix.

thibaudcolas:
That sounds really cool! Do you have some docs on how it works? The matrix.org homepage only references [this](https://matrix.org/docs/projects/as/slackbridge.html).

loicteixeira:
FWIW, this matter should be discussed at the core meeting this week.

gasman:
Having discussed this at last week's meeting, we're happy with the idea of setting up bridging between Slack and Matrix. Presumably this would be on a per-channel basis, rather than for the whole Slack instance? If so, I'd suggest that we start with the #general and #support channels and grow things from there as required. (We currently have around 30 channels, most of them inactive and created for things like working groups on sprints.)

Regarding the current PR, I think it's worth keeping mentions of Slack and Matrix in the same place - I'd suggest mentioning it on [the wiki page](https://github.com/wagtail/wagtail/wiki/Slack) and (possibly) the 'Community Support' section of the readme. If we're adding a badge, it probably makes sense to have one for Slack too.

alexgleason:
Cool, thanks!

@thibaudcolas The instructions are in the Riot client itself. From within the room in Riot you can add a bridge. It looks like this:

screenshot from 2018-02-21 11 57 44

It does have to be done on a per-room basis. Matrix also supports communities, though, which enable you to create a collection of rooms under one umbrella. I've created [+wagtail:matrix.org](https://riot.im/app/#/group/+wagtail:matrix.org) (you must be logged in to view) for this purpose. If we wanted to bridge multiple rooms, we'd just create them under the Wagtail community and bridge each room to the corresponding Slack channel.

I've granted Tom Dyson admin privileges in the Wagtail room. I've also invited him to the community and can make him an admin there once he accepts. If others would like these privileges let me know. I'll revoke my own admin privileges once we close this issue.

thibaudcolas:
👋 I’m reviewing old PRs and stumbled upon this. As stated above I don’t think we would want a badge for Matrix, but a mention alongside Slack seems appropriate.

I’ve added this mention on the Wiki page for Slack as @gasman suggests – this page also has Wagtail’s support policy, so is the one we generally link to when suggesting people get involved.

Please feel free to suggest better wording for the wiki / edit directly: https://github.com/wagtail/wagtail/wiki/Slack.

PR #4253, opened 2018-02-08
Comment Dockerfile, add docker-compose.yml

alexgleason:
First off, I changed the Dockerfile a little. Mostly I added comments. I also removed some lines:

```
ENV DJANGO_ENV dev
```
This doesn't appear to do anything. Also, I use a single Docker image for production and development and just switch between them by setting `DJANGO_SETTINGS_MODULE`, so it seems counterintuitive to declare this a development only file?

```
RUN python manage.py migrate
```
I don't think this command should be in the Dockerfile. The main reason being that it doesn't affect the built image at all, it affects the database. If you try building the image and no database is attached, the image will refuse to build because it errors out on this command. Instead, this should be run inside of the image after it's built with something like `docker run --rm wagtail python manage.py migrate`.

I also removed trailing slashes from `/code/` because they aren't needed (Docker docs doesn't use them) and they were used inconsistently.

I commented out the maintainer email because the image will build fine without it, and it seems better to build an image without a maintainer email than to build one with a placeholder email.

Next up, docker-compose. I added a file here that lets you simply `docker-compose up` to run your project locally. I did add `dj-database-url` and edit the settings. If you don't want `dj-database-url`, we could set each value distinctly, but this makes it easier and makes it Heroku-ready. The main benefit of all this is that the user can run postgres locally without a bunch of messing around.

### DIFF
--- a/wagtail/project_template/Dockerfile
+++ b/wagtail/project_template/Dockerfile
@@ -1,18 +1,22 @@
 FROM python:3.6
-LABEL maintainer="hello@wagtail.io"
+# Uncomment the line below and replace your email address
+# LABEL maintainer="your@email.com"
 
+# Enables manage.py output to be visible when using Docker
 ENV PYTHONUNBUFFERED 1
-ENV DJANGO_ENV dev
 
+# Install requirements
+# Copying in requirements.txt separately speeds up subsequent builds
 COPY ./requirements.txt /code/requirements.txt
 RUN pip install -r /code/requirements.txt
+# gunicorn is a deployment choice so it doesn't belong in requirements.txt
 RUN pip install gunicorn
 
-COPY . /code/
-WORKDIR /code/
+# Copy our project files into the image
+COPY . /code
+WORKDIR /code
 
-RUN python manage.py migrate
-
+# Run wagtail as a non-root user
 RUN useradd wagtail
 RUN chown -R wagtail /code
 USER wagtail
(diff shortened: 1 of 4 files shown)


tomdyson:
Thanks @alexgleason! Your changes to the Dockerfile all look good to me, apart from - perhaps - the removal of `RUN python manage.py migrate`. I agree with your argument, but the advantage of keeping it in is that it enables almost a single-step deploy from `wagtail start foo` to Dokku, Heroku, zeit.co/now. 

An alternative would be to include a `run.sh` with something like

```bash
#!/bin/sh
set -e

python manage.py migrate --noinput
gunicorn {{ project_name }}.wsgi:application --bind 0.0.0.0:8000 --workers 3
```

and to call `CMD "/code/run.sh"` from the end of the Dockerfile. I'm in favour of this change, even at the cost of adding a new file to the root of each new Wagtail build.

I like your docker-compose additions, but I doubt they'll be accepted by the core team. We've been careful to minimise any assumptions about developer or deployment environments in the project template, and I had to make a strong case for `Dockerfile` :) My recommendation would be to submit at PR to the Wagtail docs so users can understand and add these useful features as easily as possible.

thibaudcolas:
I’m going to close this now, because similar changes have been done as part of #5800, and the `docker-compose.yml` feels like for now it would be better done as documentation than as an additional file we have to maintain and aren’t really able to. I believe the Dockerfile changes suggeted here were taken into account when making #5800.

Alternatively we could open an issue to discuss the value in adding this – but I wouldn’t suggest doing a PR.

PR #4268, opened 2018-02-11
Added LineBreakHandler to resolve missing line breaks

### LINKED ISSUE #4266: RichText field is losing line breaks in 2.0b1
Upgraded from wagtail 2.0 alpha (no issue) to 2.0b1 and now when loading a richtext field that previously had line breaks the line breaks are missing. The initial editing/saving seems to work fine, but a subsequent edit causes the lose of the line breaks.

0. Python 3.6 and Django 2.0, Chrome 64
1. git checkout v2.0b1
2. npm install
3. npm run build
4. Start with empty richtext field on a page and save sample text to field:

The quick brown fox**⇧ + ↵**
jumps over the lazy dog.

5. Edit page again and now the field looks like this:

The quick brown foxjumps over the lazy dog.

6. Saving the page persists the missing line break(s). 

I'm not sure if this is a bug with draftail or wagtail, but there are no js errors in the console.

paperreduction:
Corrects issue where admin edit form was stripping line breaks from content. Fixes #4266

### DIFF
--- a/wagtail/admin/rich_text/converters/html_to_contentstate.py
+++ b/wagtail/admin/rich_text/converters/html_to_contentstate.py
@@ -215,11 +215,21 @@ class HorizontalRuleHandler(AtomicBlockEntityElementHandler):
         return Entity('HORIZONTAL_RULE', 'IMMUTABLE', {})
 
 
+class LineBreakHandler:
+    def handle_starttag(self, name, attrs, state, contentstate):
+        state.push()  # void tag
+        state.current_block.text += '\n'
+
+    def handle_endtag(self, name, state, contentstate):
+        state.pop()  # void tag
+
+
 class HtmlToContentStateHandler(HTMLParser):
     def __init__(self, features=()):
         self.paragraph_handler = BlockElementHandler('unstyled')
         self.element_handlers = HTMLRuleset({
-            'p': self.paragraph_handler
+            'p': self.paragraph_handler,
+            'br': LineBreakHandler(),
         })
         for feature in features:
             rule = feature_registry.get_converter_rule('contentstate', feature)
--- a/wagtail/admin/tests/test_contentstate.py
+++ b/wagtail/admin/tests/test_contentstate.py
@@ -436,6 +436,21 @@ class TestHtmlToContentState(TestCase):
             }
         })
 
+    def test_br_element_in_paragraph(self):
+        converter = ContentstateConverter(features=[])
+        result = json.loads(converter.from_database_format(
+            '''
+            <p>before<br/>after</p>
+            '''
+        ))
+        self.assertContentStateEqual(result, {
+            'entityMap': {},
+            'blocks': [
+                {'key': '00000', 'inlineStyleRanges': [], 'entityRanges': [], 'depth': 0, 'text': 'before\nafter',
+                 'type': 'unstyled'}
+            ],
+        })
+
     def test_block_element_in_empty_paragraph(self):
         converter = ContentstateConverter(features=['hr'])
         result = json.loads(converter.from_database_format(
--- a/wagtail/admin/wagtail_hooks.py
+++ b/wagtail/admin/wagtail_hooks.py
@@ -12,7 +12,7 @@ from wagtail.admin.rich_text import (
 from wagtail.admin.rich_text.converters.contentstate import link_entity
 from wagtail.admin.rich_text.converters.editor_html import LinkTypeRule, WhitelistRule
 from wagtail.admin.rich_text.converters.html_to_contentstate import (
-    BlockElementHandler, ExternalLinkElementHandler, HorizontalRuleHandler,
+    BlockElementHandler, ExternalLinkElementHandler, HorizontalRuleHandler, LineBreakHandler,
     InlineStyleElementHandler, ListElementHandler, ListItemElementHandler, PageLinkElementHandler)
 from wagtail.admin.search import SearchArea
 from wagtail.admin.utils import user_has_any_page_permission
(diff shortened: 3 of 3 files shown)


wagtail/admin/rich_text/converters/html_to_contentstate.py
```diff
@@ -215,11 +215,21 @@ def create_entity(self, name, attrs, state, contentstate):
...
+        state.pop()  # void tag
+
+
 class HtmlToContentStateHandler(HTMLParser):
     def __init__(self, features=()):
         self.paragraph_handler = BlockElementHandler('unstyled')
         self.element_handlers = HTMLRuleset({
-            'p': self.paragraph_handler
+            'p': self.paragraph_handler,
+            'br': LineBreakHandler(),
```
thibaudcolas: If we make this a configurable feature (like your changes to `wagtail_hooks.py` suggest), I don't think this should be in the default element handlers.

wagtail/admin/wagtail_hooks.py
```diff
@@ -269,6 +270,18 @@ def register_core_features(features):
         }
     })
 
+    features.register_editor_plugin(
+        'draftail', 'br', draftail_features.BooleanFeature('enableLineBreak')
+    )
```
thibaudcolas: 👍 there would also need to be some changes client-side for `enableLineBreak` to be processed [similarly to `enableHorizontalRule`](https://github.com/wagtail/wagtail/blob/6825aab5f764cb9accfd6af47b126e8e0f12f0d2/client/src/components/Draftail/index.js#L66), so its description text is translatable.

wagtail/admin/wagtail_hooks.py
```diff
@@ -269,6 +270,18 @@ def register_core_features(features):
...
+    features.register_converter_rule('contentstate', 'br', {
+        'from_database_format': {
+            'br': LineBreakHandler(),
+        },
+        'to_database_format': {
+            'entity_decorators': {'LINE_BREAK': lambda props: DOM.create_element('br')}
+        }
```
thibaudcolas: This `to_database_format` part probably doesn't do anything. In Draft.js line breaks are just `\n` in the text, not an entity, so [the exporter](https://github.com/springload/draftjs_exporter) processes them a specific mechanism.

You can see this here: https://github.com/wagtail/wagtail/blob/6825aab5f764cb9accfd6af47b126e8e0f12f0d2/wagtail/admin/rich_text/converters/contentstate.py#L52-L79

The `composite_decorators` part is what processes line breaks. You can see it's hard-coded, and there is no way to override it in the `for feature in self.features:` loop below. This PR needs to address that too.

Here is what the updated code might look like:

```py
        exporter_config = {
            'block_map': {
                'unstyled': 'p',
                'atomic': render_children,
                'fallback': block_fallback,
            },
            'style_map': {},
            'entity_decorators': {
                'FALLBACK': entity_fallback,
            },
            'composite_decorators': [],
            'engine': DOM.STRING,
        }

        for feature in self.features:
            rule = feature_registry.get_converter_rule('contentstate', feature)
            if rule is not None:
                feature_config = rule['to_database_format']
                exporter_config['block_map'].update(feature_config.get('block_map', {}))
                exporter_config['style_map'].update(feature_config.get('style_map', {}))
                exporter_config['entity_decorators'].update(feature_config.get('entity_decorators', {}))
                exporter_config['composite_decorators'] += feature_config.get('composite_decorators', [])
```

Then, here in `to_database_format`:

```py
        'to_database_format': {
            'composite_decorators': [
                {
                    'strategy': re.compile(r'\n'),
                    'component': br,
                },
            ],
        },
```

with the `br` definition moved here as well.

wagtail/admin/tests/test_contentstate.py
```diff
@@ -436,6 +436,21 @@ def test_block_element_in_paragraph(self):
...
+            '''
+        ))
+        self.assertContentStateEqual(result, {
+            'entityMap': {},
+            'blocks': [
+                {'key': '00000', 'inlineStyleRanges': [], 'entityRanges': [], 'depth': 0, 'text': 'before\nafter',
+                 'type': 'unstyled'}
+            ],
+        })
+
```
thibaudcolas: I believe we also need to test what happens if the `br` isn't inside a `p` (or any other block) tag. That will never happen with content coming from Draftail but I've seen [some weird things with Hallo.js and line breaks](https://github.com/thibaudcolas/draftjs_exporter_wagtaildbhtml#line-breaks), so this needs further consideration.

gasman:
Thanks @paperreduction!

I don't think we should make `br` a feature that's switchable through the 'features' list - at least, not for 2.0. That would be a breaking change, since any existing rich text fields that specify an explicit features list (and don't include `br` in that least, because it hasn't existed as a feature up to now) would suddenly stop allowing line breaks.

Handily, that makes this change a lot simpler - we just need to keep `LineBreakHandler` in `HtmlToContentStateHandler`'s initial `element_handlers` list, and don't need to go through `register_editor_plugin` or `register_converter_rule` (and thus avoid the need to hook everything up in the client-side code and the contentstate->html conversion as outlined in @thibaudcolas's comments). I'll go ahead and make that adjustment now.

gasman:
Merged in 54ad64e431cff78c07531514cec8b50ec8a14ba5 (master) / 4715cfc50168982b15917df23b271c92f8e41190 (2.0.x) - thanks again!

(Have added a test / fix for `<br />` elements encountered at the top level, which makes me realise we could do with some more graceful handling of inline styles and inline entities at the top level too. I'll get on to that tonight...)

paperreduction:
@gasman Okay great! That actually my preference as well. I was going to try to tackle the configurability of it over the weekend. Is that something you're interested in for a later release?

gasman:
@paperreduction Yep, definitely something we'd be interested in.

wagtail/admin/tests/test_contentstate.py
easherma-truth: I know this was old but this seemed like an odd test, since normally `<br>` is not written as `<br/>`

PR #4271, opened 2018-02-12
Update wagtail url paths for Django 2.0

marykatefain:
- `url()` is deprecated in Django 2.0 needs to be changed to `re_path()`
- show the necessary `django.url` imports for `path`, `re_path`, and `include` just in case the project does not already have them


gasman:
Thanks @marykatefain!

PR #4291, opened 2018-02-15
Add ability to search redirect to links

mwharrison:
Currently, when you search in redirects, it only returns items based on the old path. This adds a filter to the search for the redirected to link as well.

### DIFF
--- a/wagtail/contrib/redirects/tests.py
+++ b/wagtail/contrib/redirects/tests.py
@@ -291,6 +291,12 @@ class TestRedirectsIndexView(TestCase, WagtailTestUtils):
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.context['query_string'], "Hello")
 
+    def test_search_results(self):
+        models.Redirect.objects.create(old_path="/aaargh", redirect_link="http://torchbox.com/")
+        models.Redirect.objects.create(old_path="/torchbox", redirect_link="http://aaargh.com/")
+        response = self.get({'q': "aaargh"})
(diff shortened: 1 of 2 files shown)


loicteixeira:
Thanks for your contribution @mwharrison.

However if we are to include the redirected link to the search parameters, we should also include the url path of the redirected page (since you can create a redirect to a page or a link).

In addition, unless I'm mistaken, doing `filter(old_path__icontains=query_string, redirect_link__icontains=query_string)` will create an `AND` clause with `old_path` and `redirect_link` where we would want an `OR` clause.

mwharrison:
@loicteixeira - I've added the Wagtail page search code. 
Thanks for pointing out the `AND` query. I'm always forgetting the crazy `Q`. 😄 

If you think looks better, I'll attempt writing some tests for it.

loicteixeira:
It looks good to me like this @mwharrison. Let me know if you need a hand with the tests.

loicteixeira:
Thanks @mwharrison it looks good. I'll look at merging this soon.

What name should I use for the [contributors list](https://github.com/wagtail/wagtail/blob/master/CONTRIBUTORS.rst)? You're first name & last name as your on your GitHub profile page?

mwharrison:
@loicteixeira - Awesome! Yeah, you can use Michael Harrison on the contributor's list, thanks!

loicteixeira:
Merged in 4da0bd7bd2b8f15885c2678c21d369c5bbd56631. Thanks @mwharrison 🎉

Issue #3704, opened 2017-07-11
StreamField Block groups sometimes render bizarrely 


coredumperror:
I have a Page model called FlexPage, which is defined in a site-specific app. It uses a custom set of StreamField Blocks, in addition to a set of blocks defined in the "core" app which are shared with other apps.

The core app has this code block:
```
base_flexpage_widgets = [
    ('grid_layout', ColumnStreamBlock(group="Layout")),
    ('full_html', RawHTMLBlock(label='Full HTML', group="Text")),
    ('rich_text', RichTextBlock(label='Rich Text', group="Text")),
    ('image_panel', ImagePanelBlock(label='Image Panel', group="Images")),
    ('image_link', ImageLinkBlock(label='Image Link', group="Images")),
    ('captioned_image', CaptionedImageBlock(label='Captioned Image', group="Images")),
    ('news_listing', NewsListingBlock(label='News Listing', group="Listing Blocks")),
    ('spacer', SpacerBlock(label='Spacer', group="Layout")),
]
```

And the site-specific app uses this definition for its body field:
```
    body = StreamField([
        ('two_column_layout', TwoColumnBlock(group="Layout")),
        ('three_column_layout', ThreeColumnBlock(group="Layout")),
        ('four_column_layout', FourColumnBlock(group="Layout")),
        ] + base_flexpage_widgets +  # Placed here to preserve ordering in admin dashboard
        [('event_listing', EventBlock(label='Event Listing', group="Listing Blocks")),
        (
            'campus_announcements_listing',
            CampusAnnouncementsBlock(label='Campus Announcements Listing', group="Listing Blocks")
        ),
        ('section_title', SectionTitleBlock(label='Section Title', group="Text")),
        ('menu_listing', MenuListingBlock(label='Menu Listing', group="Listing Blocks")),
        ('image_slider', ImageCarouselBlock(label='Hero Image Slider', group="Images"))
    ])
```

I would expect there to be four Groups in the SteamField widget display: "Images", "Layout", "Listing Blocks", and "Text". 

Instead, I get this:
![screenshot](https://user-images.githubusercontent.com/413571/28095605-26a421ee-6658-11e7-933d-3683473ad04d.png)

Oddly enough, this only seems to affect the StreamField widget list when it's shown _after_ an existing Block in the form. The list groups itself correctly with an empty block list, or when you press the "+" button that's above the first block. But the listing is busted if you press the + button that's below any existing block.

gasman:
[...]

PR #4303, opened 2018-02-19
Use correctly sorted StreamBlock children in both prepend and append menus

timheap:
Previously blocks were only correctly sorted and grouped in the prepend menu, while they appeared in definition order in the append menu.


gasman:
Thanks @timheap! Presumably this is the cause of #3704 too.

timheap:
This is absolutely the cause of #3704. I spotted the problem myself when the interface bugged out in a similar manner.

gasman:
Merged in e8a9f9d8c924aff8659b38766ebdbe6d614adfc5

timheap:
Thanks!

PR #4304, opened 2018-02-19
Exclude the error messages from the count of inline panel elements

todd-dembrey:
When error messages are added to inline panels they are counted the same as an inline panel child element for purposes of calculating the maximum number of elements for that panel. 

This PR excludes `.error-message` from the count.

see: https://github.com/wagtail/wagtail/blob/5300b01fc4b797e355e1d08b6805b5ce5f8c45bd/wagtail/admin/templates/wagtailadmin/edit_handlers/inline_panel.html#L6-L18

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/js/page-editor.js
+++ b/wagtail/admin/static_src/wagtailadmin/js/page-editor.js
@@ -188,7 +188,7 @@ function InlinePanel(opts) {
 
     self.updateAddButtonState = function() {
         if (opts.maxForms) {
-            var forms = $('> li', self.formsUl).not('.deleted');
+            var forms = $('> li', self.formsUl).not('.deleted').not('.error-message');
             var addButton = $('#' + opts.formsetPrefix + '-ADD');
 
             if (forms.length >= opts.maxForms) {


wagtail/admin/static_src/wagtailadmin/js/page-editor.js
```diff
@@ -188,7 +188,7 @@ function InlinePanel(opts) {
 
     self.updateAddButtonState = function() {
         if (opts.maxForms) {
-            var forms = $('> li', self.formsUl).not('.deleted');
+            var forms = $('> li', self.formsUl).not('.deleted').not('.error-message');
```
thibaudcolas: Instead of adding a negative filter, would it be possible to change https://github.com/wagtail/wagtail/blob/5300b01fc4b797e355e1d08b6805b5ce5f8c45bd/wagtail/admin/templates/wagtailadmin/edit_handlers/inline_panel_child.html#L2 so we can change the initial selector?

For example, could we add a `data-inline-panel-child` attribute to the `<li>`, and do:

```js
var forms = $('> [data-inline-panel-child]', self.formsUl).not('.deleted');
// Or ideally – I'm not sure this is possible, I couldn't find where the deleted class comes from.
var forms = $('> [data-inline-panel-child]', self.formsUl);
```

todd-dembrey: Deleted gets added in the same file where i made my original change. It looks like it needs to remain unless you want to remove the `data-inline-panel-child` on delete?.

https://github.com/wagtail/wagtail/blob/5300b01fc4b797e355e1d08b6805b5ce5f8c45bd/wagtail/admin/static_src/wagtailadmin/js/page-editor.js#L112

There are actually other locations that need updating.

This looks at whether `.empty` should be added to the parent. If there is an error message i think `.empty` should not be added.
https://github.com/wagtail/wagtail/blob/5300b01fc4b797e355e1d08b6805b5ce5f8c45bd/wagtail/admin/static_src/wagtailadmin/js/page-editor.js#L95-L99

This one toggles ordering on the children, which should be updated to prevent it thinking the error message is the first orderable.
https://github.com/wagtail/wagtail/blob/5300b01fc4b797e355e1d08b6805b5ce5f8c45bd/wagtail/admin/static_src/wagtailadmin/js/page-editor.js#L181-L185

gasman:
Superseded by #4957

PR #4308, opened 2018-02-23
Fix Wagtail 2.0 Paths

loicteixeira:
While upgrading some plugins to Wagtail 2.0, I accidentally run the `wagtail updatemodulepaths` against my virtual environment and realised that some files were updated too.

Also for some reasons, `wagtail/wagtailsearch/tests/test_page_search.py` ended up not being moved to `wagtail/search`.

Edit: Also used `grep -RF --exclude-dir=releases wagtail.wagtail ./docs/` to check the documentation.


gasman:
Thanks @loicteixeira!

Issue #4296, opened 2018-02-16
Draftail remaining release blockers


BertrandBordage:
First of all, I’m very sorry to open this issue… I’m totally OK with releasing Draftail with some minor bugs since it’ll still be better than Hallo.js. The purpose of this issue is to focus only on critical, misleading and easily noticeable issues. These important issues must be fixed before 2.0, otherwise Draftail will get a ton of negative feedback and end users will remember Wagtail 2.0 as the release that changed their habits in a bad way (it’s even more important that Draftail is almost the only really noticeable change in 2.0 for end users).

# List of Draftail release blockers

Here is the current list of Draftail release blockers still present in Wagtail 2.0rc1:
- [ ] Deleting an image can delete the whole field
- [x] #4295  Draftail crash when rich text field is named "description"
- [x] The “Alt text” field should either be a modifiable input, or not an input at all. Right now it’s misleading everyone.
- [x] For some reason the “Alt text” input became huge and lost its border radius between 2.0b1 and 2.0rc1

Another very noticeable deprecated functionality is the image that no longer float in the text. While I’m strongly in favor of the Draftail behaviour, it is to be expected that a vast part of users will want to keep the old behaviour (I already got feedback on this).

# Deleting an image can delete the whole field

As listed below, I faced a case where the whole field disappeared when deleting an image. For some reason, the image was previously saved as outside of a paragraph, so the `body` field contains this in the database:
```html
<embed alt="Displaced_Persons_and_Refugees_in_Germany_BU6643.jpg" embedtype="image" format="right" id="503"/><p></p>
```
Note that the issue only occurs when you didn’t modify anything in that field since the page loaded. Adding content after the image, saving and trying to delete the image still does the same issue, even if you add an entire encyclopedia after the image.

## Let’s delete an image!
![capture d ecran 2018-02-16 18 15 19](https://user-images.githubusercontent.com/1119169/36320168-93032dfe-1345-11e8-8fe8-096fafa8dc56.png)
## No, you deleted the entire field.
![capture d ecran 2018-02-16 18 15 56](https://user-images.githubusercontent.com/1119169/36320169-9323e3fa-1345-11e8-8e62-747c6aedb9c5.png)
## Here is the corresponding JS error
![capture d ecran de 2018-02-16 18-16-04](https://user-images.githubusercontent.com/1119169/36320170-933e0b40-1345-11e8-876e-83e0c91b1116.png)

BertrandBordage:
[...]

PR #4311, opened 2018-02-24
Add spacer paragraphs around atomic blocks where necessary

gasman:
As discussed at https://github.com/wagtail/wagtail/issues/4296#issuecomment-367990405 - Draftail expects atomic blocks (such as images) to be surrounded by paragraphs for cursor interaction to work correctly. This PR updates the html->contentstate conversion logic to add spacer paragraphs where appropriate, so that content originating from hallo.js behaves correctly.

### DIFF
--- a/wagtail/admin/rich_text/converters/contentstate.py
+++ b/wagtail/admin/rich_text/converters/contentstate.py
@@ -82,10 +82,7 @@ class ContentstateConverter():
     def from_database_format(self, html):
         self.html_to_contentstate_handler.reset()
         self.html_to_contentstate_handler.feed(html)
-
-        if not self.html_to_contentstate_handler.contentstate.blocks:
-            # Draftail does not accept an empty block list as valid, but does accept 'null' as meaning "no content"
-            return 'null'
+        self.html_to_contentstate_handler.close()
 
         return self.html_to_contentstate_handler.contentstate.as_json(indent=4, separators=(',', ': '))
 
--- a/wagtail/admin/rich_text/converters/html_to_contentstate.py
+++ b/wagtail/admin/rich_text/converters/html_to_contentstate.py
@@ -20,10 +20,16 @@ class HandlerState:
         self.current_block = None
         self.current_inline_styles = []
         self.current_entity_ranges = []
+
         # what to do with leading whitespace on the next text node we encounter: strip, keep or force
         self.leading_whitespace = STRIP_WHITESPACE
         self.list_depth = 0
         self.list_item_type = None
+
+        # an atomic block which is NOT preceded by a non-atomic block must have a spacer
+        # paragraph inserted before it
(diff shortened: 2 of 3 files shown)


gasman:
Added another fix in a72196c12053baaea3d0175e5931f0ab1ddf6eca - HTML entities were being stripped on Python 3.4, because the HTML parsing defaults to `convert_charrefs=False` (which causes them to be handled outside the flow of normal text). Have bundled it here (even though it's not really related) because it seems to rely on us calling `close()` on the parser, which I wasn't doing before this fix...

gasman:
Thanks @loicteixeira! Have fixed the issue with lists in 072cd038360bc77091e98dac9266da6d74b7892f - this was due to some incorrect logic with pushing / popping state. (`has_preceding_nonatomic_block` shouldn't be included in the pushed state, because it means we end up treating the block before the list as the 'previous' one.)

Regarding the name, I think "preceding" is more correct than "current", because we check it when we encounter the image (or other atomic block) to see if the block above it is one that requires a spacer image. At that point, the relevant block has _usually_ been closed already, and `current_block` won't work because we set it back to `None` in `handle_endtag`. I tried moving the `state.has_preceding_nonatomic_block = True` line to `handle_endtag` to make the intent clearer, but then it gives the wrong result for `<p></p>` or `<p>some text here </p>` (where the image forces the start of a new block, and needs to treat the `<p>` as the preceding one).

loicteixeira:
That looks good 👌 

I did some quick testing by removing the empty paragraphs around images, embed and HRs in the database (like a butcher `UPDATE wagtailcore_pagerevision SET content_json = REPLACE(content_json, '<p></p>', '');`) and it worked as intended. Empty paragraphs where inserted as follow:

```diff
# For images:
+   Empty paragraph
    Image
    Paragraph with text
    Image
+   Empty paragraph
    Image
    H2
    Image
    UL
    Image
    OL
    Image
+   Empty paragraph

# For Embeds:
+   Empty paragraph
    Embed
    Paragraph with text
    Embed
+   Empty paragraph
    Embed
    H2
    Embed
    UL
    Embed
    OL
    Embed
+   Empty paragraph

# For HRs:
+   Empty paragraph
    HR
    Paragraph with text
    HR
+   Empty paragraph
    HR
    H2
    HR
    UL
    HR
    OL
    HR
+   Empty paragraph

# Mix
+   Empty paragraph
    HR
+   Empty paragraph
    Embed
+   Empty paragraph
    HR
+   Empty paragraph
    HR
+   Empty paragraph
    Image
+   Empty paragraph
    HR
+   Empty paragraph    
    Embed
+   Empty paragraph
    Image
+   Empty paragraph
    Embed
+   Empty paragraph
```

As for the naming, I'm not 100% convinced but I see where you're coming from. In any case, it's not worth holding anything for that reason. We're good to go 🎉

PR #4312, opened 2018-02-26
Run Django password validators against populated user models

timheap:
Previously the validators were run against empty user models, which made the UserAttributeSimilarityValidator accept passwords that would otherwise have been rejected.

Tests have been added to ensure the validator run correctly.


gasman:
Thanks @timheap! Merged in 9d8563a7442ba1a97ef114c96efec9e0ca248ddc (master) / f26491bcdd06057853271bc9d011150db24a786e (2.0.x).

PR #4314, opened 2018-02-26
Prevents accounts leak through the password reset form.

### LINKED ISSUE #4191: PasswordResetForm leaks information about registered users
### Issue Summary

The PasswordResetForm returns validation errors about whether the account exists in wagtail or not and some information about whether wagtail manages the password. 

e.g:
"This email address is not recognised."
"Sorry, you cannot reset your password here as your user account is managed by another server."

https://github.com/wagtail/wagtail/blob/master/wagtail/admin/forms.py#L87

The standard django behaviour is to not leak this information. It gets all users with the supplied email and sends them the relevant information. If the user supplies an invalid email they progress to the success screen but no email is sent.

https://github.com/django/django/blob/master/django/contrib/auth/forms.py#L282:28

### Proposed Fix

Change the behaviour of the password reset to inform the user that an email has been sent to the supplied address regardless of whether the email is registered with the site.

Send an email to accounts that do exists. and within the email include any validation messages within that email e.g. who manages the password.

BertrandBordage:
This pull request fixes #4191.

It also displays an error on invalid e-mail submit.

Note that the message `Sorry, you cannot reset your password here as your user account is managed by another server.` is no longer displayed if the site has no user with a usable password. This is normal for two reasons:
- [as mentioned in the docs](http://docs.wagtail.io/en/v1.13.1/advanced_topics/settings.html#password-management), developers should disable password management when passwords are handled externally
- assuming that a site with unusable passwords only handles password externally is logically wrong and can mislead beginner devs

### DIFF
--- a/wagtail/admin/forms.py
+++ b/wagtail/admin/forms.py
@@ -2,7 +2,6 @@ import copy
 from itertools import groupby
 
 from django import forms
-from django.contrib.auth import get_user_model
 from django.contrib.auth.forms import AuthenticationForm, PasswordResetForm
 from django.contrib.auth.models import Group, Permission
 from django.core import validators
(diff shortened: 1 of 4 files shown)


gasman:
Thanks @BertrandBordage! Merged in 98ea7bbf3fb565e2024f6e1d1ab6b476c74e1d2d (master) and d3d3f20ec8877b3258d2544a7f978bb4a8f7515e (2.0.x).

PR #4315, opened 2018-02-26
[WIP] Upgrade to latest Draftail, perf, a11y, and extension APIs

thibaudcolas:
Addresses the image block removal with the "Remove" button from #4296 by updating Draftail.

The new version also:

- Performs a bit better when typing in the editor.
- Addresses some accessibility issues with the editor.

While at it I also added new Draftail extension APIs. Wagtail doesn't use them but they will be useful for future extensions / extensions supporting Hallo and wishing to support Draftail. Edit: they aren't hooked up to the rich text features at the moment – not sure whether they should be, or how exactly.

~I'd like to add a few more things before this is merged:~

* ~[ ] Use the new `ariaDescribedBy` editor prop to point at the field's label, if possible.~
* ~[ ] Add error handling, and an error recovery UI, to the editor.~

### DIFF
--- a/client/src/components/Draftail/index.js
+++ b/client/src/components/Draftail/index.js
@@ -17,13 +17,25 @@ export { default as ModalWorkflowSource } from './sources/ModalWorkflowSource';
 const BR_ICON = 'M.436 633.471l296.897-296.898v241.823h616.586V94.117h109.517v593.796H297.333v242.456z';
 
 /**
- * Registry for client-side code of Draftail plugins.
+ * Registry for client-side code of Draftail extensions.
  */
 const PLUGINS = {};
+const DECORATORS = [];
+const CONTROLS = [];
 
 const registerPlugin = (plugin) => {
   PLUGINS[plugin.type] = plugin;
   return PLUGINS;
+};
+
+const registerDecorator = (decorator) => {
+  DECORATORS.push(decorator);
+  return DECORATORS;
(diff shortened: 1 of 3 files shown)


gasman:
Thanks again for your work on this, @thibaudcolas!

I think we should include the Draftail version upgrade in Wagtail 2.0, and leave the other updates for 2.1. The RC1 release was a kind of 'last call' for critical bugfixes, and if we consider non-essential enhancements at this point, we'd really have to put out a second release candidate and delay things further.

It's tempting to hold on for the improved error handling, but we've now accounted for the known failure cases (image deletion and the `description` naming clash) so I don't think this is essential - any further issues that come up can reasonably be handled in a 2.0.1 release if necessary. We've had to make some difficult decisions about what to leave out of 2.0, and I think now's the time to take a similarly hard line on Draftail and get us out of this "release limbo" :-)

thibaudcolas:
@gasman 👍. This is mergeable as-is, with the error case addressed. You know best what is and isn't reasonable when it comes to releases so I'll let you be the judge, and add the error handling in a separate PR.

gasman:
👍 Have cherry-picked 8648d6d451be0a071fb5013d7a714beac7097c43 into 2.0 and master - will save the rest for 2.1.

thibaudcolas:
Superseded by #4436

PR #4318, opened 2018-02-27
Add error handling component to Draftail

thibaudcolas:
This adds a "fallback" UI when the editor crashes. Builds upon #4315 (and contains some of its commits).

When the editor crashes (eg. because of a bug in Draft.js or Draftail, or in some browser API), React will remove all of the parts of the UI that are affected by the error. This is good because users won't be interacting with a broken UI, but it's even better if we provide some sort of fallback that explains what happens.

Here is what it looks like:

draftail-fallback-ui

Here is what happens when it crashes:

draftail-fallback

The "latest" content is available so users don't lose what they were typing. It's displayed as plain text in a `textarea` to reduce the risk of rendering errors. Behind the scenes, the `input` value is reset to what it was at page load, reducing the risk that error-prone / broken content will be saved.

WIP:

- [ ] Feedback on the UI and messaging would be good.
- [x] Tests are missing.

To cause a crash in the editor, put the following in the browser console (tweak the selector to pick a different editor):

```js
document.querySelector('[data-draftail-input]').draftailEditor.setState({ editorState: {}})
```

### DIFF
--- a/client/src/components/Draftail/Draftail.scss
+++ b/client/src/components/Draftail/Draftail.scss
@@ -17,6 +17,7 @@ $draftail-editor-font-family: $font-serif;
 @import '../../../../node_modules/draftail/lib/index';
 
 @import './Tooltip/Tooltip';
+@import './EditorFallback/EditorFallback';
 
 @import './decorators/TooltipEntity';
 
--- a/client/src/components/Draftail/EditorFallback/EditorFallback.js
+++ b/client/src/components/Draftail/EditorFallback/EditorFallback.js
@@ -4,6 +4,8 @@ import { convertFromRaw } from 'draft-js';
 
 import { STRINGS } from '../../../config/wagtailConfig';
 
+const MAX_EDITOR_RELOADS = 3;
+
 class EditorFallback extends PureComponent {
   constructor(props) {
     super(props);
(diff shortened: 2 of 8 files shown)


thibaudcolas:
@gasman this is working as-is. It's missing tests though, and it would be nice to get some feedback on the UI, but arguably it's already an improvement to the current behavior (field disappears).

I'll add the tests this evening if you want to wait for that.

Latest UI:

draftail-error-fallback

thibaudcolas:
Thanks for looking at this @BertrandBordage. I'll have a look at the `.full` layout issue.

For the "reload the editor" resetting the content, this is intended. I tried to make this clear with the error message saying "Content has been reset to the last saved version".

The point is that _if_ the error is due to the content, the only way to not have the editor crash again is to go back to a "safe" point. The saved content can generally be considered safer than what was just typed, if there was a crash.

I'm not entirely sure how frequently those crashes would be because of the content – my instinct tells me it could be a fair percentage, but it's hard to think quantitatively for something that's already addressing corner cases. One example that comes to mind is if people make buggy extensions for Draftail – say a better embed or image block. The images / embeds will be inserted into the content, and then the editor would crash on render, and again whenever reloaded.

Perhaps this is a matter of changing the label – say "Show latest content", and "Reload saved version" or "Reload saved content"?

BertrandBordage:
I see.
Yes, maybe changing the label would be better. Note that I probably didn’t understand the same thing as most editors will: by saved, I understood “saved by Javascript”, not “saved in the database”. So maybe it was only misleading for me, as a developer aware of the JS lying behind the scenes.

thibaudcolas:
@BertrandBordage I've addressed the layout issue on `.full` fields and also changed the button label slightly, this should be good for review again. Here's the latest WIP:

draftail-error-fallback-wip

gasman:
Merged in 7238f22ebd0af06d58a32e4253c240fb196be76a (master) / 714797ba15d7c76254d046a17d94b4710788e37a (stable/2.0.x)

PR #4322, opened 2018-02-28
Pass index name in URL to Bulk API

kaedroho:
We currently index all items in Elasticsearch using the root bulk API (at ``/_bulk``). This API is to allow inserting into multiple indices at once. However, Wagtail inserts into one index at a time so this is not needed.

If we pass the index name as a parameter in the call to ``bulk()``, the index-specific bulk API will be used instead (at ``/<index name>/_bulk``.

The advantage of this change is it makes it possible to implement access control by checking the URL an application is using. This is required in order for the Bulk API to work on certain hosts (such as Divio).

### DIFF
--- a/wagtail/search/backends/elasticsearch2.py
+++ b/wagtail/search/backends/elasticsearch2.py
@@ -806,7 +806,6 @@ class Elasticsearch2Index:
         for item in items:
             # Create the action
             action = {
-                '_index': self.name,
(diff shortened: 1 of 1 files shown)


tm-kn:
I've tested this fix on 1.13 and works good on both Divio AWS ElasticSearch and my local instance.

PR #4324, opened 2018-03-01
Use specific page url for view live link

lambertkuang:
Use specific url for  'View live' PageListingButtons - subclassed Pages weren't using the correct live url


gasman:
Hi @lambertkuang - thanks for the PR!

Can you clarify exactly where the incorrect URL is appearing, please? As far as I'm aware, the index view already handles fetching the subclassed pages (with the exception of the 'reorder' view, where it's disabled for performance reasons):
https://github.com/wagtail/wagtail/blob/master/wagtail/admin/views/pages.py#L81-L85

lambertkuang:
Ah right, I'm on an older version of wagtail that wasn't yet using the most specific pages on search - closing this PR 👍

Issue #4305, opened 2018-02-19
Perm checking consistency


sebrestin:
### Issue Summary

page_listing_more_buttons hook uses both page and page_perms to check for permissions. To keep it consistent I suggest only page_perms.page_is_root should be used.

### Technical details

* Python version: 3.6.6.
* Django version: 1.11.
* Wagtail version: 1.13.1.
* Browser version: Firefox 58.

gasman:
Fixed in #4326

PR #4326, opened 2018-03-01
Permission checking consistency #4305

sebrestin:
I encountered this while using wagtail and I think this will be useful, if someday you would like to make PagePermissionTester swappable.

### DIFF
--- a/wagtail/admin/wagtail_hooks.py
+++ b/wagtail/admin/wagtail_hooks.py
@@ -153,7 +153,7 @@ def page_listing_more_buttons(page, page_perms, is_parent=False):
             attrs={"title": _("Move page '{title}'").format(title=page.get_admin_display_title())},
             priority=10
         )
-    if not page_perms.page_is_root:
+    if page_perms.can_copy():
         yield Button(
             _('Copy'),
             reverse('wagtailadmin_pages:copy', args=[page.id]),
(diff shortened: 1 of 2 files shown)


gasman:
Hi @sebrestin - thanks for the PR!

This has come up a couple of times before (for example #2907), and unfortunately I've reached the conclusion that we can't realistically support users building their own permission backends with arbitrary customisations - it makes it impossible for us to reason about the permission logic for new features that we implement in future.

In this case, overriding the behaviour of `page_perms.page_is_root` means that you're effectively changing the definition of what it means for a page to be "the root node": there's no way for us to ensure that other parts of Wagtail (or new features developed in future) behave in a way that's consistent with that new definition.

sebrestin:
I understand that you do not plan to make it swappable, but why use page.is_root() and not page_perms everywhere, thus respecting an interface that you created internally?

gasman:
I see what you mean! We call `page_perms` to decide the visibility of the other buttons, so it makes sense for consistency to do the same here, even if the permission rule is trivial ("anyone can perform this action on any page except for root").

I don't think we should be referring to `page_perms.page_is_root` directly here, though - the `page_is_root` flag is part of the implementation of `PagePermissionTester` rather than the interface, and if we access it directly then we're not really delegating the permission logic to `PagePermissionTester`, just asking it for the answer to a question we already know... I think it would be better to add new methods `can_copy` and `can_view_revisions` to `PagePermissionTester`, which just return `not self.page_is_root`, and call those from the button code. Does that sound OK?

sebrestin:
That makes more sense. I will update the pull request.

gasman:
Thanks @sebrestin! Merged in e8ee632b296401c9ee919535ac9e0e7c4aade9a0.

PR #4332, opened 2018-03-02
Allow draftail to initialise when features are specified in the settings

todd-dembrey:
Hallo.js pops the options from the kwargs. See:
https://github.com/wagtail/wagtail/blob/76975525f7a3757b280572165f8717cc4c13bba6/wagtail/admin/rich_text/editors/hallo.py#L87

Draftail needs to do something similar (even if not using the options) in order to prevent `__init__() got an unexpected keyword argument 'options'` when calling the super classes.


thibaudcolas:
@loicteixeira @gasman any idea what's best done here?

gasman:
This looks like the right fix to me - thanks @todd-dembrey! Merged in 09f8a4f38a95f2760f38ab2f142443df93b5d8c6.

Issue #4329, opened 2018-03-01
Update datepicker to 2.5.18 to fix CSP errors due to eval()


Pomax:
Would it be possible to switch from the current jquery datepicker for page live/expiry datetime to a non-jquery library that doesn't rely on arbitray code execution? Right now wagtail requires running with `unsafe-eval` turned on for script CSP, which for production websites is an incredibly dangerous setting to have to enable.

This affects both 1.* and the just released 2.0

thibaudcolas:
Hey @Pomax, I see you've commented on #1288 as well – do you think the datepicker in particular is worth a separate issue? As far as I can see removing `unsafe-eval` would also require some work on the modal choosers for links/images/embeds/docs as well:

https://github.com/wagtail/wagtail/blob/e1325ed4d6c9e53a636d9a4fd906a6d95eefea88/wagtail/admin/static_src/wagtailadmin/js/modal-workflow.js#L54-L58

(and also on [this](https://github.com/wagtail/wagtail/blob/c7a418b1c360204ab8393386a2dd72312e21993a/wagtail/admin/static_src/wagtailadmin/js/vendor/jquery.iframe-transport.js#L209), which I don't even know what it's used for)

-----

I think we would be keen to replace the datepicker with one that's more CSP friendly if someone takes on the challenge. Same for the chooser modals, although this is more likely to happen for other reasons than just CSP.

If you can't wait for those things to happen, an alternative would be to serve the site’s back-end from a different origin than the front-end. This way you can have different CSPs for both.

gasman:
Note that there's been quite a lot of work (e.g. #2983) invested in making the current date picker work correctly with Django date formats, including custom / localised ones, so it's probably not going to be as simple as dropping in a new datepicker widget.

In fact, it looks like the offending `eval`s are part of a sub-library that deals with date string formatting and parsing - I'd be happier if we could find (or make) a direct functional replacement for that one component, as that'll give us more confidence that we're not introducing any regressions.

Pomax:
Maybe, but I just dropped in 2.5.18 rather than 2.3.8, which does not rely on `eval()`, and I am no longer seeing CSP issues but do see a perfectly working calendar widget for page publish/expiry, so it's probably worth simply doing a version bump for it and have that automatically solve this particular CSP issue.

thibaudcolas:
@Pomax that sounds great! Could you make a PR for this, highlighting the changelog in that version range?

Pomax:
[...]

PR #4337, opened 2018-03-06
Update jquery-datetimepicker to 2.5.18

Pomax:
- Fixes an `eval()` CSP violation
- includes touch scroll fix from commit 87a7a79 (L909)

closes https://github.com/wagtail/wagtail/issues/4329

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -12,6 +12,7 @@ Changelog
  * Remove outdated X-UA-Compatible meta from admin template  (Thibaud Colas)
  * Add JavaScript source maps in production build for packaged Wagtail (Thibaud Colas)
  * Removed `assert` statements from Wagtail API (Kim Chee Leong)
+ * Update `jquery-datetimepicker` dependency to make Wagtail more CSP-friendly (`unsafe-eval`) (Mike Kamermans)
  * Fix: Status button on 'edit page' now links to the correct URL when live and draft slug differ (LB (Ben Johnston))
  * Fix: Image title text in the gallery and in the chooser now wraps for long filenames (LB (Ben Johnston), Luiz Boaretto)
  * Fix: Move image editor action buttons to the bottom of the form on mobile (Julian Gallo)
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -286,6 +286,7 @@ Contributors
 * Dan Swain
 * Alexs Mathilda
 * Tony Yates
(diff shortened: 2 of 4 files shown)


Pomax:
Can you explain what you're looking for in terms of what else is included? This bumps up the datetimepicker from 2.3.8 to 2.5.18 (so a minor version bump, which should not impact the API for the datetimepicker, only the stability/internals - the changelog between those two versions [is rather large](https://github.com/xdan/datetimepicker/compare/2.3.8...2.5.18)), with the patch that @gasman pointed out needed to be preserved.

Browsers tested: Chrome 64, Firefox Nightly (60.0a1), Edge 41, Vivaldi 1.14 (I do not have access to the full list of browsers the dev docs suggests to test with). All browsers tested on Windows 10 Pro x64.

I mainly tested by clicking "edit" for a variety of `class CustomPageType(Page): ...` implementations and hitting up the "settings" tab to edit the publish/expiry datetimes for a wagtail instance we're trying to get set up at the Mozilla Foundation.

thibaudcolas:
👌 

I see the library has no explicit CHANGELOG, which would be what I'm after in order to know what to look out for when testing. It's a minor version bump but I don't think this tells much – there's still been 3+ years and 20+ versions.

No worries for the browsers you couldn't test on, we'll get through this in further reviews/tests.

Pomax:
Yeah fair - I'll see about reproducing those errors and figure out if they're easily addressed; are any of these things you caught things that can be automated? (mocha/selenium testing for instance?)

Pomax:
Looks like datepicker itself has some silly bugs around touch vs mouse events. I've filed https://github.com/xdan/datetimepicker/pull/654/files to get those resolved.

Pomax:
@thibaudcolas  My PR to datetimepicker landed, and 2.5.19 got released, which should fix all those weird issues by no longer throwing and breaking UX =)

I've updated the PR to use that version instead now if you'd like to re-review.

thibaudcolas:
@Pomax 👏 thank you for taking the time to land a PR there. I'm sure it wasn't easy, it's much appreciated. I'll get on with the review.

thibaudcolas:
It's in 🙂 scheduled for release in Wagtail 2.1

Thank you @Pomax!

Pomax:
awesome, many thanks! I also have another PR up for tackling the same issue but in modal-workflow, replacing the `eval()` call there a different way to effect the same thing, hopefully that can make it into 2.1 too =)

PR #4341, opened 2018-03-08
Add favicon to documentation

### LINKED ISSUE #4340: Please set up the Wagtail logo to appear as the favicon in readthedocs.
When I have multiple readthedocs tabs open side-by-side, I can't tell which tab is for which set of documentation.  [Here is an example of  a site that has done this](http://linuxmint-installation-guide.readthedocs.io/en/latest/).

chosak:
This change copies the Wagtail admin [favicon.ico](https://github.com/wagtail/wagtail/blob/master/wagtail/admin/static_src/wagtailadmin/images/favicon.ico) into the Sphinx build directory so that built docs use the same favicon as the admin.

Makes use of [the Sphinx `html_favicon` setting](http://www.sphinx-doc.org/en/1.5.1/config.html#confval-html_favicon) which adds this line to the generated HTML:

```html
<link rel="shortcut icon" href="_static/favicon.ico"/>
```

Fixes #4340.

Favicon looks like this in Chrome:

favicon-in-chrome


gasman:
Thanks @chosak - looks good to me!

DanAtShenTech:
Fantastic turnaround time!

DanAtShenTech:
@gasman I just noticed one other thing - the favicon is only showing in the latest version of the docs.  It would be very helpful if it could also be set to appear in the 2.0 docs and the 1.13 docs.

gasman:
@DanAtShenTech Unfortunately readthedocs makes it pretty hard to retrospectively change existing versions of the documentation, as they're pinned to git tags. There will be a 2.0.1 release soon, so the icon will be updated when that happens (and most likely a new 1.13.x and 1.12.x version, assuming I remember to merge this fix into those branches...)

DanAtShenTech:
Sounds good @gasman.  Again, I really appreciate the quick turnaround on little requests like this, and I'm sure looking forward to using Wagtail more.

PR #4351, opened 2018-03-10
Add ChooserBlock.get_queryset()

mvantellingen:
By creating the queryset to be used in to_python() and bulk_to_python()
it is more easy to override it in a subclass (for example adding
.only() or a .select_related())

### DIFF
--- a/wagtail/core/blocks/field_block.py
+++ b/wagtail/core/blocks/field_block.py
@@ -545,7 +545,7 @@ class ChooserBlock(FieldBlock):
             return value
         else:
             try:
-                return self.target_model.objects.get(pk=value)
+                return self.get_queryset().get(pk=value)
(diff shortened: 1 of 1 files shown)


gasman:
Hi @mvantellingen - please see #4157, which proposed the same idea :-)

ChooserBlock was never designed to be used directly with a Select widget, and I don't think it makes sense to go down the road of extending that functionality until we've confirmed that it actually works properly and has been fully tested and documented. Currently I believe that's blocked by #3169, and possibly #4380 (which is next on my list to review...)

mikedingjan:
@gasman I think using the ``.get_queryset()`` from the manager is better, instead of ``.objects.all()`` since that allows you to handle the select_related etc. in your custom manager. For example django-oscar uses this for products.

gasman:

Issue #4353, opened 2018-03-10
breadcrumbs don't honor overridden get_admin_display_title's


Speedy1991:
Defining a _get_admin_display_title_ doesn't get honored by breadcrumbs.
The error is here:
https://github.com/wagtail/wagtail/blob/master/wagtail/admin/templates/wagtailadmin/shared/breadcrumb.html#L12
`<li><a href="{% url 'wagtailadmin_explore' page.id %}">{{ page.get_admin_display_title }}</a></li>`

Fix (note the specific):
`<li><a href="{% url 'wagtailadmin_explore' page.id %}">{{ page.specific.get_admin_display_title }}</a></li>`

This is useful if you have multiple translated pages with same title and want to show the language code too.

E.g.: 

```
def get_admin_display_title(self):
    return "%s [%s]" % (self.title, self.slug)
```

> Awsome [EN] > Site1
> Awsome [DE] > Site1

At the moment i only receive

> Awsome > Site1
> Awsome > Site1

PR #4354, opened 2018-03-10
Fix breadcrumbs don't honor overridden get_admin_display_title's

Speedy1991:
#4353


Speedy1991:
Added same fixes to other templates

gasman:
As per https://github.com/wagtail/wagtail/issues/4396#issuecomment-373922536, I'm not sure that the extra fixes on the other templates are needed.

gasman:
Thanks @Speedy1991! Have gone with a slightly revised approach in #4415. (The reorder view listing is being addressed separately in #4410.)

PR #4355, opened 2018-03-12
Remove IE6-9 conditional comments from project template

solarissmoke:
I'd like to propose removing the shims for IE6-9 from the project template.

The rationale for this is the same as for dropping support for EOL version of Django, etc. - i.e., these versions of Internet Explorer are no longer supported by the vendor (the last version of Windows that you cannot upgrade to use IE10 or higher was Windows XP, which reached end of life in 2014). The [Wagtail documentation itself](http://docs.wagtail.io/en/v2.0/contributing/developing.html) says that IE10 and lower are not supported.

On that basis I think that Wagtail should not include these shims by default - projects that for whatever reason still need to support these browsers can easily add them to their templates.


thibaudcolas:
Thank you @solarissmoke, that seems very sensible to me. I also removed the `<meta http-equiv="X-UA-Compatible" content="IE=edge" />`, which is useless for recent versions of IE, and MS Edge. And reflected those changes on the separate `500.html` template.

PR #4356, opened 2018-03-12
Remove outdated X-UA-Compatible meta from admin template

thibaudcolas:
This meta is useless since Wagtail dropped support for IE8-9-10. Starting with IE11, edge mode is the default. The chrome=1 part is for Chrome Frame, which was only available until 2014, for IE6-9.

See also #4355.


jjanssen:
Thanks, merged with 68daec7

PR #4357, opened 2018-03-12
Rewrite modal-workflow.js to not rely on eval()

### LINKED ISSUE #4335: CSP: update modal-chooser so that it loads js source from URL using <script> rather than using eval()
the modal-chooser currently uses a mechanism that loads modal HTML+JS from a single URL as json response, then extracts the HTML as raw string and injects it into the page, and then extracts the JS as raw string and turns it into executable code using `eval()`.

The first part might be dangerous, because it *might* lead to injecting user-generated html (note: I'm not sure, didn't look deeply enough into it), but the second part violates CSP something fierce, so seeing if it's possible to rework this a little to load the associated javascript as a dedicated URL output using a `<script>` tag so that sensible CSP settings don't break wagtail would be a good idea.

I'm having a quick look at this focussing specifically on the images/chooser use case:
- leave the images/chooser route as is
- add an images/chooser/js route for exclusively serving the JS content with the correct mime-type
- keep the modal-chooser as extracting and injecting HTML 
- change the modal-chooser to load the JS by instead creating a `<script src="..."></script>` and adding that to the document `<head>`

That might still trip up CSP, but hopefully it doesn't.

Pomax:
This rewrite switches the `eval()` call used by the modal workflow (which will get blocked by CSP that do not specify `unsafe-eval`) to a DOM script insertion (which does not trigger CSP violation errors). It introduces a helper function that takes Wagtail's jsoneque html+js text response and performs minor rewrites to it to turn it into valid JSON, so that it can be run through `JSON.parse` rather than `eval()`.

This fix probably closes #4335 (it fixes the CSP error without being as invasive as the approached suggested early on in that issue).

> Do the tests still pass? (http://docs.wagtail.io/en/latest/contributing/developing.html#testing)

as far as I can tell, this code does not change the result of `python manage.py test`.

> Does the code comply with the style guide? (Run `make lint` from the Wagtail root)

as far as lint:js is concerned: yes.

> For front-end changes: Did you test on all of Wagtail’s supported browsers?

Tested against Chrome 65, Firefox nightly (61), and Edge (v41). I made sure to use only ES5 syntax so that things still work even on drastically older browsers.

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/js/modal-workflow.js
+++ b/wagtail/admin/static_src/wagtailadmin/js/modal-workflow.js
@@ -118,8 +118,11 @@ function formJSON(data) {
 
     if (data.indexOf(onloadKey) > -1) {
         var parts = data.split(onloadKey);
-        var sourceCode = parts[1].replace(/\s*\r?\n/g, '\\n').replace(/"/g, '\\"') + 'END_OF_RESPONSE';
-        sourceCode = '"' + sourceCode.replace(/}[\s\r\n]*END_OF_RESPONSE/, '"}');
+        var sourceCode = parts[1]
+        // first, turn this into JSON-valid string data
+        sourceCode = JSON.stringify(sourceCode)  + 'END_OF_RESPONSE';
+        // then, move the quote back to befor the curly bracket.
+        sourceCode = sourceCode.replace(/}"[\s\r\n]*END_OF_RESPONSE/, '"}');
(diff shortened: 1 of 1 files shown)


gasman:
Thanks @Pomax! Have now merged #4583, which completes the refactor of modal-workflow to avoid dynamic JS evaluation entirely.

Pomax:
awesome!

PR #4358, opened 2018-03-12
allow chunk_size to be set in update_index

dibell:
With certain document sets the update_index command times out. Setting the chunk_size to lower than the default 1000 fixes this issue.

This change allows the chunk_size to be set as a parameter to the command.

The current tests cover this code well and this introduces no extra paths.

### DIFF
--- a/docs/reference/management_commands.rst
+++ b/docs/reference/management_commands.rst
@@ -80,7 +80,7 @@ For example, to update just the default backend:
 
     $ python manage.py update_index --backend default
 
-The ``--chunk_size`` option can be used to set the size of chunks that are index at a time. This defaults to
+The ``--chunk_size`` option can be used to set the size of chunks that are indexed at a time. This defaults to
 1000 but may need to be reduced for larger document sizes.
 
 Indexing the schema only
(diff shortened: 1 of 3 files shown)


gasman:
Thanks @dibell! This looks good to me.

As a stylistic nitpick, I'd be inclined to keep `schema_only` and `chunk_size` as optional named arguments rather than positional ones - e.g.

    def update_backend(self, backend_name, schema_only=False, chunk_size=1000):

rather than

    def update_backend(self, backend_name, schema_only, chunk_size):

as it's a bit more explicit and readable that way. (Admittedly repeating `1000` in multiple places isn't great either, so perhaps defining a `DEFAULT_CHUNK_SIZE` constant at the top of the code would be neater...)

docs/reference/management_commands.rst
```diff
@@ -80,6 +80,8 @@ For example, to update just the default backend:
 
     $ python manage.py update_index --backend default
 
+The ``--chunk_size`` option can be used to set the size of chunks that are index at a time. This defaults to
```
tomdyson: index -> indexed

dibell:
@gasman @tomdyson changes made

gasman:
Merged in b2ef833915529cd7f4ca2655697ab2531aa33463 - thanks @dibell!

PR #4363, opened 2018-03-14
Issue 4334: Excluded fields are also excluded when a panel set is exp…

lucasmoeskops:
…licitly defined on a model.

Added functionality for removing excluded fields from model panel set, based on issue 4334. 
With test.


BertrandBordage:
Thank you Lucas ;)

favoyang:
@BertrandBordage could you revert c83a712？ It didn't take MultiFieldPanel into account, and impacted the modeladmin, settings, and snippets edit views. The original "getattr" approach works.

Issue #4200, opened 2018-01-16
When changing your password from with admin a user is presented with an edit button. 


tmsndrs:
I've just reset my password from within admin and I get this horrible large edit button. 
<img width="1267" alt="screen shot 2018-01-16 at 12 33 46" src="https://user-images.githubusercontent.com/7290588/34989276-e10b3338-fab9-11e7-9e69-05d1c7846af4.png">

Can we remove the edit button and amend the message to say 'Your details have been updated. You've been logged out for security reasons, please login to continue.' or something like this.

PR #4364, opened 2018-03-14
Fix sign-in feedback text

Caspowned:
Related to issue #4200


jjanssen:
Related to issue #4200

PR #4365, opened 2018-03-14
Fixed issue #3838: proper messages in document chooser

helmantel:
Created a simple fix for this issue, based upon existing code. Probably no tests needed (?)

- Adjusted the messages in chooser/results.html to be similar like the ones in documents/results.html
- Added class 'tab-toggle' to the upload link to open the upload tab
- Added 'current_collection' to rendering response

### DIFF
--- a/wagtail/documents/views/chooser.py
+++ b/wagtail/documents/views/chooser.py
@@ -50,9 +50,15 @@ def chooser(request):
     q = None
     if 'q' in request.GET or 'p' in request.GET or 'collection_id' in request.GET:
 
+        # Filter by collection
+        current_collection = None
         collection_id = request.GET.get('collection_id')
         if collection_id:
-            documents = documents.filter(collection=collection_id)
+            try:
+                current_collection = Collection.objects.get(id=collection_id)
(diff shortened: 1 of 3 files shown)


jjanssen:
Related to #3838

wagtail/documents/views/chooser.py
```diff
@@ -50,9 +50,15 @@ def chooser(request):
     q = None
     if 'q' in request.GET or 'p' in request.GET or 'collection_id' in request.GET:
 
+        # Filter by collection
+        current_collection = None
         collection_id = request.GET.get('collection_id')
         if collection_id:
-            documents = documents.filter(collection=collection_id)
+            try:
+                current_collection = Collection.objects.get(id=collection_id)
```
BertrandBordage: No need to fetch the current collection. We only use it to change the message in the template, so we can just check whether `collection_id` is in the GET request. I know this will not check if the id exists in the database, but it’s acceptable. The worst case scenario is that the message will not be 100% accurate if the user specifically modifies the request to ask for a non-integer or non-existing collection ID.

gasman:
Completed in #4859.

PR #4367, opened 2018-03-14
Use specific model implementation when deleting a page

### LINKED ISSUE #3827: explorer delete view does not use specific page when checking delete permission
### Issue Summary

When user wants to delete a page this page is retrieved from database as a simple `Page` model, not a specific instance of this page, so other overriden methods are not called `permissions_for_user` in my case. This is a different behaviour than in edit/create etc

kcleong:
This PR fixes #3827:

> When user wants to delete a page this page is retrieved from database as a simple Page model, not a specific instance of this pag

- Get the `specific` implementation of a model in the `delete` method
- Added test to check of the specific `admin_display_title` is shown in the delete confirmation page

### DIFF
--- a/wagtail/admin/views/pages.py
+++ b/wagtail/admin/views/pages.py
@@ -517,7 +517,7 @@ def edit(request, page_id):
 
 
 def delete(request, page_id):
-    page = get_object_or_404(Page, id=page_id)
+    page = get_object_or_404(Page, id=page_id).specific
     if not page.permissions_for_user(request.user).can_delete():
         raise PermissionDenied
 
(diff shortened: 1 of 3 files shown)


wagtail/admin/views/pages.py
```diff
@@ -517,7 +517,7 @@ def edit(request, page_id):
 
 
 def delete(request, page_id):
-    page = get_object_or_404(Page, id=page_id)
+    page = get_object_or_404(Page, id=page_id).specific
```
BertrandBordage: The `.specific` triggers a second SQL query. To avoid that, we can pass directly a queryset to `get_object_or_404` that will prefetch the specific page:
```python
page = get_object_or_404(Page.objects.specific(), id=page_id)
```

gasman: @BertrandBordage I don't think this will make any difference to the number of queries - the `specific()` method on a queryset will execute as many additional queries as there are distinct page types in the resultset (i.e. one extra query, in this case).

gasman:
Merged in 2fdb924b568a8eec4e62ae3aa899a064e6ea3ded - thanks @kcleong!

wagtail/admin/views/pages.py
BertrandBordage: Ah, you’re right, sorry about that.

PR #4368, opened 2018-03-14
4353 breadcrumbs get admin display title

helmantel:
#4353 Simple fix to display admin_display_title in breadcrumbs.


helmantel:
Was already fixed in #4354

PR #4369, opened 2018-03-14
redirect to change form if page has no subpage_types

noslouch:
The admin UX is a little weird in the explore menu for leaf nodes. If a leaf node doesn't have any children (think `BlogIndex -> BlogEntry`), clicking on the title brings up an empty explore menu. This the case even if a model has its `subpage_types` explicitly set to an empty list.

This patch checks if the retrieve `parent_page` has its `subpage_types` set and if it has been explicitly set to an empty list [as per the docs](http://docs.wagtail.io/en/stable/topics/pages.html#an-example-wagtail-page-model), then redirect the user to the edit page for that instance.

I think this matches the expectation for admin users who see a leaf node's title in the explore menu and assume that it will take them to the change form for that instance.

Happy to add tests, documentation, and changelog updates if this kind of edit is desired.

### DIFF
--- a/wagtail/admin/tests/test_pages_views.py
+++ b/wagtail/admin/tests/test_pages_views.py
@@ -3158,12 +3158,12 @@ class TestSubpageBusinessRules(TestCase, WagtailTestUtils):
         self.assertContains(response, BusinessChild.get_verbose_name())
 
     def test_business_child_subpage(self):
-        add_subpage_url = reverse('wagtailadmin_pages:add_subpage', args=(self.business_child.id, ))
+        edit_page_url = reverse('wagtailadmin_pages:edit', args=(self.business_child.id, ))
 
-        # explorer should not contain a link to 'add child page', as this page doesn't accept subpages
+        # explorer should redirect to edit view if the page doesn't accept subpages
         response = self.client.get(reverse('wagtailadmin_explore', args=(self.business_child.id, )))
-        self.assertEqual(response.status_code, 200)
-        self.assertNotContains(response, add_subpage_url)
+        self.assertEqual(response.status_code, 302)
+        self.assertEqual(response.url, edit_page_url)
 
         # this also means that fetching add_subpage is blocked at the permission-check level
         response = self.client.get(reverse('wagtailadmin_pages:add_subpage', args=(self.business_child.id, )))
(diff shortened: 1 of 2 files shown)


noslouch:
bump. is there something I can add or change to get some feedback on this feature request?

noslouch:
I can write docs too if that's a requirement for getting this merged in?

noslouch:
Bump for some feedback?

gasman:
Hi @noslouch,
Sorry for the delay in reviewing this - will get to it as soon as possible. While this is a small change at the code level, it's something we need to approach carefully, as there may be less obvious workflows through the admin interface that land on the index view of a leaf page - in those cases we need to be sure that redirecting to the edit view is the helpful (and expected) thing to do, rather than adding an extra level of unpredictability to the UX.

noslouch:
Hey @gasman. Thanks for the feedback. Based on my early look, I found that subpage types are initially None. Are there cases when the attribute is changed by other actions or other code?

gasman:
@noslouch No, as far as I can recall nothing within Wagtail itself ever populates `subpage_types` - it's only specified in user code. (It's also worth noting that some sites override related `Page` methods such as `can_create_at` to implement dynamic rules such as only allowing a page to have one child - although that's never been formally documented, so sites that do that are really doing it at their own risk...)

gasman:
Having compared the behaviour of the explorer menu with other paths through the Wagtail UI (e.g. the landing page you reach after publishing a page), it looks like the explorer menu is indeed the odd one out here - the other workflows tend to avoid dropping the user at the explorer view of a leaf page.

With that in mind, it makes sense to approach this as a proposed change to the explorer menu, and apply the fix there. (Placing a redirect on the explorer view would be overly broad, and risk being triggered in scenarios where we didn't intend it - for example, on a site with tightly locked-down permissions where a user only has permission to edit a single (leaf) page, or in cases where a page previously allowed children but a `subpage_types` rule was added later, in which case the redirect would make the existing child pages inaccessible.)

I think there are arguments both ways about which behaviour is more consistent - linking to the edit page would avoid ending up at the explorer view of a leaf page, but it would also mean that the title link goes to different places for leaf vs non-leaf pages according to rules that might not be obvious to the user. Looking back at the discussion from when the explorer menu was first developed, it looks like both options were considered, and the 'edit' icon added as a way to mitigate any confusion: https://github.com/wagtail/wagtail/pull/3012#issuecomment-288994405

Will close this ticket now as per that decision, but we'd be happy to consider any further suggestions for improvements to the UX here. Thanks @noslouch!

(cc @wagtail/ux-design)

Issue #2335, opened 2016-03-10
Generic in-admin help, which can be attached to any model and displays in page editor


davecranwell:
The configuration of some sites is such that a client, particularly someone new to the project, often has a lot of questions about the behaviour and relationships between content. It would be nice if Wagtail were able to support generic help text, not attached to fields, which could provide this.

E.g A `NewsPage` model has a relationship to multiple `news_authors` in an inline panel, linking to an `Author` model. A `NewsListingPage` allows the visitor to search the `NewsPage`s by author. It's meaningless to list _all_ the authors, particularly if most have no `NewsPage`s to their name, so the filter is programmed to only display `Authors` which have been attributed to a `NewsPage`.

As a client trying to work out why their author isn't selectable from the search filter, its a question of explaining the configuration to the client. We might write separate documentation, but this can be forgotten or ignored. What would be nicer is help text explaining this behaviour, which could be put somewhere in the admin for easy reference when confusion arises.

davecranwell:
It occurs to me that this help is most frequently required when a part of a content type is automated, alongside other content which is not. The apparent omission of the fields for the automated bit can confuse clients. 

Perhaps therefore a "HelpPanel" is a solution of sorts. It would be a Panel in the editor like any other, but rather than containing fields, it just contains text.

gasman:
Implemented in #4374

easherma-truth:
This works well except in cases where we have StructBlocks which themselves have need of placeholders for dynamic info or otherwise need extra helpful content to be shared with authors.

thibaudcolas:
@easherma-truth could you further describe your use case in a separate issue / discussion thread? This has been closed for a while so it’s unlikely lots of people will be following discussions here.

PR #4374, opened 2018-03-15
Help Panel

chungky:
This commit addresses[ this issue](https://github.com/wagtail/wagtail/issues/2335) by adding a HelpPanel. You can pass in a template or html that will show up as a panel on the admin UI to help content managers.

### DIFF
--- a/docs/reference/pages/panels.rst
+++ b/docs/reference/pages/panels.rst
@@ -243,6 +243,24 @@ SnippetChooserPanel
     See :ref:`snippets` for more information.
 
 
+HelpPanel
+---------
+
+.. module:: wagtail.admin.edit_handlers
+
+.. class:: HelpPanel(content, template)
+
+    .. attribute:: HelpPanel.content
+
+        This is string HTML that you can pass in directly to render inside the HelpPanel
+
+    .. attribute:: HelpPanel.template
+
+        Alternatively, you can pass in a template name. Specifying the template will override the content attribute.
+
+    You can also set the ``heading`` or ``classname`` attributes depending on what you want.
+
+
 Built-in Fields and Choosers
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/reference/pages/panels.rst
+++ b/docs/reference/pages/panels.rst
@@ -248,17 +248,23 @@ HelpPanel
 
 .. module:: wagtail.admin.edit_handlers
 
-.. class:: HelpPanel(content, template)
+.. class:: HelpPanel(content='', template='wagtailadmin/edit_handlers/help_panel.html', heading='', classname='')
 
     .. attribute:: HelpPanel.content
 
-        This is string HTML that you can pass in directly to render inside the HelpPanel
+        HTML string that gets displayed in the panel.
 
     .. attribute:: HelpPanel.template
 
-        Alternatively, you can pass in a template name. Specifying the template will override the content attribute.
+        Path to a template rendering the full panel HTML.
 
-    You can also set the ``heading`` or ``classname`` attributes depending on what you want.
+    .. attribute:: HelpPanel.heading (optional)
+
+        A heading for the fields
+
+    .. attribute:: HelpPanel.classname (optional)
+
+        String of CSS classes given to the panel which are used in formatting and scripted interactivity. By default, panels are formatted as inset fields.
 
 
 Built-in Fields and Choosers
--- a/docs/reference/pages/panels.rst
+++ b/docs/reference/pages/panels.rst
@@ -258,13 +258,13 @@ HelpPanel
 
         Path to a template rendering the full panel HTML.
 
-    .. attribute:: HelpPanel.heading (optional)
+    .. attribute:: HelpPanel.heading
 
-        A heading for the fields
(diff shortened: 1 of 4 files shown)


wagtail/admin/templates/wagtailadmin/edit_handlers/help_panel.html
```diff
@@ -0,0 +1,6 @@
+<fieldset>
+	{% if self.heading %}
```
BertrandBordage: This line is 2 spaces too indented.

jjanssen:
Related to #2335

wagtail/admin/templates/wagtailadmin/edit_handlers/help_panel.html
```diff
@@ -0,0 +1,6 @@
...
+    	<legend>{{ self.heading }}</legend>
+    {% endif %}
+    <div class="{{ self.classname }}">{{ self.content }}</div>
```
BertrandBordage: Could you try to put HTML tags in `content` to see how it behaves? Given the default behaviour of the Django template language escaping HTML, I think there should be a `|safe` filter here if we want to accept raw HTML.

thibaudcolas:
@chungky 👏 

To make this PR easier to review, it would be helpful if you could:

- Add a screenshot to demonstrate what this looks like for end users.
- From what I understand this is introducing reusable code – does it need to be documented somewhere? Not expecting you to know how and where to do that, but please let us know what you think would be useful.
- This introduces a bit of front-end code – please mention what browsers you tried it in (the code is fairly simple in that instance so I'm not expecting issues, but it's still good to know what has been tested so far, no matter how little).

wagtail/admin/templates/wagtailadmin/edit_handlers/help_panel.html
BertrandBordage: No, the rest was correctly indented, it’s only this line that had 2 extra spaces ;)

chungky:
screen shot 2018-03-15 at 3 19 43 am

chungky:
Tested on the latest versions of chrome, firefox, and safari.

wagtail/admin/templates/wagtailadmin/edit_handlers/help_panel.html
BertrandBordage: Ah, I see the problem, you’re using tabs. Please use an indentation base of 4 spaces instead.

docs/reference/pages/panels.rst
```diff
@@ -243,6 +243,24 @@ SnippetChooserPanel
...
+HelpPanel
+---------
+
+.. module:: wagtail.admin.edit_handlers
+
+.. class:: HelpPanel(content, template)
+
+    .. attribute:: HelpPanel.content
+
+        This is string HTML that you can pass in directly to render inside the HelpPanel
```
tomdyson: "string HTML" -> "an HTML string", perhaps?

docs/reference/pages/panels.rst
```diff
@@ -243,6 +243,24 @@ SnippetChooserPanel
...
+
+    .. attribute:: HelpPanel.template
+
+        Alternatively, you can pass in a template name. Specifying the template will override the content attribute.
+
+    You can also set the ``heading`` or ``classname`` attributes depending on what you want.
```
BertrandBordage: Please detail what these attributes are. You can shamelessly copy/paste a short definition from another panel.

docs/reference/pages/panels.rst
```diff
@@ -243,6 +243,24 @@ SnippetChooserPanel
     See :ref:`snippets` for more information.
 
 
+HelpPanel
+---------
+
+.. module:: wagtail.admin.edit_handlers
+
+.. class:: HelpPanel(content, template)
```
BertrandBordage: Please use exactly the same definition as in the code. The default values are missing, as well as the arguments `heading` and `classname`.

docs/reference/pages/panels.rst
BertrandBordage: I would opt for a simpler version, like:
`HTML string displayed in the panel.`

docs/reference/pages/panels.rst
```diff
@@ -243,6 +243,24 @@ SnippetChooserPanel
...
+
+    .. attribute:: HelpPanel.content
+
+        This is string HTML that you can pass in directly to render inside the HelpPanel
+
+    .. attribute:: HelpPanel.template
+
+        Alternatively, you can pass in a template name. Specifying the template will override the content attribute.
```
BertrandBordage: I would change this to:
`Path to a template rendering the full panel HTML.`

No need to specify how the template itself work, since there is a default template for help panel, people can just look at it to see how it works.

BertrandBordage:
Could you put as well a screenshot for the version of the panel with a title, please? :)
After all these little changes I suggested, it’ll be good to merge :D

chungky:
screen shot 2018-03-15 at 7 45 54 am

chungky:
@BertrandBordage Any other tweaks?

BertrandBordage:
Thank you very much @chungky :)

Issue #4325, opened 2018-03-01
No way to change your email address from account settings


tmsndrs:
I would expect to be able to do this from my account settings but currently the only way to do it is from the user settings.

PR #4375, opened 2018-03-15
Added change email functionality from the account settings

alejandrogarza:
Added change email functionality from the account settings as requested on 
https://github.com/wagtail/wagtail/issues/4325

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -16,6 +16,7 @@ Changelog
  * Added error notification when running the `wagtail` command on Python <3.4 (Matt Westcott)
  * `update_index` management command now accepts a `--chunk_size` option to determine the number of items to load at once (Dave Bell)
  * Added hook `register_account_menu_item` to add new account preference items (Michael van Tellingen)
+ * Added change email functionality from the account settings (Alejandro Garza, Alexs Mathilda)
  * Fix: Status button on 'edit page' now links to the correct URL when live and draft slug differ (LB (Ben Johnston))
  * Fix: Image title text in the gallery and in the chooser now wraps for long filenames (LB (Ben Johnston), Luiz Boaretto)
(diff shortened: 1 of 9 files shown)


wagtail/admin/templates/wagtailadmin/account/account.html
```diff
@@ -18,6 +18,15 @@
...
+            <li class="row row-flush">
+                <div class="col6">
+                    <a href="{% url 'wagtailadmin_account_change_email' %}" class="button button-primary">{% trans "Change email" %}</a>
+                </div>
+
+                <small class="col6">
+                    {% trans "Change email address linked to your account." %}
+                </small>
+            </li>
             {% if show_change_password %}
```
thibaudcolas: I see other blocks are wrapped in a `if show` conditional like this – would we need to do the same for email?

alejandrogarza: Hey Thibaud, added the screenshot of the link in account settings and for the form. 
image
image

Tested it on Chrome/Safari/Firefox
On the `wagtail/admin/templates/wagtailadmin/account/account.html` you can see a few `if` statements which are used for: 
*Password: When a user doesn't have a usable password, or password was never set, password can't be changed (see `has_usable_password` )
*Notifications: When a user doesn't have editing rights, so notifications aren't needed since none of his posts will be edited

From what I know, and confirming it from the `User Edit form`, changing email doesn't have any restriction. 

Cheers!

AlexsMathilda:
tested  and approved email functionality from the account settings on Windows IE11, MS Edge, latest Chrome and FF. MacOS Safari, latest Chrome and Firefox, Android Chrome 7, iOS Safari 10.

thibaudcolas:
👏 it’s in! Will make its way to a Wagtail site near you in Wagtail 2.1. Thank you @alejandrogarza.

Issue #3441, opened 2017-03-10
Snippet / Modeladmin / settings forms should use `validation_error` helper


gasman:
See #3440, which introduces a new `validation_error` helper to ensure that validation errors on the page edit form - both field-specific, and non-field errors - are displayed appropriately. After this is merged, we should update the other admin views that work with user-defined forms so that they're also using this helper:

* snippets
* ModelAdmin
* `wagtail.contrib.settings`

As part of this task, we should make sure that validation error handling is covered by unit tests - i.e. for each of these modules, there should be at least one test similar to https://github.com/wagtail/wagtail/pull/3440/files#diff-509186d17133395aafeb7c62930e9881R3964 which posts an invalid form and confirms that the form is redisplayed with the expected error message included. (There's no need to go through all permutations of field and non-field errors, since that's already covered by #3440.)

ababic:
Hi @gasman, do you know if the solution implemented for pages / proposed here will surface non-field errors from `InlinePanel` instances too?

gasman:
@ababic I suspect it won't... [`ClusterForm.is_valid`](https://github.com/wagtail/django-modelcluster/blob/fa1e2fc3194c8676e3ae5a2beeb7de25c6d8f219/modelcluster/forms.py#L258) recursively checks the `is_valid` of the associated formsets, so it will at least recognise that validation has failed. However, to actually report the errors we'd have to make the `errors` dict of the child forms/formsets bubble up to the top level (or alternatively, have the `validation_error` helper dig in to `form.formsets['foo'].errors`, but that seems like a bad idea - `validation_error` should ideally not have any special case logic for ClusterForm, and certainly shouldn't have to interface with it at such a low level).

I guess the correct behaviour for "how should errors in child formsets be surfaced in `form.errors`" has never been defined - Django doesn't have an answer to that, because child formsets are a modelcluster-specific concept - so that would be the first thing to address. Two possibilities spring to mind:

- extend the error dict object to have a `CHILD_FORM_ERRORS` property alongside `NON_FIELD_ERRORS` and the per-field keys, something like:

        event_form.errors = {
            'title': ["This field is required"],
            'NON_FIELD_ERRORS': ["End date must be after start date"],
            'CHILD_FORM_ERRORS': {
                'speakers': {
                    0: {'NON_FIELD_ERRORS': "Speaker must agree to the T+Cs"}
                }
        }

[...]

PR #4376, opened 2018-03-15
Cleanup/3441 snippet modeladmin validation error helper

Nebulosar:
Cleanup tasks for #3441 

Added the validation_message funtion to:

- Snippets

- Modeladmin

- `contrib.settings`

Also updated the tests that run through this code

### DIFF
--- a/wagtail/contrib/modeladmin/tests/test_simple_modeladmin.py
+++ b/wagtail/contrib/modeladmin/tests/test_simple_modeladmin.py
@@ -179,6 +179,8 @@ class TestCreateView(TestCase, WagtailTestUtils):
 
         # Check that a form error was raised
         self.assertFormError(response, 'form', 'title', "This field is required.")
+        self.assertContains(response, """<p class="error-message"><span>This field is required.</span></p>""",
+                            count=1, html=True)
 
     def test_exclude_passed_to_extract_panel_definitions(self):
         path_to_form_fields_exclude_property = 'wagtail.contrib.modeladmin.options.ModelAdmin.form_fields_exclude'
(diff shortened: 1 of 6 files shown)


ababic:
HI @Nebulosar,

Thank you for helping with this. I think it'll be really valuable to have validation errors handled more consistently throughout Wagtail. I don't think this is far away. However, I think there might perhaps be some misunderstanding around what the existing tests are testing, and the purpose of the `validation_error` helper added by @gasman in #3440.

Please correct me if I've missed something, but it seems you haven't added any tests that show the `validation_error` making a difference in the UI. See the [tests added in #3440](https://github.com/wagtail/wagtail/pull/3440/files#diff-509186d17133395aafeb7c62930e9881) for an example. What we should be looking to confirm is that:

1. Field-specific errors now appear in a list at the top of the page as well as next to relevant field.
2. And non-field errors are also rendered in the same spot.

Do you think you might be able to revise the tests to show this?

Nebulosar:
Hello @ababic ,

After looking at the tests in #3440 I can understand what you mend.

I will look into testing the specific tests to check for UI elements

ababic:
Okay, thanks @Nebulosar. Please let me know once you're happy with things, and I'll review things again.

gasman:
I'm happy to merge this as-is - as I said in #3441, I don't think it's necessary to re-test all the permutations of field/non-field errors as implemented by the `validation_error` helper, as that's all covered by the tests in #3440. Given that we can see the same helper in use here, and we have tests to confirm that it's hooked up correctly and responding to simple validation errors, we can reasonably take it as read that it will respond appropriately to the more esoteric error cases :-)

gasman:
Merged in a03c310858569858d70f349e3f6e3f7a4bf159e3 - thanks @Nebulosar!

Issue #4298, opened 2018-02-16
Optimise representation of deeply-nested StreamField blocks in migrations


coredumperror:
Because the actual database backend of a StreamField is just a simple `LONGTEXT` populated by json-serialized block contents, the database does not need to know about changes made to the Block structure within the StreamField.

However, since Wagtail currently does provide that information to the migrations, this causes complex StreamField instances to generate *gargantuan* migrations whenever even the tiniest change is made to a single Block used by the field. Here's an [example](https://pastebin.com/pMQriqXZ). This one model's migration code is nearly **500,000 characters**. 

And another half megabyte of text will be repeated in each subsequent migration if even a single field in a single Block has a single attribute changed. This is highly undesirable, and as far as I can tell, *completely pointless*. The auto-generated migration files don't need to care about the internal structure of a StreamField; only manually written migrations that migrate the data from an old format to a new one need to care, and those don't need to be half a megabyte of code.

I'm not yet entirely sure how best to remedy this, but I think it will have something to do with `StreamField.deconstruct()`.:

    def deconstruct(self):
        name, path, _, kwargs = super(StreamField, self).deconstruct()
        block_types = self.stream_block.child_blocks.items()
        args = [block_types]
        return name, path, args, kwargs

I don't think there's any good reason for it to care about its child blocks. Though while I do have a lot of experience with dealing with, and hacking around, StreamFields, I'm not an expert. Maybe I'm missing something?

gasman:
Hi @coredumperror,
As mentioned on #3710, this is line with Django's principle of preserving all properties of a model field within migrations, not just the properties that directly affect the database schema. This ensures that any Python code in the migration, particularly for data migrations, has access to a 'frozen' version of the ORM model as it existed at that point in the migration path, regardless of how much the real model in models.py may have moved on from there.

To be honest, your example code shows a level of configurability that goes way beyond anything I anticipated, or would recommend. The purpose of Wagtail as a CMS is to provide a separation between content editing and design decisions - hooking up every conceivable design choice to a StreamField element doesn't achieve that.

chosak:
@gasman following up on this:

> your example code shows a level of configurability that goes way beyond anything I anticipated, or would recommend

This problem (large migrations due to highly configurable StreamFields) is something that we've run into as well, and is a major pain point around maintenance of our large Wagtail site.

[...]

PR #4378, opened 2018-03-15
Cleanup for the `Block.__eq__` method

gasman:
A bit of cleanup that doesn't do very much by itself, but takes us a few baby steps towards https://github.com/wagtail/wagtail/issues/4298#issuecomment-367656028 (which I'm probably not going to get round to this week, so I didn't want to leave these commits in limbo).

TLDR: when StreamField was first implemented, we had to implement the equality operator on Block to work around [a Django bug](https://code.djangoproject.com/ticket/24340) that broke the `makemigrations` change detection logic. In theory, this code _should_ have been redundant ever since Django 1.9, but: A) it wasn't, due to a couple of silly oversights; and B) the docstring for `__eq__` still referred to that long-fixed bug.

### DIFF
--- a/wagtail/core/blocks/base.py
+++ b/wagtail/core/blocks/base.py
@@ -358,19 +358,19 @@ class Block(metaclass=BaseBlock):
 
     def __eq__(self, other):
         """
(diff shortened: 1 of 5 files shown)


BertrandBordage:
Thank you Matt :)

Issue #2309, opened 2016-03-04
Visual design of login screen


davecranwell:
On any decent-resolution monitor, both the following look a bit ridiculous. Compared to our competitors It's certainly eye-catching in an unusual way, but I don't feel it's an attractive first impression to make.

![screen shot 2016-03-04 at 10 02 36](https://cloud.githubusercontent.com/assets/352835/13523935/43a776da-e1f0-11e5-94a4-e5bddd389519.png)
![screen shot 2016-03-04 at 10 02 41](https://cloud.githubusercontent.com/assets/352835/13523936/43bf7d70-e1f0-11e5-9dba-78202ced91cf.png)

A brief skim of any "Design Inspiration" site indicates a fad of centralised boxes with giant shadows and background textures e.g  http://collectui.com/challenges/sign-up  

I'm keen to _avoid_ an approach which dates quickly, so I'm not suggesting we jump blindly onto the bandwagons illustrated above, but there are qualities in all these examples which I feel wagtail shares, which our login screen doesn't convey.

heymonkeyriot:
Agree with this, and I think this ties in with lots of other ongoing conversations within these issues at the moment. There's clearly a few things that we can improve within the UI. Some already raised as issues here on Github e.g. where to locate help text, how to articulate the explorer and how to amend StreamField; and others not, such as the offline conversations happening around the dashboard

Having said that I'm not sure about any of the examples shared from collectui though. I think they're going to get in the way of the users logging in and get in the way of developers wanting to extend that login area if it ever needs to be closer to their client's branding.

My personal feeling is that Github, Squarespace, Wordpress et al have good login patterns to use as inspiration. They're all simple screens that do their best to not get in the way of the user task. One of the things, for example, that Wordpress does well is how easy it is to 'brand' the login screen as a developer.

With that in mind some thoughts are below for discussion. Beyond styling, there's some slight structural change to the existing template: the sitename is being printed, the user and password icons have been removed, input placeholders (e.g. `enter your username`) have been removed, the Wagtail logo has been introduced and been placed under the signin button

#### 1. Sign-in {site name}
![_0003_login - idea 3 2x](https://cloud.githubusercontent.com/assets/11335309/13554336/14321d02-e39d-11e5-95c7-19e47a85e760.jpg)

![_0004_login - idea 4 2x](https://cloud.githubusercontent.com/assets/11335309/13554335/1431db44-e39d-11e5-85e1-cb7e5788288c.jpg)
[...]

PR #4379, opened 2018-03-15
Improve password reset error messages - fix #4097

### LINKED ISSUE #4097: Password reset reports 'passwords do not match' for any kind of invalidity
### Issue Summary

When entering a password that breaks validation rules (e.g. too short) on the password reset workflow, the error message is always "The passwords do not match. Please try again." regardless of the actual validation error.

### Steps to Reproduce

1. Bring up bakerydemo (or any other project with `AUTH_PASSWORD_VALIDATORS` configured)
2. Visit /admin, click 'Forgotten it?' and enter a user's email (e.g. admin@example.com)
3. Follow the link in the email, and enter `a` as the password in both fields

Expected: An error message saying that the password is too short
Actual: An error message saying "The passwords do not match. Please try again."

### Technical details

* Python version: 3.5.2
* Django version: 1.11.8
* Wagtail version: 2.0a1

lucasmoeskops:
Fixes #4097

The error messages are coming from django and are bound to fields (e.g. 'This field is required'). Therefor I put them next to the fields. 

Together with @jjanssen we came to this style, although she mentions that the general style for these forms should be improved soon.
issue-4097

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/account/password_reset/confirm.html
+++ b/wagtail/admin/templates/wagtailadmin/account/password_reset/confirm.html
@@ -4,46 +4,46 @@
 {% block bodyclass %}login{% endblock %}
 
 {% block extra_css %}
-{{ block.super }}
+    {{ block.super }}
 
-<link rel="stylesheet" href="{% static 'wagtailadmin/css/layouts/login.css' %}" type="text/css" />
+    <link rel="stylesheet" href="{% static 'wagtailadmin/css/layouts/login.css' %}" type="text/css" />
 {% endblock %}
 
 {% block furniture %}
-<div class="content-wrapper">
+    <div class="content-wrapper">
 
-    {% if form.errors %}
-    <div class="messages">
-        <ul>
+      {% if form.errors %}
+        <div class="messages">
+          <ul>
             <li class="error">{% trans 'Password reset failed. Please see the errors below.' %}</li>
-        </ul>
-        {{ form.non_field_errors }}
+          </ul>
+          {{ form.non_field_errors }}
+        </div>
+      {% endif %}
+
+        <form method="post" novalidate>
+            {% csrf_token %}
+            <h1>{% trans "Set your new password" %}</h1>
+
+            <ul class="fields">
+                <li>
+                  <div class="field">
+                      {{ form.new_password1.label_tag }}
+                      {{ form.new_password1 }}
+                      {{ form.errors.new_password1 }}
+                  </div>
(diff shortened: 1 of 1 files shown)


thibaudcolas:
👏 @lucasmoeskops could you briefly mention in what browsers you tried this?

See #2309 for the related design changes. @tmsndrs do you think we should proceed with this before #2309, or wait?

jjanssen:
To give a little insight in the changes Lucas and I made. We explicitly decided to go for a more compact error message on this page for now. Since the regular error message is a bit intimidating when multiple validation errors occur (with the possibility of even 3 errors). 

See the following example which has 2 validation errors on the password level:
image

Too weigh in: although it causes a bit of an inconsistent error message (compared to other flows), I think this would an appropriate change for the time being. Especially since we don't have an exact timeline for the related design changes to the login / password flow.

wagtail/admin/templates/wagtailadmin/account/password_reset/confirm.html
```diff
@@ -12,30 +12,32 @@
...
             <ul class="fields">
                 <li>
-                  <div class="field">
-                      {{ form.new_password1.label_tag }}
-                      {{ form.new_password1 }}
-                  </div>
+                    <div class="field">
+                        {{ form.new_password1.label_tag }}
+                        {{ form.new_password1 }}
+                        <div class="help-block help-critical">
```
jjanssen: This would need an if statement to prevent parsing empty errors.

wagtail/admin/templates/wagtailadmin/account/password_reset/confirm.html
```diff
@@ -12,30 +12,32 @@
...
                 </li>
                 <li>
-                  <div class="field">
-                      {{ form.new_password2.label_tag }}
-                      {{ form.new_password2 }}
-                  </div>
+                    <div class="field">
+                        {{ form.new_password2.label_tag }}
+                        {{ form.new_password2 }}
+                        <div class="help-block help-critical">
```
jjanssen: This would need an if statement to prevent parsing empty errors.

gasman:
+1 for going ahead with this fix rather than waiting for #2309. An odd-looking but correct error message is better than one that's flat-out wrong :-)

lucasmoeskops:
Tested it in the latest versions of Chrome, Firefox and Safari and looks as expected there.

jjanssen:
Thanks! 🎉Merged with f7b0b69

Issue #1806, opened 2015-10-09
Generic ModelChooserBlock for StreamField


LiamBrenner:
I'd like to propose a `ForeignBlock` or similar using `models.ForeignKey` as it's base for allowing integrations with other model clusters/models. It would most likely need to extend `blocks.ChooserBlock` and shouldn't be too difficult.

gasman:
The current recommended way of doing this would be to register your target model as a snippet (http://docs.wagtail.io/en/v1.1/topics/snippets.html), and use [SnippetChooserBlock](http://docs.wagtail.io/en/v1.1/topics/streamfield.html#snippetchooserblock).

However, in future we intend to move towards adopting [wagtailmodeladmin](https://github.com/ababic/wagtailmodeladmin) so that we can support arbitrary models within Wagtail admin without them having to be registered as snippets - and a general 'ModelChooserBlock' would be an important part of that.

timorieber:
+1

pymarco:
+1

Proper-Job:
I'm sorry if I missed it, but is there a way to select a ModelAdmin instance from within a StreamField now that wagtailmodeladmin has been released?

gasman:
@Proper-Job Not currently, no.

Proper-Job:
Ok, thanks.

ababic:
I think the blocker to having a ModelAdmin-based chooser has been: ModelAdmin classes know about the model they represent, but there's no easy way to access a ModelAdmin class when provided with model. Without moving to a 'django-like' registration method, where the relationships between models and ModelAdmin classes are stored in a central/reference-able place, I don't see how we can circumvent this. But that has a tonne of other implications that need thinking through. That's maybe a discussion for the developers board though, rather than here.

JoshBarr:
We at Springload have a generic ModelChooserField in the works, but it:

* Relies on the React ecosystem 
* Relies on a very basic REST API endpoint that does 'creative' things to serializers. 
* Has no tests (yet)
* Has no translations (yet)
* Has no docs (yet)

If anyone's interested in helping out with it, I'll publish the code.

kaedroho:
That's similar to the idea I have with admin view sets (#2113). This would be a place for "view sets" to be registered within Wagtail admin. This would automatically set up URLs to those views, menu items, admin API, etc. A generic chooser block could be implemented using this.

[...]

PR #4380, opened 2018-03-15
Handle ValueError in case the form send '' instead of int for ChooserBlock

Parbhat:
I encountered an issue while extending `ChooserBlock` to work with a Django model. As mentioned in Issue #1806 `SnippetChooserBlock` is the recommended method. I am not sure if this should go into Wagtail core so thought to get the views from core developers.

The following code works fine but when no service is selected Instead of giving `Validation error: Field required` got Server error `ValueError: invalid literal for int() with base 10:`.

```python
class ServiceChooserBlock(blocks.ChooserBlock):
     target_model = Service
     widget = forms.Select
```

On looking into the code, `value_from_form ` is the function which takes value from the form. When no service is selected, the value sent by form is `''` which breaks here 

```python
return self.target_model.objects.get(pk=value)
```

### DIFF
--- a/wagtail/core/blocks/field_block.py
+++ b/wagtail/core/blocks/field_block.py
@@ -571,7 +571,7 @@ class ChooserBlock(FieldBlock):
         else:
             try:
                 return self.target_model.objects.get(pk=value)
-            except self.target_model.DoesNotExist:
+            except (self.target_model.DoesNotExist, ValueError):
                 return None
 
     def clean(self, value):


gasman:
Thanks @Parbhat!
Related: #3169, #4351. (`ChooserBlock` was never designed to be used directly with a Select widget, and we'd need additional fixes / tests / documentation before we can consider this to be a supported setup.)

Parbhat:
@gasman Thanks for the review. Yes, it is better to have tests / Documentation to make it a supported feature. I am currently using `ChooserBlock` in a project by making the fix in the subclass `ServiceChooserBlock` and it works fine. Will work on the tests/docs and hopefully find more issues and fix them.

gasman:

Parbhat:
Thanks for creating a separate issue.

Issue #4199, opened 2018-01-15
Making Wagtail Accessible for users with disabilities


mush42:
Hello everyone,

After reading the docs, I've decided to use Wagtail on one of my personal projects.

Since I'm blind, I've discovered that my experience on editing content would be rather unpleasant, because the admin interface of Wagtail has some accessibility problems.

For example, the icon-font being used in the admin is not screen-readers friendly, it keeps adding random letters wherever it is used. This happen because it doesn't use the safe-unicode character range. This could be resolved by a simple javascript function which hides the icons  from screen-readers.

## Proposal:

Making Wagtail an accessible content management system. By adding necessary markup (and scripts) to wagtail's templates and outputs.

## Goal:

Enable the users of Assistive technology to efficiently manage and edit content in a Wagtail powered site.

## Rationale:

> "It is essential that the Web be accessible in order to provide equal access and equal opportunity to people with disabilities. An accessible Web can also help people with disabilities more actively participate in society. Also, web accessibility is required by laws and policies in some cases." --- Web Accessibility Initiative (https://www.w3.org/WAI/intro/accessibility.php)

## Execution:

I'll make a list of each area of Wagtail that may need some accessibility related fixes, and open an issue for each one.

## Questions:

- Is openning issues is the preferred way? or should I fix it myself and create a pull request?
- If yes, how pull request should be structured? in a batch or on a feature by feature bases?

Regards,

thibaudcolas:
Hey @mush42, thank you for reaching out and taking the time to write such a well thought-out issue. Making Wagtail accessible is something we care a lot about, even though I don't think we've stated it formally so far.

Generally, I think what we need first and foremost is to be told what to do and to define some standards for current and future developments (and enforce them in pull request reviews). This could take the form of some documentation and/or automated tests. Then, there are the existing problems with the admin, where it would help us tremendously if you took the time to give us your take – and then helped us prioritise what to fix.

The approach I would advise is for you to first make an audit of all the issues you encounter, listing them here, and indicating how problematic they are. Then we can discuss how best to address them, and make separate issues / pull requests for each problem (the smaller the PR, the easier for us to merge it). You're more than welcome to make pull requests, but I imagine it would be quite an undertaking to fix everything. I would advise you to first define what to do – then it'll be easier for others to contribute too.

---

Here are specific issues we're already aware of:

- #2224 *Certain colours combinations not validating against WCAG AA*
[...]

Issue #1224, opened 2015-04-21
Convert icon font to SVG: make icons screen-reader friendly and easier to update


JoshBarr:
We just had a user try to access the CMS from a tethered mobile device, and they were worried about all the "q" characters showing up in the admin interface before the fonts loaded... I think possibly their connection closed before the whole font downloaded, so they had a mix of icons and placeholder characters showing up in their interface.  

Is there a way we could more gracefully handle the flash of unstyled content, or preload the typefaces ahead of showing the UI?

davecranwell:
The more I read up on icon fonts the more I conclude they're actually not that great, for precisely these brittleness issues. I'd much rather convert all icon use to SVG sprites.

tmsndrs:
I agree. Inline SVG is much better. I've had a couple of sessions where the icon font just hasn't loaded at all (I didn't look into why at the time). 

Good SVG vs icon font write up here https://css-tricks.com/icon-fonts-vs-svg/

davecranwell:
That's the exact article I had in mind.

JoshBarr:
+1. We use SVG for pretty much everything.

tomdyson:
@JoshBarr is this something that you or your team could help with?

davecranwell:
The use of pseudo-elements is also a pain in the neck at the moment, due to the slightly unpredictable nature of EMs etc.

I'd imagine the fix for this might be to replace all uses of icon-specific classes with a template tag that outputs `<svg><use ...></svg>`, referencing a giant inline svg sprite at the top of the page.

Does this align with your approach @JoshBarr ?

gasman:
If we're going to be rethinking our approach to icons, then let me put in a few wishlist requests...

* If adding a new icon is going to involve updating a central resource (e.g. font / sprite), document the process for doing that
* Make it possible for end-users to define their own icons in addition to wagtailadmin's stock set, for use in StreamField blocks or custom (hook-defined) menu items

davecranwell:
Adding icons to the global set could be done with a hook which adds a user's own svg sprite inline to the top of the page. Our proposed icon template tag would continue to work as normal, because the user could just reference the IDs within their custom svg instead of the core one.

JoshBarr:
We've been using a CSS file with base64 encoded SVGs through Grunticon, like this: [Kiwibank sprite sheet](http://kiwibank.co.nz/assets-personal-banking/css/icons.data.svg.css)

[...]

PR #4381, opened 2018-03-15
Add icon template tag with accessibility options (PoC)

SanderTuit:
At the moment, the admin interface has accessibility issues, as mentioned in https://github.com/wagtail/wagtail/issues/4199. One of these issues is with icons (from the icon font): these add characters to the admin when read out loud by a screenreader. Some icons don't have an accompanying label and are not accessible at all for screenreader users. 

To address these issues, icons shouldn't be defined on the enclosing element, for instance a `<a>` tag, but should be defined in their own element, for instance a `<span>`. To facilitate this, this proof-of-concept adds a template tag for icons. The template tag generates an accessible icon by default:
- When the icon is decorative only, it will be hidden from screenreaders (`aria-hidden`)
- When the icon has a (visible) label, the label will be read out loud by screenreaders
- When the icon doesn't have a visible label but is non-decorative, the text will be hidden, but will be read out loud by screenreaders. (`'class="sr-only"`)

By using a template tag, icon usage can be standardized as a re-usable 'component'. In this proof-of-concept we chose to use a separate template tag library, which can be expanded to include other UI components like buttons. By using a template tag it's also possible to change the interface icon-by-icon, which might be necessary since switching to separate `<span>`'s requires CSS changes.

### DIFF
--- a/wagtail/admin/templatetags/wagtailui_tags.py
+++ b/wagtail/admin/templatetags/wagtailui_tags.py
@@ -3,16 +3,15 @@ from django import template
 register = template.Library()
 
 
-@register.inclusion_tag("wagtailadmin/shared/icon.html",
-                        takes_context=False)
-def wagtail_icon(classnames, title='', show_label=True, decorative=True):
+@register.inclusion_tag("wagtailadmin/shared/wagtail_icon.html", takes_context=False)
+def wagtail_icon(name=None, classname='', title=None):
     """
-    Usage: {% wagtail_icon classnames title show_label decorative %}
+    Usage: {% wagtail_icon name="cogs" classname="icon--red" title="Settings" %}
 
-    The css classes necessary to show the icon are provided as a first argument.
-    The 'title' argument is optional, but recommended. When 'show_label' is True, the title is shown next to the icon.
-    When an icon is marked decorative, it is hidden for screenreaders (a11y). If 'decorative' is False, the title is
-    visible for screenreaders, regardless whether 'show_label' is True or not.
-
+    First load the tags with {% load wagtailui_tags %}
     """
-    return {'classnames': classnames, 'title': title, 'show_label': show_label, 'decorative': decorative}
+    return {
+        'name': name,
+        'classname': classname,
+        'title': title,
+    }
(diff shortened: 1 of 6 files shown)


thibaudcolas:
🎉 🎉 will try this right away

wagtail/admin/templatetags/wagtailui_tags.py
```diff
@@ -0,0 +1,18 @@
...
+register = template.Library()
+
+
+@register.inclusion_tag("wagtailadmin/shared/icon.html",
+                        takes_context=False)
+def wagtail_icon(classnames, title='', show_label=True, decorative=True):
+    """
+    Usage: {% wagtail_icon classnames title show_label decorative %}
+
+    The css classes necessary to show the icon are provided as a first argument.
```
tomdyson: css -> CSS

wagtail/admin/templatetags/wagtailui_tags.py
```diff
@@ -0,0 +1,18 @@
...
+    The 'title' argument is optional, but recommended. When 'show_label' is True, the title is shown next to the icon.
+    When an icon is marked decorative, it is hidden for screenreaders (a11y). If 'decorative' is False, the title is
```
tomdyson: "marked decorative" -> "marked as decorative"

SanderTuit:
Thanks @thibaudcolas and @tomdyson! 

We need to decide on a couple of things for this to be implemented.

1. Is a template tag like in this proof-of-concept the right way to implement icons in the admin?
2. There's an open issue on migrating the icon font to SVG (https://github.com/wagtail/wagtail/issues/1224). Given the fact that CSS changes are necessary for this implementation, would it be better to address these problems at the same time?
3. In the current proof-of-concept implementation we use the same `classnames` argument that is used for menu items for compatibility reasons. However the new React (icon) components use a different approach where `icon icon-` is automatically added to the icon name. Would it be preferable to use the same method in both places and sacrifice backwards compatibility?

thibaudcolas:
Thank you @SanderTuit, this is amazing. I'm disappointed I didn't have more time to review / merge / help with this during the sprint, but will definitely get back to it this week.

I think this template tag is great from the call site’s perspective – I'll be merging this as soon as possible so we can start using it in the existing templates. The actual icon implementation will change to align with that of React, but I'll take care of that. SVG re-implementation isn't as big of a priority as making this accessible.

For question 3 about `classnames` vs icon `name` – good question 🙂I'll try to use this template tag across Wagtail core and we'll see what breaks!

Issue #4162, opened 2018-01-02
Feature Request: Show collection in Document chooser modal


patta42:
Currently, the Document Chooser does not show the collection of the document. If you have multiple documents with the same title or do not remember the exact title or file name, the corresponding collection might provide another hint.

Currently, one has to check every possible collection manually, which is not very helpful if you maintain several hundred collections.

I would also appreciate this for the image chooser, but I think that is not so easy to implement there.

gasman:
Completed in #4854.

PR #4383, opened 2018-03-15
Added overview of collections in the Image/Document chooser templates

alejandrogarza:
According to request in https://github.com/wagtail/wagtail/issues/4162
*On the document chooser I added a new column with the name of the collection, so we can have a clear overview of where is the document stored.  (see screenshot)
*For the image chooser I thought adding the name would create too much clutter, so thought about adding it as a `title` attribute on the image container with the format `Collection » Image name` (Currently we don't have a `title` attribute)  (see screenshot)

image
image

### DIFF
--- a/wagtail/documents/templates/wagtaildocs/documents/list.html
+++ b/wagtail/documents/templates/wagtaildocs/documents/list.html
@@ -15,6 +15,7 @@
                 {% endif %}
             </th>
             <th>{% trans "File" %}</th>
+            <th>{% trans "Collection" %}</th>
             <th>
                 {% if not is_searching %}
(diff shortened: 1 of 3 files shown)


alejandrogarza:
Forgot to mention I tested this on OSX: Chrome/Safari/Firefox

gasman:
Thanks @alejandrogarza!

Just a little request: Currently, the 'collections' feature is designed to be completely hidden within the UI whenever there are no collections set up (or, for example, the user only has permission for a single collection - in this case there's no need for the user to even know that the concept of a "collection" exists, so we can simplify the user experience by making it invisible). Please can you update the templates so that the new column (and title prefix) is hidden in this case? I _think_ this can be done by adding `{% if collections %}` around the relevant elements - see for example: https://github.com/wagtail/wagtail/blob/d188576af8e0d091f774dd837bee073656aacb11/wagtail/images/templates/wagtailimages/chooser/chooser.html#L20-L22

gasman:
Completed in #4854.

PR #4384, opened 2018-03-15
Remove assert statements from Wagtail API

### LINKED ISSUE #4186: Remove assert statements from Wagtail API
There's a few places in Wagtail API where we used assert statements to check parameters. These statements are removed when Python optimisations are enabled (see https://github.com/rohe/pysaml2/issues/451) so we should replace these.

- https://github.com/wagtail/wagtail/blob/8b420b9abcfd7c4ede23f0b52fa60d81e640eb82/wagtail/api/v2/pagination.py#L16
- https://github.com/wagtail/wagtail/blob/8b420b9abcfd7c4ede23f0b52fa60d81e640eb82/wagtail/api/v2/pagination.py#L27
- https://github.com/wagtail/wagtail/blob/fd2753667b185e66d68fe9601dce6036a768419d/wagtail/api/v2/filters.py#L146
 - https://github.com/wagtail/wagtail/blob/fd2753667b185e66d68fe9601dce6036a768419d/wagtail/api/v2/filters.py#L193
 
None of these lead to security bugs. But it's not ideal that the behaviour changes depending on how you're running Python.

kcleong:
This PR fixes #4186:

> There's a few places in Wagtail API where we used assert statements to check parameters. These statements are removed when Python optimisations are enabled so we should replace these.

This PR is best reviewed commit by commit. The asserts have been removed and tests have been added for the filters which have been changed.

### DIFF
--- a/wagtail/api/v2/filters.py
+++ b/wagtail/api/v2/filters.py
@@ -143,10 +143,11 @@ class ChildOfFilter(BaseFilterBackend):
         if 'child_of' in request.GET:
             try:
                 parent_page_id = int(request.GET['child_of'])
-                assert parent_page_id >= 0
+                if parent_page_id < 0:
+                    raise ValueError()
 
                 parent_page = self.get_page_by_id(request, parent_page_id)
-            except (ValueError, AssertionError):
+            except ValueError:
                 if request.GET['child_of'] == 'root':
                     parent_page = self.get_root_page(request)
                 else:
--- a/wagtail/api/v2/filters.py
+++ b/wagtail/api/v2/filters.py
@@ -191,10 +191,11 @@ class DescendantOfFilter(BaseFilterBackend):
                 raise BadRequestError("filtering by descendant_of with child_of is not supported")
             try:
                 parent_page_id = int(request.GET['descendant_of'])
-                assert parent_page_id >= 0
+                if parent_page_id < 0:
+                    raise ValueError()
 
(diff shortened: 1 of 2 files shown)


kcleong:
The tests are failing due to missing pages on the CircleCI tests. In the screenshot below I'm printing the result from `Page.objects.all()`:

screen shot 2018-03-15 at 16 06 01

On my dev env two pages are present:

screen shot 2018-03-15 at 16 03 34

What is the correct solution to make those pages available in the test on CircleCI? I could create the pages in `setUp()` when the pages are not present. But this doesn't seem like a proper solution.

kaedroho:
Thanks! The removal of the assert statements look good to me!

I'm not sure what the benefit of unit testing those classes is, they don't seem to be specifically checking for any issue that would be caused by the assert statements and all of this code is already covered by the functional tests.

When I developed the Wagtail API, I aimed to create a very comprehensive set of functional tests and avoided unit testing the internal implementation (except for parts where it really makes sense to do so such as parsing the syntax of the ``fields`` parameter) which makes refactoring the internal code much easier.

kcleong:
> I'm not sure what the benefit of unit testing those classes is, they don't seem to be specifically checking for any issue that would be caused by the assert statements and all of this code is already covered by the functional tests.

I've removed the tests from this branch, I agree that the usefulness of these tests is questionable.

kaedroho:
Thanks!

PR #4390, opened 2018-03-16
Improve Domain Selection for Multiple Sites

codiebeulaine:
When two sites share one root page, get_url_parts should return the page in the domain that is in the request (if any), currently it does not do that.


wagtail/core/tests/test_page_model.py
```diff
@@ -291,6 +294,19 @@ def test_urls_with_multiple_sites(self):
...
+        request = HttpRequest()
+        request.site = second_events_site
+        self.assertEqual(
+            christmas_page.get_url_parts(request=request),
+            (second_events_site.id, 'http://second_events.example.com', '/christmas/')
+        )
+
+        request.site = events_site
+        self.assertEqual(
+            christmas_page.get_url_parts(),
```
BertrandBordage: You forgot to pass the request ;)
Also, please move this test of the first event site before the test of the second event site.

wagtail/core/models.py
```diff
@@ -152,15 +152,16 @@ def clean_fields(self, exclude=None):
...
+        first - used to translate url_paths into actual URLs with hostnames
         """
         result = cache.get('wagtail_site_root_paths')
 
         if result is None:
             result = [
                 (site.id, site.root_page.url_path, site.root_url)
-                for site in Site.objects.select_related('root_page').order_by('-root_page__url_path')
+                for site in Site.objects.select_related('root_page').order_by(
+                    '-root_page__url_path')
```
BertrandBordage: Please revert this change. I know the line was too long, but since this change has no code consequence, please keep it the way it was.

BertrandBordage:
One test is failing on Python 3.4, Django 1.11 & PostgreSQL.
This is weird, I restarted the job to find out whether it was a random, and it doesn’t seem so…

Needs an investigation.

BertrandBordage:
@codiebeulaine I found the cause of the error in one of the tests. It’s because of this query in `Site.get_site_root_paths()`:
```python
Site.objects.select_related('root_page').order_by('-root_page__url_path')
```

Good news, it’s easy to fix :smile:

It sorts by `url_path`, but in what you test, `url_path` is the same for multiple sites, so when we don’t specify the request to `Page.get_url_parts(…)`, one of the multiple sites is used with no preference, so it’s not guaranteed that the chosen one will be the default site or the first in alphabetic order.

I struggle to explain because it’s a bit complex, but changing the previous query to this should work:
```python
Site.objects.select_related('root_page').order_by('-root_page__url_path', 'is_default_site', 'hostname')
```

That way, here’s the logic we get to generate the page URL, after matching multiple possible sites:
- if `request.site` is defined and matches one of the possible sites, we use its domain
- if one of the possible sites is the default site, we use its domain
- otherwise, we use the first possible domain in alphabetic order

BertrandBordage:
Thanks Codie!

PR #4397, opened 2018-03-17
2591 localize verbose names

### LINKED ISSUE #2591: Unlocalizated model label Images
Hi guys, i can't remember if someone report this issue but i found that images model hasn't translatable model name:
seleccion_093
As you can see in the next picture, the other labels are translatables:
seleccion_094

In addition, i want suggest that you can localize tha app names, for example, in spanish: Imágenes Wagtail or Páginas Wagtail, etc. But, it's only a suggestion.

Thanks guys!

allcaps:
Fix #2591 

This makes me wonder: Don't we need localised verbose_name and verbose_name_plural on **all** Wagtail models?

I don't think this needs tests. This is default Django functionality. If tests are needed, I'm happy to add them. Just need a pointer on how to test translatable verbose_names. Language override is an option. But translations are missing at this moment.

The PO files need to be updated. Is that a part of the release process? Or should I add updated PO's to this PR?

* For Python changes: Have you added tests to cover the new/fixed behaviour? NO, not needed
* For new features: Has the documentation been updated accordingly? NO, not needed


SalahAdDin:
:+1:

gasman:
Just to confirm, .po files _do_ get updated as part of the release process, so no need to do that here. Thanks @allcaps!

PR #4398, opened 2018-03-17
4391 Add invalid password reset link error message

allcaps:
@jjanssen,

The Wagtail `password_reset_confirm` view is just a small wrapper around Django `password_reset_confirm`. It has `validlink` var in the context. I used it to display the error message, in the same way this is done in Django itself.

I'm not a native speaker. But shouldn't 'The password reset link was invalid' read 'The password reset link is invalid'? The link was and _still is_ invalid.

* For front-end changes: Did you test on all of Wagtail’s supported browsers? Not required
* For new features: Has the documentation been updated accordingly? Not required

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -11,6 +11,7 @@ Changelog
  * Remove outdated X-UA-Compatible meta from admin template  (Thibaud Colas)
  * Add JavaScript source maps in production build for packaged Wagtail (Thibaud Colas)
  * Removed `assert` statements from Wagtail API (Kim Chee Leong)
+ * Add invalid password reset link error message (Coen van der Kamp)
  * Fix: Status button on 'edit page' now links to the correct URL when live and draft slug differ (LB (Ben Johnston))
  * Fix: Image title text in the gallery and in the chooser now wraps for long filenames (LB (Ben Johnston), Luiz Boaretto)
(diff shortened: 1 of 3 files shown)


jjanssen:
Hi @allcaps,
Thank you for these changes! 

As far as the grammar goes. I actually took the reference that [Django](https://github.com/django/django/blob/ff05de760cc4ef4c7f188e163c722ec3bc1f0cbf/django/contrib/admin/templates/registration/password_reset_confirm.html#L38) uses to display this information in their template. So I'm assuming this phrasing would be correct. But maybe @tomdyson can shed a little light on this.

In addition, would it be possible to add a link to the forgot-password view? So the end-user doesn't hit a dead-end.

allcaps:
@jjanssen I added the link see screenshot.

I used the button markup like described in the style guide (a tag). However, the button padding is  different to the button at /admin/password_reset/ This password reset button is an input-tag (and is not conform the style guide).

We can not use the input or button-tags they need js or a form tag. A get-form will add a `?` to the target url. So I think the a-tag is best.

What to do about the styling? I won't like this ticket to explode into a authentication views redesign ;)

example
'

allcaps:
Dear @jjanssen,

I fixed the test and page title. Also, all the authentication view buttons look the same (input => button) and a small fix in the css. This is the result:

screen shot 2018-03-25 at 00 28 44
screen shot 2018-03-25 at 00 28 30

PR #4401, opened 2018-03-19
Make update_index command quiet when calling with "--verbosity=0".

BenSturmfels:
The "self.verbosity > 0" is as used in Django's built-in management commands such as "migrate".

### DIFF
--- a/wagtail/search/management/commands/update_index.py
+++ b/wagtail/search/management/commands/update_index.py
@@ -51,20 +51,24 @@ def group_models_by_index(backend, models):
 
 class Command(BaseCommand):
     def update_backend(self, backend_name, schema_only=False):
-        self.stdout.write("Updating backend: " + backend_name)
+        if self.verbosity > 0:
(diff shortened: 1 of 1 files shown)


BenSturmfels:
The background on this is that I run `update_index` from a cron.weekly shell script, so I get mail if there's any output from the command. I could potentially redirect stdout to /dev/null, since errors would go to stderr, but the change I've made is more in line with how Django's built-in management commands work.

BenSturmfels:
For what it's worth, it looks like the test failure is a failure of the test pipeline itself, not related to this code change.

thibaudcolas:
This looks like it would be very simple for someone to take on and make the changes suggested by Andy, so I have tagged as "good first issue" if someone wants to take over.

yashlamba:
I was looking for some beginner issues to work on, @thibaudcolas I can complete this if it's still required (which I guess would be?).

lb-:
Replaced by https://github.com/wagtail/wagtail/pull/9271

PR #4404, opened 2018-03-19
in a rich text field, only generate hyperlinks to pages that are live

mwesterhof:
as far as i understand it, this code should respect the publication status of a page. The code already tries to gracefully handle nonexistent pages, and it seems to me that the rich text filter should behave the same way for pages that exist, but aren't live. Of course, if this is in error, feel free to close this PR

### DIFF
--- a/wagtail/core/rich_text/pages.py
+++ b/wagtail/core/rich_text/pages.py
@@ -22,7 +22,7 @@ class PageLinkHandler:
     @staticmethod
     def expand_db_attributes(attrs):
         try:
-            page = Page.objects.get(id=attrs['id'])
+            page = Page.objects.get(id=attrs['id'], live=True)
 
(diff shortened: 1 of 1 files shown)


gasman:
Thanks @mwesterhof!
I can't really see any benefit from making this change, since there's going to be a broken link in the resulting HTML either way - the only difference is whether it includes a real-but-unavailable URL, or a blank `<a>` tag. I can think of one (very marginal) benefit to the current behaviour: if an editor creates several interlinked pages and publishes them all at once, there'll be a brief window of time where only one of those pages is live. If the page gets visited and cached in that period, it's better for it to be the version with a real URL, so that it'll be immediately usable once the other pages go live (rather than having a cached copy of the page containing just the blank `<a>` tag, which will be unusable until the cache expires).

Happy to consider other arguments and reopen if appropriate, though!

PR #4406, opened 2018-03-19
Fixed incorrect HTML in styleguide/base.html template.

atombrella:
Just a tiny trivial fix.


atombrella:
Oh, I didn't see that this was fixed in the most recent commit.

PR #4409, opened 2018-03-21
Adds a temporary docs banner to the Kickstarter project.

BertrandBordage:
This adds a banner to the docs during the Kickstarter project.
## It looks like this:
localhost_8000_
## And like that on mobile:
localhost_8000_ galaxy s iii


docs/_static/js/kickstarter.js
```diff
@@ -0,0 +1,74 @@
...
+    '      padding-top: 75px;' +
+    '    }' +
+    '  }' +
+    '</style>' +
+    '<a href="https://www.kickstarter.com/projects/noripyt/wagtails-first-hatch" class="kickstarter">' +
+    '  <span class="close">×</span>' +
+    '  <span>Please consider supporting ' +
+    '    <strong>NoriPyt’s Wagtail Kickstarter</strong>' +
+    '  </span>' +
+    '  <img src="_static/images/kickstarter.png" width="63" height="45" />' +
```
gasman: The relative path here causes a broken image on pages within subsections :-( It looks the existing CSS/JS includes in the template are set up to output the correct number of `../`s, so hopefully it's possible to hook into that mechanism. (I guess the 'nuclear option' is to use jquery to pull out the `href` of one of those existing tags, and rewrite the URL based on that...)

BertrandBordage: Ah, right, at first I used `/_static` which wasn’t working, and then I looked at the source and saw relative `/static` on the home page and made that mistake.

Given the size of the image, the best option is probably to make it a `data:`, don’t you think?

gasman: Sure, that sounds reasonable enough!

thibaudcolas: Should be `alt=""` so screen readers don't read the image `src` out loud.

BertrandBordage:
Thank you for reviewing, @gasman & @thibaudcolas!
I fixed both issues you pointed out.

docs/_static/js/kickstarter.js
```diff
@@ -0,0 +1,74 @@
...
...etwnPEYUJ3uYt7R5yvGp6mFMfjVvQ+GFdfylhJYIr8PrXFeDbwMbCOEkNHW98Xg90kOhJYz8JjnaK7Piyhf+U6G1/K2Nqf9pxb2AKcB64a/053/8XjG7KUUNTdoy7JASMkfigsRIe5SRDM11FV+IB6P+uW6jP05V5dEQe73v21ZS5C0zvvUz14AbwmfXHX98UTnTUkxzUk2zZj26wBp1hZpSrTA52kJhJ48Q9TS/9KieJrsEgutVBApLmeBlCdBQ7CNPz7+c6EeH7QBhQDZwD3gc++jqCkODJdXkKJxhVjL9HV2RLMx6CDuKZvA7DSkVOlGQkeEpkmKMExI6X8VleMnBJqDxeP0V0e+JFtytOdDsNRqwJsRbgd8CDhC/z/Oe/rbUWwvS7fkCA6VPgGlO7/APwFnAJaDMJ9FfgA+BDPAh9hCHko6OrZ3+EVvlh08pZH25ZlSGbVqbLqooUKV0YJzkFkZKYFSj+qR7ikK6XYbP00jfLRb63xN8Y4b0gvDvOeLzTCjuAFYGiW+BuLsBRRP0/jPZ3/3pfW3cmhNXffwQMAuyBiSaBwo+uzpl8YGmGYUdpclxNYXzc6qzIGP/5Yav7V4bLz5aFyA9XhkqfqgjpXRtjJLIv3kh0bbCRDG29ffLDhK1Rh3OW+FgLsEi3xKP3N/I9vi1BuvzZbWhRtmUkd5mgRbQBx5siRVfuJbr5ILLUT3TbYm2TJzZGWNv/GvLfQ/fI/Ow0DOhy8yRejwhW+j+IYIm+Tbcl+iLEabdJntcu9zeK9kCAKl2Z/gePCvm0LolvRzFbHWy+8du64onJuq0xjjjnVM+iH24d/Q8goOujQr6uc+KI2qogY89eoDZ9C8jUlZr69k7Dc0A58F7XzjEYU4SpYpn7xY8K+bMWmyXxLdGiW+ZPi2vE38BxiK7M2/j/b6omadf2AcYAKcB24E3gzkPWZ+W3tP7WR4X824rwLlN+s0cv8tI2eQuki/H6jK7SpVfHkFLuzevM79EL+BkwFLAH3IAIIBnFMhtVP09Newv1s4FUYMSjQb7BcFlXG220OCu6srn7ObXREv0QXan+wXja2XxuLoL5z4EkHdrd47oK3yd0Bc5PIl2IJ3SlXo8/OtW+wncpJq7bIN0I0oyMHfAT4MH/oprrrMPGe+KirsWw+vk/0fIykva7pTcAAAAASUVORK5CYII=" />' +
```
thibaudcolas: Should be an empty `alt=""` since the image is only presentational flavour.

BertrandBordage:
OK, I wasn't aware that empty alt was considered a good practice!
Updated the PR.

gasman:
Merged in 011a88fb077f3b9da9b2a980df63c12ef7a120d1 (master) / 814d66590274a88161da9b2b0364b353255e61da (2.0.x). However, I've just noticed an issue with the date check - `new Date(2018, 4, 17)` is actually 17th May, because months in Javascript are zero-based (brilliant design there...). Should this be changed to `new Date(2018, 3, 17)`?

BertrandBordage:
@gasman What the… Javascript…
OK, thanks for spotting that, I make a new commit in both branches to fix that.

BertrandBordage:
Done.

Issue #4396, opened 2018-03-17
more get_admin_display_title not honored


Speedy1991:
Same setup like #4353   with similar bug:
[here](https://github.com/wagtail/wagtail/blob/master/wagtail/admin/templates/wagtailadmin/pages/listing/_page_title_choose.html) and [here](https://github.com/wagtail/wagtail/blob/master/wagtail/admin/templates/wagtailadmin/pages/listing/_page_title_explore.html) and [here](https://github.com/wagtail/wagtail/blob/master/wagtail/admin/templates/wagtailadmin/pages/listing/_page_title_move.html)

Bug:
`page.get_admin_display_title`
Fix:
`page.specific.get_admin_display_title`

Added fixes to existing PR #4354

gasman:
Hi @Speedy1991 - are you sure these are needed? We already call `specific()` on the queryset, inside the view function (except for the 'reorder' view, where it's disabled for performance reasons):
https://github.com/wagtail/wagtail/blob/72e1491f393de6c579595433f4f962c9771ffc0f/wagtail/admin/views/pages.py#L85

(Calling `specific()` on a queryset is a bit more efficient than calling it on an individual page, as it means that multiple pages of the same type are combined into one query.)

Speedy1991:
well, i needed it in the reorder view, due to translated equal titles. E.g:
```
Github
Github
Github
```

I don't get different language codes (defined in `get_admin_display_title`) so i don't know which one is which.

With the fix i get something like this:

```
Github[en]
Github[de]
Github[es]
```

And i know what i just reordered

```
Github[de]
Github[es]
Github[en]
```

gasman:
Fixing this for the reorder view is covered by the existing ticket #4300 - have now proposed a partial fix (avoiding the performance issue of calling `specific()` on very large querysets) in #4410.

Issue #4300, opened 2018-02-17
get_admin_display_title ignored in ordered list display


robslotboom:
### Issue Summary

When choosing order by order in PageAdmin, get_admin_display_title is ignored.

### Technical details

* Wagtail version: 1.13.1
![default](https://user-images.githubusercontent.com/8074866/36340232-0aa4701e-13d9-11e8-84ed-584b53344b15.jpeg)
![ordered](https://user-images.githubusercontent.com/8074866/36340238-116968dc-13d9-11e8-8d17-c32b707347b0.jpeg)

loicteixeira:
This happens because the specific version of the pages isn't retrieved when in ordering mode because it is not paginated and could be expensive.

https://github.com/wagtail/wagtail/blob/5300b01fc4b797e355e1d08b6805b5ce5f8c45bd/wagtail/admin/views/pages.py#L77-L85

robslotboom:
I’ll wait and see,

Thanks

gasman:
Have proposed a partial fix in #4410 - adding the `specific()` when there are less than 100 items.

chosak:
The fix in #4410 has been merged and custom `get_admin_display_title` methods will now be used when ordering as long as fewer than 100 pages are being sorted (this limit was chosen to avoid performance issues on extremely large querysets).

Issue #882, opened 2014-12-20
Configurable default sort order in the Explorer?


coredumperror:
Considering that the Explorer represents the contents and layout of the main site menu, the fact that the default sort order is locked as `-latest_revision_created_at` is frustrating. It means that you require an extra step (clicking "Enable ordering of child pages") to see the Menu layout. This is not good UX for site editors who aren't intimately familiar with Wagtail's admin UI.

I'd be happy to implement this myself, but I don't really see anywhere in the admin UI where one can set "site configuration". Though I suppose it could be useful to have a default that's tied to a user's role (users who don't have permission to edit the menu don't need this sort order by default), so it could go in the Group editor.

davecranwell:
The "Notification preferences" area is available at `yoursite.com/admin/account/notification_preferences/`

In my mind's eye I'd see this renamed "User preferences" and contain email notification settings, explorer ordering settings and in future also the avatar field addition see #755 and #319

stepiiicz:
Wouldn't it be better to have a possibility to set the default sorting in the page model class, e.g. something like the following?
```python
class PeopleSection(Page):
    ...
    class Meta:
        verbose_name = u"Staff"
        children_ordering = "title"
    ...
```

Nowadays I run several wagtail sites that have different sections for which I would prefer different sorting settings. E.g. in sections with articles you usually want to sort by -last_created_at, but for sections like staff you'd prefer sorting by title etc.

coredumperror:
I like that idea!

davecranwell:
That's an interesting idea. Would you see it working alongside a "user preference" or would the two be incompatible?

Perhaps the logic could be to use the `childred_ordering` if specified, failing that, a user preference if specified, failing that, last modified date.

coredumperror:
Seems like a user preference, if set, should override the `children_ordering` setting on the model. But otherwise, that sounds good to me.

gasman:
I'm in favour of having the parent page's model specify the ordering, and in fact I'd take it one step further: specifying an order should *override the manual (menu) ordering*. In other words, it's not just an admin display preference: it governs the actual ordering of pages within the database (as defined by Treebeard's `path` field), and newly created (or renamed) pages would be inserted at their appropriate position.

[...]

PR #4410, opened 2018-03-21
Use specific page objects on reorder listings with <100 items

gasman:
Currently we call `specific()` on the page explorer listing, with the exception of the 'reorder' view because it is unpaginated and the performance hit of `specific()` might be unacceptable on very large querysets.

As an increasing number of people are making use of custom `get_admin_display_title` methods, they're running into this workaround, and in some cases proposing fixes that are less performant than simply adding the `specific()` - #4300, #4396.

This PR partially mitigates this by applying the `specific()` on listings of less than 100 pages. Hopefully this will fix things for the majority of users - manually ordering a section with >100 pages is going to be a pretty horrible experience either way (and is best addressed by #882...)

### DIFF
--- a/wagtail/admin/tests/test_pages_views.py
+++ b/wagtail/admin/tests/test_pages_views.py
@@ -266,6 +266,12 @@ class TestPageExplorer(TestCase, WagtailTestUtils):
         response = self.client.get(reverse('wagtailadmin_explore', args=(self.new_event.id, )))
         self.assertContains(response, 'New event (single event)')
 
+        # Reorder view should also use specific pages
+        # (provided there are <100 pages in the listing, as this may be a significant
+        # performance hit on larger listings)
+        response = self.client.get(reverse('wagtailadmin_explore', args=(self.root_page.id, )) + '?ordering=ord')
+        self.assertContains(response, 'New event (single event)')
+
     def test_parent_page_is_specific(self):
         response = self.client.get(reverse('wagtailadmin_explore', args=(self.child_page.id, )))
         self.assertEqual(response.status_code, 200)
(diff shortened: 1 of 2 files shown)


chosak:
@gasman if the manual reordering of >100 pages is in practice discouraged, and we don't expect users to encounter that use case, then what's the downside of always including `.specific()`?

It feels a bit inconsistent and unexpected to support one kind of behavior when sorting 99 pages and another when sorting 101, especially if that behavior is fairly obscure to the end user. I might prefer to enable this across the board for simplicity and consistency.

Alternatively, what about adding a setting that users can set to enable/disable this behavior across the board?  We could also consider adding a comment to the docs [here](http://docs.wagtail.io/en/v2.0/editor_manual/finding_your_way_around/the_explorer_page.html#reordering-pages) suggesting the manual sorting is best used on a small number of pages.

gasman:
@chosak The downside is that if the query takes too long or consumes too much memory, the page will crash entirely, which is a worse outcome than simply producing less-informative-than-desired output. A `specific()` query is liable to be significantly more resource-intensive than a plain query on `Page`, since it's retrieving the complete page content rather than just title/metadata.

Admittedly I haven't done any kind of benchmarking to determine when we're likely to hit server-killing levels, and 100 is just a number I plucked out of the air, but since this is a problem people are running into right now, I think it's best to implement a stopgap fix that solves the problem for a (hopefully) large proportion of those people rather than hold off for a more rigorous solution.

chosak:
@gasman thanks for the explanation. That makes sense.

chosak:
Merged in 7818edf with additional tests in 3bd3275.

PR #4415, opened 2018-03-22
Ensure breadcrumb respects custom get_admin_display_title methods

### LINKED ISSUE #4353: breadcrumbs don't honor overridden get_admin_display_title's

gasman:
Fixes #4353.

Have gone with a slightly revised approach to #4354 / #4368 - calling `specific()` on the queryset is marginally more efficient as it allows multiple pages of the same type to be fetched in the same query, and generally it's neater to have the `specific` call as early as possible in the workflow. Test added.

### DIFF
--- a/wagtail/admin/templatetags/wagtailadmin_tags.py
+++ b/wagtail/admin/templatetags/wagtailadmin_tags.py
@@ -60,7 +60,7 @@ def explorer_breadcrumb(context, page, include_self=False):
         return {'pages': Page.objects.none()}
 
     return {
-        'pages': page.get_ancestors(inclusive=include_self).descendant_of(cca, inclusive=True)
+        'pages': page.get_ancestors(inclusive=include_self).descendant_of(cca, inclusive=True).specific()
     }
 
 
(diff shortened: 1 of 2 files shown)


gasman:
Merged in 626418d348072c7e7f08ec1bbef05f7acf97c077

PR #4416, opened 2018-03-22
Add Python 2 check to `wagtail` command

### LINKED ISSUE #4327: Add warning for pip2 users or possibly change documentation
Congrats on version 2!

Since you are completely on Python3 now, I believe [the documentation](https://wagtail.io/developers/) should either instruct the user to use `pip3` or show a warning when `pip2` is used instead.

I'm not 100% sure if it's possible for you to show a warning for `pip2`, but I just thought I'd suggest it.

For some systems -- mostly Linux distributions -- the `pip` command defaults to using `pip2`.

It turns out my system, running on Solus Linux, defaults to pip2 instead of pip3. So, I unknowingly installed Wagtail using pip2. When I used the `wagtail start` command to create a site, it created this unobvious error:

```
  File "/usr/bin/wagtail", line 11, in <module>
    load_entry_point('wagtail==2.0', 'console_scripts', 'wagtail')()
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 572, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2755, in load_entry_point
    return ep.load()
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2408, in load
    return self.resolve()
  File "/usr/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2414, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/lib/python2.7/site-packages/wagtail/bin/wagtail.py", line 230
    print(line, end='')  # NOQA
                   ^
SyntaxError: invalid syntax
```

I fixed this issue by uninstalling wagtail and using `pip3 install wagtail` instead.

gasman:
Fixes #4327. Adds an explicit error message when running the `wagtail` command on Python <3.4:

    (wagtailpy2)vagrant@vagrant-ubuntu-trusty-32:~$ wagtail start foo
    This version of Wagtail requires Python 3.4 or above - you are running 2.7

Was hoping to add this to `./manage.py check` too, but for that to work we'd have to intercept Django's loading process in such a way that the check happens before loading any *.py files that are not syntactically valid Python 2, and I can't see a way of doing that reliably. In principle it looks like it ought to be possible to catch this at the point of installation too (see https://code.djangoproject.com/ticket/28878), but that looks like a bit of a can of worms - we'd need to ask a PyPI admin to retrospectively fix the metadata on the 2.0 release, and my feeling is that the possibility of screwing things up outweighs the benefit.

### DIFF
--- a/wagtail/bin/wagtail.py
+++ b/wagtail/bin/wagtail.py
@@ -8,6 +8,15 @@ from argparse import ArgumentParser
 from difflib import unified_diff
 
 from django.core.management import ManagementUtility
+from django.utils.six import print_  # required to avoid a syntax error on Py2 when using print(..., end='')
+
+
+CURRENT_PYTHON = sys.version_info[:2]
+REQUIRED_PYTHON = (3, 4)
+
+if CURRENT_PYTHON < REQUIRED_PYTHON:
+    sys.stderr.write("This version of Wagtail requires Python {}.{} or above - you are running {}.{}\n".format(*(REQUIRED_PYTHON + CURRENT_PYTHON)))
+    sys.exit(1)
 
 
 def pluralize(value, arg='s'):
@@ -227,7 +236,7 @@ class UpdateModulePaths(Command):
         with fileinput.FileInput(filename, inplace=True) as f:
             for original_line in f:
                 line = self._rewrite_line(original_line)
-                print(line, end='')  # NOQA
+                print_(line, end='')  # NOQA
                 if line != original_line:
                     change_count += 1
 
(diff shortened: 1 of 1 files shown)


gasman:
Merged in 96061ae301d5a6469c28faefb727420b4d71a66c (master) / da45a3e4cf03fce154a23bd6a7eac6853b83d6a5 (stable/2.0.x).

PR #4418, opened 2018-03-23
Added a simple scale filter to image_operations.

oliverseal:
### Feature: simplified `scale` filter on images
Often times (especially when using the wagtail API) there is need to get an image that isn't necessarily of a specific width or height, but rather of a specific scale. The most common use case would be supporting `` or media queries in CSS. The users of the CMS may have uploaded images of varying sizes, but at render-time we just want to be able to provide responsive images.

This new `scale` filter spec is meant to facilitate that.


gasman:
Thanks @oliverseal! All looks good to me - merged in 091ea75de7f95a8d3d539c1bf1722c8d99ad0697.

kaedroho:
@gasman I thought that we had a principle of not making editors decide the size the images that appear on the site? They just upload the largest file they got and Wagtail will resize based on what's in the template.

The use of this filter now means that editors must be careful of the size the image they upload, as it could be too large or too small. You could say that the ``original`` filter sets a precedent for this, but wasn't that introduced just so we can have a link to the original image in the admin?

I also think that, due to it's name, there's a risk it could be used incorrectly by newcomers to Wagtail.

I think we should add some validation in so that this filter can only be used after a filter that defines a size for the final image to be scaled. This would ensure that the final size is not dependant on the original size (except for aspect ratio).

To nicen it up for ``srcset``, we could allow the ``{% image %}`` tag to be passed an existing rendition. This would append an additional filter onto the filter the rendition used. This could be used to allow the ``scale`` filter to create an ``srcset``:

```html+Django
{% image myimage width-1000 as large %}
{% image large scale-50 as small %}
<img src="{{ regular.url }}" srcset="{{ small.url }} 1x, {{ large.url }} 2x" />
```

And if you really want the previous behaviour, you can always chain ``scale`` after ``original``: ``original scale-50``.

gasman:
@kaedroho I agree it doesn't quite fit with how I envisage image renditions being used, but I felt that it fits neatly enough into the existing set of filters, without encroaching on any existing functionality or best practice. I think there's a place for Wagtail to push its own opinions up to a point, but I wouldn't want to take that as far as actively policing developers' decisions when there's a direct way of achieving what they want.

Issue #4412, opened 2018-03-21
Menu icon overlaps with title in modeladmin on mobile


allcaps:
The menu icon overlaps with page title in model admin on mobile.

Steps to Reproduce:
1. Open a model admin.
2. Create a small 'mobile' viewport.
3. See the overlap.

Expect that there is a margin between menu icon and title.

Technical details: Wagtail 2 / Google Chrome

<img width="969" alt="overlap" src="https://user-images.githubusercontent.com/1969342/37736918-7fe117be-2d52-11e8-92ca-76c1fe7f15e0.png">

jjanssen:
Thanks for reporting! Remember seeing something similar while back: #3703

Dropping this here as a reference.

jjanssen:
Thanks! Merged with 3bcefe3

PR #4420, opened 2018-03-23
4412 Position modeladmin title

allcaps:
Refs #4412 

`header-title` will position the title, fixing the overlap.

The same classname is used in other wagtail admin pages.

overlap be gone


jjanssen:
Thanks! Merged with 3bcefe3

PR #4421, opened 2018-03-23
Add a new hook 'register_account_menu_item'

mvantellingen:
This new hook makes it easier for third party apps to add new buttons on the 'my account' page in the Wagtail admin. Existing buttons are converted to the new hooks to make the code consistent.

An real world use case is the wagtail-2fa project (https://github.com/labd/wagtail-2fa) for which we currently need to override the complete template

This PR needs a few tests before merging, but before putting time into that I would love to get some feedback on this.

### DIFF
--- a/docs/reference/hooks.rst
+++ b/docs/reference/hooks.rst
@@ -120,6 +120,33 @@ Hooks for building new areas of the admin interface (alongside pages, images, do
 ``url`` (optional)
   A URL to an index page that lists the objects being described.
 
+.. _register_account_menu_item:
+
+``register_account_menu_item``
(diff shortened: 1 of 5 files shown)


BertrandBordage:
(Re-running the test because the setup failed)

BertrandBordage:
Thank you for this addition @mvantellingen!
Much cleaner and always nice to always get more control as developers :slightly_smiling_face:

Could you please add a very short documentation there before I can merge it please? :wink:
http://docs.wagtail.io/en/latest/reference/hooks.html#admin-modules

mvantellingen:
done :-)

PR #4427, opened 2018-03-27
Fixed #4423 -- Fixed background color in docs css.

### LINKED ISSUE #4423: visual glitch on http://docs.wagtail.io/en/v2.0/
image
Only documentation for 2.0 is affected.

It seems that it can be solved by setting `background-color` on `body` instead of `html`.

sir-sigurd:


jjanssen:
Thanks for taking the time to create a PR for this. Merged with 63e824f

PR #4434, opened 2018-03-30
Optimize the querycount for the sitemap.xml page

mvantellingen:
By optionally passing the request object to Page.get_sitemap_urls() it
will now use the cached site root on the request object instead of
retrieving it for each call. This cuts the number of queries required
for a sitemap roughly in half.

### DIFF
--- a/wagtail/contrib/sitemaps/sitemap_generator.py
+++ b/wagtail/contrib/sitemaps/sitemap_generator.py
@@ -1,13 +1,19 @@
+import warnings
+
 from django.contrib.sitemaps import Sitemap as DjangoSitemap
+
+from wagtail.core.utils import accepts_kwarg
+from wagtail.utils.deprecation import RemovedInWagtail22Warning
 
 
 class Sitemap(DjangoSitemap):
 
-    def __init__(self, site=None):
+    def __init__(self, site=None, request=None):
         self.site = site
+        self.request = request
 
     def location(self, obj):
-        return obj.url
+        return obj.get_full_url(self.request)
 
     def lastmod(self, obj):
         # fall back on latest_revision_created_at if last_published_at is null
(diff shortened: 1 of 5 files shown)


BertrandBordage:
Very nice Michael! Could you please add some `assertNumQueries` to the existing tests so we can actually test this improvement? Thank you :)

mvantellingen:
@BertrandBordage added a unittest

wagtail/contrib/sitemaps/sitemap_generator.py
```diff
@@ -1,13 +1,19 @@
...
+from wagtail.core.utils import accepts_kwarg
+from wagtail.utils.deprecation import RemovedInWagtail22Warning
+
 
 class Sitemap(DjangoSitemap):
 
-    def __init__(self, site=None):
+    def __init__(self, site=None, request=None):
         self.site = site
+        self.request = request
```
BertrandBordage: In all uses done of `Sitemap`, the site passed is `request.site`. So we actually don’t need to pass `site` anymore.

wagtail/contrib/sitemaps/tests.py
```diff
@@ -48,21 +50,40 @@ def setUp(self):
...
+
+    def test_get_urls_with_request_site_cache(self):
+        request = RequestFactory().get('/sitemap.xml')
+        req_protocol = request.scheme
+        req_site = get_current_site(request)
+
+        sitemap = Sitemap(self.site, request)
+
+        with self.assertNumQueries(16):
+            urls = [url['location'] for url in sitemap.get_urls(1, req_site, req_protocol)]
```
BertrandBordage: In fact, this pull request only removes 2 SQL queries. This number doesn’t seem to increase with the number of pages, probably simply because site fetching uses a cache system.

Of course, with the previous comment (specifying `request` instead of both `request` and `site`), this PR doesn’t add complexity, so these 2 extra queries are still good to save.

gasman:
Merged in bad95cf37c51898a67d26d4ed92ddf54bca48e21 - thanks @mvantellingen!

PR #4435, opened 2018-04-02
Feature/footerblocks

Speedy1991:
I needed some extra buttons for translation ("Create Translation Drafts") and have to override the whole template (or at least the content block). It would be useful if there is just a small block to insert new actions like it is provided by this PR.

Maybe it's also useful for others.


gasman:
Merged in 878a5beecaea7d2c5d326f811407deebb0a9759f - thanks @Speedy1991!

PR #4436, opened 2018-04-02
[WIP] Add more Draftail extension APIs for advanced extensions

thibaudcolas:
Supersedes #4315.

This adds new extension APIs for Draftail which make it possible to build more advanced extensions.

- [Draftail controls](https://github.com/springload/draftail/#other-controls) – a poor name for anything and everything that extends Draftail but doesn't directly map to a rich text format (eg. a "character count" display, or a "clear all formatting" toolbar button). The plugins over at https://github.com/vixdigital/draftail-plugins are good examples.
- [Draft.js decorators](https://github.com/springload/draftail/#custom-text-decorators) – to build syntax highlighting, or spellcheck, or any other text pattern matching and highlighting.

### DIFF
--- a/client/src/components/Draftail/index.js
+++ b/client/src/components/Draftail/index.js
@@ -19,13 +19,25 @@ import EditorFallback from './EditorFallback/EditorFallback';
 const BR_ICON = 'M.436 633.471l296.897-296.898v241.823h616.586V94.117h109.517v593.796H297.333v242.456z';
 
 /**
- * Registry for client-side code of Draftail plugins.
+ * Registry for client-side code of Draftail extensions.
  */
 const PLUGINS = {};
+const DECORATORS = [];
+const CONTROLS = [];
 
 const registerPlugin = (plugin) => {
   PLUGINS[plugin.type] = plugin;
   return PLUGINS;
+};
+
+const registerDecorator = (decorator) => {
+  DECORATORS.push(decorator);
+  return DECORATORS;
+};
+
+const registerControl = (control) => {
+  CONTROLS.push(control);
+  return CONTROLS;
 };
 
 /**
@@ -112,6 +124,8 @@ const initEditor = (selector, options, currentScript) => {
         inlineStyles={inlineStyles.map(wrapWagtailIcon)}
         entityTypes={entityTypes}
         enableHorizontalRule={enableHorizontalRule}
+        decorators={DECORATORS}
+        controls={CONTROLS}
       />
     </EditorFallback>
   );
@@ -119,7 +133,11 @@ const initEditor = (selector, options, currentScript) => {
   ReactDOM.render(editor, editorWrapper);
 };
 
-export default {
+const draftail = {
   initEditor,
   registerPlugin,
+  registerDecorator,
+  registerControl,
 };
+
+export default draftail;
--- a/client/src/components/Draftail/index.test.js
+++ b/client/src/components/Draftail/index.test.js
@@ -133,6 +133,29 @@ describe('Draftail', () => {
     });
   });
 
+  describe('#registerDecorator', () => {
+    it('works', () => {
+      const decorator = {
+        component: () => {},
+        strategy: () => {},
+      };
+
+      expect(draftail.registerDecorator(decorator)).toEqual([
+        decorator,
+      ]);
+    });
+  });
+
+  describe('#registerControl', () => {
+    it('works', () => {
+      const control = () => {};
+
+      expect(draftail.registerControl(control)).toEqual([
+        control,
+      ]);
+    });
+  });
+
   it('#ModalWorkflowSource', () => expect(ModalWorkflowSource).toBeDefined());
   it('#Link', () => expect(Link).toBeDefined());
   it('#Document', () => expect(Document).toBeDefined());
--- a/docs/advanced_topics/customisation/extending_draftail.rst
+++ b/docs/advanced_topics/customisation/extending_draftail.rst
@@ -1,9 +1,9 @@
 Extending the Draftail Editor
 =============================
 
-Wagtail’s rich text editor is built with `Draftail <https://github.com/springload/draftail>`_, and its functionality can be extended through plugins.
+Wagtail’s rich text editor is built with `Draftail <https://github.com/springload/draftail>`_, and its formatting can be extended through plugins.
 
-Plugins come in three types:
+Formatting plugins come in three types:
 
 * Inline styles – To format a portion of a line, eg. ``bold``, ``italic``, ``monospace``.
 * Blocks – To indicate the structure of the content, eg. ``blockquote``, ``ol``.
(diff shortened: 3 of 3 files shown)


thibaudcolas:
Ok. I tried my best but I'm completely unable to decide what a good integration API would be to define `controls` and `decorators`. Before I describe the problem and potential solutions, here are the use cases those APIs are meant to cover:

- Any UI that calculates some metrics on field content, like the [VIX Digital plugins](https://github.com/vixdigital/draftail-plugins) for content length, reading level, sentiment analysis (`controls`)
- Any UI that needs to add extra buttons in the editor toolbar, like a "Clear all formatting" feature, or "Import from Word" (`controls` again)
- Custom spellcheck, syntax highlighting, and autosuggest / autocomplete (`decorators`)

Note that this potential API would only integrate what Draftail already offers – `controls` and `decorators` are defined by Draftail, what we need to figure out here is "just" the integration.

The problems I see is that:

- They are quite low-level APIs (see the description above), and as such quite specific to Draftail / Draft.js.
- They don't map well to "rich text" features in the sense of "formatting" – they are more meant for IMEs or arbitrary editor widgets.
- I don't see a big reason to have much server-side configuration for those, _except_ for the ability to only enable those extensions on specific fields only (and that seems like an important one).

For solutions, I could think of:

- Using the API that comes with this PR at the moment – controls and decorators are registered and enabled client-side, for all fields without distinction. That doesn't seem ideal, but it might be good enough just to try how those features work.
- Integrating this in the "rich text features" API. Semantically it feels a bit wrong to mix editor widgets and formatting together – they are very different, and aren't at the same abstraction level (eg. "h2, h3, import-from-word, ul, clear-formatting, hr, sentiment-analysis")
- Introducing a new server-side API, with a client-side component similar to how entities work – most of the definition is done server-side, only the strict minimum client-side component registration is done client-side (and that does nothing if the corresponding entity isn't declared on a field). Sub-classing widgets might be a way to go.

---

In addition, I think a lot of people would be really keen to have *any API*, even if not the right one, just so we can actually try those types of extensions within Wagtail before we decide how exactly they are meant to work. I've also considered we could release the API coming with this PR without documentation, or already-deprecated, or with a method prefix like `draftail.experimental_registerControl` / `draftail.unstable_registerControl`.

How does all of this sound?

loicteixeira:
Calling `controls` elements which don't offer UI interactions (like a character count) bother me a little bit, but I guess nothing stops those widget from having interactions (e.g. displays word count by default but cycle between character/word/paragraph count on click) so I guess that's fine. As a side note though, `control(s)` and `decorator(s)` are already used in other places in Draftail and it's becoming a bit confusing.

Anyway, regarding how we can register such elements, I think this PR (and the snippet in your comment above) is a good first step and could be merge. Loading JS with Wagtail's hook and then registering the element directly looks fine to me.

As for how to enable/disable those elements on a per field basis, I'm not sure either. I guess whether they are widgets or formatting options, they are both features of the editor. But assuming we want them separate, I don't think it would make sense to change the API for the RichTextField to accept additional arguments (e.g. `widgets` in addition to `features`) given its implementation should be editor agnostic. The alternative could be to have a per-situation basis instead of a per-field basis (because you probably don't need to enable/disable those meta widget on a per-field basis in the end?) with the `OPTIONS` of the widget definition, something like this:

```
WAGTAILADMIN_RICH_TEXT_EDITORS = {
    'default': {
        'WIDGET': 'wagtail.admin.rich_text.DraftailRichTextArea',
    },
    'enhanced': {
        'WIDGET': 'wagtail.admin.rich_text.DraftailRichTextArea',
        'OPTIONS': {
            'controls': ['sentiment-analysis', 'reading-time'],
            'decorators': ['spelling-check'],
        },
    },
}
```

Note that:

- `default` and `enhanced` could be flipped around (e.g. `default`=>`barebones` and `enhanced`=>`default`) depending on what you want the default field to have available without having to pass `editor=...` to the RichTextField/Block constructor every time.
- In the options, I use some strings similar to a feature name rather than the name of the JS component so `registerDecorator` and `registerControl` would need to be updated to accept 2 arguments, the "slug" name and the element itself.

thibaudcolas:
Thank you @loicteixeira, you make some really good points, even though I'm not sure how to address them just yet. And I had forgotten about the `OPTIONS` widget attribute – it's very relevant indeed.

It would be nice to hear more people reflect on this before we move forward, but you're happy with the current state of the PR so I'm also tempted to merge it as-is.

Do you have any thoughts on whether we would need to specify that those APIs are somewhat unproven / subject to future change? I wondered about saying it in the docs, or explicitly using some sort of prefix (eg. `window.draftail.experimental_registerDecorator`), but I'm not sure how useful that is.

loicteixeira:
I understand that we want people to test this new API to get feedback, but if we don't consider it stable yet, we probably shouldn't document it in the first place. Otherwise, we should release it as it is now and be ready to introduce deprecation warnings just like it is done in Python when we want to change it.

As a side note to my previous comment of updating the signature of the register functions to accept a "feature" name in addition to the JS object, I reckon we actually shouldn't and instead use a `type` property, just like `registerPlugin` does:

https://github.com/wagtail/wagtail/blob/06227a11ea97f36bd0df50c00864111bc81549e9/client/src/components/Draftail/index.js#L26-L29

thibaudcolas:
I agree with you @loicteixeira, which is very annoying because I'd also like to get this done with and what you suggest means more changes 😛. I also got to think a bit more about this over the last month. 

> As a side note to my previous comment of updating the signature of the register functions to accept a "feature" name in addition to the JS object, I reckon we actually shouldn't and instead use a type property, just like registerPlugin does:

This is what I agree with the most – control over what's active/inactive should stay on the Python side, which means adding some sort of type property, and if we do that then it probably makes sense to add it in the equivalent Draftail API too (which means changes to Draftail itself).

Regarding your other points,

- `control` is a terrible name, but it's still the best I could think of. Those things basically map to items in the toolbar, which can be any kind of UI, and don't even technically have to stay within the toolbar if someone wants to build something fancier.
- `decorator` is used elsewhere, but this instance is the most similar to the Draft.js definition of a decorator. Draft.js terminology isn't great to start with unfortunately.

I quite like what you proposed to change the widget definitions over the `RichTextField` APIs. It seems more semantically correct this way, perhaps even subclassing `DraftailRichTextArea` with those extensions (ideally mixins rather than class inheritance, but oh well).

SalahAdDin:
@thibaudcolas  What's about `draftail` icons? in the example`for example` `strikethrough` have a icon while in `wagtail` we have a simple `S`.

thibaudcolas:
@SalahAdDin that should already be supported 🙂

If you look at http://docs.wagtail.io/en/v2.0.1/advanced_topics/customisation/extending_draftail.html?highlight=draftail#extending-the-draftail-editor, change the strikethrough example like this:

```diff
- 'label': 'S',
+ 'icon': ['M100 100 H 900 V 900 H 100 Z'],
```

Here this will display a square, but the SVG is completely up to you as long as it's drawn in a 1024x1024 view.

SalahAdDin:
@thibaudcolas i'm sorry but, can you provide to me the tool that you use for get the `svg` codes for icons?

Thank you very much.

thibaudcolas:
Nice! Screenshot 1 looks fine, you’d just need to provide an SVG path that looks like what you want to get instead of the square. Personally I use https://icomoon.io.

Screenshot 2 — yep, you got it. And if you want to remove the `S` just remove the `'label': 'S'`.

SalahAdDin:
@thibau
I'm sorry, i'm using this code from `icomoon`: 
```
M32 16v2h-7.328c0.86 1.203 1.328 2.584 1.328 4 0 2.215-1.146 4.345-3.143 5.843-1.855 1.391-4.29 2.157-6.857 2.157s-5.002-0.766-6.857-2.157c-1.998-1.498-3.143-3.628-3.143-5.843h4c0 2.168 2.748 4 6 4s6-1.832 6-4c0-2.168-2.748-4-6-4h-16v-2h9.36c-0.073-0.052-0.146-0.104-0.217-0.157-1.998-1.498-3.143-3.628-3.143-5.843s1.146-4.345 3.143-5.843c1.855-1.391 4.29-2.157 6.857-2.157s5.002 0.766 6.857 2.157c1.997 1.498 3.143 3.628 3.143 5.843h-4c0-2.168-2.748-4-6-4s-6 1.832-6 4c0 2.168 2.748 4 6 4 2.468 0 4.814 0.709 6.64 2h9.36z
```
I'm getting this:
nofunciona
code

thibaudcolas:
@SalahAdDin this is because the SVG you took is drawn in a 32px box, not 1024. Let's take this conversation elsewhere from now on, it has nothing to do with this PR.

SalahAdDin:
@thibaudcolas, done, thank you. I don't know very much about `SVG` icons.

thibaudcolas:
This isn't cooked enough. Closing!

PR #4443, opened 2018-04-04
Fix inconsistent order of heading features with HalloJS

loicteixeira:
Because [only headings 2-4 are set as default features][1] and they all have the same `order` value (the don't specify one and use the default value), when registering the other headings as default features, they won't be ordered properly. Depending on whether your app or wagtail.core comes first in `INSTALLED_APPS`, you would get `h1, h5, h6, h2, h3, h4` or `h2, h3, h4, h1, h5, h6`.

This PR registers the heading plugins with increasing `order` value so the registration order as default value doesn't matter.

[1]: https://github.com/wagtail/wagtail/blob/7841f54fe873d128d21a30cb19752c7081ec7b1a/wagtail/core/wagtail_hooks.py#L49


kaedroho:
Looks good to me. Please could you resolve the isort issue then merge?

loicteixeira:
Will do thanks.

loicteixeira:
Merged in b12a7b08bd3a87f0dc26829cb730d2683cb71ca5.

Issue #4455, opened 2018-04-08
'Add user' button label is trimmed when translated using single quotes


benoitvogel:
### Issue Summary

The 'Add user' button has an incorrect label in the Roles selection page when language is set to French (might as well be relevant for other languages).

### Steps to Reproduce

On a fresh Wagtail setup with French language (LANGUAGE_CODE = 'fr-fr'), go to Settings (Paramètres) > Add a user (Ajouter un utilisateur) > Roles tab (Rôles).
On this page the _Add user_ button should be translated as _Ajouter l'utilisateur_ but the label reads _Ajouter l_

The PO file from Transifex has the right translated string in it but the problem lies in the {% trans %} template tag:
 `<input type="submit" value='{% trans "Add user" %}' class="button" />` 
Adding the `|force_escape` filter to this string converts the single quote to its HTML entity equivalent, thus rendering the button correctly by avoiding the value attribute to be closed prematurely.

### Technical details

* Python version: 3.4.3
* Django version: 2.0.4
* Wagtail version: 2.1a0 (official master)
* Browser version: Firefox 59

jjanssen:
Thanks! Merged with 9a21060

PR #4456, opened 2018-04-08
Fix 'Add user' button label escaping

benoitvogel:
Prevent wrong button labelling due to quotes in translated label
Fixes issue #4455

### DIFF
--- a/wagtail/users/templates/wagtailusers/users/create.html
+++ b/wagtail/users/templates/wagtailusers/users/create.html
@@ -40,7 +40,7 @@
                 <ul class="fields">
                     {% include "wagtailadmin/shared/field_as_li.html" with field=form.is_superuser %}
                     {% include "wagtailadmin/shared/field_as_li.html" with field=form.groups %}
-                    <li><input type="submit" value='{% trans "Add user" %}' class="button" /></li>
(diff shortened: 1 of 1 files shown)


wagtail/users/templates/wagtailusers/users/create.html
```diff
@@ -40,7 +40,7 @@
                 <ul class="fields">
                     {% include "wagtailadmin/shared/field_as_li.html" with field=form.is_superuser %}
                     {% include "wagtailadmin/shared/field_as_li.html" with field=form.groups %}
-                    <li><input type="submit" value='{% trans "Add user" %}' class="button" /></li>
+                    <li><input type="submit" value='{% trans "Add user" as add_user_button_label %}{{ add_user_button_label|force_escape }}' class="button" /></li>
```
jjanssen: Hi @benoitvogel, thank you for taking the time to contribute to Wagtail. 

Instead of escaping the translation value it would be more convenient to reconsider the input submit to be a button, for eg: `<button class="button">{% trans 'Add user' %}</button>`. This way we wouldn't need to escape the value and just render it as text within the element. By default the button will have the submit type.

benoitvogel:
Oh... You're absolutely right. Much better idea. Thanks.

jjanssen:
Thanks! Your PR is now merged with 9a21060

Issue #3728, opened 2017-07-20
Wagtail search `update_index` fails when indexing `ParentalManyToManyField`


mixxorz:
`update_index` fails when it encounters a `ParentalManyToManyField`. The error is from Django

```
ValueError: '{field with ParentalManyToManyField}' does not resolve to an item that supports prefetching - this is an invalid parameter to prefetch_related().
```

I think this is caused by [get_indexed_objects](https://github.com/wagtail/wagtail/blob/1943d5a83c8a8a68345cf48bee9b739ad4001aad/wagtail/wagtailsearch/index.py#L75). Since ParentaManyToManyField is a subclass of RelatedField, this code tries to do a prefetch. That obviously doesn't work because ParentalManyToManyField does not support prefetching. (?)

Unfortunately, I don't know modelcluster well enough to suggest a solution.

### Steps to Reproduce

1. Create a model with a ParentalManyToManyField
2. Add that field to search_fields
3. Run `update_index` management command

I'm not sure if this also happens on other search backends, but it happens when using the Postgres search backend.

### Workaround

Just republish all the pages. The indexing still works that way.

### Technical details

* Python version: 3.6.0
* Django version: 1.11.3
* Wagtail version: 1.11.1
* Search backend: Postgres Search

gasman:
Hi @mixxorz - thanks for the report.

Confirmed that this bug is still present on 2.0b1, testing on a new project with the following home/models.py:

    from __future__ import absolute_import, unicode_literals

    from django import forms
    from django.db import models
    from modelcluster.fields import ParentalManyToManyField

    from wagtail.admin.edit_handlers import FieldPanel
    from wagtail.core.models import Page
    from wagtail.snippets.models import register_snippet
    from wagtail.search import index


    @register_snippet
    class BlogCategory(models.Model):
        name = models.CharField(max_length=255)

        panels = [
            FieldPanel('name'),
        ]

        def __str__(self):
            return self.name

        class Meta:
            verbose_name_plural = 'blog categories'


    class HomePage(Page):
        categories = ParentalManyToManyField('BlogCategory', blank=True)

        content_panels = Page.content_panels + [
            FieldPanel('categories', widget=forms.CheckboxSelectMultiple),
        ]

        search_fields = Page.search_fields + [
            index.RelatedFields('categories', [
                index.SearchField('name')
            ])
        ]

timonweb:
I confirm this bug is still present on Wagtail 2.0.1, I've just caught it.

[...]

PR #4458, opened 2018-04-09
Fix Wagtail search  fails when indexing

timonweb:
Fixes issue #3728  by skipping `select_related` and `prefetch_related` for `ParentalManyToManyField`.


kaedroho:
Thanks!

Issue #4463, opened 2018-04-11
Can't add new icon


xiaohuanlin:
### Issue Summary

I followed every steps of [Wiki -->Adding new icons](https://github.com/wagtail/wagtail/wiki/Adding-new-icons) instructions. But the new icon didn't show in base.html.

### Steps to Reproduce

1. The icons I want to add are from Entypo+ set, just like this picture.
![image](https://user-images.githubusercontent.com/33860497/38591161-f1de4d76-3d67-11e8-835b-030010cc0bd4.png)
2. The hex code of these icons like this:
![image](https://user-images.githubusercontent.com/33860497/38591262-7f5cbe08-3d68-11e8-9e0f-a86cc377816d.png)
3. The code block I added to **_variables-icons.scss** just like this:
```  
    ...
    'table': '\f0ce',
    'logout': '\e901',
    'open-book': '\e902',
    'feather': '\e903',
    'mobile': '\e904',
    'youtube': '\e905',
    'location': '\e906',
);
```
4. But, the icons didn't show in the base.html:
![image](https://user-images.githubusercontent.com/33860497/38591343-ca5216e2-3d68-11e8-9500-f994cafc4848.png)

### Technical details

* Python version: 3.6.4
* Django version: django 2.0.4
* Wagtail version: wagtail 2.0
* Browser version: Chrome 65

gasman:
Hi @xiaohuanlin,

The steps on the wiki were just intended as a reference for the core Wagtail team, and weren't intended to be user-facing documentation - as such, they're likely to be out of date. I believe the missing step is to run: `npm run build` (note, you'll need a Wagtail development environment as detailed in http://docs.wagtail.io/en/v2.0.1/contributing/developing.html#setting-up-the-wagtail-codebase).

jjanssen:
In addition, if you would like to customize your Wagtail setup with icons, I'd suggest to take a look at [Wagtail FontAwesome](https://github.com/alexgleason/wagtailfontawesome)

xiaohuanlin:
Thanks very much. I really appreciate what you did. This question confused me quite long time.

Issue #2349, opened 2016-03-12
Wagtail icon set should be extended and improved


heymonkeyriot:
The icon set from the Wagtail font is fine but functional, and really quite limited
![image](https://cloud.githubusercontent.com/assets/11335309/13721963/d8512a90-e82d-11e5-8f09-03abb427424c.png)
#2334 raises a problem with the spinner, in #2342 I raised how the user icon could be enhanced. It's been raised [elsewhere](https://groups.google.com/forum/#!topic/wagtail/i3aUpnOkL6U) that the icon set should be extended, but can't find an existing issue here on github. @alexgleason released https://github.com/alexgleason/wagtailfontawesome, which helps, though requires the user to then obviously use FontAwesome, which brings in an additional dependency.

I think we can do better than this and think it'll enhance the general feel of the product to have some improved icons.

gasman:
+1. As per the discussion on #2285, I think the key to a really first-rate Wagtail site is for the implementer to get involved in UI decisions, including choosing icons that are relevant to that site's business objects and thus meaningful to editors. To that end, it would make sense for Wagtail to provide a comprehensive set of icons for implementers to use, rather than just ones that represent Wagtail's own concepts.

davecranwell:
Bear in mind https://github.com/torchbox/wagtail/issues/1224

We need to replace icon fonts with SVG sprites. Not only do they have odd results when failing to load, but they have odd results with screenreaders too, of which which some attempt to read out `content` properties.

See https://css-tricks.com/icon-fonts-vs-svg/ for more.

That's not to say implementers shouldn't be allowed to use icon fonts if they want to, but it does change how the proposed extension of the icons should occur for the core icon set.

heymonkeyriot:
Wasn't aware of #1224. Completely agree that it makes sense for the icons to be svgs rather than font files, for all the reasons put forward on that other ticket.

I think we could start concurrently thinking about the other icons (and amends) to the existing set that could be introduced whilst work's happening on #1224. 

@gasman, completely agree we should try and go beyond just the Wagtail concepts (or at least our understanding of how they might be representated) but also think we should try and limit how many icons we provide initially. Perhaps we can stagger the enhancement over the course of a number of releases so that we start releasing incremental benefit to the use rather than all at once when "finished"?

gasman:
[...]

PR #4464, opened 2018-04-11
SVG icons POC

allcaps:
This is a POC to componentise icons and make them screen-reader friendly using SVG and templates.
Related issues: #4463, #2349, #1224 

Goals:
- Wagtail icon font to svg
- Insert icon via template tag
- Way to hook in extra icons (your app or third party)
- Way to override some icons (customisation)
- Way to override all icons (icon packs)
- Documentation
- See if SVG icons need to be in the base template or can live in a external file/view.

This POC:
- Convert Wagtail icon font to SVG symbols
- Each symbol its own template. (easy overrides)
- Add a hook to register new icons
- Add templatetag inserting icons (DRY)
- Add CSS to position icons
- Add templatetag to load icons into the base template
- Made small preperations for a view serving a icon sprite (as alternative to including all icons in the base template)

TODO:
- Figure out if we want all icons each request OR that they should be served via a view (and set cache headers).
- JS should also use the use-tag.
- Provide an alternative for the most popular wagtail package: wagtailfontawesome
- Add correct styling for drawer menu.
- Remove icon fonts and icon CSS.
- ...

Test drive:
```
INSTALLED_APPS += [
    'wagtail.customapp',
    'wagtail.contrib.styleguide',
]
```
- See settings => styleguide => icons. There are two sets. The icon font icons and the new SVG icons.
- The wagtail.customapp inserts a custom icon and overrides an existing one.
- The main menu and drawer have SVG-icons. (Some menu-items are replaced by JS and have the old icon font).


allcaps:
@thibaud Ping! What do you think. Should we use a separate view for an icon sprite? OR is it okay to include all these icons in the base template?

thibaudcolas:
@allcaps I think this is super duper amazing! I don't have much time to review this right now unfortunately, but I'd love to.

From the perspective of "getting this merged", the thing that's the least likely to get in as-is is the `idify` hack. It's not reliable enough to auto-convert class names to an icon identifier, we'll have to painstackingly refactor all of the existing icon call sites to refer to icons by name only.

Sorry I can't get a more thoughtful review – here are other things that come to mind:

- Try to sync this work with https://github.com/wagtail/wagtail/pull/4381#pullrequestreview-110299442
- If you want to help with that thankless refactoring of all icon call sites, have a look at https://github.com/wagtail/wagtail/issues/4199#issuecomment-379581195.
- Is it really useful to be able to override all icons?

thibaudcolas:
Closing this in favor of #4821, which includes the same commits + some more!

PR #4467, opened 2018-04-11
Expose reusable client-side code to build Draftail extensions

thibaudcolas:
This PR exposes more of Wagtail's client-side code for reuse by third-party plugins to build client-side extensions, mostly for Draftail.

On all pages, it exposes:

- [x] Wagtail’s `Portal` component
- [x] Wagtail’s `Icon` component

On pages where Draftail editors are loaded, it also exposes:

- [x] The Draftail package
- [x] Wagtail’s `Tooltip` component
- [x] Wagtail’s `TooltipEntity` component
- [x] Wagtail’s `ModalWorkflowSource` component

---

If you want to test this – the unit tests cover it all, but you can go in the browser's console and try to see if there are values for the different attributes on the `window` global (eg `window.Draftail.BLOCK_TYPE`).

Edit: I canceled the Travis build manually.

### DIFF
--- a/client/tests/stubs.js
+++ b/client/tests/stubs.js
@@ -46,6 +46,8 @@ global.wagtailConfig = {
 
 global.wagtailVersion = '1.6a1';
 
+global.wagtail = {};
+
 global.chooserUrls = {
   documentChooser: '/admin/documents/chooser/',
   emailLinkChooser: '/admin/choose-email-link/',
--- a/wagtail/admin/static_src/wagtailadmin/app/draftail.entry.js
+++ b/wagtail/admin/static_src/wagtailadmin/app/draftail.entry.js
@@ -1,6 +1,5 @@
 import * as Draftail from 'draftail';
 import draftail, {
-  ModalWorkflowSource,
   Link,
   Document,
   ImageBlock,
@@ -20,22 +19,22 @@ window.draftail = draftail;
 const plugins = [
   {
     type: 'DOCUMENT',
-    source: ModalWorkflowSource,
+    source: draftail.ModalWorkflowSource,
     decorator: Document,
   },
   {
     type: 'LINK',
-    source: ModalWorkflowSource,
+    source: draftail.ModalWorkflowSource,
     decorator: Link,
   },
   {
     type: 'IMAGE',
-    source: ModalWorkflowSource,
+    source: draftail.ModalWorkflowSource,
     block: ImageBlock,
   },
   {
     type: 'EMBED',
-    source: ModalWorkflowSource,
+    source: draftail.ModalWorkflowSource,
     block: EmbedBlock,
   },
 ];
--- a/wagtail/admin/static_src/wagtailadmin/app/draftail.entry.js
+++ b/wagtail/admin/static_src/wagtailadmin/app/draftail.entry.js
@@ -1,3 +1,4 @@
+import * as Draftail from 'draftail';
 import draftail, {
   ModalWorkflowSource,
   Link,
@@ -10,6 +11,8 @@ import draftail, {
  * Entry point loaded when the Draftail editor is in use.
  */
 
+ // Expose Draftail package as a global.
+window.Draftail = Draftail;
 // Expose module as a global.
 window.draftail = draftail;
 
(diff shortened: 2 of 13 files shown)


wagtail/admin/static_src/wagtailadmin/app/draftail.entry.js
```diff
@@ -10,29 +10,31 @@ import draftail, {
  * Entry point loaded when the Draftail editor is in use.
  */
 
+ // Expose Draftail package as a global.
+window.Draftail = Draftail;
```
loicteixeira: `Draftail` vs `draftail` is going to be rather confusing IMO.

Also it might accidentally expose some private APIs which people will start using, forcing us to maintain them and ensure backward compatibility. Maybe not such a strong argument since it's already the case in Python (we can't stop a user to import a private API but we still don't ensure backward compatibility).

thibaudcolas: I agree it's a potential source of confusion but I couldn't find an alternative of similar simplicity I found good enough. For a more long-term solution I would like to do what I highlighted in https://github.com/wagtail/wagtail/issues/2315#issuecomment-360841058 but it would take "a couple of days" I haven't been able to find yet.

thibaudcolas: For the potential private APIs / API breaking changes, [Draftail the package's public API](https://github.com/springload/draftail/blob/master/lib/index.test.js) is all documented – it might unintentionally expose some undocumented / private API, but I think this is unlikely. I tried to keep the public API as small as possible to limit this risk (only `DraftailEditor` has a big API surface, and it's already exposed outside this PR with the rich text features).

client/tests/stubs.js
```diff
@@ -46,6 +46,10 @@ global.wagtailConfig = {
 
 global.wagtailVersion = '1.6a1';
 
+global.wagtail = {
+  ui: {},
```
jjanssen: We're exposing `ui` in the stubs, but it looks like we're not using this anywhere?

wagtail/admin/static_src/wagtailadmin/app/wagtailadmin.entry.js
```diff
@@ -1,7 +1,11 @@
-import { initExplorer } from 'wagtail-client';
+import { initExplorer, Icon, Portal } from 'wagtail-client';
+
+// Expose components as globals for third-party reuse.
+window.wagtail.Icon = Icon;
+window.wagtail.Portal = Portal;
```
jjanssen: I'm not completely sure about the naming convention for the long-term if were to expose `utils` or other helpers. It would probably make more sense to incorporate `components` into this.

client/tests/stubs.js
thibaudcolas: No – like for `wagtailVersion`, this is just trying to replicate the global variables set elsewhere. We could remove this if you think it is confusing since it is unused?

wagtail/admin/static_src/wagtailadmin/app/wagtailadmin.entry.js
thibaudcolas: Do you mean as `window.wagtail.components.Icon`, and same for Portal? That seems reasonable, but I also like the idea of mimicking what a `wagtail` package would expose – you would do `import { Icon } from 'wagtail';`, without an intermediate `component` object.

jjanssen: Yeah exactly like that 😄 But I'm agreeing with you on the mimicking part for the long-haul 👍

client/tests/stubs.js
jjanssen: Maybe if there is no reason to replicate the `ui` bit then just it leave it be an empty dictionary?

thibaudcolas: 👍 check

wagtail/admin/static_src/wagtailadmin/app/wagtailadmin.entry.js
thibaudcolas: 👌 done. We'll see about the long haul when there actually is a package 😛

PR #4468, opened 2018-04-12
Images: handle all not found errors in get_file_size

### LINKED ISSUE #2166: Missing image is raising 500 on s3
In `wagtailimages/models.py` catching the OSError is not enough for non-existing images for S3 backed CDNs.

Namely the `get_file_size` method digs into s3boto storage and raises an Attribute Error here: https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto.py#L473

It's arguably an inexpressive error to raise here, however, is catching that error an option? 

This _Work on my machine_™ (only difference is the added catching of the AttributeError):

``` python
    def get_file_size(self):
        if self.file_size is None:
            try:
                self.file_size = self.file.size
            except (OSError, AttributeError):
                # File doesn't exist
                return

            self.save(update_fields=['file_size'])

        return self.file_size
```

plumdog:
To fix #2166

### DIFF
--- a/wagtail/images/models.py
+++ b/wagtail/images/models.py
@@ -97,10 +97,15 @@ class AbstractImage(CollectionMember, index.Indexed, models.Model):
     def get_file_size(self):
         if self.file_size is None:
             try:
-                self.file_size = self.file.size
-            except OSError:
-                # File doesn't exist
+                size = self.file.size
+            except Exception:
+                # File not found
+                #
+                # Have to catch everything, because the exception
+                # depends on the file subclass, and therefore the
+                # storage being used.
                 return
+            self.file_size = size
 
             self.save(update_fields=['file_size'])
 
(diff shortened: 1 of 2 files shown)


chosak:
Thanks @plumdog! Committed (with some additional unit tests) in  24712b4.

wagtail/images/models.py
```diff
@@ -97,10 +97,15 @@ def is_stored_locally(self):
     def get_file_size(self):
         if self.file_size is None:
             try:
-                self.file_size = self.file.size
-            except OSError:
-                # File doesn't exist
-                return
+                size = self.file.size
+            except Exception:
```
kaedroho: When doing catch-all except blocks like this, we must add the following block just before in order for debugging tools to work correctly:

```python
except (KeyboardInterrupt, SystemExit):
    raise
except:
    ...
```

PR #4480, opened 2018-04-18
Fix 500 errors when indexing fails on models with non-ASCII characters in their representations

Aramgutang:
When adding a model instance to the search index fails for whatever reason, Wagtail logs an exception, constructing the message using `u'%r' % instance` (the `u` is implied by `from __future__ import unicode_literals`). This throws a `UnicodeDecodeError` when the string representation of the instance contains non-ASCII characters.

For example, take an Image instance called `image` with a title of "à":
```
>>> unicode(image)
u'\xe0'
>>> repr(image)
'<Image: \xc3\xa0>'
```
This cannot be substituted into a `unicode` string using `%r`:
```
>>> '%r' % image
'<Image: \xc3\xa0>'
>>> u'%r' % image
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3...
```
Because it's the equivalent of:
```
>>> u'%s' % '\xc3\xa0'
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3..
```
This change fixes the issue while maintaing the `repr`-like representation by decoding the `repr`-generated string using UTF-8 before substituting it using `%s`:
```
>>> u'%s' % unicode(repr(image), 'utf-8')
u'<Image: \xe0>'
```

I've checked other instances of `%r` usage in the project, and none of them have this issue, since they dont try to substitute a model instance using `%r`.

Also note that this issue does not occur with Python 3, so it's not something that should be merged into the 2.x branch.

### DIFF
--- a/wagtail/wagtailsearch/index.py
+++ b/wagtail/wagtailsearch/index.py
@@ -152,7 +152,8 @@ def insert_or_update_object(instance):
                 # Catch and log all errors
                 logger.exception(
                     "Exception raised while adding %s into the '%s' search backend",
-                    six.text_type(repr(indexed_instance), 'utf-8'), backend_name
+                    unicode(repr(indexed_instance), 'utf-8') if six.PY2 else repr(indexed_instance),
+                    backend_name,
                 )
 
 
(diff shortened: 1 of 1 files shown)


Aramgutang:
The test failure above is due to NPM package installation issues, unrelated to the change in the commit.

gasman:
Thanks @Aramgutang! Merged in cfe2b4a6a2b36c6815b492fbefdcb3bba5822a36 (for 1.12.x) and 0f314e7a1228231db26818e2fbb6c71b99cf816e (for 1.13.x).

PR #4481, opened 2018-04-18
New UI for listing object uses, including in RichText and StreamField.

BertrandBordage:
This pull request replaces #3961.

The heart of this pull request is the same, except that it was rebuild from current master with all the rich text editor API changes. It is split in 5 comprehensible commits. To avoid mistakes, I did all the rebasing work manually, by coding again everything except the new collectors.py file.

I took the decision to change a bit the rich text API for two reasons:
- It was missing some critical features for this PR to work:
  - The link between rich text handlers and models
  - There was no unified way differentiating between front-end & editor HMTL generation. At first sight, it seemed to be `expand_db_attributes` for editor HTML & a top-level function for generation front-end HTML. However, `Page` link doesn’t follow this pattern, and I found some name mistakes in the tests due to the confusion introduced by this unclear architecture
- As a site developer using custom editor features, I’m lost in the current API, so I thought it wouldn’t hurt to start introducing some unification with base classes such as `LinkHandler`, especially since we expect people to reference the link between their custom features and the related models, for the collectors introduced in this PR.

And of course, it still looks the same as in #3961. It might be improved, but that’s enough for now given that this is already an important enhancement compared to the current state where we have nothing except the usage count, and 500 errors on protected items deletion.

image


gasman:
Hi @BertrandBordage,
I'm afraid the rich text API changes here are taking things in the opposite direction to what I'd planned...

Prior to Wagtail 2.0, `PageLinkHandler` (and `DocumentLinkHandler` etc) was responsible for the conversions from database-HTML representation to front-end HTML _and_ editor-HTML representation, and those things shared a lot of code (such as the `expand_db_html` method with a `for_editor` flag). However, the editor-HTML representation is specific to hallo.js, and so that whole code path is now semi-deprecated as of Wagtail 2.0. I made a start on splitting those up - hence the new top-level `page_linktype_handler` method for the front-end conversion - but I didn't manage to take that as far as I'd hoped for the 2.0 release. (Ideally, the code now residing in `PageLinkHandler` / `DocumentLinkHandler` etc would be moved somewhere under `wagtail.admin.rich_text.converters.editor_html` where we can forget about it forever... except inevitably it's not as easy as that, because `wagtail.admin` isn't 'allowed' to know about documents/images...)

I haven't looked at your subsequent commits yet, but I guess you need an object to attach the "find entities of this type" method to. `PageLinkHandler` probably looks like the most suitable candidate for that in the current codebase, but it really isn't :-) We should create a new kind of object to encapsulate the non-editor-specific logic of rich text entities: namely, the front-end HTML conversion and the new "find entities of this type" method. I'll have a ponder this weekend to see how we might perform some minimal refactoring, so that we can introduce this new object without making an even bigger mess :-)

BertrandBordage:
@gasman: Indeed, I did this because I needed a system to fetch models linked to an `embedtype` or `linktype`, fetch the object from the stored HTML attribute. This system has to be created by inheriting classes, because the editors are extendable and devs might want to create new `embedtype`s or `linktype`s. I used the HTML representation simply because… It’s what’s stored in the database, on both editors. I don’t know if you plan to change the way rich text data is stored in the database in future releases, but I assumed it would not change because that would be a dramatic source of migration issues.

I also started bringing back elements together because either we want to create an extension for Hallo or Draftail, most things are the same: the related models, the way we fetch database objects from attributes, the way we select elements in the DOM, the name of the button, etc. To me, this is going exactly the right way because we want to make a registering system that devs can extend with a clear API. The day Hallo goes away, just one or two methods would have to be removed from the registering system. Today, sorry to say it roughly, but it’s messy and I had a hard figuring out lots of details, and some still don’t make sense to me, like the class name convention for lists and dicts like `EditorHTMLEmbedConversionRule`.

I didn’t put all Draftail `linktype` and `embedtype` functions & objects in that “unified” registering system to avoid doing to much at once: I just stuck to modifying what I absolutely needed.

And to conclude, I didn’t really have a choice: I couldn’t base my work on the Draftail extension classes because they don’t exist, and because data is stored as Hallo HTML, not as Draft.js JSON. I didn’t want to introduce a new registration as well, since we obviously don’t want a third system of registering extensions in addition of Hallo’s & Draftail’s ones. Devs would forget to use it in addition of the Draftail & Hallo systems, so the uses UI wouldn’t work well with custom rich text extensions.

So how should it work? We can also use it as is and improve/change it later when we drop Hallo. I’m perfectly willing to do so.

gasman:
Have now done some refactoring in https://github.com/gasman/wagtail/commits/cleanup/richtext-converter-refactor, so that the `wagtail.*.rich_text` modules now export one handler class which is the obvious correct one to extend here. Will try to rebase this PR on top of that.

I acknowledge that the rich text handling has some severe "bus factor" issues right now... there are some design decisions (e.g. my desire to keep the dbHTML -> frontendHTML and dbHTML <-> editorHTML logic separated) that came from being immersed in the rich text logic for several weeks, but might not be so obvious in the resulting code. We should certainly review (and document!) those decisions at some point... however, given that (I believe) it's a fairly tangential part of this PR, and I don't want to hold up the progress of this feature into 2.1, I think it's best to say "just trust me on this" for now :-)

wagtail/embeds/rich_text.py
```diff
@@ -3,55 +3,50 @@
...
-    Given a dict of attributes from the <embed> tag, return the real HTML
-    representation for use on the front-end.
-    """
-    return format.embed_to_frontend_html(attrs['url'])
+class EmbedHandler(LinkHandler):
+    link_type = 'media'
 
+    @staticmethod
+    def get_model():
+        return Embed
```
gasman: I don't think it's logically correct to treat EmbedHandler as being associated with the `Embed` model - the `Embed` model is more like a cache, and deleting records from that table has no consequence on rich text fields that reference that URL. (Looking at it more abstractly: there's no reason that a linktype/embedtype _has_ to be associated with a model; it's just rewriting one kind of HTML tag into another.)

Happily, I don't think there's any negative consequence from simply returning `None` here instead.

wagtail/core/collectors.py
```diff
@@ -0,0 +1,379 @@
...
+            values = []
+            for obj in searched_objects:
+                for model, handler in handlers.items():
+                    if isinstance(obj, model):
+                        k, v = handler.get_id_pair_from_instance(obj)
+                        values.append('%s="%s"' % (k, re.escape(str(v))))
+            params['val'] = (values[0] if len(values) == 1
+                             else r'(%s)' % r'|'.join(values))
+        else:
+            pattern = r'<(a|embed) %(type)s[^>]*>'
```
gasman: From what I can see, this case exists to serve a "hidden feature" of `find_objects` where passing an empty object list returns all objects of any type referenced within rich text. What's the motivation for this feature? It seems out of place here: it's listing objects rather than finding them; it doesn't match the logical expected behaviour of find_object on an empty list; and it requires quite a bit of special-case code to make it work (and, in particular, conflicts with my earlier assertion that link/embed handlers don't have to be associated with a model)...

wagtail/embeds/rich_text.py
BertrandBordage: I made this for more than just deleting. The idea is to have a way, through collectors, to find all references of a single object. This gives a lot of possibilities, including the possibility to create a task to clean the embeds from time to time. If a reference to an embed is removed, the object still sits there in the database, useless. On huge sites, that’ll be good to have a command to remove useless embeds from time to time, hence the system to find references for it in rich text & stream fields.

Indeed, linktype/embedtypes can do something without a model, but this is not the case in all the linktype/embedtypes provided by Wagtail.

wagtail/core/collectors.py
BertrandBordage: Yes, this is indeed an unused feature for now. At first, I started with doing that wildcard search only if `searched_objects is None`.
If I remember well, I changed my mind to make a cleaner API using `.find_objects(*searched_objects)` instead of `.find_objects(searched_objects=None)`. I’m sure I thought “if you search for one or more objects, it searches them, otherwise it searches for every object”. It must have been a brain fart, as I clearly forgot that sometimes we just pass a tuple or list directly without checking whether an item exists or not.

The good news is that it cannot have any impact on the views added, but that’s definitely worth changing. I think that we should just modify line 79 with `if searched_objects is None` and reverse the condition, then modify `find_objects`, for example by adding a `find_all` boolean that could only be `True` if there’s no `searched_objects`. Or add a `_find_objects(searched_objects)` method with all the logic from `find_objects`, and create two methods `find_objects(*searched_objects)` and `find_all_objects()` (that last one passing `None` to `_find_objects`).

What do you think?

gasman: Having two separate methods for `find_objects` and `find_all_objects` definitely makes more sense to me - they're functionally very different things. I think `get_pattern_for_objects` and `get_handlers` can also be split up in the same way, to simplify the logic.

wagtail/embeds/rich_text.py
gasman: If you're using it for garbage-collecting the embeds table, then calling `get_embed` from `get_instance` will do exactly the wrong thing... for any embed not found in the table, it will make a round-trip to the embed provider and create a new record.

gasman: OK, fair point! On a related note, I think we could optimise things further by having the Link/EmbedHandler classes implement a method to check "do these tag attributes correspond to one of these model instances", so that we're not uselessly instantiating every object we encounter in rich text (only to either throw it away or match it to an object we already have). But at that point I'm gift-horsing ;-)

wagtail/core/rich_text/feature_registry.py
```diff
@@ -96,3 +99,91 @@ def get_converter_rule(self, converter_name, feature_name):
...
+    def get_model():
+        raise NotImplementedError
+
+    @classmethod
+    def get_instance(cls, attrs: dict) -> Model:
+        model = cls.get_model()
+        try:
+            return model._default_manager.get(id=attrs['id'])
+        except model.DoesNotExist:
+            pass
```
gasman: @BertrandBordage Any objection to leaving the `DoesNotExist` exception uncaught here, to be handled by the caller? I think that makes it a bit more idiomatic.

BertrandBordage: You mean put it in `find_objects_in_rich_text`? Yeah, that would be valid, I can’t remember a reason why it would be bad.

gasman:
Rebased at https://github.com/gasman/wagtail/tree/BertrandBordage-new-uses-ui-rebase with the proposed changes (not re-combining the frontendHTML and editorHTML handlers; adding separate methods for `find_all_objects`) and some additional cleanup (reorganising the removal of `USAGE_COUNT` so tests pass after every commit). I still need to finish reviewing, but hopefully the remaining bits should be the more straightforward ones (dealing with refactoring and points of integration is always more fiddly than new view code)...

BertrandBordage:
Great @gasman :)
So it’s almost there :D Thanks a lot for your review and the adjustment of the API! It’s better you did it, it’s faster than if I did another iteration 80% compliant with what you had in mind ;)

wagtail/core/collectors.py
```diff
@@ -0,0 +1,379 @@
...
+    def __init__(self, model, using=DEFAULT_DB_ALIAS):
+        self.model = model
+        self.using = using
+        self.fields = [field for field in model._meta.fields
+                       if isinstance(field, StreamField)]
+        self.field_collectors = [StreamFieldCollector(field)
+                                 for field in self.fields]
+
+    def prepare_value(self, value):
+        if isinstance(value, Model):
```
gasman: (Reposting this comment as I made it against a commit on my own branch rather than this PR, and not sure if you got the notification, @BertrandBordage...)

Do we ever call `find_objects` with values that aren't model instances? If so, it looks like we're missing tests for these cases, and we should also consider renaming `find_objects` to `find_values`. Personally I'd prefer to remove this functionality if there's no immediate need for it, though - it seems like feature creep, and adds complexity that we could do without. (I'm not entirely clear on why `"` and `|` characters in strings need to be double-escaped, or why we can't replace most of this method with `json.dumps`, and if we can avoid going down that rabbit-hole, we should :-) )

BertrandBordage: Right, so would you be OK with you if that simplification consisted in just removing `prepare_value` and just systematically apply `json.dumps(v.pk).replace('|', r'\|')` in `find_objects`?

Sorry I don’t commit anymore on this branch, I’m not quite sure what to do now that you’ve started your own ^^

gasman: Yep, I'll go ahead and remove `prepare_value` and any other code paths that become redundant as a result (I think there are a few `if isinstance(value, Model):` switches elsewhere).

I still don't understand why you're manually escaping `|` _in addition to_ calling `re.escape`. Again I'm struggling with the lack of docstring here - is the return value of `prepare_value` intended to be a literal string to match within the json (in which case it's the caller's responsibility to escape that literal string appropriately for use in a regexp, or whatever string-matching mechanism it decides to use), or a regexp pattern (in which case we should call `re.replace` on the literal string, since `|` is not the only character that has special meaning)?

BertrandBordage: Oh, you’re right, the `|` escaping must be a mistake given that I added an `re.escape`. The only purpose of this method was to prepare the value before injecting it in `stream_structure_pattern`, so it had to be escaped so it’s considered as a literal by `stream_structure_pattern`.

gasman:
Very reluctantly bumping this to 2.2 - unfortunately the reviewing work has proceeded slower than I'd hoped, and I don't think I'll be able to complete it without causing excessive delays to the 2.1 release (I'm only in for two days next week). To minimise the impact on @BertrandBordage's Kickstarter work, I'm happy to deal with any code changes that arise from the rest of the review (although, of course, that may entail Bertrand or someone else reviewing my changes...)

BertrandBordage:
Ouch, OK.

No problem for me to review or contribute to the improvements, I still reserved some Wagtail work aside from the Kickstarter work.

wagtail/core/collectors.py
```diff
@@ -0,0 +1,379 @@
...
+                yield stream
+            return
+        current_block, *block_path = block_path
+        if isinstance(current_block, StreamBlock) \
+                and isinstance(stream, StreamValue):
+            for sub_value in stream:
+                yield from self.find_values(sub_value, block_path)
+        elif isinstance(stream, StreamValue.StreamChild):
+            if stream.block == current_block:
+                yield from self.find_values(stream.value, block_path)
```
gasman: These recursion rules don't seem quite right: unwrapping `StreamValue` and `StreamChild` happens over two iterations of `find_values`, during which two blocks will be popped off `block_path` - but both wrapper elements correspond to a `StreamBlock`, so logically only one block should have been popped by that point.

(Looking closer, I think there's some inconsistency about whether the type of `stream` corresponds to the first block in `block_path`, or the block that's just been popped in the previous iteration, and my guess is that this inconsistency only surfaces in nested `StreamBlock`s, which aren't covered by the tests.)

BertrandBordage: I’m not sure I fully follow you here.

I started building this recursive method using my knowledge of the way `StructBlock`, `ListBlock` and `StreamBlock` work, but it ended up being more complex, with several intermediate classes (that could probably be refactored). I adjusted the method to work in all the *scenarii* I could think of. Could you make an example that would not be handled by this?

gasman: Failing test (throwing an 'Unexpected StreamField value' warning) here: https://github.com/gasman/wagtail/commit/4ed64fa82c72bebfb089a39e2298ebe1bbc5412e

You can verify that there's a logic error here by asking: does the value of `stream` correspond to the first item of `block_path`, or not?

* The line `if isinstance(current_block, StreamBlock) and isinstance(stream, StreamValue):` suggests that it _does_ correspond
* The `if not block_path:` section at the top suggests that it _doesn't_ correspond (`block_path` is empty, but we still have a `stream` value)
* `elif current_block.name == '' and isinstance(stream, list):` also suggests that it _doesn't_ correspond (the 'nameless' block in a ListBlock definition is the child, not the ListBlock itself)

Unfortunately at this point I really have to ask you to add docstrings to the remainder of collectors.py before I can continue reviewing. Throughout this code there's a lack of precision about the inputs / outputs of each method... up to now I was willing to approach that as a documentation issue, and fill in that missing detail as part of the review process. However, it now seems that this lack of precision is directly causing logical errors such as this one and the regexp escaping bug earlier - at this point, I don't trust my ability to catch these logical errors at the same time as figuring out the intended behaviour of the code.

BertrandBordage:
@gasman: Just added some in-depth docstrings, hope it helps :)

BertrandBordage:
Updated the PR against master.
Back to you for the review, @gasman!

gasman:
Hi @BertrandBordage - thanks for the added docstrings! Unfortunately we've now ended up with two diverging branches - my branch at https://github.com/gasman/wagtail/tree/BertrandBordage-new-uses-ui-rebase has various fixes from the earlier round of reviewing which aren't included in this PR. Also, https://github.com/wagtail/wagtail/pull/4481/files#r190932257 still needs fixing.

I'll create a rebased version of my branch with your docstrings merged in, and open a new PR to replace this one.

wagtail/core/collectors.py
```diff
@@ -129,11 +231,37 @@ def block_tree_paths(self, block, ancestors=()):
...
+        >>> from wagtail.core.blocks import CharBlock, TextBlock
+        >>> from wagtail.core.fields import StreamField
+        >>> stream_field = StreamField([
+        ...     ('title', CharBlock()),
+        ...     ('items', ListBlock(TextBlock())),
+        ... ])
+        >>> [[b.__class__.__name__ for b in path]
+        ...  for path in StreamFieldCollector(stream_field)
+        ...              .find_block_type((CharBlock, TextBlock))]
+        [['StreamBlock', 'ListBlock', 'TextBlock']]
```
gasman: Shouldn't this be `[['StreamBlock', 'CharBlock'], ['StreamBlock', 'ListBlock', 'TextBlock']]`?

gasman:
Replaced by #4702

Issue #4414, opened 2018-03-21
Redirects drop querystring parameters 


MattSegal:
### Issue Summary

Wagtail's RedirectMiddleware does not preserve querystring parameters. For example:

```
requested url          www.example.com/old-path/?foo=1&bar=2
expected redirect      www.example.com/new-path/?foo=1&bar=2
actual redirect        www.example.com/new-path/
```

I am experiencing this issue on v1.13.1, but by inspecting the code it seems that it persists in v2.

### Steps to Reproduce

- Create a `Redirect` instance from `old_path` to a `redirect_page` / `redirect_link` (eg `/old-path/` --> `/new-path/`
- Browse to the old path, with query-string parameters (eg `/old-path/?foo=1&bar=2`)
- Observe that you are redirected to `/new-path/` without the original querystring parameters.

### Technical details

I believe the issue is with `RedirectMiddleware.process_response`, where the response object is passed `redirect.link`, which does not include the querystring. It should be possible to use `request.META['QUERY_STRING']`, or similar, to pass on the querystring. 

* Python version: 3.5
* Django version: 1.11.7
* Wagtail version: 1.13.1.
* Browser version: Chromium

thibaudcolas:
Hey @MattSegal, thank you for taking the time to write such a detailed issue 🙂. It helps a lot.

I could reproduce this in Wagtail v2 – see for example: https://bakerydemo-thibaudcolas.herokuapp.com/test-redirect?test=5, which redirects to the homepage.

I see you've already attempted to look into a potential fix – would you be keen to propose a pull request?

Vamory:
As such the fix is straight-forward, in RedirectMiddleware, replacing the following:

```
        if redirect.is_permanent:
            return http.HttpResponsePermanentRedirect(redirect.link)
        else:
            return http.HttpResponseRedirect(redirect.link)
```

with 

```
        if "?" in path:
            redirect_link = redirect.link + "?" + urlparse(path).query
        else:
            redirect_link = redirect.link

        if redirect.is_permanent:
            return http.HttpResponsePermanentRedirect(redirect_link)
        else:
            return http.HttpResponseRedirect(redirect_link)
```
works but then the following tests  breaks up:
- ERROR: test_duplicate_redirects_with_query_string_when_match_is_for_generic (wagtail.contrib.redirects.tests.TestRedirects)
[...]

Issue #647, opened 2014-09-19
Wagtail redirects and GET parameters


jossingram:
Could wagtail redirects ignore GET parameters instead of trying to match the slug with parameters and producing a 404?

Case use is that our marketing people have existing links to say /law (which use to be an old page, now is a redirect) and have many old links with Google Analytics parameters on the end for tracking campaigns. They now don't work and produce a 404, as the redirect bit matches the ?utm.... bit too.

Hope that makes sense.

benje:
Any update on this? We've got a bunch of old urls that are causing 404's because of automatically added utm parameters. A fix would help us out a bunch! :+1:

gasman:
Fixed in #1677

PR #4491, opened 2018-04-23
Ensure redirect middleware preserves query strings

MattSegal:
This PR intends to resolve an issue (https://github.com/wagtail/wagtail/issues/4414) where Wagtail's RedirectMiddleware does not preserve querystring parameters. I believe that querystrings should be preserved on redirect.

Tests and linting pass locally.

My only concern with this PR is that prior authors of the redirect middleware were aware of the behaviour that I'm trying to fix. 

Looking at this relevant past issue (https://github.com/wagtail/wagtail/issues/647), I believe that my change doesn't conflict with the intent of the established logic, which stripped the querystring parameters to ensure that we can match a redirect with no params. My change preserves that behaviour.

**Acceptance Criteria**

* [ ] Querystring parameters stripped from request before matching
* [ ] Querystring parameters passed through to the redirect target
* [ ] If there are querystring parameters already specified in the `redirect_link`, then they override any params passed through in the original request

Eg. established behavior

```
old_path        /redirectme
redirect_link   /target?foo=1

request         /redirectme/
response        /target/?foo=1
```

My change intends to preserve this behaviour, but also handle this case

```
old_path        /redirectme
redirect_link   /target

request         /redirectme/?foo=1
response        /target/?foo=1
```

There's also this tricky corner case, where I default to the respecting the `redirect_link` 

```
old_path        /redirectme
redirect_link   /target?foo=1

request         /redirectme/?foo=2
response        /target/?foo=1
```

### DIFF
--- a/wagtail/contrib/redirects/tests.py
+++ b/wagtail/contrib/redirects/tests.py
@@ -120,79 +120,86 @@ class TestRedirects(TestCase):
         self.assertRedirects(response, '/redirectto', status_code=302, fetch_redirect_response=False)
 
     def test_redirect_matches_query_string(self):
-        """
-        Given there are two redirect paths: /redirectme and /redirectme?foo=Bar,
-        when I request /redirectme?foo=Bar
-        then I am matched to the redirect with the same querystring, and the querystring is stripped.
-        """
-        # Create a redirect which includes a query string
-        redirect_with_query_string = models.Redirect.objects.create(
-            old_path='/redirectme?foo=Bar',
-            redirect_link='/path-a'
-        )
-        # ... and another redirect without the query string
-        redirect_without_query_string = models.Redirect.objects.create(
-            old_path='/redirectme',
-            redirect_link='/path-b'
-        )
-        # Navigate to the redirect with the query string
-        r_matching_qs = self.client.get('/redirectme/?foo=Bar')
-        self.assertRedirects(r_matching_qs, '/path-a', status_code=301, fetch_redirect_response=False)
+        from_url_a = '/old-path?foo=Bar'
+        to_url_a = '/new-path-a'
 
-    def test_redirect_matches_query_string_with_extra_params(self):
-        """
-        Given there are two redirect paths: /redirectme and /redirectme?foo=Bar,
-        when I request /redirectme?foo=Bar&baz=1
-        then I am matched to the redirect with the no querystring, and the querystring is passed through.
-        """
-        # Create a redirect which includes a query string
-        redirect_with_query_string = models.Redirect.objects.create(
-            old_path='/redirectme?foo=Bar',
-            redirect_link='/path-a'
-        )
-        # ... and another redirect without the query string
-        redirect_without_query_string = models.Redirect.objects.create(
-            old_path='/redirectme',
-            redirect_link='/path-b'
-        )
-        # Navigate to the redirect with the query string
-        r_matching_qs = self.client.get('/redirectme/?foo=Bar&baz=1')
-        self.assertRedirects(r_matching_qs, '/path-b?foo=Bar&baz=1', status_code=301, fetch_redirect_response=False)
+        from_url_b = '/old-path'
+        to_url_b = '/new-path-b'
 
-    def test_redirect_matches_query_string_with_pass_through(self):
-        """
-        Given there are two redirect paths: /redirectme and /redirectme?foo=Bar,
-        when I request /redirectme?utm_source=irrelevant
-        then I am matched to the redirect without at querystring, and the querystring is passed through.
-        """
-        # Create a redirect which includes a query string
-        redirect_with_query_string = models.Redirect.objects.create(
-            old_path='/redirectme?foo=Bar',
-            redirect_link='/path-a'
-        )
-        # ... and another redirect without the query string
-        redirect_without_query_string = models.Redirect.objects.create(
-            old_path='/redirectme',
-            redirect_link='/path-b'
-        )
-        # Navigate to the redirect with a different query string -
-        r_no_qs = self.client.get('/redirectme/?utm_source=irrelevant')
-        self.assertRedirects(r_no_qs, '/path-b?utm_source=irrelevant', status_code=301, fetch_redirect_response=False)
+        visit_url = '/old-path/?foo=Bar'
+        redirect_url = '/new-path-a'
 
-    def test_redirect_matches_query_string_with_extended_params(self):
+        redirect_a = models.Redirect.objects.create(old_path=from_url_a, redirect_link=to_url_a)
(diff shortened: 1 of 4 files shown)


wagtail/contrib/redirects/tests.py
```diff
@@ -119,7 +119,7 @@ def test_temporary_redirect(self):
         # Check that we were redirected temporarily
         self.assertRedirects(response, '/redirectto', status_code=302, fetch_redirect_response=False)
 
-    def test_redirect_stripping_query_string(self):
+    def test_redirect_matches_query_string(self):
```
kaedroho: Often, sites are migrated from a platform where the page is identified by a query parameter. So a user might add the following redirect:

``/index.php?page_id=1`` to ``/about-us/``

But it seems the logic here will actually redirect to ``/about-us/?page_id=1``.

Is it worth adding logic in to handle this?

Also, ideally, ``/index.php?page_id=1&utm_source=foo`` should forward to ``/about-us/?utm_source=foo``

kaedroho:
Thanks for contributing this! I can't remember if there's a reason why we strip the query string other than helping it match more URLs and cleaning out crufty URL parameters.

The implementation as it stands looks good to me, just an open question about whether we should strip out URL parameters that we were expecting in the original URL.

wagtail/contrib/redirects/tests.py
MattSegal: how would we know which querystring parameters to keep and which ones to throw away?

MattSegal:
Added a defensive check against a missing `QUERYSTRING` key in `request.META`, which happened to me in testing a little while back.

wagtail/contrib/redirects/tests.py
kaedroho: I'm thinking that any query parameters that were specified in the "old path" field should be thrown away, unless also specified in the redirect field. Everything else should be passed through.

This logic might be a little confusing to users though...

jjanssen: Regarding the discussed options, I think there are two actual use-cases:
1. You have rounded up a migration to Wagtail, therefore `/index.php?page-id=5` should redirect to `/about-us/`.
2. You are currently changing URLs in your Wagtail site, therefore `/previous-search-page/?q=keyword` should redirect to `/new-search-page/?q=keyword`.

The edge-case would be `/index.php?page-id=5&utm_source=12` but supporting that would be quite a hassle and would be really confusing for the end-user to configure.

To make it feasible for both scenarios it could make sense to add a checkbox whether the editor wants to preserve querystrings while configuring the redirect, the same way it asks whether it should permanently redirect.  

Curious to see both your perspectives on this.

MattSegal: @jjanssen I think it'd be possible to handle the edge case you mentioned without too much additional craziness.

The rules in english would be:

* Specified source params are stripped if not present in the target
* Unspecified params are passed through

```
from        /index.php?page-id=5
to          /about-us/
visit       /index.php?page-id=5
redirect    /about-us/

from        /index.php?page-id=5
to          /about-us/
visit       /index.php?page-id=5&utm_source=12
redirect    /about-us/?utm_source=12

from        /previous-search-page/
to          /new-search-page/
visit       /previous-search-page/?q=keyword
redirect    /new-search-page/?q=keyword

from        /previous-search-page/?q=keyword
to          /new-search-page/?q=keyword
visit       /previous-search-page/?q=keyword
redirect    /new-search-page/?q=keyword
```

what do you think?

gasman: @MattSegal 👍 That sounds like the ideal behaviour to me.

jjanssen: Sounds like a great idea Matt!

MattSegal: Just pushed some new changes. Turns out I couldn't do everything I said that I would do. Most of it, but not everything. I think/hope my tests make the expected behavior pretty clear.

## Can do:

Pass through requested params

```
from        /previous-search-page
to          /new-search-page
visit       /previous-search-page?q=keyword
redirect    /new-search-page?q=keyword
```

Strip old params:

```
from        /index.php?page-id=5
to          /about-us
visit       /index.php?page-id=5
redirect    /about-us
```

Merge requested and specified params:

```
from        /index.php
to          /about-us?foo=1
visit       /index.php?utm_source=12
redirect    /about-us?utm_source=12&foo=1
```

Respect redirected params

```
from        /previous-search-page?q=keyword
to          /new-search-page?q=keyword
visit       /previous-search-page?q=keyword
redirect    /new-search-page?q=keyword
```

## Can't do:

If you visit a URL that doesn't match the old_path querystring exactly, then we won't be able to find a redirect. Even if we could there are potential conflicts where 2 redirects could match 1 request. As such, we can't easily do this:

```
from        /index.php?page-id=5
to          /about-us/
visit       /index.php?page-id=5&utm_source=12
redirect    /about-us/?utm_source=12
```

Instead we have to 404, or look for a redirect that doesn't have a querystring.

MattSegal:
Just pushed some new changes. Turns out I couldn't do everything I said that I would do. Most of it, but not everything. I think/hope my tests make the expected behavior pretty clear.

## Can do:

Pass through requested params

```
from        /previous-search-page
to          /new-search-page
visit       /previous-search-page?q=keyword
redirect    /new-search-page?q=keyword
```

Strip old params:

```
from        /index.php?page-id=5
to          /about-us
visit       /index.php?page-id=5
redirect    /about-us
```

Merge requested and specified params:

```
from        /index.php
to          /about-us?foo=1
visit       /index.php?utm_source=12
redirect    /about-us?utm_source=12&foo=1
```

Respect redirected params

```
from        /previous-search-page?q=keyword
to          /new-search-page?q=keyword
visit       /previous-search-page?q=keyword
redirect    /new-search-page?q=keyword
```

## Can't do:

If you visit a URL that doesn't match the old_path querystring exactly, then we won't be able to find a redirect. Even if we could there are potential conflicts where 2 redirects could match 1 request. As such, we can't easily do this:

```
from        /index.php?page-id=5
to          /about-us/
visit       /index.php?page-id=5&utm_source=12
redirect    /about-us/?utm_source=12
```

Instead we have to 404, or look for a redirect that doesn't have a querystring.

gasman:
Hi @MattSegal - just wondering what the reason is for closing this? This is still something we're keen to fix, and was on our backlog to be reviewed...

MattSegal:
Hi @gasman I'm thought that this PR had fallen by the wayside, given this is on your to-do list I'll re-open.

MattSegal:
Given merge conflicts with wagtail/contrib/redirects/middleware.py I don't want to keep this open

steverowland:
Hmm there is a proposed solution and now it's closed? We would need this one. There is no alternative?

PR #4495, opened 2018-04-26
Fix for editing document file

notvad:
When you edit document and upload a file with the same filename - file gets removed from storage after uploading.


gasman:
Thanks @notvad! Can you add a test for this, please?

kaedroho:
This seems to be the exact opposite of: https://github.com/wagtail/wagtail/pull/3821/files

gasman:
See also #4502 / #4512 - this is a `S3BotoStorage` configuration error and should be fixed by setting `AWS_S3_FILE_OVERWRITE = False` in settings.

PR #4500, opened 2018-04-28
Make slugurl template tag site-aware.

solarissmoke:
The `slugurl` tag does not currently take into account the current site - it simply returns the URL for the first page it finds in the database. This is not the behaviour you would expect - e.g., if there are two sites that both have pages with the same slug, you would expect that `slugurl` called from a given site will return the URL of the page on that site.

This patch adds filtering on the current site to `slugurl`. To preserve backward compatibility it falls back to the old behaviour (return first matching page) in the event that a page on the current site isn't found.

I have also tweaked some of the existing tests to directly test the output of `slugurl` rather than artificially rendering a template - this seems unnecessary as we are interested in the output of `slugurl` rather than testing Django's template rendering.

Please let me know if any other changes/documentation are required!

### DIFF
--- a/wagtail/core/templatetags/wagtailcore_tags.py
+++ b/wagtail/core/templatetags/wagtailcore_tags.py
@@ -34,8 +34,25 @@ def pageurl(context, page):
 
 @register.simple_tag(takes_context=True)
 def slugurl(context, slug):
-    """Returns the URL for the page that has the given slug."""
-    page = Page.objects.filter(slug=slug).first()
+    """
+    Returns the URL for the page that has the given slug.
+
+    First tries to find a page on the current site. If that fails or a request
+    is not available in the context, then returns the URL for the first page
+    that matches the slug on any site.
+    """
+
+    try:
+        current_site = context['request'].site
+    except (KeyError, AttributeError):
+        # No site object found - allow the fallback below to take place.
+        page = None
+    else:
+        page = Page.objects.in_site(current_site).filter(slug=slug).first()
+
+    # If no page is found, fall back to searching the whole tree.
+    if page is None:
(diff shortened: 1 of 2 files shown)


solarissmoke:
I've just found https://github.com/wagtail/wagtail/pull/3123, which was essentially proposing the same thing. Hopefully this patch addresses all the comments on that PR.

wagtail/core/templatetags/wagtailcore_tags.py
```diff
@@ -34,8 +34,26 @@ def pageurl(context, page):
...
+    Returns the URL for the page that has the given slug.
+
+    First tries to find a page on the current site. If that fails or a request
+    is not available in the context, then returns the URL for the first page
+    that matches the slug on any site.
+    """
+
+    try:
+        current_site = context['request'].site
+        page = Page.objects.filter(
```
kaedroho: Please change this to: ``Page.objects.in-site(current_site).filter(slug=slug).first()``

wagtail/core/templatetags/wagtailcore_tags.py
```diff
@@ -34,8 +34,26 @@ def pageurl(context, page):
...
+            path__startswith=current_site.root_page.path,
+            slug=slug
+        ).first()
+    except (KeyError, AttributeError):
```
kaedroho: These might catch exceptions raised from Django's ORM. Could we change this to something like:
```python
try:
    current_site = context['request'].site
except (KeyError, AttributeError):
    page = None
else:
    page = Page.objects...
```

solarissmoke:
@kaedroho thanks for the review - I have updated to address your comments.

Issue #3206, opened 2016-12-07
Bug in Image.get_rendition() overwrites thumbnails of other images with similar prefix


vdboor:
We've noticed that wagtailimages overwrites thumbnails of previous image versions, when they have a similar prefix. Depending on the filter settings, this may happen after 17 similar characters.

You can test this by running the `Image.get_rendition()` function with an image that has a similar prefix. In the debugger, we found that the following values are constructed in `AbstractImage.get_rendition()`:

```
spec = 'fill-720x720-c100|format-jpeg'
input_filename = 'our-story-verhaal-3.jpg'
cache_key = '2e16d0ba'   # hash of sha1('None-None-None-None').hexdigest()[:8]
```

then the magic bug happens at https://github.com/torchbox/wagtail/blob/2c9805dc3fa37e9bf27fc7593de4465673faffca/wagtail/wagtailimages/models.py#L301:

`output_filename_without_extension = input_filename_without_extension[:(59 - len(output_extension))]`

Which produces:

```
output_filename_without_extension = 'our-story-verhaal'   # <-- SILENT TRUNCATION!!
output_filename = 'our-story-verhaal.2e16d0ba.fill-720x720-c100.format-jpeg.jpg'
```

Which means that any previous image that starts with the same 17 characters ("our-story-verhaal") is overwritten with thumbnails of the next image.

gasman:
Hi @vdboor,
Are you using django-storages / S3BotoStorage as your storage backend, by any chance? Django should take care of de-duplicating filenames, but this functionality is apparently broken in django-storages - see #1686.

vdboor:
Ah, thanks I see! Yes, we're using `S3BotoStorage`.

I've enabled the `AWS_S3_FILE_OVERWRITE = False` flag now, given that it's by design (as I can see from #573)

vdboor:
I wonder whether this code should be fixed in a different way, for example caching on file hashes (like sorl-thumbnail does) to avoid the issue all together (given that it's been reported multiple times now)

tomdyson:
> I wonder whether this code should be fixed in a different way, for example caching on file hashes (like sorl-thumbnail does) to avoid the issue all together (since it's been reported multiple times now)

+1. This would also enable site implementors to set far-future cache expiry headers on /static/images/*

gasman:
[...]

PR #4502, opened 2018-04-30
Fixed document deletion issue appears when user upload exact same document in edit view

amir-qayyum-khan:
#### What are the relevant tickets?
https://github.com/mitodl/micromasters/issues/3943
https://github.com/wagtail/wagtail/issues/4512

#### What's this PR do?
it fixes an issue which appears when user re uploads exact same file on edit file. App was delete that file all together which was casing issue/exception mention on #4512

#### How should this be manually tested?
1. Login as admin (super user)
2. Go to documents and add a document lets say `x.pdf`
3. After uploading document go back to document listing page
4. Select the document from list that you just added
5. On edit page upload same document (exact same document) and press save
6. You will be navigate to the document listing page
7. Click edit on header notification or click same document that you just update

##### Expected behaviour:
- User should be able to download file that we just reupload.
- User should be able to open edit document view page.

##### Actual behaviour:
App is crashing on edit view page or downloading file.

@pdpinch 

#### Screenshots (if appropriate)
screen shot 2018-04-30 at 1 55 25 pm

### DIFF
--- a/wagtail/documents/tests/test_views.py
+++ b/wagtail/documents/tests/test_views.py
@@ -40,9 +40,12 @@ class TestEditView(TestCase, WagtailTestUtils):
         })
         self.assertRedirects(response, reverse('wagtaildocs:index'))
         self.update_from_db()
-        self.assertFalse(self.storage.exists(old_file.name))
+        # since user has reupload document with same name, Thats why asserting document name remain same
+        self.assertEqual(old_file.name, self.document.file.name)
         self.assertTrue(self.storage.exists(self.document.file.name))
-        self.assertNotEqual(self.document.file.name, 'documents/' + new_name)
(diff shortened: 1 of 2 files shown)


gasman:
Hi @amir-qayyum-khan - thanks for the bug report and the PR.

We intentionally save the new document before deleting the old one, to ensure that (on a correctly-working storage backend) it is given a new filename to avoid caching issues - see #3816 / #3821.

Unfortunately `S3BotoStorage` is broken in its default configuration, and does not de-duplicate filenames - you need to add `AWS_S3_FILE_OVERWRITE = False` to your settings to fix this. (See #3206)

amir-qayyum-khan:
thanks @gasman for the tip, this setting `AWS_S3_FILE_OVERWRITE = False` worked for me.

PR #4504, opened 2018-04-30
Validates EmbedBlock URLs against providers.

BertrandBordage:
Currently, `EmbedBlock` accepts any URL, even when no provider is related.

With this pull request, not only `EmbedBlock` checks that the URL is valid, but it also checks whether it corresponds to existing data on the third-party site by fetching data on validation instead of doing it on front-end rendering.

As a side effect, this speeds up the first front-end page load after an `EmbedBlock` value change.

### DIFF
--- a/wagtail/embeds/blocks.py
+++ b/wagtail/embeds/blocks.py
@@ -1,4 +1,5 @@
 from django.core.exceptions import ValidationError
+from django.utils.functional import cached_property
 from django.utils.translation import ugettext_lazy as _
 
 from wagtail.core import blocks
@@ -15,7 +16,10 @@ class EmbedValue:
     """
     def __init__(self, url):
         self.url = url
-        self.html = embed_to_frontend_html(url)
+
+    @cached_property
+    def html(self):
+        return embed_to_frontend_html(self.url)
 
     def __str__(self):
         return self.html
--- a/wagtail/embeds/blocks.py
+++ b/wagtail/embeds/blocks.py
@@ -1,3 +1,6 @@
+from django.core.exceptions import ValidationError
+from django.utils.translation import ugettext_lazy as _
+
 from wagtail.core import blocks
 from wagtail.embeds.format import embed_to_frontend_html
 
@@ -12,9 +15,10 @@ class EmbedValue:
     """
     def __init__(self, url):
         self.url = url
+        self.html = embed_to_frontend_html(url)
 
     def __str__(self):
-        return embed_to_frontend_html(self.url)
+        return self.html
 
 
 class EmbedBlock(blocks.URLBlock):
@@ -56,6 +60,11 @@ class EmbedBlock(blocks.URLBlock):
             return None
         else:
             return EmbedValue(value)
+
+    def clean(self, value):
+        if isinstance(value, EmbedValue) and not value.html:
+            raise ValidationError(_('Please enter a valid URL'))
+        return super().clean(value)
 
     class Meta:
         icon = "media"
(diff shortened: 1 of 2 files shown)


wagtail/embeds/blocks.py
```diff
@@ -57,5 +61,10 @@ def value_from_form(self, value):
         else:
             return EmbedValue(value)
 
+    def clean(self, value):
+        if isinstance(value, EmbedValue) and not value.html:
+            raise ValidationError(_('Please enter a valid URL'))
```
kaedroho: Could we reword this message? It says that the URL is invalid, but actually, it is valid but doesn't point to anything.

wagtail/embeds/blocks.py
```diff
@@ -12,9 +15,10 @@ class EmbedValue:
     """
     def __init__(self, url):
         self.url = url
+        self.html = embed_to_frontend_html(url)
```
kaedroho: I'm not keen on fetching the HTML on the constructor here, could we instead create ``html`` as a cached property?

wagtail/embeds/blocks.py
BertrandBordage: I just reused the message from the rich text embed feature.

wagtail/embeds/blocks.py
BertrandBordage: No problem with making it a cached property, but why would it be problematic to do it on initialisation?

kaedroho: I think the lookup should be done during the call to ``{% embed %}`` and not at the point the ``EmbedValue`` is constructed. That's where I think most developers would expect it to happen.

Also, if the embed block is called like this: ``{% embed value.url 500 %}``, it would end up fetching the embed twice.

BertrandBordage: Well, the whole point of this PR is to fetch it on save, so you can actually tell users if the URL is valid or not. Which is exactly what happens in rich text 😉

BertrandBordage: That’s indeed what I wanted to do at first.
But if we do that, we end up with an `Embed` object and we have to catch `EmbedException`.
And when we want to display it in front-end, we have to use `render_to_string(…)` on that `Embed` object, or reuse `embed_to_frontend_html`.

So… The simplest solution was to directly reuse `embed_to_frontend_html`.  Or it would require to modify the whole `wagtail.embeds.format` file, which was out of scope and not required for a clean solution.

kaedroho: > But if we do that, we end up with an Embed object and we have to catch EmbedException

Yep, what's the problem with that?

BertrandBordage: The problem is simple, if we call `get_embed` in `EmbedValue.__init__` and `embed_to_frontend_html` in `EmbedValue.__str__`, then we end up with two queries for each embed displayed in the front-end.

gasman: The issue I see with populating `self.html` on init is that we're going from a 'lazy evaluation' model to an eager one, with a consequent (potential) performance hit. Suppose someone had a 'random video' block consisting of a list of 100 Youtube URLs, from which it will pick one to display; calling `embed_to_frontend_html` on init will result in 100 queries rather than just the one.

BertrandBordage: Then the solution is simply to move `html` to a cached property, isn’t it?

BertrandBordage:
Updated!

wagtail/embeds/blocks.py
gasman: The only place `"Please enter a valid URL"` is used is https://github.com/wagtail/wagtail/blob/1247565d3da6b542f97b59d7fd62c8c7c624f6b8/wagtail/embeds/forms.py#L7-L12 , which _is_ checking URL validity. For unrecognised embed providers, the message is `"Cannot find an embed for this URL."` - I'll update this when I merge.

gasman:
Merged in 42caa586b445fa154c835802e3dbdee52c8b72d7

PR #4506, opened 2018-05-02
WIP: pageurl generates draft URLs

mikemonteith:
Clicking on links when viewing a draft page will keep the user
in draft view mode.

Is this kind of thing desirable for wagtail?
Should it be enabled by a config setting instead of being the default?

I'll spend some time doing this MR properly (tests etc) if the feature is wanted.

### DIFF
--- a/wagtail/core/templatetags/wagtailcore_tags.py
+++ b/wagtail/core/templatetags/wagtailcore_tags.py
@@ -2,6 +2,7 @@ from django import template
 from django.template.defaulttags import token_kwargs
 from django.utils.encoding import force_text
 from django.utils.safestring import mark_safe
+from django.urls import reverse
 
 from wagtail import __version__
 from wagtail.core.models import Page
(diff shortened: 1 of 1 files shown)


wagtail/core/templatetags/wagtailcore_tags.py
```diff
@@ -19,6 +20,9 @@ def pageurl(context, page):
     if not hasattr(page, 'relative_url'):
         raise ValueError("pageurl tag expected a Page object, got %r" % page)
 
+    if getattr(context['request'], 'is_preview', False):
+        return reverse('wagtailadmin_pages:view_draft', args=[page.id])
+
```
gasman: Worth noting here that 'preview' and 'draft' are two distinct concepts - the preview button on the edit page shows the result of the edits the user has just made, which have not been saved to the database at that point (not even as draft). As a result, this won't have the intended effect - `is_preview` will become false when you follow the first link, and so the links on subsequent pages will revert to the ordinary live URLs.

mikemonteith: Ah. Is there an equivalent `is_draft`?

PR #4516, opened 2018-05-04
Implement file_hash on Image

kaedroho:
This PR adds a ``file_hash`` to ``Image``. This contains a SHA-1 hexdigest of the file contents (exactly the same value you would get from running ``shasum`` on the source image file).

This hash could be factored in to the cache key (to only invalidate CDN when the file contents changes), used for duplicate detection and could also be used for import-export (by checking if an image already exists on the destination site).


gasman:
Looks good to me! Merged in c4325fd4e91488d503a76454adfeddd27cb4673b + parents.

PR #4527, opened 2018-05-08
Allow one to extend Draftail for p and br elements

jeromelebleu:
The converter rules for `p` and `br` elements are currently not taken into account when defining a new feature in order to extend the Draftail Editor. Some rules are added by Wagtail for those elements before adding features rules. As they only match on the tag name, they always take precedence.

This pull request moves the addition of the default rules for `p` and `br` after features ones. Thus it's possible to extend Draftail and set a class on a `p` element for example.

Let me know if it's a desired behavior to prevent such extensions. Also, let me know if I should try to add some tests for that... It's my first contribution, any feedback to improve myself is welcome! Thanks.

### DIFF
--- a/wagtail/admin/rich_text/converters/html_to_contentstate.py
+++ b/wagtail/admin/rich_text/converters/html_to_contentstate.py
@@ -264,15 +264,18 @@ class LineBreakHandler:
 
 class HtmlToContentStateHandler(HTMLParser):
     def __init__(self, features=()):
-        self.paragraph_handler = BlockElementHandler('unstyled')
-        self.element_handlers = HTMLRuleset({
-            'p': self.paragraph_handler,
-            'br': LineBreakHandler(),
-        })
+        self.element_handlers = HTMLRuleset()
         for feature in features:
             rule = feature_registry.get_converter_rule('contentstate', feature)
             if rule is not None:
                 self.element_handlers.add_rules(rule['from_database_format'])
+
+        # add default rules for base elements
+        self.paragraph_handler = BlockElementHandler('unstyled')
+        self.element_handlers.add_rules({
+            'p': self.paragraph_handler,
+            'br': LineBreakHandler(),
+        })
 
         super().__init__(convert_charrefs=True)
 


gasman:
Thanks @jeromelebleu! Great to see people pushing the limits of the new rich text features :-)

I don't think we should assume that `HTMLRuleset` will always match rules in the order that they were defined. When I implemented it, I deliberately left the order of precedence unspecified, since there was no immediate need for it, and I didn't want to make the code more complicated than necessary... https://github.com/wagtail/wagtail/blob/0e19076b1c587de2aadb22935825a60a9f6c142b/wagtail/admin/rich_text/converters/html_ruleset.py#L80-L85

Now that there _is_ a need for a well-defined order of precedence, it probably makes sense to implement the ordering from CSS, so that a rule with a name + attribute has priority over just a name. This way, the fix would just exist in `html_ruleset.py` (with a new test in `/wagtail/admin/tests/test_html_ruleset.py` to make sure that the precedence rules are working as intended), and there'd be no need to reorder things in `html_to_contentstate.py`.

jeromelebleu:
With pleasure! And it is, regarding the facilities that Wagtail offers, all this work, the code, the documentation... and even your detailed answer!

> Now that there is a need for a well-defined order of precedence, it probably makes sense to implement the ordering from CSS, so that a rule with a name + attribute has priority over just a name. 

It seems to be much better - as far as I know. It makes more sense for me to use the specificity of the selector for the order of precedence. If I understand correctly and with current rules types, the order would be - from higher precedence to lower:
- element with exact attribute
- element with attribute
- element

Is it right? Another nice enhancement would be to add the other [attribute selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors) but I guess it's for another request! 

Just to go deeper in the implementation, how would you handle the case when several rules of the same *precedence* - e.g. element - are set for an element? Does the addition order will influence in this case?

gasman:
It looks like CSS treats "element with attribute" and "element with exact attribute" with the same specificity, so we should probably do the same here. If multiple rules with the same precedence match, I think it's fine to leave it unspecified which one will be returned. The only really important thing here is that element+attribute matches above element only, so no need to implement behaviour we aren't using.

jeromelebleu:
I think that giving at least precedence to more specific rules - e.g. element with attribute and exact attribute - will ease the Draftail extension. As it is the case here for `p` element, we can expect than adding a specific rule - i.e. `p[class="lead"]` - will be taking into account. Unfortunately, another one is already defined for the `p` element before our rule and take the precedence. It could also allow one to define a *catch-all* for an element, if no other more specific rule matches - as it could be the case here for the `p` element.

I was asking about multiple rules with the same precedence since I think that one solution could be to mix CSS specificity and Django app definition order. Thus, a rule for an element with an attribute and an exact attribute will have precedence over a rule for an element only. Then, if multiple rules are defined, the Django app definition order takes over. What do you think about that?

I let you see and decide as I'm not part of the contributors and I don't know how you're organized for collective decision-making... Let me know if I could help for something!

gasman:
@thibaudcolas I think CSS precedence rules are a good fit for what needs to be done here. We're putting together a ruleset to say "when you encounter an HTML element that matches specification X, it becomes contentstate entity Y", and those rules are fed in from disparate places (wagtaildocs, wagtailimages etc) depending on the currently active feature list and INSTALLED_APPS.

CSS selectors are a good way of dealing with this because they allow us to represent the rules as a flat list, rather than an arbitrarily complex decision tree ("this is an `<a>` element. Does it have a `linktype` attribute? If so, look it up in our dictionary of known link types..."). When I came up with this for 2.0, I was as surprised as anyone to realise that writing (a tiny subset of) a CSS engine was the missing piece of the rich text puzzle :-) I did search PyPI for suitable libraries at the time, but didn't find anything.

When all these rules were defined within the Wagtail codebase, it was easy enough to fudge the selectors so that they didn't overlap: for example, an external link is defined as `a[href]` rather than just `a`, to fully distinguish it from the `<a linktype="...">` rules regardless of the order the tests happen in. Once user code and third-party libraries start defining their own selectors, it naturally becomes a lot harder to avoid collisions between rules - at that point, resolving those collisions through CSS precedence makes a lot of sense, as it will tend to pick the "right" rule automatically rather than relying on the user to arrange INSTALLED_APPS in a certain way.

I share your concerns about keeping a separation between the stored and rendered version of rich text. `<p class="lead">` feels a bit too much like 'presentational' HTML to me, and not something you really want to have 'baked into' the database content, as it's entirely possible that a future iteration of your site might . (Obviously, `<p class="lead">` is far better in this regard than `<p class="green-with-purple-stripes">`... :-) ) Intuitively, I'd feel more comfortable if we used an invented element like `<lead-para>` in the database representation, to make it explicit that this is an 'idealised' semantic version of the content rather than a renderable one. However:

* we don't currently have the logic in place to rewrite wrapper elements at front-end rendering time (since you need to find the corresponding closing tag, and that's too hard to do in regexps...)
* the database representation of rich text is deliberately kept somewhat close to real HTML so that the final front-end conversion step is as efficient as possible. In this case there's no _strong_ reason to deviate from real HTML, so storing `<p class="lead">` directly is acceptable in the interests of performance.

thibaudcolas:
@gasman thank you for taking the time to explain your reasoning further 🙂. I see why it's compelling to use CSS-like selectors for this. I'm a bit worried that the precedence will not be very clear when two selectors have the exact same specificity (say `a[href]` and `a[linktype]` on a link that has both attributes), but hopefully this is rare enough and simple enough to notice that it's a non-issue. Generally i find it quite hard to visualise how many conflicts there could be in practice, since as you said this will vary both based on what's installed and also what's activated in a given field.

For the separation of concerns, in my opinion the better semantics of something like `<lead-para>` is reason enough to introduce additional processing at the (hopefully low) cost of performance and complexity. For complexity, I've never seen that code so I'm a poor judge of the actual effort. For performance, my optimism is because in practice rich text isn't "rich" enough that such processing would be costly. I would expect custom formats like `lead-para` to only make a tiny portion of the rich text content of a site (see for example https://github.com/springload/draftjs_exporter/pull/89#issue-172086587), to compare to processing of say links that's most likely just as costly.

Anyway, this is all beyond the scope of this one PR 🙂

gasman:
Have now implemented the proposed fix - enforcing CSS precedence rules - in #4926.

Issue #633, opened 2014-09-15
Firefox/Chrome + S3 + Wagtail admin + webfonts = :(


BrendanBerkley:
On localhost, admin works great in Firefox and Chrome.

On our staging server, it works fine in Chrome but not in Firefox. The CSS is pointing to the proper paths which load when I copy/paste them into the URL bar, but they just don't load in Firefox. The browser never completes loading as a result, which results in a spinning icon that give me anxiety 😛

The obvious workaround is "Do admin in Chrome," but I'm a Firefox homer so I'm posting here with hopes that a solution is found.

Thanks for all of your hard work! Wagtail has been so nice to work with.

reinka:
I found this issue via stackoverflow: https://stackoverflow.com/questions/46538682/wagtail-admin-icons-disappear-when-running-collectstatic-and-putting-everything 

I'm facing the same issue described by the OP of that post, however I'm using boto3. So I tried to implement @gasman 's solution using `AWS_S3_OBJECT_PARAMETERS` however I'm getting an other error: 

`ValueError: Invalid extra_args key 'Access-Control-Allow-Origin', must be one of: ACL, CacheControl, ContentDisposition, ContentEncoding, ContentLanguage, ContentType, Expires, GrantFullControl, GrantRead, GrantReadACP, GrantWriteACP, Metadata, RequestPayer, ServerSideEncryption, StorageClass, SSECustomerAlgorithm, SSECustomerKey, SSECustomerKeyMD5, SSEKMSKeyId, WebsiteRedirectLocation`. 

I can not figure out which key is the right one for boto3, so I would really appreciate any help on this.

Also, on my s3 bucket I'm already using the follwoing CORSRules

```
<CORSConfiguration>
	<CORSRule>
		<AllowedOrigin>*</AllowedOrigin>
		<AllowedMethod>GET</AllowedMethod>
		<MaxAgeSeconds>3000</MaxAgeSeconds>
		<AllowedHeader>Authorization</AllowedHeader>
	</CORSRule>
</CORSConfiguration>`

**Comment by gasman** [comment]
@reinka Afraid we can't really help here, as this is a boto3 issue rather than a Wagtail-specific one - I'd suggest asking on Stack Overflow instead.

**Comment by SalahAdDin** [comment]
@reinka , @gasman , actually it is no needed, putting `*` could create security issues, instead you have to put your domain, it worked for me.
```
<CORSConfiguration>
	<CORSRule>
[...]

PR #4528, opened 2018-05-08
Add documentation for deploying with cloud storage

gasman:
Add a new section to the deployment page to cover common S3 gotchas (https://github.com/wagtail/wagtail/issues?utf8=%E2%9C%93&q=AWS_S3_FILE_OVERWRITE, #633).


chosak:
Late to this, but the documentation here says that

> ``AWS_S3_FILE_OVERWRITE`` must be set to ``True``

isn't it the opposite? [The default value for this setting](https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#settings) is `True`, and it needs to be `False` to ensure different filenames. As of #3821, if this value is `True`, then re-uploading a document with the same filename will delete the file you just uploaded, leaving behind a broken admin on that page.

Someone named @gasman gave similar instructions [here](https://github.com/wagtail/wagtail/pull/4502#issuecomment-386641100).

gasman:
@chosak Oops, good catch! Fixed now.

PR #4529, opened 2018-05-09
Fix the appearance of checkbox and radio in the admin

jeromelebleu:
The appearance of all inputs is [reset](https://github.com/wagtail/wagtail/blob/35367964719dc93fb8eb79501d222e15cd148df8/wagtail/admin/static_src/wagtailadmin/scss/components/_forms.scss#L58) but only set in checkbox and radio inputs for WebKit based browsers. It results in ugly inputs in other browsers - at least in Firefox ESR 52, on Debian:

ff-before

It also moves the checkbox and radio labels next to its input. It now look like this in Firefox:

ff-after

This is also because the checkbox and radio customization only works at least in Chromium - and Chrome... It could be nice to find another way to customize them which is inside CSS scope and follows its standards! Only the label position changes in Chrome with this PR:

ch-after

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/scss/components/_forms.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/components/_forms.scss
@@ -140,8 +140,7 @@ input[type=checkbox] {
 }
 
 input[type=radio] {
-    // stylelint-disable-next-line property-no-vendor-prefix
-    -webkit-appearance: radio;
+    appearance: radio;
     width: auto;
     position: relative;
     margin-right: 15px;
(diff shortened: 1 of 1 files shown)


gasman:
I'm not seeing the issue described here on Firefox 60.0.2 on OS X: the custom widget styles _do_ show up (albeit with broken vertical alignment - will open a separate PR to address that), and this PR removes them.

On current master:
screen shot 2018-06-14 at 13 01 53
screen shot 2018-06-14 at 13 02 55

With this PR applied:
screen shot 2018-06-14 at 13 00 15
screen shot 2018-06-14 at 12 59 55

jeromelebleu:
I've updated the description to precise the Firefox version I used, which was Firefox ESR 52. For now, as this version will end-up in September, this may be closed... I let you see.

gasman:
Closing in favour of #4611, which seems to be the appropriate fix for alignment issues on current versions of Firefox.

PR #4532, opened 2018-05-11
Fix bug when deleting object from DeleteView with protected OneToOneField

nolsto:
`DeleteView.post()` is raising an unhandled AttributeError when attempting to surface related object information for instances of some models.

Not every accessor descriptor is a Manager with an `all()` method. Consider this case:

```
class Member(models.Model):
    user = models.OneToOneField(User, on_delete=models.PROTECT)
```

In the case of objects with a OneToOneField with `on_delete=models.PROTECT`, the accessor value is the related object itself rather than a Manager that retrieves objects.

In `modeladmin.DeleteView`, this case was producing an unhandled AttributeError in the post.

### DIFF
--- a/wagtail/contrib/modeladmin/views.py
+++ b/wagtail/contrib/modeladmin/views.py
@@ -801,9 +801,12 @@ class DeleteView(InstanceSpecificView):
                 obj.field, ManyToManyField))
             for rel in fields:
                 if rel.on_delete == models.PROTECT:
-                    qs = getattr(self.instance, rel.get_accessor_name())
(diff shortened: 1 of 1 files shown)


kaedroho:
Thanks! This looks good to me, but it does need a test.

lb-:
@nolsto  - thanks for this, my apologies I only saw this after merging a similar PR here
https://github.com/wagtail/wagtail/pull/5437

It appears that both PRs were fixing the same issue so we will close this PR, feel free to re-open if you feel that is not the case.

PR #4536, opened 2018-05-12
updated Django Docs Links to point to stable version

engineervix:
As I was reading the wagtail docs, I noticed that many references to the Django Documentation were pointing to old versions of django (including version 1.8), while we are using Django 2.0. An example is the [Page Models](http://docs.wagtail.io/en/stable/topics/pages.html) section, where, in the second paragraph, the Django Model field reference link points to Django Docs for version 1.9.

Instead of hard linking to specific versions of Django, may I propose that we simply link to the latest stable version. This will save us from the trouble of always having to edit this when Django changes. I have therefore changed all links to Django Docs to point to [docs.djangoproject.com/en/stable](https://docs.djangoproject.com/en/stable).


gasman:
Thanks @engineervix! Merged in c38b172793cc89cb782a34fe553ffcc414ca359e.

(Up to now, we've used the link from the current version at the time the documentation was written, so that it won't break if the Django docs are reorganised at some future point. However, that seems to be a very rare occurrence, and outdated docs are probably a bigger problem, so linking to `stable` instead is a good plan 👍 )

PR #4534, opened 2018-05-11
Adds row boosting to the PostgreSQL search backend.

BertrandBordage:
Supersedes #3953.

PR #4537, opened 2018-05-12
Multiple indices support in PostgreSQL search backend

BertrandBordage:
With this pull request, we can simply configure multiple aliases in `WAGTAILSEARCH_BACKENDS` that use `wagtail.contrib.postgres_search.backend`. These backends need to have different `'INDEX'` settings just like with Elasticsearch.

This pull request contains #4534 to avoid doing a complex merge in the future. If needed, I’ll do a new pull request containing only the changes for multiple indices support.

I know I didn’t add any documentation about `'INDEX'`, but it isn’t clearly defined in Elasticsearch so I wasn’t sure of what was best to do.


BertrandBordage:
This PR will be closed after October 2021.
If you are interested in contributing to update & complete it, please comment about it!

lb-:
As per comment above, this PR should be closed.

PR #4538, opened 2018-05-13
POC implementation of the StreamField prefetcher

mvantellingen:
So this is a first pass at implementing prefetching of objects to make Wagtail significantly faster by reducing the number of queries required. A standalone poc implementation of which I did a lightning talk at Wagtail Space 2018 is available at https://gist.github.com/mvantellingen/daebda6abbaa9a5ed0888f886a77fcf0

It's not in a merge-able state due to the fact that it misses tests, documentation updates, etc. I would first like some feedback on this before I spend time on it. The earlier implemention of bulk prefetching can be removed with this change.

The idea is that on accessing a StreamField it first collects all the related models and primary keys reference and then does an in_bulk() call for each model. A data object is then passed to each get_python() call so that the block can re-use already loaded data objects.

### DIFF
--- a/wagtail/core/blocks/stream_block.py
+++ b/wagtail/core/blocks/stream_block.py
@@ -232,7 +232,7 @@ class BaseStreamBlock(Block):
 
         return StreamValue(self, cleaned_data)
 
-    def to_python(self, value):
+    def to_python(self, value, strategy=None):
         # the incoming JSONish representation is a list of dicts, each with a 'type' and 'value' field
         # (and possibly an 'id' too).
         # This is passed to StreamValue to be expanded lazily - but first we reject any unrecognised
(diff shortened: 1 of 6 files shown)


wagtail/core/blocks/stream_block.py
```diff
@@ -376,19 +388,24 @@ def __init__(self, stream_block, stream_data, is_lazy=False, raw_text=None):
...
         if i not in self._bound_blocks:
+            if self._strategy:
+                self._strategy.prefetch_data(self.stream_block, self.stream_data)
+
             if self.is_lazy:
                 raw_value = self.stream_data[i]
                 type_name = raw_value['type']
                 child_block = self.stream_block.child_blocks[type_name]
-                if hasattr(child_block, 'bulk_to_python'):
+                if False and hasattr(child_block, 'bulk_to_python'):
```
mvantellingen: Intentionally disabled here to prove that bulk_to_python is no obsolete

wagtail/core/fields.py
```diff
@@ -61,8 +96,10 @@ def deconstruct(self):
         return name, path, args, kwargs
 
     def to_python(self, value):
+        strategy = StreamFieldPrefetcher()
```
BertrandBordage: Why name it `strategy` instead of `prefetcher`?

mvantellingen: The reason is explained at https://martinfowler.com/bliki/TwoHardThings.html ;-)

gasman:
Closing this now, as there's been some progress towards more universal adoption of `bulk_to_python` (#5926, #5930) which would be preferable to building a new mechanism.

Issue #4334, opened 2018-03-05
ModelAdmin form_fields_exclude has never worked


mixxorz:
### Issue Summary

According to the docs and PR #3295, you should be able to add `form_fields_exclude` or `get_form_fields_exclude` on a `ModelAdmin` to exclude certain fields from rendering on the admin. When I tried it, it didn't work. Upon further investigation, I found out that it has never actually worked. Issue #2759 is still open.

So a couple things that stands out to me:

1. Maybe mention in the docs that this doesn't actually work
2. Find a way to make it work for all panel types

### Steps to Reproduce

1. Create a `ModelAdmin` in `wagtail_hooks.py`
2. Set `form_fields_exclude = ['some_field']` as a class attribute
3. Create a new instance of that model in the Wagtail admin
4. Look in confusion at the still present `some_field`

### Workaround

I'm currently working around the issue by creating a custom `CreateView` for my `ModelAdmin`s and overriding `get_edit_handlers` to remove panels I don't want.

```python
class PatchedCreateView(CreateView):

    def get_edit_handler_class(self):
        if hasattr(self.model, 'edit_handler'):
            edit_handler = self.model.edit_handler
        else:
            fields_to_exclude = self.model_admin.get_form_fields_exclude(
                request=self.request
            )
            panels = extract_panel_definitions_from_model_class(
                self.model,
                exclude=fields_to_exclude
            )

            # Filter out fields in fields_to_exclude
            panels = [
                panel for panel in panels
                if getattr(panel, 'field_name', None) not in fields_to_exclude
            ]

            edit_handler = ObjectList(panels)
        return edit_handler.bind_to_model(self.model)
```

Only problem is this doesn't work for panels without field names. But in my particular case that's okay.

### Technical details

* Python version: 3.6.4
* Django version: 1.11.10
* Wagtail version: 1.13.1

lucasmoeskops:
It's actually a rare use case I think, since you already have custom panels defined, but I guess you want to customise the panels per view. You can also do this by creating multiple edit handlers, but this way should technically also be possible. I'll see if I can add the functionality for this.

lucasmoeskops:
https://github.com/wagtail/wagtail/pull/4363

ababic:
Hi guys. This feature wasn't ever meant to be used in combination with a panels attribute on the model. It's really meant to serve as a shortcut to removing a field or two from the admin form when needed (which shows all editable fields by default).

[...]

PR #4539, opened 2018-05-13
Reverted c83a712

favoyang:
c83a712 breaks the modeladmin, settings, and snippets edit views when using MultiFieldPanel.


favoyang:
Closed due to #4334 may be reverted entirely. Thanks @gasman.

PR #4540, opened 2018-05-13
#278 search images and documents by tag

cho-leukeleu:
Hi,

This PR refreshes an old PR from @ashiazed (https://github.com/wagtail/wagtail/pull/3084) about tags not being considered in admin searches.

It also takes into account the requested changes, fixes one test that it broke and removes an expected failure (and therefore should not need a new test?)


cho-leukeleu:
Hi @loicteixeira and thanks for the labeling :).
Could you please also add the labels from the original issue? 
(`component: Search` and `Type: Bug` from https://github.com/wagtail/wagtail/issues/278)

gasman:
Fixed with the new database search backend in Wagtail 2.15.

PR #4545, opened 2018-05-16
Change wagtailimages.Image on_delete CASCADE to SET_NULL in docs

alexgleason:
On the latest Wagtail, it looks like you can't set `on_delete=CASCADE` anymore for ForeignKey's to wagtailimages.Image. This Wagtail-specific error now recommends using SET_NULL or PROTECT:

```
WARNINGS:
events.EventPage.image: (wagtailcore.W001) Field hasn't specified on_delete action
	HINT: Set on_delete=models.SET_NULL and make sure the field is nullable or set on_delete=models.PROTECT. Wagtail does not allow simple database CASCADE because it will corrupt its tree storage.
```

This PR changes all instances of `on_delete=CASCADE` to `on_delete=SET_NULL, null=True` in the docs.

### DIFF
--- a/docs/advanced_topics/api/v2/configuration.rst
+++ b/docs/advanced_topics/api/v2/configuration.rst
@@ -125,7 +125,7 @@ For example:
     class BlogPage(Page):
         published_date = models.DateTimeField()
         body = RichTextField()
-        feed_image = models.ForeignKey('wagtailimages.Image', on_delete=models.CASCADE, ...)
+        feed_image = models.ForeignKey('wagtailimages.Image', on_delete=models.SET_NULL, null=True, ...)
         private_field = models.CharField(max_length=255)
 
         # Export fields over the API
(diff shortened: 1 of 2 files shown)


gasman:
Hi @alexgleason - thanks for the PR!
`on_delete=CASCADE` is only invalid for foreign keys defined on `Page` models. In other cases, it may well be the appropriate setting - for example, when an image is deleted its corresponding Rendition records should be deleted too. From what I can see, the code snippets on `configuration.rst` and `i18n/index.rst` are invalid and should be fixed - the others are correct.

alexgleason:
Cool, thanks for clarifying. Updated the PR. :smile:

gasman:
Thanks @alexgleason! Merged in 1c7e90c3eeab6ae66f5028e2f4f3ddc3f3607668 (master) and d304091c425bdf2f36fabe06e4bfd5ba52e80e6a (stable/2.1.x).

PR #4546, opened 2018-05-16
Fix wording

DanAtShenTech:


gasman:
Thanks again @DanAtShenTech!

Just a little request for future PRs - please could you open them against `master`, rather than the stable branch? It's easier to manage them way - new commits will always go on master, and we'll make a case-by-case decision about whether to merge them into a stable branch each time.

DanAtShenTech:
Will do @gasman

PR #4560, opened 2018-05-23
Ability to add external pages in promoted search results

bbusenius:
### Issue Summary

At The University of Chicago Library we need the ability to add pages to our search results that do _not_ reside on our main Wagtail website. Though it might sound counter intuitive or weird to include "external pages" in site search, we have many web properties that use different content management systems. Some of those properties have the same branding as our main Wagtail website and are considered by our users to be part of our main Wagtail website even though they're hosted on different platforms. 

Though we continue to look at more involved and comprehensive solutions using Elastic Search, we have noticed that **promoted search results** would actually solve the majority of our problems if they had the ability to include external pages. In fact, in some ways promoted search results could actually be better than indexing these pages for the following reasons:

1. Even though we want to add external pages to our search results, there is a lot of content on these pages we would want to keep out of the index. Though we could index these and fine tune the results, it seems heavy handed.
2. In-site searching is used by a very small fraction of our users. The large majority of them just use Google to search our site anyhow. We're most worried about a set of known searches.

We implemented a naive proof of concept that you can see in this pull request. We'd like to use this as a starting point for getting feedback on the idea.  

external-page-search-promotion


wagtail/contrib/search_promotions/templates/wagtailsearchpromotions/list.html
```diff
@@ -18,7 +18,9 @@ <h2><a href="{% url 'wagtailsearchpromotions:edit' query.id %}" title="{% trans
                 </td>
                 <td>
                     {% for searchpick in query.editors_picks.all %}
-                        <a href="{% url 'wagtailadmin_pages:edit' searchpick.page.id %}" class="nolink">{{ searchpick.page.title }}</a>{% if not forloop.last %}, {% endif %}
+                        {% if searchpick.page %}
+                            <a href="{% url 'wagtailadmin_pages:edit' searchpick.page.id %}" class="nolink">{{ searchpick.page.title }}</a>
+                        {% endif %} {% if not forloop.last %}, {% endif %}
```
tomdyson: where do external pages get shown?

bbusenius: In this implementation they're not differentiated from normal promoted search results. So, they're just shown at the top of the page with the normal promoted search results. They're really no different from the normal promoted search results other than the fact that they go to an external page. Certainly we could discuss changing this.

tomdyson: I see - thanks for explaining!

gasman: @bbusenius I think you're describing how they get shown on the frontend, while @tomdyson is asking about the admin listing page. It looks like the listing here is missing something to output `{{ searchpick.external_page_title }}` when `searchpick.page` is blank - otherwise the entry gets skipped over entirely (but the trailing comma still gets output).

gasman:
Looks like this is currently failing due to the missing migration - please could you add this?

wagtail/contrib/search_promotions/templates/wagtailsearchpromotions/list.html
bbusenius: Thank you @gasman! Yes, I realized my response was off during my commute last night. On the front end people can implement it however they'd like. In the Wagatail Bakery, I just showed it as part of the normal search results but none of that is included as part of this pull request of course.

bbusenius:
I'm perplexed as to why this test is failing. If I change lines 247 and 248 of wagtail/contrib/search_promotions/tests.py like below, the test passes. This implies that the reordering works but the editors_picks start out in the wrong order. I don't understand why this is the case.

```
-        self.assertEqual(Query.get("Hello").editors_picks.all()[0], self.search_pick)
-        self.assertEqual(Query.get("Hello").editors_picks.all()[1], self.search_pick_2)
+        self.assertEqual(Query.get("Hello").editors_picks.all()[1], self.search_pick)
+        self.assertEqual(Query.get("Hello").editors_picks.all()[0], self.search_pick_2)
```

wagtail/contrib/search_promotions/migrations/0003_auto_20180524_1517.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.5 on 2018-05-24 15:17
+from __future__ import unicode_literals
```
lb-, line 3: Minor additional item - Django 1.x is no longer supported, nor is Python 3 so this could be revised to be simpler (code below), also it would be good to name the migration file with something easier to understand at first glance e.g. `0003_add_external_urls.py`.

```
# -*- coding: utf-8 -*-
from django.db import migrations, models
import django.db.models.deletion


class Migration(migrations.Migration):
```

lb-:
Closing this PR in favour of a new feature request raised  - anyone can pick up where this PR left off and give this another go on the latest version of Wagtail.
https://github.com/wagtail/wagtail/issues/10725

Issue #4288, opened 2018-02-14
Segregated multi-site support


DanielSwain:
The current multi-site support in Wagtail assumes that all sites are for the same organization.  If I wanted to create sites for multiple organizations of the same type (for instance, Chambers of Commerce) on a single Wagtail installation, then the images and documents uploaded by one organization are visible to all.  Also, the pages of all organizations are visible to everyone else in the Wagtail Explorer.  Creating multiple sites for organizations of the same type on a single Wagtail installation would provide the capability to view statistics across multiple organizations and to promote cross-organization communication among members (and probably other interesting things as well).  Please consider implementing segregated multi-site support in which users who are assigned editing permissions per site can only view the pages and assets of that site.

DanAtShenTech:
Looks like Django sites framework's [CurrentSiteManager](https://docs.djangoproject.com/en/2.0/ref/contrib/sites/#the-currentsitemanager) might be useful.  Perhaps the requested functionality could simply be implemented as an option in Settings.

gasman:
Hi @DanAtShenTech - thanks for the suggestion.

This is indeed something we're keen to implement, and there are various strands of work going towards this (#2869, #4206, #4207) - although it looks like we don't currently have an open issue for the feature as a whole since #2463 was closed, so I'll leave this one open.

It's possible to achieve most of this through user groups and permissions - in particular, a user who's been assigned permission over a single sub-site won't see other sites in the explorer. The main thing that's missing at the moment is introducing a 'choose' permission level so that we can restrict the pages shown in the 'choose a page' popup. Note that we want to avoid tying our model of limited-access users too closely to the concept of 'sites', since we want to keep things flexible enough to allow for other arrangements - for example, an organisation might want to enforce this kind of compartmentalisation between sections of a site, without having a distinct domain name for each one.

DanAtShenTech:
[...]

PR #4563, opened 2018-05-25
Add image permission policy to ImageChooser

ArnaudLarroche:
The images and collections shown in the `Images` admin section depend on the user permissions, but the images and collections displayed in the ImageChoosers do not, as shown below.
This PR ensures that the images and collections accessible to a user are the same in both places.

On the `Images` admin section, the images depend on the user permissions (here the `test` user group has permissions for one collection only) :
images_page

On an ImageChooser, the same `test` user has access to all images and collections :
imagechooser

### DIFF
--- a/wagtail/images/views/chooser.py
+++ b/wagtail/images/views/chooser.py
@@ -7,7 +7,6 @@ from wagtail.admin.forms import SearchForm
 from wagtail.admin.modal_workflow import render_modal_workflow
 from wagtail.admin.utils import PermissionPolicyChecker, popular_tags_for_model
 from wagtail.core import hooks
-from wagtail.core.models import Collection
 from wagtail.images import get_image_model
 from wagtail.images.formats import get_image_format
 from wagtail.images.forms import ImageInsertionForm, get_image_form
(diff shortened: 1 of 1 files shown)


gasman:
Thanks for the PR @ArnaudLarroche!

This change isn't valid, I'm afraid - a user with no permission to add or edit images should still be able to select existing images to be inserted into their pages, and so it's intentional that the chooser ignores add/edit permission. The correct way to fix this (and make it possible to limit a user's ability to choose images) would be to introduce a new 'choose' permission type - this has now been opened as a feature request at #4288.

ArnaudLarroche:
Hi @gasman,

Thanks for the feedback and sorry for the misunderstanding!

arminbw:
Hi @gasman. Imho a user who can't edit something should still be able to see/inspect it in the admin interface. At least for me it makes sense (I would need it for an admin-interface based image database). Would it be possible to add a 'view' permission?

BertrandBordage:
Note that Django 2.1 adds a `view` permission to each model to tell whether or not users can view instances in the admin without any other permission.

Maybe we should use that new permission in all admin views once we support Django 2.1?

gasman:
@arminbw In general I think it makes sense to hide items in the admin when users don't have permission - many of the items within the Settings menu (sites, redirects...) are shown or hidden depending on permission, and it would add a lot of complexity if we were to show those to 'regular' users. Adding a 'view' permission would be a good way forward.

Issue #2936, opened 2016-08-23
Stop importing _editor_js.html everywhere


gasman:
[`wagtailadmin/pages/_editor_js.html`](https://github.com/wagtail/wagtail/blob/main/wagtail/admin/templates/wagtailadmin/pages/_editor_js.html) was only ever meant to be used on the page add/edit views, but is currently imported in various templates throughout Wagtail admin because it's a convenient grab-bag of JS functions that saves the effort of actually thinking about what JS you're using. Some of the code there is generally reusable (e.g. page chooser modals) and some of it is relevant to other edit-handler-based interfaces (snippets, settings) but a lot of it is only specific to the page editor (e.g. setting up behaviour for the slug field and preview button). We really ought to refactor these templates so that they only import and run the JS that's actually meaningful to each page.

gasman:
Have been looking into what it would take to allow `AdminPageChooser` to be used on pages outside the page editor, without importing all of _editor_js.html, and it seems like it's going to be more involved than just shuffling some import lines around...

AdminPageChooser requires the following JS code to work:

- the (dynamic) snippet that sets up window.chooserUrls
- modal-workflow.js
- page-chooser.js

We could easily split these off into a new _page_chooser_js.html include, included from _editor_js.html and any other templates that require AdminPageChooser. However, this is going to lead us down a path of messy dependency tracking: a page that includes _page_chooser_js.html has to know not to import modal-workflow.js itself. And if we do the same decoupling exercise on another component, such as the image chooser, there's no way to include both of them without importing modal-workflow.js twice.

To make this work, then, we'll need a helper that keeps track of the JS imports that are already present, and de-duplicates them - something like Django's form media, but globally for the template, and able to handle arbitrary snippets such as the one for window.chooserUrls, not just `<script>` file imports. Perhaps the building block would be a template tag like:

    {% includeonce chooser_urls %}
        <script>
            window.chooserUrls = {...};
        </script>
    {% endincludeonce %}
    {% includeonce modal_workflow %}
        <script src="{% static 'modal-workflow.js' %}"></script>
    {% endincludeonce %}

thibaudcolas:
This sounds like the textbook use case for a module system and a bundler like Webpack. I think we should move the dependency graph management there, further away from Django land.

I'm not sure how this fits with dynamic snippets that generate arbitrary JS.

lb-:
[...]

PR #4566, opened 2018-05-28
Eliminate Django templating from modal-workflow javascript

gasman:
A first step in refactoring the modal-workflow mechanism so that it doesn't have to send arbitrary Javascript down the wire in response to AJAX requests. Related: #2936, #4335, #4357 

At the heart of this is a simple tweak to modal_workflow.py, to allow passing additional fields in the almost-but-not-quite-JSON response, alongside the standard `html` and `onload` fields. The `onload` handler can then retrieve the server-side data it needs from those fields, instead of having it interpolated into the JS code through Django template tags.

(The next step will be to move the now-static-JS `onload` handlers into the `ModalWorkflow` initialisation code, and have the AJAX responses include a simple `step` token that tells us which of those handlers to run. This bit is going to be more fiddly, due to there being multiple places that invoke the same modal - e.g. the page / link chooser is invoked by the AdminPageChooser widget and each rich text implementation - which now have to come up with a way of sharing that code.)

### DIFF
--- a/wagtail/admin/modal_workflow.py
+++ b/wagtail/admin/modal_workflow.py
@@ -4,7 +4,7 @@ from django.http import HttpResponse
 from django.template.loader import render_to_string
 
 
-def render_modal_workflow(request, html_template, js_template, template_vars=None):
+def render_modal_workflow(request, html_template, js_template, template_vars=None, json_data=None):
     """"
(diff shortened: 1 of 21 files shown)


gasman:
Merged as part of #4583

PR #4526, opened 2018-05-08
Add file_hash field to Document

kaedroho:
Same as https://github.com/wagtail/wagtail/pull/4516 but for documents

Needs tests, going to wait for #4517 and #4525 to be merged before implementing them

Issue #3883, opened 2017-09-28
Wagtail is not creating unique Rendition file names when using S3


nathanbegbie:
### Issue Summary

*What:* In certain cases, different Image Rendition file names are given the same name and overwriting older files where names collide.

*When:* 
- there are 2 or more images that share the same beginning of the file title e.g. `abcdefghijklmnopqrstuvwxyz.jpg` and `abcdefghijklmnopqrstuv0000.jpg`
- the image file names are very long and need to be truncated when naming the file renditions
- I am using AWS S3 for media storage

*The following happens:* When creating a rendition with the same `filter_spec`, the rendition's file names for 2 or more images, that have above properties, are allocated the same name.
This means that one or more renditions are simply overwritten with the new rendition.
I suspect that this is due to a failing check whether the image with the same name already exists.

### Steps to Reproduce

Below is the **expected** behavior when creating renditions, using local storage:
```
$ ./manage.py shell
Python 2.7.13 (default, Sep 1 2017, 11:24:58)
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.42)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> from wagtail.wagtailimages.models import *
>>> from django.conf import settings
>>> from wagtail import VERSION
>>> print VERSION
(1, 9, 1, u'final', 0)
>>> settings.DEFAULT_FILE_STORAGE
'django.core.files.storage.FileSystemStorage'
>>> Image.objects.count()
2
>>> img_1 = Image.objects.first()
>>> img_2 = Image.objects.last()
>>> img_1.file.name
u'original_images/abcdefghijklmnopqrstuvwxyz.jpg'
>>> img_2.file.name
u'original_images/abcdefghijklmnopqrstuv0000.jpg'
>>> img_1.renditions.all()
[]
>>> img_2.renditions.all()
[]
>>> rend1 = img_1.get_rendition('fill-300x150|jpegquality-60')
>>> rend1.url
'/media/images/abcdefghijklmnopqrs.2e16d0ba.fill-300x150.jpegquality-60.jpg'
>>> rend2 = img_2.get_rendition('fill-300x150|jpegquality-60')
>>> rend2.url
'/media/images/abcdefghijklmnopqrs.2e16d0ba.fill-300x150.jpegquality-60_06esx7Y.jpg'
```

Note that the filenames for the renditions, where there was a naming collision, has added on 
[...]

PR #4567, opened 2018-05-29
Adds support for overwriting file storage backends

chosak:
This change modifies the admin edit view for both documents and images to prevent deletion of uploaded files when the storage backend implements file overwriting.

See #4495 and #4502 as examples of previously reported issues where this happened with the django-storages S3 backends, which [use overwriting by default](https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#settings) (see `AWS_S3_FILE_OVERWRITE`).

The current behavior (implemented in 1.13 in #3821 and #3822) is to delete an old file when a new file is uploaded for a document or image. Normally this works fine, as Django file storage backends are supposed to give each upload a unique name (typically by appending random characters to filenames).

However, there are some backends that support overwriting of files instead, meaning that a file that gets uploaded with the same name as an existing file will overwrite that file, instead of getting stored
alongside it. In these cases, the current Wagtail logic first overwrites the file and then deletes it, leaving a broken document/image with no stored file.

This logic change simply checks to make sure that the old name and the new name are in fact different; if the backend assigned them the same name then the deletion is skipped. This shouldn't have any downside but does allow for continued use of overwriting backends.

This change includes various unit tests that check this behavior.

This change also reverts part of the documentation added in #4528, as using  django-storages S3 backends in their default configurations (with `AWS_S3_FILE_OVERWRITE` set to True) is now possible.


Scotchester:
This will be a great help to our use case at CFPB where we have relied for years on the S3 URL for linking to our documents, not the Wagtail URL, and need to be able to replace files in place when corrections are necessary.

gasman:
Thanks @chosak! A few initial observations:

* Re-uploading a document is not the only place in Wagtail where `AWS_S3_FILE_OVERWRITE=True` is broken; from a quick glance, #1686, #3108, #3206 and #3883 all describe different issues to this one (since they concern multiple distinct image/document records with clashing filenames, rather than replacing a file under the same record). If we're going to formally support `AWS_S3_FILE_OVERWRITE=True` as a valid configuration, we need to address these too.
* The use-case that @Scotchester describes is directly in contradiction to the intent of #3816 / #3821. That doesn't mean it's invalid - I can see there are arguments both ways. Given that #3821 is not a particularly robust way of performing cache-busting anyhow - it'll revert to the original filename if the file is uploaded a third time - it's probably worth taking another look at #3821 and seeing if there's a better way of cache-busting which doesn't rely on file renaming, perhaps adding a file hash as a querystring. (#3821 had the nice feature that it was a one-line fix, but it sounds like we're rapidly heading out of one-line-fix territory at this point...)
* Up to now Wagtail has never guaranteed anything about the behaviour of direct document file URLs (in fact, conventional Wagtail wisdom would dictate that document files should not have an accessible URL at all, so that permission checks can't be bypassed). However, see #1420 / #3772 which will allow configuring Wagtail to use the direct file URL as the 'canonical' document download URL.

With that out of the way - I'm wary that we may be fixing the right thing for the wrong reason here. :-) If you'd come here saying "I want to fix our handling of `AWS_S3_FILE_OVERWRITE=True` to make Wagtail more fault-tolerant (because people are inevitably going to use it regardless of what we say in the docs)", I'd be fine with that... but if you're saying "I want to fix our handling of `AWS_S3_FILE_OVERWRITE=True` because that configuration has a specific behaviour regarding file renaming that we want", that's a lot more troublesome. If Django storage backends are doing their job properly, they ought to be interchangeable and functionally identical, and it should never be the case that you need to use one specific configuration to achieve the behaviour you want. To put it another way: if not-renaming-files is indeed a desirable feature of Wagtail, then we should make that feature available to everyone, not make it conditional on using one specific quirky-and-nonstandard storage backend.

In conclusion: if your motivation really is to fix `AWS_S3_FILE_OVERWRITE=True`, then that's fine (but if so, then let's fix the other bugs with it, and make its behaviour re file renaming consistent with other backends). If the _real_ motivation is to avoid file renaming, then let's come up with an alternative cache-busting solution to replace #3821, that works equally well across all supported backends (and _could_ involve adding support for `AWS_S3_FILE_OVERWRITE=True`, but I suspect it won't; the more likely outcome is that this change will make it possible for you to use `AWS_S3_FILE_OVERWRITE=False` instead).

I suspect that getting #4526 merged will be a very good first step towards that solution - a review of that would be very welcome, if you're up for it!

chosak:
@gasman thanks for the long and thoughtful response.

I agree that using `AWS_S3_FILE_OVERWRITE=True` is a non-standard method, and that long-term the various issues you link to should be addressed to make Wagtail file serving more configurable. That being said, I do think that any time Wagtail is going to be doing file deletions, then we should err on the side of being as cautious as possible.

It may not be Wagtail's responsibility to enforce proper backend behavior, but it does feel like our responsibility to ensure that users know when an upload goes wrong. Currently, the overwritten file gets silently deleted, and the user gets returned to the document/image list view with no indication that anything went wrong. It's only when returning to the edit view that they'd see an error (for images, reliably as of #4468, I think for documents they'd still see 500s unless a similar PR is opened).

If we're not going to fix this, then we should think about better ways to let developers and users know about the risks of overwriting. Perhaps adding a check like this and raising an exception when an overwrite happens? Right now there's really no way for developers/users to know about this unless they happen to upload a new file with the same filename.

I totally get that using overwrite has its own problems -- for example, one user can overwrite another user's files -- and it's tricky to define exactly where/how much Wagtail should to do avoid this. My argument for this PR is less about supporting a particular backend and more about preventing what we know is likely to happen (and restoring behavior that worked through Wagtail 1.12), especially given the small impact and (as far as I can tell) lack of negative side effects.

gasman:
I'm happy with the idea of allowing filenames to be preserved on re-upload (and, indeed, making that the default - I'm coming round to the idea that it was a bad decision to push the 'cache-busting' behaviour on users).

What I _won't_ support is having that option conditional on using one particular storage backend setup (one that violates Django's specification for storage backends, no less). It makes my life more difficult as a maintainer (I want to be able to test on one configuration and be confident that things will work on other setups - the whole point of storage backends is that they should be interchangeable) and it's not really an acceptable solution for people who want document URLs to be persistent but don't want to use S3.

With that in mind, I'd say the steps for moving this forward are:

* Decide on, and implement, a new policy for what happens when replacing document files - considering the two cases where the newly-uploaded filename does, and doesn't, match the original. Do we always take the new filename, (except when it clashes with another document), or keep the original one? The use-cases described here (i.e. making sure existing links don't break) would suggest the latter, but that has the downside that it's impossible to fix mistakes in the filename.

* Provide an alternative fix for the caching issue described in #3816. Options for this that come to mind:
  * Make sure we're serving appropriate max-age and etag HTTP headers on documents - although we'll need to check that these are effective against browser caches, and deal with the fact that we only have full control over HTTP headers if the file is actually served through Django, which won't necessarily be the case if #5296 goes ahead. (It's probably reasonable to say that if you're hosting documents remotely and not serving them through Django, caching is your own responsibility...)
  * Add a querystring containing `file_hash` to the document URL - but see https://github.com/wagtail/wagtail/pull/5527#issuecomment-523875063 for possible issues with that.

It's tempting to add a configuration setting for this, but I'm not enormously keen on that - it feels a bit 'micro-manage-y', and it's not great to make people choose between persistent URLs and effective cache-busting. Ideally we'd find a way forward that handles common requirements out of the box, and provide customisability - through custom document models or similar - to support the edge cases.

johannesvogel:
We have a customer that requires stable document URLs, so we ran into this problem unfortunately (see my issue #5712).

Personally I believe, the behaviour should be as follows:

- If the user uploads a **new version** of an existing document with the **same filename**, the filename should stay the same as well as the URL.
- If the user uploads a **new version** of an existing document with a **different filename**, the new filename should be used and the URL should therefore change. _This is the current behaviour._
- If the user uploads a **completely new document** whose **filename collides** with an already existing document, the filename should be appended. _This is the current behaviour._

As discussed in this thread, we could run into caching issues, if the filename stays the same. Because of that I had a look at [Django's ETag implementation](https://docs.djangoproject.com/en/2.2/topics/conditional-view-processing/) and created a quick prototype for the Document serve view (`wagtail/documents/views/serve.py`):

```python
from django.views.decorators.http import etag

def doc_etag(request, document_id, document_filename):
    Document = get_document_model()
    doc = get_object_or_404(Document, id=document_id)
    return '"{}"'.format(doc.file_hash)

@etag(etag_func=doc_etag)
def serve(request, document_id, document_filename):
    Document = get_document_model()
    doc = get_object_or_404(Document, id=document_id)
    # ...
```

This seems to work! But as you might already see, it queries the database twice, so it's probably not optimal. This is necessary because the `doc_etag()` method has to also work independently from the view method (`serve()` will not be called in case the document is already cached by the client). One might attach the document to the request (`request.doc = doc`) in `doc_etag()` and then retrieve it in `serve()` but this sounds a bit hacky. Maybe there are other options I currently can't think of, so that we only query the db once?

According to caniuse using ETags for cache control [should be a save bet](https://caniuse.com/#search=etag) on the client side. But I am not so sure about service like Cloudflare etc.

tomusher:
Ran in to this myself recently when a client reported users' bookmarks had stopped working (because they were pointing at documents they had been updated).

My 'fix' was to remove this check https://github.com/wagtail/wagtail/blob/master/wagtail/documents/views/serve.py#L22, so as long as the ID is correct, the document is served - which as I understand it, shouldn't interfere with any caching or overwriting behaviour.

I appreciate the security concerns this check addresses, but is there perhaps a middle-ground here that would address the security issues but restore the 'old' behaviour?

rgs258:
I just bumped into this too. I'm just telling users to upload the file twice 🤦‍♂️

FlipperPA:
We're still running into this issue and would love to see a solution at some point. The typical scenario is that a link has to be updated in a document; the document is updated, and re-uploaded; the new filename breaks things like reporting, and looks ugly. Our editors have been deleting the original document, re-uploading the document, and then trying their best to restore the original links that were broken. This is less than ideal!

Scotchester:
> 
> 
> We're still running into this issue and would love to see a solution at some point. The typical scenario is that a link has to be updated in a document; the document is updated, and re-uploaded; the new filename breaks things like reporting, and looks ugly. Our editors have been deleting the original document, re-uploading the document, and then trying their best to restore the original links that were broken. This is less than ideal!

@FlipperPA I know it's less than ideal, and I also would like to see this get to a more permanent solution, but does the double-upload-when-updating-a-doc workaround not at least save you the work of having to restore broken links?

tjwalch:
I think the root problem here is that Wagtail is trying to solve something in #3816 (and previously) that is none of it's business. The correct way to solve it would be to trust in Django's storage system and not enforce anything, one way or another. At most is  a documentation issue, i e referencing Django's storage mechanism and how to configure it to achieve browser invalidation.

gasman:
Have now opened #9776 with a summary of the desired way forward, as per the consensus reached here - closing this PR accordingly (since giving 'preferential' treatment to overwriting backends isn't the approach we want to take).

PR #4568, opened 2018-05-30
Allow specifying js/css imports against Draftail features

gasman:
Hook up [Django form media](https://docs.djangoproject.com/en/2.0/topics/forms/media/) to Draftail feature definitions, so that additional js/css imports can be inserted whenever a feature is active.

Besides making third-party draftail add-ons simpler to write (no need to set up an `insert_editor_js` hook any more), this is also a necessary step towards refactoring modal-workflow so that it doesn't rely on pushing JS down the wire in AJAX responses (#4566) - that JS is going to exist in static files within wagtailimages / wagtaildocs etc instead, and so we need a clean way to pull those files in so that Draftail can continue to use the chooser modals.

### DIFF
--- a/wagtail/admin/rich_text/editors/draftail/__init__.py
+++ b/wagtail/admin/rich_text/editors/draftail/__init__.py
@@ -21,6 +21,12 @@ class DraftailRichTextArea(WidgetWithScript, widgets.HiddenInput):
         kwargs.pop('options', None)
         self.options = {}
 
+        self._media = Media(js=[
+            'wagtailadmin/js/draftail.js',
+        ], css={
+            'all': ['wagtailadmin/css/panels/draftail.css']
+        })
+
         self.features = kwargs.pop('features', None)
         if self.features is None:
             self.features = feature_registry.get_default_features()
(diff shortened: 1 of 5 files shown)


wagtail/admin/rich_text/editors/draftail/__init__.py
```diff
@@ -29,6 +35,7 @@ def __init__(self, *args, **kwargs):
             plugin = feature_registry.get_editor_plugin('draftail', feature)
             if plugin:
                 plugin.construct_options(self.options)
+                self._media += plugin.media
```
thibaudcolas: Does this guarantee that `draftail.js` will be loaded before the plugin's JS?

gasman: I think so, yes. It definitely does in Django 2.0: https://docs.djangoproject.com/en/2.0/topics/forms/media/#order-of-assets

Hopefully the note about earlier versions ("concatenated rather than merged") just means that 1.11 can't handle fancy cyclic dependencies like the one in the example, rather than not preserving order at all - but I'll double-check that.

gasman: Yep, on Django 1.11 the `+` operation appends items in the original order: https://github.com/django/django/blob/stable/1.11.x/django/forms/widgets.py#L104

thibaudcolas: Sweet :) thanks for tracking it down

gasman:
Merged in 607f2ec0673814a54bd8c35f7cda42a4b37f73f2

PR #4570, opened 2018-05-30
Improve Dockerfile in the project template

tm-kn:
I've updated a couple of things in the Dockerfile in the project template.

* It had a probable mistake of running `django-admin migrate` command during the build-phase which only may make sense for SQLite, but is probably an anti-pattern anyway. Moved that to the `CMD` so it runs on the container start.
* Pinned Gunicorn version.
* Configuring Gunicorn with environment variables rather than in-line arguments. Makes it more readable and more flexible for the user since they can override defaults when running the image.
  * Also using `PORT` variable makes it compatible with Heroku since that's what they set [1].
* Set default settings to production.
* Set `PYTHONPATH` to the application directory so `django-admin` works.
* Change main directory from `/code` to `/app` since that is, in my opinion, more widely used and is shorter. :stuck_out_tongue_closed_eyes: 
* Added collecting static during the build time as opposed to not collecting it at all.

I've added two settings that are read from the environment - `SECRET_KEY` and `ALLOWED_HOSTS`.

[1] https://devcenter.heroku.com/articles/runtime-principles#web-servers

### DIFF
--- a/wagtail/project_template/Dockerfile
+++ b/wagtail/project_template/Dockerfile
@@ -1,21 +1,38 @@
-FROM python:3.6
+FROM python:3.6-stretch
 LABEL maintainer="hello@wagtail.io"
 
-ENV PYTHONUNBUFFERED 1
-ENV DJANGO_ENV dev
+WORKDIR /app
 
-COPY ./requirements.txt /code/requirements.txt
-RUN pip install -r /code/requirements.txt
-RUN pip install gunicorn
+ENV PYTHONUNBUFFERED=1 \
+    # Use production settings as default for the runtime.
+    DJANGO_SETTINGS_MODULE={{ project_name }}.settings.production \
+    # Set the PYTHONPATH so django-admin can be used.
+    PYTHONPATH=/app \
+    # Set Gunicorn's workers to 2 and its port to 8000 as default.
+    WEB_CONCURRENCY=2 \
+    PORT=8000 \
+    GUNICORN_CMD_ARGS="--access-logfile -"
 
-COPY . /code/
-WORKDIR /code/
+# Expose Gunicorn's port outside of the container.
+EXPOSE 8000
 
-RUN python manage.py migrate
+# Install the application server - Gunicorn.
+RUN pip install "gunicorn>=19.8,<19.9"
 
+# Install your application's required dependencies.
+COPY requirements.txt .
+RUN pip install -r requirements.txt
+
+# Don't use the root user.
 RUN useradd wagtail
-RUN chown -R wagtail /code
+RUN chown -R wagtail .
 USER wagtail
 
-EXPOSE 8000
-CMD exec gunicorn {{ project_name }}.wsgi:application --bind 0.0.0.0:8000 --workers 3
+# Copy your application's files.
+COPY . .
+
+# Collect static assets.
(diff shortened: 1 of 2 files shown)


SalahAdDin:
why do you restrict `gunicorn` version?

tm-kn:
> why do you restrict gunicorn version?
@SalahAdDin : If someone uses this Dockerfile in two years, the environment variables like `PORT` or `WEB_CONCURRENCY` may not work any more. Also there's no proof that passing the WSGI module as the first argument of the command won't change. It's not very likely, but for example Wagtail has seen breakages with the REST framework versions which would break sites running older versions of Wagtail.

tomdyson:
Will Heroku's `PORT` take precedence over the Dockerfile `PORT`?

tm-kn:
> Will Heroku's PORT take precedence over the Dockerfile PORT?

@tomdyson: Yes, everything you pass with `-e` to `docker run` will override `ENV` directives. And yes, it works on Heroku.

wagtail/project_template/Dockerfile
```diff
@@ -1,21 +1,37 @@
 FROM python:3.6
```
tm-kn: Probably pinning to a certain Debian version is a good idea to avoid surprises.

wagtail/project_template/Dockerfile
```diff
@@ -1,21 +1,37 @@
...
+
+# Install your application's required dependencies.
+COPY requirements.txt .
+RUN pip install -r requirements.txt
 
-RUN python manage.py migrate
+# Copy your application's files.
+COPY . .
 
+# Don't use the root user.
```
mikedingjan: isn't it better to do this before adding the application files, this way the layer can't be cached in docker when application files change, or am I wrong?

mikedingjan:
I think having the `production` settings as a default isn't a good practice, I understand that having when running via docker you can only choose to run on settings module. In my opinion it would be better to have `docker settings` and have everything which is different per environment be set by Environment variables (with defaults off course).. it probably isn't a huge difference with how it's working now but naming things properly is important to..

I think running migrate in your `CMD` instruction is quite dangerous, theoretically it can result in multiple containers running different code with different migration states. Ideally migrate shouldn't be part of building your image imo.

tm-kn:
> I think having the production settings as a default isn't a good practice, I understand that having when running via docker you can only choose to run on settings module. In my opinion it would be better to have docker settings and have everything which is different per environment be set by Environment variables (with defaults off course).. it probably isn't a huge difference with how it's working now but naming things properly is important to..

@mikedingjan We probably should not care about the subjectively "best". Different people have different workflows and I think we should try to keep things simple and not opinionated. Also we should keep changes small. We could add docker settings as a separate PR. We could probably not default to the settings in the Dockerfile and instead use them only in the relevant sections, i.e. the `collectstatic` command. I can't see why defaulting to production settings is wrong though. :P If people think it's too opinionated we probably should not use it.

> I think running migrate in your CMD instruction is quite dangerous, theoretically it can result in multiple containers running different code with different migration states. Ideally migrate shouldn't be part of building your image imo.

I agree. We should probably take it out and let user migrate the database themselves. They should set it in their deployment platform.

tm-kn:
- Add dockerignore.

tm-kn:
Closing in favour of #5800.

PR #4571, opened 2018-05-31
Added ability to nest ModelAdminGroup's in the wagtail admin

Squarehost:


wagtail/contrib/modeladmin/options.py
```diff
@@ -547,7 +550,10 @@ def get_submenu_items(self):
         menu_items = []
         item_order = 1
         for modeladmin in self.modeladmin_instances:
-            menu_items.append(modeladmin.get_menu_item(order=item_order))
```
ababic: Hi @Squarehost, rather than add conditionals here, I think a better approach might be to add support from the other side, by altering `__init__()` and `get_menu_item()` on `ModelAdminGroup` to support the passing of `parent` and `order`. We probably actually want the `order` passed to ModelAdminGroup.get_menu_item() to override the value set on it, so that the order specified in `items()` is respected.

lb-:
As this has been a few years without response, I will close.

Since this time, ModelAdmin has had a lot of improvements and also Snippets has started to provide advanced menu nesting capabilities.

On top of that the Menu JS has been rewritten with the ability to be more extensible.

PR #4575, opened 2018-06-01
[WIP] Implement RFC27: search autocomplete API for Elasticsearch

kaedroho:
This PR makes a start on the new autocomplete API specified in [RFC 27](https://github.com/wagtail/rfcs/pull/27).

TODO:
 - [ ] ``partial_match`` argument on ``.search()``
 - [ ] Tests

### DIFF
--- a/wagtail/search/backends/base.py
+++ b/wagtail/search/backends/base.py
@@ -268,6 +268,7 @@ class EmptySearchResults(BaseSearchResults):
 
 class BaseSearchBackend:
     query_compiler_class = None
+    autocomplete_query_compiler_class = None
     results_class = None
     rebuilder_class = None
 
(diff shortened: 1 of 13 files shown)


wagtail/search/backends/base.py
```diff
@@ -298,7 +300,7 @@ def add_bulk(self, model, obj_list):
     def delete(self, obj):
         raise NotImplementedError
 
-    def search(self, query, model_or_queryset, fields=None, operator=None, order_by_relevance=True):
+    def search(self, query, model_or_queryset, fields=None, operator=None, order_by_relevance=True, partial_match=True, query_compiler_class=None):
```
BertrandBordage: Passing a `query_compiler_class` in that method seems very weird to me. Do we expect users to use it on the backend? This also gives some inconsistency with the `…QuerySet.search()` method that doesn’t provide this argument.

Maybe the searching logic could be moved into a separate method that both `search` and `autocomplete` would call while passing their own `query_compiler_class`.

kaedroho: Just refactored based on your comment, thanks!

gasman:
Failing against Elasticsearch 6 :-( https://travis-ci.org/wagtail/wagtail/builds/402746170

BertrandBordage:
Yes, I saw that after merging, it was not the case in the PR and could only test locally with ES 2. :'(

BertrandBordage:
Fixed by https://github.com/wagtail/wagtail/pull/4688.

PR #4579, opened 2018-06-04
Fix support for related_query_name in InlinePanel relations

Aramgutang:
When `related_query_name` is used in a `ParentalKey` of a model that's then registered as an `InlinePanel`, `InlinePanel.on_model_bound()` would throw an error, since the model would no longer have an attribute that matches the name passed to `get_field`.

Here is the offending line: https://github.com/wagtail/wagtail/blob/master/wagtail/admin/edit_handlers.py#L689

Now you'd think it should be as simple as changing it to:
```
manager = getattr(self.model, self.db_field.get_accessor_name())
```

But unfortunately, the way Django handles `related_query_name` is unintuitive, and it's unclear if this behaviour is intended, per this ticket: https://code.djangoproject.com/ticket/29458

So what happens is, `model.related_name` returns the related manager, but `model._meta.get_field('related_name')` throws an error. `model._meta.get_field('related_query_name')` will however get your the field, which can also be obtained via the `.rel` attribute of the manager.

I've updated the tests to have the `EventPageSpeaker` use `related_query_name` to test this fix, which necessitated a few slight updates to how the field is referenced elsewhere in the tests.


gasman:
Excellent, thanks @Aramgutang! Merged in c3b7a22d73650693eb4f2233f4b1d6ff2942f641.

I've also gone ahead and removed the`related` attribute of InlinePanel, since it just duplicates `db_field`. (4cdf7c474e46db819c753f945a4e242bcfaed605).

It's slightly unfortunate that `comparison.field_label()` now returns 'speaker' rather than 'speakers', since the plural is the appropriate version there... I guess we ought to pick up the label from the `InlinePanel` definition where possible. Of course, that's not a regression, since it didn't work at all before - so that can be something for another PR :-)

Issue #4274, opened 2018-02-12
Known issues with the Draftail rich text WYSIWYG editor and its dependencies


thibaudcolas:
Draftail is based on Draft.js, and most of the editor's behavior (and bugs) comes from the framework. In the interest of tracking potential problems with the end product, here are Draft.js issues known to affect Draftail within Wagtail according to my testing.

**When reporting an issue on Draftail, hopefully this issue will stand out and help reporters track down the problem.**

> Note: this is a near-copy of https://github.com/springload/draftail/issues/138

## Draftail

Have a look at the [issue tracker](https://github.com/springload/draftail/issues).

### Copy-pasting

* Pasting in from Google Docs - copying and pasting a horizontal rule/line does not work. Start an empty Google Doc, Insert > Horizontal Line, select al, copy, paste into draftail - nothing appears.

## [Draft.js](https://github.com/facebook/draft-js/issues)

In order not to spam the Draft.js issue tracker, the links are kept in a code snippet instead of being rendered.

### Copy-pasting

```markdown
* Sequential unstyled blocks are merged into the same block on paste https://github.com/facebook/draft-js/pull/927, https://github.com/facebook/draft-js/issues/738, https://github.com/facebook/draft-js/issues/523, https://github.com/facebook/draft-js/issues/1082
* Rich text pasting creates an empty block at the beginning of the pasted content https://github.com/facebook/draft-js/pull/1052
* Pasting `<pre><code>` should not add the CODE inline style https://github.com/facebook/draft-js/issues/394
* Copy pasting from one code block to another removes line breaks https://github.com/facebook/draft-js/issues/407
* Edge copy and paste issue (and IE11) https://github.com/facebook/draft-js/issues/659
* Copying text in IE adds extra new lines https://github.com/facebook/draft-js/issues/615, https://github.com/facebook/draft-js/issues/1448
* Cannot paste HTML in IE https://github.com/facebook/draft-js/issues/986
* Caret focus position incorrect after pasting text in IE10/11 https://github.com/facebook/draft-js/issues/612
* Cannot paste text with angle brackets in IE https://github.com/facebook/draft-js/issues/656
* Copy/paste of entities between editors https://github.com/facebook/draft-js/issues/787
* Copy/paste between editors strips soft returns https://github.com/facebook/draft-js/issues/1154
* Minor issue with copy-pasting rich text styles https://github.com/facebook/draft-js/issues/1163
[...]

PR #4582, opened 2018-06-04
Fix copy-paste between editors losing formatting/content #4432

### LINKED ISSUE #4432: copy paste from rich text editor X to rich text editor Y loses links/hr
I read through https://github.com/wagtail/wagtail/issues/4274 but we're running into a situation where we have a streamfield with richtext, and creating two richtext blocks, then copy/pasting from one richtext block to the other loses any link information, as well as `<hr>` elements, and from what I can tell, also empty paragraphs.

Taking content from textbox 1 that has HTML representation:
```html
<p>test1</p>
<p></p>
<hr>
<p><a href=...>test2</a></p>
```
instead becomes the following content when copy-pasted into textbox 2:
```html
<p>test1</p>
<p>test2</p>
```

thibaudcolas:
Fixes #4432. This pull request has been made possible by the backing of my [Patreon patrons](https://www.patreon.com/thibaud_colas).

The current copy-paste behavior is a mixture of the browser's default copy behavior, and the Draft.js paste processing.

With this Draftail update, we now override the defaults for both copy and paste between editors:

- Copy should fill the clipboard with the same markup as before, except with a serialised version of the Draft.js content included, and with less inline styles (can lead to some discrepancies in how word processors receive the pasted content)
- Paste should work the same in all scenarios, except when the content comes from a Draftail editor. There, it will insert the Draft.js content directly instead of its HTML representation.

draftail-copy-paste-4432

Tested in: 

- [x] Windows 10, IE11 (no support, default "plain text" behavior)
- [x] Windows 10, Edge 17
- [x] macOS 10.13, Firefox latest
- [x] macOS 10.13, Chrome latest
- [x] macOS 10.13, Safari latest
- [x] iOS 11.3 Safari
- [x] Android Chrome 7.1

---

This fixes quite a few of the known issues with Draft.js (#4274), see the potential diff over at https://github.com/springload/draftail/issues/147#issuecomment-394202357. I'll update the list myself once this PR is merged.

The underlying code changes are:

- https://github.com/thibaudcolas/draftjs-conductor/pull/2 (new copy-paste processing)
- https://github.com/thibaudcolas/draftjs-conductor/pull/3 (bug fix)
- https://github.com/springload/draftail/pull/155 (integration into the editor)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2619,23 +2619,23 @@
       }
     },
     "draftail": {
-      "version": "0.17.1",
-      "resolved": "https://registry.npmjs.org/draftail/-/draftail-0.17.1.tgz",
-      "integrity": "sha512-Yjs38vfGc+0jaT8Oe9HyRyWS/L/uJEnsW0A4IO5qD1XExbc3uVHSfEa+/ozsFC1vHTFKGDX5ixSkKPxCNTmp7A==",
+      "version": "0.17.2",
+      "resolved": "https://registry.npmjs.org/draftail/-/draftail-0.17.2.tgz",
+      "integrity": "sha512-SPmd9k9cMH/4UmXK6F7uRx1sCtebcO2qFXjXA1R6BaR9gpCqfcKefxVZbgJ9RELpZ28xvNtTEjdtJqzUI1UPaw==",
       "requires": {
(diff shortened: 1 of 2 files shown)


thibaudcolas:
🎉

PR #4583, opened 2018-06-04
Rewrite modal-workflow to not require dynamic script evaluation

gasman:
Refactor our modal-workflow code so that all Javascript onload code is declared upfront in the `ModalWorkflow` initialiser, rather than pushed as part of the AJAX response. This means that our AJAX responses become pure JSON and can be handled with `JSON.parse` instead of `eval`, eliminating a major source of CSP-non-compliance within Wagtail (#1288, #4335).

Incorporates #4566.

Since there's existing third-party code that relies on the old modal-workflow API (e.g. https://github.com/takeflight/wagtailmodelchooser), the js_template / eval route is deprecated here rather than being removed outright. However, note that any existing third-party code which calls Wagtail's built-in modal workflows using the old-style method calls will be broken by this change - alternative rich text editors are likely to fall into this category. (Are there any such packages still being developed as of Wagtail 2.0...?)


wagtail/admin/static_src/wagtailadmin/js/modal-workflow.js
```diff
@@ -51,22 +54,36 @@ function ModalWorkflow(opts) {
         });
     };
 
-    self.loadResponseText = function(responseText) {
-        var response = eval('(' + responseText + ')');
+    self.loadResponseText = function(responseText, textStatus, xhr) {
+        /* RemovedInWagtail24Warning - support for eval()-ing text/javascript responses
```
m1kola: I would drop it now. It is an internal API. People who have decided to use it, knew what they were doing.

gasman:
Merged in fee77469b381d9676729c2281ed8b985a7ae94b2 + parents

PR #4585, opened 2018-06-05
Allow users to edit first name and last name

chrisranjana:
Allow users to edit first name and last name


zerolab:
This requires input from the UX team, but the practice is to update your details in one place. It is rather odd to change your first name and last name on different pages.

I would even argue first/last name and email should all go on the same form.

gasman:
Thanks @chrisranjana!

As @zerolab says, the first name and surname fields should really be part of the same form. Admittedly the account settings area as a whole is more fragmented than we'd like (due to it being extended by different developers at different times), and fixing that would be out of scope for this pull request...

Also, please could you add test coverage for the new form, within https://github.com/wagtail/wagtail/blob/master/wagtail/admin/tests/test_account_management.py?

chrisranjana:
Hello @gasman 

I have merged first name last name into a single form and also created a new test.

lb-:
@chrisranjana - thank you for submitting this, however a more recent PR for the same fix came through and I did not see this PR at the time. Closing as this was implemented in PR #5425 & merged in via https://github.com/wagtail/wagtail/commit/31e3dab4ec3ddad9d18e342df206ca7dcc925e8f

Issue #2056, opened 2015-12-17
Tags input doesn't works properly with russian keyboard layout.


MechanisM:
You know, russian letter "Б б" is on the same key as comma in english keyboard(I'm using macbook pro), so when I'm trying to add some tag in russian and typing the word ex. "победа" it suddenly creates tag for first part of the word before letter "б" - "по" and continue to type other letters as other tag: "еда".
So instead of having just one tag as wanted "победа", I'm getting two tags: "по" and "еда".
I can only enter such tags via copy/pasting full word into field.

MechanisM:
This one works great with russian https://maxfavilli.com/jquery-tag-manager

MechanisM:
What do you think is good idea to create hallo.js plugin, to select some word in text and mark as tag?

gasman:
Let's stick to one issue at once :-)

This would be an issue with the jQuery Tag-it library, and it looks like there's a patch here that we can try: https://github.com/aehlke/tag-it/issues/324

Switching to a different tagging library is probably more trouble than it's worth - but it's something to consider if we encounter other problems with Tag-it in future (and if Tag-it isn't actively developed any more, as it seems right now).

thibaudcolas:
Hey @sky-code, this is an open-source project – if you're running into this you're more than welcome to propose a pull request with the change suggested above. I imagine this hasn't been fixed yet because there aren't many Wagtail contributors that use russian keyboard layouts, but if you're one of them we'll happily review your change and merge it if it works.

cra:
Looking at the suggested fix, I can only spot one-line difference https://github.com/wagtail/wagtail/blob/c7a418b1c360204ab8393386a2dd72312e21993a/wagtail/admin/static_src/wagtailadmin/js/vendor/tag-it.js#L244

So, a work-around might be to just comment this line out (basically, the COMMA part). I can confirm, this allows to use this with TAB and ENTER still recognized as tag tokenizer.

Is there a way to set up a flag for tagit from settings.py or somewhere python-friendly?

PR #4586, opened 2018-06-05
Add an option to disable comma on tag input (#2056)

cra:
Here's an option to disable COMMA when entering tags. It is mainly to be able to use the UI with Cyrillic layout, but as a side-effect allows for tags with commas in them.


cra:
@jjanssen is there anything I can help with to speed up the process?

PR #4587, opened 2018-06-06
Add max_length option to redirect URL field

mwharrison:
The Django URL field defaults to 200 characters. This has bitten us a few times with really long URLs.

Since Django's URL field is just a fancy Char field - this PR sets the `max_length` option to 255 (the recommended maximum to work with all supported databases).


gasman:
Merged in e2b1c66a923703ae68f09d9731393dda184cf5f2 - thanks @mwharrison!

PR #4593, opened 2018-06-08
Update search query classes for RFC 25

kaedroho:
Updating the search query classes to bring all backends in line with RFC 25: https://github.com/wagtail/rfcs/blob/master/text/025-search-query-classes-revised.md

@BertrandBordage Would it be possible to make ``PlainText`` queries map directly to ``plain_to_tsvector`` calls?

Things left to be done on search query classes:
 - [ ] Make ``PlainText()`` ``operator`` argument default to the operator used by the backend
 - [ ] Implement ``Boost()`` on database/postgresql backends
 - [ ] Documentation


kaedroho:
Probably should leave this undocumented until we implement boost in the database search backends

wagtail/contrib/postgres_search/backend.py
```diff
@@ -226,36 +225,41 @@ def get_search_field(self, field_lookup, fields=None):
...
-        if isinstance(query, Term):
-            self.check_boost(query)
-            return PostgresSearchQuery(unidecode(query.term), config=config)
+        if isinstance(query, PlainText):
+            self.check_boost(query, boost=boost)
+
+            operator = {
+                'and': AND,
+                'or': OR,
+            }[query.operator]
```
BertrandBordage: That dictionary could be moved as a class attribute for performance and consistency with `DEFAULT_OPERATOR`.

wagtail/search/backends/db.py
```diff
@@ -51,36 +51,45 @@ def build_single_term_filter(self, term):
...
         if query is None:
             query = self.query
 
+        if isinstance(query, PlainText):
+            self.check_boost(query, boost=boost)
+
+            operator = {
+                'and': AND,
+                'or': OR,
+            }[query.operator]
```
BertrandBordage: That dictionary could be moved as a class attribute for performance and consistency with `DEFAULT_OPERATOR`.

wagtail/contrib/postgres_search/backend.py
```diff
@@ -226,36 +225,41 @@ def get_search_field(self, field_lookup, fields=None):
...
+            return AND(self.build_database_query(subquery, config, boost=boost)
                        for subquery in query.subqueries)
         if isinstance(query, Or):
-            return OR(self.build_database_query(subquery, config)
+            return OR(self.build_database_query(subquery, config, boost=boost)
                       for subquery in query.subqueries)
         raise NotImplementedError(
             '`%s` is not supported by the PostgreSQL search backend.'
-            % self.query.__class__.__name__)
+            % query.__class__.__name__)
```
BertrandBordage: Nice catch! :frowning:

BertrandBordage:
Also, note that implementing `Boost` on PostgreSQL means we have to give a weight to a `tsquery` object, which is not possible using `plainto_tsquery`:
```sql
SELECT plainto_tsquery('hello:B world:A');             -- Yields "'hello' & 'b' & 'world' & 'a'"
SELECT setweight(plainto_tsquery('hello world'), 'A'); -- Raises an error
SELECT to_tsquery('hello:B & world:A');                -- Yields "'hello':B & 'world':A"
```

kaedroho:
@BertrandBordage Couldn't you multiply the rank that's output when running the query?

BertrandBordage:
@kaedroho The ranking score is computed per row, not per query part, so it doesn’t fit the `Boost` API that alllows only a piece of a query to be boosted.

kaedroho:
Is it possible to do something like this?

Convert this query:
```
PlainText("hello world") | Boost(PlainText("foo bar baz"), 10)
```

into this:
```
ts_rank(field, plainto_ts_query('hello world')) + ts_rank(field, plainto_ts_query('foo bar baz')) * 10 AS rank
```

BertrandBordage:
It’s possible, but that’s not efficient because two `ts_rank` are executed, and ranking is by far what takes the most time when searching with PostgreSQL, so I don’t advise doing it.
That being said, it’s worth benchmarking, because it’s possible that PostgreSQL optimizes the query by merging multiple `ts_rank` into a single operation. I highly doubt it, but it’s still plausible.

PR #4596, opened 2018-06-10
Update editors guide - inserting images

### LINKED ISSUE #4557: In the docs, the Editor's guide refers to image formats as "alignment"
### Issue Summary

In the Editor's guide section of the docs, [Inserting images and videos in a page](http://docs.wagtail.io/en/v2.0.1/editor_manual/new_pages/inserting_images.html?highlight=alignment) refers to the rich text image formats as "alignment" whereas all other documentation and UI uses the term "format".

### Steps to Reproduce

In the docs:

> In addition, Wagtail allows you to choose an alignment for your image.
> [...]
> The alignments available are described below:

I imagine this is because some of the default format options ("Left-aligned", "right-aligned") are alignments, but this is only the default ones, and "Full-width" can't really be thought of as an alignment.

Even the UI screenshot straight from Wagtail says "Choose a format".

lb-:
Closes #4557 - Update terminology to reflect 'format' instead of 'alignment'


thibaudcolas:
Thank you @lb- 👏

Issue #4598, opened 2018-06-11
Alternative on pageurl


Speedy1991:
I patched pageurl to take an fallback viewname if page is None.
Usecase: It's easy to implement comming soon pages or alternative pages if the page is not available (yet).  I don't know if this is useful for others.
Let me know if I should start a PR if you guys are interested.
```
@register.simple_tag(takes_context=True)
def pageurl(context, page, fallback=None):

    if not page and fallback:
        return reverse(fallback)

    try:
        current_site = context['request'].site
    except (KeyError, AttributeError):
        return page.url

    return page.relative_url(current_site, request=context.get('request'))
```
Example calls:
```
<li><a href="{% pageurl page.download_page 'coming_soon' %}">{% trans "Download" %}</a></li>
<li><a href="{% pageurl page.profile_page 'accounts:mock' %}">{% trans "Profile" %}</a></li>
```

gasman:
Thanks @Speedy1991! This looks like a useful addition, and a PR would be very welcome.

I'd suggest that the `fallback` parameter should be a keyword argument rather than a positional one, e.g. `{% pageurl page.download_page fallback='coming_soon' %}` - so that it's more understandable to someone reading the template, and leaves open the possibility of us extending the `{% pageurl %}` tag in other ways in future. (Perhaps this works already with the existing code...)

gasman:
Implemented in #4603

PR #4603, opened 2018-06-12
Added fallback to pageurl

Speedy1991:
I can add a test if you want. Just need a hint where to place it. I'm bit lost in the tests package.
#4598

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -291,7 +291,6 @@ Contributors
 * Alejandro Garza
 * Rajeev J Sebastian
 * Coen van der Kamp
-* Arthur Holzner
 
 Translators
 ===========
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -291,6 +291,7 @@ Contributors
 * Alejandro Garza
 * Rajeev J Sebastian
 * Coen van der Kamp
+* Arthur Holzner
 
 Translators
 ===========
@@ -340,4 +341,3 @@ Translators
 * Ukrainian: Viktor Shytiuk, Mykola Zamkovoi
 * Vietnamese: Luan Nguyen, Nguyễn Hồng Quân
 * Welsh: Adam Hughes
-* Arthur Holzner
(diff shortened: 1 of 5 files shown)


CONTRIBUTORS.rst
```diff
@@ -291,6 +291,7 @@ Contributors
 * Alejandro Garza
 * Rajeev J Sebastian
 * Coen van der Kamp
+* Arthur Holzner
```
gasman: You're already on the list, four names above :-)

gasman:
Thanks @Speedy1991! The test would go in `TestPageUrlTags`: https://github.com/wagtail/wagtail/blob/ac8905b125ac65960e049f714f1eff3eea9237dc/wagtail/core/tests/tests.py#L13

CONTRIBUTORS.rst
Speedy1991: eh.. yea
i ll fix this with the upcoming test

Speedy1991:
Where to put the docs? As far i can see, `pageurl` has no seperate entry in the docs. Is it worth a new section? Or just put it anywhere in the _get started_?

Speedy1991:
Friendly ping @gasman

gasman:
Hi @Speedy1991 - sorry for the delay in getting back to you. The appropriate place to document this would be the "writing templates" page: https://github.com/wagtail/wagtail/blob/master/docs/topics/writing_templates.rst#pageurl

wagtail/core/templatetags/wagtailcore_tags.py
```diff
@@ -11,11 +12,17 @@
 
 
 @register.simple_tag(takes_context=True)
-def pageurl(context, page):
+def pageurl(context, page, **kwargs):
```
gasman: `def pageurl(context, page, fallback=None)` would be better here, instead of retrieving `fallback` from `kwargs` - this way, we avoid silently accepting all keyword arguments, and users will get the expected error "'pageurl' received unexpected keyword argument" if they misspell `fallback` (for example).

gasman:
Docs added and merged in 6f6f9b8e469881b21bea22bd68ddb4ab587083e9 - thanks @Speedy1991!

PR #3802, opened 2017-08-26
675 deleting snippets enhancement

lb-:
First draft of pull request to resolve #675 

I have added tests, docs (for the editor's guide only) and functionality as per the issue.

Let me know what needs changing or if the tests are suitable.

PR #4611, opened 2018-06-14
Make checkbox/radio alignment on Firefox consistent with Webkit

gasman:
As reported at https://github.com/wagtail/wagtail/pull/3802#issuecomment-397003833, checkboxes and radio buttons are misaligned in Firefox.

(Related: #4529)

The culprit appears to be the `-webkit-appearance` rule on these fields: on the face of it this rule seems to be pretty meaningless since we go on to hide the native widget behind the fake one, but it has the effect of making it occupy a 12x12px box only on Webkit, which then becomes the basis for the relative position of the fake element. This PR forms that 12x12 box using height and margin rules instead.

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/scss/components/_forms.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/components/_forms.scss
@@ -140,11 +140,10 @@ input[type=checkbox] {
 }
 
 input[type=radio] {
-    // stylelint-disable-next-line property-no-vendor-prefix
-    -webkit-appearance: radio;
+    height: 12px;
     width: auto;
     position: relative;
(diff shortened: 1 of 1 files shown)


BertrandBordage:
Looks great from the look! Thanks @gasman!
I note a centering issue however on the selected radio button: the top padding seems to be `6px` while the bottom one seems to be `4px`. It may be coming from some antialiasing though.
41412397-c5929a54-6fd7-11e8-80d1-d0c6b1a8c7ac

gasman:
Hi @BertrandBordage,

Yep, I agree it looks off-centre, but it was equally bad before, so I regard it as out-of-scope for this ticket :-)

Issue #3897, opened 2017-10-03
Make Promoted Search Results Orderable


mope:
In the settings menu there are sections for Users, Groups, Sites, Collections, Redirects and Promoted Search Results. In the Users, Groups, Sites and Redirects sections it is possible to reorder items by clicking on column titles. This is not possible in the Collections and Promoted Search Results sections. Collections are [under development](https://github.com/wagtail/wagtail/pull/3407) and so the inconsistency can be ignored in that case. However it would be useful and consistent if Promoted Search Results could be ordered by number of views, and perhaps search terms as well.

lb-:
Implemented via #5299

PR #4614, opened 2018-06-15
added sorting facility to search promotions issue 3897

chrisranjana:
https://github.com/wagtail/wagtail/issues/3897

Added sort function to search promotions

### DIFF
--- a/wagtail/contrib/search_promotions/templates/wagtailsearchpromotions/list.html
+++ b/wagtail/contrib/search_promotions/templates/wagtailsearchpromotions/list.html
@@ -5,9 +5,23 @@
     <col />
     <thead>
         <tr>
-            <th class="title">{% trans "Search term(s)" %}</th>
+            <th class="title">
+			{% trans "Search term(s)" %}
+			{% if ordering == "query_string" %}
+                    <a href="{% url 'wagtailsearchpromotions:index' %}" class="icon icon-arrow-down-after teal"></a>
+                {% else %}
+                    <a href="{% url 'wagtailsearchpromotions:index' %}?ordering=query_string" class="icon icon-arrow-down-after"></a>
+                {% endif %}
+			</th>
             <th>{% trans "Promoted results" %}</th>
-            <th>{% trans "Views (past week)" %}</th>
+            <th>
+			{% trans "Views (past week)" %}
+			{% if ordering == "hits" %}
+                    <a href="{% url 'wagtailsearchpromotions:index' %}" class="icon icon-arrow-down-after teal"></a>
+                {% else %}
+                    <a href="{% url 'wagtailsearchpromotions:index' %}?ordering=hits" class="icon icon-arrow-down-after"></a>
+                {% endif %}
+			</th>
         </tr>
     </thead>
     <tbody>
--- a/wagtail/contrib/search_promotions/views.py
+++ b/wagtail/contrib/search_promotions/views.py
@@ -2,6 +2,7 @@ from django.shortcuts import get_object_or_404, redirect, render
 from django.urls import reverse
 from django.utils.translation import ugettext as _
 from django.views.decorators.vary import vary_on_headers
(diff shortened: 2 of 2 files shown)


chrisranjana:
Can someone restart travis-ci checks ?

wagtail/contrib/search_promotions/views.py
```diff
@@ -1,8 +1,10 @@
+from django.db.models import Sum
 from django.shortcuts import get_object_or_404, redirect, render
 from django.urls import reverse
 from django.utils.translation import ugettext as _
 from django.views.decorators.vary import vary_on_headers
 
+
```
lb-: Recommend the additional whitespace be removed, it should not be needed

wagtail/contrib/search_promotions/views.py
```diff
@@ -21,9 +23,14 @@
 def index(request):
     is_searching = False
     query_string = request.GET.get('q', "")
-
```
lb-: These changes did not work for me initially, not sure why but I have made some changes locally that appear to work, see below.

I hope this helps, there are a lot of ways to do this but I borrowed from some other ordering code in other Wagtail views.

```python
def index(request):
    is_searching = False
    query_string = request.GET.get('q', '') # unrelated - just standardising on single quotes

    # Ordering
    valid_ordering = ['query_string', '-query_string', 'ohits', '-ohits'] # set up 'valid' values early to avoid errors
    ordering = valid_ordering[0] # set the 'default' value in one place

    if 'ordering' in request.GET and request.GET['ordering'] in valid_ordering:
        ordering = request.GET['ordering'] # here is the only scenario where ordering would be used from url params

    # Query
    queries = Query.objects.filter(editors_picks__isnull=False).distinct() # set up 'base' query in one place

    if 'ohits' in ordering: # standardise on 'ohits' in both query and the annotation for clarity
        queries = queries.annotate(ohits=Sum('daily_hits__hits')) # only here do we do the annotation

    queries = queries.order_by(ordering) # finally we do the actual 'ordering' based on the above
    # ... rest below (from Search comment)
```

wagtail/contrib/search_promotions/templates/wagtailsearchpromotions/list.html
```diff
@@ -5,9 +5,23 @@
     <col />
     <thead>
```
lb-: A few comments

* There is no way to 'reverse' the ordering of each item, this is something that would be really valuable
* The nesting is a bit odd, would be good to revise and align with other template files
* Using something like `'ohits' in ordering` we can save ourselves some code duplication

See below for a bit of a suggested rework


```
    <thead>
        <tr>
            <th class="title">
                {% trans "Search term(s)" %}
                {% if "query_string" in ordering  %}
                    <a href="{% url 'wagtailsearchpromotions:index' %}?ordering={% if '-' not in ordering %}-{% endif %}query_string" class="icon icon-arrow-{% if '-' in ordering %}up{% else %}down{% endif %}-after teal"></a>
                {% else %}
                    <a href="{% url 'wagtailsearchpromotions:index' %}?ordering=query_string" class="icon icon-arrow-down-after"></a>
                {% endif %}
			</th>
            <th>{% trans "Promoted results" %}</th>
            <th>
                {% trans "Views (past week)" %}
                {% if "ohits" in ordering %}
                    <a href="{% url 'wagtailsearchpromotions:index' %}?ordering={% if '-' not in ordering %}-{% endif %}ohits" class="icon icon-arrow-{% if '-' in ordering %}up{% else %}down{% endif %}-after teal"></a>
                {% else %}
                    <a href="{% url 'wagtailsearchpromotions:index' %}?ordering=ohits" class="icon icon-arrow-down-after"></a>
                {% endif %}
			</th>
        </tr>
    </thead>
```

lb-:
Changes implemented in new PR #5299 - just adding to the work done by @chrisranjana

PR #4617, opened 2018-06-15
Remove old-skool font-support since woff and woff2 would be sufficient

jjanssen:
So this PR removes some references to some ancient browser support fonts. Nowadays woff & woff2 support would be sufficient enough according to [our browser support](https://caniuse.com/#search=woff). Woff2 is not available for icomoon unless you use a paid tier, which is essentially nothing more than a more compressed woff format.

Tested in:
- Internet Explorer 11;
- Firefox ESR (52);
- Chrome (67, latest); 
- Safari (11.1.1);
- iOS (11);


gasman:
Thanks @jjanssen!

PR #4618, opened 2018-06-15
Update Unidecode version

craigloftus:
A straightforward version bump, motivated by dependency conflicts with `django-oscar==1.6`. Unidecode has switched to semantic versioning, so the size of the version change doesn't reflect the scope of the changes.

I'm not sure if `<2.0` is the next logic place for the upper bound, but that seemed to be the convention used with the other dependencies?


gasman:
Looks good to me - thanks @craigloftus!

PR #4619, opened 2018-06-15
Improving Wagtail V2 docs

patphongs:
PR for improving wagtail upgrade documentation for users to be able to upgrade to wagtail 2.0. This includes tips and tricks for upgrading.

### DIFF
--- a/docs/upgrading_wagtail_v2/index.rst
+++ b/docs/upgrading_wagtail_v2/index.rst
@@ -8,7 +8,7 @@ Wagtail 2.0 Requirements
 * Python versions: 3.4, 3.5, 3.6
     - It's recommended that you upgrade your python version first to be compatible with python 3.4.
     
-For more requirements, see :ref:`compatible_django_python_versions` table
+For more requirements, see :ref:`Compatible Django / Python versions` table
 
 Upgrade process
 ===============
@@ -21,7 +21,7 @@ We recommend upgrading one feature release at a time, even if your project is se
 Before upgrading to a new feature release:
 
 * Check your project's console output for any deprecation warnings, and fix them where necessary;
-* Check the new version's release notes, and the :ref:`compatible_django_python_versions` table below, for any dependencies that need upgrading first;
+* Check the new version's release notes, and the :ref:`Compatible Django / Python versions` table below, for any dependencies that need upgrading first;
 * **IMPORTANT** Make a backup of your database. You may find yourself dropping and restoring your local database several times in order to get the django migrations to run correctly through the upgrades.
 
 -----------
--- a/docs/upgrading_wagtail_v2/index.rst
+++ b/docs/upgrading_wagtail_v2/index.rst
@@ -8,7 +8,7 @@ Wagtail 2.0 Requirements
 * Python versions: 3.4, 3.5, 3.6
     - It's recommended that you upgrade your python version first to be compatible with python 3.4.
     
-For more requirements, see `Compatible Django / Python versions` table
+For more requirements, see `Compatible Django / Python versions`_ table
 
 Upgrade process
 ===============
--- a/docs/upgrading_wagtail_v2/index.rst
+++ b/docs/upgrading_wagtail_v2/index.rst
@@ -67,46 +67,6 @@ For the full list of command line options, enter ``wagtail help updatemodulepath
 
 You are advised to take a backup of your project codebase before running this command. The command will perform a search-and-replace over all \*.py files for the affected module paths; while this should catch the vast majority of module references, it will not be able to fix instances that do not use the dotted path directly, such as ``from wagtail import wagtailcore``.
 
-The full list of modules to be renamed is as follows:
-
-+-----------------------------------------+-----------------------------------+-----------------------------------+
-| Old name                                | New name                          | Notes                             |
-+=========================================+===================================+===================================+
-| wagtail.wagtailcore                     | wagtail.core                      |                                   |
-+-----------------------------------------+-----------------------------------+-----------------------------------+
-| wagtail.wagtailadmin                    | wagtail.admin                     |                                   |
-+-----------------------------------------+-----------------------------------+-----------------------------------+
-| wagtail.wagtaildocs                     | wagtail.documents                 | 'documents' no longer abbreviated |
-+-----------------------------------------+-----------------------------------+-----------------------------------+
-| wagtail.wagtailembeds                   | wagtail.embeds                    |                                   |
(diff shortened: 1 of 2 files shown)


docs/upgrading_wagtail_v2/index.rst
```diff
@@ -0,0 +1,158 @@
+==============================
+Upgrading to Wagtail Version 2
+==============================
+
+Wagtail 2.0 Requirements
+========================
+* Django versions: 1.11, 2.0 
+* Python versions: 3.4, 3.5, 3.6
+    - It's recommended that you upgrade your python version first to be compatible with python 3.4.
```
allcaps: What do you exactly mean? What is compatible with 3.4? Is python 3.6 compatible? Should I downgrade? Also you write 'first' so what is 'second'? Maybe:

Make sure your project runs on Python 3.4 or higher before upgrading Django and Wagtail.

Or do you mean something else?

docs/upgrading_wagtail_v2/index.rst
```diff
@@ -0,0 +1,158 @@
...
+    
+For more requirements, see `Compatible Django / Python versions`_ table
+
+Upgrade process
+===============
+We recommend upgrading one feature release at a time, even if your project is several versions behind the current one. This has a number of advantages over skipping directly to the newest release:
+
+* If anything breaks as a result of the upgrade, you will know which version caused it, and will be able to troubleshoot accordingly;
+* Deprecation warnings shown in the console output will notify you of any code changes you need to make before upgrading to the following version;
+* Some releases make database schema changes that need to be reflected on your project by running ``./manage.py makemigrations`` - this is liable to fail if too many schema changes happen in one go.
```
allcaps: > this is liable to fail if too many schema changes happen in one go.

I don't think this is true. All migrations are executed in order. So running `manage.py migrate` is the same as running `manage.py migrate app 001 && `manage.py migrate app 002 && `manage.py migrate app 003```. But `manage.py migrate` is better as migration dependencies are respected.

gasman: This text is copied from the existing page at http://docs.wagtail.io/en/v2.1/releases/upgrading.html. It's referring to things like custom image models, where a change within Wagtail requires a corresponding migration to be created within the user project, using `./manage.py makemigrations`. This relies on Django's migration autodetection logic, which can't handle too many changes at once.

IIRC, this came up around Wagtail 1.8-1.10 when the `filter` field on image renditions was changed from a foreign key to a string - we had to add and populate the new field, switch over the unique constraints, and drop the old field, in three separate migrations over three releases. Skipping straight to the last release would mean you'd end up with one migration containing all of those changes.

docs/upgrading_wagtail_v2/index.rst
```diff
@@ -0,0 +1,158 @@
...
+========================
+* Django versions: 1.11, 2.0 
+* Python versions: 3.4, 3.5, 3.6
+    - It's recommended that you upgrade your python version first to be compatible with python 3.4.
+    
+For more requirements, see `Compatible Django / Python versions`_ table
+
+Upgrade process
+===============
+We recommend upgrading one feature release at a time, even if your project is several versions behind the current one. This has a number of advantages over skipping directly to the newest release:
```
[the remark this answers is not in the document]
gasman: This text is also taken from http://docs.wagtail.io/en/v2.1/releases/upgrading.html, which clarifies what we mean by a feature release: one that increments the second number, e.g. 1.12 -> 1.13. "One release at a time" could imply that you have to step through 1.12, 1.12.1, 1.12.2... which definitely isn't the intention :-)

gasman:
Hi @patphongs - can you give some more details about the intended purpose of this page, please?

I see it's duplicating a lot of content from the 2.0 release notes and "Upgrading Wagtail" pages, and as a result it seems to switch between 2.0-specific instructions and general advice, in ways that aren't entirely consistent. (For example, if this page is specifically about 2.0, then including the full "Compatible Django / Python versions" table isn't really necessary.)

docs/upgrading_wagtail_v2/index.rst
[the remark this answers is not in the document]
loicteixeira: Should it use semantic versioning terms then and call it `minor`?

docs/upgrading_wagtail_v2/index.rst
```diff
@@ -0,0 +1,158 @@
...
+2. Create another branch that adds the necessary failed test dependency to your intial home migration. Merge this second and deploy. 
+
+After following the 2 deployment steps above, that will upgrade you to the most recent version of Wagtail with passing tests and migrations. You wont have to do this again unless there is another major upgrade to wagtail.
+
+We've Upgraded...what now?:
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+* Take a good look at all of your custom built applications to see if they still work in Wagtail 2.
+* Hallo vs Draftail: Since you cannot run both rich text editor libraries at the same time, you will need to choose between using Hallo or Draftail. HalloPlugins have been deprecated, and will be removed in Wagtail 1.14. So it would be best that you make the effort to attempt to transfer over to draftail now, this comes by default with Wagtail 2. 
+
+Compatible Django / Python versions
```
loicteixeira: Can this link to `docs/releases/upgrading.rst#compatible-django--python-versions` instead so we don't have to maintain the lists of compatible versions in 2 places?

docs/upgrading_wagtail_v2/index.rst
```diff
@@ -0,0 +1,158 @@
+==============================
```
loicteixeira: 1. This document should also be linked from `/docs/releases/2.0.rst` I guess.
1. Is `/docs/upgrading_wagtail_v2/` the right place? We already have `/docs/releases/upgrading.rst` so maybe it should leave alongside to it?

gasman:
Unfortunately I don't think this page is going to fit well into the documentation in its current form. Re-using content from the existing "Upgrading Wagtail" and release notes pages means that we risk making the documentation more confusing to users, since there's not just one set of instructions to follow, but multiple sets of instructions giving slightly inconsistent advice.

(For example, this page is effectively saying: "Don't try to upgrade multiple versions at once... but if you do, here are some workarounds for the problems you might encounter". That's going to generate more support tickets for us than simply saying "Don't try to upgrade multiple versions at once".)

If there's anything that needs to be improved in the current upgrade documentation, I'd suggest starting out by adding it to the existing 2.0 release notes page - we can then consider splitting it off into a sub-page if there's sufficient new information.

Issue #4620, opened 2018-06-15
Dev contribution docs aren't clear about running dev version of Wagtail


Scotchester:
I'm not sure I could really capture this succinctly in the issue title, but in getting set up to contribute to Wagtail at today's Wagtail Space U.S. sprint, I and a few others ran into some confusion about getting a local site (e.g., the bakery demo) to use the local version of Wagtail we're hacking on.

The [Contributing to Wagtail – Development](http://docs.wagtail.io/en/latest/contributing/developing.html#developing) doc page takes you through setting up the local instance of Wagtail and then says

> Any Wagtail sites you start up in this virtualenv will now run against this development instance of Wagtail. We recommend using the Wagtail Bakery demo site as a basis for developing Wagtail.

The problem is that if you've not already installed bakerydemo (or some other Wagtail site) in the same virtualenv where you just set up the dev version of Wagtail, when you go to install bakerydemo, it will uninstall your dev version of Wagtail and install the current stable release.

I suggest that this doc page say before you start to set up your dev version of Wagtail that you should already have bakerydemo or some other Wagtail site up and running before you get to `pip install -e` of the dev version of Wagtail.

strotherscott:
With a lot of help from @willbarton

gasman:
I'm never really a fan of instructions that deliver information out of order, like recipes that tell you to put things in a pre-heated oven... :-) Could I suggest something like:

> Keep in mind that the setup steps for a Wagtail site may include installing a release version of Wagtail, which will override the development version you've just set up. In this case, you should install the site before running the `pip install -e` step, or re-run that step after the site is installed.

That way, hopefully people will understand _why_ things have to be done in a certain order, rather than just telling them to do so.

strotherscott:
I'm over my head as I lost me advisor - maybe use this??  I wont commit it without more help.

Any Wagtail sites you start up in this virtualenv will now run against this development instance of Wagtail.  We recommend using the Wagtail Bakery demo site as a basis for developing Wagtail. Keep in mind that the setup steps for a Wagtail site may include installing a release version of Wagtail, which will override the development version you've just set up. In this case, you should install the site before running the pip install -e step, or re-run that step after the site is installed.

strotherscott:
I had more time - so I tried a new update of the short section of text - including @gasman recommended thoughts.

gasman:
Completed in #4633

PR #4622, opened 2018-06-15
wording Edit per #4620

strotherscott:
@scotchester recommended I suggest that this doc page say before you start to set up your dev version of Wagtail that you should already have bakerydemo or some other Wagtail site up and running before you get to pip install -e of the dev version of Wagtail.


gasman:
Completed in #4633

Issue #2943, opened 2016-08-24
Admin color themes and customization?


scuerda:
I'm working on customizing the wagtail admin for a deployment and am finding it a bit challenging to catch all the locations where color schemes (background, font, hover, etc.) need to be overridden. 

The basic color variables are easy to redefine in the _variables.scss partial, but it's quite challenging to either rebuild core.css from scratch and/or to generate a custom-core.css file that doesn't contain a lot of duplicate styles. This is primarily due to the structure of the imports. I'd say that finding a way to avoid 

``` css
#wagtail {
    @import '../../../../../client/scss/style';
}
```

in core.scss would probably be the easiest first step in terms of being able to build a theming stylesheet.

Are there any roadmap items for addressing the need of users to do this type of customization and if so, are there any ways to help move this forward?

gasman:
Hi @scuerda,
There's not currently any roadmap item for this - it isn't something we've needed. We'd be happy to consider a PR that implements it, though.

One major consideration here is that a standard installation of Wagtail (as installed through `pip install wagtail`) comes with the CSS files prebuilt, and will not include the Node-based toolchain for rebuilding them from SCSS. We presumably want to avoid overwriting files within the Wagtail package, so the neatest approach is probably to write the compiled CSS to an app `/static` directory within your Wagtail project - as long as that app appears above `wagtailadmin` etc in `INSTALLED_APPS`, those files should override the ones shipped with Wagtail. This in turn will involve some rewriting of the Node build scripts, and detailed documentation for setting it all up (since the steps for building the CSS go way beyond a "standard" Wagtail install).

loleg:
I am not sure I see the need to customize everything, but it would be nice to have at least a way of specifying the admin header color to make it easier for users to tell production / staging / test servers apart.

SalahAdDin:
:+1:

JimBrouzoulis:
I'm looking to do some refactoring in the scss files to simplify the themeing process. This may be a step forward regarding this issue. 

I'm thinking of having something similar to many popular frameworks with a primary color, secondary etc. Wagtail's default theme would then use `$color-teal` and `$color-salmon`. 

However, I have a bit more difficulty understanding the usage of the greys: `$color-grey-1`-`$color-grey-5`. In total there are 6 defined greys and several more `lighten` and `darken` versions.

- Is there a semantic meaning to the different versions or is it mostly esthetics?
- Do we need that many definitions of grey, can't we make do with say three versions?

[...]

PR #4624, opened 2018-06-15
Improve color contrast compliance of admin header (using the current Wagtail color palette)

marteki:
This PR addresses some of the items brought up in #2224 by @heymonkeyriot. 

There's currently very low contrast between foreground text and background colors. As [the WAVE Web Accessibility Tool](https://wave.webaim.org/) states, "adequate contrast is necessary for all users, especially users with low vision." Recommended ways to fix this are to:
- Increase the contrast between the foreground (text) color and the background color. Text and images of text should have a contrast ratio of at least 4.5:1.
- Large text - larger than 18 point (roughly 24px) or 14 point bold (roughly 18.66 px) - does not require as much contrast as smaller text, and only needs a contrast ratio of at least 3:1.

## Proposed change

screen shot 2018-06-15 at 2 39 55 pm

With slight shifts  to other colors in the existing Wagtail brand color palette, the admin header moves into greater compliance with WCAG AA and Section 508.

## Current state

screen shot 2018-06-15 at 2 31 46 pm

## Detailed list of changes

- Changes background of admin header from `color-teal` to `color-teal-darker`
- Changes color of links in admin header from `color-teal` to `color-teal-dark`
- Changes hover color of links in admin header from `color-teal-dark` to 20% darker than dark
- Changes the `h2` in admin header to be large enough to count as "large text", while still remaining smaller than the `h1` that it's semantically lower in hierarchy than.

There are other changes that need to be done to bring the admin into greater color compliance, but they can wait for other PRs.

## Thanks
Thanks to the following folks for their consultation and review on this: @raq929 @scotchester @cfarm @FlipperPA @tomdyson

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/scss/_variables.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/_variables.scss
@@ -48,7 +48,7 @@ $color-grey-5: #fafafa;
 $color-menu-text: #cacaca; // was #aaa wich was too low contrast
 
 $color-thead-bg: $color-grey-5;
-$color-header-bg: $color-teal; // #ff6a58;
+$color-header-bg: $color-teal-darker; // #ff6a58;
 $color-fieldset-hover: $color-grey-5;
 $color-input-border: $color-grey-4;
 $color-input-focus: #f4fcfc;
@@ -62,8 +62,8 @@ $color-button-no: $color-red;
 $color-button-no-hover: darken($color-button-no, 20%);
 $color-button-warning: $color-orange;
 $color-button-warning-hover: darken($color-orange, 20%);
-$color-link: $color-teal;
-$color-link-hover: $color-teal-dark;
+$color-link: $color-teal-dark;
+$color-link-hover: darken($color-teal-dark, 20%);
 
 $color-text-base: $color-grey-2;
 $color-text-input: $color-grey-1;
--- a/wagtail/admin/static_src/wagtailadmin/scss/core.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/core.scss
@@ -198,12 +198,14 @@ footer {
     @include clearfix();
     overflow: hidden;
     padding-top: 1.4em;
+    padding-bottom: 0.65em;
+    background: $color-teal-dark;
     font-size: 0.85em;
 
     li {
         display: block;
         float: left;
-        padding: 0.5em 1.3em;
+        padding: 0.5em 1.3em 0.65em;
         position: relative;
         text-decoration: none;
         color: $color-white;
@@ -418,7 +420,6 @@ footer,
 
     .breadcrumb {
         padding-top: 0;
-        background: $color-teal-darker;
         margin-left: -($desktop-nice-padding);
         margin-right: -($desktop-nice-padding);
 
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -11,6 +11,7 @@ Changelog
  * The `{% slugurl %}` template tag is now site-aware (Samir Shah)
  * Added `file_size` field to documents (Karl Hobley)
  * Update documentation (configuring Django for Wagtail) to contain all current settings options (Matt Westcott, LB (Ben Johnston))
+ * Added `defer` flag to `PageQuerySet.specific` (Karl Hobley)
  * Fix: Handle all exceptions from `Image.get_file_size` (Andrew Plummer)
  * Fix: Fix display of breadcrumbs in ModelAdmin (LB (Ben Johnston))
 
--- a/docs/releases/2.2.rst
+++ b/docs/releases/2.2.rst
@@ -20,6 +20,7 @@ Other features
  * The ``{% slugurl %}`` template tag is now site-aware (Samir Shah)
  * Added ``file_size`` field to documents (Karl Hobley)
  * Update documentation (configuring Django for Wagtail) to contain all current settings options (Matt Westcott, LB (Ben Johnston))
+ * Added ``defer`` flag to ``PageQuerySet.specific`` (Karl Hobley)
 
 Bug fixes
 ~~~~~~~~~
--- a/wagtail/admin/static_src/wagtailadmin/scss/components/_header.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/components/_header.scss
@@ -61,7 +61,7 @@ header {
         @media screen and (max-width: $breakpoint-mobile) {
             // To all hamburger menu to be visible
             padding-left: $desktop-nice-padding;
-            padding-right: 0px;
+            padding-right: 0;
             padding-top: 11px;
         }
 
--- a/wagtail/admin/static_src/wagtailadmin/scss/components/_header.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/components/_header.scss
@@ -26,6 +26,10 @@ header {
         }
     }
 
+    h2 {
+        font-size: 1.9em;
+    }
+
     .col {
         float: left;
         margin-right: 2em;
--- a/wagtail/admin/static_src/wagtailadmin/scss/components/_header.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/components/_header.scss
@@ -61,6 +61,7 @@ header {
         @media screen and (max-width: $breakpoint-mobile) {
             // To all hamburger menu to be visible
             padding-left: $desktop-nice-padding;
+            padding-right: 0px;
             padding-top: 11px;
         }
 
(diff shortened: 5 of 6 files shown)


FlipperPA:
This looks great, @marteki - thanks for your efforts!

marteki:
@thibaudcolas Yes, these changes would pass WCAG AA for the elements changed in this PR.

You can see the differences and whether different combinations pass or fail by plugging the color hex codes into https://webaim.org/resources/contrastchecker/.

## The current state, with failing AA color combinations

White text on teal, at any size
screen shot 2018-06-16 at 9 51 34 am

Teal text on gray, at any size
screen shot 2018-06-16 at 9 52 45 am

## Changes in this PR

White text on darker teal, at increased font sizes
screen shot 2018-06-16 at 9 53 41 am

Dark teal on gray, at any size
screen shot 2018-06-16 at 9 57 09 am

marteki:
I'm looking now at shifts to make the header more complaint for other views, such as the editing page. The red warnings on this screenshot show the affected areas on the current admin:

screen shot 2018-06-16 at 10 51 18 am

@thibaudcolas and others: does it make sense to bring those into this PR, or should I keep them separate? If they should be kept apart, should I wait until this PR is fully checked and merged before filing the follow-up PR?

wagtail/admin/static_src/wagtailadmin/scss/_variables.scss
```diff
@@ -48,7 +48,7 @@ $color-grey-5: #fafafa;
 $color-menu-text: #cacaca; // was #aaa wich was too low contrast
 
 $color-thead-bg: $color-grey-5;
-$color-header-bg: $color-teal; // #ff6a58;
+$color-header-bg: $color-teal-darker; // #ff6a58;
```
marteki: I'm not sure what the comment on this line is referring to, so I didn't change it. The previous color was `#43b1b0`, and the PR changes it to `#358c8b`.

thibaudcolas:
@marteki I think that in this case a single PR will be easier to review. I'm not sure who is best placed to review this PR – probably a front-end developer and designer (ping @wagtail/ux-design).

cfarm:
With the new background in the header area, additional updates are needed to other elements on other views, such as the Edit view. Here's this PR:

screen shot 2018-06-16 at 11 42 07 am

Here's the latest version of the app:
screen shot 2018-06-16 at 11 44 34 am


To keep similar hierarchies and contrast in the color palette with this darker bg color, I'd recommend:

- Updating the background of `.breadcrumb` element at the top to be darker 
- Making the `.tab-nav` background color a darker teal
- Updating any hover styles on those elements as needed

Meteor0id:
This is a very good move to improve the default color theme of the editor interface. But I would like to point out to also keep in mind some developers might wish to alter the default color scheme. For that there is the ticket: https://github.com/wagtail/wagtail/issues/2943

It might be a good idea to write up some advise about contrast for the docs, when customization becomes available in an easy manner.

gasman:
I've now built a test instance for this PR: https://wagtail-pr-4624-stage.us.aldryn.io/en/admin/ (log in with `admin` / `changeme`)

ping @wagtail/ux-design @ollywillans

To chime in with a bit of feedback of my own - with the increased subheading size, the header bar now feels a bit unbalanced and could probably do with the text being pushed up a bit:

screen shot 2018-08-23 at 17 26 20

BertrandBordage:
The good thing about this is that we can decide to always use white as font color on this background. It’s currently not enforced, and it’s even more visible with this PR:
image
The add buttons also need to be adjusted:
image
Apart from that, I’m OK with improving the admin accessibility, but we have to make it extra beautiful, because we’ll probably have a lot of negative feedback from clients because people are used to the existing green. Currently, this PR’s rendering doesn’t look that great from my point of view. So we definitely need some designer insights, as pointed out by Thibaud & Matt! I’ll ask our internal designer for some ideas.

tmsndrs:
I'll take a look at this when I get some time. It's definitely a problem we need to solve, I'm just not sure about the current solution as it makes the interface feel a bit drab to me.

Meteor0id:
About the vertical alignment of the header: simply remove the top padding on the <h1> and leave the bottom padding. The alignment was already there though .nice-padding on <header>, but was being disrubted by its child having a top padding of its own.

Send a PR to @marteki 
https://github.com/marteki/wagtail/pull/1

Meteor0id:
fixed one of the two issues @BertrandBordage mentioned.
https://github.com/marteki/wagtail/pull/2


What I have not yet done is improve the contrast of the "add child page" plus icon background.
It puzzels me, in the variables it inherits the same dark-teal as the background itself, yet it ends up being darker-dark-teal. The proper solution would likely be to modify https://github.com/Meteor0id/wagtail/blob/e4c2383bd6a86b88e8c532c6da2a72a580648c4b/wagtail/admin/static_src/wagtailadmin/scss/components/_listing.scss#L387
and https://github.com/Meteor0id/wagtail/blob/e4c2383bd6a86b88e8c532c6da2a72a580648c4b/wagtail/admin/static_src/wagtailadmin/scss/components/_listing.scss#L478
to be another 10% darker.
These color variables are quite a mess, this needs to be refactored soon. I can not belive using varibles in this manner is making it easier for anyone.

Meteor0id:
While going though the templates, where the classes are defined for these elements, it is remarkable how similar portions of the editing interface are coded differently between views.
For example:
When editing a page, the entire top portion of the page is part of the header. But when listing pages, the top portion of the page is considered the first row of a table, `<tr class="index">`, with only the breadcrumbs being part of the page header.
The templates are inconsitent, leading to unnessesary classes, and overcomplicated CSS and SCSS rules.

Because some CSS properties are inherited, the style of the same class in a different context can differ greatly. That is what happend here with the text color. It is a symptom of a much larger issue with the front-end of Wagtail. Its templates have become overcomplicated.

tomdyson:
In the test instance at https://wagtail-pr-4624-stage.us.aldryn.io/en/admin/ (log in with admin / changeme), the button outlines are missing. Here's the proposed version:

screenshot 2018-09-19 17 10 49

and here's the current version:

screenshot 2018-09-19 17 10 57

wagtail/admin/static_src/wagtailadmin/scss/core.scss
```diff
@@ -198,12 +198,14 @@ footer {
     @include clearfix();
     overflow: hidden;
     padding-top: 1.4em;
+    padding-bottom: 0.65em;
```
allcaps: What is de idea about changing this padding? Why change any paddings? This one trows the breadcrumb/header ratio of balance. Also the breadcrumb text positioning is off.

screen shot 2018-09-20 at 15 50 03

marteki:
@tomdyson @Meteor0id @allcaps Thank you all for your review and feedback!

This PR wasn't intended to be an all-encompassing change for the header, and was focused only on increasing the font size (which is a critical accessibility bug). I also agree with some of the review comments that the PR does not take into account the effect that changing the size to the larger compliant size has on the rest of the admin.

I love that the design team is reaching out to help give more guidance around the direction certain items should go. I also hear/see that there's a strong desire to have this included in the next release with a larger look at the implications around the admin, so I can take a deeper dive into this over the next two or three weeks to include fixes for all the cascading changes to the header on different pages (and not on just the intro dashboard). I'll treat it as part of my regular sprint work.

Since the core design folks seem to be looking closer at this now, let me know how we can best collaborate together! I'd rather get your guidance earlier and identify needed items sooner than just work solo in a vacuum making guesses. (I'm also on the Wagtail Slack, where there's now an #Accessibility channel that we could use.)

tomdyson:
@marteki we really appreciate your continued attention to this. It's a very important contribution and I'm sorry that our response to it has been a bit disorganised. I've booked @benenright for some design time on Monday morning, and I know that @allcaps is intending to contribute over the next few days. Let's use the #accessibility Slack channel, as you suggest, and aim to have something that we're all happy with by the end of next week.

benenright:
Hi all

@marteki I have had a look at this and I think a good solution is in the attached:

wagtail-teal-aa

I have selected a slightly darker, richer version of the main teal. This means we can also continue to use the same colour in the button borders and hovers. I also suggest we bring the 'More' button border colour inline with the others for consistency.

I also used a subtle darker shade in the breadcrumb header - enough to distinguish it but without feeling heavy.

For ref:
Main Teal: #007D7E
Breadcrumb header: #006C6E
Button borders: #308C8C

Cheers
Ben

tomdyson:
Thanks for this, @benenright. @marteki are you happy to update the PR with these changes? Perhaps @allcaps and @nmorduch may be able to help?

allcaps:
@benenright en @tomdyson Thanks you both for pushing this forward. I'm working on it and will push my code asap.

gasman:
The colour contrast changes have now been implemented and merged in #4788. If we're still going ahead with the changes to the font size, I expect it'll be easier to open that as a new PR rather than dealing with the merge conflicts, so I'll close this one now (but happy to reopen if people think that works better).

PR #4625, opened 2018-06-15
Add download_url to the images API

mwharrison:
This adds a `download_url` to the image API to match how the document API works.

### DIFF
--- a/docs/advanced_topics/api/v2/configuration.rst
+++ b/docs/advanced_topics/api/v2/configuration.rst
@@ -228,6 +228,7 @@ This would add the following to the JSON:
             "meta": {
                 "type": "wagtailimages.Image",
                 "detail_url": "http://www.example.com/api/v2/images/12/",
+                "download_url": "/media/images/a_test_image.jpg",
                 "tags": []
             },
             "title": "A test image",
@@ -240,6 +241,12 @@ This would add the following to the JSON:
             "height": 100
         }
     }
+
+
+Note: ``download_url`` is the original uploaded file path, whereas
+``feed_image_thumbnail['url']`` is the url of rendered imaged.
+When you are using another storage backend, such as S3, ``download_url`` will return
+a URL to the image if your media files are properly configured.
 
 Additional settings
 ===================
--- a/wagtail/images/api/v2/serializers.py
+++ b/wagtail/images/api/v2/serializers.py
@@ -1,5 +1,20 @@
+from rest_framework.fields import Field
 from wagtail.api.v2.serializers import BaseSerializer
 
 
+class ImageDownloadUrlField(Field):
+    """
+    Serializes the "download_url" field for images.
+
+    Example:
+    "download_url": "/media/images/a_test_image.jpg"
+    """
+    def get_attribute(self, instance):
+        return instance
+
+    def to_representation(self, image):
+        return image.file.url
+
+
 class ImageSerializer(BaseSerializer):
-    pass
+    download_url = ImageDownloadUrlField(read_only=True)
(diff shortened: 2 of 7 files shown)


docs/advanced_topics/api/v2/configuration.rst
```diff
@@ -241,6 +242,12 @@ This would add the following to the JSON:
         }
     }
 
+
+Note: ``download_url`` is the original uploaded file path, whereas
+``feed_image_thumbnail['url']`` is the url of rendered imaged.
+When you are using another storage backend, such as AWS, ``download_url`` will return
```
tomdyson: AWS -> S3

docs/advanced_topics/api/v2/configuration.rst
```diff
@@ -241,6 +242,12 @@ This would add the following to the JSON:
...
+URL to the image if your media files are properly configured.
```
tomdyson: URL -> a URL

wagtail/images/api/v2/serializers.py
```diff
@@ -1,5 +1,21 @@
+from rest_framework.fields import Field
 from wagtail.api.v2.serializers import BaseSerializer
 
+from wagtail.api.v2.utils import get_full_url
+
+class ImageDownloadUrlField(Field):
+    """
+    Serializes the "download_url" field for documents.
```
tomdyson: for documents -> for images?

wagtail/images/api/v2/serializers.py
```diff
@@ -1,5 +1,21 @@
...
+
+    Example:
+    "download_url": "http://api.example.com/documents/1/my_document.pdf"
```
tomdyson: should this be a path to an image, e.g. `/media/images/a_test_image.jpg`?

kaedroho:
Thanks! But I'm pretty sure that we advise against downloading the original image directly like this. This is why the "original" filter exists. I can't remember what the rationale behind that position is. @gasman might remember.

gasman:
Yep - the original reasoning was that we might not want the high-resolution original image to be public, and so by configuring the web server to serve `/media/images/` but not `/media/original_images/`, you'd only be exposing the renditions that are actually used on the site.

However, we've never made this explicit in the documentation, or enforced it rigorously (for example, visiting the /admin/images edit view will create an 'original' rendition available to anyone at a guessable URL) so I'm inclined to relax that decision.

mwharrison:
This is something we found a real pain point when trying to build our FE. If we wanted that rendition URL, it would take three HTTP requests (one to generate - if it wasn't already, one to fetch the page, and one to the image API to get the link). We ended up building something to just serve up the original image URL in a hacky way. If you don't want to merge this one - I purpose at least some documentation on how it works, which is the path I was heading down before I decided to just "fix" it.

kaedroho:
@mwharrison Nothing against adding ``download_url`` to the images API. But for your particular problem, have you seen ``ImageRenditionField``? This will render the image and return its URL in the page API (no need for separate calls to the images API).

http://docs.wagtail.io/en/v2.1/advanced_topics/api/v2/configuration.html?highlight=imagerenditionfield#images-in-the-api

I'm actually thinking of changing the documentation to discourage people from enabling the images/documents APIs in favour of nesting everything that's needed in the pages API.

mwharrison:
@kaedroho - I had not seen that! Looks like it was something that was introduced after we built the site. I'll have to start implementing this.
A documentation change like that would go a long way.

gasman:
Merged in 8ef0edb371090bb2b26df915edc3dfe1b9bdc1cc - thanks @mwharrison!

PR #4626, opened 2018-06-16
Convert UUID primary key to string before serializing in json.

### LINKED ISSUE #4616: Chooser fails for custom UUID primary keys in snippets
This serialization here assumes that the primary key is readily serializable:

https://github.com/wagtail/wagtail/blob/4a7e4550044ce11cb68141b23ef8ca3884e1a07d/wagtail/snippets/views/chooser.py#L77-L82

However, when the primary key is a UUIDField you get a TypeError when a snippet is selected through a snippet chooser:

    TypeError: Object of type 'UUID' is not JSON serializable

So the ìtem.pk` must be converted explicitly to a string before trying to dump it with json.

saevarom:
Fixes #4616 .

### DIFF
--- a/wagtail/tests/settings.py
+++ b/wagtail/tests/settings.py
@@ -22,6 +22,9 @@ DATABASES = {
     }
 }
 
+if DATABASES['default']['ENGINE'] == 'sql_server.pyodbc':
+    DATABASES['default']['NAME'] = 'wagtail.db'
+
 # Add extra options when mssql is used (on for example appveyor)
 if DATABASES['default']['ENGINE'] == 'sql_server.pyodbc':
     DATABASES['default']['OPTIONS'] = {
--- a/wagtail/tests/settings.py
+++ b/wagtail/tests/settings.py
@@ -22,7 +22,7 @@ DATABASES = {
     }
 }
 
-if DATABASES['default']['ENGINE'] == 'sql_server.pyodbc':
+if DATABASES['default']['ENGINE'] == 'django.db.backends.sqlite3':
     DATABASES['default']['NAME'] = 'wagtail.db'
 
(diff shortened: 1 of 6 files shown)


chosak:
@saevarom as far as I know SQLite should have no problem with UUIDField. Running the tests on this branch locally all pass, and the default test configuration uses SQLite. The fact that the test command exits on Travis with `exited with code -11` suggests a segfault in SQLite - [this thread](https://github.com/travis-ci/travis-ci/issues/7873) suggests that others have experienced this on Travis as well. This wouldn't explain why CircleCI failed as well but might be a clue at least.

saevarom:
@chosak I tried running again with the file-based sqlite instead of the in-memory one as suggested in the thread you linked to. Now the CircleCI build works, but not Travis...

wagtail/tests/settings.py
```diff
@@ -22,6 +22,9 @@
     }
 }
 
+if DATABASES['default']['ENGINE'] == 'sql_server.pyodbc':
```
gasman: @saevarom This should be `django.db.backends.sqlite3` I think...?

saevarom: Woops, your right. Paste error :)

saevarom: I'll try again

gasman:
I think I've found the incantation to make Travis work: https://github.com/gasman/wagtail/commit/f4752d68874facbc16919b8a25830b71d287603d

Will open a rebased PR to make sure it runs on CircleCI too, and merge if all goes well.

gasman:
Merged in 6e60469142a11ac6df041e75c3537dcc538039ae - thanks @saevarom!

PR #4627, opened 2018-06-16
Expand the documentation on the EntityFeature  method

kutenai:
This is just an update to the documentation.
I found that the css keyword argument was not documented, and it differs from the js argument. This PR could be a little better if I know what the possible values for 'medium' are, but I at least documented that the css= argument needs to be a dict, and that 'all' is one value that works.


kutenai:
I just updated the documentation to include a link to the django docs here:
https://docs.djangoproject.com/en/2.0/topics/forms/media/

These docs would be helpful to understand the values for those arguments.

PR #4631, opened 2018-06-16
Add trove classfiers for Wagtail... to Wagtail.

FlipperPA:
Thanks to help from our friend Dustin Ingram, we've gotten Trove Classifiers for Wagtail added to PyPI. This adds them to Wagtail itself appears alongside the others!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -89,6 +89,8 @@ setup(
         'Framework :: Django',
         'Framework :: Django :: 1.11',
         'Framework :: Django :: 2.0',
+        'Framework :: Wagtail',
+        'Framework :: Wagtail :: 2',
         'Topic :: Internet :: WWW/HTTP :: Site Management',
     ],
     install_requires=install_requires,
--- a/setup.py
+++ b/setup.py
@@ -90,6 +90,7 @@ setup(
         'Framework :: Django :: 1.11',
         'Framework :: Django :: 2.0',
         'Framework :: Wagtail',
+        'Framework :: Wagtail :: 1',
         'Framework :: Wagtail :: 2',
         'Topic :: Internet :: WWW/HTTP :: Site Management',
     ],


FlipperPA:
Hmmm, Travis failed on the Python 3.4 download. `Unable to download 3.4 archive. The archive may not exist. Please consider a different version.`

FlipperPA:
Hmmm, Travis appears to be failing on the Python 3.4 download. Any ideas? `Unable to download 3.4 archive. The archive may not exist. Please consider a different version.`

setup.py
```diff
@@ -89,6 +89,9 @@
         'Framework :: Django',
         'Framework :: Django :: 1.11',
         'Framework :: Django :: 2.0',
+        'Framework :: Wagtail',
+        'Framework :: Wagtail :: 1',
+        'Framework :: Wagtail :: 2',
```
loicteixeira: I'm not sure about listing both `Framework :: Wagtail :: 1` and `Framework :: Wagtail :: 2` as one of the two will be wrong for every release going forward (i.e. release for `2.1.x` or `2.2` shouldn't list `Framework :: Wagtail :: 1` and LTS releases for `1.12` and `1.13` shouldn't list `Framework :: Wagtail :: 2`).

FlipperPA: That was my initial thought as well - you'll see I added `Wagtail :: 1` in a second commit. It was pointed out to me that if people are looking through PyPI, we will want Wagtail listed within the `Wagtail 1` sub-category, so they can wayfind to the latest version. What do you think?

loicteixeira: To me classifiers were supposed to be used by libraries to declare compatibility with a given Wagtail version, not as a mechanism to find Wagtail in searches.

The [classifier's documentation](https://pypi.org/classifiers/) is a bit ambiguous though:

> Each project's maintainers provide PyPI with a list of "trove classifiers" to categorize each release, describing who it's for, what systems it can run on, and how mature it is.
>
> These standardized classifiers can then be used by community members to find projects based on their desired criteria.

The first sentence kinda describe the project requirements in a way when the second sentence indeed talk about search.

FWIW, [Django does not declare its own classifiers](https://github.com/django/django/blob/5b733171813f8ddc7af84abe79f2646204b9c6ca/setup.py#L92-L109).

So in the end I don't know what to think.

thibaudcolas: @loicteixeira Django does declare its own classifier: https://github.com/django/django/blob/5b733171813f8ddc7af84abe79f2646204b9c6ca/setup.py#L95

I think it makes sense to follow the same approach without the version number – Wagtail will be findable, but won't try to convey any kind of compatibility with any specific version of itself.

loicteixeira: I don't know how I missed this one sorry :(

gasman:
Merged in 3cd18f3c1f69bb739a7f684f6bd1c62e24be786f (keeping just the non-version-specific classifier as per Django's approach). Thanks @FlipperPA!

Issue #4639, opened 2018-06-20
Update a file with the same name


csalom:
### Issue Summary
At the documents edit view, if the file has changed the original file is deleted. 
If you edit a document uploading a new version of the file with the same name, the new document will be upload to the file system and the name of the document object will not be changed.
Then, the file is deleted which the reference of the new file.

### Steps to Reproduce
1. Upload a document
2. Enter to its edit view /documents/edit/pk
3. Update the document with a new version of the previous one, with the same name.
4. Enter to the edit view of the document again /documents/edit/pk -> A 404 Error was raised when accessing to the file because of its deletion.

### Technical details

* Python version: Run `3.6`.
* Django version:  `2.0 and previous`.
* Wagtail version: `2.1 and previous`.

csalom:
I've opened a PR with a possible solution: #4640

gasman:
Covered by #4567

csalom:
Hi @gasman 
You're right, thank you!

richaagarwal:
Does it make sense to close this issue if the PRs in question have not been approved or merged?

gasman:
Hi @richaagarwal,
Unlike some other projects, we don't require PRs to have a corresponding issue open - the PR itself can serve as a problem report, that just happens to have a proposed solution attached. In this case, the existing PR #4567 had a more complete description of the problem and some further discussion, so it would be counter-productive to have another issue open covering the same ground.

PR #4640, opened 2018-06-20
documents edit view - only remove file if the name has changed

csalom:
This PR is a possible solution to issue #4639


gasman:
Hi @csalom - thanks for the PR!

This is already covered by #4567 - please see the discussion there.

Issue #4496, opened 2018-04-26
wagtailredirects.middleware causing problems with illegal strings?


benjaoming:
### Issue Summary

A malicious crawl caused this in `wagtailredirects.middleware`:

`A string literal cannot contain NUL (0x00) characters.`

Not sure if this should be the responsibility of `db.backends.utils` or `wagtailredirects`.

### Steps to Reproduce

Add a weird character in the path.

### Technical details

* Python version: 3.5.2.
* Django version: 1.11.12.
* Wagtail version: 1.13.1.
* Browser version: n/a

```

Traceback:  

File ".virtualenvs/venv/lib/python3.5/site-packages/django/core/handlers/base.py" in get_response
  131.                 response = middleware_method(request, response)

File ".virtualenvs/venv/lib/python3.5/site-packages/wagtail/wagtailredirects/middleware.py" in process_response
  51.         redirect = get_redirect(request, path)

File ".virtualenvs/venv/lib/python3.5/site-packages/wagtail/wagtailredirects/middleware.py" in get_redirect
  31.         redirect = _get_redirect(request, uri_to_iri(path))

File ".virtualenvs/venv/lib/python3.5/site-packages/wagtail/wagtailredirects/middleware.py" in _get_redirect
  19.         return models.Redirect.get_for_site(request.site).get(old_path=path)

File ".virtualenvs/venv/lib/python3.5/site-packages/django/db/models/query.py" in get
  374.         num = len(clone)

File ".virtualenvs/venv/lib/python3.5/site-packages/django/db/models/query.py" in __len__
  232.         self._fetch_all()

File ".virtualenvs/venv/lib/python3.5/site-packages/django/db/models/query.py" in _fetch_all
  1118.             self._result_cache = list(self._iterable_class(self))

File ".virtualenvs/venv/lib/python3.5/site-packages/django/db/models/query.py" in __iter__
  53.         results = compiler.execute_sql(chunked_fetch=self.chunked_fetch)

File ".virtualenvs/venv/lib/python3.5/site-packages/django/db/models/sql/compiler.py" in execute_sql
  900.             raise original_exception

File ".virtualenvs/venv/lib/python3.5/site-packages/django/db/models/sql/compiler.py" in execute_sql
  890.             cursor.execute(sql, params)

File ".virtualenvs/venv/lib/python3.5/site-packages/django/db/backends/utils.py" in execute
  64.                 return self.cursor.execute(sql, params)

[...]

PR #4652, opened 2018-06-26
Strip NULL characters when normalizing paths

acrewdson:
(see related issue https://github.com/wagtail/wagtail/issues/4496)

After migrating a Wagtail-based site from MySQL to Postgres, we noticed that malicious requests to the site that included percent-encoded Unicode NULLs (`%00`) raised a `ValueError` exception that we hadn't seen when using MySQL: `A string literal cannot contain NUL (0x00) characters.` This may relate to `psycopg2`'s decision to raise an exception in these situations, as discussed here: https://github.com/psycopg/psycopg2/issues/420

While newer versions of Django [appear to provide some field validation](https://docs.djangoproject.com/en/2.0/ref/validators/#prohibitnullcharactersvalidator) that addresses these null chars, it doesn't look like Wagtail's redirect middleware is making use of those validators, and so it seemed reasonable to clean these chars in the context of _normalizing_ the paths before looking in the db for corresponding redirects – especially since a quick investigation on the internet suggests that `U+0000` in URLs can be used as a means of attack, and also since RFC 3986 says:

> 
>    Note, however, that the "%00" percent-encoding (NUL) may require
>    special handling and should be rejected if the application is not
>    expecting to receive raw data within a component.
> 

Thanks for considering this patch.

cc: @chosak

### DIFF
--- a/wagtail/contrib/redirects/models.py
+++ b/wagtail/contrib/redirects/models.py
@@ -71,8 +71,12 @@ class Redirect(models.Model):
         parameters_components = parameters.split(';')
         parameters = ';'.join(sorted(parameters_components))
 
-        # Query string components must be sorted alphabetically
         query_string = url_parsed[4]
+
+        # Strip Unicode NULLs (U+0000) for safety and Postgres compatibility
+        query_string = query_string.replace('%00', '')
+
+        # Query string components must be sorted alphabetically
         query_string_components = query_string.split('&')
         query_string = '&'.join(sorted(query_string_components))
 
--- a/wagtail/contrib/redirects/tests.py
+++ b/wagtail/contrib/redirects/tests.py
@@ -78,6 +78,8 @@ class TestRedirects(TestCase):
 
         self.assertEqual('/', normalise_path('/'))  # '/' should stay '/'
 
+        self.assertEqual('/foo?bar=baz', normalise_path('/foo?bar=baz%00'))  # Strip NULLs
+
         # Normalise some rubbish to make sure it doesn't crash
         normalise_path('This is not a URL')
         normalise_path('//////hello/world')


benjaoming:
Seems like this is what is needed -- but should the issue be elevated to psycopg or Django's db backend? Postgres doesn't allow `\0` in general.

chosak:
@benjaoming per [the psycopg2 issue](https://github.com/psycopg/psycopg2/issues/420) mentioned by @acrewdson above, the psycopg2 code was modified [here](https://github.com/psycopg/psycopg2/pull/459) in the 2.7 release to deliberately raise an exception when a null byte is sent.

See [this Django ticket](https://code.djangoproject.com/ticket/28201) and [this django-developers thread](https://groups.google.com/forum/#!topic/django-developers/D1gvXYCezEc/discussion) for discussion about how to handle this in Django; [consensus solution](https://github.com/django/django/pull/8670) was creation of the [`ProhibitNullCharactersValidator`](https://docs.djangoproject.com/en/2.0/ref/validators/#prohibitnullcharactersvalidator) and its addition to `CharField` and `TextField`.

For our purposes here in Wagtail, we can't leverage that validator because we're not using a form when we [test the request path against redirects](https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/redirects/middleware.py#L41).

chosak:
Merged in 882f8f3; thanks @acrewdson.

gasman:
Revisiting this fix to backport it to earlier Wagtail versions, and I'm not sure if this will catch all versions of this attack:

* are there other ways of inserting null bytes other than `%00` (e.g. sending an actual 0 byte as part of the HTTP request)?
* can the null byte be passed somewhere other than the querystring?

It looks like a more robust fix would be to add

    if '\0' in path:
        return None

to the top of `_get_redirect` in middleware.py. However, I can't reproduce the original issue to test this locally. Has anyone been able to reproduce this on `./manage.py runserver`, or does it only occur when running on a 'real' webserver?

acrewdson:
@gasman I was able to repro it locally with Django's `runserver`, using the type of request that we saw, and that others reported, involving the null character being percent-encoded in the querystring. It's a good point that the character could be inserted in other contexts, not just the querystring, so it might be worth revisiting this as you suggest.

chosak:
I can reliably produce the `A string literal cannot contain NUL (0x00) characters` error using runserver following these steps:

1. `wagtail start mysite && cd mysite`
2. Edit `mysite/settings/base.py` to change default database to one using`django.db.backends.postgresql_psycopg2`. (You'll also need to `pip install psycopg2` into your environment, if you don't have it already.)
3. `./manage.py migrate`
4. `./manage.py runserver` 
5. Either of these will produce the 500:
    ```sh
    $ curl -I 'http://localhost:8000/foo%00'
    $ curl -I 'http://localhost:8000/foo/?bar=%00'
    ```

This generates 500s with all of these combinations:

- Python 2.7/Django 1.11/Wagtail 1.13
- Python 3.6/Django 1.11/Wagtail 1.13
- Python 3.6/Django 1.11/Wagtail 2.0

but not with this:

- Python 3.6/Django 2.0/Wagtail 2.0

So perhaps there's some fix in Django 2 that helps this somehow? Looking at the changelog, nothing obvious jumps out except maybe [this](https://docs.djangoproject.com/en/2.0/releases/2.0/#removed-support-for-bytestrings-in-some-places).

@gasman your idea of a more robust checking of the entire path seems like a good one.

Issue #3299, opened 2017-01-25
Make arguments of PageChooserBlock more similar to PageChooserPanel


m1kola:
In Wagtail 1.8 we introduced the the `target_model` argument in `PageChooserBlock` which takes a page model class or string representation of model like `tests.EventPage` (see #3072). It's a good idea, but some people (including me :)) were confused by interface `PageChooserBlock` provides. The `PageChooserBlock` block is very similar to the `PageChooserPanel` panel, but there are some differences in interfaces:

* `PageChooserBlock` provides the `target_model` argument to filter models in a chooser. In `PageChooserPanel` similar argument have a different name (`page_type`). Moreover we usually use the "page type" term in Wagtail (see [`Page.subpage_types`](http://docs.wagtail.io/en/v1.8/reference/pages/model_reference.html#wagtail.wagtailcore.models.Page.subpage_types) and [`Page.parent_page_types`](http://docs.wagtail.io/en/v1.8/reference/pages/model_reference.html#wagtail.wagtailcore.models.Page.parent_page_types), for example.
* `PageChooserBlock`'s `target_model` argument accepts only one page type. `PageChooserPanel` allows to pass a list or tuple of page types.

TODO:

- [x] Allow `PageChooserBlock` to take a list of page types in the `target_model` argument (PR #3297)
- [ ] Rename the `target_model` argument to `page_type` with a deprecation period

gasman:
Completed in #4653

PR #4653, opened 2018-06-27
Deprecate PageChooserBlock's `target_model` parameter in favour of `page_type`

loicteixeira:
Completes #3299.

- [X] ~~Deprecate `target_model` in favour of `page_type`~~
Allow `page_type` in addition to `target_model` (as we [need to keep](https://github.com/wagtail/wagtail/pull/4653#discussion_r198867963) `target_model` around).
- [x] Ensure it plays nice with existing migrations.
- [X] Update tests
- [x] Update documentation accordingly.

### DIFF
--- a/wagtail/core/blocks/field_block.py
+++ b/wagtail/core/blocks/field_block.py
@@ -1,5 +1,4 @@
 import datetime
-import warnings
 
 from django import forms
 from django.db.models.fields import BLANK_CHOICE_DASH
@@ -13,7 +12,6 @@ from django.utils.safestring import mark_safe
 
 from wagtail.core.rich_text import RichText
 from wagtail.core.utils import resolve_model_string
-from wagtail.utils.deprecation import RemovedInWagtail24Warning
 
 from .base import Block
 
@@ -597,12 +595,10 @@ class ChooserBlock(FieldBlock):
 
 class PageChooserBlock(ChooserBlock):
     def __init__(self, page_type=None, can_choose_root=False, target_model=None, **kwargs):
+        # We cannot simply deprecate 'target_model' in favour of 'page_type'
+        # as it would force developers to update their old migrations.
+        # Mapping the old 'target_model' to the new 'page_type' kwarg instead.
         if target_model:
-            warnings.warn(
-                "PageChooserBlock's `target_model` parameter is deprecated. "
-                "Please use `page_type` instead.",
-                category=RemovedInWagtail24Warning
-            )
             page_type = target_model
 
         if page_type:
--- a/wagtail/core/blocks/field_block.py
+++ b/wagtail/core/blocks/field_block.py
@@ -651,7 +651,7 @@ class PageChooserBlock(ChooserBlock):
     def deconstruct(self):
         name, args, kwargs = super().deconstruct()
 
-        if 'page_type' in kwargs:
+        if 'target_model' in kwargs or 'page_type' in kwargs:
             target_models = []
 
             for target_model in self.target_models:
@@ -660,6 +660,7 @@ class PageChooserBlock(ChooserBlock):
                     '{}.{}'.format(opts.app_label, opts.object_name)
                 )
 
+            kwargs.pop('target_model', None)
             kwargs['page_type'] = target_models
 
         return name, args, kwargs
--- a/wagtail/core/blocks/field_block.py
+++ b/wagtail/core/blocks/field_block.py
@@ -1,4 +1,5 @@
 import datetime
+import warnings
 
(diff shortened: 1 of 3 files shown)


wagtail/core/blocks/field_block.py
```diff
@@ -594,18 +596,23 @@ class Meta:
...
 class PageChooserBlock(ChooserBlock):
-
-    # TODO: rename target_model to page_type
-    def __init__(self, target_model=None, can_choose_root=False,
-                 **kwargs):
+    def __init__(self, page_type=None, can_choose_root=False, target_model=None, **kwargs):
         if target_model:
+            warnings.warn(
+                "PageChooserBlock's `target_model` parameter is deprecated. "
+                "Please use `page_type` instead.",
```
loicteixeira: I could not find an example of deprecation warning for renaming a method's parameter so I don't know whether this is consistent with old warnings.

wagtail/core/blocks/field_block.py
```diff
@@ -657,7 +664,7 @@ def deconstruct(self):
                     '{}.{}'.format(opts.app_label, opts.object_name)
                 )
 
-            kwargs['target_model'] = target_models
+            kwargs['page_type'] = target_models
```
loicteixeira: The fact that the `deconstruct` method is updated (and `__init__` actually) means that old migrations will need to be updated. Is it only a matter of asking the developers to update their migrations accordingly or is there a less involved way of dealing with that?

wagtail/core/blocks/field_block.py
```diff
@@ -594,18 +596,23 @@ class Meta:
 
 
 class PageChooserBlock(ChooserBlock):
-
-    # TODO: rename target_model to page_type
-    def __init__(self, target_model=None, can_choose_root=False,
-                 **kwargs):
+    def __init__(self, page_type=None, can_choose_root=False, target_model=None, **kwargs):
```
loicteixeira: Setting `page_type` as first keyword argument and moving `target_model` as last one so if the `PageChooserBlock` was created with positional argument, this should just work ™

wagtail/core/blocks/field_block.py
gasman: Requiring developers to update old migrations is something we should avoid, I think. That means we probably need the `__init__` method to accept `target_model` indefinitely, and leave out the deprecation warning (unless there's a way to suppress that warning during migrations).

(Updating `deconstruct` to use `page_type` is still the right thing to do, so that new migrations use the new name. I'm pretty sure that the migration logic will treat them as equivalent in this case - e.g. `makemigrations` won't try to create a migration if `target_model` -> `page_type` is the only change - but we should double-check that.)

loicteixeira: That's unfortunate 😞 

I'll have a look at whether it's possible to ignore the warning within the context of migrations, but I imagine it will be rather hacky and costly.

I'll also confirm whether it creates new migrations and update accordingly. Thanks for the feedback.

loicteixeira: > unless there's a way to suppress that warning during migrations

It seems that we could check if `migrate` or `showmigrations` are present in `sys.argv` but I don't think it's worth it because it will be fragile (e.g. a developper subclass one of those command and therefore names it differently) and probably not worth in the end.

loicteixeira: 9f90aab2906671413df5c09aba9ca2d5a0ef87dd tweaks the `deconstruct` method so it does not create unnecessary migrations.

I tested a combinaison of positional/keyword arguments, string/list arguments, with/without making changes after upgrade:

```diff python
  PageChooserBlock('blog.BlogPage')

  PageChooserBlock(['blog.BlogPage', 'blog.BlogIndexPage'])

  PageChooserBlock(target_model='blog.Page')  

- PageChooserBlock(target_model='blog.Page')
+ PageChooserBlock(page_type='blog.Page')

  PageChooserBlock(target_model=['blog.BlogPage', 'blog.BlogIndexPage'])

- PageChooserBlock(target_model=['blog.BlogPage', 'blog.BlogIndexPage'])
+ PageChooserBlock(page_type=['blog.BlogPage', 'blog.BlogIndexPage'])
```

---

As an aside (although outside of the scope of this PR and probably not worth changing anyway), when supplying a string as keyword argument (either `target_model` or now `page_type`), it will always be converted to a list when deconstructing while supplying it as positional arguments won't have it changed.

gasman:
Looks good to me! Merged in 53cea8bc04d56437a422b4e2cd79bd060e2b9ec9.

PR #4669, opened 2018-07-04
Bug/revisions unschedule for unpublished pages

### LINKED ISSUE #4657: Unable to remove scheduled revisions for non-live pages.
### Issue Summary

When a page is added and the first publication is scheduled for a date in the future, an admin user is unable to cancel the scheduled revision due to a 403 Forbidden error.

### Steps to Reproduce

With a working Wagtail installation:

1. Create a new page
2. Set a future Go live date
3. Publish
4. Go to revisions screen of new page
5. Click Cancel Scheduled Revision

The [view](https://github.com/wagtail/wagtail/blob/b4a2e13cb02c8c11368b3639b887066d8a3d306f/wagtail/admin/views/pages.py#L1157) that cancels scheduled revisions calls the [can_unpublish()](https://github.com/wagtail/wagtail/blob/b4a2e13cb02c8c11368b3639b887066d8a3d306f/wagtail/core/models.py#L1741) method on the `PagePermissionTester`. The `can_unpublish()` method returns `False` if the page is live, which it isn't as it hasn't been published yet.

### Technical details

* Python version: 3.6.4
* Django version: 2.0.4
* Wagtail version: 2.0.1
* Browser version: Chrome 67

malikwahab:
This PR fixes #4657 

Revisions unscheduling uses `can_unpublish` to determine if a user has permission to unschedule a revision, `can_unpublish` however checks if a page is live which causes unscheduling to fail if the page has not been published yet. This fix changes the permission check to use `can_publish` which also checks for all the necessary permissions too.

If there is a better approach to fixing this, I'm open ideas.

PR Checks

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/pages/revisions/list.html
+++ b/wagtail/admin/templates/wagtailadmin/pages/revisions/list.html
@@ -11,6 +11,7 @@
     </thead>
     <tbody>
         {% if revisions %}
+            {% page_permissions page as page_perms %}
             {% for revision in revisions %}
                 <tr {% if revision == page.get_latest_revision %}class="index"{% endif %}>
                     <td class="title">
@@ -36,7 +37,7 @@
                                 <li><a href="{% url 'wagtailadmin_pages:revisions_compare' page.id previous_revision.id revision.id %}" class="button button-small button-secondary">{% trans 'Compare with previous revision' %}</a></li>
                             {% endif %}
                             {% endwith %}
-			    {% if revision.approved_go_live_at %}
+			    {% if revision.approved_go_live_at and page_perms.can_unschedule %}
 				<li><a href="{% url 'wagtailadmin_pages:revisions_unschedule' page.id revision.id %}" class="button button-small button-secondary">{% trans 'Cancel scheduled publish' %}</a></li>
 			    {% endif %}
                         </ul>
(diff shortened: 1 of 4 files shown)


tomdyson:
Thanks @malikwahab!

gasman:
Thanks @malikwahab! Making the permission correspond to 'can publish' seems sensible to me.

Could I suggest that we add a `can_unschedule` method to `PagePermissionTester` (which can just return `self.can_publish()`, like `can_set_view_restrictions` does), and use that in the view rather than calling `can_publish` directly? That way, this business rule is kept in the same place as all the others, and it'll be easier to revise in future if necessary.

Also, I think it would be good to add this permission check to `wagtail/admin/templates/wagtailadmin/pages/revisions/list.html`, to show/hide the button where appropriate. There's a `page_permissions` tag you can use to retrieve the permission tester object: https://github.com/wagtail/wagtail/blob/c22321ee759b80f10e18b4d04fb02530165e5236/wagtail/admin/templates/wagtailadmin/pages/listing/_list.html#L48

malikwahab:
Thanks @gasman for the review. The `can_unschedule` fits better than `can_published` in terms of code readability and I forgot about hiding the button. I'll work on it.

gasman:
Thanks again @malikwahab! Merged in e53076d949f1ac15d11bfbfb3b98866efa04ce32 + parents.

PR #4671, opened 2018-07-04
Add action buttons to editing view header

### LINKED ISSUE #4599: Feature Request: Link back to page tree (to manipulate children) from page edit
While editing a page, there is not straightforward (that I've found) way to navigate the web browser to the tree navigation mode of the same page for the purpose of viewing / manipulating the page's children. One may click the breadcrumb to navigate a level up, but then must find the page they were on and select the "explore child pages" link in order to manipulate the page's children.

It would be helpful for users of my organization to be able to click a link to return to the page tree of the current page. I propose that while editing a page, the page being edited be added to the breadcrumb such that the user may click it and return to the page tree. The link of course would just remove "edit/" from the URL.

This is a mockup of proposal. In this implementation, the current page's title is a link in the breadcrumb with a lower right pencil to indicate the page is being edited:
image

nmorduch:
Fixes #4599

### Add action buttons to editing view header

- Add `editing` argument to page listing buttons template tag and hook, to indicate whether in editing view
- Action buttons in editing view have first button `Show in Explorer` instead of `Edit`. This had originally been `View in Explorer`, but `Show in Explorer` is consistent with the userbar.
- Update tests and documentation to use additional argument
- Increase dropdown menu z-index so that it is not behind editing view headings

### Other small changes to the header

- Increase font size in page title of editing view to more closely match styles in explorer view
- Move page type in editing view from page title to list of status indicators

### Potential other changes

This isn't a very elegant way of adding these buttons, but I didn't want to get into too much refactoring because I wanted to start with this being as small a change as possible. Here are a few notes on potential changes I could also make if that would be preferred:

- I noticed that the `is_parent` argument for page listing buttons is really used to determine whether to use the styles that make sense on the teal heading. This is no longer used only for parent pages, so it could make sense to rename it.
- I added the `editing` argument to `page_listing_buttons` but not to `page_listing_more_buttons`. It seems reasonable that people might want to have custom buttons for the `more` dropdown that vary based on whether one is editing, so I could add that argument to the `more` buttons as well.
- My additional CSS is not terribly well organized, but neither is the CSS it's amending. I would personally be inclined to address that CSS in a separate pull request.

### Other things I tested

- I tested the front end only in Firefox 61 for both wide and narrow screens. Testing with more browsers would not be unwise, though none of the changes use poorly-supported CSS.
- Clicking the new buttons does indeed yield a warning about navigating off of the page.

### DIFF
--- a/wagtail/admin/wagtail_hooks.py
+++ b/wagtail/admin/wagtail_hooks.py
@@ -98,14 +98,22 @@ def register_collections_menu_item():
 
 
 @hooks.register('register_page_listing_buttons')
-def page_listing_buttons(page, page_perms, is_parent=False):
+def page_listing_buttons(page, page_perms, is_parent=False, editing=False):
     if page_perms.can_edit():
-        yield PageListingButton(
-            _('Edit'),
-            reverse('wagtailadmin_pages:edit', args=[page.id]),
-            attrs={'aria-label': _("Edit '{title}'").format(title=page.get_admin_display_title())},
-            priority=10
-        )
+        if editing:
+            yield PageListingButton(
+                _('Show in Explorer'),
+                reverse('wagtailadmin_explore', args=[page.id]),
+                attrs={'aria-label': _("Show '{title}' in the Explorer").format(title=page.get_admin_display_title())},
+                priority=10
+            )
+        else:
+            yield PageListingButton(
+                _('Edit'),
+                reverse('wagtailadmin_pages:edit', args=[page.id]),
+                attrs={'aria-label': _("Edit '{title}'").format(title=page.get_admin_display_title())},
+                priority=10
+            )
     if page.has_unpublished_changes:
         yield PageListingButton(
             _('View draft'),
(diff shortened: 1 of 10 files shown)


wagtail/admin/wagtail_hooks.py
```diff
@@ -98,14 +98,22 @@ def register_collections_menu_item():
 
 
 @hooks.register('register_page_listing_buttons')
-def page_listing_buttons(page, page_perms, is_parent=False):
+def page_listing_buttons(page, page_perms, is_parent=False, editing=False):
```
thibaudcolas: Do we need to update this hook’s documentation based on this?

thibaudcolas:
Oh – ideally we would also update the screenshots in the docs. @nmorduch I’m happy to take care of this as part of merging the PR. Here are the pages I found where screenshots would need to be updated:

- http://docs.wagtail.io/en/v2.5.1/editor_manual/new_pages/selecting_a_page_type.html
- http://docs.wagtail.io/en/v2.5.1/editor_manual/new_pages/required_fields.html
- http://docs.wagtail.io/en/v2.5.1/editor_manual/editing_existing_pages.html

thibaudcolas:
Superseded by #7783

PR #4673, opened 2018-07-06
Wrap deleting page into database transaction

tm-kn:
Currently queries executed in the hooks don't run in the transaction with the page deletion query and it's harder to write hook without copying the whole view if you want to keep queries running in the hooks integral with page deletion.


giohappy:
transaction comes from django.db not django.db.models. That's the reason of the failed test

gasman:
Thanks @tm-kn! Merged in 361def81c2f9c524b4c8057841e50924d2ae7b83.

PR #4678, opened 2018-07-09
replace my name with my name

Pomax:
I do not write code under my passport name, if I have to be in a changelog, please use "Pomax" =)

(this is not merely an internet handle, this has been my name for two decades now)


gasman:
Merged in 40981b0c6550720933f5cd9e44c681ce0677fad0 (which also updates your name in contributors.txt) 👍

PR #4682, opened 2018-07-11
Reapply #4593 (Update search query classes for RFC 25)

gasman:
Rebase of #4593 on master (previously reverted in #4680 due to test breakage).

The following commits from the original PR have been omitted:

2faa90ec942c6b5f31cf1032847f2a41e8801b3c - broke several tests on Elasticsearch, evidently because the `get_children` method was still in use
3b26b3fc75e1436ffa84ee9a21b040f59749fd8c - due to dropping the above commit, `MultiOperandsSearchQueryOperator` isn't as empty any more, so may not make sense to eliminate


BertrandBordage:
Great to see this Matt, thanks for taking the time to do this.
Once this is merged, I’ll make a tiny commit to add support for the new `autocomplete()` method in PostgreSQL, following #4575.

kaedroho:
Thanks for fixing!

PR #4686, opened 2018-07-13
Update integrating_into_django.rst

Meteor0id:
flaw in url, was still referencing depreciated method, fixed.

### DIFF
--- a/docs/getting_started/integrating_into_django.rst
+++ b/docs/getting_started/integrating_into_django.rst
@@ -87,7 +87,7 @@ The URL paths here can be altered as necessary to fit your project's URL scheme.
 
 .. code-block:: python
 
-    url(r'', include(wagtail_urls)),
+    re_path(r'', include(wagtail_urls)),
 
 In this case, this should be placed at the end of the ``urlpatterns`` list, so that it does not override more specific URL patterns.
 
--- a/docs/getting_started/integrating_into_django.rst
+++ b/docs/getting_started/integrating_into_django.rst
@@ -77,6 +77,11 @@ Now make the following additions to your ``urls.py`` file:
         ...
     ]
 
+.. important::
+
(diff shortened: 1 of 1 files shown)


mikedingjan:
Wagtail supports Django 1.11 which doesn't have the ``re_path()`` the old ``url()`` function is kept in Django as alias https://docs.djangoproject.com/en/2.1/ref/urls/#url

Meteor0id:
I see. However on django 2.0.5 using wagtail 2.0.2 and phython 3.5.something url() is invalid. The Django documentation mentions support might get dropped for url().

The page http://docs.wagtail.io/en/latest/getting_started/integrating_into_django.html DOES make use of re_path in all places but one. @mikedingjan the documentation makes no sense, and seems to favor recent versions of Django in some examples, and favor old versions of Django in others.

I recommand using re_path in the examples, and adding a warning that older versions of Django use url().

url() is no longer supported in recent versions of Django, and will lead to an error when migrating.

mikedingjan:
Hi @Meteor0id ,

In the latest (2.1) version of Django ``url()`` is still supported, but raises a deprecation warning, that's something different than not being supported..

I agree with you on the part where documentation favours newer Django versions and should do so consistently. Adding a waring notice for Django 1.11 users (with a link to the Django docs) is imo the best solution here

Meteor0id:
Seems like a good solution to me. However I would like to point out that on my setup (vesion numbers have been mentioned earlier) migration did fail with an error mentioning that url() has not been imported.
Are you sure url() is still supported by Django? Are you sure it does not require a separate import?

Your proposed solution would solve it it either way, if all imports have been mentioned correctly.

Meteor0id:
I have added an important notice ot point this out, and made sure the document consitently makes use of re_path. Please check if my examples are correct, I have not tested this with Django 1.11 .

mikedingjan:
@Meteor0id thanks for the changes, looks good to me now!

The docs state that the ``url()`` is still available but will be deprecated in the future.. so when dropping Django 1.11 support we should remove the ``..important`` notice again.. :)

mikedingjan:
Your examples are correct, merged the PR btw.. thanks for contributing!

PR #4688, opened 2018-07-13
Fix autocomplete API for Elasticsearch 6

kaedroho:


BertrandBordage:
Arg, nice catch! Looks like a copy-paste issue ;)

PR #4689, opened 2018-07-13
Fix indexing related fields as FilterField in Elasticsearch

### LINKED ISSUE #4681: Facet tag search fails on Elasticsearch
`test_facet_tags` in the search backend tests fails (returning no results) on Elasticsearch. Have marked it as an expected failure for now (in 1d517dd0d8f3aef67403bacf8ff3724c747a3aca) but need to resolve this for the 2.2 release.

kaedroho:
Fixes #4681 

When looking in to #4681 I found that our support for indexing related fields by using ``FilterField`` directly on the field is a bit broken. All IDs are being coerced to strings by Elasticsearch (causing faceting on tags fields to break) and also Taggit tags fields were indexing the IDs of the through model rather than the IDs of the tags (causing faceting on tags fields to break some more).

This PR fixes these two issues. All related fields now find the actual underlying type of the foreign key and we I've added some extra logic to handle tags fields (note, other many to many fields were being indexed correctly already).

### DIFF
--- a/wagtail/search/backends/elasticsearch2.py
+++ b/wagtail/search/backends/elasticsearch2.py
@@ -263,6 +263,8 @@ class Elasticsearch2Mapping:
                     value = list(value.values_list('pk', flat=True))
                 elif isinstance(value, models.Model):
                     value = value.pk
+                elif isinstance(value, (list, tuple)):
+                    value = [item.pk if isinstance(item, models.Model) else item for item in value]
 
             doc[self.get_field_column_name(field)] = value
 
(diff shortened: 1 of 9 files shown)


BertrandBordage:
Thanks @kaedroho :)
Still one DB backend issue with the `refresh` method.

BertrandBordage:
Thanks for improving that PR, Karl!
Unfortunately, it’s still failing on test_facet_tags :cry:

kaedroho:
Damn, it works on ES6 now, but not the others...

BertrandBordage:
Do you think we could now remove ES tests from allowed failures on Travis?

kaedroho:
We switched them to allowed failure because of transient issues with ES. But yeah, I think we should try it as those issues might be fixed in recent versions.

PR #4692, opened 2018-07-13
url() is depriciated. Alias re_path()

Meteor0id:
Added important note about using url() on older versions of django, but switched the examples to re_path as to comply with Django docs for 2.0 and later.

### DIFF
--- a/docs/advanced_topics/settings.rst
+++ b/docs/advanced_topics/settings.rst
@@ -596,18 +596,23 @@ URL Patterns
   from wagtail.documents import urls as wagtaildocs_urls
 
   urlpatterns = [
-      url(r'^django-admin/', include(admin.site.urls)),
+      re_path(r'^django-admin/', include(admin.site.urls)),
 
-      url(r'^admin/', include(wagtailadmin_urls)),
-      url(r'^documents/', include(wagtaildocs_urls)),
+      re_path(r'^admin/', include(wagtailadmin_urls)),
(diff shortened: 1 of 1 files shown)


Meteor0id:
This is a continuation of https://github.com/wagtail/wagtail/pull/4686 to improve consitency and future accuracy of the docs.

mikedingjan:
Hi @Meteor0id ,

Thanks for continuing on this! However with Django 2.0 not all entries have to be ``re_path`` for example de default admin urls..

```
from django.contrib import admin
from django.urls import path

urlpatterns = [
    path('admin/', admin.site.urls),
]
```

mikedingjan:
Related PR: #4693

Meteor0id:
@mikedingjan Maybe I misunderstood the use of re_path, but the way I understand it, by using re_apth here isntead of path, all sub pages of admin/ will redirect to admin/ if the user is not logged in.

So wouldn't using re_path for all of these be the right way to do it?
Also notice taht before I edited this, the admin interface was already refered to with a regular expression. All I did was use the new method, using the same parameters.

mikedingjan:
I don't think it matters that much, will work either way..

Meteor0id:
@mikedingjan Than shall we merge it or should it be adjusted?
I don't see any issues.

mikedingjan:
I'll merge them like this, documentation which doesn't raise ``DeprecationWarnings`` is always better, thanks for your contribution! 👍

Issue #4223, opened 2018-01-29
Feature request: add API to customise front-end HTML rendering of rich text features


thibaudcolas:
### Issue Summary

New feature: add an additional "rich text features" API to give developers full control over how the front-end HTML is created, for custom and also built-in features.

At the moment, unless the feature is a link with a linktype or an embed (images and actual embeds), we can only define how it is stored as HTML in the DB. This means that the feature's "storage format" must be the same as its "front-end format". This is too limiting – for example, it should be possible to add a class attribute to the feature's front-end output without having it stored in the DB. It should be possible to store a `strikethrough` format as `<s>` tags but render them on the front-end as a more semantic `<del>`, should it make more sense for a given site.

Here is what the current APIs allows you to do:

```python
features.register_converter_rule(
        'contentstate', 'blockquote', {
            [...]
            'to_database_format': {
                'block_map': {'blockquote': {
                    'element': 'blockquote',
                    # This is in to_database_format, so the presentational attributes will be stored.
                    'props': {
                        'class': 'richtext__blockquote',
                    }
                }}
            }
        })
```

Ideally, there should be a separate `to_frontend_html` (or equivalent) to control the HTML on the front-end only. Note: it might not make sense to put this in `register_converter_rule`, since these are specific to a given converter (editorhtml for Hallo, contentstate for Draftail), whereas the new feature wouldn't be.

### Use cases

Here are the use cases we had when building [WagtailDraftail](https://github.com/springload/wagtaildraftail), beyond custom links and embeds:

```py
DRAFT_EXPORTER_BLOCK_MAP = dict(BLOCK_MAP, **{
    BLOCK_TYPES.UNORDERED_LIST_ITEM: {
        'element': 'li',
        'wrapper': 'ul',
        # Adding a class to rich text elements to make styling easier.
        'wrapper_props': {'class': 'list-styled'},
    },
    INTRO_BLOCK_ID: {
        'element': 'p',
        # Having a rich text format that's based on the same tag as other formats, + some classes.
        'props': {'class': 'intro'},
    },
})
```

### Context

This is the last missing piece to bring Wagtail's Draftail integration to parity with [WagtailDraftail](https://github.com/springload/wagtaildraftail) (and make that plugin useless, _yay!_).

It would also:

- Support what is requested in #3257 (and thus #1167).
- Help with mitigating #1214.
- Potentially replace the undocumented `register_rich_text_embed_handler` / `register_rich_text_link_handler` (#1094)

[...]

PR #4702, opened 2018-07-24
[WIP] New UI for listing object uses, including in RichText and StreamField.

gasman:
Rebase of #4481, with the following fixes applied in response to the earlier code review there:

* Better cleanup of rich text APIs (`LinkHandler` and `EmbedHandler` now deal purely with database-representation / front-end-HTML semantics, and hallo.js-specific legacy code remains elsewhere)
* the 'find all objects' mode of `ModelRichTextCollector.find_object` is split into a separate `find_all_objects` method instead of overloading
* Removed untested/undocumented logic for finding non-model values and renamed methods accordingly, e.g. `StreamFieldCollector.find_values` -> `find_objects`
* Added special cases for MySQL to account for regexp engine differences

Still to do:

* Fix logic error in `StreamFieldCollector.find_objects` when recursing over nested StreamBlocks: https://github.com/wagtail/wagtail/pull/4481/files#r190932257
* Complete review (collectors.py from `StreamFieldCollector` onward, and view-level code)


gasman:
Nested StreamBlock bug now fixed in 7ec418a8009e48220da12076d86879eac2388080.

wagtail/core/collectors.py
```diff
@@ -0,0 +1,669 @@
...
+
+        :param value: A value existing at any level of ``StreamField`` data. The type of
+            this value must match the first item in ``block_path``.
+        :param block_path: The path to a block, as a tuple of block definition objects.
+        """
+        current_block, *remaining_path = block_path
+
+        if not remaining_path:
+            # End of path reached; return model instances found within `value`
+            if isinstance(current_block, RichTextBlock) and value is not None:
```
BertrandBordage: I doubt anyone will actually do it, but what if someone writes a block type returning a `RichText` object? It will not be handled by this line, while checking for the value will handle such cases. This is why I originally checked the data types instead of the block types.

Also, is it possible to end up with a `None` value in a `RichTextBlock`, apart from a manual database rewrite?

(Making this review in the context of 19952a6bd2e95efc63af9b450183813450b93d8e)

gasman: Good point! We need to check against the block definitions for the non-leaf cases, so I was following the same pattern here for consistency... but checking the value type would indeed be more flexible. Will update...

I guess a `RichTextBlock` can't (currently) be `None`, but I think it's good to be cautious here.

BertrandBordage:
Looks great to me :smile: 
Thanks for taking over, Matt, and for the review. I know it was pretty hard to review, and I’m sorry about it, but it will be such an awesome addition to Wagtail :slightly_smiling_face: 

Later we’ll probably have to add a hook to allow users to add their own introspection rules, but that’s for another day!

wagtail/core/collectors.py
```diff
@@ -0,0 +1,669 @@
...
+            exclude = set()
+        main_use = None
+        for i, obj in enumerate(nested_list):
+            if isinstance(obj, list):
+                yield from cls.from_nested_list(
+                    obj, parent=main_use, on_delete=on_delete, exclude=exclude)
+            else:
+                use = cls(obj, parent=parent, on_delete=on_delete)
+                if use in exclude:
+                    if use not in originals:
```
gasman: @BertrandBordage If multiple objects are being passed to `get_all_uses`, it's possible for one object to be in the `originals` list and _also_ part of a `CASCADE` chain for another of those objects, isn't it? If so, we can't just skip over this logic for objects in `originals`.

gasman: ...in fact, what's the purpose of this line at all? As far as I can see, the only reason why we _might_ want to give `originals` special treatment is that they exist in the `exclude` list as plain instances rather than `Use` records, and it may or may not be a good thing to populate `Use.parent` with a plain model instance rather than a `Use`.

Given that the `parent` attribute of `Use` doesn't appear to be used, tested or documented anywhere, I'm inclined to remove it completely.

BertrandBordage: @gasman, to answer your questions (sorry in advance :slightly_frowning_face:):
- `originals` and `exclude` always contain `Use` instances, not `Model` instances.
- `Use.parent` is used in `Use.__init__`. Each `Use` instance uses the `parent` attribute of their ancestors to define its depth. I agree that it could be removed and increment `parent.depth` instead. However, I still think it’s a good thing to keep a `parent` attribute in a tree structure.
- That whole `if use in exclude: if use not in originals: […]` has a single purpose: work around the messy nested structure returned by Django’s `NestedObjects`. It can return something like this for an object `o1`: `[o1, [o5], o2, [o4], o3, [o1]]` (note that I’m simplifying because the example doesn’t contain multitable inheritance, which is correctly handled by this method). In that case:
  - `o2`, `o3` and `o5` are directly related to `o1`. Yes, it doesn’t seem to make sense because `o5` is nested while `o2` and `o3` are at the same level as `o1`. This is the main cause for the complexity of this method. `[o5]` is considered as a list of children of `o1` because it’s following `o1`. `o2` and `o3` are at a different level because they also contain related objects. I know, this is an awful tree structure, but that’s what we have to deal with.
  - `o4` is related to `o2`. We want to list it, but unlike `o5` we have to remember that `o4` is contained in `o2`, whereas `o5` was considered at the root of our uses tree structure. So this means we had to remember the previous item, `o2`, as the parent for the list of children, and therefore we store it in `main_use`. Before that, we search for `other_use`, the first instanciation of `o2` in case there is any other. This is to avoid using extra RAM because millions of uses can be found for a single deletion…
  - `o1` is related to `o3`, but we don’t want to list `o1` since we already know we want to delete it. Since `o1` is already in `exclude` from `get_all_uses`, it will not be shown. To link `o3` to `o1`, we defined the variable `main_use` containing the use preceding the list `[o1]`.
- It’s not possible for one object from `originals` to be in the `CASCADE` chain of another object. `NestedObjects` can only return an object once as a non-leaf node. For example, you couldn’t get `[o1, [o2, [o1, [o3]]]]`. There are two reasons for that, first, obviously to avoid infinite recursion, and second, because of the SQL optimisations done by `NestedObjects` to avoid triggering too many requests (even though it’s far from being optimised :frowning_face:).

**TLDR:**
**Without `if use not in originals:`** (wrong because it shouldn’t show a chevron because nothing is contained)
image
**With `if use not in originals:`** (right solution)
image

wagtail/core/collectors.py
```diff
@@ -0,0 +1,704 @@
...
+                return str(self.object)
+            url = reverse(viewset.get_url_name('edit'), args=(self.object.pk,))
+        return format_html('<a href="{}">{}</a>', url, self.object)
+
+    def get_html(self):
+        html = self.get_edit_link()
+        if self.is_root:
+            return html
+        return format_html('{}<i class="icon icon-arrow-right"></i> {}',
+                           (' ' * self.depth * 8), html)
```
BertrandBordage: Oh, this part got lost somehow, maybe during a merge: the space should be a non-breaking space, I’m sure I originally put it. @gasman please include it again, but with an explicit unicode code: `'\u00A0'`

wagtail/core/collectors.py
gasman: That's fine, but can this go in code comments please? :-)

Also, could we perhaps remove the `Use(foo) == foo` behaviour from `Use.__eq__`? I noticed that the tests are relying on that behaviour, but I have no idea if any other code is - and it's yet another thing that's making this PR extremely difficult to review, because every time I see a comparison between `Use` objects, I have to stop and ask myself if it's doing the obvious thing or whether there's some secret type-hacking going on.

BertrandBordage: Yeah, no problem with removing the ability to compare `Use` and `Model` instances.

Well, you see the problem with putting all that as code comments x)
Tricky to do such complex explanations while keeping the code readable. Do you want me to add a docstring+a few inline comments in this method? I can make a commit in my old branch that you could cherry-pick.

gasman: Yes please!

BertrandBordage:
@gasman, you can cherry-pick the latest commit I added to this branch: https://github.com/BertrandBordage/wagtail/commits/BertrandBordage-new-uses-ui-rebase-2
i.e. bbbe75dd53b6b9da90832a943ffba4fae577c8aa, e8ca0bdad492b32d82507e2838c41f559b45c246, edc723deff8433d7764d376019aa8b6821651780, 456c82eaab3476b94fc0454d6dca36e7c2be8a13 & 509e574aca50cb989c78fce720fdf6e26110b3ec

They contain all the changes we’ve discussed :wink:

gasman:
Rebased with latest fixes cherry-picked.

wagtail/core/collectors.py
```diff
@@ -555,24 +567,35 @@ def from_nested_list(cls, nested_list, parent=None, on_delete=CASCADE,
...
                 if use in exclude:
-                    if use not in originals:
-                        for other_use in exclude:
-                            if use == other_use:
-                                main_use = other_use
-                                break
+                    # When ``use`` is in ``originals``, we don’t want
+                    # to mark it as parent of the next item because we want
+                    # to display the next item as root, otherwise the first
+                    # level of deleted objects would be shown nested.
```
gasman: @BertrandBordage OK, let me see if I've got this right: this logic comes into play when `nested_list` is a structure such as:

    [a, [b], b, [c]]

where:

* `b` is one of the items from `originals`
* `b` appears in the child list of an earlier item
* the top-level occurrence of `b` in the list is immediately followed by a sub-list

In this case, we do _not_ set `parent_use` to `b` - it gets left at its existing value, which is `a`, and so `c` ends up being assigned as a child of `a`. Is that what you intended to happen? Is this _actually_ a legitimate way that NestedObjects might represent `c` being a child of `a`, and are we relying on the fact that if `c` was meant to be a child of `b`, the list would look like `[a, [b, [c]], b, [c]]` (in which case the second `c` would be skipped along with the second `b`)? Ouch...

gasman: @BertrandBordage Ah, I missed the detail that `exclude` always contains the originals whenever we call this - lines 565-566 (which are never used?) led me to understand that the initial state was the empty set. Could we make `exclude` a required argument, or at least change those lines to

    if exclude is None:
        exclude = set(originals)

so that it's not misleading?

wagtail/core/collectors.py
```diff
@@ -0,0 +1,722 @@
...
+                use = cls(obj, parent=parent, on_delete=on_delete)
+                # When ``use`` was already listed, we don’t want to list it
+                # but we want to mark it as parent of the next item, in case
+                # that next item is a list.
+                if use in exclude:
+                    # When ``use`` is in ``originals``, we don’t want
+                    # to mark it as parent of the next item because we want
+                    # to display the next item as root, otherwise the first
+                    # level of deleted objects would be shown nested.
+                    if use in originals:
```
gasman: @BertrandBordage Would it be valid to set `parent_use = None` at this point? I think this is what's hurting my head so much about this code 😀 : the fact that `parent_use` is being left at its previous value in this case, which might be something left over from way earlier in the iteration. I suspect you're relying on an unspoken rule about the behaviour of `NestedObjects` which ensures that `parent_use` is only ever `None` at this point: i.e. it will never give you a situation like

    originals = exclude = {a, b}
    nested_list = [a, c, b, [d]]

where a non-original (`c`) appears at the top level before all the originals have been output. If so, setting `parent_use = None` would make the intention clearer, and save us from having to figure out whether that rule is really true.

BertrandBordage: Yes, I’m indeed relying on that order rule, as I mentioned in my previous message.
You’re totally right, we should explicitly add another `parent_use = None` here to make it extra clear.
I tried too much to do an optimal algorithm…

wagtail/core/collectors.py
BertrandBordage: We can indeed add it. I kept it consistent with `from_flat_iterable`, but that was probably not the best decision.

wagtail/core/collectors.py
BertrandBordage: Sorry btw for the headache :frowning_face:

gasman:
Rebased on current master, to take advantage of the fact that the rich text API changes are now merged courtesy of @chosak in #5184...

There are some failing tests now, which look like they're `assertQueryCount`-related - if so, we should give those tests their own fixture file, so that they don't become a huge maintenance burden when extending the 'main' fixtures in future.

BertrandBordage:
Very cool, thank you @gasman!

gasman:
@solarissmoke has fixed the query counts in bb84a961eec9c9376c4911d4500474b3ad16fafc, but combining this PR with a couple of other patches (#5266, #5286, #5353, none of which change the test fixtures AFAICS) for a custom build has caused them to change again. Not sure what's causing the variation, but that'll definitely need to be dealt with before this can be merged.

BertrandBordage:
Alright, I will get back to this cursed PR this summer ^^"

bruecksen:
@BertrandBordage do you think there will be any progress soon? One of my clients is wanting this feature and I'm wondering if I should invest any time to make it working or wait until this PR is finished and shipped in a future version.

wagtail/snippets/templates/wagtailsnippets/snippets/confirm_delete.html
```diff
@@ -16,27 +16,29 @@
...
+                {% csrf_token %}
+                <input type="submit" value="{% trans 'Yes, delete' %}" class="button serious" />
+                <a href="{% if next %}{{ next }}{% else %}{% url 'wagtailsnippets:list' model_opts.app_label model_opts.model_name %}{% endif %}" class="button button-secondary">{% trans "No, don't delete" %}</a>
+            </form>
         {% endif %}
-        <form action="{{ submit_url }}" method="POST">
-            {% csrf_token %}
-            <input type="submit" value="{% trans 'Yes, delete' %}" class="button serious" />
-            <a href="{% if next %}{{ next }}{% else %}{% url 'wagtailsnippets:list' model_opts.app_label model_opts.model_name %}{% endif %}" class="button button-secondary">{% trans "No, don't delete" %}</a>
-        </form>
```
sir-sigurd: `{% include 'wagtailadmin/shared/uses.html' %}` should be added here.

gasman:
Now rebased at #6129 - will keep both PRs open for now so that @sir-sigurd's feedback (thanks!) doesn't end up getting lost in transmission... Copying my comment from there:

Right now I think the main thing standing in the way of merging this is that it really needs refactoring to be less "monolithic". As it stands, the logic in collectors.py is touching on the internal details of many areas of Wagtail (rich text, streamfields, foreign key on_delete behaviour...) which is going to make it hard to maintain - for example, if we extend rich text to support other kinds of object references beyond linktype and embedtype (see #4223) then it's not immediately clear what needs to change in collectors.py.

I think the key to that will be making the concept of "a collector" into a well-defined internal API, so that whenever we have a field type that can contain object references (which currently includes ForeignKey, RichTextField and StreamField), there's a corresponding Collector class that implements a few standard methods for querying those references. After the previous review on #4702 (which teased out a lot of the internal workings of collectors.py and added docstrings) we hopefully have a fairly good idea of what those methods will be - something along the lines of:

* `my_collector = RichTextCollector(model_class, field_name)` - constructor for a collector that can find object references in one particular field of one particular model
* `my_collector.find_candidate_references(obj)` - return a queryset of objects (of type `model_class`) that _may_ contain references to the given object, within the field under consideration. This will be done through a fast query (e.g. a regexp match) that may return false positives; these will subsequently be verified individually via `get_object_references`.
* `my_collector.get_object_references(instance)` - given an instance of `model_class`, return a list of `(model_class, id, reference_type)` tuples for all object references found in the field under consideration. `reference_type` will be an identifier similar to the `on_delete` parameter which indicates what the effect of a deletion will be (cascade, protect, set null, ignore etc).

Given how rich text and streamfields interact in RichTextBlock, it may also be necessary to add some lower-level methods to this API, such as "return a database-specific regexp for matching an object reference".

Hopefully splitting the logic up in this way will also help to make our tests more tightly-scoped, and pin down what's causing the changing query counts.

gasman:
An alternative mechanism for reference-counting has now been implemented in #9279. Rebuilding the deletion confirmation page from this PR on top of that mechanism should be a relatively small task, at which point this can be considered resolved.

laymonage:
Superseded by #10072. Instead of showing the references on the delete confirmation view, we're reusing the existing usage view and adding the on delete information there if you access it through a link on the delete confirmation view.

PR #4703, opened 2018-07-24
Cleanup Dockerfile and reduce the amount of intermediate layers

imanhodjaev:
In this PR I tried to enhance Dockerfile for project template by reducing the amount of intermediate layers and removing redundant commands.

Thanks.


mikedingjan:
Hi @imanhodjaev ,

Adding the requirements and a seperate ``pip install`` makes it possible to cache the layer when dependencies don't change, which in my opinion is a better approach.

Any thoughts on this?

imanhodjaev:
@mikedingjan I agree with you so we will have cached layer with dependencies and the rest can remain the same.

mikedingjan:
Hi @imanhodjaev,

This still won't work as you would expect.. since you're adding the complete folder to ``/code/`` before installing anything the layers can't be cached when any file changes..

```
COPY . /code/
RUN pip install -r /code/requirements.txt && \
    pip install gunicorn
``

However when you add the requirements.txt first and than install the dependencies these 2 layers can be cached when no requirement changes.. Having said this, I think it's better to keep the ``Dockerfile`` as is and only make a minor change to have the ``RUN`` statements combined..

For example:
```
COPY ./requirements.txt /code/requirements.txt
RUN pip install -r /code/requirements.txt && \
         pip install gunicorn

COPY . /code/
WORKDIR /code/

RUN useradd wagtail && \
         chown -R wagtail /code && \
         python manage.py migrate
```

imanhodjaev:
@mikedingjan I see now, thanks for support 🌟

PR #4709, opened 2018-07-31
Security audit of npm packages

gasman:
Upgraded gulp-sass to 4.0.1 and run `npm audit fix`. After this upgrade, the only difference to the static files generated by `npm run build` appears to be the sourcemap data, so there should be no regressions.

Before:

    npm audit on master locked deps:
    found 424 vulnerabilities (314 low, 88 moderate, 21 high, 1 critical) in 15939 scanned packages
      run `npm audit fix` to fix 386 of them.
      38 vulnerabilities require semver-major dependency updates.

After:

    found 10 vulnerabilities (1 low, 5 moderate, 4 high) in 21498 scanned packages
      run `npm audit fix` to fix 1 of them.
      5 vulnerabilities require semver-major dependency updates.
      4 vulnerabilities require manual review. See the full report for details.

From what I can see, the remaining fixes involve upgrading to gulp 4.x (which is not yet stable) and dealing with the `hoek` package pulled in by various dependencies of `gulp-sass`.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -14,19 +14,19 @@
       "resolved": "https://registry.npmjs.org/yargs/-/yargs-4.7.1.tgz",
       "integrity": "sha1-5gQyZYozh/8mnAKOrN5KUS5Djf8=",
       "requires": {
-        "camelcase": "^3.0.0",
-        "cliui": "^3.2.0",
-        "decamelize": "^1.1.1",
-        "lodash.assign": "^4.0.3",
-        "os-locale": "^1.4.0",
-        "pkg-conf": "^1.1.2",
-        "read-pkg-up": "^1.0.1",
-        "require-main-filename": "^1.0.1",
-        "set-blocking": "^1.0.0",
-        "string-width": "^1.0.1",
-        "window-size": "^0.2.0",
-        "y18n": "^3.2.1",
-        "yargs-parser": "^2.4.0"
+        "camelcase": "3.0.0",
+        "cliui": "3.2.0",
+        "decamelize": "1.2.0",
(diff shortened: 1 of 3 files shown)


BertrandBordage:
21498 dependencies… :roll_eyes:
That’s 15% of the total number of packages on PyPI!

thibaudcolas:
Sorry, I mistakenly pushed those two commits to master while reviewing them 😞. Pushed the revert right afterwards.

gasman:
@thibaudcolas Unfortunately adding the .css extension causes the `@import` line to be left intact in the final draftail.css.

Will look at upgrading to Node 10 now. Afraid I don't know enough about our node tooling to confidently make the changes to the client packages - will leave that with you if that's OK.

gasman:
Oh, that's annoying - Node 10 doesn't have a 32-bit package for Linux any more, so we'd have to either build from source (which takes ages) or switch vagrant-wagtail-develop to 64 bit (which seemed to be a resource hog last time I tried it) :-(

gasman:
@thibaudcolas It looks like upgrading to Node 10 is going to be a yak-shave too far for this PR... besides the 32-bit issue, it also breaks gulp 3.9.x (unless you do some further hacking around with dependency versions, apparently). For now I've replaced package-lock.json with the NPM 5 version (i.e. the one you end up with after running `npm install` under NPM 5) and verified that the upgrades are still applied when installing from this version - hope this is OK.

thibaudcolas:
@gasman here is the removal of the unused `client/` code that is not used, with its dependencies: https://github.com/wagtail/wagtail/pull/4723

Issue #4661, opened 2018-06-29
Upgrade CircleCI from 1 to 2


tomdyson:
We're using version 1 of CircleCI for CI on the Wagtail project. According to [their docs](https://circleci.com/docs/1.0/language-python/) August 31st 2018 will be the final day for 1.0 builds.

We'll need to port our CircleCI config from 1.0 to 2.0, following these instructions:

https://circleci.com/docs/2.0/migrating-from-1-2/

NDevox:
I've been doing a fair bit of circleci 2 recently and the config isn't complex, can move it over for you if you want.

gasman:
@NDevox Yes please - that'd be really helpful!

gasman:
Completed in #4712.

PR #4712, opened 2018-08-01
Migrate to Circle CI 2.0

jjanssen:
In order to prevent doing this last minute by the end of the month, I already prepared the Circle CI migration which was described in #4661. 

Some changes from the previous configuration: 
- There are now 2 separate jobs for the backend and the frontend. Both of them will be kickstarted by the workflow;
- [The CircleCI Python image is providing `pipenv` preinstalled](https://github.com/circleci/circleci-images/issues/168#issuecomment-391204259), I used that to execute the Python commands in order to prevent `Permission denied` errors when just running `pip install`.
- Explicit cache handlers before and after installing dependencies per branch (`pipenv`, `npm`) (v1 did this by default?);

By using the circle-cli I already tested  the two separate jobs, but currently the cli does not support testing of the workflows. Fingers crossed in what it will do when running this PR 😎


gasman:
Nice one - thanks @jjanssen!

Issue #2339, opened 2016-03-10
Focal point too abstract; need ability to crop images per-instance


alexgleason:
For example, a page model may display an image in a horizontal banner, then the same image is displayed on that page's index as a square thumbnail. There's no editorial ability to manage those separately.

I can envision two possible implementations:
1. Configure your page to have a thumbnail field and banner field separately. The client can choose to make them the same. Then, a crop can be applied _per-field_. That is, a "crop" button may appear next to "edit", "delete", "choose another" and apply that crop only to this instance in this field.
2. Allow multiple focal points to be set _per-image_. For instance, the client can crop each image for a "banner", "thumbnail", etc. then they are templated accordingly.

Any thoughts?

heymonkeyriot:
From an editor point of view I think option 1 may be easier to understand conceptually though, even if it does feel like it's leading to replication. I agree that option 2 feels more elegant, but I'm aware that there are editors that struggle with the concept of a single focal point let alone trying to introduce multiple. Would be exciting to see if we could make it work, but I think would be an interesting design/ UX challenge to pull off. @alexgleason how were you thinking you could communicate those multiple focal points for a single image?

JoshBarr:
1. Sign up for [prismic.io](http://prismic.io)
2. Use the crop feature in there for a few minutes to see how it can be 100% awesome.
3. Implement here :)

davecranwell:
Focal point is hard to explain to clients. Either their image isn't quite large enough for the use, or it's not sufficiently well composed for the focal point to have sufficient effect.

The duplication created by cropping is only a problem if you treat the crop and the original as separate images. They aren't - they're renditions of the same image, just as our existing concept of renditions doesn't create two images.

I'd expect that the image chooser field would become two fields: 1. a pointer to the image record to use and 2. a pointer to a specific rendition of that image, created manually by the user (or defaulting to the whole image otherwise).

tmsndrs:
I think there are a few issues here. 

**UI **
The focal point UI is potentially confusing in the way it shapes an editors expectations. It feel like a crop UI but actually all you're doing is defining a primary point of focus on the image. It would make more sense for the UI to use a cross hair rather than a marque to define the point IMO. 

**Context **
[...]

PR #4714, opened 2018-08-03
Browze wagtail contextual crop

browze:
**Description and status**

This pull request is for a 'contextual' cropper, also using the jcrop api, which will permit the developer to enable the editor to specify a crop for each image within the context of a page or snippet and which will be applied only within the context of the particular page or snippet being worked on.  

I have checked the image tests all still pass but not yet written documentation or tests of my own - though I've given it a fairly thorough usage testing.  I would be happy to supply these elements, perhaps after some discussion, but first I would like to submit it for comment.

**Motive**

I conceived of the feature partly to address the point raised [here](https://github.com/wagtail/wagtail/issues/2339), but also because I found that while the generic image focal point is useful where you want the image to show up in a set space (or spaces) and you want the image to be generally as well adjusted as possible with the minimum effort, it doesn't really help the editor in cases where the editor might want more control of the exact dimensions of an image for streamfield layout.  A good example might be an article.  The editor may want to use an existing image of something,  but the image may be too tall for an article banner, so they might choose to focus on a strip of the image. Using the contextual cropper they would be able to do this without having to upload another cropped copy of the original image, name it and assign it, possibly repeating the process one or more times if they want to adjust it a bit. 

**Usage**

my_image_field = SelectCropBlock(required=True, template="my_folder/my_image_spec.html")

**Basic Components and Approach:**

The main elements of the new feature under the hood are:

1. Providing a way of storing the crop data for the context rather than on the image. I found json, leveraging the great struct block feature to be a good way of simplifying this - as opposed to creating relationship models.  [SelectCropBlock, SelectCropChooserBlock]
2. Provide a new widget based on AdminImageChooser that generates the rendition for the admin based on the focal point information captured (SelectCropAdminImageChooser)
3. Allowing the chosen image previews to be defined at a larger size than 165x165 to make the interface usable in the page/snippet editing context.  This is a very small update and might stand on its own - I know there's been some discussion about making the preview size dynamic but for me a decent half way house is to make it possible to adjust this site wide in the settings.  The little css I applied here (in focal-point-chooser.css) should never impact the main image edit form as the main edit form doesn't sit within a preview block.
4. Updating the admin jcrop js to allow for multiple images per page without interference between them (I created another js crop file for the main cropping stuff: 'contextual-focal-point-chooser.js', as there was little DRY advantage and significant potential for maintenance complication in trying to just use just one set of functions, but on the other hand I was able to merely adjust the image-chooser.js to make it handle more than one image if necessary and although I added two fields to the response they won't impact the existing focal point widget in any way)
5. Adding another 'image operation' that can use information from that source ('select' - which basically leverages the 'fill' op but forces c100 and injects the contextual crop data from the struct block which it is sent during rendering (see 6))
6. I've prioritised ensuring it is applied quietly before any image operations that the developer might then define in image specifications in the template, without any need for duplication of effort by adding further template specs - after all we have a GUI with a crop select that should make that unnecessary. To do this I injected a bit of preprocessing into the wagtailimage_tags image tag, but it only operates where it finds the relevant contextual fields at the top level, so it will never affect the existing image based focus area. The impact this has on usage is the only pain point - as I noted in the code, to enable the image tag to do its work with the context, the select_crop_chooser_block from blocks.py needs to render itself to add to that context.  However, although render_child will render it, we can't just provide the template for the block, because the developer will likely want to add their own specs.  Fortunately the developer can specify a template to the block constructor. 
E.g.: my_image_field = SelectCropBlock(required=True, template="my_folder/my_image_spec.html") I took the liberty of providing a default template for the block that will display a message explaining the basic usage if its not overidden in the block constructor (select_crop_no_template.html), but if this is considered too much of a departure from the general wagtail approach (I suspect it is) we could either specify a default image spec in a default template or simply allow it to literally render a string representation of the fields by not providing a default at all.  I'd be interested in any suggestions in relation to streamlining the front end config.


browze:
I note it failed styling (sorry forgot to check).  I'd be happy to go through and make sure it passes if the PR is of interest.

tomdyson:
Wow! Thanks very much for this PR, @browze, and the detailed notes. Could you please include a couple of screenshots or a screencast, to help us evaluate it from a UI point of view, as well as a technical one?

browze:
Yep, sure np.  A random image dragged from somewhere:

Before saving:

screen shot 2018-08-03 at 12 47 44

After saving:
screen shot 2018-08-03 at 12 48 49

As it's rendered on the site (without showing any surrounding web features like menu bars etc - just the image):

screen shot 2018-08-03 at 12 50 11

BTW I factored out two routines, 'reset' and 'remove' in the contextual-focus-point-chooser and running one of them can be controlled by passing a string argument to the jcrop run function, along with the focal point block in question.  I added the 'remove' one to the image-chooser.js imageChosen callback to the runJcrop function, so it gets immediately after a new image is selected.  That basically removes the focal area so a fresh one can be selected for a fresh image, but it only affects the image changed.

browze:
@tomdyson Hi Tom, as you may have noticed I've ironed out some issues I found since my first commit of this PR.  Just wondered whether you think its likely to prove of interest?  Just a nudge, no great hurry - I understand you guys must be beyond busy.  

At first I'd written this feature to make the most of hooks and subclasses in the website code, but then I thought it was really the sort of thing that might nicely augment the wagtail feature set, especially when I saw the discussion I linked to above, and besides I was more or less forced to make at least one change to the wagtail code anyway (I think it was a small change to enable the wagtail fill operation to use the contextual data) - so I integrated it.  If it doesn't seem a good idea to offer this as part of wagtail, at some point I may get to work disentangling it and see if it could work as some sort of module/plugin, so I don't have to maintain a different wagtail.  

I can say that so far its been well received by my client :)

Thanks for your consideration and thanks for Wagtail!

tomdyson:
Hi @browze, thanks for the prompt, and sorry for the delayed response. As it stands, I think the feature is certainly of interest, but the implementation will need some work (following a review by one of my more expert colleagues). My recommendation for now would be to carry out the disentangling work you described, so that users can add this feature to their Wagtail sites with e.g. `pip install wagtail-contextual-crop`. That means that you won't have to maintain a different Wagtail, and it makes it easy for other people in the community to use it and give feedback, outside the Wagtail upgrade cycle.

Thanks again for all your work on this!

P.S. It looks like your UI refers to 'focal areas' - I think 'cropped area' would be a clearer label.

browze:
@tomdyson - my turn to apologise for the delay.  Thanks for getting back to me - I'll take the advice and turn this into a module as soon as I get a chance.  

Regarding the labelling - yes, agreed.

Naturally I'm interested in your colleague's assessment and I'm keen to know what sort of work they were referring to (I presume they were reviewing the code since the last commits on the 4th August?).  For example it would be of use to know whether they feel there may be intrinsic issues with the approach (especially keen to know what those are if that's the case) or whether there are simply better opportunities to ‘build it in’ - or other considerations (as this started website side and I moved it as explained above and I’m obviously fully familiar with all parts of the wagtail internals, I wouldn’t be totally shocked).  Of course I get that they must be very busy, but perhaps as they've already reviewed it at some point someone could provide just a little more detail?  

Cheers!

tomdyson:
Hi @browze, I'm afraid my comment wasn't very clear - I meant that I expected that implementation changes would be requested after a proper review, not that one had already happened. I've tagged @chosak in the hope that he might have some time to look at this in the coming weeks, but I recommend that you continue with converting it into a module in any case.

FWIW we've just launched a client site where this feature would have been very helpful!

luizboaretto:
I also have clients waiting for this. It would be very useful! :boom:

browze:
Ok great @Tomdyson @luizboaretto - good to know its of interest to others!  I’ll try to get to this as soon as poss.  

On reflection though, before the review is carried out, I thought it might be worth mentioning that there are at least three areas where it feels to me that it sits most ‘naturally’ in wagtail proper. I can see potential ways of knocking all the bits into a package - but the basic ideas that first occurred to me at a glance would be somewhat 'contrived' to avoid integration.  

Perhaps it might even make sense to place particular parts of it inside, allowing for more 'natural' integration of an outside package - the parts in question fall back to essential no-ops where very specific conditions in specific contexts aren't met.  

Anyway, here are the three particular bits that occurred to me when I gave returned to the code and gave it a once over:

1. My use of the wagtail image tag code (to be more precise the ImageNode class used by the image tag code) to detect and apply any contextual crop specification from the struct automatically before any other template specification supplied by the developer.  I do believe using the image tag is much better than creating a parrallel one from a usage perspective, so I'd have to import subclass, wrap and ensure that the package was installed in the correct order to avoid a fairly modest alteration to the tag.

2. 	A simple single line adaptation of the FillOperation class to pick up the focal point associated with the ImageBlock where it exists, rather than the image itself.   It may makes sense eventually to have the FIllOperation and the SelectCropOperation (or whatever we call it) invoke a shared Operation for the common bits but that would still require integration.  

3. Cache key generation for the crop in the image models get_cache_key method.

I'll happily undertake to move such things if on review its determined thats the best approach, even if only for the short term, but if a review will be made anyway, I thought it might be worth at least getting the benefit of it first before deciding if that’s the best thing to do?  I say this partly because at present the feature does seem to have been behaving itself...

Thanks!

PS In the latest update I've pushed a validation restriction in the image model to prevent a non-sensical combination of the crop operation with the image level ‘fill’ operation - it might be worth discussing at some point if raising an error or simply logging a warning and ignoring one or other operation is the way to go in this scenario.  

PPS if its desirable, once we’ve settled on an implementation, obviously **it would be very easy to add any other image operations as contextual ones**, as they’d just leverage the existing operations and the ‘contextual’ pattern we hammer out here.

browze:
@chosak @tomdyson just a nudge on this :)

browze:
@chosak no problem whatever and thank you for the review - much appreciated. I'm currently under a bit snowed under but will come back to you on all of your comments very soon!

ebertti:
👍 

Great!!! When will be merged?

browze:
@ebertti not sure :)  Based on the @chosak 's remark above I think he would prefer to see it incorporated into the standard ImageChooser widgets rather than as a special struct block, so the behaviour is uniform across wagtail.  I'm for this and I think it would add value. Hopefully what I've done already for the structblock scenario is a good part of this.  @chosak I know perhaps my response above is quite a bit to wade through, though I did try to more or less respond to your points.  I take it from your last comments you'd be happy to see me attempt to add this ability to the standard ImageChooserBlock ?  I imagine this would probably entail the extra fields necessary (select_x, select_y, select_width_x and select_width_y or similar) being added somehow as 'non model' fields? Right now the chooser I think just stores an id - a single value on the model that includes  the chooser block - I'll revisit this to remind myself very soon as its been a while since I've looked at this or done work on it. Do you have any advice on how to go about storing the extra values - how you'd like to see this done - is/has this sort of thing been done elsewhere on the project? Or shall I just experiment, probe the source and see if I can have a go at it adding the extra info from scratch?

chosak:
@browze apologies for the delay in responding; I haven't had time to look at this for a while. Thank you for your very detailed reply. I will read through it again and respond in more detail.

One quick thing: the "the edit page seems to always think that its edit state is dirty" behavior is unrelated to your PR and is covered in #4978.

chosak:
@browze still haven't had much time to look into this. Unfortunately, I notice that the proposed StreamField frontend refactor (#4942, I believe targeted for a future 2.6 release of Wagtail) will no longer support the `form_classname` and `form_template` meta attributes that you use on your SelectCropBlock here.

There's some discussion starting [here](https://github.com/wagtail/wagtail/pull/4942#issuecomment-472165937) about how to do something similar in the new (React) implementation, but I haven't looked into this too deeply and don't have a good sense about what would need to be done. Hopefully when that feature gets released it'll come with some good documentation about migrating existing blocks that use those attributes. @BertrandBordage

lb-:
Closing as this PR will need a lot of work to bring inline with changes to Wagtail since 2018.

However, I will add a comment on https://github.com/wagtail/wagtail/issues/2339 with some notes about this PR.

PR #3478, opened 2017-03-22
3227 use unlocalize

whyscream:
This PR solves https://github.com/wagtail/wagtail/issues/3227

* Do the tests still pass? [v]
* Does the code comply with the style guide? [v]
* For Python changes: Have you added tests to cover the new/fixed behaviour? [n/a]
* For new features: Has the documentation been updated accordingly? [n/a]

PR #4716, opened 2018-08-03
Convert some admin widgets to Django-1.11-style template rendering

gasman:
See #4715 for background. Convert `AdminAutoHeightTextInput`, `AdminDateInput`, `AdminTimeInput`, `AdminDateTimeInput` and `AdminTagWidget` to template-based rendering, eliminating the need for the `WidgetWithScript` helper and following standard Django idioms so that they can be subclassed and customised in the same way as Django's built-in widgets.


wagtail/admin/templates/wagtailadmin/widgets/chooser.html
```diff
@@ -8,7 +6,7 @@
     when clicked.
 {% endcomment %}
 
-<div id="{{ attrs.id|unlocalize }}-chooser" class="chooser {% block chooser_class %}page-chooser{% endblock %} {% if not value %}blank{% endif %}">
```
loicteixeira: I don't really understand why `unlocalize` was here in the first place so I don't think it would be an issue to remove it but what's the reasoning behind it? Opportunistic cleanup?

This also applies to `wagtail/contrib/table_block/templates/table_block/widgets/table.html`.

gasman: Yeah, this commit is a bit out of place due to me abandoning this strand of work at this point :-) This was going to be the first step in refactoring chooser.html to work with both the old-style and new-style template rendering, but I gave up on that idea and so this remains as a bit of standalone cleanup.

(As for why `unlocalize` was here in the first place: it was erroneously added in #3478 alongside various legitimate fixes to prevent numeric object IDs from being rendered with local number formatting.)

gasman:
@loicteixeira Now rebased, and `HalloRichTextArea` and `TableInput` conversions done.

PR #4718, opened 2018-08-03
Remaining Django 2.1 compatibility fixes

gasman:
Along with #4715 and some as-yet-unreleased patches to django-taggit (https://github.com/alex/django-taggit/pull/505 / https://github.com/alex/django-taggit/pull/518 / https://github.com/alex/django-taggit/pull/534) and django-modelcluster (https://github.com/wagtail/django-modelcluster/pull/99), this gets the test suite passing on Django 2.1 (at least when testing against sqlite / no-elasticsearch).


loicteixeira:
I imagine this PR will be updated to bump `django-taggit` and `django-modelcluster` to newer versions once the mentioned PRs are merged.

However keep in mind that alex/django-taggit#518 drops Django 1.11 support which Wagtail 2.3 will still support. This might be a bit awkward to deal with where we will leave the lower bound as it is (so it can be used with Django 1.11) but bump the upper bound (so it can be used with Django 2.1) and potentially ask developers to add an explicit requirement for the correct version of `django-taggit` for their setup 😞

loicteixeira:
Indeed. My bad sorry.

PR #4723, opened 2018-08-07
Remove unused React component generator

thibaudcolas:
Initially discussed removing this over at https://github.com/wagtail/wagtail/pull/4709#pullrequestreview-142143097.

I think this generator was added at the same time as the Webpack tooling, for the new explorer menu. The scaffolding might have worked but we've never used it much simply because we don't create new components that often. As far as I know, this was not documented anywhere anyway (beyond having the `npm run component` command), and removing it shouldn't break anything.


package.json
```diff
@@ -76,7 +76,6 @@
     "gulp-sourcemaps": "~2.6.1",
     "gulp-util": "~3.0.8",
     "jest": "^22.0.3",
-    "mustache": "^2.2.1",
```
thibaudcolas: I'm not sure what this was doing here. Removing it isn't related to the rest of the changes, but as far as I can see this dependency isn't used anywhere (anymore?).

BertrandBordage: Well, mustache was used to generate the .mst template, right?
So it’s in fact related to the other changes you did :)

thibaudcolas: Well true :) What I meant was that the `mustache` used for this generator was declared in `client/package.json` – so there shouldn't have been one for that purpose in the root package.json.

gasman:
Merged in ad5a3bbdc2ee9a3c9adb3150ed71eaa403a8390e / 6301eb4b75c7829bee33b334ef6ef0a755bd3dbd

PR #4726, opened 2018-08-09
Prevent search engines from indexing admin pages

kaedroho:
According to a google search I just did, it seems a lot of people have forgotten to add ``Disallow: /admin`` in their robots.txt (or forgot to add robots.txt) at all.

Adding this meta tag into the head of all admin pages should prevent any admin pages being indexed even if this was missed.


gasman:
Merged in ebecaf6a5b972196a5515dcbadac18a00a470c3b 👍

BertrandBordage:
Could we use `X-Robots-Tag` instead? That’s more generic, as it also covers non-HTML data.
By the way @kaedroho, lots of people willingly never use robots.txt for security.

Issue #1471, opened 2015-07-03
Remove default richtext image formats


michaelfillier:
Is there a way to remove the default image formats provided by wagtail. "Full width", "Left-aligned" and "Right-aligned".

We are using register_image_format for our custom sizes as it is a perfect way to add the right css classes and tie the images into our css framework. We are not using the default options and would like to exclude them from the admin interface, since they will produce undesired results.

gasman:
Hi @michaelfillier,
You can do this with the `unregister_image_format` function:

    from wagtail.wagtailimages.formats import unregister_image_format
    unregister_image_format('fullwidth')
    unregister_image_format('left')
    unregister_image_format('right')

(Note that this will cause any existing pages using these image formats to throw an error when viewed or edited. It's on our todo list to change this behaviour so that unregistered image formats are just hidden from the admin interface, rather than being forgotten completely...)

michaelfillier:
Thank you.

rjsnh1522:
I am still facing the issue of error page when i unregister_image_format.
is it solved . or do you have any idea how can i remove image formats without getting error for each format

PR #4728, opened 2018-08-10
Add warning about unregistering image formats.

martey:
Add warning to documentation to warn that unregistering image formats
that are being used will create an error when viewing or editing pages.

See https://groups.google.com/forum/#!topic/wagtail/X8xTUs-2npA and
https://github.com/wagtail/wagtail/issues/1471#issuecomment-118436706.


gasman:
Thanks @martey!

PR #4731, opened 2018-08-14
add ability to overwrite wagtail logo with own logo for the userbar

Meteor0id:
In the userbar the wagtail logo is loaded though a font. This behavior remains, however the font icon may now be overwritten by an image.

To acomplish this the classes which define the font icon have been split out to an icon tag which is a child of the userbar trigger div. The `<i>` tag is containted within a block called branding_logo, so the user may overwrite this block similar to overwriting the branding logo in other templates.

Please test my work, and modify if needed. (should go without saying)

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/scss/userbar.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/userbar.scss
@@ -141,7 +141,7 @@ $positions: (
         left: 0;
     }
 
-    &. #{$namespace}-icon:before {
+    .#{$namespace}-icon:before {
         transition: color 0.2s ease;
         font-size: 32px;
         width: auto;
--- a/wagtail/admin/static_src/wagtailadmin/scss/userbar.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/userbar.scss
@@ -141,7 +141,7 @@ $positions: (
         left: 0;
     }
 
-    &.#{$namespace}-icon:before {
+    &. #{$namespace}-icon:before {
         transition: color 0.2s ease;
         font-size: 32px;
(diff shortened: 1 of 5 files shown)


Meteor0id:
This also provides an alternative to overwriting the icon using a font, and could likely be condisdered a solution to issue #2353

Meteor0id:
If it gets merged, you can create a file `dashboard/templates/wagtailadmin/userbar/base.html`:
```
{% extends "wagtailadmin/userbar/base.html" %}
{% load staticfiles %}

{% block branding_logo %}
    <img src="{% static 'ffs_base/img/my_logo.svg' %}" alt="my logo" />
{% endblock %}
```

Meteor0id:
I have not fully checked if the size or dimnetions of the custom logo will be an issue. Possbile one or two lines should be added to userbar.scss to make sure the image is always containted within the "wagtail-usebar-trigger" class, and does not overflow it.
Shouldn't be hard to do if needed.

Meteor0id:
Reviewing my own pull request, I see that I made a new pull request after this one, which includes every change made in this one.

So this PR might be ignored completely, instead review PR #4741

wagtail/admin/static_src/wagtailadmin/scss/userbar.scss
```diff
@@ -137,7 +137,7 @@ $positions: (
         left: 0;
     }
 
-    &.#{$namespace}-icon:before {
+    &. #{$namespace}-icon:before {
```
thibaudcolas, line 140: Typo? The dot needs to be at the start of the class name, and `&` is useless if not directly joined with the rest of the selector.

PR #4739, opened 2018-08-17
Fix image upload failures when using django-storages (Google Cloud Storage, Azure Storage)

### LINKED ISSUE #4738: Image upload fails with django-storage backends on Wagtail 2.2
### Issue Summary

Image upload fails with `django-storage` backends on Wagtail 2.2. I can reproduce it with Google Cloud Storage backend, but there was also at least one report in Slack about the issues with Azure Storage.

I suspect projects that use other storage backends that expect file to be rewinded to the beginning before uploading it will also have issues.

It seems that the issue was introduced in Wagtail 2.2 in #4516.

The issue is that Wagtail reads the image file to generate hash, but doesn't rewind the file to the beginning (doesn't call `seek(0)`) and storage backends expect to file to be rewinded to the beginning.

### Steps to Reproduce

1. Start a new project
1. Add `django-storage` to your project and set up the `GoogleCloudStorage` backend for media files
1. Try to upload the image in the Admin UI

### Technical details

* Python version: 3.6.6.
* Django version: 2.0.8
* Wagtail version: 2.2.1

In Wagtail 2.1.1 image upload with the same version of `django-storage` works well

m1kola:
Rewind file to the beginning after reading

Fixes #4738

### DIFF
--- a/wagtail/images/views/chooser.py
+++ b/wagtail/images/views/chooser.py
@@ -150,6 +150,7 @@ def chooser_upload(request):
             # Set image file hash
             image.file.seek(0)
             image._set_file_hash(image.file.read())
+            image.file.seek(0)
 
             form.save()
 
--- a/wagtail/images/views/images.py
+++ b/wagtail/images/views/images.py
@@ -104,6 +104,7 @@ def edit(request, image_id):
                 # Set new image file hash
                 image.file.seek(0)
                 image._set_file_hash(image.file.read())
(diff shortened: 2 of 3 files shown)


gasman:
Merged in 550e8f648cdf91b8dfbcf3cdba4a84481c647463 (master) and 064d2d15444750582c8424fa725b06e03863ae47 (2.2.x) - thanks @m1kola!

PR #4741, opened 2018-08-18
Custom branding title on admin interface AND add ability to overwrite wagtail logo with own logo for the userbar

Meteor0id:
Provide an easy way of customizing the branding in the title of the admin interface, and add an example of it to the documentation. Experience with end users reveals that they tend to ask "Why does it say wagtail at the top?" Reading the name of the CMS system their organization is using has no value for the end user (an editor). They consider it clutter of their work environment with information meant for the system administrator, rather than an website editor. For them it is more important to know they are in an editors interface.

The branding can also be replaced with nothing, allowing the title of the page to only show the name of the page being edited, without any prefix.
By default it will still say Wagtail, but it can now be changed using an admin template.
**This PR essentially makes it _possible_ to hide the name Wagtail in many places (but not everywhere)**

On top of that, this PR also adds ability to overwrite wagtail logo with own logo for the userbar:

In the userbar the wagtail logo is loaded though a font. This behavior remains, however the font icon may now be overwritten by an image.
To accomplish this the classes which define the font icon have been split out to an icon tag which is a child of the userbar trigger div. The <i> tag is contained within a block called branding_logo, so the user may overwrite this block similar to overwriting the branding logo in other templates.</i>

You might wonder, why solve two issues in one PR? Because I pushed enhancement for the userbar to my master branch at the time, I made a mess and didn't know how to get a clean branch. But both are Front End enhancements, so they can be reviewed together in one test. The original PR for the userbar is #4731, but all those commits are also part of this PR.


Meteor0id:
I have tested on desktop in firefox 61.0.2 at various resolutions. I also tested in on the android version of Firefox (also 61.0.2).

thibaudcolas:
I see `branding_title` has been added already in #5344 so will close this PR in favour of #4731.

Meteor0id:
@thibaudcolas I see there is a difference between #5344 which edits the skeleton, and this PR which edits the userbar. Why is that?

PR #4743, opened 2018-08-21
Convert UUID primary key to string before serializing in json (Rebase of #4626)

gasman:


gasman:
Merged in 6e60469142a11ac6df041e75c3537dcc538039ae

PR #4746, opened 2018-08-23
Allow nav menu to take up all available space instead of scrolling

Meteor0id:
I noticed the nav-main had too much margin, always assuming the footer is in expanded position, and even than it was inaccurate. Fixed it.

To reproduce the issue, open an admin interface page with the footer not expanded, and resize your browsers height. You will notice the nav-main does not take up all the space it has, leaving a significant margin unused.

This PR fixes that.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -17,6 +17,7 @@ Changelog
  * Fix: Additional fields on custom document models now show on the multiple document upload view (Robert Rollins, Sergey Fedoseev)
  * Fix: Help text is partially hidden when using a combination of BooleanField and FieldPanel in page model (Dzianis Sheka)
  * Fix: Allow custom logos of any height in the admin menu (Meteor0id)
+ * Fix: Allow nav menu to take up all available space instead of scrolling (Meteor0id)
  * Fix: Redirects now return 404 when destination is unspecified or a page with no site (Hillary Jeffrey)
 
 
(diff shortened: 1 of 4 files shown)


Meteor0id:
example of admin nav not sizing properly


Might I add I am very suprosed a mature CMS like Wagtail has simple front-end issues like these? Glad I could help.

Meteor0id:
wagtail transparent account background issue

Also fixed this transparent background, by giving the footer a background, rather than just the submenu. commit: 1ee4bba

Meteor0id:
Right, apologies.

I have tested on desktop in firefox 61.0.2 at various resolutions. I also tested in on the android version of Firefox  (also 61.0.2). It looks, feels, acts and smells excatly as expected (it smelled refreshing).
I do not expect odd behavior on other browsers, as the css used wasn't spectaculary cutting edge, but of course it should be tested, that is what testing is for.

You meantion rather using flex-box. One of the other PR's I recnetly opened turns most of the main-nav section into a flex box, doing exactly that. This footer however still relies on the magic number approach for now. I will go see if I can include it in the flex-box when I have more time.

Never the less, I do recommand pulling this in. It is a working fix, and it will be compatible with the other PR, which turns the main-nav into a flex box. It will at least adress an ugly issue untill I get some time to refactor the whole thing.

Please do reply with any questions or feedback, I have this thing fresh in my head still.

thibaudcolas:
@Meteor0id thanks for having a shot at this. I have rebased this onto the latest `master` and will now review.

PR #4749, opened 2018-08-24
EditHandler “dynamic” forms

BertrandBordage:
Currently, `EditHandler`s are bound to a model, then forms are generated **then** the model instance and request are bound to the `EditHandler`, along with the generated form. It prevents users from generating different `EditHandler` tree structures (and therefore forms) depending on the current instance or request.

This has been a blocking limitation in numerous projects I worked on, and the common workarounds today are:
- display all fields and inform editors (not safe)
- hide some fields via JavaScript (not safe)
- server-side data validation depending on the instance or user (safe, but not as clean as “dynamic” forms)

What I call a “dynamic” form is not a form that would dynamically change in JavaScript; it’s a form that changes between one request and another.

This pull request replaces `EditHandler.bind_to_model()` and `EditHandler.bind_to_instance()` with a single `EditHandler.bind_to()` method that allows to bind any of these objects as soon as possible: `model`, `instance`, `request` & `form`.

This work also binds requests and saved model instances as soon as possible. That way, developers are able to access the request and instance (if it exists) before the form class is generated.

Here is an example generating different `content_panels` depending on a user permission, with this pull request applied:
```python
class PerUserContentPanels(ObjectList):
    def on_instance_bound(self):
        self.children = self.instance.basic_content_panels
        if self.request.user.is_superuser:
            self.children = self.instance.superuser_content_panels
        self.children = [
            child.bind_to(model=self.model, instance=self.instance,
                          request=self.request, form=self.form)
            for child in self.children]


class PerUserPageMixin:
    basic_content_panels = []
    superuser_content_panels = []

    @cached_classmethod
    def get_edit_handler(cls):
        tabs = []

        if cls.basic_content_panels and cls.superuser_content_panels:
            tabs.append(PerUserContentPanels(heading=_('Content')))
        if cls.promote_panels:
            tabs.append(ObjectList(cls.promote_panels,
                                   heading=_('Promote')))
        if cls.settings_panels:
            tabs.append(ObjectList(cls.settings_panels,
                                   heading=_('Settings'),
                                   classname='settings'))

        edit_handler = TabbedInterface(tabs,
                                       base_form_class=cls.base_form_class)

        return edit_handler.bind_to(model=cls)


class Home(PerUserPageMixin, Page):
    secret_data = TextField()

    basic_content_panels = Page.content_panels
    superuser_content_panels = basic_content_panels + [
        FieldPanel('secret_data'),
    ]
```

Obviously, this is not that simple to use for now, but at least it gives full flexibility while slightly simplifying code.

This work was commissioned by @kutenai from [ShaperTool](https://sharpertool.com/) :smiley:

### DIFF
--- a/wagtail/admin/edit_handlers.py
+++ b/wagtail/admin/edit_handlers.py
@@ -1,4 +1,5 @@
 import re
+from warnings import warn
 
 from django import forms
 from django.core.exceptions import ImproperlyConfigured
@@ -21,6 +22,7 @@ from wagtail.utils.decorators import cached_classmethod
 # DIRECT_FORM_FIELD_OVERRIDES, FORM_FIELD_OVERRIDES are imported for backwards
 # compatibility, as people are likely importing them from here and then
 # appending their own overrides
+from wagtail.utils.deprecation import RemovedInWagtail25Warning
 from .forms import (  # NOQA
     DIRECT_FORM_FIELD_OVERRIDES, FORM_FIELD_OVERRIDES, WagtailAdminModelForm, WagtailAdminPageForm,
     formfield_for_dbfield)
@@ -96,13 +98,20 @@ class EditHandler:
         self.heading = heading
         self.classname = classname
         self.help_text = help_text
+        self.model = None
+        self.instance = None
+        self.request = None
+        self.form = None
 
     def clone(self):
-        return self.__class__(
-            heading=self.heading,
-            classname=self.classname,
-            help_text=self.help_text,
-        )
+        return self.__class__(**self.clone_kwargs())
+
+    def clone_kwargs(self):
+        return {
+            'heading': self.heading,
+            'classname': self.classname,
+            'help_text': self.help_text,
+        }
 
     # return list of widget overrides that this EditHandler wants to be in place
     # on the form it receives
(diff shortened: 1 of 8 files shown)


roelio:
A welcome change, currently I'm struggling to show a panel depending on the user making the request. When the panel is hidden and the page is saved the field are emptied, because I cannot dynamically exclude fields from the form when the POST request is happening. 
Think this PR will also fix https://github.com/wagtail/wagtail/issues/4779

BertrandBordage:
That’s right @roelio, it should fix #4779.

gasman:
Have rebased this at https://github.com/gasman/wagtail/tree/BertrandBordage-edit-handler-dynamic-forms , and made a start on turning the above example code into a test (https://github.com/gasman/wagtail/commit/3b8dc08874a78ada06e66c34f0f1af39be232fcc), but it's failing with `Key 'title' not found in 'SecretPageForm'`, and I can't immediately see why :-/

gasman:
The `PerUserContentPanels` example / test case is now fixed on https://github.com/gasman/wagtail/tree/BertrandBordage-edit-handler-dynamic-forms . It was failing on page creation because it relies on the instance being bound before the form (which now happens in https://github.com/gasman/wagtail/commit/f2d1f3de4cd04f92f4d77649b4d6a7dde84b8591) - otherwise we end up constructing the form before we have the final `children` list in place. Meanwhile, it was failing on page edit because the custom `on_instance_bound` logic is only valid once both the instance and request are populated, so it has to wait for whichever is the later one of `on_instance_bound` and `on_request_bound` - see updated version in https://github.com/gasman/wagtail/commit/e68173983d4d93e2716e7cd6c45e8b0705a36b6c#diff-d2d058dc2ff8fff791e5b62fdd165cbdR1271 . I haven't figured out why binding the instance and request at the same time (`edit_handler.bind_to(request=request, instance=page)`) isn't sufficient though...

This does raise some alarm bells about maintainability - if it turns out that custom edit handlers have to rely on the `bind_to` calls happening in a precise sequence, then that's probably going to limit our ability to refactor this in future. Anyway, will pick this up in the new year.

gasman:
Merged in dc2f4873912eb537a4b2799c888cea04bb826151

BertrandBordage:
Thank you @gasman!  Sorry I couldn’t spend more time on it in January, I appreciate a lot that you took the time to finish it :heart:

stevil32:
As far as I can tell I am following this code to the tee, however I keep getting an error with the PerUserContentPanels class. It keeps telling me the request is None. Any thoughts?

PR #4751, opened 2018-08-26
Add 'add_redirect' static method

brylie:
This method was inspired by a Lacey Williams Henschel (@williln) presentation at Wagtail Space US 2018, called  'What the Wagtail docs don't tell you'.

https://www.youtube.com/watch?v=PCkxBNXWM64&t=1220s

Co-authored-by: Lacey Williams Henschel <lacey@revsys.com>

### DIFF
--- a/wagtail/contrib/redirects/models.py
+++ b/wagtail/contrib/redirects/models.py
@@ -49,7 +49,20 @@ class Redirect(models.Model):
             return cls.objects.filter(models.Q(site=site) | models.Q(site=None))
         else:
             return cls.objects.all()
-
+    
+    @staticmethod
+    def add_redirect(old_path, redirect_page, is_permanent = True):
+        redirect = Redirect()
+        
+        # Set redirect properties from input parameters
+        redirect.old_path = Redirect.normalise_path(old_path)
+        redirect.redirect_page = redirect_page
+        redirect.is_permanent = is_permanent
+        
+        redirect.save()
+        
+        return redirect
+        
     @staticmethod
     def normalise_path(url):
         # Strip whitespace
--- a/wagtail/contrib/redirects/models.py
+++ b/wagtail/contrib/redirects/models.py
@@ -51,12 +51,12 @@ class Redirect(models.Model):
             return cls.objects.all()
 
     @staticmethod
-    def add_redirect(old_path, redirect_page, is_permanent=True):
+    def add_redirect(old_path, redirect_to=None, is_permanent=True):
         """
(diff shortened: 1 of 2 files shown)


gasman:
Thanks @brylie! Could you add tests and documentation for this, please?

brylie:
OK, will work on it ASAP. Likely weekend work, as we are not using it at my day job :-)

williln:
@brylie Thanks so much for adding this! I'm so excited that you took an idea from my talk and ran with it. 😄

brylie:
I am taking a look at the project structure now, so I can write a test for this method. Where are tests defined for the Redirect class?

brylie:
@gasman I added a basic test and some documentation in the form of a method docstring. Please review.

brylie:
Ah, it looks like the `redirect_page` needs to be a Page instance.

@williln or @gasman, how should I create a page instance in the test?

gasman:
@brylie In a test like this we'd usually use one of the existing pages defined as part of the `test.json` fixture - see `test_redirect_to_page` a bit further down the same file for an example.

Since it's possible for a redirect to point to either a page or a URL path, could I suggest that we support both cases, using a signature like

    def add_redirect(old_path, redirect_page=None, redirect_link=None, is_permanent=True):

so that users can call either `Redirect.add_redirect('/foo', redirect_link='/new-foo')` or `Redirect.add_redirect('/foo', redirect_page=foo_page)`? Better to support both setups now rather than to have to go back and change it later, I think.

Thanks for adding the docstring... we'll need to link that somewhere in the docs, but as we don't currently have any developer-facing documentation for redirects (only [the page in the editor's guide](https://docs.wagtail.io/en/stable/editor_manual/managing_redirects.html)) it probably calls for a new page under [reference/contrib](https://docs.wagtail.io/en/stable/reference/contrib/index.html). I'll take care of that.

brylie:
@gasman I updated the function signature a bit, and added a Page in the test. Please review the progress, code design, docstring, etc :-)

brylie:
@gasman I updated the function signature a bit, and added a Page in the test. Please review the progress, code design, docstring, etc :-)

lb-:
@brylie thanks for your work on this, I know it has been a while.

I took a look at the discussion and what was done and finished this off + merged in.

Really appreciate the work that you and Lacey put into this.

Merged in via ff37ad6ab9aaad8f6526515623ea79fc2742a69b

@gasman  FYI I have created the reference contrib/redirects page also. Happy to flesh this out more if needed on a separate PR.

PR #4754, opened 2018-08-28
Prevent the use of a global variable.

schinckel:
Having global variables in these methods results in errors when the class is processed by BabelJS.


gasman:
Looks good, thanks! Merged in e05dc9948603d92c3112d83dadf8129cab784455.

PR #4771, opened 2018-09-10
templates/wagtailadmin/shared/field.html: use safe with help_text

blueyed:
Without the HTML on e.g. `/admin/account/change_password/` gets rendered
as-is:

> `<ul><li>Your password can&#39;t be too similar …`

Django uses the safe filter itself.

Ref: https://docs.djangoproject.com/en/2.1/ref/models/fields/#help-text

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/shared/field.html
+++ b/wagtail/admin/templates/wagtailadmin/shared/field.html
@@ -6,12 +6,12 @@
             {% block form_field %}
                 {{ field|render_with_errors }}
             {% endblock %}
-            
+
             {# This span only used on rare occasions by certain types of input #}
             <span></span>
         </div>
         {% if show_help_text|default_if_none:True and field.help_text %}
-            <p class="help">{{ field.help_text }}</p>
+            <p class="help">{{ field.help_text|safe }}</p>
         {% endif %}
 
         {% if field|has_unrendered_errors %}


gasman:
Thanks @blueyed! However, please see the discussion on #3398 - before we add `|safe` here, we need to audit all of our help text to confirm that this won't be introducing any HTML injection vulnerabilities.

blueyed:
I see.

What a patch like this then? (with a comment referring to #3398)
```diff
diff --git i/wagtail/admin/views/account.py w/wagtail/admin/views/account.py
index 2a385b324..9fe810acb 100644
--- i/wagtail/admin/views/account.py
+++ w/wagtail/admin/views/account.py
@@ -70,6 +70,11 @@ def change_password(request):
                 return redirect('wagtailadmin_account')
         else:
             form = PasswordChangeForm(request.user)
+
+        for field in form.fields.values():
+            from django.utils.safestring import mark_safe
+            mark_safe(field.help_text)
+
     else:
         form = None
```

I think it gives a bad first impression, especially since you are likely to change the password when using a demo (I am trying out the bakerydemo).

blueyed:
btw: it gets generated in https://github.com/django/django/blob/bd5ce0599bf0dc156b943ca0d03307c7e451923c/django/contrib/auth/password_validation.py#L78-L85

I think this is an issue in Django itself: it should not generate HTML for this, but rather have a list instead and then iterate over it in the template.

PR #4773, opened 2018-09-11
Add HTTP_AUTHORIZATION to values copied to dummy preview requests

### LINKED ISSUE #4772: Preview dummy requests don't include the "Authorization" header
### Issue Summary

We're using [django-basic-auth-ip-whitelist](https://pypi.org/project/django-basic-auth-ip-whitelist/) to limit access to a site. For us, this breaks page previews, because the `Authorization` header isn't included in the dummy request that Wagtail constructs. This is because our templates depend on attributes of `request` that are set by middleware. 

This is related to #3546, but I think in this case it would be reasonable to add `HTTP_AUTHORIZATION` to the list of things that Wagtail copies to the dummy request in `Page.dummy_request()`. We're already copying `HTTP_COOKIE` and `HTTP_USER_AGENT`, so this seems more like an omission than a design decision.

### Steps to Reproduce

1. Add `django-basic-auth-ip-whitelist` to a Wagtail site and set a username and password
2. Log in to set the `Authorization` header. Viewing live pages and the Wagtail admin should work fine
3. Click "preview" on any page whose template refers to `request.user` or another request attribute set by middleware.

### Technical details

* Python version: 3.6.1
* Django version: 2.0.3
* Wagtail version: 2.0
* django-basic-auth-ip-whitelist version: 0.2.1

bcdickinson:
This PR fixes #4772 by adding the `HTTP_AUTHORIZATION` header to those copied from the original request to the dummy request for previews.

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -1242,7 +1242,7 @@ class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):
 
         # Add important values from the original request object, if it was provided.
         HEADERS_FROM_ORIGINAL_REQUEST = [
-            'REMOTE_ADDR', 'HTTP_X_FORWARDED_FOR', 'HTTP_COOKIE', 'HTTP_USER_AGENT',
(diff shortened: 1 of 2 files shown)


gasman:
Looks good to me! Merged in 4c72f767ea3a532d858f857735f8d2da19f524ef.

PR #4775, opened 2018-09-13
Use better examples in ModelAdmin docs, add info about FieldPanels

alexgleason:
Updates the docs for ModelAdmin.

My main reason is that it's very hard to know how to use FieldPanels like `ImageChooserPanel` for a normal Django model with ModelAdmin.

The most user-friendly way I could imagine demonstrating this is with an example, so I added a models.py example to the index page, complete with an ImageChooserPanel and a tip about adding FieldPanels. Hopefully this will make web searches like `ModelAdmin ImageChooserPanel` link here now.

I ended up rewriting all the examples to use an analogy instead of just placeholder text. I think it's more approachable this way.

### DIFF
--- a/docs/reference/contrib/modeladmin/index.rst
+++ b/docs/reference/contrib/modeladmin/index.rst
@@ -71,9 +71,37 @@ How to use
 A simple example
 ^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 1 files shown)


gasman:
Merged in e5f9c122ab514fdfced25e57b3e827ad3216d64d.

PR #4781, opened 2018-09-18
Hooks to customise page add/edit actions

gasman:
Refactor the actions list in the footer of the page create/edit views so that rather than being a hard-coded list, it consists of individually renderable components much like the main admin menu - and provide `register_page_action_menu_item` and `construct_page_action_menu` hooks for customising it.

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/pages/action_menu/menu_item.html
+++ b/wagtail/admin/templates/wagtailadmin/pages/action_menu/menu_item.html
@@ -1,0 +1,7 @@
+<li>
+    {% if url %}
+        <a href="{{ url }}">{{ label }}</a>
+    {% else %}
+        <input type="submit" name="{{ name }}" value="{{ label }}" class="button" />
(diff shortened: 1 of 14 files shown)


wagtail/admin/templates/wagtailadmin/pages/action_menu/menu_item.html
```diff
@@ -0,0 +1,7 @@
+<li>
+    {% if url %}
+        <a href="{{ url }}">{{ label }}</a>
+    {% else %}
+        <input type="submit" name="{{ name }}" value="{{ label }}" class="button" />
```
kaedroho: Any reason why ``<input type="submit">`` is used here rather than ``<button type="submit">`` that is used elsewhere?

kaedroho:
Merged in: https://github.com/wagtail/wagtail/commit/cb8148ede2ac67ecce750136b8c6900af887ed86 + parents

Thanks!

Issue #4055, opened 2017-11-23
Improved UI visual styles of chooser fields


tmsndrs:
Current implementation feels clunky. 
<img width="396" alt="screen shot 2017-11-23 at 12 04 09" src="https://user-images.githubusercontent.com/7290588/33172064-89b392e4-d046-11e7-93a1-885fbab9d484.png">

I've put together some early thinking on how these could be improved but it requires further conversation. 
![image](https://user-images.githubusercontent.com/7290588/33172119-b11d6fe4-d046-11e7-93e6-19981eee7345.png)

lb-:
It might be good to consider #2364 relating to visual indication of loading the next page of selectable images within chooser modal, maybe filling in non-loaded images with a placeholder blank / greyed out image and also add a spinner to indicate loading.

benenright:
Hi

I've refined these designs for the chooser fields - as mocked up by @tmsndrs above. I'm attempting to reflect the global button size/style for the initial view, and then creating a more padded stroked area once the media has been chosen - to house more cleanly the media title and grouped actions associated with it ( Clear / Edit / Replace) - replacing the existing buttons: Trash button / 'Edit this page' /  'Choose another page' 

Ignore the fact the icons vary slightly from the existing ones - happy to continue using existing icons for now (wholesale change to icons would be a separate issue). But I do feel we could move towards using the white icon on teal (default btn view) and the darker shade of grey for the populated item.

One thing to bear in mind is the top aligned label. Labels are currently aligned left but this ties in to work on'WIP Streamfield UI changes' #3942 where there is a move towards top aligned labels.

![screen shot 2018-02-14 at 14 09 26](https://user-images.githubusercontent.com/4907819/36208820-f0d9d6ea-1191-11e8-9af9-2365e027938a.png)

![screen shot 2018-02-14 at 14 09 34](https://user-images.githubusercontent.com/4907819/36208819-f0c56d7c-1191-11e8-878b-ce5fe8d15fd4.png)

Ben

benenright:
Hi

The [designs](https://github.com/wagtail/wagtail/issues/4055#issuecomment-365621149) above are now agreed by the Wagtail UX/UI working group and ready for FE implementation for anyone who is keen to take this on.

Here is a mockup to show how these will work in situ with existing fields (note the key difference in label placement - vertically aligned labels in place of left-aligned which we hope to roll out globally soon to help with narrow screen alignment issues).
[...]

PR #4790, opened 2018-09-27
Fixes layout for FileField

### LINKED ISSUE #4780: Layout for the default widget for FileField looks broken
### Issue Summary

Layout for the default widget for FileField looks broken. The template seems to come from Django itself.

### Steps to Reproduce

1. Create a page with a `models.FileField`. For example:
    ```python
    header_video = models.FileField(blank=True, verbose_name='Hero video')
    ```
2. Add a `FieldPanel` into `content_panels`:
    ```python
    FieldPanel('header_video')
    ````

**Result:**

filefield

### Technical details

* Python version: 3.6.6
* Django version: 2.0.8
* Wagtail version: 2.2.2
* Browser version: Google Chrome 69.0.3497.81 and FireFox 63.0b7s and Safari 11.0.1

m1kola:
Fixes #4780 

**Tested in the following browsers:**

* Chrome 69.0.3497.100 (including "iPhone 6/7/8" and "iPad" options in dev tools)
* Safari 11.0.1
* FireFox 63.0b7 and 63.0b9

**Screenshot:**

file_field_fixed

**Note for a reviewer:**

I'm not sure about the approach. This PR is relying on the structure of HTML (partly from Wagtail and partly from Django) and existing classes which looks a bit fragile to me, but in `_forms.scss` there are a lot of overrides like this, so I wanted my changes to be consistent with the rest of the file.

An alternative approach is to override the `ClearableFileInput` widget in Wagtail and use own template with more specific CSS classes.

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/scss/components/_forms.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/components/_forms.scss
@@ -93,6 +93,27 @@ select::-ms-expand {
     display: none;
 }
 
+.file_field {
+    .input {
+        label {
+            float: none;
+            display: inline;
+            padding: 0;
+        }
+
+        input[type=checkbox] {
+            margin-top: 5px;
+        }
+
+        a {
+            &:after {
+                content: ' ';
+                display: block;
+            }
+        }
+    }
+}
+
 // select boxes
 .choice_field .input,
 .model_choice_field .input,


m1kola:
@jjanssen Looks great!

I think, this will require a custom widget with own template.

jjanssen:
Please note, I'm not a designer 😂

zerolab:
worth keeping #4055 in mind when doing this.

m1kola:
@jjanssen @zerolab does it make sense to do a simple solution now and sort it out in #4055 properly?

jjanssen:
Yes it would make sense. I think with the plain chooser we can more easily solve it now. Yet I don't know when the other ticket would get picked up. It's still on my list but haven't gotten to it just yet.

lb-:
Merged in via https://github.com/wagtail/wagtail/commit/d2d8b2c9e343e6a9532d615f079dd13e7003e80e

PR #4795, opened 2018-09-28
Gracefully handle document links with missing ID attributes in rich text

### LINKED ISSUE #4791: KeyError when saving a page containing a document link to a deleted document in a rich text field
### Issue Summary

`KeyError` when saving a page containing a document link to a deleted document in a rich text field. This affects all rich text fields regardless of whether it is a streamfield block or not.

### Steps to Reproduce

1. Edit the rich text field on a page
2. Add a document link using the 'document' button
3. Save page
4. Delete the document from the site
5. Edit the page with the document link again
6. Save it
7. You'll encounter the `KeyError 'id'` exception at location `...wagtail/documents/rich_text.py in get_attribute_data, line 65`
8. Review the previous working revision (eg. `admin/pages/<page id>/revisions/<revision id>/revert/`)
9. View page source of this edit page, you'll notice that the entity map for the deleted document has got empty data. This is probably what's causing the problem when saving the page. Eg. (the first entity is the working document link, vs the second is the deleted document with empty data)

```
        &quot;1&quot;: {
            &quot;mutability&quot;: &quot;MUTABLE&quot;,
            &quot;type&quot;: &quot;DOCUMENT&quot;,
            &quot;data&quot;: {
                &quot;id&quot;: 36,
                &quot;url&quot;: &quot;/documents/36/working.pdf&quot;,
                &quot;filename&quot;: &quot;working.pdf&quot;
            }
        },
        &quot;2&quot;: {
            &quot;mutability&quot;: &quot;MUTABLE&quot;,
            &quot;type&quot;: &quot;DOCUMENT&quot;,
            &quot;data&quot;: {}
        }
```
10. Click on 'Replace current draft', same error occurs again.

### Technical details

* Python version: 3.6.4
* Django version: 1.11.11
* Wagtail version: 2.1.1
* Browser version: firefox 61.0.2

gasman:
Fixes #4791
Previously, our rich text conversion functions handled the case where a document link specified an ID which is not found in the database. However, they failed with a KeyError when the id attribute was missing completely; links of this second type would occur whenever a link of the first type was re-saved from the Draftail editor. The fix is two-fold:

1) Catch the "missing ID attribute" case - in this case, the resulting link will be missing both the href and id attributes
2) Update the handling of the "ID present but document not found" case so that the id attribute survives the round-trip to the editor and back. The final link as rendered on the front-end will still be an attribute-less <a> element, but the id will be retained in the database (and in the versions rendered within rich text editors) which may be useful for troubleshooting.

### DIFF
--- a/wagtail/documents/rich_text.py
+++ b/wagtail/documents/rich_text.py
@@ -13,7 +13,7 @@ def document_linktype_handler(attrs):
     try:
         doc = Document.objects.get(id=attrs['id'])
         return '<a href="%s">' % escape(doc.url)
-    except Document.DoesNotExist:
+    except (Document.DoesNotExist, KeyError):
         return "<a>"
 
 
@@ -31,7 +31,9 @@ class DocumentLinkHandler:
             doc = Document.objects.get(id=attrs['id'])
             return '<a data-linktype="document" data-id="%d" href="%s">' % (doc.id, escape(doc.url))
         except Document.DoesNotExist:
-            return "<a>"
+            return '<a data-linktype="document" data-id="%s">' % attrs['id']
+        except KeyError:
+            return '<a data-linktype="document">'
 
 
 EditorHTMLDocumentLinkConversionRule = [
@@ -62,9 +64,14 @@ class DocumentLinkElementHandler(LinkElementHandler):
     def get_attribute_data(self, attrs):
         Document = get_document_model()
         try:
-            doc = Document.objects.get(id=attrs['id'])
-        except Document.DoesNotExist:
+            id = int(attrs['id'])
+        except (KeyError, ValueError):
             return {}
+
+        try:
+            doc = Document.objects.get(id=id)
+        except Document.DoesNotExist:
+            return {'id': id}
 
         return {
             'id': doc.id,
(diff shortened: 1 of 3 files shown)


wagtail/documents/rich_text.py
```diff
@@ -31,7 +31,9 @@ def expand_db_attributes(attrs):
             doc = Document.objects.get(id=attrs['id'])
             return '<a data-linktype="document" data-id="%d" href="%s">' % (doc.id, escape(doc.url))
         except Document.DoesNotExist:
-            return "<a>"
+            return '<a data-linktype="document" data-id="%s">' % attrs['id']
```
loicteixeira: It might be worth adding a comment here. Outside of this PR, I would be very puzzled as why we print out the document's ID in the output despite the document not being found.

wagtail/documents/rich_text.py
```diff
@@ -62,10 +64,15 @@ class DocumentLinkElementHandler(LinkElementHandler):
     def get_attribute_data(self, attrs):
         Document = get_document_model()
         try:
-            doc = Document.objects.get(id=attrs['id'])
-        except Document.DoesNotExist:
+            id = int(attrs['id'])
```
loicteixeira: In general, we should not cast PKs to `int` as it prevent the user from using UUIDs.

I understand that the `Document` model is built-in and therefore cannot use UUIDs, therefore it's fine here, but Wagtail makes such assumptions in different places and it has been a pain point in the past.

gasman:
Merged (with comment added) in bf870cecd3b0258f12ecbbfcd69bb0619b3f6a8b

PR #4796, opened 2018-09-28
Add template include page_switches in admin/edit.html

mikedingjan:
I'm working on a feature in ``wagtailtrans`` which allows the user to easily switch between languages in the admin. Having this block will allow me to extend the page switches which will also provide an easy method of switching when editing a page.

Related PR: https://github.com/LUKKIEN/wagtailtrans/pull/136

Edit: Updated the block to a template include, since they're easier t override.


gasman:
Looks good to me 👍

PR #4797, opened 2018-09-29
Updated _to_struct_value to work with (fully dot-pathed) string model…

c4collins:
… Meta.value_class names.


gasman:
Thanks @c4collins! Please could you add tests and documentation for this?

PR #4800, opened 2018-10-05
Properly handle links to root page and missing pages in Draftail

### LINKED ISSUE #4605: Draftail editor allows to link to root page rendering page non-functional
### Issue Summary

The following has come to my attention as a user has been complaining about a non-functional editing interface for a seemingly random sub-page.  The Draftail rich text editor allows to link to the root page of the installation:

bildschirmfoto 2018-06-13 um 15 06 05

Consequently, a non-functional link is insert as follows:

bildschirmfoto 2018-06-13 um 14 45 46

Saving this page renders an error - nevertheless the page seems to be persisted. Its impossible to edit the page thereafter.

### Steps to Reproduce

2. Edit models.py to include a page model with a RichTextField with the `link` feature enabled.
3. Add a link to the fields text clicking the "Root" page in the "Link"-popup.

I'm not entirely sure I'm using the feature correctly. Is it possible to reduce the available pages in the richtexts link popup to not include the root page?

### Technical details

* Python version: 3.6.5
* Django version: Look in your requirements.txt, or run `2.0.2`.
* Wagtail version: 2.1
* Browser version: Safari 11.1.1

gasman:
Fixes #4605, along with a similar issue to #4791 but for broken page links rather than document links: our Draftail integration uses the presence of an ID to differentiate page links from external URL links, so it makes sense to preserve the ID in the link data even if the corresponding page is missing.

### DIFF
--- a/client/src/components/Draftail/sources/ModalWorkflowSource.test.js
+++ b/client/src/components/Draftail/sources/ModalWorkflowSource.test.js
@@ -61,7 +61,13 @@ describe('ModalWorkflowSource', () => {
 
       it('page', () => {
         expect(getChooserConfig({ type: 'LINK' }, {
-          getData: () => ({ id: 1, parentId: 0 })
+          getData: () => ({ id: 2, parentId: 1 })
+        }, '')).toMatchSnapshot();
+      });
+
+      it('root page', () => {
+        expect(getChooserConfig({ type: 'LINK' }, {
+          getData: () => ({ id: 1, parentId: null })
         }, '')).toMatchSnapshot();
       });
 
--- a/wagtail/admin/rich_text/converters/html_to_contentstate.py
+++ b/wagtail/admin/rich_text/converters/html_to_contentstate.py
@@ -206,7 +206,12 @@ class PageLinkElementHandler(LinkElementHandler):
         try:
             page = Page.objects.get(id=attrs['id']).specific
         except Page.DoesNotExist:
-            return {}
+            # retain ID so that it's still identified as a page link (albeit a broken one)
+            return {
+                'id': int(attrs['id']),
+                'url': None,
+                'parentId': None
+            }
 
         parent_page = page.get_parent()
 
--- a/wagtail/admin/rich_text/converters/html_to_contentstate.py
+++ b/wagtail/admin/rich_text/converters/html_to_contentstate.py
@@ -208,10 +208,12 @@ class PageLinkElementHandler(LinkElementHandler):
         except Page.DoesNotExist:
             return {}
 
+        parent_page = page.get_parent()
+
         return {
             'id': page.id,
             'url': page.url,
-            'parentId': page.get_parent().id,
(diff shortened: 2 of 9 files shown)


client/src/components/Draftail/sources/ModalWorkflowSource.test.js
```diff
@@ -61,7 +61,13 @@ describe('ModalWorkflowSource', () => {
 
       it('page', () => {
         expect(getChooserConfig({ type: 'LINK' }, {
-          getData: () => ({ id: 1, parentId: 0 })
+          getData: () => ({ id: 2, parentId: 1 })
+        }, '')).toMatchSnapshot();
```
thibaudcolas: 👌

wagtail/admin/rich_text/converters/html_to_contentstate.py
```diff
@@ -206,12 +206,19 @@ def get_attribute_data(self, attrs):
         try:
             page = Page.objects.get(id=attrs['id']).specific
         except Page.DoesNotExist:
-            return {}
+            # retain ID so that it's still identified as a page link (albeit a broken one)
+            return {
+                'id': int(attrs['id']),
+                'url': None,
+                'parentId': None
+            }
```
thibaudcolas: From what I understand, this combination of having an id and not having a URL will only happen for broken page links. Could we leverage this to display them differently in the editor? e.g. use some kind of "warning" UI cues, and potentially have a label that says "this links look broken".

This doesn't need to be done as part of this PR of course, but it seems like a nice potential improvement.

Edit: here, https://github.com/wagtail/wagtail/blob/09cddd81f69660e5d8836eb205355c917dc9b49b/client/src/components/Draftail/decorators/Link.js#L20-L22

and then there,

https://github.com/wagtail/wagtail/blob/09cddd81f69660e5d8836eb205355c917dc9b49b/client/src/components/Draftail/decorators/TooltipEntity.js#L105-L113

we use the URL as a label for the link tooltip. I will have a look at what happens if this is set to `None`.

wagtail/admin/rich_text/converters/html_to_contentstate.py
```diff
@@ -206,12 +206,19 @@ def get_attribute_data(self, attrs):
...
+
+        parent_page = page.get_parent()
 
         return {
             'id': page.id,
             'url': page.url,
-            'parentId': page.get_parent().id,
+            'parentId': parent_page.id if parent_page else None,
```
thibaudcolas: Under what circumstances would a page exist and not have a parent page?

gasman: @thibaudcolas When it's the root page. We don't allow the root page to be picked as of dc093c96461ab1e40fe0b42bc7ee09383bf8d71d, but this way we avoid breaking on links that already existed prior to this fix (and it's more robust to have a server-side check for this case).

PR #4801, opened 2018-10-05
Reuse lazy StreamField values in get_prep_value. Fixes #2218

### LINKED ISSUE #2218: PageChooserBlock/ImageChooserBlock values go null when restoring from loaddata

gasman:
Fixes #2218 - proposed alternative fix to #4727. This version ensures that the same code path (`StreamBlock.get_prep_value` calling `StreamValue.get_prep_value`) is used for both top-level and nested StreamBlocks, and also preserves the behaviour of assigning an ID on save if one is not already present.

### DIFF
--- a/wagtail/core/blocks/stream_block.py
+++ b/wagtail/core/blocks/stream_block.py
@@ -426,11 +426,6 @@ class StreamValue(collections.Sequence):
             if self.is_lazy and i not in self._bound_blocks:
                 # This child has not been accessed as a bound block, so its raw JSONish
                 # value (stream_data_item here) is still valid
-
-                # assign a new ID if it didn't have one already
-                if not stream_data_item.get('id'):
-                    stream_data_item['id'] = str(uuid.uuid4())
-
                 prep_value_item = stream_data_item
 
             else:
@@ -439,9 +434,14 @@ class StreamValue(collections.Sequence):
                 prep_value_item = {
                     'type': child.block.name,
                     'value': child.block.get_prep_value(child.value),
-                    # assign a new ID on save if it didn't have one already
-                    'id': child.id or str(uuid.uuid4()),
+                    'id': child.id,
                 }
+
+            # As this method is preparing this value to be saved to the database,
+            # this is an appropriate place to ensure that each block has a unique id.
+            if not prep_value_item.get('id'):
+                prep_value_item['id'] = str(uuid.uuid4())
+
             prep_value.append(prep_value_item)
 
         return prep_value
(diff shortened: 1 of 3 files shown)


wagtail/core/blocks/stream_block.py
```diff
@@ -424,6 +419,33 @@ def _prefetch_blocks(self, type_name, child_block):
...
+                prep_value_item = stream_data_item
+
+            else:
+                # convert the bound block back into JSONish data
+                child = self[i]
+                prep_value_item = {
+                    'type': child.block.name,
+                    'value': child.block.get_prep_value(child.value),
+                    # assign a new ID on save if it didn't have one already
+                    'id': child.id or str(uuid.uuid4()),
```
chosak: It'd be nice if the `str(uuid.uuid4())` didn't need to be repeated. We could also have a clearer comment about why creation of this ID is desirable. What about something like:

```py
prep_value = []

for i, stream_data_item in enumerate(self.stream_data):
    if self.is_lazy and i not in self._bound_blocks:
        # This child has not been accessed as a bound block, so its raw JSONish
        # value (stream_data_item here) is still valid
        prep_value_item = stream_data_item
    else:
        child = self[i]
        prep_value_item = {
            'type': child.block.name,
            'value': child.block.get_prep_value(child.value),
            'id': child.id,
        }

    # As this method is preparing this value to be saved to the database,
    # this is an appropriate place to ensure that each block has a unique id.
    if not prep_value_item.get('id'):
        prep_value_item['id'] = str(uuid.uuid4())

    prep_value.append(prep_value_item)

...
```

loicteixeira:
Changes looks good. However it might be worth adding a test for the scenario you describe in the [comments of the previous PR](https://github.com/wagtail/wagtail/pull/4727#discussion_r223045844) as we would have missed this scenario if it wasn't for you jumping in @gasman.

chosak:
@loicteixeira great idea - I've added a couple of unit tests for this on my branch off of @gasman's [here](https://github.com/wagtail/wagtail/compare/master...chosak:fix/lazy-streamvalue-get-prep-value2#diff-9bbd817943c4bd49f80b9e42c727503eR2688).

gasman:
Merged (with @chosak's additions) in 2fe866e4849eb36b976d72c1ddee1b08ff39721b

PR #4803, opened 2018-10-09
Flexible parameters to view classes for ModelAdmin

ladrua:
Removal of hardcoded parameters to view classes and allowing additional parameters if needed.


gasman:
Thanks @ladrua! Please could you add an example of this in use to the test app (https://github.com/wagtail/wagtail/tree/master/wagtail/tests/modeladmintest)? That way we'll avoid breaking this functionality if it gets refactored in future.

ladrua:
I guess we need to create a new model and test? If one defines extra get parameters for CreateView the url naturaly changes for the test.

gasman:
@ladrua Yes please - I think a new model and test is the way to go here.

ladrua:
I've come to the realization that this does not add any practical value as one can use norma get params in cases where its needed.

Issue #4802, opened 2018-10-07
Highlight broken links to pages and documents in rich text


thibaudcolas:
### Issue Summary

As of #4795 and #4800, it is possible for rich text to contain links to pages and documents that were deleted. The rich text processing has access to the information to determine that those links point to missing things. It would be nice to surface this information to the end user so they realise the links are broken, and can hopefully remove them or replace them with links that work.

Here is what those links look like at the moment – just like any other link, except for the fact that the tooltip doesn't show any URL like for normal links:

![broken-links-wagtail](https://user-images.githubusercontent.com/877585/46586180-4a3f7780-ca83-11e8-9d5b-4aa5d2c68aa3.png)

### Steps to Reproduce

1. Enter content in a rich text field, add a link to a page of the CMS, then save.
2. Remove the page the link points to.
3. Open the rich text content to see the broken link.

---

Some potential solutions for this could be to:

- Display the links with an "error/warning" color, instead of the default teal
- Display the links with an icon that conveys they are different from normal page/document links, and might need special attention
- Add a label in the tooltip that makes it clear those links are broken ("Broken link", or "Missing document" or "Missing page")

thibaudcolas:
This would be done in the React components for links and documents, around:

https://github.com/wagtail/wagtail/blob/7cd73c472f0ad0adfe843c0eeff2d601df0198f2/client/src/components/Draftail/decorators/Document.js#L10-L22

https://github.com/wagtail/wagtail/blob/7cd73c472f0ad0adfe843c0eeff2d601df0198f2/client/src/components/Draftail/decorators/Link.js#L14-L36

and:

https://github.com/wagtail/wagtail/blob/7cd73c472f0ad0adfe843c0eeff2d601df0198f2/client/src/components/Draftail/decorators/TooltipEntity.js#L104-L114

For the potential "broken link" / "missing x" label, this would need to happen here: https://github.com/wagtail/wagtail/blob/7cd73c472f0ad0adfe843c0eeff2d601df0198f2/wagtail/admin/templates/wagtailadmin/admin_base.html#L30-L52 and also  in https://github.com/wagtail/wagtail/blob/master/client/tests/stubs.js.

PR #4813, opened 2018-10-16
Highlight broken links to pages and documents in rich text

bmoe872:
# What this does
Addresses #4802 - Adds highlighting to missing pages or documents in the WYSIWYG editor.

# Screenshots
screen shot 2018-10-16 at 8 29 24 am
screen shot 2018-10-16 at 8 29 34 am

# Additional Notes
Running tests through NPM, I was getting:
`SecurityError: localStorage is not available for opaque origins`

Which seems to be solved by adding:
`"testURL": "http://localhost:8000"` to the config for jest (I didn't commit this, because again I'm not sure why its not included already, or if its my setup, so I didn't want to introduce something that didn't make sense to someone else.) - it looks like I'm not passing the tests due to the snapshot being different with the use of the new icon, and since I'm not super familiar with Jest, I wasn't sure if it was kosher to go ahead and update the snapshot. Let me know if it is, and I can go ahead and update that!

### DIFF
--- a/client/src/components/Draftail/decorators/Document.js
+++ b/client/src/components/Draftail/decorators/Document.js
@@ -5,18 +5,32 @@ import Icon from '../../Icon/Icon';
 
 import TooltipEntity from '../decorators/TooltipEntity';
 
+import { STRINGS } from '../../../config/wagtailConfig';
+
 const documentIcon = <Icon name="doc-full" />;
+const missingDocumentIcon = <Icon name="warning" />;
 
 const Document = props => {
   const { entityKey, contentState } = props;
   const data = contentState.getEntity(entityKey).getData();
+  const url = data.url || '';
+  let icon;
+  let label;
+
+  if (!url) {
+    icon = missingDocumentIcon;
+    label = STRINGS.MISSING_DOCUMENT;
+  } else {
+    icon = documentIcon;
+    label = data.filename || '';
+  }
 
   return (
     <TooltipEntity
       {...props}
-      icon={documentIcon}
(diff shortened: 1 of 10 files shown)


client/src/components/Draftail/decorators/Document.js
```diff
@@ -5,18 +5,32 @@ import Icon from '../../Icon/Icon';
...
 
+import { STRINGS } from '../../../config/wagtailConfig';
+
 const documentIcon = <Icon name="doc-full" />;
+const missingDocumentIcon = <Icon name="warning" />;
 
 const Document = props => {
   const { entityKey, contentState } = props;
   const data = contentState.getEntity(entityKey).getData();
+  const url = data.url || null;
```
thibaudcolas: I have changed the default value for `url` to `null`, so the "Missing document" label isn’t clickable. Same for links.

client/tests/stubs.js
```diff
@@ -41,6 +41,8 @@ global.wagtailConfig = {
     SHOW_LATEST_CONTENT: 'Show latest content',
     SHOW_ERROR: 'Show error',
     EDITOR_CRASH: 'The editor just crashed. Content has been reset to the last saved version.',
+    BROKEN_LINK: 'Broken link',
+    MISSING_DOCUMENT: 'Missing document',
```
thibaudcolas: I have changed this not to use Tilte Case – was `Missing Document` before.

thibaudcolas:
Aaaand it’s merged! This will be released in Wagtail 2.5.

Issue #4489, opened 2018-04-23
Merging type_q into one Q results in "TypeError: can't pickle dict_values objects"


tm-kn:
### Issue Summary

When trying to merge `type_q` queries into one Q I get an error saying.
```
TypeError: can't pickle dict_values objects
```

### Steps to Reproduce

```python
from django.db.models import Q

from wagtail.core.models import Page

from home.models import HomePage


type_q =Page.objects.type_q(HomePage)
query = Q()
query |= type_q
```

Probably can be solved converting `dict_values` to a `list`.
https://github.com/wagtail/wagtail/blob/1849f0d54a16432eda2471fda5a602a26d5a5740/wagtail/core/query.py#L175-L179

### Technical details

* Python version: 3.6.4
* Django version: 2.0.3
* Wagtail version: 2.0
* Browser version: n/a

nott:
I'm going to work on this during the sprint in Minsk

gasman:
@nott This has already been addressed by #4814 - it's just waiting for one of us to review it (and if you want to sanity-check that PR, that would be good too :-) )

nott:
@gasman great
i'll take a look

gasman:
Fixed in #4814

PR #4814, opened 2018-10-16
returns content_types as a list instead of dict_values

bmoe872:
# What does this do?
Addresses #4489 - simply changes the return from a `dict_value` to a list, by wrapping the return in a `list()`

### DIFF
--- a/wagtail/core/query.py
+++ b/wagtail/core/query.py
@@ -178,7 +178,7 @@ class PageQuerySet(SearchableQuerySetMixin, TreeQuerySet):
             if issubclass(model, klass)
         ]).values()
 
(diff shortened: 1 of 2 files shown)


gasman:
Thanks @bmoe872! Please can you add a test for this into https://github.com/wagtail/wagtail/blob/master/wagtail/core/tests/test_page_queryset.py , so that we can verify that this fixes the issue (and so that we'll be alerted if it gets broken again in future)?

bmoe872:
@gasman should be good to go, let me know if something seems amiss!

gasman:
Thanks! Merged in 95333ba8ec5c19a039fa0328bb855d8fa95f1eec.

PR #4821, opened 2018-10-18
SVG icons refactor + SVG icons for menu

Scotchester:
This code is an extension of @allcaps's #4464 that converts all icons in the main menu to SVG. All other icons in the admin are still loaded with the icon font.

Opening it as its own PR so I can see if the CI checks will pass. I've got the tests passing locally, but I want to ensure I didn't miss any edge cases.

This should not affect users of [wagtailfontawesome](https://gitlab.com/alexgleason/wagtailfontawesome) at all, which is included in the bakery demo and still working when I test it locally.

SVG and wagtailfontawesome living in harmony

### To do:
- [ ] Cross-browser testing
- [ ] Per CodeCov, looks like we need some _new_ tests
- [ ] Update docs
- [ ] Update changelog
- [ ] ??? It's my first contribution to Wagtail; what am I missing? 😄


tomdyson:
Thanks @Scotchester! I made a new label just for you :)

thibaudcolas:
Adding the discussion from #4821 for future reference:

-----

> https://github.com/wagtail/wagtail/pull/4464#issue-180843464

This is a POC to componentise icons and make them screen-reader friendly using SVG and templates.
Related issues: #4463, #2349, #1224 

Goals:
- Wagtail icon font to svg
- Insert icon via template tag
- Way to hook in extra icons (your app or third party)
- Way to override some icons (customisation)
- Way to override all icons (icon packs)
- Documentation
- See if SVG icons need to be in the base template or can live in a external file/view.

This POC:
- Convert Wagtail icon font to SVG symbols
- Each symbol its own template. (easy overrides)
- Add a hook to register new icons
- Add templatetag inserting icons (DRY)
- Add CSS to position icons
- Add templatetag to load icons into the base template
- Made small preperations for a view serving a icon sprite (as alternative to including all icons in the base template)

TODO:
- Figure out if we want all icons each request OR that they should be served via a view (and set cache headers).
- JS should also use the use-tag.
- Provide an alternative for the most popular wagtail package: wagtailfontawesome
- Add correct styling for drawer menu.
- Remove icon fonts and icon CSS.
- ...

Test drive:
```
INSTALLED_APPS += [
    'wagtail.customapp',
    'wagtail.contrib.styleguide',
]
```
- See settings => styleguide => icons. There are two sets. The icon font icons and the new SVG icons.
- The wagtail.customapp inserts a custom icon and overrides an existing one.
- The main menu and drawer have SVG-icons. (Some menu-items are replaced by JS and have the old icon font).

@thibaudcolas Ping! What do you think? Should we use a separate view for an icon sprite? OR is it okay to include all these icons in the base template?

---

> https://github.com/wagtail/wagtail/pull/4464#issuecomment-380575587

@allcaps I think this is super duper amazing! Thank you for getting this started. I don't have much time to review this right now unfortunately, but I'd love to.

From the perspective of "getting this merged", the thing that's the least likely to get in as-is is the `idify` hack. It's not reliable enough to auto-convert class names to an icon identifier, we'll have to painstackingly refactor all of the existing icon call sites to refer to icons by name only.

Sorry I can't get a more thoughtful review – here are other things that come to mind:

- Try to sync this work with https://github.com/wagtail/wagtail/pull/4381#pullrequestreview-110299442
- If you want to help with that thankless refactoring of all icon call sites, have a look at https://github.com/wagtail/wagtail/issues/4199#issuecomment-379581195.
- Is it really useful to be able to override all icons?

allcaps:
Refs: https://www.deque.com/blog/creating-accessible-svgs/

wagtail/admin/menu.py
```diff
@@ -11,10 +11,11 @@
 class MenuItem(metaclass=MediaDefiningClass):
     template = 'wagtailadmin/shared/menu_item.html'
 
-    def __init__(self, label, url, name=None, classnames='', attrs=None, order=1000):
+    def __init__(self, label, url, name=None, classnames='', icon_name='', attrs=None, order=1000):
         self.label = label
         self.url = url
         self.classnames = classnames
+        self.icon_name = icon_name
```
thibaudcolas: Is `classnames` used for any other purpose than icons? If not we should remove it, and-or have a deprecation path then remove it.

wagtail/admin/templatetags/wagtailadmin_tags.py
```diff
@@ -392,3 +392,43 @@ def avatar_url(user, size=50):
...
+        'class_name': class_name,
+        'title': title,
+    }
+
+
+@register.inclusion_tag("wagtailadmin/shared/icons.html")
+def icons():
+    icon_hooks = hooks.get_hooks('register_icons')
+    icons = sorted(itertools.chain.from_iterable(hook([]) for hook in icon_hooks))
+    return {'icons': icons}
```
thibaudcolas: We already have an icon tag from `wagtailui_tags`, how should we go about syncing these?

wagtail/admin/templates/wagtailadmin/icons/arrow-down-big.svg
```diff
@@ -0,0 +1,4 @@
+<symbol id="arrow-down-big" viewBox="0 0 16 16">
+    <path fill="currentColor"
```
thibaudcolas: I think the fill would be better defined in CSS if possible.

client/src/components/Explorer/ExplorerItem.scss
```diff
@@ -82,6 +89,11 @@
     font-size: 1.2em;
 }
 
+.icon--item-action {
+    width: 1em;
+    height: 1em;
+}
```
thibaudcolas: This seems like a customisation of icons based on their context rather than a variation that will be reusable elsewhere, so I think it would be enough to target `.icon` within the context of this component.

client/src/components/Button/Button.js
```diff
@@ -63,10 +47,6 @@ const Button = ({
 Button.propTypes = {
   href: PropTypes.string,
   className: PropTypes.string,
-  icon: PropTypes.oneOfType([
-    PropTypes.string,
-    PropTypes.arrayOf(PropTypes.string),
-  ]),
```
thibaudcolas: I think it would still be valuable for the button to take care of rendering any icons within it, since it's quite common for buttons to have icons.

.gitignore
```diff
@@ -23,3 +23,11 @@ npm-debug.log*
...
 coverage/
 client/node_modules
+
+wagtail/contrib/wagtailstyleguide/
+wagtail/wagtailadmin/
+wagtail/wagtaildocs/
+wagtail/wagtailembeds/
+wagtail/wagtailimages/
+wagtail/wagtailsnippets/
+wagtail/wagtailusers/
```
thibaudcolas: Not sure I understand what these are for?

thibaudcolas:
Rebased onto latest master. I'll be looking into the icon’s implementation over the next few weeks, for accessibility purposes. I’ll most likely only be changing the icon font implementation to be screen-reader-friendly, not doing the switch to SVG.

zerolab:
Worth adding https://www.sarasoueidan.com/blog/accessible-icon-buttons/ to the reading mix. Hopefully there will be useful tips and tricks in that article that can be applied to this PR

thibaudcolas:
I ended up improving the current icon fonts implementation (#5359) instead of pursuing this SVG refactor. Got to experience some of the pain points of icon fonts, that show why this refactor is relevant:

- When fonts fail to load, users see a wall of "failed to render" Unicode characters
- Or they get unreliable characters from fallback fonts, https://www.zachleat.com/web/font-display-icon-fonts/ (thanks @zerolab)
- The icons were really hard to update or even just rename, because of the `.woff` binary that has to be re-generated in Icomoon.

It’s especially problematic with the current implementation that icon classes are applied on arbitrary elements (instead of having a dedicated `<span class="icon icon-potato"></span>`), as this makes it harder to provide alternative text for them.

wagtail/admin/templatetags/wagtailadmin_tags.py
willbarton: It looks implementationally similar to `wagtail_icon`. @Scotchester @allcaps is the intent here to replace that tag for the menus but keep it for everywhere it's used outside of menus for now?

Scotchester:
I've just rebased this PR on master.

Scotchester:
Force-pushed the most recent commit, amended to correct a mistake I made in rebasing.

wagtail/admin/templatetags/wagtailadmin_tags.py
Scotchester: Discussed in person: We'll keep the deletion of the `wagtailui_tags` module in this PR and update the two uses of its `wagtail_icon` icon font tag (both related to page table sorting) to use the new SVG icon tag. Done in e3060d4.

wagtail/admin/templates/wagtailadmin/icons/arrow-down-big.svg
Scotchester: This seems like a useful default to have icons render in the same color as whatever text context they find themselves in. For example, the menu icons' color matches its adjacent text no matter what state that text is in (normal, hover, focus, etc.), with no CSS to specify the icon's color at all.

client/src/components/Explorer/ExplorerItem.scss
Scotchester: I agree in principle, however, in [this submenu item template context](https://github.com/wagtail/wagtail/pull/4821/files#diff-ca00f5418069bc4ee30e2434d64a0506), there are two icons within one `a.submenu-trigger` element: the item's unique icon, and the right arrow.

```html
    <a href="#" data-nav-primary-submenu-trigger class="submenu-trigger {{ classnames }}"{{ attr_string }}>
        {% if icon_name %}{% icon icon_name 'icon--menuitem' %}{% endif %}
        {{ label }}
        {% icon 'arrow-right' 'icon--submenu-trigger' %}
    </a>
```

I suppose they could be targeted with `:first-child` and `:last-child`, but that feels a little magical or too obfuscated to me, whereas it's easy to search by those class names to connect the Less to the template code, or vice versa.

Thoughts?

wagtail/admin/menu.py
Scotchester: Two things:
1. In order to maintain compatibility for wagtailfontawesome, we'd have to do something like, if a name is passed to `menu_icon` that begins with `fa-` ([see example in bakerydemo](https://github.com/wagtail/bakerydemo/blob/99b65a5e24038e9eadcd8a3d5198efd827ab1952/bakerydemo/base/wagtail_hooks.py#L40)), then instead of using our `{% icon %}` tag, output the necessary classes. Not a huge deal, but… 
2. It is a documented API for customizing new menu items, generally: https://docs.wagtail.io/en/v2.5.1/reference/hooks.html?highlight=menuitem#register-admin-menu-item. Though it only mentions using it to add icons, I am sure there are people out there using it for more than that. We could create a deprecation path, but I'm not sure if we want to take away the ability to add custom classes to new menu items.

client/src/components/Button/Button.js
Scotchester: Discussed in person: No change to this PR. This is something for the UI refactoring team to explore further, as the current Button component is isolated to menu use cases, but we'll want a more generic button component for other parts of the UI.

wagtail/admin/templates/wagtailadmin/icons/arrow-down-big.svg
Scotchester: Done in dc12760.

client/src/components/Icon/Icon.js
```diff
@@ -2,12 +2,13 @@ import PropTypes from 'prop-types';
 import React from 'react';
 
 /**
- * Abstracts away the actual icon implementation (font icons, SVG icons, CSS sprite).
  * Provide a `title` as an accessible label intended for screen readers.
  */
 const Icon = ({ name, className, title }) => (
   <span>
-    <span className={`icon icon-${name} ${className || ''}`} aria-hidden></span>
+    <svg className={`${className || ''}`}>
```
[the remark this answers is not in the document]
Scotchester: Updated as part of dc12760.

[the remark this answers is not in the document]
Scotchester: Updated as part of dc12760.

client/src/components/Explorer/ExplorerToggle.js
```diff
@@ -5,18 +5,20 @@ import { connect } from 'react-redux';
...
 
 import Button from '../../components/Button/Button';
+import Icon from '../../components/Icon/Icon';
 
 /**
  * A Button which toggles the explorer.
  */
 const ExplorerToggle = ({ children, onToggle }) => (
   <Button
     className="submenu-trigger"
```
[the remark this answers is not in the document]
Scotchester: Removed in 3610539.

client/src/components/Button/__snapshots__/Button.test.js.snap
```diff
@@ -44,7 +44,7 @@ exports[`Button #dialogTrigger 1`] = `
 exports[`Button #icon 1`] = `
 <a
   aria-haspopup={null}
-  className=" icon icon-test-icon"
+  className=""
```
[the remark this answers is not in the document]
Scotchester: Removed in 95d0abba2947d6721a10ce364222274f9c9fca80.

client/src/components/Explorer/ExplorerHeader.js
```diff
@@ -19,7 +19,10 @@ const ExplorerHeader = ({ page, depth, onClick }) => {
       onClick={onClick}
     >
       <div className="c-explorer__header__inner">
-        <Icon name={isRoot ? 'home' : 'arrow-left'} />
+        <Icon
+          name={isRoot ? 'home' : 'arrow-left'}
+          className={'icon--explorer-header'}
```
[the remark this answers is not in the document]
Scotchester: Done in 6ee5470a0157e3794b64429ceea40af46d59260b.

Scotchester:
@thibaudcolas I've just noticed that the Draftail buttons use the `Icon` component, and are now quite broken 😆 

Screen Shot 2019-07-27 at 14 26 54

I could use your advice on how to approach solving that.

Scotchester:
> @thibaudcolas I've just noticed that the Draftail buttons use the `Icon` component, and are now quite broken 😆

Fixed in 7680354.

thibaudcolas:
Just a note for future reference that if we do go for an external SVG file we’ll need a polyfill for IE11 – https://github.com/jonathantneal/svg4everybody, or the simpler https://github.com/Keyamoon/svgxuse.

wagtail/admin/templates/wagtailadmin/shared/menu_item.html
```diff
@@ -1,3 +1,12 @@
+{% load wagtailadmin_tags %}
+
 <li class="menu-item{% if active %} menu-active{% endif %}">
-    <a href="{{ url }}" class="{{ classnames }}"{{ attr_string }}>{{ label }}</a>
+    <a href="{{ url }}"
+    {% if not icon_name %}
+       class="{{ classnames }}"
+    {% endif %}
```
thibaudcolas: I think we need to discuss the backwards-compatibility consequences of this.

wagtail/admin/templatetags/wagtailadmin_tags.py
```diff
@@ -473,3 +473,43 @@ def avatar_url(user, size=50):
...
+
+
+@register.filter
+def idify(val):
```
thibaudcolas: This looks like it is no longer used and can be removed.

wagtail/admin/views/icons.py
```diff
@@ -0,0 +1,8 @@
+from django.shortcuts import render
+from django.views.decorators.cache import cache_control
+
+
+@cache_control(max_age=3600, private=True)
+def sprite(request):
```
thibaudcolas: Is this actually used anywhere? I don’t think I’ve seen it.

allcaps: I dropped it.

wagtail/admin/templatetags/wagtailadmin_tags.py
allcaps: Dropped

wagtail/admin/templates/wagtailadmin/shared/menu_item.html
allcaps: Added.

allcaps:
@thibaudcolas and @Scotchester 

Let's get this one shipped! I've rebased this PR on master and did some clean-up.

I'll write documentation. Maybe we get better ideas about the backwards-compatibility and deprecation consequences once working with the new icons is clear.

wagtail/admin/templates/wagtailadmin/shared/icons.html
```diff
@@ -0,0 +1,11 @@
+{% spaceless %}
+<div style="display: none">
+  <svg xmlns="http://www.w3.org/2000/svg">
```
thibaudcolas, line 3: No need for `xmlns="http://www.w3.org/2000/svg"` in HTML.

wagtail/admin/templates/wagtailadmin/shared/icon.html
```diff
@@ -1,2 +1,11 @@
-<span class="{{ classnames }}" {% if decorative %}aria-hidden="true"{% else %}title="{{ title }}"{% endif %}></span>
-{% if show_label %}{{ title }}{% elif not decorative %}<span class="sr-only">{{ title }}</span>{% endif %}
+{% spaceless %}
+{# Reference implementation: components/Icon.js #}
+<svg class="icon icon-{{ name }} {{ class_name }}" aria-hidden="true">
```
thibaudcolas, line 3: We also need `focusable="false"` here for IE11 support.

thibaudcolas:
I’ve tagged as "Needs Design Decision" so we discuss in a core team meeting whether this can be merged early without documentation or not.

Scotchester:
@allcaps and @thibaudcolas I'm so sorry I've neglected this! I promise I will make time to review the latest updates and conversation tomorrow, before the core team meeting on Wednesday.

wagtail/admin/templates/wagtailadmin/shared/menu_item.html
Scotchester: > I think we need to discuss the backwards-compatibility consequences of this. Do we need to stop passing classnames if `icon_name` is set?

I think we should continue passing `classnames` for the reasons I mentioned in https://github.com/wagtail/wagtail/pull/4821/files#r307966908

Despite having written that, last summer, I had the code to fix this template to still accept `classnames` no matter what stashed and uncommitted. I will add it to the PR momentarily.

Scotchester:
> Yay @allcaps! I think this is really, really close. The only backwards-compatibility I have identified currently is with:
> 
> ```
> {% if not icon_name %}
>        class="{{ classnames }}"
> {% endif %}
> ```
> 
> If we remove that to keep `class="{{ classnames }}"` then I think that as this PR stands there shouldn’t be any other concerns. I think that `if not` might be something you added when this was based on the `idify` filter, and might not be relevant anymore.

Per my comment on the particular lines: Done!

Scotchester:
> I’ve picked up two small things while looking at this again – and one slightly bigger one: some of the `id`s for the new symbols would clash with existing `id`s used in Wagtail:
> 
> …
> 
> I’m sure we could easily rewrite that existing code to use different `id`s, but it might be easier to just prefix the `id` of all symbols with `icon-` to not have to worry about this as much?

Gooooood catch. I have a commit that could fix this for the given implementation, BUT it also ties in with the work I had in progress last  summer before it got dropped for months (which I apparently never commented on here, I think I may have just told you about it in person or over Slack, @thibaudcolas):

We can move the combined SVG markup for all the icons into an external SVG file that can be cached by browser and not bloat the markup of every page load in the admin. It would also conveniently sidestep this ID clashing problem.

The one downside is that it requires a JS polyfill for IE11 support. When I left off last summer, I recall having trouble getting the polyfill to work as expected, and in re-testing it tonight, the polyfill I was trying out ([svgxuse](https://github.com/Keyamoon/svgxuse)) works well for IE 11, but seems overly aggressive in activating itself, even in modern browsers that support external `use` references. There are some more recently-updated alternatives to consider these days, such as [SVG for Everybody](https://github.com/jonathantneal/svg4everybody) and [External SVG Polyfill](https://www.npmjs.com/package/@thasmo/external-svg-polyfill).

If we want to take a little more time to test one or both of those and see if they're more reliable, we could avoid that `id` collision issue and use the less repetitive values for the `icon_name` property, but it would also be fine to fix the ID collisions with your reasonable suggestion of prefixing with `icon-`, go ahead with this PR, and then move the SVG code to its own file in a separate PR.

Let me know what you think.

allcaps:
> The one downside is that it requires a JS polyfill for IE11 support.

I tired an external sprite but then the icons load after all other content is loaded. And this gives a nasty flash each page load.

An external sprite is a nice to have (and 90% implemented see deleted commits), but I'd prefer to keep it out of scope for now. Just get this merged.

Scotchester:
> I tired an external sprite but then the icons load after all other content is loaded. And this gives a nasty flash each page load.
> 
> An external sprite is a nice to have (and 90% implemented see deleted commits), but I'd prefer to keep it out of scope for now. Just get this merged.

I think that big delayed flash may have been due to the polyfill activating unnecessarily, if you were trying svgxuse like I was, but I agree that it's fine to keep further investigation out of this PR and move forward.

Scotchester:
> > I’ve picked up two small things while looking at this again – and one slightly bigger one: some of the `id`s for the new symbols would clash with existing `id`s used in Wagtail:
> > …
> > I’m sure we could easily rewrite that existing code to use different `id`s, but it might be easier to just prefix the `id` of all symbols with `icon-` to not have to worry about this as much?
> 
> Gooooood catch. I have a commit that could fix this for the given implementation

Pushed this in https://github.com/wagtail/wagtail/pull/4821/commits/0391fe06468a149bc529084339149a55a33cde9f

Looks like it caused some test failures. I'll try to sort those out.

wagtail/admin/templates/wagtailadmin/skeleton.html
```diff
@@ -1,5 +1,5 @@
 <!doctype html>
-{% load wagtailadmin_tags i18n %}
+{% load wagtailadmin_tags i18n static %}
```
allcaps, line 2: I don't think we have to add static here.

allcaps:
Wrap up
--------

This PR add SVG icons to the Wagtail admin interface.

The new implementation addresses accessibility concerns. Adds title attribute and prevents problems with font rendering and screen readers.

This PR
--------
- All icon font icons converted to SVG symbols
- All symbols added to `skeleton.html` to make them available in the ui.
- Single icon html template `icon.html`
- Single icon js component `Icon.js`
- `register_icons` hook to manage icons (add, overide) `wagtail/admin/wagtail_hooks.py`
- All icons in de admin interface use SVG
- Icon fonts are still compatible. See for example bakery demo, it has both Wagtail SVG and wagtail-fontawesome.

Out of scope
-------------
- Documentation how to add custom icons (keep it an internal feature for now).
- Drop icon fonts. We stay backward compatible.
- External sprite view to reduce page load. Needs poly fill to accommodate IE11. Is the next thing to tackle.

wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html
```diff
@@ -851,6 +851,111 @@ <h2>Icons</h2>
 
         </section>
 
+        <section id="icons" class="icons">
```
thibaudcolas, line 854: Not going to make much of a difference but this is another duplicate id 😅

wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html
```diff
@@ -851,6 +851,111 @@ <h2>Icons</h2>
...
+                    transform: translateY(0.6em) rotate(0deg);
+                } to {
+                    transform: translateY(0.6em) rotate(360deg);
+                }
+            }
+
+          </style>
+
+            <ul class="unlist">
+                <li>{% icon 'wagtail-icon' %} wagtail</li> {# Note: There is 'wagtail' id on the body named 'wagtail'. Should prefix (namespace) all icons? #}
```
thibaudcolas, line 875: Ah, that note was a good catch, but it’s now outdated.

wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html
```diff
@@ -851,6 +851,111 @@ <h2>Icons</h2>
...
+                <li>{% icon 'password' %} password</li>
+                <li>{% icon 'download' %} download</li>
+                <li>{% icon 'order' %} order</li>
+                <li>{% icon 'grip' %} grip</li>
+                <li>{% icon 'home' %} home</li>
+                <li>{% icon 'order-down' %} order-down</li>
+                <li>{% icon 'order-up' %} order-up</li>
+                <li>{% icon 'bin' %} bin</li>
+
+                {# Append a class name: Add both the original classname `svg-icon` and the custom classname `spinning` #}
```
thibaudcolas, line 929: This comment looks outdated?

wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html
```diff
@@ -851,6 +851,111 @@ <h2>Icons</h2>
...
+                <li>{% icon 'pilcrow' %} pilcrow</li>
+                <li>{% icon 'title' %} title</li>
+                <li>{% icon 'code' %} code</li>
+                <li>{% icon 'openquote' %} openquote</li>
+                <li>{% icon 'chain-broken' %}  chain-broken</li>
+                <li>{% icon 'table' %} table</li>
+                <li>{% icon 'logout' %} logout</li>
+
+                {# Redundant icon? #}
+                <li>{% icon 'uni52' %} uni52</li>
```
thibaudcolas, line 955: Let’s review the icons in a future PR.

wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html
Scotchester: Hmm… should we remove the above icons section with the icon font output? I'm not sure it serves a purpose to have both here in the style guide, since the main point is just to show the icons with their names.

wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html
Scotchester: Hah, yes! Will remove it.

wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html
thibaudcolas: Let’s see what @allcaps thinks, I’m happy either way.

allcaps: I think the style guide should showcase best practices. So SVG icons. Can we change this id?

wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html
allcaps: Yes remove please

wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html
allcaps: Oops, remember what I've done here.

It was never the intention to keep the icon font list. 
I just displayed both icon font and SVG icons together to compare the two.

I'd like to move forward and drop icon fonts in the future. Therefore we should show SVG only.

wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html
allcaps: Removed

wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html
allcaps: Removed

wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html
allcaps: I removed the font icon showcase from the style guide.

lb-:
@thibaudcolas / @allcaps - is this an appropriate commit message? I am attempting to sum everything up.

```
Add SVG Icons and include in admin menu
* This commit only covers the migration of the icons in the menu and draftail editor. All
other admin menu icons remain implemented as before.
* Existing font based icon support still exists.
* Convert menu icons to SVG
* This could be considered a breaking change for any Wagtail sites or
packages that are leveraging Wagtail's Button and Icon React components.
* Remove references to unused `submenu-trigger` class
* Fix size of icons in Draftail toolbar
* Fix size of link icons in Draftail editor
* Restore output of classnames on menu items at all times
* Prefix icon `id`s with `icon-` to prevent `id` collisions
* Add hooks support (not yet documented) for adding custom icons
```

allcaps:
@lb- Thank you for the review.

> * Fix size of icons in Draftail toolbar
> * Fix size of link icons in Draftail editor

Can be removed. The SVG icons  are visually the same as before. So no real change here.

> * Restore output of classnames on menu items at all times

Can be removed. Classnames where dropped and brought it back in. So no real change here.

lb-:
Merged in via eed16e00348c6ddfe8f997e1150006a3211a1216
Thanks for the amazing work on this @Scotchester @allcaps  & @thibaudcolas 

Note: I have not closed any of the related issues mentioned here https://github.com/wagtail/wagtail/pull/4821#issuecomment-433686197 as it appears they are not fully resolved as of this initial work.

PR #4824, opened 2018-10-19
language provided more cover with support Arabic

alfuhigi:


gasman:
Thanks @alfuhigi! However, the Arabic translation is currently only about 60% complete. The default list of languages only includes ones with over 90% coverage - this way we're not giving users a negative first impression by offering them incomplete translations. I've now updated the documentation to clarify this: c2a351fee1e168e130d1b3611eaec5df5c2d83dc

We'll be happy to add Arabic to the default list once it reaches 90% completion.

alfuhigi:
Thanks @gasman!,
I will happy if you add Arabic to your default list,I've now complete translation and it reach above 90% completion.

gasman:
Many thanks for your work on this! I added Arabic to the default list in the latest language update this weekend (https://github.com/wagtail/wagtail/commit/e2acd014b65e23ed6f6bf5cf95c4e4b7f8cfb799#diff-6c57a6061b56d31a714471f16352c60b), so it will be included in the next release.

PR #4826, opened 2018-10-19
Fixes docs reference to the wagtailsearch section which now results in a dead-end

jjanssen:
I was looking up some configuration settings within the documentation. Right now on the `Configuring Django for Wagtail` page the link to the search section within [Wagtail Apps](https://docs.wagtail.io/en/latest/advanced_topics/settings.html#wagtail-apps) results in a dead-end. 

image

This PR fixes this by linking to Usage Guide > Search


gasman:
Thanks @jjanssen!

PR #4832, opened 2018-10-21
create zh_Hans symlink, Fix #3668

### LINKED ISSUE #3668: Language code is not correct for Chinese

yihuang:


wagtail/api/v2/endpoints.py
```diff
@@ -294,10 +295,7 @@ def get_serializer_class(self):
         request = self.request
 
         # Get model
-        if self.action == 'listing_view':
-            model = self.get_queryset().model
-        else:
-            model = type(self.get_object())
+        model = self.get_queryset().model
```
yihuang: We can't call ``get_object`` when generates API schema. And I think there is no need to call ``get_object`` here anyway.

yihuang: But I think the model of ``get_object`` should be the same with model of queryset, isn't it?

yihuang:
I've created separate pull requests https://github.com/wagtail/wagtail/pull/4921 and https://github.com/wagtail/wagtail/pull/4922.

gasman:
Migrating to the new names `zh_Hans` and `zh_Hant` was completed in d5bc3c92d22672f06f2aeab1e1731ca4f4fcb383.

PR #4844, opened 2018-10-25
Add `rel="noopener noreferrer"` to target blank links

anselmbradford:
- Adds `rel="noopener noreferrer"` to target blank links.


jjanssen:
Merged with 303ee0f

Thank you for your work on this! I took the moment to revert the changes to the `.po` files as these files are generated during our release cycle as we pull the updated files from Transifex.

PR #4845, opened 2018-10-25
Users with blank preferred_language should use activated language (from LocaleMiddleware)

benjaoming:
Supersedes #4837

....

Hi there!

As usual, thanks for the wonderful project! :)

The behavior of most Django locale middleware is this: Activate a language according to some HTTP header, cookie, session or URL path element. Then fallback to `settings.LANGUAGE_CODE` if all this fails.

So this is the bug, that I could/should also have opened an issue for: Let's say I'm on `/en/wagtail` but my default language is `zh_CN`. In this case, all users without a preferred language on their profile will be shown the Chinese version, regardless that they navigate to `/en`.

This is very problematic because I use a manual URL navigation workflow `/<language>/wagtail` to translate using django-modeltranslation, and I have no way of switching anymore once a user has a profile object created.

Suggesting this to target `1.13.x` as it would be nice to have the support here. Am still stuck with a lot of things in Django 1.11, and it's hard to make the migration to Django 2 still. I think other users will be in the same situation.

After this change, a user can just navigate to their Language Preferences and set '---------' as preferred language. That will make them use whatever language is specified in `/XX/wagtail/`.

I tested the change locally and manually and it works. I also looked at the way that `user_profile.get_preferred_language()` and `@activate_lang` were used around the codebase, and I can't see anywhere that it would be expected to force `settings.LANGUAGE_CODE` rather than just falling back to the currently active language.

Thanks!

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -38,6 +38,11 @@ def pytest_configure(config):
     os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'wagtail.tests.settings')
     django.setup()
 
+    # Activate a language: This affects HTTP header HTTP_ACCEPT_LANGUAGE sent by
+    # the Django test client.
+    from django.utils import translation
+    translation.activate("en")
+
     from wagtail.tests.settings import MEDIA_ROOT, STATIC_ROOT
     shutil.rmtree(STATIC_ROOT, ignore_errors=True)
     shutil.rmtree(MEDIA_ROOT, ignore_errors=True)
--- a/wagtail/admin/tests/test_account_management.py
+++ b/wagtail/admin/tests/test_account_management.py
@@ -397,6 +397,9 @@ class TestAccountSection(TestCase, WagtailTestUtils):
         # Check that the language preferences are stored
         self.assertEqual(profile.preferred_language, '')
 
+        # Check that the current language is assumed as English
+        self.assertEqual(profile.get_preferred_language(), "en")
+
     @override_settings(WAGTAILADMIN_PERMITTED_LANGUAGES=[('en', 'English'), ('es', 'Spanish')])
     def test_available_admin_languages_with_permitted_languages(self):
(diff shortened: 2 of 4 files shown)


benjaoming:
It seems lots of tests are failing because the language has changed or is left non-deliberate as Spanish in the tests.

Example:

```
AssertionError: False is not true : The field 'title' on form 'form' in context 1 does not contain the error 'This field is required.' (actual errors: ['Este campo es necesario.'])
```

benjaoming:
Sorry if it looks like I'm just goofing around, but I was reproducing tests in Spanish locally, and after the change to `conftest.py` in 2b86f22, they started behaving in English.

benjaoming:
All changes squashed to a single commit... I know of nothing further to add :+1:

chimeno:
I guess it was because at that time the ``LocaleMiddleware`` was not activated by default so we didn't have that ``get_language`` and we needed a default.
In fact looking fast to [bakery demo](https://github.com/wagtail/bakerydemo/blob/428092df6d4b764d3769952ef2a6b4949767bae2/bakerydemo/settings/base.py#L70) there is no ``LocaleMiddleware`` by default.

I would also say that there is a need to announce that the  ``LocaleMiddleware``  is mandatory (it's not activated in the ``project_template`` neither).

benjaoming:
I am already using a monkeypatch of `get_preferred_language` in production and happy with the results :smile:

kaedroho:
@chimeno The ``LocaleMiddleware`` isn't enabled by default. But this change wouldn't make it mandatory, it would only make Wagtail recognise it when it is enabled.

gasman:
> @chimeno @gasman was there a reason to default to settings.LANGUAGE_CODE rather than let the middleware choose?

No, I think that was a mistake - I proposed using `LANGUAGE_CODE` as a fallback, when it would have been more accurate to say "use whatever Django's native logic picks".  Happy to go ahead with this 👍

Issue #4690, opened 2018-07-13
Have python3 installed but wagtail won't install


officerebel:
Inside virtualenv 
Python2.7 and python 3.6 installed.
$ python3 -m pip install wagtail

$ wagtail python3  start mysite
This version of Wagtail requires Python 3.4 or above - you are running 2.7

Meteor0id:
did you make that virtual environment with pip3 ? I have no idea what I am talking about, but I remember there was a seperate pip3 command for python 3 in some cases. Please look into that.

mikedingjan:
Hi @officerebel ,

Can you try running
```
python3 -m wagtail start mysite
```

officerebel:
Yes I have virtualenvironment and pip an pip3 installed wagtail

$ python3 -m wagtail start mysite

results in:
/Library/Frameworks/Python.framework/Versions/3.6/bin/python3: No module named wagtail.__main__; 'wagtail' is a package and cannot be directly executed

$wagtail start mysite
results in:
This version of Wagtail requires Python 3.4 or above - you are running 2.7

NDevox:
@officerebel sounds like you've installed wagtail as a python 2.7 package outside of your virtualenv first and that's overriding your current env. Remove it from the 2.7 one, wipe the virtualenv and try in a new one if you can.

jjanssen:
Hi @officerebel  - please use the [Wagtail tag on Stack Overflow](https://stackoverflow.com/questions/tagged/wagtail) for support queries.

alexrayu:
Installing the Wagtail as per instructions results in a non-working setup. When asking for help, users are redirected to StackOverflow instead.

gasman:
Hi @alexrayu,
It simply isn't possible for the instructions to cover every situation that might occur on every platform, such as selecting the correct Python version when multiple versions are installed, as in the case here. This is not a Wagtail-specific issue, and so it's appropriate to ask on Stack Overflow, where there's a larger pool of people available to help, and better tools for gathering the necessary information.

tomdyson:
@alexrayu thanks for taking the time to respond. We appreciate your perspective on this. FWIW, this wouldn't have helped the original issue creator, who wasn't following the instructions and had already installed Wagtail outside the virtualenv. On the other hand, we certainly want to make sure that our standard instructions work in default environments. Would you be happy to submit a PR with your suggestions? I have two points of feedback:

[...]

PR #4849, opened 2018-10-27
Clarify virtualenv steps for easier getting started

zerolab:
We recommend using Virtualenv, however newcomers seem to be stumped about this. Tweaked the docs a bit.

Reference: https://github.com/wagtail/wagtail/issues/4690#issuecomment-431240164

### DIFF
--- a/docs/getting_started/index.rst
+++ b/docs/getting_started/index.rst
@@ -10,11 +10,18 @@ We'd also recommend Virtualenv, which provides isolated Python environments:
 
  * `Virtualenv <https://virtualenv.pypa.io/en/latest/installation.html>`_
 
-
 .. important::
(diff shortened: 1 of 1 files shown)


m1kola:
Rebased on top of the `master` and merged in 3bb69841321aaa7bc5fa1468c290be8956abdd7c

gasman:
Thanks @zerolab!

I'm a little concerned that simply saying "run these lines" without explaining what they're doing might be a bit confusing, as it's something that users might want to adapt depending on their setup - for example, if they want to use a more meaningful name than `env`, or place all their virtualenvs in a central location (as virtualenvwrapper does - although if they're using virtualenvwrapper, the instructions will be different again). It's also not clear that the `activate` line needs to be re-run if you're picking up the project in a later session...

However, it's hard to say what the ideal way is of dealing with this without providing a complete guide to virtualenvs, and this is probably better than the old approach of "treat it as out of scope for the Wagtail docs and assume the user knows what they're doing already", so happy for this to be merged 👍

m1kola:
We give a link to the Virtualenv docs above, hopefully this solves the problem of explaining how Virtualenv works :)

PR #4854, opened 2018-10-27
Added overview of collections in the Image/Document chooser templates

jjanssen:
Follow up from #4383 which was missing the check on whether an end-user is actually using collections, so we could get this merged. It initially refers to issue #4162


gasman:
Merged in cea4cbc836d4cc8cad5b8e092af15838887d5b8b + parents.

Issue #2364, opened 2016-03-16
Paginating through image chooser modal offers no user feedback


davecranwell:
The Image Chooser modal lists its images paginated 12 at a time, and clicking the "Next" link loads the next page with ajax. However if the image thumbnails are yet to be generated the ajax response takes several seconds (roughly 10 on my development machine) as it waits for the images to be generated, without any feedback to the user about what's happening.

We should add spinners to all these ajax events anyway, however if there were some way of serving the HTML even if the images aren't rendered, that might also be nice.

tmsndrs:
Can anyone provide a screen recording of this so we can properly understand the problem? 

Thanks,

thibaudcolas:
![wagtail-image-chooser](https://user-images.githubusercontent.com/877585/33498500-365c2724-d6da-11e7-9e81-048791a825bf.gif)

The problems are:

- Initial load of the chooser takes forever if the thumbnails are not pre-generated.
- While loading, it is still possible to click on the button that triggers the chooser, and this will cause the chooser to open multiple time (stacked on top of one another).
- There is no loading feedback on the first load.
- No loading feedback either when clicking the pagination inside the chooser.

lb-:
See also #4006 for some detail on initial modal loading experience, it is easy to click ‘chose a ...’ multiple times on a slow connection due to no feedback that the button has been clicked / modal is loading.

thibaudcolas:
Turns out the fix for this could be rather simple, changing the modal to be in `show` mode before the content is loaded instead of `hide`:

https://github.com/wagtail/wagtail/blob/9f65e7001f17e922db32628b43028414351fe05f/wagtail/admin/static_src/wagtailadmin/js/modal-workflow.js#L22-L25

Here is what then happens when clicking on a chooser button, before the modal is fully loaded:

<img width="825" alt="modal-partial-loading" src="https://user-images.githubusercontent.com/877585/35156274-09848a0e-fd39-11e7-95d3-3a7b0f9eb57b.png">

The fix would be perfect by simply adding a loading indicator within this partially-loaded modal, which would then be replaced by the actual content.

I'd like to fix this properly so won't do this right away but still assigning to myself. If someone wants to tackle it, go for it.

PR #4856, opened 2018-10-27
2461 dynamic image url redux

zerolab:
This continues #2461 with a a few tweaks:

- rebased on and updated for latest master
- added regular template tag tests
- made the serving view an optional parameter
- updated docs

### DIFF
--- a/wagtail/images/templatetags/wagtailimages_tags.py
+++ b/wagtail/images/templatetags/wagtailimages_tags.py
@@ -113,5 +113,5 @@ class ImageNode(template.Node):
 
 
 @register.simple_tag()
-def image_url(viewname, image, filter_spec, key=None):
-    return generate_image_url(image, filter_spec, viewname, key)
+def image_url(image, filter_spec, viewname='wagtailimages_serve'):
+    return generate_image_url(image, filter_spec, viewname)
--- a/wagtail/images/templatetags/wagtailimages_tags.py
+++ b/wagtail/images/templatetags/wagtailimages_tags.py
@@ -5,7 +5,7 @@ from django.utils.functional import cached_property
 
 from wagtail.images.models import Filter
 from wagtail.images.shortcuts import get_rendition_or_not_found
-from wagtail.wagtailimages.views.serve import (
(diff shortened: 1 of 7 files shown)


wagtail/images/templatetags/wagtailimages_tags.py
```diff
@@ -107,3 +110,14 @@ def render(self, context):
...
                 resolved_attrs[key] = self.attrs[key].resolve(context)
             return rendition.img_tag(resolved_attrs)
+
+
+@register.simple_tag()
+def image_url(image, filter_spec, viewname='wagtailimages_serve'):
+    try:
+        return generate_image_url(image, filter_spec, viewname)
+    except NoReverseMatch:
+        raise template.TemplateSyntaxError(
```
m1kola: Hm. I'm not sure if `TemplateSyntaxError` is a good option. Syntax of the template is correct, but the configuration is not.

Maybe we should use Django's `ImproperlyConfigured`? We use it widely in Wagtail.

Other than this, looks good

wagtail/images/templatetags/wagtailimages_tags.py
```diff
@@ -107,3 +111,14 @@ def render(self, context):
...
+
+
+@register.simple_tag()
+def image_url(image, filter_spec, viewname='wagtailimages_serve'):
+    try:
+        return generate_image_url(image, filter_spec, viewname)
+    except NoReverseMatch:
+        raise ImproperlyConfigured(
+            "'image_url' tag requires the " + viewname + " view to be configured. Please see "
+            "http://docs.wagtail.io/en/latest/advanced_topics/images/image_serve_view.html#setup for instructions."
```
gasman: I'd suggest linking to the stable version of the docs here:
```suggestion
            "https://docs.wagtail.io/en/stable/advanced_topics/images/image_serve_view.html#setup for instructions."
```

zerolab: Done

jjanssen:
Possible next step for implementation #2364

PR #4858, opened 2018-10-27
Fix tests that check storage.exists on missing files

gasman:
Several tests that check for the deletion of underlying files of FileField were implemented incorrectly: the filename of a FileField is cleared on deletion, so they were testing the presence of None as a filename rather than testing the filename we're interested in.

Up until https://github.com/django/django/commit/bb81c22d90e5eb168544670ead1a13aa9695fed5 , None was being cast to a string and thus the assertions were always succeeding (assuming no file named "None" existed); the removal of this cast has caused the error to surface in Django 2.2a0.


zerolab:
Nice catch!

m1kola:
Rebased and merged in a5416d8604784a722f784e235a570ea3fdc0e016. Thanks!

Issue #2797, opened 2016-06-29
Enforce JS style guide (again)


gasman:
As reported in https://groups.google.com/d/msg/wagtail-developers/MjNuJnhXJv4/cPZX5LhnBgAJ - it appears that JS linting was broken when #2275 was merged (`npm run lint:js` now only checks the `client` folder). We need to:
- fix the task to run across the whole codebase
- add it to the `make lint` task https://github.com/torchbox/wagtail/blob/master/Makefile#L18-L20
- fix the inevitable style violations that have crept in since #1237
- add it to our CI setup

thibaudcolas:
I tried this very briefly yesterday, works fine but it raises upwards of 5000 linting issues.

Waiting for the next release of [Prettier](https://prettier.io) (due sometime over the next few weeks, https://github.com/prettier/prettier/issues/2589), which has interesting config options, and could allow us to reformat all of the files in a sustainable way.

lb-:
@thibaudcolas do you know of any progress on this one since you looked at this in 2017?

Prettier and Eslint have both had substantial updates in the last few years, using them together and help a lot. Eslint's fix option in conjunction with prettier can make a lot of minor linting issues automated.

I'm up to help with this if needed, I have implemented pretter/eslint configs on a few projects.

If we want to look in to pre-commit hooks, Husky seems to be a good option.

thibaudcolas:
Nope, there's been no progress that I'm aware of. One of the problems to solve here is that files in `client` are transpiled with Babel, and bundled with Webpack, while the rest of the JS isn't. This means we'd need separate ESLint/Prettier configs, so we don't treat non-transpiled code as if it was  ES2015+.

I don't think this is particularly blocking though – main reason why this hasn't happened is that it's not particularly appealing to fix Wagtail's old JS.

lb-:
@thibaudcolas - makes sense.

Hopefully this will become a bit easier when all code is moved to the webpack build process (client folder)

cc @jonnyscholes

thibaudcolas:
See #5252. I think once that's done this JS linting enforcement issue will be 99% resolved.

jonnyscholes:
Yep I'm linting the new JS as it goes in so in theory once that's all done all our JS will be consistent 😁

Although we might have to make some concessions for some jQuery plugins that whilst I'm hoping to remove, might have to stay around for a bit longer. Trying to avoid that though!

lb-:
[...]

PR #4859, opened 2018-10-27
 Use a nicer documents chooser help message if no documents exist

nott:
@m1kola reported in #3838 that the document chooser shows the following message by default when the side doesn't have any documents:

> Sorry, no documents match ""

image

@gmmoraes tried to fix this issue in #3845. He got some feedback in the PR: they suggested to add "add one document" button.

This PR implements those ideas, add tests and a JS helper function that switches to "UPLOAD" tab if available.


wagtail/documents/static_src/wagtaildocs/js/document-chooser-modal.js
```diff
@@ -11,6 +11,18 @@ DOCUMENT_CHOOSER_MODAL_ONLOAD_HANDLERS = {
                 setPage(page);
                 return false;
             });
+
+            $('a.upload-one-now').on('click', function(e) {
+                // Set current collection ID at upload form tab
+                let collection_id = $('#collection_chooser_collection_id').val();
```
davojta: looks like in wagtain js code you need to user camelCase for var names
@thibaudcolas @jjanssen  what do you think ?

jjanssen: Honestly it wouldn't surprise me if there would be a mix & match of all sorts of notations within our current javascript definitions. As our initial javascript linting accidentally got disabled some time ago (#2797). 

Looking at our more recent work regarding the integration of the Explorer menu and Draftail in React we enforce camelCasing in our variable names. But since this is a bit of legacy I'm ok with leaving this as is.

jjanssen:
Just gave it a testdrive in Chrome latest, Firefox ESR and Internet Explorer 11.
If the back-end is approved upon, I'm allright with merging this.

PR #4860, opened 2018-10-27
Define proper breakpoint definitions rather then typing @media 

jjanssen:
I decided to implement a proper mixin for mediaquery handling (based upon the thoughts of the fine people behind bootstrap). I think this would be a nice first step in the (hopefully) short-term fixing the mixed breakpoint usage.

In the morning I will perform a diff on the generated CSS, because essentially nothing should have been changed in the output 👍 

- [x] Last check on CSS output in a file-diff


jjanssen:
I did a final check on the before and after changes and I have 4 selectors with an alternate definition. Which are the breakpoints for targeting the smallest viewport and actually solved some overlapping a quirky behaviour 🎉 

**Change 1:**
image
**Change 2:**
image
**Change 3:**
image
**Change 4:**
image

The main difference in this breakpoint definition is that it no longer overlaps.
Previously the breakpoint mechanism always targeted going up and down with the same definition. 

In a nutshell the following overlaps:
```
@media (max-width:50em) {}  // 800px
@media (min-width:50em) {}  // 800px
```
In the current changes this will result in the following:
```
@media (max-width:49.9375em) {}  // 799px
@media (min-width:50em) {}  // 800px
```

These actually solves the following visual quirks:
**801px**
image
**800px**
Current quirk:
image
With the changes 🎉:
image
**799px**
image

 
￼
￼
￼
￼

jjanssen:
Different quirk with tabs is fixed as well 😄 Both on 800px.

**Before**
image

**After**
image

client/scss/_tools.breakpoints.scss
```diff
@@ -1,5 +1,5 @@
-$breakpoint-small: $breakpoint-mobile - 0.0625em;
-$breakpoint-medium: $breakpoint-mobile;
+$breakpoint-small: breakpoint-max(xs);
+$breakpoint-medium: breakpoint-min(sm);
 
 @mixin small {
```
jjanssen: An after-thought maybe we should replace these mixins with the new introduced mixins. What are you thoughts on this?

wagtail/admin/static_src/wagtailadmin/scss/core.scss
```diff
@@ -186,10 +185,8 @@ footer {
         }
     }
 
-    $width-1440px: 90em;
-
-    @media screen and (min-width: $width-1440px) {
-        width: $width-1440px;
+    @include media-breakpoint-up(xl) {
+        width: breakpoint-min(xl);
```
thibaudcolas: This seems to be the only place where the `xl` breakpoint is used. Should we re-do this with either `lg` or `xxl`?

client/scss/_tools.breakpoints.scss
thibaudcolas: Yes!

wagtail/admin/static_src/wagtailadmin/scss/core.scss
jjanssen: Ironic: it turns out it's not even called since the following is since the following from `sm` get applied.

image

client/scss/_tools.breakpoints.scss
jjanssen: In addition I also renamed and check the utility and visibility classes to have a consistent naming with the breakpoints. So `@small` became `@xs` and `@medium` became `@sm`.

jjanssen:
All done and double-checked the CSS output difference with the current master. There are no notable difference aside from the previously accidentally fixed quirks.

tomdyson:
This looks really good, @jjanssen! Thanks a lot for all your excellent work on this.

jjanssen:
Let me take a look at those last few changes. It shouldn't be on your plate just because I didn't do a proper double-check on other files.

Regarding your question. I don't think we even should have the `media-breakpoint-down` in the most ideal situation. My intention with this 1st iteration was to eventually refactor into a solution where mobile is targeted first and work our `media-breakpoint-up` from there. But that would be a hard pill too swallow just from a single pull request.

PR #4863, opened 2018-10-28
Make text 'There is one match' more translation-friendly

nott:
For languages that have more than 2 plural forms (e.g. Belarussian) it's important for all the translation to have counter variable included. Otherwise transifex won't let translators include a variable like that to the forms that need it (e.g. we can't use counter for 2, 3 or 4 in Belarussian).

image
image


m1kola:
Rebased and merged in 3a678a4e32e44ee1ccf8365bd8d26c491e992754. Thanks, Stas!

PR #4855, opened 2018-10-27
fix help text positioning to avoid text cutting with narrow field

davojta:
Fix for https://github.com/wagtail/wagtail/issues/3522

* Do the tests still pass? (http://docs.wagtail.io/en/latest/contributing/developing.html#testing)
_Yes_

* Does the code comply with the style guide? (Run `make lint` from the Wagtail root)
_Yes_

* For Python changes: Have you added tests to cover the new/fixed behaviour?
_Not applicapable_ 

* For front-end changes: Did you test on all of Wagtail’s supported browsers? **Please list the exact versions you tested**.
Chrome 70
Firefox 63.0

image

PR #4866, opened 2018-10-28
Fix StreamField full-width layout, UI regression from #4855

thibaudcolas:
Spotted by @jjanssen in https://github.com/wagtail/wagtail/pull/4855#issuecomment-433689296: 

> A regression occurred when this PR got merged which reflects to the streamfields.

> We used to render the streamfield on full-width:
image

> Now it renders collapsed:
image

---

This was due to the `> fieldset` selector, which relies on the structure of the HTML, which changed in #4855. I had seen the faulty UI but didn't realise it wasn't meant to look like that.

Tested in IE11, Edge17, Chrome, Firefox, iOS11


jjanssen:
LGTM! 👍

Issue #4087, opened 2017-12-01
Image editing screen feels under designed


tmsndrs:
When editing an image, the screen feels under designed, including tight field tittles, confusing layout and unstyled buttons. 

<img width="1256" alt="screen shot 2017-12-01 at 18 35 56" src="https://user-images.githubusercontent.com/7290588/33497767-ea1f4b7e-d6c6-11e7-8261-47c005af950b.png">

Relates to #4055 and #2339 (my comment about focal point UI)

benenright:
Here is a layout that should make this screen clearer and more usable on smaller desktop viewports. The idea is we use the 'Replace' link in place of the default 'Choose file' UI element, opening the file choose file dialog.

The styling on the 'File' field is to be consistent with the approach outlined on #4055 

![image editing 2x](https://user-images.githubusercontent.com/4907819/36314168-e1555f14-132b-11e8-8915-c24a974ac500.png)

PR #4868, opened 2018-10-28
[4087] Revamp the image editing page as far as possible

jjanssen:
Based upon the design from #4087 

**Some remarks upfront:**
- I did not change the header as it would feel weird having a different mark-up on this page. I think would be better if we would pick this up in header redesign ticket;
- I did not change the file field or the order difference between title and file, because this has a separate design proposal as defined in #4055 and it would make this change bigger than editing an image alone.

**The code changes**
- Added an additional `label-uppercase` and `label-uppercase` due to the way `field_as_li.html` is defined right now throughout the application and not change everything at once.
- Added additional utilities for global usage. Although I'm not sure whether to continue this in `admin` or within `client`.

**So what did it turn out to be?**
Without the URL generator
image

With the URL generator
image

On tablet (1024px)
image

On mobile (320px)
image
image

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/scss/_utilities.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/_utilities.scss
@@ -1,0 +1,1 @@
+@import 'wagtailadmin/scss/utilities/text';
--- a/wagtail/admin/static_src/wagtailadmin/scss/components/_forms.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/components/_forms.scss
@@ -1165,14 +1165,6 @@ select {
     @include transition(background-color 0.2s ease);
 }
 
-input[type=submit],
-input[type=reset],
-input[type=button],
-.button,
-button {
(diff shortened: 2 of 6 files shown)


jjanssen:
Tested in:
- [ ] Mobile Safari	iOS Phone	Last 2
- [ ] Mobile Safari	iOS Tablet	~~Last 2~~ latest only
- [ ] Chrome	Android	Last 2
- [ ] IE	Desktop	11
- [ ] Chrome	Desktop	Last 2
- [ ] MS Edge	Desktop	Last 2
- [ ] Firefox	Desktop	Latest
- [ ] Firefox ESR	Desktop	Latest
- [ ] Safari	macOS	~~Last 2~~ latest only

zerolab:
Ah, that is true. Missed it!
👍

jjanssen:
Note to self: update visibility utility classes once #4860 gets merged

nmorduch:
Merged in b07b7a6e2aedd6e5a1e7e0af8fe00d6da03d93f1

PR #4872, opened 2018-10-29
First positional argument for tox as test module/case path

benjaoming:
After playing a bit of the old "push and wait for Travis" game, I decided that adding this pattern would help me to quickly bootstrap a test environment in a familiar way.

Is it okay to add?


gasman:
Thanks @benjaoming! Seems like a good idea to me - merged in 5b0f2e1a546bf9b86f3c5a046e6694470ede5781.

PR #4873, opened 2018-10-29
{% include_block page.block as block_content %}

brabant:
Allows to store result of {% include_block %} in a variable for further use of filters

### DIFF
--- a/wagtail/core/templatetags/wagtailcore_tags.py
+++ b/wagtail/core/templatetags/wagtailcore_tags.py
@@ -78,15 +78,18 @@ def richtext(value):
 
 
 class IncludeBlockNode(template.Node):
-    def __init__(self, block_var, extra_context, use_parent_context):
+    def __init__(self, block_var, extra_context, use_parent_context, target_var):
         self.block_var = block_var
         self.extra_context = extra_context
         self.use_parent_context = use_parent_context
+        self.target_var = target_var
 
     def render(self, context):
         try:
             value = self.block_var.resolve(context)
         except template.VariableDoesNotExist:
+            if self.target_var is not None:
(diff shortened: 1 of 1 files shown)


gasman:
Thanks for the contribution @brabant!

Note that this will need tests and documentation before it's ready to be merged.

I'd really need to know more about the motivation for this change before we can decide it's something we want to adopt, though. Normally, the output of `{% include_block %}` would be an HTML fragment produced by a template, and it would be quite unusual to apply filters to that output - you'd normally apply those filters as part of that template. For blocks that return non-HTML values, such as DateBlock, `{{ my_block.value|date:"D d M Y" }}` would be a more natural thing to do. Is there something that can't be achieved by either of these options?

brabant:
At present I use it to wrap the h2 and h3 tags in difficult layout. 
Now headings are inserted through RichText and it would be possible to make the separate block - "Beautiful heading", but there are no wish to produce many blocks.

gasman:
Thanks for the clarification. After thinking about this some more, I think this option is likely to cause confusion for users - the natural reading of the `include_block` tag is that it will include the rendered version of the value into the page, so `{% include_block ... as var %}` is effectively saying "include this without including it".

For your application, adding the filter to `block.value` should work - e.g. `{{ block.value|transform_header_tags }}`. Django's `{% filter %}` tag would be another option: https://docs.djangoproject.com/en/2.1/ref/templates/builtins/#filter

brabant:
Really, it can create confusion. Just I modified inlcude_block under the needs and have thought that it can be useful also for others. 
Filter { { block.value|transform_header_tags } } - hasn't suited me since for some of my blocks the context, for example, a block-form insert is necessary

PR #4874, opened 2018-10-30
Collection display ordering

seb-b:
We had a client with 100+ collections for images and because the drop down is ordered by the treebeard path by default they were finding it hard to find the collection they needed.

This solution alphabetizes the relevant `Collection` queryset wherever it's used by the admin templates but still puts the Root collection at the top of the list.

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -1943,14 +1943,6 @@ class BaseCollectionManager(models.Manager):
     def get_queryset(self):
         return TreeQuerySet(self.model).order_by('path')
 
-    def for_display(self):
-        # This will output the Root collection first, and alphabetize the rest
-        return TreeQuerySet(self.model).annotate(
-            display_order=Case(
-                When(id=get_root_collection_id(), then=Value('-1')),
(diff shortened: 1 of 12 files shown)


wagtail/core/models.py
```diff
@@ -1943,6 +1943,14 @@ class BaseCollectionManager(models.Manager):
     def get_queryset(self):
         return TreeQuerySet(self.model).order_by('path')
 
+    def for_display(self):
+        # This will output the Root collection first, and alphabetize the rest
+        return TreeQuerySet(self.model).annotate(
+            display_order=Case(
+                When(id=get_root_collection_id(), then=Value('-1')),
```
gasman: `When(depth=1, then=Value('-1'))` would save a query here.

Actually, isn't this relying on the non-root collections having names that sort after the string '-1'? I'm not sure if that's something we can necessarily count on - `then=Value('')` is probably more reliable.

gasman:
Thanks @seb-b! Merged in aa2a973382f54b482eb6769984021415b52c54bc.

PR #4878, opened 2018-10-30
make img cover avatar object

alfuhigi:


gasman:
Thanks @alfuhigi! Please can you add a description of what this is fixing, perhaps with screenshots to show the 'before' / 'after' versions?

alfuhigi:
alt text


it happen when image width is not equal to height

jjanssen:
Hi @alfuhigi,

Although not directly visible in your before and after, I noticed this would result in stretching out the picture to full-height, regardless of the ratio. Which looks a bit quirky when you actually have a person or object on the uploaded picture.

For example:
image

Would result in:
image

Perhaps we should add cropping from the center when uploading an image on a fixed dimension (100x100 pixels which is also called when requesting gravatar). Just to make sure this will not stretch out the image.

alfuhigi:
Hi @jjanssen 

I thinks it need more work but for now I have simple fix by delete img tag and add avatar background image with user avatar for these files

wagtail/wagtail/admin/templates/wagtailadmin/home.html
```
   {% avatar_url user  as avatar %}
  <div class="avatar" style="background-image:url('{{avatar}}'); background-size: cover;">
```


wagtail/wagtail/admin/templates/wagtailadmin/shared/main_nav.html

```

{% avatar_url request.user  as avatar %}
 <span class="avatar square avatar-on-dark" style="background-image:url('{{avatar}}'); background-size: cover;">

```

jjanssen:
Hi @alfuhigi 

This would be a better way forward. It would mean the `.avatar` definition would get a `background-size: cover;` by default and previous image declarations would be replaced with a background-image. This way it would be compatible with and without gravatar. Is this something you can look into for this PR? Or do you need help with that?

lb-:
This ended up being raised as an issue

https://github.com/wagtail/wagtail/issues/8428

And is now fixed via 

https://github.com/wagtail/wagtail/pull/8521

So I will close this PR - thanks for the work done on this many moons ago @alfuhigi

PR #4879, opened 2018-10-31
Added last login to the admin Users list

noahbjohnson:
Added a table header and data element to the html template for users list. Table now includes a human readable last login date. Code is exactly the same as the Document list column.


noahbjohnson:
Now renders an empty <td> if there is a NULL.

screen shot 2018-10-31 at 11 05 59 am

jjanssen:
Thanks for the fast follow-up and thank you for contributing to Wagtail 🙌 
This now merged with 0dacda9 and will be available in Wagtail 2.4

PR #4883, opened 2018-11-01
Cache image renditions

tomdyson:
This small change removes one common cause of N+1 queries, by returning image rendition lookups from a cache rather than the database. Rendition caching has to be explicitly enabled, by defining a new item in `settings.CACHES` called 'renditions'.

* For Python changes: Have you added tests to cover the new/fixed behaviour? No

### DIFF
--- a/docs/advanced_topics/performance.rst
+++ b/docs/advanced_topics/performance.rst
@@ -30,6 +30,28 @@ We recommend `Redis <https://redis.io/>`_ as a fast, persistent cache. Install R
     }
 
 
+Caching image renditions
+------------------------
+
+If you define a cache named 'renditions' (typically alongside your 'default' cache), 
+Wagtail will cache image rendition lookups, which may improve the performance of pages 
+which include many images.
+
+.. code-block:: python
+
+    CACHES = {
+        'default': {...},
+        'renditions': {
+            'BACKEND': 'django.core.cache.backends.memcached.MemcachedCache',
+            'LOCATION': '127.0.0.1:11211',
+            'TIMEOUT': 600,
+            'OPTIONS': {
+                'MAX_ENTRIES': 1000
+            }
+        }
+    }
+
+
 Search
 ------
 
--- a/wagtail/images/models.py
+++ b/wagtail/images/models.py
@@ -6,6 +6,7 @@ from io import BytesIO
 
 from django.conf import settings
 from django.core import checks
+from django.core.cache import InvalidCacheBackendError, caches
 from django.core.files import File
 from django.db import models
 from django.forms.utils import flatatt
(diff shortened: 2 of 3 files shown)


gasman:
Looks good to me at first glance, thanks @tomdyson!

I'd be keen to know how the performance here compares to the database lookup - I seem to recall that we _didn't_ go with the cache for site lookups because it was marginally slower than getting the site record from the database

We'll also need test coverage for this.

tomdyson:
Performance impact will depend on the relative latencies of database and cache backends, but as a single reference point we're seeing n+1 rendition queries take ~150ms on image-heavy pages on Heroku:

screenshot 2018-11-02 12 33 03

Redis or Memcached requests are typically 1-2ms on hosted platforms like Heroku, so we should see a worthwhile improvement. It also frees up database connections, which can be a scarce resource on managed services.

I'll need some help with tests, though :)

timonweb:
I've just stumbled upon this after doing a perf review of a wagtail site, image renditions are indeed pretty expensive. That patch seems to solve the issue. Any updates? Do you need any help with it?

tomdyson:
Thanks @timonweb - did you try the patch on your site? It would be great to have your help. I think it just needs test coverage.

timonweb:
@tomdyson yes it works fine for me and has noticeably improved the performance, thank you! We really should get it into the master.

For now, I've put on my `CustomImage` model and works great in production. Hopefully, I'll have some time in the upcoming days to add some tweaks and tests here.

timonweb:
@tomdyson could you please rebase your PR to the latest master?

tomdyson:
> could you please rebase your PR to the latest master?

@timonweb done!

timonweb:
@tomdyson created a PR to this PR here: https://github.com/tomdyson/wagtail/pull/1

cyface:
Is this on the plan to merge into Wagtail 2.9?

tomdyson:
image

Local tests (using django-silk on the bakerydemo home page, with PostgreSQL and Redis) show significant savings on both query time and overall time when renditions are cached (demonstrated when the number of queries is 37 or 38 vs 60 or 61 without a cache).

zerolab:
Great idea to post the screenshot @tomdyson! This was one of the things I checked when looking at the PR, but did not think of posting any before/after screenshot 🤦‍♂️

While looking at this I noticed there's a query to collections for each image. a `select_related` will cut down the queries by another N (where N is the number of images displayed). Will submit that and others in a separate N+1 fixes PR

gasman:
Merged in f37da8f7d88701dffc6d4beb0e63270ec9470ad1 - thanks @tomdyson and @timonweb, and @zerolab for the testing!

timonweb:
Yay! Great! @gasman, thanks for merging this in! @tomdyson 👊

tomdyson:
@timonweb 👊

PR #4884, opened 2018-11-02
Allow pages to have stopword-only slugs

### LINKED ISSUE #4881: Pages named "before" raise slug validation errors on save
### Issue Summary

Naming a new page "**before**" or "**Before**" seem to leave the slug field blank on save, thus raising a page validation error. Manually entering the slug and saving the page a second time resolves the issue. I have tested this in versions 1.13.1 and 2.2.

### Steps to Reproduce

2. Create a child page (in this case a homepage but any model that inherits Page seems to have this issue)
3. Name the page **before** or **Before** and save.
4. Following error is raised "The page could not be created due to validation errors" and promote tab is highlighted.
5. Click on Promote tab and see that the slug field is empty.
 
This is a page because it doesn't generate the slug as it would for any other page title. Entering the slug manually or adding a character to 'before' ex.(before2) allows it to auto generate the page slug as expected.

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: (yes / no)
Yes

### Technical details

* Python version: 3.6.
* Django version: 1 and 2
* Wagtail version: 1.13.1 and 2.2
* Browser version: Chrome, Safari, Firefox

before_error
before_error2
before_error3

chosak:
This change modifies the [`cleanForSlug`](https://github.com/wagtail/wagtail/blob/b34b547c6abe441429d019289dc542ac76930015/wagtail/admin/static_src/wagtailadmin/js/page-editor.js#L240) function used when creating a page's slug from its title. The current behavior uses the Django [`URLify`](https://github.com/wagtail/wagtail/blob/b34b547c6abe441429d019289dc542ac76930015/wagtail/admin/static_src/wagtailadmin/js/vendor/urlify.js#L152) function, which removes stopwords like "before". If a page title consists only of such stopwords, the generated slug will be blank, thus confusingly preventing page save.

This change handles this case by falling back to an alternate slug generation approach that allows the stopwords to be used.

This does unfortunately introduce some potentially inconsistent behavior; for example, a page titled "Before me" will be given a slug of "me" and a page titled "Before" will be given a slug of "before".
(Honestly, the inclusion of "before" as a stopword is somewhat unexpected.)

Fixes #4881.

(I took a look at adding some JS unit tests for this change, but it wasn't obvious the best way to do so, given how this function is defined. I'd be happy to add tests given a pointer to a good approach.)

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/js/page-editor.js
+++ b/wagtail/admin/static_src/wagtailadmin/js/page-editor.js
@@ -411,3 +411,5 @@ $(function() {
         });
     });
 });
+
+module.exports.cleanForSlug = cleanForSlug;
--- a/wagtail/admin/static_src/wagtailadmin/js/page-editor.js
+++ b/wagtail/admin/static_src/wagtailadmin/js/page-editor.js
@@ -241,14 +241,18 @@ function cleanForSlug(val, useURLify) {
     if (useURLify) {
         // URLify performs extra processing on the string (e.g. removing stopwords) and is more suitable
         // for creating a slug from the title, rather than sanitising a slug entered manually
-        return URLify(val, 255, unicodeSlugsEnabled);
-    } else {
-        // just do the "replace"
-        if (unicodeSlugsEnabled) {
-            return val.replace(/\s/g, '-').replace(/[&\/\\#,+()$~%.'":`@\^!*?<>{}]/g, '').toLowerCase();
-        } else {
-            return val.replace(/\s/g, '-').replace(/[^A-Za-z0-9\-\_]/g, '').toLowerCase();
+        let cleaned = URLify(val, 255, unicodeSlugsEnabled);
+
+        if (cleaned) {
+            return cleaned;
         }
+    }
+
+    // just do the "replace"
+    if (unicodeSlugsEnabled) {
+        return val.replace(/\s/g, '-').replace(/[&\/\\#,+()$~%.'":`@\^!*?<>{}]/g, '').toLowerCase();
+    } else {
+        return val.replace(/\s/g, '-').replace(/[^A-Za-z0-9\-\_]/g, '').toLowerCase();
     }
 }
 
--- a/wagtail/admin/static_src/wagtailadmin/js/page-editor.js
+++ b/wagtail/admin/static_src/wagtailadmin/js/page-editor.js
@@ -243,6 +243,8 @@ function cleanForSlug(val, useURLify) {
         // for creating a slug from the title, rather than sanitising a slug entered manually
         let cleaned = URLify(val, 255, unicodeSlugsEnabled);
 
+        // if the result is blank (e.g. because the title consisted entirely of stopwords),
+        // fall through to the non-URLify method
         if (cleaned) {
             return cleaned;
         }
--- a/wagtail/admin/static_src/wagtailadmin/js/page-editor.test.js
+++ b/wagtail/admin/static_src/wagtailadmin/js/page-editor.test.js
@@ -1,0 +1,32 @@
+window.$ = require('./vendor/jquery-3.2.1.min');
+require('./vendor/urlify').default;
+
+const cleanForSlug = require('./page-editor').cleanForSlug;
+
+describe('page-editor tests', () => {
+  describe('cleanForSlug without unicode slugs enabled', () => {
+    beforeEach(() => {
+      window.unicodeSlugsEnabled = false;
+    });
+
+    it('should return a correct slug which is escaped by urlify', () => {
+      /* true triggers to use django's urlify */
+      expect(cleanForSlug('Before', true)).toBe('before');
+      expect(cleanForSlug('The', true)).toBe('the');
+      expect(cleanForSlug('Before the sun rises', true)).toBe('before-the-sun-rises');
+      expect(cleanForSlug('ON', true)).toBe('on');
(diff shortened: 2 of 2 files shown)


jjanssen:
Hi @chosak 

I'm pretty sure I've gotten some grey hairs in the past hour, but I managed to get a [working example](https://github.com/jjanssen/wagtail/pull/1/files#diff-c250150784f4b6a831d692a81e150eb2) based upon your PR with a small adjustment to get `cleanForSlug` exported and known to Jest.
If so it would be our first official test for the es5 codebase 🎉 

```
expect(cleanForSlug('Before', true)).toBe('before'); // OK
expect(cleanForSlug('The', true)).toBe('the'); // OK
expect(cleanForSlug('Before The', true)).toBe('before-the'); // OK
expect(cleanForSlug('Before The sunrise', true)).toBe('before-the-sunrise'); // Not ok
```
Where the last one results in:
```
Expected value to be:
      "before-the-sunrise"
    Received:
      "sunrise"
```
In addition it handles the `before` statement correctly, but things kind of get confusing if the string would contain one or multiple stopwords and a regular word and I'm not really sure if this would be desired behaviour? 

I really have no clue on why Django chose to use this approach and why Wagtail selectively applies the `useUrlify` boolean in slugifying and cleaning. Maybe somebody could enlighten us on this matter?

chosak:
@jjanssen thank you! `window.$ = require('./vendor/jquery-3.2.1.min');` was the magic incantation that I couldn't figure out.

Side note: when looking into this I also came across the interesting [rewire](https://github.com/jhnns/rewire) module that lets you test functions from a module without exposing them. I wasn't sure which was a worse thing to do: exposing a function just for the purposes of testing it, or testing an internal function by pretending it's exposed.

> why Wagtail selectively applies the useUrlify boolean in slugifying and cleaning

My guess from looking [here](https://github.com/wagtail/wagtail/blob/b34b547c6abe441429d019289dc542ac76930015/wagtail/admin/static_src/wagtailadmin/js/page-editor.js#L276) is that we do want to let users specify the slug themselves if they really want to, even if it contains stopwords. So a user might really really want the slug to be `before-the-sunrise`: if they title the page "Before the sunrise", see that it gets mangled to `sunrise`, and manually change it back, we want to allow that.

I agree that this is pretty confusing, I think made more so by the fact that some of the stopwords like "before", "from", "since" don't feel as throwaway as "a", "in", "is. Hopefully the fix proposed here at least prevents the issue described in #4881, pending a higher-level reevaluation of Wagtail behavior.

Thanks again for writing those tests; I can take a stab at pulling them in here unless you'd like to.

jjanssen:
>  when looking into this I also came across the interesting rewire module that lets you test functions from a module without exposing them. 

Yeah I've played around with rewire in a project of ours. It is actually pretty cool! But I'm a bit hesitant to introduce another package to our dependency listing. 

> I wasn't sure which was a worse thing to do: exposing a function just for the purposes of testing it, or testing an internal function by pretending it's exposed

I don't really have a problem with exposing it for the purpose of testing. The downside currently is that all these methods are already exposed to the window. Whereas in newer situations we could test it more easily by using ES6 classes or exporting ES6 methods properly and not exposing everything to the window object.

> Thanks again for writing those tests; I can take a stab at pulling them in here

Yes please :open_hands: This would already be a great step forward not having to the test these use-cases manually 👍

wagtail/admin/static_src/wagtailadmin/js/page-editor.js
```diff
@@ -241,15 +241,19 @@ function cleanForSlug(val, useURLify) {
...
         // for creating a slug from the title, rather than sanitising a slug entered manually
-        return URLify(val, 255, unicodeSlugsEnabled);
-    } else {
-        // just do the "replace"
-        if (unicodeSlugsEnabled) {
-            return val.replace(/\s/g, '-').replace(/[&\/\\#,+()$~%.'":`@\^!*?<>{}]/g, '').toLowerCase();
-        } else {
-            return val.replace(/\s/g, '-').replace(/[^A-Za-z0-9\-\_]/g, '').toLowerCase();
+        let cleaned = URLify(val, 255, unicodeSlugsEnabled);
+
```
gasman:
```suggestion

        // if the result is blank (e.g. because the title consisted entirely of stopwords),
        // fall through to the non-URLify method
```

chosak: Implemented in 6a5b955.

chosak:
I've added the unit tests from @jjanssen and the comment suggestion from @gasman; thanks both.

However, having sat on this PR a bit, I'm starting to have second thoughts whether this would actually be the right way to fix #4481 for Wagtail users/developers. See https://github.com/wagtail/wagtail/issues/4881#issuecomment-438710717 for more thoughts on this, and #4899 for an alternate proposal.

I've also opened #4898 to add @jjanssen's tests from this PR against the current logic, so as not to block them unnecessarily if this becomes a longer discussion.

gasman:
Merged in 44c905bf69a4ca118893b117daf55cf090c6ed65 + parents.

gasman:
Just noticed that the `module.exports.cleanForSlug = cleanForSlug;` line throws a JS error in Chrome: `Uncaught ReferenceError: module is not defined`
It should probably be wrapped in an 'if defined' check.

gasman:
Above issue fixed in #4939.

Issue #4122, opened 2017-12-13
StreamField ValidationError not shown at specific field or indicated in content tab


lb-:
### Issue Summary

When validating ``StreamField`` using the ``validators`` attribute, validation errors are indicated to the user but no field specific errors/text is shown.

This may be a related to this issue: #3441
Found this out after attempting to answer this Stack Overflow question: https://stackoverflow.com/questions/47760595/validate-no-duplicates-in-streamfield

### Steps to Reproduce

1. Start on the BakeryDemo project
2. Edit the base/models.py file - ``StandardPage`` model 
3. Create a validator as per [Django docs](https://docs.djangoproject.com/en/2.0/ref/forms/validation/#using-validators)
```
def validate_headings_are_unique(value):
    items = [data[1] for data in value.stream_data if 'heading' in data[0]]
    if len(set(items)) != len(items):
        raise ValidationError(
            'All headings must have unique values: %(value)s',
            code='invalid',
            params={'value': items},
        )
```
4. Edit the field body as such (using the validators attribute):
```
    body = StreamField([
        ('heading2', CharBlock(label='h2')),
        ('heading3', CharBlock(label='h3')),
        ('paragraph', RichTextBlock())
    ], blank=True, validators=[validate_headings_are_unique])
```
5. Edit a page using the StandardPage model
6. Enter two headings into the Body field, a H2 with value 'foo' and a H3 with value 'foo'
6a. Expected: Error treated like other validation errors on non StreamField fields...
   * Error at top: The page could not be saved due to validation errors
   * Error indicator on the content tab (1 in red circle)
   * Error on the field with the error text shown
6b. Actual: Error at the top of the page only, with no indication of what 'field' this error occurred on.
![wagtail - editing standard page - editor error](https://user-images.githubusercontent.com/1396140/33927178-5cf46644-e035-11e7-8e6a-dbcf2974e550.png)
Note: Page error only, no error on the content tab and no error at the 'Body' panel

### Technical details

* Python version: 3.4.3
* Django version: 1.11.6
* Wagtail version: 2.0a0

### Potential Solution 1 - Minor fix + Documentation of ``StreamBlockValidationError``

After digging around a bit, I can see that validation errors can be made with the ``StreamBlockValidationError`` from here:
https://github.com/wagtail/wagtail/blob/db6d36845f3f2c5d7009a22421c2efab9968aa24/wagtail/core/blocks/stream_block.py#L21-L29

The revised validator looks like:

```
[...]

PR #4887, opened 2018-11-07
Fixed tabbed streamfield error reporting

markfinger:
This fixes an issue where streamfields nested in tabs might raise validation errors that are not indicated in the tab error counts. In the case of a heavily tabbed, large admin page, this leads to a lot of user (and dev) confusion.

Re: #4122

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/js/page-editor.js
+++ b/wagtail/admin/static_src/wagtailadmin/js/page-editor.js
@@ -281,7 +281,7 @@ function initErrorDetection() {
     var errorSections = {};
 
     // first count up all the errors
-    $('.error-message').each(function() {
+    $('.error-message, .help-critical').each(function() {
         var parentSection = $(this).closest('section');
 
         if (!errorSections[parentSection.attr('id')]) {


markfinger:
For anyone else that hits the same problem, we're using the following workaround.

```
# wagtail_hooks.py

@hooks.register('insert_global_admin_js')
def global_admin_js():
    return format_html('<script src="{}"></script>', static('project/admin-tweaks.js'))
```

```
// static/project/admin-tweaks.js

$(function initErrorDetection() {
    var errorSections = {};

    // first count up all the errors
    $('.error-message, .help-critical').each(function() {
        var parentSection = $(this).closest('section');

        if (!errorSections[parentSection.attr('id')]) {
            errorSections[parentSection.attr('id')] = 0;
        }

        errorSections[parentSection.attr('id')] = errorSections[parentSection.attr('id')] + 1;
    });

    // now identify them on each tab
    for (var index in errorSections) {
        $('.tab-nav a[href="#' + index + '"]').addClass('errors').attr('data-count', errorSections[index]);
    }
});
```

lb-:
@markfinger  - thanks for this submission, I have tried this out locally and I cannot actually reproduce the scenario that this fixes.

I followed the steps to reproduce on #4122 - I still only get one global error at the top of the page and no additional error counts on the tabs or errors shown above the fields.

Could you advise as to the specific scenario that this PR fixes (or let me know what I may have missed).

It appears that the .help-critical class name only gets added by draftail, document chooser and image chooser so maybe this works for errors in those fields only?

markfinger:
Sorry, enough time's passed that I can't remember any further context about this.

lb-:
@markfinger - no problems

PR #4893, opened 2018-11-13
Use unittest.mock instead of external module.

sir-sigurd:


gasman:
Thanks @sir-sigurd! Unfortunately it looks like this is failing on Python 3.5 - any ideas?

sir-sigurd:
It turns out that 'assert_called()` and `assert_called_once()` were added in Python 3.6. They can be replaced with `assertTrue(mock.called)` and `assertEqual(mock.call_count, 1)` or this PR should be postponed to until Python 3.5 support is dropped.

Issue #929, opened 2015-01-26
"Offline" resizing of images used by the {% image %} tag


phylor:
## Problem

Images are resized during the request processing, if they do not exist yet. I'm currently having issues with this. Imagine users upload a batch of images, which are shown publicly in a grid. During the rendering of that grid, all those images need to be resized on the fly. If you are working with a slow file backend (e.g. Amazon S3), the resulting loading time isn't acceptable.

## Proposal 1

Have a `manage.py` command which:
- parses all templates for `{% image %}` tags to get required sizes
- resize all images and store them to the file backend

This command would be executed manually.

## Proposal 2

Do required resizing during the upload of an image. It would probably be necessary to specify required image sizes beforehand.

timheap:
I like proposal 2, resizing images during upload. If a list of common image sizes/filters was defined in the settings, the images could be processed through these filters ready for use in the rest of the site. Additionally, if these predefined filters were given a name, they could be used as a filter spec, keeping templates DRY:

```python
# settings.py
WAGTAILIMAGES_PREDEFINED_FILTERS = {
    'header': 'min-600x300',
    'gallery-thumb': 'fill-200x200',
    'gallery-full': 'max-1024x768',
}
```

```html+django
{# page.html #}
{% load wagtailcore_tags wagtailimages_tags %}

<h1>{{ self.title }}</h1>
{% image self.header header %}
{{ self.body|richtext }}

{% for item in self.gallery %}
    {% image item.image gallery-full as full %}
    <a href="{{ full.url }}">
        {% image item.image gallery-thumb %}
    </a>
{% endfor %}
```

kaedroho:
@timheap's solution may also be useful for the new Wagtail API. The API is designed to be generic and cannot make assumptions on the sizes of the images that the client will want. On the other hand, we probably don't want to give clients free reign on generating any renditions they want. I think letting the developer choose a list of allowed sizes and let the client pick the ones to use is a good middle ground.

The only concern I have is that it may end up generating many more renditions than are actually needed. I don't expect resizing to header or full width would be useful for someones avatar, for example. Option 2 would be much less affected by this than option 1.

gasman:
[...]

PR #4895, opened 2018-11-13
Ability for filter_spec to be a template variable

originell:
* Do the tests still pass? (http://docs.wagtail.io/en/latest/contributing/developing.html#testing) *Yes*
* Does the code comply with the style guide? (Run `make lint` from the Wagtail root) *Yes*
* For Python changes: Have you added tests to cover the new/fixed behaviour? *Yes*
* For front-end changes: Did you test on all of Wagtail’s supported browsers? **Please list the exact versions you tested**. *No frontend changes*
* For new features: Has the documentation been updated accordingly? *Not yet*

--- 

This pull requests proposes allowing `filter_spec` used in the `{% image %}` template tag to be a template variable.

Why? In most bigger projects it is usually a wish to keep a single source of truth. Now, sprinkling `width-300` and `width-325` etc. throughout your templates is not really DRY.

In order to solve this we did something along these lines:

```python
# settings.py
WAGTAIL_RENDITION_FILTER_SPECS = {
    "thumbnail": "fill-100x100",
    "small": "width-300",
    "medium": "width-600",
    "large": "width-1200",
    "x-large": "width-1600",
}
```
```python
# context_processor.py
from django.conf import settings


def standardized_image_renditions(request):
    """Exposes our ``WAGTAIL_RENDITION_FILTER_SPECS`` setting in the templates.

    So we don't have to hardcode image rendition formats throughout all of
    our (template) code.
    """
    return {"renditions": settings.WAGTAIL_RENDITION_FILTER_SPECS}
```

Finally, in our templates I wanted to replace the "hardcoded" specs with our new standardized variables:

```html
<!-- our template before -->
{% image box.image width-300 as small %}
{% image box.image width-1200 as medium %}
<img src="{{ small.url }}" srcset="{{ small.url }} 1x, {{ medium.url }} 2x" class="box__image__img" alt="" />
```
```html
<!-- our template after... -->
{% image box.image renditions.small as small %}
{% image box.image renditions.medium as medium %}
<img src="{{ small.url }}" srcset="{{ small.url }} 1x, {{ medium.url }} 2x" class="box__image__img" alt="" />
```

Sadly, this lead to an exception (invalid filter spec), as wagtail does not support template variables being the filter spec. Or I just couldn't find how to do it the right way 😅 Hence, this pull request.

I noticed the use of `@cached_property`. I guess this means that constructing the filter is rather expensive. I could imagine that you are not too happy with my proposed changes then, as they might introduce a performance hit. Not sure how big it is though.

Looking forward to comments 😄

### DIFF
--- a/docs/releases/2.4.rst
+++ b/docs/releases/2.4.rst
@@ -31,6 +31,7 @@ Other features
  * Added ``file_hash`` field to documents (Karl Hobley, Dan Braghis)
  * Added last login to the user overview (Noah B Johnson)
  * Changed design of image editing page (Janneke Janssen, Ben Enright)
+ * The ``{% image %}`` templatetag now supports passing a template variable as filter spec (Luis Nell)
 
 Bug fixes
 ~~~~~~~~~
--- a/wagtail/images/templatetags/wagtailimages_tags.py
+++ b/wagtail/images/templatetags/wagtailimages_tags.py
@@ -99,7 +99,13 @@ class ImageNode(template.Node):
         if not image:
             return ''
 
-        rendition = get_rendition_or_not_found(image, self.filter)
+        try:
+            rend_filter = Filter(
+                spec=template.Variable(self.filter_spec).resolve(context)
+            )
+            rendition = get_rendition_or_not_found(image, rend_filter)
+        except template.VariableDoesNotExist:
+            rendition = get_rendition_or_not_found(image, self.filter)
 
         if self.output_var_name:
             # return the rendition object in the given variable
--- a/wagtail/images/tests/tests.py
+++ b/wagtail/images/tests/tests.py
@@ -52,6 +52,18 @@ class TestImageTag(TestCase):
         result = self.render_image_tag(None, "width-500")
         self.assertEqual(result, '')
 
+    def test_image_tag_spec_as_variable(self):
+        temp = template.Template('{% load wagtailimages_tags %}{% image image_obj filter_spec %}')
+        context = template.Context(
+            {'image_obj': self.image, 'filter_spec': 'width-400'}
+        )
+        result = temp.render(context)
+
+        # Check that all the required HTML attributes are set
+        self.assertTrue('width="400"' in result)
+        self.assertTrue('height="300"' in result)
+        self.assertTrue('alt="Test image"' in result)
+
     def render_image_tag_as(self, image, filter_spec):
         temp = template.Template(
             '{% load wagtailimages_tags %}{% image image_obj ' + filter_spec


gasman:
Thanks @originell! This looks good to me - happy to merge this once we have documentation in place.

I don't think we've ever formally tested the performance of building the filter object - it's almost certainly insignificant compared to the actual image processing, so not something we need to worry about I reckon.

originell:
I just pushed the docs. Hope they are alright.

While writing them, I started thinking if it wouldn't be better to simply add the setting and the context processor to the `wagtail.images`? So it would be something along these lines:

1. A setting called `WAGTAIL_IMAGES_FILTER_SPECS = {…}`
2. A context processor: `wagtail.images.context_processors.filter_specs`

That would be a very opinionated approach to it though, which I personally like. However, it would cut down the documentation quite a bit.

gasman:
It's just occurred to me that this approach is going to cause problems if there's an unrelated variable called `original` floating around in the template context.

originell:
Yes there is a possibility for collisions. However:

```python
    def filter_specs(request):
        """Exposes our ``RENDITION_FILTER_SPECS`` setting in all templates.

        So we don't have to hardcode image rendition formats throughout all
        of our (template) code.
        """
        return {"filter_specs": settings.RENDITION_FILTER_SPECS}
```

As you can see, the settings that get defined are kind of namespaced. So you don't access `original`, but `filter_specs.original`. Still, you are right. There is the possibility that another view puts `filter_specs` in the template. This would then collide.

I don't know of a way to detect that automagically. I could only think of two things:

1. further refining the name.. so `wagtail_filter_specs` instead of `filter_specs`. Lowering chances.
2. putting it into the documentation I wrote as a note for the context processor.

I can see how this problem is a big contra for integrating the approach further into wagtail source, as lined out by my last comment. So maybe it would be best to simply add this as a danger-note to the documentation.

gasman:
What I mean is that if there's existing code out there which contains the tag `{% image myimage original %}` - i.e. render this image using Wagtail's built-in `original` filter - and that code also has an unrelated `original` variable in the template context, then that code will break when they upgrade to a version of Wagtail with this change in it, because the variable is used in preference to the built-in name.

We could fix that easily enough by swapping round the cases so that we try to match against a built-in rule first and then fall back to checking it as a variable name, but that's brittle too (it'll break if we introduce a new built-in filter name that collides with the variable name they're using). It would be better (and [url=https://miguelgfierro.com/blog/2018/python-pro-tips-understanding-explicit-is-better-than-implicit/]more pythonic[/url]) to explicitly distinguish between strings that are variable names and strings that are literal filter names - the "proper" way of doing that is probably to do what Django did with the `{% url %}` tag long ago, and require non-variable-name strings to be in quotes. I'm not sure if that's aesthetically the right thing to do for the `{% image %}` tag though :-/

originell:
Ahhhh ok! Now I got it. Thanks for taking the time to explain. I will give this some more thought. Maybe I can come up with something neat.

originell:
Sorry for not coming back to this sooner.

I really like the idea of having this work like the `{% url %}` tag. Sooo the plan is that first I prototype this in one of our projects here at the company.. see how well it turns out and then push this here for further discussion.

I do get the point about aesthetics, so I'm curious what it will feel like. :-) I think I will be able to start working on this in a week.

originell:
So, I finally managed to come back to this. 

In the changes you can see that there is a new template library called `wagtailimages_next`. It contains a new `image` and `image_url` tag, that works much like djangos own `url` tag. As it is a backwards incompatible change, I decided to move it into its own library.

Now, this will probably spark a lot of discussion and I realise that this might not be wagtail cores vision for the image tag. Hence, I fully understand if you decide to not accept this merge request. I just thought I'd give this another try. 

Generally, I would love to see this being the "next generation" of image tag. Plus, publishing a small python package that offers the same functionality as an official backport for wagtail versions before the one this could be included in 😃

---

Just as a quick overview and to refresh everyones memories. Instead of writing:

```html+django
{% image avatar fill-200x200 as img %}
```

You would have to explicitly use strings (single or double quote):

```html+django
{% image avatar "fill-200x200" as img %}
```

The huge benefit is, that we would finally be able to supply variables to the tag. Which in turn allows for standardisation of filter specs across a project.

It would also allow for some amazingly crazy things:

```html+django
{# assuming my_avatar_spec = "width-400|format-jpeg" #}
{% image avatar my_avatar_spec %}
<img src="..width-400.jpg" width="400" height="300" />

{% image avatar my_avatar_spec "alt"=user.display_name %}
<img src="..width-400.jpg" width="400" height="300" alt="originell" />

{% image avatar my_avatar_spec "alt"=user.display_name "aria-title"=user.display_name|title as img %}
```

You can even go as far as having the html attribute "key" (alt=) as a variable:

```html+django
{# assuming tooltip_setting = 'data-tooltip' #}
{% image avatar my_avatar_spec tooltip_setting=user.public_name %}
{# which would lead to #}
<img src="…" data-tooltip="originell" />
```

chosak:
@originell this is really great. Reading this implementation reminded me a bit of https://github.com/wagtail/wagtail/issues/929#issuecomment-71569104 (and below comments), which proposes defining the filter specs in settings to allow for bulk offline generation of renditions. This work feels well-aligned with that goal.

Personally I rather like the use of the `{% url %}` style with quotations, as unquoted strings in template tags seem to more often refer to variable names, not strings. This would be a breaking change but perhaps it could be one introduced over a few releases, with deprecation warnings if filter specs like `original` are used without quotes?

originell:
Glad you like it! Yeah, that part about "offline" generation is a nice side effect. With the new template tags, the context processor from the docs and some elbow grease, it shouldn't be too much of a challenge to offer offline generation.

---

As a matter of fact, it is what initially triggered my need to implement this. We have several projects where generating the renditions during first request is not acceptable, so we generate them on save, by using the django `post_save` signal (credit for this piece of code goes to @mr-stateradio):

```python
@receiver(post_save, sender=Image)
def create_image_renditions(sender, instance, created, **kwargs):
    for filter_spec in settings.WAGTAIL_RENDITION_FILTER_SPECS.values():
        # prerender all available renditions
        instance.get_rendition(filter_spec)
    return True
```

We had these filter specs duplicated: in all templates and then again in the settings. So you always had to modify both places. Which, to be very blunt, sucked.

The code above is quite a bit too simple to handle multi-model needs, but I could imagine one could upgrade  this to a dict structure like this:

```python
WAGTAIL_RENDITION_FILTER_SPECS = {
    "core.User": {  # django model name or maybe content type?
        "avatar_small": "fill-200x200",
        # …
    }
}
```

That would be a good basis for baked-in async/offline generations.

---

Back to topic: I just couldn't remember how django migrated their `url` tag way back when. Thankfully, the old docs are still online, so [we can even read the migration notes](https://django.readthedocs.io/en/1.4.X/ref/templates/builtins.html#url) (scroll down a bit until "Forwards compatibility"). So, they added deprecation notices to the old tags and slowly phased those old ones out over 2-3 versions. 

What I like about the way django did it is that it is really explicit. Keeping both versions for a year or more, with warnings all over the place and then simply removing it. If you didn't migrate then, breaking everything seems quite ok?

Also, the idea of a `future` library seems quite nice.. `{% load imagetags from wagtail_future %}` 😎

originell:
Hey guys, just wanted to check up if there is anything I can do to make accepting this pull request an easy decision :-)

docs/advanced_topics/images/standardizing_renditions.rst
```diff
@@ -0,0 +1,138 @@
...
+It would be much more convenient to be able to do something along these
+lines:
+
+ .. code-block:: html+Django
+
+    {% image box.image filter_specs.medium as regular %}
+    {% image box.image filter_specs.large as hidpi_retina %}
+    <img src="{{ regular.url }}" srcset="{{ regular.url }} 1x, {{ hidpi_retina.url }} 2x" />
+
+Wagtail let's you do just that, because the filter spec can be a template variable.
```
tomdyson: let's -> lets

docs/advanced_topics/images/standardizing_renditions.rst
```diff
@@ -0,0 +1,138 @@
...
+
+You can implement this any way you want. An example is illustrated below.
+
+
+Setup
+-----
+
+We are going to achieve three things:
+
+1. We will unify all filter specs in a common python file (e.g. ``settings.py``).
```
tomdyson: python -> Python

docs/advanced_topics/images/standardizing_renditions.rst
```diff
@@ -0,0 +1,138 @@
...
+2. A context processor will put these specs into our templates.
+3. As the filter specs now live in a single place (``settings.py``) you can
+   also import them in your python code. This is great if you are :ref:`image_renditions`.
```
tomdyson: python -> Python

I don't understand "This is great if you are"

docs/advanced_topics/images/standardizing_renditions.rst
```diff
@@ -0,0 +1,138 @@
...
+
+
+Common filter specs
+^^^^^^^^^^^^^^^^^^^
+
+You can put the filter specs anywhere you want. Highly recommended is a place
+that you can access freely from anywhere in your code. For most django
```
tomdyson: django -> Django

tomdyson: I suggest dropping "Highly recommended is a place that you can access freely from anywhere in your code."

docs/advanced_topics/images/standardizing_renditions.rst
```diff
@@ -0,0 +1,138 @@
...
+                ]
+            },
+        }
+    ]
+
+Using the filter specs
+----------------------
+
+Following are two examples showing how to use these newly standardized filter
+specs in your templates and in your python code.
```
tomdyson: python -> Python

docs/advanced_topics/images/standardizing_renditions.rst
```diff
@@ -0,0 +1,138 @@
...
+ 
+    from django.conf import settings
+
+    # Let's say you want to generate a tiny, low quality, version of an image.
+    low_quali_image = myimage.get_rendition(
+        "{spec}|jpegquality-25".format(
+            spec=settings.RENDITION_FILTER_SPECS['thumbnail']
+        )
+    )
+    # or if you are using python 3.6 and like f-strings
```
tomdyson: "python 3.6" -> "Python 3.6 or later,"

tomdyson:
Thanks for the prompt, @originell! I've assigned @gasman to review again.

docs/advanced_topics/images/standardizing_renditions.rst
originell: haha woops. guess I didn't finish a thought there

originell:
I was thinking on maybe adding a small section about "rendition pre-generation" (as seen in [the above comment](https://github.com/wagtail/wagtail/pull/4895#issuecomment-477734553) `create_image_renditions`). Would that be something you guys like in your docs?

originell:
I'm just putting this here for other people looking/googling. If you use `wagtailimages_next`, you can fairly easily do offline generation (as is also discussed here https://github.com/wagtail/wagtail/issues/929#issuecomment-71569104)

Assuming in your settings.py you have something like this:

```python
# settings.py
OUR_WAGTAIL_RENDITION_FILTER_SPECS = {
    "preview": "width-20|format-jpeg|jpegquality-40",
    "medium": "width-800",
    "large": "width-1600",
}
```
You can create the following management command:

```python
# management/commands/generate_renditions.py
from django.conf import settings
from django.core.management.base import BaseCommand

from wagtail.images.models import Image


class Command(BaseCommand):
    """Generates all possible image renditions."""

    help = "Generates all possible image renditions."

    def handle(self, *args, **options):
        possible_renditions = (
            len(settings.OUR_WAGTAIL_RENDITION_FILTER_SPECS)
            * Image.objects.all().count()
        )
        self.stdout.write(
            f"Generating renditions for all available images ({possible_renditions})."
        )
        i = 0
        for img in Image.objects.all().iterator():
            for (
                spec_name,
                filter_spec,
            ) in settings.OUR_WAGTAIL_RENDITION_FILTER_SPECS.items():
                i += 1
                self.stdout.write(
                    f"({i}/{possible_renditions}) Rendition {spec_name} for {img}"
                )
                img.get_rendition(filter_spec)
```

and then simply run it via `./manage.py generate_renditions`. It might take some time ;-)

originell:
hey @tomdyson any clue why this widget says "1 change requested", even though I think I've resolved all the requested changes?

image

docs/advanced_topics/images/standardizing_renditions.rst
```diff
@@ -125,14 +124,14 @@ In your code
 As before, this assumes that you have your filter specs in the ``settings.py``.
 
  .. code-block:: python
- 
+
     from django.conf import settings
 
-    # Let's say you want to generate a tiny, low quality, version of an image.
+    # Lets say you want to generate a tiny, low quality, version of an image.
```
tomdyson: In this case "Let's" was the correct version :)

tomdyson:
@originell no, as far as I can see you resolved all the issues I raised (there's a tiny new one, but not critical). I'm happy for @gasman to merge if he's OK with everything else. Thanks again for this contribution!

docs/advanced_topics/images/standardizing_renditions.rst
originell: sometimes this english not the bestest one haha

thanks for finding that one. fixed with latest commit

gasman:
Thanks @originell - this is looking good!

I'm still slightly wary about the filter becoming a full-fledged string expression (and explicitly appearing as one within the tag, as opposed to being something less obvious but still discoverable by advanced users - like, say, `{% image my_image width-{some_var} %}`). While it does feel like the natural step towards making the `{% image %}` tag more flexible (as it was with the `{% url %}` tag in Django all those years ago...), I have a feeling it might open the floodgates to people doing ill-advised things with that freedom, like exposing it as a field for editors to fill in - and breaking Wagtail's general assumption that there's going to be a small finite number of rendition sizes in use on a site.

Ultimately, though, I think making the leap is the right thing to do - it's not worth holding back a useful feature out of fear of how it could be misused.

I like the approach of introducing a `wagtailimages_next` tag library as a migration path. We should work out a plan and timeline for encouraging people to switch to the new tag - if that's a process we want to start right away (rather than, say, waiting for a Wagtail 3.0 release) then we should add deprecation warnings to the old tag library and update our existing documentation to use the new tag.

Will do some in-depth reviewing of this soon, but as an initial point of feedback - I'm not keen on having the left-hand side of kwargs treated as a string expression, i.e. the need to quote `"alt"` in `{% image myimage original "alt"="hello" %}`. It's inconsistent with how kwargs behave in tags like `{% include %}` and `{% url %}`, and the use-cases for putting an arbitrary string expression there are pretty obscure - I don't think it's worth the extra syntactic awkwardness that it adds to the more common `alt="hello"` case. If people really need to set attribute names dynamically, they can always use the `{% image ... as myvar %}` variant and build up the final `` tag within the template instead.

originell:
Thanks for taking the time!

> I have a feeling it might open the floodgates to people doing ill-advised things with that freedom

I completely understand this and I will also think some more about it. Maybe we can find a way to offer the "more freedom" behind a security door with 8 bolts and a DNA test. My first intuition would be to actually have a `{% load wagtailimages_pro %}` template library. But that on the other hand breaks with the "one way to do it" philosophy. Or having a second image tag called `{% image_var %}`, which is also kind of weird.

> It's inconsistent with how kwargs behave in tags

True that. Will change that to make it consistent! Basically just did this because I could haha

Looking forward to some in-depth feedback!

originell:
Oh and:

> Wagtail's general assumption that there's going to be a small finite number of rendition sizes in use on a site.

To give some more context about why I started this: this assumption worked pretty well for us, up until the point where we started using ``. That's when our rendition-variations exploded (multiple header image sizes, multiple user avatar sizes,…). After I caught a several wrong rendition sizes in one project's API endpoint, I started playing around and now we're here ;-)

gasman:
Going to have to side against 2019-me on this one!

The current unquoted version of the filter is an established pattern that works for the vast majority of people, and letting go of that for something that is, at best, a power-user use case is not something we should do lightly. (The alternative of supporting unquoted filters alongside quoted ones, and using heuristics to guess whether an unquoted identifier is a variable name or not, isn't sustainable either.)

I'm not sure that the analogy with Django's `{% url %}` tag going from unquoted to quoted is a valid one - notionally what the `{% image %}` tag is doing is far less of a "string processing" task, and in a tag like `{% image foo width-300 %}`, the `width-300` feels more like a syntactic element of the tag rather than an actual string.

For cases where you really do want a variable as the filter spec, it's not hard to bypass the `{% image %}` tag and come up with your own wrapper for [get_rendition](https://docs.wagtail.org/en/stable/advanced_topics/images/renditions.html#image-renditions), whether that's passing rendition objects from `get_context` or building the suggested `{% image_var %}` tag. In fact, I think there's a strong argument that _not_ providing a ready-made solution like that is in keeping with Django's stance of not making the template language too powerful, to nudge you away from putting too much logic into the template.

For the original use-case given here, assigning aliases to frequently-used filters, I'd contend that the best solution is actually to define a dedicated tag for each one using `simple_tag`, so that you'd just write `{% thumbnail_image page.photo %}` or `{% medium_image page.hero_image as img %}` or similar.

PR #4897, opened 2018-11-14
Add Slovak character map for JS slug generation

chosak:
Porting of [Django PR #9325](https://github.com/django/django/pull/9325), which fixes [Django ticket #28768](https://code.djangoproject.com/ticket/28768). These changes were made in Django 2.1.

This adds support for certain Slovak characters when auto-generating slugs for new pages. This logic is only triggered if [`settings.WAGTAIL_ALLOW_UNICODE_SLUGS`](http://docs.wagtail.io/en/v2.1.1/advanced_topics/settings.html#unicode-page-slugs) is set to `False`. 

To test, create a new page and enter a title like "Vŕba". The page's slug field will be automatically set to "vrba". In the previous behavior, this would instead be automatically set to "vba".


gasman:
Merged in 17fc5cc913d266bc584a7683ca68c3e8af66fa4a - thanks @chosak!

PR #4898, opened 2018-11-14
Add JavaScript unit tests for page slug generation

chosak:
This change adds JS unit tests for the Wagtail admin `cleanForSlug` function that generates page slugs from page titles. This function is called in two ways: once to use (a copy of) the Django `URLify` function, and once to use a more basic regex-based slug generation.

This PR adds the new JS unit tests written by @jjanssen for PR #4884 (and modified slightly here to work with the existing logic). This brings in those tests independently from the behavior change implemented on that PR, to serve as a baseline and a nice example for similar JS tests in future.


gasman:
Merged as part of #4884

PR #4900, opened 2018-11-14
allow panels & edit_handler to be defined on ModelAdmin

mr-stateradio:

* Do the tests still pass? (http://docs.wagtail.io/en/latest/contributing/developing.html#testing) *YES*
* Does the code comply with the style guide? (Run `make lint` from the Wagtail root) *YES*
* For Python changes: Have you added tests to cover the new/fixed behaviour? *YES*
* For front-end changes: Did you test on all of Wagtail’s supported browsers? **Please list the exact versions you tested**.  *No frontend changes*
* For new features: Has the documentation been updated accordingly? *Not yet*

---

This pull requests proposes to allow `edit_handler` and `panels` to be defined on a `ModelAdmin` definition. Currently it is only possible to define `edit_handler` and `panels` on a Model definition, which forces a tight coupling between Model definition and wagtail. However it would be nice and would increase decoupling if `panels` and `edit handlers` could also be defined on a `ModelAdmin` instance, since models managed via a `ModelAdmin` often live in non-cms apps.

See also this Slack thread for further details:
https://wagtailcms.slack.com/archives/C81FGJR2S/p1541582510134300

This pull request contains a modified `ModelFormView.get_edit_handler` and test cases to verify this functionality:

```python
class ModelFormView(WMABaseView, FormView):

    def get_edit_handler(self):
        if hasattr(self.model, 'edit_handler'):
            edit_handler = self.model.edit_handler
       # this elif is new...
        elif hasattr(self.model_admin, 'edit_handler'):
            edit_handler = self.model_admin.edit_handler
       # and this elif is new...
        elif hasattr(self.model_admin, 'panels'):
            panels = self.model_admin.panels
            edit_handler = ObjectList(panels)
        else:
            fields_to_exclude = self.model_admin.get_form_fields_exclude(request=self.request)
            panels = extract_panel_definitions_from_model_class(self.model, exclude=fields_to_exclude)
            edit_handler = ObjectList(panels)
        return edit_handler.bind_to_model(self.model)
```

Looking forward to comments 😄


ababic:
Hi @mr-stateradio,

Sorry I'm late to the party on this discussion, but I feel it would be helpful to outline some of the reasons why I didn't implement this myself - either while `wagtailmodeladmin` was still a separate package, or since it became `wagtail.contrib.modeladmin`.

I'm certainly all for separation of concerns and improving flexibility - However, implementing this solely into `ModelFormView` introduces a couple of inconsistencies that should be considered:

**This will work for some models, but not others**

If a model happens to inherit from `wagtail.core.Page`, then this code won't ever be used, because `modeladmin` redirects users off to the existing page management views for creating and editing pages, and they don't care about `ModelAdmin` instances at all.

**The interface for the same model will differ elsewhere within the Wagtail CMS**

E.g. in the snippet add/edit views (if the model were also registered as a snippet). And, even if `modeladmin` implemented its own set of page edit/add views to respect the `ModelAdmin` panels, the default page edit/add views would still ignore them.

IMHO, the best way to add this functionality without the inconsistencies is to take a slightly different approach. e.g.:

- Add some kind of central 'ModelAdmin registry' to `wagtail.admin` to keep track of which models have a `ModelAdmin` class registered, and can give you the `ModelAdmin` instance for a given model (kinda like Django admin does);
- Wherever panel sets are derived from models, query this central registry, so that panels defined on `ModelAdmin` classes can be respected throughout. 
- Handle panel set generation well in a central place, so that there aren't multiple places implementing their own version of this behaviour.

This central 'ModelAdmin registry' would be useful for other things too, such as helping to determine the most appropriate 'edit url' for a given model instance, or 'add url' for a given model - But, that's something for another time :)

gasman:
Hi @ababic - thanks for the feedback.
While I appreciate the concerns re consistency, I think the ability to work with models that are "out of reach" (due to being part of a third-party package, for example) is a big enough win to outweigh that - and in that particular use-case, the inconsistencies you mention generally won't arise (because the external models typically won't be pages or snippets).

I also think it may not necessarily be a bad thing to permit multiple views of the same model, each presenting a different subset of fields - for example, a user model where ordinary web editors can edit their 'profile' fields (biography, social media accounts etc) but only admins can edit their core account details.

I wouldn't be against the central registry approach - more opportunities for decoupling the admin interface for models, including the page/snippet edit interfaces, would be a good thing - but I think that's something for another time.

ababic:
Thanks @gasman.

> While I appreciate the concerns re consistency, I think the ability to work with models that are "out of reach" (due to being part of a third-party package, for example) is a big enough win to outweigh that - and in that particular use-case, the inconsistencies you mention generally won't arise (because the external models typically won't be pages or snippets).

This is a good point. Perhaps we can just outline some of the limitations in the docs/release notes to help reduce potential confusion? 

My main concern is developers who have registered models as snippets too in order to make use of `SnippetChooserPanel` and `SnippetChooserField` - which, I think is still pretty common. It should at least be made clear to them that moving the panels onto the `ModelAdmin` class instead will affect consistency.

> I also think it may not necessarily be a bad thing to permit multiple views of the same model, each presenting a different subset of fields.

I only really think it's bad when the inconsistency is unexpected, e.g. with the ModelAdmin + Snippet registered scenario. Incidentally, these changes would give developers a little more scope to offer multiple views of the same model via ModelAdmin customisation, which I'm all for.

ababic:
@mr-stateradio. If these changes are to go ahead in their current form, then I do have a suggestion that might help make things easier to override for developers...

What do you think about moving this logic over to the `ModelAdmin` class, and have `ModelFormView` refer to that? This way, developers could still override the method at view level, but would also have scope to tweak things directly on the `ModelAdmin` class, without having to take the 'custom view' route.

mr-stateradio:
@ababic just to clarify. You suggest to move the `ModelFormView.get_edit_handler()` method to the `ModelAdmin` class in order developers can override it more easily?

If I understood you correctly, that sounds good to me.

ababic:
@mr-stateradio Yes, that's pretty much it. But for backwards compatibility, the method on `ModelFormView` would stay (it would just call `ModelAdmin.get_edit_handler()` to get a result). Maybe it could pass some useful arguments in the process too, such as `request`, `instance`, and maybe an `is_new` Boolean?

mr-stateradio:
@ababic that sounds good.

Regarding the argument `is_new` you mentioned.. since we pass the argument `instance` a developer could easily check whether the instance is new by checking for `is_new = instance.id is None`. So do you think it makes sense to provide it as a separate method argument?

ababic:
@mr-stateradio, I'll leave that one up to your best judgement. It would be a nice touch, as I know switching up panels depending on whether it's a 'create' or 'edit' view is something I've done a lot of in the past - and the 'cannon' way to detect whether an object has been saved or not can be a bit illusive to even experienced Django users (I _think_ it's to use `object._state.adding`) - but I certainly wouldn't class it as necessary.

mr-stateradio:
@gasman @ababic I provided the discussed changes and moved the `ModelFormView.get_edit_handler()` method to the `ModelAdmin` class.

Let me know what you think about the suggested changes.

ababic:
Thanks for this @mr-stateradio! It'll be a few days until I'm able to play around with this fully, but from glancing over the code changes and test additions, it's looking great. Good work :) 

Part of me now wonders whether getting an edit handler object and binding it to the model might be better done in two separate methods, but this is already hugely useful as it is, and I don't want to hold things up any longer unnecessarily.

If anyone else can review more fully in the meantime, please feel free. If not, I will get to it as soon as I'm able.

wagtail/contrib/modeladmin/options.py
```diff
@@ -370,6 +371,29 @@ def delete_view(self, request, instance_pk):
         view_class = self.delete_view_class
         return view_class.as_view(**kwargs)(request)
 
+    def get_edit_handler(self, instance, request):
```
ababic: @mr-stateradio I know the original method did all of this in one method, but I feel we'd be missing a trick here if we didn't break things down a little bit further. For example, we could rename this method to `get_bound_edit_handler()` (which would return the same value it does currently), but have it hand over to a `get_edit_handler` method to get/create an `edit_handler` instance, which could in turn hand over to a `get_panels()` method to get/create a `panels` list.

There's certainly a line to be drawn with devolution, but in this case, I feel it would result in much clearer code, greatly improve the opportunities for overriding/extending, and would in turn make tests clearer and easier to write. What are your thoughts?

wagtail/tests/modeladmintest/wagtail_hooks.py
```diff
@@ -91,6 +92,41 @@ class VenuePageAdmin(ModelAdmin):
     exclude_from_explorer = True
 
 
+class PersonAdmin(ModelAdmin):
+    model = Person
+    exclude_from_explorer = True
```
ababic: The `exclude_from_explorer` option is only relevant when `model` is a subclass of `Page`. Since these new models are just standard Django models, the option should be removed for clarity.

wagtail/contrib/modeladmin/options.py
ababic: I suppose we could just move the `edit_handler.bind_to_model(self.model)` line back to `ModelFormView.get_edit_handler()`, rather than introduce separate `get_bound_edit_handler()` and `get_edit_handler()` methods.

wagtail/contrib/modeladmin/options.py
```diff
@@ -370,6 +371,29 @@ def delete_view(self, request, instance_pk):
...
+        """
+        Returns the appropriate edit_handler for this modeladmin class.
+        edit_handlers can be defined either on the model itself or on the
+        modeladmin (as property edit_handler or panels). Falls back to
+        extracting panel / edit handler definitions from the model class.
+        """
+        if hasattr(self.model, 'edit_handler'):
```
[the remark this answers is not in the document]
mr-stateradio: @ababic fixed in https://github.com/wagtail/wagtail/pull/4900/commits/ddf99d408831dac5587b11ccfaf66481629f5519

wagtail/tests/modeladmintest/wagtail_hooks.py
mr-stateradio: @ababic fixed in https://github.com/wagtail/wagtail/pull/4900/commits/61d2e4dbdd7049a9e31150f8782d74c344a60c7a

wagtail/contrib/modeladmin/options.py
mr-stateradio: @ababic fixed in https://github.com/wagtail/wagtail/pull/4900/commits/d3baf096e35d840a5c277a7c1fb2a473aa9804f2

I think the structure and responsibilities of the two `get_edit_handler` is much clearer now!

mr-stateradio:
@ababic I fixed all your comments. I hope you are happy with them!

We have a merge conflict now, do you have an idea what happened here? It looks like somebody worked in parallel on the `ModelFormView.get_edit_handler` method.

```
    def get_edit_handler(self):
<<<<<<< feature-modeladmin-panel-definitions
        edit_handler = self.model_admin.get_edit_handler(
            instance=self.get_instance(), request=self.request
        )
        return edit_handler.bind_to_model(self.model_admin.model)
=======
        if hasattr(self.model, 'edit_handler'):
            edit_handler = self.model.edit_handler
        else:
            fields_to_exclude = self.model_admin.get_form_fields_exclude(request=self.request)
            panels = extract_panel_definitions_from_model_class(self.model, exclude=fields_to_exclude)
            edit_handler = ObjectList(panels)
        return edit_handler.bind_to(model=self.model)
>>>>>>> master
```

gasman:
@mr-stateradio See #4749 (and in particular the release note https://github.com/wagtail/wagtail/blob/master/docs/releases/2.5.rst#edithandlerbind_to_model-and-edithandlerbind_to_instance-deprecated ) - `bind_to_model` has now been replaced by `bind_to(model=...)`.

mr-stateradio:
@gasman thanks for the hint. I resolved the conflict by replacing `bind_to_model` with `bind_to(model=...)`

wagtail/contrib/modeladmin/tests/test_modeladmin_edit_handlers.py
```diff
@@ -0,0 +1,136 @@
...
+        # verify that form fields are returned which have been defined
+        # in Person.edit_handler
+        response = self.client.get('/admin/modeladmintest/visitor/create/')
+
+        self.assertIn('last_name', response.content.decode('UTF-8'))
+        self.assertIn('phone_number', response.content.decode('UTF-8'))
+        self.assertIn('address', response.content.decode('UTF-8'))
+        self.assertNotIn('first_name', response.content.decode('UTF-8'))
+        self.assertEqual(
+            [ii for ii in response.context['form'].fields],
```
ababic: @mr-stateradio A very small thing, but for readability's sake, I think it could be good to use a more descriptive variable name than `ii` here. `field_name` might make a good substitute.

wagtail/contrib/modeladmin/tests/test_modeladmin_edit_handlers.py
```diff
@@ -0,0 +1,136 @@
...
+        self.assertIn('first_name', response.content.decode('UTF-8'))
+        self.assertNotIn('last_name', response.content.decode('UTF-8'))
+        self.assertIn('phone_number', response.content.decode('UTF-8'))
+        self.assertNotIn('address', response.content.decode('UTF-8'))
+        self.assertEqual(
+            [ii for ii in response.context['form'].fields],
+            ['first_name', 'phone_number']
+        )
+
+    def test_model_admin_edit_handler(self):
```
ababic: @mr-stateradio Each of these tests seems to be quite clearly split into 2 parts:

1. Post the form and check the object that gets created.
2. Loads the 'create' view and checks what fields are available. 

I would suggest that part 2 is really the most useful part here, and is probably really all that is needed. If you have any experience of using the [mock](https://docs.python.org/3/library/unittest.mock.html) library in Python, then an added bonus would be to show that both `ModelFormView.get_edit_handler()` and `ModelAdmin.get_edit_handler()` are called in order to render the page - But, I certainly wouldn't class that as necessary.

The state of a newly created instance after posting could potentially be affected by a number of things other than what `edit_handler` was used by the view (for example: default field values / custom clean/save logic on the model), and without going and checking those things, it's difficult to know what the expected output should be.

wagtail/contrib/modeladmin/tests/test_modeladmin_edit_handlers.py
```diff
@@ -0,0 +1,136 @@
...
+            last_name="Doe",
+            phone_number="+123456789",
+            address="123 Main St Anytown"
+        ).count(), 1)
+
+        # verify that form fields are returned which have been defined
+        # in Person.edit_handler
+        response = self.client.get('/admin/modeladmintest/visitor/create/')
+
+        self.assertIn('last_name', response.content.decode('UTF-8'))
```
ababic: I would also suggest that checking `response.context['form'].fields` exclusively in each of these tests would make your intentions quite a lot clearer... as the form is really the most indicative thing we have to test that the right `edit_handler` or `fields` were used.

wagtail/contrib/modeladmin/tests/test_modeladmin_edit_handlers.py
mr-stateradio: @ababic I agree and will fix it. I copied the structure and some variable names from `test_simple_modeladmin.TestCreateView.test_create`

mr-stateradio: @ababic  fixed in https://github.com/wagtail/wagtail/pull/4900/commits/57f838b2393b145bfe1655b8e01b3e01e4bdd9c4

wagtail/contrib/modeladmin/tests/test_modeladmin_edit_handlers.py
mr-stateradio: @ababic I agree and will fix it. This is also an approach I copied from `test_simple_modeladmin.TestCreateView.test_create`. But I agree that it makes the intention of the test case a lot clearer if we only check the form fields.

mr-stateradio: @ababic fixed in https://github.com/wagtail/wagtail/pull/4900/commits/68f3b9829cede196926c48a694432eb7e2a0dbe8

wagtail/contrib/modeladmin/tests/test_modeladmin_edit_handlers.py
mr-stateradio: @ababic I restricted the test cases to check only for the fields available in the 'create view', and used the mock Library to verify the appropriate `get_edit_handler` methods are called. Can you check the file `test_modeladmin_edit_handlers.TestExtractPanelDefinitionsFromModelAdmin` if we still have some potential for improvement or if we are good to go :)

ababic: @mr-stateradio This is better, thank you. Although, I'm not sure that we need to confirm that both methods are called for each different URL as a separate test. I think we can assume that the view works the same way, regardless of what the model is. I was just thinking once, at the beginning of the class, to help provide some context to the reader in regards to the other tests. Sorry if I could've been clearer there.

ababic: @mr-stateradio I could be wrong, but it appears that we're not really testing that values on the ModelAdmin class are preferred over values on the model when both are defined, which I think is probably an important detail of the implementation, and should have a test to help guard against regressions. 

Would you be okay to add something there? I'm guessing it might be easier to take a more  'unit test like' approach (create a `ModelAdmin` instance in the test, change it's attributes, play with the `get_edit_handler()` method directly).

mr-stateradio: @ababic I just posted a newer version of the `test_model_form_view_edit_handler_called` and `test_model_form_view_edit_handler_called` since the version you reviewed was not Python 3.5 compatible and had some other issues. If you check the new version you will see that I check different things there, please check again if you are fine with them now.

You are absolutely right regarding your request for adding a test to verify that values on the ModelAdmin class are preferred over values on the model. Will provide another test case for it. Will try to make this new test case more "unit test like", but will keep the others for now as they are. Maybe we adapt the other ones in the next feedback loop :)

ababic: @mr-stateradio Thank you, yes. These are much better now.

mr-stateradio:
@ababic I provided another test case to verify modeladmin.panels are prefered over model.panels: https://github.com/wagtail/wagtail/pull/4900/commits/ddec349d32e0bc9f48d2d6cdc70e05681852fc22

Are you fine with the new test case?  Do you have some more ideas for improvement?

mr-stateradio:
@ababic Yes I understand. Will provide a new version of this test case.

ababic:
Thanks @mr-stateradio! I'm happy this works as intended and is sufficiently tested, so 👍

@gasman, do you have any thoughts?

mr-stateradio:
@ababic Great! Thanks for your thorough feedback and suggestions for improvement. :pray:

mr-stateradio:
@ababic Yes, definitely. I think it was a smooth process!

ababic:
@mr-stateradio. I've just realised that there are no documentation changes yet. Would you like to try adding to the following page?
 https://github.com/wagtail/wagtail/blob/master/docs/reference/contrib/modeladmin/create_edit_delete_views.rst

mr-stateradio:
@ababic ah yes, you are right. Will provide some documentation changes today.

ababic:
@mr-stateradio okay, great. It'll probably best to leave the existing information in-tact as much as possible, and try to add the new information in separately for now. There's handy RST directive that can be used to mark the new functionality/content, which we should probably make use of here:
http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded

mr-stateradio:
@ababic I provide some docs in https://github.com/wagtail/wagtail/pull/4900/commits/2e5faa797c219012aaf5c800d8aa549d3e5cad81

image

What do you think?

docs/reference/contrib/modeladmin/create_edit_delete_views.rst
```diff
@@ -58,6 +58,38 @@ Or alternatively:
...
+Or alternatively:
+
+
+.. code-block:: python
+
+    class BookAdmin(ModelAdmin):
+        model = Book
+
+        custom_panels = [
+            FieldPanel('title', classname='fn'),
```
ababic: @mr-stateradio A small thing, but for maximum clarity, I think dropping the `classname` arguments for both these panels would be good too.

ababic:
@mr-stateradio I think the only thing missing now is a method-specific entry for `ModelAdmin.get_edit_handler()`. You okay to add that in? (this part of the docs doesn't use `autodoc`)

mr-stateradio:
@ababic You mean a `method-specific` entry like provided in this commit? https://github.com/wagtail/wagtail/pull/4900/commits/da3e04e7420e318ecdffda1470e5a8bb2ebb7b60

ababic:
@mr-stateradio Yes, that's good. Although, when discussing a method, I think the convention is to use closed brackets in the heading, for example: https://docs.wagtail.io/en/v2.4/reference/contrib/modeladmin/indexview.html#modeladmin-get-queryset

mr-stateradio:
@ababic added the closed brackets in https://github.com/wagtail/wagtail/pull/4900/commits/25ebf408d91637b6011ca50d78e069aa8dabb1da plus added a `Must return`.

mr-stateradio:
@ababic Adjusted your suggestions in https://github.com/wagtail/wagtail/pull/4900/commits/57369798a00e1b4b7b87cd19b0693177aaf8d600

Not sure about where to place the `.. versionadded:: 2.5` directive. I added it to the explanation of the `get_edit_handler` method since above it looked strangely.

It gets rendered like this now: 

image

ababic:
Thanks @mr-stateradio. I think the best thing to do in this scenario is to use the directive directly below the heading, without a description (specifying the version number only).

mr-stateradio:
@ababic agreed, looks better now:

image

ababic:
@mr-stateradio I think we're just missing the double backticks around `wagtail.admin.edit_handlers.ObjectList`. This is the last one from me, I promise!  :)

mr-stateradio:
@ababic double backticks provided in https://github.com/wagtail/wagtail/pull/4900/commits/45b89b46cceb0ee2bbfa9dbf7d3b62979a8fe4fa

gasman:
Merged in 4e8f197a9d0ac0b0d5237cf311e505764441f71d - thanks @mr-stateradio!

mr-stateradio:
You are welcome! Nice to see this PR landing in wagtail! We are looking forward to the new release.

PR #4901, opened 2018-11-14
Fixed reference to LoginView in docs.

sir-sigurd:
`LoginView` is used since a657a75cd7822bd29913f0a48e37eb97ecfd4175.


gasman:
Thanks again!

Issue #2655, opened 2016-05-31
Table block doesn't save alignment data


alexgleason:
Here's an example of what table JSON data looks like (from a StreamField form submission):

```
body-10-value:{"data":[["hello","world","sup"],["hii","yoooo",null],[null,null,null]],"first_row_is_table_header":false,"first_col_is_header":false}
```

However, you can right-click any cell, and choose Alignment > Left / Right / Center, among some vertical options, too. This data also needs to be saved into the JSON so it can be templated.

It's probably too late for every cell to have its own "options" attribute, so I'm guessing alignment will need to be passed in as a map, like

```
"alignments": [{"cell": "0,1", "alignment": "left"}, {"cell": "2,2", "alignment": "center"}]
```

etc

gasman:
I guess there's a question here, which ties in with the whole content vs presentation thing, about whether alignment is really part of the data model. It probably depends on the context: for a pure data table, such as one that gets rendered as a graph on the front-end, it doesn't belong there (and we should probably disable the alignment options entirely in that case); for a table displayed to the user, it might make sense to set alignments on a per-column or row basis. I can't think of a legitimate use for setting arbitrary alignment on a per-cell basis - I suspect that's heading too far into the domain of using tables for layout, which we want to avoid.

AtomicSpark:
Although it is implemented in the handsontable code and currently non-functioning in Wagtail, HTML align attributes are depreciated in HTML5 and [should not be used](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/table#attr-align). I vote that we disable align features in handsontable if possible, before this turns into a Wagtail image debate on whether or not we inject surprise CSS into templates.

alexgleason:
Very good points. I vote to disable alignment entirely.

Proper-Job:
I'm working on https://github.com/torchbox/wagtail/issues/2687 anyway right now and I'll do a PR for this.

Proper-Job:
https://github.com/torchbox/wagtail/pull/2701

robslotboom:
I agree with this but why not replace the html alignement attributes by alignment classes.

PR #4906, opened 2018-11-15
Alignment support for table block

smendes:
This PR is split into two parts:
- Update Handsontable to version 6.2.2
- Adding support for saving cell classes

I've found the issue https://github.com/wagtail/wagtail/issues/2655 and while I understand the motivation for uncoupling data and formatting, it would seem that being able to specify a class for the data wouldn't violate this desire.

For Twilio, we do need the ability for editors to control the alignment of some cells, but agree with the data vs. formatting principle as well. It seems that a fair compromise would be to allows specifying a "class" for the data and then allow the various presentation templates for the table handle that specific class how they see fit.

For cell-specific information Handsontable stores everything in the metadata. This implementation only keeps the `className` part of it. That attribute stores the HTML class to apply to a given cell. With this patch, table blocks get rendered with those custom classes (`htLeft`, `htRight`, etc..) but without any style rules applied to those classes. Furthermore, the alignment option for the context menu isn't activated by default. This offers the possibility for a developer to activate it, but make it their responsibility to style those classes.

- [x] For new features: Has the documentation been updated accordingly?


dprothero:
Hey team! @smendes works with us on the Twilio Developer Education team and we're keen for some kind of method for allowing different cells to be styled differently based on the desires of the editor using the CMS. Let us know what you think of this PR and if there's a different approach you'd prefer for this.

dprothero:
@gasman Do you have some time to take a look at this?

dprothero:
@gasman When do you think we might be able to expect some feedback? Is there something we can do with the PR to make your life easier?

gasman:
Hi @smendes - please can you add a section to https://github.com/wagtail/wagtail/blob/master/docs/reference/contrib/table_block.rst detailing how a developer should enable this?

dprothero:
@gasman How do the docs look?

gasman:
Merged in 51d8cac6f4f28c1b69894303122e9ff68edb12e7 + parents - thanks @smendes!

Have just made one addition in 722cd8be100f2915d76f07b14c0ce689a2fb0356 - to specify the expected HTML output directly in tests, since it doesn't really make sense to check the output against the output of another function performing the same job. (Not an issue with the current PR - the existing tests were doing it too.)

(Also, as a memo for future reference - I wasn't able to confirm that the `language` parameter is doing the right thing, as it didn't appear to have any effect on the context menu, before or after applying this PR. However, I confirmed that the Django language setting _is_ being passed on in that parameter, in the correct format according to https://handsontable.com/docs/6.2.2/Options.html#language.)

PR #4912, opened 2018-11-19
Hide "delete snippets" button when no snippets selected

### LINKED ISSUE #4904: Bulk delete button on snippets
The snippets interface has gained a bulk delete button recently (https://screenshots.firefox.com/Tfj3Gk0okTr0GMqr/localhost)

This button looks very out of place to me, and has a red hover colour when no snippets are selected (which seems weird to me as the button doesn't do anything in this state).

Could the button be hidden when no snippets are selected? and also moved above the snippets listing?

kaedroho:
Fixes #4904 

Removes the ugly disabled button from the header of the snippets section. The button is displayed when the user selects a snippet.

This is more consistent with the interface we have for form submissions (except for the button being in a different place). Also, the hiding of action buttons behaviour is used by other popular online tools such as Gmail so shouldn't be too surprising for users.

### DIFF
--- a/wagtail/snippets/static_src/wagtailsnippets/js/snippet-multiple-select.js
+++ b/wagtail/snippets/static_src/wagtailsnippets/js/snippet-multiple-select.js
@@ -19,13 +19,13 @@ var updateDeleteButton = function(anySelected, newState) {
     });
     if (anySelected) {
         // enable button and add url
-        $deleteButton.css('visibility', 'visible');
+        $deleteButton.removeClass('visuallyhidden');
         var url = $deleteButton.data('url');
         url = url + $.param({id: ids}, true);
         $deleteButton.attr('href', url);
     } else {
         // disable button and remove url
-        $deleteButton.css('visibility', 'hidden');
+        $deleteButton.addClass('visuallyhidden');
         $deleteButton.attr('href', null);
     }
 };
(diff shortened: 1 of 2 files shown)


wagtail/snippets/static_src/wagtailsnippets/js/snippet-multiple-select.js
```diff
@@ -19,13 +19,13 @@ var updateDeleteButton = function(anySelected, newState) {
     });
     if (anySelected) {
         // enable button and add url
-        $deleteButton.removeClass('disabled');
+        $deleteButton.css('visibility', 'visible');
```
zerolab: I suggest using the `.visuallyhidden` class - https://github.com/wagtail/wagtail/blob/master/wagtail/admin/static_src/wagtailadmin/scss/core.scss#L65

zerolab: This way it will be consistent with other parts of the UI.

kaedroho: Will do, thanks for pointing that out!

gasman:
Merged in 4abfbf89a0122d67fa8be0f076c97162c9167a88

PR #4915, opened 2018-11-20
Bump draftjs_exporter version to fix entities export bug

thibaudcolas:
This bumps the `draftjs_exporter` dependency requirement of Wagtail to [v2.1.5](https://github.com/springload/draftjs_exporter/blob/master/CHANGELOG.md#v215), which fixes an export bug when entities (say, links) are adjacent in the content.

Here is an example of where the problem occurs:

```html
<!-- Authoring the following in the editor, -->
<a href="/">X</a><a href="/">W</a>

<!-- Will be saved instead as, -->
<a href="/">X</a>W<a href="/"></a>
```

I wouldn't expect this to be a common issue for most Wagtail sites since it's rare/unwanted to have adjacent links like that, but for sites that have custom entities (say a custom emoji picker) it can definitely happen. See https://github.com/springload/draftjs_exporter/issues/106 for further details.

---

I'm not entirely sure whether the bump is needed (the new version was already in Wagtail's range), but I imagine people using lockfiles for their dependencies will benefit from seeing Wagtail's lower range for `draftjs_exporter` go up. And also having the mention in the release notes, for people who would want to assess whether their content has this issue.

I haven't added tests for this in Wagtail's codebase (there are tests in the exporter's), but I can if this seems useful.

* ~[ ] For Python changes: Have you added tests to cover the new/fixed behaviour?~
* ~[ ] For front-end changes: Did you test on all of Wagtail’s supported browsers? **Please list the exact versions you tested**.~
* ~[ ] For new features: Has the documentation been updated accordingly?~


loicteixeira:
As you pointed out, this bump isn't mandatory since this new version is included in the existing range but bumping the lower range is still a good thing to do as it will bring various fixes and performances improvements to everybody. Plus it doesn't have any breaking changes (it does add a required parameter to the `render_decorators` method but it isn't part of the public API) or require users to act on anything so that's all good.

As for the implications it might have, presumably it won't fix *nor worsen* already malformed data right? Let me explain, you created `<a href="/">X</a><a href="/">W</a>` but it was saved in DB Format as `<a href="/">X</a>W<a href="/"></a>`. So next time around, it will be converted back to ContentState for use in the editor with the `W` before of the `a` tag. Saving it again will not change anything to it (despite the version bump).

Otherwise I can't think of anything else. Wagtail defines links, images, documents, embed and blockquotes as entities, some are self closing some aren't. Maybe it's worth confirming that it's working with both types doubled and mixed (both order) unless you're confident in the draftjs_exporter fix (I don't have time to dig into the fix code sorry).

thibaudcolas:
Hmm, I'm quite confident in the fix but I'll do more tests specifically within the context of Wagtail so I can answer your questions confidently as well 😛.

gasman:
Merged in a3dc9ed4773857c0d09d86be5040bc8f3c0086e3.

PR #4923, opened 2018-11-23
A small refactoring to the file whitelist.py

srinivasreddy:


allcaps:
I don't see what you win here. Care to explain what makes this better? I believe readability counts.

PR #4926, opened 2018-11-26
Respect CSS precedence rules in HTMLRuleset

gasman:
Fixes #4527

Enhances the HTMLRuleset engine used in dbHTML-to-contentState conversion, so that more specific CSS rules are matched with higher priority. This makes it possible to define conversion rules for elements such as `<p class="intro">` that don't get overruled by the default `<p>` rule.

### DIFF
--- a/wagtail/admin/rich_text/converters/html_ruleset.py
+++ b/wagtail/admin/rich_text/converters/html_ruleset.py
@@ -19,7 +19,7 @@ class HTMLRuleset():
     'a[linktype="page"]' = matches any <a> element with a 'linktype' attribute equal to 'page'
     """
     def __init__(self, rules=None):
-        # mapping of element name to a list of (attr_check, result) tuples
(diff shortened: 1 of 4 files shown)


gasman:
Thanks @thibaudcolas! Agreed that we could really do with some more documentation here, but doing a good job of that will probably require more reworking of the existing docs than we can reasonably do as an add-on to this PR. The Extending Draftail page is written in a "howto" style, with lots of concepts being introduced at once before going on to progressively more complex examples - logically it makes sense to cover this at the point where we introduce `register_converter_rule` in the first example, but it'll be too overwhelming to readers if we break into reference-style documentation to describe the full functionality. (We do at least have `span[data-stock]` in the last example...) As such, I'd say this is out of scope for this PR.

gasman:
Merged in 4a5036839b4abc0f6c4b24d1e053b3ac53b067bc

PR #4927, opened 2018-11-27
Documentation on rich text internals

gasman:
Add documentation for the aspects of rich text handling which are not specific to the Draftail or Hallo editors: the in-database data format, front-end HTML rewriting, the feature registry, and the interactions between editor widgets and converter classes. The motivation for this comes from several directions:

* it provides context about Wagtail design decisions both for implementers ("why does my `<p class="intro">` get stripped out?") and core devs ("why do we have two separate but similar database-format-to-HTML converters?")
* it's necessary information for people implementing their own editor widgets (such as wagtailtinymce) and defining new link/embed types, which is currently either missing or only briefly touched on in the Draftail / Hallo customisation docs
* wagtail.images, wagtail.documents and wagtail.embeds all make extensive use of these APIs, which violates the "dogfooding" principle that these apps should be on equal standing with third-party apps and not use internal APIs
* #4702 is due to make breaking changes to link/embed handlers; in theory this is OK because it's currently an internal API, but in practice there's bound to be lots of use of it in the wild (from people cloning wagtailimages/docs) and so it's appropriate for it to go through formal deprecation. Documenting the API is a prerequisite for that, so that we can map out what is and isn't going to be broken in the transition period, and tell users what they need to change.

### DIFF
--- a/docs/advanced_topics/customisation/rich_text_internals.rst
+++ b/docs/advanced_topics/customisation/rich_text_internals.rst
@@ -5,7 +5,7 @@ At first glance, Wagtail's rich text capabilities appear to give editors direct 
 
  * The editor interface needs to filter out certain kinds of unwanted markup; this includes malicious scripting, font styles pasted from an external word processor, and elements which would break the validity or consistency of the site design (for example, pages will generally reserve the ``<h1>`` element for the page title, and so it would be inappropriate to allow users to insert their own additional ``<h1>`` elements through rich text).
  * Rich text fields can specify a ``features`` argument to further restrict the elements permitted in the field - see :ref:`rich_text_features`.
- * Enforcing a subset of HTML helps to keep presentational markup out of the database, making the site more maintainable, and making it easier to repurpose site content (including, potentially, producing non-HTML output such as `DocBook <https://docbook.org/>`_).
+ * Enforcing a subset of HTML helps to keep presentational markup out of the database, making the site more maintainable, and making it easier to repurpose site content (including, potentially, producing non-HTML output such as `LaTeX <https://www.latex-project.org/>`_).
  * Elements such as page links and images need to preserve metadata such as the page or image ID, which is not present in the final HTML representation.
 
 This requires the rich text content to go through a number of validation and conversion steps; both between the editor interface and the version stored in the database, and from the database representation to the final rendered HTML.
--- a/docs/advanced_topics/customisation/rich_text_internals.rst
+++ b/docs/advanced_topics/customisation/rich_text_internals.rst
@@ -1,0 +1,205 @@
+Rich text internals
+===================
+
+At first glance, Wagtail's rich text capabilities appear to give editors direct control over a block of HTML content. In reality, it's necessary to give editors a representation of rich text content that is several steps removed from the final HTML output, for several reasons:
+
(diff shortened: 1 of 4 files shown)


docs/advanced_topics/customisation/rich_text_internals.rst
```diff
@@ -0,0 +1,205 @@
+Rich text internals
+===================
+
+At first glance, Wagtail's rich text capabilities appear to give editors direct control over a block of HTML content. In reality, it's necessary to give editors a representation of rich text content that is several steps removed from the final HTML output, for several reasons:
+
+ * The editor interface needs to filter out certain kinds of unwanted markup; this includes malicious scripting, font styles pasted from an external word processor, and elements which would break the validity or consistency of the site design (for example, pages will generally reserve the ``<h1>`` element for the page title, and so it would be inappropriate to allow users to insert their own additional ``<h1>`` elements through rich text).
+ * Rich text fields can specify a ``features`` argument to further restrict the elements permitted in the field - see :ref:`rich_text_features`.
+ * Enforcing a subset of HTML helps to keep presentational markup out of the database, making the site more maintainable, and making it easier to repurpose site content (including, potentially, producing non-HTML output such as `DocBook <https://docbook.org/>`_).
```
tomdyson: docbook feels like a strange example - do people still use it? Perhaps we should just drop 'such as...'.

docs/advanced_topics/customisation/rich_text_internals.rst
```diff
@@ -0,0 +1,205 @@
...
+
+.. code-block:: html
+
+    <img alt="A pied wagtail" class="richtext-image left" height="294" src="/media/images/pied-wagtail.width-500_ENyKffb.jpg" width="500">
+
+Again, the ``embedtype`` attribute identifies a rule that shall be used to rewrite the tag. All tags other than ``<a linktype="...">`` and ``<embed embedtype="..." />`` are left unchanged in the converted HTML.
+
+A number of additional constraints apply to ``<a linktype="...">`` and ``<embed embedtype="..." />`` tags, to allow the conversion to be performed efficiently via string replacement:
+
+ * The tag name and attributes must be lower-case;
```
tomdyson: I love semicolons but I think they look a bit old-school in ULs :)

docs/advanced_topics/customisation/rich_text_internals.rst
gasman: I think Docbook works well as an illustrative example even if it's not a real-world one for most people - it's fairly easy to see how you'd convert rich text content to it, but only if you've been scrupulous about keeping that content as semantic markup rather than an anything-goes bucket of HTML. Hopefully it also reinforces the idea of links / bold formatting etc as being abstract entities of which HTML is just one possible representation - which then helps in understanding concepts like rich text features and contentState later on. Not sure if there are other examples that can communicate that while retaining a bit more real-world-ness - LaTeX possibly?

docs/advanced_topics/customisation/rich_text_internals.rst
gasman: Fair point! Removed in 96d81d78f1f178ed42fd90803ba61e6ba4f603ed.

docs/advanced_topics/customisation/rich_text_internals.rst
tomdyson: I agree that it's a relevant example, I'm just not sure that many people will know what it is. I was going to suggest LaTeX.

gasman: Have now updated to reference LaTeX instead.

tomdyson: LGTM

Issue #3368, opened 2017-02-14
Revision comparison view should compare StreamFields structurally


gasman:
Currently the revision comparison view [treats StreamFields as equivalent to rich text fields](https://github.com/wagtail/wagtail/blob/5aa8e00d77d20060f75b7ca65e28ef502c54245c/wagtail/wagtailadmin/compare.py#L60), and displays comparisons based on the rendered (and tag-stripped) HTML. This is far from ideal - StreamFields can contain a vast amount of structured content, not necessarily text-based (for example, the current comparison would not show a change to an ImageChooserBlock) and the comparison view should reflect that.

Doing the comparison _well_ would probably be a major undertaking (e.g. ensuring that a change to a single field of a StructBlock is displayed as a change to that one field, rather than a replacement of the entire field) but as a first pass, they should be handled similarly to InlinePanels so that changes are displayed in terms of insertions / deletions / updates of top-level blocks.

atbaker:
Just chiming in with a +1 here.

I wonder if a smarter compare would also improve the feature's performance? 90%+ of our content is in StreamFields and comparing revisions often takes a long time for us - sometimes longer than 90 seconds.

chosak:
It looks like this will be fixed by #4933 (ping @kaedroho), but I encountered a related bug that seems to fit here due to the use of rendered block HTML in comparisons.

Specifically, custom StreamField blocks that rely on parent template context variables (for example `page`) don't have those variables when they are rendered in the diff view. Consider a custom StreamField block that renders differently depending on the type of page that it is on. The StreamField docs [imply](http://docs.wagtail.io/en/v2.0/topics/streamfield.html#template-rendering) that using `{% include_block %}` and relying on parent template variables is a valid thing to do in block templates. So a developer might write their block template (or custom `get_context`) assuming that the `page` variable will be available, given that the block will only ever be inserted and rendered on pages that use `{% include_block %}`.

Unfortunately, because the `__str__`-based block rendering ([here](https://github.com/wagtail/wagtail/blob/5e5484d4d7d3fec0b860826a07903165ed6dfb31/wagtail/core/blocks/stream_block.py#L466)) ends up rendering blocks with an empty context ([here](https://github.com/wagtail/wagtail/blob/5e5484d4d7d3fec0b860826a07903165ed6dfb31/wagtail/core/blocks/base.py#L422)), that variable doesn't exist, and can cause e.g. a `ValueError` on the compare view.

[...]

PR #4933, opened 2018-11-29
Diffing StreamFields

kaedroho:
This PR implements diffing of StreamFields. It utilises the block IDs introduced in https://github.com/wagtail/wagtail/pull/3593 to work out which blocks have been created, deleted and moved.

I've currently implemented comparison logic for the rich text, image, struct and stream blocks. Other block types fall back to diffing on their HTML representation (as what happens now).

Before:

streamfield-diff-before

After:

streamfield-diff-after

TODO:

 - [x] Tests
 - [ ] ~Indicate moved blocks~ will leave for another PR as I will also refactor diffing inline panels as well
 - [x] Fallback for diffing with revision without block IDs
 - [ ] ~ListBlock~ will leave for another PR as I will also refactor diffing inline panels as well

### DIFF
--- a/wagtail/admin/compare.py
+++ b/wagtail/admin/compare.py
@@ -2,10 +2,12 @@ import difflib
 
 from bs4 import BeautifulSoup
 from django.utils.encoding import force_text
-from django.utils.html import escape
+from django.utils.html import escape, format_html, format_html_join
 from django.utils.safestring import mark_safe
 from django.utils.text import capfirst
 from django.utils.translation import ugettext_lazy as _
+
+from wagtail.core import blocks
 
 
 class FieldComparison:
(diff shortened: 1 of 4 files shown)


kaedroho:
This is ready for review now, I'd like to make improvements to the diffiing of inline panels and make it more generic so it could be used for streamfield listblocks as well. I will like to do this in a separate PR as it would be a lot of work.

wagtail/admin/compare.py
```diff
@@ -55,8 +57,155 @@ def htmldiff(self):
...
+                if current_index == index:
+                    comparisons.insert(i, comparison_class(block.block, block.value, None))
+                    block_inserted = True
+                    break
+
+                current_index += 1
+
+            # Deleted block was from the end
+            if not block_inserted:
+                comparisons.append(comparison_class(block.block, block.value, None))
```
chosak: Can avoid a bit of DRY by moving the definition of `comparison_class(block.block, block.value, None)` up to the top of the outer loop.

wagtail/admin/compare.py
```diff
@@ -55,8 +57,155 @@ def htmldiff(self):
...
+class RichTextBlockComparison(BlockComparison):
+    def htmldiff(self):
+        return diff_text(
+            BeautifulSoup(force_text(self.val_a), 'html5lib').getText(),
+            BeautifulSoup(force_text(self.val_b), 'html5lib').getText()
+        ).to_html()
+
+
+class StructBlockComparison(BlockComparison):
+    def htmldiff(self):
```
chosak: The rendering of nested StructBlocks is a little bit inconsistent.

Consider these blocks:

```py
class PersonBlock(blocks.StructBlock):
    name = blocks.CharBlock()
    email = blocks.EmailBlock()

class TeamBlock(blocks.StructBlock):
    name = blocks.CharBlock()
    leader = PersonBlock()
```

Then say a page has two PersonBlocks and an editor adds a new TeamBlock:

image

Unfortunately the diff doesn't indent the nested PersonBlock (seems like NAME and EMAIL should be indented a bit under LEADER):

image

Additionally, let's say the PersonBlock has a custom template:


```py
class PersonBlock(blocks.StructBlock):
    name = blocks.CharBlock()
    email = blocks.EmailBlock()

    class Meta:
        template = 'blocks/person_block.html'
```

```html
<!-- blocks/person_block.html -->
<a href="mailto:{{ value.email }}">Email {{ value.name }}</a>
```

Now the nested block is still using its HTML representation (note the use of the custom template there that generates an email link), but the top-level block is using the basic rendering:

image

This also unfortunately means that the bug mentioned on https://github.com/wagtail/wagtail/issues/3368#issuecomment-451496319 about custom block templates requiring template context can still cause errors when comparing.

wagtail/admin/compare.py
```diff
@@ -55,8 +57,155 @@ def htmldiff(self):
...
+            if comparison.is_new():
+                classes.append('addition')
+                block_rendered = comparison.block.render_basic(comparison.val_b)
+            elif comparison.is_deleted():
+                classes.append('deletion')
+                block_rendered = comparison.block.render_basic(comparison.val_a)
+            elif comparison.has_changed():
+                block_rendered = comparison.htmldiff()
+            else:
+                block_rendered = comparison.block.render_basic(comparison.val_a)
```
chosak: The use of `comparison.block.render_basic` for everything except the `has_changed` case means that comparison class `htmldiff` methods can't provide markup for the addition/deletion/same cases.

For example, [this logic](https://github.com/wagtail/wagtail/blob/5e5484d4d7d3fec0b860826a07903165ed6dfb31/wagtail/images/templates/wagtailimages/widgets/compare.html#L15) in the image comparison template (used [here](https://github.com/wagtail/wagtail/pull/4933/files#diff-70247e933fb4da7e546dcaf823810235R36) in this PR) doesn't ever get called, since the template isn't rendered in those cases. Instead, the `block.render_basic` rendering is used, which just renders the image at its original size. This isn't great in cases where the image is very large.

When an image is changed, the nice diff view is shown:

image

But when the image is added or deleted, it's displayed at its original size:

image

Could the `htmldiff` methods _always_ generate the rendering to be used? Even when there's no difference? They'd just render the current state of the block. Perhaps the base `BlockComparison` could provide some kind of fallback that inherited comparison classes use?

wagtail/admin/compare.py
```diff
@@ -55,8 +57,155 @@ def htmldiff(self):
...
+    if hasattr(block, 'get_comparison_class'):
+        return block.get_comparison_class()
+    elif isinstance(block, blocks.CharBlock):
+        return CharBlockComparison
+    elif isinstance(block, blocks.RichTextBlock):
+        return RichTextBlockComparison
+    elif isinstance(block, blocks.StructBlock):
+        return StructBlockComparison
+    else:
+        # As all stream field blocks have a HTML representation, fall back to diffing that.
```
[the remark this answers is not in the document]
kaedroho: Yep, that's right.

wagtail/admin/compare.py
```diff
@@ -55,8 +57,155 @@ def htmldiff(self):
         ).to_html()
 
 
-class StreamFieldComparison(RichTextFieldComparison):
-    pass
+def get_comparison_class_for_block(block):
+    if hasattr(block, 'get_comparison_class'):
```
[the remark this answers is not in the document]
kaedroho: Maybe we should cover comparisons in a whole separate document since adding comparisons for edit handlers is very similar.

wagtail/admin/compare.py
```diff
@@ -55,8 +57,155 @@ def htmldiff(self):
...
+
+    def htmldiff(self):
+        # Our method for diffing streamfields relies on the blocks in both revisions having UUIDs.
+        # But as UUIDs were added in Wagtail 1.11 we can't compare revisions that were created before
+        # that Wagtail version.
+        if self.has_block_ids(self.val_a) and self.has_block_ids(self.val_b):
+            return StreamBlockComparison(self.field.stream_block, self.val_a, self.val_b).htmldiff()
+        else:
+            # Fall back to diffing the HTML representation
+            return diff_text(
```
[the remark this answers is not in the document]
kaedroho: I'm about to make some improvements to RichTextBlockComparison but I'd still like the fallback for streamfields to keep using this simpler method since those improvements won't be useful for non-rich-text fields.

wagtail/admin/compare.py
```diff
@@ -104,17 +109,27 @@ def htmldiff(self):
 
 
 class StructBlockComparison(BlockComparison):
+    def htmlvalue(self, val):
+        htmlvalues = {}
```
chosak: [Until Python 3.7](https://mail.python.org/pipermail/python-dev/2017-December/151283.html), preservation of dict insertion order is not guaranteed, so iterating over child blocks and then doing `htmlvalues.items()` below isn't guaranteed to generate the correct order. Would it be better to use a list of `(name, html)` tuples here instead?

wagtail/admin/compare.py
```diff
@@ -172,19 +188,19 @@ def htmldiff(self):
...
-                block_rendered = comparison.block.render_basic(comparison.val_a)
+                block_rendered = comparison.htmlvalue(comparison.val_a)
             elif comparison.has_changed():
                 block_rendered = comparison.htmldiff()
             else:
-                block_rendered = comparison.block.render_basic(comparison.val_a)
+                block_rendered = comparison.htmlvalue(comparison.val_a)
 
             classes = ' '.join(classes)
             comparisons_html.append(f'<div class="{classes}">{block_rendered}</div>')
```
chosak: Travis [build is failing](https://travis-ci.org/wagtail/wagtail/builds/476294238?utm_source=github_status&utm_medium=notification) for Python < 3.6 because [f-strings aren't available](https://www.python.org/dev/peps/pep-0498/) in those older versions.

wagtail/admin/compare.py
kaedroho: A list of tuples is definitely a more appropriate type to use here too, thanks!

wagtail/admin/compare.py
kaedroho: Oops, I got into this habbit now.. Thanks!

chosak:
@kaedroho thank you! Merged with release notes in 6e2b8c2.

PR #4935, opened 2018-11-30
Add required kwarg

brylie:
I had trouble following the tutorial, as I was getting errors until I added the `to=` in front of the related model.

### DIFF
--- a/docs/getting_started/tutorial.rst
+++ b/docs/getting_started/tutorial.rst
@@ -587,7 +587,7 @@ First, alter ``models.py`` once more:
 
     class BlogPageTag(TaggedItemBase):
         content_object = ParentalKey(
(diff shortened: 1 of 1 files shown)


kaedroho:
Hey @brylie, thanks for reporting!

I'm pretty certian that this can be passed through as a positional argument as well. Please can you paste the error you were getting here?

gasman:
`ParentalKey` should indeed support passing the model name as a positional argument, and that's the version we'd want to encourage in the documentation, as it's slightly more concise and matches the version normally used for Django's [ForeignKey](https://docs.djangoproject.com/en/stable/ref/models/fields/#foreignkey). If you've found a place where it's not working as it should, please share more details of the error and we'll be happy to investigate further.

PR #4939, opened 2018-12-03
Module fix for browser module loading in cleanForSlug

jjanssen:
In addition we've forgot about the client-side module.exports handling of cleanForSlug. This is not required for the Node side of the added unit-tests but this does not help the browser. The proper way to do this is: `if (typeof module !== 'undefined' && module.exports)`

Tested in:
- Chrome latest
- Firefox ESR
- Internet Explorer 11

Related to #4884


gasman:
Thanks @jjanssen!

PR #4942, opened 2018-12-05
Implements the new StreamField based on react-streamfield.

BertrandBordage:
This pull request is the final version of the first and main goal of [the first Wagtail crowdfunding campaign](https://kickstarter.com/projects/noripyt/wagtails-first-hatch).

Main features:
- Brand new fancy design
- Straightforward structure
- Drag and drop to move around the blocks
- Blocks use as much width as possible, for a better nesting
- Duplication of blocks
- Blocks can be made collapsible for complex data structures
- Smaller add panels, perfect when using more than a dozen blocks
- Better performance by several orders of magnitude (up to 1000× faster for loading the edit page!)
- New support for `min_num` and `max_num` in `ListBlock`
- Django formsets are no longer used to handle the form data. Previously, we had to convert from the database JSON to formsets, then parse hundreds/thousands of fields with formsets to convert data to JSON. Now, a single hidden JSON input stores on the page the data that will be stored to the database, with almost no need for conversion (Draftail is an exception).
- Far better responsiveness

You can already test it by using [wagtail-react-streamfield](https://github.com/noripyt/wagtail-react-streamfield), a library that patches your existing Wagtail during the execution (a.k.a. monkey patching) to add this pull request.

As you can see, this pull request is fairly small given the amount of work required by it. This is because almost all the magic is happening in [react-streamfield](https://github.com/noripyt/react-streamfield), a JavaScript library I developed for the occasion. As of now, I spent a total of 179 hours on the campaign work (excluding preparation of the campaign), in addition to 120 hours spent benevolently on Wagtail since January.

**So yes, things got a little out of hand** and took much longer than I planned. But I still hang on and am still planning to finish everything I promised I will do.

For the integration, I had to update Babel, React and related librairies because react-streamfield relies on the latest JavaScript syntaxes and technologies.

It’s fully working, but I’m still struggling with some last tests (after updating/adding/removing dozens of them). The remaining tests rely on the old template rendering of the StreamField and are tricky to update, I still have to spend 10-30 minutes on each test.

Here is what it looks like:
image

The InlinePanel PR will follow, it’s enough to review for now ;)

@thibaudcolas My proposition of paying you to review this work is still up! If you don’t have time, please let us know so someone else can do it.


allcaps:
Hi @BertrandBordage, 

I did a test drive of wagtail-react-streamfield, with a production site and production data. It works great! Thank you for all the great work. Congratulations on getting this done the way you did it. Awesome.

I noticed layout changes of blocks when the collapsible layout is enabled. I really like the layout of the collapsible as it gives most out of every pixel on small view ports. The difference between the two is probably a well considered design choice. My gut feeling says that layout change isn't the best indicator for collapsible or not. This is a minor thing. I'm happy to make everything collapsible in my projects ;). Although, the non collapsible blocks do have small issues that the collapsible block doesn't have:

Top block is NOT collapsible rich text, bottom block is collapsible table. Not optimal use of space available.
screenshot 2018-12-06 at 22 53 33

I think this is a bug, the choosers should be visible:
screenshot 2018-12-06 at 22 57 46
screenshot 2018-12-06 at 22 58 01

Last and not least, the contrast of text and icons is too low. In chrome I ran lighthouse accessibility audit:
screenshot 2018-12-07 at 09 29 57

Let me know if you can use any help.

BertrandBordage:
Note that with the latest changes, the duplicate icon changed, I updated the icon font to add a new icon I created with Inkscape, based on the existing document icon:
image

jjanssen:
Hi @BertrandBordage

This is fantastic! 💯 🎉  In between some initial UX findings from a beta-phase we installed during development and this pull request there is not much to add from a functional perspective. 

The initial response from some our QA’ers and content managers was a bit overwhelming in terms of what was changed but it was more a moment of adapting rather than re-exploring how it works. 

Although we have a few UX related findings I would suggest to gather some feedback from the rest of the community as well (and its customers) and the UX/Design team whether we could and/or should apply these findings and suggestions. 

Perhaps it would have a better place in a new improvement issue after the merge but I’ll leave that up to you.

**Creating a new page**
When creating a new blank page it initially was a bit confusing for some that these are actual button actions to add a new block in the content area.  When there is already existing content in the streamfield there is no confusion on what this new interface does. 

image

Suggestion: perhaps it could benefit from a help text when there is no content defined yet or maybe some additional copy in the button.
￼
**Spacing**
During editing the plus icon in between the blocks is more clear to the eyes in a mobile / tablet state. Yet on desktop it took a bit of guessing what the plus on the left does. 

Suggestion: perhaps place the plus in the center in between the blocks on desktop as well.

image

**Consistency**
When there is no title the name of the block is on the left, if there is a title available the name of the black is aligned to the right. When having everything collapsed on entry it is experienced that orientating on the rendered component is delayed due to this behaviour. 

Suggestion: keep the name of the block on the left so the area of scanning the page is not disrupted.

thibaudcolas:
Hey @BertrandBordage, congratulations on pulling this off 👏 it's impressive work, to say the least.

I've been on holidays all of december and will be back at work in mid-january. Then I'm happy to have a look at this, paid or not.

client/src/components/StreamField/index.js
```diff
@@ -0,0 +1 @@
+import '../../../../node_modules/react-streamfield/dist/react-streamfield.js';
```
thibaudcolas: Judging by how this code is included, I understand that the compiled `dist/react-streamfield.js` file includes all of the library's dependencies directly bundled? If so, this means that React + Redux and all their related libraries will be imported twice, among other issues (lack of package management for transitive dependencies). It would be better if:

- Libraries already used by Wagtail were marked as peer dependencies, so we are guaranteed things like React & Redux are only included once, with the same version everywhere.
- Other libraries should be declared as `dependencies` in the package.json, so Wagtail's client-side build can leverage npm to manage their versions (de-duping, & transitive dependency updates without having to re-release `react-streamfield` every time.

I would also expect some or all of the code in https://github.com/noripyt/react-streamfield/blob/c8a42f7700724e603de494d2432e3e966c5c6ca6/src/index.js to be defined in Wagtail instead of the library. It's more common for libraries based on React components to expose the components for integration, or at least an initialisation function, so the dependencies of the library are more clearly defined. Imports should be free of side-effects.

client/src/components/StreamField/StreamField.scss
```diff
@@ -0,0 +1,82 @@
+@import '../../../../node_modules/react-streamfield/src/index';
+
+
+.children-container {
```
thibaudcolas: @BertrandBordage could you explain how this code relates to the styles defined in `react-streamfield/src/index`?

gasman:
I've put up a copy of Bakerydemo with this PR applied, for testing: https://bakerydemo-reactstreamfield.herokuapp.com/admin/

BertrandBordage:
@jjanssen Thanks for taking the time to give that detailed feedback :smiley:

**Creating a new page:** That’s right, it could be clearer, but I’m not really sure where to put the help because of the possible groups. A top right floating help text with a little question mark icon?

**Spacing:** I’m not sure what to choose, given that intermediate pluses take a lot of space, especially because we want them to be big enough for accessibility. I’ll put a survey below for people to vote for or against it.

**Consistency:** Well, the whole point of these collapsed titles is for users to quickly identify the blocks when collapsed. I’m not sure it’s that misleading to have the type that goes to the right, which is roughly where they go when `SIMPLE` layout is used.

BertrandBordage:
**EVERYONE:**
On desktop, would you prefer to have full width centered pluses over the currently left guttered ones?
Vote :-1: to keep left guttered and :+1: for full width centered pluses.
Vote :white_flag: if you don’t care :sweat_smile:  
# Current behaviour, left guttered pluses
image

# Suggested behaviour by @jjanssen, full width centered pluses
image

client/src/components/StreamField/index.js
BertrandBordage: I fully agree with you, this is the ideal solution.
I originally wanted to do exactly this, as you can see on the initial commit: https://github.com/wagtail/wagtail/pull/4942/commits/884e363da094a74d2623ce4c1bcec7c317083643

Unfortunately, I couldn’t manage to update all other projects to Babel 7, I kept facing conflicts, even after twenty hours spent on it… So I decided to make it a fully external dependency instead.

I also agree that `react-streamfield` itself must only be made of components and not have this auto-loaded logic in its `index.js`. But that’s for a scenario where the components API would be stabilized and documented, which I do not plan to do in a near future. It’s out of scope of this work, as discussed in https://github.com/noripyt/react-streamfield/issues/3. So for now, the API is just pure JSON. I’m perfectly aware that it’s not the ideal React approach, but it’s still a perfectly valid React use case.

client/src/components/StreamField/StreamField.scss
BertrandBordage: This code is specific to Wagtail. Wagtail has some global CSS that override `react-streamfield`’s assumptions on the base styling loaded in the page. This SCSS fixes conflicts with Wagtail’s CSS, and also adds some Wagtail-specific styling, like the help text that should be in blocks.

BertrandBordage:
@tmsndrs Interesting approach that Wordpress is taking! If you’re referring to the big squarish buttons with labels below, we already discussed that in the past, as it was Wagtail’s previous approach. The big downside is that your buttons are centered around icons, and most of the time we don’t have existing icons for project-specific blocks, and we don’t want to create specific icons just for one project. In our not-plugin-based case, this is not really relevant. Additionally, we ended up with issues with long labels, especially in some languages like German where you have long words.

What do you find unclear about this icon?
image
As far as I know, it’s the standard look for a copy/duplicate icon since at least 25 years, right?

Adrian-Turjak:
> **Spacing**
> During editing the plus icon in between the blocks is more clear to the eyes in a mobile / tablet state. Yet on desktop it took a bit of guessing what the plus on the left does.
> 
> Suggestion: perhaps place the plus in the center in between the blocks on desktop as well.
> 
> image

The benefit of this new StreamField is that it makes editing HUGE (and nested) pages much easier. The issue with making desktop look the same as mobile is that the added space between the blocks makes the overall size (vertically) bigger. On mobile that's fine, you'll scroll and you probably won't edit on there quite as much, but on desktop the added padding from having those pluses in the middle is frustrating.

What we can potentially do is have the mouse over effect on desktop for the plus be to increase the padding between blocks where the new block would go. The UX isn't exactly that hard to get once you've used it once and anything to decrease vertical size is good.

Adrian-Turjak:
I'd also like to point to my issue: https://github.com/noripyt/wagtail-react-streamfield/issues/31

Just because I feel like SIMPLE as the default layout kind of defeats a lot of the benefits this StreamField offers over the old one, and the current methods for controlling SIMPLE vs COLLAPSIBLE are lacking.

tmsndrs:
> @tmsndrs Interesting approach that Wordpress is taking! If you’re referring to the big squarish buttons with labels below, we already discussed that in the past, as it was Wagtail’s previous approach. The big downside is that your buttons are centered around icons, and most of the time we don’t have existing icons for project-specific blocks, and we don’t want to create specific icons just for one project. In our not-plugin-based case, this is not really relevant. Additionally, we ended up with issues with long labels, especially in some languages like German where you have long words.
> 
> What do you find unclear about this icon?
> image
> As far as I know, it’s the standard look for a copy/duplicate icon since at least 25 years, right?

I was actually just referring to the plus icon coming in on hover in Gutenburg. My personal preference would be for a smaller, centred plus symbol that is clickable from anywhere in the space between blocks. The one demoed above is too big and creates to much white space for me. 

On the duplicate icon, I think I was thrown off by the block type sitting directly below it.

tmsndrs:
> What we can potentially do is have the mouse over effect on desktop for the plus be to increase the padding between blocks where the new block would go, thus giving some visual feedback about what the button might do, and making it clearer that it is interactive.

This 👍

thibaudcolas:
I'm going to review this PR and `react-streamfield` for the front-end part, and have also deployed test environments of the bakerydemo to make it easier. They contain more advanced StreamField implementations compared to the vanilla bakerydemo (https://github.com/wagtail/bakerydemo/issues/168). Feel free to use these if you want to try it out as well:

- Wagtail 2.3 + bakerydemo + extra StreamFields: https://bakerydemo-thibaudcolas3.herokuapp.com/admin/pages/65/edit/
- Wagtail 2.5a0 + the latest PR + extra StreamFields: https://bakerydemo-thibaudcolas2.herokuapp.com/admin/pages/80/edit/

Methodology described [here](https://thib.me/recipe-dead-simple-preview-environments-with-github-travis-heroku) if anyone wants to make their own test environment with Heroku

wagtail/core/blocks/utils.py
```diff
@@ -1,23 +1,55 @@
...
-    dict_items = [
-        indent("'%s': (%s)" % (k, v))
-        for (k, v) in d.items()
-    ]
-    return "{\n%s\n}" % ',\n'.join(dict_items)
+
+def to_json_script(data, encoder=ConfigJSONEncoder):
+    return json.dumps(
+        data, separators=(',', ':'), cls=encoder
+    ).replace('<', '\\u003c')
```
loicteixeira: Wouldn't Django's built-in `escapejs` do the work?

wagtail/core/blocks/field_block.py
```diff
@@ -22,6 +22,44 @@ class FieldBlock(Block):
...
+        title_template = self.get_title_template()
+        if title_template is not None:
+            definition['titleTemplate'] = title_template
+        return definition
+
+    def get_title_template(self):
+        if isinstance(self, (CharBlock, TextBlock, FloatBlock,
+                             DecimalBlock, RegexBlock, URLBlock,
+                             DateBlock, TimeBlock, DateTimeBlock,
+                             EmailBlock, IntegerBlock)):
```
loicteixeira: A list like this will be the source of bug reports because it will always miss some class, especially user created ones.

wagtail/core/blocks/field_block.py
```diff
@@ -22,6 +22,44 @@ class FieldBlock(Block):
...
+                          type_name=None, errors=None):
+        data = super().prepare_for_react(parent_block, value,
+                                         type_name=type_name, errors=errors)
+        if errors:
+            data['html'] = self.render_form(
+                value, prefix=Block.FIELD_NAME_TEMPLATE, errors=errors)
+        return data
+
+    def get_definition(self):
+        definition = super(FieldBlock, self).get_definition()
```
loicteixeira: Since Wagtail does not support Python 2.x, there is no need to pass parameters to `super` (it's not the only occurrence). Old habits die hard 😄

wagtail/core/blocks/base.py
```diff
@@ -107,33 +112,56 @@ def set_name(self, name):
...
+                }
+            )
+
+    def get_definition(self):
+        definition = {
+            'key': self.name,
+            'label': capfirst(self.label),
+            'required': self.required,
+            'layout': self.get_layout(),
+            'dangerouslyRunInnerScripts': True,
```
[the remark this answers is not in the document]
loicteixeira: Actually we have to consider this carefully since there was a whole lot of work done in #4583 in order to eliminate all `eval`s for CSP purposes #1288.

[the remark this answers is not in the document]
thibaudcolas: See https://github.com/thibaudcolas/react-streamfield/pull/1#discussion_r249226678 for a potential solution, with an implementation that's closer to what Wagtail does at the moment (I think).

SPONSORS.md
```diff
@@ -0,0 +1,15 @@
+# Sponsors
```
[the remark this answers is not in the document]
BertrandBordage: I agree. But I don’t know which companies supported Wagtail and what was the scope of their support.

Maybe we could subdivide this file into groups like “Creator”, “Events organizers”, “Main contributors” & “Financial sponsors”?

client/src/components/StreamField/StreamField.scss
```diff
@@ -0,0 +1,82 @@
...
+    .block > .block-container {
+        > aside > .actions > button,
+        > header > aside > .actions > button {
```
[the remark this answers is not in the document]
BertrandBordage: OK, whatever you find more standard. I’m personally a little lost and I don’t think BEM is a strong enough solution (I’m more in the JSS world now ;) ). But as you have more experience in Wagtail’s own CSS, then I will follow the Wagtail rule of using BEM when possible!

client/src/components/StreamField/StreamField.scss
```diff
@@ -0,0 +1,82 @@
...
+
+    .add-panel button.type {
+        .icon,
+        .label {
+            color: unset;
+            font-size: unset;
+            font-weight: normal;
+        }
+
+        .icon > .icon {
```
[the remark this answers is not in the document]
BertrandBordage: This is not related to react-streamfield, I was just using Wagtail icons. Should I hardcode the `:before { content: }` of the icons in this SCSS? I find it a lot worse.

client/src/components/StreamField/StreamField.scss
```diff
@@ -0,0 +1,82 @@
...
+            background: unset;
+            padding: unset;
+
+            &::before {
+                margin: unset;
+            }
+        }
+    }
+
+    .field label {
```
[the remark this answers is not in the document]
BertrandBordage: I’m not sure what you mean? That we remove lines 43-52?

client/src/components/StreamField/StreamField.scss
```diff
@@ -0,0 +1,82 @@
...
+        }
+    }
+
+    .block > .block-container > header {
+        background: #fff;
+    }
+}
+
+.content-container {
+    .content {
```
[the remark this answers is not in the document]
BertrandBordage: Fine.

client/src/components/StreamField/index.js
BertrandBordage: Thank you for the PR you did on react-streamfield for that :)

wagtail/admin/static_src/wagtailadmin/js/hallo-bootstrap.js
```diff
@@ -27,15 +27,15 @@ function makeHalloRichTextEditable(id, plugins) {
     }
 
     var closestObj = input.closest('.object');
-    var isRoot = input.closest('.struct-block').length == 0;
 
     editor.hallo({
         toolbar: 'halloToolbarFixed',
-        toolbarCssClass: (closestObj.hasClass('full')) ? 'full' : (closestObj.hasClass('stream-field') && isRoot) ? 'stream-field' : '',
+        toolbarCssClass: (closestObj.hasClass('full')) ? 'full' : '',
```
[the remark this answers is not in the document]
BertrandBordage: :+1:

wagtail/admin/static_src/wagtailadmin/js/page-editor.js
```diff
@@ -25,21 +25,33 @@ function hideCurrent(current, input) {
...
         $('#' + id).datetimepicker($.extend({
             closeOnDateSelect: true,
             timepicker: false,
             scrollInput: false,
             format: 'Y-m-d',
-            onGenerate: hideCurrent
+            i18n: {
+                lang: window.dateTimePickerTranslations
+            },
+            lang: 'lang',
```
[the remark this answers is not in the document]
BertrandBordage: Good catch, thank you!

wagtail/admin/static_src/wagtailadmin/scss/components/_header.scss
```diff
@@ -50,6 +50,10 @@ header {
         margin-bottom: 0;
     }
 
+    &.tab-merged {
+        overflow-x: hidden;
+    }
```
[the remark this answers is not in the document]
BertrandBordage: Ah, this is a whole story in itself… This is to prevent the streamfield for overflowing when the summary of a collapsed block is too large for the page. This `overflow-x` doesn’t cut the streamfield, but it allows to correctly set the max width of all descendants, including React streamfields. It took me more than 2 days to narrow it down. The whole Wagtail admin structure should be rewritten with flex boxes to fix this whole mess. Because this hack is just a little fix of a lot more pre-existing hacks.

wagtail/admin/static_src/wagtailadmin/scss/layouts/page-editor.scss
```diff
@@ -199,13 +198,22 @@ $object-title-height: 40px;
...
+            padding-left: 20px;
+        }
 
         &.required .field > label:after {
             display: none;
         }
 
+        fieldset {
+            padding-bottom: 0;
+        }
```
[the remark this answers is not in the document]
BertrandBordage: OK, I will do it.

wagtail/admin/templates/wagtailadmin/block_forms/blocks_container.html
```diff
@@ -0,0 +1,12 @@
...
+        <div class="help-block help-critical">{{ error }}</div>
+    {% endfor %}
+{% endif %}
+{% if help_text %}
+    <div class="help">
+        <span class="icon-help-inverse" aria-hidden="true"></span>
+        {{ help_text }}
+    </div>
+{% endif %}
+<BlocksContainer />{# Special tag for react-streamfield #}
```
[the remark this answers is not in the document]
BertrandBordage: This whole HTML is never rendered by the browser, so it does not create any HTML validation issue.
It is directly passed in JavaScript to the React StreamField. It is a placeholder that gets replaced with a React element. This is the only way I could find to hybrid the whole work done in templates with what is rendered by React.

wagtail/core/blocks/base.py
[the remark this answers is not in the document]
BertrandBordage: See in my answer how this doesn’t seem possible to do that without an `eval` in React: https://github.com/thibaudcolas/react-streamfield/pull/1/files#discussion_r258552925

wagtail/core/blocks/base.py
```diff
@@ -485,34 +497,57 @@ def __init__(self, block_def, attrs=None):
...
+        streamfield_config = self.get_streamfield_config(value, errors=errors)
+        escaped_value = to_json_script(streamfield_config['value'],
+                                       encoder=InputJSONEncoder)
+        non_block_errors = get_non_block_errors(errors)
+        non_block_errors = ''.join([
+            mark_safe('<div class="help-block help-critical">%s</div>') % error
+            for error in non_block_errors])
+        return mark_safe("""
+        <script type="application/json" data-streamfield="%s">%s</script>
+        <textarea style="display: none;" name="%s">%s</textarea>
```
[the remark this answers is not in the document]
BertrandBordage: It did not try. I thought this was a bad practice to block DOM parsing & CSS rendering with JavaScript. This is what is happening with the old StreamField, and it completely prevents the load of the bottom of the page (like the save/delete buttons) while StreamFields are not initialized.

wagtail/snippets/static_src/wagtailsnippets/js/snippet-chooser.js
```diff
@@ -4,17 +4,21 @@ function createSnippetChooser(id, modelString) {
     var input = $('#' + id);
     var editLink = chooserElement.find('.edit-link');
 
+    function snippetChosen(snippetData, initial) {
+        if (!initial) {
```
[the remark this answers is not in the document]
BertrandBordage: Every time one of the fields in a StreamField gets modified, a boolean gets switched to `true` that will show a confirm dialog before leaving the page. So this check was done to prevent the page to be considered as modified when loading it. In addition to that, the value is already set in the `input` when rendering the template, so it does not need to be set again.

wagtail/snippets/static_src/wagtailsnippets/js/snippet-chooser.js
```diff
@@ -23,4 +27,12 @@ function createSnippetChooser(id, modelString) {
         input.val('');
         chooserElement.addClass('blank');
     });
+
+    if (input.val()) {
+        $.ajax(window.chooserUrls.snippetChooser + modelString + '/'
+               + input.val() + '/')
```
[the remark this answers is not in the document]
BertrandBordage: Indeed, nice catch.

wagtail/core/blocks/utils.py
BertrandBordage: Unfortunately not. `espacejs` would have to be used recursively on all values, but not on the JSON string itself, otherwise it’s no longer valid JSON and it would require unneeded additional `.replace` on the JS. `escapejs` is really done for JS passed in HTML tag attributes, not for JSON.

wagtail/core/blocks/field_block.py
BertrandBordage: Mmm… I don’t really see why, this method can be defined in any child class.
An alternative to this tuple would be to specify the same method in all of these blocks, but I’m not sure that’s really better.

wagtail/core/blocks/field_block.py
BertrandBordage: Aha, indeed, I forgot to replace it when making the PR from the monkey-patched application (that needs it because of monkey patching, not because of Python 2).

BertrandBordage:
Thank you very much for the review @thibaudcolas :)
Very helpful, and I’m glad to see that the issues you found are mostly related to build/webpack and CSS conventions ;)

I’m working on fixing all these starting today, with the help of @jonnyscholes for BEM (he will be paid as well with the Kickstarter money).

About the new layout of the “plus” button, I will make it default to the centered position, and make it possible to switch to the guttered solution.

wagtail/core/blocks/struct_block.py
```diff
@@ -64,60 +54,48 @@ def get_default(self):
...
-                block.bind(value.get(name, block.get_default()),
-                           prefix="%s-%s" % (prefix, name), errors=error_dict.get(name))
-            )
-            for name, block in self.child_blocks.items()
-        ])
-
-        return {
-            'children': bound_child_blocks,
-            'help_text': getattr(self.meta, 'help_text', None),
-            'classname': self.meta.form_classname,
```
lb-: will it still be possible to define a custom class name for blocks?

BertrandBordage:
This work is now fully ready for merge!
Here are the main latest changes:
- By default, the plusses are centered. A `gutteredAdd` setting is available in `react-streamfield` to have the plusses in the left gutter.
- Almost all feedback from @thibaudcolas in https://github.com/noripyt/react-streamfield/issues/4 has been tackled. All critical improvements have been done. What remains mainly is to improve error handling (when a JS widget crashes). Like for Draftail, I will add that in a later release of react-streamfield.
- In addition to rewriting the whole SCSS using BEM thanks to @jonnyscholes, I added minor styling improvements to fix all the tiny snags that nobody probably noticed ^^

@gasman It’s waiting for you now :)

BertrandBordage:
Note that there are two features that are **no longer supported** with this new StreamField:
`form_template` & `form_classname` in `StructBlock`.

I know it’s a backwards incompatibility issue, but I really can’t think of any way to properly do that. The only ways possible are extremely messy, and they will not work when the template contains random data, because now, the new StreamField uses only a single template render per block type to render all blocks on the client side.

lb-:
> Note that there are two features that are **no longer supported** with this new StreamField:
> `form_template` & `form_classname` in `StructBlock`.

No problems, I have added a note to the PR #5100 which will affect #4042

gasman:
> Note that there are two features that are no longer supported with this new StreamField:
form_template & form_classname in StructBlock.

Oh, that's bad news :-( I always intended StreamField to be a framework for developers to create new block types, as well as a set of building blocks, and StructBlock + custom `form_template` is probably the most basic form of that. This is one thing that I was concerned about with the drive to port client-side code to React - that we'd lose a lot of the "hackability" for Python developers.

If a developer has an existing StructBlock with a custom form_template, what would they have to do to migrate it to the new React implementation? If it's largely a case of translating a Django-flavoured template to a React-flavoured one, that might not be too bad.

roelio:
@BertrandBordage @gasman We're currently using it with custom React component like follows:

In the patched react-streamfield https://github.com/stamkracht/react-streamfield/tree/0.8.1.d1a98090c3f5db1561e3123201d52e56e7761d40 I basically put in some hooks that will render React components that are defined in the `window.reactStreamFieldComponentMapping` variable.

We extend the `entry,js` from wagtail-react-streamfield to provide the custom components with the store and add them to  `window.reactStreamFieldComponentMapping`

```
window.reactStreamFieldComponentMapping['Hotspot'] = (props) => (
  <Provider store={store}>
    <Hotspot {...props}/>
  </Provider>
);
```

In the backend the following is used to connect a Block to a Component.
```
class HotspotBlock(BaseQuestionBlock):
    region_selector = blocks.StaticBlock()
    regions = ListBlock(child_block=HotspotRegionBaseBlock())

    def get_definition(self):
        definition = super().get_definition()
        for child in definition["children"]:
            if child["key"] == "region_selector":
                child["reactComponent"] = {
                    "name": "Hotspot",
                    "block_type": "FieldInput",
                }

                if hasattr(self.meta, "hotspot_kwargs"):
                    child["reactComponent"]["props"] = self.meta.hotspot_kwargs
        return definition
```

The hooks aren't pretty but it seems to work, also with a custom component that wraps a DictBlock to provide some dynamic behavior to the fields in the dict. Creating the correct reducers and getting the data wanted from the state took some time to get right. 

I think that if the hooks are done correct and there is some registration interface in the frontend that will also provide the redux store to the custom component. 
In wagtail something similar as above for connecting a Block to a Component could be done. The lib folder in the patched version was necessary because in the custom components classes from react-streamfield were needed. They are not exported from the index.

wagtail/core/blocks/base.py
```diff
@@ -107,33 +113,62 @@ def set_name(self, name):
...
-        """
-        Returns a Javascript expression string, or None if this block does not require any
-        Javascript behaviour. This expression evaluates to an initializer function, a function that
-        takes the ID prefix and applies JS behaviour to the block instance with that value and prefix.
+    def prepare_for_react(self, parent_block, value,
+                          type_name=None, errors=None):
+        if type_name is None:
+            type_name = self.name
+        from .stream_block import StreamValue
+        if isinstance(value, StreamValue.StreamChild):
```
gasman: This is raising major alarm bells. It's absolutely essential to the design of StreamField that the `Block` class should define a contract for each block type to follow, _independently of any other block's implementation_. That way, we can verify the code by checking that each block implementation is following that contract, rather than having to consider every possible interaction between different parent/child block types.

No block other than `StreamBlock` should ever have to know about the `StreamChild` class. Additionally, there should never be any need for a block to vary its behaviour based on the identity of its parent, as we're doing here (`if parent_block is None`).

I suspect there is a similar logic error to https://github.com/wagtail/wagtail/pull/4481/files#r190932257 here: a `StreamBlock` value is a sequence of _wrapped_ child values (i.e. wrapped in a `StreamChild` object in the Python representation, and wrapped in a `{'type': ..., 'value': ..., 'id': ...}` dict in the JSON representation). `StreamBlock`'s `prepare_value` / `prepare_for_react` methods should be responsible for unwrapping each child value before passing it to the child block's `prepare_for_react` - that way, `Block.prepare_for_react` will consistently receive the unwrapped value and not need a special case for `StreamChild` - and `StreamBlock` will also be responsible for re-wrapping the returned value in a `BlockData` object.

Once all of the `StreamBlock`-specific logic here has been moved into the `StreamBlock` class, I believe it will reduce to:

    def prepare_for_react(self, value, errors=None):
        return self.prepare_value(value, errors=errors)

...which brings us back to the question of why we need both `prepare_value` and `prepare_for_react`, I guess :-)

wagtail/admin/static_src/wagtailadmin/js/page-chooser.js
```diff
@@ -39,4 +42,17 @@ function createPageChooser(id, pageTypes, openAtParentId, canChooseRoot, userPer
         openAtParentId = null;
         chooserElement.addClass('blank');
     });
+
+    if (input.val()) {
+        $.ajax({
+            url: window.wagtailConfig.ADMIN_API.PAGES + encodeURIComponent(input.val()) + '/',
```
gasman: This will cause an extra HTTP request for every chooser widget rendered, whether it's part of a StreamField or not (and if it isn't, the HTTP request is redundant). Chooser widgets are intended to be a standalone reusable component, so if they need a StreamField-specific customisation, then that's something we need to consider carefully.

What exactly is the technical limitation of react-streamfield that makes this change necessary? (Or, to put it another way: as a Django developer, what do I need to know to ensure that the form widgets I create will be compatible with react-streamfield?)

Is it that it has to be rendered entirely client-side, so we don't have access to any data that isn't part of the StreamField JSON (such as page title)? If so, isn't that what `prepare_value` / `prepare_for_react` are for? (How does `RichTextBlock` handle it, given that it relies heavily on server-side data conversion?)

BertrandBordage: Yes, I’m aware that it triggers a bunch of AJAX requests on page load. Here’s why:

It is not due to a technical limitation, it’s for the new duplication feature, as well as for optimization.

When you duplicate a block, you cannot copy its full JavaScript state, you have to call that page/image/document/snippet callback again. So doing that AJAX call was an easy fix to make duplication work without rewriting the whole chooser logics. And after doing that, I was amazed to see that it was actually a lot faster than I thought. Because like you, I’m not a fan of doing a ton of async requests.

So the next step for me was to try to avoid doing the Django template rendering of the same blocks over and over again, because we could just fetch only the data we need using AJAX. And the results was even more amazing, that’s why this StreamField is considerably faster to load than the old StreamField, even when you take into account the AJAX requests. For info, there is one case where I always render a block anyway: when there is an error in the block. Otherwise I only do a single render per block type.

So after these considerations I thought: why not do it this way? Since our goal with Wagtail is to go towards a headless CMS & more dynamic page editor (like the realtime preview), this approach of AJAX widget is not heretic. And since for some extreme cases it’s up to **1000 × faster**, I became actually proud of this approach ^^

gasman: > When you duplicate a block, you cannot copy its full JavaScript state, you have to call that page/image/document/snippet callback again. So doing that AJAX call was an easy fix to make duplication work without rewriting the whole chooser logics.

OK, so my question still stands: as a Django developer, what do I need to know to ensure that the form widgets I create will be compatible with react-streamfield's implementation of block duplication?

wagtail/core/blocks/base.py
BertrandBordage: That’s a good discussion to have. The main reason for this design is compatibility, as you will see. Maybe it can be improved so that there is no `StreamChild` reference, but I’m not sure how.

In theory, we should  not need to differenciate between `prepare_for_react` and `prepare_value`, as you pointed out.

I made that difference is because currently, the database representation of StreamFields is not really consistent.
Here is a comparison between database representation and react-streamfield representation, so you can understand the purpose of this:

StreamBlock of FieldBlocks in database:
```JSON
[{"id": "98a82", "type": "a", "value": 1}, {"id": "87b1", "type": "b", "value": 2}]
```

StreamBlock in react-streamfield:
```JSON
[{"id": "98a82", "type": "a", "value": 1}, {"id": "87b1", "type": "b", "value": 2}]
```

ListBlock in database:
```JSON
[1, 2]
```

ListBlock in react-streamfield:
```JSON
[{"id": "98a82", "type": "a", "value": 1}, {"id": "87b1", "type": "b", "value": 2}]
```

StructBlock in database:
```JSON
{"id": "98a82", "type": "person", "value": {"first_name": "Bertrand", "last_name": "Bordage"}}
```

StructBlock in react-streamfield:
```JSON
{"id": "b198e", "type": "person", "value": [
    {"id": "98a82", "type": "first_name", "value": "Bertrand"},
    {"id": "87b1", "type": "last_name", "value": "Bordage"}
]}
```

You can notice two main things here:
- In react-streamfield, all blocks are represented as an object, with an ID and a type. they can also have their own `hasError` and `html` keys, but that’s not really important here.
- In react-streamfield, ListBlock is exactly the same thing as a StreamBlock. Not only in terms of representation here, but also internally.

So as you can see, the poor FieldBlocks are not always considered as blocks in the database representation. The way they are treated depends on their parent. When a FieldBlock is in a StreamBlock, it’s considered as a normal block. When it’s in a ListBlock, it’s considered as a value. When it’s a StructBlock, it’s considered as a key/value pair.

I appreciate that the database representation is minimal. It doesn’t add more data than is required, which is good. But it’s truely a pain to parse because of this.

Also, there was no way I could just send the database representation to react-streamfield. We need more information than just a value in a good JavaScript StreamField. We need an `html` key in the block representation for cases where there is a special HTML for a block instance, like when it has an error. And it was pointless to try to code a structure parsing the special ListBlock & StructBlock children when it might change in the future. So I made react-streamfield generic, with a simple structure for all block representations.

To conclude, we have two possible solutions for this:
- Write a database migration to migrate all StreamFields to a consistent structure. This could be very long on some sites, but the main reason I didn’t do it is that it needs its own discussion, and it’s risky. So it’s something we might end up doing, but in like 1 or 2 years
- Change the strucutre to make StreamBlock and ListBlock iterate over children to reformat their values. I guess that’s possible, but to be confirmed. I specifically remember trying to do it, and ending up with this instead. I will try again in the next days.

wagtail/admin/static_src/wagtailadmin/js/page-chooser.js
BertrandBordage: The only thing to do is to have the same HTML code in the widget, regardless of the value. The only scenario where that can change is for displaying errors.

wagtail/core/blocks/base.py
gasman: > So as you can see, the poor FieldBlocks are not always considered as blocks in the database representation. The way they are treated depends on their parent. When a FieldBlock is in a StreamBlock, it’s considered as a normal block. When it’s in a ListBlock, it’s considered as a value. When it’s a StructBlock, it’s considered as a key/value pair.

No, that's not true - I think you've misunderstood the significance of the `{'id', 'type', 'value'}` container. This is the JSON equivalent of `StreamChild`, and is **an internal detail of StreamBlock** - no other block type will ever be required to handle it. It's not the thing that turns a block value into a "normal block". Or, to put it visually...

streamfield-mental-model

There is no inconsistency here: each block type decides on one value type to work with (which can be any Python data type), and consistently receives values of that exact type - not a wrapped version of it.

* for CharBlock, the value type is a string
* for PageChooserBlock, the value type is a `Page` object (which becomes an integer in the JSON representation)
* for ListBlock, the value type is a list of child values
* for StructBlock, the value type is a dict (or, more accurately, a `StructValue` which provides some extra semantics over `dict`) of child values
* for StreamBlock, the value type is a sequence of StreamChild-wrapped child values (which becomes a sequence of dict-wrapped child values in the JSON representation)

With a lot of persuasion, I could possibly come round to the idea of the `{'id', 'type', 'value'}` wrapper being understood and handled by each block as part of the base `Block` implementation, rather than being internal to StreamBlock. (Right now I'm not convinced that it's a worthwhile change: that wrapper dict consists of the metadata required for StreamBlock to do its job, and I can't see the same set of fields being relevant to other container types.) However, if we're doing that, we absolutely have to be explicit about the signature of each `Block` method, and whether it accepts/returns the wrapped or unwrapped version. Treating the two as interchangeable via type sniffing and undocumented `parent_block` logic is introducing vagueness to the block API, that I went to a lot of effort to avoid.

BertrandBordage: Oh. You have to admit that this completely counter-intuitive.

Also, I stand to what I said, this makes the blocks representation inconsistent. With the current approach, `StreamBlock` has a very different representation from `ListBlock`, but they are the same thing. The only difference is that `ListBlock` can assume the type of its children. But that does not explain why the structure is so different. Also, this makes only `StreamBlock` store IDs, which is not great to identify blocks and version history.

As you said, there is no need to change the database representation for now. And it’s a good thing, because we can easily switch back to the old streamfield without having to write a forward **and** backwards migration.

There is one good reason to do it in the future though: simplification. We do not need `StreamChild`, `StructValue` or `ListBlock` with the approach I took, and we will be able to finally make a system where we expose blocks directly in templates. This would be a major improvement, as it would allow us to write block methods and handle blocks in a very similar fashion to models.

wagtail/core/blocks/field_block.py
gasman: How about having a `use_as_title` flag defined on `FieldBlock`, defaulting to False but overrideable on subclasses, and checking that flag here?

wagtail/tests/utils/form_data.py
```diff
@@ -42,35 +42,6 @@ def nested_form_data(data):
     return {'-'.join(key): value for key, value in _nested_form_data(data)}
 
 
-def streamfield(items):
```
gasman: This is a public documented function, and if we're removing it, we need to remove it from the docs and add a release note:

https://docs.wagtail.io/en/v2.4/advanced_topics/testing.html#wagtail.tests.utils.form_data.streamfield

However, probably a better idea is to keep it here, and rewrite it to output the JSON value instead - that way, we avoid making a breaking change, and no longer need to update our own tests (and the tuple syntax is slightly more human-friendly than having devs write JSON directly anyhow).

wagtail/core/blocks/base.py
gasman: A quick question before I respond in full: do you consider the "block representation" of a CharBlock to be

    "hello world"

or

    {"id": "98a82", "type": "heading", "value": "hello world"}

? The examples in your previous comment are ambiguous on this point, and I'd like to make sure we're using the same terminology.

BertrandBordage: I understand that you consider that it is `"hello world"`, and I also understand that it is in a way what is currently implemented, but I consider that it would be clearer and easier to parse if we considered `{"id": "98a82", "type": "heading", "value": "hello world"}` to be the block representation.

This feels like the kind of things that made this possible in HTML5:
```
<ul>
    <li>First
    <li>Second
    <li>Third
</ul>
```
Obviously, it is “logic” that we do not have to close `li` tags because it’s an optimization that doesn’t introduce ambiguity. But on the other hand this makes it harder to understand HTML nowadays, and it makes it much harder to write a parser, because `li` is not a standalone tag, we are supposed to “guess” that the `li` gets closed by the next `li`…

It’s the same kind of feeling that I have here. OK, it’s more optimized, it may not be needed in all scenarii, but it adds complexity to parsing and understanding. Note as well that it makes it hard to cast streamfields to `JSON` or `JSONB` in PostgreSQL and search for a value. It’s probably the main reason why my streamfield collection system in https://github.com/wagtail/wagtail/pull/4702 was messy.

wagtail/tests/utils/form_data.py
BertrandBordage: OK, I revert that.

wagtail/core/blocks/base.py
```diff
@@ -107,33 +113,62 @@ def set_name(self, name):
...
-        blocks on the page.
+    def get_layout(self):
+        return self.SIMPLE
 
-        Any element IDs used in this HTML fragment must begin with definition_prefix.
-        (More precisely, they must either be definition_prefix itself, or begin with definition_prefix
-        followed by a '-' character)
-        """
-        return ''
+    def prepare_value(self, value, errors=None):
```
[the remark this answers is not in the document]
BertrandBordage: Done.

wagtail/core/blocks/base.py
BertrandBordage: Fixed by https://github.com/wagtail/wagtail/pull/4942/commits/bf862937ab62d08bf1f9f4037ce4567b4558dc85.

wagtail/tests/utils/form_data.py
BertrandBordage: Introduced back as https://github.com/wagtail/wagtail/pull/4942/commits/821d7e924e5fd9c7a90e137be9f9054f6d9bfac1. Do you think that’s enough?

BertrandBordage:
@auvipy It’s updated ;)

SPONSORS.md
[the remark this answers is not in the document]
auvipy: wagtails-first-hatch-crowd-funding-sponsors?

client/src/components/StreamField/index.js
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { createStore, applyMiddleware } from 'redux';
```
auvipy: just curious, did you ever tried Appollo client or mobx?

package.json
```diff
@@ -86,17 +86,19 @@
...
+    "custom-event-polyfill": "^1.0.6",
     "draft-js": "0.10.5",
     "draftail": "^1.2.1",
     "element-closest": "^2.0.2",
     "focus-trap-react": "^3.1.0",
-    "prop-types": "^15.6.0",
-    "react": "^16.2.0",
-    "react-dom": "^16.2.0",
+    "prop-types": "^15.6.2",
+    "react": "^16.4.0",
```
auvipy: 16.8.6 is the latest version

client/src/components/StreamField/index.js
BertrandBordage: No, I didn’t. Thank you for pointing them out, Mobx especially seems very interesting! I will probably use it in other projects.
As for Wagtail, the choice was already settled with the other React-based tools that we have that all use Redux, especially Draftail.

package.json
BertrandBordage: Indeed, but it’s extremely tricky to make React updates on Wagtail. It was already a pain to just go from React 16.2 to 16.4 and make sure everything else still built. At first I tried to build with 16.7, but there was a chain reaction of dependencies updates that implied a lot of rewrite of other modules, so I had to keep an “old” version for now.

client/src/components/StreamField/index.js
auvipy: OK thanks. I started with redux back in 2016, but later moved to mobx and recently appollo!

package.json
auvipy: OK fair enough.

BertrandBordage:
@auvipy It’s updated :)

auvipy:
is the failures anyhow related? may be not

BertrandBordage:
@auvipy Ah, it’s a damn Python < 3.6 unsorted dict "issue" (I know we are not supposed to rely on the fact that they are sorted now).

gasman:
Rebased + squashed at https://github.com/gasman/wagtail/tree/reactstreamfield-rebase, except for the addition of the 'duplicate' icon in cf4d7ffb233c1fc4dde4f1371124061b47aea172 which will need redoing in icomoon to resolve the merge conflict.

client/src/components/StreamField/index.js
```diff
@@ -0,0 +1,36 @@
...
+const store = createStore(streamFieldReducer, applyMiddleware(thunk));
+
+const init = (name, options, currentScript) => {
+  // document.currentScript is not available in IE11. Use a fallback instead.
+  const context = currentScript ? currentScript.parentNode : document.body;
+  // If the field is not in the current context, look for it in the whole body.
+  // Fallback for sequence.js jQuery eval-ed scripts running in document.head.
+  const selector = `[name="${name}"]`;
+  const field = (context.querySelector(selector)
+                 || document.body.querySelector(selector));
```
gasman: Could we select the field based on ID instead? That's the standard way we attach Javascript behaviour to widgets (see https://github.com/wagtail/wagtail/blob/master/wagtail/admin/templates/wagtailadmin/widgets/draftail_rich_text_area.html for example).

gasman:
The heading for a `ListBlock` is showing as "${}":
Screenshot 2019-07-25 at 17 32 40

Here the 'shopping list' block is defined as:

    shopping_list = ListBlock(CharBlock())

Presumably this is because the container element is trying to look up the child block by name - child blocks of `ListBlock` don't have names.

BertrandBordage:
You are right @gasman, it was because of the missing name in list block children. I made a workaround.

auvipy:
I thought this was included in wagtail 2.7!

benjaoming:
@auvipy A subset of these changes were migrated to allow for a seamless upgrade to 2.7. As I understand the Release Notes, 2.7 is LTS and the changes were quite modest and don't disturb data or affect user instructions.

https://github.com/wagtail/wagtail/pull/5476

Looking forward to seeing more work on this stuff and a big :clap: to all involved!

gasman:
A progress update on this - while the styling updates have been merged as of #5476, the deeper behavioural and architectural changes are still outstanding.

At the heart of those changes is a shift away from the server-side model, where the structured StreamField data is constructed server-side from the individual fields of the HTTP submission, in favour of a model where the data is collated client-side into a single field before submission.

This in turn means that we're breaking away from Django's form API, where the client-side HTML is effectively opaque and it's the sole responsibility of the Python-defined widget to handle the translations between HTML, form postdata and final data - because the client-side code has to have sone concept of how to extract a value from the DOM representation.

In the 'base' case, where a widget corresponds to a single HTML `<input>`, that's trivial - but however much the react-streamfield code is refined to handle 99% of the 'standard' setups, I'm wary that we're still going to get cases where someone tries to wrap a colour selector / map location picker / MathML formula editor widget that's entirely well-behaved according to the Django API, and be told "we don't handle that" or "oh that's easy, you just have to rewrite it as a React component - just read this book about the Node ecosystem." :-)

It's clear that there's strong support - both in the Wagtail core team and outside - for increasing our use of React so that we can stay ahead of the curve on rich UI. That might mean that the Django forms framework is no longer adequate for our needs, and we need to create or adopt a new framework that can do for client-side form handling what Django does for server-side... if so, we shouldn't shy away from making that leap, but we also shouldn't be making that sort of change 'through the backdoor', without a full plan for what it means for Wagtail and Django developers working with it.

I'm hoping to get up to speed on React development over the next couple of months so I can be better informed about what that framework might look like.

In the meantime, it should be possible to backport a few more of the react-streamfield features to the current implementation (such as enforcing min_num / max_num in the UI), and hopefully having the CSS in Wagtail now makes it less of a moving target for the external react-streamfield package to stay up to date with.

lb-:
@gasman - an example of the benefits of leveraging the existing Django form system is how easy it is for a non-frontend dev to add something like 'tags for streamfield'. While the full blown react implementation does solve some problems, I agree with your concern that it may create a lot more for the Wagtail assumption of 'its just Django'.

https://stackoverflow.com/questions/61231112/putting-tags-in-a-structblock/61264112

Disclaimer I put that answer up, there may be a better way to solve this in Wagtail :)

gasman:
Now completed via the streamfield rewrite in 2.13.

PR #4946, opened 2018-12-07
Allow disabling search auto-update on a per-model basis

kaedroho:
In some cases, you may want ``AUTO_UPDATE`` enabled on your Wagtail Page model but rely on ``update_index`` for a custom indexed model.

This PR implements some logic to allow auto-update to be configured on a per-model basis using the ``search_auto_update`` attribute.


gasman:
Looks good to me! Merged in 50e72dc58b4e61279dc8ef8c5da0db86d2735ddd.

PR #4950, opened 2018-12-07
Fix SERVER_PORT for Page.dummy_request() when page is attached to Site with 443 port.

sir-sigurd:


gasman:
Merged in 5f7b7230a26e244fd9c39d98b6e206902b594a9e - thanks @sir-sigurd!

PR #4951, opened 2018-12-07
Simplified Site.find_for_request().

sir-sigurd:
Handling of `AttributeError` is unneeded since support for Django 1.8 is dropped and handling of `KeyError` is unneeded because both hostname and port are obtained now via Django API (its current implementation can raise `KeyError` if `SERVER_NAME` or `SERVER_PORT` are absent, but they are required per [PEP 3333](https://www.python.org/dev/peps/pep-3333/#environ-variables)).

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -107,16 +107,8 @@ class Site(models.Model):
         still be routed to a different hostname which is set as the default
         """
 
-        try:
-            hostname = request.get_host().split(':')[0]
-        except KeyError:
-            hostname = None
-
-        try:
-            port = request.get_port()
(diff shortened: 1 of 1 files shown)


gasman:
Thanks @sir-sigurd!
Looks like this change requires us to fix up some of our tests (to include the required META fields in cases where we're constructing `HttpRequest` objects ourselves).

gasman:
Committed (with tests updated, or removed where they were explicitly testing the case where SERVER_NAME / SERVER_PORT were missing) in f271c40cea7999ffb0cc808b16854302ead841b5.

PR #4952, opened 2018-12-11
Fix: Modeladmin doesn't discover proxy models

### LINKED ISSUE #4913: Modeladmin doesn't discover proxy models
### Issue Summary

Modeladmin PermissionHelper looking up for permission objects with self.opts.app_label/self.opts.model_name, which is not correct for proxy models.

### Steps to Reproduce

1. start new project
2. declare custom model in models.py, which extends existing model using proxy model mechanism
3. declare custom model admin for that model
4. admin won't show up in the menu

looks like PermissionHelpers method get_all_model_permissions returns empty queryset, because its looking up as:
```
Permission.objects.filter(
            content_type__app_label=self.opts.app_label,
            content_type__model=self.opts.model_name,
        )
```
but permissions for proxy models would refer to the parent model (see https://stackoverflow.com/questions/15037642/django-proxy-model-permissions-do-not-appear)

so probably the correct way would be to use parent's meta in `__init__`:

```
...
        self.opts = model._meta if not model._meta.proxy else ????
...
```

???? should be replaced with a upwards lookup for **non** proxy model in the hierarchy.

### Technical details

* Python version: Run `3.6.6`.
* Django version: `2.0.9`
* Wagtail version: `2.1.3`

ababic:
Resolves #4913.

PermissionHelper.get_all_model_permissions() currently only works for concrete models, because the cross-table query created by the method presumes the model has its own content type. Although there's more work to be done in supporting proxy models more fully throughout Wagtail, there's no reason why `PermissionHelper` should get in the way here.

The fix utilises the `ContentTypeManager`'s `get_for_model()` method to fetch a `ContentType` object for the model - which, when provided with a proxy model, returns the type for the 'concrete' model instead.

Not only does this fix the issue, but it's also more consistent with how content types are used throughout Wagtail. It also simplifies the query made by `get_all_model_permissions()` slightly, and benefits from the `ContentTypeManager`'s built-in cache.


chosak:
@ababic this works well, thanks.

I did encounter some unexpected behavior, which is perhaps what you are referring to by "there's more work to be done in supporting proxy models". Specifically, given the test example you provide here, with Author and LegendaryAuthor, and ModelAdmins defined for both:

- A user having only one of "Can view author" permission or "Can view legendary author" permission  still sees (read-only) model admin entries for both in the sidebar.
- A user given Add, Edit, or Delete permission to the Author model unexpectedly gets that permission over only the proxy LegendaryAuthor model, and not the Author model.

Granted, the use case of having ModelAdmins for both models could be rare.

If there's only a single ModelAdmin defined for the proxy model, everything works fine, although there's an unexpected "Can view author" permission option in the Groups screen that doesn't seem to do anything.

A bigger problem comes if there's only a single ModelAdmin defined for the base model. The base model appears in the sidebar, but add/change/delete permissions don't seem to work (and again, there's an unexpected "Can view legendary author" permission option that doesn't do anything).

This all can perhaps be summed up as: if you want to create a proxy model on a model that's being exposed via ModelAdmin, the ModelAdmin has to be defined for the proxy model, and not the base model.

Does this match with your expectations? I'm not that familiar with Django ModelAdmin and if it has any similar constraints with proxy models. If you think it makes sense it might be good to add a note in the ModelAdmin documentation about use of proxy models if it seems likely that users might encounter these issues.

ababic:
Hi @chosak. Thanks so much for taking the time to review this, and for concise feedback.

A lot of these peculiarities are expected, but there are a couple of things that are not. Specifically these two... 

> A user given Add, Edit, or Delete permission to the Author model unexpectedly gets that permission over only the proxy LegendaryAuthor model, and not the Author model.

> A bigger problem comes if there's only a single ModelAdmin defined for the base model. The base model appears in the sidebar, but add/change/delete permissions don't seem to work. 

I'll investigate these and see if anything can be done without too much scope creep, but my gut feeling is that separate issues/PRs will be required. 

I'm a touch short on time right now, so sorry for not addressing all of your points here. I will find the time to reply more thoroughly soon. Thanks again :)

ababic:
Hi @chosak. I've come to realise there are two ways to resolve this problem... 

One that aims to support Django's default proxy model/ContentType/Permission setup (the approach taken by this PR). And another that focuses on ensuring proxy models have their own unique `ContentType` object, with correctly associated `Permission`s (the approach outlined in #4973).

Personally, I think the latter is the better approach, which would make this PR redundant. However, I'll await feedback on that issue closing this.

chosak:
@ababic thank you very much for the detailed explanation! You've taught me a thing or two about proxy models. My apologies for the delay in responding (holidays). That [10-year old Django ticket](https://code.djangoproject.com/ticket/11154) has quite an interesting history.

I agree with your definition of the ideal scenario, and also with the pain that would be required to implement it, given how Django proxy models don't have their own `ContentType`.

> I think a lot of the very weird behaviour you reported is actually more to do with what's happening in the Group permissions UI/view.

I think you're right, based on [this code](https://github.com/wagtail/wagtail/blob/882f8f3cf8ddd79c30e611a48882b309e90dad0c/wagtail/users/templatetags/wagtailusers_tags.py#L40) that creates a set of permissions based on content type. I suppose this could be improved to somehow distinguish proxy models, but you'd have to use some kind of codename-based logic (which also wouldn't properly handle custom permissions). It feels like this would just be better fixed if Django gave proxy models their own `ContentType` (or we do something as you propose in #4973).

I agree that a `ContentType`-related fix would be better than what's proposed in this PR, if we think developers are likely to run into some of the unexpected issues mentioned above (although I say that reluctantly, given that you've already done the work here). I'll comment over on #4973.

ababic:
> I agree that a ContentType-related fix would be better than what's proposed in this PR, if we think developers are likely to run into some of the unexpected issues mentioned above (although I say that reluctantly, given that you've already done the work here).

No worries :) It took doing this work to really get my head around everything, so it was all useful.

ababic:
Am closing this, as the solution relies on Django behaviour that will be changing in 2.2.

PR #4954, opened 2018-12-12
Made HTTP_HOST of page dummy request include port.

sir-sigurd:
`Host` header should include port, see https://tools.ietf.org/html/rfc7230#section-5.4.


gasman:
Thanks @sir-sigurd! Merged in 057690815b3147be1280834787e275590f71d274.

Issue #4958, opened 2018-12-15
More informative error for invalid values passed to `|richtext`


gasman:
Passing a value other than a string, `None`, or a `wagtail.core.rich_text.RichText` object to the `|richtext` template filter currently throws a hard-to-decipher error: `TypeError: expected string or bytes-like object` in wagtail/core/rich_text/rewriters.py. Reported here: https://stackoverflow.com/q/53790708/1853523

We should add a check to the template filter [here](https://github.com/wagtail/wagtail/blob/5f7b7230a26e244fd9c39d98b6e206902b594a9e/wagtail/core/templatetags/wagtailcore_tags.py#L89) to confirm that the value is a string or bytes object, and if not, throw a `TypeError` with a more meaningful error message e.g. `'richtext' template filter received an invalid value; expected string, got [value]`.

mukesh5:
Hello, I am first time contributor. Can I work on this?

gasman:
@mukesh5 Yes please!

PR #4971, opened 2018-12-30
Fix: More informative error for invalid values passed to `|richtext`

mukesh5:
More informative error for invalid values passed to `|richtext`

### DIFF
--- a/wagtail/core/templatetags/wagtailcore_tags.py
+++ b/wagtail/core/templatetags/wagtailcore_tags.py
@@ -87,7 +87,11 @@ def richtext(value):
     elif value is None:
         html = ''
     else:
-        html = expand_db_html(value)
+        if isinstance(value,str) or isinstance(value,bytes):
+            html = expand_db_html(value)
+        else:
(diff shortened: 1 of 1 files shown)


kaedroho:
Thanks @mukesh5! Please could you post the error message this currently returns for comparison?

gasman:
@kaedroho See the original report at #4958.

gasman:
Thanks @mukesh5! This looks good to me, but it's just missing a test in https://github.com/wagtail/wagtail/blob/master/wagtail/images/tests/tests.py#L30 to confirm that this is indeed throwing the expected error. (I'll add this shortly, unless you beat me to it :-) )

gasman:
Now merged in fa5deb6b44b61b3c501008797b08afcb5a1b9426, with tests added and a small adjustment to the logic (on closer inspection it turns out that bytes values are also invalid here, so those now throw the same exception).

mukesh5:
@gasman @kaedroho Sorry :( got caught up in some work.

PR #4972, opened 2018-12-30
Fix example alias URL

wlcrs:
Hi,

I noticed a small error in this example config file. To avoid confusion for future readers, I propose to fix the inconsistency.


gasman:
Looks good to me - thanks @wlcrs!

PR #4974, opened 2019-01-02
[WIP] UI overhaul

nmorduch:
So far I have:

- Rewritten `styles.scss` to have all of the includes for `core.scss` in the styles of ITCSS (still needs tweaking)
- Moved most styles to be in the new scss place, partly to make the includes in `style.scss` simpler.
- Moved the miscellaneous styles that were in `core.scss` to files in various places in preparation for rewriting them to use BEM and ~ITCSS
- Moved some styles from `components/typography` to various other files
- Made a `_helpers.scss` file for including settings and tools in scss files other 
than `core.scss`


jjanssen:
Hi @nmorduch,

This looks an impressive start of the new year! 🎆 I’m glad you found the time (and bravery 💪 ) to get your hands dirty on this. 

Within the Core UI team we’ve discussed on what we wanted the accomplish within our codebase in the long-run and this looks like a great first step. In addition to the long-run I think it would be wise to maybe setting goal(s) in what we want to accomplish during our first step and this pull request (aside from what you have done already). 

This way:
- We don’t review the irrelevant changes
- We don’t have an insanely large PR
- Could perhaps iterate faster on changes being merged (reduce the risk of [going in limbo for a long time](https://github.com/wagtail/wagtail/pull/3012))
- Divide work across members (or other community members)

The downside is that it would probably take up more time during cross-browser testing but I could actually arrange some time from our QA’ers if they know what the goals are 👍

nmorduch:
Thanks, @jjanssen! Once I have this cleaned up a little I'd love to get it merged so that further changes can be much smaller PRs.

My plan is to merge this before starting any changes that will affect styles at all. Currently the only substantial change is that `@font-face` gets into more css files than previously, which I did for simplicity but would be happy to change. I also think I might remove https://github.com/wagtail/wagtail/pull/4974/commits/8cce0ad81cacde6d74320975a66ece133b3b0c01 from this PR and save it for a later one.

nmorduch:
Closing in favor of #4979, which has a cleaner history, doesn't break watching for style changes, and and doesn't have https://github.com/wagtail/wagtail/pull/4974/commits/8cce0ad81cacde6d74320975a66ece133b3b0c01 since that was getting ahead of myself

nmorduch:
Oops thanks for closing. I must have hit the wrong button when I meant to do it.

PR #4924, opened 2018-11-26
Add hooks to page move action

maylonpedroso:
This PR adds `before_move_page` and `after_move_page` hooks to the Editor workflow.

Every editor action on a page (`create`, `edit`, `delete`, `copy`) has hooks to register new behaviors to them except for `move`, and we at Twilio need to use them all to make some verifications on the user permissions and page content before letting them proceed. 

- [x] Do the tests still pass? (http://docs.wagtail.io/en/latest/contributing/developing.html#testing)
- [x] Does the code comply with the style guide? (Run `make lint` from the Wagtail root)
- [x] For Python changes: Have you added tests to cover the new/fixed behaviour?
- [ ] For front-end changes: Did you test on all of Wagtail’s supported browsers? **No front-end changes**.
- [x] For new features: Has the documentation been updated accordingly?

PR #4975, opened 2019-01-02
Added before_move_page and after_move_page hooks to Editor page re-ordering (WIP)

joncombe:
I've been using wagtail only a tiny number of hours so I am not expecting this PR to be accepted. I add it here in case the solution is as simple as I think it might be.

Pull request https://github.com/wagtail/wagtail/pull/4924 added `before_move_page` and `after_move_page` hooks to the Editor workflow. It is my (newbie) view that these should also be called when re-ordering sibling pages, not just when moving up and down the tree. If you agree, does my PR do enough to fix the problem, please?

If what I suggest here somehow "pollutes" `before_move_page` and `after_move_page`, maybe adding something like `before_reorder_page` and `after_reorder_page` would be preferable? Advice gratefully accepted.

- [ ] For Python changes: Have you added tests to cover the new/fixed behaviour? **I think existing tests already cover this. Happy to be wrong.**
- [ ] For front-end changes: Did you test on all of Wagtail’s supported browsers? **No FE changes**

### DIFF
--- a/wagtail/admin/views/pages.py
+++ b/wagtail/admin/views/pages.py
@@ -791,6 +791,11 @@ def set_page_position(request, page_to_move_id):
     if not parent_page.permissions_for_user(request.user).can_reorder_children():
         raise PermissionDenied
 
+    for fn in hooks.get_hooks('before_move_page'):
+        result = fn(request, page_to_move, parent_page)
+        if hasattr(result, 'status_code'):
+            return result
+
     if request.method == 'POST':
         # Get position parameter
         position = request.GET.get('position', None)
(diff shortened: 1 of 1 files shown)


joncombe:
Thank you for taking the time to write such a considered response. 

My use case is simply to invalidate a cache item (you perhaps saw my question in the slack #support channel) and so a signal would clearly be better option for something as simple as that. I will look into how signals could be implemented here and likely open a new PR in due course. I'll leave you to close this ticket in case you want to add something new. Thanks again.

lb-:
Closing this PR as @joncombe advised it can be closed, plus #5169 has since been created that supersedes this.

Issue #3804, opened 2017-08-26
UI architecture technical debt


thibaudcolas:
Wagtail's UI codebase is in bad shape. This isn't really surprising – UI/CSS architecture in long-lived projects is a notoriously hard problem. It's not hopeless either. But there is work to do 🙂

I'd like to use this as a tracking issue to discuss the problems this is creating, their causes, potential solutions, and hopefully gather consensus. In my opinion, a good outcome of those conversations would be to agree on a prioritised list of actions to be taken, that can then be worked on by different contributors in separate PRs. Sweeping rewrites are not on the agenda.

> Edit: historically, the following content has been targeted towards CSS architecture in particular. But feel free to discuss any other UI maintainability issues (JS, HTML, fonts, SVG, etc) in this issue.

## Problems

- Maintenance is hard and error-prone. UI regressions are frequent, and hard to correct.
- Trying to implement new UIs, there is a big bias towards creating new styles instead of repurposing existing patterns.
- Third-party plugins have no clear point of integration with Wagtail's UI. [Anecdotal examples](https://github.com/springload/wagtaildraftail/blob/master/client/wagtaildraftail.scss#L8).

## Causes

This is a grab bag of code smells and high-level architectural issues.

- Main admin styles are kept in a few 100+ line files instead of being separated in components. (eg. `_main-nav.scss`, 479 lines)
- Too much nesting, generating overly-specific CSS selectors (eg. `.nav-main .footer-submenu a`).
- Styles rely on qualifying elements, generating overly-specific CSS selectors, making UI components harder to evolve to use different elements.
- No consistent naming convention. We are supposed to use BEM, but a lot of styles do not follow it, and it's not enforced.
- Too many magic values. This is in particular problematic for spacing, where there is no consistency between different UI components.
- Too little architecture. There is a basic separation of components in different files, layouts with their page-specific styles, but it does not go far enough. (eg. for whatever reason all of the components that are specific to the `page-editor` layout have been defined inside of it.
- No clear boundary for plugin integrations.

## Solutions

With #3773 as a foundation, I believe some of the above could easily be addressed by increasing what rules are enforced with tooling.

- Leverage more `stylelint` rules or plugins to enforce stricter guidelines
  - For qualifying elements, `selector-no-qualifying-type`.
  - For magic values, https://github.com/AndyOGo/stylelint-declaration-strict-value.
  - For naming,
[...]

PR #4979, opened 2019-01-03
Rearrange SCSS

nmorduch:
This PR is written as a step toward making changes called for in #3804. It moves the bulk of our styles to the `client` folder, and splits them into files approximately according to ITCSS.

The only changes to outputted CSS should be:
- The order of the styles
- `@font-face` is included in more CSS files because I grouped it with settings and tools. I'm not sure this is the right decision, but it does make the includes a lot simpler.

Replaces #4974


jjanssen:
Hi @nmorduch,

For the time-being I am not sure on whether `datetimepicker.scss` and `tags.tagit.scss` should be a part of our actual components folder. The context of this file is an override to an existing CSS namespace to which we have no control over, so this would be possibly confuse future contributors in how we manage this folder. Perhaps _"Generic"_ is too generic for this (🥁-tsss) but it would make sense to target a separate vendors folder for this (thus being an additional folder on top of the regular ITCSS folder structure). This way everyday knows this is not only the Wagtail-code we are targeting.
Other than this small point of interest this looks a good first iteration!

PS: 
I have asked @hvandenberg to do some cross-browser testing of this PR somewhere this week.
We’ll probably get his report on the testdrive somewhere this week within this PR 👍

nmorduch:
Sounds like a plan! I'll move those.

nmorduch:
I ended up changing the `utilities` folder to `overrides` (like I'd always wanted tbh) and giving it three subcategories: `vendor`, `utilities`, and `pages`. This way we don't have to argue about different kinds of overrides and can have them all, at least for now :)

tomdyson:
> I have asked @HvandenBerg to do some cross-browser testing of this PR somewhere this week.
We’ll probably get his report on the testdrive somewhere this week within this PR

@jjanssen do you know if there's been any progress with this?

HvandenBerg:
I've been testing and so far have found two things:
- 404 errors in console regarding fonts not being loaded
- Layout 'off' on IE11 / Too far to the right (this is not the case when not on PR)

See img's:

nmorduch:
Thank you, @HvandenBerg! I'm hoping to get to look into those problems this week.

nmorduch:
I can't say I've been looking forward to figuring out what precisely is making that happen, so I certainly wouldn't complain if you wanted to (ie yes i would love a hand on that)

jonnyscholes:
@nmorduch I'll have a look.

Is now a good time to solidify these opinions into [the CSS styleguide docs](https://github.com/wagtail/wagtail/blob/master/docs/contributing/css_guidelines.rst)? Or should we do that once the codebase is in a better state so we can more reasonably expect other developers to adhere to them...

jonnyscholes:
I couldn't reproduce the font issues showing in the console of @HvandenBerg's screenshot in IE11... I wonder if there was a build error?

jjanssen:
Hi Naomi and Jonny, 

I took the liberty of applying Jonny's PR to apply to this branch because ideally I would love to have this merged as soon as we can and I can probably can get Hugo to test it final this week. 

Aside from that I did some investigation into the font-issue. Right now this is caused as it is also rendered into subfiles like `css/layouts/home.css` whereas previously we only rendered the font-face definitions in `css/core.css`. When opening the homepage of the admin interface it tries to loads the font-set twice. Once from `css/core.css` (which resolves) and once from `css/layouts/home.css` which doesn't resolve (hence it only goes up two folders).

But I'm not sure on what the best practice would be to resolve it within this naming convention. Perhaps remove them from the general `_settings.scss` and only them to `core.scss`? I'm happy to hear your thoughts about this.

A screenshot from the difference between this branch and the latest master in terms of home.css:
image

nmorduch:
Thank you for doing this, @jjanssen.

How fonts are included is one of the few things I did actually change substantially. The way it was done before was a bit confusing to me, and maybe I need to understand a little better how we go about including css files in the templates. I was thinking that pages get  either `main.css` _or_ another css file, but I see that I was wrong and they get `main.css` _and_ other css files. I will take a look at this later today. I hope just to be able to put in a commit that fixes the font situation, but I might have further questions since I'm seeing some inconsistencies around how the css files are included in the templates.

thibaudcolas:
It's pretty cool to see all of those changes happening! What's the status of this PR @nmorduch @jjanssen @jonnyscholes?

Any other changes to Wagtail's styles will be creating conflicts for this refactor, so we should try not to keep this open for longer than necessary (otherwise it'll just become harder and harder to ever merge this).

jonnyscholes:
@thibaudcolas I actually did some work on this late last week to resolve the font issues raised by @jjanssen. I'll clean it up and commit it tomorrow - you're right it would be great to get this merged sooner rather than later!

jonnyscholes:
Turned out I'd already cleaned it up... Hope you don't mind @nmorduch but I've pushed up a fix for the fonts!

I've moved them out of `settings/_fonts.scss` and to the top of `elements/_typography.scss`. One nice side effect of this is settings now renders no CSS output so can be safely imported in any other file with no side effects. 

I did some testing and it looks like `core.css` is loaded on all pages and then some pages get an additional css file loaded. The commit just added should mean the fontface rules are only ever in `core.css` :)

jjanssen:
I think we're nearly there. Could we close this PR and create a new one so @HvandenBerg can benefit from using our new Squash tooling for that? It will save him the hassle to re-checkout this branch, build it, etcetera.

nmorduch:
This has been replaced by #5215

Issue #4970, opened 2018-12-27
{{ site_name }} is returning host name instead of site name


DanAtShenTech:
In [customizing the Wagtail login page](https://docs.wagtail.io/en/v2.4/advanced_topics/customisation/admin_templates.html?highlight=site_name%20#specifying-a-site-or-page-in-the-branding), I set up the following using {{ site_name }} as documented:

```
{% extends "wagtailadmin/login.html" %}
{% block branding_login %}Sign in to {{ site_name }}'s CMS{% endblock %}
```
I expected (and wanted) to see the site name on the login page, but instead it renders the host name.  Is this how it is supposed to behave?

### Technical details

* Python version: 3.6
* Django version: 2.1.4
* Wagtail version: 2.4
* Browser version: Chrome 71

allcaps:
Digging a bit:

Wagtail subclasses the Django LoginView.
The django.contrib.auth.views.LoginView#get_context_data uses the get_current_site method from django.contrib.sites.shortcuts this is based on the Django sites framework.

We don't need the Django Site we need a Wagtail site. Something like:
wagtail.core.sites.get_site_for_hostname or wagtail.core.models.Page#get_site

DanAtShenTech:
Thank you.  I won't have time to do anything on this for at least several months so am hopeful someone else can pick up on this.

edbrown:
Had a crack at this, check the PR here: #4987

andruwwwka:
I fix it for non-admin login view and edit test for this. Please check Pull Request #4988

gasman:
I don't think this is a bug. The documentation for [`site_name` and the related template variables](https://docs.wagtail.io/en/v2.4/advanced_topics/customisation/admin_templates.html#specifying-a-site-or-page-in-the-branding) indicates that these are only available on the admin dashboard page - in particular, their values are defined in terms of the logged-in user's permissions, which doesn't make sense for the login page.

At the moment, it's a deliberate design principle of Wagtail that there is a single 'global' instance of the admin, which provides access to the full page tree (or whatever subset of it the current user has permission over) regardless of the hostname the user is accessing it through. This has a number of benefits over conceptually giving each subsite its 'own' admin:

* it allows suitably high-level users to perform administration tasks across the whole set of sites in a single login session, instead of having to log in individually to each site
[...]

PR #4987, opened 2019-01-11
Issue #4970 Replace site_name with Wagtail site name.

edbrown:
Gave this issue #4970 a shot - replacing the site_name context with the Wagtail site name.


gasman:
Thanks @edbrown! As discussed on #4970, unfortunately there are some subtleties about the way the Wagtail admin is implemented that mean we can't make this work consistently after all (for example, when the admin is hosted on a separate domain to the sites) - but thank you for taking the time to work on this.

PR #4988, opened 2019-01-12
Issue #4970 render wagtail site name

andruwwwka:
Issue #4970 render wagtail site name to non-admin login view and test on this

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4196,14 +4196,12 @@
         "balanced-match": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
           "dev": true,
-          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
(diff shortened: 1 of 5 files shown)


gasman:
Thanks for the contribution, @andruwwwka!

My concern here is that we're adding an extra database query to the request, that most developers are unlikely to need. For some time there's been a general wish to eliminate unnecessary queries on front-end pages (see discussion on #3834) - if we add this variable to the template context now, then it's going to be hard to remove it again in future when people are relying on it.

For developers who want to display the site name on the login page, it's simple to access that from `request.site` (see the code example on https://github.com/wagtail/wagtail/issues/4970#issuecomment-455264383), so in my opinion the benefit gained from the new `site_name` variable doesn't justify the extra database hit.

One possible improvement would be to implement a template tag to fetch the site name, so that the query only happens on sites that need it:

    {% load wagtailcore_tags %}

    <h1>Log in to {% wagtail_site_name %}</h1>

or

    {% load wagtailcore_tags %}

    {% wagtail_site_name as site_name %}
    <h1>Log in to {{ site_name }}</h1>

Issue #4993, opened 2019-01-17
Syntax error in tsquery with Postgres and wagtail.contrib.postgres_search


hammygoonan:
### Issue Summary

I'm getting an error saying `django.db.utils.ProgrammingError: syntax error in tsquery: "'test\'"` whenever I use a `\` in my search term.

### Steps to Reproduce

1. Start a new project with `wagtail start myproject`
2. Configure site to use Postgres and the `wagtail.contrib.postgres_search` app.
3. Run migrations
4. log into admin
5. add a `\` character to your search term.

Running locally, you replicate the issue by using Postgres, `postgres_search` then going to: [http://127.0.0.1:8000/admin/pages/search/?q=drink%5C](http://127.0.0.1:8000/admin/pages/search/?q=drink%5C).

The error is with all the searches (page, image and document).

Any other relevant information. For example, why do you consider this a bug and what did you expect to happen instead?

I assume the `\` character should just be stripped out of search queries.

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: yes

### Technical details

* Python version: Python 3.7.2
* Django version: 2.1.5
* Wagtail version: 2.4
* Browser version: Firefox 64

BertrandBordage:
FIxed by #4994.

PR #4994, opened 2019-01-17
Tsquery fix

hammygoonan:
* For Python changes: Yes, but those tests seem to be skipped for some reason.
* For front-end changes: N/A
* For new features: N/A

A fix for [this](https://github.com/wagtail/wagtail/issues/4993) issue where backslashes break the postgres search.

### DIFF
--- a/wagtail/contrib/postgres_search/backend.py
+++ b/wagtail/contrib/postgres_search/backend.py
@@ -228,7 +228,9 @@ class PostgresSearchQueryCompiler(BaseSearchQueryCompiler):
                 return self.get_search_field(sub_field_name, field.fields)
 
     def prepare_word(self, word):
-        return unidecode(word)
+        # remove backslash from string to prevent syntax error in tsquery
+        stripped_word = word.replace("\\", "") 
+        return unidecode(stripped_word)
 
     def build_tsquery_content(self, query, group=False):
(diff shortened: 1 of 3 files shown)


wagtail/contrib/postgres_search/backend.py
```diff
@@ -228,7 +228,9 @@ def get_search_field(self, field_lookup, fields=None):
                 return self.get_search_field(sub_field_name, field.fields)
 
     def prepare_word(self, word):
-        return unidecode(word)
+        # remove backslash from string to prevent syntax error in tsquery
+        stripped_word = word.replace("\\", "")
```
BertrandBordage: This is removing an important character that can be separating words: we must escape backslashes, not remove them.
According to [the tsquery documentation](https://www.postgresql.org/docs/11/datatype-textsearch.html), backslashes must be doubled, though no example is available.
Also, a better place to put it would be in `RawSearchQuery`.
So please update your PR to remove this change and add new `.replace('\\', '\\\\')` at this line: https://github.com/wagtail/wagtail/blob/ddc5f1c21c74a1ea9a2ae819fe1dfce419655808/wagtail/contrib/postgres_search/models.py#L21

hammygoonan: Make sense. I've made that update.

PR #4998, opened 2019-01-17
Amend lookup_allowed to work with ParentalKey

jasondilworth56:
ParentalKey needs an extra remote_field step to get to the necessary rel_name used in lookup_allowed.


gasman:
Thanks @jasondilworth56! Please could you add test coverage for this?

gasman:
@jasondilworth56 Thanks for adding the test!

My first thought is that we should be checking for `ForeignKey` rather than `ParentalKey` - `ParentalKey` is a special case of `ForeignKey` whose special case behaviour is unconnected to filter lookups, so any failure cases applying to `ParentalKey` in this bit of code will apply more generally to `ForeignKey`.

Looking deeper, though, I'm wondering if `lookup_allowed` is actually working as intended here. As far as I can see, it's meant to provide some sanity-checking on lookup parameters before running `QuerySet.filter`, but the logic of what it accepts or rejects seems pretty arbitrary:

* it [allows any single-clause filter parameter](https://github.com/wagtail/wagtail/blob/be03eeee2ea882d2cbd3e8e65fe4618794f3fd2c/wagtail/contrib/modeladmin/views.py#L336-L337) (i.e. anything with no underscores), so a URL parameter like `dfghdf=dfgnjfgh` will pass the check and ultimately throw an exception during the `QuerySet.filter` call
* The closing `return clean_lookup in self.list_filter` seems to imply that we only want to allow filters that are explicitly named in `list_filter`, but that evidently isn't happening given that the new test succeeds (it's filtering on `related_links__link_page__id__exact`, which definitely isn't in EventPageAdmin's `list_filter`...)
* The `for part in parts[:-1]:` loop looks iffy - it's setting `model` and then never using it. I suspect it should really be doing something like:

        model = self.model
        for part in parts[:-1]:
            field = model._meta.get_field(part)
            if hasattr(field, 'remote_field'):
                # update 'model' to point to the next model in the sequence of relations
                model = field.remote_field.model

If I [remove the call to `lookup_allowed`](https://github.com/wagtail/wagtail/blob/be03eeee2ea882d2cbd3e8e65fe4618794f3fd2c/wagtail/contrib/modeladmin/views.py#L359-L362), all tests (including the new one) still pass, so it doesn't seem to be doing anything useful - and so I'd be inclined to go for the "nuclear option" of removing `lookup_allowed` completely. I'm not entirely keen on passing parameters from the URL into `QuerySet.filter` unchecked, but if (as I suspect) `lookup_allowed` is not effective at validating them, then it's no worse then the status quo. @ababic, what do you think?

(If we do decide to take the more thorough approach and fix up `lookup_allowed` to perform the intended validation, then it would probably be worth commenting out [the corresponding code in Django admin](https://github.com/django/django/blob/7c68cea7ee406fcb1b632e3a1a7f6f53e1423839/django/contrib/admin/views/main.py#L108-L110), seeing what tests break, and ensuring we have equivalent test coverage in modeladmin.)

ababic:
@gasman Thank you for the thorough feedback. I'll have a trawl through Django's version and see what this is really supposed to be doing, as it could well be that the method was badly ported.

jasondilworth56:
Thanks @gasman, a shame it's not a simple merge but hey ho!

@ababic I'm happy to try to help with what has potentially become a more in-depth fix than I was expecting, let me know if there's something specific I can be useful with and I'll do that so we're not duplicating work.

ababic:
Hi @jasondilworth56. If you feel confident enough in evaluating the [equivalent code in Django](https://github.com/django/django/blob/7c68cea7ee406fcb1b632e3a1a7f6f53e1423839/django/contrib/admin/views/main.py#L108-L110), and identifying what might not have been ported properly (or if indeed the code is all that important), that should would help decide what the next step is.

Personally, I'm leaning towards removing this code completely, because administrators are always authenticated when using the view, and it's difficult to see what benefit it is adding.

frmdstryr:
Ran into this error today. 

The modeladmin's `lookup_allowed` looks an awful lot like it was ripped from django-xadmin's implementation in [xadmin/plugins/filters.py](https://github.com/sshwsfc/xadmin/blob/f209bf6e1d60932a88a06a097f8b322935e78146/xadmin/plugins/filters.py) (given that the comments are even the same and xadmin's history goes back to 2013). 

Anyways the whole lookup_allowed check is disabled by the `free_query_filter` in xadmin, removing this from wagtail works for what I'm doing, so another vote for removing this.

jasondilworth56:
Also agreed, looks like @frmdstryr is right @ababic. Looks like removal is the way to go.

ababic:
Thank you for the input @frmdstryr / @jasondilworth56. Removing this will be satisfying :)

@jasondilworth56 would you be happy to submit a new PR for this?

gasman:
Have now created a replacement PR #5226 which simply removes the `lookup_allowed` validation.

PR #5002, opened 2019-01-22
Mattlinares recipe addition

mattlinares:

### DIFF
--- a/docs/reference/pages/model_recipes.rst
+++ b/docs/reference/pages/model_recipes.rst
@@ -201,7 +201,7 @@ This is just one possible way of creating a taxonomy for Wagtail objects. With a
 Have redirects created automatically when changing page slug
 ------------------------------------------------------------
 
-You may want redirects created automatically when a url gets changed in the admin so as to avoid broken links. You can add something like the following block to `/models/wagtail_hooks.py`.
+You may want redirects created automatically when a url gets changed in the admin so as to avoid broken links. You can add something like the following block to a ``wagtail_hooks.py`` within one of your project's apps.
 
 
 .. code-block:: python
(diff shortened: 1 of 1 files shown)


docs/reference/pages/model_recipes.rst
```diff
@@ -196,3 +196,31 @@ Here, ``blogs.filter(tags__name=tag)`` invokes a reverse Django QuerySet filter
 Iterating through ``page.tags.all`` will display each tag associated with ``page``, while the link(s) back to the index make use of the filter option added to the ``BlogIndexPage`` model. A Django query could also use the ``tagged_items`` related name field to get ``BlogPage`` objects associated with a tag.
 
 This is just one possible way of creating a taxonomy for Wagtail objects. With all of the components for a taxonomy available through Wagtail, you should be able to fulfil even the most exotic taxonomic schemes.
+
+
+Have redirects created automatically when changing page slug
+------------------------------------------------------------
+
+You may want redirects created automatically when a url gets changed in the admin so as to avoid broken links. You can add something like the following block to `/models/wagtail_hooks.py`.
```
gasman: `/models/wagtail_hooks.py` would only be a correct path if you had an app named "models" - I'd suggest saying "You can add something like the following block to a `wagtail_hooks.py` within one of your project's apps". (Also, note that the filename should be in double backticks rather than single - an annoying RestructuredText quirk there...)

docs/reference/pages/model_recipes.rst
```diff
@@ -196,3 +196,31 @@ Here, ``blogs.filter(tags__name=tag)`` invokes a reverse Django QuerySet filter
...
+    @hooks.register('before_edit_page')
+    def create_redirect_on_slug_change(request, page):
+        try:
+            if page.slug != request.POST['slug']:
+                Redirect.objects.create(
+                        old_path=page.url[:-1],
+                        site=page.get_site(),
+                        redirect_page=page
+                    )
+        except:
```
gasman: I'd rather not encourage people to use blanket `try` / `except` blocks, as they risk hiding legitimate errors that the developer should know about. Is there a specific exception that this code is likely to throw (and can be legitimately ignored)? If so, we should specify that exception in the `except`.

mattlinares: Yes agreed. That was initially there since this hook is called both when opening the Page edit admin page (GET request), and also when saving the page (POST), so this was an easy (bad) way to distinguish between them, as well as suppressing any unseen errors.

I have altered that, I hope that’s more appropriate?

https://github.com/wagtail/wagtail/pull/5002/files#diff-75e7c538654bfd89118b597d9090b334

docs/reference/pages/model_recipes.rst
mattlinares: Good point. Have changed that.

gasman:
Looks good to me! Merged in a0b4efd4147cf5789ba1d1c75829b6f1859667b2.

Issue #4973, opened 2019-01-01
Supporting proxy models throughout Wagtail


ababic:
### Issue Summary

While there are a number of issues/PRs relating to support of proxy models in various parts of Wagtail (#1736, wagtail/wagtail-modeladmin#16, wagtail/wagtail#1029, wagtail/wagtail#4273, wagtail/wagtail#3594), I feel that we're missing a trick by only looking at the those issues in isolation. This issue aims to discuss the matter more generally, with an aim to establishing a clean, consistent approach that can be applied throughout the wider project.

### Where are we now? 

There are a good few places in Wagtail where using proxy models fails to work as expected. The exact details vary from place to place, but I feel they can all really be summarised as follows:

1. Django creates `Permission` objects for each proxy model, but they are associated with the `ContentType` for the concrete model. 
2. When we want to identify the `ContentType` for a model, we tend to use `ContentType.objects.get_for_model()` which, when provided with a proxy model, returns the `ContentType` for the concrete model by default.
3. `ContentType.objects.get_for_model()` supports a `for_concrete_model` option (which has a default value of `True`), which can be used to retrieve a unique `ContentType` object for each proxy model instead of that of the concrete model. But, `Permission` objects for the proxy models remain associated with the `ContentType` of the concrete model.
4. Because of all of the above, identifying the correct `ContentType` and `Permission` objects for a proxy model (or worse, identifying the relevant model from a `Permission` or `ContentType` object) becomes tricky.

### Where do we want to be?

I think we can all agree that supporting proxy models is much simpler/cleaner when proxy models have their own `ContentType`, and model permissions are related to that `ContentType`. In fact, this is how Django intends for things to be from version [2.2](https://docs.djangoproject.com/en/dev/releases/2.2/) 

There's a clear desire for proxy model support in Wagtail, and the changes in Django 2.2 will make that easier, but changing our code to benefit from those changes would leave a gap in proxy model support in earlier Django versions, which would be nice to fill somehow. 

We want using proxy models in Wagtail to be simple for developers, but we don't want to force wholesale data changes on to every project where Wagtail is installed, so there should at least be an 'opt-in' step.

## How might we get there?

Below are the most up-to-date recommendations:

1. Update Wagtail to use `for_concrete_model=False` and `for_concrete_models=False` options throughout, wherever `ContentType.objects.get_for_model()` or `ContentType.objects.get_for_models()` are used (respectively).
[...]

PR #5003, opened 2019-01-23
Supporting proxy models throughout Wagtail [WIP]

ababic:
Proposed implementation for #4973

The new `wagtail/utils/contenttypes.py` module is where you should direct most attention. Everything else is basically switching out every use of `ContentType.objects.get_for_model()` and `ContentType.objects.get_for_models()` to use the equivalent methods from the new module instead (ensuring a consistent approach throughout).

Still lots to do in terms of testing, but posting up early to start collecting feedback.

To do list:
- [ ] More manual testing
- [ ] Add unit tests for `wagtail.utils.contenttypes.get_for_model()`
- [ ] Add unit tests for `wagtail.utils.contenttypes.get_for_models()`
- [ ] Update documentation to inform users of backwards incompatible changes (`ContentType` and `Permission` changes for proxy models)

## About the changes

Functionally speaking, the new methods in `wagtail/utils/contenttypes.py` work in a very similar way to the equivalent methods on `ContentTypeManager`, and utilise the manager's built-in cache to provide faster look-ups after the first time. Developers going via `ContentTypeManager` directly will also benefit from that caching, because the cache is shared.

The 2 big differences are:
 
1. The complication of `for_concrete_model`/`for_concrete_models` method arguments is removed, because (within Wagtail) we always want to identify specific ContentTypes for models, regardless of whether they are a proxy model or not (and regardless of Django version). 
2. When a `ContentType` isn't found for a model, and a new object is created (which is something the `ContentTypeManager` methods do also); For proxy models, a little extra work is done to ensure related `Permission` objects are correctly associated with the new `ContentType` (the part that was missing from #1736). From Django 2.2, this part is taken care of automatically at migration time, so we toggle this behaviour based on Django version, so as not to confuse things.

## Known limitations / issues

The following would only affect developers that are using proxy models in their Wagtail project already AND either:

1. Have custom code that relies on `Permission` objects remaining associated with the `ContentType` of the concrete model (e.g. they identify model permissions by using `ContentType.objects.get_for_model()`) [1]
2. Have created `ContentType` objects for proxy models themselves, but not reallocated the `Permission` objects [2]
3. Are using the Django admin area in addition to the Wagtail one, have Django < 2.2 installed, and proxy models are administered via the Django admin [3]

[1] Because our code only works on proxy models registered with Wagtail (rather than affecting all models) this might make identifying `Permission` and `ContentType` objects tricky, due to inconsistency.

[2] This could have been intentional, or an unintended result of calling `ContentType.objects.get_for_model(for_concrete_model=False)`. In these cases, the existing `ContentType` would be picked up and used as is, without the `Permission` reallocation logic being triggered.

[3] As yet unclear about the affect this this would have. The permission reallocation logic would leave existing `User` -> `Permission` associations in-tact, but I think the Django admin might not pick up the new `ContentType` objects in some places? @arthurio - can you perhaps share some light on this one?

### DIFF
--- a/wagtail/admin/utils.py
+++ b/wagtail/admin/utils.py
@@ -5,7 +5,6 @@ import pytz
 
 from django.conf import settings
 from django.contrib.auth import get_user_model
-from django.contrib.auth.models import Permission
 from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import PermissionDenied
 from django.core.mail import get_connection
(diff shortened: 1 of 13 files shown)


auvipy:
why this was closed?

arthurio:
@auvipy I raised few concerns on the approach in this PR in the linked issue https://github.com/wagtail/wagtail/issues/4973#issuecomment-457829035

PR #5004, opened 2019-01-23
Ignore extra end tags in contentstate converter

frmdstryr:
This adds a check if an extra end tag exists (ex if importing/migrating data from other platforms) and avoids a "pop from empty list" error when this is the case.  

<hr/>

* Do the tests still pass? (http://docs.wagtail.io/en/latest/contributing/developing.html#testing) YES


gasman:
Perfect, thanks @frmdstryr! Merged in 5fcaa64d81ecd2f3d3da796de9b1ef8b28f3faa2.

frmdstryr:
Thanks!

PR #5009, opened 2019-01-25
Add 'wagtail_update_index' as an alias for the 'update_index' command

### LINKED ISSUE #2447: Prefix management commands
The update_index management command is already clashing with haystack's.
AFAIC wagtail management commands could be prefixed with 'wt_' to prevent this in the future.

gasman:
Fixes #2447

### DIFF
--- a/docs/reference/management_commands.rst
+++ b/docs/reference/management_commands.rst
@@ -93,6 +93,14 @@ You can prevent the ``update_index`` command from indexing any data by using the
     $ python manage.py update_index --schema-only
 
 
+.. _wagtail_update_index:
+
+wagtail_update_index
+--------------------
+
+An alias for the ``update_index`` command that can be used when another installed package (such as `Haystack <http://haystacksearch.org/>`_) provides a command named ``update_index``. In this case, the other package's entry in ``INSTALLED_APPS`` should appear above ``wagtail.search`` so that its ``update_index`` command takes precedence over Wagtail's.
+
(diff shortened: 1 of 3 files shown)


kaedroho:
I've just tried locally and it looks like you can have colons in command names, so ``wagtail:update_index`` should be possible. I kinda prefer the colon as a different separator makes it clear that "wagtail" is a namespace and not actually part of the command name.

gasman:
@kaedroho Does that require putting a colon in the filename, or is there a way to set the command name in the file itself? I'd prefer to keep our filenames as "normal" as possible, given the trouble we've had with gitlab and cookiecutter's filenames with braces...

I don't see a reason to spend more time adding aliases on other commands that haven't had this issue, but I won't object if someone else wants to do it :-)

kaedroho:
It does require adding colons into filenames, and I've just found that this will probably be an issue for Windows users as well.

ababic:
I'm unable to run the command locally due to an issue with my `vagrant-wagtail-develop` setup (https://github.com/wagtail/vagrant-wagtail-develop/issues/22), but from a code/docs perspective, this looks spot on.

ababic:
Okay, local issues sorted, and this is working fine for me. Merge away :)

gasman:
Thanks @ababic! Now merged in 61e1e4c2f961d854ad8dc679f58174eacdaf940a.

PR #5013, opened 2019-01-29
Disable codecov PR checks

gasman:
Codecov checks on PRs are unnecessary (it's not essential for every PR to have test coverage equal or greater than the codebase as a whole) and are training us to ignore red crosses on PR checks, so we should disable them.

(Please don't merge this as-is, since it contains a dummy commit to confirm that the config change is indeed taking effect)

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,4 +1,8 @@
 codecov:
+  status:
+    patch:
+      default:
+        enabled: no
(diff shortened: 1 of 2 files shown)


gasman:
Okay, that's exactly the opposite of what was supposed to happen... 😠

PR #5014, opened 2019-01-29
Disable codecov PR checks

gasman:
Second attempt...

Codecov checks on PRs are unnecessary (it's not essential for every PR to have test coverage equal or greater than the codebase as a whole) and are training us to ignore red crosses on PR checks, so we should disable them.

(Please don't merge this as-is, since it contains a dummy commit to confirm that the config change is indeed taking effect)

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -2,4 +2,10 @@ codecov:
   notify:
     after_n_builds: 10
     require_ci_to_pass: no
+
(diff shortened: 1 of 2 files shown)


gasman:
Applied in 6dba61fc3ad516593344d82f32fcb1cbf9dc25f8

PR #5043, opened 2019-01-30
Add docs with instructions on how to generate urls for modeladmin

### LINKED ISSUE #4519: Document reversing ModelAdmin URLs
It's not currently explained on ModelAdmin docs (Wagtail 2.0.1).

lb-:
Documentation on how to generate URLs within model using url reversing.

Successfully built the docs with these changes and tested all internal links.

Closes #4519


ababic:
Hi @lb-. Thank you for this! There's some good and useful information here. However, I do feel like the structure is a little inconsistent with what we have already. 

In a perfect world, there'd already be a separate API reference in the docs for the various helper classes, but there isn't, and I don't think we should try to fill that gap here. Rather, I think this page would be benefit from shrugging off the structural constraints of existing content, and focus on showing specific, practical examples of how to do this. What do you think?

In regards to quoting primary key values, that should only be necessary when working with strings, that might possibly contain 'tricky' characters. In most cases, it's unnecessary (the helper should probably be a bit _more_ helpful here, and conditionally quote the value itself).

lb-:
@ababic  - thanks for the feedback, just so I understand correctly, I should do the following:

* [ ] remove the content about `ensure you wrap any data in ``quote`` to ensure it is safe to use in a URL.` and the usage of `quote` function on the primary key in examples
* [ ] Rework the `url_helper.get_action_url` section to be maybe just show one solid example of getting the edit page for an entity. (then make a note at the end that delete, inspect and choose_parent are also available
* [ ] Rework the `url_helper.index_url` to be a solid example about 'showing a list of all entities for a model' and make a small note that this is a shortcut way to a cached version of `url_helper.get_action_url('index')`
* [ ] Similar with the `url_helper.create_ur'` section, work with it as an example of how to get a url to 'add' or 'create' a entity via a form
* [ ] Remove the `**Expected value**` specifications
* [ ] In each example, show what kind of URL you will get as an output so the examples are clearer

ababic:
> Remove the content about `` ensure you wrap any data in ``quote`` to ensure it is safe to use in a URL. `` and the usage of `quote` function on the primary key in examples

I would start with a simple example, then explain why/when quoting might be necessary -  possibly showing another example using `quote()`. If the functionality is revised later, then this bit can be rejigged at a later time.

> In each example, show what kind of URL you will get as an output so the examples are clearer

Yes, I think some code examples with sample output would be incredibly useful here :)

In general, I would try to steer clear of headings like `url_helper.get_action_url()` and instead use more informal language like "Getting the list view URL" and "Getting the edit view URL for an object". We can add some more formal API documentation for the helper classes and their methods at a later time.

lb-:
Thanks, that is really helpful, will get on it soon

lb-:
@ababic  - round two just updated, would love your feedback, also found a few typos in original and one example that was wrong (the behaviour of 'choose_parent' works slightly differently as originally written).

ababic:
Thank you @lb-. I'll take a good look at this soon. It might not be for another couple of weeks though, unfortunatey, as I currently have quite a lot on.

docs/reference/contrib/modeladmin/reversing_urls.rst
```diff
@@ -0,0 +1,115 @@
...
+For example:
+
+.. code-block:: python
+
+    from .wagtail_hooks import AuthorModelAdmin
+
+    url_helper = AuthorModelAdmin().url_helper
+
+    index_url = url_helper.get_action_url('index')
+    # OR we can use the cached property which is easier
```
ababic: Suggest replacing:

```
# OR we can use the cached property which is easier
```

With 

```
# OR we can use the 'index_url' shortcut
```

docs/reference/contrib/modeladmin/reversing_urls.rst
```diff
@@ -0,0 +1,115 @@
...
+    also_index_url = url_helper.index_url # remember do not call this property as a function
+    # will output /admin/my-app/author
+
+    create_url = url_helper.get_action_url('create')
+    # OR we can use the cached property which is easier
```
ababic: Suggest replacing:

```
# OR we can use the cached property which is easier
```

With 

```
# OR we can use the 'create_url' shortcut
```

lb-:
@ababic  - great feedback, thank you for taking the time to suggest the changes and explain all of that. I will get onto this soon.

docs/reference/contrib/modeladmin/reversing_urls.rst
```diff
@@ -0,0 +1,115 @@
...
+        # only yield a button if we know an author is linked
+        if page_author:
+            url_helper = AuthorModelAdmin().url_helper # initialise modeladmin
+            edit_author_url = url_helper.get_action_url('edit', page_author.pk)
+            # edit_page_author_url will output /admin/my-app/author/edit/2/
+            yield wagtailadmin_widgets.PageListingButton(
+                'Edit Author',
+                edit_author_url,
+                priority=10
+            )
```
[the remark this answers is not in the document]
lb-: Good call on this, I did not even know about the memory implications of creating a new instance each time / nor did I know you could just create an instance here and use it in multiple functions. thanks :)

docs/reference/contrib/modeladmin/reversing_urls.rst
```diff
@@ -0,0 +1,115 @@
...
+    :depth: 1
+
+-------------------------------------------
+Getting the URL for Edit, Delete or Inspect
+-------------------------------------------
+
+In this example, our goal is to provide a quick way to ``edit`` the Author
+that is linked to a blog post from the Admin Page Listing menu. Where we
+have a ``modeladmin`` called ``AuthorModelAdmin`` which uses the model
+``Authors`` and our ``BlogPage`` model has a foreign key to a single Author.
```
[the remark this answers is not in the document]
lb-: Good call, I basically just copied some models from bakery demo, did not think about the naming conventions + explaining it to users looking at this for the first time.

lb-:
@ababic  - revision updated, take a look when you get the chance.

docs/reference/contrib/modeladmin/reversing_urls.rst
```diff
@@ -0,0 +1,126 @@
...
+    also_create_url = url_helper.create_url # note: do not call this property as a function
+    # both will output /admin/my-app/author/create
+
+.. note::
+
+    When creating a new page via ``modeladmin`` it can be created in multiple
+    places, there is an additional action ``'choose_parent'`` which is used to
+    select the parent **before** creation of a page. There should be no need to
+    navigate to this action directly as navigating to the ``create`` URL will
+    redirect the user to choose a parent if necessary.
```
ababic: I think this paragraph might be better expressed as:

```
If you have registered a page type with ``modeladmin`` (e.g. ``BlogPage``), and pages of that type can be added to more than one place in the page tree, when a user hits the `create` URL, they'll be automatically redirect to another view, prompting them to choose a parent for the new page. So, this isn't something you need to check or cater for in your own code.
```

docs/reference/contrib/modeladmin/reversing_urls.rst
```diff
@@ -0,0 +1,126 @@
...
+    # ensure our modeladmin is created
+    class AuthorModelAdmin(ModelAdmin):
+        model = Authors
+        menu_order = 200
+
+    # Creating an instance of `AuthorModelAdmin`
+    author_modeladmin = AuthorModelAdmin()
+
+    @hooks.register('register_page_listing_buttons')
+    def page_listing_buttons(page, page_perms, is_parent=False):
```
ababic: How do you feel about changing the name of this method to `add_author_edit_buttons`?

docs/reference/contrib/modeladmin/index.rst
```diff
@@ -48,6 +48,7 @@ Want to know more about customising ``ModelAdmin``?
     create_edit_delete_views
     inspectview
     chooseparentview
+    reversing_urls
```
ababic: I think it might be a good idea to create a new section here, rather than having the `reversing_urls` page at the same level. I can easily imagine other helpful pages being added in future, that might fall better into a `tips-and-tricks` or section - open to suggestions on the naming there :). What do you think @lb-?

docs/reference/contrib/modeladmin/reversing_urls.rst
```diff
@@ -0,0 +1,126 @@
...
+.. code-block:: python
+
+    # file: wagtail_hooks.py
+
+    from wagtail.admin.widgets import PageListingButton
+    from wagtail.contrib.modeladmin.options import ModelAdmin, modeladmin_register
+    from wagtail.core import hooks
+
+    # Authors & BlogPage model not shown in this example
+    from models import Authors
```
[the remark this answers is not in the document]
lb-: sweet, will fix

docs/reference/contrib/modeladmin/reversing_urls.rst
lb-: sweet done

docs/reference/contrib/modeladmin/index.rst
lb-: yeah good point, this is not really part of `Want to know more about customising ModelAdmin?` - I have put a new section at the end of the page, see what you think

lb-:
@ababic - next round up, thanks for pushing to make this awesome, let me know what you think

docs/reference/contrib/modeladmin/reversing_urls.rst
```diff
@@ -0,0 +1,126 @@
...
+
+* ``'edit'``
+* ``'delete'``
+* ``'inspect'`` Note: Requires inspect enabled in modeladmin configuration.
+
+
+.. note::
+    If your Primary Key for the model is likely to contain characters that
+    are not URL safe, ensure you wrap any data in ``quote`` from
+    ``django.contrib.admin.utils``.
```
ababic: I don't think the title casing of "Primary Key" is really needed here (lower case is fine) :) Also, I think we can be a touch more precise about what data must be wrapped. Would suggest something like:

```
If you are using string values as primary keys for you model, and it's possible key values could contain characters that are not URL safe, you should use ``django.contrib.admin.utils.quote()`` to prepare primary key values before passing them to ``get_action_url()``. Failure to do this may result in Wagtail failing to recognise the primary key when the URL is visited, resulting in a 404 error response.
```

lb-:
@ababic  - made those changes, with a few refinements from your comments.

thanks

docs/reference/contrib/modeladmin/reversing_urls.rst
```diff
@@ -0,0 +1,130 @@
...
+.. contents::
+    :local:
+    :depth: 1
+
+-------------------------------------------------------------------
+Getting the ``edit`` or ``delete`` or ``inspect`` URL for an object
+-------------------------------------------------------------------
+
+In this example, we will provide a quick way to ``edit`` the Author that is
+linked to a blog post from the Admin Page Listing menu. We have defined
```
ababic: > Admin Page Listing

I don't think the title casing is needed here.

lb-:
@ababic  - cool, thanks for picking those up, I have reworked that note section about the url encoding of primary keys, hope it makes a bit more sense now.

docs/reference/contrib/modeladmin/reversing_urls.rst
```diff
@@ -0,0 +1,131 @@
...
+=========================
+Reversing ModelAdmin URLs
+=========================
+
+It's sometimes useful to be able to derive the ``index`` (listing) or
+``create`` URLs for a model along with the ``edit``, ``delete`` or
+``inspect`` URL for a specific object in a model you have registered via
+the ``modeladmin`` app.
+
+Wagtail itself does this by instantiating of each ``ModelAdmin`` class you have
```
ababic: > instantiating of each

Should be

> instantiating each

lb-:
@ababic  - thanks for that, fixed up in latest push. let me know if we are good to go :)

ababic:
LGTM :)

lb-:
@ababic  - thanks for all the help, really appreciated it.

Merged in https://github.com/wagtail/wagtail/commit/fc21729a243e7ac76ac00b680f9a256c0bacc776

gasman:
Building the docs now gives me the warning "/vagrant/wagtail/docs/reference/contrib/modeladmin/reversing_urls.rst: WARNING: document isn't included in any toctree" - does this need to be added to an index page?

ababic:
@gasman sorry, this one's on me. will sort :)

lb-:
@ababic  - thanks for that - will leave with you

I was not sure where to put the page in the TOC as it was moved to the 'tips and tricks' section

PR #5046, opened 2019-01-31
Add MEDIA_* settings to Django integration page

davidjb:
This change adds the required MEDIA_URL/MEDIA_ROOT settings to this documentation page, so the page outlines they are required and gives an example if they're not already configured.


gasman:
Thanks @davidjb! Looks good to me - merged in f881420a98a45303c767f28d761a1791e858da7c now.

PR #5052, opened 2019-02-02
Fix for emails missing HTML messages.

tr11:
This PR fixes the `send_mail` function to send HTML emails as expected.

### DIFF
--- a/wagtail/admin/utils.py
+++ b/wagtail/admin/utils.py
@@ -222,13 +222,13 @@ def send_mail(subject, message, recipient_list, from_email=None, **kwargs):
         password=kwargs.get('auth_password', None),
         fail_silently=kwargs.get('fail_silently', None),
     )
-    kwargs = {
+    multi_alt_kwargs = {
         'connection': connection,
         'headers': {
             'Auto-Submitted': 'auto-generated',
         }
     }
-    mail = EmailMultiAlternatives(subject, message, from_email, recipient_list, **kwargs)
+    mail = EmailMultiAlternatives(subject, message, from_email, recipient_list, **multi_alt_kwargs)
     html_message = kwargs.get('html_message', None)
     if html_message:
         mail.attach_alternative(html_message, 'text/html')


overshard:
I was just hit by this problem too, had to swap to using the default django send_mail in all my code.

lb-:
@tr11 - thanks for finding this issue and supplying a fix, this is awesome!

One thing that would help is adding a test so this does not get lots / regressed in the future, here is one I roughly got working locally, are you ok to add this to the PR (please review/lint it - it may need some tweaks).

I had to look into how the [Django 'EmailMessage' object](https://github.com/django/django/blob/77d25dbd0f20d6a907c805ffae8aaadd87edbacf/django/core/mail/message.py#L415) gets set up to test the inner structure.

Let us know if you are able to update the PR.

```python
    #file: wagtail/admin/tests/tests.py
    # inside class TestSendMail(TestCase):
    def test_send_html_email(self):
        """Test that the kwarg 'html_message' works as expected on send_mail by creating 'alternatives' on the EmailMessage object"""

        send_mail("Test HTML subject", "TEXT content", ["has.html@email.com"], html_message="<h2>Test HTML content</h2>")
        send_mail("Test TEXT subject", "TEXT content", ["mr.plain.text@email.com"])

        # Check that the emails were sent
        self.assertEqual(len(mail.outbox), 2)

        # check that the first email is the HTML email
        email_message = mail.outbox[0]
        self.assertEqual(email_message.subject, "Test HTML subject")
        self.assertEqual(email_message.alternatives, [('<h2>Test HTML content</h2>', 'text/html')])
        self.assertEqual(email_message.body, "TEXT content") # this is the intentional structure of an email message
        self.assertEqual(email_message.to, ["has.html@email.com"])

        # confirm that without html_message kwarg we do not get 'alternatives'
        email_message = mail.outbox[1]
        self.assertEqual(email_message.subject, "Test TEXT subject")
        self.assertEqual(email_message.alternatives, [])
        self.assertEqual(email_message.body, "TEXT content") # this is the intentional structure of an email message
        self.assertEqual(email_message.to, ["mr.plain.text@email.com"])
```

lb-:
@tr11  all good, tests have been added and merged in.
thanks for the contribution!

lb-:
Merged in https://github.com/wagtail/wagtail/commit/4f5b11725000706db9a8baccda5ad179f5ed01a9

PR #5055, opened 2019-02-03
Set data-focal-point-* values to be '' instead of 'None' when they ar…

### LINKED ISSUE #5007: Unclear error msg when saving img after focal point edit
### Issue Summary

A summary of the issue.
It is not possible to save an image after creating a focal point and clicking next the selected area. 

### Steps to Reproduce

1. Open cms
2. Open images
3. Edit img
4. remove any focal points (if any are present)
5. create focal point (drag mouse across img)
6. click outside the focal point but still within the img 
7. The selection (focal point) now seems to be gone
8. Save the IMG

You are now shown an error msg `The image could not be saved due to errors.`

In my opinion this is a minor issue but I would expect the img to be saved without having created a focal point (as the  focal point seems to be gone after clicking outside it) OR if it is not possible to save the img after the click outside the focal point I would expect a reason for this to be included in the error msg.

### Technical details

* Django version: Look in your requirements.txt, or run `pip show django | grep Version`.
* Wagtail version: Hover over the Wagtail bird in the admin, or run `pip show wagtail | grep Version:`.
* Browser version: You can use http://www.whatsmybrowser.org/ to find this out.

asankamal:
Fixes #5007 
The issue was data-focal-point-* attributes having "None" as the value when they are empty.

### DIFF
--- a/wagtail/images/templates/wagtailimages/images/edit.html
+++ b/wagtail/images/templates/wagtailimages/images/edit.html
@@ -59,10 +59,10 @@
 
                 <div class="focal-point-chooser"
                     style="max-width: {{ rendition.width }}px; max-height: {{ rendition.height }}px;"
-                    data-focal-point-x="{{ image.focal_point_x }}"
-                    data-focal-point-y="{{ image.focal_point_y }}"
-                    data-focal-point-width="{{ image.focal_point_width }}"
-                    data-focal-point-height="{{ image.focal_point_height }}">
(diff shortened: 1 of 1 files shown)


lb-:
Also validated on Firefox 65 on macOS and now merged via https://github.com/wagtail/wagtail/commit/74218a7e43f59b329cd7cd890245bcc2764beb3a

PR #5056, opened 2019-02-04
Add filter string character check to jinja2 image template tag

jonnyscholes:
Got tripped up by a `×` instead of a `x` when I copied an aspect ratio from Chromes devtools. This should help with that :)

I considered pulling both the django and jinja2 tags out into a base function, but that seemed like more work than necessary given the django template tag is mostly parsing the tag.

### DIFF
--- a/wagtail/images/jinja2tags.py
+++ b/wagtail/images/jinja2tags.py
@@ -1,12 +1,24 @@
+import re
+
+from django import template
 from jinja2.ext import Extension
 
 from .shortcuts import get_rendition_or_not_found
 from .templatetags.wagtailimages_tags import image_url
 
 
+allowed_filter_pattern = re.compile(r"^[A-Za-z0-9_\-\.\|]+$")
+
+
 def image(image, filterspec, **attrs):
     if not image:
         return ''
+
+    if not allowed_filter_pattern.match(filterspec):
+        raise template.TemplateSyntaxError(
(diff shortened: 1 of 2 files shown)


gasman:
Thanks @jonnyscholes!

Just one quibble that might be an issue here: the Django-template version of this tag allows multiple filters to be chained together, as seen here http://docs.wagtail.io/en/v2.4/topics/images.html#changing-per-tag . Internally, these get turned into a single string joined with `|` characters:

https://github.com/wagtail/wagtail/blob/a0d6dbb5f404dbf860ac26041ea1bc09a7a47a46/wagtail/images/templatetags/wagtailimages_tags.py#L74

The [docs for the jinja2 tag](http://docs.wagtail.io/en/v2.4/advanced_topics/jinja2.html#image) don't mention how to apply multiple filters, but from a quick glance at the code it looks like you need to use the `|`-delimited version within the tag. So, while this is currently undocumented, it's quite possible that people are already making use of this, in which case we should allow `|` characters as well.

(That does mean we can't re-use the regexp from the Django template tag, but I think the partial duplication is OK here...)

With this change in place, please could you add test cases to https://github.com/wagtail/wagtail/blob/master/wagtail/images/tests/test_jinja2.py to test that multiple filters work, and that an invalid filter string raises the `TemplateSyntaxError` as expected?

jonnyscholes:
@gasman Thanks! I've implemented your feedback :)

gasman:
Thanks @jonnyscholes! Merged in 4d004ce808f43ba596ad3d756e24f6a72eec056c.

PR #5057, opened 2019-02-04
Update third_party_tutorials.rst

pvetch:
This content seemed moribund so have added some newer material and checked that the original links still work. There are probably lots of other newer examples that could be added (it might be worth weeding out some of the older content at this point too).


gasman:
Thanks @pvetch - looks good to me! Merged in 731b8fc9b44a303ac256b6e6e478c37db61e1b82.

Issue #1240, opened 2015-04-26
Supporting RTL languages


bjesus:
Hi,
I've chosen Wagtail as the CMS for a big website our company runs. I'm in the process of integrating it into our current Django website, but I need the admin interface to be in all in Hebrew, and therefore, in RTL.

I've signed up on Transifex and requested (?) Hebrew so I can do the translation. Is this the right process?
Also, I guess the best way to RTL the admin interface is to have a separate CSS file that's loaded only for RTL languages. I'd be happy to write this CSS file, but I'm not sure how and where should I include it, and what would be best as way to contribute back to the project and not only to have a local hack. Any ideas?

Thanks.

tomdyson:
Hi @bjesus, it looks like you've successfully [started the translation](https://www.transifex.com/projects/p/wagtail/language/he_IL/), which is excellent - thanks very much.

@davecranwell, can you please reply on RTL support? I hadn't realised that a separate CSS file was necessary.

davecranwell:
Hi @bjesus. My main consideration is that if you were to submit a PR for RTL support it would be difficult for the Wagtail core team to support it in ongoing work since none of us are familiar with RTL languages.  (You could always submit ongoing PRs for RTL updates, but this process would be slow and quite frustrating for you!)

Perhaps the best way might be if you create a new wagtail app which you maintain yourself outside the Wagtail repository. You'd install the app like any other in `INSTALLED_APPS` and you'd create override templates for `skeleton.html` and/or `admin_base.html` (and any others you need) in which you add the `dir` attributes to the HTML at appropriate points, and add the RTL-specific CSS file too.

We've been meaning to create a list of third party Wagtail apps, and we'd be very happy to list yours if this way suited you.

tomdyson:
Hi @bjesus. To clarify, we would love to have proper RTL support in core. We *think* that promising this to our users requires an expert contributor (you?) who can commit to keeping RTL styles up to date with any significant change in the UI. But perhaps it's simpler than we're imagining. What do you think?

bjesus:
Hi,
I do tend to believe that this issue can be solved by CSS alone, without touching the admin's template files. That's the main reason I rather not create a new django app, but just have the file included if the language is an RTL one (Arabic, Hebrew, etc), but it's really up to you of course. Another option is maybe to give a special class (named "rtl" for example) to the <html> element and then simply add .rtl specific rules to the base SCSS files.

[...]

PR #5058, opened 2019-02-04
RTL styles for admin interface

remort:
See #1240.

Styles are tested and works fine on 2.2.2 version

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/scss/core.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/core.scss
@@ -455,3 +455,6 @@ footer,
         z-index: 200;
     }
 }
+html[dir=rtl] {
+    @import "rtl.scss";
+}
--- a/wagtail/admin/static_src/wagtailadmin/scss/rtl.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/rtl.scss
@@ -1,0 +1,356 @@
+@import 'wagtailadmin/scss/variables';
+
+.page-explorer .listing .index .privacy-indicator {
+  right: auto;
+  left: 5%;
+}
+
+.logo {
+    transform: scaleX(-1);
+}
+
+.explorer {
+	right: 180px;
+}
+
+header .left:first-child {
+  float: right;
+
+    .col {
+        float: right;
+        margin-left: 2em;
+        margin-right: 0;
+    }
+
+    h1.icon:before {
+        margin-left: 0.4em;
+        margin-right: 0;
+    }
+}
+
+header .right {
+    text-align: left;
+}
+
+.wrapper {
+    padding-left: 0;
+    padding-right: 180px;
+
+    .nav-wrapper {
+        float: right;
+        right: 0;
+        left: auto;
+    }
+
+    .content-wrapper {
+        float: right;
+    }
+}
+
+
+.nav-main {
+
+     a:before {
+        margin-right: 0;
+        margin-left: 0.5em;
+        float: right;
+    }
+
+    li.menu-explorer a {
+        border-right-color: #f37e77;
+    }
+
+    a {
+        border-right: 3px solid transparent;
+        border-left: none;
+    }
+
+    a:after {
+        right: auto;
+        left: 0.5em;
+    }
+
+    a.submenu-trigger:after {
+        content: "z";
+    }
+
+    .nav-submenu {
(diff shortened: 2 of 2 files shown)


remort:
@thibaudcolas I do not know how you test CSS. All the text (except LTR fragments) should be right-aligned and goes from right to left. All the buttons should be right-aligned and going in the reverse RTL way comparing to LTR groups of buttons. Icons prepend text, like on a screenshot I've added.

I could find all the css elements in the core SCSS styles and somehow add .rtl versions for them, but I have finished the project that need RTL wagtail support and have no opportunity to return to it. And I am not sure about how to reflect `<body dir='rtl'>` in individual styles. But it's a matter of technics, so you can translate separate file into core SCSS by yourself.

thibaudcolas:
@remort 👍 you can find a link to the automated test failures in my comment above, and further info on testing in the "Contributing to Wagtail" docs over at http://docs.wagtail.io/en/v2.4/contributing/index.html. When you created the PR, there should have also been a small template explaining what we expect from people making pull requests: linting, and cross-browser testing among others.

For the styles, here is an example – at the moment the separate RTL stylesheet you added defines:

```scss
.page-explorer .listing .index .privacy-indicator {
  right: auto;
  left: 5%;
}
```

If those styles were added directly to https://github.com/wagtail/wagtail/blob/487f1f47b8a581701694cba40a594955cee01de8/wagtail/admin/static_src/wagtailadmin/scss/components/_listing.scss#L445-L451


… it would end up looking like this:

```scss
.privacy-indicator {
    font-size: 1em;
    opacity: 1;
    position: absolute;
    right: 10%;
    top: 2em;

    html[dir=rtl] & {
       right: auto;
       left: 5%;
    }
}
```

This would make it much easier to understand what components need RTL styles, since the RTL styles are co-located with the rest.

remort:
@thibaudcolas, I tried to integrate rtl.scss to original styles and failed. There are some inner css logic whether nav or explorer open or not, so style declarations duplicates and depends on media screens additionally. But it seems to be doable for those who familiar with this inner concepts of styles structure. So it would be more suitable for scss-styles authors to adopt rtl.scss. I can help/test resulting styles if any questions appears.

thibaudcolas:
Thanks @remort! I'll keep an eye out for people who might be interested in picking this up and are more familiar with Wagtail's codebase. I'm sure we'll be able to use your help in testing this.

auvipy:
Hebrew, Arabic, Persian, Urdu languages will be benefited from this feature!! more than 500 Million people using those languages

RamezIssac:
Hello @thibaudcolas 

I have an opinion here: Why making RTL css code "next" to main css code ?
Usually it's indeed in a separate file. (Django admin as an example)

I do understand your concern about rtl css falling out of date, but that's a one thing that can be tested / taken into account when changing interface.

I think we can reach consensus here and move forward with the rtl css completion and integration.

thibaudcolas:
Hey @RamezIssac. The main reason to me is maintainability – the consensus with modern front-end development is to have all of a component’s styles together in the same file, rather than separate files for the desktop styles, mobile styles, print styles, RTL styles, mobile RTL styles, etc. This makes it much easier to keep track of what a component does, without having to sort through many files.

For example, an obvious benefit of co-locating RTL styles with the components they are for is that we could use variables that are local to the component to enforce spacing is always consistent between LTR and RTL versions.

If we had RTL styles as a separate stylesheet – I can already see people making new versions of a component by only modifying the component’s own file, and then the corresponding RTL styles (in a separate file) are likely not to be touched, and missed during the review. I also wouldn’t be surprised if this single "RTL" stylesheet grew to be much larger than the current 356 lines – I don’t see how it can be good for maintainability to have such long files to troubleshoot.

Ping @wagtail/frontend if anyone else has thoughts to share on this.

---

If anyone wants to work on this, I think a more reasonable approach would be to do this component-by-component rather than "the whole admin at once". This minimises the risk of PRs falling out of date.

thibaudcolas:
I’ll close this PR now because it’s very out of date, the current implementation needs a lot of rework, and it’s been opened on the wrong target branch. I think it’s also problematic how this split discussion that was already happening in #1240 – people who commented here are very welcome to discuss this further on the issue.

If anyone wants to contribute to this please read https://github.com/wagtail/wagtail/issues/1240#issuecomment-593326944, and discuss your plans on the issue.

PR #5061, opened 2019-02-06
Update primer.rst

DanAtShenTech:
Small wording clarifications.


gasman:
Thanks @DanAtShenTech! Merged in f1e3e2e37906b0c60d702185ae1160627e7844be.

For future reference, please could you open pull requests against the `master` branch rather than a stable branch? Stable branches generally only get updated for critical bugfixes (and in those cases, we'll take care of merging into both places as appropriate).

PR #5066, opened 2019-02-08
Don't make assumptions about the username field

nanorepublica:
The USERNAME_FIELD exists to allow customisation. Therefore we should make an assumption that `.username` exists on the model. Instead, we need to pull the required value from the USERNAME_FIELD and add in a fallback default.

### DIFF
--- a/wagtail/tests/utils/wagtail_tests.py
+++ b/wagtail/tests/utils/wagtail_tests.py
@@ -33,7 +33,8 @@ class WagtailTestUtils:
         user_model = get_user_model()
         # Login
         self.assertTrue(
-            self.client.login(password='password', **{user_model.USERNAME_FIELD: user.username})
+            self.client.login(password='password', 
+                              **{user_model.USERNAME_FIELD: getattr(user, user_model.USERNAME_FIELD, 'test@email.com')})
         )
 
         return user
--- a/wagtail/tests/utils/wagtail_tests.py
+++ b/wagtail/tests/utils/wagtail_tests.py
@@ -33,7 +33,7 @@ class WagtailTestUtils:
         user_model = get_user_model()
         # Login
         self.assertTrue(
-            self.client.login(password='password', 
+            self.client.login(password='password',
                               **{user_model.USERNAME_FIELD: getattr(user, user_model.USERNAME_FIELD, 'test@email.com')})
         )
 


gasman:
Thanks @nanorepublica! We have a custom user model with a non-default `USERNAME_FIELD` defined in our test fixtures, but evidently we're not exercising that here:

https://github.com/wagtail/wagtail/blob/56f29a0825acd10ad4d1746d4c90ed4732486ca7/wagtail/tests/customuser/models.py#L85

We should add this to our test coverage.

nanorepublica:
Hi @gasman, I now have some spare time to work on this now. What tests are required to get this merged?
In my opinion, to sufficiently cover all the scenario's it would require duplicating all existing tests where the `CustomUser` is used in place of the `EmailUser`. This, however, feels extreme and naive given I have just been a user of wagtail to date and not a contributor. 

Any guidance on what tests would be required would be helpful.

gasman:
Hi @nanorepublica,
As I recall I had a brief look at adding tests for this, but found that there were other non-obvious customisations needed - to the admin login form, possibly? - to get Wagtail working with a username-less User model to the point where it could be tested. Hopefully you have more familiarity with those customisations than I do!

It won't be necessary to duplicate the full set of user-related tests - I think it would be enough to have a single test that calls the test helper's `login` method and confirms that it results in a logged-in user (e.g. by fetching `/admin/` and getting back a 200 OK response).

If you're willing to add further tests to cover the other user-management views, then I definitely wouldn't say no to that - it sounds like we're lacking in test coverage for this sort of user model, and even if they do work in the current version, it's the sort of thing that could easily get broken in a future refactor if there aren't any tests to catch it. I wouldn't say that's a requirement for the current pull request, though.

nanorepublica:
This is a note to myself and others who are interested.

So adding a test for this will be difficult. I have started by overriding `AUTH_USER_MODEL` to the `customuser.EmailUser`, however, that leads to an error in the new test due to all foreign keys (initially the PageRevision model) are pointing to `customuser.CustomUser` when the test database is set up.

My current thinking is to have a second test database for the email custom user as Django doesn't have the concept of 2 user tables in a single database. So I am thanking either to use database routers or add functionality to the runtests.py script. The functionality required is to set `AUTH_USER_MODEL` to a specified model before the migrations run for a test run.

nanorepublica:
@gasman I have been pondering over this for a couple of days and I see a few options available.

1. Add in a second test run to the CI with the `customuser.EmailUser` Model, which would run against all existing tests. This is required since migrations that a run at the beginning of a test run hook everything up to the `customuser.CustomUser` model meaning that overriding `AUTH_USER_MODEL` for one test doesn't work. This might be better as a separate PR as I would have no idea what if anything else would break as a result. If this route was chosen, it would be great to understand or pointed of which CI I ought to be adding to (circle or travis)

2. Similar to the above but switch from `customuser.EmailUser` as the default, this has the benefit not adding another whole test run, but would catch assumptions made to date about the username field.

3. Look to mock the DB layer, so a simple test changing the EmailUser model works.

4. Not test it? (This is the worst obviously)

5. Something else I haven't thought of or come across.

I would appreciate your thoughts about which you would prefer. My preference is for #1

gasman:
@nanorepublica Thanks for looking into this. Yep, it does sound like this is a job for a separate test run. It would make sense for that to be part of the Travis config, as it's a similar situation to the Elasticsearch tests (although this one _shouldn't_ be in the allow-failures list - that's only the case for elasticsearch because those are intermittently flaky...) It's also fine for it to just run the tests that are most likely to touch the user model (presumably wagtail.admin, wagtail.core and wagtail.users).

Since the tests are a bigger job than I'd anticipated, I think we shouldn't let them block this immediate fix - so I'll go ahead and merge this now so that it can be included in Wagtail 2.6.

gasman:
Merged in b9816d194a47bcb96b9e5f1d07c6e99fb76c0b3e.

nanorepublica:
Great stuff. I may have a look at the test case since I have started, but I am not promising anything soon :)

PR #5068, opened 2019-02-09
Update primer.rst

vladox:
path to overriding template doesn't match model name


gasman:
Good catch - thanks @vladox! Merged in a0d6dbb5f404dbf860ac26041ea1bc09a7a47a46.

PR #5070, opened 2019-02-11
Clarify explanation of ParentalKey within InlinePanel

DanAtShenTech:
Make it clear that the name of the relationship to relate objects to a Wagtail Page object is called ParentalKey


gasman:
Thanks again @DanAtShenTech! Merged in 56f29a0825acd10ad4d1746d4c90ed4732486ca7.

PR #5076, opened 2019-02-12
Round to integer dimensions when scaling, address #5071

### LINKED ISSUE #5071: Scaling images rarely works.
### Issue Summary

Scaling of images doesn't work if the scaling factor produces a non-integral dimension.

### Steps to Reproduce

Render a template of the form
  `{% image my_image scale-50 as my_rendition %}`
for an image with odd height or width. Pillow will generate an exception from being passed non-integral dimensions. PR forthcoming. This was tested with a custom image class, but should occur regardless.

### Technical details

* Python version: 3.6.3
* Django version: 2.1.5
* Wagtail version: 2.5
* Browser version: Chromium 65

abrunyate:
Closes #5071 

As in the title. The tests have been modified to check the new behavior, as well as to generate an exception if resizing to fractional dimensions is attempted (forbidden per the Willow docs). Also added opportunity in the basic test case to check for invalid values in image operations, but this is not used.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8141,9 +8141,9 @@
       }
     },
     "natives": {
-      "version": "1.1.6",
-      "resolved": "https://registry.npmjs.org/natives/-/natives-1.1.6.tgz",
-      "integrity": "sha512-6+TDFewD4yxY14ptjKaS63GVdtKiES1pTPyxn9Jb0rBqPMZ7VcCiooEhPNsr+mqHtMGxa/5c/HhcC4uPEUw/nA==",
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/natives/-/natives-1.1.1.tgz",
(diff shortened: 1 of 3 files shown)


lb-:
@abrunyate  - I think this looks good, could you just update the PR to remove the changes to package-lock.json, that looks like it got in there accidentally.

Once that is updated one of us can do another quick check locally but I do not see any major issues.

@kaedroho  - I am happy to check & merge this one in once updated if needed.

abrunyate:
@lb Thanks for taking a look. I've removed the change to package-lock.json (it was needed on my old machine, for some reason).

lb-:
@abrunyate  - thank you for that, this has now been merged in via
https://github.com/wagtail/wagtail/commit/d3b8bcb68987cfddf2686f754932ba967745d3f7

We really appreciate you taking the time to submit this fix.

danihodovic:
When is this estimated to be released?

lb-:
@danihodovic  - this will be in the 2.6 release, 2.5 came out late April, so 2.6 should be out sometime late June

If this is a major blocker for you, you could use a non-released version, it could be potentially unstable but might be an option.

Here is a bit more info about the release process
http://docs.wagtail.io/en/v2.5.1/releases/upgrading.html

Issue #5073, opened 2019-02-12
AttributeError on Cloudflare cache invalidation


m1kola:
### Issue Summary

As far as I understand, the idea of the following piece of code is to log errors form the API and not interrupt the editing workflow:

https://github.com/wagtail/wagtail/blob/442ab73489d1665e3c451d8356c1c9429e92e5af/wagtail/contrib/frontend_cache/backends.py#L101-L104

The problem is that `requests.exceptions.HTTPError` doesn't seem to have the `message` attribute, so `AttributeError: 'HTTPError' object has no attribute 'message'` will be raised on the line 103. And as a result the request to render the admin UI will fail.

### Technical details

* Python version: 3.6.6
* `requests` version: 2.21.0
* `Django` version: 2.1.5
* `wagtail` version: 2.4

harrislapiroff:
I've come across this a bunch! Would be thrilled to see a resolution.

gasman:
Fixed in #5150

PR #5077, opened 2019-02-13
5073 attribute error

kevinhowbrook:
#5073

### DIFF
--- a/wagtail/contrib/frontend_cache/backends.py
+++ b/wagtail/contrib/frontend_cache/backends.py
@@ -100,7 +100,7 @@ class CloudflareBackend(BaseBackend):
 
         except requests.exceptions.HTTPError as e:
             for url in urls:
-                logger.error("Couldn't purge '%s' from Cloudflare. HTTPError: %d %s", url, e.response.status_code, str(e.args[0]))
+                logger.error("Couldn't purge '%s' from Cloudflare. HTTPError: %d %s", url, e.response.status_code)
             return
 
         if response_json['success'] is False:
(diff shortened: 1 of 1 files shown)


wagtail/contrib/frontend_cache/backends.py
```diff
@@ -100,7 +100,7 @@ def purge_batch(self, urls):
 
         except requests.exceptions.HTTPError as e:
             for url in urls:
-                logger.error("Couldn't purge '%s' from Cloudflare. HTTPError: %d %s", url, e.response.status_code, e.message)
+                logger.error("Couldn't purge '%s' from Cloudflare. HTTPError: %d %s", url, e.response.status_code, str(e.args[0]))
```
m1kola: Does it make sense to replace it with something like `logging.exception("Couldn't purge '%s' from Cloudflare. HTTPError: %d", url, e.response.status_code)`?

It should log the traceback from the exception as well.

kevinhowbrook: Yea, I just saw that the message is contained in `args[0]` so I thought I would add it in. But we can just log the `status_code` if that makes more sense

m1kola: @kevinhowbrook  My point was mostly about using `logging.exception` instead of `logger.error`. The message will be logged using the same logging level (error), but exception info will also added to the logging message.

kevinhowbrook: Ah my bad, I've switched that to your rec

gasman:
Superseded by #5150

PR #5083, opened 2019-02-14
Clarify how to customize modeladmin templates

DanielSwain:
Provide clarification and an example folder structure to help explain how to override modeladmin templates.

### DIFF
--- a/docs/reference/contrib/modeladmin/primer.rst
+++ b/docs/reference/contrib/modeladmin/primer.rst
@@ -142,15 +142,20 @@ Overriding templates
 --------------------
 
 For all modeladmin views, Wagtail looks for templates in the following folders
-within your project, before resorting to the defaults:
+within your project before resorting to the defaults:
 
 1. ``/modeladmin/app-name/model-name/``
 2. ``/modeladmin/app-name/``
 3. ``/modeladmin/``
 
-So, to override the template used by ``IndexView`` for example, you'd create a
-new ``index.html`` template and put it in one of those locations.  For example,
-if you wanted to do this for an ``ArticlePage`` model in a ``news`` app, you'd
+As when you `customise the Wagtail admin templates <https://docs.wagtail.io/en/v2.4/advanced_topics/customisation/admin_templates.html?highlight=wagtailadmin#customising-admin-templates>`_, you can create a ``/templates/`` folder within one of your installed apps in order to place the ``/modeladmin/`` folder.  If, for example, you create a ``dashboard`` app for the purpose of containing all logic related to customising the Wagtail admin, then you could set up the following overall folder structure:
+
+-   ``/dashboard/templates/wagtailadmin``
+-   ``/dashboard/templates/modeladmin``
+
+To override the template used by ``IndexView``,  you'd create a
+new ``index.html`` template and put it somewhere in the ``/modeladmin/`` folder structure.  For example,
+if you wanted to override ``index.html`` for an ``ArticlePage`` model in a ``news`` app, you'd
 add your custom template as ``modeladmin/news/articlepage/index.html``.
 
 For reference, ``modeladmin`` looks for templates with the following names for


docs/reference/contrib/modeladmin/primer.rst
```diff
@@ -142,15 +142,20 @@ Overriding templates
...
+within your project before resorting to the defaults:
 
 1. ``/modeladmin/app-name/model-name/``
 2. ``/modeladmin/app-name/``
 3. ``/modeladmin/``
 
-So, to override the template used by ``IndexView`` for example, you'd create a
-new ``index.html`` template and put it in one of those locations.  For example,
-if you wanted to do this for an ``ArticlePage`` model in a ``news`` app, you'd
+As when you `customise the Wagtail admin templates <https://docs.wagtail.io/en/v2.4/advanced_topics/customisation/admin_templates.html?highlight=wagtailadmin#customising-admin-templates>`_, you can create a ``/templates/`` folder within one of your installed apps in order to place the ``/modeladmin/`` folder.  If, for example, you create a ``dashboard`` app for the purpose of containing all logic related to customising the Wagtail admin, then you could set up the following overall folder structure:
```
lb-: Can this please be updated to use the RST ref system, otherwise this link will load 'old' versions of the docs on future versions using this link.

eg.
```rst
As when you :ref:`customise the Wagtail admin templates <advanced_customising_admin_templates>`, you can create a...
```

Then on the file `docs/advanced_topics/customisation/admin_templates.rst`
add the following *two* lines to the top of the file.

```rst
.. _advanced_customising_admin_templates:

```

Note that the 'ref' links to the file via the `.. _some_name:` syntax. but the ref removes the leading underscore, finally this will not work correctly if you skip the line break after the location declaration at the top of the admin_templates file.

docs/reference/contrib/modeladmin/primer.rst
```diff
@@ -142,15 +142,20 @@ Overriding templates
 --------------------
 
 For all modeladmin views, Wagtail looks for templates in the following folders
-within your project, before resorting to the defaults:
+within your project before resorting to the defaults:
 
 1. ``/modeladmin/app-name/model-name/``
 2. ``/modeladmin/app-name/``
 3. ``/modeladmin/``
```
lb-: The updates here are helpful but also potentially add some confusion.

the following locations will all work (using the bakerydemo project)
* `bakerydemo/templates/modeladmin/index.html`
* `bakerydemo/breads/templates/modeladmin/breads/index.html`

In the case above, the default modeladmin index will now be the one in the root templates folder, but any modeladmin within breads app will use the other one.

This change kind of implies you must always put the templates folder within a Django app, when you should also be able to use the root templates folder.

Maybe... this first list needs to be updated to reflect that templates should be in a `templates` folder, but not sure if that will make it even more confusing.

It would be good to try to explain the hierarchy a bit better with these docs changes.

docs/reference/contrib/modeladmin/primer.rst
```diff
@@ -142,15 +142,20 @@ Overriding templates
...
 
-So, to override the template used by ``IndexView`` for example, you'd create a
-new ``index.html`` template and put it in one of those locations.  For example,
-if you wanted to do this for an ``ArticlePage`` model in a ``news`` app, you'd
+As when you `customise the Wagtail admin templates <https://docs.wagtail.io/en/v2.4/advanced_topics/customisation/admin_templates.html?highlight=wagtailadmin#customising-admin-templates>`_, you can create a ``/templates/`` folder within one of your installed apps in order to place the ``/modeladmin/`` folder.  If, for example, you create a ``dashboard`` app for the purpose of containing all logic related to customising the Wagtail admin, then you could set up the following overall folder structure:
+
+-   ``/dashboard/templates/wagtailadmin``
```
lb-: I recommend removing this wagtailadmin reference, while it appears to be used as an analogy of the folder structure it seems to imply that `modeladmin` templates can also sit in `wagtailadmin`, which will not work (as far as I know :) )

docs/reference/contrib/modeladmin/primer.rst
ababic: > Maybe... this first list needs to be updated to reflect that templates should be in a templates folder, but not sure if that will make it even more confusing

I think this would make a lot of sense!

docs/reference/contrib/modeladmin/primer.rst
ababic: I would also add that we might be making too big a jump here conceptually. I feel that it might be better to restore the previous content, but instead of providing just one example, show all three of the previous locations (prefixed with `templates/`) with `index.html` at the end.

Then, we can tackle the issue of "where in a project to create template folders" separately. And whatever we do here, we should provide a caveat that Django's customisable nature means that templates can technically live anywhere in a project, but common approaches include:

1. Adding a `templates/` folder to the relevant app.
2. Adding them to the project's root `templates/` folder.
3. Creating an app for the explicit purpose of overriding default Django/Wagtail behaviour, and adding override templates to a `templates/` folder in that app.

And provide a simple directory structure example for each of these, so as not to show bias for one approach or the other.

ababic: Thinking on this some more , I think we CAN safely recommend that app-specific templates are better off being placed in app-specific folders, because it improves app portability, and would make the templates easier to find, should things need tweaking in future. 

Options 2 and 3 are only really a better options for project-wide overrides, which I would imagine aren't as common.

docs/reference/contrib/modeladmin/primer.rst
ababic: I agree, but more because overriding some base templates for the entire project is conceptually very different - The `dashboard` app approach makes sense there. 9 times out of 10, I'd guess that most `modeladmin` template overrides are app or model specific, in which case, they are better off being added to app-specific template folders.

DanAtShenTech: Thank you both for your comments.  In proposing this change, I was trying to write in the same manner that the existing Wagtail documentation is written, i.e. Wagtail assumes a certain level of Django knowledge and so doesn't try to re-explain things that are found in the Django docs.  What prompted me to propose this change was that I spent longer than I should have, after following the Wagtail docs, in trying to set up a `modeladmin` customization - the docs said:

> For all `modeladmin` views, Wagtail looks for templates in the following folders within your project, ...

Not thinking (but following the docs), I just placed a `modeladmin` folder in the project, and of course Wagtail couldn't find it.  It was left unsaid in the Wagtail docs that this folder needed to be within a recognized templates folder.  In my proposed change I stated 

> As when you customise the Wagtail admin templates, you can...

and linked to the customisation of the admin templates because I thought it provided a useful cross-reference.  Perhaps, then, just a note about `modeladmin` needing to be in a recognized templates folder would be sufficient, and I definitely agree with the point that `modeladmin` customisation would more likely be done within individual apps than the way I've shown it.

Two questions then:
- If I simply removed the following from this proposed change, and let the rest, would this suffice:
> If, for example, you create a dashboard app for the purpose of containing all logic related to customizing the Wagtail admin, then you could set up the following overall folder structure:
    /dashboard/templates/wagtailadmin
    /dashboard/templates/modeladmin
- Would you be OK with letting the link to customizing the Wagtail admin in as a useful cross-reference, or would you rather see that removed?

ababic: Thanks @DanAtShenTech.

> In proposing this change, I was trying to write in the same manner that the existing Wagtail documentation is written, i.e. Wagtail assumes a certain level of Django knowledge and so doesn't try to re-explain things that are found in the Django docs

This is a good point, and worth bearing in mind here.

> If I simply removed the following from this proposed change, and let the rest, would this suffice?

If you could try just tweaking the existing content first, to better show that templates should live in `templates/` folders? I think that might possibly be enough to clear up the confusion here.

> Would you be OK with letting the link to customizing the Wagtail admin in as a useful cross-reference, or would you rather see that removed?

I would say let's keep things simple for now and leave the link out. If we can rejig the existing `ArticlePage` override example to show override templates being added to the `news` app folder, I think we'll be about there. I think most of the information on that page is very `wagtailadmin` / branding specific, so isn't actually that relevant (even if it the subject matter is similar).

lb-: no further comments from me, all makes sense. Thanks @DanAtShenTech  for wanting to make this better, and yeah when I first read the docs the other day I also assumed the templates would go in a 'modelAdmin' folder not a 'templates/modelAdmin' one so I think it is very fair that others will get confused by this and it needs improvements.

lb-:
After lots of discussion here, I think this can now be resolved, the main changes about the template path has been implemented via 4ab7080fb8e3def96ec6e82df20c9b38e1a93f2e & #6605

Looking at the threads above, it appears that the end decision was to leave out the link to the internal wagtail template docs.

If we want to improve this further we can rase a new PR.
Thanks @DanielSwain  -> we appreciate your effort here to get this moving a while ago.

PR #5091, opened 2019-02-19
Output current language code in HTML header

### LINKED ISSUE #5051: LANGUAGE_CODE doesn't affect admin interface html 'lang' attribute in skeleton template
### Issue Summary
`<html>` tag attribute 'lang' in skeleton.html admin template is not being filled with value of LANGUAGE_CODE settings variable
But it supposed to: `<html class="no-js" lang="{{ LANGUAGE_CODE|default:"en-gb" }}">`

### Steps to Reproduce

Set LANGUAGE_CODE in settings.py to, i.e., 'he_IL'.
Restart your application/Reload container and refresh admin page.
Get page source in browser. You'll get `<html lang="en-US">` at the top of the page source.
It seems that LANGUAGE_CODE variable is not accessible while in skeleton.html.

### Technical details
wagtail 2.2.2
python 3.6

gasman:
Fixes #5051. The `lang` attribute of the `<html>` tag in the admin base template was meant to vary according to the variable `LANGUAGE_CODE`, but this variable was never populated. Here we use the `get_current_language` code as per https://docs.djangoproject.com/en/2.1/topics/i18n/translation/#switching-language-in-templates

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/skeleton.html
+++ b/wagtail/admin/templates/wagtailadmin/skeleton.html
@@ -1,5 +1,6 @@
 <!doctype html>
 {% load static i18n %}
+{% get_current_language as LANGUAGE_CODE %}
 <html class="no-js" lang="{{ LANGUAGE_CODE|default:"en-gb" }}">
 <head>
     <meta charset="utf-8" />
--- a/wagtail/admin/tests/test_account_management.py
+++ b/wagtail/admin/tests/test_account_management.py
@@ -363,6 +363,9 @@ class TestAccountSection(TestCase, WagtailTestUtils):
         # Page should contain a 'Language Preferences' title
         self.assertContains(response, "Language Preferences")
 
+        # check that current language preference is indicated in HTML header
+        self.assertContains(response, '<html class="no-js" lang="en">')
+
     def test_language_preferences_view_post(self):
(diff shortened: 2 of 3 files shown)


wagtail/admin/templates/wagtailadmin/skeleton.html
```diff
@@ -1,5 +1,6 @@
 <!doctype html>
 {% load static i18n %}
+{% get_current_language as LANGUAGE_CODE %}
 <html class="no-js" lang="{{ LANGUAGE_CODE|default:"en-gb" }}">
```
ababic: I'm thinking it might be best to ditch the `|default:"en-gb"` from this line. Whether or not a current language has been specified, I think `get_current_language()` would always return the value of `settings.LANGUAGE_CODE` - and I imagine it's very rare for that not to have a value.

gasman: Good point! Looks like Django will supply a default language code of `en-us` if there isn't one specified in settings, so the default here would never get picked up even if we wanted it to...

gasman:
Merged (with `
default:"en-gb"` removed) in 8b7efc0b160ad0dcf2ca455b54a90a9bcde25f6e.

PR #5100, opened 2019-02-23
Add form_classname to ListBlock & StreamBlock

### LINKED ISSUE #4042: Support form_classname attribute on StreamBlock and ListBlock

lb-:
Full implementation of this feature, allowing `form_classname` to be set on both `ListBlock` and `StreamBlock`.

Replaces #4047
Resolves #4042

---

### DIFF
--- a/docs/topics/streamfield.rst
+++ b/docs/topics/streamfield.rst
@@ -444,6 +444,28 @@ Any block type is valid as the sub-block type, including structural types:
         ('amount', blocks.CharBlock(required=False)),
     ])))
 
+To customise the class name of a ``ListBlock`` as it appears in the page editor, you can specify a ``form_classname`` attribute as a keyword argument to the ``ListBlock`` constructor:
+
+.. code-block:: python
+    :emphasize-lines: 4
+
+    ('ingredients_list', blocks.ListBlock(blocks.StructBlock([
+        ('ingredient', blocks.CharBlock()),
+        ('amount', blocks.CharBlock(required=False)),
+    ]), form_classname='ingredients-list'))
+
+Alternatively, you can add ``form_classname`` in a subclass's ``Meta``:
+
+.. code-block:: python
+    :emphasize-lines: 6
+
+    class IngredientsListBlock(blocks.ListBlock):
+        ingredient = blocks.CharBlock()
+        amount = blocks.CharBlock(required=False)
+
+        class Meta:
+            form_classname = 'ingredients-list'
+    
 
 StreamBlock
 ~~~~~~~~~~~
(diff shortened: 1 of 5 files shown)


lb-:
@mnieber - just wanted to get your thoughts on this new PR for adding `form_classname`.

Note: I am not sure how #4942 will affect this feature once merged, it appears to remove usage of form_classname but for now I think this is still something that would be useful.

wagtail/core/blocks/list_block.py
```diff
@@ -88,6 +88,7 @@ def render_form(self, value, prefix='', errors=None):
             'help_text': getattr(self.meta, 'help_text', None),
             'prefix': prefix,
             'list_members_html': list_members_html,
+            'classname': getattr(self.meta, 'form_classname', None),
```
mnieber: Rather than renaming the attribute to something more generic (classname) I have the gut feeling that it should be renamed to something equally specific (i.e. keep the name `form_classname`) or more specific. The reason is that within the context of the template, there could be many html elements that have a customizable `classname`, so the generic name `classname` could be confusing.

lb-: I agree but that will go against the convention already set in structBlock by renaming form_classname to classnames when rendering the form. I think it should stay as is.

mnieber: Yes, that makes sense.

lb-: thanks, yeah I want other devs to be able to override this the same way they would the structBlock, including how the template variables are made available. I'll see who else from the team can review this one now.

lb-:
@loicteixeira  or @gasman  - could I please get a review on this one.

Also, not sure on the timing of the new streamfield merge on #4942 - this kind of usage may no longer be relevant after that is merged in.

Thanks

lb-:
As #4942 will be on hold for a bit, I have rebased, fixed up issues due to the new redesign and updated this PR's description.

docs/topics/streamfield.rst
```diff
@@ -444,6 +444,26 @@ Any block type is valid as the sub-block type, including structural types:
...
+
+Alternatively, you can add ``form_classname`` in a subclass's ``Meta``:
+
+.. code-block:: python
+
+    ('ingredients_list', blocks.ListBlock(blocks.StructBlock([
+        ('ingredient', blocks.CharBlock()),
+        ('amount', blocks.CharBlock(required=False)),
+    ]), form_classname='ingredients-list'))
+
```
gasman, line 466: Looks like these two examples are swapped around from what the descriptions say (and likewise for StreamBlock).

lb-: Thanks, Updated + reworked the `StreamBlock` docs part to align better with how that section is laid out.

lb-:
@gasman  - docs fixed up, merged in via 8fc722a6f3e4d25c09f99937d387ba521aa713dc

PR #5102, opened 2019-02-24
routablepageurl returns full url for first site on multisite insta…

gassan:
…llation instead of relative url if another site with the same root page exist


gasman:
Thanks @gassan! Merged in d37ce9f699223d92c88fad93aad19a5e7f9ec998 (with some more expanded tests - we want test code to be as simple and linear as possible, so I've removed the inheritance).

PR #5106, opened 2019-02-26
Fix: Update modeladmin to preserve compatibility with RelatedFieldListFilter in Django 2.2+

### LINKED ISSUE #5105: Update modeladmin to preserve compatibility with RelatedFieldListFilter / FieldListFilter in Django 2.2+
### Issue Summary

In 2.2, Django has updated it's `RelatedFieldListFilter` and `FieldListFilter` classes to be coupled to a `ModelAdmin.admin_site` property (See https://github.com/django/django/commit/6d4e5feb79f7eabe8a0c7c4b87f25b1a7f87ca0b).

`wagtail.contrib.modeladmin.view.IndexView` piggybacks on these classes to support filters added via `ModelAdmin.list_filter`. But, unlike their Django counterparts, instances of `wagtail.contrib.modeladmin.options.ModelAdmin` have no `admin_site` attribute, so an `AttributeError` is raised when certain filter types are used.

### Proposed fix

After considering the various options, it is looking like the best short-term solution is to import Django's default `admin_site` (`django.contrib.admin.site`) and make that available as `admin_site` on all instances of `ModelAdmin`.

Longer term, it might be worth defining a custom `django.contrib.admin.sites.AdminSite` class within Wagtail, and setting `admin_site` to an instance of that instead. However, reconciling the differences between how the Wagtail and Django admin areas work may prove to be more work than is worth it.

ababic:
Resolves #5105

* Do the tests still pass? **Yes**

Tests in `wagtail.contrib.modeladmin.tests.test_simple_modeladmin.TestBookIndexView` should no longer error when running against Django's `master` branch

### DIFF
--- a/wagtail/contrib/modeladmin/options.py
+++ b/wagtail/contrib/modeladmin/options.py
@@ -123,7 +123,7 @@ class ModelAdmin(WagtailRegisterable):
 
         # Needed to support RelatedFieldListFilter in Django 2.2+
         # See: https://github.com/wagtail/wagtail/issues/5105
-        self.admin_site = default_django_admin_site  
+        self.admin_site = default_django_admin_site
 
     def get_permission_helper_class(self):
         """
--- a/wagtail/contrib/modeladmin/options.py
+++ b/wagtail/contrib/modeladmin/options.py
@@ -1,4 +1,5 @@
 from django.conf.urls import url
+from django.contrib.admin import site as default_django_admin_site
 from django.contrib.auth.models import Permission
 from django.core import checks
 from django.core.exceptions import ImproperlyConfigured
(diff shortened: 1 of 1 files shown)


ababic:
@gasman Have successfully rebased onto `master` so we can see the affect on tests

This is the first time i've ever rebased without messing it up, so 😀

ababic:
@auvipy This is essentially already tested, in the sense that if you look at previous builds against Django 2.2 (e.g. https://travis-ci.org/wagtail/wagtail/jobs/500952790), there are a 9 test errors relating to this issue.

If you look at the same tests for this environment following these changes (https://travis-ci.org/wagtail/wagtail/jobs/499394096), you'll see that the `AttributeError: 'BookModelAdmin' object has no attribute 'admin_site'` errors are now gone.

gasman:
Thanks @ababic!

PR #5107, opened 2019-02-26
Update tox config to test against Django's stable/2.2 branch

ababic:
Django's master branch now contains changes intended for the `3.0` release, so I've updated tox to tests against Django's `stable/2.2` branch, so that we have a better indication about how far away we are from supporting that version specifically.


gasman:
Looks good to me - thanks @ababic!

PR #5108, opened 2019-02-26
Add ability for page.copy to copy ParentalManyToMany field relations

### LINKED ISSUE #4931: Page.copy() should copy Parental Many to Many fields
### Issue Summary

Looking at the relevant code in `Page.copy`, unfortunately it doesn't look like we're doing anything to deal with those fields when copying revisions. `exclude_fields_in_copy` does not work on Parental M2M relations also, due to this limitation.

### Steps to Reproduce

1. Open the wagtail bakery demo (admin) and edit the first 'Bread' page by adding some ingredients (as it is a parental many to many field: `ingredients = ParentalManyToManyField('BreadIngredient', blank=True)`.
2. Save changes to that page with ingredients added
3. Go back to page listing under 'breads' and copy the page
4. Open the copied page (in the page editor)
5. Creating a copy of that instance using the Page.copy() does not offer a way to not copying those relations in the revision file (ingredients field), none are selected. 

We have discussed this issue in the support slack channel with @gasman and found this to be a possible bug.

*updated*

lb-:
closes #4931 

* Do the tests still pass? ✔️ 
* Does the code comply with the style guide?  ✔️ 
* For Python changes: Tests added  ✔️

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -22,7 +22,8 @@ from django.utils import timezone
 from django.utils.functional import cached_property
 from django.utils.text import capfirst, slugify
 from django.utils.translation import ugettext_lazy as _
-from modelcluster.models import ClusterableModel, get_all_child_relations
+from modelcluster.models import (
+    ClusterableModel, get_all_child_m2m_relations, get_all_child_relations)
 from treebeard.mp_tree import MP_Node
 
 from wagtail.core.query import PageQuerySet, TreeQuerySet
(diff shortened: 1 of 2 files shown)


lb-:
@kaedroho  or  @gasman  - would love to get a review of this if possible, not sure if there is a nicer way to copy the ParentalManyToManyField relation.

lb-:
Merged in https://github.com/wagtail/wagtail/commit/fa2a0c2d02d41610d469597575e3bf644ba5fbef

PR #5109, opened 2019-02-27
Update docs for construct_homepage_panels hook

nwjlyons:
`construct_homepage_panels` hook doesn't need a `return` statement.

[Looking at the code](https://github.com/wagtail/wagtail/blob/master/wagtail/admin/views/home.py#L101-L102), the user doesn't need to return a list, they need to mutate the list passed in.


gasman:
Thanks @nwjlyons - good spot! Merged in f10c5cab58ceb9935403a3cff32a36579c8a68f8.

Issue #161, opened 2014-03-25
Allow Page Title field to be dynamically generated or relabelled


chrxr:
In some situations, Page Title is not an appropriate label for the main heading of a page.

For example, with a student or staff profile, it is the name of the person that is used for the page title. 

In addition to this top-level title there is a necessity to have separate first and surname fields. So in this situation we are duplicating information.

The co-existence of the three of these fields caused much confusion with students filling in profiles for the 2014 CCA Show.

A proposed solution would be for the Title field to able to be dynamically generated from other fields on a model-by-model basis. An example of this would be on a student's profile the page title could be generated by appending the first names to the surname field.

A less desirable solution, but acceptable, would be enabling the ability to relabel the page title field on a per-model basis. So the page title field on the student's profile could be renamed 'Full name' or similar.

gasman:
#2278 makes this possible, currently with a few rough edges...

    class PersonPage(Page):
        first_name = models.CharField(max_length=255)
        last_name = models.CharField(max_length=255)
        intro = RichTextField(blank=True)
        biography = RichTextField(blank=True)

        def clean(self):
            super(PersonPage, self).clean()
            self.title = "%s %s" % (self.first_name, self.last_name)

    PersonPage.content_panels = [
        FieldPanel('first_name'),
        FieldPanel('last_name'),
        FieldPanel('intro', classname="full"),
        FieldPanel('biography', classname="full"),
    ]

Note that `content_panels` does not include the title field - instead, that gets populated from the `clean` method, which is called before save and preview. The rough edge is that the slug doesn't get auto-populated, and has to be filled in manually.

justinoue:
Is there any resolution for this? It's been 2+ years...

Currently trying to programatically set titles and slugs but even setting slug in clean or or full_clean is causing a slug validation error.

Having a key field be auto populated by a javascript dependency on another field that might not be there seems like amateur hour.

justinoue:
I've managed to work around this with the following:

`
$(document).ready(function(){
    if($('#id_slug').val() == ''){
        $('#id_slug').val('placeholder');
    }
});
`

and in my model:

`
    def clean(self):
        super(CustomPage, self).clean()
        self.title = "{} {}".format(self.field1, self.field2)
        self.slug = text.slugify(self.title)
`

urlsangel:
Hi all

I'm coming up against exactly this issue now, and definitely don't want to duplicate title, first name and last name data.

[...]

PR #5112, opened 2019-03-01
Refs #161: Use cleaned page instance on preview

sir-sigurd:
We have setup like this:

```python
class ProfilePageForm(WagtailAdminPageForm):
        super().__init__(*args, **kwargs)
        self.fields['slug'].required = False


class ProfilePage(Page):
    first_name = models.CharField(max_length=30)
    last_name = models.CharField(max_length=30)

    content_panels = [
        FieldPanel('first_name'),
        FieldPanel('last_name'),
    ]
    base_form_class = AbstractProfilePageForm

    def full_clean(self, *args, **kwargs):
        self.title = f'{self.first_name} {self.last_name}'
        super().full_clean()
```

It works for us, but non-saved page preview crashes, because `url_path` is not set. Tests are TBD.

### DIFF
--- a/wagtail/admin/views/pages.py
+++ b/wagtail/admin/views/pages.py
@@ -636,7 +636,7 @@ class PreviewOnEdit(View):
         if not form.is_valid():
             return self.error_response(page)
 
-        form.save(commit=False)
+        page = form.save(commit=False)
         preview_mode = request.GET.get('mode', page.default_preview_mode)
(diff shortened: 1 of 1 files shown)


sir-sigurd:
It's invalid: previously `title` was generated in `clean()`, moving it to `full_clean()` was sufficient to make it work.

Issue #1985, opened 2015-12-02
UrlField.max_length is too short


balazs-endresz:
Wagtail is using the `UrlField` provided by Django, which has `max_length=200`. Sometimes this is not enough, especially for certain embeds.

Could we use a value larger than that for maybe all `UrlField`s in Wagtail? Or what would be the best way to override this on a specific website?

Related to this, someone managed to submit a page with a long url without any validation errors, which then failed when publishing it. But the draft did get saved (along with the long url) but the page editor form didn't render at all (because of an error when rendering the url field and `TEMPLATE_DEBUG=False`). Changing the embed url field length directly in the database fixed this. Although this was an older version of Wagtail, so I'll give it a try on 1.2 with richtext embeds and streamfield embed blocks too.

davecranwell:
If possible I don't think there should be any limit. It wouldn't be unreasonable for implementers to use a UrlField in a model expecting a google maps link. An event, say, where the editor should provide a link to the venue location map. You'd want it to be validated as a URL, but google maps links are in the hundreds of characters, potentially over 255 even.

balazs-endresz:
I thought mysql can handle max_length > 255 but apparently not if it has a unique constraint:

> Any fields that are stored with VARCHAR column types have their max_length restricted to 255 characters if you are using unique=True for the field.

https://docs.djangoproject.com/en/1.9/ref/databases/#character-fields

I guess that makes subclassing `TextField` more appealing.

ghostwords:
Note that this is likely to go unnoticed in development since the default SQLite database backend ignores VARCHAR length limits:

> [(9) What is the maximum size of a VARCHAR in SQLite?](http://www.sqlite.org/faq.html#q9)
> 
> SQLite does not enforce the length of a VARCHAR. You can declare a VARCHAR(10) and SQLite will be happy to store a 500-million character string there. And it will keep all 500-million characters intact. Your content is never truncated. SQLite understands the column type of "VARCHAR(N)" to be the same as "TEXT", regardless of the value of N.

thibaudcolas:
See also #4292:

> Embedding a URL longer than 200 chars results in: `DataError: value too long for type character varying(200)`.

### Steps to Reproduce

1. In a Wagtail project using `StreamField`
2. Add `('embed', EmbedBlock()),` to it
3. Create/edit a page and add an Embed block
[...]

PR #5114, opened 2019-03-01
Djangos URLFeld is limited to 200 characters, and some thumbnail valu…

kevinhowbrook:
…es are larger. This is causing an issue with embeds.


ababic:
Hi @kevinhowbrook. Django's URLField has a default `max_length` of `200`, but it isn't a hard limit. Like many of Django's built-in field types, you can specify an alternative `max_length` value as required. I think that might be a better approach to take here, rather than changing the underlying field type.

How long are the URLs you are seeing?

thibaudcolas:
Relevant: #1985.

ababic:
Thanks @thibaudcolas. I've posted some thoughts there, so that the conversation is in one place.

kevinhowbrook:
> Hi @kevinhowbrook. Django's URLField has a default `max_length` of `200`, but it isn't a hard limit. Like many of Django's built-in field types, you can specify an alternative `max_length` value as required. I think that might be a better approach to take here, rather than changing the underlying field type.
> 
> How long are the URLs you are seeing?

Largest I've seen is 230, I agree that just setting a higher limit on this will be enough. 
Having read https://github.com/wagtail/wagtail/issues/1985, I agree with:

> The simplest fix here is to set max_length=255 in the url field definition to increase the limit a little.

Because a) I haven't seen anything yet > 230 and b) MySQL knock-ons

It would be a good idea to provide better error handling here because without `debug=True` I think a publisher would just see a generic error screen. Not sure how much is going to be involved in that though, I'll investigate. For now, I'll suggest a limit on the field.

ababic:
Thanks for this @kevinhowbrook. Just to be on the safe side (and to help guard against potential regressions), would you mind adding a very basic test to show that an `Embed` object can be created with a long `url` and successfully retrieved again using `Embed.objects.get()`? This should help spot any potential database-related issues.

gasman:
This fix does feel a bit like kicking the can down the road - if 200-255-character URLs are an issue now, it's probably only a matter of time before >255-character URLs become an issue too. Since we're unlikely to be doing any lookups or validation on `thumbnail_url`, it's probably reasonable enough to change it to a `TextField`.

Having said that, there's no reason not to accept this as an immediate solution, so I'll go ahead and merge. (I don't think we need an extra test here - while I'm usually an avid proponent of testing every functional change, a test for "can a `URLField(max_length=255)` store >200 characters" is self-evident to the point of being redundant, I think :-) )

gasman:
Merged in 124827911463f0cb959edbb9d8d5685578540bd3

ababic:
Thanks @gasman.

> while I'm usually an avid proponent of testing every functional change, a test for "can a URLField(max_length=255) store >200 characters" is self-evident to the point of being redundant, I think :-) 

The recommended test addition was more to make sure that the indexes are created okay in `mysql` (because of the column being used in a `unique_together`). I'm not so bothered about the column length itself, per-se.

gasman:
@ababic We're only changing `thumbnail_url` here, which isn't part of the `unique_together` constraint. But yep, if we we're considering changing `url` too, that would be a fair point!

ababic:
@gasman ah, of course! apologies for the confusion.

PR #5115, opened 2019-03-04
Fix: still possible to create copies of pages that exceed their models' `max_count`

### LINKED ISSUE #5111: still possible to create copies of pages that exceed their models' `max_count`
### Issue Summary

In Wagtail admin, it's possible to select a page type that has hit its `max_count` limit (introduced in Wagtail 2.4) and then use the "copy" action in the dropdown menu to create another copy, even though no new pages of the type should be allowed to be created.

### Steps to Reproduce

1. Create a new model derived from wagtail.core.models.Page
2. Add a `max_count = 1` property
3. In Wagtail Admin, create a page of this type and publish it
4. In Admin, select the page's **More** > **Copy** action, choose a new slug for it
5. Publish the copy - a new page is created, no error is thrown
6. **Expected behavior**: because the existing page meets the `max_count` limit, an error should be thrown alerting the user that further copies cannot be created

### Technical details

* Python version: Python 3.7.2
* Django version: 2.1.5
* Wagtail version: 2.4
* Browser version: Chrome 72

There's no check against `max_count` in the `Page` model's `copy` method, only in `can_create_at`.

ababic:
Fixes #5111.

Updated `PagePermissionTester.can_copy_to()` to use the page class's `can_create_at()` method to check viability of the action, before checking the user's permissions

Also added some tests for this method, as it didn't seem to be covered by the current tests in `test_page_pernissions.py`

* Do the tests still pass? **Yes**
* For Python changes: Have you added tests to cover the new/fixed behaviour? **Yes**
* For front-end changes: Did you test on all of Wagtail’s supported browsers? **N/A**

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -1841,9 +1841,15 @@ class PagePermissionTester:
         if recursive and (self.page == destination or destination.is_descendant_of(self.page)):
             return False
 
-        # shortcut the trivial 'everything' / 'nothing' permissions
+        # reject inactive users early
         if not self.user.is_active:
             return False
+
+        # reject early if pages of this type cannot be created at the destination
+        if not self.page.specific_class.can_create_at(destination):
+            return False
+
+        # skip permission checking for super users
         if self.user.is_superuser:
             return True
 
--- a/wagtail/core/tests/test_page_permissions.py
+++ b/wagtail/core/tests/test_page_permissions.py
@@ -475,7 +475,11 @@ class TestPagePermissionTesterCanCopyTo(TestCase):
         # These same pages will be used for testing the result for each user
         self.board_meetings_page = BusinessSubIndex.objects.get(url_path='/home/events/businessy-events/board-meetings/')
         self.event_page = EventPage.objects.get(url_path='/home/events/christmas/')
-        self.singleton_page = SingletonPageViaMaxCount.objects.get(url_path='/home/only-one/')
+
+        # We'll also create a SingletonPageViaMaxCount to use
+        homepage = Page.objects.get(url_path='/home/')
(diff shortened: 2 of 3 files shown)


ababic:
@lb- thank you for taking the time to review this, and for the detailed feedback. 

> One suggestion would be to ensure that the page listing does not even show the actual button in the 'more drop-down.'

While I agree this would be a nice touch, I'm slightly concerned about efficiency. To work things like this out for many objects at a time could get very expensive indeed... We'd really need to think about introducing separate machinery to do that efficiently, and I imagine that could get incredibly complicated.

> First split the method can_create_at into two methods so they can be called in isolation of each other.

From a code perspective, this makes complete sense. But, in a way, I kinda like how `can_create_at()` makes you consider the parent page, as it subtly provides extra clues about how/when the method is used, and that makes it feel less 'risky' to override. For example, if you have to make one or two additional queries to work something out, it won't adversely affect performance.

What do you think?

lb-:
@ababic  - good call, I did not think about the implications of querying the total pages for that modal for *each* page listed.

The 403 page users get when they do something incorrect on this page (copy page) is a bit brutal, but that is not related to this specific issue and it is possible for devs to override this + could be improved in the future if needed.

Anyway, this is done, I will merge it in.

ababic:
thanks @lb-! yes, we should definitely create an issue to improve how the view handles cases where the page cannot be copied anywhere. This behaviour made sense when it was just about where the administrator had permission to copy the page, but not any longer.

lb-:
Merged in https://github.com/wagtail/wagtail/commit/77a8e3b7c08dafa5e18a57ce5e1d8f6ac7d54e1a

PR #5116, opened 2019-03-04
Add a setting to apply limit on number of tags that can be added to any tagged model.

contactr2m:
Hello,

I was looking for feature where i can limit the number of tags that can be applied to Tagged model.
Wagtail uses django-taggit and it is tightly coupled with it as of now.

I did not find any better way of achieving this without modifying wagtail itself.
After digging through the code i found out that wagtail uses tag-it plugin which has this setting in built.
So why not use that :)

This PR basically adds setting called TAG_LIMIT which gets applied to tag-it jQuery plugin through tag widget.

Once setting is in place, user wont be able to add tags more than no specified by TAG_LIMIT setting.

* Do the tests still pass? **yes**
* Does the code comply with the style guide? **yes**
* For Python changes: Have you added tests to cover the new/fixed behaviour?**No existing placeholder found.**
**latest**
**Safari desktop**
**Chrome desktop**
**Firefox desktop**
**Chrome ios**
**Safari ios**
* For new features: Has the documentation been updated accordingly?**yes**

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/widgets/tag_widget.html
+++ b/wagtail/admin/templates/wagtailadmin/widgets/tag_widget.html
@@ -1,1 +1,9 @@
-{% include 'django/forms/widgets/text.html' %}<script>initTagField('{{ widget.attrs.id|escapejs }}', '{{ widget.autocomplete_url|escapejs }}', {% if widget.tag_spaces_allowed %}true{% else %}false{% endif %});</script>
+{% include 'django/forms/widgets/text.html' %}
+<script>
+    initTagField(
+        '{{ widget.attrs.id|escapejs }}',
+        '{{ widget.autocomplete_url|escapejs }}',
+        {%if widget.tag_spaces_allowed %}true{% else %}false {% endif %},
+        {% if widget.tag_limit %}{{widget.tag_limit}}{%else %}null{% endif %}
+    );
+</script>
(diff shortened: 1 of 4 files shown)


wagtail/admin/templates/wagtailadmin/widgets/tag_widget.html
```diff
@@ -1 +1,9 @@
-{% include 'django/forms/widgets/text.html' %}<script>initTagField('{{ widget.attrs.id|escapejs }}', '{{ widget.autocomplete_url|escapejs }}', {% if widget.tag_spaces_allowed %}true{% else %}false{% endif %});</script>
\ No newline at end of file
+{% include 'django/forms/widgets/text.html' %}
+<script>
+    initTagField(
+        '{{ widget.attrs.id|escapejs }}',
+        '{{ widget.autocomplete_url|escapejs }}',
+        {%if widget.tag_spaces_allowed %}true{% else %}false {% endif %},
```
tomdyson: `%if` -> `% if`

wagtail/admin/templates/wagtailadmin/widgets/tag_widget.html
```diff
@@ -1 +1,9 @@
...
+        {% if widget.tag_limit %}{{widget.tag_limit}}{%else %}null{% endif %}
```
tomdyson: `%else` -> `% else`

contactr2m: fixup -> 8d8081c

Thank you for taking a quick look.

lb-:
@contactr2m  - epic stuff! Thanks for the contribution, I have added a few tests and merged this in, I put 'Mani' as your name for the changelog, let us know if that should change.

Thanks @tomdyson  for the initial review

lb-:
Merged via https://github.com/wagtail/wagtail/commit/87bbc31d88ca747584a47836985ec20677a2de82

PR #5124, opened 2019-03-07
Refactor page listing headers into a template tag

gasman:
Introduce a `table_header_label` tag to replace the template-tag-soup that generates the URLs and classnames for the ordering links in the page listing header.

(Ideally we'd go on to introduce this tag on other listings, but it seems that very few of them have fully-functional bidirectional sorting right now, so I'll leave that for another PR...)


gasman:
Thanks for the review @jonnyscholes!

PR #5128, opened 2019-03-08
Simplify pagination_nav.html include

gasman:
Rather than having messy individual cases for searching vs not searching vs no base URL, use the `{% querystring %}` tag to consistently preserve all query params. Also eliminate the `replace_page_param` tag, which just duplicated `{% querystring %}` but with less functionality.


gasman:
`wagtailadmin/pages/listing/_pagination.html` and the `{% paginate %}` tag also seem to duplicate a lot of this functionality, but that's something for another PR.

gasman:
Merged in 23a89774adef4f609e33eb3feece90d6535588ee (deprecation note added to docs, and the erroneous `RemovedInWagtail26Warning` changed to `RemovedInWagtail27Warning`)

PR #5132, opened 2019-03-11
[5130] Tweaks to text diffing

### LINKED ISSUE #5130: FieldPanel.get_comparison_class() doesn't seem to act normally
My previous messages were not clear at all. Sorry, here is a detailed issue

### Issue Summary
When you compare revisions with `CharField`, `TextField `or `RichTextField`, the output doesn't seem right.
`FieldPanel.get_comparison_class()` doesn't catch `CharField `and `TextField`, `diff_text`() doesn't work the right way.

### Steps to Reproduce

2. Add a TextField and a RichTextField to the HomePage
3. Edit it and compare revisions

### Technical details

* Python version: 3.7.2
* Django version: 2.1.7
* Wagtail version: 2.4

To solve the problem edited edit_handler.py and compare.py in wagtail/admin/
First I added a new condition at the bottom of `FieldPanel.get_comparison_class()` in edit_handler.py to catch `TextField `and `CharField `and `return compare.TextFieldComparison` if `field `is an instance of one these.
Then in compare.py, I inverted val_a and val_b in `RichTextFieldComparison.htmldiff()` and `TextFieldComparison.htmldiff()`.
Finally I inverted `'addition'` and `'deletion'` in `diff_text()`

All together:
```python
# wagtail/admin/edit_handler.py

# New Imports
from django.db.models.field import TextField, CharField

class FieldPanel(EditHandler):
    ...

    def get_comparison_class(self):
        # Hide fields with hidden widget
        widget_override = self.widget_overrides().get(self.field_name, None)
        if widget_override and widget_override.is_hidden:
            return

        try:
            field = self.db_field

            if field.choices:
                return compare.ChoiceFieldComparison

            if field.is_relation:
                if isinstance(field, TaggableManager):
                    return compare.TagsFieldComparison
                elif field.many_to_many:
                    return compare.M2MFieldComparison

                return compare.ForeignObjectComparison

            if isinstance(field, RichTextField):
                return compare.RichTextFieldComparison

            # New check - if placed before RichTextField it catches the RichTextField too
            if isinstance(field, TextField) or isinstance(field, CharField):
                return compare.TextFieldComparison
    
        except FieldDoesNotExist:
            pass

        return compare.FieldComparison

    ...

# wagtail/admin/compare.py

# invert val_b and val_a
class TextFieldComparison(FieldComparison):
    def htmldiff(self):
        return diff_text(self.val_b, self.val_a).to_html()

# invert val_b and val_a
class RichTextFieldComparison(TextFieldComparison):
    def htmldiff(self):
        return diff_text(
            BeautifulSoup(force_text(self.val_b), 'html5lib').getText(),
            BeautifulSoup(force_text(self.val_a), 'html5lib').getText()
        ).to_html()


# invert addition and deletion
def diff_text(a, b):

    def tokenise(text):
        ...

    a_tok = tokenise(a)
    b_tok = tokenise(b)
    sm = difflib.SequenceMatcher(lambda t: len(t) <= 4, a_tok, b_tok)

    changes = []

    for op, i1, i2, j1, j2 in sm.get_opcodes():
        if op == 'replace':
            for token in a_tok[i1:i2]:
                changes.append(('addition', token))
            for token in b_tok[j1:j2]:
                changes.append(('deletion', token))
        elif op == 'delete':
            for token in a_tok[i1:i2]:
                changes.append(('addition', token))
        elif op == 'insert':
            for token in b_tok[j1:j2]:
                changes.append(('deletion', token))
        elif op == 'equal':
            for token in a_tok[i1:i2]:
                changes.append(('equal', token))
    
    ...
```

the different outputs for the different steps are as follow:
comparison_wagtail

I hope it is clear enough this time.

kaedroho:
Fixes #5130

Thanks to @Zogsha for the suggestion! On merge, please credit them on the changelog.

### DIFF
--- a/wagtail/admin/edit_handlers.py
+++ b/wagtail/admin/edit_handlers.py
@@ -3,7 +3,7 @@ import re
 
 from django import forms
 from django.core.exceptions import ImproperlyConfigured
-from django.db.models.fields import FieldDoesNotExist
+from django.db.models.fields import CharField, FieldDoesNotExist, TextField
 from django.forms.formsets import DELETION_FIELD_NAME, ORDERING_FIELD_NAME
 from django.forms.models import fields_for_model
 from django.template.loader import render_to_string
@@ -501,6 +501,10 @@ class FieldPanel(EditHandler):
 
             if isinstance(field, RichTextField):
                 return compare.RichTextFieldComparison
+
+            if isinstance(field, (CharField, TextField)):
+                return compare.TextFieldComparison
+
         except FieldDoesNotExist:
             pass
 
--- a/wagtail/admin/compare.py
+++ b/wagtail/admin/compare.py
@@ -230,7 +230,14 @@ class ChoiceFieldComparison(FieldComparison):
         val_b = force_str(dict(self.field.flatchoices).get(self.val_b, self.val_b), strings_only=True)
 
         if self.val_a != self.val_b:
-            return TextDiff([('deletion', val_a), ('addition', val_b)]).to_html()
+            diffs = []
+
+            if val_a:
+                diffs += [('deletion', val_a)]
+            if val_b:
+                diffs += [('addition', val_b)]
+
+            return TextDiff(diffs).to_html()
         else:
             return escape(val_a)
 
--- a/wagtail/admin/compare.py
+++ b/wagtail/admin/compare.py
@@ -593,7 +593,7 @@ def diff_text(a, b):
         tokens = []
         current_token = ""
 
-        for c in text:
+        for c in text or "":
             if c.isalnum():
                 current_token += c
             else:
--- a/wagtail/admin/tests/test_compare.py
+++ b/wagtail/admin/tests/test_compare.py
@@ -67,9 +67,44 @@ class TestTextFieldComparison(TestFieldComparison):
         self.assertIsInstance(comparison.htmldiff(), SafeString)
         self.assertTrue(comparison.has_changed())
 
+    def test_from_none_to_value_only_shows_addition(self):
+        comparison = self.comparison_class(
+            SimplePage._meta.get_field('content'),
+            SimplePage(content=None),
+            SimplePage(content="Added content")
+        )
 
-class TestRichTextFieldComparison(TestTextFieldComparison):
+        self.assertEqual(comparison.htmldiff(), '<span class="addition">Added content</span>')
+        self.assertIsInstance(comparison.htmldiff(), SafeString)
+        self.assertTrue(comparison.has_changed())
+
+    def test_from_value_to_none_only_shows_deletion(self):
+        comparison = self.comparison_class(
+            SimplePage._meta.get_field('content'),
+            SimplePage(content="Removed content"),
+            SimplePage(content=None)
+        )
+
+        self.assertEqual(comparison.htmldiff(), '<span class="deletion">Removed content</span>')
+        self.assertIsInstance(comparison.htmldiff(), SafeString)
+        self.assertTrue(comparison.has_changed())
+
+
+class TestRichTextFieldComparison(TestFieldComparison):
     comparison_class = compare.RichTextFieldComparison
+
+    # Only change from FieldComparison is the HTML diff is performed on words
+    # instead of the whole field value.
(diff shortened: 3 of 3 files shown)


wagtail/admin/edit_handlers.py
```diff
@@ -515,6 +515,10 @@ def get_comparison_class(self):
 
             if isinstance(field, RichTextField):
                 return compare.RichTextFieldComparison
+
+            if isinstance(field, (CharField, TextField)):
```
kaedroho: Should we add a check here so that we don't use ``TextFieldComparison`` if ``choices`` are set on ``CharField``?

chosak:
@Zogsha @kaedroho could you provide some example content that can be used to duplicate the bug in #5130? I played around with a simple page with a CharField, TextField, and RichTextField and wasn't sure what problems are being referred to in that issue.

I tried the code in this PR and was able to trigger a TypeError when comparing a revision with CharField content to a revision with an empty CharField (`blank=True, null=True`). Specifically, [this line](https://github.com/wagtail/wagtail/blob/9f2e4033ea5b2258b9f594e52b46258ce19581b6/wagtail/admin/compare.py#L609) raises an exception when the value being tokenized is None.

Zogsha:
@chosak It's not really a bug. But I mentioned that text diffing didn't apply for `Charfield` and `TextField` and I didn't think it was intended. The current comparison displays all the original content in red and all the new in green, instead of showing the words that changed like in `RichTextField`. It's really hard to read for long texts.
Once you had commented I looked at the code and noticed that `TextFieldComparison` is actually never used. `RichTextFiedComparison` inherits from `TextFieldComparison` but only to override its only method. Maybe `RichTextFieldComparison `could inherit directly from `FieldComparison`.
Furthermore I didn't notice before but whenever `diff_text()` is used in the other Comparison classes, the fields pass through `force_text()`. But `force_text(None)` only returns `"None"`. So `"None"` is the content displayed in red or green wether it's a deletion or addition. I tried and that is the case. Is it the case for the other uses of `diff_text()` in the other comparison ? Probably.

I think the best way to solve this is simply to check in `diff_text()` if `text ` is None and change the value to an empty string, before tokenising it. I just tried and it seams to work.
`diff_text(None, "Modified Content").to_html()` returns `<span class="addition">Modified Content</span>`
`diff_text("Original Content", None).to_html()` returns `<span class="deletion">Original Content</span>`

I just added this to diff_text(), just before the line you mentioned.
```
if text is None:
    text = ""
```

wagtail/admin/edit_handlers.py
chosak: I don't think this check is necessary because we already handle fields with choices [above](https://github.com/wagtail/wagtail/blob/e96ad4433ca584eb469bf3278771e4d7bb6fe0a5/wagtail/admin/edit_handlers.py#L505), using `ChoiceFieldComparison` then instead.

chosak:
I've been meaning to come back to this, and got a chance to revisit it today. This change looks :+1: to me, aside from the TypeError mentioned above.

I pushed a rebased branch to my fork at [chosak/kaedroho-5130-text-diff-tweaks](https://github.com/wagtail/wagtail/compare/master...chosak:kaedroho-5130-text-diff-tweaks) that adds two additional commits to this PR.

First, I handled the TypeError so that text comparisons with `None` are properly handled: I agree with @Zogsha above that a better user experience is not to show the word "None" when comparing null text to a value, but to instead just treat it as an empty string, like this:

image

image

From the end user perspective, whether the value is stored as an empty string or a null value is an implementation detail, and it doesn't make sense to show "None" in the UI in one case but not the other.

Similarly, and again suggested by @Zogsha, I modified the behavior of `ChoiceFieldComparison` so that "None" is hidden in favor of just not showing the old/new value if it is null.

(I was wondering if this would make sense for `FieldValueComparison` in general, i.e. for all field types, but that'd be a larger change to the user experience than this PR is trying to address.)

I added a few new unit tests as well; I'd be happy to push to this PR or open a subsequent one if these changes are welcome.

chosak:
Changes and tests have been rebased and merged in 40dbb94.

Issue #5134, opened 2019-03-12
The HTML tag autocomplete=off is not recognised when two DateTimeField fields are in place


kevinhowbrook:
### Issue Summary

When using 2 instances of DateTimeField, the autocomplete=off implemented in https://github.com/wagtail/wagtail/pull/4793 is visible in the HTML for both fields, but only works for one.

![error example](https://i.imgur.com/0Cvsc0s.jpg)

### Steps to Reproduce

1. Start a new project with `wagtail start myproject`
2. Edit home.models.HomePage as follows...

```[python]
from django.db import models

from wagtail.core.models import Page
from wagtail.admin.edit_handlers import FieldPanel

class HomePage(Page):
    publication_date = models.DateTimeField(null=True, blank=True)
    updated_date = models.DateTimeField("Modified date", blank=True, null=True)

    content_panels = Page.content_panels + [
        FieldPanel('publication_date'),
        FieldPanel('updated_date'),
    ]

```
3. Run `python manage.py makemigrations` and `python manage.py migrate`
4. Observe the error by edditing the home page and click to focus between the two new date fields

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: yes

### Technical details

* Python version: 3.7.1
* Django version: 2.1.7
* Wagtail version: 2.4
* Operating System: Linux x86_64
* Browser version: Chrome 70.

gasman:
Fixed in #5136

PR #5136, opened 2019-03-13
When two date fields are in the same form, chrome ignores multiple au…

kevinhowbrook:
…tocomplete=off values.

Issue https://github.com/wagtail/wagtail/issues/5134

I've investigated [various fixes](https://stackoverflow.com/questions/15738259/disabling-chrome-autofill) for this known issue with chromes autocomplete, some workarounds suggested adding duplicate input fields, which sounds bad. My suggestion here is taken from what people suggest doing with password update fields.

### DIFF
--- a/wagtail/admin/tests/test_widgets.py
+++ b/wagtail/admin/tests/test_widgets.py
@@ -94,7 +94,7 @@ class TestAdminDateInput(TestCase):
 
         html = widget.render('test', None, attrs={'id': 'test-id'})
 
-        self.assertInHTML('<input type="text" name="test" autocomplete="off" id="test-id" />', html)
+        self.assertInHTML('<input type="text" name="test" autocomplete="new-date" id="test-id" />', html)
 
         # we should see the JS initialiser code:
         # initDateChooser("test-id", {"dayOfWeekStart": 0, "format": "Y-m-d"});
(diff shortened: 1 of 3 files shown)


gasman:
https://developers.google.com/web/fundamentals/design-and-ux/input/forms/#recommended_input_name_and_autocomplete_attribute_values seems to imply that `new-password` has a specific one-off meaning as implemented by Chrome "i.e. don't autocomplete this password field because it's where the user is meant to enter a new one" - is there any reason to believe that variants like `new-date` will work?

kevinhowbrook:
> https://developers.google.com/web/fundamentals/design-and-ux/input/forms/#recommended_input_name_and_autocomplete_attribute_values seems to imply that `new-password` has a specific one-off meaning as implemented by Chrome "i.e. don't autocomplete this password field because it's where the user is meant to enter a new one" - is there any reason to believe that variants like `new-date` will work?

My fix is based on this comment from https://bugs.chromium.org/p/chromium/issues/detail?id=468153#c164:

> As an example, if you have an address input field in your CRM tool that you don't want Chrome to Autofill, you can give it semantic meaning that makes sense relative to what you're asking for: e.g. autocomplete="new-user-street-address".  If Chrome encounters that, it won't try and autofill the field. 

I think an instance where this wouldn't work is if you have another field purposefully named 'new-date'. Then chrome may try and autofill from input to input. So another option could be `new-[random-string]` 
 per field...?

gasman:
@kevinhowbrook Thanks for the clarification - now tested and verified on Chrome 72.0.3626.121.

Merged in 7d583a52b904eb693a744210a9ab3d1af8547eb1

Issue #2901, opened 2016-08-08
Allow editors to rotate images within the Wagtail UI


tmsndrs:
Allow editors to rotate images within the Wagtail UI. It can be a pain to have to do it manually outside of the CMS.

heymonkeyriot:
To say we get this feedback quite a lot from clients.

I think it's one of those features that editors expect to see, and get quite frustrated when it isn't there. It doesn't fit their mental model of what a CMS is, and there isn't an obvious explanation to give them of why it's not there.

The most obvious out of the box example of this working is at https://en.support.wordpress.com/images/image-editing/ but it's also there in Craft, in Drupal (with a module) and a number of other CMSs

I recognise it's not an easy thing for us to introduce, but does feel like something that would bring quite a lot of value to users.

gasman:
In the spirit of "five whys" questioning... I have to wonder, what's the use case for this? (i.e. under what circumstance does a user upload a file with incorrect orientation that has to be fixed?)

If it's for photos taken in landscape vs portrait - since Wagtail 1.4 (/ Willow 0.3) we check the EXIF metadata on images and handle the rotation automatically. Is this perhaps a sign that this feature isn't working as intended? Or do we just keep encountering files where the EXIF data is missing or incorrect?

(Also, just to check - am I right in thinking that the request here is for rotating in multiples of 90˚? Or is there demand for supporting 'free' rotation too?)

mrchrisadams:
Hi there, when working on a wagtail site recently I stumbled over this too.

In my case, I was migrating an existing site from Django CMS, and I wanted to use Wagtail's image tags in templates, instead of the previous hand-rolled image tag implementation.

I was able to work out how to import the photos and associate them with an existing model instance (I've shared the [code in this gist][1], but they didn't all have correct EXIF data, so in some cases they ended up rotated 90 degrees the wrong way, and either myself or admin staff have to fix them manually when we see them.

So, the use case might be automated import of existing images for use in Wagtail, where you don't have reliable EXIF data, but want a quick way to 'fix' the image.

In my case at least, rotation in increments of 90 degrees would be plenty.

### Implementing this:

I've had quick look through Willow, and my guess is that you'd need to support these two actions to have server side rotation through steps of 90 degrees.

http://docs.wand-py.org/en/0.4.3/guide/transform.html#rotation
[...]

PR #5139, opened 2019-03-13
WIP - Add image rotation in image edit screen. (#2901)

mrchrisadams:
I've added this after a quick chat with @thibaudcolas at Wagtail space, to help resolve [issue 2901](https://github.com/wagtail/wagtail/issues/2901), giving us a way to rotate images in the server UI.

### WIP tag removal checklist

- [ ]  Do the tests still pass? (http://docs.wagtail.io/en/latest/contributing/developing.html#testing)
- [x]  Does the code comply with the style guide? (Run `make lint` from the Wagtail root)
- [x]  For Python changes: Have you added tests to cover the new/fixed behaviour?
- [x]  For front-end changes: Did you test on all of Wagtail’s supported browsers? **Please list the exact versions you tested**.
- [x]  For new features: Has the documentation been updated accordingly?

### Da current plan 

- [x] add a basic, unstyled rotate button
- [x] add some js to increment the rotate element when the button is pushed
- [x] add rotate field to the image edit form
- [x] add a code for calling `rotate()` when a user submits the form

When the button is pushed, and we update the rotate value, we use a bit of CSS to add a transform to the image, and animate it using a CSS transition to make it for a user to see what just happened (rotating is a bit jarring without this).

CSS transforms and transition animations seem to well supported, and it saves us needing much in the way of extra js dependencies:

https://caniuse.com/#search=transform
https://caniuse.com/#feat=css-animation

### DIFF
--- a/docs/editor_manual/documents_images_snippets/images.rst
+++ b/docs/editor_manual/documents_images_snippets/images.rst
@@ -28,3 +28,13 @@ ___________
 * To set the focal area, drag a marquee around the most important element of the image.
 * To remove the focal area, hit the button below the image.
 * If the feature is set up in your website, then on the front-end you will see the crop of this image focusing on your selection.
+
+Rotating images
+______________________
+
+Rotate uploaded images with the "Rotate image" button - this will rotate the image in 90 degree steps clockwise.
+
+You will need to save the image for your changes to be stored.
+
+.. Warning::
+    Changing the file rotation will change it on all pages that use the image. You will also need to update the focal point if it is already set.
--- a/rough-plan.md
+++ b/rough-plan.md
@@ -1,0 +1,7 @@
+This is just here so we have a change to make a WIP PR - nuke it afterwards
+
+- add rotate field to form
+- add single button for rotate
+- add css rotate code
+
+We'll just just do this bit for now, so we can review in person at wagtil.space
--- a/wagtail/images/forms.py
+++ b/wagtail/images/forms.py
@@ -1,4 +1,7 @@
+from io import BytesIO
+
 from django import forms
+from django.core.files.images import ImageFile
 from django.forms.models import modelform_factory
 from django.utils.text import capfirst
 from django.utils.translation import ugettext as _
@@ -6,7 +9,6 @@ from django.utils.translation import ugettext as _
 from wagtail.admin import widgets
 from wagtail.admin.forms.collections import (
     BaseCollectionMemberForm, collection_member_permission_formset_factory)
-from wagtail.images import get_image_model
 from wagtail.images.fields import WagtailImageField
 from wagtail.images.formats import get_image_formats
 from wagtail.images.models import Image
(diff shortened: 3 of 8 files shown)


thibaudcolas:
👍 current WIP for reference: https://github.com/simo97/wagtail/commit/ef404c775559722b9ad6be61a2cc6d6ca6ed8f69.

mrchrisadams:
Okay, I think I might need some pointers here on how best to call the new methods on Willow that I added in a previous PR.

### What I think I need to do

I see examples here where you encapsulate a call to willow, by using a context manager, and then calling the Willow method like so:

```python
def get_suggested_focal_point(self):
        with self.get_willow_image() as willow:
            faces = willow.detect_faces()
```

I'm assuming we'd need to do something similar - when we see a rotation in the `request.POST` dict, we'd need to: 

- call some kind of `image.rotate(DEGREES)`  method, that calls willow in a context manager the same way
- save the updated image
- make sure the updated image triggers an update of all the renditions reliant on it

Sound about right?

mrchrisadams:
Okay, spoke to @gasman .

It looks like where we currently save a new image and set hashes, these might be better off in the image. This way, when we call `form.save()` in `wagtail.images.views.edit`, we can can encapsulate the fact that we're saving a form submission, and the view method doesn't need to know whether the form is being updated vs rotated.

https://github.com/wagtail/wagtail/blob/master/wagtail/images/views/images.py#L99

In `imageform.save`, we'd move some of this logic currently in the view like this:

```
if 'file' in form.changed_data:
                # Set new image file size
                image.file_size = image.file.size

                # Set new image file hash
                image.file.seek(0)
                image._set_file_hash(image.file.read())
                image.file.seek(0)
```

Into the form behaviour, and do our check for rotation there, rather than having to mess around with with calls to `form.save()` with `commit=false` and so on,

Thanks Matt!

mrchrisadams:
Alright - I've tested this with the following browsers:

- OS X Firefox - v67
- OS X Chrome - v73
- OS X Safari - v 12.1

I've also added some basic docs - I didn't add a new screenshot, as I wasn't sure about the numbering conventions for the images because they have a numeric naming scheme like `screen32_image_edit_page.png`, I didn't want to mess with this when adding my onw.

I think the last steps here would be to clean up the commits, and see why the tests are failing - since I rebase onto master there's some new bits I think I need to update to get these green again, then I'll remove the WIP tag.

lb-:
@mrchrisadams  - thanks for all the epic work on this, would be great to see this ready to be reviewed fully if you have time to get the last bits done & cleaned up.

In regards to the docs images, I reckon just create one, if the numbers don't line up when this gets merged in one of the core team can rename then to avoid conflicts.

For now we are converting all PRs with WIP in the title to drafts, the CI will still run but helps us have a clearer view of items needing actions. 

Feel free to make a non-draft once this is good to get a full review.

lb-:
Closing as no response for a few years on this and still in draft.

Issue #4562, opened 2018-05-24
Feature request: add more built-in rich text formats


favoyang:
This is a feature request question. When browse the [draftail examples page](https://springload.github.io/draftail/examples/), I could see many formats not included in wagtail 2.1 version (under "All built-in formats" section). 

```
Code: Code
Underline: Underline
Strikethrough: Strikethrough
Mark: Mark
Quotation: Quotation
Small: Small
Sample: Sample
Insert: Insert
Delete: Delete
Keyboard: Keyboard
Superscript: Superscript
Subscript: Subscript
Blockquote
```

Is there a plan to integrate latest version for draftail into wagtail?

thibaudcolas:
Hey @favoyang, thanks for opening this. We initially only added the rich text formats Wagtail was using with its previous editor, but did discuss adding more built-in formats.

I think we would definitely welcome pull requests to add:

- Code (`<code></code>`)
- Blockquote (`<blockquote></blockquote>`)

For other formats, we haven't had a discussion about what should be built-in and what shouldn't.

-----

Coming back to your list, I should add that those things "from Draftail" are just the ones that have predefined styles https://github.com/springload/draftail/blob/df836a502cd4d833c106aa2df7f755254e06ee03/lib/api/constants.js#L53-L112 – it's nothing more than that, so shouldn't weigh in the balance that much when it comes to adding it to Wagtail or not.

I should add though that when I picked things to go on that list, I based it on HTML elements – all of those unambiguously map to an HTML tag. You can see the list of tags in the exporter: https://github.com/springload/draftjs_exporter/blob/220f2b93f2fd39329077058114616314fe495c81/draftjs_exporter/defaults.py#L36-L55.

SalahAdDin:
@favoyang  @thibaudcolas , i added six from that list using `hooks`, it's easy.

thibaudcolas:
Actually, looking at the code again I forgot that the "built-in Draftail formats" go beyond styles. They also come with pre-defined labels and descriptions in english, **and keyboard shortcuts**.

Specifically, here are the ones that come with keyboard shortcuts:

```js
KEYBOARD_SHORTCUTS[INLINE_STYLE.BOLD] = { other: 'Ctrl + B', macOS: '⌘ + B' };
KEYBOARD_SHORTCUTS[INLINE_STYLE.ITALIC] = { other: 'Ctrl + I', macOS: '⌘ + I' };
[...]

PR #5141, opened 2019-03-13
Add more built-in rich text formats

arifin4web:
As discussed in  #4562, we are adding few more additional feature formats/identifier within wagtail so that it become more easier add any/all of these features with our rich text editor. Newly added Identifiers are:

- blockquote

- code
- superscript
- subscript
- strikethrough

Screen Shot 2019-03-14 at 10 48 36


wagtail/admin/wagtail_hooks.py
```diff
@@ -459,6 +459,36 @@ def register_core_features(features):
...
+        })
+    )
+    features.register_converter_rule('contentstate', 'code-block', {
+        'from_database_format': {
+            'code': BlockElementHandler('code-block'),
+        },
+        'to_database_format': {
+            'block_map': {'code-block': 'code'}
+        }
+    })
```
thibaudcolas: We might need to have a look at whether `code-block` can be included or not. I wouldn't be surprised if there things that made it incompatible with Wagtail's rich text processing.

client/src/components/Draftail/Draftail.scss
```diff
@@ -103,3 +103,21 @@ $draftail-editor-font-family: $font-serif;
...
+        box-shadow: none;
+    }
+}
+
+.Draftail-block--blockquote {
+    border-left: 0.25em solid $color-grey-3;
+    color: $color-grey-2;
+    margin: 1em 0;
+    padding: 1em 2em;
+}
```
thibaudcolas: 👍 could you add a screenshot of what this looks like to the PR?

wagtail/admin/wagtail_hooks.py
```diff
@@ -518,3 +548,48 @@ def register_core_features(features):
...
+        })
+    )
+    features.register_converter_rule('contentstate', 'strikethrough', {
+        'from_database_format': {
+            's': InlineStyleElementHandler('STRIKETHROUGH'),
+        },
+        'to_database_format': {
+            'style_map': {'STRIKETHROUGH': 's'}
+        }
+    })
```
thibaudcolas: I think this is missing inline `code` formatting?

arifin4web: Do you mean a new identifier called `inline-code`?

wagtail/admin/wagtail_hooks.py
arifin4web: Dropping `code-block` for now. We can look into this later.

thibaudcolas:
Merged! Thank you @arifin4web 🙂 this will be part of Wagtail v2.5.

Issue #4320, opened 2018-02-27
ModelAdmin inspect view does not use short_description 


lb-:
### Issue Summary

In ModelAdmin the [short description](http://docs.wagtail.io/en/v1.13.1/reference/contrib/modeladmin/indexview.html#modeladmin-list-display) is not used when a field is referenced in the [inspect_view](http://docs.wagtail.io/en/v1.13.1/reference/contrib/modeladmin/inspectview.html).

Note: The documentation does not say this will work in inspect view, but it is somewhere between an expected behaviour and an enhancement :)

### Steps to Reproduce

1. Start a new project using the Wagtail Bakery Demo
2. Edit bakerydemo/base/models/models.py - add a new method on the `People` model.
```python
@register_snippet
class People(ClusterableModel):
    # ... all other things
    def get_full_name(self):
        return self.first_name + " " + self.last_name
    get_full_name.short_description = "Full Name"
```

3. Edit bakerydemo/base/wagtail_hooks.py - edit the `PeopleModelAdmin` class.

Enable the inspect view + use the `get_full_name` method in place of first_name & last_name (on listing view also).
```python
class PeopleModelAdmin(ModelAdmin):
    # model, menu_label, menu_icon - keep as is
    list_display = ('get_full_name', 'job_title', 'thumb_image') # replace first_name and last_name
    inspect_view_enabled = True # enable the inspect view
    inspect_view_fields = ('get_full_name', 'job_title', 'thumb_image') # add get_full_name and other fields
```

4. Open the admin and navigate to People (via Bakery Misc on the left hand side)
* /admin/base/people/
* Note that the listing view works as expected, our short_description of `Full Name` is used in the header, not `get_full_name`.
![wagtail - people](https://user-images.githubusercontent.com/1396140/36757953-94067d74-1c5e-11e8-968f-710ae4a8259b.png)

5. Open the inspect view for the first person (move mouse over row and click 'inspect')
* /admin/base/people/inspect/4/

**Expected behaviour**
* The inspect view field titles use the same `short_description` as the listing view.
* It would be fair to assume (from a developer perspective) that this setting is used throughout any representation of that field in ModelAdmin.

**Actual behaviour**
* The inspect view field title shows the original method name: eg. `get_full_name`.
![wagtail - inspecting person](https://user-images.githubusercontent.com/1396140/36757987-b4451ad2-1c5e-11e8-84a9-550380e71d8f.png)

### Potential Solution

[...]

PR #5147, opened 2019-03-14
Reuse the label_for_field function for fields in the InspectView

leewesleyv:
Summary
---------

In the InspectView the labels used for the fields will either be the field's **name** or **verbose name**. Assigning a short description to a field will have therefore have no effect for this view (#4320). Using the label_for_field function to create the labels in the inspect view will result in equal labels in the IndexView and in the InspectView, making them more predictable and consistent.

### DIFF
--- a/wagtail/contrib/modeladmin/tests/test_page_modeladmin.py
+++ b/wagtail/contrib/modeladmin/tests/test_page_modeladmin.py
@@ -157,9 +157,15 @@ class TestInspectView(TestCase, WagtailTestUtils):
         response = self.get(100)
         self.assertEqual(response.status_code, 404)
 
-    def test_label_display(self):
+    def test_short_description_is_used_as_field_label(self):
+        """
+        A custom field has been added to the inspect view's `inspect_view_fields` and since
+        this field has a `short_description` we expect it to be used as the field's label,
+        and not use the name of the function.
+        """
         response = self.client.get('/admin/modeladmintest/author/inspect/1/')
         self.assertContains(response, 'Birth information')
+        self.assertNotContains(response, 'author_birth_string')
 
 
 class TestEditView(TestCase, WagtailTestUtils):
--- a/wagtail/contrib/modeladmin/tests/test_page_modeladmin.py
+++ b/wagtail/contrib/modeladmin/tests/test_page_modeladmin.py
@@ -103,7 +103,7 @@ class TestCreateView(TestCase, WagtailTestUtils):
 
 
 class TestInspectView(TestCase, WagtailTestUtils):
-    fixtures = ['test_specific.json']
+    fixtures = ['test_specific.json', 'modeladmintest_test.json']
 
     def setUp(self):
         self.login()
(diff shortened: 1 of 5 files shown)


ababic:
@leewesleyv Please do shout if you need any help here :)

leewesleyv:
@ababic sorry for not coming back at you. I'm currently working on a test. I think I will manage but I will let you know if I get stuck. Thank you!

leewesleyv:
@ababic I guess I'm stuck in this loop I need to get out of. The problem is as follows:

I've added a function to the `EventPage` model. 

```python

def event_information_string(self):
        return "Event at {} for only {}!".format(self.location, self.cost)

event_information_string.short_description = "Information"

```

To show the field in the inspect view I have to override `inspect_view_field` in the `EventPageAdmin`.
Doing this however will defeat the purpose of the other tests in `TestInspectView` for example `test_manytomany_output`, where it is stated in the docstring:

```python
"""
Because ManyToMany fields are output InspectView by default, the
`categories` for the event should output as a comma separated list
once populated.
"""
```

So i'm considering splitting this test from the `TestInspectView` but I really think it should be part of that class. What do you think?

ababic:
@leewesleyv Perhaps try adding `categories` (and potentially other fields) to the `inspect_view_fields` list too?

leewesleyv:
That would be an option, but it's basically testing there that without defining any fields it would also get M2M relations. Or do you think that responsibility is for a different test?

ababic:
Ah, in that case... perhaps try going for a different Model/ModelAdmin class to test with? I think there are a few option in there. `EventPage` / `EventPageAdmin` are probably the most well-used.

leewesleyv:
@ababic I've processed your feedback since all the reasons were valid. I've taken a slightly different approach to write the test, I hope it is fine like this.

wagtail/tests/modeladmintest/models.py
```diff
@@ -9,6 +9,11 @@ class Author(models.Model):
     name = models.CharField(max_length=255)
     date_of_birth = models.DateField()
 
+    def author_birth_string(self):
+        return '{} was born in pallet town'.format(self.name)
```
ababic: could we lose the `self.name` from this string? This way we don't have to change the other test to make this work :)

wagtail/contrib/modeladmin/tests/test_page_modeladmin.py
```diff
@@ -157,6 +157,10 @@ def test_non_existent(self):
         response = self.get(100)
         self.assertEqual(response.status_code, 404)
 
+    def test_label_display(self):
+        response = self.client.get('/admin/modeladmintest/author/inspect/1/')
```
ababic: Could you add a short comment to this test to explain briefly explain what you're expecting to see and why? This should make maintenance a little easier.

ababic: OR rename the test to something more explanatory, e.g. `test_short_description_is_used_for_author_birth_string_label()`, and maybe also test that "author_birth_string" is NOT present?

leewesleyv:
@ababic I've resolved the comments :)

ababic:
@leewesleyv Great stuff! thank you. I'll take a final look this afternoon when I get a spare 5 minutes, but it's looking good :)

ababic:
Okay, all working as expected locally :) So, happy to approve!

Thanks again for your time and effort @leewesleyv!

jjanssen:
Hi all, just look through this one. From my guess there are no open issues left and this PR is actually ready to be merged? Is that something I can help with @ababic?

ababic:
@jjansen Yes, this one was all good to go. If you can merge for me, that'd be great (I don't have the necessary permissions yet) :)

ababic:
Merged in 5c336f2. Thanks @leewesleyv :)

PR #5148, opened 2019-03-14
Support validators on StreamField FieldBlocks

tomusher:
This adds support for `validators` on StreamField `FieldBlocks` (`CharBlock`, `TextBlock` etc.) by passing them on to the underlying Django Field to handle.

This does mean it's limited to `FieldBlocks` - I'd have liked to implement the same behaviour on other blocks but that might be for another day.

* Tests still pass
* Code complies with style guide
* Tests added for blocks that are already tested - can add tests for blocks that aren't if required
* No front-end changes
* Documentation has not yet been updated - would like some ideas on how to best approach this; Mention the availability of `validators` for each FieldBlock separately, separate them out in to a new group, or add a note that this applies to some (but not all) basic block types?

### DIFF
--- a/docs/topics/streamfield.rst
+++ b/docs/topics/streamfield.rst
@@ -87,19 +87,22 @@ A single-line text input. The following keyword arguments are accepted:
 ``help_text``
   Help text to display alongside the field.
 
+``validators``
+  A list of validation functions for the field (see `Django Validators <https://docs.djangoproject.com/en/stable/ref/validators/>`__).
+
 TextBlock
 ~~~~~~~~~
 
 ``wagtail.core.blocks.TextBlock``
 
-A multi-line text input. As with ``CharBlock``, the keyword arguments ``required`` (default: True), ``max_length``, ``min_length`` and ``help_text`` are accepted.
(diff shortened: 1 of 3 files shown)


gasman:
Thanks @tomusher - this looks good to me!

Have you tried creating a migration on a StreamField that has validators in place? I seem to recall we've had issues in the past when passing objects from `django.forms` as arguments to blocks - I think it was widget objects previously - because they don't normally appear in model definitions, and so they're missing the methods that define how they get serialised into migrations.

Regarding the documentation, I think it's best to follow the pattern we've used for the `required` argument - give it a full description under `CharBlock`, and then on the other blocks that support it, just mention its existence with no further explanation.

tomusher:
Thanks @gasman!

I've just tested creating migrations which:

* Add validators to a StreamField block without them
* Add additional validators to a StreamField block that already has them
* Removing validators from a StreamField block

These all seemed to create appropriate migrations - but I'm not sure if I should be looking for some other potential issues here?

Also updated the docs as you suggest. There is a slightly more notable change to the docs for `RichTextBlock` here, to keep formatting consistent.

gasman:
Thanks! Merged in 0f8a55a6ceb0c507d69c238c1b6f42a65b2193e0.

Issue #4445, opened 2018-04-04
Make rich text soft line breaks configurable with rich text features


thibaudcolas:
Feature request.

### Issue Summary

The rich text features used to turn on/off heading levels, list items, etc should also make it possible to toggle soft line breaks (`\n` in the editor converted to `<br>` in the HTML output). At the moment line breaks are turned on and there is no way to disable them.

See https://github.com/wagtail/wagtail/issues/4266#issuecomment-364750160 and https://github.com/wagtail/wagtail/pull/4268#issuecomment-365994579 for more context.

The current line break definition is here:

https://github.com/wagtail/wagtail/blob/7841f54fe873d128d21a30cb19752c7081ec7b1a/client/src/components/Draftail/index.js#L90-L93

yanndinendal:
It could also be useful to have an `InlineRichTextField` which would allow editors to add bold, italic and links but prevent them adding line breaks (`br`), or other blocks like like `ul`, `ol`, ...  
And especially **not** wrap the output in a `<p>` tag.

zerolab:
I can see how `InlineRichTextField` would relate to https://github.com/wagtail/rfcs/pull/60

thibaudcolas:
I believe this would be [RFC 46](https://github.com/wagtail/rfcs/pull/46). I’ve made a draft attempt at implementing this (https://github.com/wagtail/wagtail/pull/7249), it worked well, just didn’t have the time to finish it.

thibaudcolas:
Just adding notes to this specific to the page editor 2022 project:

### Tasks

- [ ] Review WIP implementation
- [ ] Decide on breaking change / no breaking change
- [ ] Rebase, tweak, merge

### Acceptance criteria

- [ ] As a developer, I can configure whether a given rich text field has soft line breaks available or not.

### Testing & review plan

- [ ] Cross-browser testing @thibaudcolas
- [ ] Code review @jacobtoppm or @gasman

david-smejkal:
I've tried adding a custom "soft" line break feature to Draftail but unfortunately I couldn't quite get it to behave properly.
I couldn't figure out how to render those line breaks in the admin area after they are saved. Specifically when it loads them from the database and renders them in the the rich text area in Wagtail's backend.

[...]

PR #5149, opened 2019-03-14
WIP Make rich text soft line breaks configurable with rich text features

arifin4web:
WIP for #4445


arifin4web:
@thibaudcolas Is this approach is the right way to go?

thibaudcolas:
Force-pushed rebased onto latest master. I'll hopefully review this on the way back home!

thibaudcolas:
@gasman starting to review this properly, I realise now that we might want to have a deprecation process for this. It's not clear to me what would be a useful deprecation though.

If this gets merged as-is, it means that in the cases where people defined their own `features` for rich text, line breaks will stop being available in the toolbar, and will stop being converted to `<br />` in the content processing. Changing how content is saved / rendered is definitely a breaking change, even if a simple one to fix (add `br` to the features list where you rely on it).

I could think of at least two ways to do this with a deprecation process over two releases:

- Docs-only. Document the need for people to opt-in to `br` feature in the "Upgrade considerations" section of the release, and in the Wagtail docs. It won't do anything until when this is actually implemented, but might be enough.
- Clever content analysis. Merge this PR with extra logic to automagically turn on the `br` feature if existing content (in a given editor) contains line breaks, and `br` isn't turned on. Logging some deprecation warnings at the same time.

I think the content analysis would solve this quite well, but it seems rather complicated for something that's only going to be used for one release as a deprecation path. The "docs-only" option is much simpler, but also means there won't be actual deprecation warnings. What do you think?

allcaps:
@thibaudcolas Is it okay to add this breaking change to a major Wagtail release? Without a deprecation process? Since we have majors more often, we could just set the milestone to 7, merge when time comes, and be done.

I'm not a fan of 'content analysis'. It will add overhead and complexity. Removing BR, because the the content doesn't contain BR, seems fuzzy. It seems perfectly possible that BR feature is needed, but just not in the content yet.

Are warnings to notify that "br" is missing from rich text features a good idea? We could keep these warnings for two releases. Additionally, we could add a "silence-no-br-warnings" feature. Integrators who review their features have two options: Add "br" or "silence-no-br-warnings". The "silence-no-br-warnings" is only a flag, the feature itself should be a no-operation.

If a design decision is made, I'm happy to get this PR over the line.

gasman:
Coming back to this with fresh eyes - I think @thibaudcolas's suggestion of turning on the `br` feature if a line break already exists in a field makes a lot of sense. Calling it "content analysis" makes it sound like an unreliable fuzzy heuristic, which I think is misleading...

As I see it, this is something we would introduce _after_ developers have been given sufficient notification that they need to add `br` as a feature if they want to continue using line breaks. This way, if developers miss the memo, there will be a period of time when the "damage" will be limited to them going "hmm, why can't I add line breaks here? Oh, I see, I need to add `br` to the features now" rather than Wagtail corrupting their existing content on save.

The question then is what counts as "sufficient notification". Personally I think that losing line breaks as a feature is a minor enough change that we could reasonably switch to the "content analysis" approach straight away, even in a minor release, with a release note and no warning. But if we want to be more cautious and have a warning for now (and maybe make the switch to "content analysis" in 7.0) I'd be fine with that too.

I don't think the `silence-no-br-warnings` pseudo-feature is a good way of handling it - it would mean that people have to add that now, and remove it again at some point in the future, triggering a migration at each step. (It reminds me of the `use_json_field` flag on StreamField, which was strung out across far too many releases...) A traditional settings flag would work, though - something like:

```
RemovedInWagtail70Warning: The BlogPage.body RichTextField [or RichTextBlock] does not have the
`br` feature enabled. In Wagtail 7.0 this will prevent inserting line breaks. To silence this
warning, set WAGTAIL_RICH_TEXT_ALLOW_NO_BR = True
```

thibaudcolas:
All makes sense to me :)

thibaudcolas:
This is ready for someone else to take over. See guidance in #4445.

PR #5150, opened 2019-03-15
Fix AttributeError on Cloudflare cache invalidation (rebase of #5077)

### LINKED ISSUE #5073: AttributeError on Cloudflare cache invalidation

gasman:
Rebase of #5077 with test added. Fixes #5073

Have also added `assertLogs` wrappers to the existing HTTPError / URLError-related tests - besides making the tests more complete this also prevents the error messages from being echoed in the test run output, which is always good :-)

### DIFF
--- a/wagtail/contrib/frontend_cache/backends.py
+++ b/wagtail/contrib/frontend_cache/backends.py
@@ -100,7 +100,7 @@ class CloudflareBackend(BaseBackend):
 
         except requests.exceptions.HTTPError as e:
             for url in urls:
-                logger.error("Couldn't purge '%s' from Cloudflare. HTTPError: %d %s", url, e.response.status_code, str(e.args[0]))
+                logger.error("Couldn't purge '%s' from Cloudflare. HTTPError: %d %s", url, e.response.status_code)
             return
 
         if response_json['success'] is False:
--- a/wagtail/contrib/frontend_cache/backends.py
+++ b/wagtail/contrib/frontend_cache/backends.py
@@ -100,7 +100,7 @@ class CloudflareBackend(BaseBackend):
 
         except requests.exceptions.HTTPError as e:
             for url in urls:
-                logger.error("Couldn't purge '%s' from Cloudflare. HTTPError: %d %s", url, e.response.status_code)
+                logging.exception("Couldn't purge '%s' from Cloudflare. HTTPError: %d", url, e.response.status_code)
             return
 
         if response_json['success'] is False:
--- a/wagtail/contrib/frontend_cache/backends.py
+++ b/wagtail/contrib/frontend_cache/backends.py
@@ -100,7 +100,7 @@ class CloudflareBackend(BaseBackend):
 
         except requests.exceptions.HTTPError as e:
             for url in urls:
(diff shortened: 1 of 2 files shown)


wagtail/contrib/frontend_cache/backends.py
```diff
@@ -100,7 +100,7 @@ def purge_batch(self, urls):
 
         except requests.exceptions.HTTPError as e:
             for url in urls:
-                logger.error("Couldn't purge '%s' from Cloudflare. HTTPError: %d %s", url, e.response.status_code, e.message)
+                logger.error("Couldn't purge '%s' from Cloudflare. HTTPError: %d", url, e.response.status_code)
```
m1kola: Any reason not to use `logging.exception` instead of `logger.error`? Similarly to `logger.error` it will log with the `ERROR` log level, but will also add exception info/traceback.

gasman: The traceback doesn't seem like particularly useful information to me, and I feel that it's just adding noise to the logs:

```
ERROR:root:Couldn't purge 'http://localhost/events/' from Cloudflare. HTTPError: 500
  File "/vagrant/wagtail/wagtail/contrib/frontend_cache/backends.py", line 89, in purge_batch
    headers=headers,
  File "/usr/lib/python3.6/unittest/mock.py", line 939, in __call__
    return _mock_self._mock_call(*args, **kwargs)
  File "/usr/lib/python3.6/unittest/mock.py", line 995, in _mock_call
    raise effect
```
Surely the important information here is that Cloudflare returned a 500 error - I don't think we gain anything from knowing that it happened on line 89 of backends.py...?

m1kola: Traceback is not very useful in tests, but without logging the traceback, how do you know in a real project what is failing? Is it Wagtail's default invalidation functionality or your [custom logic](http://docs.wagtail.io/en/v2.4/reference/contrib/frontendcache.html#invalidating-index-pages) with `PurgeBatch`? Imagine receiving a `400 Bad Request` without a traceback here given that you have custom logic in your project.

gasman: Fair enough! Changed to `logging.exception` in 2f3b47a3dfe77a2e9c7c191a135b64baecc50e61.

gasman:
Thanks @m1kola! Merged in 339b6752af88426cb4c69b964e403f57174c5ecf.

Issue #5065, opened 2019-02-08
adding a max_count for a child page?


JaredOzzy:
I would like the ability to have the max_count on a specific child.

for example i have a Page with child pages, each of the child pages can have 1 detail page of their own and many other types of pages.

here is an example of what i mean. Apologies if its difficult to understand what im trying to ask.
```
class MainPage(PagePlus):
    text = RichTextField()

    content_panels = Page.content_panels + [
        FieldPanel('text'),
    ]

    parent_page_types = ['HomePage']


class DetailPage(PagePlus):
    pass
    parent_page_types = ['MainPage']
    max_count_per_child = 1 <<<<<<<<<<<<<<<<<<<<<

class AnotherPage(Page):
    pass
    parent_page_types = ['MainPage']
```
now in the CMS it would work similarly to the newly added max_count, but only allow 1 detail page per child created. currently if i added the max_count it would allow me to make 1 detail page accross all the child pages, but i would like 1 detail page per child page.

Just a possible addition that could help in some specific cases.

zerolab:
FWIF, you can achieve this as seen in the tests - https://github.com/wagtail/wagtail/blob/master/wagtail/tests/testapp/models.py#L865-L870 and a bit of logic per page type.

gasman:
Hi @JaredOzzy - thanks for the suggestion. As @zerolab mentions, this can be achieved with a custom [`can_create_at`](https://docs.wagtail.io/en/stable/reference/pages/model_reference.html#wagtail.core.models.Page.can_create_at) method, but I expect that this is a common enough setup that it would be worth adding a standard 'shortcut' for this like `max_count`. (I'm not sure about the name, though - `max_count_per_parent` would probably be more accurate...)

JaredOzzy:
@zerolab @gasman thanks! will give it a shot.

leewesleyv:
I hope I didn't take the work from you, but I really wanted to try this out!

JaredOzzy:
> I hope I didn't take the work from you, but I really wanted to try this out!

Not at all! glad you got it working, will be using your work once its live! ;)

gasman:
Implemented in #5151.

PR #5151, opened 2019-03-15
Added an option to the page model to be able to define the amount of …

leewesleyv:
If #5065 is still relevant, I've created a PR for it.

I've defined `max_count_per_parent` on the `Page` model. In the `can_create_at` function I check whether the parent already has reached the maximum amount of children of that specific class. If so, don't let the user create another page.


gasman:
Thanks @leewesleyv - looks good to me! Just needs to be added to the documentation, at `docs/reference/pages/model_reference.rst`.

gasman:
Thanks! Merged in c831d434ab4ea67bf68a39acfd2b97ed75708626.

PR #5153, opened 2019-03-15
Remove imports of django.utils.six

gasman:
django.utils.six is being removed in Django 3.0. Now that we're hopefully on top of Django 2.2 compatibility, it can't hurt to be ahead of the game a bit :-)


kaedroho:
Merged, thanks!

PR #5155, opened 2019-03-16
Add more readable focus state for buttons

mrchrisadams:
Hi there, 

I was working on #2901 earlier today, and when testing it with just a keyboard, I notices that the focus state on inputs wasn't very readable.

You can see an example of what I mean on the user edit screen, but any input with a submit had the same issue, when they had focus:

low-contrast-focus-state

This PR adds some more contrast in the text so it's more readable:

higher-contrast-focus-state

This also applies the same `focus` pseudo class  to `<button>` elements as well now, as when I had a button in the new image rotate feature (#2901) I realised there wasn't an obvious focus state when tabbing through the UI.

> Before submitting, please review the contributor guidelines <http://docs.wagtail.io/en/latest/contributing/index.html> and checkI was workin the following:
>
> * For front-end changes: Did you test on all of Wagtail’s supported browsers? **Please list the exact versions you tested**.

I've tested this on Chrome, and Firefox, but I don't have any linux boxes with GUIs or Windows VMs on my machine, and I think someone will need to check if it looks okay there, but I'm only reusing existing colours and styles, so the chances feel quite low.

Anyway - hope it helps!

C

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/scss/components/_forms.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/components/_forms.scss
@@ -53,6 +53,7 @@ label,
 input,
 textarea,
 select,
+button,
 .halloeditor,
 .tagit {
     appearance: none;
@@ -76,6 +77,7 @@ select,
         background-color: $color-input-focus;
         border-color: $color-input-focus-border;
         outline: none;
+        color: $color-teal-darker;
     }
 
     &:disabled,


lb-:
Tested and works well on the following

* Firefox 65 (Mac OSX)
* Safari 11.1 (Mac OSX)
* Chrome 72 (Mac OSX)

PR-5155-OSX-Chrome-2019-03-19
PR-5155-OSX-Firefox-2019-03-19
PR-5155-OSX-Safari-2019-03-19

lb-:
Validated on Edge and IE 11 FWIW - works great

lb-:
@mrchrisadams  - epic stuff, thanks for finding this and fixing it.

merged in https://github.com/wagtail/wagtail/commit/768cb2a32f2e4c12c823b9da7b7a08e4a70d0922

gasman:
Unfortunately this has broken the styling of the Draftail toolbar, presumably due to the `button` element being given this entire block of CSS rules that it didn't have before:

Screenshot 2019-03-19 at 16 32 23

Reverted in ab986f53ee8a4154a32c89b0a18578d342ef85db.

mrchrisadams:
@gasman gotcha - I'll update the rule to be more specific, and resubmit 👍

lb-:
@gasman thanks for picking that up, I checked a bunch of pages but missed that one sorry.

thibaudcolas:
We’ve added much bigger (and consistent) focus outlines in #5317, so I believe this PR is no longer necessary and will now close this.

PR #5157, opened 2019-03-17
Fix/5154 add names to all action menu items

### LINKED ISSUE #5154: Add 'name' attribute to all sub-classes of ActionMenuItem
### Issue Summary

To use the hook “construct_page_action_menu” it would be nice to be able to identify the menu items in the list. So I suggest to ad an id to the ActionMenuItem.

For instance:
```
class PublishMenuItem(ActionMenuItem):
   id = 'publish'

class SubmitForModerationMenuItem(ActionMenuItem):
    id = 'submit_for_moderation'

etc.
```
Then you can do things like this:

```
def protect_homepage(menu_items, request, context):
	page = context.get('page')	
	if page.specific = HomePage: 	
		new_items = []
		for item in menu_items:
			if item.id == 'unpublish' or item.id == 'delete':
				pass
			else:
				new_items.append(item)
		menu_items[:] = [item for item in new_items]
```

### Technical details

* Wagtail version: 2.4

lb-:
Resolves #5154

* Do the tests still pass? 👍 
* For Python changes: 👍  **please check my approach is ok, could not think of a better way to validate the 'name' attribute has been set**
* For new features: Has the documentation been updated accordingly? 👍 added one minor change, to ensure that our example reflects the usage of adding a name attribute

### DIFF
--- a/docs/reference/hooks.rst
+++ b/docs/reference/hooks.rst
@@ -539,6 +539,7 @@ Hooks for customising the way users are directed through the process of creating
     from wagtail.admin.action_menu import ActionMenuItem
 
     class GuacamoleMenuItem(ActionMenuItem):
+        name = 'action-guacamole'
         label = "Guacamole"
 
         def get_url(self, request, context):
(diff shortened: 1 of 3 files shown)


gasman:
Looks good to me - thanks @lb-! Merged in a2ba2a7a8af9268f2efd0767fe055033c9e2c14a.

lb-:
@gasman  - thanks heaps

Issue #3257, opened 2017-01-04
Support custom link handler for links without a linktype.


danni:
A frequent request in sites is for external links to have the target=_blank attribute set on them. While for internal links we can set a link handler and `register_rich_text_link_handler`, you can do this for external links which don't have a `linktype` set.

Propose supporting a `None` link handler for links without `linktype` set which defaults to identity, but can be overridden as required.

gasman:
👍  This would be my preferred approach to implementing #1167, as it means the handling of the `target` attribute can be kept outside of Wagtail core and implemented as a contrib / external module instead.

thibaudcolas:
Just a note that @arifin4web has been working on this at Wagtail Space, with promising results. From what we tried it should then be very simple to implement things like #1167 server-side.

Relevant SO questions to update once all of this comes through:

- https://stackoverflow.com/questions/33300941/making-external-links-open-in-a-new-window-in-wagtail/34270862#34270862
- https://stackoverflow.com/questions/50042134/how-to-force-external-urls-inserted-in-the-wagtail-draftail-editor-to-open-on-th

thibaudcolas:
This is now implemented thanks to #5159, to be released in Wagtail v2.5. The implementation introduces support for `external` and `email` as separate link types for which handlers can be registered with `features.register_link_type`. It doesn't have a catch-all for other links without linktypes – if people write their custom links, they should use a custom link type.

Here is an example of how it works:

```python
    from django.utils.html import escape
    from wagtail.core import hooks

    def external_link_handler(attrs):
        href = attrs["href"]
        return '<a href="%s" rel="nofollow">' % escape(href)

    @hooks.register('register_rich_text_features')
    def register_external_link(features):
        features.register_link_type('external', external_link_handler)
```

Docs: https://docs.wagtail.io/en/latest/advanced_topics/customisation/rich_text_internals.html#link-rewrite-handlers

Mrfuyu:
For Wagtail 2.9 for me works:

`
#myapp/wagtail_hooks.py
from django.utils.html import escape
from wagtail.core import hooks
from wagtail.core.rich_text import LinkHandler

class NewTabExternalLinkHandler(LinkHandler):
    identifier = 'external'

    @classmethod
    def expand_db_attributes(cls, attrs):
        href = attrs["href"]
        return '<a href="%s" rel="nofollow" target="_blank">' % escape(href)

@hooks.register('register_rich_text_features')
def register_external_link(features):
[...]

PR #5159, opened 2019-03-17
Support additional custom handler for links in Rich text editor.

arifin4web:
As per the concept mention in the #3257, This PR contains the support of custom handler for the links without linktype. But instead of one generic handler, I suggest to use the specific ones based on what we provide from our link chooser inside of the rich text editor. 

The `page` linktypes are implemented (but is undocumented). On top of that we now  will support the custom handler for `external` and `email` type link as well. The appropriate type will be determined based on the provided `href` attribute. These handler will trigger only if they are registered/provided. Otherwise it'll be as previous.

This will enable user to have their own handler for links html. For example, to add a `target="_blank"` attribute in the `external` links ( #1167 ):

```python
def external_linktype_handler(attrs):
    href = attrs["href"]
    # Add `target="_blank"` to external links
    return '<a href="%s" target="_blank">' % escape(href)

@hooks.register('register_rich_text_features')
def register_external_link(features):
    features.register_link_type('external', external_linktype_handler)
```

Or may a different data attribute on `email` links:

```python
def email_linktype_handler(attrs):
    href = attrs["href"]
    return '<a href="%s" data-email="true">' % escape(href)

@hooks.register('register_rich_text_features')
def register_email_link(features):
    features.register_link_type('email', email_linktype_handler)
``` 

I'll update the documentation once we agree that this is the way to go.

### DIFF
--- a/docs/advanced_topics/customisation/rich_text_internals.rst
+++ b/docs/advanced_topics/customisation/rich_text_internals.rst
@@ -97,7 +97,7 @@ The ``register_link_type`` method allows you to define a function to be called w
     def register_report_link(features):
         features.register_link_type('report', report_link_handler)
 
-It is also possible to define link rewrite handler for Wagtail’s built-in ``external`` and ``email`` link types. For example, if you want external links to have a ``rel="nofollow"`` attribute for SEO purposes:
+It is also possible to define link rewrite handler for Wagtail’s built-in ``external`` and ``email`` links, even though they do not have a predefined ``linktype``. For example, if you want external links to have a ``rel="nofollow"`` attribute for SEO purposes:
 
 .. code-block:: python
 
--- a/docs/advanced_topics/customisation/rich_text_internals.rst
+++ b/docs/advanced_topics/customisation/rich_text_internals.rst
@@ -97,6 +97,23 @@ The ``register_link_type`` method allows you to define a function to be called w
     def register_report_link(features):
         features.register_link_type('report', report_link_handler)
 
+It is also possible to define link rewrite handler for the External Links (``external``) and Email Links (``email``). For example, if you want to open your external links in new window, You can write a custom link rewrite handler for ``external`` links.
+
+.. code-block:: python
+
+    from django.utils.html import escape
+    from wagtail.core import hooks
+
+    def external_link_handler(attrs):
+        href = attrs["href"]
+        # Add `target="_blank"` to external links
+        return '<a href="%s" target="_blank">' % escape(href)
+
+    @hooks.register('register_rich_text_features')
+    def register_report_link(features):
+        features.register_link_type('external', external_link_handler)
+
+Similarly you can use ``email`` linktype to add custom rewrite handler for Email links.
 
 Embed rewrite handlers
 ----------------------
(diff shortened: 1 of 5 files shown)


thibaudcolas:
👏 haven't reviewed your PR yet but from the discussion we had @ Wagtail Space the approach made sense to me.

Ping @gasman as well for feedback.

gasman:
Happy with this approach too 👍

docs/advanced_topics/customisation/rich_text_internals.rst
```diff
@@ -97,6 +97,23 @@ The ``register_link_type`` method allows you to define a function to be called w
     def register_report_link(features):
         features.register_link_type('report', report_link_handler)
 
+It is also possible to define link rewrite handler for the External Links (``external``) and Email Links (``email``). For example, if you want to open your external links in new window, You can write a custom link rewrite handler for ``external`` links.
```
thibaudcolas: I'll change this to be about `rel="nofollow"`. Although `target="_blank"` is the use case we've got the most requests for, it's questionable whether implementing it is a good idea (see conversation in #1167). Ideally, Wagtail docs should use examples that are uncontroversial so they are more likely to stay relevant over time.

wagtail/core/tests/test_rich_text.py
```diff
@@ -117,3 +117,61 @@ def test_missing_editor_plugin_returns_none(self):
         self.assertIsNone(
             features.get_editor_plugin('hallo', 'made_up_feature')
         )
+
+
+class TestLinkRewriterTagReplacing(TestCase):
+    def test_should_follow_default_behaviour(self):
```
thibaudcolas: 👌 thank you for taking the time to add tests for the current behavior.

wagtail/core/rich_text/rewriters.py
```diff
@@ -56,12 +56,28 @@ def replace_tag(self, match):
...
         except KeyError:
-            # return ordinary links without a linktype unchanged
-            return match.group(0)
+            link_type = None
+            href = attrs.get('href', None)
+            if href:
+                # From href attribute we tried to detect only the linktypes that we
+                # currently support (`external` and `email`. `page` have default handler)
+                # from our link chooser.
+                if href.startswith(('http', 'https')):
```
thibaudcolas: I'll make a small change to add the `:` there so we can be sure to match a protocol. It shouldn't make any difference with the current link chooser's validation, but there might be people using custom link choosers that have invalid links.

wagtail/core/tests/test_rich_text.py
```diff
@@ -117,3 +117,61 @@ def test_missing_editor_plugin_returns_none(self):
...
+        }
+        rewriter = LinkRewriter(rules)
+
+        page_type_link = rewriter('<a linktype="page" id="3">')
+        self.assertEqual(page_type_link, '<a href="/article/3">')
+
+        # It should call appropriate rule supported linktypes (external or email)
+        # based on the href value
+        external_type_link = rewriter('<a href="https://wagtail.io/">')
+        self.assertEqual(external_type_link, '<a target="_blank" href="https://wagtail.io/">')
```
thibaudcolas: Since there is separate checks for `http` and `https`, I think it's worth having tests for both here.

thibaudcolas:
Aaand it's in! Will be released in v2.5.

PR #5160, opened 2019-03-19
Improve the snippet UI by offering 'Don't delete' actions to the edit form. 

kevinhowbrook:
It's useful to have a 'No, don't delete' button on the delete snippet form. I think it's important to keep UI bits like this as similar as possible, with this addition snippet deletion looks and acts more like page 
deletion which is visually less jarring.

The snippets now have a delete/edit button on hover as well as a 'Don't delete' action on the delete form:

snippet list image

dont delete snippet

* Do the tests still pass? : Yes
* For front-end changes: Did you test on all of Wagtail’s supported browsers? Chrome 70.0.3


gasman:
Looks good to me - thanks @kevinhowbrook! Merged in 1079b7f0212f3ddedfe350a20aa929c6a9c7072b.

(Just changed the button from "No, don't delete it" to "No, don't delete", as this view gets used for multiple deletions too.)

Issue #2728, opened 2016-06-16
Signal on page moving


pyMan:
Hello.

The existing page_published and page_unpublished signals are really useful to purge pages and more related urls from a reverse proxy like Varnish Cache. But that's not enough.

I think there should be a "pre" and a "post" signal like that for page moves too.

When you move a (live) page from a parent to another or when you change the page order under the same parent, the cache about its page and the ex/new sibling ones should be purged too.

Actually I can achive this in this way:

```
class MyPage(Page):
    ....
    def move(self, target, pos = None):
       if self.live and self.get_parent() != target :
            page_unpublished.send(sender = self.specific_class, instance = self.specific)

        super(MyPage, self).move(target, pos = pos)

        if self.live :
            self = Page.objects.get(id = self.id)
            page_published.send(sender = self.specific_class, instance = self.specific, revision = self)
```

Anyway this works only when the page is changing parent, because when just changing the sort order among the same parent it's the wagtail's view "set_page_position" (in the core) who directly calls Page.move (not the specific one).

So what about modify that view to call the "specific" move or what about the pre/post move signals?

kaedroho:
Moving a page also causes the URL of it's children to be updated, so there could potentially be a lot of updates to do. The same issue occurs when a page's slug is changed as well.

If a site has index page purging set up and/or multiple languages (#3024) there moving a page with many children could potentially create a lot of cache purging requests. I think we should investigate bulk purging (#3750) before we create too many ways to generate cache purge requests.

lb-:
Resolved via #5169

PR #5169, opened 2019-03-24
Add 'pre_page_moved' and 'post_page_moved' signals

ababic:
Supersedes #4975

A client project I've been working on recently could really benefit from these changes, so I've gone ahead and implemented myself. Apologies to you, @joncombe, if you'd already been looking at this.

### DIFF
--- a/docs/reference/signals.rst
+++ b/docs/reference/signals.rst
@@ -84,16 +84,16 @@ This signal is emitted from a ``Page`` when the page is unpublished.
 :kwargs: Any other arguments passed to ``page_unpublished.send()``
 
 
-``pre_page_moved`` and ``post_page_moved``
+``pre_page_move`` and ``post_page_move``
 ------------------------------------------
 
 .. versionadded:: 2.6
 
 These signals are emitted from a ``Page`` immediately before and after it is moved.
 
-Subscribe to ``pre_page_moved`` if you need to know values BEFORE any database changes are applied. For example: Getting the page's previous URL, or that of its descendants.
+Subscribe to ``pre_page_move`` if you need to know values BEFORE any database changes are applied. For example: Getting the page's previous URL, or that of its descendants.
 
-Subscribe to ``post_page_moved`` if you need to know values AFTER database changes have been applied. For example: Getting the page's new URL, or that of its descendants.
+Subscribe to ``post_page_move`` if you need to know values AFTER database changes have been applied. For example: Getting the page's new URL, or that of its descendants.
 
 The following arguments are emitted for both signals:
 
@@ -103,7 +103,7 @@ The following arguments are emitted for both signals:
 :parent_page_after: The parent page of ``instance`` **after** moving.
 :url_path_before: The value of ``instance.url_path`` **before** moving.
 :url_path_after: The value of ``instance.url_path`` **after** moving.
-:kwargs: Any other arguments passed to ``pre_page_moved.send()`` or ``post_page_moved.send()``.
+:kwargs: Any other arguments passed to ``pre_page_move.send()`` or ``post_page_move.send()``.
 
 
 Distinguishing between a 'move' and a 'reorder'
@@ -115,7 +115,7 @@ The best way to distinguish between a 'move' and 'reorder' is to compare the ``u
 
 .. code-block:: python
 
-    from wagtail.core.signals import pre_page_moved
+    from wagtail.core.signals import pre_page_move
     from wagtail.contrib.frontend_cache.utils import purge_page_from_cache
 
     # Clear a page's old URLs from the cache when it moves to a different section
@@ -130,4 +130,4 @@ The best way to distinguish between a 'move' and 'reorder' is to compare the ``u
         purge_page_from_cache(kwargs['instance'])
 
     # Register a receiver
-    pre_page_moved.connect(clear_old_page_urls_from_cache)
+    pre_page_move.connect(clear_old_page_urls_from_cache)
--- a/docs/reference/signals.rst
+++ b/docs/reference/signals.rst
@@ -10,14 +10,14 @@ Signals are useful for creating side-effects from page publish/unpublish events.
 For example, you could use signals to send publish notifications to a messaging service, or ``POST`` messages to another app that's consuming the API, such as a static site generator.
 
 
-page_published
---------------
+``page_published``
+------------------
 
 This signal is emitted from a ``PageRevision`` when a revision is set to `published`.
 
-:sender: The page ``class``
+:sender: The page ``class``.
 :instance: The specific ``Page`` instance.
-:revision: The ``PageRevision`` that was published
+:revision: The ``PageRevision`` that was published.
 :kwargs: Any other arguments passed to ``page_published.send()``.
 
 To listen to a signal, implement ``page_published.connect(receiver, sender, **kwargs)``. Here's a simple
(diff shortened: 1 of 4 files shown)


ababic:
It's a no-brainer to me that `wagtail.api` should utilise this signal to help purge pages from the API cache (in addition to the `page_published` and `page_unpublished` signals). However, doing so resulted in a few instances of the following error:

```
File "/home/travis/build/wagtail/wagtail/wagtail/api/v2/signal_handlers.py", line 15, in purge_page_from_cache
purge_url_from_cache(base_url + reverse('wagtailapi_v2:pages:detail', args=(instance.id, )))
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

I initially tried to mitigate this by [raising a warning here](https://github.com/wagtail/wagtail/blob/master/wagtail/api/v2/signal_handlers.py#L14) when `WAGTAILAPI_BASE_URL` isn't set, instead of allowing this exception to be raised. But, I'm having doubts about whether that is really a good approach to be taking, as warnings output in test feedback is only marginally better. There also seems to be some co-dependency between `wagtail.api` and `wagtail.contrib.frontend_cache`, which I'm not quite clear on, so it's difficult to know what the consequences of this would be.

@kaedroho Do you have any thoughts on this?

wagtail/admin/views/pages.py
```diff
@@ -792,7 +792,7 @@ def move_confirm(request, page_to_move_id, destination_id):
 
 
 def set_page_position(request, page_to_move_id):
-    page_to_move = get_object_or_404(Page, id=page_to_move_id)
+    page_to_move = get_object_or_404(Page, id=page_to_move_id).specific
```
tomdyson: Why do we need `specific` here?

docs/reference/signals.rst
```diff
@@ -82,3 +82,17 @@ This signal is emitted from a ``Page`` when the page is unpublished.
...
+----------
+
+.. versionadded:: 2.5
+
+This signal is emitted from a ``Page`` when the page is moved to a different branch of the page tree, or reordered within the same branch.
+
+:sender: The page ``class``
+:instance: The specific ``Page`` instance.
+:position_page: The ``Page`` instance used in conjunction with ``position`` to specify the new tree position for ``instance`` (the intended parent or sibling page).
+:position: A string used in conjunction with ``position_page`` to specify where the new tree position for ``instance``. Common values are `'last-child'` if moving to a different branch, or `'left'` or `'right'` when reordering.
```
tomdyson: "to specify where the new tree position" -> "to specify the new tree position"

wagtail/admin/views/pages.py
ababic: Part of the problem the client is having is that they are subscribing to the `post_save` signal using their specific page types as the `sender`, with the expectation that moving will result in receipt of the signal.

While `Page.move()` does in-turn call `Page.save()`, resulting in a `post_save` signal being emitted, because we are only dealing with a vanilla `Page` object here, that signal is only emitted with `Page` as the sender, and not the fully-fledged page type.

wagtail/contrib/frontend_cache/signal_handlers.py
```diff
@@ -1,7 +1,11 @@
 from django.apps import apps
 
 from wagtail.contrib.frontend_cache.utils import purge_page_from_cache
-from wagtail.core.signals import page_published, page_unpublished
+from wagtail.core.signals import page_moved, page_published, page_unpublished
+
+
+def page_moved_signal_handler(instance, **kwargs):
```
kaedroho: Worth mentioning this separately in release notes.

wagtail/admin/views/pages.py
ababic: The current behaviour is inconsistent with the `move_page()` view just above this, which does ultimately result in a `post_save` signal being emitted for the specific page type.

docs/reference/signals.rst
```diff
@@ -82,3 +82,17 @@ This signal is emitted from a ``Page`` when the page is unpublished.
...
+page_moved
+----------
+
+.. versionadded:: 2.5
+
+This signal is emitted from a ``Page`` when the page is moved to a different branch of the page tree, or reordered within the same branch.
+
+:sender: The page ``class``
+:instance: The specific ``Page`` instance.
+:position_page: The ``Page`` instance used in conjunction with ``position`` to specify the new tree position for ``instance`` (the intended parent or sibling page).
```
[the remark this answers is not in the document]
ababic: Thanks! I actually went for the terminology used in the view here, rather than by the `move()` method itself (so, there is already some consistency). If you can confirm that you'd be happier with `target`, I will happily change :)

wagtail/admin/views/pages.py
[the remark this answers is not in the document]
ababic: @kaedroho Please see above comments. This is more to do with resolving some inconsistency between this view and the `move_confirm()` view just above it.

docs/reference/signals.rst
```diff
@@ -82,3 +82,17 @@ This signal is emitted from a ``Page`` when the page is unpublished.
 :sender: The page ``class``
 :instance: The specific ``Page`` instance.
 :kwargs: Any other arguments passed to ``page_unpublished.send()``
+
+
+page_moved
```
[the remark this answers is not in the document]
ababic: Good shout! I'll update the existing headings too, for consistency

docs/reference/signals.rst
```diff
@@ -82,3 +82,17 @@ This signal is emitted from a ``Page`` when the page is unpublished.
...
+----------
+
+.. versionadded:: 2.5
+
+This signal is emitted from a ``Page`` when the page is moved to a different branch of the page tree, or reordered within the same branch.
```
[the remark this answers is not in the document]
ababic: Understood. I'll go with "section" :)

wagtail/admin/tests/test_pages_views.py
```diff
@@ -2633,6 +2634,25 @@ def test_page_move_confirm(self):
         # Slug should be in error message.
         self.assertIn("{}".format(self.test_page_b.slug), messages[0].message)
 
+    def test_move_triggers_signal(self):
+        # Connect a mock signal handler to page_moved signal
+        mock_handler = mock.MagicMock()
+        page_moved.connect(mock_handler)
```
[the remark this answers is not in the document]
ababic: @kaedroho This is sensible, thanks. Should I update the other tests in this module while I'm here, or should that be done in a separate PR?

wagtail/contrib/frontend_cache/signal_handlers.py
ababic: > I think we should also purge the URL of the page in it's previous location too.

That would be cool. How do you envisage this being supplied to signal handlers? would supplying `old_url_path` (defined currently in `Page.move()`) be enough here?

ababic:
@kaedroho A few more questions to you above :) If you also have any thoughts on the best way to resolve [the above](https://github.com/wagtail/wagtail/pull/5169#issuecomment-475962583) too, that would be great :)

wagtail/contrib/frontend_cache/signal_handlers.py
ababic: @kaedroho I just realised something quite important. If a page has descendants, the paths/URLs of those descendants might also change as a result of `move()`. So, if we want to clear all outdated URLs from the cache, we really need the old URLs for all of those descendants, as well as that of the current page. Is it enough to provide a list of all outdated page URLs, or do you need the full page objects (in their 'before' state) in order to reliably clear things from the front-end cache?

docs/reference/signals.rst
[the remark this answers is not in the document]
kaedroho: We probably should be matching the terminology with the ``move`` method since that's already public API. I'm happier with ``target``, thanks!

ababic:
@kaedroho I've decided to scale back these changes to help with review. I'm going to remove the front-end cache integration for now to reduce sprawl, and just focus on making this signal as 'generically' helpful as possible (in a way that we might improve later)

kaedroho:
I agree, it's probably going to be tricky to do the cache stuff especially since we don't support wildcard cash invalidation yet so purging the subtree could be quite a lot of work in some cases.

We should come back to that bit once we've added support for wildcard invalidation (or, at least, async invalidation)

wagtail/admin/views/pages.py
ababic: @tomdyson / @kaedroho Have removed this change for now to simplify reviewing

ababic:
@kaedroho I believe I've now addressed all of the points raised so far. 

I'm interested to hear what you think about the addition of `changed_url_paths` though. I was trying to strike a balance between supplying all relevant information, while also keeping the number of additional queries to a minimum. 

Adding the `ChangedPageURLPath` class here might possibly be considered overkill, but I figured it'd be convenient if we ever want to build on this (e.g. add helper methods to assist with urls / full url generation)

ababic:
Related issue: #2728

ababic:
I've since realised that `changed_url_paths` here would not actually be very useful in reliably determining old url/full url path paths, because to do that reliably, you need to use `reverse('wagtail_serve')`, and that will only work if the entire database state reflects the point in time you are interested in... if all the `path` and `url_path` values have already changed (which they will have by the time this signal is received), you won't be able to work out the old urls any longer.

I think the sensible solution here is to instead have two signals:

- `pre_page_move()` will be emitted before the database state changes, allowing you to derive **old** URLs if you need to.
- `post_page_move()` will be emitted after the database state changes, allowing you to derive **new** URLs if you need to.

ababic:
This is now ready for review :)

kaedroho:
Sorry for catching this a little late. Could these be renamed to `pre_page_move` and `post_page_move` (drop the 'd'). This is more consistent with Django's wording.

Also, do you think it sounds better to have the noun at the end? `pre_move_page`

ababic:
@kaedroho Yes, I'd agree that's better way to go. Should be sorted now :)

lb-:
@ababic  - thanks so much for this epic work, this has now been merged in via 3fe8775024c4ed5b78f69bd7d3f3fe05a1bafe2d

Issue #4313, opened 2018-02-26
Saved submission and email fields mismatch for AbstractEmailForm


loicteixeira:
When submitting a form to an `AbstractEmailForm` page, the email content will [includes all the form fields](https://github.com/wagtail/wagtail/blob/96d39e45609ccf3c034c2f4267a3b6d221347bd5/wagtail/contrib/forms/models.py#L281) while the saved submission only [includes the cleaned data](https://github.com/wagtail/wagtail/blob/96d39e45609ccf3c034c2f4267a3b6d221347bd5/wagtail/contrib/forms/models.py#L200).

This is problematic for example when you have a captcha field. The result of the (Google) captcha challenge shouldn't be saved in the database nor send via email which forces you to pop it from both `form.cleaned_data` and `form.fields`. Poping a field from the later can potentially dangerous.

I think it should instead use `cleaned_data` in both cases and can submit a PR if you agree on the solution.

lb-:
@loicteixeira using `cleaned_data` has now been implemented via 165c5c0ce5138f35ba313aba54f5a40392e876cc / #5171

Issue #3733, opened 2017-07-24
Inconsistent data format in emails sent by AbstractEmailForm


m1kola:
### Issue Summary

At the moment, [`AbstractEmailForm` sends emails using raw data](https://github.com/wagtail/wagtail/blob/42d274a470fa19d3f60a24361d5153d05ce72cdf/wagtail/wagtailforms/models.py#L299) from the user's input which means that output (email body) will be inconsistent for fields like `DateField` that accept different input formats (both `%m/%d/%Y`and `%m/%d/%y` are valid, for example).

### Steps to Reproduce

1. Create a form with `DateField`.
2. Submit the form with value in format `%m/%d/%Y` (`23/01/2017`, for example)
3. Submit the form with value in format `%m/%d/%y` (`23/01/17`, for example)
4. Check both emails: they will contain `23/01/2017` and `23/01/17` respectively.

I think, both emails should be in single format (let's say in `SHORT_DATE_FORMAT` for `DateField` fields).

Some services can use emails sent by Wagtail for integration purposes, so single format for dates will make integrations easier.

### Technical details

* Wagtail version: all version (tested on 1.6 and 1.11)

lb-:
Merged in via PR #5171

PR #5171, opened 2019-03-25
Improve formatting of dates in AbstractEmailForm

haydnnewport:
In #3733 there is a request to format dates consistently in the emails that `AbstractEmailForm` sends - It seems the simplest way to do this is to use the `form.cleaned_data` of the submitted form, where those dates have already been parsed into `datetime` objects, and in doing so, could tie up #4313 as well.

In implementing this, I've split out the rendering portion of `AbstractEmailForm.send_mail`, so that it could be overridden independently.

I've also added tests to check that the fields are output in the form order in the email, and to check that various date/datetime input formats all normalise to the correct formats.

### DIFF
--- a/wagtail/contrib/forms/models.py
+++ b/wagtail/contrib/forms/models.py
@@ -281,11 +281,13 @@ class AbstractEmailForm(AbstractForm):
 
     def render_email(self, form):
         content = []
-        for field in form:
-            value = field.value()
+        for field, value in form.cleaned_data.items():
             if isinstance(value, list):
                 value = ', '.join(value)
-            content.append('{}: {}'.format(field.label, value))
+
+            label = form.fields.get(field).label
+            content.append('{}: {}'.format(label, value))
+
         return '\n'.join(content)
 
     class Meta:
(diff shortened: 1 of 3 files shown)


lb-:
Merged in via 165c5c0ce5138f35ba313aba54f5a40392e876cc

Thanks again @haydnnewport

lb-:
@gasman this change may cause some differences in form submission emails, if form is customised in a way that form cleaned_data does not match the fields shown to the user (e.g. captcha) that content will now not be in the email content by default.

Let me know if you feel this needs an upgrade consideration note in the release notes.

lb-:
Merged in via 165c5c0ce5138f35ba313aba54f5a40392e876cc

gasman:
@lb- No, I think it's fine - using `cleaned_data` seems like the overwhelmingly sensible thing to do, to the point that anyone relying on the old behaviour is bordering on https://xkcd.com/1172/ :-)

lb-:
Haha love it. Thanks for the quick reply @gasman

Issue #5176, opened 2019-03-26
Support twitter previews in Draftail embeds


fabienheureux:
### Issue Summary
Draftail's EmbedBlock have troubles displaying tweets' informations (thumb / title)
  
### Steps to Reproduce
1. Start a new project with `wagtail start myproject`
2. Add a RichTextField to an example model
3. Add a new content for this example model 
4. Click embed block in draftail toolbar
5. Add a tweet `https://twitter.com/libe/status/1110520362930974721?s=20` 

Expectations: get a sumary (title, description, thumbnail or any other relevant information) displaying in the textarea
Result: Nothing except the generic look of embeds in the editor. 
  
As discussed on slack with @thibaudcolas , this comes from here: https://github.com/wagtail/wagtail/blob/b1fab7084719ddbb88b9581567660930a6c082b4/client/src/components/Draftail/blocks/EmbedBlock.js#L13-L27  

Also, as said on slack, on good generic workaround to this case (and other future cases like this one) could be to display embed source in the preview. 

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: yes

### Technical details
* Python version: 3.7.2
* Django version: 2.1.7
* Wagtail version: 2.4

fabienheureux:
For additionnal information, we can see here that the rendered json from oembed does not return title and thumbnail for twitter. 
http://oembed.frdnspnzr.de/

thibaudcolas:
👌 as discussed, this is because currently the embed block expects embeds to have a `thumbnail` and `title` – those fields aren't always defined.

I think a good fix for this would be to introduce a new rendering of embeds for cases when no thumbnail is available; perhaps displaying a list of field values that seem relevant. We'd have to look at the oembed schema (https://oembed.com) as well as most common embeds to determine what those fields would be. The source URL of the embed would be the lowest common denominator, but we might be able to do more.

Relevant code: https://github.com/wagtail/wagtail/blob/b1fab7084719ddbb88b9581567660930a6c082b4/client/src/components/Draftail/blocks/EmbedBlock.js#L16-L27

thibaudcolas:
Assigning this to @imansyed000

imansyed000:
There are a few options here:
1) ![Screenshot 2019-08-02 at 09 36 34](https://user-images.githubusercontent.com/52247241/62357318-b9062980-b50a-11e9-9067-fdf8544723d3.png)
A html widget provided by twitter

[...]

PR #5177, opened 2019-03-26
Add providerName and url to embed preview

fabienheureux:
Pull request for #5176 
Add providerName to embed description, and add url as well. 
In case title / thumbnail is not present, this provides minimum information to the user.

### DIFF
--- a/client/src/components/Draftail/blocks/EmbedBlock.js
+++ b/client/src/components/Draftail/blocks/EmbedBlock.js
@@ -10,7 +10,7 @@ import MediaBlock from '../blocks/MediaBlock';
  */
 const EmbedBlock = props => {
   const { entity, onRemoveEntity } = props.blockProps;
-  const { url, title, thumbnail } = entity.getData();
+  const { providerName, url, title, thumbnail } = entity.getData();
 
   return (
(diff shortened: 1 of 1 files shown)


tomdyson:
Screenshot from [Squash for this PR](https://fabienheureuxmaster-bpycr.squash.io/admin/pages/3/edit/), after I added a `RichTextField` body to the home page:

image

fabienheureux:
Little auto critic: except the fact that I did not write tests. this PR need a bit of work. Suggestions:
- [ ] Add informations (url / providerName) inside the `MediaBlock` component
- [ ] Maybe display differently the embed if no thumb is provided ?

thibaudcolas:
@fabienheureux yes, see my comment in https://github.com/wagtail/wagtail/issues/5176#issuecomment-476685665.

If you want to make progress on this, I think the next step would be to decide which bits of information from the oembed schema should be displayed when there is no thumbnail. Then I imagine the blocks could display these in order (as a list?), skipping the ones that aren't available.

thibaudcolas:
Closing this in favour of #5545 from @imansyed000, which only needs test to be merge-able.

PR #5179, opened 2019-03-26
Clear pending AJAX request if error occurs on page chooser

gasman:
Little addendum to #5137: the 'pending request' flag should be cleared when an XHR error occurs, as well as on a successful response - otherwise, a temporary server error will cause subsequent requests to be blocked.

This can be verified on a dev instance by opening the page chooser, stopping the server, searching for something, restarting the server and searching again.


lb-:
Confirmed this works as expected.

Firefox seemed to not log anything to the console but Chrome would log an error before this change (in the scenario described above) and would not after the change is applied.

Merging (will update changelog/release notes in a separate commit)

PR #5181, opened 2019-03-27
Fix page chooser not rendering proper get_admin_display_title

haydnnewport:
On a project, I noticed a couple of places in the page chooser where the `get_admin_display_title` function was not showing the correct title.

This was happening both when the title appeared in the chooser breadcrumbs and when the page title appeared in the top "parent page" section. It would also show the incorrect title in the value section of a filled chooser. 

This can be reproduced by adding a page with an overridden `get_admin_display_title` function, then adding another two layers of pages underneath it.
As an example, let the tree be:
```
- Overridden page
    \- Middle page
        \- Leaf page
```
1. Using the page chooser, select the middle page
1. Click "Choose another page" on the same chooser
    The parent page entry for Overridden page at the top of the chooser will show the incorrect title
1. Using the page chooser, descend under the middle page
    The breadcrumb entry for Overridden page will show the incorrect title

This PR adds testing for both the parent page section and the breadcrumbs at the level of the rendered HTML (as extracted from the JSON response), and adds two calls to `.specific` - one in the code for `chooser.browse` and one in the template for `chooser_breadcrumb` - the ensure the override version of `get_admin_display_title` is called.

### DIFF
--- a/wagtail/admin/tests/test_page_chooser.py
+++ b/wagtail/admin/tests/test_page_chooser.py
@@ -233,15 +233,14 @@ class TestChooserBrowseChild(TestCase, WagtailTestUtils):
 
         # Use the leaf page as the chooser parent, so child is in the breadcrumbs
         response = self.client.get(
-            reverse('wagtailadmin_choose_page_child', args=(leaf_page.id,)),
-            params={'page_type': 'wagtailcore.Page'}
+            reverse('wagtailadmin_choose_page_child', args=(leaf_page.id,))
         )
         self.assertEqual(response.status_code, 200)
         self.assertTemplateUsed(response, 'wagtailadmin/chooser/browse.html')
 
         # Look for a link element in the breadcrumbs with the admin title
         self.assertTagInHTML(
-            '<a href="/admin/choose-page/{page_id}/?page_type=wagtailcore.page" class="navigate-pages">{page_title}</a>'.format(
+            '<li><a href="/admin/choose-page/{page_id}/?" class="navigate-pages">{page_title}</a></li>'.format(
                 page_id=self.child_page.id,
                 page_title=self.child_page.get_admin_display_title(),
             ),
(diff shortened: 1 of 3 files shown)


wagtail/admin/tests/test_page_chooser.py
```diff
@@ -202,6 +202,51 @@ def test_with_invalid_page_type(self):
         response = self.get({'page_type': 'foo'})
         self.assertEqual(response.status_code, 404)
 
+    def test_with_admin_display_title(self):
+        # Check the display of the child page title when it's a child
+        response = self.get({'page_type': 'wagtailcore.Page'})
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, 'wagtailadmin/chooser/browse.html')
+
+        self.assertInHTML(self.child_page.get_admin_display_title(), response.json().get('html'))
```
gasman: Rather than comparing against `self.child_page.get_admin_display_title()` here, it's better to be explicit about the result we're expecting, i.e.:

    self.assertInHTML("foobarbaz (simple page)", response.json().get('html'))

This way, it's easier to confirm that the test is doing what it set out to do, just from looking at the test itself - as it stands, I find myself asking "is this test passing because the bug is fixed, or because `get_admin_display_title` just happens to return the plain title instead of the custom one?" (For example, we might be mistakenly testing on a page model that hasn't overridden `get_admin_display_title`, or `self.child_page` might be the non-specific Page object at this point in the code - it's necessary to cross-reference with other bits of the test code to confirm that isn't the case.) Please can you update these assertions (here and in the other tests) so that we're comparing against an explicit result?

haydnnewport: Thanks, I've made that change now. I chose to leave the reference to `child_page.id` in the breadcrumb match because it' not really what we're testing here - Happy to change that too if you don't think that's the right choice

gasman: Thanks! Yep, leaving `child_page.id` in place is fine - there's no real way around it, as you can't guarantee that the database will reliably assign it the same ID each run.

gasman:
Merged in fa5c69ee006f9c701c1868fa7e13a3ac3adcd021 - thanks @haydnnewport!

PR #5183, opened 2019-03-27
Add process_child_object and exclude_fields keyword arguments to Page.copy

kaedroho:
This adds the following keyword arguments to the ``Page.copy`` method:

 - ``process_child_object`` allows a function to be passed which is called if a child object is copied. The passed arguments are ``page``, ``child_relation`` and ``child_object``.
 - ``exclude_fields`` allows a list of fields to be passed that should be excluded from copy

I've also fixed an issue where child relations were not being excluded if they were specified in excluded fields.


ababic:
Merged in 8034d00e. Thank you @kaedroho :)

PR #5184, opened 2019-03-27
Improved rich text rewrite handlers

chosak:
This PR implements the improved version of rich text rewrite handlers as implemented in #4702. It starts with @gasman's 4 commits there, makes a few edits/fixes to them, and adds tests and documentation.

In short this PR modifies how Wagtail rich text rewrite handlers are defined. Instead of being defined solely as functions that take rich text (like `<a linktype="page" id=123>`) and convert it to HTML (like `<a href="/path/to/page/123">`), handlers are now Python classes that implement that logic (as an `expand_db_attributes` method) as well as other pieces of functionality. This will be useful for determining the content being referred to by rich text; currently there's no way to know, for example, what pages are referenced by internal links, which makes it harder to know whether it's safe to delete a page.

A good summary of the changes here for developers are reflected in [this](https://docs.wagtail.io/en/stable/advanced_topics/customisation/rich_text_internals.html) modified documentation page, which can be viewed locally on this branch [here](http://localhost:8080/advanced_topics/customisation/rich_text_internals.html#rewrite-handlers).

The changes proposed in #4702 broke the existing registry API, e.g.

```py
# current signature
features.register_link_type('link_type', handler_fn)

# modified signature in #4702
features.register_link_type(CustomLinkTypeHandler)
```

Per [the deprecation policy](https://docs.wagtail.io/en/v2.4/contributing/release_process.html#deprecation-policy), we want to deprecate changes over two feature releases. For this reason, I've instead implemented a new method so as not to break the existing functionality:

```py
# keep the existing method, to be deprecated
features.register_link_type('link_type', handler_fn)

# totally new method implemented here
features.register_link_type_handler(CustomLinkTypeHandler)
```

I've implemented a kind of wrapper to make the old functionality work the new way, and have marked the old functions with `RemovedInWagtail27Warning`, in the (perhaps optimistic) hope of getting this into 2.5. @gasman I realize that this touches quite a few files; is there time enough left for that, possibly?


gasman:
Thanks @chosak! I've rebased the changes on master here - https://github.com/gasman/wagtail/tree/rich-text-rewrite-handlers-rebase - I didn't trust git to handle the merge conflicts (which now include #5159 as well as #4795) given all the code-moving, so I reapplied the changes manually (which also cleans up the commit history a bit).

Re the fallback for the old function calls, I came up with an alternative approach that avoids making the `identifier` property into a `get_identifier` method: https://github.com/gasman/wagtail/blob/ecd9a4afe4709815ae4bf98ded650c29206766af/wagtail/core/rich_text/feature_registry.py#L130-L133 . I'm also not keen on adding new method names (`register_*_type_handler`) for the new calling convention, so I've tweaked the existing `register_*_type` methods to support both versions until the old one is dropped.

gasman:
Merged (from https://github.com/gasman/wagtail/tree/rich-text-rewrite-handlers-rebase) in 12f7ee2f1a1d002e2a88aa666bb254da109bae2b.

Issue #5175, opened 2019-03-26
StreamField should accept min_num / max_num / block_counts directly


gasman:
`StreamBlock` accepts keyword arguments `min_num` / `max_num` / `block_types` to limit the number of blocks in the stream:

    body = StreamField(
        blocks.StreamBlock([
            ('heading', blocks.CharBlock()),
            ('paragraph', blocks.RichTextBlock()),
        ], max_num=3)
    )

However, these arguments are only available when using an explicit `StreamBlock` - ideally `StreamField` itself should recognise these too, so that the above could be rewritten as:

    body = StreamField([
        ('heading', blocks.CharBlock()),
        ('paragraph', blocks.RichTextBlock()),
    ], max_num=3)

This probably just requires `StreamField` to pass on these arguments when [creating its internal `StreamBlock`](https://github.com/wagtail/wagtail/blob/3134ffa119c15097c8b3c2a0422dd89f7c23304a/wagtail/core/fields.py#L48) (along with corresponding tests / documentation updates).

haydnnewport:
I've created a PR for this on #5185, but it currently doesn't update any documentation - Would adding some arguments definition to `streamfield.rst`, above the "Basic block types" section sound like roughly what you'd be looking for?

paqman85:
I understand that the only bit left on this one is to update the docs. I'll take care of it if it's still an open issue.

gasman:
@paqman85 Not quite - the issue mentioned on https://github.com/wagtail/wagtail/pull/5185#issuecomment-478083855 (ensuring max_num / min_num are respected when the first argument to StreamField is a StreamBlock class rather than a list of blocks) is still open. That may or may not be solvable without going down the rabbit-hole of #4306...

paqman85:
> @paqman85 Not quite - the issue mentioned on https://github.com/wagtail/wagtail/pull/5185#issuecomment-478083855 (ensuring max_num / min_num are respected when the first argument to StreamField is a StreamBlock class rather than a list of blocks) is still open. That may or may not be solvable without going down the rabbit-hole of #4306...

Oh boy. I'll do some digging and see if I can figure it out.

paqman85:
Yes, this is quite the rabbit hole indeed. I'm afraid it's out of my current understanding. 😞 Perhaps not the best labelling for this one?

gasman:
Fixed in #6718

Issue #4306, opened 2018-02-20
StreamField and StreamBlock have confusing behaviour with `required` and `blank` properties


nimasmi:
### Issue Summary

The documentation doesn't correspond to the actual behaviour of an optional top-level StreamBlock, and in some circumstances the behaviour is confusing even without the conflicting documentation.

From http://docs.wagtail.io/en/v1.13.1/topics/streamfield.html#streamblock:

> StreamBlock accepts the following options as either keyword arguments or Meta properties:
>
> `required` (default: True)
> If true, at least one sub-block must be supplied. This is ignored when using the `StreamBlock` as the top-level block of a StreamField; in this case the StreamField’s `blank` property is respected instead.

In fact, setting `required=False` in the StreamBlock and `blank=True` on the StreamField are both necessary to produce an optional field that also looks optional.

### Steps to Reproduce

Configure the following models.py:

``` python
from wagtail.wagtailadmin.edit_handlers import StreamFieldPanel
from wagtail.wagtailcore.blocks import StreamBlock, TextBlock
from wagtail.wagtailcore.models import Page
from wagtail.wagtailcore.fields import StreamField

class TestModel(Page):
    basic = StreamField(StreamBlock([('text', TextBlock())]))
    nonrequired = StreamField(StreamBlock([('text', TextBlock())], required=False))
    blank = StreamField(StreamBlock([('text', TextBlock())]), blank=True)
    nonrequired_blank = StreamField(StreamBlock([('text', TextBlock())], required=False), blank=True)

    content_panels = Page.content_panels + [
        StreamFieldPanel('basic'),
        StreamFieldPanel('nonrequired'),
        StreamFieldPanel('blank'),
        StreamFieldPanel('nonrequired_blank'),
    ]
```

Based on the documentation…

- `basic` should be a required field, which it is
- `nonrequired` should not be required*, and it is not (you can submit an empty field), but the red 'required' asterisk **is** shown
- `blank` should be not required _(in this case the StreamField’s `blank` property is respected instead)_, and yet it is, although the red 'required' asterisk **is not** shown
- `nonrequired_blank` should produce a blank block, but allegedly specifying `required=False` is superfluous, however this is the only way to get a non-required block that also looks optional to site editors.

\* though you'd think that a field with `blank=False` implicitly set would fail validation

### Technical details

* Wagtail version: 1.13.1

coredumperror:
[...]

PR #5185, opened 2019-03-28
Allow streamfields to accept min_num and max_num arguments directly

haydnnewport:
For #5175


haydnnewport:
I've added a second commit here, with some further testing, but I've realised that they're somewhat restating the limit constraint tests in `test_blocks` - I'm happy to prune them, if they are deemed unnecessary

gasman:
Thanks @haydnnewport! Tests look fine to me.

In the process of writing docs I realised that we should also pass these parameters to StreamBlock in the case that the block definition has been given as a StreamBlock subclass, e.g.

    carousel = StreamBlock(CarouselBlock, max_num=10)

This makes them consistent with the behaviour of `blank` vs `required`. However, fixing that has sent me down a rabbit hole of rediscovering #4306 and the fact that those _aren't_ currently consistent...

Will pick this up next week - for now my work-in-progress is at https://github.com/gasman/wagtail/tree/haydnnewport-enhancement/streamfield-minmax .

gasman:
Now re-implemented in #6718 (with tests borrowed from here...)

PR #5186, opened 2019-03-28
Re-add spacing between label and caret in listing headers

thibaudcolas:
This spacing was previously done with whitespace in the HTML, that got removed as part of #5124. This is much better done with CSS.

## Before/after #5124

whitespace-collapsing

## New spacing in CSS

5124-spacing

Tested in Chrome, Firefox, Safari on macOS.


thibaudcolas:
Will close this, as this introduces yet another regression (😐) on other templates that recycle the same CSS, but have duplicated the HTML.

PR #5187, opened 2019-03-28
Make the page reordering UI easier to find

thibaudcolas:
It seems like quite a common issue for users of Wagtail not to realise that Wagtail supports reordering pages within its explorer UI (e.g. for menus, or listings of child pages). This is because the reordering only happens in a toggle-able "reordering" mode, for which the toggle is relatively hard to spot.

This PR makes the target more obvious by adding a label, and changing the icon's glyph to something simpler. Designed and user-tested by @Katielocke3.

page-reordering-after

The toggle also has hover "toggled" states, in line with that of other column headers in the listing.

Here is what it looked like before for reference:

page-reordering-before

--- 

* [x] For front-end changes: Did you test on all of Wagtail’s supported browsers?

Tested in latest Chrome, Firefox, Safari on macOS. Latest MS Edge & IE11 on Windows 8.1. Latest Android Chrome & iOS Safari.


wagtail/admin/static_src/wagtailadmin/scss/components/_listing.scss
```diff
@@ -483,9 +490,7 @@ table.listing {
     }
 
     .table-headers {
-        .ord {
-            padding-right: 0;
-        }
+        height: 35px;
```
thibaudcolas: The height of the header row was dependent on the reordering icon. This makes it the same height whether reordering is enabled or not.

ababic:
This is working well for me locally. Thank you @thibaudcolas and @Katielocke3!

There are some obvious trade-offs here (whitespace/wording), but the overruling factor is just how much more 'findable' this makes the action. Nobody is pretending this is a perfect solution, but it's a clear incremental improvement with very little risk.

I'm happy to approve this.

loicteixeira:
As a side note, this [section](http://docs.wagtail.io/en/v2.4/editor_manual/finding_your_way_around/the_explorer_page.html#reordering-pages) of the doc will need to be updated accordingly.

thibaudcolas:
Thanks @loicteixeira, I completely missed that.

thibaudcolas:
Merged, with the updated docs!

Screenshot 2019-03-29 at 14 26 26

PR #5196, opened 2019-04-01
Formalise support for Django 2.2

gasman:
Django 2.2 has just been released; we're hopefully already compatible, so this PR is just to update CI and documentation.


robvdl:
The 2.5 readme says "This release is compatible with 2.1..." is that correct or should that be 2.2, the heading says Django 2.2 support.

PR #5197, opened 2019-04-02
updated django-taggit to 1.1

auvipy:
attempt to update django taggit version. will fix if something break with this update.

PR #5198, opened 2019-04-02
Bump django-taggit to 0.24

gasman:
Prompted by #5197 - it appears that we can't upgrade to django-taggit 1.x immediately due to breaking changes, but 0.24 is the first version to officially support Django 2.2, so we definitely want to extend our dependency range to include that.


auvipy:
great!! will try to fix 1.1 issues

PR #5201, opened 2019-04-03
Fix typo.  Small grammar changes.

DanAtShenTech:


gasman:
Thanks @DanAtShenTech! Merged in f4ada7cad520fdbfaba38f744a8300ec597a207f - I've kept the comma in the second paragraph (as it's quite an information-heavy sentence that needs the extra 'mental pause' to process, I think...), but happy with the other fixes.

PR #5203, opened 2019-04-04
Implement Page.with_content_json

### LINKED ISSUE #5199: Allow customising the Revision.as_page_object behaviour on page type basis
The ``as_page_object`` method makes sure that a bunch of internal fields are always set to the version in the live version of the page: https://github.com/wagtail/wagtail/blob/c831d434ab4ea67bf68a39acfd2b97ed75708626/wagtail/core/models.py#L1458-L1482

It's possible that a page model may have fields that need to be kept at the live versions as well.

Could we add a new method to Page called ``from_revision`` which could take a revision and return a page object? sub-classes could override this and add their own behaviour.

The initial implementation would be:

```python
@classmethod
def from_revision(cls, revision):
    obj = cls.from_json(revision.content_json)

    # Override the possibly-outdated tree parameter fields from this revision object
    # with up-to-date values
    obj.pk = revision.page.pk
    obj.path = revision.page.path
    obj.depth = revision.page.depth
    obj.numchild = revision.page.numchild

    # Populate url_path based on the revision's current slug and the parent page as determined
    # by path
    obj.set_url_path(revision.page.get_parent())

    # also copy over other properties which are meaningful for the page as a whole, not a
    # specific revision of it
    obj.draft_title = revision.page.draft_title
    obj.live = revision.page.live
    obj.has_unpublished_changes = revision.page.has_unpublished_changes
    obj.owner = revision.page.owner
    obj.locked = revision.page.locked
    obj.latest_revision_created_at = revision.page.latest_revision_created_at
    obj.first_published_at = revision.page.first_published_at

    return obj
```

And I think ``as_page_object`` could be re-implemented as follows:

```python
def as_page_object(self):
    return self.page.from_revision(self)
```

kaedroho:
Fixes #5199

This implements a new method on ``Page`` which allows subclasses to alter behavior when a page object is created from JSON content.

Currently, this is used some internal fields are set to their live values, but it's now possible to reset some fields on the sub-class as well. Also, this provides an API for plugins such as the [wagtail-headless-preview-poc](https://github.com/gasman/wagtail-headless-preview-poc/) to use to convert serialised pages back into page objects with metadata fields set correctly.

### DIFF
--- a/docs/reference/pages/model_reference.rst
+++ b/docs/reference/pages/model_reference.rst
@@ -226,6 +226,8 @@ In addition to the model fields provided, ``Page`` has many properties and metho
         Forms must be a subclass of :class:`~wagtail.admin.forms.WagtailAdminPageForm`.
         See :ref:`custom_edit_handler_forms` for more information.
 
+    .. automethod:: with_content_json
+
 .. _site-model-ref:
 
 ``Site``
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -1406,6 +1406,32 @@ class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):
         context['action_url'] = action_url
         return TemplateResponse(request, self.password_required_template, context)
 
+    def with_content_json(self, content_json):
+        obj = self.specific_class.from_json(content_json)
+
+        # Override the possibly-outdated tree parameter fields from this revision object
+        # with up-to-date values
+        obj.pk = self.pk
+        obj.path = self.path
+        obj.depth = self.depth
+        obj.numchild = self.numchild
+
+        # Populate url_path based on the revision's current slug and the parent page as determined
+        # by path
+        obj.set_url_path(self.get_parent())
+
+        # also copy over other properties which are meaningful for the page as a whole, not a
+        # specific revision of it
+        obj.draft_title = self.draft_title
+        obj.live = self.live
+        obj.has_unpublished_changes = self.has_unpublished_changes
+        obj.owner = self.owner
+        obj.locked = self.locked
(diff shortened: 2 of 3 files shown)


ababic:
@kaedroho how do you feel about introducing a class attribute here to list the fields that should retain their values?

kaedroho:
Class attributes don't work well with mixins because if multiple mixins override the attribute, the final class must manually merge the overrides or things will break.

You see this happen a lot with the `content_panels`/`promote_panels` but I think those are different because developers would normally want full control over the form, but they probably don't care which fields should retain their live values in revisions as long as they work.

A;so, class methods allow us to do things like: `obj.set_url_path(self.get_parent())`

ababic:
> Class attributes don't work well with mixins because if multiple mixins override the attribute, the final class must manually merge the overrides or things will break.

More than anything, I was thinking it would help reduce quite a bit of boilerplate in `Page. with_content_json ()`. The overridability would be more a side-benefit.

> Also, class methods allow us to do things like: obj.set_url_path(self.get_parent())

I'm not proposing we ditch the method. This line could easily stay, since it's not a straight-forward attribute replacement.

ababic:
@kaedroho I've added a test for this... also included `content_type` in the fields that should be preserved from the original page, as I think this might help in situations where revision content type ids become stale due to migration.

kaedroho:
Test looks good to me!

kaedroho:
I think this just needs a mention in the [Methods and Properties](http://docs.wagtail.io/en/v2.5.1/reference/pages/model_reference.html#methods-and-properties) section of the page model docs.

ababic:
Merged in 4f762dc. Thanks again @kaedroho!

PR #5207, opened 2019-04-05
Missing step for Migrating RichTextFields to StreamField

cssidy:
This generated block of code needs to stay or else there will be an error while running ./manage.py migrate. It is misleading that it is not included in this tutorial.

### DIFF
--- a/docs/topics/streamfield.rst
+++ b/docs/topics/streamfield.rst
@@ -993,6 +993,13 @@ If you change an existing RichTextField to a StreamField, and create and run mig
             # leave the dependency line from the generated migration intact!
             ('demo', '0001_initial'),
         ]
+        
+        # leave the generated AlterField intact!
+        migrations.AlterField(
+            model_name='BlogPage',
+            name='body',
+            field=wagtail.core.fields.StreamField([('rich_text', wagtail.core.blocks.RichTextBlock())]),
+        ),
 
         operations = [
             migrations.RunPython(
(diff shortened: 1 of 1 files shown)


docs/topics/streamfield.rst
```diff
@@ -993,6 +993,13 @@ If you change an existing RichTextField to a StreamField, and create and run mig
             # leave the dependency line from the generated migration intact!
             ('demo', '0001_initial'),
         ]
+        
+        # leave the generated AlterField intact!
+        migrations.AlterField(
```
gasman: This should be inside the `operations` list, I think.

gasman:
Looks like this is the issue I was concerned about re putting schema and data changes in one migration: https://docs.djangoproject.com/en/2.2/ref/migration-operations/#runpython

> On databases that do support DDL transactions (SQLite and PostgreSQL), `RunPython` operations do not have any transactions automatically added besides the transactions created for each migration. Thus, on PostgreSQL, for example, you should avoid combining schema changes and `RunPython` operations in the same migration or you may hit errors like `OperationalError: cannot ALTER TABLE "mytable" because it has pending trigger events.`

Since the `AlterField` doesn't actually change anything at the database level (both `RichTextField` and `StreamField` use a text field internally), I believe we should be fine in this case.

gasman:
Merged in 942979fc4ef9dd5fedcc33aadd024d796a03d0ba - thanks!

PR #5208, opened 2019-04-08
Modeladmin search backend

seb-b:
I was using modeladmin for a model that searched several fields, including related lookups. The search was timing out because the orm query that it built up was pretty big.

So I thought it would be a good idea if modeladmin could use the built in wagtail searching, since Wagtail allows the indexing of custom models.

This doesn't include is a way to turn off searching for an `Indexed` model, but I couldn't think of a use case for that. You could just override the `show_search` property.

### DIFF
--- a/docs/reference/contrib/modeladmin/indexview.rst
+++ b/docs/reference/contrib/modeladmin/indexview.rst
@@ -306,7 +306,7 @@ assuming they are indexed, you can override this property with the included
 `WagtailBackendSearchHandler` class e.g.
 
 .. code-block:: python
-    from wagtail.contrib.helpers import WagtailBackendSearchHandler
+    from wagtail.contrib.modeladmin.helpers import WagtailBackendSearchHandler
 
     from .models import IndexedModel
 
--- a/docs/reference/contrib/modeladmin/indexview.rst
+++ b/docs/reference/contrib/modeladmin/indexview.rst
@@ -307,6 +307,7 @@ assuming they are indexed, you can override this property with the included
 
 .. code-block:: python
     from wagtail.contrib.helpers import WagtailBackendSearchHandler
+
     from .models import IndexedModel
 
     class DemoAdmin(ModelAdmin):
@@ -316,6 +317,27 @@ assuming they are indexed, you can override this property with the included
 Note: when using the `WagtailBackendSearchHandler` the `search_fields` are used to limit
 the backend to search only over the specified fields, so those fields must be indexed
 by the backend.
+
+.. _modeladmin_search_handler_extra_search_kwargs:
+
+-------------------------------------------------
+``ModelAdmin.search_handler_extra_search_kwargs``
+-------------------------------------------------
+
+A dictionary of keyword arguments that are used by the specified `search_handler_class` during search.
+e.g. to override the `WagtailBackendSearchHandler` default operator you could do the following:
+
+.. code-block:: python
+    from wagtail.contrib.helpers import WagtailBackendSearchHandler
+    from wagtail.search.utils import OR
+
+    from .models import IndexedModel
+
+    class DemoAdmin(ModelAdmin):
+        model = IndexedModel
+        search_handler_class = WagtailBackendSearchHandler
+        search_handler_extra_search_kwargs = {'operator': OR}
+
 
 .. _modeladmin_ordering:
 
--- a/docs/reference/contrib/modeladmin/indexview.rst
+++ b/docs/reference/contrib/modeladmin/indexview.rst
@@ -301,7 +301,7 @@ your project is using, and without any additional setup or configuration.
 The default value for this property is the `DjangoORMSearchHandler`, which uses
 the Django ORM to perform `icontains` lookups on the fields specified by `search_fields`.
 
-If you would prefer to use the built in Wagtail search backend to search your models,
+If you would prefer to use the built-in Wagtail search backend to search your models,
 assuming they are indexed, you can override this property with the included
 `WagtailBackendSearchHandler` class e.g.
 
--- a/docs/reference/contrib/modeladmin/indexview.rst
+++ b/docs/reference/contrib/modeladmin/indexview.rst
@@ -294,9 +294,9 @@ your project is using, and without any additional setup or configuration.
 
 .. _modeladmin_search_handler_class:
 
--------------------------------------
+-----------------------------------
 ``ModelAdmin.search_handler_class``
--------------------------------------
+-----------------------------------
 
 The default value for this property is the `DjangoORMSearchHandler`, which uses
 the Django ORM to perform `icontains` lookups on the fields specified by `search_fields`.
--- a/docs/reference/contrib/modeladmin/indexview.rst
+++ b/docs/reference/contrib/modeladmin/indexview.rst
@@ -283,10 +283,39 @@ for your model. For example:
 **Expected value**: A list or tuple, where each item is the name of a model field
 of type ``CharField``, ``TextField``, ``RichTextField`` or ``StreamField``.
 
-If your model is indexed using Wagtail's search backend, setting ``search_fields``
-will determine which indexed fields are searched. Otherwise setting ``search_fields``
-will allow searching via Django's QuerySet API, rather than using Wagtail's
-search backend.
+Set ``search_fields`` to enable a search box at the top of the index page
+for your model. You should add names of any fields on the model that should
+be searched whenever somebody submits a search query using the search box.
+
+Searching is handled via Django's QuerySet API by default,
+see `ModelAdmin.search_handler_class`_ about changing this behaviour.
+This means by default it will work for all models, whatever search backend
+your project is using, and without any additional setup or configuration.
+
+.. _modeladmin_search_handler_class:
+
+-------------------------------------
+``ModelAdmin.search_handler_class``
+-------------------------------------
+
+The default value for this property is the `DjangoORMSearchHandler`, which uses
+the Django ORM to perform `icontains` lookups on the fields specified by `search_fields`.
+
+If you would prefer to use the built in Wagtail search backend to search your models,
+assuming they are indexed, you can override this property with the included
+`WagtailBackendSearchHandler` class e.g.
+
+.. code-block:: python
+    from wagtail.contrib.helpers import WagtailBackendSearchHandler
+    from .models import IndexedModel
+
+    class DemoAdmin(ModelAdmin):
+        model = IndexedModel
+        search_handler_class = WagtailBackendSearchHandler
+
+Note: when using the `WagtailBackendSearchHandler` the `search_fields` are used to limit
+the backend to search only over the specified fields, so those fields must be indexed
+by the backend.
 
 .. _modeladmin_ordering:
 
--- a/docs/reference/contrib/modeladmin/indexview.rst
+++ b/docs/reference/contrib/modeladmin/indexview.rst
@@ -283,13 +283,10 @@ for your model. For example:
 **Expected value**: A list or tuple, where each item is the name of a model field
 of type ``CharField``, ``TextField``, ``RichTextField`` or ``StreamField``.
 
-Set ``search_fields`` to enable a search box at the top of the index page
-for your model. You should add names of any fields on the model that should
-be searched whenever somebody submits a search query using the search box.
-
-Searching is all handled via Django's QuerySet API, rather than using Wagtail's
-search backend. This means it will work for all models, whatever search backend
-your project is using, and without any additional setup or configuration.
+If your model is indexed using Wagtail's search backend, setting ``search_fields``
+will determine which indexed fields are searched. Otherwise setting ``search_fields``
+will allow searching via Django's QuerySet API, rather than using Wagtail's
+search backend.
 
 .. _modeladmin_ordering:
 
@@ -646,4 +643,3 @@ See the following part of the docs to find out more:
 
 See the following part of the docs to find out more:
 :ref:`modeladmin_overriding_views`
-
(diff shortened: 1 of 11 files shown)


ababic:
Hi @seb-b. This looks like a really useful addition, thank you! Utilising the search backend where available definitely make more sense than relying on the ORM (and a sensible default). However, I do have some slight concerns regarding backwards compatibility; particularly where developers may be using filters for fields that aren't included in their model's `search_fields` list.

I think we should offer those developers a way to turn the ORM-based behaviour, in case they have issues.

I think we might also want to include a note in the documentation for `ModelAdmin.list_filter` that makes these limitations clearer when adding filters for an indexed model. 

Does this sound okay to you @seb-b?

seb-b:
@ababic sounds good to me :smile:

The other approach I started with was to add another option to `ModelAdmin` class like `search_backend = 'default'` to hide this behaviour behind a switch, also giving users the option to change to a different search backend if there was more than one. Would that make more sense? It would make this a non-breaking change but would mean that the search backend is not the default

ababic:
> The other approach I started with was to add another option to ModelAdmin class like search_backend = 'default' to hide this behaviour behind a switch, also giving users the option to change to a different search backend if there was more than one. Would that make more sense? It would make this a non-breaking change but would mean that the search backend is not the default

Yes, I think this would be a better approach. Another thing I hadn't thought about before is: if the developer is still using the default Wagtail search backend, automatically using the search backend here would create a poor experience. I don't think we really have a choice but to stick with the ORM approach by default, and have this behaviour swappable somehow.

An attribute like `search_backend` would do nicely. Though, I think I might prefer `search_helper_class ` (to follow the modeladmin terminology), and instead of introducing some arbitrary string values, we could make it reference a class that would handle the searching, e.g. `wagtail.contrib.modeladmin.helpers.DjangoORMSearchHandler` (or accept a path string and import lazily when needed). This would make the behaviour easier to customise.

Does this sound okay to you?

seb-b:
Yep sounds good.

wagtail/contrib/modeladmin/helpers/search.py
```diff
@@ -0,0 +1,49 @@
...
+        self.queryset = queryset
+        self.search_fields = search_fields
+
+    def search(self, search_term):
+        """
+        Returns a tuple containing a queryset to implement the search,
+        and a boolean indicating if the results may contain duplicates.
+        """
+        raise NotImplementedError()
+
```
ababic: @seb-b I think we can simplify thing a little here by not even instantiating these classes. If you decorate `search()` with `@classmethod` and just allow it to accept `queryset`, `search_term` and `search_fields`... we can just use that directly in `get_search_results()`.

seb-b: The final class ended up a bit different than this, the `search_fields` is the only parameter and it's used by two methods

wagtail/contrib/modeladmin/helpers/search.py
```diff
@@ -0,0 +1,61 @@
...
+                break
+        return queryset, use_distinct
+
+    @property
+    def show_search_form(self):
+        return bool(self.search_fields)
+
+
+class WagtailBackendSearchHandler(BaseSearchHandler):
+    def search_queryset(self, queryset, search_term, backend='default'):
```
ababic: Hi @seb-b - I like that each handler can accept different arguments, but I don't see a way in the current design for the `IndexView` to easily make use of this. I can't help but think additional methods/attributes on the `ModelAdmin` class would be a good way to facilitate this, as it would make things easier to override (you wouldn't have to create a custom `IndexView` class just to change the behaviour.

A common pattern for this sort of thing is to add an attribute that can be overridden to change the class-wide behaviour (e.g. `search_handler_extra_search_kwargs`), but accompany it with a method to allow more contextual overriding where needed (e.g. `get_search_handler_extra_search_kwargs(self, request, queryset, search_term)`.

Also, I see no reason why `WagtailBackendSearchHandler.search_queryset()` here shouldn't accept all of the arguments accepted by [`SearchableQuerysetMixin.search()`](https://github.com/wagtail/wagtail/blob/master/wagtail/search/queryset.py#L5)

Does this sound good to you?

seb-b:
@ababic I've just encountered a problem when using this. The results returned from `WagtailBackendSearchHandler` would be some sort of subclass of `SearchResults` and not a querset, that means that if the filters want to use `distinct` it'll blow up. So I'm thinking either the `WagtailBackendSearchHandler.search_queryset` should turn the the results back into a queryset or the `IndexView.get_queryset` is reworked to account for this (perhaps distinct is called after filtering and searching)

ababic:
Hey @seb-b, good spot! 

I would propose that  ``IndexView.get_queryset()`` be updated to apply `distinct()` beforehand if necessary (based on `filters_use_distinct`). Then pass an additional `distinct_applied` boolean to `get_search_results()`, and then on to `search_handler.search_queryset()`. The `search_queryset()` method can then use this to figure out whether or not to apply `distinct()` again, and then just return a queryset.

The `WagtailBackendSearchHandler` can probably just ignore `distinct_applied` completely, as I think most search backends only ever return unique results.

Does that sound okay to you?

wagtail/contrib/modeladmin/helpers/search.py
```diff
@@ -28,39 +27,39 @@ def show_search_form(self):
...
 
     @property
     def show_search_form(self):
         return bool(self.search_fields)
 
 
 class WagtailBackendSearchHandler(BaseSearchHandler):
-    def search_queryset(self, queryset, search_term,
+    def search_queryset(self, queryset, search_term, distinct_applied,
                         operator=None, order_by_relevance=True, partial_match=True, backend='default'):
```
ababic: Instead of adding `distinct_applied` explicitly here, could we just add `**kwargs` to the end of this signature?

wagtail/contrib/modeladmin/helpers/search.py
```diff
@@ -28,39 +27,39 @@ def show_search_form(self):
 
 
 class DjangoORMSearchHandler(BaseSearchHandler):
-    def search_queryset(self, queryset, search_term, **extra_search_kwargs):
+    def search_queryset(self, queryset, search_term, distinct_applied, **extra_search_kwargs):
```
ababic: Could we use `**kwargs` in the signature here, for consistency?

wagtail/contrib/modeladmin/helpers/search.py
```diff
@@ -28,39 +27,39 @@ def show_search_form(self):
...
         if not search_term or not self.search_fields:
-            return queryset, False
+            return queryset
 
         if extra_search_kwargs:
             queryset = queryset.filter(**extra_search_kwargs)
```
ababic: @seb-b What's the thinking behind this bit? I'm cautious of blindly interpreting `kwargs` as additional filter arguments here (i don't think the design is clear enough). Also, I'm not sure this is necessary, since `queryset` can be modified freely before passing to the method (if necessary).

wagtail/contrib/modeladmin/views.py
```diff
@@ -577,15 +576,14 @@ def get_queryset(self, request=None):
         ordering = self.get_ordering(request, qs)
         qs = qs.order_by(*ordering)
```
ababic: @seb-b Do we also need to think about preserving ordering here? `WagtailBackendSearchHandler` will just order results by best match by default, but I think there is a way to toggle that behaviour - which we might want to do if custom ordering has been specified?

ababic: This is what I was thinking of:
http://docs.wagtail.io/en/v2.5/topics/search/searching.html#custom-ordering

wagtail/contrib/modeladmin/helpers/search.py
seb-b: The idea was to be able to use the `search_handler_extra_search_kwargs` option to filter the search results, which I don't believe is possible at the moment without overriding the index class. So it should probably be moved to the top of the method def now that I think about it. 
Either way I'm not particulary attached to it, it was just a possible convenience for some devs, happy to remove.

ababic: Okay. I think developers can already override the `ModelAdmin.get_queryset()`, which receives the current `request`, so might be a better place for such customisation. Thanks for removing, anyway :)

ababic:
@seb-b. Really appreciate your time and effort on this, thank you :) It's looking really good. I especially like that you've been updating the docs as you go. I always find that a great way to sanity check my implementation.

From a testing perspective, I think it would be good to establish some separate unit tests for these handler classes, now that they are nicely abstracted? Simple input/output tests are so much more concise and easier to maintain (more fun to write too!). I know I've asked a lot already, but are you okay to make a start on this?

seb-b:
@ababic Thanks for all your feedback. I've added some test cases in the latest commit that cover most of the gotchas I think. But if there's any other functionality you think they should cover let me know :smile:

wagtail/contrib/modeladmin/tests/test_search_handlers.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def setUp(self):
+        self.login()
+
+    def get(self, **params):
+        response = self.client.get('/admin/modeladmintest/author/', params)
+        self.assertEqual(response.status_code, 200)
+        return response
+
+    def test_search_form_absent(self):
```
ababic: Because search handlers are separate components (or units), we should be able to test them in isolation. So forgetting about requests, authentication, URLs, ModelAdmin classes, and just create an instance directly, and test the behaviour matches our expectations. 

Simulating full requests and testing that everything works together like it should (feature testing) is important, but is a difficult/expensive way to test lots of input combinations. It also makes it more difficult to clearly see how the different variations of the component behave and why.

With this in mind, do you think we could make the tests in this module more 'unit test' like, and let the existing tests worry about all the bits fitting together?

ababic: In addition to the above, for `WagtailBackendSearchHandler`, we can probably mock the underlying search backend, and just ensure the correct arguments are passed to `search()`... The search backends themselves are well-tested enough. We just want to know that the handler uses the search backend correctly.

seb-b: ah I see, yep will do

seb-b:
@ababic I reverted the commits with the full test cases and added in some more specific unit tests

ababic:
@seb-b Sorry for the radio silence... It's been a busy few weeks! Things are looking much better on the testing front, thank you. I still need find some time to review this all properly in my local environment. If I have not done this and got back to you in the next 7 days, please shout at me! :)

docs/reference/contrib/modeladmin/indexview.rst
```diff
@@ -287,10 +287,58 @@ Set ``search_fields`` to enable a search box at the top of the index page
...
+
+The default value for this property is the `DjangoORMSearchHandler`, which uses
+the Django ORM to perform `icontains` lookups on the fields specified by `search_fields`.
+
+If you would prefer to use the built-in Wagtail search backend to search your models,
+assuming they are indexed, you can override this property with the included
+`WagtailBackendSearchHandler` class e.g.
+
+.. code-block:: python
+    from wagtail.contrib.helpers import WagtailBackendSearchHandler
```
ababic: The import path here should be:
```
from wagtail.contrib.modeladmin.helpers import WagtailBackendSearchHandler
```

ababic:
Hi @seb-b, I'm afraid I'm noticing a bit of an issue locally in relation to ordering. `IndexView.get_ordering()` automatically [adds `-pk` to the ordering fields when `pk` isn't already present](https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/modeladmin/views.py#L459), which tends to result in an `OrderByFieldError` when searching with the `WagtailBackendSearchHandler` activated, because `index.FilterField('pk')` itself is rarely added to a model's `search_fields`. I think it's  more common to add `index.FilterField('id')` or `index.FilterField('pk')` (or whatever the pk field is). 

To make this work, I think `IndexView.get_ordering()` needs some tweaking to identify the name of the field being used as the primary key, and add ordering for that instead. I also think we need to provide some extra warnings in the docs about using `WagtailBackendSearchHandler`, as literally every field/relationship used in ordering (any column that can be mapped to a field), filters or search_fields must be indexed for the model in order for search to work.

Do you have any thoughts on this?

seb-b:
I ran across this while testing, it didn't seem like a big problem though. Each time it happened the warnings told me exactly what to fix, what field to add and to what model. I feel a warning in the docs would be enough, since this isn't the default search handler.

```
Note:
While using WagtailSearchHandler ensure that any fields specified in the modeladmins ordering attribute are indexed using FilterFields. If no ordering is set then the models primary key must be indexed. 

e.g. #code goes here
```

ababic:
@seb-b I see your point, but I don't feel like like user's having to index `pk` specifically is something they should have to do just to make use of this functionality. Indexing the same field twice with different names certainly feels like a strange thing to force on developers, and I don't know enough about the inner workings of the various supported backends to know if it could cause other problems somewhere down the line.

Your example warning, though definitely helpful, doesn't really explain this away, because the developer isn't specifying that 'pk' be added to ordering... it's something that view is applying without their knowledge.

Don't worry about changing anything more now. I can look at this myself when I come to preparing this to be merged :)

ababic:
Hi @seb-b. How would you prefer to be named in `CONTRIBUTORS.rst`?

ababic:
Scratch that! I just realised you are already in there :)

ababic:
Merged in b839bd6. Thanks @seb-b!

gasman:
Unfortunately it looks like this has broken our Elasticsearch tests (https://travis-ci.org/wagtail/wagtail/builds/543188481 - we've had to put them as 'allow failures' because they had a tendency to fall over due to intermittent server issues). @seb-b @ababic, please could you investigate?

ababic:
> Unfortunately it looks like this has broken our Elasticsearch tests (https://travis-ci.org/wagtail/wagtail/builds/543188481 - we've had to put them as 'allow failures' because they had a tendency to fall over due to intermittent server issues)

@gasman I will look into this. Probably best to fix in a separate PR as this one is done with now.

PR #5215, opened 2019-04-10
Rearrange scss

nmorduch:
This PR is written as a step toward making changes called for in #3804. It moves the bulk of our styles to the `client` folder, and splits them into files approximately according to ITCSS.

Replaces #4979 (so that Squash will run), which replaced #4974.


nmorduch:
@jonnyscholes and @jjanssen, changes look good to me (without having tested yet). Thank you for keeping this going. Looking forward to feedback from @HvandenBerg!

HvandenBerg:
Also tested this and all looks good, no more styling issues in IE11 (and I checked a few other and older versions of a variety of browser for good measure) and no more console errors!

nmorduch:
@jjanssen Are we ready to merge this?

jjanssen:
Yeah, I think so but maybe wait on #5216 for the housekeeping, so we don't end-up in a grey area between 2.5 and 2.6. Are you in the opportunity to merge this when ready or do you like me to take a look at that?

nmorduch:
Definitely after 5216. I'm happy to do the merging, and it's also fine if someone gets to it before me. (I'll say here when I start merging, if no one else does first.)

nmorduch:
Also: thank you again, @HvandenBerg !

nmorduch:
Going to merge this now

nmorduch:
Actually, @jjanssen, would you mind merging? I went to see if any of the commits should be squashed, and there are two commits about updating the font-root, which cancel each other out, and I'm not sure whether that's correct.

jjanssen:
I'll take a look :)

jjanssen:
Squashed the commits and merged with 2668c05

Thanks @nmorduch @jonnyscholes @HvandenBerg for making this happen ❤️

PR #5226, opened 2019-04-12
Stop ModelAdmin from failing when filtering over a foreign key relation

gasman:
Supersedes #4998
As per https://github.com/wagtail/wagtail/pull/4998#issuecomment-471005219, the implementation of `lookup_allowed` is flawed and breaks on some valid lookups while allowing invalid ones. We are therefore better off removing that validation entirely.


gasman:
(merging this now, as this code and proposed fix has had enough eyes on it as part of #4998 :-) )

PR #5227, opened 2019-04-12
[5225] Take number of frames into account when limiting image size

### LINKED ISSUE #5225: Limit max number of pixels for animated GIFs
https://github.com/wagtail/wagtail/pull/5097 implemented a check to prevent images that decompress to too many pixels from being uploaded. It does this by simply multiplying the width with the height.

However, for animated GIFs, we need to also take into account the number of frames in the image.

Note: This number should be total number of pixels in an uncompressed image, so this value shouldn't take into account any savings made by the file format due to techniques like differential encoding.

kaedroho:
Fixes #5225

This depends on https://github.com/wagtail/Willow/pull/74

### DIFF
--- a/wagtail/images/fields.py
+++ b/wagtail/images/fields.py
@@ -1,5 +1,6 @@
 import os
 
+import willow
 from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.core.files.images import get_image_dimensions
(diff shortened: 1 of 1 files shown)


nmorduch:
@kaedroho does this need updates to make the checks pass?

kaedroho:
Just released Willow 1.3 which includes the `get_frame_count` operation

gasman:
Rebased at #5627

PR #5228, opened 2019-04-12
Show image size and frame count on image edit view

kaedroho:
I think the frame count of animated images is of similar importance to the width/height, so this adds a ``frame_count`` field to the ``Image`` model. and adds an "Image size" section in the sidebar which displays the total image size in pixels along with the frame count.

This depends on https://github.com/wagtail/Willow/pull/74


kaedroho:
Might remove total number of pixels, and add total length in seconds. Total length/frame count will only be shown on animated gifs so this PR would have no changes for still images.

PR #5229, opened 2019-04-13
Feature/improve attribute error message

Speedy1991:
Implemented an improved attribute error message


Speedy1991:
I didn't find a good way to keep the stacktrace and modify the message in the exception :/ Maybe someone know a better way?

gasman:
Hi @Speedy1991 - thanks for the PR. The core team discussed this last week, and while we're happy with the principle of customising this message to be more helpful to developers, the wording of the message needs some careful consideration. Simply saying "call specific" is probably not going to be meaningful to many developers, but giving more detail gets complicated quickly too, since the best-practice advice varies for different situations. (Rather than accessing `specific` on an instance, sometimes it's more appropriate to call `specific()` on a queryset, or to rearrange a query expression so that you're querying the more specific model directly.) This probably really needs a dedicated page in the docs explaining what `specific` is actually doing, and giving examples of the different scenarios - then we can link to that page in the new error message.

The stack trace is also something we should find a solution for - we want this override to have as few side effects as possible, given how low-level and frequently-accessed it is. Perhaps there's some similar functionality in the Django codebase that we can use as a reference?

PR #5232, opened 2019-04-13
Add type check to get_rendition_or_not_found

### LINKED ISSUE #5222: Image template tag should give a more helpful error message when passed a non-image value
If the `{% image %}` tag is passed a value which is truthy but not an `Image` instance (or the project's custom equivalent as returned from `get_image_model`), it will fail with an error like `'str' object has no attribute 'get_rendition'`. As we did for the `pageurl` tag in #4228, we should detect invalid values and give a more meaningful error, such as `image tag expected an Image object, got "foo"`.

denizdogan:
Fixes #5222


gasman:
Thanks @denizdogan! Merged in 0840b1a355f24c011427f617a9ba2f65a9cbc89d.

Issue #5092, opened 2019-02-20
WebP Support


edmenendez:
I'm opening this issue so that we can keep track of requirements and how support for WebP should be implemented. I'm hoping core contributors would chip in with advise on what areas of the system to look at and minimum features required for this. 

Something that is needed is for this PR to be released in Willow: https://github.com/wagtail/Willow/pull/70

Wagtail then needs to be tested against that new version of Willow to make sure it works. Additionally, there are places in the code that make some assumptions about image type that need to be fixed. i.e. https://github.com/wagtail/wagtail/blob/master/wagtail/images/models.py#L401 (thanks @gasman for pointing that out). 

I would like to contribute to this feature if there's support within the community for it.

frmdstryr:
:+1:  Changes I made to support this: https://github.com/frmdstryr/wagtail/compare/master...frmdstryr:db4271e350a8e54df05bf0b9b0fb436ec060211f (didn't update tests). It allows uploading webp but converts them to png as some browsers don't display webp images.

Also django's ImageField currently has an error due to this issue https://github.com/django/django/pull/10331. The fix is only coming in 2.2 even though it was applied last year.

onno-timmerman:
I see that this was not added in latest wagtail release? Any reason why wepb is not yet supported?

gasman:
Completed in #5249 / #5629.

PR #5249, opened 2019-04-24
Add webp support

frmdstryr:
For #5092 . Converts images to png by default.


frmdstryr:
Updated, thanks.

specialunderwear:
Merge?

docs/advanced_topics/images/image_formats.rst
```diff
@@ -0,0 +1,49 @@
+Image formats
+=========================
+
+Using the picture element
+-------------------------
+
+The `picture element <https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture>`_
+can be used with wagtails ``format-<type>`` image operation to specify different
```
zerolab: "with wagtails" → "with the Wagtail" (or just "Wagtail's")

docs/advanced_topics/images/image_formats.rst
```diff
@@ -0,0 +1,49 @@
...
+image formats and let the browser choose it's preferred one.
```
zerolab: "it's" → "its"

But more generally, "[..] let the browser choose the one it prefers" reads a bit better

docs/advanced_topics/images/image_formats.rst
```diff
@@ -0,0 +1,49 @@
...
+
+
+To do this with wagtail:
```
zerolab: Wagtail

wagtail/images/models.py
```diff
@@ -403,16 +404,23 @@ def run(self, image, output):
...
+                if not willow.has_animation():
+                    default_conversions['gif'] = 'png'
+
+                # Allow the user to override the conversions
+                conversion = getattr(settings, 'WAGTAILIMAGES_FORMATS', {})
+                default_conversions.update(conversion)
+
+                # Get the converted output format falling back to the original
+                output_format = default_conversions.get(
+                    original_format, original_format)
```
zerolab: accidental line break?

wagtail/images/models.py
```diff
@@ -403,16 +404,23 @@ def run(self, image, output):
...
+                }
 
                 # Convert unanimated GIFs to PNG as well
-                if original_format == 'gif' and not willow.has_animation():
-                    output_format = 'png'
+                if not willow.has_animation():
+                    default_conversions['gif'] = 'png'
+
+                # Allow the user to override the conversions
+                conversion = getattr(settings, 'WAGTAILIMAGES_FORMATS', {})
```
zerolab: Should probably be `WAGTAILIMAGES_FORMAT_CONVERSION` as images already uses format(s)

wagtail/images/models.py
```diff
@@ -432,6 +440,8 @@ def run(self, image, output):
                 return willow.save_as_png(output, optimize=True)
             elif output_format == 'gif':
                 return willow.save_as_gif(output)
+            elif output_format == 'webp' and hasattr(willow, 'save_as_webp'):
+                return willow.save_as_webp(output)
```
zerolab: webp support will come in the next Willow version. before that this could change as:
```python
elif output_format == 'webp':
    if hasattr(willow, 'save_as_webp'):
       return willow.save_as_webp(output)
    # Raise warning when webp was supported but we are falling back to png
    return willow.save_as_png(output)
```

wagtail/images/tests/test_image_operations.py
```diff
@@ -487,6 +494,18 @@ def test_gif(self):
 
         self.assertEqual(out.format_name, 'gif')
 
+    @unittest.skipIf(
+        not webp_supported, "Your willow version does not have webp support")
```
zerolab: TODO: Get rid of this once Willow lands webp support

docs/advanced_topics/images/image_formats.rst
```diff
@@ -0,0 +1,49 @@
+Image formats
```
kaedroho: I think this document could be merged into the main images docs. I wouldn't think this is an uncommon use case and we're documenting the settings there anyway

docs/advanced_topics/images/image_formats.rst
kaedroho: I feel like this line is already implied by the fact that it's in the Wagtail docs, could we use "For example:" instead?

docs/advanced_topics/images/image_formats.rst
zerolab: Do you mean https://docs.wagtail.io/en/stable/topics/images.html ?

I could probably go there, but then that page is turning into a rather long one

wagtail/images/models.py
```diff
@@ -403,16 +404,23 @@ def run(self, image, output):
...
-                # Default to outputting in original format
-                output_format = original_format
-
-                # Convert BMP files to PNG
-                if original_format == 'bmp':
-                    output_format = 'png'
+                # Convert bmp and webp to png by default
+                default_conversions = {
+                    'bmp': 'png',
+                    'webp': 'png',
```
kaedroho: This line feels a bit odd as WEBP is supposed to replace PNG (some day). Maybe we should document the fact that this currently happens and show an example of how to configure Wagtail so that user uploaded WEBP files remain as WEBP

wagtail/images/models.py
kaedroho: Need to add a test for this setting as well

docs/advanced_topics/images/image_formats.rst
kaedroho: We should definitely rethink that document later, but I think it's the best place to document this feature in the short term

zerolab: 👍

gasman: I suggest we name this page "Image file formats" to differentiate it from the concept of image formats in rich text.

wagtail/images/models.py
gasman: @kaedroho Isn't that already covered by https://github.com/wagtail/wagtail/pull/5249/commits/5947f9cf38aa757c6cd14129ecde9ac142945673#diff-9909cf0a580b7bd0cfd1756593179301R29 ?

gasman:
Rebased at https://github.com/gasman/wagtail/commits/feature/webp - will add the `WAGTAILIMAGES_FORMAT_CONVERSIONS` test and submit as a new PR.

gasman:
Rebased in #5629

specialunderwear:
Why do you open a new pull request after rebase? You can just rebase and push to the same branch to update the existing pull request?

gasman:
@specialunderwear If I'm not mistaken, force-pushing a rebased branch means that anyone with an existing checkout of the branch's old state will be unable to pull or push - so it's best avoided when multiple people have collaborated on a branch (especially on a fork that isn't my own).

specialunderwear:
That is very thoughtful of you!

Issue #5252, opened 2019-04-24
Reorganise/refactor all JS into wagtail-client


jonnyscholes:
As part of an overall push to bring our frontend up to date with more modern practices we're going to migrate/update all of our JS components. This is a tracking issue to cover the reorganisation of older JS (mostly in `wagtail/wagtail/admin/static_src/`) in Wagtail to the newer paradigm (see `wagtail/client`).

This will require some refactoring as some of the current code is not designed to be transpiled. That required refactoring and some older managing dependancies may make some of this work quite involved. The plan is to take it feature by feature in PRs.

Once this is done that will enable us to work on removing gulp completely and use only Webpack for our frontend pipeline.

Components to move:

- [ ] blocks/*.js
- [ ] hallo-plugins/*.js
- [ ] modal-workflow.js
- [ ] privacy-switch.js
- [ ] core.js
- [ ] page-chooser-modal.js
- [ ] upgrade_notify.js -> #5256
- [ ] expanding_formset.js
- [ ] page-chooser.js
- [x] userbar.js
- [ ] hallo-bootstrap.js
- [ ] page-editor.js
- [x] submenu.js -> #5254

For the most part I'll be working on the refactoring myself with support from @thibaudcolas - but if you're interested in picking off one of these components let me know :)

kaedroho:
Could we should go the other way and refactor the client into the wagtailadmin app? This would allow us to use the `gettext` function for translations (which must be in an app or Django won't be able to extract the strings).

thibaudcolas:
I believe this is fairly out of date now that most of the code in this list has been moved, so will close this now.

PR #5254, opened 2019-04-24
Refactor submenu JS to remove jquery dep and move to wagtail-client

jonnyscholes:
Part of #5252. This PR moves the primary navigation submenu trigger to `wagtail-client`. I've re-written it from scratch given it's simplicity including removing its jQuery dependance.

As the first component of #5252  it raises a couple of questions:

- Should `client/src/components` only be for React components? My gut feeling is that it should be the place for all JS components given we'll always have react and non-react features of the UI (at least for the foreseeable future). 
- Where possible, should I continue removing jQuery as a dependancy? Is that a long term goal? Personally I'd be in favour of removing it where possible, IE11 support is the main consideration here I think and it's not a hard browser to support using vanilla JS.

* Do the tests still pass? (http://docs.wagtail.io/en/latest/contributing/developing.html#testing) Yes.
* Does the code comply with the style guide? (Run `make lint` from the Wagtail root) Yes.
Chrome 73 MacOS
Safari 12.0 MacOS
Firefox 66.0 MacOS


client/src/components/Submenu/index.js
```diff
@@ -0,0 +1,42 @@
...
+  const primaryNavContainer = document.querySelector('[data-nav-primary]');
+
+  if (!primaryNavContainer) {
+    return;
+  }
+
+  const subMenus = document.querySelectorAll('[data-nav-primary-submenu-trigger]');
+  const activeClass = 'submenu-active';
+
+  subMenus.forEach(subMenu => {
```
lb-: @jonnyscholes - this breaks in IE 11.
I have asked in Slack but assuming we do still support IE 11, this code breaks badly (screenshot attached)

https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach

VirtualBox_MSEdge - Win10_23_05_2019_18_24_28

client/src/components/Submenu/index.js
```diff
@@ -0,0 +1,42 @@
...
+
+      document.addEventListener('mousedown', e => {
+        if (!subMenu.contains(e.target) && subMenu !== e.target) {
+          primaryNavContainer.classList.remove(activeClass);
+          subMenu.parentNode.classList.remove(activeClass);
+        }
+      });
+
+      document.addEventListener('keydown', e => {
+        if (e.keyCode === 27) {
```
lb-: minor item - for the sake of readability it might be nicer to use our core variables for the esc key

wagtail/admin/static_src/wagtailadmin/js/core.js
`wagtail.keys.ESC` - or something like that.

I know the previous code did not do this but just aids in understanding

lb-:
@jonnyscholes  - as per slack & comment, the code will need to be revised to work on IE 11 (note, the link in the code comment has a polyfill if we wanted to adopt that, which will aid in future removal of jquery.

One additional comment, I think the index.js file should be mostly reserved for exports. But looking at the other components sub-folders it appears that no index.js file is created, but simply the element/component name is created as a single file- so maybe rename `client/src/components/Submenu/index.js` to client/src/components/Submenu/Submenu.js`.

For an answer to your questions above.

1. Folder names
- I personally think components should be kept for React components only
- how about 'elements' or 'initialisers' or 'inits' - depending on what other things will go there
- this would help for separating any linting rules (if we needed in the future) as we could add an eslintrc inside the ./src/components folder and it also helps when we want to set up some testing infra to be able to provide different variables/utils to tests in those folders.
- so this would then rename to `client/src/elements/Submenu/Submenu.js`

2. Removal of jquery
- I think this is a good goal, it is very possible to do a lot with vanilla JS and a few polyfills for IE 11, the time will come when we will want to pull it out I feel, and little steps like this help.

client/src/components/Submenu/index.js
jonnyscholes: I wasn't aware of this variable - thanks! Although I have just used `event.key` in another project, and that gives you a human readable value (eg `Enter`, `Shift`). This looks to be IE9+ - I might quickly look further at that and see if there are any caveats there. If that works out we might be able to remove the `wagtail.keys` :)

client/src/components/Submenu/index.js
jonnyscholes: Ah thanks. What do you think about polyfilling IE11 holistically using `@babel/polyfill`? Would mean this issue would come up far less. It will increase the bundle size slightly, but given this is a backend where the full bundle can be cached for the whole admin (and that soonish we intend to load features via code splitting where possible) I think this is ok.

jonnyscholes:
Thanks for the review @lb- :D 

0. Ah yes good call on the `index` -> `Submenu` rename. 

1. All in favour of keeping react stuff separate - but maybe we should make it more explicit? For example we could put them in `client/src/react/components` - this would mean if we ever have react-only layouts we could then also have `client/src/react/layouts`. Maybe I'm overthinking it, but if I was looking for the notification bar (a non-react component) I'd probably look in `components` first if there was a folder called that. 

Probably worth getting input from @thibaudcolas here too.

2. Coolio. I'll keep working towards that goal :)

client/src/components/Submenu/index.js
thibaudcolas: For this specific instance I would suggest to convert the NodeList to an array, then you can iterate over it with array methods as usual. Generally I would avoid using iteration methods on NodeLists because it's quite easy to get them confused with arrays and then to have unexpected results.

So in this case this would be:

```js
const subMenus = [].slice.call(document.querySelectorAll('[data-nav-primary-submenu-trigger]'));
```

… or this same logic wrapped in a util function.

thibaudcolas: (as a side note I don't think `@babel/polyfill` includes polyfills for missing DOM methods like this that are specific to browsers, I think it's only polyfilling the JS stdlib)

thibaudcolas:
I think I find it funny as well to have not-React "vanilla JS DOM manipulation with event listeners" called `component`, as if it was a React component. I'm not sure whether this warrants having a different folder structure or nomenclature, will let you be the judge.

To me the overall goal here is to make it easier for front-end developers to understand the codebase, by separating the front-end building blocks (things in `client/`) from the places where they are used (the rest of the Wagtail codebase), and refactoring code so each building-block is self-contained. So do what feels like it would be the clearest to future readers – either have a separate folder to denote the difference in implementation, or the same folder so there is less time wondering where each building block is.

jonnyscholes:
Perhaps we should go with `includes` for the work I'm doing now - thats a fairly widely used convention. As I do all the refactoring, if common patterns emerge that warrants a different folder structure it's not hard to move files around once they're self contained. Will just leave `components` as it is :)

client/src/components/Submenu/index.js
jonnyscholes: Sure, since we're able to use ES6 features here I'd prefer to use `[...submenus]` though to avoid the need to pull out the verbosity of `[].slice.call(` into a util.

RE `@babel/polyfill` looks like they now advise the use of `core-js` directly now (which afaik does cover some but not all dom apis...). Will look into that in a separate PR :)

lb-:
Love the idea of `includes` - makes sense and if that is an existing convention, even better.

@jonnyscholes

thibaudcolas:
Hey @jonnyscholes 👋 @helenb and I will be working on menu-related accessibility improvements over the next few days, #5337 and #5336. I’d really like to get this refactoring in before we make any more changes to the menu JS – do you think you’d have a chance to address what is needed for me or @lb- to merge this?

The things I’m aware of are:

- [ ] Rebase on latest master
- [ ] IE11 support
- [ ] Potential folder structure changes

jonnyscholes:
I should be able to find some time later today to look at this. Shouldn't
take long to do those changes 😁

client/src/components/Submenu/index.js
jonnyscholes: I've used `event.key` here, annoyingly I have to check against 2 strings  - but it's still more readable :)

client/src/components/Submenu/index.js
jonnyscholes: This should now be resolved. I've not gone for a polyfill and instead cast to an Array as @thibaudcolas advised :)

jonnyscholes:
@thibaudcolas I've resolved all the issues discussed in this thread plus another that I picked up and done another round of testing. Confirm it now works as expected across the following:

- Chrome MacOS
- Firefox MacOS
- Safari MacOS
- Edge Windows 10
- IE11 Windows 10

@thibaudcolas and @helenb, I probably won't have any further time today to look at this, but if theres anything else you want me to look at ping me and I'll look tonight/tomorrow morning. Feel free to just make changes you need to too if that's going to unblock either of you :)

thibaudcolas:
Thanks @jonnyscholes! I’ll most likely pick this up either tomorrow or on thursday, do one last review + test & merge then.

thibaudcolas:
Thanks @jonnyscholes, merging now 🙂 

I have tested this in Chrome, Firefox, Safari on macOS + Safari iOS & Android Chrome, IE11, MS Edge.

PR #5256, opened 2019-04-24
Move and refactor upgrade notification JS

jonnyscholes:
Part of #5252. This PR moves the upgrade notification JS to wagtail-client. I've re-written it for the most part, given it's simplicity including, removing its jQuery dependance.

Probably look at #5254 first because it raises some related questions. 

Do the tests still pass? (http://docs.wagtail.io/en/latest/contributing/developing.html#testing) Yes.
Does the code comply with the style guide? (Run make lint from the Wagtail root) Yes.
For Python changes: Have you added tests to cover the new/fixed behaviour? N/A
For front-end changes: Did you test on all of Wagtail’s supported browsers?
Chrome 73 MacOS
Safari 12.0 MacOS
Firefox 66.0 MacOS
For new features: Has the documentation been updated accordingly? N/A


client/src/components/UpgradeNotification/index.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  /*
+  *   Expected JSON payload:
+  *   {
+  *       "version" : "1.2.3",    // Version number. Can only contain numbers and decimal point.
+  *       "url" : "https://wagtail.io"  // Absolute URL to page/file containing release notes or actual package. It's up to you.
+  *   }
+  */
+  const releasesUrl = 'https://releases.wagtail.io/latest.txt';
```
lb-: Just a thought - would there be any benefit in storing the releaseUrl on the container also?

benefits - can be customisable by devs without needing to load different JS (edge case probably), means the function itself contains less data hard coded (eventually the URL could be pulled out further into a tag or constant elsewhere.

client/src/components/UpgradeNotification/index.js
```diff
@@ -0,0 +1,40 @@
...
+  const currentVersion = container.dataset.wagtailVersion;
+
+  fetch(releasesUrl).then(response => {
```
lb-: Another IE one, just confirming we have a fetch polyfill available.

looks like we are using whatwg-fetch, but we might want to consider an upgrade (future issue probably), we are on 2.0.3

https://github.com/github/fetch/releases

client/src/components/UpgradeNotification/index.js
jonnyscholes: Yeah sure. I think this is a good idea - off the top of my head I can see how this might be useful for CodeRed style projects.

client/src/components/UpgradeNotification/index.js
jonnyscholes: Yeah we have a fetch polyfill, but like I mentioned in the other PR I think we should consider a more holistic polyfilling solution. Can continue discussion over there :)

jonnyscholes: Looking at the releases of `fetch`, 3.0 doesn't bring anything we're currently using. I think it might be wise to wait until someone requires those features before updating to reduce the surface are of this PR. Sound ok @lb- ?

lb-: @jonnyscholes  - agree 100%, that can wait until a future PR but as mentioned may not really be needed.

client/src/utils/utils.test.js
```diff
@@ -0,0 +1,18 @@
...
+    });
+    it('compares 1.5.4 and 1.5.5 correctly', () => {
+      expect(versionOutOfDate('1.5.4', '1.5.5')).toBeFalsy();
+    });
+    it('compares 1.5 and 1.5 correctly', () => {
+      expect(versionOutOfDate('1.5', '1.5')).toBeFalsy();
+    });
+    it('compares 2.6a0 and 2.4 correctly', () => {
+      expect(versionOutOfDate('2.6a0', '2.4')).toBeTruthy();
+    });
```
thibaudcolas: 👏 thank you for the tests

lb-:
This is good to go, I have merged it in with some resolved conflicts all sorted.

@jonnyscholes  - merged in via https://github.com/wagtail/wagtail/commit/c0af26b076b4561cdbdca2eb81de90a930813421

I think there were two outstanding items, not blockers though.

1. components folder housing both React and 'DOM / template style' components - maybe when more refactoring is done a good pattern will emerge and we can move things around then.
2. slightly easier future customisation of putting the releases URL in the template, so it can be accessed by its data attribute. I think that the original code did not have this so all good, plus there are other ways to override what renders in the settings bar.

Issue #5261, opened 2019-04-26
Revision history comparison breaks on streamfield after 2.5 upgrade


Adrian-Turjak:
### Issue Summary

An attempt to compare streamfield changes causes a 500 because of incorrect usage of `list.append()` getting passed in two values which I assume are meant to be a tuple.

### Steps to Reproduce

At least for me, I just had to create any page with a stream field, then edit it twice, and any attempt to compare the contents of a streamfield resulted in this error. Any streamfield that uses Structblocks will be hit by it.

### Technical details

* Python version: 3.7.3
* Django version: 2.2
* Wagtail version: 2.5

### Log:
```
app_1  | ACCESS: 10.203.0.194 - - [26/Apr/2019:03:48:53 +0000] "GET /admin/pages/120/revisions/compare/1137...1138/ HTTP/1.1" 500 27 "https://website.dev.catalystcloud.nz/admin/pages/120/revisions/" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/73.0.3683.103 Safari/537.36"
app_1  | DJANGO: ERROR 2019-04-26 03:48:53,736 /usr/local/lib/python3.7/site-packages/django/utils/log.py Internal Server Error: /admin/pages/120/revisions/compare/1137...1138/
app_1  | Traceback (most recent call last):
app_1  |   File "/usr/local/lib/python3.7/site-packages/django/core/handlers/exception.py", line 34, in inner
app_1  |     response = get_response(request)
app_1  |   File "/usr/local/lib/python3.7/site-packages/django/core/handlers/base.py", line 115, in _get_response
app_1  |     response = self.process_exception_by_middleware(e, request)
app_1  |   File "/usr/local/lib/python3.7/site-packages/django/core/handlers/base.py", line 113, in _get_response
app_1  |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
app_1  |   File "/usr/local/lib/python3.7/site-packages/django/views/decorators/cache.py", line 44, in _wrapped_view_func
app_1  |     response = view_func(request, *args, **kwargs)
app_1  |   File "/usr/local/lib/python3.7/site-packages/wagtail/admin/urls/__init__.py", line 102, in wrapper
app_1  |     return view_func(request, *args, **kwargs)
app_1  |   File "/usr/local/lib/python3.7/site-packages/wagtail/admin/decorators.py", line 34, in decorated_view
[...]

PR #5262, opened 2019-04-26
Fix streamfield structblock compare append usage

Adrian-Turjak:
Append was being called with 2 params, which clearly should
have been a tuple.

Fixes: https://github.com/wagtail/wagtail/issues/5261

### DIFF
--- a/wagtail/admin/compare.py
+++ b/wagtail/admin/compare.py
@@ -115,7 +115,7 @@ class StructBlockComparison(BlockComparison):
             label = self.block.child_blocks[name].label
             comparison_class = get_comparison_class_for_block(block)
 
-            htmlvalues.append(label, comparison_class(block, True, True, val[name], val[name]).htmlvalue(val[name]))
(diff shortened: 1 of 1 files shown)


zerolab:
A small note in that #5132 also fixes this issue in addition to other tweaks

Adrian-Turjak:
@kaedroho Do you want to add tests or should we treat that as a follow up in that other patch? I would add tests, but I don't know this code much anyway, I just fixed what appeared as an easy bug.

gasman:
Tests added in https://github.com/gasman/wagtail/commit/c5fd7f74944bb9376934be089d0d2e9526e9beb3 - will merge into master and 2.5.x once CI passes.

gasman:
Merged in 6b1930c7cf76743f5e467db9ef68d30d934144bb (stable/2.5.x) and 52fe571eb27cb68151a5c7f9dc338ff7d5def8a4 (master). Thanks @Adrian-Turjak!

PR #5266, opened 2019-04-30
Support required fields on custom image models in the multiple image uploader

### LINKED ISSUE #847: Custom Image model with additional required fields - impossible to add image in multi-add view.
It is not possible to add images in wagtail admin under images submenu, because after adding image it tries to save object, raising validation error.

gasman:
Fixes #847

If form validation fails on the multiple image uploader (due to the presence of a required field on a custom image model), we create an instance of a new model `UploadedImage` which just contains the image file field. The 'edit' sub-forms are then tweaked so that the final image object is created at the point of submitting a form with all fields filled in.


gasman:
Additional fix + tests in b9cfff6bf846ab563a8986dc4c56032cd905aeff - tags weren't being saved.

lb-:
Merged in via 4023a90d6e0b2876c560f0318c1cf6aea544e354

@gasman  - I rebased and update the release version in the docs changes, re-tested manually and checked all unit tests.

I am not sure if we should add a caveat in the final release notes that image models with required fields will now need validated data before actual saving.

gasman:
@lb- Thanks! I don't think we need a caveat in the release notes, as this would previously have failed outright - no sites would be relying on the old behaviour.

PR #1244, opened 2015-04-28
Fall back on ASCII when unicode document filenames fail - fixes #1166

gasman:

PR #5267, opened 2019-04-30
Fix non-ascii content disposition

richbrennan:
Hi @gasman, 

This is a PR for the issue with document filenames containing non-asciii characters we discussed, where a [WSGI server](https://github.com/benoitc/gunicorn/blob/19.9.0/gunicorn/http/wsgi.py#L329) might attempt to encode the header as ascii. It implements the [suggestion](https://github.com/wagtail/wagtail/pull/1244#issuecomment-97134194) to url encode the document file and add `filename*=UTF-8''` to the header.

Thanks,

Rich.

### DIFF
--- a/wagtail/documents/tests/test_views.py
+++ b/wagtail/documents/tests/test_views.py
@@ -171,7 +171,7 @@ class TestServeWithUnicodeFilename(TestCase):
         self.assertEqual(response.status_code, 200)
 
     @mock.patch('wagtail.documents.views.serve.hooks')
-    @mock.patch('wagtail.documents.views.serve.get_object_or_404') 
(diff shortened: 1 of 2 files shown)


lb-:
Tested this PR out locally, with a filename like `docsالكاتدرايية_والسوق.pdf` - working as expected.
I will merge this in with a comment added above the changed line (for future reference).

lb-:
Thanks @richbrennan ! - this has been merged in via
https://github.com/wagtail/wagtail/commit/2dbd0a49f9e73ea4d39c5c9abc1b73207a79f3c9

I had to make a few tweaks to the tests as some of that file had changed since this PR was prepared.

richbrennan:
Thanks @lb- , great to see this getting merged! 🙏 🎉

Issue #5246, opened 2019-04-23
Update front-end tooling


nmorduch:
Split from #5015, starting from https://github.com/wagtail/wagtail/issues/5015#issuecomment-482512717, where @thibaudcolas, @jonnyscholes, @lb-, and I have been discussing Prettier and stylelint rules.

nmorduch:
In response to @lb-'s comment https://github.com/wagtail/wagtail/issues/5015#issuecomment-485566213

#### Re selector ordering

I don't love alphabetizing selectors, but I'm sufficiently convinced that an enforced ordering rule is beneficial for diffs. I personally tend to prefer orderings that group "rules that affect positioning/layout/sizing" and "rules that affect presentation", though I've never been very consistent on the details of that. 

I found [stylelint-config-idiomatic-order](https://github.com/ream88/stylelint-config-idiomatic-order) via the example config for prettier-stylelint, and I haven't tested it but it looks approximately like what I would like. You can see its specific ordering in its [index.js](https://github.com/ream88/stylelint-config-idiomatic-order/blob/master/index.js#L81). It looks like it outputs a config for stylelint-order that puts positioning, display, and box model first, then everything else is alphabetized after that. Any thoughts on how well that would work for us?

#### A thought on diffs for this change

This is probably already the plan, but I'd like to ask that, in the PR to add these rules, commits/diffs are set up in such a way that it's easy to distinguish "changes made by prettier-stylelint" and "changes made by humans". This probably mostly means putting any manual fixing of things that stylelint doesn't like into a separate commit from the enormous collection of automatic changes that prettier-stylelint will make the first time it is run.

jonnyscholes:
Yeah I prefer idiomatic ordering over alphabetical too so would opt for that option if we enforce selector ordering. 

One grievance I do have with enforcing selector ordering though, is sometimes it's nice to be able to order your rules based on something specific to the design/functionality (eg themes or a complex CSS animation). For example, of late one thing I've been doing in parts of my code that have a lot of CSS transitions/animations is grouping all of the styles that are required only for the transition and are changed as part of it at the end of the rule. I've found this makes "designing" the animations much easier and ever so slightly closer to a visual timeline that one would use in something like After Effects (you can collapse all but the relevant code in your editor so you're only looking at "animation code").

[...]

PR #5281, opened 2019-05-05
Use external stylelint config

lb-:
Partial work on #5246 

* implements the usage of an external package `@wagtail/stylelint-config-wagtail`
* This config is currently a 1:1 match but does use the latest stylelint packages
* Updated the `npm run lint` script to also check scss
* Some docs refinements regarding code conventions, js one seemed to be quite out of date also

Question: - Are we happy to have this package as a scoped package or should it just be `stylelint-config-wagtail`?

Question: npm install seemed to change the package lock file drastically, I was running node 8 locally (node 10 does not appear to work with gulp), just wanted to check if this is an issue or not.


lb-:
@thibaudcolas & @nmorduch  - would love your review on this.

I have two questions on the main description, just wanted to check I have done everything right.

All of this should get us one step closer to a great set up for future devs, I will chat on the main ticket about the next steps for what rules we want to implement/change.

package.json
```diff
@@ -80,8 +81,8 @@
     "react-axe": "^3.1.0",
     "react-test-renderer": "^16.2.0",
     "redux-mock-store": "^1.3.0",
-    "stylelint": "^8.4.0",
-    "stylelint-scss": "^2.2.0",
+    "stylelint": "^10.0.1",
+    "stylelint-scss": "^3.6.1",
```
thibaudcolas: `stylelint-scss` is a dependency of `stylelint-config-wagtail`, so can be removed from here.

package.json
```diff
@@ -110,7 +111,7 @@
     "webpack:prod:build": "webpack --config ./client/webpack/prod.config.js",
     "lint:js": "eslint --max-warnings 16 ./client",
     "lint:css": "stylelint **/*.scss",
-    "lint": "npm run lint:js",
+    "lint": "npm run lint:js && npm run lint:css",
```
thibaudcolas: 👍 nice, I hadn't realised this wasn't already the case.

docs/contributing/javascript_guidelines.rst
```diff
@@ -47,10 +47,9 @@ you need to make to the codebase.
...
 
-Under the hood, the tasks use the `JavaScript Code Style <http://jscs.info/>`_ library.
+To edit ignored files, add an entry to the ``.eslintignore`` file.
 
-To edit the settings for ignored files, or to change the linting rules,
-edit the ``.jscsrc`` file in the wagtail project root.
-
-A complete list of the possible linting rules can be found here:
-`JSCS Rules <http://jscs.info/rules.html>`_
+To the configuration for the linting rules is managed in an external
```
thibaudcolas: Typo: `The configuration`.

docs/contributing/javascript_guidelines.rst
```diff
@@ -47,10 +47,9 @@ you need to make to the codebase.
 Changing the linter configuration
 ---------------------------------
 
-Under the hood, the tasks use the `JavaScript Code Style <http://jscs.info/>`_ library.
+To edit ignored files, add an entry to the ``.eslintignore`` file.
```
thibaudcolas: I know this was there before and you're just updating the docs (thank you!), but I don't think this is worth documenting.

docs/contributing/javascript_guidelines.rst
```diff
@@ -13,7 +13,7 @@ Write JavaScript according to the `Airbnb Styleguide <http://github.com/airbnb/j
 Linting and formatting code
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Wagtail provides some tooling configuration to help check your code meets the
+Wagtail uses the `eslint <https://eslint.org/>`_ linter to help check your code meets the
```
thibaudcolas: Should be "ESLint".

docs/contributing/css_guidelines.rst
```diff
@@ -49,12 +52,27 @@ Run the linter from the wagtail project root:
...
 .. code-block:: console
 
     $ npm run lint:css -- --fix
 
+Changing the linter configuration
+---------------------------------
+
+To edit ignored files, add an entry to the ``.stylelintrc.yaml`` file under ``ignoreFiles``.
+
+To the configuration for the linting rules is managed in an external
```
thibaudcolas: Typo: `The configuration`.

docs/contributing/css_guidelines.rst
```diff
@@ -49,12 +52,27 @@ Run the linter from the wagtail project root:
...
 If you want to autofix errors, you can run that command directly with:
 
 .. code-block:: console
 
     $ npm run lint:css -- --fix
 
+Changing the linter configuration
+---------------------------------
+
+To edit ignored files, add an entry to the ``.stylelintrc.yaml`` file under ``ignoreFiles``.
```
thibaudcolas: I know this has been done for JS linting, but I don't think this is worth documenting. We don't want people ignoring files via config (you can do more granularly with disable comments).

.stylelintrc.yaml
```diff
@@ -3,77 +3,5 @@ ignoreFiles:
...
-  scss/media-feature-value-dollar-variable: always
-  scss/selector-no-redundant-nesting-selector: true
-  string-no-newline: true
-  string-quotes: single
-  unit-no-unknown: true
-  unit-case: lower
-  value-no-vendor-prefix: true
-  property-no-vendor-prefix: true
+extends:
+  - '@wagtail/stylelint-config-wagtail'
```
thibaudcolas: 🎉

package.json
```diff
@@ -51,6 +51,7 @@
     ]
   },
   "devDependencies": {
+    "@wagtail/stylelint-config-wagtail": "0.0.4",
```
thibaudcolas: Two things here:

- This should be using a caret (`^0.0.4`) like other dependencies, we use a lockfile so there is no need to pin exact versions.
- This is a really minor thing but I would expect this to be versioned with a minor version number, `0.1.0`, `0.1.4`, or `0.4.0`, not with the patch number alone. This is because patches should only be used to indicate "patch"-type changes to existing minor or major releases, and 0.0.0 isn't a valid release number. SemVer guidance is to start at 0.1.0: https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase. I don't think this has any incidence in practice though.

lb-:
@thibaudcolas  - let me know if you need me to make any more changes.

PR #5285, opened 2019-05-09
Add type to chunk_size arg for update_index

seb-b:
This allows `chunk_size` to be used from the command line without having to use `management.call_command('update_index', chunk_size=x)`

A type error was thrown before this, args are strings by default.


gasman:
Thanks @seb-b! I can only assume that this behaviour must have changed in a recent Django or Python version...

Merged in 5e15984074f78d29031f2832b6918993dff4e7d7.

PR #5286, opened 2019-05-09
Output form media on add/edit image forms

### LINKED ISSUE #4466: RichTextField not visible when used as a field on custom Image model
### Issue Summary

RichTextField fields are not visible in the CMS when used as a field on custom Image model.

### Steps to Reproduce

I set up a custom image model as described [here](http://docs.wagtail.io/en/v2.0.1/advanced_topics/images/custom_image_model.html). I added a `wagtail.core.fields.RichTextField` as one of the fields, and added that field name to `admin_form_fields`. I expected the field to appear at /cms/images/{pk} but it does not. Other standard-issue Django fields added in the same way do appear in the CMS.

### Technical details

* Python version: Python 3.6.4
* Django version: Django 2.0.4
* Wagtail version: Wagtail 2.0.1
* Browser version: Safari 11.1; Chrome 65.0.3325.181

gasman:
Form media was not being output on the add/edit image forms, so any fields on custom image models which rely on additional JS/CSS (such as rich text and date fields) would fail.

Note that this isn't 100% reliable for the 'upload image' form in the image chooser modal in the case that the media is served from an external domain such as a CDN - the content of the modal is added to the page via a jQuery DOM insertion, and in this situation any script tags in that content are loaded asynchronously and not guaranteed to complete loading before inline scripts are run. (It's better than not having the includes there at all though...)

Fixes #4466 (with the proviso that certain rich text features such as links/embeds will need to be disabled on the field, as they depend on other JS which is only present on page/snippet edit views, or in the case of the image chooser popup, would spawn a modal within a modal)


gasman:
> as they depend on other JS which is only present on page/snippet edit views

More precisely, both rich text fields and AdminChooser widgets depend on `window.chooserUrls` being defined, which currently happens in `_editor_js.html`. Making them into self-sufficient widgets (with JS/CSS dependencies fully captured by their media defintions) would be a very good thing, but that's a separate issue, namely #2936...

gasman:
Rebased and fixed against current master (adding the prefix in #5298 required a test case to be updated)

johnraz:
@gasman , @thibaudcolas is anything blocking this PR outside of the time required to get it reviewed?

Also, I think adding a section to the doc explaining the limitation would be nice too - I can help with the doc part if needed.

gasman:
@johnraz It's just waiting for review, as far as I'm aware. Definitely happy to consider any additions to the docs - please go ahead!

lb-:
Merged in via https://github.com/wagtail/wagtail/commit/eaad01308142337739d7edd9f08bd8e714be23f0

thanks @gasman

zerolab:
For anyone needing links in a rich text field you can do this by creating `wagtailimages/images/edit.html` in your templates directory with

```django+html
{% extends "wagtailimages/images/edit.html" %}
{% load wagtailadmin_tags %}

{% block extra_js %}
    {{ block.super }}
    {% allow_unicode_slugs as unicode_slugs_enabled %}
<script>
    window.chooserUrls = {
        'pageChooser': '{% url "wagtailadmin_choose_page" %}',
        'externalLinkChooser': '{% url "wagtailadmin_choose_page_external_link" %}',
        'emailLinkChooser': '{% url "wagtailadmin_choose_page_email_link" %}',
        'phoneLinkChooser': '{% url "wagtailadmin_choose_page_phone_link" %}',
        'anchorLinkChooser': '{% url "wagtailadmin_choose_page_anchor_link" %}',
    };
    window.unicodeSlugsEnabled = {% if unicode_slugs_enabled %}true{% else %}false{% endif %};
</script>
<script src="{% versioned_static 'wagtailadmin/js/modal-workflow.js' %}"></script>
{% endblock %}
```

ditto for `wagtailimages/images/add.html` and `wagtailimages/multiple/add.html` with the corresponding path changed in `{% extends ... %}`

urlsangel:
If anyone else is looking for an answer to this question, I’ve fixed it by implementing the fix from here:

- https://github.com/wagtail/wagtail/pull/11754
- https://github.com/wagtail/wagtail/pull/11803/files#diff-4ee7f72163f125b407b463a6a0318dfed050983223c9af385be2f8fa631bea28

Essentially, the old implementation should be replaced with this:

```
{% block extra_js %}
    {{ block.super }}
    {{ media.js }}
    {% include "wagtailadmin/pages/_editor_js.html" %}
{% endblock %}
```

PR #5288, opened 2019-05-09
Move date / time widget initialiser JS into the widget's form media

gasman:
This allows them to work on pages that do not include _editor_js.html (or page-editor.js). ~However, wagtailadmin/shared/datetimepicker_translations.html will still be required in order to localise the date picker UI.~

(They also still depend on jquery-ui-1.12.1.min.js / jquery.datetimepicker.js, which are in the global JS.)

Another step towards #2936.


thibaudcolas:
If that's valuable, I think we might be able to get rid of `datetimepicker_translations.html`:

- Move the translated strings with other translations for client-side components in `admin_base.html`, https://github.com/wagtail/wagtail/blob/e1c2410734184cbddc8fafcf821fb637748b6e76/wagtail/admin/templates/wagtailadmin/admin_base.html#L30-L34.
- Move the datetimepicker's `wagtail_custom_locale` magic into this new `date-time-chooser.js`, reusing the labels now defined in `wagtailConfig.STRINGS`.

gasman:
@thibaudcolas Oh, nice! I was debating with myself whether we could justify moving the datetimepicker_translations.html include to admin_base.html, but given that there are a bunch of translations there already, it's really a no-brainer :-)

gasman:
Translations now moved to admin_base in 3bb25a0ea20c528137f9a0dfa587743fb66eb60c

lb-:
Tested locally, datetime fields work as expected, merged in https://github.com/wagtail/wagtail/commit/4cb308bf66ff5bbc65480058d6b63af05330c761

PR #5292, opened 2019-05-10
Add alt attributes to images that was missing them

### LINKED ISSUE #5291: Add alt attribute to all images in the CMS admin
Screen reader users cannot view images. Images need to either have a textual description, or an empty alt attribute (`alt=""`) for screen readers to ignore them – otherwise the image’s source URL gets read out loud.

Here are 4 occurences I know about, there might be more:

```
<img src="//www.gravatar.com/avatar/e64c7d89f26bd1972efa854d13d7dd61?s=100&amp;d=mm">, at #account-settings > span > img

Solution: add empty alt=””

<img>, at #id_image-chooser > div:nth-child(1) > div > img 

Solution: add empty alt=””

<img src="/media/images/vik.max-800x600_tWRrsIy.jpg" style="[…]"> at #wagtail > div:nth-child(19) > div:nth-child(2) > div > form > div:nth-child(2) > div:nth-child(2) > div:nth-child(1) > div:nth-child(2) > div:nth-child(1) > div:nth-child(1) > img

Solution: add empty alt=””
```

rinti:
Fixes #5291 - found a couple of more places where images had no alt attribute.

On a side note not related to accessibility i found that the `` tags are inconsistent in when they're `` or ``. If it's worth changing to be more consistent I'm willing to create an issue and PR.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -20,6 +20,9 @@ Changelog
  * Fix: Admin HTML now includes the correct `dir` attribute for the active language (Andreas Bernacca)
  * Fix: Fix type error when using `--chunk_size` argument on `./manage.py update_index` (Seb Brown)
  * Fix: Avoid rendering entire form in EditHandler's `repr` method (Alex Tomkins)
+ * Fix: Add empty alt attributes to HTML output of Embedly and oEmbed embed finders (Andreas Bernacca)
+ * Fix: Add empty alt attributes to all images in the CMS admin (Andreas Bernacca)
+ * Fix: Make URL generator preview image alt translateable (Thibaud Colas)
 
 
 2.5.1 (07.05.2019)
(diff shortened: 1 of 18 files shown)


thibaudcolas:
Thank you @rinti! This is going to make a big difference for screen reader users. I'll try to get this reviewed & ideally merged this week. Did you cover all of the instances of `` you found or would there be more to look at?

For `` vs ``, yes, it would be nice if these were consistent, but this has no implications in practice. If you're willing to add a commit to this PR to switch every instance to `` without the slash, I'll gladly merge it. I don't think it's worth a separate PR.

rinti:
@thibaudcolas Yes, I just did a ripgrep after '` where i _think_ attrs already contains the alt-tag. But I would need to double check this.

thibaudcolas:
👍 thanks. I just checked, if it's https://github.com/wagtail/wagtail/blob/74218a7e43f59b329cd7cd890245bcc2764beb3a/wagtail/images/templates/wagtailimages/images/edit.html#L67

this does have an `alt` coming from `rendition.attrs`.

rinti:
Yep, that's the one!

rinti:
@thibaudcolas Found a couple of other places in python code with img without alt-tags. I will fix those too in a bit.

thibaudcolas:
👌 merged! Thank you @rinti for taking care of this 🙂 and @auvipy @hippogriffic who helped with reviewing.

PR #5296, opened 2019-05-13
Implement WAGTAILDOCS_SERVE_METHOD setting

gasman:
Implements the main part of #1420, based on earlier work from @tobiasmcnulty in #3772. Document serving can now be configured to:

1. use the underlying storage's URL directly (`WAGTAILDOCS_SERVE_METHOD = 'direct'`)
2. Have the Django view serve the document (the current behaviour, `WAGTAILDOCS_SERVE_METHOD = 'serve_view'`)
3. Have the Django view redirect to the storage's URL after running permission checks and other hooks/signals (`WAGTAILDOCS_SERVE_METHOD = 'redirect'`)

with 3 becoming the new default behaviour for remote backends such as S3. This is probably the desired behaviour for most people using those backends (so that documents are served to the user from that remote storage, rather than going through Django) but is a potentially breaking change and will need to be noted in release notes.

### DIFF
--- a/docs/advanced_topics/settings.rst
+++ b/docs/advanced_topics/settings.rst
@@ -268,6 +268,24 @@ This setting lets you override the maximum number of pixels an image can have. I
 This setting enables feature detection once OpenCV is installed, see all details on the :ref:`image_feature_detection` documentation.
 
 
+Documents
+---------
+
+.. code-block:: python
+
+  WAGTAILDOCS_SERVE_METHOD = 'redirect'
+
+Determines how document downloads will be served. Normally, requests for documents are sent through a Django view, to perform permission checks (see :ref:`image_document_permissions`) and potentially other housekeeping tasks such as hit counting. To fully protect against users bypassing this check, it needs to happen in the same request where the document is served; however, this incurs a performance hit as the document then needs to be served by the Django server. In particular, this cancels out much of the benefit of hosting documents on external storage, such as S3 or a CDN.
+
+For this reason, Wagtail provides a number of serving methods which trade some of the strictness of the permission check for performance:
+
+ * ``'direct'`` - links to documents point directly to the URL provided by the underlying storage, bypassing the permission check. This is most useful when deploying sites as fully static HTML (e.g. using `wagtail-bakery <https://github.com/wagtail/wagtail-bakery>`_ or `Gatsby <https://www.gatsbyjs.org/>`_).
+ * ``'redirect'`` - links to documents point to a Django view which will check the user's permission; if successful, it will redirect to the URL provided by the underlying storage to allow the document to be downloaded. This is most suitable for remote storage backends such as S3, as it allows the document to be served independently of the Django server. Note that if a user is able to guess the latter URL, they will be able to bypass the permission check; some storage backends may provide configuration options to generate a random or short-lived URL to mitigate this.
+ * ``'serve_view'`` - links to documents point to a Django view which both checks the user's permission, and serves the document. Serving will be handled by `django-sendfile <https://github.com/johnsensible/django-sendfile>`_, if this is installed and supported by your server configuration, or as a streaming response from Django if not. When using this method, it is recommended that you configure your webserver to *disallow* serving documents directly from their location under ``MEDIA_ROOT``, as this would provide a way to bypass the permission check.
+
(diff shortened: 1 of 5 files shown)


docs/advanced_topics/settings.rst
```diff
@@ -268,6 +268,24 @@ This setting lets you override the maximum number of pixels an image can have. I
 This setting enables feature detection once OpenCV is installed, see all details on the :ref:`image_feature_detection` documentation.
 
 
+Documents
+---------
+
+.. code-block:: python
```
chosak: Can we add a link from the [`before_serve_document`](https://docs.wagtail.io/en/v2.6.1/reference/hooks.html#document-serving) hook documentation, warning that this won't be called if `WAGTAILDOCS_SERVE_METHOD = 'direct'`?

docs/advanced_topics/settings.rst
```diff
@@ -268,6 +268,24 @@ This setting lets you override the maximum number of pixels an image can have. I
...
+
+  WAGTAILDOCS_SERVE_METHOD = 'redirect'
+
+Determines how document downloads will be served. Normally, requests for documents are sent through a Django view, to perform permission checks (see :ref:`image_document_permissions`) and potentially other housekeeping tasks such as hit counting. To fully protect against users bypassing this check, it needs to happen in the same request where the document is served; however, this incurs a performance hit as the document then needs to be served by the Django server. In particular, this cancels out much of the benefit of hosting documents on external storage, such as S3 or a CDN.
```
chosak: Regarding the use of the phrase "will be served", this change alters not just the serving but also the linking. Could this say "will be linked to and served"?

More specifically, if someone chooses `'direct'`, it might not be super obvious that the _/documents/pk/filename_
URL pattern no longer appears across the site (even though, correctly, it still works, supporting, say, manual uses of `reverse` in other Django code). Specifically I think this means links both in rich text and in the documents admin views (list and edit) will point directly to the underlying storage.

wagtail/documents/tests/test_views.py
```diff
@@ -67,6 +68,64 @@ def clear_sendfile_cache(self):
...
+    """
+    def setUp(self):
+        self.document = models.Document(title="Test document")
+        self.document.file.save('example.doc', ContentFile("A boring example document"))
+        self.serve_view_url = reverse('wagtaildocs_serve', args=(self.document.id, self.document.filename))
+
+    def tearDown(self):
+        self.document.delete()
+
+    def test_url(self):
```
chosak:
```suggestion
    def test_document_url_should_point_to_serve_view(self):
```

Maybe use more descriptive test names for these, instead of a comment?

Also, it'd be nice if this test were duplicated above as part of `TestServeViewWithRedirect` to make it clear that this is the expected behavior of `Document.url` when using `'redirect'`.

wagtail/documents/tests/test_views.py
```diff
@@ -67,6 +68,64 @@ def clear_sendfile_cache(self):
...
+@override_settings(WAGTAILDOCS_SERVE_METHOD='direct')
+class TestDirectDocumentUrls(TestCase):
+    def setUp(self):
+        self.document = models.Document(title="Test document")
+        self.document.file.save('example.doc', ContentFile("A boring example document"))
+
+    def tearDown(self):
+        self.document.delete()
+
+    def test_url(self):
```
chosak:
```suggestion
    def test_url_should_point_directly_to_file_storage_url(self):
```

wagtail/documents/tests/test_views.py
```diff
@@ -67,6 +68,64 @@ def clear_sendfile_cache(self):
...
+        self.assertEqual(self.document.url, self.document.file.url)
+
```
chosak: Worth adding a test to confirm that the redirect behavior still works if someone manually uses the serve view URL?

docs/advanced_topics/settings.rst
gasman: Good shout! Added in d09445eedffab911acc40e2d9567ffa7963b594e.

docs/advanced_topics/settings.rst
gasman: Updated in d09445eedffab911acc40e2d9567ffa7963b594e - I've slightly reworded the description of `direct` to make it clearer that we're no longer linking to the Django view, but open to other suggestions here.

chosak: Looks great, thanks.

gasman:
Completed and merged in e87dcaf01590e629a306f24abdf770679c0b0ee6 + parents.

PR #5298, opened 2019-05-15
Add form field prefixes for input forms in chooser modals

gasman:
Add distinctive namespace prefixes to the input forms inside link, document, image and embed chooser modals, so that fields with common names like 'title' and 'tags' don't end up with IDs that collide with other forms on the page (resulting in invalid HTML and broken JS widgets).


lb-:
Merged in https://github.com/wagtail/wagtail/commit/c109dd3495ed67ef8a8f238f6ba7f163fbec6138

I did a bunch of random searches / checks of various forms and all appears normal, plus ran gulp build to check for any errors in the final JS.

Thanks @gasman

Issue #5191, opened 2019-03-29
Sortable headings on redirect / user / site index pages don't work


gasman:
On listing pages in the admin that have clickable column headings, the expected behaviour is that clicking once should order on that column, and clicking again should reverse the order. This doesn't happen on the following pages:

* In Settings -> Users, clicking once to order by name or username works (although this behaviour can be unclear if user accounts don't have the real name filled in, in which case it'll show the username but sort on the empty strings...) - clicking on the same heading again causes it to always revert to sorting forwards by name.
* ~In Settings -> Sites, the only sortable column is the site name and it cannot be reversed, so clicking never has any effect. Also, the heading is not highlighted in teal as it should be.~ Fixed in #7409
* In Settings -> Redirects, the only sortable column is 'From' and it cannot be reversed, so clicking never has any effect.

lb-:
@gasman  - I have a POC ready ([link](https://github.com/wagtail/wagtail/compare/master...lb-:feature/5191-sortable-headings?expand=1)) but I have a few questions:

1. Ordering of sites does not seem to work at all, its view extends `generic.IndexView` and I cannot see that ordering is available from that class and its parents. should I add the ability to order on this view or am I missing something? https://github.com/wagtail/wagtail/blob/master/wagtail/sites/views.py#L10

These others are more about consistency, I can prepare a PR to fix the issues individually quickly but the below items may take longer to address/answer.

2. should we make the title AND the chevron the link or just the chevron (it is currently inconsistent across views from what I can see), this kind of works with the 'teal' as most teal things are clickable.
3. If we are sticking to the chevron only, do we need to add a title to the link as this is not great for accessibility (maybe a separate ticket needs to be raised)
4. Is there a generic mixin or something that will make this easier? I know wagtail's modeladmin has one but not sure if there is something in Django class based views (I could not find one). Should this be investigated for this ticket?
5. Is it important to whitelist all field ordering before passing it into the query? some places do it, others do not. There will be a 500 error if someone malforms the URL manually, the benefit of not white listing is that it is easier for others to add their own ordered links.

gasman:
1 - Yep, this might well have been missed when we introduced the generic view - if the fix requires adding support for sortable listings in generic.IndexView, we should go ahead and do that.

2/3 - already fixed in #5190.

[...]

PR #5299, opened 2019-05-15
Feature - Make Promoted Search Results Orderable

### LINKED ISSUE #3897: Make Promoted Search Results Orderable

lb-:
Extends the work done on #4614 
Closes #3897

Thanks @chrisranjana - really awesome stuff.

Just adds some tests, a bit of formatting in the Django template files and a bit of a reorganisation of how the ordering url args are handled in the view.

* Do the tests still pass? ✅ 
* Does the code comply with the style guide?  ✅ 
* For front-end changes: N/A
* For new features: Has the documentation been updated accordingly? N/A - pretty much expected behaviour and other similar pages are not documented in this way.

### DIFF
--- a/wagtail/contrib/search_promotions/tests.py
+++ b/wagtail/contrib/search_promotions/tests.py
@@ -154,6 +154,70 @@ class TestSearchPromotionsIndexView(TestCase, WagtailTestUtils):
         # "aargh snake" should be the first result alphabetically
         self.assertEqual(response.context['queries'][0].query_string, "aaargh snake")
 
+    def test_results_ordering(self):
+        self.make_search_picks()
+        url = reverse('wagtailsearchpromotions:index')
+        SearchPromotion.objects.create(
+            query=Query.get("zyzzyvas"),
+            page_id=1,
+            sort_order=0,
+            description="no definition found, this is a valid scrabble word though",
+        )
+
+        SearchPromotion.objects.create(
+            query=Query.get("aardwolf"),
+            page_id=1,
+            sort_order=22,
+            description="Striped hyena of southeast Africa that feeds chiefly on insects",
+        )
+
+        popularQuery = Query.get("optimal")
+        for i in range(50):
+            popularQuery.add_hit()
+        SearchPromotion.objects.create(
+            query=popularQuery,
+            page_id=1,
+            sort_order=15,
+            description="An odly popular search term?",
+        )
+
+        popularQuery = Query.get("suboptimal")
+        for i in range(25):
+            popularQuery.add_hit()
+        SearchPromotion.objects.create(
+            query=popularQuery,
+            page_id=1,
+            sort_order=15,
+            description="Not as popular",
+        )
+
+        # ordered by querystring (reversed)
+        response = self.client.get(url + '?ordering=-query_string')
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.context['queries'][0].query_string, "zyzzyvas")
+
+        # last page, still ordered by query string (reversed)
+        response = self.client.get(url + '?ordering=-query_string&p=3')
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.context['queries'][-1].query_string, "aardwolf")
+
+        # ordered by querystring (not reversed)
+        response = self.client.get(url + '?ordering=query_string')
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.context['queries'][0].query_string, "aardwolf")
+
+        # ordered by sum of daily hits (reversed)
+        response = self.client.get(url + '?ordering=-views')
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.context['queries'][0].query_string, "optimal")
+        self.assertEqual(response.context['queries'][1].query_string, "suboptimal")
+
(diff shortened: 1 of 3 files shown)


wagtail/contrib/search_promotions/views.py
```diff
@@ -19,16 +20,30 @@
 )
 @vary_on_headers('X-Requested-With')
 def index(request):
-    is_searching = False
-    query_string = request.GET.get('q', "")
+    # Ordering
+    valid_ordering = ['query_string', '-query_string', 'ohits', '-ohits']
```
chosak: At first read the meaning of the query string parameter "ohits" isn't really clear. Since under the hood this is an annotated value that doesn't have to line up to an existing model field, it might be nicer if the query string parameter matched what is displayed in the column header, maybe "views".

wagtail/contrib/search_promotions/views.py
```diff
@@ -19,16 +20,30 @@
...
+    queries = Query.objects.filter(editors_picks__isnull=False).distinct()
 
-    queries = Query.objects.filter(editors_picks__isnull=False).distinct().order_by('query_string')
+    if 'ohits' in ordering:
+        queries = queries.annotate(ohits=functions.Coalesce(Sum('daily_hits__hits'), 0))
+
+    queries = queries.order_by(ordering)
 
     # Search
+    is_searching = False
```
chosak: The search behavior doesn't work quite as expected for me, because the AJAX call loses the ordering once you start typing.

The search behavior is a little inconsistent, although this appears to be common in other parts of Wagtail (for example the user search). I didn't see this behavior mentioned on #5191.

If you search first and then order, you lose the search term:

1. Search first, URL is http://localhost:8000/admin/searchpicks/?q=b.
2. Click on ordering, link goes to http://localhost:8000/admin/searchpicks/?ordering=ohits, losing the search.

If you order first and then search, you lose the order:

1. Order first, URL is http://localhost:8000/admin/searchpicks/?ordering=ohits.
2. Start typing a search, AJAX request is made to http://localhost:8000/admin/searchpicks/?q=b (without the ordering param), so you lose the order.

To me the expected behavior is that searches should definitely persist across ordering. I'm not sure about the other way, whether ordering should persist if you do a new search, but I think maybe it should, as the user has indicated a preferred ordering (in this case, a common one would probably be to order by most common search term).

This doesn't need to hold up this PR, but I'm curious what people think is the best way to handle this.

wagtail/contrib/search_promotions/views.py
lb-: @chosak  - thanks for the feedback, I agree that ohits was a bit confusing. I have renamed to `views`

wagtail/contrib/search_promotions/views.py
lb-: @chosak  - thanks for that, I was aware of this issue and intentionally have not fixed it in this PR. I assumed, wrongly, an issue had already been created for this (I thought I had created one a year or so ago). I searched but could not find it.

I have now created a new issue that contains your notes and a bit more.

https://github.com/wagtail/wagtail/issues/5529

wagtail/contrib/search_promotions/tests.py
```diff
@@ -154,6 +154,70 @@ def test_results_are_ordered_alphabetically(self):
...
+        for i in range(50):
+            popularQuery.add_hit()
+        SearchPromotion.objects.create(
+            query=popularQuery,
+            page_id=1,
+            sort_order=15,
+            description="An odly popular search term?",
+        )
+
+        popularQuery = Query.get("sub-optimal")
```
chosak: Looks like this needs to be changed to match what the tests expect (or vice-versa):

```suggestion
        popularQuery = Query.get("suboptimal")
```

gasman: Sounds possibly related to #5416 / #5418 - perhaps this test was written before that change was committed?

lb-:
I'll take a look - I might need to merge in the latest from master.

lb-:
@chosak  - can you take a look at this and approve if you think this is good to go, I am happy to merge this in but just wanted to make sure this gets another check through

lb-:
@chosak thanks for the suggestion, I have now added this (although I used `title` as this seemed more appropriate and useful), although this made me start down rabbit hole of accessibility support for `title` vs `aria-label`. It appears that this needs a fair bit of work throughout Wagtail (I know there are other issues for this).

Plus the way we are doing icons, as you probably know, seems to cause some issues.

happy to change back to aria-label if you think that is better, or use both (again there are various potential issues with that).

Note: I have also squashed, rebased and put all into one commit

lb-:
@chosak did a bit more digging and testing, revised to use `aria-label` as per your original suggestion.

lb-:
Merged in via 1f421e35f00b5463844171047bd099259c11e95b

Thanks again @chrisranjana  for your original work on this

PR #5305, opened 2019-05-17
Update squash user details to be more portable

thibaudcolas:
- Changes the user's email from `@myproject.com` to `@example.com`. myproject.com points to a real website, it feels better to use a domain that's definitely meant as a placeholder.
- Changes the password to `changeme`, matching the credentials used for https://github.com/wagtail/bakerydemo.

### DIFF
--- a/.squash.yml
+++ b/.squash.yml
@@ -16,7 +16,7 @@ deployments:
       - wagtail start mysite
       - cd /myproject/mysite
       - python manage.py migrate
-      - echo "from django.contrib.auth import get_user_model; User = get_user_model(); User.objects.create_superuser('admin', 'admin@myproject.com', 'password')" | python manage.py shell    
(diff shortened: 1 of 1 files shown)


tomdyson:
Thanks @thibaudcolas! I think we should document the username and password somewhere more obvious than `.squash.yml` - ideally in the Squash PR message? But this is definitely an improvement.

thibaudcolas:
👌 there are docs to customise the PR messages but they don't cover adding arbitrary text: https://www.squash.io/docs/customizing-pr-comments/.

Perhaps we could find a way to add the login details directly on the deployed site, on the homepage or login page?

tomdyson:
> Perhaps we could find a way to add the login details directly on the deployed site, on the homepage or login page?

+1. The login page would be ideal but the homepage will be easier to edit.

PR #5311, opened 2019-05-20
Optimization of UserPagePermissionsProxy.revisions_for_moderation

haplo:
`UserPagePermissionsProxy.revisions_for_moderation()` was iterating over `Page` instances only to use their `path` attribute. The optimization uses `values_list()` over the `GroupPagePermission` queryset to retrieve only the page paths without having to create the model instances in memory. This saves over 50% of the runtime, in my test case it brought the time for an user with 600 pages down from 13 to 6 seconds.

Profiling before:
revisions_for_moderation_before

Profiling after:
revisions_for_moderation_after

The other big performance hit of this method is the huge filter with hundreds of OR operators. I will try to find a solution for it, but it seems harder to tackle because of the requirement to filter paths with `startswith`.

All tests pass, but I haven't checked whether the project has any covering `revisions_for_moderation` specifically.


gasman:
Thanks @haplo!
It's probably worth mentioning that `UserPagePermissionsProxy` is designed to avoid the N+1 query problem when iterating over page listings - the plan is that the `self.permissions` is executed once per request (and is assumed to be a small query returning at most ~10 rows - unfortunately that isn't true in your case...), and then all subsequent permission tests (in `UserPagePermissionsProxy` or `PagePermissionTester`) will be derived from that query, using pure Python code wherever possible. That's why the original code was avoiding doing any further ORM operations on `self.permissions`...

However, that optimisation wasn't really relevant to `revisions_for_moderation`, since that isn't going to be called within a loop anyway, so I'm happy to accept this 👍

gasman:
Merged in 41f80af5db4c35004cc4c31b9a70cde8e945e18c

PR #5314, opened 2019-05-21
Optimize admin.navigation.get_explorable_root_page

haplo:
Testing the queryset in the if clause was causing the whole queryset to be retrieved and populated from DB, all to check whether it was empty or not (`__bool__` call on the queryset).

The optimization is to rely on the strict behavior of `first_common_ancestor`, which raises an exception if the queryset is empty.

Profiling before:
get_explorable_root_page_before

Profiling after:
get_explorable_root_page_after

All tests pass.


haplo:
@gasman Do you have any concerns regarding this change? Is there something I can do to help getting this merged?

gasman:
Sorry for the delay in reviewing - all looks good to me! Merged in baee1030249fa6b1e9d775993facbe1d62208eaf.

hh-project-config:
Hi haplo, How do you get the Profilings here? Thanks!

haplo:
> Hi haplo, How do you get the Profilings here? Thanks!

I use [django-profile-middleware](https://github.com/lavi06/django-profile-middleware) locally to get a profiling of a request (passing `?prof` query parameter, also disabling `DEBUG`) and then [SnakeViz](https://jiffyclub.github.io/snakeviz/) for visualization and navigating the results.

I might write a blog article about it, I will mention it here if I do.

PR #5317, opened 2019-05-22
Add global tab-through focus outline styles

thibaudcolas:
At the moment, the admin interface has different focus styles across its UI. Generally, those styles are quite subtle, and don't make for a good experience if the user is tabbing through a fair amount of the UI (for example the main menu).

This PR adds focus outline styles that are global, applied to all of the UI admin-wide. Those global styles only define an outline, so are only meant when tabbing through the UI, and are disabled for mouse users (who don't use tabbing, and for whom the heavy outlines might be more distracting than anything). Focus styles on form fields being interacted with are still relevant.

Here is what tabbing through the dashboard page looks like before this change:

focus-styles-before

And with the new styles (old screencap, I since improved the styles for the main menu):

focus-styles-after

---

For the choice of colour, @Katielocke3 and I tried to go for something that works well across the colour variations used in Wagtail. This yellow shade works quite well with the teal button, but also the green and red ones.

Here is a sample of me tabbing through the styleguide for reference:

focus-styles-tab-through

### DIFF
--- a/client/scss/components/_listing.scss
+++ b/client/scss/components/_listing.scss
@@ -422,6 +422,8 @@ ul.listing {
 }
 
 .image-choice {
+    // Force the link to be displayed as a block, so its focus outline has the right shape.
+    display: block;
     color: inherit;
     overflow-wrap: break-word;
     word-wrap: break-word;
--- a/client/scss/components/_main-nav.scss
+++ b/client/scss/components/_main-nav.scss
@@ -119,6 +119,10 @@
     .account {
         display: none;
     }
+
+    *:focus {
+        @include show-focus-outline-inside;
+    }
 }
 
 .submenu-trigger:after {
@@ -311,7 +315,6 @@ body.explorer-open {
 
         .footer-submenu {
             @include transition(max-height 0.2s ease);
-            overflow: hidden;
             max-height: 0;
         }
 
--- a/client/scss/components/_main-nav.scss
+++ b/client/scss/components/_main-nav.scss
@@ -62,7 +62,6 @@
         // making the text semi-transparent
         &:hover,
         &:focus {
-            outline: none;
             background-color: $nav-item-hover-bg;
             color: $color-white;
             text-shadow: -1px -1px 0 rgba(0, 0, 0, 0.3);
@@ -333,7 +332,6 @@ body.explorer-open {
             cursor: pointer;
 
             &:hover {
-                outline: none;
                 background-color: rgba(100, 100, 100, 0.15);
                 color: $color-white;
                 text-shadow: -1px -1px 0 rgba(0, 0, 0, 0.3);
--- a/client/scss/settings/_variables.scss
+++ b/client/scss/settings/_variables.scss
@@ -70,8 +70,13 @@ $color-button-no: $color-red-dark;
 $color-button-no-hover: darken($color-button-no, 20%);
 $color-button-warning: $color-orange-dark;
 $color-button-warning-hover: darken($color-button-warning, 20%);
+
 $color-link: $color-teal-darker;
 $color-link-hover: $color-teal-dark;
+
+// The focus outline color is defined without reusing a named color variable
+// because it shouldn’t be reused for anything else in the UI.
+$color-focus-outline: #ffbf47;
 
 $color-text-base: darken($color-white, 85);
 $color-text-input: darken($color-white, 90);
--- a/client/scss/settings/_variables.scss
+++ b/client/scss/settings/_variables.scss
@@ -97,6 +97,8 @@ $menu-width: 200px;
 $menu-width-max: 320px;
 $mobile-nav-indent: 50px;
 
+$focus-outline-width: 3px;
+
 $nav-wrapper-inner-z-index: 26;
 $draftail-editor-z-index: $nav-wrapper-inner-z-index + 1;
 
--- a/docs/contributing/developing.rst
+++ b/docs/contributing/developing.rst
@@ -174,6 +174,7 @@ IE 11 is gradually falling out of use, and specific features are unsupported in 
 
 * Rich text copy-paste in the rich text editor.
 * Sticky toolbar in the rich text editor.
+* Focus outline styles in the main menu & explorer menu.
 
 **Unsupported browsers / devices include:**
 
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -17,6 +17,7 @@ Changelog
  * Add form field prefixes for input forms in chooser modals (Matt Westcott)
  * Increase font-size across the whole admin (Beth Menzies, Katie Locke)
  * Improved text color contrast across the whole admin (Beth Menzies, Katie Locke)
+ * Added consistent focus outline styles across the whole admin (Thibaud Colas)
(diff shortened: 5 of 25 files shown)


zmustapha:
It looks nice and clear. Using Colorblinding extension, just as clear for deuteranopia and deuteranomaly.

zmustapha:
View in grayscale looks good too!

zmustapha:
The focus mode pops quite nicely in the Colorblinding extension for common colorblindness: red-green (deuteranopia/deuteranomaly), blue-yellow (tritanopia). Excellent!

thibaudcolas:
Thank you @zmustapha 🤘. This should now be ready to fully review, and hopefully merge. Compared to the previous WIP version I did some cleanup of the existing focus outlines, wrote more tests, and improved the styles for the main menu & explorer menu.

client/scss/components/_listing.scss
```diff
@@ -422,6 +422,7 @@ ul.listing {
 }
 
 .image-choice {
+    display: block;
```
thibaudcolas: Without this, these are inline links even though they are displayed as blocks, and their outline looks strange.

client/scss/components/_main-nav.scss
```diff
@@ -120,6 +119,10 @@
     .account {
         display: none;
     }
+
+    *:focus {
+        @include show-focus-outline-inside;
+    }
```
thibaudcolas: This improves the appearance of the outline in the main menu, which relies a lot on hidden overflows.

client/scss/components/_main-nav.scss
```diff
@@ -312,7 +315,6 @@ body.explorer-open {
 
         .footer-submenu {
             @include transition(max-height 0.2s ease);
-            overflow: hidden;
```
thibaudcolas: This doesn't seem to do anything.

docs/contributing/developing.rst
```diff
@@ -174,6 +174,7 @@ IE 11 is gradually falling out of use, and specific features are unsupported in
 
 * Rich text copy-paste in the rich text editor.
 * Sticky toolbar in the rich text editor.
+* Focus outline styles in the main menu & explorer menu.
```
thibaudcolas: IE11 doesn't support `outline-offset`, so there are focus outlines there – they just don't look as good.

client/scss/components/_listing.scss
zerolab: why not add this as an inline comment there? Will save future headaches

thibaudcolas: If you're troubleshooting the code for this bit of the UI I think it's quite obvious that those links are not meant to be inline, so I think the commit history would be enough. But yep, the only impact in practice is the outline so might be worth adding a comment since it's not too obvious when interacting with the UI.

client/scss/components/_main-nav.scss
nmorduch: Did you check that it also doesn't seem to do anything in IE? I feel like half the times when I think something doesn't do anything, it does something in IE.

client/scss/settings/_variables.scss
```diff
@@ -70,9 +70,14 @@ $color-button-no: $color-red-dark;
 $color-button-no-hover: darken($color-button-no, 20%);
 $color-button-warning: $color-orange-dark;
 $color-button-warning-hover: darken($color-button-warning, 20%);
+
 $color-link: $color-teal-darker;
 $color-link-hover: $color-teal-dark;
 
+// The focus outline color is defined without reusing a named color variable
+// because it shouldn’t be reused for anything else in the UI.
+$color-focus-outline: #ffbf47;
```
nmorduch: This color doesn't pass AA for contrast against white, but it's not text so I know that's not entirely relevant. Does anyone know of a way for assessing contrast of non-text colors?

thibaudcolas: Yes, SC 1.4.11 is specifically about contrast of UI elements, and defines a minimum contrast of 3:1: https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html. Unfortunately this shows no notion of the "size" of the elements like 1.4.3 has for text.

We went with this particular shade of yellow because it struck a good compromise between being visible and not being too distracting, after having tried all of the colors in Wagtail’s existing palette, plus custom blues and oranges, against all of the elements in the styleguide and on a few main pages. Our main goal was to have something that stood out against the dark teal used for primary buttons and in the header, and also against the dark grey of the menu. And also a bit of every other shade :)

In the end this was the shade we liked the most in a wide variety of situations, the only place where it sucked was over the "warning" orange/yellow banners used for revisions, and the orange "Save" button also for revisions. These felt like rare enough combinations to not be worth affecting the choice too much.

If you or anyone else wants to have a look at other shades, here is my "outline color switcher" code:

```scss

.outline-teal *:focus { outline-color: $color-teal !important };
.outline-teal-darker *:focus { outline-color: $color-teal-darker !important };
.outline-teal-dark *:focus { outline-color: $color-teal-dark !important };
.outline-new *:focus { outline-color: #ffbf47 !important };
.outline-blue *:focus { outline-color: $color-blue !important };
.outline-red *:focus { outline-color: $color-red !important };
.outline-red-dark *:focus { outline-color: $color-red-dark !important };
.outline-orange *:focus { outline-color: $color-orange !important };
.outline-orange-dark *:focus { outline-color: $color-orange-dark !important };
.outline-green *:focus { outline-color: $color-green !important };
.outline-salmon *:focus { outline-color: $color-salmon !important };
.outline-salmon-light *:focus { outline-color: $color-salmon-light !important };
.outline-white *:focus { outline-color: $color-white !important };
.outline-black *:focus { outline-color: $color-black !important };
```

```js
const select = document.createElement('select');
select.style = `position: fixed; bottom: 0; right: 0; width: 200px;`;

const colors = [
  'outline-new',
  'outline-blue',
  'outline-red-dark',
  'outline-orange',
  'outline-orange-dark',
];

select.addEventListener('change', (e) => {
  colors.forEach(color => {
    document.body.classList.remove(color);
  });
  document.body.classList.add(e.target.value);
});

colors.forEach(color => {
  const option = document.createElement('option');
  option.value = color;
  option.innerHTML = color;
  select.appendChild(option);
});

document.body.appendChild(select);
```

client/scss/components/_main-nav.scss
thibaudcolas: Double checked – all seems fine.

thibaudcolas:
Thanks for the review @lb-! I had a look at the breadcrumbs, they are styled in a way which makes it look like all of the darkened area is clickable, even though just the `a` tag is. This is why the outline position isn't consistent with what seems "actionable". You can see how my mouse pointer changes between clickable and not-clickable in the GIF below:

breadcrumbs-focus

I'm tempted to fix this at the same time as the focus outline rather than just make the outline a bit bigger, so will do this in a separate PR.

Other people have separately reported issues with other elements, which I'll also probably look into at the same time:

- Radio buttons
- Checkboxes

These two didn't even have outlines at all before, so their new not-so-right focus outline still feels like a step in the right direction. If anyone else notices other parts of the UI that might need special treatment, I'll include these in the PR as well.

lb-:
@thibaudcolas - awesome - thanks for the epic detail work on this ticket. Great to see it merged in!

PR #5318, opened 2019-05-23
Fixes ordering in API search with Elasticsearch

### LINKED ISSUE #3514: API search order has no effect
Steps to reproduce:
- use wagtail 1.9 & elasticsearch 2 or any version of [wagtail-pg-search-backend](https://github.com/wagtail/wagtail-pg-search-backend)
- install Wagtail’s API
- go to `http://localhost:8000/api/v2/pages/?search=a+non-empty+search&order=first_published_at` where you replace `a+non-empty+search` with any search query that returns results
- go to `http://localhost:8000/api/v2/pages/?search=a+non-empty+search&order=-first_published_at` (with the same replacement as before)
- compare the results

As you should see, ordering has no effect on the actual order of results.
I discovered this issue while trying to implement search ordering (not to be confused with ranking) in wagtail-pg-search-backend. And I couldn’t find how to do it with the current way search is implemented.

So I guess searching doesn’t handle custom ordering. Unfortunately, [the opposite is written in the docs](http://docs.wagtail.io/en/v1.9/advanced_topics/api/v2/usage.html#search-operator), even with details on how to use it.

m1kola:
Wagtail's Elasticsearch search backend doesn't currently support reverse ordering set on a `QuerySet` using the `reverse` method.

Ordering behaviour already covered by API tests. They pass on `master` because we run them with the database backend. I think it makes sense to cover support of ordering set using `QuerySet.reverse` in the Elasticsearch backends as part of #5319.

Fixes #3514

### DIFF
--- a/wagtail/api/v2/filters.py
+++ b/wagtail/api/v2/filters.py
@@ -76,23 +76,17 @@ class OrderingFilter(BaseFilterBackend):
 
                 return queryset.order_by('?')
 
-            # Check if reverse ordering is set
             if order_by.startswith('-'):
-                reverse_order = True
-                order_by = order_by[1:]
+                order_by_field = order_by[1:]
             else:
-                reverse_order = False
+                order_by_field = order_by
 
             # Add ordering
-            if order_by in view.get_available_fields(queryset.model):
+            if order_by_field in view.get_available_fields(queryset.model):
                 queryset = queryset.order_by(order_by)
             else:
                 # Unknown field
-                raise BadRequestError("cannot order by '%s' (unknown field)" % order_by)
-
-            # Reverse order
-            if reverse_order:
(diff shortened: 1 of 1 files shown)


m1kola:
@kaedroho there are tests already in place for ordering in API (see the [`TestPageListing` test case](https://github.com/wagtail/wagtail/blob/c109dd3495ed67ef8a8f238f6ba7f163fbec6138/wagtail/api/v2/tests/test_pages.py)). I don't think it makes sense to cover the Elasticsearch search backend issue in API tests. Also see my note from the original post:

> Ordering behaviour already covered by API tests. They pass on `master` because we run them with the database backend. I think it makes sense to cover support of ordering set using `QuerySet.reverse` in the Elasticsearch backends as part of #5319.

PR #5320, opened 2019-05-23
Added CI config to generate nightly builds from master

kaedroho:
This adds CI config to generate a wheel file each night (12am UTC) and upload it to a public S3 bucket. This should make it easier for developers to try out the latest master without having to make a build of Wagtail.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,9 +47,34 @@ jobs:
       - run: npm run dist
       - run: bash <(curl -s https://codecov.io/bash) -F frontend
 
+  nightly-build:
+    docker:
+      - image: circleci/python:3.7.3
+    steps:
+      - checkout
+      - run: cd ~ && wget https://nodejs.org/dist/v8.7.0/node-v8.7.0-linux-x64.tar.gz
+      - run: cd /usr/local/ && sudo tar --strip-components 1 -xzf ~/node-v8.7.0-linux-x64.tar.gz
+      - run: pip install wheel boto3
+      - run: npm install
+      - run: npm run dist
+      - run: PYTHONPATH=. python scripts/nightly/get_version.py > wagtail/__init__.py
+      - run: python setup.py bdist_wheel
+      - run: python cripts/nightly/upload.py
+
 workflows:
     version: 2
     test:
-        jobs:
-          - backend
-          - frontend
+      jobs:
+        - backend
+        - frontend
+
+    nightly:
+      jobs:
+        - nightly-build
+      triggers:
+        - schedule:
+            cron: "0 0 * * *"
+            filters:
+              branches:
+                only:
+                  - master
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,7 +59,7 @@ jobs:
       - run: npm run dist
       - run: PYTHONPATH=. python scripts/nightly/get_version.py > wagtail/__init__.py
       - run: python setup.py bdist_wheel
-      - run: python cripts/nightly/upload.py
+      - run: python scripts/nightly/upload.py
 
 workflows:
     version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,7 +57,8 @@ jobs:
       - run: pip install wheel boto3
       - run: npm install
       - run: npm run dist
-      - run: PYTHONPATH=. python scripts/nightly/get_version.py > wagtail/__init__.py
+      - run: PYTHONPATH=. python scripts/nightly/get_version.py > __init__.py
+      - run: mv __init__.py wagtail/__init__.py
       - run: python setup.py bdist_wheel
       - run: python scripts/nightly/upload.py
 
--- a/scripts/nightly/get_version.py
+++ b/scripts/nightly/get_version.py
@@ -1,0 +1,31 @@
+"""
+This script is called by the nightly build in Circle CI.
+
+It alters the version of Wagtail to include the build date
+"""
+
+import datetime
+
+from wagtail import VERSION
+
+
+INIT_TEMPLATE = """
+from wagtail.utils.version import get_semver_version, get_version
+
+# major.minor.patch.release.number
+# release must be one of alpha, beta, rc, or final
+VERSION = ({major}, {minor}, {patch}, 'dev', '{datestamp}')
+
+__version__ = get_version(VERSION)
+
+# Required for npm package for frontend
+__semver__ = get_semver_version(VERSION)
+"""
+
+
+print(INIT_TEMPLATE.format(
+    major=VERSION[0],
+    minor=VERSION[1],
+    patch=VERSION[2],
+    datestamp=datetime.date.today().strftime('%Y%m%d'),
+))
--- a/scripts/nightly/upload.py
+++ b/scripts/nightly/upload.py
@@ -1,0 +1,24 @@
+import datetime
+import pathlib
+import sys
+import json
+
+import boto3
+
+dist_folder = pathlib.Path.cwd() / 'dist'
+
+try:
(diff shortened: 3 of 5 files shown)


.circleci/config.yml
```diff
@@ -47,9 +47,34 @@ jobs:
...
+    steps:
+      - checkout
+      - run: cd ~ && wget https://nodejs.org/dist/v8.7.0/node-v8.7.0-linux-x64.tar.gz
+      - run: cd /usr/local/ && sudo tar --strip-components 1 -xzf ~/node-v8.7.0-linux-x64.tar.gz
+      - run: pip install wheel boto3
+      - run: npm install
+      - run: npm run dist
+      - run: PYTHONPATH=. python scripts/nightly/get_version.py > wagtail/__init__.py
+      - run: python setup.py bdist_wheel
+      - run: python cripts/nightly/upload.py
```
zerolab: missing "s" in `python cripts/nightly/upload.py`. Should be `python scripts/nightly/upload.py`

kaedroho: Oops, thanks!

scripts/nightly/upload.py
```diff
@@ -0,0 +1,24 @@
...
+    sys.exit()
+
+print("Uploading", f.name)
+s3 = boto3.client('s3')
+s3.upload_file(str(f), 'releases.wagtail.io', 'nightly/' + f.name, ExtraArgs={'ACL':'public-read'})
+
+print("Updating latest.json")
+
+boto3.resource('s3').Object('releases.wagtail.io', 'nightly/latest.json').put(Body=json.dumps({
+    "url": 'https://releases.wagtail.io/nightly/' + f.name,
```
m1kola: What do you think about replacing it with a `txt` containing just a link or a version of the nightly build? This way it will require less tools to get the actual URL (no need to have `jq` or use `grep`, etc): everything you need is `curl` or `wget -O-`.

So it will be something like: 

```
# URL in latest.txt
pip install $(curl https://releases.wagtail.io/nightly/latest.txt)

# Version in latest.txt
pip install https://releases.wagtail.io/nightly/package-name-$(curl https://releases.wagtail.io/nightly/latest.txt).whl

```

scripts/nightly/get_version.py
```diff
@@ -0,0 +1,31 @@
...
+
+from wagtail import VERSION
+
+
+INIT_TEMPLATE = """
+from wagtail.utils.version import get_semver_version, get_version
+
+# major.minor.patch.release.number
+# release must be one of alpha, beta, rc, or final
+VERSION = ({major}, {minor}, {patch}, 'dev', '{datestamp}')
```
gasman: Don't we need to update wagtail.utils.version to recognise `dev` as a valid release type? https://github.com/wagtail/wagtail/blob/master/wagtail/utils/version.py#L16

m1kola: Any objections to use `alpha` or `beta` instead?

m1kola: Also I'm not sure about `datestamp`. Are we following [semver](https://semver.org/) here?

zerolab: There is no consensus for that.
See https://github.com/semver/semver/issues/200 and newer, an RFC https://github.com/semver/semver/pull/519

kaedroho: `'dev'` is the PEP440 standard name to use for a "Developmental release":

> Some projects make regular developmental releases, and system packagers (especially for Linux distributions) may wish to create early releases directly from source control which do not conflict with later project releases.

https://www.python.org/dev/peps/pep-0440/#developmental-releases

I think this description best fits with nightly releases.

> Also I'm not sure about datestamp. Are we following semver here?

I'm following [PEP 440](https://www.python.org/dev/peps/pep-0440) which I think is most appropriate for python packages.

scripts/nightly/upload.py
kaedroho: I wasn't really thinking people would use this file directly. My plan is to make a static HTML page that retrieves this file and generates the correct pip install command.

scripts/nightly/get_version.py
m1kola: [Release documentation](http://docs.wagtail.io/en/v2.5.1/contributing/release_process.html#release-cadence) says that Wagtail uses "loose form" of SemVer:

> Wagtail uses a loose form of [semantic versioning](http://semver.org/). SemVer makes it easier to see at a glance how compatible releases are with each other. It also helps to anticipate when compatibility shims will be removed. It’s not a pure form of SemVer as each feature release will continue to have a few documented backwards incompatibilities where a deprecation path isn’t possible or not worth the cost.

I agree that PEP 440 is more appropriate, but I think we must to be consistent here. We should either:

* Follow existing convention which is SemVer. I guess we will need to use the [pre-release form](https://semver.org/#spec-item-9). Something like:`2.5.1-dev.20190606`
* Or change the existing convention to PEP 440. 

---

Digging deeper into this: it looks like we only say that we use SemVer in the docs, but we actually don't follow version formats. For example, `2.5rc1` I believe is not a SemVer format (SemVer would be `2.5-rc.1`). I guess it's because we copy-pasted Django's release docs and `django.utils.version` :)

Our versioning actually look more like PEP 440. Does it make sense to update misleading docs and comments in [`wagtail.utils.version`](https://github.com/wagtail/wagtail/blob/669c1af4a0f4634395a63fd8f4d6a8f42f539730/wagtail/utils/version.py) while we are here?

kaedroho: I don't think we strictly follow the format of the semver spec. That paragraph is just trying to say that we avoid breaking stuff in patch releases (or at least try to).

Since this is a python package, this version will need to be understood by setuptools, pip and PyPI. All of these expect PEP 440 formatted versions so I think that is what we should go with.

gasman: Just to be clear, my issue is not with whether `dev` follows the relevant spec: the problem is that the `wagtail.utils.version.get_version` function fails with a `KeyError` when passed a string other than 'final', 'alpha', 'beta' or 'rc':

    >>> from wagtail.utils.version import get_version
    >>> get_version([2, 6, 0, 'dev', 1])
      File "<console>", line 1, in <module>
      File "/vagrant/wagtail/wagtail/utils/version.py", line 17, in get_version
        sub = mapping[version[3]] + str(version[4])
    KeyError: 'dev'

https://github.com/wagtail/wagtail/blob/4a1f26778f1f48dc485eaa000b761dc995e40102/wagtail/utils/version.py#L15-L16

m1kola: > Since this is a python package, this version will need to be understood by setuptools, pip and PyPI. All of these expect PEP 440 formatted versions so I think that is what we should go with.

I agree. And we currently seem to use the right format. But the documentation and comments in `wagtail.utils.version` are currently misleading. I think we should just update these bits to avoid further confusion.

kaedroho: @gasman should be fixed now

wagtail/utils/version.py
```diff
@@ -3,28 +3,45 @@
...
 
+def get_complete_version(version=None):
+    """
+    Return a tuple of the Wagtail version. If version argument is non-empty,
+    check for correctness of the tuple provided.
+    """
+    if version is None:
+        from wagtail import VERSION as version
+    else:
+        assert len(version) == 5
```
m1kola: Is it ok to rely on asserts here? [They can be ignored](https://docs.python.org/3.6/using/cmdline.html#cmdoption-o).

kaedroho: This bit was taken from Django. But I don't think this would matter as any changes to the version would be run through CI and that wouldn't ignore this statement.

PR #5332, opened 2019-05-28
deprecate SiteMiddleware

aritas1:
As proposed in the comment [#3834](https://github.com/wagtail/wagtail/issues/3834#issuecomment-340009510) by @gasman i changed the wagtail behaviour to only use `Site.find_for_request(request)` and added some caching via a `request._wagtail_site` variable.

Since there is no use for the `SiteMiddleware` anymore, i pushed it for deprecation in wagtail 2.8.

Also i don't see the need for a lazy evaluation of the `SiteMiddleware` `request.site` stuff due to the possibility to remove the middleware completely.

I needed to change a lot of tests. Please take some time to review them. There was some stuff going on with requests being responses from `self.client.get()`. e.G. see the `TestCoreJinja` class changes.

- [x] checked with chrome 74.0 on the bakerydemo

### DIFF
--- a/docs/advanced_topics/settings.rst
+++ b/docs/advanced_topics/settings.rst
@@ -37,15 +37,10 @@ Middleware (``settings.py``)
     'django.middleware.clickjacking.XFrameOptionsMiddleware',
     'django.middleware.security.SecurityMiddleware',
 
-    'wagtail.core.middleware.SiteMiddleware',
-
     'wagtail.contrib.redirects.middleware.RedirectMiddleware',
   ]
 
 Wagtail requires several common Django middleware modules to work and cover basic security. Wagtail provides its own middleware to cover these tasks:
-
-``SiteMiddleware``
-  Wagtail routes pre-defined hosts to pages within the Wagtail tree using this middleware.
 
 ``RedirectMiddleware``
   Wagtail provides a simple interface for adding arbitrary redirects to your site and this module makes it happen.
@@ -692,7 +687,6 @@ These two files should reside in your project directory (``myproject/myproject/`
       'django.middleware.clickjacking.XFrameOptionsMiddleware',
       'django.middleware.security.SecurityMiddleware',
 
-      'wagtail.core.middleware.SiteMiddleware',
       'wagtail.contrib.redirects.middleware.RedirectMiddleware',
   ]
 
--- a/docs/reference/pages/model_reference.rst
+++ b/docs/reference/pages/model_reference.rst
@@ -233,7 +233,9 @@ In addition to the model fields provided, ``Page`` has many properties and metho
 
 The ``Site`` model is useful for multi-site installations as it allows an administrator to configure which part of the tree to use for each hostname that the server responds on.
 
-This configuration is used by the :class:`~wagtail.core.middleware.SiteMiddleware` middleware class which checks each request against this configuration and appends the Site object to the Django request object.
+This configuration is used by the :class:`~wagtail.core.middleware.SiteMiddleware` middleware class which checks each request against this configuration and appends the Site object to the Django request object. (deprecated)
+
+As of wagtail 2.6 use the :meth:`~wagtail.core.models.Site.find_for_request` function to get the current wagtail Site. This function will check at the first call which Site to return for a given Django request object.
 
 Database fields
 ~~~~~~~~~~~~~~~
--- a/wagtail/admin/tests/test_admin_search.py
+++ b/wagtail/admin/tests/test_admin_search.py
@@ -3,7 +3,7 @@ Tests for the search box in the admin side menu, and the custom search hooks.
 """
 from django.contrib.auth.models import Permission
 from django.template import Context, Template
-from django.test import RequestFactory, TestCase
+from django.test import RequestFactory, TestCase, override_settings
 from django.urls import reverse
 
 from wagtail.admin.utils import user_has_any_page_permission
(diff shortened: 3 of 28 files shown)


gasman:
Thanks @aritas1!
I'm a bit wary about all the changes to the tests (i.e. the need for `ALLOWED_HOSTS=['*'], and setting an explicit host/port on the request) - it seems to imply that the new implementation isn't a direct functional replacement for the old one, and might require users to make similar changes to their "real-world" code (beyond removing references to `request.site`, of course). Can you give more detail about why these changes are necessary, please?

gasman:
Have now done a bit more digging into why `ALLOWED_HOSTS=['*']` is required:

Django's test runner replaces the `ALLOWED_HOSTS` setting with `['testserver']`; however, `Page.dummy_request` will construct a request object with a hostname matching the Site record that the page belongs to (if any). In both the `testapp` fixture and `wagtailcore`'s initial migrations, this hostname is `localhost`. This means that logically any dummy requests created inside the test runner would have been failing the ALLOWED_HOSTS check all along - this change just makes that failure visible. We should probably set `ALLOWED_HOSTS = ['testserver', 'localhost']` in wagtail/tests/settings.py.

I haven't fully worked out why this wasn't causing errors up to now: the ALLOWED_HOSTS check happens in `django.http.request.HttpRequest`'s `get_host` method, and throws a `DisallowedHost` exception (which inherits from `SuspiciousOperation`) if it fails - and both the old and new code is calling `get_host` from `Site.find_for_request`. The difference seems to be that something in the middleware runner is catching and ignoring `SuspiciousOperation`.

I'd like to do some more investigation to figure out if this change has any implications on real-world sites. If the impact is confined to "the new code path makes Wagtail fail slightly more loudly when ALLOWED_HOSTS is set incorrectly", that's probably not a bad thing :-)

aritas1:
Hey @gasman ,

i dug into the code again and there are way more `get_host` calls with the new code. `wagtail.contrib.settings.context_processors.settings` may be the source for some of the `ALLOWED_HOSTS` problems. These extra calls and with the bad shaped test requests, the context_processors raises a lot of `DisallowedHost` exceptions.

Maybe the changes to `wagtail.contrib.settings.context_processors.settings` are more intrusive on the second thought and need a SimpleLazyObject wrapper to reduce the extra SQL query on none wagtail pages (#3834), pages that don't need `contrib.settings` and for some tests.

Adding localhost to `ALLOWED_HOSTS` sounds good to me.
`ALLOWED_HOSTS` is usually one of the first things to be configured if a project moves to production due to its empty default value. I don't see any harm to rely on a correct / `*` ALLOWED_HOSTS setting.

---
For more detail on the test changes:

A few of the added explicit host/port definitions are not necessary due to the behaviour of `Page.dummy_request`. I missed that while changing the tests.

The second thing i needed to change was the construction of a dummy request via `self.client.get()`. `Client.get()` dose not return a request, it returns a `TemplateResponse`.
Testing a render function with a response was working before (the `.site` was added to the response object and the context_processors picked it up). But due to the new `wagtail.contrib.settings.context_processors.settings` failing to determine the site form the response made me change this part as well.

There are two ways of fixing this. Using the `RequestFactory.get()` or in my case using `HttpRequest()` and populating the needed parts manually. The `RequestFactory` is calling `Page.dummy_request()` some how? and corrects the missing host/port on the fly. `RequestFactory` has too much side effects for my taste.

have a look at:
- test_templates.py 
- test_jinja2.py

---

I cleaned the tests to use less explicit hostnames and ports, removed a few `override_settings`, added `ALLOWED_HOSTS` to the settings.py and added `SimpleLazyObject` wrapper for the `settings.context_processors`.

timmysmalls:
Hey guys, I've had a look at this as well. I resolved the conflicts, and it's back up to date with the current master again. Here's what I found:

One of the questions raised by @gasman was regarding the `ALLOWED_HOSTS`, which had to be set manually, alongside the `HTTP_HOST` and `SERVER_PORT`. This is not necessary. It's important to realize that by default, Django will always use 'testserver', and port 80 for requests, as was pointed out. This host is allowed by default. And this is the way it happened before this issue, and it still does. If you set them mually to localhost, this will cause problems, and will lead to the behavior we saw in this PR, where we need to manually whitelist localhost. In the old tests `request.site` was set manually on every Request object (or sometimes mistakenly on Response objects that were supposed to be requests). Instead of changing `request.site` to `request._wagtail_site`, @aritas1 set the host and port manually: to localhost, which is not i allowed hosts. Since at this point `_wagtail_site` has not been set yet, when it needs to be set, it will execute the `_find_for_request` method, which will in turn call `get_host`. That's where the `DisallowedHost` exception is raised. If we simply set the `_wagtail_site`, and leave the `HTTP_HOST` alone, just like we did when the SiteMiddleware was still present, there will be no issues. The `HTTP_HOST` will default to 'testserver', which is whitelisted, and there will be no problems. That's also why the DisallowedHost error was never raised before. The host was always testserver.

I hope I'm making sense here.

The Site object that gets set will always be the same regardless of the value of `HTTP_HOST`: there is only one site object, and it has `is_default_site` set to true. So there's no need to try to explicitly force this by setting `HTTP_HOST` and `SERVER_PORT` to match the Site object we're trying to target. This is only necessary when we want to spoof a request from a different site, and check whether wagtail resolves it to the correct site object. That logic hasn't been changed in this PR. The only difference is that this logic used to be called in the middleware, and stored on `request.site`, while now `_wagtail_site` is set lazily to cache the result from the `find_for_request` method. Replacing the manual setting of these META attributes with `request._wagtail_site`, allows us to get rid of just about every override for the allowed hosts, and we can also get rid of the global allowed hosts setting this way.

I've created a new PR ( #5673 ) that should be good to go. It has the work done by @aritas1 in it, with a few commits by me on top. Let me know if you would like me to clean up the commit history a bit more, or if there are still some things that need to happen here. I hope we can resolve this one!

gasman:
Superseded by #5787

Pomax:
It looks like the 2.10 release notes forgot to mention this change in the [upgrades considerations](https://docs.wagtail.io/en/stable/releases/2.10.html#upgrade-considerations) section, can that still be added to explain how to rewrite older code? (the console log just says `RemovedInWagtail211Warning: wagtail.core.middleware.SiteMiddleware and the use of request.site is deprecated. Please update your code to use Site.find_for_request(request) in place of request.site, and remove wagtail.core.middleware.SiteMiddleware from MIDDLEWARES`  but does not say where `Site` should be imported from; also that should be `MIDDLEWARE` not `MIDDLEWARES`.

Pomax:
It looks like the 2.9 release notes forgot to mention some essentials in the [upgrades considerations](https://docs.wagtail.io/en/v2.9/releases/2.9.html#sitemiddleware-and-request-site-deprecated) section, can that still be added to explain how to rewrite older code? While it mentions what to replace, it does not mention where `Site` is supposed to be imported from.

Also, on an error notice, the console warning right now says RemovedInWagtail211Warning: wagtail.core.middleware.SiteMiddleware and the use of request.site is deprecated. Please update your code to use Site.find_for_request(request) in place of request.site, and remove wagtail.core.middleware.SiteMiddleware from MIDDLEWARES`, which should be MIDDLEWARE rather than MIDDLEWARES.

gasman:
Thanks for the suggestion @Pomax! I've added some examples in 925d64730ab5e597bf7d90bbe26bad417bafdd48 (for the master branch) and d970b2e9d2b389acd1b339c9c838d7094995226c (for the stable/2.10.x branch) now. I've also fixed the deprecation warning in ea3fcf2d6a3252418ce6013ea623bd998e2a4f62 - as development of 2.11 is now under way, that message is no longer present on master, but I noticed that another reference to `MIDDLEWARES` had snuck into the 2.11 release notes (which I've now fixed in 4f42e61a2cc7f71d4406a65afc33a292192dd1be).

Pomax:
Nice, cheers!

PR #5344, opened 2019-05-30
Implemented branding title prefix customization

dmeijboom:
Currently there is no way to override the (meta)title prefix (which is "Wagtail" by default). With these small changes it's possible to brand/customize this as well.


gasman:
Thanks @dmeijboom! Looks good to me - merged in dd1597ea37e99e4e1e1ec0ba6b9b86722736eea3.

PR #5346, opened 2019-06-02
Add template block for admin title name.

nicksergeant:
This change adds a template block around the "Wagtail" string in the skeleton title, allowing implementors to override the title instead of having to override the entire `skeleton.html` template.

Unsure if this was omitted intentionally or not (for branding?), but figured I'd open a PR to gauge interest.


zerolab:
Hey @nicksergeant, 
thank you for your pull request. 
This is a duplicate of #5344 (also very recently opened). I cannot comment from the maintainer's point of view, but it does look like there is interest for this, so imho it can't hurt

nicksergeant:
@zerolab ah great! That one has documentation also. Closing in favor of #5344.

Issue #5337, opened 2019-05-29
Implement focus management in main menu


thibaudcolas:
### Issue Summary

Tabbing through Wagtail’s main menu, it should be possible to move between all visible menu items one by one, opening submenus as required. Instead, users have to tab through submenus that are closed:

![menu-focus-management](https://user-images.githubusercontent.com/877585/58544795-eaf0cc00-81f9-11e9-9817-0407124e04df.gif)

The correct behavior would be:

- When submenus aren't open, users shouldn't have to tab through their menu items.
- Submenus can be opened by clicking, pressing Enter while focused; and closed with the Esc key or clicking outside
- Opening a submenu, the focus now moves to the submenu and is trapped there until closed
- Submenu triggers should use `button` tags instead of `a`, so screen readers announce the correct behavior (button instead of link), and browsers do not display a link target of "#" (see #5349).

### Steps to Reproduce

1. Go to the Wagtail admin
2. Navigate with a keyboard through the menu items

Related: #5336

thibaudcolas:
Added a note that we should also change submenu triggers to be `button` elements instead of `a`, both so they announce correctly to screen reader users (these are not links), and so browsers do not show the link targets for no reason (#5349).

thibaudcolas:
Note: would be better to get #5254 in before starting work on this.

thibaudcolas:
Additional task from #5274:

> Clarify that the user is focused on the side navigation by improving the voice-over command to say ‘Navigation’, for example: ‘link, Settings list navigation, 9 items

As part of this issue, I think it makes sense to implement correct labels for each of the submenus, as well as ARIA markup so they are announced as menus (`aria-haspopup="dialog"`, etc).

marteki:
I'm working on this during Wagtail Sprint US 2019.

marteki:
@thibaudcolas I'm going to take the route of using `aria-expanded="false"` or `aria-expanded="true"` to show the submenus, since the standard ARIA menu markup is meant for application menus (and not so much for navigation lists on web pages).

fabienheureux:
@marteki is there an open PR for this issue ? 
I wanted to work on it, following you approach, but if part of the work has already been done here it could be helpful 😉

PR #5349, opened 2019-06-04
Add a "Dashboard" label for logo link in the main nav for screen reader users

thibaudcolas:
This change adds a "Dashboard" label to the main nav’s "logo link" for screen reader users:

main-nav-logo-link

Up until now, for screen reader users, this link was announced as "Wagtail v2.x.y", which doesn't convey that the link takes you to the dashboard. This PR removes the `title` with the version number and adds an `aria-label`, so screen reader users have the same link label as what gets displayed on mobile viewports: "Dashboard".

I had also considered leaving the `title` as-is and just adding `aria-label` on top, but then some screen readers would announce the link as "link, Dashboard Wagtail v2.x.y", which isn't of much help.

I've updated the issue template to point users at the version number written in the Settings menu, rather than hovering Wagtail. This is also good because you could only access the version in `title` with a mouse, whereas the one under Settings is accessible with the keyboard and touch screens too.

---

Tested in macOS 10.14.4 with latest Chrome & Safari with VoiceOver, and iOS 12.3 Safari with VoiceOver. I think that's enough for cross-browser tests considering the change.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -20,5 +20,5 @@ Any other relevant information. For example, why do you consider this a bug and 
 
 * Python version: Run `python --version`.
 * Django version: Look in your requirements.txt, or run `pip show django | grep Version`.
-* Wagtail version: Hover over the Wagtail bird in the admin, or run `pip show wagtail | grep Version:`.
+* Wagtail version: Look at the bottom of the Settings menu in the Wagtail admin, or run `pip show wagtail | grep Version:`.
 * Browser version: You can use http://www.whatsmybrowser.org/ to find this out.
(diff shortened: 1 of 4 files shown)


tomdyson:
I couldn't find the version in the settings menu, until I realised that it was mostly concealed by the status bar in my browser (Chrome, OS X) -

Screenshot 2019-06-04 17 22 36

 - but this is still an improvement and we can address the positioning in a separate PR if anyone thinks it's worth it. Thanks @thibaudcolas!

thibaudcolas:
Ah good point @tomdyson, this is because focus is still on the "Settings" menu item, so Chrome displays its target of '#'. There is no good reason to use an `a` tag with `href="#"` for this, I'll add a note to change this as part of #5337.

gasman:
@thibaudcolas Just encountered an annoying edge case with the version number label, unfortunately... one of our sites has a very long settings menu that exceeds the height of the window, and the version number scrolls along with the menu, permanently obscuring one of the menu items:

Screenshot 2019-06-05 at 20 27 38

gasman:
Never mind, didn't realise that's been there forever... :-)

thibaudcolas:
👌 that probably warrants its own issue! I've noticed this as well when navigating sites in a few viewports.

PR #5353, opened 2019-06-04
Support custom add/edit page views on a per-page-type basis

gasman:
This PR converts the page creation and edit views in `wagtail.admin.views.pages` into class-based views with several overrideable methods (`get_page_instance`, `get_edit_handler`, `get_form_class`, `get_context_data`) and provides a mechanism for custom versions of those views to be hooked in on a per-page-type basis.

The motivation for this was to provide a shortcut link for editors, taking them to a version of the 'create page' view with some fields dynamically pre-populated based on a URL parameter. This apparently isn't possible (aside from questionable hacks) through the existing hooks or form / edit_handler customisation features.

This is, admittedly, a pretty heavyweight code change for that one use-case, but it seems like the most future-proof approach. Other options I considered and rejected were a new hook for modifying the page instance used to build the form (but that felt like an overly specific 'niche' thing for a hook - if every similarly low-level detail was hookable, we'd end up with millions of them...) and adding new properties / methods to the model to control how the object/form instantiation behaves (which would have coupled the model code far too tightly to the view-level behaviour - we're already pushing this probably further than necessary with the likes of `base_form_class`).

### DIFF
--- a/docs/advanced_topics/customisation/page_editing_interface.rst
+++ b/docs/advanced_topics/customisation/page_editing_interface.rst
@@ -211,3 +211,25 @@ Wagtail will generate a new subclass of this form for the model,
 adding any fields defined in ``panels`` or ``content_panels``.
 Any fields already defined on the model will not be overridden by these automatically added fields,
 so the form field for a model field can be overridden by adding it to the custom form.
+
+
+Customising page creation/edit views
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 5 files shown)


lb-:
Note - this approach would be good for scenarios like this question
https://stackoverflow.com/questions/55640685/how-do-i-customize-the-wagtail-page-copy-experience/60084836#60084836

Although that question relates to the copy view (which is not converted to a class based view in this PR), things like this would be easier in general once class based views are adopted.

ababic:
@gasman I wonder whether the class-based view changes would be better accompanied with a more generic mechanism to override the views (that didn't pollute the model classes with additional configuration), like the approach outlined in #6166?

zerolab:
Closing in favour of #6025

PR #5359, opened 2019-06-06
Change icon font implementation to be screen-reader friendly (#1224)

thibaudcolas:
This PR changes Wagtail’s icon font implementation so screen readers do not read icons out loud as if they were letters anymore. It addresses some of the issues raised in #1224, and the number 1 "icon font" issue from https://github.com/wagtail/wagtail/issues/4199#issuecomment-358640129.

The implementation change is very small: simply switching from ASCII characters to represent icons to [Private Use Areas](https://en.wikipedia.org/wiki/Private_Use_Areas) unicode code points. Private-use code points are perfect for icon fonts, where you don’t want icons failing to render to fall back to other fonts (what’s causing the issue with the current implementation).

Ideally we would want to further rework the icon implementation (#4821, or further intermediary refactoring to improve the icon font), but this PR is enough for the icons to be invisible for screen readers. I specifically tested this with:

- VoiceOver on macOS Safari
- VoiceOver on iOS Safari
- NVDA on Firefox Windows 10

For people not using screen readers, there are no differences whatsoever.

* [x] For front-end changes: Did you test on all of Wagtail’s supported browsers?

I ran cross-browser tests in Firefox, Safari, Chrome on macOS. IE11, MS Edge on Windows. iOS Safari. I also ran a visual regression test suite across the whole admin (using Chrome for rendering), which found no difference to how icons are rendered.


wagtail/images/static_src/wagtailimages/js/hallo-plugins/hallo-wagtailimage.js
```diff
@@ -15,7 +15,7 @@
                     uuid: this.options.uuid,
                     editable: this.options.editable,
                     label: 'Images',
-                    icon: 'icon-picture',
+                    icon: 'icon-image',
```
allcaps: Does changing the icon name not cause problems when integrators re-use icons for their own blocks or menu items?

thibaudcolas: Ah, yes. I’ve been careful to change as little as possible, so the only change to names is that:

- `icon-picture` is removed. Use `icon-image` instead.
- `icon-file-text-alt` is removed. Use `icon-doc-full` instead.

Both of these were only used in Hallo as far as I can see – I’ll add a note in the release notes.

tomdyson:
> I’ll add a note in the release notes.

@thibaudcolas did you do this? sorry if I've missed it.

thibaudcolas:
@tomdyson nope! Done in #5363.

PR #5361, opened 2019-06-06
Offer no delete options, as is on pages and snippets

kevinhowbrook:
Really just an idea, as we now have 'No, don't delete' buttons for snippets, it might be good to adopt this in other places so the UI is more consistent.

* Do the tests still pass? *Yes*
* Does the code comply with the style guide? (Run `make lint` from the Wagtail root) *Yes*
* For Python changes: Have you added tests to cover the new/fixed behaviour? *No, I don't think a test for this is necessary.*


gasman:
Looks good to me - thanks @kevinhowbrook! Merged in 13133dad5b7136747fb55e672516425565fbb98a + parents

PR #5368, opened 2019-06-08
Add missing class imports to hook documentation

kevinhowbrook:
Just some small documentation tweaks. Running these without the imports will throw `name 'HttpResponse' is not defined` and could cause confusion.


gasman:
Thanks @kevinhowbrook!

PR #5373, opened 2019-06-11
Expose Permission object in format_permssions tag

ghickman:
A checkbox (specifically, a `BoundWidget`) is instantiated with data from a `django.contrib.auth.models.Permission` instance.  However it doesn't receive the full object so it isn't available in the template.  This stops a user from accessing useful values like `Permission.name`.

This adds the relevant Permission object to the template in a dict for ease of access.


helenb:
@ababic please could you review this?

ababic:
Merged in 3624558. Thanks @ghickman! And congrats on your first contribution to Wagtail :)

Issue #5360, opened 2019-06-06
Per-site 404 pages


Pomax:
I can't seem to find anything in the documentation about per-site distinct 404 pages, and the standard django multisite approach (adding a 404handler to your urls.py) doesn't seem to apply here, since there isn't strictly speaking "an app" for which urls.py can be specified, the sites are fully maintained inside of the wagtail CMS.

Is there a way to specify what to server as 404 document, per site (either static, or an actual page that exists in the CMS, similar to how a root document needs to be picked)?

Pomax:
In the mean time, is there maybe a file override that I could write as a locally loaded patch file, that leaves everything intact except for the logic around where to find the 404 file? Which file would I need to override for that?

allcaps:
As far as I can see, Wagtail does use the Django default: `raise Http404()`. So this concerns mainly Django logic.

> For convenience, and because it’s a good idea to have a consistent 404 error page across your site, Django provides an Http404 exception.
https://docs.djangoproject.com/en/2.2/topics/http/views/#the-http404-exception

Raise Http404 will call the default error view.

> The default error views in Django should suffice for most Web applications, but can easily be overridden if you need any custom behavior. Simply specify the handlers as seen below in your URLconf (setting them anywhere else will have no effect).
> The page_not_found() view is overridden by handler404:
> handler404 = 'mysite.views.my_custom_page_not_found_view'

https://docs.djangoproject.com/en/2.2/topics/http/views/#customizing-error-views

I haven't tested this, but I think you should be able to specify your own `handler404` (in your ursl.py) and point it to a custom `page_not_found` view.

> page_not_found(request, exception, template_name='404.html')
https://docs.djangoproject.com/en/2.2/ref/views/#the-404-page-not-found-view

Here you can select any template based on request.get_host()
https://docs.djangoproject.com/en/2.2/ref/request-response/#django.http.HttpRequest.get_host

allcaps:
Okay, this is silly. It is exactly like described in this PR: https://github.com/wagtail/wagtail/pull/5375

Pomax:
Glad to see David filed that - as things that users don't need to learn about when using Wagtail, since Wagtail hides a lot of the Django magic, having to create files that you normally don't is 100% unexpected and needs some solid documentation to make sure people can both find the info on how to do this, and of course have an example to follow along with.

Hopefully that PR gets merged soon =)

PR #5375, opened 2019-06-12
Explain multi-site support and limitations

DanielSwain:
This PR addresses [https://github.com/wagtail/wagtail/issues/5360](https://github.com/wagtail/wagtail/issues/5360)

Also see [https://github.com/wagtail/wagtail/issues/4288](https://github.com/wagtail/wagtail/issues/4288)

### DIFF
--- a/docs/reference/index.rst
+++ b/docs/reference/index.rst
@@ -11,3 +11,4 @@ Reference
     hooks
     signals
     project_template
+    multi_site_capabilities
--- a/docs/reference/multi_site_capabilities.rst
+++ b/docs/reference/multi_site_capabilities.rst
@@ -1,0 +1,41 @@
+.. _multi_site_capabilities:
+
+Multi-Site Capabilities
+=======================
+
+Overview
+--------
+
+Wagtail provides multi-site capability via the :ref:`site-model-ref` model.  One or more sites can be created via the Sites option under the admin Settings menu.  A root page is chosen and a hostname is defined for each site that is created.
+
+:ref:`settings` can be defined by creating a model that inherits from BaseSetting.
+
+While Wagtail provides multi-site capability and adds the site object to the request object, multi-tenant capability is not currently supported.  Therefore, if more than one site is set up, any user who has access to the admin can log in through any hostname.  After logging in, however, user rights that were assigned via Groups are honored with one exception:  When a user has rights to view or edit a particular BaseSetting-derived object, all site-specific instances of that particular object are editable.  However, the BaseSetting-derived object instance that is first presented for editing will be the instance that corresponds to the hostname under which the user is logged in.
+
+Site-Specific Error Pages
+-------------------------
+
+Since Wagtail is built on Django, standard Django exceptions are raised (for example, Wagtail raises Http404 when a page is not found).  The error pages returned by these exceptions can be customized in the same way that this customization is normally done in Django - by either placing in one of the folders where templates are stored a template for a particular error message (i.e. ``404.html``) or by providing a `custom view to handle the error`_.   By using the custom view method for overriding the rendering of a particular type of error, a site-specific custom error page can be returned.  To create site-specific custom 404 pages, place the ``handler404`` directive in the same ``urls.py`` file where the Wagtail urls have been included:
+
+.. _custom view to handle the error: https://docs.djangoproject.com/en/dev/topics/http/views/#customizing-error-views
+
+.. code-block:: python
+
+    from yourapp.views import custom404_view
+
+    handler404 = 'yourapp.views.custom404_view'
+
+Then, in the views.py file for your_app, place the conditional logic to return a site-specific error page:
+
+.. code-block:: python
+
+    from django.http import HttpResponseNotFound
+    from django.template.loader import render_to_string
+
+    def custom404_view(request, exception):
+        if request.site.hostname == 'site1.com':
+            return HttpResponseNotFound(render_to_string('404_site1.html', request=request))
+        elif request.site.hostname == 'site2.com':
+            return HttpResponseNotFound(render_to_string('404_site2.html', request=request))
+        else:
+            return HttpResponseNotFound(render_to_string('404.html', request=request))
--- a/docs/reference/multi_site_capabilities.rst
+++ b/docs/reference/multi_site_capabilities.rst
@@ -6,20 +6,24 @@ Multi-Site Capabilities
 Overview
 --------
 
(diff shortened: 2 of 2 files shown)


DanielSwain:
I've submitted this PR because I think that the multi-site capability needs to be explained a bit more prominently.  The explanation includes a statement about a current problem with Wagtail multi-site.  This PR is not submitted for search engine purposes, but if someone were to search on _multi-site_ or _multi-tenant_, then this page of the documentation might come up.  

If you would rather not have the Overview statements, then I could instead submit a PR with just the example as a Recipe.

Pomax:
Any chance this can get landed? It's pretty useful documentation to have, especially for some of the work we're doing at the mozilla foundation =)

docs/reference/multi_site_capabilities.rst
```diff
@@ -0,0 +1,47 @@
...
+=======================
+
+Overview
+--------
+
+Wagtail provides multi-site capability via the :ref:`site-model-ref` model.  One or more sites can be created via the Sites option under the admin Settings menu.  A root page is chosen and a :attr:`~wagtail.core.models.Site.hostname` is defined for each site that is created.
+
+:ref:`settings` can be defined by creating a model that inherits from ``BaseSetting``.
+
+While Wagtail provides multi-site capability and adds the ``site`` object to the request, multi-tenant capability is not currently supported.  Therefore, if more than one site is set up, any user who has access to the admin can log in through any hostname that has been defined for a site.  After logging in, however, all user rights that were assigned via Groups are honored.
```
tomdyson: This point requires a very specific definition of multi-tenancy. In my opinion, Wagtail's support for multi-tenancy is not inconsistent with the current Wikipedia definition:

> "A tenant is a group of users who share a common access with specific privileges to the software instance. With a multitenant architecture, a software application is designed to provide every tenant a dedicated share of the instance - including its data, configuration, user management, tenant individual functionality and non-functional properties."

https://en.wikipedia.org/wiki/Multitenancy

DanielSwain: I understood from @gasman in [#4288](https://github.com/wagtail/wagtail/issues/4288#issuecomment-365914981) that Wagtail is not completely multi-tenant yet.  IMHO groups/permissions is not the ideal way to achieve true multi-tenancy.  Serving HTML files is not currently handled by groups/permissions.  It seems that, until `host` is central to the logic of serving model data and assets (html files, images, documents), that true multi-tenancy is not achieved.  (As I'm new to Wagtail in the last 18 months, I make that statement not as a conclusion - I have lots to learn!  I'm simply thinking that multi-tenancy should work like one of the various Django multi-tenant packages work).  Perhaps we could discuss multi-tenancy at the upcoming conference in Philadelphia - not necessarily as a discussion topic but at least informally.

tomdyson: > Serving HTML files is not currently handled by groups/permissions

If you mean serving Wagtail pages, then that's not the case; you can restrict sections of the tree (which may correspond to sites) to users in specific groups:

https://docs.wagtail.io/en/v2.5.1/advanced_topics/privacy.html

> Perhaps we could discuss multi-tenancy at the upcoming conference in Philadelphia - not necessarily as a discussion topic but at least informally

For sure! Meanwhile I'm interested in @gasman's opinion (and anyone else's).

DanielSwain: When I refer to HTML files, I'm actually referring to different versions of an HTML file being able to be used for a `Page` model.  I think that the easiest way to begin thinking about multi-tenancy is by thinking in terms of homogeneous tenants (I gave the example of Chambers of Commerce [here](https://github.com/wagtail/wagtail/issues/4288#issue-297240375)).  In this case each tenant would have basically the same information needs and so a single set of Wagtail `Page` definitions could be made to service all tenants.  However, each Chamber would want to have their own look, and so when the `template` for `HomePage(Page)` says `template = 'home.html'`, then you have to ask which Chamber's `home.html` file is to be used?  In a comment [here ](https://stackoverflow.com/questions/56449188/having-a-404-for-each-site-in-a-wagtail-multisite-setup/56526327#56526327) I stated:

> I've started using a themes folder. The themes folder is at the first level in the project and within it are folders that each hold all the assets for a particular theme. In my ideal world, you would associate a theme with a site. 

Of course `settings` have to be configured appropriately so that the templates and the static assets can be found.  Multi-tenant or not, it makes much more sense to me to have everything related to a specific site and its appearance (HTML, CSS, JS, and image files, where appropriate) in a single named theme folder within the `themes` folder.  HTML files such as `allauth` files COULD still be stored in a `templates` folder at the root level of the project, but I've even started creating an `allauth` folder within my named theme folder because invariably I have to do a little customization of the `allauth` templates too.  This serves to keep EVERYTHING related to a particular website together with the benefit that, if I ever want to share a theme, then everything is in a single folder and not scattered across several apps.

tomdyson: This project hasn't been updated for a couple of years, but - as far as I understand - is designed to solve exactly this problem:

https://github.com/moorinteractive/wagtail-themes

It may need updating to work with recent Wagtail versions, but that could be a useful first step.

foxwoods369: @tomdyson I'd say you're right - that looks like exactly the same thing I'd implemented for a test project to host multiple sites on wagtail with templates based upon which site was being served. I don't like the hacked global using thread local storage though; is there some way that the site could be passed through better to the template loader?

Pomax: This thread feels like it's diverged from the actual PR quite a bit - is there a rephrasing that can be applied to this paragraph so that this can land and wagtail users have docs to fall back on when they need to set multiple 404s? =)

tomdyson: I suggest updating line 13 to the following:

> Note that users with access to the admin can log in through every hostname that has been defined for a site. After logging in, however, all user rights that were assigned via groups are honoured.

docs/reference/index.rst
```diff
@@ -11,3 +11,4 @@ Reference
     hooks
     signals
     project_template
+    multi_site_capabilities
```
gasman: This isn't really reference documentation as such - it would probably fit better under 'advanced topics'. (Admittedly, there are quite a few existing places in the docs where the line between informational and reference docs is blurrier than it should be :-) )

gasman:
Now addressed by #7228.

PR #5378, opened 2019-06-12
Fix help-inverse icon regression introduced by icon refactorings in #5359

thibaudcolas:
Changes from #5359 broke the `help-inverse` icon used for help text in the page editing UI. I thought this icon was using the same glyph as `help` – it’s not, it’s actually just displaying the "?" character.

The icon was missing from the styleguide which is also why I missed the breakage. I’ve added it there.

Tested in Chrome, Safari, Firefox on macOS + iOS Safari & IE11.

### DIFF
--- a/client/scss/components/_icons.scss
+++ b/client/scss/components/_icons.scss
@@ -122,7 +122,6 @@
     $size: 15px;
 
     &:before {
-        content: map-get($icons, 'help');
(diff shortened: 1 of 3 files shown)


thibaudcolas:
@allcaps can I ask you to have a look at this since it’s similar changes to #5359?

Issue #5351, opened 2019-06-04
Remove tab order customisations in CMS admin


thibaudcolas:
Some parts of the Wagtail UI define a customisation of the order in which items on the page can be tabbed-through with the `tabindex` attribute. This is generally very confusing for end users, especially when they rely on the keyboard to navigate through the page, and should be removed.

An example of where this happens is the page editing actions’ dropdown, where “Publish” is before “Unpublish”:

<img width="702" alt="tabindex" src="https://user-images.githubusercontent.com/877585/58895829-bd6dbc00-86ec-11e9-9d89-8a79c0fcb1fe.png">

---

The solution for this is to remove all code that sets `tabindex` values greater than 0: https://github.com/wagtail/wagtail/search?q=tabindex&unscoped_q=tabindex

thibaudcolas:
Fixed by #5383.

PR #5383, opened 2019-06-13
Fixed Remove tab order customisations in CMS admin #5351

jordan-bauer:
This is meant to address #5351 

I have changed all tabindex values to 0 in wagtail templates.

I was unable to reproduce the example issue, as I am unable to tabindex to any options in the page editing actions’ dropdown besides "SUBMIT FOR MODERATION".
Because of this I am unable to test if this had the desired effect.

Please let me know if there are other issue examples or additional work to be done on this issue.

### DIFF
--- a/client/src/components/Draftail/blocks/MediaBlock.js
+++ b/client/src/components/Draftail/blocks/MediaBlock.js
@@ -82,6 +82,7 @@ class MediaBlock extends Component {
     return (
       <button
         type="button"
+        tabIndex={-1}
         className="MediaBlock"
         onClick={this.openTooltip}
         data-draftail-trigger
--- a/client/src/components/Draftail/blocks/MediaBlock.js
+++ b/client/src/components/Draftail/blocks/MediaBlock.js
@@ -82,7 +82,6 @@ class MediaBlock extends Component {
     return (
       <button
         type="button"
-        tabIndex={-1}
         className="MediaBlock"
         onClick={this.openTooltip}
         data-draftail-trigger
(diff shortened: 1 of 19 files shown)


jordan-bauer:
@thibaudcolas 

Removed all instances of tabindex from wagtail repo.

client/src/components/Draftail/blocks/MediaBlock.js
```diff
@@ -82,7 +82,6 @@ class MediaBlock extends Component {
     return (
       <button
         type="button"
-        tabIndex={-1}
```
thibaudcolas: This one is intentional so I’ll add it back.

client/src/components/Explorer/PageCount.js
```diff
@@ -11,7 +11,6 @@ const PageCount = ({ page }) => {
     <a
       href={`${ADMIN_URLS.PAGES}${page.id}/`}
       className="c-explorer__see-more"
-      tabIndex={0}
```
thibaudcolas: Hmm I’m not entirely sure why we set that to start with!

docs/advanced_topics/customisation/admin_templates.rst
```diff
@@ -163,7 +163,7 @@ To add extra buttons to the login form, override the ``submit_buttons`` block. Y
 
     {% block submit_buttons %}
         {{ block.super }}
-        <a href="{% url 'signup' %}"><button type="button" class="button" tabindex="4">{% trans 'Sign up' %}</button></a>
+        <a href="{% url 'signup' %}"><button type="button" class="button">{% trans 'Sign up' %}</button></a>
```
thibaudcolas: 👍 good idea to update the docs to not use this pattern. It’s also pretty strange to have a `button` inside an `a` tag, but I’m not entirely sure what this is trying to demonstrate so won’t change that now.

PR #5146, opened 2019-03-14
add form media to user add/edit forms

CherryChopsticks:
solves #5133 By adding form media to user add/edit forms.

PR #5390, opened 2019-06-17
Add form media to users/edit.html

gasman:
When I merged #5146, I added `{{ form.media.css }}` / `{{ form.media.js }}` to the extra_css / extra_js blocks as those had been moved out of _editor_css.html / _editor_js.html in the meantime. Unfortunately, it seems that I missed that on users/edit.html... 🤦‍♂


gasman:
Merged in 77a3bf60cabfaeee1a50d7ab292fd7a531b146ef

PR #5392, opened 2019-06-17
Allow the tabs content to define the tab width on xs screens to stop cutting of text

kevinhowbrook:
Tabbed title on small screens is getting cut off due to the width setting:

tab eg

This PR sets the width as auto resulting in:

tab eg 2

### DIFF
--- a/client/scss/components/_tabs.scss
+++ b/client/scss/components/_tabs.scss
@@ -148,6 +148,6 @@
     }
 
     .tab-nav li {
-        width: 25%;
+        width: auto;
     }
 }


gasman:
Thanks @kevinhowbrook! I notice that the tabs now wrap on very narrow screens (<=350px):

Screenshot 2019-08-29 at 11 50 49

\- however, the current overlapping behaviour looks pretty bad at that size too, and given that this is a clear improvement for the 350-500ish pixel range, I'm happy to go with this.

gasman:
Merged in 48c82516dba5c01acd1ed6a67534a6d742ab434f

PR #5402, opened 2019-06-19
Add dedicated accessibility section to v2.6 release notes

thibaudcolas:
This moves all of the accessibility-related changes shipping in v2.6 into their own section, to give them a bit more visibility.

### DIFF
--- a/docs/releases/2.6.rst
+++ b/docs/releases/2.6.rst
@@ -10,6 +10,40 @@ Wagtail 2.6 release notes - IN DEVELOPMENT
 What's new
 ==========
 
+Accessibility targets and improvements
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


tomdyson:
Good idea @thibaudcolas! I think we should credit DIT, something like 

"We thank the UK government, who commissioned many of these improvements".

pvetch:
> "We thank the UK government, who commissioned many of these improvements".

Could we say the UK Government (in particular the CMS team at the Department for International Trade)

thibaudcolas:
👍 thank you for the feedback, that’s a very good idea.

I’ve added it as part of the second paragraph,

> Wagtail now has official accessibility support targets: we are aiming for compliance with WCAG2.1, AA level. WCAG 2.1 is the international standard which underpins many national accessibility laws.
> 
> Wagtail isn’t fully compliant just yet, but we have made many changes to the admin interface to get there. We thank the UK Government (in particular the CMS team at the Department for International Trade), who commissioned many of these improvements.
> 
> […]

I’ll merge this now, but we can still make further changes separately if anyone has more feedback.

tomdyson:
LGTM, thanks @thibaudcolas

PR #5412, opened 2019-06-25
Added missing is_stored_locally() to AbstractDocument

jonny5532:
Unless I'm missing something obvious, there's no definition of `is_stored_locally()` in AbstractDocument, so `open_file()` fails. The functionality seems to mirror AbstractImage, so I have copied the implementation of `is_stored_locally()` from there and it now works properly.


gasman:
Thanks @jonny5532! Merged in 851bd5f24d98188ffc6ac0c83fdd62f9c666a6de.

PR #5415, opened 2019-06-27
Datetime picker dispatches 'change' event

mike-hearn:
Fixes https://github.com/wagtail/wagtail-react-streamfield/issues/41

This issue was caused because the date/time picker, when changing the value of the input field it was binded to, does not automatically fire a `change` event when the data is changed.

The react-streamfield plugin only [updates the component's value](https://github.com/wagtail/react-streamfield/blob/198f702a89e3ec32a83b79adeaec94309378cb79/src/RawHtmlFieldInput.js#L96-L110) specifically on the `change`, so this PR manually dispatches `change` using the date picker plugin's [built-in](https://xdsoft.net/jqplugins/datetimepicker/) `onChangeDateTime` method.

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/js/date-time-chooser.js
+++ b/wagtail/admin/static_src/wagtailadmin/js/date-time-chooser.js
@@ -30,14 +30,20 @@ function initDateChooser(id, opts) {
             timepicker: false,
             scrollInput: false,
             format: 'Y-m-d',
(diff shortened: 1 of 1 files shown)


gasman:
Thanks @mike-hearn - looks good to me! Merged in 10bbfec93cd919f913b0b42cfa222a43151ad82e.

PR #5423, opened 2019-07-09
Fix update docs typo

tomdyson:
Only two Ss in accessor. Also make quotes consistent.


gasman:
👍 Merged to master and stable/2.6.x.

Issue #5074, opened 2019-02-12
Template fragment caching potentially harmful


harrislapiroff:
### Issue Summary

Django provides optional [template fragment caching](https://docs.djangoproject.com/en/2.1/topics/cache/#template-fragment-caching) wherein a portion of a template is cached when it is first rendered and that cache is reused to render that portion of the page until the cache expires or is invalidated.

Wagtail provides two ways of viewing pages that have unpublished content: "Preview" and "View Draft."

Used in conjunction, these features can create scenarios in which live data may be cached and then rendered in preview OR in which draft data may be cached and rendered on the live site. The former is annoying, the latter is a potential vulnerability that could inadvertently leak unpublished information 😱 

Since this behavior exists _because_ all the parts are behaving as designed, it's not 100% clear to me what the solution should be. I can imagine a few possible solutions off the top of my head:

1. Add a note to the documentation somewhere cautioning against the use of template fragment caching for this reason
2. Somehow add Wagtail-specific behavior to the Django `{% cache %}` tag that causes it to no-op in draft and preview modes (may be possible through middleware or the cache API? not sure)
3. Create a Wagtail-specific template fragment caching system that respects draft and preview modes

### Steps to Reproduce

Start with a Wagtail project with the following files:

`models.py`

```python
from django.db import models

from wagtail.core.models import Page
from wagtail.admin.edit_handlers import FieldPanel


class HomePage(Page):
    a_charfield = models.CharField(max_length=255)
    
    content_panels = Page.content_panels + [
        FieldPanel('a_charfield')
    ]
```

`templates/home/home_page.html`

```html+django
{% load cache %}

{% cache 3600 homepage_charfield %}
  {{ self.a_charfield }}
{% endcache %}
```

(Generate and run all migrations, create necessary admin user, etc.)

From the Wagtail admin, edit and publish a new Home Page:

![screen shot 2019-02-12 at 09 52 22](https://user-images.githubusercontent.com/296631/52644139-45188c00-2eac-11e9-8d14-3c9e8e6255f4.png)

DO NOT visit this page. Edit the page again with new values:

![screen shot 2019-02-12 at 09 56 09](https://user-images.githubusercontent.com/296631/52644374-b48e7b80-2eac-11e9-8550-511154ae0a33.png)

Do not save, but click "Preview." You'll see a previewed page that says "Privileged Information." Now, load up the live page. Observe that it _also_ renders "Privileged Information" even though that is unsaved content that an editor was merely testing.

[...]

Issue #4983, opened 2019-01-08
Provide for custom dummy_request field copying (similar to request.META)


yurman-cc:
### Issue Summary

cc: @GibbsConsulting
[django-plotly-dash](https://github.com/GibbsConsulting/django-plotly-dash) issue: #101

The problem arises from the experience integrating [django-plotly-dash](https://github.com/GibbsConsulting/django-plotly-dash) server-side app (the front end is based on React) with Wagtail StreamField.

Custom template tags used by django-plotly-dash store the names of JavaScript modules required by specific Dash apps on the page in the form of an object created by the django-plotly-dash middleware that is stored in the request object. When another template tag is called, it would insert a placeholder string in the HTML that would be subsequently detected and replaced with the actual links importing the required JavaScript modules by the middleware for all the apps on the page.

This mechanism is broken by the Wagtail dummy_request as it discards this custom request field required by the middleware.

There is already a way to get custom request.META fields copied as desired by overriding Page.dummy_request() and calling Page.default_preview_mode.update()

Can we have a similar way of specifying arbitrary request fields to copy over in the similar fashion by maybe extending the same dummy_request mechanism?

If not, can you please suggest the recommended way of solving this.

### Steps to Reproduce

This is not a bug, it's a design/middleware-compatibility issue.

You can find the integrated Dash app (with a few workarounds including the one for the problem described above) [here](https://xcms.helium.red/xcmsblog/2019/01/06/launching-data-new-orbit-belka-and-strelka-/)

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: (yes)

### Technical details

(The details below, except maybe the Wagtail version are not relevant)

* Python version: 3.6.7
* Django version: 2.1.4
* Wagtail version: 2.3.1
* Browser version: Firefox 52

yurman-cc:
bumping up

PR #5427, opened 2019-07-11
Middleware errors / responses during preview should be returned to the user

### LINKED ISSUE #3546: Preview mode doesn't work under WSGI/Gunicorn
### Issue Summary

Attempt to preview any page results in exception.

### Steps to Reproduce

1. Open any page in the WagTail CMS for editing
2. Click Preview button

Site is running under nginx/gunicorn. When I run Django development server, this problem doesn't happen, preview works as expected. This is clearly related to gunicorn, but when I run local instance of gunicorn, it doesn't happen either. Only on the remote site.

It seems that tricks that Wagtail uses to load preview in a new window causes the problem.

After reading Wagtail source code I found this:
```
        preview_mode = request.GET.get('mode', page.default_preview_mode)
        response = page.serve_preview(page.dummy_request(request), preview_mode)
        response['X-Wagtail-Preview'] = 'ok'
        return response
```
in `wagtail/wagtailadmin/views/pages.py`

Apparently, differences in requests under local django devserver and gunicorn causes some malfunction in the Django authentication middleware and it fails to load currently logged in user while in preview mode.

Since this happens only in WagTail preview and all other views and dependencies functioned perfectly well in this setup for years, I assume this is a Wagtail bug.

I have "fixed" this issue in my case by changing the line from:
`response = page.serve_preview(page.dummy_request(request), preview_mode)`
to:
`response = page.serve_preview(request, preview_mode)`
On line 562 in `wagtail/wagtailadmin/views/pages.py`

I see that master branch uses the same dummy request technique that probably will also fail when running under gunicorn.

### Technical details
```
AttributeError at /cms/pages/29/edit/preview/
'WSGIRequest' object has no attribute 'user'

Request Method: POST
Request URL: https://staging.royalorchard.com/cms/pages/29/edit/preview/
Django Version: 1.10.6
Python Executable: /home/rolc/.virtualenvs/dev/bin/python3
Python Version: 3.5.2
Python Path: ['/home/rolc/dev', '/home/rolc/.virtualenvs/dev/bin', '/home/rolc/.virtualenvs/dev/lib/python35.zip', '/home/rolc/.virtualenvs/dev/lib/python3.5', '/home/rolc/.virtualenvs/dev/lib/python3.5/plat-x86_64-linux-gnu', '/home/rolc/.virtualenvs/dev/lib/python3.5/lib-dynload', '/usr/lib/python3.5', '/usr/lib/python3.5/plat-x86_64-linux-gnu', '/home/rolc/.virtualenvs/dev/lib/python3.5/site-packages']
Server time: Fri, 14 Apr 2017 03:21:22 -0400
Installed Applications:
('django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.messages',
 'django.contrib.sessions',
 'django.contrib.sites',
 'django.contrib.humanize',
 'django.contrib.staticfiles',
 'django.contrib.admindocs',
 'django.contrib.flatpages',
 'tagging',
 'django_extensions',
 'filebrowser',
 'crispy_forms',
 'floppyforms',
 'allauth',
 'allauth.account',
 'allauth.socialaccount',
 'reversion',
 'papertrail',
 'wagtail.wagtailforms',
 'wagtail.wagtailredirects',
 'wagtail.wagtailembeds',
 'wagtail.wagtailsites',
 'wagtail.wagtailusers',
 'wagtail.wagtailsnippets',
 'wagtail.wagtaildocs',
 'wagtail.wagtailimages',
 'wagtail.wagtailsearch',
 'wagtail.wagtailadmin',
 'wagtail.wagtailcore',
 'wagtail.contrib.table_block',
 'modelcluster',
 'taggit',
 'taggit_templatetags2',
 'royal_orchard.users',
 'royal_orchard.reports',
 'royal_orchard.documents',
 'royal_orchard.emails',
 'royal_orchard.blog',
 'royal_orchard.reservations',
 'royal_orchard.vendors',
 'royal_orchard.flickr',
 'royal_orchard.search',
 'royal_orchard.userpages',
 'tinymce',
 'sorl.thumbnail',
 'constance',
 'constance.backends.database',
 'djmail',
 'bootstrap_pagination')
Installed Middleware:
('django.middleware.common.CommonMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.contrib.admindocs.middleware.XViewMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'wagtail.wagtailcore.middleware.SiteMiddleware',
 'wagtail.wagtailredirects.middleware.RedirectMiddleware')


Traceback:  

File "/home/rolc/.virtualenvs/dev/lib/python3.5/site-packages/django/core/handlers/exception.py" in inner
  42.             response = get_response(request)

File "/home/rolc/.virtualenvs/dev/lib/python3.5/site-packages/django/core/handlers/base.py" in _legacy_get_response
  249.             response = self._get_response(request)

File "/home/rolc/.virtualenvs/dev/lib/python3.5/site-packages/django/core/handlers/base.py" in _get_response
  217.                 response = self.process_exception_by_middleware(e, request)

File "/home/rolc/.virtualenvs/dev/lib/python3.5/site-packages/django/core/handlers/base.py" in _get_response
  215.                 response = response.render()

File "/home/rolc/.virtualenvs/dev/lib/python3.5/site-packages/django/template/response.py" in render
  109.             self.content = self.rendered_content

File "/home/rolc/.virtualenvs/dev/lib/python3.5/site-packages/django/template/response.py" in rendered_content
  86.         content = template.render(context, self._request)

File "/home/rolc/.virtualenvs/dev/lib/python3.5/site-packages/django/template/backends/django.py" in render
  66.             return self.template.render(context)

File "/home/rolc/.virtualenvs/dev/lib/python3.5/site-packages/django/template/base.py" in render
  206.                 with context.bind_template(self):

File "/usr/lib/python3.5/contextlib.py" in __enter__
  59.             return next(self.gen)

File "/home/rolc/.virtualenvs/dev/lib/python3.5/site-packages/debug_toolbar/panels/templates/panel.py" in _request_context_bind_template
  49.         context = processor(self.request)

File "/home/rolc/dev/royal_orchard/blog/context_processors.py" in context
  10.     user = request.user

Exception Type: AttributeError at /cms/pages/29/edit/preview/
Exception Value: 'WSGIRequest' object has no attribute 'user'
Request information:
USER: Robert ***********

GET: No GET data

POST:
seo_title = ''
expire_at = ''
csrfmiddlewaretoken = '******'
body-0-type = 'paragraph'
body-count = '1'
go_live_at = ''
body-0-deleted = ''
search_description = ''
body-0-order = '0'
slug = '******-history'
title = '****** History'
next = ''
body-0-value = '<p>I feel so very privileged to be invited into your home and if we also feel a little intimidated because we are talking to folks who have generational knowledge of this wonderful house and I know as we do so we are little bit nervous about giving you advice on wh</p>'
```

* Python version: 3.5.2.
* Django version: 1.10.6.
* Wagtail version: 1.9.
* Browser version: Firefox 52.0.2 on Ubuntu Linux.
* Gunicorn: 19.7.1
* Nginx: 1.10.0

gasman:
Fixes #3546; also related to #4983, #5074 and #5332.

As noted by @chosak in https://github.com/wagtail/wagtail/issues/5074#issuecomment-462892013 , the `Page.dummy_request` method used to generate request objects for preview has some very unpleasant hidden behaviour: in the process of applying middleware to the request, it ends up running the entire request/response cycle against the page's live URL (which may or may not be a valid URL at that point), discarding this 'shadow' response, and returning the request object (which is now populated with attributes from middleware, such as `request.user` and `request.site`) to be re-used a second time.

This leads to a class of horrible-to-track-down bugs as seen in #3546:

* some sort of misconfiguration (e.g. incorrect `ALLOWED_HOSTS`, or an authentication middleware that checks an HTTP header that we don't pass on to the dummy request) causes middleware to throw an error or prematurely return a response during preview;
* This middleware processing is wrapped in Django's global error handler, so even in the case of a Python error, this will get caught by Django and converted into a 500 HTTP response which promptly gets discarded just like any other response;
* `dummy_request` merrily returns a request object with incomplete middleware applied, which usually means `request.site` is missing;
* Some part of the page rendering then tries to access `request.site` and fails with an error that's completely unrelated to the root cause.

Ultimately the design of `dummy_request` is flawed, as it relies on us being able to apply middleware to the request independently of running a view function, which Django-1.10-style middleware fundamentally doesn't permit. The fix is to replace it with a new method, `make_preview_request`, which generates the dummy request, applies middleware and calls `serve_preview` as a single operation; that way, the view function at the end of the middleware chain is the `serve_preview` method that we're actually interested in. This way, there is no longer a bogus 'background' request and no response has to be thrown away (meaning that any response returned by middleware will be correctly returned).

### DIFF
--- a/wagtail/tests/middleware.py
+++ b/wagtail/tests/middleware.py
@@ -1,0 +1,8 @@
+from django.http import HttpResponseForbidden
+from django.utils.deprecation import MiddlewareMixin
+
+
+class BlockDodgyUserAgentMiddleware(MiddlewareMixin):
+    def process_request(self, request):
+        if not request.path.startswith('/admin/') and request.META.get('HTTP_USER_AGENT') == 'EvilHacker':
+            return HttpResponseForbidden("Forbidden")
--- a/docs/releases/2.7.rst
+++ b/docs/releases/2.7.rst
@@ -28,3 +28,8 @@ Bug fixes
 
 Upgrade considerations
 ======================
+
+``Page.dummy_request`` is deprecated
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The internal ``Page.dummy_request`` method (which generates an HTTP request object simulating a real page request, for use in previews) has been deprecated, as it did not correctly handle errors generated during middleware processing. Any code that calls this method to render page previews should be updated to use the new method ``Page.make_preview_request(original_request=None, preview_mode=None)``, which builds the request and calls ``Page.serve_preview`` as a single operation.
--- a/wagtail/admin/tests/test_pages_views.py
+++ b/wagtail/admin/tests/test_pages_views.py
@@ -1,6 +1,7 @@
 import datetime
 import logging
 import os
+import unittest
 from itertools import chain
 from unittest import mock
 
@@ -5262,6 +5263,19 @@ class TestDraftAccess(TestCase, WagtailTestUtils):
 
         # User can view
         self.assertEqual(response.status_code, 200)
+
+    @unittest.expectedFailure
+    def test_middleware_response_is_returned(self):
+        """
+        If middleware returns a response while serving a page preview, that response should be
+        returned back to the user
+        """
+        self.login()
+        response = self.client.get(
+            reverse('wagtailadmin_pages:view_draft', args=(self.child_page.id, )),
+            HTTP_USER_AGENT='EvilHacker'
+        )
+        self.assertEqual(response.status_code, 403)
 
 
 class TestPreview(TestCase, WagtailTestUtils):
--- a/wagtail/admin/tests/test_pages_views.py
+++ b/wagtail/admin/tests/test_pages_views.py
@@ -1,7 +1,6 @@
 import datetime
 import logging
 import os
-import unittest
 from itertools import chain
 from unittest import mock
 
@@ -5264,7 +5263,6 @@ class TestDraftAccess(TestCase, WagtailTestUtils):
         # User can view
         self.assertEqual(response.status_code, 200)
 
-    @unittest.expectedFailure
     def test_middleware_response_is_returned(self):
         """
         If middleware returns a response while serving a page preview, that response should be
--- a/wagtail/admin/views/pages.py
+++ b/wagtail/admin/views/pages.py
@@ -586,7 +586,7 @@ def view_draft(request, page_id):
     perms = page.permissions_for_user(request.user)
     if not (perms.can_publish() or perms.can_edit()):
         raise PermissionDenied
-    return page.serve_preview(page.dummy_request(request), page.default_preview_mode)
+    return page.make_preview_request(request, page.default_preview_mode)
 
 
 class PreviewOnEdit(View):
@@ -646,8 +646,7 @@ class PreviewOnEdit(View):
 
         form.save(commit=False)
         preview_mode = request.GET.get('mode', page.default_preview_mode)
-        return page.serve_preview(page.dummy_request(request),
-                                  preview_mode)
+        return page.make_preview_request(request, preview_mode)
 
 
 class PreviewOnCreate(PreviewOnEdit):
@@ -1055,11 +1054,9 @@ def preview_for_moderation(request, revision_id):
 
     page = revision.as_page_object()
 
-    request.revision_id = revision_id
-
-    # pass in the real user request rather than page.dummy_request(), so that request.user
-    # and request.revision_id will be picked up by the wagtail user bar
-    return page.serve_preview(request, page.default_preview_mode)
+    return page.make_preview_request(request, page.default_preview_mode, extra_request_attrs={
+        'revision_id': revision_id
+    })
 
 
 @require_POST
@@ -1180,7 +1177,7 @@ def revisions_view(request, page_id, revision_id):
     revision = get_object_or_404(page.revisions, id=revision_id)
     revision_page = revision.as_page_object()
 
-    return revision_page.serve_preview(page.dummy_request(request), page.default_preview_mode)
+    return revision_page.make_preview_request(request, page.default_preview_mode)
 
 
 def revisions_compare(request, page_id, revision_id_a, revision_id_b):
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -3,6 +3,7 @@ import logging
 from collections import defaultdict
 from io import StringIO
 from urllib.parse import urlparse
+from warnings import warn
 
 from django.conf import settings
 from django.contrib.auth.models import Group, Permission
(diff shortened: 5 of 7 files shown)


wagtail/tests/middleware.py
```diff
@@ -0,0 +1,8 @@
+from django.http import HttpResponseForbidden
+from django.utils.deprecation import MiddlewareMixin
+
+
+class BlockDodgyUserAgentMiddleware(MiddlewareMixin):
+    def process_request(self, request):
+        if not request.path.startswith('/admin/') and request.META.get('HTTP_USER_AGENT') == 'EvilHacker':
```
chosak: I had to stare at this for a while to understand how this was working [in the new test](https://github.com/wagtail/wagtail/blob/92e5bf9ab5ecbb9ca7276b09bbfa8b9229bd91f1/wagtail/admin/tests/test_pages_views.py#L5266) since that test is making an admin preview request to a path that _does_ start with `/admin/`. Eventually I figured out that what's being tested is the "path" of the page being _previewed_ (in this case `/hello-world/`).

I'd be curious about real world use cases where Wagtail users might encounter non-200 responses from preview middleware. For example, say you had a middleware that redirects under certain conditions, say if a page has certain configuration settings. If you preview a page in that case, your "preview view" would end up redirecting. This makes logical sense but does feel a bit unexpected; as a Wagtail user I might expect some kind of interstitial page indicating that the preview had redirected.

gasman: @chosak Yep, the `not request.path.startswith('/admin/')` needs to be there because the same HTTP headers are reused for the real request (to the admin view) and the faked request used to perform the preview, and we're testing the case where the latter triggers a middleware response. Can add a comment to clarify this if you think it'll help.

As far as I can see, the most common reason for middleware to return a non-200 response (or indeed to return early with its own 200 response rather than proceeding to the next middleware / view) is if the user has failed an authentication check - in this case it'll probably be clear to the user what has happened. Either way, returning that response as-is will definitely be an improvement over the current behaviour (i.e. forging ahead and attempting to serve the preview using the possibly-incomplete request object).

chosak: That makes sense, thanks. A comment might help to indicate why that clause is necessary.

chosak: That makes sense, thanks. A comment might help to indicate why that clause is necessary.

gasman:
Thanks @chosak! Now merged in ada652f19f4ef1a1fc4be22b9f264cc787d1eaee (with comment added to the test middleware, and deprecation warnings updated to `RemovedInWagtail29Warning`).

Issue #5428, opened 2019-07-11
Long listing titles should break on smaller screen sizes


kevinhowbrook:
### Issue Summary

On smaller screens, listing titles aren't being _broken_:

![bug eg](https://i.imgur.com/7OzWvo6.png)

This PR will break listing titles like so:

![fix eg](https://i.imgur.com/XOmmYWW.png)

### Steps to Reproduce

1. Start a new project with `wagtail start myproject`
2. Add a page with a really long title
3. In the admin page listing view, reduce the browser down to a smaller size.
* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: yes

### Technical details
* Python version: 3.5.2
* Django version: 2.1.8
* Wagtail version: 2.7.a0
* Browser version: Chrome 75

loicteixeira:
Closing this issue because the corresponding PR has been merged.

PR #5429, opened 2019-07-11
Break listing titles by word so action buttons are visible on smaller screens

kevinhowbrook:
#5428

On smaller screens, listing titles aren't being _broken_:

bug eg

This PR will break listing titles like so:

fix eg


gasman:
Looks good to me, thanks! Merged in 7db44c71b3abbbfb274373a7d0301255b9d3a87f.

PR #5437, opened 2019-07-17
Prevent exception when deleting a model with a protected 1-to-1 relation

nealtodd:
Modeladmin handles notification to the user if a model instance has protected ForeignKey
relationships and an attempt is made to delete the instance.

However, if the protected relation is a OneToOneField it raises an exception:

```
  File ".../wagtail/wagtail/contrib/modeladmin/views.py", line 742, in post
    for obj in qs.all():
AttributeError: 'MyRelatedModel' object has no attribute 'all'
```

because qs in this case is the related instance rather than a queryset of related instances
(which is the case for a ForeignKey relation).

This commit handles the OneToOneField case as well.

* Do the tests still pass?

Yes (`wagtail.contrib.modeladmin.tests`)

* Does the code comply with the style guide?

Yes

Yes


lb-:
@nealtodd  - thanks heaps.

@ababic  has approved this one, and this has now been merged in via https://github.com/wagtail/wagtail/commit/4a934246542cd5ad5364017381353a3329191c73

PR #5439, opened 2019-07-18
Changed Snippet Chooser AJAX delay from 50ms to 200ms, to remove laggy typing.

coredumperror:
50ms is the equivalent of about 200 words per minute, so typing slower than that meant that the javascript would send an AJAX request between every single keystroke, causing terrible lag. This change makes the javascript wait for 200ms between keystrokes, which lets you finish typing the word you're looking for before it sends an AJAX request.


gasman:
Thanks @coredumperror! Merged in ef4b37cf5dd3e4b998a65487c58b043842792f4f.

coredumperror:
Yay!

PR #5440, opened 2019-07-19
Fix the required styles on title fields

jafacakes2011:
I noticed that in the latest release candidates the required asterisk wasn't appearing for the title fields.

Wagtail 2.5.1:
Screenshot 2019-07-19 at 15 32 58

Wagtail 2.6rc1
Screenshot 2019-07-19 at 15 34 27

### DIFF
--- a/client/scss/components/_forms.scss
+++ b/client/scss/components/_forms.scss
@@ -133,7 +133,8 @@ li.focused > .help {
     opacity: 1;
 }
 
-.required .field > label:after {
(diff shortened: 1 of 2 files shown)


jafacakes2011:
There wasn't any reason other than that's where I tried and it fixed the issue I was having. I am more than happy to make the suggested change.

thibaudcolas:
👌 thanks! Yes I think I’d rather do the other change, feels a bit less risky

PR #5441, opened 2019-07-19
Pin Squash docker image to a specific version

thibaudcolas:
Recent builds have been breaking due to the following error: `/post_build.sh: line 2: npm: command not found`. I’m not entirely sure this is the fix, but it looks like this `latest` tag is now resolving to `3.7.4-buster` (https://hub.docker.com/_/python). Debian Buster got released 2 weeks ago. It feels safer to have a pinned dependency, and Stretch is probably what it was resolving to before.

### DIFF
--- a/.squash.yml
+++ b/.squash.yml
@@ -1,6 +1,6 @@
 deployments:
   default:
-    dockerimage: python:latest
+    dockerimage: python:3.7.4-stretch
     build_steps:
       - apt-get update && apt-get install -y libssl-dev libpq-dev git build-essential libfontconfig1 libfontconfig1-dev curl
       - RUN bash -c "curl -sL https://deb.nodesource.com/setup_8.x | bash -"


thibaudcolas:
For future reference, I’m not entirely sure whether’s a bug with the NodeSource node distributions, or whether it’s because of a change in Buster. As far I can see (https://github.com/nodesource/distributions), the distributions are meant to install both Node and npm, but the install script in Buster outputs the following:

```
## Run `sudo apt-get install -y nodejs` to install Node.js 8.x LTS Carbon and npm
[...]
Step 5/8 : RUN apt install -y nodejs
[...]
The following additional packages will be installed:
  libc-ares2 libnode64 libuv1 nodejs-doc
Suggested packages:
  npm
The following NEW packages will be installed:
  libc-ares2 libnode64 libuv1 nodejs nodejs-doc
```

I don’t get why `npm` is listed as a separate package.

PR #5453, opened 2019-07-25
Add https support for scribd oembed provider

math-a3k:
Currently scribd defaults to https for their sharing links - http still works but if you copy&paste quickly you may think it's broken :)


gasman:
Thanks @math-a3k! Merged in 1a2bf525f2b20f142a4c1076b9b57716514e7549.

math-a3k:
:) Thanks to you too!!

PR #5456, opened 2019-07-26
Pass down process_child_object when copying pages recursively

kaedroho:
This was overlooked when we implemented the feature.

Marked as release blocker as this feature was introduced in 2.6.


gasman:
LGTM - thanks @kaedroho!

PR #5458, opened 2019-07-26
Update developing.rst

easherma:
creates a separate section for describing how to link a local wagtail installation to a local project

### DIFF
--- a/docs/contributing/developing.rst
+++ b/docs/contributing/developing.rst
@@ -12,8 +12,8 @@ If you're happy to develop on a virtual machine, the `vagrant-wagtail-develop <h
 If you'd prefer to set up all the components manually, read on. These instructions assume that you're familiar with using pip and virtualenv to manage Python packages.
 
 
-Setting up the Wagtail codebase
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Setting up the Wagtail codebase & linking it to a project
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 Install Node.js, version 8. Instructions for installing Node.js can be found on the `Node.js download page <https://nodejs.org/download/>`_.
 You can also use Node version manager (nvm) since Wagtail supplies a ``.nvmrc`` file in the root of the project with the minimum required Node version - see nvm's `installation instructions <https://github.com/creationix/nvm>`_.
@@ -45,14 +45,7 @@ Compile the assets:
 
     $ npm run build
 
-Any Wagtail sites you start up in this virtualenv will now run against this development instance of Wagtail.  We recommend using the `Wagtail Bakery demo site <https://github.com/wagtail/bakerydemo/>`_ as a basis for developing Wagtail. Keep in mind that the setup steps for a Wagtail site may include installing a release version of Wagtail, which will override the development version you've just set up. In this case, you should install the site before running the ``pip install -e`` step, or re-run that step after the site is installed.
-
-Linking your local Wagtail codebase to a local project
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-You can also link your local Wagtail codebase to a prexisting project you may have using wagtail. 
-
-To do this, you'll want to remove the wagtail version specified in your requirements. 
+You will then need to link your local Wagtail codebase to a prexisting project you may have using wagtail. 
 
 Then, with your preferred virtualenv activated, install the Wagtail package in development mode with the included testing and documentation dependencies as a dependency in your project (testing or documentation dependencies could also be considered optional depending on your needs):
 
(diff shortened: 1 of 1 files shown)


docs/contributing/developing.rst
```diff
@@ -53,6 +47,24 @@ Compile the assets:
 
 Any Wagtail sites you start up in this virtualenv will now run against this development instance of Wagtail.  We recommend using the `Wagtail Bakery demo site <https://github.com/wagtail/bakerydemo/>`_ as a basis for developing Wagtail. Keep in mind that the setup steps for a Wagtail site may include installing a release version of Wagtail, which will override the development version you've just set up. In this case, you should install the site before running the ``pip install -e`` step, or re-run that step after the site is installed.
 
+Linking your local Wagtail codebase to a local project
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can also link your local Wagtail codebase to a prexisting project you may have using wagtail.
```
tomdyson: prexisting -> pre-existing
wagtail -> Wagtail

docs/contributing/developing.rst
```diff
@@ -53,6 +47,24 @@ Compile the assets:
 
 Any Wagtail sites you start up in this virtualenv will now run against this development instance of Wagtail.  We recommend using the `Wagtail Bakery demo site <https://github.com/wagtail/bakerydemo/>`_ as a basis for developing Wagtail. Keep in mind that the setup steps for a Wagtail site may include installing a release version of Wagtail, which will override the development version you've just set up. In this case, you should install the site before running the ``pip install -e`` step, or re-run that step after the site is installed.
 
+Linking your local Wagtail codebase to a local project
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can also link your local Wagtail codebase to a prexisting project you may have using wagtail. 
+
+To do this, you'll want to remove the wagtail version specified in your requirements.
```
tomdyson: you'll want to -> you'll need to
wagtail -> Wagtail

docs/contributing/developing.rst
```diff
@@ -53,6 +47,24 @@ Compile the assets:
...
+
+To do this, you'll want to remove the wagtail version specified in your requirements. 
+
+Then, with your preferred virtualenv activated, install the Wagtail package in development mode with the included testing and documentation dependencies as a dependency in your project (testing or documentation dependencies could also be considered optional depending on your needs):
+
+.. code-block:: console
+
+    $ pip install -e '.[testing,docs]' -U
+
+If you're using pipenv, the command is slighlty different:
```
tomdyson: slighlty -> slightly

docs/contributing/developing.rst
```diff
@@ -27,12 +27,6 @@ Clone a copy of `the Wagtail codebase <https://github.com/wagtail/wagtail>`_:
     $ git clone https://github.com/wagtail/wagtail.git
     $ cd wagtail
 
-With your preferred virtualenv activated, install the Wagtail package in development mode with the included testing and documentation dependencies:
```
tomdyson: Don't we still want this section here?

willbarton: I would think so — thought it does make me think that perhaps where we mention "virtualenv" we should link to something like [the Python virtual environment docs](https://docs.python.org/3/tutorial/venv.html). It might also be a good idea to say "virtual environment" (like we do in the [README](https://github.com/wagtail/wagtail#getting-started) and the [Getting Started documentation](http://docs.wagtail.io/en/v2.5.1/getting_started/index.html#quick-install)). That way someone's not left wondering what that means.

tomdyson: +1 on both points

easherma: I moved it back to be under the same section

docs/contributing/developing.rst
```diff
@@ -51,7 +45,20 @@ Compile the assets:
 
     $ npm run build
 
-Any Wagtail sites you start up in this virtualenv will now run against this development instance of Wagtail.  We recommend using the `Wagtail Bakery demo site <https://github.com/wagtail/bakerydemo/>`_ as a basis for developing Wagtail. Keep in mind that the setup steps for a Wagtail site may include installing a release version of Wagtail, which will override the development version you've just set up. In this case, you should install the site before running the ``pip install -e`` step, or re-run that step after the site is installed.
+You will then need to link your local Wagtail codebase to a pre-existing project you may have using Wagtail.
```
gasman: This sentence seems a bit out of place here. We haven't explained what we mean by 'linking' - in reality it just means installing into the same virtualenv - and it's not obvious that the following `pip install` commands are the steps for doing that. It's a bit like saying "Instructions for frobnicating your flux capacitor: first, you should frobnicate your flux capacitor. Then, run the `frobnicate` command..." :-)

If the intent of this rewrite is to guide people through the difference between "running a Wagtail site with the Wagtail package code buried somewhere in `site-packages`" vs "running a Wagtail site against a working development copy of Wagtail", then I think it's counter-productive to talk about 'linking', as that's introducing new jargon that obscures the concepts they know already. I think the most helpful thing we can do is move the mention of bakerydemo to the top of the instructions, so we don't have the awkward situation where they've just built Wagtail and now need to follow the bakerydemo installation instructions while making sure they don't override that Wagtail codebase with a packaged version. The wording could be something like:

> When working on Wagtail's code, you'll need to have a running Wagtail project to test your changes against. We recommend using the Wagtail Bakery demo site for this, but it can be any pre-existing project. Once you have this installed and running as normal inside a virtual environment, we can go ahead and set up a working copy of the Wagtail codebase to be used instead of the pre-packaged Wagtail version. _(now continue with the node / git setup steps...)_

docs/contributing/developing.rst
```diff
@@ -51,7 +45,20 @@ Compile the assets:
...
+
+Then, with your preferred `virtual enviornment <https://docs.python.org/3/tutorial/venv.html>`_ activated, install the Wagtail package in development mode with the included testing and documentation dependencies as a dependency in your project (testing or documentation dependencies could also be considered optional depending on your needs):
```
gasman: Typo here (and in other places) - enviornment -> environment. Also, we only need to include the link on the first mention of virtual environments. (We started by saying "you should be familiar with them", after all!)

thibaudcolas:
Replaced by  #8304.

PR #5460, opened 2019-07-26
Make tabs more accessible with ARIA roles

### LINKED ISSUE #5330: Tabbed UI in editing page are not marked up as tabs for screen readers
In the editing UI, Wagtail makes it possible to use panels to separate page fields into tabs. Unfortunately those tabs are not marked up as such for screen reader users (`tab` / `tablist` / `tabpanel` ARIA roles).

wagtail-edit-tabs

harrislapiroff:
These changes improve existing ARIA attributes on tabbed interfaces in the Wagtail admin and replace the existing tab Javascript with a new script that updates those attributes accordingly.

One note on a particular quirk of this PR: In general, according to the ARIA spec, tabs should be direct children of their tablist. To avoid restructuring the HTML in this case, I have used the aria-owns property to get around this requirement and the presentation role to prevent the intermediate list items from showing up as tabs.

Resolves #5330
Addresses #5252

# Review Notes

Requesting special attention in review to these items:

* Is the javascript OK style-wise?
* Is the javascript sufficient? I've written a pretty simple and (I think) clear tab component, but I'm worried that it's too simple and there are edge cases or accessibility needs that I haven't accounted for.
* There's a decision to be made on whether or not we pre-seed the HTML with attributes that will be filled in by the javascript and I'm not sure I've done it consistently in this PR. For example, should we add `class="active"` to the HTML for the first tab in a tabbed interface even though the javascript will automatically add that if we leave it out? Ditto for `aria-selected` and maybe even `aria-owns` and `aria-controls`. I could simplify the HTML templates a lot by completely relying on the javascript for those attributes, but I'm not 100% sure whether that's the right call or not.
* I'm not 100% clear on which browsers Wagtail supports. I have not tested in any Microsoft browsers.

# Checklist

- [x] For Python changes: Have you added tests to cover the new/fixed behavior?
   * Updated existing tests
- For front-end changes: Did you test on all of Wagtail’s supported browsers?
   * Firefox 68.0.1 (macOS)
   * Safari 12.1.1 (macOS)
   * Chrome 75.0.3770.142 (macOS)


harrislapiroff:
JUST KIDDING THIS IS STILL WIP BECAUSE I BROKE MODAL TABS

harrislapiroff:
THESE ONES

Screen Shot 2019-07-26 at 3 34 17 PM

harrislapiroff:
Okay I added the changes to the images and documents modals. I'm seeing tabs in a few other places as well:

- [ ] `wagtail/users/templates/wagtailusers/users/create.html`
- [ ] `wagtail/users/templates/wagtailusers/users/edit.html`
- [ ] `wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html`

Aiming to wrap those up + test fixes tomorrow.

client/src/includes/tabs.js
```diff
@@ -0,0 +1,118 @@
...
+        }
+      } else {
+        // Remove active class and ARIA attribute from any inactive tab
+        tab.classList.remove(this.options.tabActiveClass);
+        tab.setAttribute('aria-selected', 'false');
+        // Hide inactive panes
+        pane = document.getElementById(tab.getAttribute('aria-controls'));
+        if (pane) {
+          pane.classList.remove(this.options.paneActiveClass);
+          pane.hidden = true;
```
thibaudcolas: Is this equivalent to a `pane.setAttribute('hidden', 'true');`?

client/src/includes/tabs.js
```diff
@@ -0,0 +1,118 @@
...
+        }
+      }
+    });
+  }
+}
+
+const initDefaultTabs = () => {
```
thibaudcolas: What does the `default` in the function name stand for?

wagtail/admin/static_src/wagtailadmin/js/core.js
```diff
@@ -173,46 +173,6 @@ $(function() {
...
-                    if (!$(relTarg).parents().is($dropdown)) {
-                        $dropdown.removeClass('open');
-                        $(document).off('click.dropdown.cancel');
-                    }
-                });
-            } else {
-                $(document).off('click.dropdown.cancel');
-            }
-        });
-    });
```
thibaudcolas: Goodbye jQuery and Bootstrap!

wagtail/admin/static_src/wagtailadmin/js/vendor/bootstrap-tab.js
```diff
@@ -1,138 +0,0 @@
...
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- * ======================================================================== */
-
-
-+function ($) { "use strict";
```
thibaudcolas: I’m so glad we can remove all of this! @harrislapiroff did you spend much time looking at this implementation to make sure the new one matches, or was that not worth it?

wagtail/admin/templates/wagtailadmin/edit_handlers/tabbed_interface.html
```diff
@@ -1,8 +1,8 @@
 {% load wagtailadmin_tags %}
-<ul class="tab-nav merged" role="tablist">
+<ul class="tab-nav merged" role="tablist" aria-owns="{% for child in self.children %}tab-label-{{ child.heading|cautious_slugify }} {% endfor %}">
```
thibaudcolas: We had a quick chat about potentially turning this into `div` with the `a` right within the container – I see that in the end you’ve opted for `aria-owns`, does that work better / equally as well?

wagtail/admin/templates/wagtailadmin/edit_handlers/tabbed_interface.html
```diff
@@ -1,8 +1,8 @@
...
     {% for child in self.children %}
-        <li class="{{ child.classes|join:" " }} {% if forloop.first %}active{% endif %}" role="tab" aria-controls="tab-{{ child.heading|cautious_slugify }}">
-            <a href="#tab-{{ child.heading|cautious_slugify }}" class="{% if forloop.first %}active{% endif %}">{{ child.heading }}</a>
+        <li role="presentation" {% if child.classes %}class="{{ child.classes|join:" " }}"{% endif %}>
```
thibaudcolas: I thought `role="presentation"` would prevent content within this from being displayed?

harrislapiroff: Not in the testing I did with VoiceOver and Safari. I made this choice after using Firefox's accessibility inspector and seeing these results. Without `role="presentation"` on the `li`s (notice the three `nothing` items):

Screen Shot 2019-07-27 at 8 25 00 PM

With `role="presentation"` on the `li`s:
  
Screen Shot 2019-07-27 at 8 24 24 PM

As far as I can tell though, testing with VoiceOver, there's no behavioral difference I observed.

client/src/includes/tabs.js
harrislapiroff: I wanted to leave open the possibility that we might eventually want to call (or allow third-party devs to call) this on tabs that don't have the `tab-nav` class on it, so I'm considering the `tab-nav` tabs the default ones. No strong attachment to the naming here though.

wagtail/admin/static_src/wagtailadmin/js/vendor/bootstrap-tab.js
harrislapiroff: I didn't spend a lot of time checking it bit for bit, but the tabs _seem_ to be working the same? 😬

client/src/includes/tabs.js
```diff
@@ -0,0 +1,75 @@
...
+    this.tabs.forEach(tab => {
+      let pane;
+      if (tab === this.activeTab) {
+        // Add active class and ARIA attribute to active tab
+        tab.classList.add(this.options.tabActiveClass);
+        tab.setAttribute('aria-selected', 'true');
+        // Show active pane
+        pane = document.getElementById(tab.getAttribute('aria-controls'));
+        pane.classList.add(this.options.paneActiveClass);
+        pane.hidden = false;
```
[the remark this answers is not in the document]
harrislapiroff: Yes. I have no idea why certain attributes are exposed bare to the DOM and others aren't, but https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden

Looks like it might not have IE (pre-Edge) support though, if that's a concern. The CSS classes should provide an adequate fallback, though I haven't tested.

harrislapiroff:
List of browsers is above. As far as screenreaders, I only tested VoiceOver with Safari (sort of clumsily since I'm not used to it.

I read through a lot of blog posts and examples of how to do accessible tabs, so I feel _relatively_ confident in this implementation, but it's the first real accessibility feature I've worked on and (as mentioned above) I'm pretty clumsy with a screenreader, so it's definitely possible I missed some things.

wagtail/admin/templates/wagtailadmin/edit_handlers/tabbed_interface.html
harrislapiroff: This seems like the relevant documentation. It's a little confusing to me, but it seems to be saying descendant elements only inherit the `presentation` role in special cases that don't apply here https://www.w3.org/TR/wai-aria-practices-1.1/#presentation_role_effects

Happy to defer to your judgment though on whether it's preferable to leave in, take out, of restructure the HTML and CSS 😵

wagtail/admin/templates/wagtailadmin/edit_handlers/tabbed_interface.html
harrislapiroff: Restructuring the CSS was going to be a lot of work, so I tried to kick that can down the road. 😝 As far as I can tell `aria-owns` works equally well, but it has more template writing overhead, of course. I could be open to being pressured into restructuring the HTML and CSS here, if you wanted a more thorough PR.

harrislapiroff:
Before I make revisions, curious if you have any opinions on this question, @thibaudcolas:

> There's a decision to be made on whether or not we pre-seed the HTML with attributes that will be filled in by the javascript and I'm not sure I've done it consistently in this PR. For example, should we add class="active" to the HTML for the first tab in a tabbed interface even though the javascript will automatically add that if we leave it out? Ditto for aria-selected and maybe even aria-owns and aria-controls. I could simplify the HTML templates a lot by completely relying on the javascript for those attributes, but I'm not 100% sure whether that's the right call or not.

client/src/includes/tabs.js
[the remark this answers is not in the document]
thibaudcolas: 👍 I’ll check.

wagtail/admin/static_src/wagtailadmin/js/vendor/bootstrap-tab.js
thibaudcolas: 😬 great!

wagtail/admin/templates/wagtailadmin/edit_handlers/tabbed_interface.html
thibaudcolas: Cool. I’ll need to check what the recommendations are for tab elements, basically wondering whether lists are appropriate to start with in tab markup.

wagtail/admin/templates/wagtailadmin/edit_handlers/tabbed_interface.html
thibaudcolas: Ah I see, I didn’t realise this wasn’t inherited.

thibaudcolas:
Thank you @harrislapiroff, I’ve just done a quick review to make sure you weren’t blocked and will need to spend more time on this.

For "whether to add attributes via JS only" or not, this seems ok to me, since Wagtail requires JS to work.

Two more questions:

- How much work do you think it would be to refactor the HTML/CSS for the tab list?
- Could you add links here to articles you’ve used as references for this? I’ll need to read through them as well 😄 

Two questions:

harrislapiroff:
I’ve updated the javascript and HTML to simplify the markup required and enable the javascript to take responsibility for more of the accessibility attributes. I also added keyboard shortcuts and tabindex manipulation as recommended by Mozilla.

Hard for me to estimate how long refactoring the HTML/CSS would take, but I’ve left the javascript pretty independent from it, so I’d prefer to do that in a separate PR unless the usage of  `aria-owns` and `role=“presentation”` is objectionable.

I mostly referred to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role though I was a little muddy on their usage of tabindex and whether arrow keyboard control was required, though I eventually opted to do both.

I compared what I saw in the Firefox dev tools Accessibility panel for Wagtail pages to what I saw in the MDN example and found them similar. I also checked against the examples here https://a11yproject.com/patterns/#accordions-and-tabs

I got started on writing some Jest tests for my new `TabInterface` component, but I ran into some trouble. I'll elaborate in a line comment.

client/src/includes/tabs.test.js
```diff
@@ -0,0 +1,96 @@
...
+    expect(tab2.classList.contains('active')).toEqual(true);
+    expect(tab2.getAttribute('aria-selected')).toEqual('true');
+    // Ensure properties are changed correctly on the panels
+    expect(panel1.classList.contains('active')).toEqual(false);
+    expect(panel1.hasAttribute('hidden')).toEqual(true);
+    expect(panel2.classList.contains('active')).toEqual(true);
+    expect(panel2.hasAttribute('hidden')).toEqual(false);
+  });
+
+  it('changes to next tab on right arrow press', () => {
```
harrislapiroff: For some reason this test fails not because the keydown doesn't work, but because at this point `tab2` is already selected to start with, despite the fact that I am trying to reset the DOM in `afterEach` every time. I can't figure out why, but somehow the state of which tab is selected seems to be persisting between tests! Any guidance would be welcome.

harrislapiroff:
I will also need to update python tests to match new HTML.

client/src/includes/tabs.test.js
thibaudcolas: Would this be because your code resets then HTML and re-initialises the tab behavior, but doesn’t reset `window.location.hash`?

client/src/includes/tabs.test.js
```diff
@@ -0,0 +1,96 @@
...
+    tab1 = document.getElementById('t1-label');
+    tab2 = document.getElementById('t2-label');
+    panel1 = document.getElementById('t1');
+    panel2 = document.getElementById('t2');
+    tabInterface = new TabInterface(tabList);
+  });
+
+  afterEach(() => {
+    document.getElementsByTagName('html')[0].innerHTML = '';
+    tabInterface = null;
```
thibaudcolas: Since this is also reset in the `beforeEach`, I don’t think the cleanup here is necessary.

harrislapiroff:
Okay, @thibaudcolas, this is ready for review again.

client/src/includes/tabs.js
```diff
@@ -0,0 +1,189 @@
...
+
+    // If there's a tab that already has the active class, activate it instead
+    const activeClassTab = this.el.querySelector(`${this.options.tabLinkSelector}.${this.options.tabActiveClass}`);
+    if (activeClassTab) {
+      this.activeTab = activeClassTab;
+    }
+
+    // If there's a tab that matches the location hash, activate it instead
+    if (window.location.hash) {
+      const activeHashTab = this.el.querySelector(`[href="${window.location.hash}"]`);
```
thibaudcolas, line 51: Note for future reference that the change in https://github.com/wagtail/wagtail/pull/6155 will _not_ need to be implemented here, as this isn’t using jQuery.

But it might still be worth considering "what happens if the location hash is set to something that exists but isn’t a tab" nonetheless.

thibaudcolas:
Superseded by #8266

PR #5467, opened 2019-07-27
Adds new image for accessible logo background

### LINKED ISSUE #1947: Not enough contrast between logo and background in admin
screenshot from 2015-11-20 01 58 12

The beak is barely visible on the dark background.

brianedelman:
This adds a teal background image to make the logo in the admin more accessible. This reconfigures some of the HTML to make this happen in a consistent way as well.
Fixes #1947 

The 404 page also has an animated logo which is the same as it was before but could use QA since I had to modify (go to '/admin/whateveryouwant' to trigger).

Attaching screenshots for branding sign off.

Tested on Safari, Chrome, and Firefox

Screen Shot 2019-07-27 at 10 20 39 AM
Screen Shot 2019-07-27 at 10 20 30 AM
Screen Shot 2019-07-27 at 10 52 50 AM


tomdyson:
@benenright could you let us know what you think of this, please?

wagtail/admin/static_src/wagtailadmin/images/logo-background.svg
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
brianedelman: Yeah I missed that from the svg Andrea had given me, thanks!

wagtail/admin/static_src/wagtailadmin/images/logo-background.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="80px" height="80px" viewBox="0 0 80 80" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <circle id="Oval" fill="#007D7E" cx="40" cy="40" r="40"></circle>
```
thibaudcolas:
```suggestion
    <circle fill="#007D7E" cx="40" cy="40" r="40"></circle>
```

thibaudcolas: This seems to only draw a circle. Couldn’t we do this with CSS instead?

wagtail/admin/static_src/wagtailadmin/js/core.js
```diff
@@ -115,7 +115,7 @@ $(function() {
     function initLogo() {
         var sensitivity = 8; // the amount of times the user must stroke the wagtail to trigger the animation
 
-        var $logoContainer = $('.wagtail-logo-container__desktop');
+        var $logoContainer = $('.animated-logo-container');
```
thibaudcolas:
```suggestion
        var $logoContainer = $('[data-animated-logo-container]');
```

Since this class is only used in JS, I would suggest we refactor this to a data- attribute.

wagtail/admin/templates/wagtailadmin/base.html
```diff
@@ -8,11 +8,14 @@
...
-                        <img class="wagtail-logo wagtail-logo__full" src="{% static 'wagtailadmin/images/wagtail-logo.svg' %}" alt="Wagtail" width="80" />
+                        <img class="wagtail-logo-container__background" src="{% static 'wagtailadmin/images/logo-background.svg' %}" alt=""/>
+                        <div class="wagtail-logo-container-inner">
+                          <img class="wagtail-logo wagtail-logo__full" src="{% static 'wagtailadmin/images/wagtail-logo.svg' %}" alt="Wagtail" width="80" />
+                        </div>
                     </div>
 
                     {# Desktop logo (animated): #}
-                    {% include "wagtailadmin/shared/animated_logo.html" %}
+                    {% include "wagtailadmin/shared/animated_logo_background.html" %}
```
thibaudcolas: Could we use a partial that takes a parameter (`with background=True`), or only have the one partial and add the background around it? The two variants are too similar to justify the duplication in two separate files I think.

benenright:
Hey

I prefer the use of the new Teal rather than the old, lighter one. Although the lighter one gives more contrast to the bird, it has the effect of making our new accessible teal feel drab in direct comparison. 

But I have mocked a couple of other options below, and I feel like the mid-grey version is a decent option as it gives good contrast to the bird without becoming too prominent. The full white one is definitely too distracting for an ever-present UI feature.

(I did try some other background shapes/options but nothing felt better than the circle.

I did also think that we could shift the search and menu down by about 20px to allow the circle logo some sensible breathing space - and it feels tidy aligning the search more with the tabs.

So my preference is the mid-grey version - but I might be convinced by the darker teal one if there is appetite for it :)

image (5)
image (4)

Ben

PR #5469, opened 2019-07-28
Delay dirty form check’s snapshot to avoid race conditions. Fix #4978

### LINKED ISSUE #4978: Page edit view incorrectly reporting dirty rich text
### Issue Summary

When editing a page that contains Draftail rich text (either as a RichTextField or as a RichTextBlock in a StreamField), navigating away sometimes presents a warning that the page has been changed, even though it hasn't.

This happens for me consistently in Firefox 64, although I haven't been able to reproduce in other browsers. Specifically what is happening is this:

1. The JSON representing the rich text is generated in a nicely formatted way, with line breaks and indents:

    https://github.com/wagtail/wagtail/blob/882f8f3cf8ddd79c30e611a48882b309e90dad0c/wagtail/admin/rich_text/converters/contentstate.py#L87

2. The page edit view HTML contains this JSON as the `<input value>`, including those line breaks and indents:

    image

    image

3. The dirty form check's `initialData` (generated via `$form.serialize()`) gets set with a serialized version of that initial `value`, including line breaks:

    https://github.com/wagtail/wagtail/blob/882f8f3cf8ddd79c30e611a48882b309e90dad0c/wagtail/admin/static_src/wagtailadmin/js/core.js#L59

4. Certain seemingly innocuous events cause the `<input value>` to change, removing the line breaks and indents:

    image

    I can make this happen by highlighting the `<input>` in Firefox Developer Tools and (less predictably) by clicking away to another tab and coming back. Some event seems to cause the normalization of this value to clean it up.

    At first I thought this might be some Firefox quirk around handling of special characters within `<input value>`, but I couldn't reproduce it using some manually generated HTML with the same content. So my guess is this is somehow related to how/when Draftail/Draft.js monitors/updates its state, but I don't know enough about those internals to pinpoint it further.

5. Because the `<input value>` has been changed, the subsequent dirty form check here always fails:

    https://github.com/wagtail/wagtail/blob/882f8f3cf8ddd79c30e611a48882b309e90dad0c/wagtail/admin/static_src/wagtailadmin/js/core.js#L68

    This presents the user with a warning when navigating away, even though they haven't done anything.

### Steps to Reproduce

2. Create a simple page type in models.py:

    ```py
    class TestPage(Page):
        body = RichTextField()

            FieldPanel('body'),
        ]
    ```
3. Create an instance of this page, entering some content in the body field.
4. Save the page, then, after it reloads, try navigating away. You'll see a warning even though you didn't make any further changes.

A simple fix that works for me is to modify [this line](https://github.com/wagtail/wagtail/blob/882f8f3cf8ddd79c30e611a48882b309e90dad0c/wagtail/admin/rich_text/converters/contentstate.py#L87) to remove the nice formatting (just remove the `indent=4, separators=(',', ': ')`). This makes the JSON more normalized upon first load, and so subsequent refreshes don't change it. This does have the slight downside of making the page HTML less readable, and also still relies on the Python JSON formatting exactly matching what the browser regenerates.

Another option would be to add some kind of initial refresh _before_ `initialData` is saved in the dirty check, to ensure that whatever formatting Draftail/Draft.js wants to do to its content happens once before we save the reference state.

Ping @thibaudcolas - any idea what is happening here?

### Technical details

* Python version: 3.6.3
* Django version: 2.1.4
* Wagtail version: 2.5a0 (current master)
* Browser version: Firefox 64.0

thibaudcolas:
Fixes #4978. This adds a 10-sec delay to the initialisation of the "dirty form check" displaying a "Are you sure you want to leave this page?" message in the page editing UI.

This was broken because of a race condition – the dirty form check runs "when the page is done loading" rather than after the fields’ initialisation. The delay reduces the risk of race condition issues, and it also seems like a sensible behavior from an end user’s perspective not to aggressively prompt people who have only spent a few seconds on the page, even if they did change some content.

This race condition wasn’t a problem for most fields, but anything that stores JSON data structures (like the new rich text editor) is prone to differences in serialisation between server and client code – in that case whitespace, and perhaps keys ordering.

This change doesn’t alter the behavior of the dirty form check when it is initialised in "always prompt for confirmation" mode.

---

* [ ] For front-end changes: Did you test on all of Wagtail’s supported browsers?

Tested in latest Chrome, Firefox, Safari on macOS by:

- Opening an edit page then moving elsewhere right away – no confirmation
- Opening an edit page then moving elsewhere after the 10s delay – no confirmation
- Opening an edit page, editing content in less than 10s, moving elsewhere  – no confirmation
- Opening an edit page, editing content over more than 10s, moving elsewhere - confirmation

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/js/core.js
+++ b/wagtail/admin/static_src/wagtailadmin/js/core.js
@@ -57,16 +57,23 @@ function enableDirtyFormCheck(formSelector, options) {
     var $form = $(formSelector);
     var confirmationMessage = options.confirmationMessage || ' ';
     var alwaysDirty = options.alwaysDirty || false;
-    var initialData = $form.serialize();
+    var initialData = null;
     var formSubmitted = false;
 
     $form.on('submit', function() {
         formSubmitted = true;
     });
 
+    // Delay snapshotting the form’s data to avoid race conditions with form widgets that might process the values.
+    // User interaction with the form within that delay also won’t trigger the confirmation message.
+    setTimeout(function() {
+        initialData = $form.serialize();
+    }, 1000 * 10);
+
     window.addEventListener('beforeunload', function(event) {
+        var isDirty = initialData && $form.serialize() != initialData;
         var displayConfirmation = (
-            !formSubmitted && (alwaysDirty || $form.serialize() != initialData)
+            !formSubmitted && (alwaysDirty || isDirty)
         );
 
         if (displayConfirmation) {


thibaudcolas:
@tomdyson @briaguya could either of you review this? I’d be keen to further increase the delay from 10s to 30-60ish, but that would only be to improve the UX (further reduce the frequency of useless prompts) rather than fix the current bug.

tomdyson:
It's great to have a fix for this!

I've updated the Squash instance at https://thibaudcolasbug4978-dirty-form-fa58b.squash.io/admin/pages/3/edit/ (admin/changeme) to include a rich text field, and tested it on Firefox and Safari on OS X, where the fix is working as expected.

Note that changes which are added in the first 10 seconds will not be identified by the "dirty form check", even if you wait more than 10 seconds before leaving the page. I think this is OK.

wagtail/admin/static_src/wagtailadmin/js/core.js
```diff
@@ -57,16 +57,23 @@ function enableDirtyFormCheck(formSelector, options) {
     var $form = $(formSelector);
     var confirmationMessage = options.confirmationMessage || ' ';
     var alwaysDirty = options.alwaysDirty || false;
-    var initialData = $form.serialize();
+    var initialData = null;
```
briaguya: It might be nice to change this from null to an empty string so we can see that it’s going to be a string.

Definitely not a suggestion that has any impact on functionality.

gasman: To me `null` communicates "not yet defined" better than an empty string (which implies "this is defined and has empty content"), so I'm inclined to keep this as is.

gasman:
Merged in db9b582912808f0ef4ab16f3254c7ff57e769f3d (master) / 1108eb4c6e40b77f4a265514aff62dbe68da302a (stable/2.6.x) - will also merge into stable/2.5.x shortly to prepare a 2.5.2 release.

tomdyson:
Thanks @thibaudcolas and @gasman! It's a relief to have this fixed.

PR #5470, opened 2019-07-28
Fix jcrop holder image having no alt tag. Fix part of #5309

### LINKED ISSUE #5309: Alt text / labeling issues in image cropping UI
### Issue Summary

Related: #5291. Spotted while reviewing #5292. There are three accessibility issues picked up by Axe in the image cropping UI:

image-cropping

- jcrop generates its own `img` element. That element does not even have an alt text, so the image `src` is read out loud.
- The image tag rendered server-side uses the 
`title` attribute for the alt text. While this might be appropriate on the site’s front-end, here the image is displayed in the admin, to be edited, so the `title` text does not work as an `alt` in that context. Relevant code: https://github.com/wagtail/wagtail/blob/70a749b10a1b0c4dac527da28e54cbb3581a774e/wagtail/images/templates/wagtailimages/images/edit.html#L67
- jcrop injects its own `radio` element, which is flagged as not having a label

### Steps to Reproduce

1. Go to the image editing UI on any Wagtail admin, `/admin/images/52/`
2. Navigate to the cropping UI with a screen reader

### Technical details

* Wagtail version: v2.6a0

---

I think the simplest solution for this would be to make sure the image created with jcrop has an empty `alt=""` attribute, and same for our server-generated image. This might need further consideration for us to design a more screen-reader-friendly solution.

thibaudcolas:
Fixes part of #5309. I’ve looked at the [jcrop API docs](http://deepliquid.com/content/Jcrop_API.html), this seems to be the most correct solution according to their API – manipulating the holder image in Jcrop’s "ready" callback.

Tested in Chrome & Firefox on macOS, IE11.

This can easily be tested over at https://thibaudcolasbugjcrop-img-alt-5-yms7s.squash.io/admin/images/1/.

### DIFF
--- a/wagtail/images/static_src/wagtailimages/js/focal-point-chooser.js
+++ b/wagtail/images/static_src/wagtailimages/js/focal-point-chooser.js
@@ -24,6 +24,10 @@ function setupJcrop(image, original, focalPointOriginal, fields) {
         }
     }, function() {
         jcropapi = this
+
+        // Set alt="" on the image so its src is not read out loud to screen reader users.
+        var $holderImage = $('img', jcropapi.ui.holder);
+        $holderImage.attr('alt', '');
     });
 }
 


KalobTaulien:
Looks good @thibaudcolas ! Thanks for adding this!

PR #5474, opened 2019-07-30
Make image fallback work for images with big/small aspect ratios

### LINKED ISSUE #5472: Images with 2.56+:1 aspect ratios are distored in rich text editor
See the original report of this issue at https://github.com/springload/draftail/issues/191.

### Issue Summary

In the rich text editor, images are displayed as blocks. For images that have a very big or small aspect ratio, the image block’s "failed to load image" fallback styles are interfering and making images display with the wrong aspect ratio. This is because of: https://github.com/wagtail/wagtail/blob/f53dd956967f17b202d569ed1b45a5cf1d48e1c4/client/src/components/Draftail/blocks/MediaBlock.scss#L26-L30

The `min-width` and `min-height` will be problematic for images that fall below those values because of their aspect ratio

### Steps to Reproduce

1. Go to a rich text field with images enabled
2. Insert a new image that has an aspect ratio greater than 2.56:1, or lower than 1:2.56
3. See that the image renders, but distored compared to its original aspect ratio

The image’s aspect ratio should be preserved

---

I can see a few potential solutions to this:

- Easiest – change fallback min values so more extreme aspect ratios are needed (for example `min-height: 25px;`, so 10:1 aspect ratio)
- Alternatively, refactor the code so fallback styles are separate from normal rendering. I’m not sure whether this is simple though – the whole point of fallback styles is for them to consistently be there if the image does not load

imansyed000:
Fixes #5472 

Tested on Chrome with approx  7:1 aspect ratio

ldIxFVVIRuiysz5nVRzthQ_thumb_6

Went with this option under advice of Thibaud

### DIFF
--- a/client/src/components/Draftail/blocks/MediaBlock.scss
+++ b/client/src/components/Draftail/blocks/MediaBlock.scss
@@ -25,7 +25,8 @@
 
     @mixin invalid-image-fallback {
         min-width: 256px;
-        min-height: 25px;
+        min-height: 50px;
+        object-fit: contain;
         background-color: $color-grey-1;
     }
 
--- a/client/src/components/Draftail/blocks/MediaBlock.scss
+++ b/client/src/components/Draftail/blocks/MediaBlock.scss
@@ -25,7 +25,7 @@
 
     @mixin invalid-image-fallback {
         min-width: 256px;
-        min-height: 100px;
+        min-height: 25px;
         background-color: $color-grey-1;
     }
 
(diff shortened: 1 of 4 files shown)


client/src/components/Draftail/blocks/MediaBlock.scss
```diff
@@ -25,7 +25,7 @@
 
     @mixin invalid-image-fallback {
         min-width: 256px;
-        min-height: 100px;
+        min-height: 25px;
```
thibaudcolas: Checking what this looks like in the editor, I realise that 25px would be smaller than the image block’s top-right icon and not look right 😞

PR #5478, opened 2019-08-01
Add support for phone links in rich text

mikaelengstrom:
This is mainly an updated version of @SableWalnut´s (⭐) work  (https://github.com/wagtail/wagtail/pull/3776) but with a few fixes to make it work with Draftail and recent Wagtail releases. 

Since mailto:-links is supported i think it is reasonable to support tel:-links as well. The other PR (#3776) have been stale for almost 2 years now and I would really want to see this happen since almost every client I work with is requesting this feature and I don't really find any other convenient way to implement this than via wagtail core. So if there is anything I can do to help making this happen please let me know and I´ll do my best.


KalobTaulien:
Hi @mikaelengstrom 

This looks really good 🎉I just tested it locally and it works beautifully! There's just a few conflicts that need to be fixed up first and then I think it's good to be merged in!

mikaelengstrom:
@allcaps @KalobTaulien Thank you for testing it out. I have resolved the conflicts and also tested it out on localhost and it seems to work like a charm.

KalobTaulien:
Thanks for the awesome code @mikaelengstrom! 🎉🎉

PR #5479, opened 2019-08-01
Apply escaping to JS translation strings.

### LINKED ISSUE #5477: Error in Hungarian translation
The new versions have an annoying issue with the Hungarian translations: some of the translation strings are using double quotes instead of single quotes, which causes an arror on the admin page thanks to a faulty JS translation string:

```javascript 
EDIT_PAGE: ""{title}" szerkesztése",
```
The translations are now fixed on Transifex. Since I messed it up, I corrected it - sorry.

gasman:
Fixes #5477

Rather than inserting translations individually into the template via the `{% trans %}` tag (which fails to apply Javascript escaping), write out the full dictionary in one go via `json.dumps`.

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/admin_base.html
+++ b/wagtail/admin/templates/wagtailadmin/admin_base.html
@@ -27,66 +27,7 @@
                 EXTRA_CHILDREN_PARAMETERS: '',
             };
 
-            wagtailConfig.STRINGS = {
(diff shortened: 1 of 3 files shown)


gasman:
Merged in 051494252154b9708caec6f461375c5232f87ad9 (master) / e03fa25ae9f667c04415353a7e747e71c02ec797 (stable/2.6.x)

Issue #5484, opened 2019-08-06
Python 3.9 compatibility (resolve DeprecationWarning from collections)


therefromhere:
### Issue Summary

Wagtail hits the warning `DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working` in various places.

- [x] Jinja2>2.10.1 
- [x] html5lib>1.0.1 (see https://github.com/html5lib/html5lib-python/issues/419)
- [x] beautifulsoup4>=4.8.0 (or maybe earlier)
- [x] fix wagtail/utils/l18n/translation.py:5 (#5485)

### Steps to Reproduce

1. run `tox -e py37-dj22-sqlite-noelasticsearch -- --deprecation all`
2. note deprecation warnings:

```
site-packages/jinja2/utils.py:485: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
site-packages/jinja2/runtime.py:318: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
```
resolved by https://github.com/pallets/jinja/pull/867 - the fix is on master not released yet (as of Jinja2==2.10.1)

```
site-packages/html5lib/_trie/_base.py:3: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  from collections import Mapping
```

Fixed in https://github.com/html5lib/html5lib-python/issues/402 , but not released yet as of 1.0.1 (see https://github.com/html5lib/html5lib-python/issues/419 )

```
site-packages/bs4/element.py:1134: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
  if not isinstance(formatter, collections.Callable):
```
https://bugs.launchpad.net/beautifulsoup/+bug/1778909 - resolved by beautifulsoup4>=4.8.0 (and earlier I think)

I'm also seeing one in Wagtail from my own project tests here, the above tox run didn't seem to hit it:

```
wagtail/utils/l18n/translation.py:5: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
```

See #5485 

[...]

PR #5485, opened 2019-08-06
Resolve Py3.8 DeprecationWarning

therefromhere:
Resolves "DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working" - see #5484


gasman:
Thanks @therefromhere! Merged in 3bb2ab85fdd3780d076e37bf33ab11181c659cf4.

PR #5486, opened 2019-08-06
Support twitter previews in Draftail embeds. Fixes #5176

### LINKED ISSUE #5176: Support twitter previews in Draftail embeds

imansyed000:
Fixes #5176
Tested on Chrome and Safari. 
Added new tests

image

If no thumbnails available via oembed, returns 'Provider name - Author name'

### DIFF
--- a/client/src/components/Draftail/blocks/EmbedBlock.js
+++ b/client/src/components/Draftail/blocks/EmbedBlock.js
@@ -11,7 +11,7 @@ import MediaBlock from '../blocks/MediaBlock';
 const EmbedBlock = props => {
   const { entity, onRemoveEntity } = props.blockProps;
   const { url, title, thumbnail, providerName, authorName } = entity.getData();
-  
+
   if (thumbnail) {
     return (
       <MediaBlock {...props} src={thumbnail} alt="">
@@ -34,19 +34,17 @@ const EmbedBlock = props => {
     );
   }
 
-  else {
-    return (
-      <strong>
-        <a href={url} target={'_blank'}>
-          {providerName + ' - ' + authorName + '  '}
-        </a>
+  return (
+    <strong>
+      <a href={url} target={'_blank'}>
+        {providerName + ' - ' + authorName + '  '}
+      </a>
 
-        <button className="button button-secondary no Tooltip__button" onClick={onRemoveEntity} >
-          {STRINGS.DELETE}
-        </button>
-      </strong>
-    );
-  };
+      <button className="button button-secondary no Tooltip__button" onClick={onRemoveEntity} >
+        {STRINGS.DELETE}
+      </button>
+    </strong>
+  );
 };
 
 EmbedBlock.propTypes = {
--- a/client/src/components/Draftail/blocks/EmbedBlock.js
+++ b/client/src/components/Draftail/blocks/EmbedBlock.js
@@ -12,25 +12,52 @@ const EmbedBlock = props => {
   const { entity, onRemoveEntity } = props.blockProps;
   const { url, title, thumbnail } = entity.getData();
 
-  return (
-    <MediaBlock {...props} src={thumbnail} alt="">
-      {url ? (
-        <a
-          className="Tooltip__link EmbedBlock__link"
-          href={url}
-          title={url}
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          {title}
-        </a>
-      ) : null}
+  const username = url.split("/").slice()[3]
+  const profile_image = 'https://twitter.com/' + username + '/profile_image?size=original'
+  
+  if (thumbnail) {
+    return (
+      <MediaBlock {...props} src={thumbnail} alt="">
+        {url ? (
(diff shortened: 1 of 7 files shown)


client/src/components/Draftail/blocks/EmbedBlock.js
```diff
@@ -10,26 +10,40 @@ import MediaBlock from '../blocks/MediaBlock';
  */
 const EmbedBlock = props => {
   const { entity, onRemoveEntity } = props.blockProps;
-  const { url, title, thumbnail } = entity.getData();
+  const { url, title, thumbnail, providerName, authorName } = entity.getData();
+
+  if (thumbnail) {
```
thibaudcolas: Could you turn this and the other case around? I think I’d rather have `!thumbnail` as a special case, then keep the `return` at the end of the function be the "has thumbnail" scenario.

client/src/components/Draftail/blocks/EmbedBlock.test.js
```diff
@@ -15,6 +15,8 @@ describe('EmbedBlock', () => {
                 url: 'http://www.example.com/',
                 title: 'Test title',
                 thumbnail: 'http://www.example.com/example.png',
+                authorName: 'Example author',
+                providerName: 'Example provider',
```
thibaudcolas: Could you add a separate case for this test suite that specifically tests what this component renders when there is data, but not a thumbnail?

client/src/components/Draftail/blocks/EmbedBlock.js
thibaudcolas: Could you turn this and the other case around? I think I’d rather have `!thumbnail` as a special case, then keep the `return` at the end of the function be the "has thumbnail" scenario.

client/src/components/Draftail/blocks/EmbedBlock.js
```diff
@@ -10,26 +10,40 @@ import MediaBlock from '../blocks/MediaBlock';
...
-          title={url}
-          target="_blank"
-          rel="noopener noreferrer"
-        >
-          {title}
-        </a>
-      ) : null}
-
-      <button className="button button-secondary no Tooltip__button" onClick={onRemoveEntity}>
+    <strong>
```
thibaudcolas: I think I would still like this case to look visually similar to the `MediaBlock` component, perhaps overlaying the `providerName` & `authorName` text you have here on top of the existing block visual. Could you add a prop to `MediaBlock` for that, and then use the block with that extra prop?

client/src/components/Draftail/blocks/EmbedBlock.js
```diff
@@ -10,26 +10,40 @@ import MediaBlock from '../blocks/MediaBlock';
...
+      <a href={url} target={'_blank'}>
+        {providerName + ' - ' + authorName + '  '}
```
thibaudcolas: Should be written with ES6 template strings.

client/src/components/Draftail/blocks/EmbedBlock.js
```diff
@@ -10,26 +10,40 @@ import MediaBlock from '../blocks/MediaBlock';
...
+      </a>
+
+      <button className="button button-secondary no Tooltip__button" onClick={onRemoveEntity} >
```
thibaudcolas: I think this should stay within the tooltip, so the block’s UI isn’t distracting while entering content.

client/src/components/Draftail/blocks/EmbedBlock.js
imansyed000: Applied your suggestions. Don't know how to approach it in terms of styling but we can discuss when you're back :)

imansyed000: Preview looks like this

Screenshot 2019-08-16 at 09 49 55

imansyed000: Update: Just noticed this doesn't work on Chrome

thibaudcolas:
Closing this in favour of #5545.

PR #5488, opened 2019-08-06
Upgrade django-taggit to 1.x

gasman:
django-taggit 1.x drops Python 2.x support and thus the dependency on django.utils.six; this is a prerequisite for supporting Django 3.0.
The signature of TaggableManager.value_from_object has changed to return a list of Tags (previously it was a QuerySet of TaggedItems) and so search indexing and comparison need to be updated accordingly. There is a corresponding fix to ClusterTaggableManager in django-modelcluster 5.0.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,8 +22,8 @@ except ImportError:
 
 install_requires = [
     "Django>=2.0,<2.3",
-    "django-modelcluster>=4.2,<5.0",
-    "django-taggit>=0.23,<1.0",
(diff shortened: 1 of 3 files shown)


lb-:
Tested locally, tags work as expected after upgrade, thanks @gasman 

Note: I did not add anything to the upgrade considerations in the release notes for this.

merged in via https://github.com/wagtail/wagtail/commit/34c1b4ccbe4a143492ae534e95bf61a160104a9f

PR #5489, opened 2019-08-06
Django 3.0 fixes

gasman:
Hopefully get tests passing on Django master.

Incorporates #5488

### DIFF
--- a/wagtail/admin/tests/test_account_management.py
+++ b/wagtail/admin/tests/test_account_management.py
@@ -3,6 +3,7 @@ import tempfile
 
 import pytz
 
+from django import VERSION as DJANGO_VERSION
 from django.contrib.auth import views as auth_views
 from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Group, Permission
@@ -309,7 +310,10 @@ class TestAccountSection(TestCase, WagtailTestUtils):
 
         # Check that a validation error was raised
         self.assertTrue('new_password2' in response.context['form'].errors.keys())
-        self.assertTrue("The two password fields didn't match." in response.context['form'].errors['new_password2'])
+        if DJANGO_VERSION >= (3, 0):
+            self.assertTrue("The two password fields didn’t match." in response.context['form'].errors['new_password2'])
+        else:
+            self.assertTrue("The two password fields didn't match." in response.context['form'].errors['new_password2'])
 
         # Check that the password was not changed
         self.assertTrue(get_user_model().objects.get(pk=self.user.pk).check_password('password'))
@@ -692,7 +696,12 @@ class TestPasswordReset(TestCase, WagtailTestUtils):
         self.password_reset_uid = force_text(urlsafe_base64_encode(force_bytes(self.user.pk)))
 
         # Create url_args
-        self.url_kwargs = dict(uidb64=self.password_reset_uid, token=auth_views.INTERNAL_RESET_URL_TOKEN)
+        if DJANGO_VERSION >= (3, 0):
+            token = auth_views.PasswordResetConfirmView.reset_url_token
+        else:
+            token = auth_views.INTERNAL_RESET_URL_TOKEN
+
+        self.url_kwargs = dict(uidb64=self.password_reset_uid, token=token)
 
         # Add token to session object
         s = self.client.session
@@ -772,7 +781,11 @@ class TestPasswordReset(TestCase, WagtailTestUtils):
 
         # Check that a validation error was raised
         self.assertTrue('new_password2' in response.context['form'].errors.keys())
-        self.assertTrue("The two password fields didn't match." in response.context['form'].errors['new_password2'])
+
+        if DJANGO_VERSION >= (3, 0):
+            self.assertTrue("The two password fields didn’t match." in response.context['form'].errors['new_password2'])
+        else:
+            self.assertTrue("The two password fields didn't match." in response.context['form'].errors['new_password2'])
 
         # Check that the password was not changed
         self.assertTrue(get_user_model().objects.get(username='test').check_password('password'))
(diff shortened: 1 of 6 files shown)


wagtail/admin/tests/test_account_management.py
```diff
@@ -309,7 +310,10 @@ def test_change_password_view_post_password_mismatch(self):
 
         # Check that a validation error was raised
         self.assertTrue('new_password2' in response.context['form'].errors.keys())
-        self.assertTrue("The two password fields didn't match." in response.context['form'].errors['new_password2'])
+        if DJANGO_VERSION >= (3, 0):
+            self.assertTrue("The two password fields didn’t match." in response.context['form'].errors['new_password2'])
+        else:
+            self.assertTrue("The two password fields didn't match." in response.context['form'].errors['new_password2'])
```
zerolab: I might be missing something, but the two assertions are the same

wagtail/admin/tests/test_account_management.py
```diff
@@ -772,7 +781,11 @@ def test_password_reset_confirm_view_post_password_mismatch(self):
 
         # Check that a validation error was raised
         self.assertTrue('new_password2' in response.context['form'].errors.keys())
-        self.assertTrue("The two password fields didn't match." in response.context['form'].errors['new_password2'])
+
+        if DJANGO_VERSION >= (3, 0):
+            self.assertTrue("The two password fields didn’t match." in response.context['form'].errors['new_password2'])
+        else:
```
zerolab: Same here, the assertions are the same

wagtail/admin/tests/test_account_management.py
zerolab: ah, no I take it back. it is the `didn’t` vs `didn't`

ntravis: Commenting from the bleachers here, but @gasman is there something inherent to Django 3.0 that is causing it or Wagtail to return a new character here? This seems strange, right?

https://unicode-table.com/en/0027/ - `'` Apostrophe
https://unicode-table.com/en/2019/ -  `’` Right Single Quotation Mark

wagtail/admin/tests/test_account_management.py
```diff
@@ -772,7 +781,11 @@ def test_password_reset_confirm_view_post_password_mismatch(self):
 
         # Check that a validation error was raised
         self.assertTrue('new_password2' in response.context['form'].errors.keys())
-        self.assertTrue("The two password fields didn't match." in response.context['form'].errors['new_password2'])
+
+        if DJANGO_VERSION >= (3, 0):
+            self.assertTrue("The two password fields didn’t match." in response.context['form'].errors['new_password2'])
```
ntravis: tagging similar change to above.

wagtail/admin/tests/test_account_management.py
ntravis: Commenting from the bleachers here, but @gasman is there something inherent to Django 3.0 that is causing it or Wagtail to return a new character here? I did find a perhaps related thing in the Django 3.0 [docs page](https://docs.djangoproject.com/en/dev/releases/3.0/#miscellaneous):
`django.utils.html.escape() now uses html.escape() to escape HTML. This converts ' to &#x27; instead of the previous equivalent decimal code &#39;.`

This does explain the change in the test for `wagtail/contrib/modeladmin/tests/test_page_modeladmin.py` later in this PR.

https://unicode-table.com/en/0027/ - `'` Apostrophe
https://unicode-table.com/en/2019/ -  `’` Right Single Quotation Mark <-- this is what you have in the PR right now.

wagtail/admin/tests/test_pages_views.py
```diff
@@ -3002,9 +3003,14 @@ def test_page_copy_post_invalid_slug(self):
         self.assertEqual(response.status_code, 200)
 
         # Check that a form error was raised
-        self.assertFormError(
-            response, 'form', 'new_slug', "Enter a valid 'slug' consisting of letters, numbers, underscores or hyphens."
-        )
+        if DJANGO_VERSION >= (3, 0):
+            self.assertFormError(
+                response, 'form', 'new_slug', "Enter a valid “slug” consisting of letters, numbers, underscores or hyphens."
```
ntravis: https://unicode-table.com/en/201C/ vs the standard https://unicode-table.com/en/0022/
similar question about the change here. I understand why you didn't use the double quote, but is there something that changed which caused the `'` to be an invalid test?

wagtail/admin/tests/test_account_management.py
ntravis: (lost my comment, but it's tied to my other comments above):
@gasman is there a change inherent to Django 3.0 that causes the changes you're making here with different Unicode characters? I found this in [Django 3.0 notes](https://docs.djangoproject.com/en/dev/releases/3.0/#miscellaneous) but that doesn't seem like it would cause this.
> [django.utils.html.escape()](https://docs.djangoproject.com/en/dev/ref/utils/#django.utils.html.escape) now uses [html.escape()](https://docs.python.org/3/library/html.html#html.escape) to escape HTML. This converts ' to &#x27; instead of the previous equivalent decimal code &#39;.

Here, you've got `’` [Right Single Quotation Mark](https://unicode-table.com/en/2019/) compared to the older and more standard `'` [Apostrophe](https://unicode-table.com/en/0027/)

wagtail/admin/tests/test_pages_views.py
gasman: @ntravis All of these validation messages are defined by Django, not Wagtail. I assume that the Django team has simply decided to change these messages to use curly quotes in 3.0 for aesthetic reasons, and that this is unrelated to the change to the `escape` function.

lb-:
Merged in via https://github.com/wagtail/wagtail/commit/dc780c5bb15fb0ae3c039d977e8a172edb5437ae

As mentioned in above discussion, the subtle differences in the test cases are due to different aesthetic formatting in some Django messages for 3.0

PR #5490, opened 2019-08-07
Properly check permissions when viewing revisions

### LINKED ISSUE #5426: revisions_view view should check individual page permission
The `wagtail.admin.views.pages.revisions_view` view (which displays a preview of a page at a given revision number, as linked from the page revision listing) does not apply any permission checks other than `user_has_any_page_permission` (which only checks that they have add/edit/publish/lock permission _somewhere_ in the page tree):

https://github.com/wagtail/wagtail/blob/268df8ceeb7ddc7d4e0864b03670052f4338400f/wagtail/admin/views/pages.py#L1177-L1183

This means that if a Wagtail editor is able to guess a page ID and corresponding revision ID, they would potentially be able to view private pages that they would otherwise have no access to.

The `view_draft` view explicitly checks for edit or publish permission on the requested page, and it would make logical sense to do the same here.

https://github.com/wagtail/wagtail/blob/268df8ceeb7ddc7d4e0864b03670052f4338400f/wagtail/admin/views/pages.py#L586-L588

chosak:
Currently any user with any page permission can view any page revision. This commit fixes that, and ensures that the user has publish or edit permission on the page before showing a page revision. If not, the user is presented with a 403.

This maintains the current behavior if the user has no page permissions, which is to redirect to the admin home page.

New tests have been added to cover these changes.

Fixes #5426.

### DIFF
--- a/wagtail/admin/tests/test_pages_views.py
+++ b/wagtail/admin/tests/test_pages_views.py
@@ -4229,15 +4229,51 @@ class TestRevisions(TestCase, WagtailTestUtils):
         self.assertContains(response, this_christmas_preview_url)
         self.assertContains(response, this_christmas_revert_url)
 
-    def test_preview_revision(self):
+    def request_preview_revision(self):
         last_christmas_preview_url = reverse(
             'wagtailadmin_pages:revisions_view',
             args=(self.christmas_event.id, self.last_christmas_revision.id)
         )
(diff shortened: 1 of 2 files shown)


gasman:
Thanks @chosak! Merged in 7c3418f99e7b2c2da7ad19b2840d73099dadc6a7.

PR #5491, opened 2019-08-07
Don't use page admin title when editing rich text

### LINKED ISSUE #5131: Rich text page links use admin display title instead of regular title
### Issue Summary

When inserting a link to a page from the rich text editor, the anchor text that is pre-populated (if you haven't selected some text to anchor) is generated from the admin page title (`Page.get_admin_display_title`). This is not expected behaviour, because this is not the public page title that is visible on the published page.

### Steps to Reproduce

1. Override `Page.get_admin_display_title` on a page model - for example to include some additional context about the page such as when it was created. This should only be visible when viewing pages in the admin.
2. In a rich text field, add a link to this page. The auto-populated anchor text will use the admin display title instead of just the `title` field value.

### Technical details

* Python version: 3.6
* Django version: 2.0
* Wagtail version: master

chosak:
When editing a rich text field and entering a link to a page whose Page type overrides [`get_admin_display_title`](http://docs.wagtail.io/en/v2.0/reference/pages/model_reference.html?highlight=get_admin_display_title#wagtail.core.models.Page.get_admin_display_title), the custom admin display title is used both when browsing to select the page to link to and also when viewing the rich text editor.

The first behavior is consistent with how custom admin display titles are used throughout the admin, but the second behavior is not. The Wagtail user should be able to use the rich text field as a reasonable preview of what the rendered content will look like for the end user. To do this, the "real" page title should be used, not the admin one.

This commit alters the data that gets passed to the rich text editor so that its title is the real page title and not the admin one.

I wasn't really sure how best to define a test for this, if such a thing is even easily possible. It looks to me like the current implementation was part of [a much larger set of changes](https://github.com/wagtail/wagtail/commit/8a3c47f76afcb45182aa11cc94e5e0443d2836a5#diff-ed9cd5801b0c3c10ba6eaf3878e2b7ebL9) around custom admin display titles.

Fixes #5131.

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/pages/listing/_page_title_choose.html
+++ b/wagtail/admin/templates/wagtailadmin/pages/listing/_page_title_choose.html
@@ -8,7 +8,7 @@ Expects a variable 'page', the page instance.
 
 <div class="title-wrapper">
     {% if page.can_choose %}
-        <a class="choose-page" href="#{{ page.id|unlocalize }}" data-id="{{ page.id|unlocalize }}" data-title="{{ page.get_admin_display_title }}" data-url="{{ page.url }}" data-parent-id="{{ page.get_parent.id|unlocalize }}" data-edit-url="{% url 'wagtailadmin_pages:edit' page.id %}">{{ page.get_admin_display_title }}</a>
+        <a class="choose-page" href="#{{ page.id|unlocalize }}" data-id="{{ page.id|unlocalize }}" data-title="{{ page.title }}" data-url="{{ page.url }}" data-parent-id="{{ page.get_parent.id|unlocalize }}" data-edit-url="{% url 'wagtailadmin_pages:edit' page.id %}">{{ page.get_admin_display_title }}</a>
     {% else %}
         {{ page.get_admin_display_title }}
     {% endif %}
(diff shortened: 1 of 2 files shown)


gasman:
I think the test could just consist of fetching a listing containing a `SimplePage` or `SingleEventPage` (both of which have custom `get_admin_display_title` methods) and confirming that `data-title="plain live title"` is present in the output. (For bonus points, change the page's `draft_title` beforehand to confirm that we're not picking up that change...)

Verifying that the title from `data-title` is actually being picked up by Draftail probably requires more front-end test infrastructure than we currently have, and can reasonably be skipped I think.

chosak:
@gasman regarding your comment mentioning "fetching a listing": the bug in #5131 doesn't render to, and my proposed fix here doesn't alter, the _listing_ title that shows up in the page chooser and in page listing views -- those still use the `get_admin_display_title`, which does default to the draft title. #5131 and this only address the text that gets inserted as the default link text in the rich text editor.

Writing a test for just _that_ is trickier because when the edit page is rendered, the HTML contains the hidden Draftail ContentState data, which is a bit harder to verify. In essence it'd be testing that an HTML template properly renders the template variable we pass it.

gasman:
Yep, understood. The test should check for the presence of the `data-title="plain live title"` attribute on the `<a>` element - we aren't interested in the displayed title in the listing at that point (although having said that, if we don't already have a test to confirm that we're using `get_admin_display_title` as the displayed title here, it wouldn't be a bad idea to add one).

I agree that it's more trouble than it's worth to write an end-to-end test to verify that the title from the `data-title` attribute is correctly finding its way into the rich text field.

chosak:
Thanks @gasman. Test added in a5abecd.

gasman:
Thanks @chosak - merged in 52b01cda67435b9214be6d47555d7bb436dc3472.

Issue #5438, opened 2019-07-18
Save Draft page action is not overridable


kaedroho:
Using the [`construct_page_action_menu`](http://docs.wagtail.io/en/v2.5.1/reference/hooks.html#construct-page-action-menu) hook, we can override the actions that appear in the drop-up box at the bottom of the page editor. However, this [doesn't seem to include](https://github.com/wagtail/wagtail/blob/a2ba2a7a8af9268f2efd0767fe055033c9e2c14a/wagtail/admin/action_menu.py#L123-L140) the "Save Draft" action that appears in the default position.

I think it would be nice if "Save draft" was also configurable using this hook and it was possible to choose a different action as the default one.

## Implementation

We could add a `default` flag to the `ActionMenuItem` class so you could specify which is the default?

For example, to make publish the default action:

```python
@hooks.register('construct_page_action_menu')
def make_publish_default_action(menu_items, request, context):
    menu_items[:] = [
            UnpublishMenuItem(order=10),
            DeleteMenuItem(order=20),
            SubmitForModerationMenuItem(order=30),
            SaveDraftMenuItem(order=40),
            PublishMenuItem(default=True),
    ]
```

If more than one item has `default=True` an error would be raised. If no items have `default=True` then we should automatically add a `SaveDraftMenuItem` as this is what existing implementations would expect.

Alternatively, we could pick which ever item is first (or last) by the `order` field and treat that as the default action.

rjpruitt16:
I will give it a try.

kaedroho:
Thanks @rjpruitt16!

rjpruitt16:
I finished the feature. I am adding test and updating documentation. Should submit pull request by friday.

rjpruitt16:
https://github.com/wagtail/wagtail/pull/5500

PR #5500, opened 2019-08-13
Feature for default customization in action menu

rjpruitt16:
Action menu html files were edited to be render in a list tag or just
a button to leverage existing code. Default button is rendered by
action_menu because the toggled list is rendered in same pattern.
Looking for alternative ways of writing documentation. Preferred
separate code block with instructions, but syntax would not allow.

* Do the tests still pass? (http://docs.wagtail.io/en/latest/contributing/developing.html#testing) yes
* Does the code comply with the style guide? (Run `make lint` from the Wagtail root) yes
* For Python changes: Have you added tests to cover the new/fixed behaviour? yes
* For front-end changes: Did you test on all of Wagtail’s supported browsers? **Please list the exact versions you tested**. chrome
* For new features: Has the documentation been updated accordingly? Yes, but I would like to make it look a little better. Could use help on where to read up on writing documentation.

### DIFF
--- a/docs/reference/hooks.rst
+++ b/docs/reference/hooks.rst
@@ -566,13 +566,19 @@ Hooks for customising the way users are directed through the process of creating
 
   Modify the final list of action menu items on the page creation and edit views. The callable passed to this hook receives a list of ``ActionMenuItem`` objects, a request object and a context dictionary as per ``register_page_action_menu_item``, and should modify the list of menu items in-place.
 
-  Can also be used to customize default action menu.
 
   .. code-block:: python
 
     @hooks.register('construct_page_action_menu')
     def remove_submit_to_moderator_option(menu_items, request, context):
         menu_items[:] = [item for item in menu_items if item.name != 'action-submit']
+
+
+.. _construct_page_action_menu:
+
+  Can also be used to customize default action menu button. The last item in menu_item variable is chosen as the default action. An example on changing the default to publish can be seen below.
+
+  .. code-block:: python
 
     from wagtail.admin.action_menu import UnpublishMenuItem, DeleteMenuItem, SubmitForModerationMenuItem, SaveDraftMenuItem, PublishMenuItem
 
@@ -583,10 +589,8 @@ Hooks for customising the way users are directed through the process of creating
                 DeleteMenuItem(order=20),
                 SubmitForModerationMenuItem(order=30),
                 SaveDraftMenuItem(order=40),
-                PublishMenuItem(_isDefault=True),
+                PublishMenuItem(order=50),
         ]
-
-
 
 
 .. construct_page_listing_buttons:
--- a/docs/reference/hooks.rst
+++ b/docs/reference/hooks.rst
@@ -566,11 +566,28 @@ Hooks for customising the way users are directed through the process of creating
 
   Modify the final list of action menu items on the page creation and edit views. The callable passed to this hook receives a list of ``ActionMenuItem`` objects, a request object and a context dictionary as per ``register_page_action_menu_item``, and should modify the list of menu items in-place.
 
+  Can also be used to customize default action menu.
+
   .. code-block:: python
 
     @hooks.register('construct_page_action_menu')
     def remove_submit_to_moderator_option(menu_items, request, context):
         menu_items[:] = [item for item in menu_items if item.name != 'action-submit']
+
+    from wagtail.admin.action_menu import UnpublishMenuItem, DeleteMenuItem, SubmitForModerationMenuItem, SaveDraftMenuItem, PublishMenuItem
+
+    @hooks.register('construct_page_action_menu')
+    def make_publish_default_action(menu_items, request, context):
+        menu_items[:] = [
+                UnpublishMenuItem(order=10),
+                DeleteMenuItem(order=20),
+                SubmitForModerationMenuItem(order=30),
+                SaveDraftMenuItem(order=40),
+                PublishMenuItem(_isDefault=True),
+        ]
+
+
+
 
 .. construct_page_listing_buttons:
 
--- a/wagtail/admin/action_menu.py
+++ b/wagtail/admin/action_menu.py
@@ -1,6 +1,5 @@
 """Handles rendering of the list of actions in the footer of the page create/edit views."""
 
-from django.core.exceptions import ImproperlyConfigured
 from django.forms import Media, MediaDefiningClass
 from django.template.loader import render_to_string
 from django.urls import reverse
@@ -19,10 +18,9 @@ class ActionMenuItem(metaclass=MediaDefiningClass):
     label = ''
     name = None
 
-    def __init__(self, order=None, _isDefault=False):
+    def __init__(self, order=None):
         if order is not None:
             self.order = order
-        self._isDefault = _isDefault
 
     def is_shown(self, request, context):
         """
@@ -102,7 +100,6 @@ class UnpublishMenuItem(ActionMenuItem):
         )
 
     def get_url(self, request, context):
-        print(self.is_shown(request, context), self._isDefault)
         return reverse('wagtailadmin_pages:unpublish', args=(context['page'].id,))
 
 
@@ -148,6 +145,7 @@ def _get_base_page_action_menu_items():
             DeleteMenuItem(order=20),
             PublishMenuItem(order=30),
             SubmitForModerationMenuItem(order=40),
+            SaveDraftMenuItem(order=50),
         ]
         for hook in hooks.get_hooks('register_page_action_menu_item'):
             BASE_PAGE_ACTION_MENU_ITEMS.append(hook())
@@ -162,7 +160,6 @@ class PageActionMenu:
         self.request = request
         self.context = kwargs
         self.context['user_page_permissions'] = UserPagePermissionsProxy(self.request.user)
-        self.default_item = SaveDraftMenuItem(_isDefault=True)
 
         self.menu_items = [
             menu_item
@@ -170,34 +167,25 @@ class PageActionMenu:
             if menu_item.is_shown(self.request, self.context)
         ]
 
-
-        self.menu_items.sort(key=lambda item: (item._isDefault, item.order))
+        self.menu_items.sort(key=lambda item: item.order)
 
         for hook in hooks.get_hooks('construct_page_action_menu'):
             hook(self.menu_items, self.request, self.context)
 
-        # Chcek improperly configured default menu_itmes
-        default_count = 0
-        for menu_item in self.menu_items:
-            if menu_item._isDefault:
-                self.default_item = menu_item
-                default_count += 1
+        self.default_item = self.menu_items.pop() if self.menu_items else SaveDraftMenuItem(order=50)
 
-            if default_count > 1:
-                raise ImproperlyConfigured("Only one menu item can be default")
 
     def render_html(self):
+        print(self.template)
         return render_to_string(self.template, {
+            'default_menu_item': self.default_item.render_html(self.request, self.context),
             'show_menu': bool(self.menu_items),
             'rendered_menu_items': [
                 menu_item.render_html(self.request, self.context)
                 for menu_item in self.menu_items
-                if not menu_item._isDefault and menu_item.is_shown(self.request, self.context)
+                if menu_item.is_shown(self.request, self.context)
             ],
         }, request=self.request)
-
-    def render_default_html(self):
-        return self.default_item.render_html(self.request, self.context)
 
     @cached_property
     def media(self):
--- a/wagtail/admin/action_menu.py
+++ b/wagtail/admin/action_menu.py
@@ -1,5 +1,6 @@
 """Handles rendering of the list of actions in the footer of the page create/edit views."""
 
+from django.core.exceptions import ImproperlyConfigured
 from django.forms import Media, MediaDefiningClass
 from django.template.loader import render_to_string
 from django.urls import reverse
@@ -18,9 +19,10 @@ class ActionMenuItem(metaclass=MediaDefiningClass):
     label = ''
     name = None
 
-    def __init__(self, order=None):
+    def __init__(self, order=None, _isDefault=False):
         if order is not None:
             self.order = order
+        self._isDefault = _isDefault
 
     def is_shown(self, request, context):
         """
@@ -56,6 +58,7 @@ class ActionMenuItem(metaclass=MediaDefiningClass):
 
 
 class PublishMenuItem(ActionMenuItem):
+    label = "Publish"
     name = 'action-publish'
     template = 'wagtailadmin/pages/action_menu/publish.html'
 
@@ -99,6 +102,7 @@ class UnpublishMenuItem(ActionMenuItem):
         )
 
     def get_url(self, request, context):
+        print(self.is_shown(request, context), self._isDefault)
         return reverse('wagtailadmin_pages:unpublish', args=(context['page'].id,))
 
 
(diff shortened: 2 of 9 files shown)


rjpruitt16:
https://github.com/wagtail/wagtail/issues/5438

docs/reference/hooks.rst
```diff
@@ -566,12 +566,29 @@ Hooks for customising the way users are directed through the process of creating
...
 
+  Can also be used to customize default action menu.
+
   .. code-block:: python
 
     @hooks.register('construct_page_action_menu')
     def remove_submit_to_moderator_option(menu_items, request, context):
         menu_items[:] = [item for item in menu_items if item.name != 'action-submit']
 
+    from wagtail.admin.action_menu import UnpublishMenuItem, DeleteMenuItem, SubmitForModerationMenuItem, SaveDraftMenuItem, PublishMenuItem
```
kaedroho: This would be nicer if it was it was a separate example with its own description I think

docs/reference/hooks.rst
```diff
@@ -566,12 +566,29 @@ Hooks for customising the way users are directed through the process of creating
...
+
+    @hooks.register('construct_page_action_menu')
+    def make_publish_default_action(menu_items, request, context):
+        menu_items[:] = [
+                UnpublishMenuItem(order=10),
+                DeleteMenuItem(order=20),
+                SubmitForModerationMenuItem(order=30),
+                SaveDraftMenuItem(order=40),
+                PublishMenuItem(_isDefault=True),
```
kaedroho: Could you rename this parameter to `default`, so it's consistent with naming used elsewhere

wagtail/admin/action_menu.py
```diff
@@ -147,27 +162,43 @@ def __init__(self, request, **kwargs):
...
             'show_menu': bool(self.menu_items),
             'rendered_menu_items': [
                 menu_item.render_html(self.request, self.context)
                 for menu_item in self.menu_items
-            ]
+                if not menu_item._isDefault and menu_item.is_shown(self.request, self.context)
+            ],
         }, request=self.request)
 
+    def render_default_html(self):
```
kaedroho: Could we do this rendering as part of the `render_html` method above?

wagtail/admin/tests/test_pages_views.py
```diff
@@ -2102,6 +2105,41 @@ def hook_func(request, page):
...
+        </ul>
+        '''
+        with self.register_hook('construct_page_action_menu', hook_func):
+            response = self.client.get(reverse('wagtailadmin_pages:edit',
+                                       args=(self.single_event_page.id, )))
+
+        self.assertContains(response, default_item, html=True)
+        self.assertContains(response, menu_items_html, html=True)
+
+
```
kaedroho: Also need to test:
 - Not specifying `default`
 - Using `default` more than once

gasman:
Just a thought - rather than having a `default` flag and requiring exactly one menu item to have it set, could we simply make a rule that the menu item with the lowest (or highest) `order` value is made the default? That way, we avoid having to redefine the whole list to set a new default, and it's easier to resolve clashes if these hooks are used in more than one place (for example, there are multiple third-party extensions installed that add new menu items).

wagtail/admin/tests/test_pages_views.py
rjpruitt16: I change functionality to last menu_item is default.

wagtail/admin/action_menu.py
rjpruitt16: Yes, good catch!

docs/reference/hooks.rst
rjpruitt16: Done

docs/reference/hooks.rst
rjpruitt16: I change functionality to last menu_item is default.

rjpruitt16:
I am not sure why the test is breaking. I read the documentation, and it seems like valid syntax. https://docs.djangoproject.com/en/2.2/topics/testing/tools/#django.test.SimpleTestCase.assertHTMLEqual. Does anyone have in suggestions?

rjpruitt16:
I used assertTemplateUsed instead of assertContains because templates are rendered by the database of the project. Each database renders html a little differently, so that why my test work only on sqlite. I can hardcoded each possible response for each database, but I believe this method is more maintainable.
The only downside is that my test do not proves the last item is default. It only checks to if the appropriate templates were used. if the item is default, it must be displayed first. 
I am still researching different ideas, but what do you think is best?

gasman:
> templates are rendered by the database of the project

I can promise that they aren't... :-)

We need to get to the bottom of why these differences are occurring. On Slack I suggested printing `response.content` (NOT `response.context`), as this would allow us to see the generated HTML, and therefore how it differs from the expected output. We can then decide on the best way to account for this in the tests - for example, if the only differences are in whitespace, we can easily set the tests to ignore that (although even then, I'd still want to understand why that's affected by the choice of database).

gasman:
Hi @rjpruitt16,
The `assertContains` was failing because you had the page ID 6 hard-coded in the Unpublish and Delete URLs:

            <li>	
                <a class="button" href="/admin/pages/6/unpublish/">Unpublish</a>	
            </li>	
            <li>	
                <a class="button" href="/admin/pages/6/delete/">Delete</a>	
            </li>

The `single_event_page` is created at the start of the test (in the `setUp` method), so it'll only be assigned an ID of 6 if the ID sequence is reset after every test, and apparently Postgres and MySQL don't do that. If you change this to pass in the actual ID:

        menu_items_html = '''<ul>
            <li>
                <a class="button" href="/admin/pages/{id}/unpublish/">Unpublish</a>
            </li>
            <li>
                <a class="button" href="/admin/pages/{id}/delete/">Delete</a>
            </li>
            <li>
                <input type="submit" name="action-submit" value="Submit for moderation" class="button" />
            </li>
            <li>
                <button type="submit" class="button action-save button-longrunning " data-clicked-text="Saving…" ><span class="icon icon-spinner"></span><em>Save draft</em></button>
            </li>
        </ul>'''.format(id=self.single_event_page.id)

then the test should pass on all databases.

rjpruitt16:
I saw this too. My apologies for the delay, I just started school again. I will make the change and push now.

rjpruitt16:
Is there something else to be built? @gasman @kaedroho

rjpruitt16:
@gasman @kaedroho

docs/reference/hooks.rst
```diff
@@ -566,12 +566,33 @@ Hooks for customising the way users are directed through the process of creating
...
+  Can also be used to customize default action menu button. The last item in menu_item variable is chosen as the default action. An example on changing the default to publish can be seen below.
+
+  .. code-block:: python
+
+    from wagtail.admin.action_menu import UnpublishMenuItem, DeleteMenuItem, SubmitForModerationMenuItem, SaveDraftMenuItem, PublishMenuItem
+
+    @hooks.register('construct_page_action_menu')
+    def make_publish_default_action(menu_items, request, context):
+        menu_items[:] = [
+                UnpublishMenuItem(order=10),
```
kaedroho: This list is indented to much

wagtail/admin/action_menu.py
```diff
@@ -159,13 +172,19 @@ def __init__(self, request, **kwargs):
         for hook in hooks.get_hooks('construct_page_action_menu'):
             hook(self.menu_items, self.request, self.context)
 
+        self.default_item = self.menu_items.pop() if self.menu_items else SaveDraftMenuItem(order=50)
+
+
     def render_html(self):
+        print(self.template)
```
kaedroho: Please remove this

wagtail/admin/templates/wagtailadmin/pages/action_menu/menu.html
```diff
@@ -1,6 +1,11 @@
 {% if show_menu %}
+    {{ default_menu_item }}
     <div class="dropdown-toggle icon icon-arrow-up"></div>
     <ul>
-        {% for item in rendered_menu_items %}{{ item }}{% endfor %}
+        {% for item in rendered_menu_items %}
+            <li>
+                 {{ item }}
```
kaedroho: Moving the `<li>` tags out of the template is a breaking change. I agree it's better this way though. @gasman is this worth an upgrade considerations entry?

wagtail/admin/templates/wagtailadmin/pages/create.html
```diff
@@ -26,12 +26,13 @@ <h1 class="icon icon-doc-empty-inverse">
...
                 <ul>
-                    <li class="actions">
-                        <div class="dropdown dropup dropdown-button match-width">
-                            <button type="submit" class="button action-save button-longrunning" data-clicked-text="{% trans 'Saving…' %}"><span class="icon icon-spinner"></span><em>{% trans 'Save draft' %}</em></button>
-                            {{ action_menu.render_html }}
-                        </div>
-                    </li>
+                  <li class="actions">
+                      <div class="dropdown dropup dropdown-button match-width {% if is_revision %}warning{% endif %}">
+                          {{ action_menu.render_default_html }}
```
kaedroho: Looks like this has been moved into `render_html` so we should delete this line

wagtail/admin/templates/wagtailadmin/pages/action_menu/save_draft.html
```diff
@@ -0,0 +1,2 @@
+{% load i18n %}
+<button type="submit" class="button action-save button-longrunning {% if is_revision %}warning{% endif %}" data-clicked-text="{% trans 'Saving…' %}" {% if page.locked %}disabled {% endif %}><span class="icon icon-spinner"></span><em>{% if page.locked %}{% trans 'Page locked' %}{% else %}{% if is_revision %}{% trans 'Replace current draft' %}{% else %}{% trans 'Save draft' %}{% endif %}{% endif %}</em></button>
```
kaedroho: Looks like this was copied from `edit.html`. Does it work on the create view? (just wondering if the `{% if page.locked %}` bit is a problem as the other button templates don't seem to have this).

kaedroho:
Thanks again @rjpruitt16!

wagtail/admin/templates/wagtailadmin/pages/action_menu/menu.html
gasman: @kaedroho Probably a good idea, yes. We don't document what should be in the template, but we do document the fact that the template is overrideable, which makes this feature "semi-official".

In fact, thinking about this further, this could be a problem for add-on packages that make use of this hook but want to be compatible with both Wagtail versions. Is there any way of achieving that, other than having two templates and using the Wagtail version number to choose between them?

docs/reference/hooks.rst
```diff
@@ -566,12 +566,33 @@ Hooks for customising the way users are directed through the process of creating
...
+                DeleteMenuItem(order=20),
+                SubmitForModerationMenuItem(order=30),
+                SaveDraftMenuItem(order=40),
+                PublishMenuItem(order=50),
+        ]
```
gasman: It would be good to update this example so that it doesn't require redefining the whole list. Perhaps something like:

    def make_publish_default_action(menu_items, request, context):
        for item in menu_items:
            if isinstance(item, PublishMenuItem):
                item.order = 100

(although changing the `order` on an existing menu item feels a bit sneaky, and I'm not 100% sure we should encourage it...)

rjpruitt16: Can I just add a line to the comment saying, "redefining the whole list is not necessary, but for the purposes of a clear example, the list is redefined to show why PublishMenuItem is default. " That way, the example is neither sneaky nor misleading.

gasman:
Have rebased on current master here https://github.com/gasman/wagtail/tree/rjpruitt16-SaveDraft - I'll apply the final fixes so that we can hopefully get this merged for 2.7.

wagtail/admin/templates/wagtailadmin/pages/action_menu/save_draft.html
gasman: The `{% if page.locked %}` checks shouldn't be a problem for the create view, as `page.locked` is always false for a newly-created page.

However, the current behaviour isn't quite right on the edit page when the page is locked, because `SaveDraftMenuItem` now hides itself for locked pages (the default behaviour of `ActionMenuItem.is_shown`), as do all other menu items, so the menu disappears completely. In any case, these template-level customisations to the "Save draft" button (e.g. changing its label to "Page locked") are only meaningful when it's the default button, and the whole point of this PR is that we can't assume that any more :-)

I think the most logical solution is to make the "Page Locked" button a menu item in its own right - only ever shown when the page _is_ locked, of course - and give it a very high (or low, depending on the outcome of https://github.com/wagtail/wagtail/pull/5500#pullrequestreview-286868977) order value so that if another custom button defines itself as active on locked pages for some reason, it won't become the default (unless the developer is really intent on doing so) and so the "Page Locked" label won't end up buried in the list.

gasman: Above proposal now implemented in https://github.com/gasman/wagtail/commit/5abcddfa478d526632276c4c54af0716e72246f7

docs/reference/hooks.rst
gasman: I've realised that my suggested code won't work, because `construct_page_action_menu` is called after the list has been sorted by `item.order`, and so at that point the list's actual order is meaningful rather than the `order` attribute. I think the simplest / clearest example for making Publish the default would therefore be:

    def make_publish_default_action(menu_items, request, context):
        for (index, item) in menu_items:
            if item.name == 'action-publish':
                # move to end of list
                menu_items.pop(index)
                menu_items.append(item)
                break

(The fix at https://github.com/wagtail/wagtail/pull/5500#discussion_r332783437 is a good illustration of why we shouldn't encourage redefining the whole list - anyone with that code in place would now need to update it with the new Page Locked item.)

docs/reference/hooks.rst
gasman: Fixed on https://github.com/gasman/wagtail/commits/rjpruitt16-SaveDraft

wagtail/admin/action_menu.py
gasman: Removed on https://github.com/gasman/wagtail/commits/rjpruitt16-SaveDraft

wagtail/admin/templates/wagtailadmin/pages/create.html
gasman: Removed on https://github.com/gasman/wagtail/commits/rjpruitt16-SaveDraft

gasman:
> Change the behaviour to make the lowest order the default, and set SaveDraftMenuItem to 0. (It's unlikely that any existing code has used 0 or something negative as the order.)

This approach is now implemented in https://github.com/gasman/wagtail/commit/187b78849ffaf29f4050a16c589357a76d358a9a

docs/reference/hooks.rst
gasman: Example code updated in https://github.com/gasman/wagtail/commit/6d6bda072462d6a1c859efc45aa0c3181feedcc2

wagtail/admin/templates/wagtailadmin/pages/action_menu/menu.html
gasman: Release note added in https://github.com/gasman/wagtail/commit/ad753ad9de9d7c273a28c332e9d483c29fd3a0c3. I thought about ways to make this fully backwards-compatible (e.g. giving the `<li>`-less template a distinct attribute like `button_template` and having the menu logic choose the best available template), but decided that was overkill in the end - the side effect of an extra `<li>` is minimal (an extra 1px border), and it's easy for a third-party library to add a version check if necessary.

gasman:
Have now opened an updated PR to replace this one: #5612

PR #5501, opened 2019-08-14
Fix pageurl and slugurl handling of situations where request.site is null

solarissmoke:
Hi!

I've run into an edge case when using `pageurl` and `slugurl`. 

It is possible for `SiteMiddleware` to [set `request.site` to `None`](https://github.com/wagtail/wagtail/blob/master/wagtail/core/middleware.py#L15) (as opposed to not setting the `site` attribute at all) - e.g., if no default site exists. In such cases the existing behaviour of these template tags fails, because they assume that `request.site` will always be a `Site` object.

This patch adds a check for this condition and falls back to using a non-site-aware URL as would be the case if `request.site` was not set at all.

### DIFF
--- a/wagtail/core/templatetags/wagtailcore_tags.py
+++ b/wagtail/core/templatetags/wagtailcore_tags.py
@@ -31,6 +31,10 @@ def pageurl(context, page, fallback=None):
         # request.site not available in the current context; fall back on page.url
         return page.url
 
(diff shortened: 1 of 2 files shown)


gasman:
Thanks @solarissmoke! Merged in 0ee07dddffe613d12b32bc92e4979873c822ce0c.

PR #5505, opened 2019-08-14
Internal anchor links/links with fragment identifiers. Fixes #1049

### LINKED ISSUE #1049: Not possible to add link with fragment identifier

imansyed000:
Fixes #1049

Tested on Chrome, Safari and Firefox.

image
image
image
image

User is able to input some know fragment identifier.
Anchor field currently lacks any validation.

### DIFF
--- a/client/src/components/Draftail/decorators/Link.js
+++ b/client/src/components/Draftail/decorators/Link.js
@@ -29,6 +29,9 @@ export const getLinkAttributes = (data) => {
   } else if (url.startsWith('mailto:')) {
     icon = MAIL_ICON;
     label = getEmailAddress(url);
+  } else if (url.startsWith('#')) {
+    icon = LINK_ICON;
+    label = url;
   } else {
     icon = LINK_ICON;
     label = getDomainName(url);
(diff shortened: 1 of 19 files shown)


lb-:
merged in via https://github.com/wagtail/wagtail/commit/794d40b86b3d9c652c37acbc00163890077a78b5 thanks @imansyed000

johnraz:
Awesome! Thanks a lot @imansyed000 this PR will help me understand how the all thing works and be used in my project as well! 👍 🎉

zerolab:
https://github.com/jacobtoppm/wagtail_draftail_anchors/ provides the ability to add/edit anchors so should go well with this feature

kaedroho:
Wonder if we should have this switched off by default? It's not a commonly requested feature and requires some frontend development to make it work. I worry it could actually be confusing for editors of sites that don't actually support it.

awhileback:
> Wonder if we should have this switched off by default? It's not a commonly required feature and needs some frontend development to make it work. I worry it could actually be confusing for editors of sites that don't actually support it.

There also seems to be a bug in draft vs publish. I coupled this with the package zerolab linked above your reply, which is really a requirement to use this because there's otherwise no way to insert anchor ids, and when saving a draft all of the links got converted to 'default' draftail links.  After editing them again, and not previewing them or saving a draft but rather publishing the changes directly, they stuck properly.

PR #5506, opened 2019-08-14
Rich text preserves whitespace in <pre> tag. Fixes #3864

### LINKED ISSUE #3864: Add support for code tag in rich text editor
`<code>` tag is not a built-in plugin for wagtail. We use wagtail in-house for our website and would like to be able to write blog posts that include inline code as well as code blocks. I was surprised to find that `<code>` is not currently one of the tags available by default. I know that many users of wagtail are non-technical and may not need that particular button, but it would be cool to have out of the box for those of us who are.

imansyed000:
Fixes #3864.

Tested on Safari, Firefox and Chrome.

### DIFF
--- a/wagtail/admin/rich_text/converters/html_to_contentstate.py
+++ b/wagtail/admin/rich_text/converters/html_to_contentstate.py
@@ -315,9 +315,11 @@ class HtmlToContentStateHandler(HTMLParser):
             element_handler.handle_endtag(name, self.state, self.contentstate)
 
     def handle_data(self, content):
-        # normalise whitespace sequences to a single space
+        # normalise whitespace sequences to a single space unless it is contained in <pre> tag, in
+        # which case leave it alone
         # This is in line with https://www.w3.org/TR/html4/struct/text.html#h-9.1
-        content = re.sub(WHITESPACE_RE, ' ', content)
+        if not (content.startswith('<pre>') and content.endswith('</pre>')):
+            content = re.sub(WHITESPACE_RE, ' ', content)
 
         if self.state.current_block is None:
             if content == ' ':


wagtail/admin/rich_text/converters/html_to_contentstate.py
```diff
@@ -315,9 +315,11 @@ def handle_endtag(self, name):
             element_handler.handle_endtag(name, self.state, self.contentstate)
 
     def handle_data(self, content):
-        # normalise whitespace sequences to a single space
+        # normalise whitespace sequences to a single space unless it is contained in <pre> tag, in
+        # which case leave it alone
         # This is in line with https://www.w3.org/TR/html4/struct/text.html#h-9.1
-        content = re.sub(WHITESPACE_RE, ' ', content)
+        if not (content.startswith('<pre>') and content.endswith('</pre>')):
```
gasman: This seems a little odd - the purpose of [`HTMLParser`](https://docs.python.org/3/library/html.parser.html) is to translate the string representation of HTML into a structure we can iterate over, so I'd expect it to process an element like

    <pre>if x &lt; 5:</pre>

by

* calling `handle_starttag` with the tag name `'pre'`
* calling `handle_data` with the content `"if x < 5:"` (note that the `<` is unescaped at this point, since it's giving us the tag content in plain text form, rather than the HTML serialisation of it)
* calling `handle_endtag` with the tag name `'pre'`

The only time I'd expect to see `<pre>...</pre>` within `handle_data` would be if that's part of the actual document text, i.e. the HTML is something like

    <p>&lt;pre&gt; and &lt;/pre&gt; are used to indicate preformatted text.</p>

@imansyed000, please can you give some details of how you tested this, so I can investigate further?

imansyed000: I think I've probably misunderstood the issue, which I thought was the latter description.

gasman:
Closing as invalid, as per https://github.com/wagtail/wagtail/pull/5506#discussion_r313949627.

Issue #3311, opened 2017-01-29
Licensing problem with unidecode


geonux:
Not all Wagtail dependencies have same licenses. This is not a problem in most cases, many licenses are compatible with each other, in particular non-contaminating licenses such as BSD, MIT, ...
Wagtail is made available through one of these non-contaminating license : a BSD 3 Clauses.

The problem comes from Unidecode, which is GPL v2 (https://github.com/avian2/unidecode/blob/master/LICENSE). This license is a so-called contaminant license. All derivative works resulting from the integration of the software component are therefore theoretically contaminated and should therefore be distributed with the same license. Wagtail should be GPL v2.
Unfortunately, this license is very contaminating and therefore poses many integration problems - which limits the possibilities of use and integrate your piece of software. Moreover, it is not really in the philosophy of Python that prone the opening and the non-contaminating licenses.

Moreover, the GPL v2 license is incompatible with another component : Django-treebeard, which is released under Apache v2 (https://github.com/django-treebeard/django-treebeard/blob/master/LICENSE).
More informations about that : 

- http://gplv3.fsf.org/wiki/index.php/Compatible_licenses
- https://www.apache.org/licenses/GPL-compatibility.html

Today, licenses of Wagtail and its dependencies are so incompatible.
Can you do the necessary tasks to correct this problem?
The removal of the Unidecode component will probably be the better solution. This solution keeps the current license and therefore the user community of Wagtail.

For your information, I am not the owner of one of these libraries. I just want to integrate Wagtail for my developments and I have strong constraints to produce "clean" software.

thibaudcolas:
Thanks for reporting this. It seems like others have had this problem with unidecode in the past: https://github.com/avian2/unidecode/issues/1.

> No, I have no plans to release Unidecode under a different license. See https://github.com/kmike/text-unidecode for a similar library under an Artistic License.

https://github.com/kmike/text-unidecode is also a port of Perl's `Text::Unidecode`. Its README states:

> There are other Python ports of Text::Unidecode (unidecode and isounidecode). unidecode is GPL; isounidecode doesn't support Python 3 and uses too much memory.
> This port is licensed under Artistic License and supports both Python 2.x and 3.x. If you're OK with GPL, use unidecode (it has better memory usage and better transliteration quality).

Seems to me like this is an option worth looking at.

spapas:
Hello, just to add some insights on this discussion: 

[...]

PR #5508, opened 2019-08-15
Consolidate all calls to unidecode to a single utility function 

mvantellingen:
This is in preparation for support other implementations then unidecode
since it has a GPLv2 license.

See #3311

### DIFF
--- a/wagtail/utils/text.py
+++ b/wagtail/utils/text.py
@@ -1,0 +1,5 @@
+from unidecode import unidecode
+
+
+def string_to_ascii(value) -> str:
+    return str(unidecode(value))
--- a/wagtail/contrib/forms/models.py
+++ b/wagtail/contrib/forms/models.py
@@ -6,11 +6,11 @@ from django.db import models
 from django.shortcuts import render
 from django.utils.text import slugify
 from django.utils.translation import ugettext_lazy as _
-from unidecode import unidecode
 
 from wagtail.admin.edit_handlers import FieldPanel
 from wagtail.admin.utils import send_mail
 from wagtail.core.models import Orderable, Page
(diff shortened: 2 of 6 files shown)


wagtail/utils/text.py
```diff
@@ -0,0 +1,9 @@
+try:
+    from unidecode import unidecode
+except ImportError:
+    def unidecode(value):
+        return value
+
+
+def string_to_ascii(value) -> str:
+    return str(unidecode(value))
```
gasman: I don't think this is right... If unidecode isn't installed, this function will have no effect and the string will remain as unicode, breaking any code that relies on it returning ASCII.

mvantellingen: Yeah you are right, I think we can reuse some logic from the `cautious_slugify` code

mvantellingen: I reused the fallback mechanism which was in the postgres search module. I now removed that via a force-push since unidecode is required anyway.

The idea of this PR is purely to focus on consolidating all calls to unidecode, but still leverage it

gasman: Ah, OK :-) In that case, I suggest we hold off merging this until we have a fix for #3088 in place, since this extra level of indirection makes it less obvious that wagtailforms cannot (currently) be migrated away from unidecode without losing existing data.

lb-:
Flagging as needs work, as this is blocked until we have a fix for #3088

lb-:
This PR should be superseded by https://github.com/wagtail/wagtail/pull/6093

Along with consideration for form fields in existing applications.

Note: I have avoided the try/catch wrapping of unidecode imports, but the PR will enable us to do something like that in a future release, or find a replacement.

Issue #2968, opened 2016-09-02
Elasticsearch: Unable to assign different boost values for same-named fields


m1kola:
The following definition is valid for Elasticsearch 1.*, but invalid for 2.*:

``` python
    search_fields = Page.search_fields + [
        index.SearchField('title', partial_match=True, boost=10),
        index.SearchField('intro'),
        index.SearchField('body'),
    ]
```

With Elasticsearch 2 it causes the following error:

> Mapper for [title] conflicts with existing mapping in other types:\n[mapper [title] has different [analyzer], mapper [title] is used by multiple types. Set update_all_types to true to update [boost] across all types.

I think we need to mention it somehow in documentation (maybe in the "Upgrade considerations" section?).

**Important!** The same issue can be reproduced with `wagtail.wagtailadmin.taggable.TagSearchable`. This class is deprecated, but will be removed only in Wagtail 1.8. That's why I marked this issue as "Release blocker".

Refs:
- #2570
- [Conflicts between fields in different types](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/indices-put-mapping.html#merging-conflicts)

(cc @kaedroho)

kaedroho:
To solve this, we would have to add the boosting manually to each document while indexing them instead of on the mapping. It's really annoying how Elasticsearch have added this seemingly artificial restriction in as the boost is stored on a per document-field basis anyway...

It would be better if we could perform field-boosting at query time as this would be faster and give more accurate results but would be trickier to implement.

kaedroho:
> To solve this, we would have to add the boosting manually to each document while indexing them instead of on the mapping

I can't find a way of doing this, looks like query time boosting may be the way to go

kaedroho:
One thing that might make implementing the query-time boost simpler could be to split the _all field into four separate fields for different rank levels. When indexing, each field is indexed twice, once with the usual field name and again in the global field that has the closest rank value.

When querying "all" fields, we query these four fields together setting the rank on the query.

This is basically how PostgreSQL FTS does ranking so restricting Wagtail to four levels of ranking should make it easier to integrate with that.

For example, the mapping of a page could be as follows:

```
{
    "properties" {
        "rank_a": {"type": "string"},
        "rank_b": {"type": "string"},
        "rank_c": {"type": "string"},
        "rank_d": {"type": "string"},
        "title": {"type": "string", "copy_to": "rank_a"},
        "body": {"type": "string", "copy_to": "rank_b"},
[...]

PR #5509, opened 2019-08-15
Fix doc and test on ES6 python client compat

johnraz:
This commit changes the statement that version < 6.3.1 of the
elasticsearch python client should be used to instead state
that 6.4.0 is fine.

It also update the tests to reflect the statement.

Initially the `update_all_types` argument has been used to work
around an issue described in:
https://github.com/wagtail/wagtail/issues/2968

This argument was removed in elasticsearch-py 6.3.1 and making
use of it was raising an error.

With 6.4.0 nothing is raising anymore.
It is not quite clear why it's not raising anymore but it works.

### DIFF
--- a/docs/topics/search/backends.rst
+++ b/docs/topics/search/backends.rst
@@ -113,11 +113,12 @@ Prerequisites are the `Elasticsearch`_ service itself and, via pip, the `elastic
 
 .. code-block:: sh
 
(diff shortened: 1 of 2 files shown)


lb-:
@johnraz  - thanks for this & thanks for approving /reviewing @kaedroho 

this has been merged in via https://github.com/wagtail/wagtail/commit/2de92f045cacac8ac3f29b7596fce23b090e1a96

PR #4777, opened 2018-09-14
Split up wagtail.admin.forms into submodules

gasman:
Fixes #4515 - splitting up `wagtail.admin.forms` means that `wagtail.admin.edit_handlers` can import our custom model form logic (now in `wagtail.admin.forms.models` and `wagtail.admin.forms.pages`) without also pulling in the auth forms that cause a circular dependency when custom user models are in use.

PR #5510, opened 2019-08-15
Reorganise wagtail.admin.utils and .decorators into meaningful sub-units

### LINKED ISSUE #5507: Reorganise wagtail.admin.utils
The `wagtail.admin.utils` module has become a disorganised dumping ground for random functions. Indeed, having a "utils" module at all is widely regarded as an anti-pattern:

https://lostechies.com/chrismissal/2009/06/01/anti-patterns-and-worst-practices-utils-class/
https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common

I propose to reorganise the contents of this module as follows:

**Move to wagtail.admin.locale:**

    WAGTAILADMIN_PROVIDED_LANGUAGES
    get_js_translation_strings
    get_available_admin_languages
    get_available_admin_time_zones

**Move to wagtail.admin.models:** (although 'models' is a special case within Django, as it gets preloaded on startup - will this cause any side effects?)

    get_object_usage
    popular_tags_for_model

**Move to wagtail.admin.auth:**

    users_with_page_permission
    permission_denied
    user_passes_test
    permission_required
    any_permission_required
    PermissionPolicyChecker
    user_has_any_page_permission
    - also reject_request and require_admin_access from wagtail.admin.decorators

**Move to wagtail.admin.mail (or wagtail.admin.notifications):**

    send_mail
    send_notification

**Move to wagtail.admin.navigation (which already exists):**

    get_site_for_user

As some of these are likely to be used in external code (particularly the auth decorators such as `permission_required`), I suggest we resurrect the `MovedDefinitionHandler` previously introduced in #4777 to provide a deprecation path:

https://github.com/wagtail/wagtail/commit/dc4457d4148116cd8ff91b90e4583e342692cf2a

gasman:
Fixes #5507.

@chosak Assigning you for review, as you previously did the honours for wagtail.admin.forms in #4777 and the only 'clever' bit of code here is essentially the same :-)

(Memo to self: update https://github.com/wagtail/wagtail/wiki/Managing-Wagtail-translations#fetching-new-translations-from-transifex to point to wagtail/admin/locale.py once this is merged)

### DIFF
--- a/wagtail/admin/models.py
+++ b/wagtail/admin/models.py
@@ -1,5 +1,55 @@
+from django.contrib.contenttypes.models import ContentType
+from django.db.models import Count
+
+from modelcluster.fields import ParentalKey
+from taggit.models import Tag
+
 # The edit_handlers module extends Page with some additional attributes required by
 # wagtailadmin (namely, base_form_class and get_edit_handler). Importing this within
 # wagtailadmin.models ensures that this happens in advance of running wagtailadmin's
 # system checks.
 from wagtail.admin import edit_handlers  # NOQA
+from wagtail.core.models import Page
+
+
+def get_object_usage(obj):
+    "Returns a queryset of pages that link to a particular object"
+
+    pages = Page.objects.none()
+
+    # get all the relation objects for obj
+    relations = [f for f in type(obj)._meta.get_fields(include_hidden=True)
+                 if (f.one_to_many or f.one_to_one) and f.auto_created]
+    for relation in relations:
+        related_model = relation.related_model
+
+        # if the relation is between obj and a page, get the page
+        if issubclass(related_model, Page):
+            pages |= Page.objects.filter(
+                id__in=related_model._base_manager.filter(**{
+                    relation.field.name: obj.id
+                }).values_list('id', flat=True)
+            )
+        else:
+            # if the relation is between obj and an object that has a page as a
+            # property, return the page
+            for f in related_model._meta.fields:
+                if isinstance(f, ParentalKey) and issubclass(f.remote_field.model, Page):
+                    pages |= Page.objects.filter(
+                        id__in=related_model._base_manager.filter(
+                            **{
+                                relation.field.name: obj.id
+                            }).values_list(f.attname, flat=True)
+                    )
+
+    return pages
+
+
+def popular_tags_for_model(model, count=10):
+    """Return a queryset of the most frequently used tags used on this model class"""
+    content_type = ContentType.objects.get_for_model(model)
+    return Tag.objects.filter(
+        taggit_taggeditem_items__content_type=content_type
+    ).annotate(
+        item_count=Count('taggit_taggeditem_items')
+    ).order_by('-item_count')[:count]
--- a/wagtail/admin/models.py
+++ b/wagtail/admin/models.py
@@ -5,8 +5,8 @@ from modelcluster.fields import ParentalKey
 from taggit.models import Tag
 
 # The edit_handlers module extends Page with some additional attributes required by
-# wagtailadmin (namely, base_form_class and get_edit_handler). Importing this within
-# wagtailadmin.models ensures that this happens in advance of running wagtailadmin's
+# wagtail admin (namely, base_form_class and get_edit_handler). Importing this within
+# wagtail.admin.models ensures that this happens in advance of running wagtail.admin's
 # system checks.
 from wagtail.admin import edit_handlers  # NOQA
 from wagtail.core.models import Page
--- a/wagtail/utils/deprecation.py
+++ b/wagtail/utils/deprecation.py
@@ -49,7 +49,7 @@ class MovedDefinitionHandler(object):
 
             warnings.warn(
                 "%s has been moved from %s to %s" % (name, self.real_module.__name__, new_module_name),
-                category=self.warning_class
+                category=self.warning_class, stacklevel=2
             )
 
             # load the requested definition from the module named in moved_definitions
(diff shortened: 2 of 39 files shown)


wagtail/admin/models.py
```diff
@@ -1,5 +1,55 @@
...
+
+from modelcluster.fields import ParentalKey
+from taggit.models import Tag
+
 # The edit_handlers module extends Page with some additional attributes required by
 # wagtailadmin (namely, base_form_class and get_edit_handler). Importing this within
 # wagtailadmin.models ensures that this happens in advance of running wagtailadmin's
 # system checks.
 from wagtail.admin import edit_handlers  # NOQA
+from wagtail.core.models import Page
```
chosak: Your comment on https://github.com/wagtail/wagtail/issues/5507#issue-480767983 mentions that

> 'models' is a special case within Django, as it gets preloaded on startup - will this cause any side effects?

I am not that familiar with this behavior, so don't know what side effects might happen here, but I do think that using models.py as the filename for this might cause some confusion. Developers might expect admin/models.py to contain Django models for the admin app, as opposed to utility methods about models.

Could we name this file something else, maybe model_usage.py or usage.py?

wagtail/utils/deprecation.py
```diff
@@ -7,3 +11,53 @@ class RemovedInWagtail28Warning(DeprecationWarning):
...
+                # in the stack trace
+                raise e from None
+
+            warnings.warn(
+                "%s has been moved from %s to %s" % (name, self.real_module.__name__, new_module_name),
+                category=self.warning_class, stacklevel=2
+            )
+
+            # load the requested definition from the module named in moved_definitions
+            new_module = import_module(new_module_name)
```
chosak: If there's an exception when importing the relocated module, the stack trace confusingly contains the (handled) exception about the original import:

```
$ python -Wall
>>> from wagtail.admin.utils import get_object_usage
__main__:1: RemovedInWagtail29Warning: get_object_usage has been moved from wagtail.admin.utils to wagtail.admin.models
  File "/Users/chosaka/Projects/wagtail/wagtail/utils/deprecation.py", line 40, in __getattr__
    return getattr(self.real_module, name)
AttributeError: module 'wagtail.admin.utils' has no attribute 'get_object_usage'

  File "<stdin>", line 1, in <module>
  File "/Users/chosaka/Projects/wagtail/wagtail/utils/deprecation.py", line 56, in __getattr__
    new_module = import_module(new_module_name)
   ...
  File "/Users/chosaka/.virtualenvs/testwagtail/lib/python3.6/site-packages/django/conf/__init__.py", line 64, in _setup
    % (desc, ENVIRONMENT_VARIABLE))
django.core.exceptions.ImproperlyConfigured: Requested setting INSTALLED_APPS, but settings are not configured. You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.
```

If we were to move these remaining lines out of the `except` clause, [like this](https://gist.github.com/chosak/10d9b58de8d6a1165654ff37bc11437d), that'd be properly hidden:

```
$ python -Wall
Python 3.6.8 (default, May  5 2019, 07:40:18) 
[GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from wagtail.admin.utils import get_object_usage
__main__:1: RemovedInWagtail29Warning: get_object_usage has been moved from wagtail.admin.utils to wagtail.admin.models
  File "<stdin>", line 1, in <module>
  File "/Users/chosaka/Projects/wagtail/wagtail/utils/deprecation.py", line 56, in __getattr__
    new_module = import_module(new_module_name)
   ...
  File "/Users/chosaka/.virtualenvs/testwagtail/lib/python3.6/site-packages/django/conf/__init__.py", line 64, in _setup
    % (desc, ENVIRONMENT_VARIABLE))
django.core.exceptions.ImproperlyConfigured: Requested setting INSTALLED_APPS, but settings are not configured. You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.
```

wagtail/admin/models.py
gasman: Having pondered this some more... if wagtail.admin.models _did_ have any models of its own, I wouldn't have any second thoughts about putting these functions alongside them (just as, say, wagtail.documents.models has `get_document_model`) - and I don't think the situation is substantially different when models.py is otherwise empty. So I think this is fine - it just means that the functions get loaded slightly earlier in the startup sequence.

wagtail/utils/deprecation.py
gasman: Makes sense, cheers! Overly expansive try/catch blocks are never a good thing... fixed in 7cfa1fb8c1b5001fa000a3458b856b1d236187c5.

wagtail/admin/models.py
```diff
@@ -1,5 +1,55 @@
+from django.contrib.contenttypes.models import ContentType
+from django.db.models import Count
+
+from modelcluster.fields import ParentalKey
+from taggit.models import Tag
+
 # The edit_handlers module extends Page with some additional attributes required by
 # wagtailadmin (namely, base_form_class and get_edit_handler). Importing this within
 # wagtailadmin.models ensures that this happens in advance of running wagtailadmin's
```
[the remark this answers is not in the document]
gasman: Yep, I guess this was missed in the mass-rename for 2.0 because it didn't give the full module path... Fixed in 9a7d098049e9c8a986db53dc933c95e1540c6f5a.

PR #5518, opened 2019-08-19
[DOC] Add a note about specific field widget

johnraz:
Quick doc update.

### DIFF
--- a/docs/advanced_topics/customisation/custom_user_models.rst
+++ b/docs/advanced_topics/customisation/custom_user_models.rst
@@ -31,6 +31,7 @@ Create your custom user 'create' and 'edit' forms in your app:
   from django.utils.translation import ugettext_lazy as _
 
   from wagtail.users.forms import UserEditForm, UserCreationForm
+  from wagtail.admin.widgets import AdminDateTimeInput
 
   from users.models import MembershipStatus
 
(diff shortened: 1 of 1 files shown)


docs/advanced_topics/customisation/custom_user_models.rst
```diff
@@ -31,19 +31,24 @@ Create your custom user 'create' and 'edit' forms in your app:
...
+      access_expiry = forms.DateTimeField(required=True, label=_("Access expiry"), widget=AdminDateTimeInput)
 
 
   class CustomUserCreationForm(UserCreationForm):
       country = forms.CharField(required=True, label=_("Country"))
       status = forms.ModelChoiceField(queryset=MembershipStatus.objects, required=True, label=_("Status"))
+      access_expiry = forms.DateTimeField(required=True, label=_("Access expiry"), widget=AdminDateTimeInput)
 
+**Note:**
+For some fields like the ``DateTimeField`` you need to explicitly specify the widget in this case to enable the date picker.
```
gasman: Saying that you "need" to do something doesn't seem quite right here, as it suggests that failing to do so will produce incorrect behaviour. In fact, this is just standard Django behaviour, where a date/time field uses a standard input box unless overridden - it's just that certain _other_ forms in Wagtail provide the date picker as a special case.

It's also not particularly helpful for us to say "some fields" without saying which ones. Since Wagtail's custom form widgets (DateTimeField etc, rich text, page/image/document chooser...) are largely undocumented at the moment, this probably calls for us to add a new page under Reference detailing them all. The note here can then be something like:

> As this is a regular Django form, Wagtail's custom form widgets such as date pickers will not be automatically applied to form fields; to specify these, see [Form widgets].

johnraz: @gasman thanks a lot for the review. 

As I'm not sure I'll find time to write the ref page, could we just link to the code instead? 

Or eventually skip the `; to specify these, see [Form widgets].` part and move on? 

I know it's not ideal but I think it's better to have a short form than nothing. Let me know.

gasman: I think it's important for the documentation to be consistent about where the limits of the public documented API are. In this case, the steps required for getting a date picker onto the user edit form are currently outside those limits - which, of course, isn't ideal - but until we've addressed that by properly documenting the form widgets, I think it's a bad idea to allude to them in our code examples. Effectively it's saying "there's a way to achieve this, but you'll have to figure out the details yourself", which I'd argue is worse than not documenting the possibility at all.

johnraz: Sure it’s definitely your call, I’ll try to add the ref page when I get some free time. 👍🏻

lb-:
Closing as this is quite old and all date/datetime fields will mostly automatically use the datetime widget.
See `wagtail/admin/forms/models.py`.

Although, that may change in the future to be a non-widget variant - see https://github.com/wagtail/wagtail/issues/8056 & https://github.com/wagtail/wagtail/pull/10261#issuecomment-1609293863

PR #5520, opened 2019-08-20
fix(docs/customization): Add django.db.models import in the example

CuriousLearner:
This is a minor doc fix in one of the examples where fields from `django.db.models` are used but it is not imported in the first place.

The proposed change fixes this.


gasman:
Thanks @CuriousLearner! Merged in f9753f1f2328940231e40927e54249aa708adb62.

CuriousLearner:
Thanks @gasman :)

PR #5527, opened 2019-08-21
Add cache-busting parameter to admin static files

### LINKED ISSUE #5493: Add cache-busting parameters to Wagtail admin static files
Outdated CSS/JS persisting after a Wagtail upgrade is still a regular source of hard-to-diagnose deployment issues and bug reports (#5480, #5492). Adding cache-busting parameters was previously suggested in #3700 and #4966, and the solution adopted then was to encourage use of Django's `ManifestStaticFilesStorage` (#4745). However, this isn't sufficient - it only applies to production instances, and Wagtail can't (and shouldn't) mandate the use of specific Django-level configuration anyhow.

Following discussions on Slack, we've arrived at a plan to introduce a cache-busting mechanism that works independently of the Django staticfiles configuration:

* Introduce a `versioned_static` helper function and `{% versioned_static %}` template tag which wraps Django's `static` function as follows:
  * Call `static` on the passed file path to obtain a URL
  * If this URL already includes a querystring portion, return it unchanged. (This ensures we don't interfere with existing cache-busting / verification mechanisms already in place, such as S3 authenticated URLs.)
  * Otherwise, append a querystring `?v={hash}` where `hash` is the first 5-8 characters of an SHA1 hash formed from the Wagtail version number string and the project's `SECRET_KEY`. (Using a hash ensures that we don't leak the version number to untrusted users visiting the admin login page.)
* Update all static file references within the Wagtail admin (direct uses of `{% static %}` in templates, and also form media declarations as used on widgets, StreamField blocks and elsewhere) to use this new function / tag.

Note that cache-busting alone will not address all potential configuration errors, such as forgetting to run `collectstatic` when deploying a new Wagtail version. This will be addressed with a separate version check, to be outlined in another ticket.

gasman:
Fixes #5493

To be investigated:

* are there situations where the presence of a querystring disables caching entirely (and therefore this change would cause a performance hit)?
* would it be beneficial to disable this when `ManifestStaticFilesStorage` is active?

### DIFF
--- a/docs/advanced_topics/settings.rst
+++ b/docs/advanced_topics/settings.rst
@@ -510,6 +510,14 @@ can only choose between front office languages:
     LANGUAGES = WAGTAILADMIN_PERMITTED_LANGUAGES = [('en', 'English'),
                                                     ('pt', 'Portuguese')]
 
+Static files
+------------
+
+.. code-block:: python
+
+    WAGTAILADMIN_STATIC_FILE_VERSION_STRINGS = False
+
+Static file URLs within the Wagtail admin are given a version-specific query string of the form ``?v=1a2b3c4d``, to prevent outdated cached copies of Javascript and CSS files from persisting after a Wagtail upgrade. To disable these, set ``WAGTAILADMIN_STATIC_FILE_VERSION_STRINGS`` to ``False``.
 
 API Settings
 ------------
--- a/docs/releases/2.7.rst
+++ b/docs/releases/2.7.rst
@@ -53,6 +53,12 @@ Bug fixes
 Upgrade considerations
 ======================
 
+Query strings added to static file URLs within the admin
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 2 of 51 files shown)


gasman:
Neither Chrome nor Firefox seem to have any issues with caching files with a querystring in the URL. After browsing into and out of the page editor several times, developer tools show all the static files being cached (and the `runserver` console output confirms that there are no further requests for them):

Screenshot 2019-08-22 at 12 16 10
Screenshot 2019-08-22 at 12 27 41

gasman:
Cloudflare has [a configuration option for what happens to querystrings](https://support.cloudflare.com/hc/en-us/articles/200168256-What-are-Cloudflare-s-caching-levels-) - it's not clear what the default is.

If a user has it set to "Ignore Query String" and doesn't have some other cache-busting measure in place such as `ManifestStaticFilesStorage`, then they'll continue to experience outdated files on a Wagtail upgrade - not much we can do about that.

In "Standard" mode, querystring-based cache-busting will work as intended.

In "No Query String" mode, adding the querystring will prevent caching entirely. Notably, this means that if someone is successfully using `ManifestStaticFilesStorage` in production with this setting, rolling out this change will break caching. On that basis, it _does_ make sense for us to detect when `ManifestStaticFilesStorage` is active and disable the querystring in that case.

gasman:
Cloudfront has options for selectively ignoring individual query parameters, or treating them as meaningful: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/QueryStringParameters.html  - which means that depending on how it treats the `v` parameter, it will either be effective at cache-busting or make no difference. As far as I can see there are no possible configurations where the presence of the querystring will actually harm performance (unlike Cloudflare).

gasman:
Have now added a check to disable the querystring when a storage backend with hashed filenames such as `ManifestStaticFilesStorage` is in use, and added an upgrade considerations note to flag up the remaining edge case with Cloudflare (i.e. using "No Query String" mode with some mechanism other than `ManifestStaticFilesStorage` to handle expiry, in which case we'll make things worse by adding the querystring).

Hopefully this means the above queries are now accounted for, and this is ready for review...

gasman:
Merged in e1d3390a1f62ae4b30c0ef38e7cfa5070d8565dc

Issue #4310, opened 2018-02-23
 Language chosen in user preferences persists on subsequent requests


JRafiei:
### Issue Summary

When I change the Language preferences setting from account settings in wagtail admin panel, all the translations across the website changes too. Is it the expected behavior of a bug?

### Steps to Reproduce

1. Go to the wagtail admin panel
2. Click on the Account settings section
3. Click on the Language preferences button
4. Select another language, English for example
5. Save changes and log out

Now you should see the sections that use internationalization tags like trans get changed. At least that's my experience 

### Technical details

* Python version: 3.5.2
* Django version: 1.11.7
* Wagtail version: 1.13
* Browser version: Firefox 56.0 and Chrome 61.0.3
* OS: Ubuntu 16.04

gasman:
Hi @JRafiei - I can't replicate this as described. I'm running the following steps against the [bakerydemo](https://github.com/wagtail/bakerydemo) project:

* Edit bakerydemo/templates/base.html to add `{% load i18n %}` and `<h1>{% trans "Messages" %}</h1>` to the top of the page. (I've chosen "Messages" as a string that exists in Django's own translation set)
* Temporarily change `LANGUAGE_CODE` in bakerydemo/settings/base.py from `en-us` to `de` and view http://localhost:8000/ - confirm that the text changes to "Mitteilungen"
* Change `LANGUAGE_CODE` back to `en-us`
* Log in to the admin, go to Account settings -> Language preferences and set language to German
* Visit http://localhost:8000/ - the text still says "Messages", not "Mitteilungen" (and remains that way after I log out)

This is the behaviour I'd expect - the language choice within the admin is applied using [a decorator on `/admin` URLs](https://github.com/wagtail/wagtail/blob/6dc89c1f7f7d846594d093f1e52efcf9d38ab20c/wagtail/wagtailadmin/decorators.py#L28-L29), so I can't see how that choice would "leak" into the language of the website. In fact, we've encountered the opposite problem (#3991) - the Wagtail userbar picks up the language of the website instead of the admin.

If you can provide complete steps for reproducing this bug on a fresh Wagtail instance, let me know and I'll reopen this.

JRafiei:
Hello @gasman,
I haven't checked this behavior on the localhost and after your suggestion, I retry these steps on my PC and didn't find any problem. Sorry for incomplete testing before reporting this issue.
[...]

PR #5532, opened 2019-08-22
fix: Language chosen in user preferences persists (#4310)

bmihelac:

### DIFF
--- a/wagtail/admin/auth.py
+++ b/wagtail/admin/auth.py
@@ -7,8 +7,8 @@ from django.db.models import Q
 from django.shortcuts import redirect
 from django.urls import reverse
 from django.utils.timezone import activate as activate_tz
-from django.utils.translation import activate as activate_lang
 from django.utils.translation import ugettext as _
+from django.utils.translation import override
 
 from wagtail.admin import messages
 from wagtail.core.models import GroupPagePermission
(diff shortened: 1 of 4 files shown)


wagtail/admin/auth.py
```diff
@@ -154,13 +155,19 @@ def decorated_view(request, *args, **kwargs):
...
+                if language != get_language():
+                    original_language = get_language()
+                    activate_lang(language)
                 time_zone = user.wagtail_userprofile.get_current_time_zone()
                 activate_tz(time_zone)
-            return view_func(request, *args, **kwargs)
+            response = view_func(request, *args, **kwargs)
+            if original_language:
+                activate_lang(original_language)
+            return response
```
gasman: Could we use `django.utils.translation.override` here, as you've done for the `language_preferences` view? It seems a lot neater, and I have more confidence that Django's implementation will handle the edge cases correctly (e.g. what happens if the view throws an exception) than our own code.

bmihelac: good point, I'll update ASAP

bmihelac: updated

gasman:
Thanks @bmihelac! Merged in df1608b08031d259970943adff87d2d078b77b68.

bmihelac:
Thanks for review and merging @gasman

PR #5534, opened 2019-08-23
Update release schedule documentation

gasman:
Update release schedule information to reflect the new three-month release cycle, and move the date listing to the wiki where it can be kept up to date more effectively.


nmorduch:
Merged with e263aaf74958312fa8e471c1406af1d702f6ff83

PR #5538, opened 2019-08-25
Fix compatibility with pylint-django

danihodovic:
It's not possible to use wagtail in a project with pylint and the
pylint-django plugin.

Pylint-django does not work with foreign keys that are referenced by
their string names. See: https://github.com/PyCQA/pylint-django#known-issues

If we simply replace the string name with the foreign-key class, linting
on wagtail projects is fixed.

Also see:
https://github.com/PyCQA/pylint/issues/2995
https://github.com/PyCQA/pylint-django/issues/241

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -246,7 +246,7 @@ class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):
         help_text=_("The name of the page as it will appear in URLs e.g http://domain.com/blog/[my-slug]/")
     )
     content_type = models.ForeignKey(
-        'contenttypes.ContentType',
+        ContentType,
         verbose_name=_('content type'),
         related_name='pages',
         on_delete=models.SET(get_default_page_content_type)


gasman:
Thanks @danihodovic! Merged in af4f27aa3ec6d704a9a2f71ea81b0a3572502433.

danihodovic:
Thanks @gasman when is 2.6.2 scheduled to be released?

gasman:
@danihodovic There are no current plans for a 2.6.2 release - that'll only happen if we encounter a bug that's serious enough to require a fix before the 2.7 release. The fix here will be in 2.7, which is due for release on 1st November.

For now I've merged this commit into the stable/2.6.x branch, so if a 2.6.2 release _does_ happen, it will include this fix.

danihodovic:
Awesome thanks!

Thijss:
So, does this open the door for other pull requests that change string based references to class based ones? My pylint-django config is currently failing on wagtail.contrib.settings.models.BaseSetting which uses string reference as well.

PR #5543, opened 2019-08-29
Fixed broken email header when hosting static files remotely

### LINKED ISSUE #5525: Broken email header image when hosting static files remotely
The base template for notification emails uses `settings.BASE_URL` prepended to the output of the `static` tag to produce an absolute URL:

https://github.com/wagtail/wagtail/blob/d6e4072e26ab7291d2aa56c4e87c0d82c7fb9030/wagtail/admin/templates/wagtailadmin/notifications/base.html#L89

This will produce a broken URL if `{% static %}` already returns a full URL - for example, if static assets are hosted on Amazon S3 using `django-storages`.

### Technical details

* Python version: 3.5.2
* Django version: 2.1.11
* Wagtail version: 2.7a0

edu2004eu:
Fixes #5525

* Do the tests still pass? (http://docs.wagtail.io/en/latest/contributing/developing.html#testing) **Yes**
* Does the code comply with the style guide? (Run `make lint` from the Wagtail root) **Yes**
* For Python changes: Have you added tests to cover the new/fixed behaviour? **No -- didn't think it was required for such a small change, but if needed, I can write them**
* For front-end changes: Did you test on all of Wagtail’s supported browsers? **Please list the exact versions you tested**. **N/A**

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/notifications/base.html
+++ b/wagtail/admin/templates/wagtailadmin/notifications/base.html
@@ -1,4 +1,4 @@
-{% load i18n static %}
+{% load i18n wagtailadmin_tags %}
 <!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
 <html xmlns="http://www.w3.org/1999/xhtml">
 <head>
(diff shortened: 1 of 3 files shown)


gasman:
Thanks @edu2004eu! Code here looks good.
It would be good to have a test for this, so that we've got a permanent record of the fix and to ensure it doesn't get broken again in any future refactor.

edu2004eu:
@gasman I don't really have much experience with unit testing, but I did my best. Let me know if that's OK.

I've also changed the way I get the `BASE_URL` setting, since tests failed if the settings file didn't define one.

gasman:
Thanks again! Merged in b9cfc24b37979b9c8528fd0466e39758170f90c7, with some minor tweaks to the test to make it more specific.

PR #5545, opened 2019-08-30
Display data for twitter embed block. Fixes #5176

### LINKED ISSUE #5176: Support twitter previews in Draftail embeds

imansyed000:
Replaces #5486. Fixes #5176

Tested on Chrome and Safari. 
Added new tests

image

If no thumbnails available via oembed, returns 'Provider name - Author name'


client/src/components/Draftail/blocks/EmbedBlock.js
```diff
@@ -4,16 +4,20 @@ import React from 'react';
 import { STRINGS } from '../../../config/wagtailConfig';
 
 import MediaBlock from '../blocks/MediaBlock';
+import { shortenLabel } from '../decorators/TooltipEntity'
```
thibaudcolas:
```suggestion
import { shortenLabel } from '../decorators/TooltipEntity';
```

Missing a semicolon here.

client/src/components/Draftail/decorators/TooltipEntity.js
```diff
@@ -5,7 +5,10 @@ import { Icon } from 'draftail';
 import Tooltip from '../Tooltip/Tooltip';
 import Portal from '../../Portal/Portal';
 
-const shortenLabel = (label) => {
+/**
+ * Shortens the given label if it goes beyond a predetermined limit.
+ */
+export const shortenLabel = (label) => {
```
thibaudcolas: I think we might want to write a couple of tests for this function now that it’s reused elsewhere – basically test what happens with a label shorter than 25 chars, and longer.

client/src/components/Draftail/blocks/MediaBlock.js
```diff
@@ -90,9 +90,13 @@ class MediaBlock extends Component {
         <span className="MediaBlock__icon-wrapper" aria-hidden>
           <Icon icon={entityType.icon} className="MediaBlock__icon" />
         </span>
-
         <img className="MediaBlock__img" src={src} alt={alt} width="256" />
 
+        {src ? null : (
```
thibaudcolas: We’ll need to update the test case / add a new test case for this, checking that `fallbackText` does display.

client/src/components/Draftail/blocks/EmbedBlock.js
```diff
@@ -4,16 +4,20 @@ import React from 'react';
...
   const { entity, onRemoveEntity } = props.blockProps;
-  const { url, title, thumbnail } = entity.getData();
+  const { url, title, thumbnail, providerName, authorName } = entity.getData();
+  // Fallback text is used when there is no image available for the embed.
+  // In those cases, it seems like author and provider name will always be present.
+  const fallbackText = `${shortenLabel(authorName)}\n${shortenLabel(providerName)}`;
 
   return (
-    <MediaBlock {...props} src={thumbnail} alt="">
+    <MediaBlock {...props} src={thumbnail} alt="" fallbackText={fallbackText}>
```
thibaudcolas: Might need to update the existing test case(s) to check that this prop is set as expected.

client/src/components/Draftail/blocks/EmbedBlock.js
```diff
@@ -4,16 +4,20 @@ import React from 'react';
...
 /**
  * Editor block to display media and edit content.
  */
 const EmbedBlock = props => {
   const { entity, onRemoveEntity } = props.blockProps;
-  const { url, title, thumbnail } = entity.getData();
+  const { url, title, thumbnail, providerName, authorName } = entity.getData();
+  // Fallback text is used when there is no image available for the embed.
+  // In those cases, it seems like author and provider name will always be present.
+  const fallbackText = `${shortenLabel(authorName)}\n${shortenLabel(providerName)}`;
```
thibaudcolas: I think we might want to make this a bit smarter so it works well regardless of whether those things are defined or not (_just in case_). Something like:

```suggestion
  const fallbackText = [authorName, providerName].filter(text => !!text).map(shortenLabel).join('\n');
```

thibaudcolas:
Replaced by #8167.

PR #5550, opened 2019-09-05
Replaced SafeText with SafeString.

atombrella:
https://github.com/django/django/blob/master/django/utils/safestring.py#L40


gasman:
Thanks @atombrella!

PR #5551, opened 2019-09-05
Replaced use of force_text with force_str.

atombrella:
Will be deprecated in Django 3.0.

The comment seems a bit dated, so I guess it could be removed. `SafeBytes` was removed in the removal of support for Python 2.

### DIFF
--- a/wagtail/admin/compare.py
+++ b/wagtail/admin/compare.py
@@ -1,7 +1,7 @@
 import difflib
 
 from bs4 import BeautifulSoup
-from django.utils.encoding import force_text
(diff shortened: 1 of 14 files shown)


gasman:
Thanks! Yep, it looks like that comment originally referred to the `from __future__ import unicode_literals` in https://github.com/wagtail/wagtail/blob/31972dfd9ee7b4239d3ffe7b25725ea3cdac1c10/wagtail/wagtailcore/blocks/base.py , but got orphaned when we started using `isort` on imports...

PR #5552, opened 2019-09-05
Updated the stream block id initialization

colink-eb:
The current block id generation only sets the id as the block is serialized for storage in the database, which means that the id is unavailable in the block until it is pulled back from the database. In my debugging this caused the id to be set to new values up to 3 times when saving a brand new page (each time with a new id).
This updated logic applies the new id to the actual block which makes it available right away and prevents the id from being regenerated.


gasman:
Thanks @colink-eb! Merged in 33204be10586128d44156b8a3b91a5cf2c0db4d0, with a small addition to the tests to cover this case.

PR #4721, opened 2018-08-06
Bundle the l18n library to prevent installation issues

gasman:
https://bitbucket.org/tkhyn/l18n/issues/4/setuppy-raises-unicodedecodeerror-trying prevents the l18n library from being installed on certain setups with non-Unicode locales:
https://groups.google.com/d/msg/wagtail/X5d8CL3DxUs/httaLso4DAAJ

As a temporary measure until a fixed version of l18n is released, bundle the library into wagtail.utils.l18n.

PR #5555, opened 2019-09-06
Unbundle the l18n library

gasman:
This was previously bundled in #4721 to avoid installation errors on non-Unicode locales. As of v2018.5, this has now been fixed upstream: https://bitbucket.org/tkhyn/l18n/issues/4/setuppy-raises-unicodedecodeerror-trying


lb-:
Rebased and resolved via 6ce9f3a717abd3cb7a9ae35364ae274f141c267b

Issue #5562, opened 2019-09-11
Soundcloud embed doesn't work with links to artist


kstaikov:
### Issue Summary

https://soundcloud.com/hear-the-bern us a valid url that returns an embeddable player with a list of tracks in the soundcloud oembed API. However, this is not a valid embed URL in wagtail, because it expects an extra layer of depth in the URL (hear-the-bern/something-else) - https://github.com/wagtail/wagtail/blob/master/wagtail/embeds/oembed_providers.py#L221-L226. It can currently be hacked by adding a question mark to the URL being sent in (hear-the-bern/?), but this is not very user friendly.

### Steps to Reproduce

1. In the richtext editor, try to embed https://soundcloud.com/hear-the-bern. You get an error about embeddable URL not being found.
2. See that `curl "https://soundcloud.com/oembed"  -d 'format=json' -d 'url=https://soundcloud.com/hear-the-bern/?'` returns a valid embed information.

gasman:
Fixed in #5564

PR #5564, opened 2019-09-11
Fix soundcloud embedding of artist page

kstaikov:
Fixes https://github.com/wagtail/wagtail/issues/5562. Artist page could not be embedded because URL expected extra layer of depth (i.e. soundcloud.com/artist/track worked but soundcloud.com/artist did not)


gasman:
Thanks @kstaikov! Merged in 4b692d67d5518c85c32d58c82f2cd75de838e20b.

PR #5568, opened 2019-09-13
allow image page sizes to be configurable

noslouch:
When wagtail's serving thousands of images, sometimes paging through the images (and filtered result sets) can get tedious. This feels like an unobtrusive way to allow user-land to edit the response size.


noslouch:
bump?

gasman:
Thanks @noslouch! Merged in 34f4ab8a3fcfcb532793f30cb5f6f5c3a0820253 (with some changes to the documentation wording - I think it's clearer to say 'items per page' rather than 'page size).

noslouch:
Thanks @gasman!

PR #5571, opened 2019-09-18
Fix help-text example in "Extending the Draftail Editor"

### LINKED ISSUE #5570: Example in "Extending the Draftail Editor" does not work.
Due to wrong markup in converter rule the example with help-text does not work. Using div[class=help-text] instead of div.help-text makes it work correctly.

buugaj:
Due to wrong markup in converter rule the example with help-text does not work. Using div[class=help-text] instead of div.help-text makes it work correctly.

Fixes #5570


gasman:
Thanks @buugaj! Merged in 90b7a3b0d38993f46ee585d4a4dfdfc6585cbde2.

PR #5581, opened 2019-09-25
Updating FieldDoesNotExist imports to all be django.core.exceptions.

cyface:
Django 3 removes the FieldDoesNotExist exception from the django.db.models.fields package. This PR updates all places where Wagtail was importing it from that package, and replaces it with imports from the django.core.exceptions package.  

That exception was moved to that package in 2015; so currently-compatible versions of Django should work fine with this change.

### DIFF
--- a/wagtail/admin/edit_handlers.py
+++ b/wagtail/admin/edit_handlers.py
@@ -2,8 +2,7 @@ import functools
 import re
 
 from django import forms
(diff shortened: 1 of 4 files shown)


gasman:
Thanks @cyface! Merged in f44d27642b4a6932de73273d8320bbcb76330c21 - looks like there's a corresponding fix needed in django-modelcluster, but we can deal with that separately.

PR #5584, opened 2019-09-27
Add documentation for how to use alternative image tags

kaedroho:
This adds documentation for alternative image tags such as `<picture>` and `<amp-img>`.

### DIFF
--- a/docs/topics/images.rst
+++ b/docs/topics/images.rst
@@ -204,11 +204,21 @@ Therefore, if you'd added the field ``author`` to your AbstractImage in the abov
     The image property used for the ``src`` attribute is actually ``image.url``, not ``image.src``.
 
 
-Alternative HTML tags
-`````````````````````
+The ``attrs`` shortcut
+-----------------------
 
-The `as` syntax also allows alternative HTML image tags (such as `<picture>` or `<amp-img>`) to be used.
-For example, to use the `<picture>` tag.
+You can also use the ``attrs`` property as a shorthand to output the attributes ``src``, ``width``, ``height`` and ``alt`` in one go:
+
+.. code-block:: html+django
+
+    <img {{ tmp_photo.attrs }} class="my-custom-class" />
+
+
+Alternative HTML tags
+---------------------
+
+The ``as`` syntax allows alternative HTML image tags (such as ``<picture>`` or ``<amp-img>``) to be used.
+For example, to use the ``<picture>`` tag:
 
 .. code-block:: html+django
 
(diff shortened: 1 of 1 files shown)


docs/topics/images.rst
```diff
@@ -204,6 +204,29 @@ Therefore, if you'd added the field ``author`` to your AbstractImage in the abov
...
+
+.. code-block:: html+django
+
+    <picture>
+        {% image page.photo width-800 as wide_photo %}
+        <source srcset="{{ wide_photo.url }}" media="(min-width: 800px)">
+        {% image page.photo width-400 %}
+    </picture>
+
+And to use the `<amp-img>` tag.
```
tomdyson: tag. -> tag:

docs/topics/images.rst
```diff
@@ -204,6 +204,29 @@ Therefore, if you'd added the field ``author`` to your AbstractImage in the abov
...
+
+.. code-block:: html+django
+
+    {% image page.photo width-400 as photo %}
+    <amp-img src="{{ photo.url }}" width="{{ photo.width }}"
```
tomdyson: Should we use a more detailed example, e.g. https://amp.dev/documentation/components/amp-img/#example:-specifying-a-fallback-image (although I'm not clear about the feasibility of WebP in Wagtail)?

zerolab: once #5249 gets in, webp will be relevant

docs/topics/images.rst
```diff
@@ -204,6 +204,29 @@ Therefore, if you'd added the field ``author`` to your AbstractImage in the abov
     The image property used for the ``src`` attribute is actually ``image.url``, not ``image.src``.
 
 
+Alternative HTML tags
+`````````````````````
```
gasman: Sphinx is complaining about inconsistent title levels here, presumably because we use `^` instead of backticks to indicate subheadings in the JPEG section below.

In this case I'd suggest shuffling things around so that this is the same level as its neighbours, though:

* "The attrs shortcut" should be moved above this, because the use of the `tmp_photo` variable directly follows on from the example code in "More control over the img tag"
* "Alternative HTML tags" should then be another top-level section (underlined with dashes).

docs/topics/images.rst
```diff
@@ -204,6 +204,29 @@ Therefore, if you'd added the field ``author`` to your AbstractImage in the abov
...
+
+The `as` syntax also allows alternative HTML image tags (such as `<picture>` or `<amp-img>`) to be used.
```
gasman: We should use double backticks for `<picture>` and `<amp-img>` here, so that they appear as monospace rather than italics.

gasman:
Other than the formatting tweaks, looks good to me!

kaedroho:
All review feedback applied!

docs/topics/images.rst
```diff
@@ -218,23 +228,23 @@ For example, to use the `<picture>` tag.
...
-
-The ``attrs`` shortcut
------------------------
-
-You can also use the ``attrs`` property as a shorthand to output the attributes ``src``, ``width``, ``height`` and ``alt`` in one go:
-
-.. code-block:: html+django
-
-    <img {{ tmp_photo.attrs }} class="my-custom-class" />
+    {% image image width-550 format-webp as webp_image %}
```
gasman: Is this conditional on #5249 getting merged? If so, perhaps we ought to stick with a jpeg-only example until we're confident that webp support will make it into 2.7.

tomdyson: Let's do them both! @zerolab, can you make sure that #5249 happens?

PR #5586, opened 2019-09-28
Allow the URL used by Page._get_dummy_headers() to be overridden

ababic:
So, this came about because wagtailmenus has an `AbstractLinkPage` page type with some custom previewing behaviour which breaks badly in testing with the new dummy request creation implementation. 

This is because the page type overrides the various URL methods to return the target external/page URL instead. Previewing one of these pages renders a simple string that confirms "This page redirects to http://example.com", but this now results in a 400 error in some cases, because the target URL might be on a domain not permitted by `ALLOWED_HOSTS`.

I know Wagtail cannot possibly cater for every single case where some developer/third-party app is doing something strange :) But, this small/low-risk change will allow me to fix `AbstractLinkPage` without having to override the entire `_get_dummy_headers()` method, so thought it was worth proposing?

- [ ] For Python changes: Have you added tests to cover the new/fixed behaviour?

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -1264,7 +1264,7 @@ class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):
         Return a dict of META information to be included in a faked HttpRequest object to pass to
         serve_preview.
         """
-        url = self.full_url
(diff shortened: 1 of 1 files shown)


gasman:
Looks good to me - thanks @ababic!

PR #5588, opened 2019-09-29
Add role="table" to TableBlock output

gasman:
As per https://twitter.com/SaraSoueidan/status/1177622630763028480, certain browsers apply heuristics to decide whether `<table>` elements exist for layout or data purposes, and adjust the behaviour of their accessibility features accordingly. Given that TableBlock intentionally doesn't allow markup within cells, we can be reasonably sure that any tables created with it are genuine data tables, and should therefore indicate that using `role="table"`.


gasman:
Merged in 9a21e79ff88e13d83c81d01f620313a4ddf16b14

Issue #3572, opened 2017-04-28
Feature request: ChoiceBlock should allow multiple choices


kunambi:
It could be another parameter e.g. `multiple=True`.

### Example
```
    align = blocks.ChoiceBlock(label=_('Align'), required=True, choices=[
        ('align-left', _('Left')),
        ('align-right', _('Right')),
    ], help_text=_('On which side should the image show?'), multiple=True)
```

### Expected behaviour:
Print the choices as a series of `<input type="checkbox">` in the Wagtail admin

### Expected result:
Send the chosen values as an iterable list to the template

gasman:
Thanks for the suggestion @kunambi! I think a separate `MultipleChoiceBlock` block type would be a better fit for this, since the process for outputting multiple values on the template (and storing them in the database) would be substantially different from handling a single value.

gasman:
Implemented in #5592 / #5819

Issue #4407, opened 2018-03-20
Implement a ModelChoiceBlock block type


gasman:
Wagtail currently does not provide a StreamField block equivalent of Django's ModelChoiceField (for giving the user a dropdown of model instances to choose from). People often try using the undocumented internal `ChooserBlock` class for this purpose, but this was only designed for use with our modal choosers (`AdminPageChooser` etc) and can't be relied on to work correctly with a basic `Select` widget; consequently, various PRs have been submitted to fix or extend this functionality (#2714, #3169, #4157, #4351, #4380).

Rather than trying to patch `ChooserBlock` to work with the `Select` widget, I propose that we introduce a new `ModelChoiceBlock`, for the following reasons:

* Consistency with Django's naming scheme
* Where the API of our modal chooser widgets differs from the `Select` widget, we don't have to fudge a single implementation that handles both cases, and may end up being non-optimal with modal choosers. (For example, see https://github.com/wagtail/wagtail/pull/3169#issuecomment-262815168 - our modal widgets are most efficient when the render method receives a model instance, rather than an ID, and the `bulk_to_python` optimisation relies on us being able to do this. However, the `Select` widget only accepts the ID.)
* It avoids creating the expectation that `PageChooserBlock` etc will support all options and functionality of `ModelChoiceBlock`; for example, the modal choosers can't realistically support custom `get_queryset` methods (without a LOT of extra development).

The new `ModelChoiceField` will need documentation, and tests for all methods (to cover all relevant round-trips between database, form and front-end representations); it should also implement a `get_queryset` method as per #4351. `ChooserBlock` might possibly be refactored as a subclass of `ModelChoiceField`, depending on whether there's enough shared logic to make it worthwhile; either way, it should remain as an internal-only class existing for the benefit of `PageChooserBlock` etc.

thibaudcolas:
This sounds like the following packages:

- https://github.com/springload/wagtailmodelchoosers
- https://github.com/Naeka/wagtailmodelchooser
- https://github.com/takeflight/wagtailmodelchooser

I'm not sure how relevant these are compared to your proposed implementation (I'd expect those packages to do much more UI-wise) but thought it would be worth highlighting.

gasman:
[...]

PR #5592, opened 2019-10-01
Implement MultipleChoiceBlock

SmileyJames:
Issue that requests the implemented feature: https://github.com/wagtail/wagtail/issues/3572

PR implements a multiple choice block. Also adds widget argument to ChoiceBlock and new MultipleChoiceBlock.

Tests added. Documentation updated.


pySilver:
I miss this feature in core too!

pySilver:
I believe deconstruct method shouldn’t track ‘widget’ value, otherwise it would require widget instance and all its params to be serializable too, which is not possible in many situations. Moreover widget doesn’t influence data structure, so there is no reason to track it I guess.

Someone should look at this code, I think its a valuable addition to Wagtail. 

If you are looking for Model choice or multiple model choice blocks check this out: https://github.com/wagtail/wagtail/issues/4407#issuecomment-558771219

SmileyJames:
> I believe deconstruct method shouldn’t track ‘widget’ value, otherwise it would require widget instance and all its params to be serializable too, which is not possible in many situations. Moreover widget doesn’t influence data structure, so there is no reason to track it I guess.
I think you might be right about this. Perhaps the core maintainers could weigh in on this. I am confused why blocks are saved to migrations anyway, given that they are just saved to a varchar, regardless of what the migration looks like.

> Someone should look at this code, I think its a valuable addition to Wagtail.
Very kind words, thank you.

zerolab:
@SmileyJames 

> I am confused why blocks are saved to migrations anyway, given that they are just saved to a varchar, regardless of what the migration looks like.

To quote @gasman from Slack

> They're included for the same reason that Django includes the rest of the field definition (including things that have no effect on the database structure) - it means that you have a working version of the ORM to use in your own migration code (e.g. data migrations), as it existed at that moment in time.

gasman:
Thanks @SmileyJames! Sorry for missing this earlier...

Block definitions are frozen in migrations for the same reason that Django freezes various other parameters that don't affect the database structure - it enables Django to build a (more-or-less) functional version of the ORM as it existed at that point in time. In the case of block definitions, it makes it possible to access StreamField values within a data migration and work with them in terms of blocks, rather than raw JSON. (In practice it's a bit more complicated than that - if you're writing migrations to add a StructBlock wrapper around an ImageChooserBlock, then you need to arrange things in such a way that the frozen ORM can understand both the 'before' and 'after' states of the jason. But in any case, having the block definitions in the migration is a necessary condition for doing that, if not a sufficient one...)

Now, I think it's fair to say that this design decision on Django's side isn't entirely successful: since the frozen ORM only contains the fields, not the model methods, it isn't fully functional (this is a particular annoyance for Wagtail Page models, because the frozen versions are missing the django-treebeard methods that allow us to alter tree structure without breaking data integrity), and there are fields like `help_text` where the maintenance burden of creating migrations on every minor change is clearly not worth the potential usefulness of having access to them within migrations.

So, I think there's a good argument for breaking away from Django's "freeze everything into migrations" approach in certain cases, and the widget parameter is one of those cases - I don't think it's ever going to be useful to refer to it within migration code. We just need to test it carefully, since it's breaking with Django's default behaviour - for example, we need to make sure that it doesn't see the absence of `widget` parameter as a change to the model and end up endlessly creating new migrations.

pySilver:
>So, I think there's a good argument for breaking away from Django's "freeze everything into migrations" approach in certain cases, and the widget parameter is one of those cases - I don't think it's ever going to be useful to refer to it within migration code. We just need to test it carefully, since it's breaking with Django's default behaviour - for example, we need to make sure that it doesn't see the absence of widget parameter as a change to the model and end up endlessly creating new migrations.

Thankfully not dumping `widget` does not cause endlessly creating new migrations.

wagtail/core/blocks/field_block.py
```diff
@@ -386,20 +391,22 @@ def __init__(self, choices=None, default=None, required=True, help_text=None, va
...
             self._constructor_kwargs['help_text'] = help_text
+        if widget is not None:
+            self._constructor_kwargs['widget'] = widget
 
         # We will need to modify the choices list to insert a blank option, if there isn't
         # one already. We have to do this at render time in the case of callable choices - so rather
         # than having separate code paths for static vs dynamic lists, we'll _always_ pass a callable
         # to ChoiceField to perform this step at render time.
 
         # If we have a default choice and the field is required, we don't need to add a blank option.
```
gasman: This comment line should be moved into `ChoiceBlock.get_callable_choices`.

Also, I think it's worth us renaming it to `_get_callable_choices` at this point, to clarify that this is an internal method - the fact that it's being inherited / overridden here is liable to get people thinking this is the right place to implement their custom return-a-list-of-choices logic.

wagtail/core/blocks/field_block.py
```diff
@@ -386,20 +391,22 @@ def __init__(self, choices=None, default=None, required=True, help_text=None, va
             self._constructor_kwargs['required'] = required
         if help_text is not None:
             self._constructor_kwargs['help_text'] = help_text
+        if widget is not None:
+            self._constructor_kwargs['widget'] = widget
```
gasman: As per https://github.com/wagtail/wagtail/pull/5592#issuecomment-558861585 + subsequent discussion I believe we want to leave this out of `_constructor_kwargs`, so that we don't end up including it in frozen model definitions in migrations (where it will fail because Widget objects aren't serializable).

pySilver:
@gasman @SmileyJames  I'm using MultipleChoiceBlock's very intensively in our project but it's a little bit different than this code (which I've used as a starting point). Unfortunately I cannot make a proper PR right now, but maybe you guys can review and adopt some of my code.

Bonus features:

* ModelChoiceBlock
* ModelMultipleChoiceBlock


```python
from typing import List, Optional

from django import forms
from django.db.models import BLANK_CHOICE_DASH, Model, QuerySet
from django.forms.fields import CallableChoiceIterator
from django.utils.encoding import force_str
from django.utils.functional import cached_property

from wagtail.core.blocks import FieldBlock
from wagtail.core.utils import resolve_model_string


class BaseChoiceBlock(FieldBlock):
    choices = ()

    def __init__(
        self,
        choices=None,
        default=None,
        required=True,
        help_text=None,
        widget=None,
        validators=(),
        **kwargs
    ):

        self._required = required
        self._default = default

        if choices is None:
            # no choices specified, so pick up the choice defined at the class level
            choices = self.choices

        if callable(choices):
            # Support of callable choices. Wrap the callable in an iterator so that we can
            # handle this consistently with ordinary choice lists;
            # however, the `choices` constructor kwarg as reported by deconstruct() should
            # remain as the callable
            choices_for_constructor = choices
            choices = CallableChoiceIterator(choices)
        else:
            # Cast as a list
            choices_for_constructor = choices = list(choices)

        # keep a copy of all kwargs (including our normalised choices list) for deconstruct()
        self._constructor_kwargs = kwargs.copy()
        self._constructor_kwargs["choices"] = choices_for_constructor
        if required is not True:
            self._constructor_kwargs["required"] = required
        if help_text is not None:
            self._constructor_kwargs["help_text"] = help_text

        # We will need to modify the choices list to insert a blank option, if there isn't
        # one already. We have to do this at render time in the case of callable choices - so rather
        # than having separate code paths for static vs dynamic lists, we'll _always_ pass a callable
        # to ChoiceField to perform this step at render time.

        # If we have a default choice and the field is required, we don't need to add a blank option.
        callable_choices = self.get_callable_choices(choices)
        self.field = self.get_field(
            choices=callable_choices,
            required=required,
            help_text=help_text,
            validators=validators,
            widget=widget,
        )
        super().__init__(default=default, **kwargs)

    def get_callable_choices(self, choices, blank_choice=True):
        """
        Return a callable that we can pass into `forms.ChoiceField`, which will provide the
        choices list with the addition of a blank choice (if blank_choice=True and one does not
        already exist).
        """

        def choices_callable():
            # Variable choices could be an instance of CallableChoiceIterator, which may be wrapping
            # something we don't want to evaluate multiple times (e.g. a database query). Cast as a
            # list now to prevent it getting evaluated twice (once while searching for a blank choice,
            # once while rendering the final ChoiceField).
            local_choices = list(choices)

            # If blank_choice=False has been specified, return the choices list as is
            if not blank_choice:
                return local_choices

            # Else: if choices does not already contain a blank option, insert one
            # (to match Django's own behaviour for modelfields:
            # https://github.com/django/django/blob/1.7.5/django/db/models/fields/__init__.py#L732-744)
            has_blank_choice = False
            for v1, v2 in local_choices:
                if isinstance(v2, (list, tuple)):
                    # this is a named group, and v2 is the value list
                    has_blank_choice = any([value in ("", None) for value, label in v2])
                    if has_blank_choice:
                        break
                else:
                    # this is an individual choice; v1 is the value
                    if v1 in ("", None):
                        has_blank_choice = True
                        break

            if not has_blank_choice:
                return BLANK_CHOICE_DASH + local_choices

            return local_choices

        return choices_callable

    class Meta:
        # No icon specified here, because that depends on the purpose that the
        # block is being used for. Feel encouraged to specify an icon in your
        # descendant block type
        icon = "placeholder"


class MultipleChoiceBlock(BaseChoiceBlock):
    def get_field(self, **kwargs):
        return forms.MultipleChoiceField(**kwargs)

    def get_callable_choices(self, choices, blank_choice=False):
        """ Override to default blank choice to False
        """
        return super().get_callable_choices(choices, blank_choice=blank_choice)

    def deconstruct(self):
        """
        Always deconstruct MultipleChoiceBlock instances as if they were plain
        MultipleChoiceBlocks with their choice list passed in the constructor,
        even if they are actually subclasses. This allows users to define
        subclasses of MultipleChoiceBlock in their models.py, with specific choice
        lists passed in, without references to those classes ending up frozen
        into migrations.
        """
        return (
            "wagtail.core.blocks.fields.MultipleChoiceBlock",
            [],
            self._constructor_kwargs,
        )

    def get_searchable_content(self, value):
        # Return the display value as the searchable value
        content = []
        text_value = force_str(value)
        for k, v in self.field.choices:
            if isinstance(v, (list, tuple)):
                # This is an optgroup, so look inside the group for options
                for k2, v2 in v:
                    if value == k2 or text_value == force_str(k2):
                        content.append(force_str(k))
                        content.append(force_str(v2))
            else:
                if value == k or text_value == force_str(k):
                    content.append(force_str(v))
        return content


class ModelChoiceBlock(FieldBlock):
    class Meta:
        icon = "placeholder"

    def __init__(
        self,
        target_model=None,
        queryset=None,
        to_field_name=None,
        limit_choices_to=None,
        required=True,
        help_text=None,
        validators=(),
        widget=None,
        **kwargs
    ):

        self._target_model = target_model
        self._queryset = queryset

        self._to_field_name = to_field_name
        self._limit_choices_to = limit_choices_to
        self._required = required
        self._help_text = help_text
        self._validators = validators
        self._widget = widget

        # keep a copy of all kwargs (including our normalised choices list) for deconstruct()
        self._constructor_kwargs = kwargs.copy()
        if target_model is not None:
            self._constructor_kwargs["target_model"] = target_model
        if queryset is not None:
            self._constructor_kwargs["queryset"] = queryset
        if to_field_name is not None:
            self._constructor_kwargs["to_field_name"] = to_field_name
        if limit_choices_to is not None:
            self._constructor_kwargs["limit_choices_to"] = limit_choices_to
        if required is not True:
            self._constructor_kwargs["required"] = required
        if help_text is not None:
            self._constructor_kwargs["help_text"] = help_text

        super().__init__(**kwargs)

    @cached_property
    def queryset(self) -> QuerySet:
        queryset = None
        if self._target_model is not None:
            queryset = resolve_model_string(self._target_model).objects.all()

        if self._queryset is not None:
            queryset = self._queryset
            if callable(self._queryset):
                queryset = self._queryset()

        if queryset is None:
            raise ValueError("target_model or queryset should be provided.")

        return queryset

    @cached_property
    def field(self) -> "forms.ModelChoiceField":
        return forms.ModelChoiceField(
            queryset=self.queryset,
            limit_choices_to=self._limit_choices_to,
            to_field_name=self._to_field_name,
            widget=self._widget,
            required=self._required,
            validators=self._validators,
            help_text=self._help_text,
        )

    def clean(self, value):
        if isinstance(value, self.queryset.model):
            key = self._to_field_name or "pk"
            value = getattr(value, key)
        return super().clean(value)

    def to_python(self, value) -> Optional[Model]:
        # the incoming serialised value should be None or a specific field value
        if value is None:
            return value
        else:
            try:
                key = self._to_field_name or "pk"
                return self.queryset.model.objects.get(**{key: value})
            except self.queryset.model.DoesNotExist:
                return None

    def bulk_to_python(self, values) -> List[Model]:
        """Return the model instances for the given list of primary keys.

        The instances must be returned in the same order as the values and keep None values.
        """
        key = self._to_field_name or "pk"
        objects = self.queryset.model.objects.in_bulk(values, field_name=key)
        return [
            objects.get(key) for key in values
        ]  # Keeps the ordering the same as in values.

    def get_prep_value(self, value):
        # the native value (a model instance or None) should serialise to a specific field value or None
        if value is None:
            return None
        else:
            key = self._to_field_name or "pk"
            return getattr(value, key)

    def value_from_form(self, value):
        # ModelChoiceField sometimes returns scalar value, and sometimes an instance; we want the instance
        if value is None or isinstance(value, self.queryset.model):
            return value
        else:
            try:
                key = self._to_field_name or "pk"
                return self.queryset.model.objects.get(**{key: value})
            except self.queryset.model.DoesNotExist:
                return None

    def deconstruct(self):
        return (
            "wagtail.core.blocks.fields.ModelChoiceBlock",
            [],
            self._constructor_kwargs,
        )


class ModelMultipleChoiceBlock(ModelChoiceBlock):
    class Meta:
        icon = "placeholder"

    @cached_property
    def field(self) -> "forms.ModelMultipleChoiceField":
        return forms.ModelMultipleChoiceField(
            queryset=self.queryset,
            limit_choices_to=self._limit_choices_to,
            to_field_name=self._to_field_name,
            widget=self._widget,
            required=self._required,
            validators=self._validators,
            help_text=self._help_text,
        )

    def clean(self, value):
        if isinstance(value, QuerySet):
            key = self._to_field_name or "pk"
            value = [getattr(obj, key) for obj in value]
        return super().clean(value)

    def get_prep_value(self, value):
        # the native value (a queryset or None) should serialise to a list of specific field values or None
        if value is None:
            return None
        else:
            key = self._to_field_name or "pk"
            return [getattr(obj, key) for obj in value]

    def to_python(self, value):
        # the incoming serialised value should be None or a list of specific field values
        if value is None:
            return value
        else:
            try:
                key = self._to_field_name or "pk"
                return self.queryset.model.objects.filter(**{"%s__in" % key: value})
            except self.queryset.model.DoesNotExist:
                return None

    def value_from_form(self, value):
        # ModelChoiceField sometimes returns list of scalar values, and sometimes a queryset; we want the queryset
        if value is None or isinstance(value, QuerySet):
            return value
        else:
            try:
                key = self._to_field_name or "pk"
                return self.queryset.model.objects.filter(**{"%s__in" % key: value})
            except self.queryset.model.DoesNotExist:
                return None

    def deconstruct(self):
        return (
            "wagtail.core.blocks.fields.ModelMultipleChoiceBlock",
            [],
            self._constructor_kwargs,
        )

```

gasman:
Feedback implemented in #5819 and merged in 4314f3d1a1419f7a184c858f8cc8afa1dbfb3d5c.

SmileyJames:
@gasman Sorry for my silence on this pull request - I've been busy. Thanks for getting this merged and making the changes required, much appreciated.

gasman:
@SmileyJames No worries - the lack of activity was entirely at our end :-)

PR #5597, opened 2019-10-02
Add a note in editor manual about potential disabled moderation

jacobtoppm:
This notes that the Submit for Moderation button may be disabled by the site administrator, and links to the relevant setting.


gasman:
Looks good to me, thanks @jacobtoppm!

PR #5599, opened 2019-10-03
Fix summary item counts

### LINKED ISSUE #5129: Image and document summary item don't take permissions into account for counting number of images/documents
### Summary

The admin homepage summary items for documents and images don't reflect on the permissions for collections. If a user has permission for any collection the numbers in the summary will show the total number of images/documents in the database. This leaks information to users that might not have permission to more than a single collection

### Steps to Reproduce

1. In any wagtail project with at least one collection more than the root collection create a user that does'nt have permission to the root collection but to at least one other collection
2. Ensure there are images/documents in the root collection
3. Sign in as the new user, when reaching the admin page the numbers telling the amount of images/documents in the system will render the total amount in the database, not what's available to the user

### Technical details

* Python version: 3.7.2
* Django version: 2.1.7
* Wagtail version: 2-4
* Browser version: Firefox 65.0

### Proposed solution
Update "get_context" method in "ImagesSummaryItem" in wagtail/images/wagtail_hooks.py according to:

```python
def get_context(self):

    # Fetch all collection user has any permission for to filter image counting
    collections = permission_policy.collections_user_has_any_permission_for(self.request.user, ['add', 'change', 'delete'])
    return {
        'total_images': get_image_model().objects.filter(collection__in=collections).count(),
    }
```

And similar for wagtail/documents/wagtail_hooks.py

georgeyk:
Fixes #5129 

Consider user permissions in ImagesSummaryItem and DocumentsSummaryItem counter.

I'm not very familiar with the codebase, so before I replicate the suggested fix in DocumentsSummaryItem, I would like to validate if this is going in the right direction.

~ :v:

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -43,6 +43,7 @@ Changelog
  * Fix: Prevent broken images in notification emails when static files are hosted on a remote domain (Eduard Luca)
  * Fix: Replace styleguide example avatar with default image to avoid issues when custom user model is used (Matt Westcott)
  * Fix: `DraftailRichTextArea` is no longer treated as a hidden field by Django's form logic (Sergey Fedoseev)
+ * Fix: `ImagesSummaryItem` image counter to consider user permissions on existing collections (George Y. Kussumoto)
 
 
 2.6.2 (19.09.2019)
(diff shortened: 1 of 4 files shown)


gasman:
Closing as per discussion on #5129. Thanks for your work on this, @georgeyk!

PR #5600, opened 2019-10-03
Use specific Page model in edit view

haplo:
This allows for overriding `Page` methods such as `permissions_for_user`. Same approach was already done in `create` view, but not in `edit` view, which was keeping me from implementing a custom permission system by overriding `UserPagePermissionsProxy` and `PagePermissionTester` classes.

All tests pass and this change is working on production at www.outdoorsy.com.

### DIFF
--- a/wagtail/admin/views/pages.py
+++ b/wagtail/admin/views/pages.py
@@ -321,7 +321,7 @@ def create(request, content_type_app_name, content_type_model_name, parent_page_
 
 
 def edit(request, page_id):
-    real_page_record = get_object_or_404(Page, id=page_id)
+    real_page_record = get_object_or_404(Page, id=page_id).specific
     latest_revision = real_page_record.get_latest_revision()
     page = real_page_record.get_latest_revision_as_page()
     parent = page.get_parent()


haplo:
There are more places that hardcode `UserPagePermissionsProxy`, for full customizability they should be refactored to allow for subclassing, or by introducing a new settings value. Please let me know if this would be of interest.

Also I could write some documentation on how to implement a custom permission system, although until the issue raised in the previous paragraph is addressed there will be some features that cannot be overridden.

gasman:
Hi @haplo - please see the discussion on #2907. Custom permission systems are a difficult subject - I can see that there's demand for them, but it effectively means handing over a large chunk of Wagtail's internal logic to user code, and I'm not sure that's a safe thing to do without formally defining the contract between Wagtail code and user code (for example: do we require Wagtail to _always_ go through the `Page.permissions_for_user` method when checking permissions? From my point of view that would be a bad thing, as it prevents listing views from using a single UserPagePermissionsProxy on all pages, resulting in an N+1 query scenario).

On a separate note - it looks like this change makes the logic at https://github.com/wagtail/wagtail/blob/f4cc454e13c23161490b995da970cd2f33ec92fa/wagtail/admin/views/pages.py#L530-L534 partly redundant. We ought to check whether there's a reason we avoided setting `real_page_record` to the specific page up-front on that occasion...

gasman:
It'll have to be a no this time, I'm afraid... Every change we make to the codebase, even one as small as this, has a cost (in terms of maintainability and potential for breaking things) that we have to weigh up against the benefit - and here the only benefit is to allow one particular piece of outside code to interact with an internal variable in unspecified and unknowable ways (that are liable to break as and when this code is refactored in future). As such, it's impossible for us to judge whether this is an appropriate change to make.

Issue #5358, opened 2019-06-06
Version number label obscures settings menu item


gasman:
On complex sites, the settings menu can exceed the window height, causing it to receive a scrollbar. When this happens, the version number label scrolls along with the menu items, causing it to permanently obscure one of those items:

![settings_menu_obscuring](https://user-images.githubusercontent.com/85097/59020640-52ce9480-8842-11e9-8264-56bcf63e5821.gif)

It would be better for the version number label to stay fixed on scroll, and for the menu to be given enough bottom padding to ensure that the last item can be scrolled above that label.

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: no

### Technical details

* Python version: 3.7
* Django version: 2.1.9
* Wagtail version: master
* Browser version: Chrome/74.0.3729.169 https://whatsmybrowser.org/b/ZM555Y9

lb-:
Fixed in a459e91692659aba04e662978857d14061aecaee via #5606

abeyerpath:
The PR that fixed this was reverted, so this issue should probably be reopened, no?

gasman:
@abeyerpath Good point, yep!

PR #5606, opened 2019-10-07
Fix version number label obscuring menu items in Settings submenu

Igoranze:
If you have a lot of items in the Settings submenu, then the footer breaks some of the items.
You can see in the gif below that the footer should be fixed at the bottom of the screen, but instead the footer is moving the more items are added.

bug

Fixed this by adding the `position: sticky;` and `max-height: unset;` to the new class `footer-submenu`. This way the footer is always shown after the last item on the submenu. But now when the height of the page is smaller than the items, the footer is sticky to the bottom of the browser, and none of the items are broken by the footer.

fixed

Tested with:
Safari: Version 13.0.1 (14608.2.11.1.11)
Firefox: 69.0.2 (64-bit)
Opera: Version:63.0.3368.107
Chrome: Version 77.0.3865.90 (Official Build) (64-bit)

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -4,6 +4,7 @@ Changelog
 2.7 LTS (xx.xx.xxxx) - IN DEVELOPMENT
 ~~~~~~~~~~~~~~~~~~~~
 
+ * Fix: Submenu items longer then the page height are no longer broken by the submenu footer (Igor van Spengen)
  * Improved StreamField design (Bertrand Bordage)
  * Added `construct_page_listing_buttons` hook (Michael van Tellingen)
  * Added more detailed documentation and troubleshooting for installing OpenCV for feature detection (Daniele Procida)
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -411,6 +411,7 @@ Contributors
 * Tim White
 * Mike Janger
 * Prithvi MK
+* Igor van Spengen
 
 Translators
 ===========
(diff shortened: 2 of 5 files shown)


gasman:
Thanks @Igoranze! This will fix #5358.

Igoranze:
Cool @gasman, thanks for the quick response. I was not aware of the existing issue.

rinti:
I'm not sure what the lowest version of browsers Wagtail targets, but `position: sticky` [is not available in IE11](https://caniuse.com/#feat=css-sticky).
I don't really see the value in it being sticky though, in my opinion it could just be the last item of the list - no need to attach to bottom.

gasman:
@rinti Wagtail's browser support is documented at http://docs.wagtail.io/en/stable/contributing/developing.html#browser-and-device-support - there are already a few limitations in IE11 support noted there.

lb-:
Awesome work @Igoranze  - I made one minor tweak to get this working well in IE 11 (position relative on the outer scrolling container) and validated on:

IE 11 (windows)
Edge (windows)
Firefox (mac)
Chrome (mac)


This has been merged in a459e91692659aba04e662978857d14061aecaee

gasman:
Unfortunately this change causes the footer to no longer be bottom-aligned when the menu is shorter than window height:

Screenshot 2020-01-17 at 12 51 39

While this is only a cosmetic issue, I'd rather avoid any regressions in the out-of-the-box setup given that the bug being fixed here is only really an edge case. Reverted for now in 590bea3b7b47f0bdf90a1557057bd37e04bfd0e8

Stormheg:
This was fixed by https://github.com/wagtail/wagtail/pull/5927 and released in the 2.9 release.
We can close this now :)

PR #5607, opened 2019-10-08
Page description

KalobTaulien:
* Do the tests still pass? ✅
* Have you added tests to cover the new/fixed behaviour? ✅
* Has the documentation been updated accordingly? ✅

Added a "help text"-style optional attribute to the Page model that will display the Page's helpful text in three different locations in the Wagtail admin. Previews below.

**Note:** I did _not_ add any frontend styling. I'd be happy for some constructive feedback on the design implementation and would be happy to open another PR for better design/frontend implementation. 

Screen Shot 2019-10-07 at 7 19 07 PM

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/pages/add_subpage.html
+++ b/wagtail/admin/templates/wagtailadmin/pages/add_subpage.html
@@ -20,7 +20,7 @@
                                 <a href="{% url 'wagtailadmin_pages:add' app_label model_name parent_page.id %}{% if next %}?next={{ next }}{% endif %}" class="icon icon-plus-inverse icon-larger">
                                     {{ verbose_name }}
                                     {% if page_description %}
-                                        <div><small>{{ page_description }}</small></div>
+                                        <div><span class="visuallyhidden">.</span><small>{{ page_description }}</small></div>
                                     {% endif %}
                                 </a>
                             </div>
--- a/wagtail/admin/templates/wagtailadmin/pages/add_subpage.html
+++ b/wagtail/admin/templates/wagtailadmin/pages/add_subpage.html
@@ -18,6 +18,9 @@
                         <div class="row row-flush">
                             <div class="col6">
                                 <a href="{% url 'wagtailadmin_pages:add' app_label model_name parent_page.id %}{% if next %}?next={{ next }}{% endif %}" class="icon icon-plus-inverse icon-larger">{{ verbose_name }}</a>
+                                {% if page_description %}
+                                    <small>{{ page_description }}</small>
+                                {% endif %}
                             </div>
 
                             <small class="col6" style="text-align:right">
--- a/wagtail/admin/templates/wagtailadmin/pages/add_subpage.html
+++ b/wagtail/admin/templates/wagtailadmin/pages/add_subpage.html
@@ -13,7 +13,7 @@
 
         {% if page_types %}
             <ul class="listing">
-                {% for verbose_name, app_label, model_name in page_types %}
+                {% for verbose_name, app_label, model_name, page_description in page_types %}
                     <li>
                         <div class="row row-flush">
                             <div class="col6">
--- a/wagtail/admin/templates/wagtailadmin/pages/add_subpage.html
+++ b/wagtail/admin/templates/wagtailadmin/pages/add_subpage.html
@@ -16,14 +16,16 @@
                 {% for verbose_name, app_label, model_name, page_description in page_types %}
                     <li>
                         <div class="row row-flush">
-                            <div class="col6">
-                                <a href="{% url 'wagtailadmin_pages:add' app_label model_name parent_page.id %}{% if next %}?next={{ next }}{% endif %}" class="icon icon-plus-inverse icon-larger">{{ verbose_name }}</a>
-                                {% if page_description %}
-                                    <small>{{ page_description }}</small>
-                                {% endif %}
+                            <div class="col8">
+                                <a href="{% url 'wagtailadmin_pages:add' app_label model_name parent_page.id %}{% if next %}?next={{ next }}{% endif %}" class="icon icon-plus-inverse icon-larger">
+                                    {{ verbose_name }}
+                                    {% if page_description %}
+                                        <div><small>{{ page_description }}</small></div>
+                                    {% endif %}
+                                </a>
                             </div>
 
-                            <small class="col6" style="text-align:right">
+                            <small class="col4" style="text-align:right">
                                 <a href="{% url 'wagtailadmin_pages:type_use' app_label model_name %}" class="nolink">{% blocktrans with page_type=verbose_name %}Pages using {{ page_type }}{% endblocktrans %}</a>
                             </small>
 
--- a/wagtail/admin/templates/wagtailadmin/pages/create.html
+++ b/wagtail/admin/templates/wagtailadmin/pages/create.html
@@ -16,7 +16,7 @@
                     {% trans 'New' %} <span>{{ content_type.model_class.get_verbose_name }}</span>
                 </h1>
                 {% if content_type.model_class.get_page_description %}
-                    <h3>{{ content_type.model_class.get_page_description }}</h3>
+                    <p>{{ content_type.model_class.get_page_description }}</p>
                 {% endif %}
             </div>
         </div>
--- a/wagtail/admin/templates/wagtailadmin/pages/create.html
+++ b/wagtail/admin/templates/wagtailadmin/pages/create.html
@@ -6,7 +6,6 @@
 {% block bodyclass %}page-editor create model-{{ content_type.model }}{% endblock %}
 
 {% block content %}
-
     <header class="merged tab-merged">
         {% explorer_breadcrumb parent_page include_self=1 %}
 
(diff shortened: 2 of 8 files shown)


tomdyson:
Thanks @KalobTaulien! 

Should we consider using [Meta options](https://docs.djangoproject.com/en/3.0/topics/db/models/#meta-options) for optional non-content attributes like this one?

> Model metadata is “anything that’s not a field”, such as ordering options (ordering), database table name (db_table), or human-readable singular and plural names (verbose_name and verbose_name_plural). None are required, and adding class Meta to a model is completely optional.

gasman:
I think we tried using Meta for this sort of thing in the early days of Wagtail, and found that Django made it hard to add extra properties beyond the built-in set. Maybe that's been improved now though...

wagtail/admin/templates/wagtailadmin/pages/create.html
```diff
@@ -13,7 +13,11 @@
         <div class="row row-flush">
             <div class="left col9 header-title">
                 <h1 class="icon icon-doc-empty-inverse">
-                {% trans 'New' %} <span>{{ content_type.model_class.get_verbose_name }}</span></h1>
+                    {% trans 'New' %} <span>{{ content_type.model_class.get_verbose_name }}</span>
```
jacobtoppm, line 16: My concern here is for screen readers, though I'm not sure what the solution is. I tested this briefly in Orca in Firefox, and found that the small tag did not cause a break - there's no separation between the heading and subheading. While this isn't a huge issue, it could lead to long and confusing headings for screen reader users. Perhaps this would be better outside the h1 to separate it, or by using something else to denote the heading/subheading divide?

wagtail/admin/templates/wagtailadmin/pages/edit.html
```diff
@@ -13,7 +13,11 @@
         <div class="row row-flush">
             <div class="left col9 header-title">
                 <h1 class="icon icon-doc-empty-inverse">
-                {% blocktrans with title=page.get_admin_display_title page_type=content_type.model_class.get_verbose_name %}Editing {{ page_type }} <span>{{ title }}</span>{% endblocktrans %}</h1>
+                    {% blocktrans with title=page.get_admin_display_title page_type=content_type.model_class.get_verbose_name %}Editing {{ page_type }} <span>{{ title }}</span>{% endblocktrans %}
```
jacobtoppm, line 16: This combination of page type + page name + help text could get pretty long, and doesn't look great when it wraps as far as I can tell when testing quickly - is there any way we could look at the wrapping here, or put the help text below instead?

wagtail/admin/templates/wagtailadmin/pages/create.html
zerolab: Semantically, the description does not belong in the heading, so 👍for moving it outside.
This may also move the description under "New/Editing <Page type>" making it cleaner.

wagtail/admin/templates/wagtailadmin/pages/edit.html
KalobTaulien: Hey @jacobtoppm I'm going to remove this until the new design work is in core as to prevent design conflicts.

KalobTaulien:
I rebased this off master today. 

I've removed the edit view changes since there's lots of great design work coming down the pipeline soon. 

I only added the one assert in the existing add_supage test, not sure if it requires more or not. Happy to write more if that's what we think is best. 

And I've pinged @thibaudcolas for his thoughts on accessibility in the /add_subpage/ template for the last comment. 

@gasman should I look into moving this into the Meta class as Tom had mentioned? 

Thanks for the feedback everybody!

wagtail/core/models.py
```diff
@@ -1073,6 +1073,13 @@ def get_verbose_name(cls):
         # except this doesn't convert any characters to lowercase
         return capfirst(cls._meta.verbose_name)
 
+    @classmethod
+    def get_page_description(cls):
+        """
+        Returns a page description if it's not set. eg "A multi-purpose web page".
```
thibaudcolas, line 1079: Typo?

```suggestion
        Returns a page description if it's set. eg "A multi-purpose web page".
```

wagtail/admin/templates/wagtailadmin/pages/create.html
thibaudcolas: I would suggest moving it out because I presume the text could get relatively lengthy for real-world descriptions.

If you do want to keep it in, you can add a screen-reader-only full stop `<span class="visuallyhidden">.</span>` in-between the title and description so there is a pause when announcing the text. In that case it would be nice for the `<title>` and `<h1>` to match.

wagtail/core/models.py
thibaudcolas: Bonus point for documentation – most content guidelines focused on readability suggest avoiding "eg" or other latin abbreviation and going for more plain language like "For example" instead.

wagtail/admin/templates/wagtailadmin/pages/create.html
```diff
@@ -6,13 +6,18 @@
...
 
         <div class="row">
             <div class="left col9 header-title">
-                <h1>{% trans 'New' %} <span>{{ content_type.model_class.get_verbose_name }}</span></h1>
+                <h1>
+                    {% icon "doc-empty-inverse" class_name="header-title-icon" %}
+                    {% trans 'New' %} <span>{{ content_type.model_class.get_verbose_name }}</span>
+                </h1>
+                {% if content_type.model_class.get_page_description %}
+                    <h3>{{ content_type.model_class.get_page_description }}</h3>
```
thibaudcolas, line 19: This would be better as a `p` or un-semantic `div`. There shouldn’t be two headings one after another like this.

wagtail/admin/templates/wagtailadmin/pages/add_subpage.html
```diff
@@ -13,14 +13,19 @@
...
+                {% for verbose_name, app_label, model_name, page_description in page_types %}
                     <li>
                         <div class="row row-flush">
-                            <div class="col6">
-                                <a href="{% url 'wagtailadmin_pages:add' app_label model_name parent_page.id %}{% if next %}?next={{ next }}{% endif %}" class="icon icon-plus-inverse icon-larger">{{ verbose_name }}</a>
+                            <div class="col8">
+                                <a href="{% url 'wagtailadmin_pages:add' app_label model_name parent_page.id %}{% if next %}?next={{ next }}{% endif %}" class="icon icon-plus-inverse icon-larger">
+                                    {{ verbose_name }}
+                                    {% if page_description %}
+                                        <div><small>{{ page_description }}</small></div>
```
thibaudcolas, line 23: Sorry I didn’t catch this sooner – could you add a:

```suggestion
                                        <div><span class="visuallyhidden">.</span><small>{{ page_description }}</small></div>
```

This will have screen readers have a pause between announcing the name and the description.

gasman:
> @gasman should I look into moving this into the Meta class as Tom had mentioned?

Sure - definitely worth looking, although I seem to remember that last time I checked, Django didn't let you add arbitrary new options to Meta, so don't worry too much if it can't be done.

kmagee123:
Hi @KalobTaulien here is the design that was requested for this feature. Tom Dyson has had a quick look at this yesterday and gave it the go-ahead. Let me know if you have any question. 

[]
Artboard (1)
(url)

kmagee123:
Hi @KalobTaulien here is the design that was requested for this feature. Tom Dyson has had a quick look at this yesterday and gave it the go-ahead. Let me know if you have any question.

Wagtail design

allcaps:
Only `DEFAULT_NAMES` are allowed on Meta.
https://github.com/django/django/blob/master/django/db/models/options.py#L198

This:
```
class FooPage(Page):
    class Meta:
        description = "foo"
```
Gives an error on startup:
```
TypeError: 'class Meta' got invalid attribute(s): description
```

We can extend `DEFAULT_NAMES`, but I prefer a method. 

A method allows easier interaction with the rest of the application which can be useful. Eg: Display the page usage count in the description, or get some other dynamic data.

wagtail/admin/templates/wagtailadmin/pages/create.html
```diff
@@ -6,13 +6,18 @@
...
 
         <div class="row">
             <div class="left col9 header-title">
-                <h1>{% trans 'New' %} <span>{{ content_type.model_class.get_verbose_name }}</span></h1>
+                <h1>
+                    {% icon "doc-empty-inverse" class_name="header-title-icon" %}
+                    {% trans 'New' %} <span>{{ content_type.model_class.get_verbose_name }}</span>
+                </h1>
+                {% if content_type.model_class.get_page_description %}
+                    <p>{{ content_type.model_class.get_page_description }}</p>
```
allcaps, line 19: Can we change this `p` to `div`? One might return a safe string from `get_page_description` with html structure. This allows you to display thumbnail/previews etc.

allcaps (approved):
LGTM.

Small thing, the `get_page_description` might also return a HTML structure. So I prefer wrapping the content in a div.

thibaudcolas:
This seems ready to go – I’ll review the conflicts, try to resolve them, and merge. We are going to be redesigning this page in the near future, so I think the design by Kyle can already be considered outdated unfortunately.

zerolab:
Superseded by #8065 which fixes merge conflicts

Issue #3623, opened 2017-06-01
Wagtail codebase fails import linting with isort 4.2.8


gasman:
In its default configuration, isort 4.2.8 (released yesterday) is pickier about whitespace than the previous version (4.2.5), and as a result there are now a bunch of validation failures. We need to either: 1) fix all those failures, 2) tweak the `make lint` configuration to disable the stricter checks, or 3) pin isort to the older version. In the case of 2) or 3), we should also update http://docs.wagtail.io/en/v1.10.1/contributing/python_guidelines.html#pep8 to say what our policy is.

thibaudcolas:
I had the same problem. I don't like the idea of CI builds breaking because of "patch" releases from this, so decided to pin.

gasman:
I guess pinning the version is a good idea in any case, since this clearly shows that isort isn't following semantic versioning.

gasman:
Fixed in b976db3afc456695bce57912ad8ef6aec90de5dd by pinning to 4.2.5.

PR #5608, opened 2019-10-08
isort config file & isort import ordering

KalobTaulien:
When running `make lint` I had noticed a bunch of isort ordering warnings. This commit addresses those warnings by running `isort` on those files. 

I've also added an `.isort.cfg` file (and purposely left out the `line_length` setting). The `multi_line_output` has a [number of options](https://github.com/timothycrosley/isort#multi-line-output-modes) we can run, I just chose a fairly generic option (#5). 

After running `make lint` I no longer get any isort warnings.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,6 @@
+[settings]
+indent='    '
+multi_line_output=5
+known_django=django
+sections=FUTURE,STDLIB,DJANGO,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
+skip=migrations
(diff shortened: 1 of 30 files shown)


gasman:
Hi @KalobTaulien,
Please see #3623 - it sounds like you're running a later version of isort than 4.2.5. I'm not against updating to a newer version, but I think we still need to pin against an exact release (and reflect that in the docs / setup.py), as we can't be confident that isort won't break existing configs again in future.

loicteixeira:
Also, there's already a `isort` section in `setup.cfg` so `.isort.cfg` shouldn't be needed.

KalobTaulien:
Something was up with my system because it wasn't honouring the `setup.cfg` isort config. And I was using isort 4.2.5 but it was complaining _heavily_ — I cloned and spun up a new instance of Wagtail and isort and flake8 had no complaints. Sooo ¯\_(ツ)_/¯ I have no idea. I'll close this PR though.

PR #5612, opened 2019-10-09
Support overriding default item in page action menu

### LINKED ISSUE #5438: Save Draft page action is not overridable

gasman:
Fixes #5438; rebase and update of #5500.

The behaviour has now been reversed so that the first item in the list is selected as the default, rather than the last; this way developers will have to either set a negative `order` value or explicitly reorder items in `construct_page_action_menu` to replace the Save Draft item, which makes it unlikely to happen by accident on existing uses of this hook.

Have also fixed behaviour for locked pages, updated docs/tests to avoid rewriting the whole list, and added a release note to cover the minor template change.

### DIFF
--- a/docs/reference/hooks.rst
+++ b/docs/reference/hooks.rst
@@ -524,7 +524,7 @@ Hooks for customising the way users are directed through the process of creating
 
   Add an item to the popup menu of actions on the page creation and edit views. The callable passed to this hook must return an instance of ``wagtail.admin.action_menu.ActionMenuItem``. The following attributes and methods are available to be overridden on subclasses of ``ActionMenuItem``:
 
-  :order: an integer (default 100) which determines the item's position in the menu. Can also be passed as a keyword argument to the object constructor
+  :order: an integer (default 100) which determines the item's position in the menu. Can also be passed as a keyword argument to the object constructor. The lowest-numbered item in this sequence will be selected as the default menu item; as standard, this is "Save draft" (which has an ``order`` of 0).
   :label: the displayed text of the menu item
   :get_url: a method which returns a URL for the menu item to link to; by default, returns ``None`` which causes the menu item to behave as a form submit button instead
   :name: value of the ``name`` attribute of the submit button, if no URL is specified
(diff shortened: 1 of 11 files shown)


docs/reference/hooks.rst
```diff
@@ -566,12 +566,28 @@ Hooks for customising the way users are directed through the process of creating
...
+
+  .. code-block:: python
+
+    @hooks.register('construct_page_action_menu')
+    def make_publish_default_action(menu_items, request, context):
+        for (index, item) in enumerate(menu_items):
+            if item.name == 'action-publish':
+                # move to top of list
+                menu_items.pop(index)
+                menu_items.insert(0, item)
```
kaedroho: This probably needs a note to say that this will ignore the `order` field on the menu items?

gasman: Sounds good. I'll update the text to:

The `construct_page_action_menu` hook is called after the menu items have been sorted by their `order` attributes, and so setting a menu item's `order` will have no effect at this point. Instead, items can be reordered by changing their position in the list, with the first item being selected as the default action. For example, to change the default action to Publish:

gasman:
Merged in ee68d2214af95601967ce9cfd1c423bd2606a81e + parents. Thanks again @rjpruitt16!

PR #5613, opened 2019-10-11
FixedTypo-ReadMe

kailwallin:
verb did not agree with the subject

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ We also welcome translations for Wagtail's interface. Translation work should be
 
 ### Thanks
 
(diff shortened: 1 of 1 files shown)


lb-:
@kailwallin  how would you like to be referred to in the contributors list?

gasman:
Maybe it's a peculiarly British thing to refer to companies in the plural, but "Browserstack, who provides..." sounds totally wrong to me - I'd only use that phrasing when talking about an individual person. "Browserstack, which provides..." might work, but still feels a bit odd.

lb-:
@gasman  - just merged in, I felt it read ok with the change but happy to revise if needed.

https://github.com/wagtail/wagtail/commit/6a6555eb191b32e64de656605f9f5521dff07083

Thanks for the PR @kailwallin - let me know if you want the contributors name updated :)

PR #5616, opened 2019-10-13
Refined creation of title on image & document uploads

### LINKED ISSUE #4945: Feature Request:  Refine the creation of the default file title when uploading an image
As follows:

- Remove filename extension (including the period) from default image title
- Replace underscores and dashes with spaces.  These characters are many times found in image filenames but would not typically be in a title.
- Title case the title

lb-:
Resolves #4945 

## Overview of implementation
* By default, the file name will be used to generate a title which will be the file without the file extension
* If a value is already in the applicable title field, by default this value will be kept as is
* Standardises the implementation across multiple upload forms (was previously inconsistent)
* Moves the implementation to 100% client side so user can 'edit' value before it is saved
* Adds the ability for this functionality to overridden (includes docs on how to do this)
* Contains some minor JS code fix ups

## Forms updated
- view - admin/images/multiple/add
- view - admin/images/add
- view - admin/documents/add
- view - admin/documents/multiple/add
- modal - image-chooser-modal
- modal - document-chooser-modal

## How to validate
* For each of the forms above, attempt to upload a document or image and check the title field gets pre-filled with the filename but not the extension
* For the non-multiple forms, before clicking save, select a different file to upload and confirm the title is left unchanged

## PR Checklist
* Do the tests still pass? the core.js is actually untested, I am happy to get some tests started but it appears this will be quite some effort.
* Does the code comply with the style guide?  ✅  - js lint passes
* For front-end changes: Tested on the following
  * Safari Version 12.1.1 MacOS 10.14.5
  * Firefox 69 MacOS 10.14.5
  * Chrome 77 MacOS 10.14.5

### DIFF
--- a/docs/reference/hooks.rst
+++ b/docs/reference/hooks.rst
@@ -387,12 +387,13 @@ Hooks for customising the editing interface for pages and snippets.
         js_includes = format_html_join('\n', '<script src="{0}"></script>',
             ((static(filename),) for filename in js_files)
         )
+        # remember to use double '{{' so they are not parsed as template placeholders
         return js_includes + format_html(
             """
             <script>
-                $(function() {
+                $(function() {{
                     $('button').raptorize();
-                });
+                }});
             </script>
             """
         )
--- a/wagtail/admin/static_src/wagtailadmin/js/core.js
+++ b/wagtail/admin/static_src/wagtailadmin/js/core.js
@@ -51,7 +51,6 @@ function initTagField(id, autocompleteUrl, options) {
  *    - alwaysDirty - When set to true the form will always be considered dirty,
  *      prompting the user even when nothing has been changed.
 */
-
 function enableDirtyFormCheck(formSelector, options) {
     var $form = $(formSelector);
     var confirmationMessage = options.confirmationMessage || ' ';
@@ -353,7 +352,9 @@ wagtail = (function(document, window, wagtail) {
     // Module pattern
     if (!wagtail) {
         wagtail = {
-            ui: {}
+            components: {}, // see wagtailadmin.entry.js
+            ui: {},
+            utils: {},
         };
     }
 
@@ -536,6 +537,7 @@ wagtail = (function(document, window, wagtail) {
     }
 
     $(document).ready(initDropDowns);
+
     wagtail.ui.initDropDowns = initDropDowns;
     wagtail.ui.DropDownController = DropDownController;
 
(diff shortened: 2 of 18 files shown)


lb-:
Working on the python tests now :)

DanielSwain:
The upcoming 2.7 release contains a number of great new features.  It would be great if this small but very useful feature could be included in 2.7 too.

lb-:
@DanielSwain - we have frozen the features for 2.7 (as soon as we put out an RC) so this one wont make it in but hopefully we can get this into a future release.

thibaudcolas:
Tagging this as `Accessibility`. Currently the title of an image often ends up as the image’s `alt` text. This is bad, and having a better title isn’t the correct solution, but it’s a positive improvement nonetheless.

lb-:
Cool. I'll try to rebase and open as a new pr soon.

lb-:
This has now been rebased, anyone keen to take a look?
@kaedroho  / @thibaudcolas

docs/reference/hooks.rst
```diff
@@ -388,17 +388,56 @@ Hooks for customising the editing interface for pages and snippets.
...
+                $(function() {{
                     $('button').raptorize();
-                });
+                }});
             </script>
             """
         )
 
 
+  Override how the title is set when uploading a document or image.
```
gasman, line 403: I don't think this is the right place in the docs for this - it's not describing the functionality of the `insert_editor_js` hook, but rather one particular JS customisation that you might do through it. Perhaps it should be a new subsection in advanced topics -> images (with a pointer to it from advanced topics -> documents)?

Also, I'm not sure that `insert_editor_js` will work for this, because that only gets inserted on pages that include `wagtailadmin/pages/_editor_js.html` - the multiple image uploader (for example) is not one of those. (Ideally `_editor_js.html` would only be used by the image/snippet/modeladmin edit views, although in practice it ends up being included on a bunch of others through lazy copy-and-paste)

wagtail/admin/static_src/wagtailadmin/js/core.js
```diff
@@ -538,6 +539,69 @@ wagtail = (function(document, window, wagtail) {
...
+     * @returns {Function}
+     */
+    function getPopulateTitleHandler (model, widget) {
+        return function onChangeHandler (event, data) {
+
+            var $titleField;
+            var currentTitle;
+            var filename;
+
+            if (widget === 'ADD_MULTIPLE') {
```
gasman, line 575: The branch here makes me think that there's not really much code in common between the ADD_MULTIPLE handler and the others, and we'd probably be better off making this into two separate functions.

lb-:
@gasman thanks for the review - I'll have a read and get some changes pushed up.

wagtail/admin/static_src/wagtailadmin/js/core.js
```diff
@@ -538,6 +539,69 @@ wagtail = (function(document, window, wagtail) {
...
+     * @param {Object} options 
+     * @param {string=} options.currentTitle - current entered value (if available), will be undefined for `ADD_MULTIPLE`
+     * @param {Number=} options.maxLength - maximum title length (if available), will be undefined for `ADD_MULTIPLE`
+     * @param {string} options.model - `DOCUMENT` or `IMAGE`
+     * @param {string} options.widget - `CHOOSER_MODAL`, `ADD` or `ADD_MULTIPLE`
+     * @returns {string=} if a string is not returned the title field will not be updated
+     */
+    function getTitleFromFilename (filename, options) {
+        if (options.currentTitle) return; // do not overwrite an existing entered title value
+        return filename.replace(/\.[^/.]+$/, ''); // remove file extension
```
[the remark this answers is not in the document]
lb-: nice pickup thanks

wagtail/admin/static_src/wagtailadmin/js/core.js
lb-: @gasman  I gave this a ago and personally I feel like it ends up being harder to understand, the reason is that the `onChangeHandler` needs to access the `model` & `widget` info to add to the `getTitleFromFilename` call. Splitting this two two functions ends up needing two more higher order functions which I think could create even more confusion. 

I tried a different approach, splitting the 'data getting' functions (eg. just make two small functions that each return `$titleField` & `currentTitle`). But this seems to make it confusing in a different way, plus we need to then either bind the function or pass in the `this` somehow so that the field can be used in single upload scenarios.

I have tried to isolate the 'pure data' handing into `getTitleFromFilename` and this function is what is left to smooth over the differences between the two upload widgets' internal data representation.

Plus the 'clean up' code at the end is partially shared between the various scenarios and splitting this could mean some duplication of this code that can easily get missed in the future.

docs/reference/hooks.rst
lb-: Agreed, good call, working on this now.

lb-:
@gasman updated PR with most changes as requested;

* fixed up docs locations & added more info to the docs + re-tested example code
* rebased on master
* linting re-run + manual validation re-done
* added max title length for multiple upload scenarios

lb-:
a few failing tests - will investigate

wagtail/admin/static_src/wagtailadmin/js/core.js
gasman: The problem here is that we have a centralised function in wagtailadmin's core.js, tasked with knowing the internal workings of 2-3 different form implementations from wagtailimages/wagtaildocs, so that it can extract values to plug into `getTitleFromFilename`. Those mechanisms have pretty much nothing in common - one obtains the title field from `event`, another obtains it from `data`; one gives us the filename in a clean format, the other makes us extract it from a fake path. This doesn't seem like a helpful division of responsibilities between wagtailadmin and wagtailimages/wagtaildocs - code that needs to know about the internals of an upload form should live somewhere close to that form (such as in add-multiple.js).

A useful exercise when working with the wagtailimages / wagtaildocs apps is to ask "what would a third-party app do if it wanted to do the same thing"? ([wagtailmedia](https://github.com/torchbox/wagtailmedia) is the canonical example of such an app.) As a general rule, we don't want to give wagtailimages / wagtaildocs any special treatment within wagtailadmin. `getTitleFromFilename` is mostly fine in that respect - while technically the docstring specifies that `model` has to be 'DOCUMENT' or 'IMAGE', it probably won't do any harm for a third-party app to call it and identify itself as the 'ADD_MULTIPLE' widget for the 'VIDEO' model. On the other hand, that third-party app couldn't safely call `getPopulateTitleHandler` with the same parameters, as there's no guarantee that its own multiple-uploader form has the same internals as the image/document ones. (Similarly, what if a third-party app added a new kind of upload widget, such as 'import from Google Drive', and wanted to piggyback on the get-title-from-filename logic? In this case, it couldn't make use of `getPopulateTitleHandler`, as it has no knowledge of what that form looks like.)

FWIW, the duplication of code between wagtailimages and wagtaildocs does bother me too, and I'd definitely like to see the multiple uploaders refactored into a shared generic implementation within wagtailadmin. But A) that's very much out of scope for this PR, and B) if that does happen, I'd still want to stick to the rule of "JS code that knows about form internals should live somewhere close to that form".

lb-: Hey Matt. Thanks for explaining it - I see now what you mean about splitting the utils. Makes sense.

I'll give another crack at this when I get the chance and flag for a re-review then.

lb-:
Note to self - started splitting out the functions into duplicate versions for docs/images, WIP still
https://github.com/lb-/wagtail/tree/feature/4945-ability-to-customise-title-from-filename-attempt-three

jorgenbuilder:
Looking forward to seeing this get in! I'm using something very similar to this in my own fork. This has the added benefit of working better with the postgres search backend's limitations WRT partial matches.

thibaudcolas:
Hey @lb- 👋  do you mind sharing an update on your progress with refactorings following the previous review from Matt?

lb-:
@thibaudcolas - happy to finish this off, I got stuck trying to split the functions in a way that minimised duplication but as per Matt's comments I think I will just need to be ok with the code being copied for both the documents/images variants.
I will give this another go this weekend.

thibaudcolas:
@lb- thank you! Just let me know when you get to it, even if not this weekend, and I’ll re-review afterwards.

lb-:
Replaced with #7534

PR #5623, opened 2019-10-15
Formalise Python 3.8 support

gasman:


gasman:
Merged in ddbbaea288361efb6be537932d5f609e75e2f3c7

gasman:
Rolling back, as webp (#5249) doesn't seem to be supported on Python 3.8 by the current release of Pillow.

PR #5627, opened 2019-10-16
[5225] Take number of frames into account when limiting image size (rebase of #5227)

gasman:
Rebase #5227 with fixes applied and Willow version-bumped


gasman:
Merged in b4ecead6d91573f906007f39ca7aef4fb570ad0c

PR #5633, opened 2019-10-17
 Add trimmed to blocktrans blocks, remove redundant white from msgid

allcaps:
Wagtail templates contain many message ids that have redundant whitespace: Leading and trailing line breaks. This is caused by `{% blocktrans %}` blocks and gettext which _is_ whitespace significant.

Whitespace in msgid is annoying for translators since they have to match the whitespace in their msgstr.

Screenshot 2019-10-17 at 23 50 45

In this PR I updated all blocktrans tags that cause redundant whitespace to use the `trimmed` argument. `{% blocktrans %}` => `{% blocktrans trimmed %}`

This change requires translations to be updated as wel. All that redundant whitespce needs to be deleted form the message strings. Since we value the translators efforts, I propose to only update Transifex nl_NL translations to see the impact. I hope Transifex does something smart, I hope it kan keep the msgid and msgstr together. I that is the case, updating the strings will not be hard. If the link between msgid an msgstr is lost we might not want to merge this PR.

Why nl_NL? I'm a translator and nl_NL is 100% translated. If the impact is too big we can revert the change.

I attached a diff of the English source po file. To illustrate the change in msgid
[django.po.diff.zip](https://github.com/wagtail/wagtail/files/3741448/django.po.diff.zip)

I did not include that source po file in this PR since running makemessages is part of the release process.

@gasman Is it possible to test this PR against Transifex for only the nl_NL translations?

* For new features: Has the documentation been updated accordingly? N?A

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/chooser/_search_results.html
+++ b/wagtail/admin/templates/wagtailadmin/chooser/_search_results.html
@@ -1,7 +1,7 @@
 {% load i18n wagtailadmin_tags %}
 
 <h2>
-    {% blocktrans count counter=pages.paginator.count %}
+    {% blocktrans trimmed count counter=pages.paginator.count %}
         There is {{ counter }} match
     {% plural %}
         There are {{ counter }} matches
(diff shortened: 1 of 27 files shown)


allcaps:
Redoing this will be quicker than fixing this PR.
We also need a way to prevent new occurrences.

PR #5634, opened 2019-10-18
Author specific page locking

kaedroho:
This is an initial implementation of [RFC 43](https://github.com/wagtail/rfcs/pull/43).

This changes the behaviour of page locking so that the person who locked the page can still edit it. It also makes separates the "Unlock" permission from "Lock" so that editors can be given permission to lock pages but not unlock pages that are not locked by them.

Note that the RFC is still up for discussion so please put any comments about the overall design of this feature on the [RFC PR](https://github.com/wagtail/rfcs/pull/43).

This feature was sponsored by The Motley Fool


docs/advanced_topics/settings.rst
```diff
@@ -661,6 +661,17 @@ Customise the behaviour of rich text fields. By default, ``RichTextField`` and `
...
 
+Page locking
+------------
+
+.. versionadded:: 2.8
+
+  The page locking logic has been changed so that pages are still editable
+  by the users who locked them.
+
+The ``WAGTAILADMIN_GLOBAL_PAGE_EDIT_LOCK`` can be set to ``True`` to prevent users
```
gasman: I suggest removing "The".

wagtail/admin/menu.py
```diff
@@ -116,3 +115,4 @@ def get_context(self, request):
 
 admin_menu = Menu(register_hook_name='register_admin_menu_item', construct_hook_name='construct_main_menu')
 settings_menu = Menu(register_hook_name='register_settings_menu_item', construct_hook_name='construct_settings_menu')
+reports_menu = Menu(register_hook_name='register_reports_menu_item', construct_hook_name='construct_reports_menu')
```
gasman: We should document these hooks in docs/reference/hooks.rst.

wagtail/admin/views/reports.py
```diff
@@ -0,0 +1,30 @@
...
+    paginate_by = 10
+
+    def get_context_data(self, *args, object_list=None, **kwargs):
+        context = super().get_context_data(*args, object_list=object_list, **kwargs)
+        context['title'] = self.title
+        context['header_icon'] = self.header_icon
+        return context
+
+
+class LockedPagesView(ReportView):
```
gasman: Probably worth adding a permission check for the 'unlock any page' permission to this view (in addition to the one to be added to the menu item).

wagtail/admin/action_menu.py
```diff
@@ -66,7 +66,7 @@ def is_shown(self, request, context):
             return context['user_page_permissions'].for_page(context['parent_page']).can_publish_subpage()
         else:  # view == 'edit' or 'revisions_revert'
             return (
-                not context['page'].locked
+                not context['user_page_permissions'].for_page(context['page']).page_locked()
```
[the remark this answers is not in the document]
gasman: Marking as resolved, as this probably isn't something we're going to (or need to) solve in this PR.

wagtail/admin/views/reports.py
kaedroho: Users can unlock pages that they have locked themselves without requiring this permission, should we add that to the check too?

gasman:
Merged (along with fixes from https://github.com/kaedroho/wagtail/pull/54) in e927f6269a10e3b319c90ad4351c015014109437 + parents.

PR #5636, opened 2019-10-19
Make Wagtail menu visible on non-page views.

### LINKED ISSUE #5565: Easily make Wagtail Menu visible on non-Page views/templates
The Wagtail (frontend) menu (with the Wagtail icon) currently only renders on Page model views. It would be consistent, if this menu would appear on all front-end views, showing relevant links depending on the type of view. It is possible to create a custom template that would conditionally render the Wagtail Menu on all frontend templates, but [takes a considerable amount of boilerplate](https://stackoverflow.com/a/57679157/1191545).

Rather, it would be nice to have a default, or configuration option(s), whereby the Wagtail Menu would appear on all frontend templates where the template tag is included.

con-cat:
Fixes #5565 

I have checked that this renders correctly in the front end, the tests still pass, and the code complies with the style guide. I haven't updated the tests because it doesn't look like any of them currently look at whether the view is a page or which menu items are rendered. I'm happy to add new tests for that if needed.

### DIFF
--- a/wagtail/admin/templatetags/wagtailuserbar.py
+++ b/wagtail/admin/templatetags/wagtailuserbar.py
@@ -49,21 +49,20 @@ def wagtailuserbar(context, position='bottom-right'):
     if getattr(request, 'is_preview', False):
         return ''
 
-    # Only render if the context contains a variable referencing a saved page
-    page = get_page_instance(context)
-    if page is None:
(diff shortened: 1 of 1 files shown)


lb-:
Replaced by #6018 - thanks @con-cat for getting this started

Issue #2827, opened 2016-07-06
When updating a file on an image or a document, the 'Uploaded' timestamp does not update when a file is replaced.


mwharrison:
To duplicate:
1. Upload a file to the Wagtail CMS
2. After some time passes, replace said file
3. Notice timestamp for 'Uploaded' column references original file upload time

m1kola:
It's because `craeted_at` field is used to display "uploaded" time. `craeted_at` stores time when entry in database was created.

@mwharrison, thanks for reporting!

derhedwig:
It doesn't even have `updated_at`... Maybe that would be nice to have too!

lb-:
## Update

A simple resolution created on PR #5638 - as this column is the 'created at' field, I just renamed the header to 'Created' instead of 'Uploaded'. Adding the ability to show uploaded date is actually more complex than it seems, happy to get suggestions on other ways though if I have missed something.

Note that the images listing now longer appears to contain any columns so the fix above only addresses the documents listing.

However, I think a new issue should be created to cover a better long term solution to this, if there is some consensus I will create that issue with a bit of an idea on how to proceed.

## Option A - Add 'updated_at' field using Django's `auto_now_add` feature.

* This would add a new field to both the Document & Image models, which would affect all users of Wagtail that extend this model (with potential migration conflicts to resolve). This is quite minor as it is an 'addition'
* We could then add presentation of 'updated at' in the documents listing AND the 'view image' page within admin.
* This is a reasonably simple model change and might come in handy for the future having this field, even if not always shown.
* Note: 'Updated at' would NOT be uploaded at, this date would be affected by new uploads AND changes such as a title change.

## Option B - Add a specific 'uploaded_at' field

* This could work in line with Option A, essentially be specific about this data with a new field.
* Minimal changes to the model means this is quite simple.
* Risk is that we will 'forget' to ensure this field gets kept in sync in some parts of the code.

## Option C - Modify the `uploaded_by_user` relationship

* This would modify the uploaded_by_user field to be a relationship to another table, that can store both the user AND the date the upload took place, with the potential to add extra metadata about uploads in the future (even a 'history' of changes, without actually storing the old files themselves).
* This would be a more complex change and likely break a fair few things and need some solid thought, especially as Wagtail users that have their own image/document model would be affected by this.

I would love any thoughts on this but for now the PR above should resolve the immediate issue of this ticket.

lb-:
[...]

PR #5638, opened 2019-10-20
Rename 'uploaded' column documents list to 'created'

lb-:
Quick resolution to #2827 - the column title `uploaded` in the documents list is misleading, as it actually references the field `created_at`. This is the simplest way to resolve the confusion expressed in the issue.

Notes
* The issue talks about the images list also but I cannot see any kind of table style listing of images, it is likely that this was created when the images listing was quite different.
* I have put some comments on the issue about some big picture fixes but they all involve some kind of model changes and I think a new issue should be created based on the direction we want to go.

* Do the tests still pass? 👍 
* For Python changes: Have you added tests to cover the new/fixed behaviour? N/A - sorting tests still pass
* For new features: Has the documentation been updated accordingly? N/A - listing just reflects expected behaviour now.


lb-:
@loicteixeira  - thanks for that, great pickup!
Could you take another look, if you sign off on approval I am happy to merge.
I also updated a bit of the text in the documentation listing as it appeared to be out of date with what Wagtail actually shows now.

loicteixeira:
Looks good to me :shipit:

lb-:
Thanks @loicteixeira  - merged in 57d44a38bd35f7ea04683bf81692a313ff3419fd

PR #5639, opened 2019-10-20
Add hook: get_document_form

jnsdrtlf:
Added a new wagtail hook called `get_document_form`, called when the document form is generated, e.g. when a admin user edits a document. This is especially useful for custom document models.

Is an **additional test** necessary?

**Documentation has been updated**: See 9af3950.

### DIFF
--- a/docs/reference/hooks.rst
+++ b/docs/reference/hooks.rst
@@ -909,7 +909,7 @@ Page serving
             return HttpResponse("<h1>bad googlebot no cookie</h1>")
 
 
-Document serving
+Documents
 ----------------
 
 .. _before_serve_document:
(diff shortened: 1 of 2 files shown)


jnsdrtlf:
I can add a test but I first wanted to make sure whether this hook is going to be implemented. I guess that's what the **Needs Design Decision** label is for?

loicteixeira:
Correct. Although it seems to mirror what exists for images, there are no previous issue discussing this matter so we should validate the need and the implementation first. Test creation can indeed wait for that decision.

I don't have much time to review PRs at the moment so I'm only tagging them so it's easier for other members of the core team to know what actions need to be taken. Sorry if it caused confusion.

ababic:
Hi @jnsdrtlf. Really appreciate you being proactive and getting stuck-in here (thank you!!), but this is one of those things where I think a bit of further discussion could yield a better result, so I've raised an issue (#6045) to allow for this. It would love to hear whether or not the solution proposed there would fit your needs, or if you have any other ideas on how to improve it.

Let's keep this PR open for now, as it's certainly a valid approach, and in-line with what exists already. Let's just wait for a little more input first before we go any further.

PR #5640, opened 2019-10-21
Fixed #5172 -- Always use the new page copy to create the latest revision

### LINKED ISSUE #5172: Cannot copy a draft page with OneToOneField
### Issue Summary

When making a copy of a Page with a `OneToOneField`, we get an error if we uncheck the `Publish copied page` checkbox. Making a copy with publishing turned on works as expected provided the `OneToOneField` is listed in `exclude_fields_in_copy`.  For some reason turning off publishing results in validation which ignores `exclude_fields_in_copy`.

### Steps to Reproduce

2. Edit models.py as follows:

```
from django.contrib.auth.models import User
from django.db import models

class HomePage(Page):
    foo = models.OneToOneField(User, on_delete=models.SET_NULL, null=True, blank=True)
    exclude_fields_in_copy = ['foo']

    content_panels = [
        FieldPanel('foo'),
    ]
```

3. Make a page with `foo` set to your superuser. Attempt to copy the page with `Publish copied page:` unchecked. Expect a successful copy. Result:

```
[...]
  File "/Users/lucas/.local/share/virtualenvs/test_wagtail-PLuuhPI6/lib/python3.6/site-packages/wagtail/core/models.py", line 1171, in copy
    latest_revision_as_page_revision = latest_revision.save_revision(user=user, changed=False)
  File "/Users/lucas/.local/share/virtualenvs/test_wagtail-PLuuhPI6/lib/python3.6/site-packages/wagtail/core/models.py", line 635, in save_revision
    self.full_clean()
  File "/Users/lucas/.local/share/virtualenvs/test_wagtail-PLuuhPI6/lib/python3.6/site-packages/wagtail/core/models.py", line 436, in full_clean
    super().full_clean(*args, **kwargs)
  File "/Users/lucas/.local/share/virtualenvs/test_wagtail-PLuuhPI6/lib/python3.6/site-packages/django/db/models/base.py", line 1152, in full_clean
    raise ValidationError(errors)
django.core.exceptions.ValidationError: {'foo': ['Home page with this Foo already exists.']}
```

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: 

Yes

### Technical details

* Python version: Python 3.6.1
* Django version: 2.1.7
* Wagtail version: 2.4
* Browser version: Chrome 72

czpython:
When copying a page without publishing it, the latest_revision would always contain values from the original page which would trigger a database error when trying to save the page.

I realize I'm missing tests here, I'm was hoping to get some feedback on the fix before starting on them.

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -1177,16 +1177,9 @@ class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):
 
         # Create a new revision
         # This code serves a few purposes:
-        # * It makes sure update_attrs gets applied to the latest revision
         # * It bumps the last_revision_created_at value so the new page gets ordered as if it was just created
         # * It sets the user of the new revision so it's possible to see who copied the page by looking at its history
-        latest_revision = page_copy.get_latest_revision_as_page()
-
-        if update_attrs:
-            for field, value in update_attrs.items():
-                setattr(latest_revision, field, value)
-
-        latest_revision_as_page_revision = latest_revision.save_revision(user=user, changed=False)
+        latest_revision_as_page_revision = page_copy.save_revision(user=user, changed=False)
         if keep_live:
             page_copy.live_revision = latest_revision_as_page_revision
             page_copy.last_published_at = latest_revision_as_page_revision.created_at


ababic:
Hi @czpython. Thank you very much for this. It looks as though in fixing this behaviour you have also removed the ability for override values to be specified and applied to the revision, which I don't think is something we want. I wonder whether it might be worth looking at the `Page.get_latest_revision_as_page()` and correct the behaviour if the page is an unpublished copy?

gasman:
#5172 was reportedly fixed by #8048, so closing.

Issue #5632, opened 2019-10-17
collectstatic with ManifestStaticFilesStorage fails on 2.7rc1


gasman:
Reported on Slack: after upgrading to 2.7rc1 on a project using ManifestStaticFilesStorage, `./manage.py collectstatic` fails with `ValueError: Missing staticfiles manifest entry for 'wagtailadmin/js/date-time-chooser.js'`. Stack trace: https://dpaste.de/nkYx

This is probably because as of #5527, Django's `static` handler is now being called from the module scope, while models.py is being imported. Form media definitions in admin/widgets.py previously looked like

    class SomeWidget:
        class Media:
            js = ['wagtailadmin/js/date-time-chooser.js']

relying on the path to be converted to a URL at rendering time by the Media object's own logic - but this has changed to

    class SomeWidget:
        class Media:
            js = [versioned_static('wagtailadmin/js/date-time-chooser.js')]

`versioned_static` calls `static`, and so this ends up being called during `manage.py collectstatic`, before the manifest file has been built.

gasman:
Steps to reproduce:

```
pip install wagtail==2.7rc1
wagtail start collectstaticfail
cd collectstaticfail
echo "SECRET_KEY='slgiruvbnkxjdfhbgnxjfdhgb'" > collectstaticfail/settings/local.py
rm -f static/staticfiles.json  # remove manifest file from any previous successful run
./manage.py collectstatic --settings=collectstaticfail.settings.production
```

Issue #5628, opened 2019-10-16
Admin API shouldn't be affected by global REST_FRAMEWORK configuration


kaedroho:
For example, if you remove the `SessionAuthentication` class from the `default_authentication_classes` key of the `REST_FRAMEWORK` setting, the explorer menu will break.

The admin API should override all defaults so it won't break if any are changed.

inostia:
@kaedroho I just saw this issue. See my comment here: https://github.com/wagtail/wagtail/issues/5585#issuecomment-543313909

thibaudcolas:
Closing in favour of #4083.

PR #5643, opened 2019-10-22
Ensure admin API always uses SessionAuthentication backend

### LINKED ISSUE #5585: Admin API returning no pages in Wagtail 2.6
Multiple users have reported no pages showing in the explorer menu after upgrading from Wagtail 2.5 to 2.6; checking admin API responses shows that no pages are being returned for the URL `admin/api/v2beta/pages/?child_of=1&for_explorer=1`.

https://stackoverflow.com/questions/57825908/wagtail-admin-api-call-for-fetching-child-pages-returns-0-in-the-explorer
https://stackoverflow.com/questions/58137551/menu-navigator-disappeared-after-upgrading-to-wagtail-2-6-from-2-5
and Slack support conversation copied below.

Unfortunately no-one's been able to pin this down to a reproducible test case yet. #5069 made changes to this part of the admin API in 2.6, but given that this bug occurs for superusers, it's hard to see how that can be the culprit (since `user_perms.explorable_pages()` returns early with `Page.objects.all()` for them).

```
miric 1:43 PM
I’ve just upgraded Wagtail from 2.3 to 2.6.1 and made migrations. My project is seemed to work but I’ve noticed Pages menu in Admin panel is empty now. What might be a reason of this? Response for ?child_of=1&for_explorer=1 is almost empty :disappointed:

mattwestcott  22 days ago
@miric First thing to try is to clear your browser cache / force refresh - a lot of people have reported broken menus because they had outdated javascript files cached

mattwestcott  22 days ago
Also, if this is a production site (i.e. NOT running locally with manage.py runserver), make sure you've re-run ./manage.py collectstatic

miric  22 days ago
Unfortunately they didn’t worked.

mattwestcott  22 days ago
OK - what do you mean by the response for ?child_of=1&for_explorer=1 being "almost empty"? Normally it would have one entry in it

miric  22 days ago
Here are two screenshots of responses from issue project and normal project:
<responses.jpg>

mattwestcott  22 days ago
Hmm, that's not good - pages have disappeared for some reason. Can you still get at them from /admin/pages/ and see them on the front-end?

miric  22 days ago
Yes. I see root Translatable site root page (I use WagtailTrans)

mattwestcott  22 days ago
Running manage.py fixtree might help... otherwise I'm out of ideas.

mattwestcott  22 days ago
If you can downgrade back to 2.3 and then upgrade one version at a time (2.4, 2.5, 2.6), that might give us some more clues

miric  22 days ago
I did it right after upgrade. No problems found. I’ll restore my DB and try to upgrade from very beginning

miric  22 days ago
Menu has been broken after 2.5.2>2.6 upgrade

mattwestcott  22 days ago
Are you logged in as a superuser/administrator, or a user with limited permissions? There were some changes to the permission model in 2.6 that might affect this: https://github.com/wagtail/wagtail/pull/5069

KalobTaulien  22 days ago
I had this happen once and ./manage.py fixtree is the trick for me as @mattwestcott suggested.

miric  22 days ago
Done right after 2.5.2>2.6 upgrade. No problems found (edited) 

KalobTaulien  22 days ago
Call me a tree cuz I’m stumped :thinking_face:

miric  22 days ago
@mattwestcott Yes, I’m on Admin level (edited) 

miric  22 days ago
It seems I have to roll back to 2.5.2 because I’m on budget with estimates. I’ll back to this some day if no solutions for now

miric  22 days ago
There were no any Wagtail-related migrations during 2.5.2>2.6. Is this OK? (edited) 

mattwestcott  22 days ago
@miric That's entirely possible, yes.

```

gasman:
Fixes #5585 and (at least partly) #5628. I suggest that we backport this to 2.6.x (for a 2.6.3 bugfix release) and 2.7.x (for release candidate 2, along with a fix for #5632).

I would have preferred to test this with a new test with a suitable `@override_settings` decorator, but that doesn't work, presumably because DRF only checks that setting on application startup when constructing endpoints - so we have to take the slightly more ham-fisted approach and set it in `wagtail.tests.settings` (which, without the fix, breaks the existing tests `wagtail.admin.tests.api.test_pages.TestAdminPageListing.test_for_explorer_filter` and `wagtail.core.tests.test_page_permissions.TestPagePermission.test_explorable_pages_in_explorer`).

### DIFF
--- a/wagtail/admin/api/endpoints.py
+++ b/wagtail/admin/api/endpoints.py
@@ -1,5 +1,7 @@
 from collections import OrderedDict
 
+from rest_framework.authentication import SessionAuthentication
+
 from wagtail.api.v2.endpoints import PagesAPIEndpoint
 from wagtail.api.v2.filters import (
     ChildOfFilter, DescendantOfFilter, FieldsFilter, ForExplorerFilter, OrderingFilter,
@@ -13,6 +15,7 @@ from .serializers import AdminPageSerializer
 
 class PagesAdminAPIEndpoint(PagesAPIEndpoint):
     base_serializer_class = AdminPageSerializer
+    authentication_classes = [SessionAuthentication]
 
     # Use unrestricted child_of/descendant_of filters
     # Add has_children filter
--- a/wagtail/tests/settings.py
+++ b/wagtail/tests/settings.py
@@ -220,3 +220,13 @@ WAGTAILADMIN_RICH_TEXT_EDITORS = {
         'WIDGET': 'wagtail.tests.testapp.rich_text.CustomRichTextArea'
     },
 }
+
+
+# Set a non-standard DEFAULT_AUTHENTICATION_CLASSES value, to verify that the
+# admin API still works with session-based auth regardless of this setting
+# (see https://github.com/wagtail/wagtail/issues/5585)
+REST_FRAMEWORK = {
+    'DEFAULT_AUTHENTICATION_CLASSES': [
+        'rest_framework.authentication.BasicAuthentication',
+    ]
+}


gasman:
Merged in 53f5b1a96b361884e4aab655a8e770f98b0a2b1b (2.6.x), 46615c2912fc3acb2455a2d40bb6a1e7438abad8 (2.7.x), 605cedca7897a02e2d586ddf6ff1999a3e16fd1c (master). The admin APIs for images and documents don't have behaviour that's dependent on `request.user`, so aren't essential for getting a 2.6 bugfix release out - but will leave #5628 open to cover that.

PR #5644, opened 2019-10-22
Avoid calling versioned_static from static form media definitions

### LINKED ISSUE #5632: collectstatic with ManifestStaticFilesStorage fails on 2.7rc1

gasman:
Fixes #5632

Defining media definitions with `class Media:` means that those lines are evaluated on application startup; this means that they cannot contain calls to the Django static file finder, because ManifestStaticFilesStorage doesn't work until collectstatic has completed. Fix this by making those definitions into `def media(self):` methods.

### DIFF
--- a/wagtail/snippets/widgets.py
+++ b/wagtail/snippets/widgets.py
@@ -1,5 +1,6 @@
 import json
 
+from django import forms
 from django.template.loader import render_to_string
 from django.utils.translation import ugettext_lazy as _
 
@@ -41,8 +42,9 @@ class AdminSnippetChooser(AdminChooser):
                 app=model._meta.app_label,
                 model=model._meta.model_name)))
 
-    class Media:
-        js = [
+    @property
+    def media(self):
+        return forms.Media(js=[
             versioned_static('wagtailsnippets/js/snippet-chooser-modal.js'),
             versioned_static('wagtailsnippets/js/snippet-chooser.js'),
-        ]
+        ])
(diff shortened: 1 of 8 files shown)


wagtail/snippets/widgets.py
```diff
@@ -41,8 +42,9 @@ def render_js_init(self, id_, name, value):
                 app=model._meta.app_label,
                 model=model._meta.model_name)))
 
-    class Media:
-        js = [
+    @property
+    def media(self):
```
allcaps: When someone subclassed these, will this change break their code?

Can one use the `media` property method in the same as the `Media` class?

PR #5661, opened 2019-10-27
Fixed #5537 -- Fetch deferred stream fields from db

### LINKED ISSUE #5537: KeyError exception when trying to access a streamfield in a hook
### Issue Summary

I have a page model `SimplePage` with a `content` streamfield.
Everything works fine with that page model in usual cases except when I try to access it from within a hook.

There it will raise a `KeyError` exception.

If I fetch the page from DB again it will work.

Note that doing a `refresh_from_db` will not fix the issue, only a plain `objects.get` will.

The above description shown in action in my debugger:
```
(Pdb++) page
<SimplePage: test>
(Pdb++) page.content
*** KeyError: 'content'
  File "/home/myproject/.local/share/virtualenvs/app-4PlAip0Q/lib/python3.7/site-packages/wagtail/core/fields.py", line 34, in __get__
    return obj.__dict__[self.field.name]
(Pdb++) page._meta.model.objects.get(pk=page.pk).content
[<wagtail.core.blocks.stream_block.StreamValue.StreamChild object at 0x7fba2f43fb10>]
```

### Steps to Reproduce

1. Create a Page model with a simple streamfield
2. Setup a hook and try to access that streamfield
```
@hooks.register("register_page_listing_buttons")
def page_custom_listing_buttons(page, page_perms, is_parent=False):
    page.my_stream_field
```

It should raise.

### Technical details

* Python version: 3.7.4
* Django version: 2.2.4
* Wagtail version:2.6.1
* Browser version: not relevant

czpython:
Fixed #5537
Fetches the field value from db if its not present on the object.

### DIFF
--- a/wagtail/core/fields.py
+++ b/wagtail/core/fields.py
@@ -31,7 +31,12 @@ class Creator:
     def __get__(self, obj, type=None):
         if obj is None:
             return self
-        return obj.__dict__[self.field.name]
+        field_name = self.field.name
+
+        if field_name not in obj.__dict__:
+            # Field is deferred. Fetch it from db.
+            obj.refresh_from_db(fields=[field_name])
+        return obj.__dict__[field_name]
 
     def __set__(self, obj, value):
         obj.__dict__[self.field.name] = self.field.to_python(value)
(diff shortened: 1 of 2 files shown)


gasman:
Thanks @czpython! Merged in 494260f8ee2619e9902a4d2abb9077bce2bd8b58.

PR #5666, opened 2019-10-30
Add select_related to public_q query

acarasimon96:
Prevents *n*+1 query problem when getting a list of pages with view restrictions via `PageQuerySet.public()` or `PageQuerySet.not_public()`. Sites that contain a large number of pages with view restrictions will benefit from this.

---

* For Python changes: Have you added tests to cover the new/fixed behaviour? Yes (only modified existing ones to include query counts)


lb-:
Merged in via ee99f91aa5ad526b8c15e626d5d210de5ccf5723

Issue #5166, opened 2019-03-21
instance.live set to True in "page_unpublished" signal handler


lorenzo-ange:
### Issue Summary

A page instance has the "live" property set to True in "page_unpublished" signal handler function.
This happens when a custom Page model instance is deleted.

### Steps to Reproduce

1. connect a handler function to "page_unpublished" signal with a custom Page model sender
2. from the wagtail admin delete a page of the custom Page model
3. if you debug the handler function you can see that "instance.live" is set to True instead of False

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: yes

### Technical details

I suppose the problem is caused by the commit=False in this piece of wagtail code:

wagtail.core.signal_handlers.py, line: 20
```
def pre_delete_page_unpublish(sender, instance, **kwargs):
    # Make sure pages are unpublished before deleting
    if instance.live:
        # Don't bother to save, this page is just about to be deleted!
        instance.unpublish(commit=False)
```

commit=False won't save the model instance to db, but the "page_unpublished" signal is sent passing "instance=self.specific".
**self.specific will refresh instance data from DB, resetting the "live" property to True.**

wagtail.core.models.py, line: 700
```
page_unpublished.send(sender=self.specific_class, instance=self.specific)
```

Removing commit=False should resolve the issue.

gasman:
Thanks for the report, @lorenzo-ange.
As mentioned at https://github.com/wagtail/wagtail/pull/5669#issuecomment-626613226 , removing the `commit=False` would make this fully logically correct, but would add a performance penalty since we need to make two database queries when deleting pages (one for unpublishing, then another for deletion). Can you give an example of a problem that the current implementation causes, so that we can understand the tradeoffs here and maybe find an alternative (such as setting `live=False` on the in-memory `specific` instance)?

PR #5669, opened 2019-10-30
Fixed live=true on unpublish before delete

fefi95:
Related to issue #5166 

_Removed the `commit=False` on the `pre_delete` handler. This caused the `delete_orphan_pages` command to fail because in the `clean` method is trying to get a parent (that does not have), so I added a `try/catch` for the `get_parent`_

Let me know if you think this is a good fix or any suggestions you have. Thanks!

----

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -436,7 +436,13 @@ class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):
 
     def clean(self):
         super().clean()
-        if not Page._slug_is_available(self.slug, self.get_parent(), self):
+        parent = None
+        try:
+            parent = self.get_parent()
+        except Page.DoesNotExist:
+            pass
+
+        if not Page._slug_is_available(self.slug, parent, self):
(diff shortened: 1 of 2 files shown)


gasman:
It would be useful to have an explanation of the actual problem being fixed here - Wagtail doesn't have a `delete_orphan_pages` command, so I can only assume it comes from some third-party code. As far as I can gather, this code is doing something inside a `page_unpublished` signal handler that is causing `get_parent` to fail, but without seeing that code it's impossible to know whether Wagtail or the third-party code is at fault.

I'm also not clear what the relevance of `live=True` is - the result of `get_parent` should not be affected by the `live` flag.

The `commit=False` in the original code exists as a performance optimisation - if a lot of pages are being deleted at once, we'd end up with twice as many queries as necessary, first to update the published status and then to delete the page. To be fully correct, the `page_unpublished` signal should be triggered in between these steps, but given the performance penalty, I'd like to understand what negative consequences the current (not-logically-correct) implementation has.

gasman:
Closing for now due to lack of detail on what's being fixed here - happy to reopen if you can provide further information.

PR #5672, opened 2019-11-02
Allow usage of __init_subclass__() with Page

sir-sigurd:


wagtail/core/models.py
```diff
@@ -192,22 +192,22 @@ def get_queryset(self):
 
 class PageBase(models.base.ModelBase):
     """Metaclass for Page"""
-    def __init__(cls, name, bases, dct):
-        super(PageBase, cls).__init__(name, bases, dct)
+    def __init__(cls, name, bases, dct, **kwargs):
+        super(PageBase, cls).__init__(name, bases, dct, **kwargs)
```
loicteixeira: While you're at it, you might as well remove `PageBase, cls` in `super` since we do not support Python 2.x anymore.

lb-:
Updating labels, it's been a long while without a review sorry so this PR will need some info for context and also a rebase/update.

lb-:
With no details and use cases for this feature described after a few years I think we are best to close this. 

If anyone feels otherwise, please add a comment. I'll look at closing in a week or so.

PR #5676, opened 2019-11-05
Implemented "ancestors" field in Admin API

kaedroho:
Split from: https://github.com/wagtail/wagtail/pull/3038

This adds an optional "ancestors" field to the Admin API which serializes the ancestry of the page. This is required to implement the breadcrumb on the React page chooser.

### DIFF
--- a/wagtail/admin/api/serializers.py
+++ b/wagtail/admin/api/serializers.py
@@ -104,7 +104,7 @@ class PageAncestorsField(Field):
         return instance
 
     def to_representation(self, page):
-        serializer_class = get_serializer_class(Page, ['id', 'type', 'detail_url', 'html_url', 'title'], meta_fields=['type', 'detail_url', 'html_url'], base=PageSerializer)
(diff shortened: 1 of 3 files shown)


wagtail/admin/api/serializers.py
```diff
@@ -76,8 +76,48 @@ def to_representation(self, page):
...
+            },
+            "title": "Home"
+        }
+    ]
+    """
+    def get_attribute(self, instance):
+        return instance
+
+    def serialize_page(self, page):
+        serializer_class = get_serializer_class(page.__class__, ['id', 'type', 'detail_url', 'html_url', 'title'], meta_fields=['type', 'detail_url', 'html_url'], base=PageSerializer)
```
ababic: Is it likely that this field will ever be serialising anything other than a vanilla `Page` object here? If not, I wonder if it might be better to define a single serializer class and reuse here, rather than creating a new class each time.

Also, since this is the Admin API, should we be aiming to show `draft_title` for pages rather than `title`? I don't know how widely that's enforced, but I think that's what the non-react explorer displays.

kaedroho: Thanks for the review, I've addressed both of your comments in the commits below.

kaedroho:
Thanks!

PR #5677, opened 2019-11-05
API refactoring

kaedroho:
This refactors various bits of the API backend. See commit log for details.

This is based on my previous admin API PR: https://github.com/wagtail/wagtail/pull/5676


zerolab:
From a cursory look, does this not need to go through the deprecation process? and have upgrade considerations?

kaedroho:
Ahh, yes. I think the renaming of the public API views would need to be backwards compatible. But the admin API has never been stable so we wouldn't need to implement backwards compatibility there.

kaedroho:
Adding needs work to implement deprecation warnings as @zerolab pointed out. But the rest could still be reviewed without this

kaedroho:
I've added the endpoints back now with deprecation warnings

wagtail/admin/api/endpoints.py
```diff
@@ -56,16 +47,20 @@ class PagesAdminAPIViewSet(PagesAPIViewSet):
...
+        Returns the page that is used when the `&child_of=root` filter is used.
+        """
+        return Page.get_first_root_node()
 
-        # Allow pages to be filtered to a specific type
-        try:
-            models = page_models_from_string(request.GET.get('type', 'wagtailcore.Page'))
-        except (LookupError, ValueError):
-            raise BadRequestError("type doesn't exist")
+    def get_base_queryset(self, models=None):
```
gasman: The `models` parameter is bothering me a bit here... I can see why it's a necessary evil (we need to be able to swap out the base model of the queryset so that we can filter on custom `APIField`s defined in the specific page model), but the "if len(models) == 1 then replace the base model, otherwise filter" logic in here feels like too much undocumented magic for something claiming to be a base queryset.

At the very least, I think this parameter needs to be mentioned in the docstring, because as it stands it's not obvious that the "secondary" uses of get_base_queryset (i.e. retrieving the root page for `child_of` / `descendant_of`, and calculating child/descendant counts in c67042922aa74dede6d26ab09baf6a8e48d63764) rely on us _not_ passing a specific page model here.

Beyond that, I think we could improve things a bit by having `get_base_queryset` accept a single model (defaulting to Page), and have the `if len(models)` logic in `get_queryset` so that `get_base_queryset` isn't doing any filtering (other than the implicit filtering that happens 'for free' by querying on a Page subclass, and that should hopefully be fairly clear from the method signature).

As a more drastic change (and I'm not sure how feasible this is) - perhaps instead of a `get_base_queryset`, we could have a `get_base_query_filters` method that returns a `Q()` object that the caller can apply to whatever queryset it wants?

kaedroho:
Rebased the branch to fix merge conflicts, no changes made since @gasman's review yet

wagtail/admin/api/endpoints.py
kaedroho: Done as per we discussed offline

gasman:
Looks good to me! Merged with upgrade considerations note added in 1d8f85c1c594399590c6a5b8a1e096dc134bbfa8.

PR #5679, opened 2019-11-05
Implemented explorer admin API

kaedroho:
This PR adds a separate instance of the admin API for the explorer menu (replacing the ``&for_explorer=1`` filter. Having a separate instance allows us to customise the API for this component in a neater way (we can also customise which page gets the special ID `root`, for example).

In a future PR, I'd like to make it easy for developers to create new instances of the Admin API which they could use to customise the explorer, choosers, search, etc (when we have all these in React).

This is based on my previous admin API refactoring PR: https://github.com/wagtail/wagtail/pull/5677
Which itself is based on https://github.com/wagtail/wagtail/pull/5676

### DIFF
--- a/client/src/components/Explorer/actions.js
+++ b/client/src/components/Explorer/actions.js
@@ -68,7 +68,7 @@ export function toggleExplorer(id) {
       }
 
       // We need to get the title of the starting page, only if it is not the site's root.
-      const isNotRoot = id !== 1;
+      const isNotRoot = id !== 'root';
       if (isNotRoot) {
         dispatch(getPage(id));
       }
--- a/wagtail/admin/tests/api/test_pages.py
+++ b/wagtail/admin/tests/api/test_pages.py
@@ -8,7 +8,7 @@ from django.utils import timezone
 from wagtail.api.v2.tests.test_pages import TestPageDetail, TestPageListing
 from wagtail.core.models import Page
 from wagtail.tests.demosite import models
-from wagtail.tests.testapp.models import SimplePage, StreamPage
+from wagtail.tests.testapp.models import StreamPage
 
 from .utils import AdminAPITestCase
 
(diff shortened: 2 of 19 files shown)


client/src/components/Explorer/actions.js
```diff
@@ -68,7 +68,7 @@ export function toggleExplorer(id) {
       }
 
       // We need to get the title of the starting page, only if it is not the site's root.
-      const isNotRoot = id !== 1;
+      const isNotRoot = id !== 'root';
```
gasman: If I'm not mistaken, the purpose of this line is to treat the _global_ tree root as a special case - if the explorer starts from there, we leave the menu heading at its default value ("Pages") rather than setting it to the root page's title ("Root"). Changing the test to `'root'` (i.e. the user's explorable root) breaks this in a subtle but important way: if a user's explorable root is further down the tree, they will no longer see that page title as the menu heading:

Screenshot 2019-11-28 at 00 24 36

wagtail/admin/tests/api/test_pages.py
```diff
@@ -297,42 +297,6 @@ def test_descendant_of_root_doesnt_give_error(self):
 
         self.assertEqual(response.status_code, 200)
 
-    # FOR EXPLORER FILTER
```
gasman: Rather than removing these tests, shouldn't we be keeping them and pointing them at the new explorer API endpoint instead? (Ditto for `test_explorable_pages_in_explorer` below.)

gasman: As discussed this morning, it would also be good to add a test to confirm that the explorer API works with a page `type` parameter (which wasn't the case for the old version of the code, because the `explorable_pages` filtering was trying to run a `&` operation between the specific-page queryset and a `Page` queryset).

kaedroho: These tests have been added now

gasman: Thanks - new tests look good to me.

PR #5680, opened 2019-11-05
Reference the third party embedly package requirement

tombola:
> Wagtail has built in support for fetching embeds from Embed.ly

But there is a third party dependence on embedly's own maintained python package.

I have added a reference to installing this to the documentation. 

I was not sure at what level this should be described, eg whether to detail `pip install embedly`, or to add to the project's `requirements.txt`.

I could not see any other examples, so to be brief, and because this is in the *advanced topics* I just specified to _pip_ install (intended to disambiguate against adding to INSTALLED_APPS).

### DIFF
--- a/docs/advanced_topics/embeds.rst
+++ b/docs/advanced_topics/embeds.rst
@@ -204,8 +204,10 @@ and a common video playback API which is useful if your site allows videos to
 be hosted on different providers and you need to implement custom controls for
 them.
 
-Wagtail has built in support for fetching embeds from Embed.ly.  To use it, add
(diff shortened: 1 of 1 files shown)


kaedroho:
Looks good to me!

 > I was not sure at what level this should be described, eg whether to detail pip install Embedly, or to add to the project's requirements.txt.

Happy to keep it simple and leave it as `pip install` as I think most developers would know what this means, even if they don't use pip.

gasman:
Merged in e75b74911e883ad7c2cd3829e54776d860addf16 - thanks @tombola!

PR #5681, opened 2019-11-05
Compare foreign keys using pk, not id

### LINKED ISSUE #5678: ForeignObjectComparison breaks when related model has custom primary key
### Issue Summary

Comparing page revisions raises an exception when one of the changed fields is a `ForeignKey` to a model that uses a [custom primary key](https://docs.djangoproject.com/en/2.2/topics/db/models/#automatic-primary-key-fields) (i.e. not `id`).

This looks very similar to #4180, but it's currently failing in 2.6.x.

### Steps to Reproduce

1. Create a page model with a `ForeignKey` to a model with a [custom primary key](https://docs.djangoproject.com/en/2.2/topics/db/models/#automatic-primary-key-fields).
2. Add an initial revision with a value for the related model.
3. Add another revision changing the value of the related model.
4. Try to compare both revisions.
5. Uncaught exception error: *FieldError: Cannot resolve keyword 'id' into field.*

Expected: comparing revisions works correctly, displays the changed foreign key.

I haven't checked with a fresh Wagtail install, but the error happens on our production Wagtail.

### Technical details

* Python version: 3.7.3
* Django version: 2.2.4
* Wagtail version: 2.6.1

haplo:
Trying to compare revisions of a page that includes changes to a foreign key field of a related model that declared a custom primary key failed with an uncaught exception.

The root cause was `ForeignObjectComparison` filtering by the `id` field, which is not present in models that declare a custom primary key.

The solution is simply to filter by `pk` instead of `id`, which always maps to the primary key of the corresponding model.

Include a regression unit test.

Fixes #5678.

### DIFF
--- a/wagtail/admin/compare.py
+++ b/wagtail/admin/compare.py
@@ -288,8 +288,8 @@ class TagsFieldComparison(M2MFieldComparison):
 class ForeignObjectComparison(FieldComparison):
     def get_objects(self):
         model = self.field.related_model
-        obj_a = model.objects.filter(id=self.val_a).first()
-        obj_b = model.objects.filter(id=self.val_b).first()
+        obj_a = model.objects.filter(pk=self.val_a).first()
+        obj_b = model.objects.filter(pk=self.val_b).first()
         return obj_a, obj_b
 
     def htmldiff(self):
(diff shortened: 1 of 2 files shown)


gasman:
Merged in b9f585a6c9ba157710838325ea5d7a5689975f36 - thanks @haplo!

haplo:
@kaedroho @gasman I'm sorry but I forgot to include a changelog entry.

gasman:
@haplo No problem - it's actually easier for us if you _don't_ include the changelog entry, as that way we don't have to fix a merge conflict when other entries are added in the meantime :-)

PR #5685, opened 2019-11-06
Test against Django stable/3.0.x

gasman:
Test against django stable/3.0.x branch in addition to master; also install django-rest-framework master when testing these.


gasman:
Merged in 94bb595d0c25c622fb8db36ab2b335ec5959c298

Issue #5686, opened 2019-11-06
form_template override on struct block can not accept Jinja2 html


bmoe872:
### Issue Summary

When overriding the `form_template` on a custom structblock you can not use Jinja2 templates to override the template. The admin renders the template as a string, as opposed to as HTML. I would have expected it to render the HTML, or I would hope for something in the documentation warning about this.

### Steps to Reproduce

1. Define a custom struct block
2. Use:
```
class Meta:
    form_template = "<path to jinja2 template>"
```
3. The Jinja2 template can look like this:
```
<div class="{{ classname }}">
{% for key, child in children.items() %}
    {{ child.render_form() }}
{% endfor %}
</div>
```
4. Go to admin where you entered this custom structblock, and it will render out the HTML as a string with `"`'s wrapped around it. 

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: yes

### Technical details

* Python version: `3.6.8`
* Django version: `2.2.7`
* Wagtail version: `2.7.0`
* Browser version: `Chrome Version 78.0.3904.87 (Official Build) (64-bit)`
* Jinja2 version: `2.10.1`

bmoe872:
For now I've found a workaround by overriding the `render_form` function on the structblock, but this feels hacky.

```
from django.utils.safestring import mark_safe

    def render_form(self, value, prefix='', errors=None):
        context = self.get_form_context(value, prefix=prefix, errors=errors)

        return mark_safe(render_to_string(self.meta.form_template, context))
```

gasman:
Hi @bmoe872 - thanks for the report.

I think there's a reasonable argument that jinja2 is at fault here, for not returning a SafeString like Django's own template engine does. However, I don't think there's any harm in us correcting for that by adding the `mark_safe`, as that will have no effect on a template engine that does return a SafeString. Happy to accept a PR that adds this.

bmoe872:
@gasman you got it, PR'd!

gasman:
Fixed in #5687.

PR #5687, opened 2019-11-06
Adds mark_safe to the render_form method of StructBlock

bmoe872:
# What this does
Addresses: https://github.com/wagtail/wagtail/issues/5686

Adds `mark_safe` around the `render_form` method of the struct block. 

Jinja2 doesn't seem to return "safe" html, and instead is returning a string of the content, when it really should just be returning the HTML.

### DIFF
--- a/wagtail/core/blocks/struct_block.py
+++ b/wagtail/core/blocks/struct_block.py
@@ -6,6 +6,7 @@ from django.forms.utils import ErrorList
 from django.template.loader import render_to_string
 from django.utils.functional import cached_property
 from django.utils.html import format_html, format_html_join
+from django.utils.safestring import mark_safe
 
 from wagtail.admin.staticfiles import versioned_static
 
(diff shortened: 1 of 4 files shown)


gasman:
Thanks @bmoe872! Could you add a test for this, please?

bmoe872:
@gasman I've added tests that test the Jinja2 rendering engine works with the addition of the `mark_safe` around the `render_to_string` method, and I noticed that there didn't seem to be any tests testing the functionality of a custom form using the Django template engine, but I'm running into some unexpected behavior that I was hoping you might be able to help shed some light on. 


When using the same approach as I have for testing the Jinja2 engine, the Django engine instead throws an error:
`AttributeError: 'Template' object has no attribute 'split'`


I can't seem to find much to shed any light on what is causing this to happen, and having stared at this for most of the day at this point I'm hoping you might have some insights into what's going on, or if its even needed to test custom `form_template` functionality against the Django template engine?

bmoe872:
@gasman I've added tests, and I noticed in testing that there wasn't a test for testing the functionality of adding a custom form template against the Django template engine, so I added a test for that as well.

gasman:
Thanks again @bmoe872! Merged in f8c7d7f5f77b1bb933f0d6a7899e2450f01d373f.

PR #5689, opened 2019-11-08
Update links in documentation to use HTTPS.

martey:
Use HTTPS links instead of non-secure HTTP links wherever possible in documentation.

This is a minor change, only changing the protocol used in links in order to reduce [301 redirects](https://en.wikipedia.org/wiki/HTTP_301) and improve security.

* Do the tests still pass? **yes**
* Does the code comply with the style guide? n/a
* For Python changes: Have you added tests to cover the new/fixed behaviour? n/a
* For front-end changes: Did you test on all of Wagtail’s supported browsers? n/a
* For new features: Has the documentation been updated accordingly? n/a


gasman:
Thanks @martey!

PR #5692, opened 2019-11-09
Move settings reference into separate doc

kaedroho:


gasman:
Looks like there's a link to the old location in https://github.com/wagtail/wagtail/blob/master/docs/getting_started/integrating_into_django.rst which needs to be updated. (That now raises the question of whether getting_started/integrating_into_django.rst duplicates advanced_topics/addtodjangoproject.rst - but I guess we should stick with one fix at a time :-) )

docs/advanced_topics/index.rst
```diff
@@ -8,7 +8,7 @@ Advanced topics
     images/index
     documents/index
     embeds
-    settings
+    addtodjangoproject
```
gasman: Let's use underscore-separated words here, for consistency with other pages (e.g. `third_party_tutorials`).

gasman:
Merged in 96965d2a3ae32e2ecfcde5b77c9df1304d663493 + parents - will add a redirect from advanced_topics/settings to reference/settings on readthedocs too.

PR #5694, opened 2019-11-09
Move all usage of versioned_static to media methods

### LINKED ISSUE #5693: versioned_static problems with ManifestStaticFilesStorage
### Issue Summary

Feels like a continuation of #5632.

Before Wagtail 2.7 it was possible to run commands such as `./manage.py makemigrations  --dry-run --check` *before* running collectstatic, which can be useful for testing missing migrations in CI. However since 2.7 it results in a `ValueError: Missing staticfiles manifest entry` error.

There's a few places with rich text editor plugins where versioned_static is being called when other code is imported, instead of when `.media` is being called on a form or widget. This probably shouldn't happen.

### Steps to Reproduce

```shell
pip install wagtail==2.7
wagtail start versionedstaticdemo
cd versionedstaticdemo
echo "SECRET_KEY='secret'" > versionedstaticdemo/settings/local.py
cat > home/models.py <<EOF
from django.db import models

from wagtail.core.fields import RichTextField


class HomePage(Page):
    rich_text = RichTextField()

        FieldPanel("rich_text", classname="full"),
    ]
EOF
./manage.py makemigrations --settings=versionedstaticdemo.settings.production --dry-run --check
```

Results in:

```
ValueError: Missing staticfiles manifest entry for 'wagtailadmin/js/draftail.js'
```

### Technical details

* Python version: 3.7 (any)
* Django version: 2.2 (any)
* Wagtail version: 2.7+ / master

tomkins:
To avoid staticfiles manifest file checking during loading of models, ensure that all versioned_static calls are only used inside media methods.

~Change versioned_static so that the `static()` call is done *after* the test for hashed staticfiles being used.~

Closes #5693.

To test:

```shell
pip install -e git+https://github.com/tomkins/wagtail.git@versioned-static#egg=wagtail
wagtail start versionedstaticdemo
cd versionedstaticdemo
echo "SECRET_KEY='secret'" > versionedstaticdemo/settings/local.py
cat > home/models.py <<EOF
from django.db import models

from wagtail.core.fields import RichTextField


class HomePage(Page):
    rich_text = RichTextField()

        FieldPanel("rich_text", classname="full"),
    ]
EOF
./manage.py makemigrations --settings=versionedstaticdemo.settings.production --dry-run --check
```


frmdstryr:
Related #5644, appears that was not fully complete.

wagtail/admin/rich_text/editors/hallo.py
```diff
@@ -23,7 +23,12 @@ def construct_plugins_list(self, plugins):
 
     @property
     def media(self):
-        return Media(js=self.js, css=self.css)
+        js = [versioned_static(js_file) for js_file in self.js]
```
gasman: A potential issue here:
Django's `Media` class has some conditional logic in place to [accept either a relative or absolute path](https://docs.djangoproject.com/en/2.2/topics/forms/media/#form-asset-paths) - relative paths will be passed through `static`, while absolute paths will be left unchanged. However, the `static` / `versioned_static` function itself does not include this check - passing an already-absolute path to `static` will mangle it.

In the current version of the code, we're only calling `versioned_static` on literal strings that we know are relative paths. However, by moving the `versioned_static` call here, we'll also be applying it to plugins defined by third-party code outside of Wagtail, and we can't be sure whether that third-party code is passing absolute or relative paths (since both are valid, as per the `Media` class's behaviour).

I think the best fix is to update `versioned_static` to replicate the `Media` behaviour, so that it leaves paths beginning with `/`, `http://` or `https://` unchanged. It feels slightly hacky, but there's already a bit of 'defensive' logic in place there (i.e. not adding the querystring if the result of `static` has one already) so it wouldn't be too much of a leap.

wagtail/images/wagtail_hooks.py
```diff
@@ -95,7 +94,7 @@ def register_image_feature(features):
                 'id': True,
             }
         }, js=[
-            versioned_static('wagtailimages/js/image-chooser-modal.js'),
+            'wagtailimages/js/image-chooser-modal.js',
```
gasman: As far as I can see, the `versioned_static` call that's been removed here never gets added back anywhere else, so we lose the cache-busting behaviour - `wagtail.admin.rich_text.editors.draftail.features.Feature` wraps this in a `Media` object with no modifications, and that object is then used to build `DraftailRichTextArea.plugin_media` and ultimately the result of `DraftailRichTextArea.media`. Reinstating the `versioned_static` will presumably involve refactoring `DraftailRichTextArea` similarly to `HalloRichTextArea`, so that building the media happens lazily in the `media` property rather than on `__init__`. (As an aside, we should probably use [`@cached_property`](https://docs.djangoproject.com/en/2.2/ref/utils/#django.utils.functional.cached_property) on both classes so that the evaluation only happens once.)

tomkins:
My original intention was to avoid tweaking `versioned_static` as that felt like the right thing, so the main focus was to get anything which used it into a media method to avoid being called too early.

However we could just change that function and leave the rest alone. It's a much smaller PR - and feels a lot less invasive.

I'm happy to bring back the original code and address some more of the original feedback if you prefer.

wagtail/admin/staticfiles.py
```diff
@@ -41,11 +41,11 @@ def versioned_static(path):
...
-
-    # if URL already contains a querystring, don't add our own, to avoid interfering
-    # with existing mechanisms
-    if VERSION_HASH is None or '?' in base_url:
-        return base_url
+    # if URL already contains a querystring or is an absolute path, then don't add the version
+    # querystring to avoid changing URLs which might interfere with existing mechanisms, or
+    # doesn't need cache-busting anyway
+    if VERSION_HASH is None or '?' in path or path.startswith(('http://', 'https://', '/')):
+        return path
```
gasman: Afraid this logic isn't correct - the error case occurs when `ManifestStaticFilesStorage` is active and DEBUG is False, i.e. we have a static file backend that handles cache-busting itself, so we want `versioned_static` to revert to the standard behaviour of `static`. In this case, this path is taken (because `VERSION_HASH` is None), but skipping the call to `static` is not valid as it will fail to expand local paths such as `wagtailadmin/js/draftail.js` into valid URLs.

tomkins:
Hopefully the third attempt should be a bit better:

- `versioned_static` will now immediately return the exact string back if it's an absolute URL already - which matches Django's behaviour.
- Updated `DraftailRichTextArea` a bit more, as plugin media can't be merged during `__init__` - combining media together result in the manifest being checked.
- Added `versioned_static` support to draftail features.
- Added `cached_property` to both editors media methods.

PR #5697, opened 2019-11-11
Include reference to Django's {{ block.super }}

DanielSwain:


lb-:
@DanielSwain  - thanks again! merged in via 749d017

PR #5698, opened 2019-11-11
Suppress warning when adding valid `tag_id` filter search field to the index

benpetty:
Screen Shot 2019-11-11 at 3 30 36 PM

I've been getting this warning, but everything seems to be configured correctly as far as I can tell. I have posted some info about my setup at this stack overflow question.

https://stackoverflow.com/questions/58758924/wagtail-postgresql-search-backend-shows-warning-mymodel-search-fields-contains

This PR just makes the warning go away, based on this particular case. I'm sure there is probably a better way to handle this, so if anyone has any suggestions would be awesome.

Thanks!

### DIFF
--- a/wagtail/search/index.py
+++ b/wagtail/search/index.py
@@ -112,6 +112,10 @@ class Indexed:
         for field in cls.get_search_fields():
             message = "{model}.search_fields contains non-existent field '{name}'"
             if not cls._has_field(field.field_name):
+                # Special case for tag_id field
+                from taggit.models import TaggedItemBase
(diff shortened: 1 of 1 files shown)


lb-:
@benpetty - thanks for submitting this PR. I have and a look and I think there needs to be some more investigation as to the root cause. Adding a specific case to ignore a warning for this by its field name is not ideal. 

Would you be ok to raise an issue for this (instead of a PR) and maybe attempt to work out and document the minimum steps to reproduce. E.g. is it postgres specific or maybe field name specific? Maybe the code for generating the warning does not correctly handle some relationship types. 

For now I will close this PR as it may suppress legitimate warnings.

I think the stack overflow question you posted should be added as a reference to the issue raised.

PR #5699, opened 2019-11-12
Add possibility to filter image index by tag.

Hopiu:
screenshot

I added the possibility to filter the images on the image index page by tag. A client has a lot of images and has problems to find them on the image page. This pull request helps to fix this issue.

To make clear, which tag is active, I added some extra CSS – a new class .active – to the button component.

If you're happy with this, I'd make the same changes for the document page.
(There is also a small bug in the chooser template – one is unable to disable the filtering on tags once clicked. I'd look into that too, but first I'd like to get some feedback on the UI)

### DIFF
--- a/client/scss/components/_button.scss
+++ b/client/scss/components/_button.scss
@@ -106,7 +106,7 @@
     }
 
     &.bicolor {
-        border: 0;
+        border: 1px solid transparent;
         padding-left: 3.5em;
 
         &:before {
@@ -122,6 +122,10 @@
             display: block;
             border-top-left-radius: inherit;
             border-bottom-left-radius: inherit;
+        }
+
+        &.button-secondary {
+            border: 1px solid rgba(0, 0, 0, 0.2);
         }
     }
 
@@ -230,10 +234,6 @@
         background-color: $color-white;
         border-color: $color-grey-3;
         color: $color-grey-3;
-    }
-
-    &:not(:disabled):not(.disabled).active {
-        background-color: $color-button-active;
     }
 
     &.button-nostroke {
--- a/client/scss/components/_button.scss
+++ b/client/scss/components/_button.scss
@@ -232,6 +232,10 @@
         color: $color-grey-3;
     }
 
+    &:not(:disabled):not(.disabled).active {
+        background-color: $color-button-active;
+    }
+
     &.button-nostroke {
         border: 0;
     }
--- a/client/scss/settings/_variables.scss
+++ b/client/scss/settings/_variables.scss
@@ -64,7 +64,6 @@ $color-input-error-bg: lighten(saturate($color-red, 28), 45);
 
 $color-button: $color-teal;
 $color-button-hover: $color-teal-darker;
-$color-button-active: $color-teal-dark;
 $color-button-yes: $color-green-dark;
 $color-button-yes-hover: darken($color-button-yes, 8%);
 $color-button-no: $color-red-dark;
--- a/client/scss/settings/_variables.scss
+++ b/client/scss/settings/_variables.scss
@@ -64,6 +64,7 @@ $color-input-error-bg: lighten(saturate($color-red, 28), 45);
 
 $color-button: $color-teal;
 $color-button-hover: $color-teal-darker;
+$color-button-active: $color-teal-dark;
 $color-button-yes: $color-green-dark;
 $color-button-yes-hover: darken($color-button-yes, 8%);
 $color-button-no: $color-red-dark;
--- a/wagtail/images/templates/wagtailimages/images/index.html
+++ b/wagtail/images/templates/wagtailimages/images/index.html
@@ -41,18 +41,20 @@
                     {% endif %}
                 {% endif %}
                 {% if popular_tags %}
-                <li class="taglist">
-                    <h3>{% trans 'Popular tags' %}</h3>
-                    {% for tag in popular_tags %}
-                        {% if tag.name != current_tag %}
-                            <a class="button button-small bicolor icon icon-tag" href="{% url 'wagtailimages:index' %}{% querystring tag=tag.name %}">{{ tag.name }}</a>
-                        {% else %}
-                            <a class="button button-small bicolor icon icon-tag active" href="{% url 'wagtailimages:index' %}{% querystring tag=tag.name %}">{{ tag.name }}</a>
-                        {% endif %}
-                    {% endfor %}
-                    {% if current_tag %}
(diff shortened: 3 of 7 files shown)


Hopiu:
@loicteixeira How can I help getting things foreward? Should I write some tests?

loicteixeira:
Thanks for adding the tests. I've now removed the corresponding tag, leaving only the *needs review* so hopefully another core team member can jump on the review (I unfortunately do not have enough time to review pull requests).

client/scss/components/_button.scss
```diff
@@ -232,6 +232,10 @@
         color: $color-grey-3;
     }
 
+    &:not(:disabled):not(.disabled).active {
```
lb-: I think we need a different approach for styling the 'selected' & 'non-selected'.
Currently there is only one use of this type of button (button.bicolor) and that is when you are entering tags, so this state is associated with being selected.

There is however a .button.secondary we could leverage to indicate an alternate style.

Styling changes needed
* .button.bicolor {  border: 1px solid transparent;} - this way the heights of elements do not appear different across the two states
* .button.bicolor.button-secondary {border: 1px solid #007d7e /* use variable here*/;}  - this will add a border around the tag button, otherwise it looks a bit odd.
* remove the .active class styling & the active variable below

client/scss/settings/_variables.scss
```diff
@@ -64,6 +64,7 @@ $color-input-error-bg: lighten(saturate($color-red, 28), 45);
 
 $color-button: $color-teal;
 $color-button-hover: $color-teal-darker;
+$color-button-active: $color-teal-dark;
```
lb-: as mentioned above, leveraging the .secondary convention that already exists gets us to an alternate style without adding a new convention of this kind of 'active' button.

wagtail/images/templates/wagtailimages/images/index.html
```diff
@@ -31,13 +32,31 @@
...
+                    {% if current_tag %}
+                        <input type="hidden" name="tag" value="{{ current_tag }}" />
+                    {% endif %}
+                {% endif %}
+                {% if popular_tags %}
+                <li class="taglist">
+                    <h3>{% trans 'Popular tags' %}</h3>
+                    {% for tag in popular_tags %}
+                        {% if tag.name != current_tag %}
+                            <a class="button button-small bicolor icon icon-tag" href="{% url 'wagtailimages:index' %}{% querystring tag=tag.name %}">{{ tag.name }}</a>
```
lb-: this line would become the `.secondary` variant.

```
<a class="button button-small button-secondary bicolor icon icon-tag" href="{% url 'wagtailimages:index' %}{% querystring tag=tag.name %}">{{ tag.name }}</a>
```

An added benefit of doing it this way is you get a nice hover state that reflects the 'future' state of being selected.

wagtail/images/templates/wagtailimages/images/index.html
```diff
@@ -31,13 +32,31 @@
...
+                        {% else %}
+                            <a class="button button-small bicolor icon icon-tag active" href="{% url 'wagtailimages:index' %}{% querystring tag=tag.name %}">{{ tag.name }}</a>
```
lb-: This line would just be the 'normal' button, no .active needed now
```
                             <a class="button button-small bicolor icon icon-tag" href="{% url 'wagtailimages:index' %}{% querystring tag=tag.name %}">{{ tag.name }}</a>
```

wagtail/images/templates/wagtailimages/images/index.html
```diff
@@ -31,13 +32,31 @@
...
+                        {% endif %}
+                    {% endfor %}
+                    {% if current_tag %}
+                        <a class="button button-small button-secondary" href="{% url 'wagtailimages:index' %}{% querystring tag='' %}">{% trans 'Clear choice' %}</a>
```
lb-: Might be nice to add `icon icon-cross` here to this button to now differentiate a bit better between the selected tag

wagtail/images/templates/wagtailimages/images/index.html
```diff
@@ -31,13 +32,31 @@
...
-                </ul>
-            </form>
-        {% endif %}
+                    {% if current_tag %}
+                        <input type="hidden" name="tag" value="{{ current_tag }}" />
+                    {% endif %}
+                {% endif %}
+                {% if popular_tags %}
+                <li class="taglist">
+                    <h3>{% trans 'Popular tags' %}</h3>
```
lb-: I think `h3` is probably not suitable here, as there is no other h3 on the page and the nearest h2 is in the menu.

This should probably be something like
`<label for="">{% trans 'Popular Tags:' %}</label>`

Notes;
* the for will need to go somewhere, perhaps to the 'selected' tag (using an id like `current-tag`) as you cannot reference a hidden field with a label's for attribute
* go with title case and colon `:` at the end so this matches the `Collections:` from above
* Might need to do some style tweaks so it looks the same as the `Collections:` text
* A benefit of using `label` is that it will be display block, so it will wrap nicer on small devices and again look similar to the field above

Hopiu: I tried to keep it close to how it's done in [chooser.html](https://github.com/wagtail/wagtail/blob/eaad01308142337739d7edd9f08bd8e714be23f0/wagtail/images/templates/wagtailimages/chooser/chooser.html#L28), that's why I chose an `h3` tag.
It probably makes sense, to change it there too. What do you think?

Anyway the chooser.html template is also missing a `Clear choice` button. First, I'd like to finish what I started here, before I start working on other places.

Hopiu: As to [MDN](https://developer.mozilla.org/de/docs/Web/HTML/Element/label#Attributes) the labels for attribute should reference the id of a [labelable](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Form_labelable) form-related element ( `<button>`, `<input>`, `<keygen>`, `<meter>`, `<output>`, `<progress>`, `<select>`, and `<textarea>`).

I think a `fieldset` with a `legend` element would be the proper way to go. What do you think?

lb-: @Hopiu - good investigation, I had not dug that deep into label references :) - I agree that a fieldset would make sense as we are using <a/> as buttons here (hopefully we will fix that in the future as a stand alone issue).

And yeah - it would be nice to align with the chooser here but they are in different contexts and we want to try and work with the markup/design/interaction already present on this page I reckon. plus h3 may not be suitable there either, but that can be done in a future improvement haha.

lb-:
@Hopiu - I fixed up a small padding issue (only apparent on mobile/small devices) and merged in.
Epic stuff! Thanks for your contribution.
Merged in via 233e7f5

PR #5705, opened 2019-11-13
Update document size when changing document file

### LINKED ISSUE #5704: Documents: file_size and file_hash not updated when document file changed
### Issue Summary

In `edit` function in file https://github.com/wagtail/wagtail/blob/master/wagtail/documents/views/documents.py#L143 

```
        if form.is_valid():
            doc = form.save()
            if 'file' in form.changed_data:
                doc.file_size = doc.file.size

                # Set new document file hash
                doc.file.seek(0)
                doc._set_file_hash(doc.file.read())
                doc.file.seek(0)

                # if providing a new document file, delete the old one.
                # NB Doing this via original_file.delete() clears the file field,
                # which definitely isn't what we want...
                original_file.storage.delete(original_file.name)

            # Reindex the document to make sure all tags are indexed
            search_index.insert_or_update_object(doc)

            messages.success(request, _("Document '{0}' updated").format(doc.title), buttons=[
                messages.button(reverse('wagtaildocs:edit', args=(doc.id,)), _('Edit'))
            ])
            return redirect('wagtaildocs:index')

            # --------------------------------------
            # Where doc is saved?
            # --------------------------------------
```

we see that document form is being saved... then... when "file" was changed it updates its "file_size" and "file_hash" property but it does not save "doc" object anywhere so that these properties remains unchanged

### Steps to Reproduce

1. Add new document
2. Check file_size and file_hash in database
3. Edit uploaded document - change file and save
4. File has changed but file_hash and file_size is unchanged

### Technical details

* Python version: 3.6.4
* Django version: 2.2.6
* Wagtail version: 2.6.3

rinti:
Fixes #5704

### DIFF
--- a/wagtail/documents/tests/test_admin_views.py
+++ b/wagtail/documents/tests/test_admin_views.py
@@ -321,6 +321,33 @@ class TestDocumentEditView(TestCase, WagtailTestUtils):
         self.assertContains(response, self.document.usage_url)
         self.assertContains(response, 'Used 0 times')
 
+    def test_reupload_different_file_size(self):
+        """
+        Checks that reuploading the document file with a different file
+        changes the file size (see #5704).
+        """
+        # Build a fake file, and create it through the admin view
+        # since self.document doesn't have a file_size set.
+        fake_file = SimpleUploadedFile('some_file.txt', b'this is the content')
+        post_data = {
+            'title': "My doc",
+            'file': fake_file,
+        }
+        response = self.client.post(reverse('wagtaildocs:add'), post_data)
+
+        document = models.Document.objects.get(title="My doc")
+        old_file_size = document.file_size
+
+        new_file = SimpleUploadedFile(document.filename, b'less content')
+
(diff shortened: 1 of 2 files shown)


gasman:
Thanks @rinti!

Between the `form.save()` and the `doc.save()`, the document will be saved twice, and I think that's something we should avoid if possible - for example, a developer might have registered a signal that sends a notification on save. Would it be possible to rearrange the logic to avoid this?

rinti:
@gasman Good call! Would changing `doc.save(commit=False)` solve it, or would this mean that the `pre_save`-signal gets called twice still?

gasman:
`commit=False` should be fine, yep - the signal will only be triggered on the model save, not the form.

rinti:
@gasman Cool. I had to add an else because we cannot remove the original file before the new document is commited.

lb-:
@rinti  - this is awesome, thanks so much! I tested locally before and after and this code works as expected. Thanks @gasman  for picking up the the potential issues with signals being called twice.

Merged in via 4945acf

PR #5706, opened 2019-11-13
Remove django admin for Page model

### LINKED ISSUE #5703: Slow page admin - timeout with many users in DB
### Issue Summary

I have tens of thousands users in my DB and the `page` admin change site is very slow - so slow that it timeouts.

### Steps to Reproduce

1. fill many users to your DB
2. create Wagtail page
3. go to `admin`->`Wagtail core`->`page`
4. page is very slow or timeouts

Wagtail should use `autocomplete_fields` or `raw_id_fields` for the user field.

rinti:
Fixes #5703 according to gasman's suggestion to remove it completely


lb-:
@gasman  - does this change need a deprecation note/warning in the release notes do you think?

I am happy to write it up when I merge this in but just wondered as some people may be reliant on having this available in Django admin. Additional to that, should we document how to put Page in Django admin (probably a separate issue to be created).

gasman:
@lb- Yep, I think there should be a release note detailing how to restore the current behaviour. No need for further documentation beyond that, I'd say - editing pages through django-admin isn't something we want to officially support, so if anyone wants to go down that route then it's up to them to figure it out for themselves (or, better, something for the wider community to document).

lb-:
@rinti  - thanks so much for sorting this out, really appreciate it!

Merged in via 3cb59c2667ca8a466812046bd9fd3747dbe6e209

PR #5707, opened 2019-11-13
Fix field row padding

### LINKED ISSUE #5691: Layout Bug FieldRowPanel with heading
If you make the combination with FieldRowPanel an heading the header overlaps the fields.

```
  FieldRowPanel(
            [
                FieldPanel('year'),
                FieldPanel('call')
            ], heading="year and call"
        ),
```

rinti:
Fixes #5691 - first time doing anything style related in Wagtail - so I'm not sure if this is the right approach. I just copied the same padding as the fieldset.

Before:
Before

After:
After

* For front-end changes: Did you test on all of Wagtail’s supported browsers? Tested in Chrome, Firefox and Safari.

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/scss/layouts/page-editor.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/layouts/page-editor.scss
@@ -63,6 +63,10 @@ $object-title-height: 40px;
         padding-right: 0;
     }
 
+    .field-row {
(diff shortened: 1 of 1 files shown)


lb-:
@rinti  - epic! thanks for fixing this, I have validated on multiple browsers and at different resolutions, works great!

Merged in via 5a8f468c2d62964e64c07448377b720f60392a5c

PR #5708, opened 2019-11-13
Add missing block.super in extra_css/extra_js blocks in wagtailsettings, modeladmin templates

acarasimon96:
This fixes a bug where custom styles and scripts provided in `extra_css` and `extra_js` template blocks from the user's project files aren't being applied/run in `wagtail.contrib.settings` edit pages.

In my Wagtail project, I extended the `wagtailadmin/admin_base.html` template to add my custom styles for the entire admin site like this:

```html
{% extends 'wagtailadmin/admin_base.html' %}

{% block extra_css %}
<!-- custom CSS styles go here -->
{% endblock %}

{% block extra_js %}
<!-- custom scripts go here -->
{% endblock %}
```

The [`wagtailsettings/edit.html`](https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/settings/templates/wagtailsettings/edit.html#L44-L54) template overrides the above blocks, causing my styles/scripts to not show in the page's source.

---

* For front-end changes: Did you test on all of Wagtail’s supported browsers? yes (Chrome 78, Firefox 70, Edge 18)

### DIFF
--- a/wagtail/contrib/settings/templates/wagtailsettings/edit.html
+++ b/wagtail/contrib/settings/templates/wagtailsettings/edit.html
@@ -41,12 +41,14 @@
 
 {% endblock %}
 
-{% block extra_css %}
+{% block css %}
+    {{ block.super }}
     {% include "wagtailadmin/pages/_editor_css.html" %}
     {{ edit_handler.form.media.css }}
     {{ site_switcher.media.css }}
 {% endblock %}
-{% block extra_js %}
+{% block js %}
+    {{ block.super }}
     {% include "wagtailadmin/pages/_editor_js.html" %}
     {{ edit_handler.form.media.js }}
     {{ edit_handler.html_declarations }}
--- a/wagtail/contrib/settings/templates/wagtailsettings/edit.html
+++ b/wagtail/contrib/settings/templates/wagtailsettings/edit.html
@@ -41,13 +41,13 @@
 
 {% endblock %}
 
-{% block css %}
+{% block extra_css %}
     {{ block.super }}
     {% include "wagtailadmin/pages/_editor_css.html" %}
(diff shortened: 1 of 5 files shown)


wagtail/contrib/settings/templates/wagtailsettings/edit.html
```diff
@@ -41,12 +41,14 @@ <h1 class="icon icon-cogs">
 
 {% endblock %}
 
-{% block extra_css %}
+{% block css %}
```
lb-: Please double check this, I think it should be left as `{% block extra_css %}` - otherwise the documentation will now be wrong.

wagtail/contrib/settings/templates/wagtailsettings/edit.html
```diff
@@ -41,12 +41,14 @@ <h1 class="icon icon-cogs">
...
+    {{ block.super }}
     {% include "wagtailadmin/pages/_editor_css.html" %}
     {{ edit_handler.form.media.css }}
     {{ site_switcher.media.css }}
 {% endblock %}
-{% block extra_js %}
+{% block js %}
```
lb-: same here, I am pretty sure this line should be left as `{% block extra_js %}`

acarasimon96:
@lb You're right. It turns out that even the core Wagtail admin templates such as page listing and editor views do have their specific styles and scripts in `extra_css` and `extra_js` blocks respectively, something that I overlooked when I submitted this PR. So thanks to your suggestion, this looks to be an easy fix to do.

lb-:
@acarasimon96  - merged in, thanks so much for fixing this!

1b0aed7b20ed222744bbdcf741db8119c3483b3b

PR #5720, opened 2019-11-26
Update links in docs to point at stable Django version

peteandrew:
Updated a few links in the docs that were previously linking to old or dev versions of the Django docs.


lb-:
@peteandrew  - thanks for this, has been merged in
https://github.com/wagtail/wagtail/commit/abd8320dc3795324d6d782ebc4cd18c5dbaf8d96

PR #5721, opened 2019-11-26
Chunk Cloudflare frontend cache purging to fit within API limits

tomusher:
This updates the Cloudflare frontend cache backend to chunk purge requests to clear 30 URLs at a time (as required by their API https://api.cloudflare.com/#zone-purge-files-by-url).

Expect there may be a performance hit if someone is doing a big batch purge while handling a request, but hopefully this is rare.

### DIFF
--- a/wagtail/contrib/frontend_cache/tests.py
+++ b/wagtail/contrib/frontend_cache/tests.py
@@ -203,6 +203,9 @@ class MockCloudflareBackend(CloudflareBackend):
         pass
 
     def _purge_urls(self, urls):
+        if len(urls) > self.CHUNK_SIZE:
+            raise Exception("Cloudflare backend is not chunking requests as expected")
+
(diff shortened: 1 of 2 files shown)


wagtail/contrib/frontend_cache/tests.py
```diff
@@ -198,6 +198,14 @@ def purge(self, url):
         PURGED_URLS.append(url)
 
 
+class MockCloudflareBackend(CloudflareBackend):
+    def __init__(self, config):
+        pass
+
+    def _purge_urls(self, urls):
+        PURGED_URLS.extend(urls)
```
kaedroho: Probably worth raising an exception here if `len(urls) > 30` so that the test would fail if there was no chunking

tomusher: Have added this, thanks!

PR #5730, opened 2019-12-06
Add verbose_name_plural to form submission due to translation mismatch

jjanssen:
Right now the translations for form submissions are not handled properly as it adds an additional 's' to the end. This adds a `verbose_name_plural` to the abstract model so this eventually can be translated accordingly.


gasman:
Thanks @jjanssen! Merged in 146853914eef6ddbd29233ea30c2346c3847ede8.

PR #5731, opened 2019-12-08
Update admin views to consistently return a TemplateContext and ensure preferred language applies when rendering

ababic:
As explained [in the Django docs](https://docs.djangoproject.com/en/stable/ref/template-response/#the-rendering-process), when using `TemplateResponse`, the rendering of the template is delayed until necessary, and middleware and view decorators have the opportunity to make changes before rendering happens. 

Specifically, the [`process_template_response()`](https://docs.djangoproject.com/en/stable/topics/http/middleware/#process-template-response) method on custom middleware gets to do its thing (which doesn't happen when using the `render()` shortcut), giving developers an important and convenient hook to make certain types of customisation (that might otherwise require replacing entire function-based views or monkey-patching).

Most class-based views in Wagtail subclass Django's [`TemplateResponseMixin`](http://ccbv.co.uk/projects/Django/2.2/django.views.generic.base/TemplateResponseMixin/) somewhere along the line, and so  already return a `TemplateResponse` - So these changes also greatly improve consistency.

While implementing these changes, I also noticed an issue that seems not to have been reported yet: The `wagtail.auth.require_admin_access` decorator isn't working quite as expected in many cases: Specifically, where views return a `TemplateResponse`. Because template rendering is delayed, it happens _outside_ the `translation.override()` context manager, meaning the language is only active while view logic is run - the template itself is still rendered in the default language. I've fixed this here by invoking the response's `render()` method if it has one - which means the language override is fully applied.

### DIFF
--- a/wagtail/admin/auth.py
+++ b/wagtail/admin/auth.py
@@ -1,7 +1,6 @@
 from functools import wraps
 
 import l18n
-
 from django.contrib.auth import get_user_model
 from django.contrib.auth.views import redirect_to_login as auth_redirect_to_login
 from django.core.exceptions import PermissionDenied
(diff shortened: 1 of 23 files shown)


gasman:
Thanks @ababic! Merged in 1f2b6fb449781fba0bcd313cca0158812a9bb1e6, minus the `require_admin_access` fix as (if I'm reading things right...) @jacobtoppm's fix in #5910 achieves the same thing with additional edge cases accounted for.

PR #5733, opened 2019-12-13
Move wagtail_urls to the end of template urls.py

nimasmi:
The `wagtail_urls` patterns is a catch-all list of urlpatterns, and will
prevent any patterns later in the list from being matched. The default
case when Django is in debug mode for local development is to use
`django-admin.py runserver`, and in this case the static patterns in the
example [are redundant][1]. However for anyone using a different server
for local development, this makes them work again.

[1]: https://docs.djangoproject.com/en/3.0/howto/static-files/#serving-static-files-during-development

### DIFF
--- a/wagtail/project_template/project_name/urls.py
+++ b/wagtail/project_template/project_name/urls.py
@@ -16,14 +16,6 @@ urlpatterns = [
 
     url(r'^search/$', search_views.search, name='search'),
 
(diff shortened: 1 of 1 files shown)


gasman:
Thanks @nimasmi! Merged in 7acf4889dbbda55b726c6f8dd225d10e5583ba68.

Issue #5712, opened 2019-11-18
Replacing a document file changes the filename and therefore the URL of a document


johannesvogel:
### Issue Summary

When you replace an existing document with a new version of that document which has the **same** filename, the filename will be changed from `registration_form.pdf` to something like `registration_form_JQAOxrD.pdf`. This means that the old URL of that document is no longer valid and cannot be reached anymore. This is no problem internally within Wagtail but becomes an issue when the document URL is referenced somewhere else, e.g. a printout.

### Steps to Reproduce

1. Upload a file `example.pdf` in the Wagtail admin
2. Now click on the document to edit it
3. Change the document to another file with the same filename `example.pdf` and save it.
4. As can be seen in the document list then, both the filename and the URL have now changed to something like `example_JQAOxrD.pdf`

_Any other relevant information. For example, why do you consider this a bug and what did you expect to happen instead?_

I would expect the filename and therefore the URL to be stable unless I upload a file with a different filename, then it should change to that new filename.

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: **yes**

### Technical details

* Python version: `3.7.4`
* Django version: `2.2.7`
* Wagtail version: `2.7`
* Browser version: `Chromium 78`

This is happening because the new file is being saved to the filesystem _before_ the old file has been deleted. If Wagtail would delete the old file just before the new file is being saved, then the filename and URL would stay the same.

This is happening in `wagtail/documents/views/documents.py`. There is already a PR open that  gets us halfway to a fix: #5705. Based on this PR, one only needs to move this line (https://github.com/wagtail/wagtail/pull/5705/files#diff-72e95719bafa4483a89cbb3c62068344R156) to line 144 to fix this issue. I will open a PR for this later.

gasman:
Hi @johannesvogel - thanks for the report.

This behaviour was deliberately introduced in #3821 to avoid outdated versions of documents being cached. There's some strong feeling that this isn't the preferred behaviour, but if we change it then we need to come up with an alternative fix for caching. This is discussed in #4567, and the current status of this issue is summarised at https://github.com/wagtail/wagtail/pull/4567#issuecomment-531191805.

johannesvogel:
> This is discussed in #4567, and the current status of this issue is summarised at [#4567 (comment)](https://github.com/wagtail/wagtail/pull/4567#issuecomment-531191805).

Alright, thanks for the heads-up! And I thought I used all possible permutations to find if this issue does already exist 😉

PR #5736, opened 2019-12-17
Add ETag header when serving documents

johannesvogel:
This PR uses the `file_hash` in Documents to add an `ETag` Header to the HTTP Response:
```
ETag "14e0553f9be19646cf4aa7a5ea0396f1d12e3463"
```
It utilizes Django's [Conditional View Processing](https://docs.djangoproject.com/en/3.0/topics/conditional-view-processing/) to achieve this. If a client has a certain document already cached, it can check if the Document has changed by adding `If-None-Match` to the request header:
```
If-None-Match "14e0553f9be19646cf4aa7a5ea0396f1d12e3463"
```
If the document has changed, Wagtail will serve that new file. If the document has _not_ changed, Wagtail responds with `304 Not Modified`. Then the client's cached version is used. This functionality is implemented in all modern Browsers, including Internet Explorer. It is an efficient way of "cache busting" even when the URL stays the same. I implemented this to (at least partially) solve the problem described in #4567 and #5712.

### DIFF
--- a/wagtail/documents/tests/test_views.py
+++ b/wagtail/documents/tests/test_views.py
@@ -14,7 +14,7 @@ from wagtail.documents import models
 @override_settings(WAGTAILDOCS_SERVE_METHOD=None)
 class TestServeView(TestCase):
     def setUp(self):
-        self.document = models.Document(title="Test document")
+        self.document = models.Document(title="Test document", file_hash="123456")
         self.document.file.save('example.doc', ContentFile("A boring example document"))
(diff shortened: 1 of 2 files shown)


johannesvogel:
Sorry for the accidental force-push, I messed up there 😕

gasman:
Merged in 4a1b337461b68b355827ed6d0ca101a3a1ffb52d - thanks @johannesvogel!

PR #5745, opened 2019-12-27
Use get_admin_display_title in PageChooserPanel

### LINKED ISSUE #5726: PageChooserPanel displays selected page title instead of custom get_admin_display_title
### Issue Summary

PageChooserPanel displays selected page title instead of custom get_admin_display_title value. 
When I click to select a page, the chooser displays correctly the custom get_admin_display_title in the list of pages. 

When I select the page desired from the chooser, the chosen page title doesn't display the expected value of get_admin_display_title. Instead, it displays the page title which I think it's a bug. 

I had a look at wagtail/admin/templates/wagtailadmin/widgets/page_chooser.html and it should display the get_admin_display_title right?

### Steps to Reproduce

1. Have a page with a custom get_admin_display_title
2. Have another page with a PageChooserPanel
3. Use the PageChooserPanel in the admin to select the page with custom get_admin_display_title. Once selected, the title of the page displayed is the page title instead of get_admin_display_title value

### Technical details

* Python version:  3.7.3
* Django version: 2.2.7
* Wagtail version: 2.7
* Browser version: Chrome 78

SaptakS:
Should fix #5726 . @gasman This should fix the title of the page only in the PageChooserPanel. Do you think this is a good approach? Another approach might be to remove the `pageTitle.text()` entirely since by default `page_chooser.html`  shows `get_admin_display_title`. But I think adding a new data attribute is a better way.

I think I will need to add some tests also to ensure this behaviour works. I will add the tests once you confirm that this approach looks good.

- [X] Do the tests still pass? (https://docs.wagtail.io/en/latest/contributing/developing.html#testing)
- [X] Does the code comply with the style guide? (Run `make lint` from the Wagtail root)
- [ ] For Python changes: Have you added tests to cover the new/fixed behaviour?

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/js/page-chooser.js
+++ b/wagtail/admin/static_src/wagtailadmin/js/page-chooser.js
@@ -26,7 +26,7 @@ function createPageChooser(id, pageTypes, openAtParentId, canChooseRoot, userPer
                 pageChosen: function(pageData) {
                     input.val(pageData.id);
                     openAtParentId = pageData.parentId;
-                    pageTitle.text(pageData.title);
+                    pageTitle.text(pageData.adminTitle);
                     chooserElement.removeClass('blank');
                     editLink.attr('href', pageData.editUrl);
                 }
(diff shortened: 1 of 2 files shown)


gasman:
Thanks @SaptakS! Merged in 61e817362590c12ab6e9a2920cad23b93758a720 (with a bit of additional logic + test to ensure that the display title is also used on the initial render for an already-populated field).

PR #5750, opened 2020-01-02
Make room for footer on editor pages

### LINKED ISSUE #5610: Editing pages on mobile devices
### Problem...

The only time I have an issue using Wagtail is when I am editing content from a mobile phone.  The "Drop Up" menu for want of a better word covers the last item on the editable page, making it impossible to edit (at least on my iPhone Plus).

In the image below, you can see where it says "Save Draft", is actually covering the text entry field for the body, and it is impossible to scroll down any further.

### Possible solution...

I would like the menu with "Publish, Save Draft, Submit for moderation etc" to be at the bottom of the page, rather than floating on the bottom.  This would allow users to reach the last element on the page for editing.

### Alternatives that make this non-essential ...
One solution is to rotate the phone to landscape, however in the era of mobile first development this is annoying, and isn't a very elegant solution.

image

### LINKED ISSUE #5742: bottom bar in admin pages can obscure page content
### Issue Summary

Some content is either not accessible or visually cramped:

| small| medium |
|--|--|
|image|image|

### Steps to Reproduce

Make the browser window smaller

### Technical details

* Python version: 3.7
* Django version: 2.2.9
* Wagtail version: 2.7
* Browser version: Mac Chrome 79

### Possible solutions

* Add more padding to the bottom of all editor pages
* Make the bottom bar full-width panel, and have the editor page scroll area "end" above the bar

jckpn:
Fixes #5742, fixes #5610


lb-:
This should no longer need to be applied as the footer items are no longer sticky - see #5946

@jckpn - thanks for submitting a way to fix it though.

PR #5752, opened 2020-01-03
Fix documents tests on Windows

kaedroho:
As documents are streamed, the files are not closed until the datastream has been fully consumed by the test. But most of the documents tests never do this.

On Windows, files must be closed before deleting them. This change makes the tests always consume the file so that they are closed before deleting them.

Surprisingly, this seems to be the only test that fails on Windows! (on a basic SQLite setup)


gasman:
LGTM - thanks @kaedroho!

PR #5754, opened 2020-01-06
Be more specific regarding the cached loader.

minusf:
Be more explicit about not needing to enable the `cached.Loader` manually, unless using some custom loaders.

While an explicit setting is better than an implicit one, I still think the documentation should mention that this optimization is already there for free, unless a custom loader has to be used, in which case the developer will come accross that bit of django documentation anyway.

### DIFF
--- a/docs/advanced_topics/performance.rst
+++ b/docs/advanced_topics/performance.rst
@@ -47,7 +47,7 @@ Wagtail is tested on PostgreSQL, SQLite and MySQL. It should work on some third-
 Templates
 ---------
 
-The overhead from reading and compiling templates adds up. Django's :class:`cached template loader <django.template.loaders.cached.Loader>`: stores the compiled ``Template`` in memory and returns it for subsequent requests. Unless additional custom loaders are used, the cached loader is automatically enabled when ``DEBUG`` is False. To enable it for custom loaders, use a settings like this:
+The overhead from reading and compiling templates adds up. Django's :class:`cached template loader <django.template.loaders.cached.Loader>`: stores the compiled ``Template`` in memory and returns it for subsequent requests. Unless additional custom loaders are used, the cached loader is automatically enabled when ``DEBUG`` is ``False``. To enable it for custom loaders, use a settings like this:
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


docs/advanced_topics/performance.rst
```diff
@@ -47,7 +47,7 @@ Wagtail is tested on PostgreSQL, SQLite and MySQL. It should work on some third-
 Templates
 ---------
 
-The overhead from reading and compiling templates can add up. In some cases a significant performance improvement can be gained by using :class:`Django's cached template loader <django.template.loaders.cached.Loader>`:
+The overhead from reading and compiling templates adds up. Django's :class:`cached template loader <django.template.loaders.cached.Loader>`: stores the compiled ``Template`` in memory and returns it for subsequent requests. Unless additional custom loaders are used, the cached loader is automatically enabled when ``DEBUG`` is ``False``. To enable it for custom loaders, use a settings like this:
```
zerolab: How about?

```suggestion
The overhead from reading and compiling templates adds up. Django wraps its default loaders with :class:`cached template loader <django.template.loaders.cached.Loader>`: which stores the compiled ``Template`` in memory and returns it for subsequent requests. The cached loader is automatically enabled when ``DEBUG`` is ``False``. If you are using custom loaders, update your settings to use it:
```

docs/advanced_topics/performance.rst
```diff
@@ -59,13 +59,12 @@ The overhead from reading and compiling templates can add up. In some cases a si
                 ('django.template.loaders.cached.Loader', [
                     'django.template.loaders.filesystem.Loader',
                     'django.template.loaders.app_directories.Loader',
+                    'path.to.custom.Loader',
                 ]),
             ],
         },
     }]
 
-There is a caveat associated with this loader though. Changes to a template file will not be picked up once it is cached. This means that this loader should *not* be enabled during development.
```
zerolab: I think it is worth leaving this note since specifically wrapping the loader with the cached loader will result in the caveat it describes

minusf: i feel it's a bit of a tautology as it is explained well in the linked django docs, but i have no strong preference, dont mind if it stays.

zerolab: It is, however you would be surprised how many people will overlook that!

gasman:
Merged in fc0cf64ac6615905ee6207314b935df3e979f287 - thanks @minusf!

PR #5757, opened 2020-01-09
Fix update_index throwing AttributeError when value is None

### LINKED ISSUE #4822: `update_index` throws `AttributeError` on TableBlock when value is `None`

### Issue Summary

`update_index` fails on StreamField TableBlock when its value is `None`, due to this line: https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/table_block/blocks.py#L79

Throws `AttributeError: 'NoneType' object has no attribute 'get'`

### Steps to Reproduce

1. Add a StreamField to a page
2. Add a TableBlock and leave it empty
3. Publish the page
4. Run `python manage.py update_index`, you will get a traceback leading to the above `AttributeError`

image

### Technical details

Wagtail version: 2.2.2

2pxsolidblack:
This fixes #4822 
Also avoid rendering "None" if the table value is None.

I noticed that value is None only if you save the page without touching the block form.
Not sure if that's the correct behaviour.

### DIFF
--- a/wagtail/contrib/table_block/blocks.py
+++ b/wagtail/contrib/table_block/blocks.py
@@ -83,8 +83,9 @@ class TableBlock(FieldBlock):
 
     def get_searchable_content(self, value):
         content = []
-        for row in value.get('data', []):
-            content.extend([v for v in row if v])
+        if value:
+            for row in value.get('data', []):
+                content.extend([v for v in row if v])
         return content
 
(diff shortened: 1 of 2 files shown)


gasman:
Thanks @2pxsolidblack! Merged in 4864920555d06432ed3849e9ac32f4c4daa440c5 (with test added for the `get_searchable_content` fix).

PR #5758, opened 2020-01-10
Responsive embed setting

KalobTaulien:
Added a setting to disable the `style="padding-bottom: 56.25%" class="responsive-object"` attributes around an embed elements parent `<div>`. 

By default this is enabled bu can be disabled with `WAGTAIL_ENABLE_RESPONSIVE_EMBED = False`. 

Ran `runtests.py`, no errors. Updated docs, changelog and 2.8 release notes.


zerolab:
Merged in aefb0f9

gasman:
Based on the discussions on #1495 and among the core team, I think we need to be bolder here and have the responsive CSS disabled by default. As I see it, the primary complaint on #1495 is that the out-of-the-box rendering is broken until you locate the non-obvious bit of documentation that tells you how to fix it, and that's still the case now, only the documentation now offers two fixes rather than one...

KalobTaulien:
@gasman I agree. I've swapped out the logic to be more bold. Responsive embeds are disabled by default but can be turned back on. I also changed the setting name to `WAGTAILEMBEDS_RESPONSIVE_HTML` (based on @gasman's suggestion) — this name also seems more intuitive to me.

gasman:
Have opened #5769 to cover disabling this by default, along with some further tweaks to tests / docs.

PR #5759, opened 2020-01-10
dropdown on search page by initing dropdowns on ajax success

easherma:
Tested on Chrome and Firefox.

This is my first PR, so let me know if I'm missing anything! It's a pretty small change, however. 

This fixes an issue where dropdown menus don't work on the search page after you have made a search result. The dropdown components needed to be re-initialized.


zerolab:
Merged in f5f841e

PR #5762, opened 2020-01-13
Implement methods for synchronising content between two pages

kaedroho:
This PR implements to new methods on `Page`:

`.sync_fields_with(other)` copies field content from `other` to `self`
`.sync_child_relations_with(other)` copies child relations from `other` to `self`

They must be separate because `.sync_fields_with(other)` can be run on an unsaved page but `sync_child_relations_with(other)` can't.

I've updated the implementation of `.copy()` to use these methods internally.

We need these for `wagtail-localize`. They would be used for creating and synchronising "Placeholder pages" for languages that don't currently have full translations allowing sites to be translated incrementally. WIP PR for placeholder pages here: https://github.com/wagtail/wagtail-localize/pull/47

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -1063,14 +1063,27 @@ class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):
         # Log
         logger.info("Page moved: \"%s\" id=%d path=%s", self.title, self.id, new_url_path)
 
-    def copy(self, recursive=False, to=None, update_attrs=None, copy_revisions=True, keep_live=True, user=None, process_child_object=None, exclude_fields=None):
-        # Fill dict with self.specific values
-        specific_self = self.specific
-        default_exclude_fields = ['id', 'path', 'depth', 'numchild', 'url_path', 'path', 'index_entries', 'live', 'has_unpublished_changes', 'live_revision', 'first_published_at', 'last_published_at']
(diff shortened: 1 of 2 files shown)


wagtail/core/models.py
```diff
@@ -1063,14 +1063,27 @@ def move(self, target, pos=None):
         # Log
         logger.info("Page moved: \"%s\" id=%d path=%s", self.title, self.id, new_url_path)
 
-    def copy(self, recursive=False, to=None, update_attrs=None, copy_revisions=True, keep_live=True, user=None, process_child_object=None, exclude_fields=None):
-        # Fill dict with self.specific values
-        specific_self = self.specific
-        default_exclude_fields = ['id', 'path', 'depth', 'numchild', 'url_path', 'path', 'index_entries']
-        exclude_fields = default_exclude_fields + specific_self.exclude_fields_in_copy + (exclude_fields or [])
-        specific_dict = {}
+    def sync_fields_with(self, other, exclude_fields=None):
```
jacobtoppm: I think for more general usage, it might be nicer to refactor this so it takes a list of fields to sync (so you can sync a single field between pages, for example), rather than exclude - and make the `copy` method do the exclusion part. What do you think?

kaedroho:
This requirement has now be fulfilled by [Alias pages](https://github.com/wagtail/wagtail/pull/6376)

PR #5764, opened 2020-01-14
Redirect to previous url when deleting/copying/unpublish a page

### LINKED ISSUE #3010: Deleting page from usage view sends user back to wrong place

2pxsolidblack:
This fixes #3010 appending the `next` querystring parameter to the "Delete", "Copy" and "Unpublish" actions.

Waiting for feedbak,
Carlo

### DIFF
--- a/docs/reference/hooks.rst
+++ b/docs/reference/hooks.rst
@@ -830,7 +830,7 @@ Page explorer
     from wagtail.admin import widgets as wagtailadmin_widgets
 
     @hooks.register('register_page_listing_buttons')
-    def page_listing_buttons(page, page_perms, is_parent=False):
+    def page_listing_buttons(page, page_perms, is_parent=False, next_url=None):
         yield wagtailadmin_widgets.PageListingButton(
             'A page listing button',
             '/goes/to/a/url/',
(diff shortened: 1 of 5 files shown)


lb-:
@carloratm - is this your PR? the original PR is now being submitted from a deleted user.

If so, thanks so much for submitting this.

lb-:
Merged in via 5f3c12682fe5f93de6a0f614ab4af89d8518e807

kaedroho:
I've just unexpectedly hit this change on a site:

> TypeError: page_listing_more_buttons() takes from 2 to 3 positional arguments but 4 were given

We should at least add an upgrade consideration as this is a breaking change

lb-:
@kaedroho I did not think of that when merging this in, happy to do a PR for a release notes update to reflect this.

Or do you think we should pull in `**kwargs ` instead?

any thoughts @gasman  ?

gasman:
Weird, I thought I'd replied on this, unless there was another recent PR with a similar outcome...

This should definitely go through our two-release deprecation process. When calling a user hook function, we should make use of the `accepts_kwarg` helper in `wagtail.core.utils` to see if it expects a `next_url` kwarg, and if not, output a deprecation warning and call it without `next_url`. We can then remove that check in two releases' time.

lb-:
FYI - deprecation warnings added via https://github.com/wagtail/wagtail/pull/6218 & on issue #6208

Issue #5370, opened 2019-06-10
"Unpublished" styles in pages listings are very hard / impossible to see


thibaudcolas:
### Issue Summary

In listing pages, Wagtail has styles for pages that are not published:

<img width="1155" alt="unpublished-styles" src="https://user-images.githubusercontent.com/877585/59192318-f5a44d00-8b79-11e9-8ebc-cf965acbeb84.png">

These are very hard to see – I’m not sure what value they bring, whether they should be removed altogether or switch to a visual indicator that is less subtle, and doesn’t rely on color alone.

### Steps to Reproduce

1. Go to the CMS admin
2. Create a page as draft, without publishing
3. Go to a listing view showing that page

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: no

### Technical details

* Wagtail version: 2.5a0, 2.6a0
* Browser version: latest Chrome

cfarm:
Here are some options for restyling that:

# Italicize the draft page title
![Screen Shot 2019-07-26 at 4 26 36 PM](https://user-images.githubusercontent.com/702526/61979800-d8f59480-afc2-11e9-8e66-c3486909be05.png)

# Add a background that matches the "Draft" status badge to the whole row to make the status more obvious
![Screen Shot 2019-07-26 at 4 30 35 PM](https://user-images.githubusercontent.com/702526/61979798-d85cfe00-afc2-11e9-99a1-fc6a00e823f2.png)

# Add background and italicize the title

![Screen Shot 2019-07-26 at 4 29 56 PM](https://user-images.githubusercontent.com/702526/61979799-d85cfe00-afc2-11e9-8660-a48e9c8316ed.png)

thibaudcolas:
👍 I vote for "Italicize". It seems enough for the purpose of this issue!

saevarom:
As mentioned in #5776, after discussion at Bristol sprint, we agreed that having the whole row greyed out using the diagonal line background is more visually pleasing. The proposed solution in #5776 looks like this:

<img width="1013" alt="Screenshot 2020-01-23 at 12 30 59" src="https://user-images.githubusercontent.com/143557/72985955-08088980-3ddf-11ea-8fa4-9d334d2703ed.png">

thibaudcolas:
We’ve discussed this with @benenright and agreed to simply stop highlighting unpublished pages in a special way. The "draft" status tag makes it very clear which pages aren’t published, and we don’t think this needs further calling to attention.

[...]

PR #5776, opened 2020-01-23
Emphasizing unpublished pages. Issue #5370

saevarom:
Further work on issue #5370 after discussion at Bristol sprint. We agreed that having the whole row greyed out using the diagonal line background is more visually pleasing. After that change, we felt that reduced opacity on the title would not be necessary and probably not visible enough. I also discovered that the status tag of the page is probably not WCAG compliant since it is using a very light grey on top of white background. After discussion with @thibaudcolas we agreed that these status tags and their compliance should be looked into as a separate issue, since they are used in many places and the consequences of changing them are not obvious.

The final result looks like this:
Screenshot 2020-01-23 at 12 30 59

Tested in:
* Chrome Version 79.0.3945.117 (Official Build) (64-bit) in MacOS
* Safari version 13.0.2 (15608.2.30.1.1)
* Firefox developer edition version 72.0b9

### DIFF
--- a/client/scss/components/_listing.scss
+++ b/client/scss/components/_listing.scss
@@ -363,8 +363,14 @@ ul.listing {
         padding-left: 20px;
     }
 
-    .unpublished .title-wrapper {
-        opacity: 0.7;
+    .unpublished {
+        background: #fff url('#{$images-root}bg-dark-diag.svg');
+
+        .status-tag {
+            color: $color-grey-2;
+            border: 1px solid $color-grey-2;
+            background: #fff;
+        }
     }
 
     .index {
--- a/client/scss/components/_listing.scss
+++ b/client/scss/components/_listing.scss
@@ -95,6 +95,10 @@ ul.listing {
                 border-top: 1px dashed $color-input-border;
             }
 
+            &:last-child {
+                border-bottom: 1px dashed $color-input-border;
+            }
+
             &:hover {
                 background-color: #fcfcfc;
             }
@@ -548,7 +552,6 @@ table.listing {
 
 .listing.full-width + .pagination {
     margin-top: 3em;
-    border-top: 1px dashed #d9d9d9;
     padding: 2em 50px 0;
 }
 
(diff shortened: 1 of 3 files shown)


saevarom:
@tomdyson spotted this gap when the last item in a list is in draft state:

Screenshot 2020-01-23 at 14 37 04

This is because the bottom border is part of the navigation element and the margins are calculated in relative units - resulting in different actual pixel values.

tomdyson:
Looks good! Do you think we should use the same styling on the dashboard listing, for consistency? 

image

saevarom:
Good point, that would make sense for consistency throughout the UI.

saevarom:
I see - well the reason I chose to change how borders was done was two-fold. First, some explanation: The spacing issue is caused by a relative margin of 3em on the pagination below the table, which is then negated with a -3em margin on the table above. I looks like these elements are calculating their pixel values relevant to different parent elements, resulting in a very slight pixel difference between those +3em and -3em values. This would be relatively easy to fix using rem units instead of em units, or a fixed px value. However, I thought it would be better for the table itself to define its own borders, instead of relying on a sibling element providing a border which should of course belong to the table. This also fixes another problem I found in the sorting view. The sorting view does not have a pagination sibling element and therefore doesn't have this border:

Screenshot 2020-01-27 at 09 34 08

So the reason I went this way was:

- A border of a table should belong to the table, not a different element which might or might not be present. 
- Relative units can cause issues, and if the border is moved over to the table, this is not an issue anymore.

So

saevarom:
I guess we should decide whether we try to catch all cases of double borders in lists of things in this PR, or simply revert the border fix and go for `rem` or `px` values in the margins.

saevarom:
Any opinion on which way to go @thibaudcolas ?

thibaudcolas:
I’ve finally been able to discuss this with @benenright, and we’d like to go with a much simpler approach (see [5370](https://github.com/wagtail/wagtail/wiki/Sustainability-team)). Sorry this has taken so long @saevarom 🙈

saevarom:
Oh come on @thibaudcolas, it's only been 4 years!

PR #5779, opened 2020-01-23
Accessibility improvements for react page chooser

kaedroho:
This PR implements accessibility fixes for the React page chooser.

Based on https://github.com/wagtail/wagtail/pull/3038


client/src/components/choosers/page/PageChooserResultSet.js
```diff
@@ -0,0 +1,267 @@
...
+        onChoose={onChoose}
+        onNavigate={handleNavigate}
+        pageTypes={pageTypes}
+      />
+    );
+  }
+
+  return (
+    <div className="page-results">
+      <table className="listing  chooser">
```
thibaudcolas: This would need a visually hidden `<caption>` for screen reader users.

client/src/components/choosers/page/PageChooserResult.js
```diff
@@ -0,0 +1,197 @@
...
+  renderUpdatedAt() {
+    const { page } = this.props;
+
+    if (page.meta.latest_revision_created_at) {
+      const updatedAt = moment(page.meta.latest_revision_created_at);
+
+      return (
+        <td className="updated u-vertical-align-top">
+          <div className="human-readable-date" title={updatedAt.format('D MMM YYYY h:mm a')}>
+            {updatedAt.fromNow()}
```
thibaudcolas: It feels quite expensive to add `moment` as a dependency just for that. Is the intention to use it elsewhere? Otherwise I’d suggest either doing the formatting with vanilla JS, or using `date-fns` or `luxon`, or adding a formatted version of the date field(s) to the API response.

kaedroho: I think, ideally, we would make use of Django's builtin date formats, not sure if we have a JS library for this yet?

kaedroho: Actually, `fromNow` would be a tricky one to implement with that. I wonder if we could compile `moment` in such a way that it'll only fetch the locale it needs?

thibaudcolas: It looks like there is now a way to do this in the standard library, as of ES2020: https://github.com/tc39/proposal-intl-relative-time.

kaedroho:
Abandoning this as https://github.com/wagtail/wagtail/pull/3038 was abandoned

PR #5780, opened 2020-01-23
Add basic skiplink

### LINKED ISSUE #5406: Implement a skip-link in the CMS admin so users can skip repetitive content
We need to implement a [skip link](https://en.wikipedia.org/wiki/Skip_link) in the CMS admin, so keyboard users can skip navigation menus and go straight to the main content of the page. This is useful for keyboard users who do not rely on a screen reader – screen reader users can already move around with headings and landmarks.

Here is a good example of a skip link:

skip-link

I think we will want Wagtail’s skip link to have the following characteristics:

- Only becomes visible when the user starts tabbing through the page (`visuallyhidden`, and focusable)
- Should take the user to the `main` landmark
- Should not add a fragment identifier / anchor in the URL bar
- Basically state of the art

martincoote:
Closes #5406

Tested in:

- Safari MacOS 13.0.4
- Firefox MacOS 72.0.2
- Chrome MacOS 79 all using Voiceover
- Mobile Chrome 79 on Android using TalkBack screenreader
- IE11 on windows 10 via browserstack (no screen reader testing)

Implement a JS powered skip link inspired by this post https://axesslab.com/skip-links/ and specifically this update https://axesslab.com/skip-links#update-3-a-comment-from-gov-uk.

The functionality works as follows. 
1. When the skiplink button is clicked the main element gets the attribute `tabindex="-1"` 
2. on blur or focus out the tabindex is removed so it no longer interferes with the tabbing order.

### DIFF
--- a/client/scss/settings/_variables.scss
+++ b/client/scss/settings/_variables.scss
@@ -39,6 +39,7 @@ $color-orange: #e9b04d;
 $color-orange-dark: #bb5b03;
 $color-green: #189370;
 $color-green-dark: #157b57;
+$color-green-darker: #105d42; // White has AAA compatibility when this is used in bgd.
 $color-salmon: #f37e77;
 $color-salmon-light: #fcf2f2;
 $color-white: #fff;
--- a/client/src/includes/initSkipLink.js
+++ b/client/src/includes/initSkipLink.js
@@ -1,0 +1,23 @@
+const initSkipLink = () => {
+    // Inspired by https://github.com/selfthinker/dokuwiki_template_writr/blob/master/js/skip-link-focus-fix.js
+
+    const skiplink = document.querySelector('.skiplink');
+    const main = document.querySelector('main');
+
+    const handleClick = () => {
+        main.setAttribute('tabindex', -1);
+        main.addEventListener('blur', handleBlur);
+        main.addEventListener('focusout', handleBlur);
+        main.focus();
+    };
+
+    const handleBlur = () => {
+        main.removeAttribute('tabindex');
+        main.removeEventListener('blur', handleBlur);
+        main.removeEventListener('focusout', handleBlur);
+    };
+
+    skiplink.addEventListener('click', handleClick);
+};
+
+export { initSkipLink };
--- a/client/src/includes/initSkipLink.js
+++ b/client/src/includes/initSkipLink.js
@@ -1,23 +1,23 @@
 const initSkipLink = () => {
-    // Inspired by https://github.com/selfthinker/dokuwiki_template_writr/blob/master/js/skip-link-focus-fix.js
+  // Inspired by https://github.com/selfthinker/dokuwiki_template_writr/blob/master/js/skip-link-focus-fix.js
 
-    const skiplink = document.querySelector('.skiplink');
-    const main = document.querySelector('main');
+  const skiplink = document.querySelector('.skiplink');
+  const main = document.querySelector('main');
 
-    const handleClick = () => {
-        main.setAttribute('tabindex', -1);
-        main.addEventListener('blur', handleBlur);
-        main.addEventListener('focusout', handleBlur);
-        main.focus();
-    };
+  const handleBlur = () => {
+    main.removeAttribute('tabindex');
+    main.removeEventListener('blur', handleBlur);
+    main.removeEventListener('focusout', handleBlur);
+  };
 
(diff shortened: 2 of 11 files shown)


lb-:
@martincoote thanks for your work on this, I was able to use this successfully in Chrome but could not use the skip link at all on Firefox (76.0.1 on MacOS 10.15.4). Maybe @thibaudcolas  or someone else can check on Firefox as it could be something wrong with my mac?

Also, when you do re refresh, the 'skip to content' link appears for a few ms as the element is rendered above some scripts.

It appears to be better when the link is inside the .wrapper - I think this will be ok as it is still the first focusable element in the body.


```
    <div class="wrapper">
        <a class="skiplink button" href="#main">Skip to main content</a>
       {% block furniture %}{% endblock %}
    </div>
```

Also, do you mind rebasing this on the latest master to resolve the code conflict.

lb-:
Also, just though, we probably should use translations, so the updated code would be

```
    <div class="wrapper">
        <a class="skiplink button" href="#main">{% trans 'Skip to main content' %}</a>
       {% block furniture %}{% endblock %}
    </div>
```

lb-:
I found part of the reason why tabbing was not working as expected on Firefox
https://stackoverflow.com/questions/11704828/how-to-allow-keyboard-focus-of-links-in-firefox
https://bugzilla.mozilla.org/show_bug.cgi?id=1436136

I am not sure what else we can do about this, maybe it is fair to assume users will have this turned on.

client/scss/settings/_variables.scss
```diff
@@ -39,6 +39,7 @@ $color-orange: #e9b04d;
 $color-orange-dark: #bb5b03;
 $color-green: #189370;
 $color-green-dark: #157b57;
+$color-green-darker: #105d42; // White has AAA compatibility when this is used in bgd.
```
thibaudcolas, line 42: Good call!

client/src/includes/initSkipLink.js
```diff
@@ -0,0 +1,23 @@
+const initSkipLink = () => {
+  // Inspired by https://github.com/selfthinker/dokuwiki_template_writr/blob/master/js/skip-link-focus-fix.js
+
+  const skiplink = document.querySelector('.skiplink');
```
thibaudcolas, line 4: Ideally this would be done based on a data attribute, `[data-skiplink]`.

client/src/includes/initSkipLink.js
```diff
@@ -0,0 +1,23 @@
...
+  const main = document.querySelector('main');
+
+  const handleBlur = () => {
+    main.removeAttribute('tabindex');
+    main.removeEventListener('blur', handleBlur);
+    main.removeEventListener('focusout', handleBlur);
```
thibaudcolas, line 10: Why do we need both blur and focusout?

client/src/includes/initSkipLink.js
```diff
@@ -0,0 +1,23 @@
...
+  };
+
+  const handleClick = () => {
+    main.setAttribute('tabindex', -1);
+    main.addEventListener('blur', handleBlur);
+    main.addEventListener('focusout', handleBlur);
+    main.focus();
+  };
+
+  if (skiplink) skiplink.addEventListener('click', handleClick);
```
thibaudcolas, line 20:
```suggestion
  if (skiplink) {
    skiplink.addEventListener('click', handleClick);
  }
```

thibaudcolas: Do we need to have a `&& main` in here? Or overkill?

wagtail/admin/templates/wagtailadmin/skeleton.html
```diff
@@ -19,6 +19,7 @@
     {% block branding_favicon %}{% endblock %}
 </head>
 <body id="wagtail" class="{% block bodyclass %}{% endblock %} {% if messages %}has-messages{% endif %} focus-outline-on">
+    <a class="skiplink button" href="#main">Skip to main content</a>
```
thibaudcolas, line 22: As mentioned by LB this should be:

```suggestion
    <a class="skiplink button" href="#main">{% trans 'Skip to main content' %}</a>
```

…and I haven’t tried but yep would expect this to look much better on page reloads if it was inside `wrapper`, as LB pointed out in his comment. Wagtail has a lot of synchronous JS loading, so it does make a big difference for this to be done before.

Issue #5670, opened 2019-10-31
Add a setting to disable preview / view live functionality


bsnelder:
### Problem
I want to run Wagtail as a headless CMS. In my example, I only want to use Wagtail for a mobile app. This means I don't have a frontend to use the preview functionality for pages.

### Proposed solution
Introduce a new setting `WAGTAIL_HEADLESS_DISABLE_PREVIEW` that would disable the Preview / View live buttons for pages when set to `True`.

For now, there are alternatives like serving a page with some content saying Preview is not supported but more user-friendly would be to not show the buttons at all. Another option is to override some admin templates but this is also not ideal.

kaedroho:
I would prefer to use the name `WAGTAILADMIN_DISABLE_PREVIEW` for consistency with other wagtail admin settings.

gasman:
I'd be keen to look at alternatives to adding a new setting - if Wagtail isn't catering for a particular use-case then we should improve the design of Wagtail to accommodate it, not just add a new setting and conditional on top and call it done.

As @zerolab pointed out on Slack, it would make logical sense that setting `preview_modes` to an empty list should disable previewing - it looks like it's only failing because of the template-level logic. My preference would be to fix that behaviour.

bsnelder:
> I'd be keen to look at alternatives to adding a new setting - if Wagtail isn't catering for a particular use-case then we should improve the design of Wagtail to accommodate it, not just add a new setting and conditional on top and call it done.
> 
> As @zerolab pointed out on Slack, it would make logical sense that setting `preview_modes` to an empty list should disable previewing - it looks like it's only failing because of the template-level logic. My preference would be to fix that behaviour.

Fixing the behavior of setting `preview_modes` to an empty list also makes sense. It then should not display a preview button at all. This is the first thing I tried before even asking the question on slack.

gasman:
Fixed in #5781

thibaudmartinez:
Hi there,

I am currently using Wagtail nightly `2.9.dev20200418` as I wanted to use the "disable preview" feature.

The live button is properly hidden, but I ran into the following problem.

Let's say I define the model SomePage for which I disable the preview.

````python
class SomePage(Page):
    preview_modes = []
````

I create a page from this model and save it has a draft. When I go to the admin home page and click on the button "DRAFT" of the row corresponding to the created page I get the following error.

````python
Internal Server Error: /admin/pages/4/view_draft/
Traceback (most recent call last):
[...]

PR #5781, opened 2020-01-23
Hide preview button when preview_modes is empty

caspertimmers:
Resolves https://github.com/wagtail/wagtail/issues/5670

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/pages/create.html
+++ b/wagtail/admin/templates/wagtailadmin/pages/create.html
@@ -32,24 +32,26 @@
                       </div>
                   </li>
 
-                    <li class="preview">
-                        {% trans 'Preview' as preview_label %}
-                        {% if preview_modes|length > 1 %}
-                            <div class="dropdown dropup dropdown-button match-width">
+                    {% if preview_modes %}
+                        <li class="preview">
+                            {% trans 'Preview' as preview_label %}
+                            {% if preview_modes|length > 1 %}
+                                <div class="dropdown dropup dropdown-button match-width">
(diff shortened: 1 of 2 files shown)


gasman:
Thanks @caspertimmers! Please could you add tests for this?

saeed617:
I think a variable in settings is a better solution as suggested in #5670. this variable Indicates that this library is intended to be used as a headless cms, otherwise we should change `Page` class `DEFAULT_PREVIEW_MODES` or override each page  `preview_modes` method.

```python
class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):

    ...

    DEFAULT_PREVIEW_MODES = [('', _('Default'))]

    @property
    def preview_modes(self):
        """
        A list of (internal_name, display_name) tuples for the modes in which
        this page can be displayed for preview/moderation purposes. Ordinarily a page
        will only have one display mode, but subclasses of Page can override this -
        for example, a page containing a form might have a default view of the form,
        and a post-submission 'thankyou' page
        """
        return Page.DEFAULT_PREVIEW_MODES
```

gasman:
Merged (with tests added) in 19f50c72ce4351b70941cd8611f6302bbb158c7f - thanks @caspertimmers!

@saeed617 As mentioned on #5670, I'm keen to avoid introducing new settings variables where a logical alternative exists. Adding a setting is a quick fix, but it means we miss out on a chance to think about how we can improve the design of Wagtail - if we take that route too often, it means the developer experience eventually becomes "read through a mountain of reference documentation, and hope that you can find the right combination of flags to do what you want".

PR #5783, opened 2020-01-23
Aria label added to reset password link

martincoote:
Add aria-label to login page reset password link.


wagtail/admin/templates/wagtailadmin/login.html
```diff
@@ -82,7 +82,7 @@ <h1>{% block branding_login %}{% trans "Sign in to Wagtail" %}{% endblock %}</h1
             </ul>
 
             {% if show_password_reset %}
-                <p class="help"><a href="{% url 'wagtailadmin_password_reset' %}">{% trans "Forgotten it?" %}</a></p>
+                <p class="help"><a href="{% url 'wagtailadmin_password_reset' %}" aria-label="{% trans "Forgotten your password? Request a reset email." %} ">{% trans "Forgotten it?" %}</a></p>
```
thibaudcolas: I think the general recommendation is for the link text to match the linked page’s title – so in this case `Reset your password`. This also has the advantage of already being translated, so not needing extra work for translators to add it. Do you mind switching to this @martincoote?

gasman: @thibaudcolas  Just so you're aware, @caspertimmers has already done this in #5784 :-)

martincoote: Yep, we've totally duped effort here. I've made a comment on @caspertimmers PR to move the `aria-label` onto the `a` tag not the `p` tag. If that is done I'm happy to close this PR in favour of Caspers.

martincoote:
Duplicated effort closed in favour of https://github.com/wagtail/wagtail/pull/5784

wagtail/admin/templates/wagtailadmin/login.html
thibaudcolas: Phew!

Issue #5782, opened 2020-01-23
Password reset link needs a more descriptive label for screen reader users


thibaudcolas:
Found a bug? Please fill out the sections below. 👍

### Issue Summary

As a screen reader user, it’s not clear what the "Forgotten it?" link does. The link text should convey what the link goes to without any extra context, since blind users cannot see the link is visually next to the "Password" field.

<img width="829" alt="Screenshot 2020-01-23 at 16 50 47" src="https://user-images.githubusercontent.com/877585/73006015-d2c16300-3e01-11ea-8d73-3fa0422ec9f5.png">

### Steps to Reproduce

1. Head to the login screen with a screen reader
2. Move to the link
3. See the announced label does not make it clear that this is a **password reset** link

### Technical details

* Wagtail version: latest

---

The fix for this should be to add an `aria-label` attribute on this link, with more descriptive text, e.g. `Reset your password`. Ideally reuse a similar label from elsewhere in Wagtail if possible, so we don’t have one extra label to translate.

thibaudcolas:
Thank you @caspertimmers and @martincoote!

PR #5784, opened 2020-01-24
Add aria label for password reset

caspertimmers:
Resolves https://github.com/wagtail/wagtail/issues/5782

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/login.html
+++ b/wagtail/admin/templates/wagtailadmin/login.html
@@ -82,7 +82,7 @@
             </ul>
 
             {% if show_password_reset %}
-                <p class="help"><a href="{% url 'wagtailadmin_password_reset' %}">{% trans "Forgotten it?" %}</a></p>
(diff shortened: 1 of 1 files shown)


caspertimmers:
The translation of "Reset your password" does already exist, so this does not need new translations.

wagtail/admin/templates/wagtailadmin/login.html
```diff
@@ -82,7 +82,7 @@ <h1>{% block branding_login %}{% trans "Sign in to Wagtail" %}{% endblock %}</h1
             </ul>
 
             {% if show_password_reset %}
-                <p class="help"><a href="{% url 'wagtailadmin_password_reset' %}">{% trans "Forgotten it?" %}</a></p>
+                <p class="help" aria-label="{% trans 'Reset your password' %}"><a href="{% url 'wagtailadmin_password_reset' %}">{% trans "Forgotten it?" %}</a></p>
```
martincoote: The aria-label should be on the a tag, as that is what will be focused by the screen reader.

Issue #5410, opened 2019-06-24
Page reordering UI isn’t accessible for keyboard users


thibaudcolas:
In Wagtail’s admin, child pages can be reorder with a togglable drag'n'drop UI:

![drag-n-drop-sort](https://user-images.githubusercontent.com/877585/60035150-0f7a7f80-96a4-11e9-91c3-0deb3c470c55.png)

This UI only supports drag'n'drop, thus being completely unusable for keyboard users (regardless of whether they rely on a screen reader or not). I would also suspect it’s problematic / impossible to use for people who rely on the mouse but have fine motor skills issues – e.g. Parkinson’s, as the drag'n'drop interaction has a small "drag handle" target, and requires very precise continuous movement to reach the desired "drop area" target.

---

Having discussed this issue with multiple people, it seems like the general consensus is that this feature could use a complete redesign instead of just being reworked to support other means of interaction.

thibaudcolas:
Interesting upcoming WCAG 2.2 success criterion relating to UIs relying on drag-and-drop, for future reference: https://github.com/w3c/wcag/pull/1113.

lb-:
**Additional links**

* https://www.w3.org/WAI/WCAG22/Understanding/dragging-movements
* https://medium.com/salesforce-ux/4-major-patterns-for-accessible-drag-and-drop-1d43f64ebf09
* https://www.w3.org/wiki/PF/ARIA/BestPractices/DragDrop
* https://www.w3.org/TR/aria-state/#dragdrop

lb-:
https://github.com/wagtail/wagtail/pull/8972 will be a solid improvement on this, will consider this closed once that is merged.

@thibaudcolas  it might be worth adding a future issue for drag & drop across Wagtail, we can do this better but as noted above it will need a re-think and re-build. I think the original intent of this issue will be resolved for now though.

I am hopeful that https://github.com/wagtail/rfcs/pull/78 will give us a good baseline for UI interaction like this, the current jQuery approach is very difficult to work with and no matter what (even without Stimulus) we will need to explore and find a suitable DND library.

When researching sorting for the original RFC work (inline panel POC), the only maintained library out there seemed to be 

https://github.com/SortableJS/Sortable 

Even React DND (which would require a huge amount of extra work to adopt) has not had an update in over a year.


https://www.npmjs.com/search?ranking=maintenance&q=drag

https://www.npmjs.com/search?ranking=maintenance&q=sortable

PR #5785, opened 2020-01-24
 Add functionality to reorder pages with keyboard

Thomasvdh92:
https://github.com/wagtail/wagtail/issues/5410

Added functionality to rearrange pages in the wagtail admin for keyboard-only users. To do this, you have to _tab_ through the elements after enabling ordering. Then select an element in the drop and drag area by pressing the enter key. You then selected the object and can use the up and down arrow keys to rearrange them. You can press the up or down arrow keys multiple times so you can place a page n amount of times higher or lower. Then press enter again to confirm and the page will be set accordingly in the list. You can deselect an object by pressing the escape key. 

This functionality uses the same backend call as drag and drop does.

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/pages/listing/_list.html
+++ b/wagtail/admin/templates/wagtailadmin/pages/listing/_list.html
@@ -48,7 +48,7 @@
                 {% page_permissions page as page_perms %}
                 <tr {% if ordering == "ord" %}id="page_{{ page.id|unlocalize }}" data-page-title="{{ page.get_admin_display_title }}"{% endif %} class="{% if not page.live %}unpublished{% endif %} {% block page_row_classname %}{% endblock %}">
                     {% if show_ordering_column %}
-                        <td class="ord">{% if orderable and ordering == "ord" %}<div class="handle icon icon-grip text-replace">{% trans 'Drag' %}</div>{% endif %}</td>
+                        <td class="ord">{% if orderable and ordering == "ord" %}<div tabindex="0" id="orderer" class="handle icon icon-grip text-replace">{% trans 'Drag' %}</div>{% endif %}</td>
                     {% endif %}
                     <td class="title" valign="top" data-listing-page-title>
                         {% block page_title %}
--- a/wagtail/admin/templates/wagtailadmin/pages/index.html
+++ b/wagtail/admin/templates/wagtailadmin/pages/index.html
@@ -33,6 +33,73 @@
     <script type="text/javascript">
         {% if ordering == 'ord' %}
             $(function() {
+                var currentlySelected;
+                var currentPosition;
+                var movedPageId;
+                var moveNTimesDirection = 0;
+                $("*[id*=orderer]").on("keydown", function(e) {
+                    let keyCodes = {
+                        enter: 13,
+                        upArrow: 38,
+                        downArrow: 40,
+                        escape: 27
+                    }
+
+                    if (currentlySelected) {
+                        // We want to prevent default key actions (like scrolling) when we have an object selected
+                        e.preventDefault();
+                    }
+
+                    var children = $('.listing tbody').children();
+
+                    let moveElement = function() {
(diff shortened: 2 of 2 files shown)


Thomasvdh92:
resolves #5410

wagtail/admin/templates/wagtailadmin/pages/listing/_list.html
```diff
@@ -48,7 +48,7 @@
                 {% page_permissions page as page_perms %}
                 <tr {% if ordering == "ord" %}id="page_{{ page.id|unlocalize }}" data-page-title="{{ page.get_admin_display_title }}"{% endif %} class="{% if not page.live %}unpublished{% endif %} {% block page_row_classname %}{% endblock %}">
                     {% if show_ordering_column %}
-                        <td class="ord">{% if orderable and ordering == "ord" %}<div class="handle icon icon-grip text-replace">{% trans 'Drag' %}</div>{% endif %}</td>
+                        <td class="ord">{% if orderable and ordering == "ord" %}<div tabindex="0" id="orderer" class="handle icon icon-grip text-replace">{% trans 'Drag' %}</div>{% endif %}</td>
```
zerolab: Multiple elements with the same id are not valid HTML. Could this not be replaced by a class?

thibaudcolas: `data-` attribute if this is only used for JS interaction – 
```suggestion
                        <td class="ord">{% if orderable and ordering == "ord" %}<div tabindex="0" data-orderer class="handle icon icon-grip text-replace">{% trans 'Drag' %}</div>{% endif %}</td>
```

lb-:
@PaarthAgarwal if you're up for it. Might be worth resurrecting this PR - either as a rebase or just apply the changes on main as a new branch. 

Looks like it was close and a good addition. 

I've added to the UX Unification project.

PaarthAgarwal:
> @PaarthAgarwal if you're up for it. Might be worth resurrecting this PR - either as a rebase or just apply the changes on main as a new branch.
> 
> Looks like it was close and a good addition.
> 
> I've added to the UX Unification project.

Sure. You've been working yourself on adding keyboed shortcuts. Does this PR's way of doing it goes along with yours?

lb-:
Good thinking @PaarthAgarwal  - however this keyboard behaviour is more about 'control' of the interaction and not really global keyboard shortcuts.

Have a read, in full, of the issue https://github.com/wagtail/wagtail/issues/5410 to understand the goals here.

wagtail/admin/templates/wagtailadmin/pages/listing/_list.html
lb-, line 51: @PaarthAgarwal  what if this was a `button` with `type="button"` then we do not need make it an accessible target with tabindex.

Yeah the order thing will need to be fixed but the data-attribute suggestion above makes sense.

lb-:
@PaarthAgarwal  to get this over the line we would also need to tell screen readers that this can work with the keyboard.

A simple way to do this is a message content that is in the DOM near the top of the table that explains 'focus on blah, press enter to activate, up / down'.

PaarthAgarwal:
> @PaarthAgarwal to get this over the line we would also need to tell screen readers that this can work with the keyboard.
> 
> A simple way to do this is a message content that is in the DOM near the top of the table that explains 'focus on blah, press enter to activate, up / down'.

Should I add a screen reader only message?

lb-:
Yes please Paarth, here is a rough version of the text.

> Press 'enter' to select an item, use up and down arrows to move the item, press 'enter' to deselect the item.

Note: Check that matches the actual behaviour. We can refine further on the PR when up.

lb-:
@Thomasvdh92 thanks for this PR - I know it has been a while but Paarth has taken this on board, finished it off and rebuilt via #8972 so I will close this now.

We really appreciate the original work done here and will add you to the contributors list for this feature.

PR #5786, opened 2020-01-24
Admin landmark fixes

### LINKED ISSUE #5411: Rework the admin UI’s landmark based on established best practices
As part of #5275, we’ve introduced more ARIA landmarks across the CMS admin (`<main>`, and more). Axe is now reporting issues with those new landmarks:

- [ ] Banner landmark must not be contained in another landmark (https://dequeuniversity.com/rules/axe/3.2/landmark-banner-is-top-level?application=axeAPI)
- [ ] Contentinfo landmark must not be contained in another landmark (https://dequeuniversity.com/rules/axe/3.2/landmark-contentinfo-is-top-level?application=axeAPI)
- [ ] (Additionally, it is also reporting a missing `main` landmark for the 404 page)

These first 2 warnings are because the `main` wraps both the page’s main content, as well as areas designated as header and footer:

landmarks

I’m not entirely sure which course to follow here – whether `main` should be right after `header`, and `footer` right after `main`, or whether we would want to remove some or all of those landmarks. Thoughts welcome! Tagging this as "enhancement" because Axe classifies this as a best practice rather than an accessibility compliance failure, and I’m not entirely sure of the impact on screen readers myself.

Relevant: #5275, #5406.

mixxorz:
Fixes #5411 

* Remove inappropriate `banner` roles
* Remove inappropriate `contentinfo` roles
* Add `main` tag to 404 page


lb-:
@mixxorz thanks for preparing these fixes!
@thibaudcolas as mentioned, this is approved and good to merge so I have merged in via d0a0c804c25459bbf2161c5c53f10c37ba5b6b77

I also did a quick check locally in Firefox accessibility tools and all appears to work as described.

PR #5788, opened 2020-01-24
Replace welcome page for Squash instances

tomdyson:
Update the default welcome page from `wagtail start`, for ephemeral Squash instances only.


gasman:
Thanks @tomdyson - looks good to me!

PR #5789, opened 2020-01-24
Default custom image & document model

jacobtoppm:
Addresses https://github.com/wagtail/rfcs/pull/44

This adds a default custom image, document, rendition model in the project template. To avoid reverse accessor conflicts, the abstract base classes have their related_names set to '+', and set explicitly on their children.

After discussion with @gasman about only being used as a fallback, default_alt_text is added to the new image model.

All tests/linting passing.


docs/advanced_topics/images/custom_image_model.rst
```diff
@@ -7,7 +7,10 @@ Custom image models
 The ``Image`` model can be customised, allowing additional fields to be added
 to images.
 
-To do this, you need to add two models to your project:
+For projects started from Wagtail 2.9, customisable ``Image`` and ``Rendition`` models are located in the ``media`` app
```
zerolab: Would be good to add a similar note for [custom document models](https://docs.wagtail.io/en/stable/advanced_topics/documents/custom_document_model.html)

docs/reference/contrib/modeladmin/indexview.rst
```diff
@@ -617,7 +617,7 @@ kind of interactivity using JavaScript:
 ``wagtail.contrib.modeladmin.mixins.ThumbnailMixin``
 ----------------------------------------------------
 
-If you're using ``wagtailimages.Image`` to define an image for each item in
+If you're using ``media.Image`` to define an image for each item in
```
ababic: Seeing as not all projects use the project template, I feel it might be better to say something like "If you're using `wagtailimages.Image` (or a custom Image model)" here instead.

docs/reference/contrib/modeladmin/indexview.rst
```diff
@@ -631,7 +631,7 @@ change a few attributes to change the thumbnail to your liking, like so:
 
     class Person(models.Model):
         name = models.CharField(max_length=255)
-        avatar = models.ForeignKey('wagtailimages.Image', on_delete=models.SET_NULL, null=True)
+        avatar = models.ForeignKey('media.Image', on_delete=models.SET_NULL, null=True)
```
ababic: It might be better to encourage use of `wagtail.images.get_image_model_string()` here instead

docs/reference/contrib/modeladmin/indexview.rst
```diff
@@ -646,7 +646,7 @@ change a few attributes to change the thumbnail to your liking, like so:
 
         """
         Set 'thumb_image_field_name' to the name of the ForeignKey field that
-        links to 'wagtailimages.Image'
+        links to 'media.Image'
```
ababic: Might be better as: "links to 'wagtailimages.Image' (or a custom Image model)"

docs/reference/pages/panels.rst
```diff
@@ -179,7 +179,7 @@ ImageChooserPanel
 
       class BookPage(Page):
           cover = models.ForeignKey(
-              'wagtailimages.Image',
+              'media.Image',
```
ababic: It might be better to encourage use of `wagtail.images.get_image_model_string()` here instead

docs/topics/pages.rst
```diff
@@ -40,7 +40,7 @@ This example represents a typical blog post:
         body = RichTextField()
         date = models.DateField("Post date")
         feed_image = models.ForeignKey(
-            'wagtailimages.Image',
+            'media.Image',
```
ababic: It might be better to encourage use of `wagtail.images.get_image_model_string()` here instead

wagtail/api/v2/serializers.py
```diff
@@ -16,7 +16,7 @@ class TypeField(Field):
     Serializes the "type" field of each object.
 
     Example:
-    "type": "wagtailimages.Image"
+    "type": "media.Image"
```
ababic: Since `wagtailimages.Image` is still the default image model unless using the project template (or overriding in a custom project template), I think it still makes sense to use that here.

wagtail/api/v2/serializers.py
```diff
@@ -95,7 +95,7 @@ class RelatedField(relations.RelatedField):
     "feed_image": {
         "id": 1,
         "meta": {
-            "type": "wagtailimages.Image",
+            "type": "media.Image",
```
ababic: I would suggest leaving this as is

wagtail/api/v2/serializers.py
```diff
@@ -153,7 +153,7 @@ class ChildRelationField(Field):
             "image": {
                 "id": 1,
                 "meta": {
-                    "type": "wagtailimages.Image",
+                    "type": "media.Image",
```
ababic: I would suggest leaving this as is

wagtail/contrib/modeladmin/mixins.py
```diff
@@ -9,7 +9,7 @@ class ThumbnailMixin:
     """
     Mixin class to help display thumbnail images in ModelAdmin listing results.
     `thumb_image_field_name` must be overridden to name a ForeignKey field on
-    your model, linking to `wagtailimages.Image`.
+    your model, linking to `media.Image`.
```
ababic: I feel this would be better as: "linking to `wagtailimages.Image` (or a custom Image model)"

wagtail/documents/admin.py
```diff
@@ -5,7 +5,7 @@
 
 if hasattr(settings, 'WAGTAILDOCS_DOCUMENT_MODEL') and settings.WAGTAILDOCS_DOCUMENT_MODEL != 'wagtaildocs.Document':
     # This installation provides its own custom document class;
-    # to avoid confusion, we won't expose the unused wagtaildocs.Document class
+    # to avoid confusion, we won't expose the unused wagtaildocs.LegacyDocument class
```
ababic: I would suggest leaving this as is

wagtail/images/admin.py
```diff
@@ -4,8 +4,8 @@
 from wagtail.images.models import Image
 
 if hasattr(settings, 'WAGTAILIMAGES_IMAGE_MODEL') and settings.WAGTAILIMAGES_IMAGE_MODEL != 'wagtailimages.Image':
-    # This installation provides its own custom image class;
-    # to avoid confusion, we won't expose the unused wagtailimages.Image class
+    # This installation provides its own custom image class or the newstyle;
+    # to avoid confusion, we won't expose the unused wagtailimages.LegacyImage class
```
ababic: I would suggest leaving this as is

wagtail/images/permissions.py
```diff
@@ -1,9 +1,8 @@
 from wagtail.core.permission_policies.collections import CollectionOwnershipPermissionPolicy
 from wagtail.images import get_image_model
-from wagtail.images.models import Image
 
 permission_policy = CollectionOwnershipPermissionPolicy(
     get_image_model(),
-    auth_model=Image,
+    auth_model=get_image_model(),
```
ababic: we should only change this if we're absolutely sure about it... If the permission policy has been using the permissions from the default Image model up until now, this could potentially bring about some unwanted permission-related issues on existing projects

wagtail/project_template/media/models.py
```diff
@@ -0,0 +1,58 @@
...
+from taggit.managers import TaggableManager
+
+
+class Image(AbstractImage):
+    admin_form_fields = (
+        'title',
+        'file',
+        'collection',
+        'tags',
+        'default_alt_text'
```
ababic: should this be 'generic_alt_text'?

wagtail/project_template/media/models.py
```diff
@@ -0,0 +1,58 @@
...
+
+    tags = TaggableManager(help_text=None, blank=True, verbose_name=_('tags'), related_name='images')
+    generic_alt_text = models.CharField(max_length=600, verbose_name=_('default alt text'), blank=True)
+
+    @property
+    def default_alt_text(self):
+        if self.generic_alt_text:
+            return self.generic_alt_text
+        else:
+            return super().default_alt_text
```
ababic: We could lose the else/indentation here

wagtail/search/backends/elasticsearch2.py
```diff
@@ -32,8 +32,8 @@ def get_model_root(model):
     >>> get_model_root(myapp.HomePage)
     wagtailcore.Page
 
-    >>> get_model_root(wagtailimages.Image)
-    wagtailimages.Image
+    >>> get_model_root(media.Image)
+    media.Image
```
ababic: I would suggest leaving this as is

docs/advanced_topics/api/v2/configuration.rst
```diff
@@ -127,7 +127,7 @@ For example:
     class BlogPage(Page):
         published_date = models.DateTimeField()
         body = RichTextField()
-        feed_image = models.ForeignKey('wagtailimages.Image', on_delete=models.SET_NULL, null=True, ...)
+        feed_image = models.ForeignKey('media.Image', on_delete=models.SET_NULL, null=True, ...)
```
ababic: I would suggest leaving this as is

docs/advanced_topics/api/v2/configuration.rst
```diff
@@ -228,7 +228,7 @@ This would add the following to the JSON:
         "feed_image": {
             "id": 45529,
             "meta": {
-                "type": "wagtailimages.Image",
+                "type": "media.Image",
```
ababic: I would suggest leaving this as is

vsalvino:
I am all in favor of this, however have recently experienced a big "gotcha" regarding custom images and 3rd party packages during the process of converting [coderedcms](https://github.com/coderedcorp/coderedcms) from hard-coding `Image` to using the custom image correctly.

For a 3rd party package to properly support custom images, it must do 3 things:
* In its `models.py` it must reference image foreign keys using `get_image_model_string()`
* In its migration files, it must also use `to=get_image_model_string()` in its foreign keys (rather than the default that `makemigrations` will create which is `wagtailimages.Image`)
* In its migration files, it must also specify a dynamic dependency on whatever future app will be implementing the custom image model, as so:
```python
dependencies = [
    migrations.swappable_dependency(get_image_model_string()),
]
```

The problem with this dynamic approach, is if the first-party app (I'll call it `home`) depends on the 3rd party app (`coderedcms` in my case), then creating the custom image model in the `home` app will create a circular migration dependency error at runtime. This is because `coderedcms` is now dynamically depending on `home` for the custom image, and `home` is depending on `coderedcms` for other reasons.

Therefore, to "play nice" with any 3rd party packages you might depend on, you will need to define your custom image model in its own isolated app, to break the circular dependency chain.

I see that this pull request already implements the custom media in its own separate app which is good. However it might be worth noting that NOT implementing it in a separate app will potentially break 3rd party packages.

wagtail/project_template/project_name/settings/base.py
```diff
@@ -26,6 +26,7 @@
 INSTALLED_APPS = [
     'home',
     'search',
+    'media',
```
vsalvino, line 29: Since the default media (file uploads) directory and URL is also named `media` (an old Django convention and also a Wagtail convention) - I am worried having an app named `media` will create a lot of collisions and/or confusion with this convention. For example, it is common to exclude `media/` from version control. Or newcomers might be wondering why there is python code in their `media` directory?

An alternate approach could be to change Wagtail's default MEDIA_ROOT and MEDIA_URL to `uploads` or something similar. However I would be worried about straying from the Django precedents of default `media` and `static` directories.

docs/advanced_topics/api/v2/configuration.rst
thibaudcolas: Same as Andy, it seems to me like the docs should use the actual default. Perhaps it’d help if `wagtail start` provided feedback on what it did upon successful creation of the project, so people who use it are aware of there being a custom image model.

wagtail/project_template/media/models.py
thibaudcolas: I’d prefer we leave this out entirely. I’d love to improve how we do alt text but would rather approach it more strategically ([RFC 51](https://github.com/wagtail/rfcs/pull/51) or similar), based on our best understanding possible of the best practices.

In my opinion, this PR would be plenty valuable enough even if the custom models had no custom fields defined in the base template.

wagtail/project_template/project_name/settings/base.py
thibaudcolas: Same as Vince, there are already two things called `media` in the Wagtail ecosystem so we need to think of another app name. RFC 44 proposed `assets`. I think we should get that RFC approved with a specific app name before we proceed.

thibaudcolas (changes requested):
I’d love to see this happen! Could we resurrect [RFC 44](https://github.com/wagtail/rfcs/pull/44), get it approved, and then proceed with what’s decided there + people’s feedback here?

thibaudcolas:
I’m going to close this now, on the basis that we have an RFC that needs approval before any further work can happen in Wagtail itself: [RFC 44: Make "Custom models" the default](https://github.com/wagtail/rfcs/pull/44). There’s renewed interest in default custom models in the context of [RFC 97: Alt Text Capabilities](https://github.com/wagtail/rfcs/pull/97).

PR #5790, opened 2020-01-24
Support the rel attribute on ModelAdmin buttons

chosak:
Currently it is possible to pass the `target` attribute to a button created using ModelAdmin's ButtonHelper framework. This allows you to generate button links like `<a ... target="_blank">`.

For example, if adding a new button and modeling it after the [existing `edit_button` code](https://github.com/wagtail/wagtail/blob/5e2f50403b093b651f19852e3cd1b4835b3b5125/wagtail/contrib/modeladmin/helpers/button.py#L61-L73), you can add `{'target': "_blank"}` to the returned dict and it'll get passed to the template when the button is rendered.

To be consistent with #4844, and to be consistent with what seems to be the best practice ([1](https://developers.google.com/web/tools/lighthouse/audits/noopener), [2](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/)), we should also support passing the `rel` attribute, which would allow for creation of button links like `<a ... target="_blank" rel="noopener noreferrer">`.


chosak:
Thanks @ababic, merged in 81d23fc.

I agree that it'd be nice to generalize this, but wanted to keep the PR constrained to the specific fix. Perhaps the ModelAdmin button code could somehow leverage the [`wagtail.admin.widgets.Button`](https://github.com/wagtail/wagtail/blob/70da204eb1e8b2083e9a92af05ca895c2defa5dd/wagtail/admin/widgets.py#L260) class, which supports more general setting of attributes?

ababic:
@chosak I agree, reusing existing components here would be the way to go.

PR #5797, opened 2020-01-27
Allow navigation in page chooser modal search

### LINKED ISSUE #5796: Allow switching to tree explorer modal from search modal.

Imagine a site with a large page tree structure and lots of duplicate page titles.

e.g:
```
/conditions/chickenpox/treatments/
/conditions/IBS/treatments/
/conditions/headache/treatments/
/conditions/hand-pain/treatments/
/conditions/foot-pain/treatments/
...hundreds more /conditions/* pages
```

Finding a specific treatment page in the modal page chooser is difficult because:
1. In explorer mode, the user has to trawl through a large number of conditions.
2. In search mode, "treatments" returns a large number of results.

In the page-chooser modal search, users should be able to search for a page and then switch into explorer view to find the child-page that they are interested in.

### Example images

page-chooser-1
page-chooser-3
page-chooser-4

I have some working code so I will open a pull request for this.

mikemonteith:
Fixes #5796

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/js/page-chooser-modal.js
+++ b/wagtail/admin/static_src/wagtailadmin/js/page-chooser-modal.js
@@ -75,7 +75,7 @@ PAGE_CHOOSER_MODAL_ONLOAD_HANDLERS = {
             /* pagination links within search results should be AJAX-fetched
             and the result loaded into .page-results (and ajaxified) */
             $('.page-results a.navigate-pages', modal.body).on('click', function() {
-                $('.page-results', modal.body).load(this.href, ajaxifySearchResults);
+                modal.loadUrl(this.href);
                 return false;
             });
             /* Set up parent navigation links (.navigate-parent) to open in the modal */
--- a/wagtail/admin/templates/wagtailadmin/chooser/_search_results.html
+++ b/wagtail/admin/templates/wagtailadmin/chooser/_search_results.html
@@ -9,7 +9,7 @@
 </h2>
 
 {% if pages %}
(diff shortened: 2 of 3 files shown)


wagtail/admin/static_src/wagtailadmin/js/page-chooser-modal.js
```diff
@@ -75,7 +75,7 @@ PAGE_CHOOSER_MODAL_ONLOAD_HANDLERS = {
             /* pagination links within search results should be AJAX-fetched
             and the result loaded into .page-results (and ajaxified) */
             $('.page-results a.navigate-pages', modal.body).on('click', function() {
-                $('.page-results', modal.body).load(this.href, ajaxifySearchResults);
```
mikemonteith: Deleting this line was a red flag for me, but I couldn't find where it was ever used before

thibaudcolas: The significance of `ajaxifySearchResults` is also lost on me. As far as I can see the regression in this PR doesn’t come from this. Might need further confirmation though.

mikemonteith:
After digging a bit deeper, it seems not as simple as I first thought! The modal ajax search returns a HTML response via "wagtailadmin/chooser/_search_results.html" and the explorer ajax returns a JSON response via "wagtailadmin/chooser/browse.html"

I'm not sure why the difference 🤷‍♂

thibaudcolas:
Closing in favour of #9246. We have been working on new chooser designs for a while, I don’t see ourselves making further gradual improvements to their existing implementation.

PR #5798, opened 2020-01-27
Set context variable to None if no image provided to image template tag

maylonpedroso:
This PR improves the `image` template tag in `wagtailimages_tags`.

How to reproduce the issue it solves:
1. Set a context variable with a list of images and None values
    ```
    def get_context(self, request, *args, **kwargs):
        context =  super().get_context(*args, **kwargs)
        context[images] = [Image(), None]
    ``` 
1. Use the list in a template with the template tag using a context variable
    ```
    {% load wagtailimages_tags %}
    {% for instance in images %}
        {% image instance original as image_node %}
        <img src="{{ image_node.url }}" />
    {% endfor %} 
    ```

This code will render the first image twice as the second call the template tag doesn't clean the context variable. I know this issue could be prevented from the template as I could check for the value before calling the template tag but the template tag should be consistent itself and return the right value for the supplied one.

With this fix, if the supplied value is `None` then the expected context variable it is set to `None` too.

[x] Do the tests still pass? (https://docs.wagtail.io/en/latest/contributing/developing.html#testing)
[x] Does the code comply with the style guide? (Run `make lint` from the Wagtail root)
[x] For Python changes: Have you added tests to cover the new/fixed behavior?
[ ] For front-end changes: Did you test on all of Wagtail’s supported browsers? **Doesn't apply**.
[ ] For new features: Has the documentation been updated accordingly? **Doesn't apply**

### DIFF
--- a/wagtail/images/templatetags/wagtailimages_tags.py
+++ b/wagtail/images/templatetags/wagtailimages_tags.py
@@ -97,6 +97,8 @@ class ImageNode(template.Node):
             return ''
 
         if not image:
+            if self.output_var_name:
+                context[self.output_var_name] = None
             return ''
 
         if not hasattr(image, 'get_rendition'):
--- a/wagtail/images/tests/test_templatetags.py
+++ b/wagtail/images/tests/test_templatetags.py
@@ -1,0 +1,65 @@
+from django.template import Variable
+from django.test import TestCase
+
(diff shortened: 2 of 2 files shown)


gasman:
Thanks @maylonpedroso! Merged in 78cc81c505fc46f4bdaaa2bf35a1b5421aef9e51.

Issue #2871, opened 2016-07-26
RoutablePageMixin without WAGTAIL_APPEND_SLASH


soar:
I don't like trailing slashes - they look ugly. When I use `RoutablePageMixin` with `WAGTAIL_APPEND_SLASH = False` template tag `routablepageurl` generates URLs without slash between main part and routable part of URI. For example class:

``` python
class TestPage(RoutablePageMixin, Page):
    @route(r'^tag/(\w+)/$', name='tag')
    def tag_route(self, request, tag, *args, **kwargs):
        # do something
```

For this example on page `example.com/testpage` tag `{% routablepageurl page 'tag' "test" %}` will generate URI `example.com/testpagetag/test` - this wouldn't work of course. I've tried change route to something like `@route(r'^/tag/(\w+)/$', name='tag')` - it throws 404 error.

With `WAGTAIL_APPEND_SLASH = True` - all works nice.

---

And one more problem: if I try route without ending slash:

``` python
class TestPage(RoutablePageMixin, Page):
    @route(r'^tag/(\w+)$', name='tag')
    def tag_route(self, request, tag, *args, **kwargs):
        # do something
```

It doesn't work at all and shows 404 error for both `somepage/tag/test` and `somepage/tag/test/`.

coredumperror:
I would also like to see this fixed. Fortunately, my codebase already had a workaround for the second problem that @soar mentioned (removing the end slash makes the route totally break), so we don't suffer from that. But the templatetag creating the URL wrong means we can't use it, and must build the URL ourselves in a custom templatetag, which is rather ugly:
```
news_index.url + '/' + news_index.reverse_subpage('rss_feed')
```

As you can see by the need for manually inserting a slash, when `WAGTAIL_APPEND_SLASH = False` the reverse_subpage() function doesn't prepend a separator between the route to the Page and the subroutes defined through RoutablePageMixin. 

Perhaps the `wagtailroutablepage.models.route` dectorator could be updated to take `WAGTAIL_APPEND_SLASH = False` into account, and prepend that slash onto the `pattern` argument? If you think that makes sense, I'd be happy to write up a PR.

Stormheg:
It looks like the `routablepageurl` template tag generating incorrect urls was fixed in [PR #3856]( https://github.com/wagtail/wagtail/pull/3856) which was released a while ago in Wagtail 2.0.

[...]

PR #5806, opened 2020-02-04
Allow url punctuation from RFC 3986

### LINKED ISSUE #3653: RoutablePageMixin URLs which include "." are not routable
I just lost 2 hours of work due to my ignorance of Wagtail's page routing regex. Which is unsurprising, because it's completely undocumented.

I was utterly baffled as to why the `RoutablePageMixin` code that I wrote to create an RSS feed for my NewsIndexPages was not being executed. I eventually had to fire up the debugger and dig deep into the resolver code to figure out why django was throwing a 404 from `example.com/news/feed.rss`, even though `example.com/news` worked just fine. I had the feed route very clearly defined with:

```
@route(r'^feed\.rss$')
rss_feed(self, request):
   ...
```

Turns out that the culprit is in `wagtailcore/utls.py`:
```
if WAGTAIL_APPEND_SLASH:
    serve_pattern = r'^((?:[\w\-]+/)*)$'
else:
    serve_pattern = r'^([\w\-/]*)$'
```

Because `serve_pattern` doesn't include `\.` in its regex, `example.com/news/feed.rss` will never even engage the Page serving mechanism. Django cycles through your site's URL patterns, sees nothing that matches `serve_pattern`, and gives up before calling any Wagtail code.

Now I'm not sure what the proper remedy for this would be. I'm assuming there's some sort of reason that `.` is excluded from `serve_pattern`, so simply adding it doesn't sound like a valid solution. I notice also that the javascript on the Page edit form prevents you from adding a `.` to a Page's slug.

So that means a documentation update would probably be appropriate. Perhaps the `RoutablePageMixin` instructions should state which characters are valid for use in `@route()` regexes? I'd be happy to write a PR for that if you think it's a good solution.

Note, too, that I had to contend with the bug acknowledged in #2871, since I use `WAGTAIL_APPEND_SLASH = False`. Fortunately, that wasn't nearly so frustration-inducing.

Stormheg:
This fixes #3653 by changing the regex against which urls are matched.

I took some of the allowed punctuation characters from [this comment](https://github.com/wagtail/wagtail/issues/3653#issuecomment-532596885) by @gasman 

`- . _ ~` are unreserved characters as stated in section 2.3 of [RFC 3986](https://www.ietf.org/rfc/rfc3986.txt)

I think we should allow all characters because of percentage encoding. Django sees the decoded url which could include things like spaces (`%20`). This could be used in a filename with spaces.

When I wrote the tests for this PR I also encountered issue #2871. Left a note about it in the test.

Feedback is appreciated!

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -10,6 +10,7 @@ Changelog
  * Fix: Re-add 'Close Explorer' button on mobile viewports (Sævar Öfjörð Magnússon)
  * Fix: Add a more descriptive label to Password reset link for screen reader users (Casper Timmers, Martin Coote)
  * Fix: Improve Wagtail logo contrast by adding a background (Brian Edelman, Simon Evans, Ben Enright)
+ * Fix: Change url serving mechanism to allow url punctuations from RFC 3986 section 2.3 (https://www.ietf.org/rfc/rfc3986.txt) (Storm Heg)
  * Support the rel attribute on custom ModelAdmin buttons (Andy Chosak)
 
 
--- a/wagtail/contrib/routable_page/tests.py
+++ b/wagtail/contrib/routable_page/tests.py
@@ -1,6 +1,7 @@
 from unittest import mock
 
 from django.test import RequestFactory, TestCase
+from django.test.utils import override_settings
 from django.urls.exceptions import NoReverseMatch
 
 from wagtail.contrib.routable_page.templatetags.wagtailroutablepage_tags import routablepageurl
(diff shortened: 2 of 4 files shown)


ababic:
@Stormheg Thank you very much for this! I feel you've made the right choice with going with only unreserved characters here. But at the same time, I also feel it could be a valid requirement for a project to use other characters too (if they are aware of the risks, and how they apply to their project). 

How do you feel about accompanying these changes with a Django setting that would allow developers to override this pattern completely if they needed to?

gasman:
@ababic Let's worry about that when that situation comes up - given that #3653 is mostly concerned with allowing dots, I think it's best not to overcomplicate things.

gasman:
Thanks @Stormheg! Merged in 08ef3f4840b22878d006406c548b23144b526f21.

gasman:
Unfortunately this change has broken serving static files via [whitenoise](http://whitenoise.evans.io/) - the URL pattern now matches static files, meaning that they no longer get handled by whitenoise.

Reverted in 98c2aefbe801b68346b4c345c44f840d60c911e8.

coredumperror:
Since Whitenoise serves static files, maybe Wagtail could explicitly exclude `STATIC_ROOT` from its routing regex?

pySilver:
I've simply prepended static URLs _before_ any others

gasman:
@pySilver Unfortunately that won't work for whitenoise, which handles static files through a middleware hook rather than a URL route.

I think the definitive solution to this would be to move [Wagtail's URL routing logic](https://github.com/wagtail/wagtail/blob/9359b14326eb8b7878113e59a7dfaa2edab7f926/wagtail/core/views.py#L16-L17) out of the `serve` view so that it happens during Django's URL resolution. Essentially, just as Django's `path` / `re_path` entries turn into objects that can be queried with "can you handle the current request URL or not", Wagtail's serve route would be a different kind of object that performs the same task, but with the necessary database lookups rather than string pattern matching. That way, if Wagtail fails to find a page, we drop through to the rest of the URL routes (which doesn't happen in the current approach, because we're already committed to the `serve` route by the time we make that decision).

It's a bit unorthodox, and we'd want to make sure we have some decent caching in place so that we're not doing several database lookups on every static file request, but I can't see any technical reason why it wouldn't work.

RealOrangeOne:
Just hit this problem locally, and tried the patch to fix it. More recent versions of whitenoise it seems don't have this issue, and applying the patch manually seems to allow pages to route correctly, whilst keeping static files working.

Happy to take over rebasing this patch if it's useful.

Stormheg:
@RealOrangeOne feel free to take over this patch!

RealOrangeOne:
Finally got around to doing this: https://github.com/wagtail/wagtail/pull/9947

zerolab:
Will close in favour of #9947. Thank you everyone who contributed to this.

PR #5808, opened 2020-02-05
Make server-side slug generation respect WAGTAIL_ALLOW_UNICODE_SLUGS

### LINKED ISSUE #5805: Server-side slug generation should respect WAGTAIL_ALLOW_UNICODE_SLUGS
Currently the `WAGTAIL_ALLOW_UNICODE_SLUGS` setting only affects the behaviour of the client-side slug generation on the page creation/edit view. This means that when pages are created programmatically, the slug can still contain non-ASCII characters:

```
>>> from wagtail.core.models import Page
>>> page = Page(title="A mööse bit me önce")
>>> Page.get_first_root_node().add_child(instance=page)
<Page: A mööse bit me önce>
>>> page.slug
'a-mööse-bit-me-önce'
```

In this case, the slug generation happens in `Page.full_clean`, which currently calls `slugify` with `allow_unicode=True` - this should pass the value of `WAGTAIL_ALLOW_UNICODE_SLUGS` instead.
https://github.com/wagtail/wagtail/blob/afd707f6b40e63faaccbd743bdc0fc2111af0043/wagtail/core/models.py#L434

ArkadiuszMichalRys:
This fixes #5805 by checking the WAGTAIL_ALLOW_UNICODE_SLUGS setting before the slug is created.

I added two tests, one for each case. I based the tests on the original report.

I don't know if this needs a documentation update.

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -430,6 +430,7 @@ Contributors
 * Eric Sherman
 * Martin Coote
 * Simon Evans
+* Arkadiusz Michał Ryś
 
 Translators
 ===========
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -431,7 +431,8 @@ class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):
 
         if not self.slug:
             # Try to auto-populate slug from title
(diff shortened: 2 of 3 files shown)


gasman:
Merged in 47f677577e1df485c57cd6155b9e5fa68afbd69d

PR #5813, opened 2020-02-07
Routable Page url reversing honour WAGTAIL_APPEND_SLASH setting

Stormheg:
Urls generated when the `WAGTAIL_APPEND_SLASH` is false should look like this: `/archive/year/2020`. Notice that there is no trailing slash at the end.

This works fine with the exception being the `reverse_subpage()` method on RoutablePageMixin.

This pull request fixes `reverse_subpage` to take this setting into account.


Stormheg:
Closing this after discussion with @gasman 

RoutablePage URL reversing should ignore `WAGTAIL_APPEND_SLASH` and respect what `@route` says. Which makes this PR invalid.

PR #5817, opened 2020-02-10
Relax Beautifulsoup version restriction

jeromelebleu:
The Beautifulsoup version is stuck to an old one since https://github.com/wagtail/wagtail/commit/93e079bc6c5f1b5b410d7a7115ec9762b76cf3f0 and imposes that to the whole environment, when we could want a newer version. As the [bug](https://bugs.launchpad.net/beautifulsoup/+bug/1784408) which led to the restriction of the 4.6.0 version is now resolved since the 4.6.1 version, this relaxes the Beautifulsoup required version.


gasman:
Thanks @jeromelebleu!
As I recall, there was a follow-up issue in the fixed version (4.6.2) that prevented us from upgrading further - unfortunately I can't find any record of what that issue was, but I'm pretty sure it was related to rich text having strict conditions on the HTML formatting (see https://docs.wagtail.io/en/v2.8/advanced_topics/customisation/rich_text_internals.html#data-format and the regexps at https://github.com/wagtail/wagtail/blob/master/wagtail/core/rich_text/rewriters.py#L7-L9) which were no longer being met. Judging from the wording of the release note in 2.0.2 ("further regressions in formatting empty elements"), it may have been `<embed />` elements being written as `<embed></embed>`.

Whatever the problem was, I _think_ it was caught by our tests at the time, so the tests succeeding here is an encouraging sign that it may have been fixed in a later version of beautifulsoup. I suggest we do some further testing against 4.6.2, and hopefully rediscover what the problem was - and then we can confirm whether it's fixed in current versions.

gasman:
Also related: #5484 (the outdated version of beautifulsoup is one of the remaining dependencies that may be hit by a deprecation in Python 3.9)

gasman:
Have confirmed that there are some test failures in 4.6.2 (and 4.7.x):

* test_embed_conversion_with_custom_whitelist (wagtail.admin.tests.test_rich_text.TestWidgetWhitelisting) - `<embed />` output as a tag pair
* test_embed_conversion_with_default_whitelist (wagtail.admin.tests.test_rich_text.TestWidgetWhitelisting) - `<embed />` output as a tag pair
* test_quoting (wagtail.core.tests.test_whitelist.TestWhitelister) - `` tag output without closing slash (probably harmless)

These are all fixed again in 4.8.0, which checks out with what [the changelog for that release](https://bazaar.launchpad.net/~leonardr/beautifulsoup/bs4/view/head:/CHANGELOG) says about the expanded role of Formatter objects.

Given that our usage of beautifulsoup is pretty sensitive to breaking changes, I'm wary of including 4.9 and above in the range, so I'll go ahead and set it to `>=4.8,<4.9`.

gasman:
Fixed in d192bd1d44a1c3abd98929bafefce609069dd51a.

jeromelebleu:
Thanks for your review and checks!

jeromelebleu:
Note that I was already listed in CONTRIBUTORS.rst (Jérôme Lebleu).
Thanks for taking care of that!

gasman:
Ah, thanks! I did a search for Jerome, but I guess it didn't pick up the version with the accents :-)

PR #5818, opened 2020-02-10
Prevent 'status' banners for page locking being doubled up on POST requests

jacobtoppm:
Previously, a locked page would, when a draft was saved, display double 'locked' banners. This prevents that by checking for a GET request before adding the message (eventually it would be good to move away from messages for this kind of status information, but this fixes the issue for now!)

All tests are passing.


gasman:
Merged in 3ac230f821c1a2ca8ef4cca3c2dbd97cbda25868 - thanks @jacobtoppm!

PR #5826, opened 2020-02-14
Workflow reports

kaedroho:


gasman:
Rebased on current feature/workflow in #5877

PR #5828, opened 2020-02-15
Skip field validation when unpublishing pages.

### LINKED ISSUE #2832: Unpublish with validation errors causes internal server error
I noticed that when saving or publishing a page with validation errors you are redirected to the page edit form with the fields that are invalid highlighted in red. If I try to unpublish a page with validation errors, I get an internal server error instead of the page form with the invalid fields highlighted. The validation check seems appropriate as that is what the changelog says https://github.com/torchbox/wagtail/blob/07ac5aa8ee5456d308789585d0375a2fbe68ede4/CHANGELOG.txt#L180 and the unpublish method calls save(). Also, this was caused by a direct change in the database (so it didn't go through the URLValidator for this URLField), it wasn't changed with the model api (that wouldn't have happened since the model api runs validator checks). So when trying to unpublish that page, it is not clear in the frontend why it is failing and causes an 500 error page instead.

solarissmoke:
Fixes #2832. 

Currently, if you attempt to unpublish a page that contains validation errors, you get a 500 error because saving the page triggers a `full_clean()` on it.

This situation (a page being invalid before you unpublish it) can arise, for example, if you have a nullable but non-blank foreign key field (e.g., a foreign key to `wagtailimages.Image`) which gets set to `null` when the associated object is deleted. It's this particular situation that I've run into the issue with.

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -450,8 +450,9 @@ class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):
 
     @transaction.atomic
     # ensure that changes are only committed when we have updated all descendant URL paths, to preserve consistency
-    def save(self, *args, **kwargs):
-        self.full_clean()
+    def save(self, *args, clean=True, **kwargs):
+        if clean:
+            self.full_clean()
 
         update_descendant_url_paths = False
         is_new = self.id is None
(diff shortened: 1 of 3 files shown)


wagtail/core/models.py
```diff
@@ -451,7 +451,8 @@ def clean(self):
     @transaction.atomic
     # ensure that changes are only committed when we have updated all descendant URL paths, to preserve consistency
     def save(self, *args, **kwargs):
```
ababic: Could I suggest adding the argument to the method signature instead? I feel it would help visibility of this option.

wagtail/core/models.py
```diff
@@ -703,7 +704,7 @@ def unpublish(self, set_expired=False, commit=True):
                 self.expired = True
 
             if commit:
-                self.save()
+                self.save(do_full_clean=False)
```
ababic: How do you feel about just `clean` as an option name?

solarissmoke:
Thanks @ababic . I've updated the PR to address your feedback. Added a docstring to ``unpublish()``, which hopefully is the method you were referring to? Please edit as appropriate.

ababic:
Merged in 061f0e2. Thanks again @solarissmoke :)

Issue #5811, opened 2020-02-06
Enable Pillow 7.0.0


cyface:
Hello! We have been running with Pillow 7.0.0 with our Wagtail installs for a while now, and all seems well.  Just curious if we can update the top end of the Pillow dependency to 7.0.0?  Thank you!

gasman:
Fixed in #5829.

PR #5829, opened 2020-02-16
Relax pillow version restriction

sunpoet:
It allows building with Pillow 7.x.


sunpoet:
It also fixes #5811

gasman:
Thanks @sunpoet!

sunpoet:
Thanks!

PR #5836, opened 2020-02-17
Toggle permission columns on/off

KalobTaulien:
I've had to work with lots of permissions in groups lately and found checking each box for an entire column was a bit painful as a user. 

This PR adds 3 check boxes to toggle all Add/Change/Delete column checkboxes to the Group Edit and Create views. 

Tested on:
- [x] Chrome 78
- [x] Firefox 72
- [x] Safari 12 
- [x] Edge 80 (for MacOS)
- [ ] Internet Explorer (did not test)

This is a preview of the feature. 
check-all


KalobTaulien:
I've re-written some of the JavaScript to be more IE11 friendly so this feature now works. But there seems to be a checkbox issue with IE11. I've opened a bug report in #5903. May need to fix that bug before this work can be merged in.

Scotchester:
Great idea! Would it be possible to duplicate the "Check all" row at the top of the table, as well?

KalobTaulien:
@Scotchester Yeah I don't see why we couldn't do that.

lb-:
Might be a good candidate for a future refactoring to use our `w-bulk` Stimulus controller.

See https://github.com/wagtail/wagtail/pull/10613 for a more recent usage in a PR. Should be mostly adding the check all boxes and some data attributes. 

The Controller may need to be updated to work better with sets of checkboxes though.

I'll add this to the Stimulus project for now & update this with the needs work label.

lb-:
Replaced by https://github.com/wagtail/wagtail/pull/10867

Same behaviour but using the new Stimulus shared implementation for bulk checkboxes.

PR #5837, opened 2020-02-18
Recognise custom tag models for autocompletion in AdminTagWidget

### LINKED ISSUE #3577: Support custom tag models in autocomplete view
[The tag autocomplete view in wagtailadmin](https://github.com/wagtail/wagtail/blob/85bcd5a176f54b628e6022511a25c10345fd522f/wagtail/wagtailadmin/views/tags.py) currently hard-codes the default `taggit.models.Tag` model. This means that if a [custom tag model](http://django-taggit.readthedocs.io/en/latest/custom_tagging.html#custom-tag) is in use, it will fail to provide autocompletions from the correct set of tags. Ideally, we should pick up the tag model from the field that the `AdminTagWidget` is being used on.

gasman:
Fixes #3577


gasman:
Thanks @jacobtoppm! I think `popular_tags_for_model` is out of scope here - it only comes into play on image and document models, whereas I'd say the primary use case for custom tag models is for pages and snippets. Also, I feel there are a lot of open questions about what a fixed version of `popular_tags_for_model` would need to support, given that taggit customisations are pretty open-ended - do we assume that the accessor is named `tags`, or that the tag model uses generic foreign keys, or avoid assuming anything (at which point it becomes a case of "here's a model, now do a whole load of introspection to find out how its tagging mechanism works")? Ultimately I think the answer will be to ditch it completely in favour of something better, as per https://github.com/wagtail/rfcs/pull/28#issuecomment-395026751 :-)

PR #3528, opened 2017-04-06
Fixed #3524 -- Ensured that filenames are checked when downloading documents

ellmetha:

PR #5844, opened 2020-02-18
Replace calls to django.conf.urls.url with path and re_path

### LINKED ISSUE #4433: Refactor URL confs to Django 2.0's simplified routing syntax
Django 2.0 [introduced a more readable syntax](https://docs.djangoproject.com/en/2.0/releases/2.0/#what-s-new-in-django-2-0) for URL confs:

> The new django.urls.path() function allows a simpler, more readable URL routing syntax. For example, this example from previous Django releases:
> 
>    `url(r'^articles/(?P<year>[0-9]{4})/$',` views.year_archive),
> 
> could be written as:
> 
>    `path('articles/<int:year>/',` views.year_archive),
> 

The [path()](https://docs.djangoproject.com/en/2.0/ref/urls/#path) utility function was added to django.urls and should be used from now on for situations in which no raw strings are necessary for regexp matches, while [re_path()](https://docs.djangoproject.com/en/2.0/ref/urls/#re-path) will replace django.conf.urls.url() for any regexp expressions, as [url() is likely to be eventually deprecated.](https://docs.djangoproject.com/en/2.0/ref/urls/#url)

Beyond, path() and re_path(), [include()](https://docs.djangoproject.com/en/2.0/ref/urls/#include) was also added to django.urls, though the previous one in django.conf.urls still exists to maintain backwards compatibility.

There are about 349 URL confs using the old syntax for url() - of which very few are false positives of other function calls - in the following 37 source files and one bytecode file:

admin/api/urls.py
admin/urls/__init__.py
admin/urls/collections.py
admin/urls/pages.py
admin/viewsets/model.py
api/v2/endpoints.py
api/v2/router.py
contrib/forms/urls.py
contrib/forms/wagtail_hooks.py
contrib/redirects/urls.py
contrib/redirects/wagtail_hooks.py
contrib/search_promotions/admin_urls.py
contrib/search_promotions/wagtail_hooks.py
contrib/settings/urls.py
contrib/settings/wagtail_hooks.py
contrib/styleguide/wagtail_hooks.py
core/urls.py
documents/admin_urls.py
documents/urls.py
documents/wagtail_hooks.py
embeds/urls.py
embeds/wagtail_hooks.py
images/admin_urls.py
images/tests/urls.py
images/urls.py
images/wagtail_hooks.py
project_template/project_name/urls.py
search/urls/admin.py
search/urls/frontend.py
search/wagtail_hooks.py
snippets/urls.py
snippets/wagtail_hooks.py
tests/non_root_urls.py
tests/testapp/urls.py
tests/urls.py
users/urls/users.py
users/wagtail_hooks.py
api/v2/__pycache__/router.cpython-36.pyc

I just recently started using Wagtail, and if it's OK, I'd like to do the refactoring, as I feel it's beneficial for readability sake, while leaving it be would be jarring for those coming to Wagtail after completing the Django 2.0 tutorial which already makes use of the new syntax.

For all simple URL conf replacements with no regexp involved, I'll replace url() with path(), while ones with regexp will have their expressions left as is but have url() be replaced by re_path().  The module imports of "from django.conf.urls import include, url" will be replaced with "from django.urls import include, path, re_path"

> [Note](https://docs.djangoproject.com/en/2.0/releases/2.0/#simplified-url-routing-syntax): The django.conf.urls.url() function from previous versions is now available as django.urls.re_path(). The old location remains for backwards compatibility, without an imminent deprecation. The old django.conf.urls.include() function is now importable from django.urls so you can use from django.urls import include, path, re_path in your URLconfs.

Also, the Wagtail "Getting Started -> Integrating Wagtail into a Django project" tutorial [has already started to incorporate some of these changes](http://docs.wagtail.io/en/v2.0/getting_started/integrating_into_django.html#url-configuration).  Unfortunately, the project template from "wagtail start" still generates a urls.py with the old syntax.

As for replacing the regexps of re_path() with [custom path converters](https://docs.djangoproject.com/en/2.0/topics/http/urls/#registering-custom-path-converters), I think that can be addressed later in a larger discussion about trade-offs in a separate issue.

Stormheg:
Fixes #4433

This is a (very large) PR that will replace calls to `django.conf.urls.url()` with `django.urls.path()` or `django.urls.re_path()` where appropriate.

`path()` introduces simpler routing syntax and was introduced in Django 2.0: https://docs.djangoproject.com/en/3.0/releases/2.0/#simplified-url-routing-syntax

**TODO:**
- [x] Fix import sorting issues
- [x] Simplify regexes to use [path converters](https://docs.djangoproject.com/en/3.0/releases/2.0/#simplified-url-routing-syntax) where possible.

I think especially the fairly complex url patterns benefit from this new syntax. It's much clearer although more verbose now.

I'd love to get a review and input on this!

### DIFF
--- a/wagtail/admin/views/pages.py
+++ b/wagtail/admin/views/pages.py
@@ -630,7 +630,7 @@ class PreviewOnEdit(View):
 
     def get_page(self):
         return get_object_or_404(Page,
-                                 id=self.args[0]).get_latest_revision_as_page()
+                                 id=self.kwargs["page_id"]).get_latest_revision_as_page()
 
     def get_form(self, page, query_dict):
         form_class = page.get_edit_handler().get_form_class()
@@ -672,8 +672,9 @@ class PreviewOnEdit(View):
 
 class PreviewOnCreate(PreviewOnEdit):
     def get_page(self):
-        (content_type_app_name, content_type_model_name,
-         parent_page_id) = self.args
+        content_type_app_name = self.kwargs["content_type_app_name"]
+        content_type_model_name = self.kwargs["content_type_model_name"]
+        parent_page_id = self.kwargs["parent_page_id"]
         try:
             content_type = ContentType.objects.get_by_natural_key(
                 content_type_app_name, content_type_model_name)
--- a/wagtail/documents/admin_urls.py
+++ b/wagtail/documents/admin_urls.py
@@ -1,20 +1,20 @@
-from django.conf.urls import url
+from django.urls import path, re_path
 
 from wagtail.documents.views import chooser, documents, multiple
 
 app_name = 'wagtaildocs'
 urlpatterns = [
-    url(r'^$', documents.index, name='index'),
-    url(r'^add/$', documents.add, name='add'),
-    url(r'^edit/(\d+)/$', documents.edit, name='edit'),
-    url(r'^delete/(\d+)/$', documents.delete, name='delete'),
+    path('', documents.index, name='index'),
+    path('add/', documents.add, name='add'),
+    re_path(r'^edit/(\d+)/$', documents.edit, name='edit'),
+    re_path(r'^delete/(\d+)/$', documents.delete, name='delete'),
 
-    url(r'^multiple/add/$', multiple.add, name='add_multiple'),
-    url(r'^multiple/(\d+)/$', multiple.edit, name='edit_multiple'),
-    url(r'^multiple/(\d+)/delete/$', multiple.delete, name='delete_multiple'),
+    path('multiple/add/', multiple.add, name='add_multiple'),
+    re_path(r'^multiple/(\d+)/$', multiple.edit, name='edit_multiple'),
+    re_path(r'^multiple/(\d+)/delete/$', multiple.delete, name='delete_multiple'),
 
-    url(r'^chooser/$', chooser.chooser, name='chooser'),
-    url(r'^chooser/(\d+)/$', chooser.document_chosen, name='document_chosen'),
-    url(r'^chooser/upload/$', chooser.chooser_upload, name='chooser_upload'),
-    url(r'^usage/(\d+)/$', documents.usage, name='document_usage'),
+    path('chooser/', chooser.chooser, name='chooser'),
+    re_path(r'^chooser/(\d+)/$', chooser.document_chosen, name='document_chosen'),
+    path('chooser/upload/', chooser.chooser_upload, name='chooser_upload'),
+    re_path(r'^usage/(\d+)/$', documents.usage, name='document_usage'),
 ]
--- a/wagtail/documents/admin_urls.py
+++ b/wagtail/documents/admin_urls.py
@@ -1,4 +1,4 @@
-from django.urls import path, re_path
+from django.urls import path
 
 from wagtail.documents.views import chooser, documents, multiple
 
--- a/wagtail/documents/admin_urls.py
+++ b/wagtail/documents/admin_urls.py
@@ -6,15 +6,15 @@ app_name = 'wagtaildocs'
 urlpatterns = [
     path('', documents.index, name='index'),
     path('add/', documents.add, name='add'),
-    re_path(r'^edit/(\d+)/$', documents.edit, name='edit'),
-    re_path(r'^delete/(\d+)/$', documents.delete, name='delete'),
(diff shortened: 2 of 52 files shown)


wagtail/admin/views/pages.py
```diff
@@ -672,8 +672,9 @@ def get(self, request, *args, **kwargs):
 
 class PreviewOnCreate(PreviewOnEdit):
     def get_page(self):
-        (content_type_app_name, content_type_model_name,
-         parent_page_id) = self.args
+        content_type_app_name = self.kwargs["content_type_app_name"]
+        content_type_model_name = self.kwargs["content_type_model_name"]
+        parent_page_id = self.kwargs["parent_page_id"]
```
Stormheg: I changed this because `self.args` is empty now (idk why?). We can read the variables from `self.kwargs` but not in the same way it used be.

wagtail/documents/urls.py
```diff
@@ -1,9 +1,11 @@
-from django.conf.urls import url
+from django.urls import path, re_path
 
 from wagtail.documents.views import serve
 
 urlpatterns = [
-    url(r'^(\d+)/(.*)$', serve.serve, name='wagtaildocs_serve'),
-    url(r'^authenticate_with_password/(\d+)/$', serve.authenticate_with_password,
+    re_path(r'^(\d+)/(.*)$', serve.serve, name='wagtaildocs_serve'),
```
Stormheg: I tried to change this to a `path` style url but couldn't get it to work. AFAIK there is no easy way to do an optional path argument (for the `(.*)` part).

wagtail/images/urls.py
```diff
@@ -1,7 +1,8 @@
-from django.conf.urls import url
+from django.urls import path
 
 from wagtail.images.views.serve import serve
 
 urlpatterns = [
-    url(r'^([^/]*)/(\d*)/([^/]*)/[^/]*$', serve, name='wagtailimages_serve'),
+    path('<str:signature>/<int:image_id>/<str:filter_spec>/', serve, name="wagtailimages_serve"),
+    path('<str:signature>/<int:image_id>/<str:filter_spec>/<str:filename>', serve, name="wagtailimages_serve")
```
Stormheg: This line is here twice because of the optional `filename` part. I don't really like duplicating code this way but I thought I'd let this slip because the path converter syntax is so much clearer.

wagtail/documents/admin_urls.py
```diff
@@ -1,20 +1,20 @@
...
+    path('add/', documents.add, name='add'),
+    path('edit/<int:document_id>/', documents.edit, name='edit'),
+    path('delete/<int:document_id>/', documents.delete, name='delete'),
 
-    url(r'^multiple/add/$', multiple.add, name='add_multiple'),
-    url(r'^multiple/(\d+)/$', multiple.edit, name='edit_multiple'),
-    url(r'^multiple/(\d+)/delete/$', multiple.delete, name='delete_multiple'),
+    path('multiple/add/', multiple.add, name='add_multiple'),
+    path('multiple/<str:doc_id>/', multiple.edit, name='edit_multiple'),
+    path('multiple/<str:doc_id>/delete/', multiple.delete, name='delete_multiple'),
```
gasman: These should be `int` rather than `str`

wagtail/documents/urls.py
gasman: Originally the document serve view just had the ID, then the filename was added as an optional cosmetic thing - but this was changed in #3528 so that passing a correct filename is now required. Logically, then, any tests that don't pass a filename should have started failing with #3528 (unless they somehow set a blank filename in the database too...?)

Probably not essential to investigate this as part of the current PR, unless you really want to :-)

wagtail/images/urls.py
gasman: Two URL routes pointing to the same view is fine, but are you sure it's valid to give them both the same value for `name`? Won't that cause `reverse` to fail?

wagtail/admin/views/pages.py
gasman: That makes sense - changing the URL path has changed these from positional arguments to named arguments (and with this being a class-based view, we don't have a function definition to persuade Python to treat them as positional again).

gasman:
Replaced by #6210

Stormheg:
Thanks for completing this @gasman :slightly_smiling_face:

PR #5845, opened 2020-02-19
get_indexed_models is a function, lets call it and loop through the results

specialunderwear:
Most likely nobody has ever called refresh_index, because I just tried it and found this bug.

### DIFF
--- a/wagtail/search/backends/base.py
+++ b/wagtail/search/backends/base.py
@@ -317,7 +317,7 @@ class BaseSearchBackend:
 
     def refresh_index(self):
         refreshed_indexes = []
-        for model in get_indexed_models:
+        for model in get_indexed_models():
             index = self.get_index_for_model(model)
             if index not in refreshed_indexes:
                 index.refresh()


gasman:
Thanks @specialunderwear! Merged in 7dae758ea4eb98671bc2e692c01c75f29384cf55 + parents.

I was curious why this wasn't caught by our test coverage, and it turns out the only places using `refresh_index` are two tests in `wagtail.documents.tests.test_search`and `wagtail.images.tests.test_models` which were being skipped due to a misconfiguration - they were expecting `wagtail.search.backends.elasticsearch` as the backend path, which was only correct for Elasticsearch v1. This is fixed in a8c7ce6c33fee6859ebd45cf8fe4676fc0ac6c14 along with another bug this uncovered - tags weren't being saved when they were edited at the same time as uploading a new document file.

@kaedroho - Is `refresh_index` a necessary part of the search index API, or would we be better off dropping it entirely? It's skipped on everything except elasticsearch, not documented, not used in any real code beyond these two tests (the update_index command uses rebuilder classes), and those tests still pass even with the call to `refresh_index` removed...

kaedroho:
It's necessary for testing (at least on Elasticsearch v1. Not checked if it's still required today) as documents aren't returned in search results for a few seconds after they are indexed.

Issue #5801, opened 2020-01-30
View all Users assigned to a particular Group


isandunk:
### Is your proposal related to a problem?

**When you add a User to a Group in Wagtail, you cannot easily see a list of Users that belong to that Group.**

For example, we have added a number of Users to a "Pathology" Group over a long period of time. Now, we are trying to see which Users belong to that Group. 

### Describe the solution you'd like

When viewing the Group in the wagtail admin, list the Users which have been added to the Group. The items in the list could hyperlink to the relevant User page.

### Describe alternatives you've considered

AFAIK there is only a manual (and impractical) method of sorting through all of our users to see if they have been set to an Admin, and then to check the Roles tab to see which Group they belong to. Plus, we have thousands of users, and it is not possible to sort them by "Level".

### Additional context

This issue may not be an issue on smaller sites that do not have a lot of users. On our setup, we have thousands of users.

karranb:
Hey, can I proceed to implement this issue?
What about Instead of listing the users in the group admin just adding `group` to the users' admin filter? 😄

gasman:
@karranb Sure, go ahead! Adding group as a filter on the user listing could work too, but it might get unwieldy on sites with lots of groups (>100 would not be too unusual).

lb-:
Implemented via #5854

PR #5854, opened 2020-02-22
Group users list

karranb:
### Description
This PR implements a group user listing endpoint.

### Issue

#5801

### DIFF
--- a/client/scss/components/_header.scss
+++ b/client/scss/components/_header.scss
@@ -146,7 +146,7 @@ header {
             @include column(3);
         }
 
-        .col3.addbutton {
+        .col3.actionbutton {
             width: auto;
         }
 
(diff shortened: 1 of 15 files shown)


isandunk:
Thanks so much for your work on this @karranb 

I've tested this in Squash. I have a couple of suggestions from a UX point of view:

- Perhaps we could rename the button so that it is more obvious what it does? So something like "View users of this group" or "View users of <group.id/>"
- In other sections of wagtail, like Images or Documents, they have actions like this in the header (eg. In Documents, it has "Add Document"). Perhaps this button would be better situated there?

Again, thanks for you work.

karranb:
Thanks for the review. Awesome suggestions, I'll work on them 😄

karranb:
Hey @isandunk, I had to make a small refactor on the heading component to accept other buttons besides 'add' ones. I can implement using another approach if you find one that fits better :smile:

isandunk:
Thanks @karranb! I think that's a great solution!

lb-:
merged in va 7565248438337da81d2d3be5f279d5724db0fd7e

thanks again @karranb

PR #5856, opened 2020-02-25
Workflow docs upgrade consideration

jacobtoppm:
Provides documentation for new workflow functionality (except reference on notifications system)


gasman:
Rebased at #5863

Issue #3062, opened 2016-10-09
Wagtail Admin becomes extremely slow as the number of blocks increases.


mevinbabuc:
### Issue Summary

As the number of blocks (blocks with StreamField especially) increases on a _Page_, the admin becomes extremely slow. It becomes very visible after it has more than 10 blocks with _StreamField_ inside the body field of a _Page_. The page load time increases exponentially with every block you add after that.

With around 15 blocks in your StreamField, the page takes around 30 seconds of CPU time to finish processing it and the number of static files goes to 2600+

<img width="620" alt="screen shot 2016-10-09 at 7 40 57 am" src="https://cloud.githubusercontent.com/assets/1276650/19217300/202d8364-8df5-11e6-9411-f2f0cfeade94.png">

When looking close to the source of the rendered HTML, you can see these 2600+ scripts getting loaded or rendered on to the HTML, which takes the bulk of the processing time.

<img width="758" alt="screen shot 2016-10-09 at 7 54 15 am" src="https://cloud.githubusercontent.com/assets/1276650/19217322/c03cf416-8df5-11e6-974f-67cb6f0e5bb5.png">

On further inspecting the script tags, you get to know that most of the code or templates get repeated with a different id to the script tag.
### Questions
1. Is there a way to reduce this script tag inclusions ? or tell wagtail to reuse the scripts ?
2. Can we use template fragment caching, so that these scripts don't get re-evaluated on every page reload.
### Steps to Reproduce
1. Create a simple Page with a body field (StreamField) in it.
2. Add around 15 blocks to the body field (StreamField).
3. Try creating the Page in Wagtail Admin
### Technical details
- Python version: 3.4.
- Django version: 1.8.15.
- Wagtail version: 1.6.3
- Browser version: Chrome 53.

mevinbabuc:
@gasman @m1kola Do you guys have any insight on this and any possible short term solutions for this ?

gasman:
Hi @mevinbabuc,

Your "Steps to Reproduce" doesn't include enough information to reproduce this issue. If I test on a fresh Wagtail project with the following models.py, I get 15 `<script>` blocks, as I would expect:

    from __future__ import absolute_import, unicode_literals

    from django.db import models

    from wagtail.wagtailadmin.edit_handlers import StreamFieldPanel
    from wagtail.wagtailcore import blocks
    from wagtail.wagtailcore.fields import StreamField
    from wagtail.wagtailcore.models import Page


    class HomePage(Page):
[...]

PR #5861, opened 2020-02-25
Add cache to speed up compiling media declarations for stream blocks

gasman:
As proposed by @pimarc in https://github.com/wagtail/wagtail/issues/3062#issuecomment-586331423. I've adapted it slightly to use a set rather than a dict, since they're both equivalent when we're only interested in the keys.

### DIFF
--- a/wagtail/core/blocks/base.py
+++ b/wagtail/core/blocks/base.py
@@ -73,8 +73,21 @@ class Block(metaclass=BaseBlock):
 
     def all_media(self):
         media = forms.Media()
+
+        # In cases where the same block definition appears multiple times within different
+        # container blocks (e.g. a RichTextBlock appearing at the top level of a StreamField
+        # as well as both sides of a StructBlock for producing two-column layouts), we will
+        # encounter identical media declarations. Adding these to the final combined media
+        # declaration would be redundant add processing time when determining the final media
+        # ordering. To avoid this, we keep a cache of previously-seen declarations and only
+        # add unique ones.
+        media_cache = set()
+
         for block in self.all_blocks():
-            media += block.media
+            key = block.media.__repr__()
+            if key not in media_cache:
+                media += block.media
+                media_cache.add(key)
         return media
 
     def all_html_declarations(self):


pimarc:
Thanks for the reactivity. Agree for set.

wagtail/core/blocks/base.py
```diff
@@ -73,8 +73,21 @@ def all_blocks(self):
 
     def all_media(self):
         media = forms.Media()
+
+        # In cases where the same block definition appears multiple times within different
+        # container blocks (e.g. a RichTextBlock appearing at the top level of a StreamField
+        # as well as both sides of a StructBlock for producing two-column layouts), we will
+        # encounter identical media declarations. Adding these to the final combined media
+        # declaration would be redundant add processing time when determining the final media
```
tomdyson: redundant and add

wagtail/core/blocks/base.py
```diff
@@ -73,8 +73,21 @@ def all_blocks(self):
...
+        # ordering. To avoid this, we keep a cache of previously-seen declarations and only
+        # add unique ones.
+        media_cache = set()
+
         for block in self.all_blocks():
-            media += block.media
+            key = block.media.__repr__()
```
gasman: Need to check that `repr` is unique - e.g. it doesn't get cut off after 10 items in the list

kaedroho: What about generating a hash like we do for renditions? https://github.com/wagtail/wagtail/blob/d48d3dfc5b14ef6fd66a1aa5ce652f11f27fb3a1/wagtail/images/models.py#L460

pimarc: I tested with a hash. Unfortunatly it's really too time consuming. As far I had tested (about 30 000 media entries, reduced to about 30 after the fix), __repr__() seems to be the best choice as signature of the media.

zerolab: There was a question during the core meeting whether using repr is actually unique, otherwise js files may not be loaded.

StreamBlocks use forms.Media, and looking at https://github.com/django/django/blob/master/django/forms/widgets.py#L60, I believe the key will be unique to the combination of css/js a block uses. Moreover, if other blocks have the same css/js combination, we have another win here

gasman: The `repr` method wraps the `repr` output of the css dict and js list, which both output their elements in full regardless of length (as confirmed with `repr(['foo'] * 1000)`, `repr({a: 'foo' for a in range(1, 1000)})`) so it doesn't look like this is a problem.

gasman:
Merged in f83977c29127281d7f78a9ff73c4d5f198aa23d8 - thanks @pimarc! (Let me know if you'd like your name updating in the release notes / contributors list...)

pimarc:
Hi Matt,

Thanks for the merge. Not necessary to put my name in release notes.

Regards

Pierre MARC

PR #5865, opened 2020-02-28
Add scope attribute to TableBlock generated tables for visually impaired

Quadric:
This change should not visually and functionally break anything but only support Screen Readers reading tables for people with disabilities.


gasman:
Thanks again @Quadric! Merged in 5e2d12b518d38987b6f82bc1259dcd067a56d1a0.

thibaudcolas:
I was just about to suggest this exact same change, and see it’s done already – amazing! Thank you @Quadric, that’s quite a solid improvement.

Issue #5851, opened 2020-02-21
Avoid outputting href="None" links in the admin for non-routable pages


gasman:
There are a number of situations where a page is considered live but is not accessible at any URL: namely, when the page is outside of any site root, or (as of #4276) the `wagtail_serve` URL route has been removed. Various templates in the admin do not check this when linking to live pages, resulting in links that look like `<a href="None">` - ones that I found in a very brief search were:

* wagtail/admin/templates/wagtailadmin/home/recent_edits.html
* wagtail/admin/templates/wagtailadmin/shared/page_status_tag.html
* the wagtail.contrib.sitemaps app

We should do a more thorough search for these, and fix them where found: 'view live' action buttons (as in recent_edits.html) should not be displayed if there is no URL, and status indicators (as in page_status_tag.html) should be displayed but not linked.

karranb:
Hi, Can I work on this issue?

gasman:
@karranb Yes, please do!

chosak:
I've merged @karranb's fix from #5867 in 36e777418f308fd6aa359e1c953afcf4ab041571, which partially addresses this PR.

@gasman because `wagtail_serve` either will or won't be registered for the entire Wagtail application, does it make sense to pull [this logic](https://github.com/wagtail/wagtail/blob/36e777418f308fd6aa359e1c953afcf4ab041571/wagtail/core/models.py#L832-L838) out somewhere so it can be checked directly? Relatedly, in wagtail.contrib.sitemaps, doesn't this only need to be checked once, because an entire Site either will or won't have URLs?

TUSHAR-VERMA-star:
hey !! Can I take this issue ?

gasman:
@TUSHAR-VERMA-star Yes, go ahead! The 'View live' and page status links were previously handled in #5867, but it looks like the sitemaps app is still in need of a fix, and there may be others. Even if you find that there's nothing else to fix, reporting that here will be useful information for us.

TUSHAR-VERMA-star:
Ok Sir

TUSHAR-VERMA-star:
According to me sitemaps app need not to fix.

ParitoshKabra:
Is anyone still working on this issue? Can I move on with it @TUSHAR-VERMA-star @lb-  😅 ?

SanskarGosavi2003:
Hey, is there someone still working on this? Can I start with this as my very first issue to open source?

lb-:
[...]

PR #5867, opened 2020-03-03
Hide empty 'view live' links

karranb:
This PR is related to the issue #5851

I searched for "view live" buttons in the admin panel and then:
- I removed the action buttons without URLs
- I replaced the label links without URLs by spans and
- I also removed the non-routable live pages from the sitemap generator

In order to do these items, I created a page property that checks if the page has a url.

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/home/locked_pages.html
+++ b/wagtail/admin/templates/wagtailadmin/home/locked_pages.html
@@ -31,7 +31,7 @@
                                     {% if page.has_unpublished_changes %}
                                         <li><a href="{% url 'wagtailadmin_pages:view_draft' page.id %}" class="button button-small button-secondary" target="_blank" rel="noopener noreferrer">{% trans 'Draft' %}</a></li>
                                     {% endif %}
-                                    {% if page.live %}
+                                    {% if page.live_and_routable %}
                                         <li><a href="{{ page.url }}" class="button button-small button-secondary" target="_blank" rel="noopener noreferrer">{% trans 'Live' %}</a></li>
                                     {% endif %}
                                 </ul>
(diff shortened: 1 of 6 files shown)


wagtail/admin/templates/wagtailadmin/home/locked_pages.html
```diff
@@ -31,7 +31,7 @@ <h2>{% trans "Your locked pages" %}</h2>
                                     {% if page.has_unpublished_changes %}
                                         <li><a href="{% url 'wagtailadmin_pages:view_draft' page.id %}" class="button button-small button-secondary" target="_blank" rel="noopener noreferrer">{% trans 'Draft' %}</a></li>
                                     {% endif %}
-                                    {% if page.live %}
+                                    {% if page.live_and_routable %}
```
chosak: Because `live_and_routable` also checks the page's URL, this effectively makes two duplicate lookups to `page.url`, potentially hitting the database more than once each time. How would you feel about instead doing a single lookup, for example:

```html
{% if page.live %}
{% with page_url = page.url %}
    {% if page_url is not None %}
        <li><a href="{{ page_url }}" ...></li>
    {% endif %}
{% endwith %}
{% endif %}
```

wagtail/contrib/sitemaps/sitemap_generator.py
```diff
@@ -24,14 +24,16 @@ def get_wagtail_site(self):
         return site
 
     def items(self):
-        return (
+        return list(filter(
+            lambda item: item.url is not None,
```
chosak: What's the situation that necessitates this check on a page by page basis? Or, more generally, isn't it the case that, since this code is iterating over all pages within a site, all pages either will or won't be routable, depending on if `wagtail_serve` is defined?

wagtail/admin/templates/wagtailadmin/home/locked_pages.html
karranb: Nice, this is a better approach. I'll work on this, thanks!

wagtail/contrib/sitemaps/sitemap_generator.py
karranb: You're right, I'll undo this and take a deeper look to make sure only routable URLs are being listed here. Sorry.

karranb:
Hi @chosak, I worked on the changes you suggested :smile:

chosak:
@karranb thanks! Merged in 36e777418f308fd6aa359e1c953afcf4ab041571.

PR #5868, opened 2020-03-03
Feature/workflow help text

jacobtoppm:
Adds the ability to override a FieldPanel's heading by providing it explicitly in the constructor. (Previously this would be overridden as soon as a field was bound by the form field's label)

Adds a test for this new function

Rewords workflow form text and other admin text.

Fixes duplicate disable workflow button bug


wagtail/admin/edit_handlers.py
```diff
@@ -531,7 +531,7 @@ def db_field(self):
 
     def on_form_bound(self):
         self.bound_field = self.form[self.field_name]
-        self.heading = self.bound_field.label
+        self.heading = self.heading or self.bound_field.label
```
zerolab: The changes look good to me.
Would be great to update the docs as well

jacobtoppm: Good point, thanks! I've added to the panels reference in latest commit

jacobtoppm:
Thanks! 

> Looks good to me!
> 
> I'd suggest pulling out the FieldPanel change into its own PR that we can merge into master separately, so that we don't forget to mention it in the release notes... from the look of things, that would involve cherry-picking [1f7547f](https://github.com/wagtail/wagtail/commit/1f7547fc424dbf761cc16ba94c9f0afe9f09897e), [05a1f0e](https://github.com/wagtail/wagtail/commit/05a1f0ef72ebf1345258e7755c0dfe86954fe38a), [f771d55](https://github.com/wagtail/wagtail/commit/f771d55121d4951f8a25d50aedc8196e128eec1e), [ecdfcc7](https://github.com/wagtail/wagtail/commit/ecdfcc78363f0c81c4b465603ddde685317f4a31) and [cb9245b](https://github.com/wagtail/wagtail/commit/cb9245badcfebe6fe8220fd89211f558f29eb97e).

Thanks! I've pulled out into a new PR here: https://github.com/wagtail/wagtail/pull/5870

PR #5869, opened 2020-03-04
Fix Javascript alerts reported by LGTM

gasman:


wagtail/admin/static_src/wagtailadmin/js/core.js
```diff
@@ -271,7 +267,7 @@ $(function() {
         }
 
         function getURLParam(name) {
-            var results = new RegExp('[\?&]' + name + '=([^]*)').exec(window.location.search);
+            var results = new RegExp('[\\?&]' + name + '=([^]*)').exec(window.location.search);
```
tomdyson: is this right? it looks like a backslash is being added to the match options, instead of an escaped question mark.

gasman: As I understand it, LGTM is picking up on the fact that backslash escaping happens in two places - first when interpreting the string (where you'd typically use backslashes for things like `\'` and `\n`) and secondly within the regexp engine.

We need to double up the backslash so that the string interpreter sees it as "literal backslash followed by question mark" rather than an escaped question mark (which is equivalent to a non-escaped question mark, because Javascript string syntax will happily let you escape punctuation that doesn't need escaping...) and thus the `RegExp` constructor will receive it as `[\?&]` rather than `[?&]`.

PR #5870, opened 2020-03-05
Edit handler heading override

jacobtoppm:
Pulled out from https://github.com/wagtail/wagtail/pull/5868 as requested to ensure this is added to release notes:

Adds the ability to override a FieldPanel's heading by providing it explicitly in the constructor. (Previously this would be overridden as soon as a field was bound by the form field's label)

Adds a test for this new function

### DIFF
--- a/docs/reference/pages/panels.rst
+++ b/docs/reference/pages/panels.rst
@@ -8,7 +8,7 @@ Available panel types
 FieldPanel
 ----------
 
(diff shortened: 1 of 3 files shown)


gasman:
Merged in 526649008da4c58a5c9afed263932a8791a6e9ab - thanks @jacobtoppm!

PR #5877, opened 2020-03-09
Workflow reports (rebase of #5826/#5859)

gasman:


wagtail/admin/templates/wagtailadmin/reports/workflow.html
```diff
@@ -0,0 +1,108 @@
...
+            <p>{% trans "No pages have been submitted for moderation yet" %}</p>
+        {% endif %}
+    </div>
+{% endblock %}
+
+{% block extra_js %}
+    {{ block.super }}
+
+    {# load editor js to get tooltips #}
+    {% include "wagtailadmin/pages/_editor_js.html" %}
```
gasman: Let's not use _editor_js.html as a Random Grab-Bag Of Useful Javascript for things that aren't the page editor, please. Presumably this could just be an import of wagtailadmin/js/vendor/bootstrap-tooltip.js instead...

gasman:
Minor tweaks made: imported wagtailadmin/js/vendor/bootstrap-tooltip.js instead of the whole of _editor_js.html, added a basic test and added ordering on the queries to address a warning about paginating an unordered queryset. Passing on to @jacobtoppm for re-review...

gasman:
Merged into feature/workflow in a99ef6d1aee73c384e0ed15af2e22e7855ba84f5 , along with extra test assertions (which revealed that we ought to be using `get_full_name` for the submitter - changed now) and adding the clipboard icon.

PR #5880, opened 2020-03-10
Report exports

jacobtoppm:
- Adds a new `SpreadsheetExportMixin` which provides more generic csv and xslx exports from a queryset
- Adds xlsx export functionality to Wagtail forms
- Adds spreadsheet export to locked pages report
- Adds spreadsheet export to ModelAdmin `IndexView` (by defining `list_export` similarly to `list_display`)
- Makes xlsx export the default, adjusted via a new setting, ``WAGTAIL_SPREADSHEET_EXPORT_FORMAT``, as per decision made on https://github.com/wagtail/wagtail/issues/3466 and helping address security concerns
- Adds documentation and tests

PR #5884, opened 2020-03-13
Reports docs

jacobtoppm:
Adds a new guide to ``advanced topics`` in the documentation explaining how to subclass ``ReportView`` to create a custom report and set up spreadsheet export, as well as use the relevant hooks to add it to the Reports submenu and admin urls.

(Relies on https://github.com/wagtail/wagtail/pull/5880)


docs/advanced_topics/adding_reports.rst
```diff
@@ -0,0 +1,126 @@
...
+
+A list of the fields/attributes for each model which are exported as columns in the spreadsheet view. By default,
+this is identical to the listing fields: the title, time of the last update, status, and specific type of any pages.
+
+.. attribute:: export_heading_overrides
+
+(dictionary)
+
+A dictionary of any fields/attributes in ``list_export`` for which you wish to manually specify a heading for the spreadsheet
+column, and their headings. If unspecified, heading will be taken from the field ``verbose_name`` if applicable, and the
```
tomdyson: unspecified, heading -> unspecified, the heading

docs/advanced_topics/adding_reports.rst
```diff
@@ -0,0 +1,126 @@
...
+
+
+    class CustomReportView(ReportView):
+        pass
+
+
+Defining your report
+~~~~~~~~~~~~~~~~~~~~~
+
+The most important attributes and methods to customise to define your report are:
```
tomdyson: are there other less important attributes and methods which aren't referenced here?

jacobtoppm: Yes - in addition to the standard inherited `get_context`, etc methods from `BaseListView` and `TemplateResponseMixin`, there are a number of other methods in `SpreadsheetExportMixin` which you could in principle override. I wouldn't expect this to be necessary in most cases, though.

docs/advanced_topics/adding_reports.rst
jacobtoppm: Thanks, will change!

jacobtoppm:
> This looks great @jacobtoppm! I suggest adding one complete example, with a custom report including the most important attributes and methods, and accompanying menu item and admin URL.

Thanks! That sounds good, I'll add that now.

gasman:
Looks good to me! Only thing I'd add (besides the typo fix Tom raised) is that I think we should encourage people to limit the visibility of menu items to just admins where possible. Currently that means subclassing MenuItem and overriding `is_shown`, but rather than going through that in docs, perhaps it's worth us providing a ready-made `AdminOnlyMenuItem` in `wagtail.admin.menu`.

(To do it properly, we probably ought to link that behaviour to specific permissions, so that people can make them more fine-grained than 'admin or not'. But I think that's a rabbit-hole too far at this point...)

jacobtoppm:
Thanks both! I've made the suggested changes now - ``AdminOnlyMenuItem`` sounds like a good idea to me.

gasman:
Merged in 205249b3174249e811a37a2391e13725981a44da + parents - thanks @jacobtoppm!

PR #5886, opened 2020-03-18
Do not download entire search index on search terms with no hits

specialunderwear:
When a slice is specifically set to [0:0] - which would yield an empty list - instead wagtail will clear the slice, effectively downloading the entire search
index.

This scenario happens when a search term does not return any results.
So every time you enter a search term that has no results, the entire search
index is downloaded, you don't really feel it because wagtail only downloads
the pk values, and it is unlikely a site has millions of pages, but it is still
very inefficient and unwanted.

### DIFF
--- a/wagtail/search/backends/base.py
+++ b/wagtail/search/backends/base.py
@@ -220,8 +220,8 @@ class BaseSearchResults:
 
         if isinstance(key, slice):
             # Set limits
-            start = int(key.start) if key.start else None
-            stop = int(key.stop) if key.stop else None
+            start = int(key.start) if key.start is not None else None
+            stop = int(key.stop) if key.stop is not None else None
             new._set_limits(start, stop)
 
             # Copy results cache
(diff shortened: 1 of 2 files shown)


lb-:
@specialunderwear  - thanks for submitting this PR! great to see things get more efficient if we can.

Would you be ok to add a unit test for this? this file would possibly be a suitable location to set up this scenario and test it `wagtail/search/tests/test_page_search.py`.

This way we can help ensure this improvement will not be accidentally removed in the future.

allcaps:
@specialunderwear I took the liberty to add the test.

Groet, Coen

lb-:
Merged in via 6194dcb74b71c5131a74ea81133fddccb95e8dcd

Thanks again @specialunderwear  & @allcaps

specialunderwear:
ok then @allcaps my github spam settings are very strict, so I only saw the commit, not the mentions, cheers!

PR #5891, opened 2020-03-19
Order sites in site switcher by site_name

tleguijt:
I've added this small change since managing a large set of sites can become quite difficult if these are not ordered.

### DIFF
--- a/wagtail/contrib/settings/forms.py
+++ b/wagtail/contrib/settings/forms.py
@@ -18,7 +18,7 @@ class SiteSwitchForm(forms.Form):
         initial_data = {'site': self.get_change_url(current_site, model)}
         super().__init__(initial=initial_data, **kwargs)
         sites = [(self.get_change_url(site, model), site)
-                 for site in Site.objects.all()]
+                 for site in Site.objects.all().order_by("site_name")]
         self.fields['site'].choices = sites
 
     @classmethod


allcaps:
Maybe put empty site names after and give them an alternative field to order by?
```
Site.objects.all().annotate(empty_string=Case(
            When(sitename=u'', then=Value(1)),
            default=Value(0),
            output_field=IntegerField()
            )).order_by('empty_string', Lower('sitename'), Lower('domain'))
```

allcaps:
I had a look. Site switcher lives in wagtail.contrib.sitesettings
https://docs.wagtail.io/en/v2.1.1/reference/contrib/settings.html

The best ordering would be the default Site ordering. It would give the same ordering to sites in `/admin/sites/` and in the site switcher. 

The first column in `/admin/sites/` is the field hostname. The site name is optional. So ordering by hostname is the smart thing to do. 

Only one site can be `[default]`. It is the default site when navigating to site settings. Let's show the default site on top of the list.

At the moment site objects do not have a default ordering defined. So it depends on the db used what order is used. I have the impression that sqlite uses the pk (and thereby the order of creation). For Postgres it will be more random.

We should add ordering on Site model:
```
class Meta:
    ...
    ordering = ["-is_default_site", Lower("hostname")]
```
Or on the SiteManager:
```
class SiteManager(models.Manager):
    def get_queryset(self):
        return super(SiteManager, self).get_queryset().order_by("-is_default_site", Lower("hostname"))
```

@gasman  Do you know if the Site ordering is omitted for a good reason? Do you have a preference for Meta.ordering OR SiteManager.get_queryset?

lb-:
Superseded by #6004

tleguijt:
Thanks for the work on this @allcaps! Haven't got around to it finalizing it myself

Issue #5895, opened 2020-03-19
Invalid Image property "focal_point" in template file makes notice missing on edit image view


Quadric:
### Issue Summary

When editing file there is a notice about "FOCAL POINT" and info "To define this image's most important region, drag a box over the image above.

but in file:
```
https://github.com/wagtail/wagtail/blob/master/wagtail/images/templates/wagtailimages/images/edit.html
```
there is also check for 
```
{% if image.focal_point %}({% trans "Current focal point shown" %}){% endif %}
```

Problem with that this property "focal_point" of imag does not exists - i guess this should be "has_focal_point" which exists and is used in other places.

is
![image](https://user-images.githubusercontent.com/1465170/77100207-8d8e6b80-6a15-11ea-95f3-6f7c8f41f239.png)

to be
![image](https://user-images.githubusercontent.com/1465170/77100256-a008a500-6a15-11ea-8dc1-0e90ef601100.png)

### Steps to Reproduce

1. Upload new image
2. Got to edit image
3. Set focal point
4. Save changes
5. Again go to edit view
6. You'll that there is no info about "Current focal point shown"

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: yes

### Technical details

* Python version: 3.7.4
* Django version: 3.0.3
* Wagtail version: checked on master
* Browser version: Google Chrome 80.0.3987.132

PR #5896, opened 2020-03-19
Fix focal point notice on images edit view

Quadric:
Fixes: #5895


Quadric:
i've made rebased version in https://github.com/wagtail/wagtail/pull/5897

Issue #5893, opened 2020-03-19
'Locked Pages' report overrides language preference


gasman:
### Issue Summary

[As reported by Вадим Козячий on the Google group](https://groups.google.com/forum/#!topic/wagtail/5gqURJMoMM8) - visiting the Locked Pages report overrides the logged-in user's language setting.

### Steps to Reproduce

1. On a fresh instance of bakerydemo, go to Account settings -> Language preferences and set language to German
2. Go to Berichte -> Gesperrte Seiten
3. Note that the interface is now in English
4. Browse elsewhere in the admin - the interface is now in German again

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: yes

### Technical details

* Python version: 3.6.8
* Django version: 3.0.4
* Wagtail version: master (2.9a0)
* Browser version: Chrome 80.0.3987.132

jacobtoppm:
Thanks for raising this! After testing, I believe this might actually be a problem common to views returning a `TemplateResponse`, which includes for example Collections views: the `require_admin_access` decorator we normally use to activate the language does not seem to be compatible with these. Forcing a render in the decorator worked when I tested, so this might work as a fix, or we might need to try a different translation approach - I intend to look into this a bit more.

gasman:
Fixed in #5910

PR #5910, opened 2020-03-25
Fix user preferred language override not working with TemplateResponse

jacobtoppm:
This addresses https://github.com/wagtail/wagtail/issues/5893, which turned out to be a much more general issue, where the user profile language would not be used on rendering anything returning a TemplateResponse from the Wagtail Admin, including collections and modeladmin.

This fix is intended to preserve as much as possible the ability of the TemplateResponse to be further modified by middleware if we decide to change to use these more. However, given that we largely return HttpResponses, the simpler fix of forcing the TemplateResponse to render could be preferred - I'm happy to implement that instead if that's preferred.

### DIFF
--- a/wagtail/admin/auth.py
+++ b/wagtail/admin/auth.py
@@ -1,6 +1,7 @@
 from functools import wraps
 
 import l18n
+import types
 
 from django.contrib.auth import get_user_model
 from django.contrib.auth.views import redirect_to_login as auth_redirect_to_login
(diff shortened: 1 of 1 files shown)


ababic:
Related #5731

gasman:
Thanks @jacobtoppm! Merged in df8e51a6e9c391d2b8294bb3db71bc5d194a5d4d.

PR #5921, opened 2020-03-31
Workflow implementation fixes

jacobtoppm:
Fixes:
- `get_next_task` excluding tasks erroneously (see commit for details)
- Task type selection only allowing one level of inheritance, and not preventing attempts to create abstract tasks
- Mail connection opening not wrapped in try/except block

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -2758,6 +2758,9 @@ class WorkflowState(models.Model):
                     # not STATUS_IN_PROGRESS), move to the next task
                     self.current_task_state = next_task.specific.start(self, user=user)
                     self.save()
+                    # if task has auto-approved, update the workflow again
+                    if self.current_task_state.status != self.current_task_state.STATUS_IN_PROGRESS:
+                        self.update(user=user)
                 # otherwise, continue on the current task
             else:
                 # if there is no uncompleted task, finish the workflow.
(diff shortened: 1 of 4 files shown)


wagtail/core/models.py
```diff
@@ -2758,14 +2758,17 @@ def update(self, user=None, next_task=None):
...
+                        self.update(user=user)
                 # otherwise, continue on the current task
             else:
                 # if there is no uncompleted task, finish the workflow.
                 self.finish(user=user)
 
     def get_next_task(self):
         """Returns the next active task associated with the latest page revision, which has not been either approved or skipped"""
-        return Task.objects.filter(workflow_tasks__workflow=self.workflow, active=True).exclude(Q(task_states__page_revision=self.page.get_latest_revision()), Q(task_states__status=TaskState.STATUS_APPROVED) | Q(task_states__status=TaskState.STATUS_SKIPPED)).order_by('workflow_tasks__sort_order').first()
+        return Task.objects.filter(workflow_tasks__workflow=self.workflow, active=True).exclude(task_states__in=TaskState.objects.filter(Q(page_revision=self.page.get_latest_revision()), Q(status=TaskState.STATUS_APPROVED) | Q(status=TaskState.STATUS_SKIPPED))).order_by('workflow_tasks__sort_order').first()
```
gasman, line 2771: This could really do with being split onto multiple lines...
```suggestion
        return (
            Task.objects.filter(workflow_tasks__workflow=self.workflow, active=True)
            .exclude(
                task_states__in=TaskState.objects.filter(
                    Q(page_revision=self.page.get_latest_revision()),
                    Q(status=TaskState.STATUS_APPROVED) | Q(status=TaskState.STATUS_SKIPPED)
                )
            ).order_by('workflow_tasks__sort_order').first()
        )
```

jacobtoppm: Good point, thanks! Happy to commit that

PR #5923, opened 2020-03-31
Facet search: make 'All' translatable

luizboaretto:
Minor fix allowing "All" to be translatable on wagtail search results


kaedroho:
Thanks!

PR #5928, opened 2020-04-02
Update Page.objects.specific() to warn instead of erroring when specific page data cannot be found

ababic:
Imagine you're working on a project with multiple branches, in various different development states. You create a new page type in one branch, create/apply the relevant database migrations as usual, and create an instance to play around with. Then, you find yourself needing to switch to a branch with a different migration state, so reverse the migrations in the current branch before switching over. 

Now you load up wagtail again to find a different page, but the explorer widget and the pages section no longer work, because the table that contained the 'specific' data for the previously create page no longer exists. And even if you switch back to the other branch and run migrations to add the model back in, the record won't exist in the newly created database table.

In earlier versions of Django, it seemed that removing a model via a migration would also delete the associated `ContentType` from the database, triggering [this line](https://github.com/wagtail/wagtail/blob/master/wagtail/core/models.py#L268) to kick in and update the `content_type` of the affected pages to `wagtailcore.Page`. However, since Django 1.11 and onwards, it seems that the content type is no longer deleted, and this line is no longer triggered - which makes Page.objects.specific() quite a bit more fragile.

The approach I've take here is, instead of raising an `KeyError` when a specific page cannot be found, I supplement the result with vanilla `Page` objects, mimicking the result we used to get under such circumstances.

I've also added a warning to help allude developers to the fact that specific pages are missing in these circumstances, and provided a short description of what might have caused it.

I was also thinking about coupling these changes with a 'cleanup' command to set the content types of any problematic pages to `wagtailcore.Page`.

Any thoughts/comments would be much appreciated :)

### DIFF
--- a/wagtail/core/query.py
+++ b/wagtail/core/query.py
@@ -1,4 +1,5 @@
 import posixpath
+import warnings
 from collections import defaultdict
 
 from django.apps import apps
@@ -373,6 +374,8 @@ def specific_iterator(qs, defer=False):
 
     This should be called from ``PageQuerySet.specific``
     """
+    from wagtail.core.models import Page
+
     annotation_aliases = qs.query.annotations.keys()
     values = qs.values('pk', 'content_type', *annotation_aliases)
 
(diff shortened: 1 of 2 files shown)


zerolab:
Looks like LGTM has updated its checks

ababic:
@zerolab Okay. I have evaluated all of those, but it doesn't look like any of those new alerts are due to changes in this PR. I can see this having the potential to be a little annoying going forward :)

ababic:
@gasman This has now been rebased and updated to maintain compatibility with the recent work to preserve annotations

gasman:
Merged in 58741d3989ecea7c44b0f12fd780a9673ec750f7

Issue #5926, opened 2020-04-01
StreamField should consistently use bulk_to_python() to improve efficiency


ababic:
### Is your proposal related to a problem?

It's very common in projects that I work on to define blocks that have a heading of some kind, followed by a curated list of pages, snippets or other objects (maybe up to 10 or so). Currently, when rendering a page with such a block in place, as many queries as there are items are made to fetch the necessary data, which clearly isn't as efficient as it could be.

### Describe the solution you'd like

Nested instances of ListBlock should utilise `bulk_to_python()` to fetch multiple objects at once.

An added bonus would be adding some simple way of optimising further, by allowing some control over how 'prefetch_related' and 'select_related' are used in `bulk_to_python()`. Perhaps they could be supplied as arguments, or each block could look for `streamvalue_select_related` and `streamvalue_prefetch_related` attributes on the target model, or a `for_streamvalue()` method on the queryset itself?

### Additional context

It looks as though `StreamValue`[ should already support this](https://github.com/wagtail/wagtail/blob/e1d3390a1f62ae4b30c0ef38e7cfa5070d8565dc/wagtail/core/blocks/stream_block.py#L408), but it seems not to be working past the first level for me when testing locally. The built-in `SnippetChooserBlock`, `PageChooserBlock`, `ImageChooserBlock` and `DocumentChooserBlock` classes already implement a `bulk_to_python()` method to support such optimisation. 

Below is a generic block example that can be used to test this:

```
from wagtail.images.blocks import ImageChooserBlock
from wagtail.core.blocks import CharBlock, ListBlock, StructBlock

class GalleryBlock(StructBlock):
    heading = CharBlock(required=False)
    images = ListBlock(ImageChooserBlock())
```

chosak:
I've opened #5930 that partially addresses this by allowing each ListBlock to retrieve its items in a single query. This doesn't address the more general desire to do bulk loading across ListBlocks; for example, in the example you provide above @ababic it'd be nice to be able to load _all_ images across _all_ GalleryBlocks at once.

[...]

PR #5930, opened 2020-04-02
Allow ListBlocks to leverage child bulk_to_python

chosak:
Currently a select set of StreamField blocks like PageChooserBlock expose a bulk_to_python method that is used to optimize their retrieval from the database. As reported in #5926, ListBlock could take advantage of this so that its child items are loaded at once, instead of one at a time.

This change modifies how ListBlock.to_python works so that it calls its child block's bulk_to_python, if defined. This allows for a single query instead of one query per child item.

Note that this change doesn't add `bulk_to_python` to ListBlock itself, meaning that individual ListBlocks in a StreamField or StructBlock are still retrieved independently. But it does optimize the lookup for each ListBlock.

### DIFF
--- a/wagtail/core/tests/test_blocks.py
+++ b/wagtail/core/tests/test_blocks.py
@@ -3275,6 +3275,26 @@ class TestPageChooserBlock(TestCase):
             'wagtail.core.blocks.PageChooserBlock',
             (), {'page_type': ['tests.SimplePage', 'tests.EventPage']}))
 
+    def test_bulk_to_python(self):
+        page_ids = [2, 3, 4, 5]
+        expected_pages = Page.objects.filter(pk__in=page_ids)
+        block = blocks.PageChooserBlock()
+
+        with self.assertNumQueries(1):
+            pages = block.bulk_to_python(page_ids)
+
+        self.assertSequenceEqual(pages, expected_pages)
+
+    def test_bulk_to_python_when_child_of_listblock(self):
+        page_ids = [2, 3, 4, 5]
+        expected_pages = Page.objects.filter(pk__in=page_ids)
+        block = blocks.ListBlock(blocks.PageChooserBlock())
+
+        with self.assertNumQueries(1):
+            pages = block.to_python(page_ids)
+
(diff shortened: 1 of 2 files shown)


wagtail/core/tests/test_blocks.py
```diff
@@ -3275,6 +3275,26 @@ def test_deconstruct_target_model_multiple_literals(self):
...
+
+    def test_bulk_to_python_when_child_of_listblock(self):
+        page_ids = [2, 3, 4, 5]
+        expected_pages = Page.objects.filter(pk__in=page_ids)
+        block = blocks.ListBlock(blocks.PageChooserBlock())
+
+        with self.assertNumQueries(1):
+            pages = block.to_python(page_ids)
+
+        self.assertSequenceEqual(pages, expected_pages)
```
ababic, line 3296: It is nice having these two tests together, as the patterns complement each other nicely. However, I think this last test would better belong to under a new `TestListBlock` class

ababic:
@chosak wow, thanks so much for the lightning fast response. Just one minor point from me, but otherwise looking great. I take my hat off to you :)

wagtail/core/tests/test_blocks.py
chosak: Agreed -- there is already a [`TestListBlock`](https://github.com/wagtail/wagtail/blob/a87b3fce9bf33fde53008d5987b28a22970cf22e/wagtail/core/tests/test_blocks.py#L1942) but I didn't add the test there because it's a SimpleTestCase and doesn't load the test.json fixture like TestPageChooserBlock does. I didn't want to convert it to a TestCase and thus add the database fixture loading unnecessarily to all of its tests. Maybe I can just add another TestListBlockWithFixture to isolate just this one test.

ababic: > Maybe I can just add another TestListBlockWithFixture to isolate just this one test.

Yep, this sounds ideal :)

chosak:
Thanks for the reviews! Merged in 3797132b4.

PR #5932, opened 2020-04-03
Allow select_related() to be utilised when fetching setting model objects for a site

ababic:
Builds on the work done in #5931 (to be merged first) to provide a convenient way to utilise select_related() when fetching setting model instances for a site/request. Simply set a list of the foreign key fields you want to be fetched to your class, and the related object data for those fields will be fetched in the same query when looking up the object, reducing the need for additional queries to access the related objects later on.

- [x] Have you added tests to cover the new/fixed behaviour?
- [x] Has the documentation been updated accordingly?

### DIFF
--- a/docs/reference/contrib/settings.rst
+++ b/docs/reference/contrib/settings.rst
@@ -222,3 +222,47 @@ Or, alternately, using the ``set`` tag:
 .. code-block:: html+jinja
 
     {% set social_settings=settings("app_label.SocialMediaSettings") %}
+
+
+Utilising ``select_related`` to improve efficiency
+--------------------------------------------------
+
+For models with foreign key relationships to other objects (e.g. pages),
+which are very often needed to output values in templates, you can set
+the ``select_related`` attribute on your model to have Wagtail utilise
+Django's `QuerySet.select_related() <https://docs.djangoproject.com/en/stable/ref/models/querysets/#select-related>`_
+method to fetch the settings object and related objects in a single query.
+With this, the initial query is more complex, but you will be able to
+freely access the foreign key values without any additional queries,
+making things more efficient overall.
+
+Building on the ``ImportantPages`` example from the previous section, the
+following shows how ``select_related`` can be set to improve efficiency:
+
+.. code-block:: python
+    :emphasize-lines: 4,5
+
+    @register_setting
+    class ImportantPages(BaseSetting):
+
+        # Fetch these pages when looking up ImportantPages for or a site
+        select_related = ["donate_page", "sign_up_page"]
+
+        donate_page = models.ForeignKey(
+            'wagtailcore.Page', null=True, on_delete=models.SET_NULL, related_name='+')
+        sign_up_page = models.ForeignKey(
+            'wagtailcore.Page', null=True, on_delete=models.SET_NULL, related_name='+')
+
+        panels = [
+            PageChooserPanel('donate_page'),
+            PageChooserPanel('sign_up_page'),
+        ]
+
+With these additions, the following template code will now trigger
+a single database query instead of three (one to fetch the settings,
+and two more to fetch each page):
+
+.. code-block:: html
+
+    {{ settings.app_label.ImportantPages.donate_page.url }}
+    {{ settings.app_label.ImportantPages.sign_up_page.url }}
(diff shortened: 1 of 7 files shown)


docs/reference/contrib/settings.rst
```diff
@@ -108,17 +108,20 @@ Settings are designed to be used both in Python code, and in templates.
 Using in Python
 ---------------
 
-If access to a setting is required in the code, the :func:`~wagtail.contrib.settings.models.BaseSetting.for_site` method will retrieve the setting for the supplied site:
+If you require access to a setting in a view, the :func:`~wagtail.contrib.settings.models.BaseSetting.for_request` allows you to retrieve the relevant settings for the current request:
```
zerolab, line 111: Missing "method" before "allows you to...". Or, can be viewed as having an extra "the" in front of `:func:...`

docs/reference/contrib/settings.rst
```diff
@@ -219,3 +222,47 @@ Or, alternately, using the ``set`` tag:
...
     {% set social_settings=settings("app_label.SocialMediaSettings") %}
+
+
+Utilising ``select_related`` to improve efficiency
+--------------------------------------------------
+
+For models with foreign key relationships to other objects (e.g. pages),
+which are very often needed to output values in templates, you can set
+the ``select_related`` attribute on your model class to have Wagtail
+fetch the related data for those fields when looking up setting for a site.
```
zerolab, line 233: `setting` -> `settings` or `a setting`

docs/reference/contrib/settings.rst
```diff
@@ -219,3 +222,47 @@ Or, alternately, using the ``set`` tag:
...
+
+Doing this results in a more complex initial query, but means that further
```
zerolab, line 235: This reads overly complex. How about something like:
"By doing so, the initial query will be more complex, but it will reduce additional database queries when accessing the related field values." ?

And maybe link to Django's `select_related` docs for further reading?

docs/reference/contrib/settings.rst
```diff
@@ -219,3 +222,47 @@ Or, alternately, using the ``set`` tag:
...
+additional queries will not be required to access the field values,
+meaning fewer database queries overall.
+
+Reusing the ``ImportantPages`` example from the previous section, here
+is how ``select_related`` can be used to improve efficiency:
+
+.. code-block:: python
+    :emphasize-lines: 4,5
```
zerolab, line 243: 👍

docs/reference/contrib/settings.rst
```diff
@@ -219,3 +222,47 @@ Or, alternately, using the ``set`` tag:
...
+            'wagtailcore.Page', null=True, on_delete=models.SET_NULL, related_name='+')
+        sign_up_page = models.ForeignKey(
+            'wagtailcore.Page', null=True, on_delete=models.SET_NULL, related_name='+')
+
+        panels = [
+            PageChooserPanel('donate_page'),
+            PageChooserPanel('sign_up_page'),
+        ]
+
+With these additions, the following template code would now only
```
zerolab, line 261: "would now only result in a single database query" -> "will now generate a single database query"?

wagtail/contrib/settings/tests/test_model.py
```diff
@@ -0,0 +1,85 @@
...
+        request = self.get_request()
+
+        self._create_importantpages_object()
+
+        # force site query beforehand
+        Site.find_for_request(request)
+
+        # fetch settings and access foreiegn keys
+        with self.assertNumQueries(expected_queries):
+            settings = ImportantPages.for_request(request)
```
zerolab, line 72: should this not be `TestSetting.for_request()`?

ababic: I had to add the `ImportantPages` model specifically to test this feature, as `TestSetting` doesn't have any ForeignKey fields on it :(

wagtail/contrib/settings/models.py
```diff
@@ -13,16 +13,63 @@ class BaseSetting(models.Model):
...
+        return queryset
+
     @classmethod
     def for_site(cls, site):
         """
-        Get an instance of this setting for the site.
+        Get or create an instance of this setting for the site.
+        """
+        queryset = cls.base_queryset()
+        return queryset.get_or_create(site=site)[0]
```
zerolab, line 51: noticed this moves away from the `instance, created =  Model.get_or_create()` pattern. Would be good to be consistent. `get_or_create()[0]` is more cryptic

PR #5933, opened 2020-04-04
Provide a more convenient way to access page URLs from site settings

ababic:
Building on the progress in #5931 and #5932, this PR adds a `page_url` attribute to setting models, which provides a cleaner (and in many cases, more efficient) way to access URLs for pages referenced by settings.

- [x] Do the tests still pass? (https://docs.wagtail.io/en/latest/contributing/developing.html#testing)
- [x] Does the code comply with the style guide? (Run `make lint` from the Wagtail root)
- [x] For Python changes: Have you added tests to cover the new/fixed behaviour?
- [x] For new features: Has the documentation been updated accordingly?

PR #5934, opened 2020-04-04
[WIP] Fix LGTM Python complaints

ababic:
Starting with this list, which is causing unnecessary noise on a couple of PRs:
https://lgtm.com/projects/g/wagtail/wagtail/rev/pr-a1a8ef0f423e143c25c3b8034d11c506d66b17f9

Then I plan to look at the other alerts/warnings from the full list:
https://lgtm.com/projects/g/wagtail/wagtail/alerts/?mode=list


thibaudcolas:
I’ve done a bit of research on LGTM and,

- Posted a topic on the community forum to ask whether anyone ran into similar problems as the "new" alerts we got on #5933, https://discuss.lgtm.com/t/unclear-how-lgtm-determines-alerts-are-new/2783
- As far as I can see, posting more than one comment may be a bug – https://discuss.lgtm.com/t/dont-comment-on-every-push-to-a-wip-branch-or-pr/2589/3, https://discuss.lgtm.com/t/github-ci-status-instead-of-comments/1543/6

I also looked into why LGTM was reporting alerts in this PR for lines clearly marked with ignore comments. This is actually intentional, and ignore comments will work as expected once this is merged:

> As it stands, alerts are not suppressed in pull requests to avoid suppressions being introduced without reviewers noticing them… We are aware that this is not always clear and intuitive, so we’re working on improving the way this is communicated to the user.
> https://discuss.lgtm.com/t/broken-suppression-of-false-positives/2443

setup.py
```diff
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
-
-import sys
+import sys  # lgtm [py/unused-import]
```
thibaudcolas, line 2: Not sure I understand why we’re not just removing this? If it’s a similar type of hack to `multiprocessing` below we should add a comment.

wagtail/project_template/home/models.py
```diff
@@ -1,4 +1,4 @@
-from django.db import models
+from django.db import models  # lgtm [py/unused-import]
```
thibaudcolas, line 1: For the project template I think we should remove this so the template output passes basic linting, and we don’t risk littering other people’s codebases with ignore comments for LGTM.

gasman:
Closing as [lgtm.com is imminently being shut down](https://github.blog/2022-08-15-the-next-step-for-lgtm-com-github-code-scanning/). Some of this may be applicable to our current CodeQL setup (https://github.com/wagtail/wagtail/blob/main/.github/workflows/codeql-analysis.yml), but given the age of this PR it'll almost certainly be less work to start afresh.

PR #5935, opened 2020-04-05
Update imagerendition api alt

### LINKED ISSUE #5816: Have ImageRenditionField return ImageRendition alt property

The ImageRenditionField does not return the alt attribute in its [to_representation](https://github.com/wagtail/wagtail/blob/70da204eb1e8b2083e9a92af05ca895c2defa5dd/wagtail/images/api/fields.py#L31). Alt tags are useful for accessibility, and this feels like something that should be included in the "out of the box" solution Wagtail provides.

Ideally, I think ImageRenditionField could just return the [attrs_dict](https://github.com/wagtail/wagtail/blob/70da204eb1e8b2083e9a92af05ca895c2defa5dd/wagtail/images/models.py#L500) property of the AbstractRendition class. However, these have slightly differing keys - "src" vs "url", so that would be a breaking change.

A simpler alternative would be adding the "alt" key to the OrderedDict constructed by ImageRenditionField.

mail-liam:

Yes
Yes
See comment below
N/A
Yes

I was unable to locate any existing tests for the API output of `ImageRenditionField`. I'll add a check to the existing test if it exists. Otherwise, please let me know where such a test case would live so I can include it.

This PR aims to close #5816. An additional field is added to the `OrderedDict` constructed by `ImageRenditionField` over the alternative use of `image.attrs_dict` for backwards compatibility concerns. This may be a worthwhile change for a potential v3 of the wagtail API, however.


kaedroho:
Thanks @mail-liam!

This looks good to me, please could you add a test for it?

mail-liam:
Hi @kaedroho, thanks for the review.

I've added a testcase for the ImageRenditionField and had a stab at where it should live. Please let me know if you think there is a more appropriate location though as I'm not 100% certain where everything should sit in the project hierarchy.

lb-:
@mail-liam  - thanks for updating the tests, I could not see a better home for them myself so left as is.

Really appreciate you taking the time to submit this PR, great to see this improvement in place.

Merged in via 52cb7ab77b8e9c64b2426b04165dc031395c279c

Issue #5937, opened 2020-04-06
Page manager implements the `delete()` method


nimasmi:
### Issue Summary

Django's `QuerySet.delete` method has the attribute `queryset_only = True` [(source)][1]. This means that when `Manager.from_queryset(SomeQuerySet)` is used, the `delete()` method is not copied as a manager method [(source)][2].

Django [documents this behaviour as a sort of safety feature](https://docs.djangoproject.com/en/3.0/topics/db/queries/#deleting-objects).

Wagtail's `PageQuerySet.delete` method is originally defined on Treebeard's [MPNodeQuerySet][3], and not further modified in Wagtail's source. There the `queryset_only` attribute is not set.

This means that in Wagtail we can do:

```
>>>  Page.objects.delete()
[2020-01-29 00:00:02,000][951][INFO][wagtail.core] Page unpublished: "Root" id=1
[2020-01-29 00:00:02,000][951][INFO][wagtail.core] Page unpublished: "Home" id=3
[2020-01-29 00:00:02,000][951][INFO][wagtail.core] Page deleted: "Home" id=3
[2020-01-29 00:00:02,000][951][INFO][wagtail.core] Page deleted: "Root" id=1
```

whereas for any vanilla Django models, even within a Wagtail project, we get:

```
>>> MyModel.objects.delete()
*** AttributeError: 'Manager' object has no attribute 'delete'
```

[1]: https://github.com/django/django/blob/b9cf764be62e77b4777b3a75ec256f6209a57671/django/db/models/query.py#L667
[2]: https://github.com/django/django/blob/b9cf764be62e77b4777b3a75ec256f6209a57671/django/db/models/manager.py#L92
[3]: https://github.com/django-treebeard/django-treebeard/blob/master/treebeard/mp_tree.py#L72. 

I consider this a bug, because it deviates from the Django documentation, and is not itself documented.

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: yes

To fix this, either

- decide it's not a bug
- document it
- add `.queryset_only` attribute to the `MPNodeQuerySet.delete` method
- defer to Treebeard (I have opened this as https://github.com/django-treebeard/django-treebeard/issues/160)

### Technical details

* Django version: 2.2
* Wagtail version: 2.7

nimasmi:
[...]

PR #5938, opened 2020-04-07
Set TreeQueryset.delete.queryset_only = True

nimasmi:
Fixes wagtail/wagtail#5937

This reverts Wagtail's behaviour to match Django's, where an error is
raised as a safety mechanism.

Projects relying on the non-safe behaviour should update e.g.
`MyModel.objects.delete()` to `MyModel.objects.all().delete()`.

* Do the tests still pass? (https://docs.wagtail.io/en/latest/contributing/developing.html#testing) **Yes.**
* Does the code comply with the style guide? (Run `make lint` from the Wagtail root) **Yes.**
* For Python changes: Have you added tests to cover the new/fixed behaviour? **Yes.**


gasman:
Merged in 555adccc2a6ff22db63f5d943a2ded48688936cf

PR #5940, opened 2020-04-07
Workflow reports filtering (rebase of #5892)

gasman:


client/scss/components/_button-select.scss
```diff
@@ -0,0 +1,17 @@
+.button-select {
+    &__option {
+        display: block;
+        width: 100%;
+        margin-left: 0 !important;
```
gasman, line 5: TODO: fix reliance on `!important`, as it's disallowed by our linting rules

gasman:
TODO: update https://github.com/wagtail/wagtail/blob/master/docs/advanced_topics/adding_reports.rst to reflect the fact that base_report.html no longer provides a page listing by default (that's now in base_page_report.html).

wagtail/admin/views/reports.py
```diff
@@ -186,17 +186,6 @@ class ReportView(SpreadsheetExportMixin, TemplateResponseMixin, BaseListView):
...
-        "latest_revision_created_at": _("Updated"),
-        "status_string": _("Status"),
-        "content_type.model_class._meta.verbose_name.title": _("Type"),
-    }
-    list_export = [
-        "title",
-        "latest_revision_created_at",
-        "status_string",
-        "content_type.model_class._meta.verbose_name.title",
-    ]
```
gasman, line 199: This is another breaking change - the 'adding reports' example in the docs expects these to be defined on ReportView. Perhaps we need an intermediate `PageReportView` that both the docs example and LockedPagesView inherit from?

wagtail/admin/filters.py
```diff
@@ -67,6 +68,128 @@ def decompress(self, value):
...
+
+            subgroup = []
+            if isinstance(option_label, (list, tuple)):
+                group_name = option_value
+                subindex = 0
+                choices = option_label
+            else:
+                group_name = None
+                subindex = None
+                choices = [(option_value, option_label)]
```
jacobtoppm, line 119: It looks like sub-choices can't have a separate filter value to the rest of the group. Is this intentional then?

wagtail/admin/filters.py
```diff
@@ -67,6 +68,128 @@ def decompress(self, value):
...
+        self.widget.filter_field = filter_field
+
+    def get_filter_value(self, obj):
+        # filter_accessor identifies a property or method on the instances being listed here,
+        # which gives us a queryset of related objects. Turn this queryset into a list of IDs
+        # that will become the 'data-filter-value' used to filter this listing
+        queryset = getattr(obj, self.filter_accessor)
+        if callable(queryset):
+            queryset = queryset()
+
```
jacobtoppm, line 184: This limits it to methods/attributes of the instance itself, as far as I can see? In this case that works really well, I think it's a good addition to Task, but would it be worth adding a way to specify a queryset/filter/function more generally here, that just takes the instance as an argument? Eg that would let me specify the `Workflow.objects.filter(workflow_tasks__task=instance)`?

wagtail/admin/templates/wagtailadmin/reports/workflow_tasks.html
```diff
@@ -1,74 +1,60 @@
...
+        <a href="{% url 'wagtailadmin_reports:workflow' %}" class="button">By Workflow</a>
+    </div>
+    {{ block.super }}
+{% endblock %}
 
-    <div class="nice-padding">
-        {% if object_list %}
-            <table class="listing">
-                <thead>
+{% block results %}
```
jacobtoppm, line 11: This report doesn't seem to be paginated (unless I'm missing something) - I think that would be worth adding if not.

jacobtoppm: (and I think same with the other)

wagtail/admin/filters.py
gasman: I didn't consider how this would interact with grouped options - the logic for dealing with them is only here because the base Select implementation funnels everything through the `optgroups` method, and I wanted to make the changes to the original method as minimal as possible. As it stands, I think you're right - the effect of that is that we only accept a filter_value at the top level of `choices` (i.e. an optgroup or ungrouped choice). Can look at supporting filter_value within optgroups too, if you think that's a useful addition?

wagtail/admin/templates/wagtailadmin/reports/workflow_tasks.html
gasman: Pagination is handled by Django's `BaseListView`, which `ReportView` inherits from:

https://github.com/django/django/blob/master/django/views/generic/list.py#L118

jacobtoppm: Sorry, I should have been clearer - I meant frontend pagination tags like in `BasePageReport`, eg https://github.com/wagtail/wagtail/blob/860588c80763059c75abe7512dd6be597abed4b3/wagtail/admin/templates/wagtailadmin/reports/base_page_report.html

wagtail/admin/filters.py
jacobtoppm: I don't think it's a massively important feature, but it would be nice to support it if it's not too much effort, just for continuity. Otherwise I'd suggest just making a note in comment or docstring that this isn't currently supported, as I can see it being a bit confusing to narrow that down - I'd be happy with either!

wagtail/admin/templates/wagtailadmin/reports/workflow_tasks.html
gasman: Ah, I see, yep - that `{% paginate %}` tag probably ought to be moved out into base_report...

wagtail/admin/filters.py
gasman: Makes sense, yep - it wouldn't be a massive leap to allow passing a callable for `filter_accessor`, so that you could write something like:

    task = FilteredModelChoiceFilter(
        queryset=Task.objects.all(),
        filter_field='id_workflow',
        filter_accessor=(lambda task:Workflow.objects.filter(workflow_tasks__task=task))
    )

At this point I think the harder task is going to be documenting what's going on, since this is quickly outgrowing "a bit of supporting code to make this dropdown work" into a reusable component (and, quite possibly, its own mini-library outside of Wagtail...)

jacobtoppm: True, it is growing! I think having it reusable would be great (but yeah, documenting it somewhere does sound good). Possibly the documentation section for reports filtering, once we make that public?

wagtail/admin/templates/wagtailadmin/reports/workflow_tasks.html
gasman: Fixed in 996d4806f1a9f07d7691710414c30952be01016a

wagtail/admin/filters.py
gasman: optgroup support added in 3f18f0f8a2aa4c1f6859ae1adfabde5cdf90423d

wagtail/admin/filters.py
gasman: Now accepts relations and arbitrary callables as of caa81cc807b4ff65de35038a1bb11d3cc7510ab1.

PR #5942, opened 2020-04-08
Suggeted new COMMANDS for "bin/wagtail.py" tool.

ahke-sied:
COMMANDS = {
    'start': CreateProject(),
    'updatemodulepaths': UpdateModulePaths(),
    # add new commands, why not?
    #'createsuperuser':like_CreateSuperUser(), #execute `python manage.py createsuperuser`
    #'runserver':like_RunServer(), #execute `python manage.py runserver`
    #'migrate':like_Migrate(), #execute `python manage.py migrate`
}
This is just a suggestion and has no effects on the code. I'm sorry for not have implementing my idea because I just get started discovering this awesome framework.

### DIFF
--- a/wagtail/bin/wagtail.py
+++ b/wagtail/bin/wagtail.py
@@ -245,6 +245,10 @@ class UpdateModulePaths(Command):
 COMMANDS = {
     'start': CreateProject(),
     'updatemodulepaths': UpdateModulePaths(),
+    # add new commands, why not?
+    #'createsuperuser':like_CreateSuperUser(), #execute `python manage.py createsuperuser`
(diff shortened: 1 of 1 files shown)


gasman:
Hi @ahke-sied - thanks for the suggestion!

I think it's important for people to understand that the normal development practices from Django are still valid in Wagtail, and for that reason I'd say it's best that we don't duplicate Django's standard commands in Wagtail. For example: if someone is adding Wagtail to an existing Django project, and all the documentation tells them to use `wagtail runserver`, they might assume that Wagtail is doing something non-standard and that they have to update their existing processes that use ./manage.py.

PR #5944, opened 2020-04-09
Raise 404 if failing to find specified page or task for previewing during moderation

jacobtoppm:
Previously, `get` was used to fetch Pages and Tasks for the preview view, without catching errors. This should probably raise a 404 similarly to the existing `preview_for_moderation` view, using `get_object_or_404`.


gasman:
Makes sense 👍 Merged in fdc5b9fc9cc10143bfbc245e0703e5117a6b3bd7.

PR #5946, opened 2020-04-10
Remove sticky footer on small devices

### LINKED ISSUE #4641: Edit pages’ sticky footer hides critical parts of the UI
### Issue Summary

You can't add new streamfield blocks at the bottom of a edit page on mobile, and this hides various parts of the UI at other screen sizes.

### Steps to Reproduce

1. On a IOS device open a browser and go to an edit page in the admin.
2. Make sure the last field on that page is a streamfield. Containing a block.
3. Try to add a block below. You can't. The `+` button is behind the save buttons.

I expect the save buttons to go below the form/content. 
I also wonder why the save buttons are always visible. They take up a lot op space. Up to 1/4 of the screen on smaller devices.

In this example the 'Add another' button - for a list stream field item - is also below the save buttons.
iphone_6s_-_11_4

### Technical details

* Wagtail version: 2.1
* Browser version: IOS Safari 11.4

saeed617:
The default layout is the mobile one and “up from mobile” gets the overlap added.

Result:
localhost_8000_admin_pages_34_edit_(iPhone 6_7_8)
localhost_8000_admin_pages_34_edit_(iPhone 6_7_8) (1)

Fixes #4641

### DIFF
--- a/client/scss/components/_footer.scss
+++ b/client/scss/components/_footer.scss
@@ -4,12 +4,8 @@ footer {
     border-radius: 3px 3px 0 0;
     box-shadow: 0 0 2px rgba(255, 255, 255, 0.5);
     background: $color-grey-1;
-    position: fixed;
-    bottom: 0;
-    padding: 0.5em;
-    width: 90%;
-    margin: 0 5%;
     color: $color-white;
+    margin-top: $mobile-nice-padding;
 
     ul {
         @include unlist();
@@ -72,5 +68,8 @@ footer {
         margin-left: $desktop-nice-padding;
         margin-right: $desktop-nice-padding;
         width: calc(100% - #{$menu-width} - #{2 * $desktop-nice-padding});
+        position: fixed;
+        bottom: 0;
+        padding: 0.5em;
     }
 }
--- a/client/scss/components/_grid.legacy.scss
+++ b/client/scss/components/_grid.legacy.scss
@@ -19,8 +19,11 @@
     background: $color-white;
     border-top: 0 solid $color-grey-5; // this top border provides space for the floating logo to toggle the menu
     min-height: 100%;
-    padding-bottom: 4em;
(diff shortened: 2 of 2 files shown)


lb-:
@saeed617 thanks for submitting this, along with screenshots also!

@thibaudcolas - I tested this locally, works as expected with the footer now at the bottom, will leave you to review in full.

My only thoughts are that the user might need to scroll a fair bit on a long page to get to the save button. I also agree that it can be a bit full on having the entire footer showing at all times.

Ideas for refinement
* keep the action buttons 'sticky' but put the last modified/revisions text at the bottom
* maybe try to fit two buttons in on mobile side by side, I think this is doable but that is assuming an english translation with the words being small and keep them sticky
* harder to do - make the actions on mobile ONE dropdown (drop up on mobile)  and keep them sticky

saeed617:
What about this style? I think in this way it's not on screen all the times and you don't need to scroll down the page to reach buttons and with a simple scroll up you have them.  I saw this in many places and users are already familiar with it.

wagtail

lb-:
@saeed617 - I personally am not a fan of that style of footer (or header) as it is prone to issues and a sense of fiddly-ness. I thought maybe something akin to the wagtail user bar (smaller button, always there (sticky) but clicking it opens up a menu with options.

This means two clicks to save though, and also resorts to icons to represent things instead of text.

I probably would defer to some of the other Wagtail team for ideas though.

 @wagtail/ux-design ?

lb-:
@saeed617 thanks again for your work on this, we discussed this at the core meeting and agreed that your initial solution is a good way forward.

We can add improvements as we go in the future, one idea was to add a 'go to bottom' type link that floats on mobile view.

I will take another look at this soon and do another review.

lb-:
Merged in via 44c1f1cab2bdbc4bec95d62a876bd87e279d7ac3

PR #5948, opened 2020-04-12
Fix typo in translation file

### LINKED ISSUE #5947: Typo in translation file
In file `wagtail/admin/locale/ru/LC_MESSAGES/django.po` on line `361` should be "Воскресенье" instead of "Воскрксенье"

toptalo:
Close #5947


zerolab:
Thank you for this @toptalo. Would you be so kind to submit these changes through https://www.transifex.com/torchbox/wagtail/ , so that they're kept in sync with other translation updates?

Note that this is also referenced in the docs: https://docs.wagtail.io/en/v2.8/contributing/index.html#translations

PR #5949, opened 2020-04-13
Add redirects_importer module

marteinn:
Hi there! I was asked by Tom Dyson and the core team to submit a PR where [wagtail-redirect-importer](https://github.com/Frojd/wagtail-redirect-importer/) is included by default in Wagtail, so here we go :)

First, to answer the contribution questions:
- [x] For front-end changes: Did you test on all of Wagtail’s supported browsers?
    - Firefox 75.0b6 (Mac)
    - Chrome 80.0.3987.163 (Mac)
    - Chrome 80.0.3987.162 (Chrome OS)
    - Chrome 80.0.3987.149 (Android 8.1)
    - IE 11 (Windows 8.1)
    - Edge 80 (Windows 8.1)
    - Safari 13.1 (13609.1.20.111.8) (Mac)
    - Safari 13 (iOS 13.3)
- [ ] For new features: Has the documentation been updated accordingly? **Here I need input, see my comments below**

Before submitting this PR I’ve done some preparation work that are described here: https://github.com/Frojd/wagtail-redirect-importer/issues/6 (TLDR; Improved tests coverage, offset/limit feature, bug squashing)

How this PR differs from the original wagtail-redirect-importer:
- I’ve renamed the app to redirects_importer (this also applies to templates and static)
- Changed path to wagtail.contrib.redirects_importer
- Changed url namespace to wagtailredirectsimporter
- Fixed a bug where the error section was always visible even if there were no errors
- Fixed a bug on the summary page (that requires POST data), if reloaded it now redirects the user back to import redirect start

A couple of things to note:
- I’ve kept wagtail_redirect_importer as a separate app
- This PR introduces a new dependency (tablib with xls and xlsx support)

Things I need help with / broader questions:
- Improving docs. This is a new feature, but are currently unsure what kind of info we need to add. Any pointer in the right direction is appreciated. Will probably need to point out how to activate the app.
- (Core team advice) Today redirect importer has its own menu item, but interface wise it might be better to see a “Import redirects” besides “Add redirect” but there are some limitations in the templating system that only allows one action button. Any thoughts on this? This can of course be done is a separate PR.
- (Core team advice) Keeping it a separate app / merging into redirects? Its separate right now because it's easier for me to maintain, but architecture wise it might be good to have one app that handles all redirect related matters. Could also be done in a different PR.
- The app has its own demoapp as a test harness, any pointers how the wagtail project looks at constructing test environments for apps or is it an ok practice within the project?

Anyhow, let's start here and feedback is much appreciated!


jacobtoppm:
Thanks for making this PR - it'll be great to get this functionality in Wagtail!
I would lean towards bringing it in with redirects as much as possible, both app and menu item/buttons - for modeladmin xlsx/csv downloads in 2.9, we're adding a download button next to the add button, so I think adding an import button for redirects would probably be a good call.

gasman:
Agreed - I can't see a situation where someone would have redirects enabled in their project and _not_ welcome this option being available, so from an end-user perspective it would make sense for this to be part of the standard redirects module rather than an add-on. (It'll also make it less fiddly to get the UI more tightly integrated - no need to come up with a way to 'hook in' extra action buttons from outside the app.)

marteinn:
Thanks for the feedback @gasman and @jacobtoppm! My next step is to rebase this PR to include Wagtail 2.9 and find a good merge of redirects and redirects-importer. Will ping you once I have a review ready.

marteinn:
Hello! Just pushed out a new update, I’ve taken the feedback provided into account and unified the importer with the redirects app.

Design decisions I’ve made:
- Moved the import view functions into `wagtail.contrib.wagtailredirects.views.py`
- Moved and renamed `wagtail.contrib.wagtailredirects.tests` into `wagtail.contrib.wagtailredirects.views.test_redirects.py`. I also added the remaining tests from redirect_importer into this dir.
- Moved import view functions into `wagtail.contrib.wagtailredirects.views`
- Changed the url schema so the importer lives at /redirects/import/*
- Added a button titled “Import redirects” in redirects index header
- Inlined header code from header partial into redirects index to make it possible to add the additional “Import redirects” button (I saw the same pattern in `modeladmin`)
- Changed “i’m done” and cancel buttons to link to wagtailredirects:index
- Renamed templates and views to clarify they manage import (choose_file.html -> choose_import_file and start -> start_import)

Small fixes:
- Fixed a couple of missing translations
- Error summary was always showing (even if there were no errors)
- Lots of renaming to better fit the redirects app conventions

Advice needed
- I’ve added the button “Import redirects” in the header, but design wise it does not look that great (no margin/padding - see my embedded screenshot). I tried @jacobtoppm example for the download button, but it generated the same layout. My question is, should the buttons be stacked horizontally or vertically and any advice on design and/or helper css classes I could look at?

Looking forward to getting additional feedback!

Screenshot 2020-05-10 at 09 47 00

lb-:
FYI @marteinn there is now a docs page for the contrib.redirects module in master.
https://github.com/wagtail/wagtail/blob/master/docs/reference/contrib/redirects.rst

wagtail/contrib/redirects/templates/wagtailredirects/index.html
```diff
@@ -15,10 +15,38 @@
 
 {% block content %}
     {% trans "Redirects" as redirects_str %}
-    {% trans "Add redirect" as add_str %}
     {% if user_can_add %}
         {% url "wagtailredirects:add" as add_link %}
-        {% include "wagtailadmin/shared/header.html" with title=redirects_str icon="redirect" action_url=add_link action_text=add_str search_url="wagtailredirects:index" %}
+        {% trans "Add redirect" as add_str %}
+        {% url "wagtailredirects:start_import" as import_link %}
+        {% trans "Import redirect" as import_str %}
```
tomdyson, line 22: trans "Import redirect" -> trans "Import redirects"

tomdyson:
Thanks @marteinn! I agree that the vertical stacking without margin doesn't look right. Paging @benenright for his opinion on a better approach.

benenright:
Hiya

I'd go for something like this.

Screenshot 2020-05-11 at 11 08 40

Ben

wagtail/contrib/redirects/templates/wagtailredirects/index.html
marteinn: Well spotted :)

marteinn:
Thanks @lb-! I went through the docs and added documentation for the `import_redirects` management command. I also started writing a guide for how to import redirects from the interface, but realized its better suited in the editors guide, but in the end decided not to for now, (I think its better to do it when this PR is merged and interface is finalized so I can add proper screenshots).

marteinn:
Thanks @benenright and @tomdyson for the input! I have updated the button placement and added a css fix for this particular module, it can be found here: https://github.com/wagtail/wagtail/blob/7e1a9a02623dcb50406d253eeffaed6977b39d51/wagtail/contrib/redirects/static/wagtailredirects/css/index.css#L7-L17

I have tested it on a couple of different browsers and seems to be working fine:
- Firefox 77.0b1 (64-bit) (MacOS)
- Version 81.0.4044.138 (Official Build) (64-bit) (MacOS)
- Safari Version 13.1 (MacOS)
- IE 11 (Windows 10)

I’m not 100% happy with this solution though, I would rather find a solution that resides in `_header.scss` and can be applied to all headers that contain multiple action buttons. But I think that is a separate issue. (Perhaps already is?)

marteinn:
I’ve continued to work on this PR and I think it's ready for review, some docs are missing but I’m planning on doing those once the code has been reviewed.

In my latest set of updates I have done the following:

- Solved issue https://github.com/Frojd/wagtail-redirect-importer/issues/5 reported by @zerolab where the admin import did not work on horizontally scaled apps. I solved it by introducing the storage method “cache”, which utilizes the django cache to keep the state between servers.
- Added the setting `WAGTAIL_REDIRECTS_FILE_STORAGE` so horizontally scaled apps can change storage method, the default method is still using a temporary file.
- Fixed so only the file formats csv, tsv, xls and xlsx are allowed, mainly to make it easier for us when dealing with support issues and debugging.

I’m marking this as needs review. Cheers!

wagtail/contrib/redirects/base_formats.py
```diff
@@ -0,0 +1,228 @@
...
+        dataset.headers = [cell.value for cell in next(rows)]
+
+        for row in rows:
+            row_values = [cell.value for cell in row]
+            dataset.append(row_values)
+        return dataset
+
+
+#: These are the default formats for import and export. Whether they can be
+#: used or not is depending on their implementation in the tablib library.
```
lb-, line 218: This comment references export, is this just for future or is there a way to export (I could not see it)

wagtail/contrib/redirects/views.py
```diff
@@ -126,3 +133,183 @@ def add(request):
...
+
+    try:
+        data = file_storage.read(input_format.get_read_mode())
+        if not input_format.is_binary() and from_encoding:
+            data = force_str(data, from_encoding)
+        dataset = input_format.create_dataset(data)
+    except UnicodeDecodeError as e:
+        return HttpResponse(_(u"<h1>Imported file has a wrong encoding: %s</h1>" % e))
+    except Exception as e:  # pragma: no cover
+        return HttpResponse(
```
lb-, line 188: Is it possible for both of these error responses to use the response.messages system and land you back on the import page with the errors in red at the top of the page? Being shown these full page errors, outside of the Wagtail interface is a bit heavy.

wagtail/contrib/redirects/views.py
```diff
@@ -126,3 +133,183 @@ def add(request):
...
+            "redirect_link": to_link,
+            "is_permanent": config["permanent"],
+        }
+
+        if config["site"]:
+            data["site"] = config["site"].pk
+
+        form = RedirectForm(data)
+        if not form.is_valid():
+            error = form.errors.as_text().replace("\n", "")
```
lb-, line 303: These render as a row of errors with the `*` character in front.
e.g.
> * redirect_link * Enter a valid URL.* __all__ * A redirect with this path already exists.

It would be great if this could be slightly nicer to read, maybe pull out the leading `*` and join with a comma. so the output would be more like.
> redirect_link, Enter a valid URL., __all__, A redirect with this path already exists.

More could probably done to avoid these being too cryptic but I think this is much better than saying nothing.

wagtail/contrib/redirects/templates/wagtailredirects/choose_import_file.html
```diff
@@ -0,0 +1,43 @@
...
+            targetOutput: "#redirects-results"
+        }
+    </script>
+{% endblock %}
+
+{% block content %}
+    {% trans "Import redirects" as import_str %}
+    {% include "wagtailadmin/shared/header.html" with title=import_str icon="redirect" %}
+
+    {% if form.non_field_errors %}
```
lb-, line 20: I think this form, on initial load, needs a little more info for the user.

It is unclear that you will be taken to a next step that will let you validate/approve/ see errors, it is also unclear that the format used can be any header name (allocatable later).

I know that there are docs underway, however there could be a little more help given to the user.

One idea would be to add a help block here - for example:
```
    <div class="nice-padding">
        <div class="help-block help-info">
            {% blocktrans %}
                <p>Select a file that contains a column with header <pre>from</pre> and <pre>to</pre>, with each row being one redirect.</p>
                <p>... file formats are available.</p>
            {% endblocktrans %}
        </div>
    </div>
```

Alternatively, adding some help text to the import file field might be enough, something along the lines of 'once submitted, you will be given the change to select the from/to mapping and confirm before adding'.

wagtail/contrib/redirects/templates/wagtailredirects/choose_import_file.html
```diff
@@ -0,0 +1,43 @@
...
+            url: "{% url 'wagtailredirects:index' %}",
+            termInput: "#id_q",
+            targetOutput: "#redirects-results"
+        }
+    </script>
+{% endblock %}
+
+{% block content %}
+    {% trans "Import redirects" as import_str %}
+    {% include "wagtailadmin/shared/header.html" with title=import_str icon="redirect" %}
```
lb-, line 18: I think each page in this imports redirect flow should have the title AND the subtitle to help guide the user through.

e.g.
```
    {% trans "Import redirects" as header_title %}
    {% trans "Choose a file" as header_subtitle %}
    {% include "wagtailadmin/shared/header.html" with title=header_title subtitle=header_subtitle icon="redirect" %}
```

wagtail/contrib/redirects/templates/wagtailredirects/confirm_import.html
```diff
@@ -0,0 +1,55 @@
+{% extends "wagtailadmin/base.html" %}
+{% load i18n %}
+
+{% block titletag %}{% trans "Confirm import" %}{% endblock %}
+{% block content %}
+    {% trans "Confirm import" as confirm_str %}
+    {% include "wagtailadmin/shared/header.html" with title=confirm_str icon="redirect" %}
```
lb-, line 7: Similar to the other headers, having a consistent main header and adding a subtitle may provide a bit more of a unified feel to this workflow.

```
    {% trans "Import redirects" as header_title %}
    {% trans "Confirm import" as header_subtitle %}
    {% include "wagtailadmin/shared/header.html" with title=header_title subtitle=header_subtitle icon="redirect" %}
```

wagtail/contrib/redirects/templates/wagtailredirects/import_summary.html
```diff
@@ -0,0 +1,40 @@
+{% extends "wagtailadmin/base.html" %}
+{% load i18n %}
+{% block titletag %}{% trans "Summary" %}{% endblock %}
+{% block content %}
+    {% trans "Summary" as summary_str %}
+    {% include "wagtailadmin/shared/header.html" with title=summary_str icon="redirect" %}
```
lb-, line 6: Similar to the other headers, having a consistent main header and adding a subtitle may provide a bit more of a unified feel to this workflow.

```
    {% trans "Import redirects" as header_title %}
    {% trans "Summary" as header_subtitle %}
    {% include "wagtailadmin/shared/header.html" with title=header_title subtitle=header_subtitle icon="redirect" %}
```

wagtail/contrib/redirects/base_formats.py
```diff
@@ -0,0 +1,228 @@
...
+DEFAULT_FORMATS = [fmt for fmt in (
+    CSV,
+    XLS,
+    XLSX,
+    TSV,
+    ODS,
```
lb-, line 224: I did not see `ODS` listed in the UI, how come this is excluded?
It would be nice to see this format available, but not a big problem if not.

wagtail/contrib/redirects/templates/wagtailredirects/import_summary.html
```diff
@@ -0,0 +1,40 @@
...
+    <div class="nice-padding">
+        <section id="summary">
+            <h2>{% trans "Summary" %}</h2>
+            <h3>{% blocktrans with total=import_summary.total successes=import_summary.successes errors=import_summary.errors_count %}Found {{ total }} redirects, created {{ successes }} and found {{ errors }} errors.{% endblocktrans %}</h3>
```
lb-, line 10: I am not sure that `h3` is appropriate here, this is not really a heading but more a summary of a success state.

on top of that, h2 is now a duplicate of the page header (or subtitle if used as per suggested).

A potential rework of this would be to use the `<p class="help-block help-info">` or potentially the response.messages to show this as successful.

I know we need a page here to show the potential errors, but maybe if there is NO errors, we could go back to the redirects index listing and show the success as a messages and only show this summary page if there were issues with the import.

wagtail/contrib/redirects/templates/wagtailredirects/import_summary.html
```diff
@@ -0,0 +1,40 @@
...
+
+            <a href="{% url 'wagtailredirects:index' %}" class="button">{% trans "Continue" %}</a>
```
lb-, line 12: I think the continue button should show at the bottom, this is in line with many other final actions on a page in Wagtail.

Having continue last, after the errors, means it will hopefully be more likely for users to read the errors before clicking the 'continue' button.

marteinn:
Thank you so much for the feedback @lb, much appreciated! Starting to address your comments now.

wagtail/contrib/redirects/base_formats.py
marteinn: @lb- So this module is a special case, since its actually from another Django library and I've tried to not to make any changes to it, mostly to make updating easier. Would you be ok keeping the comment based on my reasoning?

wagtail/contrib/redirects/views.py
marteinn: @lb- Good point, here I hijack the form validation and it's not always pretty to the user (especially the `__all__`key bit). Will apply some string parsing to enhance user experience.

wagtail/contrib/redirects/base_formats.py
marteinn: @lb- I excluded a couple of formats like `ODS` (but also `json` and `yaml`) for a couple of reasons. I skipped `ODS` because I hard a hard time finding tooling to construct `ODS` files and that blocked me from writing the necessary tests. I would love to have more formats in the future. Right now I restricted it to the common formats csv, tsv, xls and xlsx.

wagtail/contrib/redirects/views.py
marteinn: @lb- Good point! Its definitely doable. Will solve in my next update.

wagtail/contrib/redirects/forms.py
```diff
@@ -42,3 +44,51 @@ def clean(self):
     class Meta:
         model = Redirect
         fields = ('old_path', 'site', 'is_permanent', 'redirect_page', 'redirect_link')
+
+
+class ImportForm(forms.Form):
+    import_file = forms.FileField(label=_("File to import"))
+    input_format = forms.ChoiceField(label=_("Format"), choices=[],)
```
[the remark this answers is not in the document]
marteinn: @lb- No problem at all, I actually do this in the cli, so this would make it more consistent. I'm considering if we should keep the `input_format` field, but make it optional so we can support cases where the user wants to override the file format. Any thoughts on this?

wagtail/contrib/redirects/base_formats.py
lb-: Oh yep. I can see that now. 100% agree. 
Also, we might even want to bring this util into core eventually and use for similar shared import/export functionality. For now - this is great, leave as is.

wagtail/contrib/redirects/base_formats.py
lb-: Great. Makes sense. Leave as is.

wagtail/contrib/redirects/forms.py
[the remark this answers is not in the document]
lb-: I think less fields is better, if the user wants to upload a CSV with the extension .tsv or a .txt - I think that is a really small edge case and most average users would not even understand that extensions can be anything.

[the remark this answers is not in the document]
lb-: One additional thought - It would be good to restrict the file extensions in the field upload widget to what is supported only. 

Similar to this - https://stackoverflow.com/a/43737771/8070948

wagtail/contrib/redirects/views.py
lb-: Also, not sure if response.messages is the best, maybe non-field form errors.

wagtail/contrib/redirects/views.py
```diff
@@ -126,3 +133,183 @@ def add(request):
...
     })
+
+
+@permission_checker.require_any("add")
+def start_import(request):
+    from_encoding = "utf-8"
+    query_string = request.GET.get('q', "")
+
+    if not request.POST:
+        SUPPORTED_FORMATS = ["CSV", "TSV", "XLS", "XLSX"]
```
lb-, line 144: Another suggestion - as this is the canonical source for what formats the user can upload, it might be best to move this to a class property or static method if possible. 

As this is a function view, maybe on the redirect model or just a global in this file. 

Just somewhere central, I think it will come in handy having this outside of a branch within the view.

wagtail/contrib/redirects/views.py
```diff
@@ -126,3 +133,183 @@ def add(request):
...
+                'search_form': SearchForm(
+                    data=dict(q=query_string) if query_string else None, placeholder=_("Search redirects")
+                ),
+                "form": ImportForm(accepted_formats),
+            },
+        )
+
+    form_kwargs = {}
+    form = ImportForm(
+        DEFAULT_FORMATS, request.POST or None, request.FILES or None, **form_kwargs
```
lb-, line 161: Similar here - having a central source of truth in this file of what formats will be supported vs which formats are 'default' could make more sense. 

Again, I may be missing something but at first glance it looks like we are getting the supported formats in separate ways throughout this view.

wagtail/contrib/redirects/forms.py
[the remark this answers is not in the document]
marteinn: Awesome, then we drop the field and use a customized upload widget instead.

wagtail/contrib/redirects/views.py
marteinn: Agree! I'll make it a module property.

wagtail/contrib/redirects/management/commands/import_redirects.py
```diff
@@ -0,0 +1,182 @@
...
+        )
+        parser.add_argument(
+            "--ask", help="Ask before creating", default=False, type=bool,
+        )
+
+        available_formats = [key for key in registry._formats]
+        parser.add_argument(
+            "--format",
+            help="Source file format (.csv, .xls etc)",
+            choices=available_formats,
```
lb-, line 49: The available formats here are ALL formats that the parse util uses, not the subset that the UI interface uses. 

I recommend that the subset be somewhere central (as per comments on the view function), as this CLI format support should match the UI, as we have only tested the sub-set of supported formats.

wagtail/contrib/redirects/management/commands/import_redirects.py
```diff
@@ -0,0 +1,182 @@
...
+            "--from_index",
+            help="The column where to read from link",
+            default=0,
+            type=int,
+        )
+        parser.add_argument(
+            "--to_index", help="The column where to read to link", default=1, type=int,
+        )
+        parser.add_argument(
+            "--dry_run",
```
lb-, line 36: Not sure on the best way forward here, but it appears that Django commands all use `kebab-case` for their commands, so `--dry-run` instead of `--dry_run`. However, after digging into other Wagtail commands, they appear to use `underscore_case`.

This might cause confusion for Django users, as `--dry-run` seems a pretty standard usage as kebab case across other commands. 

e.g. https://docs.djangoproject.com/en/3.0/ref/django-admin/#cmdoption-makemigrations-dry-run

wagtail/contrib/redirects/management/commands/import_redirects.py
```diff
@@ -0,0 +1,182 @@
...
+            "--src", help="Path to file", type=str, required=True,
+        )
+        parser.add_argument(
+            "--site_id", help="The site where redirects will be associated", type=int,
+        )
+        parser.add_argument(
+            "--permanent",
+            help="Save redirects as permanent redirects",
+            type=bool,
+            default=True,
```
lb-, line 24: I recommend that all bool args be converted to optional args.

So instead of having to type
`python manage.py import_redirects --src ./Untitled\ spreadsheet\(1\).xlsx --permanent TRUE --dry_run TRUE`

You can write 
`python manage.py import_redirects --src ./Untitled\ spreadsheet\(1\).xlsx --permanent --dry_run`

Simply excluding the arg will be the false scenario, however as permanent is default true it might make sense to leave this as an arg. Unless the arg becomes `--temporary`, where not including that arg will make the redirect permanent. Your call, as it is probably best for the user to be explicit here. `dry_run` should be an optional arg though.

see - https://docs.djangoproject.com/en/3.0/howto/custom-management-commands/#accepting-optional-arguments

wagtail/contrib/redirects/management/commands/import_redirects.py
```diff
@@ -0,0 +1,182 @@
...
+            "--to_index", help="The column where to read to link", default=1, type=int,
+        )
+        parser.add_argument(
+            "--dry_run",
+            default=False,
+            help="Run only in test mode, will not create redirects",
+            type=bool,
+        )
+        parser.add_argument(
+            "--ask", help="Ask before creating", default=False, type=bool,
```
lb-, line 42: Same here, this could be an optional arg, meaning you can type `--ask` instead of `--ask TRUE`

https://docs.djangoproject.com/en/3.0/howto/custom-management-commands/#accepting-optional-arguments

wagtail/contrib/redirects/management/commands/import_redirects.py
```diff
@@ -0,0 +1,182 @@
...
+            help="Source file format (.csv, .xls etc)",
+            choices=available_formats,
+            type=str,
+        )
+
+        parser.add_argument(
+            "--offset", help="Import starting with index", type=int, default=-1
+        )
+        parser.add_argument(
+            "--limit", help="Limit import to num items", type=int, default=-1
```
lb-, line 57: the default here might be confusing, if the arg is not defined, does that mean that only the first row will be included.

Maybe the default could be `None` and if `None`, all lines will be imported.

wagtail/contrib/redirects/management/commands/import_redirects.py
```diff
@@ -0,0 +1,182 @@
...
+            "--site_id", help="The site where redirects will be associated", type=int,
+        )
+        parser.add_argument(
+            "--permanent",
+            help="Save redirects as permanent redirects",
+            type=bool,
+            default=True,
+        )
+        parser.add_argument(
+            "--from_index",
```
lb-, line 27: for simplicity, and to avoid further confusion about kebab/underscore case args, maybe change this arg to simply `--from`

wagtail/contrib/redirects/management/commands/import_redirects.py
```diff
@@ -0,0 +1,182 @@
...
+
+class Command(BaseCommand):
+    help = "Imports redirects from .csv, .xls, .xlsx"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--src", help="Path to file", type=str, required=True,
+        )
+        parser.add_argument(
+            "--site_id", help="The site where redirects will be associated", type=int,
```
lb-, line 18: for simplicity, and to avoid further confusion about kebab/underscore case args, maybe change this arg to simply `--site`

wagtail/contrib/redirects/management/commands/import_redirects.py
```diff
@@ -0,0 +1,182 @@
...
+            default=True,
+        )
+        parser.add_argument(
+            "--from_index",
+            help="The column where to read from link",
+            default=0,
+            type=int,
+        )
+        parser.add_argument(
+            "--to_index", help="The column where to read to link", default=1, type=int,
```
lb-, line 33: for simplicity, and to avoid further confusion about kebab/underscore case args, maybe change this arg to simply `--to`

wagtail/contrib/redirects/views.py
```diff
@@ -126,3 +133,183 @@ def add(request):
...
+        ]
+        return render(
+            request,
+            "wagtailredirects/choose_import_file.html",
+            {
+                'search_form': SearchForm(
+                    data=dict(q=query_string) if query_string else None, placeholder=_("Search redirects")
+                ),
+                "form": ImportForm(accepted_formats),
+            },
```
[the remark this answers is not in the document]
marteinn: Good feedback, there are some state passing between the views which makes the view harder to read. Will do another pass and try to simplify.

wagtail/contrib/redirects/management/commands/import_redirects.py
marteinn: In that case I'll add support for both kebab and snake to keep everyone happy (I'm personally not a `kebab-fan` casing fan since it makes copying harder)..

wagtail/contrib/redirects/management/commands/import_redirects.py
marteinn: I usually use `-1` as a replacement to `None` when dealing with positive ints, mainly to improve type integrity. So -1 in this case will import all rows. But in this case I think you are right, since -1, in the context of value operations in python could mean we want to import the last row. I'm changing to None.

wagtail/contrib/redirects/management/commands/import_redirects.py
lb-: I actually agree with that, double clicking on the text to copy :) 
However, alas, dealing with legacy conventions across Django and Wagtail means compromises here and there.

Thanks

wagtail/contrib/redirects/views.py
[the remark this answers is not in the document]
marteinn: I managed to find a good way to combine the initial render and validation error render by adding a if statement to set the proper form. I think it looks better now.

wagtail/contrib/redirects/management/commands/import_redirects.py
marteinn: I centralized the extension/format config into a function in utils.

wagtail/contrib/redirects/templates/wagtailredirects/choose_import_file.html
marteinn: What I did here was to add a info block on top and requirements in the file uploader help text.

marteinn:
Again, thank you so much @lb- for the feedback, super valuable and I think it really improved the user experience.

I just made a commit that should address all the issues you raised, (I went through them one by one and resolved).

Any additional feedback is of course much appreciated!

lb-:
Awesome! I'll try do do another full review over the weekend

marteinn:
Fantastic @lb-, much appreciated!

docs/reference/contrib/redirects.rst
```diff
@@ -40,6 +40,44 @@ Page model recipe of to have redirects created automatically when changing a pag
...
+
+This command imports and creates redirects from a file supplied by the user.
+
+Options:
+
+ - **src**
+   This is the path to the file you wish to import redirects from.
+
+ - **site_id**
+   This is the **site_id** for the site you wish to save redirects to.
```
tomdyson, line 62: What's the behaviour if `site_id` is not supplied?

marteinn: The `site` argument (it has been renamed from `site_id` but I missed to change it in the docs, will fix it) has the same behavior as "From site" when you create redirects manually. 

Reading through the description I think it needs a little rewording, perhaps something along these lines? _The is the site where redirects come from, if omitted it will be applied to all sites._

marteinn: I realized I never answered your question, if not supplied the redirect will work on all sites.

lb-:
@marteinn  - awesome stuff!
I did another round of validation and this feels much better, the error, info & success messages make it really clear what is happening.

I made a few small tweaks locally, fixed one typo and have merged in.

via 301d1bc7f52fffc18b8cb34fbde01633bc1599ee

lb-:
@gasman  - I have not added this to the 'What's new' section of the 2.10 release notes in this commit but figured we would probably want to add this as a higher profile feature but was not sure what to write.

marteinn:
Fantastic @lb-! Many thanks for the review and merging.

PR #5953, opened 2020-04-14
Some cleanups in PostgreSQL search module

kaedroho:
This reduces the amount of raw SQL in the PostgreSQL search module and makes it easier to follow (I think anyway).

The new ``query.py`` is taken from https://github.com/django/django/pull/8313, I plan to add tests/docs for this and resubmit it to Django, so it will only live in Wagtail temporarily.

### DIFF
--- a/docs/reference/contrib/postgres_search.rst
+++ b/docs/reference/contrib/postgres_search.rst
@@ -4,7 +4,7 @@
 PostgreSQL search engine
 ========================
 
-This contrib module provides a search engine backend for Wagtail using
+This contrib module provides a search engine backend using
 `PostgreSQL full-text search capabilities <https://www.postgresql.org/docs/current/static/textsearch.html>`_.
 
 .. warning::
@@ -13,23 +13,20 @@ This contrib module provides a search engine backend for Wagtail using
 
 **Features**:
 
-- Supports all the search features available in Wagtail.
+- It supports all the search features available in Wagtail.
 - Easy to install and adds no external dependency or service.
-- Excellent performance for sites with up to 200 000 pages.
-  Stays decent for sites up to a million pages.
-- Faster to reindex than Elasticsearch if you use PostgreSQL 9.5 or more.
+- Excellent performance for sites with up to 200 000 pages and stays decent for sites up to a million pages.
+- Faster to reindex than Elasticsearch, if you use PostgreSQL 9.5 or higher.
 
-The only known **downsides** concern :
+**Drawbacks**:
 
-**Downsides**:
-
-- ``SearchField(partial_match=True)`` is not handled.
+- Partial matching (``SearchField(partial_match=True)``) is not supported
 - Due to a PostgreSQL limitation, ``SearchField(boost=…)`` is only partially
-  respected. It is changed so that there can only be 4 different boosts.
-  If you define 4 or less different boosts,
+  respected. It is changed so that there can only be four different boosts.
+  If you define four or less different boosts,
   everything will be perfectly accurate.
   However, your search will be a little less accurate if you define more than
-  4 different boosts. That being said, it will work and be roughly the same.
+  four different boosts. That being said, it will work and be roughly the same.
 - When :ref:`wagtailsearch_specifying_fields`, the index is not used,
   so it will be slow on huge sites.
 - Still when :ref:`wagtailsearch_specifying_fields`, you cannot search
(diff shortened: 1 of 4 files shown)


docs/reference/contrib/postgres_search.rst
```diff
@@ -13,23 +13,20 @@ This contrib module provides a search engine backend for Wagtail using
 
 **Features**:
 
-- Supports all the search features available in Wagtail.
+- It supports all the search features available in Wagtail.
 - Easy to install and adds no external dependency or service.
-- Excellent performance for sites with up to 200 000 pages.
-  Stays decent for sites up to a million pages.
-- Faster to reindex than Elasticsearch if you use PostgreSQL 9.5 or more.
+- Excellent performance for sites with up to 200 000 pages and stays decent for sites up to a million pages.
```
tomdyson, line 18: Do we have any data to support this?

docs/reference/contrib/postgres_search.rst
```diff
@@ -13,23 +13,20 @@ This contrib module provides a search engine backend for Wagtail using
...
 
-**Downsides**:
-
-- ``SearchField(partial_match=True)`` is not handled.
+- Partial matching (``SearchField(partial_match=True)``) is not supported
 - Due to a PostgreSQL limitation, ``SearchField(boost=…)`` is only partially
-  respected. It is changed so that there can only be 4 different boosts.
-  If you define 4 or less different boosts,
+  respected. It is changed so that there can only be four different boosts.
+  If you define four or less different boosts,
```
tomdyson, line 26: If you define four or less different boosts -> If you define up to four different boosts

wagtail/contrib/postgres_search/backend.py
```diff
@@ -192,106 +247,136 @@ def __str__(self):
...
+                new_lexeme = Lexeme(term, invert=invert)
+
+                if query.operator == 'and':
+                    lexemes &= new_lexeme
+                else:
+                    lexemes |= new_lexeme
+
+            return lexemes
+
+        elif isinstance(query, Boost):
```
tomdyson, line 306: If boosts aren't supported here, should we raise a warning?

wagtail/contrib/postgres_search/query.py
```diff
@@ -0,0 +1,124 @@
+# From https://github.com/django/django/pull/8313
```
tomdyson, line 1: Should we include a reminder here that this can be removed when/if the Django PR is accepted?

docs/reference/contrib/postgres_search.rst
kaedroho: ping @BertrandBordage

wagtail/contrib/postgres_search/query.py
```diff
@@ -0,0 +1,125 @@
...
+
+        rsql, params = compiler.compile(self.rhs)
+        value_params.extend(params)
+
+        combined_sql = '({} {} {})'.format(lsql, self.connector, rsql)
+        combined_value = combined_sql % tuple(value_params)
+        return '%s', [combined_value]
+
+
+class RawSearchQuery(SearchQueryCombinable, Expression):
```
gasman, line 88: @kaedroho Please can you explain what this class is doing that Django's own `SearchQuery` class doesn't do? My best guess is that it's the equivalent of the two-line patch to `SearchQuery` in https://github.com/django/django/pull/12727 , but that implementation has changed so drastically between Django 3.0 and master that I can't see how one relates to the other.

docs/reference/contrib/postgres_search.rst
BertrandBordage: Yes, I made a test using data from a ~ 200 000 pages Wagtail site from Torchbox that @gasman handed me during the Wagtail space where we created this. Response time was still under a second for search queries on those 200 000 pages, including ranking (which is the most expensive operation).
I then extrapolated the data by cloning existing pages until I reached 2 millions if I remember correctly. It was struggling a bit after around a million pages, where response time went over 5 seconds.

But I don’t have a more detailed report nor did a clean script to create “realistic” fake data of that size.

Also, I deleted the data, I don’t remember which site it was from, as it didn’t matter that much.

So if you prefer being cautious about this, you can remove this claim. Even though it was true at that time, it may not stay true in the future.

wagtail/contrib/postgres_search/query.py
kaedroho: Up until Django 3.1, it wasn't possible to pass in expressions as the query string: https://github.com/django/django/commit/3baf92cf8230ad3a932986170fd07c8feae7ff2f#diff-81374e85d785546664c64c9ecb89b085

wagtail/contrib/postgres_search/backend.py
kaedroho: Added this, thanks!

PR #5955, opened 2020-04-14
Switch order of hook injected css in admin_base.html

gcmurphy:
When trying to fully skin the admin login pages I noticed that the
`login.css` file is injected via `{% block extra_css %}` which exists
below the `{% hook_output 'insert_global_admin_css' %}`. This means
that any customization to the login / password_reset pages will need
to create a `.html` file with their own `{% block extra_css %}` to
override colors and styles rather than just relying on the hook
to inject custom styles.

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/admin_base.html
+++ b/wagtail/admin/templates/wagtailadmin/admin_base.html
@@ -5,9 +5,11 @@
     <link rel="stylesheet" href="{% versioned_static 'wagtailadmin/css/vendor/jquery-ui/jquery-ui-1.10.3.verdant.css' %}" />
     <link rel="stylesheet" href="{% versioned_static 'wagtailadmin/css/vendor/jquery.tagit.css' %}">
     <link rel="stylesheet" href="{% versioned_static 'wagtailadmin/css/core.css' %}" type="text/css" />
-    {% hook_output 'insert_global_admin_css' %}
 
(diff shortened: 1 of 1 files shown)


lb-:
@gcmurphy  - thanks for the PR, I had a look and feel this could cause a problem for the assumptions about how `extra_css` blocks work on other pages. From what I can gather the `insert_global_admin_css` hook is for importing extra items and not necessarily overriding existing styles.

It might be best to raise a bug for the underlying issue found for discussion, I do agree the login.scss styles are a big heavy handed and would be hard to override.

An alternative approach for your problem would be to customise just one template
`wagtail/admin/templates/wagtailadmin/admin_base.html`
and simply add something like the following
```
{% block css %}
  {% block.super %}
  ... add more global styles here that will be added after extra_css
{% endblock %}
```

gasman:
I think it's worth adding that both the hooks and the `extra_css` block can potentially be used within end-user code, third-party add-ons and Wagtail itself - it would be incorrect to assume that `extra_css` only includes core Wagtail styles and `insert_global_admin_css` only includes custom user CSS to override them.

Swapping them round is liable to cause the opposite problem: if your project includes a custom-built admin page with specific CSS defined in `{% block extra_css %}`, you don't want a third-party module to be able to overwrite those styles through the `insert_global_admin_css` hook.

PR #5959, opened 2020-04-16
Filtering for locked page report

gasman:
Allow the locked pages report to be filtered by 'locked at', 'locked by' and 'is live'.

Split out from #5940, to cover the parts of report filtering that do not depend on the workflow branch. Incorporates #5954.


gasman:
Test added in 3c9c75817da34b1c6536336c92b07578d3bdfd2c.

PR #5960, opened 2020-04-16
Implement autocomplete API for PostgreSQL

kaedroho:
This PR finishes the implementation of the autocomplete API ([RFC 27](https://github.com/wagtail/rfcs/blob/master/text/027-autocomplete.md)) for PostgreSQL.

### DIFF
--- a/wagtail/contrib/postgres_search/backend.py
+++ b/wagtail/contrib/postgres_search/backend.py
@@ -30,10 +30,11 @@ class ObjectIndexer:
     """
     Responsible for extracting data from an object to be inserted into the index.
     """
-    def __init__(self, obj, search_config):
+    def __init__(self, obj, backend):
         self.obj = obj
         self.search_fields = obj.get_search_fields()
-        self.search_config = search_config
+        self.config = backend.config
+        self.autocomplete_config = backend.autocomplete_config
 
     def prepare_value(self, value):
         if isinstance(value, str):
@@ -74,15 +75,17 @@ class ObjectIndexer:
                 for sub_field in field.fields:
                     yield from self.prepare_field(sub_obj, sub_field)
 
-    def as_vector(self, texts):
+    def as_vector(self, texts, for_autocomplete=False):
         """
         Converts an array of strings into a SearchVector that can be indexed.
         """
         if not texts:
             return EMPTY_VECTOR
 
+        search_config = self.autocomplete_config if for_autocomplete else self.config
+
         return ADD([
-            SearchVector(Value(text, output_field=TextField()), weight=weight, config=self.search_config)
+            SearchVector(Value(text, output_field=TextField()), weight=weight, config=search_config)
             for text, weight in texts
         ])
 
@@ -117,7 +120,7 @@ class ObjectIndexer:
                 if isinstance(current_field, AutocompleteField):
                     texts.append((value, boost))
 
-        return self.as_vector(texts)
+        return self.as_vector(texts, for_autocomplete=True)
 
 
 class Index:
@@ -231,7 +234,7 @@ class Index:
         if not search_fields:
             return
 
-        indexers = [ObjectIndexer(obj, self.backend.config) for obj in objs]
+        indexers = [ObjectIndexer(obj, self.backend) for obj in objs]
 
         # TODO: Delete unindexed objects while dealing with proxy models.
         if indexers:
@@ -268,6 +271,9 @@ class PostgresSearchQueryCompiler(BaseSearchQueryCompiler):
                 field_lookup: self.get_search_field(field_lookup, fields=search_fields)
                 for field_lookup in self.fields
             }
+
+    def get_config(self, backend):
+        return backend.config
 
     def get_search_field(self, field_lookup, fields=None):
         if fields is None:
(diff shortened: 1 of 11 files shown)


wagtail/contrib/postgres_search/backend.py
```diff
@@ -92,25 +95,25 @@ def id(self):
...
     @cached_property
     def autocomplete(self):
         """
-        Returns all values to index as "body". This is the value of all SearchFields that have partial_match=True
+        Returns all values to index as "autocomplete". This is the value of all AutocompleteFields
         """
         texts = []
         for field in self.search_fields:
             for current_field, boost, value in self.prepare_field(self.obj, field):
-                if isinstance(current_field, SearchField) and current_field.partial_match:
```
gasman: I'm okay with us breaking backwards compatibility on `AutocompleteField`, but isn't this also breaking the behaviour of `SearchField(partial_match=True)`, which _is_ documented? Shouldn't we be indexing both `AutocompleteField` and `SearchField(partial_match=True)` into the autocomplete index, until the latter is formally deprecated?

kaedroho: The autocomplete field is only used on its own when using the undocumented `autocomplete()` API. When users use the `.search()` API it's just merged back into `body` at runtime and treated no differently.

gasman: Ahh, I see... I missed the fact that the Postgres backend has never supported the current documented functionality of `partial_match=True` (i.e. returning partial matches on `search()`), so any references to `partial_match` that do exist in the code are only there to support the previous iteration of `autocomplete()`, and are fair game to be removed.

Having said that, it looks like there's one remaining reference to `partial_match` in the backend after this PR:

https://github.com/kaedroho/wagtail/blob/d536e73baf017898b3bd6570d4221d227709ea60/wagtail/contrib/postgres_search/backend.py#L472-L473

I guess this should be updated to iterate over `self.queryset.model.get_autocomplete_search_fields()` instead?

wagtail/contrib/postgres_search/backend.py
```diff
@@ -52,6 +52,9 @@ def prepare_field(self, obj, field):
             yield (field, get_weight(field.boost),
                    self.prepare_value(field.get_value(obj)))
 
+        elif isinstance(field, AutocompleteField):
+            yield (field, 'D', self.prepare_value(field.get_value(obj)))
```
gasman, line 57: Is there a reason we're hard-coding boost "D" here, rather than calling `get_weight`? (If so, I'm fine with that, but we should probably explain why in a comment.)

wagtail/contrib/postgres_search/backend.py
kaedroho: Ahh, yep that's right will fix!

I think the tests have been working because the title field in the test model has both `partial_match=True` and is an `AutocompleteField`

wagtail/contrib/postgres_search/backend.py
kaedroho: There's no boost on `AutocompleteField`. Autocomplete should be just one field (either title, name, etc)

gasman: Is there a fundamental reason not to support multiple `AutocompleteField`s on a model? I could see a use for something like a Person model with AutocompleteFields on `first_name` and `last_name`, so that `Person.objects.autocomplete('matthe')` returns both Matthew Westcott and Bernard Matthews (and potentially gives the latter a higher ranking, if there's a boost on `last_name`).

(Happy to treat this as out of scope for the current PR though...)

wagtail/contrib/postgres_search/backend.py
gasman: Proposed fix in a329870a8e18fa7faea04139ee44b9a250cf04d5.

It turns out my "why can't we use boost on AutocompleteFields?" tangent is somewhat relevant here after all - having patched `PostgresAutocompleteQueryCompiler` to build `self.search_fields` as a lookup of `AutocompleteField`s rather than `SearchField`s, it turns out that `get_fields_vector` also needs patching to hard-code a weight of 'D', rather than calling `get_weight(search_field.boost)` (which is not currently valid on AutocompleteField).

If AutocompleteField supported a `boost` field - and I can't see a logical reason why it shouldn't - we could get away without overriding this. Happy to leave that for a future PR though.

gasman:
The new test for `autocomplete` with a fields argument (which is necessary to provide test coverage of `PostgresAutocompleteQueryCompiler.get_fields_vector`) is now failing on Elasticsearch:

    self.backend.autocomplete("Georg", models.Author, fields=['name'])

(there are also various failures of test_get_mapping / test_get_document which are expected, given the model change I made to add `index.AutocompleteField('name')` on Author)

Haven't had chance to dig into this in detail, but I think it might be due to this line where an explicit fields mapping takes precedence over the one that tells it to use edgengrams:

https://github.com/wagtail/wagtail/blob/1958bba2476c32069acafd6837514749a4d3ceea/wagtail/search/backends/elasticsearch2.py#L623

@kaedroho, can you advise on whether passing `fields` to autocomplete is something that we'd want to support, or is that not feasible on Elasticsearch? If not, I guess the best thing to do is leave the fixed version of `PostgresAutocompleteQueryCompiler.get_fields_vector` in place, but leave it untested (or make that a Postgres-specific test).

kaedroho:
It is feasible on elasticsearch, we could index the autocomplete twice. Both in the `_edgengrams` field but also as each field individually. But I don't think we should support this, I can't think of a single moment I ever needed to do that (usually autocomplete is only added to the title anyway).

kaedroho:
Latest changes look good to me, thanks!

gasman:
Thanks! Haven't added any release note / changelog entry for this, as this is still an undocumented API - we should probably plan for making it official once the dust has settled.

PR #5965, opened 2020-04-18
Fix draft view when preview modes are disabled

zerolab:
Ref: #5670

View draft uses `page.make_preview_request(request, page.default_preview_mode)`, as does viewing a revision

This PR fixes `default_preview_mode` to fall gracefully to returning `None` when no preview modes are defined


gasman:
I think the more correct thing to do here is to hide the 'view draft' / 'view revision' link when preview_modes is empty. While these are notionally not previews from an end-user/programmer perspective, in practice they are the same mechanism, and the situations where you'd want to disable previewing (i.e. headless setups where front-end page rendering is outside the control of Wagtail/Django) are ones where you'd also want to disable viewing of drafts / revisions.

gasman:
Alternative fix submitted: #5971

PR #5971, opened 2020-04-20
Hide preview / view draft buttons throughout admin UI when preview_modes is empty

gasman:
Ref: https://github.com/wagtail/wagtail/issues/5670#issuecomment-615836390
Add an `is_previewable` method to Page (which accounts for being called either on the base Page model or the specific subclass, as either may be in use on listing pages that display these buttons), and check this when rendering buttons.

Additionally, fix the 'pages for moderation' panel to check that revision.user is defined before trying to output a username - this field is intentionally nullable to allow for scripted processes. Without this check, under normal template-rendering rules the whole panel is liable to be silently dropped.

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -1402,6 +1402,10 @@ class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):
 
     @property
     def default_preview_mode(self):
+        """
+        The preview mode to use in workflows that do not give the user the option of selecting a
+        mode explicitly, e.g. moderator approval. Will raise IndexError if preview_modes is empty
+        """
         return self.preview_modes[0][0]
 
     def is_previewable(self):
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -1404,6 +1404,20 @@ class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):
     def default_preview_mode(self):
         return self.preview_modes[0][0]
 
+    def is_previewable(self):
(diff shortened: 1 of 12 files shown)


wagtail/core/models.py
```diff
@@ -1404,6 +1404,20 @@ def preview_modes(self):
     def default_preview_mode(self):
         return self.preview_modes[0][0]
 
+    def is_previewable(self):
```
zerolab, line 1407: 👍💯

gasman:
Tests added in cbf4ba973f9a210637bf9d82cdeebb6ae3c0a2bf - thanks @zerolab!

I'm not sold on the idea of keeping the view_draft view working if people visit it directly - if a developer sets `preview_modes` to empty, that means they've effectively opted out of the standard template rendering mechanism, and trying to recover from user error by overriding that preference and running the template rendering anyhow seems like the "wrong sort of helpful" :-) I think it does make sense for visiting view_draft to be an error condition, although serving a more meaningful error (a 403 perhaps?) is not a bad idea.

tomdyson:
> if a developer sets preview_modes to empty, that means they've effectively opted out of the standard template rendering mechanism

I strongly agree! 403 feels right to me.

gasman:
Updated in 4a654b4dbcea05ec5eee7735f395c7978fbd08ef to return a 403 on these endpoints.

zerolab:
> I think it does make sense for visiting view_draft to be an error condition, although serving a more meaningful error (a 403 perhaps?) is not a bad idea.

👍

Issue #4871, opened 2018-10-28
[Discussion] choose appropriate tool for WCAG auto testing / visual regression tests


davojta:
Problem:
*  Wagtail need a tool for auto frontend tests
* wagtail need auto testing for wcag

Here is the cypress test for wagtail - https://github.com/davojta/test-cypress-for-wagtail

There are several solutions for wcag testing:

* pa11yjs
* axe-core

There also some options for visual regression testing

* cypress
* web driver
* puppeteer 

To be updated later )

thibaudcolas:
To add a bit more context, here are the things we want to test accessibility-wise at the moment:

- Icon font replacement to SVG, https://github.com/wagtail/wagtail/issues/4199#issuecomment-358640129, #4821
- Document outline & landmarks,  https://github.com/wagtail/wagtail/issues/4199#issuecomment-358640129
- Textual labels to icon-only elements, https://github.com/wagtail/wagtail/issues/4199#issuecomment-358640129
- Focus management, https://github.com/wagtail/wagtail/issues/4199#issuecomment-358640129
- Contrast ratio (WCAG 2.0 AA level), #2224, #4788 

For the actual accessibility tester, I feel pretty strongly about picking aXe since it's battle-tested, and is know to have a "no false positives" policy for use in automated tests.

thibaudcolas:
Looking into this further, I see the need for at least two tools / sets of tools, for two separate scenarios:

- Dev-time "static analysis" / linting, that can give developers some feedback as they work on their change (e.g. a PR to add a new view in the admin). And can also run through the source code in CI.
- Automated E2E integration testing, with browser automation tools, that runs a more comprehensive set of checks.

For the E2E accessibility tests, the most mature is [pa11y](https://github.com/pa11y/pa11y), already mentioned above. They are working on an Axe integration that would make it perfect (https://github.com/pa11y/pa11y/pull/408). If this doesn't work out, I think the next best would be to have our own Puppeteer + Axe integration (pa11y is an abstraction layer for Puppeteer interactions).

For the dev-time static analysis, I haven't found that many options. Common HTML linting tools do not have good support for fragments, or templating syntax, and do not have mature rules ecosystems like ESLint/stylelint/flake8.

[...]

PR #5977, opened 2020-04-21
Revert "Add role="table" to TableBlock output"

thibaudcolas:
This reverts #5588, commit 9a21e79ff88e13d83c81d01f620313a4ddf16b14. cc @gasman and @loicteixeira for thoughts.

I’m proposing this revert for a few reasons:

First, `role="table"` is the default for `table` elements, so HTML5 validation raises the following warning when stumbling upon this: `Warning: The table role is unnecessary for element table`.
See for example https://www.buckinghamshire.gov.uk/libraries/library-membership/library-charges-and-concessions/ in https://html5.validator.nu/.

From my brief testing, it looks like the browser’s (Chrome) heuristics also take the presence of a `<caption>` on a table to determine it’s a data table. Adding a caption is a much more desirable if people experience this issue on their tables – as accessibility standards mandate usage of (data) table captions anyway, and forbid captions usage for layout tables.
- I get the same results with the table containing a `thead` / `th` elements. Which are also mandatory / (highly recommended?) by accessibility guidelines.

Additionally, I don’t see why we would want to override browser’s heuristics for this to start with. Even when the heuristics aren’t good enough (and as far as I can tell, they are), announcing a small data table as if it was presentational really doesn’t do that much harm - it just announces the content following the language’s directionality. On the other hand announcing a presentation table as data means having a much more verbose output than warranted.

In essence I don’t see how this can be a good idea as a blanket rule – if people use tables for layout then this change is harmful. If people use tables for data then they need to put a caption, th, thead anyway. And if they get that wrong, the browser can decide based on its heuristics, which seem quite good.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -13,6 +13,7 @@ Changelog
  * `AbstractEmailForm` now has a separate method (`render_email`) to build up email content on submission emails (Haydn Greatnews)
  * Fix: Ensure link to add a new user works when no users are visible in the users list (LB (Ben Johnston))
  * Fix: `AbstractEmailForm` saved submission fields are now aligned with the email content fields, `form.cleaned_data` will be used instead of `form.fields` (Haydn Greatnews)
+ * Fix: Removed ARIA `role="table"` from TableBlock output (Thibaud Colas)
 
 
 2.9 (xx.xx.xxxx) - IN DEVELOPMENT
--- a/docs/releases/2.10.rst
+++ b/docs/releases/2.10.rst
@@ -27,6 +27,7 @@ Bug fixes
 
  * Ensure link to add a new user works when no users are visible in the users list (LB (Ben Johnston))
  * ``AbstractEmailForm`` saved submission fields are now aligned with the email content fields, ``form.cleaned_data`` will be used instead of ``form.fields`` (Haydn Greatnews)
+ * Removed ARIA `role="table"` from TableBlock output (Thibaud Colas)
 
 
 Upgrade considerations
--- a/wagtail/contrib/table_block/templates/table_block/blocks/table.html
+++ b/wagtail/contrib/table_block/templates/table_block/blocks/table.html
@@ -1,6 +1,6 @@
 {% load table_block_tags %}
 
-<table role="table">
+<table>
     {% if table_caption %}
        <caption>{{ table_caption }}</caption>
     {% endif %}
--- a/wagtail/contrib/table_block/tests.py
+++ b/wagtail/contrib/table_block/tests.py
@@ -37,7 +37,7 @@ class TestTableBlock(TestCase):
         block = TableBlock()
         result = block.render(value)
         expected = """
-            <table role="table">
+            <table>
                 <tbody>
                     <tr><td>Test 1</td><td>Test 2</td><td>Test 3</td></tr>
                     <tr><td></td><td></td><td></td></tr>
@@ -61,7 +61,7 @@ class TestTableBlock(TestCase):
         block = TableBlock()
         result = block.render(value)
         expected = """
-            <table role="table">
+            <table>
                 <thead>
                     <tr><th scope="col">Test 1</th><th scope="col" class="htLeft">Test 2</th><th scope="col">Test 3</th></tr>
                 </thead>
@@ -90,7 +90,7 @@ class TestTableBlock(TestCase):
             ]
         })
         expected = """
-            <table role="table">
+            <table>
                 <tbody>
                     <tr><td></td><td></td><td></td></tr>
                     <tr><td></td><td></td><td></td></tr>
@@ -110,7 +110,7 @@ class TestTableBlock(TestCase):
                           [None, None, None]]}
 
         expected = """
-            <table role="table">
+            <table>
                 <tbody>
                     <tr><td>&lt;p&gt;&lt;strong&gt;Test&lt;/strong&gt;&lt;/p&gt;</td><td></td><td></td></tr>
                     <tr><td></td><td></td><td></td></tr>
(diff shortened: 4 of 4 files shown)


loicteixeira:
As stated in the initial PR, the table bloc only accept raw text, so I doubt it will ever be used for layout purposes. However, if the default heuristics are good enough, then sure, let's remove the unnecessary markup.

I have to admit that I hadn't tested the original PR and simply relied on the argument which sounded like a good idea. I guess it's a reminder that this should be tested manually 😞

thibaudcolas:
@loicteixeira to be fair I doubt anyone ever runs HTML validation as part of reviewing PRs. Automated accessibility testing would have caught this but it’s stuck in limbo at #4871.

Again, to me the point is that we shouldn’t try to override the browsers / screen readers’ heuristics even if we think we know better, as the consequences of getting this wrong on our side are much worse than the potential improvement when the heuristics are wrong. And the current approach comes at the cost of a clean HTML validation report.

Here is an interesting read on the same topic: https://webaim.org/techniques/tables/

gasman:
> First, `role="table"` is the default for table elements, so HTML5 validation raises the following warning when stumbling upon this: `Warning: The table role is unnecessary for element table.`

If indeed there are cases where real-world browsers override the ARIA role as a result of deciding that a table is presentational, then clearly the table role _isn't_ unnecessary, and the validator is either flat out wrong here, or validating according to a spec that isn't respected by real-world browsers. Either way, I'd side with the actual browser behaviour and the advice of several accessibility experts (assuming I'm not misinterpreting that advice, which is a possibility)...

https://codepen.io/dexterperrin/full/ExYMvRV
https://speakerdeck.com/edds/what-even-is-a-table-a-quick-look-at-accessibility-apis

> From my brief testing, it looks like the browser’s (Chrome) heuristics also take the presence of a <caption> on a table to determine it’s a data table. Adding a caption is a much more desirable if people experience this issue on their tables – as accessibility standards mandate usage of (data) table captions anyway, and forbid captions usage for layout tables.

One of the key design principles of Wagtail is that we shouldn't rely on editors to be knowledgeable about web best practices - roughly speaking, our aim is to take input from editors whose tech expertise is on the level of "being able to prepare a Word document" and turn that into a standards-compliant website (subject to the site implementer doing their job properly too, of course). The fact that an editor _can_ take steps to make their content accessible is no guarantee that they actually _will_, and I believe there's value in Wagtail taking this extra step to compensate.

(Half-baked idea: what if we made caption a required field unless explicitly overridden in the TableBlock definition? That would make the out-of-the-box configuration accessible by default, without the need for the explicit `role`.)

> announcing a small data table as if it was presentational really doesn’t do that much harm - it just announces the content following the language’s directionality.

According to https://speakerdeck.com/edds/what-even-is-a-table-a-quick-look-at-accessibility-apis?slide=63 , JAWS has a shortcut for skipping to the next table, which would presumably fail if a table was wrongly flagged as presentational.

> if people use tables for layout then this change is harmful.

As @loicteixeira mentions, TableBlock's functionality is limited (intentionally) to make it unsuitable for use for presentational templates. Admittedly, that doesn't mean editors won't _try_ to do that... I guess my argument hangs on the assertion that "this table was created with TableBlock, therefore it's probably a data table" is a better heuristic than the one the browser is using. I'm open to being persuaded otherwise on that :-)

thibaudcolas:
Thank you for taking the time to detail your thoughts @gasman, that really helps, and this slide deck is invaluable as well.

I think I need to learn more about this before I can have a more informed opinion. Currently I’m 100% sure I’m concerned about Wagtail rendering HTML that doesn’t pass validation without warnings, in a way that is not override-able by site authors (without essentially re-implementing the table block’s rendering). HTML validation warnings are very valuable, so I don’t like the idea of anything that desensitises developers to seeing them.

Here is the spec on this for reference,

- https://w3c.github.io/html-aria/#rules-wd
- https://w3c.github.io/html-aria/#docconformance

thibaudcolas:
Looking at this a bit more from the Wagtail perspective, I see that [TableBlock](https://docs.wagtail.io/en/v2.8.1/reference/contrib/table_block.html?highlight=tableblock) supports defining the first row as header, or first column as header, and the caption.

Could we at least make it so `role="table"` isn’t added if neither of these three fields are used? If the user hasn’t bothered to set a header row/column, I don’t see how we can reasonably infer that their table is "data" enough to override browser heuristics. If we really want to enforce TableBlock is used for data tables only (which is effectively what `role="table"` amounts to), then we could switch this to a "Table header" dropdown field with values "First row | First column" so it’s more explicit, and so the HTML is valid.

gasman:
From a quick play on [the codepen example](https://codepen.io/dexterperrin/full/ExYMvRV), it looks like a `<thead>` or just a single `<th>` is sufficient for Chrome's heuristics to recognise it as a data table, so adding `role="table"` just for these cases would be functionally the same as the current proposal to roll it back completely.

I'm now coming round to the idea that the caption/header options provide enough of a 'nudge' to editors to create accessibility-friendly tables - to the point that if an editor chooses not to use them, that's suspicious enough that Wagtail shouldn't be authoritatively stating "yes, this really is a data table". So yep, +1 for merging.

thibaudcolas:
👍 let’s see whether @loicteixeira has more thoughts on this and I’ll merge for release in 2.10. I have found other issues with the tableblock (its `th` aren’t scoped, it would be nice for it to support `summary`), so might invest some time into making a proper test suite.

thibaudcolas:
Looking around the spec more, the heuristics are actually documented in it:

> User agents, especially those that do table analysis on arbitrary content, are encouraged to find heuristics to determine which tables actually contain data and which are merely being used for layout. This specification does not define a precise heuristic, but the following are suggested as possible indicators:
> 
> [Table of heuristics]
> https://html.spec.whatwg.org/multipage/tables.html#the-table-element

The first heuristic for "non-layout tables" is "The use of caption, thead, or th elements", so I think this corroborates your tests @gasman, and the idea that we could enforce TableBlock tables have a caption and row or column headers. That’s probably for another day though.

loicteixeira:
> let’s see whether @loicteixeira has more thoughts on this

Nothing else to add no, you two made an amazing work with all that testing and spec reading 👍

thibaudcolas:
Thank you both – merged for release in Wagtail 2.10.

PR #5981, opened 2020-04-24
Fix code example in image file formats docs

sephii:
[The code in the docs](https://docs.wagtail.io/en/v2.8.1/advanced_topics/images/image_file_formats.html#using-the-picture-element) result in a `Rendition object (xxx)` because the `{% image %}` template tag used with `as` returns a rendition object instead of an `img` tag.


lb-:
Merged via 2055da6a3aab8302ee16fcd2cc72901b6159b8a0

Thanks

PR #5993, opened 2020-04-28
Documentation: Adding Orderable superclass in code example in "Bindin…

tjwalch:
Documentation: Adding Orderable superclass in code example in "Binding Pages to Snippets". If this is not included sorting in InlinePanel won't work so good to include it in documentation.


gasman:
Merged in 72ee81fbe5e4d52083935d75200bf0d359cbc464.

PR #5997, opened 2020-04-30
Accept ipv6 as domain

allcaps:

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -16,6 +16,7 @@ from django.db import models, transaction
 from django.db.models import Case, Q, Value, When
 from django.db.models.functions import Concat, Substr
 from django.http import Http404
+from django.http.request import split_domain_port
 from django.template.response import TemplateResponse
 from django.urls import NoReverseMatch, reverse
 from django.utils import timezone
@@ -122,7 +123,7 @@ class Site(models.Model):
 
     @staticmethod
     def _find_for_request(request):
-        hostname = request.get_host().split(':')[0]
+        hostname = split_domain_port(request.get_host())[0]
         port = request.get_port()
         site = None
         try:
--- a/wagtail/core/tests/test_sites.py
+++ b/wagtail/core/tests/test_sites.py
@@ -56,7 +56,7 @@ class TestSiteNameDisplay(TestCase):
         self.assertEqual(site.__str__(), 'example.com:8080 [default]')
 
 
-@override_settings(ALLOWED_HOSTS=['example.com', 'unknown.com'])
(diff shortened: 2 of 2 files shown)


allcaps:
Issue https://github.com/wagtail/wagtail/issues/2375
Original pr: https://github.com/wagtail/wagtail/pull/2376
I didn't know how to update the existing PR ^, so I created a new one.

Thanks to @alexgleason for the issue and solution. I just added the tests.

allcaps:
Ref: https://docs.djangoproject.com/en/3.0/ref/settings/
Ref: https://tools.ietf.org/html/rfc3986.html#section-3.2.2

wagtail/core/models.py
```diff
@@ -122,7 +123,7 @@ def find_for_request(request):
 
     @staticmethod
     def _find_for_request(request):
-        hostname = request.get_host().split(':')[0]
+        hostname = split_domain_port(request.get_host())[0]
         port = request.get_port()
```
zerolab, line 127: why not do `hostname, port = split_domain_port(request.get_host())` ?

allcaps: Because the implementation if `request.get_port` and `split_domain_port` are different.

Get port uses the request meta headers.
```
    def get_port(self):
        """Return the port number for the request as a string."""
        if settings.USE_X_FORWARDED_PORT and 'HTTP_X_FORWARDED_PORT' in self.META:
            port = self.META['HTTP_X_FORWARDED_PORT']
        else:
            port = self.META['SERVER_PORT']
        return str(port)
```
Split domain port does something different:
```
def split_domain_port(host):
    """
    Return a (domain, port) tuple from a given host.

    Returned domain is lowercased. If the host is invalid, the domain will be
    empty.
    """
    host = host.lower()

    if not host_validation_re.match(host):
        return '', ''

    if host[-1] == ']':
        # It's an IPv6 address without a port.
        return host, ''
    bits = host.rsplit(':', 1)
    domain, port = bits if len(bits) == 2 else (bits[0], '')
    # Remove a trailing dot (if present) from the domain.
    domain = domain[:-1] if domain.endswith('.') else domain
    return domain, port
```
So I do not dare to touch it.

zerolab: 👍

zerolab: TIL: `request.get_host` uses `split_domain_port` under the hood - https://github.com/django/django/blob/e348ab0d4382b0d7cb0cab9d1261c916c3d0ce6c/django/http/request.py#L128

All good then!

lb-:
Merged in via b98e6d7549e8ad8935c261104a1ca268725733f3

lb-:
Thanks @allcaps  & @alexgleason for putting this together!

PR #5999, opened 2020-05-01
Allow page's variable name in context to be configurable

### LINKED ISSUE #5213: Set context_object_name when defining model?
When defining a Wagtail Page model, it would be nice to be able to set the `context_object_name`, e.g. as a `Meta:` field. This would allow the use of meaningful context object names without needing to define a view.

danbentley:
`Page.context_object_name` can be used to define a more meaningful name
for the page variable in the template.

If defined a `context_object_name` is defined, the default `page` and `self` variables are still available for use in templates.

I might have missed a trick adding the variable to the Page's Meta class but `django.db.models.options.Options`'s `contribute_to_class` method [raises a TypeError](https://github.com/django/django/blob/b4068bc65636cca6c2905aa8c40bea69bb0e4245/django/db/models/options.py#L198) for any unexpected properties found on the Meta.

Instead, `context_object_name` lives on the `Page` and is configurable on a per-page basis.

Resolves #5213

### DIFF
--- a/docs/reference/pages/model_reference.rst
+++ b/docs/reference/pages/model_reference.rst
@@ -153,6 +153,10 @@ In addition to the model fields provided, ``Page`` has many properties and metho
 
     .. automethod:: serve
 
+    .. autoattribute:: context_object_name
+
+        Custom name for page instance in page's ``Context``.
+
     .. automethod:: get_context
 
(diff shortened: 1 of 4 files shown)


danbentley:
Thanks both!

PR #4591, opened 2018-06-07
Publish and Unpublish page hooks

jonatasbaldin:
Hello there!

I know this PR has no test nor documentation, but I opened it with demonstration purpose, because I think we still need to discuss if this is needed/wanted. The discussion is at #4590.

Cheers!

Issue #4590, opened 2018-06-07
Feature request: Hooks for publishing and unpublishing pages


jonatasbaldin:
### Hooks for publishing and unpublish pages

Today the only mechanism to act upon the Publish and Unpublish Page actions are `Signals`, but I don't think they are useful because:
1. You can only execute code _after_ the action, not before
2. You don't have access to the `request` object

That said, I think we could use the `Hooks` mechanism to enable code execution before/after the publish/unpublish actions.

I created a small PR (#4591) to demonstrate how that could work. I didn't write any tests or documentation because I don't know if this interests the maintainers/users, but if so, I could improve the PR with that.

What do you think?

Cheers!

gasman:
Thanks @jonatasbaldin! Happy with the principle of these hooks, but I think we need to be careful about how we specify and document their behaviour, since the 'publish' button on the edit page isn't the only point where a page might get published: it could be submitted for moderation and approved by a moderator, or it could be scheduled for publishing on a future date (in which case the actual publishing will happen from the `publish_scheduled_pages` management task, where there isn't a `request` object available at all).

(Also, if a page is 'published' from the edit page, but has a go-live date in the future, does that count as triggering the hook or not?)

Originally the `before_edit_page` / `after_edit_page` hooks were implemented as a way to modify the admin workflow: e.g. "if the user editing the page is in the 'restricted users' group, redirect them to a custom dashboard page instead of the standard explorer view". I think that's a useful guideline to stick to: anything related to the view / workflow level should be done with hooks, and anything at the model level (e.g. sending notifications, pushing the published pages to an external server...) should be done with signals.

If we're following that convention, then the behaviour as implemented in #4591 is mostly correct (but we should trigger the hook from the 'add' view as well as the 'edit' view). We should make this distinction clear in the documentation, though.

gasman:
Hi @linxuanmax - please use the [Wagtail tag on Stack Overflow](https://stackoverflow.com/questions/tagged/wagtail) for support queries.

lb-:
Merged in via #6000

PR #6000, opened 2020-05-01
Publish unpublish page hooks

allcaps:
Adds tests and docs to PR https://github.com/wagtail/wagtail/pull/4591
I don't know how to update the existing pr. I hope this new pr (including commits from 4591) is the way to go.

Thanks to @jonatasbaldin for adding the new hooks.

### DIFF
--- a/docs/reference/hooks.rst
+++ b/docs/reference/hooks.rst
@@ -505,6 +505,46 @@ Hooks for customising the way users are directed through the process of creating
   Uses the same behaviour as ``before_create_page``.
 
 
+.. _after_publish_page:
+
+``after_publish_page``
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+  Do something with a ``Page`` object after it has been published via page create view or page edit view.
+
+  The function does not have to return anything, but if an object with a ``status_code`` property is returned, Wagtail will use it as a response object and skip the rest of the view.
+
+
+.. _before_publish_page:
+
+``before_publish_page``
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+  Do something with a ``Page`` object before it has been published via page create view or page edit view.
+
+  The function does not have to return anything, but if an object with a ``status_code`` property is returned, Wagtail will use it as a response object and skip the rest of the view.
+
+
+.. _after_unpublish_page:
+
+``after_unpublish_page``
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+  Called after unpublish action in "unpublish" view passing in the request and the page object.
+
+  The function does not have to return anything, but if an object with a ``status_code`` property is returned, Wagtail will use it as a response object and skip the rest of the view.
+
+
+.. _before_unpublish_page:
+
+``before_unpublish_page``
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+  Called before unpublish action in "unpublish" view passing in the request and the page object.
+
+  The function does not have to return anything, but if an object with a ``status_code`` property is returned, Wagtail will use it as a response object and skip the rest of the view.
+
+
 .. _after_copy_page:
 
 ``after_copy_page``
--- a/wagtail/admin/views/pages.py
+++ b/wagtail/admin/views/pages.py
@@ -748,6 +748,11 @@ def unpublish(request, page_id):
     if request.method == 'POST':
         include_descendants = request.POST.get("include_descendants", False)
 
+        for fn in hooks.get_hooks('before_unpublish_page'):
+            result = fn(request, page)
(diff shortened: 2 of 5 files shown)


wagtail/admin/views/pages.py
```diff
@@ -399,11 +399,21 @@ def edit(request, page_id):
...
+                    result = fn(request, page)
+                    if hasattr(result, 'status_code'):
+                        return result
+
                 revision.publish()
                 # Need to reload the page because the URL may have changed, and we
                 # need the up-to-date URL for the "View Live" button.
                 page = page.specific_class.objects.get(pk=page.pk)
 
+                for fn in hooks.get_hooks('after_publish_page'):
```
lb-, line 412: @allcaps the publish hooks are not tested or documented in this revised PR.

However, maybe we should just get the unpublish hooks merged in and leave the publish hooks for a separate piece of work.

see Matt's comments here
https://github.com/wagtail/wagtail/issues/4590#issuecomment-395806974

If you agree with this, I will get this merged in without the publish hooks.

docs/reference/hooks.rst
```diff
@@ -505,6 +505,26 @@ Hooks for customising the way users are directed through the process of creating
...
 
+.. _after_unpublish_page:
+
+``after_unpublish_page``
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+  Called after unpublish action in "unpublish" view passing in the request and the page object.
+
+  The function does not have to return anything, but if an object with a ``status_code`` property is returned, Wagtail will use it as a response object and skip the rest of the view.
+
```
lb-, line 516: We should add a note that this will not be called for scheduled unpublishing.

docs/reference/hooks.rst
```diff
@@ -505,6 +505,26 @@ Hooks for customising the way users are directed through the process of creating
...
+.. _before_unpublish_page:
+
+``before_unpublish_page``
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+  Called before unpublish action in "unpublish" view passing in the request and the page object.
+
+  The function does not have to return anything, but if an object with a ``status_code`` property is returned, Wagtail will use it as a response object and skip the rest of the view.
+
```
lb-, line 526: We should add a note that this will not be called for scheduled unpublishing.

allcaps:
> If you are ok, I will merge this in without the publish hooks and with that docs change.

I happy to apply your docs suggestion myself and will add the publish hooks too. It makes sense to keep all this together in one pr.

I also realised that the before_unpublish hook, when it returns a response, will prevent the original unpublish to be executed. Is that obvious? Maybe I have to add a line explaining that. Maybe:

When a response is returned, the original unpublish will not be called. Add 'page.unpublish()' to your hook if needed.

@lb- ^ Does that make sense?

lb-:
No problems, makes sense. 

In regards to the before hooks, I would assume this would be good for people wanting to add some kind of step (maybe an additional confirmation page) before the unpublish happens. Then on that page they would re-post to the same url - and maybe update the request Param / data with whatever additional check being passed. 

I'll take another look once your update is in. 

Also - have a read of Matt's comment linked above, we may need to docs that adding a page that is published immediately will NOT trigger the hook.

gasman:
> Also - have a read of Matt's comment linked above, we may need to docs that adding a page that is published immediately will NOT trigger the hook.

I'd much rather we fix this (i.e. add the hook to the `create` view too) rather than just documenting the behaviour - if the hook behaves differently between "create page + immediately publish" and "create page, save draft + publish" then I'd consider that a bug.

lb-:
Agreed, thanks Matt. 

Let me know if you need a hand Coen

wagtail/admin/views/pages.py
allcaps: I added:
- Create hooks
- Create hooks tests
- Create hooks docs
- Edit hooks tests
- Edit hooks docs

allcaps:
> ... if the hook behaves differently between "create page + immediately publish" and "create page, save draft + publish" then I'd consider that a bug.

The hooks behave the same. But when a hook returns a response, that response will be served before shadowed code is executed.

Create https://github.com/wagtail/wagtail/pull/6000/files#diff-7ec76fd4042b11064109bfef8030e8f8R258-R327

Edit https://github.com/wagtail/wagtail/pull/6000/files#diff-7ec76fd4042b11064109bfef8030e8f8R427-R581

lb-:
Merged in via a7f58821a7c899857abc745b1eedc71d93ba575a

Thanks @allcaps  & @jonatasbaldin

jonatasbaldin:
awesome ✨

PR #6002, opened 2020-05-03
Docs - revise bg colour & wagtail on hover styles

lb-:
- background teal colour  to match latest Wagtail admin styles
* small on hover animation (similar to the wagtail admin one but more basic)


allcaps:
I did a testdrive and this is a nice change!
Screenshot 2020-05-03 at 13 50 53

PR #6004, opened 2020-05-05
Site switcher, site ordering

allcaps:
Inspired by the work in https://github.com/wagtail/wagtail/pull/5891

Sites do not have a default order. Sites do not have a consistent string representation (site name if given, else hostname). That makes selecting sites in the sites list view and site switcher hard. Especially with many sites.

This PR adds:
- ordering to sites.
- displays the hostnames in site switcher (not the site name).

The ordering of sites in site list view and site switcher are now consistent.

### DIFF
--- a/wagtail/contrib/settings/forms.py
+++ b/wagtail/contrib/settings/forms.py
@@ -17,9 +17,17 @@ class SiteSwitchForm(forms.Form):
     def __init__(self, current_site, model, **kwargs):
         initial_data = {'site': self.get_change_url(current_site, model)}
         super().__init__(initial=initial_data, **kwargs)
-        sites = [(self.get_change_url(site, model), site)
-                 for site in Site.objects.all()]
-        self.fields['site'].choices = sites
+        self.fields["site"].choices = [
+            (
+                self.get_change_url(site, model),
+                (
+                    site.hostname + " [default]"
+                    if site.is_default_site
+                    else site.hostname
+                ),
+            )
+            for site in Site.objects.all()
+        ]
 
     @classmethod
     def get_change_url(cls, site, model):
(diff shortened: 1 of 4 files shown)


allcaps:
### Before

before

### After

after

wagtail/contrib/settings/forms.py
```diff
@@ -17,9 +17,17 @@ def media(self):
...
         initial_data = {'site': self.get_change_url(current_site, model)}
         super().__init__(initial=initial_data, **kwargs)
-        sites = [(self.get_change_url(site, model), site)
-                 for site in Site.objects.all()]
-        self.fields['site'].choices = sites
+        self.fields["site"].choices = [
+            (
+                self.get_change_url(site, model),
+                (
+                    site.hostname + " [default]"
```
lb-, line 24: Do we need to pass this `default` through a translation?

wagtail/core/models.py
```diff
@@ -42,6 +42,9 @@
 
 
 class SiteManager(models.Manager):
+    def get_queryset(self):
+        return super(SiteManager, self).get_queryset().order_by(Lower("hostname"))
+
```
lb-, line 47: Great central solution, love it!

wagtail/contrib/settings/forms.py
allcaps: > 1. Confirming that this will supersede #5891 ?
Yes.

> 2. On that other PR you suggested that the default site still always be at the top and then sorted all others after, how come the change?

Because When I looked at it is seemed silly. Mainly because the site list has a column for default. Breaking out alphabetical order, just seems confusing. So I decided to keep it simple.

wagtail/core/models.py
allcaps: Thank you for the review. I love it when I get quick feedback!

lb-:
Thanks heaps @allcaps & @tleguijt for your initial work on this - merged in via 758ebce865c2ac5eb10d1be57d5503d8b3906740

tleguijt:
Thanks @lb- for picking this up and @allcaps for finalising this!

PR #6015, opened 2020-05-08
improve Adding Reports documentation

phette23:
* `reverse` was used but not imported in the wagtail_hooks.py examples, causing them to be invalid
* I switched the Django `url` calls to use `path` instead since that seems both cleaner and officially recommended, the url function is "likely to be deprecated in a future release" per Django docs and  `path` is starting to replace it in urls.py files now


lb-:
@phette23 - thank you for the fixes for the docs, this has been merged in. Really appreciate your time on making these improvements.

PR #6021, opened 2020-05-12
Use max_count = 1 for HomePage model (Docs - getting started section)

brylie:
Since there is generally only one "Home Page", show the use of the new `max_count` feature, setting the maximum instances to one.

### DIFF
--- a/docs/getting_started/tutorial.rst
+++ b/docs/getting_started/tutorial.rst
@@ -146,6 +146,11 @@ Edit ``home/models.py`` as follows, to add a ``body`` field to the model:
         content_panels = Page.content_panels + [
             FieldPanel('body', classname="full"),
         ]
+        
+        # There should be only one Home page
+        # This restricts the maximum number of Home Page
+        # instances to one
+        max_count = 1
 
 ``body`` is defined as ``RichTextField``, a special Wagtail field. You
 can use any of the :doc:`Django core fields <django:ref/models/fields>`. ``content_panels`` define the


gasman:
Thanks for the suggestion @brylie! I'm undecided about whether this is a good idea, as there are quite a few subtle points to consider:

* Someone just starting the tutorial will probably feel like they're being bombarded with new information... is this the best place to introduce another piece of information that isn't strictly necessary?
* There is a valid use case for having multiple homepages - namely, multi-site setups. Obviously that's not something that a new user following the tutorial will be concerned about, but there's a risk that we create a "cargo cult" effect, where people habitually leave `max_count` in place because that's what the documentation tells you to do, and then wonder why setting up a second site doesn't work.
* At this point in the tutorial, HomePage is the only available page type, which leads to some minor changes of behaviour in the admin - the "add page" button skips past the "choose a page type" view and takes you directly to creating a page of that one available type. Off the top of my head, I'm not sure what happens in the admin when the one available type is blocked by `max_count` - is it smart enough to hide the 'add' button, or does it end up taking you to an empty "choose a page type" screen? We need to be careful not to do something that breaks user expectations. To be fair, the current behaviour isn't free of confusion either - people are sometimes confused that they're able to create pages, but it leads to them creating another homepage under the homepage. Indeed, that may well be what prompted you to propose this change :-) We ought to have a discussion about whether that's better or worse than not being able to create pages at all.

brylie:
> There is a valid use case for having multiple homepages - namely, multi-site setups.

Ah dang. I didn't realize the `max_count` would not allow an instance of HomePage to be created for each site in a multi-site setup. My intuition would be that `max_count` would be used to ensure that each site would have a maximum of instances based on that setting.

gasman:
Will close this for now, but happy to revisit if there are strong feelings that we should encourage new users to make use of this setting.

PR #6022, opened 2020-05-12
Snippet hooks

KalobTaulien:

### DIFF
--- a/docs/reference/hooks.rst
+++ b/docs/reference/hooks.rst
@@ -49,7 +49,6 @@ The available hooks are listed below.
     :local:
     :depth: 1
 
-
 Admin modules
 -------------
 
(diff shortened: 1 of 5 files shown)


wagtail/snippets/tests.py
```diff
@@ -300,6 +318,25 @@ def test_edit_with_tags(self):
...
+            url='http://www.example.com/',
+            text='Test hook',
+        )
+
+        def hook_func(request, instance):
+            self.assertIsInstance(request, HttpRequest)
+            self.assertQuerysetEqual(instance, ["<Advert: Test hook>"])
+            return HttpResponse("Overridden!")
+
+        with self.register_hook('after_delete_snippet', hook_func):
```
zerolab, line 332: this tests the hook deletion, rather than edit :)

docs/reference/hooks.rst
```diff
@@ -976,3 +975,51 @@ Document serving
...
+
+  .. code-block:: python
+
+    from wagtail.core import hooks
+
+    @hooks.register('after_delete_snippet')
+    def after_snippet_delete(request, instances):
+        # "instances" is a QuerySet
+        for instance in instances:
+            print(instance.pk)
```
zerolab, line 1025: The page CRUD hooks use `HttpResponse` as an example. Might be worth doing that instead of one the print statements to have a more realistic code example?

docs/reference/hooks.rst
```diff
@@ -976,3 +975,51 @@ Document serving
...
+  Called when a Snippet is edited. The callable passed into the hook will receive the model instance, the request object. If the callable returns an ``HttpResponse``, that response will be returned immediately to the user, and Wagtail will not proceed to call ``redirect()`` to the listing view.
+
+  .. code-block:: python
+
+    from wagtail.core import hooks
+
+    @hooks.register('after_edit_snippet')
+    def after_snippet_update(request, instance):
+        # your code here
+        # print(instance.pk)
```
lb-, line 999: Might be good to add something a bit more specific here as per other docs examples for hooks
e.g.
```
from django.http import HttpResponse
...
return HttpResponse(f"Congrats on editing a snippet with id {instance.pk}", content_type="text/plain")
```

docs/reference/hooks.rst
lb-: e.g.
```
from django.http import HttpResponse
...
total = len(instances)
return HttpResponse(f"{total} have been deleted", content_type="text/plain")
```

wagtail/snippets/tests.py
KalobTaulien: LOL This is why we do code review. Copy pasta strikes again. Will adjust in the next commit. Thanks!

KalobTaulien:
@zerolab @lb- thanks for the feedback. Made the changes you both suggested. :tada:

lb-:
Thanks @KalobTaulien ! looks great, tested again and works.
Thanks for fixing up those minor items.

merged in via f757324039f67aa0f816ac48946418edf3b880bc

PR #6024, opened 2020-05-13
Fix typo in ModelAdminGroup example

jafacakes2011:
I noticed a typo in the code example for the `ModelAdminGroup` where it shows you can alter the text that will appear in the menu item, this should be `menu_label` and not `label`


lb-:
@jafacakes2011  - thanks heaps, great find!

PR #6028, opened 2020-05-14
Add new "Icons" page for icons customisation and reuse across the admin interface

allcaps:
Docs draft for https://github.com/wagtail/wagtail/pull/4821

### DIFF
--- a/docs/advanced_topics/icons.md
+++ b/docs/advanced_topics/icons.md
@@ -1,0 +1,73 @@
+(icons)=
+
+# Icons
+
+Wagtail comes with an icon set. The icons are used throughout the admin interface.
+
+Elements that use icons are:
+
+- [Various menu items]() <admin_hooks>`{.interpreted-text role="ref"}
+- [ModelAdmin menu items](modeladmin_menu_icon)
+- [Streamfield blocks](basic_block_type_icon)
+- [Rich text editor toolbar buttons](extending_the_draftail_editor)
+
+This document describes how to choose and add icons.
+
+## Available icons and their names
+
+TODO, add a overview of icons that ship with Wagtail (preferably something that keeps in sync with code)
+
+Enable the [styleguide](styleguide) to view icons and their names.
+
+Alternatively inspect the source code `wagtail/admin/wagtail_hooks.py`. The filename without `.svg` is the icon name.
+
+## Add a custom icon
+
+Draw or download an icon.
+
+The SVG should:
+
+-   Have a 1:1 ratio
+-   Contain a `symbol` tag
+-   Have `id="icon-<name>"` attribute
+-   Not set the `fill` attribute
+
+Add the icon to the set with the `register_icons` hook.
+
+```python
+@hooks.register("register_icons")
+def register_icons(icons):
+    return icons + ['path/to/rocket.svg']
+```
+
+## Changing icons via template override
+
+When several applications provide different versions of the same template, the application listed first in `INSTALLED_APPS` has precedence.
+
+Place your app before any Wagtail apps in `INSTALLED_APPS`.
+
+Wagtail icons live in `wagtail/admin/templates/wagtailadmin/icons/`.
+Place your own SVG files in `<your_app>/templates/wagtailadmin/icons/`.
+
+## Changing icons via hooks
+
+```python
+@hooks.register("register_icons")
+def register_icons(icons):
+    icons.remove("wagtailadmin/icons/time.svg")  # Remove the original icon
+    icons.append("path/to/time.svg")  # Add the new icon
+    return icons
+```
+
+## Icon template tag
+
+Use an icon in a custom template:
+
+```html+django
+{% load wagtailadmin_tags %}
+{% icon name="rocket" classname="..." title="Launch" %}
+```
+
+## Icon font support
+
+Use the `insert_global_admin_css` and reference your icons via `class_names`.
--- a/docs/advanced_topics/icons.md
+++ b/docs/advanced_topics/icons.md
@@ -6,33 +6,50 @@ Wagtail comes with an icon set. The icons are used throughout the admin interfac
 
 Elements that use icons are:
 
-- [Various menu items]() <admin_hooks>`{.interpreted-text role="ref"}
-- [ModelAdmin menu items](modeladmin_menu_icon)
-- [Streamfield blocks](basic_block_type_icon)
+- [Register Admin Menu Item](register_admin_menu_item)
+- [Clientside components](extending_clientside_components)
 - [Rich text editor toolbar buttons](extending_the_draftail_editor)
+- [ModelAdmin menu](modeladmin_menu_icon)
+- [Streamfield blocks](custom_streamfield_blocks)
 
 This document describes how to choose and add icons.
 
 ## Available icons and their names
 
-TODO, add a overview of icons that ship with Wagtail (preferably something that keeps in sync with code)
+Icons are registered in `wagtail/admin/wagtail_hooks.py`. 
+The _filename_ without the extension is the icon name.
 
 Enable the [styleguide](styleguide) to view icons and their names.
 
-Alternatively inspect the source code `wagtail/admin/wagtail_hooks.py`. The filename without `.svg` is the icon name.
+[//]: # (The code is included to present an up-to-date icon name list)
+
+```{eval-rst}
+.. literalinclude:: ../../wagtail/admin/wagtail_hooks.py
+   :language: python
+   :pyobject: register_icons
+```
 
 ## Add a custom icon
 
 Draw or download an icon.
 
-The SVG should:
+The `svg` tag should:
 
--   Have a 1:1 ratio
--   Contain a `symbol` tag
--   Have `id="icon-<name>"` attribute
--   Not set the `fill` attribute
+- Set `id="icon-<name>"` attribute, icons are referenced by this name 
+- Set `xmlns="http://www.w3.org/2000/svg"` attribute
+- Set `viewBox="..."` attribute
+- Include license information if applicable
 
-Add the icon to the set with the `register_icons` hook.
+Set `<path fill="currentColor"` to give the icon the current color.
+
+Example:
+
+```{eval-rst}
+.. literalinclude:: ../../wagtail/admin/templates/wagtailadmin/icons/angle-double-left.svg
+   :language: xml
+```
+
+Add the icon to the icon set with the `register_icons` hook.
 
 ```python
 @hooks.register("register_icons")
--- a/docs/advanced_topics/icons.md
+++ b/docs/advanced_topics/icons.md
@@ -2,69 +2,65 @@
 
 # Icons
 
-Wagtail comes with an icon set. The icons are used throughout the admin interface.
+Wagtail comes with an SVG icon set. The icons are used throughout the admin interface.
 
 Elements that use icons are:
 
 - [Register Admin Menu Item](register_admin_menu_item)
-- [Clientside components](extending_clientside_components)
+- [Client-side components](extending_clientside_components)
 - [Rich text editor toolbar buttons](extending_the_draftail_editor)
 - [ModelAdmin menu](modeladmin_menu_icon)
-- [Streamfield blocks](custom_streamfield_blocks)
+- [StreamField blocks](custom_streamfield_blocks)
 
-This document describes how to choose and add icons.
+This document describes how to choose, add and customise icons.
 
 ## Available icons and their names
 
-Icons are registered in `wagtail/admin/wagtail_hooks.py`. 
-The _filename_ without the extension is the icon name.
+An icon name corresponds with the value of `id` in the SVG file: `<svg id="<name>"`.
 
-Enable the [styleguide](styleguide) to view icons and their names.
+Icons are SVG files in the [Wagtail admin template folder](https://github.com/wagtail/wagtail/tree/main/wagtail/admin/templates/wagtailadmin/icons). 
 
-[//]: # (The code is included to present an up-to-date icon name list)
-
-```{eval-rst}
-.. literalinclude:: ../../wagtail/admin/wagtail_hooks.py
-   :language: python
-   :pyobject: register_icons
-```
+Alternatively, enable the [styleguide](styleguide) to view the available icons and their names.
 
 ## Add a custom icon
 
-Draw or download an icon.
+Draw or download an icon and save it in a template folder:
+
+```xml
+# app/templates/app_name/toucan.svg
+
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 800 800" id="icon-toucan">
+  <!-- This work is licensed under a CC0 license. https://creativecommons.org/publicdomain/zero/1.0/ -->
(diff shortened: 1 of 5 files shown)


docs/advanced_topics/icons.rst
```diff
@@ -0,0 +1,85 @@
+.. _icons:
+
+=====
+Icons
+=====
+
+Wagtail comes with an icons set.
```
tomdyson, line 7: icons set -> icon set

docs/advanced_topics/icons.rst
```diff
@@ -0,0 +1,85 @@
...
+
+Elements that use icons are:
+
+- :ref:`Various menu items <admin_hooks>`
+- :ref:`ModelAdmin menu items <modeladmin_menu_icon>`
+- :ref:`Streamfield blocks <basic_block_type_icon>`
+- :ref:`Rich text editor toolbar buttons <extending_the_draftail_editor>`
+- ...
+
+This document describes how to choose an add icons.
```
tomdyson, line 18: an add -> and add

docs/advanced_topics/icons.rst
```diff
@@ -0,0 +1,85 @@
...
+-----------------
+
+Draw or download an icon.
+
+The SVG should:
+
+ - Have a size of 16x16 points
+ - Contain a ``symbol`` tag
+ - Have ``id="icon-<name>"`` attribute
+ - Do not set the ``fill`` attribute
```
tomdyson, line 38: Do not set -> not set

docs/advanced_topics/icons.rst
```diff
@@ -0,0 +1,85 @@
...
+Changing icons via template override
+------------------------------------
+
+When several applications provide different versions of the same template,
+the application listed first in ``INSTALLED_APPS`` has precedence.
+
+Place your app before any Wagtail apps in ``INSTALLED_APPS``.
+
+Wagtail icons live in ``wagtail/admin/templates/wagtailadmin/icons/``.
+Place your own svg files in ``<your_app>/templates/wagtailadmin/icons/``.
```
tomdyson, line 58: svg -> SVG

docs/advanced_topics/icons.rst
```diff
@@ -0,0 +1,85 @@
...
+ - Have a size of 16x16 points
+ - Contain a ``symbol`` tag
+ - Have ``id="icon-<name>"`` attribute
+ - Do not set the ``fill`` attribute
+
+Add the icon to the set with the ``register_icons`` hook.
+
+.. code-block:: python
+
+    @hooks.register('register_icons')
```
tomdyson, line 44: Should we use `black`-style formatting for all our code examples from now on?

docs/advanced_topics/icons.rst
```diff
@@ -0,0 +1,85 @@
...
+    def register_icons(icons):
+        icons.remove("wagtailadmin/icons/time.svg")  # Remove the original icon
+        icons.append("path/to/time.svg")  # Add the new icon
+        return icons
+
+Icon template tag
+-----------------
+
+Use an icon in custom template:
```
tomdyson, line 74: in custom -> in a custom

docs/advanced_topics/icons.rst
```diff
@@ -0,0 +1,85 @@
...
+The filename without ``.svg`` is the icon name.
+
+Add a custom icon
+-----------------
+
+Draw or download an icon.
+
+The SVG should:
+
+ - Have a size of 16x16 points
```
zerolab, line 35: It is unclear to me what 16x16 points means. Is it the viewbox?

allcaps: Yes. I added an example.

zerolab: Cheers @allcaps!
I feel this may prove tricky for non-SVG proficient developers who just want to use a FontAwesome icon (e.g. https://fontawesome.com/icons/clipboard-list has a viewbox of `0 0 384 512` (and I, for example don't have the tools to convert it to a `0 0 16 16` viewbox. Changing the value ends up with a blank icon)

allcaps: I'm not sure the same size matters that much as the final size is defined in CSS. 
Maybe the ratio is significant. But I'm not sure about that too. I have to experiment.

zerolab: aside from this line which is confusing, the rest LGTM and ready to be merged on my opinion

docs/advanced_topics/icons.rst
```diff
@@ -0,0 +1,104 @@
...
+- :ref:`Streamfield blocks <basic_block_type_icon>`
+- :ref:`Rich text editor toolbar buttons <extending_the_draftail_editor>`
+- ...
+
+This document describes how to choose and add icons.
+
+Available icons and their names
+-------------------------------
+
+Enable the :ref:`styleguide <styleguide>` to view icons and their names.
```
thibaudcolas, line 23: Note from https://github.com/wagtail/wagtail/pull/6493#discussion_r588925559 – as part of this documentation work, we could consider adding the icon names directly to the docs.

zerolab:
Hey @adsee42,

The switch to SVG is mostly done. See #6107 for the full list and progress.

What would help get everything over the finish is line is submitting PRs for the remaining items, and reviewing/testing related pull requests such as this one.

docs/advanced_topics/icons.rst
```diff
@@ -0,0 +1,104 @@
...
+    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 16 16">
+      <path fill="black" d="M0 0h16v16H0z"/>
+    </svg>
+
+
+Converted to a SVG symbol:
+
+.. code-block:: html
+
+    <symbol id="icon-square" viewBox="0 0 16 16">
```
kaedroho, line 54: Any reason why we shouldn't do this programmatically? There are JS svg sprite libraries out there that could be ported to Python.

By documenting the fact that users must do this manually, this probably closes the door on developing a tool like that

docs/advanced_topics/icons.rst
```diff
@@ -0,0 +1,104 @@
...
+        <path d="M0 0h16v16H0z"/>
+    </symbol>
+
+
+Add the icon to the set with the ``register_icons`` hook.
+
+.. code-block:: python
+
+    @hooks.register("register_icons")
```
kaedroho, line 63: Any reason to not pick up icon files automatically, like how Django does it with templates?

gasman:
One thing we've tended to miss when adding icons to Wagtail is ensuring that any necessary attribution is in place (which is a requirement for Font Awesome v5 - see #7377). It would be good if we could find somewhere in these docs to put a reminder of this.

This might be something that's better suited for a page in "Contributing to Wagtail" though - it looks like these docs are primarily aimed at people adding icons for their own purposes (including third-party packages).

docs/advanced_topics/icons.rst
```diff
@@ -0,0 +1,104 @@
...
+
+Add a custom icon
+-----------------
+
+Draw or download an icon.
+
+The SVG should:
+
+ - Have a size of 16x16 points
+ - Contain a ``symbol`` tag
```
zerolab, line 36: Note that #8222 changes this to `<svg>`

docs/advanced_topics/icons.rst
```diff
@@ -0,0 +1,104 @@
...
+
+Use an icon in a custom template:
+
+.. code-block:: html+django
+
+    {% load wagtailadmin_tags %}
+    {% icon name="rocket" classname="..." title="Launch" %}
+
+
+Icon font support
```
vsalvino, line 101: Should add a note here that icon fonts cannot be used for `icon` attributes on Streamfield blocks (not sure if this also applies to other icon attributes defined in Python code)

docs/advanced_topics/icons.rst
vsalvino: I am experiencing the same issue. Trying to use fork-awesome icons which look like this:
```svg
<svg width="1536" height="1536" xmlns="http://www.w3.org/2000/svg">
  <path d="M1401 1547l-6 6c-75 ... " />
</svg>
```

Simply removing the width, height, and xmlns attributes and adding `viewbox="0 0 16 16"` does not scale it correctly. I'm not sure what kind of tools or attributes should be used to convert an existing SVG into a wagtail-flavored SVG.

zerolab:
@allcaps this could use a rebase and I think it is high time we merge this in as menus, streamfields and quite a lot of UI components use the SVG icons now

allcaps:
@zerolab Rebased!

I've included code snippets to give easy access to the available icon names and provide a example.

Should we drop the section about icon fonts?

docs/advanced_topics/icons.md
```diff
@@ -0,0 +1,90 @@
...
+## Available icons and their names
+
+Icons are registered in `wagtail/admin/wagtail_hooks.py`. 
+The _filename_ without the extension is the icon name.
+
+Enable the [styleguide](styleguide) to view icons and their names.
+
+[//]: # (The code is included to present an up-to-date icon name list)
+
+```{eval-rst}
```
zerolab, line 26: This is cool! 👏

https://wagtail--6028.org.readthedocs.build/en/6028/advanced_topics/icons.html#available-icons-and-their-names

Any chance the whole block can we wrapped in `<details>` or similar? it is fairly large and takes a fair bit of scrolling to get the core of the doc

docs/advanced_topics/icons.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Add a custom icon
+
+Draw or download an icon.
+
+The `svg` tag should:
+
+- Set `id="icon-<name>"` attribute, icons are referenced by this name 
+- Set `xmlns="http://www.w3.org/2000/svg"` attribute
+- Set `viewBox="..."` attribute
```
zerolab, line 40:
```suggestion
- Set the `id="icon-<name>"` attribute. Icons are referenced by this name 
- Set the `xmlns="http://www.w3.org/2000/svg"` attribute
- Set the `viewBox="..."` attribute
```

zerolab (changes requested):
Oh, this looks so much better!

Rendered: https://wagtail--6028.org.readthedocs.build/en/6028/advanced_topics/icons.html

Left a number of notes/suggestions

docs/advanced_topics/icons.md
```diff
@@ -0,0 +1,90 @@
...
+Use an icon in a custom template:
+
+```html+django
+{% load wagtailadmin_tags %}
+{% icon name="rocket" classname="..." title="Launch" %}
+```
+
+## Icon font support
+
+Use the `insert_global_admin_css` and reference your icons via `class_names`.
```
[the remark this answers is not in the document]
lb-: Agreed. Sure people can do this kind of thing but let's not make it officially supported I think.

docs/advanced_topics/icons.md
```diff
@@ -0,0 +1,90 @@
...
+    return icons
+```
+
+## Icon template tag
+
+Use an icon in a custom template:
+
+```html+django
+{% load wagtailadmin_tags %}
+{% icon name="rocket" classname="..." title="Launch" %}
```
lb-, line 85: I'll do my best to get a PR up so that the icon template supports `classname` Correctly. 

As per

- https://docs.wagtail.org/en/latest/contributing/general_guidelines.html#use-classname-in-python-html-template-tag-variables
- https://github.com/wagtail/wagtail/issues/6107#issuecomment-1325635281

docs/advanced_topics/icons.md
```diff
@@ -0,0 +1,90 @@
...
+
+```{eval-rst}
+.. literalinclude:: ../../wagtail/admin/wagtail_hooks.py
+   :language: python
+   :pyobject: register_icons
+```
+
+## Add a custom icon
+
+Draw or download an icon.
```
vsalvino, line 34: A huge "gotcha" is that the icon file will need to be put in a TEMPLATE folder, not a static folder. This took me several hours of head-scratching to figure out for the first time. I think it would be important to instruct the reader where to save the SVG file.

docs/advanced_topics/icons.md
vsalvino: Is there any guidance on the contents of `viewBox`? I have done several of these icons so far, and every time have to just make up guesses as to what numbers to put here until it eventually looks OK for that specific icon. It's an extremely frustrating developer experience. Many existing SVGs do not have this field, so the developer will likely have to guess.

docs/advanced_topics/icons.md
gasman: There was some talk in https://github.com/wagtail/wagtail/issues/7511#issuecomment-1284975058 about moving the icon files out of templates and into static files, and from the sound of things it might not even be a huge task at this point... it would be really, really nice if we could make that change before this becomes a publicly documented API.

docs/advanced_topics/icons.md
lb-: PR up for this https://github.com/wagtail/wagtail/pull/9769

docs/advanced_topics/icons.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+Add the icon to the icon set with the `register_icons` hook.
+
+```python
+@hooks.register("register_icons")
+def register_icons(icons):
+    return icons + ['path/to/rocket.svg']
+```
+
```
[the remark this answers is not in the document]
allcaps: > I think this needs a note that <name> should match the filename without an extension, no?

This is actually a convention, the real referencing is by `id` in the code, not the filename. I'll adjust.

docs/advanced_topics/icons.md
allcaps: Thanks, I like it too, however... I linked to the source code, and like a simple link even better.

docs/advanced_topics/icons.md
allcaps: @vsalvino A tool like https://jakearchibald.github.io/svgomg/ has the option `Prefer viewBox to width/height`. 

AFAIK the viewbox attribute works better when scaling the icon to the target size.

docs/advanced_topics/icons.md
allcaps: Icons are stored as individual files (easier to maintain, override and extend) but are served as _a single sprite_. We don't need (or want?) to serve individual icon files.

I do get the confusion. But I fail to see the benefits of handling icons as static files. The staticfiles app just brings redundant features and complexity.

I've re-written the docs and explicitly mention that the SVG should go into a template folder.

Maybe I'm missing something. I'm happy to be convinced.

zerolab (approved):
LGTM! :shipit:

lb- (approved):
Just had another read - also looks great. Thanks @allcaps !

lb-:
Does anyone know why we did not get this into 4.2? I'm happy to merge now but not sure if we were still waiting on something. 

@zerolab / @allcaps 

Maybe it's worth putting this in 4.2 stable if we do merge.

zerolab:
@lb- uh-oh, I have failed to add to the 4.2 milestone, so it got lost in translation.

I don't see any reason not to, but pinging @gasman

gasman:
@zerolab Go for it 👍

thibaudcolas:
Merging now! I believe there is still work to be done on this to be honest, but we’re hoping to push for this to happen by the next release, and if this fails we can add a disclaimer.

I wouldn’t recommend us backporting this. The documented setup here isn’t compatible with all of our icons, which is part of the reason why #6107 is still only halfway through the list. We have two solid approaches to solving this mentioned in #7511 that should work without changes to how icons are registered, but they’re just ideas/prototypes at this stage.

lb-:
Thanks Thibaud

PR #6029, opened 2020-05-14
docs - revise heading levels for feature detection (images)

lb-:
* recent docs update had some headings level issues
* revise headings & remove the installation header as it only related to the open cv one and was a bit confusing to read


lb-:
@thibaudcolas  - thanks for merging

PR #6031, opened 2020-05-14
Add export_filename option to modeladmin

jacobtoppm:
Allows modeladmin users to specify a filename for their spreadsheet exports without needing
to add a custom `IndexView`

### DIFF
--- a/docs/reference/contrib/modeladmin/indexview.rst
+++ b/docs/reference/contrib/modeladmin/indexview.rst
@@ -293,6 +293,21 @@ for your model. For example:
         list_filter = ('is_staff', 'company')
 
 
+.. _modeladmin_export_filename:
+
+------------------------------
+``ModelAdmin.export_filename``
+------------------------------
+
+**Expected value**: A string specifying the filename of an exported spreadsheet,
+without file extensions.
+
+.. code-block:: python
+
+    class PersonAdmin(ModelAdmin):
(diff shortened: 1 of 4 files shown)


jacobtoppm:
> @jacobtoppm - this sounds like an awesome feature.
> 
> Are you able to add a test for this.
> 
> Questions:
> 
>     1. What happens if someone puts `people_spreadsheet.xls`, will the download be `people_spreadsheet.xls.csv`?
> 
> 
> Note: filename.ext.ext.ext is still a valid filename, just wanted to flag that this might be something easily misunderstood.
> 
>     1. String validation/parsing
>        Should we add any cleaning of the input (e.g. other invalid filename scenarios) or trust that the developers doing this change understand these things.

@lb- : Thanks, added them to the tests now!

1. Yes. I've tried to flag up that this should be the filename without extensions in the documentation/example, but I'm very open to rewording this if you can think of something clearer.
2. I'd be inclined to trust developers on this - if it was coming from the admin I'd definitely suggest validation, but so far I don't think we've tended to do much validation of values in code. You could easily add an invalid value to `list_filter` in the same modeladmin without validation, for example.

lb-:
Thanks @KalobTaulien for merging in & @jacobtoppm  thanks for fixing the items up.

I agree with your answers, makes sense, I think the wording is fine as is.

PR #5401, opened 2019-06-19
Upgrade to Jest 23 for JS unit tests

thibaudcolas:
This upgrades the Jest version used in Wagtail. @helenb and I were having some issues updating snapshot tests because of a bug in Jest, that has been fixed in v23.0.0.

Jest 24 is out, but upgrading to that would also require we upgrade Babel from v6 to v7 – which is quite a bit of work.

After the upgrade, I also had to change how some of the Draftail initialisation tests were written. I couldn’t pinpoint which change caused that but I wouldn’t be surprised if `jsdom` had changed the behavior of how they parse script tags in HTML that caused those tests to break.

PR #6033, opened 2020-05-15
Switch Sass compilation from node-sass to dart-sass

thibaudcolas:
**Includes the changes from #5401**. I had trouble with dependencies shared between node-sass and Jest 22 (jsdom 11.5) – this resolves those problems quite conveniently.

This replaces gulp-sass with a fork using the new Dart implementation of Sass, https://github.com/sass/sass. This implementation is now the canonical implementation for Sass, so is more up to date (e.g. `@use`) – and also much easier to keep up to date on projects, since it’s compiled to JavaScript for the npm package, not needing native modules like `node-sass` does.

This should make it much easier for contributors to switch between working on a Vagrant box, Docker, their own computer – node-sass is our only dependency requiring native modules (there’s `fsevents` too but it’s optional). It also makes it easier to upgrade Node versions, since we don’t have to worry about node-sass having prebuilt binaries for the versions we upgrade to.

This also makes the Draftail.css compilation warning go away.

---

As part of the upgrade I had to rewrite two bits of code that were using now-deprecated Sass features. The output CSS for those two features. The output CSS is overall nearly identical – there are a few things that are ordered differently in a way that’s not significant, and a different level of precision for math, not in a way that’s significant.

I tested this with my UI regression test suite from https://github.com/thibaudcolas/wagtail-tooling.

---

* ~~[x] For Python changes: Have you added tests to cover the new/fixed behaviour?~~
* [x] For front-end changes: Did you test on all of Wagtail’s supported browsers? **Please list the exact versions you tested**.

---

Once this is merged we can remove: https://github.com/wagtail/docker-wagtail-develop/blob/eef59861b590ffabb54fd7c29271a4a39d9de071/Dockerfile.frontend#L9-L10

and might also be able to change: https://github.com/wagtail/vagrant-wagtail-develop/blob/master/Vagrantfile#L35.


lb-:
Merged in via 7eeb44ad0408cf72060f19db577de2159859cd74

PRs created as noted to remove node-sass specific build workarounds in docker/vagrant
* https://github.com/wagtail/docker-wagtail-develop/pull/11
* https://github.com/wagtail/vagrant-wagtail-develop/pull/27

Issue #1214, opened 2015-04-18
Some fields insert unnecessary markup into templates


JoshBarr:
Current:

``` html
<div class="rich_text"><p>hello world</p></div>
```

Ideal:

``` html
<p>hello world</p>
```

timheap:
This also happens with StreamBlocks, where each item is wrapped in a `<div class="block-text">`. I've yet to find a use for these blocks, and in fact have written code to *disable* them every time I use a StreamBlock.

gasman:
I think this comes under the category of "it seemed like a good idea at the time" :-)

As I recall, the rationale for this was that you often want to give images (and potentially other elements) CSS rules such as `max-width: 100%` within rich text areas, without having to make those rules universal for all images across the site. The assumption was that the extra div wouldn't interfere with the styling for people who didn't want to make use of it (unless their CSS was particularly odd!) and therefore it would be better to include it as standard, rather than require template developers to insert a wrapper element manually when they did need one.

(I think it may also have been related to the ongoing difficulty in ensuring that rich-text content is consistently wrapped in `<p>` tags - the wrapper element gives an extra place to fix up margins.)

However, it does break our rule of Wagtail not dictating your front-end markup.

Clearly this isn't going to be straightforward to change, given that there's plenty of existing CSS in the wild that relies on that element. We'd probably want a flag in settings to restore the old behaviour. (Going through the formal deprecation process would be even more of a faff - having the flag set to the old behaviour by default, and then explicitly setting it to new behaviour in the settings file of the project template. That's probably overkill.)


My gut feeling is that the situation with StreamBlock is a bit different - I think that's more akin to Django forms, where you get a certain default rendering if you output it as a single unit (`{{ self.body }}`), and if you don't like that, you have the option to iterate over it instead. Admittedly, I can't think of many situations where the `<div class="block-foo">` wrapper is absolutely necessary to distinguish between different blocks for CSS purposes - if you're working with StructBlocks then you'll be specifying your own template (which can include all the identifying element classes you want), so it's only relevant if you have, say, two different 'flavours' of RichTextBlock.

It's worth noting that if we make both of these proposed changes, then in the default rendering, there will be no difference in markup between a single RichTextBlock containing two paragraphs, and two consecutive RichTextBlocks with one paragraph each. I can't decide whether that's a good thing or not!

davecranwell:
[...]

PR #6036, opened 2020-05-16
Template render richtext, allows template override

allcaps:
Resolves https://github.com/wagtail/wagtail/issues/1214

* For new features: Has the documentation been updated accordingly? YES


wagtail/core/templates/wagtailcore/shared/richtext.html
```diff
@@ -0,0 +1 @@
+<div class="rich-text">{{ html|safe }} LALA</div>
```
zerolab, line 1: I love your test content :)

` LALA` makes the tests fail.

Other than that, the approach here is 🔥 for not adding yet another setting.

gasman:
Nice one @allcaps!
I think we should make a plan for dropping the `<div>` in the default configuration - enough people have complained about it that it wouldn't be right to leave it as a recipe that developers have to specifically seek out.

I'd be OK with making that change in a single release as we did for responsive embeds, rather than going through the full multi-release deprecation - in this case, that would mean removing the div from Wagtail's version of the template. If we do that, though, we should really make it as easy as possible for people to revert to the old behaviour, and getting people to set up a template override manually is probably too much of a hoop to jump through. So perhaps we should ship a contrib app that just contains a template override with the div in it, and tell people to add that to their installed apps if they want to keep the old behaviour?

(I have something similar in mind for when we drop SiteMiddleware, moving it to something like `wagtail.contrib.legacy.sitemiddleware` to give people an extra nudge to migrate away from it - although in this case there's less reason to actively discourage people from using it.)

allcaps:
This PR is updated. The suggestions from @gasman are applied:
- Default `richtext` filter does _not_ render the wrapper.
- Add `wagtail.contrib.legacy.richtext`to `INSTALLED_APPS` before `wagtail.core` to get the legacy markup.
- Tests are added.
- Docs are added.
- Tests pass.

docs/reference/contrib/index.rst
```diff
@@ -72,3 +73,9 @@ Provides a TableBlock for adding HTML tables to pages.
 -----------------------
 
 Provides a way to manage redirects.
+
+
+:doc:`legacy_richtext`
+-----------------------
+
+Provides legacy richtext wrapper.
```
zerolab, line 81:
```suggestion
Provides the legacy richtext wrapper (``<div class="rich-text"></div>``).
```

lb-:
Merged in via a1a2c35c1c1e78efa38bed0074ae0c7968770e84

lb-:
Epic stuff @allcaps  @zerolab  - awesome to see this in

PR #6040, opened 2020-05-18
WebP images quality/lossless compression parameters

mozgsml:
It will work after wagtail/Willow v1.4 is released


wagtail/images/tests/test_image_operations.py
```diff
@@ -580,7 +580,7 @@ def test_default_quality(self):
         save.assert_called_with(f, 'JPEG', quality=85, optimize=True, progressive=True)
 
     def test_jpeg_quality_filter(self):
-        fil = Filter(spec='width-400|jpegquality-40')
+        fil = Filter(spec='width-400|quality-40')
```
kaedroho, line 583: Is it possible that someone would want to specify different qualities for JPEG/WEBP?

wagtail/images/wagtail_hooks.py
```diff
@@ -116,7 +116,10 @@ def register_image_operations():
         ('width', image_operations.WidthHeightOperation),
         ('height', image_operations.WidthHeightOperation),
         ('scale', image_operations.ScaleOperation),
-        ('jpegquality', image_operations.JPEGQualityOperation),
+        # For backward compatibility
+        ('jpegquality', image_operations.QualityOperation),
```
kaedroho, line 120: Need a test to check that this filter still works

wagtail/images/wagtail_hooks.py
```diff
@@ -116,7 +116,10 @@ def register_image_operations():
...
+        ('quality', image_operations.QualityOperation),
+        ('webplossless', image_operations.WebPLossless),
```
kaedroho, line 122: I feel like this could be a nice extension to the `format` filter: `format-webp-lossless`

wagtail/images/tests/test_image_operations.py
mozgsml: Yes, it is possible. But keep in mind that you set this setting for a specific image and it is either JPEG or WebP. You would never use both of them at the same time.

kaedroho: That's the case if the template is only used in one place on the site. But if the template is used multiple times, you would have different images using the same filter.

It's probably fine the way it is though as long as the scale is always 1 - 100. But I wonder why the defaults are different for the two formats?

mozgsml: JPEG and WebP quality visually differs with the same quality setting. That's why it seemed more practical to choose different default settings. I agree with you. I will specify different filters for JPEG and WebP.

kaedroho: Thanks!

mozgsml:
Is my realization suitable?

wagtail/images/wagtail_hooks.py
```diff
@@ -117,6 +117,8 @@ def register_image_operations():
         ('height', image_operations.WidthHeightOperation),
         ('scale', image_operations.ScaleOperation),
         ('jpegquality', image_operations.JPEGQualityOperation),
+        ('webpquality', image_operations.WebPQualityOperation),
+        ('webplossless', image_operations.WebPLossless),
```
kaedroho, line 121: Is this tag redundant now that we have `format-webp-lossless`?

docs/topics/images.rst
```diff
@@ -349,20 +361,21 @@ done from the Django shell:
...
-It's also possible to have different JPEG qualities on individual tags by using
-the ``jpegquality`` filter. This will always override the default setting:
+It's also possible to have different JPEG and WebP qualities on individual tags
+by using ``jpegquality`` and ``webpquality`` filters. This will always override
+the default setting:
 
 .. code-block:: html+Django
 
-    {% image page.photo width-400 jpegquality-40 %}
+    {% image page.photo width-400 quality-40 %}
```
kaedroho, line 370: We need to change this example back to `jpegquality` and add an example with `webpquality`

docs/topics/images.rst
```diff
@@ -349,20 +361,21 @@ done from the Django shell:
...
 Note that this will have no effect on PNG or GIF files. If you want all images
-to be low quality, you can use this filter with ``format-jpeg`` (which forces
-all images to output in JPEG format):
+to be low quality, you can use this filter with ``format-jpeg`` or ``format-webp``
+(which forces all images to output in JPEG or WebP format):
 
 .. code-block:: html+Django
 
-    {% image page.photo width-400 format-jpeg jpegquality-40 %}
+    {% image page.photo width-400 format-jpeg quality-40 %}
```
kaedroho, line 378: Same as above

wagtail/images/image_operations.py
```diff
@@ -248,22 +248,43 @@ def construct(self, quality):
         self.quality = int(quality)
 
         if self.quality > 100:
-            raise ValueError("JPEG quality must not be higher than 100")
+            raise ValueError("Quality must not be higher than 100")
```
kaedroho, line 251: This could also go back the way it was before

docs/topics/images.rst
```diff
@@ -295,6 +296,16 @@ For example, to make the tag always convert the image to a JPEG, use ``format-jp
 
 You may also use ``format-png`` or ``format-gif``.
 
+WebP lossless
```
kaedroho, line 299: Lossless WebP

You can encode the image losslessly into WebP format by using the ``format-webp-lossless`` filter:

kaedroho:
Almost there now, just a handful of few minor tweaks

Thanks again!

pascalw:
Nice work @mozgsml! Looking forward to this.
Could this make it into 2.10?

pascalw:
@kaedroho it seems your remarks have been addressed, could you have a look again? Anything else I can help with to drive this home?

mozgsml:
@kaedroho I have addressed all your remarks, what do we do next?

kaedroho:
I've added release notes here: https://github.com/wagtail/wagtail/commit/8d84f6a4152ff82c908ed3bc590570a96b30c3a8

I've added your name to `CONTRIBUTORS` as "mozgsml". Let me know if you would like to use a different name

mozgsml:
I would prefer Nikolay Lukyanov

PR #6051, opened 2020-05-21
Update signals.rst for Python 3 compatibility

GTpyro:
The send_to_slack example was not compatible with Python 3. Example is updated using the requests & json libraries (instead of urllib & urllib2)

### DIFF
--- a/docs/reference/signals.rst
+++ b/docs/reference/signals.rst
@@ -26,8 +26,8 @@ example showing how you might notify your team when something is published:
 .. code-block:: python
 
     from wagtail.core.signals import page_published
-    import urllib
-    import urllib2
+    import requests
+    import json
 
 
     # Let everyone know when a new page is published
@@ -41,9 +41,8 @@ example showing how you might notify your team when something is published:
             "icon_emoji": ":octopus:"
         }
 
-        data = urllib.urlencode(values)
-        req = urllib2.Request(url, data)
-        response = urllib2.urlopen(req)
+        data = json.dumps(values)
+        response = requests.post(url, data)
 
     # Register a receiver
     page_published.connect(send_to_slack)
--- a/docs/reference/signals.rst
+++ b/docs/reference/signals.rst
@@ -27,7 +27,6 @@ example showing how you might notify your team when something is published:
 
     from wagtail.core.signals import page_published
     import requests
-    import json
 
 
     # Let everyone know when a new page is published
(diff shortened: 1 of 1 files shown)


docs/reference/signals.rst
```diff
@@ -41,9 +41,8 @@ example showing how you might notify your team when something is published:
             "icon_emoji": ":octopus:"
         }
 
-        data = urllib.urlencode(values)
-        req = urllib2.Request(url, data)
-        response = urllib2.urlopen(req)
+        data = json.dumps(values)
```
marteinn, line 44: A small note, the `json.dumps` step is redundant as `requests.post` accepts json serializable datatypes (such as dict). See this example: https://requests.readthedocs.io/en/master/user/quickstart/#more-complicated-post-requests

docs/reference/signals.rst
```diff
@@ -26,8 +26,8 @@ example showing how you might notify your team when something is published:
 .. code-block:: python
 
     from wagtail.core.signals import page_published
-    import urllib
-    import urllib2
+    import requests
```
marteinn, line 29: Could you add a comment that indicates that requests is not a part of the standard libraries? See the geocoder example here https://docs.wagtail.io/en/stable/advanced_topics/customisation/page_editing_interface.html#customising-generated-forms

tomdyson: `requests` is required by Wagtail, so it might not be important to make the distinction here:

https://github.com/wagtail/wagtail/blob/master/setup.py#L36

tomdyson:
Thank you for the PR, @GTpyro, and for your review, @marteinn!

docs/reference/signals.rst
tomdyson: `requests` is required by Wagtail, so the distinction may not be important here:

https://github.com/wagtail/wagtail/blob/master/setup.py#L36

marteinn: @tomdyson is right, requests is always present. Let's skip this requirement.

docs/reference/signals.rst
GTpyro: thanks for the suggestion @marteinn! I learned something new today. I'll make the change right now and resubmit

marteinn:
@GTpyro I'm about to merge this, but before I do I want to know your preferred name so I can add you to our https://github.com/wagtail/wagtail/blob/master/CONTRIBUTORS.rst document :)

marteinn:
@GTpyro I'm using your github name for now, just let me know if you want it changed!

marteinn:
Merged in https://github.com/wagtail/wagtail/commit/147fda12fabb681cf0c2a80b7a4d9b11b06f589b

Issue #5454, opened 2019-07-25
AdminDateTime widget problem with localized datetimes


roodie:
### Issue Summary

If I have the locale/datetime format settings set according to my native language (Hungarian), the AdminDateTime widget stops working properly. 

Every time  I enter the field the date and time selectors pops up, I select a date and time, and seemingly everything is correct,  but as soon as I leave the field, it writes back the current date and time.

The formats are actually correct - the initial current date and time, the one after I use the date and time picker, and the current date and time written back, all are in the expected  "2019.07.25. 11:15" format.

The separate Date and Time fields are working properly though.

### Steps to Reproduce

1. Start a new project as described in the documentetion
2. Set the LANGUAGE_CODE in settings to 'hu-HU'
3. Make sure that both USE_I18N and USE_L10N is set to True
4. Add the following 2 lines to settings:

```python 
  WAGTAIL_DATETIME_FORMAT = '%Y.%m.%d. %H:%M'
  DATETIME_INPUT_FORMATS = ['%Y.%m.%d. %H:%M']
```
5. Login into admin and try to schedule the Home page

### Technical details

* Python version: 3.6.5
* Django version: 2.2.3
* Wagtail version: 2.5.1
* Browser version: Googe Chrome/75.0.3770.142

cyface:
I have the same problem with USA AM/PM datetime formats:

DATETIME_INPUT_FORMATS = ["%m-%d-%Y %I:%M %p"]
WAGTAIL_DATETIME_FORMAT = "%m-%d-%Y %I:%M %p"

But rather than reverting to the current date and time in the field, it subtracts an hour from whatever I picked when I click out of the field.

Also, the date/time picker doesn't switch to AM/PM correctly, but the format looks correct in the field.

linuxsoftware:
Do you have USE_L10N set to True?  If so Django will prefer your locale format's DATETIME_INPUT_FORMATS over your own. See https://docs.djangoproject.com/en/2.2/topics/i18n/formatting/.

cyface:
@linuxsoftware Unfortunately, the same behavior happens with USE_L10N set to False. :(

vupham04:
I try to replicate the bug. Something that I have noticed:

1. The problem seems not to be because of `LANGUAGE_CODE`. I tried 'en-US' instead of 'hu-HU' and it returns the same result.
[...]

Issue #4846, opened 2018-10-25
Local date time issue for AdminDateTimeInput


robslotboom:
### Issue Summary

After updating to wagtail 2.3 my local date time works only partial.

#### settings.py

```
TIME_ZONE = 'Europe/Amsterdam'
LANGUAGE_CODE = 'nl'
LANGUAGES = WAGTAILADMIN_PERMITTED_LANGUAGES = [
	('nl', 'Nederlands'),
]
USE_I18N = True
USE_L10N = True
USE_TZ = True

WAGTAIL_DATETIME_FORMAT = '%d %b %Y %H:%M'

```
#### formats.py for dutch

```
DATETIME_INPUT_FORMATS = [
    '%d-%m-%Y %H:%M:%S', '%d-%m-%y %H:%M:%S', '%Y-%m-%d %H:%M:%S',
    '%d/%m/%Y %H:%M:%S', '%d/%m/%y %H:%M:%S', '%Y/%m/%d %H:%M:%S',
    '%d-%m-%Y %H:%M:%S.%f', '%d-%m-%y %H:%M:%S.%f', '%Y-%m-%d %H:%M:%S.%f',
    '%d/%m/%Y %H:%M:%S.%f', '%d/%m/%y %H:%M:%S.%f', '%Y/%m/%d %H:%M:%S.%f',
    '%d-%m-%Y %H.%M:%S', '%d-%m-%y %H.%M:%S',     # '20-01-2009 15.23:35', '20-01-09 15.23:35'
    '%d/%m/%Y %H.%M:%S', '%d/%m/%y %H.%M:%S',     # '20/01/2009 15.23:35', '20/01/09 15.23:35'
    '%d-%m-%Y %H.%M:%S.%f', '%d-%m-%y %H.%M:%S.%f',   # '20-01-2009 15.23:35.000200', '20-01-09 15.23:35.000200'
    '%d/%m/%Y %H.%M:%S.%f', '%d/%m/%y %H.%M:%S.%f',   # '20/01/2009 15.23:35.000200', '20/01/09 15.23:35.000200'
    '%d-%m-%Y %H:%M', '%d-%m-%y %H:%M', '%Y-%m-%d %H:%M',   # '20-01-2009 15:23', '20-01-09 15:23', '2009-01-20 15:23'
[...]

PR #6057, opened 2020-05-22
Timepicker fixes

jacobtoppm:
- Resolves the issue raised by @cyface on the first comment of https://github.com/wagtail/wagtail/issues/5454, where the datetimepicker would reduce the time selected by an hour each time it lost focus when using 12 hour times
- Adds a new setting, `WAGTAIL_TIME_FORMAT` to specify the admin time input format in both the picker widget and field, allowing 12 hour times to be used in the widget as well as the output, for instance
- Makes the timepicker widget time column flexible-width, to cope with longer time formats
- Allows `USE_L10N=True` to override the default format with a locale-specific one

### DIFF
--- a/wagtail/admin/widgets.py
+++ b/wagtail/admin/widgets.py
@@ -47,10 +47,7 @@ class AdminDateInput(widgets.DateInput):
         if attrs:
             default_attrs.update(attrs)
         if fmt is None:
-            if getattr(settings, 'USE_L10N', False):
-                fmt = get_format('DATE_INPUT_FORMATS')[0]
-            else:
-                fmt = getattr(settings, 'WAGTAIL_DATE_FORMAT', DEFAULT_DATE_FORMAT)
+            fmt = getattr(settings, 'WAGTAIL_DATE_FORMAT', DEFAULT_DATE_FORMAT)
         self.js_format = to_datetimepicker_format(fmt)
         super().__init__(attrs=default_attrs, format=fmt)
 
@@ -81,10 +78,7 @@ class AdminTimeInput(widgets.TimeInput):
             default_attrs.update(attrs)
         fmt = format
         if fmt is None:
-            if getattr(settings, 'USE_L10N', False):
-                fmt = get_format('TIME_INPUT_FORMATS')[0]
-            else:
-                fmt = getattr(settings, 'WAGTAIL_TIME_FORMAT', DEFAULT_DATETIME_FORMAT)
+            fmt = getattr(settings, 'WAGTAIL_TIME_FORMAT', DEFAULT_TIME_FORMAT)
         self.js_format = to_datetimepicker_format(fmt)
         super().__init__(attrs=default_attrs, format=fmt)
 
@@ -115,16 +109,10 @@ class AdminDateTimeInput(widgets.DateTimeInput):
         if attrs:
             default_attrs.update(attrs)
         if fmt is None:
-            if getattr(settings, 'USE_L10N', False):
-                fmt = get_format('DATETIME_INPUT_FORMATS')[0]
-            else:
-                fmt = getattr(settings, 'WAGTAIL_DATETIME_FORMAT', DEFAULT_DATETIME_FORMAT)
+            fmt = getattr(settings, 'WAGTAIL_DATETIME_FORMAT', DEFAULT_DATETIME_FORMAT)
         time_fmt = time_format
         if time_fmt is None:
-            if getattr(settings, 'USE_L10N', False):
-                time_fmt = get_format('TIME_INPUT_FORMATS')[0]
-            else:
-                time_fmt = getattr(settings, 'WAGTAIL_TIME_FORMAT', DEFAULT_TIME_FORMAT)
+            time_fmt = getattr(settings, 'WAGTAIL_TIME_FORMAT', DEFAULT_TIME_FORMAT)
         self.js_format = to_datetimepicker_format(fmt)
         self.js_time_format = to_datetimepicker_format(time_fmt)
         super().__init__(attrs=default_attrs, format=fmt)
--- a/wagtail/admin/widgets.py
+++ b/wagtail/admin/widgets.py
@@ -23,6 +23,7 @@ from wagtail.utils.widgets import WidgetWithScript
 
 DEFAULT_DATE_FORMAT = '%Y-%m-%d'
 DEFAULT_DATETIME_FORMAT = '%Y-%m-%d %H:%M'
+DEFAULT_TIME_FORMAT = '%H:%M'
 
 
 class AdminAutoHeightTextInput(widgets.Textarea):
@@ -46,7 +47,10 @@ class AdminDateInput(widgets.DateInput):
         if attrs:
             default_attrs.update(attrs)
         if fmt is None:
-            fmt = getattr(settings, 'WAGTAIL_DATE_FORMAT', DEFAULT_DATE_FORMAT)
+            if getattr(settings, 'USE_L10N', False):
+                fmt = get_format('DATE_INPUT_FORMATS')[0]
+            else:
+                fmt = getattr(settings, 'WAGTAIL_DATE_FORMAT', DEFAULT_DATE_FORMAT)
         self.js_format = to_datetimepicker_format(fmt)
         super().__init__(attrs=default_attrs, format=fmt)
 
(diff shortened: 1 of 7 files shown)


wagtail/admin/widgets.py
```diff
@@ -71,11 +75,29 @@ def media(self):
 class AdminTimeInput(widgets.TimeInput):
     template_name = 'wagtailadmin/widgets/time_input.html'
 
-    def __init__(self, attrs=None, format='%H:%M'):
-        default_attrs = {'autocomplete': 'off'}
+    def __init__(self, attrs=None, format=None):
+        default_attrs = {'autocomplete': 'new-time'}
```
gasman, line 79: autocomplete="off" is the current preferred attribute, as per 928e79ed3a119bf376a69b613517c6523100068b.

jacobtoppm: Thanks for catching! Must have accidentally overridden this: I've changed back in the new fixup commit.

wagtail/admin/widgets.py
```diff
@@ -46,7 +47,10 @@ def __init__(self, attrs=None, format=None):
         if attrs:
             default_attrs.update(attrs)
         if fmt is None:
-            fmt = getattr(settings, 'WAGTAIL_DATE_FORMAT', DEFAULT_DATE_FORMAT)
+            if getattr(settings, 'USE_L10N', False):
+                fmt = get_format('DATE_INPUT_FORMATS')[0]
```
gasman, line 51: Letting `DATE_INPUT_FORMATS` take precedence over `WAGTAIL_DATE_FORMAT` would make `WAGTAIL_DATE_FORMAT`'s behaviour consistent with Django's `DATE_FORMAT`, but it is a breaking change that would affect existing projects (see for example #4846 which involves setting Wagtail's format to an entry which is not the first one in the list). Will need to ponder some more to decide if this is a positive change or not - worth noting that this would rule out the possibility of setting Wagtail's format to something different to the Django default (to allow dates to be presented differently in the Wagtail admin vs the site frontend).

jacobtoppm: I see what you mean, but from my testing at least, using `USE_L10N=True` with a `WAGTAIL_DATE_FORMAT` (and without this override) just made the field unusable as the validation would break - the advice in the issues I've seen is to turn that setting off to fix it, which seems awkward? Happy to be corrected on this - perhaps I've had something setup improperly, but that's why I added this L10N specific override.

gasman: My understanding is that `WAGTAIL_DATE_FORMAT` does work with `USE_L10N=True` enabled, but [as per the docs](https://docs.wagtail.io/en/stable/reference/settings.html#date-and-datetime-inputs) the chosen format needs to be one of the formats in the `DATE_INPUT_FORMATS` list, and with `USE_L10N` enabled there are extra hoops to jump through to find out what that list is (and doubly so if you actually want to override it).

What we could do is add a startup check that warns if `WAGTAIL_DATE_FORMAT` has been set to something not on the `DATE_INPUT_FORMATS` list, and perhaps have our documentation emphasise a bit more that if you intend to change `WAGTAIL_DATE_FORMAT` from its default, you'll probably have to faff around with locales...?

gasman:
Other than `USE_L10N` overriding the `WAGTAIL_*_FORMAT` settings, these changes look good to me and are working nicely in my tests. If you're OK with the alternative suggestion of adding a startup check, I'd be happy to go ahead and implement that.

jacobtoppm:
@gasman Yes, I'd be very happy with that suggestion - if you're happy to add that would be very welcome (thanks!) but otherwise happy to implement as well.

gasman:
Rebased and checks added as discussed in 86377db8d0f0ccb048ca79537c6285c4cc957234 - @jacobtoppm, please can you re-review?

wagtail/admin/checks.py
```diff
@@ -150,3 +152,64 @@ def check_panels_in_model(cls, context='model'):
...
+
+        errors.append(
+            Warning(
+                "WAGTAIL_DATE_FORMAT is not a recognised input format",
+                hint=error_hint,
+                id='wagtailadmin.W003',
+            )
+        )
+
+    datetime_input_formats = get_format('DATETIME_INPUT_FORMATS')
```
jacobtoppm, line 179: Generally this looks great. However, `get_format` only gets it for the current locale/language - if a user sets up a language preference that isn't the server default in their profile for the Wagtail Admin, I believe it would still be possible for validation to fail without this check catching it. Potentially we could use the `lang=` kwarg to loop over the allowed user profile languages and check `WAGTAIL_*_FORMAT` is consistent with all if `USE_L10N=True`?

jacobtoppm:
> Rebased and checks added as discussed in [86377db](https://github.com/wagtail/wagtail/commit/86377db8d0f0ccb048ca79537c6285c4cc957234) - @jacobtoppm, please can you re-review?

@gasman Re-reviewed now - let me know your thoughts on the admin localization concern above, but otherwise looks good.

wagtail/admin/checks.py
gasman: Aaah, that's annoying - I hadn't made the connection that the *_INPUT_FORMATS lists are liable to vary on a per-user basis. Effectively, this means that the only way to do something not-totally-standard with date formats (e.g. 12-hour times) while USE_L10N is enabled is to supply a custom locale file (as per https://github.com/wagtail/wagtail/issues/4846#issuecomment-433618730) for every locale that a user might conceivably be using.

I think your original tack was the logically correct one then - having WAGTAIL_*_FORMAT configurable as a global setting is fundamentally at odds with USE_L10N. There's a fair argument to be made that anyone who enables Django locale support is disqualifying themselves from being able to say "I want times to display in 12-hour format", unless what they _really_ mean is "I want to provide my users with one or more custom locale options that use 12-hour times" :-) I guess we'll need to take some time to figure out the best compromise between what Django conventions dictate, and what Wagtail users already expect to be able to do.

jacobtoppm: How would you feel about merging the uncontroversial part of this PR (ie minus the `USE_L10N` exception) for now until we make that decision, so we get at least the 12 hour, L10N off, support in 2.10? That way we're not changing Wagtail's behaviour from standard on this issue (though I agree we need a better solution soon, because right now I agree the global setting approach just doesn't work with L10N). Then maybe discuss the L10N issue in the next core team meeting and I can create a separate PR with whatever we decide?

gasman: Sounds good 👍

jacobtoppm: Great, thanks! Are you happy to tweak and merge or shall I?

gasman: I'll do that now.

allcaps:
@gasman and @jacobtoppm 

I did an attempt to add a stat-up check.

```
# wagtail/core/apps.py
class WagtailCoreAppConfig(AppConfig):
    name = 'wagtail.core'
    label = 'wagtailcore'
    verbose_name = _("Wagtail core")

    def ready(self):
        from wagtail.core.signal_handlers import register_signal_handlers
        register_signal_handlers()

        if settings.USE_I18N:
            for code, label in settings.LANGUAGES:
                with translation.override(code):
                    if settings.WAGTAIL_TIME_FORMAT not in settings.TIME_INPUT_FORMATS:
                        raise ImproperlyConfigured(
                            f"WAGTAIL_TIME_FORMAT {settings.WAGTAIL_TIME_FORMAT} "
                            f"must be in TIME_INPUT_FORMATS for language {code} {label}."
                        )

```

This works for settings defined in settings. But I also like it to work with [formats.py](https://docs.djangoproject.com/en/3.0/topics/i18n/formatting/#creating-custom-format-files).
This makes it possible to define multiple date/time settings, one per language. With that one could resolve setting conflicts.

# Config

```
LANGUAGE_CODE = 'en'
TIME_ZONE = 'UTC'
USE_I18N = True
USE_L10N = True
USE_TZ = True

from django.utils.translation import gettext_lazy as _

LANGUAGES = [
    ('nl', _('Dutch')),
    ('en', _('English')),
]

FORMAT_MODULE_PATH = [
    'bakerydemo.formats',
]

WAGTAIL_TIME_FORMAT = '%H:%M'
TIME_INPUT_FORMATS = ['%H:%M']
```
Formats files:
```
.
│   ├── formats
│   │   ├── __init__.py
│   │   ├── en
│   │   │   ├── __init__.py
│   │   │   └── formats.py
│   │   └── nl
│   │       ├── __init__.py
│   │       └── formats.py
```

Issue #6059, opened 2020-05-22
Automatic code formatting with Prettier


yhoiseth:
Extracted from #6056 where @thibaudcolas wrote:

> For reference, we’ve also discussed using Prettier as well (for JS, SCSS, Markdown, JSON, YAML). Here I think it’s just a matter of someone putting a small proposal together (in a new issue), and then making the change – we’re all in agreement we’d love to see this happen.

I’ll give it a go.

yhoiseth:
Okay… I’m getting some mysterious CI failures in #6061. Going to try with a more incremental approach.

gasman:
Those CI failures are from a separate ongoing issue - #6063

yhoiseth:
@gasman I see. Thanks for letting me know 🙂

thibaudcolas:
Hey @yhoiseth, just looking at your PRs – at this stage I think it’s too early to make PRs for this. As I said in my comment what we really need right now is a proposal so we can discuss what we want to do.

If we take the black DEP as a reference – I think what we need is just the [Reference implementation](https://github.com/django/deps/blob/master/accepted/0008-black.rst#reference-implementation), with details of the steps involved.

Here’s what to cover:

- How we should configure Prettier, based on our existing styleguide(s) and best practices
- Updating our ESLint and stylelint configurations to be compatible
- Formatting the code – how we do (personally I’d rather have this *not* done in the same PR as with the other changes, as it will make it go outdated all the time)
- Enforcement - where do we run Prettier in CI, and what changes do we need to make
- Documentation – where do we need to upgrade our code styleguide, and where do we need instructions on Prettier usage.

yhoiseth:
Ah, I see. I should have read your comment more thoroughly. 

As a first step, I think it would be useful to know the _why_ here. What is insufficient about the current setup?

thibaudcolas:
👍 I think we can pretty much paraphrase Django’s black DEP in this case, along with Prettier’s own "Why" https://prettier.io/docs/en/why-prettier.html.

thibaudcolas:
Merged!

Here are the updated steps I would suggest to rebase existing PRs with the reformatting:

```
# Make a copy of your branch
git branch save/my-existing-branch my-existing-branch
# Make sure you have the latest `main` from Wagtail
git remote add upstream git@github.com:wagtail/wagtail.git
git checkout main
git pull upstream main
# Rebase your work onto the commit preceding the reformatting to resolve conflicts that would have been present anyway.
[...]

PR #6061, opened 2020-05-22
6059 prettier

yhoiseth:
See #6059.


thibaudcolas:
It looks like 	#6065 replaces this so I’ll close this PR. Please don’t leave PRs open if they are replaced by other ones.

PR #6063, opened 2020-05-22
Fix incompatibility issue with django-taggit 1.3.0

FlipperPA:
`django-taggit` 1.3.0 was released today, which causes an incompatibility issue with Wagtail, as it introduced new arguments to the API. See: https://github.com/coderedcorp/coderedcms/issues/321

This PR locks Wagtail to < 1.3.0 as a quick fix.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ except ImportError:
 install_requires = [
     "Django>=2.1,<3.1",
     "django-modelcluster>=5.0,<6.0",
-    "django-taggit>=1.0,<2.0",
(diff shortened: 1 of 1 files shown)


rgs258:
nice

gasman:
Thanks @FlipperPA! The fix for this issue is now released as django-modelcluster 5.0.2, so any environments that don't have dependencies pinned to exact versions (and would thus have been affected by the 1.3.0 breakage) should now pick up the fixed version.

Have opened #6071 to address the remaining test failure (which is specific to the test suite and will not affect real-world installations).

Issue #2893, opened 2016-08-04
Add support for readonly fields (ReadOnlyFieldPanel) on modeladmin


rafaponieman:
I believe the easiest way would be to add a ReadOnlyFieldPanel and style it accordingly. Only thing would be that it should somehow be able to use a callable on the model or admin itself instead of an actual model field the same way readonly_fields does.

melinath:
Ugly workaround. Only works with fields, not methods etc. Also missing other fields, so YMMV, but hey, it's something.

```python
# edit_handlers.py
from django.forms.utils import pretty_name
from django.template.loader import render_to_string
from django.utils.safestring import mark_safe
from wagtail.wagtailadmin.edit_handlers import EditHandler


class BaseReadOnlyPanel(EditHandler):
    def __init__(self, instance=None, form=None):
        super(BaseReadOnlyPanel, self).__init__(instance=instance, form=form)
        self.heading = pretty_name(self.field_name)

    template = 'core/edit_handlers/read_only_panel.html'

    def render(self):
        return mark_safe(render_to_string(self.template, {
            'self': self,
            'value': getattr(self.instance, self.field_name),
        }))


class ReadOnlyPanel(object):
    def __init__(self, field_name):
        self.field_name = field_name

    def bind_to_model(self, model):
        base = {
            'model': model,
            'field_name': self.field_name,
        }

        return type(str('_ReadOnlyPanel'), (BaseReadOnlyPanel,), base)
```

If you're using this in a ModelAdmin, you can then use this for the model by overriding the view(s) with custom versions like this:

```python
# views.py
from wagtail.contrib.modeladmin.views import EditView
from wagtail.wagtailadmin.edit_handlers import (
    ObjectList,
    FieldPanel,
    get_form_for_model,
)

from edit_handlers import ReadOnlyPanel


class MyModelEditView(EditView):
    def get_form_class(self):
        return get_form_for_model(
            MyModel,
            fields=['readable_field'],
        )

    def get_edit_handler_class(self):
        edit_handler = ObjectList([
            ReadOnlyPanel('unreadable_field'),
            FieldPanel('readable_field'),
        ])
        return edit_handler.bind_to_model(self.model)
```

```python
# wagtail_hooks.py
from wagtail.contrib.modeladmin.options import (
    ModelAdmin,
    modeladmin_register,
)

from views import MyModelEditView


class MyModelAdmin(ModelAdmin):
    edit_view_class = MyModelEditView


[...]

PR #6067, opened 2020-05-23
Add ReadOnlyPanel

vmavromatis:
https://github.com/wagtail/wagtail/issues/2893

### DIFF
--- a/wagtail/admin/edit_handlers.py
+++ b/wagtail/admin/edit_handlers.py
@@ -8,6 +8,7 @@ from django.forms.formsets import DELETION_FIELD_NAME, ORDERING_FIELD_NAME
 from django.forms.models import fields_for_model
 from django.template.loader import render_to_string
 from django.utils.functional import cached_property
+from django.utils.html import format_html
 from django.utils.safestring import mark_safe
 from django.utils.translation import gettext_lazy
 from taggit.managers import TaggableManager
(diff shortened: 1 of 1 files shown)


wagtail/admin/edit_handlers.py
```diff
@@ -844,3 +845,38 @@ def id_for_label(self):
...
+            heading=self.heading,
+            classname=self.classname,
+            help_text=self.help_text,
+        )
+
+    def render(self):
+        value = getattr(self.instance, self.attr)
+        if callable(value):
+            value = value()
+        return format_html('<div style="padding-top: 1.2em;">{}</div>', value)
```
zerolab, line 867: Can you please move styling a more relevant place?

kaedroho:
Thanks @vmavromatis!

I'm intrigued to know how to use this panel, please could you add an example into the [panels docs](https://github.com/wagtail/wagtail/blob/master/docs/reference/pages/panels.rst)?

wagtail/admin/edit_handlers.py
brylie: @zerolab what is a specific suggestion?

brylie: Are there any CSS helper classes that could be used here instead?

zerolab: A tough one. I would add it as an utility class under `client/scss/overrides/_utilities.legacy.scss`

brylie: What about just removing the style for now?

```suggestion
        return format_html('<div>{}</div>', value)
```

brylie:
Ping @vmavromatis

wagtail/admin/edit_handlers.py
ababic: @thibaudcolas Would you be able to suggest the best place to add a style for this, and a sensible class name? It's a while since I've done anything with static files in Wagtail.

bedrockjohan:
Gave this a shot and worked very nice until I used it on a nested `InlinePanel`.

```python
class Order(ClusterableModel):
    # additional order fields
    panels = [
        # additional edit handlers from the Order
        edit_handlers.MultiFieldPanel([
            edit_handlers.InlinePanel("lines")
        ], heading=_("Order lines"))
    ]

class OrderLine(ClusterableModel):
    order = ParentalKey(Order, on_delete=models.CASCADE, related_name="lines")
    quantity = models.PositiveSmallIntegerField(_("quantity"), default=1)
    # more order line product details

    panels = [
        # more edit handlers for product details
        edit_handlers.FieldPanel("quantity"),
        edit_handlers.InlinePanel("fields"),
    ]

class OrderLineField(models.Model):
    order_line = ParentalKey(OrderLine, on_delete=models.CASCADE, related_name="fields")
    field = models.ForeignKey("some field model")
    value = models.CharField(_("value"), max_length=255, blank=True)

    panels = [
        ReadOnlyFieldPanel("field"),
        ReadOnlyFieldPanel("value"),
    ]

```

This works but the nested `InlinePanel` renders the fields twice. Once as read only and once as only an input.
image

Defining the panels on the `OrderLineField` model or as a `panels` argument on  the `InlinePanel` makes no difference.

thibaudcolas:
Replaced by #6194.

PR #6068, opened 2020-05-24
Fix that proper document icon are shown

marteinn:
I stumbled upon this while researching another issue, it seems like the header for document multiple add (`/admin/images/multiple/add/`) referred to a non-existing icon. This commit makes sure it uses the same class name as the rest of the document headers.

Before:
Screenshot 2020-05-24 at 15 36 33
 
After:
Screenshot 2020-05-24 at 15 46 37


marteinn:
Merged in https://github.com/wagtail/wagtail/commit/f4542c8ffc93633c0e8751bd2860f440d44b1fa7

PR #6070, opened 2020-05-25
Hide "add-buttons" in StreamBlock when 'max_num' child blocks is reached

Maxxen:
This change hides the add-buttons in the `StreamBlock` interface when `max_num` number of child blocks is reached. This also adds two new callbacks to sequence.js, onDisableAdd and onEnableAdd that is called whenever the number of members reach or dip below the maximum number of allowed children.

Setting `max_num` to 1 now makes it simple and intuitive to create the commonly requested "choose one"-style block, which allows users to pick one block from a set. This is functionally similar to the type of conditional rendering other CMS-interfaces regularly provide.

If #2379 or #3914 ever gets resolved, it would be trivial to make the ListBlock interface behave similarly by hooking into the callbacks.

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/js/blocks/list.js
+++ b/wagtail/admin/static_src/wagtailadmin/js/blocks/list.js
@@ -9,6 +9,7 @@
         return function(elementPrefix) {
             var sequence = Sequence({
                 prefix: elementPrefix,
+				maxNumChildBlocks: Infinity,
                 onInitializeMember: function(sequenceMember) {
                     /* initialize child block's JS behaviour */
                     if (opts.childInitializer) {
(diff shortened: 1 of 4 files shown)


Maxxen:
@gasman Fixed now! I'm not used to working with rebase so let me know if I messed up.

Maxxen:
Ok, reversed the rebase since something was obviously wrong, now only contains my changes

gasman:
Thanks again @Maxxen! Merged in 4afdbdea5609cfcbb648bf53db23c3dd488425fe (with one last-minute tweak - added a check for the presence of `onEnableAdd` / `onDisableAdd` before calling them).

PR #6073, opened 2020-05-26
A couple of minor tweaks to the admin search

kaedroho:
Set the `zero_terms` parameter to `MATCH_ALL` so filters in the query string run if there are no search terms.
Remove the quotes from around the query string in the "Sorry, no pages match" message. This makes the output less messy when doing phrase searches.

### DIFF
--- a/wagtail/admin/views/pages.py
+++ b/wagtail/admin/views/pages.py
@@ -1033,7 +1033,7 @@ def search(request):
             pagination_query_params['q'] = q
 
             # Parse query
-            filters, query = parse_query_string(q, operator='and')
+            filters, query = parse_query_string(q, operator='and', zero_terms=MATCH_ALL)
 
             # Live filter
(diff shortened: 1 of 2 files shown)


wagtail/admin/views/pages.py
```diff
@@ -29,7 +29,7 @@
 from wagtail.admin.navigation import get_explorable_root_page
 from wagtail.core import hooks
 from wagtail.core.models import Page, PageRevision, UserPagePermissionsProxy
-from wagtail.search.query import MATCH_ALL
+from wagtail.search.query import MATCH_ALL, MATCH_NONE
```
zerolab, line 32: is `MATCH_NONE` an unused import?

kaedroho: Yep, fixed now, thanks!

gasman:
Thanks @kaedroho!

PR #6075, opened 2020-05-27
Fix RestaurantTag migration tests introducing an extra migration

thibaudcolas:
Follow-up for https://github.com/wagtail/django-modelcluster/pull/123, fixing another change in django-taggit 1.3.0 that breaks our tests. For this one I’m pretty sure this just breaks Wagtail’s tests but has no other consequences for websites (well, except for the extra migration they’ll get as well).

Since those migrations are just here for tests I’ve assumed it was ok to edit the existing migration as the only change is with `verbose_name`.

Change from django-taggit: https://github.com/jazzband/django-taggit/commit/90c7224018c941b9a260c8e8bed166536f5870df

---

* ~[ ] For front-end changes: Did you test on all of Wagtail’s supported browsers? **Please list the exact versions you tested**.~
* ~[ ] For new features: Has the documentation been updated accordingly?~

### DIFF
--- a/wagtail/tests/testapp/migrations/0047_restaurant_tags.py
+++ b/wagtail/tests/testapp/migrations/0047_restaurant_tags.py
@@ -28,8 +28,8 @@ class Migration(migrations.Migration):
             name='RestaurantTag',
             fields=[
                 ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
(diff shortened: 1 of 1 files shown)


thibaudcolas:
🎉 that was quick!

gasman:
I already fixed this in #6071 :-) I don't think there's any problem with having both versions of the fix in place, though.

thibaudcolas:
Damn, sorry! I must have been running with tests with an out-of-date `master`.

PR #6078, opened 2020-05-27
Drop null=True from Site.site_name

allcaps:
> Avoid using null on string-based fields such as CharField and TextField. If a string-based field has null=True, that means it has two possible values for “no data”: NULL, and the empty string. In most cases, it’s redundant to have two possible values for “no data;” the Django convention is to use the empty string, not NULL. One exception is when a CharField has both unique=True and blank=True set. In this situation, null=True is required to avoid unique constraint violations when saving multiple objects with blank values.

https://docs.djangoproject.com/en/3.0/ref/models/fields/#django.db.models.Field.null

* Do the tests still pass? Y
* Does the code comply with the style guide? Y
* For Python changes: Have you added tests to cover the new/fixed behaviour? N
* For new features: Has the documentation been updated accordingly? N


lb-:
Merged in via eb9cff7bf388735e988bfd9f084d7a5d34f0ba42
Thanks heaps!

Issue #6076, opened 2020-05-27
AdminModel list_export doesn't work for methods


Cheradenine:
Found a bug? Please fill out the sections below. 👍

### Issue Summary
2.9 introduced list_export in ModelAdmin. Ideally should work like list_display, but crashes if tuple contains ModelAdmin methods

### Steps to Reproduce
1. Create a new ModelAdmin and register it (e.g. for Contact model)
2. Define a lookup method on ModelAdmin class similar to list_display
`
def city(self, obj):
    return obj.address.city
`
3. Include list_export
`
list_export = (..., 'city') # same as list_display
`
4. Load in admin and try to export to csv or xlsx

Result:
exception thrown in
`
File "/Users/me/project/lib/python3.7/site-packages/django/contrib/admin/utils.py", line 341, in label_for_field
    raise AttributeError(message)
AttributeError: Unable to lookup 'city' on Contact
`
I think that the problem is in [IndexView.get_heading](https://github.com/wagtail/wagtail/blob/314a926f75f96275ccd9c84a568d7e588f13b258/wagtail/contrib/modeladmin/views.py#L277)

When calling label_for_field it should pass self.model_admin? (have not verified that yet)

Expected list_export to work like list_display.

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: no

### Technical details

* Python version: 3.7.7
* Django version: 3.0.6
* Wagtail version: 2.9
* Browser version: Chrome 83.0.4103.61 (Official Build) (64-bit)

PR #6081, opened 2020-05-27
Fix for IndexView to allow AdminModel methods for fields #6076

Cheradenine:
Fixes issue #6076
For xlsx and csv export, list_export should behave like list_display. The bug was that the model_admin wasn't passed along to allow field lookups to be resolved by model_admin methods.

- Added method to do lookup on Book model to Author.date_of_birth
- Added tests to cover this case
- Verified that test fail
- Added fix in IndexView.get_heading
- Verified that tests now pass

### DIFF
--- a/wagtail/contrib/modeladmin/views.py
+++ b/wagtail/contrib/modeladmin/views.py
@@ -274,7 +274,7 @@ class IndexView(SpreadsheetExportMixin, WMABaseView):
         heading_override = self.export_headings.get(field)
         if heading_override:
             return force_str(heading_override)
-        return force_str(label_for_field(field, model=self.model).title())
+        return force_str(label_for_field(field, model=self.model, model_admin=self.model_admin).title())
 
     def to_row_dict(self, item):
         """ Returns an OrderedDict (in the order given by list_export) of the exportable information for a model instance"""
@@ -283,7 +283,7 @@ class IndexView(SpreadsheetExportMixin, WMABaseView):
             f, attr, value = lookup_field(field, item, self.model_admin)
             if not value:
                 value = getattr(attr, 'empty_value_display', self.model_admin.get_empty_value_display(field))
-            row_dict[field] = value
+            row_dict[field] = force_str(value)
 
         return row_dict
 
--- a/wagtail/contrib/modeladmin/views.py
+++ b/wagtail/contrib/modeladmin/views.py
@@ -283,7 +283,7 @@ class IndexView(SpreadsheetExportMixin, WMABaseView):
             f, attr, value = lookup_field(field, item, self.model_admin)
             if not value:
                 value = getattr(attr, 'empty_value_display', self.model_admin.get_empty_value_display(field))
-            row_dict[field] = force_str(value)
+            row_dict[field] = value
 
         return row_dict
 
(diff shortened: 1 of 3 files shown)


wagtail/contrib/modeladmin/views.py
```diff
@@ -283,7 +283,7 @@ def to_row_dict(self, item):
             f, attr, value = lookup_field(field, item, self.model_admin)
             if not value:
                 value = getattr(attr, 'empty_value_display', self.model_admin.get_empty_value_display(field))
-            row_dict[field] = value
+            row_dict[field] = force_str(value)
```
jacobtoppm, line 286: If you see https://github.com/wagtail/wagtail/blob/master/wagtail/admin/views/reports.py, `force_str` is currently chosen as the default in `get_preprocess_function` anyway, which allows the normal `SpreadsheetExportMixin` overrides (by value type and by field name) to be used, whereas adding it here would prevent those from taking place. I'd suggest removing it here.

Cheradenine:
Removed force_str - that makes sense. Stepped through the relevant code. Then I went down a rabbit hole with Excel and openpyxl where dates before 1900 are borked. They don't properly parse in Excel itself. So I added a comment, linked to the openpyxl issue. Goofy workaround to get tests to pass. Happy to discuss other options.

jacobtoppm:
> Removed force_str - that makes sense. Stepped through the relevant code. Then I went down a rabbit hole with Excel and openpyxl where dates before 1900 are borked. They don't properly parse in Excel itself. So I added a comment, linked to the openpyxl issue. Goofy workaround to get tests to pass. Happy to discuss other options.

Hmm, that was an interesting rabbit hole, thanks for catching that! The workaround functions for tests, but it seems that Excel won't properly display dates before 1900 anyway, so I think we need another option more generally eventually, and I'm not sure we want to be testing for this odd behaviour. For eventual solutions, one would be to replace

```
        (datetime.date, datetime.time): {FORMAT_XLSX: None},
```

in `SpreadsheetExportMixin.custom_value_preprocess` with a function that forces the date to a string if pre-1900. This isn't ideal, as if you had dates spanning the year 1900, some would be coerced to strings and some would remain in standard Excel date time format.

Another option would be to coerce all dates to strings. This is consistent, but means you don't have nice Excel dates automatically when exporting - especially relevant since most dates are likely to be post-1900 anyway!

A third would be to check all dates for pre-1900 values before outputting anything, and determining string vs Excel date format based on that.. This does result in an otherwise-unnecessary second loop through all data (as with things like modeladmin methods, we can't use a simple database query for the minimum), so isn't great for efficiency.

For now, until we make a further decision, I'd lean towards:
- Output in Excel datetime format as we do now (as the dates for builtin reports will all be post-1900, and for now we can encourage people to use `custom_field_preprocess` for custom reports)
- For tests: add an override to `custom_field_preprocess` on an `IndexView` to force the date to string, or to your date extracting method, for the purpose of testing the modeladmin methods for fields behaviour rather than this particular oddity
- We can then get this merged, and raise a separate issue for the 1900 date behaviour

Would you be happy to modify your tests slightly for now?

Cheradenine:
Moved the wonky date off by one code from the test, and as suggested we now use custom serialization for excel to iso format strings. This does seem cleaner - agree having that in the test was odd.

PR #6084, opened 2020-05-28
[Workflow] Audit log

zerolab:
This PR adds the ability to have an audit trail for actions on Wagtail objects. The current focus is pages, however the `LogEntry` model allows for any Django model.

Core model actions are logged at the model, rather than view level to account for management commands (e.g. publishing scheduled revisions), custom code and complex decision branches such as those in `PageRevision.publish()`. Where appropriate, the methods allow skipping logging.

Additionally, a new Page History "report" replaces the revisions list.

Last, but not least, this PR builds upon or touches on several RFCs:

- [RFC 24: Enhance PageRevision to include release notes](https://github.com/wagtail/rfcs/pull/24)
- [RFC 29: Immutable Revision Events](https://github.com/wagtail/rfcs/pull/29)
- [RFC 47: Proposal for supporting comments when submitting pages for moderation](https://github.com/wagtail/rfcs/pull/47)
- [RFC 45: Workflow](https://github.com/wagtail/rfcs/pull/24)

Diff with original, generic FK implementation: https://gist.github.com/zerolab/31605438a6ac72dcf64d24e320ea0247

* For front-end changes: Did you test on all of Wagtail’s supported browsers? **Please list the exact versions you tested**. ✅


docs/advanced_topics/audit_log.rst
```diff
@@ -0,0 +1,32 @@
+.. _audit_log:
+
+Audit log
+=========
+
+Wagtail provides a mechanism to log actions performed on its objects. Common activities like page creation, update, deletion,
+locking and unlocking, revision scheduling, privacy changes are automatically logged at the model level.
```
tomdyson, line 7: revision scheduling, privacy -> revision scheduling and privacy

docs/advanced_topics/audit_log.rst
```diff
@@ -0,0 +1,32 @@
...
+
+The Wagtail admin uses the action log entries to provide a site-wide and page specific history of changes. It uses a
+a registry of 'actions' that provide additional context for the logged action.
```
tomdyson, line 10: a a -> a

docs/advanced_topics/audit_log.rst
```diff
@@ -0,0 +1,32 @@
...
+The audit log-driven Page history replaces the revisions list page, but provide a filter for revision-specific entries.
+
+.. note:: The audit log does not replace revisions
+
+To provide additional logging for your site or package, invoke the :meth:`~LogEntryManger.log_action` manager method via ``LogEntry.objects.log_entry(object_instance, action)``
+and register a ``register_log_actions`` hook to describe your action (see :ref:`register_log_actions`).
+
+You can provide additional metadata by passing additional parameters:
+
+- `user`` - a user object.
```
tomdyson, line 21: `user`` -> ``user`` (double backtick before user - Github markdown escapes this comment)

zerolab: Thank you @tomdyson. Sharp-eyed as always!

docs/reference/pages/model_reference.rst
```diff
@@ -780,3 +780,102 @@ Database fields
...
+
+        The date/time when the entry was created.
+
+    .. attribute:: revision
+
+        (foreign key to :class:`PageRevision`)
+
+        Optional foreign key to the current page revision, when logging for :class:`Page` objects.
+
+    .. attribute:: created
```
kaedroho, line 838: We probably don't need these anymore. Thanks to micrologging and inheritance (that we could implement later).

Just to get it in writing, the think the idea of inheritance we had was to allow actions to be given parents when they are registered. So for example:

 - `wagtail.copy` inherits `wagtail.create`
 - `wagtail_localize.copy_for_translation` inherits `wagtail.copy`
 - `wagtail.revert` inherits `wagtail.save_draft`

Then with some Python trickery, we turn this into a tree structure and make all queries looking for `wagtail.create` logs actually search for `wagtail.create`, `wagtail.copy` and `wagtail.copy_for_translation`.

wagtail/admin/models.py
```diff
@@ -8,12 +8,12 @@
 # wagtail admin (namely, base_form_class and get_edit_handler). Importing this within
 # wagtail.admin.models ensures that this happens in advance of running wagtail.admin's
 # system checks.
-from wagtail.admin import edit_handlers  # NOQA
+from wagtail.admin import edit_handlers, log_action_registry  # NOQA
```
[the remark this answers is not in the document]
zerolab: this was actually an unused import, so removed

docs/reference/pages/model_reference.rst
zerolab: We do need `content_changed` and `deleted` for filtering purposes, and we discussed we'd leave all of them until we implement the "inheritance" idea.

wagtail/admin/__init__.py
```diff
@@ -1 +1,6 @@
+from .log_action_registry import LogActionRegistry
```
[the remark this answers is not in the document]
zerolab: I agree, but I can't think of a better way. imho, when we import models in `admin/log_action_registry.py` anything that needs to explode will do, so maybe we cross that bridge at that point?

wagtail/core/models.py
```diff
@@ -670,13 +671,47 @@ def save(self, clean=True, **kwargs):
                 self.url_path
             )
 
+        if log_action is not None:
+            # The default for log_action is False. i.e. don't log unless specifically instructed
+            # Page creation is a special case that we want logged by default, but allow skipping it
+            # explicitly by passing log_action=None
+            if is_new:
```
zerolab, line 678: > Don't we need to disable this for page copy?

@kaedroho, we can't do this without a context manager (which we agreed would be overkill). *sigh*

1. we want log page creation generally
2. it probably is a good idea to disable this for page copy, however the copy happens via `Page._copy()` which calls `_save_copy_instance()` which calls `to.add_child()` (from mp_tree) which then calls its own `process()`, which finally does `obj.save()`

I'm open for suggestions

wagtail/admin/__init__.py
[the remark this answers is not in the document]
zerolab: Sorted!

wagtail/core/models.py
```diff
@@ -3185,6 +3471,178 @@ def get_comment(self):
...
+
+class LogEntry(models.Model):
+    content_type = models.ForeignKey(
+        ContentType,
+        models.SET_NULL,
+        verbose_name=_('content type'),
+        blank=True, null=True,
+        related_name='+',
+    )
+    object_id = models.CharField(max_length=255, blank=False, db_index=True)
```
gasman, line 3651: I'm thinking this generic foreign key is probably more trouble than it's worth:

* There are lots of places (e.g. `LogEntriesView`) where we're filtering on `object_id` but don't have a corresponding filter on `content_type`, which need fixing if we're ever going to have log entries for non-page types
* It gets in the way of efficient querying - e.g. `LogEntriesView` has to retrieve the entire `explorable_pages()` dataset (which may contain many thousands of IDs) to convert to strings to pass back to the next query. With a non-generic foreign key, Django would (hopefully) be able to compile this as a subquery that the database can optimise.
* Many of the definitions here are page-specific - realistically, if we extended logging to non-page models, isn't it more likely that they'd use a separate model (possibly with a common abstract base model / mixin)?

zerolab: I have debated this for a bit. Decided to make it generic to accomodate non-integer PKs.
FWIW, `LogEntriesView` does `LogEntry.objects.get_pages()` which limits the qs to `Page`-derived content types

> It gets in the way of efficient querying
Great shout on the `explorable_pages()` dataset.

> Many of the definitions here are page-specific - realistically, if we extended logging to non-page models, isn't it more likely that they'd use a separate model (possibly with a common abstract base model / mixin)?

I see where you are going with this. In hindsight a common abstract model with non-page model specific implementations is probably more elegant. So `LogEntry` becomes `BaseLogEntry`, with a `PageLogEntry` models with a page FK with `on_delete=models.DO_NOTHING`.

Can work with that

zerolab: I guess the downside of moving to an abstract model / mixin approach is that pulling it all together in "site history" requires jumping through hoops. 

Do we tackle that when we introduce logging for non-Page model? e.g. by adding a separate filter that picks up the various instances (page, image, doc etc) and provide the appropriate querysets, defaulting to `PageLogEntry` objects?

zerolab: Went with abstract base class.

wagtail/core/models.py
zerolab: @kaedroho this is the last bit that I can't quite fix elegantly, but @gasman gave his blessing for the PR. So... file under "won't fix" ?

zerolab:
Note: this needs an upgrade consideration that one needs to run `./manage.py create_log_entries_from_revisions` to populate page history from existing revisions

wagtail/admin/views/page_privacy.py
```diff
@@ -27,10 +27,13 @@ def set_privacy(request, page_id):
...
                 # remove any existing restriction
                 if restriction:
-                    restriction.delete()
+                    restriction.delete(user=request.user)
             else:
                 restriction = form.save(commit=False)
                 restriction.page = page
+                # pass the current user as restriction attribute since the model form calls a plain instance.save()
+                # the attribute is removed in PageViewRestriction.save()
+                restriction.user = request.user
```
kaedroho, line 36: I'm pretty sure the only thing `commit=False` doesn't do is run `instance.save()` so could we just call `restriction.save()` here?

wagtail/core/management/commands/create_log_entries_from_revisions.py
```diff
@@ -0,0 +1,49 @@
...
+
+                if previous_revision is not None:
+                    comparison = get_comparison(revision.page, current_revision_as_page, previous_revision.as_page_object())
+                    has_content_changes = len(comparison) > 0
+
+                if is_new_page or has_content_changes or published:
+                    if is_new_page:
+                        action = 'wagtail.create'
+                    elif published:
+                        action = 'wagtail.publish'
```
kaedroho, line 35: I think we should treat publish as a separate event like it is in the view. So if the page was created and published, you have two log entries.

wagtail/core/management/commands/create_log_entries_from_revisions.py
```diff
@@ -0,0 +1,49 @@
...
+        previous_revision = None
+        for revision in PageRevision.objects.order_by('page_id', 'created_at').select_related('page').iterator():
+            is_new_page = revision.page_id != current_page_id
+            has_content_changes = False
+            current_page_id = revision.page_id
+
+            current_revision_as_page = revision.as_page_object()
+
+            if not PageLogEntry.objects.filter(revision=revision).exists():
+                published = revision.id == revision.page.live_revision_id
```
kaedroho, line 25: We did have some code that detected past published revisions by looking at what the `live_revision_id` was in the `content_json` field. Did we remove that?

wagtail/core/models.py
```diff
@@ -3185,6 +3538,201 @@ def get_comment(self):
...
+
+class PageLogEntryManager(BaseLogEntryManager):
+
+    def log_action(self, instance, action, **kwargs):
+        kwargs.update(page=instance)
+        return super().log_action(instance, action, **kwargs)
+
+
+class BaseLogEntry(models.Model):
+    content_type = models.ForeignKey(
```
kaedroho, line 3628: Is this field now redundant since we have type-specific `LogEntry` models?

wagtail/core/models.py
```diff
@@ -3185,6 +3538,201 @@ def get_comment(self):
...
+        models.SET_NULL,
+        verbose_name=_('content type'),
+        blank=True, null=True,
+        related_name='+',
+    )
+    label = models.TextField()
+
+    action = models.CharField(max_length=255, blank=True, db_index=True)
+    data_json = models.TextField(blank=True)
+    timestamp = models.DateTimeField("Timestamp (UTC)")
```
kaedroho, line 3639: We should wrap this with `gettext_lazy` if it's displayed anywhere

wagtail/core/models.py
```diff
@@ -3185,6 +3538,201 @@ def get_comment(self):
...
+        next_task = self.workflow_state.get_next_task()
+        next_task_data = None
+        if next_task:
+            next_task_data = {
+                'id': next_task.id,
+                'title': next_task.name
+            }
+        PageLogEntry.objects.log_action(
+            instance=page,
+            action='wagtail.workflow.{}'.format(action),
```
kaedroho, line 3553: Do we still support custom actions on custom task types?

wagtail/core/management/commands/create_log_entries_from_revisions.py
zerolab: We did - https://gist.github.com/zerolab/31605438a6ac72dcf64d24e320ea0247#file-audit-log-original-diff-L381 - it was setting the plublished flag to true at the end of the process, however we dropped the flag, so no need.

There's additional logic (around `if is_new_page or has_content_changes or published`) that deals with published revisions and it is a simpler process.

kaedroho: Ahh, yep. We are now adding an additional event for published revisions aren't we? Do we want to have a full history of publish events?

zerolab: sorted (and fixed the comparison too)

wagtail/core/models.py
zerolab: This is useful when the page is deleted and `object_verbose_name` uses it

wagtail/core/models.py
zerolab: yes, `TaskState.approve/reject()` call `TaskState.log_state_change_action()` with approve/reject as the action. Any custom task state should call `log_state_change_action`. One option would be to move this on `TaskState.on_action()`, however this messes up with the order of logged actions. and.. nothing prevents the custom `on_action` to skip the parent on_action entirely. Maybe this needs documenting?

wagtail/core/tests/test_audit_log.py
```diff
@@ -0,0 +1,274 @@
...
+        self.home_page.save()
+        revision.publish()
+
+        self.assertEqual(PageLogEntry.objects.count(), 4)
+        self.assertEqual(PageLogEntry.objects.filter(action='wagtail.publish').count(), 2)
+        self.assertEqual(PageLogEntry.objects.filter(action='wagtail.rename').count(), 1)
+
+    def test_page_unpublish(self):
+        self.home_page.unpublish()
+        self.assertEqual(PageLogEntry.objects.count(), 2)
```
kaedroho, line 116: Could we add an assertion for the other log entry as well?

kaedroho: If this other log entry is created by `.add_child` in the `setUp()`, maybe we should delete all log entries at the end of `setUp()` so only relevant entries exist?

wagtail/core/models.py
```diff
@@ -1830,8 +1976,45 @@ def is_latest_revision(self):
...
+        :param previous_revision: indicates a revision reversal. Should be set to the previous revision instance
+        """
         page = self.as_page_object()
+
+        def log_scheduling_action(revision, previous_revision=None, user=None, changed=changed):
+            if not previous_revision:
+                action = 'wagtail.schedule.publish'
+                revision_for_data = self
+            else:
+                action = 'wagtail.schedule.revert'
```
kaedroho, line 1999: Do we need a separate action for scheduling a revert? The revert has already been logged as `wagtail.revert`

wagtail/core/models.py
```diff
@@ -1830,8 +1976,45 @@ def is_latest_revision(self):
...
+            performs several actions: "publish", "revert" (and publish the reverted revision),
+            "schedule publishing with a live revision", "schedule revision reversal publishing, with a live revision",
+            "schedule publishing", "schedule revision reversal publishing"
+        :param previous_revision: indicates a revision reversal. Should be set to the previous revision instance
+        """
         page = self.as_page_object()
+
+        def log_scheduling_action(revision, previous_revision=None, user=None, changed=changed):
+            if not previous_revision:
+                action = 'wagtail.schedule.publish'
```
kaedroho, line 1996: When I first saw this, I assumed it meant "published using a wagtail schedule". Could we rename this to `wagtail.schedule`?

wagtail/core/models.py
```diff
@@ -1830,8 +1976,45 @@ def is_latest_revision(self):
...
+        :param log_action:
+            flag for the logging action. Pass False to skip logging. Cannot pass an action string as the method
+            performs several actions: "publish", "revert" (and publish the reverted revision),
+            "schedule publishing with a live revision", "schedule revision reversal publishing, with a live revision",
+            "schedule publishing", "schedule revision reversal publishing"
+        :param previous_revision: indicates a revision reversal. Should be set to the previous revision instance
+        """
         page = self.as_page_object()
+
+        def log_scheduling_action(revision, previous_revision=None, user=None, changed=changed):
```
kaedroho, line 1994: On scheduling, I think we might want to set the `user` attribute here: https://github.com/wagtail/wagtail/blob/master/wagtail/core/management/commands/publish_scheduled_pages.py#L108

To match the user in the revision?

Also, should we have separate action types for publishing from a schedule and a page expiring? (we display expiry differently in the UI to unpublish)

wagtail/core/models.py
zerolab: we do for publish, so it makes sense for this

wagtail/core/models.py
```diff
@@ -1379,6 +1476,22 @@ def copy(self, recursive=False, to=None, update_attrs=None, copy_revisions=True,
...
+                action=log_action,
+                user=user,
+                data={
+                    'page': {
+                        'id': page_copy.id,
+                        'title': page_copy.get_admin_display_title()
+                    },
+                    'source': {'id': parent.id, 'title': parent.get_admin_display_title()} if parent else None,
+                    'destination': {'id': to.id, 'title': to.get_admin_display_title()} if to else None,
+                    'keep_live': page_copy.live and keep_live
```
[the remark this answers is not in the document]
zerolab: We discussed this previously and decided against it. The action is not copy & publish, but copy (and maintain published status if set). We could vary the displayed message since we have the metadata.

I'm in two minds about it. @gasman break the tie?

wagtail/core/models.py
kaedroho: Oh, I didn't realise we did it for publishing too. What's the advantage of having a separate revert+publish event? We don't have a separate event for edit+publish do we?

wagtail/core/models.py
zerolab: renamed to `wagtail.publish.schedule` as per Slack convo

wagtail/core/models.py
zerolab: dropped `wagtail.schedule.revert` in favour of a single `wagtail.publish.schedule`

wagtail/core/models.py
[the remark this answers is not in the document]
gasman: I think _not_ having a separate log entry makes sense, since this is a log of 'actions performed', not 'state changes that happened as a result of actions'. If we were to start logging implied sub-actions in their own right, then logically you could make a similar case that copying a page also includes a 'create page' action, or that deleting a published page includes an 'unpublish' action, and it seems to me that going down that route would harm the usefulness of the logs.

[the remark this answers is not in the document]
kaedroho: > I think not having a separate log entry makes sense, since this is a log of 'actions performed', not 'state changes that happened as a result of actions'.

When Dan and I discussed this, we came up with "micro-logging" where we make multiple logs for each action instead of trying to encode every possible combination of actions in Wagtail. This included making some actions more specific versions of others:

The Hierachy I'm thinking of is as follows:

```
wagtail.create
    wagtail.copy
        wagtail.copy_for_translation
    wagtail_content_import.import
    wagtail_transfer.import
wagtail.edit
    wagtail.revert
    wagtail_transfer.update
wagtail.publish
    wagtail.schedule.publish
wagtail.unpublish
    wagtail.schedule.expire
wagtail.delete
```

This heirachy would be encoded in the registry so that queries can be easily made to find all creations no matter what action was actually performed to create them.

If you copy a page and keep it live, I think it should emit both a `wagtail.copy` and a `wagtail.publish` log. This is to match how creating + publish is logged which emits a `wagtail.create` and `wagtail.publish` log

wagtail/core/models.py
zerolab: Partially solved in a79b2fb

wagtail/core/models.py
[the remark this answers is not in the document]
zerolab: As discussed this morning, added the publish log on copy. I've also updated the documentation to list the current actions and when they are logged + the rule of thumb for logging

gasman:
Doh, just noticed something after merging...

https://github.com/wagtail/wagtail/blob/feature/workflow/wagtail/admin/views/reports.py#L386

We shouldn't be using `Site.find_for_request(self.request)` within the admin - we don't assume that the hostname a user is using to access the admin corresponds to a site.

@zerolab, I half remember us discussing whether non-superusers should be able to see logs for deleted pages, and while I can't remember the rationale, I think we settled on users _with permission over the whole tree_ being able to see them. In which case, we can change this to a check on the tree root page (`Page.get_first_root_node()`) instead. Does that sound right?

zerolab:
@gasman having checked my notes, your summary of the decision is spot on and `Page.get_first_root_node()` sounds perfect. Will submit a new PR

PR #6088, opened 2020-05-29
Revise RoutablePageMixin default index route to call parent's serve

### LINKED ISSUE #6062: AbstractEmailForm and RoutablePageMixin did not work together
### Issue Summary

Trying to mix `AbstractEmailForm` and `RoutablePageMixin`  to provide a form and a custom route does not work - only one of these two functionalities does work.

### Steps to Reproduce

Add a page model with a _Form builder_-form and a (secondary) custom route

Depending on the order of inheritance:
``` python
class MyPage(AbstractEmailForm, RoutablePageMixin, Page):
    @route(r'^ics/$')
    def get_ics_for_event(self, request):
        return ical_response(
            events=self.get_events,
            filename='{}.ics'.format(self.slug)
        )
```
vs. 

``` python
class MyPage(RoutablePageMixin, AbstractEmailForm, Page):
    @route(r'^ics/$')
    def get_ics_for_event(self, request):
        return ical_response(
            events=self.get_events,
            filename='{}.ics'.format(self.slug)
        )
```
I get a form **or** my custom `@route`, but not both. 

I did not overwrite the default route and assumed to get both the form and the route working:

> By default a route for r'^$' exists, which serves the content exactly like a regular Page would. It can be overridden by using @route(r'^$') on any other method of the inheriting class.

_(from https://docs.wagtail.io/en/v2.9/reference/contrib/routablepage.html)_

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: 
sorry, not yet

### Technical details

* Python version: Python 3.8.3rc1
* Django version: Django               3.0.6
* Wagtail version: wagtail              2.9

lb-:
Currently, `RoutablePageMixin` will not use any existing parent's serve method, which might make sense as it has its own method but it can cause confusion and appears to not align with how the docs explain this mixin. This means `RoutablePageMixin` is not compatible with `AbstractEmailForm` out of the box, nor is it compatible with pages that extend classes that also have their own serve method.

Reading the item from the docs 'exactly like any regular Page would' could be seen as it should honour existing serve methods if present, on the index route created.

> By default a route for r'^$' exists, which serves the content exactly like a regular Page would. It can be overridden by using @route(r'^$') on any other method of the inheriting class.
> https://docs.wagtail.io/en/stable/reference/contrib/routablepage.html

This PR adds a bit of indirection, and as per [Matt's comments](https://github.com/wagtail/wagtail/issues/6062#issuecomment-634640233) it may mix up levels of abstraction. However, i think it might be worth considering.

@gasman  - my comments on the original issue were probably not very nuanced, I do not think we need to support `serve` methods on page models that use `RoutablePageMixin` but instead support serve models on classes that are extended by the model that is in turn using `RoutablePageMixin`. I do not know a more articulate way to explain it, but the added tests will show the scenario I am trying to explain.

**Details**
- makes RoutablePageMixin compatible with AbstractEmailForm & AbstractForm
- allows RoutablePageMixin to be used with any custom Page models that have their own serve method by default
- resolves #6062 
* Do the tests still pass? ✅ 


gasman:
With the current version of the code, there's an unambiguous chain of command - the `route` method finds a view method that matches the URL, the `serve` method passes control to the view method, and the view method generates the response. My concern is that this is no longer true now that `serve` is serving a dual function - it works (for this particular inheritance chain) because there's a path through `serve` that avoids going round in a loop, but that seems to be more by accident than by design. (It looks like the 'call super().serve if view is None' case is actually dead code at the moment - that was added to prevent an error if an index route was not defined (#2585), but since then we added a default index route to RoutablePageMixin (#3344).)

I think this is going to harm future maintainability of this code - any changes to `serve` will need to make sure that not only the primary functionality (i.e. delegating control to the relevant view method) still works, but also that the 'hidden path' through it is kept open. It'll no longer be possible to reason about the correctness of an individual method of RoutablePageMixin based on its inputs and outputs, only by treating the whole class as a complex interconnected bundle.

lb-:
@gasman - I agree, this would make it harder to reason about in the future, I will close this PR.

PR #6091, opened 2020-05-30
Tweak Wagtail logo hover animation in docs

acarasimon96:
I think our little pet Wagtail logo bobs down too slow when you hover it, so I made it behave exactly as in the Wagtail admin (without the "playful" mode of course).

Also, isn't the "pet the wagtail" action on the Wagtail admin logo an Easter egg?

### DIFF
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -34,8 +34,7 @@ div.wy-side-nav-search {
 }
 
 .logo:hover {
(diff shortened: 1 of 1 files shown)


lb-:
@acarasimon96 thanks for the tweak, merged in. 
I think its nice to have a bit of an animation on the docs also, also serves as a visual hover state of an interactive element :)

PR #6099, opened 2020-06-02
Strip away HTML tags from RichText searchable content

### LINKED ISSUE #6098: RichTextField / RichTextBlock should strip out HTML tags from search-indexable content
### Issue Summary

Currently, when indexing RichTextFields, or RichTextBlocks within StreamFields, the raw content including HTML tags is passed to the index. Since HTML tag names / attributes are not meaningful searchable content*, it would be better to provide a `get_searchable_content` method on these that applies Django's `striptags` filter.

More importantly, if you customise the Elasticsearch backend (or run raw Elasticsearch queries) to take advantage of [Elasticsearch's highlighting support](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/search-request-highlighting.html) (see #5340), you currently end up with either stray HTML tags or spurious formatting in the results, depending on whether you passed `'encoder': 'html'` in the query.

(* This isn't totally true - for example, you could make a good case for indexing the alt text on images - so an ideal solution would probably involve being able to specify custom rules at the richtext feature level to generate a plain text representation. However, on balance, I think stripping tags out is better than keeping them in.)

### Steps to Reproduce

Search for 'h2' on https://www.rca.ac.uk/ . Observe that the results include many pages that contain an `<h2>` element but not the text "h2"...

acarasimon96:
Fixes #6098

So far I only stripped HTML tags in searchable contents of `RichTextBlock`s per @gasman's proposal in the issue. Stripping out tags from `RichTextField`s are ~~currently a work in progress, so I'll remove the draft status from this PR once that's done.~~

**Edit:** `RichTextField.get_searchable_content` has been added (and was easier to implement than I thought).

---

### DIFF
--- a/wagtail/core/fields.py
+++ b/wagtail/core/fields.py
@@ -1,7 +1,10 @@
 import json
+from html import unescape
 
 from django.core.serializers.json import DjangoJSONEncoder
 from django.db import models
+from django.utils.encoding import force_str
+from django.utils.html import strip_tags
 
 from wagtail.core.blocks import Block, BlockField, StreamBlock, StreamValue
 
@@ -18,6 +21,11 @@ class RichTextField(models.TextField):
         defaults = {'widget': get_rich_text_editor_widget(self.editor, features=self.features)}
         defaults.update(kwargs)
         return super().formfield(**defaults)
+
+    def get_searchable_content(self, value):
+        # Strip HTML tags to prevent search backend to index them
+        source = force_str(value)
+        return [unescape(strip_tags(source))]
 
 
 # https://github.com/django/django/blob/64200c14e0072ba0ffef86da46b2ea82fd1e019a/django/db/models/fields/subclassing.py#L31-L44
--- a/wagtail/core/tests/test_blocks.py
+++ b/wagtail/core/tests/test_blocks.py
@@ -549,6 +549,20 @@ class TestRichTextBlock(TestCase):
         with self.assertRaises(ValidationError):
             block.clean(RichText('<p>bar</p>'))
 
+    def test_get_searchable_content(self):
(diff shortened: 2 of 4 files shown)


wagtail/core/fields.py
```diff
@@ -19,6 +22,11 @@ def formfield(self, **kwargs):
         defaults.update(kwargs)
         return super().formfield(**defaults)
 
+    def get_searchable_content(self, value):
+        # Strip HTML tags to prevent search backend to index them
```
tomdyson, line 26: search backend to index them -> search backend from indexing them

wagtail/core/tests/test_blocks.py
```diff
@@ -549,6 +549,20 @@ def validate_contains_foo(value):
         with self.assertRaises(ValidationError):
             block.clean(RichText('<p>bar</p>'))
 
+    def test_get_searchable_content(self):
+        block = blocks.RichTextBlock()
+        value = RichText(
+            '<p>Merry <a linktype="page" id="4">Christmas</a>!</p>\n'
```
gasman, line 555: Just a small follow-up suggestion - perhaps we should include an `&amp;` in this test, to check that the `unescape` is being applied correctly...
```suggestion
            '<p>Merry <a linktype="page" id="4">Christmas</a> &amp; a happy new year!</p>\n'
```

lb-:
@acarasimon96 thanks for this, an awesome improvement.

I made those minor changes suggested by @tomdyson & @gasman  and merged in via 48511a7457c5e25b9cdda48f593664e2a82e8851

acarasimon96:
Wow, I didn't expect this PR to get merged overnight! Thank you everyone who got involved! You guys are awesome! :heart:

PR #6101, opened 2020-06-02
Update npm packages to address reported vulnerabilities

thibaudcolas:
This updates a lot of our transitive dependencies, and direct development dependencies, in order to address security vulnerabilities reported by `npm audit`. None of the changes here should have any incidence on the compiled JS & CSS shipped to users – this is all tooling only. As far as I could see all of the vulnerabilities also were with our development tooling, rather than code shipped to users of Wagtail.

Before:

```
$ npm install
found 1429 vulnerabilities (457 low, 26 moderate, 946 high)
```

Then I did:

```
$ npm audit fix
fixed 1326 of 1429 vulnerabilities in 2220 scanned packages
  6 vulnerabilities required manual review and could not be updated
  7 package updates for 97 vulns involved breaking changes
```

And then proceeded to upgrade a few packages manually,

- Latest `stylelint` – currently this causes a dependency warning on `npm install`, will be gone once https://github.com/wagtail/stylelint-config-wagtail/pull/1 is merged.
- Latest `gulp-size` and `gulp-cssnano` – no changes worth reporting
- Latest `gulp-autoprefixer` – this is why I had to move the autoprefixer configuration from the gulp tasks to the package.json.
- Latest `react-streamfield` – this fixes an issue with that package’s dependencies. Again no changes to how this is actually used in Wagtail.

---

Tested with the UI regression test suite in [wagtail-tooling](https://github.com/thibaudcolas/wagtail-tooling).

---

* [x] Do the tests still pass? (https://docs.wagtail.io/en/latest/contributing/developing.html#testing)
* [x] Does the code comply with the style guide? (Run `make lint` from the Wagtail root)
* [x] For Python changes: Have you added tests to cover the new/fixed behaviour?
* [x] For front-end changes: Did you test on all of Wagtail’s supported browsers?  Chrome only.
* [x] For new features: Has the documentation been updated accordingly?

PR #6105, opened 2020-06-03
Webpack & Node tooling upgrades

thibaudcolas:
This PR was originally about updating our development tools to reduce the number of vulnerabilities we may be exposed to, as a follow-up to #6101. Now it also feels like an important next step from #6472.

Here are the changes:

- Upgrading to Webpack 4 to fix security vulnerabilities, and so we can eventually use `ts-loader` with TypeScript v4 for JS & TS compilation.
- Switching from `gulp-autoprefixer` and `gulp-cssnano` to `gulp-postcss`, and the standalone `cssnano` and `autoprefixer` pacakges. This means we can use the latest versions of those packages, and it’s also taking us closer to moving stylesheets compilation to Webpack eventually (working on this in a separate branch).
- A few other package upgrades to use versions with no known vulnerabilities.

There is now only 1 reported vulnerability in a transitive dependency of the Draftail editor, for which I’m certain we don’t use any of the vulnerable code.

---

* [x] For front-end changes: Did you test on all of Wagtail’s supported browsers?  Chrome and IE11.


client/webpack/dev.config.js
```diff
@@ -1,23 +0,0 @@
...
-
-// add poll-options for in vagrant development
-// See http://andrewhfarmer.com/webpack-watch-in-vagrant-docker/
-config.watchOptions = {
-  poll: 1000,
-  aggregateTimeout: 300,
-};
-
-// Set process.env.NODE_ENV to development to enable JS development aids.
-config.plugins.push(new webpack.DefinePlugin({
```
kaedroho, line 17: What about these plugins?

client/webpack/prod.config.js
```diff
@@ -1,30 +0,0 @@
-var webpack = require('webpack');
-var base = require('./base.config');
-var config = base('production');
-
-
-// production overrides go here
-config.plugins.push(new webpack.DefinePlugin({
```
kaedroho, line 7: What about these plugins?

kaedroho (commented):
Changes look good! Wondering what we're doing about the plugins that were removed in the second commit?

I've just done some squashing of the commits in: https://github.com/kaedroho/wagtail/commits/chore/major-npm-updates

client/webpack/dev.config.js
thibaudcolas: This is automatically defined in Webpack 4 based on the `mode`.

kaedroho (approved):

PR #6110, opened 2020-06-04
Make pages/find by html_path aware of redirects

adonig:
This pull requests makes `/api/v2/pages/find/?html_path=/foo` aware of any (page) redirects.

### DIFF
--- a/wagtail/api/v2/views.py
+++ b/wagtail/api/v2/views.py
@@ -456,9 +456,9 @@ class PagesAPIViewSet(BaseAPIViewSet):
         if 'html_path' in request.GET and site is not None:
             path = request.GET['html_path']
 
-            # Check whether there is a matching redirect and set path accordingly.
+            # Check whether there is a matching page redirect and set path accordingly.
             try:
-                redirect = Redirect.objects.get(old_path=path)
+                redirect = Redirect.objects.exclude(redirect_page__isnull=True).get(old_path=path)
                 path = urlparse(redirect.link).path
             except Redirect.DoesNotExist:
                 pass
--- a/wagtail/api/v2/views.py
+++ b/wagtail/api/v2/views.py
@@ -1,4 +1,5 @@
 from collections import OrderedDict
+from urllib.parse import urlparse
 
 from django.conf.urls import url
 from django.core.exceptions import FieldDoesNotExist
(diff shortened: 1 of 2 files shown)


wagtail/api/v2/views.py
```diff
@@ -453,6 +455,14 @@ def find_object(self, queryset, request):
         site = Site.find_for_request(request)
         if 'html_path' in request.GET and site is not None:
             path = request.GET['html_path']
+
+            # Check whether there is a matching page redirect and set path accordingly.
+            try:
```
kaedroho, line 460: This check should only happen if the page routing has failed. So if there is both a page and a redirect for the same URL, we get the page.

adonig: Hi Karl! Thank you for looking into this. I have made some changes to reflect your proposals. I placed a TODO where I thought would be a good place to add the header, because I did not know how to set it.

jaap3:
@gasman Would a standalone endpoint be acceptable? Something like this:

```python
class RedirectSerializer(serializers.ModelSerializer):
    class Meta:
        model = Redirect
        fields = ('old_path', 'is_permanent', 'link')


class RedirectAPIViewSet(viewsets.ReadOnlyModelViewSet):
    serializer_class = RedirectSerializer

    def get_object(self):
        request = self.request
        redirect = None

        if 'html_path' in request.GET:
            redirect = get_redirect(request, Redirect.normalise_path(request.GET['html_path']))

        if redirect is None:
            raise Http404()
        return redirect

    @classmethod
    def get_urlpatterns(cls):
        """
        This returns a list of URL patterns for the endpoint
        """
        return [
            url(r'^$', cls.as_view({'get': 'retrieve'})),
        ]

api_router = WagtailAPIRouter('wagtailapi')
api_router.register_endpoint('redirects', RedirectAPIViewSet)
```

gasman:
@jaap3 That would be fine by me - I don't know whether it solves the problem that the original submitter intended to solve, though.

jaap3:
I think it would, in a slightly more convoluted way. Instead of returning the `Page` instance the redirects point to it would be up to the consumer to issue a redirect and do a lookup from there. (Come to think of it, I think the initial solution fails to address redirects to external urls).

I'll see if I can whip up a PR with my suggestion after I'm done implementing it in our project.

lb-:
Just saw that https://github.com/wagtail/wagtail/pull/8842 replaces this PR and best to only have one open. Same goal but implemented without the mixing of contrib requirements.

lb-:
Thank you @adonig 

Took a bit but others have picked this up and prepared some follow up PRs and it just recently got merged in. I have added you to the contributors list though.

See 996abeae8eaeb96677a0e6f07e6addf25ca8f698 / https://github.com/wagtail/wagtail/pull/11510

PR #6115, opened 2020-06-07
Remove visible focus states from .visiblyhidden util

### LINKED ISSUE #5903: IE11 checkboxes aren't clickable
### Issue Summary

When trying to check a checkbox in IE11 it doesn't check, and displays the hidden help text around it instead. 

### Steps to Reproduce

1. Start a new project, or open an existing project. 
2. Run `python manage.py runserver 0.0.0.0:8000` so your site is accessible to your browser
3. Open a page with checkboxes in BrowserStack or LambdaTest. This is the page I've been using: http://localhost:8000/admin/groups/2/ (but any Group Edit view will work)

I think this is a bug somewhere because checking a checkbox in any browser should be supported as there isn't too much going on behind the scenes here. 

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: **Yes**, used a fresh install with Wagtail 2.8, Python 3.7.2, and Django 2.2.11 and got the same issue. 

### Technical details

* Python version: 3.6.9
* Django version: 2.2.9
* Wagtail version: 2.8
* Browser version: www.whatsmybrowser.org/b/58AG7

Visual of the issue
ie-checkbox-issue

lb-:
- visuallyhidden should keep elements hidden even when focused
- Fixes #5903 (IE 11 will trigger :focus on the label when checkbox is in focus & click event would be lost)
- see: https://github.com/wagtail/wagtail/issues/5903#issuecomment-639380475
- note: I have a separate PR coming that will resolve the issue with the usage of .visuallyhidden on snippets index (see comment above)

## Checklist

* Do the tests still pass? ✅ 
  *  ✅ Mac Firefox 76
  *  ✅ Mac Chrome 83
  *  ✅ Mac Safari 13.1
  *  ✅ Windows IE 11
  *  ✅ Windows Edge 17

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -74,6 +74,7 @@ Changelog
  * Fix: Make "Site" chooser in site settings translateable (Andreas Bernacca)
  * Fix: Add missing dropdown icons to image upload, document upload, and site settings screens (Andreas Bernacca)
  * Fix: Prevent snippets’ bulk delete button from being present for screen reader users when it’s absent for sighted users (LB (Ben Johnston))
+ * Fix: Fix group permission checkboxes not being clickable in IE11 (LB (Ben Johnston))
 
 
 2.9.3 (20.07.2020)
--- a/client/scss/tools/_mixins.general.scss
+++ b/client/scss/tools/_mixins.general.scss
@@ -67,16 +67,6 @@
     padding: 0;
     position: absolute;
     width: 1px;
-
-    &:active,
-    &:focus {
-        clip: auto;
-        height: auto;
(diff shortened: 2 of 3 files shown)


lb-:
Also, when this is resolved it should unblock #5836 FYI @KalobTaulien

lb-:
Accidentally closed, but just reopened.

lb-:
Thanks @thibaudcolas - yeah as it's a util I was not sure whether it should be removed or kept

thibaudcolas:
👍  things work as expected without it, and the IE11 issue is fixed, so all good. Thanks again!

PR #6116, opened 2020-06-07
Hide inactive delete button with .u-hidden instead of .visuallyhidden

lb-:
- .visuallyhidden meant that this button was visible to assistive technologies when the button would not work
- instead of hiding with that util, use .u-hidden (essentially display: none)
- relates to issue where the focus/active states in the visuallyhidden mixin were breaking some items in IE 11 - see #5903
* To test - confirm that you can select one or all of the items in a snippet type list and the delete button should show/hide once one or more items are selected.

## Checklis
* Do the tests still pass? ✔️ 
* Does the code comply with the style guide? ✔️ 
    ✅ Mac Firefox 76
    ✅ Mac Chrome 83
    ✅ Mac Safari 13.1
    ✅ Windows IE 11
    ✅ Windows Edge 17


lb-:
Accidentally deleted the branch, reopened now.

lb-:
thanks @thibaudcolas

PR #6117, opened 2020-06-07
Fix gulp watch task

lb-:
- revise watch & task syntax for Gulp v4 requirements (regression from upgrading to v4)
- to validate; run `npm run start` and make a change to a scss or js file - expectation is that the relevant build task should run and refreshing the browser should reflect that change

### DIFF
--- a/gulpfile.js/tasks/watch.js
+++ b/gulpfile.js/tasks/watch.js
@@ -2,16 +2,22 @@ var gulp = require('gulp');
 var path = require('path');
 var config = require('../config');
 
+
+const paths = config.apps.reduce((_, app) => ({
+    'styles:sass': [...(_['styles:sass'] || []), path.join('./client/**/*.scss'), path.join(app.sourceFiles, '*/scss/**')],
+    'styles:css': [...(_['styles:css'] || []), path.join(app.sourceFiles, '*/css/**')],
+    'scripts': [...(_['scripts'] || []), path.join(app.sourceFiles, '*/js/**')],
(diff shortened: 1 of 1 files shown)


zerolab:
Kudos @lb- !

I did scratch my head at the en of last week when I needed to do some style changes and watch was not reacting.

Tried this locally and watch is back working again. It does, interestingly, start 9 threads of `styles:sass` when saving a single scss file. I wonder if that is by design or there's a tweak that can be made

lb-:
Hmm. Good point. Maybe the series for each should be parallel, from what I read the old syntax would do series. 

Not sure if this used to work the same with multiple threads of scams builds. 

I also noticed it seems to take a while, 30s on my VM. 

I have not had much experience with full config so not sure what else to try. 

@zerolab

lb-:
@zerolab can you give it another try, I have consolidated all the watchers that get generated, hopefully this will be better.

zerolab:
🎉 that does it!

```bash
$ npm run watch
...
[09:23:34] Finished 'build' after 3.61 s

// modify and save an scss file
[09:23:34] Starting '<anonymous>'...
[09:23:50] Starting 'styles:sass'...
[09:23:52] Wagtail CSS all files 143.84 kB
[09:23:52] Finished 'styles:sass' after 1.4 s

// undo change, save
[09:24:00] Starting 'styles:sass'...
[09:24:01] Wagtail CSS all files 143.84 kB
[09:24:01] Finished 'styles:sass' after 1.09
```

zerolab:
Merged. Thank you @lb-

lb-:
Awesome thanks @zerolab

PR #6118, opened 2020-06-07
I would like only add my own editor without to redefine a default one.

gassan:

Yes
yes
yes


gasman:
Merged in a4e05f453865fc1617e5880968503d658b4a2aaa (with documentation updated) - thanks @gassan!

PR #6119, opened 2020-06-07
Add Azure CDN and Front Door front-end cache invalidation support

tm-kn:
This PR adds support for invalidating cache on two platforms:

* Azure CDN
* Azure Front Door

This is based on [wagtail-azure-cdn](https://github.com/torchbox/wagtail-azure-cdn) with multi site support taken out as it would only make this PR more complex.

I am not sure if this is a rich approach to have a massive `backends.py` file. Would it help to first create a PR that splits individual backends into separate files?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,6 +50,8 @@ testing_extras = [
     'freezegun>=0.3.8',
     'openpyxl>=2.6.4',
     'Unidecode>=0.04.14,<2.0',
+    'azure-mgmt-cdn>=5.1,<6.0',
+    'azure-mgmt-frontdoor>=0.3,<0.4',
 
     # For coverage and PEP8 linting
     'coverage>=3.7.0',
--- a/wagtail/contrib/frontend_cache/backends.py
+++ b/wagtail/contrib/frontend_cache/backends.py
@@ -3,7 +3,7 @@ import uuid
 
 from collections import defaultdict
 from urllib.error import HTTPError, URLError
-from urllib.parse import urlparse, urlunparse
+from urllib.parse import urlparse, urlsplit, urlunparse, urlunsplit
 from urllib.request import Request, urlopen
 
 import requests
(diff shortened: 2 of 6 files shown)


setup.py
```diff
@@ -49,6 +49,8 @@
     'boto3>=1.4,<1.5',
     'freezegun>=0.3.8',
     'openpyxl>=2.6.4',
+    'azure-mgmt-cdn',
```
kaedroho, line 52: Could you version checks here?

wagtail/contrib/frontend_cache/backends.py
```diff
@@ -199,3 +199,169 @@ def _create_invalidation(self, distribution_id, paths):
                     e.response['Error']['Code'],
                     e.response['Error']['Message']
                 )
+
+
```
kaedroho, line 203: We've tightened up the flake8 config recently so it won't allow three spaces here anymore, please could you rebase this on the latest master?

kaedroho (approved):
Nice work! Just a couple of minor things to do before merge.

> I am not sure if this is a rich approach to have a massive backends.py file. Would it help to first create a PR that splits individual backends into separate files?

Happy for it to be a big file for now, we can split it up after merge

setup.py
zerolab: I would very much like for us to move for optional dependencies where possible.
i.e. if you need the Azure CDN backend you'd do `pip install wagtail[cdn-azure]` or similar?

This is probably a large discussion to have, so not a blocker on this per se

kaedroho:
Rebased and made amendments

setup.py
kaedroho: Yep, I agree. We probably want to do something similar for cloudflare, elasticsearch, boto3, etc. Shall we tackle all these in one go in a separate PR?

lb-:
Flagging this for discussion at the next core team meeting.

lb-:
Good to go, action for LB to rebase/fix conflicts and Merge in.

tm-kn:
Just to let you know, I made changes to use the newer Azure SDK - torchbox/wagtail-azure-cdn#4

setup.py
lb-: @kaedroho  / @zerolab  I have split this out to a new issue, please can either of you update with a bit more context https://github.com/wagtail/wagtail/issues/7753

lb-: Note: the original issue raised has been resolved

lb-:
@tm-kn thanks for this, it has been merged in.
As for https://github.com/torchbox/wagtail-azure-cdn/pull/4 -> I thought it was best to add additional versions support in a new PR, it would be awesome if you could raise a new one when you get the chance.

Thanks @kaedroho  for reviewing.

Issue #6066, opened 2020-05-23
Users should be able to configure API renderer_classes


narun4sk:
### Issue Summary

Users should be able to configure API `renderer_classes`, currently these are hardcoded to `[JSONRenderer, BrowsableAPIRenderer]`: https://github.com/wagtail/wagtail/blob/1958bba2476c32069acafd6837514749a4d3ceea/wagtail/api/v2/views.py#L26

### Steps to Reproduce

Configure pages endpoint as usual, but do not include `rest_framework` in the `INSTALLED_APPS`. Try to open configured endpoint in the browser, the app will throw an exception:

```
Internal Server Error: /rest/p/
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 145, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 143, in _get_response
    response = response.render()
  File "/usr/local/lib/python3.8/site-packages/django/template/response.py", line 105, in render
    self.content = self.rendered_content
  File "/usr/local/lib/python3.8/site-packages/rest_framework/response.py", line 70, in rendered_content
    ret = renderer.render(self.data, accepted_media_type, context)
  File "/usr/local/lib/python3.8/site-packages/rest_framework/renderers.py", line 723, in render
    template = loader.get_template(self.template)
  File "/usr/local/lib/python3.8/site-packages/django/template/loader.py", line 19, in get_template
    raise TemplateDoesNotExist(template_name, chain=chain)
django.template.exceptions.TemplateDoesNotExist: rest_framework/api.html
```

### Technical details

- Python Version: 3.8.2
- Django Version: 3.0
- Wagtail Version: 2.9

---

Currently I'm using this nasty trick:

```
from django.conf import settings
from wagtail.api.v2.views import PagesAPIViewSet
from wagtail.api.v2.router import WagtailAPIRouter

if not settings.DEBUG:
    from rest_framework.renderers import JSONRenderer

    class ProdPagesAPIViewSet(PagesAPIViewSet):
        renderer_classes = [JSONRenderer]

    PagesAPIViewSet = ProdPagesAPIViewSet

# Create the router. "wagtailapi" is the URL namespace
api_router = WagtailAPIRouter("wagtailapi")

[...]

PR #6120, opened 2020-06-07
Make wagtail api renderer_classes configurable

marteinn:
This PR originates from https://github.com/wagtail/wagtail/issues/6066  and acts as a proposal on how we can make the `renderer_classes` configurable from the settings in a similar way as DRF works today.

This is my setting proposal;

```python
WAGTAILAPI_DEFAULT_RENDERER_CLASSES = [
    'rest_framework.renderers.JSONRenderer',
    'rest_framework.renderers.BrowsableAPIRenderer',
]
```

Some considerations I’ve had while doing this:

**Using the built in settings in DRF. (REST_FRAMEWORK.DEFAULT_RENDERER_CLASSES)**
This works by just removing the renderer_classes in the View definition, it works well but will most likely introduce bugs for users using custom `renderer_classes` that will start to affect the wagtail api.

**Using the setting design proposed in the originating issue. (WAGTAIL_API.DEFAULT_RENDERER_CLASSES)**
I’ve decided against using a multi key dictionary for definition since there are already flat level  `WAGTAILAPI_*` settings in the api app.

### DIFF
--- a/wagtail/api/v2/tests/test_renderer_classes.py
+++ b/wagtail/api/v2/tests/test_renderer_classes.py
@@ -1,0 +1,22 @@
+from django.test import TestCase
+from django.test.utils import override_settings
+from rest_framework.renderers import BrowsableAPIRenderer, JSONRenderer
+
+from wagtail.api.v2.utils import get_default_renderer_classes
+
+
+class TestPageListing(TestCase):
+    @override_settings(WAGTAILAPI_DEFAULT_RENDERER_CLASSES=[
+        'rest_framework.renderers.JSONRenderer',
+    ])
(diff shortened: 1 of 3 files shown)


lb-:
I think it's best to close this issue now, it's been a while and the original issue referenced has since been closed as done.

The solution was different to what's proposed here, instead of a setting there is better documentation for the existing approach that worked as is. See https://github.com/wagtail/wagtail/pull/9902

Thanks for raising this @marteinn and if you feel that it's worth a fresh discussion, can I ask that you raise a new issue with an overview of the proposal.  Alternatively, a discussion about how to better make the API endpoints configurable may be good.

PR #6122, opened 2020-06-07
Allow registering ModelAdmin with custom menu hook

tm-kn:
* This makes ModelAdmin more extensible to allow customising Wagtail hooks registration behaviour rather than having one big method. This will make overriding specific Wagtail hooks much easier.
* Add new `menu_hook` setting to `ModelAdmin`. It's not added to `ModelAdminGroup` as Wagtail admin won't support a third column, but maybe we should allow it as it is a more of a super user feature.

Related issue - wagtail/wagtail-modeladmin#18

### DIFF
--- a/docs/reference/contrib/modeladmin/menu_item.rst
+++ b/docs/reference/contrib/modeladmin/menu_item.rst
@@ -69,7 +69,19 @@ If you'd like the menu item for your model to appear in Wagtail's 'Settings'
 sub-menu instead of at the top level, add ``add_to_settings_menu = True`` to
 your ``ModelAdmin`` class.
 
-This will only work for indivdual ``ModelAdmin`` classes registered with their
+This will only work for individual ``ModelAdmin`` classes registered with their
 own ``modeladmin_register`` call. It won't work for members of a
 ``ModelAdminGroup``.
 
+------------------------
+``ModelAdmin.menu_hook``
+------------------------
+
+**Expected value**: A hook name for registering the new menu item.
+
+If you'd like menu item to be registered in a custom ``Menu``, set
+``menu_hook`` to the value from Menu's ``register_hook_name``.
+
+This will only work for individual ``ModelAdmin`` classes registered with their
+own ``modeladmin_register`` call. It won't work for members of a
+``ModelAdminGroup``.
--- a/wagtail/contrib/modeladmin/options.py
+++ b/wagtail/contrib/modeladmin/options.py
@@ -28,32 +28,40 @@ class WagtailRegisterable:
     exclude_from_explorer = False
 
     def register_with_wagtail(self):
+        self.register_permissions()
+        self.register_admin_urls()
+        self.register_admin_menu_item()
+        self.register_explorer_page_queryset()
+        self.register_checks()
 
-        @hooks.register('register_permissions')
-        def register_permissions():
-            return self.get_permissions_for_registration()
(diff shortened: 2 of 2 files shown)


wagtail/contrib/modeladmin/options.py
```diff
@@ -556,12 +561,23 @@ def modify_explorer_page_queryset(self, parent_page, queryset, request):
 
     def register_with_wagtail(self):
         super().register_with_wagtail()
+        self.register_modeladmin_model_check_with_wagtail()
```
ababic, line 564: We could avoid this override on both ModelAdmin and ModelAdminGroup by adding a `register_modeladmin_model_check_with_wagtail()` method to the base class that doesn't do anything (just `pass`)

wagtail/contrib/modeladmin/options.py
```diff
@@ -28,33 +28,37 @@ class WagtailRegisterable:
     exclude_from_explorer = False
 
     def register_with_wagtail(self):
+        self.register_permissions_with_wagtail()
+        self.register_admin_urls_with_wagtail()
+        self.register_admin_menu_item_with_wagtail()
+        self.register_explorer_page_queryset_with_wagtail()
```
ababic, line 34: Descriptive naming is good, but a `ModelAdmin` class only registers itself with Wagtail, so I think we can probably lose the `_with_wagtail` from these new method names.

wagtail/contrib/modeladmin/options.py
ababic: It might be nice to shorten the name to 'register_checks' too, as they are just regular django checks, that don't necessarily have to be about the model.

docs/reference/contrib/modeladmin/menu_item.rst
```diff
@@ -73,3 +73,15 @@ This will only work for indivdual ``ModelAdmin`` classes registered with their
...
+------------------------
+``ModelAdmin.menu_hook``
+------------------------
+
+**Expected value**: A hook name for registering the new menu item.
+
+If you'd like menu item to be registered in a custom ``Menu``, set
+``menu_hook`` to the value from Menu's ``register_hook_name``.
+
+This will only work for indivdual ``ModelAdmin`` classes registered with their
```
ababic, line 85: indivdual > individual

ababic:
@tm-kn thanks for the updates. All we really need now is a short unittest to show that the relevant hook method gets called, depending on the value of `menu_hook`.

vupham04:
Hi, I'm trying to write unittest for this pull request because I notice that it needs an unittest.

I am trying to replicate the behavior of the code, and when I run the tests (without adding new test) after adding the code from this pull request, the tests show 2 error tests like this: 
 
image

Is this an expected behavior for the code? If there's no warning to be there, am I supposed to just delete the two tests in the images above?

ababic:
Hi @vupham04,

This PR is really quite dated, so would probably benefit from a rebase  onto `main`, which will likely prove tricky with the code restructure that happened recently :(

laymonage:
Hello, I'm closing this as we no longer accepting PRs for new ModelAdmin features in Wagtail as per wagtail/rfcs#85. If you'd like, you could create a new PR to the [wagtail-modeladmin](https://github.com/wagtail-nest/wagtail-modeladmin) repository. Thank you!

PR #6128, opened 2020-06-08
Add more detail to RoutablePageMixin example

gasman:
The RoutablePageMixin documentation doesn't fully explain that view methods should be written as standard Django views, and stops short of giving a full example, which seems to have led people to come up with ill-advised solutions involving `Page.serve`.

This PR extends the `EventPage` example to include example view methods, and clarifies the situations where calling `serve` is valid.


gasman:
Thanks @kaedroho! Added `live()` and merged in e9e6a8d5e7ada3e2475fc99036ef16779489fc70.

PR #6129, opened 2020-06-08
New UI for listing object uses, including in RichText and StreamField --> rebase 3

kedmundson:
Rebase of #4702, with an additional commit of test fixes and updates vis a vis current `master`:

- Updates query counts for collector tests -- note that the counts have simply been changed so that tests pass
- Adds test app merge migration `0052_merge_20200603_2337.py`
- Loads missing `wagtailadmin_tags` in image templates, removes an obsolete test and unused imports


gasman:
Thanks for keeping this up to date @kedmundson!

Right now I think the main thing standing in the way of merging this is that it really needs refactoring to be less "monolithic". As it stands, the logic in collectors.py is touching on the internal details of many areas of Wagtail (rich text, streamfields, foreign key on_delete behaviour...) which is going to make it hard to maintain - for example, if we extend rich text to support other kinds of object references beyond linktype and embedtype (see #4223) then it's not immediately clear what needs to change in collectors.py.

I think the key to that will be making the concept of "a collector" into a well-defined internal API, so that whenever we have a field type that can contain object references (which currently includes ForeignKey, RichTextField and StreamField), there's a corresponding Collector class that implements a few standard methods for querying those references. After the previous review on #4702 (which teased out a lot of the internal workings of collectors.py and added docstrings) we hopefully have a fairly good idea of what those methods will be - something along the lines of:

* `my_collector = RichTextCollector(model_class, field_name)` - constructor for a collector that can find object references in one particular field of one particular model
* `my_collector.find_candidate_references(obj)` - return a queryset of objects (of type `model_class`) that _may_ contain references to the given object, within the field under consideration. This will be done through a fast query (e.g. a regexp match) that may return false positives; these will subsequently be verified individually via `get_object_references`.
* `my_collector.get_object_references(instance)` - given an instance of `model_class`, return a list of `(model_class, id, reference_type)` tuples for all object references found in the field under consideration. `reference_type` will be an identifier similar to the `on_delete` parameter which indicates what the effect of a deletion will be (cascade, protect, set null, ignore etc).

Given how rich text and streamfields interact in RichTextBlock, it may also be necessary to add some lower-level methods to this API, such as "return a database-specific regexp for matching an object reference".

Hopefully splitting the logic up in this way will also help to make our tests more tightly-scoped, and pin down what's causing the changing query counts.

Siburg:
I have been gratefully using the functionality from `collectors.py` in this PR for a project I'm working on. Will just note that I had to add following lines to `find_objects_in_rich_text` function to avoid a `KeyError: 'data-app-name'`.
```
            if hasattr(handler, 'identifier') and handler.identifier == 'snippet':
                # The SnippetEmbedHandler and SnippetLinkHandler use the
                # attrs below. So add them with appropriate values.
                attrs['data-app-name'] = attrs['app-name']
                attrs['data-model-name'] = attrs['content-type-name']
``` 
Those lines go immediately after `handler = registry[attrs[name_attr]]`.

Wishing you all the best for progressing this functionality.

gasman:
An alternative mechanism for reference-counting has now been implemented in #9279. Rebuilding the deletion confirmation page from this PR on top of that mechanism should be a relatively small task, at which point this can be considered resolved.

gasman:
Completed in #10072

PR #6140, opened 2020-06-11
Fix another typo in 2.10 changelog/release notes

acarasimon96:
This time, it's a misspell of XLS(X) as "XLX(S)," which the latter doesn't exist for an Excel spreadsheet file.


lb-:
Thanks for picking that up @acarasimon96

Issue #5482, opened 2019-08-04
Feature request: Allow adding extra permission types to per-object permissions table


dmeijboom:
Currently the "per object" permissions include "change", "delete" and "add" but "view" is also a default permission in django.contrib.auth. Is it possible to add this as well? I think it's nothing more than adding an entry to this array: https://github.com/wagtail/wagtail/blob/fdd8ddc2b183421eeb61231e8a724cdb34aa490c/wagtail/users/templatetags/wagtailusers_tags.py#L60

Wagtail version: 2.6

gasman:
Adding the checkboxes to the 'edit group permissions' form would be easy enough, but that might be misleading to users, as those permissions aren't enforced anywhere in Wagtail, and we'd have to decide what they mean in the context of Wagtail, especially since Wagtail has a separate concept of 'view restrictions' on pages and documents.

One requested feature that might align well with this is a 'choose' permission (#4288, #4488) to determine which items appear in the 'choose page' / 'choose image' etc popups.

We also need to keep in mind that view permissions were [added to Django in 2.1](https://docs.djangoproject.com/en/2.2/releases/2.1/#model-view-permission) - we need to ensure that things continue to work correctly in Django 2.0, or wait for a future release where we drop Django 2.0 support.

dmeijboom:
Hmm, maybe I can open a PR which simply allows adding more permissions next to "change" & "delete"? Note that I created this issue specifically for permissions of user-defined django models (and not pages/images).

gasman:
@dmeijboom Sure, that sounds like a worthwhile addition.

lb-:
Resolved via #7488

PR #6143, opened 2020-06-12
Django view permission intergration using InspectView

Joep110:
This PR fixes issue: #5482, integrated view Permission included in Django >= 2,1 using the InspectView. This will standard enable InspectView for all objects, and moved the inspect button to the last in the list.


gasman:
Hi @Joep110 - thanks for the contribution.
Unfortunately this doesn't really address the concerns raised in #5482 - it's imposing one specific interpretation of "view permission" that is only meaningful to users who have modeladmin enabled. For users who don't manage pages through modeladmin, this new permission will have no effect and is likely to be confused with Wagtail's "private pages" feature.

PR #6146, opened 2020-06-12
[Workflow] Allow editing of groups after GroupApprovalTask creation

jacobtoppm:
Allow editing GroupApprovalTask groups after task creation - after user feedback, this option seems preferred.


zerolab:
Fixed in #6172, so closing

PR #6149, opened 2020-06-12
ImageChooser will be opened at the same place where it was closed 

gassan:
The  GET 'q', 'tag', 'p' and 'collection_id' parameters are stored in the session.

My point: I have tons of images that all have the same names like IMG_0000.jpg. So I can't really make use of searching (with documents and snippets that may be different), so if I need for my new article 5 pictures which are on page 10, I have to navigate 5 times to page 10, because the chooser always opens on page 1.
The parameters WAGTAILIMAGES_CHOOSER_STORE_GET_PARAMS = False allows User to switch to old behavior and the session remains almost maiden clean.

P.S. with a request modifications hook i could not solve this problem. As soon as one of the GET parameters ('q', 'tag', 'p' and 'collection_id') is set, the server assumes that the chooser is already opened (which is not true) and sends only a part of the chooser content.

yes
* Does the code comply with the style guide? (Run `make lint` from the Wagtail root)pip
yes
yes
No, that must be done,

Please feel free to change the variables or constant names. they may be not best choice

### DIFF
--- a/wagtail/images/tests/test_admin_views.py
+++ b/wagtail/images/tests/test_admin_views.py
@@ -791,6 +791,40 @@ class TestImageChooserView(TestCase, WagtailTestUtils):
         self.assertEqual(len(response.context['images']), 1)
         self.assertEqual(response.context['images'][0], image)
 
+    @override_settings(WAGTAILIMAGES_CHOOSER_PAGE_SIZE=4)
+    def test_stored_params_q_and_p(self):
+        for i in range(1, 15):
+            self.image = Image.objects.create(
+                title="Very nice image %i" % i,
+                file=get_test_image_file(size=(1, 1)),
+            )
+
+        response = self.get()
+        self.assertEqual(response.status_code, 200)
+        response_json = json.loads(response.content.decode())
+        self.assertEqual(response_json['step'], 'chooser')
+        self.assertTemplateUsed(response, 'wagtailimages/chooser/chooser.html')
+
+        response = self.get({'q': "Very nice image", 'p': 2})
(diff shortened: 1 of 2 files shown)


lb-:
@gassan thanks for submitting this, I have flagged this for review.

Could you maybe update the description with a bit more explanation about the problem this solves and how to validate the changes. Also, why a new settings item is needed `WAGTAILIMAGES_CHOOSER_STORE_GET_PARAMS`.

Is there an equivalent problem in document chooser modal or snippet chooser modal?

wagtail/images/tests/test_admin_views.py
```diff
@@ -791,6 +791,40 @@ def filter_images(images, request):
...
+        self.assertTemplateUsed(response, 'wagtailimages/chooser/chooser.html')
+        self.assertIn('value="Very nice image"', response_json['html'])
+        self.assertIn("Page 2 of 2", response_json['html'])
+
+        # response_json = json.loads(response.content.decode())
+        # self.assertEqual(response_json['step'], 'chooser')
+        # self.assertTemplateUsed(response, 'wagtailimages/chooser/chooser.html')
+
+        # # draftail should NOT be a standard JS include on this page
+        # self.assertNotIn('wagtailadmin/js/draftail.js', response_json['html'])
```
lb-, line 826: small note - commented out test items here, can these be uncommented if need or removed if not.

gassan: ups. sorry, the lines are removed.

gassan:
I hope I've explained the problem properly. Englisch is not one of my strong points.

PR #6159, opened 2020-06-18
Feature/workflow permission registration

jacobtoppm:
Registers the workflow permissions properly so they can be added to groups in the admin, allowing non-superusers to set up workflows

Only shows the workflow menu icon if the index view is accessible

Adds tests for accessing workflow views with and without permissions in the Wagtail Admin


kaedroho:
There's a good amount of unit tests so should be easy to integrate this with my branch, thanks!

Issue #6162, opened 2020-06-20
Add a UUIDField to the Page model


ababic:
### Is your proposal related to a problem?

As systems become more and more distributed these days, it seems there is a general consensus in the tech industry that it's a bad idea for public APIs to expose too much information about a system's inner workings - using page's `AutoID` values as identifiers in the API could be classed by some as exposing too much.

Also, we've seen recently (e.g. in wagtail-transfer) that there is value in being able to identify pages across multiple environments, and relying on `AutoID` values in such a scenario is simply not viable.

### Describe the solution you'd like

I'm certainly not suggesting that we replace the `AutoIDField` with a `UUIDField`. It does the job - it's efficient, it's used a lot internally, and it's handy to have a short, human-friendly identifier in a lot of situations. Not to mention it would be practically impossible to pull off from a migration/backwards-compatibility perspective.

What I AM suggesting though, is that we have a separate UUID field (I would suggest `public_id` as the name, but that's by-the-by). The field would have a unique constraint, but it wouldn't replace the primary key. It would be automatically generated at initialisation and available before saving a page (possibly useful in some scenarios), and would serve as an alternative way of referring to the page when using the `AutoID` isn't so ideal.

You could argue that having more than one identifier for a page is a bit redundant. Some might even use words like 'messy' or 'ludicrous' - and I would have to agree to a point; I wouldn't bake this into something by design. But, I believe the pluses outweigh the negatives here. The overhead of generating and storing a single UUID value for each instance isn't really worth mentioning, and there's nothing to say that we have to make any other changes immediately (although it would be nice to update a few places to support them instead of / as well as the id).

### Describe alternatives you've considered

The alternative is that when a project needs to have a UUID for a page, it is added to a custom page model, and the API functionality is overridden. And in apps like `wagtail-transfer`, the value is generated and stored in a separate table (which must certainly have an impact on join performance at the database level).

thenewguy:
Awesome! Do you plan to transparently replace the id in the API using this as the natural key?

ababic:
@thenewguy that is the idea, yes. but I thought it best to leave that to a separate PR/discussion and keep this change simple.

thibaudcolas:
For anyone interested in this, I think [RFC 115: Write API](https://github.com/wagtail/rfcs/pull/115) includes enough API-related changes that this could be bundled in. I don’t think the API needs UUIDs but they’d certainly be useful.

PR #6164, opened 2020-06-21
Add uuid field to Page model

ababic:
Related issue: #6162

Possibly resolves: #6917.

Takes care of adding the field and generating initial values (based on `url_path`) via a data migration.

Any further changes relating to #6162 can be added via separate PRs.

The most contentious part of this is probably the data migration, as that will quite obviously take a while to run on sites with large numbers of pages. Thankfully, we're only supporting Django 2.2 and later now, so we can utilise `bulk_update()`. We can likely increase the batch size used here in order to blast through things more quickly. Does anybody know if we have any 'minimum requirements' anywhere that would help indicate how much memory we have to play with?


wagtail/core/migrations/0047_add_page_uuid.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('wagtailcore', '0046_site_name_remove_null'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='page',
+            name='uuid',
+            field=models.UUIDField(editable=False, null=True),
```
kaedroho, line 16: I noticed this isn't `unique=true`. So I'm wondering if this could be used by wagtail-localize to link translations of the same page?

Currently, this is done with a shared UUID: https://github.com/wagtail/wagtail-localize/blob/master/wagtail_localize/models.py#L97

kaedroho:
Thanks for submitting this @ababic!

This actually overlaps a bit with a proposal that I wanted to make at some point in the future, which was to introduce a `locale` (FK to a Locale model) and `translation_key` (UUID) fields on `Page`. In my proposal, the UUID would only be unique within a locale and not for the whole instance (the same UUID will be shared between translations of the same page), but a locale/UUID combination would be globally unique. Effectively, the UUID would still represent a single page, but that page could exist in multiple languages on a Wagtail instance.

ababic:
Cheers for taking a look @kaedroho. I considered the translation stuff when adding this, and I think the use-cases are similar, but different. This is more like an 'alternative primary key' - so I believe the value here should be universally unique. `translation_key` in your example has different requirements, and so would simply be separate field. 

Off-topic, but I wonder whether introducing a `PageTranslation` model would be valuable there, and `translation_key` would become a ForeignKey to that model? That way, you could store a reference to the 'original language version' there (amongst other things), and you could use `translation_id` to identify translations just as efficiently?

kaedroho:
> Off-topic, but I wonder whether introducing a PageTranslation model would be valuable there, and translation_key would become a ForeignKey to that model? That way, you could store a reference to the 'original language version' there (amongst other things), and you could still use the translation_id column on the Page model to identify related pages as efficiently?

Sort of like the `TranslatableObject` model in the translation app? https://github.com/wagtail/wagtail-localize/blob/master/wagtail_localize/translation/models.py#L39

I had considered this, but I went with the translation key approach mainly because it was easy to set the default value, so there won't be any cases where the field is `None` to handle. This might not be much of a problem if this was implemented in core though!

ababic:
> EDIT: Hmm, actually, you got me thinking now. None would be beneficial as it would tell us the page has no translations, which might optimise some queries... We only need to set it when copy_for_translation is called.

This exactly :) Also, it could be convenient to have a model at that level in case it were needed in future (to store data relevant to all translations of a page)

ababic:
Closing due to lack of interest

PR #6173, opened 2020-06-23
Workflow reports UI tweaks

kaedroho:
Small tweaks to the workflow reports UI, as discussed in standup this morning.

See commit log for details.


kaedroho:
Thanks for the quick review @zerolab!

The path can't be a sub-path of `/workflows` because this will make both menu items display as active when someone is looking at workflow tasks.

zerolab:
> The path can't be a sub-path of /workflow because this will make both menu items display as active when someone is looking at workflow tasks.

Ah, I knew I was missing something obvious! :shipit:

PR #6175, opened 2020-06-24
Disable password auto-completion on UserForm password fields.

solarissmoke:
If you use the web browser's password manager to store the login credentials for a Wagtail site, it will auto-populate these credentials into the password field of the `UserForm` when you go to edit a user. This is not the desired behaviour, because the fields are for editing another user's password rather than logging in with your own.

This patch adds the `autocomplete="off"` [attribute](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion) to these inputs to prevent auto-completion. (We could potentially also use `autocomplete="new-password"`, but I didn't do this because the password suggestions that modern browsers provide would be different for both password1 and password2 fields, which may cause confusion).

### DIFF
--- a/wagtail/users/forms.py
+++ b/wagtail/users/forms.py
@@ -83,11 +83,11 @@ class UserForm(UsernameForm):
 
     password1 = forms.CharField(
         label=_('Password'), required=False,
-        widget=forms.PasswordInput,
+        widget=forms.PasswordInput(attrs={'autocomplete': 'new-password'}),
         help_text=_("Leave blank if not changing."))
     password2 = forms.CharField(
         label=_("Password confirmation"), required=False,
-        widget=forms.PasswordInput,
+        widget=forms.PasswordInput(attrs={'autocomplete': 'new-password'}),
         help_text=_("Enter the same password as above, for verification."))
 
     is_superuser = forms.BooleanField(


gasman:
Thanks @solarissmoke!

> We could potentially also use `autocomplete="new-password"`, but I didn't do this because the password suggestions that modern browsers provide would be different for both password1 and password2 fields, which may cause confusion.

Can you elaborate on this, please? I've [spent some time wrestling with Chrome's autofill/autocomplete implementation](https://stackoverflow.com/a/57810447/1853523), and while I haven't tested it on passwords, my understanding is that saved passwords on Chrome are handled by Autofill, which does not respect `autocomplete="off"`. I believe `autocomplete="new-password"` is therefore more likely to be effective on Chrome (and follows the Chrome dev team's previous advice to use the autocomplete attribute to give valid semantic meaning), so I'd prefer to use that unless there's a solid reason against it. Surely a modern browser that recognises `autocomplete="new-password"` will use that as a cue to _not_ provide suggestions?

solarissmoke:
> Can you elaborate on this, please?

@gasman Yeah, sorry, I could have explained this better. On some modern browsers (e.g., Firefox), if `autocomplete="new-password"` is set, then the browser will offer to populate a (new) randomly generated password for you, instead of you having to come up with your own. (Here's the [ticket in which this feature](https://bugzilla.mozilla.org/show_bug.cgi?id=1565407) was added to Firefox, version 70. I haven't checked how other browsers behave.)

This would be confusing here because we have two password fields (password and confirmation fields). The browser doesn't recognise this and will suggest different randomly generated passwords for each field (verified on Firefox), which is of course going to fail validation if submitted. Hence my suggestion that `off` is a better choice in this case.

> Surely a modern browser that recognises autocomplete="new-password" will use that as a cue to not provide suggestions?

I think there's an important distinction between auto-completion and suggestion. `autocomplete="new-password"` appears to be a signal that the browser should in fact suggest a new, secure password.

gasman:
Wow, that's pretty alarming... surely 99% of "new password" forms on the web must have a 'confirm new password' field that would be affected by that?

solarissmoke:
> Surely 99% of "new password" forms on the web must have a 'confirm new password' field that would be affected by that?

So, I just tested again and it turns out that Firefox populates the same value into both fields, at least in this particular instance (I don't know why I got different results the first time - unless I just wasn't paying attention). So `new-password` would actually not cause a problem here. Should I switch to using that?

gasman:
Yes please!

solarissmoke:
Done.

gasman:
Merged in30419231e58bc5fc3897a3b61c795caaa3314f05.

PR #6157, opened 2020-06-17
WIP: Workflow management UI changes

kaedroho:
This PR updates the workflow management UI to reflect the designs that have been done recently.

Improvements include:

 - Made the workflow tasks view accessible from settings menu
 - Added number of pages a workflow applies to column to the index
 - Redesigned how tasks are displayed on workflows index
 - Added chooser modal for tasks, that allows creating and editing tasks
 - Improved assigning to page UI to work as a formset on the workflow edit view (instead of separate view)
 - Cleaned up usage of edit handlers
   - Tasks use plain Django forms, similar to Images
   - All code related to edit handlers moved and buried in forms/workflows.py as it's internal

This is still WIP. I'm posting this to make use of CI.

PR #6178, opened 2020-06-25
[Workflow management UI changes 5] Added workflow usage

kaedroho:
This adds a usage view and usage info to the workflow index

Separated from: #6157


kaedroho:
> But I do wonder what the impact of .all_pages() will be on large sites

I wonder too, but I think the complexity of the query mainly depends on the total number of workflows and not much on the total number of pages. I don't think sites would have very many workflows, even if they have millions of pages.

PR #6181, opened 2020-06-25
Only run checks on children for tabs that have children

seb-b:
Kind of an edge case but it was a bit annoying. I wanted to use a bare `EditHandler` subclass to display some info to a content editor in a separate tab:

```python

class ProjectVotesTab(EditHandler):
    template = 'edit_handlers/project_votes.html'
    def render(self):
        return mark_safe(render_to_string(self.template, {
            'self': self,
            'instance': self.instance,
        }))

class DemoPage(Page):
    # Usual business here
        ObjectList(content_panels, heading='Content'),
        ProjectVotesTab(heading='Votes'),
        ObjectList(Page.promote_panels, heading='Promote'),
        ObjectList(Page.settings_panels, heading='Settings', classname="settings"),
    ])
```

This triggered a `AttributeError: 'ProjectVotesTab' object has no attribute 'children'` from the changed code. The workaround was to add a dud attribute `children = []`, this change drops the need for that.

### DIFF
--- a/wagtail/admin/checks.py
+++ b/wagtail/admin/checks.py
@@ -93,7 +93,7 @@ def inline_panel_model_panels_check(app_configs, **kwargs):
 def check_panels_in_model(cls, context='model'):
     """Check panels configuration uses `panels` when `edit_handler` not in use."""
     from wagtail.core.models import Page
-    from wagtail.admin.edit_handlers import InlinePanel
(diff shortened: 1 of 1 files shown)


lb-:
@seb-b  thanks for this!
merged in b9347ee97047175a47d0b88889c4461d22847dd3

PR #6183, opened 2020-06-26
Purge image rendition cache on rendition deletion

### LINKED ISSUE #6096: Broken image tags when image is changed with rendition caching enabled
### Issue Summary

With rendition caching enabled, changing an image results in broken image tags (old file paths remain cached).

### Steps to Reproduce

2. [Enable rendition caching](https://docs.wagtail.io/en/v2.9/advanced_topics/performance.html?highlight=RENDITIONS#caching-image-renditions)
3. Edit home/models.py as follows
```python
from django.db import models

from wagtail.images.edit_handlers import ImageChooserPanel


class HomePage(Page):
    image = models.ForeignKey(
        'wagtailimages.Image', on_delete=models.SET_NULL, related_name='+', null=True, blank=True
    )

        ImageChooserPanel('image')
    ]
```
4. Edit home/templates/home/home_page.html as follows
```html
{% extends "base.html" %}
{% load static wagtailimages_tags %}

{% block content %}
{% image page.image width-200 %}
{% endblock content %}
```
5. Open the page in the frontend, ensure the rendition is cached
6. Change the image in the Wagtail admin (Change image file)
7. Reload the page from step 5, a broken image tag is now rendered because an old image path is still cached

I would expect the cache to be cleared when the image is changed.

### Technical details

* Python version: 3.8.2
* Django version: 3.0.5
* Wagtail version: 2.9.0

gasman:
Fixes #6096

### DIFF
--- a/wagtail/images/models.py
+++ b/wagtail/images/models.py
@@ -278,12 +278,12 @@ class AbstractImage(CollectionMember, index.Indexed, models.Model):
             filter = Filter(spec=filter)
 
         cache_key = filter.get_cache_key(self)
-
+        Rendition = self.get_rendition_model()
 
         try:
             rendition_caching = True
             cache = caches['renditions']
-            rendition_cache_key = "image-{}-{}-{}".format(
+            rendition_cache_key = Rendition.construct_cache_key(
                 self.id,
                 cache_key,
                 filter.spec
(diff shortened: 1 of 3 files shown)


tomdyson:
Thank you @gasman, this looks great.

pascalw:
Great, thanks! 👍

PR #6186, opened 2020-06-26
Add "used on workflows" column on Task index and task chooser

kaedroho:


zerolab:
Cheers @kaedroho, tweaked and merged in 39b84fc15dd

PR #6190, opened 2020-06-29
Extend fixtree to fix path problems and apply to Collections

gasman:
Extend the `manage.py fixtree` command to fix both Page and Collection trees, and fix path problems (holes in the path sequence, and nodes not ordered with respect to `node_order_by`). The latter is copied from https://github.com/django-treebeard/django-treebeard/pull/165, pending that PR making it into a release of django-treebeard.

As per https://github.com/wagtail/wagtail/pull/6141#discussion_r445483522, this is a prerequisite for #6141, since nested collections mean we can no longer order collections by name at query time, and so we need to provide a migration path for people relying on that ordering (#4874); we can do that by adding `node_order_by=['name']` to the Collection model, and instructing them to run `./manage.py fixtree --full` to apply name ordering to their existing collections.


coredumperror:
Out of curiosity, why doesn't this PR itself take the step of setting `node_order_by=['name']` on the Collection model?

gasman:
> Out of curiosity, why doesn't this PR itself take the step of setting `node_order_by=['name']` on the Collection model?

We generally want PRs to do a single standalone thing, for ease of reviewing and ensuring we're not dropping in too much code at once. In this case, the improvements to the fixtree command are independent from the new collections behaviour. (Admittedly, the benefits of this change are somewhat limited until we start using node_order_by - it might possibly have some relevance to #882 though...)

ababic:
Merged in 8f8f2e1. Thank you @gasman!

PR #6191, opened 2020-06-30
Remove Django 2.1 from install_requires

mikemonteith:
Since Wagtail 2.9, Django >=2.2 is required.

See documentation: https://docs.wagtail.io/en/v2.9/releases/upgrading.html#compatible-django-python-versions


gasman:
Thanks @mikemonteith!

PR #6194, opened 2020-07-01
ReadOnlyPanel continued

brylie:
Continued from #6067 based on the discussion in #2893

### DIFF
--- a/docs/reference/pages/panels.rst
+++ b/docs/reference/pages/panels.rst
@@ -276,6 +276,13 @@ SnippetChooserPanel
 
     See :ref:`snippets` for more information.
 
+ReadOnlyPanel
+----------
+
+.. class:: ReadOnlyPanel(field_name)
+
+    This panel will render a read-only representation of a field. E.g. when you don't expect user input.
+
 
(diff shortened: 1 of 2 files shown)


marteinn:
Thanks for the work so far! Some comments and ideas I have.

- Is there a reason why heading is not auto-populated as with `FieldPanel`? 
- We should follow the existing pattern in the edit_handler module and move markup to external template files.
- I think we should support more fields types (right now when I try to use a image nothing gets shown). I'm pretty certain we will otherwise get a lot of questions on this. 
- We should add tests to simplify future maintenance.

Pomax:
gentle bump to hopefully not have this work get lost - it'd be super valuable still =)

jeffykle:
How can I render the help_text?

brylie:
@KalobTaulien any progress on reviewing this PR?

thibaudcolas:
For anyone interested in this, more recent implementation: #9129

Pomax:
@thibaudcolas I assume this can be closed, then?

zerolab:
Let's mark this as closed. Good shout @Pomax

PR #6196, opened 2020-07-01
Add non-field errors to image/document add/edit forms

gasman:
The upload and edit views for images and documents don't currently output `form.non_field_errors` on the template. As a result, if a custom image/document model includes validation that triggers such an error, such as a unique_together constraint on title and collection, the user will receive a generic "the image could not be updated due to errors" message but no specific message saying what the error is (or, in the case of the inline edit forms for the multiple uploader, no message at all).


gasman:
Merged in 34785b9085465f19b0666bffd3d1946292ae2eb5 + parents

PR #6199, opened 2020-07-03
[Workflow] Only show active workflows in task usage

zerolab:
and cleans up a duplicate property


zerolab:
The Travis build passed, so merged in 77c8b71

PR #6204, opened 2020-07-03
[Workflow] Add requested by filter on the workflow report

zerolab:
Adds a "requested by" user filter to the workflow report, limited to the set of user that initiated a workflow, rather than all the site users


zerolab:
Heh, indeed. The purpose of the filtering is avoid it becoming unwieldy on big sites. The user chooser widget is a stretch task (so as to avoid an ever expanding scope)

PR #6205, opened 2020-07-03
Django 3.1 compatibility fixes

gasman:
Add Django's stable/3.1.x branch to tests in preparation for the 3.1 release, and apply the fairly simple fixes to get them passing again.

django-rest-framework requires additional fixes, which are in master but not yet released.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,8 @@ matrix:
      python: 3.8
    - env: TOXENV=py38-dj30stable-postgres-noelasticsearch
      python: 3.8
+   - env: TOXENV=py38-dj31stable-postgres-noelasticsearch
+     python: 3.8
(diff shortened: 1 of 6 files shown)


gasman:
Looks like the failures in the postgres search were introduced in this commit: https://github.com/django/django/commit/3baf92cf8230ad3a932986170fd07c8feae7ff2f

gasman:
Based on the above Django commit, I checked the codebase for references to `resolve_expression` and landed at this comment from @kaedroho:

https://github.com/wagtail/wagtail/blob/34785b9085465f19b0666bffd3d1946292ae2eb5/wagtail/contrib/postgres_search/query.py#L91

I made the suggested change, and it worked. Of course. :-)

PR #6210, opened 2020-07-06
Replace calls to django.conf.urls.url with path and re_path - rebase/update of #5844

gasman:


gasman:
Merged in 8c306910dd86e09cea11196715da47c6a54c722b + parents.

PR #6213, opened 2020-07-08
Model autocomplete widget

zerolab:
Adds a model autocomplete widget and changes the locked pages "locked_by" filter to use it. Currently supports only single choice.

Screenshot 2020-07-08 at 14 41 40

Tried to keep it as lightweight as I could. All feedback is welcome.

Note: ideally we would fold https://github.com/wagtail/wagtail-autocomplete into core and expand it to allow a separate lookup field and work without needing `autocomplete_label`. If that is deemed the preferred approach, can try to look into getting that package ready for core.

**edit:** related issue #4798

### DIFF
--- a/wagtail/admin/filters.py
+++ b/wagtail/admin/filters.py
@@ -1,11 +1,12 @@
 import django_filters
 from django import forms
+from django.contrib.auth import get_user_model
 from django.utils.translation import gettext_lazy as _
 from django_filters.widgets import SuffixedMultiWidget
 
 from wagtail.core.models import Page
 
-from .widgets import AdminDateInput
+from .widgets import AdminDateInput, AutocompleteWidget
 
 
 class ButtonSelect(forms.Select):
@@ -90,6 +91,11 @@ class WagtailFilterSet(django_filters.FilterSet):
 
 
 class LockedPagesReportFilterSet(WagtailFilterSet):
+    locked_by = django_filters.ModelChoiceFilter(
+        field_name='locked_by',
+        queryset=get_user_model().objects.all(),
+        widget=AutocompleteWidget(get_user_model(), lookup_field='username')
+    )
     locked_at = django_filters.DateFromToRangeFilter(widget=DateRangePickerWidget)
 
     class Meta:
--- a/wagtail/admin/filters.py
+++ b/wagtail/admin/filters.py
@@ -94,7 +94,7 @@ class LockedPagesReportFilterSet(WagtailFilterSet):
     locked_by = django_filters.ModelChoiceFilter(
         field_name='locked_by',
         queryset=get_user_model().objects.all(),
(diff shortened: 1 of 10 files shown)


ababic:
Suggestion: Introduce an `AutocompleteConfig` class, which could be registered by a developer to customise lookup behaviour for a specific model in a declarative way? With this, you could probably remove some complexity from the widget itself, and allow the developer to control:
1. Which fields should be searched
2. What the ordering of results should be
3. Whether different objects need to be surfaced for different users (via a get_queryset(request) method that could be easily overridden)

A few things that I'm not very keen on at all in this implementation are:
1. The way it assumes the existence of a 'title' field. If that isn't passed to the widget, nothing should be passed in the autocomplete URL, and the view could figure something out (maybe utilising `Model._meta.get_fields()` to pick out `CharField` instances?).
2. Limiting to a single search field by design. It would be very little work, and infinitely more useful to support searching on multiple fields out of the box. 
3. How the text representation/label is worked out. We should be going with the standard Django convention here and using `str(instance)` by default (and encourage users to override `__str__()` on their model class)

zerolab:
Thank you for taking the time to look at this @ababic.

> The way it assumes the existence of a 'title' field. If that isn't passed to the widget, nothing should be passed in the autocomplete URL, and the view could figure something out (maybe utilising `Model._meta.get_fields()` to pick out CharField instances?).

It defaults to title as that is what pages have. I thought of picking up `CharField` and (TextField/etc) instances, but it is better to be explicit about the lookup fields and get a 400 when the supplied lookup fields are wrong.

Sorted out 2 and 3.

--

I like the idea of `AutocompleteConfig`, but maybe we should really look into #4798 and implement it as part of that?

ababic:
> It defaults to title as that is what pages have. I thought of picking up CharField and (TextField/etc) instances, but it is better to be explicit about the lookup fields and get a 400 when the supplied lookup fields are wrong.

Instead of failing at the autocomplete view level, how about checking that the value(s) are actual model fields on widget initialisation (using `Model._meta.get_field()`)? If the exception sufficiently explained what the issue was and how to resolve it, that would alleviate my concerns around the default value.

michal-demecki-profil-software:
Any updates on this feature?

thibaudcolas:
FYI we’re considering this as part of the [page editor 2022](https://github.com/wagtail/wagtail/discussions/7739) project – at this stage we’ll review the UX/design implications, and see whether we can make this work better with our new designs.

phildexter:
I don't see this as part of the scope of the page editor, though I do fully support getting something like this in place to help with filtering. Removing from the project list of issues for now, having discussed with @zerolab.

zerolab:
closing this based on https://github.com/wagtail/wagtail/issues/4798#issuecomment-1111448062

PR #6214, opened 2020-07-08
Workflow icon fixes

gasman:


wagtail/admin/templates/wagtailadmin/base.html
```diff
@@ -35,6 +35,8 @@
                               {% if message.level_tag == "error" %}
                                 {# There is no error icon, use warning icon instead #}
                                 {% icon name="warning" class_name="messages-icon" %}
+                              {% elif message.extra_tags == "workflow" %}
```
gasman, line 38: We also need a corresponding special case for the 'lock' / 'unlock' icon, so I'll revise this to move the icon-picking logic into a new template tag rather than letting this if/else sequence get out of control. (It looks like the `{% message_tags %}` tag is also 99% redundant now that we're no longer handling the icons through CSS.)

wagtail/admin/templates/wagtailadmin/pages/workflow_history/list.html
```diff
@@ -35,9 +35,9 @@ <h2>{{ workflow_name }}</h2>
                         {% for task in workflow_state.all_tasks_with_status %}
                             <span data-wagtail-tooltip="{{ task.name }}: {{ task.status_display }}">
                                 {% if task.status == 'approved' %}
-                                    {% icon "success" title=approved_title %}
+                                    {% icon "success" title=approved_title class_name="uniform" %}
                                 {% else %}
-                                    {% icon "radio-empty" title=incomplete_title %}
```
gasman, line 40: Now handled in feature/workflow branch via 949d559cb81070623e36f8541827f85150a4f341.

gasman:
Replaced by #6246 (workflow no longer uses its own message styles, but lock/unlock does, so needs a fix targeted at master).

PR #6216, opened 2020-07-09
Reimplement PageQuerySet.specific() to utilise select_related() instead of making additional queries

ababic:
Using an approach inspired by [Jeff Elmore's](https://jeffelmore.org/2010/11/11/automatic-downcasting-of-inherited-models-in-django/) implementation of [`InheritanceManager`](https://django-model-utils.readthedocs.io/en/stable/managers.html#inheritancemanager) from `django-model-utils`, this PR utilises `Model._meta` and `QuerySet.select_related()` to fetch specific pages in a single query instead of making separate queries to fetch the pages for each type.

### What are the benefits?

1. Improved speed: Fewer queries overall means less time spent communicating with the database and interpreting the results.
2. Fetching the page data in a single query means that combining with `prefetch_related()` becomes viable, which is obviously very useful for query optimisation.
3. Any other values requested via `select_related()` and `annotate()` are also included on the returned values.
5. Because the database result returns all of the data needed for each row (in the order it is needed), we can use a 'generator style' to return results (instead of building an in-memory list of combined results). This alllows `specific()` to be combined with `iterator()` to vastly reduce peak memory usage when processing with large querysets.
6. In cases where the specific version of a page cannot be found, no additional queries are needed to fetch the generic page data (to return as a backup), as the generic page instance is already present in the result.
7. The implementation penalises the developer less when `specific()` is used in places where it would be clearly of no benefit.

### What are the negatives?

1.  Memory usage over time in python is higher, because using `select_related()` activates a lot of additional processing before the query, and while processing it. In addition to this, a full generic instance is always created in-memory in order to 'get to' the specific one. 

### DEPENDS ON

To be ticked off as merged:

- [x] #6845 
- [x] #6846
- [x] #6847
- [x] #6850

### TO DO

- [x] Do some benchmarking to compare performance against existing implementation
- [x] Remove existing `specific()` method and iterators
- [x] Address failing tests and add new ones
- [ ] Update documentation

### DIFF
--- a/wagtail/core/query.py
+++ b/wagtail/core/query.py
@@ -137,16 +137,24 @@ class PageQuerySet(SearchableQuerySetMixin, TreeQuerySet):
     def __init__(self, *args, **kwargs):
         """Set custom instance attributes"""
         super().__init__(*args, **kwargs)
-        # set by defer_streamfields()
+        # custom iterable to conditionally apply upcasting
+        self._iterable_class = PageIterable
+        # PageIterable utilizes the following values:
+        # updated by specific() and generic()
+        self._yield_specific_instances = False
+        self._defer_specific_fields = False
+        # updated by defer_streamfields()
         self._defer_streamfields = False
-        # set by type(), exact_type() or page()
+        # updated by type(), exact_type() or page()
         self._types_requested = set()
-        # set by not_type() or not_exact_type()
+        # updated by not_type() or not_exact_type()
         self._types_excluded = set()
 
     def _clone(self):
         """Ensure clones inherit custom attribute values."""
         clone = super()._clone()
+        clone._yield_specific_instances = self._yield_specific_instances
+        clone._defer_specific_fields = self._defer_specific_fields
         clone._defer_streamfields = self._defer_streamfields
         # these sets must be recreated in order to avoid
         # unintended cross-queryset contamination
@@ -154,14 +162,20 @@ class PageQuerySet(SearchableQuerySetMixin, TreeQuerySet):
         clone._types_excluded = self._types_excluded.copy()
         return clone
 
-    def __and__(self, other):
+    def __and__(self, other: 'PageQuerySet'):
         combined = super().__and__(other)
+        combined._yield_specific_instances = self._yield_specific_instances or other._yield_specific_instances
+        combined._defer_specific_fields = self._defer_specific_fields or other._defer_specific_fields
+        combined._defer_streamfields = self._defer_streamfields or other._defer_streamfields
         combined._types_requested = self._types_requested & other._types_requested
         combined._types_excluded = self._types_excluded & other._types_excluded
         return combined
 
-    def __or__(self, other):
+    def __or__(self, other: 'PageQuerySet'):
         combined = super().__or__(other)
+        combined._yield_specific_instances = self._yield_specific_instances or other._yield_specific_instances
+        combined._defer_specific_fields = self._defer_specific_fields or other._defer_specific_fields
+        combined._defer_streamfields = self._defer_streamfields or other._defer_streamfields
         combined._types_requested = self._types_requested | other._types_requested
         combined._types_excluded = self._types_excluded | other._types_excluded
         return combined
@@ -408,10 +422,11 @@ class PageQuerySet(SearchableQuerySetMixin, TreeQuerySet):
         field values will be loaded and all specific fields will be deferred.
         """
         clone = self._clone()
+        clone._yield_specific_instances = True
         if defer:
-            clone._iterable_class = DeferredSpecificIterable
+            clone._defer_specific_fields = True
         else:
-            clone._iterable_class = SpecificIterable
+            clone._defer_specific_fields = False
         return clone
 
     def in_site(self, site):
@@ -445,6 +460,21 @@ class PageQuerySet(SearchableQuerySetMixin, TreeQuerySet):
         from the results.
         """
         return self.exclude(self.translation_of_q(page, inclusive))
+
+
+class PageIterable(ModelIterable):
+    """
+    A custom iterable class used by ``PageQuerySet`` to support on-the-fly
+    upcasting of generic ``Page`` instances to their 'most specific' type.
+    """
+    def __iter__(self):
+        if not self.queryset._yield_specific_instances:
+            iterable = super()
+        if self.queryset._defer_specific_fields:
+            iterable = DeferredSpecificIterable(self.queryset)
+        else:
+            iterable = SpecificIterable(self.queryset)
+        yield from iterable
 
 
 class SpecificIterable(ModelIterable):
--- a/wagtail/core/query.py
+++ b/wagtail/core/query.py
@@ -1,13 +1,12 @@
 import posixpath
 import warnings
 
-from collections import defaultdict
-
 from django.apps import apps
 from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import ObjectDoesNotExist
 from django.db.models import CharField, Q
 from django.db.models.functions import Length, Substr
-from django.db.models.query import BaseIterable, ModelIterable
+from django.db.models.query import ModelIterable
 from treebeard.mp_tree import MP_NodeQuerySet
 
 from wagtail.search.queryset import SearchableQuerySetMixin
@@ -229,7 +228,7 @@ class PageQuerySet(SearchableQuerySetMixin, TreeQuerySet):
         This filters the QuerySet to exclude any pages which are an instance of the specified model(s)
         (matching the model exactly, not subclasses).
         """
-        return self.exclude(self.exact_type_q(*types))
+        return self.exclude(self.exact_type_q(*types, excluding=True))
 
     def public_q(self):
         from wagtail.core.models import PageViewRestriction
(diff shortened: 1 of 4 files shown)


wagtail/core/query.py
```diff
@@ -179,43 +198,61 @@ def not_page(self, other):
...
+        This filters the QuerySet to only contain pages that are direct
+        children of the specified page.
+
+        Data from the parent page's class is used to limit the 'upcast'
+        lookups, to only include joins for page types that are allowed
+        to exist in this part of the tree.
+        """
+        parent_class = parent_page.specific_class
+        if parent_class is not None:
+            self._specified_classes = parent_class.allowed_subpage_models()
```
kaedroho, line 254: I don't think we can rely on `allowed_subpage_models` for covering all of the possible models. This is only ever checked by the Wagtail admin and they can be changed over time without any cleanup.

kaedroho:
Does this allow us to defer the fields of the specific models?

For example, I'm thinking a listing of 100 blog posts would get quite slow if we're now required to load their body text which could be quite long streamfields.

wagtail/core/query.py
ababic: That is a shame :( child_of() is probably the most commonly used filter (especially within wagtail itself) - and it provides an excellent opportunity simplify the query. when `subpage_classes` is used well.

ababic: I suppose we could do a little query to identify all content_types used at this level, but I would guess that in sites with quite a small number of page types, that might be less efficient than just including joins for all types.

ababic:
@kaedroho 

> Does this allow us to defer the fields of the specific models?
> For example, I'm thinking a listing of 100 blog posts would get quite slow if we're now required to load their body text which could be quite long streamfields.

In cases like that, wouldn't you start with `BlogPage.objects.all()` and apply `values()` or `defer()` to select only what you want?

wagtail/core/query.py
kaedroho: If you don't mind a bit of complexity, we could possibly use a bloom filter for this. 

This would mean adding a 32-bit integer on each page. Whenever a new page is created, we set a bit on the parent page's bloom filter corresponding to the ID of the content type we just created wrapped to 32 bits. (eg `parent.bloom |= 1 << (new_page.content_type_id % 32)`)

Then in the query, you do a little calculation to find which page types have content type IDs that would match that filter.

I know it's a bit complicated to what we would usually do, but could give up to a 32x improvement if these `JOIN`s turn out to slow things down.

ababic:
If there's a strong need to be able to deal with specific page instances, but not bother with the database joins, that's totally possible.

ababic:
@kaedroho The changes in 66441f8 _should_ do the job - However, we'll obviously want to add some tests

wagtail/core/query.py
ababic: @kaedroho I do really like this idea - However, I think I'd quite like to tackle that in a separate PR, as this one is already likely to get a little 'sprawly' once I start digging in to remove unnecessary uses of `specific()`. Sounds like there would be some additional considerations there too, like what to do for existing pages, so probably best tackled separately.

kaedroho:
> In cases like that, wouldn't you start with BlogPage.objects.all() and apply values(), only() or defer() to select only what you want?

I was thinking about the page explorer, where it usualyl doesn't use any body fields so having them by default might impact performance here

ababic:
@kaedroho Yep, the plan is to use `generic()` in a few places, and I'll also continue to support the `defer` argument for `specific()` in order to maintain backwards compatibility. However, I'm still unsure about that as an approach. Normally, if I was going to override `get_admin_display_title()` or `get_url_parts()`, it would probably be to make use of custom field data - by only fetching the base page data, but calling those methods, it's probably more likely to negatively impact performance, due to the n queries needed to fetch that data for each individual object.

kaedroho:
I don't think pulling in every specific page field for the sake of a few cases where someone might've overridden that method is a good tradeoff since that would negatively impact performance for everyone.

I tested this when I first implemented the deferring, the non-deferred version used 4x more memory than the deferred version https://github.com/wagtail/wagtail/pull/3355

Is there a way we could let someone specify the fields they need if they have overridden `get_admin_display_title` and we add those to a `.extra()` call?

ababic:
@kaedroho Or where listing lots of items, identify the models up-front, use `Model._meta.get_fields` to identify any streamfields, and automatically exclude them with `defer()`?

ababic:
@kaedroho Please see latest additions in c3949f5. I've implemented a `defer_large_fields()` method which does pretty much what I mentioned in my last comment. I've also added a 'threshold' attribute, which should mark the 'tipping point' by which doing an extra query to identify the types will likely be more efficient than adding a huge number of joins. This should mean that we're alway taking the most efficient approach, and should benefit pretty much all scenarios except where the entire tree if being fetched (which should rarely be the case)

ababic:
Have restored the old behaviour for now to help facilitate side-by-side comparisons

ababic:
Some collected benchmark results can be seen in the following Google sheet:
https://docs.google.com/spreadsheets/d/1hTdLvpHzBZbgycPnl4YFxtVW2xWlPb8BUUNddBZ4iqI/edit?usp=sharing

Some observations I have made:
1. The new implementation is always much faster (40% -> 90%)
2. Memory usage in the new implementation is lower when dealing with smaller numbers of pages, but this lead tales off as we get into processing larger result sets, because the memory that is initially saved by making/processing fewer queries is eventually cancelled out by the memory used to assemble a full vanilla Page instance AND a full specific page instance for each item in the result. 
3. In contrast to the new solution, where the pattern's performance degrades with larger result numbers, the original solution is much more sensitive to variations in page type, which is always going to be less easy to predict. I feel the former is easier to explain and circumvent than the latter.
4. Surprisingly, using `exact_types()` to reduce the number of `select_related` items applied to the query seemed to make very little difference to speed or memory usage - it appears that the number of tables being joined only noticeably slows things down when there's actually joining of data to be done. 
5. I'm quite pleased with the results for `defer_large_fields()` - excluding streamfields in particular makes a big difference not just because less data needs to be transmitted between the database and the app, but also because there's no need for the field to apply `json.loads()` to each value when converting the data into a page object (which, via a bit of profiling, I notice can be quite a big contributor to inconsistencies in the measurements, both in time and memory use)
6. I've tested with larger numbers of pages (500), and the general trend in performance seemed to continue in line with point 2, but memory usage for both solutions really become too unpredictable to be useful in the reporting figures. Using the averages for comparison helps slightly, but even with that, different runs can still produce sporadically better/worse results overall.

ababic:
Another point: The new solution can be combined with `iterator()` in cases where caching of the result is not required, and memory overload is a concern. However, because of the higher volume of data being fetched for each item, this tends to use quite a bit more memory overall - However, that is spread out far more consistently across the overall execution time.

ababic:
A couple more benefits I neglected to mention above:
- The new implementation does not penalise you for calling specific() on a queryset of 'already specific' pages. If the manager's model class has no subclasses, no select_related joins are added to the query, and the queryset evaluates normally.
- In cases where the specific part of the page doesn't exist for some reason (a rare case, admittedly, but still one we cater for) no additional queries are needed to fetch the generic version of the page, as it is already to hand.

ababic:
@gasman / @kaedroho Hold fire on the testing. Just realised there's a bug in the code which is preventing the subclasses from models being identified. I'll have to redo the benchmarks

ababic:
Things are working properly now, but the latest benchmarks seem to indicate a memory leak somewhere, which I'll need to time to look into. Embarrassingly, a lot of my observations above were based on the faulty code (a classic example of the importance of unit tests), so i'll repost some general observations again once I've figured out what's happening on the memory front (execution time improvements are still pretty decent - but not quite as good as erroneously reported)

ababic:
@gasman / @kaedroho. Okay, this is now ready for another look. It turns out there wasn't a memory leak in the end - it's just that using `select_related()` uses some extra memory.

I think it's quite easy to look at the benchmarks and think "The original is so much better", but the benchmark is only a partial truth... using `select_related()` does indeed use a lot more memory than the original, but I don't think we're capturing memory used outside the Python process. Is memory used by the database being included in this? That could throw things out quite a bit. The test suite is also not real life - In production, the database can often be a cloud service on a completely different server... add a bit of distance and I'd imagine the performance of the original starts to drop significantly, and the benefit of making fewer queries starts to shine. Overall memory usage probably isn't a great indicator of anything anyhow. Far more important than that is utilisation of memory over time... moving to a generator-style approach that can yield results and forget about them immediately (when `prefetch_related` isn't used - that requires caching of the result) should make things much more resilient in general.

ababic:
Okay, there seems to be something funny going on with models that have more than one level of multi-table inheritance (as indicated by the test failures, which all seem to be using a subclass of Event Page). I have added a todo to the description to look at this.

ababic:
Okay, I now have things working for multiple levels of inheritance. It turns out you cannot access, say, a third-level page type directly from a generic `Page` - the in-between relationships must be traversed, and each 'interim' model instance loaded into memory in order to get to the next. This seems a little wasteful, but I feel it's unlikely to affect performance detrimentally in most cases. It would likely only be a problem if specific() were being used on large querysets in combination with 'prefetch_related' - as the iterator would cache the entire result in order to join everything together. Maybe it's enough to just warn users against this... or perhaps we need to offer a way to fall back to the previous behaviour if needed (we even could do this automatically by counting the  '__' joins in related names)? I'm open to feedback here.

Anyhow, I feel the time has definitely come for this to be independently field-tested. @gasman / @kaedroho are you still able to help in this regard?

balazs-endresz:
> Memory usage is python is higher, because using select_related() activates a lot of additional processing before the query, and when processing it. In addition to this, the full 'generic' page AND a full 'specific' page object is created in memory for each result during processing, regardless of whether the generic instance is needed.

I just want to comment on the **sitemap** generator specifically because that already uses a lot of memory. Would it be possible to simply drop `.specific()` from `Sitemap.items`? And might as well add `.iterator(chunk_size=500)` at the same time? 

Even if `Sitemap.items` used `defer_streamfields()` (I'm not sure if it does in this PR) I guess `select_related` would still increase peak memory usage, which would make it even worse especially on sites that run on 500MB ram dynos for example.

ababic:
Hey @balazs-endresz!

> Would it be possible to simply drop .specific() from Sitemap.items?

I don't think we could do this, since it's possible to for specific classes to override the methods that determine their URLs, as well as how they contribute to the site map itself. 

> And might as well add .iterator(chunk_size=500) at the same time?

In this new version of `specific()` `iterator()` is applied automatically when evaluating a queryset that is not using `prefetch_related`, so as long as the sitemap is generated in a way that doesn't hold onto object references (allowing for garbage collection) , peak memory usage should be minimal here already (`defer_streamfields()` will be applied too, but in a separate MR).

ababic:
Hi @gasman. Putting a pin in this one for now until dust has settled on other MRs. Essentially, I'm aiming to get as many of the improvements here in as separate MRs, so that we can benefit from them sooner, and this can be 'purified' somewhat to only include the changes to `specific()`

ababic:
@gasman I think I've uncovered the reason for the test failures on this PR: 

The specifying of custom `page_ptr` `OneToOneField` fields on models with `related_name="+"` specified - making it impossible to span the relationship from the parent model in queries.

Do you happen to know under what circumstances specifying a `page_ptr` field be beneficial over allowing Django to generate it automatically, or how testing/support for this became so prevalent in Wagtail? Is it something we could phase out somehow?

ababic:
Closing this because, after doing more testing, it revealed a few issues:

- The extra joins in the query do not come for free: After testing with postgres instead of SQLLite, this became far more apparent, with query performance taking an obvious hit.
- The `ptr_id` field is sometimes overridden with a "+" `related_name` value to get around model name clashes (at least in the Wagtail test app), which prevents access to that model's data with this approach. While we could close that loophole with some extra work, I'm not sure it's worth it, given the first point.

I think the focus should instead be on enhancing the current approach. For example:

#8271

PR #6218, opened 2020-07-12
Add fallback for old signature page listing button hooks

### LINKED ISSUE #6208: Add fallback for old signature of register_page_listing_more_buttons hook
As per https://github.com/wagtail/wagtail/pull/5764#issuecomment-651721677

lb-:
- hook - `register_page_listing_buttons`
- hook - `register_page_listing_more_buttons`
- fixes  #6208
- relates to the changes in #5764
- Note: I have added upgrade considerations to the release notes (hopefully this reads ok).

### DIFF
--- a/docs/releases/2.10.rst
+++ b/docs/releases/2.10.rst
@@ -104,3 +104,9 @@ Python 3.5 is no longer supported as of this release; please upgrade to Python 3
 
 In previous releases, rich text values were enclosed in a ``<div class="rich-text">`` element when rendered; this element has now been removed.
 To restore the old behaviour, see :ref:`legacy_richtext`.
+
+
+Hooks functions ``register_page_listing_buttons`` & ``register_page_listing_more_buttons``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Page listing button hook functions now accept an additional kwarg `next_url`.
--- a/wagtail/admin/templatetags/wagtailadmin_tags.py
+++ b/wagtail/admin/templatetags/wagtailadmin_tags.py
@@ -1,6 +1,7 @@
 import itertools
 import json
 
+import warnings
 from urllib.parse import urljoin
 
 from django import template
@@ -25,8 +26,9 @@ from wagtail.core import hooks
 from wagtail.core.models import (
     CollectionViewRestriction, Page, PageViewRestriction, UserPagePermissionsProxy)
 from wagtail.core.utils import cautious_slugify as _cautious_slugify
-from wagtail.core.utils import camelcase_to_underscore, escape_script
+from wagtail.core.utils import accepts_kwarg, camelcase_to_underscore, escape_script
 from wagtail.users.utils import get_gravatar_url
+from wagtail.utils.deprecation import RemovedInWagtail212Warning
 
 register = template.Library()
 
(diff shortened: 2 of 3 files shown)


wagtail/admin/templatetags/wagtailadmin_tags.py
```diff
@@ -430,9 +432,21 @@ def paginate(context, page, base_url='', page_key='p',
...
-        for hook in button_hooks))
+
+    buttons = []
+    for hook in button_hooks:
+        if accepts_kwarg(hook, 'next_url'):
+            button = hook(page, page_perms, is_parent, next_url)
+        else:
+            warnings.warn(
+                'register_page_listing_buttons hooks will require an additional kwarg `next_url` in a future release. '
+                'Please update your hook function accept `next_url`.',
```
zerolab, line 443:
```suggestion
                'Please update your hook function to accept `next_url`.',
```

wagtail/admin/widgets.py
```diff
@@ -379,6 +382,18 @@ def show(self):
...
-            hook(self.page, self.page_perms, self.is_parent, self.next_url)
-            for hook in button_hooks))
+
+        buttons = []
+        for hook in button_hooks:
+            if accepts_kwarg(hook, 'next_url'):
+                button = hook(self.page, self.page_perms, self.is_parent, self.next_url)
+            else:
+                warnings.warn(
+                    '%s hooks will require an additional kwarg `next_url` in a future release. Please update your hook function accept `next_url`.' % self.hook_name,
```
zerolab, line 392:
```suggestion
                    '%s hooks will require an additional kwarg `next_url` in a future release. Please update your hook function accept to `next_url`.' % self.hook_name,
```

docs/releases/2.10.rst
```diff
@@ -104,3 +104,9 @@ Python 3.5 is no longer supported as of this release; please upgrade to Python 3
 
 In previous releases, rich text values were enclosed in a ``<div class="rich-text">`` element when rendered; this element has now been removed.
 To restore the old behaviour, see :ref:`legacy_richtext`.
+
+
+Hooks functions ``register_page_listing_buttons`` & ``register_page_listing_more_buttons``
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Page listing button hook functions now accept an additional kwarg `next_url`.
```
zerolab, line 112:
```suggestion
Page listing button hook functions now accept an additional kwarg ``next_url``. Please update your hook function definitions to accept the new ``next_url`` kwarg
```

wagtail/admin/widgets.py
```diff
@@ -379,6 +382,18 @@ def show(self):
...
     def dropdown_buttons(self):
         button_hooks = hooks.get_hooks(self.hook_name)
-        return sorted(itertools.chain.from_iterable(
-            hook(self.page, self.page_perms, self.is_parent, self.next_url)
-            for hook in button_hooks))
+
+        buttons = []
+        for hook in button_hooks:
+            if accepts_kwarg(hook, 'next_url'):
+                button = hook(self.page, self.page_perms, self.is_parent, self.next_url)
```
gasman, line 389: This is a bit deceptive... the hook actually returns an iterable of buttons (one for every `yield` in the hook function - there may be multiple) and the `itertools.chain.from_iterable` takes care of turning this iterable of iterables into a final sequence that we can sort. As a result, for someone casually reading this code, this change is inadvertently making the `itertools` line even more voodoo magic than it was before :-)

This could be rectified by renaming `button` to `hook_result` or similar - however, since the 'fancy' code is really just going out of its way to avoid building an intermediate list - which we are now doing - I think it'd be better to turn it fully into non-fancy code...

* this line (and the corresponding one in the 'else' branch) becomes `buttons.extend(hook(self.page, self.page_perms, self.is_parent, self.next_url))` - i.e. append everything returned by the hook to the end of `buttons`
* instead of the `itertools` incantation, the ending just becomes
  ```
  buttons.sort()
  return buttons
  ```

docs/releases/2.10.rst
gasman: In applying this change I realise that we should document the purpose of `next_url`, since it's something that third-party code using these hooks might want to make use of. In turn, that makes me think that next_url should be the _unescaped_ version of the URL, since the escaping is just an implementation detail of how Wagtail's own functions pass that URL on to subsequent views.

Will create a follow-up PR to address this, so that it's not holding up this fix.

gasman:
Merged (with fixes as per my + @zerolab's feedback) in 450f1bef3445f1cf1bb70532f52b79482d8ebcdc

lb-:
Thanks @gasman  & @zerolab

PR #6219, opened 2020-07-13
Clarify origin of From address used in password reset emails

stvnw:
Add a minor clarification that the From address used by the password reset mechanism is set by `DEFAULT_FROM_EMAIL`. Prompted by a recent conversation on Slack where the user was under the false impression this would be governed by `WAGTAIL_NOTIFICATION_FROM_EMAIL`.

(Plus pedantic clarification in the following section about what owns `DEFAULT_FROM_EMAIL`.)


gasman:
Thanks @stvnw! Merged in 5611f5ef6c7947ed1393c83730d1bcda08da39cb.

PR #6222, opened 2020-07-13
[RFC 54] Internationalisation API changes

kaedroho:
This implements the following new features on the pages endpoint:

 - Logic to pull in pages from other locale trees
 - A `locale` filter that filters pages by a Locale's `language_code`
 - A `translation_of` filter that takes a page id and returns the pages that are translations of that page
 - A `meta.locale` field to show the Locale of a page
 - A `meta.translations` field that nests data about a pages translations in each page (admin API only due to performance)

### DIFF
--- a/wagtail/api/v2/filters.py
+++ b/wagtail/api/v2/filters.py
@@ -187,6 +187,37 @@ class DescendantOfFilter(BaseFilterBackend):
         return queryset
 
 
+class TranslationOfFilter(BaseFilterBackend):
+    """
+    Implements the ?translation_of filter which limits the set of pages to translations
+    of a page.
+    """
+    def filter_queryset(self, request, queryset, view):
+        if 'translation_of' in request.GET:
+            try:
+                page_id = int(request.GET['translation_of'])
+                if page_id < 0:
+                    raise ValueError()
+
+                page = view.get_base_queryset().get(id=page_id)
+            except ValueError:
+                if request.GET['translation_of'] == 'root':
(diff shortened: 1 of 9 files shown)


wagtail/api/v2/filters.py
```diff
@@ -187,14 +187,50 @@ def filter_queryset(self, request, queryset, view):
...
+
+
 class LocaleFilter(BaseFilterBackend):
     """
     Implements the ?locale filter which limits the set of pages to a
     particular locale.
     """
     def filter_queryset(self, request, queryset, view):
         if 'locale' in request.GET:
+            _filtered_by_child_of = getattr(queryset, '_filtered_by_child_of', None)
```
gasman, line 228: Maybe out of scope for this PR, but worth considering for cleanup... as far as I can see, `_filtered_by_child_of` is only used to prevent people from passing both `child_of` and `descendant_of` parameters. Doing that is a bit pointless, but surely not logically invalid or ambiguous (all other conditions are ANDed together, so these should be too) - so why are we going to such great lengths (and cluttering up unrelated filter code) to disallow it?

gasman (approved):

wagtail/api/v2/filters.py
kaedroho: I agree, I do like the way Wagtail API is good at giving you errors when you do something wrong but it definitely goes too far in a couple of places! I'll clean this up in another PR!

PR #6225, opened 2020-07-14
Further fixups for button hook next_url parameter

gasman:
Follow-up to #6218:

* Add test for the original issue #3010
* Refactor `next_url` so that the value being passed is the unescaped URL, rather than one embedded into a querystring (which unnecessarily ties it to Wagtail's own parameter-passing scheme and makes it less useful to third-party code)
* Document the purpose of next_url in reference docs and release notes

### DIFF
--- a/docs/reference/hooks.rst
+++ b/docs/reference/hooks.rst
@@ -878,6 +878,13 @@ Page explorer
             priority=10
         )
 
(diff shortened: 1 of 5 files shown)


lb-:
Merged in via 9917d7d2e99d0f0ec90fa0a7b04f149bab2d0792 & parents

PR #6226, opened 2020-07-15
Prepare react-streamfield JS integration

BertrandBordage:
This pull request integrates inside Wagtail all the JS compatibility fixes implemented in [wagtail/wagtail-react-streamfield](https://github.com/wagtail/wagtail-react-streamfield).

The **choosers fixes** are there to fix block duplication. Block duplication in react-streamfield is done by:
1. copying the block template (by “template” I mean the rendered block with a `__PREFIX__` ID in it)
2. running its inline `<script>`
3. changing its value to apply the value fetched from the original block

If we tried instead to copy the modified HTML and running the JS, it would result in a broken state for the block.

Before those chooser fixes, step 3 didn’t work because applying the original value only changed the hidden input state. It was not fetching the thumbnail and file/page/snippet name.

If you are concerned about the extra cost of the AJAX requests, we can always add conditional processing to page chooser Python views to prevent extra server and network cost. But keep in mind that with react-streamfield, StreamFields are considerably faster because we don’t render templates for each block, so the cost of the extra requests stays negligible, even for a huge StreamField.

The **table fix** is here to fix the width of the table that was overflowing from the block. That fix is perfectly compatible with the current StreamField.

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/js/page-chooser.js
+++ b/wagtail/admin/static_src/wagtailadmin/js/page-chooser.js
@@ -4,6 +4,16 @@ function createPageChooser(id, pageTypes, openAtParentId, canChooseRoot, userPer
     var input = $('#' + id);
     var editLink = chooserElement.find('.edit-link');
 
+    function pageChosen(pageData, isInitial) {
+        if (!isInitial) {
+          input.val(pageData.id);
+        }
+        openAtParentId = pageData.parentId;
+        pageTitle.text(pageData.title);
+        chooserElement.removeClass('blank');
+        editLink.attr('href', pageData.editUrl);
+    }
+
     $('.action-choose', chooserElement).on('click', function() {
         var initialUrl = window.chooserUrls.pageChooser;
         if (openAtParentId) {
@@ -23,13 +33,7 @@ function createPageChooser(id, pageTypes, openAtParentId, canChooseRoot, userPer
             urlParams: urlParams,
             onload: PAGE_CHOOSER_MODAL_ONLOAD_HANDLERS,
             responses: {
-                pageChosen: function(pageData) {
-                    input.val(pageData.id);
-                    openAtParentId = pageData.parentId;
-                    pageTitle.text(pageData.title);
-                    chooserElement.removeClass('blank');
-                    editLink.attr('href', pageData.editUrl);
-                }
+                pageChosen: pageChosen,
             }
         });
     });
@@ -39,4 +43,18 @@ function createPageChooser(id, pageTypes, openAtParentId, canChooseRoot, userPer
         openAtParentId = null;
         chooserElement.addClass('blank');
     });
+
+    if (input.val()) {
+        $.ajax({
+            url: window.wagtailConfig.ADMIN_API.PAGES + encodeURIComponent(input.val()) + '/',
+        }).done(function (data) {
+            pageChosen({
+                id: data.id,
+                title: data.admin_display_title,
+                parentId: (data.meta.parent && data.meta.parent.id) ? data.meta.parent.id : null,
+                editUrl: window.wagtailConfig.ADMIN_URLS.PAGES + data.id + '/edit/',
+            }, true
+            );
+        });
+    }
 }
--- a/wagtail/contrib/table_block/static/table_block/js/table.js
+++ b/wagtail/contrib/table_block/static/table_block/js/table.js
@@ -63,13 +63,11 @@ function initTable(id, tableOptions) {
     }
 
     if (!tableOptions.hasOwnProperty('width') || !tableOptions.hasOwnProperty('height')) {
-        // Size to parent .sequence-member-inner width if width is not given in tableOptions
         $(window).on('resize', function() {
             hot.updateSettings({
-                width: getWidth(),
+                width: '100%',
                 height: getHeight()
             });
-            resizeWidth('100%');
         });
     }
 
--- a/wagtail/contrib/table_block/static/table_block/js/table.js
+++ b/wagtail/contrib/table_block/static/table_block/js/table.js
@@ -21,9 +21,6 @@ function initTable(id, tableOptions) {
     var dataForForm = null;
     var isInitialized = false;
 
-    var getWidth = function() {
-        return $('.widget-table_input').closest('.sequence-member-inner').width();
-    };
     var getHeight = function() {
         var tableParent = $('#' + id).parent();
         return tableParent.find('.htCore').height() + (tableParent.find('.input').height() * 2);
--- a/wagtail/documents/static_src/wagtaildocs/js/document-chooser.js
+++ b/wagtail/documents/static_src/wagtaildocs/js/document-chooser.js
@@ -4,17 +4,21 @@ function createDocumentChooser(id) {
     var input = $('#' + id);
     var editLink = chooserElement.find('.edit-link');
 
+    function documentChosen(docData, isInitial) {
+        if (!isInitial) {
+          input.val(docData.id);
+        }
+        docTitle.text(docData.title);
+        chooserElement.removeClass('blank');
+        editLink.attr('href', docData.edit_link);
+    }
+
     $('.action-choose', chooserElement).on('click', function() {
         ModalWorkflow({
             url: window.chooserUrls.documentChooser,
             onload: DOCUMENT_CHOOSER_MODAL_ONLOAD_HANDLERS,
             responses: {
-                documentChosen: function(docData) {
-                    input.val(docData.id);
-                    docTitle.text(docData.title);
-                    chooserElement.removeClass('blank');
-                    editLink.attr('href', docData.edit_link);
-                }
+                documentChosen: documentChosen,
             }
         });
     });
@@ -23,4 +27,11 @@ function createDocumentChooser(id) {
         input.val('');
         chooserElement.addClass('blank');
     });
+
+    if (input.val()) {
+        $.ajax(window.chooserUrls.documentChooser + encodeURIComponent(input.val()) + '/')
+            .done(function (data) {
+                documentChosen(data.result, true);
+            });
+    }
 }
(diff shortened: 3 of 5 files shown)


gasman:
Hi @BertrandBordage,

In the interests of keeping the different components of Wagtail loosely coupled, it's not enough to think in terms of "these are the changes needed to make choosers work with react-streamfield" - I need to know the contract that react-streamfield expects chooser widgets to follow. That way, if I change the chooser code in future, or you change the react-streamfield code, then we're clear on exactly how far we can change things without breaking the contract, and we don't have to keep both halves of the implementation in our heads while doing that...

To the best of my understanding, the contract is something like:

IF a form component makes changes to its own HTML in response to taking on different values, that go beyond simply updating the value of an `<input>` / `<select>` / `<textarea>` element,
THEN it must have an inline `<script>` tag as part of its rendered HTML
AND running the code in that `<script>` tag after modifying the component's `<input>` / `<select>` / `<textarea>` values must trigger the corresponding changes to the HTML.

Does that match your view? Please feel free to rewrite as needed. (To be clear: I'm not saying that this version _should_ be the contract - there are quite a few details I'd disagree with, which is why I'm keen to hear your interpretation rather than my strawman version of it :-) )

BertrandBordage:
Hello @gasman!

Thanks for taking the time to review and explain your position :)

The contract is actually a little different, it’s not necessarily a matter of dynamically changed HTML like for the chooser blocks. It can also be a difference in JavaScript state, even though I didn’t face this case in the existing widgets.

Here is the logic to fully cover what a block should provide:
- **[For duplication]** The HTML & JavaScript state of `YourBlock.render_form(some_value)` interpreted by the browser (script included) must be perfectly equivalent to the HTML & JavaScript state of `YourBlock.render_form(None)` interpreted by the browser (script included) with followed by `$('#your-block').val(some_value)`.
- **[For duplication and StreamField serialization]** Each block input value change must either:
  - change the `value` attribute of the input for `type="hidden"` inputs
  - trigger a JavaScript `change` event for other input types

These are the rules that I want to put in the documentation with the integration of react-streamfield. I want to put it in the `Custom block types` section, with a new subtitle `JavaScript considerations` or something like that. It would explain the first part of the "contract": that you have to include an inline `script` that should base itself on the block ID etc. And of course the rest of the contract above.

What do you think about it?

gasman:
I'm going to ask a very unfair-sounding question... does the react-streamfield widget itself follow this contract?

I ask that not because I seriously expect someone to try embedding BlockField into a FieldBlock... but rather, if we're placing conditions on form widgets to give them standardised client-side semantics, then I'd like the benefits of that process to be available universally across Wagtail, not just in the parts that are going to be used within StreamField.

For example, one item on our wishlist is auto-saving on the page form - it would be great if we could do that by serialising the form content to JSON client-side and pushing that to a REST API endpoint in the background. If we're _only_ considering simple form widgets like Textarea and PageChooserWidget - the sort of things that you'd put into a FieldBlock - then it's easy to assume that rules-of-thumb like "`$('#some-field-id').val()` will return the serializable value" will work (or can easily be hacked into working with some ad-hoc Javascript). If we need it to work on complex components like StreamField itself, that's a more risky assumption.

(In this particular case, I guess react-streamfield _would_ work, since it continuously updates a hidden element with an ID that matches the field name. Classic streamfield definitely wouldn't :-) So really the implication is that if we ever build something in future at the same level of complexity as streamfield, it _has_ to be built using the same programming model as react-streamfield. That might not be a bad thing, but I don't want to be locked in to it either...)

I think the solution I'm really fishing for is for every Django widget object to have a corresponding JS 'adapter' object that knows enough about the client-side representation to perform the tasks that would traditionally be server-side only - tasks like:

* render a fully-initialised instance of this widget with name `foo`, ID `id-foo` and initial value `"hello world"` into container `bar`
* extract the JSON-serializable value of a widget instance
* extract the internal state of a widget instance, i.e. all data required to replicate this widget's value in another instance (- for example, for PageChooserWidget, this might be a dict of page ID, display title, and parent page ID)
* set this widget instance's internal state (as defined above) to the given value

For 'simple' widgets that map directly to HTML inputs, these operations would be trivial one-liners like `$('#foo').val()` - I think we could reasonably provide a default adapter that does this whenever there isn't a more specific one registered. For slightly more complex widgets - like, say, a date picker consisting of separate dropdowns for day, month and year - this would be custom, but still straightforward code (and, I think, simpler than trying to rig up hidden fields and event handlers to make it behave like a single input). And something on the scale of Streamfield can be built by nesting adapters inside other adapters, probably a lot more cleanly and performantly than the current HTML-fragment-soup of classic Streamfield. (Perhaps react-streamfield doesn't need that, but I think it's important to keep the option open...)

To me the key properties of the adapter approach is that it doesn't lock us in to inline `<script>` tags (see: CSP, [RFC 33](https://github.com/wagtail/rfcs/pull/33)...), doesn't assume that the field prefix/ID will directly match an HTML form element, and distinguishes between the serializable value and the internal state needed to fully render the widget. On that last point - I know you ran the benchmarks to verify that the overhead of an AJAX fetch to turn a raw ID into renderable data wasn't a major performance hit, but I'm not keen on gambling that that will _always_ be the case for every widget type. People have constantly surprised me with how much complexity they're prepared to load into Streamfield, and the steady improvements to `bulk_to_python` (#5926, #5976, #6328) show that there's a desire to avoid individual lookup queries.

I hope this doesn't come across as obstructive, or over-engineering... I really do think the decisions we make here are going to go far beyond "tweaks to make react-streamfield work", and will determine our long-term approach to client-side form handling - probably the biggest unsolved problem that Wagtail faces today - so it's worth being cautious, and making sure that the approach we take is as robust and future-proof as possible.

BertrandBordage:
Thanks for the detailed reply @gasman!

I really like your idea, it would indeed be a clean way to have declarative system instead of our good old inline `<script>` that forced me to create `dangerouslyRunInnerScripts`in react-streamfield. It would be definitively an improvement.

To answer the first question: no, react-streamfield doesn’t fully comply with the rules I explained above, so if we wanted to nest StreamFields inside StreamFields (which is impossible, but equivalent to StreamBlocks inside StreamFields) the duplication wouldn’t work. But it would be trivial to make react-streamfield follow that rule. The only thing that is missing is when you run `$('[name="your-streamfield"]').val('[]');`, the value changed is ignored by react-streamfield. But actually, the Redux state of react-streamfield is just the deserialized input value. So there wouldn’t be much to do add, a `JSON.parse` and a value change listener.

But you are right, that feels a bit unfair because this declarative JS system is a refactoring improvement of Wagtail, it’s not directly because of react-streamfield, it is because react-streamfield’s improvements increase the limits of using inline `<script>`s. This was undocumented and already limiting in some ways (you have to know that you need to put a reference to the block ID in the `<script>`) before react-streamfield.
On the other hand, I agree that it’s a bit annoying for the rare third party JS blocks to have to adjust to react-streamfield then potentially adjust a second time later when we have a declarative JS system.

In a way I’m in favor of merging react-streamfield as is, at the expense of just a few third party JS blocks. Because it’s been a while, because it’s already a big improvement, and because I want to finally turn the page.
On the other hand I absolutely don’t want to force your hand. I see the huge long-term benefit of such a declarative system. It’s mandatory to keep gradually shifting Wagtail to a more dynamic editing admin.

So I want to help building that declarative system, if that’s the last thing we need to do to integrate react-streamfield. But I need help I think. Not so much technically, but to think, organize this and make it great.

If you want, we can discuss and build this system together, I would be very happy about that! Or if anyone wants to build that system for Wagtail, I would use it with pleasure in react-streamfield and remove `dangerouslyRunInnerScripts`.

gasman:
This was a deeper rabbit-hole than I expected, but I think I'm getting somewhere... https://github.com/gasman/telepath-poc

The hardest bit to solve was the 'bootstrapping' problem - how do you obtain a Javascript object describing the behaviour of widget X, when that behaviour is partly defined by Python-side configuration? - and it wouldn't surprise me if there's existing tech to deal with that problem, if only I knew what to look for... With that out of the way, it looks like the operations we need to support in the form field API are pretty simple - `render`, `getValue`, `getState` and `setState`. (There may be some additions to deal with the equivalent of `id_for_label` and rendering error messages, but nothing major.) Keen to hear your thoughts on whether it's feasible to retro-fit this into react-streamfield...

BertrandBordage:
Thank you very much Matt for pushing the concept as far :)
It's indeed not so simple, for multiple reasons.

The first blocker in my opinion (in the current state) is that telepath needs to be easily accessible from a NodeJS context in order to unpack data from react-streamfield. Which implies making an NPM distribution of telepath.

Another blocker that can be resolved is adjusting the react-streamfield API to allow creating for example a subclass of the main StreamField component that would set/get state. That being said, it's non trivial as we need to be able to load a state per StreamField and per block (for page choosers, for example).

And finally, the main part that doesn't seem right is this: “Output the resulting JSON on the template somewhere where Javascript code can access it (I suggest a data- attribute on an appropriate HTML element), along with js_context.media”
I know it's a common way to load data using jQuery, but it's considered as an anti-pattern in React. It's also not generic: sometimes we need to initialize a widget, but sometimes we want to update it via an AJAX request (for page choosers for example). I suggest we only pass IDs to widgets, then always load their state through AJAX requests (or websockets in the future). While that may sound awful and extremely slow (I didn't want to do that at first for wagtail-react-streamfield), it ends up being very fast and easier to maintain. There is also 2 ways ro optimize this: using conditional processing, and fetching multiple widget states at the same time.

All that being said, I know it's a heavy paradigm switch for Wagtail…

gasman:
In the context of StreamField, "Output the resulting JSON on the template somewhere where Javascript code can access it" probably means one JSON structure representing the entire StreamField configuration. I believe it fits neatly into what react-streamfield is already doing with the `Block.definition` property, so instead of doing:

    class NewFieldBlock(FieldBlock):
        @cached_property
        def definition(self):
            definition = super(FieldBlock, self).definition
            definition['html'] = self.render_form(self.get_default(), prefix=self.FIELD_NAME_TEMPLATE)

it would become something like:

    class NewFieldBlock(FieldBlock):
        def definition(self, js_context):
            definition = super(FieldBlock, self).definition
            definition['field'] = js_context.pack(self.field)

Your JS code can then call `telepath.unpack` on the `field` value to get back the Javascript object of the field, and then use that for the element construction and initialisation rather than working with HTML fragments and input `.value`s directly.

gasman:
Now completed as the telepath rewrite in 2.13.

PR #6227, opened 2020-07-15
Add duplicate icon from wagtail-react-streamfield

BertrandBordage:
In the effort to keep gradually integrating wagtail-react-streamfield into Wagtail, here is an icon needed by wagtail-react-streamfield. This is one of the tricky-to-maintain overrides done by wagtail-react-streamfield, that’s why it’s easier to integrate it earlier than the rest.


BertrandBordage:
Note that the icon is integrated both in the icons font & as SVG, to makes the transition as smooth as possible.

PR #6229, opened 2020-07-15
Serve PDFs inline in the browser and make content-type / content-disposition configurable

### LINKED ISSUE #1158: Any hook to set content-disposition in wagtail?
I want to serve some pdfs and I want them to be opened in the browser by default. I know in django development response['content-disposition'] does this trick, and I wonder if there is a hook in wagtail that has the same effect.

gasman:
Fixes #1158
Add config options WAGTAILDOCS_CONTENT_TYPES and WAGTAILDOCS_INLINE_CONTENT_TYPES to determine the Content-Type and Content-Disposition headers returned for documents when using the Django serve view, and default to application/pdf being served inline.

### DIFF
--- a/docs/reference/settings.rst
+++ b/docs/reference/settings.rst
@@ -180,6 +180,26 @@ For this reason, Wagtail provides a number of serving methods which trade some o
 
 If ``WAGTAILDOCS_SERVE_METHOD`` is unspecified or set to ``None``, the default method is ``'redirect'`` when a remote storage backend is in use (i.e. one that exposes a URL but not a local filesystem path), and ``'serve_view'`` otherwise. Finally, some storage backends may not expose a URL at all; in this case, serving will proceed as for ``'serve_view'``.
 
+.. _wagtaildocs_content_types:
+
+.. code-block:: python
+
+  WAGTAILDOCS_CONTENT_TYPES = {
+      'pdf': 'application/pdf',
+      'txt': 'text/plain',
+  }
+
+Specifies the MIME content type that will be returned for the given file extension, when using the ``serve_view`` method. Content types not listed here will be guessed using the Python ``mimetypes.guess_type`` function, or ``application/octet-stream`` if unsuccessful.
+
+.. _wagtaildocs_inline_content_types:
+
+.. code-block:: python
+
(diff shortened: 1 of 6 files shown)


lb-:
@gasman  I checked this out locally but I cannot seem to validate the changes manually.

Chrome
* Master branch - Will open public facing links in a tab with an inline preview and not download (seems to ignore the content disposition), checking out this PR and restarting the server makes no difference, it still opens inline (tab preview)
* Links within the documents listing in admin always trigger a download in either branch

Firefox
* Master or this branch - public facing links will never open in the preview (inline) mode and always download as a file
* Same in the documents listing - always triggers a download

I might be missing something when trying to validate. The code seems to make sense so maybe it is just the way different browsers honour this. Maybe it is because I am running the Bakery demo locally and the file serving is slightly different (e.g. from filesystem).

Also, is this change meant to affect the links in the Documents (admin) listing?

gasman:
@lb- I've retested, and all seems to work as expected, although browser caching does get in the way a bit - it helps to test with the developer tools open at the Network tab and "Disable Cache" ticked...

* from a clean install of bakerydemo, go to Documents -> Add a document and upload / save a PDF
* go to the edit view for that document (`/admin/documents/edit/1/`) and click the link against the "File:" label - this points to the `/documents/1/` serve view so should match the document links used on the front-end
* On both Firefox and Chrome, with the master branch checked out this gives a download prompt; with this PR branch checked out it opens the PDF in-browser.

Note that this change only applies when documents are served through Django itself, which is the default for local storage backends and can be enabled on other backends (S3 etc) by setting `WAGTAILDOCS_SERVE_METHOD = 'serve_view'`.

gasman:
@lb- Any chance you'd be able to take another look at this, please? I'm 90% sure that the issues you were seeing earlier were down to browser caching (which is less intuitive than normal here, since you can't force-refresh a URL that's being served as a download)...

lb-:
@gasman thanks heaps, this is working great, sorry it took be a bit to retest. I will merge this in shortly.

lb-:
Merged in via 8edf16e5ff03cb9aec25f15fa201c7345732d120

PR #6231, opened 2020-07-15
Add icon sprite

allcaps:
This PR moves the SVG icons from the body of each page to an SVG sprite.

- Sprite lives at `/admin/icon-[random string].svg`
- Has one year cache. Both view cache and browser cache.
- The random string is part of the filename. `#` is needed to reference the symbols by id.
- Random string changes when server restarts. This invalidates cache and makes easy development possible.
- In multi instance setups, each instance might have its own random string. And additional url pattern serves the sprite no matter the random string.
- Admin urls are never cached, but I'd wanted `/admin/sprite-[random].svg` to be cached. Fixed it by adding the url pattern after the `decorate_urlpatterns` `never_cache`.
- Pattern is added before `decorate_urlpatterns` `display_custom_404`. To prevent 404.
- Added svg4everybody to polyfill ie11.
- Icon template tag accepts a sprite argument. This was needed for the edit bird. There is no polyfill in the frontend, so symbols are in the body. So we do not need the sprite url here.
- JS can reference the correct sprite url via `wagtailConfig.ADMIN_URLS.SPRITE`

* Do the tests still pass? ✅
* For Python changes: Have you added tests to cover the new/fixed behaviour?❌ TODO
  - View test
     - 200
     - content type
     - cache header
  - URL test
    - Reverse gives pattern with random sting.
    - Calling multiple times gives same filename.
    - Match no matter the random string.

* For front-end changes: Did you test on all of Wagtail’s supported browsers? ✅

  - Mobile Safari | iOS Phone | Last 2✅  IOS 13.5 Safari 13.1
  - Mobile Safari | iOS Tablet | Last 2
  - Chrome | Android | Last 2
  - IE | Desktop | 11 ✅
  - Chrome | Desktop | Last 2 ✅  83.0.4103.116
  - MS Edge | Desktop | Last 2
  - Firefox | Desktop | Latest ❌  78.0.2   Icons flash in admin on load. Edit bird is okay.
  - Firefox ESR | Desktop | Latest❌  68.10.0esr   Icons flash in admin on load. Edit bird is okay.
  - Safari | macOS | Last 2 ✅ 13.1.1

Finally 
- I saw that the Page menu item chevron flashes in _all_ browsers. That has nothing to do with this PR. Probably a JS thing.
- I need some helpt to debug the Firefox flash.

### DIFF
--- a/client/src/components/Icon/Icon.js
+++ b/client/src/components/Icon/Icon.js
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types';
 import React from 'react';
+import { ADMIN_URLS } from '../../config/wagtailConfig';
 
 /**
  * Provide a `title` as an accessible label intended for screen readers.
@@ -7,7 +8,7 @@ import React from 'react';
 const Icon = ({ name, className, title }) => (
   <span>
     <svg className={`icon icon-${name} ${className || ''}`} aria-hidden="true">
-      <use href={`#icon-${name}`}></use>
+      <use href={`${ADMIN_URLS.SPRITE}#icon-${name}`}></use>
     </svg>
     {title ? (
       <span className="visuallyhidden">
--- a/client/src/utils/polyfills.js
+++ b/client/src/utils/polyfills.js
@@ -8,3 +8,10 @@ import 'core-js/shim';
 import 'whatwg-fetch';
 // IE11.
 import 'element-closest';
+// IE11.
+import { IS_IE11 } from '../config/wagtailConfig';
+import svg4everybody from 'svg4everybody';
+
+if (IS_IE11) {
+  svg4everybody();
+}
--- a/wagtail/admin/templates/wagtailadmin/shared/icon.html
+++ b/wagtail/admin/templates/wagtailadmin/shared/icon.html
@@ -2,7 +2,7 @@
 {# Reference implementation: components/Icon.js #}
 {% if wrapped %}<span class="icon-wrapper">{% endif %}
 <svg class="icon icon-{{ name }} {{ class_name }}" aria-hidden="true" focusable="false">
-  <use href="#icon-{{ name }}"></use>
+  <use href="{{ sprite }}#icon-{{ name }}"></use>
 </svg>
 {% if title %}
   <span class="visuallyhidden">
--- a/wagtail/admin/templates/wagtailadmin/skeleton.html
+++ b/wagtail/admin/templates/wagtailadmin/skeleton.html
@@ -13,6 +13,7 @@
     <script src="{% versioned_static 'wagtailadmin/js/vendor/modernizr-2.6.2.min.js' %}"></script>
 
     <link rel="stylesheet" href="{% versioned_static 'wagtailadmin/css/normalize.css' %}" />
+    <link rel="preload" href="{% url 'wagtailadmin_sprite' %}">
 
     {% block css %}{% endblock %}
 
@@ -28,8 +29,6 @@
             Here are the <a href="https://www.enable-javascript.com/" target="_blank" rel="noopener noreferrer">instructions how to enable JavaScript in your web browser</a>.
         {% endblocktrans %}
     </noscript>
-
-    {% icons %}
 
     {% block js %}{% endblock %}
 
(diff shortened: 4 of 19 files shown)


client/src/utils/polyfills.js
```diff
@@ -8,3 +8,7 @@ import 'core-js/shim';
 import 'whatwg-fetch';
 // IE11.
 import 'element-closest';
+// IE11.
+import svg4everybody from 'svg4everybody';
+
+svg4everybody();
```
thibaudcolas, line 14: Should we wrap this in our `IS_IE11`? https://github.com/wagtail/wagtail/blob/882f8f3cf8ddd79c30e611a48882b309e90dad0c/client/src/config/wagtailConfig.js

wagtail/admin/views/home.py
```diff
@@ -144,3 +146,15 @@ def default(request):
     redirected to the login page.
     """
     raise Http404
+
+
+@cache_page(60 * 60 * 24 * 365)  # seconds * minutes * hours * days
+def sprite(request):
+    response = render(request, "wagtailadmin/icons/sprite.svg")
+    response["Content-Type"] = "image/svg+xml"
+    # TODO: We do not want to set cache in debug mode.
```
thibaudcolas, line 155: I think we do, if possible? I imagine it’s going to be really annoying for Wagtail devs to see the icons flash all the time.

client/src/components/Icon/Icon.js
```diff
@@ -7,7 +7,8 @@ import React from 'react';
 const Icon = ({ name, className, title }) => (
   <span>
     <svg className={`icon icon-${name} ${className || ''}`} aria-hidden="true">
-      <use href={`#icon-${name}`}></use>
+      {/* TODO: Should not hardcode the admin path ... */}
```
thibaudcolas, line 10: 😄  a new attribute for https://github.com/wagtail/wagtail/blob/master/wagtail/admin/templates/wagtailadmin/admin_base.html#L32?

wagtail/admin/templates/wagtailadmin/skeleton.html
```diff
@@ -13,6 +13,7 @@
     <script src="{% versioned_static 'wagtailadmin/js/vendor/modernizr-2.6.2.min.js' %}"></script>
 
     <link rel="stylesheet" href="{% versioned_static 'wagtailadmin/css/normalize.css' %}" />
+    <link rel="preload" href="{% url 'wagtailadmin_sprite' %}">
```
allcaps, line 16: @thibaudcolas 

- This prevents a flash when in development the sprite isn't cached. 
- It also removes the initial flash.


But we still need

wagtail/admin/views/home.py
```diff
@@ -144,3 +146,12 @@ def default(request):
...
     """
     raise Http404
+
+
+@cache_page(60 * 60 * 24 * 365)  # seconds * minutes * hours * days
+def sprite(request):
+    response = render(request, "wagtailadmin/icons/sprite.svg")
+    response["Content-Type"] = "image/svg+xml"
+    if not settings.DEBUG:
+        response["Cache-Control"] = f"max-age={60 * 60 * 1}"  # seconds * minutes * hours
```
allcaps, line 156: Normal assets are under webserver cache policy (if configured). But this is a Django view, default without browser cache.

I experimented with infinite browser cache and invalidating the sprite with a hash in the filename. I think that will lead to many problems. Especially with multi worker setups. Each instance need to use the same hash. Keeping that in sync is a pain.

We need to update the filename `sprite-[hash].svg` as `#` is used to reference the symbols by id.

The sprite is only 80 kilobytes. But the whole idea of a sprite is to stop bytes to go over the wire each request.

So I choose one hour browser cache in production mode. Is that sensible?

wagtail/admin/templates/wagtailadmin/shared/icon.html
```diff
@@ -1,7 +1,7 @@
 {% spaceless %}
 {# Reference implementation: components/Icon.js #}
 <svg class="icon icon-{{ name }} {{ class_name }}" aria-hidden="true" focusable="false">
-  <use href="#icon-{{ name }}"></use>
+  <use href="{% url 'wagtailadmin_sprite' %}#icon-{{ name }}"></use>
```
allcaps, line 4: Check if the IE11 polyfill isn't loaded in `wagtail-userbar-nav`.

allcaps:
Superseded by https://github.com/wagtail/wagtail/pull/6243

PR #6233, opened 2020-07-16
Workflow docs updates

jacobtoppm:
Updates the reference and custom tasks documentation for consistency with the new models

Also updates the purge_revisions management command to not purge revisions attached to NEEDS_CHANGES workflows (as this state is non terminal)


gasman:
LGTM - merged in fa84f0c2ebc941a3b111765ce157988bdfb10a4d.

PR #6234, opened 2020-07-16
Replace page explorer pushPage/popPage with gotoPage

kaedroho:
The current actions only allow navigating into children and back up to the parent. This PR replaces the existing actions with one that allows navigating to any part of the tree.

This is a prerequisite for the "locale" filter that I am implementing in the explorer (see https://github.com/wagtail/rfcs/blob/internationalisation/text/054-internationalisation.md#page-explorer).

I've also removed "path" from the reducer and replaced this with "depth" and "page". Page nodes now have a "parent" attribute which we use for navigating back up the tree

 - "page" is the ID of the current page
 - "depth" is an integer, when you navigate down it's incremented, when you navigate up it's decremented. This is used by the `ExplorerPanel` to work on which direction to run the animation. The absolute value doesn't mean anything; we only care about how much this value changes between page transitions.

### DIFF
--- a/client/src/components/Explorer/ExplorerHeader.js
+++ b/client/src/components/Explorer/ExplorerHeader.js
@@ -9,32 +9,27 @@ import Icon from '../../components/Icon/Icon';
  * The bar at the top of the explorer, displaying the current level
  * and allowing access back to the parent level.
  */
-const ExplorerHeader = ({ page, depth, onClick }) => {
-  const isRoot = depth === 1;
-
-  return (
-    <Button
-      href={page.id ? `${ADMIN_URLS.PAGES}${page.id}/` : ADMIN_URLS.PAGES}
-      className="c-explorer__header"
-      onClick={onClick}
-    >
-      <div className="c-explorer__header__inner">
-        <Icon
-          name={isRoot ? 'home' : 'arrow-left'}
-          className="icon--explorer-header"
-        />
-        <span>{page.data && page.data.admin_display_title || STRINGS.PAGES}</span>
-      </div>
-    </Button>
-  );
-};
+const ExplorerHeader = ({ page, onClick }) => (
+  <Button
+    href={page.id ? `${ADMIN_URLS.PAGES}${page.id}/` : ADMIN_URLS.PAGES}
+    className="c-explorer__header"
+    onClick={onClick}
+  >
+    <div className="c-explorer__header__inner">
+      <Icon
+        name={page.parent ? 'arrow-left' : 'home'}
+        className="icon--explorer-header"
+      />
+      <span>{page.data && page.data.admin_display_title || STRINGS.PAGES}</span>
+    </div>
+  </Button>
+);
 
 ExplorerHeader.propTypes = {
   page: PropTypes.shape({
     id: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
     admin_display_title: PropTypes.string,
   }).isRequired,
-  depth: PropTypes.number.isRequired,
   onClick: PropTypes.func.isRequired,
 };
 
--- a/client/src/components/Explorer/ExplorerHeader.js
+++ b/client/src/components/Explorer/ExplorerHeader.js
@@ -11,7 +11,7 @@ import Icon from '../../components/Icon/Icon';
  */
 const ExplorerHeader = ({ page, onClick }) => (
   <Button
-    href={page.id ? `${ADMIN_URLS.PAGES}${page.id}/` : ADMIN_URLS.PAGES}
+    href={page.data.id ? `${ADMIN_URLS.PAGES}${page.data.id}/` : ADMIN_URLS.PAGES}
     className="c-explorer__header"
     onClick={onClick}
   >
--- a/client/src/components/Explorer/ExplorerHeader.js
+++ b/client/src/components/Explorer/ExplorerHeader.js
@@ -23,7 +23,7 @@ const ExplorerHeader = ({ page, depth, onClick }) => {
           name={isRoot ? 'home' : 'arrow-left'}
           className="icon--explorer-header"
         />
-        <span>{page.admin_display_title || STRINGS.PAGES}</span>
+        <span>{page.data && page.data.admin_display_title || STRINGS.PAGES}</span>
       </div>
     </Button>
   );
--- a/client/src/components/Explorer/ExplorerPanel.js
+++ b/client/src/components/Explorer/ExplorerPanel.js
@@ -30,8 +30,8 @@ class ExplorerPanel extends React.Component {
   }
 
   componentWillReceiveProps(newProps) {
-    const { path } = this.props;
-    const isPush = newProps.path.length > path.length;
+    const { depth } = this.props;
+    const isPush = newProps.depth > depth;
 
     this.setState({
       transition: isPush ? PUSH : POP,
@@ -70,23 +70,22 @@ class ExplorerPanel extends React.Component {
   }
 
   onItemClick(id, e) {
-    const { pushPage } = this.props;
+    const { gotoPage } = this.props;
 
     e.preventDefault();
     e.stopPropagation();
 
-    pushPage(id);
+    gotoPage(id, 1);
   }
 
   onHeaderClick(e) {
-    const { path, popPage } = this.props;
-    const hasBack = path.length > 1;
+    const { page, gotoPage } = this.props;
 
-    if (hasBack) {
+    if (page.parent) {
       e.preventDefault();
       e.stopPropagation();
 
-      popPage();
+      gotoPage(page.parent, -1);
     }
   }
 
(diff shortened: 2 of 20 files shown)


client/src/components/Explorer/reducers/nodes.js
```diff
@@ -19,8 +18,10 @@ const node = (state = defaultPageState, { type, payload }) => {
     return state || defaultPageState;
 
   case 'GET_PAGE_SUCCESS':
-    return Object.assign({}, state, payload.data, {
+    return Object.assign({}, state, {
       isError: false,
+      parent: (payload.data.meta.parent && payload.data.meta.parent.id) || null,
+      data: payload.data,
```
thibaudcolas, line 24: It looks like `parent` is just derived from information that was already accessible in the reduced state. Could we avoid adding this computed data here, and instead retrieve the info as needed where it is needed?

client/src/components/Explorer/ExplorerHeader.js
```diff
@@ -9,32 +9,27 @@ import Icon from '../../components/Icon/Icon';
...
-};
+const ExplorerHeader = ({ page, onClick }) => (
+  <Button
+    href={page.id ? `${ADMIN_URLS.PAGES}${page.id}/` : ADMIN_URLS.PAGES}
+    className="c-explorer__header"
+    onClick={onClick}
+  >
+    <div className="c-explorer__header__inner">
+      <Icon
+        name={page.parent ? 'arrow-left' : 'home'}
```
thibaudcolas, line 20: Are there any implications to this for cases where the user only has access to some sub-pages in the site’s page tree?

thibaudcolas (changes requested):
I’ve had a brief look, CI is complaining about broken unit tests and I think you likely will have to update, add or remove unit tests as part of this PR.

The animation for pop and push is pretty clear – can you explain what animation we’re expecting when navigating to another page that’s at the same level as the current page?

kaedroho:
>The animation for pop and push is pretty clear – can you explain what animation we’re expecting when navigating to another page that’s at the same level as the current page?

This currently just loads the new results and sets the without any animation

client/src/components/Explorer/ExplorerHeader.js
kaedroho: Yep! IIRC, this is handled by the admin API (but I might be misremembering some other work I did in https://github.com/wagtail/wagtail/pull/5679). Been so long since I worked on any of that so will have to check

kaedroho:
@thibaudcolas I've updated all the snapshots now, so the tests are passing again

client/src/components/Explorer/reducers/nodes.js
kaedroho: Mind if I leave this one until after the move to TypeScript?

client/src/components/Explorer/ExplorerPanel.test.js
```diff
@@ -72,38 +72,40 @@ describe('ExplorerPanel', () => {
...
+      expect(mockProps.gotoPage).toHaveBeenCalled();
     });
 
-    it('does not call popPage for first page', () => {
+    it('does not call gotoPage for first page', () => {
+      mockProps.page.parent = null;
+
       shallow((
-        <ExplorerPanel {...mockProps} path={[1]} />
+        <ExplorerPanel {...mockProps} depth={1} />
```
thibaudcolas, line 93: `depth` is already set to 1 in mockProps, so what I think you need here is just to override the `page` prop so it has no `parent`.

client/src/components/Explorer/__snapshots__/ExplorerHeader.test.js.snap
```diff
@@ -46,7 +20,7 @@ exports[`ExplorerHeader #page 1`] = `
       title={null}
     />
     <span>
-      test
+      Pages
```
thibaudcolas, line 23: This looks like a bug, but most likely only because the test case needs updating?

client/src/components/Explorer/actions.test.js
```diff
@@ -65,32 +65,22 @@ describe('actions', () => {
...
-  describe('pushPage', () => {
-    it('exists', () => {
-      expect(actions.pushPage).toBeDefined();
+      expect(actions.gotoPage).toBeDefined();
     });
 
     it('creates action', () => {
       const store = mockStore(stubState);
-      store.dispatch(actions.pushPage(5));
+      store.dispatch(actions.gotoPage(5));
```
thibaudcolas, line 75: Here and below – it looks like `gotoPage`’s transition argument needs to be defined?

client/src/components/Explorer/reducers/explorer.test.js
```diff
@@ -20,13 +20,7 @@ describe('explorer', () => {
     expect(explorer(initialState, { type: 'CLOSE_EXPLORER' })).toEqual(initialState);
   });
 
-  it('PUSH_PAGE', () => {
+  it('GOTO_PAGE', () => {
     expect(explorer(initialState, { type: 'PUSH_PAGE', payload: { id: 100 } })).toMatchSnapshot();
```
thibaudcolas, line 24:
```suggestion
    expect(explorer(initialState, { type: 'GOTO_PAGE', payload: { id: 100 } })).toMatchSnapshot();
```

You’ll also need to add the `transition` I assume.

client/src/components/Explorer/reducers/nodes.js
thibaudcolas: Hmm, I’m not sure. If you want to reduce refactoring work I would suggest to:

- Keep existing components using `pushPage` and `popPage` action creators, just redefine them to be based on the `GOTO_PAGE` action under the hood.
- Stop changing the shape of this `nodes` reducer – there is no need for the extra `data` structure if we just access the parent data from `page.meta.parent && page.meta.parent.id`.
- Could be even tidier by upgrading to a recent version of Babel so we could do `page.meta.parent?.id`

client/src/components/Explorer/ExplorerHeader.js
thibaudcolas: It looks like the admin API returns the details of the `parent` page even if the user has no permissions on it, so this no longer works as it used to:

blogger only

This is with a user who only has access to the Blog page in the bakerydemo. Ideally it should show the Home icon, and more importantly this shouldn’t be clickable (causes a JS error currently). I think this might be fixable simply enough by basing this on `depth`, so it’s relative to the user’s navigation, wherever it happens to start from.

client/src/components/Explorer/ExplorerPanel.js
```diff
@@ -70,23 +70,22 @@ class ExplorerPanel extends React.Component {
...
+    gotoPage(id, 1);
   }
 
   onHeaderClick(e) {
-    const { path, popPage } = this.props;
-    const hasBack = path.length > 1;
+    const { page, gotoPage } = this.props;
 
-    if (hasBack) {
+    if (page.parent) {
```
thibaudcolas, line 84: Related to my comment in `ExplorerHeader` – this check isn’t equivalent to the previous one, as pages have a defined `parent` even if the current user technically doesn’t have permissions over that parent.

Here, this means clicking the header causes a JS error (trying to load the details of a parent page the user has no access to) instead of opening the link to the user’s top-level page. I imagine we could fix it by having this be based on `depth` rather than `page.parent`?

thibaudcolas (changes requested):
@kaedroho I think this is pretty close, there are still some changes to do in the tests, and a bug to fix for users who only have access to a sub-tree of the site.

For the tests – I agree they feel overkill (particularly the Redux ones), and it would probably be better to rely less on snapshots (or use inline snapshots if we had Prettier set up). I’m open to suggestions on which ones we should can, but for now I feel like it’s better for us to have those tedious tests than no tests at all. Integration tests would be the best for this explorer menu, it’s very easy to test the back / forward functionality with browser automation, but we don’t have that set up.

client/src/components/Explorer/ExplorerHeader.js
kaedroho: Ahh, I think I changed this logic in https://github.com/wagtail/wagtail/pull/5679 but forgot that isn't merged yet. Thanks for spotting!

client/src/components/Explorer/reducers/nodes.js
kaedroho: > Stop changing the shape of this nodes reducer – there is no need for the extra data structure if we just access the parent data from page.meta.parent && 

While implementing https://github.com/wagtail/wagtail/pull/6474 I was reminded why I changed this. The nodes reducer mixes content from the API with local state. If we added a field to the API that clashed with a local one, weird stuff might happen.

kaedroho: > Keep existing components using pushPage and popPage action creators, just redefine them to be based on the GOTO_PAGE action under the hood.

One issue with this is popPage doesn't have any arguments so there's no way of knowing what the destination page should be.

In the last, this was done by tracking all ancestors, but this gets tricky to do when we allow people to jump to any postition in the tree.

lb-:
Superseded by #6478 /  f0ce1f4775927b1fe38042e756e880e753a0baf1

kaedroho:
Thanks!

PR #6241, opened 2020-07-19
Fix Python 3.9 Deprecation Warning

loicteixeira:
While working on wagtail-bakery, I came across a deprecation warning for Python 3.9 on Wagtail itself:

> DeprecationWarning: Using or importing the ABCs from 'collections'
> instead of from 'collections.abc' is deprecated since Python 3.3,
> and in 3.9 it will stop working

Since we only support Python 3.5+, it there shouldn't be any compatibility issue.

I haven't checked if there are other deprecation warnings just yet.


gasman:
Thanks @loicteixeira!

PR #6243, opened 2020-07-19
Add icon sprite via local storage

allcaps:
This PR stores the sprite in local storage and injects it into the page body. 

A previous attempt has an external sprite with IE11 polyfill. https://github.com/wagtail/wagtail/pull/6231 
Unfortunately, it leads to flashing icons in Firefox. :/

This PR:
- Inspired by https://osvaldas.info/caching-svg-sprite-in-localstorage
- Uses local storage
- Injects the sprite into the body
- The filename contains a hash to invalidate local storage.
- The hash is based on the sprite contents and secret key (inspired by versioned_static).
- No polyfill at all 🎉
- No flash on load 🎉

Checks:
* Do the tests still pass? ✅
* For front-end changes: Did you test on all of Wagtail’s supported browsers? ✅
  - ✅ IE 11
  - ✅ Edge 83
  - ✅ Chrome 84.0.4147.89
  - ✅ Firefox 78.0.2
  - ✅  Firefox 68.10.0esr
  - ✅ Safari 13.1.2
  - ✅ Safari 13.1 iPad


wagtail/admin/static_src/wagtailadmin/js/core.js
```diff
@@ -83,6 +83,57 @@ function enableDirtyFormCheck(formSelector, options) {
...
+                    insert();
+                    if (isLocalStorage) {
+                        localStorage.setItem(dataKey, data);
+                        localStorage.setItem(revisionKey, spriteURL);
+                    }
+                }
+            }
+            request.send();
+        } catch (e) {
+        }
```
Scotchester, line 132: Is there anything we can do if the XHR fails to provide a fallback or a notice, or are we good with just ignoring it?

wagtail/admin/static_src/wagtailadmin/js/core.js
```diff
@@ -83,6 +83,57 @@ function enableDirtyFormCheck(formSelector, options) {
...
+    // Load icon sprite and cache it in local storage.
+    function loadIconSprite() {
+        var spriteURL = wagtailConfig.ADMIN_URLS.SPRITE;
+        var revisionKey = 'spriteRevision';
+        var dataKey = 'spriteData';
+
+        // Feature check
+        if (!document.createElementNS || !document.createElementNS('http://www.w3.org/2000/svg', 'svg').createSVGRect) {
+            return true;
+        }
```
allcaps, line 95: Remove, all Wagtail supported browsers support SVG anyway.

thibaudcolas, line 95: It looks like we don’t need this, since all our target browsers support SVG (https://caniuse.com/#search=createSVGRect).

wagtail/admin/static_src/wagtailadmin/js/core.js
```diff
@@ -83,6 +83,57 @@ function enableDirtyFormCheck(formSelector, options) {
...
+
+
+        var isLocalStorage = 'localStorage' in window && window.localStorage !== null;
```
thibaudcolas, line 98: I would be more comfortable with this being:

```suggestion
        var isLocalStorage = 'localStorage' in window && typeof window.localStorage !== 'undefined';
```

… so this works in more cases. I’m not too familiar with what `window.localStorage` would register as for browsers that don’t have it, so that check seems more comprehensive.

wagtail/admin/static_src/wagtailadmin/js/core.js
```diff
@@ -83,6 +83,57 @@ function enableDirtyFormCheck(formSelector, options) {
...
+        var request;
```
thibaudcolas, line 99: I don’t think this needs to be defined here, can be defined below only.

```suggestion
```

wagtail/admin/static_src/wagtailadmin/js/core.js
```diff
@@ -83,6 +83,57 @@ function enableDirtyFormCheck(formSelector, options) {
...
+        if (isLocalStorage && localStorage.getItem(revisionKey) === spriteURL) {
+            data = localStorage.getItem(dataKey);
+            if (data) {
+                insert();
+                return true;
+            }
+        }
+
+        try {
+            request = new XMLHttpRequest();
```
thibaudcolas, line 118:
```suggestion
            var request = new XMLHttpRequest();
```

wagtail/admin/static_src/wagtailadmin/js/core.js
```diff
@@ -83,6 +83,57 @@ function enableDirtyFormCheck(formSelector, options) {
...
+        // Feature check
+        if (!document.createElementNS || !document.createElementNS('http://www.w3.org/2000/svg', 'svg').createSVGRect) {
+            return true;
+        }
+
+
+        var isLocalStorage = 'localStorage' in window && window.localStorage !== null;
+        var request;
+        var data;
+        var insertIt = function () {
```
thibaudcolas, line 101:
```suggestion
        var insertIt = function (data) {
```

wagtail/admin/static_src/wagtailadmin/js/core.js
```diff
@@ -83,6 +83,57 @@ function enableDirtyFormCheck(formSelector, options) {
...
+            document.body.insertAdjacentHTML('afterbegin', data);
+        }
+        var insert = function () {
+            if (document.body) insertIt();
+            else document.addEventListener('DOMContentLoaded', insertIt);
```
thibaudcolas, line 106:
```suggestion
        var insert = function (data) {
            if (document.body) {
                insertIt(data) 
            };
            else {
                document.addEventListener('DOMContentLoaded', insertIt.bind(null, data));
            }
```

wagtail/admin/static_src/wagtailadmin/js/core.js
```diff
@@ -83,6 +83,57 @@ function enableDirtyFormCheck(formSelector, options) {
...
+        }
+
+        if (isLocalStorage && localStorage.getItem(revisionKey) === spriteURL) {
+            data = localStorage.getItem(dataKey);
+            if (data) {
+                insert();
```
thibaudcolas, line 112:
```suggestion
                insert(data);
```

wagtail/admin/static_src/wagtailadmin/js/core.js
thibaudcolas: I would suggest `console.error(e);`

gasman: What kind of errors are we anticipating here to require the try/catch? Since XHR is asynchronous, any HTTP errors resulting from the sprite fetch will happen after we've already left this try/catch block - all that will happen in that case is that the `request.onload` callback will fail to trigger. (If we need to catch those errors - and I'm not sure we do - we'd need a `request.onerror`.)

wagtail/admin/urls/__init__.py
```diff
@@ -82,9 +86,23 @@
...
 
+sprite_hash = None
+
+
+def get_sprite_hash():
+    global sprite_hash
+    if not sprite_hash:
+        sprite = render_to_string("wagtailadmin/shared/icons.html", icons())
+        sprite_hash = hashlib.sha1(
+            (sprite + settings.SECRET_KEY).encode("utf-8")
```
gasman, line 97: What's the purpose of SECRET_KEY here?

allcaps: I did the same as in https://github.com/wagtail/wagtail/blob/33e4c0ec7b889131740ee66a954f4ca2293fc2dd/wagtail/admin/staticfiles.py#L31-L36

I assume this is to keep the Wagtail version a secret.

allcaps:
All feedback is applied.

I moved the script into the page body. This way it is executed on load. All icons are directly available for paint.
This removes the annoying flash in Firefox! 🎉

What do we do with the catched error? Not displaying icons is an obvious fail. I propose we don't handle the error at all.

thibaudcolas:
I’ll try and review this over the next two weeks. @allcaps is there anything in particular you know might not be ready to go?

allcaps:
Ready!

wagtail/admin/views/home.py
```diff
@@ -144,3 +146,10 @@ def default(request):
     redirected to the login page.
     """
     raise Http404
+
+
+@cache_page(31536000)  # 1 year
```
kaedroho, line 151: To cache in browsers, we should use the [`cache_control`](https://docs.djangoproject.com/en/3.1/topics/http/decorators/#django.views.decorators.cache.cache_control) header. This one is for caching server-side.

wagtail/admin/views/home.py
```diff
@@ -144,3 +146,10 @@ def default(request):
...
+def sprite(request, **kwargs):
+    response = render(request, "wagtailadmin/icons/sprite.svg")
+    response["Content-Type"] = "image/svg+xml"
```
kaedroho, line 154: Since we're embedding this onto the page with .innerHTML, I think we should change this MIME type to text/html, or change that embed code to use an  tag instead. This makes it clear what the format is since HTML has some extra security concerns that we don't usually need to consider for SVG.

wagtail/admin/views/home.py
allcaps: Yes, it is intended to be server-side. We don't want to render that template each request. 
It loads icons template tag which triggered hooks. 

Matt already experienced the performance hit (currently the same code runs for adding the sprite to the base template). He added this:
https://github.com/wagtail/wagtail/blob/master/wagtail/admin/templatetags/wagtailadmin_tags.py#L551-L562

And that makes this sever-side cache redundant. So I'll drop it.

wagtail/admin/views/home.py
allcaps: Good catch! I believe this is an artefact of earlier attempts. I tried to implement SVG as an external file.

I will drop this line. A Django view default type is `text/html`.

wagtail/admin/views/home.py
kaedroho: Oh, I see, thanks!

wagtail/admin/views/home.py
kaedroho: Thanks! It might also be worth changing the template name from `.svg` to `.html` to make it clear there too

allcaps:
I rebased and simplified the code a bit. I dropped the template tag. Now the view itself is cached. This is less code and easier to understand.

allcaps:
Ready again :D

wagtail/admin/urls/__init__.py
```diff
@@ -84,9 +86,23 @@
...
+        content = str(home.sprite(None).content, "utf-8")
+        sprite_hash = hashlib.sha1(
+            (content + settings.SECRET_KEY).encode("utf-8")
+        ).hexdigest()[:8]
+    return sprite_hash
+
 
 # These url patterns do not require an authenticated admin user
 urlpatterns += [
+    path(f"sprite-{get_sprite_hash()}.svg", home.sprite, name="wagtailadmin_sprite"),
```
thibaudcolas, line 104: This ends in `.svg` but is technically HTML. Should we switch to `.html`?

wagtail/admin/templates/wagtailadmin/skeleton.html
```diff
@@ -19,6 +19,54 @@
...
+                        data = request.responseText;
+                        insert(data);
+                        if (isLocalStorage) {
+                            localStorage.setItem(dataKey, data);
+                            localStorage.setItem(revisionKey, spriteURL);
+                        }
+                    }
+                }
+                request.send();
+            } catch (e) {}
```
thibaudcolas, line 65: Could be worth a `console.error(e)` so as not to completely swallow JS bugs with the code?

```suggestion
            } catch (e) {
                console.error(e);
            }
```

wagtail/admin/templates/wagtailadmin/skeleton.html
```diff
@@ -19,6 +19,54 @@
     {% block branding_favicon %}{% endblock %}
 </head>
 <body id="wagtail" class="{% block bodyclass %}{% endblock %} {% if messages %}has-messages{% endif %} focus-outline-on">
+    <div data-sprite></div>
+    <script>
+        function loadIconSprite() {
+            var spriteURL = '{% url "wagtailadmin_sprite" %}';
+            var revisionKey = 'spriteRevision';
+            var dataKey = 'spriteData';
```
thibaudcolas, line 27: I could see this potentially colliding with Wagtail sites’ code since Wagtail is generally served from the same domain as sites (at least in development). I think it would be worth prefixing both of those keys with `wagtail:` or similar, so they are less likely to be un-intentionally shared?

I could see myself stealing this icon sprites implementation for other Wagtail-based projects 😉 .

```suggestion
            var revisionKey = 'wagtail:spriteRevision';
            var dataKey = 'wagtail:spriteData';
```

thibaudcolas (changes requested):
@allcaps this is looking really good! The icons are missing for a small amount of time on the first page load, but this is similar to an icon font’s performance, and for subsequent page loads the SVG icons seem to load faster than the icon fonts’.

I’ve suggested a few more changes but they are very minor.

For future reference, I tried to improve the first load performance with `<link rel="preload" href="{% url 'wagtailadmin_sprite' %}" as="fetch">` towards the top of `skeleton.html`. This works as expected, however it has the drawback of always being there, even after the first load, so feels very wasteful for a one-off performance gain. I don’t think this is worthwhile.

thibaudcolas (approved):
Wooo!

thibaudcolas:
Rebased, merged, and ready for release in 2.11! Thank you @allcaps, it looks like a lot of thought went into this, and it’s working perfectly as far as I can tell. I haven’t done any thorough benchmarking but this looks like it will improve first page load performance to be as good as the best external sprites implementation, while subsequent page loads will be as fast as the best inline sprites implementations – all the while preserving compatibility with IE11. Cool stuff :)

PR #6244, opened 2020-07-20
Replace unidecode with anyascii in wagtail.core.utils.string_to_ascii

rjmackay:
- Remove unidecode because it is licensed under GPL and replace with anyascii which has a compatible license.
- Update wagtail.core.utils.string_to_ascii to use anyascii.
- Anyascii has a similar but not exactly the same encoding - see updates to tests.

Refs https://github.com/wagtail/wagtail/issues/3311

  - I've updated the existing tests


setup.py
```diff
@@ -31,12 +31,12 @@
     "Pillow>=4.0.0,<8.0.0",
     "beautifulsoup4>=4.8,<4.9",
     "html5lib>=0.999,<2",
-    "Unidecode>=0.04.14,<2.0",
```
rjmackay, line 34: I'm not sure if this was correct or if I should leave unidecode in the requirements till a future release can remove it?

It doesn't appear to be required but rather it now an optional dependency you could install to use AbstractFormField `_migrate_legacy_clean_name`.

gasman:
Looks good to me - thanks @rjmackay! Merged in 437e3b9258c87229498c43a70b43ea8739eca988 now - I've left unidecode in place as a dependency but added a note to remove it for 2.12 (since by then developers will have had enough time to deploy 2.10 / 2.11 and have their old form data migrated).

lb-:
Awesome news, great to see this finished off. Thanks @gasman  & @rjmackay

PR #6220, opened 2020-07-13
[RFC 54] Internationalisation models

kaedroho:
This PR implements the models as specified in [RFC 54](https://github.com/wagtail/rfcs/pull/54). This excludes any UI improvements, but these are not required for a basic implementation and wouldn't matter if they had to be merged into a future release.

This is sponsored by the Mozilla Foundation.

TODO:

 - [x] Get it working with existing tests
 - [x] Add new tests
   - [x] `Locale` model
   - [x] `TranslatableMixin`
   - [x] `copy_for_translation`
   - [x] `get_translatable_models`
   - [x] Default locale
   - [x] `find_available_slug`
   - [x] `can_move_to` logic
   - [x] `.localized`
   - [x] Routing
   - [x] Auto-localization of rich text links
   - [x] Default locale set on page create
   - [x] `WAGTAIL_CONTENT_LANGUAGES`
 - [x] Find a new way to do routing
 - [x] `WAGTAIL_CONTENT_LANGUAGES` setting

Future PRs:

 - Documentation #6377
 - API changes - #6222
 - Search changes
 - Sitemaps changes
 - UI changes - #6230
 - Locale management UI - https://github.com/wagtail/wagtail/pull/6245
 - Logging support - https://github.com/wagtail/wagtail/pull/6262

PR #6245, opened 2020-07-20
[RFC 54] Locale management UI

kaedroho:
Builds on #6220


wagtail/locales/templates/wagtaillocales/create.html
```diff
@@ -0,0 +1,6 @@
+{% extends "wagtailadmin/generic/create.html" %}
+
+{% block extra_js %}
+    {{ block.super }}
+    {% include "wagtailadmin/pages/_editor_js.html" %}
```
gasman, line 5: Is this needed?

gasman (changes requested):
Looks pretty good - just a minor quibble about the use of `_editor_js.html` (and imports need fixing to placate isort).

wagtail/locales/templates/wagtaillocales/create.html
kaedroho: nope, removed!

gasman:
Looks like `wagtail.locales.utils.get_locale_usage` isn't playing well with [the one-off invalid model created in `TestSystemChecks`](https://github.com/wagtail/wagtail/blob/738003726900295328f263bfe91686026f70f596/wagtail/core/tests/test_translatablemixin.py#L142) - once that test has run, the model is returned in `get_translatable_models`, causing the subsequent query to fail as it doesn't have a table.

I'm thinking the best way around that is going to be either: find a way to un-register the invalid model after use (which looks a bit messy, involving fiddling around in internal properties of `AppConfig` and clearing the lru_cache on `apps.get_models`), or tweaking the test so that it temporarily modifies a real model from testapp.models rather than creating a new one.

kaedroho:
@gasman thanks! That issue is fixed now

PR #6250, opened 2020-07-21
Postgres search fixes

kaedroho:
This PR contains a couple of fixes for the Postgres search backend. These issues were caused by recent improvements to ranking.

- No longer rewrites the whole table when new entries are added
- No longer catches indexing errors for the Postgres backend (because when it does, Django crashes anyway)


gasman:
Merged in b8848ae2f5b2594c2162cee2d76ef92fcc2ed670+parents

PR #6253, opened 2020-07-22
Add lock message when the page is in a workflow task that requires review

zerolab:
Screenshot 2020-07-22 at 19 47 51


gasman:
Thanks @zerolab - looks good to me! Merged in f26bd62cfeec627c9730713dd2b37fe757dc45c8 (with a small fix to add missing punctuation on the "awaiting moderation" message).

PR #6262, opened 2020-07-28
[RFC 54] Internationalisation logging

kaedroho:
Split from: https://github.com/wagtail/wagtail/pull/6220

We need to make some changes to audit logging in order to integrate this properly:

 - Stop `Page.copy` from returning a `wagtail.create` log. This is so we can override it in a consistent way. `wagtail.copy` is a specific case of `wagtail.create`. So `wagtail.copy_for_translation` can become a specific case of `wagtail.copy`. The alternative would be to create three logs every time we copy for translation.
 - Use the same log action for descendants (currently always uses `wagtail.copy` for descendants no matter what `log_action` is)
 - Allow extra data to be added (source locale)


kaedroho:
Superseded by https://github.com/wagtail/wagtail/pull/7352

PR #6264, opened 2020-07-28
Fix/streamfield form has changed

jacobtoppm:
This fixes a couple of issues with nonlinear workflows ('WAGTAIL_WORKFLOW_REQUIRE_REAPPROVAL_ON_EDIT=True`):

1. Due to checking `form.has_changed` when deciding whether to save a new revision on a workflow action from the edit view, and the fact that pages with StreamFields always returned True for this, this meant that nonlinear workflows would always act as if edited when being approved, and so return to earlier tasks to be reapproved. This PR adds a custom `has_changed` method to `BlockField` to resolve this.

2. The action menu button for the final workflow step would include "and publish" for the approve action. However, if edits are made, this is misleading, as the page will actually return to previous tasks for reapproval first. This removes "and publish" text for nonlinear workflows for now, although in future we could consider doing something with the form check used to provide the "unsaved changes" message on leaving the edit view.

### DIFF
--- a/wagtail/core/blocks/base.py
+++ b/wagtail/core/blocks/base.py
@@ -542,6 +542,9 @@ class BlockField(forms.Field):
     def clean(self, value):
         return self.block.clean(value)
 
+    def has_changed(self, initial_value, data_value):
+        return self.block.get_prep_value(initial_value) != self.block.get_prep_value(data_value)
+
 
 DECONSTRUCT_ALIASES = {
     Block: 'wagtail.core.blocks.Block',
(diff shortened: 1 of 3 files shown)


wagtail/core/blocks/base.py
```diff
@@ -542,6 +542,9 @@ def __init__(self, block=None, **kwargs):
     def clean(self, value):
         return self.block.clean(value)
 
+    def has_changed(self, initial_value, data_value):
```
zerolab, line 545: was that all it took to fix `form.has_changed` being True for StreamFields? 🤯

zerolab: With this change, https://github.com/wagtail/wagtail/blob/master/wagtail/admin/views/pages.py#L503 should also change to something along the lines of:

> if there are content changes then create a new revision, otherwise use `latest_revision`, regardless of wether or not this is a workflow action

I think that is, but it has been a long day so will revisit tomorrow

PR #6267, opened 2020-07-29
Add hooks for snippet listing buttons

kaedroho:
This PR implements two new hooks to allow customising the listing buttons on snippets:
 - `register_snippet_listing_buttons`
 - `construct_snippet_listing_buttons`

These are based on  `register_page_listing_buttons` and  `construct_page_listing_buttons`.

I need the `register_page_listing_buttons` hook in order to add a "Translate" button for snippets in `wagtail-localize`.

Sponsored by The Mozilla Foundation.

This depends on https://github.com/wagtail/wagtail/pull/6268

PR #6268, opened 2020-07-29
Add hooks.register_temporarily for testing hooks

kaedroho:
There's currently a number of places where we test hooks by registering
hook functions from test methods. These are never cleaned up so that
hook remains registered in future tests. This can cause issues with
tests not working consistently when you run a subset of the tests.

This adds a `register_temporarily` function to the hooks module. This
function can be used either as a decorator or a context manager. It'll
always remove the hook after the decorated function or with block exits,
even if it exits through an exception.

This is required in order to get tests working in https://github.com/wagtail/wagtail/pull/6267

### DIFF
--- a/wagtail/core/hooks.py
+++ b/wagtail/core/hooks.py
@@ -1,3 +1,4 @@
+from contextlib import ContextDecorator
 from operator import itemgetter
 
 from wagtail.utils.apps import get_app_submodules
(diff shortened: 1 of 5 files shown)


wagtail/core/hooks.py
```diff
@@ -32,6 +33,48 @@ def decorator(fn):
     _hooks[hook_name].append((fn, order))
 
 
+class TemporaryHook(ContextDecorator):
```
zerolab, line 36: I like the idea, but should this not be in `tests/utils/hooks.py`, similar to `tests/utils/form_data.py`?

Unless there's wider scope to support this in core..

kaedroho: I think it's best to keep all registration logic in the main hooks module. It directly modifies the `_hooks` global variable that I think should be considered internal to the hooks module.

Also, I plan to document this, third party modules could use it for testing their custom hooks.

PR #6269, opened 2020-07-29
Add before_{create,edit,delete}_snippet hooks

kaedroho:
Similar to how we already have `before_{create,edit,delete}_snippet` hooks, implement the before variants of these hooks like we have for pages.

This is required for implementing a new translation UI in `wagtail-localize` which is sponsored by The Mozilla Foundation.

### DIFF
--- a/docs/reference/hooks.rst
+++ b/docs/reference/hooks.rst
@@ -1032,7 +1032,7 @@ Hooks for working with registered Snippets.
     @hooks.register('before_edit_snippet')
     def block_snippet_edit(request, instance):
         if isinstance(instance, RestrictedSnippet) and instance.prevent_edit:
-            return HttpResponse(f"Sorry, you can't edit this snippet", content_type="text/plain")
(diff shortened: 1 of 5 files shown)


docs/reference/hooks.rst
```diff
@@ -1016,6 +1016,24 @@ Hooks for working with registered Snippets.
...
+  .. code-block:: python
+
+    from django.http import HttpResponse
+
+    from wagtail.core import hooks
+
+    @hooks.register('before_edit_snippet')
+    def block_snippet_edit(request, instance):
+        if isinstance(instance, RestrictedSnippet) and instance.prevent_edit:
+            return HttpResponse(f"Sorry, you can't edit this snippet", content_type="text/plain")
```
KalobTaulien, line 1035: Do we need an `f` string here?

kaedroho: nope, thanks!

PR #6273, opened 2020-07-30
Add system check to make sure Orderable takes effect

kaedroho:
I hit a weird issue when working on tests for Internationalisation. When I added `TranslatableMixin` to a child object, the ordering of that child object would change.

It took me a while to figure it out, but it turns out this is because `TranslatableMixin` defines a `Meta` and when both `TranslatableMixin` and `Orderable` are used on the same model, only the `Meta` object from the first item in the bases will be used.

This PR adds a system check to warn developers that their `Orderable` objects are not being sorted correctly.

Turns out this is an issue on bakerydemo with `FormField`!


ababic:
Hi @kaedroho. This is fantastic work, so don't want to hold things up if you're short on time. However, we could take a more 'corrective' approach here and use a (python) metaclass to add `ordering` back on to `Meta` in cases where it has been accidentally overridden. For example, Mezzanine's version of `Orderable` does this, and I never got caught out after years of working with it: https://github.com/stephenmcd/mezzanine/blob/master/mezzanine/core/models.py#L383

ababic:
Come to think of it - Maybe these two solutions compliment one another nicely. If `ordering` is missing, it can be added, or if it has been deliberately overridden (making `Orderable` redundant), `check()` can issue the warning?

kaedroho:
Thanks @ababic!

Is it possible for a model to have two metaclasses? We have a similar issue on `TranslatableMixin` where the `unique_together` can be removed if the child model has overridden meta. It's possible that a model could have both `Orderable` and `TranslatableMixin` at the same time, so if we went with this solution for both of these mixins we might end up with a model with two separate metaclasses.

ababic:
I don't think that would be a problem.

I've used metaclasses on page types before to autogenerate sets of fields,
and that didn't prevent the metaclass from the generic page model from
doing it's thing. As far as I'm aware, subclasses always inherit any/all
metaclasses from their parents.

kaedroho:
> I've used metaclasses on page types before to autogenerate sets of fields, and that didn't prevent the metaclass from the generic page model from doing its thing. As far as I'm aware, subclasses always inherit any/all metaclasses from their parents.

I think this is a bit different though. In that case, you're inheriting a metaclass so you can override it. But here, we need to somehow integrate with metaclasses in other mixins (not parent classes) that we're not aware of.

ababic:
@kaedroho I think there could be a little confusion around this... `Orderable` would have a single metaclass that would set `ordering`  on the Meta class if it was missing, and `TranslatableMixin` would have a single metaclass that would set `unique_together` on the Meta class if it was missing.

If you defined a model that subclassed both `Orderable` and `TranslatableMixin`, both their metaclasses should still activate.... Just like whenever you subclass Django's `Model` class to create a model and add a custom metaclass (like `Page` does), Django's `Model` metaclass still runs and turns the field definitions (added as class attributes) to real functioning fields (and moves them to the classes `_meta` attribute).

ababic:
@kaedroho Apologies for this nonsense

> If you defined a model that subclassed both Orderable and TranslatableMixin, both their metaclasses should still activate.... Just like whenever you subclass Django's Model class to create a model and add a custom metaclass (like Page does), Django's Model metaclass still runs and turns the field definitions (added as class attributes) to real functioning fields (and moves them to the classes _meta attribute).

You need to subclass `django.db.models.base.ModelBase` to define your metaclasses in a way that preserve Django's behaviour. So, applying this for `Orderable` would make sense, because that isn't used in combination with pages (and if a different metaclass happened to knock it out, it isn't the end of the world - as the check would still do its thing).

If you wanted to do this for `TranslatableMixin`, you could define two versions: `TranslatableMixin` and `TranslatablePageMixin` - one of which would have a metaclass based on `django.db.models.base.ModelBase`, and the other would have a metaclass based on `wagtail.core.models.PageBase`

gasman:
Happy with the principle of the system check, but I think this might be slightly mis-diagnosed...

When defining a `class Meta` on a model whose superclass already has one, the proper thing to do is [define the inheritance on `Meta` too](https://docs.djangoproject.com/en/3.1/topics/db/models/#meta-inheritance):

    class Address(Orderable):
        ...
        class Meta(Orderable.Meta):
            verbose_name_plural = 'Addresses'

([The `ordering = []` in `TestOrderableSystemChecks`](https://github.com/wagtail/wagtail/pull/6273/files#diff-8dc3f4807891f1a9a565b83b9aa4f91aR335) is redundant - the presence of `class Meta` is itself enough to clobber the `ordering` property, and I'd suggest that replacing it with a plain `pass` or an unrelated property like `verbose_name` would make it a more representative test case.)

I suppose the corollary to that is that if you're inheriting from multiple superclasses that don't know about each other, you need to do this even if you're not setting any Meta options yourself:

    class Address(TranslatableMixin, Orderable):
        ...
        class Meta(TranslatableMixin.Meta, Orderable.Meta):
            pass

Ideally I think we'd want to encourage developers to go this route rather than redefining `ordering` themselves - and likewise for [the corresponding check in TranslatableMixin](https://github.com/wagtail/wagtail/blob/738003726900295328f263bfe91686026f70f596/wagtail/core/models.py#L373-L382). That does make the advice in the warning a bit more convoluted, because we have to differentiate between:

1) the developer added a class Meta for an unrelated reason (e.g. `verbose_name`), and now needs to add `Orderable.Meta`to fix it
2) the developer added a class Meta (with or without the `Orderable.Meta`) with the express intention of overriding `ordering`, and now needs to be told to add `'sort_order'` back
3) the developer is extending both TranslatableMixin and Orderable, and is quite possibly unaware of the need for a class Meta at all, and now needs to be told to add one

(I think we can reasonably leave out the "remove inheritance from Orderable" suggestion from the message - it seems pretty unlikely that that will ever be the right thing to do...)

I think this might call for a two-step process in the check method, in both Orderable and TranslatableMixin: check the inheritance directly (`issubclass(cls.Meta, Orderable.Meta)`), and if that succeeds, go on to check the property (`'sort_order' in cls._meta.ordering`). This would make the check noisier, as it will flag up code that reinstates `sort_order` in the "wrong" way ([like wagtailforms](https://github.com/wagtail/wagtail/blob/738003726900295328f263bfe91686026f70f596/wagtail/contrib/forms/models.py#L177-L179)...), but maybe that's not a bad thing.

thibaudcolas:
We discussed this with the core team today and not sure where to take this next :) It’s been a while, seems like a neat improvement but apparently complex enough for us not to be sure how to proceed. Any thoughts?

lb-:
@thibaudcolas  @gasman 

I will close this PR and create an enhancement request from it, hopefully someone else can pick this up in the future.

lb-:
OK, I have done my best to try to bring in the various comments & original goal into an issue.

It would be great if anyone can review (core team, just edit the issue) so it's solid enough for a future contribution or PR.

https://github.com/wagtail/wagtail/issues/13316

PR #6275, opened 2020-07-31
Move SpreadsheetExportMixin to wagtail.admin.views.mixins

### LINKED ISSUE #6265: 2.10 Regression - Importing SubmissionsListView causes error
### Issue Summary

`from wagtail.contrib.forms.views import SubmissionsListView` throws an error.

### Steps to Reproduce

1. Follow the instructions to create a custom `SubmissionListView` here from `wagtail.contrib.forms.views import SubmissionsListView`
2. `from wagtail.contrib.forms.views import SubmissionsListView` causes an error
3. error: `django.core.exceptions.AppRegistryNotReady: Models aren't loaded yet.`

### Technical details
* Wagtail version: 2.10rc1

### Recommendations

* Investigate if we can split out the views in a different way to make this easier to use
* We should update the docs about any changes to imports that need to be done for 2.10

gasman:
Fixes #6265. The new django-filter code in wagtail.admin.views.reports is not compatible with loading that module before models have been loaded; this broke the recipe at https://docs.wagtail.io/en/stable/reference/contrib/forms/customisation.html#customise-form-submissions-listing-in-wagtail-admin , as it involved loading wagtail.contrib.forms.views at model loading time, and that was importing SpreadsheetExportMixin from wagtail.admin.views.reports. Moving this out of wagtail.admin.views.reports breaks the deadlock.


lb-:
@gasman  thanks heaps for fixing this, nice solution.

Issue #6271, opened 2020-07-30
Nested inline panels are ignored by page copy


kaedroho:
# Issue Summary

[Nested inline panels](https://github.com/wagtail/wagtail/pull/5566) were introduced in Wagtail 2.8. However, there are no tests for if page.copy() copies them and it appears to be broken on Wagtail 2.11.

Not sure if this was broken by the recent refactoring that's been done on the copy method.

### Steps to Reproduce

1. Create an insteance of `testapp.EventPage` with a speaker and award
2. Call `.copy()` on that page
3. The resulting page will have the speaker but not the award

### Technical details

* Python version: 3.8
* Django version: 3.0
* Wagtail version: 2.11 Alpha
* Browser version: Firefox 71

gasman:
FWIW, nested inline panels were downgraded to 'experimental' in 7a92ca5ad916d7e37de11a24080f5b0ca06dbd10 as there are still some outstanding issues - in particular, ID assignment and JS initialisation are still erratic because the `<script>` blocks containing the HTML fragments get nested, resulting in IDs with multiple levels of `__prefix__` placeholders and no way for the handling code to know which ones refer to the outer or inner fragment.

PR #6277, opened 2020-08-01
Simplify Page.copy()

kaedroho:
This PR refactors the copy code to remove some of the spaghetti and hopefully make it more reusable (it still only works with Clusterable models though!). I've also moved bits into Django modelcluster itself (https://github.com/wagtail/django-modelcluster/pull/128 and https://github.com/wagtail/django-modelcluster/pull/129)

This also means that the nested copying feature https://github.com/wagtail/wagtail/issues/6271 can now be completely handled by Django modelcluster so I'll fix that there.

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -51,7 +51,7 @@ PAGE_TEMPLATE_VAR = 'page'
 
 
 def _extract_field_data(source, exclude_fields=None):
-    """Get dictionaries representing the model: one with all non m2m fields, and one containing the m2m fields"""
+    """Get dictionaries representing the model's field data. Excluding many to many fields (which are handled by _copy_m2m_relations)'"""
     exclude_fields = exclude_fields or []
     data_dict = {}
 
@@ -84,21 +84,6 @@ def _extract_field_data(source, exclude_fields=None):
     return data_dict
 
 
-def _make_copy(source, data_dict, update_attrs=None):
-    """Create a copy instance (without saving) from dictionaries of the model's fields, and update any attributes in update_attrs"""
-
-    if not update_attrs:
-        update_attrs = {}
-
-    target = source.__class__(**data_dict)
-
-    if update_attrs:
-        for field, value in update_attrs.items():
-            setattr(target, field, value)
-
-    return target
-
-
 def _copy_m2m_relations(source, target, exclude_fields=None, update_attrs=None):
     """Copy non-ParentalManyToMany m2m relations"""
     update_attrs = update_attrs or {}
@@ -126,7 +111,12 @@ def _copy_m2m_relations(source, target, exclude_fields=None, update_attrs=None):
 
 def _copy(source, exclude_fields=None, update_attrs=None):
     data_dict = _extract_field_data(source, exclude_fields=exclude_fields)
-    target = _make_copy(source, data_dict, update_attrs=update_attrs)
+    target = source.__class__(**data_dict)
+
+    if update_attrs:
+        for field, value in update_attrs.items():
+            setattr(target, field, value)
+
     child_object_map = source.copy_all_child_relations(target, exclude=exclude_fields)
     return target, child_object_map
 
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -131,34 +131,18 @@ class MultiTableCopyMixin:
 
     def _copy_child_objects_to_instance(self, instance, exclude_fields=None, process_child_object=None):
         """Copy objects linked to the model by a ParentalKey, and set this to the new revision"""
-
-        # A dict that maps child objects to their new ids
-        # Used to remap child object ids in revisions
-        child_object_id_map = defaultdict(dict)
-        exclude_fields = exclude_fields or []
         specific_self = self.specific
-        for child_relation in get_all_child_relations(specific_self):
-            accessor_name = child_relation.get_accessor_name()
 
-            if accessor_name in exclude_fields:
-                continue
+        child_object_map = specific_self.copy_all_child_relations(instance, exclude=exclude_fields)
 
-            parental_key_name = child_relation.field.attname
-            child_objects = getattr(specific_self, accessor_name, None)
+        # Save copied child objects and run process_child_object on them if we need to
+        child_object_id_map = defaultdict(dict)
+        for (child_relation, old_pk), child_object in child_object_map.items():
+            if process_child_object:
+                process_child_object(specific_self, instance, child_relation, child_object)
 
-            if child_objects:
-                for child_object in child_objects.all():
-                    old_pk = child_object.pk
-                    child_object.pk = None
-                    setattr(child_object, parental_key_name, instance.id)
-
-                    if process_child_object is not None:
-                        process_child_object(specific_self, instance, child_relation, child_object)
-
-                    child_object.save()
-
-                    # Add mapping to new primary key (so we can apply this change to revisions)
-                    child_object_id_map[accessor_name][old_pk] = child_object.pk
+            child_object.save()
+            child_object_id_map[child_relation.get_accessor_name()][old_pk] = child_object.pk
 
         return child_object_id_map
 
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -128,22 +128,7 @@ class MultiTableCopyMixin:
 
         return instance
 
-    def _copy_child_objects_to_instance(self, instance, exclude_fields=None, process_child_object=None):
-        """Copy objects linked to the model by a ParentalKey, and set this to the new revision"""
-        specific_self = self.specific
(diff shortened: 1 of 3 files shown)


kaedroho:
@gasman since this should work with any clusterable model and doesn't actually depend on anything page/taskstate specific anymore (eg multi table inheritance). Shall I make all this copy code a utility in modelcluster itself?

wagtail/core/models.py
```diff
@@ -84,6 +84,8 @@ def _get_field_dictionaries(self, exclude_fields=None):
                             continue
                     except AttributeError:
                         pass
+
+                    # TODO: Wouldn't this reassign the objects to the new page rather than copy them?
```
kaedroho: Note: Not checked this yet. Let me know if you know if this is an issue or not!

wagtail/core/models.py
```diff
@@ -48,131 +48,92 @@
...
+                try:
+                    # Do not copy m2m links with a through model that has a ParentalKey to the model being copied - these will be copied as child objects
+                    through_model_parental_links = [field for field in field.through._meta.get_fields() if isinstance(field, ParentalKey) and (field.related_model == source.__class__ or field.related_model in source._meta.parents)]
+                    if through_model_parental_links:
+                        continue
+                except AttributeError:
+                    pass
 
-        copy_instance = specific_class(**specific_dict)
+                # TODO: Wouldn't this reassign the objects to the new page rather than copy them?
```
kaedroho, line 84: Note: Not checked this yet. Let me know if you know if this is an issue or not!

kaedroho:
@gasman 

This one is ready for review now. Tests are currently breaking because it depends on the latest modelcluster.

I didn't get time to finish the move to `copy_cluster`, but I think this can be done separately. I've moved progress on that into https://github.com/kaedroho/wagtail/pull/56

wagtail/core/models.py
```diff
@@ -3630,11 +3581,10 @@ def cancel(self, user=None, resume=False, comment=''):
     def copy(self, update_attrs=None, exclude_fields=None):
         """Copy this task state, excluding the attributes in the ``exclude_fields`` list and updating any attributes to values
         specified in the ``update_attrs`` dictionary of ``attribute``: ``new value`` pairs"""
-        copy_instance, _ = self._copy(exclude_fields, update_attrs)
-        return copy_instance
-
-    def _save_copy_instance(self, instance, **kwargs):
+        exclude_fields = self.default_exclude_fields_in_copy + self.exclude_fields_in_copy + (exclude_fields or [])
+        instance, _ = _copy(self.specific, exclude_fields, update_attrs)
```
zerolab, line 3585: @kaedroho wouldn't the use of `_` conflict with the `from django.utils.translation import gettext_lazy as _` import?

kaedroho: oh yes, nice spot!

kaedroho: This is on master too: https://github.com/wagtail/wagtail/blob/master/wagtail/core/models.py#L3633

wagtail/core/models.py
```diff
@@ -87,6 +87,8 @@ def _get_field_dictionaries(self, exclude_fields=None):
                             continue
                     except AttributeError:
                         pass
+
+                    # TODO: Wouldn't this reassign the objects to the new page rather than copy them?
```
gasman: @kaedroho Isn't that what you want to happen for an M2M relation? If a page has an M2M relation linking it to several categories, then you want the copied page to be linked to those same categories - you don't want it to duplicate the category objects.

kaedroho: I was just wondering if this would remove them from the source page if we were passing in the through model instances here?

gasman: I don't think so - on the line below we're calling `all()` on the relation, which skips over the through model and returns a queryset of the model at the far end of the relation. When these are assigned back to the relation on the copied page, it will create new through models for them.

That does have me wondering what should happen if there's an explicit through model with extra fields, though: as it stands, that extra data will be lost on the copied page (or fail if any of those fields are required). In fact, at this point I'm not even sure what the proper use-case is for a non-parental ManyToMany field on a ClusterableModel, and whether it's really appropriate to copy these relations across at all. There's the one special case being addressed directly above, where an M2M relation is being set up as a convenience for accessing a classic ParentalKey/InlinePanel relation ("get all of this image gallery's Image objects"), which makes sense - but other than that, it seems to me that non-parental ManyToMany relations aren't going to be used for actual page content, and their usefulness is confined to things like "user X has bookmarked page Y". And in that case, you really don't want the relation to be copied...

kaedroho: Given this is a refactoring PR, shall we open a new PR to change that behaviour?

gasman: Sure, that sounds fair (and if non-parental M2Ms aren't really used in practice, maybe this isn't a priority to fix anyhow). This comment stopped me in my tracks because it looked like you might have gone on to wrongly fix this by duplicating the objects on the other side of the relation... if that's not the case, then I'm happy to continue reviewing!

wagtail/core/models.py
```diff
@@ -3604,13 +3602,14 @@ def cancel(self, user=None, resume=False, comment=''):
     def copy(self, update_attrs=None, exclude_fields=None):
         """Copy this task state, excluding the attributes in the ``exclude_fields`` list and updating any attributes to values
         specified in the ``update_attrs`` dictionary of ``attribute``: ``new value`` pairs"""
+        def _save_copy_instance(instance):
+            # Called from self._copy when it's ready to save an instance
+            instance.save()
+            return instance
+
         copy_instance, _ = self._copy(exclude_fields, update_attrs)
```
gasman: I think you forgot to update the call to `_copy` to pass in `_save_copy_instance` here. (Looks like this gets refactored out completely later on, though, so perhaps that doesn't matter...)

gasman: yep, https://github.com/wagtail/wagtail/pull/6277/commits/4fea8cd71bc1e8963bc11d9f2224433f4f920766#diff-984f91690294c77757dbe73c72481f9bR3595 is where it would have been removed again.

kaedroho: Oh, sorry. I was sort of making this up as I went along! Feel free to squash all this at the end though

wagtail/core/models.py
```diff
@@ -74,17 +73,6 @@ def _extract_field_data(source, exclude_fields=None):
                     values = parental_field.all()
                     if values:
                         data_dict[field.name] = values
```
gasman, line 75: Now that we're no longer building `m2m_dict` here, the comment above this block (`# Otherwise add them to the m2m dict to be set after saving`) needs updating.

wagtail/core/models.py
```diff
@@ -106,29 +94,41 @@ def _make_copy(source, data_dict, m2m_dict, update_attrs=None):
 
     if update_attrs:
         for field, value in update_attrs.items():
-            if field in m2m_dict:
-                continue
```
gasman, line 144: As far as I can see, this check hasn't been accounted for in the new code. If `update_attrs` contains any (non-parental) M2M fields, then the following `setattr` will fail because `target` is unsaved at this point.

gasman: Replacing this with

    if field not in data_dict:
        continue

should work, I think, since `data_dict` should contain exactly the fields that are legal to update at this point. (Unless there's a discrepancy between `foo` and `foo_id`, possibly...)

gasman: actually, that will cause any unrecognised items in `update_attrs` to be silently discarded rather than raising an exception, which probably isn't ideal :-/

gasman:
Besides the comment about the m2m dict in `_extract_field_data` and the missing check for m2m fields in `update_attrs`, everything else looks good to me!

kaedroho:
@gasman Addressed your comments in the last commit

wagtail/core/models.py
```diff
@@ -115,6 +120,8 @@ def _copy(source, exclude_fields=None, update_attrs=None):
 
     if update_attrs:
         for field, value in update_attrs.items():
+            if field not in data_dict:
+                continue
```
gasman, line 124: I think there's still a possible failure case when `update_attrs` includes a foreign key field referenced as `foo_id`, because `data_dict` only includes the `foo` variant. However, in the case of pages at least, it looks like this fixes itself further down when we create a new revision:
https://github.com/wagtail/wagtail/blob/5e6a6746869d351eb2c55e092d6e669da1af896f/wagtail/core/models.py#L1493-L1495

Given the amount of reaching I'm doing to try (and fail) to turn this into a failing test case, and the unlikeliness of this happening in the real world (I'm pretty sure that `update_attrs` is only going to be used on well-established fields like `slug`), I'm going to call this good enough :-)

PR #6282, opened 2020-08-01
Move log registry to wagtailcore and validate log actions

kaedroho:
This PR moves the log registry into wagtailcore and adds validation to make sure log actions are registered at the point of inserting new logs. I've also renamed the registry so that we can have multiple registries for different object types.

This PR:

- [x] Move log registry to core
- [x] Make log registry specific to pages (so pages can have a different set of actions to images, for example)
- [x] Validate log action on save

Future PRs:

- Polymorphic log actions (where a `wagtail.copy` acts as a specific version of `wagtail.create`)
- Grouping logs (Logs that were generated at the same time are linked in the database and displayed as one)
- Auto-assign user if log generated in Wagtail admin (in case this is forgotten, or not possible due to the way the view is coded)


wagtail/core/tests/test_audit_log.py
```diff
@@ -1,280 +1,265 @@
...
-        self.assertEqual(PageLogEntry.objects.get_for_user(self.user).count(), 1)  # the create from setUp
+        with self.register_hook('register_log_actions', test_hook):
+            log_actions = LogActionRegistry('register_log_actions')
+            self.assertIn('test.custom_action', log_actions.get_actions())
+            self.assertEqual(log_actions.format_message(log_entry), "Tested!")
+            self.assertEqual(log_actions.get_action_label('test.custom_action'), 'Custom action')
 
 
-class TestAuditLog(TestCase):
+class TestAuditLogAdmin(TestCase, WagtailTestUtils):
```
kaedroho, line 61: Oops

wagtail/core/wagtail_hooks.py
```diff
@@ -181,9 +181,9 @@ def rename_message(data):
         except KeyError:
             return _('Renamed')
 
-    actions.register_action('wagtail.rename', _('Rename'), rename_message)
-    actions.register_action('wagtail.revert', _('Revert'), revert_message)
-    actions.register_action('wagtail.copy', _('Copy'), copy_message)
+    actions.register_action('wagtail.rename', _('Rename'), rename_message, parent='wagtail.edit')
```
kaedroho, line 184: Should be separate.

wagtail/core/tests/test_audit_log.py
```diff
@@ -24,19 +24,19 @@ def setUp(self):
     def test_register_log_actions_hook(self):
         # testapp/wagtail_hooks.py defines a 'blockquote' rich text feature with a hallo.js
```
jacobtoppm, line 26: I don't think this comment applies here, unless I'm missing some rich text integration...

wagtail/core/tests/test_audit_log.py
```diff
@@ -1,280 +1,265 @@
-from datetime import datetime, timedelta
+from datetime import timedelta
```
gasman, line 1: It appears that you've moved wagtail/admin/tests/test_audit_log.py here, and overwritten the existing wagtail/core/tests/test_audit_log.py.

wagtail/core/tests/test_audit_log.py
gasman: This was already in the old code in wagtail/admin/tests/test_audit_log.py.

gasman (changes requested):

kaedroho:
Sorry for the mistake, this is ready for another review now!

gasman (approved):

wagtail/core/wagtail_hooks.py
```diff
@@ -92,3 +93,193 @@ def describe_collection_children(collection):
...
+
+    def convert_alias_message(data):
+        try:
+            return _("Converted the alias '%(title)s' into a regular page") % {
+                'title': data['page']['title'],
+            }
+        except KeyError:
+            return _("Converted an alias into a regular page")
+
+    def move_message(data):
```
zerolab, line 145: note: this PR doesn't have the changes from #6762

gasman:
Merged in 7423d72efe58d38e9a287999940b9b302cac58e4 + parents

PR #6284, opened 2020-08-01
Version command

KalobTaulien:
A small --version command for the global wagtail command.

image

### DIFF
--- a/wagtail/bin/wagtail.py
+++ b/wagtail/bin/wagtail.py
@@ -255,7 +255,7 @@ class Version(Command):
 COMMANDS = {
     'start': CreateProject(),
     'updatemodulepaths': UpdateModulePaths(),
-    'version': Version()
+    '--version': Version()
 }
 
 
--- a/wagtail/bin/wagtail.py
+++ b/wagtail/bin/wagtail.py
@@ -243,7 +243,7 @@ class UpdateModulePaths(Command):
 
 
 class Version(Command):
-    description = "List which version of watch you are using"
+    description = "List which version of Wagtail you are using"
 
     def run(self):
         import wagtail
@@ -255,7 +255,7 @@ class Version(Command):
 COMMANDS = {
     'start': CreateProject(),
     'updatemodulepaths': UpdateModulePaths(),
-    '--version': Version()
+    'version': Version()
 }
 
 
--- a/wagtail/bin/wagtail.py
+++ b/wagtail/bin/wagtail.py
@@ -242,9 +242,20 @@ class UpdateModulePaths(Command):
         return change_count
 
 
+class Version(Command):
+    description = "List which version of watch you are using"
+
+    def run(self):
+        import wagtail
+        version = wagtail.get_version(wagtail.VERSION)
+
+        print("You are using Wagtail %(version)s" % {'version': version})
+
+
 COMMANDS = {
     'start': CreateProject(),
     'updatemodulepaths': UpdateModulePaths(),
+    '--version': Version()
 }
 
 


wagtail/bin/wagtail.py
```diff
@@ -242,9 +242,20 @@ def _rewrite_file(self, filename):
         return change_count
 
 
+class Version(Command):
+    description = "List which version of watch you are using"
```
tomdyson, line 246: watch -> Wagtail?

wagtail/bin/wagtail.py
```diff
@@ -242,9 +242,20 @@ def _rewrite_file(self, filename):
...
+        import wagtail
+        version = wagtail.get_version(wagtail.VERSION)
+
+        print("You are using Wagtail %(version)s" % {'version': version})
+
+
 COMMANDS = {
     'start': CreateProject(),
     'updatemodulepaths': UpdateModulePaths(),
+    '--version': Version()
```
tomdyson, line 258: I think this should be simply 'version'; the '--' suggests it's an argument for other commands, rather than a command itself.

wagtail/bin/wagtail.py
KalobTaulien: Haha yeeeep no idea what my fingers were doing there. Will fix. Thanks!

wagtail/bin/wagtail.py
```diff
@@ -242,9 +242,20 @@ def _rewrite_file(self, filename):
...
 
 
+class Version(Command):
+    description = "List which version of Wagtail you are using"
+
+    def run(self):
+        import wagtail
+        version = wagtail.get_version(wagtail.VERSION)
+
+        print("You are using Wagtail %(version)s" % {'version': version})
```
marteinn, line 252: I would replace `print()` with `self.stdout.write("XXX", ending='')`, as its recommended in the django docs:

**When you are using management commands and wish to provide console output, you should write to self.stdout and self.stderr**

wagtail/bin/wagtail.py
Scotchester: I respectfully disagree with you, @tomdyson. I think using the double-hyphen for the version output command is very common. For example, big names that are a part of our stack use it: `python --version`, `node --version`, `npm --version`. Using the double-hyphen style would maintain consistency with those.

wagtail/bin/wagtail.py
KalobTaulien: I didn't know this! I'll update it shortly. Thanks!

wagtail/bin/wagtail.py
tomdyson: You've convinced me @Scotchester! I withdraw my comment, @KalobTaulien :)

KalobTaulien: Heh that's why I went with --version because Python does it. I'll revert shortly. Thanks both of you!

wagtail/bin/wagtail.py
KalobTaulien: @marteinn I just looked into this, and this isn't using a Django command, it's using a standard Python class. `self.stdout.write()` doesn't exist. Thoughts on keeping this as a print statement as these are Python commands?

marteinn: @KalobTaulien Doh! Its not a management command. In that case I say go with print() as there is already a precedent in the module at other places like here: https://github.com/wagtail/wagtail/blob/4ad4a33bfb9fedc87587a774ed5113abdedefd5a/wagtail/bin/wagtail.py#L96

marteinn:
LGTM 👍

wagtail/bin/wagtail.py
kaedroho: I still wouldn't call this it's own "sub command" though. If you ever run `--help` on a tool that uses subcommands, `--version` is always listed as an option and not a command.

`npm` does actually have a `version` command but this is different to `npm --version`

PR #6286, opened 2020-08-03
Cache the SVG icon HTML fragment

gasman:
The `{% icons %}` tag runs on every admin page render, and without caching, this will perform a template fetch for every .svg file - which is exceptionally slow on environments with shared filesystems (e.g. Vagrant). Adding this cache reduces the runtime of `wagtail.admin.tests.pages.test_create_page` from 89 to 57 seconds.


gasman:
Note that #6243 will also solve this issue (and improve performance further, as there is no need to include the SVG data in the page response) - get_sprite_hash still needs to load all .svg files, but as this happens at the point of building the urlconf, it will happen once at server startup rather than on every page request.

PR #6290, opened 2020-08-04
Fix WAGTAILFRONTENDCACHE_LANGUAGES

kaedroho:
The code was incorrectly expecting a 2-tuple so it was purging URLs with the first letter of the language code when the langauge code had two characters. It crashed with more than two letters.


kaedroho:
Thanks @gasman !

PR #6295, opened 2020-08-05
Stronger check on blocks names

fpoulain:
Ref: https://github.com/wagtail/wagtail-react-streamfield/issues/54

This PR stronger check on block names as stated in the [docs](https://docs.wagtail.io/en/stable/topics/streamfield.html).
> ‘name’ is used to identify the block type within templates and the internal JSON representation (and should follow standard Python conventions for variable names: lower-case and underscores, no spaces)

Since it may break existing code I made use of a warning rather than an error. Let me know how do you think about it.
Maybe only in a second phase it should become a strong error and could replace others name checks.

Thanks for Wagtail! 🎉

### DIFF
--- a/wagtail/core/blocks/base.py
+++ b/wagtail/core/blocks/base.py
@@ -1,4 +1,5 @@
 import collections
+import re
 from importlib import import_module
 
 from django import forms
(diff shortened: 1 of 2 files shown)


khashashin:
Thanks for this, I lost yesterday so much time because I didn't read the docs carefully and have definitely the name with brackets: `cards_(_left_side_icons_)`. There wasn't any warning or errors during migrations, and only checking the front end of wagtail admin interface displayed the error in the console.

gasman:
Thanks @fpoulain! Code looks good to me.

I think this should be an error, not just a warning. There's at least one place in the design of StreamField where the block name needs to be a valid Python identifier - namely, being able to write `{{ value.some_child_block }}` within templates for StructBlocks - so letting developers use invalid names to a limited extent isn't really doing them any favours.

fpoulain:
> I think this should be an error, not just a warning.

Ok I will change this.

What about the redondancy of the errors? The first check (name empty) could be left as is but the others (dash/space/numeric) could be removed since they are covered by my check.

gasman:
The redundancy of the checks did cross my mind, but it's probably a good thing that the more common mistakes (using spaces / dashes) give you a more focused error message telling you exactly what to change, rather than just "you must follow this format". So, I think this can stay as it is 👍

fpoulain:
Ok. I updated the PR. I let you commit the changelog because my english is too approximative. Moreover, since it will breaks code for some users which --like me-- didn't read carefully the docs, imho this change should be told in the release notes.

gasman:
Merged in 0c6815910b460f29784c6d8c3ef15d5c3e7eeaf0.

PR #6296, opened 2020-08-05
Split up wagtail.admin.views.pages

gasman:
`wagtail.admin.views.pages` is too big. Let's split it up.

No code changes have been made to the views, and `from wagtail.admin.views.pages.foo import *` lines have been added to keep all imports in their current locations.


cnk:
LGTM. Having separate files for these views will make things more approachable. 

I did notice that you moved `get_valid_next_url_from_request` into a `utils.py` but then did not import that in your `__init__.py` The only current use of `get_valid_next_url_from_request` in wagtail itself is in these views. But if anyone starts using that in their own code, they will need to alter their import statement. 

Also you are importing that function into other views as an absolute rather than relative import. Is that the preferred coding style? `from wagtail.admin.views.pages.utils import get_valid_next_url_from_request` vs `from .utils import get_valid_next_url_from_request`

gasman:
Thanks for reviewing so quickly, @cnk and @allcaps!

> I did notice that you moved `get_valid_next_url_from_request` into a `utils.py` but then did not import that in your `__init__.py` The only current use of `get_valid_next_url_from_request` in wagtail itself is in these views. But if anyone starts using that in their own code, they will need to alter their import statement.

That's a fair point! Our official line is that any undocumented internal methods can change at any time with no notice, and any third-party code that uses them has to accept the possibility of them being a moving target. In practice we're a bit more accommodating than that, particularly on APIs that are needed for wagtailimages / wagtaildocs and likely to be replicated elsewhere (like, say, opening modal windows). In this case, though, I think it's fair to regard this as an internal helper only intended for use within wagtail.admin.views.pages - even as an internal function I'm a bit sceptical of its usefulness (it's really just moving data from one variable to another, and the cognitive load of seeing that function call in the code and knowing what it does is little or no gain over just having that code inline...), and I wouldn't be keen to see it being imported elsewhere.

> Also you are importing that function into other views as an absolute rather than relative import. Is that the preferred coding style? `from wagtail.admin.views.pages.utils import get_valid_next_url_from_request` vs `from .utils import get_valid_next_url_from_request`

I don't think we have a stated preference either way... I'd tend to prefer the explicitness of the absolute imports, but I also wouldn't reject code that uses the relative format. In a similar situation [the Django codebase uses absolute imports](https://github.com/django/django/blob/master/django/forms/__init__.py), if that helps to cast a deciding vote :-)

> * Should we run black on these? It might be a good moment.
> * I'm not sure about `import *`. But understand it is best for backwards compatibility.

Personally I don't think converting to black formatting is a priority, but I wouldn't be against it. I would indeed like to drop the `import *` lines, but wanted to keep the changes as local as possible here. (I may well do that in a follow-up PR... hopefully the only other change will be updating the corresponding imports in urls.py.)

PR #6298, opened 2020-08-06
Quick fix for inconsistent margins between menu icons and labels

### LINKED ISSUE #6297: Inconsistent spacing between menu icons and labels
I noticed that certain menu icons have bigger right margins than others. They are supposed to be equal like in versions of Wagtail before 2.10.

### Steps to Reproduce

1. Log in to the Wagtail admin
2. Note that the space (margins) between the Reports menu item and its icon is twice as much as the rest of the items. This also occurs with site settings menu entries, where the cog icon has more spacing before the label.

In the screenshot below, I used Firefox Developer Tools to highlight the box model dimensions in each menu item. Note that the first three menu items with the cog icon use `wagtail.contrib.sitesettings`, and the "Cache" menu item is from the `wagtailcache` package.

Screenshot_20200805_133053

### Further details

It appears that this is caused by ~~extra whitespace before the actual label text, and removing them manually using DevTools in the browser brought the labels vertically aligned with the others.~~ This behavior can be reproduced reliably in both Firefox and Chrome (or any Chromium-based) browser.

**Update:** The cause of this bug isn't caused by whitespace but rather inconsistent `margin-right` values in the CSS files. It turns out the right margins of `<svg>` icon elements that are used typically for single menu items without submenus are 0.25em. But the items that use icons through CSS (i.e. their parent `<a>` elements have `icon icon-some-action` classes) use 0.5em instead.

I think this bug was introduced in eed16e0.

---

I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: **yes**

### Technical details

* Python version: 3.8.5
* Django version: 3.0.9
* Wagtail version: 2.10-rc2
* Browser version: Firefox 79, Chrome 84

acarasimon96:
Fixes #6297

* Do the tests still pass? **N/A** since this only includes a change in the SCSS files
* For front-end changes: Did you test on all of Wagtail’s supported browsers? **Yes** -- Firefox 79, Chrome 84

By the way, congratulations and thanks in advance for the big 2.10 release, Wagtail core team!

### DIFF
--- a/client/scss/components/_main-nav.scss
+++ b/client/scss/components/_main-nav.scss
@@ -76,7 +76,7 @@
         &:before {
             font-size: 1rem;
             vertical-align: -15%;
-            margin-right: 0.5em;
+            margin-right: 0.25em;
         }
 
         // only really used for spinners and settings menu


allcaps:
I close this one, we should solve this by using SVG everywhere.

PR #6299, opened 2020-08-06
Django 3.1 support for 2.10

gasman:
Backport fixes from #6205, upgrade django-rest-framework to 3.11.1 and update CI and docs.


setup.py
```diff
@@ -21,11 +21,11 @@
 
 
 install_requires = [
-    "Django>=2.2,<3.1",
+    "Django>=2.2,<3.2",
     "django-modelcluster>=5.0,<6.0",
     "django-taggit>=1.0,<2.0",
     "django-treebeard>=4.2.0,<5.0",
-    "djangorestframework>=3.7.4,<4.0",
+    "djangorestframework>=3.11.1,<4.0",
```
zerolab, line 28: DRF 3.8.0 and 3.11 introduce some breaking changes. May be worth adding a note for people to go check the DRF release notes - https://www.django-rest-framework.org/community/release-notes/

gasman: Fair point. Do you have an idea of how likely it is for people to encounter these breaking changes in practice? If they only really apply to obscure combinations of settings, then I'd be more inclined to leave it out, because you can't really give any actionable advice here - people who only use DRF through the Wagtail API won't know if the information in the release notes applies to them.

zerolab: We've encountered them on non Wagtail projects. It is not impossible, especially if people add their own views/serializers etc. Fair point on leaving this out.

PR #6300, opened 2020-08-06
Adds PageQuerySet.viewable_by_user()

ababic:
Implements a more comprehensive pattern for finding pages affected/unaffected by page view restrictions and checking in bulk which ones a user passes or fails. Page view restrictions can now effectively be applied to multiple nodes in a page's ancestry - with only the 'closest restriction' to a page affecting its 'view-ability'. For example, you can now have a subsection of public pages within a password/login protected section, then another restricted section of pages within that.

`PageQuerySet.viewable_by_user()` and `PageQuerysSet.not_viewable_by_user()` are the primary methods for accessing this functionality, with the heavy lifting mostly done on `BaseViewRestriction`. `PageViewRestriction` and `CollectionViewRestriction` override just the bare minimum to make the generic functionality on `BaseViewRestriction` work for them.

Currently, `public()` and `not_public()` methods return pages unaffected by ANY restriction. These have both been updated to piggy-back on `viewable_by_user()` and `not_viewable_by_user()` - giving them the ability to identify public pages nested within one or more private sections.

### DIFF
--- a/docs/reference/pages/queryset_reference.rst
+++ b/docs/reference/pages/queryset_reference.rst
@@ -20,7 +20,13 @@ Examples
 
       events = EventPage.objects.live().descendant_of(events_index)
 
-- Getting a list of menu items
+ - Selecting only pages that the current user is permitted to view
+
+   .. code-block:: python
+
+       events = Page.objects.viewable_by_user(user, request)
+
+ - Getting a list of menu items
 
   .. code-block:: python
 
@@ -190,6 +196,30 @@ Reference
 
     .. automethod:: not_public
 
+    .. automethod:: viewable_by_user
+
+        See: :ref:`private_pages`
+
+        .. note::
+
+            This doesn't filter out unpublished pages. If you want to only have published public pages, use ``.live().viewable_by_user(user, request)``
+
+        Examples:
+
+        .. code-block:: python
+
+            # Find all the pages that are viewable by the user making the request
+            all_pages = Page.objects.live().viewable_by_user(user, request)
+
+            # Find all the pages that are viewable by a user without a request
+            all_pages = Page.objects.live().viewable_by_user(user)
+
+            # Find pages viewable by several users:
+            fake_request = HttpRequest()  # to facilitate restriction caching
+            user_a_viewable = Page.objects.live().viewable_by_user(user_a, fake_request)
+            user_b_viewable = Page.objects.live().viewable_by_user(user_b, fake_request)
+            user_c_viewable = Page.objects.live().viewable_by_user(user_c, fake_request)
+
     .. automethod:: search
 
         See: :ref:`wagtailsearch_searching_pages`
@@ -200,6 +230,25 @@ Reference
 
             # Search future events
             results = EventPage.objects.live().filter(date__gt=timezone.now()).search("Hello")
+
+    .. automethod:: view_restricted
+
+        Example:
+
+        .. code-block:: python
+
+            # Find pages afffected by view restrictions
+            pages = Page.objects.view_restricted()
+
+    .. automethod:: not_view_restricted
+
+        Example:
+
+        .. code-block:: python
+
+            # Find pages unafffected by view restrictions
+            pages = Page.objects.not_view_restricted()
+
 
     .. automethod:: type
 
--- a/wagtail/models/__init__.py
+++ b/wagtail/models/__init__.py
@@ -26,7 +26,7 @@ from django.core.handlers.base import BaseHandler
 from django.core.handlers.wsgi import WSGIRequest
 from django.core.serializers.json import DjangoJSONEncoder
 from django.db import models, transaction
-from django.db.models import DEFERRED, Q, Value
+from django.db.models import DEFERRED, Q, QuerySet, Value
 from django.db.models.expressions import OuterRef, Subquery
 from django.db.models.functions import Concat, Substr
 from django.dispatch import receiver
@@ -2775,6 +2775,16 @@ class PageViewRestriction(BaseViewRestriction):
     class Meta:
         verbose_name = _("page view restriction")
         verbose_name_plural = _("page view restrictions")
+
+    @classmethod
+    def get_all_queryset(cls) -> QuerySet:
+        return super().get_all_queryset().select_related("page").order_by("-page__path")
+
+    def get_affected_objects_q(self) -> Q:
+        return Q(path__startswith=self.page.path)
+
+    def is_descendant_of(self, other: "PageViewRestriction") -> bool:
+        raise self.page.is_descendant_of(other.page)
 
     def save(self, user=None, **kwargs):
         """
--- a/wagtail/models/__init__.py
+++ b/wagtail/models/__init__.py
@@ -2784,7 +2784,7 @@ class PageViewRestriction(BaseViewRestriction):
         return Q(path__startswith=self.page.path)
 
     def is_descendant_of(self, other: "PageViewRestriction") -> bool:
-        raise self.page.is_descendant_of(other.page)
+        return self.page.is_descendant_of(other.page)
 
     def save(self, user=None, **kwargs):
         """
--- a/wagtail/models/collections.py
+++ b/wagtail/models/collections.py
@@ -1,5 +1,6 @@
 from django.contrib.auth.models import Group, Permission
 from django.db import models
+from django.db.models import Q, QuerySet
 from django.utils.html import format_html
 from django.utils.safestring import mark_safe
 from django.utils.translation import gettext_lazy as _
@@ -33,6 +34,21 @@ class CollectionViewRestriction(BaseViewRestriction):
     class Meta:
         verbose_name = _("collection view restriction")
         verbose_name_plural = _("collection view restrictions")
+
+    @classmethod
+    def get_all_queryset(cls) -> QuerySet:
+        return (
+            super()
+            .get_all_queryset()
+            .select_related("collection")
+            .order_by("-collection__path")
+        )
+
+    def get_affected_objects_q(self) -> Q:
+        return models.Q(collection__path__startswith=self.collection.path)
+
+    def is_descendant_of(self, other: "CollectionViewRestriction") -> bool:
+        raise self.collection.is_descendant_of(other.collection)
 
 
 class Collection(TreebeardPathFixMixin, MP_Node):
--- a/wagtail/models/view_restrictions.py
+++ b/wagtail/models/view_restrictions.py
@@ -5,9 +5,13 @@ but the definitions here should remain generic and not depend on the base wagtai
 module or specific models defined there.
 """
 
+from typing import Iterator
+
 from django.conf import settings
 from django.contrib.auth.models import Group
 from django.db import models
+from django.db.models import Q, QuerySet
+from django.http import HttpRequest
 from django.utils.translation import gettext_lazy as _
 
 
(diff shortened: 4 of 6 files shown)


wagtail/core/query.py
```diff
@@ -237,6 +272,71 @@ def not_public(self):
...
+
+        View restrictions apply to the selected page AND their descendants, so for each
+        page, it's the restriction applied to the 'closest' ancestor that determines
+        what restrictions apply to it.
+        """
+        for vr in self.all_view_restrictions:
+            if page.id == vr.page.id or page.is_descendant_of(vr.page):
+                return vr
+
+    def for_visitor(self, user=None, include_password_protected=False):
```
zerolab, line 287: ❤️ the idea.
suggestion: `for_visitor` -> `for_user`

ababic: That was my first choice, but I felt it might be able to distinguish between 'a CMS user' and 'a user of the site'. I suppose if the docstring/accompanying documentation explains things well enough, `for_user()` might be clearer.

wagtail/core/query.py
```diff
@@ -237,6 +272,71 @@ def not_public(self):
...
+
+        View restrictions apply to the selected page AND their descendants, so for each
+        page, it's the restriction applied to the 'closest' ancestor that determines
+        what restrictions apply to it.
+        """
+        for vr in self.all_view_restrictions:
+            if page.id == vr.page.id or page.is_descendant_of(vr.page):
+                return vr
+
+    def for_user(self, user=None, include_password_protected=False):
```
kaedroho, line 287: I'm confused about what `include_password_protected` means. Does it unlock all the password-restricted pages?

zerolab: the docs/code further down means it includes them because you can't know whether the user knows the password or not

wagtail/core/query.py
```diff
@@ -237,6 +272,71 @@ def not_public(self):
...
+        # widen criteria for authenticated users
+        if user is not None and user.is_authenticated:
+            user_group_pks = user.groups.all().values_list("pk", flat=True)
+            non_restrictive_types.append(PageViewRestriction.LOGIN)
+
+        # NOTE: we only need the id plus values used by Page.is_descendant_of()
+        for page in self.not_public().only("pk", "path", "depth").distinct().iterator():
+
+            # get the most relevant restriction for this page
+            closest_restriction = self._get_closest_view_restriction(page)
```
kaedroho, line 318: `.not_public()` also performs a query to get all view restrictions so it might be good to make it only do this once

wagtail/core/query.py
```diff
@@ -237,6 +272,71 @@ def not_public(self):
...
+        Return a queryset of public and private pages that should be
+        accessible to the provided ``user``, according to their group
+        memberships and current page view restriction settings.
+
+        It is impossible to know whether the user knows the password
+        to access a password-protected page/section, so those pages are
+        excluded by default. However, you can use the
+        ``include_password_protected`` keyword argument to override this
+        behaviour and include them if needed.
+        """
```
kaedroho, line 298: I wonder if we could cache the list of protected paths that apply to this user on the request?

Eg, the first time this is called, it fetches all page view restrictions and removes any that user passes. Then saves the page path's that are still restricted for subsequent calls.

Is there any particular reason why we're doing this for just the user and not the whole request? (that would also help us with the password-restricted views)

wagtail/core/query.py
ababic: @kaedroho Say you wanted to filter search results on a site to make them 'applicable' to a user/request - By default, I exclude any pages with a password restriction. However, it might be that the developer is okay with password-protected pages being included in results, and the user getting the 'Enter the password' prompt if they click on the result and are taken to the page.

wagtail/core/query.py
ababic: Ah yes. As you can probably tell from `get_view_restrictions()`, I was trying to make it so that existing users of `public()` and `not_public()` weren't penalised by the extra query to prefetch the groups. That's probably a bit over-cautious though, thinking about - and actually means two queries being made here. I feel the 'right' solution is to let `public()` and `not_public()` users take the hit, and keep this simple.

wagtail/core/query.py
ababic: In the project I implemented this on, it's only used in a couple of places (once per request), so there's no real advantage to bringing requests into it. However, I can totally see the benefit of that in a Wagtail-native solution. This has all of a sudden become much more exciting :D thanks @kaedroho

gasman:
The way this is currently written, the business rules for what the different restriction types actually mean (LOGIN, GROUPS, PASSWORD) are distributed between the `PageViewRestriction` model and the `PageQuerySet` class. This is going to make it harder to make restriction types extensible in future (https://github.com/wagtail/rfcs/pull/32)...

It would be really good if we could keep those business rules internal to the `PageViewRestriction` model. I'm thinking that could possibly be done by adding a method to `PageViewRestriction` that accepts a user object (or a request object if we go with @kaedroho's suggestion) and returns a `Q()` object representing all pages NOT blocked for the user by that restriction. `PageQuerySet.for_user` would then just need to build a combined query of all of those results. Alternatively, perhaps `PageViewRestriction` could have a static method that returns all of the `PageViewRestrictions` that the current user fails (and `PageQuerySet.for_user` returns everything that's not a child of those pages)?

ababic:
@kaedroho / @gasman okay, things are officially rejigged, taking into your feedback, and the conversations on wagtail/rfcs#32.

I decided to stick with `for_user()` in the end, as I want this to be able to cater for scenarios where `user != request.user`. However, if `request.user` and `user` are one and the same, we tap into `accept_request()` for some fun default behaviour.

ababic:
@gasman, just to clarify:

> I'm thinking that could possibly be done by adding a method to PageViewRestriction that accepts a user object (or a request object if we go with @kaedroho's suggestion) and returns a Q() object representing all pages NOT blocked for the user by that restriction. PageQuerySet.for_user would then just need to build a combined query of all of those results. Alternatively, perhaps PageViewRestriction could have a static method that returns all of the PageViewRestrictions that the current user fails (and PageQuerySet.for_user returns everything that's not a child of those pages)?

I have kept the checking of PageViewRestriction suitability at the instance-level for now. I feel like refactoring the entire way page view restrictions work here is very much out of scope for this PR. Realistically, I doubt I'd finish that for a very long time (heck, I'm struggling to see when I'll get a decent enough chunk of time to cover this well enough from a test perspective). For now, if this solution works, could we roll with it and work on the API later? The only public-facing change here is the new queryset method itself (as far as I can tell), so we can effectively replumb things however we see fit going forward.

ababic:
Hi @gasman. I've refactored this to be much more class-driven now. The only thing I'm struggling with is the new `get_permitted_objects_q()` method. Before this refactor, we essentially tested every page one-by-one, fetching the 'closest' (and therefore the most appropriate) restriction for each page, and testing that. `get_permitted_objects_q()` is not as reliable currently, and will not cater for nested restrictions very well. I will have another think about this might be refactored, but in the meantime, I'd be grateful for any suggestions.

ababic:
Okay, I think I've figured out a way for this to work. I can continue to include items from below passing restrictions, but for the failing ones, exclusions need to be effectively 'snipped' where appropriate, so that exclusion rules don't cancel out inclusive once applied lower down in the tree.

ababic:
Greetings @gasman. This has now undergone a significant refactor to simplify things and keep as much filtering logic as possible on the `BaseViewRestriction` class to simplify the planned refactor. As stated in the description (an illustrated by tests), this makes wagtail capable of supporting more complex nesting of page restriction rules (including public pages within protected sections, and visa versa), which I feel will make view restriction even more useful going forward.

ababic:
NOTE: in an earlier iteration, this method had the name `for_user`, but I decided on `viewable_by_user` on the end, as it's more descriptive. I'd also quite like to implement other permission-related methods, such as `editable_by_user`, `publishable_by_user` and `deletable_by_user` to provide more convenient access to some of Wagtail's lesser-known APIs.

kaedroho:
👍  for `viewable_by_user`

As for the other permissions, we already have ``UserPagePermissionsProxy`` which provides methods for getting ``PageQuerySet``s based on edit permissions, and these can be combined with other ``PageQuerySet``s using the ``&`` operator.

Is it worth having two API's for this? I wonder if making the ``PageQuerySet`` and ``PagePermissionsProxy`` dependant on each other is a good idea.

ababic:
Going slightly off-topic for this PR, but completely valid questions. Some quick-fire answers below, which I'll copy over to the relevant PR (when that gets created)...

> Is it worth having two API's for this? I wonder if making the PageQuerySet and PagePermissionsProxy dependant on each other is a bad idea?

These proposals are really just about putting the functionality where developers expect to find it. We'll have a few inline imports to avoid circular dependencies, but I can't forsee that causing any problems at this stage. 

> I wonder if we should be applying the "separation of policy and mechanism" principle here? So PageQuerySet just deals with querying pages and policies are implemented at some higher level?

I'm absolutely happy for the various rules/logic to be implemented elsewhere in separate components, and for devs to continue using them directly if that's what feels right to them. It's just that, currently, these various APIs are tricky to find and utilise. Django already has these well-established patterns for finding individual/multiple objects that meet certain criteria, and it's a natural place for developers look. Plus, having it directly accessible from the model class (well, via Page.objects) is great for all sorts of reasons... Cleaner code, fewer imports, easier to test things out in the Django shell - all of which carry their own significant benefits.

> It might be nice to someday allow more flexible policies to be plugged in, but maybe we shouldn't make PageQuerySet depend on any customisable behaviour?

It may take some thinking, but I'm sure we can find a design that allows for such things, but retain a consistent 'entry point'. It's just a case of handing off at the right point in the process. For example, this addition would cater just fine for custom view restriction types... the implementor just needs to override the method that tests whether a given user passes or fails.

Scotchester:
Hi @ababic! I've tried to rebase this locally and test it out (rebasing on v2.15.3 so I can test it in an existing app that is on that version of Wagtail), but it's proving more difficult than I expected. Would you mind rebasing and I'll try it out on a fresh project running main?

Scotchester:
I guess posting that comment was the motivation my brain needed to figure out why my local-rebase-on-2.15.3 setup wasn't working. Of course, it was a silly rebasing mistake. Got it sorted!

Would love to squeeze this into the 3.0 RC, but it may be too late for that, and I'm not sure if any changes are needed based on the discussion.

ababic:
@Scotchester 

> I guess posting that comment was the motivation my brain needed to figure out why my local-rebase-on-2.15.3

Always happy to be a rubber 🦆 :)

> I've tried to rebase this locally and test it out

I'm more than happy to rebase and re-push, if you'd find that helpful? Just let me know :)

Scotchester:
> I'm more than happy to rebase and re-push, if you'd find that helpful? Just let me know :)

@ababic That would be great, thank you!

wagtail/core/query.py
```diff
@@ -218,25 +220,85 @@ def not_exact_type(self, model):
...
+        for restriction in PageViewRestriction.get_all(cache_target):
             q &= ~self.descendant_of_q(restriction.page, inclusive=True)
         return q
 
-    def public(self):
+    def not_view_restricted(self, cache_target=None):
+        """
+        Filters the queryset to only include pages unaffected by page view restrictions.
+        This differs from ``public()`` in that the result will not include pages
+        affected by restrictions of type ``BaseViewRestriction.NONE``.
```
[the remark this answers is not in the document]
ababic: @gasman I think we can safely assume that people aren't using `BaseViewRestriction.NONE` in the wild in this way now - However, since this PR paves the way for more complex configurations, it could technically be used to denote a public section within a private one, as I have done so in this test here: https://github.com/wagtail/wagtail/pull/6300/files#diff-0982e81a156f529061ddfffaa49f59c5f598ba50daf8234340e8d37febff420bR653

If this restriction has special meaning elsewhere though, perhaps we should add something new to allow that to be reflected without fear of it being deleted by accident :)

ababic:
@Scotchester Rebase completed, good sir.

ababic:
Apologies @Scotchester. The rebase definitely didn't go to plan! And in my effort to restore what was there, I've clearly missed something, as test failures are rife... even though the test suite was meaningfully extended and everything was passing before. I'm afraid I'm not finding myself with the time/bandwidth to fully throw myself at this problem again immediately, so it may have to sit for a while unless someone else can pick up the mantle?

ababic:
Closing due to lack of interest.

PR #6301, opened 2020-08-06
Avoid retranslating month / weekday names that Django already provides

### LINKED ISSUE #6292: Wagtail should not change month name translations
### Issue Summary

Wagtail translations overrides month name translations (at least for Slovenian language) which changes how dates are formatted.

### Steps to Reproduce

With wagtail installed:

```python
>>> from django.utils.translation import activate                                                                                                
>>> activate("sl")                                                                                                                               
>>> from django.utils import formats 
...                                                                                                                                              
>>> from datetime import date                                                                                                                    
>>> formats.date_format(date.today())                                                                                                            
'5. Avgust 2020'
```

It should be (and without wagtail installed it is) `5. avgust 2020`.

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: (yes / no)

yes

### Technical details

Python 3.7.1

Django version: Look in your requirements.txt, or run `pip show django | grep Version`.

Version: 2.2.14

* Wagtail version: Look at the bottom of the Settings menu in the Wagtail admin, or run `pip show wagtail | grep Version:`.

Version: 2.9.2

gasman:
Fixes #6292. Having literal gettext calls here means that makemessages will add these to wagtailadmin's dictionary of strings for translation, creating redundant work for translators (and overriding Django's translations if they differ).

Note: this will not fix the specific behaviour described in #6292 until we've rebuilt, pushed and re-fetched translations from Transifex - up to that point, wagtailadmin's .po / .mo files will still contain our own translations, and the call to gettext will prioritise our translations over Django's, regardless of the fact that we're running it on a string that comes from the Django codebase rather than ours.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -34,6 +34,7 @@ Changelog
  * Fix: `Page.copy()` now raises an error if the page being copied is unsaved (Anton Zhyltsou)
  * Fix: `Page.copy()` now triggers a `page_published` if the copied page is live (Anton Zhyltsou)
  * Fix: The Elasticsearch `URLS` setting can now take a string on its own instead of a list (Sævar Öfjörð Magnússon)
+ * Fix: Avoid retranslating month / weekday names that Django already provides (Matt Westcott)
 
 
 2.10.2 (25.09.2020)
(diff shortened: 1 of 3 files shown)


bmihelac:
Looks good, I wonder why are names converted to str?

gasman:
@bmihelac The original code here was using `gettext` rather than `gettext_lazy`, so the `str` is to match the existing behaviour of returning a list of strings rather than lazy translation objects. I didn't look into why the old code was using gettext, so maybe that's not necessary.

bmihelac:
Oh yes I see now :)

thibaudcolas (approved):
Thank you @gasman, this works exactly as expected. As you mention the underlying issue will only be fixed once we do update our translation files, as part of the release process for 2.11, but I’ll mark #6292 as "fixed" anyway.

For future reference, I tried to remove the `str` conversion and the JSON serialization wasn’t happy about that, "Object of type '__proxy__' is not JSON serializable".

PR #6305, opened 2020-08-07
 Add ability to export redirects using reports

marteinn:
A common task I have is to export and send a list of the current active Wagtail redirects to various SEO partners, for this I usually roll my own export view or management command. This pull request introduces a new item in the redirects index header, along side of "Import Redirects" (in a dropdown) called "Export Redirects", it uses the report system to enable the view for filtering and data export.

Some things to note:
- This view `/redirects/report` is a report view, but are not placed in the "Reports" section in the menu. The reason is that I want to keep the redirects namespace isolated under `/redirects`. In my experiments (when I did place the report under /reports/redirects) it also caused a bug in the menu highlight, which marked both settings and reports active at the same time (when it parses the current url path).
- I reused the dropdown component from the recent reports feature to enable ui, this is the first time the dropdown is used for navigation (besides the file download actions).

Very open to feedback! The main feature I wanted is to be able to download csv/xls of redirects, the report view is just a bonus. Cheers!

### DIFF
--- a/wagtail/contrib/redirects/filters.py
+++ b/wagtail/contrib/redirects/filters.py
@@ -1,0 +1,25 @@
+import django_filters
+from django.utils.translation import gettext as _
+
+from wagtail.admin.filters import WagtailFilterSet
+from wagtail.admin.widgets import ButtonSelect
+from wagtail.core.models import Site
+
+
+class RedirectsReportFilterSet(WagtailFilterSet):
+    is_permanent = django_filters.ChoiceFilter(
+        label=_("Type"),
+        method="filter_type",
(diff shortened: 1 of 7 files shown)


gasman (approved):
Looks good to me! I'm not 100% sold on the option being part of a dropdown with Import, but I don't have any better ideas, and having the functionality in place is the more important thing, so happy to go with this.

gasman:
Merged in c0a84975e734ddcd5e733d9a27bacc417f4afab9

marteinn:
Thanks for merging @gasman! I totally agree with your thoughts about the dropdown, I had a hard time finding a good ux-component for the header. It's been a while since I made this PR, so might start to iterate on in soon.

PR #6310, opened 2020-08-08
Better Admin-User Index View sorting

zeroepix:
Added a whole bunch of sorting options for the default admin user index view.

The template now cycles between ascending sort and descending sort when clicked, and the arrows now make sense depending on the way the column is sorted. 

## Primary sorting
?ordering=x | how it sorts | which way the arrow points
-|-|-
name | last_name, first_name, username | up, because the names are ascending
-name | -last_name, -first_name, -username | down, because the names are descending
username | User.USERNAME_FIELD | up, ascending
-username | '-'+User.USERNAME_FIELD | down, descending
level | superusers, then staff, then others (-secondary_sort per group) | down, because it's descending power
-level | others, then staff, then superusers (+secondary_sort per group)| up, because it's ascending power
status | is_active, -secondary_sort | down, although this is probably not much use
-status | -is_active, +secondary_sort | up
last_login | latest login first, and puts people who have never logged in at the end (+secondary_sort) | up, because time since login is ascending
-last_login | oldest login first, and puts people who have never logged in at the end (-secondary_sort) | down, because time since login is descending

## Secondary sorting
Where the primary order is ambiguous, it will sort by last_name, first_name, then username (for the case of blank or common names).
secondary sorting | sort order | purpose
-|-|-
+secondary_sort | last_name, first_name, username | ascending order, added to resolve ambiguities
-secondary_sort | -last_name, -first_name, -username | descending order, added to resolve ambiguities

**Special note:** Django doesn't let you chain order_bys, so it's not simple to factorize the code. It looks a bit verbose and I'm not sure if there's a better way to chain it. Open to suggestions.


wagtail/users/views/users.py
```diff
@@ -90,9 +92,32 @@ def index(request, *args):
...
-        if ordering == 'username':
+        if ordering == "-name":
+            users = users.order_by('-last_name', '-first_name')
+        elif ordering == 'username':
             users = users.order_by(User.USERNAME_FIELD)
+        elif ordering == '-username':
+            users = users.order_by('-'+User.USERNAME_FIELD)
+        elif ordering == 'level':
+            superusers = users.filter(is_superuser=True)
+            staff = users.filter(is_staff=True).exclude(id__in=superusers)
```
jsonn, line 103: Why not `filter(is_staff=True, is_superuser=False)`?

jsonn (commented):
Generally, when the sorting criterion doesn't have a natural order, I would expect using either `name` or `username` as secondary order in the appropriate direction. That avoids entries getting shuffled around based on the disk order and is more deterministic in general. Over all, I like the idea.

wagtail/users/views/users.py
zeroepix: oo yes that would be better, I'll make those changes.

zeroepix: with the natural order thing - taking the is_active ascending as an example, do you mean how all the active users are sorted against each other? If so I think I get you, I'll add that but let me know if you meant something else.

zeroepix: I've added a secondary sort to all the columns and I do agree it's a big improvement. I've updated the main comment to reflex the changes. I'm worried the code is a little verbose as Django clears the order when you execute a new order_by. Not sure if there's a more elegant way than the way I've done it.

gasman (changes requested):
Thanks @zeroepix! Consistency in the ordering UI for listings has been a long-running issue in Wagtail, and this will help a lot.

One problem with this as it stands is that we need to support [custom user models](https://docs.wagtail.io/en/stable/advanced_topics/customisation/custom_user_models.html), and there's no guarantee that they will have `first_name` or `last_name` fields - at minimum they must inherit from `AbstractBaseUser`, which only guarantees that it provides a `get_full_name` method.

The old code checks for the presence of `last_name` and `first_name` before applying the order, and we'd need to do something similar in the new code to ensure that it doesn't fail if those fields are absent. A `get_full_name` method doesn't tell us anything about how to construct a suitable `order_by` clause, so there may not be anything useful we can do in that situation - we should probably fall back on ordering just by USERNAME_FIELD instead, or perhaps disable ordering on the name column.

gasman:
Also, just a note that this will address part of #5191.

zerolab:
Closing this as solved by #11823

PR #6315, opened 2020-08-11
Fix all `E303: Too many blank lines` Flake8 errors

kaedroho:
I think this is quite a useful rule to have. It makes flake8 check that the spacing between module level things and class methods are consistent.


kaedroho:
Thanks for the quick reviews!

PR #6316, opened 2020-08-11
Stop ignoring Flake8 `First argument of a method should be named 'self'`

kaedroho:
Already no breakages of this rule in the whole of Wagtail!


kaedroho:
I'm not picking up any errors for the D1XX codes for missing docstrings and I'm sure we have missing docstrings all over the place! Not sure why

PR #6318, opened 2020-08-11
Allow nested sections in tabs

kaedroho:
This CSS rule was hiding all nested sections, not just the one the tabs are using.


jacobtoppm (approved):
Looks good to me!

PR #6319, opened 2020-08-11
Fix page history view on custom user models without username field

### LINKED ISSUE #6317: [2.10] FieldError on page history view when custom user class doesn't have the username field
### Issue Summary

I'm getting an error on wagtail admin's `pages/{id}/history/` route:

```
FieldError at /wagtail/pages/12/history/

Cannot resolve keyword 'username' into field. Choices are: ...
```

### Steps to Reproduce

1. Create a wagtail project with custom AUTH_USER_MODEL which is missing the `username` field but defines the USERNAME_FIELD.
2. Navigate to page history view in wagtail admin.

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: **no**
(but I'm pretty sure this is a real and reproducible issue)

### Technical details

* Python version: 3.8.3
* Django version: 3.1
* Wagtail version: 2.10 (I didn't have this problem with 2.9)
* Browser version: latest Firefox on macOS

### Other notes

The stack trace mentions [wagtail/admin/templates/wagtailadmin/skeleton.html](https://github.com/wagtail/wagtail/blob/master/wagtail/admin/templates/wagtailadmin/skeleton.html) which doesn't mention username at all.

<s>The culprit is probably:
* the code in [wagtail/admin/templates/wagtailadmin/pages/history.html](https://github.com/wagtail/wagtail/blob/master/wagtail/admin/templates/wagtailadmin/pages/history.html)
* and possibly [wagtail/admin/templates/wagtailadmin/reports/site_history.html](https://github.com/wagtail/wagtail/blob/master/wagtail/admin/templates/wagtailadmin/reports/site_history.html)

The line `{% include "wagtailadmin/shared/user_avatar.html" with user=entry.user username=entry.username %}` is suspicious, but I haven't checked yet if replacing `username` with `get_username` fixes the problem. (Upd: this might be wrong, since it's entry.username and not user.username... I'm not sure and I'm looking into the code right now.)</s>

berekuk:
Username models without username field caused an exception on page history views.

This commit fixes #6317.

### DIFF
--- a/wagtail/admin/filters.py
+++ b/wagtail/admin/filters.py
@@ -112,9 +112,10 @@ class LockedPagesReportFilterSet(WagtailFilterSet):
 
 
 def get_requested_by_queryset(request):
-    return get_user_model().objects.filter(
+    User = get_user_model()
+    return User.objects.filter(
         pk__in=set(WorkflowState.objects.values_list('requested_by__pk', flat=True))
-    ).order_by('username')
+    ).order_by(User.USERNAME_FIELD)
 
 
 class WorkflowReportFilterSet(WagtailFilterSet):
@@ -176,9 +177,10 @@ class WorkflowTasksReportFilterSet(WagtailFilterSet):
 
 
 def get_audit_log_users_queryset(request):
-    return get_user_model().objects.filter(
+    User = get_user_model()
+    return User.objects.filter(
         pk__in=set(PageLogEntry.objects.values_list('user__pk', flat=True))
-    ).order_by('username')
+    ).order_by(User.USERNAME_FIELD)
 
 
 class SiteHistoryReportFilterSet(WagtailFilterSet):


robmoorman:
Ah also noticed this one which fixes #6326

gasman:
Thanks @berekuk! The fix here looks good - I'm now investigating whether we can add this case to our test coverage, since this is clearly a common pattern that's easy for us to miss in future development. We do have [a custom user model without a username as part of our test models](https://github.com/wagtail/wagtail/blob/b9a784adf02ebbe9d37cd40026fa83f3c86f3362/wagtail/tests/customuser/models.py#L85), but naturally it's not as simple as setting `@override_settings(AUTH_USER_MODEL='customuser.EmailUser')`, because our migrations also need to run with the relevant `AUTH_USER_MODEL` active in order to set up foreign keys. It looks like the best approach will be to add a new option to `runtests.py` that runs the test suite with AUTH_USER_MODEL = EmailUser.

gasman:
Now merged in e984190870da3929a16d76f22fafe392335fee90 (master) / ab8007ea2ff7a194fe18b5842b9e5de6a858479c (stable/2.10.x). In #6330 I've added full test coverage for user models without a username field, but as that's a much more complex change that will probably take a while to review, I think it makes sense to go ahead and merge this so that we can get a 2.10.1 patch release out.

PR #6320, opened 2020-08-12
Fix Jinja2 to include_block not including local variables

### LINKED ISSUE #5279: `include_block block with context` does not include variables defined in templates
Wagtails `include_block` tag is meant to match as closely as possible Jinja2's `include` tag 
- so one would expect that when you include the context the local template variables are also includes. However the following code does not make `foo` available to the template that `block` renders:

```
{% with foo="bar" %}
        {% include_block block with context %}
{% endwith %}
```

Looks like this is because `jinja2.nodes.ContextReference()` does not currently include template defined variables (which we use in [wagtail/core/jinja2tags.py](wagtail/core/jinja2tags.py#L39) to get the context). There is an issue relating [to this on Jinja2's repo](https://github.com/pallets/jinja/issues/860).

We could provide our own Jinja2 environment with a patched `CodeGenerator` fairly easily to include template variables (the method described in the issue above does work and django is nice enough to expose a way to load custom environments).

I'd be keen to submit a PR for this - but I'm unsure if we want to ship a patched `CodeGenerator` class? If we were to do so we'd probably want to pin the Jinja2 version in Wagtail a little more tightly to avoid breakages.

jonnyscholes:
Fixes #5279 

I wasn't quite sure what our policy is on supporting previous Jinja2 versions, so I've made it backwards compatible with >=2.8 but updated our tests to use 2.11.x

I found a file that wasnt used anywhere, the PR also removes that file as it was confusing -> `wagtail/tests/testapp/jinja2_templates/tests/jinja2/include_block_only_test.html`

* For Python changes: Have you added tests to cover the new/fixed behaviour? Yes.
* For new features: Has the documentation been updated accordingly? Correct documentation exists.


kaedroho (approved):
Looks good to me, thanks!

PR #6328, opened 2020-08-14
Implement bulk to python in struck block

jmfederico:
* Do the tests still pass? - Yes
* Does the code comply with the style guide? (Run `make lint` from the Wagtail root) - Yes
* For Python changes: Have you added tests to cover the new/fixed behaviour? - Yes

`bulk_to_python` has been an amazing addition to WagTail performance, and this MR aims to implement it in `StructBlock`.

`StructBlock` is a source of performance problems when multiple `ChooserBlock` are used in them, causing as many requests as instances of `ChooserBlock` added to the content of the field.

With this MR, `StructBlock` has its own `bulk_to_python` that in turn calls its child's `bulk_to_python` method (when available) to improve performance when loading data.

I see one possible problem with this change. The signature for `to_python` in `StructBlock` changes, adding a new optional argument.

I have this already implemented on a production site and has been working fine, and reduced the number of queries to just the number of `ChooserBlock` defined in the `StructBlock`.

### DIFF
--- a/wagtail/core/blocks/struct_block.py
+++ b/wagtail/core/blocks/struct_block.py
@@ -137,17 +137,52 @@ class BaseStructBlock(Block):
 
         return self._to_struct_value(result)
 
-    def to_python(self, value):
-        """ Recursively call to_python on children and return as a StructValue """
-        return self._to_struct_value([
-            (
-                name,
-                (child_block.to_python(value[name]) if name in value else child_block.get_default())
-                # NB the result of get_default is NOT passed through to_python, as it's expected
-                # to be in the block's native type already
+    def bulk_to_python(self, values):
+        """ Bulk load the data for fields that implement bulk_to_python """
+        bulk_in_python = {}
+
+        # Go through each child_block and pre_load data if `bulk_to_python` is available.
+        for name, field in self.child_blocks.items():
+            if not hasattr(field, 'bulk_to_python'):
+                continue
+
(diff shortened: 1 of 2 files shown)


gasman:
Hi @jmfederico - thanks for the contribution! However, this has already been worked on as part of #5976.

jmfederico:
That is actually awesome!

Well, thanks for the info 😄

wagtail/core/blocks/struct_block.py
```diff
@@ -137,17 +137,52 @@ def clean(self, value):
...
+        """ Bulk load the data for fields that implement bulk_to_python """
+        bulk_in_python = {}
+
+        # Go through each child_block and pre_load data if `bulk_to_python` is available.
+        for name, field in self.child_blocks.items():
+            if not hasattr(field, 'bulk_to_python'):
+                continue
+
+            bulk_in_python[name] = field.bulk_to_python(
+                [value[name] for value in values if name in value]
```
gasman, line 150: This is liable to leave the element indexes out-of-step, if one of the keys is missing from a dict at the beginning or middle of the list (which could happen if a new field is added to an existing StructBlock definition).

jmfederico: Should it use `None` for when no value comes? It is not possible to use `child_block.get_default()` as that is already a python object (value).

PR #6330, opened 2020-08-16
Run test suite against a user model with no username

### LINKED ISSUE #6317: [2.10] FieldError on page history view when custom user class doesn't have the username field

gasman:
Fixes #6317 / incorporates #6319 

Add two extra test runs (standard and elasticsearch) to Travis CI to run tests against an `EmailUser` model which uses the email field as the username, so that we can catch places where we've inadvertently hard-coded references to a `username` field. Wrapper functions for `create_superuser`, `create_user` and `client.login` have been added to `WagtailTestUtils to support creating users with the appropriate signature.


kaedroho (approved):
Looks good to me! I've left a couple of comments about this on https://github.com/wagtail/wagtail/pull/6354#pullrequestreview-475481809

gasman:
Merged via #6354.

PR #6335, opened 2020-08-18
Ensure block level elements in rich text are whitespace-separated in indexed content

### LINKED ISSUE #6312: #6099 can generate wrong search index content
### Issue Summary

#6099 strips out HTML tags without considering that some create line breaks. This can generate searchable content that doesn't exist in the field.

### Steps to Reproduce

Type two lines in a paragraph in a RichTextBlock:
line one
line two

The paragraph's value will be: 
```
<p>line one</p><p>line two</p>
```

get_searchable_content() will return 3 words: 'line oneline two'

gasman:
Fixes #6312

### DIFF
--- a/wagtail/core/blocks/field_block.py
+++ b/wagtail/core/blocks/field_block.py
@@ -1,5 +1,4 @@
 import datetime
-from html import unescape
 
 from django import forms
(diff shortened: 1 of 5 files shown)


kaedroho (approved):

PR #6336, opened 2020-08-18
Correctly output Draft vs Published when no revisions exist

### LINKED ISSUE #6324: Live pages without revisions reported as "Draft" in page editor
### Issue Summary

Live pages that don't have any revisions are having their status reported in the meta bar as "Draft" instead of "Published".

### Steps to Reproduce

1. Open up Bakerydemo with Wagtail 2.10 installed
2. Visit any page in the editor
3. Look at the status in the meta bar

### Technical details

Logic for this implemented in: https://github.com/wagtail/wagtail/blob/1f8fc4b6d8624d5062af255a0f917a63b6ff60ce/wagtail/admin/templates/wagtailadmin/shared/workflow_status.html#L24-L36

PS: we should stop calling `.get_latest_revision` so many times!

* Wagtail version: 2.10 (i18n branch, based on master from a couple of days ago)

gasman:
Fixes #6324
Also removed repeated calls to page.get_latest_revision.

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/shared/workflow_status.html
+++ b/wagtail/admin/templates/wagtailadmin/shared/workflow_status.html
@@ -21,18 +21,24 @@
 </li>
 {% else %}
 <li class="header-meta--status">
-    {% if page.get_latest_revision %}
-        {% if page.get_latest_revision.user %}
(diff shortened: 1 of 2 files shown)


gasman:
Thanks for the review @ababic! Merged in 8e44476e4f812261d897107dc88c8300c525880e (master) / 18a6dc1866d7eda1ee067e0bb67ba0465dcd89ca (stable/2.10.x).

PR #6337, opened 2020-08-18
Introduce a separate form for document uploads

jsonn:
Custom document models may have additional fields. Before, all those
fields had to be null=blank=True if there were in admin_form_fields
as the multiple/add view only provided values for some fields and all
others had the default value overwritten by None. This change introduces
a separate ModelForm reducted to the fields provided by the uploader,
preserving the default values for other fields.

Yes.

flake8 passes, isort hits various unrelated things.

Yes.

The documentation already specifies the behavior here.


lb-:
Sorry this never had a review, flagging as needs work as this will require a rebase /redo on the latest Wagtail to be reviewed.

thibaudcolas:
Reviewed with the core team – this needs a better explanation of the problem, and likely a new approach with refactorings over the last 4 years.

Suggestion from @laymonage: revisit this after we’ve further migrated the documents support towards generic views.

Assigning to @gasman to check whether this might have already been addressed

gasman:
I _think_ this may have been solved independently in https://github.com/wagtail/wagtail/commit/b13641e9367b62eaedef85c8900bbb88efdf9354 - will take a deeper look to see if there's anything I missed.

PR #6339, opened 2020-08-19
django 3.1 deprecated django-admin.py

minusf:
Mechanical change to use `django-admin` instead of `django-admin.py` as Django 3.1 deprecated the later.

I did not change the last remaining location: `etc/uwsgi.conf.sample` because to be honest that comment makes no sense to me, but I am not a native speaker.

Announcement: https://docs.djangoproject.com/en/3.1/releases/3.1/#id2

### DIFF
--- a/docs/contributing/developing.rst
+++ b/docs/contributing/developing.rst
@@ -100,7 +100,7 @@ You can create migrations for the test app by running the following from the Wag
 
 .. code-block:: console
 
-    $ django-admin.py makemigrations --settings=wagtail.tests.settings
+    $ django-admin makemigrations --settings=wagtail.tests.settings
 
 
 Testing against PostgreSQL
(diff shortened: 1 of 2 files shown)


kaedroho (approved):

kaedroho:

> I did not change the last remaining location: etc/uwsgi.conf.sample because to be honest that comment makes no sense to me, but I am not a native speaker.

This comment makes no sense to me too, but it's just a comment and this wouldn't change the meaning of it anyway so happy to leave this out

allcaps (approved):

gasman:
Thanks @minusf! I think the comment in uwsgi.conf.sample is meant to say "Set $PYTHONPATH so that 'django-admin.py' _works_"... I'll update that when I merge.

Issue #5719, opened 2019-11-25
Replace template references to settings.BASE_URL with template tag call


thenewguy:
A hardcoded string reference to the settings.BASE_URL isn't very flexible and is problematic for cleanly serving two domains from one running codebase.

There is a template tag [base_url_setting()](https://github.com/wagtail/wagtail/blob/master/wagtail/admin/templatetags/wagtailadmin_tags.py#L199) that could be used instead.

It would also be nice if this template tag would accept a callable for a dynamic value.

gasman:
Hi @thenewguy - can you clarify which particular templates you're referring to?

thenewguy:
@gasman I hit the issue when attempting to preview page changes.  Then I did a search against the project code for `settings.BASE_URL`.  Unfortunately, I did not keep the list.  After letting this roll around in the back of my head for a few days, I am leaning towards using a custom object for  `settings.BASE_URL` that converts to the appropriate string value.  It won't have access to the request, which is probably generally useful, but I think I can get around that for my needs

gasman:
OK - will close, as it sounds like isn't something for Wagtail itself to deal with.

PR #6341, opened 2020-08-20
Force string url before passing to urlparse

thenewguy:
I am hitting an issue that I cannot really work around as this setting needs to change during runtime.

urlparse assumes input is either a str or bytes.  If it isn't str, it assumes bytes and calls the decode method.  The problem is this sets the return value to bytes.  This then breaks at the string concatenation line.

Similar to https://github.com/wagtail/wagtail/issues/5719 but due to this problem, providing an object that changes when cast doesn't work.

### DIFF
--- a/wagtail/api/v2/tests/tests.py
+++ b/wagtail/api/v2/tests/tests.py
@@ -1,6 +1,5 @@
 from unittest import TestCase
 
-from django.conf import settings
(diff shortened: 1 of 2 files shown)


kaedroho (approved):
LGTM! I'm not sure if this is the best way to support dynamic base paths though, but this is a really small change so happy to merge it, thanks!

PR #6349, opened 2020-08-24
WIP: Css variable color theming

JNaftali:
re: https://github.com/wagtail/wagtail/issues/2943#issuecomment-655568497

An taste of what it would look like to refactor the WagtailAdmin styles to use css variables instead of sass variables. Just open up wagtail admin in your browser + add a new css rule to the `:root` selector - `--color-primary-hue: 25`. Change that number and watch all the header and button colors update in real time.

Things I did and did not do:
* I renamed `$color-teal` to `--color-primary` because the whole point of this is it might not always be teal
* I added `postcss-custom-properties` which means that the wagtail admin should still work in Internet Exporer (although custom themes will not)
* I defined some sass functions/mixins to make working with colors in css variables easier

Things left to discuss:
* Adding `postcss-custom-properties` meant adding postcss to gulp. Some other parts of the gulp task (autoprefixer and cssnano) could maybe be done by postcss as well, which might reduce the amount of time it takes for sass compilation to occur during dev. I tried this and I couldn't get it to work - some incompatibility between cssnano and postcss

### New sass mixins/functions I made.
I managed to both keep the relationships between lighter/darker variants of the same color AND avoid the syntax getting too ugly. To do so, I made these, many of which operate on tuples of 3 css variables (which represent hsl colors). Putting them all together, the variable declarations look like this:
```scss
:root {
    @include define-color('primary', #007d7e);
    @include define-color('primary-darker', css-darken(css-adjust-hue(get-color('primary'), 1), 4%));
}

button {
    // can be referenced as normal css variables
    background: var(--color-primary);

    $:hover {
        // can be gotten as a tuple if they need to be transformed inline
        background: hsl( css-darken( get-color('primary'), 10 )... );
    }
}
```

#### `define-color` mixin
Takes a `$name` and a `$color` arguments and set the following css variables: `--color-#{name}-hue`, `--color-#{name}-saturation`, `--color-#{name}-lightness`, and `--color-#{name}`. The last variable passes the other 3 variables to the native css hsl function. `$color` can be either a sass color or a tuple of hsl values

#### `get-color` function
Given the `$name` of a color (e.g. `'white'`) it returns a tuple with the variables as defined  by the above mixins (e.g. `var(--color-white-hue), var(--color-white-saturation), var(--color-white-lightness)`). Useful when you need to transform a color into another color

#### `css-darken` function,  with equivalent for `lighten`, `saturate`, `desaturate`, and `adjust-hue`
Each function works the same as the sass equivalent except that it takes and returns an hsl tuple instead of a sass color

#### `css-transparentize` function
Takes an hsl tuple, returns a color (expressed as `hsla`). Does not return a tuple.

Important housekeeping yet to be done:
- [ ] Update docs
- [ ] Test on all browsers
- [ ] Possibly update style guide?


lb-:
Quick question, is there a way for these to be compiled via SASS for older browser support?
From what I can see, this will not work for IE 11 (which we still need to [support in Wagtail](https://docs.wagtail.io/en/stable/contributing/developing.html?#browser-and-device-support))
https://caniuse.com/#feat=css-variables

Also, for tracking we will mark this WIP PR as a draft (CI will still run on draft PRs).

JNaftali:
@lb- I forgot draft PRs were a thing or else I'd have done that myself 😆 thanks.

Tldr; the admin will not be theme-able in IE, but it will continue working like it does today. You can test the squash preview, it should already work in IE.

As for IE support, that's what `postcss-custom-properties` does. If you open the Squash preview and inspect (for example) the `<header>` element you'll see duplicate declarations for `color` and `background-color` - `background-color: #007f80` and `background-color: var(--color-header-bg)`. The hard coded values are the default wagtail colors as calculated by the postcss plugin at build time. Wherever custom properties aren't supported, those declarations will be ignored. Wherever they ARE supported, the declarations that use custom properties will override the hard coded ones.

lb-:
Awesome, I didn't know that about postcss custom properties, thanks for clarifying.

JNaftali:
@lb- Hello! I did some more work on this and am fairly happy with where it is. Is there anything I could do to get a review and/or start a discussion about actually merging this?  Obviously I understand that this is not on anyone's priority list, but I'd really appreciate some guidance on how to move forward.

wagtail/images/static_src/wagtailimages/scss/add-multiple.scss
```diff
@@ -94,7 +94,7 @@
         width: 1em;
         font-size: 10em;
         line-height: 1.4em;
-        color: lighten($color-grey-4, 4%);
+        color: hsl(css-lighten(get-color('grey-4'), 4%)...);
```
thibaudcolas, line 97: I’ve never seen this syntax before, can you explain what’s going on here?

wagtail/admin/static_src/wagtailadmin/scss/layouts/compare-revisions.scss
```diff
@@ -1,13 +1,16 @@
...
-$color-addition-light: #ebffeb;
-$color-deletion-dark: #f8cbcb;
-$color-deletion-light: #ffebeb;
+
+:root {
+    --color-addition-dark: #a6f3a6;
+    --color-addition-light: #ebffeb;
+    --color-deletion-dark: #f8cbcb;
+    --color-deletion-light: #ffebeb;
+}
```
thibaudcolas, line 9: I don’t really see the point in those colors being theme-able? At least to me they shouldn’t be defined at the root level, we don’t want them reused elsewhere.

client/src/components/Explorer/Explorer.scss
```diff
@@ -1,12 +1,15 @@
...
 $menu-footer-height: 50px;
 
 @import 'ExplorerItem';
 
+:root {
+    --c-explorer-bg: #4c4e4d;
+    --c-explorer-bg-dark: var(--nav-grey-1);
+    --c-explorer-bg-active: rgba(0, 0, 0, 0.425);
+    --c-explorer-secondary: #a5a5a5;
+}
```
thibaudcolas, line 11: I don’t think this should be defined at the root level, we don’t want any other component to reuse the explorer’s colors.

client/src/components/Draftail/Draftail.scss
```diff
@@ -26,6 +22,13 @@ $draftail-editor-font-family: $font-serif;
 @import './blocks/ImageBlock';
 @import './blocks/EmbedBlock';
 
+:root {
+    --draftail-editor-text: var(--color-text-input);
+    --draftail-editor-chrome: var(--color-white);
+    --draftail-editor-chrome-text: var(--color-grey-2);
+    --draftail-editor-chrome-active: var(--color-grey-2);
+}
```
thibaudcolas, line 30: I would prefer for these not to be defined at the root level.

client/scss/settings/_variables.scss
```diff
@@ -106,21 +46,153 @@ $draftail-editor-z-index: $nav-wrapper-inner-z-index + 1;
...
 
-// Nav search
-$nav-search-color: darken($color-white, 20);
-$nav-search-border: darken($color-white, 40);
-$nav-search-bg: $nav-grey-1;
-$nav-search-hover-bg: $nav-item-hover-bg;
-$nav-search-focus-color: $color-white;
-$nav-search-focus-bg: $nav-item-hover-bg;
+
+@function css-darken($hsl-tuple, $darken-by) {
```
thibaudcolas, line 54: This and all other functions below shouldn’t be in `variables.scss`, they need their own file.

client/scss/settings/_variables.scss
```diff
@@ -106,21 +46,153 @@ $draftail-editor-z-index: $nav-wrapper-inner-z-index + 1;
...
+    --color-#{$name}-saturation: #{$s};
+    --color-#{$name}-lightness: #{$l};
+    --color-#{$name}: hsl(#{ var(--color-#{$name}-hue), var(--color-#{$name}-saturation), var(--color-#{$name}-lightness) });
+}
+
+@function get-color($name) {
+    @return (var(--color-#{$name}-hue), var(--color-#{$name}-saturation), var(--color-#{$name}-lightness));
+}
+
+:root {
```
thibaudcolas, line 117: `variables.scss` shouldn’t contain any CSS styles, otherwise they will be duplicated wherever it’s imported. It should be pure Sass exports only.

client/scss/settings/_variables.scss
```diff
@@ -106,21 +46,153 @@ $draftail-editor-z-index: $nav-wrapper-inner-z-index + 1;
...
+    // colours
+    // Defines: --color-primary-hue, --color-primary-saturation, --color-primary-lightness, and --color-primary
```
thibaudcolas, line 119: This is a general point, not specifically about this comment – do we really need to create those four separate variables? It feels to me like it would be better to just refactor existing colors to not rely on modifying the darkness/lightness/saturation. Otherwise we end up with 20 shades that are all slightly different.

I can see how the above would be useful for transparency, but that’s about it?

thibaudcolas (changes requested):
Thank you for taking a stab at this @JNaftali! I haven’t tried this in a browser yet but it looks very promising.

It also looks like a gigantic refactoring that’s unlikely to be merged in one go if it’s so big though. I have two things I’d like you to consider:

First, would it be possible to do this for the teal / primary color only? Or perhaps teal / primary color, and the shades of grey? Split this into two PRs (or more), so we can at least take the colors people are most likely to want to customize all the way through, and that will make it much easier to review and QA.

It’s also not clear to me that we do want all colors to be theme-able with that much granularity – it feels like ideally as a user who makes a theme, you’d want to set the primary color, maybe a couple of highlights here and there, and that should just trickle down to the rest of the UI without you having to override each and every value. Wagtail’s primary color is somewhat arbitrary, but the grey scale, state, warning, error colors – have all been picked so they work well together.

---

For your approach – I was assuming we wouldn’t have to rewrite that much code. You essentially did this across the codebase:

```scss
// Before
$color-teal: #007d7e;
$color-teal-darker: darken(adjust-hue($color-teal, 1), 4);
$color-teal-dark: darken(adjust-hue($color-teal, 1), 7);

// After
:root {
    @include define-color('primary', #007d7e);
    @include define-color('primary-darker', css-darken(css-adjust-hue(get-color('primary'), 1), 4%));
    @include define-color('primary-dark', css-darken(css-adjust-hue(get-color('primary'), 1), 7%));
```

And then we have to rewrite all code that uses those variables:

```diff
    &:hover {
-        color: $color-teal;
+        color: var(--color-primary);
    }
```

I was assuming we’d be able to keep using Sass variables:

```scss
// Before
$color-teal: #007d7e;
$color-teal-darker: darken(adjust-hue($color-teal, 1), 4);
$color-teal-dark: darken(adjust-hue($color-teal, 1), 7);

// After
$color-primary: var(--color-primary);
$color-primary-darker: var(--color-primary-darker);
$color-primary-dark: var(--color-primary-dark);

:root {
    @include define-color('primary', #007d7e);
    @include define-color('primary-darker', css-darken(css-adjust-hue(get-color('primary'), 1), 4%));
    @include define-color('primary-dark', css-darken(css-adjust-hue(get-color('primary'), 1), 7%));
}
```

…so we wouldn’t have to change all existing code using Sass variables, unless we decided to rename them as well. Does that seem workable?

wagtail/images/static_src/wagtailimages/scss/add-multiple.scss
JNaftali: Assuming you mean the `...`, that's spreading a list of 3 values onto the 3 arguments of sass's hsl function. Can't find the sass documentation but it seems to work the same as [the JS equivalent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax). More detailed explanation below in case that's not what you meant.

`get-color` (defined [here](https://github.com/JNaftali/wagtail/blob/e17455a4866f663f76be4f30eb5247d604e389e6/client/scss/settings/_variables.scss#L113-L115)) returns a list of 3 css variables (in this case `var(--color-grey-4-hue)` followed by the equivalent for saturation and lightness)
`css-lighten` (defined [here](https://github.com/JNaftali/wagtail/blob/e17455a4866f663f76be4f30eb5247d604e389e6/client/scss/settings/_variables.scss#L60-L65)) takes that list and returns a new list, replacing the last element (`var(--color-grey-4-lightness)`) with a calc statement that lightens it (`calc( var(--grey-4-lightness) - 4% )`). In effect the same as the sass lighten function that it's replacing, except the calculation happens at runtime (in the browser) not at build time.
`hsl` is the sass hsl function. It takes 3 arguments, so the `...` spreads the list of 3 values returned by `css-lighten` out onto those 3 arguments.

wagtail/admin/static_src/wagtailadmin/scss/layouts/compare-revisions.scss
JNaftali: I went about this with the goal of replacing every sass variable with css variables, and the postcss polyfill plugin requires that all css variables be defined inside `:root`. We could explore other plugins that don't have that restriction (although they have other footguns).

My thinking in converting these was that if someone were to make a dark theme or some other more complete and ambitious style override they might need to override these colors as well.

client/src/components/Explorer/Explorer.scss
JNaftali: see https://github.com/wagtail/wagtail/pull/6349#discussion_r494494951 re: restrictions coming from the postcss plugin

client/scss/settings/_variables.scss
JNaftali: It's really nice being able to change one color and have other colors change in coordination with it... but ditching that that it would certainly make all this MUCH simpler.

JNaftali:
> First, would it be possible to do this for the teal / primary color only? Or perhaps teal / primary color, and the shades of grey? Split this into two PRs (or more), so we can at least take the colors people are most likely to want to customize all the way through, and that will make it much easier to review and QA.

Definitely possible! I can make a separate PR for just the primary color.
 
> It’s also not clear to me that we do want all colors to be theme-able with that much granularity – it feels like ideally as a user who makes a theme, you’d want to set the primary color, maybe a couple of highlights here and there, and that should just trickle down to the rest of the UI without you having to override each and every value. Wagtail’s primary color is somewhat arbitrary, but the grey scale, state, warning, error colors – have all been picked so they work well together.

Changing anything besides the primary color would definitely be a lot of work for anyone building a theme, but still more feasible than overriding all the styles individually. 

> I was assuming we’d be able to keep using Sass variables:
> 
> ```css-scss
> // Before
> $color-teal: #007d7e;
> $color-teal-darker: darken(adjust-hue($color-teal, 1), 4);
> $color-teal-dark: darken(adjust-hue($color-teal, 1), 7);
> 
> // After
> $color-primary: var(--color-primary);
> $color-primary-darker: var(--color-primary-darker);
> $color-primary-dark: var(--color-primary-dark);
> 
> :root {
>     @include define-color('primary', #007d7e);
>     @include define-color('primary-darker', css-darken(css-adjust-hue(get-color('primary'), 1), 4%));
>     @include define-color('primary-dark', css-darken(css-adjust-hue(get-color('primary'), 1), 7%));
> }
> ```
> 
> …so we wouldn’t have to change all existing code using Sass variables, unless we decided to rename them as well. Does that seem workable?

I experimented with something like that for a bit but didn't like how it played out - it was no longer obvious which `$color` variables could safely be passed to sass functions like `darken` or `transparentize`, and it would mean that adding/changing color variants had to be coordinated across 2 different files (I hate stuff like that). Definitely workable if that's preferable to y'all.

wagtail/images/static_src/wagtailimages/scss/add-multiple.scss
thibaudcolas: Thank you, I had never seen `...` and can’t even find documentation about it on the Sass website.

wagtail/admin/static_src/wagtailadmin/scss/layouts/compare-revisions.scss
thibaudcolas: Very good point. I had completely missed that PostCSS plugin only supports colors on `:root`, that’s annoying.

client/scss/settings/_variables.scss
thibaudcolas: Yeah, I’m a bit torn about this. It feels neat for people to be able to change one "root" color and have other shades derived from it, but I don’t see anyone wanting to change say only the hue.

thibaudcolas:
Thank you for the valuable replies @JNaftali, I’ll follow up on #6409 which looks equally as promising as this first draft.

PR #6353, opened 2020-08-25
Fix timesince_last_update for naive datetimes (USE_TZ=False)

gasman:
Prevent timesince_last_update template tag from failing when passed a naive datetime, which happens when USE_TZ is set to False. Fixes #6345

I'll open a separate PR against master, built on top of #6330, which adds test coverage for this by adding a new test run for USE_TZ=False (just as #6330 added a new test run for a username-less user model). In the interests of not holding up the 2.10.1 release, I propose that we don't wait for #6330 and the new PR to be reviewed in full, but they can still serve as verification that this fix is valid.

PR #6354, opened 2020-08-25
Run test suite with USE_TZ=False

gasman:
Incorporates #6330 and #6353. Add a test run with USE_TZ=False so that we can catch any future code changes that break under this configuration.


wagtail/tests/utils/wagtail_tests.py
```diff
@@ -36,6 +36,32 @@ def login(self, user=None):
...
+
+        kwargs['email'] = email or "%s@example.com" % username
+        kwargs['password'] = password
+        if User.USERNAME_FIELD != 'email':
+            kwargs[User.USERNAME_FIELD] = username
+
+        return User.objects.create_user(**kwargs)
+
+    @staticmethod
+    def create_superuser(username, email=None, password=None, **kwargs):
```
kaedroho, line 65: Could we update `create_test_user` to use this method?

wagtail/tests/customuser/fixtures/test_explorable_pages.json
```diff
@@ -0,0 +1,112 @@
+[
```
kaedroho, line 1: Probably for future PR, but is it really necessary for this test case to have its own users fixture?

wagtail/contrib/forms/tests/test_views.py
```diff
@@ -663,11 +663,11 @@ def test_list_submissions_csv_export(self):
         self.assertEqual(response.status_code, 200)
         data_lines = response.getvalue().decode().split("\n")
 
-        self.assertEqual(data_lines[0], 'Username,Submission date,Your email,Your message,Your choices\r')
+        self.assertEqual(data_lines[0], 'User email,Submission date,Your email,Your message,Your choices\r')
```
kaedroho, line 699: Should this remain the username for regular users? If not, this would be a breaking change, wouldn't it?

kaedroho (approved):
This looks great! Sorry I didn't realise this was based on another PR until I got about halfway through and hadn't seen any timezone-specific changes. So this review counts for https://github.com/wagtail/wagtail/pull/6330 as well

wagtail/contrib/forms/tests/test_views.py
gasman: I changed this in the definition of `CustomFormPageSubmission` - see https://github.com/wagtail/wagtail/commit/87bc95f34e60c112f3b2c7052c66a06a57681c0f#diff-d2d058dc2ff8fff791e5b62fdd165cbd . I figured that since the intention of this test was to make sure we could capture data from `request.user` in the submission, it didn't really matter _which_ field of the user model we captured - I could have kept it closer to the original by doing

        form_data.update({
            'username': self.user.get_username(),
        })

but then we would have needed separate cases for the subsequent assertEqual lines that check the CSV data, depending on whether CustomUser or EmailUser was active. Changing it to consistently use the email field meant that we could avoid that.

Issue #5005, opened 2019-01-23
Can modeladmin edit_handlers be bound to an instance before checking `form.is_valid()` on Create and Edit views?


frmdstryr:
This is not an issue, more of a feature request to add the ability to further customize form handling in the admin (mainly to support an InlineFormPanel for OneToOneFields & ForeignKeys).

Currently the `edit_handler.bind_to_instance` is not called before the ModelFormView subclasses (CreateView and EditView)  call `form.is_valid()` for `put` & `post` requests. 

This prevents using panels that may customize the form on a "per instance" basis (such as a which related object is chosen).

I currently have to use custom Create & Edit views like this:

```python
class BoundFormMixin:
    """ The bound form lets edit handlers customize the form
    on a per instance basis.
    """

    def get_bound_form(self):
        edit_handler = self.get_edit_handler().bind_to_instance(
            instance=self.get_instance(),
            form=self.get_form(),
            request=self.request)
        return edit_handler.form


class CreateView(BaseCreateView, BoundFormMixin):

    @transaction.atomic
    def post(self, request, *args, **kwargs):
        form = self.get_bound_form()
        if form.is_valid():
            return self.form_valid(form)
        else:
            return self.form_invalid(form)


class EditView(BaseEditView, BoundFormMixin):

    @transaction.atomic
    def post(self, request, *args, **kwargs):
        form = self.get_bound_form()
        if form.is_valid():
            return self.form_valid(form)
        else:
            return self.form_invalid(form)

```

It would be nice if this were built in. I would also like to contribute an `InlineFormPanel` to support OneToOneFields and ForeignKeys (similar to a ChooserPanel)

ababic:
Thanks for raising @frmdstryr! This makes a lot of sense, and would be super easy to implement (literally a one-line change to `ModelFormView.get_edit_handler()`). It would nicely compliment the work being done in #4900 too. 

However, we would likely have to stick to specifying `request` and `instance` only, as calling `get_form()` from within ``get_edit_handler()`` would create an infinite loop (`get_form()` calls `get_form_class()`, which calls `get_edit_handler()`)

frmdstryr:
Agree, I like the new change to use `bind_to` and #4900.

Thoughts on making [get_form](https://github.com/django/django/blob/master/django/views/generic/edit.py#L29-L33) return a "bound form" instead of in `get_edit_handler`?  

[...]

PR #6355, opened 2020-08-25
Clean ModelAdmin form only once at creation and edit

jeromelebleu:
The same form instance is used during the whole create and edit view process to ensure the form is cleaned only once. This will improve performances when form cleaning requires extra database queries or external requests for example.

### DIFF
--- a/wagtail/contrib/modeladmin/tests/test_simple_modeladmin.py
+++ b/wagtail/contrib/modeladmin/tests/test_simple_modeladmin.py
@@ -266,6 +266,13 @@ class TestCreateView(TestCase, WagtailTestUtils):
                 self.assertTrue(mock_form_fields_exclude.called)
                 m.assert_called_with(Book, exclude=mock_form_fields_exclude.return_value)
 
+    def test_clean_form_once(self):
+        with mock.patch('wagtail.tests.modeladmintest.wagtail_hooks.PublisherModelAdminForm.clean') as mock_form_clean:
+            response = self.client.post('/admin/modeladmintest/publisher/create/', {'name': ''})
+            self.assertEqual(response.status_code, 200)
+
+            mock_form_clean.assert_called_once()
+
 
 class TestInspectView(TestCase, WagtailTestUtils):
     fixtures = ['modeladmintest_test.json']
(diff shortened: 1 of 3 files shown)


ababic:
Hi @jeromelebleu! Thanks very much for the submission. I've had a quick look, and it feels like the discussion on #5005 is very much relevant here. `get_context_data()` should definitely reuse the `form` instance if one is passed to it, though we also need to change the order of events a little and reduce the number of edit_handler instances that are created overall. I think these changes might do the trick...

```
class ModelFormView(WMABaseView, FormView):

    def get_form_class(self):
        self.edit_handler = self.get_edit_handler().
        return self.edit_handler.get_form_class()

    def get_form(self):
        form = super().get_form()
        self.edit_handler = self.edit_handler.bind_to(form=form)
    
    def get_context_data(self, form=None, **kwargs):
        if form is None:
            form = self.get_form()
        kwargs.update({
            'is_multipart': form.is_multipart(),
            'edit_handler': self.edit_handler,
            'form': form,
        })
        return super().get_context_data(**kwargs)
```

jeromelebleu:
Hi @ababic, actually I had seen this issue but didn't really know how to address it at the same time, so thanks a lot for your review and this much better solution. Is it better to close this PR and let you commit these changes in order to fix this issue?

By the way, I am wondering if something similar should be done for `instance` in case of a creation in order to use the same object in `get_edit_handler()` and `get_form_kwargs`?

ababic:
You're welcome @jeromelebleu! 

> Is it better to close this PR and let you commit these changes in order to fix this issue?

It would probably be tidier to open up a separate PR referencing this one, so that the back-story isn't lost (unless you want to rebase/force push this branch - it's completely up to you).

> By the way, I am wondering if something similar should be done for instance in case of a creation in order to use the same object in get_edit_handler() and get_form_kwargs()?

I wouldn't worry about this for `ModelAdmin`. For some reason, the instance is identified on view initialisation, and has already been added as `self.instance` to the view by the time anything is done with it (`get_instance()` will always return that same attribute value) .

jeromelebleu:
@ababic I let you have a look and review #6362!

PR #6356, opened 2020-08-26
tweak Japanese translation

snj:
The current translation is not very suitable for most Japanese users.


marteinn:
Hi @snj! Thanks for your contribution. As per the README, translations should be submitted through Transifex. For Japanese, go to https://www.transifex.com/torchbox/wagtail/language/ja/. (We then sync translations from Transifex on a regular basis).

snj:
@marteinn OK! I'll close this PR. thank you for the reply.

PR #6362, opened 2020-08-27
Instantiate edit_handler and form once in ModelAdmin views

### LINKED ISSUE #5005: Can modeladmin edit_handlers be bound to an instance before checking `form.is_valid()` on Create and Edit views?

jeromelebleu:
Replace #6355 with @ababic changes proposal to fix #5005 by the way.

### DIFF
--- a/wagtail/contrib/modeladmin/views.py
+++ b/wagtail/contrib/modeladmin/views.py
@@ -108,6 +108,14 @@ class WMABaseView(TemplateView):
 
 
 class ModelFormView(WMABaseView, FormView):
+    def setup(self, request, *args, **kwargs):
+        super().setup(request, *args, **kwargs)
+        self.edit_handler = self.get_edit_handler()
+
+    def get_form(self):
+        form = super().get_form()
+        self.edit_handler = self.edit_handler.bind_to(form=form)
+        return form
 
     def get_edit_handler(self):
         edit_handler = self.model_admin.get_edit_handler(
@@ -116,7 +124,7 @@ class ModelFormView(WMABaseView, FormView):
         return edit_handler.bind_to(model=self.model_admin.model)
 
     def get_form_class(self):
-        return self.get_edit_handler().get_form_class()
+        return self.edit_handler.get_form_class()
 
     def get_success_url(self):
         return self.index_url
@@ -136,15 +144,12 @@ class ModelFormView(WMABaseView, FormView):
             js=self.model_admin.get_form_view_extra_js()
         )
 
-    def get_context_data(self, **kwargs):
-        instance = self.get_instance()
-        edit_handler = self.get_edit_handler()
-        form = self.get_form()
-        edit_handler = edit_handler.bind_to(
-            instance=instance, request=self.request, form=form)
+    def get_context_data(self, form=None, **kwargs):
+        if form is None:
+            form = self.get_form()
         context = {
             'is_multipart': form.is_multipart(),
-            'edit_handler': edit_handler,
+            'edit_handler': self.edit_handler,
             'form': form,
         }
         context.update(kwargs)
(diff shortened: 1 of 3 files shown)


wagtail/contrib/modeladmin/views.py
```diff
@@ -136,19 +144,15 @@ def media(self):
...
-        instance = self.get_instance()
-        edit_handler = self.get_edit_handler()
-        form = self.get_form()
-        edit_handler = edit_handler.bind_to(
-            instance=instance, request=self.request, form=form)
-        context = {
+    def get_context_data(self, form=None, **kwargs):
+        if form is None:
+            form = self.get_form()
+        kwargs.update({
```
ababic, line 150: I think we need to keep the previous pattern here of defining a dictionary, then updating it with `kwargs` - otherwise, subclasses lose the ability to override these values by providing alternative values using `kwargs`.

ababic (changes requested):
Thanks a lot @jeromelebleu. This is looking really good! just one small change request from me, and we should be good :)

wagtail/contrib/modeladmin/views.py
jeromelebleu: I tried to keep that in mind actually and I think it is still possible at some point:
* `form`: an alternative value can still be defined in keyword arguments of `get_context_data` - and it will now only be created if its value is None.
* `is_multipart`: as it is directly linked to `form`, I am not sure if it is safe to override this value here.
* `edit_handler`: it is indeed not possible anymore to override its value in `get_context_data`. However, one can now instead provide an alternative value either in `get_edit_handler` or in `edit_handler` object's property. I think providing an alternative value only in `get_context_data` should be discouraged since this new possibilities ensure at least consistency. What do you think?

ababic: @jeromelebleu This is more about backwards compatibility than anything - Currently, if a user was doing something like the following, it would no longer work as expected:

```
MyEditView(EditView):
    def get_context_data(self, **kwargs):
        return super().get_context_data(
            edit_handler=some_alternative_handler
        )
```

There are a lot of things that vary throughout wagtail, but this pattern is something we have tried to stick to elsewhere when it comes to `get_context_data()` - it would be good to keep this element as it was for consistency.

jeromelebleu: Thanks for this explanations, I understand your point and totally agree in fact with the backward compatibility argument. I have made requested changes!

ababic:
@jeromelebleu Thank you kindly for time time and effort on this. Merged in 80cc70b

jeromelebleu:
@ababic Sorry, I was away from my keyboard, thanks for the fix and the merge!

PR #6363, opened 2020-08-27
Add note about RoutablePageMixin not calling get_context et al

### LINKED ISSUE #6346: get_context() not being called for routable page
### Issue Summary

I'm creating a blog as part of a client's site.  I have created a `BlogPage` and set it up with `RoutablePageMixin`.  Different routes render with different templates (`blog-posts-listing.html`  and  `blog-post.html`).  In the footer of each template I want to show links to the blog archives by year, so I set up `BlogPage`'s `get_context()` method to return this information.  When I render any route for the page, the `get_context()` method is never called.  I have confirmed that this issue can be reproduced as described on another Wagtail project that uses `RoutablePageMixin`, and I have confirmed that `get_context()` **IS** called when the page is rendered by setting its `template =xxxxxxxxxxx` attribute and removing the Base Route noted below.

### Steps to Reproduce (models)

```
class BlogPost():
	title = models.CharField(_("Title"), max_length=255, default='')
	slug = models.SlugField(_("Slug"), max_length=255, unique=True, db_index=True, blank=True)
	content = RichTextField(default='', features=['italic', 'ol', 'ul', 'undo', 'redo', 'link', 'document-link'])
	image = models.OneToOneField('wagtailimages.Image', null=True, blank=True, on_delete=models.SET_NULL, related_name='+',  help_text='Image for post (optional)')
        date_created = models.DateTimeField(auto_now_add=True, editable=False)

	def __str__(self):
		return self.title

	panels = [
		MultiFieldPanel([
			FieldPanel('title'),
			ImageChooserPanel('image'),
			FieldPanel('content'),
		])		
	]


class BlogPage(RoutablePageMixin, Page):

	def get_context(self, request):
		context = super().get_context(request)
		posts_by_year = BlogPost.objects.all().annotate(year=TruncYear).values('date_created').annotate(Count('items'))
		context['posts_by_year'] = posts_by_year
		return context

        # Base Route
	@route(r'^$')
	def posts(self, request, name='post_list'):
		blog_posts = BlogPost.objects.all()
		page = request.GET.get('page', 1)
		paginator = Paginator(blog_posts, 12)
		try:
			blog_posts = paginator.page(page)
		except PageNotAnInteger:
			blog_posts = paginator.page(1)
		except EmptyPage:
			blog_posts = paginator.page(paginator.num_pages)
		return render(request, 'blog-posts-listing.html', {
			'page': self,
			'blog_posts': blog_posts
		})

	@route(r'^(?P<slug>[\w-]+)/$', name='single_post')
	def single_post(self, request, slug):
		post = get_object_or_404(BlogPost, slug=slug)
		return render(request, 'blog-post.html', {
			'page': self,
			'post': post,
		})	

	@route(r'^year/(\d+)/$')
	@route(r'^year/current/$')
	def posts_for_year(self, request, year=None):
		if year is None:
			year = datetime.date.today().year

		blog_posts = BlogPost.objects.filter(date_created__year=year)

		return render(request, 'blog-post-list.html', {
			'page': self,
			'blog_posts': blog_posts
		})
```

**Caveat:**  The code for `posts_by_year` is untested.

### Why I Think This is a Bug
The [documentation here](https://docs.wagtail.io/en/v2.10/topics/pages.html#more-control-over-page-rendering) says:

> All page classes have a serve() method that internally calls the get_context and get_template methods and renders the template.

I have verified that the `serve()` method is, in fact, called for each routable page route.  While I can see that `serve()` should not call `get_template()` for routable page routes, is there any reason `serve()` should not call `get_context()` for routable page routes?  Based upon the documentation above, I was thinking that `get_context()` **WOULD** be called.  If it is not useful to consider modifying Wagtail to call `get_context()` for routable page routes, would a note in the documentation about this behavior be in order?  I don't know if this is really a bug, but to me it certainly was unexpected behavior.

### Technical details

* Python version: 3.6
* Django version: 3.1
* Wagtail version: 2.10
* Browser version: N/A

gasman:
Fixes #6346. The relationship between RoutablePageMixin and the standard Wagtail page serving chain-of-command is a frequent point of confusion (see #6062 for example) - hopefully this clears things up a bit.

### DIFF
--- a/docs/reference/contrib/routablepage.rst
+++ b/docs/reference/contrib/routablepage.rst
@@ -32,6 +32,10 @@ The basics
 
 To use ``RoutablePageMixin``, you need to make your class inherit from both :class:`wagtail.contrib.routable_page.models.RoutablePageMixin` and :class:`wagtail.core.models.Page`, then define some view methods and decorate them with ``wagtail.contrib.routable_page.models.route``. These view methods behave like ordinary Django view functions, and must return an ``HttpResponse`` object; typically this is done through a call to ``django.shortcuts.render``.
 
+.. note::
+   Just like a Django view function, a view method is responsible for the end-to-end process of taking a request and returning a corresponding response - therefore, methods that would normally play a part in serving a standard Wagtail page, such as ``get_context`` and ``get_template``, will be skipped (unless the view method explicitly calls them). This also means that other modules which customise the page serving behaviour, such as the :ref:`form_builder`, will generally not work in conjunction with ``RoutablePageMixin``.
+
+
 Here's an example of an ``EventIndexPage`` with three views, assuming that an ``EventPage`` model with an ``event_date`` field has been defined elsewhere:
 
 .. code-block:: python


ababic:
@gasman Crazy idea: How about moving the following code out of `Page.serve()` into a new `Page.render()` method, then updating the example to show that being used?

```
    return TemplateResponse(
            request,
            self.get_template(request, *args, **kwargs),
            self.get_context(request, *args, **kwargs)
        )
```

gasman:
@ababic Given that the whole point of RoutablePage is to do different things for different URL endpoints, I think the usefulness of funnelling those different functions into a shared render / get_context method is a bit debatable - if we steer people towards that as best practice, they'll end up wanting to know how to pass in extra variables (or use an alternative template) for this one view.

Any abstractions / helpers that we demonstrate in the docs will tend to lead to people using those things unquestioningly, whether or not it actually makes sense for their use-case. I think it's more helpful to tell (and show) people that these view methods work just like Django views, and then they can build their own abstractions on top of that if they feel that's appropriate.

ababic:
@gasman Okay, I see your point, but perhaps the example help there by also showing a custom route doing something completely different?

gasman:
@ababic If you can think of something, feel free to open a PR :-)

gasman:
Superseded by #6370

PR #6364, opened 2020-08-28
Add note about programmatically creating categories

PresidentNick:

### DIFF
--- a/docs/getting_started/tutorial.rst
+++ b/docs/getting_started/tutorial.rst
@@ -877,6 +877,8 @@ Finally, we can update the ``blog_page.html`` template to display the categories
 .. figure:: ../_static/images/tutorial/tutorial_10.jpg
    :alt: A blog post with categories
 
+To programmatically add a category to a page, refer to the django documentation for `adding related objects <https://docs.djangoproject.com/en/dev/ref/models/relations/#django.db.models.fields.related.RelatedManager.add>`_
+
 
 Where next
 ----------


gasman:
Thanks for the suggestion @PresidentNick! However, I don't think the tutorial is the right place for this - adding categories programmatically is a fairly specialist task that won't come up for the majority of users, and given that the tutorial is already introducing a lot of new concepts, we don't want to overwhelm new users with information that doesn't apply to them.

Working with Wagtail data programmatically is also a big topic to cover in full - a user wanting to do this would need to understand the appropriate Django ORM methods, but depending on what they're trying to achieve, they might also need to know about the distinction between saving and publishing, working with the page tree, management commands, data migrations and many other things. Pointing them to the documentation for just one aspect of this information is liable to leave them more lost than they were before. Perhaps this calls for a new sub-section of the documentation?

PresidentNick:
Those are all good points I hadn't considered! I've been working on a django management command that loads data into a wagtail instance and I am doing lots of programmatic creation so I could include these notes on a sub-section like you said. What exactly does a subsection look like in the documentation? Do you have a recommendation of where I should put it?

gasman:
I think it would belong under 'advanced topics' - creating a new `.rst` page within [docs/advanced_topics](https://github.com/wagtail/wagtail/tree/master/docs/advanced_topics) and adding an entry to index.rst would do it.

PR #6366, opened 2020-08-29
Specify required python versions

vsalvino:
I was recently bitten by the fact that you *can* install wagtail 2.10 in Python 3.5, it just breaks when you try to use it.

This change adds the `python_requires` argument in setup.py to properly specify the compatible versions for the pip package manager. It might be a good idea to also update previous wagtails (2.9 at least) to also include this with their respective python supports

See: https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -115,12 +115,12 @@ https://github.com/wagtail/wagtail/.",
         'Framework :: Wagtail',
         'Topic :: Internet :: WWW/HTTP :: Site Management',
     ],
+    python_requires='>=3.6',
     install_requires=install_requires,
     extras_require={
         'testing': testing_extras,
         'docs': documentation_extras
     },
-    python_requires='>=3.6, <3.9',
     entry_points="""
             [console_scripts]
             wagtail=wagtail.bin.wagtail:main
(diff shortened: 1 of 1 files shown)


allcaps:
Just as a reference, this is how Django handles Python version check.

https://github.com/django/django/blob/master/setup.py#L8

vsalvino:
Yes, they have that warning for users with older versions of pip that do not understand the `python_requires` attribute. Django also uses the same attributed I am proposing. See: https://github.com/django/django/blob/3d16496037fbb8a6bbc6b6e354fa4f5eb65e6cea/setup.cfg

allcaps (approved):
LGTM

gasman:
Thanks @vsalvino! I'd be in favour of removing the upper bound - it's pretty rare that a new Python version breaks compatibility, and I think the usefulness of being able to run against a newer Python version (even if it's not officially supported yet) outweighs the small chance of it catching a user error.

allcaps (approved):

gasman:
Merged in 7ec4b516fa087b32832bfc55a4237dae0a468082

Issue #6340, opened 2020-08-20
Unreadable focused button label in admin


jsonn:
### Issue Summary

Buttons in the admin interface in the default theme use white on dark green text. When the button is focused using tab, this changes to white on white. This makes the text unreadable.

### Steps to Reproduce

1. Start a new project and add a superuser, runserver
2. Go to admin / document / add document
3. Click on the tags input, hit tab

Observe the submit changing from readable "UPDATE" to a white box.

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: (yes / no)

Yes.

### Technical details

Wagtail 2.10 or master as of today. Rest is not relevant.

PR #6369, opened 2020-09-01
Fix admin submit inputs focus

karranb:
In this PR I'm trying to solve issue #6340. If there is anyone working on this, I'm sorry I did not ask previously :sweat: 

I tested this on chrome and firefox, but this affects a lot of pages and I think I might have missed some of them.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -17,6 +17,7 @@ Changelog
  * Fix: Update oEmbed endpoints to use https where available (Matt Westcott)
  * Fix: Revise `edit_handler` bind order in ModelAdmin views and fix duplicate form instance creation (Jérôme Lebleu)
  * Add support for hierarchical/nested Collections (Robert Rollins)
+ * Fix: Button focus state (Karran Besen)
 
 
 2.10.2 (xx.xx.xxxx) - IN DEVELOPMENT
--- a/client/scss/elements/_forms.scss
+++ b/client/scss/elements/_forms.scss
@@ -86,7 +86,6 @@ label,
 }
 
 input:not([type=submit]),
-input:not([type=button]),
 textarea,
 select,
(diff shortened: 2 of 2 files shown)


allcaps (approved):
Thank you @karranb for your contribution.

I dropped `input:not([type=button])` because:

1. Wagtail does not use `<input type="button" ...`. Only `<button type="button"`
2. I believe the correct way to chain `not` is `input:not([type=submit]):not([type=button])`. In the old situation the one would cancel out the other.

I leave the PR open for another set of eyes.

Browsers tested:
- IE11, Win10
- Chrome 85, OSX
- Firefox 80

karranb:
thanks for these great catches @allcaps

allcaps:
@nmorduch Can you have a look at this?

allcaps (approved):

nmorduch (approved):
I will note that comprehensive testing isn't my strongest area, but this looks good to me and I did check that `input`s with `type="submit"` still are getting from elsewhere styles for a disabled state, which they are.

Also: It looks like we _do_ use `<input type="button" ...` in the style guide ([L508](https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html#L508), [L519](https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html#L519), [L530](https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html#L530)), which we probably should stop doing given that it says on that very same same page not to do so.

allcaps:
@karranb and @nmorduch thank you! 

I opened an issue https://github.com/wagtail/wagtail/issues/6388 for the input type button.

And will merge this!

PR #6370, opened 2020-09-01
Add RoutablePage.render() to facilitate rendering of custom routes

ababic:
An alternative fix for #6346 (superseeding #6363).

The relationship between RoutablePageMixin and the standard Wagtail page serving chain-of-command is a frequent point of confusion (see #6062 for example). By offering a separate documented mechanism for rendering custom routes, and fleshing out the documentation with more varied examples, I feel we can alleviate some of this confusion


gasman (approved):
Thanks @ababic! I'm a little wary that if we don't take the hard line and say "there is no magic here - it's just a Django view function" then developers are just going to go further down the path of trying plausible-looking things from elsewhere in Wagtail (like adding form builder mixins) and being disappointed that they don't work. But maybe that's just a slippery slope argument... at this point there's enough evidence of people wanting template serving to be more "Wagtail-like" that it makes sense to pave that cowpath, and the docs here do a good job of describing exactly how much "magic" the `render` method is adding - so I'm willing to give this a go and see what effect it has on people's usage and expectations of RoutablePage.

gasman (changes requested):
Actually, just one request - please could you add a test case to `wagtail/tests/routablepage` to cover a route using the new `self.render` method?

ababic:
Thank you @gasman. Tests have now been added.

PR #6373, opened 2020-09-04
Update hooks.rst

Jean-Zombie:
This update intends to add the answer I got on my slack support request (https://wagtailcms.slack.com/archives/C81FGJR2S/p1599209092087000).

### DIFF
--- a/docs/reference/hooks.rst
+++ b/docs/reference/hooks.rst
@@ -477,6 +477,15 @@ Hooks for customising the way users are directed through the process of creating
     def do_after_page_create(request, page):
         return HttpResponse("Congrats on making content!", content_type="text/plain")
 
+  Setting attributes on a ``Page`` object requires to call ``save_revision()`` since the edit and index view pick up their data from the revisions table rather than the actual saved page record.
+  
+  .. code-block:: python
+  
+      @hooks.register('after_create_page')
+      def set_attribute_after_page_create(request, page)
+         page.title = 'Persistent Title'
+         page.save_revision()  # instead of just calling save()
+
 
 .. _before_create_page:
 
--- a/docs/reference/hooks.rst
+++ b/docs/reference/hooks.rst
@@ -477,14 +477,14 @@ Hooks for customising the way users are directed through the process of creating
     def do_after_page_create(request, page):
         return HttpResponse("Congrats on making content!", content_type="text/plain")
 
-  Setting attributes on a ``Page`` object requires to also call ``save_revision()`` since the edit and index view pick up their data from the revisions table rather than the actual saved page record.
+  If you set attributes on a Page object, you should also call save_revision(), since the edit and index view pick up their data from the revisions table rather than the actual saved page record.
   
   .. code-block:: python
   
       @hooks.register('after_create_page')
       def set_attribute_after_page_create(request, page)
          page.title = 'Persistent Title'
-         page.save_revision()  # instead of just calling save()
+         page.save_revision()
          page.save()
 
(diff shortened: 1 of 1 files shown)


docs/reference/hooks.rst
```diff
@@ -477,6 +477,15 @@ Hooks for customising the way users are directed through the process of creating
     def do_after_page_create(request, page):
         return HttpResponse("Congrats on making content!", content_type="text/plain")
 
+  Setting attributes on a ``Page`` object requires to also call ``save_revision()`` since the edit and index view pick up their data from the revisions table rather than the actual saved page record.
```
tomdyson, line 480: Setting attributes on a ``Page`` object requires to also call ``save_revision()`` since the edit -> If you set attributes on a ``Page`` object, you should also call ``save_revision()``, since the edit

docs/reference/hooks.rst
```diff
@@ -477,6 +477,15 @@ Hooks for customising the way users are directed through the process of creating
...
+  
+  .. code-block:: python
+  
+      @hooks.register('after_create_page')
+      def set_attribute_after_page_create(request, page)
+         page.title = 'Persistent Title'
+         page.save_revision()  # instead of just calling save()
```
tomdyson, line 487: i don't think the comment is necessary, as you've explained it just above.

tomdyson (commented):
Thank you @Jean-Zombie!

docs/reference/hooks.rst
Jean-Zombie: Better! I will reword it.

docs/reference/hooks.rst
Jean-Zombie: True ;-)

Jean-Zombie:
I added the proposed changes.

docs/reference/hooks.rst
tomdyson: Thanks @Jean-Zombie. You removed the double backticks around `Page` and `save_revision()` - was that deliberate? Sorry to be picky :)

Jean-Zombie: Ha, no. I copied your sentence and overlooked it. Sry for being sloppy ;-)

tomdyson (approved):

docs/reference/hooks.rst
```diff
@@ -477,6 +477,15 @@ Hooks for customising the way users are directed through the process of creating
...
 
+  If you set attributes on a ``Page`` object, you should also call ``save_revision()``, since the edit and index view pick up their data from the revisions table rather than the actual saved page record.
+  
+  .. code-block:: python
+  
+      @hooks.register('after_create_page')
+      def set_attribute_after_page_create(request, page)
+         page.title = 'Persistent Title'
+         page.save_revision()
+         page.save()
```
kaedroho, line 488: No need to call `.save()` here since the new title will already be saved. If you meant to publish the changes, you should call `.publish()` on the saved revision: `page.save_revision().publish()`

Jean-Zombie: Ah, I guess going for the `title` attribute as an example was the wrong choice. I actually tested this with an attribute set on my sub-classed model `MyPage` (i.e. inheriting from `Page`). There, my attribute does not get set if I don't call `.save()` on top of `.save_revision()`.

gasman:
Merged in b5cfca31eb4c75dc78401f6b6a310dd4b3800d0a (with the save/publish logic adjusted as per https://github.com/wagtail/wagtail/pull/6373#discussion_r484475265).

PR #6376, opened 2020-09-09
[RFC 58] Alias pages

kaedroho:
Implementation of https://github.com/wagtail/rfcs/pull/58


wagtail/core/models.py
```diff
@@ -1930,6 +1930,147 @@ def copy(self, recursive=False, to=None, update_attrs=None, copy_revisions=True,
...
+                    'page': {
+                        'id': alias.id,
+                        'title': alias.get_admin_display_title()
+                    },
+                    'source': {'id': source_parent.id, 'title': source_parent.get_admin_display_title()} if source_parent else None,
+                    'destination': {'id': parent.id, 'title': parent.get_admin_display_title()} if parent else None,
+                },
+            )
+            if alias.live:
+                # Log the publish if the use chose to keep the copied page live
```
gasman, line 2149: "if the use chose to keep the copied page live" can be deleted here, as there is no `keep_live` option.

wagtail/admin/views/pages/copy.py
```diff
@@ -51,16 +51,24 @@ def copy(request, page_id):
...
-                recursive=form.cleaned_data.get('copy_subpages'),
-                to=parent_page,
-                update_attrs={
-                    'title': form.cleaned_data['new_title'],
-                    'slug': form.cleaned_data['new_slug'],
-                },
-                keep_live=keep_live,
-                user=request.user,
-            )
+            if can_publish and form.cleaned_data['alias']:
```
gasman, line 54: It took me a while to make sense of the `can_publish` here - I think it would be helpful to add a comment here and on the form definition, along the lines of: "only users with publish permission are given the option of creating an alias, because aliases must keep the same published state as the original".

wagtail/admin/views/pages/convert_alias.py
```diff
@@ -0,0 +1,65 @@
...
+from wagtail.core.models import Page, PageLogEntry
+
+
+def convert_alias(request, page_id):
+    page = get_object_or_404(Page, id=page_id, alias_of_id__isnull=False).specific
+    if not page.permissions_for_user(request.user).can_edit():
+        raise PermissionDenied
+
+    with transaction.atomic():
+        for fn in hooks.get_hooks('before_convert_alias_page'):
```
gasman, line 19: This hook (and `after_convert_alias_page`) should be added to the docs.

gasman (changes requested):
Code all looks good to me! Just a couple of commenting / documentation suggestions...

kaedroho:
@gasman Thanks! I've addressed your comments and squashed them back into the commits above

gasman (approved):

docs/reference/hooks.rst
```diff
@@ -614,6 +614,31 @@ Hooks for customising the way users are directed through the process of creating
 
   Uses the same behaviour as ``before_create_page``.
 
+
+.. _before_convert_alias_page:
+
+``before_convert_alias_page``
+~~~~~~~~~~~~~~~~~~~~~~~~~
```
gasman, line 621: Just noticed from CI that these underlines need extending.

kaedroho: Oops. fixed!

kaedroho:
Rebased to resolve merge conflict

PR #6383, opened 2020-09-14
Add support for bulk sending notification emails

tomusher:
On sites with lots of admin notification recipients, submitted a page for moderation/in to a workflow can be quite slow as each emails is localised and sent individually.

This PR adds a new setting which sets a threshold at which Wagtail will bulk send these emails (in the BCC field) instead.

As we're sending a single email to multiple users, the emails can't be personalised or translated, so the default for this settings is set quite high (30 recipients) so it shouldn't affect sites that aren't already seeing this issue - but I'm not sure if this is something we need to think more carefully about, or turn it off altogether by default.

Appreciate if this is not the right solution to the problem - if it's too specific an issue I could separate it out in to a package and update Wagtail core to allow specifying a custom notifier, or provide a set of built-in notifiers adjustable with a setting.

### DIFF
--- a/docs/reference/settings.rst
+++ b/docs/reference/settings.rst
@@ -236,6 +236,12 @@ Notification emails are sent in `text/plain` by default, change this to use HTML
 
 Notification emails are sent to moderators and superusers by default. You can change this to exclude superusers and only notify moderators.
 
+.. code-block:: python
+
+  WAGTAILADMIN_NOTIFICATION_BULK_SEND_THRESHOLD = 30
+
+Wagtail will usually send notifications one-by-one so that the notifications can be personalised with a user's name and language which can be slow on sites with lots of recipients. This setting sets a threshold at which Wagtail will send these emails in bulk instead. Emails sent in bulk will not be personalised or translated.
+
 .. _update_notifications:
 
 Wagtail update notifications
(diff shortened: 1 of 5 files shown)


jacobtoppm (changes requested):
Hi Tom,

Thanks for contributing this! I think this could be really useful for big sites, but I have a couple of comments on the general approach:
- Translation. Generally we're trying to support internationalization more in Wagtail, and I think switching email language based on a threshold that's not visible to the user getting the emails could lead to confusion. Instead, would it be possible to group the emails and batch send by language? Admittedly this would be a smaller performance improvement in some cases, but would still be useful except when sending 30 emails where every recipient is using a different language, which is an unlikely scenario - and in cases where everyone is using the default language, which I suspect are the most common, still leads to the original improvement.
- There will be some emails that absolutely require personalisation - for example sending review tokens in Wagtail Review. I think that there should be some kind of per-notifier flag that can be used to turn off depersonalisation to cope with these scenarios. This should also probably be mentioned in the documentation.

tomusher:
Thanks for the feedback @jacobtoppm! That definitely makes sense - I was a bit unsure about this approach anyway so having that clarity is very useful. I'll take another look.

tomusher:
I think this is likely to be replaced by something async when https://github.com/wagtail/rfcs/pull/72 is implemented, so I'll close this one.

PR #5522, opened 2019-08-20
Adds `ModelAdmin.prepopulated_fields` feature  from django.admin

pySilver:
This PR brings auto slug functionality from `django.contrib.admin` described [here](https://docs.djangoproject.com/en/2.2/ref/contrib/admin/#django.contrib.admin.ModelAdmin.prepopulated_fields) into Wagtail modeladmin.

It works fine with modeladmin integrations, however I havent tested it with inlines (not sure inlines are supported within modeladmin). It obviosly won't work for snippets, but thats expected.

I'm sorry for lack of documentation and tests and I hope someone will help me the following:

* Do the tests still pass? (http://docs.wagtail.io/en/latest/contributing/developing.html#testing)
* Does the code comply with the style guide? (Run `make lint` from the Wagtail root)
* For Python changes: Have you added tests to cover the new/fixed behaviour?
* For front-end changes: Did you test on all of Wagtail’s supported browsers? **Please list the exact versions you tested**.make 
* For new features: Has the documentation been updated accordingly?

PR #6386, opened 2020-09-16
Adds `ModelAdmin.prepopulated_fields` feature from django.admin - with tests

dbramwell:

Hi,

I've taken a crack at adding tests for [#5522](https://github.com/wagtail/wagtail/pull/5522). Pretty new to python, wagtail and django, so let me know if I've missed anything or done anything daft.

I've added a small bit of documentation. There's more documentation on it in the django docs but I wasn't really sure where to wedge it in in the wagtail docs. Let me know where and I'll copy paste it somewhere.

### DIFF
--- a/wagtail/contrib/modeladmin/tests/test_simple_modeladmin.py
+++ b/wagtail/contrib/modeladmin/tests/test_simple_modeladmin.py
@@ -272,6 +272,12 @@ class TestCreateView(TestCase, WagtailTestUtils):
 
             mock_form_clean.assert_called_once()
 
+    def test_in_context(self):
+        response = self.get()
+
+        self.assertEqual(response.status_code, 200)
+        self.assertIn('data-prepopulated-fields="[{&quot;id&quot;: &quot;#id_title&quot;, &quot;name&quot;: &quot;title&quot;, &quot;dependency_ids&quot;: [&quot;#id_author&quot;], &quot;dependency_list&quot;: [&quot;author&quot;], &quot;maxLength&quot;: 255, &quot;allowUnicode&quot;: false}]"', response.content.decode('UTF-8'))
+
 
 class TestInspectView(TestCase, WagtailTestUtils):
     fixtures = ['modeladmintest_test.json']
--- a/docs/reference/contrib/modeladmin/create_edit_delete_views.rst
+++ b/docs/reference/contrib/modeladmin/create_edit_delete_views.rst
@@ -199,6 +199,31 @@ excluded from the form. This is particularly useful when registering ModelAdmin
 classes for models from third-party apps, where defining panel configurations
 on the Model itself is more complicated.
 
+.. _modeladmin_prepopulated_fields:
+
+-----------------------------------
+``ModelAdmin.prepopulated_fields``
+-----------------------------------
+
+**Expected value**: A dict mapping prepopulated fields to a tuple of fields to
+prepopulate from
+
+When using CreateView or EditView to create or update model instances, the
+fields corresponding to the keys in the dict are prepopulated using the fields
+in the corresponding tuple. The main use for this functionality is to
+automatically generate the value for SlugField fields from one or more other
+fields. The generated value is produced by concatenating the values of the
+source fields, and then by transforming that result into a valid slug (e.g.
+substituting dashes for spaces; lowercasing ASCII letters; and removing various
+English stop words such as ‘a’, ‘an’, ‘as’, and similar).
+
+Prepopulated fields aren’t modified by JavaScript after a value has been saved.
+It’s usually undesired that slugs change (which would cause an object’s URL to
+change if the slug is used in it).
+
+prepopulated_fields doesn’t accept DateTimeField, ForeignKey, OneToOneField,
+and ManyToManyField fields.
+
 
 -----------------------------------
 ``ModelAdmin.get_edit_handler()``
--- a/docs/reference/contrib/modeladmin/index.rst
+++ b/docs/reference/contrib/modeladmin/index.rst
@@ -87,6 +87,7 @@ to create, view, and edit ``Book`` entries.
 
     class Book(models.Model):
         title = models.CharField(max_length=255)
+        slug = models.SlugField()
         author = models.CharField(max_length=255)
         cover_photo = models.ForeignKey(
             'wagtailimages.Image',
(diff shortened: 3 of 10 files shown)


wagtail/tests/modeladmintest/models.py
```diff
@@ -124,3 +124,8 @@ class Friend(models.Model):
 
     def __str__(self):
         return self.first_name
+
+
+class PrepopulatedFieldModel(models.Model):
+    name = models.CharField(max_length=100)
+    slug = models.SlugField()
```
ababic, line 131: If at all possible, I'd like to avoid adding another model to the test suite to specifically test this feature. Technically, there's no reason why the target field needs to be a slug, is there? Maybe we could just take an existing model and add `prepopulated_fields` to an existing `ModelAdmin` class, with any old field being set from some others?

ababic, line 131: If at all possible, I'd like to avoid adding another model to the test suite to specifically test this feature. Technically, there's no reason why the target field needs to be a `SlugField`, is there? Maybe we could just take an existing model (like `Book`) and add `prepopulated_fields` to an existing `ModelAdmin` class, with any old field being set from some others?

wagtail/contrib/modeladmin/tests/test_simple_modeladmin.py
```diff
@@ -695,3 +695,18 @@ def test_model_with_single_tabbed_panel_and_edit_handler(self):
...
+        self.login()
+
+    def get(self, **params):
+        return self.client.get('/admin/modeladmintest/prepopulatedfieldmodel/create/', params)
+
+    def test_in_context(self):
+        response = self.get()
+
+        self.assertEqual(response.status_code, 200)
+        self.assertIn('data-prepopulated-fields="[{&quot;id&quot;: &quot;#id_slug&quot;, &quot;name&quot;: &quot;slug&quot;, &quot;dependency_ids&quot;: [&quot;#id_name&quot;], &quot;dependency_list&quot;: [&quot;name&quot;], &quot;maxLength&quot;: 50, &quot;allowUnicode&quot;: false}]"', response.content.decode('UTF-8'))
```
ababic, line 712: Instead of writing a whole new class for this feature, I feel we can get away with just adding a single test to the existing `TestCreateView` is this module (if we also go ahead and reuse the `Book` model here. Does that sound okay to you @dbramwell ?

ababic (changes requested):
Hi @dbramwell. Thanks very much for picking this up, and sorry it's taken a while to get back to you. While there's functionally nothin wrong with what you've added here, it just feels like we're making a bit too much of a special case here for what is quite a simple addition - it'd would be nice if we could utilise a bit more of what is there already.

dbramwell:
Thanks for the guidance @ababic !

I've give it a go. When I changed [this](https://github.com/wagtail/wagtail/pull/6386/files#diff-7bd1c624b692a608ef52fa74232b3472R74) I started getting an error within the two test_exclude_passed_to_extract_panel_definitions tests:

```
KeyError: "Key 'title' not found in 'BookForm'. Choices are: ."
```
I changed the list comprehension in views.py to only add pre-populated fields when the field actually exists in the form, see [here](https://github.com/wagtail/wagtail/pull/6386/commits/57b75b72483106e7dfcef8d0a024b3c25fb97ef6). Hope this is the right approach

ababic:
No problem @dbramwell. Your additions there look really good! The only thing I would change is: If any of the dependency fields aren't present in the form, I think we need to avoid including the entry all together, as it could produce some surprising behaviour for the developer (it might also be helpful to raise a warning in such cases). I have a feeling we'll need to deviate away from the double dict comprehension in order to do this.... it might even be worth moving the new code to a separate method, then calling that in `get_context_data()`? What do you think?

dbramwell:
Yeah, good idea. I've moved it out into its own function and tweaked the logic.

`it might also be helpful to raise a warning in such cases` - not really sure how to do this, is there an example somewhere else in the codebase?

ababic (changes requested):
@dbramwell. Sure thing. Here's an an example you can copy from: https://github.com/wagtail/wagtail/blob/master/wagtail/core/query.py#L447

dbramwell:
Thanks for all the help with this. I've changed to `self.model_admin.get_prepopulated_fields(self.request)`, is that just to ensure that we always have at least an empty dictionary and not `None` (or anything else)?

Have added a warning and deleted that test class also.

ababic:
Marvelous! thanks @danbentley.

> I've changed to self.model_admin.get_prepopulated_fields(self.request), is that just to ensure that we always have at least an empty dictionary and not None (or anything else)?

Yes, that ^, and also it's a pattern that WagtailModelAdmin uses quite a lot generally, so it's good for consistency - And you never really know what users want to do with a feature like this, so giving the developer the ability to return different values depending on the `request` leaves things nice and flexible.

ababic:
@dbramwell would you be okay to add some documentation to this page to cover this? It's probably enough to keep it just about the attribute, then just mention that the method can be overridden to cater for more complex cases.

dbramwell:
I've given it a go. Let me know if I've used the incorrect terminology for any of it. I'm not really sure which function you mean to mention can be overridden, [this](https://github.com/wagtail/wagtail/pull/6386/files#diff-a682530f2e1f114037f34e183048e293R163)?

ababic:
@dbramwell The documentation changes are excellent, thank you! :) 

> I'm not really sure which function you mean to mention can be overridden, this?

Sorry, I could've been clearer there. This documentation is about attritutes on the `ModelAdmin` class, so I meant `ModelAdmin.get_prepopulated_fields() `. That's not strictly necessary though, so don't worry too much.

dbramwell:
Cool, is this good to go then? Anything else I need to do before it's merged in?

ababic:
@dbramwell I think it's all good now, thanks! I just need to find a little time to run through our commit process and update the release notes etc.

dbramwell:
Awesome, that's for all your guidance with it!

ababic (approved):

dbramwell:
Do I need to fix the conflicts here?

ababic:
@dbramwell, no it's fine. I can sort those.

ababic:
Merged in 9c56dfc5d / e28aa6b19. Thanks @dbramwell! And sorry about the delay (you caught me amidst a house move)

dbramwell:
Awesome, thanks @ababic ! Enjoy the new place!

pySilver:
@dbramwell I'm glad you've done it!

PR #6387, opened 2020-09-17
deleted the div that conatined the actionbutton class in Model Admin template

Josh-Woodcock:
Deleted the div that conatined the actionbutton class in the Model Admin template

It was creating styling issues in the Wagtail admin when I was using wagtail hooks to expose a model to the admin for moderatioon and export.

Before (w/ actionbutton)

Screenshot_2020-09-17 Wagtail - Village Stories

After (without the div with actionbutton class)

Screenshot_2020-09-17 Wagtail - Village Stories(1)

I had a look at adding the same class to the 

` {% if view.list_export %}` 

however it required additional styling to create space with the button above. I figured this solution would be simplest and most effective.

### DIFF
--- a/wagtail/contrib/modeladmin/templates/modeladmin/index.html
+++ b/wagtail/contrib/modeladmin/templates/modeladmin/index.html
@@ -26,9 +26,7 @@
                 {% block header_extra %}
                     <div class="right">
                         {% if user_can_create %}
-                            <div class="actionbutton">
                                 {% include 'modeladmin/includes/button.html' with button=view.button_helper.add_button %}
-                            </div>
                         {% endif %}
                         {% if view.list_export %}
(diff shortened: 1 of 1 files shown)


jnns:
Does anything prevent this from being merged?

lb- (approved):
@Josh-Woodcock thanks so much for this contribution a long time ago.

Apologies that it did not get a review in a short time but we really appreciate the time taken to get this together.

The header layout has changed a bit so the eventual change including this is a bit different - see 8a4a6b4ceda2efc71bc6499c2d8ccc43e6140af7

Nonetheless, it should now be resolved and thanks for making your first contribution to Wagtail!

lb-:
Merged in via 8a4a6b4ceda2efc71bc6499c2d8ccc43e6140af7

PR #6389, opened 2020-09-17
Avoid 'ZZZZ' tree paths as placeholders on preview

### LINKED ISSUE #5807: PathOverflow: Path Overflow from: '**********ZZZZ'
### Issue Summary

Creating a new page from the administration area raises a PathOverflow error.

```
PathOverflow: Path Overflow from: '000100020003ZZZZ'
  File "django/core/handlers/exception.py", line 34, in inner
  File "django/core/handlers/base.py", line 126, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "django/core/handlers/base.py", line 124, in _get_response
  File "django/views/decorators/cache.py", line 44, in _wrapped_view_func
  File "wagtail/admin/urls/__init__.py", line 102, in wrapper
  File "wagtail/admin/decorators.py", line 34, in decorated_view
  File "wagtail/admin/views/pages.py", line 224, in create
    parent_page.add_child(instance=page)
  File "treebeard/mp_tree.py", line 1013, in add_child
    return MP_AddChildHandler(self, **kwargs).process()
  File "treebeard/mp_tree.py", line 387, in process
    newobj.path = self.node.get_last_child()._inc_path()
  File "treebeard/mp_tree.py", line 1114, in _inc_path
    raise PathOverflow(_("Path Overflow from: '%s'" % (self.path, )))
```

The only information I could find potentially regarding the issue was on SO at the following link: https://stackoverflow.com/questions/54166821/wagtail-pathoverflow-on-adding-new-child-page

There are only a handful of pages, so I'm not sure that particular issue applies though.

### Steps to Reproduce

I haven't been able to reproduce this myself, however a client has been seeing this behavior, which at the time I've just resolved by modifying the ZZZZ at the end to be numeric. I'm in the process of gathering some steps they may have taken to replicate it from this end (where they tried to add the page from)

### Technical details

* Python version: 3.6.8
* Django version: 2.1.4
* Wagtail version: 2.4
* Browser version: N/A

gasman:
Fixes #5807. If the page instance being previewed gets saved (no confirmation that Wagtail itself is doing this, but it could easily happen as a result of some misplaced logic in `get_context` or similar), it will block further page creations in a way that's difficult to recover from.

(no tests added, as this should not cause any functional change on a correctly configured Wagtail site)

### DIFF
--- a/wagtail/admin/views/pages/preview.py
+++ b/wagtail/admin/views/pages/preview.py
@@ -113,9 +113,15 @@ class PreviewOnCreate(PreviewOnEdit):
         # hopefully enable tree traversal operations
         # to at least partially work.
         page.depth = parent_page.depth + 1
-        # Puts the page at the maximum possible path
+        # Puts the page at the next available path
         # for a child of `parent_page`.
-        page.path = Page._get_children_path_interval(parent_page.path)[1]
+        if parent_page.is_leaf():
+            # set the path as the first child of parent_page
+            page.path = page._get_path(parent_page.path, page.depth, 1)
+        else:
+            # add the new page after the last child of parent_page
+            page.path = parent_page.get_last_child()._inc_path()
+
         return page
 
     def get_form(self, page, query_dict):


chosak (approved):
LGTM. It's too bad that treebeard doesn't expose versions of `add_child` and `add_sibling` that create the node without persisting to the database.

PR #6398, opened 2020-09-23
Locale components

kaedroho:
Locale components are extra models that you can register that add additional information to a `Locale`. The form for this model is displayed in the locale's UI.

For example, the following code will add a "flag" choice field to the locale form:

```python

from wagtail.locales.views import register_locale_component

@register_locale_component
class LocaleFlag(models.Model):
    locale = models.OneToOneField(Locale)
    flag = models.CharField(choices=[...])
```


kaedroho:
I've rebased this on the [docs branch](https://github.com/wagtail/wagtail/pulls) and added some docs now. Ready for review!

kaedroho:
Abandoning this in favour of moving the entire locale sync feature to Wagtail core, this would allow it to be used with other translation plugins too.

There's a bunch of fixes that need to be made first though, these are documented in https://github.com/wagtail/wagtail-localize/issues/386

PR #6400, opened 2020-09-23
Consistently show user's full name (where available) in reports

gasman:
Currently, views that list PageLogEntry records display the user's `get_username` - a client pointed out that this is inconsistent with other views, which show the full name. While looking into this, I found that `{{ user.get_full_name|default:user.get_username }}` is a heavily-used pattern in templates, so this seemed like a good opportunity to move that into a template filter.

### DIFF
--- a/wagtail/admin/templatetags/wagtailadmin_tags.py
+++ b/wagtail/admin/templatetags/wagtailadmin_tags.py
@@ -628,3 +628,24 @@ def minimum_collection_depth(collections: QuerySet) -> int:
     use {% format_collection collection min_depth %}.
     """
     return collections.aggregate(Min('depth'))['depth__min'] or 2
+
+
+@register.filter
+def user_display_name(user):
+    """
+    Returns the preferred display name for the given user object: the result of
+    user.get_full_name() if implemented and non-empty, or user.get_username() otherwise.
+    """
+    try:
+        full_name = user.get_full_name()
+        if full_name:
+            return full_name
+    except AttributeError:
+        pass
+
+    try:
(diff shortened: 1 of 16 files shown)


wagtail/admin/templatetags/wagtailadmin_tags.py
```diff
@@ -637,7 +637,7 @@ def user_display_name(user):
     user.get_full_name() if implemented and non-empty, or user.get_username() otherwise.
     """
     try:
-        full_name = user.get_full_name()
+        full_name = user.get_full_name().strip()
```
jacobtoppm, line 640: Maybe worth adding this to `get_username` line as well for consistency? Admittedly you'd need to override the default validator to get spaces in the username, but it happens.

jacobtoppm (approved):
Looks great! Will be nice to cut this pattern out of the templates

wagtail/admin/templatetags/wagtailadmin_tags.py
gasman: The `strip()` is needed here because if `get_full_name` has been implemented naively as `return self.first_name + ' ' + self.last_name`, then a blank name will be returned as a single space, and it will fail to fall back on the username. There's no equivalent scenario where `get_username` would 'accidentally' return whitespace (and if it does, there's no further fallback we can provide other than returning that whitespace username) so I'd say `strip()` is unnecessary there.

gasman:
Merged in 3300e3b851d1a0a31873fdf65176ac10bee9abc0

Issue #6388, opened 2020-09-17
Stop using <input type="button" ...


allcaps:
### Issue Summary

It looks like we _do_ use `<input type="button" ...` in the style guide ([L508](https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html#L508), [L519](https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html#L519), [L530](https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html#L530)), which we probably should stop doing given that it says on that very same same page not to do so.

### Steps to Reproduce

1. Search te code for `<input type="button"

Refs: https://github.com/wagtail/wagtail/pull/6369#pullrequestreview-489947548

Scotchester:
Closed via https://github.com/wagtail/wagtail/commit/3ab18aa901ddf6b7152e6d0a547cf7c2a7d054e3 in #6388

PR #6401, opened 2020-09-23
Issue #6388 --> Changed <input type=button> to <button type=button>

paqman85:

### DIFF
--- a/wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html
+++ b/wagtail/contrib/styleguide/templates/wagtailstyleguide/base.html
@@ -505,7 +505,7 @@
             <div class="row">
                 <div class="col3">
                     <div class="dropdown dropdown-button match-width">
-                        <button type="button" value="drop down" class="button" />
+                        <button type="button" value="drop down" class="button"></button>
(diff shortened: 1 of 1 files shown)


Scotchester (approved):
Thank you, @paqman85! We appreciate it.

Scotchester (changes requested):
Whoops, too hasty with my approval. `<button>` is not a self-closing tag, so these tags need to be converted to this format:

```html
<button type="button" class="button">value</button>
```

(replacing the `value` with the value of the `value` attribute from before)

paqman85:
No prob! I'll update it.

paqman85:
I may have missed  something again, did you want the `value="drop up"` to be between the button tags, not as a tag attribute?

Scotchester:
> I may have missed something again, did you want the `value="drop up"` to be between the button tags, not as a tag attribute?

Yes, the attribute should go away, and just `drop up` should appear between the tags.

paqman85:
Ok I think I got them all now. Sorry about the extra commits!

Scotchester (approved):

Scotchester:
Beautiful, thanks again, @paqman85!

PR #6405, opened 2020-09-24
[Comments] Initial models

jacobtoppm:
Adds first draft of models, tests, and the logic for updating comments when revisions linked to them are deleted


wagtail/core/models.py
```diff
@@ -4220,3 +4237,83 @@ def __str__(self):
...
+    # In most cases, revisions will be purged oldest-first, so deleting the comment when the revision is deleted is the correct behaviour as the resolved
+    # comment is now inaccessible. However, in cases where the deleted revision_resolved is not the oldest revision, the revision_resolved needs to be
+    # changed instead. This is done in PageRevision.delete()
+
+    class Meta:
+        verbose_name = _('comment')
+        verbose_name_plural = _('comments')
+
+    def __str__(self):
+        return _("Comment on Page '{0}', left by {1}: '{2}'").format(self.page, self.user, self.text)
```
gasman, line 4273: I'd suggest not marking these strings as translatable, since they're presumably not going to appear in user-facing UI and are just creating extra work for translators. (If we do keep them translatable, they should use old-style `%` string formatting so that Transifex / makemessages can catch any discrepancies between placeholders in the original and translated strings.)

gasman (approved):
LGTM! Just a mostly-stylistic nitpick about translations on `__str__` strings.

Issue #6212, opened 2020-07-07
Privacy settings aren't copied on page copy


jessemenn:
### Issue Summary

Copying a page doesn't copy the original's privacy settings, but appears to inherit from the new parent's privacy settings. 

That is, if page A1 is public, and its child page A1a has privacy set to "private, accessible to...," copying page A1a to A1b causes A1b to be set to public. I feel there's a general expectation that the original's privacy settings would copy over to the newly created page. 

### Steps to Reproduce

1. Set a page's privacy setting to public and a child page to private.
2. Copy private page.

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: yes

### Technical details

* Python version: 3.7.3, 3.8.2
* Django version: 3.0.8, 2.2.10
* Wagtail version: 2.9.2
* Browser version: Chrome, Firefox, Safari.

zerolab:
The reason the privacy settings are not copied is that the copy mechanism deals with fields on the Page model. When one changes the privacy settings from public to anything else, a `PageViewRestriction` model instance is created.

To get this working, we should consider the following scenarios:
- The privacy settings are on the page being copied. On copy:
   - New parent has privacy settings. Do nothing
   - New parent has no privacy settings. Copy the settings (that is, create a new `PageViewRestriction` with the `page` foreign key set to the newly created page)
- The privacy settings are on the page parent. On copy: do nothing. If the new parent has privacy settings,

Skchoudhary:
Hello @zerolab , I would like to pick up this issue.

bilalshaikh292:
Hello @zerolab , I would like to pick up this issue.

ArijeetBanerjee:
I want to work on this issue. kindly give me permission and access to work on this issue

gasman:
Sorry for rethinking this after it's been open for so long, but I'm not sure if this is the desired behaviour. If your site has two blog sections, "Skunkworks Engineering Blog" (with strongly locked-down privacy settings) and "Corporate Blog" (with lesser or no privacy restrictions), then the natural interpretation of someone copying an article "iPhone 15 tech specs" from Skunkworks Engineering Blog to Corporate Blog is that the audience for that information has changed, and therefore it should inherit the privacy settings of the new parent, rather than copying its settings across.

On the other hand, if you're copying the page that the privacy settings are explicitly attached to, then I think it does make sense for those settings to be copied across - and that indeed doesn't happen at the moment.

julianwachholz:
[...]

PR #6406, opened 2020-09-24
Privacy setting copy

Skchoudhary:
#6212 

Need help to review whether I am doing this in right way or not.


gasman:
Thanks @Skchoudhary!

I don't think splitting the logic between views/page_privacy.py and views/pages/copy.py is quite right here - having a `copy_privacy` method in views/page_privacy.py would imply that there's a URL that users can visit in the admin specifically for copying privacy settings.

In fact, I don't think you need to change view-level code at all - if a developer was to call `page.copy(...)` from their own code (for example, a maintenance script), we'd still want privacy settings to be copied, even though that's not going through the admin views at all. So, all of the new code should be within the `Page.copy` method in wagtail/core/models.py.

Skchoudhary:
okay @gasman, will move the logic to wagtail/core/models.py.

Skchoudhary:
Hello @gasman 
Can you please review the commit.

wagtail/core/models.py
```diff
@@ -2049,6 +2050,29 @@ def copy(self, recursive=False, to=None, update_attrs=None, copy_revisions=True,
                 page_copy.numchild = numchild
                 page_copy.save(clean=False, update_fields=['numchild'])
 
+            # Copy privacy permission
+            if self.get_view_restrictions().exists() and to:
```
gasman, line 2054: `self.get_view_restrictions()` will return all view restrictions that apply to the page, including ones that are defined on a parent or ancestor - we don't want to copy them in that case. `self.view_restrictions.all()` will give the view restrictions defined on this page specifically, so that's what we should use here.

Also, if `to` isn't defined, it means that we're copying the page into the same parent, and in this case we still want to copy the page's restrictions - we shouldn't skip over this block.

wagtail/core/models.py
```diff
@@ -2049,6 +2050,29 @@ def copy(self, recursive=False, to=None, update_attrs=None, copy_revisions=True,
...
+                # On copy:
+                # 1.New parent has privacy settings. Do nothing
+                # 2. New parent has no privacy settings. Copy the settings (that is, create a new
+                # PageViewRestriction with the page foreign key set to the newly created page)
+                # The privacy settings are  on the page parent. On copy: do nothing.
+                # If the new parent has privacy settings, they will be inherited.
+
+                parent_page_restriction = to.get_view_restrictions()
+                if not parent_page_restriction.exists():
+                    privacy_settings_to_copy = get_object_or_404(PageViewRestriction, page_id=self.id)
```
gasman, line 2066: `get_object_or_404` would not be correct here - model code doesn't know anything about HTTP requests/responses, so it shouldn't be responsible for deciding whether to return a 404. However, this isn't really necessary anyway, since we're checking for the existence of a view restriction above - we can combine "does a view restriction exist?" with "retrieve the view restriction".

Overall, the new logic would look something like this:

    if to:
        parent_page_restriction = to.get_view_restrictions()
    else:
        parent_page_restriction = self.get_parent().get_view_restrictions()

    if not parent_page_restriction.exists():
        for privacy_settings_to_copy in self.view_restrictions.all():
            # copy the setting, as before

gasman (changes requested):
Thanks for the updates! Just a couple more changes required here.

Also, this now needs some test cases adding to [`TestCopyPage`](https://github.com/wagtail/wagtail/blob/906380486d8fa2da7e668e64ae2fd28a5961fb0d/wagtail/core/tests/test_page_model.py#L916), to confirm that this is working as expected.

Skchoudhary:
Hello @gasman 
I have done the requested changes, but bit confused about the test cases and have come up with these two test cases.

```
    def test_copy_page_copies_restriction_when_parent_page_have_no_restriction(self):
        # Test case where Page we are copy from have restriction and parent have no restriction
        ...

    def test_copy_page_doesnt_copies_restriction_when_parent_page_have_restriction(self):
        # Test case where Page we are copy from have restriction and parent have restriction
        ...
```

On looking some other test cases code, found a way to create Page object with `SimplePage`, but in this we cannot add restriction, can you point to some code where from I can take reference.

gasman:
@Skchoudhary Thanks for the updates!

Adding a view restriction inside test code can be done the same way as in any other code, by creating a `PageViewRestriction` object. There's an example here: https://github.com/wagtail/wagtail/blob/b0fa9c30d77972d90e0961fa6b42d4447e211e21/wagtail/core/tests/test_page_queryset.py#L363-L369

Skchoudhary:
Hello @gasman
Test cases added for the changed code, please review the changes. Will rebase the branch with the master after that :)

lb-:
@Skchoudhary been a while, are you able to rebase this on `main`. I know that you were waiting on some additional feedback, sorry that never happened.

There is another PR that aims to make [page copying better](https://github.com/wagtail/wagtail/pull/10145) and it reminded me of this PR.

Skchoudhary:
Hello @lb- I will rebase the code onto the main branch by the end of this week.

lb-:
Epic - thanks @Skchoudhary

Skchoudhary:
Hello @lb- PR is ready for the review and I have tested the changes on the squash too.

lb-:
Thanks. I've labelled as needs review again. I'll try to get to it if @gasman doesn't soon.

wagtail/tests/test_page_model.py
```diff
@@ -2141,6 +2142,41 @@ def test_copy_alias_page(self):
...
+
+        # check that the child_page_2 is in the result
+        self.assertTrue(PageViewRestriction.objects.filter(page=child_page_2).exists())
+
+    def test_copy_page_not_copies_restriction_when_parent_have_restriction(self):
+        """ Test case where Page we are copy from have restriction and parent have  restriction"""
+
+        homepage = Page.objects.get(url_path='/home/')
+        child_page_1 = SimplePage(title="Child Page 1", slug="child-page-1", content="hello child page 1")
+        homepage.add_child(instance=child_page_1)
```
gasman, line 2167: This test is a bit misleading because the same parent page is being used as the source and destination. As per @zerolab's [original specification](https://github.com/wagtail/wagtail/issues/6212#issuecomment-655377744), there are three different situations to consider when copying page X from parent A to parent B, which I think should become three distinct tests here:

1. Page X has a restriction, but neither A nor B have one; page X's restriction should be copied
2. Page A has a restriction; this should not be copied over when copying page X
3. Page X has a restriction, and page B has a restriction; in this case page X's restriction should not be copied over (because then the resulting copy would have multiple restrictions applied to it).

As it stands, the test above is testing situation 1, and this one is testing a mixture of 2 and 3.

PR #6409, opened 2020-09-25
Css variables part 1

JNaftali:
A much less ambitious version of https://github.com/wagtail/wagtail/pull/6349

This PR:
* Adds sass functions and mixins for easier authoring/modifying of CSS variables, as documented in the other MR
* Adds `postcss-custom-properties` to the sass gulp task so that the admin will continue working on IE despite the lack of support for css variables
* changes the `$color-teal` sass variable (and all the variables that depend on it) to reference the `--color-primary` css variable


JNaftali:
@thibaudcolas As far as your comment:
> It feels to me like it would be better to just refactor existing colors to not rely on modifying the darkness/lightness/saturation

How would you recommend dealing with the one-off colors in _listing.scss and _tabs.scss? Add a new color variable (`$color-teal-darkerer`? `$color-teal-darkest`?) that's close enough to those values that I can replace them?

allcaps (changes requested):
@JNaftali Thank you for the PR. Great effort! Also on the more advanced one :clap:

I've been requested to review your PR. Maybe because I contributed the previous color changes. When we went from low contrast lime to the higher contrast teal.

How does this work in practice? What does an integrator need to know to change the admin colours? The docs are missing. I need the guidance too ;)

I do get that branding is important. More important is that Wagtail strives to be WCAG2.1, AA level compatible. The Wagtail color scheme and contrasts ratios are carefully crafted to comply to Web Content Accessibility Guidelines. WCAG 2.2 AA. I'd like the docs to have a warning about braking contrast and how to avoid that.

JNaftali:
> @JNaftali Thank you for the PR. Great effort! Also on the more advanced one 👏
> 
> I've been requested to review your PR. Maybe because I contributed the previous color changes. When we went from low contrast lime to the higher contrast teal.

Thanks for the review!

> How does this work in practice? What does an integrator need to know to change the admin colours? The docs are missing. I need the guidance too ;)

I had been holding off on docs while I wasn't sure if the PR was going to go through any major changes. I'll have a go at writing some tonight or tomorrow.

In the mean time - to customize the admin colors you need to insert a stylesheet into the wagtail admin (often using https://docs.wagtail.io/en/v2.10.2/reference/hooks.html#insert-global-admin-css ). That stylesheet needs to have a :root selector that defines some css variables. For example:

```css
:root {
  --color-primary-hue: 45;
}
```

This PR adds 5 customizable colors: `color-primary`, `color-primary-darker`, `color-primary-dark`, `color-input-focus`, and `color-input-focus-border`. Ideally these should be set using 3 css variables per color - `color-primary-hue` (a unitless number), `color-primary-saturation` (percentage), and `color-primary-lightness` (percentage) (with equivalents for each of the other 4 colors). Assuming you choose a good color for `color-primary`, you shouldn't have to touch any of the other colors.

> I do get that branding is important. More important is that Wagtail strives to be WCAG2.1, AA level compatible. The Wagtail color scheme and contrasts ratios are carefully crafted. I'd like the docs to have a warning about braking contrast and how to avoid that.

That was absolutely something that I planned to do. Do you have a preferred browser extension for checking contrast values? If not I'll probably call out Axe.

Will try to update the PR with docs tonight

client/scss/tools/_various.colors.scss
```diff
@@ -0,0 +1,62 @@
...
+
+@function get-color($name) {
+    @return (var(--#{$name}-hue), var(--#{$name}-saturation), var(--#{$name}-lightness));
+}
+
+@function css-darken($hsl-tuple, $darken-by) {
+    $h: nth($hsl-tuple, 1);
+    $s: nth($hsl-tuple, 2);
+    $l: nth($hsl-tuple, 3);
+    @return ($h, $s, calc(#{$l} - #{$darken-by + 0%}));
```
thibaudcolas, line 31: Is this and the other HSL shifts really equivalent to the Sass color functions they replace?

thibaudcolas (commented):
Looking as amazing as the last PR @JNaftali. There is one thing I’d like to confirm – this approach relies on HSL, and as-is it wouldn’t be possible for people to only set `color-primary` in RGB hex and have it trickle down to the other shades?

Can you confirm that the cases below do behave as I’d expect?

```scss
    :root {
        // As you documented – this will auto-magically adjust all dark, darker, input, input border shades.
        --color-primary-hue: 25;

        // This one correctly overrides the primary color, but the dark and darker variations will still be based on our teal
        // since they are defined based on the -hue, -saturation, -lightness variables.
        --color-primary: #ff0000;
    }
```

So practically, for real-world theming, people will either have to know which primary color they want to use in HSL notation, and know how to split a normal HSL value into the three components:

```scss
:root {
    // This will trickle down to the dark and darker shades as expected.
    --color-primary-hue: 25;
    --color-primary-saturation: 25;
    --color-primary-lightness: 25;
}
```

…or if they want to use RGB / hex values, then they have to override _all_ of the colors:

```scss
:root {
    --color-primary: #ff0000;
    --color-primary-dark: #cc0000;
    --color-primary-darker: #990000;
    --color-input-focus: #dd00f0;
    --color-input-focus-border: #dd00ff;
}
```

Is that correct? I’m still uncertain of whether the HSL trick really is worthwhile or not.

client/scss/tools/_various.colors.scss
JNaftali: Near as I can tell, yep! Run the following through sass (I did it on sassmeister.com to check easily) and you'll see the two colors end up identical:
```scss
.test {
    color: darken(red, 10%);
    background-color: hsl(hue(red), saturation(red), lightness(red) - 10%);
}
```

JNaftali:
Yup, everything you wrote there is correct. I don't know of any work coming up to standardize anything like this - since coming up with all this I've run across https://una.im/css-color-theming/ so I know that I'm not the only one to imagine doing stuff this way. It wouldn't shock me if there ends up being a way to make color functions in css whenever houdini support becomes the norm but I haven't seen anything written about that specifically.

Seems like even sass is trying to de-emphasize darken in favor of `color.scale` - https://sass-lang.com/documentation/modules/color#darken

JNaftali:
@allcaps Hey! Hate to be a bother, but are the docs that I added sufficient? Or are there further changes you would like to request?

docs/advanced_topics/customisation/admin_templates.rst
```diff
@@ -91,6 +91,24 @@ To replace the welcome message on the dashboard, create a template file ``dashbo
 
     {% block branding_welcome %}Welcome to Frank's Site{% endblock %}
 
+Custom branding colors
+======================
+
+To customize the primary color used in the admin UI, inject a css file using the hook :ref:`insert_global_admin_css` and override the variables within the ``:root`` selector:
```
allcaps, line 97: `UI` => `user interface`. No abbrs plz.

docs/advanced_topics/customisation/admin_templates.rst
```diff
@@ -91,6 +91,24 @@ To replace the welcome message on the dashboard, create a template file ``dashbo
...
+
+.. code-block:: css
+
+    :root {
+        --color-primary-hue: 25;
+    }
+
+``color-primary`` is an `hsl color <https://en.wikipedia.org/wiki/HSL_and_HSV>`_ composed of 3 css variables - ``--color-primary-hue`` (a unitless number), ``--color-primary-saturation`` (a percentage), and ``--color-primary-lightness`` (also a percentage). Separating the color into 3 allows us to calculate variations on the color to use alongside the primary color. If needed, you can also control those variations manually by setting ``hue``, ``saturation``, and ``lightness`` variables for the following colors: ``color-primary-darker``, ``color-primary-dark``, ``color-primary-darkest``, ``color-input-focus``, and ``color-input-focus-border``.
```
allcaps, line 105: H of HSL is a 360' number of the color wheel starting at red.

`(a unitless number)` => `(0-360 without unit)`

docs/advanced_topics/customisation/admin_templates.rst
```diff
@@ -91,6 +91,24 @@ To replace the welcome message on the dashboard, create a template file ``dashbo
...
+
+Warnings
+--------
+1. CSS variables are not supported in Internet Explorer, so the admin will appear with the default colors when viewed in that browser.
+2. The default Wagtail colors conform to the WCAG2.1 AA level color contrast requirements. When customizing the admin colors you should test the contrast using tools like `Axe <https://www.deque.com/axe/browser-extensions/>`_.
```
allcaps, line 110: Please use the warning directive here.

```
Warnings
---------
```
=>
```
.. warning::
    This is warning text. 
```
Just use two times the warning directive. No numbering.

allcaps (approved):
Sorry to let you hanging. Thank you for your patience.

Overall it looks great. Just some nitpicks.

docs/advanced_topics/customisation/admin_templates.rst
allcaps: `css` => `CSS`

docs/advanced_topics/customisation/admin_templates.rst
allcaps: `css` => `CSS`

thibaudcolas:
Thank you @JNaftali, honestly this is so future-thinking I wasn’t too sure how to review it. I wasn’t expecting something so fancy! I’ll do one more round of tests on Friday, and hopefully merge this then.

JNaftali:
Thanks for the review @allcaps! Made those changes.

And that's super exciting to hear, @thibaudcolas! This is my first time contributing a feature to an open source project that I use, nice to hear it described like that

allcaps:
Again, thank you! This is a much requested feature. I'm sure many people are happy to see it realised. :D 

I'll leave the final review and merge to @thibaudcolas.

client/scss/settings/_variables.scss
```diff
@@ -28,9 +28,10 @@ $breakpoints: (
 );
 
 // colours
-$color-teal: #007d7e;
-$color-teal-darker: darken(adjust-hue($color-teal, 1), 4);
-$color-teal-dark: darken(adjust-hue($color-teal, 1), 7);
+$color-teal: var(--color-primary);
+$color-teal-darker: var(--color-primary-darker);
+$color-teal-dark: var(--color-primary-dark);
+$color-teal-darkest: var(--color-primary-darkest);
```
thibaudcolas, line 34: I would welcome a follow-up PR to search-and-replace all the Sass variables so they switch to `color-primary` as well. But glad this isn’t in the same PR for now!

thibaudcolas (changes requested):
This looks great in Chrome, unfortunately it’s broken in IE11 as far as I can see. The fallback for the primary color is there, but the `calc` fails for the darker shades. Did you do any testing there?

ie11-calc

I’ll have a look over the weekend or next week to understand what the issue is.

docs/advanced_topics/customisation/admin_templates.rst
```diff
@@ -91,6 +91,42 @@ To replace the welcome message on the dashboard, create a template file ``dashbo
...
+        --color-primary-darkest-hue: 22;
+        --color-primary-darkest-saturation: 100%;
+        --color-primary-darkest-lightness: 10%;
+    }
+
+.. warning::
+    CSS variables are not supported in Internet Explorer, so the admin will appear with the default colors when viewed in that browser.
+
+.. warning::
+    The default Wagtail colors conform to the WCAG2.1 AA level color contrast requirements. When customizing the admin colors you should test the contrast using tools like `Axe <https://www.deque.com/axe/browser-extensions/>`_.
```
thibaudcolas, line 128: I think all of these should be the first thing in this section, not last. And combined in a single warning?

docs/advanced_topics/customisation/admin_templates.rst
```diff
@@ -91,6 +91,42 @@ To replace the welcome message on the dashboard, create a template file ``dashbo
...
+
+To customize the primary color used in the admin user interface, inject a CSS file using the hook :ref:`insert_global_admin_css` and override the variables within the ``:root`` selector:
+
+.. code-block:: css
+
+    :root {
+        --color-primary-hue: 25;
+    }
+
+``color-primary`` is an `hsl color <https://en.wikipedia.org/wiki/HSL_and_HSV>`_ composed of 3 CSS variables - ``--color-primary-hue`` (0-360 with no unit), ``--color-primary-saturation`` (a percentage), and ``--color-primary-lightness`` (also a percentage). Separating the color into 3 allows us to calculate variations on the color to use alongside the primary color. If needed, you can also control those variations manually by setting ``hue``, ``saturation``, and ``lightness`` variables for the following colors: ``color-primary-darker``, ``color-primary-dark``, ``color-primary-darkest``, ``color-input-focus``, and ``color-input-focus-border``:
```
thibaudcolas, line 105: This all sounds very advanced – I think we’re missing an intermediary example of "here is how you can set all the colors manually with hex codes".

docs/advanced_topics/customisation/admin_templates.rst
JNaftali: done

docs/advanced_topics/customisation/admin_templates.rst
JNaftali: added such an example to the end

JNaftali:
@thibaudcolas I did not actually test on IE11 - I don't have a great way to run it, so your help debugging that would be very much appreciated.

thibaudcolas:
According to [caniuse](https://caniuse.com/calc) it looks like IE11 doesn’t support `calc` in color functions. No matter, the usage of `calc` in the IE11-only fallback really is just a remnant of the custom properties setup and shouldn’t actually be needed. I’ll try to use https://github.com/postcss/postcss-calc to remove the unneeded & incompatible `calc`s and report back.

client/scss/settings/_variables.scss
```diff
@@ -59,8 +60,8 @@ $color-header-bg: $color-teal;
 
 $color-fieldset-hover: $color-grey-5;
 $color-input-border: $color-grey-4;
-$color-input-focus: lighten(desaturate($color-teal, 40), 72);
-$color-input-focus-border: lighten(saturate($color-teal, 12), 10);
+$color-input-focus: var(--color-input-focus);
+$color-input-focus-border: var(--color-input-focus);
```
thibaudcolas, line 64: Looks like this should be:

```suggestion
$color-input-focus-border: var(--color-input-focus-border);
```

client/scss/elements/_root.scss
```diff
@@ -0,0 +1,9 @@
+:root {
+    @include define-color('color-primary', #007d7e);
+    @include define-color('color-primary-darker', css-darken(css-adjust-hue(get-color('color-primary'), 1), 4%));
+    @include define-color('color-primary-dark', css-darken(css-adjust-hue(get-color('color-primary'), 1), 7%));
+    @include define-color('color-primary-darkest', css-darken(get-color('color-primary-darker'), 9%));
+
+    @include define-color('color-input-focus', css-lighten(css-desaturate(get-color('color-primary'), 40%), 72%));
+    @include define-color('color-input-focus-border', css-lighten(css-desaturate(get-color('color-primary'), 12%), 10%));
```
thibaudcolas, line 8: Based on the previous Sass definition this should be:

```suggestion
    @include define-color('color-input-focus-border', css-lighten(css-saturate(get-color('color-primary'), 12%), 10%));
```

thibaudcolas (changes requested):

JNaftali:
@thibaudcolas Still haven't managed to get IE running to test there but I added postcss-calc as you suggested and it seems to have removed calc from the backup declarations. So hopefully fixed whatever issues existed on IE

thibaudcolas:
👍  thank you, I’ll assign this back to myself and see whether that was enough / take over from there.

dest81:
Thank you that doing this @JNaftali! I have played with your branch and  it works very well. 
So far I found one thing - in the stream field add block button on hover has old default wagtail color #007d7e. The problem is [here](https://github.com/wagtail/react-streamfield/blob/master/src/scss/_variables.scss#L19).

JNaftali:
> So far I found one thing - in the stream field add block button on hover has old default wagtail color #007d7e. The problem is [here](https://github.com/wagtail/react-streamfield/blob/master/src/scss/_variables.scss#L19).

Thanks for testing this out!, @dest81!

Given that react-streamfield is a separate package I don't think I want to mess with that until this PR is merged in... But! Since css variables exist at runtime, react-streamfield could fairly easily reference wagtail's color variables in their css. Thanks for making me aware of this, I might take a swing at that next :)

dest81:
@thibaudcolas I wonder if/when this PR can be merged?

thibaudcolas:
@dest81 when I get time to test it in IE11 and it works there as well – last time I tested this the styles were completely broken in that browser.

JNaftali:
@thibaudcolas Obviously you should test this in your own time, but I managed to get VirtualBox up and running on my machine and it seems to be running well on IE! Screenshot attached (of IE accessing the Squash branch).

Screen Shot 2020-12-12 at 2 08 11 PM

dest81:
I have tested on IE11 via BrowserStack and can confirm it works.

dest81:
Hi guys,
Sorry that annoying you but any news here? 
I still see "needs work" label on this PR, shouldn't be changed to "Needs review"?
I know your are all busy but can you give me rough estimate when it can be reviewed and possibly merged?

thibaudcolas:
Hey @dest81, you’re right, I’ve updated the labels. I have the branch checked out locally and am halfway through reviewing it. There is a blocker with https://github.com/wagtail/react-streamfield/pull/14 but I managed to find another way to do this that doesn’t involve changes to this separate package, so we’re in a good place. I’ll close that PR once I’ve confirmed my approach does work.

I still have more testing to do – I wouldn’t have noticed https://github.com/wagtail/react-streamfield/pull/14 had @JNaftali not reported there, so I want to make sure I’m not missing other things like this.

One other thing I’m still unsure of is `color-primary-darkest` – it’s very annoying to have to introduce this new shade for just 2 places out of the whole of Wagtail, and I think I found a way to change how those styles are made so we don’t need this. But here as well, more testing to do.

dest81:
@thibaudcolas thank you very much for the quick response, if you need any help (maybe with testing) I would be very glad to help ...

JNaftali:
> One other thing I’m still unsure of is `color-primary-darkest` – it’s very annoying to have to introduce this new shade for just 2 places out of the whole of Wagtail, and I think I found a way to change how those styles are made so we don’t need this. But here as well, more testing to do.

I look forward to hearing what you come up with! Let me know if there's some way I can help

thibaudcolas (approved):
All working as expected across the browsers I tested! And yep, the `darkest` variant shouldn’t be needed in the end.

Thank you @JNaftali for making this happen! This took a fair amount of back and forth but the end result really is remarkable, and considering how frequently this had been requested I’m sure lots of people will find this useful. I’ll follow up in #2943 for any further work in this area.

thibaudcolas:
@dest81 this is now published in Wagtail 2.12rc1 if you want to try it out: https://pypi.org/project/wagtail/2.12rc1/. Release notes: https://docs.wagtail.io/en/latest/releases/2.12.html#admin-color-themes.

dest81:
@thibaudcolas I did a quick test and it works as expected. Will work with it more closely after wagtail 2.12 release. Thank you  @JNaftali and @thibaudcolas for the great job

PR #6410, opened 2020-09-27
Update urlify.js to retain stopwords and fix handling of Turkish İ

### LINKED ISSUE #4899: Stop removing stopwords from auto-generated page slugs
### Issue Summary

When a new Wagtail page is saved without the user entering a slug, its slug is auto-generated from the page's title. Wagtail's slug generation uses a copy of the logic used by [Django's `SlugField`](https://docs.djangoproject.com/en/2.1/ref/contrib/admin/#django.contrib.admin.ModelAdmin.prepopulated_fields) (the Wagtail codebase contains a copy of the "bit of JavaScript" mentioned there). This logic includes the removal of [certain stopwords](https://github.com/wagtail/wagtail/blob/b34b547c6abe441429d019289dc542ac76930015/wagtail/admin/static_src/wagtailadmin/js/vendor/urlify.js#L163) if the page title contains only ASCII characters. For example, a page with title "To be or not to be, that is the question" will be given a slug "be-or-not-be-question", not "to-be-or-not-to-be-that-is-the-question" as one might expect (this can be entered manually if desired).

This is longstanding behavior (since #8) but I'm opening this issue to solicit feedback on whether Wagtail should stop removing these stopwords, motivated in particular by #4881 and related discussion in #4884.

---

Here is the logic that performs stopword stripping, which includes the list of words that are removed:

https://github.com/wagtail/wagtail/blob/882f8f3cf8ddd79c30e611a48882b309e90dad0c/wagtail/admin/static_src/wagtailadmin/js/vendor/urlify.js#L158-L170

This behavior comes from Django's [urlify.js](https://github.com/django/django/blob/b9cf764be62e77b4777b3a75ec256f6209a57671/django/contrib/admin/static/admin/js/urlify.js), which is used in the Django admin in conjunction with [`ModelAdmin.prepopulated_fields`](https://docs.djangoproject.com/en/2.1/ref/contrib/admin/#django.contrib.admin.ModelAdmin.prepopulated_fields) on [`SlugField`](https://docs.djangoproject.com/en/2.1/ref/models/fields/#slugfield) fields, of which `Page.slug` is one. There's no backend code related to removal of these words.

As @jjanssen mentions on #4884, the behavior of having a title like "Before the Sunrise" converted to a slug of just plain "sunrise" feels quite unexpected. Another example would be three pages named "Before Wagtail Space", "At Wagtail Space", and "Since Wagtail Space". All three of these would be given the same auto-generated slug: "wagtail-space".

Another downside of the existing logic is the unequal treatment given to English over other languages. Only English stopwords are removed, and if a title contains even a single Unicode character, no stopwords are. Worse, stopwords can be removed even though they may be valid words in other languages (see [django#12905](https://code.djangoproject.com/ticket/12905)).

Granted, the current behavior does given users the ability to override the removal of stopwords if desired -- by retyping the slug -- but this is an extra step they'd have to remember to do. Especially given the complexity introduced by something like #4881, might it be better to default to a slightly simpler behavior?

---

The most straightforward change would be to remove the lines excerpted above, but keep the rest of the slug generation logic, which would maintain similarity with Django.

One consideration is around deviating from Django's logic. Because `Page.slug` is a `SlugField`, it might be best to stay in step with Django slug creation, perhaps proposing any change in functionality there. On the other hand, Wagtail should be first and foremost focused on the needs of users, not on maintaining compatibility with Django admin behavior. Additionally, Wagtail is already using a copy of the logic instead of directly pulling in the Django JS file (and has to work to keep it up-to-date with Django changes, see for example #4897). The fact that Django's stopword list [hasn't changed since 2005](https://github.com/django/django/blob/dd5320d1d56ca7603747dd68871e72eee99d9e67/media/js/urlify.js) (!) might be a sign that the Django community is happy with the existing logic -- or the use of Django `SlugField`s with auto-generation might be an edge case that doesn't get much attention.

Wagtail could instead consider making this logic customizable, so that users can define their own slugging behavior if it differs from the Wagtail default. This is at least theoretically possible now by including an alternate `wagtailadmin/js/vendor/urlify.js` file in a project's templates, although maybe not as simple or well-contained as one might like; it'd be nicer if you could override just the stopwords piece while using the default unicode character handling, for example. Customization of urlify.js logic in Django seems to have been discussed a fair bit in Django tickets and on the django-developers mailing list, but [the decision was made](https://code.djangoproject.com/ticket/2282#comment:19) not to implement.)

Other potential questions to consider:
- Do other CMSes remove these words from URLs? [Wordpress doesn't seem to](https://developer.wordpress.org/reference/functions/wp_insert_post/). [Django CMS does seem to](https://github.com/divio/django-cms/commit/ff968036495acba6def19634830a8c0efa2d7966#diff-9f0a5a2ac54c8ae5b2c87ff730211c5dR30), referencing Django's JS file directly.
- Are there modern SEO concerns around stopword removal? [Here's](https://onlinemediamasters.com/why-you-shouldnt-remove-stop-words-from-urls/) an arbitrary SEO page that seems to think stopwords are okay. [Here's](https://cseo.com/blog/google-stop-words/) another that advises against -- and look at the bottom of that page for a longer list of what they consider stopwords in 2018. The Django list that Wagtail copies is unchanged since 2005 -- is it still reasonable and effective to filter out this specific list? [Here's](https://stackoverflow.com/questions/9734970/better-seo-to-remove-stop-words-from-an-articles-url-slug/47957580#47957580) an arbitrary SO answer that comes down somewhere in the middle.
- How frequently in practice do Wagtail users manually edit the auto-generated slug when creating a new page? It'd be interesting to create a script that one could run against a Wagtail DB to compute this. One example would be a post like [What’s It Like For A Developer Moving To Wagtail From Drupal?](https://wagtail.io/blog/whats-it-like-for-a-developer-moving-to-wagtail-from-drupal/) on the Wagtail blog. This currently uses the slug "whats-it-like-for-a-developer-moving-to-wagtail-from-drupal", instead of what would be the current default "whats-like-developer-moving-wagtail-drupal".

### Steps to Reproduce

2. `./manage.py migrate && ./manage.py createsuperuser`
3. `./manage.py runserver`, visit http://localhost:8000/admin, and login with the superuser you just created.
4. [Add a new child page under Home](http://localhost:8000/admin/pages/add/home/homepage/3/) and give it a title like "To be or not to be, that is the question". Note in the Promote tab that the autogenerated slug is "be-or-not-be-question".

### Technical details

* Python version: 3.6.3
* Django version: 2.1.3
* Wagtail version: 2.3

gasman:
Fixes #4899 and the İstanbul issue raised on Slack and at https://github.com/django/django/pull/12237.

I reapplied @chosak / @Scotchester's fix from https://github.com/django/django/commit/62f1655a64795d055f72e53557fb8404c5430963 manually rather than just taking the latest Django version, as there were a few ES6-specific fixes in the interim that could break IE11 compatibility (which Django has dropped, but we haven't).

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/js/page-editor.test.js
+++ b/wagtail/admin/static_src/wagtailadmin/js/page-editor.test.js
@@ -13,10 +13,10 @@ describe('page-editor tests', () => {
       /* true triggers to use django's urlify */
       expect(cleanForSlug('Before', true)).toBe('before');
       expect(cleanForSlug('The', true)).toBe('the');
-      expect(cleanForSlug('Before the sun rises', true)).toBe('sun-rises');
+      expect(cleanForSlug('Before the sun rises', true)).toBe('before-the-sun-rises');
       expect(cleanForSlug('ON', true)).toBe('on');
-      expect(cleanForSlug('ON this day in november', true)).toBe('day-november');
-      expect(cleanForSlug('This & That', true)).toBe('this--that');
+      expect(cleanForSlug('ON this day in november', true)).toBe('on-this-day-in-november');
+      expect(cleanForSlug('This & That', true)).toBe('this-that');
     });
 
     it('should return a correct slug which is escaped by urlify', () => {
--- a/wagtail/admin/static_src/wagtailadmin/js/vendor/urlify.js
+++ b/wagtail/admin/static_src/wagtailadmin/js/vendor/urlify.js
@@ -128,7 +128,7 @@
 
     var Downcoder = {
         'Initialize': function() {
-            if (Downcoder.map) {  // already made
+            if (Downcoder.map) { // already made
                 return;
             }
             Downcoder.map = {};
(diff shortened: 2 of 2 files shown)


Scotchester (approved):
Woohoo! I created a test page in Squash to confirm that the lack of stopword removal works as expected.

Scotchester:
Needs a little JS unit test adjustment, though.

gasman:
Thanks @Scotchester! Force-pushed with unit test fixed now...

gasman:
Merged in 2590fab0b6444809df21fb1579cc845f161ee99b

PR #6411, opened 2020-09-28
Fix #6408: InvokeViaAttributeShortcut makes it impossible to pickle objects

### LINKED ISSUE #6408: InvokeViaAttributeShortcut makes it impossible to pickle objects
### Issue Summary

`wagtail.core.utils.InvokeViaAttributeShortcut` makes it impossible to pickle `BaseSetting` object.

So it effectively kills object level caching.

### Steps to Reproduce

```
import pickle
some_settings_obj = SomeSettings.objects.get(pk=1)
picke.dumps(some_settings_obj, -1)
```
would produce `TypeError: 'str' object is not callable`

ababic:
Resolves #6408

### DIFF
--- a/wagtail/contrib/settings/models.py
+++ b/wagtail/contrib/settings/models.py
@@ -1,4 +1,5 @@
 from django.db import models
+from django.utils.functional import cached_property
 
 from wagtail.core.models import Site
 from wagtail.core.utils import InvokeViaAttributeShortcut
@@ -81,11 +82,14 @@ class BaseSetting(models.Model):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        # Allows get_page_url() to be invoked using
-        # `obj.page_url.foreign_key_name` syntax
-        self.page_url = InvokeViaAttributeShortcut(self, 'get_page_url')
         # Per-instance page URL cache
         self._page_url_cache = {}
+
+    @cached_property
+    def page_url(self):
+        # Allows get_page_url() to be invoked using
+        # `obj.page_url.foreign_key_name` syntax
+        return InvokeViaAttributeShortcut(self, 'get_page_url')
 
     def get_page_url(self, attribute_name, request=None):
         """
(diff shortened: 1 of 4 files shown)


pySilver:
5c: do we really need slots here? Having '__getstate__' override in BaseSetting seems like a half measure, since you'll have to repeat the same workaround in every implementation that needs a pickle support. It would be nicer to have it fixed on a definition level. I'd say pickle expects this.

ababic:
@pySilver I feel we are right to exclude `page_url` when pickling site settings, because the InvokeViaAttributeShortcut instance does not contribute anything meaningful to the state (it can be recreated like-for-like from non-instance specific data).

Whether or not `InvokeViaAttributeShortcut` uses `__slots__` is mostly irrelevant I thing - because it doesn't ever make sense to pickle an `InvokeViaAttributeShortcut` instance on an object (it isn't meaningful). If it did, it would include another representation of the object, which has a reference to the `InvokeViaAttributeShortcut` instance, creating an infinite loop.

pySilver:
@ababic its almost correct

1. Yes there is no need to pickle page_url, we can simply drop it (I do this for all my Page models usually by checking isinstance cached_property in `__getstate__`)

2. Pickle handles circular references just fine. Its just a gotcha - if you use `__slots__` & together with `__getattr__` then you need to implement `__getstate__` & `__setstate__` In absense of this methods, pickle would fallback to call `__getattr__` and create an infinity loop here.

Either way if there are no plans using `InvokeViaAttributeShortcut` anywhere else, then `cached_property` & state exclusion will work just fine.

ababic:
> If you use __slots__ & together with __getattr__ then you need to implement __getstate__ & __setstate__ In absence of this methods, pickle would fallback to call __getattr__ and create an infinity loop here

Just to clarify for anyone stumbling across this conversation: `__slots__` hasn't got anything to do with the pickle failure here (an instance of `InvokeViaAttributeShortcut` will not pickle, regardless of whether `__slots__` is present or not). The problem is the `__getattr__` override. Adding `__getstate__` & `__setstate__` does resolve the issue though.

ababic:
@gasman this solves the issue of settings objects no longer being pickle-able, as well as making `InvokeViaAttributeShortcut` instances themselves pickle-able (in case they are used anywhere else where pickle-ability is important). Regression tests also included.

gasman:
Thanks @ababic! Merged in 29d564c238c5c6577b4b1f75203c1c77acdb35dc.

PR #6412, opened 2020-09-28
Allow ignoring a page when finding an available slug

kaedroho:
This is useful for when using this function to update a pages slug


gasman (approved):
LGTM!

kaedroho:
Thanks!

PR #6414, opened 2020-09-29
Add a comment to describe why _filtered_by_child_of exists

kaedroho:
I was just going to remove this as per https://github.com/wagtail/wagtail/pull/6222#discussion_r493740072 but I notuced that we actually use this attribute to pass the parent page into the `construct_explorer_page_queryset` hook.

This adds a comment to document this a bit better.


zerolab (approved):
Nice catch!

gasman (approved):
Another reason to hate `construct_explorer_page_queryset`! Annoying, but fair enough...

PR #6415, opened 2020-09-29
Make the API's get_base_url tests use Django TestCase

kaedroho:
In my review of https://github.com/wagtail/wagtail/pull/6341, I completely missed the fact that the test case was using `unittest.TestCase` and not Django's `TestCase`. But the tests were passing at the time I reviewed it.

After merge though, this caused some problems because this test case modifies data but Django is unable to roll it back properly causing strange test failures.

This PR fixes this test case to be an ordinary Django `TestCase`. Note that this also changes a little bit of logic so that `WAGTAILAPI_BASE_URL = None` which is interpreted the same way as if you didn't specify this setting at all. This brings it in line with other settings in Wagtail.


kaedroho:
Travis is taking forever so I've reverted the problematic commit and added it into this PR

gasman (approved):

PR #6417, opened 2020-09-30
Add migration to apply alphabetical collection ordering

gasman:
When I proposed adding `node_order_by = ['name']` to the Collection model in #6141, I didn't realise that it would break creation of new collections for people who hadn't run the `fixtree` command. The release note did mention that fact, but it feels a bit clunky for a Wagtail upgrade to break a routinely-used feature and require a manual step to fix it. (I like to think of the 'Upgrade considerations' section as a sort of sliding scale - the more esoteric feature-hacking you do, the more on the hook you are for paying attention to it :-) )

So, I've added a migration to ensure that the fixtree process runs as part of the standard 2.11 upgrade process. I've had to slightly break migration best practices and import the real Collection model (as opposed to the migration-frozen one) to get at the `fix_tree` method, but I believe that should be safe - `fix_tree` only touches the Treebeard-specific fields via `values_list` and `update`, so it shouldn't be sensitive to schema changes elsewhere.

Unfortunately this doesn't make the change completely transparent - sites that assign `path` directly after the migrations have run (like bakerydemo, which loads them from a fixture) will still be bitten by it, so I've updated the release note accordingly. Also took the opportunity of making nested collections a headline feature for 2.11, which saves having to put the 'remind me what collections are again...' detail in the upgrade note :-)


ababic:
@gasman looks good to me! I would make one suggestion though...

> fix_tree only touches the Treebeard-specific fields via values_list and update, so it shouldn't be sensitive to schema changes elsewhere.

I feel it would be sensible to add a comment to this effect to the data migration.

gasman:
Thanks @ababic! Merged (with comment added) in 7e74b2b5d978b11c4f13a2d18d11b7d54d85c834.

PR #6419, opened 2020-09-30
Pass chooser URLs to widgets via data attributes instead of global JS vars

gasman:
Chipping away at `_editor_js.html` bit by bit...

Update chooser widgets so that the URLs to the modal views are passed via a `data-` attribute on the widget's root element, instead of depending on the global `chooserUrls` object in _editor_js.html.

Draftail (and hallo.js) still relies on `chooserUrls`, so it can't be eliminated _just_ yet.

### DIFF
--- a/client/tests/stubs.js
+++ b/client/tests/stubs.js
@@ -61,7 +61,6 @@ global.chooserUrls = {
   externalLinkChooser: '/admin/choose-external-link/',
   imageChooser: '/admin/images/chooser/',
   pageChooser: '/admin/choose-page/',
-  snippetChooser: '/admin/snippets/choose/',
 };
 
 /* use dummy content for onload handlers just so that we can verify that we've chosen the right one */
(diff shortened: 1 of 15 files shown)


gasman:
(Next on the agenda: there are probably a lot of views, such as wagtailsites/create.html and wagtailsites/edit.html, which don't make use of rich text or expanding formsets, and are only importing _editor_js to make the page chooser widget work - these can now ditch _editor_js 🎉 )

client/tests/stubs.js
```diff
@@ -61,7 +61,6 @@ global.chooserUrls = {
   externalLinkChooser: '/admin/choose-external-link/',
   imageChooser: '/admin/images/chooser/',
   pageChooser: '/admin/choose-page/',
-  snippetChooser: '/admin/snippets/choose/',
```
kaedroho, line 64: Do we still need the other choosers in this file?

kaedroho (approved):
Looks good to me!

PR #6420, opened 2020-10-01
Fixed an issue with copying unsaved Pages

### LINKED ISSUE #6221: Copying unsaved pages works
Just discovered an accidental Wagtail feature. Not sure if it's a bug. You can copy an unsaved page. For example:

```python
mypage = SimplePage(title="Foo page", slug="foo")
mypage.copy(to=homepage)
```

This actually works, it'll create the SimplePage. There aren't any tests for this, except for one test that appears to use it accidentally: https://github.com/wagtail/wagtail/blob/11561822839c0ad7d8a8c454eb3e2800e2fd78bd/wagtail/core/tests/test_page_model.py#L1196-L1202

I'm thinking we should block this, it's easy enough to create pages properly and this will probably lead to confusing logs (after the logging PR is merged).

samgans:
Resolves #6221

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -1756,6 +1756,11 @@ class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):
         :param log_action flag for logging the action. Pass None to skip logging.
             Can be passed an action string. Defaults to 'wagtail.copy'
         """
+
+        # If the model doesn't exist in the DB, the operation will not work
+        if self._state.adding:
+            raise Exception('You cannot copy the unsaved object')
+
         exclude_fields = self.default_exclude_fields_in_copy + self.exclude_fields_in_copy + (exclude_fields or [])
         specific_self = self.specific
         if keep_live:
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -1757,9 +1757,8 @@ class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):
             Can be passed an action string. Defaults to 'wagtail.copy'
         """
 
-        # If the model doesn't exist in the DB, the operation will not work
         if self._state.adding:
-            raise Exception('You cannot copy the unsaved object')
+            raise RuntimeError('Page.copy() called on an unsaved page')
 
         exclude_fields = self.default_exclude_fields_in_copy + self.exclude_fields_in_copy + (exclude_fields or [])
         specific_self = self.specific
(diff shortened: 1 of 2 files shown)


wagtail/core/models.py
```diff
@@ -1756,6 +1756,11 @@ def copy(self, recursive=False, to=None, update_attrs=None, copy_revisions=True,
         :param log_action flag for logging the action. Pass None to skip logging.
             Can be passed an action string. Defaults to 'wagtail.copy'
         """
+
+        # If the model doesn't exist in the DB, the operation will not work
+        if self._state.adding:
+            raise Exception('You cannot copy the unsaved object')
```
kaedroho, line 1762: Could we change this to a [`RuntimeError`](https://docs.python.org/3/library/exceptions.html#RuntimeError) instead?

kaedroho, line 1762: Could we change this to a [`RuntimeError`](https://docs.python.org/3/library/exceptions.html#RuntimeError) and reword the message to "`Page.copy()` called on an unsaved page"?

wagtail/core/models.py
```diff
@@ -1756,6 +1756,11 @@ def copy(self, recursive=False, to=None, update_attrs=None, copy_revisions=True,
         :param log_action flag for logging the action. Pass None to skip logging.
             Can be passed an action string. Defaults to 'wagtail.copy'
         """
+
+        # If the model doesn't exist in the DB, the operation will not work
```
kaedroho, line 1760: The operation does work, it's just weird that it does. Maybe get rid of this comment as the code itself reads quite well

kaedroho (changes requested):
Thanks! Just a couple of minor tweaks and this would be ready for merge!

wagtail/core/tests/test_page_model.py
```diff
@@ -1484,6 +1484,12 @@ def test_copy_page_with_excluded_parental_and_child_relations(self):
             # reset excluded fields for future tests
             EventPage.exclude_fields_in_copy = []
 
+    def test_copy_unsaved_page(self):
+        """Test that unsaved page will not be copied."""
+        new_page = SimplePage.objects.get(url_path='/home/about-us/')
+        with self.assertRaises(Exception):
```
kaedroho, line 1490: Just need to update this line to check for the `RuntimeError` then this is good to go!

kaedroho (approved):
Thanks! Will merge once tests pass

kaedroho:
@samgans Congratulations on your first contribution! Do you have a preferred name for [CONTRIBUTORS.rst](https://github.com/wagtail/wagtail/blob/master/CONTRIBUTORS.rst)? (You can use your GitHub username if you want to)

Issue #6197, opened 2020-07-01
Control over documents extensions


starkyller:
### Is your proposal related to a problem?

I tell people to only upload pdf files but they sometimes upload .docx

### Describe the solution you'd like

A simple key on the settings file like:
WAGTAIL_DOCS_EXTENSIONS = ['pdf', 'another_extension']

and then on the files upload validate if the inserted file has one of the defined extensions

django already has the FileExtensionValidator so I'm thinking a small update in the AbstractDocument ?

meghanabhange:
Can I pick this up?

tomdyson:
@meghanabhange go for it!

loicteixeira:
Implemented in #6423. Great contribution 👏

PR #6423, opened 2020-10-01
Add validation for file extension in AbstractDocument

meghanabhange:
**About this PR**
Added validation for the uploaded documents with reference to #6197. 

**What this does?**
Checks if the uploaded document complied with allowed documentation extension mentioned in settings as `WAGTAIL_DOCS_EXTENSIONS=["pdf", "docx"]`. Also ensured that it doesn't break(allowed all uploads) if `WAGTAIL_DOCS_EXTENSIONS` isn't provided at all in the settings.

When the allowed extensions are provided, it should show an error like the following if an invalid file is uploaded. 

CleanShot 2020-10-02 at 01 42 26@2x

I would also like to point out some shortcomings of this feature that were mentioned in Django docs for [FileExtentionValidator](https://docs.djangoproject.com/en/dev/ref/validators/#fileextensionvalidator)
> The extension is compared case-insensitively with allowed_extensions.

> Warning: Don’t rely on the validation of the file extension to determine a file’s type. Files can be renamed to have an extension no matter what data they contain.


meghanabhange:
- Changing the name of settings variable for file extension validation from `WAGTAIL_DOCS_EXTENSIONS` to `WAGTAILDOCS_EXTENSIONS` to maintain consistency across all the variable names in documents.
- The migrations added in b25aad6 and aaacbaf are now redundant once the validator was moved into a clean() function so removed them in 5419309

wagtail/documents/models.py
```diff
@@ -23,7 +24,10 @@ class DocumentQuerySet(SearchableQuerySetMixin, models.QuerySet):
 
 class AbstractDocument(CollectionMember, index.Indexed, models.Model):
     title = models.CharField(max_length=255, verbose_name=_('title'))
-    file = models.FileField(upload_to='documents', verbose_name=_('file'))
+    file = models.FileField(
+        upload_to="documents",
+        verbose_name=_("file")
+    )
```
allcaps, line 30: This does not change anything, does it?

allcaps (approved):
Thank you @meghanabhange,

This is very useful! One minor thing is the file field changes that don't change anything.

LGTM

wagtail/documents/models.py
meghanabhange: Yes, this doesn't change everything. My bad it got reformatted in the adding and removing validators. 
Should I change it back to the formatting it was before or this is okay?

allcaps: Yes please. 🙇

docs/reference/settings.rst
```diff
@@ -199,6 +199,16 @@ Specifies the MIME content type that will be returned for the given file extensi
...
 
+.. _wagtaildocs_extensions:
+
+.. code-block:: python
+
+  WAGTAILDOCS_EXTENSIONS = ['pdf', 'docx']
+
+A list of allowed document extensions that will be validated during document uploding. 
+If this isn't supplied all the document extensions are allowed. 
+Warning : this doesn't always ensure that the uploded file is a valid file as files can
```
jacobtoppm, line 210: Small typo here: "uploded" should be "uploaded"

wagtail/documents/models.py
```diff
@@ -53,6 +54,21 @@ class AbstractDocument(CollectionMember, index.Indexed, models.Model):
...
+        based on allowed extensions that were specified.
+        This doesn't always ensure that the uploded file is a valid file
+        as files can be renamed to have any extension no matter what
+        data they contain.
+
+        More info : https://docs.djangoproject.com/en/3.1/ref/validators/#fileextensionvalidator
+        """
+        allowed_extensions = getattr(settings, "WAGTAILDOCS_EXTENSIONS", None)
+        if allowed_extensions:
+            validatate = FileExtensionValidator(allowed_extensions)
```
jacobtoppm, line 69: Just a small typo again: validatate -> validate

wagtail/documents/models.py
```diff
@@ -53,6 +54,21 @@ class AbstractDocument(CollectionMember, index.Indexed, models.Model):
         index.FilterField('uploaded_by_user'),
     ]
 
+    def clean(self):
```
jacobtoppm, line 57: I think this could be worth making a `clean_file` method rather than just `clean`, as it only validates a single field. Also, I think it would be worth adding a call to `full_clean` to the `save` method, to catch programmatically created models too.

These are just minor points though!

jacobtoppm (changes requested):
Looks great overall! Just a few typos to fix, and potentially making the clean method a bit more specific

wagtail/documents/models.py
meghanabhange: So I after adding `full_clean` to `save` and changing the name of `clean` function to `clean_file` seems to be breaking. The function `clean_file` isn't being called during the full clean and I am not able to figure out why it's behaving that way :((

Also, save with full_clean seems to be breaking multiple testcases most of them giving the following error `django.core.exceptions.ValidationError: {'file': ['This field cannot be blank.']}`

Test results : 

```
FAILED (failures=1, errors=76, skipped=389, expected failures=14)
```

jacobtoppm: Ah, that's on me! I used the form syntax instead of the model - apologies.

Looks like that's just because our existing test cases don't provide a file. (We have a function for it `get_test_document_file` in tests/utils.py, but we only use it in one place). Since that's a lot of tests to rewrite, happy to leave that out for now.

jacobtoppm (approved):

docs/reference/settings.rst
```diff
@@ -199,6 +199,16 @@ Specifies the MIME content type that will be returned for the given file extensi
...
 A list of MIME content types that will be shown inline in the browser (by serving the HTTP header ``Content-Disposition: inline``) rather than served as a download, when using the ``serve_view`` method. Defaults to ``application/pdf``.
 
+.. _wagtaildocs_extensions:
+
+.. code-block:: python
+
+  WAGTAILDOCS_EXTENSIONS = ['pdf', 'docx']
+
+A list of allowed document extensions that will be validated during document uploading. 
+If this isn't supplied all the document extensions are allowed.
```
tomdyson, line 209: all the document -> all document

tomdyson (commented):
Thanks @meghanabhange!

tomdyson (approved):
Thank you @meghanabhange, this looks great!

Issue #6170, opened 2020-06-23
Drop support for IE11


thibaudcolas:
This is a tracking issue to discuss what it would look like for Wagtail to drop its support for IE11. 

Edit: [November 2020 update](https://github.com/wagtail/wagtail/issues/6170#issuecomment-722665899). **We are now through the process of gradually removing support for IE11**.

Edit: [September 2020 update](https://github.com/wagtail/wagtail/issues/6170#issuecomment-693523810). As of September 2020, we intend to gradually remove support for IE11 over the coming Wagtail releases. If this is something you rely on we would love to hear from you.

~**As of June 2020 there are no short-term plans to remove support, but if this is something you rely on we would love to hear from you.**~

This is prompted by a similar change to the Django admin – view [Django 3.1’s release notes](https://docs.djangoproject.com/en/3.1/releases/3.1/#contrib-admin-browser-support). Although I assume the Django admin might still work mostly as expected in IE11, it’s a good milestone for us to pay attention to, and feels like a good time for us to start planning for this in Wagtail.

**Particularly, it would be great to hear from people who do rely on IE11 in their organisation**, as the cost of supporting IE11 is pretty clear, while the benefits aren’t always so apparent. If you’re in this situation please take a moment to describe your situation – which Wagtail version you’re using, why you’re still using IE11, whether there are alternatives available, and what’s your plan with other suppliers/technologies dropping support.

---

### Known bugs

- #5914
- #5903
- #5297
- Some of #4274
- #6447

### Potential improvements

- #6107 some of the implementation options would require more work for IE11 support
- #1240 could switch to logical properties for RTL styles
- #2943 could start using CSS variables for easy theming
- #6130 could remove `.woff` font files 
- #3804 Would speed up code reviews
- #6059 We might be able to set up Prettier on the whole codebase, with ES2017 function parameter trailing commas for un-compiled code too.
- #5015 could use CSS grid and flexbox as part of our refactoring.
- Could reduce our static files’ size by switching to compiling JS for only the browser support we need.
- Could start loading Wagtail’s JS code as ES modules, with asynchronous loading, without having to polyfill / compile this.
- JS polyfills we include are only needed for IE11 support
- We could likely remove Modernizr (might be able to do this already regardless)

### Tech debt cleanup

- [GitHub search for "IE11"](https://github.com/wagtail/wagtail/search?q=IE11&unscoped_q=IE11)
- Remove legacy IE conditional comments (can do this regardless)

---

## IE11 market share

[...]

PR #6425, opened 2020-10-02
Implement admin-only IE11 support warning message, linking to support plans

thibaudcolas:
First step towards removing support, as discussed in https://github.com/wagtail/wagtail/issues/6170#issuecomment-693523810. Right now this is an admin-only, IE11-only, dashboard-only message. In the future, we will have this be visible for all users, and further down the line, on all pages like Wagtail’s "JavaScript is required" message.

The message is remove-able with the `construct_homepage_panels` hook, like all home panels. I didn’t find tests for these so I haven’t added any, happy to add a test if someone points me in the right direction.

In addition to what we discussed, I decided to add this to the documentation as well. It felt important to have this written somewhere more visible than the issue tracker, and is also a natural place for us to add more guidance about our plans.

Here is the message in its current iteration, displayed right after the admin-only upgrade notification:

ie-warning

---

* For Python changes: Have you added tests to cover the new/fixed behaviour? No, I couldn’t find existing tests I could add to for the other panels (might have missed them!)
* For front-end changes: Did you test on all of Wagtail’s supported browsers? **Please list the exact versions you tested**. Tested in latest Chrome macOS, latest Safari macOS, IE11 Win 10, latest MS Edge Win 10

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/home/ie11_warning.html
+++ b/wagtail/admin/templates/wagtailadmin/home/ie11_warning.html
@@ -1,0 +1,5 @@
+{% load i18n wagtailcore_tags %}
+
+<div data-ie11-warning hidden class="panel nice-padding">
+    <div class="help-block help-info">{% blocktrans %}Wagtail is gradually removing support for IE11, which may affect you or your organisation. <a href="https://docs.wagtail.io/en/stable/editor_manual/browser_issues.html#ie11">Learn more about our IE11 support plans</a>.{% endblocktrans %}</div>
+</div>
--- a/client/src/includes/initIE11Warning.js
+++ b/client/src/includes/initIE11Warning.js
@@ -1,0 +1,19 @@
+import { IS_IE11 } from '../config/wagtailConfig';
+
+/**
+ * Toggles IE11-only warning messages, if any.
+ */
+const initIE11Warning = () => {
+  if (!IS_IE11) {
+    return;
+  }
+
+  const warnings = [].slice.call(document.querySelectorAll('[data-ie11-warning]'));
+
+  warnings.forEach(warning => {
+    // eslint-disable-next-line no-param-reassign
+    warning.hidden = false;
+  });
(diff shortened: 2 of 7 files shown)


wagtail/admin/templates/wagtailadmin/home/ie11_warning.html
```diff
@@ -0,0 +1,5 @@
+{% load wagtailcore_tags %}
+
+<div data-ie11-warning hidden class="panel nice-padding">
+    <div class="help-block help-info">Wagtail is gradually removing support for IE11, which may affect you or your organisation. <a href="https://docs.wagtail.io/en/stable/editor_manual/browser_issues.html#ie11">Learn more about our IE11 support plans</a>.</div>
```
allcaps, line 4: This needs a {% blocktrans trimmed %}...

allcaps (changes requested):
Just the blocktrans.

wagtail/admin/templates/wagtailadmin/home/ie11_warning.html
thibaudcolas: @allcaps the existing "Upgrade notifications" is english-only, which is why I had this in english only as well: https://github.com/wagtail/wagtail/blob/master/wagtail/admin/templates/wagtailadmin/home/upgrade_notification.html

I’m not sure _why_ though, so yep feels like it could be nicer translated.

thibaudcolas: Thinking about this more, it does feel to me like I overthought this (…), and it’s probably just a mistake for the upgrade notification to be english only.

thibaudcolas:
Now good to go with the extra `blocktrans`.

gasman:
Merged in 6e404740c7834520e220c7d9b5fb0fd6e0465451 (with a small adjustment to the blocktrans tag so that the URL is not included in the translation string).

PR #6429, opened 2020-10-03
Lint docs with doc8

kaedroho:
This checks for trailing whitespace and missing newlines at ends of files (and a few other things). Some people configure their editor likes to fix these issues automatically so fixing them all now should reduce noise in the future.


kaedroho:
Merged as part of https://github.com/wagtail/wagtail/pull/6430

PR #6430, opened 2020-10-03
Spelling fixes

kaedroho:
Depends on my previous PR https://github.com/wagtail/wagtail/pull/6429


wagtail/tests/search/migrations/0001_initial.py
```diff
@@ -54,7 +54,7 @@ class Migration(migrations.Migration):
             name='ProgrammingGuide',
             fields=[
                 ('book_ptr', models.OneToOneField(auto_created=True, on_delete=django.db.models.deletion.CASCADE, parent_link=True, primary_key=True, serialize=False, to='searchtests.Book')),
-                ('programming_language', models.CharField(choices=[('py', 'Python'), ('js', 'Javascript'), ('rs', 'Rust')], max_length=255)),
+                ('programming_language', models.CharField(choices=[('py', 'Python'), ('js', 'JavaScript'), ('rs', 'Rust')], max_length=255)),
```
allcaps, line 57: Do we want to change migrations? I'm fine as choices don't trigger db changes. Just wanted to mention it.

allcaps (approved):
- Changing a migration is a bit strange. But can't really hurt since choices do not alter db.
- Do we need to update Transifex? Because `msgid`s changed.

Minor things, so approved.

wagtail/tests/search/migrations/0001_initial.py
kaedroho: These are just test migrations and this change doesn't actually alter schema anyway

kaedroho:
Thanks for the fast review!

> Do we need to update Transifex? Because msgids changed.

Yep, we change msgid's all the time but don't usually update the PO files until just before a release. I'm hoping that this would avoid requiring translators to retranslate those strings just because of capitalisation mistake

PR #6431, opened 2020-10-03
Fix incorrect reference to edit fields in image documentation

### LINKED ISSUE #6375: Editor's guide for images mentions non-existent fields
From https://docs.wagtail.io/en/v2.10/editor_manual/documents_images_snippets/images.html:

> Clicking an image will allow you to edit the data associated with it. This includes the Alt text, the photographers credit, the medium of the subject matter and much more.
> Changing the alt text here will alter it for all occurrences of the image in carousels, but not in inline images, where the alt text can be set separately.

The basic image model in an out-of-the-box Wagtail installation does not have fields for alt text, credit or medium, and carousels are not a core Wagtail concept, so how they behave is site-specific. (The inline model used for a carousel could, and probably should, have its own context-specific alt text field...) This text is presumably a hangover from the original Wagtail site built for [RCA](https://www.rca.ac.uk/), (which implements these fields with a [custom image model](https://docs.wagtail.io/en/stable/advanced_topics/images/custom_image_model.html)) and should be rewritten or removed.

saevarom:
Fixes #6375 

Took the liberty of mentioning the focal point editing instead of the non-existent fields, as well as simply removing the warning as it does not make any sense.

### DIFF
--- a/docs/editor_manual/documents_images_snippets/images.rst
+++ b/docs/editor_manual/documents_images_snippets/images.rst
@@ -5,10 +5,7 @@ If you want to edit, add or remove images from the CMS outside of the individual
 
 .. image:: ../../_static/images/screen31_images_page.png
 
-* Clicking an image will allow you to edit the data associated with it. This includes the Alt text, the photographers credit, the medium of the subject matter and much more.
-
-.. Warning::
-    Changing the alt text here will alter it for all occurrences of the image in carousels, but not in inline images, where the alt text can be set separately.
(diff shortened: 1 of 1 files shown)


kaedroho (approved):

gasman (changes requested):
Photographers credit is a non-core field too - perhaps this should be "the title, the focal point of the image and more"?

saevarom:
@gasman correct, I've changed it accordingly

gasman (approved):

PR #6433, opened 2020-10-04
Change docs code block style to 'default'

### LINKED ISSUE #6432: Change code block style in the docs
The colours in our existing code blocks fail WCAG AA on contrast: https://webaim.org/resources/contrastchecker/?fcolor=408090&bcolor=EEFFCC

See an example here: https://docs.wagtail.io/en/stable/advanced_topics/performance.html#cache

It looks like ``sphinx-rtd-theme`` uses a different style for their own docs: https://sphinx-rtd-theme.readthedocs.io/en/latest/demo/demo.html#code-blocks so maybe we should switch to that.

kaedroho:
Fixes #6432 

I think this highlighting style looks better and is more readable.

Before:
image

After:
image

We could probably do even better. There's a list of code block styles at https://help.farbox.com/pygments.html, or we could make our own!

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -115,7 +115,7 @@ exclude_patterns = ['_build']
 #show_authors = False
 
 # The name of the Pygments (syntax highlighting) style to use.
-pygments_style = 'sphinx'
+pygments_style = 'default'
 
 # A list of ignored prefixes for module index sorting.
 #modindex_common_prefix = []


Scotchester (approved):
Much better!

tomdyson:
Thanks @kaedroho, this looks much better!

PR #6434, opened 2020-10-05
(FIX): Page Copy does not trigger page_published signal

### LINKED ISSUE #5766: Page Copy, does not trigger page_published signal.
### Issue Summary

The `page_published` signal has been implemented, and when I go to copy a page and select "Publish Pages" I would expect this signal to get triggered during the copy. However it does not. 

### Steps to Reproduce

1. Assuming a basic project setup. 
2. implement a `page_published` signal following a similar guideline found here:
http://docs.wagtail.io/en/v2.1.1/reference/signals.html#page-published
3. Go to any page, and click "copy" then make sure "Publish Copied Page" is selected. 
4. The slack message will not be sent. 
5. I've implemented this a second way without the slack API as well, where Im simply trying to track the number of times a page was published. This is what that might look like:
```
from .models import HomePage
from .snippets import UsageStatistics

def track_published_pages(sender, **kwargs):

    page_instance = kwargs['instance']

    if page_instance.specific_class == HomePage:
        user = page_instance.owner

        stats, created = UsageStatistics.objects.get_or_create(page=page_instance,  user=user)

        if created:
            stats.published = 1
        else:
            stats.published += 1

        stats.save()


page_published.connect(track_published_pages)
```
6. This also doesn't trigger an update to the `UsageStatistics` model, like I would think it should. 

### Technical details

* Python version: 3.6.8
* Django version: 2.2.9
* Wagtail version: 2.7.1
* Browser version: Chrome 79

samgans:
Fixed #5766


wagtail/core/tests/test_page_model.py
```diff
@@ -1490,6 +1491,43 @@ def test_copy_unsaved_page(self):
...
+            nonlocal signal_page
+            signal_fired = True
+            signal_page = instance
+
+        page_published.connect(page_published_handler)
+        copy_page = christmas_page.copy(
+            update_attrs={'title': "New christmas", 'slug': 'new-christmas'},
+        )
+
+        self.assertIs(signal_fired, True)
```
kaedroho, line 1512: You can replace this with ``self.assertTrue(signal_fired)``

wagtail/core/tests/test_page_model.py
```diff
@@ -1490,6 +1491,43 @@ def test_copy_unsaved_page(self):
...
+
+        signal_fired = False
+
+        def page_published_handler(sender, instance, **kwargs):
+            nonlocal signal_fired
+            signal_fired = True
+        page_published.connect(page_published_handler)
+
+        homepage.copy(update_attrs={'slug': 'new_slug'})
+        self.assertIs(signal_fired, False)
```
kaedroho, line 1529: You can replace this with ``self.assertFalse(signal_fired)``

wagtail/core/tests/test_page_model.py
```diff
@@ -1490,6 +1491,43 @@ def test_copy_unsaved_page(self):
...
+
```
kaedroho, line 1530: I think we should add a test for the case where the page being published is live, but the ``keep_live`` flag was set to ``False``, please could you add that?

kaedroho (commented):
Apart from a couple of minor tweaks, this looks good to me, thanks again!

kaedroho (approved):
Nice work! Thanks again

PR #6437, opened 2020-10-05
Update wagtail app list to match installed apps in documentation.

### LINKED ISSUE #6435: The names of the Wagtail apps in the "How to add Wagtail into an existing Django project" docs are outdated
See: https://docs.wagtail.io/en/stable/advanced_topics/add_to_django_project.html#wagtail-apps

This app listing doesn't match the names in ``INSTALLED_APPS``, this is because the app listing wasn't updated when we renamed them in Wagtail 2.0 (see [RFC 12](https://github.com/wagtail/rfcs/blob/master/text/012-rationalise-apps.md)).

saevarom:
Updated the app names to match `INSTALLED_APPS`.
The `wagtail.sites` app was missing so I added a reference to that too.

Fixes #6435

### DIFF
--- a/docs/advanced_topics/add_to_django_project.rst
+++ b/docs/advanced_topics/add_to_django_project.rst
@@ -83,34 +83,37 @@ Wagtail requires several Django app modules, third-party apps, and defines sever
 Wagtail Apps
 ------------
 
-``wagtailcore``
+``wagtail.core``
   The core functionality of Wagtail, such as the ``Page`` class, the Wagtail tree, and model fields.
 
-``wagtailadmin``
+``wagtail.admin``
   The administration interface for Wagtail, including page edit handlers.
 
-``wagtaildocs``
+``wagtail.documents``
   The Wagtail document content type.
 
-``wagtailsnippets``
+``wagtail.snippets``
   Editing interface for non-Page models and objects. See :ref:`Snippets`.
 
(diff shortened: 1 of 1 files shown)


docs/advanced_topics/add_to_django_project.rst
```diff
@@ -83,34 +83,37 @@ Wagtail requires several Django app modules, third-party apps, and defines sever
...
+``wagtail.embeds``
   Module governing oEmbed and Embedly content in Wagtail rich text fields. See :ref:`inserting_videos`.
 
-``wagtailsearch``
+``wagtail.search``
   Search framework for Page content. See :ref:`wagtailsearch`.
 
-``wagtailredirects``
+``wagtail.sites``
+  The Wagtail site content type.
```
kaedroho, line 111: The content type is actually in Wagtail core. This app just provides a management UI

saevarom: Oh, right. I just thought it would make sense to include it in the list, since it was in `INSTALLED_APPS`. We could change the text to "Management UI for Wagtail sites"?

kaedroho: Yep, that sounds good to me, thanks!

kaedroho (approved):

PR #6440, opened 2020-10-06
move generate_signature and verify_signature from images/views/serve …

### LINKED ISSUE #4347: wagtail.images.views.serve.generate_signature should be relocated
I've encountered an issue with wagtail.images.views.serve.generate_signature.  When importing it into a models.py file, the class wagtail.images.views.serve.ServeView is created, which invokes a call to get_image_model, which in turn leads to a AppRegistryNotReady exception.

The functions generate_signature and verify_signature are not used at all in wagtail.images.views.serve. In fact, generate_signature is used in wagtail itself only in wagtail.images.views.images, and verify_signature is not used at all. It would make more sense to move these two functions to wagtail.images.views, or wagtail.images.utils, so that importing them does not result in the aforementioned side-effects.

One could then import these functions into wagtail.images.views.serve so as not to break existing code.

nthall:
…to images/utils

Fixes #4347

### DIFF
--- a/wagtail/images/tests/test_admin_views.py
+++ b/wagtail/images/tests/test_admin_views.py
@@ -9,7 +9,7 @@ from django.utils.http import RFC3986_SUBDELIMS, urlquote
 
 from wagtail.core.models import Collection, GroupCollectionPermission
 from wagtail.images.models import UploadedImage
-from wagtail.images.views.serve import generate_signature
+from wagtail.images.utils import generate_signature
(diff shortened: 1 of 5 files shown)


tomdyson:
Thanks @nthall!

kaedroho (approved):
LGTM, thanks!

kaedroho:
@nthall 

Sorry I forgot to ask earlier. How would you like your name written in the release notes?

nthall:
@kaedroho the way you already have it in that commit is fine 👍

Issue #6439, opened 2020-10-06
Checkbox and radio inputs visually off in admin


cole:
### Issue Summary

Checkboxes and radio buttons look broken in 2.10.2. Downgrading to 2.10.1 fixes the issue.

<img width="206" alt="Screen Shot 2020-10-05 at 7 08 07 PM" src="https://user-images.githubusercontent.com/134694/95151131-b884a680-073e-11eb-96ef-4d896c35ee75.png">

Notice the grey box and the alignment of the checkbox.
This seems to have been caused by adding `:not([type=submit])` to inputs, increasing the selector specificity.

### Steps to Reproduce

1. Have a page with a boolean field visible via Wagtail admin.
2. View the checkbox on the admin page

### Technical details

* Python version: 3.7.7
* Django version: 3.0.10
* Wagtail version: 2.10.2
* Browser version: Safari 14 / Chrome 85 / Firefox 81

lb-:
Fixed via #6441 

![Screenshot_2020-11-22 Wagtail - Editing Form page Contact Us](https://user-images.githubusercontent.com/1396140/99899995-107c6a00-2cf9-11eb-9a69-8aae886a924d.png)
![Screenshot_2020-11-22 Wagtail - Editing Breads index page Breads](https://user-images.githubusercontent.com/1396140/99899997-11ad9700-2cf9-11eb-87c6-af1278a26ae5.png)

Issue #6421, opened 2020-10-01
Checkboxes broken on Edge (legacy)


kaedroho:
Checkboxes are being displayed like this on Edge (and they don't appear to do anything when clicked):

![image](https://user-images.githubusercontent.com/1093808/94793681-c4850700-03d2-11eb-8639-0d529a34459f.png)

I'm testing with:
Microsoft Edge 44.19041.423.0
Microsoft EdgeHTML 18.19041

Wagtail master with i18n and alias pages

Note that Edge Legacy is the default browser that ships with Windows 10

loicteixeira:
Maybe related to #6439?

kaedroho:
I think it might be actually! I'll give that PR another try on IE and see if it fixes it

kaedroho:
This doesn't affect IE11, or the Chromium version of Edge. Only Edge Legacy

lb-:
Fixed with #6441 
Screenshot below (Edge 18)
![download](https://user-images.githubusercontent.com/1396140/99899944-90560480-2cf8-11eb-9783-c8419fbf7a04.png)

PR #6441, opened 2020-10-06
Reset padding for checkbox & radio inputs

cole:
Previously, the normalize stylesheet covered this, however that's no longer applicable due to selector specificity.

Works in Safari 14, Chrome 85 and Firefox 81.

Note; this doesn't fix the `full` versions of these input types, which also may look odd(ref: https://github.com/wagtail/wagtail/commit/1bbf541d7aa14b51a113c074dce9f17f97e43ab1). I'm not sure those ever were properly styled, though.

### DIFF
--- a/client/scss/elements/_forms.scss
+++ b/client/scss/elements/_forms.scss
@@ -154,6 +154,7 @@ input[type=checkbox] {
     border-radius: 0;
     cursor: pointer;
     border: 0;
+    padding: 0;
 }
 
 input[type=radio] {


cole:
Fixes #6439.

kaedroho:
Just checked and this fixes https://github.com/wagtail/wagtail/issues/6421 too!

kaedroho (approved):

kaedroho:
@gasman this one solves an issue that was introduced in 2.10.2 so it might be worth backporting if we decide to make another patch release

PR #6445, opened 2020-10-06
[FIX]: Document chooser’s "Search" tab broken after a failed upload

### LINKED ISSUE #6126: Document chooser’s "Search" tab broken after a failed upload
### Issue Summary

The document chooser has two separate tabs: "Search" to select existing documents, and "Upload" to create a new document. If the user goes to the upload tab and encounters a validation error, then moves back to Search, the tab is partially broken:

document-chooser-bypass

- The search form disappeared
- Collections filtering disappeared
- Pagination no longer works – all of the site’s documents are displayed at once (on larger sites with hundreds of documents this might make the upload form validation response take a lot of time)
- Documents are displayed without having been processed by the [`construct_document_chooser_queryset` hook](https://docs.wagtail.io/en/v2.9/reference/hooks.html#construct-document-chooser-queryset)
- We’re also missing a few more things that are cosmetic only (`documents_exist`, `is_searching`)

This is due to the chooser’s implementation which always has to re-create all tabs whenever interacting with either of the tabs. The upload view’s "failed validation" response does not correctly set the template context for all of the above to work:

https://github.com/wagtail/wagtail/blob/6e9bcef5db84d254686811b27fee0bf8c0210f9a/wagtail/documents/views/chooser.py#L154-L160

### Steps to Reproduce

1. On any Wagtail site, open the document chooser. See what’s displayed in the "Search" tab.
2. Switch to the Upload tab
3. Submit, receiving a validation error
4. Switch to the Search tab

### Technical details

* Wagtail version: v2.7.1 and v2.9

---

## Security considerations & suggested solution

There are security implications due to bypassing `construct_document_chooser_queryset` – this makes it possible for a malicious user to see all documents on a given Wagtail instance, even if otherwise set up with collection-level permissions and the queryset hook to restrict this for the choosers as well. Discussing this with @gasman, the hook doesn’t come with particular security guarantees so this is only tracked as a bug rather than "security vulnerability".

In an ideal world we would change the choosers’ API to only re-render separate tabs rather than the whole modal. Practically, fixing the existing implementation is a matter of replicating processing done in the main chooser view in the `upload` view:

https://github.com/wagtail/wagtail/blob/6e9bcef5db84d254686811b27fee0bf8c0210f9a/wagtail/documents/views/chooser.py#L54-L58

and 

https://github.com/wagtail/wagtail/blob/6e9bcef5db84d254686811b27fee0bf8c0210f9a/wagtail/documents/views/chooser.py#L91-L111

Similar changes have been done to the images’ chooser view (which had the same issues, but had been fixed) – so it might be worth borrowing the image chooser’s implementation, and tests

- https://github.com/wagtail/wagtail/blob/6e9bcef5db84d254686811b27fee0bf8c0210f9a/wagtail/images/views/chooser.py
- https://github.com/wagtail/wagtail/issues/4548
- https://github.com/wagtail/wagtail/pull/4553
- https://github.com/wagtail/wagtail/commit/cd0f8381cddb875172d15bc904fca5ebeadc43fe

samgans:
Fixes #6126

### DIFF
--- a/wagtail/documents/views/chooser.py
+++ b/wagtail/documents/views/chooser.py
@@ -17,7 +17,7 @@ from wagtail.search import index as search_index
 permission_checker = PermissionPolicyChecker(permission_policy)
 
 
-def get_chooser_js_data():
+def get_chooser_json_data():
     """construct context variables needed by the chooser JS"""
     return {
         'step': 'chooser',
@@ -42,18 +42,45 @@ def get_document_result_data(document):
     }
 
 
-def get_chooser_context(request):
+def get_chooser_context(request, documents, uploadform, **kwargs):
     """Helper function to return common template context variables for the main chooser view"""
-    collections = Collection.objects.all()
-    if len(collections) < 2:
-        collections = None
-
-    return {
-        'searchform': SearchForm(),
+    context = {
+        'uploadform': uploadform,
         'is_searching': False,
-        'query_string': None,
-        'collections': collections,
     }
+
+    for hook in hooks.get_hooks('construct_document_chooser_queryset'):
+        documents = hook(documents, request)
+
+    collection_id = request.GET.get('collection_id')
+    if collection_id:
+        documents = documents.filter(collection=collection_id)
+        context['collection_id'] = collection_id
+    else:
+        collections = Collection.objects.all()
+        if len(collections) > 2:
+            context['collections'] = collections
+        else:
+            context['collections'] = None
+
+    context['documents_exist'] = documents.exists()
+
+    searchform = SearchForm(request.GET)
+    if searchform.is_valid():
+        q = searchform.cleaned_data['q']
+        documents = documents.search(q)
+
+        context['query_string'] = q
+        context['is_searching'] = True
+    else:
+        context['searchform'] = SearchForm
+
+    paginator = Paginator(documents, per_page=10)
+    documents = paginator.get_page(request.GET.get('p'))
+
+    context['documents'] = documents
+    context.update(**kwargs)
+    return context
 
 
 def chooser(request):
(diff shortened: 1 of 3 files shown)


wagtail/documents/views/chooser.py
```diff
@@ -17,7 +17,7 @@
 permission_checker = PermissionPolicyChecker(permission_policy)
 
 
-def get_chooser_context():
+def get_chooser_js_data():
```
kaedroho, line 20: Could we change ``js`` to ``json`` here?

wagtail/documents/views/chooser.py
```diff
@@ -151,8 +165,21 @@ def chooser_upload(request):
 
     documents = Document.objects.order_by('title')
 
+    for hook in hooks.get_hooks('construct_document_chooser_queryset'):
```
kaedroho, line 168: Wondering if we could merge this in to ``get_chooser_context()`` as well as the search/pagination code from the ``chooser`` view so this isn't repeated?

kaedroho (commented):
Thanks again! Looks like we could move some more logic out into the ``get_chooser_context`` function so that there's no need to repeat that logic twice? Let me know if that isn't as easy as I think!

wagtail/documents/views/chooser.py
samgans: I've made this name to conform with the image's chooser method. So I think it'll be better to change this method for images chooser too. Is it ok?

kaedroho: Oh, I see. Yep I think it makes sense to have the same name in both places. Thanks!

wagtail/documents/views/chooser.py
samgans: I think so. Actually, the same changes can be made to the images chooser. Is it better to make changes in this pull request or create another one after the documents chooser will be updated?

kaedroho: It's slightly nicer to have them in the same PR so that changes to both get merged at the same time

views
```diff
@@ -0,0 +1 @@
+System check identified no issues (0 silenced).
```
kaedroho, line 1: Looks like this was accidentally committed

kaedroho:
Thanks for the updates! This is a bit of a tricky one to review now because the commits are mixing a lot of refactoring with code changes. Would it be possible to rearrange the commits so that there's one change per commit, with refactoring changes first?

I'd be able to rebase-merge rather than squash the commits if they were organised in this way

samgans:
I think I make wrong the history of commits. Can you give me any advice to edit all of this?

lb-:
The referenced issue was fixed in 4.0 - see;

- https://github.com/wagtail/wagtail/issues/6126#issuecomment-1326392333
- https://github.com/wagtail/wagtail/pull/8776

PR #6450, opened 2020-10-08
[FIX]: Inconsistent Handling of permission denied errors

### LINKED ISSUE #6352: Inconsistent Handling of permission denied errors
There's a couple of methods we use for handling permission denied errors:

 - When I attempt to edit a page without edit permission, I am given a 403 "Permission denied" response
 - When I attempt to edit a snippet without edit permission, I am redirected to the dashboard with a "Sorry, you do not have permission to access this area." message at the top.

I think we should use the latter approach in general, but perhaps not always go back to the dashboard (ideally should go to a passed `next_url`/`back_url` so the user is sent back to the view they came from)

samgans:
Fixed #6352 

Added:

- Tests of permission denied errors handling for images, documents, and snippets;

- Restriction on the snippets index view that doesn't let people without any valid permission on snippets view the snippets index page.


kaedroho:
Thanks for implementing this!

Just wondering if it would be simpler to catch ``PermissionDenied`` errors in our ``required_admin_access`` decorator, which is applied to all views. This would allow third party apps to continue to use ``PermissionDenied``. Also, exceptions can be a much better way to handle errors like this, if they are raised in a deep bit of logic, they won't need to be converted by the view.

I think we could replace [Line 168 of auth.py](https://github.com/wagtail/wagtail/blob/80c65390a4a9f67a38672c2d604ee97b441fdce5/wagtail/admin/auth.py#L168) with:

```python
try:
except PermissionDenied:
    if request.is_ajax():
        raise

    response = permission_denied()
```

This would allow third-party apps to continue to raise ``PermissionDenied`` errors, but now they are automatically converted to a redirect.

samgans:
> Thanks for implementing this!
> 
> Just wondering if it would be simpler to catch `PermissionDenied` errors in our `required_admin_access` decorator, which is applied to all views. And convert it to a redirect automatically?
> 
> I think we could replace [Line 168 of auth.py](https://github.com/wagtail/wagtail/blob/df8e51a6e9c391d2b8294bb3db71bc5d194a5d4d/wagtail/admin/auth.py#L168) with:
> 
> ```python
> try:
>     response = view_func(request, *args, **kwargs)
> except PermissionDenied:
>     if request.is_ajax():
>         raise
> 
>     response = permission_denied()
> ```
> 
> This would allow third-party apps to continue to raise `PermissionDenied` errors, also exceptions can be a lot simpler to use when the permissions logic isn't all in the view.

Yes, it would be simpler, but there are two types of permission errors in the views: one that related with user permissions and others related to the ability of creation child pages (the restriction given when we define the page). So handling the second types of errors with 'You don't have permissions to access this area" will be a little inconsistent too, I think. Do you have any suggestion of how to resolve those types of PermissionDenied?

gasman:
@samgans I don't think there's any inconsistency there - in the approach that @kaedroho is proposing, all types of 'permission denied' condition will raise a PermissionDenied exception and be caught by `required_admin_access`. However, it is a more wide-ranging change since it affects all the other apps (modeladmin, wagtail.contrib.forms) that currently raise PermissionDenied. I'm working on that now, and will raise a new PR based on this one shortly.

gasman:
Updated version now opened: #6459

Issue #5950, opened 2020-04-13
Unable to add parameters to mailto links


phette23:
### Issue Summary

Using the rich text editor, you cannot add additional URI parameters to a `mailto` link such as subject, bcc, cc, etc. You can only pass a single email address.

### Steps to Reproduce

1.  Create a page or other model with a rich text field, then edit an instance of it
2. Add text, then hyperlink it
3. Try to add an email link with additional parameters e.g. `help@example.com?subject=Help+Me` or `help@example.com?cc=support@example.com`
4. You're told the email is invalid
5. You're also unable to add an analogous URI but with the "mailto:" scheme under External Link

I think the solution could be to validate not against an email address, but a mailto URI. I looked at the Draftail repo first but decided to open the issue here, since it seems that the link selection dialog is native to Wagtail and not the editor itself; forgive me if I'm wrong about that.

Any other relevant information. For example, why do you consider this a bug and what did you expect to happen instead?

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: (yes / no)

No, but we are on the latest release (2.8).

### Technical details

* Python version: 3.5.2
* Django version: 2.2.11
* Wagtail version: 2.8
* Browser version: Chrome 80

gasman:
Hi @phette23 - thanks for the suggestion. A potential issue with having users enter mailto: links is that editor-level users are not necessarily technical, and may not know how to construct them. I think the way forward would be to have a separate optional field for subject / cc in the link chooser, which would then store the link internally as something like `<a linktype="email" email="foo@example.com" subject="Help me">` - we can then use the existing link-rewriting logic (as used for page and document links) to convert this into a working mailto: URL.

lb-:
PR raised a while ago with a partial solution - would be good if anyone is keen to pick that up and redo on the latest code + fix one of the minor feedback items.

https://github.com/wagtail/wagtail/pull/6451

lb-:
#10920 has implemented the subject/body request in this issue, we will close when merged.

If support for cc/bcc is still requested, that can be requested as a future enhancement issue, it will likely involve a bit of InlinePanel to allow for multiple adding/re-ordering of a CC and/or BCC address.

PR #6451, opened 2020-10-08
5950 add parameters to mailto links

mekanics:
This PR adds a `subject` and `body` field to the email link in the RichText Editor. 

It does not completely solve #5950 as I didn't add the `cc` & `bcc` fields. The reason for this, is that the standard allows multiple email address in the 'cc' and 'bcc' (actually in the `to` as well – and it should be optional, but that's another story 😉 )

### DIFF
--- a/wagtail/admin/views/chooser.py
+++ b/wagtail/admin/views/chooser.py
@@ -15,10 +15,10 @@ from wagtail.core.utils import resolve_model_string
 try:
     import urlparse
     from urllib import urlencode
-except: # For Python 3
+except ImportError:  # For Python 3
     import urllib.parse as urlparse
     from urllib.parse import urlencode
-    
+
 
 def shared_context(request, extra_context=None):
     context = {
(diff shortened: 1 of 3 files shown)


jandor64:
Will we be seeing this in a version soon? Thanks!

daveclarkit:
+ 1 for getting this into release!

wagtail/admin/views/chooser.py
```diff
@@ -12,6 +12,13 @@
 from wagtail.core.models import Page, UserPagePermissionsProxy
 from wagtail.core.utils import resolve_model_string
 
+try:
+    import urlparse
+    from urllib import urlencode
+except ImportError:  # For Python 3
```
zerolab, line 18: drive-by comment: we only support Python 3

lb-:
@mekanics sorry this never got a review, are you keen on rebasing/pursuing this still? Would be great to get this up to date and fix the feedback item above.

chrismattmann:
we could really use this in my project at the day job. Any ideas on a workaround until this PR is merged?

lb-:
Superseded by #10920

PR #6453, opened 2020-10-09
Fix privacy switcher panel padding

### LINKED ISSUE #6446: Privacy button cut off in settings if there are additional panels
### Issue Summary

Privacy button is cut off at the bottom if there are more panels after it. This is in part because of `.field-row`'s negative margin bottom, but removing that still leaves it without white space between the button and the next panel.
 
image

### Steps to Reproduce

Add another `FieldPanel` to the `settings_panels` after the privacy panel

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: **No**

### Technical details

* Python version: 3.6.4
* Django version: 3.0.7
* Wagtail version: 2.10.2
* Browser version: Firefox 82, Chrome 85

zerolab:
This PR fixes #6446

The issue was in a class combination (resulting in negative bottom margin being applied)

Here's a screenshot with the fix applied.
Screenshot 2020-10-09 at 18 41 59

I tested across several browsers, but worth double checking

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/scss/layouts/page-editor.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/layouts/page-editor.scss
@@ -127,7 +127,8 @@
     }
 
     fieldset,
-    .field-row {
+    .field-row,
+    .object-content-wrapper {
         padding-top: $object-title-height + 12px;
     }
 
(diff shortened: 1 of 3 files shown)


zerolab:
@KalobTaulien #6454 provides an alternative implementation, if you don't mind checking it out too and weigh in to which one to use..

KalobTaulien:
@zerolab I just reviewed #6454 and it works well as well. I'm not 100% familiar with the accessibility parts, but it certainly acts the way it should in that PR. This PR works well too. I'll let you decide which one gets merged ;)

gasman:
Fixed in #6454

PR #6454, opened 2020-10-09
Fix spacing around the privacy indicator panel.

### LINKED ISSUE #6446: Privacy button cut off in settings if there are additional panels

saevarom:
Fixes #6446 

I did two things:
* Wrapped the whole privacy indicator in a `fieldset` tag. It seems that most field panels do this anyway so this creates a uniform way of spacing things for a panel. 
* Removed the `field-row` class which gives a negative bottom margin.

After the change, this is how the panel looks with additional `FieldPanels`:

Screenshot 2020-10-09 at 17 40 07

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/pages/privacy_switch_panel.html
+++ b/wagtail/admin/templates/wagtailadmin/pages/privacy_switch_panel.html
@@ -6,22 +6,24 @@
 {% endif %}
 
 {% if not page.is_root %}
-<div class="field-row privacy-indicator {% if is_public %}public{% else %}private{% endif %}">
-    <p>
-        {% trans "Current status:" %}
+<div class="privacy-indicator {% if is_public %}public{% else %}private{% endif %}">
+    <fieldset>
+        <p>
+            {% trans "Current status:" %}
 
-        <span class="label-public icon icon-view" aria-label="{% trans 'Page privacy. Current status: Public' %}">
-             {% trans "Public" %}
-        </span>
-        <span class="label-private icon icon-no-view" aria-label="{% trans 'Page privacy. Current status: Private' %}">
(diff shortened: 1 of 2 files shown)


zerolab:
:) looks like we cross-PR-ed (is that a thing?) - #6453
I started off with wrapping the panel content in `fieldset`, but the spec/a11y is a bit ambiguous so decided againsts it as the contents are not rield fields

@thibaudcolas maybe you can advise? happy with either solution

thibaudcolas:
Semantically I don’t think `fieldset` is correct, but the screen reader testing I did indicates it could be used nonetheless as it’s ignored. I assume due to there not being a `legend`. I think I’d go with whatever feels the most pragmatic. Those styles are in bad need of a rewrite, it’s completely crazy to apply styles to bare elements like this in a UI that’s going to be extended by other people like field panels.

KalobTaulien (approved):
This works well, thank you for this PR! 🎉

zerolab:
@gasman this or #6453 would be good to get in 2.11 as it fixes a styling issue

gasman:
Merged in b7643f869d64e02624724eabb3a6b9ef567249be (went for this one as it sounds like it's had slightly more eyes on it :-) )

PR #6456, opened 2020-10-12
Add greeting for contributors on raising their first PR or issue

winter-code:
Messages ( for issues and PRs) can be customized. This is recommended since a hello message on successfully raising a PR will motivate new contributors. :rocket:

### DIFF
--- a/.github/workflows/greetings.yml
+++ b/.github/workflows/greetings.yml
@@ -10,4 +10,4 @@ jobs:
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
         issue-message: 'Thanks for submitting your first issue to Wagtail!'
-        pr-message: 'Thanks for submitting your first PR to Wagtail!'
+        pr-message: 'Thanks for submitting your first PR to Wagtail! We know how hard you worked on this. Please give us some time to review, and we'll be in touch as soon as we can.'
--- a/.github/workflows/greetings.yml
+++ b/.github/workflows/greetings.yml
@@ -9,5 +9,5 @@ jobs:
     - uses: actions/first-interaction@v1
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        issue-message: 'Here's a :candy: while we look into your'' first issue'
-        pr-message: 'Kudos! You have successfully raised your'' first pr'
+        issue-message: 'Thanks for submitting your first issue to Wagtail!'
+        pr-message: 'Thanks for submitting your first PR to Wagtail!'
--- a/.github/workflows/greetings.yml
+++ b/.github/workflows/greetings.yml
@@ -1,0 +1,13 @@
+name: Greetings
+
+on: [pull_request, issues]
+
+jobs:
+  greeting:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/greetings.yml
```diff
@@ -0,0 +1,13 @@
...
+on: [pull_request, issues]
+
+jobs:
+  greeting:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/first-interaction@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        issue-message: 'Here's a :candy: while we look into your'' first issue'
```
tomdyson, line 12: I think the apostrophe in `Here's` needs to be escaped with a second single quote, and there shouldn't be two single quotes after `your`. Colons also have tricky escape sequences:

https://yaml.org/spec/1.2/spec.html#id2788097

Perhaps we should go for a more straightforward

> 'Thanks for submitting your first issue to Wagtail!'

and 

> 'Thanks for submitting your first PR to Wagtail!'

tomdyson (changes requested):
Thanks @winter-code! This will be a nice welcome to new contributors. I've suggested a simpler message, partly to help with tricky yaml escaping.

.github/workflows/greetings.yml
winter-code: Looks good! This is my first time working with yaml files and I did try out these messages on my own repos but wasn't sure what was wrong with the message that the :candy: didn't appear (which was sad)

Making the changes as suggested by you! :)

lb-:
@vossisboss - when you get a min - do you mind reviewing this in the context of community. 

I think it would be great to have this in but it's been a while since this was created so maybe cast your eyes over it.

.github/workflows/greetings.yml
```diff
@@ -0,0 +1,13 @@
...
+
+jobs:
+  greeting:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/first-interaction@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        issue-message: 'Thanks for submitting your first issue to Wagtail!'
+        pr-message: 'Thanks for submitting your first PR to Wagtail!'
```
vossisboss, line 13: I think this is a great idea! I'd just like to suggest we give the PR submitters some reassurance that we will review their work as soon as we can.

```suggestion
        pr-message: 'Thanks for submitting your first PR to Wagtail! We know how hard you worked on this. Please give us some time to review, and we'll be in touch as soon as we can.'
```

vossisboss (changes requested):
Requesting one small change. Otherwise I think this is a great contribution!

winter-code:
I will make the necessary changes! :D

thibaudcolas:
Thank you for giving this a go @winter-code! We’ve just discussed this at our weekly core contributor team meetings, and have decided to not go further with this. We think it’s better for us to personalise our greetings / thanks to contributors, particularly new contributors.

There is a chance we might not do so consistently but we think this is worth the tradeoff, since personal greetings by other contributors will be more meaningful.

PR #6457, opened 2020-10-12
Update help text to use SVG icons

Scotchester:
This PR updates the icons in help text to use SVG icons.

As it turns out, the existing icon was some sort of construction made only with CSS (source [here](https://github.com/wagtail/wagtail/blob/82a2002f2d7463a5082908479c878c30e4a61192/client/scss/components/_icons.scss#L140-L156), [here](https://github.com/wagtail/wagtail/blob/82a2002f2d7463a5082908479c878c30e4a61192/wagtail/admin/static_src/wagtailadmin/scss/layouts/page-editor.scss#L158-L162), and [here](https://github.com/wagtail/wagtail/blob/82a2002f2d7463a5082908479c878c30e4a61192/client/scss/settings/_variables.icons.scss#L34-L35)). The icon font did not have a circled dark-on-light question mark, and the new set of SVGs does not, either.

So, I guess, strictly speaking, this update was not _necessary,_ except insofar as we plan to remove the whole suite of old icon-generation code when we finish the SVG transition. The approach I've taken here is to replace the constructed icon with the light-on-dark question mark SVG icon. This makes the icon more prominent, but I personally think that that's a usability improvement. I am not sure that the help text called enough attention to itself previously. But if anyone objects to this, we can roll it back.

before | after
--- | ---
existing inverse icon | new svg icon

- [x] For front-end changes: Did you test on all of Wagtail’s supported browsers?
  - [x] Chrome
  - [x] Firefox
  - [x] IE 11
  - [x] Safari
  - [x] Edge 18
  - [x] iOS Safari
  - [x] Android Chrome

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/scss/layouts/page-editor.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/layouts/page-editor.scss
@@ -151,6 +151,10 @@
         padding-left: 3em;
         opacity: 1;
 
+        .icon-help {
+            margin-left: -1.75em;
+        }
+
         .icon-help-inverse {
             float: left;
             margin-left: -2em;
--- a/wagtail/admin/static_src/wagtailadmin/scss/layouts/page-editor.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/layouts/page-editor.scss
@@ -154,12 +154,6 @@
         .icon-help {
             margin-left: -1.75em;
         }
-
-        .icon-help-inverse {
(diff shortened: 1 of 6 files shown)


allcaps (approved):
LGTM, and the icon change is an improvement 👍

allcaps (approved):

Scotchester:
Added a commit to remove the `help_text`-specific Sass for the previous implementation. Thibaud approved verbally via Slack. Merging!

PR #6459, opened 2020-10-13
Consistently redirect to admin home on permission denied

### LINKED ISSUE #6352: Inconsistent Handling of permission denied errors

gasman:
As per https://github.com/wagtail/wagtail/pull/6450#issuecomment-706556190. Supersedes #6450; fixes #6352.

### DIFF
--- a/wagtail/admin/auth.py
+++ b/wagtail/admin/auth.py
@@ -157,35 +157,43 @@ def require_admin_access(view_func):
             return reject_request(request)
 
         if user.has_perms(['wagtailadmin.access_admin']):
-            preferred_language = None
-            if hasattr(user, 'wagtail_userprofile'):
-                preferred_language = user.wagtail_userprofile.get_preferred_language()
-                l18n.set_language(preferred_language)
-                time_zone = user.wagtail_userprofile.get_current_time_zone()
-                activate_tz(time_zone)
-            if preferred_language:
-                with override(preferred_language):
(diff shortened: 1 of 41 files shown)


lb-:
Merged in via 41eccc72bcc68924d801e4048eef872ec5826897
Thanks @gasman

 @samgans  - I put your name as `Samgans` in the contributors list, is that ok or would you prefer something else?

samgans:
> Merged in via [41eccc7](https://github.com/wagtail/wagtail/commit/41eccc72bcc68924d801e4048eef872ec5826897)
> Thanks @gasman
> 
> @samgans - I put your name as `Samgans` in the contributors list, is that ok or would you prefer something else?

Hi! My name is already in the list of contributors, so there is no need to do so.

kaedroho:
Sorry, I should've made that clearer when I added it earlier. Fixed in https://github.com/wagtail/wagtail/commit/aa4a1e37b686cb7efcc90bb4ae2ec4d92eadf5b8

PR #6462, opened 2020-10-14
Customisable media forms

### LINKED ISSUE #6444: Support large document uploads
I asked a question in slack and didn't get a response.   I occasionally have large documents (i.e. product pdfs and such) that should be downloadable.  However, I don't want to allow arbitrarily large uploads through nginx or whatever frontend is in use.  I do not see a simple way to hook in here.

In [wagtailmedia we allowed the site to provide a custom base form](https://github.com/torchbox/wagtailmedia/blob/master/wagtailmedia/forms.py#L47) to set custom widgets to handle this case.  Would you accept this approach?

Since very large documents are out of the ordinary, it doesn't concern me that they are not compatible with the multi-uploader javascript so I would be happy if it was restricted to the single document upload form.  But it would be helpful if there was a handy link on the multi-upload form users reach by default to upload a single document.  And perhaps a way to nudge users to use this form if their upload fails because the file is too large

zerolab:
This PR introduces two new settings that allows customising the image/document forms. Closes #6444

### DIFF
--- a/docs/reference/settings.rst
+++ b/docs/reference/settings.rst
@@ -127,6 +127,14 @@ This setting lets you provide your own image model for use in Wagtail, which mig
 
 .. code-block:: python
 
+  WAGTAILIMAGES_IMAGE_FORM_BASE = 'myapp.forms.MyImageBaseForm'
+
+This setting lets you provide your own image base form for use in Wagtail, which might extend the built-in ``BaseImageForm`` class or replace it entirely.
+You can use it to specify or override the widgets to use in the admin form.
+
+
+.. code-block:: python
+
     WAGTAILIMAGES_MAX_UPLOAD_SIZE = 20 * 1024 * 1024  # i.e. 20MB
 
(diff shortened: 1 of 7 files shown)


thenewguy:
What is holding this up? Anything I can do to get it merged?

Scotchester:
Hey @thenewguy, could you test out this branch and let us know if it solves the issue for you?

thenewguy:
@Scotchester sure thing - could this make it into 2.11?

Scotchester:
@thenewguy Sorry, I don't believe so. The first release candidate is already out, so I think everything is frozen other than regressions or critical bug fixes.

gasman (approved):

gasman:
Merged in 05587e0d84ffd8795d461ac0ac7dc94ab1188296.

kaedroho:
I really hope we get to the point someday where we can solve issues like this more generally by letting people subclass the views/viewsets instead of having to add new settings/hooks every time a bit of one of these interfaces needs to be changed by someone!

PR #6463, opened 2020-10-14
Upgrade isort and add pre-commit support

zerolab:
This PR:
* upgrades isort to latest version
* and adds https://pre-commit.com/ configuration to aid local development.
  Time and time again I find myself forgetting to lint. With this I can `pre-commit install` then forget until it complains about the wrong sort order or missing imports.

Note: this is now  using the latest isort (5.x, [upgrade notes[(https://pycqa.github.io/isort/docs/upgrade_guides/5.0.0/))

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,13 @@
+repos:
+  - repo: https://github.com/timothycrosley/isort
+    # isort config is in setup.cfg
+    rev: 5.6.4
+    hooks:
+      - id: isort
+        language_version: python3
+  - repo: https://gitlab.com/pycqa/flake8
+    # flake8 config is in setup.cfg
+    rev: 3.8.4
+    hooks:
+      - id: flake8
+        language_version: python3
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,6 +20,8 @@ multi_line_output=4
 skip=migrations,project_template
 known_first_party=wagtail
 default_section=THIRDPARTY
+lines_between_types=1
+lines_after_imports=2
 
 [tool:pytest]
 django_find_project = false
--- a/setup.py
+++ b/setup.py
@@ -3,8 +3,9 @@
 from wagtail import __version__
 from wagtail.utils.setup import assets, check_bdist_egg, sdist
 
+
 try:
-    from setuptools import setup, find_packages
+    from setuptools import find_packages, setup
 except ImportError:
     from distutils.core import setup
 
@@ -55,7 +56,7 @@ testing_extras = [
     # For coverage and PEP8 linting
     'coverage>=3.7.0',
     'flake8>=3.6.0',
-    'isort==4.2.5',
+    'isort>=5.6.0,<6.0',
     'flake8-blind-except==0.1.1',
     'flake8-print==2.0.2',
     'doc8==0.8.1',
--- a/tox.ini
+++ b/tox.ini
@@ -60,5 +60,5 @@ setenv =
 
 [testenv:flake8]
 basepython=python3.6
-deps=flake8>=2.2.0
+deps=flake8>=3.6.0
 commands=flake8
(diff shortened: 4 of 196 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,13 @@
+repos:
+  - repo: https://github.com/timothycrosley/isort
+    # isort config is in setup.cfg
+    rev: 5.6.4
```
kaedroho, line 4: We still use version 4.2.5 which isn't compatible with isort 5.x

.pre-commit-config.yaml
```diff
@@ -0,0 +1,13 @@
...
+    hooks:
+      - id: isort
+        language_version: python3
+  - repo: https://gitlab.com/pycqa/flake8
+    # flake8 config is in setup.cfg
+    rev: 3.8.4
```
kaedroho, line 10: Same here, we have flake8 2.0.2 in setup.py. This is probably incompatible with 3.8

kaedroho:
>  May be time to bump up the version in setup.py (https://pycqa.github.io/isort/docs/upgrade_guides/5.0.0/)

Yep, sorry I looked at the code before reading the description. Definitely think they should match versions though!

.pre-commit-config.yaml
zerolab: `flake8>=3.6.0`, according to https://github.com/wagtail/wagtail/blob/master/setup.py#L57

.pre-commit-config.yaml
zerolab: updated to match.

.pre-commit-config.yaml
kaedroho: Oops, read the wrong line. Sorry!

kaedroho:
Just tried editing a random file (`wagtail/core/compat.py`) by adding `# foo` to the end, but `isort` gave an error.

```
$ pre-commit run
isort....................................................................Failed
- hook id: isort
- files were modified by this hook
flake8...................................................................Passed
```

I think most of our files are not necessarily compliant with all of isort's rules as it seems to have a concept of separate "soft" and "hard" errors. Normally, it seems to only report the soft errors if a hard error occured in that file. But pre-commit is reporting these errors if any change was made to the file.

Maybe it's time to update isort, fix all of these soft-errors, and disable this behaviour?

zerolab:
Flake8 should only give an error when one is found, but it will not do anything.
The isort check will fail if any issues, but it will also run isort, so trying to commit gain should work.

Will have a look at upgrading isort and isorting everything this afternoon.

Thank you for taking the time to look at this

zerolab:
@kaedroho ok, bit the bullet and bumped us to latest isort.
Needed a couple of config tweaks to bring it close to what we have now (but even with that, there were a number of file with inconsistent ordering / spacing)

kaedroho (approved):
Seems to work great now, thanks!

gasman:
Sorry to not catch this earlier...

* We should pin isort to a specific version in setup.py, because it has a habit of changing its rules between patch releases - see #3623
* The isort version number mentioned at in https://docs.wagtail.io/en/stable/contributing/python_guidelines.html needs updating
* On my vagrant box, upgrading from 4.2.5 to 5.6.4 causes it to go from 19 seconds to 2 and a half minutes :-(

gasman:
> * On my vagrant box, upgrading from 4.2.5 to 5.6.4 causes it to go from 19 seconds to 2 and a half minutes :-(

Looks like this is a combination of two factors:

* In 4.3.0 (https://github.com/PyCQA/isort/pull/604), they made the dubious and non-overrideable decision to start scanning non-`.py` files just in case they contained a `#!/usr/bin/python` line
* Our isort config in setup.cfg has a list of directories to skip, overriding the default setting which includes some that we definitely do want to skip, such as `node_modules` and `.git`

Updating setup.cfg with more aggressive `skip` and `blocked_extensions` settings gets it back under 25 seconds. Will open a PR shortly...

Issue #4488, opened 2018-04-21
Limit collections in image/document chooser based on permissions


robslotboom:
### Summary
When a user is assigned to a group with permissions to add/edit images and/or documents to a collection named “Collection One”, the chooser still has a select with options for all collections.

### Quick and dirty solution:

wagtail/images/views/chooser.py

```
def chooser(request):
	...
	collections = None
	groups = request.user.groups.all()
	if groups:
		collection_permissions = GroupCollectionPermission.objects.filter(group__in=groups)
		if collection_permissions:
			collections = [cp.collection for cp in collection_permissions.all()]

	if not collections:
		collections = Collection.objects.all()
	
	if len(collections) < 2:
		collections = None

	...
```

### Technical details

* Python version: 3.6.4.
* Django version: 2.0.4.
* Wagtail version: 2.0.1.

gasman:
Hi @robslotboom - thanks for the report.

The change you're proposing wouldn't be entirely correct - a user with no permission to add or edit images should still be able to select existing images to be inserted into their pages, and so it's intentional that the chooser ignores add/edit permission.

The correct way to fix this (and make it possible to limit a user's ability to choose images) would be to introduce a new 'choose' permission type. This was mentioned in passing in #4288, but it looks like we don't currently have an issue open to address this for images/documents, so I'll leave this one open as a feature request.

robslotboom:
Thanks Matt,

I see the problem.

The project I am working on does has a kind of ‘choose’ permission.
It’s a portal with sites and each site has a corresponding group and collection…

I’ll wait and hope there will be a ‘choose’ permission type in the future :-)

Best regards,


Robert Slotboom

marqpdx:
Yes, this is an important feature on one of our sites too. Users create a Member Profile for themselves, and need access to **only** their images, not that of all users.

+1 for this. 

thx,
m

marqpdx:
OK, followup.

The core of the issue seems to be that there doesn't appear to be a way to turn off permissions for a User/Group for the Root level Collection. This i think should be added.

For the code above, the following line returns Root collection, plus whatever other Collections the user has permissions to. 
```
	if len(collections) < 2:
		collections = None
```
So even if a user is part of a specific collection, the drop down will always include the Root collection. This inclusion of root level is hard-coded in:`wagtail/images/templates/wagtailimages/chooser/chooser.html` which includes `

So it seems we could add a permission to deny access to the Root Collection, and we'd be a lot closer.

[...]

Issue #4444, opened 2018-04-04
Built-in command `remove_stale_contenttypes` removes `wagtail_admin` permissions


emorozov:
### Issue Summary

When running `remove_stale_contenttypes` it offers to remove `wagtail_admin.admin` content type and all related permissions and all assigned permissions making access to Wagtail admin impossible to anyone but superuser.

### Steps to Reproduce

Run `./manage.py remove_stale_contenttypes`. It will offer:

```
Some content types in your database are stale and can be deleted.
Any objects that depend on these content types will also be deleted.
The content types and dependent objects that would be deleted are:

    - Content type for wagtailadmin.admin
    - 1 auth.Permission object(s)
    - 2 auth.Group_permissions object(s)

This list doesn't include any cascade deletions to data outside of Django's
models (uncommon).

Are you sure you want to delete these content types?
If you're unsure, answer 'no'.
Type 'yes' to continue, or 'no' to cancel:
```

### Technical details

* Python version: 3.5.3
* Django version: 2.0.3
* Wagtail version: 2.0

I suggest using more robust way of creating `wagtail_admin` content type because it is very easy to miss and accidentally remove it along with the permissions if there're lots of stale content types and you just type `yes` to all questions thinking that there're no content types that are actually used by the project.

Maybe there should be a model in Wagtail or wagtailadmin permissions should be tied to some existing model, like `Page`, so that Django doesn't consider wagtailadmin content type unused, stale and safe to remove.

gasman:
For anyone who's encountered this, you can restore the missing permission types by running the following from the `./manage.py shell` command line:

    from django.contrib.contenttypes.models import ContentType
    from django.contrib.auth.models import Permission
    wagtailadmin_content_type, _ = ContentType.objects.get_or_create(app_label='wagtailadmin', model='admin')
    Permission.objects.get_or_create(content_type=wagtailadmin_content_type, codename='access_admin', name='Can access Wagtail admin')

adsk2050:
I faced the same issue, but I never tried the remove_stale_contenttypes command. I have implemented a CustomUser model though. I am trying to find the steps to reproduce. But for sure this issue is deeper than this. Will update once I find the problem

richkirsch:
Would be great to see this fixed. Running `remove_stale_contenttype` is useful in many automated deployment scenarios.

As it stands this issue leads to an extremely opaque debugging problem ... I was relieved to find this thread.

berekuk:
[...]

PR #6466, opened 2020-10-15
Implemented a 'choose' permission for Images and Documents

coredumperror:
This PR adds the 'choose' permission discussed in #4488. It limits Image and Document choosers to showing only those Collections (and their contents) for which the current user has the Choose permission.

It grants Choose permissions on the Root Collection to every existing Group that has the "Can access Wagtail admin" perm, via data migration. So there will be no change from previous behavior until the user goes in and alters the Choose perms for their site's Groups.

This PR has no effect on anything except choosers. The Image and Document index pages are still controlled solely by the 'add' and 'change' permissions, as those are only accessible by users who have those permissions in the first place. This means that it's now possible to configure a Group to allow its members to choose images from only specific Collections when editing pages, *without* them being able to edit said images or add new ones to said Collections.

However, this does add one wrinkle that I'd like to fix, but I'm not entirely sure how. I'd like for the `AdminImageChooser` and `AdminDocumentChooser` widgets to hide the "Edit this image/document" button if the user only has Choose permission. But since low level stuff like the widget constructor code doesn't have access to the Django request, I'm not sure how to actually check if the current user has said permission, because I have no access to the current user. There's no obvious way I can see to pass the request (or just the user) from the view down into the widget constructor (or the `render_html()` method).

Anyway, I've written the code that will be able to implement edit button hiding, *if* I can find a way to get the User object. But since I don't know how, it's commented out.

I will add docs and tests for the new choose permission functionality in the next commit. One existing test fails due to my changes, which I will fix ASAP.

There are also several linter errors, but only in files I didn't touch, so I'm going to leave them alone.


coredumperror:
OK, new commit fixes the lint errors by merging with the newest master, and fixes the error with the existing test by having that test code add the necessary Choose permission.

coredumperror:
By the way, I discovered a bug in how the Group Collection permissions formsets are rendered.

I discovered the issue because of how I originally wrote the changes to `documents/forms.py` and `images/forms.py`. `('choose_image', _("Choose"), _("Select images in Choosers")),` was originally at the top of the permissions list, but that caused a really baffling bug in the form. It displayed the checkboxes as "Choose   Add   Edit", but when you actually checked them and saved the form, it would save the permissions as if the order was "Add   Edit   Choose".

Thus, checking "Choose" would actually give the "Add" permission. Figuring out why the heck this was happening was a frustrating process, but I eventually hypothesized that the issue might be due to alphabetical storing, so I put "choose_image" *after* "Edit" (which is actually "change_image"), and that fixed it.

My guess is that this has always been a bug in `collection_member_permission_formset_factory()`, but it's never manifested because the only two permissions that existed previously were coincidentally placed in alphabetical order. I believe the perpetrator is the first code in that function: 
```    
    permission_queryset = Permission.objects.filter(
        content_type__app_label=model._meta.app_label,
        codename__in=[codename for codename, short_label, long_label in permission_types]
    ).select_related('content_type')
```

This ends up making the form code ignore the order from `permission_types` and instead use the default ordering of the `Permission` model, which orders permissions for the same model in "codename" order. However, the template that renders the checkbox's labels gets them directly from `permission_types`. Thus, the order is preserved for the table headers, but not the checkboxes.

Since this PR is why this bug was discovered, should I fix it in this PR? Or should it be fixed in a separate one? For now, the PR can work despite the bug, by just putting "Choose" at the end of the list of checkboxes. Though it does feel more natural to put it at the start.

ababic:
@coredumperror Thanks very much for getting going with this. This is valuable start, but I still think there are some conversations to be had about how we this should work.

1.  Should a user need the the 'choose' permission explicitly in order to choose from a Collection?

If a user has the `create` or `change` permission for a collection (either directly, or via permissions set on an ancestor), couldn't we assume the `choose` permission? This would make the 'choose' permission a bit of a special case (not unlike Django's recently added 'view' permission), in that it is a lesser permission than `create` or `change`, that is only really ever meaningfully granted to users who aren't trusted enough to wield 'manage' the media. I personally feel this is okay, but would am keen to hear other points of view.

2. Should all of the functionality behind the 'Images' and 'Documents' menu items be reserved for those with the `create` or `change` permission for a collection, or would there be value in a 'read-only' version of those views?

This isn't necessarily something that would need to be done in the same PR, but in Layman's terms, by granting the 'choose' permission, aren't we essentially saying that users in that group can "See that the collection exists" and "Use media from the collection". When we think about that user's experience within the CMS, are there things that could be useful for them to 'see' in regards to that media, that might not be surfaced in the 'choose' UI? And is `choose` the correct verb for this permission after all (perhaps `view` or `browse` makes more sense)?

wagtail/documents/migrations/0011_add_choose_permissions.py
```diff
@@ -0,0 +1,92 @@
...
+    Permission = apps.get_model('auth.Permission')
+    document_content_type = ContentType.objects.get(
+        model='document',
+        app_label='wagtaildocs',
+    )
+    # This cascades to Group
+    Permission.objects.filter(
+        content_type=document_content_type,
+        codename='choose_document'
+    ).delete()
```
ababic, line 43: It's fine to create the permission to use in the data migration (as usually, it wouldn't be created until Django's 'post_migrate' signal is emitted), but I don't think we want to DELETE it here. We're better off letting Django take care of that.

ababic:
Should officially define the new permission on the `Meta` class of `Image` and `Document`? This way, if for any reason the `Permission` object is deleted, Django should automatically detect that and recreate it. Or, perhaps we can just reuse the `view_image` and `view_document` permissions that Django will have created automatically for these models?

wagtail/documents/migrations/0011_add_choose_permissions.py
coredumperror: I just copied what the images and documents apps do with the original `add_document`, `change_image`, etc. permissions in their respective backwards migrations. See `wagtail/images/migrations/0001_squashed_0021.py` and `wagtail/documents/migrations/0002_initial_data.py`. I wrote this PR's migration based on what those two files do.

As for "letting Django take care of that", what do you mean? In my experience, Django is notorious for *not* deleting permissions for models that have been removed from the database.

coredumperror:
> Should we officially define the new permission on the Meta class of `Image` and `Document`?  

The Image and Document models don't currently define the Add and Edit permissions in their Metas. Should those be defined there, too? That's probably the "right" way to do it, but I'm not sure how it'd interact with established Wagtail projects. And defining permissions in multiple different ways sounds like a disaster waiting to happen.

> Should a user need the 'choose' permission explicitly in order to choose from a Collection?

Having the Add or Edit permissions act as a proxy for the Choose permission just sounds confusing to me, now that a separate Choose permission actually exists. An admin with permission to edit Groups might look at a Group's perms and say "Wait a minute, Suzy told me she can choose images from this Collection, but her group doesn't have the Choose perm!"

Sure, having Edit and _not_ Choose would be weird a weird state, but this PR already gives everyone the Choose permission on the Root Collection (to retain existing behavior), so why would anyone remove Choose but leave Add or Edit?

> Should all of the functionality behind the 'Images' and 'Documents' menu items be reserved for those with the add or change permission for a collection, or would there be value in a 'read-only' version of those views?

I considered this, but treated it as outside the scope of this PR. I figure that the entire purpose of the Index views for Images and Documents are for adding new ones and editing existing ones. We'd need essentially a whole new "read only index view" for people who have Choose but neither Edit nor Add, and we'd want to make it really clear that the read-only one a completely different from the add/edit one, or users are likely to get quite confused. 

Plus, what about the case where Suzy has add/edit permission Collection 2, but *only* Choose on Collection 3 (which is a sibling of Collection 2)? The Index view would need to show Images in Collection 3 as "read only" but images from Collection 2 as editable, even on the "view all images" version of the index. That'd get real confusing, real fast.

This is one of the reasons that I figure "Choose" is the right name for this permission, because it's *specifically* for Choosers. It is not affiliated with the Index pages at all, so there's no way for confusion to occur on the Index pages for people with disparate permissions on separate Collections. 

This also ties back into the "Should a user need the 'choose' permission explicitly in order to choose from a Collection?" question. This PR takes the implied "Choose" permission out of the Edit and Add perms and makes it a completely separate thing.

coredumperror:
It's been a while since the last comment around here. I'm really hoping we can get this PR integrated into Wagtail 2.12, so if there's anything I need to do to help that happen, please let me know. :)

gasman (approved):
> > Should we officially define the new permission on the Meta class of `Image` and `Document`?
> 
> The Image and Document models don't currently define the Add and Edit permissions in their Metas. Should those be defined there, too? That's probably the "right" way to do it, but I'm not sure how it'd interact with established Wagtail projects. And defining permissions in multiple different ways sounds like a disaster waiting to happen.

Looking at https://docs.djangoproject.com/en/3.1/topics/auth/default/#default-permissions and https://docs.djangoproject.com/en/3.1/topics/auth/customizing/#custom-permissions , Django makes a distinction between 'default permissions' (add, edit, delete) and custom permissions, so based on that, defining the choose permission (but not the others) in Meta would fit with how Django expects things to work. It's probably not strictly necessary given the `get_or_create` in the migration, but it can't hurt, in case there is some unforeseen mechanism that garbage-collects them (see [remove_stale_contenttypes](https://github.com/wagtail/wagtail/issues/4444)...)

> > Should a user need the 'choose' permission explicitly in order to choose from a Collection?
> 
> Having the Add or Edit permissions act as a proxy for the Choose permission just sounds confusing to me, now that a separate Choose permission actually exists. An admin with permission to edit Groups might look at a Group's perms and say "Wait a minute, Suzy told me she can choose images from this Collection, but her group doesn't have the Choose perm!"
> 
> Sure, having Edit and _not_ Choose would be weird a weird state, but this PR already gives everyone the Choose permission on the Root Collection (to retain existing behavior), so why would anyone remove Choose but leave Add or Edit?

I agree - to quote the Zen of Python, "explicit is better than implicit". Even if there's no obvious reason why an administrator might give a user add/edit but not choose permission, the best thing Wagtail can do in that situation is to behave as instructed rather than try to second-guess what they meant.

I'm happy with the overall approach here, and with the Meta permission lines added and the small tweaks below addressed, I'm happy to approve this. Would be good to find a fix for the checkbox ordering bug - I agree that it feels more natural for 'choose' to be the first option - but that shouldn't be a blocker to getting this merged.

gasman:
Merged in d18ffb0233dfb01c5fc876fa7e9b755aa9a256f0, with the proposed fixes applied and some additional documentation. Thanks @coredumperror!

coredumperror:
Woooo!! 🥳

PR #6469, opened 2020-10-19
Update Webpack and enable TypeScript

kaedroho:
TODO:
 - [ ] Rewrite plugins config


kaedroho:
Seems like Jest doesn't use Webpack so will have to upgrade babel and use that instead

Issue #6192, opened 2020-06-30
Consistently support form_classname on block definitions


gasman:
The `classname` kwarg on FieldBlock is mentioned in documentation examples (`('heading', blocks.CharBlock(classname="full title"))`) but not actually described. Coupled with the fact that `StructBlock`, `ListBlock` and `StreamBlock` implement the same thing as `form_classname` (#5100), this may mislead people into thinking it applies to the front-end rendering (see #6184).

I propose that we standardise on a `form_classname` kwarg for all block types, and list this under the "Basic block types" options in the docs.

To be determined:

* should we formally deprecate `classname` (and thus drop it in two releases' time), or continue to accept it as an alias of `form_classname` to minimise disruption?
* as a public service to everyone who hates streamfield migrations: can we tweak the deconstruction logic so that changing `classname` to `form_classname` doesn't trigger a migration?

loicteixeira:
> * as a public service to everyone who hates streamfield migrations: can we tweak the deconstruction logic so that changing `classname` to `form_classname` doesn't trigger a migration?

It should be possible, similarly to #4653, we will need to continue supporting the old kwarg name so old migrations don't need to be updated.

meghanabhange:
Heya, 

Wanted to pick this issue. 
If we've reached a consensus on whether to deprecate the variable name or to keep it, could I pick this up?

gasman:
@meghanabhange Sure, go ahead!
I think we should go the simpler route and keep `classname` working as a fallback to `form_classname` - we can always deprecate it later on if we want to free up that name for something else.

PR #6470, opened 2020-10-19
Adding support for `form_classname` meta in FieldBlock

meghanabhange:
Changed the meta input for FormField to take in both `classname` and `form_classname` with reference to #6192

I didn't see any other blocks that use `self.meta.classname` so I believe that FormField is the only block that requires the changes. If I missed anything please do point me to the right place.

### Checklist ✨

- [x] Do the tests still pass?
- [x] Does the code comply with the style guide? 

### DIFF
--- a/docs/topics/streamfield.rst
+++ b/docs/topics/streamfield.rst
@@ -29,7 +29,7 @@ Using StreamField
         author = models.CharField(max_length=255)
         date = models.DateField("Post date")
         body = StreamField([
-            ('heading', blocks.CharBlock(classname="full title")),
+            ('heading', blocks.CharBlock(form_classname="full title")),
             ('paragraph', blocks.RichTextBlock()),
             ('image', ImageChooserBlock()),
         ])
@@ -411,7 +411,7 @@ This defines ``PersonBlock()`` as a block type that can be re-used as many times
 .. code-block:: python
 
     body = StreamField([
-        ('heading', blocks.CharBlock(classname="full title")),
+        ('heading', blocks.CharBlock(form_classname="full title")),
         ('paragraph', blocks.RichTextBlock()),
         ('image', ImageChooserBlock()),
         ('person', PersonBlock()),
--- a/wagtail/core/blocks/base.py
+++ b/wagtail/core/blocks/base.py
@@ -97,6 +97,13 @@ class Block(metaclass=BaseBlock):
         return mark_safe('\n'.join(declarations))
 
     def __init__(self, **kwargs):
+        if 'classname' in self._constructor_args[1]:
+            # Adding this so that migrations are not triggered
+            # when form_classname is used instead of classname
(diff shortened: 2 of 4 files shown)


gasman:
Thanks @meghanabhange!
As mentioned on #6192, it would be good to arrange the migration logic so that people can switch from `classname` to `form_classname` without it triggering a new migration - I think that can be done by modifying `_constructor_kwargs` within the `__init__` method so that it always appears as `form_classname` there even when the block was constructed with `classname`. The details are a bit fiddly, though, so I'm happy to pick that task up...

meghanabhange:
So from what I understand, would we want something like 

```
def __init__(self, **kwargs):
    self.form_classname = kwargs.pop('form_classname', kwargs.get('classname'))
```

and use `self.form_classname` [here](https://github.com/wagtail/wagtail/blob/8d8a609fbc31620ac843762050f2ef783e3535ea/wagtail/core/blocks/field_block.py#L42)?

gasman:
@meghanabhange I think it'll need to be something like:

    def __init__(self, **kwargs):
        if 'classname' in self._constructor_args[1]:
            classname = self._constructor_args[1].pop('classname')
            self._constructor_args[1].setdefault('form_classname', classname)

`_constructor_args` is a copy of the original args/kwargs that were passed when constructing the block, [saved in `Block.__new__`](https://github.com/wagtail/wagtail/blob/cbf65ba645b0d399595c3a0278e4fa22b47e17c7/wagtail/core/blocks/base.py#L60-L65) and [used again in `Block.deconstruct`](https://github.com/wagtail/wagtail/blob/cbf65ba645b0d399595c3a0278e4fa22b47e17c7/wagtail/core/blocks/base.py#L382-L386) to generate the definition to be used in migrations. Tweaking the definition to be the same will hopefully mean that Django's migration logic will see no change to the field when changing `classname` to `form_classname`, so it should be possible to do something like:

(on an old version of Wagtail)
* create a model with a StreamField containing `CharBlock(classname='title')`
* run `manage.py makemigrations` to get an initial migration

(on the new version of Wagtail)
* change `CharBlock(classname='title')` to `CharBlock(form_classname='title')`
* run `manage.py makemigrations` and get a "no changes detected" result

meghanabhange:
@gasman Made the changes in the base class `Block` for the above and tested it on BakeryDemo to see if the migrations were triggered. It shows `no changes detected` now. 

Regarding the documentation, I changed the mentions of `classname` to `form_classname` in e7cda2b. I am slightly confused about where to explicitly mention the change.

gasman:
Thanks again! Now merged in 3a2f2a6ea3c78de1ca0b0847f42054f78e0f653b - the old `classname` argument was not really properly documented (other than the code example using it in passing), so I've now added proper documentation for the new argument and mentioned the change in the release notes.

PR #6471, opened 2020-10-19
Update to Babel 7

kaedroho:

PR #6472, opened 2020-10-19
Enable TypeScript

kaedroho:
Depends on: https://github.com/wagtail/wagtail/pull/6471

This PR enables TypeScript through babel. I've disabled `noImplicitAny` for now to improve interopability with existing JS.


client/src/components/Draftail/decorators/TooltipEntity.js
```diff
@@ -80,8 +80,6 @@ class TooltipEntity extends Component {
     } = this.props;
     const { showTooltipAt } = this.state;
 
-    // Contrary to what JSX A11Y says, this should be a button but it shouldn't be focusable.
```
kaedroho, line 83: Removed due to Eslint update. It returned this error:

```
/home/karl/projects/bakerydemo-i18n/wagtail/client/src/components/Draftail/decorators/TooltipEntity.js
  84:5  error  Definition for rule 'springload/jsx-a11y/interactive-supports-focus' was not found  springload/jsx-a11y/interactive-supports-focus
```

thibaudcolas (approved):
Looking great @kaedroho. I think it would be good to switch fully to TypeScript for the compilation & tests, but as we discussed that’s a bit more involved so I’ll open a separate PR for that.

PR #6474, opened 2020-10-19
Convert the React Page Explorer to TypeScript

kaedroho:
Based on: https://github.com/wagtail/wagtail/pull/6472


client/src/utils/actions.ts
```diff
@@ -1,14 +1,26 @@
-const identity = func => func;
+// Returns the value of the first argument. All others are ignored.
+function identity<T extends any[]>(...func: T): T[0] {
```
kaedroho, line 2: Note that this does exactly the same thing as the last implementation. Just written out like this to allow it to accept a splat and pass through the type of the first argument

client/src/components/Explorer/Explorer.tsx
```diff
@@ -1,12 +1,24 @@
-import PropTypes from 'prop-types';
+/* eslint-disable react/prop-types */
+
```
lb-, line 2: Will prop warnings still trigger? Just checking we have a way for these components to be used outside of typescript land (ie. the ones added to global) and still have the benefit of prop warnings.

client/src/components/Explorer/ExplorerToggle.tsx
```diff
@@ -7,10 +8,15 @@ import * as actions from './actions';
 import Button from '../../components/Button/Button';
 import Icon from '../../components/Icon/Icon';
 
+interface ExplorerToggleProps {
+  onToggle(): void;
+  children: any[];
```
lb-, line 13: Should we add more specifics to this type?
maybe `children: React.ReactNode`
While we may pass children in as an array here, there is no reason it should not compile if in the future we pass in a single child.

Note: from what I can see this `children` is a react element rendered between the icon menu items and not the children (as in page children data).

lb- (changes requested):
A few minor thoughts on some of the code, we may want to check our linting/formatting rules for the ts files as we write more ts (seemed to be a few inconsistencies).

Naming is great, thanks heaps.

I will do a quick smoke test for basic functionality but so far the code seems ok.

Package lock will need updating but I assume you are waiting on this to be approved?

lb-:
@kaedroho  looks good, see comment above, let me know what you think about the comments, my main concern is that we might want to check that the runtime validation of prop types still works (I am not sure if that is something Typescript does out of the box sorry) as these components may be used outside of the compile time checks of typescript.

lb-:
Had a quick play around with the updated Typescript based elements, could not see anything breaking.

client/src/components/Explorer/Explorer.tsx
kaedroho: You won't get runtime warnings anymore for components that have been switched to TypeScript and have had prop types removed.

It looks like it's possible to derive propt ypes from TypeScript interfaces using a plugin: https://github.com/milesj/babel-plugin-typescript-to-proptypes but it might be simpler to just switch everything to TypeScript?

client/src/components/Explorer/ExplorerToggle.tsx
kaedroho: Yep, I'll fix this up!

kaedroho:
Thanks @lb- 
I've just merged some work Thibaud did last night on the TypeScript configuration ino master. There are a few errors with this PR that are raised with the new config:
1. Strict null checks wasn't actually being run, so there's a few "this could be null" errors
2. The new config can derive typescript types from proptypes, so the call to ``PublicationStatus`` (which is still implemented in JS) is now giving an error because I fogot to add the ``meta.status.status`` field to the ``WagtailPageAPI`` interface.

I'll rebase this PR on master now, and the rebase will include fixes for the above and https://github.com/wagtail/wagtail/pull/6474#discussion_r509188085

lb-:
Awesome. What about the package lock update? Will that be on this PR also?

Let me know when this is good for another review, it's late here so I might not get to look at it again until tomorrow my time.

kaedroho:
Just finished the rebasing and committed the new ``package-lock.json``. Thanks!

I'll write an issue so we can discuss linting of TypeScript. Would we need to update ``eslint-config-wagtail`` to support this as well?

client/src/components/Explorer/Explorer.tsx
lb-: ok, we might need to work out what we want to do for components that are shared at some point, all good on these ones though.

lb- (approved):

lb-:
@kaedroho  no release notes added, that can be done separately, rebased & merged via github.

kaedroho:
Thanks!

PR #6477, opened 2020-10-20
Customizable Group create/edit page

seifert:
Hello, we need to customize Group create/edit page in Wagtail admin. However, Waigtail does not support it. I thing, the easiest way is allow customize GroupViewSet by WAGTAIL_GROUP_VIEW_SET setting. My patch solves this. If the changes are ok, I will complete documentation too.


gasman:
Thanks @seifert!
For a long time we've been trying to move away from using settings variables for customisations (because they're a big disorganised list that just keeps growing whenever someone has a new requirement for customisation, with no planning or consistency), and make the process of customising an app more like subclassing it - viewsets were a step towards this goal. What you're proposing here is a new kind of customisation, so I'm thinking this is a good opportunity to take that a step further and use Django's appconf mechanism to manage it.

What do you think about following [this example from the Django docs](https://docs.djangoproject.com/en/3.1/ref/applications/#for-application-users), and making it possible to override the viewset like this:

    # myproject/apps.py
    from wagtail.users.apps import WagtailUsersAppConfig
    from myproject.users import CustomGroupViewSet

    class CustomUsersAppConfig(WagtailUsersAppConfig):
        group_viewset = CustomGroupViewSet

    # myproject/settings.py
    INSTALLED_APPS = [
        'myproject.apps.CustomUsersAppConfig',
        # ...
    ]

We should also standardise on the spelling of 'viewset'. Since we already have `register_admin_viewset` and `wagtail.admin.viewsets`, I suggest that we continue using ViewSet when it's in camel case (e.g. for class names), but write it as one word in all other situations (`group_viewset`, `WAGTAIL_GROUP_VIEWSET`). I think this matches how Django uses 'formset'.

seifert:
Hi @gasman, thank for review. I'm sorry, I was snowed under with other work. I have integrated your suggests and have written documentation. There is one difference:
```
class CustomUsersAppConfig(WagtailUsersAppConfig):
    group_viewset = "path.to.GroupViewSetClass"
```
`group_viewset` is a string, not a class, because the class caused `AppRegistryNotReady` exception (because of model in ModelForm). So lazy import solves it. Current pull request is rebased to master branch.

PR #6479, opened 2020-10-21
Switch to using TypeScript without Babel

thibaudcolas:
Follow-up to #6105. Includes one final commit which finishes the move from Babel to TypeScript, for tests (via ts-jest) and build (via ts-loader). There will be more TypeScript configuration tweaks to do, but with this we’ll already have type errors in CI rather than just in IDEs.

* [x] For front-end changes: Did you test on all of Wagtail’s supported browsers? Chrome 


kaedroho (approved):
Works great! Thanks!

thibaudcolas:
Wooo!

PR #6485, opened 2020-10-21
Make StreamField values mutable and deprecate stream_data

### LINKED ISSUE #2876: Make streamvalue mutable

gasman:
Fixes #2876 and supersedes #2886. StreamValue now implements the `MutableSequence` interface so that it can be modified in place, rather than having to reassign the entire list to update it:

    # Replace the first block with a new block of type 'heading'
    my_page.body[0] = ('heading', "My story")

    # Delete the last block
    del my_page.body[-1]

    # Append a block to the stream
    my_page.body.append(('paragraph', "<p>And they all lived happily ever after.</p>"))

    # Save the updated data back to the database
    my_page.save()

StreamValue has undergone some internal refactoring to make this possible: previously the internal `stream_data` property contained the underlying data in one of two possible formats, depending on whether the data came from the database ('lazy' mode, where it's stored as JSONish dicts and converted to native Python values on first access) or from Python code ('non-lazy' mode, where it's stored in tuples already containing the native Python values). This setup doesn't work for mutable sequences (where we might have some data items originating from the database, and others overwritten in python code) and is pretty messy anyhow.

Unfortunately, `stream_data` has found its way into a lot of end-user code that doesn't properly account for the lazy versus non-lazy cases (which usually means it mostly works, but breaks on page previews) and so this needs to go through the deprecation process. I've added a backwards-compatible shim for `stream_data`, along with a new accessor `raw_data` which consistently returns the JSONish data (i.e. it does what people _think_ `stream_data` did, and for 99% of cases people should be able to fix their code by replacing `stream_data` with `raw_data`).

~~Note: I've written the deprecation warnings and docs as if this is a 2.11 feature, but I'm not expecting it to make the deadline :-) Will update these to target 2.12 once the necessary bits of release housekeeping have been done.~~ (done)

### DIFF
--- a/docs/topics/streamfield.rst
+++ b/docs/topics/streamfield.rst
@@ -1024,6 +1024,30 @@ Your extended value class methods will be available in your template:
     </div>
 
 
+Modifying StreamField data
+--------------------------
+
+A StreamField's value behaves as a list, and blocks can be inserted, overwritten and deleted before saving the instance back to the database. A new item can be written to the list as a tuple of *(block_type, value)* - when read back, it will be returned as a ``BoundBlock`` object.
+
+.. code-block:: python
+
+    # Replace the first block with a new block of type 'heading'
+    my_page.body[0] = ('heading', "My story")
+
+    # Delete the last block
+    del my_page.body[-1]
+
+    # Append a block to the stream
+    my_page.body.append(('paragraph', "<p>And they all lived happily ever after.</p>"))
+
+    # Save the updated data back to the database
+    my_page.save()
+
+
+.. versionadded:: 2.12
+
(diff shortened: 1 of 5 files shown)


gasman:
Now rebased on current master to target 2.12.

kaedroho (approved):

gasman:
Merged in 93a80efe5061e5f0f31b7eb8fb2b36dc83cd3e6c with release notes added in 8d9e864581e1772885ace7efbf9c7253bf989a93

PR #6487, opened 2020-10-22
Fix steps for adding translations to an existing snippet

gasman:
* Fix BootstrapTranslatableMixin to skip TranslatableModel's system check for the missing unique constraint
* Fix incorrect import for BootstrapTranslatableModel and use of `--empty` in final schema migration
* Clarify handling of Meta classes, the fact that you need to run makemigrations for step 1, and what to do for the non-nullable 'locale' prompt


kaedroho (approved):
These changes look good to me, thanks!

PR #6481, opened 2020-10-21
Implement locale selector in React page explorer

kaedroho:
This PR implements a locale switcher in the react page explorer.

image

Based on: https://github.com/wagtail/wagtail/pull/6488

PR #6488, opened 2020-10-22
Add locale indication to the React page explorer

kaedroho:
This PR adds locale indication into the react page explorer. I've created this hoping it would be much quicker to review than the full implementation in https://github.com/wagtail/wagtail/pull/6481. Hopefully we could get both in to 2.11, but if not, we should be OK if we only get this one merged!

At the root level, the locale is displayed next to each page:
image

When you navigate into a section, the locale moves into the header:
image

### DIFF
--- a/client/src/api/admin.ts
+++ b/client/src/api/admin.ts
@@ -15,6 +15,7 @@ export interface WagtailPageAPI {
     parent: {
       id: number;
     } | null;
+    locale?: string;
   };
   /* eslint-disable-next-line camelcase */
   admin_display_title?: string;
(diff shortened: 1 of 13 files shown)


lb-:
Hey @kaedroho the code looks good but I was unable to validate, I tried to follow the instructions in docs/advanced_topics/i18n.rst and could not get the translations to work as desired.

The language showed for the home page only, and did NOT move to the header when I went down a level.

I may not get time to review this again to the weekend sorry, I probably just missed some settings.

I had set the following, restarted, bakery demo, when I went to create a new home page it only gave me the English option.
```
USE_L10N = True
USE_I18N = True
WAGTAIL_I18N_ENABLED = True

WAGTAIL_CONTENT_LANGUAGES = LANGUAGES = [
    ('en', "English"),
    ('it', "Italian"),
]
```

allcaps (approved):
LGTM

It took me some time to realise that I need an external package to setup pages in various locales. But that is outside the scope of this PR. So, this LGTM! :clap:

kaedroho:
@lb-  thanks for looking into it! It may have been that you haven't installed the ``wagtail.locales`` package and set up a locale? 

@allcaps  Thanks!

PR #6489, opened 2020-10-22
Make React locale selector act like a filter on the root page

kaedroho:


allcaps (approved):
LGTM, I did the test-drive and works :clap:

thibaudcolas (changes requested):
I can’t get it to work, and I have no idea why! :( I’m not entirely sure what this is meant to do, @kaedroho I think it would be worth either you adding a description / screenshot to the MR, or talking me through it.

I was expecting the locale selector to render for the root page, but it doesn’t for two reasons:

- There’s a `isSiteRoot` check that prevents rendering (but that check looks like it might be using the wrong id?)
- There is no `meta.locale` defined that I can see.

Changing the code, I think I could get this to work as expected (including reproducing the bug with hyphens in locales), but that involved completely reworking the "render yes/no" check for SelectLocale.

thibaudcolas:
This has been open for a while and I wasn’t clear whether the PR was suitable for inclusion back then. This feels a bit dated as an approach considering we have largely moved away from React for this type of feature. So – will close now, but happy for a clearer feature request for this in the pages explorer menu (and PR after we discuss the approach).

PR #6493, opened 2020-10-23
WIP: Add support for SVG icons to SearchArea subclasses

thibaudcolas:
WIP

---


Scotchester:
@thibaudcolas Ready for review now?

thibaudcolas:
@Scotchester almost – I updated the wrong test case!

thibaudcolas:
@Scotchester should be good to go now I believe. I’ve updated the description with the test cases I used in bakerydemo.

docs/reference/hooks.rst
```diff
@@ -287,7 +287,8 @@ Hooks for building new areas of the admin interface (alongside pages, images, do
   :label: text displayed in the "Other Searches" option box.
   :name: an internal name used to identify the search option; defaults to the slugified form of the label.
   :url: the URL of the target search page.
-  :classnames: additional CSS classnames applied to the link, used to give it an icon.
+  :classnames: arbitrary CSS classnames applied to the link
+  :icon_name: icon to display next to the label.
```
Stormheg, line 291: **Question (non-blocking):** do we have an overview of acceptable icon names? Maybe we could link to that?

Stormheg (approved):
Thanks @thibaudcolas! Looks good to me :+1: 

Left a comment for a potential documentation improvement.

I've tested this change on the following browsers
- Firefox 86
- Chrome 89
- Safari on iPadOS 14.4 (iPad 6th gen)

I don't have access to BrowserStack so unfortunately I'm not able to test all browsers listed in [Wagtail's browser and device support](https://docs.wagtail.io/en/stable/contributing/developing.html#browser-and-device-support).

docs/reference/hooks.rst
Scotchester: The only place I'm aware of that lists them is the Styleguide app that one can turn on in their Wagtail installation. Would be nice to have them on a page in the docs, too.

thibaudcolas:
Thank you @Stormheg :) 

I decided to _not_ add a note about how to find a list of existing icons in the end – I think I’d rather see us do that more consistently as part of #6028.

Issue #6499, opened 2020-10-27
`before_delete_page` hook documentation clarification needed


KalobTaulien:
On https://docs.wagtail.io/en/stable/reference/hooks.html#before-delete-page it says:

> Called at the beginning of the “delete page” view passing in the request and the page object.
> Uses the same behaviour as before_create_page.

That's referring to [this area](https://docs.wagtail.io/en/stable/reference/hooks.html#before-create-page) and the following code:
```python
from wagtail.core import hooks

from .models import AwesomePage
from .admin_views import edit_awesome_page

@hooks.register('before_create_page')
def before_create_page(request, parent_page, page_class):
    # Use a custom create view for the AwesomePage model
    if page_class == AwesomePage:
        return create_awesome_page(request, parent_page)
```

However, according to https://github.com/wagtail/wagtail/blob/master/wagtail/admin/views/pages/delete.py#L20 the `before_delete_page ` hook only takes `request` and `page` as parameters.

I suggest the documentation around the `before_delete_page` hook be updated, ideally with a code example (even if it's somewhat duplicating another similar example). 

This is a good first issue for anybody to pick up.

Tanmay-47:
Hey, If this issue is still open, I'm willing to take this up.

lb-:
fixed via 92381698d0215d9f26c71a52c61ddeb429fb96e7  & https://github.com/wagtail/wagtail/pull/6500

PR #6500, opened 2020-10-29
Add code example to documentation around `before_delete_page` hook.

janeliutw:
issue #6499. Let me know if anything needs to be updated!

### DIFF
--- a/docs/reference/hooks.rst
+++ b/docs/reference/hooks.rst
@@ -533,6 +533,16 @@ Hooks for customising the way users are directed through the process of creating
 
   Uses the same behaviour as ``before_create_page``.
 
+ .. code-block:: python
+
+    from wagtail.core import hooks
+
+    from .models import AwesomePage
+    from .admin_views import edit_awesome_page
+
+    @hooks.register('before_delete_page')
+    def before_delete_page(request, page):
+        return delete_awesome_page(request, page)
 
 .. _after_edit_page:
 


docs/reference/hooks.rst
```diff
@@ -533,6 +533,16 @@ Hooks for customising the way users are directed through the process of creating
...
+ .. code-block:: python
+
+    from wagtail.core import hooks
+
+    from .models import AwesomePage
+    from .admin_views import edit_awesome_page
+
+    @hooks.register('before_delete_page')
+    def before_delete_page(request, page):
+        return delete_awesome_page(request, page)
```
tomdyson, line 545: I know this is also in the `before_create_page` example, but I'm not sure why we import `edit_awesome_page` then call `delete_awesome_page`.

tomdyson (commented):
Thanks @janeliutw!

docs/reference/hooks.rst
janeliutw: Yeah I was actually wondering about that as well, but I am not super familiar with the project so I figured I just kept it. Happy to remove it in both of the examples though!

lb-:
@janeliutw  thanks so much for this PR, I know it has been a long while but we have now merged this in with a bit of refinement via 92381698d0215d9f26c71a52c61ddeb429fb96e7

Really appreciate your time taken on this, thank you for contributing to Wagtail.

PR #6501, opened 2020-10-30
Fix ModelAdmin documentation example

naglis:
The `get_extra_class_names_for_field_col` seems to have an `if` missing and uses an undefined `balance`.


gasman:
Thanks @naglis! Looks like the `Decimal('-50.00')` is wrong too, as that would never be reached, so I've fixed that too. Merged in 74c692ba213279300db42797b2aeced5626b3071

Issue #6467, opened 2020-10-16
Embed's thumbnail_url model field is too short


GTpyro:
### Issue Summary

The thumbnail_url model field in embeds/models.py is too short. Currently, the character limis is set to 255 however some providers return a longer url (e.g. Instagram). This does not affect sqlite backends as sqlite does not enforce field length limits. When this error occurs, it results in the following stack trace:
```
[16/Oct/2020 02:59:41] "GET /admin/embeds/chooser/ HTTP/1.1" 200 1556
Internal Server Error: /admin/embeds/chooser/upload/
Traceback (most recent call last):
  File "/home/someuser/code/wagtail_scratch/bugreport/venv/lib/python3.8/site-packages/django/db/models/query.py", line 573, in get_or_create
    return self.get(**kwargs), False
  File "/home/someuser/code/wagtail_scratch/bugreport/venv/lib/python3.8/site-packages/django/db/models/query.py", line 429, in get
    raise self.model.DoesNotExist(
wagtail.embeds.models.Embed.DoesNotExist: Embed matching query does not exist.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/someuser/code/wagtail_scratch/bugreport/venv/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.StringDataRightTruncation: value too long for type character varying(255)


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/someuser/code/wagtail_scratch/bugreport/venv/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/someuser/code/wagtail_scratch/bugreport/venv/lib/python3.8/site-packages/django/core/handlers/base.py", line 179, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/someuser/code/wagtail_scratch/bugreport/venv/lib/python3.8/site-packages/django/views/decorators/cache.py", line 44, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "/home/someuser/code/wagtail_scratch/bugreport/venv/lib/python3.8/site-packages/wagtail/admin/urls/__init__.py", line 110, in wrapper
    return view_func(request, *args, **kwargs)
[...]

Issue #1745, opened 2015-09-28
MultipleObjectsReturned: get() returned more than one Embed


lee-kagiso:
We are getting this quite a lot.

We have  a Sentry report over here: https://beta.getsentry.com/share/group/34313638322e3833333433303131/

```
MultipleObjectsReturned: get() returned more than one Embed -- it returned 2!
  File "django/core/handlers/base.py", line 137, in get_response
    response = response.render()
  File "django/template/response.py", line 103, in render
    self.content = self.rendered_content
  File "django/template/response.py", line 80, in rendered_content
    content = template.render(context)
  File "django/template/base.py", line 148, in render
    return self._render(context)
  File "django/template/base.py", line 142, in _render
    return self.nodelist.render(context)
  File "django/template/base.py", line 844, in render
    bit = self.render_node(node, context)
  File "django/template/base.py", line 858, in render_node
    return node.render(context)
  File "django/template/loader_tags.py", line 126, in render
    return compiled_parent._render(context)
  File "django/template/base.py", line 142, in _render
    return self.nodelist.render(context)
  File "django/template/base.py", line 844, in render
    bit = self.render_node(node, context)
  File "django/template/base.py", line 858, in render_node
    return node.render(context)
  File "django/template/loader_tags.py", line 65, in render
    result = block.nodelist.render(context)
  File "django/template/base.py", line 844, in render
    bit = self.render_node(node, context)
  File "django/template/base.py", line 858, in render_node
    return node.render(context)
  File "django/template/base.py", line 904, in render
    return render_value_in_context(output, context)
  File "django/template/base.py", line 881, in render_value_in_context
    value = force_text(value)
  File "django/utils/encoding.py", line 83, in force_text
    s = six.text_type(s)
  File "wagtail/wagtailcore/blocks/stream_block.py", line 311, in __str__
    return self.stream_block.render(self)
  File "wagtail/wagtailcore/blocks/base.py", line 216, in render
    return self.render_basic(value)
  File "wagtail/wagtailcore/blocks/stream_block.py", line 201, in render_basic
    [(force_text(child), child.block_type) for child in value]
  File "wagtail/wagtailcore/blocks/stream_block.py", line 201, in <listcomp>
    [(force_text(child), child.block_type) for child in value]
  File "django/utils/encoding.py", line 83, in force_text
    s = six.text_type(s)
  File "wagtail/wagtailcore/blocks/stream_block.py", line 251, in __str__
    return self.block.render(self.value)
[...]

PR #6502, opened 2020-11-01
Remove max_length form embed url fields

allcaps:
This PR replaces the URL fields on embeds for an URL fields without `max_length`.

Fixes: #6467 and #1985

The custom replacement URL field is a subclass of `TextField` and applies the `URLValidator` and in forms uses the `form.URLField`.


allcaps:
> _mysql_exceptions.OperationalError: (1170, "BLOB/TEXT column 'url' used in key specification without a key length")
https://travis-ci.org/github/wagtail/wagtail/jobs/740595278

ababic:
Hey @allcaps,

Thank you for starting work on this. The above error is to be expected really. As noted on #1985, mysql has some limitations when it comes to the length of index values. 

In short, what we want to do is:

1. Add `hash = models.CharField(max_length=32, null=True)` as a new model field (and create a migration)
3. Add a reusable function to `wagtail.embeds.embeds` to take care of hashing, e.g.:
    ```
    from hashlib import md5

    get_embed_hash(url, max_width):
        h = md5()
        h.update(url.encode('utf-8'))
        h.update(max_width)
        return h.hexdigest() 
    ```
4. Update [`get_embed()`](https://github.com/wagtail/wagtail/blob/master/wagtail/embeds/embeds.py#L8) to create a `hash` from `url` and `max_width`, and use this to looking existing values, and in the `get_or_create()` method nearer to the end of the `get_embed()`.
5.  Write a data migration to populate the `hash` field value on all existing `Embed` objects, using `builk_update()` to do the saving in batches (e.g. https://github.com/wagtail/wagtail/pull/6164/files#diff-2cac7a6a3576444b76d1ae7c28cdba967ddc3938442e84a14c3d7f5d9d643898)
6. Update the hash field to:  `hash = models.CharField(max_length=32, unique=True)`
7. Change the `url` and `thumbnail_url` fields to use `models.TextField` instead of `models.URLField`(*)
8. Remove the  `unique_together` from the model `Meta`.
7. Generate a final migration.

(*) `Embed` instances are always created by Ajax requests, where there is no way of surfacing validation errors - so, having this validation at the model-level isn't useful. We might as well accept any old string value.

allcaps:
@ababic Thank you for your detailed plan. I did read your comment and was aware of your preferred solution.

Before I give your preferred solution a try, I wanted to try to swop out `unique_together` for custom code that would stay as close as possible to how unique together works. This is also the reason why I added the form validation. Not because it is needed in Wagtail or not, but because unique_together does raise validation errors in forms. I know it is redundant in the embed scenario. I'm happy to drop form related parts.

I broke MySQL on my dev env, so I'm abusing Travis. I'll fix my dev to keep the noice down.

wagtail/embeds/models.py
```diff
@@ -58,3 +60,18 @@ def is_responsive(self):
...
+    def validate_url_max_width_unique(self, error):
+        qs = self.__class__.objects.filter(url=self.url, max_width=self.max_width)
+        if self.id:
+            qs = qs.exclude(id=self.id)
+        if qs.exists():
+            raise error(_('Embed with this Url and Max width already exists.'))
+
+    def clean(self):
+        self.validate_url_max_width_unique(ValidationError)
+        super().clean()
```
ababic, line 73: Might it be worth overriding `validate_unique()` instead of clean() here? https://github.com/django/django/blob/master/django/db/models/base.py#L1009

wagtail/embeds/models.py
```diff
@@ -58,3 +60,18 @@ def is_responsive(self):
 
     def __str__(self):
         return self.url
+
+    def validate_url_max_width_unique(self, error):
+        qs = self.__class__.objects.filter(url=self.url, max_width=self.max_width)
+        if self.id:
+            qs = qs.exclude(id=self.id)
+        if qs.exists():
+            raise error(_('Embed with this Url and Max width already exists.'))
```
ababic, line 69: Would suggest "Embed with this url and max_width value already exists."

wagtail/embeds/models.py
```diff
@@ -22,20 +25,19 @@ class Embed(models.Model):
     If an instance of this model is deleted, it will be automatically refetched
     next time the embed code is needed.
     """
-    url = models.URLField()
+    url = URLField()
```
ababic, line 28: As mentioned previously, we might as well use `TextField` here and drop the custom field/related tests

wagtail/embeds/tests/test_embeds.py
```diff
@@ -572,3 +574,40 @@ def test_clean_invalid_url(self, get_embed):
...
+    def test_unique_constraint_form_raised_validation_error(self):
+        url = "https://www.youtube.com/watch?v=-GPbywRWkIU"
+        max_width = 600
+        form = EmbedForm(data=dict(url=url, max_width=max_width))
+        self.assertTrue(form.is_valid())
+        form.save()
+
+        form = EmbedForm(data=dict(url=url, max_width=max_width))
+        self.assertFalse(form.is_valid())
+        self.assertEqual(form.errors, {'__all__': ['Embed with this Url and Max width already exists.']})
```
ababic, line 606: Could this test be simplified to just call `full_clean()` on an Embed instance?

wagtail/embeds/tests/test_embeds.py
```diff
@@ -572,3 +574,40 @@ def test_clean_invalid_url(self, get_embed):
...
+class EmbedForm(forms.ModelForm):
+    class Meta:
+        model = Embed
+        fields = ["url", "max_width"]
+
+
+class TestEmbedModel(TestCase):
+    def test_url_fields(self):
+        self.assertIsInstance(Embed._meta.get_field('url'), URLField)
+        self.assertIsInstance(Embed._meta.get_field("thumbnail_url"), URLField)
```
ababic, line 588: This test can be safely removed, I think. It would be very odd indeed for Django to not respect our model field definitions.

ababic (commented):
Hi @allcaps. Noticed a couple of bits here. In general - the one thing we lose with this approach is a DB index, which will have an impact on lookup speed at the database level. Maybe that's entirely acceptable though, given often embeds are really looked up from the DB.

gasman (changes requested):
A problem with emulating unique_together with a validator function is that [the `get_or_create` call in `get_embed`](https://github.com/wagtail/wagtail/blob/330f0dde152ea23ae5072e03508a5d3c15b03bcb/wagtail/embeds/embeds.py#L42) relies on the unique_together constraint - internally Django's get_or_create implementation is more like "try to create but catch the integrity error". Since the embed fetch could take a second or more, there's a real possibility of a race condition where two calls to `get_embed` are made in close succession, they both fail the initial lookup, and end up creating a duplicate entry.

I'd also echo @ababic's comment re losing the database index. Based on https://stackoverflow.com/questions/7830091/django-does-unique-together-imply-db-index-true-in-the-same-way-that-foreignkey it's slightly unclear to me whether unique_together automatically gives you an index, or whether that's only the case for Postgres. If we _don't_ currently have an index in place, we should probably fix that...

I'd favour the hash-based method. I think that would also fix https://github.com/wagtail/wagtail/issues/1745#issuecomment-151131584, which has been open for far too long...

allcaps:
Okay, that settles it. Stubborn me is convinced. I'll implement the hash with an index plan.

ababic:
@allcaps stubbornness is underrated :) I'm glad we explored the options here.

allcaps:
Superseded by https://github.com/wagtail/wagtail/pull/6635

PR #6506, opened 2020-11-02
More prominent note about wagtail-localize in i18n docs

### LINKED ISSUE #6504: Internationalisation documentation issue 2.11rc1
Maybe docs need to improve slightly.
I have setup wagtail. According docs. International seems to work. I see in admin hen creating a page choice for language.  I also see in API the locale tag. So I think I set everything correct up.

However Its unclear how to translate a page. I don't see any button or any way to translate.

I do the following:
1. create homepage (english)
2. link that homepage in wagtail setting as root page
3. ???

- tried to copy the page
- tried to make extra homepage in dutch
- tried to create 2 wagtail configurations
- tried to reconfigure, restart project etc ... 

Nothing works to setup the translation. Nor the ?translation_of=... gives me translations

I'm missing some crucial information how to start and how the wagtail tree is copied over. I more ore less expect a button so I can jump from the english version to the dutch version.

Or some setting in base.py where you can say copy_tree=True ... 

My request how to make docs a bit more clear? I must doing something wrong?

The docs say

_Translated homepages
_When you set up a site in Wagtail, you select the site’s homepage in the ‘root page’ field and all requests to that site’s root URL will be routed to that page._

_Multi-lingual sites have a separate homepage for each locale that exist as siblings in the page tree. Wagtail finds the other homepages by looking for translations of the site’s ‘root page’._

_This means that to make a site available in another locale, you just need to translate and publish its homepage in that new locale._

_If Wagtail can’t find a homepage that matches the user’s language, it will fall back to the page that is selected as the ‘root page’ on the site record, so you can use this field to specify the default language of your site._

gasman:
Fixes #6504 - clarify the difference in roles between Wagtail's i18n support and wagtail-localize, and emphasise that you need wagtail-localize if you're translating content between languages. A few people have missed this detail and been sent on the wrong track - I suspect that describing it as a "translation plugin" is leading people to imagine something like a Google Translate widget, rather than the whole admin UI for translation.

I've also removed the "such as..." wording, since wagtail-localize is currently the only app that integrates with the i18n model described here - we can reinstate it if and when alternative apps appear.

### DIFF
--- a/docs/advanced_topics/i18n.rst
+++ b/docs/advanced_topics/i18n.rst
@@ -18,9 +18,13 @@ Overview
 
 Out of the box, Wagtail assumes all content will be authored in a single language.
 This document describes how to configure Wagtail for authoring content in
-multiple languages. For translating content between different languages, a translation
-plugin (such as `wagtail-localize <https://github.com/wagtail/wagtail-localize>`_)
-must be installed separately.
+multiple languages.
+
+.. note::
+    Wagtail provides the infrastructure for creating and serving content in multiple languages,
+    but does not itself provide an admin interface for managing translations of the same content
+    across different languages. For this, the `wagtail-localize <https://github.com/wagtail/wagtail-localize>`_
+    app must be installed separately.
 
 This document only covers the internationalisation of content managed by Wagtail.
 For information on how to translate static content in template files, JavaScript


tomdyson (approved):
LGTM

PR #6515, opened 2020-11-03
Additional safeguard to check if a cached site is from an old Wagtail version

### LINKED ISSUE #6513: Upgrading to 2.11 breaks websites relying on caching
### Issue Summary

When upgrading to Wagtail 2.11 on websites using caching, specifically Redis in the case we tested, breaks some, or worst case, all rendering of the website.

### Steps to Reproduce

1. Run a previous version of wagtail, e.g. 2.10.2
2. Configure Redis caching
3. Make sure the :1:_wagtail_site_root_paths key has been generated in redis
4. Upgrade to 2.11
5. Visit the admin. You should see the error 

```
Internal Server Error: /admin/
  File "/Users/som/Virtualenvs/wagtailredistest/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
  File "/Users/som/Virtualenvs/wagtailredistest/lib/python3.8/site-packages/django/core/handlers/base.py", line 179, in _get_response
  File "/Users/som/Virtualenvs/wagtailredistest/lib/python3.8/site-packages/django/views/decorators/cache.py", line 44, in _wrapped_view_func
  File "/Users/som/Virtualenvs/wagtailredistest/lib/python3.8/site-packages/wagtail/admin/urls/__init__.py", line 127, in wrapper
  File "/Users/som/Virtualenvs/wagtailredistest/lib/python3.8/site-packages/wagtail/admin/auth.py", line 191, in decorated_view
  File "/Users/som/Virtualenvs/wagtailredistest/lib/python3.8/site-packages/wagtail/admin/views/home.py", line 187, in home
    site_details = get_site_for_user(request.user)
  File "/Users/som/Virtualenvs/wagtailredistest/lib/python3.8/site-packages/wagtail/admin/navigation.py", line 36, in get_site_for_user
    root_site = root_page.get_site()
  File "/Users/som/Virtualenvs/wagtailredistest/lib/python3.8/site-packages/wagtail/core/models.py", line 1678, in get_site
    url_parts = self.get_url_parts()
  File "/Users/som/Virtualenvs/wagtailredistest/lib/python3.8/site-packages/wagtail/core/models.py", line 1565, in get_url_parts
    possible_sites = [
  File "/Users/som/Virtualenvs/wagtailredistest/lib/python3.8/site-packages/wagtail/core/models.py", line 1567, in <listcomp>
    for pk, path, url, language_code in self._get_site_root_paths(request)
ValueError: not enough values to unpack (expected 4, got 3)
```

In the [release notes](https://docs.wagtail.io/en/v2.11/releases/2.11.html#site-get-site-root-paths-now-returns-language-code), we have what I suspect is the reason for this error. 
The addition of this fourth argument, the language, to _get_site_root_paths causes [this line](https://github.com/wagtail/wagtail/blob/12e8ee97bbe2957aa27a9c783680f38307e72056/wagtail/core/models.py#L1567) to break. 

I would expect Wagtail internals to handle this kind of upgrade internally, the object stored in the cache is created by Wagtail internally, and accessed by Wagtail internally. 

A simple fix would be to count the parameters returned from cache, and simply invalidate it if they do not match the expected value.

### Technical details

* Python version: 3.8.5
* Django version: 3.1.3
* Wagtail version: 2.11

saevarom:
Fixes #6513. 
This proposed PR fixes the immediate problem that I had with upgrading to Wagtail 2.11.

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -299,19 +299,11 @@ class Site(models.Model):
         """
         result = cache.get('wagtail_site_root_paths')
 
-        # Wagtail 2.11 changed the way site root paths were stored which would cause
-        # an upgraded 2.11 site to break when loading cached site root paths that were cached with 2.10.2
-        # or older versions of Wagtail.
-        # The code below checks if the any of the cached site urls is consistent with an older version of
-        # Wagtail and busts the cache.
-        if result is not None:
-            for site_root_paths in result:
-                if len(site_root_paths) == 3:
-                    cache.delete('wagtail_site_root_paths')
-                    result = None
-                    break
-
-        if result is None:
+        # Wagtail 2.11 changed the way site root paths were stored. This can cause an upgraded 2.11
+        # site to break when loading cached site root paths that were cached with 2.10.2 or older
+        # versions of Wagtail. The line below checks if the any of the cached site urls is consistent
+        # with an older version of Wagtail and invalidates the cache.
+        if result is None or any(len(site_record) == 3 for site_record in result):
             result = []
 
             for site in Site.objects.select_related('root_page', 'root_page__locale').order_by('-root_page__url_path', '-is_default_site', 'hostname'):
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -301,6 +301,7 @@ class Site(models.Model):
 
         # Wagtail 2.11 changed the way site root paths were stored which would cause 
         # an upgraded 2.11 site to break when loading cached site root paths that were cached with 2.10.2
+        # or older versions of Wagtail.
         # The code below checks if the any of the cached site urls is consistent with an older version of 
         # Wagtail and busts the cache.
         if result is not None:
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -299,6 +299,17 @@ class Site(models.Model):
         """
         result = cache.get('wagtail_site_root_paths')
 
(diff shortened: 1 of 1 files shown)


wagtail/core/models.py
```diff
@@ -299,6 +299,18 @@ def get_site_root_paths():
...
+        # or older versions of Wagtail.
+        # The code below checks if the any of the cached site urls is consistent with an older version of
+        # Wagtail and busts the cache.
+        if result is not None:
+            for site_root_paths in result:
+                if len(site_root_paths) == 3:
+                    cache.delete('wagtail_site_root_paths')
+                    result = None
+                    break
+
```
gasman, line 313: Thanks @saevarom!

Just wondering - given that we're immediately setting the cache key again in the code below, deleting it is probably redundant... Could we achieve the same effect just by changing the line below to:

`if result is None or any(len(site_record == 3) for site_record in result):`

saevarom: You're right, this is cleaner. There is a slight parenthesis mismatch in your line so I will change it to:

`if result is None or any(len(site_record) == 3 for site_record in result):`

And commit a new change.

gasman:
Thanks again! Merged in a5415104170b880037b3efcb5148b85d75c551ac (for master) and 3ec0d405edba443dc5609c0ce2b97f6721f8714c (for 2.11.x).

PR #6519, opened 2020-11-04
Switch from Travis CI to Github Actions

gasman:
Replace our Travis CI setup with Github Actions, since Travis CI has scaled back their resources for open-source builds to the point of unusability.

Elasticsearch doesn't provide Docker builds for 2.x, which makes it decidedly non-trivial to test here, but we're well overdue for dropping support for that anyway (so I'll do that in a follow-up PR). For the purposes of the 2.11.x LTS branch (which does still need to support Elasticsearch 2) we'll keep Travis for just the Elasticsearch 2.x test, and so the last commit here should be omitted when merging into stable/2.11.x.


gasman:
Just to add (since it won't run on this PR)... you can see this in action on my fork at https://github.com/gasman/wagtail/actions.

gasman:
Merged via #6520

Issue #6511, opened 2020-11-03
Changing language in admin breaks the page explorer menu


gasman:
### Steps to Reproduce

On a fresh Wagtail install, log in to the admin and go to Account settings -> Language preferences and set language to French. Opening the page explorer then gives a server error, with this error in the console:

```
Internal Server Error: /admin/api/main/pages/
Traceback (most recent call last):
  File "/Users/matthew/.virtualenvs/langwithoutwagtail/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/Users/matthew/.virtualenvs/langwithoutwagtail/lib/python3.9/site-packages/django/core/handlers/base.py", line 179, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/Users/matthew/.virtualenvs/langwithoutwagtail/lib/python3.9/site-packages/django/views/decorators/cache.py", line 44, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "/Users/matthew/.virtualenvs/langwithoutwagtail/lib/python3.9/site-packages/wagtail/admin/urls/__init__.py", line 127, in wrapper
    return view_func(request, *args, **kwargs)
  File "/Users/matthew/.virtualenvs/langwithoutwagtail/lib/python3.9/site-packages/wagtail/admin/auth.py", line 170, in decorated_view
    response = view_func(request, *args, **kwargs)
  File "/Users/matthew/.virtualenvs/langwithoutwagtail/lib/python3.9/site-packages/wagtail/api/v2/router.py", line 63, in wrapped
    return func(request, *args, **kwargs)
  File "/Users/matthew/.virtualenvs/langwithoutwagtail/lib/python3.9/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File "/Users/matthew/.virtualenvs/langwithoutwagtail/lib/python3.9/site-packages/rest_framework/viewsets.py", line 125, in view
    return self.dispatch(request, *args, **kwargs)
  File "/Users/matthew/.virtualenvs/langwithoutwagtail/lib/python3.9/site-packages/rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "/Users/matthew/.virtualenvs/langwithoutwagtail/lib/python3.9/site-packages/wagtail/api/v2/views.py", line 116, in handle_exception
[...]

PR #6521, opened 2020-11-05
Prevent errors on get_url_parts when a non-content-language locale is active

### LINKED ISSUE #6510: Page.get_url fails on non-default locales when not using WAGTAIL_I18N_ENABLED
### Issue Summary

With WAGTAIL_I18N_ENABLED = False, `Page.get_url` fails with a LookupError when a locale other than LANGUAGE_CODE is active.

### Steps to Reproduce

2. `./manage.py migrate` / `./manage.py createsuperuser`
3. Add a LANGUAGES setting to myproject/settings/base.py:

       LANGUAGES = (
           ("en", "English"),
           ("fr", "French"),
       )

4. Update home/models.py:

       class HomePage(Page):
           def serve(self, request):
               activate('fr')
               link = self.get_url()

               return super().serve(request)

5. Visit http://localhost:8000/ and receive the error:

```
  File "/Users/matthew/.virtualenvs/langwithoutwagtail/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
  File "/Users/matthew/.virtualenvs/langwithoutwagtail/lib/python3.9/site-packages/django/core/handlers/base.py", line 179, in _get_response
  File "/Users/matthew/.virtualenvs/langwithoutwagtail/lib/python3.9/site-packages/wagtail/core/views.py", line 24, in serve
    return page.serve(request, *args, **kwargs)
  File "/Users/matthew/Development/tbx/wagtail/debug/langwithoutwagtail/home/models.py", line 10, in serve
    link = self.get_url()
  File "/Users/matthew/.virtualenvs/langwithoutwagtail/lib/python3.9/site-packages/wagtail/core/models.py", line 1641, in get_url
    url_parts = self.get_url_parts(request=request)
  File "/Users/matthew/.virtualenvs/langwithoutwagtail/lib/python3.9/site-packages/wagtail/core/models.py", line 1587, in get_url_parts
    if get_supported_content_language_variant(translation.get_language()) == language_code:
  File "/Users/matthew/.virtualenvs/langwithoutwagtail/lib/python3.9/site-packages/wagtail/core/utils.py", line 271, in get_supported_content_language_variant
    raise LookupError(lang_code)

Exception Type: LookupError at /
Exception Value: fr
```

This bug occurs because `wagtail.core.utils.get_content_languages()` defaults to a list consisting of just LANGUAGE_CODE when WAGTAIL_CONTENT_LANGUAGES is not set. Since existing sites may have multi-language features not built using Wagtail 2.11's i18n support, it's not valid to assume that the active locale is always going to correspond to LANGUAGE_CODE or WAGTAIL_CONTENT_LANGUAGES.

### Technical details

* Python version: 3.9
* Django version: 3.1.3
* Wagtail version: 2.11
* Browser version: Chrome 86

gasman:
Fixes #6510 and #6511

Catch the LookupError from get_supported_content_language_variant so that if the active language is not one that is recognised as a content language, the URL prefix will be left unchanged from the page's reported locale.

Additionally, skip the `translation.override` when WAGTAIL_I18N_ENABLED is false; this accounts for the fact that existing sites may have used alternative / custom i18n implementations involving i18n_patterns, and on these sites we can't rely on the page's locale field to be meaningful. That way, we restore the pre-2.11 behaviour: if the existing i18n code was using some kind of wrapper around the get_url_parts call, then that will work again; and if not, the URL will reflect the current active locale, which is no worse than before.

### DIFF
--- a/wagtail/admin/tests/api/test_pages.py
+++ b/wagtail/admin/tests/api/test_pages.py
@@ -2,6 +2,7 @@ import collections
 import datetime
 import json
 
+from django.contrib.auth import get_user_model
 from django.urls import reverse
 from django.utils import timezone
 
@@ -9,6 +10,7 @@ from wagtail.api.v2.tests.test_pages import TestPageDetail, TestPageListing
 from wagtail.core.models import Locale, Page
 from wagtail.tests.demosite import models
 from wagtail.tests.testapp.models import SimplePage, StreamPage
+from wagtail.users.models import UserProfile
 
 from .utils import AdminAPITestCase
 
@@ -100,6 +102,20 @@ class TestAdminPageListing(AdminAPITestCase, TestPageListing):
         response = self.get_response()
         content = json.loads(response.content.decode('UTF-8'))
         self.assertEqual(content['meta']['total_count'], new_total_count)
+
+    def test_get_in_non_content_language(self):
+        # set logged-in user's admin UI language to Swedish
+        user = get_user_model().objects.get(email='test@email.com')
+        UserProfile.objects.update_or_create(user=user, defaults={'preferred_language': 'se'})
+
+        response = self.get_response()
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response['Content-type'], 'application/json')
+
+        # Will crash if the JSON is invalid
+        content = json.loads(response.content.decode('UTF-8'))
+        self.assertIn('meta', content)
 
     # FIELDS
 
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -1585,16 +1585,28 @@ class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):
             else:
                 site_id, root_path, root_url, language_code = possible_sites[0]
 
-        # If the active language code is a variant of the page's language, then
-        # use that instead
-        # This is used when LANGUAGES contain more languages than WAGTAIL_CONTENT_LANGUAGES
-        if get_supported_content_language_variant(translation.get_language()) == language_code:
-            language_code = translation.get_language()
+        use_wagtail_i18n = getattr(settings, 'WAGTAIL_I18N_ENABLED', False)
+
+        if use_wagtail_i18n:
+            # If the active language code is a variant of the page's language, then
+            # use that instead
+            # This is used when LANGUAGES contain more languages than WAGTAIL_CONTENT_LANGUAGES
+            try:
+                if get_supported_content_language_variant(translation.get_language()) == language_code:
+                    language_code = translation.get_language()
+            except LookupError:
+                # active language code is not a recognised content language, so leave
+                # page's language code unchanged
+                pass
 
         # The page may not be routable because wagtail_serve is not registered
         # This may be the case if Wagtail is used headless
         try:
-            with translation.override(language_code):
+            if use_wagtail_i18n:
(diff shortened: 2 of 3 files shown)


wagtail/admin/tests/api/test_pages.py
```diff
@@ -101,6 +103,20 @@ def test_private_pages_appear_in_list(self):
...
+        user = get_user_model().objects.get(email='test@email.com')
+        UserProfile.objects.update_or_create(user=user, defaults={'preferred_language': 'se'})
+
+        response = self.get_response()
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response['Content-type'], 'application/json')
+
+        # Will crash if the JSON is invalid
+        content = json.loads(response.content.decode('UTF-8'))
```
loicteixeira, line 117: FWIW, I believe a crash will result in an error rather than a failure which isn't quite the same thing (or is it only pytest making that difference?) but at the end of the day, the test will fail and that's what we want.

wagtail/core/models.py
```diff
@@ -1585,16 +1585,28 @@ def get_url_parts(self, request=None):
...
+            # If the active language code is a variant of the page's language, then
+            # use that instead
+            # This is used when LANGUAGES contain more languages than WAGTAIL_CONTENT_LANGUAGES
+            try:
+                if get_supported_content_language_variant(translation.get_language()) == language_code:
+                    language_code = translation.get_language()
+            except LookupError:
+                # active language code is not a recognised content language, so leave
+                # page's language code unchanged
+                pass
```
loicteixeira, line 1600: I find that `contextlib.suppress` better carry the intend (and unlike `except/pass`, it won't be picked up by static analysis tools as a potential error). For example:

```suggestion
            # If the active language code is a variant of the page's language, use that instead.
            # This is used when LANGUAGES contain more languages than WAGTAIL_CONTENT_LANGUAGES.
            # LookupError will be raised if the active language code is not a recognised content language
            # so leave the page's language code unchanged
            with contextlib.suppress(LookupError):
                if get_supported_content_language_variant(translation.get_language()) == language_code:
                    language_code = translation.get_language()
```

loicteixeira (approved):
I left a couple comments but approving nonetheless as it is already functional.

wagtail/core/models.py
gasman: @loicteixeira Can you give more details about the issue with static analysis tools? I'm not sure about the `contextlib.suppress` version being an improvement (it's an extra import, and I think that having both comments up-front makes the code less immediately understandable).

loicteixeira: Some tools might raise warnings when all there is to an `except` block is `pass` statement. Now typically it is reserved for bare excepts (i.e. `except:`) or too broad excepts (i.e. `except Exception:`) but some tools also check for any typed exceptions. [Bandit only checks for bare/board exception by default](https://docs.openstack.org/bandit/1.4.0/plugins/try_except_pass.html), so it wouldn't have picked up this one.

I don't think an extra import can be an issue really. I still believe `suppress` better express the intent from a code POV, but I can see your point of having two separate comments being more easily understandable.

gasman:
Merged in f378f48e148f12e09a0f429cf84f2b2b91a017f4 (master) and 34572e3129f5d1d5e685bd1ac2f06abaa424d1ad (stable/2.11.x)

PR #6522, opened 2020-11-05
Avoid circular import between wagtail.admin.auth and custom user models

### LINKED ISSUE #6518: Circular Import Error in wagtail.cotrib.forms.models for a FormPage in 2.11
### Issue Summary

When upgrading from a previous version of Wagtail, the form pages are not working anymore. I'm getting a circular import error for code that previously worked. The ONLY change was upgrading to 2.11

### Steps to Reproduce

1. Update virtual env to use Wagtail 2.11
2. Startup runserver
3. Now getting the following error: 

`ImportError: cannot import name 'FORM_FIELD_CHOICES' from partially initialized module 'wagtail.contrib.forms.models' (most likely due to a circular import) (/home/user/sandbox/clients/client/project/web/env/envname/lib/python3.8/site-packages/wagtail/contrib/forms/models.py)`

My expectation is that the code would run as before.

### Technical details

* Python version: 3.8.
* Django version: 3.1.3.
* Wagtail version: 2.11.
* Browser version: Version 86.0.4240.111

gasman:
Fixes #6518

If the models.py containing a custom user model directly or indirectly imports wagtail.admin.auth, this causes a circular dependency because wagtail.admin.auth attempts to import django.contrib.auth.views, which in turn depends on the custom user model. Fix this by moving the django.contrib.auth.views import inside reject_request.

### DIFF
--- a/wagtail/admin/auth.py
+++ b/wagtail/admin/auth.py
@@ -5,7 +5,6 @@ from functools import wraps
 import l18n
 
 from django.contrib.auth import get_user_model
-from django.contrib.auth.views import redirect_to_login as auth_redirect_to_login
 from django.core.exceptions import PermissionDenied
 from django.db.models import Q
 from django.shortcuts import redirect
@@ -145,6 +144,9 @@ def reject_request(request):
     if request.is_ajax():
         raise PermissionDenied
 
+    # import redirect_to_login here to avoid circular imports on model files that import
+    # wagtail.admin.auth, specifically where custom user models are involved
+    from django.contrib.auth.views import redirect_to_login as auth_redirect_to_login
     return auth_redirect_to_login(
         request.get_full_path(), login_url=reverse('wagtailadmin_login'))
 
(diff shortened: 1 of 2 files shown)


wagtail/admin/auth.py
```diff
@@ -145,6 +144,7 @@ def reject_request(request):
     if request.is_ajax():
         raise PermissionDenied
 
+    from django.contrib.auth.views import redirect_to_login as auth_redirect_to_login
```
zerolab, line 147: Could use a comment re: circular imports. Otherwise, great 🕵️‍♂️ work!

gasman: Fair point - comment now added!

zerolab (approved):

gasman:
Merged in 3c984b9871157b3a75d345a1d299aa209a0a3b0a (master) and 268663dea5ba28c8126b1637acd2d500e634cbb9 (stable/2.11.x)

PR #6528, opened 2020-11-09
Add contrib simple translation

allcaps:
This PR adds `wagtail.contrib.simple_translation` to Wagtail:
- The simplest form of translation UI, no translation, just duplicating the content in another locale.
- Moves Pages to another locale
- Moves Snippets to another locale

* For front-end changes: Did you test on all of Wagtail’s supported browsers? **Please list the exact versions you tested**. Chrome Version 89
* For new features: Has the documentation been updated accordingly? YES

# New 'translate' option in the page header
Screenshot 2021-04-12 at 22 04 49

# New 'translate' option in the page list
Screenshot 2021-04-12 at 22 05 05

# One locale to translate into.
Screenshot 2021-04-12 at 22 06 05

# Multiple locales to translate into (adds select all)
Screenshot 2021-04-12 at 22 05 40

# Snippet translate button
Screenshot 2021-04-12 at 22 12 51

# Snippet single locale
Screenshot 2021-04-12 at 22 13 01

# Snippet multiple locales
Screenshot 2021-04-12 at 22 13 37


wagtail/contrib/simple_translation/views.py
```diff
@@ -0,0 +1,153 @@
...
+from django.utils.translation import gettext_lazy
+from django.views.generic import TemplateView
+from django.views.generic.detail import SingleObjectMixin
+from wagtail.admin.views.pages.utils import get_valid_next_url_from_request
+from wagtail.contrib.simple_translation.forms import SubmitTranslationForm
+from wagtail.core.models import Locale, Page, TranslatableMixin
+from wagtail.snippets.views.snippets import get_snippet_model_from_url_params
+
+
+class SubmitTranslationView(SingleObjectMixin, TemplateView):
```
kaedroho, line 18: Perhaps rather than calling it "Submit translation", which I think implies there's some more complex translation management going on, could we rename this to something "Create translated page"/"Copy for translation"?

wagtail/contrib/simple_translation/views.py
```diff
@@ -0,0 +1,153 @@
...
+            args=[
+                self.kwargs["app_label"],
+                self.kwargs["model_name"],
+                self.kwargs["pk"],
+            ],
+        )
+
+    def get_success_message(self, locales):
+        return _(
+            "The {model_name} '{object}' was successfully submitted for translation into {locales}"
```
kaedroho, line 148: I think we should avoid the term "submitted" as I think this implies there's a workflow. Maybe reword this to "Successfuly created translated versions of {model_name} '{object}' in {locales}"?

wagtail/contrib/simple_translation/wagtail_hooks.py
```diff
@@ -0,0 +1,105 @@
...
+from .views import SubmitPageTranslationView, SubmitSnippetTranslationView
+
+
+# from simple_translation.models import TranslationSource
+
+
+# The `wagtail.snippets.action_menu` module is introduced in https://github.com/wagtail/wagtail/pull/6384
+# FIXME: Remove this check when this module is merged into master
+try:
+    from wagtail.snippets.action_menu import ActionMenuItem as SnippetActionMenuItem
```
kaedroho, line 23: We can get rid of this check now

mixxorz:
@kaedroho Anything I can help with here to get this across the line?

allcaps:
I do not get why RTD fails. :/ Suggestions?

zerolab:
@allcaps rebase on latest `main`? #6972 fixed it, and I don't see it on your branch

wagtail/contrib/simple_translation/views.py
```diff
@@ -0,0 +1,145 @@
...
+                                    child_page.copy_for_translation(locale)
+
+                                    if child_page.numchild:
+                                        _walk(child_page)
+
+                            _walk(self.object)
+                    else:
+                        self.object.copy_for_translation(
+                            locale
+                        ).save()  # pragma: no cover  # TODO snippet trans.
```
kaedroho, line 67: What's left to do here?

docs/reference/contrib/simple_translation.rst
```diff
@@ -0,0 +1,31 @@
+.. _simple_translation:
+
+Simple translation
+==================
+
+Simple Translation is a user interface that allows to create copies of translatables (pages and snippets) into another language.
```
kaedroho, line 6: Maybe start this off like "The ``simple_translation`` module provides a user interface that allows users to copy pages and translatable snippets into another language"?

docs/reference/contrib/simple_translation.rst
```diff
@@ -0,0 +1,31 @@
...
+
+- Copies are in the source language (not translated)
+- Copies of pages are in draft mode
```
kaedroho, line 9: "draft status" is the usual term we use here

docs/reference/contrib/simple_translation.rst
```diff
@@ -0,0 +1,31 @@
+.. _simple_translation:
+
+Simple translation
+==================
+
+Simple Translation is a user interface that allows to create copies of translatables (pages and snippets) into another language.
+
+- Copies are in the source language (not translated)
```
kaedroho, line 8: How about "Copies are *created* in the source language"

wagtail/contrib/simple_translation/wagtail_hooks.py
```diff
@@ -0,0 +1,78 @@
...
+        path(
+            "submit/snippet/<slug:app_label>/<slug:model_name>/<str:pk>/",
+            SubmitSnippetTranslationView.as_view(),
+            name="submit_snippet_translation",
+        ),
+    ]
+
+    return [
+        path(
+            "localize/",
```
kaedroho, line 31: Probably should rename this to ``translation``?

wagtail/contrib/simple_translation/models.py
```diff
@@ -0,0 +1,16 @@
+from django.db.models import Model
+
+
+class SimpleTranslation(Model):
```
kaedroho, line 4: Wonder if it's worth having this permission in core so that it could be shared? But leave it up to the translation packages to actually register it so it won't appear in the admin unless you have one installed?

kaedroho (commented):
Generally looks good to me!

wagtail/contrib/simple_translation/models.py
allcaps: I added the permission to https://github.com/wagtail/wagtail/blob/main/wagtail/admin/models.py#L21

I like the idea because you do not have to migrate anymore. The permission will exist before adding simple_translation to installed apps.

Is this what you where looking for?

kaedroho: Yep, thanks!

allcaps:
@kaedroho I tried to move the permission to https://github.com/wagtail/wagtail/blob/main/wagtail/admin/models.py#L21
But it touches a lot of code and it feels weird. Maybe it is better if all apps have their own permissions. Then third-party apps also have more control over the name and label of their perms.

I think this is good to merge. If you have anything, let me know!

kaedroho (approved):

Issue #5913, opened 2020-03-27
Document popup paginator BUG


onno-timmerman:
# Issue Summary
When you have many documents in a "collection". And want to select a document in that collection on page 2 or higher then its unreachable. 
The paginator regresses to the general paginator.

### Steps to Reproduce

1. upload "many" documents
2. put "many" document under a collection
3. See bottom how many pages in collection if you see 1 of 2
4. Go to page 2 you will notice that the paginator suddenly is the root collection and will show page 1 of 10

Any other relevant information. For example, why do you consider this a bug and what did you expect to happen instead?
It's kinda annoying bug. I haven't confirmed the bug yet in 2.8 but my guess is this will be still there.

### Technical details

* Python version: 3.7.
* Django version: 2.2
* Wagtail version: 2.7.1
* Browser version: latest chrome

thibaudcolas:
Thank you for reporting this @onno-timmerman, I’ve merged #6529 which should address this, and the fix will be released in Wagtail 2.12 in February.

PR #6529, opened 2020-11-09
added collection_id param to setPage in document-chooser-modal.js

z1r0-:
The setPage function in document-chooser-modal.js does not send the collection_id parameter, so you cannot use the pagination for collections of documents. image-chooser-modal.js looks good.

Fixes https://github.com/wagtail/wagtail/issues/5913

Tested in:
  - Chrome 86.0.4240.183
  - Safari 14.0.1

Didn't have more browsers to test with, but should even work in IE11.


thibaudcolas (approved):
Thank you @z1r0-, it’s great to have this fixed. I’ve tested in Chrome and IE11, see no reason for this not to work everywhere.

PR #6544, opened 2020-11-13
Fix issues in React explorer when there are many languages

kaedroho:
This PR fixes two bugs with the react explorer, both issues occur on sites using i18n and have many locales.

Requesting translations to be nested with each page causes major performance problems when there are many languages. Mozilla has 61 languages, each with a separate homepage that's a translation of the others, so the response for this query contains details of 61 * 61 pages.

This isn't actually required since we only need to know the translations if the user navigates into a page, and the translations are requested again anyway by the ``getPageTranslations`` function.

The other issue is that it wasn't fetching subsequent pages of translations for the translation picker.

### DIFF
--- a/client/src/api/admin.test.js
+++ b/client/src/api/admin.test.js
@@ -20,28 +20,28 @@ describe('admin API', () => {
   describe('getPageChildren', () => {
     it('works', () => {
       getPageChildren(3);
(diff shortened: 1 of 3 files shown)


gasman (approved):

gasman:
Merged in 414571c (master) and 734199b (stable/2.11.x).

PR #6548, opened 2020-11-13
Add combined index for Postgres search backend. (#6546)

### LINKED ISSUE #6546: postgres_search does not use index
### Issue Summary

Whenever you do a search and don't specify any fields, Wagtail concatenates the index fields together into SQL like `("postgres_search_indexentry"."autocomplete" || "postgres_search_indexentry"."body") AS "_vector_"`.  However, each field is indexed individually.  Since the SQL generates a concatenated string the field's indexes are not available for use in the search query. 

An index should be created that covers the default case of the fields being concatenated.

```sql
CREATE INDEX postgres_search_indexentry_index_test ON postgres_search_indexentry
        USING gin (("postgres_search_indexentry"."autocomplete" || "postgres_search_indexentry"."body"));
```

We're on Wagtail 2.7.4 so we don't have a title column yet, but it looks like this is still an issue.  The index should include all three fields instead of the 2.

### Steps to Reproduce

1. Have a project with lots of pages
2. Search
3.  Wait a really long time

wgiddens:
This adds an index on the concatenated `title` and `body` as it is used in normal searches for the Postgres backend.

This is the SQL generated for this query:
```python
Page.objects.search('query')
```
```SQL
SELECT "wagtailcore_page"."id",
       "wagtailcore_page"."path",
       ...
       "wagtailcore_page"."live_revision_id",
       ("postgres_search_indexentry"."title" || "postgres_search_indexentry"."body") AS "_vector_"
FROM "wagtailcore_page"
         LEFT OUTER JOIN "postgres_search_indexentry"
                         ON ((("postgres_search_indexentry"."content_type_id" IN
                               (1, 8, 9, 10, 12, 13, 14, 18, 19, 23, 24) AND
                               "postgres_search_indexentry"."object_id" =
                               ("wagtailcore_page"."id")::text)))
WHERE ("postgres_search_indexentry"."title" || "postgres_search_indexentry"."body") @@
      to_tsquery('english'::regconfig, '''query''')
ORDER BY ((ts_rank('{0.009900990099009901,0.009900990099009901,0.20792079207920794,1.0}',
                   "postgres_search_indexentry"."title",
                   to_tsquery('english'::regconfig, '''query''')) *
           "postgres_search_indexentry"."title_norm") +
          (ts_rank('{0.009900990099009901,0.009900990099009901,0.20792079207920794,1.0}',
                   "postgres_search_indexentry"."body",
                   to_tsquery('english'::regconfig, '''query''')) * 1.0)) DESC,
         "wagtailcore_page"."id" DESC
LIMIT 21
```
Postgres could not use the existing indices since `title` and `body` are concatenated everywhere it is used in the query.
This query plan shows it can use the new index:
```
QUERY PLAN                                                  
----------------------------------------------------------------------------------------------------
 Limit  (cost=776.08..776.13 rows=21 width=182) (actual time=0.064..0.064 rows=0 loops=1)
   ->  Sort  (cost=776.08..776.29 rows=82 width=182) (actual time=0.064..0.064 rows=0 loops=1)
         Sort Key: (((ts_rank('{0.00990099,0.00990099,0.207921,1}'::real[], postgres_search_indexentry.title, '''queri'''::tsquery) * postgres_search_indexentry.title_norm) + (ts_rank('{0.00990099,0.00990099,0.207921,1}'::real[], po
stgres_search_indexentry.body, '''queri'''::tsquery) * '1'::double precision))) DESC, wagtailcore_page.id DESC
         Sort Method: quicksort  Memory: 25kB
         ->  Hash Join  (cost=238.59..773.87 rows=82 width=182) (actual time=0.039..0.039 rows=0 loops=1)
               Hash Cond: ((wagtailcore_page.id)::text = postgres_search_indexentry.object_id)
               ->  Seq Scan on wagtailcore_page  (cost=0.00..430.45 rows=16445 width=142) (actual time=0.011..0.011 rows=1 loops=1)
               ->  Hash  (cost=237.56..237.56 rows=82 width=47) (actual time=0.013..0.013 rows=0 loops=1)
                     Buckets: 1024  Batches: 1  Memory Usage: 8kB
                     ->  Bitmap Heap Scan on postgres_search_indexentry  (cost=12.64..237.56 rows=82 width=47) (actual time=0.013..0.013 rows=0 loops=1)
                           Recheck Cond: ((title || body) @@ '''queri'''::tsquery)
                           Filter: (content_type_id = ANY ('{1,8,9,10,12,13,14,18,19,23,24}'::integer[]))
                           ->  Bitmap Index Scan on postgres_search_indexentry_title_body_concat_search  (cost=0.00..12.61 rows=82 width=0) (actual time=0.013..0.013 rows=0 loops=1)
                                 Index Cond: ((title || body) @@ '''queri'''::tsquery)
 Planning time: 3.270 ms
 Execution time: 0.178 ms
```

Fixes #6546


minusf:
i can confirm that adding this index is a big performance improvement to the full text search, but wouldn't it be more explicit to add it to the model instead?

i am a bit ouf of my depth here but i am confused about adding
```
        indexes = [GinIndex(fields=['autocomplete']),
                   GinIndex(fields=['title']),
                   GinIndex(fields=['body'])]
```
field based indexes (with rather specific names in the migration: `postgres_se_autocom_ee48c8_gin`, `postgres_se_body_aaaa99_gin`, and `postgres_se_title_b56f33_gin`) when the docs say indexes are ignored when fields are chosen? What is the benefit of having these?

wgiddens:
Normally adding indexes to the model is the best way to do it. However, Django does not support computed indexes, so it has to be done in SQL.  

The existing `GinIndex(fields=['title'])` and `GinIndex(fields=['body'])` indexes would only be used if fields were passed into a `search()` call. This probably doesn't happen often so I would guess those indices probably don't have much value, but I'll let Wagtail weigh in here. Someone may be relying on that index and I wouldn't want to break stuff by removing them.

wgiddens:
While I'm here I'll share this graph of CPU usage from one of our database servers. We added the concatenated index on the 16th.

image

kaedroho (approved):

kaedroho:
> The existing GinIndex(fields=['title']) and GinIndex(fields=['body']) indexes would only be used if fields were passed into a search() call. This probably doesn't happen often so I would guess those indices probably don't have much value, but I'll let Wagtail weigh in here. Someone may be relying on that index and I wouldn't want to break stuff by removing them.

Yep, I think these still may be used by PostgreSQL for when searching those fields individually (unless PostgreSQL tags each term with the field, but I think this is unlikely). GIN indexes are usually a lot smaller than their source content so I think it's fine to add another.

I think the main disadvantage would be index performance, do you have any data on whether indexing has taken longer since this change?

kaedroho:
@wgiddens :heart: that chart! Just interested in knowing how much data is that? And how many search queries do you have per day?

wgiddens:
It has made staying with Postgres for search feasible so we're pretty excited. We don't normally run `update_index` since we're using Postgres, but we haven't noticed any issues related to indexing from the db. The migration to add the index took about 45 seconds for 100k pages. 


This database powers about 10 wagtail installations that range from around 20k - 100k pages each.  It averages around 1800 searches per day not counting searches in the admin.

minusf:
> Normally adding indexes to the model is the best way to do it. However, Django does not support computed indexes, so it has to be done in SQL.

right.  maybe a comment on the model meta could be still nice?

minusf:
> Yep, I think these still may be used by PostgreSQL for when searching those fields individually (unless PostgreSQL tags each term with the source field, but I think this is unlikely). GIN indexes are usually a lot smaller than their source content so I think it's fine to add another.

i could not make postgres use these indexes when specifying fields for `search()` (`title`, `body`).  it's also one the prominent caveats in the docs:

> When Specifying the fields to search, the index is not used, so it will be slow on huge sites.

wgiddens:
@minusf, I added a comment to the model. That makes sense.

wgiddens:
@minusf 
I did some testing and also couldn't find a use for the `title` and `body` indexes.  The fields passed into `search()` are Page fields, not IndexEntry fields so it wouldn't be used by actual searches. It could be when updating the index, but I did not look into that. 
I can take them out if @kaedroho wants to. Just let me know.

minusf:
I guess the generated but unused field indexes and the not generated composite index are signs of postgres' stellar performance combined with no wikipedia sized wagtail installations :}

The nicest solution would be to have these indexes work for sites where search by fields is a thing. But tbh I prefer combined searches with boosts for title, etc and will be using the composite index instead.

For this reason I think if the field indexes could be made to work, they should be optional and the site owner should be able to decide what they want to index and what not as these indexes overlap and GIN is fast but disk happy...

In case they cannot be made to work I would definitely argue for their removal, as it's a waste of disk space and cpu on every insert and update.

Out of curiosity, how big were these field indexes on your sizeable instances?  Just for a 1000 articles I was approaching 10 MB already.

Because of my ocd I would also very much welcome not hardcoding the automplete index name and let django generate a name for it, as with the other Model indexes. What is this index used for btw?

wgiddens:
This particular instance runs wagtail 2.7, but it should still give an idea. For a database with about 95,000 pages:

Number of rows: 212613
Size of postgres_search_indexentry table: 556 MB
Size of the combined (autocomplete||body) index: 77 MB
Size of autocomplete index: 82 MB
Size of body index: 184 MB

wagtail/contrib/postgres_search/migrations/0004_concat_index.py
```diff
@@ -0,0 +1,22 @@
...
+
+    dependencies = [
+        ('postgres_search', '0003_title'),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            'CREATE INDEX {0}_title_body_concat_search ON {0} '
+            'USING GIN(( title || body));'.format(table),
+            'DROP INDEX {0}_title_body_concat_search;'.format(table),
```
minusf, line 20: maybe `IF EXISTS` would be nice here, django also uses it.

minusf:
This index might be a good candidate also for `CONCURRENTLY` but it's a very new postgres feature and cannot be added unconditionally...

wagtail/contrib/postgres_search/migrations/0004_concat_index.py
wgiddens: That makes sense. I'll add it.

gasman:
Merged in 2da410b2e0221f084aa7c87f49cc9793be40f4ce - thanks @wgiddens!

PR #6565, opened 2020-11-19
Change default WAGTAIL_REDIRECTS_FILE_STORAGE setting to cache

marteinn:
Hi! 

This PR comes from a discussion I had with @kaedroho and @pySilver on how the redirect import treats the uploaded file. 

How it works: To enable a preview of the redirects before import we need to keep the file around after initial upload, there are two methods, as a temporary file or in the builtin cache. The default method is as a temporary file. The temporary file is more effective but does not always work on load balanced environments (such as Heroku).

So in this PR I'm suggesting that we change the default method from using a temporary file to using the builtin cache. This will guarantee that the feature works out of the box on all environments.

The change is minimal. Feedback are much appreciated, cheers!

### DIFF
--- a/docs/reference/settings.rst
+++ b/docs/reference/settings.rst
@@ -655,13 +655,13 @@ Redirects
 
 .. code-block:: python
 
-   WAGTAIL_REDIRECTS_FILE_STORAGE = 'tmp_file'
+   WAGTAIL_REDIRECTS_FILE_STORAGE = 'cache'
 
-By default the redirect importer keeps track of the uploaded file as a temp file, but on certain environments (load balanced/cloud environments), you cannot keep a shared file between environments. For those cases you can use the built-in cache to store the file instead.
+By default the redirect importer keeps track of the uploaded file by storing it in the built-in cache. It's less effective then handling the upload as a actual file, but lets us support load balanced environment out of the box. If you are on a single server environment you can use the more effective tmp_file method instead.
 
 .. code-block:: python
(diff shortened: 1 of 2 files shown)


docs/reference/settings.rst
```diff
@@ -655,13 +655,13 @@ Redirects
 
 .. code-block:: python
 
-   WAGTAIL_REDIRECTS_FILE_STORAGE = 'tmp_file'
+   WAGTAIL_REDIRECTS_FILE_STORAGE = 'cache'
 
-By default the redirect importer keeps track of the uploaded file as a temp file, but on certain environments (load balanced/cloud environments), you cannot keep a shared file between environments. For those cases you can use the built-in cache to store the file instead.
+By default the redirect importer keeps track of the uploaded file by storing it in the built-in cache. It's less effective then handling the upload as a actual file, but lets us support load balanced environment out of the box. If you are on a single server environment you can use the more effective tmp_file method instead.
```
gasman, line 660: @marteinn What exactly does "less effective" mean here?

lb-:
@marteinn I know it has been a while, but are you still keen to pick this one up?
There is a question on the PR above, plus it appears that many of the CI checks are failing (but this may just need to be rebased on `main`)

thibaudcolas:
Checking with Martienn, I have converted this to issue #12797. The proposed change still makes sense, but needs someone else to pick it up and take it through.

PR #6566, opened 2020-11-19
Check if menu_item is not None

Vadim-Karpenko:
For some reason when viewing Snippets in the latest Wagtail version I'm getting this error message -
image

I'm not sure what caused this, maybe that I used a new localization feature in the project (not on this snippet though), maybe this is the problem.

To resolve this I made a simple checkup in the **wagtail/snippets/action_menu.py** that prevents calling the **is_shown** method on the None variables.

Please, let me know if you have any questions.

BR,
Vadim.

### DIFF
--- a/wagtail/admin/action_menu.py
+++ b/wagtail/admin/action_menu.py
@@ -289,7 +289,9 @@ def _get_base_page_action_menu_items():
             PageLockedMenuItem(order=10000),
         ]
         for hook in hooks.get_hooks('register_page_action_menu_item'):
-            BASE_PAGE_ACTION_MENU_ITEMS.append(hook())
+            action_menu_item = hook()
+            if action_menu_item:
+                BASE_PAGE_ACTION_MENU_ITEMS.append(action_menu_item)
 
     return BASE_PAGE_ACTION_MENU_ITEMS
 
--- a/wagtail/snippets/action_menu.py
+++ b/wagtail/snippets/action_menu.py
@@ -119,7 +119,7 @@ class SnippetActionMenu:
         self.menu_items.extend([
             menu_item
             for menu_item in get_base_snippet_action_menu_items(self.context['model'])
-            if menu_item.is_shown(self.request, self.context)
+            if menu_item and menu_item.is_shown(self.request, self.context)
         ])
 
         self.menu_items.sort(key=lambda item: item.order)
--- a/wagtail/snippets/action_menu.py
+++ b/wagtail/snippets/action_menu.py
@@ -100,7 +100,9 @@ def get_base_snippet_action_menu_items(model):
     ]
 
     for hook in hooks.get_hooks('register_snippet_action_menu_item'):
-        menu_items.append(hook(model))
+        action_menu_item = hook(model)
+        if action_menu_item:
+            menu_items.append(action_menu_item)
 
     return menu_items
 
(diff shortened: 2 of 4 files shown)


gasman:
Hi @Vadim-Karpenko - thanks for the PR.
An error like this would indicate that some code has wrongly inserted a None into the list of menu items, through the `register_snippet_action_menu_item` or `construct_snippet_action_menu` hook. In this case, I think it's better to fix that bug at its source, rather than silently masking it when rendering the menu.

Vadim-Karpenko:
> register_snippet_action_menu_item

Turns out it was a problem with the [Wagtail-localize](https://github.com/wagtail/wagtail-localize), thank you a lot for pointing out where the problem might be. I'll prepare a pull request for wagtail-localize as soon as possible.

kaedroho:
According to https://github.com/wagtail/wagtail-localize/pull/257, this menu item is being registered with ``register_snippet_action_menu_item``. The hook function will return ``None`` if the snippet is not translatable.

Any reason we can't add some logic into the code that builds the list of menu items to ignore hook functions that returned ``None``?

gasman:
My original intention with `register_page_action_menu_item` and `register_admin_menu_item` was that you would globally register any menu item that potentially _could_ be shown, and use the `is_shown` method to determine whether it _is_ shown on a given request, just as you suggested in https://github.com/wagtail/wagtail-localize/pull/257#discussion_r527675285.

It looks like `register_snippet_action_menu_item` might have muddied the waters a bit by calling the hook function once per model, rather than once globally, and pasing `model` as a parameter (which, incidentally, isn't mentioned in [the example in the docs](https://docs.wagtail.io/en/stable/reference/hooks.html#register-snippet-action-menu-item) - that should probably be fixed). If it wasn't for that parameter, the need to handle `None` responses wouldn't arise, because it would always be trivial to rewrite

    @hooks.register('register_page_action_menu_item')
    def add_menu_item():
        if some_conditional_logic():
            return MyMenuItem()

into

    if some_conditional_logic():
        @hooks.register('register_page_action_menu_item')
        def add_menu_item():
            return MyMenuItem()

I'd definitely prefer it if wagtail-localize used `is_shown` for this - but given that `register_snippet_action_menu_item` has now introduced a valid reason for handling `None` responses, and it wouldn't break anything to do so, I guess we should go ahead and add that too.

So, in other words: let's go ahead with this fix, but put it [in this spot instead](https://github.com/wagtail/wagtail/blob/2cefe4ebd9f2d691e4cb963ac816ccbc62e8a697/wagtail/snippets/action_menu.py#L103), and do the same for the other `register_*_menu_item` hooks (and add tests).

Vadim-Karpenko:
I've added a fix [here ](https://github.com/wagtail/wagtail/blob/2cefe4ebd9f2d691e4cb963ac816ccbc62e8a697/wagtail/snippets/action_menu.py#L103)and [here](https://github.com/wagtail/wagtail/blob/30a09ae0e6f360e31b52491dc750879c6762ab62/wagtail/admin/action_menu.py#L294) that prevents appending to the list None values. Is there any other places where we need to add a similar fix? Please let me know if so.
Also, about testing, I don't have much experience in this and I would appreciate it if someone will find a time for it.

Please, let me know if you have any questions/concerns.

Vadim-Karpenko:
I have found some time to dive into the tests, could you please take a look at the code and let me know if everything was done correctly?

Vadim-Karpenko:
@gasman sorry for bothering, any update on this?

Vadim-Karpenko:
@gasman Hi, could you please take a look at my recent commit?

gasman:
Thanks - all looks good to me! Merged in 4cfd303db9c29ae582d35820c6e6e4edfd23ec6f.

PR #6568, opened 2020-11-20
Support calling get_image_model and get_document_model at import time

gasman:
Attempting to call get_image_model or get_document_model at the top level of a models.py file currently fails with "Models aren't loaded yet". This can be avoided by passing `require_ready=False` to apps.get_model.

This change makes it possible for third-party apps to define abstract models with foreign key references to the possibly-custom image or document model (which can then be subclassed into concrete models in the project itself - defining concrete models in the third-party app probably still isn't safe, as the model will end up being baked into the third-party app's migrations).

### DIFF
--- a/wagtail/documents/__init__.py
+++ b/wagtail/documents/__init__.py
@@ -23,7 +23,7 @@ def get_document_model():
     from django.apps import apps
     model_string = get_document_model_string()
     try:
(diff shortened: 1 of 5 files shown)


lb-:
Merged in via 7f27e0abbc4e0c011338425968c952db0c63703d
Thanks @gasman

PR #6573, opened 2020-11-22
Update pagination icons to use SVGs

Scotchester:
Updates pagination left and right arrow icons to use SVG icons.

I couldn't figure out how to invoke the `icon` template inclusion tag from inside the `pagination_link_previous` and `pagination_link_next` tags defined in `modeladmin_tags.py`, so I settled for just rendering the `icon.html` template into a variable. Definitely open to hearing better ideas on this!

The other mild concern I had was the fact that there are three templates that create very very similar pagination code. I didn't have time to fully digest the differences, but I see in the release notes for 2.5 that it was definitely a conscious decision to split out `ajax_pagination_nav` from `pagination_nav`. What I'm really not sure about is why `_pagination` and `pagination_nav` both exist. The main difference appears to be how the URLs for the pagination links are constructed, but I'm not sure why they are different or if they _need_ to be. @gasman, looks like you're the primary contributor to both files. Would you have any insight into that?

---

  - There's a failing Draftail Jest snapshot, but I don't believe that would have anything to do with what I've done here
- For Python changes: Have you added tests to cover the new/fixed behaviour?
  - N/A
- For front-end changes: Did you test on all of Wagtail’s supported browsers? **Please list the exact versions you tested**.
  - No, but should work the same as other SVG icon changes
- For new features: Has the documentation been updated accordingly?
  - N/A

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/pages/listing/_pagination.html
+++ b/wagtail/admin/templates/wagtailadmin/pages/listing/_pagination.html
@@ -11,12 +11,18 @@ Pagination for page listings. Used by the `{% paginate %}` template tag.
     <ul>
         <li class="prev">
             {% if page.has_previous %}
-                <a data-page="{{ page.previous_page_number }}" href="{{ base_url }}{% pagination_querystring page.previous_page_number page_key=page_key %}" class="icon icon-arrow-left {{ classnames }}">{% trans "Previous" %}</a>
+                <a data-page="{{ page.previous_page_number }}" href="{{ base_url }}{% pagination_querystring page.previous_page_number page_key=page_key %}" class="{{ classnames }}">
+                    {% icon name="arrow-left" class_name="default" %}
+                    {% trans "Previous" %}
+                </a>
             {% endif %}
         </li>
         <li class="next">
             {% if page.has_next %}
-                <a data-page="{{ page.next_page_number }}" href="{{ base_url }}{% pagination_querystring page.next_page_number page_key=page_key %}" class="icon icon-arrow-right-after {{ classnames }}">{% trans "Next" %}</a>
+                <a data-page="{{ page.next_page_number }}" href="{{ base_url }}{% pagination_querystring page.next_page_number page_key=page_key %}" class="{{ classnames }}">
+                    {% trans "Next" %}
+                    {% icon name="arrow-right" class_name="default" %}
+                </a>
             {% endif %}
         </li>
     </ul>
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -5,6 +5,7 @@ Changelog
 ~~~~~~~~~~~~~~~~~
 
  * Added support for Python 3.9
+ * Switch pagination icons to use SVG instead of icon fonts (Scott Cranfill)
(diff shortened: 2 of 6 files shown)


gasman:
>  @gasman, looks like you're the primary contributor to both files. Would you have any insight into that?

From what I can remember, before I split it up into multiple templates it was a nasty hodgepodge of `{% if %}` / `{% else %}` cases with not much in common, so it was really one template doing several different jobs that looked superficially similar. I suspect I was in the middle of some wider template refactoring and didn't want to get sidetracked into rewriting the view code to make everything follow the same pagination logic - as far as I'm concerned there's no fundamental reason to keep these separated, though, and I'd be very happy to see this cleaned up further!

Scotchester:
> > @gasman, looks like you're the primary contributor to both files. Would you have any insight into that?
> 
> From what I can remember, before I split it up into multiple templates it was a nasty hodgepodge of `{% if %}` / `{% else %}` cases with not much in common, so it was really one template doing several different jobs that looked superficially similar. I suspect I was in the middle of some wider template refactoring and didn't want to get sidetracked into rewriting the view code to make everything follow the same pagination logic - as far as I'm concerned there's no fundamental reason to keep these separated, though, and I'd be very happy to see this cleaned up further!

Thanks for that! I'll see if Thibaud and/or Coen think it's worth trying to do that here, or if we should just keep the scope restricted to the icon changes.

thibaudcolas:
This looks simple to review, should we try to get this in as it is and you do the refactoring in a separate PR? Or would the refactoring drastically change how the icons are defined?

Scotchester:
@thibaudcolas Yeah, let's hold the refactoring for another PR. It wouldn't change the icon updates, really.

wagtail/admin/templates/wagtailadmin/pages/listing/_pagination.html
```diff
@@ -11,12 +11,18 @@
     <ul>
         <li class="prev">
             {% if page.has_previous %}
-                <a data-page="{{ page.previous_page_number }}" href="{{ base_url }}{% pagination_querystring page.previous_page_number page_key=page_key %}" class="icon icon-arrow-left {{ classnames }}">{% trans "Previous" %}</a>
+                <a data-page="{{ page.previous_page_number }}" href="{{ base_url }}{% pagination_querystring page.previous_page_number page_key=page_key %}" class="{{ classnames }}">
+                    {% icon name="arrow-left" class_name="default" %}
```
thibaudcolas, line 15: I’m not a big fan of this `default` class name (default what?), but this was here before so oh well.

thibaudcolas (approved):
This looks good to me, with the caveat that the icons are much bigger than before – I think I like it like that, but just wanted to confirm it’s intentional and we’re happy for it to stay.

Before/after:

pagination-icons

Tested in IE11 and Chrome, with the image listings, image chooser, and modeladmin listings.

Scotchester:
Yes, the larger size was intentional and I forgot to comment on it! Sorry about that. Glad you like it :)

Yeah, the `default` class name is not great, but it's already in use to give SVG icons a "default" size that is actually more reasonable than what the SVG would produce on its own (which is very large), so I used that here, too.

Issue #6393, opened 2020-09-21
Migration fails with MS SQL 


rjschave:
### Issue Summary

Wagtail versions < 2.10 have worked with Microsoft SQL Server.  Wagtail 2.10 introduced a change (related to configurable moderation workflow) that is not supported by Microsoft SQL Server.  Specifically, Microsoft SQL Server does not support the OR keyword in the WHERE clause for unique indexes.

Migration 0050 in Wagtail Core (issue #6047 and commit 74a8cf8) only allows one record with status of In Progress **OR** Needs Changes per page. 

```
migrations.AddConstraint(
    model_name='workflowstate',
    constraint=models.UniqueConstraint(condition=models.Q(('status', 'in_progress'), ('status', 'needs_changes'), _connector='OR'), fields=('page',), name='unique_in_progress_workflow'),
),
```

This migration generates the following SQL statement, which is invalid:

```
CREATE UNIQUE INDEX [unique_in_progress_widget]
    ON [widgets_widgetstate] ([widget_id])
    WHERE ([status] = 'in_progress' OR [status] = 'needs_changes');
```

### Steps to Reproduce

1. Start a new project with `wagtail start myproject`.
2. Configure Django to connect to MS SQL server.

Example:
```
DATABASES = {
    'default': {
        'ENGINE': 'sql_server.pyodbc',
        'HOST': 'localhost',
        'PORT': '',
        'NAME': 'database-name',
        'USER': 'username',
        'PASSWORD': 'password',
        'OPTIONS': {
            'driver': 'ODBC Driver 17 for SQL Server',
        },
    },
}
```

3. Run `migrate` command.

Error message:

```
 Applying wagtailcore.0050_workflow_rejected_to_needs_changes...Traceback (most recent call last):
  File "/Users/rjschave/pycharm-projects/clients/rjs/wagtailtest2/venv/lib/python3.8/site-packages/django/db/backends/utils.py", line 86, in _execute
    return self.cursor.execute(sql, params)
  File "/Users/rjschave/pycharm-projects/clients/rjs/wagtailtest2/venv/lib/python3.8/site-packages/sql_server/pyodbc/base.py", line 553, in execute
    return self.cursor.execute(sql, params)
pyodbc.ProgrammingError: ('42000', "[42000] [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Incorrect syntax near the keyword 'OR'. (156) (SQLExecDirectW)")
```

[...]

PR #6585, opened 2020-11-26
Improve test database settings to fix SQLite error when creating migrations

davidjb:
This change makes several improvements:

* Use in-memory SQLite database for test migrations

  The default database is SQLite but its `NAME` (which SQLite uses as the filename) was `wagtail`, which isn't valid since the wagtail codebase already has a `wagtail/` directory.  Trying to run migration creation commands (https://docs.wagtail.io/en/latest/contributing/developing.html#running-migrations-for-the-test-app-models) produced an error:

  ```
    File "wagtail/lib/python3.9/site-packages/django/db/backends/base/base.py", line 219, in ensure_connection
      self.connect()
    File "wagtail/lib/python3.9/site-packages/django/utils/asyncio.py", line 26, in inner
      return func(*args, **kwargs)
    File "wagtail/lib/python3.9/site-packages/django/db/backends/base/base.py", line 200, in connect
      self.connection = self.get_new_connection(conn_params)
    File "wagtail/lib/python3.9/site-packages/django/utils/asyncio.py", line 26, in inner
      return func(*args, **kwargs)
    File "wagtail/lib/python3.9/site-packages/django/db/backends/sqlite3/base.py", line 207, in get_new_connection
      conn = Database.connect(**conn_params)
  sqlite3.OperationalError: unable to open database file
  ```

  because of this conflict. (This traceback also causes a subsequent error `django.db.utils.OperationalError: unable to open database file`)

  This change uses `:memory` as the default database filename if SQLite is used, thereby fixing the above issue and avoiding a potential for an empty SQLite database file to be created (if DATABASE_NAME was the path to a real file; as it was in the first version of this PR).

  Other non-SQLite engines continue to use the original `wagtail` name, meaning that the `test_wagtail` database gets created just as before.

* Modifies the default values for database USER, PASSWORD, HOST etc to being an empty string rather than None, to match Django's [defaults](https://docs.djangoproject.com/en/stable/ref/settings/#host).  This helps avoid any potential issues as some parts of Django rely on certain values being a string (e.g. NAME where `test_` gets appended).

* Adds documentation to `developing.rst` regarding installation of required database modules and available environment variables for database connection customisation (`PGPORT`, `DATABASE_*` etc)

* Normalises the `DATABASE_PASSWORD` tests environment variable to match the name in Django's database settings object

----

* [ ] For front-end changes: Did you test on all of Wagtail’s supported browsers? **Please list the exact versions you tested**.

### DIFF
--- a/docs/contributing/developing.rst
+++ b/docs/contributing/developing.rst
@@ -106,6 +106,10 @@ You can create migrations for the test app by running the following from the Wag
 Testing against PostgreSQL
 --------------------------
 
+.. note::
+
+   In order to run these tests, you must install the required modules for PostgreSQL as described in Django's `Databases documentation`_.
+
 By default, Wagtail tests against SQLite. You can switch to using PostgreSQL by
 using the ``--postgres`` argument:
 
@@ -113,10 +117,14 @@ using the ``--postgres`` argument:
 
     $ python runtests.py --postgres
 
-If you need to use a different user, password or host. Use the ``PGUSER``, ``PGPASSWORD`` and ``PGHOST`` environment variables.
+If you need to use a different user, password, host or port, use the ``PGUSER``, ``PGPASSWORD``, ``PGHOST`` and ``PGPORT`` environment variables respectively.
 
 Testing against a different database
 ------------------------------------
+
+.. note::
+
+   In order to run these tests, you must install the required client libraries and modules for the given database as described in Django's `Databases`_ documentation or 3rd-party database backend's documentation.
 
 If you need to test against a different database, set the ``DATABASE_ENGINE``
 environment variable to the name of the Django database backend to test against:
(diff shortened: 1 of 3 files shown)


gasman:
Thanks @davidjb! Do the other commands given in the docs (e.g. `python runtests.py --postgres`) still work with this change?

davidjb:
@gasman I've just updated and pushed a new iteration of this PR; I hadn't fully understood the scope of how the `NAME` was utilised through the tests on different databases.  So, now `:memory:` is used just for SQLite databases but for _any_ other engines, the original `wagtail` name is used -- so to answer your question, it won't affect them and tests do all run okay (tested on SQLite, MySQL, Postgres and working on MS SQL Server support as per #6393) 👍 

The PR also sets the default values for these database variables to be an empty string (matching Django's defaults) rather than `None` - I encountered a crash inside Django where the previous default None was trying to be appended to a string.  I took the opportunity to add documentation about the environment variables used in testing and normalised one of their names (`DATABASE_PASSWORD`) to match Django's docs.

Let me know if any me any changes are needed.  The changes are all related and somewhat intertwingled (e.g. can't run tests until the default env variables get set) -- otherwise I would have broken them up more.

gasman:
All looks good to me, thanks! Merged in 67235accfcd59ba9e2fd0a2d86dfb388fac4be38.

PR #6586, opened 2020-11-26
Let modeladmin customize field display if needed

frmdstryr:
This lets each `modeladmin` customize the display value for specific fields if needed.

The primary motivation for this is so the display can be customized __without__ breaking sorting since using a method on the model admin disables the sorting link.

```python

    def get_display_for_field(self, obj, value, field, empty_value_display):
        if field.name == "due_date":
            return naturaltime(value)
        return super().get_display_for_field(obj, value, field, empty_value_display)
```

it also makes it easier to customize certain values, eg 

```python
    def get_display_for_field(self, obj, value, field, empty_value_display):
        if isinstance(value, Image):
            return get_image_thumbnail(value)
        return super().get_display_for_field(obj, value, field, empty_value_display)

```

### DIFF
--- a/wagtail/contrib/modeladmin/options.py
+++ b/wagtail/contrib/modeladmin/options.py
@@ -3,7 +3,7 @@ from django.contrib.admin.utils import display_for_field
 from django.contrib.auth.models import Permission
 from django.core import checks
 from django.core.exceptions import ImproperlyConfigured
-from django.db.models import Model
+from django.db.models import Model, ManyToOneRel
 from django.urls import re_path
 from django.utils.safestring import mark_safe
 
@@ -218,11 +218,16 @@ class ModelAdmin(WagtailRegisterable):
         """
         return mark_safe(self.empty_value_display)
 
-    def get_display_for_field(self, value, field, empty_value_display):
+    def get_display_for_field(self, instance, value, field, empty_value_display):
         """
         Returns the field display value. Override this to
         customize the field display for a specific model admin.
         """
+        if isinstance(field, ManyToOneRel):
+            field_val = getattr(instance, field.name)
+            if field_val is None:
+                return empty_value_display
+            return field_val
         return display_for_field(value, field, empty_value_display)
 
     def get_list_filter(self, request):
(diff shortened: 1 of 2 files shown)


jacobtoppm (approved):
Thanks for contributing! This looks good to me - nice little win for customisation

ababic:
NOTE: The recommended way to retain sorting for custom methods in `list_display` is to add an `admin_order_field` attribute to the method. For example:

```
from django.utils.html import format_html
from wagtail.contrib.modeladmin.options import ModelAdmin

class PersonModelAdmin(ModelAdmin):
    model = Person
    list_display = ['coloured_first_name', 'last_name']

    def colored_first_name(self, obj):
        return format_html(
            '<span style="color: #{};">{}</span>',
            obj.color_code,
            obj.first_name,
        )
    colored_first_name.admin_order_field = 'first_name'
```

frmdstryr:
> The recommended way to retain sorting for custom methods in list_display is to add an admin_order_field attribute to the method.

Oh nice, didn't know that... thanks for sharing! Is there a decorator to tag the modeladmin methods instead of having to repeat the name, eg `@admin_field(short_description="...")` or something?

This still has some other use cases (I also use it to auto-generate links to the modeladmin for related fields) but I ended up just copying the modeladmin template tags and overriding the template in my project so it's no big deal if it doesn't get merged.

ababic:
You're welcome @frmdstryr! This is something that was copied over from the Django's version of modeladmin, do you can do the same thing there too.

> Is there a decorator to tag the modeladmin methods instead of having to repeat the name, eg @admin_field(short_description="...") or something?

There isn't, but I can see how that would be useful. I don't suppose you'd be interested in writing one? (I'd be happy to review). 

> This still has some other use cases (I also use it to auto-generate links to the modeladmin for related fields)

This should be easy enough to achieve with a regular ModelAdmin method like that in the above example. You just need to use [format_html](https://docs.djangoproject.com/en/3.1/ref/utils/#django.utils.html.format_html) to create the return value, or wrap your html string with [mark_safe](https://docs.djangoproject.com/en/3.1/ref/utils/#django.utils.safestring.mark_safe) to prevent escaping.

PR #6589, opened 2020-11-27
Changes repr method of Phrase to label the object Phrase instead of PlainText

alistairclark:
Fixes an error in the repr method of Phrase


kaedroho:
Thanks!

PR #759, opened 2014-10-27
Management command to generate thumbnails that may be missing

damianmoore:
Occasionally, you can create problems when syncing data from production environments or you just want to clear out unused thumbnails. I thought this seemed like a useful tool that others might need.

PR #6595, opened 2020-11-28
rm_cropped_images from wagtail

onno-timmerman:
_My first pull request please double check it!_

This is a command will remove all rendered images. Other packages have similar commands like sorl thumbnailing.
I hope this command is complete enough to cover situations with and without custom image model.


wagtail/core/management/commands/rm_cropped_images.py
```diff
@@ -0,0 +1,53 @@
...
+            msg.append(self.style.WARNING('MAKE SURE THE PATH POINTS TO YOUR CROPPED IMAGES!!!\n'))
+            msg.append('There are %(count)d images cropped refrenced in DB.\n' % {'count': images.count()})
+            msg.append('These cropped images will be deleted by deleting the image folder.\n')
+            msg.append('Don\'t worry on visiting the website they will be recreated on the fly.\n')
+            msg.append('Are you sure you want to delete? yes / no: ')
+            if input(''.join(msg)) == 'yes':
+                # empty db from cropped images
+                images.delete()
+                # removing directory with cropped images
+                shutil.rmtree(image_root)
```
jacobtoppm, line 48: How does this interact with using non-local storage options? Is there a reason you don't want to rely on the signal handlers https://github.com/wagtail/wagtail/blob/master/wagtail/images/signal_handlers.py to do the file cleanup?

wagtail/core/management/commands/rm_cropped_images.py
```diff
@@ -0,0 +1,53 @@
+import os, shutil
+from django.conf import settings
+from django.apps import apps
+from django.core.management.base import BaseCommand, CommandError
+
+# we check if there is a custom image model and rendition model
+# if a custom image model exist we introduce a extra setting
+if hasattr(settings, 'WAGTAILIMAGES_IMAGE_MODEL_RENDITION'):
+    app, model = settings.WAGTAILIMAGES_IMAGE_MODEL_RENDITION.split('.')
```
jacobtoppm, line 9: Can you get the Rendition model using its foreign key to a custom image model instead? This extra setting for one small use case seems redundant. If it is necessary, it should be documented in the settings reference.

jacobtoppm (changes requested):
Thanks for contributing! There's a few wording/clarity issues we can address later, but wanted to add some initial feedback. In addition to the main comments I've added, you use "rendition", "cropped image", and "compiled image" pretty interchangably here. I think it would be good to just use "rendition". (There's also a couple of flake8 issues which the backend tests are flagging up - should be pretty easy to fix)

thibaudcolas:
Thank you for looking into this @onno-timmerman :) Note for future reference – this looks like it could replace #759.

wagtail/core/management/commands/rm_cropped_images.py
onno-timmerman: I'm afraid I'm not aware of the post delete signal handlers. Is this a problem?

Naming, I'm dyslectic so anything that improves spelling etc is super...

wagtail/core/management/commands/rm_cropped_images.py
onno-timmerman: I tried, and this is annoying but I didn't see an good way to link the two models and detact what the model of the renditions would be? Do you have any suggestions so we can drop that extra Wagtail setting.

wagtail/core/management/commands/rm_cropped_images.py
jacobtoppm: I guess I just mean a) this might be redundant and b) I suspect it may not work for people using remote image storage. Could you try it without the directory removal and see if it still works (as those signal handlers I mentioned should take care of file deletion when the rendition itself is deleted)?

wagtail/core/management/commands/rm_cropped_images.py
jacobtoppm: When you've retrieved the image model, I think you should be able to use its classmethod `get_rendition_model`

wagtail/core/management/commands/rm_cropped_images.py
jacobtoppm: As for naming - no worries, let's get the functionality figured out and then I'm happy to do a pass with you on the wording/spelling.

wagtail/core/management/commands/rm_cropped_images.py
onno-timmerman: ah if so I'll look into it this weekend then can drop the extra setting :)

jacobtoppm: Great, thank you!

docs/reference/management_commands.rst
```diff
@@ -125,3 +125,12 @@ search_garbage_collect
     $ ./manage.py search_garbage_collect
 
 Wagtail keeps a log of search queries that are popular on your website. On high traffic websites, this log may get big and you may want to clean out old search queries. This command cleans out all search query logs that are more than one week old (or a number of days configurable through the :ref:`WAGTAILSEARCH_HITS_MAX_AGE <wagtailsearch_hits_max_age>` setting).
+
+rm_cropped_images
```
zerolab, line 129: I suggest this is renamed to `remove_image_renditions`, or `remove_generated_images`

zerolab:
@onno-timmerman I was wondering if you have time to revisit this PR? It feels like a good addition that would help lots of Wagtail users.

The changes needed:
- [ ] Make use of `Image.get_rendition_model()` (see https://github.com/wagtail/wagtail/blob/571b9e1918861bcbe64a9005a4dc65f3a1fe7a15/wagtail/images/signal_handlers.py#L26-L27)
- [ ] drop the `shutil.rmtree(image_root)` bit, as the actual file deletion is done via signals.
- [ ] add a test to confirm the deletion
- [ ] revisit the spelling

lb-:
Not sure where this PR is at but I have created https://github.com/wagtail/wagtail/issues/8107 as a issue to cover the general request for this kind of management command.

docs/advanced_topics/api/v2/usage.rst
```diff
@@ -478,6 +478,9 @@ This can also be used across relationships. For example,
 ``?fields=body,feed_image(width,height)`` will nest the ``width`` and ``height``
 of the image in the response.
 
+Notice if you use fields in combination with ``child_of`` you also need the ``type``
+parameter to use fields.
+
```
lb-, line 483: Just reading this, it might be best to split this change out into its own PR, will be a small one and can be reviewed in isolation.

lb-:
superseded by #8166

PR #6598, opened 2020-11-29
A basic template suitable for use as WAGTAIL_FRONTEND_LOGIN_TEMPLATE

damascene:


lb-:
@damascene - thanks for this draft PR - I know it has been a while but just checking in to see if you are still wanting to pursue this change (the template example).

Otherwise maybe it would be good to close this and just raise a documentation enhancement request with your WIP so far.

Relevant parts of the docs for reference

* https://docs.wagtail.org/en/stable/advanced_topics/privacy.html#setting-up-a-login-page
* https://docs.wagtail.org/en/stable/reference/settings.html#wagtail-frontend-login-template

lb-:
Just reviewed this again, it appears that everything this PR attempts to do (in regards to the login template) has been covered in the docs already. Looks like it was done after this PR was created.

There is one small addition I will cherry pick from this though - the note about being authenticated and still being prompted for access.

Thanks @damascene  - this is done via 5d78e03084bcdfbea3bb743de8d374e7478b608a

As for the signals docs - we have a basic covering of this also in the docs and a link to the Django docs for signals for anyone needing to get started.

Issue #6563, opened 2020-11-19
StreamValue._prefetch_blocks is not called in preview


pySilver:
### Issue Summary

Optimization that has recently landed into Wagtail StreamFields https://github.com/wagtail/wagtail/pull/5976 https://github.com/wagtail/wagtail/issues/5926 is awesome but works only for serve mode and isn't triggered in preview which causes confusion.

Use case:

* Very complex page with multiple widgets that triggers external storage like ElasticSearch for every StructBlock instance.

While serving page to users we can implement bulk data loading in `MyWidgetBlock.bulk_to_python`. However this method is not called when page is in preview mode, so we have to  implement duplicate code that runs on preview mode only (and this implementation will be slow).

### Technical details

* Python version: 3.8.x
* Django version: 3.1.x
* Wagtail version: 2.11.2

gasman:
`bulk_to_python` is not called on preview because it's responsible for converting the database JSON representation to native Python objects, and on preview the data isn't coming from database JSON - it's coming from a form submission. (One that's been serialised/deserialised in the session, in fact, but that's not important here...) The relevant Block method for handling those form submissions is `value_from_datadict`.

I can see roughly how a `bulk_value_from_datadict` method analogous to `bulk_to_python` might work: it accepts a `data` and `files` dict, and a _list_ of prefix strings, and returns a list of Python block values obtained from the form fields at those prefixes. One possible tricky bit is that FieldBlock typically delegates the final 'extract value from form field' task to a form widget object, so you may end up having to hack a `bulk_value_from_datadict` method into those too.

gasman:
Actually, now I think of it, the upcoming [Telepath rewrite of Streamfield](https://wagtail.io/blog/telepath/) is likely to completely change how form processing works (possibly not initially, but in the second phase of development) - so maybe best to hold off on this for now...

pySilver:
@gasman yeah, for the time being I've implemented support for both preview/non-preview mode so production is not failing and at least works fast for customer facing traffic.

Telepath seems to be very interesting :)

PR #6603, opened 2020-12-01
Split out reusable _map helper methods from bulk_to_python

gasman:
A bit of refactoring done as foundational work for the Telepath streamfield rewrite, although it may turn out not to be needed... putting it here so it doesn't get lost.

This allows the logic behind the StructBlock / ListBlock / StreamBlock implementations of `bulk_to_python` (where the values to be converted are distributed to the appropriate child blocks and recombined) to be re-used for other kinds of value conversion. Possibly also relevant to #6563.

### DIFF
--- a/wagtail/core/blocks/list_block.py
+++ b/wagtail/core/blocks/list_block.py
@@ -144,11 +144,20 @@ class ListBlock(Block):
         return self.child_block.bulk_to_python(value)
 
     def bulk_to_python(self, values):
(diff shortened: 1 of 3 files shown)


lb-:
Do we still need this PR @gasman ?

gasman:
@lb- Nah, we can close this. It's hard to say how much benefit we'll get from pulling out the recursive block methods into a common pattern, and this code will still be around if we decide we do need it.

PR #6605, opened 2020-12-02
Removing ambiguity in the Template Override ModelAdmin documentation

JoeWHoward:
As it stands, the docs on this topic (Overriding Templates in ModelAdmin) seem to indicate that the /modeladmin/ directory should be placed in the project root folder.  I think a rewording of line 144 is in order as well, but I don't know a good way to phrase that.

### DIFF
--- a/wagtail/search/query.py
+++ b/wagtail/search/query.py
@@ -45,7 +45,7 @@ class Phrase(SearchQuery):
         self.query_string = query_string
 
     def __repr__(self):
-        return '<PlainText {}>'.format(repr(self.query_string))
+        return '<Phrase {}>'.format(repr(self.query_string))
 
 
 class MatchAll(SearchQuery):


jacobtoppm:
Thanks for contributing! I definitely think this clears up some ambiguity. For line 144, what about 'project's apps' - do you think that's clearer? I agree that bit still sounds a little like it's referring to the root.

thibaudcolas:
👋  does this replace #5083 or is that other PR still relevant?

docs/reference/contrib/modeladmin/primer.rst
```diff
@@ -143,14 +143,14 @@ Overriding templates
 For all modeladmin views, Wagtail looks for templates in the following folders
 within your project, before resorting to the defaults:
```
JoeWHoward, line 144: within your project or app, before resorting to the defaults:

JoeWHoward (commented):
From the discussion in #5083 it seems that the templates/modeladmin folder can exist in either the project directory or an app directory, but the inclusion of the templates parent folder is the same either way.

I understand that this is something that should be "inferred" from working with Django, but I think the example paths are misleading if no mention of the templates directory is made on the page.

lb-:
@JoeWHoward  I made that last little change locally, thanks so much for taking the time to improve the docs.

This has now been merged in via 4ab7080fb8e3def96ec6e82df20c9b38e1a93f2e

PR #6607, opened 2020-12-03
Change UniqueConstraint for SQL Server (and Postgres) database support

### LINKED ISSUE #6393: Migration fails with MS SQL

davidjb:
This fixes #6393 by modifying the constraint to use an IN condition which supports both Postgres and SQL Server. Previously, the `|` (OR) condition was only supported by Postgres because SQL Server only supports AND conditions, meaning that migrations would crash mid-way through on the latter.

The implementation follows suggestions from @gasman in https://github.com/wagtail/wagtail/issues/6393#issuecomment-732161057:

* Migration 0050 is modified to not break on SQL Server
* Added migration 0060 to add or replace the constraint in a manner that supports both of these databases

Additionally, this allows for and documents a `DATABASE_DRIVER` env variable to be set for testing, to allow a different SQL Server driver (e.g. FreeTDS on Mac/Linux); and adds the specific `host_is_server` option for FreeTDS (won't affect SQL Server Native Client on CI).

Note: 8 other database-related tests are failing (see https://gist.github.com/davidjb/be2849b5bbb5d4c3e3db1e317d644a17 for log) when tested against SQL Server; this could be a FreeTDS driver issue or something else but they don't seem related to this PR.

Other database backends (SQLite, MySQL) see their tests pass.

This PR depends on changes in #6585, hence why it has 2 commits present.  Once that other is merged, it's just one.

----

* [x] Do the tests still pass? (https://docs.wagtail.io/en/latest/contributing/developing.html#testing) - yes existing tests do on _other_ database engines; please see comments
* [x] For Python changes: Have you added tests to cover the new/fixed behaviour? - the fact the tests run?


tomdyson:
Thanks @davidjb! This looks really good. Can you give us any advice on how to test against SQL Server, either locally or in CI? I don't think any of the core team have much SQL Server expertise.

davidjb:
@tomdyson Thanks for the feedback. 

For sure. Here’s my cheat sheet for running wagtail tests on all the various db backends: https://gist.github.com/davidjb/643272824f469386f933e7a4e081a81e - SQL Server is down the bottom. 

I spin up a Docker container for each server - thankfully SQL Server was released for Linux, install the requisite engine and client and run the tests. In this case, the install instructions are for Mac but adaptable for Linux - not sure about Windows. For CI, you could run the Docker container and then the tests; I’ve done this style of thing on Travis.  If that doc is helpful, feel free to take it for the wagtail docs or I can submit a PR. 

You’ll notice things are a little messy at the moment on the latest version of Django 3.1 because of the engine needing fixes for db internal changes. I’ve got pull requests open to the engine repo at https://github.com/ESSolutions/django-mssql-backend/pull/86 which will be merged at some point (code is broken in 3.1+ without it), hence the temporary use of a branch in my fork. Otherwise you could use the backend’s current master branch with Django 3.0 in tests.

davidjb:
Rebased on the latest `master` so it's able to be merged; anything else you'd like me to change or update?

gasman:
Thanks @davidjb - all looks good to me! Merged in 3a5b7255adfb7eefbd9a48b56fdb7654b636d7cb.

davidjb:
Thanks @gasman - much appreciated.  I also created the work-in-progress https://github.com/ESSolutions/django-mssql-backend/pull/97 with the database backend to catch situations where an unsupportable constraint is present.

PR #6609, opened 2020-12-03
Add {% if page %} around language selector example

gasman:
As suggested here: https://stackoverflow.com/questions/65065301/wagtail-not-redirecting-500-error-instead-of-404-when-debug-false/65065608?noredirect=1#comment115125518_65065608


kaedroho:
Thanks!

Araoo:
It's exactly the same issue if you use wagtail-metadata. You need to add the {% if page %} around {% meta_tags %} in order to make 404 works.

PR #6612, opened 2020-12-03
Add system check to validate that a model `edit_handler` is actually an EditHandler

frmdstryr:
This adds a system check that makes sure the `edit_handler` (if any) assigned on a model that is registered using a modeladmin is actually EditHandler instance.  It also will warn if you accidentally assign `edit_handlers` which I seem to do by accident a lot...

For example
```
ERRORS:
menu.Menu: (wagtailadmin.E003) edit_handler is not an EditHandler instance
        HINT: Ensure that Menu.edit_handler is an instance of wagtail.admin.edit_handler.EditHandler. It is currently: `<class 'tuple'>`

website.Configuration: (wagtailadmin.W003) edit_handlers has no effect
        HINT: Did you mean to define an `edit_handler`?
```

Is there a way to test this?

### DIFF
--- a/wagtail/admin/checks.py
+++ b/wagtail/admin/checks.py
@@ -151,3 +151,33 @@ There are no default tabs on non-Page models so there will be no \
         errors.append(error)
 
     return errors
+
+
+def check_edit_handler_on_model(cls, context='model'):
+    """Check that the edit_handler is an instance of EditHandler."""
+    from wagtail.admin.edit_handlers import EditHandler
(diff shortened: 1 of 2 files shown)


frmdstryr:
Guess this still needs some work since certain edit handlers subclasses are still invalid.

 eg if I use a `MultiFieldPanel` there's still an error `'MultiFieldPanel' object has no attribute 'get_form_class'`

PR #6614, opened 2020-12-04
Move all legacy JS into client folder

kaedroho:
This allows us to start refactoring these bits to use modern JS. Each file is now a new webpack entry point, but we can combine these into larger bundles in a future PR.


kaedroho:
Looks like that code might need a bit of cleanup as it wasn't being linted previously. The main thing being the indentation level of the ``client`` folder is 2, but most of the files in wagtailadmin/js were indented at 4.

gasman (approved):
Webpack isn't really my field of expertise, but all makes sense and works nicely in my tests!

gasman:
Merged in 27133f957074635bf6efe461eff9f61d206d8443.

PR #6617, opened 2020-12-07
Feature/comment saving

jacobtoppm:
Note: this PR depends on https://github.com/jacobtoppm/wagtail-comment-frontend/pull/2 (`npm link` it locally), which also needs reviewing for the FE changes.

This PR integrates the commenting frontend and backend, so that comments can be saved to and loaded from the database. Currently, we only deal with field level comments, which have only been added to choosers so far (as text fields will later be getting inline comments instead). Comments are implemented via a new panel, `CommentPanel`, and their models are changed to use `ParentalKey` rather than `ForeignKey` to `page` in order to simplify saving along with the page/revision. Similar to attributes like `draft_title`, comments are kept in sync across all revisions of a page (positions will later be allowed to vary when inline commenting is implemented). Comments are saved using formsets. (Ideally we'd be sticking with the API driven commenting approach of wagtail-review, but unfortunately they need to be saved with the revision until we bring autosaving into Wagtail)

Some creating of form subclasses within the edit handler methods is currently necessary to allow user-specific comment validation on save - if anyone has alternative suggestions for doing this, I'm definitely open!

Comment models are also adjusted such that resolving a comment is equivalent to deleting it, so comments are for now a simple feedback mechanism and we don't have to deal with the UI of showing resolved comments on old revisions in the first version. This is something we can definitely revisit later.

Known issues:
- Not IE11 compatible due to use of shadow-dom
- Historical wagtail review code needs to be reformatted to match Wagtail's linting config, and eventually brought into Wagtail. Reformatting will be done in the following ticket.
- Many elements of the UI are placeholders (the top 'show comments' bar, the comment styling, the annotation button being permanently visible rather than on hover/focus), and will be replaced in time.


wagtail/admin/forms/comments.py
```diff
@@ -0,0 +1,47 @@
...
+        cleaned_data = super().clean()
+        user = self.user
+
+        if not self.instance.pk:
+            self.instance.user = user
+        elif self.instance.user != user:
+            # trying to edit someone else's comment
+            if any(field for field in self.changed_data if field != DELETION_FIELD_NAME):
+                # users can delete each other's base comments, as this is just the "resolve" action
+                self.add_error(None, error=ValidationError(_("You cannot edit another user's comment")))
```
tm-kn, line 45: **Suggestion:** `error` is just a positional parameter - I can't see a reason why you'd explicitly add `error=` keyword argument. https://github.com/django/django/blob/148702e7258eafb27d5488b33a723c87d898e22b/django/forms/forms.py#L306
Also, full stop?

wagtail/admin/edit_handlers.py
```diff
@@ -789,8 +792,89 @@ def render(self):
...
+    def get_context(self):
+        user = getattr(self.request, 'user', None)
+        authors = {user.pk: user_display_name(user)} if user else {}
+        serialized_comments = []
+        for comment in self.comments:
+            # iterate over comments to retrieve user display names and serialized versions
+            if comment.user_id not in authors:
+                authors.update({comment.user.pk: user_display_name(comment.user)})
+            serialized_comments.append(comment.serializable_data())
+            for reply in comment.replies.all():
```
tm-kn, line 851: **Thought:** Could do a simple optimisation? Not sure how well this works with modelcluster in the context you are here in.

```suggestion
            for reply in comment.replies.select_related('user'):
```

_This would be redundant if you prefetch it level higher, then you must use `all`._

I am not sure how possible it is to prefetch replies if we are not fetching all of this stuff from a database table, but it might be a nice optimisation.

client/src/components/CommentApp/comments.js
```diff
@@ -28,13 +29,13 @@ function getContentPath(fieldNode) {
     return '';
   }
   let element = fieldNode.closest('[data-contentpath]');
-  const contentPaths = [];
+  const contentpaths = [];
```
tm-kn, line 32: **Question:** Are we suddenly re-branding "contentpath" to be a one word? :smile: Let's stay consistent everywhere if possible.

wagtail/admin/edit_handlers.py
```diff
@@ -789,8 +792,89 @@ def render(self):
...
+        for form in comment_forms:
+            instance = form.instance
+            instance.replies = [reply_form.instance for reply_form in form.formsets['replies'].forms]
+            self.comments.append(instance)
+
+    def get_context(self):
+        user = getattr(self.request, 'user', None)
+        authors = {user.pk: user_display_name(user)} if user else {}
+        serialized_comments = []
+        for comment in self.comments:
```
tm-kn, line 846: **Thought:** If I could only do something like this... I don't quite get whether all of this is coming from a database or not.

```python
pks = [comment.pk for comment in self.comments]
qs = Comment.objects.filter(pk__in=pks).select_related('user').prefetch_related('replies__user').in_bulk()
for qs[pk] in pks:
```
Alternatively maybe prefetch the users in one go to save some queries.

```python
user_pks = []
for comment in self.comments:
    user_pks.append(comment.user_id)
     # all() should hopefully mean this will be cached for the future if you don't use filter etc.
    user_pks += [reply.user_id for reply in comment.replies.all()]
users = get_user_model().objects.filter(pk__in=user_pks).in_bulk()

# Down the line
users[reply.user_id]
users[comment.user_id]
```

tm-kn (changes requested):

wagtail/admin/edit_handlers.py
```diff
@@ -789,8 +792,89 @@ def render(self):
...
+    def __init__(self, *args, **kwargs):
+        self.comments = []
+        super().__init__(*args, **kwargs)
+
+    def required_formsets(self):
+        # add the comments formset
+        # we need to pass in the current user for validation on the formset
+        # this could alternatively be done on the page form itself if we added the
+        # comments formset there, but we typically only add fields via edit handlers
+        current_user = getattr(self.request, 'user', None)
```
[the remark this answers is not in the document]
jacobtoppm: Not really, but before the edit handler's bound to the request, `self.request` can be `None`.

client/src/components/CommentApp/comments.js
jacobtoppm: Yeah, this was an effort to standardise all usages between this and the FE PR. Went with 'contentpath' as the model's original version, and changed everything else to match that.

wagtail/admin/edit_handlers.py
jacobtoppm: The first one I think won't work because `self.comments` can include unsaved edited comments from the form (not the db), so fetching the comments from the db isn't guaranteed to be the same, but you're right that I could optimise more here! I'll have an experiment with this (I did originally have some prefetching, I'll see if I can bring that back in)

wagtail/admin/edit_handlers.py
```diff
@@ -789,8 +792,89 @@ def render(self):
...
+
+    template = "wagtailadmin/edit_handlers/comments/comment_panel.html"
+    js_template = "wagtailadmin/edit_handlers/comments/comment_panel.js"
+    declarations_template = "wagtailadmin/edit_handlers/comments/comment_declarations.html"
+
+    def html_declarations(self):
+        return render_to_string(self.declarations_template)
+
+    def on_form_bound(self):
+        self.comments = []
```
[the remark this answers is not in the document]
jacobtoppm: I've changed this to build the list first, then set `self.comments` to it instead to make this clearer. Basically, it's just in case `on_form_bound` is called multiple times.

wagtail/admin/forms/comments.py
```diff
@@ -0,0 +1,47 @@
...
+        user = self.user
+
+        if not self.instance.pk:
+            self.instance.user = user
+        elif self.instance.user != user:
+            # trying to edit someone else's comment reply
+            if any(field for field in self.changed_data):
+                # includes DELETION_FIELD_NAME, as users cannot delete each other's individual comment replies
+                # if deleting a whole thread, this should be done by deleting the parent Comment instead
+                self.add_error(None, error=ValidationError(_("You cannot edit another user's comment")))
```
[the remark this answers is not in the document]
jacobtoppm: Thanks, fixed.

wagtail/admin/forms/comments.py
jacobtoppm: Thanks! Fixed

client/src/components/CommentApp/comments.js
```diff
@@ -13,10 +13,11 @@ function initComments(initialComments) {
   };
   document.addEventListener('DOMContentLoaded', () => {
     const commentsElement = document.getElementById('comments');
-    const author = JSON.parse(document.getElementById('comments-author').textContent);
+    const commentsOutputElement = document.getElementById('comments-output');
+    const data = JSON.parse(document.getElementById('comments-data').textContent);
```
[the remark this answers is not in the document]
jacobtoppm: I think it's okay to throw an error and let it crash respectively, I've added this in the latest commit. It's in a separate bundle and the dummy commentApp is already in place, so all other js code should function fine (I've tested with the chooser comment widgets).

wagtail/admin/edit_handlers.py
jacobtoppm: I've added your second suggestion. Unfortunately `prefetch_related_objects` didn't seem to work for this.

wagtail/core/models.py
```diff
@@ -4754,7 +4740,7 @@ def __str__(self):
 
 
 class CommentReply(models.Model):
-    comment = models.ForeignKey(Comment, on_delete=models.CASCADE, related_name='replies')
+    comment = ParentalKey(Comment, on_delete=models.CASCADE, related_name='replies')
```
kaedroho, line 4743: Django model cluster doesn't actually fully support tested parental keys. I think one place this might be an issue is with page copying, it'll copy the comments but not the replies. Could you check that?

kaedroho (approved):
LGTM, just one comment about whether the comment replies would be copied when the page is copied, but maybe we should delete all comments on page copy?

wagtail/core/models.py
jacobtoppm: Yep, as we discussed, added a ticket in to delete all comments on copy in the future.

PR #6559, opened 2020-11-18
Fix missing locale in the project template's homepage migration

zerolab:
Fixes #6557

Added "upgrade considerations", which may be a bit on the controversial side so input is more than welcome

PR #6620, opened 2020-12-08
Fix create_homepage migration running out of order

frmdstryr:
For new projects the template should ensure that the create_homepage migration is run before the locale field is made required. Currently it doesn't give any requirement for this so if for whatever reason django decides to run those later it can lead to an integrity error:

```python-traceback  
  Applying wagtailcore.0057_page_locale_fields_notnull... OK
  Applying wagtailcore.0058_page_alias_of... OK
  Applying wagtailcore.0059_apply_collection_ordering... OK
  Applying home.0001_initial... OK
  Applying home.0002_create_homepage...Traceback (most recent call last)
  File "venv/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
  File "venv/lib/python3.8/site-packages/django/db/backends/sqlite3/base.py", line 413, in execute
    return Database.Cursor.execute(self, query, params)
sqlite3.IntegrityError: NOT NULL constraint failed: wagtailcore_page.locale_id
```

After

```
  Applying wagtailcore.0055_page_locale_fields... OK
  Applying home.0001_initial... OK
  Applying home.0002_create_homepage... OK
  Applying wagtailcore.0056_page_locale_fields_populate... OK

```

### DIFF
--- a/wagtail/project_template/home/migrations/0002_create_homepage.py
+++ b/wagtail/project_template/home/migrations/0002_create_homepage.py
@@ -52,6 +52,10 @@ class Migration(migrations.Migration):
     dependencies = [
         ('home', '0001_initial'),
     ]
+    
+    run_before = [
+        ('wagtailcore', '0056_page_locale_fields_populate'),
+    ]
 
     operations = [
         migrations.RunPython(create_homepage, remove_homepage),


gasman:
Thanks @frmdstryr! This is a known issue, being handled at #6559.

frmdstryr:
Oops, missed that. Sorry!

PR #6633, opened 2020-12-12
Add Javascript API for page chooser widget

gasman:
Some non-destructive (and hopefully safe to merge into master now) refactoring to AdminPageChooser to make its client-side functionality more hackable to client-side code (including Telepath) - createPageChooser now returns an object with accessors for populating the chooser state, retrieving its value, and triggering the modal. (This object will also be used as the Telepath client-side representation of the widget, in the follow-up PR I'm about to open.) Since constructing a state to pass to `chooser.setState` involves some non-trivial logic (such as retrieving the parent page and constructing an edit-view URL), I've extracted a `get_value_data` method from `render_html` and `render_js_init` so that we can re-use that logic.

(I've also done a small part of the latter refactoring on the other choosers - image, document, snippet and task - but didn't get time to finish the job off for those...)

### DIFF
--- a/client/src/entrypoints/page-chooser.js
+++ b/client/src/entrypoints/page-chooser.js
@@ -1,5 +1,5 @@
 /* eslint-disable */
-function createPageChooser(id, pageTypes, openAtParentId, canChooseRoot, userPerms) {
+function createPageChooser(id, openAtParentId, config) {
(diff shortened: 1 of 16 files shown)


kaedroho (approved):
Looks good to me!

I noticed you mentioned that this API is public, so might want to have some docs for this at some point if that's the case!

PR #6634, opened 2020-12-12
Implement PageChooserBlock for telepath

gasman:
Define a telepath adapter for AdminPageChooser and hook up the value conversion to make PageChooserBlock work in telepath. Built on top of #6633 which does most of the heavy lifting - only the last commit is new.


client/src/entrypoints/telepath/widgets.js
```diff
@@ -54,3 +54,21 @@ class RadioSelect extends Widget {
...
+        this.html = html;
+        this.idForLabel = idForLabel;
+        this.config = config;
+    }
+
+    render(placeholder, name, id) {
+        var html = this.html.replace(/__NAME__/g, name).replace(/__ID__/g, id);
+        var dom = $(html);
+        $(placeholder).replaceWith(dom);
+        /* the chooser object returned by createPageChooser also serves as the JS widget representation */
```
kaedroho, line 70: Nice!

wagtail/admin/widgets/chooser.py
```diff
@@ -166,3 +167,23 @@ def media(self):
...
+    def js_args(self, widget, context):
+        return [
+            widget.render_html('__NAME__', None, attrs={'id': '__ID__'}),
+            widget.id_for_label('__ID__'),
+            widget.client_config,
+        ]
+
+    @property
+    def media(self):
+        return AdminPageChooser().media + forms.Media(js=[
```
kaedroho, line 184: Is the order here significant? If `AdminPageChooser` depends on `widgets.js`, should `widgets.js` be added first so it's loaded before `AdminPageChooser`'s js?

wagtail/core/blocks/field_block.py
```diff
@@ -721,6 +721,18 @@ def widget(self):
         return AdminPageChooser(target_models=self.target_models,
                                 can_choose_root=self.can_choose_root)
 
+    def get_form_state(self, value):
+        value_data = self.widget.get_value_data(value)
+        if value_data is None:
+            return None
+        else:
+            return {
+                'id': value_data['id'],
```
kaedroho, line 730: For future: We probably should make this representation to match the Admin API so the same code could work with both

kaedroho (approved):
Looks good to me, just a few suggestions for the API

client/src/entrypoints/page-chooser.js
```diff
@@ -1,44 +1,85 @@
...
         input.val('');
-        openAtParentId = null;
         chooserElement.addClass('blank');
-    });
+      }
+
+      state = newState;
+    },
+
+    openChooserModal: () => {
```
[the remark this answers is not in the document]
gasman: As I see it, chooser widgets and chooser modals are two distinct APIs (the rich text editor also uses the latter independently, and wagtail-generic-chooser treats them as separate concepts), and I made a point of leaving the modal API unchanged here. Things that are not using the form widgets should continue to call ModalWorkflow themselves - the presence of the openChooserModal method here doesn't change that.

I agree with your overall point though - we really could do with a higher-level API for modals rather than an ugly boilerplate-y ModalWorkflow() call. I think a good way of doing that would be to turn ModalWorkflow into a base class, and have page-chooser-modal.js etc export a subclass of it (rather than the dict of onload callbacks to pass to ModalWorkflow that they currently 'export'). That'll be an easier task now that we're moving things into /client, and as a bonus it'll stop eslint complaining about ModalWorkflow being capitalised as a class despite not being one...

I like the idea of making it promise-based! In theory ModalWorkflow supports multiple response callbacks, which wouldn't work with promises, but I can't see anything making use of that in practice.

wagtail/admin/widgets/chooser.py
gasman: I'd say the dependency is the other way round (i.e. the current ordering is correct): AdminPageChooser's media contains the base implementation of the widget that comes into play regardless of whether we instantiated it through telepath or oldschool server-side rendering (with an inline `<script>createPageChooser(...)</script>` call), and telepath/widgets.js is the wrapper around it to make it interoperate with telepath.

(It just happens that the internal `chooser` object from createPageChooser is exactly what telepath needs, and so that wrapper is more or less transparent.)

client/src/entrypoints/page-chooser.js
```diff
@@ -1,44 +1,85 @@
...
-    var pageTitle = chooserElement.find('.title');
-    var input = $('#' + id);
-    var editLink = chooserElement.find('.edit-link');
-
-    $('.action-choose', chooserElement).on('click', function() {
-        var initialUrl = chooserElement.data('chooserUrl');
-        if (openAtParentId) {
-            initialUrl += openAtParentId + '/';
-        }
+function createPageChooser(id, openAtParentId, config) {
```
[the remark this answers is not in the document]
gasman: Makes sense - updated in 668a962c2d89ce9e5ce9380392fc48a2981d980a.

gasman:
Merged into feature/telepath in a2fe224d9c389f73c679bb178d998d9acc08a106

PR #6635, opened 2020-12-12
Allow long embed urls, add unique hash

allcaps:
This PR removes the max_length constraint from embed url fields: `Embed.url` and `Embed.thumbnail_url` now accept any length of string.

Fixes: #6467 and #1985

Idea @ababic:
https://github.com/wagtail/wagtail/blob/master/wagtail/embeds/embeds.py#issuecomment-720111428

Note:
- Embed url+max_width uniqueness is now guaranteed by `Embed.hash`
- The hash field has a db index

Question:
- The migration combines structure, data and structure again in a single file. Is the right thing to do? Or should this migration be split into 3 files?

Checks
* Do the tests still pass? (https://docs.wagtail.io/en/latest/contributing/developing.html#testing)  YES

### DIFF
--- a/wagtail/embeds/embeds.py
+++ b/wagtail/embeds/embeds.py
@@ -1,4 +1,5 @@
 from datetime import datetime
+from hashlib import md5
 
 from .exceptions import EmbedUnsupportedProviderException
 from .finders import get_finders
@@ -6,14 +7,17 @@ from .models import Embed
 
 
 def get_embed(url, max_width=None, finder=None):
+    embed_hash = get_embed_hash(url, max_width)
+
     # Check database
     try:
-        return Embed.objects.get(url=url, max_width=max_width)
+        return Embed.objects.get(hash=embed_hash)
     except Embed.DoesNotExist:
         pass
 
     # Get/Call finder
     if not finder:
+
         def finder(url, max_width=None):
             for finder in get_finders():
                 if finder.accept(url):
@@ -40,9 +44,12 @@ def get_embed(url, max_width=None, finder=None):
 
     # Create database record
     embed, created = Embed.objects.get_or_create(
-        url=url,
-        max_width=max_width,
-        defaults=embed_dict,
+        hash=embed_hash,
+        defaults=dict(
+            url=url,
+            max_width=max_width,
+            **embed_dict,
+        )
     )
 
     # Save
@@ -50,3 +57,11 @@ def get_embed(url, max_width=None, finder=None):
     embed.save()
 
     return embed
+
+
+def get_embed_hash(url, max_width=None):
+    h = md5()
+    h.update(url.encode("utf-8"))
+    if max_width is not None:
+        h.update(str(max_width).encode("utf-8"))
+    return h.hexdigest()
--- a/wagtail/embeds/migrations/0006_add_embed_hash.py
+++ b/wagtail/embeds/migrations/0006_add_embed_hash.py
@@ -1,33 +1,5 @@
 from django.db import migrations, models
 
-from wagtail.embeds.embeds import get_embed_hash
-
-
-def migrate_forwards(apps, schema_editor):
-    Embed = apps.get_model("wagtailembeds.Embed")
-
-    # based each embed being ~500b, available memory being
-    # to 2MB, and leaving plenty of headway for app itself
-    batch_size = 1500
-
-    # allows small batches to be kept in memory in order
-    # to utilise bulk_update()
-    batch = []
-
-    for embed in Embed.objects.all().only("id", "url", "max_width").iterator():
-
-        embed.hash = get_embed_hash(embed.url, embed.max_width)
-        batch.append(embed)
-
-        if len(batch) == batch_size:
-            # save and reset current batch
-            Embed.objects.bulk_update(batch, ["hash"])
-            batch.clear()
-
-    # save any leftovers
-    if batch:
-        Embed.objects.bulk_update(batch, ["hash"])
-
 
 class Migration(migrations.Migration):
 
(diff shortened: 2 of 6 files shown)


allcaps:
TODO:
`django.db.utils.OperationalError: (1170, "BLOB/TEXT column 'url' used in key specification without a key length")`

Change migration order. `Char => Text`, must come later

wagtail/embeds/models.py
```diff
@@ -22,22 +23,23 @@ class Embed(models.Model):
     If an instance of this model is deleted, it will be automatically refetched
     next time the embed code is needed.
     """
-    url = models.URLField()
+
+    url = models.TextField(validators=[URLValidator()])
```
ababic, line 27: I know this sounds counterintuitive, but we actually don't ever invoke field or model-level validation for embeds, so we can safely drop the validators (for both fields)

wagtail/embeds/tests/test_embeds.py
```diff
@@ -134,6 +134,12 @@ def dummy_finder(self, url, max_width=None):
             'html': "<p>Blah blah blah</p>",
         }
 
+    def test_get_embed_hash(self):
+        url = "www.test.com/1234"
+        self.assertEqual(get_embed_hash(url), "9a4cfc187266026cd68160b5db572629")
+        self.assertEqual(get_embed_hash(url, 0), "be54b69772d5e086ec07748741455736")
+        self.assertEqual(get_embed_hash(url, 1), "601c15cca60eb068ebcc166b9587ed63")
+
```
ababic, line 142: I feel this test might benefit from its own `TestEmbedHash` class

wagtail/embeds/migrations/0006_add_embed_hash.py
```diff
@@ -0,0 +1,72 @@
...
+
+from wagtail.embeds.embeds import get_embed_hash
+
+
+def migrate_forwards(apps, schema_editor):
+    Embed = apps.get_model("wagtailembeds.Embed")
+
+    # based each page being ~500b, available memory being
+    # to 2MB, and leaving plenty of headway for app itself
+    batch_size = 1500
```
ababic, line 13: based each page -> based on each embed

ababic (commented):
Thanks @allcaps! Just a couple of minor nitpicks, but otherwise looking great 👍

allcaps:
@ababic Thank you for the review. I applied your feedback and squashed the commits.

wagtail/embeds/migrations/0006_add_embed_hash.py
```diff
@@ -0,0 +1,70 @@
...
+        migrations.AlterField(
+            model_name="embed",
+            name="thumbnail_url",
+            field=models.TextField(
+                blank=True,
+                default="",
+            ),
+            preserve_default=False,
+        ),
+    ]
```
ababic, line 70: Has this migration been run in reverse? I have a feeling that might throw up some issues. Django does recommend [not including schema changes and RunPython actions in the same migration](https://django.readthedocs.io/en/latest/ref/migration-operations.html#runpython), so may be safer to err on the side of cautiom here. Excerpt below:

> On databases that do support DDL transactions (SQLite and PostgreSQL), RunPython operations do not have any transactions automatically added besides the transactions created for each migration. Thus, on PostgreSQL, for example, you should avoid combining schema changes and RunPython operations in the same migration or you may hit errors like OperationalError: cannot ALTER TABLE "mytable" because it has pending trigger events.

ababic (changes requested):

wagtail/embeds/models.py
```diff
@@ -22,22 +22,23 @@ class Embed(models.Model):
...
+    url = models.TextField()
     max_width = models.SmallIntegerField(null=True, blank=True)
+    hash = models.CharField(max_length=32, unique=True, db_index=True)
     type = models.CharField(max_length=10, choices=EMBED_TYPES)
     html = models.TextField(blank=True)
     title = models.TextField(blank=True)
     author_name = models.TextField(blank=True)
     provider_name = models.TextField(blank=True)
-    thumbnail_url = models.URLField(max_length=255, null=True, blank=True)
+    thumbnail_url = models.TextField(blank=True)
```
ababic, line 34: @allcaps Just noticed we have lost the `null=True` here. This makes sense from a consistency point of view, but we'll want to check whether the ORM is checking for null values anywhere before merging.

ababic:
Just one last thing to check, and I think we'll be good to go. Thank you very much for your time and patience with this one @allcaps!

wagtail/embeds/models.py
allcaps: The Django docs mentions not add `null=True` to string based fields:

> Avoid using null on string-based fields such as CharField and TextField. If a string-based field has null=True, that means it has two possible values for “no data”: NULL, and the empty string. In most cases, it’s redundant to have two possible values for “no data;” the Django convention is to use the empty string, not NULL. One exception is when a CharField has both unique=True and blank=True set. In this situation, null=True is required to avoid unique constraint violations when saving multiple objects with blank values.
https://docs.djangoproject.com/en/3.1/ref/models/fields/#null

Adding `null=True` on string based fields is a common mistake. I can't see any good reason because `thumbnail_url` does not have unique constraints.

Our migration already has `default=""`, AFAIK any `Null` values will be converted to empty strings.

Is searched `wagtail/wagtail/embeds` module for `isnull`. Zero matches.
I also searched for `None`, 72 matches, as far as I can see, not related to `thumbnail_url`.
@ababic Will you also search the code whether the ORM is checking for null values?

ababic: @allcaps yep, I'll have a quick look. I imagine all will be fine! I'm totally with you on the reasoning behind the change - it just usually pays to be thorough where field changes are concerned.

ababic: @allcaps LGTM :)

ababic (approved):
LGTM

wagtail/embeds/embeds.py
```diff
@@ -40,13 +44,24 @@ def finder(url, max_width=None):
...
     embed.save()
 
     return embed
+
+
+def get_embed_hash(url, max_width=None):
+    h = md5()
+    h.update(url.encode("utf-8"))
+    if max_width is not None:
+        h.update(str(max_width).encode("utf-8"))
```
gasman, line 66: Super-paranoid nitpick: putting the two `update`s together means that `get_embed_hash('https://example.com/embed', 400)` and `get_embed_hash('https://example.com/embed400')` return the same result. Easy enough to fix - I'll just add a `\n` delimiter between them...

gasman (approved):
Looks good to me, thanks @allcaps! Will fix my super-pedantic nitpick about the hash function when I merge...

gasman:
Merged in c0c7f1bd5deca217776e6f264200d37a09f195e4

pySilver:
Great! Thank you!

pySilver:
Looks like another bug has been introduced #6788

PR #6639, opened 2020-12-14
Special case date lookups for searching

### LINKED ISSUE #5967: 'ExtractYear' object has no attribute 'target'
### Issue Summary
Calling `search` queryset method on querysets that are filtered by exact year of a date throws AttributeError. This happens not only with `__year` but other exact matches also, e.g `__month`, `__day`. 

### Steps to Reproduce
1. `wagtail start wagtail-test`
2. `python manage shell`
3. `from wagtail.core.models import Page`
4. `Page.objects.filter(last_published_at__year=2020).search('hello')`

Here is the traceback: 

```
  File "<console>", line 1, in <module>
  File "/Users/gadirrustamli/anaconda/envs/wagtail-test/lib/python3.8/site-packages/wagtail/search/queryset.py", line 11, in search
    return search_backend.search(query, self, fields=fields,
  File "/Users/gadirrustamli/anaconda/envs/wagtail-test/lib/python3.8/site-packages/wagtail/search/backends/base.py", line 363, in search
    return self._search(
  File "/Users/gadirrustamli/anaconda/envs/wagtail-test/lib/python3.8/site-packages/wagtail/search/backends/base.py", line 358, in _search
    search_query.check()
  File "/Users/gadirrustamli/anaconda/envs/wagtail-test/lib/python3.8/site-packages/wagtail/search/backends/base.py", line 157, in check
    self._get_filters_from_where_node(self.queryset.query.where, check_only=True)
  File "/Users/gadirrustamli/anaconda/envs/wagtail-test/lib/python3.8/site-packages/wagtail/search/backends/base.py", line 108, in _get_filters_from_where_node
    child_filters = [self._get_filters_from_where_node(child) for child in where_node.children]
  File "/Users/gadirrustamli/anaconda/envs/wagtail-test/lib/python3.8/site-packages/wagtail/search/backends/base.py", line 108, in <listcomp>
    child_filters = [self._get_filters_from_where_node(child) for child in where_node.children]
  File "/Users/gadirrustamli/anaconda/envs/wagtail-test/lib/python3.8/site-packages/wagtail/search/backends/base.py", line 91, in _get_filters_from_where_node
    field_attname = where_node.lhs.target.attname
AttributeError: 'ExtractYear' object has no attribute 'target'

```

### Technical details

* Python version: 3.8
* Django version: 2.2.8
* Wagtail version: 2.8.1

seb-b:
Fixes #5967. I didn't delve too deep into why date lookups have nested `lhs` clauses, but this seems to fix the problem.

Yes
Yes
Yes
N/A
N/A

### DIFF
--- a/wagtail/search/backends/base.py
+++ b/wagtail/search/backends/base.py
@@ -1,6 +1,6 @@
 from warnings import warn
 
-from django.db.models.lookups import Lookup
+from django.db.models.lookups import Lookup, YearLookup
 from django.db.models.query import QuerySet
 from django.db.models.sql.where import SubqueryConstraint, WhereNode
 
@@ -88,7 +88,10 @@ class BaseSearchQueryCompiler:
     def _get_filters_from_where_node(self, where_node, check_only=False):
         # Check if this is a leaf node
         if isinstance(where_node, Lookup):
-            field_attname = where_node.lhs.target.attname
+            if isinstance(where_node, YearLookup):
+                field_attname = where_node.lhs.lhs.target.attname
+            else:
+                field_attname = where_node.lhs.target.attname
             lookup = where_node.lookup_name
             value = where_node.rhs
 
--- a/wagtail/search/backends/base.py
+++ b/wagtail/search/backends/base.py
@@ -1,6 +1,8 @@
 from warnings import warn
 
-from django.db.models.lookups import Lookup, YearLookup
+from django.db.models.functions.datetime import Extract as ExtractDate
+from django.db.models.functions.datetime import ExtractYear
+from django.db.models.lookups import Lookup
 from django.db.models.query import QuerySet
 from django.db.models.sql.where import SubqueryConstraint, WhereNode
 
@@ -88,8 +90,14 @@ class BaseSearchQueryCompiler:
     def _get_filters_from_where_node(self, where_node, check_only=False):
         # Check if this is a leaf node
         if isinstance(where_node, Lookup):
-            if isinstance(where_node, YearLookup):
-                field_attname = where_node.lhs.lhs.target.attname
+            if isinstance(where_node.lhs, ExtractDate):
+                if isinstance(where_node.lhs, ExtractYear):
+                    field_attname = where_node.lhs.lhs.target.attname
+                else:
+                    raise FilterError(
+                        'Cannot apply filter on search results: "' + where_node.lhs.lookup_name + 
+                        '" queries are not supported.'
+                    )
             else:
                 field_attname = where_node.lhs.target.attname
             lookup = where_node.lookup_name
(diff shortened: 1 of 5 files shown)


wagtail/search/tests/elasticsearch_common_tests.py
```diff
@@ -176,6 +176,12 @@ def test_slice_to_next_page(self):
         results = self.backend.search(MATCH_ALL, models.Book)[110:]
         self.assertEqual(len(results), 54)
 
+    def test_search_with_date_filter(self):
+        after_1900 = models.Book.objects.filter(publication_date__year__gt=1900)
```
kaedroho, line 180: Could you add tests for months/days as well?

wagtail/search/backends/base.py
```diff
@@ -88,7 +88,10 @@ def _process_filter(self, field_attname, lookup, value, check_only=False):
     def _get_filters_from_where_node(self, where_node, check_only=False):
         # Check if this is a leaf node
         if isinstance(where_node, Lookup):
-            field_attname = where_node.lhs.target.attname
+            if isinstance(where_node, YearLookup):
```
kaedroho, line 91: Surprised this works without any further changes! I'm guessing it's a special kind of range query?

We should add tests for the query builder as well: https://github.com/wagtail/wagtail/blob/master/wagtail/search/tests/test_elasticsearch7_backend.py#L22

kaedroho (changes requested):
Thanks for submitting this!

I think we'd want to add support for year/month/day at the same time, since it might be confusing for developers to only partially support date filters. Also, I think we need tests for the query builder so we can check exactly what it's converting these filters into.

wagtail/search/tests/elasticsearch_common_tests.py
seb-b: Turns out this doesn't work for months/days, I'm going to keep looking into it

seb-b:
So I think the problem here is that elastic can't do a query like "after the 15th of every month" (e,g, `Querset.filter(day__gt=15)`). If I'm reading [this bit of the docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-range-query.html#missing-date-components) right sending something like ` gt: 15||//d` (15th day) translates to `after the the 15th of Jan`. 

So this fix will work for years, since es will round down or up based on the [query](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-range-query.html#range-query-date-math-rounding), but not won't work for wildcard type queries for days and months. Maybe the best we can do is make it work for years and throw an error when a queryset is filtered by day or month? Or maybe someone who is better at constructing elastic queries can see how to make it work

kaedroho:
Thanks for looking into this!

> Maybe the best we can do is make it work for years and throw an error when a queryset is filtered by only day or month?

This sounds fine to me! I'm not sure if having the ability to filter by a day of the month in all months is useful in many cases.

seb-b:
Thanks for you feedback @kaedroho, I've added some extra tests and also some exception throwing on bad queries.

kaedroho (approved):
Looks good to me, thanks!

PR #6645, opened 2020-12-17
fix ModelAdmin example in docs

Bohreromir:
There is small typo that prevents one from using the example code without modifying it.
I tried to fix it. Its my first contribution here, let me know if i made a mistake


gasman:
Thanks @Bohreromir! Merged in 3d18986610c147299e3cf497efdfb948ae336da0.

Issue #6107, opened 2020-06-03
Phase out icon font and use SVG everywhere


allcaps:
In https://github.com/wagtail/wagtail/pull/4821 the menu font icons are replaced with SVG icons. It is a nice start. In this issue an overview of work to be done to have SVG icons everywhere.

We choose to switch to SVG icons for accessibility (title attr and aria labels). But SVG icons are also easier to maintain, customise and extend. All icons are all-ready converted to SVG. We need to dorp the icon font reference (CSS) and use the SVG icon like this:

    {% load wagtailadmin_tags %}
    {% icon name="rocket" classname="..." title="Launch" %}

The intention is to drop Wagtail font icon and CSS from future Wagtail releases. We need to decide what is needed for a proper deprecation process.

I'd propose to keep this issue as a meta issue to keep track of progress. Work can be done in separate issues and pull requests. To keep pull requests reviewable.

## Priority

 - [x] Images in rich text have oversized image icon (SVG regression) #6142
 - [x] Menu icon positioning https://github.com/wagtail/wagtail/issues/6297 Will be solved by switching to SVG.

## Icon font usage that needs to be refactored

- [x] Header icons ``<h1 class="icon icon-image">Images</h1>`` https://github.com/wagtail/wagtail/pull/6121
- [x] Messages framework messages ``messages success`` https://github.com/wagtail/wagtail/pull/6124
- [x] Button plus ``button bicolor icon icon-plus`` #6151
- [x] Dropdown buttons ``dropdown-toggle icon icon-arrow-down``  #6151 
- [x] More buttons ``o-icon  c-dropdown__toggle  [ icon icon-arrow-down``  #6151 
- [x] Wagtail user bar and menu items `wagtail-icon wagtail-icon-wagtail` https://github.com/wagtail/wagtail/pull/6125
- [x] `MenuItem`s (including `ModelAdminMenuItem` and `GroupMenuItem`) #6402 
- [x] Help text ``help-block help-critical`` @Scotchester #6457 
- [x] Breadcrumb (home icon) #6438 @allcaps
- [x] Pagination, previous and next chevrons @Scotchester #6573
- [x] SearchAreas API @thibaudcolas #6493
- [x] Settings menu item @thibaudcolas #6649
- [x] All choosers ``.page-chooser .unchosen:before`` @Stormheg 
- [x] StreamField buttons #7565
- [x] Spinner @stevedya 
[...]

PR #6649, opened 2020-12-18
Switch register_setting, register_settings_menu_item to use SVG icons (#6107)

thibaudcolas:
Switches `register_setting` and `SettingMenuItem`’s `icon` attribute to use SVG icons, as opposed to font icons. Part of #6107.

This is a breaking change for end users who set custom icons that aren’t part of Wagtail’s default icon set. Users who want to keep using custom icon fonts will need to update their code from `icon='my-icon-name'` to `classnames='icon icon-my-icon-name'`, essentially matching what the menu item used to generate.

To minimise the breakage, I’ve added a special check for Font Awesome (`fa-`) icons, which still will work as-is without any further changes. This should eventually be phased out when changes from #6107 are closer to completion.

Here are notes I would recommend adding to the release’s Upgrade considerations when this is merged:

```rst
Switched ``SettingMenuItem`` and ``register_setting`` to use SVG icons
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Setting menu items now use SVG icons by default. For sites reusing built-in Wagtail icons, no changes should be required. For sites using custom icons, update the menu items’ definition to use the ``classnames`` attribute:

.. code-block:: python

    # With register_setting,
    # Before:
    @register_setting(icon='custom-cog')
    # After:
    @register_setting(icon='', classnames='icon icon-custom-cog')

    # Or with SettingMenuItem,
    @hooks.register('register_settings_menu_item')
    def register_frank_menu_item():
        # Before:
        return SettingMenuItem(CustomSetting, icon='custom-cog')
        # After:
        return SettingMenuItem(CustomSetting, icon='', classnames='icon icon-custom-cog')
```

---

* ~For front-end changes: Did you test on all of Wagtail’s supported browsers? **Please list the exact versions you tested**.~

---

Here is code to add to `wagtail_hooks.py` to test all variants locally. Make sure to also add `wagtail.contrib.settings` to the `INSTALLED_APPS`.

```py
from django.db import models

from wagtail.admin.menu import MenuItem

from wagtail.contrib.settings.registry import SettingMenuItem
from wagtail.contrib.settings.models import BaseSetting, register_setting


@register_setting(icon="tag")
class IconSetting(BaseSetting):
    title = models.CharField(max_length=100)


@register_setting(icon="fa-suitcase")
class SuitcaseSetting(BaseSetting):
    title = models.CharField(max_length=100)


class IconUnregisteredSetting(BaseSetting):
    title = models.CharField(max_length=100)


@hooks.register('register_settings_menu_item')
def register_frank_menu_item():
    return SettingMenuItem(IconUnregisteredSetting, icon='fa-cutlery', classnames='test-class')


@hooks.register('register_settings_menu_item')
def register_herbert_menu_item():
    return SettingMenuItem(IconUnregisteredSetting, icon='tag', classnames='test-class')
```

This includes both icons defined via the `register_setting` decorator and via the `register_settings_menu_item` with the `SettingMenuItem` class, although really following through the code these end up being processed the exact same way.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -12,6 +12,7 @@ Changelog
  * Update `PageQueryset.specific(defer=True)` to only perform a single database query (Andy Babic)
  * Add `PageQueryset.defer_streamfields()` (Andy Babic)
  * Utilize `PageQuerySet.defer_streamfields()` to improve efficiency in a few key places (Andy Babic)
+ * Switch ``register_setting``, ``register_settings_menu_item`` to use SVG icons (Thibaud Colas)
  * Fix: StreamField required status is now consistently handled by the `blank` keyword argument (Matt Westcott)
  * Fix: Show 'required' asterisks for blocks inside required StreamFields (Matt Westcott)
  * Fix: Make image chooser "Select format" fields translatable (Helen Chapman, Thibaud Colas)
--- a/docs/releases/2.13.rst
+++ b/docs/releases/2.13.rst
@@ -21,7 +21,7 @@ Other features
 * Support passing multiple models as arguments to ``type()``, ``not_type()``, ``exact_type()`` and ``not_exact_type()`` methods on ``PageQuerySet`` (Andy Babic)
 * Update default attribute copying behaviour of ``Page.get_specific()`` and added the ``copy_attrs_exclude`` option (Andy Babic)
 * Update ``PageQueryset.specific(defer=True)`` to only perform a single database query (Andy Babic)
-
+* Switched ``register_setting``, ``register_settings_menu_item`` to use SVG icons (Thibaud Colas)
 
 Bug fixes
 ~~~~~~~~~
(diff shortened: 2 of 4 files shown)


Stormheg (approved):
Thanks @thibaudcolas! Icons and the special font awesome check work as expected :slightly_smiling_face: 

I like that there is some backwards compatibility in place for e.g. [wagtailfontawesome](https://gitlab.com/alexgleason/wagtailfontawesome) users :ok_hand: 

Would it be a good idea to create an issue on removing the font awesome check you're introducing? That way we can keep track of it and won't forget to remove it when #6107 is completed.

---

I've tested this change on the following browsers
- Firefox 86
- Chrome 89
- Safari on iPadOS 14.4 (iPad 6th gen)

I don't have access to BrowserStack so unfortunately I'm not able to test all browsers listed in [Wagtail's browser and device support](https://docs.wagtail.io/en/stable/contributing/developing.html#browser-and-device-support).

thibaudcolas:
Thank you @Stormheg, I added a note directly in #6107 based on your suggestion.

PR #6651, opened 2020-12-18
Refactor draftail ModalWorkflowSource for extensibility

gasman:
Currently, the important work in draftail's ModalWorkflowSource.js is handled by two functions `getChooserConfig` and `filterEntityData` which consist of a switch/case over the four built-in entity types: images, embeds, links and documents. This makes it impossible to extend draftail with new modal-powered entity types without editing (or duplicating) this file. To address this, I've refactored ModalWorkflowSource into an abstract class which is subclassed for each entity type.

~Along the way I've relaxed the no-unused-vars eslint rule so that it doesn't complain about unused function parameters - I think this is needlessly restrictive, since a case like this (where multiple classes are implementing the same interface) is a valid place where unused parameters can arise.~ Now rebased to use `"args": "after-used"` and eslint-ignore as per Slack discussion...

### DIFF
--- a/client/src/components/Draftail/sources/ModalWorkflowSource.js
+++ b/client/src/components/Draftail/sources/ModalWorkflowSource.js
@@ -17,138 +17,6 @@ MUTABILITY[DOCUMENT] = 'MUTABLE';
 MUTABILITY[ENTITY_TYPE.IMAGE] = 'IMMUTABLE';
 MUTABILITY[EMBED] = 'IMMUTABLE';
 
-export const getChooserConfig = (entityType, entity, selectedText) => {
-  let url;
-  let urlParams;
-
-  switch (entityType.type) {
-  case ENTITY_TYPE.IMAGE:
-    if (entity) {
-      const data = entity.getData();
-      url = `${global.chooserUrls.imageChooser}${data.id}/select_format/`;
-      urlParams = {
-        format: data.format,
-        alt_text: data.alt,
-      };
-    } else {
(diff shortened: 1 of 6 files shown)


client/src/components/Draftail/sources/ModalWorkflowSource.js
```diff
@@ -161,10 +29,20 @@ class ModalWorkflowSource extends Component {
     this.onClose = this.onClose.bind(this);
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  getChooserConfig(entity, selectedText) {
```
jacobtoppm, line 33: Is it worth using Typescript abstract class for this, and making this an abstract method for instance? See https://www.typescriptlang.org/docs/handbook/2/classes.html#abstract-classes-and-members

jacobtoppm (approved):
Generally looks good to me, though I think an alternative approach would be to use Typescript abstract classes

client/src/components/Draftail/sources/ModalWorkflowSource.js
gasman: Sounds good - will give that a go!

gasman: Gave it a try, but `npm run test` says it isn't valid outside of a Typescript file, and if I rename the file to `ModalWorkflowSource.ts` I get failures elsewhere (`Property 'jQuery' does not exist on type 'Global & typeof globalThis'`). I'll leave that improvement for someone who knows Typescript :-)

PR #6652, opened 2020-12-19
Prevent remove_stale_contenttypes from removing the access_admin permission

### LINKED ISSUE #4444: Built-in command `remove_stale_contenttypes` removes `wagtail_admin` permissions

gasman:
Fixes #4444. Thanks to @berekuk for the pointer to how to fix this cleanly.


allcaps (approved):
Works as advertised. Approved!

PR #6653, opened 2020-12-20
Initial linting pass over all legacy JS

kaedroho:
Just fixing the most common errors like:

 - Indent with 2 spaces
 - Replace ``var`` with ``let``/``const``
 - Replace ``function()`` expressions with arrow functions
 - A few other trivial tweaks like spacing

I've made eslint ignore all non-trivial issues for now


client/src/entrypoints/core.js
```diff
@@ -51,296 +51,302 @@ window.initTagField = initTagField;
...
+    } else {
+      $('body').removeClass('nav-closed');
+    }
+  });
 
-    $(window).on('resize', function() {
-        fitNav();
-    });
+  // Enable toggle to open/close user settings
+  $(document).on('click', '#account-settings', () => {
```
gasman, line 107: We need to watch out for inline functions that refer to `this`, since the semantics are different for arrow functions - in this case, the arrow icon fails to switch direction as the panel is opened/closed.

gasman (changes requested):

client/src/entrypoints/core.js
kaedroho: All fixed now, I redid the linting from scratch but left the arrow function change entirely to ``eslint --fix`` this time and it appears to err on the safe side

client/src/entrypoints/core.js
```diff
@@ -369,220 +369,217 @@ $(() => {
...
-    },
-
-    add(dropDown) {
-      this.dropDowns.push(dropDown);
-    },
-
-    get() {
-      return this.dropDowns;
-    },
+const wagtail = window.wagtail || {};
```
gasman, line 372: Not really sure about this change - by removing the function wrapper, don't we end up dumping a load of constants into the global scope (including some with pretty generic names like `keys`)? Is there a chance that this could clash with other code elsewhere using those variables as non-constants?

(Happy with everything else in the PR though!)

kaedroho: It just moves it up to module scope, webpack encapsulates the modules so this shouldn't affect any other files (hence why we need to explicitly add stuff to the ``window`` object now when we want to create globals).

gasman (approved):
LGTM! In the process of testing I noticed that the userbar has stopped working since the js was moved into client (27133f957074635bf6efe461eff9f61d206d8443), so will merge now and open a new issue for that.

kaedroho:
Thanks! Will look into that

PR #6654, opened 2020-12-20
Support custom document models with required fields

gasman:
Same approach as 4023a90d6e0b2876c560f0318c1cf6aea544e354 for images - define an UploadedDocument model to hold the uploaded file up to the point where the required metadata has been supplied.

(Refactoring the image and document views to share the same base code will be the next job - it makes sense to get the feature sets of the two apps back in sync before tackling that.)


gasman:
Merged via #6658 in 758fa2b8631829cb52f422d2f6dda81c126e37c0.

PR #6658, opened 2020-12-23
Add shared class-based views for multiple image / document uploaders

gasman:
Based on #6654. Extract shared view code from the multiple image / document uploader views into a set of generic class-based views.

(The view modules can probably be simplified further with viewsets, and the template / front-end code could benefit from some de-duplication too, but this is enough for one PR...)


kaedroho (approved):
Looks good to me! Added a few suggestions but mainly stuff that could be looked at in another MR.

One minor issue that could arise is the new "uploaded_documents" folder it creates only when there is a validation error. It'll be tricky to test this when configuring a new site, in theory, someone could've set up their S3 credentials to only allow access to the usual images/documents folders.

gasman:
Merged in 758fa2b8631829cb52f422d2f6dda81c126e37c0

PR #6660, opened 2020-12-27
Update webpack to version 5 and fix userbar

### LINKED ISSUE #6657: Userbar / edit bird broken
### Issue Summary

The userbar no longer opens since the JS was moved to `/client` (160d66ca93fb071c71785897eddf4d638a3c6d6c / 27133f957074635bf6efe461eff9f61d206d8443).

### Steps to Reproduce

1. Install bakerydemo in a virtualenv as per https://github.com/wagtail/bakerydemo#setup-with-virtualenv
2. `pip uninstall wagtail` then install a build of current Wagtail master with `pip install -e .`
3. Log in to http://localhost:8000/admin/
4. Visit http://localhost:8000/ , click on the bird icon in the bottom right; menu fails to appear

### Technical details

* Python version: 3.7.2
* Django version: 3.1.4
* Wagtail version: master fe79fbb48f807da0f10d23548019ff8b496afcf8
* Browser version: Chrome 87.0.4280.88

kaedroho:
Fixes #6657 and updates Webpack to v5

The problem turned out to be the fact that we we're adding ``polyfills.js`` to the bundle and this imports stuff from ``node_modules``. Our webpack config bundles up anything that came from ``node_modules`` into a separate ``vendor.js`` bundle and it's all linked at runtime so  ``userbar.js`` wasn't being executed because ``vendor.js`` wasn't loaded and webpack was waiting for it.

Since it's written in a very portable way anyway, I've removed the polyfills from this bundle to remove this dependency and added a note at the top of ``userbar.py``.

The Webpack upgrade turned out to be unnecessary, but thought might as well submit it anyway so we don't have to do it again later.

### DIFF
--- a/client/src/entrypoints/userbar.js
+++ b/client/src/entrypoints/userbar.js
@@ -1,3 +1,7 @@
+// This entrypoint is not bundled with any polyfills to keep it as light as possible
+// Please stick to old JS APIs and avoid importing anything that might require a vendored module
+// More background can be found in webpack.config.js
+
 document.addEventListener('DOMContentLoaded', (e) => {
   const userbar = document.querySelector('[data-wagtail-userbar]');
   const trigger = userbar.querySelector('[data-wagtail-userbar-trigger]');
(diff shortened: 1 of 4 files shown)


gasman:
Looks good to me and works nicely! Rebased and merged in 5e54382aae48162152fccded7f56de19219d9899.

PR #6661, opened 2020-12-29
Add Page.specific_deferred

ababic:
Adds a `specific_deferred` property method to the `Page` model, allowing developers to efficiently access specific page instances without a database query to fetch specific model field values.

Originally implemented as part of #6216, but separated out to simplify review.

TODO:

- [x] Add tests to cover the new behaviour
- [x] Update documentation

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -18,7 +18,7 @@ from django.core.exceptions import PermissionDenied, ValidationError
 from django.core.handlers.base import BaseHandler
 from django.core.handlers.wsgi import WSGIRequest
 from django.db import migrations, models, transaction
-from django.db.models import Q, Value
+from django.db.models import DEFERRED, Q, Value
 from django.db.models.expressions import OuterRef, Subquery
 from django.db.models.functions import Concat, Lower, Substr
 from django.db.models.signals import pre_save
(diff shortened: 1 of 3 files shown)


wagtail/core/models.py
```diff
@@ -1189,10 +1202,52 @@ def specific(self):
...
+            )
+            # Create object from known attribute values
+            specific_obj = model_class(*values)
+            specific_obj._state.adding = self._state.adding
+        else:
+            # Fetch object from database
+            specific_obj = model_class._default_manager.get(id=self.id)
+
+        # Copy additional attribute values
+        if copy_attrs:
```
kaedroho, line 1225: Since we know all the valid attributes a generic ``Page`` instance should have, is there any reason we can't do this automatically?

wagtail/core/models.py
```diff
@@ -1189,10 +1202,52 @@ def specific(self):
...
+            for attr in copy_attrs:
+                if attr in self.__dict__:
+                    setattr(specific_obj, attr, getattr(self, attr))
+
+        return specific_obj
+
+    @cached_property
+    def specific(self):
+        """
+        A wrapper for ``self.get_specific()`` that returns this page in its
```
kaedroho, line 1235: Could we remove "A wrapper for ``self.get_specific()`` that" from these two methods? Feel like jumping straight into what it does makes it a bit easier to read.

kaedroho (commented):
Looks good to me, just a couple of minor questions. Thanks for implementing this!

wagtail/core/models.py
ababic: My only concern there would be copying accross calculated cached_property values that  may be calculated differently on the target class. Not such a big deal if going from `Page` to another type, but could be problematic in longer inheritance chains.

kaedroho (approved):

gasman:
Merged in f905e705da9ce94f93e2a7d0dc7328c9a3f32d81

ababic:
Wow, thanks @gasman!

PR #6665, opened 2021-01-04
Telepath: Support for standard and rich text blocks

kaedroho:
This PR implements some tweaks to allow standard and rich text blocks to render properly:

 - Implements ``autosize()`` for text input
 - Adds the logic to calculate the default classnames that are usually added to fields
 - Implements ``get_form_state`` for Rich Text Block
 - Widget media is now passed through (allowing draftail to work)

### DIFF
--- a/wagtail/core/blocks/field_block.py
+++ b/wagtail/core/blocks/field_block.py
@@ -60,6 +60,9 @@ class FieldBlock(Block):
         # a FieldBlock is required if and only if its underlying form field is required
         return self.field.required
 
+    def get_form_state(self, value):
+        return self.field.widget.format_value(self.value_for_form(value))
+
     class Meta:
         # No icon specified here, because that depends on the purpose that the
         # block is being used for. Feel encouraged to specify an icon in your
--- a/wagtail/core/blocks/field_block.py
+++ b/wagtail/core/blocks/field_block.py
@@ -12,7 +12,7 @@ from django.utils.safestring import mark_safe
 from wagtail.admin.staticfiles import versioned_static
 from wagtail.core.rich_text import RichText, get_text_for_indexing
 from wagtail.core.telepath import Adapter, register
-from wagtail.core.utils import resolve_model_string
+from wagtail.core.utils import camelcase_to_underscore, resolve_model_string
 
 from .base import Block
 
@@ -72,10 +72,23 @@ class FieldBlockAdapter(Adapter):
     js_constructor = 'wagtail.blocks.FieldBlock'
 
     def js_args(self, block, context):
+        classname = [
+            'field',
+            camelcase_to_underscore(block.field.__class__.__name__),
+            'widget-' + camelcase_to_underscore(block.field.widget.__class__.__name__),
+            'fieldname-' + block.name,
+            # TODO: if errors then add 'error'
+        ]
+
         return [
             block.name,
             context.pack(block.field.widget),
-            {'label': block.label, 'required': block.required, 'icon': block.meta.icon},
+            {
+                'label': block.label,
+                'required': block.required,
+                'icon': block.meta.icon,
+                'classname': ' '.join(classname),
+            },
         ]
 
     class Media:
--- a/client/src/entrypoints/admin/telepath/blocks.js
+++ b/client/src/entrypoints/admin/telepath/blocks.js
@@ -20,16 +20,14 @@ function initBlockWidget(id) {
   const blockDef = window.telepath.unpack(blockDefData);
   const blockValue = JSON.parse(body.dataset.value);
 
-  // replace the 'body' element with the unpopulated HTML structure for the block
-  const block = blockDef.render(body, id);
-  // populate the block HTML with the value
-  block.setState(blockValue);
+  // replace the 'body' element with the populated HTML structure for the block
+  blockDef.render(body, id, blockValue);
 }
 window.initBlockWidget = initBlockWidget;
 
 
 class FieldBlock {
-  constructor(blockDef, placeholder, prefix) {
+  constructor(blockDef, placeholder, prefix, initialState) {
     this.blockDef = blockDef;
     this.type = blockDef.name;
 
@@ -47,6 +45,7 @@ class FieldBlock {
     $(placeholder).replaceWith(dom);
     const widgetElement = dom.find('[data-streamfield-widget]').get(0);
     this.widget = this.blockDef.widget.render(widgetElement, prefix, prefix);
+    this.widget.setState(initialState);
   }
 
   setState(state) {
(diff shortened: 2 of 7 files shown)


wagtail/core/blocks/field_block.py
```diff
@@ -91,6 +91,10 @@ def js_args(self, block, context):
             },
         ]
 
+    def get_media(self, block, context):
+        media = super().get_media(block, context)
+        return media + block.field.widget.media
```
gasman, line 96: I think this should happen in the widget's own adapter rather than FieldBlock's. Handily you've just implemented [the TODO I left there](https://github.com/wagtail/wagtail/blob/f6f06175e0362d04140eaa6c4972c9e9bf8d0a21/wagtail/core/widget_adapters.py#L23-L25), so that should be a straightforward fix :-)

gasman (changes requested):

wagtail/core/blocks/field_block.py
kaedroho: Oh nice! Forgot widgets had their own adapters!

kaedroho: Fixed! (squashed into the last commit)

kaedroho:
Rebased

wagtail/core/blocks/field_block.py
```diff
@@ -545,6 +558,9 @@ def get_prep_value(self, value):
         # the JSONish representation
         return value.source
 
+    def get_form_state(self, value):
+        return value.source
```
gasman, line 562: This method should be returning the value that's directly usable/renderable by the client-side component, which for Draftail is the ContentState representation. As it stands, this block saves to the database correctly but doesn't re-render the value when the form is reshown.

Looking at the rich text widget code, I think we should be calling the widget's `format_value` method here - and since that's a core Django method that could come into play on any form widget, that should happen on FieldBlock.

gasman (changes requested):

wagtail/core/blocks/field_block.py
kaedroho: This should be fixed now. But there's something weird going on with images. For some reason, it saves images with the ```` tag in the HTML, and this doesn't get converted back into the contentstate in this branch, but it does on master.

No idea what's causing this, I had a poke around the internals and found there isn't an element handler for ```` and the ``ImageElementHandler`` is registered against ``<embed>``.

Are there any differences in the rich text configuration on the telepath branches?

gasman:
Rebased on current feature/telepath.

client/src/entrypoints/draftail.js
```diff
@@ -40,3 +40,33 @@ const plugins = [
...
+  constructor(options) {
+      this.options = options;
+  }
+
+  render(container, name, id) {
+    const options = this.options;
+
+    return {
+      getValue() {
+        return input.value;
```
gasman, line 54: `input` is not defined here.

client/src/entrypoints/draftail.js
```diff
@@ -40,3 +40,33 @@ const plugins = [
...
+        input.id = id;
+        input.name = name;
+        input.value = state;
+        container.appendChild(input);
+        draftail.initEditor('#' + id, options, document.currentScript);
+      }
+    };
+  }
+}
+window.telepath.register('wagtail.widgets.DraftailRichTextArea', DraftailRichTextArea);
```
gasman, line 72: Not sure it's safe for this to go in draftail.js - draftail as a Django widget is likely to appear on views that don't include telepath.js.

gasman (changes requested):

wagtail/core/blocks/field_block.py
gasman: Now fixed in 7c1bba6be7bd76383815671a9fc74bea941f062e - `FieldBlock.get_form_state` now passes the block value through both `block.value_for_form` (which retrieves the .source of the RichText value) and `widget.format_value` (which performs the dbHTML -> contentState conversion). Missing the first step meant that the contentState conversion was being applied to the front-end-rendered version of the field instead.

client/src/entrypoints/draftail.js
```diff
@@ -40,3 +40,33 @@ const plugins = [
...
+
+    return {
+      getValue() {
+        return input.value;
+      },
+      getState() {
+        return input.value;
+      },
+      setState(state) {
+        // We must initialise it here because Draftail doesn't provide a way to change state after initialisation
```
gasman, line 60: Hmm. As far as I can think of, nothing in StreamField's current or planned feature set requires us to be able to change the state of an already-initialised widget. (For example, copying blocks will only require us to read the state of an existing block and write that to a newly-initialised one.)

However, the getState / setState API is designed on the assumption that this _is_ possible, and if that's not the case then we should probably revise the API (i.e. have the `render` method accept an initial state) rather than have Draftail pretend it supports the API as currently designed (and then have it come back to bite us when we're using that API more intensively)

I have to wonder how much work it would be to update Draftail to accommodate this, but either way that doesn't change the basic facts: if it's non-trivial to get Draftail to co-operate, then chances are there are going to be other Django widgets in the wild that are similarly difficult or impossible to fit into telepath's world view. It's in our interests to keep telepath as permissive as possible when it comes to integrating 'difficult' widgets - we don't want "rewrite your Django widgets in telepath" to be the new "rewrite your Django widgets in React" :-)

With that in mind, I think the way forward is:

* Add an `initialState` argument to the `render` method in telepath's widget API, and update the StreamField infrastructure to use that rather than calling setState on newly-initialised widgets
* Keep setState as an optional method of the API. Widgets can still use it in their own internals (as PageChooser does, for example) and outside code can call it _if_ it knows that it's dealing with a widget type that implements it. Container blocks like StructBlock can still implement setState recursively, but it will only be functional if all descendant blocks also implement it.

It's quite possible that in future, requirements will come up that _do_ require setState to be implemented more generally - for example, I can imagine people wanting a general-purpose "if field X is filled in, then prefill field Y" mechanism as per the title/slug fields. Establishing setState as an optional non-essential method now will make it easier to reason about those interactions and the limitations of them when they do come up (i.e. it won't work for prefilling a rich text field until we go ahead and implement that properly in Draftail).

gasman:
Rebased and revised to address https://github.com/wagtail/wagtail/pull/6665#discussion_r553310883 . A bit of a yak-shave I'm afraid, but a productive one: telepath/blocks.py has been linted and reorganised so that each block type has a distinct definition class (the one that we had before, which maps to the Python-side class) and block class (the thing that actually holds a block value and has getState / setState accessors - previously this was just the return value of `render`). We then update blocks and widgets so that an initial state can be passed on initialisation, and that allows us to implement the Draftail widget cleanly without having to hack around the initialisation sequence. Ready for re-review!

kaedroho:
Changes look good to me! I cannot approve as it is not my PR though

PR #6666, opened 2021-01-04
Linting Telepath JS

kaedroho:
This PR contains some linting for the Telepath JavaScript. I've avoided large-scale issues such as indentation to reduce merge conflicts.

I had to change the eslint parser to ``babel-eslint`` in order to support class properties. See https://stackoverflow.com/questions/60046847/eslint-does-not-allow-static-class-properties

### DIFF
--- a/client/src/entrypoints/telepath/widgets.js
+++ b/client/src/entrypoints/telepath/widgets.js
@@ -1,4 +1,7 @@
-/* eslint-disable */
+/* eslint-disable indent, vars-on-top, no-warning-comments, max-len */
+
+/* global $ */
+
 class BoundWidget {
     constructor(element, name) {
         var selector = ':input[name="' + name + '"]';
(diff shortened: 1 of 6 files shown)


gasman (approved):

client/src/entrypoints/telepath/widgets.js
```diff
@@ -21,16 +24,17 @@ class Widget {
...
 
-    boundWidgetClass = BoundWidget;
+    static boundWidgetClass = BoundWidget;
 
     render(placeholder, name, id) {
         var html = this.html.replace(/__NAME__/g, name).replace(/__ID__/g, id);
         var dom = $(html);
         $(placeholder).replaceWith(dom);
+        // eslint-disable-next-line new-cap
         return new this.boundWidgetClass(dom, name);
```
gasman, line 34: Oops, missed this... making boundWidgetClass a static property means that `this.boundWidgetClass` on an instance is no longer valid.

gasman: Linting still passes with `static` removed, so have reverted this in 8746596544ac68616adba46fffda8e2d68cdf906.

PR #6670, opened 2021-01-06
Telepath: Date/Time inputs

kaedroho:


client/src/entrypoints/admin/date-time-chooser.js
```diff
@@ -1,4 +1,5 @@
 import $ from 'jquery';
+import { Widget } from './telepath/widgets';
```
gasman, line 2: Importing telepath/widgets and calling `window.telepath.register` from this file will make it fail on any view that doesn't include the main telepath.js module (which currently means any view without a streamfield component). Widget registration code should be in its own file, which is then specified in the Python-side widget adapter's media.

gasman:
Merged in b79637b4f9f2d4edfc13e0907d1315155c31c216 (with JS telepath adapters moved to telepath/widgets.js, and a few functions fixed to deal with API changes since this was opened). Thanks @kaedroho!

PR #6672, opened 2021-01-07
WIP: RTD docs test

thibaudcolas:
**Do not merge**. Feedback welcome ! This PR is to trial two Sphinx and Read the Docs features:

### [Autobuild Documentation for Pull Requests](https://docs.readthedocs.io/en/stable/guides/autobuild-docs-for-pull-requests.html)

This is now working exactly as intended, with fast build times due to skipping some of the build (search index, and offline docs formats as well?), a GitHub PR status check linking straight to the build logs while in progress, and to the built documentation once finished: https://wagtail--6672.org.readthedocs.build/en/6672/.

There is a nice "this is a PR" warning on all of the pages, and indexing is disabled with robots.txt, as well as having the canonical meta pointing to the live docs.

### Sphinx Markdown support

- Markdown source: see changes in this PR
- Sphinx / RTD output: https://wagtail--6672.org.readthedocs.build/en/6672/reference/contrib/legacy_richtext.html

I’m still exploring the capabilities / limitations here but this is already very promising.

[recommonmark](https://recommonmark.readthedocs.io/en/latest/index.html) does all of the heavy lifting. It really is just a matter of adding the extension in Sphinx and configuring it. Then we can mix and match Markdown and rST within the same project documentation.

There are a few more things to enable to also be able to have rST snippets within Markdown files. This means we could still use Sphinx autodoc where we want to, or admonitions, while using Markdown for more free-form content. This feels a bit strange but would help tremendously in having a more gradual migration from one format to the next.

**Specific point of feedback I’m particularly interested in**: I think the whole premise of switching to Markdown is to make documentation contributions more approachable, so I’d be keen to hear people’s thoughts on whether the RST embeds are desirable in that respect, or whether we’d rather rewrite docs to rely less on this more complex formatting.


docs/releases/2.10.rst
```diff
@@ -81,7 +81,7 @@ Other features
  * Add ``after_edit_snippet``, ``after_create_snippet`` and ``after_delete_snippet`` hooks and documentation (Kalob Taulien)
  * Improve performance of empty search results by avoiding downloading the entire search index in these scenarios (Lars van de Kerkhof, Coen van der Kamp)
  * Replace ``gulp-sass`` with ``gulp-dart-sass`` to improve core development across different platforms (Thibaud Colas)
- * Remove markup around rich text rendering by default, provide a way to use old behaviour via ``wagtail.contrib.legacy.richtext``. See :ref:`legacy_richtext`. (Coen van der Kamp, Dan Braghis)
+ * Remove markup around rich text rendering by default, provide a way to use old behaviour via ``wagtail.contrib.legacy.richtext``. See :doc:`legacy_richtext`. (Coen van der Kamp, Dan Braghis)
```
thibaudcolas, line 84: This and below aren’t working as expected, https://wagtail--6672.org.readthedocs.build/en/6672/releases/2.10.html. It might need the full path rather than just the filename.

docs/reference/contrib/legacy_richtext.md
```diff
@@ -0,0 +1,26 @@
+# Legacy richtext
```
thibaudcolas, line 1: toctree index working as expected with no changes: https://wagtail--6672.org.readthedocs.build/en/6672/reference/contrib/index.html

docs/reference/contrib/legacy_richtext.rst
```diff
@@ -1,26 +0,0 @@
-.. _legacy_richtext:
-
-=====================
-Legacy richtext
-=====================
-
-.. module:: wagtail.contrib.legacy.richtext
```
thibaudcolas, line 7: On its own this didn’t seem to do anything? https://docs.wagtail.io/en/stable/reference/contrib/legacy_richtext.html

docs/reference/contrib/legacy_richtext.md
```diff
@@ -0,0 +1,61 @@
...
+
+Potato `(っ◕‿◕)っ`
+
+Test eval RST
```
thibaudcolas, line 5: This and below currently cause a warning in the Sphinx build, but as far as I can see it’s only because of the content being duplicated between two files rather than any syntax issue.

docs/reference/contrib/legacy_richtext.rst
allcaps: It think it only converts to a link if the reference (autodoc) has this module. I'm not 100% sure. But can have a look.

In RST/Sphinx there is also a way to cross reference external docs. For example point to Django docs in the same way.

docs/reference/contrib/legacy_richtext.md
thibaudcolas: One added note – I think the whole premise of switching to Markdown is to make documentation contributions more approachable, so I’d be keen to hear people’s thoughts on whether the RST embeds are desirable in that respect, or whether we’d rather rewrite docs to rely less on this more complex formatting

thibaudcolas:
We discussed this in today’s team meeting, and I’ll close this now. Overall we want to proceed with the tooling changes. For content changes, we still need to review formatting features of rst and Markdown and have some high-level agreement on what we want to use or avoid.

PR #6673, opened 2021-01-07
Telepath: TableBlock

kaedroho:
This adds an adapter for ``TableInput`` so that it can render table blocks.

Note, this requires widget media to be pulled in which is currently only supported in https://github.com/wagtail/wagtail/pull/6665


gasman:
Looks good to me! Merged in 4769e1e8537a1d8253796c98927fcd0fcb67b88b (with the necessary API updates again).

Issue #6675, opened 2021-01-08
Multiple clicks on image "Update" button results in duplicate images in the system


m3brown:
Found a bug? Please fill out the sections below. 👍

### Issue Summary

It is possible for consecutive clicks to create duplicate images in the create_from_uploaded_image form. In my case, this issue is exacerbated because of a post_save signal on Image creation, which pauses the form UI for a second or two. In other words, it's a race condition that probably doesn't affect most users.

I imagine the same issue exists for documents.

This issue is related to #305.

### Steps to Reproduce

1. Create a post_save signal to pause image creation for a second or two
2. Upload an image
3. At the `Upload successful. Please update this image...` prompt, click "Update" several times until the form disappears.
4. Visit the image list view, and multiple copies of the image will be present.

Any other relevant information. For example, why do you consider this a bug and what did you expect to happen instead?

From my observations, the root of the issue is that the form button is not disabled on click, leaving the door open to more than one click. Since the transaction is usually pretty fast, this isn't typically an issue, but I still think it should be fixed for use cases where the backend may not be as responsive as we'd like.

The CSS styling of this solution isn't create, but one example of a workaround is to add the following line above the post in [add-multiple.js](https://github.com/wagtail/wagtail/blob/cf38b2457dba521d029d3a6e5024a7d0c3f5faf6/wagtail/images/static_src/wagtailimages/js/add-multiple.js#L137):

```javascript
$(this).find(':submit').prop("disabled", true);
```

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: no

### Technical details

* Python version: 3.6.
* Django version: 2.2.17
* Wagtail version: 2.11.3
* Browser version: Edge 87

gasman:
Fixed in #6676.

PR #6676, opened 2021-01-08
Disable image/document Update button on click

m3brown:
This is one solution for resolving #6675, although I'll be the first to admit it's not the most elegant PR.

### General notes
- There is no need to implement a hook to re-enable if the form succeeds or fails. On success the form is deleted, on failure it is re-rendered from scratch.
- Screenshots of the styling of the button are provided below

### Discussion topics
- It didn't seem reasonable for me to add tests confirming this race condition, since I couldn't find any existing tests to leverage as a starting point.
- The styling of the disabled button is not ideal during hover. I'm not familiar with the styling options available in the admin, suggestions are appreciated if there's an obvious class to add.
- A more elegant solution might be to add handling to the UploadedImage model itself to prevent duplicates at the model layer (i.e. a semaphore approach)

### Screenshots

after click, without mouseover:
image

after click, with mouseover (arrow mouse icon):
image

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/generic/multiple_upload/edit_form.html
+++ b/wagtail/admin/templates/wagtailadmin/generic/multiple_upload/edit_form.html
@@ -1,4 +1,4 @@
-{% load i18n %}
+{% load wagtailadmin_tags i18n %}
 
 {% include "wagtailadmin/shared/non_field_errors.html" %}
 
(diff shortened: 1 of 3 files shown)


gasman (changes requested):
Thanks @m3brown! There's an existing `button-longrunning` class used in a few other places for this purpose:
https://github.com/wagtail/wagtail/blob/14f63519c2c355637cfcb20715971dfaf9346fbf/client/src/entrypoints/admin/core.js#L311

I suspect it might need a bit of re-plumbing to make that work, though, since in the current version the event handler gets attached on page load, and these forms are dynamically inserted after that point - would you be happy to pick that up?

m3brown:
@gasman Thanks for the insight. It turns out the button-longrunning hooks wire up fine. Here are the updated screenshots:

image

image

gasman:
Thanks again @m3brown! Merged in 30ec4631d21a732ab309c7891ddbd3babdf9cb76.

PR #6678, opened 2021-01-10
Use .specific_deferred when calling .get_admin_display_title()

kaedroho:
There are quite a few places where we are not calling ``.get_admin_display_title()`` with a specific page instance so custom implementations of this method weren't being used.

@ababic's [`.specific_deferred`](https://github.com/wagtail/wagtail/pull/6661) method allows us to get a specific instance without an extra query. So custom implementations of ``.get_admin_display_title`` will be called without much of a performance hit for those who don't use this feature (hopefully).


gasman:
Merged in 44f3731b1cd458f1fce0dbf87507805fce5fea4f

PR #6681, opened 2021-01-11
signature for `get` includes `request`

minusf:
a trivial change to add `request` to the `get` signature.  it's nicer to override too...


gasman:
Hi @minusf,
I don't see how this improves things for either Wagtail core development or people overriding this method - the override method can still use `def get(self, request, *args, **kwargs)` even if the base method doesn't - but happy to reconsider if you can provide more details.

minusf:
no worries, i thought it was an oversight.

(i like consistency, and `get` uses `request` multiple times)

PR #6683, opened 2021-01-11
Apply scss styling to collection_choice_field

### LINKED ISSUE #6682: Chevron missing in collection_choice_field select widgets

### Issue Summary

Select elements in wagtail forms that render a collection dropdown are missing the chevron icon on the right.

image

Expected style:

image

It appears this issue is related to #6141, where the css class is now rendered as "collection_choice_field".

### Steps to Reproduce

2. Go to the Images page from the left panel
3. Add an image
4. Edit the image
5. Observe the "collection" dropdown does not have chevron icon

The missing icon is in several other places, notably:
- in the "add an image" form, which looks to be a case of a missing span that is not the same as this bug.
- in the right hand sidebar form in Reports (site history, etc.), which is a different problem than this bug.

### Technical details

* Python version: 3.6
* Django version: 2.2.17
* Wagtail version: 2.11.3
* Browser version: Edge 87, Firefox 83

m3brown:
This is a simple change to include collection_choice_field in the set of class names that are styled with a chevron icon. This change notably impacts the Image and Document edit forms.

Before:
image

After:
image

This PR resolves #6682.

### DIFF
--- a/client/scss/components/_forms.scss
+++ b/client/scss/components/_forms.scss
@@ -77,6 +77,7 @@ select::-ms-expand {
 }
 
 // select boxes
+.collection_choice_field .input,
 .choice_field .input,
 .model_choice_field .input,
 .typed_choice_field .input {


gasman:
Thanks @m3brown - good catch! Merged in cadb5456c032829ca907fb0af60700d2e6a713f8.

PR #6688, opened 2021-01-12
Update telepath serialisation format to pack/unpack arguments recursively

gasman:
As per https://paper.dropbox.com/doc/I-think-we-should-redesign-telepath--BDFQDvczLORqNyxworD6uNCiAg-jprxEyru9gLGWIxS7zxin

### DIFF
--- a/wagtail/core/telepath.py
+++ b/wagtail/core/telepath.py
@@ -4,7 +4,73 @@ from django.forms import MediaDefiningClass
 from wagtail.admin.staticfiles import versioned_static
 
 
-adapters = {}
+DICT_RESERVED_KEYS = ['_type', '_args', '_dict']
+
+
+class BaseAdapter:
+    """Handles serialisation of a specific object type"""
+    def pack(self, obj, context):
+        """
+        Translates obj into serialisable form. Any media declarations that will be required for
+        deserialisation of the object should be passed to context.add_media().
+
+        This base implementation handles simple JSON-serialisable values such as strings, and
+        returns them unchanged.
+        """
+        return obj
+
+
+class DictAdapter(BaseAdapter):
+    """Handles serialisation of dicts"""
+    def pack(self, obj, context):
+        packed_obj = {
+            str(key): context.pack(val)
+            for key, val in obj.items()
+        }
+        if any(reserved_key in packed_obj for reserved_key in DICT_RESERVED_KEYS):
+            # this dict contains keys such as _type that would collide with our object notation,
+            # so wrap it in an explicit _dict to disambiguate
+            return {'_dict': packed_obj}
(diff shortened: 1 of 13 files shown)


wagtail/core/telepath.py
```diff
@@ -4,7 +4,73 @@
...
+    def pack(self, obj, context):
+        context.add_media(self.get_media(obj))
+        return {
+            '_type': self.js_constructor,
+            '_args': [context.pack(arg) for arg in self.js_args(obj)]
+        }
+
+
+adapters = {
+    # Primitive value types that are unchanged on serialisation
```
kaedroho, line 62: Where would be the best place to register adapters for standard library types like dates and Decimals?

wagtail/core/telepath.py
```diff
@@ -16,21 +82,27 @@ def __init__(self):
...
+        try:
+            return [self.pack(item) for item in obj]
+        except TypeError:
+            pass
 
-
-class Adapter(metaclass=MediaDefiningClass):
-    def get_media(self, obj, context):
-        return self.media
+        raise Exception("don't know how to pack object: %r" % obj)
```
kaedroho, line 108: Could we use a more specific exception type? (probably a subclass of ``LookupError``?)

kaedroho (approved):
Looks good to me! just a couple of questions that I think can be answered later.

wagtail/core/telepath.py
gasman: Same as we've done for Django widgets (wagtail.core.widget_adapters) - some module that gets loaded on startup via the appconf. It would probably make sense for dates (and possibly decimals? A bit rusty on how they work in Python vs JS) to be built in to the base telepath module as standard though.

wagtail/core/telepath.py
gasman: Sure - json.dumps uses TypeError for the equivalent failure case, so I'd be inclined to do the same.

gasman: ...ah, but then the problem with TypeError is that it'll be caught and masked by the iterable handler (so it'll end up reporting a failure on the top-level list rather than the offending object) 🤔 

        # as fallback, try handling as an iterable
        try:
            return [self.pack(item) for item in obj]
        except TypeError:
            pass

        raise TypeError("don't know how to pack object: %r" % obj)

gasman: Fixed in 7b62bfca2950f95e17c17bda09be06463616188c, although I might be overthinking this for the sake of keeping it as a TypeError - @kaedroho what do you think?

kaedroho: Looks good to me!

PR #6695, opened 2021-01-15
Telepath - packing / unpacking objects by reference

gasman:
Optimise the telepath representation of values that reference the same object multiple times - rather than duplicating that object in the data, we tag it with an ID on the first occurrence and reference that ID on subsequent uses. This will be especially important for complex StreamField definitions that re-use the same block in multiple places (e.g. a teaser that can appear at the top level, or as a child of a two-column layout block in either column).

(Note that it doesn't support cyclic references, e.g. an object containing a reference to itself in its arguments. But that's fine, because neither does StreamField...)


client/src/components/Telepath/Telepath.js
```diff
@@ -0,0 +1,137 @@
+/* eslint-disable dot-notation, no-param-reassign */
```
kaedroho, line 1: I think this belongs in either `utils` or its own root folder rather than in components

client/src/components/Telepath/Telepath.test.js
```diff
@@ -0,0 +1,158 @@
...
+    this.artists = artists;
+  }
+}
+
+telepath.register('music.Album', Album);
+
+describe('Telepath', () => {
+  it('can unpack objects', () => {
+    const beyonce = telepath.unpack({ _type: 'music.Artist', _args: ['Beyoncé'] });
+    expect(beyonce.name).toBe('Beyoncé');
```
kaedroho, line 27: Probably should have a `expect(beyonce).toBeInstanceOf(Artist)` here?

kaedroho (approved):
This is really awesome! Just a couple of minor comments but otherwise looks good to me!

client/src/components/Telepath/Telepath.test.js
gasman: Good call. This was my first time using jest, so I wasn't being very adventurous :-)

PR #6696, opened 2021-01-15
Refined creation of title on image upload

### LINKED ISSUE #4945: Feature Request:  Refine the creation of the default file title when uploading an image

lb-:
Resolves #4945 & is a smaller scoped version of #5616 that only covers images (not documents).

## Overview of implementation
* New behaviour, by default -the file name will be used to generate a title which will be the file without the file extension
* If a value is already in the applicable title field, by default this value will be kept as is.
* Moves the implementation to 100% client side so user can 'edit' value before it is saved (or in the case of multiple upload, the initial title will based on the util)
* Adds the ability for this functionality to overridden (includes docs on how to do this)

## Forms updated
- view - admin/images/multiple/add
- view - admin/images/add
- modal - image-chooser-modal

## How to validate
* For each of the forms above, attempt to upload an image and check the title field gets pre-filled with the filename but not the extension.
* For the non-multiple forms, before clicking save, select a different file to upload and confirm the title is left unchanged.

## PR Checklist
* Do the tests still pass? the core.js is actually untested, I am happy to get some tests started but it appears this will be quite some effort.
* Does the code comply with the style guide?  ✅  - js lint passes
* For front-end changes: Tested on the following
  * Safari Version 12.1.1 MacOS 10.14.5
  * Firefox 69 MacOS 10.14.5
  * Chrome 77 MacOS 10.14.5

### DIFF
--- a/client/src/entrypoints/admin/core.js
+++ b/client/src/entrypoints/admin/core.js
@@ -582,4 +582,6 @@ function initButtonSelects() {
 
 $(document).ready(initButtonSelects);
 
+window.wagtail.utils = {};
+
 window.wagtail = wagtail;
--- a/docs/advanced_topics/images/custom_title_generation_from_filename.rst
+++ b/docs/advanced_topics/images/custom_title_generation_from_filename.rst
@@ -1,0 +1,48 @@
+.. _image_custom_title_generation_from_filename:
+
+Custom Title Generation From Filename
+=====================================
+
+Override how the title is set when uploading single files, multiple files or uploading a file within a chooser modal.
+
+.. note::
+    This will not apply to editing images (e.g. uploading a replacement image file while editing an existing image).
+
+``wagtail.utils.getImageUploadTitle`` can be overridden with a custom function,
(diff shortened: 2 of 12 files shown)


lb-:
@thibaudcolas I have basically rewritten the code changes from #5616 to simplify this whole approach, this PR only covers images (if this is merged, I will do a similar one for documents), removed a bunch of the abstraction and minimised the changes across other client code. It would be great if you could review this (I will keep an eye on the CI for any fixes I need to make).

@gasman  hopefully this is more inline with how we want to approach images/documents in isolation, plus each widget/form implementation is now isolated and 'closer' to the code that uses it (except for the smaller global filename to title util).

comment ref for context. https://github.com/wagtail/wagtail/pull/5616#pullrequestreview-435860898

lb-:
Replaced with #7534

PR #6698, opened 2021-01-15
Bugfix: PROTECTED with related_name='+'

m3brown:
This PR is an attempt to resolve wagtail/wagtail-modeladmin#30.

Changes in this PR:
- Make adjustments to views.py to handle `related_name="+"`
- Add test cases for ForeignKey and OneToOneField
- Refactor some of the surrounded code to properly support cases where there are multiple "+" relationships
  - The old code would only pick one, since the keys in `fields_map` would conflict

* Does the code comply with the style guide? I ran flake8 on all the modified files


wagtail/contrib/modeladmin/views.py
```diff
@@ -798,20 +807,23 @@ def post(self, request, *args, **kwargs):
...
-            fields = self.model._meta.fields_map.values()
-            fields = (obj for obj in fields if not isinstance(
-                obj.field, ManyToManyField))
-            for rel in fields:
+            for rel in self._get_related_fields():
                 if rel.on_delete == models.PROTECT:
                     if isinstance(rel, OneToOneRel):
                         try:
-                            obj = getattr(self.instance, rel.get_accessor_name())
+                            if rel.get_accessor_name() != '+':
```
m3brown, line 814: Arguably, the performance of both `if` and `else` OneToOneField implementations are similar, they're both a "get" operation, so we could potentially drop the `if` and use the `else` implementation all the time...

In contrast, for the implementation for ManyToOneFields, I assume a releated_name lookup is more efficient than doing a filter of the related model? In other words, I imagine leaving the `if` and `else` implementations makes sense for ManyToOneField.

laymonage:
Hello, I'm closing this as it's quite dated and we no longer accepting PRs for new ModelAdmin features in Wagtail as per wagtail/rfcs#85. If you'd like, you could create a new PR to the [wagtail-modeladmin](https://github.com/wagtail-nest/wagtail-modeladmin) repository. If the bug fix is critical, we might consider backporting it to the Wagtail repo during the deprecation period. Thank you!

Issue #6647, opened 2020-12-17
UserCreationForm requires custom user models to have certain fileds


dakaza98:
According to the documentation about custom user models, [found here](https://docs.wagtail.io/en/v2.11.3/advanced_topics/customisation/custom_user_models.html), the minimum for using UserCreationForm is to inherit AbstractBaseUser and PermissionsMixin:

> Create a custom user model. This must at minimum inherit from AbstractBaseUser and PermissionsMixin.

However, the UserCreationForm still has some standard fields defined that all custom user models must have:
https://github.com/wagtail/wagtail/blob/e48c1704cb13e222740d1ccc1ebbdc0e1ca92f44/wagtail/users/forms.py#L184

The standard fields are defined here
https://github.com/wagtail/wagtail/blob/e48c1704cb13e222740d1ccc1ebbdc0e1ca92f44/wagtail/users/forms.py#L32

This means that if I have a user model that does not use `first_name` and `last_name`, I stumble upon many problems. I assume that this is something that remains from when the model was required to be an AbstractUser (which was changed in [this commit](https://github.com/wagtail/wagtail/commit/d65532dd0f0bf297523a4925fd79286b1d403a5d#diff-e22c4c9af0117a237de54b50fc964ef9a06e475500c0b6c596833da251424576R10)). Or perhaps it is a typo in the documentation? Either way this would be a good thing to sort out

favoyang:
The docs said,

> Create a custom user model. This must at minimum inherit from AbstractBaseUser and PermissionsMixin. In this case we extend the AbstractUser class and add two fields. The foreign key references another model (not shown).

The example is based on extending from AbstractUser class, which provided all the standard_fields you mentioned. If you inherit from AbstractBaseUser and PermissionsMixin, then you need to create these forms from scratch.

dakaza98:
Ah of course! They are just forms that can be overridden. Not the most optimal way to do it but it works. Thanks for pointing me in the right direction!

PR #6702, opened 2021-01-16
Combined account settings into a single form

kaedroho:
This PR combines all builtin account settings into a single form and adds a mechanism for adding new panels to this form. The old hook is still supported with links appearing after the form in a "More actions" section.

image


kaedroho:
Some tweaks from an internal chat:
- [ ] Use salmon bars to distinguish the sections
- [ ] Add tabs and put notifications in a separate tab. default tab should be called "personal details"
- [ ] Merge language/timezone into a single panel called "locale"

kaedroho:
We have a design for this now in Figma, but I might not have time to implement this in the next few weeks so let me know if anyone else wants to pick this up!

zerolab:
@kaedroho it is worth sharing the designs, so whoever picks this up can hit the ground running

kaedroho:
@zerolab 

It's in Figma which makes it very difficult to screenshot and I will have to share directly (not sure if I can make it public)

kaedroho:
Ready for review!

kaedroho:
Rebased. @gasman any chance you could review this?

wagtail/admin/views/account.py
```diff
@@ -44,80 +49,207 @@ def password_reset_enabled():
...
-    if not email_management_enabled():
-        raise Http404
-    if request.method == 'POST':
-        form = EmailForm(request.POST, instance=request.user)
+@sensitive_post_parameters()
+def account(request):
+    # Fetch the user and profile objects once and pass into each panel
+    # We need to use the same instances for all forms so they don't overwrite each other
+    user = request.user
+    profile = UserProfile.get_for_user(request.user)
```
zerolab, line 195:
```suggestion
    profile = UserProfile.get_for_user(user)
```

since `user = request.user` just above

wagtail/admin/views/account.py
```diff
@@ -44,80 +49,207 @@ def password_reset_enabled():
...
+    panels = [
+        NameEmailSettingsPanel(request, user, profile),
+        AvatarSettingsPanel(request, user, profile),
+        NotificationsSettingsPanel(request, user, profile),
+        LocaleSettingsPanel(request, user, profile),
+        ChangePasswordPanel(request, user, profile),
+    ]
+    for fn in hooks.get_hooks('register_account_settings_panel'):
+        panel = fn(request, user, profile)
+        if panel:
```
zerolab, line 207:
```suggestion
        if panel and panel.is_active():
```

which would reduce the need for `panels = [panel for panel in panels if panel.is_active()]` below

wagtail/admin/views/account.py
```diff
@@ -44,80 +49,207 @@ def password_reset_enabled():
...
-        'form': form,
-    })
+        if all(form.is_valid() or not form.is_bound for form in panel_forms):
+            with transaction.atomic():
+                for form in panel_forms:
+                    if form.is_bound:
+                        form.save()
 
+            # Prevent a password change from logging this user out
+            update_session_auth_hash(request, user)
```
zerolab, line 233: question: does it make sense to run this when pass management is disabled and not `user.has_usable_password` ?

zerolab (commented):
This looks and works so much better, thank you @kaedroho.

Got a few minor suggestions and a question, otherwise LGTM

wagtail/admin/static_src/wagtailadmin/scss/layouts/account.scss
```diff
@@ -0,0 +1,38 @@
+@import '../../../../../../client/scss/settings/variables';
+
+.top-padding {
```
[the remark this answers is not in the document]
kaedroho: I avoided putting this into the main CSS because it feels like a hacky workaround. Ideally, the salmon bars should allow space for the content beneath.

wagtail/admin/views/account.py
kaedroho: It updates a value in the user's current session to a key that is derived from their password. if the password is not changed, this would just set that value to what it is currently and have no effect.

zerolab: Thanks Karl. I do get the mechanics of this (and followed the code to refresh my memory before asking the question). Was more curious than anything if the additional processing made sense. At the same time we'd need our own additional processing to check if the password was changed so it doesn't matter :)

wagtail/admin/templates/wagtailadmin/account/settings_panels/avatar.html
```diff
@@ -0,0 +1,37 @@
...
+                <img src="{% avatar_url request.user 100 %}" alt="">
+            </div>
+        </div>
+        <div class="avatar-panel__form">
+            <label for="id_avatar-avatar">{% trans "Upload a profile picture:" %}</label>
+            <input type="file" name="avatar-avatar" accept="image/*" id="id_avatar-avatar">
+            <input type="hidden" name="avatar-avatar-clear" id="avatar-avatar-clear_id">
+
+            {% if form.avatar.value %}
+                <button class="avatar-panel__revert-to-gravatar button button-secondary" data-gravatar-url="{% avatar_url request.user 100 gravatar_only=True %}">{% trans "Revert to gravatar" %}</button>
```
gasman, line 15: I think we should avoid explicitly referring to gravatars in the UI. Ever since we started supporting uploadable avatars, there's no reason for the typical user to know or care what a gravatar is - I'd consider the fact that we support gravatar at all to be a delighter / easter egg for the 1% who happened to have one previously set up. It's also possible for site owners to replace or disable gravatar support via WAGTAIL_GRAVATAR_PROVIDER_URL, which would make the label inaccurate.

Maybe "reset to default" is the simplest way of covering all bases?

zerolab: Came here to raise a point about `WAGTAIL_GRAVATAR_PROVIDER_URL` (prompted by the CSP issue) and, here it is, a true delighter from @gasman ;)

kaedroho: Thanks I agree! I do think there's some value in telling users where their avatar is coming from though, could we make it say "gravatar" if they haven't changed this setting, but "default" otherwise?

gasman: No, I don't think so - that means it's still there in the out-of-the-box setup, and presenting users with a term they don't understand and don't _need_ to understand is simply bad UX. For those users, it might as well say "frobnicate warp drive". :-)

The one case where it might be justifiable to mention Gravatar is if we detect that the user does have one, in which case it could be part of some tooltip help saying "this profile picture has been imported automatically from Gravatar" or similar, possibly with the option to drop it and fall back to the plain 'user' icon (does that option exist already?). That's probably overkill for this PR though.

kaedroho: Just made this change now, let me know if there's anything else!

gasman:
One thing that might need some attention here is how we handle custom user models that don't have first_name / last_name fields - i.e. ones that extend `AbstractBaseUser` but not `AbstractUser`. This is [permitted according to our custom user model docs](https://docs.wagtail.io/en/stable/advanced_topics/customisation/custom_user_models.html) and [apparently not totally unknown in the real world](https://github.com/wagtail/wagtail/issues/6647). In the existing 'edit name' view/form code I can't see any guard against this, and the documentation / discussion about this sort of customisation only ever seems to mention the user management forms in wagtailusers rather than this account area - so it's possible that this form has never worked in those custom setups, and people have just conditioned themselves to not click on that link... in which case we probably need a proper fix now that it's an unavoidable part of the page. Will do some further experimenting.

wagtail/users/forms.py
```diff
@@ -394,7 +395,11 @@ class Meta:
 
 
 def _get_language_choices():
-    return sorted(BLANK_CHOICE_DASH + get_available_admin_languages(),
+    language_choices = [
+        (lang_code, get_language_info(lang_code)['name_local'])
+        for lang_code, lang_name in get_available_admin_languages()
```
gasman, line 400: 👍  ... if we're not using lang_name here any more, then we can hopefully remove [the localisations on that list](https://github.com/wagtail/wagtail/blob/main/wagtail/admin/localization.py#L11) that were probably creating mostly-useless work for translators :-)

gasman:
Small-but-important UI issue: the 'old password' field is set to autofocus, which means that on successful submission, the page will tend to jump down to that field, obscuring the "Your account settings have been changed successfully!" message and making it look like the submission has failed... along with the password fields wrongly having a required asterisk, it creates the impression that you're being forced to change your password.

wagtail/users/forms.py
gasman: done in https://github.com/gasman/wagtail/commit/1e9d32915b8c95167256d17ba1bdc95b2295d941

gasman:
> Small-but-important UI issue: the 'old password' field is set to autofocus, which means that on successful submission, the page will tend to jump down to that field, obscuring the "Your account settings have been changed successfully!" message and making it look like the submission has failed... along with the password fields wrongly having a required asterisk, it creates the impression that you're being forced to change your password.

done in https://github.com/gasman/wagtail/commit/d5e9f084a3af13d9083e998f3308cdd956fe4770

gasman:
> One thing that might need some attention here is how we handle custom user models that don't have first_name / last_name fields - i.e. ones that extend AbstractBaseUser but not AbstractUser.

This is not as bad as I thought - if the user model is missing those fields, then the ModelForm treats them as additional non-database-backed fields, so the user needs to fill them in to pass validation, but on submission they're immediately forgotten. Which is not brilliant, but not a release blocker either.

There's a bit of related weirdness with the default user model: name and email are required at the form level but not the model level, and a user created through `./manage.py createsuperuser` won't have them - which means an admin is forced to fill them in the first time they use the account settings form. (As with the other issue, this isn't a change of behaviour - it's just that it was easier to ignore when the form was on its own page.) I think this _is_ the desired default behaviour - name/email being required is a good thing, and if anything `createsuperuser` is the outlier - but I anticipate we'll get requests to make this customisable sooner rather than later :-) Again, not a blocker.

Will open a rebased PR with the fixes on https://github.com/gasman/wagtail/tree/kaedroho-combined-account-settings once commenting is merged.

kaedroho:
Thanks for looking into that @gasman!

gasman:
@kaedroho No problem! If you're able to give https://github.com/gasman/wagtail/commit/d5e9f084a3af13d9083e998f3308cdd956fe4770 a quick sanity check, that would be good :-)

gasman:
Rebased at #7050

PR #6711, opened 2021-01-20
Configure Sphinx for Markdown support

thibaudcolas:
In preparation for the upcoming documentation sprint. This will allow us to have Markdown docs alongside rst.

The main change is the Sphinx build configuration, and I’ve also moved one page to Markdown to simplify testing.


docs/reference/contrib/legacy_richtext.md
```diff
@@ -0,0 +1,24 @@
+# Legacy richtext
```
allcaps, line 1: It's a nitpick. But I'm just curious...

Do we want to keep the section directive on the first line?

    ```eval_rst
    .. _legacy_richtext:
    ```
In that case, we don't have to update the links in the release notes.

Can we point from inline markdown links to this section marker? 🤔

docs/releases/2.10.rst
```diff
@@ -81,7 +81,7 @@ Other features
  * Add ``after_edit_snippet``, ``after_create_snippet`` and ``after_delete_snippet`` hooks and documentation (Kalob Taulien)
  * Improve performance of empty search results by avoiding downloading the entire search index in these scenarios (Lars van de Kerkhof, Coen van der Kamp)
  * Replace ``gulp-sass`` with ``gulp-dart-sass`` to improve core development across different platforms (Thibaud Colas)
- * Remove markup around rich text rendering by default, provide a way to use old behaviour via ``wagtail.contrib.legacy.richtext``. See :ref:`legacy_richtext`. (Coen van der Kamp, Dan Braghis)
+ * Remove markup around rich text rendering by default, provide a way to use old behaviour via ``wagtail.contrib.legacy.richtext``. See :doc:`/reference/contrib/legacy_richtext`. (Coen van der Kamp, Dan Braghis)
```
allcaps, line 84: We don't have to update these links if we keep the section marker.

allcaps (approved):
Hey Thibaud,

Great work. I really like the documentation preview build 👏 .

Just a nitpick: Is leaving out the section marker intentional? 

LGTM! 👋

docs/reference/contrib/legacy_richtext.md
thibaudcolas: I know you’ve already approved the PR but I think it’s a good point for us to discuss further and make sure we approach this consistently for other documents.

I’ve removed it for two reasons:

1. The reference is at the very top of the document, so practically the same as linking straight to the document.
2. Since the result is the same, I felt like we’d prefer the way that requires no reStructuredText.

As you mention the drawback is that we need to update links, which could be annoying if it creates merge conflicts. But ultimately for new content / after our migration, I think we would want to encourage linking straight to documents rather than the additional ref syntax? Unless of course the reference is further down the page and we want to link straight to a section.

---

For links in Markdown, yes, here is what a reference link looks like:

```markdown
[Link to a reference](register_reports_menu_item)

# Equivalent to :ref:`Link to a reference <register_reports_menu_item>`
# It’s unclear to me whether this only requires `.. _register_reports_menu_item:`, or also a heading with the same name.
```

And a doc link:

```markdown
[Link to document (with wrong extension!)](/reference/hooks.md)

# Equivalent to :doc:`/reference/hooks`
# Note how the extension is wrong, I’m not too sure why
```

cnk: Regarding that last point: I think it would be great to configure Sphinx to automatically add reference targets from document headings. I love docs with that feature and often use it to send someone a link to exactly the section of docs I think they need.

thibaudcolas: @cnk at least half of this should already be possible and will keep on working as-is – that additional configuration is essentially so we can also create the same links from within documentation sources.

See for example https://wagtail--6711.org.readthedocs.build/en/6711/reference/contrib/legacy_richtext.html#module-wagtail.contrib.legacy.richtext.

thibaudcolas:
Alright, I think the point about links can be addressed separately so I’ll merge this as-is and move on to creating a formatting reference.

PR #6712, opened 2021-01-20
Unit tests for date/time and table widgets

kaedroho:


client/src/entrypoints/contrib/table_block/table.js
```diff
@@ -224,10 +224,10 @@ class TableInput {
 
     const widget = {
       getValue() {
-        return input.value;
+        return JSON.parse(input.value);
       },
       getState() {
-        return input.value;
+        return JSON.parse(input.value);
```
gasman, line 230: Good catch!

gasman (approved):

PR #6713, opened 2021-01-20
Fix typo in generic EditView

helderco:
The generic EditView has a trailing comma in get_context_data which is making can_delete always True.


kaedroho (approved):

kaedroho:
@gasman should we packport this to 2.12?

kaedroho:
@gasman I've added this to both 2.11 and 2.12 branches

PR #6717, opened 2021-01-21
Assign file information to form object without using file.save

cnk:
I changed the way file information is transferred from the UploadedImage model to the form that will create the normal image model.

For most sites, my change will not make any difference but it fixes a very specific issue we started seeing in our code base between Wagtail 2.10 and 2.11. Uploads using the single image form such as is used in the Image chooser modal window worked fine but uploads using the image multiple upload path started failing with an error in our custom `get_upload_to` method which uses the image's collection name as the folder name when storing the image file. The problem only occurs when there is a custom image model with a required field AND has a get_upload_to method that uses the collection name AND the person uploading the image only has write access to a single collection. When a user only has one collection, the image upload form removes the collection field from the form and then reassigns it in `BaseCollectionMemberForm#save`. 

The current save_object method of CreateFromUploadedImageView calls file.save(...save=False) to move the file data from UploadedFile to the site's image model. For my situation, that causes `get_upload_to` to be called before `BaseCollectionMemberForm#save` assigns the user's single collection to the image instance. This patch assigns the file name and file data to the image instance without the use of file.save and so defers calling get_upload_to until after the correct collection is available in the image instance. This makes the multiple upload path behave more like the single upload path with all the form saving deferred until the end.

* For front-end changes: Did you test on all of Wagtail’s supported browsers? no frontend changes in this PR
* For new features: Has the documentation been updated accordingly? no new features in this PR

### DIFF
--- a/wagtail/images/views/multiple.py
+++ b/wagtail/images/views/multiple.py
@@ -114,7 +114,8 @@ class CreateFromUploadedImageView(BaseCreateFromUploadView):
         # assign the file content from uploaded_image to the image object, to ensure it gets saved to
         # Image's storage
 
-        self.object.file.save(os.path.basename(self.upload.file.name), self.upload.file.file, save=False)
+        self.object.file = self.upload.file.file
+        self.object.file.name = os.path.basename(self.upload.file.name)
         self.object.uploaded_by_user = self.request.user
         self.object.file_size = self.object.file.size
         self.object.file.open()


gasman:
Thanks @cnk! Just tested with the setup you describe, and this fix is working well (although I still need to get my head around the exact sequence of events that's being changed here, to be sure that we aren't introducing a regression in some other setup...)

Just to clarify - are you sure this is a regression in Wagtail 2.11? I'm seeing the same broken behaviour in 2.10, and can't see any changes to the code that would contribute to this (just the change to support nested collections, which doesn't seem to be directly related). I suspect it's behaved this way ever since we introduced the UploadedImage model and started supporting required fields on image models - in which case this fix probably wouldn't be a candidate for back-porting.

cnk:
I suspect the code has behaved this way ever since we introduced UploadedImage as a way to handle required fields in image models. I first noticed it when we upgraded to 2.11 but did not have time to characterize the issue so added a monkey patch to my code to work around it. And I now have this fix monkey patched into the 2.12.x upgrade branch for my project. So if this only goes into 2.13, that's fine for me.

cnk:
It does not trigger in my work project, but I notice that the same issue will likely apply to documents saved via UploadedDocument. I will mark this PR as draft until I have had a chance to update it to do the same thing to that code

cnk:
Upgraded to 2.12.3 and now I cannot reproduce this issue. Closing this PR.

PR #6718, opened 2021-01-22
Consistently pass validation kwargs from StreamField to the top-level block

### LINKED ISSUE #4306: StreamField and StreamBlock have confusing behaviour with `required` and `blank` properties

gasman:
Fixes #4306 and #5175; supersedes (and incorporates tests from) #5185:

`min_num`, `max_num` and `block_counts` can now be set on the StreamField constructor, allowing definitions like

    body = StreamField([
        ('paragraph', RichTextBlock()),
    ], min_num=3)

where previously it was necessary to use an explicit top-level StreamBlock to hang the `min_num` argument on. The kwargs on StreamField take precedence over any `min_num`, `max_num`, or `block_counts` meta attributes set on the top-level block (at the class or instance level).

The `blank` kwarg is also now propagated to the top-level block's `required` attribute, avoiding the situation (described in #4306) where the former controlled the asterisk on the form field but the latter controlled the actual validation. The default of `blank=False` is respected - even if `blank` is omitted - meaning that the initial `required` meta attribute on the top-level block is always ignored. This matches what the documentation always said was supposed to happen (but actually only happened in the specific setup where you pass a block class: `body = StreamField(MyStreamBlock)`).

These changes mean that we have to relax the rule that meta properties are considered immutable (because things like `StreamField(MyStreamBlock(), min_num=3)` need to set `min_num` on the `MyStreamBlock` instance that already exists by that point) - the definition of `__eq__` on blocks has been tweaked accordingly.


docs/topics/streamfield.rst
```diff
@@ -47,7 +47,24 @@ The parameter to ``StreamField`` is a list of ``(name, block_type)`` tuples. 'na
 
 This defines the set of available block types that can be used within this field. The author of the page is free to use these blocks as many times as desired, in any order.
 
-``StreamField`` also accepts an optional keyword argument ``blank``, defaulting to false; when this is false, at least one block must be provided for the field to be considered valid.
+``StreamField`` also accepts the following optional keyword arguments:
+
+``blank`` (default: False)
```
kaedroho, line 52: ``False`` should be backticked

docs/topics/streamfield.rst
```diff
@@ -47,7 +47,24 @@ The parameter to ``StreamField`` is a list of ``(name, block_type)`` tuples. 'na
...
+
+``max_num``
+  Maximum number of sub-blocks that the stream may have.
+
+``block_counts``
+  Specifies the minimum and maximum number of each block type, as a dictionary mapping block names to dicts with (optional) ``min_num`` and ``max_num`` fields.
+
+.. versionadded:: 2.13
+
+    The ``min_num``, ``max_num`` and ``block_counts`` arguments were added. Previously, these were only available on the StreamBlock definition.
```
kaedroho, line 66: ``StreamBlock`` should be backticked

docs/topics/streamfield.rst
```diff
@@ -47,7 +47,24 @@ The parameter to ``StreamField`` is a list of ``(name, block_type)`` tuples. 'na
...
+  When false, at least one block must be provided for the field to be considered valid.
+
+``min_num``
+  Minimum number of sub-blocks that the stream must have.
+
+``max_num``
+  Maximum number of sub-blocks that the stream may have.
+
+``block_counts``
+  Specifies the minimum and maximum number of each block type, as a dictionary mapping block names to dicts with (optional) ``min_num`` and ``max_num`` fields.
```
kaedroho, line 62: I think an example would definitely help with this one

wagtail/tests/testapp/models.py
```diff
@@ -1007,12 +1007,38 @@ class StreamModel(models.Model):
...
+    )
+
+
+class BlockCountsStreamModel(models.Model):
+    body = StreamField([
+        ('text', CharBlock()),
+        ('rich_text', RichTextBlock()),
+        ('image', ImageChooserBlock()),
+    ],
+        block_counts={
```
kaedroho, line 1027: There's a couple of indentation issues here, surprised that flake8 didn't catch this!

wagtail/core/blocks/base.py
```diff
@@ -121,6 +127,16 @@ def set_name(self, name):
         if not self.meta.label:
             self.label = capfirst(force_str(name).replace('_', ' '))
 
+    def set_meta_options(self, options):
```
kaedroho, line 130: Could this have a more specific name to make it clear it's only for the top-loevel block? eg ``set_root_block_meta_options``

wagtail/core/fields.py
```diff
@@ -51,13 +51,31 @@ def __set__(self, obj, value):
...
         elif isinstance(block_types, type):
-            self.stream_block = block_types(required=not self.blank)
+            # block passed as a class - instantiate it
+            self.stream_block = block_types()
         else:
-            self.stream_block = StreamBlock(block_types, required=not self.blank)
+            # construct a top-level StreamBlock from the list of block types
+            self.stream_block = StreamBlock(block_types)
+
+        self.stream_block.set_meta_options(block_opts)
```
kaedroho, line 78: What if the options were specificed that aren't supported by the root block?

Maybe we should pass them as kwargs here so a ``TypeError`` is raised?

kaedroho (changes requested):
Just some minor tweaks (some are debatable), but otherwise looks good to me

docs/topics/streamfield.rst
gasman: I think we need to split this page into separate tutorial/howto and reference pages before we can do that - hopefully that's something the upcoming documentation sprint can take care of. Introducing these options before we've explained the available block types already feels a bit clunky, and I think the best compromise is to get through the 'reference-y' bits as quickly as possible.

wagtail/core/blocks/base.py
gasman: Actually I think the comment is the thing that needs updating here... initially I'd called this something like `set_options_from_field`, with the plan that the `StreamField` constructor wouldn't care about the semantics of the additional kwargs, and would just pass them here unfiltered (e.g. `blank=True` rather than `required=False`) for the block to use as it sees fit. However, that didn't work out because `StreamField` needs to pull out the StreamBlock-related kwargs before calling `super()` - so at that point it made sense for `StreamField` to translate them into StreamBlock terms (`blank` -> `required`). As a result, there's nothing field-specific here any more, so no reason why it couldn't be used for other blocks besides the top-level one.

wagtail/core/fields.py
gasman: Passing as kwargs is messy because we need to distinguish between `max_num=None` (allow unlimited blocks, overriding any maximum set at the class level) and not passing `max_num` (leave the setting unchanged from what the class definition says). Best to cross that bridge when we start supporting other top-level block types besides StreamBlock, I think...

gasman: Actually, scratch that - I now see that `StreamBlock.set_meta_options` is really just duplicating the information in `MUTABLE_META_ATTRIBUTES`, so we can make the base `set_meta_options` method a bit smarter and fix this and the code comment issue at the same time :-)

gasman:
@kaedroho Updated!

kaedroho (approved):

gasman:
Merged in df125b165209d4c80082cff5502c4e22dc3ad3fd + parents.

PR #6721, opened 2021-01-22
Telepath: include help text, required asterisk and label 'for' attribute in rendering

gasman:


wagtail/core/blocks/base.py
```diff
@@ -512,6 +512,18 @@ def has_changed(self, initial_value, data_value):
         return self.block.get_prep_value(initial_value) != self.block.get_prep_value(data_value)
 
 
+_help_icon = None
+
+
+def get_help_icon():
```
kaedroho, line 518: Would be slightly neater to decorate this with `@cache`

```python
from functools import lru_cache

@lru_cache(maxsize=1)
def get_help_icon():
    return render_to_string("wagtailadmin/shared/icon.html", {
        'name': 'help', 'class_name': 'default'
    })
```

kaedroho (approved):

gasman:
Merged in 1e41fa1587af0eb90fb459c15bfaeac1d3d7c17e.

PR #6723, opened 2021-01-23
added tel type to form pages

### LINKED ISSUE #6611: Add field type "tel" to form pages

Currently, when an editor is creating a form using the form page, they need to use the "number" field type (`<input type="number">`) if they would like to ask for telephone or cellphone numbers. This makes it difficult for users of the form who are using a mobile devices because `type="tel"` and `type="number"` pull up different keyboards on mobile devices. (Sample images of the keyboards below.)

I would like for the "tel" field to be added to the list of `FORM_FIELD_CHOICES` ([relevant code](https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/forms/models.py)). This change will likely need some migrations, I don't think it would require new HTML files.

The alternative to this would likely be on the project side– creating subclasses of the abstract form field and the abstract form to add "tel", which isn't ideal.

### Additional context

- Browsers that don't support "tel" will fall back to being a "text" input.
- "tel" does not automatically validate to a format because phone number formats vary a lot. (Maybe only for now?)
- Semantic HTML == better :)

Sample images of the keyboards on iOS:
IMG_7460 IMG_7459

adenhaus:
# Added "tel" field to form pages

Added "tel" field to FORM_FIELD_CHOICES in https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/forms/models.py
Added a `create_tel_function` in https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/forms/forms.py
Created a custom TelephoneInput widget in utils/widgets.py
Added the necessary migrations to a number of files in wagtail/tests/testapp/migrations

Added this to the changelog and added myself as a contributor.

Closes #6611


adenhaus:
Okay I see the PR has failed some tests because of migrations, can someone please run through how I can solve those issues? I'm a new contributor, thanks!

gasman:
Hi @adenhaus - thanks for the PR!
Did you test this change? It looks like there ought to be a corresponding `create_tel_field` method in https://github.com/wagtail/wagtail/blob/master/wagtail/contrib/forms/forms.py.

adenhaus:
Hi @gasman
Right. Sorry about that, I am working on adding it.
However, I have a question about creating a create_tel_field function:
With others, for example create_number_field, there is a django.forms.DecimalField already, but there is no field for telephone numbers. The list of fields are here: https://docs.djangoproject.com/en/3.1/ref/forms/fields/#combofield
So, I have found a possible solution. Pypi's django-phonenumber-field, which can be seen here: https://pypi.org/project/django-phonenumber-field/
It seems like it will provide the necessary field and act just like the others, but it will need to be installed.
Can I proceed to implement this?

adenhaus:
Hey again @gasman I added the migrations and all tests pass.

gasman:
Thanks for the updates!
For something that's only going to be used infrequently, the dependency on django-phonenumber-field is a bit heavyweight - we're only really interested in setting the `type="tel"` attribute, not the national phone number validation that python-phonenumbers gives us. I think it should be possible to do what we need by defining a custom widget type for type="tel" (this should only take a couple of lines, [like the ones built into Django](https://github.com/django/django/blob/2b4b6c8af0aae8785bc1347cf1be2e8e70fd5ff3/django/forms/widgets.py#L308-L310)) and passing that in create_tel_field (much like create_multiline_field does).

adenhaus:
Okay @gasman happy to try that. Alternatively, there is a lightweight version of `django-phonenumber-field` which we could use instead?
Otherwise, should `create_tel_field` return `django.forms.CharField(widget=django.forms.Textarea, **options)` like `create_multiline_field`?

gasman:
`create_tel_field` will return something like `django.forms.CharField(widget=TelephoneInput, **options)`, where `TelephoneInput` is the custom widget type you're going to create.

adenhaus:
Right, thanks for the guidance! Done.

wagtail/contrib/forms/forms.py
```diff
@@ -47,6 +48,9 @@ def create_url_field(self, field, options):
     def create_number_field(self, field, options):
         return django.forms.DecimalField(**options)
 
+    def create_tel_field(self, field, options):
+        return django.forms.CharField(widget=TelephoneInput, **options)
```
kaedroho, line 52: I really think a phone number field should provide a way to add validation so that only phone numbers for a particular country can be inserted.

Would it be possible to allow people to use ``PhoneNumberField`` from [``django-phonenumber-field``](https://github.com/stefanfoulis/django-phonenumber-field) here?

adenhaus: Hi @kaedroho, that's exactly what I did originally, but if you look a bit earlier in this conversation, you'll see @gasman said that was too heavyweight for this purpose and that we don't need the validation, just the `type="tel"` attribute, and asked me to make a custom widget instead.

adenhaus: > Thanks for the updates!
> For something that's only going to be used infrequently, the dependency on django-phonenumber-field is a bit heavyweight - we're only really interested in setting the `type="tel"` attribute, not the national phone number validation that python-phonenumbers gives us. I think it should be possible to do what we need by defining a custom widget type for type="tel" (this should only take a couple of lines, [like the ones built into Django](https://github.com/django/django/blob/2b4b6c8af0aae8785bc1347cf1be2e8e70fd5ff3/django/forms/widgets.py#L308-L310)) and passing that in create_tel_field (much like create_multiline_field does).

kaedroho: Does `type="tel"` provide any validation if you don't specify a pattern? According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/tel), it behaves the same as a text field on desktop, but on mobile phones it shows a keypad when you edit it.

gasman: @kaedroho Yes, that's exactly what was requested in #6611.

kaedroho: @gasman sorry I can't see any mention of using `pattern` there?

adenhaus: @kaedroho as I understand it, it has nothing to do with `pattern` or validation. The problem was that  `<type="number">` was bringing up the wrong keyboard on mobile, so the addition of `<type="tel">` is just intended to make the correct keyboard display. That's all.

kaedroho: If we provide a "telephone number" option, wouldn't user's expect that to include some sort of validation?

adenhaus: I think that might be beyond the scope of what I can answer... Perhaps @gasman or @SharmaineLim (who created the issue #6611) can decide which approach is best. If you want validation, you can use https://github.com/wagtail/wagtail/pull/6723/commits/c8420ff62a09328bd1148d85af762ff5ec59aaee (which builds on changes made in https://github.com/wagtail/wagtail/pull/6723/commits/c249429d383ae44ee0311bebe9d8853a70ad80be), if you don't want validation, you can use https://github.com/wagtail/wagtail/pull/6723/commits/77bf795352f09e2f6271db97a969d97113f4317b.

kaedroho: We discussed this in todays core team meeting. Here's an approach that we've come up with:

 - Add a setting to allow a regex pattern to be specified for phone number validation
 - Add a warning into the form builder docs and in the 2.13 upgrade considerations to warn developers that there will be no phone number validation if this setting isn't used

Let us know what you think! Sorry for all the back and forth on this

adenhaus: Okay cool, will do!

adenhaus: Hi @kaedroho I would like to finish this, but could I get some guidance on how to go about it please?

lb-:
Thanks @adenhaus - sorry this ended up being a bit more complex than originally thought.
I will close this PR now as it's been a long while but if you are keen on giving this another go, please reach out on the issue.

PR #6727, opened 2021-01-26
Telepath: Merge ListChild and StreamChild into a BaseSequenceChild class

kaedroho:
This is the first part of refactoring ListBlock/StreamBlock to share most of their implementation


gasman (approved):
Looks good to me!
With the addition of the type field to the ListBlock rendering (and, particularly, adding `${this.blockDef.meta.label}` to the header, which will generally be redundant since ListBlock children typically don't have names or labels) I'm slightly wary that we're artificially making the HTML of ListBlock and StreamBlock identical, when we'd be better served by making the common base class general enough to accommodate differences between the two. (Compare the old implementation, where the sequence_member.html template defined blocks for the subclasses to insert their own hidden fields and header/footer controls.)

Having said that, I don't think that's a reason to hold this up... sharing the same HTML clearly works for now (and actually improves ListBlock, with the ability to insert a block anywhere...) so happy to go ahead with this, and revisit the ability to customise the base HTML as and when the need for it comes up.

PR #6738, opened 2021-01-28
Telepath: Break up blocks.py into multiple files

kaedroho:
This moves the block components and definition classes into ``components/StreamField/blocks``. It also splits up the tests into separate files.

I rewrote the block definitions in tests to instantiate the definition classes directly instead of using telepath (so the blocks themselves don't actually depend on telepath, but it might also be helpful for type checking if we add that later).


gasman (approved):
LGTM! Making the tests not depend on telepath for object instantiation is definitely a good move.

Issue #6737, opened 2021-01-28
Image chooser modal form fields aren’t translatable


thibaudcolas:
### Issue Summary

In the image chooser modal, some of the fields are missing translations due to them not being marked for translations:

- The Collection field on the Upload form. I wouldn’t be surprised if this was the case for all Upload forms, as well as the one for documents, as they all inherit from the same base form.
- The Format and Alt text fields on the `ImageInsertionForm`.

![translations-chooser-upload](https://user-images.githubusercontent.com/877585/106126793-c038e280-6155-11eb-86cd-2d9247890592.png)

![translations-chooser-format](https://user-images.githubusercontent.com/877585/106126803-c3cc6980-6155-11eb-90c7-778e9fe2cac8.png)

### Steps to Reproduce

1. Use Wagtail in any other language than english
2. Open the image chooser modal from a rich text editor
3. Note the field labels are in english.

Any other relevant information. For example, why do you consider this a bug and what did you expect to happen instead?

All of Wagtail’s admin UI defined in core should be translate-able.

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: no

### Technical details

* Wagtail version: Wagtail v.2.13a0

thibaudcolas:
Fixed in #6730 for the "select format" step fields. The Collection issue is more widespread so I’ll open a separate PR just for that.

PR #6739, opened 2021-01-29
Make all Collection form fields translate-able

thibaudcolas:
Initially reported in #6737 for the image chooser, but I then realised the issue was more widespread. The "Collection" field wasn’t set up for translations in:

- Document upload form
- Image edit form
- Image multi-upload form
- Image chooser upload form
- Document chooser upload form

The fix is just to override the auto-generated label with our translate-able equivalent, just like we do for other form fields.

Additionally, the "Parent" field also needed the same treatment in the collection  create/edit form.

I’ve only tested this in Chrome by navigating to some of the relevant forms. Existing translations already have a "collection" string, so the translated label should appear without further work. For "Parent", this will need to be translated in Transifex.

### DIFF
--- a/wagtail/admin/forms/collections.py
+++ b/wagtail/admin/forms/collections.py
@@ -68,6 +68,7 @@ class CollectionChoiceField(forms.ModelChoiceField):
 
 class CollectionForm(forms.ModelForm):
     parent = CollectionChoiceField(
+        label=_("Parent"),
         queryset=Collection.objects.all(),
         required=False,
         help_text=_(
@@ -279,6 +280,7 @@ def collection_member_permission_formset_factory(
         (i.e. group or user) for a specific collection
         """
         collection = CollectionChoiceField(
+            label=_("Collection"),
             queryset=Collection.objects.all().prefetch_related('group_permissions'),
             empty_label=None
         )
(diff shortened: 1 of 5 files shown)


wagtail/admin/forms/collections.py
```diff
@@ -68,6 +68,7 @@ def label_from_instance(self, obj):
 
 class CollectionForm(forms.ModelForm):
     parent = CollectionChoiceField(
+        label=_("Parent"),
```
kaedroho, line 71: Looks like this file imports ``gettext`` as ``_``. In this instance, we need to use ``gettext_lazy`` or this would be evaluated on startup.

kaedroho (changes requested):
Thanks! Looks good apart from we should use ``gettext_lazy`` for the change in ``collections.py``

wagtail/admin/forms/collections.py
thibaudcolas: Ah nice catch, completely missed that.

Thank you for the review!

thibaudcolas:
@kaedroho this should be ready for another review when you have the chance. Right below the label you flagged, there was a `help_text` using the eager `gettext`, so I took the liberty to also switch that to `gettext_lazy`.

kaedroho (approved):
Thanks @thibaudcolas, this looks good to me!

thibaudcolas:
🤘 thank you, adding release notes and merging this now.

PR #6740, opened 2021-01-29
Improve wagtailforms accessibility guidance

thibaudcolas:
Rewording the existing content so the guidance is more actionable, and adding a new line item about Number fields.

This is a pretty small docs-only change so I will merge later today without further review.

Review link: https://wagtail--6740.org.readthedocs.build/en/6740/advanced_topics/accessibility_considerations.html


helenb:
Changes LGTM :)

PR #6741, opened 2021-01-29
Defer rendering StreamBlock menu until opened

gasman:
A small optimisation for large StreamBlocks - most stream menus will never be expanded, so it's wasted work to render the menu contents up-front.


gasman:
Merged via #6742

PR #6743, opened 2021-02-01
Telepath: Backend tests for structural block types

kaedroho:


wagtail/core/tests/test_blocks.py
```diff
@@ -1710,21 +1644,19 @@ class Meta:
...
-            'title': "Wagtail site",
-            'link': 'http://www.wagtail.io',
-        }), prefix='mylink')
 
-        self.assertInHTML('<div class="help"> <svg class="icon icon-help default" aria-hidden="true" focusable="false"><use href="#icon-help"></use></svg>  Self-promotion is encouraged</div>', html)
+        block.set_name('test_structblock')
+        js_args = StructBlockAdapter().js_args(block)
 
-        # check it can be overridden in the block constructor
-        block = LinkBlock(help_text="Self-promotion is discouraged")
```
gasman, line 1721: It would be good to include the equivalent of this in the new test - i.e. confirm that when `LinkBlock(help_text="Self-promotion is discouraged")` is passed to `StructBlockAdapter.js_args`, the overridden help text comes through in the result.

gasman (changes requested):

wagtail/core/tests/test_blocks.py
kaedroho: Added, thanks!

gasman (approved):
LGTM! (Tests are now failing due to the addition of the strings dict in #6747, but I'm happy to fix that and merge)

gasman:
Merged in dcd2322fc8f22f0498d8c4707866cce22a2923b6. (Along the way I noticed that `TestStreamBlock.test_adapt` had got cut off somehow, but I filled in the gap...

PR #6753, opened 2021-02-02
WIP: Try out DocSearch integration

thibaudcolas:
**Do not merge**. Quick & dirty prototyping implementation so we can see how DocSearch works.

Online demo: https://wagtail--6753.org.readthedocs.build/en/6753/

### DIFF
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -5,8 +5,14 @@ body.wy-body-for-nav {
     background-color: #e6e6e6; /*$color-grey-4*/
 }
 
-div.wy-nav-side {
+nav.wy-nav-side {
     background-color: #333;  /*color-grey-1*/
+    position: absolute;
+    overflow: inherit;
+}
+
+div.wy-side-scroll {
+    overflow: inherit;
 }
 
 div.wy-side-nav-search {
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -112,8 +112,3 @@ a.wagtailspace:hover svg {
   -webkit-animation-name: bounce;
   animation-name: bounce;
 }
-
-.algolia-autocomplete .algolia-docsearch-suggestion--highlight {
-  color: #2980B9;
-  background: #F1C40F;
-}
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -112,3 +112,8 @@ a.wagtailspace:hover svg {
   -webkit-animation-name: bounce;
   animation-name: bounce;
 }
+
+.algolia-autocomplete .algolia-docsearch-suggestion--highlight {
+  color: #2980B9;
+  background: #F1C40F;
+}
--- a/docs/_static/css/docsearch.overrides.css
+++ b/docs/_static/css/docsearch.overrides.css
@@ -1,0 +1,4 @@
+.algolia-autocomplete .algolia-docsearch-suggestion--highlight {
+  color: #2980B9;
+  background: #F1C40F;
+}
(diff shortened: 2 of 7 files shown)


chosak:
@thibaudcolas I've pushed a few commits that do the following:

- Make it possible to run locally (since `window.READTHEDOCS_DATA` isn't defined)
- Allow the sidebar to scroll, so that the dropdown isn't cut off (thanks to @Scotchester for help with this). This is temporary for the purposes of exploring/demoing this capability. It's a little tricky to get the popup to appear while still preserving the RTD sidebar scroll, and, if we might be migrating to a new design with a non-scrolling sidebar, this seemed like a reasonable change to make for now. (This also doesn't work well on mobile right now.)
- Experiment with customizing the dropdown styles, adding highlighting from our [current search results page](https://docs.wagtail.io/en/stable/search.html?q=test&check_keywords=yes&area=default). Further docs for how to do this are on Algolia's site [here](https://docsearch.algolia.com/docs/styling).

The biggest limitation I see so far with DocSearch is I can't see how we might setup a search results page. This is not only useful to allow people to browse the full set of search results, but also (as mentioned by @tomdyson) provides a fixed URL to link to at times. This is mentioned at https://github.com/algolia/docsearch/issues/424.

chosak:
Gitlab is listed as one of the users on the [DocSearch home page](https://docsearch.algolia.com/), and they have a [search results page](https://docs.gitlab.com/search/?query=test). Happily they have [some documentation](https://gitlab.com/gitlab-org/gitlab/-/blob/master/doc/development/documentation/site_architecture/index.md#algolia-search-engine) on their DocSearch integration, and its code is open source: [template](https://gitlab.com/gitlab-org/gitlab-docs/blob/master/layouts/instantsearch.html), [JS](https://gitlab.com/gitlab-org/gitlab-docs/-/blob/master/content/frontend/search/index.js). This looks promising -- I'll see if I can try something similar here.

thibaudcolas:
Note: need to evaluate whether the results are indeed better! For example `WAGTAILIMAGES_INDEX_PAGE_SIZE`.

tbrlpld:
> Note: need to evaluate whether the results are indeed better! For example `WAGTAILIMAGES_INDEX_PAGE_SIZE`.

When I search something like "page" or "pages", then the result of the Algolia search is much more what I would expect/find helpful.

tbrlpld:
Ok, looks like the code blocks are ignored on purpose. 

> Can you index code samples?
>
> Yes, but we do not recommend it.
> 
> Code samples are a great way for humans to understand how people use a specific method. It often requires boilerplate code though, repeated across examples, which adds noise to the results.
> 
> What we recommend instead is to exclude the code blocks from the indexing (by using the selectors_exclude option in your config), and instead structure your content so the method names are present in the headers.

-- https://docsearch.algolia.com/docs/faq/#can-you-index-code-samples

I guess that makes sense. Adding more headings for things like the settings also makes sense, because then we make it possible to link to it.

thibaudcolas:
👍  that is likely the issue with `WAGTAILIMAGES_INDEX_PAGE_SIZE`, it’s in a code block only. You’re right, we’d likely want to have one heading per setting for linking purposes.

For "pages", the results are what I’d expect so I don’t see the issue? The previous search had 50 results, and ranked "Private Pages" higher than "Pages".

I’ll see if I can figure out how to get the most popular searches from our current analytics, so we can compare the results before/after.

tbrlpld:
Sorry, yes that is what I mean. The Algolia search returns must better results on things like "pages".

thibaudcolas:
Closing in favour of #6833.

PR #6755, opened 2021-02-02
Fix: 500 error when searching in admin

cspollar:
# Current behavior:
Malformed query strings cause the app to crash while filtering search results in the admin.

Manually adding a period to the `content_type` in the URL's query string will cause the admin site to 500.

## Working Query

http://127.0.01:8000/admin/pages/search/?content_type=demosite.standardpage

## Broken Query

http://127.0.01:8000/admin/pages/search/?content_type=demosite.standardpage.broken

# Expected behavior:
Malformed URLS should 404.

# Cause

Too many values to unpack.

`app_label, model_name = request.GET['content_type'].split('.')`

Source: https://github.com/wagtail/wagtail/blob/main/wagtail/admin/views/pages/search.py#L46

# Fix:
Add a try/except block around `app_label, model_name = request.GET['content_type'].split('.')`

# Reproduce

This is reproducible on the demo site by using the following path.
`http://127.0.01:8000/admin/pages/search/?content_type=demosite.standardpage.error`

----

### DIFF
--- a/wagtail/admin/tests/pages/test_page_search.py
+++ b/wagtail/admin/tests/pages/test_page_search.py
@@ -175,3 +175,12 @@ class TestPageSearch(TestCase, WagtailTestUtils):
         response = self.get({'q': 'Lunar', 'ordering': '-live'})
         page_ids = [page.id for page in response.context['pages']]
         self.assertEqual(page_ids, [live_event.id, draft_event.id])
+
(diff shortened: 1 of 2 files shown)


gasman (approved):

gasman:
Merged in 5f7181e05d402c5c71852ce5dfd03c98aa6caebf - thanks @cspollar!

PR #6756, opened 2021-02-03
Publish aliases when the source page is published

### LINKED ISSUE #6749: Draft state is not removed from page aliases when the source page is published
### Issue Summary

There’s an issue with page aliases, which are used by wagtail-localize. When a source page goes to draft state at one point then gets published, the page aliases (used for each locale until there is an actual translation) remain in draft state with no way to publish them. Wagtail should propagate the publish state to the page aliases as well.

cc @kaedroho 

### Steps to Reproduce

1. Set up wagtail-localize
2. Add at least one locale to Settings > Locales
3. Save a new page as draft or Unpublish an existing page
4. Publish the page again

At step 3, the `Draft` state gets propagated to the page alias in the French tree. (works as expected)
At step 4, the `Draft` state is NOT removed from the page alias (broken)

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: No, but I’ve reproduced locally with a test DB.

### Technical details

* Python version: 3.7.8
* Django version: 2.2.13
* Wagtail version: 2.11.2
* Wagtail-localize version: 0.9.3

kaedroho:
Fixes #6749

Aliases are currently not being published if they were created in draft

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -1469,6 +1469,10 @@ class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):
             # Copy field content
             alias_updated = alias.with_content_json(_content_json)
 
+            # Publish the alias if it's currently in draft
+            alias_updated.live = True
+            alias_updated.has_unpublished_changes = False
+
             # Copy child relations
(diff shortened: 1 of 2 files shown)


gasman (approved):

kaedroho:
Thanks for the quick review @gasman!

PR #6757, opened 2021-02-03
Make privacy rules also apply to aliases

### LINKED ISSUE #6750: Page aliases are not inheriting the source page privacy settings
### Issue Summary

This issue may be related to #6749
When using page aliases (e.g. with wagtail-localize), the source page privacy settings are not reflected on the page aliases. For instance, setting a source page in English to be password-protected will still be accessible in other languages

cc @kaedroho 

### Steps to Reproduce

1. Set up wagtail-localize
2. Add at least one locale to Settings > Locales
3. Set a password to a source page 
4. Open /*your locale*/*your page slug*/, the page is  still public.

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: No, but I’ve reproduced locally with a test DB.

### Technical details

* Python version: 3.7.8
* Django version: 2.2.13
* Wagtail version: 2.11.2
* Wagtail-localize version: 0.9.3

kaedroho:
Fixes #6750

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -2690,7 +2690,7 @@ class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):
         add_page_to_check_list(self)
 
         # Check each ancestor for view restrictions as well
-        for page in self.get_ancestors():
+        for page in self.get_ancestors().only('alias_of'):
             add_page_to_check_list(page)
 
         return PageViewRestriction.objects.filter(page_id__in=page_ids_to_check)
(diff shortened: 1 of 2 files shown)


wagtail/core/models.py
```diff
@@ -2668,8 +2668,32 @@ def get_prev_siblings(self, inclusive=False):
...
+            if page.alias_of:
+                add_page_to_check_list(page.alias_of)
+            else:
+                page_ids_to_check.add(page.id)
+
+        # Check current page for view restrictions
+        add_page_to_check_list(self)
+
+        # Check each ancestor for view restrictions as well
+        for page in self.get_ancestors():
```
gasman, line 2693: Since this query is fairly performance-critical, could we change this to:
```suggestion
        for page in self.get_ancestors().only('alias_of'):
```

gasman (changes requested):
Just one performance-related suggestion.

To confirm: are we happy that view restrictions on ancestors of the page being aliased do NOT get passed on? For example, if secret-organisation is restricted, and public-organisation/news is an alias of secret-organisation/news, then public-organisation/news/product-launch? I think it does make sense that way (if we consider aliasing to be a lightweight 'copy' operation), but figured I'd better check.

gasman (approved):

PR #6760, opened 2021-02-03
Telepath: Animate new list block items

kaedroho:


gasman (approved):
Nice! That was easier than I expected :-)

Issue #6761, opened 2021-02-03
Reordered pages not logged properly


Stormheg:
### Issue Summary

Reordering a page shows up in Wagtail's Site History report as having been performed by the 'system' user. Instead of the user that has actually performed the action.

The description of the reordering action could also be improved upon. 'Moved from [parent page] to [parent page].' is not very clear to the end user. '[reordered page] was reordered.' or something like that would seem clearer to me (suggestions welcome).

Screenshot of the issue:
![Screenshot from 2021-02-03 16-50-21](https://user-images.githubusercontent.com/13856515/106772089-f38cdd00-663f-11eb-8de9-1295869e6a82.png)

### Steps to Reproduce

1. Install [Wagtail Bakerydemo](https://github.com/wagtail/bakerydemo) and update it to Wagtail 2.12 (The bakerydemo is still on 2.11 as of writing).
2. Go to 'Breads' page in the Wagtail admin.
3. Click the 'Sort' link (screenshot included). ![Screenshot from 2021-02-03 16-56-22](https://user-images.githubusercontent.com/13856515/106773262-2b485480-6641-11eb-81f1-2426aa1948ed.png)
4. Try dragging a page to a new position. Move the Bagel page to the top for example :).
5. Now go to the Site History report (`Reports -> Site History` in the menu bar).

The report should show your action as having been performed by the 'system' user. It should also show the confusing description as stated in the issue summary.

**Expectation:** the report should show the actual user who has performed the action, not 'system'.

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: yes

### Technical details

* Python version: 3.9.1.
* Django version: 3.1.6.
* Wagtail version: 2.12 final
* Browser version: Chrome 88 on Fedora Linux.

Stormheg:
The offending code is in [`wagtail/admin/views/pages/ordering.py`](https://github.com/wagtail/wagtail/blob/master/wagtail/admin/views/pages/ordering.py#L8).

`set_page_position()` should pass the current user to the `page_to_move.move()`.

zerolab:
Nice find @Stormheg
Thank you in advance for the PR. I think pos `left` and `right` can count as re-order rather than move and I wonder if the log message should reflect that.

Stormheg:
[...]

PR #6762, opened 2021-02-03
Add `wagtail.reorder` audit log action

Stormheg:
Include `request.user` so `Page.move()` can include it in it's log entry. This will prevent the log entry from having no acting user associated with it (#6761).

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -2062,7 +2062,8 @@ class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):
         # Log
         PageLogEntry.objects.log_action(
             instance=self,
-            action='wagtail.move',
+            # Check if page was reordered (reordering doesn't change the parent)
+            action='wagtail.reorder' if parent_before.id == target.id else 'wagtail.move',
             user=user,
             data={
                 'source': {
--- a/docs/advanced_topics/audit_log.rst
+++ b/docs/advanced_topics/audit_log.rst
@@ -72,6 +72,7 @@ Action                               Notes
 ``wagtail.revert``                   The page was reverted to a previous draft
 ``wagtail.copy``                     The page was copied to a new location
 ``wagtail.move``                     The page was moved to a new location
+``wagtail.reorder``                  The order of the page under it's parent was changed
 ``wagtail.view_restriction.create``  The page was restricted
 ``wagtail.view_restriction.edit``    The page restrictions were updated
 ``wagtail.view_restriction.delete``  The page restrictions were removed
--- a/wagtail/admin/views/pages/ordering.py
+++ b/wagtail/admin/views/pages/ordering.py
@@ -34,11 +34,11 @@ def set_page_position(request, page_to_move_id):
             # right. If left, then left.
             old_position = list(parent_page.get_children()).index(page_to_move)
             if int(position) < old_position:
-                page_to_move.move(position_page, pos='left')
+                page_to_move.move(position_page, pos='left', user=request.user)
             elif int(position) > old_position:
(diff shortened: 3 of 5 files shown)


Stormheg:
Should there be tests for this?

We currently only test if log entries get created when calling `Page.move()`  (we don't test if they have the proper user set here) https://github.com/wagtail/wagtail/blob/master/wagtail/core/tests/test_audit_log.py#L194

Perhaps we should test in the admin if a `PageLogEntry` with the proper user is created? I think the right place for that would be in this test: https://github.com/wagtail/wagtail/blob/master/wagtail/admin/tests/pages/test_move_page.py#L148
This would have caught `set_page_position()` not supplying the correct user to `Page.move()`.

Any input?

zerolab:
@Stormheg the test should go in `test_audit_log`'s `test_page_move`

Stormheg:
@zerolab I've added the test. Thank you for your guidance. :slightly_smiling_face: 

This is ready for review.

Stormheg:
@zerolab Is it okay if I assign you to review? Feel free to reassign to someone else.

zerolab:
Hey @Stormheg, the changes look good to me as they are, however having had time to think of it, the log messages are misleading in this case in that this is not a move per se, but a re-order.

2 options:
1. change to not log anything
2. update https://github.com/wagtail/wagtail/blob/master/wagtail/admin/wagtail_hooks.py#L861 to check if `                'data['source']['id']` is the same as 'data['destination']['id']` and tweak the message to say "Reordered under %(parent_title)s" or similar

Stormheg:
Lets go with option 2 and add a proper log message.

Stormheg:
@zerolab thanks, I've added the log message. Could you review this again please?

zerolab (commented):
That reads/feels so much better, thank you Storm.
As noted in Slack, would be nice to have a test for the reorder. Then we're good to go

Stormheg:
@zerolab and I discussed this on Slack. We've decided that adding a separate `wagtail.reorder` action is the best option for the following reasons:

* It allows for filtering on reorder log entries with `PageLogEntry.objects.filter(action='wagtail.reorder')`.
* It allows us to easily test the reorder by checking if `PageLogEntry(action='wagtail.reorder')` objects are created.

wagtail/core/models.py
```diff
@@ -2060,21 +2060,40 @@ def move(self, target, pos=None, user=None):
...
+                        'id': parent_after.id,
+                        'title': parent_after.specific_deferred.get_admin_display_title()
+                    }
                 }
-            }
-        )
+            )
+        # Page was moved
+        else:
+            PageLogEntry.objects.log_action(
```
zerolab, line 2082: The two `log_action` calls are essentially the same, with the only thing changed being the action value..
suggestion: use `action = "wagtail.reorder" if parent_before.id == target.id else "wagtail.move"` and do a single `PageLogEntry.objects.log_action()` call

Stormheg: Makes sense, will fix!

Stormheg:
Thanks @zerolab :smile: 

As this is nearing completion I've squashed some of the commits.

This pull requests does two things now:

- It fixes #6762 by where page reordered log entries would display as performed by the system.
- It adds a new `wagtail.reorder` action to improve the message displayed in the audit log and to make it possible to filter on page reordered actions.

I don't think it is necessary to have a data migration to convert `wagtail.move` actions to `wagtail.reorder`

zerolab (approved):
LGTM!
Thank you @Stormheg

Stormheg:
Merged in 520fa23 and 1ff679f :tada:

Issue #5989, opened 2020-04-27
Feature request: Switch TableBlock’s header controls to a field that requires user input


thibaudcolas:
### Is your proposal related to a problem?

Related: #5865, #5977. Wagtail’s [TableBlock](https://docs.wagtail.io/en/v2.8.1/reference/contrib/table_block.html#tableblock) has support for defining the first row, or column, as table headers. 

While it’s not mandatory for a table to have header cells, if the table isn’t meant for layout then using header cells is a usability and accessibility benefit (or requirement) – header cells are generally displayed differently (e.g. bold), and screen readers will use the header cell’s label when announcing items is the header for. I believe failing to mark up header cells as such is a failure of [WCAG2.1’s Success Criterion 1.3.1 (Info and Relationships)](https://www.w3.org/TR/WCAG20-TECHS/F91.html).

I think the current form makes it a bit too easy to miss those checkboxes and leave the table without headers.

### Describe the solution you'd like

I would suggest changing the header cells’ two checkboxes to a single "Header cells" dropdown, with the following options:

- [Choose] (Default)
- First row
- First column
- First row and first column
- No header cells

This field would be required. The main improvement here is the fact that contrary to checkboxes, the field has to be interacted with for the user to select the "No header cells" option that we want to discourage the use of (but should still be an option).

### Describe alternatives you've considered

One alternative would be to do this as a per-project customization, but it feels like an improvement that could easily live in Wagtail directly. Other changes to the TableBlock to remove the ability to have no header cells aren’t particularly compelling, because we still want the ability to select row/column/both.

### Additional context

Additionally to the above, as part of this change we could consider:

- Improving the help text for this field, to clarify its usage and the accessibility consequences
- Adding a straightforward way to customize this to disable the "No header cells" option – in practice all data tables should have header cells, and a table that doesn’t have any would be better as a list, or heading sections

In terms of migration, this looks like it could be easy:

- First row checkbox -> First row dropdown option
- First column checkbox -> First column dropdown option
- Both checkboxes -> "Both" dropdown option
- No checkboxes -> "No header cells" option

thibaudcolas:
cc @zmustapha and @rjpruitt16 if you have any thoughts on this.

gasman:
[...]

PR #6763, opened 2021-02-04
Converted column and row headers from checkboxes to a dropdown menu

cnk:
This PR addresses issue #5989 

NOTE this commit will change the default behavior for TableBlocks to having a header row. If we want to change this, we can either add an option that is no choice OR we could set the default to 'neither' which mimics not having either checkbox checked. 

* For Python changes: Have you added tests to cover the new/fixed behaviour? I added some tests but did not specifically start testing the admin UI implementation which moved from checkboxes to a select widget
* For front-end changes: Did you test on all of Wagtail’s supported browsers? No. I tested:  Chrome 88.0.4324.96 and FIrefox 84.0.2 on MacOS 10.15.7
* For new features: Has the documentation been updated accordingly? Yes, including updating the image showing the admin UI.


thibaudcolas:
Amazing! I’ll review this as soon as I can.

For the default value – re-reading #5989, I was keen for the default to be a "no choice" option so we force users to make a choice (making sure the select is required with normal form validation). Would that seem reasonable to you?

docs/reference/contrib/table_block.rst
```diff
@@ -99,13 +100,15 @@ When defining a TableBlock, Wagtail provides the ability to pass an optional ``t
 Configuration Options
 ^^^^^^^^^^^^^^^^^^^^^
 
-Every key in the ``table_options`` dictionary maps to a `handsontable <https://handsontable.com/>`_ option. These settings can be changed to alter the behaviour of tables in Wagtail. The following options are available:
+The ``table_header_choice`` option determines if rows or columns are treated as headings. Valid values are ``row``, ``column``, ``both``, and ``neither``. The default is ``row``. This will cause the elements in the first row of your table to be rendered as ``th`` elements with ``scope = row``.
```
thibaudcolas, line 103: The distinction isn’t that big in my mind but generally I only see these referred to as "headers".

```suggestion
The ``table_header_choice`` option determines if rows or columns are treated as headers. Valid values are ``row``, ``column``, ``both``, and ``neither``. The default is ``row``. This will cause the elements in the first row of your table to be rendered as ``th`` elements with ``scope = row``.
```

docs/reference/contrib/table_block.rst
```diff
@@ -99,13 +100,15 @@ When defining a TableBlock, Wagtail provides the ability to pass an optional ``t
...
+
+The rest of the keys in the ``table_options`` dictionary each map to a `handsontable <https://handsontable.com/>`_ option. These settings can be changed to alter the behaviour of tables in Wagtail. The following options are available:
 
 * `minSpareRows <https://handsontable.com/docs/6.2.2/Options.html#minSpareRows>`_ - The number of rows to append to the end of an empty grid. The default setting is 0.
 * `startRows <https://handsontable.com/docs/6.2.2/Options.html#startRows>`_ - The default number of rows for a new table.
 * `startCols <https://handsontable.com/docs/6.2.2/Options.html#startCols>`_ - The default number of columns for new tables.
-* `colHeaders <https://handsontable.com/docs/6.2.2/Options.html#colHeaders>`_ - Can be set to ``True`` or ``False``. This setting designates if new tables should be created with column headers. **Note:** this only sets the behaviour for newly created tables. Page editors can override this by checking the the “Column header” checkbox in the table editor in the Wagtail admin.
```
thibaudcolas, line 107: Should we keep that note or is it no longer relevant? I’m not entirely sure I understand it but would be keen to hear more.

docs/reference/contrib/table_block.rst
```diff
@@ -99,13 +100,15 @@ When defining a TableBlock, Wagtail provides the ability to pass an optional ``t
...
-* `rowHeaders <https://handsontable.com/docs/6.2.2/Options.html#rowHeaders>`_ - Operates the same as ``colHeaders`` to designate if new tables should be created with the first column as a row header. Just like ``colHeaders`` this option can be overridden by the page editor in the Wagtail admin.
+* `colHeaders <https://handsontable.com/docs/6.2.2/Options.html#colHeaders>`_ - Can be set to ``True`` or ``False``. If this is set to ``True``, tables will display the default column headings ('A', 'B', 'C'...) in the admin interface. This does not add headings to the public display of your table block.
```
thibaudcolas, line 110:
```suggestion
* `colHeaders <https://handsontable.com/docs/6.2.2/Options.html#colHeaders>`_ - Can be set to ``True`` or ``False``. If this is set to ``True``, tables will display the default column headers ('A', 'B', 'C'...) in the admin interface. This does not add headers to the public display of your table block.
```

docs/reference/contrib/table_block.rst
```diff
@@ -99,13 +100,15 @@ When defining a TableBlock, Wagtail provides the ability to pass an optional ``t
...
+* `rowHeaders <https://handsontable.com/docs/6.2.2/Options.html#rowHeaders>`_ - Can be set to ``True`` or ``False``. If this is set to ``True``, tables will display the default row headings ('1', '2', '3'...) in the admin interface. This does not add row headings to the public display of your table block.
```
thibaudcolas, line 111:
```suggestion
* `rowHeaders <https://handsontable.com/docs/6.2.2/Options.html#rowHeaders>`_ - Can be set to ``True`` or ``False``. If this is set to ``True``, tables will display the default row headers ('1', '2', '3'...) in the admin interface. This does not add row headers to the public display of your table block.
```

docs/reference/contrib/table_block.rst
```diff
@@ -120,23 +123,18 @@ A `complete list of handsontable options <https://handsontable.com/docs/6.2.2/Op
 Changing the default table_options
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-To change the default table options just pass a new table_options dictionary when a new TableBlock is declared.
+To change the default table options, pass a new table_options dictionary with the values you would like to override when you define a your TableBlock.
```
thibaudcolas, line 126: Nice! Much clearer like this.

wagtail/contrib/table_block/blocks.py
```diff
@@ -10,6 +10,7 @@
 
 
 DEFAULT_TABLE_OPTIONS = {
+    'table_header_choice': 'row',
```
thibaudcolas, line 13: At first glance I think it would be better for there to be no default so editors are forced to choice.

wagtail/contrib/table_block/static/table_block/js/table.js
```diff
@@ -43,6 +45,24 @@ function initTable(id, tableOptions) {
         parentDiv.find('.field-content').width(width);
         parentDiv.find('.fieldname-table .field-content .field-content').width('80%');
     }
+    var setHiddenHeaderValues = function(headerChoice) {
+        if (headerChoice == 'row') {
```
thibaudcolas, line 49: Nitpick:

```suggestion
        if (headerChoice === 'row') {
```

Same for all below.

wagtail/contrib/table_block/templates/table_block/widgets/table.html
```diff
@@ -1,21 +1,24 @@
...
+            <option value="both">
+                {% trans 'Display the first row AND first column as headers' %}
+            </option>
+            <option value="neither">
+                {% trans 'No headers' %}
+            </option>
+        </select>
+        <p class="help">{% trans 'Which cells should be displayed as headers?' %}</p>
+        <input type="hidden" id="{{ widget.attrs.id }}-handsontable-header" name="handsontable-header"/>
+        <input type="hidden" id="{{ widget.attrs.id }}-handsontable-col-header" name="handsontable-col-header"/>
```
thibaudcolas, line 21: Do we really need these two hidden fields? Reading through the JS, it seems like we could also use the value of the `-table-header-choice` select directly inside the `persist` function?

thibaudcolas (commented):
@cnk I couldn’t resist having a quick look now… I’ve made some suggestions and also have a few questions. This is based on the code only, I’ll hopefully find some time to test this next week.

docs/reference/contrib/table_block.rst
cnk: Removing all reference to colHeaders and rowHeaders might be the best option. After all, we do not document every Handsontable configuration option; we only include the ones most relevant to our use of the library.

thibaudcolas: Indeed, that good to me.

cnk: I have removed colHeaders and rowHeaders from our docs. Mentioning them was confusing.

wagtail/contrib/table_block/blocks.py
cnk: I removed having a default for all TableBlocks but will probably add back a way for developers to set a default when defining their own table blocks

wagtail/contrib/table_block/templates/table_block/widgets/table.html
cnk: Good point. I moved the mapping from table_header_choice back to the original persisted options (first_row_is_table_header and first_col_is_header) from the JS layer to the field's clean method.

cnk:
I have a test to fix (and perhaps an additional one to write) but am ready for some help with the error handling.

Adding a blank option to our select means we need to require the user to choose one of the other options. I have put a check for this in the the block's `clean` method but I am unclear on how to get the error message to show only for the select item. The comment at wagtail/core/blocks/list_block.py line 136 leads me to believe this is a solved problem but I don't know what the solution is.

salty-ivy:
@cnk  does this PR needs anymore work or just error handling ? i am following through this PR to create a new one if i may ?

cnk:
@salty-ivy Thanks for taking this up. This PR was written against Wagtail 2.x so the first thing that needs to be done is to update it for Wagtail 4.1+. At that point, you will need to add the error handling mentioned above and then see what the test coverage looks like. 

Given the changes to Wagtail's code organization and UI, if you find it easier to start a new branch from the current main branch and then copy / adapt the changes here, I am fine with closing this PR in favor of a new one you write.

salty-ivy:
> @salty-ivy Thanks for taking this up. This PR was written against Wagtail 2.x so the first thing that needs to be done is to update it for Wagtail 4.1+. At that point, you will need to add the error handling mentioned above and then see what the test coverage looks like.
> 
> Given the changes to Wagtail's code organization and UI, if you find it easier to start a new branch from the current main branch and then copy / adapt the changes here, I am fine with closing this PR in favor of a new one you write.

Great, i would need some references from this PR let it open for the time being until i prepare a new one, thanks a lot

thibaudcolas:
Closing in favour of #9673 which I believe takes the feedback in this PR into account.

PR #6770, opened 2021-02-04
Support skipping CI

tomdyson:
Don't run CI if the commit message includes "[skip ci]" (like this one!)

There may be a cleaner way of applying this generally, rather than adding a conditional to each set of tests, but this should work for now.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,6 +29,7 @@ on:
 
 jobs:
   test-sqlite:
+    if: "!contains(github.event.head_commit.message, '[skip ci]')"
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


gasman (approved):

tomdyson:
This helps, but didn't work completely: It looks like the `pull_request` event doesn't have the `head_commit` key available, so it's only working on the `push` events (see https://github.com/wagtail/wagtail/pull/6777 for an example of a PR with partially skipped checks).

tomdyson:
This discussion suggests

`if: "!contains(toJSON(github.event.commits.*.message), '[skip-ci]')"`

rather than 

`if: "!contains(github.event.head_commit.message, '[skip ci]')"`

PR #6775, opened 2021-02-04
Changed the title for Wagtail AMP support

estaar:


.gitignore
```diff
@@ -2,6 +2,7 @@
 *.pyc
 .DS_Store
 /.coverage
+/false/
```
tomdyson, line 5: What is this for @estaar?

estaar: My development machine creates that folder everytime am running a virtual environment with pycharm. something to do with Pipenv

tomdyson: Is it possibly from npm?

https://github.com/npm/npm/issues/14514

estaar: I think so. I use Pipenv which has some dependency on node.js for running interactive maps. I will relook at this from my side and see how I can clean it. so in all my .gitignore files am adding this folder.

thibaudcolas: As a workaround you can set up a global "gitignore" for your whole computer: https://stackoverflow.com/a/7335487/1798491.

I would only recommend doing this for files that are specific to your computer and projects are unlikely to have, so looks like this might be a good case.

docs/advanced_topics/amp.rst
```diff
@@ -1,5 +1,5 @@
-Building a site with AMP support
-================================
+How to Build a site with AMP support
```
thibaudcolas, line 1:
```suggestion
How to build a site with AMP support
```

I think we should keep this all lowercase like other pages.

thibaudcolas (changes requested):
Thank you @estaar :) Great to see you opening that PR so early in the sprint.

I’ve suggested two changes, please have a look and ping me when you’ve made the changes.

docs/advanced_topics/amp.rst
estaar: Okay. I am redoing it now.

thibaudcolas (approved):
Thank you!

PR #6778, opened 2021-02-04
Telepath: ListBlock random insert

kaedroho:
This PR adds the ability to insert items into list blocks at any position. The implementation is copied from stream block:
image

### DIFF
--- a/client/src/components/StreamField/blocks/ListBlock.js
+++ b/client/src/components/StreamField/blocks/ListBlock.js
@@ -47,6 +47,14 @@ class InsertPosition {
       }
     });
   }
+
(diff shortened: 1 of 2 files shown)


gasman:
Thanks @kaedroho! This overlaps a fair bit with the ListBlock / StreamBlock refactoring I was in the middle of, but that's probably not a bad thing - it was a bit of a toss-up whether it made more sense to make the two classes match and then extract the common bits into a base class, or to extract the shareable functionality from StreamBlock into a base class first and then apply it to ListBlock.

gasman (approved):

PR #6774, opened 2021-02-04
Update promote tab field information

phildexter:
Includes best practice changes, as well as expected updates to field titles (meta description and title tag), expected to be done by Kalob during this sprint.

PR #6780, opened 2021-02-04
Promote panel changes

KalobTaulien:
This is in conjunction with PR #6774 

This new documentation references new changes that are reflected in this PR.

image


tomdyson (commented):
Thanks @KalobTaulien, this is clearer and the changes LGTM.

My concern is that these changes might give the impression that just by setting these values, 'The descriptive text [will be] displayed underneath a headline in search engine results' etc., which isn't really the case; they need to set these values and then whoever is responsible for the front-end has to make sure that they're exposed in the correct way.

I appreciate that this is too subtle a point to make in the UI, but we should be clear about this in the docs.

I'm also in favour of moving the associated docs change into this PR.

KalobTaulien:
Awesome, thanks for checking this out @tomdyson 

I'll merge in that docs PR into this one and reconnect with that team around wording and clarity.

KalobTaulien:
Notes for @phildexter and @LisaBallam — we should make it extra clear in the docs that the Title tag and Meta description are the responsibility of the developer. The rationale is: just because it's what wagtail suggests, doesn't mean the developers will necessarily follow that guideline, or that it will "just magically work".

phildexter:
@tomdyson I guess you're talking here about search on the frontend of the website?

Here I think I (we) meant, search engines such as google or bing. These things should get picked up automatically I think?

tomdyson:
> Here I think I (we) meant, search engines such as google or bing. These things should get picked up automatically I think?

No, they won't get picked up automatically - that's the point I want to make. Setting them in the back end has no effect unless a front-end developer (either using Wagtail templates, or possibly something completely separate in the case of a headless site) chooses to output them with the appropriate `<meta name="description" ...>` etc. tags.

wagtail/core/migrations/0061_auto_20210204_1531.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 3.0.12 on 2021-02-04 15:31
```
gasman, line 1: Please give this migration file a meaningful filename - more than anything, it helps us to troubleshoot when people accidentally introduce stray migrations as a result of monkeypatching the page model.

phildexter:
I see, thanks Tom. I'll make some amends to the docs!

wagtail/core/migrations/0061_auto_20210204_1531.py
KalobTaulien: Done! Thank you for catching this

phildexter:
This has been updated now @tomdyson and ready for your re-review.

PR #6782, opened 2021-02-04
Skip PRs as well as pushes with [skip ci]

tomdyson:
An attempt to handle the partial skipping behaviour detailed in https://github.com/wagtail/wagtail/pull/6770#issuecomment-773385420

This approach should work if `[skip ci]` is in the last commit of any push or PR.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,7 +29,7 @@ on:
 
 jobs:
   test-sqlite:
-    if: "!contains(github.event.commits[0].message, '[skip ci]')"
+    if: "!contains(toJSON(github.event.commits[0].message), '[skip ci]')"
     runs-on: ubuntu-latest
     strategy:
       matrix:
@@ -57,7 +57,7 @@ jobs:
           DATABASE_ENGINE: django.db.backends.sqlite3
 
   test-postgres:
-    if: "!contains(github.event.commits[0].message, '[skip ci]')"
+    if: "!contains(toJSON(github.event.commits[0].message), '[skip ci]')"
     runs-on: ubuntu-latest
     continue-on-error: ${{ matrix.experimental }}
     strategy:
@@ -112,7 +112,7 @@ jobs:
           DISABLE_TIMEZONE: ${{ matrix.notz }}
 
   test-mysql:
-    if: "!contains(github.event.commits[0].message, '[skip ci]')"
+    if: "!contains(toJSON(github.event.commits[0].message), '[skip ci]')"
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


tomdyson:
https://github.com/wagtail/wagtail/pull/6786 might mean this is redundant, although a manual [skip ci] will still help reduce cycles in some cases, e.g. in the case of an update to README.md.

Scotchester:
Hmm… I think this should have worked to skip the PR checks on this very PR. I wonder why it didn't. I wonder if this approach needs the `toJSON` suggested in [this comment](https://stackoverflow.com/questions/59759921/how-to-skip-github-actions-job-on-push-event/59775665#comment110153861_59775665) to the answer you were basing this change off of.

gasman (approved):
LGTM... I'll leave it to @Scotchester to merge this or not, depending on how it fits in with the rest of the changes.

Scotchester:
Well, it appears that `github.event.commits` simply does not exist for workflows triggered by `pull_request` events, so all permutations of the suggestions listed in [yesterday's comment](https://github.com/wagtail/wagtail/pull/6770#issuecomment-773392214) are invalid.

GHA output showing github.event.commits is null on a pull request event

See: https://github.com/wagtail/wagtail/pull/6782/checks?check_run_id=1840232907#step:2:1

I'll keep digging. There are a couple prebuilt Actions that might help: https://github.com/atlwendy/retrieve-commit-messages-from-pr and https://github.com/mstachniuk/ci-skip

Scotchester:
Another possible solution not involving loading a third-party action: https://github.community/t/accessing-commit-message-in-pull-request-event/17158/9

Scotchester:
That works!! It adds the minor noise of the preparation "jobs", but it appears to reliably ascertain whether `[skip ci]` is in the commit message for both push and PR events.

Cleaning up and applying to all jobs now.

Scotchester:
🎉 

@tomdyson and @gasman, are you all cool with this approach?

gasman:
@Scotchester Sure, let's go for it!

tomdyson:
I'm not sure about this. It adds a couple of jobs - admittedly small, at ~5 seconds each - to each CI run, to check for the possibility of a [skip ci] which may be used very rarely. I think your excellent auto-skipping approach will solve the problem of unnecessary builds 95% of the time, so perhaps we should leave it at that.

Scotchester:
One bit of food for thought before we ditch the idea, Tom:

Regarding the auto-skipping based on changed paths I implemented in #6786 – It appears that the `pull_request` event always looks at the total change set in the PR, not just the most recent commit. As a result, it will still run on later commits to a PR that didn't change relevant files, even if the check had passed previously. See, for example, my [final commit](https://github.com/wagtail/wagtail/pull/6790/commits/08de57661e6a5991c880c2ba047752ac20b174fd) on #6790, which despite only changing an ignored path, still triggered the test workflow for the `pull_request` event (but not the `push` event, which just looks at the one commit). In cases like that, or things like cleaning up typos in a comment or something, having the ability to throw `[skip ci]` on it would still be useful.

Does that sway you any?

Stormheg:
Is this PR still valid now that Github Actions [supports skip-ci out-of-the-box](https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/)?

Scotchester:
> Is this PR still valid now that [Github Actions supports skip-ci out-of-the-box](https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/)?

No! Sorry I kept forgetting to close it 😄

PR #6786, opened 2021-02-04
Limit GitHub Actions docs and test workflows depending on changed paths

Scotchester:
This should result in the docs workflow only running if files in the `docs/` folder have changed, and only run the test workflow if files NOT in the `docs/` folder have changed.

This first commit itself should therefore, I think, only trigger the test workflow. I will follow up with test commits for the other scenarios (docs-only and both).

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,7 +2,12 @@ name: Wagtail Docs
 
 on:
   push:
+    paths:
+    - 'docs/**'
   pull_request:
+    paths:
+    - 'docs/**'
+
 
 jobs:
   build-docs:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,7 +2,11 @@ name: Wagtail CI
 
 on:
   push:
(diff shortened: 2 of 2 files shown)


Scotchester:
It appears that the first commit has not queued the docs workflow, as expected.

Now realizing that we'll also want to see if we can apply similar filters to our third party integrations that run checks.

Scotchester:
Without breaking down the code changes into frontend/backend, I think this is the basic breakdown of when we want these builds/checks to run:

Service/Workflow | Docs changes only | Code changes only | Docs and code changes
-- | -- | -- | --
GitHub Actions Docs | ✅ | ⛔️ | ✅
GitHub Actions CI | ⛔️ | ✅ | ✅
Squash | ⛔️ | ✅ | ✅
LGTM | ⛔️ | ✅ | ✅
Circle CI | ⛔️ | ✅ | ✅
Read the Docs | ✅ | ⛔️ | ✅

tomdyson (approved):
LGTM but we won't know until merging!

PR #6789, opened 2021-02-04
Move "writing documentation" page into Contributing

gasman:
While we very much want to introduce an "explanation" section in some form, I don't think this should be the inaugural item in that section, as it's 'meta' documentation about the Wagtail documentation process, rather than information directed at Wagtail site creators.


tomdyson (approved):
I agree.

PR #6790, opened 2021-02-04
Only run Circle CI jobs if appropriate files have been changed

Scotchester:
Testing out the technique described https://github.com/labs42io/circleci-monorepo/tree/v2

🤞


Scotchester:
Looks like that worked! No reports from Circle CI on a98d2c3.

Going to try adding some temporary commits with front-end and back-end changes and see if the Circle CI builds run correctly.

Scotchester:
Mmmnope. Failed with "ENV variable `CIRCLE_USER_TOKEN` is empty. Please provide a user token."

I'll come back to this tomorrow.

Scotchester:
Ready for review!

zerolab:
I missed this during the documentation sprint, but I do wonder if it is worth converting the CircleCI config to a GitHub Action workflow?

Scotchester:
@zerolab I feel like I heard some reason for why we want to continue to do some stuff on CircleCI, but I can't pinpoint what that was.

thibaudcolas:
I believe the reason is that it gives us much faster feedback, since CircleCI builds are only competing with one-another.

If we moved everything to GitHub Actions, we would have those basic "linting + tests" checks competing for execution time with the Python/Django/DB compatibility check jobs of other PRs.

.circleci/config.yml
```diff
@@ -1,6 +1,29 @@
...
 
 jobs:
+  trigger-workflows:
+    docker:
+      - image: cimg/base:stable
+    steps:
+      - checkout
+      - run:
+          name: Trigger workflows
+          command: chmod +x .circleci/conditional-builds.sh && .circleci/conditional-builds.sh
```
thibaudcolas, line 25: Could commit the file to git with the +x permissions added already?

.circleci/conditional-builds.sh
```diff
@@ -0,0 +1,255 @@
+#! /usr/bin/env bash
+
+set -e
```
thibaudcolas, line 3: I think it would be nice to add a comment here explaining where this code is from

Scotchester: Yes! I'll do that.

.circleci/config.yml
Scotchester: Good call! I'll give it a try.

.github/workflows/test.yml
```diff
@@ -3,10 +3,30 @@ name: Wagtail CI
...
+      - 'docs/**'
+      - '.readthedocs.yml'
+      - '.squash.yml'
+      - 'CHANGELOG.txt'
+      - 'CODE_OF_CONDUCT.md'
+      - 'CONTRIBUTORS.rst'
+      - 'LICENSE'
+      - 'README.md'
+      - 'SECURITY.md'
+      - 'SPONSORS.md'
```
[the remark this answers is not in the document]
Scotchester: It might be… but one could argue that it's more conservative to use a blocklist, so that if new paths that should be checked are added, someone doesn't have to know to update the allowlist.

Scotchester:
All comments addressed except for the allowlist vs. blocklist and rebased on main. Looks like the CircleCI check is working as expected. Let me know what you think about that remaining issue, @thibaudcolas.

thibaudcolas:
@Scotchester sorry to make this drag on… re-reviewing this, I have concerns about the `conditional-builds` script 😐 . I tried to run it on a disconnected fork (https://github.com/thibaudcolas/wagtail-fork) I use to test CI setups, and realised this is relying on the CircleCI API and an API token with write permissions. Looking at our CircleCI project settings, we are _not_ allowing "Pass secrets to builds from forked pull requests", so we’d have to turn this on for this implementation to work with the majority of pull requests. This would be incompatible with our nightly builds storing their output in S3 via credentials also set up as environment variables.

Trying to look into this further, it looks like CircleCI now has an official [`path-filtering`](https://circleci.com/developer/orbs/orb/circleci/path-filtering) orb which we could alternatively use for this. Here’s an example from Electron: https://github.com/electron/electron/blob/d657cd8ed66aebe303caa26d620081249300556c/.circleci/config.yml#L132-L144.

This seems worth a shot to me, so we bypass the need for an API token. What do you think?

thibaudcolas:
I think I got this working at https://app.circleci.com/pipelines/github/thibaudcolas/wagtail-fork?filter=all / https://github.com/thibaudcolas/wagtail-fork/commit/fb2cb2d3fc93bce24a99698752d2d39950e0195a. Keen to hear your thoughts on this.

From what I understand this "dynamic configuration" setup _doesn’t_ support scheduled workflows, so we’d need to migrate to [scheduled pipelines](https://circleci.com/docs/2.0/scheduled-pipelines/). It looks like we’d have to do this anyway though as scheduled workflows are going to be [removed in June](https://circleci.com/docs/2.0/workflows/#scheduling-a-workflow).

thibaudcolas:
FYI, I can confirm the scheduled pipelines work with minimal changes. See the recent runs in https://app.circleci.com/pipelines/github/thibaudcolas/wagtail-fork?filter=all (only failing because of the missing environment variables). Here is the configuration I used:

scheduled-pipelines-wagtail-6790

Scotchester:
@thibaudcolas Fantastic! I don't know if that orb was publicly available when I initially opened that PR, but it certainly looks like a much cleaner solution.

My only question about your POC is whether or not we need the additional `build_config.yml` file like Electron uses or if we can still do it all in one. The info on the `config-path` option in the orb docs says that it's not required, and implies you might want to use it for organizational purposes, but I don't know if our setup is complex enough to warrant that?

thibaudcolas:
@Scotchester my impression from reading the source of the orb is that it’s required for it to be a separate file – the `config-path` is just optional because they default to `.circleci/continue_config.yml` if unset. I haven’t tried to have it all in one file and see if that works, might be worth a shot 😄

lb-:
What should we do with this PR? Is it something we should raise an issue for instead? It's been a few years with no comments, plus it's connected to a branch on the Wagtail repo which would be good to remove for spring cleaning.

Scotchester:
@thibaudcolas Think the solution you came up with 3 years ago is still valid? Perhaps we should try transferring it from your fork into this branch.

thibaudcolas:
I’ll close this and have opened #13706 as a replacement – definitely something we still want, but it likely needs a new approach (what I found, or something else altogether).

Issue #6788, opened 2021-02-04
oEmbed failures related to thumbnails


pySilver:
### Issue Summary

Trying to embed 2 items to test fixes recently introduced to Wagtail 2.12 (thumbnail hash):

https://twitter.com/CyberpunkGame/status/1354769450726285312
https://www.facebook.com/martyna.world/posts/256722362490410

Both fails with the following error:

```
Traceback (most recent call last):
django_1          |   File "/opt/venv/lib/python3.8/site-packages/django/db/models/query.py", line 573, in get_or_create
django_1          |     return self.get(**kwargs), False
django_1          |   File "/opt/venv/lib/python3.8/site-packages/cacheops/query.py", line 353, in get
django_1          |     return qs._no_monkey.get(qs, *args, **kwargs)
django_1          |   File "/opt/venv/lib/python3.8/site-packages/django/db/models/query.py", line 429, in get
django_1          |     raise self.model.DoesNotExist(
django_1          | wagtail.embeds.models.Embed.DoesNotExist: Embed matching query does not exist.
django_1          | 
django_1          | During handling of the above exception, another exception occurred:
django_1          | 
django_1          | Traceback (most recent call last):
django_1          |   File "/opt/venv/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
django_1          |     return self.cursor.execute(sql, params)
django_1          | psycopg2.errors.NotNullViolation: null value in column "thumbnail_url" violates not-null constraint
django_1          | DETAIL:  Failing row contains (23, https://twitter.com/CyberpunkGame/status/1354769450726285312, null, rich, <blockquote class="twitter-tweet"><p lang="en" dir="ltr">Hotfix ..., , Cyberpunk 2077, Twitter, null, 550, null, 2021-02-04 20:10:53.143013+00, bbcc5b725a47235ef04e65ddf094e6fb).
django_1          | 
django_1          | 
django_1          | The above exception was the direct cause of the following exception:
django_1          | 
django_1          | Traceback (most recent call last):
django_1          |   File "/opt/venv/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
django_1          |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
django_1          |   File "/usr/local/lib/python3.8/contextlib.py", line 75, in inner
django_1          |     return func(*args, **kwds)
[...]

PR #6791, opened 2021-02-04
Store empty string if embed does not have thumbnail_url

cnk:
Fixes issue #6788 

* For Python changes: Have you added tests to cover the new/fixed behaviour? Modified one existing test to cover this
* For new features: Has the documentation been updated accordingly?  N/A

### DIFF
--- a/wagtail/embeds/embeds.py
+++ b/wagtail/embeds/embeds.py
@@ -42,6 +42,10 @@ def get_embed(url, max_width=None, finder=None):
     if 'html' not in embed_dict or not embed_dict['html']:
         embed_dict['html'] = ''
 
+    # If the finder does not return an thumbnail_url, convert null to '' before inserting into the db
+    if 'thumbnail_url' not in embed_dict or not embed_dict['thumbnail_url']:
+        embed_dict['thumbnail_url'] = ''
+
     # Create database record
     embed, created = Embed.objects.get_or_create(
         hash=embed_hash,
--- a/wagtail/embeds/migrations/0009_convert_null_to_empty_string.py
+++ b/wagtail/embeds/migrations/0009_convert_null_to_empty_string.py
@@ -1,0 +1,18 @@
+from django.db import migrations
+
+
+def migrate_forwards(apps, schema_editor):
+    Embed = apps.get_model("wagtailembeds.Embed")
+
+    Embed.objects.filter(thumbnail_url__isnull=True).all().update(thumbnail_url='')
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("wagtailembeds", "0008_allow_long_urls"),
+    ]
+
+    operations = [
+        migrations.RunPython(migrate_forwards, migrations.RunPython.noop),
+    ]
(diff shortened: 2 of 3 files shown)


ababic:
@cnk how do you feel about also adding a data migration to turn blank string values back to `None`? Not strictly necessary, but I think it would be have consistent falsey values at the DB level.

cnk:
@ababic I like the idea of consistent values but looking at my existing data, it looks like the consistent way to do this would be to keep the not null but default the value to ''. I'll looks some more and amend this PR.

allcaps:
If we want consistency for empty values, I agree with @cnk and suggest to migrate to empty string. The oEmbed providers who provide Null values should be cast to empty string on save too.

In this case, from the Django field perspective, there isn't any good reason to store Null.

pySilver:
+1 for @allcaps - most of the time nullable char fields brings nothing but the pain.

wagtail/embeds/embeds.py
```diff
@@ -42,6 +42,10 @@ def finder(url, max_width=None):
     if 'html' not in embed_dict or not embed_dict['html']:
```
pySilver, line 42: I guess its matter of taste but you can shortcut this (as well as `thumbnail_url` check/conversion) to smth like

```python
embed_dict['html'] = embed_dict.get('html') if isinstance(embed_dict.get('html'), str) else ''
embed_dict['thumbnail_url'] = embed_dict.get('thumbnail_url') if isinstance(embed_dict.get('thumbnail_url'), str) else ''
```

but again, thats just matter of taste. maybe your variant is more readable. 

you can always simply check if value is a string, because I guess we don't care if its _falsey_  (empty string, False, None, 0 etc) but rather if its a string type.

cnk: Sorry I didn't see this before I pushed my (hopefully) last commit on this.

cnk:
@allcaps @ababic OK I have added a data migration to convert null thumbnail_urls to '' and updated a couple of tests to demonstrate the behavior when thumbnail_url is missing in the data from the embed provider

wagtail/embeds/migrations/0009_convert_null_to_empty_string.py
```diff
@@ -0,0 +1,18 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("wagtailembeds", "0008_allow_long_urls"),
+    ]
+
+    operations = [
+        migrations.RunPython(migrate_forwards, migrations.RunPython.noop),
+    ]
```
ababic, line 18: Hi @cnk. Thank you very much for this, but I think this migration might be redundant. I could have probably done a better job of outlining that we would only need this if adding `null=True` back to the field... The AlterField operation in the previous migration should have already taken care of converting NULL values to blank strings.

gasman (approved):
Thanks @cnk - code looks good to me. Have just checked that on both postgresql and sqlite, the existing migration for adding the NOT NULL constraint does indeed change nulls to empty strings as per @ababic's comment, so the new migration isn't necessary - will go ahead and merge without the migration.

gasman:
Merged in afc3349c94ea3756f25b6444eea585ebd0c02fdf (master) and d0e6d230b21ce9ca65dd28d93022b01b90cffc6a (stable/2.12.x).

PR #6794, opened 2021-02-05
Include privacy within the settings tab section

phildexter:
Done by Phil and Abigail! This should have been updated with 2.10 when we brought privacy into this section of the docs.


tomdyson (approved):
Thanks @phildexter and Abigail! LGTM.

PR #6796, opened 2021-02-05
Create DOCUMENTATION.md

JoanEliot:


.github/ISSUE_TEMPLATE/DOCUMENTATION.md
```diff
@@ -0,0 +1,53 @@
...
+  Select one of the following options by placing an 'x' inside the corresponding brackets, or just delete the other options. 
+-->
+
+|  x  | Type of change  | Fix/enhancement |
+| --- | ------------- | ------------- |
+| [ ] | Copy edit | Fix spelling or grammar errors, improve clarity
+| [ ] | Content correction | Correct inaccurate explanations, instructions, or code 
+| [ ] | New content | Add a new topic or provide more detailed instructions or examples 
+| [ ] | Reorganization | Move topic(s) 
+| [ ] | Other suggestion | Anything else
```
thibaudcolas, line 34: This feels a bit overkill / hard to edit / hard to read as well.

thibaudcolas (changes requested):
Thank you for starting this @JoanEliot! I’ve reviewed and requested a few changes, most of which are about keeping issues easier to fill. We’ve generally had good documentation issues in the past, so I think it’s good to keep it simple to not put off potential contributors.

.github/ISSUE_TEMPLATE/DOCUMENTATION.md
thibaudcolas, line 34:
```suggestion
```

I don’t think we need contributions to be precisely categorised like this, this should hopefully be clear enough from the title and the details.

.github/ISSUE_TEMPLATE/DOCUMENTATION.md
```diff
@@ -0,0 +1,51 @@
...
+(Write your answer here.)
+
+### Proposed change (for brief corrections/revisions)
+
+<!--
+  - If you're suggesting a correction or revision of existing passage, write the corrected/revised passage here.
+?  -  If you want to contribute new or reorganized content please create a pull request...
+-->
+
+(Write your answer here.)
```
thibaudcolas, line 51: Even for documentation changes, I think we should encourage people to use pull requests only. They are much easier for us to manage than code snippets in issues. I would suggest removing this section.

```suggestion

<!--
  If you're suggesting a very specific change to the documentation, feel free to directly submit a pull request.
-->
```

.github/ISSUE_TEMPLATE/DOCUMENTATION.md
```diff
@@ -0,0 +1,51 @@
+---
+name: "📜📖📄📘 📚  Documentation enhancement"
```
thibaudcolas, line 2:
```suggestion
name: "📚 Documentation"
```

Changing this to match the format of other issue templates, with a single emoji, and the title matches the file name.

thibaudcolas (approved):
I think this is very helpful already so I’ll apply my suggestions that generally make the template smaller, and merge this.

PR #6799, opened 2021-02-08
Telepath: Validation of block counts

kaedroho:
This adds the logic for disabling/enabling creation and duplication of blocks based on the values of `maxNum` and `blockCounts[blockType].max_num`.

### DIFF
--- a/client/src/components/StreamField/blocks/StreamBlock.js
+++ b/client/src/components/StreamField/blocks/StreamBlock.js
@@ -98,10 +98,38 @@ class StreamBlockMenu {
         });
       });
     });
+
+    // Disable buttons for any disabled block types
+    this.disabledBlockTypes.forEach(blockType => {
+      $(`button.action-add-block-${h(blockType)}`, this.innerContainer).attr('disabled', 'true');
+    });
   }
 
   setIndex(newIndex) {
     this.index = newIndex;
+  }
+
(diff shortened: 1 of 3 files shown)


client/src/components/StreamField/blocks/StreamBlock.js
```diff
@@ -98,12 +98,40 @@ class StreamBlockMenu {
...
+    });
   }
 
   setIndex(newIndex) {
     this.index = newIndex;
   }
 
+  setNewBlockRestrictions(canAddBlock, disabledBlockTypes) {
+    this.canAddBlock = canAddBlock;
+    this.disabledBlockTypes = disabledBlockTypes;
```
gasman, line 114: We should set initial values for these in the constructor so that we're not implicitly relying on the parent block to call this before renderMenu or open. ~~(Additionally we might want to split this into separate setter functions for canAddBlock and disabledBlockTypes, since canAddBlock belongs on the base InsertionControl class whereas disabledBlockTypes is specific to the stream menu)~~ that can come later - for now we can keep this StreamBlock-specific

gasman (approved):
LGTM! Just one quibble about setting initial states for canAddBlock / disabledBlockTypes, but I've got a merge conflict to sort anyhow, so I'll fix that when I merge.

gasman:
Merged in 5b865d0540cb110e31854c20a8b9b3f0dbb327c6

PR #6805, opened 2021-02-10
Telepath: move telepath into an external library

gasman:


wagtail/contrib/table_block/blocks.py
```diff
@@ -44,14 +44,17 @@ def __init__(self, table_options=None, attrs=None):
...
-    class Media:
-        css = {'all': [
-            versioned_static('table_block/css/vendor/handsontable-6.2.2.full.min.css')
-        ]}
-        js = [
-            versioned_static('table_block/js/vendor/handsontable-6.2.2.full.min.js'),
-            versioned_static('table_block/js/table.js')
-        ]
+    @cached_property
+    def media(self):
```
kaedroho, line 48: is there a way we can detect these and raise an error/warning?

kaedroho (approved):
Looks good to me! Just one minor note

wagtail/contrib/table_block/blocks.py
nimasmi: I second this question. I was caught out in CI recently for having used `static` instead of `versioned_static` on a project. I have now switched away from using `class Media` and instead using the property method approach as here. However, the Wagtail docs for Telepath currently recommend the Media inner class approach, with no mention of versioned static file paths: https://docs.wagtail.io/en/stable/reference/streamfield/widget_api.html?highlight=adapter#form-widget-client-side-api.

gasman: @nimasmi Can you elaborate on 'caught out' please? `versioned_static` is an internal Wagtail implementation detail that external code should ideally not have to care about - it appends a querystring derived from the Wagtail version number, as a guard against people who don't have cache-busting set up properly so that they aren't left with stale JS on Wagtail upgrades. As such, it's not meaningful to use it on files that aren't shipped with Wagtail itself.

I'm aware that there are cases where you might end up referencing a Wagtail-supplied file in a media definition, so that things get output in the right order - I think (and hope) that these cases are avoidable, and if they aren't, I'd like to know about them.

(I'm also open to the possibility that `versioned_static` is more trouble than it's worth... it seems to have curbed the false bug reports a fair bit, but it's still possible for sufficiently ingenious fools to defeat it, e.g. by forgetting to run collectstatic in production, and I've noticed a bit of a resurgence of those reports in 2.13.)

nimasmi: (Discussed outside of GitHub). My mistake, thanks Matthew.

PR #6811, opened 2021-02-15
Don't focus date pickers when adding new blocks

gasman:
While focusing on the first child field of a new block is generally a good thing, focusing a date field opens the date picker, which looks clunky when you've not explicitly requested it. Introduce a 'soft' focus flag to indicate that we only want to focus the field if it can be done unobtrusively.


kaedroho (approved):
Looks good to me, thanks!

PR #6815, opened 2021-02-16
Remove duplicates before applying unique constraint on embed hash

### LINKED ISSUE #6764: Embed Uniqueness Issues on 2.12
### Issue Summary

Embed uniqueness was not enforced at a db level pre-2.12, so it is possible for https://github.com/wagtail/wagtail/blob/master/wagtail/embeds/migrations/0008_allow_long_urls.py to fail during upgrade. (Additionally, because the uniqueness is enforced on the hash, rather than the url, there seems to be an additional issue with certain url strings that are apparently unique in postgres (a `url_count` annotation is 1 for both) but equal in Python, but I haven't worked out exactly what the problem is here)

### Steps to Reproduce

1. Start a new project with `wagtail start myproject` on Wagtail 2.11, making sure the embeds app is added
2. Enter the django shell
3. Create two Embeds objects with the same url using `Embed.objects.create`
4. Attempt the 2.12 upgrade

### Technical details

* Python version: 3.6.4
* Django version: 2.2.13
* Wagtail version: 2.11.3 -> 2.12
* Browser version: N/A

gasman:
Fixes #6764

### DIFF
--- a/wagtail/embeds/migrations/0007_populate_hash.py
+++ b/wagtail/embeds/migrations/0007_populate_hash.py
@@ -1,4 +1,5 @@
 from django.db import migrations
+from django.db.models import Count, Min
 
 from wagtail.embeds.embeds import get_embed_hash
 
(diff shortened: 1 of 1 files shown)


wagtail/embeds/migrations/0007_populate_hash.py
```diff
@@ -28,6 +29,14 @@ def migrate_forwards(apps, schema_editor):
     if batch:
         Embed.objects.bulk_update(batch, ["hash"])
 
+    # delete duplicates
+    duplicates = Embed.objects.values('hash').annotate(
+        hash_count=Count('id'), min_id=Min('id')
```
jacobtoppm, line 34: Suggestion: Should this be `Count('hash')`?

jacobtoppm (approved):
This looks good to me!

gasman:
Merged in bc1458f9e35c72070787380964eaf9b0806dca4a (master) / 9ecfbfa (stable/2.12.x)

PR #6818, opened 2021-02-17
Docs: Include custom permissions hook documentation

danthedeckie:
This includes a demonstration of using the register_permissions admin hook to display permissions for custom objects that don't show up automatically.


lb- (approved):
@danthedeckie thanks for this, I know it has been a while since you put this PR up but I took a look and it looks great.

I have moved the example to the hooks page and instead just put a reference in the permissions page & pushed into your branch.

Will merge in shortly.

PR #6826, opened 2021-02-19
Allow html formatted error messages in javascript validation from the multi image uploader

### LINKED ISSUE #6824: Allow html formatted error messages
It would be nice if error messages in the `fileupload()` javascript could include html.

[Code block of interest](https://github.com/wagtail/wagtail/blob/3a057f2fd805b5040a9b88e770c7c15bca7d6e93/wagtail/images/static_src/wagtailimages/js/add-multiple.js#L62)

You can use `mark_safe()` on the error message but the use of [text](https://api.jquery.com/text/) includes the html literally.

This would allow linking the user to an alternate action - for example, pointing the user to the single image upload form that can handle large file uploads.

Example of current behavior:
wagtail-text-error

thenewguy:
Fixes #6824


lb-:
@thenewguy thanks for this, I have validated locally and it works as expected.
I have rebased, squashed and added release notes to your branch and will get this merged in.

If you get the chance, it would be good to do a similar change for the multiple document uploads and also look at how the single upload pages for documents/images + upload modals handle html. It would be good to align all of them if possible, however for now this is a backwards compatible change and the potential inconsistency would only impact those that customise these error messages.

PR #6827, opened 2021-02-20
Telepath - collapsible blocks

gasman:


client/src/components/StreamField/blocks/StructBlock.js
```diff
@@ -113,6 +113,19 @@ export class StructBlock {
     return value;
   }
 
+  getTextLabel(opts) {
+    /* Use the text label of the first child block to return one */
```
kaedroho, line 117: Idea: Might be nice to allow a developer to specify a field name to use?

client/src/entrypoints/admin/page-chooser.js
```diff
@@ -42,6 +42,14 @@ function createPageChooser(id, openAtParentId, options) {
 
       state = newState;
     },
+    getTextLabel: (opts) => {
+      if (!state) return null;
+      const result = state.adminTitle;
+      if (opts && opts.maxLength && result.length > opts.maxLength) {
```
kaedroho, line 48: Would it be possible to move this truncation logic into the caller so it doesn't need to be repeated?

kaedroho (approved):
Just a couple of little suggestions, but looks good to me as it is!

client/src/components/StreamField/blocks/StructBlock.js
gasman: Sure - there's a lot of scope for making this more customisable, perhaps with some limited templating like `"${first_name} ${last_name}"`. This will do as a first pass though!

client/src/entrypoints/admin/page-chooser.js
gasman: I think there are a couple of advantages in doing the truncation at the child level:

- it means we can have block-specific behaviour for the truncation - e.g. ListBlock will stop adding items at the point where it can't fit them in full, rather than cutting them off mid-label
- it means that we can return as soon as we reach maxLength, and save some redundant work calling getTextLabel on later child blocks that are just going to get cut off

gasman:
Merged in eb18d37986970c0b1fafb0f943593482d879c3b2 + parents

PR #6828, opened 2021-02-21
Add How-To instructions for landing-page redirects

nimasmi:
### Pertinent section of the Wagtail docs 

https://docs.wagtail.io/en/stable/advanced_topics/index.html  
https://docs.wagtail.io/en/stable/reference/contrib/forms/customisation.html

### Suggestion type

|  x  | Type of change  | Fix/enhancement |
| --- | ------------- | ------------- |
| [ ] | Copy edit | Fix spelling or grammar errors, improve clarity
| [ ] | Content correction | Correct inaccurate explanations, instructions, or code 
| [x] | New content | Add a new topic or provide more detailed instructions or examples 
| [ ] | Reorganization | Move topic(s) 
| [ ] | Other suggestion | Anything else

### Details

This uses both the forms and routable_page contrib modules to redirect
without using a separate landing page.

There is a similar guide in the Form builder documentation at https://github.com/wagtail/wagtail/blame/master/docs/reference/contrib/forms/customisation.rst#L552, but it requires adding a separate landing page. I have long thought that the non-redirecting behaviour of the standard Form builder pages was counter to the fairly standard practice of redirecting after a POST request, to prevent issues with refreshing the pages. This was raised in #2834, but closed because the use of the GET parameter to provide the form submission was implementation dependent. It looks unlikely to become part of Wagtail core this way, but the recipe ought to be documented.

Because this example spans two modules, I have not added it to the form builder documentation directly. I have also used the new 'How To' label from the January 2021 documentation sprint.

### DIFF
--- a/docs/advanced_topics/formbuilder_routablepage_redirect.md
+++ b/docs/advanced_topics/formbuilder_routablepage_redirect.md
@@ -1,0 +1,55 @@
+# How to use a redirect with Form builder to prevent double submission
+
+It is common for form submission HTTP responses to be a `302 Found` temporary redirection to a new page.
+By default `wagtail.contrib.forms.models.FormPage` success responses don't do this, meaning there is a risk that users will refresh the success page and re-submit their information.
+
+Instead of rendering the `render_landing_page` content in the POST response, we will redirect to a `route` of the `FormPage` instance at a child URL path.
+The content will still be managed within the same form page's admin.
+This approach uses the additonal contrib module `wagtail.contrib.routable_page`.
+
+An alternative approach is to redirect to an entirely different page, which does not require the `routable_page` module.
+See [Custom landing page redirect](../reference/contrib/forms/customisation.html#custom-landing-page-redirect).
+
+Make sure `"wagtail.contrib.routable_page"` is added to `INSTALLED_APPS`, see [Installation](../reference/contrib/routablepage.html#installation) documentation.
+
+```python
+from django.shortcuts import redirect
+
+from wagtail.contrib.forms.models import AbstractEmailForm
+from wagtail.contrib.routable_page.models import RoutablePageMixin, route
+
+
+class FormPage(RoutablePageMixin, AbstractEmailForm):
+
+    # fields, content_panels, …
+
+    @route(r"^$")
+    def index_route(self, request, *args, **kwargs):
+        """Serve the form, and validate it on POST"""
+        return super(AbstractEmailForm, self).serve(request, *args, **kwargs)
+
+    def render_landing_page(self, request, form_submission, *args, **kwargs):
+        """Redirect instead to self.thank_you route"""
+        url = self.reverse_subpage("thank_you")
+        # If a form_submission instance is available, append the ID to URL.
+        if form_submission:
+            url += "?id=%s" % form_submission.id
+        return redirect(self.url + url, permanent=False)
+
+    @route(r"^thank-you/$")
+    def thank_you(self, request):
+        """Return the superclass's landing page, after redirect."""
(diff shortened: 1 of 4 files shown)


docs/advanced_topics/formbuilder_routablepage_redirect.md
```diff
@@ -0,0 +1,50 @@
+# How to… use a redirect with Form builder to prevent double submission
```
lb-, line 1: Not sure we need the `...` here.
```suggestion
 # How to use a redirect with Form builder to prevent double submission
```

docs/advanced_topics/formbuilder_routablepage_redirect.md
```diff
@@ -0,0 +1,50 @@
...
+
+It is common for form submission HTTP responses to be a `302 Found` temporary redirection to a new page, so that a user who hits refresh doesn't submit their information again, because the URL has changed.
```
lb-, line 3: Maybe reverse the wording
```suggestion
When a user submits a form, if they refresh the page it will re-submit their information again, to work around this it is common for form submission HTTP responses to be a `302 Found` temporary redirection to a new page.
```

docs/advanced_topics/formbuilder_routablepage_redirect.md
```diff
@@ -0,0 +1,50 @@
...
+
+This approach uses two contrib modules `wagtail.contrib.routable_page` and `wagtail.contrib.forms`.
+For a solution which redirects to an entirely different page, and doesn't require use of routablepages, see [Custom landing page redirect](../reference/contrib/forms/customisation.html#custom-landing-page-redirect).
```
lb-, line 6: Not 100% sure but just trying to reword this to be a bit clearer that we are presenting what the tutorial does vs what is already documents.

```suggestion
Instead of showing the `render_landing_page` content on the POST response we can instead redirect to an entirely different URL with that same content managed by our `FormPage` class.

An alternative approach to this, that does not use `RoutablePageMixin` but instead redirects to an entirely different `Page` see [Custom landing page redirect](../reference/contrib/forms/customisation.html#custom-landing-page-redirect).
```

lb- (changes requested):
@nimasmi thanks so much for submitting this, I know it has been a while but I think this is looking really good, just a few recommended changes.

Let us know what you think.

docs/advanced_topics/formbuilder_routablepage_redirect.md
nimasmi: Thanks, I have mostly adopted this change.

docs/advanced_topics/formbuilder_routablepage_redirect.md
nimasmi: I like your suggestion, however I think it inadvertently suggests that Wagtail FormPages' default behaviour is the correct one. I have retained my ordering to emphasise that FormPages are in the deficient minority here, but separated the sentences as you had, which is much clearer.

docs/advanced_topics/formbuilder_routablepage_redirect.md
nimasmi: Thanks, I've mostly adopted this change.

lb- (approved):
looks good to me

lb-:
Just tested the code out as is (copy/paste) and works great.
Note network tab - 302 redirect to new 200 for the new URL under sub-route.
Screen Shot 2021-10-04 at 2 04 20 pm

lb-:
Pushed some refinements (whitespace change, changelog, added page to index.rst) to the branch, rebased and merged in.

Issue #6793, opened 2021-02-05
Documentation design changes


nmorduch:
## Done during sprint

- [x] TOC, hide show in desktop/mobile
- [x] TOC, hide om some pages (index/landing pages)
- [x] Remove quote block wrapping code.
- [x] Version picker.
- [x] Version warning banner check (is this JS injected?). Checked. Yes, JS
- [x] Remove Typo3 refs hardcoded and via conf.
- [x] White on white text. /advanced_topics/images/custom_image_model.html
- [x] Load Roboto font

## Should

- [x] Increase total width
- [ ] Version picker styling improvements.
- [ ] TOC, to the right, sticky and scrollable, if possible.
- [ ] Branding, font-size, colour scheme. See [Olly's design](https://www.figma.com/proto/hFjWqQeEjGd8NtTsEHcc4i/Wagtail-Docs?node-id=1%3A2&viewport=983%2C953%2C1&scaling=min-zoom&hide-ui=1)
- [ ] Beter understand template overrides, maybe override only some blocks and not complete templates.

# Changes to #6792

 From Slack

- [x] Remove view source button
- [ ] the contrast of the inline code elements needs to be much better
- [ ] Should the main navigation on the left be sticky, too? Decision needed.
- [ ] Change the "system" font stack on body to include a better system font than Arial on Windows. Use 'Segoe UI'.
- [ ] syntax highlighting failing to initialize due to a JS syntax error. there is some template syntax issue here: https://github.com/allcaps/wagtail/blob/docs-theme/docs/_templates/layout.html#L175-L178
- [ ] Enable default styling so code is readable even if the syntax highlighting fails to initialise.
- [x] Tone down the github button (could style it like the view source button)
- [ ] Reduce the padding of the code blocks so they don’t nearly crash into each other vertically. if we can do this and slightly increase the font size of the blocks that’d be great.
- [ ] Increase line height
- [ ] Slightly reduce the blackness of the large code blocks (they feel a bit heavy)
- [ ] Consider going a bit less black with the Page contents section, to make it less distracting and indicate that it’s ‘secondary’ content
- [ ] Colours of notes and info blocks clash with Wagtail teal
- [x] If possible, remove the Page contents block when there’s only one heading
- [x] Hover state in navigation is missing.
- [ ] Tab focus styling
[...]

PR #6833, opened 2021-02-25
Docsearch w/ Search Results Page

tbrlpld:
I have build on #6753 by @thibaudcolas and extended it to generate a search results page from the Algolia Docsearch results. 
The search results page has the advantage of being available as a link target (and you can keep it open in your browser as you check the different results). 

This is only an extension of the Algolia dropdown. The dropdown still exists and shows the results as you type. Once you hit `enter` the results page is loaded. 

I have tried to keep the styling of the results page close to the "Read the docs" search. 

Currently, only the first 50 results are returned and no pagination is enabled. 

Please let me know what you think and if any changes are needed. 

---

Here is a screenshot of a results page:
Results page screenshot

### DIFF
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -41,6 +41,8 @@
         facetFilters: [getVersionFacetFilter()],
       },
       autocompleteOptions: {
+        // Do NOT automatically select the first suggestion in the dropdown.
+        // https://github.com/algolia/autocomplete/blob/45fa32d008620cf52bf4a90530be338543dfba7f/README.md#global-options
         autoSelect: false
       },
       debug: true // Set debug to true if you want to inspect the dropdown
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -29,6 +29,15 @@
     return 'version:' + getReadTheDocsVersion()
   }
 
+  /**
+   * Return true (debug: on) when not hosted on Readthedocs.
+   *
+   * The debug mode allows inspection of the dropodown.
+   */
+  function getSearchDebugMode() {
+    return window.READTHEDOCS_DATA === undefined
+  }
+
   function docSearchReady(script) {
     /**
     * Configure Algolia DocSearch.
@@ -45,7 +54,7 @@
         // https://github.com/algolia/autocomplete/blob/45fa32d008620cf52bf4a90530be338543dfba7f/README.md#global-options
         autoSelect: false
       },
-      debug: true // Set debug to true if you want to inspect the dropdown
+      debug: getSearchDebugMode(),
     });
 
     return search
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -11,6 +11,7 @@
 {% block footer %}
 {{ super() }}
 <script async defer data-domain="docs.wagtail.io" src="https://plausible.io/js/plausible.js"></script>
+<script src="{{ docsearch_base }}docsearch.min.js"></script>
 <script>
   /**
    * Get version of the currently served docs.
@@ -60,9 +61,6 @@
     return search
   }
 
-  const s = document.createElement('script')
-  s.src = '{{ docsearch_base }}docsearch.min.js'
-  s.onload = docSearchReady
-  document.body.appendChild(s)
+  window.onload(docSearchReady())
 </script>
 {% endblock %}
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -23,7 +23,7 @@ function docSearchReady(script) {
   const rtd_version = (window.READTHEDOCS_DATA || {}).version || 'latest';
   const version = rtd_version.match(/^\d+$/) ? 'latest' : rtd_version;
 
-  docsearch({
+  const search = docsearch({
     apiKey: '8325c57d16798633e29d211c26c7b6f9',
     indexName: 'wagtail',
     inputSelector: '#rtd-search-form [name="q"]',
@@ -35,6 +35,8 @@ function docSearchReady(script) {
     },
     debug: true // Set debug to true if you want to inspect the dropdown
   });
+
+  return search
 }
 
 const s = document.createElement('script');
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -1,5 +1,10 @@
 {% extends "!layout.html" %}
 
+{% block extrahead %}
+<link rel="stylesheet" href="{{ docsearch_base ~ 'docsearch.min.css' }}"/>
+<link rel="stylesheet" href="{{ pathto('_static/css/docsearch.overrides.css', 1) }}" />
+{% endblock %}
+
 {% block footer %}
 {{ super() }}
 <script async defer data-domain="docs.wagtail.io" src="https://plausible.io/js/plausible.js"></script>
@@ -17,12 +22,18 @@ function docSearchReady(script) {
    */
   const rtd_version = (window.READTHEDOCS_DATA || {}).version || 'latest';
   const version = rtd_version.match(/^\d+$/) ? 'latest' : rtd_version;
+
   docsearch({
     apiKey: '8325c57d16798633e29d211c26c7b6f9',
     indexName: 'wagtail',
     inputSelector: '#rtd-search-form [name="q"]',
-    algoliaOptions: { 'facetFilters': [`version:${version}`] },
-    debug: true
+    algoliaOptions: {
+      facetFilters: ['version:' + version],
+    },
+    autocompleteOptions: {
+      autoSelect: false
+    },
+    debug: true // Set debug to true if you want to inspect the dropdown
   });
 }
 
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -12,14 +12,14 @@
 {{ super() }}
 <script async defer data-domain="docs.wagtail.io" src="https://plausible.io/js/plausible.js"></script>
 <script>
+  /**
+   * Get version of the currently served docs.
+   *
+   * PR builds have their version set to the PR ID (e.g. "6753").
+   * If the docs are built for a PR, use the "latest" search index.
+   * Otherwise, use the search index for the current version.
+   */
   function getReadTheDocsVersion() {
-    /**
-    * Get version of the currently served docs.
-    *
-    * PR builds have their version set to the PR ID (e.g. "6753").
-    * If the docs are built for a PR, use the "latest" search index.
-    * Otherwise, use the search index for the current version.
-    */
     const rtd_version = (window.READTHEDOCS_DATA || {}).version || 'latest';
     const version = rtd_version.match(/^\d+$/) ? 'latest' : rtd_version;
     return version
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -20,8 +20,8 @@
    * Otherwise, use the search index for the current version.
    */
   function getReadTheDocsVersion() {
-    const rtd_version = (window.READTHEDOCS_DATA || {}).version || 'latest';
-    const version = rtd_version.match(/^\d+$/) ? 'latest' : rtd_version;
+    const rtd_version = (window.READTHEDOCS_DATA || {}).version || 'latest'
+    const version = rtd_version.match(/^\d+$/) ? 'latest' : rtd_version
     return version
   }
 
@@ -55,14 +55,14 @@
         autoSelect: false
       },
       debug: getSearchDebugMode(),
-    });
+    })
 
     return search
   }
 
-  const s = document.createElement('script');
-  s.src = '{{ docsearch_base }}docsearch.min.js';
-  s.onload = docSearchReady;
-  document.body.appendChild(s);
+  const s = document.createElement('script')
+  s.src = '{{ docsearch_base }}docsearch.min.js'
+  s.onload = docSearchReady
+  document.body.appendChild(s)
 </script>
 {% endblock %}
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -1,5 +1,8 @@
 {% extends "!layout.html" %}
 
+{% set docsearch_version = '2' %}
+{% set docsearch_base = 'https://cdn.jsdelivr.net/npm/docsearch.js@' ~ docsearch_version ~ '/dist/cdn/' %}
+
 {% block extrahead %}
 <link rel="stylesheet" href="{{ docsearch_base ~ 'docsearch.min.css' }}"/>
 <link rel="stylesheet" href="{{ pathto('_static/css/docsearch.overrides.css', 1) }}" />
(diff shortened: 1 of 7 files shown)


docs/_templates/search.html
```diff
@@ -0,0 +1,109 @@
...
+}
+
+function extractPageLink(url) {
+  const endPageUrlIndex = url.lastIndexOf('#')
+  return url.slice(0, endPageUrlIndex)
+}
+
+window.onload = runSearchPageSearch
+</script>
+{% endblock %}
```
thibaudcolas, line 109:
```suggestion
{% endblock %}

```

docs/_templates/search.html
```diff
@@ -0,0 +1,109 @@
...
+
+  return hitListElement
+}
+
+function extractPageLink(url) {
+  const endPageUrlIndex = url.lastIndexOf('#')
+  return url.slice(0, endPageUrlIndex)
+}
+
+window.onload = runSearchPageSearch
```
thibaudcolas, line 107: This should be on a `DOMContentLoaded` listener rather than override `onload`.

docs/_templates/search.html
```diff
@@ -0,0 +1,109 @@
...
+      facetFilters: [getVersionFacetFilter()]
+    }
+  )
+  .then(({ hits }) => addResultsList(hits, query, searchResultsContainer))
+  .catch((error) => console.log(error))
+}
+
+function addHeadingForQuery(query, parentElement) {
+  const searchHeading = document.createElement('h2')
+  searchHeading.textContent = `Search Results for "${query}"`
```
thibaudcolas, line 44:
```suggestion
  searchHeading.textContent = `Search results for “${query}”`
```

docs/_templates/search.html
```diff
@@ -0,0 +1,109 @@
...
+      hitsPerPage: 50,
+      facetFilters: [getVersionFacetFilter()]
+    }
+  )
+  .then(({ hits }) => addResultsList(hits, query, searchResultsContainer))
+  .catch((error) => console.log(error))
+}
+
+function addHeadingForQuery(query, parentElement) {
+  const searchHeading = document.createElement('h2')
```
thibaudcolas, line 43:
```suggestion
  const searchHeading = document.createElement('h1')
```

I assume you made this a `h2` to match the current implementation but really should be `h1`.

docs/_templates/search.html
```diff
@@ -0,0 +1,109 @@
...
+  searchHeading.textContent = `Search Results for "${query}"`
+  parentElement.appendChild(searchHeading)
```
thibaudcolas, line 45: It would be nice if we could set `document.title` to the same text, but only if we can keep the title’s suffix with the documentation version.

docs/_templates/search.html
```diff
@@ -0,0 +1,109 @@
...
+  const hitListElement = document.createElement('li')
+
+  let firstHierarchyLevel = undefined
+  let lastHierarchyLevel = undefined
+  for (levelLabel in result.hierarchy) {
+    if (firstHierarchyLevel === undefined ) {
+      firstHierarchyLevel = result.hierarchy[levelLabel]
+    }
+    lastHierarchyLevel = result.hierarchy[levelLabel]
+  }
```
thibaudcolas, line 73: I’m not entirely sure why this is needed (would warrant a comment) but feels overkill to get the first and last item.

```suggestion
  const hierarchies = Object.values(result.hierarchy);
  const firstHierarchyLevel = hierarchies[0];
  const lastHierarchyLevel = hierarchies[hierarchies.length - 1];
```

docs/_templates/search.html
```diff
@@ -0,0 +1,109 @@
...
+    const hitElement = createHitElement(hit, query)
+    searchResultsList.appendChild(hitElement)
+  }
+  parentElement.appendChild(searchResultsList)
+}
+
+function createHitElement(hitData, query) {
+  const pageURL = extractPageLink(hitData.url) + '?highlight=' + query
+  const anchor = hitData.anchor
+  const anchorURL = pageURL + '#' + anchor
```
thibaudcolas, line 61: I think we should use the URL API directly for this:

```suggestion
  const pageURL = new URL(hitData.url)
  pageURL.hash = '';
  pageURL.searchParams.set('highlight', query);
  const anchorURL = new URL(hitData.url);
  anchorURL.searchParams.set('highlight', query);
```

Otherwise we’d need to add URL encoding.

docs/_templates/searchbox.html
```diff
@@ -0,0 +1,6 @@
+<div role="search">
+  <form id="rtd-search-form" class="wy-form" action="{{ pathto('search') }}" method="get">
+    <input type="text" name="q" placeholder="{{ _('Search docs') }}" />
```
thibaudcolas, line 3:
```suggestion
    <input type="search" name="q" placeholder="{{ _('Search docs') }}" />
```

In addition I think this template would warrant a comment to explain what we’re overriding from the built-in template.

docs/_templates/searchbox.html
```diff
@@ -0,0 +1,6 @@
+<div role="search">
+  <form id="rtd-search-form" class="wy-form" action="{{ pathto('search') }}" method="get">
```
thibaudcolas, line 2: The search role isn’t valid on a `div`. Hopefully no styles are attached to this? If possible we should also remove the unneeded div.

```suggestion
<div>
  <form id="rtd-search-form" class="wy-form" action="{{ pathto('search') }}" method="get" role="search">
```

docs/conf.py
```diff
@@ -187,7 +187,7 @@
 # html_domain_indices = True
 
 # If false, no index is generated.
-# html_use_index = True
+html_use_index = False
```
thibaudcolas, line 190: I don’t understand how this relates to the other changes?

thibaudcolas (changes requested):
Thank you @tbrlpld! This is looking pretty amazing in the preview. I’ve suggested a few changes, most around code structure, a few around accessibility.

thibaudcolas:
@tbrlpld I’ll remove the "WIP" in the PR title as I think it looks ready-to-ship to me, minus the code review feedback of course. Was there anything you thought was left to do? Any nice-to-have we should consider in a future PR?

docs/conf.py
chosak: If I recall, this determines whether Sphinx generates an index of its own ([docs here](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-html_use_index)). If we're using Docsearch then we don't need to do that - I suppose it doesn't hurt to leave this on, although removing it might incrementally speed up the build.

docs/_templates/layout.html
```diff
@@ -1,6 +1,57 @@
...
+      debug: true // Set debug to true if you want to inspect the dropdown
+    });
+
+    return search
+  }
+
+  const s = document.createElement('script');
+  s.src = '{{ docsearch_base }}docsearch.min.js';
+  s.onload = docSearchReady;
+  document.body.appendChild(s);
```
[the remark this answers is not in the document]
tbrlpld: This is how it came in the setup from @chosak. It took me a moment to even figure out what is going on there. 

@chosak Can you maybe explain the thought behind this setup?

docs/_templates/layout.html
```diff
@@ -1,6 +1,57 @@
...
+    */
+    const search = docsearch({
+      apiKey: '8325c57d16798633e29d211c26c7b6f9',
+      indexName: 'wagtail',
+      inputSelector: '#rtd-search-form [name="q"]',
+      algoliaOptions: {
+        facetFilters: [getVersionFacetFilter()],
+      },
+      autocompleteOptions: {
+        autoSelect: false
```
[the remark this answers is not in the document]
tbrlpld: Do you refer to the `facetFilters` or the `autocompleteOptions`? Or both?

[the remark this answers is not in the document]
tbrlpld: I have added a comment for the `autoSelect` in 282bc64318d94cf17df5ec4bfd07d199be185669. 

I did not know before either (that also came from @chosak ). It is actually a quite necessary option to have to keep the search bar triggering the search page on `Enter`.

docs/_templates/layout.html
```diff
@@ -1,6 +1,57 @@
...
+      },
+      debug: true // Set debug to true if you want to inspect the dropdown
```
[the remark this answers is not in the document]
tbrlpld: Can we define the contents of `READTHEDOCS_DATA`? I currently fail to see and option to distinguish between production and PR preview. 

For now I will enable debug when not on Read The Docs (`window.READTHEDOCS_DATA === undefined`)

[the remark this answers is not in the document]
tbrlpld: First step is added in e9d21581bfbffe24632efadd875af18b0553185f, but that does not differentiate between preview and production.

docs/_templates/layout.html
```diff
@@ -1,6 +1,57 @@
...
+    /**
+    * Get version of the currently served docs.
+    *
+    * PR builds have their version set to the PR ID (e.g. "6753").
+    * If the docs are built for a PR, use the "latest" search index.
+    * Otherwise, use the search index for the current version.
+    */
+    const rtd_version = (window.READTHEDOCS_DATA || {}).version || 'latest';
+    const version = rtd_version.match(/^\d+$/) ? 'latest' : rtd_version;
+    return version
```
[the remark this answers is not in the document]
tbrlpld: All semicolons removed in 03098b5ebe464be41debac118e9d7f822fae6ce5

docs/_templates/search.html
tbrlpld: Oh yes. That could have used some refactoring. I thought at first that the properties would exist but empty. 

Your way is obviously way more elegant!

docs/_templates/searchbox.html
tbrlpld: This section came from @chosak

docs/_templates/searchbox.html
tbrlpld: Seems like it works without the `div` and the role moved.

tbrlpld:
> @tbrlpld I’ll remove the "WIP" in the PR title as I think it looks ready-to-ship to me, minus the code review feedback of course. Was there anything you thought was left to do? Any nice-to-have we should consider in a future PR?

@thibaudcolas I think pagination could be nice, in case somebody needs more than 50 results. Also styling of the search dropdown. But I think that might be better added in #6793

docs/conf.py
thibaudcolas: Thank you. I think this would be worth an additional comment.

docs/_templates/layout.html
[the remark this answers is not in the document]
thibaudcolas: Thank you.

docs/_templates/layout.html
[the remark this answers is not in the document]
thibaudcolas: Thank you. To distinguish between preview and production, I think the easiest would be to do a version check similar to what we do here:

```
const rtd_version = (window.READTHEDOCS_DATA || {}).version || 'latest'
const version = rtd_version.match(/^\d+$/) ? 'latest' : rtd_version
```

If the version is a number, then it means this is a PR preview build, and `debug` should be true.

docs/_templates/layout.html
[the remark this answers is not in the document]
thibaudcolas: This PR was something I threw together as quickly as I could and then Andy took over, there is no reason why we’d want the `script` tag to be defined in JS like this.

docs/_templates/searchbox.html
thibaudcolas: @tbrlpld could you compare this with the built-in template and check what the differences might be? If there are none, we might as well keep the built-in template.

docs/_templates/layout.html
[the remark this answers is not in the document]
tbrlpld: Moved to separate `script` tag in c5ef5400e514c0d2c77c636f67384b6548f5f02b

tbrlpld:
One more nice-to-have:
Currently, when loading the page, the search bar element show a visual jump. I guess this is when our overriding element replaces the original one. For some reason, our overridden searchbar is smaller than the original one. 

I think it might be nicer to not show the original one at all, or make sure that our search bar looks the same to avoid the jump on loading.

tomdyson:
This is looking great! Thanks @tbrlpld, @thibaudcolas and @chosak.

> I have tried to keep the styling of the results page close to the "Read the docs" search.

I dislike the standard RTD search term highlighting, and I usually remove the `?highlight=video` param before sharing search results, e.g. in responses to support questions on Slack. I'd be in favour of removing that styling here - what do others think?

docs/_templates/layout.html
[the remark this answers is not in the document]
tbrlpld: I have added the check in 4b8eb8405d88641a2e71908e1f6143a1059f92ae

tbrlpld:
> I dislike the standard RTD search term highlighting, and I usually remove the `?highlight=video` param before sharing search results, e.g. in responses to support questions on Slack. I'd be in favour of removing that styling here - what do others think?

I agree the style is not great, but I appreciate the highlighting of the search terms. I often find myself doing a browser search on a resulting page for the term I have just searched 😅 Especially on longer pages. 

But I am open to remove it.

docs/_templates/searchbox.html
tbrlpld: This is the original from the current production site
```html
<div role="search">
  <form id="rtd-search-form" class="wy-form" action="../search.html" method="get">
    <input type="text" name="q" placeholder="Search docs">
    <input type="hidden" name="check_keywords" value="yes">
    <input type="hidden" name="area" value="default">
  </form>
</div>
```

It looks like this is where the `<div role=search>` is coming from.

tbrlpld: I guess we are removing the two `hidden` inputs and add a `submit` input.

tbrlpld: Original template: https://github.com/readthedocs/sphinx_rtd_theme/blob/master/sphinx_rtd_theme/searchbox.html

docs/conf.py
tbrlpld: I have added a comment in c25f8afb1ec8a0f9c9aa0a6a8f963378ae808cd2.

Let me know if that is expressive enough.

docs/_templates/search.html
```diff
@@ -0,0 +1,106 @@
...
+    contextElement.appendChild(contentElement)
+  }
+
+  return hitListElement
+}
+
+function extractPageLink(url) {
+  const endPageUrlIndex = url.lastIndexOf('#')
+  return url.slice(0, endPageUrlIndex)
+}
```
thibaudcolas, line 102:
```suggestion
```

Looks like this is no longer used.

thibaudcolas (commented):
@tbrlpld could you take care of the 5 remaining suggested code changes? Then I think this is ready to be merged, and we can create a new issue for the nice-to-haves.

docs/_templates/search.html
tbrlpld: Yes, that's right. The elegant use of the URL API made this pointless.

tbrlpld:
> @tbrlpld could you take care of the 5 remaining suggested code changes? Then I think this is ready to be merged, and we can create a new issue for the nice-to-haves.

Maybe I missed it, but what are the other 4 suggested changes? I accepted the one regarding the unused function. I don't see anything else suggested in GitHub.

docs/_templates/search.html
tbrlpld: Yup. 👍

docs/_templates/search.html
tbrlpld: That makes way more sense!

tbrlpld: Fixed in 7b1b31a112db20e609c193f3bb767afdd71804ee

tbrlpld:
Sorry, this was my first larger review on Github. I was expecting to see the requested changes in this overview page. I did not know they are all visible on the "Files changed" page.

@thibaudcolas I have updated according to you suggestions.

docs/_templates/search.html
tbrlpld: New line already added.

tbrlpld (commented):
@thibaudcolas changes according to review are added.

docs/_templates/layout.html
```diff
@@ -1,6 +1,77 @@
...
+      // like 'stable', 'latest', 'v2.12', etc. In that case `Number()` is `NaN`.
+      const versionNumber = Number(window.READTHEDOCS_DATA.version)
+      debug = !isNaN(versionNumber)
+    }
+    return debug
+  }
+
+  function docSearchReady(script) {
+    /**
+    * Configure Algolia DocSearch.
```
thibaudcolas, line 55:
```suggestion
    * Configure Algolia DocSearch.
    * See https://github.com/algolia/docsearch-configs/blob/master/configs/wagtail.json for index configuration.
```

thibaudcolas (commented):
Cosmetic changes to comments

tbrlpld (commented):
Looks good. Definitely makes the comments more useful.

tbrlpld:
@thibaudcolas Not sure why, but I see the message "1 change requested", but I can't find what change that is. All conversations are resolved as far as I can see.

thibaudcolas (approved):
Good question @tbrlpld – I think this is good to go 😄

thibaudcolas (approved):
Good question @tbrlpld – I think this is good to go 😄

thibaudcolas:
Well, this was so good to go I even approved it twice.

Here are follow-up issues from the nice-to-haves above:

- https://github.com/wagtail/wagtail/issues/6885
- https://github.com/wagtail/wagtail/issues/6886

Thank you for taking this through @tbrlpld!

Stormheg:
This is the first time I'm trying the Algolia Docsearch introduced in this PR and these are my thoughts. 

- The autocomplete feature is a very helpful and one that we were lacking - good to have this now! :tada: 
- I wish the search results would show more of the actual content of the docs (like google search results), not just the headings. Not sure if that is possible with Docsearch.
image
Search results with descriptions from Google:
image

tbrlpld:
@Stormheg Yea, I know what you mean. This is depended on the results that come back from Docsearch. If you look in the PR message, there is a screenshot where I searched for "test". The lower results show a body text. 

I think this has to do with where the term was found. If there is a page with the term in the title, I will only return the heading and nothing more. 

Maybe there is an option in Docsearch that always requests a body snippet too, but I have not looked for that yet.

PR #6839, opened 2021-02-26
Docs: Convert "getting started - demo site" from RST to MD

tbrlpld:
Partially replaces #6834 

GitHub diff does not quite work with renamed files. The following `git` command is nicer to identify the changes:
```sh
$ git diff master:./docs/getting_started/demo_site.rst ./docs/getting_started/demo_site.md
```

PR #6840, opened 2021-02-26
Docs: Convert "getting started - integrating into django" from RST to MD

tbrlpld:
Partially replaces #6839 

Diff locally with 
```
git diff master:./docs/getting_started/integrating_into_django.rst ./docs/getting_started/integrating_into_django.md
```


docs/getting_started/integrating_into_django.md
```diff
@@ -0,0 +1,121 @@
...
+
+Wagtail provides the `wagtail start` command and project template to get you started with a new Wagtail project as quickly as possible, but it\'s easy to integrate Wagtail into an existing Django project too.
+
+Wagtail is currently compatible with Django 2.2, 3.0 and 3.1. First, install the `wagtail` package from PyPI:
+
+```sh
+$ pip install wagtail
+```
+
+or add the package to your existing requirements file. This will also install the **Pillow** library as a dependency, which requires libjpeg and zlib - see Pillow\'s [platform-specific installation instructions](https://pillow.readthedocs.org/en/latest/installation.html#external-libraries).
```
thibaudcolas, line 14: Are these escapes really needed? If so I think I’d prefer we use the correct apostrophes instead, to keep the source Markdown easier to read.

```suggestion
or add the package to your existing requirements file. This will also install the **Pillow** library as a dependency, which requires libjpeg and zlib - see Pillow’s [platform-specific installation instructions](https://pillow.readthedocs.org/en/latest/installation.html#external-libraries).
```

tbrlpld: Right. Sorry, that's how it's coming out of Pandoc. I will remove those.

thibaudcolas (approved):
Looking good :)

Issue #6641, opened 2020-12-15
Pass the focal point information through to the image tag


helenb:
### Is your proposal related to a problem?

I frequently use images in wagtail templates as background images, for example:

```
{% image image width-765 as card_image %}
<div class="card__image" style="background-image:url('{{ card_image.url }}')"></div>
```

The CSS might then crop the image as follows:

```
height: 250px;
width: 30%;
background-position: center center;
background-repeat: no-repeat;
background-size: cover;
```

The result will be an image that is cropped differently at different screen sizes. This frequently causes issues as we do not know where to set the focal point of the image (in the example above, the centre of the image is used). Sometimes if an image of a person is used, their head might get cropped off at certain screen sizes.

### Describe the solution you'd like

When users upload images to wagtail, they get to set a focal area. This is used when generating image renditions to ensure that the most important area of the image is not cropped. Could this information be passed through to the image tag in a such a was that it could be used in a CSS `background-position` property to set a focal point? For example if the focal point was 20% from the left edge of the image and 45% from the top of the image, we'd want to end up with:

```
background-position: 20% 45%;
```

It would also be possible for this information to be set in px. See https://developer.mozilla.org/en-US/docs/Web/CSS/background-position.

 I presume this would need to be set once the rendition had been generated, as otherwise it might be incorrect. It would need to be based on the centre point of the focal area. Ideally I'd then be able to use code like this:

```
{% image image width-765 as card_image %}
<div class="card__image" style="background-image:url('{{ card_image.url }}') background-position:{{ card_image.focalpoint_x}}{{ card_image.focalpoint_y}}"></div>
```

### Describe alternatives you've considered

(Write your answer here.)

### Additional context

(Write your answer here.)

thenewguy:
I haven't investigated this but off hand from memory can't you use the backwards relation to the image model for the data you seek?

Like `card_image.image.focal_point_x` - Pretty sure you've just got a reference to the rendition

gasman:
@thenewguy As I understand it, the problem is that the process of cropping/resizing to generate the rendition is liable to change the position of the focal point _within that particular rendition_ in a hard-to-calculate way, and this adjusted focal point is the one we need for the CSS logic.

thenewguy:
@gasman Ah - I read the example with one of the proportional scaling filters in mind - in this case it should be straightforward to determine the focal point (maybe no change necessary with % units?) - but if the focal point is being obfuscated I guess that would be more challenging =)

[...]

PR #6843, opened 2021-02-26
Image transforms

kaedroho:
Fixes https://github.com/wagtail/wagtail/issues/6641

This separates the 'transform' image operations from the 'filter' so that we can calculate the transform separately and apply this to the focal point on its own, allowing us to calculate a new focal point to be passed into the image tag on the frontend.

I originally tried to implement this using an affine transform matrix and perform all the transformations with a single call Pillow's ``transform()`` method. However, the resampling was pretty horrible (see the middle image on https://pythontic.com/image-processing/pillow/rotate) so I replaced this with tracking the offset and scale instead and use Pillow's ``crop()``/``resize()`` methods.

### DIFF
--- a/wagtail/images/image_operations.py
+++ b/wagtail/images/image_operations.py
@@ -1,7 +1,7 @@
 import inspect
 
 from wagtail.images.exceptions import InvalidFilterSpecError
-from wagtail.images.rect import Rect
+from wagtail.images.rect import Rect, Vector
 from wagtail.images.utils import parse_color_string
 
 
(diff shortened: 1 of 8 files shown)


kaedroho:
@jacobtoppm, will you be able to find time to review this soon?

kaedroho:
@helenb 

I think we could add a ``.background_position_style`` attribute to make it easier to add the styling into the tag:

```
    <img
        src="{{ myrendition.url }}"
        alt="{{ myimage.title }}"
        {% if myrendition.focal_point %}
            style="{{ myrendition.focal_point.background_position_style }}"
        {% endif %}
    />
```

and this would generate a style like: ``style="background-position: 20% 80%;"``

What do you think?

helenb:
> @helenb
> 
> I think we could add a `.background_position_style` attribute to make it easier to add the styling into the tag:
> 
> ```
>              src="{{ myrendition.url }}"
>         alt="{{ myimage.title }}"
>         {% if myrendition.focal_point %}
>             style="{{ myrendition.focal_point.background_position_style }}"
>         {% endif %}
>     />
> ```
> 
> and this would generate a style like: `style="background-position: 20% 80%;"`
> 
> What do you think?

Hi @kaedroho, that would be brilliant and certainly would fit my use case.

kaedroho:
Implemented this now. I've put this directly on the rendition since we need to know the size of the rendition in order to calculate the percentages: ``style="{{ myrendition.background_position_style }}"``

wagtail/images/image_operations.py
```diff
@@ -29,8 +29,97 @@ def construct(self, *args):
...
+        Returns a new vector.
+        """
+        return Vector(
+            (vector.x + self.offset[0]) * self.scale[0],
+            (vector.y + self.offset[1]) * self.scale[1]
+        )
+
+    def untransform_vector(self, vector):
+        """
+        Transforms a back to the images coordinate space.
```
jacobtoppm, line 92: A few words missing here. In addition, in this and the `transform_vector` docstring, I think it's worth noting what space it's being transformed to (presumably real space?)

wagtail/images/image_operations.py
```diff
@@ -29,8 +29,97 @@ def construct(self, *args):
...
+
+    def get_rect(self):
+        """
+        Returns a Rect representing the region of the original image to be cropped.
+        """
+        return Rect(
+            -self.offset[0],
+            -self.offset[1],
+            -self.offset[0] + self.size[0] / self.scale[0],
+            -self.offset[1] + self.size[1] / self.scale[0]
```
jacobtoppm, line 109: Should this not be `self.scale[1]` to scale by the correct dimension?

wagtail/images/image_operations.py
```diff
@@ -29,8 +29,97 @@ def construct(self, *args):
 # Transforms
 
 
+class ImageTransform:
+    """
+    Tracks transformations that are performed on an image.
+
+    This allows multiple transforms to be processed in a single operation and also
+    provides a matrix that can be used to transform the focal point of the image.
```
jacobtoppm, line 37: Where is the matrix here?

wagtail/images/models.py
```diff
@@ -579,6 +579,28 @@ def focal_point(self):
             transform = self.filter.get_transform(self.image)
             return image_focal_point.transform(transform)
 
+    @property
+    def background_position_style(self):
+        """
+        Returns a `background-position` rule to be put in the inline style of an element which uses the rendition for its background.
+
+        This positions the rendition according to the value of the focal point. This is helpful for when the element does not have
+        the same aspect ratio as the rendtion.
```
jacobtoppm, line 588: 'rendtion` -> `rendition`

jacobtoppm (changes requested):
Generally looks great! A few typos/points/questions and should be good to go

wagtail/images/image_operations.py
kaedroho: Oops, yep it should be! This wouldn't have any effect on our built-in operations as none of them can change the aspect ratio

wagtail/images/image_operations.py
kaedroho: Oops, I removed the matrix (reasoning in PR description)

kaedroho:
Thanks @jacobtoppm. I've fixed the issues that you've spotted and squashed them back into the original commits

jacobtoppm (approved):
LGTM!

PR #6844, opened 2021-02-26
fix small youtube embed - allow to override max_width/max_height in EmbedBlock

PetrDlouhy:
YouTube embeds are ending up to be very small:
Snímek obrazovky_2021-02-26_17-42-48

This is because YouTube oEmbed changed defaults to smaller size.
This PR allows to set `max_width` and `max_height` parameters to EmbedBlock init in order to set up bigger embeds.

### DIFF
--- a/docs/advanced_topics/embeds.rst
+++ b/docs/advanced_topics/embeds.rst
@@ -33,6 +33,9 @@ nest the embed code.
 The :class:`~wagtail.embeds.block.EmbedBlock` block type allows embeds
 to be placed into a ``StreamField``.
 
+The `max_width` and `max_height` init attributes will be passed to
+the oEmbed as `maxwidth` and `maxheight`.
+
 For example:
 
 .. code-block:: python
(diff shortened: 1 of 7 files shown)


gasman:
Thanks @PetrDlouhy! It looks like there are a few test failures here, and a couple of debugging `print` statements have been left in - would you be able to take another look at this please?

PetrDlouhy:
Hi @gasman, everything is done.

gasman:
Thanks @PetrDlouhy! I think this is a worthwhile change, but we'll need to take additional care not to break backwards compatibility with embed finders defined outside of the Wagtail codebase (for example, people using [the custom Youtube finder from this blog post](https://wagtail.io/blog/fixing-youtubes-oembed-implementation-custom-finder-class/)) - as it stands, any finder that isn't expecting to be passed a `max_height` parameter will fail. This can hopefully be fixed by checking the `find_embed` function with [`accepts_kwarg`](https://github.com/wagtail/wagtail/blob/b1b69360a334b947881679e044fdc8c14fead658/wagtail/core/utils.py#L139) before calling it.

PetrDlouhy:
@gasman I have added the compatibility code now.

gasman:
Thanks again! Merged in 50e4f5151d5b6517725eae295732640a55e53f2f.

PR #6845, opened 2021-02-27
DeferredSpecificIterable efficiency improvements

ababic:
Now that we have `Page.get_specific()` in place, the `PageQuerySet.specific(defer=True)` option can be updated to utilise that, preventing unnecessary queries on evaluation.

NOTE: Because we're now only evaluating a single queryset, we can use `ModelIterable` as a base - meaning `specific(defer=True)` will now work happily in combination with `select_related()`, `prefetch_related()`. Though, I don't think this is something we should shout too much about until we have like-for-like behaviour whether `defer=True` is provided or not.


jacobtoppm (approved):
Nice reduction in queries! Looks great to me, with the caveat that I believe this addresses the comment https://github.com/wagtail/wagtail/blob/50a1cf2c3b5c056e27d80ecbf75d28ae0c7eb677/wagtail/core/query.py#L356, so maybe remove that?

PR #6846, opened 2021-02-27
Add PageQuerySet.defer_streamfields()

ababic:
Originally implemented in #6216.

Adds a `defer_streamfields()` method to PageQuerySet that can be used in situations where specific page instance are required, but only in a 'listing' capacity, where `Streamfield` values are not unlikely to be needed. For example: Page explorer views, API list views, sitemap generation.

It can also be used as an alternative to `specific(defer=True)` in cases where overridden methods are likely to be utilising specific field values, but the likelihood of those fields being streamfields is slim.


jacobtoppm (approved):
The code looks good to me! I feel like the utility is mostly already covered by the new defer specific option, or simply `only` for a version where you do want some child model data but only specific fields - but I think this could save a fair bit of boilerplate for people with huge streamfields, so probably a good addition anyway!

ababic:
@jacobtoppm Thank you matey!

PR #6847, opened 2021-02-27
Utilise PageQuerySet.defer_streamfields()

ababic:
Originally implemented in #6216

Utilises the newly implemented `PageQuerySet.defer_streamfields()` method to save resources in several places where `StreamField` values are unlikely to be needed.

### DIFF
--- a/wagtail/admin/views/chooser.py
+++ b/wagtail/admin/views/chooser.py
@@ -93,8 +93,8 @@ def browse(request, parent_page_id=None):
 
     parent_page = parent_page.specific
 
-    # Get children of parent page
-    pages = parent_page.get_children().specific()
+    # Get children of parent page (without streamfields)
+    pages = parent_page.get_children().defer_streamfields().specific()
 
     # allow hooks to modify the queryset
     for hook in hooks.get_hooks('construct_page_chooser_queryset'):
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -11,6 +11,7 @@ Changelog
  * Update default attribute copying behaviour of `Page.get_specific()` and added the `copy_attrs_exclude` option (Andy Babic)
  * Update `PageQueryset.specific(defer=True)` to only perform a single database query (Andy Babic)
  * Add `PageQueryset.defer_streamfields()` (Andy Babic)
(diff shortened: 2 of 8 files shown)


kaedroho:
I think this adds a lot of complexity to work around a performance issue that is probably better worked around with other means. For example, `get_admin_display_title` probably should be a field as we already discussed.

I'd really like us to stop using specific page models in listings in the Wagtail admin so these sort of optimisations wouldn't be necessary.

ababic:
@kaedroho Your point is valid - Wagtail should be aiming for efficiency at the feature level. However, I think it's worth highlighting that this is actually rather simple in terms of moving parts. And while we can control how Wagtail works, I come across things regularly in day-to-day work that would benefit from this kind of easy-to-apply / self-explanatory optimisation. For example:
- Generating navigation menus
- Index pages
- Curated / automated "You may also want to check out these pages..."  sections

wagtail/admin/views/chooser.py
```diff
@@ -93,8 +93,8 @@ def browse(request, parent_page_id=None):
 
     parent_page = parent_page.specific
 
-    # Get children of parent page
-    pages = parent_page.get_children().specific()
+    # Get children of parent page (without streamfields)
+    pages = parent_page.get_children().defer_streamfields().specific()
```
jacobtoppm, line 97: Question: is the reason you're using this rather than `specific(defer=True)` in case something like the `admin_display_title` method on the specific model actually references one of the deferred fields, causing a query per row? Or is there another consideration here?

ababic: You are totally right, yes. This is so that we can still effectively support those kinds of overrides for the time being. Hopefully we can avoid `specific()` completely at a later time, but for now, this should alleviate things somewhat

jacobtoppm: Hmm, looks like my previous comment didn't save! But anyway - in that case, that sounds like a good measure. I agree with Karl that it would be nice to remove specific usage entirely from the admin eventually, but this sounds like a really good idea in the short term that doesn't require the same kind of deprecation path.

jacobtoppm (approved):

PR #6849, opened 2021-02-28
Update non-field attribute copying behaviour in Page.get_specific()

ababic:
I have been thinking about this some more since introducing it not long ago, and I feel it makes more sense to copy all non-field attribute values to the newly returned object by default.

This *could* potentially catch people out, for example, if: 
- The subclass's `cached_property` method would normally return a different value on evaluation, but the version on the generic version has already been evaluated.
- An attribute on the original instance is referencing something large that would be garbage-collected if the reference weren't copied to the specific version.

However, in 99% cases I can think of, the `specific` or `specific_deferred` cached properties are the first thing that are called on generic instance, and any further interaction is with the returned value, so the above is unlikely to be a problem. As long the default behaviour is explained well, and we offer a convenient way to prevent copying of anything risky (as `copy_attrs_exclude` does), I think we should be good. 

It's also worth bearing in mind that the method was originally designed to be used in planned (at least be me) changes to `PageQuerySet.specific()`, where we quite obviously want annotations and other values to be preserved where possible. Technically, it would be possible for those methods to specify a full list of attributes to be copied, but I think that's unnecessarily verbose. Plus, when we (hopefully) get to supporting `select_related()` alongside `PageQuerySet.specific()`, that code will become more and more long-winded.


wagtail/core/models.py
```diff
@@ -1169,23 +1169,36 @@ def _update_descendant_url_paths(self, old_url_path, new_url_path):
...
+        from current object to the returned one. This includes:
+        * Values set by a queryset, for example: annotations, or values set as
+          a result of using ``select_related()`` or ``prefetch_related()``.
+        * Any ``cached_property`` values that have been evaluated.
+        * Attributes set elsewhere in Python code.
+
+        For fine-grained control over which non-field values are copied to the
+        returned object, you can use ``copy_attrs`` to specify a complete list
+        list of attribute names to include. Alternatively, you can use
+        ``copy_attrs_exclude`` to specify a list attribute names to exclude.
```
jacobtoppm, line 1201:
```suggestion
        ``copy_attrs_exclude`` to specify a list of attribute names to exclude.
```

wagtail/core/models.py
```diff
@@ -1225,10 +1238,17 @@ def get_specific(self, deferred=False, copy_attrs=None):
...
 
-        # Copy additional attribute values
-        for attr in copy_attrs or ():
-            if attr in self.__dict__:
+        # Copy non-field attribute values
+        if copy_attrs is not None:
+            for attr in (attr for attr in copy_attrs if attr in self.__dict__):
                 setattr(specific_obj, attr, getattr(self, attr))
+        else:
+            exclude = copy_attrs_exclude or ()
```
jacobtoppm, line 1246: We don't exclude the fields attributes here, so we're essentially resetting all the fields, so I'm not sure this "non-field" comment is accurate, as it's copying the fields too (unless I'm missing something)

jacobtoppm: If we wanted to prevent setting them again, we could always exclude anything in `._meta.get_fields()`, but I don't suppose there's much overhead from setting them

jacobtoppm (commented):
Looking good! Tiny docs tweak and a comment clarification

wagtail/core/models.py
ababic: Yes, I tried a few different versions of this. In the end, it seemed clearest to avoid too many conditionals and go with `setdefault` for the setting, which would naturally avoid resetting anything unnecessary. I've added a comment in to help make this clearer.

ababic:
Thanks for the review @jacobtoppm! much appreciated!

Issue #6758, opened 2021-02-03
Crash when copying alias pages


kaedroho:
When you try to copy an alias page, you get the following error:

```
RuntimeError at /admin/pages/107/copy/

save_revision() was called on an alias page. Revisions are not required for alias pages as they are an exact copy of another page.
```

We should instead prevent aliases from being copied. Or require copies to become aliases.

PR #6854, opened 2021-03-01
Fix crash when copying an alias page 

kaedroho:
Fixes https://github.com/wagtail/wagtail/issues/6758

This sets ``alias_of`` to null on all copied pages. So if someone copies an alias page, the new page will not be an alias.


jacobtoppm (approved):
LGTM!

kaedroho:
Thanks!

gasman:
Tagging for the next 2.11 (and 2.12) patch release, as this sounds like one that ought to be backported (unless I hear otherwise).

kaedroho:
Yep, I didn't realise there was going to be another 2.11 release, but I've already put it on the 2.12 branch!

PR #6856, opened 2021-03-02
Feature/redux comments

jacobtoppm:
This PR:
- Refactors field level comment widget to subscribe to the redux commenting state directly, so as to act more similarly to how commenting needs to be handled in Draftail
- Creates a new `CommentableEditor` component wrapping `DraftailEditor` as a controlled component when commenting is turned on, and adds the initial functionality for inline comments in Draftail
It does not include:
- Full styling for Draftail - text contrast is inadequate for instance (this will be done in a future ticket)
- Accessibility for comment annotations (though any suggestions on making comment highlights simultaneously editable and able to navigate easily to the comment in question are very welcome, I'm aware the current solution isn't yet adequate for this)
- Preventing comment decoration of links or other entities, so they remain accessible (https://github.com/wagtail/wagtail/pull/6861
- Darker highlights for overlapping comments (https://github.com/wagtail/wagtail/pull/6861)
- Saving and loading Draftail comments from the database (future ticket)
- Tests for CommentableEditor - these are needed, but it would be great to get an overall assessment here first

This should be reviewed with https://github.com/jacobtoppm/wagtail-comment-frontend/pull/7 `npm link`ed, and this should be reviewed at the same time

One issue that needs fixing: using the object spread operator for eg props seems to have broken eslint. Is this something we're not supporting, or do we need to adjust our config?


client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -0,0 +1,217 @@
+import { DraftailEditor, ToolbarButton, createEditorStateFromRaw, serialiseEditorStateToRaw } from 'draftail';
+import { EditorState, Modifier, RichUtils, SelectionState } from 'draft-js';
+import { filterInlineStyles } from "draftjs-filters";
```
thibaudcolas, line 3: This should be added as a direct dependency in `package.json` if you use it here. Draftail won’t stop depending this anytime soon, but it’s more correct to declare all dependencies.

client/src/components/CommentApp/comments.js
```diff
@@ -81,34 +124,73 @@ class FieldLevelCommentWidget {
...
+      }
+    });
     this.commentAdditionNode.addEventListener('click', () => {
-      makeComment(this.getAnnotationForComment(), this.contentpath);
+      // Make the widget button clickable to add a comment
+      const annotation = this.getAnnotationForComment();
+      const localId = this.commentApp.makeComment(annotation, this.contentpath);
+      annotation.subscribeToUpdates(localId);
     });
+    return unsubscribeWidget; // TODO: listen for widget deletion and use this
```
thibaudcolas, line 185: Should this TODO be addressed before merging?

client/src/components/CommentApp/comments.js
```diff
@@ -66,8 +102,15 @@ class BasicFieldLevelAnnotation {
...
-  setOnClickHandler(handler) {
-    this.node.addEventListener('click', handler);
+  setOnClickHandler(localId) {
+    this.node.addEventListener('click', () => {
+      this.commentApp.store.dispatch(
+        this.commentApp.actions.setFocusedComment(localId)
+      );
+      this.commentApp.store.dispatch(
+        this.commentApp.actions.setPinnedComment(localId)
+      );
```
thibaudcolas, line 112: I can see that in two cases those two actions are used at the same time. If that’s always the case, that should be a single action to not lead to potentially inconsistent state.

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -0,0 +1,217 @@
...
+import { STRINGS } from '../../../config/wagtailConfig';
+import Icon from '../../Icon/Icon';
+
+const COMMENT_STYLE_IDENTIFIER = 'COMMENT-'
+
+function usePrevious(value) {
+  const ref = useRef();
+  useEffect(() => {
+    ref.current = value;
+  }, [value]); // Only re-run if value changes
```
thibaudcolas, line 16:
```suggestion
  }, [value]);
```

On this very line I don’t think this comment is useful enough, but if you think it helps overall I’d suggest adding it as a JSDoc comment for the function.

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -0,0 +1,217 @@
+import { DraftailEditor, ToolbarButton, createEditorStateFromRaw, serialiseEditorStateToRaw } from 'draftail';
```
thibaudcolas, line 1: Now that we’re using TypeScript in Wagtail, wouldn’t you want to write this file in TypeScript?

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -0,0 +1,217 @@
...
+      this.inlineRefs = new Set()
+    }
+    addRef(ref) {
+      this.inlineRefs.add(ref);
+    }
+    removeRef(ref) {
+      this.inlineRefs.delete(ref);
+    }
+    getDesiredPosition() {
+      const nodeTops = Array.from(this.inlineRefs).map(ref => ref.current).filter(node => node).map(node => node.getBoundingClientRect().top)
```
thibaudcolas, line 32: This would warrant a comment. And it’s not super clear to me why the `filter` is needed? wouldn’t there not be a `ref` in the set if it has no value?

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -0,0 +1,217 @@
...
+      if (nodeTops.length > 0) {
+        return nodeTops.reduce((a,b) => a + b, 0)/nodeTops.length + document.documentElement.scrollTop
```
thibaudcolas, line 34: I think this would warrant either a comment, or a variable name to explain what this is? I’m surprised we’d want an average here.

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -0,0 +1,217 @@
...
+  return ({getEditorState, onChange}) => {
+    return <ToolbarButton
+      name='comment'
+      active={false}
+      title={STRINGS.ADD_A_COMMENT}
+      icon={<Icon name="comment"/>}
+      onClick={() => {
+        const annotation = new DraftailInlineAnnotation(fieldNode)
+        const commentId = commentApp.makeComment(annotation, contentPath);
+        onChange(RichUtils.toggleInlineStyle(getEditorState(), COMMENT_STYLE_IDENTIFIER + commentId))
```
thibaudcolas, line 65: Only use `+` for math.

```suggestion
        onChange(RichUtils.toggleInlineStyle(getEditorState(),`${COMMENT_STYLE_IDENTIFIER}${commentId}`))
```

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -0,0 +1,217 @@
...
+
+function getCommentDecorator(commentApp) {
+  const CommentDecorator = ({ contentState, children }) => {
+    // The comment decorator makes a comment clickable, allowing it to be focused.
+    // It does not provide styling, as draft-js imposes a 1 decorator/string limit, which would prevent comment highlights
+    // going over links/other entities
+    const blockKey = children[0].props.block.getKey()
+    const start = children[0].props.start
+
+    const commentId = useMemo(() => parseInt(contentState.getBlockForKey(blockKey).getInlineStyleAt(start).find((style) => style.startsWith(COMMENT_STYLE_IDENTIFIER)).slice(8)), [blockKey, start]);
```
thibaudcolas, line 83: Hmm this looks dangerous, but indeed I wouldn’t be too surprised if it was a necessary hack.

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -0,0 +1,217 @@
...
+      // that a previous style won't persist without filtering everywhere, which seems a bit too heavyweight
+      // This hack can be removed when draft-js triggers inline style rerender on props change
+      setEditorState(editorState => forceResetEditorState(editorState, 
+        Modifier.applyInlineStyle(
+          filteredContent,
+          getFullSelectionState(filteredContent),
+          'STYLE_RERENDER_'+uniqueStyleId
+        )
+      ))
+      setUniqueStyleId((id) => (id + 1) % 200);
```
thibaudcolas, line 183: Why modulo 200?

thibaudcolas (commented):
Bit of a drive-by code review, I’ll spend more time on this in the browser on Monday.

What’s the plan for unit tests? We have pretty good coverage in our React code currently.

client/src/components/CommentApp/comments.js
jacobtoppm: This isn't always the case - they're used separately in the commenting system itself. For widgets, it usually makes sense to do both together, so I could combine them in one utility action if you prefer for their use, but it doesn't lead to inconsistent state if done separately.

client/src/components/CommentApp/comments.js
jacobtoppm: This is only needed within streamfield. As @kaedroho is coming onto the project this week, and we're going to need to integrate with Telepath anyway, I think this is something to address when we integrate into the new streamfield system. The note is really just to make sure we do this before we go live.

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
jacobtoppm: Good point, thanks! Will address this

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
jacobtoppm: Good point, the filter is a holdout from when I hadn't added the unmount logic - I'll remove the filter and add a comment

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
jacobtoppm: I'll add a comment - you're right, it's not necessarily that clear. Why are you surprised we'd want an average, if you don't mind me asking? To me if you've got a few different highlights, it makes sense for the comment to try and float vertically in the middle of the set.

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
jacobtoppm: Thanks, will fix!

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
jacobtoppm: I couldn't find another way of getting the offset (and thus style) in the decorator in the current draft version - it looks like the next version provides the start in the props, but unfortunately we're not there yet! Once we update in Draftail, this is a hack we can remove (but if you can find an alternative to this hack in the short term, I'd love to hear it of course, that would be great)

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
jacobtoppm: This is pretty arbitrary, I just didn't want the style numbers to get too big if you stayed on the page for a while. To trigger the rerender, the added style can't be the same as one that was just filtered out, but there's no other constraints.

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -0,0 +1,217 @@
...
+
+    return <DraftailEditor
+    ref={editorRef}
+    onChange={setEditorState}
+    editorState={editorState}
+    controls={enabled ? [CommentControl] : []}
+    decorators={enabled ? [{
+      strategy: findCommentStyleRanges,
+      component: CommentDecorator 
+    }] : []}
```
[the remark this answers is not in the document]
jacobtoppm: So I should declare it as a variable in the function rather than inline here? Or would I need to memoise it to fix any equality checks by reference failing?

package.json
```diff
@@ -54,7 +54,7 @@
     "enzyme": "^3.9.0",
     "enzyme-adapter-react-16": "^1.9.1",
     "enzyme-to-json": "^3.3.0",
-    "eslint": "^7.11.0",
+    "eslint": "^7.20.0",
```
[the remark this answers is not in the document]
jacobtoppm: Thanks, will fix!

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
jacobtoppm: Probably a good idea! Was trying to keep with the rest of the Draftail stuff for now, but it's probably best just to switch

jacobtoppm:
Re tests: yep, they're needed, but given the changes we've previously needed when doing Draftail comments, it would be good to get a nod on the rest of the code first to okay the general approach

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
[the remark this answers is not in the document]
thibaudcolas: My bad, I had missed `CommentDecorator` was declared in this function. Ignore unless there seems to be performance issues! If so, yes, we’d want to memoise.

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
thibaudcolas: This is really just the usual "mean vs median" question in statistics. Looking at the particular consequences here, both suffer from the problem that you can have inline annotations quite far apart from one-another in a long field.

I’m not sure I fully understand the comment positioning logic so I’ll only state what I think we want:

- If a comment is attached to multiple inline annotations, I think it should either be displayed next to the first of those annotations, or next to the "median" annotation. Currently using the average means that comments can be placed alongside seemingly unrelated text.
- Comments should always be displayed in the viewport when focused. 

Here is an extreme case for reference:

comments-position

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -0,0 +1,217 @@
...
+}
+
+function getFullSelectionState(contentState) {
+  // Get a selection state corresponding to the full contentState
+  const lastBlock = contentState.getLastBlock();
+  let fullSelectionState = SelectionState.createEmpty();
+  fullSelectionState = fullSelectionState.set('anchorKey', contentState.getFirstBlock().getKey());
+  fullSelectionState = fullSelectionState.set('focusKey', lastBlock.getKey());
+  fullSelectionState = fullSelectionState.set('anchorOffset', 0);
+  fullSelectionState = fullSelectionState.set('focusOffset', lastBlock.getLength());
```
thibaudcolas, line 51: Nitpick:

```suggestion
  fullSelectionState = fullSelectionState.set('anchorKey', contentState.getFirstBlock().getKey());
  fullSelectionState = fullSelectionState.set('anchorOffset', 0);
  fullSelectionState = fullSelectionState.set('focusKey', lastBlock.getKey());
  fullSelectionState = fullSelectionState.set('focusOffset', lastBlock.getLength());
```

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -0,0 +1,217 @@
...
+      const fieldNode = this.field
+      if (fieldNode) {
+        return fieldNode.getBoundingClientRect().top + document.documentElement.scrollTop
+      }
+      return 0
+    }
+}
+
+function getFullSelectionState(contentState) {
+  // Get a selection state corresponding to the full contentState
```
thibaudcolas, line 45:
```suggestion
/**
 * Get a selection state corresponding to the full contentState.
 */
function getFullSelectionState(contentState) {
```

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -0,0 +1,217 @@
...
+}
+
+function getCommentControl(commentApp, contentPath, fieldNode) {
+  return ({getEditorState, onChange}) => {
+    return <ToolbarButton
+      name='comment'
+      active={false}
+      title={STRINGS.ADD_A_COMMENT}
+      icon={<Icon name="comment"/>}
+      onClick={() => {
```
thibaudcolas, line 62: I think this needs extra logic so clicking this button multiple times in succession doesn’t create multiple empty comments.

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -0,0 +1,217 @@
...
+    // TODO: determine the correct way to make this accessible, allowing both editing and focus jumps
+    return (
+      <a 
+        type="button"
+        ref={annotationNode}
+        onClick={onClick}
+        data-annotation
+      >
+        {children}
+      </a>
```
thibaudcolas, line 117:
```suggestion
      <span 
        role="button"
        ref={annotationNode}
        onClick={onClick}
        data-annotation
      >
        {children}
      </span>
```

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -0,0 +1,217 @@
...
+    const focusedId = useSelector(commentApp.selectors.selectFocused);
+
+    const ids = useMemo(() => comments.map(comment => comment.localId), [comments]);
+
+    const commentStyles = useMemo(() => ids.map(id => {return {
+      type: COMMENT_STYLE_IDENTIFIER + id,
+      style: enabled ? {
+        'background-color': (focusedId !== id) ? '#01afb0' : '#007d7e'
+      } : {}
+    }}), [ids, enabled]);
```
thibaudcolas, line 155: I find this style of inline `return` a bit hard to read.

```suggestion
    const commentStyles = useMemo(() => ids.map(id => ({
      type: `${COMMENT_STYLE_IDENTIFIER}${id}`,
      style: enabled ? {
        'background-color': (focusedId !== id) ? '#01afb0' : '#007d7e'
      } : {}
    })), [ids, enabled]);
```

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -0,0 +1,217 @@
...
+        filterInlineStyles(inlineStyles.map(style => style.type), editorState.getCurrentContent())
+      );
+      timeoutRef.current = window.setTimeout(
+        onSave(serialiseEditorStateToRaw(filteredEditorState)),
+        250,
+      );
+      return () => {
+        window.clearTimeout(timeoutRef.current);
+      }
+    }, [editorState, inlineStyles]);
```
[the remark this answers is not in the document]
jacobtoppm: I don't think we'd need it directly in Draftail, since Draftail is doing a sensible thing here: it's filtering to the inline styles it knows about, it's just we don't want it to save all of those legitimate styles because some are comments. This is probably also where we'd want to put the comment position updating logic, when we serialise it. So yep, will add a comment to that effect!

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
jacobtoppm: What would you suggest? A cooldown?

client/src/components/CommentApp/comments.js
```diff
@@ -42,13 +33,58 @@ function getContentPath(fieldNode) {
...
+    this.setOnClickHandler(localId);
   }
   onDelete() {
     this.node.remove();
+    if (this.unsubscribe) {
+      this.unsubscribe();
+    }
   }
   onFocus() {
     this.node.classList.remove('button-secondary');
```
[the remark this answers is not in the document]
jacobtoppm: Thanks for catching!

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
jacobtoppm: Plan is now to do this in a future ticket, so resolving this for now

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
jacobtoppm: Aiming to support multiple field level comments on non Draftail as well after discussing with Karl, so resolving this because creating multiple unhighlighted comments will be consistent

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
jacobtoppm: Removed, thanks!

client/src/components/CommentApp/comments.js
```diff
@@ -19,10 +11,9 @@ function initComments() {
       throw new Error('Comments app failed to initialise. Missing HTML element');
     }
     const data = JSON.parse(dataElement.textContent);
-    window.commentApp = initCommentsApp(
+    window.commentApp.renderApp(
       commentsElement, commentsOutputElement, data.user, data.comments, new Map(Object.entries(data.authors)), STRINGS
```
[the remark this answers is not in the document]
jacobtoppm: I think this is a good idea, but I'm going to add this to the list of things to do on separate PRs instead, as this one's become a bit unwieldy and the parameters weren't changed here.

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -20,21 +20,60 @@ function usePrevious(value) {
...
+    static getMedianRef(refArray) {
+      const refs = refArray.sort((a, b) => this.getHeightForRef(a) - this.getHeightForRef(b))
+      const length = refs.length
+      if (length > 0) {
+        return refs[Math.ceil(length/2 - 1)]
       }
+      return null
+    }
+    getDesiredPosition(focused = false) {
+      // The comment should always aim to float by an annotation, rather than between them, so uses the median height decorator ref
```
kaedroho, line 63: "median height decorator ref" doesn't make a lot of sense to me!

Does this sound accurate?

> The comment should always aim to float by an annotation, rather than between them, so calculate which annotation is the median one by height and float the comment by that

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -20,21 +20,60 @@ function usePrevious(value) {
...
     constructor(field) {
       this.field = field;
-      this.inlineRefs = new Set()
+      this.inlineRefs = new Map();
+      this.focusedBlockKey = '';
+      this.cachedMedianRef = null;
     }
-    addRef(ref) {
-      this.inlineRefs.add(ref);
+    addRef(ref, blockKey) {
```
kaedroho, line 36: The name ``ref`` isn't very clear since it could point to anything. Are they pointing to "decorators"? If so, could we rename this to ``decoratorRef``?

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
jacobtoppm: Addressed now

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
jacobtoppm: Changed, thanks!

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
jacobtoppm: Done!

client/src/components/Draftail/CommentableEditor/CommentableEditor.tsx
```diff
@@ -325,7 +359,7 @@ function CommentableEditor({
     setUniqueStyleId((id) => (id + 1) % 200);
   }, [focusedId, enabled, inlineStyles, ids, editorState]);
 
-  const timeoutRef = useRef();
+  const timeoutRef = useRef<ReturnType<typeof setTimeout> | undefined>();
```
kaedroho, line 362: For clarity, is there an actual type we can use here?

kaedroho:
Typescript conversion all looks good to me!

PR #6857, opened 2021-03-02
Change webmaster to admin

nmorduch:
Working to decrease potentially-insensitive language as per [RFC 63](https://github.com/wagtail/rfcs/blob/master/text/063-avoid-insensitive-language.md).

The primary change here is `contact your webmaster` → `contact your website administrator`. `webmaster` contains `master` and is also a title that is increasingly outdated; `website administrator` is still not a very precise term, but I think it should work fine for this. I'm not confident that `admin@localhost` is an amazing name choice, but I also don't think that matters enormously.

## Updating for translation

I did not change the translated versions of this phrase, so some of them still have variations on 'webmaster', but some of them already had some variation on 'administrator'. Retranslating this phrase may not be a high priority, given that only the wording has changed, not the meaning.

**After merging this PR, push changes to Transifex.**

## Checks

* [x] Do the tests still pass? Yes
* [x] Does the code comply with the style guide? No linting errors in things I changed
* [x] For Python changes: I updated tests
* ~~For front-end changes: Did you test on all of Wagtail’s supported browsers?~~

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -38,6 +38,7 @@ Changelog
  * Implemented a locale switcher on the forms listing page in the admin (Dan Braghis)
  * Implemented a locale switcher on the page chooser modal (Dan Braghis)
  * Implemented the `wagtail_site` template tag for Jinja2 (Vladimir Tananko)
+ * Change webmaster to website administrator in the admin (Naomi Morduch Toubman)
  * Fix: Accessibility fixes for Windows high contrast mode; Dashboard icons colour and contrast, help/error/warning blocks for fields and general content, side comment buttons within the page editor, dropdown buttons (Sakshi Uppoor, Shariq Jamil, LB (Ben Johnston), Jason Attwood)
  * Fix: Rename additional 'spin' CSS animations to avoid clashes with other libraries (Kevin Gutiérrez)
  * Fix: `default_app_config` deprecations for Django >= 3.2 (Tibor Leupold)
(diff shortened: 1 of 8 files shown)


ababic (commented):
Hey @nmorduch! This is great to see, thank you :) The email address change will need including in the upgrade considerations section of the release notes though, as it classes as a configurational change for existing sites (even if the likelihood of sites NOT setting `DEFAULT_FROM_EMAIL` is slim).

nmorduch:
It turns out that it would be very difficult to ever be in a situation where `DEFAULT_FROM_EMAIL` is not set, since you would have to remove that setting entirely, and it's one of Django's core settings (see [django/conf/global_settings.py L223](https://github.com/django/django/blob/stable/3.1.x/django/conf/global_settings.py#L223)).

In light of this, I don't think it's worth making a change that would require upgrade consideration. My inclination is instead to:
- leave our fallback the same as Django's default (with a comment in the code noting that it is Django's default)
- remove the mention of our fallback in the docs, because:
  - it would be extremely difficult to remove the setting, and if you managed to, we just use the same default as Django
  - it has a name I think it would be ideal to avoid

Does that sound sensible to you, @ababic?

ababic:
@nmorduch yes, those suggestions are great. Fingers crossed Django will make that change eventually anyway, and we can benefit without doing anything.

thibaudcolas:
@nmorduch wondering if you’ve considered proposing this change to Django itself? It looks like this was discussed relatively recently https://groups.google.com/g/django-developers/c/tctDuKUGosc/m/0OD-u5uPBQAJ, I can’t see any discussion for `webmaster` in particular.

thibaudcolas (approved):
Thank you @nmorduch, looks spot on. I have rebased, and removed the changes to the `.po` files – we’re pretty close to submitting recent changes for Wagtail 2.16 to Transifex, so I have assumed it would be better to only do that once.

PR #6861, opened 2021-03-03
Feature/advanced annotation

jacobtoppm:
This PR improves the logic around annotating rich text with inline comments:
- Handling styling for overlapping comments
- Adding logic to make sure decorators for overlapping comments try and make all comment highlights clickable at least in part (note this isn't perfect: perfectly overlapping comments for instance will still be problematic, but I'm wary of making the logic too heavy considering how often it runs)
- Ensuring entities aren't decorated by the comments system (but still remain editable), allowing eg commented links to be editable
- Preventing undo/redo from resurrecting invalid comment styles
It builds on https://github.com/wagtail/wagtail/pull/6856

### DIFF
--- a/client/scss/elements/_root.scss
+++ b/client/scss/elements/_root.scss
@@ -2,6 +2,8 @@
     @include define-color('color-primary', #007d7e);
     @include define-color('color-primary-darker', css-darken(css-adjust-hue(get-color('color-primary'), 1), 4%));
     @include define-color('color-primary-dark', css-darken(css-adjust-hue(get-color('color-primary'), 1), 7%));
+    @include define-color('color-primary-lighter', css-lighten(css-adjust-hue(get-color('color-primary'), 1), 10%));
+    @include define-color('color-primary-light', css-lighten(css-adjust-hue(get-color('color-primary'), 1), 15%));
 
     @include define-color('color-input-focus', css-lighten(css-desaturate(get-color('color-primary'), 40%), 72%));
     @include define-color('color-input-focus-border', css-lighten(css-saturate(get-color('color-primary'), 12%), 10%));
--- a/client/scss/settings/_variables.scss
+++ b/client/scss/settings/_variables.scss
@@ -28,6 +28,8 @@ $breakpoints: (
 );
 
 // colours
+$color-teal-light: var(--color-primary-light);
+$color-teal-lighter: var(--color-primary-lighter);
 $color-teal: var(--color-primary);
 $color-teal-darker: var(--color-primary-darker);
 $color-teal-dark: var(--color-primary-dark);
--- a/client/src/components/Draftail/CommentableEditor/CommentableEditor.tsx
+++ b/client/src/components/Draftail/CommentableEditor/CommentableEditor.tsx
@@ -162,15 +162,25 @@ function getCommentControl(commentApp: CommentApp, contentPath: string, fieldNod
 }
 
 function findCommentStyleRanges(contentBlock: ContentBlock, callback: (start: number, end: number) => void) {
+  // Find comment style ranges that do not overlap an existing entity
+  const entityRanges: Array<[number, number]> = [];
+  contentBlock.findEntityRanges(character => character.getEntity() !== null, (start, end) => entityRanges.push([start, end]));
   contentBlock.findStyleRanges(
-    (metadata) =>
-      metadata
-        .getStyle()
-        .some((style) => style !== undefined && style.startsWith(COMMENT_STYLE_IDENTIFIER)),
+    (metadata) => metadata.getStyle().some((style) => style !== undefined && style.startsWith(COMMENT_STYLE_IDENTIFIER)),
     (start, end) => {
-      callback(start, end);
-    }
-  );
+      const interferingEntityRanges = entityRanges.filter(value => value[1] > start).filter(value => value[0] < end);
+      let currentPosition = start;
+      interferingEntityRanges.forEach((value) => {
+        const [entityStart, entityEnd] = value;
+        if (entityStart > currentPosition) {
+          callback(currentPosition, entityStart);
+        }
+        currentPosition = entityEnd;
+      });
+      if (currentPosition < end) {
+        callback(start, end);
+      }
+    });
 }
 
 interface DecoratorProps {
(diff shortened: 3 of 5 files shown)


client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -201,14 +237,45 @@ function CommentableEditor({commentApp, fieldNode, contentPath, rawContentState,
...
+          // There is at least one comment in the range
+          const commentIds = commentStyles.map(style => parseInt(style.slice(8)));
+          let background = '#01afb0';
+          if (commentIds.has(focusedId)) {
+            // Use the focused colour if one of the comments is focused
+            background = '#007d7e';
+          } else if (numStyles > 1) {
+            // Otherwise if we're in a region with overlapping comments, use a slightly darker colour than usual
+            // to indicate that
+            background = '#01999a';
```
zerolab, line 270: suggestion: use classes so that the colours are consistent and in one place. Especially now when the admin can have colour themes ([docs](https://docs.wagtail.io/en/stable/advanced_topics/customisation/admin_templates.html#custom-user-interface-colors))

jacobtoppm: The issue with this is that as it's inside Draftail, we can't apply classes directly. (We can using a decorator, but due to the need to comment across entities, we can't guarantee a decorator will be used for the whole comment annotation)

That said, I do agree that these should be taken from somewhere, rather than hardcoded - as noted in the other PR, styling here is still placeholder, so this isn't something I've looked into in detail yet, how to retrieve these colours from the css,

jacobtoppm: possibly just creating and retrieving the styles from an element with the right class...

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -80,7 +97,29 @@ function getCommentDecorator(commentApp) {
...
+            findCommentStyleRanges(block, () => {counter = counter + 1}, (metadata) => metadata.getStyle().some(rangeStyle => rangeStyle === style));
+            return [style, counter]
+          }).sort((a, b) => a[1] - b[1])
+        
+          styleToUse = styleFreq.first()[0];
+
+        } else {
+          styleToUse = styles.first();
+        }
+        return parseInt(styleToUse.slice(8));
```
thibaudcolas, line 121: suggestion: I don’t think `8` as a magic value is understandable enough

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -201,14 +237,45 @@ function CommentableEditor({commentApp, fieldNode, contentPath, rawContentState,
 
     return <DraftailEditor
     ref={editorRef}
-    onChange={setEditorState}
+    onChange={(editorState) => {
+      let newEditorState = editorState;
+      if (['undo', 'redo'].includes(editorState.getLastChangeType())) {
```
thibaudcolas, line 242: So comment additions / removals can’t be undone or redone?

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
```diff
@@ -201,14 +237,45 @@ function CommentableEditor({commentApp, fieldNode, contentPath, rawContentState,
...
+        newEditorState = forceResetEditorState(editorState, filteredContent);
+      }
+      setEditorState(newEditorState);
+    }
+    }
     editorState={editorState}
     controls={enabled ? [CommentControl] : []}
     decorators={enabled ? [{
-      strategy: findCommentStyleRanges,
+      strategy: (block, callback) => findCommentStyleRanges(block, callback),
```
thibaudcolas, line 252: Is this really needed?

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
thibaudcolas: You should be able to replace these with CSS variables, we added them to Wagtail not so long ago.

thibaudcolas (commented):
Looking pretty good :)

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
jacobtoppm: Agreed, I'll change this!

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
jacobtoppm: You can undo adding the highlight, reverting the new comment to a field level comment. But the actual comment needs to be deleted in the comments system itself, which I don't think is a huge burden (we could look at doing this in specific cases - eg your comment is still in "creating" mode - but given that I don't think we can get this to work for deletions, I'm a bit cautious here)

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
jacobtoppm: This is because otherwise it gets a third argument (which isn't the `filterFn` argument) - `ContentState` I believe

client/src/components/Draftail/CommentableEditor/CommentableEditor.js
jacobtoppm: Okay great, this is something I'll look into then!

client/scss/elements/_root.scss
```diff
@@ -2,6 +2,8 @@
     @include define-color('color-primary', #007d7e);
     @include define-color('color-primary-darker', css-darken(css-adjust-hue(get-color('color-primary'), 1), 4%));
     @include define-color('color-primary-dark', css-darken(css-adjust-hue(get-color('color-primary'), 1), 7%));
+    @include define-color('color-primary-lighter', css-lighten(css-adjust-hue(get-color('color-primary'), 1), 10%));
```
kaedroho, line 5: Are these colours from the design system?

jacobtoppm: Not yet - we have an additional Draftail styling ticket for tweaking the colours/design, these are just "roughly similar" placeholders.

PR #6872, opened 2021-03-04
Prevent error on page editing after changing LANGUAGE_CODE

### LINKED ISSUE #6860: Edit page will get error after changing LANGUAGE_CODE in settings

### Issue Summary
When I set LANGUAGE_CODE = 'zh-hans', then edit page will ocurred error like

`can't currently handle on_delete types other than CASCADE, SET_NULL and DO_NOTHING`

### Steps to Reproduce
1.set LANGUAGE_CODE = 'zh-hans' in settings/base.py
2. runserver anyway
3. Try to edit page in nemu

### Technical details

* Python version: 3.9.1
* Django version: 3.1.7
* Wagtail version: 2.12.2.
* Browser version: Edge 88.

gasman:
Fixes #6860 as per https://github.com/wagtail/wagtail/issues/6860#issuecomment-790599722.

### DIFF
--- a/wagtail/admin/tests/pages/test_edit_page.py
+++ b/wagtail/admin/tests/pages/test_edit_page.py
@@ -1000,6 +1000,58 @@ class TestPageEdit(TestCase, WagtailTestUtils):
 
         self.assertEqual(response.status_code, 405)
 
+    def test_edit_after_change_language_code(self):
+        """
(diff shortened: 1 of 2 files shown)


kaedroho (approved):
This looks good to me, thanks for fixing it!

The reason I added this was to remove invalid locales from all the menus. But it's probably better to display the broken locales in the menus as this would make it easier to tell what's going on and the logic is simpler!

PR #6873, opened 2021-03-04
Implemented Translatable settings models

kaedroho:
This PR adds the ability to define translatable settings models. These models have one instance per site and locale rather than just one instance per site.

I've also changed the look of the site switcher to look the same as the locale switcher:

image

TODO:

 - [ ] Tests
 - [ ] Docs


zerolab:
missed the fact this is still a draft (hence the `needs review` label dance). if you rebase it, I will make time to review/test it this week

zerolab:
Closing in favour of #9390

Issue #6876, opened 2021-03-04
Large amount of queries in wagtail admin


dest81:
### Issue Summary

I have noticed that sometimes wagtail admin is very slow when DEBUG=True. Quick look into django-debug-toolbar shows hundreds of queries.

### Steps to Reproduce

1. Install wagtail with debug_toolbar
2. Add simple custom Page into models.py
```
    class MyPage(Page):
          pass
```
3.  Run server
4.  Open wagtail admin and create MyPage
5.  Save MyPage and open edit view again  - you will see around 60 SQL queries in the debug_toolbar
6.  Submit for moderators approvals and then open page in edit view - now you will see around 400 queries

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: Yes

### Technical details

* Python version:  3.8.2.
* Django version:  3.1.6.
* Wagtail version: 2.12.

Short investigation showed that function [current_workflow_state](https://github.com/wagtail/wagtail/blob/main/wagtail/core/models.py#L2834) is called few hundreds times. 
For example function [current_workflow_task_state](https://github.com/wagtail/wagtail/blob/main/wagtail/core/models.py#L2844) trigger function current_workflow_state 4 times per one call.

If I wrap current_workflow_task in cached_property decorator then number of queries goes down to 60.

Is it known issue? Shell I send a PR?

femto113:
I wonder if this is something that recently got worse?  I just noticed it as well and put in this PR #6878

KalobTaulien:
Hi @dest81 

This is a good note, thanks for bringing this up. 

So this is going to largely depend on the page you're viewing. If you're viewing the Images section (listing page) with 100 images per page, that's going to get very costly very quickly (especially when running renditions against each one for the first time). 

Likewise, on views that list lots of pages, or semi-custom views like using Wagtails ModelAdmin where you can define your own fields, this can get expensive as well. 

Middleware and context processors can play a role in adding queries as well. 

On a fresh install with ~100 pages and ~100 images and ~100 documents (any number is fine to test) are you still seeing 100's of database queries? Or is this only on an existing installation?

dest81:
Hi @KalobTaulien, 

This is on fresh installed instance. I added simple custom page model (without images) into models.py and create one object of it. Issue I am talking about can be seen only if you open EditPage view and page is submitted for moderators approvals.
So it's not about list view, that's why I found it very strange. @femto113 PR would work, I was going to send similar.
[...]

PR #6878, opened 2021-03-05
make Page.current_workflow_state a cached_property

femto113:
Trying to diagnose general slowness with the page editor I noticed many identical queries being issued in a row of the form

```
SELECT
   "wagtailcore_workflowstate"."id",
   ...
FROM
   "wagtailcore_workflowstate"
WHERE ((
      "wagtailcore_workflowstate"."status" = 'in_progress'
       OR "wagtailcore_workflowstate"."status" = 'needs_changes'
    ) AND "wagtailcore_workflowstate"."page_id" = 105
) LIMIT 21;
```

Which is generated by a query in `Page`'s `current_workflow_state` property.  This is compounded by the fact that the `current_workflow_task_state` property can reference it up to 4 times in a single call!  By making `current_workflow_state` a `@cached_property` the total number of queries is reduced dramatically.  There was also a very similar query used by the `workflow_in_progress` query, so that property has been rewritten to reuse the `current_workflow_state` rather than issuing its own query.

I don't see any obvious reason why this should break anything (assuming the page's workflowstate won't be changing in the middle of a request) and the core tests pass, but it's conceivable some explicit cache invalidation may need to be handled somewhere.

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -2786,9 +2786,9 @@ class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):
     @property
     def workflow_in_progress(self):
         """Returns True if a workflow is in progress on the current page, otherwise False"""
-        return WorkflowState.objects.filter(page=self, status=WorkflowState.STATUS_IN_PROGRESS).exists()
+        return self.current_workflow_state is not None and self.current_workflow_state.status == WorkflowState.STATUS_IN_PROGRESS
 
-    @property
+    @cached_property
     def current_workflow_state(self):
         """Returns the in progress or needs changes workflow state on this page, if it exists"""
         try:


gasman:
Thanks @femto113! Looks like there are a few failing tests in the wagtail.admin module - could you take a look please?

jacobtoppm (changes requested):
Thanks for contributing, I agree with optimising this! However, assuming that it won't change in the request cycle is incorrect: for example, performing a workflow action is something that you can do in a request, so at that point the cache does need to be invalidated, at the point you cancel, end, or start a workflow. I believe this is the cause of the test failures - once that's addressed, I would be happy to approve, since we did originally have this as a `cached_property` and it would be good to get that back!

femto113:
I've only barely scratched the surface of the wagtail code (only enough to figure out why this particular slowness is happening) and don't have time at the moment to dig into this deep enough to fix the remaining tests.  Perhaps someone more familiar with the area can take this as inspiration.  If directly caching `current_workflow_state` proves to be overly complex perhaps some of the higher level properties that reference it could be more easily cached for a whole request.  There was some discussion of what paths are invoking this in #6876

zerolab:
Closing in favour of #6946. Thank you for your work on this @femto113

PR #6880, opened 2021-03-05
Fixed pagination url for 'view users in a group'

### LINKED ISSUE #6874: When viewing users in a group, clicking next page brings you to list of all users
### Issue Summary

When clicking to view users within a group. The pagination brings you to the list of ALL users, not the users within a specific group.

### Steps to Reproduce

1. Go to edit a group within wagtail
2. Click "View users in this group" in the top right.
3. If you have enough users in this group to necessitate pagination, click "next"

You'll now be viewing *all* users, not just the users within that group.

### Technical details

* Python version: 3.7.7
* Django version: 3.1.6.
* Wagtail version: 2.12

I believe it's due to this line:
https://github.com/wagtail/wagtail/blob/main/wagtail/users/templates/wagtailusers/users/results.html#L17

Where it's passing in the link url as `wagtailusers_users:index` - I think this needs to be a different named url, however I could not find the named url that matches the pattern: `wagtail/groups/<group id here>/users/`

sagar-wal:
I too was able to reproduce this bug. 

Fixed it by changing pagination url of 'view users in a group' from 
`wagtailusers_users:index` to `wagtailusers_groups:index group`

Closes #6874 

@allcaps  Kindly review.

### DIFF
--- a/wagtail/users/templates/wagtailusers/users/results.html
+++ b/wagtail/users/templates/wagtailusers/users/results.html
@@ -14,7 +14,7 @@
 
     {% include "wagtailusers/users/list.html" %}
 
-    {% include "wagtailadmin/shared/pagination_nav.html" with items=users linkurl="wagtailusers_users:index" %}
+    {% include "wagtailadmin/shared/pagination_nav.html" with items=users linkurl="wagtailusers_groups:index group" %}
 {% else %}
     {% if is_searching %}
          <h2 role="alert">{% blocktrans %}Sorry, no users match "<em>{{ query_string }}</em>"{% endblocktrans %}</h2>


gasman:
Hi @sagar-wal,
The `wagtailusers/users/results.html` template is [also used for the main users listing](https://github.com/wagtail/wagtail/blob/main/wagtail/users/templates/wagtailusers/users/index.html#L28) - have you checked that this change doesn't break that listing?

sagar-wal:
Yes I've checked, the main users listing doesn't break.

gasman:
I was puzzled how changing the URL could work even though the include is used in two different views with different URLs, and it turns out the answer is in pagination_nav.html...

https://github.com/wagtail/wagtail/blob/571b9e1918861bcbe64a9005a4dc65f3a1fe7a15/wagtail/admin/templates/wagtailadmin/shared/pagination_nav.html#L4-L11

`linkurl="wagtailusers_groups:index group"` is invalid here - you can't pass the `group` parameter as part of the URL route string like that - so it's failing to generate the URL path, and outputting a URL like `<a href="?p=2">`, which works on both views. Omitting `linkurl` completely has the same effect.

Since omitting `linkurl` is the 'official' documented way of using pagination_nav.html on URLs with parameters (like `/admin/groups/<group_id>/users/`, I'll use that version - that way the code is a bit less misleading for people reading it in future. Thanks @sagar-wal!

gasman:
Merged in f9d1ec0bf70d9071a7daebbb8b7e49f9a0d04377

PR #6881, opened 2021-03-06
Fix typo in site settings documentation example

naglis:
Lowercase `i` is used in the model definition and in other examples.


thibaudcolas:
Thank you!

PR #6883, opened 2021-03-08
Fix the DocSearch search input appearing smaller after initialization

thibaudcolas:
Follow-up to #6833. This is pretty small so assuming it works I’ll merge right away.


thibaudcolas:
It works – tested in latest Chrome, Safari, Firefox on macOS.

PR #6889, opened 2021-03-08
Telepath - animation fixes

gasman:


gasman:
Broke a few tests here due to snapshot changes - will fix tomorrow.

kaedroho (approved):
Looks good to me! There's just a bit of inconsistency with the way we're deconstructing the ``opts`` argument, but that's quite minor.

PR #6891, opened 2021-03-09
Refactor get_settings template tag to return a value instead of writing directly to the context.

### LINKED ISSUE #4753: Make the name of the Site settings configurable in the templates' context.
### Issue Summary

As mentioned in the [Site settings docs](https://docs.wagtail.io/en/v2.2.1/reference/contrib/settings.html#using-the-settings), it's possible to use `{{ settings }}` (or `{{ settings(...) }}` for jinja2). However, in our case, we already use `settings` in our context to access the django settings. I know we can do all sorts of workaround to make it work in our case, but I think it would be nicer (and not too hard) to have a wagtail settings to set the name of the context variable holding the settings instead of hardcoding it to `settings` (it could default to `settings` too).

#### Django

```python
@register.simple_tag(takes_context=True)
def get_settings(context, use_default_site=False):
    if use_default_site:
        site = Site.objects.get(is_default_site=True)
    elif 'request' in context:
        site = context['request'].site
    else:
        raise RuntimeError('No request found in context, and use_default_site '
                           'flag not set')
    context_key = getattr(settings, 'WAGTAILCONTRIB_SITE_SETTINGS_CONTEXT_KEY', 'settings')
    context[context_key] = SettingsProxy(site)
    return ''
```

#### Jinja

```python
class SettingsExtension(Extension):
    def __init__(self, environment):
        super(SettingsExtension, self).__init__(environment)
        context_key = getattr(settings, 'WAGTAILCONTRIB_SITE_SETTINGS_CONTEXT_KEY', 'settings')
        self.environment.globals.update({
            context_key: get_setting,
        })
```

_**Note**: `WAGTAILCONTRIB_SITE_SETTINGS_CONTEXT_KEY` is a suggestion._

I'd be happy to do a fork/pull request, but is it possible to base it out of `v1.13.4` (We haven't migrated to python 3.0)?

solarissmoke:
Fixes #4753, and addresses #4753 for Django's template engine (not for Jinja2 though - I'm not sure it is possible there).

This PR changes `get_settings` to return a value (which can be assigned in the template) instead of writing directly to the context. This avoids the risk of collisions with other context variables ([example](https://github.com/pydata/conf_site/pull/320)). Thus the recommended usage of the tag changes to:

```
{% load wagtailsettings_tags %}

{% get_settings as wagtail_settings %}
{{ wagtail_settings.foo.SomeClass.text }}
```

In order to preserve some degree of backward compatibility, it's still writing to the context if no `settings` variable already exists - I've documented this as deprecated behaviour (but welcome input on whether this is the right approach).

Also refactored the tests to directly test the return value of the function instead of Django's template rendering.

### DIFF
--- a/wagtail/contrib/settings/templatetags/wagtailsettings_tags.py
+++ b/wagtail/contrib/settings/templatetags/wagtailsettings_tags.py
@@ -1,4 +1,5 @@
-from django.template import Library
+from django.template import Library, Node
+from django.template.defaulttags import token_kwargs
 
 from wagtail.core.models import Site
 
@@ -8,14 +9,33 @@ from ..context_processors import SettingsProxy
 register = Library()
 
 
-@register.simple_tag(takes_context=True)
-def get_settings(context, use_default_site=False):
-    if use_default_site:
-        site = Site.objects.get(is_default_site=True)
-        context['settings'] = SettingsProxy(site)
-    elif 'request' in context:
-        context['settings'] = SettingsProxy(context['request'])
-    else:
-        raise RuntimeError('No request found in context, and use_default_site '
-                           'flag not set')
-    return ''
+class GetSettingsNode(Node):
+    def __init__(self, kwargs, target_var):
+        self.kwargs = kwargs
+        self.target_var = target_var
+
+    @staticmethod
+    def get_settings_object(context, use_default_site=False):
+        if use_default_site:
+            site = Site.objects.get(is_default_site=True)
+            return SettingsProxy(site)
+        if 'request' in context:
+            return SettingsProxy(context['request'])
+
+        raise RuntimeError('No request found in context, and use_default_site flag not set')
+
+    def render(self, context):
+        resolved_kwargs = {k: v.resolve(context) for k, v in self.kwargs.items()}
+        context[self.target_var] = self.get_settings_object(context, **resolved_kwargs)
+        return ''
+
+
+@register.tag
+def get_settings(parser, token):
+    bits = token.split_contents()[1:]
+    target_var = 'settings'
+    if len(bits) >= 2 and bits[-2] == 'as':
+        target_var = bits[-1]
+        bits = bits[:-2]
+    kwargs = token_kwargs(bits, parser) if bits else {}
+    return GetSettingsNode(kwargs, target_var)
--- a/docs/reference/contrib/settings.rst
+++ b/docs/reference/contrib/settings.rst
@@ -170,6 +170,15 @@ If there is no ``request`` available in the template at all, you can use the set
 
 .. note:: You can not reliably get the correct settings instance for the current site from this template tag if the request object is not available. This is only relevant for multisite instances of Wagtail.
 
+By default, the tag will create or update a ``settings`` variable in the context. If you want to
+assign to a different context variable instead, use ``{% get_settings as other_variable_name %}``:
+
+.. code-block:: html+django
+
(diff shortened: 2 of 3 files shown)


wagtail/contrib/settings/templatetags/wagtailsettings_tags.py
```diff
@@ -12,10 +12,14 @@
...
-        context['settings'] = SettingsProxy(context['request'])
+        settings = SettingsProxy(context['request'])
     else:
         raise RuntimeError('No request found in context, and use_default_site '
                            'flag not set')
-    return ''
+    # For backward compatibility with the old behaviour of this tag
+    # set context['settings'] if it isn't already set
+    if 'settings' not in context:
+        context['settings'] = settings
```
RealOrangeOne, line 24: Python has a built-in mechanism for inserting into a dict only when the key doesn't exist

```suggestion
    context.setdefault('settings', settings)
```

solarissmoke: Thanks!

wagtail/contrib/settings/templatetags/wagtailsettings_tags.py
```diff
@@ -12,10 +12,13 @@
...
-        context['settings'] = SettingsProxy(context['request'])
+        settings = SettingsProxy(context['request'])
     else:
         raise RuntimeError('No request found in context, and use_default_site '
                            'flag not set')
-    return ''
+    # For backward compatibility with the old behaviour of this tag
+    # set context['settings'] if it isn't already set
+    context.setdefault('settings', settings)
+    return settings
```
gasman, line 24: Won't this cause the settings object to be rendered on the page, in the case that the old `{% get_settings %}` variant is used?

gasman (changes requested):
Thanks @solarissmoke!
Rather than changing the behaviour of the old `{% get_settings %}` version, could we make it so that we detect whether we've been called with or without `as foo`, and treat `{% get_settings %}` as equivalent to `{% get_settings as settings %}`? That way, if there's someone who really does rely on the tag overwriting an existing `settings` variable, we won't be breaking their code - and then the new preferred `{% get_settings as some_var %}` version doesn't have to touch `settings` at all.

Annoyingly I don't think we can do that with `simple_tag` (because [we don't have access to SimpleNode.target_var within our tag function](https://github.com/django/django/blob/ff514309e178e3955012050ead9b8fc66dc21a5b/django/template/library.py#L193)) so we probably have to use the lower-level tag parsing API, but that'll mostly be a copy-and-paste of the `simple_tag` implementation.

wagtail/contrib/settings/templatetags/wagtailsettings_tags.py
solarissmoke: Hah, yes, of course it would! I will rework this to use the approach you've suggested. Do you want to retain the current tag behaviour (writing to the context) indefinitely or should we add a deprecation warning if it is used without an assignment to a variable?

gasman: I think we should continue to allow the current syntax without a deprecation warning - if someone's happily using the `settings` variable without clashes then there's no real harm in letting them continue doing that.

gasman (approved):
All looks good to me, thanks!

gasman:
Merged in 0daae4a2b829445ee97906dd4393398778881fcb

PR #6893, opened 2021-03-09
Fix CI builds against development Django version

solarissmoke:
Django has [renamed](https://github.com/django/django/commit/d9a266d657f66b8c4fa068408002a4e3709ee669) its master branch to main, causing the CI build that was fetching from the old branch to fail.


gasman:
Thanks @solarissmoke!

PR #6902, opened 2021-03-10
Documentation Search: Adding pagination for results page

### LINKED ISSUE #6886: Documentation search: pagination for results page
Follow-up to #6833 – it would be nice to add pagination to our search results page, which is currently limited to the top 50 results: https://docs.wagtail.io/en/latest/search.html?q=pages.

The results page is mostly custom, and currently "50" is hard-coded. We’d need to implement the pagination logic, and UI: https://github.com/wagtail/wagtail/blob/23f146c7e115fc1daa342b6084ee94531cbe44dd/docs/_templates/search.html#L34.

Tijani-Dia:
Hello I tried to implement the pagination logic and UI for the docs search results page.
This is my first contribution ever so any help or suggestion will be appreciated.
I tested it against Mozilla and Chrome.

Fixes #6886

### DIFF
--- a/docs/_static/css/pagination.css
+++ b/docs/_static/css/pagination.css
@@ -1,0 +1,27 @@
+#pagination {
+    display: flex;
+    justify-content: center;
+}
+
+.pagination-previous {
+    margin-right: 2rem;
+}
+
+.pagination-next {
+    margin-left: 2rem;
+}
+
+.pagination-list {
+    display: flex;
+}
+
+.pagination-list li {
+    margin: auto 0.5rem;
+}
+
+.is-current {
+    border: 1px;
+    padding: 0.25rem;
+    background-color: #3091d1;
+    color: #d0d6da;
+}
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -6,6 +6,7 @@
 {% block extrahead %}
 <link rel="stylesheet" href="{{ docsearch_base ~ 'docsearch.min.css' }}"/>
 <link rel="stylesheet" href="{{ pathto('_static/css/docsearch.overrides.css', 1) }}" />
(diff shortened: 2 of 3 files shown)


thibaudcolas (changes requested):
@Tijani-Dia I’m having issues adding review comments to this, could you re-create a branch with those changes? Generally it’s a bad idea to delete your branch if there is a PR open for it.

I think this is working pretty well but the UI needs further work, and there are a few tidbits to change.

General comments:

- Could you add screenshots to your PR?
- Could you share search terms you’ve used this with that return different amounts of results

Comments file-by-file:

- `pagination.css` – missing newline at the end of the file
- `search.html` line 39 – I think we should retain our `hitsPerPage` of 50, I don’t see why we’d want to have so few results per page as 10
- `search.html` line 52/55 – we should use the `hidden` property rather than CSS for this. It should also be set by default (like 17) so we don’t display the pagination if JS fails to initialise.
- `search.html` lines 71 – 78 – this should be refactored to set the `disabled` values based on the expressions in the `if` conditions.
- `search.html` line 86 – we should either use a `button` element, or set a `href` on those links (and make sure we load the correct paginated page when navigating by URL)

On the UI,

pagination-ui

- We should style those prev/next buttons a bit more so they look more consistent with the rest of the page. Ideally reuse the same button styles as elswhere
- The page number styles are good, but I don’t think we’d want list items displayed there
- The whole component needs spacing below, and I think all buttons/links should be centered both horizontally and vertically.

thibaudcolas:
Just gave this another go in another browser and I still can’t add review comments to this PR due to the branch having been deleted. I don’t think this is salvageable without opening a new PR so I’ll close this now and make a comment on #6886 as to the state of this PR.

PR #6908, opened 2021-03-11
Fix reverse migration errors in images and documents

### LINKED ISSUE #6907: Reverse migration for documents 0011 and images 0023 fail

### Issue Summary

The reverse migration for documents.migrations.0011 and images.migrations.0023 both fail for the same reason.

These migrations were introduced by #6466 

Here are the stack traces for each:

```
  File "/lib/python3.6/site-packages/wagtail/documents/migrations/0011_add_choose_permissions.py", line 80, in remove_choose_permission_from_collections
    GroupCollectionPermission.objects.filter(permission=choose_permission).delete()
    ...
    'The QuerySet value for an exact lookup must be limited to '
ValueError: The QuerySet value for an exact lookup must be limited to one result using slicing.
```

```
  File "/lib/python3.6/site-packages/wagtail/images/migrations/0023_add_choose_permissions.py", line 80, in remove_choose_permission_from_collections
    GroupCollectionPermission.objects.filter(permission=choose_permission).delete()
    ...
    'The QuerySet value for an exact lookup must be limited to '
ValueError: The QuerySet value for an exact lookup must be limited to one result using slicing.
```

### Steps to Reproduce

1. Install wagtail 2.12.3
2. Migrate forward
3. Run `manage.py migrate documents 0010` or `manage.py migrate images 0022`

### Technical details

* Python version: 3.6.8
* Django version: 2.2.17
* Wagtail version: 2.12.3
* Browser version: N/A

m3brown:
This PR resolves #6907 

The root of the issue is that get_choose_permission() returns a Queryset.

### DIFF
--- a/wagtail/documents/migrations/0011_add_choose_permissions.py
+++ b/wagtail/documents/migrations/0011_add_choose_permissions.py
@@ -77,7 +77,7 @@ def remove_choose_permission_from_collections(apps, _schema_editor):
     GroupCollectionPermission = apps.get_model('wagtailcore.GroupCollectionPermission')
     choose_permission = get_choose_permission(apps)
 
-    GroupCollectionPermission.objects.filter(permission=choose_permission).delete()
+    GroupCollectionPermission.objects.filter(permission__in=choose_permission).delete()
 
 
 class Migration(migrations.Migration):
--- a/wagtail/documents/migrations/0011_add_choose_permissions.py
+++ b/wagtail/documents/migrations/0011_add_choose_permissions.py
@@ -54,7 +54,7 @@ def get_choose_permission(apps):
     return Permission.objects.filter(
         content_type=document_content_type,
         codename__in=['choose_document']
-    )
+    ).first()
 
 
 def copy_choose_permission_to_collections(apps, _schema_editor):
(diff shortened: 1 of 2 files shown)


ababic:
Hi @m3brown. Thank you for this! I was just looking at the code here and wondered whether the implementor really intended for `get_choose_permission()` to return a queryset, or whether they may have just forgot to replace `filter()` in that method with `get()` to return a single object. What do you think?

m3brown:
> Hi @m3brown. Thank you for this! I was just looking at the code here and wondered whether the implementor really intended for `get_choose_permission()` to return a queryset, or whether they may have just forgot to replace `filter()` in that method with `get()` to return a single object. What do you think?

I assumed the query set was intentional because of the for loop on line 67.

A logical explanation is that the function was initially a singular implementation, and the implementer updated the function to return a queryset but forgot to update one of the two scenarios that called it.

ababic:
Yes, perhaps that is the case. Either way, I think most sensible fix here is actually to update `get_choose_permission()` to use `first()`, making it return the correct `Permission` object (if it exists), or `None` (if it doesn't). Then, `copy_choose_permission_to_collections()` should be updated accordingly.

m3brown:
@ababic I updated the PR to reflect your suggestion. Note the new implementation has the risk of regressions to the forward migration if the queryset were ever to return more than one result. I trust your insight is better than mine on this matter.

ababic:
That's brilliant. Thanks very much @m3brown! I think we should be okay, since Permissions have a [unique_together restraint](https://github.com/django/django/blob/main/django/contrib/auth/models.py#L71) on `content_type` and `codename`:

ababic (approved):

PR #6909, opened 2021-03-12
Sanitize return_url in authenticate_with_password functions

sdreher:
Addressing unvalidated url redirects in the `authenticate_with_password` functions in the `core` and `documents` modules. I'm following Django's pattern in [LoginView](https://github.com/django/django/blob/stable/2.2.x/django/contrib/auth/views.py#L65) by first checking to see if the `return_url` [is safe](https://github.com/django/django/blob/stable/2.2.x/django/contrib/auth/views.py#L73), and if not, redirecting to the `LOGIN_REDIRECT_URL`.

Note:  [is_safe_url](https://github.com/django/django/blob/stable/3.0.x/django/utils/http.py#L330) is deprecated as of Django 3.0 and will be removed in Django 4.0. `is_safe_url` should be updated to [url_has_allowed_host_and_scheme](https://github.com/django/django/blob/stable/3.0.x/django/utils/http.py#L301) when Django 2.2 support is dropped.

### Checks
 
√ Do the tests still pass? (https://docs.wagtail.io/en/latest/contributing/developing.html#testing)
√ Does the code comply with the style guide? (Run `make lint` from the Wagtail root)
√ For Python changes: Have you added tests to cover the new/fixed behaviour?
~~For front-end changes: Did you test on all of Wagtail’s supported browsers? **Please list the exact versions you tested**.~~
~~For new features: Has the documentation been updated accordingly?~~

### DIFF
--- a/wagtail/core/views.py
+++ b/wagtail/core/views.py
@@ -1,6 +1,8 @@
+from django.conf import settings
 from django.http import Http404, HttpResponse
 from django.shortcuts import get_object_or_404, redirect
 from django.urls import reverse
+from django.utils.http import is_safe_url
 
 from wagtail.core import hooks
 from wagtail.core.forms import PasswordViewRestrictionForm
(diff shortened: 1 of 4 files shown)


wagtail/core/views.py
```diff
@@ -1,6 +1,8 @@
+from django.conf import settings
 from django.http import Http404, HttpResponse
 from django.shortcuts import get_object_or_404, redirect
 from django.urls import reverse
+from django.utils.http import is_safe_url
```
jaap3, line 5: Please note: `django.utils.http.is_safe_url()` is deprecated in favor of `url_has_allowed_host_and_scheme()`

jaap3 (changes requested):

wagtail/core/views.py
sdreher: 👍 I'll make the change.

gasman:
Thanks @sdreher! Merged in 09431f7b22b51e29d91fdbbe3e64f046103fafb0 (with @jaap3's suggested fix for the deprecated imports).

PR #6911, opened 2021-03-12
Ignore venv from stylelint

RealOrangeOne:
`coverage` ships with a `style.scss`, which `stylelint` attempts to lint.

https://github.com/nedbat/coveragepy/tree/master/coverage/htmlfiles


kaedroho:
Thanks!

PR #6921, opened 2021-03-16
Add option to set name to MenuItems generated by modeladmin

### LINKED ISSUE #6859: Customizable name for the MenuItem in modeladmin
There is currently no way to explicitly set the name of the `MenuItem` generated by `modeladmin`. This can be an issue if we want to, say, register two ModelAdmin's with the same label (for example for superuser and one for others, with different permissions). Then the name of the `MenuItem` is autogenerated from the label which leads to conflicts. An easy solution would be to add a `menu_name` to `ModelAdmin` which then gets passed to the `MenuItem` constructor in `ModelAdminMenuItem`. I could make a PR for this.

ARogovskyy:
This PR fixes #6859 by adding a `menu_name` attribute to `ModelAdmin` (and to `ModelAdminGroup`)

### DIFF
--- a/docs/reference/contrib/modeladmin/menu_item.rst
+++ b/docs/reference/contrib/modeladmin/menu_item.rst
@@ -73,3 +73,13 @@ This will only work for individual ``ModelAdmin`` classes registered with their
 own ``modeladmin_register`` call. It won't work for members of a
 ``ModelAdminGroup``.
 
+.. _modeladmin_menu_name:
+
+-----------------------------------
+``ModelAdmin.menu_name``
+-----------------------------------
+
+**Expected value**: A string or ``None``
+
+This will be passed to the ``name`` attribute of the corresponding ``MenuItem``.
+
--- a/wagtail/contrib/modeladmin/menus.py
+++ b/wagtail/contrib/modeladmin/menus.py
@@ -18,6 +18,7 @@ class ModelAdminMenuItem(MenuItem):
             icon_name = menu_icon
         super().__init__(
             label=model_admin.get_menu_label(), url=url,
+            name=model_admin.get_menu_name(),
             classnames=classnames, icon_name=icon_name, order=order)
 
     def is_shown(self, request):
@@ -40,6 +41,7 @@ class GroupMenuItem(SubmenuMenuItem):
             icon_name = menu_icon
         super().__init__(
             label=modeladmingroup.get_menu_label(), menu=menu,
+            name=modeladmingroup.get_menu_name(),
             classnames=classnames, icon_name=icon_name, order=order, )
 
     def is_shown(self, request):
--- a/wagtail/contrib/modeladmin/options.py
+++ b/wagtail/contrib/modeladmin/options.py
@@ -70,6 +70,7 @@ class ModelAdmin(WagtailRegisterable):
 
     model = None
     menu_label = None
+    menu_name = None
     menu_icon = None
     menu_order = None
     list_display = ('__str__',)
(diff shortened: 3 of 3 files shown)


ababic:
Hi @abogushov. Thank you for contributing to Wagtail! I'm curious to understand more about the use case that drove this requirement. I'm currently struggling to think of a situation where controlling the name attribute separately would be useful.

ARogovskyy:
Hi! This is more of a general thing that could come in handy rather than a specific use case. In my opinion, it makes sense to expose the full API of `MenuItem` instead of hiding it, especially since only few non-intrusive changes are required and `modeladmin` is explicitly focused on customizeability.
That being said, one use case that I can think of is hiding a modeladmin MenuItem in the `construct_main_menu` hook. Perhaps the modeladmin comes from an external app that you have little influence on, and thus you cannot change the `PermissionHelper`. Inside the hook identifying the `MenuItem` by its slugified label is problematic because the label might be localized, so a unique name might be handy.

ababic:
Okay, thanks for explaining @ARogovskyy. I'm happy with this in principle, but I'm worried about the potential confusion of having these two very similarly named things doing very different things, and the value requirements being quite different. For `name` to be used as an HTML attribute value, there are certain restrictions that apply to it, that aren't really being enforced or hinted at here.  How do you feel about renaming the attribute to something more descriptive, like `menu_item_html_name`, and maybe also using Django's checks framework to check that the value is suitable?

ARogovskyy:
I am not quite sure what you mean with HTML attribute. I am only talking about the `name` parameter of `MenuItem`, as defined in `wagtail/admin/menu.py`:
```python
class MenuItem(metaclass=MediaDefiningClass):
    template = 'wagtailadmin/shared/menu_item.html'

    def __init__(self, label, url, name=None, classnames='', icon_name='', attrs=None, order=1000):
        ...
```

My idea would be to have the `menu_name` passed on as this parameter.

ababic:
> I am not quite sure what you mean with HTML attribute. I am only talking about the name parameter of MenuItem

Ah, okay. I was under the impression that was making its way to the template as an attribute, but I see that isn't the case. How do you feel about `menu_item_name` as the name? 

Also, I think we could afford to be a little more explicit about what happens to this value in various places. For example: 

> Passed on as the 'name' parameter when initialising the ``MenuItem`` for this class, becoming the `name` attribute value for that instance.

lb-:
@ARogovskyy I know it has been a little while but I wanted to check in on this and see if you are ok to make the changes suggested by @ababic  above?

Also, would you be able to add a unit test to cover this usage so that it does not accidentally break in the future?

vupham04:
Hi, because nobody has updated this for a long time, I will create a new request for the issue with updated code, unless @ARogovskyy still wants to work on it

vupham04:
Hi, I'm trying to write unit tests for this code - but I don't know which file I should start. I wonder if anyone can assist me where I should start?

Also, I notice that there is no testing for menu_label or menu_icon, so is it necessary to add unit tests for menu_item_name?

ababic:
Hi @vupham04,

Due to the age of these changes I would highly recommend rebasing the branch onto `main` and trying to address the merge conflicts first.

As for where to add tests, I would probably add a new class to `wagtail.contrib.modeladmin.tests.test_simple_modeladmin` for this. 

Ultimately, the test just needs to show the attribute being set on a ModelAdmin class/instance, and that affecting to return value of `get_menu_item()`. So, you'll likely want to write two tests:

1. Showing the default behaviour (no value specified).
2. Showing the customisation affecting the result.

gasman:
Completed via #8553

PR #6922, opened 2021-03-16
Simplify oembed dict missing key fallback

jaap3:
Minor cleanup of fallback to empty string using the default fallback of `dict.get`.

This could also be written as `oembed.get('title') or ''` if `None` values are prohibited. However that would be a change is behaviour. They way it is written now, `None` values can pass through if the key is present in the `oembed` dict and the value is `None`.

### DIFF
--- a/wagtail/embeds/finders/oembed.py
+++ b/wagtail/embeds/finders/oembed.py
@@ -72,9 +72,9 @@ class OEmbedFinder(EmbedFinder):
 
         # Return embed as a dict
         return {
-            'title': oembed['title'] if 'title' in oembed else '',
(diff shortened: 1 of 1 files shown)


RealOrangeOne:
(Test issues are fixed by updating from `main`, 421bcbf9c33b0f9e5a0169c6fff6b30d8e85cce9)

RealOrangeOne (approved):

jaap3:
I didn't know wagtail also changed to `main`. I basically came across this bit of code by Google'ing for the Wagtail oembed provider. which pointed me to `master`.

jaap3:
I've properly rebased now, so the CI should pass 🤞

gasman:
Thanks @jaap3!

PR #6928, opened 2021-03-17
Make streamfields the same size as everything else

kaedroho:
To provide room for comments to be added on the right

image

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/scss/layouts/page-editor.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/layouts/page-editor.scss
@@ -271,8 +271,6 @@
     // cursory styling for streamfield. Main styling in client/src/components/StreamField/StreamField.scss
     &.stream-field {
         background-color: #f6f6f6;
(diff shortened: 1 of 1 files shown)


zerolab:
👍 for consistency. I wonder how does this look like for nested blocks?

kaedroho:
Thanks @zerolab, here's a screenshot with three levels of nesting:

image

zerolab:
Looks pretty good. The one consideration is the grey background -- there's a lot more of it now especially on wide screen. Do we need input from design?

Screenshot 2021-03-17 at 17 15 19

kaedroho:
Thanks, @zerolab!

Discussed this on a call with Ben:

The stream field should take up as much available space as possible, but when we implement comments, it should provide room for those to appear on the right.

We will remove the grey background, and instead use borders/shadows to make it easier to distinguish the blocks from the background. This styling will match what he's designed for the comments.

kaedroho:
Given the change of approach, I'll close this PR and submit a new one

PR #6930, opened 2021-03-17
Add Sphinx Wagtail Theme

allcaps:
This PR adds [Sphinx Wagtail Theme](https://github.com/wagtail/sphinx_wagtail_theme).

Preview: https://wagtail--6930.org.readthedocs.build/en/6930/index.html

- The theme is pinned in setup.py AND docs/requirements.txt (RTD doesn't read the setup.py)
- The title reads 'Wagtail Documentation' instead of 'Wagtail'
- Copyright is 'YYYY, Torchbox and contributors' instead of 'YYYY, Torchbox' (like the licence)
- Searchbox template is updated to reflect the theme (Bootstrap) markup
- Sphinx disables the TOC on pages that have a single heading. We added `:hidetoc: 1` to disable the TOC on pages with multiple headings, but where a TOC is still redundant.

Screenshot 2021-04-18 at 22 07 06

Most work is done on the [theme](https://github.com/wagtail/sphinx_wagtail_theme) and according to the [design prototype](https://www.figma.com/proto/hFjWqQeEjGd8NtTsEHcc4i/Wagtail-Documentation-(wagtail.io)?node-id=1%3A2&viewport=983%2C953%2C1&scaling=min-zoom&hide-ui=1)

The new theme:
- is based on Bootstrap
- is responsive
- takes 100% of the viewport
- has a table of contents
- does not reload the page when open/close main menu items

I expect the commits to be squashed on merge.

- Edge 89
- IE 11 ([search doesn't work](https://discourse.algolia.com/t/algolia-instant-search-v4-5-0-not-working-in-ie11/10492)) Note [the latest docs](https://docs.wagtail.io/en/latest/search.html?q=foo#) also doesn't work on IE11.
- Chrome  90.0.4430.72 OSX
- FireFox 86 OSX
- FireFox 87.0 OSX
- IPad Pro 12.9

### DIFF
--- a/docs/_static/css/docsearch.overrides.css
+++ b/docs/_static/css/docsearch.overrides.css
@@ -7,3 +7,11 @@
 .algolia-autocomplete {
     width: 100%;
 }
+
+#search-results a,
+a.algolia-docsearch-suggestion {
+    border-bottom: 0;
+}
+#search-results  h2 {
+    display: none;
+}
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -8,8 +8,7 @@
 <link rel="stylesheet" href="{{ pathto('_static/css/docsearch.overrides.css', 1) }}" />
 {% endblock %}
 
-{% block footer %}
-{{ super() }}
+{% block search %}
 <script async defer data-domain="docs.wagtail.io" src="https://plausible.io/js/plausible.js"></script>
 <script src="{{ docsearch_base }}docsearch.min.js"></script>
 <script>
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -58,7 +58,7 @@
     const search = docsearch({
       apiKey: '8325c57d16798633e29d211c26c7b6f9',
       indexName: 'wagtail',
-      inputSelector: '#rtd-search-form [name="q"]',
+      inputSelector: '#searchbox [name="q"]',
       algoliaOptions: {
         facetFilters: [getVersionFacetFilter()],
       },
--- a/docs/_templates/searchbox.html
+++ b/docs/_templates/searchbox.html
@@ -1,0 +1,4 @@
+<form id="rtd-search-form" class="wy-form" action="{{ pathto('search') }}" method="get" role="search">
+  <input type="text" name="q" placeholder="{{ _('Search docs') }}" />
+  <input type="submit" style="visibility:hidden;position:absolute">
+</form>
--- a/docs/_templates/searchbox.html
+++ b/docs/_templates/searchbox.html
@@ -1,4 +1,8 @@
-<form id="rtd-search-form" class="wy-form" action="{{ pathto('search') }}" method="get" role="search">
-  <input type="text" name="q" placeholder="{{ _('Search docs') }}" />
-  <input type="submit" style="visibility:hidden;position:absolute">
-</form>
+{%- if builder != "singlehtml" %}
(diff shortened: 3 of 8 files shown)


Stormheg:
https://wagtail--6930.org.readthedocs.build/en/6930/index.html

allcaps:
[ ] Fix override search
[ ] Drop banner.js

tbrlpld:
The version picker at the bottom appears broken. Maybe some of the removed theme css was injected into it?

tbrlpld:
- [ ] Update project `name` in `conf.py`

tbrlpld (approved):
LGTM 👍

Stormheg (approved):
One small nitpick, but apart from that this looks very complete to me! :smile: 

Nitpick: I think we should make it more clear that the version in the footer is the version of sphinx_wagtail_theme. Not the version of Wagtail itself.
image

tbrlpld:
> One small nitpick, but apart from that this looks very complete to me! 😄
> 
> Nitpick: I think we should make it more clear that the version in the footer is the version of sphinx_wagtail_theme. Not the version of Wagtail itself.

Hm, I feed like a reference to the theme in the docs would be odd. I think we are better off just removing that and only showing the project name or nothing in the footer.

Stormheg:
@tbrlpld I think a reference to the version of the documentation is useful for us (wagtail developers) in case we need to publish a new version of the theme.

allcaps:
I changed the line `Wagtail Documentation 5.0.0` into `Sphinx Wagtail Theme 5.0.1` and set display none.

Updated this pr.

thibaudcolas (commented):
Things that stand out to me and I think are mandatory to fix:

- The URL to the "Edit on GitHub" link is wrong.

Things I’d prefer to see fixed:

- In the search, I have additional (browser?) suggestions displaying underneath the Algolia dropdown. If I navigate search suggestions with keyboard keys, that will trigger the native search form instead of going to the Algolia result I selected.
- Also in the search, if pressing "Enter" on the search field to get the full list of results in the separate template, there are specific queries that break the page. For example "testintegerblock". Potentially related, I can see this still queries RTD’s search backend as well as Algolia.
- There is a hidden link without text in the header 

Nice to have:

- I don’t think we should have a link / version number for the docs theme in the footer.
- On mobile viewports, placement of the "hamburger" menu toggle button could be better
- Sidebar’s Toctree "Expand" buttons are impossible to use with a keyboard

thibaudcolas:
Re-reviewing this I can confirm there is something funny going on with the search. It looks like our custom Algolia search isn’t completely overriding the built-in one, which causes both search issues I reported above. I think this should be fixed before we merge. Other than that, the "Edit on GitHub" link should be easy to change but I can’t figure out where to set it (cc @allcaps).

Theme default we want to override:

```
github_url = https://github.com/wagtail/sphinx_wagtail_theme/blob/main/docs/
```

allcaps (commented):
LGTM, 

we should squash all commits on merge.

thibaudcolas:
@allcaps with the latest version there are two issues:

- The "expand" buttons are gone from the sidebar
- The version picker has also disappeared

allcaps:
> * The "expand" buttons are gone from the sidebar
> * The version picker has also disappeared

@thibaudcolas Speed over quality. Sorry! Updated the theme and fixed the issue.

Stormheg:
Looks like all feedback has been implemented. I'm going to squash and merge this now.

PR #6936, opened 2021-03-19
Show current page in breadcrumb

KalobTaulien:
Enabled the current page to show up in the breadcrumb by default.

Example image:
image

* For Python changes: N/A
* For front-end changes: Chrome, FF, Safari
* For new features: N/A

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/pages/create.html
+++ b/wagtail/admin/templates/wagtailadmin/pages/create.html
@@ -8,7 +8,7 @@
 {% block content %}
 
     <header class="merged tab-merged">
-        {% explorer_breadcrumb parent_page include_self=1 %}
+        {% explorer_breadcrumb parent_page include_self=1 trailing_arrow=True %}
 
         <div class="row">
             <div class="left col9 header-title">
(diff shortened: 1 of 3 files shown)


zerolab:
Thank you @KalobTaulien, what is the rationale behind this change?

KalobTaulien:
Hey @zerolab 

We want to be able to get back to the page we're on — when you're editing a page there's no visual way to get to the parent view (to see child pages) without going to the parent page, then into the page your editing, to see childpages.

zerolab:
Ah, that brought back memories of frustrations. Makes perfect sense.

We have 3 templates that use `explorer_breadcrumb` - https://github.com/wagtail/wagtail/search?q=explorer_breadcrumb - and only one uses `include_self=1` at the moment. Would be good to check the other instances (and update if needed)

KalobTaulien:
Hey @zerolab 

Index view shows the current page in the breadcrumb. That's not necessarily a bad thing. 
image

The edit view shows the current page we're on, which is the big one. This is what we primarily what to adjust. 
image

And this is the create view, which now shows its new "parent" page. 
image

I don't think we need to update the template tags from what they currently are. But I'm happy to make an adjustment if you think we should.

kaedroho:
Hey Kalob, do you think we need the final right arrow now? I think that was there before because the current page title was on the following line

benenright:
Hi Kalob

Chatted briefly with Karl on this - I usually always suggest on normal websites that showing the page you are on in the crumb is pretty worthless. But this makes total sense here.

I don’t think it makes sense to have an arrow after the page you are on, as the arrow kind of implies it’s a parent of the page you are on, so I agree with Karl and suggest we remove it off the end (except in the "create view" case where the final item in the crumb _is_ the actual parent).

Ben

KalobTaulien:
Hey @zerolab @kaedroho and @benenright I have a small update here. Only the Create View has the trailing arrow now. 

Edit view:
image

Create view:
image

List view:
image

kaedroho (approved):

rgs258:
This probably closes https://github.com/wagtail/wagtail/issues/4599 - nice work, folks!

PR #6938, opened 2021-03-19
Make use of listing button attrs in ButtonWithDropdownFromHook

kaedroho:
The ``attrs`` attribute on ``ButtonWithDropdownFromHook`` and all buttons within it was being ignored, which makes it impossible to do things like make an item in the "more" menu open something in a new tab.

This PR makes it pass through all ``attrs`` to the HTML. It also renames the ``title`` attr to ``aria-label`` for consistency with the other page listing buttons that are not in the "more" menu


zerolab:
@KIRA009 this looks similar to the changes you made in #7340
Worth ensuring consistency cc @jacobtoppm @Stormheg

PR #6850, opened 2021-02-28
Update type-related filters to support multiple models/arguments

ababic:
Originally implemented in #6216.

Updates the following `PageQuerySet` methods to support multiple models as arguments:
- `type()`
- `not_type()`
- `exact_type()`
- `not_exact_type()`

PR #6943, opened 2021-03-22
Remove redundant 'filter_page_type' helper methods

ababic:
#6850 updated the type() filter to support multiple models as arguments, so these views can utilise that now, instead of a custom `filter_page_type()` function.


kaedroho (approved):
Nice one, thanks!

PR #6945, opened 2021-03-23
Collection management changes 

cnk:
# Collection Management Pull Request

This implements [RFC 67](https://github.com/wagtail/rfcs/blob/collection-management/text/067-colection-management.md).

GroupEditScreen

* Do the tests still pass?  - Yes except for the ones I marked 'skip' pending discussion on use of CollectionPermissionPolicy
* Does the code comply with the style guide? (Run `make lint` from the Wagtail root) - yes
* For Python changes: Have you added tests to cover the new/fixed behaviour? - yes
* For front-end changes: Did you test on all of Wagtail’s supported browsers?  - no frontend changes
* For new features: Has the documentation been updated accordingly? - no I will add documentation once I get some feedback on the approach I took.

## Admin UI Changes

1. Changed group creation form to move Collections from the Object Permissions section (where groups are given permissions on all instances of an object type) to its own section like Pages, Documents, and Images.
2. Assigning a group permission to manage part of the collection hierarchy causes the collections item to appear in the group's Settings menu.
3. The collection management pages behave as before with two small differences for non-superusers:
    * Non-superusers cannot move a collection that is the apex of their collection hierarchy. For those collections, the 'parent' field changes from a select widget to a hidden field.
    * To prevent users locking themselves out of collection management, we do not let users delete the collection where their permissions are assigned.

CollectionEditChanges

## Questions

1. Is it OK that I used the existing CollectionPermissionPolicy from wagtail.core.permission_policies.collections import? It does what I want and does not appear to have ever been used within Wagtail itself.
1. I used hooks to add Collection Management to the group edit form. That put it at the end of the list but I would kind of like for it to appear between Pages and Documents. What do you think? And any tips about getting the page ordered that way?
1. Do you agree with not letting users move or delete the collections that provide their group collection management permissions (item 3 above)?

## Still to do

* Add docs
* Fix TestCollectionPermissionPolicy to be about collection management rather than about management of items within a collection


ababic:
Hi @cnk. I'll refrain from making an in-depth review for now (at least until the RFC has had a week or so to collect comments), but I can attempt to answer some of your questions:
1. Personally, I would prefer to create an entirely new class than change this existing one. `CollectionPermissionPolicy` is really provided to support users who want to use collections for grouping other things (custom models).
3. This sounds in-line with what we do for pages, so is at least consistent. I wonder whether slightly different wording on the labels might help clarify things? E.g. "Add descendants", "Edit descendants", "Delete descendants".

2 needs a little more investigation from my side, but I'm sure we can find a way!

cnk:
I finally got back to this. This PR is ready for review. 

I created a new permission policy: CollectionMangementPermissionPolicy which enforces the rules for add, change, and delete permissions when editing the collection hierarchy. I have updated the docs and tests and created a migration to give all groups that currently have permission to manage collections the same permissions on the root collection. 

About the migration, when we implemented collection hierarchies, we had a bug report about the assumption of a single root collection. A wagtail site should only have a single root collection (named Root) but given the previously report issue, should I code this migration more defensively? perhaps with a console message if there is more than one root collection? I am also a little concerned about deleting the original rows after new ones are set up (both directions) but also do not want to leave stale permission information in the database. So it would be great to have some skeptical eyes on the migration before it goes out.

wagtail/admin/views/collections.py
```diff
@@ -74,9 +118,17 @@ def form_valid(self, form):
...
-    def get_queryset(self):
-        # Only return descendants of the root node, so that the root is not editable
-        return Collection.get_first_root_node().get_descendants().order_by('path')
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        if self.request.user.is_superuser:
+            can_delete = True
+        else:
+            # Can not delete the collection where the delete permission is assigned
+            can_delete = self.permission_policy.descendants_of_collections_with_user_perm(
```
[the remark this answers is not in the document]
cnk: OK I refactored this so the query needed to enforce "can't move the collection giving you permissions" rule is a method in the collection edit view: `_user_may_move_collection` and the other uses of non-standard PermissionPolicy methods where changed to use the standard set.

wagtail/admin/views/collections.py
```diff
@@ -57,6 +70,37 @@ class Edit(EditView):
...
+
+    def get_form(self, form_class=None):
+        form = super().get_form(form_class)
+        user = self.request.user
+        # If this instance is a collection used to assign permissions for this user,
+        # do not let the user move this collection.
+        if self.permission_policy.user_has_any_permission_directly_on_instance(
+            user, ['add', 'change', 'delete'], form.instance
+        ):
+            form.fields['parent'].widget = HiddenInput()
```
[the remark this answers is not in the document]
cnk: I tried this but if I disable the parent form field, then it does not get submitted at all - which means I would have to make the parent field optional in the form. That is how the CollectionForm was originally written but then we needed to fill in the parent (see the diff for wagtail/admin/forms/collections.py in this PR). So instead of adding disabled = True, I added code in the `form_valid` method to prevent someone from posting a parent id they should not be able to send from the front end form. My test cases cover the hacks I could think of.

If you would like me to go back to the original form __init__ logic plus disabled = True to handle this, let me know and I will rearrange.

cnk:
Thanks for the comments @gasman and @ababic. I think this is ready for a re-review.

wagtail/admin/views/collections.py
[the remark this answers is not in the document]
ababic: Okay, so, in these cases, maybe you can just remove the field completely? That way, it isn't going to display, won't factor into validation, and the instance will just keep it's current value.

If you couple that with `self.instance.parent = value` for new objects, so that they have a value, I think all should be well.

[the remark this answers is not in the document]
cnk: Disabling a form field after the field is initialized does disable it in the rendered html. Unfortunately it doesn't tell the form class that the field is disabled so the logic that always keeps the initial value doesn't work. I solved this by putting the logic I had originally put in the create and edit views into the CollectionForm where it should have been in the first place. 

The one slightly unusual aspect of this is that because the logic depends on the current user and current permission policy, I had to pass both of those items from the view into the form itself.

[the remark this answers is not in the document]
ababic: @cnk did you see my suggestion about simply removing the field? Forms only update model fields which they themselves have a field for, so juts popping it within `__init__()` would totally work. 

With this new approach, I feel like the distinction between view/form has become a bit blurry. I expect much the same could be achieved by keeping the user-specific logic in the view, and passing more refined values to the form, e.g a queryset for the parent field.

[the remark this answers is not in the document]
cnk: I guess I do not understand what you mean by "removing the field". Do you mean something like `form.fields['parent'].pop()` in the view's `get_form` method? Does that work? I'll fall back a commit and try that approach. 

From the reading I did, pushing all the logic into the form object is the preferred django approach - my only concern was that it involved changing the signature of `CollectionForm.__init__`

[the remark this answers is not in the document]
cnk: So removing the field from the form does work and gives the desired behavior. This is ready for review (assuming the automated checks all pass).  (Thank goodness we use 'squash' when merging PRs. I would hate for the history of this PR to end up the the project history)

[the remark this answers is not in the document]
ababic: @cnk I feel like you took my idea and made it even better! This latest iteration feels a lot clearer.  Are you happier with it?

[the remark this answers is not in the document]
cnk: I am shocked that removing the field works but am pleased with the result. It took care of all the problems I was having when just disabling the field. I am satisfied with this iteration. Ready for review.

wagtail/admin/views/collections.py
```diff
@@ -57,26 +64,59 @@ class Edit(EditView):
...
+            form.fields.pop('parent')
+        # If this instance is a collection used to assign permissions for this user,
+        # do not let the user move this collection.
+        elif not self._user_may_move_collection(user, form.instance):
+            form.fields.pop('parent')
+        else:
+            # Filter collections offered in parent field by current user's add permissions
+            collections = self.permission_policy.instances_user_has_permission_for(user, 'add')
+            form.fields['parent'].queryset = collections
+            form.fields['parent'].disabled_queryset = form.instance.get_descendants(inclusive=True)
```
ababic, line 101: Is this attribute still used?

wagtail/admin/views/collections.py
```diff
@@ -57,26 +64,59 @@ class Edit(EditView):
...
+            form.fields['parent'].empty_label = None
```
ababic, line 102: It's difficult to ascertain why this line is here too. It feels like it might be more at home on the form?

wagtail/admin/views/collections.py
```diff
@@ -57,26 +64,59 @@ class Edit(EditView):
...
+        form.initial['parent'] = form.instance.get_parent().pk
+        return form
+
     def save_instance(self):
         instance = self.form.save()
-        parent_pk = self.form.data.get('parent')
-        if parent_pk and parent_pk != instance.get_parent().pk:
-            instance.move(Collection.objects.get(pk=parent_pk), 'sorted-child')
+        parent = self.form.cleaned_data.get('parent')
+        if parent and parent.pk != instance.get_parent().pk:
```
ababic, line 110: Would this work here?
```suggestion
        if 'parent' in form.changed_data:
```

wagtail/admin/views/collections.py
```diff
@@ -57,26 +64,59 @@ class Edit(EditView):
...
+            instance.move(Collection.objects.get(pk=parent.pk), 'sorted-child')
```
ababic, line 111: Is there a reason for refetching the object from the database here? I figure we could save ourselves an extra query here and just reuse `form.cleaned_data['parent']`?

cnk: Agreed.

wagtail/admin/views/collections.py
cnk: Yes that is much clearer.

wagtail/admin/views/collections.py
cnk: I could move `form.fields['parent'].disabled_queryset` and `form.initial['parent'] = form.instance.get_parent().pk` into the form's `__init__` method. But the lines only apply to the 'edit' case and I personally think keeping all the changes to the parent field in the same method makes it easier to see what is going on.

wagtail/admin/tests/test_collections_views.py
```diff
@@ -231,3 +464,56 @@ def test_post_collection_with_descendant(self):
 
         # Check that the collection was not deleted
         self.assertTrue(Collection.objects.get(id=self.collection.id))
+
+
+class TestDeleteCollection(CollectionInstanceTestUtils, TestCase, WagtailTestUtils):
```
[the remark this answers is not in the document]
cnk: Good catch. I improved test coverage for wagtail/admin/views/collections.py to 100% - which was very helpful when making changes Andy requested.

wagtail/admin/tests/test_collections_views.py
```diff
@@ -155,7 +309,86 @@ def test_post(self):
...
+
+    def test_marketing_user_cannot_move_collection_permissions_are_assigned_to(self):
+        response = self.get(collection_id=self.marketing_collection.id)
+        self.assertEqual(response.status_code, 200)
+        form_fields = response.context['form'].fields
+        # Can edit the name of the collection but not it's parent because this
+        # is the collection that gives me permission on a section of hierarchy
+        self.assertEqual(type(form_fields['name'].widget).__name__, 'TextInput')
+        self.assertEqual(type(form_fields['parent'].widget).__name__, 'HiddenInput')
+        self.assertNotContains(response, "Delete collection")
```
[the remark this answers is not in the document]
cnk: Andy suggested removing the field when it is not editable. With that change, the form ignores submitted parent ids unless the user should be able to move the collection.

wagtail/admin/views/collections.py
cnk: Yes.

ababic: Ah, I see. `disabled_queryset` is property setter that affects the widget choices!

wagtail/core/migrations/0063_collection_management_permissions.py
```diff
@@ -0,0 +1,76 @@
...
+
+    for perm in stock_permissions:
+        for group in perm.group_set.all():
+            gcp = GroupCollectionPermission(
+                group_id=group.id,
+                collection_id=root_collection.id,
+                permission_id=perm.id
+            )
+            gcp.full_clean()
+            gcp.save()
```
ababic, line 26: @cnk We should avoid cleaning here, as there's no handling for `ValidationError` in migrations. We just need to be confident that these changes do what we expect. We could maybe shorten to...
```suggestion
            GroupCollectionPermission.objects.create(
                group_id=group.id,
                collection_id=root_collection.id,
                permission_id=perm.id
            )
```

wagtail/core/migrations/0063_collection_management_permissions.py
```diff
@@ -0,0 +1,76 @@
...
+from django.db import migrations
+
+
+def grant_instance_level_collection_management_permissions(apps, schema_editor):
+    Collection = apps.get_model('wagtailcore.Collection')
+    GroupCollectionPermission = apps.get_model('wagtailcore.GroupCollectionPermission')
+    Permission = apps.get_model('auth.Permission')
+
+    # Give all groups who currently manage collections permission to manage root collection
+    root_collection = Collection.objects.filter(depth=1).order_by('path').first()
```
ababic, line 11: If we expect the collection to always exist here, we should use `get()` - If not, we should handle the `root_collection is None` case and gracefully bow out here :)

wagtail/core/migrations/0063_collection_management_permissions.py
```diff
@@ -0,0 +1,76 @@
...
+    Collection = apps.get_model('wagtailcore.Collection')
+    ContentType = apps.get_model('contenttypes.ContentType')
+    GroupCollectionPermission = apps.get_model('wagtailcore.GroupCollectionPermission')
+    Permission = apps.get_model('auth.Permission')
+
+    collection_content_type = ContentType.objects.get(
+        model='collection',
+        app_label='wagtailcore'
+    )
+    root_collection = Collection.objects.filter(depth=1).order_by('path').first()
```
ababic, line 41: https://github.com/wagtail/wagtail/pull/6945/files#r707668901 applies here too.

wagtail/core/migrations/0063_collection_management_permissions.py
```diff
@@ -0,0 +1,76 @@
...
+    group_collection_permissions = GroupCollectionPermission.objects.filter(
+        collection=root_collection,
+        permission__in=stock_permissions
+    ).prefetch_related('group')
+
+    for row in group_collection_permissions.all():
+        perm = Permission.objects.get(
+            content_type=collection_content_type,
+            codename=row.permission.codename
+        )
```
ababic, line 58: We're likely not talking huge numbers of operations here, but could fetch the `Permission` objects in a single query ahead of this loop, then perhaps pull them out of a dictionary here based on codename?

wagtail/core/permission_policies/collections.py
```diff
@@ -107,6 +107,22 @@ def _users_with_perm(self, actions, collection=None):
...
+        """
+        Return a queryset of all collections in which the given user has
+        permission to perform any of the given actions
+        """
+        if user.is_active and user.is_superuser:
+            # active superusers can perform any action (including unrecognised ones)
+            # in any collection
+            return Collection.objects.all()
+
+        elif not user.is_authenticated:
```
ababic, line 120:
```suggestion
        if not user.is_authenticated:
```

wagtail/core/permission_policies/collections.py
```diff
@@ -107,6 +107,22 @@ def _users_with_perm(self, actions, collection=None):
...
+            return Collection.objects.none()
+
+        else:
+            return self._collections_with_perm(user, actions)
```
ababic, line 124:
```suggestion
        return self._collections_with_perm(user, actions)
```

ababic (commented):
@cnk Thanks very much for all your hard work on this. I will have a real play with this later in the week and let you know if anything crops us - in the meantime, I picked up on a few small things.

wagtail/admin/views/collections.py
ababic: Likewise for the `disabled_queryset`... Something like this would do wonders:
```
# let the CollectionChoiceField know which options to disable
form.fields['parent'].disabled_queryset = form.instance.get_descendants(inclusive=True)
# prevent the misleading empty/blank option from appearing
form.fields['parent'].empty_label = None

```

cnk: I removed the `empty_label` line - I don't think it bought us anything. 

Hopefully the updated comments for this section playing with the parent widget are now clearer.

wagtail/core/migrations/0063_collection_management_permissions.py
cnk: That is much cleaner. I have incorporated the direct `create` in my other changes to the migration.

wagtail/core/migrations/0063_collection_management_permissions.py
cnk: The way I changed the migrations, they just do nothing if there are no root collections.

The case I had had in mind when I wrote that line was a problem someone had in the support channel in slack - where they had more than one root collection in their app. With the changes in our collection management views, I think everything should work even if an app somehow ends up with more than one 'root' collection (in TreeBeard's modeling, that is any collection with depth = 1) so I altered the migrations so they will run no matter how many root collections an app already has - even though the expectation is still that each wagtail installation will have a single root collection.

wagtail/core/migrations/0063_collection_management_permissions.py
cnk: Agreed. The new version of the backwards migration prefetches group and permission objects for use creating model-level permissions.

cnk:
@ababic Thanks once again for your suggestions. Going over the migrations again, I found another logic error that needed correcting - in addition to the items you pointed out. The forward migration was adding all three stock permissions to the instance-level permissions table - even if a group had had only one or two of them at the model level. I fixed that while incorporating your other comments.

wagtail/core/migrations/0063_collection_management_permissions.py
```diff
@@ -3,60 +3,54 @@
...
-    stock_permissions = Permission.objects.filter(
-        content_type__app_label='wagtailcore',
-        content_type__model='collection',
-        codename__in=['add_collection', 'change_collection', 'delete_collection']
-    ).prefetch_related('group_set').all()
+    groups_w_permissions = Group.objects.filter(
+        permissions__content_type__app_label='wagtailcore',
+        permissions__content_type__model='collection',
+        permissions__codename__in=['add_collection', 'change_collection', 'delete_collection']
+    ).values('id', 'name', 'permissions__id', 'permissions__codename')
```
ababic, line 18: This is beautiful :)

wagtail/core/migrations/0063_collection_management_permissions.py
```diff
@@ -3,60 +3,54 @@
...
+    root_collections = Collection.objects.filter(depth=1).all()
     group_collection_permissions = GroupCollectionPermission.objects.filter(
-        collection=root_collection,
-        permission__in=stock_permissions
-    ).prefetch_related('group')
+        permission__content_type__app_label='wagtailcore',
+        permission__content_type__model='collection',
+        permission__codename__in=['add_collection', 'change_collection', 'delete_collection'],
+        collection__in=root_collections,
+    ).prefetch_related('group', 'permission')
```
ababic, line 50: Since the `Group` and `Permission` objects are pretty tiny (in memory usage terms), I think we could safely get away with `select_related()` here instead. Doing so should reduce this from 3 database queries to 1.

cnk: Ahh good catch. Corrected.

cnk:
@gasman I think Andy and I are finally happy with this code. I believe I have addressed your change requests but let me know if there are more changes you would like to see before this gets merged in. I would really love to get it into 2.15.

gasman:
Merged in f7fcae0a1d6640458afc6876c718dd54aa50e069 + parents.

PR #6946, opened 2021-03-23
Improve db queries

### LINKED ISSUE #6876: Large amount of queries in wagtail admin

dest81:
Resolves #6876 
With those changes number of db queries goes down from 400 to 60.
For more information look into issue #6876

### DIFF
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -3312,7 +3312,6 @@ class PagePermissionTester:
     def user_has_lock(self):
         return self.page.locked_by_id == self.user.pk
 
-    @cached_property
     def page_locked(self):
         current_workflow_task = self.page.current_workflow_task
         if current_workflow_task:
@@ -3405,7 +3404,7 @@ class PagePermissionTester:
             return False
         if (not self.page.live) or self.page_is_root:
             return False
-        if self.page_locked:
+        if self.page_locked():
             return False
 
         return self.user.is_superuser or ('publish' in self.permissions)
@@ -3419,7 +3418,7 @@ class PagePermissionTester:
         return self.user.is_superuser or ('publish' in self.permissions)
 
     def can_submit_for_moderation(self):
-        return not self.page_locked and self.page.has_workflow and not self.page.workflow_in_progress
+        return not self.page_locked() and self.page.has_workflow and not self.page.workflow_in_progress
 
     def can_set_view_restrictions(self):
         return self.can_publish()
--- a/wagtail/core/models.py
+++ b/wagtail/core/models.py
@@ -2839,21 +2839,23 @@ class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):
     def current_workflow_state(self):
         """Returns the in progress or needs changes workflow state on this page, if it exists"""
         try:
-            return WorkflowState.objects.active().get(page=self)
+            return WorkflowState.objects.active().select_related("current_task_state__task").get(page=self)
         except WorkflowState.DoesNotExist:
             return
 
     @property
     def current_workflow_task_state(self):
         """Returns (specific class of) the current task state of the workflow on this page, if it exists"""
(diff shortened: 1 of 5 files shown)


wagtail/core/models.py
```diff
@@ -2839,21 +2839,23 @@ def workflow_in_progress(self):
...
+            return WorkflowState.objects.active().select_related("current_task_state__task").get(page=self)
         except WorkflowState.DoesNotExist:
             return
 
     @property
     def current_workflow_task_state(self):
         """Returns (specific class of) the current task state of the workflow on this page, if it exists"""
-        if self.current_workflow_state and self.current_workflow_state.status == WorkflowState.STATUS_IN_PROGRESS and self.current_workflow_state.current_task_state:
-            return self.current_workflow_state.current_task_state.specific
+        current_workflow_state = self.current_workflow_state
```
femto113, line 2849: I'm guessing a huge share of the difference in queries comes from this one line, would be interesting to test with and without it to verify.

dest81: @femto113 Yes, if we fix only this  method number goes down from 400 to 200. 
also as you saw PR has two commits: 

1. first commit reduces number db queries to 90
2. second commit reduces to 60

Let me know which variant do you prefer and I will update PR

wagtail/core/models.py
```diff
@@ -3310,9 +3312,11 @@ def __init__(self, user_perms, page):
     def user_has_lock(self):
         return self.page.locked_by_id == self.user.pk
 
+    @cached_property
```
zerolab, line 3315: question: does this fall under the same problem area as https://github.com/wagtail/wagtail/pull/6878#pullrequestreview-604997462 ? cc @jacobtoppm

zerolab (commented):
@dest81 thank you for this PR. it does make quite a difference!

Screenshot 2021-04-14 at 17 29 49

Need to do further checking making `page_locked` a cached property. It certainly saves ~30 queries. At the same time it checks the workflow state so may fall under a similar scenario as https://github.com/wagtail/wagtail/pull/6878#pullrequestreview-604997462

There is also a minor suggestion for the comment in the `ActionMenuItem.is_shown` docblock, but that can be tackled during the merge.

wagtail/core/models.py
jacobtoppm: I believe so, yes. At least I'm thinking about the scenario where someone approves a page (not locked) and it moves into another task where the page is now locked for them, and I'm not sure this would behave correctly in that case without clearing the cached property.

zerolab (changes requested):
@dest81 based on https://github.com/wagtail/wagtail/pull/6946#discussion_r613924845, could you please undo the `cached_property` changes? Then we should be good to get this in

dest81:
@zerolab I have reverted back cached_property changes. There is one failed check but it doesn't look related to the changes in this PR.

zerolab (approved):
LGTM!

zerolab:
Merge in b215573 + parents.

Thank you!

PR #6952, opened 2021-03-30
New styling for commenting

kaedroho:
Adds support for https://github.com/jacobtoppm/wagtail-comment-frontend/pull/11


jacobtoppm (changes requested):
Generally looking great, but 2 issues right now:

- Adding a new reply causes no avatar to show on the added reply (all other avatars on existing replies display fine)
- On Chrome, at various levels of zoom, the circles surrounding the avatar seem to have gaps in some places between the avatar and the border, which look a little odd. The border itself also looks a little pixellated - any way we could smooth this a little more? (This is a minor visual bug, but thought I'd mention)

jacobtoppm (approved):

PR #6957, opened 2021-04-01
Commenting: Delete comments when block is deleted

kaedroho:


client/src/components/CommentApp/main.tsx
```diff
@@ -204,6 +205,20 @@ export class CommentApp {
     this.store.dispatch(setFocusedComment(commentId, { updatePinnedComment: true }));
     return commentId;
   }
+  invalidateContentPath(contentpath: string) {
```
jacobtoppm, line 208: This feels like it could be an action, rather than dispatching lots of sequential actions (which we can't always avoid, but would be nice to where possible). 

If we do leave it as a function that creates lots of actions, we already have a selector for contentpath, so should probably use that rather than relying on the exact shape of the state here

jacobtoppm (changes requested):

client/src/components/CommentApp/main.tsx
kaedroho: This makes sense, I'll change it, thanks!

kaedroho: Done!

jacobtoppm (changes requested):
Deletion looks good, aside from the reassignment the linter is failing on!

I think we needed to get the widget unsubscription for field level commenting in here as well though, unless you wanted to do that in a separate PR?

kaedroho:
Rebased!

PR #6959, opened 2021-04-01
Bulk actions for pages, with search and select functionality

KIRA009:


ababic:
Hey there @KIRA009! This is shaping up really nicely. I know this is a WIP, so you may have these things on your radar already, but I have a few comments/suggestions:

1. For the bulk action views themselves, I feel like class-based views could be really valuable there. Not only would this allow you to reduce some repetition from this code, but it would give users a much more convenient starting point for defining their own bulk actions. Classes also have attributes - which are a great way to support customisation with this kind of thing.
2. I could be missing something, but we appear to have lost the ability to reorder pages here. Is that intentional?
3. There appear to be two concepts here: filtering, and the bulk actions themselves. It's difficult to tell from just looking at the code, but are filter options only visible when using bulk actions, or are they visible all the time? I'm feeling like a way to define and apply filters to the explorer is a really useful feature in its own right, and would definitely have value outside of bulk actions. I wonder if that might make for a separate PR, to allow you to focus on the two things separately?
4. On the subject of filtering, Django already has quite a well-defined approach to this in its admin app (which the IndexView in Wagtail's modeladmin app also utilizes). I wonder whether there could be something there worth reusing here too?

KIRA009:
Thanks for the suggestions @ababic. I would like to let you know that this pr is part of my GSOC proposal. My [final proposal](https://iamkira99-personal.s3.ap-south-1.amazonaws.com/gsoc/GSOC.pdf) included several refactors I would like to do on this pr to meet the requirements. I will list the relevant details here

1. Yes, that is a great suggestion. Using class based views would make it more readable and easier for the users.
However I had a slightly different idea of implementation, one that still uses function based views, but all the actual processing is done by a class that is provided by the user. From my proposal:
***
We have to refactor the bulk actions feature so as to allow user to add custom actions, which will be displayed as options under the `More` dropdown in the admin dashboard. There is a similar decorator, `register_snippet` which when applied to a Django model, allow it to be accessible via the Snippets UI. We can make a similar approach for allowing the user to add custom bulk actions.
```python
def register_custom_bulk_actions(cls):
    custom_bulk_actions.append(cls)
    """
    a class would be a better implementation for bulk actions
    because it can contain the display name of the action, and the
    corresponding method for performing the action. This would
    also allow us to define an abstract base class like a
    CustomBulkAction, which can be used by the end user for designing
    their bulk actions.

    class CustomBulkAction(ABC):
        @abstractproperty
        def display_name(self):
            pass

        @abstractstaticmethod
        def act():
            # this method performs the action
            pass
    """
```
The bulk actions UI in my [pr](https://github.com/wagtail/wagtail/pull/6959) is currently displayed by fetching the buttons using the `register_bulk_action_choices` hook. The custom actions could then be fetched in this hook and displayed in the `More` dropdown.  to something which process the action based on the type of the action. 
***
We could store all the bulk action classes in a dict with their name as the key, and when a request is made to /multiple/<action_name>/, we could then retrieve the corresponding bulk action class and pass the data to that class for the processing.
2. I had some meets with olly, benright and karl. It was concluded, that the 'Ordering' column was not in place with the other columns, since that column alone was one which performed an 'action'. IIRC, it was decided that it would be moved to the upper panel, alongside buttons like bulk actions.
3. and 4. The filters are currently designed to work like they do in github, supporting syntax like "is:page status:draft". It is independent of bulk actions, so it can definitely be addressed in a separate pr, and yes, I came across a similar thing in wagtail after I made the pr. I read the code and it seems like it supports filters of specific categories only. In my proposal, I have illustrated on how I plan to implement it in a way that it can be extendable by users, allowing them to define their own filters. From my proposal
***
The `Filters` feature works by defining a dict of filters, e.g.
```python
filters = {
        'status': {
            'live': Q(live=True),
            'draft': Q(live=False),
        }
}
```
This filters are then added to a `filters_to_be_applied` dict using a filter string of the form `filter1:value1 filter2: value2`. This makes the filters easily extendable, so one can do,
```python
filters.update(**{
    'content_type': {
        'json': Q(content_type='application/json'),
        'html': Q(content_type='html')
    }
})
```
This will then allow filter strings like `content_type:html status:live` to work as expected.
***
5. Yes, that is also part of my GSOC proposal.
***
Implement custom bulk actions for snippets similar to the previous bulk actions, and bring the snippets design in line with the pages design.
***

Once again thanks for the feedback, please let me know if there are any improvements to be made from my proposal.

ababic:
@KIRA009 ah, great! I hadn't seen your full proposal, so that fills in a lot of gaps for me, thanks :)

1. The approach you have in mind sounds great. I think there are a few places in Wagtail where class-based views are instantiated to handle a request from a function (e.g. https://github.com/wagtail/wagtail/blob/main/wagtail/contrib/modeladmin/options.py#L348-L412) so that should work nicely! Since those classes need to do things that views generally do (check permissions, process form posts, respond with a HttpResponse), so using a generic Django view class as a base may be beneficial there.
2. Ordering: Oh, excellent. That ordering option is a long-standing bugbear of mine too. Will be great to see that have a more natural home.
3 & 4. Filtering: I do like the simplicity of the approach in your proposal. I would like to see better feature-parity in general across the explorer/snippet/modeladmin views though. I know there is some UI work happening in the background (I'm due to have a catch-up with the team tomorrow) - I'm sure we'll figure it out between us :)
5. Excellent!

wagtail/admin/wagtail_hooks.py
```diff
@@ -127,6 +127,59 @@ def register_workflow_tasks_menu_item():
...
+        priority=30
+    )
+
+
+@hooks.register('register_bulk_action_choices')
+def bulk_action_choices(page, is_parent=False, next_url=None):
+    if page:
+        yield PageListingButton(
+            _('Move'),
+            reverse('wagtailadmin_bulk_delete', args=[page.id]) + '?' + urlencode({'next': next_url}),
```
Stormheg, line 159: Looks like this should be `wagtailadmin_bulk_move` which has not been implemented yet.

wagtail/admin/templates/wagtailadmin/pages/listing/_table_headers_explore.html
```diff
@@ -17,24 +17,22 @@
 {% endcomment %}
 
 <tr class="table-headers">
-    {% if show_ordering_column %}
```
Stormheg, line 20: **Question:** how does one change the order of the page tree without this option?

Stormheg (commented):
Thanks @KIRA009 this looks great already!

Three general notes

1. Looks like bulk move hasn't been implemented yet. The move button behaves like a delete button.
2. How does one change the order of pages in the page tree? The button for that was removed.

3. Regarding the dropdown arrow, I'm not sure it's visible enough.
image

Reading back the slack chat in #gsoc-bulk-actions it looks like @benenright mentioned this too. https://wagtailcms.slack.com/archives/C01RAKTN269/p1616501091018700

> thinking a little bit about mobile: I do wonder if we should combine the checkbox and little arrow into one button (the current dual-button sizes are too small for touch devices) that forces you to choose from the dark dropdown (‘ALL’ being the first option anyway).

So far for my first pass. Keep up the good work Shohan!

wagtail/admin/templates/wagtailadmin/pages/listing/_table_headers_explore.html
```diff
@@ -17,24 +17,22 @@
...
-    {% endif %}
-    <th class="title">
-        {% trans 'Title' as title_label %}
-        {% page_table_header_label label=title_label sortable=sortable sort_field='title' parent_page_title=parent_page.title %}
+    <th class="bulk-actions-filter-checkbox">
+        <div>
+            <input type="checkbox" aria-label="Bulk action checkbox" />
+            {% bulk_action_filters %}
+            <div class="bulk-actions-choices u-hidden">
+                <span>All 8 on this page selected. Select all 43 in listing |</span>
```
jacobtoppm: Should we be hardcoding the 8 and 43 here? These seem quite specific. Additionally, text needs to be translatable.

wagtail/admin/templates/wagtailadmin/pages/listing/_table_headers_explore.html
```diff
@@ -17,24 +17,22 @@
...
+                <ul>{% bulk_action_choices %}</ul>
+            </div>
+        </div>
+    </th>
+    <th class="search">
+        <div class="nav-search">
+            <button class="button" type="submit">Search</button>
+            <label for="page-search-q">Search</label>
+            <input aria-label="Search text" type="text" id="page-search" name="q" placeholder="Search">
```
jacobtoppm: Placeholder should presumably be translatable.

client/src/entrypoints/admin/bulk-actions.js
```diff
@@ -0,0 +1,56 @@
...
+    document.querySelector(`.${BULK_ACTION_CHOICES_CLASS}`).classList.add('u-hidden');
+    document.querySelectorAll(`.${BULK_ACTION_CHECKBOX_CLASS}`).forEach(el => el.classList.remove('show'));
+    document.querySelector(`.${BULK_ACTION_CHECKBOX_FILTER_CLASS}`).setAttribute('colspan', '1');
+  } else if (checkedState.checkedPages.size === document.querySelectorAll(`.${BULK_ACTION_CHECKBOX_CLASS}`).length) {
+    // all checkboxes are checked
+    document.querySelector(`.${BULK_ACTION_CHECKBOX_FILTER_CLASS} input`).checked = true;
+  } else if (checkedState.checkedPages.size === 1 && prevLength === 0) {
+    // 1 checkbox is checked for the first time
+    document.querySelectorAll(`.${BULK_ACTION_CHECKBOX_CLASS}`).forEach(el => {
+      el.classList.remove('show');
```
jacobtoppm, line 45: Why is 'show' being removed only to be added again?

wagtail/admin/templates/wagtailadmin/pages/listing/_list.html
```diff
@@ -79,3 +77,8 @@
         {% endif %}
     </tbody>
 </table>
+
+<script src="{% versioned_static 'wagtailadmin/js/bulk-actions.js' %}"></script>
+<script>
+    window.AddBulkActionCheckboxEventListeners()
+</script>
```
jacobtoppm, line 83: No newline

wagtail/admin/templates/wagtailadmin/pages/listing/_list.html
```diff
@@ -2,9 +2,7 @@
 {% load l10n %}
 {% load wagtailadmin_tags %}
 <table class="listing {% if full_width %}full-width{% endif %} {% block table_classname %}{% endblock %}">
-    {% if show_ordering_column %}
-        <col width="65px" />
-    {% endif %}
+    <col width="10px" />
```
jacobtoppm, line 5: Now this no longer depends on a variable from Django, I think it would be nicer to move this into css.

client/src/entrypoints/admin/bulk-actions.js
```diff
@@ -1,56 +1,147 @@
...
+
+  if (checkedState.checkedPages.size === checkedState.numPages) {
+    /* when all checkboxes in the page are checked */
+    document.querySelector(`.${BULK_ACTION_SELECT_ALL_CHECKBOX_TH} input`).checked = true;
+  }
+
+  if (checkedState.checkedPages.size > 0) {
+    /* Update text on number of pages */
+    document.querySelector(`.${BULK_ACTION_NUM_PAGES_SPAN}`).textContent =
+    `${checkedState.checkedPages.size === checkedState.numPages ? 'All ' : ''} ${checkedState.checkedPages.size}`;
```
jacobtoppm, line 62: This needs to be translatable.

client/src/entrypoints/admin/bulk-actions.js
```diff
@@ -1,56 +1,147 @@
...
+
+/* Gets the value of given name from the query string in url */
+function getParameterByName(name) {
+  var match = RegExp('[?&]' + name + '=([^&]*)').exec(window.location.search);
+  return match && decodeURIComponent(match[1].replace(/\+/g, ' '));
+}
+
+
+/* Updates the content of BULK_ACTION_NUM_PAGES_IN_LISTING_SPAN with the new count of pages */
+function udpateNumPagesInListing(filterQueryString) {
```
jacobtoppm: udpate => update

client/src/entrypoints/admin/bulk-actions.js
```diff
@@ -1,56 +1,147 @@
...
 }
 
-function AddBulkActionCheckboxEventListeners() {
-  document.querySelectorAll(`.${BULK_ACTION_CHECKBOX_CLASS}`)
-    .forEach(el => el.addEventListener('change', SelectBulkActionsCheckboxes));
-  document.querySelector(`.${BULK_ACTION_CHECKBOX_FILTER_CLASS}`).addEventListener('change', SelectBulkActionsFilter);
+
+/* Gets the value of given name from the query string in url */
+function getParameterByName(name) {
+  var match = RegExp('[?&]' + name + '=([^&]*)').exec(window.location.search);
```
jacobtoppm, line 69: Is there a reason you're using `var` rather than `const` here?

wagtail/admin/templates/wagtailadmin/pages/bulk_actions/confirm_bulk_delete.html
```diff
@@ -0,0 +1,33 @@
+{% extends "wagtailadmin/base.html" %}
+{% load i18n wagtailadmin_tags %}
+{% block titletag %}{% blocktrans count counter=pages|length with title=page.get_admin_display_title  %}Delete 1 page {% plural %}Delete {{ counter }} pages{% endblocktrans %}{% endblock %}
+
+{% block content %}
+    {% trans "Delete" as del_str %}
+    {% include "wagtailadmin/shared/header.html" with title=del_str subtitle=page.get_admin_display_title icon="doc-empty-inverse" %}
+
+    <div class="nice-padding">
+        <p>{% blocktrans with snippet_type_name=model_opts.verbose_name_plural %}Are you sure you want to delete these pages?{% endblocktrans %}</p>
```
jacobtoppm, line 10: This doesn't use the `with` variable

wagtail/admin/views/bulk_actions/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .delete import delete
+from .unpublish import unpublish
+from .publish import publish
```
jacobtoppm: Needs a newline

wagtail/admin/views/bulk_actions/delete.py
```diff
@@ -0,0 +1,66 @@
...
+            ),
+        })
+    elif request.method == 'POST':
+        num_parent_pages = 0
+        num_child_pages = 0
+        with transaction.atomic():
+            for page in pages:
+                num_parent_pages += 1
+                num_child_pages += page.get_descendant_count()
+                for fn in hooks.get_hooks('before_delete_page'):
```
jacobtoppm: I think we need to think about these hooks. There's a few considerations here, for one, if one wants to return a response for `after_delete_page`, for instance, it could shortcut the rest of the deletions, unlike if someone is using the same result in the single deletion view. Something for us to talk about in the meeting, I think! One solution could be running the hooks with lists at start and end of all deletions.

wagtail/admin/views/pages/listing.py
```diff
@@ -11,6 +12,40 @@
 from wagtail.core.models import Page, UserPagePermissionsProxy
 
 
+def apply_filters(filter_query):
+    # filter_query should be of the format 'filter1:value1 filter2:value2'
+    filter_query = filter_query.strip()
+    if not filter_query:
+        return dict()
+    filter_query = filter_query.split(' ')
```
jacobtoppm: I think we should consider sharing a filter function with https://github.com/wagtail/wagtail/blob/571b9e1918861bcbe64a9005a4dc65f3a1fe7a15/wagtail/admin/views/pages/search.py#L67

wagtail/admin/views/bulk_actions/move.py
```diff
@@ -0,0 +1,109 @@
...
+    child_pages = set()
+    if dest_page_id:
+        viewed_page = get_object_or_404(Page, id=dest_page_id)
+    else:
+        viewed_page = Page.get_first_root_node()
+
+    for _page in get_list_or_404(Page, id__in=page_ids):
+        page_to_move = _page.specific
+        page_perms = page_to_move.permissions_for_user(request.user)
+        if not page_perms.can_move():
```
jacobtoppm, line 31: Can you select a page you don't have permissions to move? This could be an awkward user experience

jacobtoppm (commented):
Generally looking great so far, nice work! In general I think we should discuss splitting this into smaller commits or smaller PRs, as some of these are getting rather weighty

wagtail/admin/views/bulk_actions/move.py
jacobtoppm: (as discussed in the meeting, these permissions/inability to move parts should probably be explicitly mentioned to the user on the confirmation view for each one)

wagtail/admin/templates/wagtailadmin/pages/bulk_actions/confirm_bulk_delete.html
zerolab: @jacobtoppm `blocktrans` does use `with` in its syntax - https://docs.djangoproject.com/en/3.0/topics/i18n/translation/#std:templatetag-blocktrans. Unless I misunderstood your comment

jacobtoppm: Sorry, don't think I was clear enough. I know `with` is valid syntax, but `snippet_type_name` isn't actually being used inside the tag.

wagtail/admin/views/bulk_actions/delete.py
```diff
@@ -0,0 +1,65 @@
...
+                num_parent_pages += 1
+                num_child_pages += page.get_descendant_count()
+                page.delete(user=request.user)
+
+                for fn in hooks.get_hooks('after_delete_page'):
+                    result = fn(request, page)
+                    if hasattr(result, 'status_code'):
+                        return result
+
+        messages.success(request, _(f'You have successfully deleted {num_parent_pages} pages including {num_child_pages} child pages.'))
```
zerolab, line 64: suggestion: use proper pluralisation - https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#pluralization

The project convention is not to use f-strings where translation is needed. Instead use % formatting. More details in #6238

_("%(time_period)s ago") % {'time_period': time_period}

Additionally, the message could read something along the lines of `_("%(parent_pages)d pages and %(child_pages)d child pages" % {'parent_pages': num_parent_pages, 'child_pages': num_child_pages })` -- most action success messages are impersonal. ie. "Page 'TITLE' deleted. " or "Page 'TITLE' has been published." instead of `You have successfully deleted page "TITLE"`

I suggest changing for all actions

wagtail/admin/views/bulk_actions/move.py
```diff
@@ -69,35 +72,39 @@ def move_confirm(request, parent_page_id, dest_page_id):
...
             'destination': destination,
             'submit_url': reverse('wagtailadmin_bulk_move_confirm', args=[parent_page_id, dest_page_id])
             + '?' + urlencode([('id', page_id) for page_id in page_ids]),
         })
     else:
         for page_to_move in pages_to_move:
+            for fn in hooks.get_hooks('before_move_page'):
+                result = fn(request, page_to_move, destination)
+                if hasattr(result, 'status_code'):
+                    return result
```
zerolab, line 108: @KIRA009 can you add a note to revisit this? We discussed we'd cover this during the next core meeting.
I am not convinced this should apply for bulk actions but we're split on this so wider consensus is needed

wagtail/admin/templates/wagtailadmin/pages/listing/_list.html
```diff
@@ -20,6 +18,7 @@
         {% if parent_page %}
             {% page_permissions parent_page as parent_page_perms %}
             <tr class="index {% if not parent_page.live %} unpublished{% endif %} {% block parent_page_row_classname %}{% endblock %}">
+                <td></td>
                 <td class="title" {% if show_ordering_column %}colspan="2"{% endif %}>
```
zerolab, line 22:
```suggestion
                <td class="title" colspan="2">
```
would keep the page title consistent and in line with the left edge of the breadcrumb as well as the first column below

wagtail/admin/views/pages/bulk_actions/delete.py
```diff
@@ -0,0 +1,58 @@
+from django.utils.translation import gettext_lazy as _
+from django.utils.translation import ngettext
+
+from wagtail.admin.views.pages.bulk_actions.page_bulk_action import PageBulkAction
+from wagtail.core import hooks
+
+
+class DeleteBulkAction(PageBulkAction):
```
zerolab, line 8: suggestion: update `action_priority` so the actions follow the listing actions (move, copy, delete, publish, unpublish)

Also it would be great if the delete bulk action button had the `.no` or `.serious` class. That is not happening on the page listing as delete is under "more"

zerolab (changes requested):
Lots of great progress @KIRA009!

Added my first pass of notes and suggestions. Mostly templates and functionality.
Will continue tomorrow with bulk action hooks and actions

wagtail/admin/templates/wagtailadmin/pages/bulk_actions/confirm_bulk_move.html
```diff
@@ -0,0 +1,82 @@
...
+    <div class="nice-padding">
+        <span>{% trans "Are you sure you want to move these pages?" %}</span>
+        <ul>
+            {% for page in pages %}
+            <li>
+                <a href="{% url 'wagtailadmin_pages:edit' page.page.id %}" target="_blank" rel="noopener noreferrer">{{ page.page.title }}</a>
+                {% if not page.page.is_leaf %}
+                <ul>
+                    <li>{% trans "This page has child pages" %}</li>
+                </ul>
```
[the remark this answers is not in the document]
KIRA009: Do you suggest mentioning the number of child pages, or removing this line itself?

client/src/entrypoints/admin/bulk-actions.js
```diff
@@ -0,0 +1,165 @@
...
+    if (checkedState.checkedPages.size === checkedState.numPages) {
+      document.querySelector(`.${BULK_ACTION_NUM_PAGES_IN_LISTING}`).classList.remove('u-hidden');
+      if (checkedState.numPages === 1) {
+        numPagesSelected = wagtailConfig.STRINGS.NUM_PAGES_SELECTED_SINGULAR;
+      } else {
+        numPagesSelected = wagtailConfig.STRINGS.NUM_PAGES_SELECTED_ALL.replace('{0}', checkedState.checkedPages.size);
+      }
+    } else {
+      document.querySelector(`.${BULK_ACTION_NUM_PAGES_IN_LISTING}`).classList.add('u-hidden');
+      if (checkedState.checkedPages.size === 1) {
```
jacobtoppm, line 76: The `numPagesSelected` logic seems to be repeated twice. I would suggest refactoring to only find it once.

wagtail/admin/templates/wagtailadmin/pages/bulk_actions/confirm_bulk_delete.html
```diff
@@ -0,0 +1,49 @@
...
+            </li>
+            {% endfor %}
+        </ul>
+        {% endif %}
+        {% if pages_with_no_access %}
+        <p>{% blocktrans count counter=pages_with_no_access|length %}The following page cannot be deleted{% plural %}The following pages cannot be deleted{% endblocktrans %}</p>
+        <ul>
+            {% for page in pages_with_no_access %}
+            <li>
+                <a href="{% url 'wagtailadmin_pages:edit' page.id %}" target="_blank" rel="noopener noreferrer">{{ page.title }}</a>
```
jacobtoppm, line 34: Feels like we shouldn't offer the edit link if they don't have edit permissions (this is also what we do in the explore listing, see https://github.com/wagtail/wagtail/blob/main/wagtail/admin/templates/wagtailadmin/pages/listing/_page_title_explore.html)

wagtail/admin/templates/wagtailadmin/pages/bulk_actions/confirm_bulk_delete.html
```diff
@@ -0,0 +1,49 @@
...
+                    This will also delete {{ counter }} more subpages.
+                    {% endblocktrans %}
+                    </p>
+                {% endif %}
+            </li>
+            {% endfor %}
+        </ul>
+        {% endif %}
+        {% if pages_with_no_access %}
+        <p>{% blocktrans count counter=pages_with_no_access|length %}The following page cannot be deleted{% plural %}The following pages cannot be deleted{% endblocktrans %}</p>
```
jacobtoppm, line 30: Can we be clearer? "You don't have permission to delete...", or are there other reasons a user might not be able to delete?

wagtail/admin/templates/wagtailadmin/pages/bulk_actions/confirm_bulk_move.html
```diff
@@ -0,0 +1,93 @@
...
+        {% endif %}
+
+        {% if pages %}
+        <form action="{{ submit_url }}" method="POST">
+            {% csrf_token %}
+            <ul class="fields">
+            {% include "wagtailadmin/shared/field_as_li.html" with field=form.chooser %}
+            {% if pages_without_destination_access or pages_with_duplicate_slugs %}
+            <li>
+                <div class="field boolean_field checkbox_input">
```
jacobtoppm, line 69: Is there a pattern you could use here? Ie another `field_as_li`?

wagtail/admin/templates/wagtailadmin/pages/bulk_actions/confirm_bulk_unpublish.html
```diff
@@ -0,0 +1,62 @@
...
+                        This page has {{ counter }} subpages
+                        {% endblocktrans %}
+                    {% endif %}
+                </p>
+            </li>
+            {% endfor %}
+        </ul>
+        {% endif %}
+        {% if pages_with_no_access %}
+        <p>{% blocktrans count counter=pages_with_no_access|length %}The following page cannot be unpublished{% plural %}The following pages cannot be unpublished{% endblocktrans %}</p>
```
jacobtoppm, line 29: If this is only going to be permission related, could we be clearer?

wagtail/admin/views/pages/bulk_actions/move.py
```diff
@@ -0,0 +1,100 @@
...
+        super().__init__(*args, **kwargs)
+        self.fields['chooser'] = forms.ModelChoiceField(
+            initial=destination,
+            queryset=Page.objects.all(),
+            widget=widgets.AdminPageChooser(can_choose_root=True, user_perms='move_to'),
+            label=_("Select a new parent page"),
+        )
+
+
+def before_bulk_move(request, action_type, pages, action_instance):
```
jacobtoppm, line 25: I don't think this should be done via a hook, because we're writing the original view in the first place.

jacobtoppm (changes requested):

wagtail/admin/views/bulk_action.py
```diff
@@ -99,8 +106,14 @@ def get_context_data(self, **kwargs):
             'submit_url': self.request.path + '?' + self.request.META['QUERY_STRING']
```
jacobtoppm, line 106: Probably nicer to avoid string concatenation when possible.

client/src/entrypoints/admin/bulk-actions.js
jacobtoppm: @KIRA009 just tagging you because this might have been overlooked from a while back

wagtail/admin/templates/wagtailadmin/pages/bulk_actions/confirm_bulk_move.html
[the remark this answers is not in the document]
zerolab: showing the number of page would be a neater thing to do

zerolab:
@KIRA009 thank you for this! We can now move on with the smaller PRs

Issue #6848, opened 2021-02-27
Can't specify ordering on construct_explorer_page_queryset hook


amxfonseca:
### Issue Summary

The wagtail page explorer sidebar ignores the ordering of the QuerySet returned by the `construct_explorer_page_queryset` hook.

Since wagtail 2.8 (probably due to the migration of the admin pages api) the ordering of the items for the explorer items is broken. Any custom ordering is ignored and the items are always ordered by `path`.

The problem is caused by the following line: https://github.com/wagtail/wagtail/blob/3c040acc88abd052c9d12c1d3b94a54166e2eb6a/wagtail/admin/api/filters.py#L39

Then `&` removes any existing order and sets it to `path`, because that's the default value that is present on the manager:
https://github.com/wagtail/wagtail/blob/3c040acc88abd052c9d12c1d3b94a54166e2eb6a/wagtail/core/models.py#L676

The order can still be set for the main explorer, only the sidebar explorer is broken since the main page listing does not use the api.

### Technical details

* Python version: N/A
* Django version: N/A
* Wagtail version: Wagtail >= 2.8
* Browser version: N/A

amxfonseca:
I can see some possible quick solutions for this issue:

1. Change the place where we apply  the`queryset = queryset & user_perms.explorable_pages()` operation by doing it before we call the hooks
2. Make sure that the `user_perms.explorable_pages()` doesn't return any order information (`user_perms.explorable_pages().order_by()` restores the old behaviour)
3. Reverse the order of the operands of the `&` operation (`queryset = user_perms.explorable_pages() & queryset`)

The first approach is easier but it introduces some possible issues, because we may end up displaying pages that the user may not be allowed to see.

The second approach keeps the current filtering logic intact, but it requires some extra changes to the ` UserPagePermissionsProxy`

The last approach is also quite simple. I don't know the internal application of the `&` operator, but I assume it is commutative, so it should not impact the final result (when it comes to filtering)

Stormheg:
Fixed in https://github.com/wagtail/wagtail/commit/6e77e0d415a9e127bb3593db9c69bc60b88e06ed (main) and https://github.com/wagtail/wagtail/commit/fa94ffc76b52f35e918c35e222b099a12b6cde08 + parents (stable/2.13.x) :tada:

PR #6962, opened 2021-04-01
Fix ordering behaviour when using construct_explorer_page_queryset hook

amxfonseca:
Fixes https://github.com/wagtail/wagtail/issues/6848

Since wagtail 2.8 the behaviour of doing `queryset.order_by('some_field')` inside a **construct_explorer_page_queryset** hook handler is broken (at least partially, depending if they are returned through the template or API).

This was introduced when migrating the backend API for the page explorer.

At this moment any ordering preference that is set on the queryset is lost. This is caused by the `queryset & user_perms.explorable_pages()` operation on the **ForExplorerFilter** class. Since the `explorable_pages()` returns some default ordering, this always takes precedence over the now modified queryset.

The pragmatic approach here was to simply reverse the operands. This has no effect on the filtering behaviour part of the operation, since the `&` operator is commutative (`A & B <=> B & A`). But now the **order_by** information of the rightmost operator is taken instead. In the situation that no **order_by** is present, then the one from the leftmost operator is still taken, so no breaking changes are being introduced.

This is a bit of a edge case, because sorting on that hook is not a common pattern.  But some users really want different default ordering of the explorer pages depending on the parent page type (for example, blog posts being sorted by publish date, but the other pages being sorted by creation date). This was always possible, but since the recent versions it completely broke. And since the hook takes and returns a queryset, it makes sense that none of the information is not silently discarded (like is currently happening with the **order_by**).

* For front-end changes: Did you test on all of Wagtail’s supported browsers? **N/A**

### DIFF
--- a/wagtail/admin/api/filters.py
+++ b/wagtail/admin/api/filters.py
@@ -36,6 +36,6 @@ class ForExplorerFilter(BaseFilterBackend):
                 queryset = hook(parent_page, queryset, request)
 
             user_perms = UserPagePermissionsProxy(request.user)
-            queryset = queryset & user_perms.explorable_pages()
+            queryset = user_perms.explorable_pages() & queryset
 
         return queryset
--- a/wagtail/admin/tests/api/test_pages.py
+++ b/wagtail/admin/tests/api/test_pages.py
@@ -7,6 +7,7 @@ from django.urls import reverse
 from django.utils import timezone
 
 from wagtail.api.v2.tests.test_pages import TestPageDetail, TestPageListing
+from wagtail.core import hooks
 from wagtail.core.models import Locale, Page
 from wagtail.tests.demosite import models
 from wagtail.tests.testapp.models import SimplePage, StreamPage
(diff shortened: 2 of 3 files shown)


Stormheg:
Thank you for your contribution @amxfonseca and congratulations on your first one to Wagtail👍 

I’ll try to review your pull request in the next couple of days!

Stormheg (approved):
Ordering the sidebar page explorer works like a charm now :tada: 

Thank you for fixing @amxfonseca! :+1: 

image

Stormheg:
Merged in https://github.com/wagtail/wagtail/commit/6e77e0d415a9e127bb3593db9c69bc60b88e06ed (main) and https://github.com/wagtail/wagtail/commit/fa94ffc76b52f35e918c35e222b099a12b6cde08 + parents (stable/2.13.x) :tada:

gasman:
Just to note that I've rolled this back from the 2.13 branch (and moved the release note accordingly) - it's a bit of a borderline case here, but we want to limit the changes that happen between 2.13rc1 and 2.13 final so that it's just fixes to 2.13 functionality.

PR #6966, opened 2021-04-02
Fix issue 6664 - oEmbed cache_age

### LINKED ISSUE #6664: Respect cache_until when retrieving embeds
[The oEmbed spec](https://oembed.com/) defines `cache_age` as 

> The suggested cache lifetime for this resource, in seconds. Consumers may choose to use this value or not.

My initial search hasn't pulled up any details about this being used.  It would be helpful if this was stored on the embed model.  This could be used in conjunction with `last_modified` for automatically refreshing stale content.

Since `last_modified` doesn't necessarily translate to when the data was fetched, there could ideally be an additional field that captures the fetch time.

This could also be combined into one DateTime field `cache_until` which would simplify queries and simply be set to the fetch time plus the cache age.

thenewguy:
Fixes #6664


jaap3:
I have a minor code cleanup PR that would clash with the changes proposed here: #6922

Any chance that mine could get merged before or rolled into this one?

gasman (approved):

gasman:
Merged in 3c6ac4b5026a8cd4ed8d979a701699389c417e7d - thanks @thenewguy!

Issue #5914, opened 2020-03-27
Streamfield issues IE11


lennarddevries:
### Issue Summary

While editing a page in IE11 I'm unable to add blocks to a StreamFields. Clicking the '+' button only triggers focus outline for the block above
### Steps to Reproduce

1. Set up fresh bakerydemo instance
2. Create a new blogpost instance
3. Add a block through the + button

![ezgif com-video-to-gif](https://user-images.githubusercontent.com/7124952/77750214-b4572e00-7023-11ea-963d-8e006e1606d2.gif)

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: **yes**
### Technical details

* Python version: 3.6.0 / 3.6.9 .
* Django version: 2.2.9 / 3.0.4.
* Wagtail version: 2.7.1 / 2.8`.
* Browser version: IE11 (version: 11.379.17763.0)

Stormheg:
I can reproduce this issue.

Here are my findings:

Using the select element tool I cannot hover the add button as the containing element behind it receives focus. This probably explains why clicking the button triggers the focus outline for the block.

<img width="1507" alt="Screenshot 2020-04-14 at 10 49 43" src="https://user-images.githubusercontent.com/13856515/79205136-d6cbb280-7e3d-11ea-87a5-a651e00737dc.png">

When I change the `pointer-events` property on the `.c-sf-add-button--visible` CSS class from `unset` to `visible` the button works again and I can add streamfields. It appears IE11 doesn't recognise `unset` as a valid property value.

<img width="1395" alt="Screenshot 2020-04-14 at 11 00 51" src="https://user-images.githubusercontent.com/13856515/79207281-cec14200-7e40-11ea-8bfd-f2f1420d058c.png">

I am short on time to investigate this issue in more depth but hopefully this is useful.

paul-apacio:
I can second this as an issue.

thibaudcolas:
I stumbled upon this as well and was about to create a new issue before I found this one. I wouldn’t be surprised if this was broken since the new StreamField UI was introduced in Wagtail 2.7, https://docs.wagtail.io/en/stable/releases/2.7.html#improved-streamfield-design.

It does look like the issue is with the elements’ stacking order as mentioned above, so might need to be resolved in the stylesheet for https://github.com/wagtail/react-streamfield, which is where those new styles are defined.

Tested in Wagtail 2.10 (current master) for me.

[...]

PR #6968, opened 2021-04-03
Compile ES6 to ES5 to fix IE11 support

### LINKED ISSUE #6961: IE11 support currently broken for Draftail/telepath
We switched to using the TypeScript compiler for building both JavaScript and TypeScript to make it easy to integrate the two. The problem is TSC is not converting ES6 JavaScript to ES5 JavaScript, as it only compiles TypeScript to ES5.

This means that ES6 JS components (telepath and Draftail) are currently broken on IE11.

kaedroho:
Partially fixes #6961

This mostly fixes IE11 support by compiling all ES6 Javascript in Wagtail itself to ES5 using babel-loader. This is not a complete fix though as  ``telepath.js`` is still giving a syntax error, I think this is because the distributed version of ``telepath-unpack`` is in ES6 and it seems the best way to solve that is adding babel to ``telepath-unpack``.

### DIFF
--- a/client/webpack.config.js
+++ b/client/webpack.config.js
@@ -88,6 +88,7 @@ module.exports = function exports() {
 
   return {
     entry: entry,
+    target: ['web', 'es5'],  // Note: can remove after dropping IE11
     output: {
       path: path.resolve('.'),
       publicPath: '/static/js/'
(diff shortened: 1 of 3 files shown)


thibaudcolas (changes requested):
@kaedroho I don’t really see why this would be needed. tsc will transpile ES6+ JS to ES5 just fine – I can’t see any issues testing Wagtail v2.12 in IE11 as an example. You can see this by inspecting the JS bundles from the project, or trying `tsc` manually.

Trying the current `main` in IE11, the only issue I can see is that [telepath-unpack](https://github.com/wagtail/telepath-unpack/blob/f0d120599698a6edc9f3ce7506da6b13cb4c1099/index.js) throws JS syntax errors due to its use of `class` (among other things). This is to be expected, we don’t convert our dependencies, and generally the expectation is that all dependencies should be published as either ES5 CommonJS, or ES5 with ES modules.

gasman:
Have released an ES5-transpiled version of telepath-unpack as 0.0.3, and version bumped to that in 10d106ab43bd3fb67a1cd03237f9c246bdc1e2d7 - let me know if that's done the job...

kaedroho:
Nice one @gasman! Will check tomorrow morning

gasman:
Testing on IE11 against current main (with the telepath-unpack 0.0.3 upgrade), the only regression I can see is that adding a new rich text block to a streamfield results in a Draftail crash:

Screenshot 2021-04-19 at 16 47 16

(This itself is fairly well hidden due to #5914 making the keyboard the only way to add blocks on IE11...)

Applying this PR doesn't fix that crash, so on the available evidence @thibaudcolas is correct that the telepath-unpack fix was the necessary one here.

gasman:
With the above error fixed in #7033, will close this on the assumption that recompilation isn't needed.

PR #6970, opened 2021-04-04
Important fix to index.md

Pleskan:
Add forgotten build static assets step

### DIFF
--- a/docs/getting_started/index.md
+++ b/docs/getting_started/index.md
@@ -40,6 +40,7 @@ Inside your `mysite` folder, run the setup steps necessary for any Django projec
 $ pip install -r requirements.txt
 $ ./manage.py migrate
 $ ./manage.py createsuperuser
+$ ./manage.py collectstatic
 $ ./manage.py runserver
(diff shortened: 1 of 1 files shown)


tomdyson:
Thanks for this, @Pleskan, but `./manage.py collectstatic` shouldn't be necessary for local development, which is what these instructions are designed for.

Pleskan:
> Thanks for this, @Pleskan, but `./manage.py collectstatic` shouldn't be necessary for local development, which is what these instructions are designed for.

OK, then would you like to make a remark that for production deployment assets building is mandatory step?

gasman:
Hi @Pleskan - the appropriate place in the docs for advice on deployment is https://docs.wagtail.io/en/stable/advanced_topics/deploying.html , and we do link to Django's documentation on handling static files there.

It's not really possible for us to give step-by-step instructions for production deployments because every setup is different, but if you can see a better way of mentioning the need for `collectstatic`, we'd be happy to consider a PR for that.

PR #6971, opened 2021-04-06
Commenting controls 

kaedroho:
This PR implements the commenting controls in the header:

image

- Clicking the comment icon shows/hides comments
- When comments are visible, the "Comment notifications" switch appears (does nothing at the moment)
- Number of comments stays up to date


client/src/components/CommentApp/selectors/index.ts
```diff
@@ -47,3 +47,9 @@ export const selectIsDirty = createSelector(
       return Array.from(comment.replies.values()).some(reply => reply.deleted || reply.originalText !== reply.text);
     });
   });
+
+export const selectCommentCount = (state: State) => (
```
jacobtoppm, line 51: This should probably exclude resolved as well, I think?

jacobtoppm (changes requested):
One minor change to account for resolved, but otherwise LGTM!

client/src/components/CommentApp/selectors/index.ts
kaedroho: Fixed, thanks!

PR #6972, opened 2021-04-06
Pin docutils to 0.16

kaedroho:


zerolab:
@kaedroho https://github.com/sphinx-doc/sphinx/issues/9049 recommends pinning `docutils==0.16`

kaedroho:
Nice one @zerolab, that's fixed it!

PR #6974, opened 2021-04-06
Add a switch widget

kaedroho:
This PR adds a new switch widget for boolean fields, and makes use of it for notification preferences.

image

image

Design by @benenright

### DIFF
--- a/client/scss/components/_switch.scss
+++ b/client/scss/components/_switch.scss
@@ -1,0 +1,84 @@
+$switch-width: 40px;
+$switch-height: 20px;
+$switch-border: 2px;
+$switch-outline: 3px;
+
+$switch-border-radius: ($switch-height + $switch-border * 2) / 2;
+$switch-outline-radius: $switch-border-radius + $switch-outline;
+
+// All the greys in Wagtail are really dark or really bright
+$switch-color-middle-grey: #777;
+
+.switch {
+    display: inline-flex;
+    align-items: center;
+    margin: 5px 0;
+
+    // Disable forms styling that's applied to the <label> tag
+    width: unset;
+    float: unset;
+
+    &__toggle {
+        position: relative;
+        cursor: pointer;
+
+        &::before,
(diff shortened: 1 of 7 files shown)


client/scss/components/_switch.scss
```diff
@@ -0,0 +1,83 @@
+$switch-width: 40px;
+$switch-height: 20px;
+$switch-border: 2px;
+$switch-outline: 3px;
+
+$switch-border-radius: ($switch-height + $switch-border * 2) / 2;
+$switch-outline-radius: $switch-border-radius + $switch-outline;
+
+// All the greys in Wagtail are really dark or really bright
+$switch-color-middle-grey: #777777;
```
jacobtoppm, line 10: This feels like it could be useful more generally - should we add it to global grey colour variables instead?

wagtail/users/forms.py
```diff
@@ -385,7 +385,12 @@ def __init__(self, *args, **kwargs):
 
     class Meta:
         model = UserProfile
-        fields = ("submitted_notifications", "approved_notifications", "rejected_notifications")
+        fields = ['submitted_notifications', 'approved_notifications', 'rejected_notifications']
```
jacobtoppm, line 388: Why has this line changed from a tuple to a list?

jacobtoppm (approved):
Two minor comments, but otherwise looks good!

wagtail/users/forms.py
kaedroho: Because it should be a list!

jacobtoppm: Haha, fair enough!

client/scss/components/_switch.scss
kaedroho: Would be nice to, not sure what the most consistent way to do this would be though, the current greys are numbered 1 - 5 so I guess we would either need to shift some of them or come up with a new way to label them

PR #6976, opened 2021-04-07
Include submitter in workflow submitted email notification

gasman:
Add back the name of the submitter to the "submitted for moderation" email - this was present in the pre-workflow implementation.

### DIFF
--- a/wagtail/admin/mail.py
+++ b/wagtail/admin/mail.py
@@ -331,6 +331,11 @@ class WorkflowStateSubmissionEmailNotifier(BaseWorkflowStateEmailNotifier):
 
         return recipients
 
+    def get_context(self, workflow_state, **kwargs):
+        context = super().get_context(workflow_state, **kwargs)
+        context['requested_by'] = workflow_state.requested_by
+        return context
+
 
 class BaseGroupApprovalTaskStateEmailNotifier(EmailNotificationMixin, Notifier):
     """A base notifier to send email updates for GroupApprovalTask events"""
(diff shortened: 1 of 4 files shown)


zerolab (approved):
LGTM.
Does this need a test since the requester was in the pre-workflow implementation, thus technically a regression?

jacobtoppm (approved):
This looks good to me! I think at some point we could move some of the templatetags/filters out so we don't need to redo them each time when sending bulk emails, but that's something for the future

KalobTaulien (approved):

gasman:
Actually, when trying to add a test here it seems to be more complicated than I thought... in [the `submit` method of TestSubmitToWorkflow](https://github.com/wagtail/wagtail/blob/2fe56837449a17e11ef6e517307363f9f7a4713b/wagtail/admin/tests/test_workflows.py#L709), it seems that three emails get sent:

```
moderator@email.com - The page "Hello world! (simple page)" has been submitted for approval in moderation stage "test_task_1"
superuser@email.com - The page "Hello world! (simple page)" has been submitted for approval in moderation stage "test_task_1"
superuser@email.com - The page "Hello world! (simple page)" has been submitted to workflow "test_workflow"
```

I guess the first two only happen for multi-stage workflows, and for those, it's not entirely clear who should count as the submitter on tasks after the first - is it the editor who initiated the workflow, or the moderator who took it to the next task? Since that's a more niche setup, I'll go ahead with this as is (then at least it's not a regression in the default setup).

gasman:
Merged in 255a1e8641e083f27e5d2686b1329b1f97e3cbb7

PR #6978, opened 2021-04-07
Commenting: Tweak the sizes to more closely match the design

kaedroho:


jacobtoppm (changes requested):
Generally looking great, but a few little visual glitches/differences from designs.
- On Chrome, creating a comment, there is an odd straight line above the textarea that then diverges at the corners
- The designs suggest you should be able to get at least 3 lines of text without a scrollbar, but that's not the case here
- The date is further below the username than the designs - could this be moved up?

kaedroho:
@jacobtoppm Thanks! I've fixed those issues now, please could you review the last commit?

client/src/components/CommentApp/components/Comment/index.tsx
```diff
@@ -281,11 +274,11 @@ export default class CommentComponent extends React.Component<CommentProps> {
       <>
         <CommentHeader commentReply={comment} store={store} strings={strings} />
         <form onSubmit={onSave}>
-          <textarea
+          <TextArea
```
jacobtoppm, line 277: You've doubled up on the focusing logic here - we already focus in `ComponentDidMount`. Bearing in mind we're going to need to move keyboard focus to other areas of the comment in the accessibility pass (ie the whole comment box when viewing an existing comment), I think it might make more sense to handle focus at the comment level - if you see my ref logic in https://github.com/wagtail/wagtail/pull/6983, I use a ref for this - and instead of moving focus logic into the textarea, get it to forward a ref to the comment that can be used to focus. What do you think? At the very least we should remove the double focusing

jacobtoppm (changes requested):
Visually looks great! Just an issue with doubling up on the focusing logic, but otherwise LGTM

client/src/components/CommentApp/components/Comment/index.tsx
kaedroho: Oops, thanks for spotting that!

I copied this from Wagtail Localize for the auto-resize behaviour, just thought we might as well reuse the focusing logic too! happy to completely remove this if it makes things easier for #6983?

PR #6980, opened 2021-04-07
Apply chevron padding fix to all applicable `select` elements

Scotchester:
Most `select` elements in the admin do not currently have adequate padding on the right-hand side to avoid long text colliding with the down-pointing chevron icon. A fix was previously added for this, but it was too narrowly scoped. This PR should, I believe, apply it to all cases where the chevron appears.

### Before:

Long select option text colliding with chevron icon

### After:

The same long text is cut off before colliding with the icon

---

Also tagging for review: @rinti

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -34,6 +34,7 @@ Changelog
  * Fix: Validate host/scheme of return URLs on password authentication forms (Susan Dreher)
  * Fix: Reordering a page now includes the correct user in the audit log (Storm Heg)
  * Fix: Reverse migration errors in images and documents (Mike Brown)
+ * Fix: Apply enough chevron padding to all applicable select elements (Scott Cranfill)
 
 
 2.12.3 (05.03.2021)
--- a/client/scss/components/_forms.scss
+++ b/client/scss/components/_forms.scss
@@ -107,11 +107,11 @@ select::-ms-expand {
             display: none;
         }
     }
-}
 
-// the site setting dropdown is auto width, so the chevron will overlap with text if not padded
-.choice_field .setting-site-switch-form .input select {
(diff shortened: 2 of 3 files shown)


Scotchester:
Weird sidenote: I am at a complete loss to explain the line-height discrepancy between my Before and After shots that causes the letter descenders to be cut off in the After shot.

Before is on a Gitpod environment, which shows a default, user-agent supplied line-height of 22px. After is me running the Wagtail Bakery demo locally, which shows a default, user-agent supplied line-height of 18px. Both viewed in Google Chrome 89.0.4389.114. The Squash Labs environment that was spun up for this PR looks like the Gitpod environment.

Scotchester:
> Weird sidenote: I am at a complete loss to explain the line-height discrepancy between my Before and After shots that causes the letter descenders to be cut off in the After shot.
> 
> Before is on a Gitpod environment, which shows a default, user-agent supplied line-height of 22px. After is me running the Wagtail Bakery demo locally, which shows a default, user-agent supplied line-height of 18px. Both viewed in Google Chrome 89.0.4389.114. The Squash Labs environment that was spun up for this PR looks like the Gitpod environment.

I guess it was some browser caching issue. After hard reloading the Gitpod and Squash environments, all show 18px now. Yay for consistency, but boo for descenders getting cut off everywhere :(

This appears to have been caused by a change to Chrome's interpretation of `line-height` set to the `normal` keyword. The line height is adequate in Fireox:

The text is not cut off in Firefox

thibaudcolas (approved):
Well – this looks pretty spot on! I’m not quite sure why we didn’t go with this from the get-go 😛 .

I can confirm I had the same cropping of descenders in Chrome.

PR #6983, opened 2021-04-08
Feature/inline comment focus on add

jacobtoppm:
Ensures newly added comments in Draftail are focused and have keyboard focus moved to the text input

Focusing logic will be extended in accessibility pass

### DIFF
--- a/client/src/components/Draftail/CommentableEditor/CommentableEditor.tsx
+++ b/client/src/components/Draftail/CommentableEditor/CommentableEditor.tsx
@@ -449,8 +449,6 @@ function CommentableEditor({
     comments,
   ]);
 
-  const controls = useMemo(() => (enabled ? [CommentControl] : []), [enabled, CommentControl]);
-
   const commentStyles: Array<InlineStyle> = useMemo(
     () =>
       ids.map((id) => ({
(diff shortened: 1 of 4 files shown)


client/src/components/Draftail/CommentableEditor/CommentableEditor.tsx
```diff
@@ -445,6 +449,8 @@ function CommentableEditor({
     comments,
   ]);
 
+  const controls = useMemo(() => (enabled ? [CommentControl] : []), [enabled, CommentControl]);
```
kaedroho, line 452: Why do we need a `useMemo` here? This seems like it'll be pretty fast

kaedroho (approved):

client/src/components/Draftail/CommentableEditor/CommentableEditor.tsx
jacobtoppm: Yep, this can be probably be removed actually. I was testing whether using an inline array here caused some of the event listener problems, but I don't think this did much!

kaedroho: Feel free to merge once this is done

PR #6984, opened 2021-04-08
Only show comments for current tab

kaedroho:


client/src/components/CommentApp/utils/layout.ts
```diff
@@ -103,79 +105,97 @@ export class LayoutController {
...
 
-            // If this block contains the focused comment, position it so
-            // the focused comment is in it's desired position
+    // Get location of pinned comment
+    const pinnedCommentPosition = this.pinnedComment ?
+      this.commentDesiredPositions.get(this.pinnedComment) : undefined;
+    const pinnedCommentTab = this.pinnedComment ?
+      this.commentTabs.get(this.pinnedComment) : undefined;
+
+    // For each tab, resolve positions of all the comments
```
jacobtoppm, line 122: Do we need to resolve positions on all tabs? Could we do just the current?

jacobtoppm (commented):
One quick question, but also looks like some of the edit handler tests are failing right now

client/src/components/CommentApp/utils/layout.ts
kaedroho: Helps with initial page load to resolve them all, I didn't think it was worth optimising this much since it's pretty fast

kaedroho:
Tests should be fixed now!

Issue #6836, opened 2021-02-26
Migration fails on MySQL when upgrading to Wagtail>=2.12rc1


stvnw:
### Issue Summary

The [wagtailembeds.0008_allow_long_urls](https://github.com/wagtail/wagtail/blob/master/wagtail/embeds/migrations/0008_allow_long_urls.py) migration added in 2.12rc1 (https://github.com/wagtail/wagtail/commit/32ade8672fe32223b6ebdcb2255bf4de16908f97) results in the following error on MySQL 8.0 (and presumably also older versions):

    MySQLdb._exceptions.OperationalError:
    (1101, "BLOB, TEXT, GEOMETRY or JSON column 'thumbnail_url' can't have a default value")

This is expected behaviour. According to https://dev.mysql.com/doc/refman/8.0/en/blob.html:

    BLOB and TEXT columns cannot have DEFAULT values.

### Steps to Reproduce
_Update: Migration from 2.11 to 2.12 is not needed; can be reproduced more simply in steps 1–5 using 2.12rc1<=wagtail<=2.12.2_

This issue can be reproduced on a fresh Wagtail project as follows:

1. Install Wagtail ~~< 2.12~~ >= 2.12rc1 with ~~`pip install 'wagtail<2.12'`~~ `pip install 'wagtail>=2.12rc1<=2.12.2'` 
2. Start a new project `wagtail start testproject`
3. Configure the `DATABASES` setting to use a MySQL database
```
DATABASES = {
    'default': {
        'ENGINE': 'django.db.backends.mysql',
        'NAME': 'testproject',
        'USER': 'testuser',
        'PASSWORD': 'testpass'
    }
}
```
4. Prepare the database
```
mysql> create database testproject;
mysql> create user 'testuser'@'localhost' identified by 'testpass';
mysql> grant all privileges on testproject.* to 'testuser'@'localhost';
```
5. Run initial migrations `./manage.py migrate`
~~6. Upgrade to Wagtail 2.12rc1 or greater with, e.g. `pip install --upgrade wagtail==2.12.2`~~
~~7. Re-run migrations `./manage.py migrate`~~

### Additional Notes

Prior to MySQL 8.0.13, BLOB and TEXT could not be assigned DEFAULT values. As of 8.0.13, they can be assigned defaults, but only in the form of expressions as noted in https://dev.mysql.com/doc/refman/8.0/en/data-type-defaults.html#data-type-defaults-explicit

### Technical details
* Python 3.8.3
[...]

PR #6985, opened 2021-04-08
Formalise support for Django 3.2

### LINKED ISSUE #6981: Formalise Django 3.2 support
Django 3.2 has just been [released](https://www.djangoproject.com/weblog/2021/apr/06/django-32-released/)!
Although Wagtail seems to support it (I've seen tests are run against Django master) it is excluded in `setup.py` (`"Django>=2.2,<3.2"`).

It is an issue because the [Poetry](https://python-poetry.org/) package manager cannot install packages declared as being incompatible ([source](https://github.com/python-poetry/poetry/issues/697#issuecomment-470431668)). That means that people currently using Wagtail and Poetry (me for example :D) cannot upgrade to Django 3.2 (not easily at least).

gasman:
Fixes #6981


gasman:
The remaining failure on MySQL is caused by this commit: https://github.com/django/django/commit/779e615e362108862f1681f965ee9e4f1d0ae6d2. The docs for that commit don't indicate that this is a breaking change, so it's probably an unintended regression in Django.

(#6836 also makes it difficult to test against current MySQL versions - it may be possible to work around that by retrospectively changing the initial migration in wagtail.embeds so that thumbnail_url is a TextField already, but that needs further investigation.)

gasman:
Updated with fix from #6999 and workaround for the order_by/update issue from https://github.com/django/django/commit/779e615e362108862f1681f965ee9e4f1d0ae6d2 - will investigate the latter further and file an issue with Django if necessary.

gasman:
MySQL order_by/update bug now reported to Django in https://code.djangoproject.com/ticket/32645.

jpmckinney:
Some warnings that'd be nice to avoid in `wagtail/core/signals.py`:

> RemovedInDjango40Warning: The providing_args argument is deprecated. As it is purely documentational, it has no replacement. If you rely on this argument as documentation, you can move the text to a code comment or docstring.

Basically, `providing_args=[…]` can be removed with no consequence.

gasman:
@jpmckinney Good point... removed those now, cheers!

auvipy (approved):

gasman:
Merged in f8299885aa916becb9109c881f3ab28408af794e + parents.

PR #6986, opened 2021-04-08
Dynamically resize fields when comments are enabled

kaedroho:


jacobtoppm (changes requested):
The add buttons should probably remain centered over the streamfield at smaller screen widths, but right now they stay centered on the original width instead

Otherwise, this looks good! Though I kind of feel all fields should scale the same way in commenting mode... Will this empty bar width be reduced when the comment width reducing PR goes in?

kaedroho:
Fixed now and ready for another review!

jacobtoppm (changes requested):
Looking good with streamfield! However, it looks like this now makes ordinary fields smaller than streamfield at desktop widths with comments enabled, which looks odd

(also, tests still need a fix)

PR #6989, opened 2021-04-09
Split out CommentableEditor utility functions, and add basic tests

jacobtoppm:


client/src/components/Draftail/CommentableEditor/CommentableEditor.tsx
```diff
@@ -291,6 +291,39 @@ function updateCommentPositions({ editorState, comments, commentApp }:
   });
 }
 
+export function findLeastCommonCommentId(block: ContentBlock, offset: number) {
```
kaedroho, line 294: I think we could do with a comment here to describe what this function is meant to do

client/src/components/Draftail/CommentableEditor/CommentableEditor.tsx
```diff
@@ -507,30 +540,10 @@ function CommentableEditor({
...
-          hasUpdated = true;
-        });
-      } catch (err) {
-        console.error(`Error loading comment position for comment ${comment.localId}`);
-        console.error(err);
-      }
-    });
-    if (hasUpdated) {
-      setEditorState(forceResetEditorState(editorState, contentState));
+    const contentState = editorState.getCurrentContent();
```
kaedroho, line 543: This might be why I get Draftail messages every time I type a character in a comment. This code is probably being run for every comment change, not just ones that are relevant for Draftail

client/src/components/Draftail/CommentableEditor/CommentableEditor.tsx
```diff
@@ -394,6 +398,35 @@ function forceResetEditorState(editorState: EditorState, replacementContent: Con
...
 
+export function addCommentsToEditor(
+  contentState: ContentState,
+  comments: Comment[],
+  commentApp: CommentApp,
+  getAnnotation: () => Annotation
+) {
+  let newContentState = contentState;
+  comments.filter(comment => !comment.annotation).forEach((comment) => {
+    commentApp.updateAnnotation(getAnnotation(), comment.localId);
```
kaedroho, line 409: Would firing off lots of redux actions on initial load cause a lot of re-renders?

kaedroho (approved):
Just some minor comments about the existing code, but this looks like a good change

client/src/components/Draftail/CommentableEditor/CommentableEditor.tsx
jacobtoppm: This particular bit should only run:
- For comments with a relevant draftail contentpath
- Once per comment, as it runs for comments with no annotation and then adds an annotation
So I think it's unlikely to be the culprit

That said, I think if you're changing the comment text for a draftail comment, the other `useEffect` hook could potentially trigger to rerender Draftail, which we could probably fix by changing the id equality check to not work by reference only, so that could be the culprit

client/src/components/Draftail/CommentableEditor/CommentableEditor.tsx
jacobtoppm: Yeah, this is something we could optimize

client/src/components/Draftail/CommentableEditor/CommentableEditor.tsx
jacobtoppm: Good point, thanks!

Issue #6991, opened 2021-04-10
Improving signals examples


damascene:
### Documentation on signals needs improvements 

https://docs.wagtail.io/en/latest/reference/signals.html#page-published

### Details

I would like to see clear example about how to do test. The current example have multiple issues:

1. Example is using an external service that it requires API access from an external provider account and setup to work
2. It doesn't specify where code should be inserted

allcaps:
This is a reference section. It is about technical details, not a tutorial or how-to. The introduction to signals and the example should go into a separate how-to listen to signals page.

Signals are a Django concept. We could add a link to the Django documentation to learn about the basics.

PR #6992, opened 2021-04-10
Created example for creating signals

damascene:
https://github.com/wagtail/wagtail/issues/6991

### DIFF
--- a/docs/advanced_topics/adding_signals.rst
+++ b/docs/advanced_topics/adding_signals.rst
@@ -1,0 +1,68 @@
+==================
+How to use Signals
+==================
+
+Wagtail's :ref:`page-revision-model-ref` and :ref:`page-model-ref` implement
+:doc:`Signals <topics/signals>` from ``django.dispatch``.
+Signals are useful for creating side-effects from page publish/unpublish events.
+
+***************
+Example for using signals
+***************
+
+To use signals you need to create/modify three files inside your app:
+
+1. singals.py
+2. apps.py
+3. __init__.py
+
+Here is a short example that prints "Hello world!" when a page is published.
+
+We have:
+
+- App named: ``blog``
+- Model named: ``BlogPage``
+
+We configure the files 
+
+`singals.py`:
+===============
+
+.. code-block:: python
+
+    from wagtail.core.signals import page_published
+    from blog.models import BlogPage
+    
+    # Do something clever for each model type
+    def receiver(sender, **kwargs):
+        # Do something with blog posts
+        print("Hello world!")
+    
+    # Register listeners for each page model class
+    page_published.connect(receiver, sender=BlogPage)
+    
+
+
+`apps.py`:
+===============
+
+.. code-block:: python
+
+    from django.apps import AppConfig
+    
+    
+    class BlogConfig(AppConfig):
+        name = 'blog'
+    
+        def ready(self):
+            import blog.signals  # noqa
+    
+`__init__.py`:
+===============
+
+.. code-block:: python
+
+    default_app_config = 'blog.apps.BlogConfig'
+
+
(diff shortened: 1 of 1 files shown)


docs/advanced_topics/adding_signals.rst
```diff
@@ -0,0 +1,68 @@
+==================
+How to use Signals
+==================
+
```
allcaps, line 4: Can we start with a line that points the user to the Django documentation on signals?
https://docs.djangoproject.com/en/3.1/topics/signals/

docs/advanced_topics/adding_signals.rst
```diff
@@ -0,0 +1,68 @@
...
+        print("Hello world!")
+    
+    # Register listeners for each page model class
+    page_published.connect(receiver, sender=BlogPage)
+    
+
+
+`apps.py`:
+===============
+
```
allcaps, line 48: Signals might be wired in other locations. Using the app ready is just a convention. We should make that clear.

docs/advanced_topics/adding_signals.rst
```diff
@@ -0,0 +1,68 @@
...
+    
+        def ready(self):
+            import blog.signals  # noqa
+    
+`__init__.py`:
+===============
+
+.. code-block:: python
+
+    default_app_config = 'blog.apps.BlogConfig'
```
allcaps, line 65: This is not best practice.
https://docs.djangoproject.com/en/3.2/ref/applications/#configuring-applications last p
> Changed in Django 3.2:
In previous versions, a default_app_config variable in the application module was used to identify the default application configuration class.

allcaps (commented):
Thank you for bringing up this topic and improving the docs.

Since Signals is a Django concept, we should point to their documentation a bit more.
- It will be up to date
- It will explain better, with more detail

I do like to have a Signals section, but I think this should focus on Wagtail specifics and for the rest point to Django docs.

Also touching this topic is the difference between Wagtail hooks and Signals.

Confusing to a lot of integrators is the difference between signals and hooks.
- Signals should be used on db operations. The also fire when an object is modified via the ORM.
- Hooks are limited, to their scope. For example only fire when a object is saved via the Wagtail admin.

damascene:
@allcaps Thank you for reviewing this. Please do the required changes if you see fit. I'm not very knowledgeable on the subject. I just wanted a way to make signals works on Wagtail after searching and asking for long time. When I got it to work I wanted to share this short How to with the community. So please feel free to add to it.

lb-:
@damascene thanks so much for submitting this based on what you have learnt, it has been some time and no one else has picked up these changes.

As you noted, you may not be confident to prepare this PR to be ready for review.

I will close this for now and may I suggest you write your learnings up as a blog post, this is a great way to be involved in the community, share what you have learned and maybe the blog post can be added to the This week in Wagtail newsletter or to https://github.com/springload/awesome-wagtail or https://docs.wagtail.org/en/stable/advanced_topics/third_party_tutorials.html

I hope this does not deter you from being involved.

damascene:
Thanks. For some reason I didn't see the comments on my PR.

lb-:
Not sure i follow, your response above implied that you saw the review comments and were not able to continue with this PR yourself.

> @allcaps Thank you for reviewing this. Please do the required changes if you see fit. I'm not very knowledgeable on the subject. I just wanted a way to make signals works on Wagtail after searching and asking for long time. When I got it to work I wanted to share this short How to with the community. So please feel free to add to it.

See the comment above from Coen https://github.com/wagtail/wagtail/pull/6992#pullrequestreview-633029394

damascene:
My bad. almost a year have passed and I've acquired some more info about the subject. I might reopen it in the feature or write a blog page as you have suggested. 
Thank you

lb-:
awesome, sometimes a blog post is a great way to start, it gives you a chance to condense your thoughts into something clear and from there  it can be shared and you can get feedback.

Many of the 'how to' sections in the Wagtail documentation started out as blog posts.

PR #6996, opened 2021-04-12
Update links to Diátaxis framework

evildmp:
Information about the documentation system is now maintained and updated
at https://diataxis.fr.


kaedroho:
Thanks, @evildmp. The site looks great by the way!

Just noticed Wagtail isn't listed on https://diataxis.fr/adoption/

We also use it on:
https://www.wagtail-localize.org/
https://willow.readthedocs.io/en/latest/index.html

PR #7001, opened 2021-04-13
Commenting: refactoring

kaedroho:
See commit log for details. Main changes are:

 - Moved ``initComments`` to run after the form is rendered
 - Moved field level annotations to the ``FieldPanel``/``FieldBlock`` level rather than on widgets

### DIFF
--- a/client/src/components/CommentApp/comments.js
+++ b/client/src/components/CommentApp/comments.js
@@ -1,29 +1,35 @@
 import { initCommentApp } from './main';
 import { STRINGS } from '../../config/wagtailConfig';
 
-function initComments() {
+function initComments(formElem) {
   window.commentApp = initCommentApp();
-  document.addEventListener('DOMContentLoaded', () => {
-    const commentsElement = document.getElementById('comments');
-    const commentsOutputElement = document.getElementById('comments-output');
-    const dataElement = document.getElementById('comments-data');
-    if (!commentsElement || !commentsOutputElement || !dataElement) {
-      throw new Error('Comments app failed to initialise. Missing HTML element');
-    }
-    const data = JSON.parse(dataElement.textContent);
-    window.commentApp.renderApp(
-      commentsElement, commentsOutputElement, data.user, data.comments, new Map(Object.entries(data.authors)), STRINGS
-    );
-  });
-}
 
-function attachTabNav(tabNavElem) {
-  // Attaches the commenting app to the given tab navigation element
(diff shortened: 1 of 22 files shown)


jacobtoppm:
@kaedroho I'm a bit confused why we're changing to a callback approach here, which seems to be causing a few problems in the later commits. Why not just initialise the commentapp as before and add the form element reliant parts to the existing dom loaded callback?

kaedroho:
My aim there was to reduce the amount of magic going on and things happening on module load. But yeah, this does make the draftail initialisation a little awkward (I didn't know how deeply draftail is coupled with the internals of comment app until now!).

I'm thinking of changing my approach with this, instead of initialising the comment app in ``edit.html``, instead move it to being initialised in the entrypoint, but use a javascript hook class to remove the magic initialisation. I think we should avoid having two separate APis for this, so maybe we should make ``initComments`` return an API for the initialised comments system, and make that be the thing that is accessed from ``window.comments``?

jacobtoppm:
> My aim there was to reduce the amount of magic going on and things happening on module load. But yeah, this does make the draftail initialisation a little awkward (I didn't know how deeply draftail is coupled with the internals of comment app until now!).
> 
> I'm thinking of changing my approach with this, instead of initialising the comment app in `edit.html`, instead move it to being initialised in the entry point again, but use a javascript hook class so it's more like a component. I think we should avoid exposing two separate APis for it, so maybe we should make `initComments` return the public API for the initialised comments system, and make that be the thing that is accessed from `window.comments`?
> 
> So `comments.js` would be something like:
> 
> ```js
> import initComments from 'CommentApp';
> 
> (() => {
>     const element = document.querySelector('.js-comments-app');
>     if (element) {
>         window.comments = initComments(element);
>     }
> })();
> ```
> 
> I'd also move all the logic for handling the comments toggles into `comments.js`, like above
> 
> What do you think?

Yeah, this is something I grappled with before, and why I split the commentApp init into the initial (redux store etc only, that anything else like Draftail/widgets could then use without relying on callbacks) and rendering/loading initial data phases.

I'm not completely sure I get the component approach here - surely this would still need the callbacks for widgets to work?

I think the idea of simplifying the API is good, but tbh there's not much that's not already on commentApp, so we could just reduce down to that...
Let's talk about this in standup tomorrow!

kaedroho:
Superseded by https://github.com/wagtail/wagtail/pull/7005

PR #7002, opened 2021-04-13
Use django.shortcuts.resolve_url() instead of reverse() for pageurl fallbacks

ababic:
`reverse()` will only work for named Django urls (anything else fails with a `django.urls.exceptions.NoReverseMatch` exception)

I think it's quite common to just want to specify a fallback string (e.g. `"/"` or `"."`) that will be taken as is.

`django.shortcuts.resolve_url()` is what gives `redirect()` it's nice, non-fussy logic, which can be reused here, allowing the fallback to be any of the following:
1. A model instance with a `get_absolute_url()` method
2. A named url pattern string
3. A string containing `"/"` or `"."`
4. A Promise that evaluates to a string meeting the requirements of 2 or 3

### DIFF
--- a/wagtail/core/templatetags/wagtailcore_tags.py
+++ b/wagtail/core/templatetags/wagtailcore_tags.py
@@ -1,5 +1,5 @@
 from django import template
-from django.shortcuts import reverse
+from django.shortcuts import resolve_url
 from django.template.defaulttags import token_kwargs
 from django.template.loader import render_to_string
 from django.utils.encoding import force_str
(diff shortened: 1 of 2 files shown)


gasman:
Seems like a good idea to me! Would be good to have some extra tests alongside https://github.com/wagtail/wagtail/blob/4f8ef843d03e6a081e084faeb089abb84f18aab0/wagtail/core/tests/tests.py#L24-L32 to cover the new possible fallback types though.

ababic:
@gasman Tests now in place, sir.

gasman:
Merged in 9630967e0e3de3469fd858d2cf6fd656be7c491e

PR #7003, opened 2021-04-13
Make wagtail admin icon visable

patphongs:
When we upgraded to Wagtail version 2.11, we noticed that the Wagtail admin icon was missing. I saw that there was a fix in the [2.10 release version](https://github.com/wagtail/wagtail/commit/884c78aa530539d884b5c67aae9f96e8e2cd919c) but then noticed that the fix had the SVG hidden by default. When I switch aria-hidden to false, the icon appears as expected. 

No additional tests needed as this is a cosmetic change that I'm requesting. Thank you!

## Before
Screen Shot 2021-04-13 at 2 43 57 PM

## After
Screen Shot 2021-04-13 at 2 44 55 PM

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/userbar/base.html
+++ b/wagtail/admin/templates/wagtailadmin/userbar/base.html
@@ -18,7 +18,7 @@
                         </defs>
                     </svg>
                   </div>
-                  <svg class="wagtail-userbar-icon" aria-hidden="true" focusable="false">
+                  <svg class="wagtail-userbar-icon" aria-hidden="false" focusable="false">
                     <use href="#icon-wagtail-icon"></use>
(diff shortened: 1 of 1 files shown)


thibaudcolas (commented):
Hi @patphongs, thank you for looking into this. `aria-hidden="true"` is meant to hide the icon for screen reader users, it should have no impact whatsoever for sighted users.

Based on your report, it looks like for some reason your site might have styles that hide elements set with `aria-hidden="true"`? If so I think you should consider removing those styles – if an element is meant to be hidden for both sighted users and screen reader users, the `hidden` attribute is much more appropriate than `aria-hidden="true"` (and doesn’t require any additional CSS).

PR #7005, opened 2021-04-14
Add comments on FieldBlock and FieldPanel instead of widgets

kaedroho:


client/src/components/CommentApp/comments.js
```diff
@@ -14,6 +14,9 @@ function initComments() {
     window.commentApp.renderApp(
       commentsElement, commentsOutputElement, data.user, data.comments, new Map(Object.entries(data.authors)), STRINGS
     );
+
+    // eslint-disable-next-line @typescript-eslint/no-use-before-define
```
jacobtoppm, line 18: Is there a reason we're disabling this instead of just defining it first?

client/src/components/CommentApp/comments.js
```diff
@@ -14,6 +14,9 @@ function initComments() {
...
+    document.querySelectorAll('[data-comment-add]').forEach(initAddCommentButton);
```
jacobtoppm, line 19: If we're using this attribute to initialise field level widgets, we should remove it from the draftail control and use a different data-attribute to ensure clicking it doesn't remove focus.

jacobtoppm (approved):
One little tweak, but then good to go!

client/src/components/CommentApp/comments.js
kaedroho: The next PR moves this down, I thought I'd just avoid adding too much noise for now

jacobtoppm: Great, just wanted to check.

client/src/components/CommentApp/comments.js
kaedroho: Good spot! Changed it here

PR #7014, opened 2021-04-16
Commenting: Field level commenting fixes

kaedroho:


jacobtoppm (changes requested):
Works for pages with existing comments, but if you open the edit view for a page with no comments, there are no field level comment buttons

kaedroho:
I've copied in some code that puts the add comment button on field comment panels.

Note, it needs to be moved out of ``field.html`` for these as they sometimes have a grey background (eg the title) and the button needs to appear outside of this

PR #7020, opened 2021-04-16
Feature/commenting aria fixes

jacobtoppm:
Fixes a few errors flagged by aXe, and adds aria-describedby attributes to textboxes when we will focus them directly

### DIFF
--- a/client/src/entrypoints/admin/core.js
+++ b/client/src/entrypoints/admin/core.js
@@ -299,8 +299,8 @@ $(() => {
   const showTab = (tabButtonElem) => {
     $(tabButtonElem).tab('show');
 
-    // Update data-current-tab attribe on the .tab-nav element
-    const tabNavElem = tabButtonElem.closest('.tab-nav');
+    // Update data-current-tab attribute on the [data-tab-nav] element
+    const tabNavElem = tabButtonElem.closest('[data-tab-nav]');
     tabNavElem.dataset.currentTab = tabButtonElem.dataset.tab;
 
     // Trigger switch event
@@ -313,7 +313,7 @@ $(() => {
   }
 
   // eslint-disable-next-line func-names
-  $(document).on('click', '.tab-nav a', function (e) {
+  $(document).on('click', '[data-tab-nav] a', function (e) {
     e.preventDefault();
     showTab(this);
     window.history.replaceState(null, null, $(this).attr('href'));
@@ -322,7 +322,7 @@ $(() => {
   // eslint-disable-next-line func-names
   $(document).on('click', '.tab-toggle', function (e) {
     e.preventDefault();
-    $('.tab-nav a[href="' + $(this).attr('href') + '"]').trigger('click');
+    $('[data-tab-nav] a[href="' + $(this).attr('href') + '"]').trigger('click');
   });
 
   // eslint-disable-next-line func-names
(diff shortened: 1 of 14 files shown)


wagtail/admin/templates/wagtailadmin/edit_handlers/tabbed_interface.html
```diff
@@ -1,20 +1,21 @@
...
+    <ul data-tab-nav role="tablist" data-current-tab="{{ self.children.0.heading|cautious_slugify }}">
+        {% for child in self.children %}
+            <li class="{{ child.classes|join:" " }} {% if forloop.first %}active{% endif %}" role="tab" aria-controls="tab-{{ child.heading|cautious_slugify }}">
+                <a href="#tab-{{ child.heading|cautious_slugify }}" class="{% if forloop.first %}active{% endif %}" data-tab="{{ child.heading|cautious_slugify }}">{{ child.heading }}</a>
+            </li>
+        {% endfor %}
+    </ul>
     {% if self.show_comments_toggle %}
-        <li class="right wide">
+        <span class="right wide">
```
kaedroho, line 11: Probably should be a div?

wagtail/admin/templates/wagtailadmin/edit_handlers/tabbed_interface.html
```diff
@@ -1,20 +1,21 @@
...
             <div class="comments-controls">
                 <label class="comments-toggle" aria-label="{% trans 'Show comments' %}">
                     <input type="checkbox" checked>
                     {% icon name="comment" class_name="comments-toggle__icon" %}
                     <div class="comments-toggle__count">{# populated in page-editor.js #}</div>
                 </label>
-
+
```
kaedroho, line 18: Trailing space, you can set up your editor to remove this automatically

client/src/entrypoints/admin/core.js
```diff
@@ -299,8 +299,8 @@ $(() => {
   const showTab = (tabButtonElem) => {
     $(tabButtonElem).tab('show');
 
-    // Update data-current-tab attribe on the .tab-nav element
-    const tabNavElem = tabButtonElem.closest('.tab-nav');
+    // Update data-current-tab attribute on the [data-tab-nav] element
+    const tabNavElem = tabButtonElem.closest('[data-tab-nav]');
```
kaedroho, line 303: Are we OK to make this breaking change? (ping @gasman @thibaudcolas) I think there are third-party apps that use the tabs, but this is undocumented

wagtail/admin/templates/wagtailadmin/edit_handlers/tabbed_interface.html
jacobtoppm: Thanks, should be removed now! Thought I had got that set up, but apparently only on certain file types - hopefully my editor will catch it now.

wagtail/admin/templates/wagtailadmin/edit_handlers/tabbed_interface.html
jacobtoppm: Fixed!

client/src/entrypoints/admin/core.js
kaedroho: Could we change the selector to something like `'ul.tab-nav .tab-nav > ul'` so It caters for both?

jacobtoppm: Is this a big issue? I'd thought we generally tried to avoid attaching too much js using pure css class selectors, so I thought changing to a data attribute was clearer. Also quite easy to change any external dependencies, just change the selector.

But happy to change our css selector instead of using the data attribute if that's easier?

PR #7023, opened 2021-04-18
Wagtail API pages endpoint to respond with a 400 error (instead of 500 error) when entering an invalid value to a filter

### LINKED ISSUE #7011: Wagtail API pages endpoint responds with a 500 error when entering an invalid value to a filter
### Issue Summary

When querying the Wagtail API and using filters, invalid filter values cause the server to respond with a 500 error.

Given a `BlogPage` with an `author` ForeignKey.

e.g. `/api/v2/pages/?type=blog.BlogPage&author=a`

This returns a 500 error.

### Steps to Reproduce

1. Create any page type that has a foreign key
2. Query that page type using the Wagtail API, and pass an invalid value to a filter (e.g. a letter to a foreign key field)
3. Server Error 500

* This is not a server error. It's a client error.
* We already have this error message for invalid IDs:

```
HTTP 400 Bad Request
Allow: GET, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "message": "field filter error. 'a' is not a valid value for id (invalid literal for int() with base 10: 'a')"
}
```

* A 400 error is expected.

### Technical details

* Python version: 3.8.6.
* Django version: 3.1.
* Wagtail version: 2.12.
* Browser version: Firefox 87

Tijani-Dia:
Hello, I think Wagtail API pages endpoint responds with a 500 error when entering an invalid value to a filter because no check is made on models.ForeignKey [here](https://github.com/wagtail/wagtail/blob/42ddd8ad554f02e9761f1a36f45e162d3edfa7c1/wagtail/api/v2/filters.py#L37).

I added models.ForeignKey and it seems to fix it.

* Do the tests still pass? (https://docs.wagtail.io/en/latest/contributing/developing.html#testing) Yes
    

Fixes #7011

### DIFF
--- a/wagtail/api/v2/filters.py
+++ b/wagtail/api/v2/filters.py
@@ -34,7 +34,7 @@ class FieldsFilter(BaseFilterBackend):
                 try:
                     if isinstance(field, (models.BooleanField, models.NullBooleanField)):
                         value = parse_boolean(value)
-                    elif isinstance(field, (models.IntegerField, models.ForeignKey, models.AutoField)):
+                    elif isinstance(field, (models.IntegerField, models.AutoField)):
                         value = int(value)
                     elif isinstance(field, models.ForeignKey):
                         value = field.get_prep_value(value)
--- a/wagtail/api/v2/filters.py
+++ b/wagtail/api/v2/filters.py
@@ -36,6 +36,8 @@ class FieldsFilter(BaseFilterBackend):
                         value = parse_boolean(value)
                     elif isinstance(field, (models.IntegerField, models.ForeignKey, models.AutoField)):
                         value = int(value)
+                    elif isinstance(field, models.ForeignKey):
+                        value = field.get_prep_value(value)
                 except ValueError as e:
                     raise BadRequestError("field filter error. '%s' is not a valid value for %s (%s)" % (
                         value,
--- a/wagtail/api/v2/filters.py
+++ b/wagtail/api/v2/filters.py
@@ -37,7 +37,7 @@ class FieldsFilter(BaseFilterBackend):
                     elif isinstance(field, (models.IntegerField, models.AutoField)):
                         value = int(value)
                     elif isinstance(field, models.ForeignKey):
-                        value = field.get_prep_value(value)
+                        value = field.target_field.get_prep_value(value)
                 except ValueError as e:
                     raise BadRequestError("field filter error. '%s' is not a valid value for %s (%s)" % (
                         value,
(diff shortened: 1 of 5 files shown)


wagtail/api/v2/filters.py
```diff
@@ -34,7 +34,7 @@ def filter_queryset(self, request, queryset, view):
                 try:
                     if isinstance(field, (models.BooleanField, models.NullBooleanField)):
                         value = parse_boolean(value)
-                    elif isinstance(field, (models.IntegerField, models.AutoField)):
+                    elif isinstance(field, (models.IntegerField, models.ForeignKey, models.AutoField)):
```
kaedroho, line 37: ``ForeignKey``'s could sometimes be text or a UUID internally.

Could you try using `field.get_prep_value(value)` here? this should raise a `ValueError` if the value is invalid

kaedroho (changes requested):

wagtail/api/v2/filters.py
Tijani-Dia: I thought about it but I didn't know how to handle it.
I made the changes requested.
Is that it?

Tijani-Dia: Tests aren't passing for mysql 3.6 appearly. 
How can it be fixed?

kaedroho: I'm not sure what's going on there, maybe we could try performing `get_prep_value` on the remote field?

So try: `value = field.remote_field.get_prep_value(value)`

Tijani-Dia: Seems that it fixes the problem. However the error message for mysql 3.65 is different which makes the tests fail. How can that be fixed? 
If we only make assertions for status code, tests should pass but it is the right solution?

kaedroho: Maybe we could test the prefix? ``self.assertEqual(content['message'][:62], 'field filter error. 'abc' is not a valid value for feed_image')``

Tijani-Dia: Tests pass now!

kaedroho: Nice one, thanks!

kaedroho:
@gasman is it worth putting this in 2.13 as it is a bugfix?

gasman:
@kaedroho No, I think it's best not to (unless it was a regression from 2.12) - even small fixes can have unintended knock-on effects, and we want the release candidate process to pick that up (which won't happen if we make further fixes after rc1 is out).

kaedroho:
@Tijani-Dia Thanks for your first contribution! How would you like your name to appear in the release notes?

Tijani-Dia:
@kaedroho Thanks for your feedback too! Tidjani Dia is fine.

PR #7024, opened 2021-04-19
Add PageSubscription model and hook up the comment notifications switch

kaedroho:


wagtail/admin/edit_handlers.py
```diff
@@ -876,6 +876,21 @@ def render(self):
         panel = render_to_string(self.template, self.get_context())
         return panel
 
+
+class PageSubscriptionPanel(EditHandler):
```
jacobtoppm, line 880: Since this is so minimal, could we just combine it with `CommentPanel` for now and split it out later if we get more notification types?

kaedroho: Thanks! Fixed + squashed

jacobtoppm (approved):

PR #7025, opened 2021-04-19
Add initial draft of commenting update notes

jacobtoppm:

### DIFF
--- a/docs/releases/2.13.rst
+++ b/docs/releases/2.13.rst
@@ -17,6 +17,12 @@ The StreamField editing interface has been rebuilt on a client-side rendering mo
 
 This feature was developed by Matt Westcott and Karl Hobley and sponsored by `YouGov <https://yougov.co.uk/>`_, inspired by earlier work on `react-streamfield <https://github.com/wagtail/wagtail-react-streamfield>`_ completed by Bertrand Bordage through the `Wagtail's First Hatch <https://www.kickstarter.com/projects/noripyt/wagtails-first-hatch>`_ crowdfunder.
 
+Commenting
+~~~~~~~~~~
+
+The page editor now supports leaving comments on fields and StreamField blocks, by entering commenting mode (using the button in the top right of the editor). Inline comments are available in rich text fields using the Draftail editor.
+
+This feature was developed by Jacob Topp-Mugglestone, Karl Hobley and Simon Evans and sponsored by `The Motley Fool <https://www.fool.com/>`_.
 
 Other features
 ~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


docs/releases/2.13.rst
```diff
@@ -95,3 +102,16 @@ Setting menu items now use SVG icons by default. For sites reusing built-in Wagt
...
+~~~~~~~~~~~~
+
+``Page.content_panels`` now includes `CommentPanel`, which is used to save and load comments. If you are overriding page edit handlers and
+want to use the new commenting system, your list of edit handlers should be updated to include ``CommentPanel``.
+
+New data attribute in rich text
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To allow paragraphs in rich text to be identified consistently by the commenting system, the unique block key used in the Draftail editor
+for each block level element is now saved in the generated database html in the ``data-block-key`` attribute.
```
gasman, line 117: Who does this affect, and what action do they need to take?

docs/releases/2.13.rst
```diff
@@ -95,3 +102,16 @@ Setting menu items now use SVG icons by default. For sites reusing built-in Wagt
         return SettingMenuItem(CustomSetting, icon='custom-cog')
         # After:
         return SettingMenuItem(CustomSetting, icon='', classnames='icon icon-custom-cog')
+
+
+CommentPanel
+~~~~~~~~~~~~
+
+``Page.content_panels`` now includes `CommentPanel`, which is used to save and load comments. If you are overriding page edit handlers and
+want to use the new commenting system, your list of edit handlers should be updated to include ``CommentPanel``.
```
gasman, line 111: Could do with a code example here, to show what the import and declaration for `CommentPanel` look like, and clarify that you only need it if your `content_panels` definition _doesn't_ incorporate `Page.content_panels`.

gasman (changes requested):

docs/releases/2.13.rst
jacobtoppm: This is one I wasn't sure about whether to include. It affects everyone using Draftail, but I can't really think of any circumstances under which it would break anything - it's more just a heads up that there will be an extra attribute in the html. Maybe it makes more sense just to remove this note?

docs/releases/2.13.rst
jacobtoppm: Thanks, will add some code.

jacobtoppm: Added now

docs/releases/2.13.rst
jacobtoppm: Removed the note for now.

gasman (approved):

PR #7028, opened 2021-04-19
Notification emails for commenting

kaedroho:


wagtail/admin/views/pages/edit.py
```diff
@@ -55,6 +56,45 @@ def add_save_confirmation_message(self):
...
+        # Get changes
+        comments_formset = self.form.formsets['comments']
+        new_comments = comments_formset.new_objects
+        deleted_comments = comments_formset.deleted_objects
+
+        # Assume any changed comments that are resolved were only just resolved
+        resolved_comments = []
+        for changed_comment, changed_fields in comments_formset.changed_objects:
+            if changed_comment.resolved_at and 'resolved' in changed_fields:
+                resolved_comments.append(changed_comment)
```
jacobtoppm, line 76: This doesn't include replies as far as I can see - new replies are also something people should be notified about, and I believe the idea was that even with "page level" notifications off, anyone in the thread would get reply notifications.

wagtail/admin/views/pages/edit.py
```diff
@@ -55,6 +56,45 @@ def add_save_confirmation_message(self):
...
+
+        # Skip if no changes were made
+        if not new_comments and not deleted_comments and not resolved_comments:
+            return
+
+        # Get subscribers
+        subscribers = PageSubscription.objects.filter(page=self.page, comment_notifications=True).select_related('user')
+        recipient_users = [subscriber.user for subscriber in subscribers if subscriber.user != self.request.user]
```
jacobtoppm, line 84: This doesn't account for the userprofile opt out setting

jacobtoppm (changes requested):
Generally looking good (we don't use `send_notifications` much any more, the `Notifier` class gets used more, but that's fine), but needs to account for replies and opt outs. In the future we should also have a link to scroll directly to the comment, but that seems too much for getting this merged quickly!

wagtail/admin/views/pages/edit.py
kaedroho: This is accounted for in ``send_notification``

jacobtoppm: Oh yep, my mistake, sorry! Was looking for the name explicitly, I see the check now. (Trying to code review in a hurry, oops)

PR #7040, opened 2021-04-20
Feature/comment reply notifications

jacobtoppm:

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/notifications/updated_comments.txt
+++ b/wagtail/admin/templates/wagtailadmin/notifications/updated_comments.txt
@@ -19,5 +19,13 @@
  - "{{ comment.text|safe }}"
 {% endfor %}{% endif %}
 
+{% if replied_comments %}
+{% trans "New replies" %}
+{% for thread in replied_comments %}
+{% trans 'New replies to:' %} {{ thread.comment.text }}
+{% for reply in thread.replies %}
+   - {{ reply.text }}
+{% endfor %}{% endfor %}{% endif %}
+
 {% trans "You can edit the page here:" %} {{ settings.BASE_URL }}{% url 'wagtailadmin_pages:edit' page.id %}
 {% endblock %}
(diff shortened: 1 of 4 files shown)


wagtail/admin/views/pages/edit.py
```diff
@@ -79,15 +81,46 @@ def send_commenting_notifications(self):
         if not new_comments and not deleted_comments and not resolved_comments:
             return
 
-        # Get subscribers
+        relevant_comment_ids = [comment.pk for comment in new_comments + resolved_comments]
```
kaedroho: Not sure why we need to add new_comments here since this is list is only used for handling emails about individual threads?

wagtail/admin/views/pages/edit.py
```diff
@@ -79,15 +81,46 @@ def send_commenting_notifications(self):
...
+        )
 
         # Skip if no recipients
-        if not recipient_users:
+        if not (global_recipient_users or thread_users):
             return
+        thread_users = [(user, set(list(user.comment_replies.values_list('comment_id', flat=True)) + list(user.comments.values_list('pk', flat=True)))) for user in thread_users]
+        mailed_users = set()
+
+        for current_user, current_threads in thread_users:
```
kaedroho, line 117: This is really nice! But I think it's worth commenting here that we are trying to avoid calling ``send_notification`` for each user for performance reasons

wagtail/admin/templates/wagtailadmin/notifications/updated_comments.txt
```diff
@@ -19,5 +19,13 @@
  - "{{ comment.text|safe }}"
 {% endfor %}{% endif %}
 
+{% if replied_comments %}
+{% trans "New replies" %}
+{% for thread in replied_comments %}
+{% trans 'New replies to:' %} - {{ thread.comment.text }}
```
kaedroho, line 25: Could we remove the hyphen here? I think the colon is enough

kaedroho (approved):

wagtail/admin/views/pages/edit.py
jacobtoppm: Thanks! Will add a comment

wagtail/admin/templates/wagtailadmin/notifications/updated_comments.txt
jacobtoppm: Yep, will do, thanks

wagtail/admin/views/pages/edit.py
jacobtoppm: Good point, we don't! Will remove

PR #7042, opened 2021-04-20
Feature/commenting update comment icons

SimonDEvans:
Updates icons to use correct types and states.


jacobtoppm (changes requested):
It looks like this is missing the corresponding streamfield change: see https://github.com/wagtail/wagtail/blob/feature/commenting-update-icons/client/src/components/StreamField/blocks/FieldBlock.js#L57

PR #7054, opened 2021-04-21
Add po files to simple_translation

allcaps:
Added the locales to simple_translation. I took the same locales as in `wagtail/core/locale`.

@gasman These need to be synced with Transifex. Since there are no changes to main that mutate translation keys, that should not be a problem. Maybe there is a way to only sync `wagtail.contrib.simple_translation`?

* Does the code comply with the style guide? (Run `make lint` from the Wagtail root) N/A


gasman:
I don't think we want to create .po files for all languages at this point - the usual way is to just create the English one, and the others will get added as and when translators complete them. There's also a bit of config to update https://github.com/wagtail/wagtail/wiki/Managing-Wagtail-translations#setting-up-translations-for-a-new-app - I'll sort this.

PR #7057, opened 2021-04-21
IE11 support for commenting

kaedroho:
It doesn't look great, but all the functionality works


jacobtoppm (changes requested):
Generally looks good, but I'm not sure about the details/summary fix. In order to address the focusing issue, it looks like it will be necessary to use details/summary as a controlled component: https://github.com/wagtail/wagtail/pull/7060/commits/bfd922a2f48cd7205c91e917f5205dffc6078f2d

I'm happy to rebase this on top of the work here when needed, however the way in which `Summary` uses `open` here rather than `Details` feels confusing to me. If you take the controlled component approach from the other commit, could we not replace `Details` with something that takes the `open` prop that we pass in and adjust the css class based on that?

kaedroho:
@jacobtoppm rebased!

jacobtoppm (approved):
LGTM!

PR #7058, opened 2021-04-21
Feature/commenting make comment buttons focusable

SimonDEvans:
@jacobtoppm it doesn't look like the button inside Draftail's toolbar is reachable, but I think that's likely a separate issue or user error on my part - I can't tab to any of Draftail's toolbar buttons.


jacobtoppm (approved):
Great, thanks for fixing this!

PR #7060, opened 2021-04-21
Close the comment menu when unfocused

jacobtoppm:


client/src/components/CommentApp/components/CommentHeader/index.tsx
```diff
@@ -53,6 +54,24 @@ export const CommentHeader: FunctionComponent<CommentHeaderProps> = ({
...
+      setMenuOpen(false);
+    }
+  }, [focused]);
+  const menuRef = useRef<HTMLDivElement>(null);
+
+  const toggleMenu = (e: React.MouseEvent) => {
+    e.preventDefault();
+    if (!menuOpen) {
+      setMenuOpen(true);
+      setTimeout(() => menuRef.current?.focus(), 1);
```
kaedroho, line 69: Could use use a `useEffect` for this?

jacobtoppm: For the focus part specifically? Yep, I think that could be moved into the `useEffect` hook. Or the toggling as well?

kaedroho: Just the focusing part

kaedroho (approved):

PR #7063, opened 2021-04-21
Remove the comment formset if the management form is missing

### LINKED ISSUE #7061: CommentPanel breaks assertCanCreate in existing tests
The addition of CommentPanel to the base Page panels definition in 1ca48eb31f9a3bcf673d78379f5f23f7d727c639 adds a formset to the posted data; this means that any existing test code that posts to the page edit view (for a page model that does not override this) now fails with

    django.core.exceptions.ValidationError: ['ManagementForm data is missing or has been tampered with']

This issue has been masked in 63c536eef3525d9d4bb322b3740078694570c9a2 by modifying `settings_panels` on all test page models, but this is not a viable option in real-world projects, and posting to the page edit view in tests is standard practice as codified by [the assertCanCreate method](https://docs.wagtail.io/en/latest/advanced_topics/testing.html#wagtail.tests.utils.WagtailPageTests.assertCanCreate). An example of a project where this is failing on current main is wagtail-webstories:

https://github.com/torchbox/wagtail-webstories/blob/main/tests/tests/test_editing.py#L24-L57

At minimum, we need to update documentation and add an upgrade note to tell developers to add the comment formset to their postdata; preferably, we should update the CommentPanel logic to gracefully handle the absence of this formset in the postdata.

jacobtoppm:
Fixes #7061

This means that test data does not need to be updated/`CommentPanel` removed from test models

### DIFF
--- a/wagtail/admin/edit_handlers.py
+++ b/wagtail/admin/edit_handlers.py
@@ -852,7 +852,9 @@ class CommentPanel(EditHandler):
         user_pks = {user.pk}
         serialized_comments = []
         bound = self.form.is_bound
-        for form in self.form.formsets['comments'].forms if self.form else []:
+        comment_formset = self.form.formsets.get('comments')
+        comment_forms = comment_formset.forms if comment_formset else []
+        for form in comment_forms:
             # iterate over comments to retrieve users (to get display names) and serialized versions
             replies = []
             for reply_form in form.formsets['replies'].forms:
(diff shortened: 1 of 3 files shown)


zerolab:
Fixes #7061 ?

jacobtoppm:
Thanks, yep, that's the intent!

gasman (approved):
Thanks @jacobtoppm!

PR #7067, opened 2021-04-21
Update Makefile

jaap3:
This file doesn't seem to have been used by anyone in a while. I tend to always use Makefiles in my projects and just stumbled upon this by executing `make` out of habit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,11 @@
-.PHONY: clean-pyc develop
+.PHONY: clean-pyc develop lint test coverage
 
 help:
 	@echo "clean-pyc - remove Python file artifacts"
+	@echo "develop - install development dependencies"
 	@echo "lint - check style with flake8"
-	@echo "test - run tests quickly with the default Python"
-	@echo "coverage - check code coverage quickly with the default Python"
+	@echo "test - run tests"
+	@echo "coverage - check code coverage"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -13,6 +14,7 @@ clean-pyc:
 
 develop: clean-pyc
 	pip install -e .[testing,docs]
+	nvm install
```
kaedroho, line 17: Is nvm shipped with node.js? If not, could we check for it first?

jaap3: I don't think it's shipped with node.js. I just copied this additional instruction from the developing documentation here: https://github.com/wagtail/wagtail/blob/main/docs/contributing/developing.rst

jaap3: I think it might be better to just remove this line. Upon re-reading the developing docs I see that `nvm` is mentioned as an optional way to install the correct node version. I don't really feel checking for the existence of `nvm` should be part of this `Makefile`. All other tasks operate under the assumption the dev envoriment is set up correctly (e.g. a Python virtualenv exists and is of the correct python version, dependencies are installed etc.).

gasman:
Thanks @jaap3! Merged in 5808a4940751f05d7836b394615e2be68eebd895 (with nvm line removed).

PR #7069, opened 2021-04-21
Redirect to parent page when a child page is deleted from its listing view

### LINKED ISSUE #6527: Delete page in admin and get 404
### Issue Summary
When I delete the root page (a new project) I get a 404 in admin as after a delete Wagtail tries to go back to that page.

### Steps to Reproduce

1. start wagtail project
2. delete homepage that is made
3. get 404 page

I would expect to back to the three or see a screen that the page is deleted.

### Technical details

* Wagtail version: 2.11.1

Tijani-Dia:
Fixes #6527 

Do the tests still pass? Yes
    Does the code comply with the style guide? Yes
    For Python changes: Have you added tests to cover the new/fixed behaviour? No, I'm not really sure how to test it since it involves the frontend.


allcaps:
@Tijani-Dia Thank you!

Can you add a test? When under test we are sure this never breaks again.

Maybe we should add a comment?
# After deleting the page, it is impossible to redirect to it.

When if statements are nested and there is no else, it is better to combine the statements:

if next and next != ...:

Tijani-Dia:
> Can you add a test? When under test we are sure this never breaks again.
 Sure! Where can I add the tests? I can't figure out.

> Maybe we should add a comment?
> > After deleting the page, it is impossible to redirect to it.
Ok.

> When if statements are nested and there is no else, it is better to combine the statements:
OK

allcaps:
Can you search the test folder for page_listing_more_buttons? There should be existing tests, and it is likely that you can extend the existing test.

We are only looking for a unit test, a test that shows your added if statement works. We don't need an integration test (a test that checks the html output and click flow).

allcaps:
It seems that a test should go in this test class `wagtail.admin.tests.pages.test_delete_page.TestPageDelete`

You can also do something like this (pseudo code):

from wagtail.admin.wagtail_hooks import page_listing_more_buttons

class PagePerms:
   def can_move(self):
      return false
   def can_copy(self):
      return false
   def can_copy(self):
      return True
   # False for other needed methods.

page_perms=PagePerms()
page = ... # see other tests.
next_url = "/not/very/relevant/slugs/"
# This will return our delete button!
button = page_listing_more_buttons(page, page_perms, is_parent=False, next_url=next_url)
assert button.url == "/something/?next=/not/very/relevant/slugs/"

next_url = reverse("admin:something", args=(page.id))
button = page_listing_more_buttons(page, page_perms, is_parent=False, next_url=next_url)
assert button.url == "/something-you/d/expect/here/"

Tijani-Dia:
Alright, I am going to do it.

allcaps:
Thank you @Tijani-Dia , for your quick follow-up. I'm going to leave the review to someone else, to get a new pair of eyes on this.

wagtail/admin/tests/pages/test_delete_page.py
```diff
@@ -202,3 +221,31 @@ def hook_func(request, page):
...
+    def test_delete_button_next_url(self):
+        page_perms = PagePerms()
+        page = self.child_page
+        base_url = reverse('wagtailadmin_pages:delete', args=[page.id])
+
+        next_url = "a/random/url/"
+        full_url = base_url + '?' + urlencode({'next': next_url})
+
+        # page_listing_more_button generator yields only `Delete button`
+        delete_button = next(page_listing_more_buttons(
```
kaedroho, line 234: We shouldn't be testing ``page_listing_more_buttons`` here. We should either test this behaviour at the view level (by asserting on the HTML output or template context), or we should move this into a separate unit test suite for `page_listing_more_buttons`

wagtail/admin/wagtail_hooks.py
```diff
@@ -210,7 +210,9 @@ def page_listing_more_buttons(page, page_perms, is_parent=False, next_url=None):
         )
     if page_perms.can_delete():
         url = reverse('wagtailadmin_pages:delete', args=[page.id])
-        if next_url:
+
+        # After deleting the page, it is impossible to redirect to it.
+        if next_url and next_url != reverse('wagtailadmin_explore', args=[page.id]):
```
kaedroho, line 215: Just an opinion, it feels like this bit of logic should be in its own block for added clarity to the comment:

```python
# After deleting the page, it is impossible to redirect to it.
if next_url != reverse('wagtailadmin_explore', args=[page.id]):
    next_url = None

if next_url:
```

kaedroho (changes requested):

wagtail/admin/tests/pages/test_delete_page.py
Tijani-Dia: Can I move the tests into wagtail/admin/tests/test_button_hooks.py ?

kaedroho: Yep, that sounds good to me, thanks!

Tijani-Dia: It's done!

kaedroho: awesome, thanks!

kaedroho (approved):

PR #7071, opened 2021-04-22
Fix for CheckboxSelectMultiple widget

coredumperror:
Without this, any project that uses `django.forms.widgets.CheckboxSelectMultiple` crashes needlessly in any step that touches the database, because of Telepath's "don't know how to pack object" error. 

Simply applying WidgetAdapter to CheckboxSelectMultiple seems to work fine, though I haven't tested it in any robust way. I just went in to my site that uses CheckboxSelectMultiple in several places, and confirmed that it still works.

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -104,3 +104,8 @@ source_file = wagtail/contrib/table_block/locale/en/LC_MESSAGES/django.po
 source_lang = en
 type = PO
 
+[wagtail.wagtailsimpletranslation]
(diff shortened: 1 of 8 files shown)


coredumperror:
ARG damnit. I did my PR wrong... There's only suppose to be a one-line change in one file.

coredumperror:
Nevermind, I clearly did this all wrong. Ignore this.

PR #7072, opened 2021-04-22
Apply WidgetAdapter to CheckboxSelectMultiple.

coredumperror:
Without this, any project that uses `django.forms.widgets.CheckboxSelectMultiple` crashes needlessly in any step that touches the database, because of Telepath's "don't know how to pack object" error.

Simply applying WidgetAdapter to CheckboxSelectMultiple seems to work fine, though I haven't tested it in any robust way. I just went in to my site that uses CheckboxSelectMultiple in several places, and confirmed that it still functions.


gasman (approved):
Thanks @coredumperror! Just tested here too, and all working well.

Issue #7064, opened 2021-04-21
Comments float in strange places when written in a collapsed MultiFieldPanel


KalobTaulien:
When writing a comment a MultiFieldPanel that's collapsed by default in Wagtail 2.13rc1 the comment boxes float in slightly strange places.

Preview: 

https://user-images.githubusercontent.com/4743971/115608882-11afc400-a2a4-11eb-87b4-f1c0c6103c3b.mp4

And when I delete a comment that's inside a MultiFieldPanel, the _other_ comment (from the video demo) jumps to the top. Example below:
![image](https://user-images.githubusercontent.com/4743971/115609191-779c4b80-a2a4-11eb-9cbd-057b3963b4bb.png)

jacobtoppm:
Thanks for catching this! This was also a problem with collapsed streamfield blocks. Hopefully the link PR will address this now.

PR #7078, opened 2021-04-22
Account for collapsed panels when positioning comments

jacobtoppm:
Fixes https://github.com/wagtail/wagtail/issues/7064

### DIFF
--- a/client/src/components/CommentApp/main.tsx
+++ b/client/src/components/CommentApp/main.tsx
@@ -354,6 +354,15 @@ export class CommentApp {
         }
       }
     });
+
+    document.body.addEventListener('commentAnchorVisibilityChange', () => {
+      // If any streamfield blocks or panels have collapsed or expanded
+      // check if we need to rerender
+      this.layout.refreshDesiredPositions(this.store.getState().settings.currentTab);
+      if (this.layout.refreshLayout()) {
+        render();
+      }
+    });
   }
 }
 
--- a/client/src/components/CommentApp/utils/annotation.ts
+++ b/client/src/components/CommentApp/utils/annotation.ts
@@ -1,4 +1,4 @@
 export interface Annotation {
   getTab(): string | null | undefined;
-  getDesiredPosition(focused: boolean): number;
+  getAnchorNode(focused: boolean): Element;
 }
--- a/client/src/components/CommentApp/utils/layout.ts
+++ b/client/src/components/CommentApp/utils/layout.ts
@@ -51,9 +51,11 @@ export class LayoutController {
       return;
     }
 
+    const currentNodeTop = annotation.getAnchorNode(commentId === this.pinnedComment).getBoundingClientRect().top;
+
     this.commentDesiredPositions.set(
       commentId,
-      annotation.getDesiredPosition(commentId === this.pinnedComment) + OFFSET
+      currentNodeTop !== 0 ? currentNodeTop + document.documentElement.scrollTop + OFFSET : 0
     );
   }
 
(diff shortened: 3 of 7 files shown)


client/src/components/CommentApp/utils/annotation.ts
```diff
@@ -1,4 +1,4 @@
 export interface Annotation {
   getTab(): string | null | undefined;
-  getDesiredPosition(focused: boolean): number;
+  getDesiredPositionNode(focused: boolean): Element;
```
kaedroho, line 3: How about `getAnchorNode`?

client/src/components/CommentApp/utils/layout.ts
```diff
@@ -51,9 +51,19 @@ export class LayoutController {
       return;
     }
 
+    // If the annotation's desired position node is not visible,
```
kaedroho, line 54: Wondering if we should hide the comments if their fields are hidden?

Otherwise, would it be easy to expand any collapsed panels when the comment is focused?

jacobtoppm: I'd be inclined to leave the comments shown - it doesn't look bad, and comments are the kind of thing we want to be easily found. If you don't want to see comments, then you can always leave comment mode!

Easyish, but I think we'd probably want to do it as a new subscriber for a) streamfield block and b) multifieldpanel listening for the focused comment being in the list of comments with contentpath starting with their contentpath. I don't think it would be too hard, I can think how to do it, but it is extra code that interacts with eg telepath - wondering if this is a step too far for the release candidate level of tweaks?

client/src/components/CommentApp/utils/annotation.ts
jacobtoppm: Yep, that's better!

client/src/components/CommentApp/utils/layout.ts
jacobtoppm: Hmm, though with the first, I suppose for big streamfields with lots of comments it could be useful...

client/src/entrypoints/admin/page-editor.js
```diff
@@ -264,17 +264,26 @@ function initCollapsibleBlocks() {
...
-        $fieldset.hide('slow');
+        $fieldset.hide({
+          duration: 'slow',
+          complete: onAnimationComplete
+        });
       } else {
         $li.removeClass('collapsed');
-        $fieldset.show('show');
+        $fieldset.show({
+          duration: 'show',
```
kaedroho, line 284: Wonder if this was 'show' by mistake, and should instead be 'slow' as above?

client/src/components/CommentApp/main.tsx
```diff
@@ -354,6 +354,15 @@ export class CommentApp {
         }
       }
     });
+
+    document.body.addEventListener('visibilityChange', () => {
```
kaedroho, line 358: Could we name this something like `commentAnchorVisibilityChange` so people know it needs to be fired when comment anchors change?

kaedroho (approved):
LGTM, just two minor comments

client/src/entrypoints/admin/page-editor.js
jacobtoppm: Good point, changed!

client/src/components/CommentApp/main.tsx
jacobtoppm: Yep, can do - we can rename it more generically when we actually use it for something else

PR #7081, opened 2021-04-22
Start 2.14

kaedroho:


gasman (approved):
We should also add a line in https://github.com/wagtail/wagtail/blob/main/docs/releases/upgrading.rst (just copy the one from 2.13 for now), but other than that, looks good to me!

(I thought we had some test coverage for `StreamValue.stream_data` which would also need dropping now, but I guess we'll find out from CI...)

kaedroho:
Added, thanks!

PR #7089, opened 2021-04-23
Add warning to bottom of draft comments

kaedroho:
The page needs to be saved after inserting new comments or those comments will be lost. This PR adds an info message to the bottom of draft comments to help new users to not forget this.

image


jacobtoppm (changes requested):
Looks good for new comments! However we also need to cover:
- Edited comments
- New replies
- Edited replies

jacobtoppm (changes requested):
Maybe worth checking with Ben, but I think the reply warning was meant to be on the reply, rather than the parent comment.

Also we should probably use the edit warning for edited replies. We can use the same text for comment and comment replies though!

jacobtoppm (approved):
Looks good to me! Minor tweak: in the designs we grey the text of any unsaved comments slightly.

PR #7093, opened 2021-04-23
Fix BooleanBlock for telepath

### LINKED ISSUE #7083: blocks.BooleanBlock not working in Streamfield in 2.13rc1
### Issue Summary

Whenever I select blocks.BooleanBlock in a streamfield, when saved, the value is not stored.

### Steps to Reproduce

```
from django.utils.translation import ugettext_lazy as _

from wagtail.core import blocks


class BlockPluginWrapper(blocks.StructBlock):
    plugin_background = blocks.BooleanBlock(
        required=False,
        default=False,
        help_text=_('if true, the background will be gray.'),
        group='wrapper'
    )
    plugin_title = blocks.CharBlock(
        max_length=150,
        required=False,
        help_text=_('optional'),
        group='wrapper'
    )

    def clean(self, value):
        print(value)

        return super().clean(value)

    class Meta:
        form_template = 'blocks/block_plugin_wrapper.html'
```

In the clean method the following is printed, even though plugin_background should be True:

```
StructValue([('plugin_background', False), ('plugin_title', 'test')])
```

Streamfield:

```
content = StreamField(
        [
            ('test', BlockPluginWrapper())
        ],
        null=True,
        blank=True
    )
```

### Technical details

* Python version: 3.7.9
* Django version: 3.1.8
* Wagtail version: 2.13rc1

gasman:
Fixes #7083. Override default FieldBlock / BoundWidget behaviour so that:

* BooleanBlock.get_form_state bypasses CheckboxInput.format_value (which Django uses to determine the 'value' HTML attribute rather than the checked state)
* BoundCheckboxInput.get_state / set_state / get_value work with the 'checked' attribute rather than 'value'

### DIFF
--- a/client/src/entrypoints/admin/telepath/widgets.js
+++ b/client/src/entrypoints/admin/telepath/widgets.js
@@ -52,6 +52,26 @@ class Widget {
 window.telepath.register('wagtail.widgets.Widget', Widget);
 
 
+class BoundCheckboxInput extends BoundWidget {
+    getValue() {
+        return this.input.is(':checked');
+    }
+    getState() {
+        return this.input.is(':checked');
+    }
+    setState(state) {
+        // if false, set attribute value to null to remove it
+        this.input.attr('checked', state || null);
+    }
+}
+
+
+class CheckboxInput extends Widget {
+    boundWidgetClass = BoundCheckboxInput;
+}
+window.telepath.register('wagtail.widgets.CheckboxInput', CheckboxInput);
+
+
 class BoundRadioSelect {
     constructor(element, name, idForLabel, initialState) {
         this.element = element;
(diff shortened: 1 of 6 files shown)


kaedroho (approved):

client/src/entrypoints/admin/telepath/widgets.js
```diff
@@ -52,6 +52,26 @@ class Widget {
 window.telepath.register('wagtail.widgets.Widget', Widget);
 
 
+class BoundCheckboxInput extends BoundWidget {
+    getValue() {
+        return this.input.is(':checked');
```
kaedroho, line 57: Just a minor comment, would be nice to remove jQuery from this API if it's considered public, this would make it easier for us to remove jQuery later (which we've already decided to start phasing out)

gasman: The [public API](https://docs.wagtail.io/en/latest/reference/streamfield/widget_api.html) is `getValue` / `getState` / `setState`, which doesn't expose anything jQuery-specific, so I think we're good here.

PR #7094, opened 2021-04-23
Replacement for js_initialize: a hook to run custom form-side javascript upon StructBlock subclass render.

coredumperror:
I spoke with @gasman about this concept on Slack today, and he helped me a TON with getting the basic idea in place. This is all very preliminary at the moment, but my small amount of testing in Bakeydemo is very promising so far. 

With the following code added to their repo, and this PR's small tweak to StructBlock.js, developers can now get custom javascript to run after their StructBlock subclasses get rendered in the form.

This is added to base/blocks.py in Bakerdemo:
```python 
from django import forms
from django.templatetags.static import static
from django.utils.functional import cached_property
from wagtail.core.blocks.struct_block import StructBlockAdapter
from wagtail.core.telepath import register

class ImageBlockAdapter(StructBlockAdapter):

    def js_args(self, block):
        # Gets the base StructBlockAdapter args, then adds 'onBlockRender' to the meta arg.
        args = super().js_args(block)
        args[2]['onBlockRender'] = 'image_block_render_func'
        return args

    @cached_property
    def media(self):
        return forms.Media(js=[static('js/image-block.js')])


register(ImageBlockAdapter(), ImageBlock)
```

And here's `static/js/image-block.js` (just a proof of concept function):
```
function image_block_render_func(prefix, block) {
  alert('This is the prefix: ' + prefix);
  console.log(block);
}
```

This allows `image_block_render_func(prefix, block)` to trigger every time an ImageBlock gets added to a StreamField (or when the editor for a Page with a StreamField containing one or more ImageBlocks gets displayed). 

This is meant as a replacement for the old `js_initializer()` mechanism that StreamFields used to use for this purpose. Users who have implementations of that will need to be told that they will cease to function, and will need to be migrated to either this solution, or something like it. So this PR really needs at least that documentation added to it, plus tests, before it can be considered complete. I intend to add those next week.

### DIFF
--- a/client/src/components/StreamField/blocks/StructBlock.js
+++ b/client/src/components/StreamField/blocks/StructBlock.js
@@ -147,6 +147,6 @@ export class StructBlockDefinition {
     if (this.meta.onBlockRender && typeof window[this.meta.onBlockRender] === 'function') {
       window[this.meta.onBlockRender](prefix, block);
     }
-    return block
+    return block;
   }
 }
--- a/client/src/components/StreamField/blocks/StructBlock.js
+++ b/client/src/components/StreamField/blocks/StructBlock.js
@@ -143,8 +143,8 @@ export class StructBlockDefinition {
 
   render(placeholder, prefix, initialState, initialError) {
     const block = new StructBlock(this, placeholder, prefix, initialState, initialError);
-    if (this.meta.onBlockRender) {
-      // If a function name was specified for onBlockRender, look it up in the window scope and call it.
+    // If meta.onBlockRender was specified, look it up in the window scope and call it if it's a function.
+    if (this.meta.onBlockRender && typeof window[this.meta.onBlockRender] === 'function') {
       window[this.meta.onBlockRender](prefix, block);
     }
     return block
--- a/client/src/components/StreamField/blocks/StructBlock.js
+++ b/client/src/components/StreamField/blocks/StructBlock.js
@@ -142,6 +142,11 @@ export class StructBlockDefinition {
   }
 
   render(placeholder, prefix, initialState, initialError) {
-    return new StructBlock(this, placeholder, prefix, initialState, initialError);
+    const block = new StructBlock(this, placeholder, prefix, initialState, initialError);
+    if (this.meta.onBlockRender) {
+      // If a function name was specified for onBlockRender, look it up in the window scope and call it.
+      window[this.meta.onBlockRender](prefix, block);
+    }
+    return block
   }
 }


gasman:
Thanks for looking into this @coredumperror!
I had second thoughts about the suitability of a hook / callback for this - I'd much rather leverage the language's capabilities for extending code, rather than coming up with a new mechanism that will need keeping up to date as the code evolves.

After a bit more experimenting, it looks like the only missing piece is [making Wagtail's block definitions available as global variables](https://github.com/gasman/wagtail/commit/e80e99b7ccc33b7ad59d1bc08d811701b25b8ccc) - with that in place, the same thing can be achieved as follows:

blocks.py:
```
from wagtail.core.blocks.struct_block import StructBlockAdapter
from wagtail.core.telepath import register
from django.utils.functional import cached_property
from django.templatetags.static import static
from wagtail.admin.staticfiles import versioned_static


class ImageBlockAdapter(StructBlockAdapter):
    js_constructor = 'bakerydemo.blocks.ImageBlock'

    @cached_property
    def media(self):
        return super().media + forms.Media(js=[
            versioned_static('wagtailadmin/js/telepath/blocks.js'),
            static('js/image-block.js')
        ])


register(ImageBlockAdapter(), ImageBlock)
```

image-block.js (not IE11 compatible in this form, but can hopefully be rewritten to classic prototype-based inheritance without too much bother):
```
class ImageBlockDefinition extends window.wagtailStreamField.blocks.StructBlockDefinition {
    render(placeholder, prefix, initialState, initialError) {
        const block = super.render(placeholder, prefix, initialState, initialError);
        console.log('Do stuff with block:', block, prefix);
        return block;
    }
}
window.telepath.register('bakerydemo.blocks.ImageBlock', ImageBlockDefinition);
```

The one bit that bugs me here is the need to explicitly include `wagtailadmin/js/telepath/blocks.js` in the media - even though we're extending a definition that includes it, the ordering isn't guaranteed (and we need it that way to ensure StructBlockDefinition is defined for image-block.js). Hopefully there's a neater workaround there...

Will do some more fine-tuning and get the new global definitions included in 2.13rc2, and try and get some proper documentation in place for 2.13 final.

gasman:
Media definition that I'm slightly happier with...

```
class ImageBlockAdapter(StructBlockAdapter):
    js_constructor = 'bakerydemo.blocks.ImageBlock'

    @cached_property
    def media(self):
        structblock_media = super().media
        return forms.Media(
            js=structblock_media._js + ['js/image-block.js'],
            css=structblock_media._css
        )
```

gasman:
Closing in favour of #7120.

PR #7097, opened 2021-04-24
Storybook

kaedroho:
This PR adds storybook to our frontend tooling, this can be invoked with either ``npm run storybook`` (live server) or ``npm run build-storybook`` (one off build).

I've updated our commenting stories to work with it too.

Currently, icons are not supported as these are managed in Django. @thibaudcolas @allcaps, would you, in principle, be in favour of moving SVG icon management to the frontend tooling as well (as long as we had a way to extend it nicely)?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ npm-debug.log*
 /*.egg/
 /.cache/
 /.pytest_cache/
+/storybook-static
 
 ### JetBrains
 .idea/
--- a/client/.storybook/main.js
+++ b/client/.storybook/main.js
@@ -8,5 +8,9 @@ module.exports = {
   ],
   "core": {
     "builder": "webpack5"
-  }
+  },
+  "webpackFinal": (config) => {
+    config.resolve.fallback.crypto = false;
+    return config;
+  },
 }
--- a/client/.storybook/main.js
+++ b/client/.storybook/main.js
@@ -1,0 +1,12 @@
+module.exports = {
+  "stories": [
+    "../src/**/*.stories.@(js|jsx|ts|tsx)"
+  ],
+  "addons": [
+    "@storybook/addon-links",
+    "@storybook/addon-essentials"
+  ],
+  "core": {
+    "builder": "webpack5"
+  }
+}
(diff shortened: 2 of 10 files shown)


allcaps:
> (as long as we had a way to extend it nicely)?

The SVG icons are Django templates (override) and Wagtail hooks (override/extend). These are default Django and Wagtail concepts. Easy to explain in docs and maintain in core. But more important, easy to understand and be used by integrators.

I'm not against changing the Wagtail internal icon management, as long as _third-party packages_ and _integrators_ can extend it nicely 👍

kaedroho:
Alternatively, this could be a stop-gap to help us with things like commenting and the menu rebuild in the short term

For the long term, we could improve the existing style guide to support React widgets as well? (and also organise it a bit better, make it extendable, and add reloading) The advantage of this is it'll help us maintain Django widgets too, some of which are duplicated in React, and a side-by-side view might be useful for helping us maintain consistency.

allcaps (approved):
I gave this a run, and LGTM.

In the storybook the items "Saving" and "Deleting" show a red box, with an error:

> Your focus-trap must have at least one container with at least one tabbable node in it at all times ...

Is this expected?

kaedroho:
Yep, I got that too. Focus trap was introduced after the merge into Wagtail so those two tests are broken. Neither of those screens are used right now so we could remove those tests temporarily (or completely remove support for those screens, but we would probably use them later)

allcaps:
I'm fine with those views erroring if we use them later. 👍 

I do like storybook and storybook-django and think it is the way to get a grip on frontend development and make Wagtail frontend testable and reliable. The style guide page is never kept in sync and can't be used for testing. I never saw it as a reliable development tool.

Maybe we can have Wagtail output a 'mock sprite' via a management command? And have it under version control? Then the frontend should use the mock sprite if Python/Django/Wagtail is not available.

Do you have other/better options?

kaedroho:
Thanks @allcaps!

I had a chat with @thibaudcolas about this yesterday, he's going to look into creating an icon loader for storybook to allow it to load the icons from the files in Django.

Shall we merge this in the mean time?

allcaps:
Yes, go for it!

PR #7101, opened 2021-04-26
Styling fixes for the "Are you sure?" prompt when deleting a comment/reply

kaedroho:
image

image


jacobtoppm (changes requested):
Looks great with replies, but on comments, the `cancel` and `delete` buttons both appear in the middle of the dividing line between a comment and its replies (both chrome and firefox).

kaedroho:
Just added clearfixes which hopefully fixes that issue! (I couldn't reproduce though)

jacobtoppm (approved):
LGTM now!

PR #7103, opened 2021-04-26
Fix jitter on draftail comment button icon

SimonDEvans:
Fixes the issue in the gif below.

Kapture 2021-04-26 at 15 19 35


jacobtoppm (approved):
LGTM, thanks!

PR #7104, opened 2021-04-26
Lighten comment more actions icon

kaedroho:
image


jacobtoppm (changes requested):
I know this is the design, but it fails the colour contrast check, sorry! Think we need to go back to design on this one.

jacobtoppm (approved):

PR #7105, opened 2021-04-26
Put notices straight after the relevant comment/reply

kaedroho:
image

image


jacobtoppm (approved):
Looks great, just needs to pass linting and then good to go!

PR #7109, opened 2021-04-26
Fix inconsistent StreamField ValidationError nesting

### LINKED ISSUE #7086: StreamField validation error messages appear as single characters
Screenshot 2021-04-23 at 13 08 11

~(Could be related to the #7059 / #7074 fix? These error messages were presumably localized strings all along, so maybe the error message code was over-correcting for them being returned as lists...)~ No, same thing happens with telepath==0.1.

gasman:
Fixes #7086. As per https://github.com/wagtail/wagtail/issues/7086#issuecomment-826945031, ensure that .as_data() is consistently called when telepath-packing ErrorList objects (so that we preserve any embedded ValidationError objects instead of casting them to strings), and introduce an explicit ValidationError class on the client side to make mismatches more obvious (and for future extensibility in case we need to attach more fancy logic to ValidationError).

Also add tests for setError, and fix rendering of StreamBlock non-field errors (selector to clear old errors was incorrect, and jest apparently doesn't support innerText).

### DIFF
--- a/client/src/components/StreamField/blocks/FieldBlock.js
+++ b/client/src/components/StreamField/blocks/FieldBlock.js
@@ -27,7 +27,7 @@ export class FieldBlock {
       // eslint-disable-next-line no-console
       console.error(e);
       this.setError([
-        ['This widget failed to render, please check the console for details']
+        { messages: ['This widget failed to render, please check the console for details'] }
       ]);
       return;
     }
@@ -78,7 +78,7 @@ export class FieldBlock {
 
       const errorElement = document.createElement('p');
       errorElement.classList.add('error-message');
-      errorElement.innerHTML = errorList.map(error => `<span>${h(error[0])}</span>`).join('');
+      errorElement.innerHTML = errorList.map(error => `<span>${h(error.messages[0])}</span>`).join('');
       this.element.querySelector('.field-content').appendChild(errorElement);
     } else {
       this.element.classList.remove('error');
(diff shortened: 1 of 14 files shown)


client/src/components/StreamField/blocks/FieldBlock.js
```diff
@@ -27,7 +27,7 @@ export class FieldBlock {
       // eslint-disable-next-line no-console
       console.error(e);
       this.setError([
-        ['This widget failed to render, please check the console for details']
+        { messages: ['This widget failed to render, please check the console for details'] }
```
gasman, line 30: Sorry, I know this faking of a ValidationError is a bit hacky (and likewise the duplication of the ValidationError definition in the tests), but making a single ValidationError definition visible to both the StreamField components and the wagtailadmin/js/telepath/widgets.js endpoint would have involved more code reshuffling than I'm happy to do post-RC...

wagtail/core/blocks/struct_block.py
```diff
@@ -27,7 +27,17 @@ class StructBlockValidationErrorAdapter(Adapter):
...
+        if error.block_errors is None:
+            return [None]
+        else:
+            result = [
+                {
+                    name: error_list.as_data()
+                    for name, error_list in error.block_errors.items()
+                }
+            ]
+            print("block_errors: %r" % (result, ))
```
kaedroho, line 39: Stray print

kaedroho (approved):
Looks good to me! just a stray print

wagtail/core/blocks/struct_block.py
gasman: Oops, good catch!

gasman:
Merged (with `print` fix) in 1061caa5eff53f450520284b618e615fc2a809c8 (main) and 498fcbd6cabad21868f94aeda7207c6e60ce2ae7 (stable/2.13.x).

PR #7119, opened 2021-04-27
Remove comment styles from further typing on enter, right/left arrow

jacobtoppm:
On enter:
- Remove comment style from further typing if not splitting an existing comment
On right/left:
- Remove comment style if at the end of the content and pressing in the same direction as text


kaedroho (approved):
Works well for creating a new paragraph, couldn't get it to drop the comment styling on right arrow though

PR #7120, opened 2021-04-27
Make StreamField block objects available globally

gasman:
As discussed on #7094 - making these definitions available as global variables means that we can provide a reasonably simple upgrade path for people who previously customised StructBlock (or others) with custom JS via the `js_initializer` method.

(Would have liked to make these available under the `window.wagtail.blocks` namespace, but it seems there's a previously un-noticed issue where `window.wagtail` gets overwritten by the `<body id="wagtail">` element, so it's not available reliably. Doh!)

Will aim to get this properly documented in time for 2.13 final.

### DIFF
--- a/client/src/entrypoints/admin/telepath/blocks.js
+++ b/client/src/entrypoints/admin/telepath/blocks.js
@@ -1,8 +1,36 @@
-import { FieldBlockDefinition } from '../../../components/StreamField/blocks/FieldBlock';
-import { StaticBlockDefinition } from '../../../components/StreamField/blocks/StaticBlock';
-import { StructBlockDefinition, StructBlockValidationError } from '../../../components/StreamField/blocks/StructBlock';
-import { ListBlockDefinition, ListBlockValidationError } from '../../../components/StreamField/blocks/ListBlock';
-import { StreamBlockDefinition, StreamBlockValidationError } from '../../../components/StreamField/blocks/StreamBlock';
+import { FieldBlock, FieldBlockDefinition } from '../../../components/StreamField/blocks/FieldBlock';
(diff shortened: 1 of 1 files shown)


kaedroho:
> (Would have liked to make these available under the window.wagtail.blocks namespace, but it seems there's a previously un-noticed issue where window.wagtail gets overwritten by the <body id="wagtail"> element, so it's not available reliably. Doh!)
> Will aim to get this properly documented in time for 2.13 final.

For this reason, do you think it's worth leaving it undocumented until 2.14 so that we can sort the namespacing out properly?

gasman:
@kaedroho I'd prefer not to - I think it's important for us to make the transition to telepath StreamFields as smooth and well-documented as possible now, so that people with non-trivial customisations don't end up 'stranded' on the pre-telepath version.

coredumperror:
> @kaedroho I'd prefer not to - I think it's important for us to make the transition to telepath StreamFields as smooth and well-documented as possible now, so that people with non-trivial customisations don't end up 'stranded' on the pre-telepath version.

Hear, hear! :)

kaedroho (approved):

zerolab:
> @kaedroho I'd prefer not to - I think it's important for us to make the transition to telepath StreamFields as smooth and well-documented as possible now, so that people with non-trivial customisations don't end up 'stranded' on the pre-telepath version.

Amen to that! case in point: https://github.com/ixc/wagtail-instance-selector/issues/11#issuecomment-826447149

kaedroho:
I feel firmly put in my place now :)

I agree with @gasman's comment above, but I also think that in the long term, it's good to have definitions follow a consistent pattern. So I hope we can get to refactor this (as well as some other APIs) to follow a scheme that's similar to the original comment.

Issue #5331, opened 2019-05-28
Admin URLs don't respect APPEND_SLASH setting


harrislapiroff:
### Issue Summary

Even with `APPEND_SLASH` set to `True` and `'django.middleware.common.CommonMiddleware'` in `MIDDLEWARE`, Wagtail admin URLs don't respect the [behavior described in the Django docs](https://docs.djangoproject.com/en/2.2/ref/middleware/#module-django.middleware.common).

### Steps to Reproduce

1. With any Wagtail project ensure that settings have `APPEND_SLASH = True` and `MIDDLEWARE` contains `'django.middleware.common.CommonMiddleware'`
2. Observe that `http://localhost:8000/admin/pages/` works
3. Observe that `http://localhost:8000/admin/pages` yields a 404

Expected behavior should be for `http://localhost:8000/admin/pages` to redirect to `http://localhost:8000/admin/pages/`

@gasman theorized in Slack that it may be the custom admin 404 that's interfering with the behavior here.

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: yes

### Technical details

* Python version: 3.5.6
* Django version: 2.1.7
* Wagtail version: 2.3

johnraz:
I can confirm the behavior described above.

I also noticed that going to `http://localhost:8000/admin` would properly redirect to `http://localhost:8000/admin/`.

I believe that redirect is handled by Django alone and probably add to the theory that Wagtail's custom admin 404 is getting in the way for any subtree pages.

Tijani-Dia:
Django's [should_redirect_with_slash](https://github.com/django/django/blob/6e742dabc95b00ba896434293556adeb4dbaee8a/django/middleware/common.py#L63) function:

> """
        Return True if settings.APPEND_SLASH is True and appending a slash to
        the request path turns an invalid path into a valid one.
        """

To do so, it makes sure that the requested url is not a valid one and only then does it try to append a slash.
The catch here is that all URLs are considered valid([see here](https://github.com/wagtail/wagtail/blob/5271b4e97af5dee7e097fdfb16b3ff7d38b8b06a/wagtail/admin/urls/__init__.py#L97-L99)). Therefore, Django doesn't take a step further to see if appending a slash to the request path turns it into a valid one. That's why no slash is appended in admin URLs even if APPEND_SLASH is set to True.

[...]

Issue #5814, opened 2020-02-07
RoutablePage routes without trailing slashes don't work


Stormheg:
### Issue Summary

Use case: I want to use my wagtail website without trailing slashes in the url.
(`WAGTAIL_APPEND_SLASH = False` and `APPEND_SLASH = False`)

When I create a RoutablePage with a `@route` without a trailing slash in the regex, the route will not resolve. This issue was also mentioned in https://github.com/wagtail/wagtail/issues/2871.

### Steps to Reproduce

1. Start a new project with `wagtail start myproject`
2. Edit models.py as follows
```python
from django.db import models
from django.shortcuts import HttpResponse

from wagtail.core.models import Page
from wagtail.contrib.routable_page.models import RoutablePageMixin, route


class HomePage(Page):
    pass

class TestPage(RoutablePageMixin, Page):
    @route(r"^year/(\d+)$")  # r"^year/(\d+)/$" does work however!
    def year(self, request, year):
        return HttpResponse("This is the page for year {}".format(year))
```
3. Add this to `base.py`
```python
WAGTAIL_APPEND_SLASH = False
APPEND_SLASH = False
```
4. Migrate and create a new TestPage with slug `test` in the Wagtail admin.
5. Visiting http://localhost:8000/test/year/2020 won't work, 404 error.
Visiting http://localhost:8000/test/year/2020/ (with a trailing slash) won't work either.

I tried setting `WAGTAIL_APPEND_SLASH = True` `APPEND_SLASH = True` but that doesn't matter.

I think we should be able to define routable page routes without a trailing slash.

### Technical details

* Python version: 3.8.1.
* Django version: 3.0.3.
* Wagtail version: 2.8.
* Browser version: Chrome 79 on macOS Catalina

Tijani-Dia:
I think this happens because a slash is automatically appended [here](https://github.com/wagtail/wagtail/blob/59f63abdad3a5430c62dde62a665cb86bc236a94/wagtail/contrib/routable_page/models.py#L105).

[Here](https://github.com/Tijani-Dia/wagtail/tree/routable-slash) is a possible fix.

I wanted to add a unit test, but whatever I do WAGTAIL_APPEND_SLASH is True so the serve_pattern is not properly set in the urlconf.
I tried to override settings, mock wagtail.core.utils.WAGTAIL_APPEND_SLASH but none of this work.

Thoughts?

tbrlpld:
I am currently running into this issue as well

[...]

PR #7126, opened 2021-04-27
Admin URLs respect APPEND_SLASH setting

Tijani-Dia:
See #5331

### DIFF
--- a/wagtail/admin/tests/tests.py
+++ b/wagtail/admin/tests/tests.py
@@ -344,6 +344,24 @@ class Test404(TestCase, WagtailTestUtils):
         self.assertRedirects(response, reverse('wagtailadmin_login') + '?next=/admin/sdfgdsfgdsfgsdf')
 
 
+class TestAdminURLAppendSlash(TestCase, WagtailTestUtils):
+    def setUp(self):
+        # Find root page
+        self.root_page = Page.objects.get(id=2)
+
+    @override_settings(APPEND_SLASH=True)
+    def test_return_correct_view_for_correct_url_without_ending_slash(self):
+        self.login()
+
+        # Remove the ending slash from URL
+        response = self.client.get(reverse('wagtailadmin_explore_root')[:-1], follow=True)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertTemplateUsed(response, 'wagtailadmin/pages/index.html')
+        self.assertEqual(Page.objects.get(id=1), response.context['parent_page'])
+        self.assertTrue(response.context['pages'].paginator.object_list.filter(id=self.root_page.id).exists())
+
+
 class TestRemoveStaleContentTypes(TestCase):
     def test_remove_stale_content_types_preserves_access_admin_permission(self):
         call_command('remove_stale_contenttypes', interactive=False)
--- a/wagtail/admin/tests/tests.py
+++ b/wagtail/admin/tests/tests.py
@@ -349,13 +349,14 @@ class TestAdminURLAppendSlash(TestCase, WagtailTestUtils):
         # Find root page
         self.root_page = Page.objects.get(id=2)
 
-    @override_settings(APPEND_SLASH=True)
     def test_return_correct_view_for_correct_url_without_ending_slash(self):
         self.login()
 
+        # APPEND_SLASH is True by default
         # Remove the ending slash from URL
         response = self.client.get(reverse('wagtailadmin_explore_root')[:-1], follow=True)
 
+        # Check that correct page is returned after CommonMiddleware redirect
         self.assertEqual(response.status_code, 200)
         self.assertTemplateUsed(response, 'wagtailadmin/pages/index.html')
         self.assertEqual(Page.objects.get(id=1), response.context['parent_page'])
--- a/wagtail/admin/tests/tests.py
+++ b/wagtail/admin/tests/tests.py
@@ -355,7 +355,7 @@ class TestAdminURLAppendSlash(TestCase, WagtailTestUtils):
         self.login()
 
         # APPEND_SLASH is True by default
-        # Remove the ending slash from URL
+        # Remove trailing slash from URL
         response = self.client.get(reverse('wagtailadmin_explore_root')[:-1], follow=True)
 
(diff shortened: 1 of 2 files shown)


wagtail/admin/tests/tests.py
```diff
@@ -333,15 +333,36 @@ def test_no_permissions(self):
...
         self.login()
-        response = self.client.get('/admin/sdfgdsfgdsfgsdf')
+        response = self.client.get('/admin/sdfgdsfgdsfgsdf', follow=True)
+
+        # Check 400 error after CommonMiddleware redirect
         self.assertEqual(response.status_code, 404)
         self.assertTemplateUsed(response, 'wagtailadmin/404.html')
 
     def test_not_logged_in_redirect(self):
-        response = self.client.get('/admin/sdfgdsfgdsfgsdf')
```
Tijani-Dia, line 341: Otherwise it redirects first to `/admin/sdfgdsfgdsfgsdf/` and then `reverse('wagtailadmin_login') + '?next=/admin/sdfgdsfgdsfgsdf/`

wagtail/admin/tests/tests.py
```diff
@@ -333,15 +333,36 @@ def test_no_permissions(self):
 class Test404(TestCase, WagtailTestUtils):
     def test_admin_404_template_used(self):
         self.login()
-        response = self.client.get('/admin/sdfgdsfgdsfgsdf')
+        response = self.client.get('/admin/sdfgdsfgdsfgsdf', follow=True)
+
+        # Check 400 error after CommonMiddleware redirect
```
Stormheg, line 338: Suggestion: I think 400 should be 404 as that's what we are checking for.

wagtail/admin/tests/tests.py
```diff
@@ -333,15 +333,36 @@ def test_no_permissions(self):
...
+    def setUp(self):
+        # Find root page
+        self.root_page = Page.objects.get(id=2)
+
+    def test_return_correct_view_for_correct_url_without_ending_slash(self):
+        self.login()
+
+        # APPEND_SLASH is True by default
+        # Remove trailing slash from URL
+        response = self.client.get(reverse('wagtailadmin_explore_root')[:-1], follow=True)
```
Stormheg, line 359: Suggestion: could we explicitly set `APPEND_SLASH`  instead of assuming that it is set to true? I feel like that would make this test more robust.

Overriding settings in Django's TestCase can be done using the `settings()` context manager. [See overriding settings in the Django docs](https://docs.djangoproject.com/en/3.2/topics/testing/tools/#overriding-settings).

It should look somewhat like this:

```python
with self.settings(APPEND_SLASH=True):
    response = self.client.get(reverse('wagtailadmin_explore_root')[:-1], follow=True)
```

wagtail/admin/tests/tests.py
```diff
@@ -333,15 +333,36 @@ def test_no_permissions(self):
 class Test404(TestCase, WagtailTestUtils):
     def test_admin_404_template_used(self):
```
Stormheg, line 334: Suggestion: If I'm understanding this correctly, it looks like we are only testing the 404 view using the default settings (`APPEND_SLASH = True`).

However, depending on the value of `APPEND_SLASH` we use a slightly different urlconfig. Can we add a test case for `APPEND_SLASH = False` as well? Even though it seems rather insignificant it adds to the test coverage. See my comment below on mocking settings for more information on how to do this.

Stormheg (commented):
Thank you for fixing this @Tijani-Dia! :+1: 

I can confirm that the issue reported in #5331 is resolved with your fix :tada: 
My only request would be to add an additional test case for when `APPEND_SLASH = False`

Good work - keep it coming!

wagtail/admin/tests/tests.py
Tijani-Dia: Thanks for the feedback @Stormheg!
So trying to do this, I faced the same issue I mentioned in the following [comment](https://github.com/wagtail/wagtail/issues/5814#issuecomment-830854643).
I will put the [link to Django docs](https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#use-of-django-conf-settings) which explains why:

> Modules should not in general use settings stored in django.conf.settings at the top level 

And that's what we are doing here; which makes mocking settings harder.

So I tried to write a test but it fails - even tough we have the expected behaviour when running a Wagtail project.
I inspected the `tried` argument provided by ResolverMatch and found that the URL_CONF is set with APPEND_SLASH=True.
I forced the APPEND_SLASH setting to be False in my local  django and the test passes.

Should we rewrite it considering django's advice to use:

> a level of laziness or indirection ..., such as django.utils.functional.LazyObject, django.utils.functional.lazy() or lambda.?

Or is there a way i can rewrite the test ?

Stormheg: Thank you for your explanation @Tijani-Dia. I did not consider that when writing the suggestion. I'm afraid I don't know of an any easy way to work around this issue so I'd be happy to accept it as-is.

Of course you're free to try but I don't think it's worth the effort. Not sure how other core team members feel about this though - calling for @gasman's expert advice here :wink:. Maybe reloading the urlconfig during the test to make sure `settings.APPEND_SLASH` is reevaluated might work? But that sounds like quite a hassle...

Tijani-Dia: Alright! I made the changes you suggested, waiting for the expert's advice!

gasman: The only non-hacky way I can think of is to add a separate test run for the APPEND_SLASH=False setup, a bit like we do for [testing custom user models and USE_TZ=False](https://github.com/wagtail/wagtail/blob/9d29ba862c94a81d6cddb4010b179c6fce450799/.github/workflows/test.yml#L20-L32), and that seems like overkill here. I agree, it's not worth the hassle - I'm OK with leaving the APPEND_SLASH=False case untested.

Stormheg: Thank you both for your input! Marking this as resolved :heavy_check_mark:

Tijani-Dia:
Wrong manipulation!

Stormheg (approved):
LGTM

PR #7137, opened 2021-04-29
Audit logging for commenting

kaedroho:


wagtail/admin/tests/pages/test_edit_page.py
```diff
@@ -2344,14 +2395,90 @@ def test_new_reply(self):
...
+            'comments-0-position': '',
+            'comments-0-replies-TOTAL_FORMS': '1',
+            'comments-0-replies-INITIAL_FORMS': '1',
+            'comments-0-replies-MIN_NUM_FORMS': '0',
+            'comments-0-replies-MAX_NUM_FORMS': '',
+            'comments-0-replies-0-id': str(reply.id),
+            'comments-0-replies-0-text': 'an edited reply',
+        }
+
+        with enable_comment_panel(SimplePage):
```
jacobtoppm, line 2445: Do you need `enable_comment_panel` now we don't require the comments data in the formset? Could we just have the comments panel inherited via `settings_panels` as standard?

wagtail/admin/views/pages/edit.py
```diff
@@ -161,49 +166,49 @@ def send_commenting_notifications(self):
...
         """
-        # Skip if this page does not have CommentPanel enabled
-        if 'comments' not in self.form.formsets:
-            return
-
-        new_comments, deleted_comments, resolved_comments, edited_comments, new_replies, deleted_replies, edited_replies = self.get_commenting_changes()
-
         # Skip if no changes were made
-        if not new_comments and not deleted_comments and not resolved_comments and not edited_comments and not new_replies and not deleted_replies and not edited_replies:
+        if (not changes['new_comments']
```
jacobtoppm, line 201: This is more explicit, but quite long. If you don't think the explicitness is really needed, we could shorten to a simple `not any(changes.values())`?

wagtail/admin/views/pages/edit.py
```diff
@@ -647,8 +656,10 @@ def submit_action(self):
             previous_revision=(self.previous_revision if self.is_reverting else None)
         )
 
-        self.log_commenting_changes(revision)
-        self.send_commenting_notifications()
+        if 'comments' in self.form.formsets:
```
jacobtoppm, line 583: We save the page in a few other places, including `perform_workflow_action`, but also on workflow cancel/restart. We should make sure changes are logged and notifications sent there too (and probably consolidate some of this)

jacobtoppm (changes requested):
Generally looking great! A couple of possible tweaks, and one case of us not audit logging everywhere we save

wagtail/admin/views/pages/edit.py
kaedroho: Thanks! Just realised that this check is completely unnecessary anyway as all the for loops will just skip.

kaedroho:
@jacobtoppm ready for another review now!

jacobtoppm (approved):
LGTM! I think it would probably be worth us consolidating that revision saving + notification sending logic in one place in the future, but doesn't need to be here - we could also add the "check if it has changes before creating a new revision" to all of these methods as well, as we've been holding off on that since 2.10 and it would cut down on excess revisions.

PR #7138, opened 2021-04-29
Fix usage of html syntax instead of html+django

fabienheureux:
~The documentation currently use `html` for code blocks.~
~I noticed using `jinja` makes it a lot easier to read.~ 

EDIT: see https://github.com/wagtail/wagtail/pull/7138#issuecomment-829885447

### DIFF
--- a/docs/advanced_topics/customisation/extending_draftail.rst
+++ b/docs/advanced_topics/customisation/extending_draftail.rst
@@ -304,7 +304,7 @@ Finally, we register the JS components of our plugin:
 
 And that’s it! All of this setup will finally produce the following HTML on the site’s front-end:
 
-.. code-block:: html
+.. code-block:: html+django
 
     <p>
         Anyone following Elon Musk’s <span data-stock="TSLA">$TSLA</span> should also look into <span data-stock="BTC">$BTC</span>.
--- a/docs/advanced_topics/customisation/extending_draftail.rst
+++ b/docs/advanced_topics/customisation/extending_draftail.rst
@@ -304,7 +304,7 @@ Finally, we register the JS components of our plugin:
 
 And that’s it! All of this setup will finally produce the following HTML on the site’s front-end:
 
-.. code-block:: jinja
+.. code-block:: html
 
     <p>
         Anyone following Elon Musk’s <span data-stock="TSLA">$TSLA</span> should also look into <span data-stock="BTC">$BTC</span>.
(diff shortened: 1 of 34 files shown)


gasman:
jinja2 is not the same syntax as Django templates, so this seems risky. Can you give an example of a place in the docs where `jinja+html` is an improvement over `django+html`?

fabienheureux:
Sorry these changes were a bit brutal, what I intended to improve is syntax highlighting for example on this page: https://docs.wagtail.io/en/latest/advanced_topics/adding_reports.html

image
vs
image

It appears that it had nothing to do with `html+django` code blocks but places where `html` syntax was used instead of `html+django` 

I have reverted my first commit and committed a new one that should make more sense.

docs/advanced_topics/customisation/extending_draftail.rst
```diff
@@ -304,7 +304,7 @@ Finally, we register the JS components of our plugin:
 
 And that’s it! All of this setup will finally produce the following HTML on the site’s front-end:
 
-.. code-block:: html
+.. code-block:: html+django
```
fabienheureux, line 307: This block is only `html`, should I revert this ?

gasman: Yes, we should only use `html+django` for template code - this is showing the output of the template, so `html` is correct.

docs/reference/contrib/settings.rst
```diff
@@ -205,21 +205,21 @@ Add ``wagtail.contrib.settings.jinja2tags.settings`` extension to your Jinja2 se
 
 Then access the settings through the ``settings()`` template function:
 
-.. code-block:: html+jinja
+.. code-block:: html+django
```
gasman, line 208: `html+jinja` is correct for this section - it's talking about jinja2 template tags.

gasman (changes requested):

docs/reference/contrib/settings.rst
```diff
@@ -213,13 +213,13 @@ Then access the settings through the ``settings()`` template function:
 
 This will look for a ``request`` variable in the template context, and find the correct site to use from that. If for some reason you do not have a ``request`` available, you can instead use the settings defined for the default site:
 
-.. code-block:: html+jinja
+.. code-block:: html+django
```
gasman, line 216: This and the two below should also be reverted to `html+jinja`.

gasman (changes requested):

docs/reference/contrib/settings.rst
```diff
@@ -228,7 +228,7 @@ You can store the settings instance in a variable to save some typing, if you ha
 
 Or, alternately, using the ``set`` tag:
 
-.. code-block:: html+jinja
+.. code-block:: html+django
```
gasman, line 231: This should stay as jinja too

gasman (changes requested):

gasman (approved):
Thanks @fabienheureux!

PR #7142, opened 2021-04-30
Update accessibility contribution guidance in bug report and PR templates

thibaudcolas:
Following a @wagtail/accessibility team discussion. This is to reflect our [accessibility support targets](https://docs.wagtail.io/en/stable/contributing/developing.html#accessibility-targets) as introduced a year or so ago. Just like for browser support, we need to ensure every code change to Wagtail is appropriately tested for compatibility issues.

Tagging @wagtail/wagtail-core-team as well because it does mean we expect the core team all reviewers with merge rights to either:

- Ensure the change has been appropriately tested, and do all additional testing required before merging
- Get in touch with @wagtail/accessibility for anything that requires particular attention / expertise

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.md
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.md
@@ -30,4 +30,4 @@ Any other relevant information. For example, why do you consider this a bug and 
 * Python version: Run `python --version`.
 * Django version: Look in your requirements.txt, or run `pip show django | grep Version`.
 * Wagtail version: Look at the bottom of the Settings menu in the Wagtail admin, or run `pip show wagtail | grep Version:`.
-* Browser version: You can use https://www.whatsmybrowser.org/ to find this out.
+* Browser / operating system / assistive technology versions: You can use https://www.whatsmybrowser.org/ to find this out.
--- a/.github/ISSUE_TEMPLATE/BUG_REPORT.md
+++ b/.github/ISSUE_TEMPLATE/BUG_REPORT.md
@@ -30,4 +30,4 @@ Any other relevant information. For example, why do you consider this a bug and 
 * Python version: Run `python --version`.
 * Django version: Look in your requirements.txt, or run `pip show django | grep Version`.
 * Wagtail version: Look at the bottom of the Settings menu in the Wagtail admin, or run `pip show wagtail | grep Version:`.
-* Browser / operating system / assistive technology versions: You can use https://www.whatsmybrowser.org/ to find out more about your browser environment.
+* Browser version: You can use https://www.whatsmybrowser.org/ to find this out.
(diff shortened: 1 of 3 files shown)


thibaudcolas:
I’ve made a few tweaks based on today’s discussion.

Scotchester (approved):

.github/ISSUE_TEMPLATE/BUG_REPORT.md
```diff
@@ -30,4 +30,4 @@ Any other relevant information. For example, why do you consider this a bug and
 * Python version: Run `python --version`.
 * Django version: Look in your requirements.txt, or run `pip show django | grep Version`.
 * Wagtail version: Look at the bottom of the Settings menu in the Wagtail admin, or run `pip show wagtail | grep Version:`.
-* Browser version: You can use https://www.whatsmybrowser.org/ to find this out.
+* Browser / operating system / assistive technology versions: You can use https://www.whatsmybrowser.org/ to find out more about your browser environment.
```
Stormheg, line 33: How about this? 

Make it clearer that we only expect the bug reporter to include the assistive technology if necessary. I honestly don’t expect many people reporting issues with accessibility.

```suggestion
* Browser / operating system versions: You can use https://www.whatsmybrowser.org/ to find out more about your browser environment.
* Assistive technology: please include this if your bug report is about an accessibility issue. Examples of assistive technology: NVDA 2020.4, VoiceOver, Windows Magnifier...
```

Stormheg (approved):
One minor suggestion that might be helpful. I’ll leave it up to you whether it’s worth to include it 👍

.github/ISSUE_TEMPLATE/BUG_REPORT.md
thibaudcolas: We only have a single bug report template for all issues so I’m hesitant to lengthen it that much for something that will only be useful in a fraction (<1%) of the issues. Let’s see what others think though.

thibaudcolas:
Thank you both! I’ll leave this open at least over the weekend to give others a chance to take a look.

.github/ISSUE_TEMPLATE/BUG_REPORT.md
loicteixeira: On one hand, longer "forms" tend to be more daunting, so the fewer fields the better.

On the other hand, whatsmybrowser.org does not help in finding the version of the assistive technology being used. Some users might not even know what assistive technology are, get confused and skip the field entirely.

So I guess I would be in favour of splitting the fields.

However, if we don't have an easy and universal way of helping them find what software and version they use, I would skip the whole help text and simply use `Assistive technology software (and version) if applicable:`.

Accessibility still not seeing the integration it should, if someone requires assistive technologies, they have to seek it, so I imagine they would at least know the name of the software. But don't take my word for it, I really have no idea.

kbayliss: Out of the two options, I'd favour separating the assistive tech into a separate question for clarity, though I'm unsure about including it at all...

Is the lack of this question currently causing a problem? If we know accessibility issues account for less than 1% of all issues raised, then we'd be lengthening the process for 99% of issues by asking it (in either format). Let's ask for it if/when it gets missed, and re-evaluate this if it becomes burdensome?

PR #7150, opened 2021-05-04
Restructure reports views/filters

kaedroho:
- Moves the non-reusable reports filters into the views files, to co-locate the code better and reduce clutter in admin/filters.py.
- Splits the reports into multiple files to make room for more reports to be added and make it easier to find all the code for a particular report.


jacobtoppm (approved):
Looks like a sensible refactor to me!

wagtail/admin/views/pages/history.py
```diff
@@ -8,12 +11,32 @@
...
 from wagtail.admin.auth import user_has_any_page_permission, user_passes_test
-from wagtail.admin.filters import PageHistoryReportFilterSet
+from wagtail.admin.filters import DateRangePickerWidget, WagtailFilterSet
 from wagtail.admin.views.reports import ReportView
+from wagtail.core.log_actions import page_log_action_registry
 from wagtail.core.models import (
     Page, PageLogEntry, PageRevision, TaskState, UserPagePermissionsProxy, WorkflowState)
 
 
+def get_audit_log_users_queryset(request):
```
zerolab, line 21: question: this is also defined in `wagtail/admin/views/reports/audit_logging.py`. is that intentional?

kaedroho: Didn't actually notice that, thanks for spotting!

This should probably be moved to a method on ``PageLogEntry``?

zerolab: That is a sensible suggestion!

kaedroho:
@jacobtoppm @zerolab Please could either of you review the final commit?

jacobtoppm (approved):
Final commit LGTM

zerolab (approved):
LGTM too.
Note that this could use a note in the upgrade considerations, in case someone was importing any of reports/filter sets to extend/build on

gasman:
> Note that this could use a note in the upgrade considerations, in case someone was importing any of reports/filter sets to extend/build on

I'd disagree here... the class definitions that have moved are clearly specific to Wagtail's own reports and are outside of [the documented reports API](https://docs.wagtail.io/en/stable/advanced_topics/adding_reports.html), so there's no reasonable expectation for them to be supported in user code - it's very much "use at your own risk".

I think it's good for us to get out of the habit of mentioning internal API changes in upgrade considerations "just in case", because 1) it creates a TL;DR effect that makes people skip over the more important notes, and 2) it makes us overly cautious about important code refactoring, which means we end up not moving as fast as we should on that.

zerolab:
> I think it's good for us to get out of the habit of mentioning internal API changes in upgrade considerations "just in case", because 1) it creates a TL;DR effect that makes people skip over the more important notes, and 2) it makes us overly cautious about important code refactoring, which means we end up not moving as fast as we should on that.

Cannot disagree with that. Thanks @gasman

PR #7152, opened 2021-05-04
Prevent StreamField initialisation from being broken by colliding HTML IDs

### LINKED ISSUE #7146: StreamFields named 'main' are broken in page editor
Starting from a new 2.13rc3 project, the code at https://gist.github.com/hazho/e43007ab966e06b2365c7ba84528f420 fails to render the StreamField when editing the homepage.

This is because the call to `initBlockWidget` identifies the StreamField placeholder div by id, and there's an element with `id="main"` elsewhere on the page. We should ensure that the top-level StreamField element is given a unique prefix/suffix, or use a more specific selector to retrieve it (e.g. give it a `data-streamfield` attribute and select on both that and ID).

gasman:
Fixes #7146 - the root StreamField element is now matched by ID _and_ having a data-block attribute. (Duplicate IDs are invalid HTML and the 'proper' fix would be to add a global prefix to form fields, particularly since this problem isn't necessarily specific to StreamFields - but that would be a bigger breaking change.)

### DIFF
--- a/client/src/entrypoints/admin/telepath/blocks.js
+++ b/client/src/entrypoints/admin/telepath/blocks.js
@@ -43,7 +43,7 @@ function initBlockWidget(id) {
     data-value: JSON-encoded value for this block
   */
 
-  const body = document.getElementById(id);
+  const body = document.querySelector('#' + id + '[data-block]');
 
   // unpack the block definition and value
   const blockDefData = JSON.parse(body.dataset.block);


kaedroho (approved):

client/src/entrypoints/admin/telepath/blocks.js
```diff
@@ -43,7 +43,7 @@ function initBlockWidget(id) {
     data-value: JSON-encoded value for this block
   */
 
-  const body = document.getElementById(id);
+  const body = document.querySelector('#' + id + '[data-block]');
```
kaedroho, line 46: I think it's still invalid HTML to have two elements with the same ID, so we should make it use a different ID for the long term

zerolab: It is, as per (all and latest) spec https://www.w3.org/TR/html51/dom.html#the-id-attribute
Also things like `getElementById` and `querySelector` are specced to return the first one https://dom.spec.whatwg.org/#dom-nonelementparentnode-getelementbyid / https://dom.spec.whatwg.org/#dom-parentnode-queryselector

PR #7156, opened 2021-05-04
Add titlesonly to reference / topics top-level TOCs

gasman:
This prevents duplicated headings in places where the second-level index page has intro blurbs for each subsection (and also means we're not listing out long multi-section pages in full in the index, but given how big these indexes are that's probably not a bad thing).

topics (before):
Screenshot 2021-05-04 at 16 10 08

reference (before):
Screenshot 2021-05-04 at 16 10 27

### DIFF
--- a/docs/reference/index.rst
+++ b/docs/reference/index.rst
@@ -4,6 +4,7 @@ Reference
 
 .. toctree::
     :maxdepth: 2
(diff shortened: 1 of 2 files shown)


zerolab (approved):
LGTM!

PR #7157, opened 2021-05-04
Add filter to hide all commenting actions from history

kaedroho:


wagtail/admin/views/reports/audit_logging.py
```diff
@@ -15,15 +16,27 @@
 
 class SiteHistoryReportFilterSet(WagtailFilterSet):
     action = django_filters.ChoiceFilter(choices=page_log_action_registry.get_choices)
+    hide_commenting_actions = django_filters.BooleanFilter(
```
jacobtoppm, line 19: The two filtersets here seem to share pretty much everything but the `label` filter - do you think it would be worth refactoring out the common bit/ making `SiteHistoryReportFilterSet` inherit from the other to avoid repetition?

kaedroho: I'd rather we duplicated them than coupled two separate view modules together. In this case though, it looks like the two modules do basically the same thing at different levels, so I wonder if it makes sense to merge them both into one `history.py` that contains both views?

jacobtoppm: Yes, I think merging makes sense here - feels like they're doing almost exactly the same thing.

zerolab: The Page and Site history reports are both audit logs, just at a different granularity level. so either `history.py` or `audit_logging.py` works

jacobtoppm (approved):

PR #7158, opened 2021-05-05
Change 'iff' to 'if and only if' to avoid confusion.

kevinhowbrook:


gasman (changes requested):
'iff' means 'if and only if'. I think you're not the first person to be confused by this, so I'd be happy to accept a PR to change it to 'if and only if' :-)

gasman (approved):
Thanks @kevinhowbrook!

PR #7169, opened 2021-05-07
Prevent JS error on URL hashes that are not tabs

gasman:
Fixes a bug introduced in e0d4381661ea81394b3558cf33a4a2f18b42d9d2.

The tab code looks for the presence of a URL hash and calls showTab on a link with that href, without checking that the link A) actually exists and B) is a tab as identified by the data-tab attribute. As a result, non-tab-related uses of URL hashes (such as simply visiting `/admin/#foo`) throw a JS error.

### DIFF
--- a/client/src/entrypoints/admin/core.js
+++ b/client/src/entrypoints/admin/core.js
@@ -308,8 +308,10 @@ $(() => {
   };
 
   if (window.location.hash) {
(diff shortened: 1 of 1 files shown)


jacobtoppm (approved):
Thanks for catching this! This fix looks good to me

PR #7172, opened 2021-05-10
[prototype] sidebar rewrite in react

kaedroho:
Not for review, just posting here to have something to link other devs to

This will be broken down, polished and submitted in small chunks


kaedroho:
All separated out into separate PRs now

PR #7174, opened 2021-05-10
[slim-sidebar 2]  Initial frontend set up

kaedroho:
This PR adds empty JS and CSS entrypoints for the new sidebar, moves the legacy sidebar JS into a separate file and introduces a `WAGTAIL_EXPERIMENTAL_FEATURES` setting.


client/src/index.test.js
```diff
@@ -34,16 +34,4 @@ describe('wagtail package API', () => {
   it('has Transition', () => {
     expect(Transition).toBeDefined();
   });
-
```
fabienheureux, line 37: `Explorer` `ExplorerToggle` and `initExplorer` imports are not needed anymore in this file

fabienheureux (changes requested):

client/src/index.test.js
kaedroho: Oops, fixed!

kaedroho:
Stacked this on top of https://github.com/wagtail/wagtail/pull/7181 so that my next PR can use both

wagtail/admin/templatetags/wagtailadmin_tags.py
```diff
@@ -673,6 +673,11 @@ def locales():
     ])
 
 
+@register.simple_tag()
+def slim_sidebar_enabled():
```
jacobtoppm, line 677: Maybe you could reuse this function in the if statement for the other two tags too (just a quick amendment to avoid repeating logic, `if slim_sidebar_enabled()` reads okay to me)?

jacobtoppm (approved):
LGTM! Just one tweak suggested

wagtail/admin/templatetags/wagtailadmin_tags.py
kaedroho: Good idea, thanks!

PR #7179, opened 2021-05-11
Remove redundant badges from readme

gasman:
* Codecov is only tracking JS coverage, so displaying a coverage figure for the whole project is misleading
* wagtail-slack.now.sh appears to be dead


kaedroho (approved):

kaedroho:
Thanks!

PR #7180, opened 2021-05-11
Added full_url property for in AbstractRendition model.

### LINKED ISSUE #6803: Add a full_url property to image rendition objects

To include a Twitter 'share' image on a template, you'd typically do something like:

    {% image page.social_image fill-1200x1200 as social_img %}
    <meta name="twitter:image" content="{{ social_img.url }}">

However, this may return either a local URL (`/media/images/foo.jpg`) or an absolute URL (`https://media.example.com/images/foo.jpg`) depending on how media files are configured at the Django level. There's no good way to consistently return a full absolute URL, as required for the 'share' image to work.

Add a `full_url` property to the image Rendition model, which prepends `settings.BASE_URL` to the rendition's `url` property if it begins with a `/`, and returns the url unchanged if not. This will then allow us to write

    <meta name="twitter:image" content="{{ social_img.full_url }}">

(Using `settings.BASE_URL` isn't perfect - it's only an informal convention for Wagtail sites to define it, and it doesn't account for multi-site installations - but it's probably good enough for most purposes.)

shreyashsrivastava:
Resolves #6803

* Do the tests still pass? 
   Yes

   Yes

* For Python changes: Have you added tests to cover the new/fixed behavior?
   Yes, added the tests for the same, but it might need review.

   Yes, I updated the documentation accordingly.

### DIFF
--- a/docs/topics/images.rst
+++ b/docs/topics/images.rst
@@ -187,7 +187,7 @@ Wagtail can assign the image data to another variable using Django's ``as`` synt
 
     {% image page.photo width-400 as tmp_photo %}
 
-    <img src="{{ tmp_photo.url }}" width="{{ tmp_photo.width }}"
+    <img src="{{ tmp_photo.full_url }}" width="{{ tmp_photo.width }}"
         height="{{ tmp_photo.height }}" alt="{{ tmp_photo.alt }}" class="my-custom-class" />
 
 
@@ -201,7 +201,7 @@ Therefore, if you'd added the field ``author`` to your AbstractImage in the abov
 
 
 .. Note::
-    The image property used for the ``src`` attribute is actually ``image.url``, not ``image.src``.
+    The image property used for the ``src`` attribute is actually ``image.full_url``, not ``image.src``.
 
 
 The ``attrs`` shortcut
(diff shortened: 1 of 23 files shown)


shreyashsrivastava:
Hey @gasman, can you please review this pr.

docs/topics/images.rst
```diff
@@ -201,7 +201,7 @@ Therefore, if you'd added the field ``author`` to your AbstractImage in the abov
 
 
 .. Note::
-    The image property used for the ``src`` attribute is actually ``image.url``, not ``image.src``.
+    The image property used for the ``src`` attribute is actually ``image.full_url``, not ``image.src``.
```
gasman, line 204: The `full_url` property isn't intended to replace `url` - it's provided as an alternative, and `url` will still be preferred for images shown on page - so it'd better to add a new section covering `full_url`.

gasman (commented):
Thanks @shreyashsrivastava! Just a couple of minor issues which I'll deal with now.

gasman:
Merged in 08be45dbd9f6b110101eee0183ac6eb02608dda1 + parents.

PR #7185, opened 2021-05-13
Update eslint config

kaedroho:
Updated to the latest version of ``wagtail-eslint-config``.

Turns out we were quite behind so there were a lot of peer dependencies to update as well.


.eslintrc
```diff
@@ -4,7 +4,7 @@
     "@typescript-eslint"
   ],
   "extends": [
-    "wagtail",
+    "@wagtail/eslint-config-wagtail",
```
kaedroho, line 7: Note: the plugin isn't working as documented: https://www.npmjs.com/package/@wagtail/eslint-config-wagtail

kaedroho:
@gasman also had trouble with updating it https://github.com/wagtail/wagtail/pull/7247 and I can't find someone on slack who can answer questions about this set-up.

Closing as I think the better option here would be to rethink this setup. I'd like to see if we can remove the dependency on ``wagtail-eslint-config`` altogether and replace it with something standard.

PR #7191, opened 2021-05-16
[slim-sidebar 4] Allow footer text to be set on sub menus

kaedroho:
This adds a ``footer_text`` attribute to ``SubMenuItem``, which allows adding text to the bottom of any sub-menu. This removes the need for a separate ``SettingsSubMenuItem`` type

### DIFF
--- a/client/src/components/Sidebar/menu/SubMenuItem.tsx
+++ b/client/src/components/Sidebar/menu/SubMenuItem.tsx
@@ -88,6 +88,7 @@ export const SubMenuItem: React.FunctionComponent<SubMenuItemProps> = (
         <ul aria-labelledby={`wagtail-sidebar-submenu${path.split('.').join('-')}-title`}>
           {renderMenu(path, item.menuItems, slim, state, dispatch, navigate)}
         </ul>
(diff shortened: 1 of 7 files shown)


client/src/components/Sidebar/menu/SubMenuItem.tsx
```diff
@@ -88,6 +88,7 @@ export const SubMenuItem: React.FunctionComponent<SubMenuItemProps> = (
         <ul aria-labelledby={`wagtail-sidebar-submenu${path.split('.').join('-')}-title`}>
           {renderMenu(path, item.menuItems, slim, state, dispatch, navigate)}
         </ul>
+        {item.footerText && <p className="sidebar-sub-menu-panel__footer">{item.footerText}</p>}
```
zerolab, line 91: question: should there be a limit for the footer text? how would this look with a long phrase?

gasman (approved):

client/src/components/Sidebar/menu/SubMenuItem.tsx
kaedroho: It should wrap on to multiple lines if it's too long

PR #7192, opened 2021-05-16
[slim-sidebar 5] Page explorer

kaedroho:
This PR makes a copy of the existing page explorer for the new UI.


Stormheg (approved):
Thanks @kaedroho 🚀 

You brought the page explorer back from the dead 😄 
There are some deprecation warnings thrown by react regarding calls to `componentWillReceiveProps` and `componentWillMount ` but we'll leave that for cleanup later.

There's currently an issue with the z-index of the sidebar but that should be fixed by https://github.com/wagtail/wagtail/pull/7265

PR #7198, opened 2021-05-18
WIP: Add SVG icon sprite into storybook

kaedroho:
TODO: Find a way to extract this from Django without committing it to the repo


allcaps:
Maybe it is easiest to have a Django management command that outputs the `client/src/components/SVGIconSprite/index.tsx` file or only update the `const spriteHtml =` lines.

Whenever icons are updated, the developer should run the management command to update the storybook and commit.

It might not be ideal, but workable.

thibaudcolas:
Superseded by #8025.

PR #7208, opened 2021-05-24
Reorganise snippet admin URLs to avoid ambiguous paths

gasman:
Snippet admin views allow for arbitrary strings as primary keys, and the current URL patterns don't adequately namespace these from reserved words like 'add' - for example, a snippet with the primary key 'add' would have an edit URL that collides with the add view at `/admin/snippets/foo/bar/add/`.

This is unlikely to come up in practice, but it does mean that our urlconf is more sensitive to ordering than it needs to be. Rearrange so that the verb (add, edit, delete) consistently comes before the pk, and add redirects to handle the legacy URLs.


Stormheg (approved):
LGTM - thanks @gasman! 👍

Stormheg:
Rebased and release notes added. Will merge as soon as the CI passes.

PR #7209, opened 2021-05-24
Eliminate request.is_ajax from snippet listing view

gasman:
Django 4.0 [is dropping `request.is_ajax()`](https://docs.djangoproject.com/en/3.2/releases/3.1/#id2), so it's worth us making a start on removing it from the Wagtail admin. The snippet listing view currently makes use of it for the header search, returning just the 'results' include if the request comes from AJAX - this PR turns that into a class-based view so that we can easily split off a dedicated URL endpoint for that instead.

Incorporates #7208.

### DIFF
--- a/wagtail/snippets/urls.py
+++ b/wagtail/snippets/urls.py
@@ -12,6 +12,7 @@ urlpatterns = [
     path('choose/<slug:app_label>/<slug:model_name>/<str:pk>/', chooser.chosen, name='chosen'),
 
     path('<slug:app_label>/<slug:model_name>/', snippets.ListView.as_view(), name='list'),
+    path('<slug:app_label>/<slug:model_name>/results/', snippets.ListResultsView.as_view(), name='list_results'),
     path('<slug:app_label>/<slug:model_name>/add/', snippets.create, name='add'),
     path('<slug:app_label>/<slug:model_name>/edit/<str:pk>/', snippets.edit, name='edit'),
     path('<slug:app_label>/<slug:model_name>/multiple/delete/', snippets.delete, name='delete-multiple'),
--- a/wagtail/snippets/urls.py
+++ b/wagtail/snippets/urls.py
@@ -12,7 +12,10 @@ urlpatterns = [
     path('choose/<slug:app_label>/<slug:model_name>/<str:pk>/', chooser.chosen, name='chosen'),
 
     path('<slug:app_label>/<slug:model_name>/', snippets.ListView.as_view(), name='list'),
-    path('<slug:app_label>/<slug:model_name>/results/', snippets.ListResultsView.as_view(), name='list_results'),
+    path(
+        '<slug:app_label>/<slug:model_name>/results/',
+        snippets.ListView.as_view(results_only=True), name='list_results'
+    ),
     path('<slug:app_label>/<slug:model_name>/add/', snippets.create, name='add'),
     path('<slug:app_label>/<slug:model_name>/edit/<str:pk>/', snippets.edit, name='edit'),
     path('<slug:app_label>/<slug:model_name>/multiple/delete/', snippets.delete, name='delete-multiple'),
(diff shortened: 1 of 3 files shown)


wagtail/snippets/urls.py
```diff
@@ -11,10 +11,17 @@
     path('choose/<slug:app_label>/<slug:model_name>/', chooser.choose, name='choose'),
     path('choose/<slug:app_label>/<slug:model_name>/<str:pk>/', chooser.chosen, name='chosen'),
 
-    path('<slug:app_label>/<slug:model_name>/', snippets.list, name='list'),
+    path('<slug:app_label>/<slug:model_name>/', snippets.ListView.as_view(), name='list'),
+    path('<slug:app_label>/<slug:model_name>/results/', snippets.ListResultsView.as_view(), name='list_results'),
```
ababic, line 15: I'm just wondering if separate views are the best way to go here? It certainly makes things quite clear-cut, but it feels like a big departure. Perhaps we could reuse the same view class, but in a different 'mode'? e.g.

```
path('<slug:app_label>/<slug:model_name>/results/', snippets.ListView.as_view(results_only=True), name='list_results'),
```

wagtail/snippets/views/snippets.py
```diff
@@ -74,96 +75,107 @@ def index(request):
         raise PermissionDenied
 
 
-def list(request, app_label, model_name):
-    model = get_snippet_model_from_url_params(app_label, model_name)
+class BaseListView(TemplateView):
```
ababic, line 78: If we're going to go with generics here, is there any reason we don't go with `ListView` for this, or perhaps even our own [`wagtail.admin.views.generic.model.IndexView`](`https://github.com/wagtail/wagtail/blob/main/wagtail/admin/views/generic/models.py#L14`)?

It feels like a missed opportunity to take this step without breaking the code down into more digestible (and ultimately more overridable) chunks.

ababic (commented):
Thanks @gasman. Nothing blocking here per se, but just a couple of questions on approach.

wagtail/snippets/views/snippets.py
gasman: No, the missed opportunity would have been if I'd sat on this PR indefinitely because I didn't have several days free to do all the refactoring that could be done here :-) Incremental changes are the way to go (and more friendly to code reviewers).

There is indeed a lot more that could be done here, and I hope that in time this will eventually become a generic listing view. I think the best way to get to that point without breaking anything is to start from the lowest level (initially just moving the existing code into a `dispatch` method, not even splitting it into GET/POST cases) and progressing to higher levels from there.

wagtail/snippets/urls.py
gasman: I think the benefits of separate views will come further down the line once we split the logic out further into separate methods - the results-only view's `get_context` method can leave out a lot of the stuff that's only used for page furniture.

ababic: Okay, but it's looking to me like pretty much everything currently happening in `get_context_data()` is relevant to both views. All that really needs to change is the template.

wagtail/snippets/views/snippets.py
ababic: Okay, happy with that approach.

wagtail/snippets/urls.py
gasman: OK, that's fair... I was imagining that the code for handling the search form would drop out, but (for now at least) we still need the form object in order to validate the incoming URL parameter. So yep, the `can_add_snippet` permission check is probably the only real potential saving...

So it's really just a question of whether we do the template swap with two different view classes, or by checking a `results_only` parameter in a `get_template_names` method. I don't see a huge amount of difference there, but happy to go with the latter if that will make this mergeable :-)

ababic: I feel like the `results_only` option is 'ever-so-slightly' closer to the original solution. The concept of a view using a different template based on some value is a familiar one, and I can't think of a compelling reason to move away from it here. We're really splitting hairs by this point though :-) Go with whatever you're happy with.

gasman: OK - updated to use a `results_only` parameter in 286ee6d412e572650a92cb0f6f4ff1083c5cfa5d.

ababic (approved):
LGTM

PR #7212, opened 2021-05-25
Hide comment notifications checkbox when commenting is disabled

### LINKED ISSUE #7203: UI Issue - Empty checkbox is shown while customising tabbed interface
### Issue Summary

I've recently upgraded to 2.13. When customising the tabbed interface, if I commented out the settings panel from the edit handler, an empty checkbox (comment_notifications) will appear on the bottom left of the page.

### Steps to Reproduce

1. In any Page, customise the edit_handler and leave out settings_panels from the list:
```
        ObjectList(content_panels, heading='Content'),
        ObjectList(promote_panels, heading='Promote'),
        # ObjectList(Page.settings_panels, heading='Settings')
    ])
```
Screenshots are given below:

ss1
ss2

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: Yes

### Technical details

* Python version: 3.9.1
* Django version: 3.2.3
* Wagtail version: 2.13
* Browser version: Safari 14

kaedroho:
Fixes #7203

### DIFF
--- a/wagtail/admin/edit_handlers.py
+++ b/wagtail/admin/edit_handlers.py
@@ -2,6 +2,7 @@ import functools
 import re
 
 from django import forms
+from django.apps import apps
 from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.exceptions import FieldDoesNotExist, ImproperlyConfigured
@@ -373,6 +374,18 @@ class TabbedInterface(BaseFormEditHandler):
             self.show_comments_toggle = show_comments_toggle
         else:
             self.show_comments_toggle = 'comment_notifications' in self.required_fields()
+
+    def get_form_class(self):
+        form_class = super().get_form_class()
+
+        # Set show_comments_toggle attibute on form class
+        return type(
+            form_class.__name__,
+            (form_class, ),
+            {
+                'show_comments_toggle': self.show_comments_toggle
+            }
+        )
 
     def clone_kwargs(self):
         kwargs = super().clone_kwargs()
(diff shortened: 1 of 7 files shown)


gasman (changes requested):
Looks like the Javascript side also needs updating - with the checkbox removed, [the "missing HTML element" exception in comments.js](https://github.com/wagtail/wagtail/blob/main/client/src/entrypoints/admin/comments.js#L255) now gets thrown on page load.

kaedroho:
> Looks like the Javascript side also needs updating - with the checkbox removed, the "missing HTML element" exception in comments.js now gets thrown on page load.

This is fixed now, thanks!

gasman:
Some weirdness going on with tests here :-/ All seems fine on main, but the new tests fail on the stable/2.13.x branch, and only when running `./runtests.py wagtail.admin.tests.pages.test_edit_page` in full (not `wagtail.admin.tests.pages.test_edit_page.TestPageSubscriptionSettings` on its own).

gasman:
I think this is probably something to do with 19671c7073b92b23c1c7457422c89a91ee9e60e5. @kaedroho, what's the best way forward here? Is it possible to remove the `enable_comment_panel` hack on the 2.13.x branch too, or does that rely on other changes that are too big to backport (in which case we need to update the tests from this PR to play nicely with that hack instead, I guess...)?

gasman:
Looks like [removing the enable_comment_panel context manager](https://github.com/wagtail/wagtail/commit/d3309f927be41a24daffa8280d8bcdf47d06fc42) has sorted it (which makes sense now, because [the "let's not override panels on all of our test models" update](https://github.com/wagtail/wagtail/pull/7063) happened in time for 2.13).

PR #7213, opened 2021-05-25
Allow a `ModelAdmin` to handle root pages

RealOrangeOne:
Previously, `user_can_copy_obj` would raise an exception, as `parent_page` would be `None`. Now it handles that, and assumes it can copy.

Fixes an issue discovered in the [Wagtail Slack](https://wagtailcms.slack.com/archives/C81FGJR2S/p1621923985127700).

### DIFF
--- a/wagtail/contrib/modeladmin/options.py
+++ b/wagtail/contrib/modeladmin/options.py
@@ -240,6 +240,9 @@ class ModelAdmin(WagtailRegisterable):
         ordering = self.get_ordering(request)
         if ordering:
             qs = qs.order_by(*ordering)
+        if self.is_pagemodel:
+            # If we're listing pages, exclude the root page
+            qs = qs.exclude(depth=1)
         return qs
 
     def get_search_fields(self, request):
--- a/wagtail/contrib/modeladmin/tests/test_page_modeladmin.py
+++ b/wagtail/contrib/modeladmin/tests/test_page_modeladmin.py
@@ -54,6 +54,11 @@ class TestIndexView(TestCase, WagtailTestUtils):
         # There should still be four results
         self.assertEqual(response.context['result_count'], 4)
 
+    def test_using_core_page(self):
+        # The core page is slightly different to other pages
+        response = self.client.get('/admin/wagtailcore/page/')
+        self.assertEqual(response.status_code, 200)
+
 
 class TestExcludeFromExplorer(TestCase, WagtailTestUtils):
     fixtures = ['modeladmintest_test.json']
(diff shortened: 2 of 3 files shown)


wagtail/contrib/modeladmin/helpers/permission.py
```diff
@@ -162,4 +162,6 @@ def user_can_publish_obj(self, user, obj):
 
     def user_can_copy_obj(self, user, obj):
         parent_page = obj.get_parent()
+        if parent_page is None:
+            return True  # If there is no parent, assume they can copy it
```
RealOrangeOne, line 166: Meta: Is `True` correct here? It feels like it probably is, as making another root page won't really do anything unless it's also attached to a `Site`.

gasman:
Generally, the root page should not be accessible or editable as a normal page, except in limited circumstances such as setting permissions where "encompassing everything in the page tree" is a meaningful concept. For example, it's not counted as part of the page count shown on the admin homepage. It definitely shouldn't be possible to copy it - that's likely to break internal Wagtail business logic that assumes that only one root will exist (e.g. the aforementioned permission logic). In this case, I'd say the correct solution here would be for modeladmin to exclude the root page from its listing queries.

RealOrangeOne:
>  In this case, I'd say the correct solution here would be for modeladmin to exclude the root page from its listing queries.
Makes sense. Thoughts on how to implement? It's going to be a pretty special case anyway, but perhaps a custom `get_queryset` which checks if the model is a `Page`, and either emits a warning, or just does the filtering itself (the latter feels like something which would confuse people and just lean towards more error reports)?

gasman:
I'd just stick a line [in ModelAdmin.get_queryset](https://github.com/wagtail/wagtail/blob/4dc68550bcbc310bac0673005fd86869277b1379/wagtail/contrib/modeladmin/options.py#L243) saying `if self.is_pagemodel: qs = qs.filter(depth__gt=1)`. Omitting the root page from the listing is 99.9% certain to be the right thing to do, and if someone really wants the opposite, I'm happy to make them jump through an extra hoop.

wagtail/contrib/modeladmin/helpers/permission.py
RealOrangeOne: Resolved with change of approach

ababic:
@RealOrangeOne this is looking good, thank you :) would you mind just adjusting the test to be a bit more meaningful? For example, checking that the root page isn't in the results?

gasman:
Thanks @RealOrangeOne! Merged in d5b955b0e496bc997c65266b3518c7cce8fca6d7.

PR #7218, opened 2021-05-26
Allow workflows to be disabled globally

gasman:
Introduce a WAGTAIL_WORKFLOW_ENABLED setting; when false, the workflow report and settings menus are hidden, permissions are not registered, moderation-related dashboard panels on the homepage are skipped, workflow actions on add/edit page are hidden, and model methods such as page.current_workflow_state return None / False immediately without any db queries.

(Normally I'd be reluctant to use settings for this, and would prefer to just have the workflow UI 'disappear' when all workflows are disabled within the admin - however, I think the performance boost from eliminating the otherwise hard-to-avoid `current_workflow_state` queries as per #6878 makes it worthwhile.)

### DIFF
--- a/docs/reference/settings.rst
+++ b/docs/reference/settings.rst
@@ -714,6 +714,12 @@ Workflow
 
 .. code-block:: python
 
+  WAGTAIL_WORKFLOW_ENABLED = False
+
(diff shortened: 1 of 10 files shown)


jacobtoppm (approved):
This looks good to me! I do think we can still improve the `current_workflow_state` queries by caching/invalidating with a bit more care, but for sites that never want to use workflow, this seems like a nice performance improvement anyway.

gasman:
Merged in de9588590bbabffbbe30e8830d34b59d8da6512e

PR #7220, opened 2021-05-26
Avoid n+1 query on recent edits dashboard panel

gasman:
The use of Page.objects.in_bulk to retrieve all page data for the panel in a single query was being thwarted by our use of revision.page.id when reading them back (which ends up loading the whole page object just to look at the id...)

### DIFF
--- a/wagtail/admin/tests/pages/test_dashboard.py
+++ b/wagtail/admin/tests/pages/test_dashboard.py
@@ -90,3 +90,28 @@ class TestRecentEditsPanel(TestCase, WagtailTestUtils):
         self.assertEqual(panel.last_edits[0][0].page, Page.objects.get(pk=self.child_page.id))
         # check if the page in this list is the specific page of this revision
         self.assertEqual(panel.last_edits[0][1], Page.objects.get(pk=self.child_page.id).specific)
+
+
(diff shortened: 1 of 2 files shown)


wagtail/admin/tests/pages/test_dashboard.py
```diff
@@ -90,3 +90,28 @@ def test_panel(self):
...
+    def test_panel_query_count(self):
+        # fake a request object with bob as the user
+        self.client.user = self.bob
+        with self.assertNumQueries(4):
+            # Instantiating RecentEditsPanel should not generate N+1 queries -
+            # i.e. any number less than 6 would be reasonable here
+            panel = RecentEditsPanel(self.client)
+
+        # check that the panel is still actually returning results
+        html = panel.render()
```
kaedroho, line 116: Is it worth putting this in the with block to check if any n+1 queries occur in the template?

gasman: Would love to, if I could :-) However, there are indeed n+1 queries there (including [the notorious `current_workflow_state`](https://github.com/wagtail/wagtail/pull/6878)...), which are probably going to be tougher to crack.

kaedroho (approved):

zerolab (approved):
Excellent catch! 💪🏼

PR #7221, opened 2021-05-27
Check that page.save_revision() was given a specific page object

kaedroho:
Currently it's possible to call `.save_revision()` on the generic version of the page. This results in all the specific content being deleted.

This PR adds a check to make sure the specific page is always used.


zerolab:
@kaedroho looks like there a number of tests failing after this change. Other than that it is unclear what else needs working on

PR #7224, opened 2021-05-28
Fix ClusterTaggableManager copy regression in _copy_m2m_relations

cspollar:
This pull request fixes a regression that was introduced during the https://github.com/wagtail/wagtail/pull/6277 refactor. Currently copying a page that inherits from a page that inherits from a `TaggedPage` (a page with ClusterTaggableManager) fails with the following error.

`ValueError: Cannot add [] (<class 'modelcluster.queryset.FakeQuerySet'>). Expected <class 'django.db.models.base.ModelBase'> or str.` when copying a `SubBlogPage`.

This is because `_copy_m2m_relations` ([source](https://github.com/wagtail/wagtail/blob/7c6db207a8bad92d1a84facd933a27334fa8aa9f/wagtail/core/models.py#L127)) uses `source._meta.parents` instead of `source._meta.source._meta.get_parent_list()`.

----

* For front-end changes: Did you test on all of [Wagtail’s supported environments] (https://docs.wagtail.io/en/latest/contributing/developing.html#browser-and-device-support)? N/A


cspollar:
It may also be helpful to see a full broken example. I've created a [proof of concept ](https://github.com/cspollar/broken-wagtail-copy-demo)app that is unable to copy.

In this example trying to copy a `SubBlogPage` will raise the following error.

```
`ValueError: Cannot add [] (<class 'modelcluster.queryset.FakeQuerySet'>). Expected <class 'django.db.models.base.ModelBase'> or str.` when copying a `SubBlogPage`.
```

Once I switch to a patched version of wagtail that uses `source._meta.source._meta.get_parent_list()` all is well.

```
# 
from django.db import models
from modelcluster.contrib.taggit import ClusterTaggableManager
from taggit.models import TaggedItemBase


class BlogPageTag(TaggedItemBase):
    content_object = ParentalKey(
        Page, on_delete=models.CASCADE, related_name="tagged_items"
    )


class HomePage(Page):
    pass


class BlogPage(Page):
    tags = ClusterTaggableManager(through=BlogPageTag, blank=True)

    promote_panels = Page.promote_panels + [
        FieldPanel("tags"),
    ]


class SubBlogPage(BlogPage):
    pass
```

I'm happy to open an issue or make changes if needed. Thank you!

wagtail/core/models.py
```diff
@@ -124,7 +124,7 @@ def _copy_m2m_relations(source, target, exclude_fields=None, update_attrs=None):
         if field.many_to_many and field.name not in exclude_fields and not field.auto_created and not isinstance(field, ParentalManyToManyField):
             try:
                 # Do not copy m2m links with a through model that has a ParentalKey to the model being copied - these will be copied as child objects
-                through_model_parental_links = [field for field in field.through._meta.get_fields() if isinstance(field, ParentalKey) and (field.related_model == source.__class__ or field.related_model in source._meta.parents)]
+                through_model_parental_links = [field for field in field.through._meta.get_fields() if isinstance(field, ParentalKey) and (field.related_model == source.__class__ or field.related_model in source._meta.get_parent_list())]
```
ababic, line 127: This could be a red herring, but maybe we could simplify things slightly here?:

```suggestion
                through_model_parental_links = [field for field in field.through._meta.get_fields() if isinstance(field, ParentalKey) and issubclass(source.__class__, field.related_model)]
```

ababic (commented):
Hi @cspollar. Thank you very much for the thorough write-up and submission - this is high quality stuff :) I just have one suggestion, which might make for slightly easier reading.

cspollar:
> Hi @cspollar. Thank you very much for the thorough write-up and submission - this is high quality stuff :) I just have one suggestion, which might make for slightly easier reading.

Thanks @ababic your change makes sense to me (and the tests still pass).

ababic:
@cspollar FYI, I raised this in today's core team meeting to see where this can fit in the release cycle. Because this change has been in place for a few versions, it falls out of the usual pattern of deserving it's own release. However, it looks like there are a couple of other things we might want to release next week, and this may be able to slide in alongside those. 

Thanks again for such an excellent contribution! I'll let @gasman pick things up from here.

gasman (approved):

gasman:
Thanks @cspollar! This fix looks good to me.

It looks like the incorrect parent model logic was also present in the pre-#6277 version of this code, and your test case does indeed fail when run against the 2.10.x branch too, so it seems that this bug isn't a regression - it's basically been around forever... On that basis, this will go through our regular release cycle and be included in the 2.14 release, rather than being backported to existing releases (which are generally limited to regressions, in the interest of not introducing unintended changes).

gasman:
Merged in c9a55d8b1bf519f97fab32c854b4332f415b5176

cspollar:
@gasman Thanks for getting this merged. I upgraded from the 2.9.x branch (where the tests do pass). It looks like the regression was probably added in the 2.10.x branch. Regardless, I created a fork and will run from there until 2.14 is released. In the future I'll try to keep our version of Wagtail more up to date.

Issue #947, opened 2015-01-30
Modify image.get_rendition() to take advantage of prefetch_related


thenewguy:
The [Picturefill polyfill](http://scottjehl.github.io/picturefill/) makes it possible to output responsive img tags using srcset with cross browser support.

image.get_rendition() does not take advantage of [prefetch_related](https://docs.djangoproject.com/en/1.7/ref/models/querysets/#prefetch-related) because of the way the rendition lookup is written.  (ref: [models.py#L172](https://github.com/torchbox/wagtail/blob/master/wagtail/wagtailimages/models.py#L172))

Since a responsive img tag links to many renditions, rendering a lot of images greatly increases the number of database queries.  I.e. to output an image gallery with 30 images using 10 possible renditions per srcset suddenly takes 300 roundtrips to the database for renditions.

It would be beneficial to use 

```
            renditions = self.renditions.all()
            rendition = python_selection_func(renditions, filter, vary_key)
```

instead of 

```
            rendition = self.renditions.get(
                filter=filter,
                focal_point_key=vary_key,
            )
```

so that [prefetch_related](https://docs.djangoproject.com/en/1.7/ref/models/querysets/#prefetch-related) could be utilized.

davecranwell:
@kaedroho we'll want to be using picture elements (or similar polyfills) in some projects coming soon, could you investigate and respond on this? You seem most eligible.

kaedroho:
Thanks @thenewguy 

Another possibility on the API for this could be to create a new method on ``Image`` which takes a list of filters and returns a queryset of renditions. All of the optimisations can then be implemented in this method.

Also, we need to take into the account the fact that renditions may not exist and need to be generated on the fly. If none of the images that your gallery page needs are generated yet, that could cause 300 resizes to take place within the request. I'd be suprised if the request doesn't timeout before Wagtail has a chance to give a response.

thenewguy:
A new method would be handy if you want to add one to the api.  It would also be good to refactor image and filter creation out of the current method at the same time for DRYness.  Otherwise both methods will need to duplicate that code.

I am currently generating the images in advance with celery.  Because of this I decide how many and what dimensions of renditions to offer the client automatically based on source dimensions alone.  Ideally the rendition url would be known in advance so the url could be returned without creating the rendition and renditions generated on demand.  But then there would also need to be a way to verify valid dimension choices based on source dimensions before creating and serving a rendition.  So generating them with celery in advance is good enough for now.

[...]

PR #7225, opened 2021-05-29
Implemented prefetch_renditions method on ImageQuerySet

kaedroho:
This should be useful for improving performance on listings with a lot of images. Instead of looking them up individually from a cache or by running lots of database queries, this method allows a developer to fetch many renditions for a queryset of images using just a single query.

In theory, this should be much faster than looking up each one from a cache as there will be fewer round-trips to an external server (the performance gain from this will be much higher in a production environment than locally). This is should also be a big improvement for those who don't use redis.

### DIFF
--- a/wagtail/images/models.py
+++ b/wagtail/images/models.py
@@ -36,7 +36,24 @@ class SourceImageIOError(IOError):
 
 
 class ImageQuerySet(SearchableQuerySetMixin, models.QuerySet):
-    pass
+    def prefetch_renditions(self, *filters):
+        """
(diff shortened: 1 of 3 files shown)


ababic:
@kaedroho this is looking great! Would be fantastic to see this applied to Wagtail's image list / picker views.

kaedroho:
Thanks! I completely forgot about the chooser views!

zerolab:
Related: #947

kaedroho:
Implemented in https://github.com/wagtail/wagtail/pull/8584

PR #7227, opened 2021-05-29
Use specific_deferred when generating links to pages in rich text

kaedroho:
Currently, we fetch whole page objects when generating URLs to them in rich text. This can generate a lot of unnecessary large queries (especially if those pages have a lot of content).

This PR makes these queries use specific_deferred instead. This will cause a performance disadvantage for those who have overridden the URL generation logic though. But this is generally rare and I think we should discourage this.

### DIFF
--- a/wagtail/core/rich_text/pages.py
+++ b/wagtail/core/rich_text/pages.py
@@ -13,12 +13,12 @@ class PageLinkHandler(LinkHandler):
 
     @classmethod
     def get_instance(cls, attrs):
-        return super().get_instance(attrs).specific
+        return super().get_instance(attrs).specific_deferred
 
     @classmethod
     def expand_db_attributes(cls, attrs):
         try:
             page = cls.get_instance(attrs)
-            return '<a href="%s">' % escape(page.localized.specific.url)
+            return '<a href="%s">' % escape(page.localized.specific_deferred.url)
         except Page.DoesNotExist:
             return "<a>"


ababic:
@kaedroho this one got me thinking. Maybe we could add a couple of things to support custom routing without needing specific pages?
- Allow '/' characters in `slug`, so that pages can occupy more than one segment (e.g. An event could have something like 'year/month/usual-slug') - and adjust routing to account for that.
- Add a non-editable `custom_url_path` field that can be set by devs in cases where pages need to live in a certain structure for organisation, but the URL does not need to reflect that. We could look for this value in `Page.url_parts()` and `Page.route()` as well as the standard `url_path` and `slug`. How devs set this would be up to them.

kaedroho:
Thanks @ababic! That generally sounds like good plan to me.

In addition, we also have `.localize` here which would still generate a query for each page. Wonder if there's a way we could bulk load these? (same applies for image renditions and embeds too)

ababic:
> there's probably a few details to work out though

Definitely. Feeling like a short RFC would be best - I don't want to hijack this PR on vaguely related tangent :)

> We also have .localize here which would still generate a query for each Page. Wonder if there's a way we could bulk load these?

I would imagine you could annotate results with the generic localized Page easily enough, but getting the specific instance would be the challenge, and may still require an additional query. Would probably be best to keep the priority on simplicity in this case?

chosak:
> Wonder if there's a way we could bulk load these? (same applies for image renditions and embeds too).

If it's helpful, I suggested an approach to bulk loading rich text links in #5875.

zerolab:
I've tested this PR locally and it makes plenty of sense.

IMHO, the hit is acceptable for builds that customise the URL generation. And should be followed by the RFC/approach that @ababic suggest.

Both this and #5875 would be great additions. @kaedroho, one to maybe discuss at the next performance sub-team?

kaedroho:
@ababic I've submitted a proposal based on your comment [here](https://github.com/wagtail/wagtail/discussions/7812)

Let me know what you think!

lb-:
Been a while since the last action here and this PR is unlikely to be picked up any time soon.

Unless there is a push back I will close this PR and update comments on the already linked discussions and other PRs, potentially creating a new issue if needed to track the desire to improve the performance here.

zerolab:
I've suggested we use `specific_deferred` in #5875, so will close it as @lb suggests

PR #7228, opened 2021-05-30
Add background information on multi site, multi instance and multi tenancy

allcaps:
Add background information on multi site, multi instance and multi tenancy

[Rendered](https://wagtail--7228.org.readthedocs.build/en/7228/advanced_topics/multi_site_multi_instance_multi_tenancy.html)

* Do the tests still pass? Y
* Does the code comply with the style guide? Y
* For front-end changes: Did you test on all of [Wagtail’s supported environments] N/A

### DIFF
--- a/docs/advanced_topics/multi_site_multi_instance_multi_tenancy.rst
+++ b/docs/advanced_topics/multi_site_multi_instance_multi_tenancy.rst
@@ -1,0 +1,82 @@
+Multi-site, multi-instance and multi-tenancy
+============================================
+
+This page gives background information on how to run multiple Wagtail sites (with the same source code).
+
+.. contents::
+   :local:
+
+Multi-site
+----------
+
+Multi-site is a Wagtail project configuration where content creators go into a single admin interface and manage the content of multiple websites. Permission to manage specific content, and restricting access to other content, is possible to some extent.
+
+Multi-site configuration is a single code base, on a single server, connecting to a single database. Media is stored in a single media root directory. Content can be shared between sites.
+
+Wagtail supports multi-site out of the box: Wagtail comes with a :class:`site model <wagtail.core.models.Site>`. The site model contains a hostname, port, and root page field. When a URL is requested, the request comes in, the domain name and port are taken from the request object to look up the correct site object. The root page is used as starting point to resolve the URL and serve the correct page.
+
+Wagtail also comes with :ref:`site settings <settings>`. *Site settings* are 'singletons' that let you store additional information on a site. For example, social media settings, a field to upload a logo, or a choice field to select a theme.
+
+Model objects can be linked to a site by placing a foreign key field on the model pointing to the site object. A request object can be used to look up the current site. This way, content belonging to a specific site can be served.
+
+User, groups, and permissions can be configured in such a way that content creators can only manage the pages, images, and documents of a specific site. Wagtail can have multiple *site objects* and multiple *page trees*. Permissions can be linked to a specific page tree or a subsection thereof. Collections are used to categorize images and documents. A collection can be restricted to users who are in a specific group.
+
+Some projects require content editors to have permissions on specific sites and restrict access to other sites. Splitting *all* content per site and guaranteeing that no content 'leaks' is difficult to realize in a multi-site project. If you require full separation of content, then multi-instance might be a better fit...
+
+Multi-instance
+--------------
+
+Multi-instance is a Wagtail project configuration where a single code base is used by multiple websites. Each website has its server process, settings file, and a dedicated database and media directory. This guarantees the total separation of all content.
+
+Assume the domains a.com and b.com. Settings files can be `base.py`, `acom.py`, and `bcom.py`. The base settings will contain all settings like normal. The contents of site-specific settings override the base settings:
+
+.. code-block:: python
+
+   # acom.py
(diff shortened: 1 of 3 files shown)


docs/advanced_topics/multi_site_multi_instance_multi_tenancy.rst
```diff
@@ -0,0 +1,82 @@
...
+
+User, groups, and permissions can be configured in such a way that content creators can only manage the pages, images, and documents of a specific site. Wagtail can have multiple *site objects* and multiple *page trees*. Permissions can be linked to a specific page tree or a subsection thereof. Collections are used to categorize images and documents. A collection can be restricted to users who are in a specific group.
+
+Some projects require content editors to have permissions on specific sites and restrict access to other sites. Splitting *all* content per site and guaranteeing that no content 'leaks' is difficult to realize in a multi-site project. If you require full separation of content, then multi-instance might be a better fit...
+
+Multi-instance
+--------------
+
+Multi-instance is a Wagtail project configuration where a single code base is used by multiple websites. Each website has its server process, settings file, and a dedicated database and media directory. This guarantees the total separation of all content.
```
ababic, line 29: each website has its own

ababic: Also... Is there any particular reason why separate settings files are needed here? I would've thought you could get mostly the same level of control using a single one, with configuration at the env level (which would require no code changes to deploy a new site).

allcaps: The point is, that you can run with multiple settings. Maybe that is possible with environment variables. But adding env vars to the mix doesn't make it clearer what happens. 

Also, env vars are always the same right? How would you differentiate? Maybe 'a single code base' isn't clear enough. I mean, one directory on the server containing the source code, powering all instances. Not a single repo cloned in multiple directories.

docs/advanced_topics/multi_site_multi_instance_multi_tenancy.rst
```diff
@@ -0,0 +1,82 @@
...
+  - arbitrary sections of the page tree
+  - sections of the collection tree (coming soon)
+  - one or more collections of documents and images
+
+- The page API is automatically scoped to the host used for the request
+
+But several features do not currently support multi-tenancy:
+
+- Snippets are global pieces of content so not suitable for multi-tenancy but any model that can be registered as a snippet can also be managed via the Wagtail model admin. You can add a site_id to the model and then use the model admin get_queryset method to determine which site can manage each object. The built-in snippet choosers can be replaced by `modelchooser <https://pypi.org/project/wagtail-modelchooser/>`_ that allows filtering the queryset to restrict which sites may display which objects.
+- Site, site setting, user, and group management. At the moment, your best bet is to only allow superusers to manage these objects.
```
ababic, line 71: Maybe we could add something here as to why this is best left to superusers? Maybe "Permission configuration for built-in models like Sites, Site settings and Users are not site-specific, so any user with permission to edit a single entry can edit them all. This limitation can be mostly circumvented by only allowing superusers to manage these models.

docs/advanced_topics/multi_site_multi_instance_multi_tenancy.rst
ababic: Okay, maybe I'm misunderstanding this setup in terms of how it actually runs. How do you control which settings file is used by each process?

ababic: I was imagining some exotic Kubernetes setup with the same docker image powering differently configured nodes.

allcaps: With uWSGI, the acom.ini contains 
`env = DJANGO_SETTINGS_MODULE=settings.acom`
And bcom.ini contains:
`env = DJANGO_SETTINGS_MODULE=settings.bcom`

With a development server, you would run 
`./manage.py 8000 --settings settings.acom`
`./manage.py 8001 --settings settings.bcom`

Using environment variable wound't get you very far. All is executed in the same directory on exactly the same code. Wouldn't the env vars would be similar?

Of course the cost of cloning the repo a second time and have your project code multiple times on disk isn't very expensive. Disk space is cheep. But deploying many instances might be extra work. A single project dir running multiple sites, means a single deploy.

Does that make sense?

ababic: Yes, that makes sense. I think we could maybe just do with a tweak a here to make that clearer in the text?

docs/advanced_topics/multi_site_multi_instance_multi_tenancy.rst
allcaps: Done.

docs/advanced_topics/multi_site_multi_instance_multi_tenancy.rst
allcaps: @ababic I improved this section. Can you have another look?

ababic (approved):
LGTM 👍 thank you for your awesomeness!

gasman:
Merged in 3f24beaae60e281417b7818c5204c447c7989885

PR #7232, opened 2021-06-02
Don't set fixed width on the comment notifications dropdown

krzysztofjeziorny:
Longer words (as in German) can break the dropdown's layout.

Tested with Firefox 89 and Chromium 93 on macOS 11.4


krzysztofjeziorny:
The translation has been corrected as in 2.14.1, but the width is still set and the content overflows.

Screenshot 2021-08-16 at 11 04 22

lb-:
@krzysztofjeziorny thanks so much for taking the time to submit a PR, sorry it took so long to get a review.
I have merged this in via cf064f2b9990e90f5c5ca5d4405fe75a16604e0a
I just made one small refinement, changing the width to min-width instead of removing it, so it works nicely for super short translations also :) 

I have added your name to the contributors list, please check and let me know if you would like it changed.

Honestly, thanks again for submitting a fix.

krzysztofjeziorny:
@lb- Glad I could help! Yeah, setting `min-width` is a better way to fix it.

PR #7239, opened 2021-06-04
[experimental] Bulk load all external references in streamfields

kaedroho:
The current ``bulk_to_python`` methods in streafields only bulk load on one field at a time. This PR is a possible replacement for this that supports bulk loading referenced instances in the entire streamfield in one go.

**How it works**

Currently, we insert the IDs into the object returned by `get_prep_value` and this is stored in the database without any further conversion.

    [
        {
            "type": "image",
            "value": 1,
        },
        {
            "type": "image",
            "value": 2,
        },
        {
            "type": "image",
            "value": 1,
        }
    ]

This PR changes the` get_prep_value` implentation on ``ChooserBlock`` to return actual instances, and these get handled in a separate step. So the example above becomes:

    [
        {
            "type": "image",
            "value": <Image id=1>,
        },
        {
            "type": "image",
            "value": <Image id=2>,
        },
        {
            "type": "image",
            "value": <Image id=1>,
        }
    ]

Before serializing into JSON, these instances are replaced with `_ref` objects which contain an index to a "refs" object at the top of the streamfield value:

    {
        "blocks": [
            {
                "type": "image",
                "value": { "_ref": 0 }
            },
            {
                "type": "image",
                "value": { "_ref": 1 }
            },
            {
                "type": "image",
                "value": { "_ref": 0 }
            }
        ],
        "refs": [
            ["wagtailimages.Image", 1],
            ["wagtailimages.Image", 2]
        ]
    }

On retrieval, we bulk load the objects in “refs” and reinsert them back into the JSON object, to get back to the state we had in step 2. We then pass this value into `to_python`.

This “refs” key could be used later for quickly calculating object usage from streamfields too.

This would replace the existing and rather complicated `bulk_to_python` methods, which could mean that this change would reduce the complexity of the code overall. However, since it might take some time for existing data to be updated, we should keep this around for a couple of releases.

### DIFF
--- a/wagtail/core/blocks/field_block.py
+++ b/wagtail/core/blocks/field_block.py
@@ -665,6 +665,11 @@ class ChooserBlock(FieldBlock):
         # the incoming serialised value should be None or an ID
         if value is None:
             return value
+
+        elif isinstance(value, self.target_model):
+            # Value bulk loaded by streamfield
+            return value
+
         else:
             try:
                 return self.target_model.objects.get(pk=value)
@@ -684,7 +689,7 @@ class ChooserBlock(FieldBlock):
         if value is None:
             return None
         else:
-            return value.pk
+            return value
 
     def value_from_form(self, value):
         # ModelChoiceField sometimes returns an ID, and sometimes an instance; we want the instance
@@ -707,6 +712,9 @@ class ChooserBlock(FieldBlock):
         if isinstance(value, self.target_model):
             value = value.pk
         return super().clean(value)
+
+    def get_api_representation(self, value, context=None):
+        return value.pk if value else None
 
     class Meta:
         # No icon specified here, because that depends on the purpose that the
--- a/wagtail/core/blocks/stream_block.py
+++ b/wagtail/core/blocks/stream_block.py
@@ -6,6 +6,7 @@ from collections.abc import MutableSequence
 
 from django import forms
 from django.core.exceptions import NON_FIELD_ERRORS, ValidationError
+from django.db import models
 from django.forms.utils import ErrorList
 from django.utils.functional import cached_property
 from django.utils.html import format_html_join
@@ -13,6 +14,7 @@ from django.utils.translation import gettext as _
 
 from wagtail.admin.staticfiles import versioned_static
 from wagtail.core.telepath import Adapter, register
+from wagtail.core.utils import resolve_model_string
 
 from .base import Block, BoundBlock, DeclarativeSubBlocksMetaclass, get_help_icon
 
@@ -177,11 +179,18 @@ class BaseStreamBlock(Block):
 
         return StreamValue(self, cleaned_data)
 
-    def to_python(self, value):
+    def to_python(self, value, root=False):
         # the incoming JSONish representation is a list of dicts, each with a 'type' and 'value' field
         # (and possibly an 'id' too).
         # This is passed to StreamValue to be expanded lazily - but first we reject any unrecognised
         # block types from the list
+
+        if root and isinstance(value, dict):
+            return StreamValue(self, [
+                child_data for child_data in value['value']
+                if child_data['type'] in self.child_blocks
+            ], is_lazy=True, refs=value['refs'])
+
         return StreamValue(self, [
             child_data for child_data in value
             if child_data['type'] in self.child_blocks
@@ -238,7 +247,7 @@ class BaseStreamBlock(Block):
             for block_map in block_maps
         ]
 
-    def get_prep_value(self, value):
+    def get_prep_value(self, value, root=False):
         if not value:
             # Falsy values (including None, empty string, empty list, and
             # empty StreamValue) become an empty stream
@@ -247,7 +256,7 @@ class BaseStreamBlock(Block):
             # value is a StreamValue - delegate to its get_prep_value() method
             # (which has special-case handling for lazy StreamValues to avoid useless
             # round-trips to the full data representation and back)
-            return value.get_prep_value()
+            return value.get_prep_value(root=root)
 
     def get_form_state(self, value):
         if not value:
@@ -416,7 +425,7 @@ class StreamValue(MutableSequence):
         def __repr__(self):
             return repr(list(self))
 
-    def __init__(self, stream_block, stream_data, is_lazy=False, raw_text=None):
+    def __init__(self, stream_block, stream_data, is_lazy=False, refs=None, raw_text=None):
         """
         Construct a StreamValue linked to the given StreamBlock,
         with child values given in stream_data.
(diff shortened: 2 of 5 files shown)


gasman:
Thanks @kaedroho - this is a really interesting approach. However, I'm reluctant to change the JSON data representation, for a few reasons:

* There's a lot of code in the wild that works directly with the current JSON representation, because in many respects it's more self-explanatory and simpler than wrangling the native-Python types (image objects, RichText values etc). Requiring all of that code to be updated to work with a more complex representation - one that only really benefits Wagtail's internal code - would be an unreasonable burden, one that goes beyond what we can reasonably ask people to make through the normal deprecation process.
* It turns the semantics of `get_prep_value` from something that's easily understandable in one line (convert a native Python value for this block into something JSON-serialisable) to something that's full of exceptions and edge cases.
* Ultimately, it's not necessary - if we're capable of walking the data structure to build a flat list of object references at the point of saving the data, we can equally do that when reading the data back. (Yes, reads will happen more often than writes, but walking the tree is not the bottleneck here - it's the database lookups, and we can achieve the same optimisation there either way.)

I acknowledge that understanding _why_ bulk_to_python works is complicated, but it's a self-contained algorithmic kind of complexity that doesn't 'leak out' into the rest of the design of StreamField. Structural blocks such as StreamBlock are only concerned with organising the data of their children - they don't care whether that data comes from a database or has undergone rich text conversion, or whether they're the top-level block or not. As far as bulk_to_python is concerned, its goal is to convert all of the incoming data in as few calls to its children's bulk_to_python methods as possible. (It doesn't care _why_ fewer, big calls to bulk_to_python is better than lots of small ones.)

Conversely, blocks like ImageChooserBlock that do database lookups _only_ do database lookups, and aren't concerned with the bigger structure. If it receives a list of 100 IDs in bulk_to_python, it doesn't care what contortions StreamBlock went through to pull those IDs into a flat list.

What we're in danger of doing here is trading that 'algorithmic' complexity for a "keeping lots of concepts in your head" cognitive-load kind of complexity, where every component needs to know about everything else in the stack: chooser blocks have to know when it's safe to return non-JSON values, and StreamBlocks need to know how to do database lookups (which isn't the case right now - they do pure data structure manipulation), and whether they're passing data back to the StreamField or to another parent block. I do believe that the current separation of concerns is the key to how StreamField has managed to rein in complexity over time, where other similarly-clever bits of code in the Wagtail ecosystem (like modelcluster and wagtail-transfer) haven't.

Now, having said that, this PR does provide a key improvement over the bulk_to_python approach, namely that all lookups of the same object type can be batched together, not just values that are to be handled by the same specific block instance (i.e. values at the same structural position in the tree). I'm confident that we can find an implementation that captures that idea without breaking the separation of concerns. I have a hunch that it's something along the lines of: rather than each block instance implementing its own to_python / bulk_to_python methods, it delegates that to a dedicated 'preloader' object. In the case of something like ImageChooserBlock, that would be a singleton object shared by all ImageChooserBlock instances. There will then be some mechanism (which I haven't figured out yet) where those preloader objects will bubble up to the top level, and the higher-level blocks will identify where the same preloader is in use in multiple places, and ensure that all requests to that preloader are handled in a single call.

kaedroho:
Thanks for your very detailed reply @gasman! I generally agree with all that you've said.

One reason I liked the "refs" approach is that I think it would make it easier to find the number of StreamFields that reference a particular object, but there are probably other ways to store that.

I like the general idea of using pre-loader objects. One thing that might be able to do, that this PR wasn't able to, is pre-loading objects that aren't directly referenced from the JSON such as image renditions and embeds.

Another extension to that, which might not be difficult to implement but would be really helpful, could be to allow people to collect together these pre-loaders for multiple StreamFields and run them together. This would allow us to nest data about images and embeds in the API (we can't do this currently as we don't allow N+1 queries) and make Wagtail more efficient for bulk static-site generation.

gasman:
I've got a bit more of an idea of how this could work, and the missing piece is a mini-library to encapsulate the general concept of a 'batch operation': something that takes an input and generates a result, and is most efficient when performed in bulk. For example, a database lookup. This might exist already, and we should check before reinventing the wheel :-)

This library defines two types:

`Operation` - represents some task that can be applied to an input, such as "fetch a page object with a given ID". Operation objects are immutable, hashable, and implement equality such that two operations that do the same thing compare as equal. This makes it possible to use them as dictionary keys. Operation objects implement a `run_batch` method which takes a `set` of inputs and returns a `dict` mapping inputs to outputs. Wagtail would provide a subclass `DatabaseLookup`:

    page_lookup = DatabaseLookup(Page, only=['title'])
    page_lookup.run_batch({1, 2, 3})  # returns {1: <Page "foo">, 2: <Page "bar">, 3: <Page "baz">}

`Batch` - represents a batch of work to be performed by multiple Operations. Internally this is represented as a dict mapping `Operation` objects to a `set` of inputs:

    my_batch = Batch({
        page_lookup: {1, 2, 3},
        image_lookup: {4, 5, 6},
    })

Batch objects can be added together, to return a new Batch object with the dictionaries merged. Finally, they implement a `run()` method that returns a dict of the results of `run_batch` on all operations:

    my_batch.run()
    => {
        page_lookup: {1: <Page "foo">, 2: <Page "bar">, 3: <Page "baz">},
        image_lookup: {4: <Image "foo.jpg">, 5: <Image "bar.jpg">, 6: <Image "baz.jpg>},
    }

This would be used by StreamField as follows:

Blocks would implement a `prefetch(val)` method which returns a Batch object representing the lookups that it would like to happen in advance of running `to_python`. Structural blocks would implement this by calling `prefetch` on all children and adding the resulting Batch objects together. The top-level block initiating the `to_python` operation would then call `run()` on the final batch, before passing the results to each child `to_python` method (or, more likely, a new method `to_python_with_prefetch` for that purpose).

    class Block:
        def prefetch(self, value):
            # by default, blocks do not prefetch anything
            return Batch({})

        def to_python_with_prefetch(self, value, prefetch_result):
            # by default, blocks do not make use of the prefetch result
            return self.to_python(value)


    class ImageChooserBlock(Block):
        image_lookup = DatabaseLookup(Image)

        def prefetch(self, value):
            return Batch({
                self.image_lookup: {value}
            }

        def to_python_with_prefetch(self, value, prefetch_result):
            return prefetch_result[self.image_lookup][value]


    class ListBlock(Block):
        def prefetch(self, value):
            result = Batch({})
            for item in value:
                result += self.child_block.prefetch(item)
            return result

        def to_python_with_prefetch(self, value, prefetch_result):
            return [
                self.child_block.to_python_with_prefetch(item, prefetch_result)
                for item in value
            ]

        def to_python(self, value):
            # once all container block types have adopted this optimisation, we would expect to_python
            # to only ever be called on the top-level StreamBlock. However, the implementation will be
            # the same in all cases...
            prefetch_result = self.prefetch(value).run()
            return self.to_python_with_prefetch(value, prefetch_result)

PR #7247, opened 2021-06-10
Fix eslint warnings/errors

### LINKED ISSUE #7177: eslint failure in client/src/components/CommentApp/utils/storybook.tsx
The workaround in d50da87cf4df580c0c46c67016bba96ec32bd6bd to stop the core.css import from failing during CI causes linting to fail when running on a development instance that _does_ have static assets built, because it's treated as an unnecessary directive in that case:

```
/Users/matthew/Development/tbx/wagtail/devscript/wagtail/client/src/components/CommentApp/utils/storybook.tsx
  24:1  error  Unused eslint-disable directive (no problems were reported from 'import/no-unresolved')
```

gasman:
I tried and failed to upgrade to eslint-config-wagtail 0.3.0 (I suspect packaging issues with eslint-config-airbnb et al due to them being moved from dependencies to devDependencies...) so as a stopgap to make our linting usable again, I've fixed the code to comply with our existing rules (including `eslint-disable-next-line no-warning-comments` against the TODO lines, which I think is a necessary evil at this point).

Have also added an `npm run build` step to CircleCI to fix #7177 (but will only be able to test that at the point of opening this PR, so here's hoping...)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -41,6 +41,7 @@ jobs:
           paths:
             - ~/project/node_modules/
           key: node-v1-{{ .Branch }}-{{ checksum "package-lock.json" }}
+      - run: npm run build
       - run: npm run lint:js
       - run: npm run lint:css
       - run: npm run test:unit:coverage -- --runInBand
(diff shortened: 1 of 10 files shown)


gasman:
Have now added a fix for the long-broken cleanForSlug tests, so now `npm run test` is usable again as well :-)

kaedroho (approved):

PR #7251, opened 2021-06-14
Refactor images index into CBV and eliminate is_ajax

gasman:
Basically a repeat of #7209 for the images index view (but this time it does make sense to split off a separate subclass for the results-only view, since that allows us to skip some page-furniture-only logic...)


ababic (approved):
Good stuff @gasman. Thanks for this! I agree that there's more of a case here for the separate views. One additional consideration is: Whatever we do here, we should try to mimic for documents, as it'll help with refactoring later on.

PR #7254, opened 2021-06-14
Split up wagtail.core.models

gasman:
About time, I think. :-)

Everything page-related is still in the top-level module (which will hopefully reduce the number of merge conflicts with current PRs), and 'noqa' imports have been added to ensure that everything is still importable from their old locations. (Third-party code could still potentially break if tries to import things from wagtail.core.models that actually exist in e.g. wagtail.core.utils, but at least it will break in an obvious way...)

### DIFF
--- a/wagtail/core/models/__init__.py
+++ b/wagtail/core/models/__init__.py
@@ -9,7 +9,6 @@ from urllib.parse import urlparse
 from django import forms
 from django.apps import apps
 from django.conf import settings
-from django.contrib.auth import get_user_model
 from django.contrib.auth.models import Group
 from django.contrib.contenttypes.models import ContentType
 from django.core import checks
@@ -52,6 +51,7 @@ from wagtail.core.utils import (
     get_supported_content_language_variant, resolve_model_string)
 from wagtail.search import index
 
+from .audit_log import BaseLogEntry, BaseLogEntryManager, LogEntryQuerySet  # noqa
 from .collections import (  # noqa
     BaseCollectionManager, Collection, CollectionManager, CollectionMember,
     CollectionViewRestriction, GroupCollectionPermission, GroupCollectionPermissionManager,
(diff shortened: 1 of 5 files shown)


wagtail/core/models/__init__.py
```diff
@@ -3680,12 +3470,13 @@ def save(self, user=None, specific_instance=None, **kwargs):
                 }
             )
 
-    def delete(self, user=None, specific_instance=None, **kwargs):
+    def delete(self, user=None, **kwargs):
```
ababic, line 3473: Not quite sure of the consequences of this change. It might be safer to keep functional changes out of this PR?

gasman: I tried to keep the code changes to a minimum, but this one was needed so that BaseViewRestriction didn't have a dependency on PageLogEntry (which would prevent us from moving it out into its own module). I don't think there's a functional change here - the only reasonable circumstance where BaseViewRestriction.save and BaseViewRestriction.delete would be passed a `specific_instance` kwarg is in the `super` call within PageViewRestriction. If anything, I'd say that the presence of page-related code in BaseViewRestriction rather than PageViewRestriction was a (benign) bug.

ababic: Ah, I see that now. Thanks for clarifying!

zerolab (approved):
Excellent split up and tidy up! Thank you @gasman 

non-blocking question: would it help devs to add a note in `.audit_log` / `.view_restrictions` that specific implementations are in the relevant modules?

gasman:
> non-blocking question: would it help devs to add a note in `.audit_log` / `.view_restrictions` that specific implementations are in the relevant modules?

Good idea! The direction of dependencies here (submodules providing the base implementations for the page-specific ones in wagtail.core.models to use) seems fairly obvious now, but it's exactly the sort of thing that's likely to slip over time if we don't document it explicitly... will add that when I merge.

PR #7257, opened 2021-06-15
Split translation models out into wagtail.core.models.i18n

gasman:


wagtail/core/models/translation.py
```diff
@@ -0,0 +1,331 @@
+import uuid
```
kaedroho, line 1: Could we name this `i18n` short for Internationalisation? Since these models are mainly to allow us to organise a site by locales, the actual translation part is done by `wagtail-localize` and `wagtail.contrib.simple_translation`

wagtail/core/models/copying.py
```diff
@@ -0,0 +1,96 @@
+from django.db import models
```
kaedroho, line 1: Wonder if this should be moved to `wagtail/core/utils/copying.py` instead? Since there are no models here?

wagtail/core/models/translation.py
gasman: Sure, sounds good. That'll also get around the need for an alias when importing django.utils.translation...

wagtail/core/models/copying.py
gasman: I'm not really a fan of 'utils' modules - they tend to become a bit of a dumping ground. Moving it outside of core.models also raises questions about whether we expect this to be re-used in other apps/modules, and if so, whether they should become un-underscored... I think that's a yak-shave too far for this PR, and for now we should stick to just splitting up wagtail.core.models rather than actually moving things out of it.

wagtail/core/models/translation.py
gasman: Updated to use wagtail.core.models.i18n now.

kaedroho (approved):

wagtail/core/models/__init__.py
```diff
@@ -248,6 +237,25 @@ def __str__(self):
         return force_str(self.get_display_name() or self.language_code)
 
 
+@receiver(pre_validate_delete, sender=Locale)
```
zerolab, line 82: question: should this not live in `core.signal_handlers`? or this is too edge case for that?

gasman: I debated that a bit... there's precedent for putting signal handlers here (`set_locale_on_new_instance`, now in models/i18n.py) and intuitively it does feel like these are part of the model definition in a way that more conventional uses of signals (e.g. sending a notification) wouldn't be. The Django docs [do advise against putting signal handlers in models.py](https://docs.djangoproject.com/en/3.2/topics/signals/#connecting-receiver-functions) (see under "Where should this code live?"), but I don't entirely follow the reasoning there, and in this case it wouldn't be a problem if it did end up being registered multiple times. I don't have a particularly strong feeling either way, and I'd be happy to move it to signal_handlers if that's the consensus.

PR #7258, opened 2021-06-16
docs: update custom_account_settings.rst

swilltec:
Modify CustomSettingsForm import in hooks.py in user model usage


gasman (approved):
Thanks @swilltec!

PR #7262, opened 2021-06-18
Give 404 pages a meaningful HTML title in project template

gasman:
Improve the generation of `<title>` tags as follows:

* use `page` in preference to `self` (self has been semi-deprecated ever since we added jinja2 support)
* Retrieve current site with `{% wagtail_site %}` rather than page.get_site so that it works on non-pages such as 404s
* Fill in the 'title' block on 404.html


jacobtoppm (approved):
This all looks good to me!

gasman:
Merged in a28376a5c9ef990e9eb070bb6171cff968d6e75d

Issue #7240, opened 2021-06-06
A shortcut for accessing StreamField blocks by name


gasman:
### Is your proposal related to a problem?

It's common for people to want to retrieve a block of a particular type from a StreamField value - for example, if 'hero image' is one of the block types defined in a page's content, a developer may want to pull out the first such hero image for use in listings or as a social share image. ([An example of this in the wild](https://stackoverflow.com/questions/67855617/cannot-get-blog-listing-using-get-context-to-work-properly/67860016#67860016)...)

This can be done in user code (either in Python or Django template code) by looping over the StreamField value until the appropriate block is found - however, I think this is a common enough scenario that Wagtail should provide a standard method / idiom for it.

### Describe the solution you'd like

The `StreamValue` class could implement a `by_type` method that returns a list of blocks matching the given type:

    page.body.by_type('hero_image')  # returns a list of all `hero_image` blocks in the `body` StreamField

Passing a parameter to `by_type` in this way would not be possible within a template, so as an additional feature, `by_type` could return a (lazily-evaluated?) dict consisting of all blocks organised by type when invoked with no parameters. This would make it possible to write something like:

    <h1>{{ page.body.by_type.heading.0 }}</h1>

The `by_type` method will need tests and documentation.

### Describe alternatives you've considered

StreamValue could be updated to behave as something like a dict as well as a list, e.g.`page.body['hero_image']` - however, this feels a bit ambiguous about whether it should return a single item or a list. (Also, in template code where `foo.bar` is equivalent to `foo['bar']`, it means the block names end up in the same namespace as methods on StreamValue, which feels icky.)

Given that people will probably be using this for retrieving a single item most of the time, we could borrow BeautifulSoup's pattern of `page.body.find('hero_image')` (or `page.body.find(type='hero_image')`?) which returns the first item or None, and `page.body.find_all('hero_image')` which returns a list. It's not so obvious how to make this syntax work nicely within templates, though.

### Additional context

Perhaps something for a later phase, but it's worth bearing in mind that for this kind of StreamField access, optimisations like `bulk_to_python` and #7239 are actually counter-productive, since they're pre-fetching all database objects in the stream but we only want one of them. Maybe this calls for us to treat StreamValues more like querysets, where the actual data retrieval happens as late as possible and these kinds of 'filter' operations act as modifiers to that final lookup.

Scotchester:
[...]

PR #7266, opened 2021-06-21
Shortcut for accessing StreamField blocks by block name/type

Tijani-Dia:
See #7240 

This PR adds three methods/shortcuts to `StreamValue` class.
- `blocks_by_type` Arguments: None
This method returns a  dict consisting of all blocks organised by type.
It allows writing something like the following in templates:
`<h1>{{ page.body.blocks_by_type.heading.0 }}</h1>
`

- `blocks_of_type` Arguments: `block_type`
This method returns all blocks of type `block_type`.

- `first_block_of_type` Arguments: `block_type`
This method returns the first block of type `block_type`

* Do the tests still pass?
  Yes
* Does the code comply with the style guide? 
  Yes
  Yes
  Not yet. I will add it if the implementation is approved.

### DIFF
--- a/wagtail/core/blocks/stream_block.py
+++ b/wagtail/core/blocks/stream_block.py
@@ -544,13 +544,13 @@ class StreamValue(MutableSequence):
         return prep_value
 
     def blocks_by_name(self, block_name=None):
-        blocks = defaultdict(list)
+        blocks = {block: [] for block in self.stream_block.child_blocks}
         for item in self:
             blocks[item.block_type].append(item)
 
         if block_name is None:
             return blocks
-        return blocks[block_name]
+        return blocks.get(block_name, [])
 
     def first_block_by_name(self, block_name):
         if block_name not in self.stream_block.child_blocks:
--- a/wagtail/core/blocks/stream_block.py
+++ b/wagtail/core/blocks/stream_block.py
@@ -547,12 +547,10 @@ class StreamValue(MutableSequence):
         blocks = {}
         for item in self:
             block_type = item.block_type
-            block_value = item.value
-
             if block_type in blocks:
-                blocks[block_type].append(block_value)
+                blocks[block_type].append(item)
             else:
-                blocks[block_type] = [block_value]
+                blocks[block_type] = [item]
 
         return blocks
 
@@ -568,7 +566,7 @@ class StreamValue(MutableSequence):
 
         for item in self:
             if item.block_type == block_type:
-                return item.value
+                return item
 
     def __eq__(self, other):
         if not isinstance(other, StreamValue) or len(other) != len(self):
--- a/wagtail/core/blocks/stream_block.py
+++ b/wagtail/core/blocks/stream_block.py
@@ -558,7 +558,7 @@ class StreamValue(MutableSequence):
         if block_type not in self.stream_block.child_blocks:
             return []
 
-        return self.blocks_by_type()[block_type]
+        return self.blocks_by_type().get(block_type, [])
 
     def first_block_of_type(self, block_type):
         if block_type not in self.stream_block.child_blocks:
--- a/wagtail/core/blocks/stream_block.py
+++ b/wagtail/core/blocks/stream_block.py
@@ -588,9 +588,21 @@ class StreamValue(MutableSequence):
         if block_name not in self.stream_block.child_blocks:
             return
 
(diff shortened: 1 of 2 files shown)


wagtail/core/blocks/stream_block.py
```diff
@@ -543,6 +543,33 @@ def get_prep_value(self):
...
+
+        return self.blocks_by_type()[block_type]
+
+    def first_block_of_type(self, block_type):
+        if block_type not in self.stream_block.child_blocks:
+            return
+
+        for item in self:
+            if item.block_type == block_type:
+                return item.value
```
gasman, line 571: I think it would be better for these methods to return the block object (i.e. `return item` here) rather than the value. There are some cases where the block object has some additional functionality that we want to keep - for example, if we have a block like

    heading = CharBlock(template='blocks/heading.html')

where the template is `<h1>{{ value }}</h1>`, then calling {% include_block page.body.blocks_by_type.heading.0 %} should include the `<h1>` tags. This works if we return the block object, but not if we only have the string value.

wagtail/core/blocks/stream_block.py
```diff
@@ -543,6 +543,33 @@ def get_prep_value(self):
...
+
+            if block_type in blocks:
+                blocks[block_type].append(block_value)
+            else:
+                blocks[block_type] = [block_value]
+
+        return blocks
+
+    def blocks_of_type(self, block_type):
+        if block_type not in self.stream_block.child_blocks:
```
gasman, line 560: I don't think this test is quite right: if `block_type` is not a recognised block type for the StreamField (as in your `invalid_type` test), it will do the right thing, but if `block_type` _is_ a recognised type but there are no instances of it in the stream, then it will pass this test and `self.blocks_by_type()[block_type]` will still fail with a KeyError.

gasman (changes requested):
Thanks @Tijani-Dia!

wagtail/core/blocks/stream_block.py
Tijani-Dia: Ah ok. Alright, I will do that.

wagtail/core/blocks/stream_block.py
Tijani-Dia: Ok, so should I do a try/except and if we catch a `KeyError` we return None or an empty list depending on the case?

gasman: Yes, something like that. I don't think we need to distinguish between the "unrecognised block type" and "no instances found" cases - we can just return the 'empty' result (None or empty list). So I think we can leave out the `if block_type not in self.stream_block.child_blocks` check entirely and just do something like: `return self.blocks_by_type().get(block_type, [])`

lb-:
@Tijani-Dia let us know if you want to get another review on this or if you are still good to work through the changes requested.

Tijani-Dia:
@lb- I think I made the changes requested. I was also waiting to see a feedback on [this comment](https://github.com/wagtail/wagtail/issues/7240#issuecomment-865346461).

wagtail/core/blocks/stream_block.py
gasman: Now that you've updated the `return` line, this check can be removed completely.

wagtail/core/blocks/stream_block.py
```diff
@@ -543,6 +543,31 @@ def get_prep_value(self):
 
         return prep_value
 
+    def blocks_by_type(self):
+        blocks = {}
```
gasman, line 547: Taking another look at this... I think it makes sense for the return value here to include an empty list for any block type that is part of the definition but has no blocks. That also means that we can simplify the code below...

    def blocks_by_type(self):
        blocks = {block.name: [] for block in self.stream_block.child_blocks}
        for item in self:
            blocks[item.block_type] = item
        return blocks

Tijani-Dia:
@gasman I applied the different suggestions made about the naming, implementation...

I kept it simple up to [this commit](https://github.com/wagtail/wagtail/pull/7266/commits/12410065189c9da7ee54246e9bb7d72f036d57f6).

Beyond that, I was trying to see if I could also implement the lazy evaluation stuff. If that's not what you meant by lazy evaluation, we can still revert to the commit mentionned above.

gasman:
Rebased at #8735

Issue #7267, opened 2021-06-21
Commenting JS blocks loading of Draftail in a custom image model's RichTextField


Scotchester:
### Issue Summary

If I add a `RichTextField` to a custom image model, when the form renders (either post-upload or by going to the image’s edit page), I am getting this JavaScript console error:

```
Uncaught TypeError: window.comments is undefined
    initEditor index.js:111
    <anonymous> (index):591
index.js:111:8
```

Switching to a plain `CharField`, the error goes away.

### Steps to Reproduce

1. Start a new project with `wagtail start myproject`
2. Edit `home/models.py` as follows:
   ```python
   from django.db import models
   
   from wagtail.core.fields import RichTextField
   from wagtail.core.models import Page
   from wagtail.images.models import AbstractImage, AbstractRendition, Image
   
   
   class HomePage(Page):
       pass
   
   
   class CustomImage(AbstractImage):
       caption = RichTextField(
           features=['bold', 'italic'],
           blank=True,
           null=True,
       )
   
       admin_form_fields = Image.admin_form_fields + (
           'caption',
       )
   
   
   class CustomRendition(AbstractRendition):
       image = models.ForeignKey(
           CustomImage,
           on_delete=models.CASCADE,
           related_name='renditions'
       )
   
       class Meta:
           unique_together = (
               ('image', 'filter_spec', 'focal_point_key'),
           )
   ```
3. `makemigrations` and `migrate`
4. Open the admin
5. Upload an image
6. See that the Draftail field for `caption` does not render
7. Open the console and see the error mentioned above

Alternatively, I'm not sure how long this will stay available, but [log into this Gitpod I set up](https://8080-indigo-sheep-q5kf5o4o.ws-us08.gitpod.io/admin/images/1/) and open one of the images for editing to see the error in the console.

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: **yes**

I also tried turning off the commenting feature in settings with `WAGTAILADMIN_COMMENTS_ENABLED = False`, but that didn’t get rid of the error.

@jacobtoppm's initial thoughts from Slack:

> Sounds like a bug to me! Tested on snippets but not images, which of course don't use the editor js... Just needs a check for `window.comments` in the Draftail init code before it decides to create `CommentableEditor` instead of plain Draftail, I think.

### Technical details

* Python version: 3.8.10
* Django version: 3.2.4
* Wagtail version: 2.13.2
* Browser version: Firefox 78.11.0esr

PR #7268, opened 2021-06-22
Don't assume window.comments exists during Draftail init

jacobtoppm:
This may break use of Draftail widgets in custom image models, where the usual editor js is not used.

Fixes https://github.com/wagtail/wagtail/issues/7267


Scotchester (approved):
Well, I had this working in a Gitpod yesterday and just didn't get the PR in before the end of the day, but this is a much more elegant solution than I had anyway 😆 TIL about optional chaining in JS! Very cool.

PR #7269, opened 2021-06-22
[Docs] Incorrect field name in example (tiny)

danthedeckie:
Tiny documentation fix, `body` instead of `book_text` - as referenced by the `content_panels` line straight afterwards.


kaedroho:
Thanks!

PR #7272, opened 2021-06-22
RichTextField when cloned should include features & editor settings

danthedeckie:
A `RichTextField` when `clone`d should also include the `features` & `editor` kwargs it was originally created with.

### DIFF
--- a/wagtail/core/fields.py
+++ b/wagtail/core/fields.py
@@ -13,12 +13,12 @@ class RichTextField(models.TextField):
         self.editor = kwargs.pop('editor', 'default')
         self.features = kwargs.pop('features', None)
         super().__init__(*args, **kwargs)
-        
-    def deconstruct(self):
-        name, path, args, kwargs = super().deconstruct()
-        kwargs['features'] = self.features
-        kwargs['editor'] = self.editor
-        return name, path, args, kwargs
+
+    def clone(self):
+        name, path, args, kwargs = self.deconstruct()
+        kwargs["features"] = self.features
+        kwargs["editor"] = self.editor
+        return self.__class__(*args, **kwargs)
 
     def formfield(self, **kwargs):
         from wagtail.admin.rich_text import get_rich_text_editor_widget
--- a/wagtail/core/fields.py
+++ b/wagtail/core/fields.py
@@ -12,8 +12,13 @@ class RichTextField(models.TextField):
     def __init__(self, *args, **kwargs):
         self.editor = kwargs.pop('editor', 'default')
         self.features = kwargs.pop('features', None)
-        # TODO: preserve 'editor' and 'features' when deconstructing for migrations
         super().__init__(*args, **kwargs)
+        
+    def deconstruct(self):
+        name, path, args, kwargs = super().deconstruct()
+        kwargs['features'] = self.features
+        kwargs['editor'] = self.editor
+        return name, path, args, kwargs
 
     def formfield(self, **kwargs):
         from wagtail.admin.rich_text import get_rich_text_editor_widget


gasman:
Hi @danthedeckie - thanks for the PR.
Do you have a specific use-case in mind for this? Despite what the TODO note says, I'm thinking it might actually make more sense to leave this as-is, since changing this will cause new migrations to be generated on all rich text fields when users next upgrade Wagtail (which is what's causing the test failures here - there are changes in the test models that aren't accounted for by migrations), and I suspect there are very few situations where having an accurate `features` / `editor` value inside migrations would be any benefit.

danthedeckie:
The specific use case that made me notice this was we have a site with a few languages all on the same model (pre wagtail multilingual support) and we have a `TranslatedField` which works kinda similar to `ArrayField`:
```python
class WhateverPage(Page):
    translated_title = TranslatedField(models.CharField(max_length=100, blank=True))
    translated_subtitle = TranslatedField(RichTextField(features=['bold', 'italic', 'link'], blank=True))
```

kind of thing - then the TranslatedField generates all the underlying fields on the model - equivalent of:
```
    spanish_subtitle=RichTextField(features=['bold', 'italic', 'link'], blank=True)
    german_subtitle=RichTextField(features=['bold', 'italic', 'link'], blank=True)
    english_subtitle=RichTextField(features=['bold', 'italic', 'link'], blank=True)
```
and then the accessor on `translated_subtitle` returns the appropriate field depending on your current language.  It works pretty well - except that the RichTextField, missing the features, ends up creating all the fields w/o features set.

We've got around it by subclassing `RichTextField` with the changes in the PR.

danthedeckie:
Our `TranslatedField` is something like:
```
 class TranslatedField:
    def __init__(self, base_field, other_options):
        self.base_field = base_field

    def contribute_to_class(self, cls, name):
        for language in SITE_LANGUAGE_CODES:
            field = self.base_field.clone()
            field_name = f"{name}_{language}"
            setattr(cls, field_name, field)
            field.contribute_to_class(cls, field_name)
```

(There's a bit more to it than that the codebase, handling extra options, over-rides, re-using existant fields, keeping track of the new fields for accessing, etc.)

So when the field is initialised by django, it clones the base field, and attaches them to the model.

gasman:
Ah, I see - [`self.base_field.clone()` makes use of `deconstruct`](https://github.com/django/django/blob/43d1ea6e2ff1982d52faf6b04d613390220e5d7a/django/db/models/fields/__init__.py#L507-L513), so this is about making `clone` behave correctly and not about migrations as I thought...

So, the question remains about how to deal with the migrations that we'll end up creating on everyone's existing RichTextFields due to this change of representation. I can see three possibilities:

1) accept that it will happen (and fix up Wagtail's own test migrations accordingly)
2) make the `deconstruct` logic a bit smarter so that it only adds the `features` and `editor` kwargs to the deconstructed representation if they differ from the system-wide default
3) Instead of changing `deconstruct`, override `RichTextField.clone` to pass on the `features` and `editor` kwargs - that way `clone` will work as intended without any impact on migrations

I'm thinking that option 3 is the way forward - what do you think?

For the record, I suspect there are other places in Wagtail where the `deconstruct` result is not a 100% accurate copy of the field - for example, [not preserving the `widget` kwarg on choice blocks within StreamField](https://github.com/wagtail/wagtail/blob/9f3e6585afd07bab15c22257ecf4ed4eab5ab0dd/wagtail/core/blocks/field_block.py#L421-L423) - it's possible that your code may trip up on these too.

danthedeckie:
I've updated the PR to approach 3 like you suggested. It still seems to work correctly for us here 👍 

Would you like me to squash the PR completely?

gasman:
Thanks again @danthedeckie! Now merged in fee65a93458b0c4b0b68dfefade13fda9c6291cb. What name would you like to be credited under in CONTRIBUTORS.rst?

danthedeckie:
> Thanks again @danthedeckie! Now merged in [fee65a9](https://github.com/wagtail/wagtail/commit/fee65a93458b0c4b0b68dfefade13fda9c6291cb). What name would you like to be credited under in CONTRIBUTORS.rst?

"Daniel Fairhead" would be great!
Thanks @gasman

PR #7273, opened 2021-06-22
Remove ability to add arbitrary classes to menu items from slim menu

kaedroho:
This was added to allow users to add custom icons to menu items. This is no longer needed as ``icon_name`` can be used to provide an icon now.

The slim menu won't become a mandatory feature until at least 2.16, so we don't need to deprecate this in the slim menu.

TODO:
 - [ ] Deprecation warning for people using custom classes on the non-slim menu

### DIFF
--- a/client/src/components/Sidebar/menu/LinkMenuItem.tsx
+++ b/client/src/components/Sidebar/menu/LinkMenuItem.tsx
@@ -36,7 +36,7 @@ export const LinkMenuItem: React.FunctionComponent<MenuItemProps<LinkMenuItemDef
 
   return (
     <li className={className}>
-      <a href="#" onClick={onClick} className={item.classNames}>
+      <a href="#" onClick={onClick}>
         {item.iconName && <Icon name={item.iconName} className="icon--menuitem" />}
         <span className="menuitem-label">{item.label}</span>
       </a>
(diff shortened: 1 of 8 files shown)


kaedroho:
test

kaedroho:
test

kaedroho:
test

lb-:
Not sure on the status of this, but I think that adding classes to menu items would be a valuable feature, for example a post I wrote relies on it 
https://dev.to/lb/how-to-build-an-interactive-guide-for-users-in-the-wagtail-cms-admin-2dcp

Note: This also relies on adding data attributes which will no longer work but still I think adding some DOM attributes (at a minimum classes) is a very simple thing and a very powerful way to support customisations from those working with Wagtail.

thibaudcolas:
@lb- I think the status of this is up in the air. We have more work planned on the sidebar in the next release. If there is value in adding arbitrary attributes to those menu items – sounds like something we could go ahead and implement?

lb-:
I personally think dropping the ability to make your own MenuItem with any html is a bit of a backwards step. In 2.15 we added a formal Wagtail Component (Python driven) for a consistent approach to html with JS things.  but then in 2.16 we removed the ability to use html in our menus  

All that being said, I understand why and I also know the new sidebar has been in the code for some time and have not seen similar feedback. Also, as per other discussions, at the moment our only option is to make new things with React and it does not play nicely with arbitrary html provided in templates. 

So maybe classes and data attributes are a middle ground - additional JS can be hooked in (to some extent) this way but anything too complex will just be fighting with the React re-renders (e.g. adding additional content in a menu item won't work but adding event listeners may work sometimes). 

Anyway - rant over and happy to help with whatever direction we go towards.

lb-:
Just deleted my comment, probably not the best place to have a rant. 

or now I think keeping classes supported would be great. Data attributes would be a nice enhancement on the new sidebar to support some kinds of customisations. 

We still have the custom branding and other template overrides for other ways to get custom content in if needed.

kaedroho:
Going to close this as it didn't make it in to 2.17 so would be awkward to remove, but I think we should leave this as an undocumented/unsupported feature as classnames are quite difficult to maintain backwards-compatibility for.

PR #7279, opened 2021-06-23
Respect embed cache_until (fixes #6664)

### LINKED ISSUE #6664: Respect cache_until when retrieving embeds

thenewguy:
Fixes #6664


gasman:
Thanks @thenewguy! Merged in a64eb1f2f3d63e3366460495e8de69b1794f1e71.

PR #7285, opened 2021-06-24
Ensure comparison view picks up request/instance-specific EditHandler customisations

gasman:
Some edit handlers, such as the 'unofficial' PerUserContentPanels recipe from #4749, vary their field list according to the current request/instance by hooking into bind_to. This was not being called on the comparison view, meaning that when these edit handlers are in use, the field list was never getting populated and so the view was wrongly reporting no changes.

Note that the bind_to method also allows binding a form, which we do still skip (since the comparison view doesn't construct one).


kaedroho (approved):

gasman:
Merged in ac00945d3bebae836c01f99ce51b5d41172a81e0

PR #7291, opened 2021-06-28
Add headings for all settings

### LINKED ISSUE #6887: New documentation search lacking results for Wagtail settings
With the new Algolia DocSearch merged (#6833), there are a few cases where the results aren’t as good as they could be – in this instance, Wagtail settings such as `WAGTAILIMAGES_INDEX_PAGE_SIZE`, because they’re only used in code blocks, which aren’t indexed.

We should add a heading for each setting on https://docs.wagtail.io/en/stable/reference/settings.html, so the setting variable is in the non-code-block content, and so we can link to the documentation about a specific setting. I would suggest keeping the existing headings as-is, and just adding sub-headings, so we can keep all existing links.

### Additional context

- Current DocSearch results: https://docs.wagtail.io/en/latest/search.html?q=WAGTAILIMAGES_INDEX_PAGE_SIZE
- Previous RTD results: https://docs.wagtail.io/en/stable/search.html?q=WAGTAILIMAGES_INDEX_PAGE_SIZE

thibaudcolas:
Fixes #6887. I chose to add headings for all settings, with the exception of `WAGTAILFRONTENDCACHE_LOCATION` since it seems to not be recommended anymore, and is present in the textual content of the page already.

There were a lot of cases where the current heading would be very close / identical to the one I added, for example `Tag Limit` and `TAG_LIMIT`. I chose to add a heading in those cases nonetheless so we consistently have settings names as `<h3>`, while `<h2>` is for sections (even if they only contain a single heading). I think it looks a bit neater in the side navigation.

Test page: [ReadTheDocs deploy preview](https://wagtail--7291.org.readthedocs.build/en/7291/reference/settings.html).

I didn’t test this with the search indexing as I don’t think there is a way – but we already know this will work, as existing "settings headings" are properly indexed, such as [`WAGTAIL_I18N_ENABLED`](https://wagtail--7291.org.readthedocs.build/en/7291/search.html?q=WAGTAIL_I18N_ENABLED).

* ~~For front-end changes: Did you test on all of [Wagtail’s supported environments](https://docs.wagtail.io/en/latest/contributing/developing.html#browser-and-device-support)?~~

### DIFF
--- a/docs/reference/settings.rst
+++ b/docs/reference/settings.rst
@@ -58,16 +58,6 @@ Search
 
 Define a search backend. For a full explanation, see :ref:`wagtailsearch_backends`.
 
-``WAGTAILSEARCH_RESULTS_TEMPLATE``, ``WAGTAILSEARCH_RESULTS_TEMPLATE_AJAX``
----------------------------------------------------------------------------
-
-.. code-block:: python
-
-  WAGTAILSEARCH_RESULTS_TEMPLATE = 'myapp/search_results.html'
-  WAGTAILSEARCH_RESULTS_TEMPLATE_AJAX = 'myapp/includes/search_listing.html'
-
-Override the templates used by the search front-end views.
-
 .. _wagtailsearch_hits_max_age:
 
 ``WAGTAILSEARCH_HITS_MAX_AGE``
--- a/docs/reference/settings.rst
+++ b/docs/reference/settings.rst
@@ -7,6 +7,9 @@ Wagtail makes use of the following settings, in addition to :doc:`Django's core 
 Site Name
 =========
 
+``WAGTAIL_SITE_NAME``
+---------------------
+
 .. code-block:: python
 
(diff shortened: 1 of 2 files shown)


docs/reference/settings.rst
```diff
@@ -49,6 +58,9 @@ Search
 
 Define a search backend. For a full explanation, see :ref:`wagtailsearch_backends`.
 
+``WAGTAILSEARCH_RESULTS_TEMPLATE``, ``WAGTAILSEARCH_RESULTS_TEMPLATE_AJAX``
```
gasman, line 61: These can be removed (and also from `docs/advanced_topics/add_to_django_project.rst`) - search views were moved into the project template long ago, so there's nothing to override from the wagtailsearch app.

gasman (approved):

docs/reference/settings.rst
thibaudcolas: Nice catch :) Removing

PR #7292, opened 2021-06-28
Feature add facebook instagram oembed v11 support

tkremmel:

Yes
Yes
Yes - I adjust existing tests.
I checked the docs and I think the existing documentation in `/docs/embeds.rst` is sufficient.

### DIFF
--- a/docs/advanced_topics/embeds.rst
+++ b/docs/advanced_topics/embeds.rst
@@ -213,7 +213,7 @@ the `oEmbed Read` feature. Furthermore the app must be reviewed and accepted
 by Facebook. You can find the announcement in the API `changelog
 <https://developers.facebook.com/docs/graph-api/changelog/version11.0/#oembed>`.
 Apps that activated the `oEmbed Product` before June 8, 2021 need to activate
-the `oEmbed Read` feature and review their app until September 7, 2021. 
+the `oEmbed Read` feature and review their app until September 7, 2021.
 
 Once you have your app access tokens (App ID and App Secret), add the Facebook and/or
 Instagram finders to your ``WAGTAILEMBEDS_FINDERS`` setting and configure them with
--- a/docs/advanced_topics/embeds.rst
+++ b/docs/advanced_topics/embeds.rst
@@ -199,17 +199,17 @@ For example, this is how you can instruct Youtube to return videos in HTTPS
 Facebook and Instagram
 ----------------------
 
-As of September 2021, your Facebook app has to be reviewed and accepted by
-Facebook. You can find the announcement in the API `changelog
-<https://developers.facebook.com/docs/graph-api/changelog/version11.0/#oembed>`.
-
 As of October 2020, Facebook deprecated their public oEmbed APIs. If you would
 like to embed Facebook or Instagram posts in your site, you will need to
 use the new authenticated APIs. This requires you to set up a Facebook
 Developer Account and create a Facebook App that includes the oEmbed Product.
-Instructions for creating the neccessary app are in the requirements sections of the
+Instructions for creating the necessary app are in the requirements sections of the
 `Facebook <https://developers.facebook.com/docs/plugins/oembed>`_
 and `Instagram <https://developers.facebook.com/docs/instagram/oembed>`_ documentation.
+
+As of September 2021, your Facebook app has to be reviewed and accepted by
+Facebook. You can find the announcement in the API `changelog
+<https://developers.facebook.com/docs/graph-api/changelog/version11.0/#oembed>`.
 
 Once you have your app access tokens (App ID and App Secret), add the Facebook and/or
 Instagram finders to your ``WAGTAILEMBEDS_FINDERS`` setting and configure them with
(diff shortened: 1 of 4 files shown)


zerolab:
For reference: https://developers.facebook.com/docs/graph-api/changelog/version11.0/#oembed
Also, according to https://developers.facebook.com/docs/graph-api/changelog/version9.0 v9 will be supported until the beginning of 2023

tkremmel:
> For reference: https://developers.facebook.com/docs/graph-api/changelog/version11.0/#oembed
> Also, according to https://developers.facebook.com/docs/graph-api/changelog/version9.0 v9 will be supported until the beginning of 2023

@zerolab Thanks for the hint. We started working on this PR because we thought that as of 7. September 2021 one has to use a Facebook app that has been reviewed AND one has to use the v11 api version. After your comment we started to discuss internally on how to interpret the  change log and it seems that one just has to use a reviewed Facebook app, but can continue to use the v9 api version until 2023.

So it is probably not urgent to merge this PR 😅

zerolab:
oEmbed Read

> Applies to all versions.

As far as I can tell, the difference after September is that you need to enable "oEmbed Read" in the registered Facebook app (and I suppose complete the [app review](https://developers.facebook.com/docs/app-review) step).

This PR still makes sense, but as you say, not an urgent one

gasman:
So, if I'm understanding correctly:

* Anyone who set up the old oEmbed product before June 8, 2021 can continue using it for now, but needs to switch to the oEmbed Read feature (which requires app review) before September 7, 2021.
* Anyone setting up oEmbed right now is supposed to use the oEmbed Read feature. (The message on https://developers.facebook.com/docs/plugins/oembed implies that the old oEmbed product is unavailable to new apps created after June 8, 2021, although it doesn't say so outright...)
* The above is true regardless of whether you're using the `v9.0` or `v11.0` URL endpoints.

It would be worth double-checking that last point, I think. @cnk - am I right in thinking that you have an existing site set up using the pre-June-2021 oEmbed product, that hasn't gone through Facebook app review? If so, please could you check whether it still works with the URL endpoint changed to `v11.0`?

If that does work, then it looks like there's no reason not to go ahead and merge this now. If it _doesn't_ work, we should hold off on merging it until after September (by which time Facebook will have forced everyone to switch to the new oEmbed Read feature).

Either way, it sounds like the mention of September 2021 in the docs here should be fixed (=> as of June 2021 for new apps, and September 2021 for existing ones), and it would also be a good idea to add a note about the Facebook policy change in the 2.14 release notes. (While that change isn't tied to any code changes on the Wagtail side or any particular Wagtail upgrade, it's something that developers will need to know about.)

cnk:
Unfortunately I only had a Facebook/Instagram app for testing the Wagtail embed code. Since I haven't been using it, it expired and now I will need to verify my identity before being able to use it. As part of that verification process, FB is automatically upgrading me to v 11. So sorry. No testing help here.

tkremmel:
@gasman We have one "old" Facebook/Instagram app. Unfortunately it has not the _old oEmbed product_ activated, and the old oEmbed product is also not available for selection.

When using the credentials of this app to fetch the embed code, we get the following response:

```
{
	"error": {
		"message": "(#10) To use 'Oembed Read', your use of this endpoint must be reviewed and approved by Facebook. To submit this 'Oembed Read' feature for review please read our documentation on reviewable features: https://developers.facebook.com/docs/apps/review.",
		"type": 'OAuthException',
		"code": 10,
		'fbtrace_id': 'xxx'
	}
}
```

The response is the same for the api endpoints `v9.0` and `v11.0`.

So I would say that it is safe to say that for the time being Facebooks oembed permission handling is independent from whether one is using the `v9.0` or the `v11` api endpoint.

So I think the [comment](https://github.com/wagtail/wagtail/pull/7292#issuecomment-869740361) from @zerolab is still true.

> As far as I can tell, the difference after September is that you need to enable "oEmbed Read" in the registered Facebook app (and I suppose complete the app review step).
> This PR still makes sense, but as you say, not an urgent one

tkremmel:
@gasman I updated the docs in this [commit](https://github.com/wagtail/wagtail/pull/7292/commits/d2fd2fe79ffeeb595d9bbe5a7015b5137371554c).

I hope you are fine with the changes.

cnk:
> the `oEmbed Read` feature and review their app until September 7, 2021. 

@tkremmel do you mean "until"? I think you mean: 

Apps that activated the `oEmbed Product` before June 8, 2021 need to activate the `oEmbed Read` feature and submit their app for review before September 7, 2021.

tkremmel:
@cnk You are absolutely right. English is - as easily detectable - not my mother tongue. Thanks for the correction! Fixed in this [commit](https://github.com/wagtail/wagtail/pull/7292/commits/e6c072a0e8b6fa7f4f8960edad5818fc2612fea7).

gasman:
Thanks @tkremmel! Merged in be857cc5e98a73023249198b98fe54838ed9397c , with upgrade consideration note added to the 2.14 release notes.

Issue #7248, opened 2021-06-11
Nested StructBlocks in a StreamBlock are not showing the ValidationError in the field


jkevingutierrez:
Found a bug? Please fill out the sections below. 👍

### Issue Summary

In Wagtail 2.13.1, having a required field in a nested StructBlock shows the message "The page could not be saved due to validation errors" at the top of the page, but it doesn't show the message "This field is required." in the failing field.

Looks like the issue is happening with any ValidationError, not just the required error. I also tested it with a custom RegexValidator and it didn't show the error in the field

![image](https://user-images.githubusercontent.com/6220055/121612387-ecc40b80-ca1f-11eb-875a-a7510b127c76.png)

![image](https://user-images.githubusercontent.com/6220055/121612373-e5046700-ca1f-11eb-8d9f-c87e8d6990ec.png)

### Steps to Reproduce

1. Create a StructBlock with some required fields. eg:
```
class CopyBlock(blocks.StructBlock):
    text = blocks.RichTextBlock(required=True)
```
2. Create a StructBlock that will accept other struct blocks using a StreamBlock. eg:
```
class GridLayoutBlock(blocks.StructBlock):
    body = blocks.StreamBlock(
        [
            ('copy', CopyBlock()),
        ]
    )
```
3. Following the example, add GridLayoutBlock to some page StreamField
```
class CustomPage(Page):
    body = StreamField(
        [
            ('grid', GridLayoutBlock()),
        ]
    )
```
4. Create a CustomPage, add a GridLayoutBlock and inside that add a CopyBlock filling the text field.
5. Refresh the page and delete the text field data. Try to save again. It will show the "The page could not be saved due to validation errors" message at the top of the page, but it won't show the message "This field is required." in the field

Also, it doesn't "count" that error in the CONTENT tab.

![image](https://user-images.githubusercontent.com/6220055/121612529-29900280-ca20-11eb-9734-2b1fa0aecd5e.png)

Expected behaviour:

It should show the "This field is required" message in the empty field.

![image](https://user-images.githubusercontent.com/6220055/121612669-770c6f80-ca20-11eb-8531-04d1018a465d.png)

And count the error in the CONTENT tab

![image](https://user-images.githubusercontent.com/6220055/121612624-61974580-ca20-11eb-835f-b7b996c2db6d.png)

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: (yes / no) yes

### Technical details

* Python version: Run `python --version`. 3.9.5
[...]

PR #7295, opened 2021-06-28
Correctly handle nulls in ListBlock validation errors

gasman:
Fixes @cnk's test case from #7248. blockErrors within a ListBlockValidationError is an array (with nulls for the items with no errors), but setError was treating it as a dict keyed by block index, which meant it was erroneously passing nulls to the child setError method. FieldBlock handles this gracefully, but other blocks such as StructBlock don't.

### DIFF
--- a/client/src/components/StreamField/blocks/ListBlock.js
+++ b/client/src/components/StreamField/blocks/ListBlock.js
@@ -130,12 +130,13 @@ export class ListBlock extends BaseSequenceBlock {
     }
     const error = errorList[0];
 
-    // eslint-disable-next-line no-restricted-syntax
-    for (const blockIndex in error.blockErrors) {
-      if (error.blockErrors.hasOwnProperty(blockIndex)) {
-        this.children[blockIndex].setError(error.blockErrors[blockIndex]);
+    // error.blockErrors = a list with the same length as the data,
+    // with nulls for items without errors
+    error.blockErrors.forEach((blockError, blockIndex) => {
+      if (blockError) {
+        this.children[blockIndex].setError(blockError);
       }
-    }
+    });
   }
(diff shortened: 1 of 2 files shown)


cnk:
I tried this as a patch in my work project and this fixes the issue I was seeing.

client/src/components/StreamField/blocks/ListBlock.js
```diff
@@ -130,12 +130,13 @@ export class ListBlock extends BaseSequenceBlock {
     }
     const error = errorList[0];
 
-    // eslint-disable-next-line no-restricted-syntax
-    for (const blockIndex in error.blockErrors) {
-      if (error.blockErrors.hasOwnProperty(blockIndex)) {
-        this.children[blockIndex].setError(error.blockErrors[blockIndex]);
+    // error.blockErrors = a list with the same length as the data,
+    // with nulls for items without errors
+    error.blockErrors.forEach((blockError, blockIndex) => {
```
kaedroho, line 135: Does the forEach here work on IE11?

kaedroho (commented):
Looks good to me! Just wondering if this switching to forEach an effect on IE11 support

client/src/components/StreamField/blocks/ListBlock.js
gasman: We're using forEach extensively elsewhere in the StreamField implementation, so if that's causing problems with IE11, that ship has sailed by now :-) Doesn't babel / core-js deal with that, though? (Our webpack.config.js adds `client/src/utils/polyfills.js` to all bundles, which in turn imports `core-js/shim`.)

kaedroho (approved):

PR #7298, opened 2021-06-29
Prevent comments CSS from overriding custom colours

### LINKED ISSUE #7294: Elements that were styled with custom colours in 2.12.x revert to wagtail teal in 2.13.x
### Issue Summary

We have added a custom theme to wagtail by setting a primary colour hue to a colour that matches our brand. We have just updated to 2.13.2, and some elements have become unthemed

### Steps to Reproduce

1. install wagtail 2.12.3 (also tested with 2.12.5)

2. create a css file containing

```css
:root {
    --color-primary-hue: 237;
}
```

3. include it with `@hooks.register("insert_global_admin_css")` as per documentation

4. Observe that e.g. the tab bar on the edit page is Blue matching the hue specified

5. Update to Wagtail 2.13.0 (also tested with 2.13.2)

Expected: the e.g. the tab bar on the edit page should remain as it was in wagtail 2.12
Actual:  certain UI elements (possibly anything set to "primary-color-darker"?) are now teal

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: No

### Technical details

* Python version: 3.7.9
* Django version: 2.2.24
* Wagtail version: 2.13.0,2.13.2
* Browser version: Chrome 91.0.4472.114

gasman:
Fixes #7294. CommentApp/main.scss mistakenly redefined the SCSS colour variables to their Wagtail defaults, overriding the ability to customise them as per https://docs.wagtail.io/en/stable/advanced_topics/customisation/admin_templates.html#custom-user-interface-colors

The changes I've made here are limited to removing the duplicate declarations (additional variables in CommentApp/main.scss that weren't in Wagtail's main _variables.scss have been left in place) and changing hard-coded colours within the commenting CSS into their closest variables (so that they respect custom colour schemes rather than clashing horribly). This could probably be done better by a real front-end person - in particular, when the primary colour is darkened as per #7294 then the inactive comment icon becomes nearly invisible and probably ought to have a custom variable of its own - but this deals with the regression, at least.

### DIFF
--- a/client/src/components/CommentApp/main.scss
+++ b/client/src/components/CommentApp/main.scss
@@ -1,55 +1,14 @@
-$color-teal: #007d7e;
-$color-teal-darker: darken(adjust-hue($color-teal, 1), 4);
-$color-teal-dark: darken(adjust-hue($color-teal, 1), 7);
+@import '../../../scss/settings/variables';
 
-$color-blue: #71b2d4;
-$color-red: #cd3238;
-$color-red-dark: #b4191f;
 $color-red-very-dark: #901419;
-$color-orange: #e9b04d;
-$color-orange-dark: #bb5b03;
-$color-green: #189370;
-$color-green-dark: #157b57;
-$color-salmon: #f37e77;
-$color-salmon-light: #fcf2f2;
-$color-white: #fff;
-$color-black: #000;
 $color-amber-0: #e9b04d;
 $color-amber-1: #faecd5;
 
-// darker to lighter
-$color-grey-1: darken($color-white, 80);
-$color-grey-2: darken($color-white, 70);
 $color-grey-25: #626262;
-$color-grey-3: darken($color-white, 15);
-$color-grey-4: darken($color-white, 10);
-$color-grey-5: darken($color-white, 2);
 $color-grey-7: #f2f2f2;
 $color-grey-8: #fbfbfb;
 
-$color-fieldset-hover: $color-grey-5;
-$color-input-border: $color-grey-4;
-$color-input-focus: lighten(desaturate($color-teal, 40), 72);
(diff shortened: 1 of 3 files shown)


client/src/components/CommentApp/main.scss
```diff
@@ -1,56 +1,15 @@
-$color-teal: #007d7e;
-$color-teal-darker: darken(adjust-hue($color-teal, 1), 4);
-$color-teal-dark: darken(adjust-hue($color-teal, 1), 7);
+@import '../../../scss/settings/variables';
 
-$color-blue: #71b2d4;
-$color-red: #cd3238;
-$color-red-dark: #b4191f;
 $color-red-very-dark: #901419;
```
zerolab, line 3: question: should these make it into variables.scss and make use of CSS vars as the other `$color-` ones?

zerolab: ah, looking at https://github.com/wagtail/wagtail/blob/bd091b3784171da59cd30c3c155860c0eaa5525f/client/scss/settings/_variables.scss -- not all use css vars

gasman: Yep, there's definitely a second phase of work needed to make the commenting app a "good citizen" regarding colour customisations - rather than just existing alongside them without breaking - but I think that requires insights into the front-end code that I don't have, for example knowing when these definitions belong in the main _variables.css, and whether they should have their own CSS vars or be derived from others with `darken()` etc.

jacobtoppm: Regarding being a "good citizen", I think this is true with much of Wagtail's CSS - we have comparatively few CSS variables in general, almost all non-teals aren't able to be modified. This feels like an improvement to be made across the admin, to be honest.

jacobtoppm (approved):

Issue #7143, opened 2021-04-30
UI for the delete button is confusing


Ermite-bossu:
### Is your proposal related to a problem?

When hovering the delete button, the color change from red to green, and it's pretty confusing

### Describe the solution you'd like

As you can see on my exemple, the best practice is that the delete button is filled with red instead of green

_P.S. I'm a graphic designer and a noob with Github, please don't flog me_ 

![Good_Bad_button](https://user-images.githubusercontent.com/83450946/116711950-a4064500-a9d3-11eb-93bc-68596b2e9584.gif)

tbrlpld:
This appears to be closed by #7302

PR #7184, opened 2021-05-11
Update _listing.scss

mcmikashi:
solution for the  #7143 issue
adding the scss selector 
https://github.com/wagtail/wagtail/issues/7143#issue-872721877

Thanks for contributing to Wagtail! 🎉

Before submitting, please review the contributor guidelines <https://docs.wagtail.io/en/latest/contributing/index.html> and check the following:

* Do the tests still pass? (https://docs.wagtail.io/en/latest/contributing/developing.html#testing)
* Does the code comply with the style guide? (Run `make lint` from the Wagtail root)
* For Python changes: Have you added tests to cover the new/fixed behaviour?
* For front-end changes: Did you test on all of [Wagtail’s supported environments](https://docs.wagtail.io/en/latest/contributing/developing.html#browser-and-device-support)?
    * **Please list the exact browser and operating system versions you tested**.
    * **Please list which assistive technologies you tested**.
* For new features: Has the documentation been updated accordingly?

PR #7302, opened 2021-06-30
Fix delete button color with no class (#7143)

BrandonMurch:
This pull request is to fix bug #7143.  Building on #7184 from @mcmikashi . I added the no class to the delete button within the snippets list. This allows the delete button to have the proper color both when hovered and not.

### DIFF
--- a/wagtail/snippets/wagtail_hooks.py
+++ b/wagtail/snippets/wagtail_hooks.py
@@ -58,5 +58,6 @@ def register_snippet_listing_buttons(snippet, user, next_url=None):
             _('Delete'),
             reverse('wagtailsnippets:delete', args=[model._meta.app_label, model._meta.model_name, quote(snippet.pk)]),
             attrs={'aria-label': _("Delete '%(title)s'") % {'title': str(snippet)}},
-            priority=20
+            priority=20,
+            classes={'no'}
         )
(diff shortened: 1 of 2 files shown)


wagtail/snippets/wagtail_hooks.py
```diff
@@ -58,5 +58,6 @@ def register_snippet_listing_buttons(snippet, user, next_url=None):
             _('Delete'),
             reverse('wagtailsnippets:delete', args=[model._meta.app_label, model._meta.model_name, quote(snippet.pk)]),
             attrs={'aria-label': _("Delete '%(title)s'") % {'title': str(snippet)}},
-            priority=20
+            priority=20,
+            classes={'no'}
```
kaedroho, line 62: Could we use a list instead of a set? This is for consistency, also I think that sets don't maintain ordering which would be important when using multiple classes

kaedroho (changes requested):
Thanks @BrandonMurch!

This doesn't appear to fix #7143 though. This adds the red border to the delete buttons on snippets but it still changes to teal when you hover. This issue also happens in model admin (see Bread Categories on the bakery demo) which is where that issue was initially reported

BrandonMurch:
Thanks @kaedroho for having a look at this. I have made both requested changes. The button is red on hover in both snippets, and model admin.

kaedroho (approved):
Looks good to me, thanks!

PR #7304, opened 2021-06-30
`select_related` collection on image listing

RealOrangeOne:
`images/results.html` shows the collection if it's available, which results in `O(n)` queries.

Before: 70 queries
After: 50 queries

(20 being the page size, so definitely working)

Makes a minimal difference to performance (~200ms locally), but hey, every little helps!


zerolab (approved):
LGTM!

(as an aside, I used this very view over a year or so ago when I was demoing django-silk and never got to submit a PR 🤦‍♂️)

PR #7305, opened 2021-07-01
PostgreSQL backend

ACMCMC:
This is the full implementation of the PostgreSQL backend for the search app. There's a new migration that will have to be edited later, when there's support for the MySQL and SQLite backends as well.


wagtail/search/utils.py
```diff
@@ -179,3 +180,8 @@ def get_search_fields(search_fields):
         elif isinstance(search_field, RelatedFields):
             for sub_field in get_search_fields(search_field.fields):
                 yield sub_field
+
+
+def get_postgresql_connections():
```
kaedroho, line 185: Looks like this is unused

wagtail/contrib/postgres_search/apps.py
```diff
@@ -1,13 +1,20 @@
...
+
 from .utils import get_postgresql_connections, set_weights
 
 
 class PostgresSearchConfig(AppConfig):
     name = 'wagtail.contrib.postgres_search'
 
     def ready(self):
+
+        warnings.warn('The postgres_search app will be deprecated in a future release. Please update your search backend to \'wagtail.search.backends.database\', which will pick up a Postgres backend automatically.', RemovedInWagtail217Warning)
```
kaedroho, line 16: I think we should change ``postgres_search`` to ``'wagtail.contrib.postgres_search'`` since this is how it would be registered in settings files

kaedroho (changes requested):
Only a couple of minor comments, but otherwise looks great! Nice work!

wagtail/search/utils.py
ACMCMC: It's used in some tests (`wagtail.search.tests.test_backends.TestBackendLoader.test_get_search_backends_with_no_default_defined, for example`)

wagtail/contrib/postgres_search/apps.py
ACMCMC: Changed! 🙏🏻

PR #7307, opened 2021-07-02
Refactor StreamField action buttons into their own objects

gasman:
Refactor the implementations of the move-up / move-down / duplicate / delete buttons on StreamBlock / ListBlock children so that they exist as distinct objects, rather than being hard-coded in BaseSequenceChild. This will potentially make it possible for external code to define additional action buttons (although I've not yet decided on the neatest way to expose this as an external API - a global `window.wagtailStreamField.ACTION_BUTTONS` array that third-party code can append to perhaps?)

### DIFF
--- a/client/src/components/StreamField/blocks/BaseSequenceBlock.js
+++ b/client/src/components/StreamField/blocks/BaseSequenceBlock.js
@@ -22,6 +22,22 @@ class ActionButton {
 
   render(container) {
     $(container).append(this.dom);
+
+    if (this.enableEvent) {
+      this.sequenceChild.on(this.enableEvent, () => {
+        this.enable();
+      });
+    }
+
+    if (this.disableEvent) {
+      this.sequenceChild.on(this.disableEvent, () => {
+        this.disable();
+      });
+    }
+
+    if (this.initiallyDisabled) {
+      this.disable();
+    }
   }
 
   enable() {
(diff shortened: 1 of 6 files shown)


client/src/components/StreamField/blocks/BaseSequenceBlock.js
```diff
@@ -2,23 +2,111 @@
...
+  constructor(sequenceChild) {
+    this.sequenceChild = sequenceChild;
+  }
+
+  render(container) {
+    const label = this.sequenceChild.strings[this.labelIdentifier] || this.labelIdentifier;
+
+    this.dom = $(`
+      <button type="button" class="c-sf-block__actions__single" title="${h(label)}">
+        <i class="icon icon-${h(this.icon)}" aria-hidden="true"></i>
```
zerolab, line 18: question: in light of #6107, would this not be the perfect place to make the switch for the StreamField action buttons? Doing it now would mean less end-dev maintenance

gasman: Makes sense, yep... I hadn't made the connection that this was still using the icon font! I guess that involves coming up with some new infrastructure for passing SVGs to front-end code though - maybe something like the 'strings' dict for translations. (Which, come to think of it, is also not brilliantly placed to serve 'pluggable' mechanisms like this, where the third-party code will want to use strings / icons outside of the set that's been [picked out by the core code](https://github.com/wagtail/wagtail/blob/9f3e6585afd07bab15c22257ecf4ed4eab5ab0dd/wagtail/core/blocks/stream_block.py#L593-L598).)

zerolab: it does indeed. the regular templates have `{% icon name=".." %}` which uses https://github.com/wagtail/wagtail/blob/main/wagtail/admin/templates/wagtailadmin/shared/icon.html

I suppose adding an icon class similar to https://github.com/wagtail/wagtail/blob/main/client/src/components/Icon/Icon.tsx would solve that bit.

kaedroho:
> (although I've not yet decided on the neatest way to expose this as an external API - a global window.wagtailStreamField.ACTION_BUTTONS array that third-party code can append to perhaps?)

If we make this a global constant, will there be a problem if a bundle loads late? For example, if someone modifies it in a ``DOMContentLoaded`` event handler, or uses ``async``/``defer`` on their script tag. Perhaps a global function would be better as that would be able to insert it into the streamfield if it's already loaded?

kaedroho (approved):

PR #7309, opened 2021-07-02
`select_related` user profile when listing users

RealOrangeOne:
The `avatar_url` template tag uses the `wagtail_userprofile` to find an attached avatar (if there is one), which was resulting in `n` queries.

Before: 45 queries in 1.06s
After: 26 queries in 0.63s


gasman:
Thanks @RealOrangeOne!

PR #7313, opened 2021-07-05
Fix crash when user copies a page from a section they can't publish

### LINKED ISSUE #7293: Error while copying pages when the user does not have publishing permissions for the original location
### Issue Summary

Error during article copy if you have a particular set of permissions - if the given user has only editing (and no publishing) permissions on the page tree they copying FROM, but has publication permissions on the page tree they copying TO, then the copy form omits the "alias" checkbox, thus the copy action fails with the following error message:  

`Traceback (most recent call last):
  File "C:\virtualenv\wagtail211\lib\site-packages\django\core\handlers\exception.py", line 47, in inner
  File "C:\virtualenv\wagtail211\lib\site-packages\django\core\handlers\base.py", line 179, in _get_response
  File "C:\virtualenv\wagtail211\lib\site-packages\django\views\decorators\cache.py", line 44, in _wrapped_view_func
  File "C:\virtualenv\wagtail211\lib\site-packages\wagtail\admin\urls\__init__.py", line 127, in wrapper
  File "C:\virtualenv\wagtail211\lib\site-packages\wagtail\admin\auth.py", line 193, in decorated_view
  File "C:\virtualenv\wagtail211\lib\site-packages\wagtail\admin\auth.py", line 60, in wrapped_view_func
  File "C:\virtualenv\wagtail211\lib\site-packages\wagtail\admin\views\pages\copy.py", line 56, in copy
    if can_publish and form.cleaned_data['alias']:

Exception Type: KeyError at /admin/pages/7/copy/
Exception Value: 'alias'`

### Steps to Reproduce

2. Create a simple model (for example "ArticlePage", even without a content field), and in the admin, create 2 page trees (lets name them Tree A and Tree B) with an another page under Tree A
3. Create a group where you give view/edit permissions to Tree A, and view/edit/publish permission to Tree B
4. Create a new admin user and assign the group you created
5. Now navigate to Tree A, and try to copy the article to Tree B
6. You will get the error message after the copy form (which does not show the 'alias' checkbox)
7. If you give publish permissions to Tree A, then the alias checbox appears upon copy, and everything works

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: YES

### Technical details

* Django version: 3.1.12
* Wagtail version: 2.12.5 and 2.13.2

kaedroho:
Fixes #7293.

We probably should rethink the UX here. The "alias" and "publish subpages" checkboxes show/hide based on their permissions on the source page but not the destination.

### DIFF
--- a/wagtail/admin/tests/pages/test_copy_page.py
+++ b/wagtail/admin/tests/pages/test_copy_page.py
@@ -556,7 +556,7 @@ class TestPageCopy(TestCase, WagtailTestUtils):
         # treebeard should report no consistency problems with the tree
         self.assertFalse(any(Page.find_problems()), 'treebeard found consistency problems')
 
-    def test_page_copy_alias_post_without_destination_publish_permission(self):
+    def test_page_copy_alias_post_without_source_publish_permission(self):
         # Check for issue #7293 - If the user has permission to publish at a destination, but not the source.
         # Wagtail would crash on attempt to copy
 
--- a/wagtail/admin/tests/pages/test_copy_page.py
+++ b/wagtail/admin/tests/pages/test_copy_page.py
@@ -555,3 +555,35 @@ class TestPageCopy(TestCase, WagtailTestUtils):
 
         # treebeard should report no consistency problems with the tree
         self.assertFalse(any(Page.find_problems()), 'treebeard found consistency problems')
+
+    def test_page_copy_alias_post_without_destination_publish_permission(self):
+        # Check for issue #7293 - If the user has permission to publish at a destination, but not the source.
+        # Wagtail would crash on attempt to copy
+
+        # Create a new section
+        self.destination_page = self.root_page.add_child(instance=SimplePage(
+            title="Destination page",
+            slug='destination-page',
+            content="hello",
+            live=True,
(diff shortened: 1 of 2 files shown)


wagtail/admin/tests/pages/test_copy_page.py
```diff
@@ -555,3 +555,35 @@ def test_page_copy_alias_post_copy_subpages(self):
 
         # treebeard should report no consistency problems with the tree
         self.assertFalse(any(Page.find_problems()), 'treebeard found consistency problems')
+
+    def test_page_copy_alias_post_without_destination_publish_permission(self):
```
gasman, line 559: Name is the wrong way round:
```suggestion
    def test_page_copy_alias_post_without_source_publish_permission(self):
```

gasman (changes requested):
Looks good to me other than the test name.

Agreed that checking publish permission on the source page doesn't seem to make sense - happy to fix the immediate bug and save that for a future fix.

gasman (approved):

gasman:
Merged in 7d237092d19d08c40dcdfada4983641c9ad5afea - thanks @kaedroho!

PR #7315, opened 2021-07-05
Use `select_related` when viewing the Site History report

### LINKED ISSUE #7284: Site history report n+1s
While checking Scout APM for a client project with lots of different authors and overall activity, I've noticed that the site history report has a few N+1s.

wagtail-log-views-report-n-plus-one

zerolab:
Fixes #7284

With the build that generated the n+1 trace from #7284 (my local build still at 2.12). Before:
~1s overall
163ms on queries
196 queries

after:
528ms overall
101ms on queries
46 queries

With my bakery setup. Before:

1090ms overall
165ms on queries
170 queries

After:
649ms overall
38ms on queries
31 queries

### DIFF
--- a/wagtail/admin/views/reports/audit_logging.py
+++ b/wagtail/admin/views/reports/audit_logging.py
@@ -88,7 +88,7 @@ class LogEntriesView(ReportView):
                 PageLogEntry.objects.filter(deleted=True).values('page_id')
             ))
 
-        return PageLogEntry.objects.filter(q)
+        return PageLogEntry.objects.filter(q).select_related('page', 'user', 'user__wagtail_userprofile')
 
     def get_action_label(self, action):
         from wagtail.core.log_actions import page_log_action_registry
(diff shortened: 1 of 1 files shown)


wagtail/admin/views/reports/audit_logging.py
```diff
@@ -88,7 +86,11 @@ def get_queryset(self):
                 PageLogEntry.objects.filter(deleted=True).values('page_id')
             ))
 
-        return PageLogEntry.objects.filter(q)
+        # Using prefech_related() on page, as select_related() generates an INNER JOIN,
+        # which filters out entries for deleted pages
+        return PageLogEntry.objects.filter(q).select_related(
+            'user', 'user__wagtail_userprofile'
+        ).prefetch_related('page')
```
RealOrangeOne, line 93: [Looks like](https://github.com/wagtail/wagtail/blob/main/wagtail/core/models/__init__.py#L3944) `page` is a ForeignKey, so you should be able to `select_related` it here instead, resulting in 1 fewer query

RealOrangeOne (approved):

zerolab:
@gasman would be curious to see if this approach works on your PoC model audit log branch

PR #7318, opened 2021-07-07
Improve admin listing performance

tomusher:
This change removes 4 N+1 queries on Wagtail admin page listings, reducing the number of queries on a 'fully loaded' (50 items) listing view in my bakerydemo test from 300+ to 46.

With Workflow disabled, there is still a notable improvement here, reducing from ~200 queries to 44.

I am not a huge fan of the pattern of annotating querysets and later accessing these 'cached' values in properties later on, which introduces duplication and indirection, so I'm hoping for some feedback on the implementation here.

* For Python changes: Have you added tests to cover the new/fixed behaviour? No, pending discussion

### DIFF
--- a/wagtail/admin/views/pages/listing.py
+++ b/wagtail/admin/views/pages/listing.py
@@ -1,7 +1,6 @@
 from django.conf import settings
 from django.core.paginator import Paginator
-from django.db.models import Count, Prefetch
-from django.db.models.expressions import Exists, OuterRef
+from django.db.models import Count
 from django.shortcuts import get_object_or_404, redirect
 from django.template.response import TemplateResponse
 from django.urls import reverse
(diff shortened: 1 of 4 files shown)


wagtail/core/query.py
```diff
@@ -418,6 +420,53 @@ def not_translation_of(self, page, inclusive=False):
         """
         return self.exclude(self.translation_of_q(page, inclusive))
 
+    def with_workflow_state(self):
```
kaedroho, line 423: Since these methods do slightly different things, could we replace ``with_`` with something more descriptive?

For example, ``prefetch_workflow_states()``, ``annotate_approved_schedule``, etc?

tomusher: Thanks Karl - that makes sense. I've renamed them and moved them to the appropriate view file.

wagtail/admin/views/pages/listing.py
```diff
@@ -84,6 +132,13 @@ def index(request, parent_page_id=None):
     for hook in hooks.get_hooks('construct_explorer_page_queryset'):
         pages = hook(parent_page, pages, request)
 
+    # Annotate queryset with various states to be used later for performance optimisations
+    pages = annotate_site_root_state(
+        annotate_approved_schedule(
+            prefetch_workflow_states(pages)
```
gasman, line 138: Could we skip `prefetch_workflow_states` if WAGTAIL_WORKFLOW_ENABLED is false?

wagtail/core/models/__init__.py
```diff
@@ -450,7 +450,12 @@ def is_site_root(self):
 
         This includes translations of site root pages as well.
         """
-        return Site.objects.filter(root_page__translation_key=self.translation_key).exists()
+        if hasattr(self, "_is_site_root"):
```
gasman, line 453: I'm not enormously keen on the Page model code having special knowledge of an optimisation that only exists in wagtail.admin view code. Perhaps we could make `prefetch_workflow_states`, `annotate_approved_schedule` and `annotate_site_root_state` into methods of PageQuerySet instead (they'll probably need the imports of Site / WorkflowState / PageRevision to be inside the function bodies to avoid a circular import, but that should be fine), and add a comment here to clarify that that's where this attribute comes from?

gasman (changes requested):

wagtail/core/models/__init__.py
tomusher: Sure, this is how had it [originally](https://github.com/wagtail/wagtail/pull/7318/commits/870949cd4074a09f8c4b86df309e6046d209e4f2) but Karl [suggested otherwise](https://github.com/wagtail/wagtail/pull/7318#discussion_r667759669), as they're likely only to be used in this view.

kaedroho: Sorry I missed this bit!

If we move them back, could we make it clear by the naming/docstrings which properties on the ``Page`` model are optimised by each method?

tomusher: I've moved these back to the queryset and clarified where they are set in docstrings/comments in both directions.

gasman (approved):

PR #7320, opened 2021-07-07
Use date format from current locale for xlsx export

thatguystone:
The current default looked broken to me when I opened up a report in Excel. Django has locales with date formats, so it might be a good idea to use that instead of a hard-coded default.

Some implementation notes:

* Excel has very limited date formatting capabilities, so some cases that django provides can't be covered.
* `SHORT_DATETIME_FORMAT` seems to provide a reasonable middle ground for coverage and usability.
* flake8 hated the new class, but this is the most readable version I could come up with (that didn't use `__getattr__`).

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -23,6 +23,7 @@ Changelog
  * Support calling `get_image_model` and `get_document_model` at import time (Matt Westcott)
  * When copying a page, default the 'Publish copied page' field to false (Justin Slay)
  * Open Preview and Live page links in the same tab, except where it would interrupt editing a Page (Sagar Agarwal)
+ * Added `ExcelDateFormatter` to `wagtail.admin.views.mixins` so that dates in Excel exports will appear in the locale's `SHORT_DATETIME_FORMAT` (Andrew Stone)
  * Fix: Delete button is now correct colour on snippets and modeladmin listings (Brandon Murch)
  * Fix: Ensure that StreamBlock / ListBlock-level validation errors are counted towards error counts (Matt Westcott)
  * Fix: InlinePanel add button is now keyboard navigatable (Jesse Menn)
--- a/docs/releases/2.15.rst
+++ b/docs/releases/2.15.rst
@@ -34,6 +34,7 @@ Other features
  * Support calling ``get_image_model`` and ``get_document_model`` at import time (Matt Westcott)
  * When copying a page, default the 'Publish copied page' field to false (Justin Slay)
  * Open Preview and Live page links in the same tab, except where it would interrupt editing a Page (Sagar Agarwal)
+ * Added ``ExcelDateFormatter`` to ``wagtail.admin.views.mixins`` so that dates in Excel exports will appear in the locale's ``SHORT_DATETIME_FORMAT`` (Andrew Stone)
 
 Bug fixes
 ~~~~~~~~~
--- a/wagtail/admin/tests/test_reports_views.py
+++ b/wagtail/admin/tests/test_reports_views.py
@@ -3,11 +3,14 @@ import datetime
 from io import BytesIO
 
 from django.conf import settings
+from django.conf.locale import LANG_INFO
 from django.test import TestCase
+from django.test.utils import override_settings
 from django.urls import reverse
-from django.utils import timezone
+from django.utils import timezone, translation
 from openpyxl import load_workbook
 
+from wagtail.admin.views.mixins import ExcelDateFormatter
 from wagtail.core.models import Page, PageLogEntry
 from wagtail.tests.utils import WagtailTestUtils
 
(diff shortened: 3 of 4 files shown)


thatguystone:
A simpler solution might also be to use the date format "YYYY MMM. DD HH:MM:SS", which translates to eg. "2020 Jul. 29 09:36:12". With this format, at least, there's text disambiguation between month and day, so it should be more readable in more locales.

lb-:
@thatguystone  I have been taking a look at this and from what I can understand `default_date_format` is the format used when date values are stored in the generated rows.

Am I correct in understanding that the goal is:
* To ensure that an exported spreadsheet with dates, or datetimes, should be able to be opened by anyone and Excel can parse those dates and use its own formatter in Excel to present the dates to the user (or the user could highlight the column and format their own way).

My concern with the current PR approach is that the export of excel dates being pinned to Django locale could make things worse (e.g. exported on a US timezone server and attempting to open in the UK could mix up month/day values).

Maybe a better option is to keep a static format but align with the ISO standard, so that no matter where Excel is opened it should parse the date/time as the same value.

* Current: `"dd/mm/yy hh:mm:ss"`
* Proposed: `"yyyy-MM-dd HH:mm:ss"` (Note: I have not tested this) 

This would not solve the nuance of timezones, however that is not covered in the existing implementation and probably out of scope for a smaller change. To solve this, maybe it is worth pulling out the code:
```
            {
                "in_memory": True,
                "constant_memory": True,
                "remove_timezone": True,
                "default_date_format": "dd/mm/yy hh:mm:ss",
            },
```
And making a new method `get_xlsx_workbook_options` on `SpreadsheetExportMixin`, so that it can be overridden for more complex use cases. Unit tests would be needed for a change like this though.

Links
* https://en.wikipedia.org/wiki/ISO_8601
* https://stackoverflow.com/a/810118/8070948
* https://xlsxwriter.readthedocs.io/working_with_dates_and_time.html
* https://xlsxwriter.readthedocs.io/workbook.html (if adding a new method, would be good to reference this doc in it)

gasman:
@lb- I started to review this, but didn't quite get round to finishing the job :-)

The details confused me a bit at first, but the relevant bit of documentation is https://xlsxwriter.readthedocs.io/working_with_dates_and_time.html - in short, when we pass a Python date object to the Excel file as we're doing here, it will get saved as an unambiguous numeric format, so there's no risk of it misinterpreting an mm/dd/yyyy date as dd/mm/yyyy. The format we specify here is just cosmetic, for the benefit of the person viewing the file, and on that basis I think it does make sense to try and match the format from the current Django locale, as long as we're happy that the translation from Django to Excel's date format is suitably robust.

lb-:
@gasman ahhh, I see it now. Thanks

Ok in that case - @thatguystone are you able to add some unit tests for this change? We want to ensure we are able to handle multiple variants of date format strings that could be from Django/Python. 

Also, __getattr__ might actually make this implementation more readable, but that is not a biggie.

thatguystone:
I've added a test case that makes sure each locale that Django defines outputs _something_. I was toying with the idea of making sure that unsupported formats would cause the test to fail, but I don't know what value that would add since it's not really feasible to fix such failures due to excel's lack of support.

For the current locales, the following are generated:

```10 Sep 2021, 22:32
10 Sep. 2021 22:32
10 September 2021 22:32
10-9-2021 22:32
10.9.2021 22.32
10.9.2021. 22:32
10.09.2021 22:32
10.09.2021. 22:32
10.Sep.2021 22:32
10 Sep 2021 22:32
10-09-2021  22:32
10-09-2021 22.32
10-09-2021, 22:32
10.09.2021 22:32
10.09.2021 22:32
10.09.2021, 22:32
10/09/2021 22:32
10/09/2021 22:32
10/09/2021 10:32 PM
22:32 10-09-2021
09/10/2021 10:32 PM
2021-9-10 22:32
2021-09-10 22:32
2021.09.10. 22:32
2021/9/10،‏ 22:32
2021/09/10 22:32
2021年9月10日 22:32
```

lb-:
@thatguystone I did a quick test just now and not sure this is working as expected

In the image below, the one underneath is after checking out your branch (also rebased on main), you can see that the date format is not correct.

Before the change `11/09/21 04:05:42`
After the change is applied & download again `09/254/2021 04:09 am`

[workflow-report-2021-09-18(1).xlsx](https://github.com/wagtail/wagtail/files/7189819/workflow-report-2021-09-18.1.xlsx)
[workflow-report-2021-09-18.xlsx](https://github.com/wagtail/wagtail/files/7189820/workflow-report-2021-09-18.xlsx)
Screen Shot 2021-09-18 at 9 35 49 pm
.

I have attached the two excel files for reference, the one with `(1)` is after the code change applied.

thatguystone:
I just checked this in Libreoffice 7.1, and it looks fine to me (below). What's the software did you use?

Screenshot-1631979999

lb-:
@thatguystone thanks for the quick reply, I also just tested now on Excel (online) + Google Drive with the new date format and also logged out the format and it's all working fine.

The screenshots from above are from Mac Preview, so not sure why it is not using the date format correctly but actual Excel is working fine.

I will merge this in soon, I think this is looking good, thanks for adding the unit tests + reworking the formatter util

lb-:
Just pushed to your branch with release notes / changelog + a typo fix in the word `Formatter`, will check the CI checks go through and then merge in.

Thanks again @thatguystone

Issue #5300, opened 2019-05-15
Validation errors applied to ListBlocks inside StructBlocks don't get rendered.


coredumperror:
I've got a custom `clean()` method on my StructBlock subclass, `TestBlock`. `TestBlock` contains two child blocks: `text_field` (a CharBlock) and `list_field` (a ListBlock of CharBlocks). My custom `clean()` method ensures that the user provides at least three elements to `list_field`, otherwise it throws a `ValidationError`. Here's the code:

    class TestBlock(blocks.StructBlock):

        text_field = blocks.CharBlock(
            label='Text',
            max_length=255
        )
        list_field = blocks.ListBlock(
            blocks.CharBlock(max_length=255),
            default=[],
        )

        class Meta:
            label = 'Test Block'
            template = 'app/blocks/TestBlock.html'
            form_classname = 'test-block struct-block'

        def clean(self, value):
            result = super().clean(value)
            errors = {}
            list_field_count = len(value['list_field'])
            if list_field_count < 3:
                errors['list_field'] = ErrorList(['You must provide at least three.'])
            if errors:
                # The message here is arbitrary - StructBlock.render_form will suppress it
                # and delegate the errors contained in the 'params' dict to the child blocks
                raise ValidationError('TableBlock received invalid data', params=errors)
            return result

The problem is that while the validator will successfully mark the form with an error if the user provides less than 3 elements for `list_field`, the form will not render the "You must provide at least three" message anywhere. Here's a screenshot of what I mean:
<img width="1595" alt="Screen Shot 2019-05-15 at 3 34 56 PM" src="https://user-images.githubusercontent.com/413571/57814162-a2c8a880-7727-11e9-9b43-f23d5f2b57d5.png">

However, by changing `errors['list_field']` to `errors['text_field']` in the `clean()` method, you get this:
<img width="1606" alt="Screen Shot 2019-05-15 at 3 40 39 PM" src="https://user-images.githubusercontent.com/413571/57814272-07840300-7728-11e9-9c08-03eb34a0c3c7.png">

So the error rendering is functioning properly on CharBlock. It's just not working on ListBlock, for some reason.

komailhaider10p:
Any update on this issue? I am facing same issue.

loicteixeira:
Looking at the `ListBlock` source code, I reckon `errors` (in your code snippet) should be a list of `ErrorList` (instead of a simple `list`) and should have the same length as the number of elements (using `None` when there are no errors): 

[...]

PR #7322, opened 2021-07-08
Support non-block validation errors on ListBlock

gasman:
Ref #5300 - allow non-field errors to be attached to ListBlock validation errors, as we do for StreamBlock. As it stands Wagtail will not generate any of these (but the next step will be to use this for min_num / max_num validation as per #3914). Tested against bakerydemo with this block definition:

```
from django.forms.utils import ErrorList
from wagtail.core.blocks import (
    CharBlock, StreamBlock,
    ListBlock, ListBlockValidationError
)


class FooListBlock(ListBlock):
    error = 'At least one block must say "foo"'

    def clean(self, value):
        value = super().clean(value)
        if not any(child == 'foo' for child in value):
            raise ListBlockValidationError(non_block_errors=ErrorList([self.error]))
        return value

class BaseStreamBlock(StreamBlock):
    # ...
    foo_list = FooListBlock(CharBlock())
```

### DIFF
--- a/wagtail/core/blocks/list_block.py
+++ b/wagtail/core/blocks/list_block.py
@@ -17,16 +17,26 @@ __all__ = ['ListBlock', 'ListBlockValidationError']
 
 
 class ListBlockValidationError(ValidationError):
-    def __init__(self, block_errors):
-        self.block_errors = block_errors
-        super().__init__('Validation error in ListBlock', params=block_errors)
+    def __init__(self, block_errors=None, non_block_errors=None):
+        self.non_block_errors = non_block_errors or ErrorList()
+        self.block_errors = block_errors or []
+
+        params = {}
+        if block_errors:
+            params['block_errors'] = block_errors
+        if non_block_errors:
+            params['non_block_errors'] = non_block_errors
+        super().__init__('Validation error in ListBlock', params=params)
 
 
 class ListBlockValidationErrorAdapter(Adapter):
     js_constructor = 'wagtail.blocks.ListBlockValidationError'
 
     def js_args(self, error):
-        return [[elist.as_data() if elist is not None else elist for elist in error.block_errors]]
+        return [
+            [elist.as_data() if elist is not None else elist for elist in error.block_errors],
+            error.non_block_errors.as_data(),
+        ]
 
     @cached_property
     def media(self):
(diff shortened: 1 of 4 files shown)


wagtail/core/blocks/list_block.py
```diff
@@ -17,16 +17,26 @@
 
 
 class ListBlockValidationError(ValidationError):
-    def __init__(self, block_errors):
-        self.block_errors = block_errors
-        super().__init__('Validation error in ListBlock', params=block_errors)
+    def __init__(self, block_errors=None, non_block_errors=None):
+        self.non_block_errors = non_block_errors or ErrorList()
```
kaedroho, line 21: On ``StreamBlockValidationError``, we don't set defaults here. So wonder why we need these for list block?

wagtail/core/blocks/list_block.py
```diff
@@ -79,6 +89,7 @@ def value_omitted_from_data(self, data, files, prefix):
     def clean(self, value):
         result = []
         errors = []
+        non_block_errors = ErrorList()
```
kaedroho, line 92: ``non_block_errors`` doesn't seem to ever be set in this method

kaedroho (approved):
This looks good to me, just a couple of very minor comments

wagtail/core/blocks/list_block.py
gasman: I think StreamBlockValidationError is slightly in the wrong in this case - if non_block_errors isn't explicitly passed when constructing one, then the telepath adapter will fail when it tries to call `error.non_block_errors.as_data()`. That's not a big deal when validation errors are being thrown 99% of the time within StreamBlock.clean's own logic, but less ideal for things that are subclassing `clean`. (The fact that ListBlock's convention for block errors is slightly weirder - a list of mostly `None` values` - also means that letting user code avoid that detail is a good idea...)

wagtail/core/blocks/list_block.py
gasman: Yep, this is a bit of a placeholder right now until we have the min_num / max_num params to make use of it.

gasman:
Merged in e0ac8ae73ddba34bcfcc995c1e5ebb24de0a0e26

PR #7324, opened 2021-07-08
Add missing max_length on temporary thumbnail_url migration

### LINKED ISSUE #7323: Allow long URLs migration fails upgrading wagtail to 2.13.3
### Issue Summary

Hello, I'm upgrading an existing Wagtail site 2.7.3 to 2.13.3. All worked fine on my test database, but when running the migrations on production data I encountered the following error when `wagtailembeds.0008_allow_long_urls` was applied:

```
psycopg2.errors.StringDataRightTruncation: value too long for type character varying(200)
```

After a bit of digging, I found that my production data has a couple of Embed objects with `thumbnail_url` lengths greater than 200.

This seemed OK, as Wagtail 2.7.3 has `thumbnail_url` max length of 255:

```
thumbnail_url = models.URLField(max_length=255, null=True, blank=True)
```

I tried a quick fix by updating the `wagtailembeds.0008_allow_long_urls` migration by adding `max_length=255`. It seemed to fix the problem and the migration ran OK on my production data.

However, I'm not sure if this is the preferred or right solution:

```
migrations.AlterField(
    model_name="embed",
    name="thumbnail_url",
    field=models.URLField(
        blank=True,
        default="",
        max_length=255,
    ),
    preserve_default=False,
),
```

### Steps to Reproduce

I have only tried this on a existing Wagtail instance, but I suppose it could be recreated on a fresh (but older 2.7.3?) instance of Wagtail, creating an Embed object with a `thumbnail_url` greater that 200 in length and then migrating to the latest Wagtail (2.13.3).

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: No

### Technical details

* Python version: Python 3.6.4
* Django version: 3.2.5
* Wagtail version: 2.13.3
* Browser version: N/A

gasman:
Fixes #7323 - my previous fix in #6999 annoyingly missed [the `max_length=255` on the CharField](https://github.com/wagtail/wagtail/blob/4eb7c2c0198d786d33e2257436d346774de21698/wagtail/embeds/migrations/0005_specify_thumbnail_url_max_length.py#L16), so while the whole intent of this migration is to change the field to a TextField where max_length isn't an issue, it introduced a brief window where 200-255 char URLs became invalid. 😖

### DIFF
--- a/wagtail/embeds/migrations/0008_allow_long_urls.py
+++ b/wagtail/embeds/migrations/0008_allow_long_urls.py
@@ -34,6 +34,7 @@ class Migration(migrations.Migration):
             field=models.URLField(
                 blank=True,
                 default="",
+                max_length=255,
             ),
             preserve_default=False,
         ),


jacobtoppm (approved):
Looks good to me!

PR #7330, opened 2021-07-09
Update accessibility targets with links to known issues & audit

thibaudcolas:
This updates documentation we had already:

- Converts the table of assistive technologies to a list to hopefully make it a bit easier to understand, and add to.
- Adds Windows high-contrast mode, as we’re now testing with it and treating any issues as bugs

And the bigger change – moving our current progress to its own section, with different links. This is primarily intended so people contributing to Wagtail know where we’re at – there is a place to check what we want to support, and also links to what problems we know about already, and what we haven’t tested yet (in the audit).

The intention is also to make the work of the @wagtail/accessibility more transparent – having the majority of our progress in a spreadsheet is nice for auditing and reporting, but falls a bit short in openness compared to how we normally manage Wagtail work.

cc @wagtail/accessibility who I’d like to give feedback on this.

Preview link: [contributors – accessibility targets](https://wagtail--7330.org.readthedocs.build/en/7330/contributing/developing.html#accessibility-targets), [editor’s guide – assistive technologies](https://wagtail--7330.org.readthedocs.build/en/7330/editor_manual/browser_issues.html#assistive-technologies).

### DIFF
--- a/docs/contributing/developing.rst
+++ b/docs/contributing/developing.rst
@@ -213,25 +213,30 @@ Safari         Windows        All
 Accessibility targets
 ---------------------
 
-We want to make Wagtail accessible for users of a wide variety of assistive technologies. The specific standard we aim for is `WCAG2.1 <https://www.w3.org/TR/WCAG21/>`_, AA level. Wagtail’s administration user interface isn’t accessible at the moment (see `issue #4199 <https://github.com/wagtail/wagtail/issues/4199>`_), but here are specific assistive technologies we aim to test for, and ultimately support:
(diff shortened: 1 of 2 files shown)


Stormheg (approved):
I think this is spot-on @thibaudcolas. Thank you! 🙌 

One thing that comes to mind (out of scope for this PR) is that we could add a section in the contributing section on how to contribute to the accessibility of Wagtail. This could be fixing issues from the backlog or, even better,  being an assistive technology user willing to engage with the Wagtail accessibility team. Or are the docs not the right place to 'advertise' this?

thibaudcolas:
@Stormheg thank you 🙌 I like the idea of having this kind of call to action in the contributing docs. The other option would be our [CONTRIBUTING.md](https://github.com/wagtail/wagtail/blob/main/.github/CONTRIBUTING.md), or both. Let’s see what others think.

Issue #7111, opened 2021-04-27
Enforce the use of a single string formatting mechanism for translation source strings


loicteixeira:
### Is your proposal related to a problem?

Python has numerous ways of formatting strings which are all surfaced differently within Transifex.

Across the codebase, translation strings use different formatting: `{}`, `{0}`, `{named}`, `%s` and `%(named)s`. This can not only be confusing for new translation but all those tags aren’t treated equally in Transifex.

In addition to the different formatting styles, there is a case to be made for positional placeholders (i.e. {}, {0} and %s) which are problematic as it does not allow the target language to reorder them.

### Describe the solution you'd like

Enforce named and old style formatting:

```python
# In Python
_("Welcome to the %(site_name)s Wagtail CMS") % {"site_name": request.site.site_name}
```

```
<!-- In Templates -->
{% blocktrans %}Welcome to the {{ site_name }} Wagtail CMS{% endblocktrans %}
```

### Additional context

Formatting Option | Screenshot | Notes
-- | -- | --
{}, {0} and {named} |![Brackets_Positional_Empty](https://user-images.githubusercontent.com/6834938/116188305-801ed700-a727-11eb-98ea-a523dc32ed98.png)  ![Brackets_Positional_Numbers](https://user-images.githubusercontent.com/6834938/116188357-92991080-a727-11eb-8272-078f7655be3f.png) ![Brackets_Named](https://user-images.githubusercontent.com/6834938/116188377-97f65b00-a727-11eb-94ab-76bb471f0e9c.png) ![Brackets_Missing](https://user-images.githubusercontent.com/6834938/116188403-a17fc300-a727-11eb-87a4-c24e468d7f74.png)| There is no special treatment about them.When the name of a placeholder matches a glossary term, it displays an unrelevant warning.
%s and %(named)s |![PercentS_Positional](https://user-images.githubusercontent.com/6834938/116188451-b5c3c000-a727-11eb-893c-a02d97f7512f.png) ![PercentS_Named](https://user-images.githubusercontent.com/6834938/116188470-bf4d2800-a727-11eb-8878-ace745cfcd1b.png) ![PercentS_Question](https://user-images.githubusercontent.com/6834938/116188479-c4aa7280-a727-11eb-9175-37c60bf232b9.png)| The placeholders are clickable, it is clear they are not meant to be translated.Note that there were questions around the use of %s, but it was raised before placeholders were made clickable, so it probably isn’t an issue any more.
[...]

Issue #5539, opened 2019-08-26
ModelAdmin delete error 


roodie:
### Issue Summary

For some reason, if I have a model which is registered as a snippet, and also has it's custom ModelAdmin defined (don't ask...:-D), then if I try to delete an instance using the modeladmin, I get the following error message:

```python

File "C:\virtualenv\wagtail\lib\site-packages\django\core\handlers\exception.py" in inner
  34.             response = get_response(request)

File "C:\virtualenv\wagtail\lib\site-packages\django\core\handlers\base.py" in _get_response
  115.                 response = self.process_exception_by_middleware(e, request)

File "C:\virtualenv\wagtail\lib\site-packages\django\core\handlers\base.py" in _get_response
  113.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "C:\virtualenv\wagtail\lib\site-packages\django\views\decorators\cache.py" in _wrapped_view_func
  44.         response = view_func(request, *args, **kwargs)

File "C:\virtualenv\wagtail\lib\site-packages\wagtail\admin\urls\__init__.py" in wrapper
  102.             return view_func(request, *args, **kwargs)

File "C:\virtualenv\wagtail\lib\site-packages\wagtail\admin\decorators.py" in decorated_view
  34.             return view_func(request, *args, **kwargs)

File "C:\virtualenv\wagtail\lib\site-packages\wagtail\contrib\modeladmin\options.py" in delete_view
  397.         return view_class.as_view(**kwargs)(request)

File "C:\virtualenv\wagtail\lib\site-packages\django\views\generic\base.py" in view
  71.             return self.dispatch(request, *args, **kwargs)

File "C:\virtualenv\wagtail\lib\site-packages\django\utils\decorators.py" in _wrapper
  45.         return bound_method(*args, **kwargs)

File "C:\virtualenv\wagtail\lib\site-packages\django\contrib\auth\decorators.py" in _wrapped_view
  21.                 return view_func(request, *args, **kwargs)

File "C:\virtualenv\wagtail\lib\site-packages\wagtail\contrib\modeladmin\views.py" in dispatch
  713.         return super().dispatch(request, *args, **kwargs)

File "C:\virtualenv\wagtail\lib\site-packages\django\utils\decorators.py" in _wrapper
  45.         return bound_method(*args, **kwargs)

File "C:\virtualenv\wagtail\lib\site-packages\django\contrib\auth\decorators.py" in _wrapped_view
[...]

PR #7334, opened 2021-07-09
Implement min_num / max_num on ListBlock

### LINKED ISSUE #5300: Validation errors applied to ListBlocks inside StructBlocks don't get rendered.

gasman:
Includes both server-side and client-side validation.

Fixes #5300 and #2379; supersedes #3914.

### DIFF
--- a/wagtail/core/blocks/list_block.py
+++ b/wagtail/core/blocks/list_block.py
@@ -98,6 +98,16 @@ class ListBlock(Block):
             else:
                 errors.append(None)
 
+        if self.meta.min_num is not None and self.meta.min_num > len(value):
+            non_block_errors.append(ValidationError(
+                _('The minimum number of items is %d') % self.meta.min_num
+            ))
+
+        if self.meta.max_num is not None and self.meta.max_num < len(value):
+            non_block_errors.append(ValidationError(
+                _('The maximum number of items is %d') % self.meta.max_num
+            ))
+
         if any(errors) or non_block_errors:
             raise ListBlockValidationError(block_errors=errors, non_block_errors=non_block_errors)
 
@@ -172,6 +182,10 @@ class ListBlock(Block):
         # descendant block type
         icon = "placeholder"
         form_classname = None
+        min_num = None
+        max_num = None
+
+    MUTABLE_META_ATTRIBUTES = ['min_num', 'max_num']
 
 
 class ListBlockAdapter(Adapter):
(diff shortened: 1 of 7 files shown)


wagtail/core/blocks/list_block.py
```diff
@@ -98,6 +98,16 @@ def clean(self, value):
             else:
                 errors.append(None)
 
+        if self.meta.min_num is not None and self.meta.min_num > len(value):
+            non_block_errors.append(ValidationError(
+                _('The minimum number of items is %d') % self.meta.min_num
```
kaedroho, line 103: Could we standardise on the brace format? I think we use this most often elsewhere and it might make the translators job a bit easier if they didn't need to know about all the older formats

kaedroho (approved):
Looks good to me, just a minor comment about string formatting

wagtail/core/blocks/list_block.py
gasman: That's covered by #7111 - in the past the brace format caused problems because Transifex didn't validate that the placeholders matched in the translated string, leading to issues like https://github.com/wagtail/wagtail/issues/5539#issuecomment-527554273 - but that may have changed now. Either way, this is just copying the string that's already used in StreamBlock so it won't be adding any new work for translators.

gasman:
Merged in f98bd6870d4d60803a91578b598cb704c490c3f5

PR #7337, opened 2021-07-12
Ordinary

nimasmi:
This is a pedantic submission of some documentation and UI wording changes, based on the language guidance at https://docs.wagtail.io/en/latest/contributing/general_guidelines.html which prefers British English for user-facing text. _Regular_ in British English doesn't have the US English meaning of _usual_ or _ordinary_.

The page aliases UI uses the term "regular page" to refer to a page that is not an alias. This MR changes the terminology to refer to _ordinary_ pages. Elsewhere  I have also replaced some uses of _regular_ with _normal_, _basic_, _ordinary_, _conventional_, or _standard_, depending on the context.

One use of _regularly_ has been changed with _often_, based more on context than on a dialect difference.

The guidance on language is slightly more ambiguous about dialect choice for text that is not user facing and not a built-in language keyword. I've been conservative, so this MR does not include changes to:

- comments and docstrings in Python code
- strings in test suites
- Python variable names
- text in release notes, as they might be considered immutable

### DIFF
--- a/docs/advanced_topics/amp.rst
+++ b/docs/advanced_topics/amp.rst
@@ -223,7 +223,7 @@ Using a different page template when AMP mode is active
 -------------------------------------------------------
 
 You're probably not going to want to use the same templates on the AMP site as
-you do on the regular web site. Let's add some logic in to make Wagtail use a
+you do on the normal web site. Let's add some logic in to make Wagtail use a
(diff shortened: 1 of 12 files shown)


gasman (approved):

gasman:
Thanks @nimasmi!

PR #7340, opened 2021-07-14
[feat] Move actions bar to footer and add bulk actions to search page

KIRA009:
This pull request also includes commits that will be used by other pull requests to reuse certain logic for creating bulk action functionality for objects like documents, users and images.


client/scss/components/_listing.scss
```diff
@@ -487,6 +487,60 @@ ul.listing {
...
+
+        ul {
+            bottom: 50px;
+            left: 30px;
+            flex-direction: column;
+        }
+    }
+}
+
+.footer__container {
```
zerolab, line 512: This impacts the regular action bar on the edit form. I suggest you test the changes together. The best thing to do is namespace it in `.bulk-actions-choices`

Screenshot 2021-07-14 at 16 51 12

client/scss/components/_listing.scss
```diff
@@ -487,6 +487,60 @@ ul.listing {
     }
 }
 
+footer.bulk-actions-choices {
+    @include transition(bottom 0.5s ease 0.5s);
```
zerolab, line 491: This is very slow. Maybe just `bottom 0.1s`

zerolab: The action bar requires further work on mobile. Please use the edit form action bar as a guide.

Screenshot 2021-07-14 at 16 54 26
Screenshot 2021-07-14 at 16 54 18

The delete button hover text is all white, so unreadable :/ There's an existing issue with the wrong colours on hover in listings (#7143, fixed in #7302) so worth checking the solution and the overlap there

zerolab (changes requested):
Thank you for this @KIRA009 
left a number of comments on the look/interactions

wagtail/admin/templatetags/wagtailadmin_tags.py
```diff
@@ -510,19 +510,44 @@ def page_bulk_action_choices(context):
...
             attrs={'aria-label': action.aria_label},
-            priority=action.action_priority
+            priority=action.action_priority,
+            classes=action.classes
         ) for action in bulk_actions_list
     ]
 
+    if bulk_action_more_list:
+        more_button = ButtonWithDropdown(
+            label=_("MORE"),
```
zerolab, line 534:
```suggestion
            label=_("More"),
```

KIRA009: I kept it as 'MORE' to match the style of the page listing actions, where it is achieved using css

Stormheg: Could we do the same thing with css and commit Dan’s suggestion? Screen readers tend to spell words written in all capital letters, which is a bit confusing.

FlipperPA:
I'm not sure if this pain point has already been addressed, but wanted to mention it here.

Some of my users need to frequently move pages in the admin interface. We have several thousand pages, so this can be an arduous process: it is easy enough to find the page to be moved, but when hunting for the destination page, there is no search or filter options. Then have to manually click "NEXT" through the pages until the find the right destination parent page.

It would be fantastic if there was some sort of filtering / search that could allow them to target the set of potential destination pages. Happen to do a video demo if it would help, but I hope I've explained it well enough. Cheers.

client/src/entrypoints/admin/bulk-actions.js
```diff
@@ -32,30 +32,31 @@ function SelectBulkActionsCheckboxes(e) {
   if (e.target.checked) checkedState.checkedPages.add(+e.target.dataset.pageId);
   else {
     /* unchecks `Select all` checkbox as soon as one page is unchecked */
-    document.querySelector(`.${BULK_ACTION_SELECT_ALL_CHECKBOX_TH} input`).checked = false;
+    document.querySelectorAll(`.${BULK_ACTION_SELECT_ALL_CHECKBOX_TH} input`).forEach(_el => {
+      const el = _el;
```
jacobtoppm, line 36: Is this necessary?

client/src/entrypoints/admin/bulk-actions.js
```diff
@@ -32,30 +32,31 @@ function SelectBulkActionsCheckboxes(e) {
...
-    document.querySelector(`.${BULK_ACTION_CHOICES_DIV}`).classList.remove('u-hidden');
-    document.querySelector(`.${BULK_ACTION_SELECT_ALL_CHECKBOX_TH}`).setAttribute('colspan', '6');
+    document.querySelector(`.${BULK_ACTION_CHOICES_DIV}`).classList.remove('hidden');
   }
 
   if (checkedState.checkedPages.size === checkedState.numPages) {
     /* when all checkboxes in the page are checked */
-    document.querySelector(`.${BULK_ACTION_SELECT_ALL_CHECKBOX_TH} input`).checked = true;
+    document.querySelectorAll(`.${BULK_ACTION_SELECT_ALL_CHECKBOX_TH} input`).forEach(_el => {
+      const el = _el;
```
jacobtoppm, line 57: Similarly to above, is this reassignment necessary?

client/src/entrypoints/admin/bulk-actions.js
```diff
@@ -79,7 +80,7 @@ function SelectBulkActionsCheckboxes(e) {
 function selectAllPageIdsInListing() {
   checkedState.selectAllInListing = true;
   document.querySelector(`.${BULK_ACTION_NUM_PAGES_SPAN}`).
-    textContent = wagtailConfig.STRINGS.NUM_PAGES_SELECTED_ALL_IN_LISTING;
+    textContent = wagtailConfig.STRINGS.NUM_PAGES_SELECTED_ALL_IN_LISTING + '.';
```
jacobtoppm, line 89: Wagtail generally tries to do string operations without concatenation if possible - you could use a template literal instead?

zerolab:
Hey @FlipperPA,

Thank you for your suggestion! This point has not yet been addressed. This piece of work is happening through Google Summer of Code (GSoC) and we did make a point of re-using as much of the existing functionality as possible.
Will add it as a stretch goal. At the same time if you guys spent time looking into this, a PR against the single page move would be very welcome. And that can then be helpful for bulk actions too.

client/src/entrypoints/admin/bulk-actions.js
```diff
@@ -110,7 +111,10 @@ function FilterEventListener(e) {
     }
   } else {
     /* If filter string is empty, select all checkboxes */
-    document.querySelector(`.${BULK_ACTION_SELECT_ALL_CHECKBOX_TH}`).checked = true;
+    document.querySelectorAll(`.${BULK_ACTION_SELECT_ALL_CHECKBOX_TH}`).forEach(_el => {
+      const el = _el;
```
jacobtoppm, line 121: Similar to above.

client/src/entrypoints/admin/bulk-actions.js
```diff
@@ -139,14 +143,18 @@ function addBulkActionListeners() {
...
     });
-  document.querySelector(`.${BULK_ACTION_SELECT_ALL_CHECKBOX_TH}`).addEventListener('change', SelectBulkActionsFilter);
+  document.querySelectorAll(`.${BULK_ACTION_SELECT_ALL_CHECKBOX_TH} input`).forEach(el => {
+    el.addEventListener('change', SelectBulkActionsFilter);
+  });
   document.querySelectorAll(`.${BULK_ACTION_FILTERS_CLASS}`).forEach(
     elem => elem.addEventListener('click', FilterEventListener)
   );
-  document.querySelectorAll(`.${BULK_ACTION_CHOICES_DIV} > ul > li > a`).forEach(
+  document.querySelectorAll(`.${BULK_ACTION_CHOICES_DIV} ul > li > a`).forEach(
```
jacobtoppm, line 158: I don't really love this dependence on the html structure to this extent. We could consider using a data attribute on the choice itself, or at least simplify this selector.

wagtail/admin/templates/wagtailadmin/pages/index.html
```diff
@@ -20,6 +20,16 @@ <h1 class="visuallyhidden">Explorer</h1>
             {% url 'wagtailadmin_explore' parent_page.id as pagination_base_url %}
             {% paginate pages base_url=pagination_base_url %}
         {% endif %}
+        <footer class="footer bulk-actions-choices hidden">
+            <div class="footer__container">
+                <div class="bulk-actions-filter-checkbox">
+                    <input type="checkbox" aria-label="Bulk action checkbox" />
```
jacobtoppm, line 26: Aria labels need to be translatable too

jacobtoppm (changes requested):

client/src/entrypoints/admin/bulk-actions.js
KIRA009: A linting issue shows up if I try to modify the property of the function parameter

client/scss/components/_listing.scss
```diff
@@ -487,6 +487,66 @@ ul.listing {
     }
 }
 
+footer.bulk-actions-choices {
+    @include transition(bottom 0.1s ease 0.5s);
```
zerolab, line 491: discussed in our catch up today:

```suggestion
    @include transition(bottom 0.1s ease 0.1s);
```

or plain `@include transition(bottom 0.1s);` for a quicker/smoother transition

client/src/entrypoints/admin/bulk-actions.js
```diff
@@ -28,59 +28,66 @@ function SelectBulkActionsFilter(e) {
...
 
-  if (checkedState.checkedPages.size === checkedState.numPages) {
+  if (checkedState.checkedObjects.size === checkedState.numObjects) {
     /* when all checkboxes in the page are checked */
-    document.querySelector(`.${BULK_ACTION_SELECT_ALL_CHECKBOX_TH} input`).checked = true;
+    document.querySelectorAll(`.${BULK_ACTION_SELECT_ALL_CHECKBOX_TH} input`).forEach(_el => {
+      const el = _el;
+      el.checked = true;
+    });
+    if (checkedState.shouldShowAllInListingText) {
```
zerolab, line 60: discussed during catch up: when the number of items in the listing fits on one page, i.e. no pagination, the "Select all in listing" option should not appear as it is confusing.
The listing uses the paginator so you could check for next/prev page existence in https://github.com/wagtail/wagtail/pull/7340/files#diff-5ae4cc8a762e77181ca3abcd029bdd0e759928f484d7a72457749ad1b9980412R23 `pages.has_previous and pages.has_next`

wagtail/admin/widgets/button.py
```diff
@@ -65,12 +65,35 @@ def __init__(self, *args, **kwargs):
...
+    template_name = 'wagtailadmin/pages/listing/_button_with_dropdown.html'
+
+    def __init__(self, *args, **kwargs):
+        self.btn_classes = kwargs.pop('btn_classes', set())
+        self.buttons_data = kwargs.pop('buttons_data', [])
+        super().__init__(*args, **kwargs)
+
+    def get_context_data(self):
+        context = super().get_context_data()
+        context['btn_classes'] = self.btn_classes
```
zerolab, line 90: nitpick: although we use `.btn` in CSS, I feel this can be more explicit `button_classes`

zerolab (changes requested):
A few minor tweaks from my side based on functional testing. And a nitpick. Other than that LGTM

Stormheg (changes requested):
Thanks @KIRA009. Looks like you are on the right track with this PR! 🛤️  🚅  🚀 
The backend code for this looks good to me.

I've left a few tiny comments about the ARIA labels.

However, I'd like to request changes on the frontend. This looks slightly different than the design. Unless we're deviating from the design which I'm not aware of.

Here's the design for reference:
image

How this PR looks on squash:
image

Notable differences:
- The design has more whitespace and feels less 'cramped'.
- The bulk action buttons in the design are slightly different.
  - Your version has the delete button in red.
  - Eyeballing it, it looks like your version has a coloured border applied and slightly bolder font for the button labels.
- The 2 pages selected text in the design is bold.
- When some of the checkboxes are selected, the design features an indeterminate checkbox (with a minus sign). I'm not sure if the styles for this have been implemented in Wagtail, but I would guess not.

Stormheg:
@KIRA009 I'd be happy for you to make the requested frontend changes in a follow up PR if it means we can get this merged!

wagtail/admin/views/bulk_action.py
```diff
@@ -53,14 +53,16 @@ def get_queryset(cls, object_ids):
...
         return True
 
-    @classmethod
-    def execute_action(cls, objects):
+    def execute_action(self, objects):
         raise NotImplementedError("execute_action needs to be implemented")
 
     @classmethod
     def execute(cls, object_ids):
+        # this method can be called to execute the action without instantiating the class.
```
jacobtoppm, line 61: This feels like quite a confusing way of doing this to me: this should either be a classmethod, or it shouldn't be. What you could do is keep it a classmethod, but pass in the data it needs through keyword arguments, for instance.

wagtail/admin/templates/wagtailadmin/bulk_actions/confirmation/form_with_fields.html
```diff
@@ -0,0 +1,9 @@
+{% extends './form.html' %}
+
+{% block form_fields %}
+<ul class="fields">
```
jacobtoppm, line 4: Looks like this is pretty similar to the `form_as_ul` template - I would suggest reusing the existing one

wagtail/admin/templates/wagtailadmin/bulk_actions/confirmation/form.html
```diff
@@ -0,0 +1,6 @@
+<form action="{{ submit_url }}" method="POST">
```
jacobtoppm, line 1: You should also include the `non_field_errors` template

wagtail/admin/templates/wagtailadmin/bulk_actions/confirmation/go_back.html
```diff
@@ -0,0 +1,2 @@
+{% load i18n %}
+<a href="{{ next }}" class="button button-secondary">{% trans "Go back" %}</a>
```
jacobtoppm, line 2: This feels pretty slight for its own template - maybe move to the base one?

jacobtoppm (changes requested):

zerolab:
@Stormheg "Your version has the delete button in red." I have previously requested that the delete button is specifically red as it is a destructive action cc @benenright 

There is, of course, the whole red/green in the context of colour blindness (e.g. https://baselinehq.com/blog/colourblindness-information-ui-design-red-green-problems-tips-tricks.html) but I think this is something we should tackle throughout the UI. Did the accessibility group look into this particular issue (cc @thibaudcolas)

wagtail/admin/templates/wagtailadmin/bulk_actions/confirmation/go_back.html
KIRA009: I am not sure how that will work, since this is conditionally rendered, and the condition will be different for pages, documents and images. Defining it in the base template would mean we would have to overwrite it every time using 
`{{ block go_back }}{{ endblock }}` if we have to not display it.

thibaudcolas:
@zerolab as long as there is a confirmation step and some indication in the button label that the action is destructive ("Delete"), I think the current situation is ok. It would be nice for Wagtail’s teal and red to not have such close saturation and lightness, but it’s not for this PR to worry about.

One thing I wonder about accessibility-wise is the placement of the button – feels like a destructive action shouldn’t be in the middle of the group, as anything in the middle is more likely to be mis-clicked? I don’t think it’s that big of a deal though if there are reasons why this order makes sense.

Are the 4-5 side-by-side buttons going to work for languages that have longer words in english? For example all languages where there is no single-word equivalent of "Unpublish"?

client/src/entrypoints/admin/bulk-actions.js
```diff
@@ -4,83 +4,98 @@ const BULK_ACTION_PAGE_CHECKBOX_INPUT = 'bulk-action-checkbox';
 const BULK_ACTION_SELECT_ALL_CHECKBOX_TH = 'bulk-actions-filter-checkbox';
 const BULK_ACTION_FILTERS_CLASS = `${BULK_ACTION_SELECT_ALL_CHECKBOX_TH} .c-dropdown__item a`;
 const BULK_ACTION_CHOICES_DIV = 'bulk-actions-choices';
-const BULK_ACTION_NUM_PAGES_SPAN = 'num-pages';
-const BULK_ACTION_NUM_PAGES_IN_LISTING = 'num-pages-in-listing';
-const TABLE_HEADERS_TR = 'table-headers';
+const BULK_ACTION_NUM_OBJECTS_SPAN = 'num-objects';
+const BULK_ACTION_NUM_OBJECTS_IN_LISTING = 'num-objects-in-listing';
```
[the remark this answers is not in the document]
KIRA009: Should I add `data-` attributes to each of the above elements then?

wagtail/admin/templatetags/wagtailadmin_tags.py
```diff
@@ -502,27 +502,58 @@ def bulk_action_filters(context):
...
+            priority=action.action_priority,
+            classes=action.classes | {'bulk-action-btn'},
         ) for action in bulk_actions_list
     ]
 
+    if bulk_action_more_list:
+        more_button = ButtonWithDropdown(
+            label=_("More"),
+            attrs={
+                'target': '_blank', 'rel': 'noopener noreferrer',
```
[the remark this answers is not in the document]
KIRA009: Nope, it doesn't actually. I think I copy pasted these from the other buttons without noticing. Thanks for pointing it out.

wagtail/admin/templates/wagtailadmin/bulk_actions/footer.html
```diff
@@ -0,0 +1,13 @@
+{% load wagtailadmin_tags i18n %}
+<footer class="footer bulk-actions-choices hidden">
```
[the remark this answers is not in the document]
KIRA009: Yes, adding `visibility: hidden` will stop readers from reading this when hidden, and also make it not navigatable using keyboard

client/scss/overrides/_utilities.dropdowns.scss
```diff
@@ -57,6 +62,10 @@
 // =============================================================================
 // Dark theme
 // =============================================================================
+.t-dark {
+    position: relative;
+}
```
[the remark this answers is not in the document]
KIRA009: I did this because the default behavior made the 'More' dropdown in bulk actions slightly aligned to left of where it would look more natural, due to which I had to use the `left` css property (not used in any other dropdowns, hence why this never came up), and because of that I had to make the sure the parent container had `position: relative`. 

The default one doesn't look way out of line, so I will revert back to that, and if there are changes to be made, we can do it in a future pr

wagtail/admin/templates/wagtailadmin/bulk_actions/footer.html
```diff
@@ -0,0 +1,13 @@
...
+    <div class="footer__container">
+        <div class="bulk-actions-filter-checkbox">
+            <input type="checkbox" aria-label='{% trans "Bulk action checkbox" %}' />
+        </div>
+        <ul>{% bulk_action_choices bulk_action_register_hook %}</ul>
+        <span class="num-objects"></span>
+        {% if select_all_obj_text and objects.has_other_pages %}
+        <a class="num-objects-in-listing u-hidden" href='#'>{{ select_all_obj_text }}</a>
```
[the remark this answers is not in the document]
KIRA009: I used an `a` tag because the default style matched the one in the designs. Will change that

client/scss/components/_listing.scss
```diff
@@ -487,6 +487,66 @@ ul.listing {
     }
 }
 
+footer.bulk-actions-choices {
```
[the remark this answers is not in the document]
KIRA009: The only purpose of `.footer` is to override the default footer transition style

client/src/entrypoints/admin/bulk-actions.js
```diff
@@ -1,11 +1,12 @@
...
 
-const BULK_ACTION_PAGE_CHECKBOX_INPUT = 'bulk-action-checkbox';
-const BULK_ACTION_SELECT_ALL_CHECKBOX_TH = 'bulk-actions-filter-checkbox';
-const BULK_ACTION_FILTERS_CLASS = `${BULK_ACTION_SELECT_ALL_CHECKBOX_TH} .c-dropdown__item a`;
-const BULK_ACTION_CHOICES_DIV = 'bulk-actions-choices';
-const BULK_ACTION_NUM_OBJECTS_SPAN = 'num-objects';
-const BULK_ACTION_NUM_OBJECTS_IN_LISTING = 'num-objects-in-listing';
+const BULK_ACTION_PAGE_CHECKBOX_INPUT = '[data-bulk-action-checkbox]';
+const BULK_ACTION_SELECT_ALL_CHECKBOX = '[data-bulk-action-select-all-checkbox]';
+const BULK_ACTIONS_CHECKBOX_PARENT = '[data-parent-id]';
```
zerolab, line 5: suggestion: namespace them this and the vars below with `data-bulk-actions-` to ensure there is no clash with other elements (e.g. from packages, or possibly future core ones)

wagtail/admin/templates/wagtailadmin/bulk_actions/confirmation/form.html
zerolab: @jacobtoppm this seems to have been added. below `{% include 'wagtailadmin/shared/non_field_errors.html' %}`

jacobtoppm: Yep, I think this is resolved now!

wagtail/admin/views/pages/bulk_actions/move.py
```diff
@@ -72,28 +71,43 @@ def prepare_action(self, pages):
...
 
-        for page in pages:
-            if not page.permissions_for_user(cls.request.user).can_move_to(destination):
+    @classmethod
+    def execute_action(cls, objects, **kwargs):
+        destination = kwargs.get('destination', None)
+        if destination is None:
+            return
+        user = kwargs.get('user', None)
+        if user is None:
```
zerolab, line 101: note: we discussed that user is not mandatory in `execute_action` especially when used in a management command context. This and `destination` contradict that 🤔 (this applies to the other `execute_action`s
One to sort out with documentation in a separate PR

wagtail/admin/views/bulk_action.py
```diff
@@ -100,16 +99,22 @@ def get_context_data(self, **kwargs):
...
     def prepare_action(self, objects):
         return
 
-    def post(self, request):
+    def get_execution_context(self):
+        return {}
+
+    def form_valid(self, form):
+        request = self.request
+        self.cleaned_form = form
```
zerolab, line 117: nitpick: this is used to pass particular parameters (`include_descendants`, `move_applicable`, `chooser`). I would prefer we either set those directly in each class, or.. let's make this `self.form` cc @jacobtoppm

wagtail/admin/views/bulk_action.py
```diff
@@ -118,11 +123,14 @@ def post(self, request):
...
             after_hook_result = self.__run_after_hooks(self.action_type, request, objects)
             if after_hook_result is not None:
                 return after_hook_result
             success_message = self.get_success_message()
             if success_message is not None:
                 messages.success(request, success_message)
         return redirect(self.next_url)
+
+    def form_invalid(self, form):
+        return super().form_invalid(form)
```
zerolab, line 136: question: this does nothing (that `FormView` doesn't) , why is it required?

zerolab (changes requested):
99% there @KIRA009. a few additional notes from me

wagtail/admin/views/bulk_action.py
KIRA009: Yes, I kept it there for some custom logic, then decided against it, and forgot to remove it.

wagtail/admin/views/pages/bulk_actions/move.py
KIRA009: I agree that requiring to input user in the management command is not ideal, but the requirement of user cannot be abstracted out from the action, because other than permission checks we also need the user for logging purposes in some cases. 
Maybe instead of making the default as `None` in `kwargs.get('user', None)`, we can default to any superuser in the database.

wagtail/admin/views/bulk_action.py
KIRA009: are you suggesting to rename `cleaned_form` to `form`?

wagtail/admin/views/pages/bulk_actions/move.py
jacobtoppm: Which actions is the user required for? As far as I know, all of the fundamental operations they run don't require a user.

I don't think defaulting to a random superuser is a good idea here - that will make for very confusing logs.

zerolab: @jacobtoppm see `execute_action`s in this PR (delete, move, publish, unpublish)

KIRA009: Ah I assumed that for logging purposes, a user is must. Since that is not the case, in the cases where there is no 'user' in kwargs where there should have been, will it be safe to assume that the classmethod is being called from a management command, and treat all permission checks as true?

jacobtoppm: @zerolab Yep, I'm referring more to the operations in core that they're calling, sorry for the confusion

client/src/entrypoints/admin/bulk-actions.js
KIRA009: Should this be addressed in this pr, or a future pr?

zerolab:
We discussed that the the final comments will be addressed in a follow up PR. Thank you @KIRA009

Issue #6592, opened 2020-11-27
Hitting enter in a char block opens the page privacy menu


kaedroho:
Seems to happen on every Wagtail build I know of, including a new 2.11 one I just created with Wagtail start.

1) Create a page with a stream field, add an instance of a char block and save it
2) On the edit view, put your cursor in that char block and press enter
3) The page privacy modal will open

zerolab:
Did a quick check: the "set page privacy: element is a `button` which seems to take precedence over the page action buttons. Changing to an `a` element solves this

paqman85:
I'd like to help with this one if it's still open.

@zerolab I believe I found the element you are referencing.

wagtailadmin>pages>privacy_switch_panel.html
`        {% if page.id and page_perms.can_set_view_restrictions %}
            <button data-url="{% url 'wagtailadmin_pages:set_privacy' page.id %}" class="button action-set-privacy">
                {% trans "Set page privacy" %}
            </button>
        {% endif %}`

and this can be changed to 
`        {% if page.id and page_perms.can_set_view_restrictions %}
            <a data-url="{% url 'wagtailadmin_pages:set_privacy' page.id %}" class="button action-set-privacy">
                {% trans "Set page privacy" %}
            </a>
        {% endif %}`

to solve the issue. Is this correct?

rinti:
I would argue that a `<button>` is semantically correct here, for accessibility reasons, since it performs an action. The `<a>` tag should never be used without `href`.

zerolab:
@rinti I concur. that is why it was used in the first place. Did some more digging. In order to prevent `<button>` elements from capturing the enter key event, they need to be defined as `<button type="button>". TIL!

sagar-wal:
Hi ! Can I work on this ?

I tried it out locally, changing `<button>` to `<a>` , or to `<button type="button">` , both cause the 'Save Draft' button to be pressed on hitting Enter.

gasman:
Fixed in #6882

PR #7346, opened 2021-07-16
Converts inline panel anchor to button to make keyboard focusable.

jessemenn:
This allows the InlinePanel button to be keyboard navigable and targetable with VoiceOver.

        * macOS Chrome, Firefox ESR, Safari
        * VoiceOver on macOS with Safari

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/edit_handlers/inline_panel.html
+++ b/wagtail/admin/templates/wagtailadmin/edit_handlers/inline_panel.html
@@ -24,8 +24,8 @@
 </script>
 
 <p class="add">
-    <a class="button bicolor button--icon" id="id_{{ self.formset.prefix }}-ADD">
+    <button class="button bicolor button--icon" id="id_{{ self.formset.prefix }}-ADD">
         {% icon name="plus" wrapped=1 %}
         {% blocktrans with label=self.label|lower %}Add {{ label }}{% endblocktrans %}
-    </a>
+    </button>
 </p>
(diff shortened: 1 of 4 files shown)


wagtail/admin/templates/wagtailadmin/edit_handlers/inline_panel.html
```diff
@@ -24,8 +24,8 @@
 </script>
 
 <p class="add">
-    <a class="button bicolor button--icon" id="id_{{ self.formset.prefix }}-ADD">
+    <button class="button bicolor button--icon" id="id_{{ self.formset.prefix }}-ADD">
```
zerolab, line 27: suggestion: use

```suggestion
    <button type="button" class="button bicolor button--icon" id="id_{{ self.formset.prefix }}-ADD">
```

otherwise pressing enter anywhere on the form will trigger a submit. See #6592 for context

zerolab: Somehow my second half of the comment got lost somwhere -- just wanted to say thank you for this seemingly small change, but nonetheless quite important for making the Wagtail admin more accessible

jessemenn: Oops! I entirely forgot about that, good catch. Also makes any change to expandind_formset.js unnecessary. Thanks 😄

Stormheg:
Thanks @jessemenn 🚀 ⌨️ 

This looks good and I can confirm this solves the issue.

A few notes during testing with bakery demo, but out of scope for this PR:
- 'ADD' is spelled instead of pronounced as a word. I think this is a common issue in Wagtail in general.
- There is no feedback after clicking the button. I would expect focus to shift to the new item and VoiceOver to announce a new item has been added.

Stormheg (approved):

PR #7349, opened 2021-07-19
Add sidebar slim state preference

Stormheg:
The slim sidebar will now remember if it is collapsed ('slim') or not when the page is loaded.

Preferences are saved in the session.

**Note:** I've added `window.wagtailConfig.CSRF_TOKEN` as an easy way to get the CSRF token. Another way would be to read the `csrftoken` cookie set by Django but that could be set to httponly, making it impossible to read from JS.

### DIFF
--- a/client/src/components/Sidebar/index.tsx
+++ b/client/src/components/Sidebar/index.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import ReactDOM from 'react-dom';
 
 import { Sidebar } from './Sidebar';
+import { SidebarPreferencesDefinition } from './SidebarPreferences';
 
 export function initSidebar() {
   const element = document.getElementById('wagtail-sidebar');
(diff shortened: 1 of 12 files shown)


client/src/components/Sidebar/index.tsx
```diff
@@ -20,6 +21,12 @@ export function initSidebar() {
   if (element instanceof HTMLElement && element.dataset.props) {
     const props = window.telepath.unpack(JSON.parse(element.dataset.props));
 
+    const preferences: SidebarPreferencesDefinition = props.preferences;
+
+    if (preferences.collapsed) {
+      document.body.classList.add('sidebar-collapsed');
```
Stormheg, line 27: There is a slight issue with this code: it can sometimes run a bit late. This means that when the sidebar is collapsed and the page is loaded, a transition is visible where the main content expands to fill the whitespace left by the collapsed sidebar. This looks a bit odd and can be fixed by doing this check inside the Django template, not here in the frontend.

Stormheg:
@kaedroho and I have decided to take a different approach that uses cookies instead of a REST endpoint.

PR #7350, opened 2021-07-19
Bulk actions for documents

KIRA009:
This PR adds bulk actions for the Documents listing

image

### DIFF
--- a/wagtail/admin/templatetags/wagtailadmin_tags.py
+++ b/wagtail/admin/templatetags/wagtailadmin_tags.py
@@ -505,6 +505,7 @@ def bulk_action_filters(context):
 def bulk_action_choices(context, hook_name):
     corresponding_urls = {
         'register_page_bulk_action': 'wagtailadmin_page_bulk_action',
+        'register_document_bulk_action': 'wagtaildocs:document_bulk_action',
     }
     if hook_name not in corresponding_urls:
         return {'buttons': []}
--- a/wagtail/documents/templates/wagtaildocs/documents/list.html
+++ b/wagtail/documents/templates/wagtaildocs/documents/list.html
@@ -1,5 +1,6 @@
 {% load i18n %}
 <table class="listing">
+    <col width="10px" />
     <col />
     <col  />
     {% if collections %}
--- a/wagtail/documents/templates/wagtaildocs/documents/list.html
+++ b/wagtail/documents/templates/wagtaildocs/documents/list.html
@@ -8,6 +8,7 @@
     <col width="16%" />
     <thead>
         <tr class="table-headers">
+            {% include 'wagtailadmin/bulk_actions/select_all_checkbox_cell.html' %}
             <th>
                 {% if not is_searching %}
                     <a href="{% url 'wagtaildocs:index' %}{% if not ordering == "title" %}?ordering=title{% endif %}" class="icon icon-arrow-down-after {% if  ordering == "title" %}teal{% endif %}">
@@ -33,8 +34,10 @@
         </tr>
     </thead>
     <tbody>
+        {% trans "Document select checkbox" as checkbox_aria_label %}
         {% for doc in documents %}
             <tr>
+                {% include 'wagtailadmin/bulk_actions/listing_checkbox_cell.html' with obj_type='document' obj=doc aria_label=checkbox_aria_label %}
                 <td class="title">
                     {% if choosing %}
                         <div class="title-wrapper"><a href="{% url 'wagtaildocs:document_chosen' doc.id %}" class="document-choice">{{ doc.title }}</a></div>
--- a/wagtail/documents/templates/wagtaildocs/documents/list.html
+++ b/wagtail/documents/templates/wagtaildocs/documents/list.html
@@ -34,7 +34,7 @@
         </tr>
     </thead>
     <tbody>
-        {% trans "Document select checkbox" as checkbox_aria_label %}
+        {% trans "Select document" as checkbox_aria_label %}
         {% for doc in documents %}
             <tr>
                 {% include 'wagtailadmin/bulk_actions/listing_checkbox_cell.html' with obj_type='document' obj=doc aria_label=checkbox_aria_label %}
--- a/wagtail/documents/views/bulk_actions/add_to_collection.py
+++ b/wagtail/documents/views/bulk_actions/add_to_collection.py
@@ -33,11 +33,17 @@ class AddToCollectionBulkAction(DocumentBulkAction):
             'user': self.request.user
         }
 
-    def execute_action(self, documents):
-        self.collection = self.cleaned_form.cleaned_data['collection']
-        for document in documents:
-            self.num_parent_objects += 1
-            document.collection = self.collection
+    def get_execution_context(self):
+        return {
+            'collection': self.cleaned_form.cleaned_data['collection']
+        }
+
+    @classmethod
+    def execute_action(cls, objects, **kwargs):
+        cls.collection = kwargs.get('collection', None)
+        for document in objects:
+            cls.num_parent_objects += 1
+            document.collection = cls.collection
             document.save()
 
     def get_success_message(self):
--- a/wagtail/documents/views/bulk_actions/add_to_collection.py
+++ b/wagtail/documents/views/bulk_actions/add_to_collection.py
@@ -1,0 +1,55 @@
+from django import forms
+from django.utils.translation import gettext_lazy as _
+from django.utils.translation import ngettext
+
+from wagtail.core import hooks
+from wagtail.core.models.collections import Collection
+from wagtail.documents.permissions import permission_policy
+from wagtail.documents.views.bulk_actions.document_bulk_action import DocumentBulkAction
+
+
+class CollectionForm(forms.Form):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields['collection'] = forms.ModelChoiceField(
+            queryset=Collection.objects.all()
+        )
+
+
+class AddToCollectionBulkAction(DocumentBulkAction):
+    display_name = _("Add to collection")
+    action_type = "add_to_collection"
+    aria_label = _("Add documents to collection")
+    template_name = "wagtaildocs/bulk_actions/confirm_bulk_add_to_collection.html"
+    action_priority = 30
+
+    def check_perm(self, document):
+        return permission_policy.user_has_permission_for_instance(self.request.user, 'change', document)
+
+    def execute_action(cls, documents):
+        collection_id = int(cls.request.POST.get('collection'))
+        cls.collection_name = Collection.objects.get(id=collection_id).name
+        for document in documents:
+            cls.num_parent_objects += 1
+            document.collection_id = collection_id
+            document.save()
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context['form'] = CollectionForm()
+        return context
+
+    def get_success_message(self):
+        return ngettext(
+            "%(num_parent_objects)d document has been added to %(collection)s",
+            "%(num_parent_objects)d documents have been added to %(collection)s",
+            self.num_parent_objects
+        ) % {
+            'num_parent_objects': self.num_parent_objects,
+            'collection': self.collection_name
+        }
+
+
+@hooks.register('register_document_bulk_action')
+def add_to_collection(request):
+    return AddToCollectionBulkAction(request)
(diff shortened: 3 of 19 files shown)


wagtail/documents/views/bulk_actions/delete.py
```diff
@@ -0,0 +1,37 @@
...
+    action_type = "delete"
+    aria_label = _("Delete documents")
+    template_name = "wagtaildocs/bulk_actions/confirm_bulk_delete.html"
+    action_priority = 10
+    classes = {'serious'}
+
+    def check_perm(self, document):
+        return permission_policy.user_has_permission_for_instance(self.request.user, 'delete', document)
+
+    def execute_action(cls, documents):
```
jacobtoppm, line 20: Why is this `cls` rather than `self`? This doesn't seem to be a classmethod.

wagtail/documents/views/bulk_actions/add_to_collection.py
```diff
@@ -0,0 +1,55 @@
...
+
+class AddToCollectionBulkAction(DocumentBulkAction):
+    display_name = _("Add to collection")
+    action_type = "add_to_collection"
+    aria_label = _("Add documents to collection")
+    template_name = "wagtaildocs/bulk_actions/confirm_bulk_add_to_collection.html"
+    action_priority = 30
+
+    def check_perm(self, document):
+        return permission_policy.user_has_permission_for_instance(self.request.user, 'change', document)
```
jacobtoppm, line 27: I think this permissions check is incomplete. The user also needs the `add` permission for the new collection.

wagtail/documents/views/bulk_actions/add_to_collection.py
```diff
@@ -0,0 +1,55 @@
...
+from wagtail.core.models.collections import Collection
+from wagtail.documents.permissions import permission_policy
+from wagtail.documents.views.bulk_actions.document_bulk_action import DocumentBulkAction
+
+
+class CollectionForm(forms.Form):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields['collection'] = forms.ModelChoiceField(
+            queryset=Collection.objects.all()
```
jacobtoppm, line 15: The user might not have permissions for all of these collections - see below.

wagtail/documents/views/bulk_actions/add_to_collection.py
```diff
@@ -0,0 +1,55 @@
...
+    action_type = "add_to_collection"
+    aria_label = _("Add documents to collection")
+    template_name = "wagtaildocs/bulk_actions/confirm_bulk_add_to_collection.html"
+    action_priority = 30
+
+    def check_perm(self, document):
+        return permission_policy.user_has_permission_for_instance(self.request.user, 'change', document)
+
+    def execute_action(cls, documents):
+        collection_id = int(cls.request.POST.get('collection'))
```
jacobtoppm, line 30: I feel like we should start working more with Django's form system, using `is_valid()` and the like and then retrieving data rather than getting it directly from `request.POST` - it lets us handle errors a bit better and return an invalid form to the user.

jacobtoppm (changes requested):
Generally looking good, but I think the permissions checks need to be looked at

wagtail/documents/views/bulk_actions/add_to_collection.py
jacobtoppm: For example:
```
form = CollectionForm(request.POST) # This gets you a form with all of the data in
if form.is_valid(): # Check if the form is valid
    collection_id = form.cleaned_data['collection'] # retrieve data from form.cleaned_data instead
    # now do other stuff
else:
    # return a response with the form in context, render its errors in the template
```

To avoid boilerplate, you may find `FormView` or similar handy, see https://docs.djangoproject.com/en/3.2/topics/class-based-views/generic-editing/

In your particular case, you may need to use `user` in the form to construct the set of collections. You can make the constructor accept `user` as well to do this.

KIRA009: > To avoid boilerplate, you may find FormView or similar handy, see https://docs.djangoproject.com/en/3.2/topics/class-based-views/generic-editing/

Should this be changed at the base bulk action class, where we can add the `FormView` as another base class in addition to the `TemplateView`? Because more or less all actions have at least one input element

jacobtoppm: If you end up finding `FormView` useful then yes, I'd suggest it adding it at the base class level.

wagtail/documents/views/bulk_actions/delete.py
KIRA009: Actually it is a classmethod. It is used by another classmethod `execute` which is used to perform bulk actions using a management command. This can be changed to a simple method though.

wagtail/documents/views/bulk_actions/add_to_collection.py
KIRA009: The `user_has_permission_for_instance` [here](https://github.com/wagtail/wagtail/blob/99f05d67cb97465127c095fc2664ff161e1b049a/wagtail/core/permission_policies/collections.py#L254), checks for the 'add' permission as well, if 'change' permission is checked for. Is it necessary to mention explicitly?

wagtail/documents/views/bulk_actions/delete.py
jacobtoppm: Should it not get the classmethod decorator in that case?

wagtail/documents/views/bulk_actions/add_to_collection.py
jacobtoppm: Ah, bit of miscommunication here, I'm referring to the new collection. The user needs to have change permission on the existing instance in the old collection, but also needs to have permission to add a document to the new collection. (This doesn't necessarily have to be done here - it could be done by limiting the selectable collections in the form as I mention below - but it should be done somewhere)

wagtail/documents/views/bulk_actions/delete.py
KIRA009: Yes, I'll add that

zerolab:
@KIRA009 now that #7340 is merged, you can rebase this as well

wagtail/documents/templates/wagtaildocs/documents/list.html
```diff
@@ -33,8 +34,10 @@
         </tr>
     </thead>
     <tbody>
+        {% trans "Document select checkbox" as checkbox_aria_label %}
```
zerolab, line 37:
```suggestion
        {% trans "Select document" as checkbox_aria_label %}
```

wagtail/documents/views/bulk_actions/dispatcher.py
```diff
@@ -0,0 +1,14 @@
+from django.http.response import Http404
+
+from wagtail.core import hooks
+
+
+def index(request, action):
+    bulk_actions = hooks.get_hooks('register_document_bulk_action')
```
zerolab, line 7: suggestion: the body of this method is similar to that for pages, except for the hook. Maybe move it to base as 

```python
# admin/views/bulk_actions/dispatcher.py
def bulk_actions_index(request, action, hook):
    bulk_actions = hooks.get_hooks(hook)
    for action_func in bulk_actions:
        _action = action_func(request)
        if _action.action_type != action:
            continue
        return _action.dispatch(request)

    return Http404()
```

and then this could be something like

```python
from wagtail.admin.views.bulk_actions.dispatcher import bulk_actions_index

def index(request, action):
    return bulk_actions_index(request, action, "register_document_bulk_action")
```

ditto for the pages dispatcher

wagtail/documents/views/bulk_actions/document_bulk_action.py
```diff
@@ -0,0 +1,19 @@
+from wagtail.admin.views.bulk_action import BulkAction
+from wagtail.documents import get_document_model
+from wagtail.documents.permissions import permission_policy as documents_permission_policy
+
+
+class DocumentBulkAction(BulkAction):
```
zerolab, line 6: this and `PageBulkAction` don't seem to implement `display_name`, `action_type` nor `aria_label`.  feels like some tidy up is needed

wagtail/documents/views/bulk_actions/document_bulk_action.py
```diff
@@ -0,0 +1,19 @@
...
+    permission_policy = documents_permission_policy
+    model = get_document_model()
+    object_key = 'document'
+
+    def get_context_data(self, **kwargs):
+        ctx = super().get_context_data(**kwargs)
```
zerolab, line 12:
```suggestion
        context = super().get_context_data(**kwargs)
```

`ctx` -> `context`
a tad more vebose, but easier to parse

wagtail/documents/views/bulk_actions/document_bulk_action.py
```diff
@@ -0,0 +1,19 @@
...
+        ctx['documents_with_no_access'] = [
+            {
+                'document': document,
+                'can_edit': self.permission_policy.user_has_permission_for_instance(self.request.user, 'delete', document)
```
zerolab, line 16: I am not sure this is the correct permission here. "change" feels more suitable

zerolab (changes requested):
Taking great shape, @KIRA009 

A few changes and questions this time

wagtail/admin/templatetags/wagtailadmin_tags.py
```diff
@@ -505,6 +505,7 @@ def bulk_action_filters(context):
 def bulk_action_choices(context, hook_name):
     corresponding_urls = {
         'register_page_bulk_action': 'wagtailadmin_page_bulk_action',
+        'register_document_bulk_action': 'wagtaildocs:document_bulk_action',
```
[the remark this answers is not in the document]
KIRA009: Yes, I was going to address that soon. A new hook maybe?

wagtail/documents/views/bulk_actions/document_bulk_action.py
KIRA009: These aren't meant to be used as bulk actions, rather templates for the specific actions. Trying to create an object from either of these classes will raise an Exception

wagtail/documents/views/bulk_actions/document_bulk_action.py
KIRA009: Yes, I don't know why I put 'delete' there. Will change that

wagtail/documents/views/bulk_actions/dispatcher.py
KIRA009: I was thinking of maybe addressing this in a separate pr. I don't want this pr to contain refactoring commits which are not directly related to documents bulk actions functionality. The action-bar pr did become quite large because of that. Is that fine?

zerolab: sounds like a plan!

wagtail/admin/templatetags/wagtailadmin_tags.py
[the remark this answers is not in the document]
zerolab: It is probably our best bet. @jacobtoppm @Stormheg any objections?
(note: this would come in a separate PR)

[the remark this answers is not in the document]
jacobtoppm: Addressing this in a separate PR sounds good.

That said, I'm still wondering if we can do this action registration + url endpoints in a slightly more generic way - against a list of model classes for instance, and have a url that takes the model name + the bulk action or similar. Will need to put some more thought into it though.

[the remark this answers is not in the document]
KIRA009: Yeah structuring it that way would make it very easy to integrate bulk actions into modeladmin.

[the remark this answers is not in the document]
Stormheg: Hook sounds fine, this is how most customisations are made to Wagtail anyway. Looking forward to seeing that in a follow up PR 🙂

wagtail/documents/templates/wagtaildocs/documents/list.html
```diff
@@ -8,6 +8,7 @@
     <col width="16%" />
     <thead>
         <tr class="table-headers">
+            {% include 'wagtailadmin/bulk_actions/select_all_checkbox_cell.html' %}
```
Stormheg, line 11: These cells are wider than they should be for me.
image

I would expect them to be similarly sized as in the page explorer.

Stormheg (commented):
Thanks @KIRA009 - this looks good!

I have a few comments about the frontend, could you fix those before we merge?

The checkbox cells are really wide and there is also a filter with options that belong to the page explorer, see my comments.
One more thing is that the bulk action bar should use 'documents' instead of 'pages':
image

Stormheg (approved):
Thanks for making those changes @KIRA009 

Let's merge 😁 👍

PR #7352, opened 2021-07-20
Add a log action for copy_for_translation operation

kaedroho:
Currently, when a user copies a page for translation, the 'copy' log action is used.

This adds a more specific 'copy_for_translation' operation to be used instead when the user is translating.

### DIFF
--- a/docs/advanced_topics/audit_log.rst
+++ b/docs/advanced_topics/audit_log.rst
@@ -71,6 +71,7 @@ Action                               Notes
 ``wagtail.rename``                   A page was renamed
 ``wagtail.revert``                   The page was reverted to a previous draft
 ``wagtail.copy``                     The page was copied to a new location
+``wagtail.copy_for_translation``     The page was copied into a new locale for translation
 ``wagtail.move``                     The page was moved to a new location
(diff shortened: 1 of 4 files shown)


gasman (approved):

docs/advanced_topics/audit_log.rst
```diff
@@ -71,6 +71,7 @@ Action                               Notes
 ``wagtail.rename``                   A page was renamed
 ``wagtail.revert``                   The page was reverted to a previous draft
 ``wagtail.copy``                     The page was copied to a new location
+``wagtail.copy_for_translation``     The page was copied into a new locale for translation
```
gasman, line 74: Just a thought... any reason not to follow the hierarchical pattern of the other action names, and make it something like `wagtail.copy.for_translation`?

kaedroho: It seems a bit inconsistent the way it's done already. For example `wagtail.revert` is a type of edit, and ``wagtail.publish.schedule`` is not publishing the page, just scheduling it

gasman:
Merged in bea17ac8b6831cae936c6e97b5c24c9af24304e2.

PR #7361, opened 2021-07-22
Ensure error count badge counts StreamField non-block errors

### LINKED ISSUE #7353: Using non_block_errors in ListBlock
While testing 2.14rc1, I saw this in the release notes: "Allow ListBlock to raise validation errors that are not attached to an individual child block"  I thought I would try converting some code to using it. I am not 100% sure I am using it correctly, but I did notice one oddity. 

Page validation correctly notes (at the top of the edit form): The page could not be saved due to validation errors. And the error message I raised is nicely displayed at the top of the list block. But I do not have the usual circle with error count on the content tab. Is that a bug in the feature as released in 2.14? Or am I using it incorrectly? This is most logically something people will be doing with the upcoming min/max settings for ListBlock so feel free to close this with a note to wait for that feature if that is the preferred solution.

```python
class AirspaceStatBlock(blocks.StructBlock,):

    stat = blocks.CharBlock(
        help_text="The big statistic.",
        label="Statistic"
    )
    desc = blocks.CharBlock(
        label="Short Description",
        help_text=escape("The description underneath the big statistic. You may use these tags in this field: "
                         "<a>, <b>, <br>, <i>, <em>, <strong>, <sup>, <sub>")
    )

class AirspaceStatsBlock(blocks.StructBlock):
    STYLES = [
        ('h2', 'Heading 2'),
        ('h3', 'Heading 3'),
        ('h4', 'Heading 4'),
        ('h5', 'Heading 5'),
        ('h6', 'Heading 6 (all uppercase)'),
    ]

    title = blocks.CharBlock(
        required=False
    )
    title_style = blocks.ChoiceBlock(
        choices=STYLES,
        requried=True,
        default='h3'
    )
    title_centered = blocks.BooleanBlock(
        verbose_name='Center title',
        required=False,
        default=False,
        help_text='Check this box to center the title. Otherwise, it will be left-aligned.'
    )
    stats = blocks.ListBlock(
        AirspaceStatBlock(),
        label="Statistics"
    )

    def clean(self, value):
        result = []  # build up a list of (name, value) tuples to be passed to the StructValue constructor
        errors = {}
        for name, val in value.items():
            try:
                result.append((name, self.child_blocks[name].clean(val)))
            except Exception as e:
                errors[name] = ErrorList([e])

        if len(value['stats']) > 4:
            errors['stats'] = ErrorList(
                [ListBlockValidationError(
                    non_block_errors=ErrorList(['Please use no more than 4 Statistics per block.'])
                )]
            )

        if errors:
            raise StructBlockValidationError(errors)

        return self._to_struct_value(result)

```

Screen Shot 2021-07-20 at 2 14 49 PM

gasman:
Fixes #7353

The error counter was only counting elements with class error-message, but non-block errors on ListBlock / StreamBlock use a different styling, help-critical.

(For the record, I've checked that the uses of help-critical elsewhere in the codebase are also error messages, so this selector isn't being over-broad. Although hypothetically if there were any items that weren't actual error messages but warranted an "it's critically important that you read this advice" class, flagging them up via the badge is probably the right thing anyway...)

### DIFF
--- a/client/src/entrypoints/admin/page-editor.js
+++ b/client/src/entrypoints/admin/page-editor.js
@@ -222,7 +222,7 @@ function initErrorDetection() {
 
   // first count up all the errors
   // eslint-disable-next-line func-names
-  $('.error-message').each(function () {
+  $('.error-message,.help-critical').each(function () {
     const parentSection = $(this).closest('section');
 
     if (!errorSections[parentSection.attr('id')]) {


cnk:
Thanks this works for me.

Issue #7364, opened 2021-07-23
Make more panels collapsible


fabienheureux:
### Is your proposal related to a problem?

Currently, it seems that only `MultiFieldPanel` are collapsible. 
https://docs.wagtail.io/en/stable/reference/pages/panels.html#multifieldpanel

However, more panel types can support being collapsible, for example using `StreamFieldPanel("body", classname="collapsible multi-field")` to match [this selector ](https://github.com/wagtail/wagtail/blob/35132296cc45ea22a5f7db35df75685f2c3465be/client/src/entrypoints/admin/page-editor.js#L245)makes the panel collapsible. 

### Describe the solution you'd like

Updating the selector to match `.object.collapsible` should do the trick

### Describe alternatives you've considered

Update the documentation to suggest using the `multi-field` classname in the panel definition. However...this is more a hack than a proper way to achieve this.

gasman:
Thanks for the suggestion @fabienheureux! Would you be willing to submit a PR with this change?

fabienheureux:
Absolutely, I wanted to first check that it was not working this way on purpose.

zerolab:
#6187 discusses this in more detail and #6342 is the implementation which is very similar to #7365. @fabienheureux  can you check #6342 and bring in the home template changes and see double check the documentation changes?

fabienheureux:
Thanks for this comment @zerolab, sure I will update my PR shortly. 
I searched for open issues about this topic but did not think of searching in PR...

PR #7365, opened 2021-07-23
Fix/7364 make more panels collapsible

fabienheureux:
Fixes https://github.com/wagtail/wagtail/issues/7364

# Make more fields panels collapsible

The selector for collapsible panels now targets not only `MultiFieldPanel` but every panel with `classname="collapsible"`#7340 

Also in this PR: fields with `help_text` could not be collapsed, because only `fieldset` html tag was targetted by the JavaScript.
This is fixed by targeting not only `fieldset` but the whole `.object-layout` class (the field content). 

Below a screencast of the fix in action. 

https://user-images.githubusercontent.com/1702255/126796639-d7ea3c63-b692-4d5c-833b-63190c8998dc.mp4

### DIFF
--- a/docs/reference/pages/panels.rst
+++ b/docs/reference/pages/panels.rst
@@ -85,6 +85,11 @@ InlinePanel
 
     This is a powerful but complex feature which will take some space to cover, so we'll skip over it for now. For a full explanation on the usage of ``InlinePanel``, see :ref:`inline_panels`.
 
+.. topic:: Collapsing InlinePanels to save space
+
+    Note that you can use ``classname="collapsible collapsed"`` to load the panel collapsed under its heading in order to save space in the Wagtail admin.
+    See :ref:`my-reference-label` for more details on ``collapsible`` usage.
+
 
 
 FieldRowPanel
@@ -290,6 +295,8 @@ Titles
 Use ``classname="title"`` to make Page's built-in title field stand out with more vertical padding.
 
 
+.. _collapsible:
+
 Collapsible
 ~~~~~~~~~~~
 
@@ -297,9 +304,9 @@ By default, panels are expanded and not collapsible.
 Use ``classname="collapsible"`` to enable the collapse control.
 Use ``classname="collapsible collapsed"`` will load the editor page with the panel collapsed under its heading.
 
-
 You must define a ``heading`` when using ``collapsible`` with ``MultiFieldPanel``.
 You must defined a ``heading`` or ``label`` when using ``collapsible`` with ``InlinePanel``.
+
 
 .. code-block:: python
 
--- a/docs/reference/pages/panels.rst
+++ b/docs/reference/pages/panels.rst
@@ -291,7 +291,7 @@ Use ``classname="title"`` to make Page's built-in title field stand out with mor
 Collapsible
 ~~~~~~~~~~~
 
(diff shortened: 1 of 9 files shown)


fabienheureux:
Following #7364 comment, I have updated this PR to include #6342 changes @zerolab 
I kept the `Collapsible` section in the docs but added a note regarding `Collapsing InlinePanels to save space` section from #6342.

Let me know if this requires some updates

fabienheureux:
@allcaps any chance you could take a look at this PR before 2.15 release ? Thanks !

allcaps:
This PR request is automatically assigned to me. I'm short on time and this isn't my expertise. So, I requested a review from @Scotchester 🙏

wagtail/admin/templates/wagtailadmin/home.html
```diff
@@ -34,25 +34,5 @@ <h1>{% block branding_welcome %}{% blocktrans %}Welcome to the {{ site_name }} W
...
-                $target.addClass('collapsed');
-                $content.hide('slow');
-            } else {
-                $target.removeClass('collapsed');
-                $content.show('show');
-            }
-        });
-    });
-});
-</script>
```
Scotchester, line 57: I don't think we want to take this out of this template and load the entire `page-editor.js`. There's a lot of JS in that file that is totally useless on the Wagtail Admin homepage.

Scotchester (commented):
It looks great at a glance of the code, but I haven't had a chance to test it in a running environment yet. Here's one comment for your consideration in the meanwhile.

wagtail/admin/templates/wagtailadmin/home.html
fabienheureux: I extracted this from another PR but it makes totally sense, sorry for this change. 
Should we continue copy-paste this snippet with a comment, with a warning in both places in order to keep these in sync ?
Or do you think it would be better to extract this in its own js file ?

gasman: I'd be strongly in favour of moving this into its own `collapsible.js` file, and importing that both here and in wagtailadmin/pages/_editor_js.html. (I'm pretty sure that every view that currently depends on page-editor.js just includes the whole of _editor_js.html - a bad habit in the codebase that I'm trying to eliminate... - so that should preserve backwards compatibility.)

fabienheureux: Thanks for your feedback, I will update following these shortly

fabienheureux: I have updated accordingly, let me know if this need more work

docs/reference/pages/panels.rst
```diff
@@ -305,6 +295,34 @@ Titles
...
+
+Collapsible
+~~~~~~~~~~~
+
+By default, panels are expanded and not collapsible.
+Use ``classname="collapsible"`` to enable the collapse control.
+Use ``classname="collapsible collapsed"`` will load the editor page with the panel collapsed under its heading.
+
+You must define a ``heading`` when using ``collapsible`` with ``MultiFieldPanel``.
+You must defined a ``heading`` or ``label`` when using ``collapsible`` with ``InlinePanel``.
```
zerolab, line 308:
```suggestion
You must define a ``heading`` or ``label`` when using ``collapsible`` with ``InlinePanel``.
```

Scotchester:
@fabienheureux I'm working on rebasing this and getting it merged. There's a semi-significant conflict in `page-editor.js` to resolve.

Scotchester:
OK, finally all cleaned up! 😄 I'll squash-merge after the checks pass.

PR #7368, opened 2021-07-23
Update title of documentation page

jsma:
The title of this page mentioned videos because in the original version of this document it _did_ cover adding embedded videos. However, this page no longer mentions videos aside from the title.

### DIFF
--- a/docs/editor_manual/new_pages/inserting_images.rst
+++ b/docs/editor_manual/new_pages/inserting_images.rst
@@ -1,7 +1,7 @@
 .. _inserting_images:
 
 Inserting images in a page
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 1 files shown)


RealOrangeOne (approved):

docs/editor_manual/new_pages/inserting_images.rst
```diff
@@ -1,6 +1,6 @@
 .. _inserting_images:
 
-Inserting images and videos in a page
+Inserting images in a page
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
gasman, line 4: Please can you reduce the length of the underline to match the text? RestructuredText is picky about this...

gasman (changes requested):

docs/editor_manual/new_pages/inserting_images.rst
jsma: Done!

PR #7372, opened 2021-07-26
Allow subclasses of PagesAPIViewSet override default Page model.

hpoul:
I wanted to create an API endpoint with custom filtering for a specific page type. I thought the easiest would be to subclass `PagesAPIViewSet` - the only problem is that it would always use `Page` object even when overriding `get_base_queryset` or setting the `model` attribute.

imho it is also sub-optimal that when no `type` GET parameter is given, it still dynamically looks up `wagtailcore.Page` - imho this created a completely unused code fragment, because i don't see how `page_models_from_string(request.GET.get('type', 'wagtailcore.Page'))` could possibly return an False/empty value in the current implementation.

### DIFF
--- a/wagtail/api/v2/views.py
+++ b/wagtail/api/v2/views.py
@@ -441,7 +441,7 @@ class PagesAPIViewSet(BaseAPIViewSet):
         parent pages when using the child_of and descendant_of filters as well.
         """
         # Get live pages that are not in a private section
-        queryset = Page.objects.all().public().live()
+        queryset = self.model.objects.all().public().live()
 
         # Filter by site
         site = Site.find_for_request(self.request)
@@ -465,7 +465,8 @@ class PagesAPIViewSet(BaseAPIViewSet):
 
         # Allow pages to be filtered to a specific type
         try:
-            models = page_models_from_string(request.GET.get('type', 'wagtailcore.Page'))
+            models_type = request.GET.get('type', None)
+            models = models_type and page_models_from_string(models_type) or []
         except (LookupError, ValueError):
             raise BadRequestError("type doesn't exist")
 


wagtail/api/v2/views.py
```diff
@@ -441,7 +441,7 @@ def get_base_queryset(self):
         parent pages when using the child_of and descendant_of filters as well.
         """
         # Get live pages that are not in a private section
-        queryset = Page.objects.all().public().live()
+        queryset = self.model.objects.all().public().live()
```
kaedroho, line 444: This method is used to fetch children, ancestors, etc of the page as well, would filtering this prevent those related pages from showing up if they are not the same type?

Instead, I think we could change the below like that you've added to: ``models = models_type and page_models_from_string(models_type) or [self.model]``

kaedroho (changes requested):
Thanks! Please could you add some tests? You could probably copy our existing page tests, point it a new test endpoint, and remove any tests that aren't relevant

lb-:
If you are still keen - go for it @Jordan-Rob

NXPY123:
I'll look into this

NXPY123:
Can't the following code:
```python
elif len(models) == 1:
            # If a single page type has been specified, swap out the Page-based queryset for one based on
            # the specific page model so that we can filter on any custom APIFields defined on that model
            return models[0].objects.filter(
                id__in=self.get_base_queryset().values_list("id", flat=True)
            )
```
be replaced with:
```python
elif len(models)==1:
            return self.get_base_queryset().type(model)
```

Shouldn't the results be the same?

NXPY123:
@kaedroho are you talking about using the tests in the TestPageListing class in wagtail/wagtail/api/v2/tests/test_pages.py to test for a model type?

lb-:
Replaced by https://github.com/wagtail/wagtail/pull/10902

PR #7375, opened 2021-07-27
Ensure admin tab events are handled on page load

thatguystone:
The `shown.bs.tab` event was being bound after showTab() was already called, so on page load, the tabs weren't being updated correctly.

### DIFF
--- a/client/src/entrypoints/admin/core.js
+++ b/client/src/entrypoints/admin/core.js
@@ -295,6 +295,16 @@ $(() => {
     $(this).closest('li').removeClass('focused');
   });
 
+  /* Functions that need to run/rerun when active tabs are changed */
+  $(document).on('shown.bs.tab', () => {
+    // Resize autosize textareas
+    // eslint-disable-next-line func-names
+    $('textarea[data-autosize-on]').each(function () {
+      // eslint-disable-next-line no-undef
+      autosize.update($(this).get());
+    });
+  });
+
   /* tabs */
   const showTab = (tabButtonElem) => {
     $(tabButtonElem).tab('show');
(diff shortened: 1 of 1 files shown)


gasman:
Hi @thatguystone - thanks for the PR. Can you clarify what the specific bug being fixed here is, please?

As far as I can see, the event handler only exists to ensure that text areas that were previously initialised with the jquery.autosize plugin are refreshed when we switch to their tab (because the plugin isn't able to determine the correct height when those text areas are hidden inside a non-default tab). If so, it doesn't matter that the event handler isn't triggered on the initial page load, because any text areas inside the initially-shown tab will get correctly initialised and resized [when they are first rendered and have `autosize` applied to them](https://github.com/wagtail/wagtail/blob/b349ba436eeb853a8c2b2ff749f173a1f7d26d59/wagtail/admin/templates/wagtailadmin/widgets/auto_height_text_input.html).

If the lack of `shown.bs.tab` event on page load was a problem, I'd expect that problem to show up as text areas in the "Content" tab of the page editor failing to be resized correctly (e.g. [the 'introduction' field of a blog page on bakerydemo](https://github.com/wagtail/bakerydemo/blob/d3e8caedd728b7770b1bef917a7cb4f157ad20c7/bakerydemo/blog/models.py#L59)). Those work fine for me, though - maybe I'm missing something?

thatguystone:
Sorry about that, I should have described this better.

Try this: Edit Page > Promote > Meta Description: throw something really long in there so the box expands, and save it, then immediately hit back. The Promote tab should be focused, but the textarea won't be expanded.

lb-:
@thatguystone thanks for this, I have been able to reproduce the bug with the steps you provided.
This has been merged via 7d9e90fdbdf7db9657274d8f37d14eba29ffb7f2

PR #7377, opened 2021-07-27
Add attribution for Font Awesome v5 icons

gasman:
As per https://fontawesome.com/license/free:

> Attribution is required by MIT, SIL OFL, and CC BY licenses. Downloaded Font Awesome Free files already contain embedded comments with sufficient attribution, so you shouldn't need to do anything additional when using these files normally.

It looks like these comments are only present in the .svg files in the full Font Awesome package and not when downloading them individually from the Font Awesome site, which probably explains why we're missing them.

For Font Awesome v3 and v4, attribution was not required "but much appreciated" - which is just as well, because icomoon didn't preserve metadata about where they came from, and it doesn't seem to be the case that they all came from Font Awesome (e.g. the lock icon doesn't seem to match anything in any version of FA). With that in mind, I've gone through all icons added after 30 Jan 2018 (the first public release of FA5) and added the attribution where I've found a definite match in Font Awesome 5.

The ones where I didn't find a match are:

superscript, subscript, strikethrough - similar but not identical to ones that existed in Font Awesome 3
draft, resubmit - look like edited versions of existing icons from pre v5


kaedroho (approved):

kaedroho:
Looks good to me, thanks!

PR #7379, opened 2021-07-29
Mass cleanup of snippet chooser modal code

gasman:
Started out with the aim of phasing out ajax_pagination_nav.html - ended up going down quite a bit more of a rabbithole :-)

@kaedroho I've assigned you to review, as it encroaches on the locale filter code.

### DIFF
--- a/wagtail/admin/templatetags/wagtailadmin_tags.py
+++ b/wagtail/admin/templatetags/wagtailadmin_tags.py
@@ -15,6 +15,7 @@ from django.template.defaultfilters import stringfilter
 from django.template.loader import render_to_string
 from django.templatetags.static import static
 from django.urls import reverse
+from django.urls.exceptions import NoReverseMatch
 from django.utils import timezone
 from django.utils.encoding import force_str
 from django.utils.html import avoid_wrapping, format_html, format_html_join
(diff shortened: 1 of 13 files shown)


wagtail/snippets/urls.py
```diff
@@ -7,7 +7,6 @@
 urlpatterns = [
     path('', snippets.index, name='index'),
 
-    path('choose/', chooser.choose, name='choose_generic'),
```
gasman, line 10: To clarify, choose_generic was never a usable URL route in its own right - it only served to generate the partial URL to pass to JS (where the model parameter would then get tacked on).

wagtail/admin/templatetags/wagtailadmin_tags.py
```diff
@@ -716,3 +717,19 @@ def menu_props(context):
 @register.simple_tag
 def get_comments_enabled():
     return getattr(settings, 'WAGTAILADMIN_COMMENTS_ENABLED', True)
+
+
+@register.simple_tag
+def resolve_url_if_not_absolute(url):
```
kaedroho, line 723: Django has a similar built-in function https://github.com/django/django/blob/4b6208ffdd93d5d379d247ec094d567ef208effd/django/shortcuts.py#L102

Have you considered using this?

kaedroho (approved):

wagtail/admin/templatetags/wagtailadmin_tags.py
gasman: Ah, nice! Will update to use that...

PR #7381, opened 2021-07-29
Simplify generic view templates

gasman:
Some little enhancements to make generic views nicer to work with (e.g. adding a base class to make `page_title` and `page_subtitle` accessors consistently available, rather than them being implemented ad-hoc on each view) and reduce duplication.

(a bit of gardening in preparation for some deeper changes to the generic index/listing view, to follow)


lb-:
@gasman ran this locally and tested a bunch of scenarios, I did find an issue on create pages where there is a button in the DOM that is not showing correctly, I think this button is not meant to show at all as it is the link to the 'add' which will be the page the user is already on.

First is the current (main), and second image is after this change is applied.
Screen Shot 2021-08-23 at 8 55 11 pm

Examples with DOM detail
Screen Shot 2021-08-23 at 8 58 06 pm
Screen Shot 2021-08-23 at 8 58 28 pm

wagtail/admin/views/generic/models.py
```diff
@@ -58,6 +55,11 @@ def get_error_message(self):
             return None
         return self.error_message
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context['action_url'] = self.get_add_url()
+        return context
+
```
lb-, line 62: I am guessing the issue is this line, however I did not want to just remove it, I assume context is needed but not sure why the add url was added to the create view.

```
context['action_url'] = self.get_add_url()
```

gasman: @lb- Well caught! `action_url` here is used as the POST target for the 'add' form, but wagtailadmin/shared/header.html is picking it up to be used as the action button. `action_url` is really too generic a name for that, and we should probably get in the habit of using `{% include ... only %}` when including the header to stop unrelated variables from leaking into it - at least on these generic views.

gasman: Addressed in 1fe36d2597dfae558c979150651feff0f949ca03.

lb-: Nice, I will take another look

lb-:
Added in via 5d2794b9321b0e5453cae13ba3b6f851b78a30c6 & parents.
I had to add two more `only` usages for the workflow creation and user group creation, all others look good.

PR #7382, opened 2021-07-30
Bulk actions for users

KIRA009:
This PR adds bulk actions for the Users listing

image


wagtail/users/views/bulk_actions/toggle_activity.py
```diff
@@ -0,0 +1,80 @@
...
+
+    @classmethod
+    def execute_action(cls, objects, **kwargs):
+        cls.mark_as_active = kwargs.get('mark_as_active', False)
+        user = kwargs.get('user', None)
+        if user is not None:
+            objects = list(filter(lambda x: x.pk != user.pk, objects))
+        for user in objects:
+            user.is_active = cls.mark_as_active
+            user.save()
```
jacobtoppm, line 63: We should consider using `bulk_update` for these - let's bring up in relation to logging at the core team meeting as we discussed

wagtail/users/views/bulk_actions/toggle_activity.py
```diff
@@ -0,0 +1,80 @@
...
+                    context['users'] = list(filter(lambda x: x['user'].pk != user.pk, context['users']))
+                    context['mark_self_as_inactive'] = [user]
+                    return TemplateResponse(self.request, self.template_name, context)
+
+    @classmethod
+    def execute_action(cls, objects, **kwargs):
+        cls.mark_as_active = kwargs.get('mark_as_active', False)
+        user = kwargs.get('user', None)
+        if user is not None:
+            objects = list(filter(lambda x: x.pk != user.pk, objects))
```
jacobtoppm, line 60: I feel like permissions checks should already have taken place at this point: this is the part that could be run from a management command, so should just assume the user can perform the action

wagtail/users/views/bulk_actions/toggle_activity.py
```diff
@@ -0,0 +1,80 @@
...
+from django.template.response import TemplateResponse
+from django.utils.translation import gettext_lazy as _
+from django.utils.translation import ngettext
+
+from wagtail.core import hooks
+from wagtail.core.compat import AUTH_USER_APP_LABEL, AUTH_USER_MODEL_NAME
+from wagtail.users.views.bulk_actions.user_bulk_action import UserBulkAction
+
+
+change_user_perm = "{0}.change_{1}".format(AUTH_USER_APP_LABEL, AUTH_USER_MODEL_NAME.lower())
```
jacobtoppm, line 11: Why not import from https://github.com/wagtail/wagtail/blob/main/wagtail/users/views/users.py#L29?

wagtail/users/views/bulk_actions/assign_role.py
```diff
@@ -0,0 +1,73 @@
...
+from django.template.response import TemplateResponse
+from django.utils.translation import gettext_lazy as _
+from django.utils.translation import ngettext
+
+from wagtail.core import hooks
+from wagtail.core.compat import AUTH_USER_APP_LABEL, AUTH_USER_MODEL_NAME
+from wagtail.users.views.bulk_actions.user_bulk_action import UserBulkAction
+
+
+change_user_perm = "{0}.change_{1}".format(AUTH_USER_APP_LABEL, AUTH_USER_MODEL_NAME.lower())
```
jacobtoppm, line 12: You could import this from https://github.com/wagtail/wagtail/blob/main/wagtail/users/views/users.py#L29

jacobtoppm (changes requested):
Generally looking great! A few things to clear up however

wagtail/users/views/bulk_actions/toggle_activity.py
KIRA009: It is checked in the `prepare_action` method, but the list of users that is passed on to `execute_action` is not modified. So, we have to remove the user from the list of users again

wagtail/users/views/bulk_actions/assign_role.py
```diff
@@ -44,7 +44,7 @@ def prepare_action(self, objects):
 
     @classmethod
     def execute_action(cls, objects, **kwargs):
-        cls.role = role = kwargs.get('role', None)
+        role = kwargs.get('role', None)
```
jacobtoppm, line 47: I feel like it's clearer to have these explicitly spelled out in the kwargs if they're a key part of the action

wagtail/users/views/bulk_actions/toggle_activity.py
```diff
@@ -49,17 +49,17 @@ def prepare_action(self, objects):
...
             objects = list(filter(lambda x: x.pk != user.pk, objects))
         for user in objects:
-            user.is_active = cls.mark_as_active
+            user.is_active = mark_as_active
             user.save()
             cls.num_parent_objects += 1
 
     def get_success_message(self):
-        activity_status = "active" if self.mark_as_active else "inactive"
+        activity_status = "active" if self.cleaned_form.cleaned_data['mark_as_active'] else "inactive"
```
jacobtoppm, line 58: This won't be translated, will it? I think you'll need to write `get_success_message` in a way where both the `activity_status` and the rest of the string will be translated, optimally together

jacobtoppm (changes requested):
Very close now! A couple of minor bugfixes

jacobtoppm (approved):

zerolab:
Thank you @KIRA009 and @jacobtoppm

PR #7383, opened 2021-07-30
Make multigetattr from SpreadsheetExportMixin into a reusable utility

gasman:
and steal Django's implementation for template variable resolution so that it's more robust and supports dict/list lookups too.

Planning to re-use this functionality for a new version of the generic index / listing view, so it makes sense to have a definitive implementation of it in core.utils.

### DIFF
--- a/wagtail/admin/views/mixins.py
+++ b/wagtail/admin/views/mixins.py
@@ -10,6 +10,7 @@ from django.utils.translation import gettext as _
 from xlsxwriter.workbook import Workbook
 
 from wagtail.admin.forms.search import SearchForm
+from wagtail.core.utils import multigetattr
 from wagtail.search.backends import get_search_backend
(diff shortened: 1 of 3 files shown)


jacobtoppm (approved):
Looks great, thanks for expanding this! I think it's fine as is, but tiny nitpick: similar to Django here (https://github.com/django/django/blob/f331eba6d576752dd79c4b37c41d981daa537fe6/django/template/base.py#L847) and our no-param-reassign rule in js, I do think it would be slightly clearer to read if we had some kind of current value variable instead of just using `item` (since if a function takes a param and then returns it I personally tend to assume it's probably doing some kind of pass by reference modification or something... maybe just me though!)

gasman:
Thanks @jacobtoppm! Merged in ae14c8c718909c6610f3397c6c06607759b0b72e with your suggested readability tweak applied.

Issue #3672, opened 2017-06-21
API :: Private pages not listed when authenticated


spoetnik:
### Issue Summary

I have set some pages, with childpages, to private. Ofter logging in on the website, these pages are shown. When browsing the /pages endpoint in the API, those private pages are not listed. The DRF interface shows my logged in username, so I am authenticated.

### Steps to Reproduce
1) Create a group called 'Private'
2) Create a user called 'Private', which is member of the 'Private' group.
3) Create a page, add subpages if you feel like it, and set the privacy to 'Private, accessible to users in specific groups', and select the group 'Private'. Save
4) In the Live-view, go to the created page with the Privacy settings. You are redirected to a login page.
5) Login with the user 'Private'
6) The page is available to you, and can be seen.
7) Go to the /pages API endpoint, /api/v1/pages/
8) Find the page with the Privacy settings. Bang head to the wall...
9) Notice in the upper right corner you are authenticated as user 'Private'

I would expect the 'Private' page to be available in the public API after authentication. Only Public pages are shown.

### Technical details

* Python version: Python 3.6.1
* Django version: Django==1.11.2
* Wagtail version: wagtail==1.10.1

gasman:
Thanks for the report @spoetnik! I can confirm that the API doesn't currently support authentication to access private pages - it just filters to public/live pages:

https://github.com/wagtail/wagtail/blob/deeee4a5978be50eb6f12132ca1f00d8cdc19f05/wagtail/api/v2/endpoints.py#L402-L403

jonathanfmachado:
Hi @spoetnik. Did you manage to find a way around this issue?

> I would expect the 'Private' page to be available in the public API after authentication. Only Public pages are shown.

peteyb:
Hi @jonathanfmachado 

Not sure if this is of help to you but a relatively simple solution is to subclass the PagesAPIEndpoint class and override the get_queryset() method to check if the user is authenticated.

It does not take into account the various private options, such as being part of a group, but this may suffice for your purposes.

```
from wagtail.api.v2.endpoints import PagesAPIEndpoint
from wagtail.api.v2.utils import (BadRequestError, filter_page_type,
                                  page_models_from_string)
from wagtail.core.models import Page


class MyPagesAPIEndpoint(PagesAPIEndpoint):
    """
    Override stock wagtail pages api to handle private pages
    """
    def get_queryset(self):
        request = self.request

        # Allow pages to be filtered to a specific type
        try:
[...]

PR #7385, opened 2021-07-31
Fix/3672 private pages api

NabilMostafa:
Fix issue #[3672](https://github.com/wagtail/wagtail/issues/3672)
# Changes 

- [x]  If user is not authenticated, return public pages only
- [x]  If user is authenticated, return pages that has login restriction. Also, check if there are any group restriction and check if user belongs to any of the groups in the page restriction groups
- [x] Cover changes in tests


wagtail/api/v2/views.py
```diff
@@ -440,8 +441,16 @@ def get_base_queryset(self):
...
-        queryset = Page.objects.all().public().live()
+        # Get all live pages
+        queryset = Page.objects.all().live()
+        # if user is not authenticated return only public pages
+        if not self.request.user.is_authenticated:
+            queryset &= queryset.public()
+        else:  # get pages that are accessible to logged-in users or users in specific groups
+            queryset &= queryset.filter(Q(view_restrictions__restriction_type=PageViewRestriction.LOGIN)
+                                        | Q(view_restrictions__groups__in=self.request.user.groups.all())
+                                        | Q(view_restrictions__restriction_type=None))
```
RealOrangeOne, line 452: Indentation here looks a little off
```suggestion
            queryset &= queryset.filter(
                Q(view_restrictions__restriction_type=PageViewRestriction.LOGIN) |
                Q(view_restrictions__groups__in=self.request.user.groups.all()) |
                Q(view_restrictions__restriction_type=None)
            )
```

NabilMostafa: You're right, but your suggestion broke this rule https://www.flake8rules.com/rules/W504.html, I think it should be like that ->
```suggestion
          queryset &= queryset.filter(
                Q(view_restrictions__restriction_type=PageViewRestriction.LOGIN)
                | Q(view_restrictions__groups__in=self.request.user.groups.all())
                | Q(view_restrictions__restriction_type=None)
            )
```

wagtail/api/v2/tests/test_pages.py
```diff
@@ -81,6 +84,48 @@ def test_private_pages_dont_appear_in_list(self):
         content = json.loads(response.content.decode('UTF-8'))
         self.assertEqual(content['meta']['total_count'], new_total_count)
 
+    def test_private_pages_with_user_login(self):
+        client = APIClient()
```
kaedroho, line 88: Any reason to not use ``self.client`` here?

kaedroho (changes requested):

wagtail/api/v2/tests/test_pages.py
NabilMostafa: For some reason when I used `self.client` and `self.login` the user wasn't logged in the views and always returned `anonymous` user type

lb-:
@kaedroho  just a ping to see if you can re-review this, looks like changes requested have been addressed

kaedroho (approved):

PR #7390, opened 2021-08-03
Fix broken references in Pages topic documentation

chosak:
The "Page models" documentation page in the Wagtail usage guide has several broken links (see current page at
https://docs.wagtail.io/en/stable/topics/pages.html).

These links work when developing the documentation locally, but not when served on ReadTheDocs (docs.wagtail.io)

The broken links are all of the form `/some/page.html#anchor` - somehow this format doesn't translate properly to get the necessary RTD prefix (for example `/en/stable/`).

I've modified these links to use RST references, which will also make them more robust to future changes.

Edit: Verify that links are fixed on this page at the PR preview at https://wagtail--7390.org.readthedocs.build/en/7390/topics/pages.html.

### DIFF
--- a/docs/reference/pages/model_reference.rst
+++ b/docs/reference/pages/model_reference.rst
@@ -170,6 +170,8 @@ In addition to the model fields provided, ``Page`` has many properties and metho
 
     .. automethod:: get_url
 
+    .. automethod:: get_full_url
+
     .. autoattribute:: full_url
 
     .. automethod:: relative_url
--- a/docs/topics/pages.md
+++ b/docs/topics/pages.md
@@ -1,6 +1,8 @@
 # Page models
 
-Each page type (a.k.a. content type) in Wagtail is represented by a Django model. All page models must inherit from the [`wagtail.core.models.Page`](/reference/pages/model_reference.html#wagtail.core.models.Page) class.
+```eval_rst
+Each page type (a.k.a. content type) in Wagtail is represented by a Django model. All page models must inherit from the :class:`wagtail.core.models.Page` class.
+```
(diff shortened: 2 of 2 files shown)


gasman:
One of the biggest motivators for switching to Markdown was the less horrible link syntax, so if we're going to drop back into RST for links then that largely defeats the object of using Markdown at all. Maybe we should use relative links instead? i.e.

    [`wagtail.core.models.Page`](../reference/pages/model_reference.html#wagtail.core.models.Page)

cc @tbrlpld

chosak:
@gasman I recall some discussion about RST vs Markdown during the documentation sprint earlier this year. The current [Documentation guidelines](https://docs.wagtail.io/en/stable/contributing/documentation_guidelines.html) say

>  We encourage writing documentation in Markdown first, and only reaching for more advanced reStructuredText formatting if there is a compelling reason.

Personally I think being able to write

``` :class:`wagtail.core.models.Page` ```

is a "compelling reason" over writing

```[`wagtail.core.models.Page`](../reference/pages/model_reference.html#wagtail.core.models.Page)```

since the latter requires duplication and an understanding of the relationship between two pages that is much more brittle when the linked page inevitably changes or moves. But I am happy to change this PR to use Markdown if that is the preference.

The main issue is that absolute references like `/reference/pages/model_reference.html` will unfortunately seem to work fine in one's development environment but then only cause a problem once deployed to ReadTheDocs.

gasman:
OK, fair point... I see that this page is more "API documentation"-oriented than most, so using the rst-specific syntax for class/method links is justified here.

PR #7391, opened 2021-08-03
Add a debug logger around rendition generation

RealOrangeOne:
This allows insight into which images are taking the longest to generate, which fail to render at all, and potentially which images are causing crashes (as in they start, but never stop).

The logging is intentionally only on DEBUG level, so it's opt-in, and is also reasonably quiet so it doesn't bloat logs.

Relevant Slack discussion: https://wagtailcms.slack.com/archives/C023WG6FB7C/p1627922996003600.

### DIFF
--- a/wagtail/images/models.py
+++ b/wagtail/images/models.py
@@ -1,5 +1,7 @@
 import hashlib
+import logging
 import os.path
+import time
 
 from collections import OrderedDict
 from contextlib import contextmanager
@@ -27,6 +29,9 @@ from wagtail.images.image_operations import FilterOperation, ImageTransform, Tra
 from wagtail.images.rect import Rect
 from wagtail.search import index
 from wagtail.search.queryset import SearchableQuerySetMixin
+
+
+LOGGER = logging.getLogger("wagtail.images")
 
 
 class SourceImageIOError(IOError):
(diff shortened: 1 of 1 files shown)


wagtail/images/models.py
```diff
@@ -303,7 +308,23 @@ def get_rendition(self, filter):
             )
         except Rendition.DoesNotExist:
             # Generate the rendition image
-            generated_image = filter.run(self, BytesIO())
+            try:
+                LOGGER.debug("Generating '%s' rendition for image %d", (
```
RealOrangeOne, line 312: I'm not entirely sure if this log before is necessary.

My main intention is to cover the case where generating an image uses so much memory that the application either locks up or is OOM killed (or platform-specific kill). I suspect an exception handler probably wouldn't catch that.

Question is, is it important enough to have 2 log messages for each generation as opposed to just 1?

wagtail/images/models.py
```diff
@@ -29,6 +31,9 @@
 from wagtail.search.queryset import SearchableQuerySetMixin
 
 
+LOGGER = logging.getLogger("wagtail.images")
```
RealOrangeOne, line 34: Question: Should we add some explicit documentation about the existence of this logger, and how to enable / use it?

gasman:
Looks good to me... Merged in 90c86e76d0d6f46208eb1efb2918026d61ee3326.

jgeskens:
Hi, this patch seems to break the generation of image renditions. This logger statement uses incorrect syntax for the format arguments (a tuple instead of *args), triggering an exception: https://github.com/wagtail/wagtail/commit/90c86e76d0d6f46208eb1efb2918026d61ee3326#diff-c195bc8bd4d5a11ae02d0451edb4dd4a8e7b0014398a384a1be3ff3c355a5ee8R312
Before this change, everything still worked fine. So for now I have to downgrade.

lb-:
@jgeskens thanks for the comment, are you able to raise a bug for this, it will help us action this sooner, comments against closed PRs are easy to get lost.

RealOrangeOne:
Nice catch - i'm not sure how I missed that, or how CI never caught it. I've opened a fix as #7694

lb-:
All good - fix PR is already up https://github.com/wagtail/wagtail/pull/7694

PR #7394, opened 2021-08-03
Fix bug that caused disconnect between ui and javascipt state in bulk action confirmation cancel

KIRA009:
When using the navigate back button of the browser to cancel a bulk action from the confirmation page, the user landed on the previous page with the objects still checked, but the javascript state would be empty. This pr fixes this issue

## Before
https://user-images.githubusercontent.com/40872556/128392403-ca6d7f3e-0d3e-4300-916a-84a87fd9dac1.mp4

## After
https://user-images.githubusercontent.com/40872556/128392463-bc229769-384b-4824-a6ff-444b6e852ca8.mp4

### DIFF
--- a/client/src/entrypoints/admin/bulk-actions.js
+++ b/client/src/entrypoints/admin/bulk-actions.js
@@ -168,6 +168,7 @@ function onClickActionButton(e) {
  * Adds all event listeners
  */
 function addBulkActionListeners() {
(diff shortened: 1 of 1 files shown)


Stormheg (approved):
Nice one Shohan!

This has fixed the issue, tested on Chrome 92 and Safari 14.1 on macOS 11.5.

PR #7395, opened 2021-08-03
Namespace data attributes for bulk actions

KIRA009:
All `data-` attributes required for bulk actions are namespaced with `data-bulk-action-` as discussed [here](https://github.com/wagtail/wagtail/pull/7340#discussion_r677222405)

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/pages/listing/_button_with_dropdown.html
+++ b/wagtail/admin/templates/wagtailadmin/pages/listing/_button_with_dropdown.html
@@ -10,7 +10,7 @@
         <ul class="c-dropdown__menu u-toggle  u-arrow u-arrow--tl u-background">
         {% for button in buttons %}
             <li class="c-dropdown__item ">
-                <a href="{{ button.url }}" {% if button.attrs.filter %}data-filter={{button.attrs.filter}}{% endif %} aria-label="{{ button.attrs.title }}" class="u-link is-live {{ button.classes|join:' ' }}">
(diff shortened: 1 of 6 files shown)


wagtail/admin/templates/wagtailadmin/pages/listing/_button_with_dropdown.html
```diff
@@ -10,7 +10,7 @@
         <ul class="c-dropdown__menu u-toggle  u-arrow u-arrow--tl u-background">
         {% for button in buttons %}
             <li class="c-dropdown__item ">
-                <a href="{{ button.url }}" {% if button.attrs.filter %}data-filter={{button.attrs.filter}}{% endif %} aria-label="{{ button.attrs.title }}" class="u-link is-live {{ button.classes|join:' ' }}">
+                <a href="{{ button.url }}" {% if button.attrs.filter %}data-bulk-action-filter={{button.attrs.filter}}{% endif %} aria-label="{{ button.attrs.title }}" class="u-link is-live {{ button.classes|join:' ' }}">
```
zerolab, line 13: This is the only place I am not 100% sure, now that I see the changes in a small PR.
button with dropdown could be used outside of the bulk actions context.

suggestion: make it accept an attrs.data dict with key for the data attribute and value

```python
attrs = {
    "data": {"bulk-action-filter": "something"}
}
```
and here `{% if button.attrs.data %}{% for data_key, data_value in button.attrs.data.items %} ...`

zerolab (changes requested):
Excellent changes, bar one on the button with dropdown that needs reworking.
The case there is.. data-filter or data-bulk-actions-filter are very specific, but we should make it generic enough for other uses too

zerolab (approved):
LGTM

PR #7397, opened 2021-08-04
Template components

gasman:
Implements a standard documented pattern for "objects that know how to render themselves onto templates". This pattern has been reimplemented numerous times throughout Wagtail's development, for things like homepage panels and action buttons, in various levels of completeness - and as a result, some of them provide ready-made mechanisms for rendering templates, some provide a get_context method, some allow specifying external css / js, some of them can pick up context variables from the calling template, and so on... Standardising these will make these features consistently available, and make our APIs easier to learn (no more "is it `render` or `render_html`?")

Things that use components as of this PR:
* Page action menu items (register_page_action_menu_item)
* Snippet action menu items (register_snippet_action_menu_item)
* Homepage panels (construct_homepage_panels)
* Summary items (construct_homepage_summary_items)

Other things that could potentially be migrated to components in future:
* Sidebar menu items (need to consider how this aligns with the telepath / slim sidebar rewrite, but I believe these concepts can coexist)
* the page / snippet action menus as a whole (PageActionMenu, SnippetActionMenu as distinct from ActionMenuItem)
* Userbar / edit bird menu items
* User account settings panels
* Search areas (wagtail.admin.search)
* wagtail.admin.widgets.button (as used for buttons / button menus within listings)

### DIFF
--- a/docs/extending/template_components.rst
+++ b/docs/extending/template_components.rst
@@ -1,0 +1,165 @@
+Template components
+===================
+
+Working with objects that know how to render themselves as elements on an HTML template is a common pattern seen throughout the Wagtail admin. For example, the admin homepage is a view provided by the central ``wagtail.admin`` app, but brings together information panels sourced from various other modules of Wagtail, such as images and documents (potentially along with others provided by third-party packages). These panels are passed to the homepage via the :ref:`construct_homepage_panels` hook, and each one is responsible for providing its own HTML rendering. In this way, the module providing the panel has full control over how it appears on the homepage.
+
+Wagtail implements this pattern using a standard object type known as a **component**. A component is a Python object that provides the following methods and properties:
+
+.. method:: render_html(self, request, parent_context)
+
+Given a request object and a dict of context variables from the calling template, returns the string representation to be inserted into the template. This will be subject to Django's HTML escaping rules, so a return value consisting of HTML should typically be returned as a :py:mod:`SafeString <django.utils.safestring>` instance.
+
+.. attribute:: media
+
+A (possibly empty) :doc:`form media <django:topics/forms/media>` object defining JavaScript and CSS resources used by the component.
+
+.. note::
+   Any object implementing this API can be considered a valid component; it does not necessarily have to inherit from the ``Component`` class described below, and user code that works with components should not assume this (for example, it must not use ``instanceof`` to check whether a given value is a component).
+
+
+Creating components
+-------------------
+
+The simplest way to create a component is to define a subclass of ``wagtail.admin.components.Component`` and specify a ``template`` attribute on it. The rendered template will then be used as the component's HTML representation:
+
+.. code-block:: python
+
+    from wagtail.admin.components import Component
+
+    class WelcomePanel(Component):
+        template = 'my_app/panels/welcome.html'
+
+
+    my_welcome_panel = WelcomePanel()
+
+``my_app/templates/my_app/panels/welcome.html``:
+
+.. code-block:: html+django
+
+    <h1>Welcome to my app!</h1>
+
+
+For simple cases that don't require a template, the ``render_html`` method can be overridden instead:
+
+.. code-block:: python
+
+    from django.utils.html import format_html
+    from wagtail.admin.components import Component
+
+    class WelcomePanel(Component):
+        def render_html(self, request, parent_context):
+            return format_html("<h1>{}</h1>", "Welcome to my app!")
+
+
+Passing context to the template
+-------------------------------
+
+The ``get_context`` method can be overridden to pass additional context variables to the template. As with ``render_html``, this receives the request object and the context dictionary from the calling template:
+
+.. code-block:: python
+
+    from wagtail.admin.components import Component
+
+    class WelcomePanel(Component):
+        template = 'my_app/panels/welcome.html'
+
+        def get_context(self, request, parent_context):
+            context = super().get_context(request, parent_context)
+            context['username'] = request.user.username
+            return context
+
+``my_app/templates/my_app/panels/welcome.html``:
+
+.. code-block:: html+django
+
+    <h1>Welcome to my app, {{ username }}!</h1>
+
+
+Adding media definitions
+------------------------
+
+Like Django form widgets, components can specify associated JavaScript and CSS resources using either an inner ``Media`` class or a dynamic ``media`` property:
+
+.. code-block:: python
+
+    class WelcomePanel(Component):
+        template = 'my_app/panels/welcome.html'
+
+        class Media:
+            css = {
+                'all': ('my_app/css/welcome-panel.css',)
+            }
+
+
+Using components on your own templates
+--------------------------------------
+
+The ``wagtailadmin_tags`` tag library provides a ``{% component %}`` tag for including components on a template. This takes care of passing context variables from the calling template to the component (which would not be the case for a basic ``{{ ... }}`` variable tag). For example, given the view:
+
+
+.. code-block:: python
+
+    from django.shortcuts import render
+
+    def welcome_page(request):
+        panels = [
+            WelcomePanel(),
+        ]
+
+        render(request, 'my_app/welcome.html', {
+            'panels': panels,
+        })
+
+the ``my_app/welcome.html`` template could render the panels as follows:
+
+.. code-block:: html+django
+
+    {% load wagtailadmin_tags %}
+    {% for panel in panels %}
+        {% component panel %}
+    {% endfor %}
+
+
+Note that it is your template's responsibility to output any media declarations defined on the components. For a Wagtail admin view, this is best done by constructing a media object for the whole page within the view, passing this to the template, and outputting it via the base template's ``extra_js`` and ``extra_css`` blocks:
+
+.. code-block:: python
+
+    from django.forms import Media
+    from django.shortcuts import render
+
+    def welcome_page(request):
+        panels = [
+            WelcomePanel(),
+        ]
+
+        media = Media()
+        for panel in panels:
+            media += panel.media
+
+        render(request, 'my_app/welcome.html', {
+            'panels': panels,
+            'media': media,
+        })
+
+``my_app/welcome.html``:
+
+.. code-block:: html+django
+
+    {% extends "wagtailadmin/base.html" %}
+    {% load wagtailadmin_tags %}
+
+    {% block extra_js %}
+        {{ block.super }}
+        {{ media.js }}
+    {% endblock %}
+
+    {% block extra_css %}
+        {{ block.super }}
+        {{ media.css }}
+    {% endblock %}
+
+    {% block content %}
+        {% for panel in panels %}
+            {% component panel %}
+        {% endfor %}
+    {% endblock %}
--- a/docs/extending/template_components.rst
+++ b/docs/extending/template_components.rst
@@ -17,6 +17,8 @@ A (possibly empty) :doc:`form media <django:topics/forms/media>` object defining
    Any object implementing this API can be considered a valid component; it does not necessarily have to inherit from the ``Component`` class described below, and user code that works with components should not assume this (for example, it must not use ``instanceof`` to check whether a given value is a component).
 
 
+.. _creating_template_components:
+
 Creating components
 -------------------
 
--- a/docs/extending/template_components.rst
+++ b/docs/extending/template_components.rst
@@ -14,7 +14,7 @@ Given a request object and a dict of context variables from the calling template
 A (possibly empty) :doc:`form media <django:topics/forms/media>` object defining JavaScript and CSS resources used by the component.
 
 .. note::
-   Any object implementing this API can be considered a valid component; it does not necessarily have to inherit from the ``Component`` class described below, and user code that works with components should not assume this (for example, it must not use ``instanceof`` to check whether a given value is a component).
+   Any object implementing this API can be considered a valid component; it does not necessarily have to inherit from the ``Component`` class described below, and user code that works with components should not assume this (for example, it must not use ``isinstance`` to check whether a given value is a component).
 
 
 .. _creating_template_components:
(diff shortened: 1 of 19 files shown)


gasman:
Rendered docs: https://wagtail--7397.org.readthedocs.build/en/7397/extending/template_components.html

ababic:
Possibly off-topic, and no way easy way of backtracking now, but I wanted to ask the reasoning for deviating away from the Django tropes of specifying templates with `template_name` and using `get_context_data(self, *args, **kwargs)` for putting together the context (with the 'just call super() with extra kwargs to add more items)?

I've been working with Django and Wagtail for a long time, and the former is still something that regularly trips me up (using the wrong convention in the wrong place), and the latter is a pattern I often long for when creating custom components like streamfied blocks. 

To a lesser extent, this confusion also extends to developers of third-party apps in the Wagtail/Django space, with devs having to second-guess which conventions their intended audience might be more familiar with - or maybe just using the ones they are more familiar while another might be more appropriate.

In short, I'm not saying I 'prefer' the Django's naming conventions, but reducing the number of ways in which Django and Wagtail differ are can only be a good thing for developers - it's the 'kinder' thing to do.

gasman:
@ababic All good points - I'm keen to get these details right, as it's going to be a hassle to change them later. Besides `template` / `template_name` and `get_context` / `get_context_name`, I was also hung up over whether the render method should be `render` or `render_html` (I also briefly toyed with `as_html`).

Up to this point I've been mostly taking a "paving the cowpaths" stance and formalising what Wagtail does already, to minimise the disruption in making this change. But that might not be the wisest approach:

* "what Wagtail does already" is disproportionately driven by my own personal biases
* As Wagtail shifts towards being a development platform with its own conventions (rather than just an app with one-off extension points), the differences with Django conventions are going to become more conspicuous
* The people currently using these hooks are probably more on the 'hardcore' end of the Wagtail developer scale, and more accustomed to breaking changes than most...

Changing `template` to `template_name`, I think I can get on board with. It's slightly fiddly to do the deprecation, but probably no more so than the `render` -> `render_html` change in SummaryItem - I think we just need to get render_html to check for the presence of `template` and give it precedence over `template_name`, for now, while outputting a deprecation warning.

get_context is trickier. It's worth noting that Django CBVs are in a different position here, as they don't have to deal with the phenomenon of picking up variables from the calling template's context. I'm still not sure whether having the base `get_context` method [copy those into the context as standard](https://github.com/wagtail/wagtail/blob/5e920327330ccf58e5117e3fada75208e2b3cd11/wagtail/admin/components.py#L12) is the right move - I think there's a risk of variables from the outer template polluting the context, if the component's template does something like `<div {% if classname %}class="{{ classname }}"{% endif %}>` and get_context doesn't make a point of explicitly setting classname to something meaningful or None. (But if we don't do that, we'll have to be more explicit about what happens to `request` - I think having that available as standard is a good thing.) At the very least, having the parent context variables passed in an explicitly-identified dict so that the component code knows where they came from, and can deal with them accordingly, _feels_ like the pragmatic move... once you've made **kwargs into a free-for-all and surrendered all control over what might be passed there, it's hard to put that genie back into the bottle.

Having said that, I'm very conscious of the need to make the get_context API as programmer-friendly as possible. Already in my first day of writing component code, I've done the following several times:

    def get_context(self, request, parent_context):
        context = super().get_context(request, parent_context)
        context['foo'] = self.foo

and forgotten the `return context` at the end - so I'm definitely open to considering better idioms here.

If we _do_ reach the conclusion that components are in a different situation to CBVs here, then I think there's a case to be made that giving it a different name (`get_context` rather than `get_context_data`) is justified, so that people don't conflate the two.

Although, another minor point there: in certain places within Django, such as inside template tags, [a Context object](https://docs.djangoproject.com/en/3.2/ref/templates/api/#playing-with-context-objects) is a stack-like data structure rather than the plain dict as expected by `render` / `render_to_string` etc. It may be that Django has adopted the convention of referring to the dict as "context data" to distinguish it from "a context object", and if so, we should probably think twice about treading on that convention.

docs/extending/template_components.rst
```diff
@@ -0,0 +1,167 @@
+Template components
```
tbrlpld, line 1: **Question** Just wondering why this new page for the docs is started in RST? As I understand we would like to migrate all documentation from RST to MD (with a few pages already started). It might be a helpful nudge in that direct to start all new documentation pages in MD directly.

tbrlpld:
Is there a particular reason that the components are stored in the admin app? It seems like the component concept could be helpful for developing Wagtail "frontend" (public facing) pages with that approach as well.

docs/extending/template_components.rst
gasman: Mostly just inertia :-) In this case, there are a few function definitions and Intersphinx cross-links to the Django docs - looking up how to do those in MD (or whether I should revert to RST for those) would have been one extra thing to think about. Happy to switch if people think that would take things in a positive direction though!

gasman:
@tbrlpld If it does find more widespread use - and I hope it does - then it would be just as applicable to Django in general, and at that point it would make sense to release as a standalone package. Right now, though, I see the Wagtail admin as being a sort of sandbox for testing out and refining the idea.

ababic:
> It may be that Django has adopted the convention of referring to the dict as "context data" to distinguish it from "a context object", and if so, we should probably think twice about treading on that convention.

Yes, I observed recently when writing some template tags that, when using `@register.simple_tag(takes_context=True)`, the `context` object received is actually a `django.template.context.RequestContext` instance, which would be the case for components rendered from the `{% component %}` tag too. If the component is then starting another rendering process, where the result of `get_context_data()` is used to create a new `Context` or `RequestContext` instance, I think 'get_context_data()' as a name makes sense.

It's a similar thing with `template_name`. The value isn't an actual `Template` instance until something initiates `django.template.Engine.get_template(name)` to take the string value and return one (something usually done by a `get_template()` method).

ababic:
@gasman How would you feel about this?

```
def get_context_data(self, parent_context: Context, **kwargs):
    value = {}    # whatever I want to add here
    value.update(kwargs)
    request = context.get('request')
    if request:
        # do something request-specific
    else:
        # the component was potentially rendered outside a request/response cycle
    return value
```

ababic:
> I'm still not sure whether having the base get_context method copy those into the context as standard is the right move - I think there's a risk of variables from the outer template polluting the context, if the component's template does something like <div {% if classname %}class="{{ classname }}"{% endif %}> and get_context doesn't make a point of explicitly setting classname to something meaningful or None

I suspect that selectively keeping some things in `context` and not others might play with expectations a bit. Maybe having the ability to affect rendering of a component in the above way could be useful in some cases? If the component absolutely cannot have certain values 'trickle down' from the parent context, it can always reset anything important with its own defaults (blank/None).

ababic:
> I was also hung up over whether the render method should be render or render_html (I also briefly toyed with as_html).

I guess my question would be: Are components always rendering HTML? 

If they are, then maybe the `_html` portion of the method name is redundant?

If not, then does `render_html` set too many expectations? can a component not render something else entirely (e.g. JS/CSS/XML)?

gasman:
> @gasman How would you feel about this?

That code seems a lot more 'heavyweight' than I'd like - or is that because it's a "demonstrating everything" example rather than a realistic one? Ideally the developer experience when writing this method should be more about "stuffing things into a dict to make them available on the template" and less about unpacking parameters and passing them up and down an inheritance tree.

The fact that you need a `context.update(kwargs)` line to support subclassing is a red flag for me (and yes, I'm not a fan of Django CBVs doing similar things) - it makes way more sense to me for subclassing to work as "call `super()`, then work with the result".

I need to do some more thinking about whether it's a good thing or not for `get_context_data` to receive a Context object rather than a dict. I get the feeling that Context objects give you more low-level control (which is why template tags expose them), but if developers are going to have to routinely call `push()` or `flatten()` to do anything useful to it, it might be a bit too esoteric. (Again, I'd like developers to be able to approach this at a level of "just stuffing variables into a dict" and not have to think about anything deeper than that :-) )

Maybe a good guideline to follow is: components should work as closely as possible to an `inclusion_tag` with `takes_context=True`. Which implies:
* yes, receiving a Context object is fine
* variables don't just get automatically inherited from the parent context (not even `request`) - you have to explicitly include them in the dict you return from `get_context_data`
* the Components mechanism itself doesn't dictate which variables are available to the component (again, not even `request`) - that's an informal contract between the component and the template it's being used in, just like it would be for one template `{% include %}`-ing another

gasman:
> I guess my question would be: Are components always rendering HTML?

As far as the components API is concerned: yes. Rendering HTML is the specific problem components are designed to solve - anything else would be feature creep.

But - 'component' is just an interface that a given object can choose to implement, or not. It would be perfectly reasonable to have an object that implements the component API and _also_ provides `as_json`, a Telepath adapter, `export_as_knitting_pattern` or whatever. :-) So, by that logic, `render_html` does make sense.

gasman:
@ababic Thanks for the clarification! I'm not sold on the idea that components should introduce a new calling convention at the point that the developer decides they're "designed to be subclassed" - it seems like that brings us back to the situation where developers have to keep track of two different conventions, except instead of "is this `template` or `template_name`", it'll be "is this a pass-variables-to-super component or a pass-variables-to-context-update component".

In a way this circles back to the original motivation for components, where we've got the same pattern implemented in various places across Wagtail, all in different stages of evolution from less flexible to more flexible because in certain places the need for the more flexible version hasn't come up yet - so we're providing a ready-made fully-realised version of the pattern to make the same capabilities consistently available everywhere. If there's an optional extra feature that can be enabled with a few lines of additional plumbing as and when the need arises, then only the well-trafficked bits of Wagtail will get that feature, and we'll be back to inconsistent capabilities for the lesser-used bits.

There's still room for specific components to build their own extensions on top of the components API - so a particular component could choose to implement **kwargs passing - but as far as the base Component class goes, I think it's important to preserve the principle that if a developer writes a component X in the normal idiomatic way without giving any special thought to inheritance, then another developer can subclass it into component Y and still have the same APIs available that the developer of component X had.

Thanks again for all the feedback - I think there's now enough refinements on the table for me to come up with a second draft...

ababic:
No problem at all @gasman. A good, healthy exchange of thought here :)

gasman:
New plan, borrowing ideas from `inclusion_tag`...

```
from django.forms import MediaDefiningClass
from django.template import Context
from django.template.loader import get_template


class Component(metaclass=MediaDefiningClass):
    def get_context_data(self, parent_context: Context):
        return {}

    def render_html(self, parent_context=None):
        if parent_context is None:
            parent_context = Context()
        context_data = self.get_context_data(parent_context)
        if context_data is None:
            raise TypeError("Expected a dict from get_context_data, got None")
        new_context = parent_context.new(context_data)

        template = get_template(self.template_name)
        return template.render(new_context)
```

Changes:
* names now follow Django's precedent from CBVs: `template_name` and `get_context_data`
* context variables from the calling template are no longer pulled through to the component template automatically - if you want a variable to be available, you have to explicitly set it in get_context_data
* `render_html` and `get_context_data` now receive `parent_context` as a Context object rather than a dict. I don't think there's much difference in terms of how it ultimately gets used, but if nothing else it saves the `{% component %}` template tag from having to do any `flatten`-ing or copying to get it into the right format, for something that may not even get used...
* `request` is no longer an argument of its own. I only really had it that way because `render_to_string` wanted it to be separate, and it looks like that's only so that it can decide whether to instantiate a Context or a RequestContext. The implications of that choice are fairly irrelevant here, and completely moot when we have a parent context to work with and can just do `parent_context.new({...})`, like inclusion_tag does. A component that wants access to the request object on the template can still do `{'request': parent_context['request']}` - assuming the caller actually passes it, which is a matter for the component and caller to agree amongst themselves.
* `parent_context` is now optional on `render_html` (but not `get_context_data`). While we'd expect it to be always present in normal use (e.g. when invoking via the `{% component %}` tag), I think it'll come in useful for debugging to be able to throw in a quick `print(foo.render_html())` or `{{ foo.render_html }}`. (That does rely on component authors writing them robustly enough that they can gracefully handle missing context vars, but it's not the end of the world if they don't.)
* Added a check for `context_data is None`, so that a missing `return context` in get_context_data fails in a more obvious way :-)

Miscellaneous TODOs:
* change docs to Markdown
* move the module to `wagtail.admin.ui.components`

ababic:
> Added a check for context_data is None, so that a missing return context in get_context_data fails in a more obvious way :-)

Very nice touch :)

gasman:
> `render_html` and `get_context_data` now receive parent_context as a Context object rather than a dict.

Going to change this slightly: parent_context can now be _either_ a Context object or a dict. If you treat it as a read-only dict lookup (which I expect 99% of components to do) it makes no difference, but components that do non-standard things with it (e.g. override `render_html` to bypass get_context_data and pass the whole context directly to a `render_to_string` call) will have to do some additional type-checking before trying to call `.flatten()` or `.copy()` or whatever on it. This is prompted by the following findings while rebuilding this according to the new plan:

* Context objects are not as ubiquitous to the whole template-rendering process as I thought. The `render_html` code I posted above is incorrect - the Template object returned by `get_template` expects its `render` method to receive a plain dict, so the `new_context = parent_context.new(context_data)` step is unwanted. ([The one that inclusion_tag works with](https://github.com/django/django/blob/3a6431db5461b88e691d770abdba6a2793ef699d/django/template/library.py#L226-L235) is apparently a different kind of Template object, and judging from the `context.template.engine.get_template` business, a more low-level / internal one that we'd do better to steer clear of.)
* There are cases when the input to be passed to `render_html` / `get_context_data` is naturally available as a dict - e.g. [during unit testing](https://github.com/wagtail/wagtail/blob/c6ffafcedf8ee51373c0a8d01a177e2670958454/wagtail/admin/tests/pages/test_dashboard.py#L89) or [when it's being built up from Python code that isn't (yet) a component itself](https://github.com/wagtail/wagtail/blob/5e920327330ccf58e5117e3fada75208e2b3cd11/wagtail/admin/action_menu.py#L323-L328) - and forcing it to be passed as a Context object is an unnecessary extra hurdle.

gasman:
Superseded by #7425

PR #7398, opened 2021-08-04
Add missing args/kwargs to Page documentation

chosak:
Documentation examples of `Page.get_context` and `Page.get_template` lack `*args` and `**kwargs` parameters (which were added way back in 8c4c26864103890897ca63708867621f16938a25).

This commit adds those missing parameters.

See live documentation page [here](https://docs.wagtail.io/en/stable/topics/pages.html#customising-template-context) and [here](https://docs.wagtail.io/en/stable/topics/pages.html#dynamically-choosing-the-template). Edit: and compare to the documentation page for this PR [here](https://wagtail--7398.org.readthedocs.build/en/7398/topics/pages.html#customising-template-context) and [here](https://wagtail--7398.org.readthedocs.build/en/7398/topics/pages.html#dynamically-choosing-the-template).

### DIFF
--- a/docs/topics/pages.md
+++ b/docs/topics/pages.md
@@ -252,8 +252,8 @@ To add more variables to the template context, you can override this method:
 class BlogIndexPage(Page):
     ...
 
-    def get_context(self, request):
-        context = super().get_context(request)
+    def get_context(self, request, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


gasman:
Thanks @chosak! I don't _think_ these would be needed unless the developer overrides the `route` method, but it certainly can't hurt...

PR #7403, opened 2021-08-05
Bulk actions - action bar UI fixes

KIRA009:
This pr brings the action bar in bulk actions closer to the actual design

## Design
image

## Implementation
image

### DIFF
--- a/client/scss/components/_listing.scss
+++ b/client/scss/components/_listing.scss
@@ -500,10 +500,14 @@ ul.listing {
     }
 
     .button {
-        font-weight: 700;
+        font-weight: 600;
     }
 
     .bulk-actions-more {
+        .button {
+            border: 0;
+        }
+
         .button:not(:hover) {
             color: $color-teal;
         }
--- a/client/scss/components/_listing.scss
+++ b/client/scss/components/_listing.scss
@@ -530,8 +530,9 @@ ul.listing {
     justify-content: space-around;
     width: 100%;
     align-items: center;
-    padding: 1.5em 2em;
-    border-radius: 0.5em 0.5em 0 0;
+    padding: 1.25em;
+    border-radius: 4px 4px 0 0;
+    margin-left: 30px;
 
     .bulk-actions-buttons {
         border-left: 1px solid $color-grey-2;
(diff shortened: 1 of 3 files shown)


zerolab:
Hey @KIRA009,

this certainly needs more work:

1. left margin should ~80px to align with the titles in the listing for pages
2. `.footer_container` needs less padding (~1.25em all around. and `4px 4px 0 0` for border radius)
3. the buttons do not have borders in the designs. at the moment there's a slight hint of the border but it gets lost in the dark bar background

zerolab:
@KIRA009 this is still marked as draft. Is it ready for re-review?

KIRA009:
Yes @zerolab

Stormheg (commented):
This looks good @KIRA009!

The amount of whitespace is like the designs, well done 👍 

Here are some minor nitpicks I'd like to address before merging.

- Buttons labels have `font-weight: 700`. The design uses `font-weight: 600` here - meaning labels appear slightly too thick in the current implementation. _Can we reduce the font weight?_
- Observation: the design has the number of objects (e.g. 'All 11 pages') highlighted by making it bold (`font-weight: 700`). _Is there a reason this isn't implemented?_
- The 'more' button has a gray border - unlike the other buttons which have no border. _Can we remove the border?_
- Observation: the 'select all X pages in listing' link (only appears on paginated listings) doesn't mention the total number of objects. _Is there a reason the number of objects is missing?_

KIRA009:
@Stormheg 

> Observation: the design has the number of objects (e.g. 'All 11 pages') highlighted by making it bold (font-weight: 700). Is there a reason this isn't implemented?

The string is part of a whole sentence, so making some words bold would require putting those words in a <span> . We have to fetch the two phrases 2 pages and on this screen selected as separate strings, they’ll need to be translated separately as well. I figured it was too much complexity for a simple effect. That is the reason I omitted that part. Any thoughts @jacobtoppm @zerolab ?

> Observation: the 'select all X pages in listing' link (only appears on paginated listings) doesn't mention the total number of objects. Is there a reason the number of objects is missing?

I believe we decided to remove the number a while back, probably because of translation inconsistencies, and , as far as I remember

Stormheg (approved):
LGTM - thanks Shohan! 💯

PR #7407, opened 2021-08-05
Tiny cleanup for site root page chooser

gasman:
* Pulling in _editor_js.html is unnecessary - the only JS dependency that isn't in form media is now modal_workflow.js. (So close to being able to ditch the template override entirely!)
* Omit the 'clear' button, as this is a required field.


zerolab (approved):
\o/ for this kind of tidy ups!
Tested locally and LGTM

gasman:
Merged in 07402a545a1225160e26c57dc5dac1bff417a0d3

PR #7408, opened 2021-08-05
Makes the title of ModelAdmin items into links

### LINKED ISSUE #7333: ModelAdmin IndexView not keyboard navigable

### Issue Summary

The default ModelAdmin IndexView isn't keyboard navigable. Without using a mouse, there's no way to use the edit or delete (or, presumably, other added buttons of that view) of a particular item. This also affects screen reader users.

### Steps to Reproduce

1. Use ModelAdmin to register a menu item. 
2. Try to tab to an item's edit or delete buttons. 
3. Alternatively, in the bakery demo, navigate to anything under Bread Categories or Bread Misc, and then edit something via keyboard. 

At first glance, changing these to operate in the same manner as the page and snippets indexview works seems appropriate, where each item is a link that can be targetted, thus triggering the `.focus-within` of the `td` or `tr`.  

### Technical details

* Wagtail version: 2.13.3
* Browser version: Chrome, Safari, FirefoxESR (all Mac OS). I presume IE and Edge, as well.

SaptakS:
Fixes #7333 

As suggested by @thibaudcolas in the issue, I converted the title of the item into a link (structured it using the same HTML as page listing to maintain the same style). This allows for keyboard navigation to the action buttons for each item using the `:focus-within` like rest of admin.

      * Firefox, Chrome on Debian, safari on MacOS
      * VoiceOver on MacOS

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -12,6 +12,7 @@ Changelog
  * Fix: Ensure that StreamBlock / ListBlock-level validation errors are counted towards error counts (Matt Westcott)
  * Fix: InlinePanel add button is now keyboard navigatable (Jesse Menn)
  * Fix: Remove redundant 'clear' button from site root page chooser (Matt Westcott)
+ * Fix: Make ModelAdmin IndexView keyboard-navigable (Saptak Sengupta)
 
 
 2.14 (02.08.2021)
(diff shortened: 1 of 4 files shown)


thibaudcolas (commented):
Looking at the code only, this seems very solid. I’ll test this locally next.

thibaudcolas (approved):
Working great :)

Issue #7400, opened 2021-08-04
Sorting/filtering throughout the admin


nmorduch:
### Problem

It can be frustrating to try to find things like:
- All active users
- All users with admin permissions
- All active users with admin permissions
- Active users who haven't logged in in the past year
- All pages of Foo type that are children of a specific page
- In the history of a page, only the times it has been published, or only actions by a specific person

### Solution I'd like

- [ ] Every table of information in the admin to be sortable by every heading
- [ ] Ideally add filtering as well as sorting, for most headings
  - [ ] For things like page type and user, a list of checkboxes for the ones you want to filter by
  - [ ] For things like date, a range to filter for (ideally with an option for not having a date)
  - [ ] Should be able to filter by multiple things at a time

### Alternatives I've considered

- Just add filtering to a few specific places, such as viewing users
- Add a "show inactive users" button, equivalent to "show disabled tasks" on the workflow tasks index page

gasman:
Currently working on some improvements to the [generic index view](https://github.com/wagtail/wagtail/blob/main/wagtail/admin/views/generic/models.py#L14-L30) that should make it much easier to build views with properly-working sortable tables, and adding filtering shouldn't be a massive leap from there. Some preliminary work on this is already in flight on #7381 and #7397 - the rest to follow shortly...

nmorduch:
I'm excited to hear that! Thanks @gasman!

thibaudcolas:
I believe this is covered by #10446 :) @nmorduch if you see a need for more specific functionality not covered in that project, I think it’d now be better to have individual issues for specific sorting/filtering requirements (specific fields, specific views).

PR #7409, opened 2021-08-05
Generic framework for object listings

gasman:
Adds a `wagtail.admin.tables` module, based on template components (#7397), that lets you construct listing tables given a set of column definitions (which are easily subclassable to define new styles), and handles the rendering.

Hooks this up to wagtail.admin.views.generic.IndexView so that it's possible to build index views for arbitrary models, including pagination and sortable headings, with no template code or method overriding at all.

Incorporates #7381 and #7397
Partially addresses #5191 and #7400

(Marking as draft as it depends on the template components PR, which is still in flux...)


gasman:
By the way, I'm aware that this will probably collide with the bulk actions work, but I believe it will be straightforward to accommodate that when it lands - basically just adding the checkbox column as another standard column type in the definition.

wagtail/admin/tables.py
```diff
@@ -0,0 +1,190 @@
+"""Helper classes for formatting data as tables"""
```
kaedroho, line 1: Why not put this in ``wagtail/admin/ui/``?

gasman: Sounds good! Wasn't sure if admin/ui was intended for more front-end / react-y things, but if you think it's a good fit, let's go with that.

gasman: Sounds good! Wasn't sure if admin/ui was intended for more front-end / react-y things, but if you think it's a good fit, let's go with that.

gasman:
Rebuilt on top of new template components version (#7425).

gasman:
Rebased on top of current main (which includes the final template components PR #7425) and the fixed #7381.

wagtail/sites/views.py
```diff
@@ -1,5 +1,6 @@
-from django.utils.translation import gettext_lazy
+from django.utils.translation import gettext_lazy as _
```
kaedroho, line 1: I think it makes sense to call ``gettext_lazy`` with something different than ``gettext`` to make it less likely to use the wrong one. What do you think?

wagtail/users/views/groups.py
```diff
@@ -57,8 +58,12 @@ class IndexView(mixins.SearchableListMixin, generic.IndexView):
     search_fields = ['name']
     context_object_name = 'groups'
     paginate_by = 20
-    page_kwarg = 'p'
     ordering = ['name']
+    default_ordering = 'name'
+
+    columns = [
+        TitleColumn('name', label=_("Name"), sort_key='name', url_name='wagtailusers_groups:edit'),
```
kaedroho, line 65: Should be `gettext_lazy` here

wagtail/admin/ui/tables.py
```diff
@@ -0,0 +1,213 @@
...
+            'value': self.get_value(instance),
+            'request': parent_context.get('request'),
+        }
+
+    def render_cell_html(self, instance, parent_context):
+        """
+        Renders a table cell containing data for the given instance
+        """
+        context = self.get_cell_context_data(instance, parent_context)
+        return render_to_string(self.cell_template_name, context)
```
kaedroho, line 97: This is probably going to run extremely slowly since every cell will load, parse, and render a template (and Django templates are really slow as we've seen with Streamfield!)

wagtail/admin/views/generic/models.py
```diff
@@ -16,15 +17,65 @@ class IndexView(PermissionCheckedMixin, WagtailAdminTemplateMixin, BaseListView)
...
+    page_kwarg = 'p'
+    default_ordering = None
+
+    def get(self, request, *args, **kwargs):
+        if not hasattr(self, 'columns'):
+            self.columns = self.get_columns()
+
+        return super().get(request, *args, **kwargs)
+
+    def get_columns(self):
```
kaedroho, line 32: Would all index views be required to use this? Would it make sense to make this a mixin?

wagtail/admin/views/generic/models.py
```diff
@@ -16,15 +17,65 @@ class IndexView(PermissionCheckedMixin, WagtailAdminTemplateMixin, BaseListView)
...
+        try:
+            return self.columns
+        except AttributeError:
+            return [
```
kaedroho, line 36: Any reason to not set the default value of the ``columns`` attribute to this? I can imagine this being useful for overriding as we do with similar attributes on page:

```
class MyIndexView(IndexView):
    columns = IndexView.columns + [
        # ...
    ]
```

kaedroho (changes requested):

wagtail/admin/views/generic/models.py
gasman: The problem is that we can't reference `self.get_edit_url` from a class-level definition.

wagtail/admin/views/generic/models.py
gasman: I'd say it's as much a requirement as `get_form` is for an edit view... it's a core part of what the view is designed to do, and the expectation is that you will use it, but you have a decent amount of freedom to subclass and override it if you have non-standard requirements. If you don't want the provided table functionality, then I think you'd just swap out the template with one that doesn't use the `table` variable.

I think it's more likely that you'd just roll your own view class based on `WagtailAdminTemplateMixin` in that case though - there's nothing in viewsets or similar that requires you use IndexView, right?

wagtail/admin/ui/tables.py
gasman: Good point! We could certainly cache the compiled template on the Column object here.

wagtail/sites/views.py
gasman: The makemessages command only recognises `_` and the original function names, so we don't have a lot of freedom here. (I seem to remember us trying to use `_` and `__` in the distant past, and it not working at all...) From what I can see, we've adopted an informal convention of using `_` for whichever one is most useful within that file:

* `from django.utils.translation import gettext_lazy as _` has 94 matches across the codebase
* `from django.utils.translation import gettext as _` has 53 matches

and if the generic views are doing their job properly, there'll be a lot more strings in class-level configuration than inside methods. (Besides, is there any real penalty for using `gettext_lazy` when `gettext` would have worked...?)

zerolab: I could not find anything against using `gettext_lazy`. Do be doubly sure I tester the changes with a different admin language and it is all working as advertised/expected.

zerolab (approved):
Took this for a local test and re-checked the code. LGTM :shipit:

gasman:
Thanks @zerolab! Now merged in 4db9be469fba2810d837028bfaf94c3048173ba0.

PR #7416, opened 2021-08-08
Fix Uncaught ReferenceError when editing links in Hallo

cnk:
When editing links with the hallo rich text editor, I get the following error if the link has been saved (either as draft or by publishing):

```javascript
Uncaught ReferenceError: assignment to undeclared variable parentPageId
    populateToolbar hallo-wagtaillink.js:48
```

I see this on Wagtail 2.14 and with all 3 browsers I have tried (Safari, Chrome, and Firefox). This PR fixes that problem by declaring the parentPageId variable. 

(I also see this problem on 2.12 and 2.13 but because those are not receiving support and Hallo is deprecated, I don't see that we need to backport to those releases. But if there is going to be a 2.14.1 release, I would be grateful if this could be included there.)


kaedroho (approved):
Thanks!

PR #7418, opened 2021-08-09
Bulk actions - refactor execute_action

KIRA009:
This pr refactors several methods in `BulkAction` class so that `execute_action` can deal with the objects supplied to it without going through additional permission checks


jacobtoppm (approved):
Looks good to me!

KIRA009:
Can we merge this after other listings are merged, so if any refactors are required in those, I can add them in this pr itself

wagtail/admin/views/bulk_action.py
```diff
@@ -88,7 +88,9 @@ def get_actionable_objects(self):
                 objects_with_no_access.append(obj)
             else:
                 objects.append(obj)
-        return objects, objects_with_no_access
+        return objects, {
+            '{}s_with_no_access'.format(self.object_key): objects_with_no_access
```
zerolab, line 92: question: why not have this generic (e.g. `items_with_no_access`) cc @jacobtoppm

jacobtoppm: Yep, this replicated the existing template variables, but I agree a generic version would be better - it's not like we use the key name here except in code.

KIRA009: Should I remove the `self.object_key` altogether then?

jacobtoppm: It seems to only be used for changing dictionary key names, so I think so, yes. If you think through the process of extending a bulk action, having this just adds extra complexity - because you need to think about what the object name would be in your action's case - without any big benefit (maybe it's slightly more intuitive to write `pages_with_no_access` on a template, but when you're wondering why your bulk action template for your custom snippet type x doesn't work even though you just copied your other template for a different model, I think it winds up more on the "extra confusion" side).

Issue #4201, opened 2018-01-16
WAGTAIL_APPEND_SLASH is not applied to /_util/* URLs


coredumperror:
If `WAGTAIL_APPEND_SLASH = False` is set in the django settings, I would expect all Wagtail-served URLs to obey it (as [the docs](http://docs.wagtail.io/en/latest/advanced_topics/settings.html#append-slash) say they should). However, neither the `/admin/*` URLs nor (as I discovered today) the `/_util/*` URLs obey this setting. 

I recognize that making every single admin URL obey it might be a pain in the butt, but there are only two `/_util` URLs (`_util/authenticate_with_password/(\d+)/(\d+)/` and `_util/login/`. _And_ they're defined in the same file where `WAGTAIL_APPEND_SLASH` is used to implement its feature (`watailcore/urls.py`).

Heck, it might be as simple as making the ending slash be always optional. Like this:

```
urlpatterns = [
    url(r'^_util/authenticate_with_password/(\d+)/(\d+)/?$', views.authenticate_with_password,
        name='wagtailcore_authenticate_with_password'),
    url(r'^_util/login/?$', auth_views.login, {'template_name': WAGTAIL_FRONTEND_LOGIN_TEMPLATE},
        name='wagtailcore_login'),

    # Front-end page views are handled through Wagtail's core.views.serve
    # mechanism
    url(serve_pattern, views.serve, name='wagtail_serve')
]
```

Or, if you want to limit it with `WAGTAIL_APPEND_SLASH`, you could do this:
```
slash = '/?' if not WAGTAIL_APPEND_SLASH else '/'
urlpatterns = [
    url(r'^_util/authenticate_with_password/(\d+)/(\d+){}$'.format(slash), 
        views.authenticate_with_password,
        name='wagtailcore_authenticate_with_password'),
    url(r'^_util/login{}$'.format(slash), auth_views.login, 
        {'template_name': WAGTAIL_FRONTEND_LOGIN_TEMPLATE},
        name='wagtailcore_login'),

    # Front-end page views are handled through Wagtail's core.views.serve
    # mechanism
    url(serve_pattern, views.serve, name='wagtail_serve')
]
```

Is there any downside to doing that? I understand why there would be a downside to doing this on Page URLs (SEO issues), but these are internal utility URLs. I'm not sure what problems this could possibly cause.

gasman:
Hi @coredumperror,

[...]

PR #7421, opened 2021-08-10
Make WagtailAPIRouter obey WAGTAIL_APPEND_SLASH.

coredumperror:
WagtailAPIRouter assumes that all API urls have to end in `/`, which should not be true when `WAGTAIL_APPEND_SLASH=False`. This PR fixes that assumption.

The tests I added are not the best... but I couldn't come up with a better way to confirm that the end-`/` is not necessary in the URL when `WAGTAIL_APPEND_SLASH=False`. I'm sure there's a better way to do this.

### DIFF
--- a/wagtail/api/v2/views.py
+++ b/wagtail/api/v2/views.py
@@ -13,6 +13,7 @@ from rest_framework.viewsets import GenericViewSet
 
 from wagtail.api import APIField
 from wagtail.core.models import Page, Site
+from wagtail.core.utils import WAGTAIL_APPEND_SLASH
 
 from .filters import (
     AncestorOfFilter, ChildOfFilter, DescendantOfFilter, FieldsFilter, LocaleFilter, OrderingFilter,
@@ -336,10 +337,16 @@ class BaseAPIViewSet(GenericViewSet):
         """
         This returns a list of URL patterns for the endpoint
         """
+        if WAGTAIL_APPEND_SLASH:
+            detail_path = '<int:pk>/'
+            find_path = 'find/'
+        else:
+            detail_path = '/<int:pk>'
+            find_path = '/find'
         return [
             path('', cls.as_view({'get': 'listing_view'}), name='listing'),
-            path('<int:pk>/', cls.as_view({'get': 'detail_view'}), name='detail'),
-            path('find/', cls.as_view({'get': 'find_view'}), name='find'),
+            path(detail_path, cls.as_view({'get': 'detail_view'}), name='detail'),
+            path(find_path, cls.as_view({'get': 'find_view'}), name='find'),
         ]
 
     @classmethod
--- a/wagtail/api/v2/router.py
+++ b/wagtail/api/v2/router.py
@@ -2,6 +2,7 @@ import functools
 
 from django.urls import include, re_path
 
+from wagtai.core.utils import WAGTAIL_APPEND_SLASH
 from wagtail.utils.urlpatterns import decorate_urlpatterns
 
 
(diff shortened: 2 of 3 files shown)


coredumperror:
Huh, odd. Not sure why the tests I added are failing. They pass just fine when I run the test suite on my dev box.

coredumperror:
Turns out that I missed the detail_view and find_view paths. But with those "fixed", I'm really not sure I'm doing this properly at all. My solution for those two URLs just feels wrong. But I'm not sure how to articulate why.

wagtail/api/v2/views.py
```diff
@@ -336,10 +337,16 @@ def get_urlpatterns(cls):
         """
         This returns a list of URL patterns for the endpoint
         """
+        if WAGTAIL_APPEND_SLASH:
+            detail_path = '<int:pk>/'
+            find_path = 'find/'
+        else:
+            detail_path = '/<int:pk>'
+            find_path = '/find'
```
allcaps, line 345: The Django's `APPEND_SLASH=True` setting is not intended to force all urls to end with a slash. The CommonMiddleware docs explain it well:

> If APPEND_SLASH is True and the initial URL doesn’t end with a slash, and it is not found in the URLconf, then a new URL is formed by appending a slash at the end. If this new URL is found in the URLconf, then Django redirects the request to this new URL. Otherwise, the initial URL is processed as usual.
> For example, foo.com/bar will be redirected to foo.com/bar/ if you don’t have a valid URL pattern for foo.com/bar but do have a valid pattern for foo.com/bar/.
> https://docs.djangoproject.com/en/3.2/ref/middleware/#django.middleware.common.CommonMiddleware

`WAGTAIL_APPEND_SLASH` should behave similar to Django’s `APPEND_SLASH`. That is: if a url pattern without trailing slash exists and it matches the requested url, just serve it.

coredumperror: I guess the code doesn't make it clear, but that should already be true. Requesting the API views that I created in my Wagtail project with or without an end-slash both work. That _may_ be due to a custom middleware that we use, but I'm not sure.

gasman:
I'll echo what @allcaps said here, and also my reply on #4201...

Django's APPEND_SLASH setting is not intended to override the choice of trailing-slash-or-not as made by the developer(s) who wrote the urlconfig - all it does is perform a redirect in the specific case that a user enters a slash-less URL that doesn't exist, but the corresponding URL with trailing slash does.

Wagtail as a product defines certain built-in URL patterns, including the admin, _util and API. For these, we've standardised on trailing slashes. While we _could_ make that configurable according to the end-user's preferences, that would add a maintenance burden for very limited benefit, and - crucially - the existence of APPEND_SLASH doesn't set a precedent that a configuration setting for that is expected or desirable.

In the specific case of Wagtail page URL paths, we recognise that a site owner is defining their own URL scheme through the act of creating pages, and it makes sense that they should be able to set their own convention about trailing slashes, just as they would for a vanilla Django site where they're writing their own urlconf - so we provide WAGTAIL_APPEND_SLASH as a concession to that. Again, the fact that a setting exists for that one purpose doesn't set a precedent that trailing slashes in general should be configurable.

Finally, if we were to make the trailing slash configurable for API URLs, doing it through WAGTAIL_APPEND_SLASH would introduce a breaking change - on a site with the API enabled and WAGTAIL_APPEND_SLASH = False, the URL `/api/v2/pages/81/` (with trailing slash) currently works, and this PR makes it a 404. It's likely that sites exist in the wild with WAGTAIL_APPEND_SLASH = False and API clients that _do_ currently set the trailing slash when making API requests.

PR #7425, opened 2021-08-10
Template components (take 2)

gasman:
Rebuild of #7397 with feedback applied as summarised in https://github.com/wagtail/wagtail/pull/7397#issuecomment-894325899.

Implements a standard documented pattern for "objects that know how to render themselves onto templates". This pattern has been reimplemented numerous times throughout Wagtail's development, for things like homepage panels and action buttons, in various levels of completeness - and as a result, some of them provide ready-made mechanisms for rendering templates, some provide a get_context method, some allow specifying external css / js, some of them can pick up context variables from the calling template, and so on... Standardising these will make these features consistently available, and make our APIs easier to learn (no more "is it `render` or `render_html`?")

Things that use components as of this PR:

* Page action menu items (register_page_action_menu_item)
* Snippet action menu items (register_snippet_action_menu_item)
* Homepage panels (construct_homepage_panels)
* Summary items (construct_homepage_summary_items)

Other things that could potentially be migrated to components in future:

* Sidebar menu items (need to consider how this aligns with the telepath / slim sidebar rewrite, but I believe these concepts can coexist)
* the page / snippet action menus as a whole (PageActionMenu, SnippetActionMenu as distinct from ActionMenuItem)
* Userbar / edit bird menu items
* User account settings panels
* Search areas (wagtail.admin.search)
* wagtail.admin.widgets.button (as used for buttons / button menus within listings)

### DIFF
--- a/docs/extending/template_components.md
+++ b/docs/extending/template_components.md
@@ -1,0 +1,159 @@
+# Template components
+
+Working with objects that know how to render themselves as elements on an HTML template is a common pattern seen throughout the Wagtail admin. For example, the admin homepage is a view provided by the central `wagtail.admin` app, but brings together information panels sourced from various other modules of Wagtail, such as images and documents (potentially along with others provided by third-party packages). These panels are passed to the homepage via the [`construct_homepage_panels`](../reference/hooks.html#construct-homepage-panels) hook, and each one is responsible for providing its own HTML rendering. In this way, the module providing the panel has full control over how it appears on the homepage.
+
+Wagtail implements this pattern using a standard object type known as a **component**. A component is a Python object that provides the following methods and properties:
+
+```eval_rst
+.. method:: render_html(self, parent_context=None)
+
+Given a context dictionary from the calling template (which may be a :py:class:`Context <django.template.Context>` object or a plain ``dict`` of context variables), returns the string representation to be inserted into the template. This will be subject to Django's HTML escaping rules, so a return value consisting of HTML should typically be returned as a :py:mod:`SafeString <django.utils.safestring>` instance.
+
+.. attribute:: media
+
+A (possibly empty) :doc:`form media <django:topics/forms/media>` object defining JavaScript and CSS resources used by the component.
+
+.. note::
+   Any object implementing this API can be considered a valid component; it does not necessarily have to inherit from the ``Component`` class described below, and user code that works with components should not assume this (for example, it must not use ``isinstance`` to check whether a given value is a component).
+```
+
+
+## Creating components
+
+The simplest way to create a component is to define a subclass of `wagtail.admin.ui.components.Component` and specify a `template_name` attribute on it. The rendered template will then be used as the component's HTML representation:
+
+```python
+from wagtail.admin.ui.components import Component
+
+class WelcomePanel(Component):
+    template_name = 'my_app/panels/welcome.html'
+
+
+my_welcome_panel = WelcomePanel()
+```
+
+`my_app/templates/my_app/panels/welcome.html`:
+
+```html+django
+<h1>Welcome to my app!</h1>
+```
+
+For simple cases that don't require a template, the `render_html` method can be overridden instead:
+
+```python
+from django.utils.html import format_html
+from wagtail.admin.components import Component
+
+class WelcomePanel(Component):
+    def render_html(self, parent_context):
+        return format_html("<h1>{}</h1>", "Welcome to my app!")
+```
+
+
+## Passing context to the template
+
+The `get_context_data` method can be overridden to pass context variables to the template. As with `render_html`, this receives the context dictionary from the calling template:
+
+```python
+from wagtail.admin.ui.components import Component
+
+class WelcomePanel(Component):
+    template_name = 'my_app/panels/welcome.html'
+
+    def get_context_data(self, parent_context):
+        context = super().get_context_data(parent_context)
+        context['username'] = parent_context['request'].user.username
+        return context
+```
+
+`my_app/templates/my_app/panels/welcome.html`:
+
+```html+django
+<h1>Welcome to my app, {{ username }}!</h1>
+```
+
+
+## Adding media definitions
+
+Like Django form widgets, components can specify associated JavaScript and CSS resources using either an inner `Media` class or a dynamic `media` property:
+
+```python
+class WelcomePanel(Component):
+    template_name = 'my_app/panels/welcome.html'
+
+    class Media:
+        css = {
+            'all': ('my_app/css/welcome-panel.css',)
+        }
+```
+
+
+## Using components on your own templates
+
+The `wagtailadmin_tags` tag library provides a `{% component %}` tag for including components on a template. This takes care of passing context variables from the calling template to the component (which would not be the case for a basic `{{ ... }}` variable tag). For example, given the view:
+
+```python
+from django.shortcuts import render
+
+def welcome_page(request):
+    panels = [
+        WelcomePanel(),
+    ]
+
+    render(request, 'my_app/welcome.html', {
+        'panels': panels,
+    })
+```
+
+the `my_app/welcome.html` template could render the panels as follows:
+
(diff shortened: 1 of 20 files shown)


gasman:
Rendered docs: https://wagtail--7425.org.readthedocs.build/en/7425/extending/template_components.html

docs/extending/template_components.md
```diff
@@ -0,0 +1,161 @@
...
+```python
+from wagtail.admin.ui.components import Component
+
+class WelcomePanel(Component):
+    template_name = 'my_app/panels/welcome.html'
+
+    def get_context(self, parent_context):
+        context = super().get_context(parent_context)
+        context['username'] = parent_context['request'].user.username
+        return context
```
ababic, line 68:
```suggestion
    def get_context_data(self, parent_context):
        context = super().get_context_data(parent_context)
        context['username'] = parent_context['request'].user.username
        return context
```

docs/reference/hooks.rst
```diff
@@ -1244,16 +1253,13 @@ Hooks for working with registered Snippets.
   :icon_name: icon to display against the menu item
   :classname: a ``class`` attribute value to add to the button element
   :is_shown: a method which returns a boolean indicating whether the menu item should be shown; by default, true except when editing a locked page
-  :template: the path to a template to render to produce the menu item HTML
-  :get_context: a method that returns a context dictionary to pass to the template
-  :render_html: a method that returns the menu item HTML; by default, renders ``template`` with the context returned from ``get_context``
-  :Media: an inner class defining Javascript and CSS to import when this menu item is shown - see `Django form media <https://docs.djangoproject.com/en/stable/topics/forms/media/>`_
 
-  The ``get_url``, ``is_shown``, ``get_context`` and ``render_html`` methods all accept a request object and a context dictionary containing the following fields:
+  The ``get_url``, ``is_shown``, ``get_context`` and ``render_html`` methods all accept a context dictionary containing the following fields:
```
ababic, line 1257:
```suggestion
  The ``get_url``, ``is_shown``, ``get_context_data`` and ``render_html`` methods all accept a context dictionary containing the following fields:
```

wagtail/admin/ui/components.py
```diff
@@ -0,0 +1,18 @@
+from django.forms import MediaDefiningClass
```
ababic, line 1:
```suggestion
from typing import Mapping

from django.forms import MediaDefiningClass
```

wagtail/admin/ui/components.py
```diff
@@ -0,0 +1,18 @@
...
+from django.template import Context
+from django.template.loader import get_template
+
+
+class Component(metaclass=MediaDefiningClass):
+    def get_context_data(self, parent_context: Context):
+        return {}
+
+    def render_html(self, parent_context=None):
```
ababic, line 10:
```suggestion
    def render_html(self, parent_context: Mapping[str,Any] = None) -> str:
```

ababic:
Hey @gasman. Feel free to ignore the type-hint tweaks above if you feel it's out-of-line with what we normally go for. With a very new, bare-bones Python abstraction like this though, I think a few type hints can make a huge difference.

gasman:
Thanks @ababic! Rebased with feedback applied.

docs/extending/template_components.md
```diff
@@ -0,0 +1,161 @@
...
+    template_name = 'my_app/panels/welcome.html'
+
+    class Media:
+        css = {
+            'all': ('my_app/css/welcome-panel.css',)
+        }
+```
+
+
+## Using components on your own templates
```
ababic, line 93:
```suggestion
## Using components in your own templates
```

docs/extending/template_components.md
```diff
@@ -0,0 +1,161 @@
+# Template components
+
+Working with objects that know how to render themselves as elements on an HTML template is a common pattern seen throughout the Wagtail admin. For example, the admin homepage is a view provided by the central `wagtail.admin` app, but brings together information panels sourced from various other modules of Wagtail, such as images and documents (potentially along with others provided by third-party packages). These panels are passed to the homepage via the [`construct_homepage_panels`](../reference/hooks.html#construct-homepage-panels) hook, and each one is responsible for providing its own HTML rendering. In this way, the module providing the panel has full control over how it appears on the homepage.
```
ababic, line 3:
```suggestion
Working with objects that know how to render themselves as HTML using a template is a common pattern seen throughout the Wagtail admin. For example, the admin homepage is a view provided by the central `wagtail.admin` app, but brings together information panels sourced from various other modules of Wagtail, such as images and documents (potentially along with others provided by third-party packages). These panels are passed to the homepage via the [`construct_homepage_panels`](../reference/hooks.html#construct-homepage-panels) hook, and each one is responsible for providing its own HTML rendering. In this way, the module providing the panel has full control over how it appears on the homepage.
```

wagtail/tests/testapp/wagtail_hooks.py
```diff
@@ -189,3 +192,33 @@ class FavouriteColourPanel(BaseSettingsPanel):
     order = 500
     form_class = FavouriteColourForm
     form_object = 'user'
+
+
+class ClippyPanel(Component):
+    order = 50
+
+    def render_html(self, parent_context):
+        return mark_safe("<p>It looks like you're making a website. Would you like some help?</p>")
```
ababic, line 201: Love it :)

ababic (changes requested):
One more tiny wording thing here. Maybe my suggestions could need some extra tweaking, but "on a template" didn't feel quite right

docs/extending/template_components.md
gasman: The latter version here isn't strictly true - components don't necessarily use templates for their own rendering. (They might just return `mark_safe("<p>some static HTML</p>")` from the `render_html` method instead.) The message I'm aiming to get across here is that we're often faced with a page template, with a bunch of objects to render into it, and we _aren't_ particularly concerned with where those objects come from or how they arrive at their HTML.

docs/extending/template_components.md
gasman: Mmm... to me "on your own templates" feels like it's more closely pinpointing the action of putting a tag on a page - which is what this section is about - while "in your own templates" would cover the broader concepts of using components, which would include things like setting up a hook to collect components from outside code. That might just be a personal quirk though, and I'm not married to it :-)

allcaps: `your own` -> `custom`?

docs/extending/template_components.md
```diff
@@ -0,0 +1,161 @@
...
+.. note::
+   Any object implementing this API can be considered a valid component; it does not necessarily have to inherit from the ``Component`` class described below, and user code that works with components should not assume this (for example, it must not use ``isinstance`` to check whether a given value is a component).
+```
+
+```eval_rst
+.. _creating_template_components:
+```
+## Creating components
+
+The simplest way to create a component is to define a subclass of `wagtail.admin.ui.components.Component` and specify a `template_name` attribute on it. The rendered template will then be used as the component's HTML representation:
```
allcaps, line 25: If this is the simplest way, and Wagtail itself uses this way... Should we mention this is the _preferred_ way?

allcaps (approved):

docs/extending/template_components.md
gasman: This section of the docs is aimed at developers building their own areas of the admin (e.g. as third-party packages), and it feels to me like that's beyond the point where it can reasonably be called "custom".

docs/extending/template_components.md
gasman: Yes, I think we can say that! In #7409 the Cell and Header objects don't use the Component base class, because while they outwardly behave as components they actually delegate the HTML rendering to the Column object - and I wanted to avoid implying that was the "wrong" way to do things. But that's really a bit of a special case, and it's fair to describe the Component base class as the preferred way even if there are occasional exceptions.

PR #7426, opened 2021-08-11
Bulk actions - Refactor action registration

KIRA009:
This pr changes the way actions are registered, making the integration of bulk actions with modeladmin easier


wagtail/admin/views/bulk_action/utils.py
```diff
@@ -0,0 +1,22 @@
...
+
+from wagtail.core import hooks
+
+
+def get_bulk_actions_from_model(app_label, model_name):
+    bulk_actions_list = []
+    bulk_actions = hooks.get_hooks('register_bulk_action')
+    model = apps.get_model(app_label, model_name)
+
+    for action_class in bulk_actions:
```
zerolab, line 11: suggestion: fail loudly if the `action_class` is not a sublcass of `BulkAction`

jacobtoppm (changes requested):
Looking good! A few suggested tweaks.

Additionally, I think it would be good to think about efficiency here. Currently we loop through all the bulk actions every time we want to find the list of actions for a model, or a specific action. It might be worth thinking about whether we can subdivide the bulk actions so future operations are a bit more efficient when we register them for the first time (see https://github.com/wagtail/wagtail/blob/main/wagtail/core/rich_text/feature_registry.py for a class that handles registration of a similar feature) - have a think about how you might do that?

wagtail/users/views/bulk_actions/delete.py
```diff
@@ -16,9 +16,17 @@ class DeleteBulkAction(UserBulkAction):
...
+        return {
+            **super().get_execution_context(),
+            'model': self.model
+        }
+
     @classmethod
-    def execute_action(cls, objects, **kwargs):
-        cls.model.objects.filter(pk__in=[obj.pk for obj in objects]).delete()
+    def execute_action(cls, objects, model=None, **kwargs):
+        if model is None:
```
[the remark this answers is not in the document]
KIRA009: >  we should fail loudly

I agree with this. The only case where `model` can be `None` is when its called from a management command, and the `model` argument is not added. I am not sure if we should assume the model from the `models` list, even if `len(models)` is 1

wagtail/users/views/bulk_actions/toggle_activity.py
```diff
@@ -45,7 +45,10 @@ def prepare_action(self, objects):
     @classmethod
     def execute_action(cls, objects, mark_as_active=False, model=None, **kwargs):
         if model is None:
-            return
+            if len(cls.models) == 1:
```
jacobtoppm, line 48: This feels like this could end up in a lot of actions - maybe something to factor out to reduce the boilerplate a bit?

jacobtoppm: Additionally, does this need to go in a few more actions here? Feels like this should be applicable to more actions

jacobtoppm (changes requested):

zerolab:
@KIRA009 heads up, #7440 introduced a few conflicts.

jacobtoppm (approved):
Nice job! Looking great now you've made those final tweaks.

PR #7428, opened 2021-08-12
Bulk actions - appropriate string for listing

KIRA009:
This pr is to display different strings on the action bar based on the listing

### DIFF
--- a/wagtail/admin/localization.py
+++ b/wagtail/admin/localization.py
@@ -109,10 +109,38 @@ def get_js_translation_strings():
         'WEEKDAYS_SHORT': [str(WEEKDAYS_ABBR[d % 7]) for d in range(-1, 6)],
 
         # used by bulk actions
-        'NUM_PAGES_SELECTED_SINGULAR': _('1 page selected'),
-        'NUM_PAGES_SELECTED_PLURAL': _("{0} pages selected"),
-        'NUM_PAGES_SELECTED_ALL': _("All {0} pages on this screen selected"),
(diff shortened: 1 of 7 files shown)


wagtail/admin/localization.py
```diff
@@ -109,10 +109,38 @@ def get_js_translation_strings():
         'WEEKDAYS_SHORT': [str(WEEKDAYS_ABBR[d % 7]) for d in range(-1, 6)],
 
         # used by bulk actions
-        'NUM_PAGES_SELECTED_SINGULAR': _('1 page selected'),
-        'NUM_PAGES_SELECTED_PLURAL': _("{0} pages selected"),
-        'NUM_PAGES_SELECTED_ALL': _("All {0} pages on this screen selected"),
-        'NUM_PAGES_SELECTED_ALL_IN_LISTING': _("All pages in listing selected"),
+        'BULK_ACTIONS': {
+            'page': {
```
jacobtoppm, line 113: We haven't nested `STRINGS` before, but I think that's okay here (the alternative would be to store this information on the template). However, I think the change in case is confusing: either `BULK_ACTIONS` and `page` (for instance) should both be capitalised, or neither should, as they both do a similar thing here, exposing another strings object.

jacobtoppm (changes requested):
Just a tiny tweak for consistency, and I think we're good to merge.

wagtail/admin/localization.py
KIRA009: Okay, I'll change them to all caps

jacobtoppm (approved):

Issue #7430, opened 2021-08-12
"Blog Index and Posts" section of tutorial: specify 'PUBLISH' when making new child pages


jacksorjacksor:
In the "Blog Index and Posts" section of the "Getting Started" tutorial, please add "Make sure you specify 'PUBLISH' when making new child pages instead of 'SAVE DRAFT' for this tutorial" or similar.

### Pertinent section of the Wagtail docs 

https://docs.wagtail.io/en/stable/getting_started/tutorial.html#blog-index-and-posts

### Details

In the "Getting Started" tutorial, the learner is instructed to "create a few blog posts" - however, the default option when publishing is "Save Draft". Because I'm logged in and am just exploring things locally, I hadn't noticed this/the importance of everything I've made so far being a "Draft".

This only becomes relevant when working through the "Overriding Context" section, which uses:
```python
blogpages = self.get_children().live().order_by('-first_published_at')
```
This was returning empty for me, so after a bit of hacking I realised it was because none of my child pages satisfied "live()", as they were all drafts. For this to work successfully the pages need to have "PUBLISH" selected, which I don't believe is currently specified.

Thanks!

Rich

CullenSharp:
Hello, I'd like to take this issue on.

jacobtoppm:
Addressed in https://github.com/wagtail/wagtail/commit/2d741386c4f384f1825b9d45522e03806391c239, thanks for raising!

PR #7431, opened 2021-08-12
Added 'PUBLISH' when making new child pages #7430

DeviousLab:
Added extra instruction to publish a blog post rather than save as a draft for the overriding section.

Aims to solve issue #7430

### DIFF
--- a/docs/getting_started/tutorial.md
+++ b/docs/getting_started/tutorial.md
@@ -349,6 +349,9 @@ other page type.
 
 ![](../_static/images/tutorial/tutorial_5.png)
 
+Make sure you specify 'PUBLISH' when making new blog posts instead of
+'SAVE DRAFT' for this tutorial at the bottom of the page.
+
 You should now have the very beginnings of a working blog.
 Access the `/blog` URL and you should see something like this:
 
--- a/docs/getting_started/tutorial.md
+++ b/docs/getting_started/tutorial.md
@@ -159,7 +159,7 @@ Run `python manage.py makemigrations` (this will create the migrations file), th
 changes). You must run the above commands each time you make changes to
 the model definition.
 
-You can now edit the homepage within the Wagtail admin area (go to Pages, Homepage, then Edit) to see the new body field. Enter some text into the body field, and publish the page.
+You can now edit the homepage within the Wagtail admin area (go to Pages, Homepage, then Edit) to see the new body field. Enter some text into the body field, and publish the page by selecting *Publish* at the bottom of the page editor, rather than *Save Draft*.
 
 The page template now needs to be updated to reflect the changes made
 to the model. Wagtail uses normal Django templates to render each page
@@ -349,8 +349,7 @@ other page type.
 
 ![](../_static/images/tutorial/tutorial_5.png)
 
-Make sure you specify 'PUBLISH' when making new blog posts instead of
-'SAVE DRAFT' for this tutorial at the bottom of the page.
+Be sure to *Publish* the blog post after you are done.
 
 You should now have the very beginnings of a working blog.
 Access the `/blog` URL and you should see something like this:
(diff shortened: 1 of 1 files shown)


docs/getting_started/tutorial.md
```diff
@@ -349,6 +349,9 @@ other page type.
 
 ![](../_static/images/tutorial/tutorial_5.png)
 
+Make sure you specify 'PUBLISH' when making new blog posts instead of
+'SAVE DRAFT' for this tutorial at the bottom of the page.
```
jacobtoppm, line 353: I don't think this is very clear: it sounds like the tutorial is at the bottom of the page. Perhaps instead we should be changing "Now create a few blog posts as children of `BlogIndexPage`" -> "Now create and publish a few blog posts as children of `BlogIndexPage`", as by this stage in the tutorial, they have already been asked to publish a page (the index). If any further instructions on how to publish are needed, they should probably go up with the original direction to publish the index.

jacobtoppm (changes requested):
Thanks for contributing! I think this will definitely make the tutorial easier to follow, but I think we need to think about the wording a bit more.

docs/getting_started/tutorial.md
DeviousLab: Appreciate the advice, would you say it would be better to say:
"Now create and publish a few blog posts as children of `BlogIndexPage` by specifying 'PUBLISH' when making new blog posts instead of 'SAVE DRAFT'."?

jacobtoppm: I would say:
- Just say "publish" here, because they've already published a page at this point in the tutorial.
- Add the guidance about "by clicking 'Publish' in the menu at the bottom of the page editor, rather than 'Save draft'" or similar to the instructions about publishing the index page, since that's the first time they're doing it.
I'd also avoid the caps, for preference - I don't think we use them elsewhere in the docs (see https://github.com/wagtail/wagtail/blob/main/docs/editor_manual/new_pages/previewing_and_submitting_for_moderation.rst for how we talk about the buttons elsewhere)

DeviousLab: I think the following should be fine now?
"Now create and publish a few blog posts as children of `BlogIndexPage` by switching to *Publish* at the bottom of the page editor, rather than *Save Draft*."

jacobtoppm: Haha, sorry, I think we might be looking at slightly different sections. If you see https://github.com/wagtail/wagtail/blame/main/docs/getting_started/tutorial.md#L162, you can see that to get to this bit of the tutorial, they've already published a page. I think that's where "by selecting Publish at the bottom of the page editor, rather than Save Draft." belongs, because that's the first time they're performing the action.

Down here, I think we just need to say that the child pages need to be published, rather than created. Hope that makes sense!

DeviousLab: Ah I see what you mean now! I'll make the necessary changes then for you to review.

jacobtoppm: Awesome, thank you!

docs/getting_started/tutorial.md
```diff
@@ -349,6 +349,8 @@ other page type.
 
 ![](../_static/images/tutorial/tutorial_5.png)
 
+Be sure to *Publish* the blog post after you are done.
```
jacobtoppm, line 352: How about "*Publish* each blog post after you are done editing."? (Just because there's not just a single blog post). But other than that, looks great!

jacobtoppm:
Thanks @DeviousLab! Are you happy to be listed in the contributors file, and if so, would that be as Himesh Samarasekera?

DeviousLab:
Yes that would be fine with me!

jacobtoppm:
Added in https://github.com/wagtail/wagtail/commit/2d741386c4f384f1825b9d45522e03806391c239, thanks @DeviousLab!

PR #7440, opened 2021-08-16
Bulk actions - Minor fixes

KIRA009:
This pr contains minor refactors discussed in previous prs:
- Uses bulk updates and deletes wherever possible
- form fields are added as class variables wherever applicable
- Required kwargs are explicitly mentioned when overriding `execute_action`
- Page bulk actions work even when `user` is `None`, useful for management commands

### DIFF
--- a/wagtail/admin/views/pages/bulk_actions/move.py
+++ b/wagtail/admin/views/pages/bulk_actions/move.py
@@ -99,8 +99,6 @@ class MoveBulkAction(PageBulkAction):
         if destination is None:
             return
         user = kwargs.get('user', None)
-        if user is None:
-            return
         for page in objects:
             if not page.permissions_for_user(user).can_move_to(destination):
                 continue
--- a/wagtail/admin/views/pages/bulk_actions/move.py
+++ b/wagtail/admin/views/pages/bulk_actions/move.py
@@ -10,6 +10,11 @@ from wagtail.core.models import Page
 
 
 class MoveForm(forms.Form):
+    move_applicable = forms.BooleanField(
+        label=_("Move only applicable pages"),
+        required=False
+    )
+
     def __init__(self, *args, **kwargs):
         destination = kwargs.pop('destination')
         super().__init__(*args, **kwargs)
@@ -18,10 +23,6 @@ class MoveForm(forms.Form):
             queryset=Page.objects.all(),
             widget=widgets.AdminPageChooser(can_choose_root=True, user_perms='move_to'),
             label=_("Select a new parent page"),
-        )
-        self.fields['move_applicable'] = forms.BooleanField(
-            label=_("Move only applicable pages"),
-            required=False
         )
 
 
--- a/wagtail/admin/views/pages/bulk_actions/move.py
+++ b/wagtail/admin/views/pages/bulk_actions/move.py
@@ -94,7 +94,7 @@ class MoveBulkAction(PageBulkAction):
         }
 
     @classmethod
-    def execute_action(cls, objects, destination=None, user=None):
+    def execute_action(cls, objects, destination=None, user=None, **kwargs):
         if destination is None:
             return
         for page in objects:
--- a/wagtail/admin/views/pages/bulk_actions/move.py
+++ b/wagtail/admin/views/pages/bulk_actions/move.py
@@ -94,11 +94,9 @@ class MoveBulkAction(PageBulkAction):
         }
 
     @classmethod
-    def execute_action(cls, objects, **kwargs):
-        destination = kwargs.get('destination', None)
+    def execute_action(cls, objects, destination=None, user=None):
         if destination is None:
             return
-        user = kwargs.get('user', None)
         for page in objects:
             if not page.permissions_for_user(user).can_move_to(destination):
                 continue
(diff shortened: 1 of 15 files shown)


wagtail/admin/views/pages/bulk_actions/publish.py
```diff
@@ -32,19 +32,15 @@ def get_execution_context(self):
...
+    def execute_action(cls, objects, include_descendants=False, user=None):
         for page in objects:
             revision = page.save_revision(user=user)
             revision.publish(user=user)
             cls.num_parent_objects += 1
 
             if include_descendants:
                 for draft_descendant_page in page.get_descendants().not_live().defer_streamfields().specific():
-                    if draft_descendant_page.permissions_for_user(user).can_publish():
+                    if user is None or draft_descendant_page.permissions_for_user(user).can_publish():
```
zerolab, line 43: At first this bit here feelt a tad 🤔 but it works with the idea that `{type}BulkAction.execute_action()` will just do the thing if called directly in code. The same applies to Page `publish()`, so all good!

wagtail/documents/views/bulk_actions/add_tags.py
```diff
@@ -32,8 +28,7 @@ def get_execution_context(self):
         }
 
     @classmethod
-    def execute_action(cls, objects, **kwargs):
-        tags = kwargs.get('tags', [])
+    def execute_action(cls, objects, tags=[]):
```
zerolab, line 31: suggestion: return early if no tags are provided

wagtail/images/views/bulk_actions/add_tags.py
```diff
@@ -32,8 +28,7 @@ def get_execution_context(self):
         }
 
     @classmethod
-    def execute_action(cls, images, **kwargs):
-        tags = kwargs.get('tags', [])
+    def execute_action(cls, images, tags=[]):
```
zerolab, line 31: ditto: if no tags, return early

wagtail/admin/views/pages/bulk_actions/publish.py
```diff
@@ -32,19 +32,15 @@ def get_execution_context(self):
         }
 
     @classmethod
-    def execute_action(cls, objects, **kwargs):
-        include_descendants = kwargs.get('include_descendants', False)
-        user = kwargs.get('user', None)
-        if user is None:
-            return
+    def execute_action(cls, objects, include_descendants=False, user=None):
```
zerolab, line 35: Should we change the base class `get_execution_context()` output, this will explode with a `TypeError`. 

suggestion: leave `**kwargs` in since `BulkAction` has the following method definition: `def execute_action(cls, objects, **kwargs)`

zerolab (changes requested):
Lots of nice tweaks @KIRA009 

Added a few questions and some tweak suggestions

wagtail/admin/views/pages/bulk_actions/move.py
```diff
@@ -19,10 +24,6 @@ def __init__(self, *args, **kwargs):
             widget=widgets.AdminPageChooser(can_choose_root=True, user_perms='move_to'),
             label=_("Select a new parent page"),
         )
```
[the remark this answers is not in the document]
KIRA009: Looks like I missed this!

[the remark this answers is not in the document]
KIRA009: It uses the `destination` variable
```
self.fields['chooser'] = forms.ModelChoiceField(
            initial=destination,
```

wagtail/documents/views/bulk_actions/add_to_collection.py
```diff
@@ -39,8 +39,8 @@ def get_execution_context(self):
         }
 
     @classmethod
-    def execute_action(cls, objects, **kwargs):
-        cls.collection = kwargs.get('collection', None)
+    def execute_action(cls, objects, collection=None):
+        cls.collection = collection
```
jacobtoppm, line 43: There seems to still be a lot of storing variables on the class here. Is that necessary? If we can, I think we should just use local variables here (and in the other bulk actions that use class attributes)

zerolab: I had a similar thought about this specific one. The collection name is used in the success message. Two options:
1. make the success message generic
2. come up with a different way of passing context

jacobtoppm: Personally, I'd favour 2 - since `collection` is being passed in, it seems reasonable to set eg `self.collection` elsewhere, and use that in the success message (or give that the `execution_context` as kwargs as well, I suppose, but doing it on the instance is probably simpler). As for the others that store the number of updated objects on the class, that seems like a reasonable variable for `execute_action` to return. What do you think?

KIRA009: > As for the others that store the number of updated objects on the class, that seems like a reasonable variable for execute_action to return

Are you referring to the `num_parent_objects` and `num_child_objects` variables? If so, then yes I think returning them as output for the `execute_action` method makes sense

jacobtoppm: Yes, that's what I was thinking of.

PR #7444, opened 2021-08-18
Use a mixin instead of properties for various image file related methods

fabienheureux:
# Description

In order to use Willow on rendtion, I had to copy-paste methods set on `AbstractImage` on my `CustomRenditionModel`. 
It could be useful to have these methods defined for both `AbstractImage` and `AbstractRendition`. 

Nothing has been changed in these methods, they have just been moved.

## `ImageFileMixin`
A new mixin has been added, that could now be used by both `AbstractImage`  and `AbstractRendition` models.  
Everything in this mixin is related to image files, hence can be useful both for images and renditions.


lb-:
Looking good in principle, ned to do some local testing still though.
@zerolab  noted that this might be useful for https://github.com/GrappleGQL/wagtail-grapple/pull/182

lb-:
@fabienheureux thanks so much for taking the time to put this PR up, I have validated a few image handling scenarios and looks great. I pushed to your branch + rebased and have merged it into main.

fabienheureux:
Thanks @lb- ! 
Yes I did this initially for a grapple PR I need to finish. 
This will help and reduce the added code on Grapple's side.

PR #7445, opened 2021-08-18
MySQL Search Backend

ACMCMC:
This PR adds the implementation of the MySQL search backend, which completes the implementation of the new backends for the out-of-the-box search capabilities that Wagtail will now have.


wagtail/search/backends/database/mysql/query.py
```diff
@@ -0,0 +1,208 @@
...
+        combined_sql = '({}{} {}{})'.format(lhs_connector, lsql, rhs_connector, rsql)  # if self.connector is '+' (AND), then both terms will be ANDed together. We need to repeat the connector to make that work.
+        combined_value = combined_sql % tuple(value_params)
+        return '%s', [combined_value]
+
+
+class MatchExpression(Expression):
+    filterable = True
+    template = 'MATCH (%s) AGAINST (%s IN BOOLEAN MODE)'
+
+    def __init__(self, query: SearchQueryCombinable, columns: List[str] = ['title', 'body'], output_field: Field = BooleanField()) -> None:
```
kaedroho, line 198: The best practice in Python is to not use lists as default parameters since they are initialised only once and reused (so if it's modified in any way, the next call will get the modified list and not a fresh one).

wagtail/search/backends/database/mysql/mysql.py
```diff
@@ -0,0 +1,587 @@
...
+        if isinstance(self.query, Not):  # If the outermost operator is a Not, we invert the query. This is done because, if every search term is negated, the Not() will return no results, an we want to match all results instead.
+            query = self.query.subquery
+            negated = True
+        else:
+            query = self.query
+            negated = False
+
+        search_query = self.build_search_query(query)
+        match_expression = MatchExpression(search_query, output_field=BooleanField())  # For example: MATCH (`title`, `body`) AGAINST ('+query' IN BOOLEAN MODE)
+        score_expression = MatchExpression(search_query, output_field=FloatField())  # Same as above, but with a different output_field
```
kaedroho, line 411: Looks like ``title_norm`` isn't being used. Is it possible to create a separate ``MatchExpression`` for `body` and `title` and multiply the `title` by `title_norm`?

in the postgres backend, it's done here: https://github.com/wagtail/wagtail/blob/1b50c744240b1f6ddaf5abd99c2ea26d17c5cdfb/wagtail/contrib/postgres_search/backend.py#L494


As a total guess (sorry if this is wrong!), but do you think something like this would work?

```python
score_expression = (
    MatchExpression(search_query, columns=['title'], output_field=FloatField()) * F('title_norm') +
    MatchExpression(search_query, columns=['body'], output_field=FloatField())
)
```

kaedroho (approved):
Apart from a couple of small things, this looks good to me, thanks!

There's definitely lots of repeated code between the backends, but I think it's fine to merge them as they are for now and clean this up later

kaedroho:
Rebased on `gsoc-search` here: https://github.com/wagtail/wagtail/tree/gsoc-search-mysql

ACMCMC:
I have fixed a bug that appeared when we started doing separate `MATCH` queries on `title` and `body` for searching 👍🏻

PR #7446, opened 2021-08-18
Refactor documents index into CBV and eliminate is_ajax

gasman:
Repeat of #7251 (and indirectly #7209), for documents this time.

### DIFF
--- a/wagtail/documents/views/documents.py
+++ b/wagtail/documents/views/documents.py
@@ -5,7 +5,9 @@ from django.core.paginator import Paginator
 from django.shortcuts import get_object_or_404, redirect
 from django.template.response import TemplateResponse
 from django.urls import reverse
+from django.utils.decorators import method_decorator
 from django.utils.translation import gettext as _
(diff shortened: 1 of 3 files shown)


wagtail/documents/views/documents.py
```diff
@@ -22,74 +23,83 @@
 permission_checker = PermissionPolicyChecker(permission_policy)
 
 
-@permission_checker.require_any('add', 'change', 'delete')
-@vary_on_headers('X-Requested-With')
-def index(request):
-    Document = get_document_model()
+class BaseListingView(TemplateView):
+    @method_decorator(permission_checker.require_any('add', 'change', 'delete'))
+    def get(self, request):
```
zerolab, line 28: question: given all this does is call `TemplateView`'s counterparts, is it needed?

lb-:
Looks great @gasman - I ran the code locally & tested a bunch of scenarios, works great. Love that the index & search results are separated now, and both have their own URL name

PR #7461, opened 2021-08-19
Fix typo in StructBlock documentation

jnns:
The StructValue class `LinkValue` is refrerenced as `LinkStructValue` in the Meta class.


gasman:
Thanks @jnns!

PR #7464, opened 2021-08-24
update stylelint & config-stylelint-wagtail

lb-:
On a path to resolve the linting warning `Deprecation Warning: 'declaration-property-value-blacklist' has been deprecated. Instead use 'declaration-property-value-disallowed-list'` See [7](https://github.com/wagtail/stylelint-config-wagtail/issues/7).

However it appears that we are not on the latest stylelint or wagtail stylelint config so updating those first.


thibaudcolas (approved):
Thank you @lb-!

PR #7465, opened 2021-08-25
Remove is_ajax checks from multiple upload views

gasman:
is_ajax is scheduled for removal in Django 4.0, since Django no longer considers the X-Requested-With header as a definitive indicator of whether the request came from Javascript. (See under https://docs.djangoproject.com/en/3.2/releases/3.1/#id2.) On that logic, there is no real value in enforcing it in our AJAX-only views, and there's no harm in removing it (it just means that if anyone ever does make a non-AJAX request to that URL, they'll get an HTML fragment back rather than a 400 error, which is equally good at communicating the message that the URL endpoint isn't directly usable in the browser).

### DIFF
--- a/wagtail/admin/views/generic/multiple_upload.py
+++ b/wagtail/admin/views/generic/multiple_upload.py
@@ -119,9 +119,6 @@ class AddView(PermissionCheckedMixin, TemplateView):
         }
 
     def post(self, request):
-        if not request.is_ajax():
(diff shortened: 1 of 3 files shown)


lb-:
Thanks @gasman  - looks good to me, merging in.

Will be great to get our tests running without these warnings eventually also haha.

PR #7467, opened 2021-08-27
Fix home icon in breadcrumbs for limited-permission users, fixes #7348

### LINKED ISSUE #7348: Breadcrumbs: home icon not displayed for restricted site tree
### Issue Summary

If a user only has access to a subtree, the breadcrumbs at the top don't show the home icon correctly.
Superuser:
Superuser

User with subtree-only access:
User with subtree-only access

### Possible Fix

It seems like [this line](https://github.com/wagtail/wagtail/blob/ec0daf604a366b18c842a13e1821391a5e2abdcd/wagtail/admin/templates/wagtailadmin/shared/breadcrumb.html#L13) could be synced with [that line](https://github.com/wagtail/wagtail/blob/ec0daf604a366b18c842a13e1821391a5e2abdcd/wagtail/admin/templates/wagtailadmin/shared/breadcrumb.html#L9), so:

- remove `class="text-replace"`
- move `{% icon name="arrow-right" class_name="arrow_right_icon"%}` into the `<a>`-tag

If I do this in the browser's dev-tools, the "home"-link works again and it looks like this:
fixed breadcrumbs template

I don't know, if there are any issues with the above changes, but it seems to work for me.

### Steps to Reproduce

1. Start a new project with `wagtail start myproject` + migrate + createsuperuser
2. Create a page beneath the created home page
3. Create a new user, assign him to the moderators group
4. Change the moderators group: remove the root page from the site permissions and add the created sub page instead
5. With the new moderators-user, open his "root" page in the cms

### Technical details

* Wagtail version: 2.13.4
* Browser version: Firefox 90, Google Chrome 91

th3hamm0r:
The html has just been synced with the html for unlimited users.

See #7348

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -24,6 +24,7 @@ Changelog
  * Fix: Admin fields and Icon visibility issues in Windows high contrast mode (Dmitrii Faiazov)
  * Fix: `blocks.MultipleChoiceBlock`, `forms.CheckboxSelectMultiple` and `ArrayField` checkboxes will now stack instead of display inline to align with all other checkboxes fields (Seb Brown)
  * Fix: Screen readers can now access login screen field labels (Amy Chan)
+ * Fix: Admin breadcrumbs home icon now shows for users with access to a subtree only (Stefan Hammer)
 
 
 2.14.1 (12.08.2021)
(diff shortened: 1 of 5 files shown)


lb-:
@th3hamm0r thanks for this fix, works as expected, really appreciate the detailed steps on the bug you raised.

I have pushed to your branch (with release notes & contributors added), please let me know if you want to see a different name on the contributors listing.

Screen Shot 2021-08-30 at 7 53 36 pm

PR #7468, opened 2021-08-27
Update form label css to be readable by screen reader

### LINKED ISSUE #7432: Screenreader: Login screen username label doesn't exist
### Issue Summary

When trying to login to wagtail using JAWs the username field doesn't have a label, making it tricky to understand what's required

### Steps to Reproduce

1. Load a wagtail admin login screen in JAWS (/admin usually)
2. Navigate to the username field
3. Only "Edit" will be announced for the input field, 

This should be "Enter username"

Image showing what's displayed in JAWs:
image (2)

Inspector in Chrome/Firefox reports that the labels don't meet WCAG requirements - they seem to be display: none at the moment:
image (3)

amychan331:
Fixes #7432
The label can't be read by screen reader since it was set to display none. This hides it without using display none.

* For front-end changes:
    * Please list the exact browser and operating system versions you tested.
      * Firefox, Chrome, and Safari on MacOS
    * Please list which assistive technologies you tested.
      * VoiceOver and Firefox accessibility inspector.

### DIFF
--- a/wagtail/admin/static_src/wagtailadmin/scss/layouts/login.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/layouts/login.scss
@@ -82,12 +82,7 @@ input[type=checkbox]:before {
         padding: 0;
 
         label {
-            position: absolute;
-            left: -10000px;
-            top: auto;
-            width: 1px;
-            height: 1px;
-            overflow: hidden;
+            @include visuallyhidden;
         }
 
         input {
--- a/wagtail/admin/static_src/wagtailadmin/scss/layouts/login.scss
+++ b/wagtail/admin/static_src/wagtailadmin/scss/layouts/login.scss
@@ -82,7 +82,12 @@ input[type=checkbox]:before {
         padding: 0;
 
         label {
-            display: none;
+            position: absolute;
+            left: -10000px;
+            top: auto;
+            width: 1px;
+            height: 1px;
+            overflow: hidden;
         }
 
         input {


wagtail/admin/static_src/wagtailadmin/scss/layouts/login.scss
```diff
@@ -82,7 +82,12 @@ input[type=checkbox]:before {
         padding: 0;
 
         label {
-            display: none;
+            position: absolute;
+            left: -10000px;
+            top: auto;
+            width: 1px;
+            height: 1px;
+            overflow: hidden;
```
lb-, line 90: Just wondering if the mixin `visuallyhidden` would be best to use here, to avoid duplication?

See definition - client/scss/tools/_mixins.general.scss

lb- (changes requested):

wagtail/admin/static_src/wagtailadmin/scss/layouts/login.scss
amychan331: I did not realized there is a mixin for it already. Thank you for the heads up. I just added the commit with visuallyhidden change.

lb- (approved):

lb-:
@amychan331 thanks for the fix! this is great, I have validated and it works as expected.

I have merged this in via f1b3bbc52f2944dde0a37d1c33b47f44d03d4a76 (a single commit with release notes & changelog, plus have added you to the list of contributors, let me know if you want me to change the name).

We really appreciate you taking the time to dig into this and find a fix, accessibility is super important.

Screen Shot 2021-08-30 at 7 11 15 pm

PR #7469, opened 2021-08-27
Frontend tweaks for new slim sidebar

kaedroho:


jacobtoppm (approved):
LGTM! Only thing is I think the circular Wagtail background looks a little weird in slim mode, but that's something for design

kaedroho:
> LGTM! Only thing is I think the circular Wagtail background looks a little weird in slim mode, but that's something for design

Yep, I couldn't find a better way but we can always adjust this later. Thanks for the review!

PR #7470, opened 2021-08-27
Typed table block

gasman:
A contrib app providing a new experimental StreamField block type, `TypedTableBlock`. This addresses the common request for a TableBlock that supports rich text (and more), without just exposing a free-for-all HTML field. It does this by embedding an arbitrary StreamField block in each cell. (Some block types will work better than others, of course...)

TypedTableBlock is passed a list of child blocks, much like StreamBlock (and can also be subclassed like StreamBlock):

    # add 'wagtail.contrib.typed_table_block' to INSTALLED_APPS, then

    from wagtail.contrib.typed_table_block.blocks import TypedTableBlock

    # and within a StreamField definition:

    table = TypedTableBlock([
        ('text', CharBlock()),
        ('rich_text', RichTextBlock(features=['bold', 'italic', 'ol', 'ul', 'link', 'document-link'])),
        ('number', FloatBlock()),
        ('country', ChoiceBlock(choices=[
            ('be', 'Belgium'),
            ('fr', 'France'),
            ('de', 'Germany'),
            ('nl', 'Netherlands'),
            ('pl', 'Poland'),
            ('uk', 'United Kingdom'),
        ])),
    ])

Adding one of these blocks to a StreamField will give an initially empty table with no columns. On adding a new column, the editor is prompted to select a block type as the type for that column. All cells within that column will then use the editing interface for that block.

Screenshot 2021-08-27 at 18 54 17

TODO for MVP:

- [x] Front-end rendering
- [x] Inserting / deleting rows and columns in arbitrary positions
- [x] Validation (i.e. the block `clean` method)
- [x] Styling
- [x] getTextLabel / focus
- [x] Tests - backend
- [x] Tests - frontend
- [x] Docs

Other more complex potential improvements (some of these will probably need new block APIs to support interaction between blocks):

* Row / column reordering
* A single shared toolbar for all rich text cells, rather than each one having its own toolbar
* Keyboard navigation between cells (e.g. arrow keys)

### DIFF
--- a/client/src/entrypoints/contrib/typed_table_block/__snapshots__/typed_table_block.test.js.snap
+++ b/client/src/entrypoints/contrib/typed_table_block/__snapshots__/typed_table_block.test.js.snap
@@ -8,7 +8,7 @@ exports[`wagtail.contrib.typed_table_block.blocks.TypedTableBlock it renders cor
         <div class=\\"typed-table-block__wrapper\\">
           <table>
             <thead>
-              <tr><th><input type=\\"hidden\\" name=\\"mytable-column-0-type\\" value=\\"test_block_a\\"><input type=\\"hidden\\" name=\\"mytable-column-0-order\\" value=\\"0\\"><button type=\\"button\\" class=\\"button button-small button-secondary prepend-column\\" aria-label=\\"Insert column\\" title=\\"Insert column\\">+</button><input name=\\"mytable-column-0-heading\\" class=\\"column-heading\\" placeholder=\\"Column heading\\"><button type=\\"button\\" class=\\"button button-secondary button-small button--icon text-replace white no delete-column\\" aria-label=\\"Delete column\\" title=\\"Delete column\\">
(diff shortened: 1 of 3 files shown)


kaedroho:
A couple of questions:

Would we remove the list of options if there is only one cell type?
If yes, could this be a replacement for `TableBlock`?

I'm wondering, if the answers to the previous questions are yes, if we should define this in Wagtail core and call it `TableBlock`, but keep the contrib app around for backwards compatibility? Then we would have one block type instead of two doing almost the same thing (but hard to migrate between)

gasman:
@kaedroho I'm hesitant to push this as a direct replacement for TableBlock, as the UI isn't quite as fluid (you just know people are going to be judging it on how close it is to Excel, after all :-) ) and it's designed to deal with a narrower use-case than TableBlock - namely the 'SQL-like' perspective where rows are 'things' and columns are properties of those things, rather than the other way round. While I can't think of any scenarios where TypedTableBlock would prevent doing something that was possible with TableBlock, it feels like the sort of thing where as soon as we go ahead with the change, real-world problems will surface where the new solution isn't as good.

gasman:
Arbitrary row/column insertion/deletion is now implemented.

Caylu:
So the development work on this request is completed? 
Thank you to all those who are working on this important feature.

gasman:
@Caylu It's not completed yet - see the TODO checkboxes above.

Caylu:
@gasman thank you. I didn't see it.

Caylu:
@gasman Is it okay to ask if there is any timeline for this feature to be available?

allcaps:
Reference, I made a PR against gasman:feature/typed-table-block at https://github.com/gasman/wagtail/pull/6

PR #7471, opened 2021-08-29
Clean up document chooser modal

gasman:
Eliminate a lot of the spaghetti code in the document chooser modal (and fix a few bugs that were caused by said spaghetti), following the strategy previously used for the snippet chooser in #7379:

* the listing template no longer piggybacks on the index page's listing template (which means it no longer has column reordering links that point to the wrong URL, and will make both views easier to convert to the generic listing framework #7409)
* the 'results HTML fragment only' endpoint used when paginating / searching is now a distinct view, rather than an alternative behaviour that kicks in in the presence of certain URL parameters (and the contexts of those two cases have been made consistent so that the Collection column doesn't disappear when you start searching / paginating)
* pagination links now use the standard pagination_nav.html include, instead of the hacky ajax_pagination_nav.html that needs a bunch of supporting JS
* the upload form only re-renders its own tab on validation errors, rather than the whole modal (which meant having to replicate the state of the listing - which it did very badly, and which has led to wagtail-generic-chooser becoming way too overengineered while trying to replicate this pattern...)

### DIFF
--- a/client/src/entrypoints/documents/document-chooser.js
+++ b/client/src/entrypoints/documents/document-chooser.js
@@ -12,7 +12,7 @@ function createDocumentChooser(id) {
   createDocumentChooser. State is either null (= no document chosen) or a dict of id, title and
   edit_link.
 
-  The result returned from the document chooser modal (see get_document_result_data in
+  The result returned from the document chooser modal (see get_document_chosen_response in
   wagtail.documents.views.chooser) is a superset of this, and can therefore be passed directly to
   chooser.setState.
   */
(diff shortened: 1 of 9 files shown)


lb-:
@gasman  spent a bit of time validating the changes across various document chooser scenarios (normal field, rich text, block) and reading through the code, this looks great  and I have merged in.

I have just rebased as the commits were useful in their granular state I thought, hope that is ok.

I have not updated changelog/release notes but let me know if you want me to do that also.

PR #7477, opened 2021-08-30
Fix: Update elasticsearch path in documentation

cspollar:
New location: https://github.com/wagtail/wagtail/blob/main/wagtail/search/backends/elasticsearch7.py


lb-:
@cspollar  thanks so much for this typo fix!

PR #7480, opened 2021-08-31
add support for non_field_errors rendering in workflow action modal

### LINKED ISSUE #7479: Add support for non_field_errors in Workflow action modal form

* When adding [non field errors](https://docs.djangoproject.com/en/3.2/ref/forms/api/#django.forms.Form.non_field_errors) to a Workflow action form (e.g. on the form.clean method), these validation errors do not show up in the UI.
* This means the user cannot submit the form but is not provided any feedback as to why.

* By default, the template `wagtail/admin/templates/wagtailadmin/pages/workflow_action_modal.html` should contain the ability to render non field validation errors.
* The simplest way to do this would be to add `{% include "wagtailadmin/shared/non_field_errors.html" %}` as per other forms within the Wagtail admin UI.
* Example screenshot:
Screen Shot 2021-08-31 at 12 16 27 pm

* Customised Wagtail instances can easily override the template mentioned above and do this themselves but this comes with a lot of extra boilerplate template code needing to be copied.
* Never use non-field errors, only use field errors when validating a form.

### Additional context

To reproduce this scenario, a custom `Task` will need to be created, below is the code to add to your `models.py` to create a simple form that requires a specific string to be submitted. I know this can be done with field validation but it is just the simplest example I can think of.

Once the model is created, create a new workflow that uses a new instance of this Task type.

Alternatively, check out [this branch](https://github.com/lb-/bakerydemo/tree/bug-report/steps-for-workflow-form-error) and run migrations.

```python
from django.forms import CharField, ValidationError

from wagtail.core.forms import TaskStateCommentForm
from wagtail.core.models import Task


class ApprovalConfirmationForm(TaskStateCommentForm):

    confirmation_agreement = CharField(
        label="Agreement",
        help_text="Type in 'I AGREE'",
        required=True,
    )

    def clean(self):
        cleaned_data = super().clean()

        if "confirmation_agreement" in cleaned_data:
            if cleaned_data["confirmation_agreement"] != "I AGREE":
                # ValidationError should  appear in non_field_errors
                raise ValidationError('Must enter "I AGREE"')

            del cleaned_data["confirmation_agreement"]

        return cleaned_data


class RequiredApprovalTask(Task):
    def get_actions(self, page, user):

        return [
            ("approve", "Approve", True),
        ]

    def get_form_for_action(self, action):
        return ApprovalConfirmationForm
```

lb-:
* fixes #7479
* Do the tests still pass? 👍 
* For Python changes: Have you added tests to cover the new/fixed behaviour? 👍 
* Note: I tried for a while to get a basic scenario working in the tests for failed form submission but could not, hopefully the test added (which checks that the template is included) is sufficient.

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/pages/workflow_action_modal.html
+++ b/wagtail/admin/templates/wagtailadmin/pages/workflow_action_modal.html
@@ -2,6 +2,8 @@
 {% include "wagtailadmin/shared/header.html" with title=action_verbose icon="clipboard-list" %}
 
 <div class="nice-padding">
+    {% include "wagtailadmin/shared/non_field_errors.html" %}
+    
     <form action="{{ submit_url }}" method="POST" novalidate>
         {% csrf_token %}
         <ul class="fields">
--- a/wagtail/admin/tests/test_workflows.py
+++ b/wagtail/admin/tests/test_workflows.py
@@ -1100,6 +1100,7 @@ class TestApproveRejectWorkflow(TestCase, WagtailTestUtils):
         response = self.client.get(reverse('wagtailadmin_pages:collect_workflow_action_data', args=(self.page.id, 'approve', self.page.current_workflow_task_state.id)))
         self.assertEqual(response.status_code, 200)
         self.assertTemplateUsed(response, 'wagtailadmin/pages/workflow_action_modal.html')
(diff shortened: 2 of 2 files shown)


zerolab (approved):

lb-:
Thanks @zerolab  - I will merge in and do release notes as a separate commit.

PR #7481, opened 2021-08-31
Completely hide sidebar in mobile mode

kaedroho:
This PR completely hides the sidebar on mobile, and adds a hamburger icon to allow toggling it.

image
image

Extracted from https://github.com/wagtail/wagtail/pull/7362
Based on https://github.com/wagtail/wagtail/pull/7469


client/src/components/Sidebar/Sidebar.tsx
```diff
@@ -180,8 +184,9 @@ export const Sidebar: React.FunctionComponent<SidebarProps> = (
       <aside
```
jacobtoppm, line 184: Question: would `<nav>` be more appropriate than `<aside>` here?

client/src/components/Sidebar/Sidebar.tsx
```diff
@@ -58,6 +58,11 @@ export const Sidebar: React.FunctionComponent<SidebarProps> = (
   // space next to the content
   const [peeking, setPeeking] = React.useState(false);
 
+  // 'open' indicates whether the sidebar is visible or not.
+  // this is meant to be used to show/hide the menu on small screens.
+  // it only has effect on small screens.
+  const [open, setOpen] = React.useState(false);
```
jacobtoppm, line 64: Could we rename `open` to something more intuitive here? I feel like the set of variables we've got to describe the menu state is getting slightly confusing with this addition (since I'd probably expect `open` to be eg the opposite of `collapsed`, but they're really describing different things). Maybe `visible`?

wagtail/admin/templates/wagtailadmin/admin_base.html
```diff
@@ -35,6 +35,7 @@
             wagtailConfig.LOCALES = {{ locales|safe }};
 
             wagtailConfig.STRINGS = {% js_translation_strings %};
+            wagtailConfig.CSRF_TOKEN = '{{ csrf_token }}'
```
jacobtoppm, line 38: Has this been included accidentally? I can't see its use here.

jacobtoppm (commented):
Some thoughts:
- Is it necessary for `open` to be separate from `collapsed`? A few scenarios.
  - I'm working solely on mobile, whether `open`==`collapsed` doesn't affect me
  - I'm working solely at desktop width, similarly unaffected
  - I'm resizing my screen. Now, if I have the opened/wide menu at mobile width and I expand again, the menu collapses, which feels weird. I feel like just using the `isMobile` and `collapsed` we could do away with `open` alltogether, but I may be missing some reason it's essential.
  
 Generally this is looking great though, and worked very well when I was testing

wagtail/admin/templates/wagtailadmin/admin_base.html
kaedroho: Oops, yes it has

client/src/components/Sidebar/Sidebar.tsx
kaedroho: `<aside>` is what we use for the existing sidebar. But perhaps we should be using `<nav>` for the main menu portion of it?

jacobtoppm: That feels right to me

client/src/components/Sidebar/Sidebar.tsx
kaedroho: Oh, good point! Yep I've renamed this to `visibleOnMobile`

kaedroho:
> Is it necessary for open to be separate from collapsed? A few scenarios. 

I think so, since ``collapsed`` is persisted in a cookie but I don't think we want to do this with open

jacobtoppm (approved):

PR #7485, opened 2021-08-31
Change initial value of publish_copies to False

jslay-excella:
When copying a live page, it is more likely that one would not like to publish on copy, therefore, this PR changes the initial value of `publish_copies`/"Publish copied page" to `False`. Opening PR for discussion and consideration. 

image

### DIFF
--- a/wagtail/admin/forms/pages.py
+++ b/wagtail/admin/forms/pages.py
@@ -53,7 +53,7 @@ class CopyForm(forms.Form):
                         pages_to_publish_count) % {'count': pages_to_publish_count}
 
                 self.fields['publish_copies'] = forms.BooleanField(
-                    required=False, initial=True, label=label, help_text=help_text
+                    required=False, initial=False, label=label, help_text=help_text
                 )
 
             # Note that only users who can publish in the new parent page can create an alias.


Scotchester:
I would agree that not wanting it to be immediately published seems likely to be the more common case, since there is no opportunity to change any of the existing content.

niqjohnson:
> I would agree that not wanting it to be immediately published seems likely to be the more common case, since there is no opportunity to change any of the existing content.

Agreed. This change came from a request from our content management lead to not publish copied pages by default. He said that in the "vast majority of cases" they need to edit the copied page's content before publishing, and having to uncheck the "published copied page" box every time was an extraneous step.

lb-:
I think this makes sense also, have put a question to the core team to get some additional feedback, we may need to consider the implications of this in terms of backwards compatibility (e.g. Wagtail users that expect this default).

lb-:
@jslay-excella thanks for thinking of this and taking the time to put up a PR, and congrats on your first commit to Wagtail.

This PR has been merged in by a separate commit 85faf192112271e71bc7e2afee0060e3bc52cd0e (it contains release notes /changelog entries).

Please check your name as listed in the contributors and let me know if you want it changed to something else.

Really appreciate your time here, thanks again.

PR #7486, opened 2021-08-31
workflow - only show active Tasks in the chooser modal

### LINKED ISSUE #7401: Hide disabled workflow tasks when adding tasks to a workflow
### Problem

This isn't really a big deal for me, but it's a little annoying: I've disabled many builtin workflow tasks and added a small number that we actually use. In the list of workflow tasks (/admin/workflows/tasks/index/), the disabled tasks are hidden unless I click "show disabled tasks," which is a setup that works well for me. However, when adding tasks to a workflow, the list of existing workflow tasks includes all of the disabled ones, and there are a lot of them to sort through. It also makes it feel inaccurate to describe those workflow tasks as "disabled;" it's more like "hidden from the main list."

### Desired solution

Hide disabled workflow tasks in the chooser when adding/editing tasks in a workflow. Could also add a button to show the disabled tasks, but it seems reasonable to me that one would need to re-enable a task in order for it to show there.

lb-:
* Implemented the desired solution (first part, simply not showing disabled tasks in the chooser modal), this is simpler to implement and makes more sense as adding a disabled task to a workflow will mean it always gets skipped.
* fixes #7401 
* When validating, be sure to also do a search in the chooser modal, the query code is different.
* Do the tests still pass? ✅ 
* Does the code comply with the style guide?  ✅ 
* For front-end changes: N/A

Screenshot of functionality working
Screen Shot 2021-09-01 at 7 29 10 am

### DIFF
--- a/wagtail/admin/tests/test_workflows.py
+++ b/wagtail/admin/tests/test_workflows.py
@@ -1532,7 +1532,7 @@ class TestDisableViews(TestCase, WagtailTestUtils):
         response = self.client.get(reverse('wagtailadmin_workflows:disable_task', args=(self.task_1.pk,)))
 
         self.assertTemplateUsed(response, "wagtailadmin/workflows/confirm_disable_task.html")
-        self.assertEqual(response.context['warning_message'], "This task is in progress on 1 page. Disabling this task will cause it to be skipped in the moderation workflow.")
+        self.assertEqual(response.context['warning_message'], "This task is in progress on 1 page. Disabling this task will cause it to be skipped in the moderation workflow and not be listed for selection when editing a workflow.")
 
         # create a new, unused, task and check the warning message is accurate
         unused_task = GroupApprovalTask.objects.create(name='unused_task_3')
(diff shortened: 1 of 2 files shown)


wagtail/admin/tests/test_workflows.py
```diff
@@ -1535,6 +1538,8 @@ def test_get(self):
         self.assertTemplateUsed(response, "wagtailadmin/workflows/task_chooser/includes/results.html")
         self.assertTemplateNotUsed(response, "wagtailadmin/workflows/task_chooser/includes/create_form.html")
         self.assertFalse(response.context['searchform'].is_searching())
+        # check that only active (non-disabled) tasks are listed
+        self.assertEqual([_.name for _ in response.context['tasks'].object_list], ['Enabled foo', 'Moderators approval'])
```
nmorduch, line 1542: I find it confusing to name this variable `_` since that makes me expect it will be ignored. Would it be clearer if we called it `task`?

nmorduch (changes requested):
This is great, just one small variable name that it would probably be helpful to change. Thanks @lb-!

wagtail/admin/tests/test_workflows.py
lb-: yep can do, bad habit from my Javascript life haha

nmorduch (approved):
Perfect, thank you!

nmorduch:
Closed in d6bb9bb5c0507e781060f3ca12696c136792934f

Issue #6810, opened 2021-02-14
dumpdata fails with "no such table: wagtailadmin_admin"


stefan-walluhn:
### Issue Summary

On wagtail 2.12, running `manage.py dumpdata` fails with `CommandError: Unable to serialize database: no such table: wagtailadmin_admin`, even on freshly installed wagtail site. `manage.py dumpdata` is needed for test fixture generation.

### Steps to Reproduce

1. Start a new project with `wagtail start myproject`
2. Run migrations `manage.py migrate`
3. Try to dump data `manage.py dumpdata`
4. `CommandError: Unable to serialize database: no such table: wagtailadmin_admin`

Running on a freshly installed wagtail from within a python 3.8.7 virtualenv. Dumpdata and fixture generation worked well on wagtail 2.11.3.

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: yes

### Technical details

```
./bin/pip freeze
anyascii==0.1.7
asgiref==3.3.1
beautifulsoup4==4.8.2
certifi==2020.12.5
chardet==4.0.0
Django==3.1.6
django-filter==2.4.0
django-modelcluster==5.1
django-taggit==1.3.0
django-treebeard==4.4
djangorestframework==3.12.2
draftjs-exporter==2.1.7
et-xmlfile==1.0.1
html5lib==1.1
idna==2.10
jdcal==1.4.1
l18n==2020.6.1
openpyxl==3.0.6
Pillow==8.1.0
pytz==2021.1
requests==2.25.1
six==1.15.0
soupsieve==2.2
sqlparse==0.4.1
tablib==3.0.0
urllib3==1.26.3
wagtail==2.12
webencodings==0.5.1
Willow==1.4
xlrd==2.0.1
XlsxWriter==1.3.7
xlwt==1.3.0
```

```
sqlite> .tables
auth_group                                  
auth_group_permissions                      
auth_permission                             
auth_user                                   
auth_user_groups                            
auth_user_user_permissions                  
django_admin_log                            
django_content_type                         
django_migrations                           
django_session                              
home_homepage                               
taggit_tag                                  
taggit_taggeditem                           
wagtailcore_collection                      
wagtailcore_collectionviewrestriction       
wagtailcore_collectionviewrestriction_groups
wagtailcore_groupapprovaltask               
wagtailcore_groupapprovaltask_groups        
wagtailcore_groupcollectionpermission       
wagtailcore_grouppagepermission             
wagtailcore_locale                          
[...]

PR #7488, opened 2021-09-01
Add "other permissions" to group permissions table

exler:
Resolves issue #5482 by replacing the "Other permissions" section with a new column in the "Object permissions" table.

Permissions not in the main categories (add, change, delete) will be placed in this column as an option in a multiple select box.


lb- (changes requested):
Hey @exler thanks for taking the time to submit this PR, code wise it is fine however user experience wise I think this may need some work.

## Feedback
Not sure on the best way forward but here is the feedback:
* The multi-select is a bit unintuitive as when you click something it goes grey (selected), but it is actually quite hard to see at a glance whether this grey is disabled/selected/unselected, especially when there are not many values listed.
* The text value says something like 'Can view Person' which makes sense when they were sitting in their own stand-alone list, however in the multi-select, this is quite verbose as you are already viewing the Person row.
* When even one permission exists, there is a large ~5 row select box shown.
* In general, I would be concerned that this UI will make it harder to manage permissions and not make mistakes, obviously it should be super clear what permissions are included or not.

## Suggestions
Again, no specific approach is recommended, just trying to get an understanding of what the problem is and how else we could solve this for CMS users and developers.

* It might be good to get a bit more clarity on what kind of problem is being solved here, do you have some mocks/screenshots of your scenarios (maybe with a really long permission per object list it is better with this new approach).
* What happens when there are zero other permissions for other objects, does the extra column still show? Maybe it can be removed if not applicable
* Instead of a multi-select, what about a multi-checkbox, this aligns with the other checkbox usage and is a lot harder to accidentally select/unselect something.
* What if this area was maybe just easier to customise, maybe including some `block` syntax so the template can be overridden in parts to allow for specific use cases to be presented.
* How do other UIs solve this, can we get some inspiration from them.

## Screenshots (reference)

The changes (Firefox) for the multi-select box look like this.

Screen Shot 2021-09-19 at 2 26 50 pm

The original looks like this.
Screenshot 2021-09-19 at 14-27-45 Wagtail - Editing Editors

exler:
Hello @lb-, thanks for the feedback!

I updated the PR to remove the model name from permission name to make it less verbose. I did not change the styles as I'm not quite sure what would be the proper design - there are libraries made with styled multiselect fields such as [Select2](https://select2.org/), but I'm not sure if it's worth adding just for such a simple feature. However, I don't think the default widget is confusing as there are only 2 states - selected/unselected.

The problem I'm solving here is the fact that permissions at this moment are unorganized. In the company I'm working we are using Wagtail and clients are constantly complaining that they have to search for the permissions where adding "Other permissions" as another column solves their problem. In the company fork we also added "View" as one of the primary columns (next to "Edit, Create, Delete"), but I decided not to add it due to [this comment](https://github.com/wagtail/wagtail/issues/5482#issuecomment-518201651). So the issue here is not a long list of permissions (which is also a problem because we frequently need to scroll through the whole page), but the ability to have all permissions for one model in one place.

I'm not sure how other UIs solve this as I'm not using any other CMS solutions, but I'm open to suggestions.

wagtail/users/templates/wagtailusers/groups/includes/formatted_permissions.html
```diff
@@ -38,30 +40,17 @@ <h2>{% trans "Object permissions" %}</h2>
...
                 </td>
+                <td>
+                    {% if content_perms_dict.other %}
+                    <select name="permissions" multiple>
+                        {% for other_perm in content_perms_dict.other %}
+                        <option value="{{ other_perm.perm.id }}" {% if other_perm.selected %}selected{% endif %}>{% trans other_perm.name %}</option>
+                        {% endfor %}
+                    </select>
+                    {% endif %}
+                </td>
```
lb-, line 51: I have not tested this but it might be better to use a multi-checkbox as below (as per discussion, I agree that using select2 is overkill)

```
                <td>
                    {% if content_perms_dict.other %}
                    <fieldset class="other-permissions">
                        <legend>{% trans "Other permissions" %}</legend> 
                        {% for other_perm in content_perms_dict.other %}     
                        <label class="other-permissions-item">
                            <input type="checkbox" name="permissions" value="{{ other_perm.perm.id }}" {% if other_perm.selected %}checked{% endif %}>
                            {{ other_perm.name }}
                        </label>
                        {% endfor %}
                    </fieldset>
                    {% endif %}
                </td>
```

Also, I do not think the `other_perm.name` needs translation, I would assume that would already be provided as a translated value.

The `legend` can be hidden but should be there for screen readers.

As noted above, the `td` would also need to be not rendered if not needed.

lb- (changes requested):
A few more suggestions, main one though is the potential to use a set of checkboxes instead of multi-select.

exler:
Thanks for the review @lb-! 

I have added the proposed changes. The checkboxes look better, however I am concerned whether this will be readable with long lists of permissions. With 2-3 checkboxes it looks fine, but there will be a lot of whitespace if there are more permissions. Below I added the screenshot of the new changes:

obraz

gasman:
Is there any way we can avoid the "Admin -> Can access Wagtail" permission being shown as an "object permission"? The fact that it's attached to a model is only an internal implementation detail - you can't define a permission without a content type, and a content type with no model creates other problems (see #6810).

exler:
Hey @gasman - sure we can. I implemented it just for this case, because I'm not sure if there will be other "other" permissions to put in the same section. This is how it looks now:

obraz

lb- (approved):

lb- (changes requested):
@exler this is looking heaps better, thank you for taking the time to refine this solution.

Are you able to flesh out some additional unit tests for this?
* custom permissions exist / not exist
* a group that has some custom permissions checked (to test checked state)
* the other permissions grouping (probably already covered in unit tests)

I know this may take a bit of trial and error but adding unit tests for these new features is the best way to ensure that they do not break in the future.

exler:
Hey @lb-, I've looked into writing some tests for this PR, but I have no idea how can I programmatically register permissions to show up in the Groups view. Normally I add the permissions to `@hooks.register('register_permissions')` in the app's `wagtail_hooks.py` so that they show up as custom permissions, but I cannot do that from the test suite. Can you point me in the right direction?

exler:
Thanks for the help @lb-! I was able to write some simple tests to cover the cases you mentioned. 

Is there anything else I should add?

wagtail/users/static_src/wagtailusers/scss/groups_edit.scss
```diff
@@ -16,6 +16,14 @@
         // stylelint-disable-next-line declaration-no-important
         font-size: 2.5em !important;
     }
+
+    .custom-permissions {
+        padding-bottom: 0;
```
lb-, line 21: The default bold feels a bit off here, maybe let's  make that look the same as other content.

```suggestion
        padding-bottom: 0;
        
        label {
            font-weight: inherit;
        }
```

lb- (approved):

wagtail/users/static_src/wagtailusers/scss/groups_edit.scss
```diff
@@ -16,6 +16,18 @@
...
+
```
lb-, line 22:
```suggestion

```

lb-:
@exler  sorry, could you please update your branch and fix the linting issue, I attempted to resolve in git but whitespace is causing issues.

This is looking really good, if all other tests pass we should be good to get this in soon.

@gasman any other thoughts on this one?

exler:
Sure thing @lb-, I updated the `.scss` file.

lb-:
@exler thanks heaps for your time on this, merged in via 8d42bec

PR #7491, opened 2021-09-02
Sidebar overlays to make page explorer usable on mobile

kaedroho:
This PR implements an overlays feature that allows the page explorer menu to display on top of everything if the screen size is too small to display it alongside

Screen Shot 2021-09-02 at 18 01 54

### DIFF
--- a/client/src/components/PageExplorer/PageExplorer.scss
+++ b/client/src/components/PageExplorer/PageExplorer.scss
@@ -8,8 +8,10 @@ $menu-footer-height: 50px;
 @import 'PageExplorerItem';
 
 .c-page-explorer {
-    width: 485px;
+    position: relative;
+    width: 100vw;
     height: 100vh;
+    max-width: 485px;
     background: $c-page-explorer-bg;
     overflow: hidden;
     flex: 1;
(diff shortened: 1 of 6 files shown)


nmorduch (changes requested):
This is an exciting step in a good direction! Some things I'm running into while trying it out:

- When I open the page explorer, it's confusing how it feels like the sidebar suddenly jumps to full width. Would it make sense to have the whole sidebar menu be 100vw wide instead of just the page explorer, and have a transition to the page explorer submenu? And do you have plans to extend this solution to the other submenus?
- It took me a while to find the button to close the page explorer. I kept trying to hit the `🏠 Pages` link and it was not taking me where I wanted. My first thought is to give it an ✖️ for closing the page explorer, which I would have in the same place as the one for closing the menu so it's easy to hit twice in a row if you want to close the whole menu. I would also probably change the text to something more verbose like  "Close page explorer" or "Back to menu", aligned left.
- Ignorable nitpick: The buttons for ✏️  and ➡️ have more padding than I think they need, so the page title can still end up rather squished on a small screen. (They're also different widths from each other, which I don't think they were previously.)
- Ignorable nitpick: I know this isn't a change you made, but the 🏠 icon is inconsistent with the root page having the 🌎 icon the rest of the time

And I made an issue for the slim sidebar not scrolling vertically, which isn't new in this PR: #7546

Note: this doesn't show in Squash because it requires the slim sidebar setting

thibaudcolas:
Completed in #7844.

PR #7493, opened 2021-09-03
Make hamburger menu a button for tab accessibility

### LINKED ISSUE #7327: Hamburger menu doesn’t work for keyboard and high-contrast mode users
The hamburger markup is not a link nor button, which breaks keyboard support, and styling for high-contrast mode users.

See for example with high-contrast mode – the button should be styled with an "active" color:

Current

---

The solution for this is to change the markup so this toggle is a `<button>` element.

amychan331:
Fixes #7327

      * Firefox, Chrome, and Safari in MacOS
      * Keyboard tabing


wagtail/admin/templates/wagtailadmin/base.html
```diff
@@ -54,7 +54,7 @@
             </div>
 
             {% if not slim_sidebar_enabled %}
-            <div id="nav-toggle" class="nav-toggle icon text-replace">{% trans "Menu" %}</div>
+            <button id="nav-toggle" class="nav-toggle icon text-replace">{% trans "Menu" %}</button>
```
zerolab, line 57: suggestion:

```suggestion
            <button type="button" id="nav-toggle" class="nav-toggle icon text-replace">{% trans "Menu" %}</button>
```

reference: https://github.com/wagtail/wagtail/issues/6592#issuecomment-735226755

amychan331: Thank you for pointing out the missing type! Commit suggestion applied

lb-:
Validated in Firefox (mac) and  Edge with NVDA, keyboard tabbing works great, focus outline visible and works in high contrast mode. I made one tweak to the border so that it shows an actual border again in windows high contrast.

I will rebase and update changelog/release notes on your remote branch and then merge in.

Thanks so much @amychan331 @zerolab  - a small change and a big accessibility improvement.

Note: The padding around the button on the right/bottom is tight against the icon, however after playing with this a bit I think this is the best way forward for now as we do not want to impact the size of the click target at the top left or increase the size at the bottom right I think.

**Validation screenshot**
Screen Shot 2021-09-19 at 2 06 50 pm

PR #7496, opened 2021-09-03
Document how to limit a page parent to root

### LINKED ISSUE #2768: Need ability to restrict parent page types to "root" page
I want the editor to be able to add a homepage as a child of the "root" page only. Currently there's no way to do that.

The workaround is to use a fixture (or migration) to make the homepage, then restrict its subpage and parent page types to `[]`. But that doesn't stop the editor from deleting the homepage. If they do, they can't add one back again.

The correct behavior should be to set `parent_page_types = ['root']` (maybe a special string) to restrict pages from only being able to be added to the root page.

tbrlpld:
Closes #2768

Issue #2768 was created because a way how to limit a page to be only
available under the root page was unknown.

The implementation has allowed this for a while now, but the issue was
not closed (presumably due to missing documentation).

The documentation of the `parent_page_types` filed now includes this
"special" case.

### DIFF
--- a/docs/reference/pages/model_reference.rst
+++ b/docs/reference/pages/model_reference.rst
@@ -256,6 +256,13 @@ In addition to the model fields provided, ``Page`` has many properties and metho
             class HiddenPage(Page):
                 parent_page_types = []
 
+        To allow for a page to be only created under the root page (e.g. for ``HomePage`` models) set the ``parent_page_type`` to ``['wagtailcore.Page']``.
+
+        .. code-block:: python
+
+            class HomePage(Page):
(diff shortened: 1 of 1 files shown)


brylie (approved):

tbrlpld:
Thank you for the review @brylie! I don't have access to merge the PR. Would be great if you could help me out there is you have a another moment 😅

brylie:
I don't have merge privileges. Perhaps @gasman or @kaedroho can take a look?

gasman (approved):
Looks good to me, thanks!

PR #7501, opened 2021-09-07
Fix typos on site settings docs

RealOrangeOne:
- the `.url` property is needed to get the actual page URL 
- Use correct path to page objects from settings (stray `.page_url`)
- Remove duplicate word

### DIFF
--- a/docs/reference/contrib/settings.rst
+++ b/docs/reference/contrib/settings.rst
@@ -308,7 +308,7 @@ Using the ``page_url`` shortcut has a few of advantages over using the tag:
     in more than one place (e.g. in a form and in footer navigation), using
     the ``page_url`` shortcut will be more efficient.
 3.  It's more concise, and the syntax is the same whether using it in templates
-    or views (or other Python code), allowing you to write more more consistent
+    or views (or other Python code), allowing you to write more consistent
     code.
 
 When using the ``page_url`` shortcut, there are a couple of points worth noting:


docs/reference/contrib/settings.rst
```diff
@@ -296,8 +296,8 @@ You could write:
 
 .. code-block:: html+django
 
-    {{ settings.app_label.ImportantPages.page_url.donate_page }}
-    {{ settings.app_label.ImportantPages.page_url.sign_up_page }}
+    {{ settings.app_label.ImportantPages.donate_page.url }}
+    {{ settings.app_label.ImportantPages.sign_up_page.url }}
 
 Using the ``page_url`` shortcut has a few of advantages over using the tag:
```
lb-, line 302: Should this line also be updated, along with the description underneath, it refers to the comparison between `page_url` the tag and the attribute/method on the `Page`

docs/reference/contrib/settings.rst
```diff
@@ -296,8 +296,8 @@ You could write:
 
 .. code-block:: html+django
 
-    {{ settings.app_label.ImportantPages.page_url.donate_page }}
-    {{ settings.app_label.ImportantPages.page_url.sign_up_page }}
+    {{ settings.app_label.ImportantPages.donate_page.url }}
+    {{ settings.app_label.ImportantPages.sign_up_page.url }}
```
lb-, line 300: Just confirming, I have not run this myself, but should this be `get_url`
https://docs.wagtail.io/en/stable/reference/pages/model_reference.html#wagtail.core.models.Page.get_url
?

lb- (changes requested):
Just checking a few things.

docs/reference/contrib/settings.rst
zerolab: drive-by comment: I believe `page_url` here relates to https://github.com/wagtail/wagtail/blob/main/wagtail/contrib/settings/models.py#L86 so the notation there is correct

lb-: ahhh yep, that makes sense.

zerolab: updated my previous comment to note that the original notation was correct. It also caches the results

lb- (approved):

lb-:
@RealOrangeOne  thanks for this, I have just pushed to your branch to remove the `     {{ settings.app_label.ImportantPages.donate_page.url }} ` change, and just kept the change to remove the duplicate 'more'.

Hope that is ok, please raise another PR if you think that the docs are actually broken for this example.

RealOrangeOne:
I did not know about the `.page_url` property on settings, interesting! Yeah in that case definitely fine with just the duplicate word change.

Might push another PR to better explain the existence of the `.page_url` property, but that's definitely OOS of this PR!

PR #7504, opened 2021-09-08
Admin URL finder mechanism

gasman:
A mechanism for obtaining the admin edit URL for an arbitrary model instance, optionally applying permission checks. 

    url_finder = AdminURLFinder(request.user)
    url_finder.get_edit_url(some_page)  # => "/admin/pages/123/edit/"
    url_finder.get_edit_url(some_image)  # => "/admin/images/456/"
    url_finder.get_edit_url(some_site)  # => None (user does not have edit permission for sites)

Handlers for new models can be registered via register_admin_url_finder:

    class SprocketAdminURLFinder(ModelAdminURLFinder):
        edit_url_name = 'wagtailsprockets:edit'

    register_admin_url_finder(Sprocket, SprocketAdminURLFinder)

I haven't added any documentation yet, as nothing is making use of AdminURLFinder yet (other than tests) - but the extended Site History report will make use of this, and it'll also be useful for things like "deleting this object will also delete the following..." prompts (#6129).

### DIFF
--- a/wagtail/admin/admin_url_finder.py
+++ b/wagtail/admin/admin_url_finder.py
@@ -1,0 +1,107 @@
+from django.contrib.admin.utils import quote
+from django.core.exceptions import ImproperlyConfigured
+from django.urls import reverse
+
+from wagtail.core.hooks import search_for_hooks
+
+
+"""
+A mechanism for finding the admin edit URL for an arbitrary object instance, optionally applying
+permission checks.
+
+    url_finder = AdminURLFinder(request.user)
+    url_finder.get_edit_url(some_page)  # => "/admin/pages/123/edit/"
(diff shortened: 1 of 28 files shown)


kaedroho (approved):
This looks good to me!

Do you think we could extend this to support other actions and use for the Admin API as well?

gasman:
> Do you think we could extend this to support other actions and use for the Admin API as well?

Sure! There's an open question there about how we handle different actions for the same model being registered in separate calls to register_admin_url_finder - e.g. wagtail.contrib.modeladmin will register itself as the edit view, but presumably won't provide an API endpoint so that will get registered separately - but we can cross that bridge when we come to it.

gasman:
Merged in de549b428863e8d1cd58a0c1ce1ed46af4c9f749

PR #7507, opened 2021-09-08
Fix userbar tabbing behaviour

### LINKED ISSUE #7290: Wagtail userbar should support tab navigation matching ARIA practices for menu
### Issue Summary

Follow-up to #6108 / #6994. Our user bar follows the ARIA menu [keyboard interactions](https://w3c.github.io/aria-practices/#keyboard-interaction-12), with the exception of Tab and Shift-Tab handling:

> Tab: Moves focus to the next element in the tab sequence, and if the item that had focus is not in a menubar, closes its menu and all open parent menu containers.
>
> Shift + Tab: Moves focus to the previous element in the tab sequence, and if the item that had focus is not in a menubar, closes its menu and all open parent menu containers.

### Steps to Reproduce

1. Go to a page with the userbar
2. Open the userbar
3. Tab or shift-tab

---

I think Shift-Tab should move focus back to the user bar trigger. For Tab, it should move focus to the next focusable element on the page after the trigger. If that’s not possible, we could also move focus back to the trigger. And in both cases, close the userbar list.

Stormheg:
Tabbing (navigation using Tab or Shift + Tab keys) will now close
the menu and move to the next focusable element on the page instead
of focusing on the next menu item.

The previous behaviour was a deviation from the ARIA menu practices:
https://w3c.github.io/aria-practices/#menu

Further changes/cleanup:

* Consume keyboard events like arrow down to prevent the browser from interpreting them. See: https://developer.mozilla.org/en-US/docs/Web/Accessibility/Keyboard-navigable_JavaScript_widgets#prevent_used_key_events_from_performing_browser_functions
* Refactor repeated setTimeout and `.focus()` calls into single
  `focusElement(el)` function. Let's keep it DRY!

Fixes #7290

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -43,6 +43,7 @@ Changelog
  * Fix: Disable Task confirmation now shows the correct value for quantity of tasks in progress (LB Johnston)
  * Fix: Page history now works correctly when it contains changes by a deleted user (Dan Braghis)
  * Fix: Add `gettext_lazy` to `ModelAdmin` built in view titles so that language settings are correctly used (Matt Westcott)
+ * Fix: Tabbing and keyboard interaction on the Wagtail userbar now aligns with ARIA best practices (Storm Heg)
 
 
 2.14.1 (12.08.2021)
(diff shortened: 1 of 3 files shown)


lb-:
@Stormheg works great! this is a much better interaction, thanks for taking the time to work on this.
I have pushed release notes / changelog and will get this merged in.

lb- (approved):

PR #7509, opened 2021-09-09
Allow specifying a format for the label of a collapsed StructBlock

gasman:
Currently, the label shown for a StructBlock's collapsed representation takes its content from the first sub-block of the StructBlock, which isn't always what you want. Add a new `label_format` meta option to StructBlock to allow customising this - e.g. `label_format = "Profile for {first_name} {surname}"`

Tested on bakerydemo by adjusting HeadingBlock in bakerydemo/base/blocks.py:

    class Meta:
        icon = "title"
        template = "blocks/heading_block.html"
        label_format = "{heading_text} (size {size})"

Screenshot 2021-09-09 at 22 45 25

(Prompted by a question on the #support Slack channel, which made me realise that customising this is currently a lot harder than it should be :-) )

### DIFF
--- a/client/src/components/StreamField/blocks/StructBlock.js
+++ b/client/src/components/StreamField/blocks/StructBlock.js
@@ -114,7 +114,23 @@ export class StructBlock {
   }
 
   getTextLabel(opts) {
-    /* Use the text label of the first child block to return one */
+    if (this.blockDef.meta.labelFormat) {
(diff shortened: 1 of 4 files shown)


kaedroho (approved):

gasman:
Merged in 0380f01b245615e5ec2a818e5e8dabf20250223f

PR #7514, opened 2021-09-12
Add the ability to pass a config object to register_snippet & configure per page count on listing view

### LINKED ISSUE #7399: Make number of snippets shown in list view adjustable via `SnippetViewSet`

It would be great to be able to adjust the number of snippets shown in the list page. Sometimes being able to see ALL the snippets is helpful. 

I think adjusting the code here:
https://github.com/wagtail/wagtail/blob/b8a86ac15de331654024f3b4590088f0b88626fc/wagtail/snippets/views/snippets.py#L145

To accept a setting would make this work. If the setting is not available we could default to 20

@gasman has made a better suggestion below. Stating that this could be added to the model similarly to how ModelAdmin handles it with a config object.  

I thought about overwriting the view myself, but there's a lot there, and I think this could be better served through a setting in settings.py

lb-:
## Implementation of the following
* `register_snippet` now allows for an optional `config` class, and a default `DefaultSnippetConfig` will be used when no config is supplied.
* The config contains one attribute `list_per_page` (same name as what ModelAdmin uses) which will determine the per_page value in the Snippet's list view.
* `register_snippet` when called but then used as a decorator (e.g. `@register_snippet()`) will now work, this would be an accidental usage but it is more likely now that we are adding a config arg.

## Notes
* Inspiration for the code implementation of this decorator/function usage is from https://github.com/django/django/blob/main/django/template/library.py#L54
- resolves #7399
* Do the tests still pass? 👍 
* Does the code comply with the style guide?  👍 
* For Python changes: Have you added tests to cover the new/fixed behaviour?  👍 
* For new features: Has the documentation been updated accordingly?  👍


lb-:
Rebased on main

lb-:
Note to self. Rebase on main after Black.

lb-:
ok, having a fair bit of trouble with this with the module rework, gonna put back to needs work and pick up when I have a bit more time.

lb-:
Closing this for now - there are a lot of changes happening to Snippets for now and this may not be the direction we want to head in.

@kaedroho feel free to pull in parts of this if it is useful.

PR #7520, opened 2021-09-15
Use gettext_lazy in modeladmin views where required

gasman:
Class-level strings were using gettext, which meant that they were not respecting user language settings.


lb-:
@gasman  looks great, I tested the behaviour before and after and this change works well.
Thanks heaps, I have pushed release notes to your branch and will rebase/merge

PR #7523, opened 2021-09-17
Explicitly use port 80 for squash.io

Stormheg:
Hopefully prevents scenarios where Squash prompts for a port number and a
wrong port number is entered. This leads to the Squash deployment never finishing and eventually timing out.


Scotchester (approved):
It works! Thanks, @Stormheg.

PR #7524, opened 2021-09-17
Fix page history with delete user

zerolab:
From [Slack message in the #support channel](https://wagtailcms.slack.com/archives/C81FGJR2S/p1631891808385600):

> We're currently working on a site and have found that if a user has edited a page, and then the user is deleted, then we can no longer view the pages history. Rather we get the following exception 'NoneType' object has no attribute 'get_username' .

Upon checking it looks like `self._meta.get_field('user').related_model.DoesNotExist` doesn't get raised, so changed the approach and added tests for page and site history.

Of note is the change of the user id placeholder to a string one to accomodate for non-integer IDs


lb-:
@zerolab thanks for this, saw that comment on Slack, nice turn around :) 
Tested the behaviour before and after the fix, works great!

I pushed the release notes/changelog to your branch and have merged in

zerolab:
@lb- thank you for the swift review!

gasman:
Marking as 2.14.2, as this looks like a worthwhile candidate for a backport.

Issue #7366, opened 2021-07-23
Page editing actions dropdown doesn’t support keyboard or screen reader navigation


thibaudcolas:
### Issue Summary

Initially reported by @jessemenn. On the page editing UI, the actions dropdown is very brittle with keyboard and screen reader navigation:

- The dropdown’s toggle button isn’t focusable
- Items within the dropdown are focusable even when hidden.

### Steps to Reproduce

1. Go to the page editing UI on any Wagtail site
2. Reach the "Save draft" button
3. Try to reach any action within the dropdown

---

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: yes

### Technical details

Tested in Chrome macOS & Safari VoiceOver. I’d expect this to be an issue for all browsers, all screen readers, in all Wagtail releases over the last few years at least.

thibaudcolas:
Based on reviewing available resources, it seems there are two documented ARIA patterns we could use for this:

- Progressive disclosure, similar to https://www.w3.org/TR/wai-aria-practices-1.1/examples/disclosure/disclosure-navigation.html
- Or the menubar & menu pattern: https://www.w3.org/TR/wai-aria-practices-1.1/examples/menubar/menubar-1/menubar-1.html

In both cases there are many more keyboard interactions to support than my initial issue description suggests.

thibaudcolas:
As part of refactoring this we should also make sure to consistently use the same element (button or link) for all menu items, so screen reader announcements and high-contrast mode display is more consistent:

![7366](https://user-images.githubusercontent.com/877585/129975685-4b93eb30-903f-409a-8364-fa5c579b3367.png)

Desaiakshata:
I can work on this issue

lb-:
Some improvement has been made to this, thanks to @Desaiakshata  https://github.com/wagtail/wagtail/commit/997ac738b0bd04db0e737a11fa6304586865eab5

thibaudcolas:
@thibaudcolas to re-test

PreciousArinda:
Hello @thibaudcolas  I'm seeing this issue is to be re-tested. 
I believe I can contribute to this issue.
Can I take it on?

ArijeetBanerjee:
I want to work on this issue. kindly give me permission and access to work on this issue

PR #7525, opened 2021-09-19
Keyboard accessibility: Edit page actions dropdown

Desaiakshata:
Tested on: Windows 10
Microsoft Edge: latest version
Google Chrome: Latest version

This is a fix for issue https://github.com/wagtail/wagtail/issues/7366

Added keyboard navigation to the edit page actions dropdown. 
Fixed windows high contrast mode display for the dropdown.


client/scss/components/_dropdown.legacy.scss
```diff
@@ -123,6 +123,29 @@
 
         li {
             border-width: 0 0 1px;
+
+            a:focus,
+            button:focus {
+                border: 3px solid #ffbf47;
```
lb-, line 129: It would be good to use the scss variable `$color-focus-outline` here.
(1 styles/focus)

client/scss/components/_dropdown.legacy.scss
```diff
@@ -123,6 +123,29 @@
...
+            }
+        }
+
+        @media(forced-colors:$media-forced-colours) {
```
lb-, line 133: to make it easier to find these kinds of changes in the code and also explain 'why', it would be good to change this to 

```
// Media for Windows High Contrast

@media (forced-colors: $media-forced-colours) {
```
So that there is a comment above the change and also a space between the colon & before the bracket.

(1 styles/focus)

wagtail/admin/templates/wagtailadmin/pages/edit.html
```diff
@@ -57,7 +57,7 @@ <h1 class="left col9 header-title" title="{% blocktrans with title=page.get_admi
             <ul>
                 <li class="footer__container">
                     <nav aria-label="{% trans 'Actions' %}">
-                        <ul>
+                        <ul id="exTest" class="disclosure-nav">
```
lb-, line 60: I might recommend instead of grabbing the `ul` it might be better to base the JS work on the `div.dropdown` as this appears to be the more generic usage of this kind of thing, and is will mean that the JS can work for more cases, not just when it is in a `ul`.

(2 js/keyboard)

client/scss/components/_dropdown.legacy.scss
```diff
@@ -123,6 +123,29 @@
 
         li {
             border-width: 0 0 1px;
+
```
lb-, line 126: Putting the styles here means they will only work with the dropup variant, not the dropdown variant.

You may be able to find a spot higher up or elsewhere in this stylesheet that works for both.

You can test the dropdown usage in the reporting section of Wagtail.

(1 focus/styles)

wagtail/admin/templates/wagtailadmin/pages/action_menu/menu.html
```diff
@@ -3,7 +3,7 @@
     {{ default_menu_item }}
 {% endif %}
 {% if show_menu %}
-    <div class="dropdown-toggle">{% icon name="arrow-up" %}</div>
+    <div class="dropdown-toggle" aria-expanded="true" aria-controls="other-actions" tabindex="0">{% icon name="arrow-up" %}</div>
```
lb-, line 6: I think an initial step to this would be to convert the dropdown-toggle to a `button` (remember to add `type="button"` on it also), then ensure that the styling works, and this would need to be done in all usage (about~30 in the current code).

This should solve the issue of this click target being focusable and avoids having to add JS work arounds for 'act like a button but is not really a button' issues.

That would be a recommended step 1 (focus/styles) and then the step 2 part can add the relevant aria props, currently there is nothing with an id `other-actions` that I can see so this aria-controls may not do what we want it to do.

However, let's try to solve the conversion a to an actual button first (for all usage).

wagtail/admin/templates/wagtailadmin/pages/edit.html
lb-: Also, best to remove this id, it does not appear to be part of the implementation.

client/src/entrypoints/admin/disclosureMenu.js
```diff
@@ -0,0 +1,168 @@
+// JavaScript source code
```
lb-, line 1: There are other notes in the main feedback about this JS, however it looks like we already have a 'home' for the dropdown code.

see client/src/entrypoints/admin/core.js
about line 341... `$('.dropdown').each(function () {`

It would be good to try to update that code  in place, instead of having extra code that works on once instance of the dropdown.

I know this increases the 'scope' of this work but well worth a bit of a refactor here.
Plus - as noted, similar code could be leveraged that was built for https://github.com/wagtail/wagtail/pull/7507

lb- (changes requested):
@Desaiakshata  thank you so much for taking the time to submit this PR.

After having a bit of a look at this with another team member I think it might be best to approach this a bit differently. I hope this feedback is helpful, our goal is to get incremental improvements into the code but we need to ensure they all work as expected.

Firstly, I recommend that this PR be either closed or reworked to cover the focus/styling aspects only. I have more context about this below & in individual comments, but even with that change, there are lots of cases where this dropdown menu is used and we want to ensure they are all covered - not just the edit page menu if possible.

Secondly, the keyboard interaction JS looks like code from an external resource, I know the copyright disclaimer is there but that opens up a bit of an unknown box of issues from an open source perspective. However, a very similar interaction has been recently built via https://github.com/wagtail/wagtail/pull/7507 (see the changes there). Maybe this could be pulled out into its own file and reused elsewhere for these drop-down buttons OR be coped from that source into the admin/core.js file instead of from code that is external..

This keyboard interaction work could be a second PR and might be easier to review in isolation, plus it means we can get the easier one out of the way and into the code base.

As for specific feedback, I have included some per line recommendations along with a note about which PR work the change could be part of.

In general, I know the original bug says page edit menu, but it would be good to consider the other cases of this `dropdown` component throughout the UI.

Please let us know if you would be ok to work through these changes.

@thibaudcolas is away at the moment, but when he is back on deck he might also have some other input.

@Stormheg  feel free to add any thoughts also.

lb-:
Hey @Desaiakshata 
I found this article which also might be helpful to implement this solution https://web.dev/building-a-split-button-component/ 

Also, the comment about moving the code to `admin/core.js` should be revised, after chatting with some of the core team, we are actually hoping to move away from that so ideally the code for the dropdown that exists there would be pulled out into its own file.

I know there is a lot to take in for changes requested on this PR but the best way to move forward would be to solve smaller parts at a time on separate PRs (as per original comment, maybe let's start by adjusting styling first).

Desaiakshata:
> Hey @Desaiakshata
> I found this article which also might be helpful to implement this solution https://web.dev/building-a-split-button-component/
> 
> Also, the comment about moving the code to `admin/core.js` should be revised, after chatting with some of the core team, we are actually hoping to move away from that so ideally the code for the dropdown that exists there would be pulled out into its own file.
> 
> I know there is a lot to take in for changes requested on this PR but the best way to move forward would be to solve smaller parts at a time on separate PRs (as per original comment, maybe let's start by adjusting styling first).

Sorry for following up quite late. I have updated the changes requested for styling.

wagtail/admin/templates/wagtailadmin/pages/action_menu/menu.html
Desaiakshata: Could you let me know where else (some instances) the dropdown toggle is used (as mentioned around ~30). As the original issue focuses on the edit page, I have worked to make changes only to that page. But I would like to follow up on the others as well.

Desaiakshata:
I can work on creating a separate file for the dropdown js part and will follow up for issues

lb-:
@Desaiakshata did you want to try to solve the JS & CSS items in one PR or were you happy to split these out?

If you split the work it might be easier to assess the PR.

To do this though you will need to create a new branch on your code (using git), or you could just manually copy the JS work to a different folder, then remove the JS changes from this branch you are working on and push the changes up.

Let me know if you need a hand working through this, thanks for taking the time to review and give it another go.

thibaudcolas:
I believe this has been replaced by #7650.

PR #7528, opened 2021-09-20
Allow wildcards in redirect paths

RealOrangeOne:
Currently, it's not possible to specify redirects with wildcards in them, which allow matching anything under a path. There's definitely an argument for this being handled by something further downstream (reverse proxy, CDN etc).

I've intentionally tried to break this in terms of SQLi, but couldn't find anything, which is good! Using percent symbols may be strange to some, but we could rewrite these if necessary. I intentionally didn't use full regexes, as the possibility for DoS based on that is far too high, not to mention them being completely overkill.

Still very much in progress, but opening as a draft to gauge opinions. Currently it's just the data model and tests, as opposed to the middleware integration. 

- If a URL matches multiple patterns, which should it redirect to? And how do sites play into this?


lb-:
@RealOrangeOne would you be up for closing this PR (as it's an older draft and has need for a rebase), instead maybe open an issue that raises the concerns and references this PR as a starting point.

This way it may be a better place to gauge opinions and feedback from the community and core team.

PR #7529, opened 2021-09-20
Introduce LogFormatter class for formatting 

gasman:
An API improvement as part of model audit logging - rather than the log action registry keeping track of several lookups for log action labels, messages and comments, define a LogFormatter object for each action type. Code that uses log messages (e.g. reports) can now access the `message` and `comment` properties on the LogEntry object, rather than having to work with the registry.

Note: I haven't preserved undocumented behaviour, specifically the `comments` kwarg on register_log_action and the `takes_log_entry` flag.

### DIFF
--- a/docs/reference/hooks.rst
+++ b/docs/reference/hooks.rst
@@ -1379,8 +1379,27 @@ Audit log
         def additional_log_actions(actions):
             actions.register_action('wagtail_package.echo', _('Echo'), _('Sent an echo'))
 
-            def callback_message(data):
-                return _('Hello %(audience)s') % {
-                    'audience': data['audience'],
-                }
-            actions.register_action('wagtail_package.with_callback', _('Callback'), callback_message)
+
+    Alternatively, for a log message that varies according to the log entry's data, create a subclass of ``wagtail.core.log_actions.LogFormatter`` that overrides the ``format_message`` method, and use ``register_action`` as a decorator on that class:
+
+    .. code-block:: python
+
+        from django.utils.translation import gettext_lazy as _
+
+        from wagtail.core import hooks
+        from wagtail.core.log_actions import LogFormatter
+
+        @hooks.register('register_log_actions')
+        def additional_log_actions(actions):
+            @actions.register_action('wagtail_package.greet_audience')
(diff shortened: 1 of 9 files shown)


docs/reference/hooks.rst
```diff
@@ -1379,8 +1379,27 @@ Audit log
...
+    .. code-block:: python
+
+        from django.utils.translation import gettext_lazy as _
+
+        from wagtail.core import hooks
+        from wagtail.core.log_actions import LogFormatter
+
+        @hooks.register('register_log_actions')
+        def additional_log_actions(actions):
+            @actions.register_action('wagtail_package.greet_audience')
```
kaedroho, line 1394: Any reason to not allow this at the module level? This would be consistent with how [modeladmin classes are registered](https://docs.wagtail.io/en/stable/reference/contrib/modeladmin/index.html#a-more-complicated-example)

docs/reference/hooks.rst
```diff
@@ -1379,8 +1379,27 @@ Audit log
...
+                label = _('Greet audience')
+
+                def format_message(self, log_entry):
+                    return _('Hello %(audience)s') % {
+                        'audience': log_entry.data['audience'],
+                    }
+
+    .. versionchanged:: 2.15
+
+      The LogFormatter class was introduced. Previously, dynamic messages were achieved by passing a callable as the ``message`` argument to ``register_action``.
```
kaedroho, line 1405: Missing backticks around ``LogFormatter``

kaedroho (approved):

docs/reference/hooks.rst
gasman: No reason that I can see. This is the "eliminate hooks that just register something on startup" task that I put forward for the forthcoming Python refactoring sprint, and `register_log_actions` is on my hit list for that :-)

(you may have noticed that I pointedly avoided creating a new hook for admin URL finders in #7504, in favour of just calling the register function at the module level - so if that doesn't blow up, it should serve as a testing-the-water exercise in removing them more widely!)

gasman:
Merged in 3b972775fc63fd1dd0e8f30e65a6096a64e6c727

Issue #7328, opened 2021-07-09
High-contrast mode: Breadcrumb separator should match link text colour


thibaudcolas:
This is a bug specific to [Windows high-contrast mode](https://blogs.windows.com/msedgedev/2020/09/17/styling-for-windows-high-contrast-with-new-standards-for-forced-colors/). cc @kbayliss who did the testing for this.

---

Current: `color: #00585a`
Proposed: `color: LinkText`

(will also need hover/active styles)

Current:

![Current](https://user-images.githubusercontent.com/877585/125083195-fb095400-e0bf-11eb-8b81-f7da400ac5ff.png)

Proposed:

![Proposed](https://user-images.githubusercontent.com/877585/125083203-fcd31780-e0bf-11eb-905a-06145005df92.png)

thibaudcolas:
Marking as a good first issue – I think the proposed fix could work well with the [forced-colors](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors) media query, without expert knowledge of Wagtail required. Note testing this does require having access to Windows high-contrast mode.

surajaiswal13:
Hello Sir, Can I work on this Issue, This could be my Open Source Contribution

surajaiswal13:
Sir " Can you Suggest something what can i do for hover/active styles " 
I have Implemented 

    .arrow_right_icon {
        @include svg-icon(1em);
        color: LinkText;
        transform: scale(1.75) translate(0.3em, 0.1em);
    }

also it is working fine in high contrast mode When normal it is in Yellow color and when hovered it gets black in color

lb-:
This has been fixed via fb66c1fd1732f462a1983edb5410192a9b8963da
<img width="1520" alt="Screen Shot 2021-09-30 at 7 46 58 pm" src="https://user-images.githubusercontent.com/1396140/135429446-a3177091-43a8-4a46-81d5-94f805ad9d05.png">

Issue #7453, opened 2021-08-18
High-contrast mode: active tab can’t be determined


thibaudcolas:
This is a bug specific to [Windows high-contrast mode](https://blogs.windows.com/msedgedev/2020/09/17/styling-for-windows-high-contrast-with-new-standards-for-forced-colors/). cc @kbayliss who did the testing for this.

In layouts using tabs, it’s impossible to tell which tab the user is currently in:

![7453](https://user-images.githubusercontent.com/877585/129971971-ce3d5a1c-155e-49f5-b84c-b95bad1349b1.png)

PR #7532, opened 2021-09-22
Fixed High-contrast mode: Breadcrumb separator should match link text…

surajaiswal13:
Fixed issue #7328 According to the Instructions


lb-:
@surajaiswal13 thanks so much for this, however it looks like the issue of the breadcrumbs 'arrow' and the home icon was already fixed via fb66c1fd1732f462a1983edb5410192a9b8963da 

Feel free to pick up another issue though, may I suggest #7453 -> i can confirm that is not yet fixed.

PR #7535, opened 2021-09-23
update documentation & 2.15 release to drop IE11 support

### LINKED ISSUE #6170: Drop support for IE11

lb-:
Our docs previously stated that IE11 browser support would be dropped in 2.14, however we never added that to the release notes.

We want to ensure that this gets put into 2.15 as it is an LTS.

Beyond docs changes & `browserslist` there are no actual JS changes done in this PR, I think it would be best to pull out IE11 specific work-arounds on a case by case basis as other code is cleaned up and worked on.

- fixes #6170
- note for any backport fixes - v2.11 LTS, supports IE11, and will be supported until February 2022
- as of August 2021, IE11 usage is at 0.58% https://gs.statcounter.com/browser-market-share

### DIFF
--- a/docs/contributing/developing.rst
+++ b/docs/contributing/developing.rst
@@ -183,7 +183,6 @@ Browser        Device/OS      Version(s)
 Mobile Safari  iOS Phone      Last 2
 Mobile Safari  iOS Tablet     Last 2
 Chrome         Android        Last 2
-IE             Desktop        11
 Chrome         Desktop        Last 2
 MS Edge        Desktop        Last 2
 Firefox        Desktop        Latest
@@ -191,9 +190,9 @@ Firefox ESR    Desktop        Latest
 Safari         macOS          Last 2
 =============  =============  =============
 
-We aim for Wagtail to work in those environments. Our development standards ensure that the site is usable on other browsers **and will work on future browsers**. To test on IE, install virtual machines `made available by Microsoft <https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/>`_.
+We aim for Wagtail to work in those environments. Our development standards ensure that the site is usable on other browsers **and will work on future browsers**.
 
-IE 11 is gradually falling out of use, and specific features are unsupported in this browser:
+IE 11 support has been officially dropped in 2.15 as it is gradually falling out of use. Features already known not to work include:
 
 * Rich text copy-paste in the rich text editor.
 * Sticky toolbar in the rich text editor.
@@ -206,7 +205,7 @@ IE 11 is gradually falling out of use, and specific features are unsupported in 
 Browser        Device/OS      Version(s)
 =============  =============  =============
 Stock browser  Android        All
-IE             Desktop        10 and below
+IE             Desktop        All
 Safari         Windows        All
 =============  =============  =============
 
--- a/wagtail/admin/templates/wagtailadmin/skeleton.html
+++ b/wagtail/admin/templates/wagtailadmin/skeleton.html
@@ -84,7 +84,7 @@
     {% block ie11warning %}
         <div data-ie11-warning hidden>
             <div class="capabilitymessage">
-                {% blocktrans with url="https://docs.wagtail.io/en/stable/editor_manual/browser_issues.html#ie11" %}Wagtail is gradually removing support for IE11, which may affect you or your organisation. <a href="{{ url }}">Learn more about our IE11 support plans</a>.{% endblocktrans %}
+                {% blocktrans with url="https://docs.wagtail.io/en/stable/editor_manual/browser_issues.html#ie11" %}Wagtail has removed support for IE11. <a href="{{ url }}">Learn more about IE11 alternatives</a>.{% endblocktrans %}
             </div>
         </div>
     {% endblock %}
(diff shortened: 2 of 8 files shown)


docs/contributing/developing.rst
```diff
@@ -183,17 +183,16 @@ Browser        Device/OS      Version(s)
...
 Firefox        Desktop        Latest
 Firefox ESR    Desktop        Latest
 Safari         macOS          Last 2
 =============  =============  =============
 
-We aim for Wagtail to work in those environments. Our development standards ensure that the site is usable on other browsers **and will work on future browsers**. To test on IE, install virtual machines `made available by Microsoft <https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/>`_.
+We aim for Wagtail to work in those environments. Our development standards ensure that the site is usable on other browsers **and will work on future browsers**.
 
-IE 11 is gradually falling out of use, and specific features are unsupported in this browser:
+IE 11 support has officially dropped in 2.15 as it is gradually falling out of use, many specific features are completely unsupported in this browser:
```
gasman, line 195: has officially dropped -> has been officially dropped

I'd be inclined to remove the list of things that don't work - it creates an expectation that things _not_ on the list ought to still work in IE11, and we're no longer promising that (and we're likely to break a lot more once we start removing compatibility shims). Or at least, we should say "features known not to work include..." - saying "unsupported" is understating things.

wagtail/admin/templates/wagtailadmin/skeleton.html
```diff
@@ -84,7 +84,7 @@
     {% block ie11warning %}
         <div data-ie11-warning hidden>
             <div class="capabilitymessage">
-                {% blocktrans with url="https://docs.wagtail.io/en/stable/editor_manual/browser_issues.html#ie11" %}Wagtail is gradually removing support for IE11, which may affect you or your organisation. <a href="{{ url }}">Learn more about our IE11 support plans</a>.{% endblocktrans %}
+                {% blocktrans with url="https://docs.wagtail.io/en/stable/editor_manual/browser_issues.html#ie11" %}Wagtail is has removed support for IE11, <a href="{{ url }}">learn more about IE11 alternatives</a>.{% endblocktrans %}
```
gasman, line 87:
```suggestion
                {% blocktrans with url="https://docs.wagtail.io/en/stable/editor_manual/browser_issues.html#ie11" %}Wagtail has removed support for IE11. <a href="{{ url }}">Learn more about IE11 alternatives</a>.{% endblocktrans %}
```

gasman (changes requested):

docs/contributing/developing.rst
lb-: Done, thanks.

wagtail/admin/templates/wagtailadmin/skeleton.html
lb-: Done thanks.

gasman (approved):

PR #7536, opened 2021-09-23
Enable basic js linting other folders

lb-:
* Partial work towards #2797
* The big benefit of this is that our linter will pick up the things that really could break code (shadow variables, triple vs double equal) with minimal diffs that might add friction to existing work in progress.
* Enable linting globally but disable a bunch of rules for wagtail folder and docs folder (plus add globals)
* Still some work to do but it means we can pick a rule one by one and disable them until we get to the big one (formatting)
* Also, with IE11 support dropped in 2.15 we can use the arrow function rule


lb-:
Rebased on main

thibaudcolas (approved):
Thank you @lb- 🤘 this looks spot-on to me. I couldn’t tell which of those changes were from `--fix` and which you did manually, but either way all of those look completely safe.

lb-:
@thibaudcolas thanks for reviewing.

PR #7538, opened 2021-09-24
Add missing bind_to call to PreviewOnEdit, fixes #7537

### LINKED ISSUE #7537: Preview error when using StreamField with EditHandler on per-request basis
### Issue Summary

We're using `EditHandler`s, which show/hide fields depending on the user's permission. This is based on the example in wagtail's tests (see https://github.com/wagtail/wagtail/blob/main/wagtail/tests/testapp/models.py#L1418).

**Problem:**
Unfortunately, there are multiple places in wagtail's codebase, where `BaseFormEditHandler.get_form_class()` is called, without binding the current request (and also instance etc). One example is the preview-mode (see [preview.py](https://github.com/wagtail/wagtail/blob/main/wagtail/admin/views/pages/preview.py#L51)).

As a result, a user which can only view a subset of the full admin form of a page for example, actually validates his POST-data (which only contains a subset of the fields) with the full form, since the logic that would hide some fields based on the request isn't called.

This probably works for simple cases like [wagtail's test](https://github.com/wagtail/wagtail/blob/main/wagtail/tests/testapp/models.py#L1442) and described in multiple tickets (e.g. https://github.com/wagtail/wagtail/issues/5546), but in our case it results in crashes when showing/hiding `StreamField`s.

**Solution:**
At least bind `instance` and `request` before `get_form_class()` is called, where possible, similar to how the normal edit-view does it: https://github.com/wagtail/wagtail/blob/main/wagtail/admin/views/pages/edit.py#L311

I will try to add those calls to all the places I can find and let the tests run and I hope this can be fixed quickly.

### Steps to Reproduce

I don't have a simple example available yet, but basically we have a custom `BaseCompositeEditHandler` which may skip some of its children, if the user is not a superuser. This is very similar to the [sample](https://github.com/wagtail/wagtail/blob/main/wagtail/tests/testapp/models.py#L1393) in wagtail's tests, but instead of switching between two panel-lists, we have one default list, and based on the field-keys we may exclude some of those fields.

So this all works for the superuser, since he can edit all of the fields. But if one of those excluded fields is a `StreamField` and the non-superuser clicks on the preview button, the server crashes with something like that:

    2021-09-24 11:43:13 ERROR django.request: Internal Server Error: /admin/content/pages/add/base/testpage/1337/preview/ (log.py:224)
      File "/****/lib/python3.8/site-packages/django/utils/datastructures.py", line 76, in __getitem__
        list_ = super().__getitem__(key)
    KeyError: 'form_element_blocks-count'

    ...****...
      File "/****/lib/python3.8/site-packages/django/views/decorators/cache.py", line 44, in _wrapped_view_func
      File "/****/lib/python3.8/site-packages/wagtail/admin/urls/__init__.py", line 121, in wrapper
      File "/****/lib/python3.8/site-packages/wagtail/admin/auth.py", line 193, in decorated_view
      File "/****/lib/python3.8/site-packages/django/views/generic/base.py", line 70, in view
        return self.dispatch(request, *args, **kwargs)
      File "/****/lib/python3.8/site-packages/django/views/generic/base.py", line 98, in dispatch
        return handler(request, *args, **kwargs)
      File "/****/lib/python3.8/site-packages/wagtail/admin/views/pages/preview.py", line 64, in post
        form = self.get_form(self.get_page(), request.POST)
      File "/****/lib/python3.8/site-packages/wagtail/admin/views/pages/preview.py", line 129, in get_form
        if form.is_valid():
      File "/****/lib/python3.8/site-packages/wagtail/admin/forms/pages.py", line 150, in is_valid
        return super().is_valid()
      File "/****/lib/python3.8/site-packages/modelcluster/forms.py", line 316, in is_valid
        form_is_valid = super(ClusterForm, self).is_valid()
      File "/****/lib/python3.8/site-packages/django/forms/forms.py", line 175, in is_valid
        return self.is_bound and not self.errors
      File "/****/lib/python3.8/site-packages/django/forms/forms.py", line 170, in errors
        self.full_clean()
      File "/****/lib/python3.8/site-packages/django/forms/forms.py", line 372, in full_clean
        self._clean_fields()
      File "/****/lib/python3.8/site-packages/django/forms/forms.py", line 384, in _clean_fields
        value = field.widget.value_from_datadict(self.data, self.files, self.add_prefix(name))
      File "/****/lib/python3.8/site-packages/wagtail/core/blocks/base.py", line 535, in value_from_datadict
        return self.block_def.value_from_datadict(data, files, name)
      File "/****/lib/python3.8/site-packages/wagtail/core/blocks/stream_block.py", line 97, in value_from_datadict
        count = int(data['%s-count' % prefix])
      File "/****/lib/python3.8/site-packages/django/utils/datastructures.py", line 78, in __getitem__
        raise MultiValueDictKeyError(key)
    django.utils.datastructures.MultiValueDictKeyError: 'form_element_blocks-count'
    2021-09-24 11:43:13 ERROR django.channels.server: HTTP POST /admin/content/pages/add/base/testpage/1337/preview/ 500 [0.15, 127.0.0.1:44074] (runserver.py:160)

In the above stacktrace, the `form_element_blocks` (which is a `StreamField`) are not visible to the non-superuser user in the frontend. But the preview-request actually creates a form with the `StreamField` (because our code in `on_request_bound()` isn't called), which fails when validating against the POST-data, which doesn't contain any of the `StreamField`'s POST-data.

### Technical details

* Python version: 3.8.10
* Django version: 3.2.7
* Wagtail version: 2.14.1

th3hamm0r:
see #7537

### DIFF
--- a/wagtail/admin/tests/pages/test_preview.py
+++ b/wagtail/admin/tests/pages/test_preview.py
@@ -1,13 +1,17 @@
 import datetime
 
+from functools import wraps
+from unittest import mock
+
 from django.test import TestCase
 from django.urls import reverse
 from django.utils import timezone
 from freezegun import freeze_time
 
+from wagtail.admin.edit_handlers import FieldPanel, ObjectList, TabbedInterface
 from wagtail.admin.views.pages.preview import PreviewOnEdit
 from wagtail.core.models import Page
-from wagtail.tests.testapp.models import EventCategory, SimplePage, StreamPage
+from wagtail.tests.testapp.models import EventCategory, EventPage, SimplePage, StreamPage
 from wagtail.tests.utils import WagtailTestUtils
 
 
@@ -55,6 +59,21 @@ class TestIssue2599(TestCase, WagtailTestUtils):
         self.assertEqual(response.context['self'].depth, homepage.depth + 1)
         self.assertTrue(response.context['self'].path.startswith(homepage.path))
         self.assertEqual(response.context['self'].get_parent(), homepage)
+
+
+def clear_edit_handler(page_cls):
+    def decorator(fn):
+        @wraps(fn)
+        def decorated(*args, **kwargs):
+            # Clear any old EditHandlers generated
+            page_cls.get_edit_handler.cache_clear()
+            try:
+                fn(*args, **kwargs)
+            finally:
+                # Clear the bad EditHandler generated just now
+                page_cls.get_edit_handler.cache_clear()
+        return decorated
+    return decorator
 
 
 class TestPreview(TestCase, WagtailTestUtils):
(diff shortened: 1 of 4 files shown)


lb- (changes requested):
Hey @th3hamm0r this looks good in principle, I have checked your original issue and the comments there and I think it makes sense to add this.

However, it would be really great if you can take a bit more time to add some unit tests for this so that the behaviour does not get lost in the future.

You could do this by
* wagtail/admin/tests/pages/test_preview.py (adding a test here, within `TestPreview`
* see wagtail/admin/tests/test_edit_handlers.py -> TestPageEditHandlers.test_custom_edit_handler_form_class for an example of customising a Page's edit_handler temporarily for a test
* Remember to 'reset' the cached `get_edit_handler` method by copying the code from `clear_edit_handler` in that same edit handlers test file, you may not need the decorator approach but we do not want to pollute other tests

Please reach out if you get stuck, and thanks again for taking the time to investigate the root cause here.

th3hamm0r:
@lb- Thanks for your detailed guide, that helped a lot! :)

Some notes:
- I re-used the code of the existing tests in `TestPreview` to get working test quickly. There is also a `SecretPage` model in the testapp, which already uses a custom panel, which I could have used, but it was easier for me to reuse the lines from the other preview tests, which use the `EventPage`. That allowed me to reuse the defined `post_data` for example.
- I followed your hint to copy the `clear_edit_handler` decorator, but I think it would be a good point to extract it into a utility module. Since I'm too new to wagtail's test setup, I didn't fell confident enough to do that in the first place.
- Actually I don't fully understand, why the `test_custom_edit_handler_form_class` also sets the `edit_handler` before mocking it, so I left that out in my test. Hope I didn't miss something here. https://github.com/wagtail/wagtail/blob/27c3e7921a6c31664abf8587b491861e62f4e74c/wagtail/admin/tests/test_edit_handlers.py#L249

lb- (approved):

lb-:
@th3hamm0r looking great, for some reason the ` ci/circleci: backend — Your tests failed on CircleCI ` failed.
Looks like there needs to be a line break after ` import datetime`

```
ERROR: /home/circleci/project/wagtail/admin/tests/pages/test_preview.py Imports are incorrectly sorted and/or formatted.
--- /home/circleci/project/wagtail/admin/tests/pages/test_preview.py:before	2021-09-27 11:54:04.911275
+++ /home/circleci/project/wagtail/admin/tests/pages/test_preview.py:after	2021-09-27 11:55:38.731172
@@ -1,4 +1,5 @@
 import datetime
+
 from functools import wraps
 from unittest import mock
 
Skipped 956 files

Exited with code exit status 1


```

wagtail/admin/tests/pages/test_preview.py
```diff
@@ -1,13 +1,16 @@
 import datetime
```
lb-, line 1:
```suggestion
import datetime

```

lb-:
@th3hamm0r thanks again, all merged in, really appreciate the time taken to add unit tests and investigate the other potential missing bind_to calls in the codebase.

Issue #6130, opened 2020-06-08
Ways to speed up Wagtail installs


thibaudcolas:
### Is your proposal related to a problem?

I just came out of a workshop where we had participant install Wagtail on Windows, and we noticed quite slow install times, up to 5-10 minutes for a participant. I suppose an element of this might just be slower machines, and @tomdyson let me know on Slack that it’s also important to use an up-to-date version of pip as older versions might not be able to use wheels. I wonder if there are other things we could do to speed up the install though, even if only by a few seconds.

### Describe the solution you'd like

Inspecting a Wagtail install (on macOS),

#### Wheels

It looks like three of Wagtail’s dependencies don’t have prepublished wheels. Should we consider making PRs?

```
Building wheel for l18n (setup.py) ... done
Building wheel for django-treebeard (setup.py) ... done
Building wheel for draftjs-exporter (setup.py) ... done
```

For `draftjs-exporter` – I can easily publish as a wheel, not sure why it’s not set up yet (https://github.com/springload/draftjs_exporter/issues/132).

#### static_src

There is a line in our MANIFEST.in excluding a static_src folder – however it’s using a path that no longer exists since wagtail 2.0. Should we have all `static_src` folders there?

https://github.com/wagtail/wagtail/blob/e10a9f26e7c944fd3a037a0459d371ab122ff96d/MANIFEST.in#L3

```
global-exclude static_src
```

#### tests

We could also probably not publish our tests with Wagtail, hopefully just with:

```
global-exclude tests
global-exclude tests.py
```

#### Source maps

Wagtail only ships three source maps, but together those files account for about 4MB. I think I liked the idea of shipping these with Wagtail so users could report on JS errors that were intelligible, but in practice I haven’t seen this done that much since we started shipping those files.

```
2.1M	admin/static/wagtailadmin/js/draftail.js.map
1.1M	admin/static/wagtailadmin/js/vendor.js.map
440K	admin/static/wagtailadmin/js/wagtailadmin.js.map
```

To exclude these,

```
global-exclude *.map
```

#### Other files

Here are other patterns we should be able to exclude, resulting in small savings:

```
global-exclude .gitignore
# Single file but is quite big and does nothing at runtime.
global exclude wagtail-icomoon.json

(Describe your proposed solution here.)

---

Here is the change with all those modifications done.

```
# Before
29MB, 1742 directories, 2803 files
# After
19MB, 1647 directories, 2356 files
```

Not night-and-day, but worth considering

[...]

PR #7543, opened 2021-09-27
Replace l18n library with built-in translations

vsalvino:
Potentially controversial change, but the `l18n` library needed to be removed in my opinion for a few reasons:
* **Performance**: `l18n` does not provide a wheel package and has been unresponsive to the request to do so. This means that on every install (or cloud deploy, docker image build, test run, etc.) `l18n` must have its wheel built from source by pip. This is actually somewhat time consuming and slows down the Wagtail installation measurably. Over thousands of Wagtail deployments within our CodeRed Cloud hosting service alone, `l18n` is the largest bottleneck of installation speed.
* **Language coverage**: `l18n` only provides translations for a small subset of Wagtail's available languages.
* **Maintenance**: `l18n` does not provide a way for Wagtail to add translations (other than through PR), and after diving into its source code, is a bit convoluted to understand based on what it does. Including flat message files in Wagtail seems a much more simple way to continue adding language support.
* "The dependency of my dependency is my enemy" —Ancient proverb 😜

### Description of this change:
* Includes `.po` and `.mo` files of the timezone translations provided in each of Wagtail's supported languages. These translations are sourced from the [Unicode Consortium CLDR project](https://cldr.unicode.org/).
* ``l18n`` has been replaced with a simple lookup of these provided translation files, using the Python standard library.
* An included utility script, `translate.py` builds these `.po`  and `.mo` files from the list of `pytz` timezones and CLDR translations of the respective regions and exemplar cities.

### Caveats:
* Some translations may be missing - notably the oceans (which are sadly not included in CLDR): Atlantic, Pacific, Indian, Arctic. It should be a simple task to manually translate these in the `.po` files, or add a lookup dictionary to the translate script.
* Translations are provided via a custom message domain (`timezones.po`) rather than the default `django.po` file. As such Django is not aware of them. Therefore Python gettext functionality is used to load them. Since this only occurs on one page in the Wagtail admin, I don't see it being a problem, but suggestions are welcome!

### Reference:
See: #6130 

Opinions and feedback welcome! I would be happy to add unit tests and/or Ocean translations if the core team is willing to accept this change.


thibaudcolas:
Hey @vsalvino 👋  we discussed this in today’s core team meeting.

There was general agreement l18n isn’t currently working as well as we’d like it to for the reasons you mention, however we weren’t clear on whether replacing l18n with what you’re proposing here warrants the maintenance burden for the project.

Personally I’m keen to consider generating localised labels client-side in JS, since it seems we currently only use those labels on `/admin/account/current_time_zone/`, where it should be possible to leverage browser APIs to generate labels. Here is what it would look like:

```
Africa/Abidjan: GMT (Greenwich Mean Time)
Africa/Accra: GMT (Greenwich Mean Time)
Africa/Addis_Ababa: GMT+3 (East Africa Time)
Africa/Algiers: GMT+1 (Central European Standard Time)
Africa/Asmara: GMT+3 (East Africa Time)
Africa/Bamako: GMT (Greenwich Mean Time)
Africa/Bangui: GMT+1 (West Africa Standard Time)
Africa/Banjul: GMT (Greenwich Mean Time)
[…]
UTC: UTC (Coordinated Universal Time)
US/Alaska: AKDT (Alaska Daylight Time)
US/Arizona: MST (Mountain Standard Time)
US/Central: CDT (Central Daylight Time)
US/Eastern: EDT (Eastern Daylight Time)
US/Hawaii: HST (Hawaii-Aleutian Standard Time)
US/Pacific: PDT (Pacific Daylight Time)
US/Mountain: MDT (Mountain Daylight Time)
```

For each time zone name, this displays the identifier, the "short localised form", and the "long localised form". This relies on the [DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat.) JS API. Browser support seems pretty good, except IE11. There are more timezone name formatting options being worked on in: https://github.com/tc39/proposal-intl-extend-timezonename, but browser support for those additional options isn’t there yet.

If anyone wants to try this out on /admin/account/current_time_zone/, here is code to copy-paste in the JS console:

```js
function getTZLabel(timeZone, timeZoneName) {
    // Replace `en-US` with any locale identifier, or with an empty array `[]` to use the current locale.
    const df = new Intl.DateTimeFormat("en-US", {timeZone, timeZoneName});
    const parts = df.formatToParts(new Date())
    return parts.find(p => p.type === 'timeZoneName').value
}

[...document.querySelector('[name="current_time_zone"]').options].forEach(o => {
    const timeZone = o.value;
    if (timeZone) {
        o.innerHTML = `${timeZone}: ${getTZLabel(timeZone, 'short')} (${getTZLabel(timeZone, 'long')})`;
    }
})```

**Comment by vsalvino** [comment]
@thibaudcolas the JavaScript solution looks ideal! I was also looking for a JavaScript solution first but missed this API. Translating timezones does feel like re-inventing the wheel, so I am glad to see there is an API for it.

Given that Microsoft themselves has already dropped IE support on their own products, I think that this would be an acceptable situation for "graceful degradation", as the dropdown could remain functional in IE11 showing the original values.

**Comment by zerolab** [comment]
Thank you @thibaudcolas for summing up the core team discussion and for the alternative. Native browser APIs FTW!

Re: IE11, we will no longer support IE11 with 2.15 (#7535), so I would not worry about it.

**Comment by thibaudcolas** [comment]
For the sake of completeness I thought I should add an example of the results with another language than english – same example as above but in French:

```
Africa/Abidjan: UTC (heure moyenne de Greenwich)
Africa/Accra: UTC (heure moyenne de Greenwich)
Africa/Addis_Ababa: UTC+3 (heure normale d’Afrique de l’Est)
Africa/Algiers: UTC+1 (heure normale d’Europe centrale)
Africa/Asmara: UTC+3 (heure normale d’Afrique de l’Est)
Africa/Bamako: UTC (heure moyenne de Greenwich)
Africa/Bangui: UTC+1 (heure normale d’Afrique de l’Ouest)
Africa/Banjul: UTC (heure moyenne de Greenwich)
[…]
UTC: UTC (Temps universel coordonné)
US/Alaska: UTC−8 (heure d’été de l’Alaska)
US/Arizona: UTC−7 (heure normale des Rocheuses)
US/Central: UTC−5 (heure d’été du Centre)
US/Eastern: UTC−4 (heure d’été de l’Est)
US/Hawaii: UTC−10 (heure normale d’Hawaii - Aléoutiennes)
US/Pacific: UTC−7 (heure d’été du Pacifique)
US/Mountain: UTC−6 (heure d’été des Rocheuses)
```

This makes one of the limitations of the JS approach a bit more apparent – we can’t translate the "zone" names with geographical information, hence why I had appended the short "UTC offset" label, as well as the "long localised form".

**Comment by vsalvino** [comment]
Well, if the JS approach with the non-translated region names is not acceptable to your users, then I would be happy to split this PR out into a separate package if that would make it more doable regarding the maintenance burden.

At that point it would be trading `l18n` for `new-package`. The advantage being having a wheel for faster install and more complete language support. And I would be happy to give wagtail team commit rights.

Still hoping the JS option is acceptable though!

**Comment by gasman** [comment]
Looks like l18n has installation issues on Python 3.10 right now, apparently due to its setup.py looking for a bdist_wininst module that no longer exists...

    ERROR: Command errored out with exit status 1:
     command: /Users/matthew/.virtualenvs/wagtail310/bin/python -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/p6/_51b92q88xj9kk0059bc8f3h0000gn/T/pip-install-7hh6q49_/l18n_b55933a1cdc542e985e2a9ccf5bfd2e9/setup.py'"'"'; __file__='"'"'/private/var/folders/p6/_51b92q88xj9kk0059bc8f3h0000gn/T/pip-install-7hh6q49_/l18n_b55933a1cdc542e985e2a9ccf5bfd2e9/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /private/var/folders/p6/_51b92q88xj9kk0059bc8f3h0000gn/T/pip-pip-egg-info-89zlj7f0
         cwd: /private/var/folders/p6/_51b92q88xj9kk0059bc8f3h0000gn/T/pip-install-7hh6q49_/l18n_b55933a1cdc542e985e2a9ccf5bfd2e9/
    Complete output (12 lines):
      File "/private/var/folders/p6/_51b92q88xj9kk0059bc8f3h0000gn/T/pip-install-7hh6q49_/l18n_b55933a1cdc542e985e2a9ccf5bfd2e9/setup.py", line 74, in <module>
        cmd_module = getattr(__import__('setuptools.command', fromlist=[cmd]),
    AttributeError: module 'setuptools.command' has no attribute 'bdist_wininst'

      File "<string>", line 1, in <module>
      File "/private/var/folders/p6/_51b92q88xj9kk0059bc8f3h0000gn/T/pip-install-7hh6q49_/l18n_b55933a1cdc542e985e2a9ccf5bfd2e9/setup.py", line 78, in <module>
        cmd_module = getattr(__import__('distutils.command', fromlist=[cmd]),
    AttributeError: module 'distutils.command' has no attribute 'bdist_wininst'
    ----------------------------------------

It would be nice (but certainly not a blocker) for the forthcoming Wagtail 2.15 release to support Python 3.10, so that will depend on either l18n releasing a fix or us getting a replacement rolled out.

**Comment by thibaudcolas** [comment]
I’ll ping the core team again to check whether we’re in agreement the JS approach is reasonable, despite the labels not being as good as l18n or the proposed replacement from Vince.

**Comment by gasman** [comment]
The maintainer of l18n has now put out a wheel of the latest release https://github.com/tkhyn/l18n/issues/7 , which solves the 3.10 compatibility issue as well as @vsalvino's original issue :-) This doesn't mean it's not worth switching for the other reasons, but it's no longer urgent.

(PR for 3.10 support is now ready for review: #7577)

**Comment by gasman** [comment]
Apparently l18n uses `getdefaultlocale` which has been deprecated in Python 3.11 and scheduled for removal in 3.13 - so that would be one more reason to investigate alternatives.

**Comment by thibaudcolas** [comment]
I think we should proceed with the JS approach as per https://github.com/wagtail/wagtail/pull/7543#issuecomment-930533955. There are now [even more formats](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat) for us to choose from. According to the TC39 [Extend TimeZoneName Option Proposal](https://github.com/tc39/proposal-intl-extend-timezonename#extend-timezonename-option-proposal), there are translations for 476 locales.

**Comment by vsalvino** [comment]
Re-reviewing this after a few years...

I do think we should replace the l18n package. Reasons:
* Languages - l18n only supports 5 of the 35 Wagtail languages (!!!)
* Performance - one less pip install
* Deprecations - l18n has python deprecation warnings since 3.11

That being said, I'm all for a JavasScript solution, if someone is willing to do the work. 

However, since the work in this branch is already done, I'm also fine with merging it in. The rest of Wagtail uses python translations, so this would be more of the same, easy to maintain, etc. And assuming that timezones don't change their names, we shouldn't have to touch it very often.

File: setup.py [hunk 1]
```diff
@@ -66,6 +65,8 @@
     "factory-boy>=3.2",
     # For running tests in parallel
     "tblib>=2.0,<3.0",
+    # For generating timezone translations.
+    "polib>=1.2,<2.0",
```
zerolab, line 69: we have polib on line 64 - https://github.com/wagtail/wagtail/blob/main/setup.py#L64

vsalvino: Good catch, thanks!

thibaudcolas:
Just noting we discussed this in a core team meeting in November, and agreed this needs to progress. @vsalvino can work on taking this PR further. @thibaudcolas is keen to work on a JS version. Then we’ll get to compare both approaches.

Hopefully one for early 2024.

thibaudcolas:
Re-reviewed today and @laymonage is keen to work on this, to hopefully be able to drop `pytz` (as Django itself no longer depends on it).

vsalvino:
Thanks. I am still very happy with this implementation, because it follows the typical Python/Django translation process, and makes it easy to support all the languages in the admin. New languages can be easily added manually (typical translation process) or via the import script which parses them from the Unicode CLDR.

translate.py
```diff
@@ -0,0 +1,249 @@
...
+
+# Setup django so we can import things from Wagtail.
+import django  # noqa
+
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "wagtail.test.settings")
+django.setup()
+
+
+import polib  # noqa
+import pytz  # noqa
```
laymonage, line 24: We've switched from `pytz` to `zoneinfo.available_timezones` in #12155, so this will need to be updated

wagtail/admin/localization.py
```diff
@@ -117,3 +120,22 @@ def get_available_admin_time_zones():
...
+        # Get current language, and fallback to non-specific version of language.
+        langs = [get_language()] + [get_language().split("-")[0]]
+        lc_path = os.path.join(os.path.dirname(__file__), "locale")
+        _gt = gettext.translation(
+            domain,
+            localedir=lc_path,
+            languages=langs,
+            fallback=True,
+        )
+        return _gt.gettext(message)
```
laymonage, line 140: I have a slight concern that this isn't very performant, since we reconstruct the translation object every time we call `gettext_domain()`. [Django's implementation](https://github.com/django/django/blob/main/django/utils/translation/trans_real.py) has quite a lot of work put into caching the translations per thread, so I assume it could get a bit heavy. Especially since we call this function for every timezone when loading the accounts view.

I wonder whether we can leverage Django's implementation, or maybe somehow make the strings generated as part of the `django` domain generated for Wagtail by `makemessages`. Failing that, maybe add a simple cache that keeps the translation object for each locale.

laymonage (commented):
Thanks for working on this! I've rebased the branch and spent some time investigating. There's still some minor work to do here if we want to proceed with this implementation, especially after #12155. I'll try to work on the JS-based alternative, and if that takes too much time I'll continue this PR again.

translate.py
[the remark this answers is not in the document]
vsalvino: Good idea

wagtail/admin/localization.py
vsalvino: It's been too many years so now I forget why I implemented it this way 😵 But that is a valid point - surely there must be a way to hook into Django's translations.

laymonage:
I've opened #12751 based on Thibaud's solution in https://github.com/wagtail/wagtail/pull/7543#issuecomment-930533955. I think I'm quite happy with that approach – it's much easier to maintain, and we'll automatically get any updates as the browsers get updated. That said, thanks @vsalvino for getting the ball rolling and providing this initial approach!

PR #7545, opened 2021-09-28
Make wagtail compatible with django 4.0

ephes:
The two points I tried to fix were:

* Replaced django.utils.http.urlquote with urllib.parse.quote
* Replaced request.is_ajax() with request.headers.get('x-requested-with') == 'XMLHttpRequest'

All tests run by `python runtests.py` passed, but some were expected failures (didn't install all the javascript stuff etc..).

### DIFF
--- a/wagtail/admin/auth.py
+++ b/wagtail/admin/auth.py
@@ -35,7 +35,7 @@ def users_with_page_permission(page, permission_type, include_superusers=True):
 
 def permission_denied(request):
     """Return a standard 'permission denied' response"""
-    if request.headers.get('x-requested-with') == 'XMLHttpRequest' 
+    if request.headers.get('x-requested-with') == 'XMLHttpRequest':
         raise PermissionDenied
 
     from wagtail.admin import messages
--- a/wagtail/admin/auth.py
+++ b/wagtail/admin/auth.py
@@ -35,7 +35,7 @@ def users_with_page_permission(page, permission_type, include_superusers=True):
 
 def permission_denied(request):
     """Return a standard 'permission denied' response"""
-    if request.is_ajax():
+    if request.headers.get('x-requested-with') == 'XMLHttpRequest' 
         raise PermissionDenied
 
     from wagtail.admin import messages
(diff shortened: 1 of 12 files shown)


lb- (changes requested):
@ephes thanks so much for taking the time to submit this PR, however to make it easier to get this merged in could you please help with the following;

## 1. Split out the URL parse change into its own PR
* It is much easier to assess changes for a smaller subset of work
* This will be the easiest one to merge in as there is a 1:1 replacement by the looks of things
* Also, can you please include a link to the Django release notes in the description of that PR for context https://docs.djangoproject.com/en/dev/releases/3.0/#id3

## 2. Rework the Ajax change
* This one may need a bit more work
* If you look at https://github.com/wagtail/wagtail/pull/7465 (a similar fix) we have opted to no longer check the x-requested-with but simply remove the is_ajax check completely
* However there might be some nuance with these cases and maybe we can consider the recommendation in the docs https://docs.djangoproject.com/en/3.2/releases/3.1/#id2 to use  request accepts https://docs.djangoproject.com/en/3.2/ref/request-response/#django.http.HttpRequest.accepts for when we need to differentiate between a json or html response.
* This also may require some JS changes, while we do use jquery lots we do not want to  be tied into that way of requesting an endpoint
* Hence why, splitting this ajax change to its own PR can help to get some discussion going on the individual cases

ephes:
Thanks for your response :). To move the URL parse change into it's own PR should be no problem, I'll do that a little bit later today.

I was unaware of the is_ajax change being a nontrivial change and I have to read a little bit about this :).

gasman:
Just to note, while I've been working on gradually eliminating is_ajax (and currently have a PR in progress for updating the workflow task chooser), I can't guarantee I'll be done with that in time for Django 4.0 being released - so I'd be OK with the x-requested-with change as a stopgap fix.

lb- (approved):
As per the comment from @gasman  - let's get this in as an initial round of work, I will rebase and merge this in.

@ephes it would be great if you get the chance to prepare a new PR with the request.accepts usage instead, let us know if you need a hand.

lb-:
Tested
* User groups search
* User search
* Inline panel creation
* Search promotions search
* Redirects search
* Modeladmin url generation

lb-:
Tested
* User groups search
* User search
* Inline panel creation
* Search promotions search
* Redirects search
* Modeladmin url generation

lb-:
@ephes thanks heaps for this, merged in via a8636781b0d0fd92cc9c7da2075e4b5c758b38a2 & parents

PR #7547, opened 2021-09-29
fix: Replacement file deletion in document edit

vbudovski-alliance:
Fixes a bug in document edit where uploading a new file with the same filename will cause the new upload to be deleted
from storage.

### DIFF
--- a/wagtail/documents/views/documents.py
+++ b/wagtail/documents/views/documents.py
@@ -162,10 +162,13 @@ def edit(request, document_id):
                 doc.save()
                 form.save_m2m()
 
-                # If providing a new document file, delete the old one.
-                # NB Doing this via original_file.delete() clears the file field,
-                # which definitely isn't what we want...
-                original_file.storage.delete(original_file.name)
+                # Only delete the original file if it was uploaded to a different path.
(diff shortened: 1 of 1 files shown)


lb-:
@vbudovski-alliance thanks for taking the time to put together a PR, however I am trying to work out the bug that this is trying to fix.

Could you please update the description with the steps to reproduce the bug that this fixes.

I have tried;
1. Edit an existing doc & upload a file with the same name -> it will upload the replacement file fine and append a hash to the file.
2. Upload a NEW document with the same filename as one that already exists -> same behaviour as item 1 (it will update the document with the same file name and then show a message ' document updated'). This was done via multiple upload.

Just trying to get to the bottom of the issue and how this fixes it. thanks.

Also, assuming this fix is needed, it would be great to have some unit tests added so that there is not a future change that breaks it again.

gasman:
I think this is the same django-storages issue as previously reported in #4502 and #4495 - you need to set `AWS_S3_FILE_OVERWRITE = False` in settings (or the equivalent for your chosen storage backend), as the default settings provided by django-storages break Django's spec.

See #4567 for discussion about a proper long-term fix for this.

PR #7550, opened 2021-09-29
Model audit log - add logging

gasman:
Extend the audit logging support to cover logging non-Page models, via a new ModelLogEntry model with a generic foreign key. The log registry is responsible for directing logs to the appropriate table, and as such it now provides a general-purpose `log` method (and the old method of calling `PageLogEntry.objects.log_action` is deprecated).

Logging has been added to all sub-apps of the Wagtail admin except for images / documents (which would need their own dedicated LogEntry model to deal with filtering reports by the user's collection permissions - this is out of scope for the current development).

No changes to the reports have been made here - those will follow in another PR.


wagtail/contrib/search_promotions/views.py
```diff
@@ -167,7 +189,10 @@ def delete(request, query_id):
     query = get_object_or_404(Query, id=query_id)
 
     if request.method == 'POST':
-        query.editors_picks.all().delete()
+        editors_picks = query.editors_picks.all()
+        for search_pick in editors_picks:
+            log(search_pick, 'wagtail.delete')
```
zerolab, line 194: question: could `log()` take on a queryset or a list to do the heavy lifting? or is this done to be explicit in the places that we need it to log in bulk?

wagtail/users/views/users.py
```diff
@@ -15,6 +15,7 @@
 from wagtail.admin.forms.search import SearchForm
 from wagtail.core import hooks
 from wagtail.core.compat import AUTH_USER_APP_LABEL, AUTH_USER_MODEL_NAME
+from wagtail.core.log_actions import log
```
zerolab, line 18: question: are images/documents outside of the scope of this PR?

zerolab (commented):
Lots of excellent changes here @gasman 🎉
Got a couple question for the firs pass. Will test locally in the next few days

wagtail/contrib/search_promotions/views.py
gasman: That could be a future enhancement, yep. I don't know if we'd gain much from it, though - I think the places like this where the bulk action code _doesn't_ already have a loop to put the `log` line in are likely to be pretty rare.

wagtail/users/views/users.py
gasman: Yep - there's a bit of extra work involved for those, since we'd want the Site History report to only include images/documents in collections that the viewing user has 'choose' permission on, and in turn that means we need to store the collection ID as part of the log model for efficient filtering (and thus a dedicated MediaLogEntry model). Logging for images/documents wasn't part of Motley Fool's requirements, so hasn't been included in the development they're sponsoring.

(Having said that, the lack of logging on those models feels like a conspicuous absence when everything else has it, so I may well try and fit that in if there's time before 2.15rc1!)

hazho:
I think, if we create a module or app or hook_registery to log every non-page activities automatically, like how the django registers model-admins of all installed apps' models in just simple iteration, like:
```
from django.apps import apps
models = apps.get_models()
for model in models:
    admin.site.register(model)
```

hazho:
> I think, if we create a module or app or hook_registery to log every non-page activities automatically, like how the django registers model-admins of all installed apps' models in just simple iteration, like:
> 
> ```
> from django.apps import apps
> models = apps.get_models()
> for model in models:
>     admin.site.register(model)
> ```
Of course for our case would be much complex registry manager, something like:
```
from wagtail.util import register_logs
@hooks.register('loging_acitivities'):
def log_following():
    models=['app_name.modelname_1',''app_name.modelname_2'',...etc]
    for model in models:
        register_logs(model)
```
        
can be collected into a table that has something like the following fields:

- action_title
- action_type
- changer
- affected_amodels
- date
- time
- ip_address
- browser_agent
- extra_notes

hazho:
> @hazho Logging has been added to the snippets and modeladmin apps, so for most situations that a Wagtail site implementer is dealing with, it will happen automatically with no need to register anything new. If they create their own admin views, then they'll have to add their own logging, but that's just a single line of code.

do you mean in the current version 2.14.1? if yes, where can an audit or admin see the recorded logs?

gasman:
@hazho No, I mean in the current PR.

jacobtoppm (commented):
Generally I think this is looking great! I was unsure about the thread local context pattern initially when it was just the user, but I think for the UUID it definitely makes sense now.

Added one concern about transactions, but otherwise I think this is good to go.

wagtail/contrib/search_promotions/views.py
```diff
@@ -77,11 +78,30 @@ def save_searchpicks(query, new_query, searchpicks_formset):
             # Make sure the form is marked as changed so it gets saved with the new order
             form.has_changed = lambda: True
 
+        # log deleted items before saving, otherwise we lose their IDs
```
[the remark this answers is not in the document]
gasman: Sounds fair enough! Transactions added in 2ef42e9c8c2e08ec4d0547f13864e403c6ed91cc.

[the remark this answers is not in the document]
zerolab: This fits quite neatly with #7556

jacobtoppm (approved):
Looks good to me, thanks for adding those transactions!

PR #7551, opened 2021-09-29
Model audit log - site history report

gasman:
Incorporates #7550 - extend the Site History report to include logs from non-Page models.


wagtail/admin/views/reports/audit_logging.py
```diff
@@ -158,17 +159,22 @@ def decorate_paginated_queryset(self, queryset):
...
                 .prefetch_related('user__wagtail_userprofile', 'content_type')
-                .in_bulk(pks)
+                .filter(pk__in=pks)
+                .with_instances()
             )
-            for pk, log_entry in log_entries.items():
-                object_lookup[(log_model_index, pk)] = log_entry
+            for log_entry, instance in log_entries:
+                # annotate log entry with an 'edit_url' property
+                log_entry.edit_url = url_finder.get_edit_url(instance)
```
jacobtoppm, line 176: Could there be some value in extending the `AdminURLFinder` to accept just an id + model type at some point? It feels like we mostly need the instance for media and page models, but lots of modeladmin registered models operate using a much simpler permissions policy that doesn't need the instance. Still, then we'd probably need to add bulk operations to it to fetch the instances efficiently in the case it did need them, so maybe this is more pain than it's worth.

jacobtoppm (approved):
All looks good to me!

wagtail/admin/views/reports/audit_logging.py
gasman: I went round in circles a bit on this :-) Originally I tried to avoid fetching object instances on the site history report, since in all other respects the LogEntry model goes to a lot of trouble to store all the required information within the log entry. But in the end, you need to do the object lookup to find out whether it's been deleted (and thus whether to actually output the link or not). I suspect that any other use-case for AdminURLFinder (e.g. outputting "deleting this object will delete the following related objects" for a confirmation page) will either run into the same thing, or have to retrieve the object for some other reason (e.g. to display its string representation), so I'd hold off on that until a case for it comes up.

gasman:
Merged in d67fdd1b8b216f9c917e9b09f83ff89fd466fb11

wagtail/admin/views/reports/audit_logging.py
jacobtoppm: Good point, I hadn't thought about deletion in this context!

PR #7552, opened 2021-09-29
Link to Github discussion boards

gasman:


lb- (commented):
I think it would be good to also replace the reference to Google Groups in docs/contributing/issue_tracking.rst

```
Do not use issues for support queries or other questions ("How do I do X?" - although "Implement a way of doing X" or "Document how to do X" could well be valid issues). These should be asked on `Stack Overflow <https://stackoverflow.com/questions/tagged/wagtail>`_, or for discussions that do not fit Stack Overflow's question-and-answer format, the `Wagtail Support Google group <https://groups.google.com/forum/#!forum/wagtail>`_.
```

Maybe the Github discussions would be a good replacement in that line?

lb-:
Can add more links elsewhere in the docs if needed in due course. merging in.

PR #7563, opened 2021-10-01
Fix link to node.js download

gasman:
https://nodejs.org/download/ just points to a directory listing now; the download link is now on the https://nodejs.org/ homepage, and there are no instructions to speak of.

### DIFF
--- a/docs/contributing/developing.rst
+++ b/docs/contributing/developing.rst
@@ -15,7 +15,7 @@ If you'd prefer to set up all the components manually, read on. These instructio
 Setting up the Wagtail codebase
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Install Node.js, version 14. Instructions for installing Node.js can be found on the `Node.js download page <https://nodejs.org/download/>`_.
+Install `Node.js <https://nodejs.org/>`_, version 14.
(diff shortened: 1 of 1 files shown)


docs/contributing/developing.rst
```diff
@@ -15,7 +15,7 @@ If you'd prefer to set up all the components manually, read on. These instructio
 Setting up the Wagtail codebase
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Install Node.js, version 14. Instructions for installing Node.js can be found on the `Node.js download page <https://nodejs.org/download/>`_.
+Install `Node.js <https://nodejs.org/>`_, version 14.
```
RealOrangeOne, line 18: Question: Would linking to https://nodejs.org/en/download/ instead be better? Not only does it note all the possible download / install options, but also links off to previous versions and other potentially useful resources. It's also a page designated for download links, rather than the homepage which is more likely (if however unlikely) to change in future.

gasman: I figure that whatever thought processes led the node.js folks to select those particular links as the most useful ones to put on the homepage - e.g. not wanting to bombard people with unnecessary choices - probably apply to people installing it for Wagtail too, so we should trust their judgment. Anyone who doesn't want to go with the suggested option can find the downloads page easily enough.

(Plus, this way we're not imposing the English version of the page on people if that's not what they want...)

Issue #3657, opened 2017-06-14
Wagtailusers app cannot be disabled as wagtailadmin depends on it


kaedroho:
``wagtailadmin/utils.py`` currently imports ``wagtail.wagtailusers.models`` which makes it impossible to remove the app to disable user account management.

IIRC, the reason why it was made a separate app was so that the UAM features could be switched off. The ``UserProfile`` model is required by ``wagtailadmin`` but is unrelated to the UAM views also provided by that app so maybe that model belongs in ``wagtailadmin``.

gasman:
👍 . Putting the UserProfile model in `wagtailusers` was a misstep...

robinzachmann:
Is there any workaround for this?

gasman:
@robinzachmann You could remove the Users menu item via the `construct_main_menu` hook: https://docs.wagtail.org/en/stable/reference/hooks.html#construct-main-menu

kaedroho:
This will be fixed by https://github.com/wagtail/wagtail/pull/7564

thibaudcolas:
We’ve decided to close #7564 as it’s fallen out of date, and from recent developments it seems clear we would be more keen for `UserProfile` to go to `wagtail.admin`. So this is ready for anyone to pick up. Aside from the code contributions, it would be great to hear more from anyone who has a need to disable `wagtailusers` – what’s the rationale?

This seems desirable in the context of security-hardened sites that run as few "admin" apps as possible in production, with Django in read-only mode. Are there other use cases we should be aware of? This all feels interesting and aligns with Wagtail’s focus on security, but also relatively niche.

laymonage:
I'd just like to add that if we do migrate the `UserProfile` into `wagtail.admin`, we should learn from our experience in moving the models from `wagtail.search` to `wagtail.contrib.search_promotions`. For example, there was #10847.

We also often see cases where people use those models but they forgot to add the `search_promotions` app into the `INSTALLED_APPS` list when upgrading. It's unlikely in this case as `wagtail.admin` comes by default and devs who don't have it in their `INSTALLED_APPS` would probably run into other issues already, but it's something to keep in mind when writing the docs.

PR #7564, opened 2021-10-01
Move UserProfile into wagtailcore

kaedroho:


wagtail/core/migrations/0064_copy_userprofiles.py
```diff
@@ -0,0 +1,38 @@
...
+        ],
+        ignore_conflicts=True
+    )
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('wagtailcore', '0063_userprofile'),
+        ('wagtailusers', '0011_userprofile_related_name_change'),
```
gasman, line 31: Doesn't this dependency screw up the possibility of removing wagtail.users from INSTALLED_APPS? (i.e. if someone does that then tries installing their project again on a fresh db, this will refer to an unrecognised migration)

kaedroho:
This has now been rebased

kaedroho:
Rebased

wagtail/migrations/0071_copy_userprofiles.py
```diff
@@ -0,0 +1,38 @@
...
+        [
+            UserProfile(
+                user_id=user_profile.user_id,
+                submitted_notifications=user_profile.submitted_notifications,
+                approved_notifications=user_profile.approved_notifications,
+                rejected_notifications=user_profile.rejected_notifications,
+                updated_comments_notifications=user_profile.updated_comments_notifications,
+                preferred_language=user_profile.preferred_language,
+                current_time_zone=user_profile.current_time_zone,
+                avatar=user_profile.avatar,  # TODO does this copy the avatar?
```
kaedroho, line 22: I'll check this today

wagtail/migrations/0071_copy_userprofiles.py
```diff
@@ -0,0 +1,38 @@
...
+            )
+            for user_profile in OldUserProfile.objects.all()
+        ],
+        ignore_conflicts=True,
```
jacobtoppm, line 26: `Enabling this parameter disables setting the primary key on each model instance` from the Django docs - is there a reason we need this, and don't want to try and set the pk to make migrating any FKs to `UserProfile` easier? Probably a pretty minor concern here though. (I was interested to find out that the OneToOneField on `UserProfile` isn't the primary key though! Had always assumed it would be)

jacobtoppm (commented):
Looks good to me! One quick question, and the avatar copying to check, but otherwise all looks good.

wagtail/migrations/0071_copy_userprofiles.py
jacobtoppm: I don't, I'm afraid! Happy for it to stay as is unless we do make the `OneToOneField`-> primary key change here, in which case we should check the key is still set properly

I think that would make sense! Tbh I spotted that problem while assuming that `OneToOneField` was the primary key - that way round seems pretty natural to me

wagtail/migrations/0071_copy_userprofiles.py
kaedroho: This is all good, it reuses the existing file instead of copying it

wagtail/migrations/0071_copy_userprofiles.py
kaedroho: Done!

thibaudcolas:
Input from @laymonage Now that we have other models in `wagtail.admin`, this will be a better place for user profile data than `core`?

From core team discussions – assigning to myself to convert this to an issue. Some uncertainty as to whether we even should support running Wagtail without `wagtailusers` (@RealOrangeOne needs this!)

thibaudcolas:
Closing in favour of #3657

PR #7566, opened 2021-10-01
Removing restrictions should not show "Removed Public restriction"

hazho:
currently when the page editor removes the restrictions, the page and site history shows like (Removed the 'Public' view restriction), and this off course is not making any sense, this PR is just simply removing the variables and shows that the restrictions are removed


hazho:
I think my forked repo is not a mirrored as the wagtail:main I forgot to to pull changes before I make my changes, forgive me about this, I'll close this one and create new PR after fixing this

Issue #5338, opened 2019-05-29
Implement focus management for chooser modals


thibaudcolas:
Originally reported in #4199. Related: #5336, #5337

It should be possible for screen reader users to interact with chooser dialogs for images, documents, snippets, embeds, links.

Context from #4199:

> Blind people rely entirely on the keyboard to access their system, so the mouse is out of the question. They use Tap and the arrow keys to find their way around, while the browser respond to these navigation commands by setting the keyboard focus on each element according to its tabindex

> If you want to embed an image on a page, a file browser dialog opens, again, the user is unaware of its appearance because the focus is still. Once the focus is inside the modal, it should be trapped there. There should also be a way to move between tabs within the modal.

At the moment, here are specific issues users encounter:

- When opening a chooser dialog, there is no indication that the dialog has opened
- Focus doesn't move to the chooser dialog once opened
- Focus isn't trapped in the chooser modals once there

thibaudcolas:
Katie from Torchbox has had a look at some potential improvements we could make to those modals to make them more accessible & usable. I’ll copy her recommendations over here for future reference. Some might go beyond strict "focus management" but they feel relevant nonetheless.

## Image & document upload modal behaviour

Issues:

 - The user is not aware of the modal as the focus is still when opening.
 - The voice over announcement is not as helpful as what it could be.
 - ESC is not closing the modal

Recommendations:

 - Ensure that the focus is recognising there is a modal shown on the UI
 - Give context on what’s going on, i.e “You are now on ‘choose an image’ modal” before listing actions within that modal. 

![Screenshot 2019-06-24 at 16 59 54](https://user-images.githubusercontent.com/877585/60034025-9d08a000-96a1-11e9-9f57-9e31ae837310.png)

## Image & document upload modal tab order

Recommended tab order:

- Once the user has learned from the screen reader what the modal is about, they should have the -option to close the modal within the first few tabs rather having to go all the way to the bottom to close the modal.
- We should make sure that ESC on the keyboard closes the modal. 

![Screenshot 2019-06-24 at 17 02 43](https://user-images.githubusercontent.com/877585/60034157-dd681e00-96a1-11e9-9b6b-3456e42b1f9f.png)

## Image & document upload modal: other identified issues 

Issues:

 - The screen reader does not give help text context that’s related to the field
 - The screen reader doesn’t say there is a ‘Choose file’ button and says there is invalid data would could confuse the user

Recommendations:

[...]

PR #7568, opened 2021-10-03
modal-workflow - block additional opens & add trigger focus management

### LINKED ISSUE #4006: Image, Page, Document, Snippet chooser modal loads more than once on multiple clicks
### Issue Summary

When using one of the in line Document or Image choosers on a slow internet connection, the user might click 'Choose An Image' multiple times.

Clicking multiple times loads the modal in the background multiple times, meaning the user has to close each one to get back to the 'base' one before choosing something will actually work.

### Steps to Reproduce

It's essential that you provide enough information for someone else to replicate the problem you're seeing. Simply describing something that's broken on your current project is not enough!

1. Start a project using the wagtaildemo
2. Log into Admin
3. Edit the Blog page which contains an image link
4. On Chrome open developer tools, click the Network tab and change speed to 'slow 3G'
5. Click 'Choose an image' three times quickly (the modal will not appear for about 1500ms
6. Attempt to select an image (or upload one)
6a. Expected - image is selected and modal is closed.
6b. Actual - image is selected, background gets slightly lighter, and you are left with the 'underneath' modal without any selection.

Note: This is even more of an issue if you upload a new image because you will need to upload it multiple times without realising what is happening.

### Technical details

* Python version: 3.4.3
* Django version: 1.11.6
* Wagtail version: 1.13
* Browser version: Chrome 61 on Mac

### Potential Solution - First resolve why Modals can be opened again.

The code for `ModalWorkflow` looks like it should remove existing modals when a new one is opened.

`$('body > .modal').remove();`
[modal-workflow.js](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailadmin/static_src/wagtailadmin/js/modal-workflow.js)

### Potential Solution - Disable Button Until Modal Closed

Whenever `ModalWorkflow` is triggered on `action-choose` buttons we could possibly disable the button until the modal is loaded (or closed).

This also visually indicates that something is happening, maybe we could add a loading indicator in the future. Maybe look at a way to make this easier as part of the `ModalWorkflow` library, eg. add something to opts like an additional function to run on modal close.

```
    $('.action-choose', chooserElement).click(function(event) {
        buttonClicked = event.target;
        buttonClicked.addClass('disabled'); // disables the button immediately on click
        ModalWorkflow({
            url: window.chooserUrls.imageChooser,
            responses: {
                imageChosen: function(imageData) {
                    buttonClicked.removeClass('disabled'); // enable the button when image selected
                    input.val(imageData.id);
                    previewImage.attr({
                        src: imageData.preview.url,
                        width: imageData.preview.width,
                        height: imageData.preview.height,
                        alt: imageData.title
                    });
                    chooserElement.removeClass('blank');
                    editLink.attr('href', imageData.edit_link);
                }
            }
        });
    });
```

The equivalent will need to be reviewed and updated wherever the ModalWorkflow is created, here are some - need to check fully for any others.

* [image-chooser.js](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailimages/static_src/wagtailimages/js/image-chooser.js)
* [document-chooser.js](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtaildocs/static_src/wagtaildocs/js/document-chooser.js)
* [snippet-chooser.js](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailsnippets/static_src/wagtailsnippets/js/snippet-chooser.js)
* [page-chooser.js](https://github.com/wagtail/wagtail/blob/master/wagtail/wagtailadmin/static_src/wagtailadmin/js/page-chooser.js)

lb-:
## Overview

- ensure that when the modal opens it keeps track of the element that triggered it
- disable that element so that it cannot accidentally be clicked twice
- re-enable the element when closing modal and then move focus back to the trigger when closed

## Notes

- added code to remove the modal from DOM when closed (existing behaviour was to just call the modal's hide function and it would still be in DOM, however it seems better to remove it fully unless I am missing something)
- a reminder that there is still a short delay between pressing the button and the modal showing (as discussed on #5338 ), this PR does not address this but does consider this scenario, focus can remain on a disabled button and the focus will be allowed as the button will be re-enabled in the step before the actual modal is fully removed from DOM so that there should not be a visible delay to re-enabling it.
- fixes #4006 + partial work towards #5338

## Recording

* See recording, on Slow 3G network throttling.

https://user-images.githubusercontent.com/1396140/135741248-80475d2f-beb7-4fbd-933d-9899428b1f86.mp4

## Checks

* Do the tests still pass? 👍 
* For front-end changes, checked Firefox 92, Chrome 94, Safari 14 on MacOS Big Sur
* Tested keyboard control using the above browsers, tabbing through to multiple modal chooser activations (draftail, workflow chooser, snippet chooser, page chooser) then pressing enter or space or 'esc' while in the modal to confirm the focus moves back to the trigger button.

### DIFF
--- a/client/src/entrypoints/admin/modal-workflow.test.js
+++ b/client/src/entrypoints/admin/modal-workflow.test.js
@@ -1,0 +1,133 @@
+import $ from 'jquery';
+window.$ = $;
+window.jQuery = $;
+
+import '../../../../wagtail/admin/static_src/wagtailadmin/js/vendor/bootstrap-modal';
+import './modal-workflow';
+
+$.get = jest.fn().mockImplementation((url, data, cb) => {
+  cb(JSON.stringify({ html: `<div id="url">${url}</div>`, data, step: 'start' }));
+  return { fail: jest.fn() };
+});
+
+describe('modal-workflow', () => {
+  beforeEach(() => {
+    document.body.innerHTML =
+      '<div id="content"><button class="button action-choose" id="trigger">Open</button></div>';
+  });
+
+  it('exposes module as global', () => {
+    expect(window.ModalWorkflow).toBeDefined();
+  });
+
+  it('should generate a modal when called that is removed when closed', () => {
+    let modalWorkflow;
+
+    const openModal = () => {
+      // eslint-disable-next-line new-cap
+      modalWorkflow = window.ModalWorkflow({ url: 'path/to/endpoint' });
+    };
+
+    expect(document.getElementsByClassName('modal')).toHaveLength(0);
+
+    const triggerButton = document.getElementById('trigger');
+
+    triggerButton.addEventListener('click', openModal);
+    triggerButton.dispatchEvent(new MouseEvent('click'));
+
+    // check that modal has been created in the DOM
+    const modal = document.getElementsByClassName('modal');
+    expect(modal).toHaveLength(1);
+    expect(modal).toMatchSnapshot();
+
+    // check that modalWorkflow returns self
+    expect(modalWorkflow).toBeInstanceOf(Object);
+    expect(modalWorkflow).toHaveProperty('close', expect.any(Function));
+
+    // close modal & check it is removed from the DOM
+    modalWorkflow.close();
+    expect(document.getElementsByClassName('modal')).toHaveLength(0);
+  });
+
+  it('should pull focus from the trigger element to the modal and back to the trigger when closed', () => {
+    let modalWorkflow;
+
+    const openModal = () => {
+      // eslint-disable-next-line new-cap
+      modalWorkflow = window.ModalWorkflow({ url: 'path/to/endpoint' });
+    };
+
+    expect(document.getElementsByClassName('modal')).toHaveLength(0);
+
+    const triggerButton = document.getElementById('trigger');
+
+    triggerButton.focus();
+    triggerButton.addEventListener('click', openModal);
+
+    expect(document.activeElement).toEqual(triggerButton);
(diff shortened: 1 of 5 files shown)


lb-:
Rebased on main

client/src/entrypoints/admin/modal-workflow.test.js
```diff
@@ -0,0 +1,133 @@
...
+    triggerButton.dispatchEvent(new MouseEvent('click'));
+
+    // check that modal has been created in the DOM
+    const modal = document.getElementsByClassName('modal');
+    expect(modal).toHaveLength(1);
+    expect(modal).toMatchSnapshot();
+
+    // check that modalWorkflow returns self
+    expect(modalWorkflow).toBeInstanceOf(Object);
+    expect(modalWorkflow).toHaveProperty('close', expect.any(Function));
```
thibaudcolas, line 45: I would say this is superfluous since we’re calling it right after, but at least the error message will be easier to understand.

thibaudcolas (approved):
Thank you @lb-, this is excellent. I particularly appreciate you took the time to add tests for `onload` and the URL params, on top of the open/close focus management!

client/src/entrypoints/admin/modal-workflow.test.js
lb-: Yeah. This is kind of a habit of mine, first check for presence of the thing and then check the thing has a specific part we want. Makes it easier to debug and also when writing tests. Often I'll remove the first check before committing but forgot here.

PR #7569, opened 2021-10-04
add tags to existing Django registered checks

lb-:
- add tags to existing Django `checks` registration calls, makes testing easier and also ensures that specific checks can be run in the command line in isolation
- see https://docs.djangoproject.com/en/3.2/topics/checks/#registering-and-labeling-checks
* Do the tests still pass? 👍 


lb-:
Thanks @nmorduch !

Issue #7556, opened 2021-09-30
Add transaction.atomic to modifying views


th3hamm0r:
### Is your proposal related to a problem?

Currently, the majority of all modifying views in wagtail, which often contain many different db reads and writes, don't use a wrapping DB-transaction. So errors in the middle of the processing could result in a partially saved state.

For example:
Publishing a page (implemented [here](https://github.com/wagtail/wagtail/blob/4ab7080fb8e3def96ec6e82df20c9b38e1a93f2e/wagtail/admin/views/pages/edit.py#L471)) saves the revision, publishes the page, updates comments, sends notifications, calls thirdparty hooks, etc. If any of those steps fails, the previous modifications are already written to the db. Wrapping it in django's `transaction.atomic` would be an easy fix.

Another example:
If a developer uses the recipe to automatically create redirects (https://docs.wagtail.io/en/stable/reference/pages/model_recipes.html#have-redirects-created-automatically-when-changing-page-slug), when pages are moved, theoretically it could happen, that pages get moved, but due to an error in a code after the page has been moved, eg. in the hook itself, the page would be moved, but no redirect would have been created. A wrapping transaction would ensure that either all the db writes are successfully written, or none of them.

Existing `transaction.atomic` calls:
There are some places, that have been improved, for example deleting pages uses a transaction (added with https://github.com/wagtail/wagtail/pull/4673): https://github.com/wagtail/wagtail/blob/4ab7080fb8e3def96ec6e82df20c9b38e1a93f2e/wagtail/admin/views/pages/delete.py#L18
But I think that even that code could be improved by also wrapping the SELECT-queries too, so we can profit from the DB's serialization functionality which could detect parallel modifications of the same objects better.

### Describe the solution you'd like

Similar to django ensures transaction safety in its `ModelAdmin` by wrapping `changeform_view` (see [here](https://github.com/django/django/blob/492ed60f236d770eb9a6d56d85ff2550bb1ecfff/django/contrib/admin/options.py#L1548)) and `delete_view` (see [here](https://github.com/django/django/blob/492ed60f236d770eb9a6d56d85ff2550bb1ecfff/django/contrib/admin/options.py#L1863)) in a `transaction.atomic` call, wagtail could also add this to its modifying views.

I know, there are many places which must be edited, so I don't know exactly how to approach this issue.
But I really want to help improving this, so I'm looking forward to get some feedback here.

[...]

PR #7571, opened 2021-10-05
Added more wrapping database transactions to the page admin (#7556)

th3hamm0r:
To avoid too many changes I started with the page administration and added wrapping `transaction.atomic` calls.
This should avoid possible data inconsistencies and also allow developers to add `on_commit`-functionality, that is only executed on success of the whole transaction (see #7556).

**TBD:**
To be consistent with existing `transaction.atomic` usages, for example in [delete.py](https://github.com/wagtail/wagtail/blob/9aa2f68dfa04d5ab21eab059fc5432e7fcfee883/wagtail/admin/views/pages/delete.py#L18), I've wrapped the "before"-hooks too. This was easy for the function-based views, but for the class-based views, eg. the EditView, I basically would have to wrap the whole `dispatch()`-call with `@transaction.atomic`.
I avoided that for now, but that results in some minor inconsistencies, eg. using the `on_commit`-hook in `before_edit_page` won't be part of the newly added transaction. What do you think?

### DIFF
--- a/wagtail/admin/views/pages/copy.py
+++ b/wagtail/admin/views/pages/copy.py
@@ -1,4 +1,5 @@
 from django.core.exceptions import PermissionDenied
+from django.db import transaction
 from django.shortcuts import redirect
 from django.template.response import TemplateResponse
 from django.utils.translation import gettext as _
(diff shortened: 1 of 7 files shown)


jacobtoppm (commented):
Thanks for contributing!

- We need to think about those inconsistencies. My gut feeling is that we should probably standardise whether the before hooks are inside or outside a transaction.
- If we wrap things in transactions, I think we should probably be using `on_commit()` for the notifications - this could be very confusing to users otherwise.
- I think this would need an upgrade consideration written. Some hooks will no longer be possible without using `on_commit`, as this is written (for example, anything relying on an external service to be able to read the current database state).

th3hamm0r:
This was one of my early PRs, and there is obviously more thinking and work to do, so I'm marking it as draft for now.

lb-:
@th3hamm0r It's been a while on this PR, to avoid long lived open PRs may I suggest you close this and raise a new issue or  maybe discussion presenting a bit of a proposal for what we could do here.

I do not know feel confident enough about the impacts/big picture of this specific PR to create that discussion. It probably does not need an RFC by the looks of things but will need some buy in on the approach and probably a discussion or two at the core team meetings.

th3hamm0r:
Ok, I'm closing this PR. I think the related issue #7556 is still valid (I've just removed a clearly outdated part).

PR #7572, opened 2021-10-05
Deployment docs - update wagtail getting started

lb-:
* existing link does not work, just redirects to div.io home page
* New link https://docs.divio.com/en/latest/introduction/wagtail/


lb-:
Thanks @zerolab

PR #7573, opened 2021-10-05
Bulk actions

zerolab:
This PR adds bulk actions for pages, images, documents and users. It was done as part of this year's GSoC by @KIRA009. It provides

- A mechanism for registering bulk actions
- hook up with the UI
- documentation
- and tests

The work follows the initial [design guidance](https://www.figma.com/proto/q6DeuoBXGCtYdf7GdGJs8D/Wagtail-Design-Toolkit?node-id=819%3A474&viewport=783%2C670%2C0.25&scaling=min-zoom&hide-ui=1) with tweaks and deviations along the way

To-Do
- [ ] check and add the new audit log entry points where relevant
- [ ] check and tweak for the new generic list views

Additional thanks to @jacobtoppm and @Stormheg for the mentoring and reviews (along with many others for the input)


gasman (changes requested):
This is looking really good! Really impressed with how well thought-out the UX is here - the way it's unobtrusive but easily findable, the intuitive behaviour of the select all / select none checkboxes, the summary of changes on the confirmation page, getting the pluralisation right everywhere...

Just a couple of niggles:

* The checkboxes show up in the page chooser modal when they shouldn't (due to it sharing some of the listing includes with the index page - we really need to clean up that spaghetti!)
* On the 'toggle activity' action for users, it took me a long time to understand what the "Activate / Deactivate" checkbox meant. To me, "toggle" implies that the end state will be the opposite of the initial state (so if I apply the action to two users, one of which is active and the other is inactive, I should end up deactivating the first one and activating the second) - and a checkbox should always be a choice between "do the labelled thing" versus "don't do the thing", never a choice between two opposing actions. Also, I'm not keen on the word "activity" here - it sounds like it's something to do with a log of actions the user has performed. (It's a shame "activeness" isn't a word...) I'd suggest renaming the action to "set active state", and changing the checkbox to a dropdown or pair of radio buttons with options Activate / Deactivate (or Active / Inactive).

zerolab:
Thank you for the feedback @gasman!

The checkboxes showing in the page chooser is a known issue (just checked our internal tracker). In addition to that we need to revise the checkbox borders, they fail contrast checks. Will aim to submit additional PRs this weekend

kaedroho:
>  check and add the new audit log entry points where relevant

This seems to be working for pages already. Images, Documents and Users do not currently use audit logging

jacobtoppm:
@thibaudcolas Thanks for flagging, approved!

jacobtoppm:
Just a note that we will need to add upgrade considerations, as the new bulk actions could allow users to bypass previous functionality implemented using hooks for actions like copy, etc.

PR #7388, opened 2021-08-01
Add a style fallback to avoid invalid inline styles

tomkins:
Closes #7387

This is mostly copying the behaviour of https://github.com/springload/draftjs_exporter/blob/v2.1.7/example.py#L130 - if an inline style has been added which isn't valid with the site configuration, then the style will be dropped and the contents will be kept.

To test:

- Try and hack in an underline block (similar to #7387)
- A couple of warnings will appear on save:
```
WARNING:root:Missing config for "UNDERLINE". Deleting style.
WARNING:root:Missing config for "UNDERLINE". Deleting style.
[01/Aug/2021 21:42:01] "POST /admin/pages/4/edit/ HTTP/1.1" 302 0
```
- Content will appear without inline style:
Screenshot 2021-08-01 at 22 42 51

PR #7578, opened 2021-10-06
Capture log output from style fallback unit test

gasman:
For some reason the log output generated during the test added in #7388 causes subsequent tests to be much more noisy - wrap it in self.assertLogs to silence this.

(also change logging.warn to logging.warning - the former is obsolete as per https://docs.python.org/3.10/library/logging.html#logging.warning )


lb-:
Thanks @gasman I did not know about the assert logs approach.

PR #7580, opened 2021-10-07
Use SVG icon in homepage's site summary items

jeromelebleu:
To continue to contribute to #6107, this replace font icon by SVG icon in the site summary items of the homepage. As one may have defined custom items using `li.icon`, I kept those styles but tell me if you find that it can be removed.

This also changes the display of the items to have 2 of them per row on small screen instead of 3 - even it should be 4?


lb- (approved):
Looks great @jeromelebleu - I think 50% works well for smaller devices on the item listing, icons are working well.

PR #7581, opened 2021-10-07
added missing cache control to preview routes

th3hamm0r:
I found two bugs in the code where, if django's [per-site cache](https://docs.djangoproject.com/en/stable/topics/cache/#the-per-site-cache) is enabled, the page preview gets cached by the backend once for each page, and this cache entry may be used for the normal page request in the frontend too.

The default page preview correctly disables caching in `serve_preview()` ([see here](https://github.com/wagtail/wagtail/blob/5522992c2923276fca40417401e8fb2c536b4b4f/wagtail/core/models/__init__.py#L2135)), but the `RoutablePageMixin` ([here](https://github.com/wagtail/wagtail/blob/5522992c2923276fca40417401e8fb2c536b4b4f/wagtail/contrib/routable_page/models.py#L166)) and the "landing" preview-mode of the `AbstractForm` ([here](https://github.com/wagtail/wagtail/blob/5522992c2923276fca40417401e8fb2c536b4b4f/wagtail/contrib/forms/models.py#L319)) don't disable it.

This leads to multiple issues if the per-site cache is enabled:
* `RoutablePageMixin`: if you use the `RoutablePageMixin`, the default preview (which renders the index route) will probably cache the first preview once, so any dynamic content that has been changed afterwards won't show up in a following preview request. Further the same cached content may be shown to the same user if he opens the page normally, without using the preview. This is because the preview simply calls the middleware stack with the `full_url` of the page, so the per-site cache will cache the content for that url, which is also used if the user just opens the url normally.
* `AbstractForm`: I have not tested it, but I assume that previewing the "landing" page first and the form afterwards would show the landing page for the form-preview, since the cached content of the first "landing"-mode request will be used

**Fix:**
Based on the `Page`s `serve_preview()` I've added the missing calls to `patch_cache_control(response, private=True)`.

I really tried hard to also write a test for the form page, but I failed to implement the required preview requests in the tests :(
I've tried multiple things but I'm assuming that the simple POST-data containing "title" and "slug" was not enough.

**Important:**
This is only a small fix, but the real problem is that the preview just uses the global django cache and it is very easy to fill the cache for the public frontend with content generated during the preview, so this can result in really harmful issues ([related info here](https://github.com/wagtail/wagtail/pull/5427)). The related issue #5074 **has not been solved yet**, as there is no easy fix I think.

**Related:**
https://github.com/wagtail/wagtail/issues/5074
https://github.com/wagtail/wagtail/pull/5427

### DIFF
--- a/wagtail/contrib/forms/models.py
+++ b/wagtail/contrib/forms/models.py
@@ -6,6 +6,7 @@ from django.core.serializers.json import DjangoJSONEncoder
 from django.core.validators import validate_email
 from django.db import models
 from django.template.response import TemplateResponse
+from django.utils.cache import patch_cache_control
 from django.utils.formats import date_format
 from django.utils.translation import gettext_lazy as _
 
@@ -291,7 +292,9 @@ class AbstractForm(Page):
         if mode_name == "landing":
             request.is_preview = True
             request.preview_mode = mode_name
-            return self.render_landing_page(request)
+            response = self.render_landing_page(request)
+            patch_cache_control(response, private=True)
+            return response
         else:
             return super().serve_preview(request, mode_name)
 
--- a/wagtail/contrib/routable_page/models.py
+++ b/wagtail/contrib/routable_page/models.py
@@ -2,6 +2,7 @@ from django.http import Http404
 from django.template.response import TemplateResponse
 from django.urls import URLResolver, re_path
 from django.urls.resolvers import RegexPattern
+from django.utils.cache import patch_cache_control
 
 from wagtail.models import Page
 from wagtail.url_routing import RouteResult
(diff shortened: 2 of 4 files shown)


th3hamm0r:
TL;DR: rebased the PR, which adds missing response cache headers in 2 places, and a test for one case.

Note: this PR could be marked as "Bug"

th3hamm0r:
Update: This has been fixed for wagtail >= 4 with https://github.com/wagtail/wagtail/commit/ae8f0514bc534904200f742482c0957127eb5bea (#8709)
It it is still a bug for wagtail < 4, but since the support for wagtail < 4 probably ends with feb 2023 (2.15 LTS), I will close this.
If I find some time I may open a PR for the test I've wrote, so the energy invested for that was not in vain.

lb-:
I've raised an issue from this draft PR that was created out of this PR. 

https://github.com/wagtail/wagtail/issues/12981

@th3hamm0r - your efforts are never in vain, thank you for pursuing these problems. 

For now, let's see what will come of the issue raised, maybe with further discussion and input we can land on a direction. Even if it's a documentation update initially.

PR #7585, opened 2021-10-09
preventing the same collection name within the same family(depth) 

hazho:
as a logic, it shouldn't be there any collections' name same as another collection, but this is much better to avoid when the collection name duplication to be avoided within one parent, thus having unique_together validation, would make sure that the same name won't be used, even mistakenly


hazho:
after testing:
- the restriction applied.
- the form is not validating the unique_together, hense throws this error 
image

TODO:
- adding the validation on the browser's Form level.
- testing for any indirect impact on other wagtail_builtin apps

Issue #7483, opened 2021-08-31
UX and security  Improvement


hazho:
in the Forms listing page of WagtailAdmin, it is better to change the (Origin) to (actions) and the origin text that indicates the location of model and parent App, to text of each Actions (currently "Edit"), isn't it?
![image](https://user-images.githubusercontent.com/20362375/131545748-eb43dc35-1a9b-43f7-bf29-7ec4cda4bbe5.png)

hazho:
The Origin members are links to edit pages that been built by, therefore the "Edit" text is better..!
or at least this change should be for non-SuperUsers view

lb-:
@hazho I agree that the form pages listing could do with some refinements, but at the moment we have a pretty consistent behaviour of the first column on the row linking to the 'thing'. In other cases where we do have buttons we have them inside the main thing row with buttons on hover.

See modeladmin, snippet listing and others for existing conventions.

For now I think this needs some further ideas and input.

## Other listing with 'actions' in Wagtail.
<img width="1493" alt="Screen Shot 2022-11-20 at 9 36 54 pm" src="https://user-images.githubusercontent.com/1396140/202899883-c2adfa09-f07a-4fef-8dca-1271baad219c.png">
<img width="1493" alt="Screen Shot 2022-11-20 at 9 36 47 pm" src="https://user-images.githubusercontent.com/1396140/202899884-68826606-9cc2-4a4d-856a-b1f296758a05.png">
<img width="1493" alt="Screen Shot 2022-11-20 at 9 37 23 pm" src="https://user-images.githubusercontent.com/1396140/202899876-0141dd4c-ddb3-49a5-8d80-f53d2becc87c.png">
<img width="1154" alt="Screen Shot 2022-11-20 at 9 41 40 pm" src="https://user-images.githubusercontent.com/1396140/202900032-385e705c-c2ad-46e1-87ca-9ab5250eb7ac.png">

<img width="1150" alt="Screen Shot 2022-11-20 at 9 45 58 pm" src="https://user-images.githubusercontent.com/1396140/202900167-ac016c6c-4596-4338-af73-0c2f5834f0f6.png">

## Potential path forward - treat this like a reports page

If we look at this page as a special report listing of pages (and maybe even use the reports system) we get a nice set of conventions that helps us out more.

Columns

* A. Page with status/translation labels then  'edit' + 'view submissions' on hover?
* B. Type (e.g. `Form Page (base.formpage)`)
[...]

PR #7586, opened 2021-10-09
for Issue #7483

hazho:
Adding actions as buttons, removing the anchor link from the formPage_TITLE and adding the location collumns in order to destinguish between the same FormPages of different site (on multi_site projects), with some priviliges for the superUsers only
image

### DIFF
--- a/wagtail/contrib/forms/templates/wagtailforms/list_forms.html
+++ b/wagtail/contrib/forms/templates/wagtailforms/list_forms.html
@@ -1,21 +1,33 @@
 {% load i18n %}
 <table class="listing">
-    <col width="50%"/>
+    <col width="25%"/>
+    <col width="25%"/>
     <col width="50%"/>
     <thead>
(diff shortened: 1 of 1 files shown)


lb-:
@hazho  thanks for giving a go at improving this, I agree that the current state of the form page listing is a bit confusing.

However, I do not think this approach aligns with many of the other listing / reporting interfaces we have in Wagtail at the moment.

Please see comment - https://github.com/wagtail/wagtail/issues/7483#issuecomment-1321110707

I think for now it is best to either close this until there is a clearer direction for how to improve this page or if you want to leave it open I have just added a label to flag this as not ready for a review.

lb-:
Sorry this took so long, I think it would have been better to get a bit of feedback on the issue first before going ahead with a solution. I know it takes a while sometimes though to get feedback.

Thanks for wanting to make Wagtail better.

Issue #7476, opened 2021-08-30
Namespace conflict between "comment" / "comments" and admin comments introduced in 2.13


awhileback:
Found a bug? Please fill out the sections below. 👍

### Issue Summary

This has been reported (#7343) but the person reporting it did not tag it as a bug and I think it should be a bug.

### Steps to Reproduce

1. install wagtail, make a basic page model
2. install [django-comments-dab](https://github.com/Radi85/Comment), add it to that page model to add comment functionality to said page
3. add some comments to an instance of the wagtail page
4. edit the instance of the wagtail page in the wagtail admin

Any other relevant information. For example, why do you consider this a bug and what did you expect to happen instead?

```
Internal Server Error: /admin/pages/45/edit/
Traceback (most recent call last):
  File "/Users/ausername/Dropbox/wagdev/siteenv/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/Users/ausername/Dropbox/wagdev/siteenv/lib/python3.8/site-packages/django/core/handlers/base.py", line 204, in _get_response
    response = response.render()
  File "/Users/ausername/Dropbox/wagdev/siteenv/lib/python3.8/site-packages/wagtail/admin/auth.py", line 187, in overridden_render
    return render()
  File "/Users/ausername/Dropbox/wagdev/siteenv/lib/python3.8/site-packages/django/template/response.py", line 105, in render
    self.content = self.rendered_content
  File "/Users/ausername/Dropbox/wagdev/siteenv/lib/python3.8/site-packages/django/template/response.py", line 83, in rendered_content
    return template.render(context, self._request)
  File "/Users/ausername/Dropbox/wagdev/siteenv/lib/python3.8/site-packages/django/template/backends/django.py", line 61, in render
    return self.template.render(context)
  File "/Users/ausername/Dropbox/wagdev/siteenv/lib/python3.8/site-packages/django/template/base.py", line 170, in render
    return self._render(context)
  File "/Users/ausername/Dropbox/wagdev/siteenv/lib/python3.8/site-packages/django/test/utils.py", line 100, in instrumented_test_render
    return self.nodelist.render(context)
  File "/Users/ausername/Dropbox/wagdev/siteenv/lib/python3.8/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
[...]

PR #7591, opened 2021-10-13
Change 'comments' relation name to avoid collisions

### LINKED ISSUE #7343: Namespace reverse accessor for comments on Page and User models
This is something I previously raised on Slack, but it's been nagging away at me so I wanted the input of the larger community before I worked on any sort of PR - I hope that's okay.

### The problem

Wagtail 2.13 introduced commenting within the Page editor as a feature (and it's ace!). The `Comment` model creates a relationship between `User` and `Page` using `related_name="comments"`:

```
class Comment(ClusterableModel):
    """
    A comment on a field, or a field within a streamfield block
    """
    page = ParentalKey(Page, on_delete=models.CASCADE, related_name='comments')
    user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE, related_name='comments') 
```

Thus reserving `comments` as a related name for those common models.

On sites that already make use of commenting functionality (for example, a blog that allows users to comment on articles), this can be problematic, as a comment is almost invariably linked to a `User` and `Page`. While I completely accept that Wagtail can't be expected to make concessions to specific user cases, commenting feels like commonplace enough functionality that this will be a problem other users encounter when upgrading.

### Proposed solution

I think we should consider namespacing the `reverse_name` to specifically reference that these are Wagtail comments, by prefixing with app name:

```
class Comment(ClusterableModel):
    """
    A comment on a field, or a field within a streamfield block
    """
    page = ParentalKey(Page, on_delete=models.CASCADE, related_name='wagtail_comments')
    user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE, related_name='wagtail_comments') 
```

And updating all references to `comments` throughout Wagtail to use `wagtail_comments`.

The alternative is that users rewrite the codebase for Wagtail based projects that already make use of `reverse_name="comments"`, however I feel it's much cleaner for future development if Wagtail keeps a clear division between editor functionality and that of first-party code when it comes to reserving common terminology (especially when dealing with non-Wagtail models like `User`).

Maybe I'm alone in this, but either way would definitely like to hear from the the core team and the wider community about the implications of this, and whether there is a need for a namespacing strategy on common models going forward.

Thanks!

gasman:
Fixes #7343 / #7476. Requires a corresponding update to django-modelcluster: https://github.com/wagtail/django-modelcluster/pull/143

Change the 'comments' relation name on the Page and User models to wagtail_admin_comments, and also make this configurable via the setting `WAGTAIL_COMMENTS_RELATION_NAME`. I propose that we apply this to 2.15 as is (with a version bump to django-modelcluster, and a release note advising developers that if they have any references to the 'comments' relation in their code, they should update it or set `WAGTAIL_COMMENTS_RELATION_NAME = 'comments'`), and for the 2.13 / 2.14 backports, apply a version with the setting defaulting to 'comments'. This will ensure that developers who weren't able to upgrade from 2.12 due to having a conflicting 'comments' relation can unblock themselves by changing this setting.

As far as I can see, this change shouldn't break wagtail-transfer, provided both instances set WAGTAIL_COMMENTS_RELATION_NAME to the same thing. The tests break because the fixtures refer to the 'comments' relation; setting `WAGTAIL_COMMENTS_RELATION_NAME = 'comments'` in the test settings fixes them.

Note: changing the relation name does have the side effect of changing the key used for the comments relation in the JSON data of page revisions, which means it won't be able to retrieve this data from old revisions when [updating comment positions](https://github.com/gasman/wagtail/blob/73810a944e0de3625959e991b1a7057aaa71c635/wagtail/core/models/__init__.py#L2292-L2301). Maybe this is obscure enough to let it slide? (Or should we hard-code 'comments' as a fallback?)

### DIFF
--- a/wagtail/core/models/__init__.py
+++ b/wagtail/core/models/__init__.py
@@ -2259,7 +2259,21 @@ class Page(AbstractPage, index.Indexed, ClusterableModel, metaclass=PageBase):
         * ``wagtail_admin_comments`` (COMMENTS_RELATION_NAME)
         """
 
-        obj = self.specific_class.from_json(content_json)
+        data = json.loads(content_json)
+
+        # Old revisions (pre Wagtail 2.15) may have saved comment data under the name 'comments'
+        # rather than the current relation name as set by COMMENTS_RELATION_NAME;
+        # if a 'comments' field exists and looks like our comments model, alter the data to use
+        # COMMENTS_RELATION_NAME before restoring
+        if (
+            COMMENTS_RELATION_NAME not in data and 'comments' in data
+            and isinstance(data['comments'], list) and len(data['comments'])
+            and isinstance(data['comments'][0], dict) and 'contentpath' in data['comments'][0]
+        ):
+            data[COMMENTS_RELATION_NAME] = data['comments']
+            del data['comments']
+
+        obj = self.specific_class.from_serializable_data(data)
 
         # These should definitely never change between revisions
         obj.id = self.id


gasman:
wagtail-localize currently hard-codes a check for the name `comments`:
https://github.com/wagtail/wagtail-localize/blob/8ee52c861b07309a6def36d55efde6d9c67b9ebd/wagtail_localize/fields.py#L173

To fix this in a backwards-compatible way (including current 2.13 - 2.14 releases), we're going to need a reliable way for non-core code to find out what the relation name is. I think the least worst solution is to make it a constant on the wagtail.core.models module, so the backwards-compatible incantation becomes

    try:
        from wagtail.core.models import COMMENTS_RELATION_NAME
    except ImportError:
        COMMENTS_RELATION_NAME = 'comments'

I thought about making it an attribute of the Page model (somewhat like User.USERNAME_FIELD), but I think that will irrevocably couple comments to the Page model in a way that we don't want (presently, all logic relating to the comment relation lives either outside the Page model, or in methods we'd like to move out of the Page model in our upcoming Python refactoring) and logically we'd have to make it available on User too, which is even more hacky.

jacobtoppm:
That makes sense to me. Currently, comments are fairly linked to Page, but there's no reason we couldn't broaden that concept to "a clusterable model with revisions" in the future, so the constant makes sense to me (having it on the Page model would I think make most sense if you could set a per-model reverse relation name, and that definitely seems like overkill to me)

Issue #7584, opened 2021-10-09
Admin filters don't use the color-primary variables


Ash-Crow:
Hi !

### Issue Summary
<img align="right" width="200" src="https://user-images.githubusercontent.com/765956/136667005-c7132877-101a-46f0-8199-dca401db6cad.png">
The filters in the admin panel have hard-coded colors in the CSS, so they stay teal even if the primary color is changed.

The colors are a bit lighter than the ones defined by the variables (see the attached picture), but I think using the existing ones (so making the filters use the same color as the main interface) would be better than adding two new variations on the registered colors.

That means:
- replacing #43b1b0 (used for the border/text of unselected filters and the background of selected ones) with color-primary
- replacing #358c8b (used for the :hover pseudo-class) with color-primary-dark

This is the first issue I post here, so I'm not sure how to proceed: can I just submit a PR that makes the change described above or should this change be discussed somewhere beforehand (and if so, where?)

### Steps to Reproduce

1. Install a new wagtail and wagtailbakery
2. Customize the admin templates colors as specified in the [documentation](https://docs.wagtail.io/en/latest/advanced_topics/customisation/admin_templates.html#custom-user-interface-colors) so that they are another color than teal, for example this dark blue :

```
:root {
  --color-primary-hue: 240;
  --color-primary-saturation: 100%;
  --color-primary-lightness: 21%;
}
```
3. Navigate to a page that has a filter sidebar (for example, [Bakery Misc > People](http://127.0.0.1:8000/admin/base/people/) in the wagtailbakery demo site)
4. See that the colors of the filters is still teal, while the rest of the interface uses the defined color:

![Capture d’écran de 2021-10-09 19-09-47](https://user-images.githubusercontent.com/765956/136667910-ced8604f-a7cb-4594-8772-b2409b190302.png)

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: yes

### Technical details

* Python version: 3.9.5
* Django version: 3.2.8
* Wagtail version: 2.14.1
* Browser version: Firefox 93

PR #7612, opened 2021-10-15
replace hard-coded colors for admin filters with variables per #7584

nthall:
This commit is a small straightforward fix for issue #7584.


tomdyson:
Also many thanks to @Ash-Crow, whose detailed description of the issue made it easy for us to understand the problem and suggest an approach: https://github.com/wagtail/wagtail/issues/7584

PR #7629, opened 2021-10-18
Display the select option's text instead of value in telepath

jeromelebleu:
`BoundWidget.getTextLabel()` returns the value of the widget, which is the value of the option for a `<select>` element. This adds a new widget for `<select>` elements to display instead the text of the option. It would be more relevant for the user when the block is collapsed.


gasman (approved):
LGTM - thanks @jeromelebleu!

gasman:
Merged in e1ecca2f3c7a693fbc0f10cfa54cf996011ebbda

PR #7638, opened 2021-10-21
Add more SketchFab oembed patterns

tomusher:
Adding a few additional patterns for the Sketchfab oEmbed provider.

- The existing `/show` pattern is not referenced anywhere on the Sketchfab site, but still seems to work when passed to the oEmbed endpoint.
- The `/models` pattern is what is documented in the [Sketchfab oEmbed docs](https://sketchfab.com/developers/oembed).
- The `/3d-models` pattern is what you'd get by copying and pasting the URL from browsing the site - this seems to work when passed to the oEmbed endpoint too.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -5,6 +5,7 @@ Changelog
 ~~~~~~~~~~~~~~~~~
 
  * Added Aging Pages report (Tidjani Dia)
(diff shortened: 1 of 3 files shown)


lb-:
@tomusher thanks for this & thanks for reviewing @zerolab 
merging in.

PR #7649, opened 2021-10-23
Fix gulp watch to correctly watch files

Wicked7000:
This is an issue I ran into while trying to modify the style for some of the high contrast issues. Gulp didn't seem to be watching files correctly (it was correctly building them on the first time around but any subsequent changes weren't picked up). This turned out to be down to the cwd not being set correctly. 

Also the client styles path was being added numerous times due to it being used in a reduce statement, so I pulled it out and just added it once after the reduce is done.

Let me know if there is any concerns with the changes below!


lb- (approved):
@Wicked7000 validated locally, working great thanks.

I have also taken this opportunity to format this js file in a consistent way & pushed the changelog commit to your remote branch.

PR #7650, opened 2021-10-24
Navigation support for page-editing dropdown

Desaiakshata:
Follow up for #7525
Tested on: Windows 10
Google Chrome: Latest version
Frontend tests pass

Added keyboard navigation to the edit page actions dropdown. 
Added a separate file for dropdowns

### DIFF
--- a/client/scss/components/_dropdown.legacy.scss
+++ b/client/scss/components/_dropdown.legacy.scss
@@ -9,7 +9,7 @@
     .button {
         padding: 0 5em 0 1em;
         display: block;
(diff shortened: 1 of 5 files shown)


thibaudcolas:
Note #7525 already has a code review from LB.

thibaudcolas:
@Desaiakshata I’m going to close this now as I don’t see ourselves wanting to invest more time in this approach. We’d rather be rebuilding the widget with more proper foundations, which is already under way in #9030.

PR #7653, opened 2021-10-26
Fixes #7450 Issues with Dropdowns in High contrast mode

### LINKED ISSUE #7450: High-contrast mode: dropdown menu needs custom styles
This is a bug specific to [Windows high-contrast mode](https://blogs.windows.com/msedgedev/2020/09/17/styling-for-windows-high-contrast-with-new-standards-for-forced-colors/). cc @kbayliss who did the testing for this.

There are two issues – first, the trigger needs a border so it can be identified as a button:

7450 png

Then, we need to customise the styles of the opened dropdown so its contents stands out better compared to the top-level button:

Wicked7000:
Fixes #7450 Issue with high contrast examples of how the changes look are shown below:
Chrome:
DropdownFix

DropdownFix2

As with my previous MR its worth noting there are significant differences in Firefox as it seems to be a lot more restrictive with what custom styling it allows, which is seemingly due to the fact it doesn't currently support forced-color-adjust. (Although hopefully once it begins to support that all the changes that apply to Chrome/Safari should easily transfer over)

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -34,7 +34,7 @@ Changelog
  * Improve display of image listing for long image titles (Krzysztof Jeziorny)
  * Use SVG icons in admin home page site summary items (Jérôme Lebleu)
  * Ensure site summary items wrap on smaller devices on the admin home page (Jérôme Lebleu)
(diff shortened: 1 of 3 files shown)


lb- (approved):
Works great, thanks @Wicked7000  - I appreciate that you added recordings of before/after - really helps.

PR #7657, opened 2021-10-26
Make cross-link URLs on generic model views optional

gasman:
Normally, generic model views will be spun up through wagtail.admin.viewsets, which takes care of configuring the add_url_name / edit_url_name etc attributes to ensure that they are cross-linked appropriately. But if you're defining one in isolation, that's an unnecessary hoop to jump through, and the error messages if you don't specify them are fairly opaque. Fix this so that:

1. Non-essential cross-links (e.g. the Add button on the index view, and the Delete button on the edit view) are gracefully omitted if the URLs are undefined;
2. Essential cross-links (e.g. the destination of the redirect after a successful form submission) raise an informative error telling the developer what needs to be defined.


gasman:
This should fix #7758.

lb- (approved):
Tested on
* Collections index, inc. add/edit/delete
* Snippets index, inc. add/edit/delete
* Collections index, inc. add/edit/delete
* Workflows index, inc. add/edit/delete
Looks good

PR #7659, opened 2021-10-28
rename admin/expanding-formset.js (using kebab-case file)

lb-:
* small JS file clean up - the file was named `expanding_formset.js` not `expanding-formset.js` (all other similar JS files are kebab-case not snake_case).


lb-:
@gasman agreed. Thanks - I'll add that to this PR

PR #7663, opened 2021-10-29
Fixes #7241 and improves error visibility in high-contrast

### LINKED ISSUE #7241: Improve visibility of error messages

When ever I have validation errors when trying to save a page (or other model) in the Wagtail admin, I find it hard to identify the field with the error. 

I believe this is particularly hard for me as I have a form of "deuteranomalia" which causes "red" for me to be less saturated than for other people. As a result, the small red error text appears basically grey to me and is hard to distinguish from the rest of the page content. 

On simple field like text inputs this is less annoying since the entire field changes its styling. But, especially on chooser fields there is no indication apart from the small text. 

I could imagine the style to be updated in the following order of usefulness (most to least):

- Add a icon at the beginning of the error text (something like this maybe ❌)
- Use a bold font. 
- Increase the font size. 
- Use a brighter red.

I understand that might be ugly from a design perspective, but this is only visible when something is amiss and important information needs to be transported to the user. 

Not much I can do but slowly scroll the pages. 

### Additional context

Here is an example comparison:

Screenshot 2021-06-07 at 13 22 07

Screenshot 2021-06-07 at 13 23 48

Here is the changed `.error-message` class I have used:

```css
.error-message {
    color: #d93030;
    font-weight: bold;
    font-size: small;
}
```

Wicked7000:
This fixes the accessibility issue mentioned in #7241 and also fixes issue with error visibility in high-contrast mode (before this the message was white and indistinguishable from other text on the page)

Before:
image
image

After:
image
image

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -19,6 +19,7 @@ Changelog
  * Improve CircleCI frontend & backend build caches (Thibaud Colas)
  * Add a 'remember me' checkbox to the admin sign in form, if unticked (default) the auth session will expire if the browser is closed (Michael Karamuth, Jake Howard)
  * When returning to image or document listing views after editing, filters (collection or tag) are now remembered (Tidjani Dia)
+ * Improve the visibility of field error messages, in Windows high-contrast mode and out (Jason Attwood)
  * Fix: Accessibility fixes for Windows high contrast mode; Dashboard icons colour and contrast (Sakshi Uppoor)
  * Fix: Rename additional 'spin' CSS animations to avoid clashes with other libraries (Kevin Gutiérrez)
  * Fix: `default_app_config` deprecations for Django >= 3.2 (Tibor Leupold)
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -557,6 +557,7 @@ Contributors
 * Michael Karamuth
 * Vu Pham
 * Khanh Hoang
+* Jason Attwood
 
 Translators
 ===========
--- a/client/scss/components/_forms.scss
+++ b/client/scss/components/_forms.scss
@@ -123,7 +123,20 @@ select::-ms-expand {
 }
 
 .error-message {
-    color: $color-red;
+    font-size: 1em;
+    font-weight: bold;
+    color: $color-text-error;
+
+    @media (forced-colors: $media-forced-colours) {
+        forced-color-adjust: none;
+        color: $color-text-error-forced-color;
+    }
+
+    &::before {
+        font-family: wagtail;
+        vertical-align: -10%;
+        content: map-get($icons, 'cross');
+    }
 }
 
 .help {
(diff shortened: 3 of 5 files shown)


nmorduch (changes requested):
Thanks for making this change! It looks great.

One request: I'd like to put all new hex colors in [variables.scss](https://github.com/wagtail/wagtail/blob/main/client/scss/settings/_variables.scss) for consistency, and ideally base them off of the existing colors. I would add something like the following in `variables.scss` (but simply moving the hex colors there would also work), then reference those colors here.

``` scss
$color-text-error: change-color($color-red, $saturation: 69, $lightness: 52);
$color-text-error-forced-color: change-color($color-red, $saturation: 100, $lightness: 50);
```

Let me know if any of that isn't clear enough, and I'm happy to clarify it :)

Sorry I missed the end of October for this, if when it's reviewed makes a difference! If you no longer have time to devote to this, let me know and I'd be happy to update the PR with this change.

Wicked7000:
> Thanks for making this change! It looks great.
> 
> One request: I'd like to put all new hex colors in [variables.scss](https://github.com/wagtail/wagtail/blob/main/client/scss/settings/_variables.scss) for consistency, and ideally base them off of the existing colors. I would add something like the following in `variables.scss` (but simply moving the hex colors there would also work), then reference those colors here.
> 
> ```css-scss
> $color-text-error: change-color($color-red, $saturation: 69, $lightness: 52);
> $color-text-error-forced-color: change-color($color-red, $saturation: 100, $lightness: 50);
> ```
> 
> Let me know if any of that isn't clear enough, and I'm happy to clarify it :)
> 
> Sorry I missed the end of October for this, if when it's reviewed makes a difference! If you no longer have time to devote to this, let me know and I'd be happy to update the PR with this change.

No worries! I've updated to put the adjusted colours in the variables file.

I forced pushed (as I assume you'd likely want commit history to be 1 Commit per MR/Feature) so I can't mark the requested change as completed.

Let me know if there is any other additional changes needed!

nmorduch:
Looks good to me! Thank you! I'm going to ask that someone who knows more about the Windows high contrast setup review that aspect :)

vsalvino:
I can test out Windows High contrast stuff if needed... let me know what areas to hit (if you have any kind of check-list) and I'd be happy to help.

Wicked7000:
> I can test out Windows High contrast stuff if needed... let me know what areas to hit (if you have any kind of check-list) and I'd be happy to help.

The main area I used for testing was by using the form shown in Settings -> Users -> Add a user 
Then clicking over to the roles tab without correctly filling out some or all fields and you should get the error notification to come up.

However I'm sure this is just one of many forms on the Wagtail admin page so I'm sure others can give pointers to others if needed to check over multiple form areas.

vsalvino:
Tested this out, error messages on the fields seem to be visible, always in red regardless of varying high contrast setting. The error banner at the top could probably use some visual treatment but is perfectly readable at least. Looks good to me overall.

Here is what I am seeing when making a page edit with error:

High Contrast \#1:

image

High Contrast \#2:

image

High Contrast Black:
image

High Contrast White:

image

vsalvino (approved):

thibaudcolas (approved):
Thank you! Looks amazing, I’m very happy we get this in. Font icons are going away pretty soon (#6107), so we’ll need to switch this to SVGs, but this should be no problem when that happens.

Issue #7664, opened 2021-10-31
Inconsistent output from RichTextField and RichTextBlock


gajus11:
### Issue Summary

Inconsistent output from RichTextField and RichTextBlock (Wagtail API v2).
 
For example if we use Internal link handled by PageLinkHandler:

RichTextField output:
`<html><body><p data-block-key=\"ca20z\">Test internal link</p><p data-block-key=\"d12am\"><a href=\"/\">Home</a></p></body></html>`

RichTextBlock output:
`<p data-block-key=\"s1g1a\">Test internal link</p><p data-block-key=\"b5i7\"><a id=\"5\" linktype=\"page\">Home</a></p>`

As you can see `a` tags are different.
RichTextBlock doesn't return actual path.

### Steps to Reproduce

1. Insert "Internal link" into RichTextField
2. Insert the same "Internal link" into RichTextBlock
3. Check outputs

gasman:
This is covered by #2695 (and the discussion there gives some workarounds).

PR #7665, opened 2021-10-31
Fix Inconsistent output from RichTextField and RichTextBlock #7664

gajus11:
Fix Inconsistent output from RichTextField and RichTextBlock #7664


gasman:
Thanks @gajus11! This fix isn't correct, though - as per https://docs.wagtail.io/en/stable/advanced_topics/customisation/rich_text_internals.html#data-format, rich text data is intentionally stored in the database in 'symbolic' form (e.g. with page IDs instead of URLs, so that they don't break if pages are moved or renamed) and converted to real HTML when output on the front-end. Changing `get_prep_value` in this way would mean that we start storing the converted HTML in the database instead.

Issue #7646, opened 2021-10-22
Accessibility and UI issues with typed table block


Scotchester:
Starting this issue to catalogue issues that remain in the [typed table block](https://docs.wagtail.io/en/latest/reference/contrib/typed_table_block.html) UI, especially accessibility-related ones.

I am already working on a few of these, denoted below.

**Note:** We also know we need to take a harder look at how the ARIA Grid pattern might serve this interface well, and using that may nullify some of the issues below.

## Keyboard navigation

- [ ] After creating a new column, focus should either move to the heading input of that column or back to the new column button (for easier insertion of multiple columns)
  - [ ] After creating a new row, focus should move to the first column input in that row, if the decision in the item above was to move to the heading input
- [ ] Should use ARIA Grid pattern

## Screen reader support

- [ ] After inserting a column, there is no indication of what type of column it is
- [ ] Should read the current row and column positions (e.g., "row 2, column 3") whenever changing cells or after inserting/removing rows/columns if that action changes the current position
- [ ] Column heading inputs need labels @Scotchester 
- [ ] Add column menu should have an `aria-expanded` update, defaulting to `false` and updating to `true` when open @Scotchester 

## General UI

- [x] Use SVG plus icons for add buttons for better visibility @Scotchester 
- [x] Add borders between rows to help know which row the delete button applies to
- [ ] Add missing bottom border in Safari https://github.com/wagtail/wagtail/pull/7670#pullrequestreview-853537536
- [ ] Confirmation of row/column deletion would be nice, to protect against accidental data loss
- [ ] Change plus icon on add column button to an x when clicked to indicate that you click that button again to close the add column menu

mohammadareeb95:
@Scotchester I would like on work on some General UI fixes

lb-:
@mohammadareeb95 I encourage to to read through the whole issue, all comments and all related PRs - see what gaps there are and maybe suggest something specific you can reproduce and give a go at fixing.

thibaudcolas:
The change proposed in #7667 is ready for someone else to take on, with a few tweaks:

- The whole label needs to be set up for translation, _not_ concatenation of translateable text.
- We need to add the index of the column. Something like `Column heading for column {index}, {column type}`

PR #7667, opened 2021-10-31
TypedTableBlock: Add `aria-label` to column heading input

Scotchester:
All inputs need a valid label of some kind, so this PR adds an `aria-label` attribute to the column heading inputs, which previously didn't have a label.

(Part of #7646)

---

    - Firefox 88 on macOS 10.15.7


client/src/entrypoints/contrib/typed_table_block/typed_table_block.js
```diff
@@ -235,6 +235,9 @@ export class TypedTableBlock {
     column.headingInput.name = this.prefix + '-column-' + column.id + '-heading';
     column.headingInput.className = 'column-heading';
     column.headingInput.placeholder = this.blockDef.meta.strings.COLUMN_HEADING;
+    column.headingInput.setAttribute(
+      'aria-label', `${h(this.blockDef.meta.strings.COLUMN_HEADING)} for ${h(blockDef.meta.label)} column`
```
gasman, line 239: The `COLUMN_HEADING` string (and potentially blockDef.meta.label) is translated, so for people using other languages, this is likely to end up as a mix of English and their own language. This would be fixable by adding 'column heading for X' as [another translatable string](https://github.com/wagtail/wagtail/blob/24b99a5ebc4b25986b89bbbfa0de6e93acfe6dcc/wagtail/contrib/typed_table_block/blocks.py#L286) - we don't currently have any provision for placeholders in those strings, but it would be easy enough to implement one with a simple string search and replace.

I'm wondering if `blockDef.meta.label` is the most useful thing to use to identify the column, though, given that it's typically going to be something generic like "rich text column"? I'm assuming we can't just say "column heading" because we're expected to make them unique, but adding the column type doesn't guarantee that either. Maybe we can use the column number instead - "column heading for column 1"?

thibaudcolas: Yes, I think further work is needed here. Even without the `aria-label`, the field is already announced as "Column heading" thanks to its placeholder, so it’s not so bad. Having the column number would indeed be nice so it’s possible to distinguish between them when navigating the page linearly.

I’d recommend running a few options past @wagtail/accessibility, as it doesn’t feel clear to me which labels would be most helpful.

thibaudcolas (changes requested):

client/src/entrypoints/contrib/typed_table_block/typed_table_block.js
SaptakS: Placeholders are really not prefered as the text label to be used by screen readers since it is not a standard, so preferably a `<label>` tag with some class to hide it visually if needed, or at least an aria-label is definitely needed. PS: https://www.htmhell.dev/24-a-placeholder-is-not-a-label/ , https://www.smashingmagazine.com/2018/06/placeholder-attribute/ , https://codepen.io/stevef/post/placeholder-the-piss-take-label

thibaudcolas: @SaptakS by "which labels would be most helpful", I meant the textual content rather than mean of implementation. In a table like this it feels like each cell field should announce its row/column number (or "column 1 heading" in the case of the `th`), but this could get pretty verbose.

The placeholder is taken into account in the accessible name computation (see [HTML-AAM](https://www.w3.org/TR/html-aam-1.0/#accessible-name-and-description-computation)) as the last resort, hence why I think this can wait until we have the best label possible. I still wouldn’t recommend placeholder over `aria-label`, but at least there is _something_ announced currently.

SaptakS: Agree with everything you said. Just wanted to note that we should finally at some point replace placeholder with proper label, since many screen readers don't respect placeholders.

thibaudcolas:
Assigning to @Scotchester and I to take to the next accessibility team discussions.

lb-:
Has this been discussed at any of the accessibility meetings?

thibaudcolas:
Closing with guidance in #7646

PR #7668, opened 2021-10-31
TypedTableBlock: Add `aria-expanded` attribute to add new column button

Scotchester:
Adding `aria-expanded` attribute to the add new column menu trigger button so that assistive technology knows when it has opened its menu.

(Part of #7646)

---

    * Firefox 88 on macOS 10.15.7

### DIFF
--- a/client/src/entrypoints/contrib/typed_table_block/typed_table_block.js
+++ b/client/src/entrypoints/contrib/typed_table_block/typed_table_block.js
@@ -141,6 +141,7 @@ export class TypedTableBlock {
   showAddColumnMenu(baseElement, callback) {
     this.addColumnMenuBaseElement = baseElement;
     baseElement.after(this.addColumnMenu);
+    baseElement.attr('aria-expanded', 'true');
     this.addColumnMenu.show();
     this.addColumnCallback = callback;
   }
@@ -221,7 +222,8 @@ export class TypedTableBlock {
     const prependColumnButton = $(`<button type="button"
       class="button button-secondary button-small button--icon text-replace prepend-column"
       aria-label="${h(this.blockDef.meta.strings.INSERT_COLUMN)}"
-      title="${h(this.blockDef.meta.strings.INSERT_COLUMN)}">
+      title="${h(this.blockDef.meta.strings.INSERT_COLUMN)}"
+      aria-expanded="false">
         <svg class="icon icon-plus icon" aria-hidden="true" focusable="false"><use href="#icon-plus"></use></svg>
       </button>`);
     $(newHeaderCell).append(prependColumnButton);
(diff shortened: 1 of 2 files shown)


client/src/entrypoints/contrib/typed_table_block/typed_table_block.js
```diff
@@ -141,6 +141,7 @@ export class TypedTableBlock {
   showAddColumnMenu(baseElement, callback) {
     this.addColumnMenuBaseElement = baseElement;
     baseElement.after(this.addColumnMenu);
+    baseElement.attr('aria-expanded', 'true');
```
gasman, line 144: Presumably there ought to be a corresponding `baseElement.attr('aria-expanded', 'false');` on `hideAddColumnMenu`.

I think that when I originally wrote this code and chose the name `baseElement`, I didn't necessarily expect that to refer to the trigger button element - I think it was just the element that the menu dropdown was positioned relative to. Now that this change is effectively setting in stone that it is indeed the button, this is probably a good time to rename it accordingly - `baseElement` -> `triggerElement` and `addColumnMenuBaseElement` -> `addColumnMenuTrigger`?

client/src/entrypoints/contrib/typed_table_block/typed_table_block.js
```diff
@@ -221,7 +222,8 @@ export class TypedTableBlock {
     const prependColumnButton = $(`<button type="button"
       class="button button-secondary button-small button--icon text-replace prepend-column"
       aria-label="${h(this.blockDef.meta.strings.INSERT_COLUMN)}"
-      title="${h(this.blockDef.meta.strings.INSERT_COLUMN)}">
+      title="${h(this.blockDef.meta.strings.INSERT_COLUMN)}"
+      aria-expanded="false">
         <svg class="icon icon-plus icon" aria-hidden="true" focusable="false"><use href="#icon-plus"></use></svg>
```
gasman, line 227: There's also an `appendColumnButton` (`data-append-column` in the initial HTML fragment near the top of the script) that needs the same change. Perhaps it'll be clearer to set this attribute alongside the point where we attach the onclick handler (`this.appendColumnButton.on('click', ...)`, `this.prependColumnButton.on('click', ...)`) rather than putting it in the initial HTML fragments?

gasman (changes requested):

lb-:
Labelling as needs work, review was done (some feedback given) and now needs a rebase. 

Are you still up for this @Scotchester or maybe someone else can pick up from here?

Scotchester:
I'm happy for anyone to pick it up, if they like. If I find time to work on it again myself, I'll post another comment here.

ayaan-qadri:
Going to give go to redoing this, I have already started !

ayaan-qadri:
I found few things while working on this:

1. `hide()`
    - I noticed that jQuery hide method uses inline css to hide or unhide element
    - I think this will break if CSP is enabled, I'm not sure about this but once i got to know from @lb-  https://github.com/wagtail/sphinx-wagtail-theme/issues/305#issuecomment-2466612908

<details>
<summary>Screenshot</summary>

https://github.com/user-attachments/assets/7e34db36-5c96-4909-9053-e8bff9e580ff

</details>

2. On some point opening w-combobox goes out of the visible area
<details>
<summary>Screenshot</summary>

     image

</details>

3. There should be a message closing button because someone like me, who is used to closing message boxes, will find it irritating not to have that option. currently, we have to reload the page to remove the message box.

<details>
<summary>Screenshot</summary>

image

</details>

gasman:
Superseded by #12680

PR #7669, opened 2021-10-31
Add new field  `alias_of` to default endpoint fields

### LINKED ISSUE #7009: Add alias_of to page meta in api response
We started using aliases and needed to find a way to determine if the current page is an alias in order to set a rel=canonical header pointing to the original page.

The API currently doesn't provide a way to distinguish a normal page from an alias.

As noted in [this stackoverflow answer](https://stackoverflow.com/questions/66581694/expose-page-alias-in-wagtail-api/67107694) this can be achieved be serializing the alias_of relation.

```
from rest_framework import serializers

from wagtail.api import APIField

class AliasPageSerializer(serializers.ModelSerializer):
    class Meta:
        model = Page
        fields = ('id', 'full_url')

class BasePage(Page):
    class Meta:
        abstract = True

    api_meta_fields = [
        APIField('alias_of', serializer=AliasPageSerializer()),
    ]
```

Would you consider adding this to Wagtail?

scientia-ac-labore:
This PR adds a new field `alias_of` for api. If the page marked as an alias `alias_of` return original page dictionary contain `id` and `full_url`. If not return `null`. Docs and tests were changed according to this field. Close #7009

### DIFF
--- a/docs/advanced_topics/api/v2/usage.rst
+++ b/docs/advanced_topics/api/v2/usage.rst
@@ -585,6 +585,10 @@ Pages
         Nests some information about the parent page (only available on detail
         views)
 
+    ``meta.alias_of`` (dictionary)
+
+        If the page marked as an alias return original page id and full url
+
 Images
 ------
 
--- a/wagtail/admin/tests/api/test_pages.py
+++ b/wagtail/admin/tests/api/test_pages.py
@@ -173,7 +173,7 @@ class TestAdminPageListing(AdminAPITestCase, TestPageListing):
 
         for page in content['items']:
             self.assertEqual(set(page.keys()), {'id', 'meta', 'title', 'admin_display_title', 'date', 'related_links', 'tags', 'carousel_items', 'body', 'feed_image', 'feed_image_thumbnail'})
-            self.assertEqual(set(page['meta'].keys()), {'type', 'detail_url', 'show_in_menus', 'first_published_at', 'seo_title', 'slug', 'parent', 'html_url', 'search_description', 'locale', 'children', 'descendants', 'ancestors', 'translations', 'status', 'latest_revision_created_at'})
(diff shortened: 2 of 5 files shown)


kaedroho (approved):
Sorry for the delay in reviewing this. Looks good to me!

jaap3:
Just noticed that this now uses `full_url` instead of `html_url`, this is inconsistent with how the urls are serialized in Wagtail. Other places always use the `PageHtmlUrlField` to serialize the `full_url`.

gasman:
> Just noticed that this now uses full_url instead of html_url, this is inconsistent with how the urls are serialized in Wagtail. Other places always use the PageHtmlUrlField to serialize the full_url.

@kaedroho Is this something to worry about? I guess this is going to be hard to change retrospectively after it ships.

jaap3:
It could be made more consistent by repeating what is done in `PageParentField` (declared right above `PageAliasOfField`) which is more complex, but that might be because of good reasons, i.e.:

```
class PageAliasOfField(relations.RelatedField):
    """
    Serializes the "alias_of" field on Page objects.
    """
    def get_attribute(self, instance):
        return instance

    def to_representation(self, value):
        serializer_class = get_serializer_class(value.__class__, ['id', 'type', 'detail_url', 'html_url', 'title'], meta_fields=['type', 'detail_url', 'html_url'], base=PageSerializer)
        serializer = serializer_class(context=self.context)
        return serializer.to_representation(value)
```

kaedroho:
> @kaedroho Is this something to worry about? I guess this is going to be hard to change retrospectively after it ships.

Sorry I missed this in review. Yes it would be hard for us to change this once it ships

scientia-ac-labore:
@kaedroho What should I do now? Should I make changes as @jaap3 propose or leave this to somebody with experience.

kaedroho:
@scientia-ac-labore That would be great if you can make the changes that @jaap3 described. Thanks!

> which is more complex, but that might be because of good reasons, i.e.

I think this was to reuse the serializers for those fields to keep things consistent if we ever changed them

Issue #7336, opened 2021-07-12
Change release notification to link to latest major release notes


kaedroho:
Wagtail has a notifcation on the dashboard that is shown if the Wagtail installation is out of date:

![image](https://user-images.githubusercontent.com/1093808/125284628-95aca180-e311-11eb-82cf-797a38e2d7d1.png)

This notification always links to the latest patch release. So clicking the link in the above screenshot will take the user to https://docs.wagtail.io/en/stable/releases/2.13.3.html

The problem with this is that this Wagtail site is a major version behind, so the release notes of the point release aren't relevant to this user since we only fix new regressions in point releases.

We should instead be pointing to the [2.13 release notes](https://docs.wagtail.io/en/stable/releases/2.13.html), which is much more relevant and interesting to someone on Wagtail 2.12.

kaedroho:
@gasman please could you add a separate link to the major release notes to releases.wagtail.io/latest.txt ?

gasman:
@kaedroho Sure. I figure we might as well add the data for the latest LTS release at the same time, so that we can potentially add a settings flag for people on LTS releases to ignore notifications for non-LTS upgrades - how about the following structure?

```
{
    "version": "2.13.4",
    "url":     "https://docs.wagtail.io/en/stable/releases/2.13.4.html",
    "majorUrl": "https://docs.wagtail.io/en/stable/releases/2.13.html",
    "lts": {
        "version": "2.11.8",
        "url": "https://docs.wagtail.io/en/stable/releases/2.11.8.html",
        "majorUrl": "https://docs.wagtail.io/en/stable/releases/2.11.html"
    }
}
```

If that looks good, let me know and I'll upload that to mark today's (non-urgent) 2.13.4 release.

kaedroho:
@gasman Looks good to me!

gasman:
OK, done (but it may take 24h for the cache to clear).

tbrlpld:
@gasman I have created #7671 to adjust the `latest.txt` according to your comment. Let me know if that works.

tbrlpld:
@gasman Do you have a preferred name for the LTS upgrade notification flag setting? 

Or should that whole topic be covered in a separate PR?

gasman:
@tbrlpld A separate PR for the LTS handling makes sense, I think.

tbrlpld:
Added the follow-up PR with the LTS: #7752

PR #7671, opened 2021-11-01
Apply semver naming for minor release url

tbrlpld:
The current naming is conflicting with how the version number parts are
called accoring to semantic versioning. Wagtail does not strictly follow
semantic versioning, but it still seems reasonable to adopt the same
terminology as the rest of the world.

See related discussion in #7336 (in particular [this comment by @gasman](https://github.com/wagtail/wagtail/issues/7336#issuecomment-954223018))

### DIFF
--- a/scripts/latest.txt
+++ b/scripts/latest.txt
@@ -1,10 +1,10 @@
 {
     "version": "2.14.2",
     "url":     "https://docs.wagtail.io/en/stable/releases/2.14.2.html",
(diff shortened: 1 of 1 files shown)


tbrlpld:
@gasman @kaedroho I was wondering if one of you could take a look at this one to unblock #7336

gasman (approved):

tbrlpld:
Thanks @gasman

PR #7675, opened 2021-11-03
Provide and use get_add_url method in generic index view

jeromelebleu:
This will allow one to have more control over the `add_url` construction - e.g. reverse the URL with arguments - and comply with other URLs resolution - i.e. a `get_<view>_url` is always used instead of accessing `<view>_url_name` directly.

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/generic/index.html
+++ b/wagtail/admin/templates/wagtailadmin/generic/index.html
@@ -4,9 +4,8 @@
 {% block titletag %}{{ page_title }} {{ page_subtitle }}{% endblock %}
 
 {% block content %}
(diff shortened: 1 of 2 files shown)


gasman:
Thanks @jeromelebleu! This is actually already included in #7657 along with a few other fixes, so I'll close this as a duplicate.

jeromelebleu:
Oops, sorry, I had not seen it. This looks much better, do you think it will be included in 2.15 please?

gasman:
No, 2.15 has been in Release Candidate stage since October 15th so it'll only be receiving critical bug fixes - any new feature code will be in 2.16 now.

PR #7677, opened 2021-11-03
Add Assistiv Labs sponsorship info to the README

thibaudcolas:
Advertising how [Assistiv Labs](https://assistivlabs.com/) sponsors Wagtail. Having this here is a precondition so our account gets created, once merged I’ll reply back and share the account details on Slack.

The image will only render once this is merged to `main`.


Stormheg (approved):

zerolab:
Nice one, @thibaudcolas 

Would be great to follow this up with a change to make the all logos have a more consistent size

Screenshot 2021-11-03 at 20 37 32

PR #7682, opened 2021-11-05
Implement CopyPageAction and admin API for copying pages

kaedroho:
This PR introduces a new "actions" folder in Wagtail core that contains classes to hold logic for page actions such as copying, publishing, moving, deleting, etc.

The purpose of these action classes are to:

- Provide one implementation of the action that can be called from both Django views and an API
- Reduce the bloat of the models by moving the whole implementation into the action class (the model will call the action class for backwards compatibility)
- Provide an object that can be passed to a background worker for asynchronous execution of actions (for the future)

On the admin API side, this will add the first writeable views. We don't plan to use these anywhere yet, but later they will become part of a documented and stable API that can be used from external applications. The plan for now is to not implement any documentation or authentication until after we've implemented all the actions like this and implemented autosave for the page editor.

This PR provides a template for reimplementing the other actions. Those other actions are:

- [ ] ``MovePageAction``  #7707
- [ ] ``DeletePageAction``  #7723
- [ ] ``PublishPageRevisionAction``  #7728
- [ ] ``RevertToPageRevisionAction``
- [ ] ``UnpublishPageAction``  #7740 
- [ ] ``CopyForTranslationPageAction``
- [ ] ``CreateAliasPageAction``
- [ ] ``ConvertAliasPageAction``

The following operations won't be reimplemented as action classes:

 - Page creation and editing. We will think about what to do with this when we implement autosave since it's important the Admin API works with that (even if it isn't implemented on the Admin API)
 - Approve / Reject moderation - These are legacy actions which may be removed some day
 -  Workflow actions - I'm not sure if these will fit into this model since workflow actions have their own way of being defined

### DIFF
--- a/wagtail/admin/api/actions/base.py
+++ b/wagtail/admin/api/actions/base.py
@@ -1,0 +1,6 @@
+class APIAction:
+    serializer = None
+
+    def __init__(self, view, request):
+        self.view = view
+        self.request = request
--- a/wagtail/admin/api/actions/copy.py
+++ b/wagtail/admin/api/actions/copy.py
@@ -1,0 +1,67 @@
+from django.core.exceptions import ValidationError as DjangoValidationError
+from django.shortcuts import get_object_or_404
+from rest_framework import fields, status
+from rest_framework.exceptions import ValidationError
+from rest_framework.response import Response
+from rest_framework.serializers import Serializer
+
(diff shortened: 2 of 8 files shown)


gasman (approved):
All looks good to me!

kaedroho:
Thanks!

PR #7299, opened 2021-06-29
Change spin animation name to make it less prone to be overridden 

jkevingutierrez:
The animation `spin` is too generic, and it can be easily overridden by other libraries (eg. the code used in [django-json-widget](https://github.com/jmrivas86/django-json-widget/blob/master/django_json_widget/static/dist/jsoneditor.min.css)) or by custom code without intention, affecting some animations, like the saving animation.

Thanks for contributing to Wagtail! 🎉

Before submitting, please review the contributor guidelines <https://docs.wagtail.io/en/latest/contributing/index.html> and check the following:

* Do the tests still pass? (https://docs.wagtail.io/en/latest/contributing/developing.html#testing) yes
* Does the code comply with the style guide? (Run `make lint` from the Wagtail root) yes
* For Python changes: Have you added tests to cover the new/fixed behaviour? N/A
* For front-end changes: Did you test on all of [Wagtail’s supported environments](https://docs.wagtail.io/en/latest/contributing/developing.html#browser-and-device-support)?
    * **Please list the exact browser and operating system versions you tested**.
    * **Please list which assistive technologies you tested**.
* For new features: Has the documentation been updated accordingly? N/A

PR #7683, opened 2021-11-05
Change spin animation name to make it less prone to be overridden

jkevingutierrez:
Continuing with the work done in #7299

The animation spin is too generic, and it can be easily overridden by other libraries (eg. the code used in django-json-widget) or by custom code without intention, affecting some animations, like the saving animation.


lb-:
@jkevingutierrez thanks for finding these and updating to align with the previous work, looks good - I have pushed some release notes to your branch & will merge in.

Issue #1250, opened 2015-04-29
Filter PageChooserPanel queryset


rmartins90:
It would be great if you could filter Pages on PageChooserPanel using a queryset.

Something like:
PageChooserPanel('link_page', queryset=EventPage.objects.live().descendant_of(events_index)),

MechanisM:
Yea! I thought about it aswell. For peeps from torchbox, it should be known when adding authors for blog posts in wagtail-torchbox :smile:

gasman:
Interesting idea! Clearly you can't pass an arbitrary queryset as part of the URL to /admin/pages/chooser, so you'd probably have to pass it a reference to the original model field (something like `/admin/pages/chooser/?target_field=demo.BlogPost.author`) and have the chooser view extract the queryset from there. You'd have to make sure it also works when the field is part of an inline child rather than the base Page model - and if you want it to work within StreamField too (where the chooser could be any number of levels deep), that's another layer of complexity altogether...

You'd also need to define the queryset in a way that avoids evaluating it at server startup - it would probably have to be a callable (e.g. a lambda function) that returns a queryset.

In fact, now that I think about it some more - this is an unsolved problem within Django, too. [limit_choices_to](https://docs.djangoproject.com/en/1.8/ref/models/fields/#django.db.models.ForeignKey.limit_choices_to) is Django's attempt at limiting the values of foreign key fields, but it doesn't work with arbitrary querysets (only filter dicts and Q objects), and it doesn't integrate fully with [raw_id_fields](https://docs.djangoproject.com/en/1.8/ref/contrib/admin/#django.contrib.admin.ModelAdmin.raw_id_fields) in Django admin, which is Django admin's equivalent of Wagtail's PageChooserPanel. In other words - there's an opportunity here for someone to solve this problem for both Wagtail *and* Django :-)

But to bring this back down to earth... maybe this is all overkill, and the easy solution is to make the page chooser interface a bit smarter. If the page chooser is asking for an EventPage, and all of our EventPages exist in a single section, then there's no point dropping the user at the root level - we should just take them straight to that section.

jorge-marques:
Maybe it would be possible to drop the user at the shallowest level where a page of the specified type appears? A simple list of pages would also work nicely for most cases. IMO, the search is main advantage over a simple select field.

gasman:
Related / alternative suggestion: #670

nek4life:
[...]

PR #7684, opened 2021-11-06
Phase out special-purpose field panel types

gasman:
This change makes it possible to use `FieldPanel` for all field types, making `StreamFieldPanel`, `RichTextFieldPanel`, `ImageChooserPanel`, `DocumentChooserPanel` and `SnippetChooserPanel` obsolete and `PageChooserPanel` _almost_ obsolete (it's still provided as a convenience for passing the `page_type` or `can_choose_root` arguments, although even this can be replaced with a `FieldPanel` with an appropriate `widget`). This is achieved in part by a new extension to `WagtailAdminModelForm`, where apps can register their own form widgets to be used for a given model field type (so that, for example, a foreign key to Image always uses the AdminImageChooser widget, without having to specify it explicitly via the form class or EditHandler definition).

One thing that might need more consideration is the removal of docs for ImageChooserPanel and DocumentChooserPanel - these currently seem to be written as if they're the first introduction to the image and document models. If this is indeed the case (which wouldn't be great, as these are reference docs rather than howto docs...) we should probably add them back as a new first section on topics/images, and a new topics/documents page.

We might also want to note that if people have _deliberately_ chosen to use FieldPanel instead of a chooser panel (or, for non-edit-handler-based forms, omitted specifying a chooser widget) because a plain dropdown works better for their use case, they'll now need to specify the Select widget explicitly.

### DIFF
--- a/docs/extending/forms.md
+++ b/docs/extending/forms.md
@@ -1,0 +1,58 @@
+# Forms
+
+[Django's forms framework](https://docs.djangoproject.com/en/stable/topics/forms/) can be used within Wagtail admin views just like in any other Django app. However, Wagtail also provides various admin-specific form widgets, such as date/time pickers and choosers for pages, documents, images and snippets. By constructing forms using `wagtail.admin.forms.models.WagtailAdminModelForm` as the base class instead of `django.forms.models.ModelForm`, the most appropriate widget will be selected for each model field. For example, given the model and form definition:
+
+```python
+from django.db import models
+
+from wagtail.admin.forms.models import WagtailAdminModelForm
+from wagtail.images.models import Image
+
+
+class FeaturedImage(models.Model):
+    date = models.DateField()
+    image = models.ForeignKey(Image, on_delete=models.CASCADE)
+
+
+class FeaturedImageForm(WagtailAdminModelForm):
+    class Meta:
+        model = FeaturedImage
+```
+
+the `date` and `image` fields on the form will use a date picker and image chooser widget respectively.
+
+
+## Defining admin form widgets
+
+If you have implemented a form widget of your own, you can configure `WagtailAdminModelForm` to select it for a given model field type. This is done by calling the `wagtail.admin.forms.models.register_form_field_override` function, typically in an `AppConfig.ready` method.
+
+```eval_rst
+.. function:: register_form_field_override(model_field_class, to=None, override=None, exact_class=False)
+
+   Specify a set of options that will override the form field's defaults when ``WagtailAdminModelForm`` encounters a given model field type.
+
+   :param model_field_class: Specifies a model field class, such as ``models.CharField``; the override will take effect on fields that are instances of this class.
+   :param to: For ``ForeignKey`` fields, indicates the model that the field must point to for the override to take effect.
+   :param override: A dict of keyword arguments to be passed to the form field's ``__init__`` method, such as ``widget``.
+   :param exact_class: If true, the override will only take effect for model fields that are of the exact type given by ``model_field_class``, and not a subclass of it.
+```
+
+For example, if the app `wagtail.videos` implements a `Video` model and a `VideoChooser` form widget, the following AppConfig definition will ensure that `WagtailAdminModelForm` selects `VideoChooser` as the form widget for any foreign keys pointing to `Video`:
+
+```python
+from django.apps import AppConfig
+from django.db.models import ForeignKey
+
+
+class WagtailVideosAppConfig(AppConfig):
+    name = 'wagtail.videos'
+    label = 'wagtailvideos'
+
+    def ready(self):
+        from wagtail.admin.forms.models import register_form_field_override
+        from .models import Video
+        from .widgets import VideoChooser
+        register_form_field_override(ForeignKey, to=Video, override={'widget': VideoChooser})
+```
+
+Wagtail's edit views for pages, snippets and ModelAdmin use `WagtailAdminModelForm` as standard, so this change will take effect across the Wagtail admin; a foreign key to `Video` on a page model will automatically use the `VideoChooser` widget, with no need to specify this explicitly.
--- a/docs/extending/forms.md
+++ b/docs/extending/forms.md
@@ -1,4 +1,4 @@
-# Forms
+# Using forms in admin views
 
 [Django's forms framework](https://docs.djangoproject.com/en/stable/topics/forms/) can be used within Wagtail admin views just like in any other Django app. However, Wagtail also provides various admin-specific form widgets, such as date/time pickers and choosers for pages, documents, images and snippets. By constructing forms using `wagtail.admin.forms.models.WagtailAdminModelForm` as the base class instead of `django.forms.models.ModelForm`, the most appropriate widget will be selected for each model field. For example, given the model and form definition:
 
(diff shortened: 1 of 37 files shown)


lb-:
Is it possible to also do away with PagePanel by making the can choose root and page type options pass through via some other kwarg?

Maybe attrs or similar?

gasman:
The only added functionality in PageChooserPanel now is to pass the `page_type` and `can_choose_root` arguments to the widget (as `target_models` and `can_choose_root` respectively), so a line like

    PageChooserPanel('related_page', page_type=[EventPage, BlogPage])

can be rewritten as:

    FieldPanel('related_page', widget=AdminPageChooser(target_models=[EventPage, BlogPage]))

The latter is more wordy and 'low-level', so I'm hesitant to push it as a recommended replacement for PageChooserPanel in that situation. I'd also be reluctant to hard-code any page-chooser-specific special cases in FieldPanel, which would rule out something like

    FieldPanel('related_page', page_type=[EventPage, BlogPage])

and while there might be a middle ground like

    FieldPanel('related_page', widget_args={'target_models': [EventPage, BlogPage]})

...I'm not sure if that's enough of an improvement to be worthwhile, or whether the `page_type` option is even widely used enough to worry about this in the first place. If we _could_ embed this information into the model field definition via `limit_choices_to`, that would neatly sidestep the problem, but that's easier said than done - see #1250.

As a bit of additional context - I'm intending that the next step on this path will be to make `'foo'` an alias of `FieldPanel('foo')` so that for the simplest setups (think snippets rather than pages), the panel definition will just be equivalent to writing a `fields = ['first_name', 'last_name']` list on a ModelForm. Only when you want to customise the form layout would you start adding EditHandler-ish idioms:

    panels = [
        MultiFieldPanel(['first_name', 'last_name']),  # actually we'd probably just rename it 'Row' or something...
        'photo',
    ]

...and hopefully along the way we'll introduce template components so that defining a wrapper like MultiFieldPanel is more a case of overriding a template and less deep Wagtail "bind an instance to a form" voodoo.

In that future scenario, then, I think PageChooserPanel in its current form will sit nicely as a 'power user' feature that you'll drop in if you want more control over how the form behaves - for a simple setup it'll be

    content_panels = [
        'title', 'body', 'related_page'
    ]

but the 'advanced' version is:

    content_panels = [
        'title', 'body',
        PageChooserPanel('related_page', page_types=[EventPage, BlogPage])
    ]

ababic:
@gasman We could adopt something similar to the pattern used by [django.contrib.admin.Modeladmin.fieldsets](https://docs.djangoproject.com/en/4.0/ref/contrib/admin/#django.contrib.admin.ModelAdmin.fieldsets)

- A tuple containing a dictionary is interpreted as a MultiFieldPanel, where the first item is interpreted as the panel heading
- A tuple that contains only strings or FieldPanel instances is interpreted as a RowFieldPanel

Example:

```
content_panels = [
    (
        "Header",   # This is a multifieldpanel / queryset
        {
            "fields": [
                "title", 
                "banner_image",
                ("main_category", "secondary_categoy"),  # This is a row
                "publish_date".
            ]
        },  
    ),
    "body",
]
```

gasman:
@ababic I see the panels definition as being a 'pluggable' thing, where it would be a routine task for developers to create their own panel types if they so wish - so I'm reluctant to attach any syntactic special status to the built-in ones. With some well-thought-out API design I think we can make things readable enough that that isn't necessary:

```
from wagtail.panels import Fieldset, Row

...
    content_panels = [
        Fieldset("Header", [
            "title", 
            "banner_image",
            Row("main_category", "secondary_category"),
            "publish_date",
        ]),
        "body",
    ]
```

kaedroho (changes requested):
Nice work! It's really great to see such a large reduction in complexity too

Only thing I wanted to pick up on was the title new forms doc. Also, do you think we should follow the [how-to guide](https://documentation.divio.com/how-to-guides/) recommendations for this?

docs/extending/forms.md
```diff
@@ -0,0 +1,58 @@
+# Forms
```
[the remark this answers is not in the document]
gasman: I've gone for "Using forms in admin views", as the instructions for overriding widgets are only one section of it. I think there's a risk with going overly specific on these titles, where people have to do extra thinking to figure out if it corresponds to the problem they're trying to solve - if they're coming to the docs with the question "how do I build a form with the Wagtail look and feel", they might not immediately make the connection to "overriding form widgets".

wagtail/admin/tests/test_edit_handlers.py
```diff
@@ -983,22 +940,21 @@ def test_autodetect_page_type(self):
         self.assertIn(expected_js, result)
 
     def test_target_models(self):
```
zerolab, line 942: Before this PR, 

if you have a FK to a specific model, `target_models` includes it.

e.g. for

```python
    page = models.ForeignKey(
        "tests.EventPage", help_text="more help text", on_delete=models.CASCADE
    )
```

the following test is successful

```python
    def test_target_models_on_fk_with_specific_model(self):
        result = PageChooserPanel(
            'page',
        ).bind_to(model=EventPageChooserModel).target_models()
        self.assertEqual(result, [EventPage])
```


after this PR:

```python
    def test_target_models_on_fk_with_specific_model(self):
        panel = PageChooserPanel("page").bind_to(
            model=EventPageChooserModel
        )
        widget = panel.widget_overrides()["page"]
        self.assertEqual(widget.target_models, [EventPage])
```

fails with:

```
ERROR: test_target_models_on_fk_with_specific_model (wagtail.admin.tests.test_edit_handlers.TestPageChooserPanel)
  File "wagtail/admin/tests/test_edit_handlers.py", line 953, in test_target_models_on_fk_with_specific_model
    widget = panel.widget_overrides()["page"]
KeyError: 'page'
```

gasman: This is expected. WagtailAdminModelForm now [has its own rule](https://github.com/wagtail/wagtail/blob/ef2fa2e3944a7844b08aa070552f5046d5433a7f/wagtail/admin/forms/models.py#L66-L68) to select an AdminPageChooser with an appropriate `target_models` argument when it encounters a foreign key to a Page model - as such, it's no longer PageChooserPanel's responsibility to specify an override for that widget (unless PageChooserPanel is configured with an explicit `page_type` or `can_choose_root` argument, which WagtailAdminModelForm won't know about from looking at the ForeignKey alone).

To preserve the intent of that test, I would build the final form and confirm that the expected widget appears there - something like:

    from wagtail.admin.forms.models import WagtailAdminModelForm

    class EventPageChooserForm(WagtailAdminModelForm):
        class Meta:
            model = EventPageChooserModel
            fields = ['page']

    ...

        def test_target_models_on_fk_with_specific_model(self):
            form = EventPageChooserForm()
            widget = form.fields['page'].widget
            self.assertEqual(widget.target_models, [EventPage])

zerolab: Argh, I missed that. Thank you Matthew.

Looks like we'll want to be smarter about in on https://github.com/wagtail/wagtail-localize/blob/main/wagtail_localize/views/edit_translation.py#L266-L280

PR #7687, opened 2021-11-07
Hide bulk actions in report pages

### LINKED ISSUE #7681: Stray checkbox on locked pages report
### Issue Summary

The Locked Pages report (and possibly others inheriting from PageReportView) has a stray additional column with a checkbox, presumably from the bulk actions feature, which causes the column headings to get out of step:

Screenshot 2021-11-04 at 14 38 33

Thanks to @Tijani-Dia for reporting.

### Steps to Reproduce

1. Start a new project with `wagtail start myproject` / `./manage.py migrate` / `./manage.py createsuperuser`
2. Log in, edit the homepage, select 'Lock' from the action menu
3. Go to Reports -> Locked pages

### Technical details

* Python version: 3.8.0
* Django version: 3.2.8
* Wagtail version: 2.15
* Browser version: Chrome 95.0.4638.54

laymonage:
Fix #7681.

I didn't find any other pages that extend `PageReportView` or use the `base_page_report.html` template, but I have added `with hide_bulk_actions=1` in `base_page_report.html` just in case.

This means that new pages that extend `base_page_report.html` and have the bulk actions feature will have to override the `listing` block, e.g.

```html
{% block listing %}
    {% include "wagtailadmin/reports/listing/_list_page_report.html" %}
{% endblock %}
```

(This change is made in a separate commit for easier review).

---

Result:

image

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/reports/base_page_report.html
+++ b/wagtail/admin/templates/wagtailadmin/reports/base_page_report.html
@@ -6,7 +6,7 @@
         <div id="page-results">
             {% if pages %}
                 {% block listing %}
-                    {% include "wagtailadmin/reports/listing/_list_page_report.html" with hide_bulk_actions=1 %}
+                    {% include "wagtailadmin/reports/listing/_list_page_report.html" %}
                 {% endblock %}
             {% else %}
                 {% block no_results %}
--- a/wagtail/admin/templates/wagtailadmin/reports/base_page_report.html
+++ b/wagtail/admin/templates/wagtailadmin/reports/base_page_report.html
@@ -6,7 +6,7 @@
         <div id="page-results">
             {% if pages %}
                 {% block listing %}
-                    {% include "wagtailadmin/reports/listing/_list_page_report.html" %}
+                    {% include "wagtailadmin/reports/listing/_list_page_report.html" with hide_bulk_actions=1 %}
                 {% endblock %}
             {% else %}
                 {% block no_results %}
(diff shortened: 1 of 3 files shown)


zerolab:
Hey @laymonage,

Thank you for this. The changes LGTM. I do wonder if perhaps adding `hide_bulk_actions` in the template context from the report view would help with any custom reports that extend `base_page_report.html` ?

laymonage:
> Hey @laymonage,
> 
> Thank you for this. The changes LGTM. I do wonder if perhaps adding `hide_bulk_actions` in the template context from the report view would help with any custom reports that extend `base_page_report.html` ?

Thanks, that's a good idea. Changing it from the view does change `hide_bulk_actions` for all report pages. If that's desirable, I've pushed a commit to do it that way. We can revert to the first approach if that's preferred.

gasman:
Having a `hide_bulk_actions` flag in the first place seems like a code smell to me. It's an established fact that the page listing includes in `wagtailadmin/pages/listing` are used for more than just the page explorer index view - we previously had to fix this for the page chooser view, and it's entirely possible that third-party code elsewhere is using them.

`_list_explore.html` is the version of the include used specifically for the page explorer - ideally, everything related to the bulk action checkboxes should live in there. Failing that, they can go in `_list.html`, ONLY if they're behind a flag that the explore view opts _in_ to, like the existing `show_ordering_column` flag. `_list.html` is in active use by other views, including these reports - the root cause of this bug is that we made a breaking change there that requires those other views to add a flag to restore the old behaviour.

gasman:
Closing in favour of #7701, which addresses the problem for all listing pages using the `wagtailadmin/pages/listing` includes, not just reports.

PR #7694, opened 2021-11-09
Use correct syntax for image debug logging

RealOrangeOne:
Fixes an issue in #7391 which used the incorrect logging syntax

This should only affect users which have debug logging enabled for wagtail. I don't believe it affects all users.

### DIFF
--- a/wagtail/images/models.py
+++ b/wagtail/images/models.py
@@ -311,19 +311,21 @@ class AbstractImage(ImageFileMixin, CollectionMember, index.Indexed, models.Mode
         except Rendition.DoesNotExist:
             # Generate the rendition image
             try:
-                logger.debug("Generating '%s' rendition for image %d", (
+                logger.debug(
+                    "Generating '%s' rendition for image %d",
                     filter.spec,
                     self.pk,
-                ))
+                )
 
                 start_time = time.time()
                 generated_image = filter.run(self, BytesIO())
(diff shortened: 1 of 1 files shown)


wagtail/images/models.py
```diff
@@ -311,21 +311,23 @@ def get_rendition(self, filter):
...
+                logger.debug(
+                    "Generated '%s' rendition for image %d in %.1fms",
                     filter.spec,
                     self.pk,
                     (time.time() - start_time) * 1000
-                ))
+                )
             except:  # noqa:B901,E722
-                logger.debug("Failed to generate '%s' rendition for image %d: %s", filter.spec, self.pk)
+                logger.debug("Failed to generate '%s' rendition for image %d", filter.spec, self.pk)
```
RealOrangeOne, line 330: Question: Should this be `logger.exception`? It feels like something generally useful to know regardless of any changes to the logging level, and isn't _necessarily_ additionally debugging information.

gasman: We are re-raising the exception here, to be handled by whatever mechanism is set up site-wide for handling exceptions generally, and I see no reason to treat this differently to any other exception. I assumed the only reason we're catching this at all is to make sure that the "generating rendition" log entry has a corresponding end state logged.

gasman (approved):

PR #7696, opened 2021-11-09
Aging pages report

Tijani-Dia:
The goal of this PR is to add a report to track pages that haven't been updated recently.

* Do the tests still pass? (https://docs.wagtail.io/en/latest/contributing/developing.html#testing) Yes
* For new features: Has the documentation been updated accordingly? Not yet

### DIFF
--- a/wagtail/admin/templates/wagtailadmin/reports/aging_pages.html
+++ b/wagtail/admin/templates/wagtailadmin/reports/aging_pages.html
@@ -1,0 +1,57 @@
+{% extends 'wagtailadmin/reports/base_report.html' %}
+{% load i18n wagtailadmin_tags %}
+
+{% block results %}
+    {% if object_list %}
+        <table class="listing">
+            <thead>
+                <tr>
+                    <th class="title">
+                        {% trans 'Title' %}
+                    </th>
+                    <th class="status">
+                        {% trans 'Status' %}
+                    </th>
+                    <th class="updated">
+                        {% trans 'Last published' %}
+                    </th>
+                    <th>{% trans 'Last published by' %}</th>
+                    <th class="type">
+                        {% trans 'Type' %}
+                    </th>
+                </tr>
+            </thead>
+            <tbody>
+                {% for page in object_list %}
+                    <tr>
+                        <td class="title" valign="top">
+                            {{ page.title }}
+                        </td>
+                        <td class="status" valign="top">
+                            {% include "wagtailadmin/shared/page_status_tag.html" with page=page %}
+                        </td>
+                        <td class="updated" valign="top">
+                            <div class="human-readable-date" title="{{ page.last_published_at|date:'DATETIME_FORMAT' }}">
+                                {% blocktrans with time_period=page.last_published_at|timesince %}
+                                    {{ time_period }} ago
+                                {% endblocktrans %}
+                            </div>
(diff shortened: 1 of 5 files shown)


gasman (commented):
This is looking really good! I think the only thing that's really missing is a link from the 'title' field to the edit page. With that added, and some basic tests, I think this will be good to go.

Tijani-Dia:
Thanks @gasman. I've added the link to edit the page and some tests.

gasman:
We shouldn't just sweep the test failure under the carpet by changing the result :-) It's important that the tests should work individually as well as when running the whole test suite.

After some experimenting, I found that the original test failed (returning `Page` instead of `wagtailcore | page`) when you run [TestAuditLogAdmin](https://github.com/wagtail/wagtail/blob/cb70218166c3a0454a5352b780f5472cddae9d88/wagtail/admin/tests/test_audit_log.py#L14) before it. The problem is this code in the site history report:

https://github.com/wagtail/wagtail/blob/cb70218166c3a0454a5352b780f5472cddae9d88/wagtail/admin/views/reports/audit_logging.py#L98-L105

It's modifying the `custom_field_preprocess` attribute to get the nicer content type representation - but this attribute is defined on the `ReportView` class, so it ends up overwriting this setting for _all_ reports. Adding this line above that code fixes it:

        self.custom_field_preprocess = self.custom_field_preprocess.copy()

We probably do want the aging pages report to use the `get_content_type_label` representation, so the same code should be copied into `AgingPagesView` too...

    def __init__(self, **kwargs):
        super().__init__(**kwargs)

        self.custom_field_preprocess = self.custom_field_preprocess.copy()
        self.custom_field_preprocess['content_type'] = {
            self.FORMAT_CSV: get_content_type_label,
            self.FORMAT_XLSX: get_content_type_label
        }

Tijani-Dia:
Alright @gasman, I discovered the root of the problem after changing the tests.

I've updated it. Thanks.

gasman:
The test failure is only happening on Django 3.0, and I think it might be caused by this:

https://docs.djangoproject.com/en/3.2/releases/3.1/#forms

> django.forms.DateTimeField now accepts dates in a subset of ISO 8601 datetime formats, including optional timezone, e.g. 2019-10-10T06:47, 2019-10-10T06:47:23+04:00, or 2019-10-10T06:47:23Z. The timezone will always be retained if provided, with timezone-aware datetimes being returned even when USE_TZ is False.

In the test, you're passing `self.aboutus_page.last_published_at` as the value of the `last_published_at` parameter, and since that's a datetime value, passing it as a URL parameter probably causes it to be converted to one of the 'raw' timestamp strings above (which is only supported by DateTimeField from Django 3.1 onward). This isn't a problem in real-world usage of the report, of course, because the `last_published_at` parameter will normally come from the form in simple yyyy-mm-dd format. Changing the test to

        response = self.get(
            params={"last_published_at": '2015-01-01'}
        )

(a date which is definitely in the past, but after the `last_published_at` date for the homepage in the test fixtures) fixes the test for all Django versions. I'll make that change and merge now.

Tijani-Dia:
Oh alright, great you caught that.

gasman:
Merged in 660b00b7a6ac52fb7475b0446e6dcc8b90048d89

PR #7702, opened 2021-11-10
Fix bulk page publishing

### LINKED ISSUE #7690: Wagtail 2.15 Bulk Publish for Inherited Page

### Issue Summary
Assume we have two classes:
class A(Page):

class B(A):
  nickname = models.CharField(max_length=255)

When do bulk action of publishing the pages with B, an error will occur:
image

The existing solution I think is to customize bulk action according to https://docs.wagtail.io/en/stable/extending/custom_bulk_actions.html#custom-bulk-actions
But I would like to know if there is other solution?

gasman:
Fixes #7690. We need to retrieve the specific page instance before `save_revision`, or we'll end up clearing the subclass's fields (resulting in a ValidationError if any were required, or a published version with blanked fields if not).


zerolab (approved):
Once again excellent 🕵️‍♂️!

Tested this locally and it works a treat.

Issue #7700, opened 2021-11-10
Wagtail 2.15: Image renditions for re-oriented images are scaled and cropped using un-oriented dimensions


jmichalicek:
Found a bug? Please fill out the sections below. 👍

### Issue Summary

Renditions of images which need rotated by `willow.auto_orient()` are being created based on the dimensions of the unoriented image. For example an image from Android phone held in portrait creates the image with exif data stating the landscape dimensions and an orientation value of 6. So my phone says portrait photos are 4032x3024 with an orientation of 6.

This results in photos which are portraits being cut off at the bottom and having a large black bar on the right side.

I have done some digging and fixed this issue locally and so would like to make a PR with the fix. I have detailed what seems to be happening/need done to reesolve below.

### Steps to Reproduce

1. Take a picture with a device which can do portrait or landscape, but always saves the exif data for only one of those with an exif orientation value applied. I used a Samsung Galaxy S8 but I expect most phones are doing this.
2. Upload the image to wagtail images
3. Look at the images on the images list, in the image detail, and any generated for actual pages have the wrong dimensions

What appears to be the case is that the `willow.auto_orient()` call performs as expected and rotates the image. In the 2.15 code an ImageTransform is created and is used to get the rectangle dimensions of the image. Rather than using the re-oriented image, it is still working off of the original. The result is that it is cropping and scaling based on a 4032x3024 landscape photo in my case, so photos which are actually portrait get cropped based on those dimensions rather than the rotated dimensions of 3024x4032.

I have found that in `wagtail.images.models.Transform.run()` if I use the returned value from `willow.get_pillow_image()` after `willow.auto_orient()` is called everywhere which `image` which was passed into the method is currently used then everything appears to work correctly.

It does not look like that is going to work if Wand is being used since `get_pillow_image()` is specific to willow's `PillowImage` class. I will test just using `willow.image` this evening. My other thought would be to update Willow to add a plain `get_image()` to the plugins and register that with `@Image.operation` so that there is a consistently named operation to get at the image if not directly accessing `willow.image` is preferred.

I am attaching the actual original image and the rendition using the `original` format spec.

Original:
![wagtail_renditions_bug](https://user-images.githubusercontent.com/1059070/141117543-0d09245e-23b8-4052-8a45-b5d17138285e.jpg)

Rendition:
![wagtail_renditions_bug original](https://user-images.githubusercontent.com/1059070/141117572-06547567-e122-40ab-bf09-18499fba51a6.jpg)

[...]

PR #7704, opened 2021-11-11
Fix transform operations in Filter.run() when image has been re-oriented

jmichalicek:
This is a proposed fix for https://github.com/wagtail/wagtail/issues/7700. It adds an optional `transform_width` and `transform_height` arguments to `wagtail.images.models.Filter.get_transform().
If both are passed in, then use them otherwise image.width and image.height for the ImageTransform.

Would it be preferred to pass these in as a tuple to ensure both are passed? Or raise an exception if both are not passed in? What about checks for invalid values such as negative numbers?

Are there any new tests which could be useful for this? I think they would need to compare the width and height of the original and the transform and make sure they swapped.

* Do the tests still pass? (https://docs.wagtail.io/en/latest/contributing/developing.html#testing): yes
* Does the code comply with the style guide? (Run `make lint` from the Wagtail root): yes


wagtail/images/models.py
```diff
@@ -459,7 +459,10 @@ def get_transform(self, image):
          - .matrix - An affine transformation matrix that combines any
            transform/scale/rotation operations that need to be applied to the image
         """
-        transform = ImageTransform((image.width, image.height))
```
kaedroho, line 462: I think this also highlights a minor bug with our image width/height values not accounting for the EXIF orientation, which i think are displayed in the admin.

Anyone think we should solve this too?

This wouldn't change anything for this PR though since we'd obviously still have to deal with legacy images with incorrect width/height

wagtail/images/models.py
```diff
@@ -450,7 +450,7 @@ def filter_operations(self):
             if isinstance(operation, FilterOperation)
         ]
 
-    def get_transform(self, image):
+    def get_transform(self, image, transform_width=None, transform_height=None):
```
kaedroho, line 453: Just for consistency, shall we pass the width and height together as a tuple?

kaedroho (changes requested):
This looks good, thanks! Just a minor tweak consistency. Also, please could you add a test? (you can find some sample images with different orientations [here](https://github.com/wagtail/Willow/tree/master/tests/images/orientation))

wagtail/images/models.py
jmichalicek: > Just for consistency, shall we pass the width and height together as a tuple?

Makes sense to me. I asked about that up in the main comments. I wasn't sure if a tuple or named arguments would be preferred since the named parameters make it extra explicit what those are, but do leave more room for mistakes as well.

I'd be happy to make it be a tuple(width, height) and then just document that in the docstring. I should double check for other docs which need updated as well. I wanted to get this draft up and make sure this was an ok approach before changing much else. I'll try to get that and tests tonight - woorst case it'll be tomorrow night.

wagtail/images/models.py
jmichalicek: I'm not sure if anything needs done for that. It does show the correct width and height of the `original` format rendition in the image admin. That rendition is just created with an incororect width and height on 2.15 if `willow.auto_orient()` had to rotate the image.

You know far more about the internals of this than I do. If there is more to that part than I'm seeing, I'd love to stay in the loop and, time allowing, help with resolving it.

kaedroho: Oh, I didn't know we did that! This is fine then, thanks!

jmichalicek:
I think I have things in a good place. Updated test cases coming in the morning after I double check them with a fresh brain.

wagtail/images/models.py
jmichalicek: @kaedroho I've added tests for the Rendition model. I borrowed the tests from Willow which you pointed me at and tweaked them to check the initial dimensions, then generate a rendition, and then check the dimensions of the rendition model. It also goes through Willow to open the files so that it can verify the actual image file's dimensions are still correct and that the correct colored pixels are where they belong.

jmichalicek:
Should I be merging `main` into this branch or rebase it on top of `main`? I didn't catch it in the documentation about contributing. Other PRs I've looked at did not appear to have any merge commits for `main` into the feature branch.

jacobtoppm:
@jmichalicek Just to jump in to answer your last question: we normally encourage rebasing on top of main, thanks for checking!

jmichalicek:
> @jmichalicek Just to jump in to answer your last question: we normally encourage rebasing on top of main, thanks for checking!

Thanks @jacobtoppm , just rebased and pushed up. I'll keep it up to date regularly until this gets merged in.

kaedroho (commented):
(forgot to leave review message before merging!)

This looks good to me, thanks for fixing!

jmichalicek:
Glad to help out, I've gotten a lot of use out of Wagtail personally and for work.

PR #7707, opened 2021-11-11
Implement MovePageAction and admin API for moving pages

Tijani-Dia:
This a continuation of the actions works started in #7682.

### DIFF
--- a/wagtail/admin/api/actions/move.py
+++ b/wagtail/admin/api/actions/move.py
@@ -1,0 +1,44 @@
+from django.core.exceptions import ValidationError as DjangoValidationError
+from django.shortcuts import get_object_or_404
+from rest_framework import fields, status
+from rest_framework.exceptions import ValidationError
+from rest_framework.response import Response
+from rest_framework.serializers import Serializer
+
+from wagtail.core.actions.move_page import MovePageAction
+from wagtail.core.models import Page
+
+from .base import APIAction
+
+
+class MovePageAPIActionSerializer(Serializer):
+    target_page_id = fields.IntegerField(required=True)
+    position = fields.CharField(required=False)
+
+
+class MovePageAPIAction(APIAction):
+    serializer = MovePageAPIActionSerializer
+
+    def _action_from_data(self, instance, data):
+        target_page_id = data["target_page_id"]
+        target = get_object_or_404(Page, id=target_page_id)
+
+        return MovePageAction(
+            page=instance, target=target, pos=data.get("pos"), user=self.request.user
+        )
+
+    def execute(self, instance, data):
+        serializer = self.serializer(data=data)
+        if not serializer.is_valid():
+            return Response(serializer.errors, status=400)
+
+        action = self._action_from_data(instance, serializer.data)
+
+        try:
+            action.execute()
+        except DjangoValidationError as e:
+            raise ValidationError(e.message_dict)
+
+        instance.refresh_from_db()
+        serializer = self.view.get_serializer(instance)
+        return Response(serializer.data, status=status.HTTP_200_OK)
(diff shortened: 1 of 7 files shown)


wagtail/core/actions/copy_page.py
```diff
@@ -0,0 +1,257 @@
...
+                to = self.page.get_parent()
+
+            if not self.page.permissions_for_user(self.user).can_copy_to(to, self.recursive):
+                raise CopyPagePermissionError("You do not have permission to copy this page")
+
+            if self.keep_live:
+                destination_perms = UserPagePermissionsProxy(self.user).for_page(self.to)
+
+                if not destination_perms.can_publish_subpage():
+                    raise CopyPagePermissionError("You do not have permission publish a page at the destination")
```
jacobtoppm, line 73: Missing a "to"

wagtail/core/actions/copy_page.py
```diff
@@ -0,0 +1,257 @@
...
+
+    def _copy_page(self, page, to=None, update_attrs=None, exclude_fields=None, _mpnode_attrs=None):
```
jacobtoppm, line 75: I feel like this or `execute` need wrapping in a transaction - we're doing a lot of successive database operations, we should probably make sure we don't end up with pages without their M2M relations or similar

jacobtoppm (changes requested):

wagtail/core/actions/copy_page.py
kaedroho: Feel like should be added to ``execute()`` since adding it to ``_copy_page`` would create a save point before copying each page

wagtail/admin/api/actions/move.py
```diff
@@ -0,0 +1,40 @@
...
+
+from wagtail.core.actions.move_page import MovePageAction
+from wagtail.core.models import Page
+
+from .base import APIAction
+
+
+class MovePageAPIActionSerializer(Serializer):
+    target_page_id = fields.IntegerField(required=True)
+    position = fields.CharField(required=False)
```
kaedroho, line 16: Could we make this a ``ChoiceField`` to add some validation?

```python
position = fields.ChoiceField(required=False, choices=['left', 'right', 'first-child', 'last-child', 'first-sibling', 'last-sibling'])
```

kaedroho (changes requested):
Nice work! Just one small tweak and then I think this is good to go!

wagtail/core/actions/move_page.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class MovePageAction:
+    def __init__(self, page, target, pos=None, user=None):
+        self.page = page
+        self.target = target
+        self.pos = pos
+        self.user = user
+
+    def check(self, skip_permission_checks=False):
```
kaedroho, line 29: Is it worth checking that the page can be moved to the new position based on the values of the ``sub_page_types`` and ``parent_page_types`` attributes on the parent and moved page model?

wagtail/admin/api/actions/move.py
```diff
@@ -0,0 +1,40 @@
...
+from rest_framework.serializers import Serializer
+
+from wagtail.core.actions.move_page import MovePageAction
+from wagtail.core.models import Page
+
+from .base import APIAction
+
+
+class MovePageAPIActionSerializer(Serializer):
+    target_page_id = fields.IntegerField(required=True)
```
kaedroho, line 15: I think for consistency with copy, we should rename this to ``destination_page_id``

Tijani-Dia: Alright!

wagtail/admin/api/actions/move.py
Tijani-Dia: Sure!

kaedroho (approved):
Looks good to me now! Just needs merging main or rebasing and it'll be ready to merge

Tijani-Dia:
I've merged main branch in this branch.

PR #7708, opened 2021-11-12
Duplicate queries in admin

Tijani-Dia:
I've done here some initial work about duplicate queries in the admin.

I'm adding first failing tests to showcase the problem.


zerolab:
@Tijani-Dia I believe various other PRs have addressed some, if not all of the changes in this. Will you be able to revise it, or shall we close it and come back fresh?

PR #7713, opened 2021-11-17
Prevent duplicate nested streamblocks to have duplicated UUIDs - Fix #7712

### LINKED ISSUE #7712: Duplicate streamfield block duplicates UUID
### Issue Summary
I noticed streamfield's  blocks uuid can be not unique. 

It can be an issue, in my case in a headless setup, I cache blocks based on their UUIDs.

### Steps to Reproduce
For example, let's consider a streamfield, containing a structblock, containing a streamblock.
Duplicating an instance of this structblock, with existing streamblocks, cause the nested streamblock UUIDs to remain identical to the original ones

I created an example repo here https://github.com/fabienheureux/wagtail-duplicate-streamfield-uuid

1. `git clone https://github.com/fabienheureux/wagtail-duplicate-streamfield-uuid`
2. `./manage.py runserver`
3. `./manage.py shell`
4. `from home.models import HomePage`
5. `print(HomePage.objects.first().blocks.raw_data)`

```json
[
   {
      "type":"simple_question",
      "value":{
         "extended_answer":{
            "body":[
               {
                  "type":"text",
                  "value":"<p data-block-key=\"a7uom\">Test 1 2 1 2</p>",
                  "id":"ff7f1a7a-42a6-46c9-915c-5ca453b23698" 
               }
            ]
         }
      },
      "id":"520971d8-3d49-4779-b833-46222fdd48e1"
   },
   {
      "type":"simple_question",
      "value":{
         "extended_answer":{
            "body":[
               {
                  "type":"text",
                  "value":"<p data-block-key=\"a7uom\">Coucou</p>",
                  "id":"ff7f1a7a-42a6-46c9-915c-5ca453b23698"
               }
            ]
         }
      },
      "id":"ec72605c-b58a-4d57-9ead-2af3196575b4"
   }
]
```
➡️ The body's UUIDs are the same

### How to reproduce from Wagtail admin
1. Create a Streamfield block
2. Fill some data for the StreamBlock inside this block
3. Duplicate the Streamfield block

### Technical details

* Python version: any version
* Django version: 3.2.9
* Wagtail version: 2.15.1

fabienheureux:
Fix #7712 

# Description
When duplicating a block in a StreamField, nested StreamBlock UUIDs remain. 
This change removes the UUIDs from nested blocks so that they are generated server-side on Page's save. 

This is still a work in progress as I am not sure if I made this change in the right place. Once I get confirmation that this properly fixes the issue, I will add some tests. 

# TODO
- [x] Add a description
- [ ] Write tests

### DIFF
--- a/client/src/components/StreamField/blocks/StreamBlock.js
+++ b/client/src/components/StreamField/blocks/StreamBlock.js
@@ -292,13 +292,12 @@ export class StreamBlock extends BaseSequenceBlock {
     Object.keys(childState).forEach(key => {
       const value = childState[key]
       if (Array.isArray(value)) {
-        // In case the child is an array, like a list of blocks
+        // The value can contain a list of blocks in which we need to set id to null.
         newChildState[key] = value.map(item => this._removeIdsFromChildState(item));
       } else if (typeof value === 'object' && value !== null) {
-        // In case the child is an object, like a structblock
+        // The value can be an object, a StructBlock for example.
         newChildState[key] = this._removeIdsFromChildState(childState[key]);
       } else if (key === 'id' && !!childState[key]) {
-        // In case the key is "id", we remove it
         newChildState[key] = null;
       }
     });
--- a/client/src/components/StreamField/blocks/StreamBlock.js
+++ b/client/src/components/StreamField/blocks/StreamBlock.js
@@ -287,34 +287,27 @@ export class StreamBlock extends BaseSequenceBlock {
     return [blockDef, initialState, uuidv4()];
   }
 
-  removeIdsFromChildState(childState) {
-    console.log("Initial childState :", childState)
-    for (var i in childState) {
-      // In case the child is an array, like a list of blocks
+  _removeIdsFromChildState(childState) {
+    const newChildState = { ...childState };
+    Object.keys(childState).forEach(i => {
       if (Array.isArray(childState[i])) {
-        childState[i] = childState[i].map(item => this.removeIdsFromChildState(item))
+        // In case the child is an array, like a list of blocks
+        newChildState[i] = childState[i].map(item => this._removeIdsFromChildState(item));
+      } else if (typeof childState[i] === 'object' && childState[i] !== null) {
+        // In case the child is an object, like a structblock
+        newChildState[i] = this._removeIdsFromChildState(childState[i]);
+      } else if (i === 'id' && !!childState[i]) {
+        // In case the key is "id", we remove it
+        newChildState[i] = null;
       }
+    });
 
-      // In case the child is an object, like a structblock
-      if (typeof childState[i] === "object" && childState[i] !== null) {
-        childState[i] = this.removeIdsFromChildState(childState[i])
-      }
-
-      // In case the key is not an object, and not id, we do nothing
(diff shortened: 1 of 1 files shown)


gasman (changes requested):
Thanks for looking into this @fabienheureux!
It's an important part of StreamField's design that each block type is only responsible for managing its own part of the data structure, and doesn't try to 'guess' how to step through its child objects based on their data type. That means the general pattern for this method ought to be something like getState or getValue - blocks that manage child blocks (e.g. StreamBlock, StructBlock and ListBlock [1]) should call a method on those child blocks to perform whatever manipulation is required.

This potentially means an addition to the block API, and we need to think quite carefully about what that addition should look like - it has to be something 'high-level' enough that it makes sense for any block to implement it without having to think about an internal detail of StreamBlock (so a method called 'fixUUIDs' or whatever would not be good).

The root problem here is that StreamBlock implements block duplication by calling `getState` on the child block and creating a new block with that state, on the assumption that it's always desirable for the copied block to exactly match the existing block's state. We now know that isn't true - we don't want the IDs in nested StreamBlocks to match.

I think the neatest way to address that is with a new method `getDuplicatedState()` that behaves like `getState`, but returns the state that we want a new copy of this block to have. For a simple block like CharBlock, `getDuplicatedState` would be identical to `getState` - and since we don't really want to have to go through _all_ existing blocks adding a new method, I suggest that we fall back on `getState` whenever `getDuplicatedState` isn't defined.

A container block like StructBlock would implement `getDuplicatedState` recursively, by calling `getDuplicatedState` on each child (falling back on `getState` if that's not defined) and building a data structure out of the results. StreamBlock would do the same, with the additional step of stripping out UUIDs (at its own top level only - the recursion would do the rest). Then, duplicateBlock would be changed to call `getDuplicatedState` instead of `getState`, and - I think - it would all fall into place from there.


[1] As of 2.15, we can add TypedTableBlock to that list. Unfortunately I don't think we'll be able to make this work for third-party container block types (do any exist?) without code changes.

fabienheureux:
Thanks a lot for your review/insight @gasman !
I will look into this, and try to get a better understanding of current StreamField frontend architecture before pushing new changes.

jacobtoppm (changes requested):
I think new UUIDs should be generated on the frontend rather than simply being stripped out - commenting uses them to find the contentpath for the streamfield block, which sets out which block or field the comment is on (and I think it's reasonable to expect every block to have a unique path/uuid combo to track it in code - we might well use this elsewhere in the future).

gasman:
@jacobtoppm FWIW, I think the existing version of this code does the same thing - see the `childState.id = null;` line that's been removed here. Looks like bd606ab15cc847fec2d1666bbd9a87350192a7ee added client-side UUID generation, but only did it for insertions (or maybe I'm reading the code wrong, and `_getChildDataForInsertion` does actually get called on duplication further down the line).

jacobtoppm:
@gasman Good point! Clearly something we missed. But if we're revisiting this point, I would suggest doing client side id generation for duplication as well.

fabienheureux:
Closing as it was fixed in #9108

PR #7717, opened 2021-11-17
Drop Python 3.6 support

gasman:
Python 3.6 reaches end of life on 23 Dec 2021, so it'll be out of support by the time Wagtail 2.16 happens.

Also take the opportunity to remove appveyor.yml - this was going to be a CI setup for Windows and MSSQL, but that never happened, and apparently this file hasn't been touched since Python 3.5 and Django 1.x.


zerolab (approved):
Can't find any fault in this. :shipit:

PR #7719, opened 2021-11-18
Search with MySQL backend errors if query has more than 2 words

cnk:
When using the new MySQL search backend, the search throws the following exception if my search contains more than 2 words. Works fine if those words are a phrase (enclosed in double quotes). I am opening this PR to see if I can replicate the behavior I am seeing locally in our TOX suite - 3 word search works OK in Postgres and SQLite but not in MySQL.

One other odd thing, self.connector always seems to be empty. 

```
  File "/vote-ve/lib64/python3.7/site-packages/django/db/models/expressions.py", line 480, in as_sql
    sql, params = compiler.compile(self.lhs)
  File "/vote-ve/lib64/python3.7/site-packages/django/db/models/sql/compiler.py", line 445, in compile
    sql, params = node.as_sql(self, self.connection)
  File "/vote-ve/lib64/python3.7/site-packages/wagtail/search/backends/database/mysql/query.py", line 206, in as_sql
    compiled_query = compiler.compile(self.query)  # Compile the query to a string
  File "/vote-ve/lib64/python3.7/site-packages/django/db/models/sql/compiler.py", line 445, in compile
    sql, params = node.as_sql(self, self.connection)
  File "/vote-ve/lib64/python3.7/site-packages/wagtail/search/backends/database/mysql/query.py", line 153, in as_sql
    sql, params = compiler.compile(self.value)
  File "/vote-ve/lib64/python3.7/site-packages/django/db/models/sql/compiler.py", line 445, in compile
    sql, params = node.as_sql(self, self.connection)
  File "/vote-ve/lib64/python3.7/site-packages/wagtail/search/backends/database/mysql/query.py", line 87, in as_sql
    if self.lhs.invert and self.connector == '+':  # NOTE: This is a special case for MySQL. If either side's operator is AND (+), and it is inverted, the operator should become NOT (-). If we did nothing, the result could would be '+X +(-Y)' for And(X, Not(Y)), which seems correct, but produces a wrong result. The solution is to turn the query into '+X -Y', which does work, and therefore this is done here.

Exception Type: AttributeError at /search/
Exception Value: 'CombinedLexeme' object has no attribute 'invert'
```

### DIFF
--- a/wagtail/search/tests/test_backends.py
+++ b/wagtail/search/tests/test_backends.py
@@ -82,7 +82,7 @@ class BackendTests(WagtailTestUtils):
 
     def test_ranking(self):
         # Note: also tests the "or" operator
-        results = list(self.backend.search("JavaScript The Definitive", models.Book, operator='or'))
+        results = list(self.backend.search("JavaScript Definitive", models.Book, operator='or'))
         self.assertUnsortedListEqual([r.title for r in results], [
             "JavaScript: The good parts",
             "JavaScript: The Definitive Guide"
--- a/wagtail/search/tests/test_backends.py
+++ b/wagtail/search/tests/test_backends.py
@@ -82,7 +82,7 @@ class BackendTests(WagtailTestUtils):
 
     def test_ranking(self):
         # Note: also tests the "or" operator
-        results = list(self.backend.search("JavaScript Definitive", models.Book, operator='or'))
+        results = list(self.backend.search("JavaScript The Definitive", models.Book, operator='or'))
         self.assertUnsortedListEqual([r.title for r in results], [
             "JavaScript: The good parts",
             "JavaScript: The Definitive Guide"
@@ -106,7 +106,7 @@ class BackendTests(WagtailTestUtils):
 
     def test_search_and_operator(self):
         # Should not return "JavaScript: The good parts" as it does not have "Definitive"
-        results = self.backend.search("JavaScript Definitive", models.Book, operator='and')
+        results = self.backend.search("JavaScript The Definitive", models.Book, operator='and')
         self.assertUnsortedListEqual([r.title for r in results], [
             "JavaScript: The Definitive Guide"
         ])


cnk:
@kaedroho @ACMCMC I am sorry I didn't get any real world testing done until now. I don't see why this error is happening. I am wondering if the connector being blank means the problem is earlier than when this error is thrown. The problem happens with 'and' and 'or' queries, though I only provided a failing test for the 'and' condition. Any ideas?

ACMCMC:
Hi @cnk ! I need to look into it and can't really provide an answer right now, as I am travelling and don't have my laptop or access to a computer right now. I will get home on Monday, and I hope I can find out a proper answer then. Thanks so much for bringing this up! I hope this is not too urgent 🙏

cnk:
@ACMCMC Not urgent at all. I am just embarrassed I can't see what the problem is. Enjoy your travels!

ACMCMC:
Perfect, thanks 🤞 I will get back to you on this issue as soon as I can look into it @cnk 🙌

ACMCMC:
> Perfect, thanks 🤞 I will get back to you on this issue as soon as I can look into it @cnk 🙌

I have been able to reproduce your problem, and I believe I have solved it. I can either make a PR to this branch (`mysql-search-problem`), or open an issue with a PR solving it directly on the `wagtail` repo.

In a side note, please be aware that the test that you have specified `self.backend.search('Javascript The Definitive', models.Book, operator='and')` will not 'work' as expected. More specifically, it will not return the 'JavaScript: The Definitive Guide' `Book`. That's because MySQL doesn't index stop words ('The'). So, as our search query will be '+Javascript **+The** +Definitive', it's impossible that any entry contains 'The', and therefore it's impossible that any entry matches the three specified search terms. However, `self.backend.search('Javascript Definitive Guide', models.Book, operator='and')` will indeed work as expected. A possible improvement would be to remove stopwords from search queries. A similar case is found [here](https://github.com/wagtail/wagtail/blob/131cd076b77fa2e4c79f546d975ee4e52cb0ce32/wagtail/search/tests/test_mysql_backend.py#L51).

cnk:
@ACMCMC Why don't we just close this PR and you start a fresh one? The only reason I created this as a PR was so I could get GitHub to run the combinations of django versions + different search backends. 

And I am glad to hear MySQL doesn't index stop words. I was quite surprised to find that ElasticSearch did index 'the'.

Issue #7725, opened 2021-11-23
Search fails on MySQL with missing `invert` attribute


ACMCMC:
### Issue Summary

When doing a three-word (or more) search query, such as:

```python3
search("W1 W2 W3", myModel, operator='and')
```

It errors with `AttributeError: 'CombinedLexeme' object has no attribute 'invert'`.

Originally reported by @cnk on #7719.

### Steps to Reproduce

1. In the `wagtail.search.tests.test_backends.BackendTests.test_search_and_operator` method, change the search query to `"JavaScript Definitive Guide"` instead of `"JavaScript Definitive"`.
2. Run the test on a MySQL database environment.

* I have confirmed that this issue can be reproduced as described on a fresh Wagtail project: yes

### Technical details

* Python version: 3.9.1
* Django version: 3.1.13
* Wagtail version: 2.16a0

ACMCMC:
I have found the solution. I will submit a PR solving the issue and reference it.

gasman:
Fixed in #7726

PR #7726, opened 2021-11-23
Solved missing `invert` attribute bug

ACMCMC:
Solves issue #7725 🙌

### DIFF
--- a/wagtail/search/backends/database/mysql/query.py
+++ b/wagtail/search/backends/database/mysql/query.py
@@ -103,7 +103,7 @@ class CombinedLexeme(LexemeCombinable):
             rsql, params = compiler.compile(self.rhs)
         value_params.extend(params)
 
-        combined_sql = '{}{} {}{}'.format(lhs_connector, lsql, rhs_connector, rsql)  # if self.connector is '+' (AND), then both terms will be ANDed together. We need to repeat the connector to make that work.
+        combined_sql = '({}{} {}{})'.format(lhs_connector, lsql, rhs_connector, rsql)  # if self.connector is '+' (AND), then both terms will be ANDed together. We need to repeat the connector to make that work.
         combined_value = combined_sql % tuple(value_params)
         return '%s', [combined_value]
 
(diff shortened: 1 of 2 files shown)


lb-:
@ACMCMC - thanks for taking the time to report and debug/resolve this.

Would you be able to look at creating a unit test to cover this DB specific issue, it would be very easy for something like this to be missed/inverted ;)  in the future and it would be great to have a test to cover it.

ACMCMC:
> Would you be able to look at creating a unit test to cover this DB specific issue, it would be very easy for something like this to be missed/inverted ;) in the future and it would be great to have a test to cover it.

Yes! Of course, that's a great idea 🙌 I will add a commit to this PR with that unit test. However, do you think it's better to add it only to the MySQL backend (`test_mysql_backend.py`), or to instead create a general test that applies to every backend (`test_backends.py`), and specialize that test in the MySQL backend, to add other MySQL-specific features to the test?

gasman:
@ACMCMC Since the bug occurs when running a standard non-database-specific query, I think it'd be most useful to write it as a general test for all backends - that way we'll catch any similar bugs if they exist on the other backends.

ACMCMC:
The latest commit on the branch is the requested change on the unit tests. I chose not to add another test, which would contribute to a higher delay when running all the suite of tests of Wagtail. Instead, I just modified the `test_plain_text_multiple_words_and` test, to test for a 3+ words case.

The tests are now failing because of it, as described in #7727 . I will submit another commit with the fix for that issue.

ACMCMC:
The latest commit on the branch is the requested change on the unit tests. I chose not to add another test, which would contribute to a higher delay when running all the suite of tests of Wagtail. Instead, I just modified the `test_plain_text_multiple_words_and` test, to test for a 3+ words case.

The tests are now failing because of it, as described in #7727 . I will submit another commit with the fix for that issue.

gasman (approved):
Looks good to me - thanks @ACMCMC!

PR #7728, opened 2021-11-23
Implement PublishPageRevisionAction and admin API for publishing pages

Tijani-Dia:
See #7682 

* For Python changes: Have you added tests to cover the new/fixed behaviour? Yes but may need more

### DIFF
--- a/wagtail/admin/api/actions/publish.py
+++ b/wagtail/admin/api/actions/publish.py
@@ -1,6 +1,5 @@
 from django.core.exceptions import ValidationError as DjangoValidationError
-from django.shortcuts import get_object_or_404
-from rest_framework import fields, status
+from rest_framework import status
 from rest_framework.exceptions import ValidationError
 from rest_framework.response import Response
 from rest_framework.serializers import Serializer
@@ -11,37 +10,13 @@ from wagtail.core.actions.publish_page_revision import PublishPageRevisionAction
 from .base import APIAction
 
 
-class PublishPageAPIActionSerializer(Serializer):
-    changed = fields.BooleanField(default=True, required=False)
-    log_action = fields.BooleanField(default=True, required=False)
-    previous_revision_id = fields.IntegerField(required=False)
-
-
 class PublishPageAPIAction(APIAction):
-    serializer = PublishPageAPIActionSerializer
+    serializer = Serializer
 
     def _action_from_data(self, instance, data):
-        previous_revision = None
         user = self.request.user
-        log_action = data["log_action"]
-
-        previous_revision_id = data.get("previous_revision_id", None)
-        if previous_revision_id:
-            previous_revision = get_object_or_404(
-                instance.revisions, id=previous_revision_id
-            )
-
-        revision = instance.save_revision(
-            user=user, log_action=log_action, previous_revision=previous_revision
-        )
-
-        return PublishPageRevisionAction(
-            revision,
-            user=user,
-            changed=data["changed"],
-            log_action=log_action,
-            previous_revision=previous_revision,
-        )
+        revision = instance.get_latest_revision() or instance.save_revision(user=user)
+        return PublishPageRevisionAction(revision, user=user)
 
     def execute(self, instance, data):
         try:
--- a/wagtail/admin/api/actions/publish.py
+++ b/wagtail/admin/api/actions/publish.py
@@ -1,0 +1,59 @@
+from django.core.exceptions import ValidationError as DjangoValidationError
+from django.shortcuts import get_object_or_404
+from rest_framework import fields, status
+from rest_framework.exceptions import ValidationError
+from rest_framework.response import Response
+from rest_framework.serializers import Serializer
+
+from wagtail.api.v2.utils import BadRequestError
+from wagtail.core.actions.publish_page_revision import PublishPageRevisionAction
+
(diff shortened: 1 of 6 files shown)


zerolab:
@kaedroho I won't get to this until next week, but more than happy to look at it then

wagtail/admin/api/actions/publish.py
```diff
@@ -0,0 +1,59 @@
...
+
+from wagtail.api.v2.utils import BadRequestError
+from wagtail.core.actions.publish_page_revision import PublishPageRevisionAction
+
+from .base import APIAction
+
+
+class PublishPageAPIActionSerializer(Serializer):
+    changed = fields.BooleanField(default=True, required=False)
+    log_action = fields.BooleanField(default=True, required=False)
```
kaedroho, line 16: I think we shouldn't allow changing the ``log_action`` on the admin API, since we don't currently have a way to validate that the provided action is a variant of a "publish" action. (a malicious user could pass "delete" here).

wagtail/admin/api/actions/publish.py
```diff
@@ -0,0 +1,59 @@
...
+    previous_revision_id = fields.IntegerField(required=False)
```
kaedroho, line 17: I think the way the reversion feature was implemented is quite confusing and I think we should make it a separate action later.

If everyone agrees with that. Could we leave it out of the admin API for now?

wagtail/admin/api/actions/publish.py
```diff
@@ -0,0 +1,59 @@
...
+from rest_framework.serializers import Serializer
+
+from wagtail.api.v2.utils import BadRequestError
+from wagtail.core.actions.publish_page_revision import PublishPageRevisionAction
+
+from .base import APIAction
+
+
+class PublishPageAPIActionSerializer(Serializer):
+    changed = fields.BooleanField(default=True, required=False)
```
kaedroho, line 15: I think this will always be ``False`` for a publish action on the API. Since it wouldn't be possible to edit and publish the page with the same request like it is in the page editor.

I'm not sure if it's correct for Wagtail editor to ever set this to ``True`` either since it should emit an "edit" log with the "publish" log if the content was edited when it was published.

kaedroho (changes requested):
Thanks! This is nice work!

I feel that the three parameters should be left out of the admin API for now, for different reasons. But these are fine to add to the ``PublishPageAction`` class.

wagtail/admin/api/actions/publish.py
```diff
@@ -0,0 +1,59 @@
...
+            previous_revision = get_object_or_404(
+                instance.revisions, id=previous_revision_id
+            )
+
+        revision = instance.save_revision(
+            user=user, log_action=log_action, previous_revision=previous_revision
+        )
+
+        return PublishPageRevisionAction(
+            revision,
```
kaedroho, line 39: Wonder if we should use ``instance.get_latest_revision()`` here?

wagtail/admin/api/actions/publish.py
Tijani-Dia: I agree with that. I'll remove it for now waiting to hear others' opinion.

wagtail/admin/api/actions/publish.py
Tijani-Dia: It makes sense.

wagtail/admin/api/actions/publish.py
Tijani-Dia: And use `save_revision` if there is no revision at all?

Tijani-Dia:
Thanks! I've updated it.

wagtail/admin/api/actions/publish.py
```diff
@@ -11,37 +10,13 @@
...
-        )
-
-        return PublishPageRevisionAction(
-            revision,
-            user=user,
-            changed=data["changed"],
-            log_action=log_action,
-            previous_revision=previous_revision,
-        )
+        revision = instance.get_latest_revision() or instance.save_revision(user=user)
```
kaedroho, line 18: I think the ``or`` operator will always execute both sides of the operator regardless of the value of the first one.

We should change this to:
revision = instance.get_latest_revision()
if revision is None:
    revision = instance.save_revision(user=user)

kaedroho: Actually, I might be wrong here:

```python
>>> print("foo") and print("bar")
foo
>>> print("foo") or print("bar")
foo
bar
```

wagtail/admin/api/actions/publish.py
kaedroho: Yep, thanks!

kaedroho (approved):
This looks good to me now! Would you mind merging master into this branch or rebasing it?

Tijani-Dia:
I've merged main branch here.

lb-:
@Tijani-Dia I have just rebased this branch, pushed to your remote branch, once the checks have passed I will merge in.

lb-:
@kaedroho I could not see any release notes / changelog entries for all other other API changes so I have not done them for this one either, will leave to you (I assume it will happen at the end when all are in).

@Tijani-Dia thanks for the great work here.

Tijani-Dia:
Thanks @lb- !

PR #7737, opened 2021-11-26
List blocks representation

Tijani-Dia:

* For Python changes: Have you added tests to cover the new/fixed behaviour? No
* For front-end changes: Did you test on all of [Wagtail’s supported environments]

### DIFF
--- a/wagtail/core/blocks/list_block.py
+++ b/wagtail/core/blocks/list_block.py
@@ -1,4 +1,3 @@
-import itertools
 import uuid
 
 from collections.abc import MutableSequence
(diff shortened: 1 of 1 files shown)


zerolab:
Note: this is an implementation for https://github.com/wagtail/rfcs/pull/65

gasman:
Thanks @Tijani-Dia! Merged into the feature/list-block-ids branch now, along with tests and a small update to the logic (we should check that a value matches all the requirements listed at https://github.com/wagtail/rfcs/blob/main/text/065-listblock.md#upgrading-existing-data before treating it as the new data format, to avoid the risk of matching old-style values that happen to be in a similar format).

PR #7738, opened 2021-11-29
collapse option on StreamField, StreamBlock, ListBlock

### LINKED ISSUE #7705: Auto-collapsing StreamField blocks in the admin while editing a page

My clients love the Wagtail CMS but are getting increasingly frustrated when editing their pages. The major complain is that it's very difficult to have an overview of all the StreamField blocks as their pages become more content-rich.

Previously we used [wagtail-react-streamfield](https://github.com/wagtail/wagtail-react-streamfield) which had the exact feature I'd like re-implemented—on page load it's **auto-collapsing all the StreamField blocks** which allows faster scrolling and overview.

Since wagtail-react-streamfield isn't supported any longer, there's no way to have clean overview.

Another alternative would perhaps to have a "remote control" in the top right corner of the page with in-page anchors to every StreamField block. However, this remote control would have to be updated dynamically every time a block is moved, added, or deleted.

### Additional context

Decided to create a feature request after discussion on Slack, worrying that the conversation would disappear into the void.

gasman:
Fixes #7705

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -6,6 +6,7 @@ Changelog
 
  * Added Aging Pages report (Tidjani Dia)
  * Add more SketchFab oEmbed patterns for models (Tom Usher)
+ * Add collapse option to `StreamField`, `StreamBlock`, and `ListBlock` which will load all sub-blocks initially collapsed (Matt Westcott)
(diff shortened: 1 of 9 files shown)


lb- (approved):
Looks great @gasman  - tested a bunch of cases and a few different browsers, will merge in soon.

* Checked `StreamField`
* Checked `StreamBlock`
* Checked `ListBlock`

PR #7741, opened 2021-11-30
Add UUIDs to ListBlock items (RFC 65)

gasman:
Implements [RFC 65](https://github.com/wagtail/rfcs/blob/main/text/065-listblock.md) - the database representation of a ListBlock value now saves an id alongside the value of each child. To preserve these IDs during Python-side processing, we introduce a new `ListValue` type; this behaves like a list of child values (so code that iterates over it or performs indexing on it should continue to work, provided it doesn't explicitly expect a `list` instance e.g. `isinstance(value, list)` or `value == ['foo', 'bar']`) but additionally provides a `bound_blocks` property to access the bound block objects with IDs.

Note: the RFC proposed `data` as the name of this property; I went with `bound_blocks` for consistency with StructValue.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -5,6 +5,7 @@ Changelog
 ~~~~~~~~~~~~~~~~~
 
  * Removed support for Python 3.6
+ * Added persistent IDs for ListBlock items, allowing commenting and improvements to revision comparisons (Matt Westcott, Tidjani Dia)
  * Added Aging Pages report (Tidjani Dia)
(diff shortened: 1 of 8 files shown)


kaedroho (approved):
This all looks good to me, nice work!

kaedroho:
> and presumably translatable?

This limitation only affects Wagtail Localize (and probably ``wagtail-xliff-translation`` too) since it breaks pages down into individual translatable strings rather than requiring users to retranslate the whole page every time.

PR #7746, opened 2021-12-02
Improve CircleCI frontend & backend build caches

thibaudcolas:
This changes our CircleCI backend and frontend builds to make better use of the cache.

## backend

- The backend build had caching setup but it didn’t do anything – there’s a typo (missing `s`) between the cache key used for saving and that used for restoring.
- Even just fixing the typo, it doesn’t seem like we’d save much processing by caching the pip cache in `~/.local/` only

I’ve done three changes:

- The cache key is based on a hash of `setup.py`, rather than per-branch. So the cache will only be discarded when dependencies change, rather than for each and every branch / PR.
- We cache the contents of the virtual environment after installation rather than the pip dependencies cache.
- We completely skip the installation step if the `.venv` has been populated from the cache.

I would expect this will speed up `backend` jobs by 1 minute or so – from 1min 15s installing all dependencies afresh every time, to 2-5s restoring the cache.

## frontend

Changes aren’t so dramatic there since the cache was set up correctly. I’ve only made two of the above changes:

- Caching based on `package-lock.json` hash only, rather than per branch. The lockfile is the only thing that matters for `npm install`, so segregating caches per branch doesn’t achieve anything useful.
- We can also completely skip the installation step if `node_modules` was restored from cache.

In this case I’d expect the savings to be on the order of 30s per build – from 45s restoring cache, installing, saving cache – to 10-15s restoring the cache only.

---

Sample build with both improvements visible (on a fork): https://app.circleci.com/pipelines/github/thibaudcolas/wagtail-fork/58/workflows/59d0fe54-6ca7-4420-b195-ecbad3804f3e.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,20 +36,20 @@ jobs:
     steps:
       - checkout
       - restore_cache:
-          keys:
-            - node-v1-{{ .Branch }}-{{ checksum "package-lock.json" }}
-            - node-v1-{{ .Branch }}-
-            - node-v1-
-      - run: npm install --no-save
+          key: frontend-v1-{{ checksum "package-lock.json" }}
+      # Only install if node_modules wasn’t cached.
+      - run: |
+            if [[ ! -e "node_modules" ]]; then
+                npm install --no-save --no-optional --no-audit --no-fund --progress=false
+            fi
       - save_cache:
           paths:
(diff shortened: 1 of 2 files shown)


lb- (commented):
@thibaudcolas looks great, approved and merged.

PR #7749, opened 2021-12-02
Use translations for permission names

exler:
Adding translation tags for mechanism created in https://github.com/wagtail/wagtail/pull/7488.


gasman:
I don't think this will work - the `{% trans %}` tag needs to be passed a string literal, not a variable, in order to be picked up by the `manage.py makemessages` command and added to the translation dictionary. As a result, when it encounters one of these tags at runtime, it will try to look up a string such as "Can access Wagtail admin" in the translation dictionary, but no translation will be available.

As per https://groups.google.com/g/django-developers/c/ZzeKqqkUC3g, Django currently has no mechanism for translating permission labels - since they're stored in the database rather than code - and until that happens I don't think it'll be feasible to provide translations within Wagtail either.

Issue #5271, opened 2019-05-01
Feature request: UI testing environment


thibaudcolas:
Working on Wagtail’s admin UI, I find it unnecessarily hard to make sure that I take all of the variations in UI components into account. It's a problem with all projects that have fairly big UIs, but is compounded in Wagtail by the fact that so much of the UI is dependent on what content has been created in the CMS, and on how the site has been set up. It also doesn't help that Wagtail's default dev environment is first meant to be a good demo of how Wagtail would be used on a small-scale website, rather than a dev playground (https://github.com/wagtail/bakerydemo/issues/175).

Having a more comprehensive UI "test" environment would help in a number of scenarios:

- Reviewing the existing UI, for example as part of a UX review, accessibility testing, compliance, or just when testing the impact of a specific change.
- Changing existing UI components, and reviewing changes. It's really hard to know what impact a UI change might have, if you have to know how to use the CMS for bits of its UI to appear. This is particularly problematic when multiple people work on the same feature, and they all have to do the same error-prone setup on their local copies of Wagtail for them to test their changes meaningfully
- Setting up automated UI testing in continuous integration – accessibility testing (https://github.com/wagtail/wagtail/issues/4871#issuecomment-486648541), visual regression testing. These are impossible to set up if we don't have a standardised test environment to build the test suites for.

Related: #3804, https://github.com/wagtail/bakerydemo/issues/175, https://github.com/wagtail/wagtail/issues/4871#issuecomment-486648541, https://github.com/wagtail/rfcs/pull/37

---

This was previously discussed at a Wagtail team meeting, and the conclusion there was that it should be done as a separate environment / project, rather than extending `bakerydemo`. It's not entirely clear how this project could then be used in the Wagtail CI builds. Thoughts welcome!

jonnyscholes:
I'm super keen for this too. Working on refactoring some of our older JS, and some of it really could use with some tests.

I'm keen to take a stab at this as part of that refactor as now really is the right time for me to be writing tests whilst I know about the edge cases. Rough thoughts on a path forward:

- Take bakerydemo as a starting point
- Remove vagrant but keeping the docker setup with the intention of running our tests inside docker
- Keep the models as a starting point, but implement testdata in a way that can easily be modified (probably by creating some JSON of just the data, and the importing the models and programatically creating them)

thibaudcolas:
[...]

PR #7751, opened 2021-12-02
Automated browser accessibility test suite in CircleCI

thibaudcolas:
Depends on #7746. Addresses #4871 and #5271, by making Wagtail’s existing `wagtail/tests` suitable for UI integration tests. Compared to other approaches we’ve tried in the past, this has the immense advantage that the test suite is fully contained within the repository – so the tests can run in pull requests with little magic needed.

I chose [Puppeteer](https://github.com/puppeteer/puppeteer) for the browser automation, because it’s very popular and battle-tested, particularly on CircleCI and with Jest, which we use already.

- Puppeteer has a [Percy](https://docs.percy.io/docs/puppeteer-tutorial) (from Browserstack) integration, if/when we want to do visual regression tests.
- It also has good [Axe](https://github.com/WordPress/gutenberg/tree/trunk/packages/jest-puppeteer-axe) integrations, which I’m making use of here.

This PR contains:

- A baseline end-to-end browser integration test setup with Puppeteer & Jest
- Enough configuration changes to the `wagtail/tests` suite to make it suitable for those browser tests in CI
- CircleCI configuration to run the new test suite for all pull requests
- Automated accessibility tests with Axe for a selection of Wagtail views
- Basic documentation on how to get this test suite running.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,7 +83,7 @@ jobs:
       # Only install if node_modules wasn’t cached.
       - run: |
             if [[ ! -e "client/tests/integration/node_modules" ]]; then
-                npm --prefix ./client/tests/integration install --no-save --no-optional --no-audit --no-fund --progress=false
+                npm --prefix ./client/tests/integration ci
             fi
       - save_cache:
           key: ui_tests-npm_integration-v1-{{ checksum "client/tests/integration/package-lock.json" }}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,10 +47,65 @@ jobs:
             - node_modules
           key: frontend-v1-{{ checksum "package-lock.json" }}
       - run: npm run dist
+      # Save static files for subsequent jobs.
+      - persist_to_workspace:
+          root: ~/project
+          paths:
+            - wagtail
       - run: npm run lint:js
       - run: npm run lint:css
       - run: npm run test:unit:coverage -- --runInBand
       - run: bash <(curl -s https://codecov.io/bash) -F frontend
+
+  ui_tests:
+    docker:
+      - image: cimg/python:3.8.11-browsers
+    environment:
+      PIPENV_VENV_IN_PROJECT: true
+      DJANGO_SETTINGS_MODULE: wagtail.tests.settings_ui
+    steps:
+      - checkout
+      - attach_workspace:
+          at: ~/project
+      - restore_cache:
+          key: pipenv-v1-{{ checksum "setup.py" }}
+      # Only install if .venv wasn’t cached.
+      - run: |
+            if [[ ! -e ".venv" ]]; then
+                pipenv install -e .[testing]
+            fi
+      - save_cache:
+          key: pipenv-v1-{{ checksum "setup.py" }}
+          paths:
+            - .venv
+      - restore_cache:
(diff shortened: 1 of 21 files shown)


.circleci/config.yml
```diff
@@ -32,22 +36,77 @@ jobs:
...
-          keys:
-            - node-v1-{{ .Branch }}-{{ checksum "package-lock.json" }}
-            - node-v1-{{ .Branch }}-
-            - node-v1-
-      - run: npm install --no-save
+          key: frontend-v1-{{ checksum "package-lock.json" }}
+      # Only install if node_modules wasn’t cached.
+      - run: |
+            if [[ ! -e "node_modules" ]]; then
+                npm install --no-save --no-optional --no-audit --no-fund --progress=false
```
fabienheureux, line 43: Is there a reason you don't use `npm ci` here ?

client/tests/integration/editbird.test.js
```diff
@@ -0,0 +1,16 @@
+describe.skip('Editbird', () => {
+  beforeAll(async () => {
+    await page.goto('http://localhost:8000/');
```
fabienheureux, line 3: It could be useful to have wagtail's base url in an environment variable. 
I often expose wagtail's port on another port than 8000 in order to try things.

I am not sure if it is relevant here though

fabienheureux (commented):
I asked two question, however it looks good 👍 
I did not run it locally yet.

client/tests/integration/editbird.test.js
thibaudcolas: Good point, definitely!

.circleci/config.yml
thibaudcolas: Last time I tried it, it was slower than this command, which does the same thing anyway when there isn’t a `node_modules` folder. I’ll try again and see how they compare these days.

fabienheureux: I think - it might have changed though - that `npm ci` throws an error if one of the dependencies is missing from the `package-lock.json` file. 

I switched to `npm ci` in a project - a while back - because a minor version of an npm package was removed from the registry and the CI installed the next version, causing difference between local package version and the one installed in the CI, introducing bugs on production that were not present locally. 
However, it was also caused by badly fixed version in the `package.json`...

fabienheureux: https://docs.npmjs.com/cli/v8/commands/npm-ci
> If dependencies in the package lock do not match those in package.json, npm ci will exit with an error, instead of updating the package lock.

Well, it seems it is still the case. However better speed might be more important in that case than covering a situation that would almost never happen.

client/tests/integration/editbird.test.js
thibaudcolas: I’ve added a `TEST_ORIGIN=http://localhost:9000` environment variable so we can both test on different ports when running locally, and test remote Wagtail instances.

.circleci/config.yml
thibaudcolas: From quick timing, `npm ci` was 1s faster. Time to switch 😄

thibaudcolas:
Now updated with feedback from Fabien, and full documentation on how to run this locally.

lb-:
@thibaudcolas  - #7746 is merged in, this will need to be rebased/conflicts resolved.

thibaudcolas:
Thank you @lb-! All rebased and working.

Issue #7754, opened 2021-12-03
simple_translation does not use latest revision of draft document


gnomeby:
### Issue Summary

When I translate a draft document with only draft revisions the first draft revision will be used for making a copy. So simple_translation does not use latest revision of draft document

### Configure a project

1. Start a new project with `wagtail start myproject`
2. Enable wagtail.locales and wagtail.contrib.simple_translation apps
3. Configure languages

```
WAGTAIL_I18N_ENABLED = True
WAGTAIL_CONTENT_LANGUAGES = LANGUAGES = [
    ('en', "English"),
    ('ru', "Russian"),
]
```

4. `./manage.py migrate`
5. `./manage.py createsuperuser`
6. `./manage.py runserver`
7. Login to admin panel
8. Add second language
9. Translate Homepage and publish it for the second language

### Steps to Reproduce
1. Add child page for the english Homepage, set title "Title1" and save it as draft
2. Edit the new child page and change the title to "Title2" and save it as draft
3. Translate the new child page to the second language

Result: A copy of the child page will have the "Title1"
Expected: A copy of the child page must have the "Title2"

### Technical details

* Python version: 3.8
* Django version: 3.2.9
* Wagtail version: 2.14.2
* Browser version: Any

gnomeby:
A possible fix https://github.com/wagtail/wagtail/pull/7755

PR #7755, opened 2021-12-03
Use the latest draft for translation (#7754)

gnomeby:
Fix for #7754


lb-:
@gnomeby thanks for taking the time to contribute a fix for #7754 - In principle this looks good but we will get another review from a core team member also.

However, we ask that you take a bit more time to contribute a unit test that covers this scenario with the fix. This way it will be much harder to break this in the future.

gnomeby:
OK, I'll add them.

allcaps:
@gnomeby Thank you for your contribution to Wagtail. The code looks sound. However, I doubt this is the desired behaviour. 

Why is copying the last revision more correct than copying the current page? Copying the last revision might come as a  surprise to some users. Or am I missing something? Is there a scenario I did not consider? Why is the the latest revision a better source for a translation than the current page?

Revisions are the last 'draft' saved by anyone. That doesn't mean the content is more correct than the current page. The revision might be a rough -and failed- rewrite of the page. That should not be published. Ever. Let alone be translated. I consider the current page as the most correct content. The content most suitable to be translated.

TLDR; The latest revision doesn't have to be 'more correct' than current page.

I'm not 100% sure about my own argumentation, so I'm happy to convinced otherwise.

FYI, simple translation is a perfect starting point for your own translation app. Copy `simple_translation` from `site-packages/wagtail/contrib/simple_translation` into your project root and apply your changes. Change installed apps "wagtail.contrib.simple_translation" to "simple_translation". You can take it a step further, add a choice field to the form to ask the user which 'source' to use.

allcaps (commented):
I don't think last revision is 'better' content than the current page. A revision might even be a failed draft. Therefore not a suitable source for translations.

gnomeby:
@allcaps please take a look to description of issue:
https://github.com/wagtail/wagtail/issues/7754 
I'm trying to translate document with only draft revisions.
But for the case when we have published+draft versions you are right, my fix is not suitable and should be improved.

gasman:
I think this might be related to the common "gotcha" with pages that have never been published, seen most recently in https://github.com/wagtail/wagtail/issues/7690#issuecomment-1004975409 - if a page has never been published, then the 'live' page record will reflect the state of the page when it was first created and saved for the first time - subsequent 'save draft' actions will only affect page revisions, so it will never move on from that state. In this case, the outcome is less desirable (and more surprising) than selecting the latest draft revision.

Detecting pages that have _never_ been published is a bit of a black art, so I'd suggest that we also select the latest draft for pages that were published and then unpublished. (In these cases, the 'live' page record will be frozen at the last published version, which may be far out of date.)

In other words - when copying a page for translation, I'd suggest the following logic:

* does the page's real database record have `live=True`?
  * if yes, copy the page data from the real (live) database record
  * if no, copy the page data from the latest (draft) revision

gnomeby:
@gasman agree

gnomeby:
I've updated PR using @gasman advice

lb-:
There has been some more recent work on how non-saved drafts are treated. This needs some rework and maybe some discussion before it's able to be reviewed again. Flagging as needs work / needs design decision.

wagtail/actions/copy_for_translation.py
```diff
@@ -146,7 +148,11 @@ def execute(self, skip_permission_checks=False):
         self.check(skip_permission_checks=skip_permission_checks)
 
         translated_page = self._copy_for_translation(
-            self.page, self.locale, self.copy_parents, self.alias, self.exclude_fields
+            self.page.live and self.page or self.page.get_latest_revision_as_page(),
```
gasman, line 151: `self.page if self.page.live else self.page.get_latest_revision_as_page()` is the more familiar syntax I think, but happy with this logic.

gasman (approved):

gasman:
Rebased and merged in 32c48e149d83f8ed8157372be0ec41098b296c70 - thanks @gnomeby!

